Issue #2515, opened 2016-10-07
google-cloud-speech pip install not working


puneith:
pip is unable to find google-cloud-speech 

$ pip install --upgrade google-cloud-speech
Collecting google-cloud-speech
  Could not find a version that satisfies the requirement google-cloud-speech (from versions: )
No matching distribution found for google-cloud-speech

Same is true when I try to install google-cloud

$ python -c "from google.cloud import speech"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: cannot import name speech

dhermes:
@puneith We have not yet released a version of `speech` yet nor is it yet included in the umbrella `google-cloud` package (notice it is not listed on https://pypi.python.org/pypi/google-cloud/0.20.0). @daspecster may have more to say there?

daspecster:
@puneith I've been working on adding the `stream_recognize()` method and implementing GRPC support for `sync_recognize()` and `async_recognize()` before it's fully released.

You could play with speech by installing the library from source. If you have any questions just let me know!

Thanks @dhermes!

gw00207:
@dhermes In the [docs](https://googlecloudplatform.github.io/google-cloud-python/stable/speech-usage.html) there is a Speech API section which does say you can use:

```
>>> from google.cloud import speech
>>> client = speech.Client()
```

However I get the same error as puneith.  @daspecster do you know how soon Speech will be implemented in google-cloud-python? Thanks :-)

dhermes:
@gw00207 It's alread (partially) "implemented in google-cloud-python". You can install from source and play with the current implementation:

```
$ git clone https://github.com/GoogleCloudPlatform/google-cloud-python
$ cd google-cloud-python/speech
$ python setup.py install
```

daspecster:
@gw00207, we have a list of things to do for speech before we release. You can follow along on #2522.

mlissner:
Can we please remove this from the docs until #2522 is done or put a loud note to this effect? I just spent time:
 - trying to figure out why the import in the docs was failing
 - realizing that I needed to install a special package
 - learning that that package was failing to install
 - finding my way here after googling.

A simple notice would have prevented all this.

gw00207:
@dhermes I'm trying to run the speech API inside a GCE VM, can I run those commands from the Cloud Shell? Thanks.

PR #2620, opened 2016-10-26
Add message about API not being available yet.

daspecster:
A small message stating that `google-cloud-speech` isn't released yet.

See #2515.

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -7,6 +7,12 @@ base.
 
 .. warning::
 
+    This API, ``google-cloud-speech`` has not been released yet. In order to
+    test and explore this feature, you must install ``google-cloud`` from
+    source.
+
+.. warning::
+
     This is a Beta release of Google Speech API. This
     API is not intended for real-time usage in critical applications.
 


tseaver:
LGTM.  You need to add reverting it to your "release speech" checklist somewhere. ;)

daspecster:
@tseaver thanks! I added it here https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2522

dhermes:
@daspecster You should also remove the RST docs from the build (you could just delete them and put them back, you'll need to temporarily ignore the modules in the `verify_included_modules.py` file as well.

daspecster:
So you would opt for not having the warning but just removing the docs until the release?

dhermes:
I (stupidly) assumed you updated the `README`, not the RST file. My bad.

I'd say it'd be better not to have the docs show up before release (especially in our `/stable/` docs) but it's your call.

daspecster:
Makes sense, I think this is good to hash out moving forward as well. As new services are added it would be good to make it clear for users.

Issue #1838, opened 2016-05-31
CORS documentation missing setter


pdknsk:
https://googlecloudplatform.github.io/gcloud-python/stable/storage-buckets.html#gcloud.storage.bucket.Bucket.cors

> Retrieve CORS policies configured for this bucket.

It doesn't mention that it can also be set.

pdknsk:
(I think this also applies to a few other properties.)

daspecster:
@tseaver https://github.com/GoogleCloudPlatform/gcloud-python/blob/master/gcloud/storage/bucket.py#L509 is it just not showing up in the docs because of the decorator?

tseaver:
@daspecster Docstrings for `@property` objects are derived from their getters:

```python
>>> class Foo(object):
...     @property
...     def bar(self):
...         """bar getter"""
...     @bar.setter
...     def bar(self, value):
...         """bar setter"""
...     @bar.deleter
...     def bar(self):
...         """bar deleter"""
... 
>>> Foo.bar
<property object at 0x7f8fe9a856d8>
>>> Foo.bar.__doc__
'bar getter'
```

daspecster:
Cool, what if we just combine the docs in some way that makes sense in the getter?

```python
 @property
    def cors(self):
        """Get/Set CORS policies configured for this bucket.
        See: http://www.w3.org/TR/cors/ and
             https://cloud.google.com/storage/docs/json_api/v1/buckets

        :setter: Set CORS policies for this bucket.
        :getter: Gets the CORS policies for this bucket.

        """
        return [copy.deepcopy(policy)
                for policy in self._properties.get('cors', ())]

    @cors.setter
    def cors(self, entries):
        """Set CORS policies configured for this bucket.
        See: http://www.w3.org/TR/cors/ and
             https://cloud.google.com/storage/docs/json_api/v1/buckets
        :type entries: list of dictionaries
        :param entries: A sequence of mappings describing each CORS policy.
        """
        self._patch_property('cors', entries)
```

mlissner:
I filed #2610 where the same issue applies to the setter for `lifecycle_rules`. Seems like a very simple solution here is what @daspecster proposed. It's not perfect, but if it had been done back in June when it was proposed, I would have been a happier customer.

daspecster:
@mlissner, sorry you lost cycles on this.

[...]

Issue #2610, opened 2016-10-25
Setting lifecycle rules totally undocumented


mlissner:
It took me quite a while to figure out how to set lifecycle rules. If you look at the code, it's undocumented:

```
@lifecycle_rules.setter
def lifecycle_rules(self, rules):
    self._patch_property('lifecycle', {'rule': rules})
```

I found [some tests for this](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/090c5d583da82daabf05c52e8f72b858a0bb09c2/storage/unit_tests/test_bucket.py#L588-L596), but even those don't seem to use the `lifecycle_rules` method. Not having a documented method for doing this is very frustrating since this is supposed to be a library that supports all of the features.

daspecster:
@mlissner, thanks for pointing this out.

This library is still under heavy development and hasn't reached 1.0 yet, which is why some areas are missing or light on documentation. There are also a number of features that aren't yet supported by this library.

dhermes:
@mlissner That was the setter, you want the [getter][1]. Is your concern that the [documentation][2] for the property is insufficient / doesn't indicate that there **is** a setter?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/b6fe01ec4934d7744b5df8a7e64531f4c1386a9e/storage/google/cloud/storage/bucket.py#L563-L571
[2]: https://googlecloudplatform.github.io/google-cloud-python/0.20.0/storage-buckets.html#google.cloud.storage.bucket.Bucket.lifecycle_rules

tseaver:
In fact, the `@property` descriptor explicitly takes the `__doc__` from the getter:  any docstring on from the setter or deleter are not discoverable at runtime.

mlissner:
> Is your concern that the documentation for the property is insufficient / doesn't indicate that there is a setter?

Yes, exactly. I want to set the lifecycle rules for a bucket, but there's no documentation of how to do so.

mlissner:
OK, I don't know where else to put this, so I'm putting it here. I've struggled mightily to figure out this API and I hope this feedback will be helpful. Two things:

1. To set the `lifecycle_rules`, you need to set them before you create the bucket:

        b = client.bucket('my-bucket-name')
        b.exists()  # False
        b.lifecycle_rules = [{
            'action': {'type': 'Delete'},
            'condition': {'age': 1},
        }]
        b.create()

[...]

PR #2623, opened 2016-10-26
Mention that cors property has a setter.

daspecster:
Bucket CORS property docs only say retrieve but it actually has a setter as well.

See #1838.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -511,10 +511,13 @@ class Bucket(_PropertyMixin):
 
     @property
     def cors(self):
-        """Retrieve CORS policies configured for this bucket.
+        """Retrieve or set CORS policies configured for this bucket.
 
         See: http://www.w3.org/TR/cors/ and
              https://cloud.google.com/storage/docs/json_api/v1/buckets
+
+        :setter: Set CORS policies for this bucket.
+        :getter: Gets the CORS policies for this bucket.
 
         :rtype: list of dictionaries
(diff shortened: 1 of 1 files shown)


mlissner:
Do you want to add docs for the `lifecycle_rules` setter too, as in #2610?

daspecster:
@mlissner, I think that will be a separate PR actually.

storage/google/cloud/storage/bucket.py
```diff
@@ -511,11 +511,14 @@ def rename_blob(self, blob, new_name, client=None):
...
     @property
     def cors(self):
-        """Retrieve CORS policies configured for this bucket.
+        """Retrieve or set CORS policies configured for this bucket.
 
         See: http://www.w3.org/TR/cors/ and
              https://cloud.google.com/storage/docs/json_api/v1/buckets
 
+        :setter: Set CORS policies for this bucket.
+        :getter: Gets the CORS policies for this bucket.
```
dhermes: Are these valid Sphinx directives?

tseaver:
@dhermes There is a [SO answer](http://stackoverflow.com/questions/17478040/how-can-i-generate-documentation-for-a-python-property-setter-using-sphinx) which shows `:getter:` and `:setter:` being rendered.

daspecster:
I'm generating the docs real quick and I'll screenshot it.

dhermes:
Got it, so they aren't official directives but they are "generally supported" syntax

daspecster:
@dhermes @tseaver is this good to merge?

dhermes:
:+1:

PR #2628, opened 2016-10-26
Add AVRO as a valid BigQuery SourceFormat

jeffk:
Fixes GoogleCloudPlatform/google-cloud-python#2627, and brings the function up to date with the [BigQuery Documentation](https://cloud.google.com/bigquery/loading-data#loading_avro_files).


jeffk:
I signed it!

tseaver:
@jeffk Thanks for the patch!  Interesting that we had support for it for the `desitnationFormat` field for extracts, but not the `sourceFormat` field for loads:  I almost feel like we've been gaslighted again.

PR #2633, opened 2016-10-27
Converting Pub/Sub client->list_subscriptions to iterator.

dhermes:
There is surprisingly no system test for `Client.list_subscriptions()`. After seeing how different the implementations of `Client.list_subscriptions()` and `Topic.list_subscriptions()` are, I am strongly for adding such a system test.

~~Sending this out without fixing unit tests to get it in the hands of reviewers. The unit tests may take me a non-trivial time, but I am working on them.~~

### DIFF
--- a/pubsub/google/cloud/pubsub/_gax.py
+++ b/pubsub/google/cloud/pubsub/_gax.py
@@ -218,8 +218,9 @@ class _PublisherAPI(object):
         iter_kwargs = {}
         if page_size:  # page_size can be 0 or explicit None.
             iter_kwargs['max_results'] = page_size
-        iterator = GAXIterator(self._client, page_iter,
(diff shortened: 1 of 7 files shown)


dhermes:
@tseaver @daspecster PTAL

pubsub/google/cloud/pubsub/_gax.py
```diff
@@ -313,7 +315,7 @@ def subscription_create(self, subscription_path, topic_path,
             if exc_to_code(exc.cause) == StatusCode.FAILED_PRECONDITION:
                 raise Conflict(topic_path)
             raise
-        return _subscription_pb_to_mapping(sub_pb)
+        return MessageToDict(sub_pb)
```
daspecster: 👍

PR #2636, opened 2016-10-28
Converting Logging client->list_entries to iterator.

dhermes:
@waprin @jonparrott I just realized how weird the logging HTTP/JSON API is. For `/entries:list`, [POST is used](https://cloud.google.com/logging/docs/api/reference/rest/v2/entries/list) instead of GET, and the "query params" come in the payload instead of as query params.

~~Just as #2633 started, I'm sending this out without fixing unit tests to get it in the hands of reviewers. The unit tests may take me a non-trivial time, but I am working on them.~~

### DIFF
--- a/core/google/cloud/iterator.py
+++ b/core/google/cloud/iterator.py
@@ -298,6 +298,7 @@ class HTTPIterator(Iterator):
     _PAGE_TOKEN = 'pageToken'
     _MAX_RESULTS = 'maxResults'
     _RESERVED_PARAMS = frozenset([_PAGE_TOKEN, _MAX_RESULTS])
+    _HTTP_METHOD = 'GET'
 
     def __init__(self, client, path, item_to_value,
                  items_key=DEFAULT_ITEMS_KEY,
@@ -378,9 +379,19 @@ class HTTPIterator(Iterator):
         :rtype: dict
         :returns: The parsed JSON response of the next page's contents.
         """
-        return self.client.connection.api_request(
-            method='GET', path=self.path,
-            query_params=self._get_query_params())
+        params = self._get_query_params()
+        if self._HTTP_METHOD == 'GET':
+            return self.client.connection.api_request(
+                method=self._HTTP_METHOD,
+                path=self.path,
+                query_params=params)
+        elif self._HTTP_METHOD == 'POST':
+            return self.client.connection.api_request(
+                method=self._HTTP_METHOD,
+                path=self.path,
+                data=params)
+        else:
+            raise ValueError('Unexpected HTTP method', self._HTTP_METHOD)
 
 
 class GAXIterator(Iterator):
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -1,0 +1,48 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Common logging helpers."""
+
+
+from google.cloud.logging.entries import ProtobufEntry
+from google.cloud.logging.entries import StructEntry
+from google.cloud.logging.entries import TextEntry
+
+
+def entry_from_resource(resource, client, loggers):
+    """Detect correct entry type from resource and instantiate.
+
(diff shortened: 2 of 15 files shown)


jonparrott:
iirc logging has some very strange behavior with listing entries, I think @waprin is pretty familiar with it.

waprin:
Yes specifically it uses the page tokens as a progress indicator, not just as a means of breaking results into pages. This means you can get 0 results on first page, but then 10 results on the next page, so you really always want to check the page token. This is especially true for slower queries, which are most of them without an index e.g. logName=  .

dhermes:
@tseaver @daspecster PTAL

dhermes:
@waprin So what are the implications for using an Iterator? It'll just take a long time consuming empty pages until the results are available?

waprin:
@dhermes yeah, if anything I'm thinking this is a very good change because with the iterator people are less likely to get no results on the first page and think there are no results.

jonparrott:
@waprin is there a possibility the iterator will never finish?

logging/google/cloud/logging/_helpers.py
```diff
@@ -0,0 +1,48 @@
...
+from google.cloud.logging.entries import ProtobufEntry
+from google.cloud.logging.entries import StructEntry
+from google.cloud.logging.entries import TextEntry
+
+
+def entry_from_resource(resource, client, loggers):
+    """Detect correct entry type from resource and instantiate.
+
+    :type resource: dict
+    :param resource: one entry resource from API response
```
daspecster: Nit, but there are a few :param:'s that could be spruced up.
```diff
- one entry resource from API response
+ One entry resource from API response.
```

logging/google/cloud/logging/_helpers.py
```diff
@@ -0,0 +1,48 @@
...
+    :returns: The entry instance, constructed via the resource
+    """
+    if 'textPayload' in resource:
+        return TextEntry.from_api_repr(resource, client, loggers)
+    elif 'jsonPayload' in resource:
+        return StructEntry.from_api_repr(resource, client, loggers)
+    elif 'protoPayload' in resource:
+        return ProtobufEntry.from_api_repr(resource, client, loggers)
+
+    raise ValueError('Cannot parse log entry resource')
```
daspecster:
```diff
- 'Cannot parse log entry resource'
+ 'Cannot parse log entry resource.'
```

logging/unit_tests/test__gax.py
```diff
@@ -97,33 +115,47 @@ def _list_entries_with_paging_helper(self, payload, struct_pb):
...
         SIZE = 23
         TOKEN = 'TOKEN'
         NEW_TOKEN = 'NEW_TOKEN'
         resource_pb = MonitoredResource(type='global')
-        timestamp_pb = _datetime_to_pb_timestamp(
-            datetime.datetime.utcnow())
-        entry_pb = LogEntry(log_name=self.LOG_NAME,
+        timestamp = datetime.datetime.utcnow().replace(tzinfo=UTC)
+        timestamp_pb = _datetime_to_pb_timestamp(timestamp)
+        entry_pb = LogEntry(log_name=self.LOG_PATH,
```
daspecster: `log_name=self.LOG_PATH`?
I know this is from [here](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2636/files#diff-41b5ab9e98931dbeb0d6ea726a819a1eR42), but it just reads in an unclear way.

Seeing how these are used later on, I'm not really sure I have any good suggestions on how to make this clearer.

core/google/cloud/iterator.py
```diff
@@ -378,9 +379,19 @@ def _get_next_page_response(self):
...
+        if self._HTTP_METHOD == 'GET':
+            return self.client.connection.api_request(
+                method=self._HTTP_METHOD,
+                path=self.path,
+                query_params=params)
+        elif self._HTTP_METHOD == 'POST':
+            return self.client.connection.api_request(
+                method=self._HTTP_METHOD,
+                path=self.path,
+                data=params)
```
tseaver: I'm waiting for one of the back-end developers to release an API which uses `POST` with query strings.

core/google/cloud/iterator.py
```diff
@@ -378,9 +379,19 @@ def _get_next_page_response(self):
         :rtype: dict
         :returns: The parsed JSON response of the next page's contents.
         """
-        return self.client.connection.api_request(
-            method='GET', path=self.path,
-            query_params=self._get_query_params())
+        params = self._get_query_params()
+        if self._HTTP_METHOD == 'GET':
```
dhermes: Not really. The `HTTPIterator` makes a "list" type HTTP request. Previously it was **always** a GET request but now it is **sometimes** a GET request. The base connection doesn't do anything other than `api_request` and it's up to the caller to provide the HTTP method.

waprin:
@jonparrott can't think of a way

dhermes:
@tseaver LGTY to merge?

dhermes:
As discussed with @daspecster, going to make the docstring updates in a follow-up PR

PR #2637, opened 2016-10-28
Actual type_url URI is type.googleapis.com.

daspecster:
Usually `prefix` is passed to `def _compute_type_url(klass, prefix)` so this should be a low/no impact change.

However I don't see where Bigtable passes `prefix`...

### DIFF
--- a/core/google/cloud/operation.py
+++ b/core/google/cloud/operation.py
@@ -17,7 +17,7 @@
 from google.longrunning import operations_pb2
 
 
-_GOOGLE_APIS_PREFIX = 'types.googleapis.com'
+_GOOGLE_APIS_PREFIX = 'type.googleapis.com'
 
 _TYPE_URL_MAP = {
 }


tseaver:
@daspecster The `_GOOGLE_APIS_PREFIX` constant is only used as the default value for the `prefix` argument to `_compute_type_url`.  Outside its own tests, that helper is called only to generate type URLs for Bigtable's `UpdateClusterMetadata` and `CreateClusterMetadata`:  the effect of borking the URL is that callers won't be able to fetch convert the metadata for those operations.

tseaver:
LGTM

daspecster:
@tseaver it doesn't look like there are system tests for Bigtable Clusters?
I checked the `type_url`s for the Bigtable system tests but I'm just realizing that there are no cluster type_urls when I looked before.

I also don't see `CreateClusterMetadata` in the docs https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.CreateClusterRequest.

daspecster:
I'm going to take your LGTM though. Googling for `types.googleapis.com` comes up with another typo I believe.

dhermes:
@daspecster It'd help if you posted a snippet that verified for Bigtable, similar to https://gist.github.com/dhermes/09c964d6d27003ae817b650424fda7c3 for speech

daspecster:
Ok, I'll see what I can do for that.

dhermes:
It doesn't have to be complex / overly reproducible. Whatever you used to show me

```
response {
 type_url: "type.googleapis.com/google.bigtable.admin.v2.Instance"
 value: "\n2projects/ferrous-arena/instances/new-1477676796867\022\021new-1477676796867\030\001"
}
```

from before will be fine.

daspecster:
I put `import pdb; pdb.set_trace()` after this line
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/40932e9307066987bf7e325bbc87b267d6d22f0e/core/google/cloud/operation.py#L104
and checked the value of `type_value` from running `tox -e system-test -- bigtable`.

dhermes:
Ha, I like it!

daspecster:
😄

PR #2638, opened 2016-10-28
Clean up async docs and unify for future GAPIC async.

daspecster:
@dhermes mentioned this in a previous review.
This is correct and I also cleaned up the docs.

In my testing with async, I haven't gotten more than one alternative to show up, even whith `max_alternatives=2`. So that's why there isn't two in the doc example.

### DIFF
--- a/speech/google/cloud/speech/operation.py
+++ b/speech/google/cloud/speech/operation.py
@@ -123,8 +123,9 @@ class Operation(operation.Operation):
         raw_results = response.get('response', {}).get('results', None)
         results = []
         if raw_results:
-            for result in raw_results[0]['alternatives']:
-                results.append(Transcript.from_api_repr(result))
+            for result in raw_results:
+                for alternative in result['alternatives']:
+                    results.append(Transcript.from_api_repr(alternative))
         if metadata:
             self._metadata = Metadata.from_api_repr(metadata)
 
--- a/speech/unit_tests/test_operation.py
+++ b/speech/unit_tests/test_operation.py
@@ -36,7 +36,7 @@ class OperationTests(unittest.TestCase):
 
     def test_from_api_repr(self):
         from unit_tests._fixtures import OPERATION_COMPLETE_RESPONSE
-        from google.cloud.speech.operation import Transcript
+        from google.cloud.speech.transcript import Transcript
         from google.cloud.speech.metadata import Metadata
         RESPONSE = OPERATION_COMPLETE_RESPONSE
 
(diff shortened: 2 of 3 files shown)


speech/google/cloud/speech/operation.py
```diff
@@ -123,8 +123,9 @@ def _update(self, response):
         raw_results = response.get('response', {}).get('results', None)
```
dhermes: Future PR: Let's collapse some / all of this into `google.cloud.operation.Operation`. This class is for the HTTP version of a long-running operation, but it's still the exact same proto: https://github.com/googleapis/googleapis/blob/050400df0fdb16f63b63e9dee53819044bffc857/google/longrunning/operations.proto#L80

It should be fine since it's already a subclass?

speech/google/cloud/speech/result.py
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Representation of a Speech API Result."""
+
+from google.cloud.speech.transcript import Transcript
+
+
+class SpeechResult(object):
```
dhermes: I am :-1: on this class. The [proto][1] **only** defines the `alternatives` fields, we could achieve the same effect by just returning a list.

[1]: https://github.com/googleapis/googleapis/blob/1eb6f564981cd745495ebb0c299e7deda801184b/google/cloud/speech/v1beta1/cloud_speech.proto#L391-L395

daspecster: If we do that, then I will have to traverse all the alternatives in all the results and make them `Transcript` objects.

```python
results = []
for result in raw_results:
    for alternative in result['alternatives']:
        alternatives.append(Transcript.from_api_repr(alternative))
    results.append({'alternatives': alternatives})
```

However that forces the docs to access the alternatives with `__getitem__` instead of it being a property.

daspecster: Could do something like this I guess...
```python
result = lambda: None
result.alternatives = Transcript.from_api_repr(alternative)
```

dhermes: I am :-1: on us preserving `alternatives`. Your helper class still traverses all the alternatives and makes them `Transcript` objects. Why do you need `results.append({'alternatives': alternatives})`?

daspecster: My goal is to match the usage of the JSON version of the API to the gRPC version of the API so they're the same.

Also this pattern logically follows the other two methods(`sync_recognize`, `streaming_recognize`).

speech/google/cloud/speech/operation.py
```diff
@@ -123,8 +123,9 @@ def _update(self, response):
...
         results = []
         if raw_results:
-            for result in raw_results[0]['alternatives']:
-                results.append(Transcript.from_api_repr(result))
+            for result in raw_results:
+                for alternative in result['alternatives']:
+                    results.append(Transcript.from_api_repr(alternative))
```
dhermes: @daspecster I finally understand what you were worried about: a [list][1] inside a (different) [list][2]. However, I still think this is the right approach.

[1]: https://github.com/googleapis/googleapis/blob/050400df0fdb16f63b63e9dee53819044bffc857/google/cloud/speech/v1beta1/cloud_speech.proto#L249
[2]: https://github.com/googleapis/googleapis/blob/050400df0fdb16f63b63e9dee53819044bffc857/google/cloud/speech/v1beta1/cloud_speech.proto#L394

speech/unit_tests/test_operation.py
```diff
@@ -46,10 +46,12 @@ def test_from_api_repr(self):
         self.assertEqual('123456789', operation.name)
         self.assertTrue(operation.complete)
 
+        self.assertIsInstance(operation.results, list)
```
dhermes: Add a length assertion (the type assertion isn't very valuable?

PR #2640, opened 2016-10-28
Add _make_streaming_request, formerly _make_streaming_config.

daspecster:
Since this actually returned `StreamingRecognizeRequest`, I renamed it to `_make_streaming_request`.

### DIFF
--- a/speech/unit_tests/test__gax.py
+++ b/speech/unit_tests/test__gax.py
@@ -1,0 +1,80 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+
+class TestSpeechGAX(unittest.TestCase):
+    SAMPLE_RATE = 16000
+    HINTS = ['hi']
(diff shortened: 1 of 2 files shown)


speech/unit_tests/test__gax.py
```diff
@@ -0,0 +1,49 @@
...
+            SpeechContext)
+        from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
+            StreamingRecognizeRequest)
+
+        sample = Sample(source_uri=self.AUDIO_SOURCE_URI,
+                        encoding=speech.Encoding.FLAC,
+                        sample_rate=self.SAMPLE_RATE)
+        streaming_request = self._callFUT(sample, 'US-en', 2, True,
+                                          SpeechContext(), True, True)
+        self.assertIsInstance(streaming_request, StreamingRecognizeRequest)
```
dhermes: You don't want to make any assertions about the properties?

speech/unit_tests/test__gax.py
```diff
@@ -39,11 +38,38 @@ def test_ctor(self):
...
+        speech_context = SpeechContext(phrases=self.HINTS)
+        single_utterance = True
+        interim_results = False
+
+        streaming_request = self._callFUT(sample, language_code,
+                                          max_alternatives, profanity_filter,
+                                          speech_context, single_utterance,
+                                          interim_results)
         self.assertIsInstance(streaming_request, StreamingRecognizeRequest)
+        self.assertIsInstance(streaming_request.audio_content, bytes)
```
dhermes: It's overkill to check simple types like this, why not just check the value?

daspecster: So at this point, `streaming_request.audio_content` should be empty `b''`. The first `StreamingRecognizeRequest` **must** not have `audio_content` set and only `streaming_config` set.
The second `StreamingRecognizeRequest` can only have `audio_content` set and not `streaming_config.

I added a comment about this in the test for now.

daspecster:
Also just squashed.

PR #2641, opened 2016-10-28
Re-factoring Operation base class.

dhermes:
This is in preparation to support JSON/HTTP operations as well and also to ensure that **all** of the operation PB is parsed when polling.


core/google/cloud/operation.py
```diff
@@ -116,22 +163,46 @@ def complete(self):
         """
         return self._complete
 
+    def _get_operation_rpc(self):
+        """Checks the status of the current operation.
```
tseaver: "Polls" or "Retrieves"  would seem clearer than  "Checks".

PR #2644, opened 2016-10-29
Add _stream_requests() for managing speech streaming configuration

daspecster:
This method creates the configuration request before making content requests and yields them to the caller.

### DIFF
--- a/speech/unit_tests/test__gax.py
+++ b/speech/unit_tests/test__gax.py
@@ -15,10 +15,10 @@
 import unittest
 
 
-class TestSpeechGAX(unittest.TestCase):
+class TestSpeechGAXMakeRequests(unittest.TestCase):
     SAMPLE_RATE = 16000
     HINTS = ['hi']
-    AUDIO_CONTENT = '/9j/4QNURXhpZgAASUkq'
+    AUDIO_CONTENT = b'/9j/4QNURXhpZgAASUkq'
 
     def _callFUT(self, sample, language_code, max_alternatives,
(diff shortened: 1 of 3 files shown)


speech/unit_tests/test__gax.py
```diff
@@ -78,3 +78,63 @@ def test_ctor(self):
...
+        single_utterance = True
+        interim_results = False
+        streaming_requests = self._callFUT(sample, language_code,
+                                           max_alternatives, profanity_filter,
+                                           speech_context, single_utterance,
+                                           interim_results)
+        all_requests = []
+        for streaming_request in streaming_requests:
+            self.assertIsInstance(streaming_request, StreamingRecognizeRequest)
+            all_requests.append(streaming_request)
```
dhermes: No need for a change, just a note here. You can consume an iterator by "casting" to a list or other container type, e.g.:

```python
all_requests = list(streaming_requests)
```

speech/unit_tests/test__gax.py
```diff
@@ -78,3 +78,63 @@ def test_ctor(self):
...
+
+        self.assertEqual(len(all_requests), 2)
+
+        config_request = all_requests[0]
+        streaming_request = all_requests[1]
+        self.assertIsInstance(config_request, StreamingRecognizeRequest)
```
dhermes: You already checked this in the loop above.

daspecster:
Chat LGTM from @dhermes.

PR #2646, opened 2016-10-31
Allow Transcript to handle missing 'confidence' fields

daspecster:
When `max_alternitives` > 1, the `confidence` field is omitted entirely. So `Transcript` needs to set `confidence` to `None`.

```
{
 "results": [
  {
   "alternatives": [
    {
     "transcript": "hello thank you for using Google Cloud platform",
     "confidence": 0.95869493
    },
    {
     "transcript": "thank you for using Google Cloud platform"
    },
    {
     "transcript": "hello thank you for using Google Cloud platforms"
    }
   ]
  }
 ]
}
```

### DIFF
--- a/speech/google/cloud/speech/transcript.py
+++ b/speech/google/cloud/speech/transcript.py
@@ -38,7 +38,7 @@ class Transcript(object):
         :rtype: :class:`Transcript`
         :returns: Instance of ``Transcript``.
         """
-        return cls(transcript.get('transcript'), transcript.get('confidence'))
+        return cls(transcript['transcript'], transcript.get('confidence'))
 
     @classmethod
     def from_pb(cls, transcript):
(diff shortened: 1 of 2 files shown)


speech/google/cloud/speech/transcript.py
```diff
@@ -38,7 +38,7 @@ def from_api_repr(cls, transcript):
         :rtype: :class:`Transcript`
         :returns: Instance of ``Transcript``.
         """
-        return cls(transcript['transcript'], transcript['confidence'])
+        return cls(transcript.get('transcript'), transcript.get('confidence'))
```
tseaver: Can the `transcript` key actually be missing? If so, you need to add a test for that case.

daspecster: Actually, I don't think so. There would be no result at all. Maybe I should switch `transcript`  back to an index then?

daspecster: I put the indexing back since that doesn't really have anything to do with the issue.
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2646/commits/f7fb81895d6f433b6561f909fb7a695d78bd42cc

PR #2647, opened 2016-10-31
Expose connection property on _JSONSpeechAPI.

daspecster:
`Operation.poll` uses the `connection` to make the request.  This was missed in the tests since they set `connection` to `_Connection()`.


daspecster:
Thanks @tseaver! I have system tests that these are blocking! More to come!

PR #2648, opened 2016-10-31
Add speech async JSON/REST system tests.

daspecster:
This adds JSON/REST system tests for speech's `async_recognize()` method.

PR #2649, opened 2016-10-31
Add speech sync JSON/REST system tests.

daspecster:
This adds the speech `sync_recognize()` JSON/REST system test.

Based from https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2648

### DIFF
--- a/system_tests/speech.py
+++ b/system_tests/speech.py
@@ -63,6 +63,7 @@ class Config(object):
 def setUpModule():
     Config.CLIENT = speech.Client()
     Config.USE_GAX = Config.CLIENT._use_gax
+    Config.CLIENT = speech.Client
     # Now create a bucket for GCS stored content.
     storage_client = storage.Client()
     bucket_name = 'new' + unique_resource_id()
@@ -91,8 +92,8 @@ class TestSpeechClient(unittest.TestCase):
             value.delete()
 
     def _make_sync_request(self, content=None, source_uri=None,
-                           max_alternatives=None):
-        client = Config.CLIENT
+                           max_alternatives=None, use_gax=True):
+        client = Config.CLIENT(use_gax=use_gax)
         sample = client.sample(content=content,
                                source_uri=source_uri,
                                encoding=speech.Encoding.LINEAR16,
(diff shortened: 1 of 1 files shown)


system_tests/speech.py
```diff
@@ -37,7 +37,7 @@ class Config(object):
 
 
 def setUpModule():
-    Config.CLIENT = speech.Client()
+    Config.CLIENT = speech.Client
```
dhermes: Don't do this. You should add a boolean to `Config` that allows you to tell if using gRPC / GAX:

```python
Config.CLIENT = speech.Client()
Config.USE_GAX = Config.CLIENT._use_gax
```

then you can use `unittest.TestCase.skipTest` ([e.g.][1]) to avoid tests that can't run without GAX.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/319fd8f7e8a77a899747911bdd0b9dc9c357bb2f/system_tests/bigtable.py#L313-L318

system_tests/speech.py
```diff
@@ -66,8 +66,8 @@ def tearDown(self):
             value.delete()
 
     def _make_sync_request(self, content=None, source_uri=None,
-                           max_alternatives=None):
-        client = Config.CLIENT
+                           max_alternatives=None, use_gax=True):
+        client = Config.CLIENT(use_gax=use_gax)
```
dhermes: See my comment above.

system_tests/speech.py
```diff
@@ -125,6 +125,32 @@ def test_sync_recognize_gcs_file(self):
                                          max_alternatives=1)
         self._check_best_results(result)
 
+    def test_sync_recognize_local_file_rest(self):
+        with open(AUDIO_FILE, 'rb') as file_obj:
+            results = self._make_sync_request(content=file_obj.read(),
+                                              max_alternatives=2,
+                                              use_gax=False)
+            second_alternative = results[1]
```
dhermes: Definitely let the file context manager close before doing the entire test. Like in #2648, just read the file in the context manager and then call `_make_sync_request()` after the context manager.

system_tests/speech.py
```diff
@@ -125,6 +125,32 @@ def test_sync_recognize_gcs_file(self):
...
+            self.assertEqual(len(results), 2)
```
dhermes: This check should come **before** you access `results[1]`, otherwise you'd get an `IndexError` instead of a failed assertion.

daspecster:
I'm not sure how https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2649/commits/63cb2fa42e07746a45720324327c7a3edc352742 got in here? I rebased master in...

daspecster:
There we go!

daspecster:
Actually..this is pointless now since the switch to the env usage.

PR #2651, opened 2016-10-31
Passing Client through to low-level logging API objects.

dhermes:
This is in preparation for making `list_sinks()` and `list_metrics()` use `Iterator`.


daspecster:
LGTM

PR #2652, opened 2016-10-31
Updating list_sinks() to Iterator pattern.

dhermes:
~~**NOTE**: Has #2651 as diffbase. Unit test changes will be forthcoming.~~


dhermes:
@tseaver @daspecster PTAL this is ready to go

logging/google/cloud/logging/_helpers.py
```diff
@@ -24,7 +24,7 @@ def entry_from_resource(resource, client, loggers):
     """Detect correct entry type from resource and instantiate.
 
     :type resource: dict
-    :param resource: one entry resource from API response
+    :param resource: One entry resource from API response.
```
daspecster: 👍  😄

logging/unit_tests/test_connection.py
```diff
@@ -319,17 +322,33 @@ def test_list_sinks_no_paging(self):
...
         api = self._makeOne(client)
 
-        sinks, token = api.list_sinks(self.PROJECT)
+        iterator = api.list_sinks(self.PROJECT)
+        page = six.next(iterator.pages)
+        sinks = list(page)
+        token = iterator.next_page_token
 
-        self.assertEqual(sinks, RETURNED['sinks'])
+        # First check the token.
```
daspecster: I see you have these a few places. Is there a reason to note it more than just having the assert there?

daspecster:
This LGTM as well.

logging/unit_tests/test_connection.py
dhermes: The visual separation

logging/unit_tests/test__gax.py
```diff
@@ -612,56 +612,72 @@ def test_ctor(self):
...
+        api = self._makeOne(gax_api, client)
 
-        sinks, token = api.list_sinks(self.PROJECT)
+        iterator = api.list_sinks(self.PROJECT)
+        page = six.next(iterator.pages)
+        sinks = list(page)
+        token = iterator.next_page_token
 
-        self.assertEqual(sinks, SINKS)
+        # First check the token.
```
tseaver: Given that you just consumed the whole iterator above, asserting that the token is None seems an irrelevant implementation detail.

logging/unit_tests/test_client.py
```diff
@@ -375,25 +377,42 @@ def test_list_sinks_no_paging(self):
...
+            'sinks': SINKS,
+            'nextPageToken': TOKEN,
+        }
+        client.connection = _Connection(returned)
 
-        sinks, token = client.list_sinks()
+        iterator = client.list_sinks()
+        page = six.next(iterator.pages)
+        sinks = list(page)
+        token = iterator.next_page_token
```
tseaver: Should we have a test for `list_sinks` which ignores all paging (either passing the token or checking it)?

logging/unit_tests/test__gax.py
dhermes: Indeed. I have been trying to minimally change the old behavior without thinking about it.

logging/unit_tests/test_client.py
dhermes: Sure. Not sure what the value is that the two existing tests don't already provide?

tseaver: ISTM that they both exercise the "paged mode", which is likely to be somewhat uncommon compared to the "just iterate all the items" mode.

dhermes: I'm confident in the core tests for the iterator base to exercise those modes, I just tried to mimic the intent of the original test, which was inherently a single page.

PR #2653, opened 2016-10-31
Updating list_metrics() to Iterator pattern.

dhermes:
~~**NOTE**: Has #2651 as diffbase. Unit test changes will be forthcoming.~~

### DIFF
--- a/logging/unit_tests/test_client.py
+++ b/logging/unit_tests/test_client.py
@@ -485,33 +485,26 @@ class TestClient(unittest.TestCase):
         self.assertEqual(metric.project, self.PROJECT)
 
     def test_list_metrics_no_paging(self):
-        import six
         from google.cloud.logging.metric import Metric
 
-        PROJECT = 'PROJECT'
-        TOKEN = 'TOKEN'
-        METRICS = [{
+        metrics = [{
             'name': self.METRIC_NAME,
             'filter': self.FILTER,
             'description': self.DESCRIPTION,
         }]
-        client = self._makeOne(project=PROJECT, credentials=_Credentials(),
-                               use_gax=False)
+        client = self._makeOne(
+            project=self.PROJECT, credentials=_Credentials(),
+            use_gax=False)
         returned = {
-            'metrics': METRICS,
-            'nextPageToken': TOKEN,
+            'metrics': metrics,
         }
         client.connection = _Connection(returned)
 
         # Execute request.
         iterator = client.list_metrics()
-        page = six.next(iterator.pages)
-        metrics = list(page)
-        token = iterator.next_page_token
+        metrics = list(iterator)
 
-        # First check the token.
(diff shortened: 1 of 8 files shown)


dhermes:
@daspecster @tseaver PTAL

tseaver:
pyilnt failure:

logging/google/cloud/logging/_gax.py:500:1: E303 too many blank lines (3)

dhermes:
Ahhhh! Darn rebases. PTAL.

Was that your only issue?

tseaver:
I didn't really review it in detail since Travis was red.

dhermes:
OK well it's good to go now. LMK what you think

logging/unit_tests/test_client.py
```diff
@@ -519,22 +539,40 @@ def test_list_metrics_with_paging(self):
...
         }]
-        client = self._makeOne(project=PROJECT, credentials=_Credentials())
-        api = client._metrics_api = _DummyMetricsAPI()
-        api._list_metrics_response = METRICS, None
+        client = self._makeOne(project=PROJECT, credentials=_Credentials(),
+                               use_gax=False)
+        returned = {
+            'metrics': METRICS,
+        }
+        client.connection = _Connection(returned)
```
tseaver: *This* version tests the "list all items" mode in both test cases, but not the "page-at-a-time" mode.

logging/unit_tests/test_client.py
```diff
@@ -519,22 +539,40 @@ def test_list_metrics_with_paging(self):
...
 
         # Execute request.
-        metrics, token = client.list_metrics(PAGE_SIZE, TOKEN)
-        # Test values are correct.
+        iterator = client.list_metrics(PAGE_SIZE, TOKEN)
+        metrics = list(iterator)
+        token = iterator.next_page_token
+
+        # First check the token.
+        self.assertIsNone(token)
```
tseaver: Likewise, I don't think we care about the token in "list-all-items" mode

logging/unit_tests/test_client.py
dhermes: I don't have any preferences. What would you like it to do?

logging/unit_tests/test_client.py
dhermes: Happy to delete this line (and the one that accesses the token).

logging/unit_tests/test_client.py
tseaver: Perhaps we should adjust `test_list_metrics_with_paging` to read only a page, both passing one token and checking the returned token.  `test_list_metrics_wo_paging` can then ignore all features of paging (page size, tokens, and using the `iterator.pages` attribute).

dhermes: Done

dhermes:
@tseaver PTAL

PR #2655, opened 2016-10-31
Using Logging client's use_gax attr. in API properties.

dhermes:
These changes are also in #2652 (just so the tests would pass) but they are simple enough to stand on their own.

/cc @waprin


waprin:
lgtm

PR #2656, opened 2016-10-31
Add datastore.__init__ to docs.

dhermes:
This in turn required a fix in verify_included_modules to turn `__init__.py` files into the correct module / package name.

This is towards getting doctests running in `datastore`

### DIFF
--- a/datastore/google/cloud/datastore/__init__.py
+++ b/datastore/google/cloud/datastore/__init__.py
@@ -14,38 +14,40 @@
 
 """Shortcut methods for getting set up with Google Cloud Datastore.
 
-You'll typically use these to get started with the API::
+You'll typically use these to get started with the API:
 
-    >>> from google.cloud import datastore
-    >>>
-    >>> client = datastore.Client()
-    >>> key = client.key('EntityKind', 1234)
-    >>> entity = datastore.Entity(key)
-    >>> query = client.query(kind='EntityKind')
+.. code-block:: python
+
+  from google.cloud import datastore
+
+  client = datastore.Client()
+  key = client.key('EntityKind', 1234)
+  entity = datastore.Entity(key)
+  query = client.query(kind='EntityKind')
 
 The main concepts with this API are:
 
-- :class:`google.cloud.datastore.connection.Connection`
+- :class:`~google.cloud.datastore.connection.Connection`
   which represents a connection between your machine and the Cloud Datastore
   API.
 
-- :class:`google.cloud.datastore.client.Client`
+- :class:`~google.cloud.datastore.client.Client`
   which represents a project (string) and namespace (string) bundled with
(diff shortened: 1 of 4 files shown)


datastore/google/cloud/datastore/__init__.py
```diff
@@ -14,38 +14,40 @@
...
-    >>> entity = datastore.Entity(key)
-    >>> query = client.query(kind='EntityKind')
+.. code-block:: python
+
+  >>> from google.cloud import datastore
+  >>>
+  >>> client = datastore.Client()
+  >>> key = client.key('EntityKind', 1234)
+  >>> entity = datastore.Entity(key)
+  >>> query = client.query(kind='EntityKind')
```
tseaver: `.. code-block:: python` is not normally used with the REPL prompt(s) (that would be `.. doctest::`).

scripts/verify_included_modules.py
```diff
@@ -131,7 +130,12 @@ def get_public_modules(path, base_package=None):
                 if base_package is not None:
                     rel_path = os.path.join(base_package, rel_path)
                 # Turn into a Python module rather than a file path.
-                result.append(rel_path.replace(os.path.sep, '.'))
+                rel_path = rel_path.replace(os.path.sep, '.')
+                if mod_name == '__init__':
+                    # len('.__init__') == 9
+                    result.append(rel_path[:-9])
```
tseaver: Why not compute it?  E.g.:

```python
                    result.append(rel_path[:-len('.__init__')])
```

Or store it in a local variable, if microoptimization makes your boat float. :)

dhermes: Isn't hard-coding 9 the most optimal microoptimization? FWIW I realized this function is not so great when in there, but tried to change as little as possible.

datastore/google/cloud/datastore/__init__.py
dhermes: I'm happy to change them all back to the explicit block `::`. I see this as a temporary band-aid anyhow, since I will be introducing a check that all literal blocks are valid doctests (or bash or something else)

tseaver: Or you could leave in `.. code-block:: python` and delete the REPL prompts, which would be trivial given that no output is rendered.

dhermes: :+1: Will do

scripts/verify_included_modules.py
tseaver: *i* don't care about the immeasurably-small performance difference here. ;)  My point was that 9 is a majyk constant:  computing it either ahead or in place makes it clearer, and removes the need for the comment.

dhermes: OK updated.

PR #2657, opened 2016-11-01
Removing explicit (and implicit) doctest blocks from Sphinx docs.

dhermes:
This way we can gradually turn them on with **doctest** and make sure they work piece by piece.

Also converted some implicit code blocks (`::`) and some implicit doctest blocks (`:` followed by `>>>`) into explicit code blocks.

Related to #2656.

-----

Aside: Where did these changes come from? I started a branch to run datastore doctests side-by-side with system tests. Before over-engineering some handrolled solution I just wanted to get `sphinx-build -b doctest` working. So, each of these changes came from modules that resulted in failed doctests. Not all of the changes were actually interpreted as doctests (e.g. implicit code blocks via `::`) however, isolating them / designating them as `.. code-block:: python` gives a clearer target to come back at a later time and convert into a `doctest`.

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -447,29 +447,35 @@ class Client(_BaseClient, _ClientProjectMixin):
 
         Passes our ``project``.
 
-        Using query to search a datastore::
+        Using query to search a datastore:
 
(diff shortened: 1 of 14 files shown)


dhermes:
@tseaver @daspecster PTAL

PR #2660, opened 2016-11-01
Add a "How to start a new service" branch with sample commits and code

dhermes:
@jonparrott and @waprin have both suggested something like this and my discussion with @mckoss today pushed me over the "OK Jon and Bill" edge.

I am not quite done yet, but wanted to get it in the hands of those "in the know" now, so you can give ideas and feedback as it becomes a living thing.

- The commit content and commit messages are actually important, so please review them
- Should this go in it's own branch (referred to from CONTRIBUTING.rst) or should we check it in to `master` (checking in to master will help keep it from going stale, but may make the commits harder to find)
- Is there an actual fake service we can talk to? @bjwatson do you know

PS: H/T to http://unicorn.namegeneratorfun.com/

### DIFF
--- a/shiny/setup.py
+++ b/shiny/setup.py
@@ -1,0 +1,68 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from setuptools import find_packages
+from setuptools import setup
+
+
+PACKAGE_ROOT = os.path.abspath(os.path.dirname(__file__))
+
+with open(os.path.join(PACKAGE_ROOT, 'README.rst')) as file_obj:
+    README = file_obj.read()
+
+# NOTE: This is duplicated throughout and we should try to
+#       consolidate.
+SETUP_BASE = {
+    'author': 'Google Cloud Platform',
+    'author_email': 'jjg+google-cloud-python@google.com',
+    'scripts': [],
+    'url': 'https://github.com/GoogleCloudPlatform/google-cloud-python',
+    'license': 'Apache 2.0',
+    'platforms': 'Posix; MacOS X; Windows',
+    'include_package_data': True,
+    'zip_safe': False,
+    'classifiers': [
+        'Development Status :: 4 - Beta',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apache Software License',
+        'Operating System :: OS Independent',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.7',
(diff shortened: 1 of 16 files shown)


jonparrott:
> Is there an actual fake service we can talk to? 

You could always write an emulator for your fake service (in both HTTP/JSON and gRPC).

jonparrott:
> Should this go in it's own branch (referred to from CONTRIBUTING.rst) or should we check it in to master (checking in to master will help keep it from going stale, but may make the commits harder to find)

My initial thought is to check it into master. The commits themselves are less important than the content here, and I'm concerned not checking it in will make it drift out of date quickly.

waprin:
Looks really helpful!

Makes sense to go into master, makes less sense to me to be at a top level, rather than maybe in some subdirectory. Maybe demonstrate system tests?

daspecster:
Ditto what @waprin said. I don't think some other branch makes a lot of sense.
There should definitely be a reference to it in `CONTRIBUTING.rst` as well.

shiny/setup.py
```diff
@@ -0,0 +1,68 @@
...
+    'author': 'Google Cloud Platform',
+    'author_email': 'jjg+google-cloud-python@google.com',
+    'scripts': [],
+    'url': 'https://github.com/GoogleCloudPlatform/google-cloud-python',
+    'license': 'Apache 2.0',
+    'platforms': 'Posix; MacOS X; Windows',
+    'include_package_data': True,
+    'zip_safe': False,
+    'classifiers': [
+        'Development Status :: 4 - Beta',
```
bjwatson: Should probably be Alpha for a new API.

dhermes:
FYI: https://github.com/dhermes/google-shiny-service

bjwatson:
@dhermes Is "Shiny New Service" something that the user would deploy in their own GCP project?

bjwatson:
BTW, my wife and I enjoyed the unicorn name picker link. :)

mckoss:
Should there be Sphinx docs generation templates included?

dhermes:
@mckoss Yes and no. Yes, our sample package should show people how docs work. No, if I checked in `rst` files, this fake service would end up in the real docs.

But, I've been leaning this way for a while and recent discussions with @mckoss have solidified it: (I think) the current approach is unsustainable at scale and we should switch to using `sphinx-apidoc` as [`oauth2client`][2] and [`google-auth-library-python`][1] (due to @jonparrott) do.

[1]: https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/cadec899657213dfff65299567225df6e2ba239f/tox.ini#L62
[2]: https://github.com/google/oauth2client/blob/92e4ad332992b91d4674335f3d83a7181e38967d/scripts/build_docs.sh#L22

lukesneeringer:
@dhermes Is this still a thing / in progress? Or is it dead? If it is dead for all intents and purposes, we should close it.

dhermes:
@lukesneeringer I would like for it to be a thing, but we had a previous discussion where you indicated that it shouldn't be a thing?

lukesneeringer:
I might not be understanding the context of what this is, then? (I recall no such discussion.)

lukesneeringer:
This issue may no longer be relevant due to its age. Feel free to re-open.

dhermes:
@lukesneeringer You and I have had two or three conversations about the value (or lack of value) for such a "fake package" to exist.

I'm happy to "update" it if you think it's worthwhile, but will leave it alone otherwise.

We **previously** had 4 or 5 not so great experiences with other teams doing their own implementations and wanted a boilerplate we could point them at.

PR #2661, opened 2016-11-02
Fix pycodestyle failing when removing files from repo.

daspecster:
I noticed this after removing some unneeded pb files for speech.

Example:
```bash
Using files changed relative to origin/master:
------------------------------------------------------------
docs/index.rst
docs/speech-metadata.rst
docs/speech-operation.rst
docs/speech-usage.rst
speech/google/cloud/speech/_gax.py
speech/google/cloud/speech/client.py
speech/google/cloud/speech/metadata.py
speech/google/cloud/speech/operation.py
speech/setup.py
speech/unit_tests/test_client.py
speech/unit_tests/test_metadata.py
speech/unit_tests/test_operation.py
system_tests/speech.py
------------------------------------------------------------
checking speech/google/cloud/speech/_gax.py
checking speech/google/cloud/speech/client.py
checking speech/google/cloud/speech/metadata.py
speech/google/cloud/speech/metadata.py:1:1: E902 IOError: [Errno 2] No such file or directory: 'speech/google/cloud/speech/metadata.py'
checking speech/google/cloud/speech/operation.py
speech/google/cloud/speech/operation.py:1:1: E902 IOError: [Errno 2] No such file or directory: 'speech/google/cloud/speech/operation.py'
checking speech/setup.py
checking speech/unit_tests/test_client.py
checking speech/unit_tests/test_metadata.py
speech/unit_tests/test_metadata.py:1:1: E902 IOError: [Errno 2] No such file or directory: 'speech/unit_tests/test_metadata.py'
checking speech/unit_tests/test_operation.py
speech/unit_tests/test_operation.py:1:1: E902 IOError: [Errno 2] No such file or directory: 'speech/unit_tests/test_operation.py'
checking system_tests/speech.py
ERROR: InvocationError: '/Users/daspecster/Documents/google/google-cloud-py-2/.tox/lint/bin/python /Users/daspecster/Documents/google/google-cloud-py-2/scripts/pycodestyle_on_repo.py'
__________________________________________________________________________________ summary __________________________________________________________________________________
ERROR:   lint: commands failed
```

PR #2663, opened 2016-11-02
Add speech async gapic

daspecster:
~~This will need #2661 in order for travis-ci to pass since I don't need `metadata.py` and `operation.py` after @dhermes refactored the core `operation.py`@ Thanks!~~

@dhermes, I think we talked about this back and forth a little bit, but as you can see from the system tests, the surface is much different between `sync_recognize()`'s results and `async_recognize()`'s results.

There will be some updates for the gRPC version of `sync_recognize()` that will follow this as well.

### DIFF
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -14,23 +14,36 @@
 
 """GAX/GAPIC module for managing Speech API requests."""
 
+
 from google.cloud.gapic.speech.v1beta1.speech_api import SpeechApi
-from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import SpeechContext
-from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import RecognitionConfig
 from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import RecognitionAudio
+from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import RecognitionConfig
+from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import SpeechContext
 from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
     StreamingRecognitionConfig)
 from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
     StreamingRecognizeRequest)
+from google.longrunning import operations_grpc
 
+from google.cloud._helpers import make_secure_stub
+from google.cloud.connection import DEFAULT_USER_AGENT
 
 from google.cloud.speech.alternative import Alternative
+from google.cloud.speech.operation import Operation
+
+OPERATIONS_API_HOST = 'speech.googleapis.com'
 
 
 class GAPICSpeechAPI(object):
     """Manage calls through GAPIC wrappers to the Speech API."""
-    def __init__(self):
+    def __init__(self, client=None):
+        self._client = client
         self._gapic_api = SpeechApi()
+        self._operations_stub = make_secure_stub(
+            self._client.connection.credentials,
+            DEFAULT_USER_AGENT,
+            operations_grpc.OperationsStub,
+            OPERATIONS_API_HOST)
 
     def async_recognize(self, sample, language_code=None,
                         max_alternatives=None, profanity_filter=None,
@@ -72,9 +85,21 @@ class GAPICSpeechAPI(object):
                                and phrases. This can also be used to add new
                                words to the vocabulary of the recognizer.
 
-        :raises NotImplementedError: Always.
+        :rtype: :class:`~google.cloud.speech.operation.Operation`
+        :returns: Instance of ``Operation`` to poll for results.
         """
-        raise NotImplementedError
+        config = RecognitionConfig(
+            encoding=sample.encoding, sample_rate=sample.sample_rate,
+            language_code=language_code, max_alternatives=max_alternatives,
+            profanity_filter=profanity_filter,
+            speech_context=SpeechContext(phrases=speech_context))
+
+        audio = RecognitionAudio(content=sample.content,
+                                 uri=sample.source_uri)
+        api = self._gapic_api
+        response = api.async_recognize(config=config, audio=audio)
+
+        return Operation.from_pb(response, self)
 
     def sync_recognize(self, sample, language_code=None, max_alternatives=None,
                        profanity_filter=None, speech_context=None):
--- a/speech/setup.py
+++ b/speech/setup.py
@@ -52,7 +52,6 @@ SETUP_BASE = {
 REQUIREMENTS = [
     'google-cloud-core >= 0.20.0',
     'gapic-google-cloud-speech-v1beta1 >= 0.11.1, < 0.12.0',
-    'grpc-google-cloud-speech-v1beta1 >= 0.11.1, < 0.12.0',
 ]
 
 setup(
--- a/speech/unit_tests/test_client.py
+++ b/speech/unit_tests/test_client.py
@@ -21,6 +21,28 @@ class TestClient(unittest.TestCase):
     AUDIO_SOURCE_URI = 'gs://sample-bucket/sample-recording.flac'
     AUDIO_CONTENT = '/9j/4QNURXhpZgAASUkq'
 
+    @staticmethod
+    def _make_result(alternatives):
+        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+
+        return cloud_speech_pb2.SpeechRecognitionResult(
+            alternatives=[
+                cloud_speech_pb2.SpeechRecognitionAlternative(
+                    transcript=alternative['transcript'],
+                    confidence=alternative['confidence'],
+                ) for alternative in alternatives
+            ],
+        )
+
+    def _make_sync_response(self, *results):
+        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+
+        response = cloud_speech_pb2.SyncRecognizeResponse(
+            results=results,
+        )
+
+        return response
+
     def _getTargetClass(self):
         from google.cloud.speech.client import Client
 
@@ -69,15 +91,15 @@ class TestClient(unittest.TestCase):
 
     def test_sync_recognize_content_with_optional_params_no_gax(self):
         from base64 import b64encode
-        from google.cloud._helpers import _to_bytes
-        from google.cloud._helpers import _bytes_to_unicode
 
+        from google.cloud._helpers import _bytes_to_unicode
+        from google.cloud._helpers import _to_bytes
         from google.cloud._testing import _Monkey
-        from google.cloud.speech import client as MUT
+
         from google.cloud import speech
+        from google.cloud.speech import client as MUT
         from google.cloud.speech.alternative import Alternative
         from google.cloud.speech.sample import Sample
-
         from unit_tests._fixtures import SYNC_RECOGNIZE_RESPONSE
 
         _AUDIO_CONTENT = _to_bytes(self.AUDIO_CONTENT)
@@ -131,8 +153,9 @@ class TestClient(unittest.TestCase):
 
     def test_sync_recognize_source_uri_without_optional_params_no_gax(self):
         from google.cloud._testing import _Monkey
-        from google.cloud.speech import client as MUT
+
         from google.cloud import speech
+        from google.cloud.speech import client as MUT
         from google.cloud.speech.alternative import Alternative
         from google.cloud.speech.sample import Sample
         from unit_tests._fixtures import SYNC_RECOGNIZE_RESPONSE
@@ -174,8 +197,9 @@ class TestClient(unittest.TestCase):
 
     def test_sync_recognize_with_empty_results_no_gax(self):
         from google.cloud._testing import _Monkey
-        from google.cloud.speech import client as MUT
+
         from google.cloud import speech
+        from google.cloud.speech import client as MUT
         from google.cloud.speech.sample import Sample
         from unit_tests._fixtures import SYNC_RECOGNIZE_EMPTY_RESPONSE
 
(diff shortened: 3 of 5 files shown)


speech/setup.py
```diff
@@ -51,6 +51,8 @@
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.20.0',
+    'grpcio >= 1.0.0, < 2.0dev',
+    'google-gax >= 0.14.1, < 0.15dev',
```
daspecster: I don't think these are needed actually.

speech/google/cloud/speech/_gax.py
```diff
@@ -23,13 +29,23 @@
 from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
     StreamingRecognizeRequest)
 
-
 from google.cloud.speech.alternative import Alternative
+from google.cloud._helpers import make_secure_stub
+from google.cloud.connection import DEFAULT_USER_AGENT
+from google.cloud.speech.operation import Operation
+from google.cloud.operation import register_type
```
dhermes: Put the non-`google.cloud.speech` imports above (and put them in alphabetical order)

speech/google/cloud/speech/_gax.py
```diff
@@ -23,13 +29,23 @@
...
+
+
+OPERATIONS_API_HOST = 'speech.googleapis.com'
+
+register_type(AsyncRecognizeMetadata)
+register_type(AsyncRecognizeResponse)
```
dhermes: This isn't needed anymore, they have been registered in `speech.operation`

speech/google/cloud/speech/_gax.py
```diff
@@ -72,9 +88,26 @@ def async_recognize(self, sample, language_code=None,
                                and phrases. This can also be used to add new
                                words to the vocabulary of the recognizer.
 
-        :raises NotImplementedError: Always.
+        :rtype: :class:`~google.cloud.operation.Opeartion`
```
dhermes: 1. typo
1. return a `google.cloud.speech.operation.Operation`

speech/google/cloud/speech/_gax.py
```diff
@@ -72,9 +88,26 @@ def async_recognize(self, sample, language_code=None,
...
+            language_code=language_code, max_alternatives=max_alternatives,
+            profanity_filter=profanity_filter,
+            speech_context=SpeechContext(phrases=speech_context))
+
+        audio = RecognitionAudio(content=sample.content,
+                                 uri=sample.source_uri)
+        api = self._gapic_api
+        response = api.async_recognize(config=config, audio=audio)
+
+        self._client._operations_stub = make_secure_stub(
```
dhermes: This shouldn't happen here. Instead do it in the constructor for `GAPICSpeechAPI`. Technically you don't need to pass a client to `Operation.from_pb()`.

If `self._operations_stub` was defined (in the constructor) for this `GAPICSpeechAPI` object then you could just use

```python
Operation.from_pb(response, self)
```

speech/setup.py
```diff
@@ -52,7 +52,6 @@
 REQUIREMENTS = [
     'google-cloud-core >= 0.20.0',
     'gapic-google-cloud-speech-v1beta1 >= 0.11.1, < 0.12.0',
-    'grpc-google-cloud-speech-v1beta1 >= 0.11.1, < 0.12.0',
```
dhermes: Why is this removed?

speech/unit_tests/test_client.py
```diff
@@ -199,11 +199,10 @@ def test_sync_recognize_with_empty_results_gax(self):
         credentials = _Credentials()
         client = self._makeOne(credentials=credentials, use_gax=True)
         client.connection = _Connection()
+        _MockGAPICSpeechAPI._results = []
```
dhermes: Why is this necessary? Can't you just create an instance of `_MockGAPICSpeechAPI`?

speech/unit_tests/test_client.py
```diff
@@ -277,13 +274,16 @@ def test_async_recognize_with_gax(self):
         credentials = _Credentials()
         client = self._makeOne(credentials=credentials)
         client.connection = _Connection()
+        client.connection.credentials = credentials
```
dhermes: Why do the credentials matter?

speech/unit_tests/test_client.py
```diff
@@ -277,13 +274,16 @@ def test_async_recognize_with_gax(self):
...
         sample = client.sample(source_uri=self.AUDIO_SOURCE_URI,
                                encoding=speech.Encoding.LINEAR16,
                                sample_rate=self.SAMPLE_RATE)
         with _Monkey(MUT, SpeechApi=_MockGAPICSpeechAPI):
-            with self.assertRaises(NotImplementedError):
-                client.async_recognize(sample)
+            operation = client.async_recognize(sample)
+
+        self.assertFalse(operation.complete)
+        self.assertIsNone(operation.response)
```
dhermes: Check the type as well?

speech/unit_tests/test_client.py
```diff
@@ -321,28 +321,40 @@ class _MockGAPICAlternative(object):
     confidence = 0.95234356
 
 
+class _MockGAPICMetadata(object):
+    type_url = None
```
dhermes: Don't mock this, just use the PB types. The excessive mocking can be problematic.

[Example][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6c5975502fcfab7cde6d9d34c433c26761505a72/speech/unit_tests/test_operation.py#L43-L73

speech/unit_tests/test_client.py
daspecster: I can override `GAPICSpeechAPI._gapic_api` instead I think.
But it is this way now because `SpeechApi` gets called in the `GAPICSpeechAPI` constructor. `self._gapic_api = SpeechApi()`.

speech/setup.py
daspecster: It's a dependency of `gapic-google-cloud-speech-v1beta1` so it gets installed automatically.

speech/unit_tests/test_client.py
daspecster: It's needed for `make_secure_stub`.
```bash
    def __init__(self, client=None):
        self._client = client
        self._gapic_api = SpeechApi()
        self._operations_stub = make_secure_stub(
>           self._client.connection.credentials,
            DEFAULT_USER_AGENT,
            operations_grpc.OperationsStub,
            OPERATIONS_API_HOST)
E       AttributeError: '_Connection' object has no attribute 'credentials'
```

daspecster:
@dhermes I think there are just thow two open questions. Did you have anything else?

speech/unit_tests/test_client.py
```diff
@@ -21,6 +21,26 @@ class TestClient(unittest.TestCase):
     AUDIO_SOURCE_URI = 'gs://sample-bucket/sample-recording.flac'
     AUDIO_CONTENT = '/9j/4QNURXhpZgAASUkq'
 
+    @staticmethod
+    def _make_result(transcript, confidence):
+        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+        return cloud_speech_pb2.SpeechRecognitionResult(
```
dhermes: Nit: empty line between the import and the code

speech/unit_tests/test_client.py
```diff
@@ -21,6 +21,26 @@ class TestClient(unittest.TestCase):
...
+            alternatives=[
+                cloud_speech_pb2.SpeechRecognitionAlternative(
+                    transcript=transcript,
+                    confidence=confidence,
```
dhermes: ISTM you could benefit from a unit test that has a result with multiple `alternatives`, no?

speech/unit_tests/test_client.py
```diff
@@ -21,6 +21,26 @@ class TestClient(unittest.TestCase):
...
+                ),
+            ],
+        )
+
+    def _make_sync_response(self, *results):
+        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+        response = cloud_speech_pb2.SyncRecognizeResponse(
```
dhermes: Nit: empty line between the import and the code

speech/unit_tests/test_client.py
```diff
@@ -192,45 +214,48 @@ def test_sync_recognize_with_empty_results_no_gax(self):
 
     def test_sync_recognize_with_empty_results_gax(self):
         from google.cloud._testing import _Monkey
-        from google.cloud.speech import _gax as MUT
+
         from google.cloud import speech
+        from google.cloud.speech import _gax as MUT
```
dhermes: Nit: `MUT` stands for module-under-test. So in `test_client` the module-under-test is `client`, not `_gax`.

speech/unit_tests/test_client.py
```diff
@@ -192,45 +214,48 @@ def test_sync_recognize_with_empty_results_no_gax(self):
...
         from google.cloud.speech.sample import Sample
 
         credentials = _Credentials()
         client = self._makeOne(credentials=credentials, use_gax=True)
         client.connection = _Connection()
+        client.connection.credentials = credentials
+
+        _MockGAPICSpeechAPI._response = self._make_sync_response()
```
dhermes: Don't use a mutable global class here. You can monkey-patch `SpeechApi` with a function instead of a class, and that function could spit out an instance of your `_MockGAPICSpeechAPI` class that has `_response` equal to  `self._make_sync_response()`.

speech/unit_tests/test_client.py
dhermes: 👍

speech/unit_tests/test_client.py
```diff
@@ -316,33 +348,23 @@ def test_speech_api_preset(self):
...
+        operation = Operation()
+        return operation
 
     def sync_recognize(self, config, audio):
         self.config = config
         self.audio = audio
-        mock_response = self._response
-        mock_response.results = self._results
-        return mock_response
+        # self._response.results = self._results
```
dhermes: No commented out code please

speech/unit_tests/test_client.py
```diff
@@ -316,33 +348,23 @@ def test_speech_api_preset(self):
...
-
 class _MockGAPICSpeechAPI(object):
     _requests = None
-    _response = _MockGAPICSpeechResponse()
-    _results = [_MockGAPICSyncResult()]
+    _response = None
+    _results = None
+
+    def async_recognize(self, config, audio):
+        from google.longrunning.operations_pb2 import Operation
```
dhermes: Nit: empty line after import

speech/unit_tests/test_client.py
daspecster: How about....
```diff
- MUT  # Module under test
+ MUM  # Module under mock
```

I guess the real point here is that I need to mock out `SpeechApi`, I guess I could try and override it through `client`.

speech/unit_tests/test_client.py
```diff
@@ -237,25 +240,39 @@ def test_sync_recognize_with_gax(self):
...
-        result = self._make_result(transcript, confidence)
-        _MockGAPICSpeechAPI._response = self._make_sync_response(result)
-
-        with _Monkey(MUT, SpeechApi=_MockGAPICSpeechAPI):
+        alternatives = [{
+            "transcript": 'testing 1 2 3',
+            "confidence": 0.9224355
+        }, {
+            "transcript": 'testing 4 5 6',
+            "confidence": 0.0123456
```
dhermes: 1. single quotes
1. trailing comma `0.0123456,`

speech/unit_tests/test_client.py
```diff
@@ -216,18 +218,19 @@ def test_sync_recognize_with_empty_results_gax(self):
...
         client.connection = _Connection()
         client.connection.credentials = credentials
 
-        _MockGAPICSpeechAPI._response = self._make_sync_response()
+        def speech_api():
+            return _MockGAPICSpeechAPI(response=self._make_sync_response())
 
         with self.assertRaises(ValueError):
-            with _Monkey(MUT, SpeechApi=_MockGAPICSpeechAPI):
+            with _Monkey(_gax, SpeechApi=speech_api):
```
dhermes: Re-factor this as follow:

```
# Monkey-patch while manually setting the speech API on client
with _Monkey(_gax, SpeechApi=speech_api):
    client._speech_api = _gax.GAPICSpeechAPI()

sample = Sample(source_uri=self.AUDIO_SOURCE_URI,
                encoding=speech.Encoding.FLAC,
                sample_rate=self.SAMPLE_RATE)

    client.sync_recognize(sample)
```

speech/unit_tests/test_client.py
```diff
@@ -237,25 +240,39 @@ def test_sync_recognize_with_gax(self):
...
-            self.assertEqual(results[0].transcript, transcript)
-            self.assertEqual(results[0].confidence, confidence)
+            self.assertEqual(results[0].transcript,
+                             alternatives[0]['transcript'])
+            self.assertEqual(results[0].confidence,
+                             alternatives[0]['confidence'])
+            self.assertEqual(results[1].transcript,
+                             alternatives[1]['transcript'])
+            self.assertEqual(results[1].confidence,
+                             alternatives[1]['confidence'])
```
dhermes: 1. Don't create the `sample` in the context manager or do the asserts in the context manager.
1. Assert `len(results) == 2`
1. Do the same monkey-patch on the `client` object suggested above.

speech/setup.py
jonparrott: @dhermes the gapic one brings in the grpc one, right? In that case, I'd avoid specifying the transitive dependency.

daspecster:
Travis has gone on strike again...

daspecster:
Finally travis ran this job. @dhermes, I'm merging(in 15minutes) since I got your suggestions in.

PR #2664, opened 2016-11-02
Re-factoring Speech operation to use the base class.

dhermes:
Also:

- Fixing `script_utils.get_affected_files` to only return non-deleted files


daspecster:
Gah, jinx.

speech/google/cloud/speech/operation.py
```diff
@@ -14,120 +14,55 @@
 
 """Long running operation representation for Google Speech API"""
 
-from google.cloud.speech.metadata import Metadata
-from google.cloud.speech.transcript import Transcript
+from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
```
dhermes: @tseaver @daspecster This effectively makes the `_HAVE_GAX` / `ImportError` check in `client` pointless. And it is anyway since we have the deps in `setup.py`. I am :+1: on removing the check and only using the env. var. for `_USE_GAX`.

dhermes:
@daspecster Sorry, I guess I was unclear after our discussion yesterday.

speech/google/cloud/speech/operation.py
```diff
@@ -14,120 +14,55 @@
...
-            for result in raw_results:
-                for alternative in result['alternatives']:
-                    results.append(Transcript.from_api_repr(alternative))
-        if metadata:
-            self._metadata = Metadata.from_api_repr(metadata)
-
-        self._results = results
-        self._complete = response.get('done', False)
+        result = pb_results[0]
+        self.results = [Transcript.from_pb(alternative)
```
daspecster: Do we really need to use `Transcript` for this now? Since it's going to get renamed to `Alternative` which would pretty much be the same as [google.cloud.speech.v1beta1.SpeechRecognitionAlternative(https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.SpeechRecognitionAlternative)?

PR #2665, opened 2016-11-02
Removing ImportError check in Speech client.

dhermes:
I mentioned this in #2664 and this delivers on it.

We should also get rid of all the instances of this in other packages. LMK what you both think (and @jonparrott I'd be interested in what you think, i.e. importing `grpc` on GAE)

### DIFF
--- a/speech/google/cloud/speech/client.py
+++ b/speech/google/cloud/speech/client.py
@@ -22,23 +22,15 @@ from google.cloud._helpers import _bytes_to_unicode
 from google.cloud.client import Client as BaseClient
 from google.cloud.environment_vars import DISABLE_GRPC
 
+from google.cloud.speech._gax import GAPICSpeechAPI
 from google.cloud.speech.connection import Connection
 from google.cloud.speech.encoding import Encoding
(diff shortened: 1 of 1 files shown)


jonparrott:
It's not ideal for GAE where the gRPC won't be available for some time, but we've already crossed that bridge it seems.

dhermes:
I'm going to leave this open before merging to let @tseaver weigh in.

----

> we've already crossed that bridge it seems.

@jonparrott How do you mean?

jonparrott:
> we've already crossed that bridge it seems.

There are other locations where grpc is not import checked, right?

dhermes:
> There are other locations where grpc is not import checked, right?

Not that I know of, except a change I made earlier today in `speech`. Unfortunately the PB definitions are bundled with `grpc`, so it's tough to parse the `Any` fields in the operations returned from the API (gRPC or JSON-over-HTTP) without importing the PB types.

jonparrott:
@dhermes Gotcha. In that case I'm fine with this outside of datastore, logging, pubsub, storage, and logging.

PR #2671, opened 2016-11-03
Adding Python 3.5 on Travis.

dhermes:
~~Also removing unit tests on Travis for Python 3.4 (to keep the builds fast, if possible).~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,20 @@
 language: python
 sudo: false
 
+addons:
+  apt:
+    sources:
+    - deadsnakes
+    packages:
+    - python3.5
+
 install:
   - pip install --upgrade pip tox
 
 script:
   - python2.7 scripts/run_unit_tests.py
   - python3.4 scripts/run_unit_tests.py
+  - python3.5 scripts/run_unit_tests.py
   - python scripts/run_unit_tests.py --tox-env cover
   - tox -e lint
   - tox -e system-tests


dhermes:
@tseaver Any reservations about this change?

.travis.yml
```diff
@@ -1,12 +1,19 @@
 language: python
 sudo: false
 
+addons:
+  apt:
+    sources:
+    - deadsnakes
+    packages:
+    - python3.5
+
```
dhermes: Empirically that isn't the case.

dhermes: See: https://github.com/travis-ci/travis-ci/issues/4794

dhermes:
@tseaver So we good to merge? No problem with **only** running 3.5 (rather than 3.4 and 3.5)?

tseaver:
@dhermes Are we picking up 3.4 under coverage?  Or just deciding not to support it?

dhermes:
@tseaver Neither. It's supported, just not tested on Travis. I can add it back if you think 3.5 and 3.4 will be sufficiently different. We also run 3.4 on AppVeyor (on merges to master).

tseaver:
If we don't test it, it isn't supported, because no one will hear the screams.

dhermes:
They will hear the screams on AppVeyor?

dhermes:
@tseaver Bump

daspecster:
What about running both 3.4 and 3.5 only on master builds?

dhermes:
Just left 3.4 in there to avoid the conflicting opinions holding this up. If we find Travis getting slow we can revisit.

PR #2676, opened 2016-11-03
Speech unit tests cleanup.

daspecster:
Some test cleanup that was missed or not relevant on the #2663 PR.

Issue #2522, opened 2016-10-10
Tracking Speech API development.


daspecster:
These are the items required to reach feature completion.
- [x] Use Transcript object for sync and async requests. (#2613)
- [x] GAPIC Sync Recognize (#2615)
- [x] GAPIC Async Recognize(#2638, #2663)
- [x] Streaming recognize (#2523, #2640,#2644, #2680)
- [x] Add stability information to streaming interim_results (https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2702)
- [x] Add `sync_recognize` system tests (#2547)
- [x] Add `async_recognize` system tests (#2648)

## Cleanup
- [x] Update `SpeechContext` in tests. See: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2700. (#2717)
- [x] Add PyPI [badges](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2685) (#2716)
- [x] Rename `Transcript` to `Alternative`. (#2666)
- [x] Test cleanup (#2676)
- [x] Run skipped tests(https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2648#discussion_r85783495)
- [x] Move sample to Client property. See: #2516
- [x] Remove warning about `google-cloud-speech` not yet released. See: #2620 (#2732)

### Future Updates:
- Clearer way to send stream to `Sample.content` (https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2708)
-  Move result access to Operation handler https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2638#discussion_r85596232
- Add snippet tests (optional for release)

daspecster:
@dhermes I commented on your gist from today but I'll copy it here as well...
Ref: https://gist.github.com/dhermes/09c964d6d27003ae817b650424fda7c3

--------

Thank you for this!

I have another question.
As you saw in your output here, the final result has a `response` attribute.

In the [sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/d7dc5780cd975ed5927e68bd5dd5a4a69ce5efb5/speech/grpc/transcribe_async.py#L94-L95) the way that you unpack the response is via
```python
response = cloud_speech_pb2.AsyncRecognizeResponse()
operation.response.Unpack(response)
```

But that has to be done after the Operation is completed.
How would we handle that in `google-cloud-python`?

[...]

PR #2680, opened 2016-11-04
Add speech streaming recognition.

daspecster:
Towards #2522. Add `streaming_recognition()` to speech API.

~~I will rebase #2676 before merging.~~

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -151,5 +151,85 @@ words to the vocabulary of the recognizer.
     transcript: Hello, this is a test
     confidence: 0.81
 
+
+Streaming Recognition
+---------------------
+
+The :meth:`~google.cloud.speech.Client.streaming_recognize` method converts
+speech data to possible text alternatives on the fly.
+
+.. note::
+    Streaming recognition requests are limited to 1 minute of audio.
+
+    See: https://cloud.google.com/speech/limits#content
+
+.. code-block:: python
+
+    >>> from google.cloud import speech
+    >>> client = speech.Client()
+    >>> with open('./hello.wav', 'rb') as stream:
+    ...     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    ...     alternatives = list(client.streaming_recognize(sample))
+    >>> print(alternatives[0].transcript)
+    'hello'
+    >>> print(alternatives[0].confidence)
+    0.973458576
+
+
+By default the API will perform continuous recognition
+(continuing to process audio even if the speaker in the audio pauses speaking)
+until the client closes the output stream or until the maximum time limit has
+been reached.
+
+If you only want to recognize a single utterance you can set
+ ``single_utterance`` to :data:`True` and only one result will be returned.
+
+See: `Single Utterance`_
+
+.. code-block:: python
+
+    >>> with open('./hello_pause_goodbye.wav', 'rb') as stream:
+    ...     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    ...     responses = client.streaming_recognize(sample,
+    ...                                            single_utterance=True)
+    ...     alternatives = list(responses)
+    >>> print(alternatives[0].transcript)
+    hello
+    >>> print(alternatives[0].confidence)
+    0.96523453546
+
+
+If ``interim_results`` is set to :data:`True`, interim results
+(tentative hypotheses) may be returned as they become available.
+
+.. code-block:: python
+
+    >>> from google.cloud import speech
+    >>> client = speech.Client()
+    >>> with open('./hello.wav', 'rb') as stream:
+    ...     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    ...     for alternatives in client.streaming_recognize(sample,
+    ...                                                    interim_results=True):
+    ...         print('=' * 20)
+    ...         print(alternatives[0].transcript)
+    ...         print(alternatives[0].confidence)
+    ====================
+    'he'
+    None
+    ====================
+    'hell'
+    None
+    ====================
+    'hello'
+    0.973458576
+
+
+.. _Single Utterance: https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#streamingrecognitionconfig
 .. _sync_recognize: https://cloud.google.com/speech/reference/rest/v1beta1/speech/syncrecognize
 .. _Speech Asynchronous Recognize: https://cloud.google.com/speech/reference/rest/v1beta1/speech/asyncrecognize
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -106,6 +106,97 @@ class GAPICSpeechAPI(object):
 
         return Operation.from_pb(response, self)
 
+    def streaming_recognize(self, sample, language_code=None,
+                            max_alternatives=None, profanity_filter=None,
+                            speech_context=None, single_utterance=False,
+                            interim_results=False):
+        """Streaming speech recognition.
+
+        .. note::
+
+            Streaming recognition requests are limited to 1 minute of audio.
+            See: https://cloud.google.com/speech/limits#content
+
+        Yields :class:`~streaming_response.StreamingSpeechResponse` containing
+        results and metadata from the streaming request.
+
+        :type sample: :class:`~google.cloud.speech.sample.Sample`
+        :param sample: Instance of ``Sample`` containing audio information.
+
+        :type language_code: str
+        :param language_code: (Optional) The language of the supplied audio as
+                              BCP-47 language tag. Example: ``'en-GB'``.
+                              If omitted, defaults to ``'en-US'``.
+
+        :type max_alternatives: int
+        :param max_alternatives: (Optional) Maximum number of recognition
+                                 hypotheses to be returned. The server may
+                                 return fewer than maxAlternatives.
+                                 Valid values are 0-30. A value of 0 or 1
+                                 will return a maximum of 1. Defaults to 1
+
+        :type profanity_filter: bool
+        :param profanity_filter: If True, the server will attempt to filter
+                                 out profanities, replacing all but the
+                                 initial character in each filtered word with
+                                 asterisks, e.g. ``'f***'``. If False or
+                                 omitted, profanities won't be filtered out.
+
+        :type speech_context: list
+        :param speech_context: A list of strings (max 50) containing words and
+                               phrases "hints" so that the speech recognition
+                               is more likely to recognize them. This can be
+                               used to improve the accuracy for specific words
+                               and phrases. This can also be used to add new
+                               words to the vocabulary of the recognizer.
+
+        :type single_utterance: bool
+        :param single_utterance: (Optional) If false or omitted, the recognizer
+                                 will perform continuous recognition
+                                 (continuing to process audio even if the user
+                                 pauses speaking) until the client closes the
+                                 output stream (gRPC API) or when the maximum
+                                 time limit has been reached. Multiple
+                                 SpeechRecognitionResults with the is_final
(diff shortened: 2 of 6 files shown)


daspecster:
@dhermes, I assume that this travis build won't happen until tonight. If you think it's worthwhile can you review now? Then after I push your suggestions I'll cancel this build and start a new one.

system_tests/speech.py
```diff
@@ -175,3 +185,43 @@ def test_async_recognize_gcs_file(self):
...
+            for results in self._make_streaming_request(file_obj):
+                self._check_results(results)
+
+    def test_stream_recognize_interim_results(self):
+        if not Config.USE_GAX:
+            self.skipTest('gRPC is required for Speech Streaming Recognize.')
+
+        # These extra words are interim_results that the API returns as it's
+        # deciphering the speech audio. This has a high probability of becoming
+        # out of date and causing the test to fail.
```
daspecster: Nope, it's been highly variable. I've gotten 5 responses and the last time I ran it I got 27...

speech/google/cloud/speech/sample.py
```diff
@@ -47,8 +47,8 @@ class Sample(object):
     default_encoding = Encoding.FLAC
     default_sample_rate = 16000
 
-    def __init__(self, content=None, source_uri=None,
-                 encoding=None, sample_rate=None):
+    def __init__(self, content=None, source_uri=None, encoding=None,
+                 sample_rate=None):
```
daspecster: Yeah I had another input type `stream` but I took it out and reformatted. Apparently this was wrongly formatted in `master`.

speech/google/cloud/speech/_gax.py
```diff
@@ -101,6 +101,95 @@ def async_recognize(self, sample, language_code=None,
...
+                                returned.
+
+        :raises: :class:`EnvironmentError` if gRPC is not enabled and
+            :class:`ValueError` if stream has closed.
+
+        :rtype: :class:`~google.cloud.grpc.speech.v1beta1\
+                       .cloud_speech_pb2.StreamingRecognizeResponse`
+        :returns: ``StreamingRecognizeResponse`` instances.
+        """
+        if sample.content.closed:
```
daspecster: Should I add a 3rd input type? `source_uri`, `content` and `stream`?
I had that but I took it out because I thought it wasn't needed.

speech/google/cloud/speech/client.py
```diff
@@ -159,6 +159,91 @@ def speech_api(self):
...
+                                                        profanity_filter,
+                                                        speech_context,
+                                                        single_utterance,
+                                                        interim_results)
+        for response in responses:
+            results = getattr(response, 'results', [])
+            if results or interim_results:
+                for result in results:
+                    yield [Alternative.from_pb(alternative)
+                           for alternative in result.alternatives]
```
daspecster: Yeah, I've been trying to figure out a clearer way to do this.

So if `interim_results` is `True` then more than one response will have results defined.
If `interim_results` is `False`(The default) then only the last response will have results defined, but there will be other responses defining the `endpointer_type`.

Does that clear it up or was there a different part that was confusing?

speech/google/cloud/speech/client.py
```diff
@@ -159,6 +159,91 @@ def speech_api(self):
...
+            raise EnvironmentError('gRPC is required to use this API.')
+
+        responses = self.speech_api.streaming_recognize(sample, language_code,
+                                                        max_alternatives,
+                                                        profanity_filter,
+                                                        speech_context,
+                                                        single_utterance,
+                                                        interim_results)
+        for response in responses:
+            results = getattr(response, 'results', [])
```
daspecster: See below (https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2680#discussion_r86615897)

docs/speech-usage.rst
```diff
@@ -151,5 +151,84 @@ words to the vocabulary of the recognizer.
...
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    ...     response = list(client.streaming_recognize(sample))
+    ...     print(response[0].transcript)
+    'hello'
+    ...     print(response[0].confidence)
+    0.973458576
+
+
+By default the recognizer will perform continuous recognition
```
daspecster: Switched to "API".

docs/speech-usage.rst
```diff
@@ -151,5 +151,84 @@ words to the vocabulary of the recognizer.
...
+(continuing to process audio even if the user pauses speaking) until the client
+closes the output stream or when the maximum time limit has been reached.
```
daspecster: >until the client closes the output stream or until the maximum time limit has
been reached.

docs/speech-usage.rst
```diff
@@ -151,5 +151,84 @@ words to the vocabulary of the recognizer.
...
+    >>> client = speech.Client()
+    >>> with open('./hello.wav', 'rb') as stream:
+    ...     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    ...     for response in client.streaming_recognize(sample,
+    ...                                                interim_results=True):
+    ...         print('=' * 20)
+    ...         print(response[0].transcript)
+    ...         print(response[0].confidence)
```
daspecster: `response` is probably a bad name. It's really `alternatives`.

docs/speech-usage.rst
```diff
@@ -151,5 +151,84 @@ words to the vocabulary of the recognizer.
...
+    >>>     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    >>>     response = client.streaming_recognize(sample,
+    ...                                           single_utterance=True)
+    >>> results = list(response)
+    >>> print(results[0].transcript)
+    hello
+    >>> print(results[0].confidence)
+    0.96523453546
```
daspecster: Although I haven't seen it, it could be 2 if `max_alternatives` is set to 2. In practice I've just seen one.

daspecster:
@dhermes LMK if you have anything else for this. I've tried to give answer to the questions you had. The main outstanding issue has to do with managing the content coming into the [`Sample`](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2680#discussion_r86612107).

docs/speech-usage.rst
```diff
@@ -151,5 +151,85 @@ words to the vocabulary of the recognizer.
...
+
+See: `Single Utterance`_
+
+.. code-block:: python
+
+    >>> with open('./hello_pause_goodbye.wav', 'rb') as stream:
+    >>>     sample = client.sample(content=stream,
+    ...                            encoding=speech.Encoding.LINEAR16,
+    ...                            sample_rate=16000)
+    >>>     response = client.streaming_recognize(sample,
```
dhermes: 1. `>>>` --> `...`
1. `response` isn't the greatest name here is it? It's an iterator and could be many "responses"?

docs/speech-usage.rst
```diff
@@ -151,5 +151,85 @@ words to the vocabulary of the recognizer.
...
+
+If you only want to recognize a single utterance you can set
+ ``single_utterance`` to :data:`True` and only one result will be returned.
+
+See: `Single Utterance`_
+
+.. code-block:: python
+
+    >>> with open('./hello_pause_goodbye.wav', 'rb') as stream:
+    >>>     sample = client.sample(content=stream,
```
dhermes: `>>>` --> `...`

speech/google/cloud/speech/_gax.py
dhermes: Probably? But that should be it's own tiny PR so we can discuss the merits. Or you could add some features to check if `content` is a stream or not? Or you could just convert it to a stream if it isn't?

speech/google/cloud/speech/client.py
dhermes: It does not clear it up. Let's chat in Hangouts about this (I just want to make real API calls with the GAPIC layer to get a feel for it).

speech/google/cloud/speech/sample.py
dhermes: Just remove this change from the PR. It's noise.

speech/unit_tests/test_client.py
```diff
@@ -33,15 +26,26 @@ def _make_result(alternatives):
                 ) for alternative in alternatives
             ],
         )
+    else:
+        return cloud_speech_pb2.SpeechRecognitionResult(alternatives=[])
```
dhermes: Competing implementation. Rather than put the `return` in two branches, do the following at the beginning:

```python
if alternatives is None:
    alternatives = []
```

or even better just use an empty tuple (isn't mutable) as the default for `alternatives`

speech/unit_tests/test_client.py
```diff
@@ -15,16 +15,9 @@
...
-    AUDIO_SOURCE_URI = 'gs://sample-bucket/sample-recording.flac'
-    AUDIO_CONTENT = '/9j/4QNURXhpZgAASUkq'
-
-    @staticmethod
-    def _make_result(alternatives):
-        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
-
+def _make_result(alternatives=None):
+    from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+    if alternatives:
```
dhermes: Always check for `None` instead of checking false-y:

```python
if alternatives is not None:
    ...
```

See [PEP8][1]:

> Comparisons to singletons like `None` should always be done with `is` or `is not`, never the equality operators.
>
> Also, beware of writing `if x` when you really mean `if x is not None`

[1]: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

speech/unit_tests/test_client.py
```diff
@@ -33,15 +26,26 @@ def _make_result(alternatives):
...
-        from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
 
+def _make_sync_response(*results):
+    from google.cloud.grpc.speech.v1beta1 import cloud_speech_pb2
+    if results:
         response = cloud_speech_pb2.SyncRecognizeResponse(
             results=results,
         )
+    else:
+        response = cloud_speech_pb2.SyncRecognizeResponse()
```
dhermes: This branch isn't needed. The default of `results` is an empty tuple.

speech/unit_tests/test_client.py
```diff
@@ -33,15 +26,26 @@ def _make_result(alternatives):
...
+    return response
 
-        return response
+
+class TestClient(unittest.TestCase):
+    SAMPLE_RATE = 16000
+    HINTS = ['hi']
```
dhermes: Python tip: don't use mutables (`list` here) as class attributes unless you meant to do it on purpose. A `tuple` would be better here. (It's a test, so no need to change if you don't feel like it, but it's a good habit to get into.)

system_tests/speech.py
dhermes: There has to be a better way. Do we have a contact with the backend team?

daspecster: I guess we do in that case? I mean, we're receiving hypothesis results so I don't know if those can really be absolutely predictable. Maybe with an emulator though?

speech/google/cloud/speech/client.py
daspecster: OK just ping me whenever.

speech/unit_tests/test_client.py
daspecster: I made this to circle back to it. https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2700

daspecster:
@dhermes I believe I addressed everything we talked about over chat.

Here are the things I think that *could* be outstanding.
- https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2680#discussion_r86839838
- https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2680#discussion_r86838227

daspecster:
Squashed!

dhermes:
LGTM*

daspecster:
* There is still some concern with the surface and how the API responses should be passed on to the consumer of our API.

See: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2709

PR #2681, opened 2016-11-04
Use mock library in natural language tests.

dhermes:
~~Also using PEP8 valid names in place of `_makeOne` and `_getTargetClass` and making all the `_get_target_class` methods a `@staticmethod` (since they don't use self).~~

@tseaver LMKWYT, I know you are somewhat unsure of the `mock` library, so I wanted to give it a spin here to see what you think.

### DIFF
--- a/language/unit_tests/test_client.py
+++ b/language/unit_tests/test_client.py
@@ -15,6 +15,15 @@
 import unittest
 
 
+def make_mock_credentials():
+    import mock
+    from oauth2client.client import GoogleCredentials
+
+    credentials = mock.Mock(spec=GoogleCredentials)
+    credentials.create_scoped_required.return_value = False
+    return credentials
+
+
 class TestClient(unittest.TestCase):
 
     @staticmethod
@@ -28,7 +37,7 @@ class TestClient(unittest.TestCase):
     def test_ctor(self):
         from google.cloud.language.connection import Connection
 
-        creds = _Credentials()
+        creds = make_mock_credentials()
         http = object()
         client = self._make_one(credentials=creds, http=http)
         self.assertIsInstance(client._connection, Connection)
@@ -38,7 +47,7 @@ class TestClient(unittest.TestCase):
     def test_document_from_text_factory(self):
         from google.cloud.language.document import Document
 
-        creds = _Credentials()
+        creds = make_mock_credentials()
         client = self._make_one(credentials=creds, http=object())
 
         content = 'abc'
@@ -53,7 +62,7 @@ class TestClient(unittest.TestCase):
         self.assertEqual(document.language, language)
 
     def test_document_from_text_factory_failure(self):
-        creds = _Credentials()
+        creds = make_mock_credentials()
         client = self._make_one(credentials=creds, http=object())
 
         with self.assertRaises(TypeError):
@@ -62,7 +71,7 @@ class TestClient(unittest.TestCase):
     def test_document_from_html_factory(self):
         from google.cloud.language.document import Document
 
-        creds = _Credentials()
+        creds = make_mock_credentials()
         client = self._make_one(credentials=creds, http=object())
 
         content = '<html>abc</html>'
@@ -77,7 +86,7 @@ class TestClient(unittest.TestCase):
         self.assertEqual(document.language, language)
 
     def test_document_from_html_factory_failure(self):
-        creds = _Credentials()
+        creds = make_mock_credentials()
         client = self._make_one(credentials=creds, http=object())
 
         with self.assertRaises(TypeError):
@@ -86,7 +95,7 @@ class TestClient(unittest.TestCase):
     def test_document_from_url_factory(self):
         from google.cloud.language.document import Document
 
-        creds = _Credentials()
+        creds = make_mock_credentials()
         client = self._make_one(credentials=creds, http=object())
 
         gcs_url = 'gs://my-text-bucket/sentiment-me.txt'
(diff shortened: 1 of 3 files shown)


daspecster:
I'm really curious about a critique of the mock lib. I've been wanting to use this here but vaguely remember it being discouraged.

dhermes:
Nope, not discouraged. It became very mature and useful in Python 3 as `unittest.mock` and then was back-ported to Python 2.7 as https://pypi.python.org/pypi/mock

language/unit_tests/test_client.py
```diff
@@ -17,28 +17,33 @@
 
 class TestClient(unittest.TestCase):
 
-    def _getTargetClass(self):
+    @staticmethod
+    def _get_target_class():
```
tseaver: If we're going down this path (of debatable value), then it would seem better to just do it via a script across the whole repo.  Same for the `_makeOne` -> `_make_one` change below, and I assume `_callFUT` -> `_call_fut` where appropriate.

tseaver:
@dhermes Summary:  I particularly dislike the "magic" bits of `mock`, which always leaves the test assertions in a "maze of twisty passages, all alike."  I like mocked-up things to fail with AttributeErrors, etc.

If used carefully in non-magic mode, the constructed mocks are similar to our hand-rolled ones:  they take fewer lines, at the cost of some instant comprehension.

Using `mock` as a stdlib-based way to do what our current `_Monkey` does seems fine.

language/unit_tests/test_client.py
```diff
@@ -17,28 +17,33 @@
...
         self.assertIs(client.connection.http, http)
 
     def test_document_from_text_factory(self):
+        import mock
         from google.cloud.language.document import Document
 
-        creds = _Credentials()
-        client = self._makeOne(credentials=creds, http=object())
+        creds = mock.Mock()
+        creds.create_scoped_required.return_value = False
```
tseaver: Not a win over our `_Credentials` class IMO:  it is noisier and less comprehensible.  Also, without setting a `spec` on the mock, it is willing to return attrs promiscuously, which is a Bad Thing.

language/unit_tests/test_document.py
```diff
@@ -228,9 +259,10 @@ def test_analyze_entities(self):
             ],
             'language': 'en-US',
         }
-        connection = _Connection(response)
-        client = _Client(connection=connection)
-        document = self._makeOne(client, content)
+        connection = mock.Mock()
+        connection.api_request.return_value = response
+        client = mock.Mock(connection=connection)
```
tseaver: Promiscuous again (I won't repeat further).

language/unit_tests/test_document.py
```diff
@@ -242,10 +274,10 @@ def test_analyze_entities(self):
...
 
         # Verify the request.
-        self.assertEqual(len(connection._requested), 1)
-        req = connection._requested[0]
-        self.assertEqual(req['path'], 'analyzeEntities')
-        self.assertEqual(req['method'], 'POST')
+        expected = self._expected_data(
+            content, encoding_type=Encoding.UTF8)
+        connection.api_request.assert_called_once_with(
+            path='analyzeEntities', method='POST', data=expected)
```
tseaver: Not terrible, but not really much improvement, either.

language/unit_tests/test_client.py
dhermes: 1. Will do it with a script
1. I am unclear why it is of debatable value. PEP8 exists.

language/unit_tests/test_client.py
dhermes: I am refactoring with a `spec`.

dhermes:
@tseaver PTAL

language/unit_tests/test_client.py
```diff
@@ -26,19 +26,25 @@ def _make_one(self, *args, **kw):
         return self._get_target_class()(*args, **kw)
 
     def test_ctor(self):
+        import mock
+        from oauth2client.client import GoogleCredentials
         from google.cloud.language.connection import Connection
 
-        creds = _Credentials()
+        creds = mock.Mock(spec=GoogleCredentials)
+        creds.create_scoped_required.return_value = False
```
tseaver: Factor this into a helper method / function?

tseaver: And use keyword args to the `Mock` constructor?

dhermes: @tseaver Before I write the helper I want to confirm how you mean with `kwargs`. ISTM it must be a layer of nesting:

```python
creds = mock.Mock(
    spec=GoogleCredentials,
    create_scoped_required=mock.Mock(return_value=False))
```

unless there is some other syntax I'm unaware of?

tseaver: Ignore the kwargs bit for the return-value parts:  the helper function / method would just take the desired 'create_scope_required` value and configure it appropriately.

dhermes: Gotcha. I though

> And use keyword args to the `Mock` constructor

was referring to how I use `mock.Mock`

dhermes:
After rebase (and #2703) the `spec` saved the day:

```
AttributeError: Mock object has no attribute '_connection'
```

dhermes:
@tseaver I have re-factored as requested (note that there is no need for kwargs since "the desired 'create_scope_required` value" is always `False`). PTAL.

tseaver:
LGTM

PR #2689, opened 2016-11-04
Adding docs for the Runtime Config API.

dhermes:
The runtimeconfig-usage doc (or the `runtimeconfig/google/cloud/runtimeconfig/__init__.py` file) should add some examples of library usage.

@tswast once this is merged, care to take the reins on adding some usage examples?

### DIFF
--- a/docs/runtimeconfig-variable.rst
+++ b/docs/runtimeconfig-variable.rst
@@ -1,0 +1,6 @@
+Variables
+~~~~~~~~~
+
+.. automodule:: google.cloud.runtimeconfig.variable
+  :members:
+  :show-inheritance:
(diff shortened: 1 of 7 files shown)


dhermes:
Note the only failure here has been resolved in #2020, which @daspecster just merged. This is good to merge if anyone can review.

docs/runtimeconfig-variable.rst
```diff
@@ -0,0 +1,6 @@
+Variables
+~~~~~~~~~
```
daspecster: Why not `=`'s?

dhermes: No reason in particular. `dns-zone.rst` use `~` and I used that as a template.

Issue #2606, opened 2016-10-25
Drop Connection classes (or at least make them non-public)


dhermes:
The Logging / Pub/Sub / Datastore packages all have custom `API` objects that do that actual communication with the server. Now `Connection` is just a container for some members that those API objects require. Should we remove these base classes and make any shared behavior from the core `Connection` into helper **functions** instead of a shared base?

dhermes:
Just chatted with @tseaver about this and we both seem to agree, the answer is "yes" to "Should we...".

Sending a PR for BigQuery, Datastore, Logging, Pub/Sub and Storage ASAP.

jonparrott:
With the switch to google-auth, there is even less code in the `Connection` class - more fuel for this fire.

PR #2690, opened 2016-11-04
Renaming connection module as _http in 5 packages.

dhermes:
The packages are BigQuery, Datastore, Logging, Pub/Sub and Storage.
    
The rename is in advance of a larger re-factor (#2606). But so long as the connections are not public, the re-factor can happen without user-facing implications.

I'd also like to make all the `connection` attributes on instances into `_connection`. @tseaver Is it OK to put that change in this PR?


datastore/google/cloud/datastore/batch.py
```diff
@@ -107,7 +107,7 @@ def namespace(self):
     def connection(self):
         """Getter for connection over which the batch will run.
 
-        :rtype: :class:`google.cloud.datastore.connection.Connection`
+        :rtype: :class:`google.cloud.datastore._http.Connection`
         :returns: The connection over which the batch will run.
         """
         return self._client.connection
```
tseaver: Should this property even exist, given that we are hiding connections?

dhermes: Nope. Will follow up with some more `connection`-privatization in another PR to keep the scope low here.

PR #2696, opened 2016-11-06
Handle missing nanos.

dhermes:
See https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1783#issuecomment-258663547


daspecster:
This LGTM but I imagine you'll want @tseaver to look at it as well.

dhermes:
I doubt there is a single person or team responsible.

PR #2703, opened 2016-11-08
Making base connection module non-public and making connection attribute non-public

dhermes:
I have held off on renaming `google.cloud.${PACKAGE}.connection` to `_http` for the 5 packages in #2690, but I can do that in this PR if desired. The PR is already pretty large though.


dhermes:
@tseaver PTAL?

tseaver:
Hmm, I thought we'd approved this already.

dhermes:
You're probably thinking about #2690. Shall I merge then?

tseaver:
Yes.

PR #2706, opened 2016-11-08
Migrate datastore iterator to use base iterator class

dhermes:
I considered

- Making `offset` and `end_cursor` read-only public properties on the subclass
- Making a `@property` alias `start_cursor` that just returns the value of `next_page_token`
- Making a `@property` alias `limit` that just returns the value of `max_results`

I elected not to make anything **new** public since the original implementation didn't.

@jonparrott PTAL as well (do you think this will be too jarring a change?).

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -464,6 +464,23 @@ class Client(_BaseClient, _ClientProjectMixin):
           >>> for entity in query_iter:
           ...     do_something(entity)
 
+        or manually page through results
+
+        .. code-block:: python
+
+          >>> query_iter = query.fetch(start_cursor='2mdd223i944')
+          >>> pages = query_iter.pages
+          >>>
+          >>> first_page = next(pages)
+          >>> first_page_entities = list(first_page)
+          >>> query_iter.next_page_token
+          'abc-some-cursor'
+          >>>
+          >>> second_page = next(pages)
+          >>> second_page_entities = list(second_page)
+          >>> query_iter.next_page_token is None
+          True
+
         Under the hood this is doing:
 
         .. code-block:: python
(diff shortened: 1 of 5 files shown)


datastore/google/cloud/datastore/client.py
```diff
@@ -456,19 +456,13 @@ def query(self, **kwargs):
           >>> query = client.query(kind='MyKind')
           >>> query.add_filter('property', '=', 'val')
 
-        Using the query iterator's
-        :meth:`~google.cloud.datastore.query.Iterator.next_page` method:
+        Using the query iterator
```
jonparrott: Since using cursors is a very common pattern in datastore, can you add a quick sample here showing how to accomplish that?

dhermes: @jonparrott PTAL at the latest commit

jonparrott:
@dhermes thanks, the same is clear and doesn't make a common use case too difficult. I'm cool with this.

dhermes:
@daspecster @tseaver PTAL

datastore/google/cloud/datastore/client.py
```diff
@@ -456,19 +456,30 @@ def query(self, **kwargs):
...
+          >>> first_page_entities = list(first_page)
+          >>> query_iter.next_page_token
+          'abc-some-cursor'
+          >>>
+          >>> second_page = next(pages)
+          >>> second_page_entities = list(second_page)
+          >>> query_iter.next_page_token is None
+          True
 
         Under the hood this is doing:
```
daspecster: @dhermes does this section need to be updated with these changes?

dhermes: Nope.

dhermes:
Going to merge (waiting 24h for other feedback)

PR #2710, opened 2016-11-09
Use PEP8 names in test helper methods

dhermes:
From discussion with @tseaver in #2681

Most of this PR was created from the command line, see the commit messages for the commands used to create the commits.

I also wanted to turn off the `no-self-use` disable in the test Pylint RC but it had too many offenders left to put in this PR.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -174,7 +174,7 @@ Coding Style
 
 Exceptions to PEP8:
 
-- Many unit tests use a helper method, ``_callFUT`` ("FUT" is short for
+- Many unit tests use a helper method, ``_call_fut`` ("FUT" is short for
   "Function-Under-Test"), which is PEP8-incompliant, but more readable.
   Some also use a local variable, ``MUT`` (short for "Module-Under-Test").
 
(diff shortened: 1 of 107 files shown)


daspecster:
@dhermes was this scripted at all?

dhermes:
> Also, my internet may be slow or something but I can't even get the files view to load ATM.

It's probably not your internet, just the browser failing because of too much memory needed. May be easier to check out my branch locally.

----

> was this scripted at all?

See [my comment][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2710#issue-188167835

daspecster:
@dhermes oh I didn't expand the commit messages. That's cool.

daspecster:
What would it take to enable `invalid-name`, which we [ignore](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1abfcbec1b44f58096d7a6d340c0799f8fdf1f1b/scripts/run_pylint.py#L58), after this?

We could just update the regex that `invalid-name` uses. Just a thought.

daspecster:
Also this LGTM after the rebase and travs going green.

dhermes:
> What would it take to enable `invalid-name`, which we [ignore](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1abfcbec1b44f58096d7a6d340c0799f8fdf1f1b/scripts/run_pylint.py#L58), after this?

I'd imagine a lot. We could probably get some mileage out of [tweaking regexes][1].

[1]: https://github.com/dhermes/bezier/blob/99b47998745df66986728560349cd33f2e0a80cf/pylintrc#L25-L26

dhermes:
Just rebased. Will merge if/when at least one CI goes green.

PR #2711, opened 2016-11-09
Dropping usage of _Monkey in favor of mock.patch.

dhermes:
This was done only in bigquery, datastore and storage packages. Still needs updates in bigtable, core, logging, monitoring, pubsub and speech.

From discussion with @tseaver in #2681.

### DIFF
--- a/datastore/unit_tests/test__http.py
+++ b/datastore/unit_tests/test__http.py
@@ -181,8 +181,7 @@ class Test_DatastoreAPIOverGRPC(unittest.TestCase):
         return _DatastoreAPIOverGRPC
 
     def _make_one(self, stub, connection=None, secure=True, mock_args=None):
-        from google.cloud._testing import _Monkey
-        from google.cloud.datastore import _http as MUT
+        import mock
 
         if connection is None:
             connection = _Connection(None)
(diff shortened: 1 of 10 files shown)


datastore/unit_tests/test__http.py
```diff
@@ -417,16 +420,18 @@ def mock_api(connection):
             connections.append(connection)
             return return_val
 
-        with _Monkey(MUT, _DatastoreAPIOverHttp=mock_api):
+        patch = mock.patch(
```
daspecster: Why not...
```python
method_name = 'google.cloud.datastore._http._DatastoreAPIOverHttp'
with mock.patch(method_name, new=mock_api):
```

dhermes: @daspecster That's fine with me. I have no preference here.

daspecster: Ok, I was mostly curious if you had a reason for expanding it out.

dhermes:
@tseaver Any reservations?

datastore/unit_tests/test__http.py
dhermes: Nope no reason.

tseaver:
@dhermes LGTM after rebase, Travis OK.

Issue #2697, opened 2016-11-06
vision API manual detection documented but not implemented


ptone:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/docs/vision-usage.rst#manual-detection

wah?

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/vision/google/cloud/vision/image.py#L88

documentation written by @daspecster
https://github.com/GoogleCloudPlatform/google-cloud-python/commit/95a55c145bd2e16e80e34e180682932c8a16199a

daspecster:
@ptone, sorry about that. There was a lot of froth in deciding the surface and this slipped by me.

In the meantime you can use [`client.annotate`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/15eb09183333f4c3cebc59bd9600a43ae6c3715c/vision/google/cloud/vision/client.py#L85) to do multiple detections.  I will get a PR to correct this ASAP!

PR #2712, opened 2016-11-09
Refactor _detect_annotation() to support all annotation types.

daspecster:
This is a step towards adding the manual detect method.

See: #2697

### DIFF
--- a/vision/google/cloud/vision/image.py
+++ b/vision/google/cloud/vision/image.py
@@ -27,6 +27,25 @@ from google.cloud.vision.color import ImagePropertiesAnnotation
 from google.cloud.vision.safe import SafeSearchAnnotation
 
 
+_FACE_DETECTION = 'FACE_DETECTION'
+_IMAGE_PROPERTIES = 'IMAGE_PROPERTIES'
+_LABEL_DETECTION = 'LABEL_DETECTION'
+_LANDMARK_DETECTION = 'LANDMARK_DETECTION'
+_LOGO_DETECTION = 'LOGO_DETECTION'
+_SAFE_SEARCH_DETECTION = 'SAFE_SEARCH_DETECTION'
+_TEXT_DETECTION = 'TEXT_DETECTION'
+
+_REVERSE_TYPES = {
+    _FACE_DETECTION: 'faceAnnotations',
+    _IMAGE_PROPERTIES: 'imagePropertiesAnnotation',
+    _LABEL_DETECTION: 'labelAnnotations',
+    _LANDMARK_DETECTION: 'landmarkAnnotations',
+    _LOGO_DETECTION: 'logoAnnotations',
+    _SAFE_SEARCH_DETECTION: 'safeSearchAnnotation',
+    _TEXT_DETECTION: 'textAnnotations',
+}
+
+
 class Image(object):
     """Image representation containing information to be annotate.
 
@@ -85,28 +104,25 @@ class Image(object):
         """
         return self._source
 
-    def _detect_annotation(self, feature):
+    def _detect_annotation(self, features):
         """Generic method for detecting a single annotation.
 
-        :type feature: :class:`~google.cloud.vision.feature.Feature`
-        :param feature: The ``Feature`` indication the type of annotation to
-                        perform.
+        :type features: list
+        :param features: List of :class:`~google.cloud.vision.feature.Feature`
+                         indicating the type of annotations to perform.
 
         :rtype: list
         :returns: List of
-                  :class:`~google.cloud.vision.entity.EntityAnnotation`.
+                  :class:`~google.cloud.vision.entity.EntityAnnotation`,
+                  :class:`~google.cloud.vision.face.Face`,
+                  :class:`~google.cloud.vision.color.ImagePropertiesAnnotation`,
+                  :class:`~google.cloud.vision.sage.SafeSearchAnnotation`,
         """
-        reverse_types = {
-            'LABEL_DETECTION': 'labelAnnotations',
-            'LANDMARK_DETECTION': 'landmarkAnnotations',
-            'LOGO_DETECTION': 'logoAnnotations',
-            'TEXT_DETECTION': 'textAnnotations',
-        }
         detected_objects = []
-        result = self.client.annotate(self, [feature])
-        for response in result[reverse_types[feature.feature_type]]:
-            detected_object = EntityAnnotation.from_api_repr(response)
-            detected_objects.append(detected_object)
+        results = self.client.annotate(self, features)
+        for feature in features:
+            detected_objects.extend(
+                _entity_from_response_type(feature.feature_type, results))
         return detected_objects
 
     def detect_faces(self, limit=10):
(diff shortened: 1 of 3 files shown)


vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
         return self._source
 
-    def _detect_annotation(self, feature):
+    def _detect_annotation(self, features):
         """Generic method for detecting a single annotation.
 
-        :type feature: :class:`~google.cloud.vision.feature.Feature`
-        :param feature: The ``Feature`` indication the type of annotation to
-                        perform.
+        :type features: list of:class:`~google.cloud.vision.feature.Feature`
```
dhermes: Just use `list` here and explain the type of each element in the list in the `param` doc

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
         """
         return self._source
 
-    def _detect_annotation(self, feature):
+    def _detect_annotation(self, features):
```
dhermes: I can't find any place in this PR where you've passed in a list with anything other than one element. Is the change to `features` not YAGNI?

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+        if len(detected_objects) == 1:
+            return detected_objects[0]
+        return detected_objects
+
+    @staticmethod
+    def _entity_from_response_type(feature_type, results):
+        """Convert a JSON result to an entity type based on the feature."""
+
         reverse_types = {
+            'FACE_DETECTION': 'faceAnnotations',
```
dhermes: I realize this was already here, but please move this map to module-level scope and name it `_REVERSE_TYPES` (module-level mutables are scary, but it's non-public so if users break it they buy it)

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+        results = self.client.annotate(self, features)
+        for feature in features:
+            detected_objects.extend(
+                self._entity_from_response_type(feature.feature_type, results))
+
+        if len(detected_objects) == 1:
+            return detected_objects[0]
+        return detected_objects
+
+    @staticmethod
```
dhermes: Prefer helper function over `staticmethod`

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+                  :class:`~google.cloud.vision.sage.SafeSearchAnnotation`,
         """
+        detected_objects = []
+        results = self.client.annotate(self, features)
+        for feature in features:
+            detected_objects.extend(
+                self._entity_from_response_type(feature.feature_type, results))
+
+        if len(detected_objects) == 1:
+            return detected_objects[0]
```
dhermes: Don't do this. Just return a list every time.

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+            for face in results[feature_key]:
+                detected_objects.append(Face.from_api_repr(face))
+        elif feature_type == 'IMAGE_PROPERTIES':
+            detected_objects.append(
+                ImagePropertiesAnnotation.from_api_repr(results[feature_key]))
+        elif feature_type == 'SAFE_SEARCH_DETECTION':
+            result = results[feature_key]
+            detected_objects.append(SafeSearchAnnotation.from_api_repr(result))
+        else:
+            for result in results[feature_key]:
```
dhermes: So this branch is meant to cover 4 feature types?

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
             'TEXT_DETECTION': 'textAnnotations',
         }
         detected_objects = []
-        result = self.client.annotate(self, [feature])
-        for response in result[reverse_types[feature.feature_type]]:
-            detected_object = EntityAnnotation.from_api_repr(response)
-            detected_objects.append(detected_object)
+        feature_key = reverse_types[feature_type]
+
+        if feature_type == 'FACE_DETECTION':
```
dhermes: Make this string a module-level (non-public) constant since you re-use it. Ditto on the two below that you use twice.

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+            for face in results[feature_key]:
+                detected_objects.append(Face.from_api_repr(face))
```
dhermes: Equivalent code using a generator expression (if you like):

```python
detected_objects.extend(
    Face.from_api_repr(face) for face in results[feature_key])
```

vision/google/cloud/vision/image.py
```diff
@@ -85,28 +85,58 @@ def source(self):
...
+        elif feature_type == 'IMAGE_PROPERTIES':
+            detected_objects.append(
+                ImagePropertiesAnnotation.from_api_repr(results[feature_key]))
+        elif feature_type == 'SAFE_SEARCH_DETECTION':
+            result = results[feature_key]
+            detected_objects.append(SafeSearchAnnotation.from_api_repr(result))
+        else:
+            for result in results[feature_key]:
+                detected_objects.append(EntityAnnotation.from_api_repr(result))
```
dhermes: Ditto here on the `extend()` with generator expression if you're into that

vision/google/cloud/vision/image.py
daspecster: Yeah, `LABEL_DETECTION`, `LANDMARK_DETECTION`, `LOGO_DETECTION` and `TEXT_DETECTION` all return `EntityAnnotation`s.

vision/google/cloud/vision/image.py
daspecster: Ok yeah, not my best pick for splitting things up.

daspecster:
Regarding your comment about `features = [features]`, if I do that, then I'll need to put a check there as well.

```python
if not isinstance(features, list):
    features = [features]
```

vision/google/cloud/vision/image.py
```diff
@@ -27,6 +27,21 @@
...
+
+_REVERSE_TYPES = {
+    _FACE_DETECTION: 'faceAnnotations',
+    _IMAGE_PROPERTIES: 'imagePropertiesAnnotation',
+    'LABEL_DETECTION': 'labelAnnotations',
+    'LANDMARK_DETECTION': 'landmarkAnnotations',
+    'LOGO_DETECTION': 'logoAnnotations',
+    _SAFE_SEARCH_DETECTION: 'safeSearchAnnotation',
+    'TEXT_DETECTION': 'textAnnotations',
+}
```
tseaver: Why not define constants for all of them?

dhermes:
> if I do that, then I'll need to put a check there as well.

No, just make your callers obey the signature. So if it accepts one feature (which all the current callers so) then they send one feature.

However, as I mentioned before, just leave the change in.

daspecster:
@dhermes this is lacking clarity as you mentioned before because I'm preparing for #2697. 
`detect()` will need to pass a list of features. Which will be clear in the next PR for this.

daspecster:
Squished. Once the build goes green I'll merge.

Issue #885, opened 2015-05-19
Add language switcher drop-down to landing page


jgeewax:
The other guys (ie http://googlecloudplatform.github.io/gcloud-node/#/) link to us on their landing page, might be nice to return the favor.

ludoch:
That would be needed for all the runtimes to be consistent. Changing runtimes from there should in fact not induce UI/css style difference.

jgeewax:
@ludoch : That makes the assumption that all of the libraries follow the exact same styling. Go and Java, (I believe, as of now) don't. I think it's a reasonable ask that the site in gcloud-python has the dropdown, but we can debate the others later on in their respective repositories.

dhermes:
Which ones are public and can be linked to? Shouldn't more than just Ruby + Python + Node be linked?

jgeewax:
Right now, Node + Ruby + Python are linkable. We have to get Java and Go into a linkable state before we can add them.

PR #2713, opened 2016-11-09
Use v1 of Resource Manager

carise:
Resource Manager (for both organizations and projects) is now v1, updating the API paths to reflect that.

reference: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1789

@dhermes @jgeewax


dhermes:
Ha, that was easy! Do you mind squashing your commits / or we can do it in the GitHub UI.

I am worried: hasn't the interface changed.

Since we don't have system tests for resource manager (I believe it is because they didn't support service accounts when the original implementation was done) this won't quickly detect any breakages.

carise:
Correct, the interface hasn't changed noticeably, which worries me as well. I am pretty new to GCP open source contrib, but I would be happy to work on a system test... How should I do that?

dhermes:
Yeah there might be a lot of work here (beyond just the version change). If you want to create a system test you could peak at one of the smaller ones as an example: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/cd9ac338065e7c46e7a3176864f62ad00b9ace83/system_tests/language.py

carise:
No worries, I am happy to work on it. Thanks for the pointers :)

carise:
Just a small update here. I am still working on this, trying to get v1 working for projects. I've hit a little roadblock as I've hit the max # of projects that my service account can create; trying to work it out with support.

dhermes:
Thanks for the heads up. @jonparrott may be able to help with whitelisting?

carise:
I think I might get it handled soon :) but I will reach out if I do need additional help. Thanks!

carise:
Hmm. So it looks like I need to submit a design doc to the team handling the whitelisting and set up a meeting with them... Do you think I should ask @jonparrott if I can cc him on the issue, to see if we can keep it moving?

dhermes:
Indeed. Hopefully the presence of this library (with 450+ stars and 200+ forks) merits the work, unclear why so much red tape needed.

carise:
So, good news. I got my quota raised! Currently trying to finish up the system tests. I am estimating that it will be ready for review in the next day or so.

carise:
I'm running into a few issues with the tests -- sorry in advance for the noob confusion :(

1. I can't figure out why the python3.5 unit tests are failing. It seems to be failing at the install step. https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/jobs/183792322#L276 
2. Likewise, the system-tests3 are failing, but I don't understand why. https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/jobs/183792322#L427 

Has anyone run into those issues?

As part of using v1 for projects.create(), I had to add 2 new classes, Operation and Resource. (https://cloud.google.com/resource-manager/reference/rest/v1/projects/create) This is because calling projects.create() returns an Operation vs the v1beta1 which returns a Project. (https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/create)

Also, when creating a project, I had to declare the project's parent (organization) explicitly when testing with the service account. I am wondering, how would I make google-cloud-python its own module so I can try it out (as if I am the end user trying to use it)? I would like to test it using default application credentials.

Lastly, I will work on updating documentation, but should I commit those changes into this (already pretty large) PR?

tseaver:
@carise 

> I'm running into a few issues with the tests -- sorry in advance for the noob confusion :(
>
>   I can't figure out why the python3.5 unit tests are failing. It seems to be failing at the install step. https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/jobs/183792322#L276
>   Likewise, the system-tests3 are failing, but I don't understand why. https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/jobs/183792322#L427
>
> I have a virtualenv (from using mkvirtualenv) where I installed tox and with that virtualenv activated, I run "tox -e py35 -- resource_manager". But that fails with the same error as point 1 (above). It also fails when I am in the master branch of my fork.

> Has anyone run into those issues?

#2856 is the culprit here, which tracks [setuptools issue #885](https://github.com/pypa/setuptools/issues/885).  It appears we will have an updated setuptools release with a fix for that issue very soon.

tseaver:
> As part of using v1 for projects.create(), I had to add 2 new classes, Operation and Resource. (https://cloud.google.com/resource-manager/reference/rest/v1/projects/create) This is because calling projects.create() returns an Operation vs the v1beta1 which returns a Project. (https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/create)

We already have a [class for managing `google.longrunning.operations_pb2.Operation`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/aea151b83893656e43d563935de8ce93a46f310a/core/google/cloud/operation.py#L84).

carise:
@tseaver Thanks for the pointers. I'll fix the code to use the existing Operation!

carise:
Oh no, I just found this: https://github.com/googleapis/googleapis/issues/206

I'm running into an issue with the Operation not being able to find type.googleapis.com/google.cloudresourcemanager.v1.ProjectCreationStatus -- sounds like a missing protobuf descriptor (like [this issue](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2572)). Going to follow up with the other people on the thread...

dhermes:
RE: "Oh no", that isn't a problem. That PR is for a low-level surface, not the high-level one here.

RE: protobuf issue, that isn't the same problem. Our operation class relies on manually registered types. For example:

carise:
Oh, I see. So the correct thing to do is manually register the google.cloudresourcemanager.v1.ProjectCreationStatus type which is in the google.cloudresourcemanager.v1 pb?

dhermes:
Correct

carise:
Ok, thank you. And for the google.cloudresourcemanager.v1 pb, I should be able to find that in the [googleapis repo](https://github.com/googleapis/googleapis), correct?

dhermes:
Ah, it seems to not exist yet on PyPI, though the protos are probably public? /cc @bjwatson

daspecster:
@carise it looks like the Python 3.5 issues have been resolved!

bjwatson:
@dhermes @carise We've not published generated Python code yet for `cloudresourcemanager`. There are quite a few APIs ahead of it in our priority queue, and my understanding is that we're waiting until v2 to do work on this.

@carise @adamtanner I cc'd you on the relevant internal bug.

dhermes:
@bjwatson Sure thing. Not "high priority" for us either, just @carise took it on as a project.

Doesn't look like the protos are available publicly:
https://github.com/googleapis/googleapis/tree/5a903752cf7967059acdfbd19af295d671826332/google

If they were, I could help and make a script to add a `_generated` sub-package like we do in datastore and bigtable (though hopefully those will go away very soon).

carise:
@bjwatson Thanks for doing that. I'll follow up internally.

@daspecster Awesome! I'll merge in the new changes.

@dhermes It sounds like we have a few options here: 1) Abandon this PR and redo it when v2 is out, 2) Keep using v1beta1 for create and everything else on v1 (I think this is not really a good option), or 3) Don't include v1 projects.create() but keep everything else (delete, update, undelete). Not sure if I am missing anything here. What do you think?

dhermes:
@carise Let's chat privately? I don't quite know the lay of the land so would like to get a clearer picture

carise:
@dhermes Sure, I will contact you through the email in your github profile.

carise:
Closing this as it has been determined that the Resource Manager team will wait until v2 to publish the necessary protos.

PR #2714, opened 2016-11-10
Add stability information to streaming results.

daspecster:
Towards #2522.

This changes the surface, since `is_final` and `stability` are per-result and not a per-alternative.

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -171,10 +171,10 @@ speech data to possible text alternatives on the fly.
     ...     sample = client.sample(content=stream,
     ...                            encoding=speech.Encoding.LINEAR16,
     ...                            sample_rate=16000)
-    ...     alternatives = list(client.streaming_recognize(sample))
-    >>> print(alternatives[0].transcript)
+    ...     results = list(client.streaming_recognize(sample))
+    >>> print(results[0].alternatives[0].transcript)
     'hello'
-    >>> print(alternatives[0].confidence)
+    >>> print(results[0].alternatives[0].confidence)
     0.973458576
 
 
@@ -196,10 +196,10 @@ See: `Single Utterance`_
     ...                            sample_rate=16000)
     ...     responses = client.streaming_recognize(sample,
     ...                                            single_utterance=True)
-    ...     alternatives = list(responses)
-    >>> print(alternatives[0].transcript)
+    ...     results = list(responses)
+    >>> print(results[0].alternatives[0].transcript)
     hello
-    >>> print(alternatives[0].confidence)
+    >>> print(results[0].alternatives[0].confidence)
     0.96523453546
 
 
@@ -214,20 +214,28 @@ If ``interim_results`` is set to :data:`True`, interim results
     ...     sample = client.sample(content=stream,
     ...                            encoding=speech.Encoding.LINEAR16,
     ...                            sample_rate=16000)
-    ...     for alternatives in client.streaming_recognize(sample,
-    ...                                                    interim_results=True):
+    ...     for results in client.streaming_recognize(sample,
+    ...                                                interim_results=True):
     ...         print('=' * 20)
-    ...         print(alternatives[0].transcript)
-    ...         print(alternatives[0].confidence)
+    ...         print(results[0].alternatives[0].transcript)
+    ...         print(results[0].alternatives[0].confidence)
+    ...         print(results[0].is_final)
+    ...         print(results[0].stability)
     ====================
     'he'
     None
+    False
+    0.113245
     ====================
     'hell'
     None
+    False
+    0.132454
     ====================
     'hello'
     0.973458576
+    True
+    0.982345
 
 
 .. _Single Utterance: https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#streamingrecognitionconfig
--- a/speech/google/cloud/speech/client.py
+++ b/speech/google/cloud/speech/client.py
@@ -27,6 +27,7 @@ from google.cloud.speech.alternative import Alternative
 from google.cloud.speech.connection import Connection
 from google.cloud.speech.encoding import Encoding
 from google.cloud.speech.operation import Operation
+from google.cloud.speech.result import StreamingSpeechResult
 from google.cloud.speech.sample import Sample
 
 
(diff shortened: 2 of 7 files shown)


docs/speech-usage.rst
```diff
@@ -196,10 +196,10 @@ See: `Single Utterance`_
     ...                            sample_rate=16000)
     ...     responses = client.streaming_recognize(sample,
     ...                                            single_utterance=True)
-    ...     alternatives = list(responses)
-    >>> print(alternatives[0].transcript)
+    ...     results = list(responses)
+    >>> print(results[0].alternatives[0].transcript)
     hello
-    >>> print(alternatives[0].confidence)
+    >>> print(results[0].alternatives[0].confidence)
```
dhermes: This change indicates the snippet was broken before?

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
                                   profanity_filter, speech_context)
 
 
+class StreamingSpeechResult(object):
```
dhermes: Does this not want its own module?

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+    """Streaming speech result representation.
+
+    :type alternatives: list
+    :param alternatives: List of protobuf speech alternatives.
```
dhermes: You want the constructor to take pbs?

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+
+    :type is_final: bool
+    :param is_final: Boolean indicator of results finality.
+
+    :type stability: float
+    :param stability: 0.0-1.0 stability score for the results returned.
+
+    :rtype: :class:`~google.cloud.speech.client.StreamingSpeechResult`
+    :returns: Instance of ``StreamingSpeechResult``.
```
dhermes: You don't need / want an `rtype` and `returns` in the class docstring

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+    """
+    def __init__(self, alternatives, is_final=False, stability=0.0):
+        self._alternatives = [Alternative.from_pb(alternative)
+                              for alternative in alternatives]
```
dhermes: Move this into `from_pb`

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+        self._is_final = is_final
+        self._stability = stability
+
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct instance of ``StreamingSpeechResult``.
+
+        :type response: :class:`~google.cloud.grpc.speech.v1beta1\
+                               .cloud_speech_pb2.StreamingRecognizeResult`
+        :param response: Istance of ``StreamingRecognizeResult`` protobuf.
```
dhermes: Istance

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+
+        :rtype: :class:`~google.cloud.speech.client.StreamingSpeechResult`
```
dhermes: Since you are in the same module as the referenced class (obviously, since this classmethod is part of the class) you can refer to it as

```
:class:`StreamingSpeechResult`
```

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+    def alternatives(self):
+        """List of alternative transcripts.
+
+        :rtype: list of :class:`~google.cloud.speech.alternative.Alternative`
+        :returns: List of ``Alternative`` instances.
+        """
+        return self._alternatives
+
+    @property
+    def is_final(self):
```
dhermes: This (making it read-only) is overkill. Just set `self.is_final` on the instance.

speech/google/cloud/speech/client.py
```diff
@@ -299,6 +298,74 @@ def sync_recognize(self, sample, language_code=None,
...
+
+        :rtype: bool
+        :returns: True if this result is final and no more processing will
+                  occur. False if more processing can will be done and results
+                  may change.
+        """
+        return self._is_final
+
+    @property
+    def stability(self):
```
dhermes: Ditto on overkill

speech/unit_tests/test_client.py
```diff
@@ -521,15 +525,27 @@ def speech_api(channel=None):
 
         results = list(client.streaming_recognize(sample,
                                                   interim_results=True))
-        self.assertEqual(results[0], [])
-        self.assertEqual(results[1][0].transcript,
+
+        self.assertIsInstance(results[0], StreamingSpeechResult)
```
dhermes: Please add a length assertion for `results`

daspecster:
I'm totally open to ideas(regarding `StreamingSpeechResult` class). I had this class in a previous branch, and from discussions before, it sounded like you were for bringing protobuf data into our own classes as an abstraction.

docs/speech-usage.rst
daspecster: Nope, this is the surface change I was talking about.

This is an example of the current structure this PR changes vs before.
```diff
+ {
+     is_final: True
+     stability: 0.9234354
+     alternatives. []
+ }
- [
-   {
-      transcript: 'some transcript'
-      confidence: 0.9124354 
-    },
- ]
```

dhermes: I think I get what you're saying but a Python snippet wouldn't been more helpful

daspecster: I think you meant "would" instead of "wouldn't" so I'll make a python snippet.

```python
#  Previously 
for alternatives in client.streaming_recognize(sample, interim_results=True):
    print('=' * 20)
    print(alternatives[0].transcript)
    print(alternatives[0].confidence)
    ====================
    'hello'
    0.973458576
```

```python
# Now
for results in client.streaming_recognize(sample, interim_results=True):
    print('=' * 20)
    print(results[0].alternatives[0].transcript)
    print(results[0].alternatives[0].confidence)
    print(results[0].is_final)
    print(results[0].stability)
====================
'hello'
0.973458576
True
0.982345
```

Since the `is_final` and `stability` values don't make sense directly in the `alternative` data, they *should* go one level up AFAICT.

daspecster:
@dhermes CircleCI is winning my love.

Did you have anything else for this?

dhermes:
You also forgot to check in a new RST file for `result.py`

dhermes:
LGTM once you fix your b0rken docs situation

daspecster:
Everything went green. Squashed and I'll just make sure it goes green again and then I'll merge.

PR #2716, opened 2016-11-10
Add PyPI badges for speech

daspecster:
This adds the PyPI badges to the speech README.

I was waiting on this but since we have the placeholder package on PyPI I guess this can go anytime.


dhermes:
LGTM

PR #2720, opened 2016-11-11
Passing through COVERALLS_REPO_TOKEN to tox coveralls env.

dhermes:
Also re-factoring the tox config a little bit to reuse a base passenv rather than passing around the system-tests passenv everywhere.

See https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/119 for an example failure that shouldn't have failed (since the env. var. is set in the CircleCI UI).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,11 @@ envlist =
 [testing]
 deps =
     pytest
+passenv =
+    CI_*
+    CIRCLE*
+    GOOGLE_*
+    TRAVIS*
 localdeps =
     pip install --upgrade \
         {toxinidir}/core \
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -236,7 +248,9 @@ commands =
     python {toxinidir}/system_tests/attempt_system_tests.py {posargs}
 deps =
     {[testing]deps}
-passenv = CI_* CIRCLE* GOOGLE_* TRAVIS* encrypted_*
+passenv =
+    {[testing]passenv}
+    encrypted_*
```
daspecster: @dhermes what's this for?

dhermes: System tests. These are env. vars. set automatically by the Travis CLI too. See https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f348615f6e1eec92cedbbe4ea1327ba27d7100ee/system_tests/attempt_system_tests.py#L74-L76 and the module docstring in the same file

PR #2721, opened 2016-11-11
Adding quiet flag to pip command for local deps.

dhermes:


dhermes:
@daspecster This `--quiet` flag is only use for our local deps, so the versions are just what's in the source tree.

PR #2722, opened 2016-11-11
Fixing now failing language system tests.

dhermes:


system_tests/language.py
```diff
@@ -88,7 +88,7 @@ def _check_analyze_entities_result(self, entities):
         self.assertEqual(entity2.metadata, {})
         # Verify entity 3.
         self.assertEqual(entity3.name, self.NAME3)
-        self.assertEqual(entity3.entity_type, EntityType.WORK_OF_ART)
+        self.assertEqual(entity3.entity_type, EntityType.EVENT)
```
dhermes: Hilariously, this used to be `EVENT` and it flaked its way to `WORK_OF_ART`, which we had to address in #2587

Maybe the better fix would be:

```python
choices = (EntityType.EVENT, EntityType.WORK_OF_ART)
self.assertIn(entity3.entity_type, choices)
```

daspecster: Like what JJ said last Monday, these ML tests change as the systems learn. So as long as they return something that is one of the expected responses, I think it's good.  So I'm 👍 for the `choices` approach.

dhermes:
@daspecster PTAL

PR #2723, opened 2016-11-11
Removing local dependencies from lint tox env.

dhermes:
Also skipping the install (which would install from PyPI if the localdeps weren't added). The deps are not needed since the lint tools just reads the source.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -135,6 +135,7 @@ skip_install =
     py34: True
     py35: True
     cover: True
+    lint: True
 passenv =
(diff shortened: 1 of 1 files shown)


tseaver:
Hmm, I thought pylint was verifying imports:  does it skip even trying them (as opposed to reporting them) when we disable via `import-error`?

dhermes:
Pylint is mostly static analysis by design, I'm not sure how it verifies imports, but yes, turning off `import-error` is probably why this works.

PR #2724, opened 2016-11-11
Move speech methods to sample.

### LINKED ISSUE #2516: Move speech methods to `Sample`.
Speech API calls should be from the `Sample` object instead of directly from the client.

``` python
# Before
client.sync_recognize(sample)

# After
sample.sync_recognize()
```

See: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2483#discussion_r82463142

daspecster:
Cleans up the surface a little bit but moving `sync_recognize`, `async_recognize` and `streaming_recognize` to the `Sample` class.

Fixes #2516.

### DIFF
--- a/speech/google/cloud/speech/sample.py
+++ b/speech/google/cloud/speech/sample.py
@@ -15,6 +15,7 @@
 """Sample class to handle content for Google Cloud Speech API."""
 
 from google.cloud.speech.encoding import Encoding
+from google.cloud.speech.result import StreamingSpeechResult
 
 
 class Sample(object):
@@ -43,12 +44,16 @@ class Sample(object):
                         to 16000 Hz. If that's not possible, use the
                         native sample rate of the audio source (instead of
                         re-sampling).
+
+    :type client: :class:`~google.cloud.speech.client.Client`
+    :param client: (Optional) The client that owns this instance of sample.
     """
     default_encoding = Encoding.FLAC
     default_sample_rate = 16000
 
     def __init__(self, content=None, source_uri=None,
-                 encoding=None, sample_rate=None):
+                 encoding=None, sample_rate=None, client=None):
+        self._client = client
 
         no_source = content is None and source_uri is None
         both_source = content is not None and source_uri is not None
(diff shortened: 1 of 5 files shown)


speech/google/cloud/speech/sample.py
```diff
@@ -43,12 +44,16 @@ class Sample(object):
                         to 16000 Hz. If that's not possible, use the
                         native sample rate of the audio source (instead of
                         re-sampling).
+
+    :type client: :class:`~google.cloud.client.Client`
```
dhermes: Don't you want to refer to the speech client?

speech/google/cloud/speech/sample.py
```diff
@@ -43,12 +44,16 @@ class Sample(object):
...
+    :param client: Instance of ``Client``
```
dhermes: This is redundant. How about "the client that owns the current sample".

Also it is "(Optional) "

speech/google/cloud/speech/sample.py
```diff
@@ -112,3 +117,180 @@ def encoding(self):
         :returns: String value of Encoding type.
         """
         return self._encoding
+
+    def async_recognize(self, language_code=None, max_alternatives=None,
+                        profanity_filter=None, speech_context=None):
+        """Asychronous Recognize request to Google Speech API.
```
dhermes: 1. A ll just copy-pasta or do i need to read?
1.  Why not just put the remaining four args (`language_code`, etc.) into the constructor? Or if not in the constructor make a "query" class (or similar) since both sync and async have the same args

    ```python
    query = sample.query(
        language_code='en-US', max_alternatives=4)

    # async is a keyword in Python3.5 so might need a diff name
    query.async()
    query.sync()
    ```

speech/google/cloud/speech/sample.py
```diff
@@ -112,3 +117,180 @@ def encoding(self):
...
+
+        api = self._client.speech_api
+        responses = api.streaming_recognize(self, language_code,
+                                            max_alternatives, profanity_filter,
+                                            speech_context, single_utterance,
+                                            interim_results)
+        for response in responses:
+            for result in response.results:
+                if result.is_final or interim_results:
+                    yield StreamingSpeechResult.from_pb(result)
```
dhermes: Holy nesting

speech/google/cloud/speech/sample.py
daspecster: 1. Yeah I just copied them over and deleted references to the `sample` param.
2. I could, however, those settings don't really have to do with the `sample` so much as they do the request. Semantics but should I feel like those settings belong in `Sample`?

daspecster: Yeah, I think this is a consequence of supporting both GAX and REST, having two branches?

daspecster: There's a bunch of ways to package the data up and pass it around but I can't think of any that are good.

daspecster:
Squashed.

daspecster:
@dhermes did this look good to you to merge?

dhermes:
GLTM

PR #2727, opened 2016-11-11
Removing use of _Monkey in pubsub and logging.

dhermes:
PS I didn't run `tox -e lint` locally, so fingers crossed. Naughty naughty.

### DIFF
--- a/logging/unit_tests/test_client.py
+++ b/logging/unit_tests/test_client.py
@@ -54,8 +54,7 @@ class TestClient(unittest.TestCase):
         self.assertIs(again, api)
 
     def test_logging_api_w_gax(self):
-        from google.cloud.logging import client as MUT
-        from google.cloud._testing import _Monkey
+        import mock
 
         clients = []
         api_obj = object()
@@ -68,7 +67,10 @@ class TestClient(unittest.TestCase):
         client = self._make_one(project=self.PROJECT, credentials=creds,
                                 use_gax=True)
 
-        with _Monkey(MUT, make_gax_logging_api=make_api):
+        patch = mock.patch(
+            'google.cloud.logging.client.make_gax_logging_api',
+            new=make_api)
+        with patch:
             api = client.logging_api
 
         self.assertIs(api, api_obj)
@@ -78,12 +80,14 @@ class TestClient(unittest.TestCase):
         self.assertIs(again, api)
 
     def test_no_gax_ctor(self):
-        from google.cloud._testing import _Monkey
-        from google.cloud.logging import client as MUT
+        import mock
         from google.cloud.logging._http import _LoggingAPI
 
         creds = _Credentials()
-        with _Monkey(MUT, _USE_GAX=True):
+        patch = mock.patch(
+            'google.cloud.logging.client._USE_GAX',
+            new=True)
+        with patch:
             client = self._make_one(project=self.PROJECT, credentials=creds,
                                     use_gax=False)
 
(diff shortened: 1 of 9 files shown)


logging/unit_tests/test_client.py
```diff
@@ -92,11 +96,10 @@ def test_no_gax_ctor(self):
...
     def test_sinks_api_wo_gax(self):
         from google.cloud.logging._http import _SinksAPI
-        from google.cloud.logging import client as MUT
-        from google.cloud._testing import _Monkey
 
-        with _Monkey(MUT, _USE_GAX=False):
-            client = self._make_one(self.PROJECT, credentials=_Credentials())
+        client = self._make_one(
+            self.PROJECT, credentials=_Credentials(),
+            use_gax=False)
```
tseaver: ISTM that this should really be testing the environment sniff, rather than the explicit constructor parameter.

logging/unit_tests/test_client.py
```diff
@@ -133,11 +138,10 @@ def make_api(client_obj):
...
     def test_metrics_api_wo_gax(self):
         from google.cloud.logging._http import _MetricsAPI
-        from google.cloud.logging import client as MUT
-        from google.cloud._testing import _Monkey
 
-        with _Monkey(MUT, _USE_GAX=False):
-            client = self._make_one(self.PROJECT, credentials=_Credentials())
+        client = self._make_one(
+            self.PROJECT, credentials=_Credentials(),
+            use_gax=False)
```
tseaver: Likewise here.

pubsub/unit_tests/test_client.py
```diff
@@ -32,12 +32,12 @@ def _make_one(self, *args, **kw):
...
-        from google.cloud.pubsub import client as MUT
-        from google.cloud._testing import _Monkey
+
         creds = _Credentials()
 
-        with _Monkey(MUT, _USE_GAX=False):
-            client = self._make_one(project=self.PROJECT, credentials=creds)
+        client = self._make_one(
+            project=self.PROJECT, credentials=creds,
+            use_gax=False)
```
tseaver: And here.

pubsub/unit_tests/test_client.py
```diff
@@ -98,12 +101,11 @@ def __init__(self, _wrapped, client):
...
-        from google.cloud.pubsub import client as MUT
-        from google.cloud._testing import _Monkey
-        creds = _Credentials()
 
-        with _Monkey(MUT, _USE_GAX=False):
-            client = self._make_one(project=self.PROJECT, credentials=creds)
+        creds = _Credentials()
+        client = self._make_one(
+            project=self.PROJECT, credentials=creds,
+            use_gax=False)
```
tseaver: And here.

logging/unit_tests/test_client.py
dhermes: [This][1] is a property test, not a constructor test.

The "environment" sniff [happens only in the constructor][3] and [there is already a test for that][2].

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/115748814bf6a59d99c13e070dfdcfbdec5e865d/logging/unit_tests/test_client.py#L93
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/115748814bf6a59d99c13e070dfdcfbdec5e865d/logging/unit_tests/test_client.py#L80
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/115748814bf6a59d99c13e070dfdcfbdec5e865d/logging/google/cloud/logging/client.py#L79-L82

PR #2731, opened 2016-11-14
Update translation

dhermes:


translate/google/cloud/translate/client.py
```diff
@@ -173,7 +185,11 @@ def translate(self, values, target_language=None, format_=None,
                                   for translation. Sets the ``cid`` parameter
                                   in the query.
 
-        :rtype: str or list list
+        :type model: str
+        :param model: (Optional) The model used to translate the text. The
+                      only accepted values are ``base`` and ``nmt``.
```
jonparrott: Should we have constants for these?

translate/google/cloud/translate/connection.py
```diff
@@ -20,11 +20,14 @@
...
+    API_BASE_URL = 'https://translation.googleapis.com'
     """The base of the API call URL."""
 
     API_VERSION = 'v2'
     """The version of the API, used in building the API call's URL."""
 
     API_URL_TEMPLATE = '{api_base_url}/language/translate/{api_version}{path}'
     """A template for the URL of a particular API call."""
+
+    SCOPE = ('https://www.googleapis.com/auth/cloud-platform',)
```
dhermes: I don't think the service has its own scope yet.

jonparrott: Ew. Alright.

translate/google/cloud/translate/client.py
jonparrott: Please do, I definitely prefer that over magic strings. Unless there is some way to get the list of models from the API?

dhermes:
@jonparrott PTAL

dhermes:
We'll sort out the API key issue after the release (joy).

PR #2732, opened 2016-11-14
Remove speech package not yet released warning.

daspecster:
Last checkbox for releasing was to remove the warning about `google-cloud-speech` not being available on PyPI.

Do not merge until **after** release to PyPI.

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -7,12 +7,6 @@ base.
 
 .. warning::
 
-    This API, ``google-cloud-speech`` has not been released yet. In order to
-    test and explore this feature, you must install ``google-cloud`` from
-    source.
-
-.. warning::
-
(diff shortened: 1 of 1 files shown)


tseaver:
Seems like we shouldn't merge this PR until we have actually pushed a release to PyPI.

daspecster:
@tseaver, actually probably just before right?

tseaver:
> actually probably just before right?

I was figuring that he sdist/bdist_wheel distributions don't contain the docs, so there is no particular reason to have them tell "proactive truth":  the only people who will see the docs changes will be looking at `gh-pages` or RTD, and so it would be better if the PyPI release had distributions when the docs change materialized.  I guess the "stable" docs should show the change, which is a point in your favor.

dhermes:
This discussion luckily culminates in #2733!

PR #2735, opened 2016-11-14
Allow using a metric type in the Query method `select_metrics`.

supriyagarg:
This is useful when a user wants to apply the exact same filters (duration, aggregation, resource labels) for a different metric type.

### DIFF
--- a/monitoring/google/cloud/monitoring/query.py
+++ b/monitoring/google/cloud/monitoring/query.py
@@ -297,6 +297,8 @@ class Query(object):
 
             query = query.select_metrics(instance_name='myinstance')
             query = query.select_metrics(instance_name_prefix='mycluster-')
+            query = query.select_metrics(
+                metric_type='compute.googleapis.com/instance/cpu/utilization')
 
         A keyword argument ``<label>=<value>`` ordinarily generates a filter
         expression of the form::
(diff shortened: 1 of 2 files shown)


supriyagarg:
@rimey: Please take a look.

monitoring/google/cloud/monitoring/query.py
```diff
@@ -630,6 +648,10 @@ def select_metrics(self, *args, **kwargs):
 
         See :meth:`Query.select_metrics`.
         """
+        new_metric_type = kwargs.get('metric_type')
+        if new_metric_type is not None:
+          self.metric_type = new_metric_type
+          kwargs.pop('metric_type', None)
```
dhermes: Just `pop` at the outset.

monitoring/google/cloud/monitoring/query.py
```diff
@@ -314,6 +316,20 @@ def select_metrics(self, *args, **kwargs):
 
             metric.label.<label> = ends_with("<value>")
 
+        As a special case, ``"metric_type"`` is treated as a special
+        pseudo-label corresponding to the filter object ``metric.type``.
+        For example, ``metric_type=<value>`` generates::
```
dhermes: Sorry I forgot to mention this, can you change to the explicit syntax:

```
        For example, ``metric_type=<value>`` generates::

        .. code-block:: python

           metric.type = "<value>"
```

I am currently at the very beginning of a project to make sure all of our snippets get tested with doctest and the explicit blocks are part of the project.

lukesneeringer:
Hi @supriyagarg,
It looks like we did you a disservice here, by approving your PR but failing to merge it. Sadly, now that nearly a year as past, the code has moved significantly.

We are still willing to accept this in principle, but can not make changes to your PR (since you have it configured not to allow this). Therefore, I am going to close this issue as stale for now. If you would like to make the necessary updates, though, we would be happy to consider doing this.

PR #2737, opened 2016-11-14
Accounting extra data in language system test.

dhermes:
The API has started returning metadata, e.g.

    {u'mid': u'/m/020bg'}

and has added "painter" as a mention of Michelangelo.


daspecster:
I think this strategy is the only way to go with these services.

PR #2738, opened 2016-11-15
Adding ability to run doctests with datastore system tests.

dhermes:
Since Travis / CircleCI won't run this by default, here is some sample output:

```
$ tox -e system-tests -- datastore
GLOB sdist-make: .../setup.py
system-tests inst-nodeps: .../.tox/dist/google-cloud-0.21.0.zip
...
system-tests runtests: commands[1] | python 
    .../system_tests/attempt_system_tests.py datastore
test_it (datastore.TestDoctest) ... Running Sphinx v1.4.8
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [doctest]: targets for 1 source files that are out of date
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] contents
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
running tests...

Document: contents
------------------
1 items passed all tests:
   9 tests in constructors
9 tests in 1 items.
9 passed and 0 failed.
Test passed.

Doctest summary
===============
    9 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build succeeded.
ok

Ran 1 test in 1.088s

OK
____________________________________ summary ____________________________________
  congratulations :)
```

### DIFF
--- a/datastore/google/cloud/datastore/__init__.py
+++ b/datastore/google/cloud/datastore/__init__.py
@@ -16,14 +16,26 @@
 
 You'll typically use these to get started with the API:
 
-.. code-block:: python
+.. testsetup:: constructors
 
-  from google.cloud import datastore
+  import os
+  os.environ['GOOGLE_CLOUD_PROJECT'] = u'my-project'
 
-  client = datastore.Client()
-  key = client.key('EntityKind', 1234)
-  entity = datastore.Entity(key)
-  query = client.query(kind='EntityKind')
+.. doctest:: constructors
+
+  >>> from google.cloud import datastore
+  >>>
+  >>> client = datastore.Client()
+  >>> print(client.project)
+  my-project
+  >>> key = client.key('EntityKind', 1234)
+  >>> key
+  <Key('EntityKind', 1234), project=my-project>
+  >>> entity = datastore.Entity(key)
+  >>> entity['answer'] = 42
+  >>> entity
+  <Entity('EntityKind', 1234) {'answer': 42}>
+  >>> query = client.query(kind='EntityKind')
 
 The main concepts with this API are:
 
--- a/datastore/google/cloud/datastore/entity.py
+++ b/datastore/google/cloud/datastore/entity.py
@@ -144,7 +144,7 @@ class Entity(dict):
 
     def __repr__(self):
         if self.key:
-            return '<Entity%s %s>' % (self.key.path,
+            return '<Entity%s %s>' % (self.key._flat_path,
                                       super(Entity, self).__repr__())
         else:
-            return '<Entity %s>' % (super(Entity, self).__repr__())
+            return '<Entity %s>' % (super(Entity, self).__repr__(),)
--- a/datastore/google/cloud/datastore/key.py
+++ b/datastore/google/cloud/datastore/key.py
@@ -380,7 +380,7 @@ class Key(object):
         return self._parent
 
     def __repr__(self):
-        return '<Key%s, project=%s>' % (self.path, self.project)
+        return '<Key%s, project=%s>' % (self._flat_path, self.project)
 
 
 def _validate_project(project, parent):
--- a/datastore/unit_tests/test_entity.py
+++ b/datastore/unit_tests/test_entity.py
@@ -190,10 +190,13 @@ class TestEntity(unittest.TestCase):
 
     def test___repr___w_key_non_empty(self):
         key = _Key()
-        key._path = '/bar/baz'
+        flat_path = ('bar', 12, 'baz', 'himom')
+        key._flat_path = flat_path
         entity = self._make_one(key=key)
-        entity['foo'] = 'Foo'
-        self.assertEqual(repr(entity), "<Entity/bar/baz {'foo': 'Foo'}>")
+        entity_vals = {'foo': 'Foo'}
+        entity.update(entity_vals)
+        expected = '<Entity%s %s>' % (flat_path, entity_vals)
+        self.assertEqual(repr(entity), expected)
 
 
 class _Key(object):
@@ -206,7 +209,3 @@ class _Key(object):
 
     def __init__(self, project=_PROJECT):
         self.project = project
-
-    @property
-    def path(self):
-        return self._path
--- a/system_tests/datastore.py
+++ b/system_tests/datastore.py
@@ -14,6 +14,8 @@
 
 import datetime
 import os
+import pkgutil
+import tempfile
 import unittest
 
 import httplib2
@@ -29,6 +31,23 @@ import clear_datastore
 import populate_datastore
 from system_test_utils import EmulatorCreds
 from system_test_utils import unique_resource_id
+
+
+SPHINX_CONF = """\
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx.ext.doctest',
+]
+"""
+
+SPHINX_SECTION_TEMPLATE = """\
+Section %02d
+===========
+
+.. automodule:: google.cloud.%s
+  :members:
+
+"""
 
 
 class Config(object):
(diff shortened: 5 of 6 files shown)


system_tests/datastore.py
```diff
@@ -495,3 +514,59 @@ def test_failure_with_contention(self):
...
+            self._add_section(0, '__init__', file_obj)
+            for index, datastore_module in enumerate(datastore_modules):
+                self._add_section(index + 1, datastore_module, file_obj)
+
+        return docs_dir
+
+    def test_it(self):
+        from sphinx import application
+
+        docs_dir = self._make_temp_docs()
```
dhermes: @tseaver It's worth noting that this will only pick up doctests in source files. I think this is a fine start, and have some ideas how we can have automated checks that make sure we don't miss any tests.

system_tests/datastore.py
```diff
@@ -495,3 +514,59 @@ def test_failure_with_contention(self):
...
+        datastore_modules = self._submodules()
+        with open(index_file, 'w') as file_obj:
+            self._add_section(0, '__init__', file_obj)
+            for index, datastore_module in enumerate(datastore_modules):
+                self._add_section(index + 1, datastore_module, file_obj)
+
+        return docs_dir
+
+    def test_it(self):
+        from sphinx import application
```
dhermes: I considered adding a check here and gracefully bailing if Sphinx wasn't installed. We'll probably (long term) want to add a flag to allow turning off doctest runs.

dhermes:
@tseaver PTAL

dhermes:
@tseaver PTAL

datastore/google/cloud/datastore/key.py
```diff
@@ -380,7 +380,7 @@ def parent(self):
         return self._parent
 
     def __repr__(self):
-        return '<Key%s, project=%s>' % (self.path, self.project)
+        return '<Key%s, project=%s>' % (self._flat_path, self.project)
```
tseaver: Why is this a better repr?

datastore/google/cloud/datastore/__init__.py
```diff
@@ -16,14 +16,26 @@
...
+  >>> print(client.project)
+  my-project
+  >>> key = client.key('EntityKind', 1234)
+  >>> key
+  <Key('EntityKind', 1234), project=my-project>
+  >>> entity = datastore.Entity(key)
+  >>> entity['answer'] = 42
+  >>> entity
+  <Entity('EntityKind', 1234) {'answer': 42}>
+  >>> query = client.query(kind='EntityKind')
```
tseaver: I *really* hate having tests depend on a specific repr.

datastore/google/cloud/datastore/entity.py
```diff
@@ -144,7 +144,7 @@ def exclude_from_indexes(self):
 
     def __repr__(self):
         if self.key:
-            return '<Entity%s %s>' % (self.key.path,
+            return '<Entity%s %s>' % (self.key._flat_path,
                                       super(Entity, self).__repr__())
         else:
-            return '<Entity %s>' % (super(Entity, self).__repr__())
+            return '<Entity %s>' % (super(Entity, self).__repr__(),)
```
tseaver: If we *are* going to use `repr` for testing, then we need to take hash-order out of the equation.

datastore/unit_tests/test_entity.py
```diff
@@ -190,10 +190,13 @@ def test___repr___no_key_empty(self):
...
         key = _Key()
-        key._path = '/bar/baz'
+        flat_path = ('bar', 12, 'baz', 'himom')
+        key._flat_path = flat_path
         entity = self._make_one(key=key)
-        entity['foo'] = 'Foo'
-        self.assertEqual(repr(entity), "<Entity/bar/baz {'foo': 'Foo'}>")
+        entity_vals = {'foo': 'Foo'}
+        entity.update(entity_vals)
+        expected = '<Entity%s %s>' % (flat_path, entity_vals)
```
tseaver: I don't see that the tuple form is any improvement in "comprehensibility" over the path form, at least for "human" consumption.

datastore/google/cloud/datastore/key.py
dhermes: I didn't do it because it was better, I did it to make it deterministic since `repr(self.path)` depends on `PYTHONHASHSEED` (`path` is a dict).

However, it **is** also better, because a key is really just `(kind, id_or_name, kind, id_or_name, ...)` and the fact that we use a dictionary for each `kind / id_or_name` pair is just an implementation detail.

datastore/google/cloud/datastore/__init__.py
dhermes: I am not attached to this test, I just wanted something easy as the first test. Suggestions?

datastore/google/cloud/datastore/entity.py
dhermes: Where does that come in here? (It does matter above.)

dhermes:
@tseaver Only comments on the doctest itself / associated `repr()` changes.

Any comments on the approach of making a fake docs directory and then running doctest there?

tseaver:
> Any comments on the approach of making a fake docs directory and then running doctest there?

We're going to need a way to test all the examples, not just those embedded in docstrings:  do you have a plan for that?

tseaver:
@dhermes Memory eludes me: what stops us from just running `sphinx-build -b doctest` in place in the normal `docs` directory, rather than synthesizing one?

dhermes:
> We're going to need a way to test all the examples, not just those embedded in docstrings: do you have a plan for that?

Yes, though that is "beyond" the scope of this PR. There are two possible approaches that I can see. The first is to "guess" which hand-written RST files apply to a given package (I am not "pro" on this approach). The second is as follows:

-   Discontinue handwritten RST files and use `sphinx-apidoc` to generate our docs (I wrote the docs gen for `oauth2client` using this and @jonparrott "copied" that approach in https://github.com/GoogleCloudPlatform/google-auth-library-python/)
    - In this world, our "package usage" docs will go in `google/cloud/foo/__init__.py`
-   Add checks that literal blocks are as we expect, e.g. no python literal blocks that aren't tested


----

> Memory eludes me: what stops us from just running `sphinx-build -b doctest` in place in the normal `docs` directory, rather than synthesizing one?

The synthesized one is an attempt to isolate on a per-package basis (just as is done with the system tests). It's not strictly necessary.

The "real" reason I did it was because our docs build takes quite a long time (since the project is so large), and I wanted to have a snappier doctest build.

dhermes:
@tseaver Can we keep moving on this?

tseaver:
@dhermes I'd still say that running `sphinx-build -b doctest` (maybe in a separate `[testenv:doctest]` environment w/ appropriate system-tests deps) would be the right way to move forward:  it will pick up any snippets in non-autodoc files, for instance.

datastore/google/cloud/datastore/__init__.py
```diff
@@ -16,14 +16,26 @@
 
 You'll typically use these to get started with the API:
 
-.. code-block:: python
+.. testsetup:: constructors
 
-  from google.cloud import datastore
+  import os
+  os.environ['GOOGLE_CLOUD_PROJECT'] = u'my-project'
```
tseaver: How is that possibly going to work for making API requests?

dhermes: 1. This isn't a global `testsetup`, just for a single `doctest` block
1. I was thinking about this yesterday and just using `...` where the project goes is probably better

dhermes:
> I'd still say that running `sphinx-build -b doctest` (maybe in a separate `[testenv:doctest]` environment w/ appropriate system-tests deps) would be the right way to move forward:  it will pick up any snippets in non-autodoc files, for instance.

This is fine, though I see it having two issues:

- We can't cherry pick which packages we run it for (this is actually really nasty because the docs build is very slow)
- We have to set up and teardown system test-like behavior yet again in CI (as opposed to the scenario where doctests run with the system tests)

dhermes:
Plan moving forward:

- use `sphinx-apidoc` for most RST files
- leave usage docs in handwritten RST files
- make a (potentially "bit-rot"-able) script to pick up usage docs along with other package files when running isolated doctest
- write TOC by hand

tseaver:
LGTM.

PR #2743, opened 2016-11-16
Set core version compatible specifier to packages.

### LINKED ISSUE #2742: All google-cloud-* packages broke due to poor dependency management
All `google-cloud-*` packages broke due to poor dependency management when version 0.21.0 was released.

All `google-cloud-*` packages depend on `google-cloud-core >= <current_version>`. Version 0.21.0 was a breaking change, compared to 0.20.0.

As a result, a dependency on a pinned down version is automatically broken as soon as a new version is released.

The workaround is fairly simple. If one depends on `google-cloud-*`, one *must* explicitly specify the version of `google-cloud-core` to depend on.

The real issue lies at dependency management of the `google-cloud-*` packages. Dependencies should be locked down in the packages that use them. So the requirements should be changed from `google-cloud-core >= <current_version>` to `google-cloud-core ~= <current_version>` (assuming semver is used and patch versions are non-breaking).

1. OS type and version
Ubuntu 16.0.4
2. Python version and virtual environment information `python --version`
```sh
$ python --version
Python 3.5.1+
```
3. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`
```sh
$ pip list
Package                  Version    
------------------------ -----------
...
google-cloud-core        0.21.0     
google-cloud-datastore   0.20.1     
google-cloud-storage     0.20.0
...
```
4. Stacktrace if available
```
      File ".tox/docs/lib/python3.5/site-packages/google/cloud/datastore/__init__.py", line 55, in <module>
        from google.cloud.datastore.connection import Connection
      File ".tox/docs/lib/python3.5/site-packages/google/cloud/datastore/connection.py", line 23, in <module>
        from google.cloud import connection as connection_module
    ImportError: cannot import name 'connection'
```
5. Steps to reproduce
```sh
pip install google-cloud-datastore~=0.20.0
```

daspecster:
Installing a prior version of a package or umbrella pulls in an incompatible version of `google-cloud-core` since it has a version specifier of `>=`.

Ref: https://www.python.org/dev/peps/pep-0440/#compatible-release
Fixes #2742.

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
 ]
 
 setup(
--- a/bigtable/setup.py
+++ b/bigtable/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
     'grpcio >= 1.0.0, < 2.0dev',
 ]
 
--- a/datastore/setup.py
+++ b/datastore/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
     'grpcio >= 1.0.0, < 2.0dev',
 ]
 
--- a/dns/setup.py
+++ b/dns/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
 ]
 
 setup(
--- a/error_reporting/setup.py
+++ b/error_reporting/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
     'google-cloud-logging >= 0.21.0',
 ]
 
--- a/language/setup.py
+++ b/language/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.21.0, < 0.22dev',
 ]
 
 setup(
--- a/logging/setup.py
+++ b/logging/setup.py
@@ -50,7 +50,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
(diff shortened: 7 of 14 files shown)


daspecster:
Should the reference in the root [`setup.py`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/b92f0cf5af596dd41951a8505a7eea037fd1a48a/setup.py#L56) change as well?

Alternatively I think we could use `==` but there would be a lot more maintenance and also could be prone to human error.

dhermes:
@daspecster This is not the right way to go. We should just establish an upper-bound.

daspecster:
@dhermes so something like `>= 0.20.*, < 0.21.0`?

remcohaszing:
@daspecster 

> so something like `>= 0.20.*, < 0.21.0`?

That's exactly what `~= 0.20.0` does. :wink:

dhermes:
> That's exactly what `~= 0.20.0` does. :wink: 

I didn't realize, TIL (or more like YIL).

However, I don't think that's a well known syntax, so I lean towards the one that is more "obvious".

I defer to @jonparrott when there are packaging things I am unsure of.

jonparrott:
I definitely prefer the range specification.

daspecster:
Thanks @jonparrott!

Updates pushed.

Anything else we should consider for this?

bigquery/setup.py
```diff
@@ -50,7 +50,7 @@
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.20.*, < 0.21.0',
```
dhermes: Please follow the pattern we've already established:

```
>= 0.20.0, < 0.21dev'
```

@jonparrott confirm?

jonparrott: right, please do that. Sometimes we do need a minimum patch version.

daspecster: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2743/commits/bb6a3ae04232a878e2c51cac870db3418aaa9af2

bigquery/setup.py
```diff
@@ -50,7 +50,7 @@
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.21.0',
+    'google-cloud-core >= 0.20.0, < 0.21dev',
```
daspecster: Thanks! I squashed btw.

dhermes:
@daspecster Would you like to tackle https://github.com/GoogleCloudPlatform/google-cloud-python-happybase as well? (It hasn't been upgraded to depend on the new `core`, it may be in a b0rken state?)

daspecster:
@dhermes sure I'll check it out!

daspecster:
@dhermes when I run the tests for google-cloud-python-happybase, it's pulling in the latest `core` (0.21.0) and the unit tests and system tests are passing.

```bash
tox -e system-tests
...
futures==3.0.5,
google-cloud-bigtable==0.21.0,
google-cloud-core==0.21.0,
google-cloud-happybase==0.20.0,
googleapis-common-protos==1.5.0,
grpcio==1.0.1,
...
```

dhermes:
@daspecster That's good to know. We should still set the upper bound on the dependency though.

daspecster:
@dhermes  https://github.com/GoogleCloudPlatform/google-cloud-python-happybase/pull/18

PR #2755, opened 2016-11-18
datastore: Support decoding empty arrays

udoprog:
When the `array_value` is empty, the set also becomes empty which triggers the existing condition `len(exclude_values) != 0`.

This patch changes this condition to permit empty array values to pass through.

### DIFF
--- a/datastore/google/cloud/datastore/helpers.py
+++ b/datastore/google/cloud/datastore/helpers.py
@@ -138,12 +138,12 @@ def entity_from_protobuf(pb):
         if is_list:
             exclude_values = set(value_pb.exclude_from_indexes
                                  for value_pb in value_pb.array_value.values)
-            if len(exclude_values) != 1:
+            if len(exclude_values) > 1:
                 raise ValueError('For an array_value, subvalues must either '
                                  'all be indexed or all excluded from '
                                  'indexes.')
 
-            if exclude_values.pop():
+            if len(exclude_values) > 0 and exclude_values.pop():
                 exclude_from_indexes.append(prop_name)
         else:
             if value_pb.exclude_from_indexes:


udoprog:
I signed it!

dhermes:
@udoprog Can you give an example that currently fails, but would work under the changes in this PR?

udoprog:
The python client doesn't permit writing an empty array, so you'll have to do that using another client (we do it using Java).

Or, go into the cloud console and add an entity with an `Array` property that has the following value:

```
{
  "values": []
}
```

After this, run the following code:
```python
from google.cloud import datastore

if __name__ == "__main__":
    client = datastore.Client('my-project-1')

    k = client.key('test', 'test')
    e = client.get(k)

    print(e)
```

Without this patch, you should see:

```
  File "example.py", line 9, in <module>
    e = client.get(k)
  File "/home/udoprog/.local/lib/python3.5/site-packages/google/cloud/datastore/client.py", line 256, in get
    deferred=deferred, transaction=transaction)
  File "/home/udoprog/.local/lib/python3.5/site-packages/google/cloud/datastore/client.py", line 316, in get_multi
    for entity_pb in entity_pbs]
  File "/home/udoprog/.local/lib/python3.5/site-packages/google/cloud/datastore/client.py", line 316, in <listcomp>
    for entity_pb in entity_pbs]
  File "/home/udoprog/.local/lib/python3.5/site-packages/google/cloud/datastore/helpers.py", line 142, in entity_from_protobuf
    raise ValueError('For an array_value, subvalues must either '
ValueError: For an array_value, subvalues must either all be indexed or all excluded from indexes.
```

dhermes:
Thanks @udoprog!

In order to actually merge this in, we'll need some unit tests. You can write them yourself, or I am happy to write them, since you've allowed maintainers to push commits to your PR branch.

daspecster:
@dhermes won't google CLA bot wig out if multiple people commit?

dhermes:
@daspecster It may, but we the humans can ignore it.

udoprog:
My apologies I can't take a more active part in this right now. Feel free to just see this as a bug report and fix things whatever way suits you best.

datastore/google/cloud/datastore/helpers.py
```diff
@@ -138,12 +138,12 @@ def entity_from_protobuf(pb):
         if is_list:
             exclude_values = set(value_pb.exclude_from_indexes
                                  for value_pb in value_pb.array_value.values)
-            if len(exclude_values) != 1:
+            if len(exclude_values) > 1:
```
daspecster: @dhermes if `len(exclude_values) == 0`, won't that cause trouble when it *should* have raised? I'm thinking this line should be reverted back?

dhermes: Come up with sample data?

daspecster: I keep playing with this from time to time but I'm having a hard time testing it.
You can't load an empty value into the generated proto objects so I need to figure out how to mock the response from `Connection` in order to replicate the situation. I haven't dug into the protos to figure out what the response should look like yet.

lukesneeringer:
See also https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3152, where I attempted to fix this problem and hit another problem.

We likely do not need two issues for it, though, so I am closing this one.

PR #2758, opened 2016-11-19
Fixed grammar in comments and docs.

mbrukman:
Summary of edits:
* "a instance" -> "an instance" consistency (some are correct, some not)
* removed extraneous words, added missing words in sentences
* minor grammatical edits for readability


dhermes:
Thanks @mbrukman. This likely happened via a find-and-replace when instance came into existence as a concept (used to be a cluster in those modules?)

mbrukman:
@dhermes – yes, you're correct, clusters used to be stand-alone, and _instance_ was introduced in the summer as a new object, so that's probably what happened.

PR #2764, opened 2016-11-23
Manually adding docs for 0.21.0.

### LINKED ISSUE #2763: 0.21.0 documentation is missing
0.21.0 was released recently (~1 week ago).  It introduces several backward-incompatible changes, which are only mentioned in the release notes.

However the documentation has not been updated, and still references 0.20.0 and the old API:

https://googlecloudplatform.github.io/google-cloud-python/stable/

It is also unclear what version of the library the documentation is for by reading the docs, because the Sphinx flags for version numbering have not been used.

dhermes:
Fixes #2763.


daspecster:
Could we just restart the build for the release instead?

dhermes:
@daspecster What's the point, the changes are already ready to go?

daspecster:
That way it would ensure that it's possible to fully build the 0.21.0 tag and release the docs at that point. It just seems like the proper path to validate and tie docs to a tag.
If you're not concerned with that then these LGTM.

dhermes:
> If you're not concerned with that

I'm not :grinning:

PR #2773, opened 2016-11-28
Fix double encoding of messages when commit() fails.

### LINKED ISSUE #2771: PubSub JSON Double base64 Encode
We recently upgraded from 0.18.1 to 0.21.0 and we do *not* use gRPC/GAX/Protobuf fun times, but we use the JSONPublisherApi.

Maybe we are using the topic.Batch().Commit() method wrong, but we have been using it in our retry strategy to just simply call Commit() again.

With the change in #2577 it seems that Commit() will double encode *if* you just call Commit() again on a network failure.

Is this the intended design? Are we supposed to keep another copy of the message batch or create another message batch?

daspecster:
Fixes #2771 

This copies the `messages` variable instead of transforming it in place so that if a `commit()` fails, the messages aren't double encoded.

### DIFF
--- a/pubsub/google/cloud/pubsub/_http.py
+++ b/pubsub/google/cloud/pubsub/_http.py
@@ -15,6 +15,7 @@
 """Create / interact with Google Cloud Pub/Sub connections."""
 
 import base64
+import copy
 import functools
 import os
 
@@ -203,9 +204,10 @@ class _PublisherAPI(object):
         :rtype: list of string
         :returns: list of opaque IDs for published messages.
         """
-        _transform_messages_base64(messages, _base64_unicode)
+        messages_to_send = copy.deepcopy(messages)
+        _transform_messages_base64(messages_to_send, _base64_unicode)
         conn = self._connection
-        data = {'messages': messages}
+        data = {'messages': messages_to_send}
         response = conn.api_request(
             method='POST', path='/%s:publish' % (topic_path,), data=data)
         return response['messageIds']
--- a/pubsub/unit_tests/test__http.py
+++ b/pubsub/unit_tests/test__http.py
@@ -284,10 +284,32 @@ class Test_PublisherAPI(_Base):
         msg_data = connection._called_with['data']['messages'][0]['data']
         self.assertEqual(msg_data, B64_PAYLOAD)
 
-    def test_topic_publish_miss(self):
-        from google.cloud.exceptions import NotFound
+    def test_topic_publish_twice(self):
+        import base64
+
         PAYLOAD = b'This is the message text'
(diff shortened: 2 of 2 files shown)


pubsub/google/cloud/pubsub/_http.py
```diff
@@ -203,9 +204,10 @@ def topic_publish(self, topic_path, messages):
         :rtype: list of string
         :returns: list of opaque IDs for published messages.
         """
-        _transform_messages_base64(messages, _base64_unicode)
+        messages_to_send = copy.deepcopy(messages)
```
dhermes: D'oh! Where is the caller of this?

daspecster: It's [`Topic.commit()`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/8001d45c2f6cb3fcf583280874b141c83541c50a/pubsub/google/cloud/pubsub/topic.py#L444)

dhermes: Got it, so the caller sends a shallow copy: `self.messages[:]`.

This fix is just fine, though I wonder if we should just create a new dictionary on the fly in `_transform_messages_base64`. It's already doing a loop over all the data, so the copy effectively makes us do that loop twice.

daspecster: How big and how many messages can there be in one commit?

daspecster: I can't find anything about size in https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage.

If the messages could be considerable in size then we might want to really think about this more.

dhermes: We don't set a hard cap, but we could. I seem to recall the backend having a limit, maybe it's size-based (e.g. 1MB) or number (e.g. 10000 messages) or time (e.g. 60s request).

daspecster: So I was just changing `_transform_messages_base64` and I realized it's used in `subscription_pull()` as well. 
We might want to edit the messages in place instead of copying for *pull* because it should increase the performance?

So I propose the current changes to be used.

dhermes: I don't fully parse what you're trying to say, but I agree this performance discussion shouldn't block the PR.

Merge as-is and file a follow-up bug?

dhermes: Feel free to merge

HakShak: As a note here, a single request is limited to 1000 messages and/or 10MB in size.

PR #2776, opened 2016-11-29
Add support for query parameters

### LINKED ISSUE #2551: BigQuery: Add support for query parameters
The CLI command `bq query` now supports query parameters for Standard SQL queries and the API does as well (https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#request-body). It would be great to be able to provide these via google.cloud.

tseaver:
Closes #2551.

### DIFF
--- a/bigquery/google/cloud/bigquery/__init__.py
+++ b/bigquery/google/cloud/bigquery/__init__.py
@@ -28,3 +28,6 @@ from google.cloud.bigquery.dataset import AccessGrant
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import Table
+from google.cloud.bigquery._helpers import ArrayQueryParameter
+from google.cloud.bigquery._helpers import ScalarQueryParameter
+from google.cloud.bigquery._helpers import StructQueryParameter
--- a/bigquery/google/cloud/bigquery/__init__.py
+++ b/bigquery/google/cloud/bigquery/__init__.py
@@ -23,11 +23,11 @@ The main concepts with this API are:
 """
 
 
+from google.cloud.bigquery._helpers import ArrayQueryParameter
+from google.cloud.bigquery._helpers import ScalarQueryParameter
+from google.cloud.bigquery._helpers import StructQueryParameter
 from google.cloud.bigquery.client import Client
 from google.cloud.bigquery.dataset import AccessGrant
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import Table
-from google.cloud.bigquery._helpers import ArrayQueryParameter
-from google.cloud.bigquery._helpers import ScalarQueryParameter
-from google.cloud.bigquery._helpers import StructQueryParameter
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -409,3 +409,76 @@ class ArrayQueryParameter(AbstractQueryParameter):
         if self.name is not None:
             resource['name'] = self.name
         return resource
+
+
+class StructQueryParameter(AbstractQueryParameter):
+    """Named / positional query parameters for struct values.
+
+    :type name: str or None
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
+                 paramter can only be addressed via position (`?`).
+
+    :type sub_parms: tuple of :class:`ScalarQueryParameter`
+    :param sub_parms: the sub-parameters for the struct
+    """
+    def __init__(self, name, *sub_parms):
+        self.name = name
+        self._order = [sub.name for sub in sub_parms]
+        self.struct_types = {sub.name: sub.type_ for sub in sub_parms}
+        self.struct_values = {sub.name: sub.value for sub in sub_parms}
+
+    @classmethod
+    def positional(cls, *sub_parms):
+        """Factory for positional paramters.
+
+        :type sub_parms: tuple of :class:`ScalarQueryParameter`
+        :param sub_parms:s the sub-parameters for the struct
+
+        :rtype: :class:`StructQueryParameter`
+        :returns: instance w/o name
+        """
+        return cls(None, *sub_parms)
+
+    @classmethod
+    def from_api_repr(cls, resource):
+        """Factory: construct paramter from JSON resource.
+
+        :type resource: dict
+        :param resource: JSON mapping of parameter
+
+        :rtype: :class:`StructQueryParameter`
+        :returns: instance
+        """
+        name = resource.get('name')
+        instance = cls(name)
+        types = instance.struct_types = {
+            item['name']: item['type']
+            for item in resource['parameterType']['structTypes']
+        }
+        struct_values = resource['parameterValue']['structValues']
+        values = instance.struct_values = {}
+        for key, value in struct_values.items():
+            values[key] = _CELLDATA_FROM_JSON[types[key]](value, None)
+        return instance
+
+    def to_api_repr(self):
+        """Construct JSON API representation for the parameter.
+
+        :rtype: dict
+        :returns: JSON mapping
+        """
+        types = [
+            {'name': name, 'type': self.struct_types[name]}
+            for name in self._order
+        ]
+        resource = {
+            'parameterType': {
+                'structTypes': types,
+            },
+            'parameterValue': {
+                'structValues': self.struct_values,
+            },
+        }
+        if self.name is not None:
+            resource['name'] = self.name
+        return resource
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -243,7 +243,29 @@ def _build_udf_resources(resources):
     return udfs
 
 
-class ScalarQueryParameter(object):
+class AbstractQueryParameter(object):
+    """Base class for named / positional query parameters.
+    """
+    @classmethod
+    def from_api_repr(cls, resource):
+        """Factory: construct paramter from JSON resource.
+
+        :type resource: dict
+        :param resource: JSON mapping of parameter
+
+        :rtype: :class:`ScalarQueryParameter`
+        """
+        raise NotImplementedError
+
+    def to_api_repr(self):
+        """Construct JSON API representation for the parameter.
+
+        :rtype: dict
+        """
+        raise NotImplementedError
+
+
+class ScalarQueryParameter(AbstractQueryParameter):
     """Named / positional query parameters for scalar values.
 
     :type name: str or None
@@ -275,6 +297,9 @@ class ScalarQueryParameter(object):
         :type value: str, int, float, bool, :class:`datetime.datetime`, or
                      :class:`datetime.date`.
         :param value: the scalar parameter value.
+
+        :rtype: :class:`ScalarQueryParameter`
+        :returns: instance w/o name
         """
         return cls(None, type_, value)
 
@@ -284,6 +309,9 @@ class ScalarQueryParameter(object):
 
         :type resource: dict
         :param resource: JSON mapping of parameter
+
+        :rtype: :class:`ScalarQueryParameter`
+        :returns: instance
         """
         name = resource.get('name')
         type_ = resource['parameterType']['type']
@@ -292,7 +320,11 @@ class ScalarQueryParameter(object):
         return cls(name, type_, converted)
 
     def to_api_repr(self):
-        """Construct JSON API representation for the parameter."""
+        """Construct JSON API representation for the parameter.
+
+        :rtype: dict
+        :returns: JSON mapping
+        """
         resource = {
             'parameterType': {
                 'type': self.type_,
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -500,7 +500,7 @@ class QueryParametersProperty(object):
                          the class).
 
         :type value: list of instances of classes derived from
-                :class:`AbstractQueryParameter`.
+                     :class:`AbstractQueryParameter`.
         :param value: new query parameters for the instance.
         """
         if not all(isinstance(u, AbstractQueryParameter) for u in value):
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -14,6 +14,8 @@
 
 """Shared helper functions for BigQuery API classes."""
 
+from collections import OrderedDict
+
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _date_from_iso8601_date
 
@@ -407,8 +409,8 @@ class StructQueryParameter(AbstractQueryParameter):
     """
     def __init__(self, name, *sub_parms):
         self.name = name
-        self._order = [sub.name for sub in sub_parms]
-        self.struct_types = {sub.name: sub.type_ for sub in sub_parms}
+        self.struct_types = OrderedDict(
+            (sub.name, sub.type_) for sub in sub_parms)
         self.struct_values = {sub.name: sub.value for sub in sub_parms}
 
     @classmethod
@@ -452,8 +454,8 @@ class StructQueryParameter(AbstractQueryParameter):
         :returns: JSON mapping
         """
         types = [
-            {'name': name, 'type': self.struct_types[name]}
-            for name in self._order
+            {'name': key, 'type': value}
+            for key, value in self.struct_types.items()
         ]
         resource = {
             'parameterType': {
@@ -474,6 +476,13 @@ class QueryParametersProperty(object):
     def __get__(self, instance, owner):
         """Descriptor protocol:  accessor
 
+        :type instance: :class:`QueryParametersProperty`
+        :param instance: instance owning the property (None if accessed via
+                         the class).
+
+        :type owner: type
+        :param owner: the class owning the property.
+
         :rtype: list of instances of classes derived from
                 :class:`AbstractQueryParameter`.
         :returns: the descriptor, if accessed via the class, or the instance's
@@ -485,6 +494,10 @@ class QueryParametersProperty(object):
 
     def __set__(self, instance, value):
         """Descriptor protocol:  mutator
+
+        :type instance: :class:`QueryParametersProperty`
+        :param instance: instance owning the property (None if accessed via
+                         the class).
 
         :type value: list of instances of classes derived from
                 :class:`AbstractQueryParameter`.
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -472,13 +472,24 @@ class QueryParametersProperty(object):
     """Custom property type, holding query parameter instances."""
 
     def __get__(self, instance, owner):
-        """Descriptor protocol:  accessor"""
+        """Descriptor protocol:  accessor
+
+        :rtype: list of instances of classes derived from
+                :class:`AbstractQueryParameter`.
+        :returns: the descriptor, if accessed via the class, or the instance's
+                  query paramters.
+        """
         if instance is None:
             return self
         return list(instance._query_parameters)
 
     def __set__(self, instance, value):
-        """Descriptor protocol:  mutator"""
+        """Descriptor protocol:  mutator
+
+        :type value: list of instances of classes derived from
+                :class:`AbstractQueryParameter`.
+        :param value: new query parameters for the instance.
+        """
         if not all(isinstance(u, AbstractQueryParameter) for u in value):
             raise ValueError(
                 "query parameters must be derived from AbstractQueryParameter")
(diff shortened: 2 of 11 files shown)


bigquery/google/cloud/bigquery/__init__.py
```diff
@@ -28,3 +28,6 @@
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import Table
+from google.cloud.bigquery._helpers import ArrayQueryParameter
```
daspecster: Shouldn't these be at the top. I would think that `bigquey._helpers` would be before `bigquery.client`.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -228,16 +228,259 @@ def __set__(self, instance, value):
...
+        :rtype: dict
+        """
+        raise NotImplementedError
+
+
+class ScalarQueryParameter(AbstractQueryParameter):
+    """Named / positional query parameters for scalar values.
+
+    :type name: str or None
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
```
daspecster: Double backticks are probably what you want?

Throughout the changed/added files.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -228,16 +228,259 @@ def __set__(self, instance, value):
...
+    """
+    def __init__(self, name, type_, value):
+        self.name = name
+        self.type_ = type_
+        self.value = value
+
+    @classmethod
+    def positional(cls, type_, value):
+        """Factory for positional paramters.
+
```
daspecster: two newlines, probably just want one.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -228,16 +228,259 @@ def __set__(self, instance, value):
...
+        :type type_: str
+        :param type_: name of paramter type.  One of `'STRING'`, `'INT64'`,
+                      `'FLOAT64'`, `'BOOLEAN'`, `'TIMESTAMP'`, or `'DATE'`.
+
+        :type value: str, int, float, bool, :class:`datetime.datetime`, or
+                     :class:`datetime.date`.
+        :param value: the scalar parameter value.
+
+        :rtype: :class:`ScalarQueryParameter`
+        :returns: instance w/o name
```
daspecster: `:returns:` ends up publically visible, so we might want it to be...
```
Instance without name.
```

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -228,16 +228,259 @@ def __set__(self, instance, value):
...
+        self._order = [sub.name for sub in sub_parms]
+        self.struct_types = {sub.name: sub.type_ for sub in sub_parms}
+        self.struct_values = {sub.name: sub.value for sub in sub_parms}
+
+    @classmethod
+    def positional(cls, *sub_parms):
+        """Factory for positional paramters.
+
+        :type sub_parms: tuple of :class:`ScalarQueryParameter`
+        :param sub_parms:s the sub-parameters for the struct
```
daspecster: `:param sub_parms:s`

bigquery/unit_tests/test__helpers.py
```diff
@@ -491,6 +491,363 @@ def test_instance_setter_w_bad_udfs(self):
         self.assertEqual(instance.udf_resources, [])
 
 
+class Test_AbstractQueryParameter(unittest.TestCase):
+
+    @staticmethod
+    def _get_target_class():
+        from google.cloud.bigquery._helpers import AbstractQueryParameter
```
daspecster: Throughout the tests, can we add a space between the imports and the following code?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -241,3 +241,66 @@ def _build_udf_resources(resources):
         udf = {resource.udf_type: resource.value}
         udfs.append(udf)
     return udfs
+
+
+class ScalarQueryParameter(object):
+    """Named / positional query parameters for scalar values.
+
+    :type name: str or None
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
```
dhermes: Sphinx wants two backticks for code font

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -241,3 +241,66 @@ def _build_udf_resources(resources):
         udf = {resource.udf_type: resource.value}
         udfs.append(udf)
     return udfs
+
+
+class ScalarQueryParameter(object):
+    """Named / positional query parameters for scalar values.
+
+    :type name: str or None
```
dhermes: In general, the `or None` isn't needed, instead we can say "(Optional) " in the prose (also `None` isn't a type)

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -241,3 +241,66 @@ def _build_udf_resources(resources):
...
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
+                 paramter can only be addressed via position (`?`).
```
dhermes: "paramter" --> "parameter"

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -241,3 +241,66 @@ def _build_udf_resources(resources):
...
+    @classmethod
+    def from_api_repr(cls, resource):
+        """Factory: construct paramter from JSON resource.
+
+        :type resource: dict
+        :param resource: JSON mapping of parameter
+        """
+        name = resource.get('name')
+        type_ = resource['parameterType']['type']
+        value = resource['parameterValue']['value']
```
dhermes: Can we just pre-emptively prepare for `KeyError`s? Have been bitten too many times

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -241,3 +241,66 @@ def _build_udf_resources(resources):
...
+
+    :type type_: str
+    :param type_: name of parameter type.  One of `'STRING'`, `'INT64'`,
+                  `'FLOAT64'`, `'BOOLEAN'`, `'TIMESTAMP'`, or `'DATE'`.
+
+    :type value: str, int, float, bool, :class:`datetime.datetime`, or
+                 :class:`datetime.date`.
+    :param value: the scalar parameter value.
+    """
+    def __init__(self, name, type_, value):
```
dhermes: Why not put `name` at the end and make it optional?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -243,7 +243,29 @@ def _build_udf_resources(resources):
     return udfs
 
 
-class ScalarQueryParameter(object):
+class AbstractQueryParameter(object):
+    """Base class for named / positional query parameters.
+    """
```
dhermes: If you ain't gonna use it, put the `"""` on the same line

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -336,3 +336,76 @@ def to_api_repr(self):
...
+
+class ArrayQueryParameter(AbstractQueryParameter):
+    """Named / positional query parameters for array values.
+
+    :type name: str or None
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
+                 paramter can only be addressed via position (`?`).
+
+    :type array_type: str
+    :param array_type:
```
dhermes: This is the `T` in `list<T>`?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -336,3 +336,76 @@ def to_api_repr(self):
...
+        :param resource: JSON mapping of parameter
+
+        :rtype: :class:`ArrayQueryParameter`
+        :returns: instance
+        """
+        name = resource.get('name')
+        array_type = resource['parameterType']['arrayType']
+        values = resource['parameterValue']['arrayValues']
+        converted = [
+            _CELLDATA_FROM_JSON[array_type](value, None) for value in values]
```
dhermes: Can a list contain a list? Does `_CELLDATA_FROM_JSON` need to know how to handle this recursively?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -409,3 +409,76 @@ def to_api_repr(self):
...
+
+class StructQueryParameter(AbstractQueryParameter):
+    """Named / positional query parameters for struct values.
+
+    :type name: str or None
+    :param name: Parameter name, used via `@foo` syntax.  If None, the
+                 paramter can only be addressed via position (`?`).
+
+    :type sub_parms: tuple of :class:`ScalarQueryParameter`
+    :param sub_parms: the sub-parameters for the struct
```
dhermes: Chicken parm? `parms` --> `params`

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -409,3 +409,76 @@ def to_api_repr(self):
...
+    """
+    def __init__(self, name, *sub_parms):
+        self.name = name
+        self._order = [sub.name for sub in sub_parms]
```
dhermes: `collections.OrderedDict` landed in Python 2.7

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -409,3 +409,76 @@ def to_api_repr(self):
...
+
+        :rtype: :class:`StructQueryParameter`
+        :returns: instance
+        """
+        name = resource.get('name')
+        instance = cls(name)
+        types = instance.struct_types = {
+            item['name']: item['type']
+            for item in resource['parameterType']['structTypes']
+        }
```
dhermes: Doesn't this seem to indicate that the constructor is deficient / wrong?

bigquery/unit_tests/test__helpers.py
tseaver: This isn't our convention, in general.  E.g., try the following:

```bash
$ find . -name "test_*.py" | grep -v "tox" | xargs grep -C1 "from .* import" | less
```

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -467,3 +467,20 @@ def to_api_repr(self):
...
+
+    def __get__(self, instance, owner):
+        """Descriptor protocol:  accessor"""
+        if instance is None:
+            return self
+        return list(instance._query_parameters)
+
+    def __set__(self, instance, value):
+        """Descriptor protocol:  mutator"""
+        if not all(isinstance(u, AbstractQueryParameter) for u in value):
```
dhermes: Can you replace `u` with something else? I don't even know what types `u` or `value` are supposed to be (other than `value` is iterable).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1035,6 +1047,15 @@ def _populate_config_resource(self, configuration):
                 {udf_resource.udf_type: udf_resource.value}
                 for udf_resource in self._udf_resources
             ]
+        if len(self._query_parameters) > 0:
+            configuration[self._QUERY_PARAMETERS_KEY] = [
+                query_parameter.to_api_repr()
+                for query_parameter in self._query_parameters
+            ]
+            if self._query_parameters[0].name is None:
```
dhermes: The assumption is that every query param in `self._query_parameters` have the same mode. What enforces that assumption?

bigquery/unit_tests/test_job.py
```diff
@@ -1287,12 +1287,31 @@ def _verifyIntegerResourceProperties(self, job, config):
         else:
             self.assertIsNone(job.maximum_bytes_billed)
 
+    def _verifyUDFResources(self, job, config):
```
dhermes: Y u no PEP8?

bigquery/unit_tests/test_job.py
```diff
@@ -1561,6 +1604,107 @@ def test_begin_w_bound_client_and_udf(self):
         self.assertEqual(req['data'], SENT)
         self._verifyResourceProperties(job, RESOURCE)
 
+    def test_begin_w_named_query_parameter(self):
+        from google.cloud.bigquery._helpers import ScalarQueryParameter
+        query_parameters = [ScalarQueryParameter('foo', 'INT64', 123)]
+        PATH = 'projects/%s/jobs' % self.PROJECT
```
dhermes: explicit tuple to `%`

docs/bigquery_snippets.py
```diff
@@ -461,13 +461,9 @@ def do_something_with(_):
...
 
     # [START client_list_jobs]
-    jobs, token = client.list_jobs()   # API request
-    while True:
-        for job in jobs:
-            do_something_with(job)
-        if token is None:
-            break
-        jobs, token = client.list_jobs(page_token=token)  # API request
+    job_iterator = client.list_jobs()   # API request
```
dhermes: 1. My bad
1. The API request doesn't happen until the iterator is consumed and there might be multiple requests.

docs/bigquery_snippets.py
```diff
@@ -490,6 +486,30 @@ def client_run_sync_query(client, _):
 
 
 @snippet
+def client_run_sync_query_w_param(client, _):
+    """Run a synchronous query using a query parameter"""
+    QUERY_W_PARM = (
```
dhermes: Chicken parm again?

bigquery/unit_tests/test__helpers.py
daspecster: Yeah, I think @dhermes has [asked](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2663#discussion_r86286081) me to in the [past](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2770#discussion_r90081046). I guess we just need to *commit* to one or the other. (*pun somewhat intended*)

dhermes: Indeed. Just because we've done it before doesn't mean we meant to.

tseaver: I'd argue against it, myself, and I certainly don't think it makes sense to hold up unrelated PRs to enforce a new and debatable style rule.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: By inspection, `value` is a sequence, and its items are expected to be instances of classes derived from `AbstractQueryParameter`.  Using a longer name for the placeholder iteration value doesn't add information, and makes the expression wrap, which then puts us into lint hell.

I will update the docstrings to make the sense more obvious.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: I'd rather have the named form be "normal", and have the `positional` factory mark the other usage.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: The API's representation of query parameters is twisted:  it splits the parameter type apart from the value, and makes each one a "union type".  I decided that it made more sense to have the user construct the individual sub-parameters (which know their own type and value) and synthesize the `StructQueryParameter` from them.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: The [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays) say "arrays of arrays are not supported."

bigquery/google/cloud/bigquery/job.py
tseaver: Nothing.  If the user screws it up, the back-end will barf.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -228,16 +230,280 @@ def __set__(self, instance, value):
...
+class ScalarQueryParameter(AbstractQueryParameter):
+    """Named / positional query parameters for scalar values.
+
+    :type name: str or None
+    :param name: Parameter name, used via ``@foo`` syntax.  If None, the
+                 paramter can only be addressed via position (``?``).
+
+    :type type_: str
+    :param type_: name of parameter type.  One of `'STRING'`, `'INT64'`,
+                  `'FLOAT64'`, `'BOOLEAN'`, `'TIMESTAMP'`, or `'DATE'`.
```
tseaver: Hmm, the [API docs for schema definition](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#schema.fields.type) use `BOOLEAN`:  is `BOOL` the "standard SQL" alias (like `INT64` for `INTEGER`)?

bigquery/unit_tests/test__helpers.py
dhermes: 1. That's fine, we don't need to hold this PR up
1. It would be very easy for you to just add the one extra character
1. There will never be "related" PRs

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -452,8 +454,8 @@ def to_api_repr(self):
         :returns: JSON mapping
         """
         types = [
-            {'name': name, 'type': self.struct_types[name]}
-            for name in self._order
+            {'name': key, 'type': value}
+            for key, value in self.struct_types.items()
```
dhermes: I encourage the use of `six.iteritems()`, though not sure what the expected size of items is to know if a copy is prohibitive

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -472,13 +474,35 @@ class QueryParametersProperty(object):
...
     def __set__(self, instance, value):
-        """Descriptor protocol:  mutator"""
+        """Descriptor protocol:  mutator
+
+        :type instance: :class:`QueryParametersProperty`
+        :param instance: instance owning the property (None if accessed via
+                         the class).
+
+        :type value: list of instances of classes derived from
+                :class:`AbstractQueryParameter`.
```
dhermes: This indent doesn't make sense? (Change it if you like, leaving it won't make Sphinx mad, and I am "over it" on "style" that can't be enforced with a tool.)

tseaver: I pushed the indentation change and cancelled the Travis / Circle builds.  Any remaining issues before squashing merging?

bigquery/unit_tests/test__helpers.py
tseaver: If we want to adopt the rule that imports-inside-testcases must be followed by a blank line, we should have a monster PR which fixes all the places where we *don't* have one, and then start enforcing it.

bigquery/google/cloud/bigquery/_helpers.py
dhermes: @tseaver This now contradicts the constructor, where `struct_types` is an OrderedDict

dhermes: Suggest: feed the actual values to the constructor instead of monkey patching after construction

bigquery/google/cloud/bigquery/_helpers.py
tseaver: Yes,

bigquery/google/cloud/bigquery/_helpers.py
tseaver: I can't imagine ever seeing more than a dozen struct elements in a parameter.

bigquery/google/cloud/bigquery/_helpers.py
dhermes: DERP. My bad

tseaver:
> especially if no human is every going to use the constructor

I do expect humans to use the constructor, e.g.:

```python
from google.cloud.bigquery import Client
from google.cloud.bigquery import ScalarQueryParameter
from google.cloud.bigquery import StructQueryParameter


birth_state = ScalarQueryParameter('birth_state', 'STRING', 'VA')
death_state = ScalarQueryParameter('death_state', 'STRING', 'MS')
birth_death = StructQueryParameter(
    'birth_death', birth_state, death_state)

client = Client()

query = client.run_sync_query(
    'SELECT * FROM mortality WHERE birth_death = @birth_death')
```

avitzavi:
does this mean we can use @parameters in the BigQuery UI ?? I cant seem to figure out how to declare them correctly.

tseaver:
@avitzavi I'm not sure what you are asking.  Does the [example in the docs](https://googleapis.github.io/google-cloud-python/latest/bigquery/usage/queries.html#run-a-query-using-a-named-query-parameter) help?

avitzavi:
I’m trying to use variables like I do in sql server:
“Declare @variable INT = 10”

Can I do this in the BigQuery UI?

tseaver:
The `@parameter` placeholders added by this PR are used for passing in values from the client to be substituted into a query:  they aren't meant to be used that way.

Also, I don't know what you mean by "BigQuery UI" -- is that the https://console.cloud.google.com/ interface?

avitzavi:
Yes it is the console. 

The Declare/Set methods from SQL are very valuable for me as a data analyst. 
Seems like a really big missing feature...

I create many reports (based off scheduled queries) and not having these parameters makes it much more challenging to do. 

Especially given the Console's lack of basic editor features like find/replace.

tseaver:
@avitzavi This repository is solely about the Python client libraries which wrap the various Google Cloud APIs.  Please check out the [BigQuery support page](https://cloud.google.com/bigquery/support) for ways to get support for the back-end API itself.

Issue #2753, opened 2016-11-18
Tracking Vision Development.


daspecster:
There are some things missing in vision that should be added to complete the service.

- [x] Add support for `image = client.image(filename='./my.jpg')`. (#2778, #2792)
- [x] Make public method for multiple annotation detection(#2770).
- [x] Move emotion attributes directly to face. `face.joy` instead of `face.emotions.joy`. (#2782)
- [x] Add system tests #2526. (#2788, #2827, #2830, #2834, #2835, #2836, #2839)
- [x] Add GAPIC support. (#2914, #2918, #2922, #2937, #2938)
- [x] Add batch processing, annotation multiple images with one request.(#2897, #2967, #2978)
- [ ] Rename Face.Landmark to something like Face.FacialFeature(https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2833#discussion_r91315815).
- [ ] Move snippets to `docs/vision_snippets.py`.
- [ ] Add more examples and update documentation(#2754).
- [ ] Refactor unit tests.

lukesneeringer:
Long since launched.

PR #2778, opened 2016-11-30
Add loading image from filename.

daspecster:
Towards #2753

Adding support to pass in a filename when instantiating an `Image`.

### DIFF
--- a/docs/vision-usage.rst
+++ b/docs/vision-usage.rst
@@ -1,8 +1,11 @@
+####################
 Using the Vision API
-====================
+####################
 
+
+********************************
 Authentication and Configuration
---------------------------------
+********************************
 
 - For an overview of authentication in ``google-cloud-python``,
   see :doc:`google-cloud-auth`.
@@ -31,8 +34,43 @@ or pass in ``credentials`` and ``project`` explicitly.
      >>> client = vision.Client(project='my-project', credentials=creds)
 
 
+*****************************************************
+Creating an :class:`~google.cloud.vision.image.Image`
+*****************************************************
+
+The :class:`~google.cloud.vision.image.Image` class is used to load image
+data from sources such as a Google Cloud Storage URI, raw bytes, or a file.
+
+
+From a Google Cloud Storage URI
+===============================
+
+.. code-block:: python
+
+    >>> from google.cloud import vision
+    >>> client = vision.Client()
+    >>> image = client.image(source_uri='gs://my-test-bucket/image.jpg')
+
+
+From a filename
+===============
+
+.. code-block:: python
+
+    >>> image = client.image(filename='image.jpg')
+
+From raw bytes
+==============
+
+.. code-block:: python
+
+    >>> with open('./image.jpg', 'rb') as image_file:
+    ...     bytes_image = client.image(content=image_file.read())
+
+
+****************
 Manual Detection
-~~~~~~~~~~~~~~~~
+****************
 
 You can call the detection method manually.
 
@@ -60,8 +98,9 @@ You can call the detection method manually.
     'github'
 
 
+**************
 Face Detection
-~~~~~~~~~~~~~~
+**************
 
 :meth:`~google.cloud.vision.image.Image.detect_faces` will search for faces in
 an image and return the coordinates in the image of each `landmark type`_ that
@@ -87,8 +126,9 @@ was detected.
     0.02545464
 
 
+***************
 Label Detection
-~~~~~~~~~~~~~~~
+***************
 
 :meth:`~google.cloud.vision.image.Image.detect_labels` will attempt to label
 objects in an image. If there is a car, person and a dog in the image, label
@@ -107,8 +147,9 @@ certainty from ``0.0 to 1.0``.
     0.9863683
 
 
+******************
 Landmark Detection
-~~~~~~~~~~~~~~~~~~
+******************
 
 :meth:`~google.cloud.vision.image.Image.detect_landmarks` will attempt to
 detect landmarks such as "Mount Rushmore" and the "Sydney Opera House". The API
@@ -133,8 +174,9 @@ will also provide their known geographical locations if available.
     162
 
 
+**************
 Logo Detection
-~~~~~~~~~~~~~~
+**************
 
 With :meth:`~google.cloud.vision.image.Image.detect_logos`, you can identify
 brand logos in an image. Their shape and location in the image can be found by
@@ -162,8 +204,9 @@ iterating through the detected logo's ``vertices``.
     62
 
 
+*********************
 Safe Search Detection
-~~~~~~~~~~~~~~~~~~~~~
+*********************
 
 :meth:`~google.cloud.vision.image.Image.detect_safe_search` will try to
 categorize the entire contents of the image under four categories.
@@ -192,8 +235,9 @@ categorize the entire contents of the image under four categories.
     'LIKELY'
 
 
+**************
 Text Detection
-~~~~~~~~~~~~~~
+**************
 
 :meth:`~google.cloud.vision.image.Image.detect_text` performs OCR to find text
 in an image.
@@ -213,8 +257,9 @@ in an image.
     'some other text in the image'
 
 
+****************
 Image Properties
-~~~~~~~~~~~~~~~~
+****************
 
 :meth:`~google.cloud.vision.image.Image.detect_properties` will process the
 image and determine the dominant colors in the image.
@@ -238,8 +283,9 @@ image and determine the dominant colors in the image.
     0.758658
 
 
+****************
 No results found
-~~~~~~~~~~~~~~~~
+****************
 
 If no results for the detection performed can be extracted from the image, then
 an empty list is returned. This behavior is similiar with all detection types.
--- a/vision/google/cloud/vision/image.py
+++ b/vision/google/cloud/vision/image.py
@@ -30,6 +30,9 @@ class Image(object):
     :type content: bytes
     :param content: Byte stream of an image.
 
+    :type filename: str
+    :param filename: Filename to image.
+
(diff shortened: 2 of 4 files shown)


docs/vision-usage.rst
```diff
@@ -31,8 +31,38 @@ or pass in ``credentials`` and ``project`` explicitly.
      >>> client = vision.Client(project='my-project', credentials=creds)
 
 
+Creating an :class:`~google.cloud.vision.image.Image` object.
```
dhermes: 1. Does a header need to end in a period?
1. Isn't ending with `Image` clear? (i.e. don't say "object")

docs/vision-usage.rst
```diff
@@ -31,8 +31,38 @@ or pass in ``credentials`` and ``project`` explicitly.
...
+-------------------------------------------------------------
+
+The :class:`~google.cloud.vision.image.Image` object is used to load your image
```
dhermes: 1. s/object/class
1. ~~your~~

docs/vision-usage.rst
```diff
@@ -31,8 +31,38 @@ or pass in ``credentials`` and ``project`` explicitly.
...
+data from sources such as a Google Cloud Storage URI, raw bytes, or a file.
+
+
+**From a Google Cloud Storage URI**
```
dhermes: Use a sub-heading?

docs/vision-usage.rst
```diff
@@ -214,7 +244,7 @@ in an image.
 
 
 Image Properties
-~~~~~~~~~~~~~~~~
+----------------
```
dhermes: Changes made per the recs? http://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#headings

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,28 @@ class Image(object):
...
+    def __init__(self, client, content=None, filename=None, source_uri=None):
         self.client = client
-        self._content = None
-        self._source = None
 
-        if source_uri:
-            self._source = source_uri
-        else:
-            self._content = _bytes_to_unicode(b64encode(_to_bytes(content)))
+        if filename:
```
dhermes: Always remember your None-checks: `if filename is not None`

nunchucks

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,28 @@ class Image(object):
...
+            with open(filename, 'rb') as file_obj:
+                content = file_obj.read()
+
+        if content:
```
dhermes: Ditto here on None-check

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,28 @@ class Image(object):
...
+
     :type source_uri: str
     :param source_uri: Google Cloud Storage URI of image.
 
     :type client: :class:`~google.cloud.vision.client.Client`
     :param client: Instance of Vision client.
     """
 
-    def __init__(self, client, content=None, source_uri=None):
+    def __init__(self, client, content=None, filename=None, source_uri=None):
```
dhermes: You need an explicit check that makes sure exact one of `content`, `filename` and `source_uri` is not None. The first thing that comes to mind is:

```python
(int(content is not None) + int(filename is not None) + 
 int(source_uri is not None)) == 1
# OR FEWER CHARS
(int(content is None) + int(filename is None) + 
 int(source_uri is None)) == 2
```

though it feels kind of icky.

Maybe @jonparrott or @tseaver can make me feel better about writing code like that / suggest something less icky.

vision/tox.ini
```diff
@@ -7,6 +7,7 @@ localdeps =
     pip install --quiet --upgrade {toxinidir}/../core
 deps =
     {toxinidir}/../core
+    mock
```
dhermes: Why am I the only one that has screwed up k-l-M-n-o-P?

vision/unit_tests/test_image.py
```diff
@@ -36,26 +36,26 @@ def _make_one(self, *args, **kw):
     def test_image_source_type_content(self):
         image = self._make_one(CLIENT_MOCK, content=IMAGE_CONTENT)
 
-        _AS_DICT = {
+        _as_dict = {
```
dhermes: Since you're editing here, you could add a trailing comma (if it pleases you)?

```python
_as_dict = {
    'content': B64_IMAGE_CONTENT',
}
# or go one-line
_as_dict = {'content': B64_IMAGE_CONTENT'}
```

vision/unit_tests/test_image.py
```diff
@@ -36,26 +36,26 @@ def _make_one(self, *args, **kw):
...
         self.assertEqual(B64_IMAGE_CONTENT, image.content)
         self.assertEqual(None, image.source)
-        self.assertEqual(_AS_DICT, image.as_dict())
+        self.assertEqual(image.as_dict(), _as_dict)
 
     def test_image_source_type_google_cloud_storage(self):
         image = self._make_one(CLIENT_MOCK, source_uri=IMAGE_SOURCE)
 
-        _AS_DICT = {
+        _as_dict = {
```
dhermes: Also, why `_as_dict` instead of `as_dict`? Who are we hiding from in a unit test?

docs/vision-usage.rst
daspecster: Sure, I can use `~~~~~~` but that looks really close in size to `-------` when rendered.

daspecster: NM, I found the rest of the header sizes.

docs/vision-usage.rst
daspecster: Yeah just to make it match up. But looking at that, I will redo it with the proper markup starting with `H1`'s

vision/google/cloud/vision/image.py
daspecster: Maybe a helper function but what about something like...

```python
sources = [content, filename, source_uri]
if sum(map(bool, sources)) <> 1:
   print('Please specify one source.'
```

vision/unit_tests/test_image.py
daspecster: search/replace/destroy.

vision/tox.ini
daspecster: Should I just try and pull in the deps from the root `tox.ini`?

vision/google/cloud/vision/image.py
tseaver:
```python
sources = set(source for source in (content, filename, source_uri)
              if source is not None)
if len(sources) != 1:
    raise ValueError(
        "Specify exactly one of 'content', 'filename', or 'source_uri'.")
```

daspecster: I added this in so we could look at it. Also I added the tests as well to look at.

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2778/commits/a431df1373f7cc0021186bdffc89bfe1dcc0ec08

dhermes: @tseaver Casting to a set isn't necessary / correct, the `if source is not None` will filter out everything needed.

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,35 @@ class Image(object):
...
+        sources = set(source for source in (content, filename, source_uri)
+                      if source is not None)
```
dhermes: As above, don't use `set()`, it's "wasteful" (though negligible, since it is creating a hash table from at most 3 values).

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,35 @@ class Image(object):
...
+        if len(sources) != 1:
+            raise ValueError(
+                'Specify exactly one of \'content\', \'filename\', or'
+                '\'source_uri\'.')
```
dhermes: Instead of 6 backslashes, pick one of:

```python
raise ValueError(
    'Specify exactly one of "content", "filename", or '
    '"source_uri".')
# OR
raise ValueError(
    "Specify exactly one of 'content', 'filename', or "
    "'source_uri'.")
```

(Also when you do line concat of strings, you should make sure you have space chars in betwee, i.e. `'... or '` not `'... or'`)

vision/google/cloud/vision/image.py
```diff
@@ -30,22 +30,35 @@ class Image(object):
...
-        self._content = None
-        self._source = None
 
-        if source_uri:
-            self._source = source_uri
-        else:
-            self._content = _bytes_to_unicode(b64encode(_to_bytes(content)))
+        if filename is not None:
+            with open(filename, 'rb') as file_obj:
+                content = file_obj.read()
```
dhermes: Do this after the `content is not None` check? What's the point of the transformation below?

vision/tox.ini
dhermes: No. You did the right thing, `mock < pytest` in alphabetical order. I screwed it up elsewhere.

vision/google/cloud/vision/image.py
daspecster: Since `filename` is a helper, I just grab the data from the file and put into `content`.
Then `content` is used as it normally would.

So if you pass in an empty file, that is when `content` would be `None`.

dhermes: OK

vision/google/cloud/vision/image.py
daspecster: I switched the `set()` to a `[]`. I think that was the implication there?

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2778/commits/b77ed781fcfb48389d289f9e97cd64b7602707d7

daspecster:
Thanks @dhermes! I cancelled travis and rerun after squashing.

Issue #2526, opened 2016-10-11
Add Vision system tests


daspecster:
Sadly, Vision has no system tests. They must be added.

daspecster:
Also forgot to close this.

PR #2788, opened 2016-11-30
Add system test for Vision Logo detection.

daspecster:
Towards: #2753 and #2526.

I will add the filename test after #2778 is merged.
To get this kicked off though, here are some system tests for `detect_logo()`.


system_tests/run_system_test.py
```diff
@@ -27,11 +27,13 @@
 import storage
 import system_test_utils
 import translate
+import vision
 
 
 TEST_MODULES = {
     'datastore': datastore,
     'speech': speech,
+    'vision': vision,
```
dhermes: Why here vs. somewhere else in the dict? (Not that it matters)

system_tests/vision.py
```diff
@@ -0,0 +1,98 @@
...
+
+from system_test_utils import unique_resource_id
+from retry import RetryErrors
+
+
+class Config(object):
+    CLIENT = None
+    TEST_BUCKET = None
+
+    LOGO_FILE = os.path.join(os.path.dirname(__file__), 'data', 'logo.png')
```
dhermes: This doesn't need to be in `Config`, the `Config` is intended for run-time only values, but this one can be determined at import time without issue. It helps to include the docstring on `Config` to clarify that goal.

Also, filesystem paths can be annoying sometimes. I recommend the following (as globals, not as attrs on `Config`):

```python
_SYS_TESTS_DIR = os.path.abspath(os.path.dirname(__file__))
LOGO_FILE = os.path.join(_SYS_TESTS_DIR, 'data', 'logo.png')
```

system_tests/vision.py
```diff
@@ -0,0 +1,98 @@
...
+
+def tearDownModule():
+    # 409 Conflict if the bucket is full.
+    # 429 Too Many Requests in case API requests rate-limited.
+    bucket_retry = RetryErrors(
+        (exceptions.TooManyRequests, exceptions.Conflict))
+    bucket_retry(Config.TEST_BUCKET.delete)(force=True)
+
+
+class TestSpeechClient(unittest.TestCase):
```
dhermes: You can probably just say `TestVision` here?

system_tests/vision.py
```diff
@@ -0,0 +1,98 @@
...
+    def setUp(self):
+        self.to_delete_by_case = []
+
+    def tearDown(self):
+        for value in self.to_delete_by_case:
+            value.delete()
+
+    def _assert_logo(self, logo):
+        self.assertEqual(logo.description, 'Google')
```
dhermes: Add an `assertIsInstance` too?

system_tests/run_system_test.py
daspecster: So I read `MODULES = (  # ordered from most to least stable` [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/7cd15697f802252ab5c4ac062a0f3c4b2b80856b/system_tests/attempt_system_tests.py#L59) and I figured that speech and vision should be similarly stable. These seemed to be in the same order so that's the only reason.

system_tests/vision.py
daspecster: >Also, filesystem paths can be annoying sometimes. 

By this, do you mean that `abspath` should be used?

system_tests/vision.py
daspecster: O.o....

system_tests/run_system_test.py
dhermes: But this isn't that variable, this is a dictionary, which has no ordering

system_tests/vision.py
dhermes: Yes

system_tests/run_system_test.py
dhermes: See: "(Not that it matters)"

daspecster:
I added the filename test as well.

```bash
system-tests runtests: commands[1] | python /Users/daspecster/Documents/google/google-cloud-python/system_tests/attempt_system_tests.py vision
test_detect_logos_content (vision.TestVisionClient) ... ok
test_detect_logos_filename (vision.TestVisionClient) ... ok
test_detect_logos_gcs (vision.TestVisionClient) ... ok

Ran 3 tests in 5.540s

OK
____________________________________________________ summary _____________________________________________________
  congratulations :)
```

daspecster:
Squashed!

PR #2789, opened 2016-12-01
Add support for standard SQL 'BOOL' type.

tseaver:
Alias for 'BOOLEAN' in legacy SQL.


tseaver:
h/t to @blowmage for [pointing out the alias](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2776#discussion_r90483766).

PR #2793, opened 2016-12-02
Fix Markdown syntax in README to RST

waprin:
@dhermes sorry about that, this time I looked and it looks right.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -109,7 +109,7 @@ See `CONTRIBUTING`_ for more information on how to get started.
 Community
 ---------
 
-Google Cloud Platform Python developers hang out in `Slack`_ in the ```#python``
+Google Cloud Platform Python developers hang out in `Slack`_ in the ``#python``
 channel, click here to `get an invitation`_.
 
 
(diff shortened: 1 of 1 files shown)


dhermes:
Indeed, looks good: https://github.com/waprin/google-cloud-python/blob/fix_rst/README.rst

I am happy to merge, though should `#python` be in code font? (Your call)

README.rst
```diff
@@ -109,8 +109,12 @@ See `CONTRIBUTING`_ for more information on how to get started.
 Community
 ---------
 
-Google Cloud Platform Python developers hang out in [Slack](https://googlecloud-community.slack.com) in the #python channel (get an invitation [here](https://gcp-slack.appspot.com/)). 
+Google Cloud Platform Python developers hang out in `Slack`_ in the #python
+channel, click here to `get an invitation`_.
 
+
+.. _Slack: https://googlecloud-community.slack.com
+.. _get an invitation: https://gcp-slack.appspot.com/
```
daspecster: If you're making the codefont change that @dhermes mentioned then I think it might be good to add a space between this line and the `License` heading as well.

waprin:
done and done

PR #2795, opened 2016-12-02
Upgrading GAPIC deps to 0.14 series.

dhermes:
In the process, ditching **hard** deps on protobufs, `grpcio`, and `google-gax` in those packages. Also had
to upgrade calls to `list_log_entries()` based on: https://github.com/googleapis/api-client-staging/pull/118/files/2bcd2875a578ae525d2aabb862cf9c131b4665f5#r90422054

/cc @geigerj @bjwatson 

@waprin I could not for the life of me figure out how to use `resource_names` instead of the deprecated `project_ids`. Can you explain? (Trying `foo` --> `project/foo` did not work)

### DIFF
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -16,19 +16,19 @@
 
 import functools
 
-from google.cloud.gapic.logging.v2.config_service_v2_api import (
-    ConfigServiceV2Api)
-from google.cloud.gapic.logging.v2.logging_service_v2_api import (
-    LoggingServiceV2Api)
-from google.cloud.gapic.logging.v2.metrics_service_v2_api import (
-    MetricsServiceV2Api)
+from google.cloud.gapic.logging.v2.config_service_v2_client import (
+    ConfigServiceV2Client)
+from google.cloud.gapic.logging.v2.logging_service_v2_client import (
+    LoggingServiceV2Client)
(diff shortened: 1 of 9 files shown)


tseaver:
@dhermes Should we work out what the story is supposed to be for `resource_names` before merging?

dhermes:
@tseaver The system tests are still passing, so I don't think it's necessary to block on that, though we should get it straightened out. (Was there a version bump for this API I wasn't aware of that deprecated the field?)

waprin:
@dhermes did you try `foo` -> `projects/foo` ? (note the plural projects). Apologies if you just had a typo in your original message, but I do hope that was the problem since it's an easy fix :)

dhermes:
This is what I did before (I know because my chat log with @geigerj)

```python
resource_names = ['project/' + project for project in projects]
```

@waprin Should we allow callers to send **other** resource types? I checked and only project IDs come through in the system tests.

geigerj:
@dhermes `project/` -> `projects/`?

dhermes:
@geigerj Right that's what @waprin just said (and I confirmed I was holding the gun wrong before)

waprin:
Docs look like there are some "orgazinations/" thing but I'm not familiar with it tbh, will have to ask around.

dhermes:
@tseaver Let's merge this and continue this discussion elsewhere (like a PR which converts `projects` to `resource_names`)

tseaver:
LGTM

PR #2800, opened 2016-12-02
Removing os.environ patching in doctest.

dhermes:
It turns out Sphinx allows this outside of a sandbox.


dhermes:
See breakage: https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/180845187#L3513

Merging now

PR #2803, opened 2016-12-03
Coerce query parameter values from native types to JSON-appropriate types

### LINKED ISSUE #2799: Query paramters need to coerce native Python values
If the user passes in non-string values (e.g., for `INT64`, which must be tunnelled via strings because JS reasons, or `TIMESTAMP`, `DATETIME`, `DATE`), the parameter must coerce the value appropriately inside its `to_api_repr()` method.

h/t to @dwsupplee for the prod.

tseaver:
Also, distinguish parsing 'DATETIME' vs. 'TIMESTAMP' properly when processing row data.

Closes #2799.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -15,9 +15,12 @@
 """Shared helper functions for BigQuery API classes."""
 
 from collections import OrderedDict
+import datetime
 
-from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _date_from_iso8601_date
+from google.cloud._helpers import _datetime_from_microseconds
+from google.cloud._helpers import _datetime_to_rfc3339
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -43,13 +46,20 @@ def _bool_from_json(value, field):
...
     if _not_null(value, field):
         # value will be a float in seconds, to microsecond precision, in UTC.
         return _datetime_from_microseconds(1e6 * float(value))
 
 
+def _datetime_from_json(value, field):
+    """Coerce 'value' to a datetime, if set or not nullable."""
+    if _not_null(value, field):
+        # value will be a string, in YYYY-MM-DDTHH:MM:SS form.
+        return datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S')
```
dhermes: @tseaver Don't we already have this template in `google.cloud._helpers`? If not, please add it there.

PR #2805, opened 2016-12-05
Allow timestamp strings in 'Table.insert_data'.

### LINKED ISSUE #1382: Support both seconds from epoch and timestamp string
If someone already has a timestamp string, we shouldn't force them to convert it to `datetime.datetime` so that we can in turn [convert that](https://github.com/GoogleCloudPlatform/gcloud-python/blob/0c74a599b6aa0a16e21575998e1770182c05e12c/gcloud/bigquery/table.py#L665-L669) into seconds. (Inspired by #1381.)

From [the docs](https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatype)

> TIMESTAMP data types can be described in two ways: UNIX timestamps or calendar datetimes. BigQuery stores TIMESTAMP data internally as a UNIX timestamp with microsecond precision.

so we could allow the user to just pass a string and we "take their word for it" that it's a valid datetime. Since `Table.insert_data` returns a list of errors anyhow, we wouldn't have to do much/anything to enforce this.

tseaver:
Closes #1382.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -738,11 +738,6 @@ class Table(object):
         rows_info = []
         data = {'rows': rows_info}
 
-        def _convert_timestamp(value):
-            if isinstance(value, datetime.datetime):
-                value = _microseconds_from_datetime(value) * 1e-6
-            return value
-
         for index, row in enumerate(rows):
             row_info = {}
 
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -738,15 +738,20 @@ def insert_data(self,
         rows_info = []
         data = {'rows': rows_info}
 
+        def _convert_timestamp(value):
```
dhermes: Please define this in the global scope

dhermes:
@tseaver Easy enough change to review, ping me once new function is defined as a global

tseaver:
@dhermes ping

shealutton:
Thanks! I needed exactly this change. Much obliged.

PR #2806, opened 2016-12-05
Marshal 'BYTES' and 'TIME' column / paramter types

### LINKED ISSUE #2229: New Standard SQL Data Types Not Supported in gcloud._helpers._rows_from_json() / _CELLDATA_FROM_JSON()
Standard SQL dialect support was added in 0.18.0, but current release does not support queries that return results that include data types that are not supported by Legacy SQL.

E.g. executing the following query using Standard SQL:

```
SELECT PARSE_DATE('%Y%m%d', '20160810') AS table_date;
```

results in 

```
rows = query.rows
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gcloud/bigquery/query.py", line 204, in rows
    return _rows_from_json(self._properties.get('rows', ()), self.schema)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gcloud/bigquery/_helpers.py", line 85, in _rows_
from_json
    converter = _CELLDATA_FROM_JSON[field.field_type]
KeyError: u'DATE'
```

Not a significant issue since the Legacy SQL Data types will work for my needs, but wanted to report it all the same.

tseaver:
Closes #2229.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -14,6 +14,7 @@
 
 """Shared helper functions for BigQuery API classes."""
 
+import base64
 from collections import OrderedDict
 import datetime
 
@@ -50,6 +51,12 @@ def _bool_from_json(value, field):
 def _string_from_json(value, _):
     """NOOP string -> string coercion"""
     return value
+
+
+def _bytes_from_json(value, field):
+    """Base64-decode value"""
+    if _not_null(value, field):
+        return base64.decodestring(value)
 
 
 def _timestamp_from_json(value, field):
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -64,9 +77,17 @@ def _datetime_from_json(value, field):
 def _date_from_json(value, field):
     """Coerce 'value' to a datetime date, if set or not nullable"""
     if _not_null(value, field):
+        # value will be a string, in YYYY-MM-DD form.
```
daspecster: I know this method is private but you could just add
```
:type value: str
:param value: String in YYYY-MM-DD form.
```

dhermes: :+1:

tseaver:
I will merge as soon as either Circle CI or Travis passes the Py3k fix I just pushed.

tseaver:
Travis is showing [build failures for docs](https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/181434795#L1784-L1785) which don't happen on Circle or on my local machine.

PR #2807, opened 2016-12-05
Add back NO COVER to gRPC ImportError branch

jonparrott:


core/google/cloud/_helpers.py
```diff
@@ -34,7 +34,7 @@
     import grpc
     from google.auth.transport.grpc import (
         AuthMetadataPlugin)  # pragma: NO COVER
-except ImportError:
+except ImportError:  # pragma: NO COVER
```
dhermes: Doesn't this mean the one above doesn't belong

jonparrott: No, because-

case 1: grpc isn't available, `import grpc` raises `ImportError` and the second import is never executed.
case 2: grpc is available, both imports succeed and the `except` is never executed.

To remove the one above, I'd need to make a case where `import grpc` works but `import google.auth.transport.grpc` doesn't, which for all practical purposes is meaningless.

dhermes: But the presence of your `# pragma: NO COVER` here indicates `grpc` **is** present in the relevant environment

jonparrott: It is in one of the several environments you run coverage in, but not in another.

dhermes: So let's put it in all them?

dhermes:
@jonparrott Let's close this?

PR #2808, opened 2016-12-05
Make make_secure_channel use google-auth

### LINKED ISSUE #2797: Drop _helpers.make_secure_channel in favor of google.auth.transports.grpc.secure_authorized_channel
Unless there is a compelling reason to keep it @dhermes?

jonparrott:
Resolves #2797

### DIFF
--- a/core/tox.ini
+++ b/core/tox.ini
@@ -4,9 +4,9 @@ envlist =
 
 [testing]
 deps =
+    grpcio
     mock
     pytest
-    grpcio
 covercmd =
     py.test --quiet \
       --cov=google.cloud \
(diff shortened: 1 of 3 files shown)


jonparrott:
@dhermes note that this adds `grpc` as a test dep for core.

core/tox.ini
```diff
@@ -6,6 +6,7 @@ envlist =
 deps =
     mock
     pytest
+    grpcio
```
dhermes: Alphabet?

jonparrott: Fixed.

PR #2809, opened 2016-12-05
Allow overriding dataset's project during construction.

### LINKED ISSUE #2118: bigquery dataset should not always take project argument from client
The project in which a dataset resides is not necessarily the project in which jobs involving the dataset will be executed. For example, if querying the bigquery public data hacker news dataset the following will error

```
client = bigquery.Client(project='bigquery-public-data')
dataset = client.dataset('hacker_news')
client.extract_table_storage(..., dataset.table('comments'))
```

Because the user can't create a job in the project 'bigquery-public-data'. Instead the fairly convoluted:

```
client = bigquery.Client(project='my-project')
client.extract_table_storage(..., bigquery.Client(project='bigquery-public-data').dataset('hacker_news').table('comments'))
```

Despite the fact that the client involved will never be used.

The proposed fix is to add a kwarg `project` to the Dataset class which overrides the originating project.

tseaver:
Closes #2118.

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -115,7 +115,7 @@ class Dataset(object):
         self._properties = {}
         # Let the @property do validation.
         self.access_grants = access_grants
-        self._project = project
+        self._project = project or client.project
 
     @property
     def project(self):
@@ -124,7 +124,7 @@ class Dataset(object):
         :rtype: str
         :returns: the project (derived from the client).
         """
-        return self._project or self._client.project
+        return self._project
 
     @property
     def path(self):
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -119,7 +124,7 @@ def project(self):
         :rtype: str
         :returns: the project (derived from the client).
         """
-        return self._client.project
+        return self._project or self._client.project
```
dhermes: Shouldn't this just happen once in the constructor?

tseaver: Maybe so, given that a client instance normally won't change its project (leaving it dynamic allows for that possibility).

tseaver:
@daspecster We don't have any system tests which use datasets other than those they create / tear down.  I guess we could write one which uses one of the "public" datasets.

daspecster:
@tseaver if you think there's value in having that system test. The reason I thought of it was due to your assert on `dataset.path` in the unit tests. Since that references the remote resource it might be good to test that for real as well as the mock.

Issue #2810, opened 2016-12-05
Tagging Strategy with sub-packages


dhermes:
I used the "question" label but this is more like an informative issue. Supersedes #2462.

Tag names will be as follows:

- `{pkg}-0.x.y` for releases to the `google-cloud-{pkg}` PyPI package, e.g. `core-0.21.1`, `datastore-0.22.0`, etc.
- `0.x.y` for releases of the umbrella package (as we had already been doing)

----

- For [hand-rolled stable][2] builds (i.e. `gh-pages`), we'll have a custom script which install from the latest version of each package on PyPI, rather than from the source tree (this may be a chicken-and-egg problem on tag builds that would both push docs **and** push to PyPI)
- We ~~will be disabling~~ have disabled (as of 1:47pm Pacific, Dec. 5, 2016) [RTD stable][1] builds: we can't have local filesystem [deps][3] on `latest` builds and then have PyPI deps on `stable` builds (RTD [subprojects][4] may help)

[1]: http://google-cloud-python.readthedocs.io/en/stable/
[2]: https://googlecloudplatform.github.io/google-cloud-python/stable/
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6916814c94f6329ce036c01093e7728961c24789/docs/requirements.txt
[4]: http://read-the-docs.readthedocs.io/en/latest/faq.html#how-do-i-host-multiple-projects-on-one-cname

PR #2812, opened 2016-12-05
Removing links to "stable" build of RTD.

dhermes:
See #2810 for context.


dhermes:
The Travis breakage was just fixed in #2811. Merging.

PR #2815, opened 2016-12-05
Cutting new releases for BigQuery, Language and Core.

dhermes:
H/T to

- @jmdobry for #2798 forcing the NL release
- @tseaver for #2809, #2806, #2805, #2803, #2776, #2787 and #2789 forcing the BQ release
- @jonparrott #2808, #2801, #2726 forcing the core release

Proposed tags after merge:

- `bigquery-0.22.0`
- `core-0.22.0`
- `language-0.22.0`

I'll start going through `$ git log 0.21.0..HEAD -- bigquery/ core/ language/` now to figure out release notes.

PR #2817, opened 2016-12-05
Put hard upper-bounds on umbrella package.

dhermes:
This is needed to that new versions of sub-packages don't pull in conflicting versions of core or other
dependencies.

This needs to get pushed to PyPI when #2815 does

----

Proposed tag: `0.21.1`


setup.py
```diff
@@ -50,27 +50,27 @@
...
-    'google-cloud-monitoring >= 0.21.0',
-    'google-cloud-pubsub >= 0.21.0',
-    'google-cloud-resource-manager >= 0.21.0',
-    'google-cloud-storage >= 0.21.0',
-    'google-cloud-translate >= 0.21.0',
-    'google-cloud-vision >= 0.21.0',
-    'google-cloud-runtimeconfig >= 0.21.0',
+    'google-cloud-bigquery >= 0.21.0, < 0.22dev',
+    'google-cloud-bigtable >= 0.21.0, < 0.22dev',
+    'google-cloud-happybase >= 0.21.0, < 0.22dev',
```
dhermes: Note: I **just** released this https://pypi.python.org/pypi/google-cloud-happybase/0.21.0

PR #2818, opened 2016-12-05
Fix 'BYTES' field handling on Py3k.

tseaver:
JSON decodes the base64-encoded bits as text, which cannot be passed to 'base64.decodestring' on Py3k.

`tox -e system-tests3` was failing for me locally before this patch, and passes with it.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -18,12 +18,15 @@ import base64
 from collections import OrderedDict
 import datetime
 
+import six
+
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud._helpers import _microseconds_from_datetime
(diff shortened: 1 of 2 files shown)


dhermes:
@tseaver: See "but please use the core helper"

tseaver:
@dhermes I [updated it](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2818/files#diff-7df7ab17bcc39fd5c3ead6404a2b90b9R64) (via `git commit --amend`) to do so.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -57,7 +60,8 @@ def _string_from_json(value, _):
 def _bytes_from_json(value, field):
     """Base64-decode value"""
     if _not_null(value, field):
-        return base64.decodestring(value)
+        return base64.decodestring(
+            _to_bytes(value) if isinstance(value, six.text_type) else value)
```
dhermes: @tseaver The `if isinstance()` check happens within `_to_bytes`, i.e. that's what it's for

Sorry I wasn't clear on that.

PR #2823, opened 2016-12-06
Removing support for API_KEY auth in translate.

### LINKED ISSUE #2734: Remove support for API_KEY in translation
See #2731

dhermes:
Fixes #2734.

/cc @jonparrott @jgeewax

### DIFF
--- a/translate/google/cloud/translate/client.py
+++ b/translate/google/cloud/translate/client.py
@@ -15,7 +15,6 @@
 """Client for interacting with the Google Cloud Translate API."""
 
 
-import httplib2
 import six
 
 from google.cloud._helpers import _to_bytes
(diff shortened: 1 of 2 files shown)


lukesneeringer:
Hey folks -- what was the reason why we did this? I went down the rabbit hole and I can see that everyone agreed it was a good idea, but nowhere can I find an explanation as to why.

Cards on table: I had a "bug" filed on this recently, and so I need to grok what the appropriate answer is (e.g. whether to put this support back, or say that we are not going to do it because ______).

daspecster:
@lukesneeringer, as I recall, it was mostly because the API key auth was outside of the normal authentication methods and I believe it was said that API key based auth was going away..or they(backend) were trying to not support it in the future.

tseaver:
@lukesneeringer the description @daspecster gave matches my memory. Sort of an extension of "Special cases aren't special enough to break the rules."

jonparrott:
API keys are strongly being discouraged for Cloud APIs, especially for client libraries targeting backend languages. Omar has more details. :)

lukesneeringer:
Thanks. :-)

PR #2825, opened 2016-12-06
Support authenticated local non-TLS for datastore and pubsub

jcoatgoogle:
Hello, this is a very early stab at the foundation of a change in line with some internal discussion.

My goal is to
1. validate the approach
2. make sure it works for HTTP, not just GRPC (I wanted to validate it first)
3. get guidance on how to test this
4. extend this to the rest of veneer

Implementation wise, one remaining thing is to add a catchall env variable.... but that is trivial to add. I wanted to validate the approach.

And 3... I'm not sure if we have integration tests or what. Things like this change are always tricky to test, so any guidance is very welcome!

### DIFF
--- a/core/google/cloud/_helpers.py
+++ b/core/google/cloud/_helpers.py
@@ -27,6 +27,7 @@ import re
 from threading import local as Local
 
 import google.auth
+from google.cloud.environment_vars import AUTH_LOCAL
 from google.protobuf import timestamp_pb2
 import google_auth_httplib2
 
@@ -545,6 +546,40 @@ def make_insecure_stub(stub_class, host, port=None):
         target = '%s:%d' % (host, port)
     channel = grpc.insecure_channel(target)
     return stub_class(channel)
+
+
+def get_authorization_metadata(credentials, host):
+  """Gets Bearer token for authorization header.
+
+  This is a helper method which allows getting access to the Bearer token for
+  special use cases. Having access to the authorization header allows for
+  authenticated plaintext traffic. Note: this token is sensitive! As such, we
+  only return the token when an environment veriable (defined by AUTH_LOCAL) is
+  true, and the connection is with localhost (minimizing the chance of leaking
+  the token externally). Having access to the authorization header allows for
+  authenticated plaintext traffic. If you use this method, you MUST take
+  special care to ensure you do not leak your authentication credentials.
+
+  :type credentials: :class:`oauth2client.client.OAuth2Credentials`
+  :param credentials: The OAuth2 Credentials to use for this connection.
+
+  :type host: string
+  :param host: the host to route traffic to
+
+  :rtype: (string, string) or None
+  :returns: authorization header (if appropriate)
+  """
+  auth_local = os.environ.get(AUTH_LOCAL)
+  parse_host = re.match("(?:http://)?(\w+)(?:\d*)?", host)
+  is_local = (parse_host is not None and
+              parse_host.group(0) == 'localhost')
+  if (is_local and auth_local is not None and
+      auth_local.lower() in ['1', 't', 'true']):
+    access_token = credentials.get_access_token().access_token
+    return ('authorization', 'Bearer ' + access_token)
+  else:
+    return None
+
 
 
 try:
--- a/core/google/cloud/environment_vars.py
+++ b/core/google/cloud/environment_vars.py
@@ -18,6 +18,10 @@ These enable many types of implicit behavior in both production
 and tests.
 """
 
+AUTH_LOCAL = 'GOOGLE_AUTH_LOCAL_TRAFFIC'
+"""Environment variable which, if true, will allow authorization header to be
+present in insecure local traffic."""
+
 GCD_DATASET = 'DATASTORE_DATASET'
 """Environment variable defining default dataset ID under GCD."""
 
--- a/datastore/google/cloud/datastore/_http.py
+++ b/datastore/google/cloud/datastore/_http.py
@@ -15,12 +15,14 @@
 """Connections to Google Cloud Datastore API servers."""
 
 import contextlib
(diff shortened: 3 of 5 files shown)


core/google/cloud/_helpers.py
```diff
@@ -547,6 +548,40 @@ def make_insecure_stub(stub_class, host, port=None):
...
+  auth_local = os.environ.get(AUTH_LOCAL)
+  parse_host = re.match("(?:http://)?(\w+)(?:\d*)?", host)
+  is_local = (parse_host is not None and
+              parse_host.group(0) == 'localhost')
+  if (is_local and auth_local is not None and
+      auth_local.lower() in ['1', 't', 'true']):
+    access_token = credentials.get_access_token().access_token
+    return ('authorization', 'Bearer ' + access_token)
+  else:
+    return None
```
tseaver: Please indent four spaces (`tox -e lint` in the root would highlight that, plus other "style" requirements).

jonparrott:
afaik there's no way to attach credentials to a grpc channel without TLS.

core/google/cloud/_helpers.py
tseaver: Given that the result is basically an HTTP header tuple, would it make more sense to return a possibly-empty list of two-tuples?  An empty would be the "normal" case.  It would make the calling code below simpler, and should be equally simple to handle in the stub-invocation code.  Making it a list would also allow other headers to be added, if needed, in the futurue.

jcoatgoogle:
Oh, and the main testing I have done is setting the respective variable to localhost:8000 which puts it into non-secure more, and then setting up a local proxy that routes the request to GCP. Thus it's non-TLS locally, TLS to GCP, so Veneer currently would not add this authorization information, but with this set, it does, and it works.

jcoatgoogle:
@jonparrott: you're wrong. Look at the change to find out how :)

core/google/cloud/_helpers.py
```diff
@@ -547,6 +548,40 @@ def make_insecure_stub(stub_class, host, port=None):
...
+  :returns: authorization header (if appropriate)
+  """
+  auth_local = os.environ.get(AUTH_LOCAL)
+  parse_host = re.match("(?:http://)?(\w+)(?:\d*)?", host)
+  is_local = (parse_host is not None and
+              parse_host.group(0) == 'localhost')
+  if (is_local and auth_local is not None and
+      auth_local.lower() in ['1', 't', 'true']):
+    access_token = credentials.get_access_token().access_token
+    return ('authorization', 'Bearer ' + access_token)
```
jonparrott: This will fail without recourse after an hour.

core/google/cloud/_helpers.py
```diff
@@ -547,6 +548,40 @@ def make_insecure_stub(stub_class, host, port=None):
...
+  This is a helper method which allows getting access to the Bearer token for
+  special use cases. Having access to the authorization header allows for
+  authenticated plaintext traffic. Note: this token is sensitive! As such, we
+  only return the token when an environment veriable (defined by AUTH_LOCAL) is
+  true, and the connection is with localhost (minimizing the chance of leaking
+  the token externally). Having access to the authorization header allows for
+  authenticated plaintext traffic. If you use this method, you MUST take
+  special care to ensure you do not leak your authentication credentials.
+
+  :type credentials: :class:`oauth2client.client.OAuth2Credentials`
```
jonparrott: We don't use `oauth2client` anymore.

core/google/cloud/_helpers.py
jcoatgoogle: Not sure what you mean

jcoatgoogle: Oh, in that it needs to be refreshed?

core/google/cloud/_helpers.py
jcoatgoogle: will fix style... new to working with tox, so thanks for the heads up. I agree that returning a list is a good choice, will update accordingly.

core/google/cloud/_helpers.py
jonparrott: Yep. Which is exactly what things like the [AuthMetadataPlugin](http://google-auth.readthedocs.io/en/latest/reference/google.auth.transport.grpc.html#google.auth.transport.grpc.AuthMetadataPlugin) typical handle for channel/call credentials.

dhermes:
@jcoatgoogle High level questions:

- Why is "authenticated local non-TLS" necessary?
- Who is it necessary for?
- How is this different than what we [already do][1] with the auth plugin?

[1]: https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/443df66cb6e7960205c1d2108e67220d27408efe/google/auth/transport/grpc.py#L49-L61

dhermes:
daniel.j.hermes@gmail.com (on my profile: https://github.com/dhermes)

dhermes:
> The auth plugin only works if the traffic is TLS

This was an ad-hoc decision by me when implementing emulator support, e.g. [here][1]. We can de-couple the two so that you can turn off TLS without turning off the plugin.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/5f68b4a2d997cbc2fcfa1af586f7cdf9cab36ec0/pubsub/google/cloud/pubsub/_gax.py#L529-L534

jcoatgoogle:
Ah, the referenced doc is internal (it's always weird when internal discussion leads to an outward change).

So yeah, the explanations above will have to suffice (unless I can get permission to share more... will look into that. it's not particularly sensitive)

We want the ability to communicate with a local proxy without having to speak TLS (it's kind of silly to have to do TLS with a local process).

lukesneeringer:
This issue may no longer be relevant due to its age. Feel free to re-open.

PR #2826, opened 2016-12-06
Switch to enums and clean up names.

daspecster:
Towards #2753
Switching `Likelihood` and `LandmarkTypes` to `Enum`'s.

### DIFF
--- a/vision/tox.ini
+++ b/vision/tox.ini
@@ -7,6 +7,7 @@ localdeps =
     pip install --quiet --upgrade {toxinidir}/../core
 deps =
     {toxinidir}/../core
+    enum34
     mock
     pytest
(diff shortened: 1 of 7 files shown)


vision/tox.ini
```diff
@@ -7,6 +7,7 @@ localdeps =
     pip install --quiet --upgrade {toxinidir}/../core
 deps =
     {toxinidir}/../core
+    enum34
```
dhermes: @daspecster You don't need this here, it's in the package requirements

daspecster: Thanks!
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2829

PR #2827, opened 2016-12-06
Add system tests for detect_faces().

daspecster:
Towards #2753 

This adds system test for `detect_faces()`.

```bash
test_detect_faces_content (vision.TestVisionClientFace) ... ok
test_detect_faces_filename (vision.TestVisionClientFace) ... ok
test_detect_faces_gcs (vision.TestVisionClientFace) ... ok
test_detect_logos_content (vision.TestVisionClientLogo) ... ok
test_detect_logos_filename (vision.TestVisionClientLogo) ... ok
test_detect_logos_gcs (vision.TestVisionClientLogo) ... ok

Ran 6 tests in 9.324s

OK
________________________________________________________ summary ________________________________________________________
  congratulations :)
```


system_tests/vision.py
```diff
@@ -108,3 +109,104 @@ def test_detect_logos_gcs(self):
...
+        from google.cloud.vision.face import Landmark
+        from google.cloud.vision.face import FaceLandmarkTypes
+
+        self.assertIsInstance(landmark, Landmark)
+
+        valid_landmark_type = getattr(FaceLandmarkTypes,
+                                      landmark.landmark_type, False)
+        if valid_landmark_type:
+            return True
+        return False
```
tseaver: - No testcase calls this method.
- Its name implies it should raise an exception or return silently (the return value doesn't make sense).

system_tests/vision.py
```diff
@@ -108,3 +109,104 @@ def test_detect_logos_gcs(self):
...
+        self._assert_likelihood(face.anger)
+        self._assert_likelihood(face.joy)
+        self._assert_likelihood(face.sorrow)
+        self._assert_likelihood(face.surprise)
+        self._assert_likelihood(face.image_properties.blurred)
+        self._assert_likelihood(face.image_properties.underexposed)
+        self._assert_likelihood(face.headwear)
+        self.assertGreater(abs(face.angles.roll), 0.0)
+        self.assertGreater(abs(face.angles.pan), 0.0)
+        self.assertGreater(abs(face.angles.tilt), 0.0)
```
tseaver: How meaningful is the assertion?  Would `assertNotEqual(..., 0.0)` be equally descriptive?

system_tests/vision.py
```diff
@@ -108,3 +109,104 @@ def test_detect_logos_gcs(self):
...
+    def setUp(self):
+        self.to_delete_by_case = []
+
+    def tearDown(self):
+        for value in self.to_delete_by_case:
+            value.delete()
+
+    def _assert_coordinate(self, coordinate):
+        if coordinate is None:
+            return True
```
tseaver: As below, just raise if invalid (nobody checks the return value).

system_tests/vision.py
```diff
@@ -108,3 +109,104 @@ def test_detect_logos_gcs(self):
...
+
+        self.assertIn(type(coordinate), [int, float])
```
dhermes: Use `assertIsInstace(coordinate, (int, float))`

system_tests/vision.py
dhermes: @tseaver If this branch was gone and `coordinate` were invalid, it's fail the first assertion

daspecster: So in this case `None` is a valid value but the following assertions should be skipped if it is `None`.

tseaver: So invert the test (`if coordinate is not None:`) and make the assertions inside its suite.  Nobody is checking the return value, which makes it pointless / distracting.

system_tests/vision.py
```diff
@@ -121,9 +121,8 @@ def tearDown(self):
 
     def _assert_coordinate(self, coordinate):
         if coordinate is None:
-            return True
-
-        self.assertIn(type(coordinate), [int, float])
+            return
```
tseaver: OK

daspecster:
Travis is green, merging.

PR #2828, opened 2016-12-06
Raise ValueError if credentials are not from google-auth

### LINKED ISSUE #2814: [Discussion] oauth2client credentials in a post-oauth2client world.
We recently switched `core` to use [`google-auth`](https://google-auth.readthedocs.io) in #2639 and completely removed support for `oauth2client`.

It is possible that some users are explicitly passing `oauth2client` credentials into `Client` constructors, e.g.:

```python
credentials = oauth2client.contrib.gce.AppAppsertionCredentials()
storage_client = storage.Client(credentials=credentials)
```

After `google-cloud-core==0.22.0` is released this code will no longer work. Only credentials from `google-auth` will work. There is a potential that we could break some user code and there may be some valid reasons for sticking with oauth2client in the near-term (user credentials and storage, for example).

`google-auth` offers a way to [`convert`](https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/443df66cb6e7960205c1d2108e67220d27408efe/google/auth/_oauth2client.py#L136) credentials:

```python
import google.auth._oauth2client

new_credentials = google.auth._oauth2client.convert(old_credentials)
```

We have two options:

1. Tell users who still want to use oauth2client to use `google.auth._oauth2client` to convert their credentials.
2. Add code in our `Client` base class to automatically convert credentials and issue a warning.

I strongly prefer (1) so that we do not prolong oauth2client's lifetime and so that we can more quickly free ourselves to switch http libraries.

jonparrott:
Resolves #2814

### DIFF
--- a/core/google/cloud/client.py
+++ b/core/google/cloud/client.py
@@ -14,12 +14,20 @@
 
 """Base classes for client used to interact with Google Cloud APIs."""
 
+import google.auth.credentials
 from google.oauth2 import service_account
 import six
 
 from google.cloud._helpers import _determine_default_project
 from google.cloud._http import Connection
 from google.cloud.credentials import get_credentials
+
+
+_GOOGLE_AUTH_CREDENTIALS_HELP = (
+    'This library only supports credentials from google-auth-library-python. '
+    'See https://google-cloud-python.readthedocs.io/en/latest/'
+    'google-cloud-auth.html for help on authentication with this library.'
+)
 
 
 class _ClientFactoryMixin(object):
@@ -83,6 +91,10 @@ class Client(_ClientFactoryMixin):
     _connection_class = Connection
 
     def __init__(self, credentials=None, http=None):
+        if (credentials is not None and
+                not isinstance(
+                    credentials, google.auth.credentials.Credentials)):
+            raise ValueError(_GOOGLE_AUTH_CREDENTIALS_HELP)
         if credentials is None and http is None:
             credentials = get_credentials()
         self._connection = self._connection_class(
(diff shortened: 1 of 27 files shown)


jonparrott:
@dhermes PTAL - core only changes right now. Once you're cool with those I'll start fixing the unit tests for all other packages.

core/google/cloud/client.py
```diff
@@ -22,6 +23,13 @@
 from google.cloud.credentials import get_credentials
 
 
+_GOOGLE_AUTH_CREDENTIALS_HELP = (
+    'This library only supports credentials from google-auth-library-python. '
```
dhermes: Nit: Aren't newlines good for this text? Wouldn't it be better to use a triple-quoted / multiline string?

core/google/cloud/client.py
```diff
@@ -83,6 +91,10 @@ class Client(_ClientFactoryMixin):
     _connection_class = Connection
 
     def __init__(self, credentials=None, http=None):
+        if (credentials is not None and
+                not isinstance(
+                    credentials, google.auth.credentials.Credentials)):
```
dhermes: Gah this indenting hurts my eyes!

jonparrott: Got an alternative recommendation?

jonparrott: You should know how I feel about import aliasing.

dhermes: I feel `as` if it is the best choice here

daspecster: helper function anyone? `_check_credentials_type`

dhermes:
Seems CircleCI has given up on pull requests, I'd guess yesterday's outage b0rked something

PR #2831, opened 2016-12-07
Converting remaining datastore snippets to doctests

dhermes:
It's worth noting that I delegated clean-up to `system_tests/datastore.py::tearDownModule`. This is because (in my experimenting) I discovered that the `sphinx.ext.doctest` directive `testcleanup` doesn't run when the thing being tested fails. (FWIW, I think it'd be fairly easy for one of us to send a PR to the Sphinx project and fix that behavior, but that was outside the scope of this PR.)

----

To make dev a little faster I did the following:

```
$ tox -e system-tests --recreate --notest
$ source .tox/system-tests/bin/activate
$ GOOGLE_CLOUD_DISABLE_GRPC=true \
> .tox/system-tests/bin/py.test system_tests/datastore.py \
> -k TestDoctest -s -vv
```

and every time I made a change I just re-installed into the env

```
$ .tox/system-tests/bin/pip install --upgrade datastore/
```

Might be worth turning this into a script someday? But I don't want to design it before we know how we'll use it. Just putting it here for posterity.

----

There is a ``code-block`` snippet in `datastore._http` but that module is non-public so I left it alone.

----

This package could use some more snippets, but for now at least they are all covered.

### DIFF
--- a/datastore/google/cloud/datastore/__init__.py
+++ b/datastore/google/cloud/datastore/__init__.py
@@ -18,17 +18,17 @@ You'll typically use these to get started with the API:
 
 .. doctest:: constructors
 
-  >>> from google.cloud import datastore
-  >>>
-  >>> client = datastore.Client()
-  >>> key = client.key('EntityKind', 1234)
-  >>> key
-  <Key('EntityKind', 1234), project=...>
-  >>> entity = datastore.Entity(key)
-  >>> entity['answer'] = 42
-  >>> entity
-  <Entity('EntityKind', 1234) {'answer': 42}>
-  >>> query = client.query(kind='EntityKind')
+   >>> from google.cloud import datastore
+   >>>
+   >>> client = datastore.Client()
+   >>> key = client.key('EntityKind', 1234)
+   >>> key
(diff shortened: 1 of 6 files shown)


daspecster:
I like it!

dhermes:
Blerg! Python 2 and 3 strike again:
https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/182425545#L2280

```
Failed example:
    client.get(key)
Expected:
    <Entity(u'EntityKind', 1234L) {u'property': 'value'}>
Got:
    <Entity('EntityKind', 1234) {'property': 'value'}>
```

Proposed short-term "fix": Only run snippet tests in Python 2? @jonparrott and @tseaver any better ideas?

tseaver:
@dhermes I hate any testing dependency on repr, period (and tracebacks even worse).

dhermes:
@tseaver Yes I'm aware of your strong opinion. Any construction suggestion?

tseaver:
@dhermes Either change the example not to use the repr, or change the repr not to use the "dict literal" repr of the keys / values.

PR #2833, opened 2016-12-07
Update reference to bounds.

daspecster:
Towards #2753 

Old references to `bounding_poly` -> `bounds`.

### DIFF
--- a/docs/vision-usage.rst
+++ b/docs/vision-usage.rst
@@ -169,9 +169,9 @@ will also provide their known geographical locations if available.
     -33.857123
     >>> landmarks[0].locations[0].longitude
     151.213921
-    >>> landmarks[0].bounding_poly.vertices[0].x_coordinate
+    >>> landmarks[0].bounds.vertices[0].x_coordinate
     78
-    >>> landmarks[0].bounding_poly.vertices[0].y_coordinate
+    >>> landmarks[0].bounds.vertices[0].y_coordinate
     162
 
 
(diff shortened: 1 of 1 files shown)


daspecster:
Cancelled CI since it's just a docs change.

docs/vision-usage.rst
```diff
@@ -169,9 +169,9 @@ will also provide their known geographical locations if available.
     -33.857123
     >>> landmarks[0].locations[0].longitude
     151.213921
-    >>> landmarks[0].bounding_poly.vertices[0].x_coordinate
+    >>> landmarks[0].bounds.vertices[0].x_coordinate
     78
-    >>> landmarks[0].bounding_poly.vertices[0].y_coordinate
+    >>> landmarks[0].bounds.vertices[0].y_coordinate
```
tseaver: I was confused here that these were instances of `face.Landmark`, which have no `bounds` attribute:  but these are `entity.EntityAnnotation` instances, right?  Is it just that the API reuses the term confusingly, and we follow suit?

daspecster: `Landmark` actually uses [`Position`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/5d6148e645a38691eedcd78ae20fd011aa001bb6/vision/google/cloud/vision/geometry.py#L109) which is a 3D point.  [`Bounds`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/5d6148e645a38691eedcd78ae20fd011aa001bb6/vision/google/cloud/vision/geometry.py#L18) is actually a container for a list of points which make a path that encompasses the boundaries of the detected object or feature.

It is complex, and if you have ideas to make it simpler I can make an issue and I'm more than happy to implement ideas!

daspecster: @tseaver I updated #2753 to mention this.

daspecster:
Yeah, I'm trying to clean those up as I go.

PR #2836, opened 2016-12-07
Add Vision detect_text() system tests.

daspecster:
Towards #2753

```bash
test_detect_text_content (vision.TestVisionClientText) ... ok
test_detect_text_filename (vision.TestVisionClientText) ... ok
test_detect_text_gcs (vision.TestVisionClientText) ... ok

Ran 18 tests in 24.979s

OK
________________________________________________________ summary ________________________________________________________
  congratulations :)
```


system_tests/vision.py
```diff
@@ -392,3 +393,64 @@ def test_detect_safe_search_filename(self):
         self.assertEqual(len(safe_searches), 1)
         safe_search = safe_searches[0]
         self._assert_safe_search(safe_search)
+
+
+class TestVisionClientText(unittest.TestCase):
+    DESCRIPTIONS = (
+        'Do',
```
tseaver: Why are the component words here? AFAICT, you only need to test against the full phrase.

daspecster: I could limit the maxResults to 1 and do that but the API returns each word in order to provide it's position in the image. For instance if you have a sign or logo with stylized text in different locations, you might want to know if one of the words was in the right place.

daspecster:
Conflict resolved!

PR #2837, opened 2016-12-07
Adding helpers to push to PyPI on tags.

dhermes:
See #2810 for discussion of which tags are used.

`twine upload` will "just work" since we have set the ``TWINE_USERNAME`` and ``TWINE_PASSWORD`` env. vars on CircleCI.

### DIFF
--- a/scripts/circleci_twine_upload.sh
+++ b/scripts/circleci_twine_upload.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


scripts/circleci_twine_upload.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# H/T: http://stackoverflow.com/a/246128/1068170
+SCRIPT="$(dirname "${BASH_SOURCE[0]}")/circleci_tagged_pkg.py"
+# Determine the package directory being deploying on this tag.
+PKG_DIR="$(python ${SCRIPT})"
+
+# Move into the package, build the distribution and upload.
+cd ${PKG_DIR}
+python setup.py sdist bdist_wheel
+twine upload dist/*
```
tseaver: Where are the PyPI credentials being stored / passed?

tseaver: Oops, my bad.  LGTM

PR #2841, opened 2016-12-08
Update language docs to use explicit path to Document.

mckoss:
@dhermes Eliminates accidental conflict on common class name, 'Document'.


daspecster:
@mckoss were these references broken?

The current links should be ok since `.Document` should be referencing, *I believe*, the imported [`Document`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/c4aa9dfec758b55354272ec055dab8179dfa0bb9/language/google/cloud/language/client.py#L20).

mckoss:
Not broken now - just future-proofing.

dhermes:
Thanks @mckoss!

PR #2844, opened 2016-12-08
Explicitly putting 1.0.2rc0 lower bound on grpcio.

dhermes:
Labeled "don't merge" in the hopes that `1.0.2` will be released soon and we don't need to depend on a release candidate. With this release, running e.g.  system test eliminates the previous error (and hang on exit)

```
$ tox -e system-tests -- datastore
...
E1208 10:30:09.454683798   17019 network_status_tracker.c:48] Memory leaked as all network endpoints were not shut down
```

/cc @bjwatson @nathanielmanistaatgoogle

### DIFF
--- a/core/tox.ini
+++ b/core/tox.ini
@@ -4,7 +4,7 @@ envlist =
 
 [testing]
 deps =
-    grpcio
+    grpcio >= 1.0.2rc0
     mock
     pytest
 covercmd =
--- a/bigtable/setup.py
+++ b/bigtable/setup.py
@@ -51,7 +51,7 @@ SETUP_BASE = {
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.21.0, < 0.22dev',
-    'grpcio >= 1.0.0, < 2.0dev',
+    'grpcio >= 1.0.2rc0, < 2.0dev',
 ]
 
 setup(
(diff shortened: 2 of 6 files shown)


core/tox.ini
```diff
@@ -4,7 +4,7 @@ envlist =
 
 [testing]
 deps =
-    grpcio
+    grpcio >= 1.0.2rc0
     mock
```
tseaver: Shouldn't this pin `< 2.0dev` as well?

dhermes: I am :-1: on upper bounds in **testing**. It's good to have a canary **somewhere** that tells us new versions break us.

daspecster: Changing the version would have to be a deliberate thing anyway and the CI would have to pass before a release.

Pinning this now would keep us from worrying about the future `latest grpcio` being flakey. If grpcio does become flakey down the road we could pin this at that time.

jonparrott: I agree with Danny, typically I only specify minimum bounds on testing dependencies.

dhermes:
I deleted the CircleCI webhook, turned off the project and turned it back on, but CircleCI still won't trigger a build. What the heck

daspecster:
@dhermes their support email is sayhi@circleci.com.

dhermes:
@daspecster Already sent an email, thanks though

dhermes:
Now that 1.0.2 is out I'm putting it into #2846

PR #2847, opened 2016-12-08
Make logging unit test file a temp file

### LINKED ISSUE #2843: Logging unit tests create a file name with PID and then don't clean it up.
@waprin Can you fix?

For example

```
$ git st
On branch upgrade-grpcio
nothing to commit, working tree clean
$ tox -e py27
...
 $ git st
On branch upgrade-grpcio
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        14920

nothing added to commit but untracked files present (use "git add" to track)
$ cat 14920 | wc
      0       0       0
```

waprin:
Fixes #2843. This doesn't clean up the file after but it does make it a temp file, @dhermes let me know if you think this is sufficient which I think it is.

### DIFF
--- a/logging/unit_tests/test_client.py
+++ b/logging/unit_tests/test_client.py
@@ -553,6 +553,7 @@ class TestClient(unittest.TestCase):
 
     def test_get_default_handler_app_engine(self):
         import os
+        import tempfile
         from google.cloud._testing import _Monkey
         from google.cloud.logging.client import _APPENGINE_FLEXIBLE_ENV_VM
         from google.cloud.logging.handlers import app_engine as _MUT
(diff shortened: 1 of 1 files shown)


logging/unit_tests/test_client.py
```diff
@@ -562,7 +563,8 @@ def test_get_default_handler_app_engine(self):
                                 credentials=_make_credentials(),
                                 use_gax=False)
 
-        with _Monkey(_MUT, _LOG_PATH_TEMPLATE='{pid}'):
+        temp_log_path = os.path.join(tempfile.mkdtemp(), '{pid}')
```
tseaver: This leaks tempdirs every time the tests are run. :<

daspecster: Well crap...
>The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.
- https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp

Fixing...

waprin: Ya I mentioned that in PR comment, thought it wasn't a big deal cause it's in tmp but I supposed might as well clean it up, ty for fixing.

daspecster: Yeah I saw that but I what I didn't realize was...
>The directory is readable, writable, and searchable only by the creating user ID.

@waprin thanks for pointing it out in your notes and thanks @tseaver for the fix!

PR #2848, opened 2016-12-09
Plug tempdir leak

tseaver:
Addresses:
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2847/files#r91715191

### DIFF
--- a/core/google/cloud/_testing.py
+++ b/core/google/cloud/_testing.py
@@ -54,6 +54,24 @@ class _NamedTemporaryFile(object):
         os.remove(self.name)
 
 
+def _tempdir_maker():
+    import contextlib
+    import shutil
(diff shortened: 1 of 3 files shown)


dhermes:
@tseaver it looks like you just invented a `py.test` fixture

tseaver:
@dhermes it's been around since a lot longer than our switch to `py.test`.

dhermes:
Ha, that's not what I meant, I just meant that's what (user-defined) fixtures usually look like

FYI, this fix doesn't work on Windows:
https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-python/build/1.0.971.master/job/iij8tk8b3l3obob5

PR #2849, opened 2016-12-09
Cleanup left over temp directory.

daspecster:
See: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2847#discussion_r91715191

@tseaver LMK if there's a different way you would like to see this done.

I looked at tempfile.NamedTemporaryFile() since it cleans up on it's own but I'm not sure if there is a way to make that work with the `{pid}` formatting in the file name.


daspecster:
Ha! Just saw yours @tseaver.

Issue #2842, opened 2016-12-08
Track implement the leftover "future" items for Speech


daspecster:
The future is now!

### Future Updates:
- [x] Clearer way to send stream to `Sample.content`. See #2708.  (#2850)
- [ ] Move result access to Operation handler https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2638#discussion_r85596232
- [ ] Add snippet tests (optional for release)

PR #2850, opened 2016-12-09
Make sample.content handle bytes and files.

daspecster:
Towards: #2842

See: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2680#discussion_r86612107

This PR attempts to handle both bytes and file objects passed to `Sample.content`.

I added `Sample.stream` to distinguish between raw bytes and a file like object.
`Sample.stream` cannot be set, and is derived from `Sample._content`, which can be either bytes or a file.

I'm sure you guys will have some great feedback for this. I'm totally open to suggestions.

### DIFF
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -182,9 +182,7 @@ class GAPICSpeechAPI(object):
                        .cloud_speech_pb2.StreamingRecognizeResponse`
         :returns: ``StreamingRecognizeResponse`` instances.
         """
-        if getattr(sample.content, 'closed', None) is None:
-            raise ValueError('Please use file-like object for data stream.')
-        if sample.content.closed:
+        if sample.stream.closed:
             raise ValueError('Stream is closed.')
 
         requests = _stream_requests(sample, language_code=language_code,
@@ -252,7 +250,6 @@ class GAPICSpeechAPI(object):
             language_code=language_code, max_alternatives=max_alternatives,
             profanity_filter=profanity_filter,
             speech_context=SpeechContext(phrases=speech_context))
-
         audio = RecognitionAudio(content=sample.content,
                                  uri=sample.source_uri)
         api = self._gapic_api
@@ -337,7 +334,7 @@ def _stream_requests(sample, language_code=None, max_alternatives=None,
     yield config_request
 
     while True:
-        data = sample.content.read(sample.chunk_size)
+        data = sample.stream.read(sample.chunk_size)
         if not data:
             break
         yield StreamingRecognizeRequest(audio_content=data)
--- a/speech/google/cloud/speech/client.py
+++ b/speech/google/cloud/speech/client.py
@@ -66,12 +66,12 @@ class Client(BaseClient):
         else:
             self._use_gax = use_gax
 
-    def sample(self, content=None, source_uri=None, encoding=None,
+    def sample(self, content=None, source_uri=None, stream=None, encoding=None,
                sample_rate=None):
         """Factory: construct Sample to use when making recognize requests.
 
         :type content: bytes
-        :param content: (Optional) Byte stream of audio.
+        :param content: (Optional) Bytes containing audio data.
 
         :type source_uri: str
         :param source_uri: (Optional) URI that points to a file that contains
@@ -79,6 +79,9 @@ class Client(BaseClient):
                            Currently, only Google Cloud Storage URIs are
                            supported, which must be specified in the following
                            format: ``gs://bucket_name/object_name``.
+
+        :type stream: file
+        :param stream: (Optional) File like object to stream.
 
         :type encoding: str
         :param encoding: encoding of audio data sent in all RecognitionAudio
@@ -97,7 +100,7 @@ class Client(BaseClient):
         :rtype: :class:`~google.cloud.speech.sample.Sample`
         :returns: Instance of ``Sample``.
         """
-        return Sample(content=content, source_uri=source_uri,
+        return Sample(content=content, source_uri=source_uri, stream=stream,
                       encoding=encoding, sample_rate=sample_rate, client=self)
 
     @property
--- a/speech/google/cloud/speech/sample.py
+++ b/speech/google/cloud/speech/sample.py
@@ -14,6 +14,7 @@
 
 """Sample class to handle content for Google Cloud Speech API."""
 
+
 from google.cloud.speech.encoding import Encoding
 from google.cloud.speech.result import StreamingSpeechResult
 
@@ -22,7 +23,7 @@ class Sample(object):
     """Representation of an audio sample to be used with Google Speech API.
 
     :type content: bytes
-    :param content: (Optional) Byte stream of audio.
+    :param content: (Optional) Bytes containing audio data.
 
     :type source_uri: str
     :param source_uri: (Optional) URI that points to a file that contains
@@ -30,6 +31,9 @@ class Sample(object):
                        Currently, only Google Cloud Storage URIs are
                        supported, which must be specified in the following
                        format: ``gs://bucket_name/object_name``.
+
+    :type stream: file
+    :param stream: (Optional) File like object to stream.
 
     :type encoding: str
     :param encoding: encoding of audio data sent in all RecognitionAudio
@@ -51,17 +55,19 @@ class Sample(object):
     default_encoding = Encoding.FLAC
     default_sample_rate = 16000
 
-    def __init__(self, content=None, source_uri=None,
+    def __init__(self, content=None, source_uri=None, stream=None,
                  encoding=None, sample_rate=None, client=None):
         self._client = client
 
-        no_source = content is None and source_uri is None
-        both_source = content is not None and source_uri is not None
-        if no_source or both_source:
-            raise ValueError('Supply one of \'content\' or \'source_uri\'')
+        sources = [content is not None, source_uri is not None,
+                   stream is not None]
+        if sources.count(True) != 1:
+            raise ValueError('Supply exactly one of '
+                             '\'content\',  \'source_uri\', \'stream\'')
 
         self._content = content
         self._source_uri = source_uri
+        self._stream = stream
 
         if sample_rate is not None and not 8000 <= sample_rate <= 48000:
             raise ValueError('The value of sample_rate must be between 8000'
@@ -108,6 +114,15 @@ class Sample(object):
         :returns: Integer between 8000 and 48,000.
         """
         return self._sample_rate
+
+    @property
+    def stream(self):
+        """Stream the content when it is a file-like object.
+
+        :rtype: file
+        :returns: File like object to stream.
+        """
+        return self._stream
 
     @property
     def encoding(self):
--- a/speech/unit_tests/test_client.py
+++ b/speech/unit_tests/test_client.py
@@ -72,7 +72,7 @@ class TestClient(unittest.TestCase):
     SAMPLE_RATE = 16000
     HINTS = ['hi']
     AUDIO_SOURCE_URI = 'gs://sample-bucket/sample-recording.flac'
-    AUDIO_CONTENT = '/9j/4QNURXhpZgAASUkq'
+    AUDIO_CONTENT = b'testing 1 2 3'
 
     @staticmethod
     def _get_target_class():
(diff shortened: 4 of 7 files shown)


daspecster:
I should have fixed this quicker.  But I think I have at least a working path that we can discuss now.

LMKWYT!

speech/google/cloud/speech/_gax.py
```diff
@@ -99,8 +99,7 @@ def async_recognize(self, sample, language_code=None,
             profanity_filter=profanity_filter,
             speech_context=SpeechContext(phrases=speech_context))
 
-        audio = RecognitionAudio(content=sample.content,
-                                 uri=sample.source_uri)
+        audio = RecognitionAudio(content=sample.content, uri=sample.source_uri)
```
dhermes: This change has nothing to do with the PR?

speech/google/cloud/speech/_gax.py
```diff
@@ -252,9 +249,7 @@ def sync_recognize(self, sample, language_code=None, max_alternatives=None,
             language_code=language_code, max_alternatives=max_alternatives,
             profanity_filter=profanity_filter,
             speech_context=SpeechContext(phrases=speech_context))
-
-        audio = RecognitionAudio(content=sample.content,
-                                 uri=sample.source_uri)
+        audio = RecognitionAudio(content=sample.content, uri=sample.source_uri)
```
dhermes: Ditto here: this change is just cosmetic?

speech/google/cloud/speech/client.py
```diff
@@ -67,8 +67,8 @@ def sample(self, content=None, source_uri=None, encoding=None,
                sample_rate=None):
         """Factory: construct Sample to use when making recognize requests.
 
-        :type content: bytes
-        :param content: (Optional) Byte stream of audio.
+        :type content: bytes or file
+        :param content: (Optional) Byte stream of audio or file like object.
```
dhermes: The "steam" in "Byte stream" already sounds like it's a file-like object. Instead: "Bytes object containing audio or file-like object which yields bytes."

speech/google/cloud/speech/sample.py
```diff
@@ -14,15 +14,19 @@
...
 from google.cloud.speech.result import StreamingSpeechResult
 
 
 class Sample(object):
     """Representation of an audio sample to be used with Google Speech API.
 
-    :type content: bytes
-    :param content: (Optional) Byte stream of audio.
+    :type content: bytes or file
+    :param content: (Optional) Byte stream of audio or file like object.
```
dhermes: Ditto here on the description

speech/google/cloud/speech/sample.py
```diff
@@ -14,15 +14,19 @@
...
 from google.cloud.speech.encoding import Encoding
 from google.cloud.speech.result import StreamingSpeechResult
 
 
 class Sample(object):
     """Representation of an audio sample to be used with Google Speech API.
 
-    :type content: bytes
-    :param content: (Optional) Byte stream of audio.
+    :type content: bytes or file
```
dhermes: Is there a reason to allow `content` to be two different types? Is it a large drag on users to just have two different input params?

@tseaver @jonparrott WDYT?

speech/google/cloud/speech/sample.py
```diff
@@ -95,10 +99,15 @@ def source_uri(self):
...
+        :rtype: bytes or file
+        :returns: Byte stream of audio content or file like object.
         """
-        return self._content
+        if self._content is None:
+            return self._content
+        content = self._content
+        if getattr(self._content, 'read', None) is not None:
+            content = self._content.read()
+        return content
```
dhermes: You could shorten this up a lot. The entire body of this method / property can be

```python
try:
    return self._content.read()
except AttributeError:
    return self._content
```

speech/google/cloud/speech/sample.py
```diff
@@ -95,10 +99,15 @@ def source_uri(self):
...
-        :returns: Byte stream of audio content.
+        :rtype: bytes or file
+        :returns: Byte stream of audio content or file like object.
         """
-        return self._content
+        if self._content is None:
+            return self._content
+        content = self._content
+        if getattr(self._content, 'read', None) is not None:
+            content = self._content.read()
```
dhermes: This method will consume the "entire" stream without going back to the beginning. That's not always bad, but here, you don't make any "note" of that on the object, so callers have no real way of knowing if they are getting the whole thing.

We need a better strategy here. A "simple" answer might be to just over-write the file-like object with the consumed content returned by `.read()`, though that doesn't fit your target examples of a stream from a microphone (since there is no such thing as "all" or "the end").

speech/google/cloud/speech/sample.py
```diff
@@ -110,6 +119,18 @@ def sample_rate(self):
...
     @property
+    def stream(self):
+        """Stream the content when it is a file-like object.
+
+        :rtype: file
+        :returns: File like object to stream.
+        """
+        stream = self._content
+        if stream is not None and getattr(stream, 'read', None) is None:
+            stream = BytesIO(_to_bytes(stream))
```
dhermes: What is wrong with returning `None` when there is no attached stream?

speech/unit_tests/test_client.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import unittest
-
```
dhermes: This empty line "belongs" here because `unittest` is in the stdlib and `mock` is not (`unittest.mock` **is**, but only in Python 3)

speech/unit_tests/test_sample.py
```diff
@@ -42,6 +42,39 @@ def test_content_and_source_uri(self):
...
 
+    def test_content_to_stream(self):
+        from google.cloud.speech.encoding import Encoding
+
+        content = 'testing 1 2 3 4'
+        sample = self._make_one(content=content,
+                                encoding=Encoding.FLAC,
+                                sample_rate=self.SAMPLE_RATE)
+
+        self.assertIsNotNone(getattr(sample.stream, 'read', None))
```
dhermes: Or:

```python
with self.assertRaises(AttributeError):
    getattr(sample.stream, 'read')
```

speech/unit_tests/test_sample.py
```diff
@@ -42,6 +42,39 @@ def test_content_and_source_uri(self):
...
+        from google.cloud.speech.encoding import Encoding
+
+        test_bytes = b'testing 1 2 3 4'
+        content = BytesIO(test_bytes)
+        sample = self._make_one(content=content,
+                                encoding=Encoding.FLAC,
+                                sample_rate=self.SAMPLE_RATE)
+        self.assertEqual(sample.content, test_bytes)
+
+    def test_bytes_converts_to_file_like_object(self):
```
dhermes: I don't understand the name of this test. Did you mean to check that `.stream` is a `BytesIO`?

speech/google/cloud/speech/sample.py
daspecster: So don't convert `content`?

```python
if getattr(self._content, 'read', None) is None:
    return None
return self._content
```

speech/unit_tests/test_sample.py
daspecster: So depending on how the `stream` property ends up working, this could be removed.

I had setup `stream` to check if `content` was set and then just convert that to a file-like object through `BytesIO`

speech/google/cloud/speech/sample.py
daspecster: Based on the comment below and this one, maybe I shouldn't try and convert `content` to a stream if it's not already one?

speech/unit_tests/test_sample.py
daspecster: Now it just returns `None` if it's not a file like object.
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2850/commits/bd7130537a3ea852593966997245c6d9dd0235a3#diff-37554bc08e0770c52c2125924d18ae12R126

speech/google/cloud/speech/sample.py
daspecster: @tseaver @jonparrott bump! :)

jonparrott: audio sample should just be bytes.

daspecster: @jonparrott, so `Sample` should have three ways to pass in data. `Sample.content`, `Sample.stream` and `Sample.source_uri`?

SGTM, I'll update the PR.

speech/google/cloud/speech/_gax.py
```diff
@@ -182,9 +182,7 @@ def streaming_recognize(self, sample, language_code=None,
                        .cloud_speech_pb2.StreamingRecognizeResponse`
         :returns: ``StreamingRecognizeResponse`` instances.
         """
-        if getattr(sample.content, 'closed', None) is None:
-            raise ValueError('Please use file-like object for data stream.')
-        if sample.content.closed:
+        if sample.stream is None or sample.stream.closed:
             raise ValueError('Stream is closed.')
```
dhermes: Now this error should be different, i.e. if the stream is `None`

speech/google/cloud/speech/sample.py
```diff
@@ -51,17 +55,21 @@ class Sample(object):
...
+    def __init__(self, content=None, source_uri=None, stream=None,
                  encoding=None, sample_rate=None, client=None):
         self._client = client
 
-        no_source = content is None and source_uri is None
-        both_source = content is not None and source_uri is not None
+        no_source = content is None and source_uri is None and stream is None
+        both_source = (content is not None
+                       and source_uri is not None
+                       and stream is not None)
```
dhermes: This isn't what you want. You are checking `000` and `111` among the 8 possible null/non-null triples. You are accidentally allowing `011`, `101` and `110`. Instead:

```python
non_null = (int(content is not None) +
            int(source_uri is not None) +
            int(stream is not None))
if non_null != 1:
    raise ValueError('...')
```

Also "both" would no longer apply here.

(ISTM your unit tests that test those situations should fail, so maybe I am misunderstanding something.)

speech/google/cloud/speech/sample.py
```diff
@@ -22,7 +23,7 @@ class Sample(object):
     """Representation of an audio sample to be used with Google Speech API.
 
     :type content: bytes
-    :param content: (Optional) Byte stream of audio.
+    :param content: (Optional) Bytes object containing audio data.
```
dhermes: Lose the "object" (here and elsewhere)?

speech/google/cloud/speech/sample.py
```diff
@@ -110,6 +118,17 @@ def sample_rate(self):
...
     @property
+    def stream(self):
+        """Stream the content when it is a file-like object.
+
+        :rtype: file
+        :returns: File like object to stream.
+        """
+        if getattr(self._stream, 'read', None) is None:
+            return None
+        return self._stream
```
dhermes: ISTM we can safely just return `self._stream`? "Trust" that the user passed a file-like. @jonparrott WDYT?

speech/google/cloud/speech/sample.py
```diff
@@ -51,17 +55,21 @@ class Sample(object):
...
 
-        no_source = content is None and source_uri is None
-        both_source = content is not None and source_uri is not None
+        no_source = content is None and source_uri is None and stream is None
+        both_source = (content is not None
+                       and source_uri is not None
+                       and stream is not None)
         if no_source or both_source:
-            raise ValueError('Supply one of \'content\' or \'source_uri\'')
+            raise ValueError('Supply one of '
```
dhermes: "Supply **exactly** one of"

speech/google/cloud/speech/sample.py
daspecster: That's weird that the tests pass...I'll check that out.

WDYT of...
```python
one_source = [content is not None, source_uri is not None,
              stream is not None].count(True)
if one_source != 1:
    raise ValueError('Supply exactly one of '
                     '\'content\',  \'source_uri\', \'stream\'')
```

speech/google/cloud/speech/sample.py
```diff
@@ -51,17 +55,19 @@ class Sample(object):
...
                  encoding=None, sample_rate=None, client=None):
         self._client = client
 
-        no_source = content is None and source_uri is None
-        both_source = content is not None and source_uri is not None
-        if no_source or both_source:
-            raise ValueError('Supply one of \'content\' or \'source_uri\'')
+        sources = [content is not None, source_uri is not None,
+                   stream is not None]
+        if sources.count(True) != 1:
```
dhermes: @daspecster My assumption was that creating the list is slower. However, I was wrong:

```
$ python -m timeit "A = B = None; C = 1; N = [A is not None, B is not None, C is not None].count(True)"
10000000 loops, best of 3: 0.182 usec per loop
$ python -m timeit "A = B = None; C = 1; N = int(A is not None) + int(B is not None) + int(C is not None)"
1000000 loops, best of 3: 0.306 usec per loop
```

daspecster: Interesting!

daspecster:
Squashed! Will merge per offline discussion once CI passes.

PR #2851, opened 2016-12-10
Force close logging handler.

tseaver:
Allow tempdir cleanup to work as expected on Windows.


dhermes:
Something very wrong with Python 3:

- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/549
- https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-python/build/1.0.992.master

Happened after this merged, but I don't think it was caused by this PR

daspecster:
@dhermes yeah I've been looking into this but I haven't found the cause other than what's in the trace.

Let's chat about it here https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2856.

PR #2853, opened 2016-12-10
Fixing regex for package deployment.

dhermes:
@tseaver I have a feeling you could craft better regexes than me here if you feel like it.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,6 +24,7 @@ general:
 
 deployment:
   release:
+    # See "scripts/circleci_tagged_pkg.py" for info on REGEX
     tag: /(([a-z]+)-)*([0-9]+)\.([0-9]+)\.([0-9]+)/
     owner: GoogleCloudPlatform
     commands:
--- a/circle.yml
+++ b/circle.yml
@@ -24,7 +24,7 @@ general:
 
 deployment:
   release:
-    tag: /(([a-z]+)-)?([0-9]+)\.([0-9]+)\.([0-9]+)/
+    tag: /(([a-z]+)-)*([0-9]+)\.([0-9]+)\.([0-9]+)/
     owner: GoogleCloudPlatform
     commands:
       - pip install --upgrade twine
--- a/scripts/circleci_tagged_pkg.py
+++ b/scripts/circleci_tagged_pkg.py
@@ -24,11 +24,16 @@ import re
 import sys
 
 
-RE_TXT = r'^(?P<pkg>(([a-z]+)-)*)([0-9]+)\.([0-9]+)\.([0-9]+)$'
-TAG_RE = re.compile(RE_TXT)
+TAG_RE = re.compile(r"""
+    ^
+    (?P<pkg>
+        (([a-z]+)-)*)            # pkg-name-with-hyphens- (empty allowed)
+    ([0-9]+)\.([0-9]+)\.([0-9]+)  # Version x.y.z (x, y, z all ints)
+    $
+""", re.VERBOSE)
 TAG_ENV = 'CIRCLE_TAG'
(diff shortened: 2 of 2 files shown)


dhermes:
For a bit of sanity / checking that this works:

```
$ CIRCLE_TAG=error-reporting-0.22.0 python scripts/circleci_tagged_pkg.py
/home/dhermes/da_googz/worktree_regex/error_reporting
$ CIRCLE_TAG=0.22.0 python scripts/circleci_tagged_pkg.py
/home/dhermes/da_googz/worktree_regex/
$ CIRCLE_TAG=lots-of-hyphens-0.22.0 python scripts/circleci_tagged_pkg.py
/home/dhermes/da_googz/worktree_regex/lots_of_hyphens
$ CIRCLE_TAG=onehyphen-0.22.0 python scripts/circleci_tagged_pkg.py
/home/dhermes/da_googz/worktree_regex/onehyphen
```

circle.yml
```diff
@@ -24,7 +24,7 @@ general:
 
 deployment:
   release:
-    tag: /(([a-z]+)-)?([0-9]+)\.([0-9]+)\.([0-9]+)/
+    tag: /(([a-z]+)-)*([0-9]+)\.([0-9]+)\.([0-9]+)/
```
tseaver: Probably don't want to make the first alpha token optional (unless I misunderstand something).  At a minimum, we should be documenting the different kinds of tags from which we expect to build releases.

dhermes: - Some expected tags are in my [2nd comment][1]. The reason we allow it to be empty is because umbrella tags are expected to be `0.x.y` (discussed in #2810)
- Where did you imagine the documenting happens? Here or in another file?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2853#issuecomment-266242369

scripts/circleci_tagged_pkg.py
```diff
@@ -24,7 +24,7 @@
 import sys
 
 
-RE_TXT = r'^((?P<pkg>[a-z]+)-)?([0-9]+)\.([0-9]+)\.([0-9]+)$'
+RE_TXT = r'^(?P<pkg>(([a-z]+)-)*)([0-9]+)\.([0-9]+)\.([0-9]+)$'
```
tseaver: Could maybe use `re.VERBOSE` form with comments on each part of the pattern?

circle.yml
tseaver: > Some expected tags are in my 2nd comment. The reason we allow it to be empty is because umbrella tags are expected to be 0.x.y (discussed in #2810)

OK, I missed that this could be applied to the umbrella tags.  What does the script produce when the `CIRCLE_TAG` variable is set to an umbrella tag?

> Where did you imagine the documenting happens? Here or in another file?

dhermes: > What does the script produce when the `CIRCLE_TAG` variable is set to an umbrella tag?

From [referred to comment][1]

```
$ CIRCLE_TAG=0.22.0 python scripts/circleci_tagged_pkg.py
/home/dhermes/da_googz/worktree_regex/
```

In [the source][2] it just returns the `git` root without trying to append a subdir

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2853#issuecomment-266242369
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/b004bd43ecc0d2eb80367afc9615e135a95726d5/scripts/circleci_tagged_pkg.py#L52-L53

scripts/circleci_tagged_pkg.py
dhermes: OK SGTM

dhermes:
@tseaver PTAL

tseaver:
LGTM

PR #2860, opened 2016-12-13
Add a timestamp parameter to the logger.

Fkawala:
The timestamp parameter is useful when one sends batches of log records. In this case, every log record in a batch will have the same timestamp. The timestamp equals the date of insertion in the google cloud logging API. 

If one uses the `asctime` attribute from the [LogRecord](https://docs.python.org/2/library/logging.html#logging.LogRecord) instances as timestamp value, a log record would have a propre date.

### DIFF
--- a/logging/google/cloud/logging/logger.py
+++ b/logging/google/cloud/logging/logger.py
@@ -17,6 +17,7 @@
 import json
 
 from google.protobuf.json_format import MessageToJson
+from google.cloud._helpers import _datetime_to_rfc3339
 
 
 class Logger(object):
@@ -120,7 +121,8 @@ class Logger(object):
         :type http_request: dict
         :param http_request: (optional) info about HTTP request associated with
                              the entry
-        :type timestamp: str
+
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
 
         :rtype: dict
@@ -158,7 +160,7 @@ class Logger(object):
             resource['httpRequest'] = http_request
 
         if timestamp is not None:
-            resource['timestamp'] = timestamp
+            resource['timestamp'] = _datetime_to_rfc3339(timestamp)
 
         return resource
 
@@ -190,7 +192,7 @@ class Logger(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         client = self._require_client(client)
@@ -227,7 +229,7 @@ class Logger(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         client = self._require_client(client)
@@ -264,7 +266,7 @@ class Logger(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         client = self._require_client(client)
@@ -373,7 +375,7 @@ class Batch(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         self.entries.append(
@@ -399,7 +401,7 @@ class Batch(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         self.entries.append(
@@ -425,7 +427,7 @@ class Batch(object):
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
-        :type timestamp: str
+        :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
         """
         self.entries.append(
--- a/logging/google/cloud/logging/logger.py
+++ b/logging/google/cloud/logging/logger.py
@@ -455,8 +455,8 @@ class Batch(object):
             kwargs['labels'] = self.logger.labels
 
         entries = []
-        for entry_type, entry, labels, iid, severity, http_req, timestamp in \
-                self.entries:
+        for (entry_type, entry, labels, iid, severity, http_req,
+             timestamp) in self.entries:
             if entry_type == 'text':
                 info = {'textPayload': entry}
             elif entry_type == 'struct':
--- a/logging/google/cloud/logging/logger.py
+++ b/logging/google/cloud/logging/logger.py
@@ -379,7 +379,8 @@ class Batch(object):
         :param timestamp: (optional) timestamp of event being logged.
         """
         self.entries.append(
-            ('text', text, labels, insert_id, severity, http_request, timestamp))
+            ('text', text, labels, insert_id, severity, http_request,
+             timestamp))
 
     def log_struct(self, info, labels=None, insert_id=None, severity=None,
                    http_request=None, timestamp=None):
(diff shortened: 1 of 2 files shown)


Fkawala:
I signed it!

daspecster:
@Fkawala, I noticed that timestamp is always `str` in the docs. Do you think it would actually be a `datetime.datetime` at some point?

There are helper functions in [`google.cloud.core._helpers`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/aea151b83893656e43d563935de8ce93a46f310a/core/google/cloud/_helpers.py#L239) that might make things easier too.

tseaver:
@Fkawala, @daspecster The `timestamp` paramter to `google.cloud.logging.entries._BaseEntry` is expected to be an instance of `datetime.datetime`.  It would be best if the equivalent parameter for `Logger.log_*` was also an instance of `datetime.datetime`.

dhermes:
@waprin PTAL

Fkawala:
@daspecster @tseaver I'll update the PR to use `datetime.datime` instances instead of `str`. I have a question though. Where should we convert the `datetime.datime` to RFC3339 UTC "Zulu" format ([docs](https://cloud.google.com/logging/docs/api/reference/rest/v2/LogEntry))? It could be done in the [write_entry](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/115748814bf6a59d99c13e070dfdcfbdec5e865d/logging/google/cloud/logging/_http.py#L131) method. This would look like:

```python
def write_entries(self, entries, logger_name=None, resource=None, labels=None):
        data = {'entries': [_prepare_entry(entry) for entry in entries]}
    
        if logger_name is not None:
            data['logName'] = logger_name
    
        if resource is not None:
            data['resource'] = resource
    
        if labels is not None:
            data['labels'] = labels
    
        self._connection.api_request(method='POST', path='/entries:write',
                                     data=data)
```
With `_prepare_entry` a function that would rely on the [_datetime_to_rfc3339](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/aea151b83893656e43d563935de8ce93a46f310a/core/google/cloud/_helpers.py#L311) function to convert timestamp into RFC3339.

Would you have a more elegant solution ?

daspecster:
@Fkawala I think you should convert it as close to receiving it as possible.

@tseaver in thinking about this more, should the `log_*` methods support both `str` timestamps and `datetime.datetime` and just properly convert if needed?

Otherwise the example would be something like
```python
from datetime import datetime

logger.log_text("A simple entry", timestamp=datetime.utcnow())
```
Or something similar to that.

Fkawala:
@daspecster Can I get closer than I did in 7ac9c0a ?

daspecster:
@Fkawala if a datetime object gets passed to the `log_*` methods then I think it's good. If a `str` ends up getting accepted in the `log_*` methods then we may want to convert them from `str`->`datetime` right in those methods.

As I mentioned above, this depends on what we want our usage to look like.

tseaver:
@Fkawala The `_make_entry_resource` helper method is the right place to convert from `datetime.datetime` to a string:  its job is to convert "native" Python objects to their JSON equivalents before putting the entry's repr on the wire.

Fkawala:
@daspecster ok got it, thanks for the explanation. I'll update the PR If the string compatibility is needed for the `log_*` methods. 

Is there something I could do to for the PR to pass the CI tests ?

daspecster:
@Fkawala I think the issue with the Python 3 tests has been resolved. I'll restart the build and see if they clear up.

waprin:
lgtm

logging/unit_tests/test_logger.py
```diff
@@ -484,7 +485,7 @@ def test_log_struct_defaults(self):
         batch = self._make_one(logger, client=client)
         batch.log_struct(STRUCT)
         self.assertEqual(batch.entries,
-                         [('struct', STRUCT, None, None, None, None)])
+                         [('struct', STRUCT, None, None, None, None, None)])
```
tseaver: Too long:  best bet would be to wrap everything, e.g.:

```python
        self.assertEqual(
            batch.entries,
            [('struct', STRUCT, None, None, None, None, None)])
```

logging/unit_tests/test_logger.py
```diff
@@ -499,13 +500,14 @@ def test_log_struct_explicit(self):
...
             'status': STATUS,
         }
+        TIMESTAMP = '2016-10-12T15:01:23.045123456Z'
         client = _Client(project=self.PROJECT, connection=_make_credentials())
         logger = _Logger()
         batch = self._make_one(logger, client=client)
         batch.log_struct(STRUCT, labels=LABELS, insert_id=IID,
-                         severity=SEVERITY, http_request=REQUEST)
+                         severity=SEVERITY, http_request=REQUEST, timestamp=TIMESTAMP)
         self.assertEqual(batch.entries,
```
tseaver: Too long:  wrap to next line.

logging/unit_tests/test_logger.py
```diff
@@ -499,13 +500,14 @@ def test_log_struct_explicit(self):
...
-                         [('struct', STRUCT, LABELS, IID, SEVERITY, REQUEST)])
+                         [('struct', STRUCT, LABELS, IID, SEVERITY, REQUEST, TIMESTAMP)])
```
tseaver: Too long:  wrap to next line.

logging/unit_tests/test_logger.py
```diff
@@ -531,13 +533,14 @@ def test_log_proto_explicit(self):
             'requestUrl': URI,
             'status': STATUS,
         }
+        TIMESTAMP = '2016-10-12T15:01:23.045123456Z'
         client = _Client(project=self.PROJECT, connection=_make_credentials())
         logger = _Logger()
         batch = self._make_one(logger, client=client)
         batch.log_proto(message, labels=LABELS, insert_id=IID,
-                        severity=SEVERITY, http_request=REQUEST)
+                        severity=SEVERITY, http_request=REQUEST, timestamp=TIMESTAMP)
```
tseaver: Too long:  wrap to next line.

logging/unit_tests/test_logger.py
```diff
@@ -531,13 +533,14 @@ def test_log_proto_explicit(self):
...
         self.assertEqual(batch.entries,
-                         [('proto', message, LABELS, IID, SEVERITY, REQUEST)])
+                         [('proto', message, LABELS, IID, SEVERITY, REQUEST, TIMESTAMP)])
```
tseaver: Too long:  wrap to next line.

daspecster:
@Fkawala now we're cooking with Ci!

You're missing some test coverage in `logger.py`.
https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/183920499#L378

And the issues that @tseaver mentioned can be seen here
https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/183920499#L423-L432

You can run the tests, lint and coverage commands as mentioned in the [`CONTRIBUTING.rst`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/CONTRIBUTING.rst).

```bash
$ tox -e py27 -- logging
$ tox -e py35 -- logging
$ tox -e cover -- logging
$ tox -e lint
```

Fkawala:
@daspecster Thanks for the explanation! 
@tseaver I think that I addressed all your points.

logging/google/cloud/logging/logger.py
```diff
@@ -452,7 +455,8 @@ def commit(self, client=None):
             kwargs['labels'] = self.logger.labels
 
         entries = []
-        for entry_type, entry, labels, iid, severity, http_req, timestamp in self.entries:
+        for entry_type, entry, labels, iid, severity, http_req, timestamp in \
+                self.entries:
```
tseaver: I hate to nitpick, but we really try hard to avoid backslash continuations.  Can you wrap the unpack-variables in parentheses instead?  E.g:

```python
    for (entry_type, entry, labels, iid, severity, http_req,
         timestamp) in self.entries:
```

tseaver:
@Fkawala Thank you for the patch!

Fkawala:
@tseaver I'm happy to contribute, you do truly great work with here, thanks!

Fkawala:
@tseaver, I don't know about the release patterns for this project, when should I expect a release for google.cloud.logging ?

dhermes:
Releasing is pretty easy for us, though we probably won't do it until the new year.

tseaver:
@dhermes do we need to release `core` as well, due to the connection refactoring in #2870 / #2875?

dhermes:
@tseaver That's pretty likely. Though I want to break off some more chunks before doing a release (e.g. make `Connection.api_request` a function rather than an instance method)

Fkawala:
Hello guys, first thing first happy new year. I don't mean to be pushy, but it would be great to have this released, is there something I can do ?

Fkawala:
Hi, any update on the expected release of #2860 ?

Issue #2863, opened 2016-12-14
Branding updates required for Translate API client libraries


danoscarmike:
On November 15, 2016, Google launch a rebranding of Google Translate API to Google Cloud Translation API. This does not impact the method names but the URLbase as well as readmes, samples and comments.

URLbase has been already updated to translation.googleapis.com

We need to have update the Client library, the readme, client library docs, samples and code comments to use "Google Cloud Translation API" instead of Google Translate API. The short version of the new name is "Cloud Translation API" or "Translation API"

Note: the product Google Translate still exists. It is the API Google Translate API that has been re-branded.

daspecster:
Thanks @danoscarmike! This will get resolved in #2865.

PR #2865, opened 2016-12-14
Update Translation API branding.

daspecster:
Fixes: #2863

### DIFF
--- a/translate/setup.py
+++ b/translate/setup.py
@@ -56,7 +56,7 @@ REQUIREMENTS = [
 setup(
     name='google-cloud-translate',
     version='0.22.0',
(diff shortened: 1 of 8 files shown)


translate/setup.py
```diff
@@ -56,7 +56,7 @@
 setup(
     name='google-cloud-translate',
```
dhermes: So the actual PyPI package name is good to remain unchanged @danoscarmike?

bondareko: ATLAS_TOKEN='$ATLAS_TOKEN'

bondareko:
unnamed

PR #2870, opened 2016-12-15
Remove connection from base Client class

dhermes:
This is part of #2606. The idea behind this PR: we can drop support for `Connection` classes on a package-by-package basis rather than all at once.

/cc @tseaver and @mckoss this will have teeny implications on downstream clients

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -58,20 +58,25 @@ class Client(JSONClient):
                     passed when creating a dataset / job.  If not passed,
                     falls back to the default inferred from the environment.
 
-    :type credentials: :class:`oauth2client.client.OAuth2Credentials` or
-                       :class:`NoneType`
-    :param credentials: The OAuth2 Credentials to use for the connection
-                        owned by this client. If not passed (and if no ``http``
-                        object is passed), falls back to the default inferred
-                        from the environment.
+    :type credentials: :class:`~google.auth.credentials.Credentials`
+    :param credentials: (Optional) The OAuth2 Credentials to use for this
+                        client. If not passed (and if no ``http`` object is
+                        passed), falls back to the default inferred from the
+                        environment.
 
-    :type http: :class:`httplib2.Http` or class that defines ``request()``.
-    :param http: An optional HTTP object to make requests. If not passed, an
+    :type http: :class:`~httplib2.Http`
+    :param http: (Optional) HTTP object to make requests. Can be any object
+                 that defines ``request()`` with the same interface as
+                 :meth:`~httplib2.Http.request`. If not passed, an
                  ``http`` object is created that is bound to the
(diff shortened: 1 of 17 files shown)


core/google/cloud/client.py
```diff
@@ -72,33 +71,32 @@ def from_service_account_json(cls, json_credentials_path, *args, **kwargs):
...
         if (credentials is not None and
                 not isinstance(
                     credentials, google.auth.credentials.Credentials)):
             raise ValueError(_GOOGLE_AUTH_CREDENTIALS_HELP)
         if credentials is None and http is None:
             credentials = get_credentials()
-        self._connection = self._connection_class(
-            credentials=credentials, http=http)
+        self._credentials = credentials
+        self._http = http
```
daspecster: What if `http` is None? Where/how does the default for that get set?

dhermes: In the `Connection` base class: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/8e258d5d50c909d1125d58e44a82a86f148c4f91/core/google/cloud/_http.py#L96-L108

It's worth noting: this code does exactly what it did before the change, but now the subclasses call the `Connection` constructor instead of the base class (and also the base class stores two new attributes `_credentials` and `_http`)

daspecster:
I would assume you would want @tseaver to take a look too, but this looks pretty straight forward and LGTM.

dhermes:
@daspecster Yup I'll wait for @tseaver. Thanks

datastore/google/cloud/datastore/client.py
```diff
@@ -157,26 +157,29 @@ class Client(_BaseClient, _ClientProjectMixin):
...
+                 that defines ``request()`` with the same interface as
+                 :meth:`~httplib2.Http.request`. If not passed, an
                  ``http`` object is created that is bound to the
                  ``credentials`` for the current object.
     """
-    _connection_class = Connection
 
     def __init__(self, project=None, namespace=None,
                  credentials=None, http=None):
         _ClientProjectMixin.__init__(self, project=project)
```
tseaver: Not an issue with this PR, but a  mix-in whose ctor messes up `super` is a code smell.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -58,20 +58,25 @@ class Client(JSONClient):
...
+                        passed), falls back to the default inferred from the
+                        environment.
+
+    :type http: :class:`~httplib2.Http`
+    :param http: (Optional) HTTP object to make requests. Can be any object
+                 that defines ``request()`` with the same interface as
+                 :meth:`~httplib2.Http.request`. If not passed, an
                  ``http`` object is created that is bound to the
                  ``credentials`` for the current object.
     """
```
dhermes: @jonparrott has made a `copy_docstring` decorator in `google-auth`

PR #2874, opened 2016-12-15
Removing most (direct) connection usage in Pub / Sub

dhermes:
~~**NOTE**: Has #2870 as diffbase.~~

### DIFF
--- a/pubsub/unit_tests/test__gax.py
+++ b/pubsub/unit_tests/test__gax.py
@@ -900,9 +900,9 @@ class Test_SubscriberAPI(_Base, unittest.TestCase):
 @unittest.skipUnless(_HAVE_GAX, 'No gax-python')
 class Test_make_gax_publisher_api(_Base, unittest.TestCase):
 
-    def _call_fut(self, connection):
+    def _call_fut(self, *args, **kwargs):
         from google.cloud.pubsub._gax import make_gax_publisher_api
-        return make_gax_publisher_api(connection)
+        return make_gax_publisher_api(*args, **kwargs)
 
     def test_live_api(self):
         from google.cloud.pubsub._gax import DEFAULT_USER_AGENT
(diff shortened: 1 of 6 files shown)


pubsub/unit_tests/test__gax.py
```diff
@@ -953,13 +951,12 @@ def mock_insecure_channel(host):
...
 
         host = 'CURR_HOST:1234'
-        connection = _Connection(in_emulator=True, host=host)
         patch = mock.patch.multiple(
             'google.cloud.pubsub._gax',
             PublisherClient=mock_publisher_api,
             insecure_channel=mock_insecure_channel)
         with patch:
-            result = self._call_fut(connection)
+            result = self._call_fut(None, host=host, secure=False)
```
tseaver: Would making `credentials` optional and using its absence to signal insecure clarify?  Essentially it would become "pass exactly one of `credentials` or `host`."

tseaver:
@dhermes I'm OK to merge without the change I suggested, if you don't think it would be an improvement.

pubsub/unit_tests/test__gax.py
dhermes: I'm fine doing that, though it makes the calling code a bit "worse":

```python
make(self.creds, host=self.conn.host, secure=not self.conn.in_emul)
```

becomes

```python
kwargs = {'host': self.conn.host}
if not self.conn.in_emul:
    kwargs['creds'] = self.creds
make(**kwargs)
```

dhermes:
@tseaver WDYT of my most recent comment [above][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2874#discussion_r92695114

tseaver:
@dhermes It looks to me as though `host` unused in the non-emulated case.  If so, then the call site becomes:

```python
if self.conn.in_emul:
    channel = make_foo(host=emulator_host)
else:
    channel = make_foo(credentials=credentials)
```

dhermes:
@tseaver PTAL

tseaver:
LGTM

PR #2876, opened 2016-12-15
Using `gapic-google-cloud-datastore-v1` instead of `datastore._generated`

dhermes:
This doesn't fully switch over to using the GAPIC client, but at least gets rid of **all** references to the generated code.

/cc @bjwatson


Makefile.datastore
```diff
@@ -1,53 +0,0 @@
...
-	GRPCIO_VIRTUALENV="$(GRPCIO_VIRTUALENV)" \
-	GENERATED_SUBDIR=$(GENERATED_SUBDIR) \
-		python scripts/make_datastore_grpc.py
-	# Rewrite the imports in the generated *_pb2.py files.
-	python scripts/rewrite_imports.py $(DATASTORE_DIR)/*pb2.py
-
-clean:
-	rm -fr $(GENERATED_DIR)
-
-.PHONY: generate clean
```
tseaver: W00t!

scripts/make_datastore_grpc.py
```diff
@@ -1,151 +0,0 @@
...
-        file_obj.write('# BEGIN: Imports from datastore_pb2\n')
-        for name in get_pb2_message_types():
-            import_line = IMPORT_TEMPLATE % (name,)
-            file_obj.write(import_line)
-        file_obj.write('#   END: Imports from datastore_pb2\n')
-        file_obj.write(''.join(grpc_only_lines))
-
-
-if __name__ == '__main__':
-    main()
```
tseaver: Double w00t!

PR #2879, opened 2016-12-16
Do not attempt to acknowledge 0 messages

### LINKED ISSUE #2878: If AutoAck pull returns empty then exception raised on context exit
If the subscription pull returns 0 messages in the __enter__ method in the AutoAck class then on __exit__ the acknowledge call raises an exception on the HTTP call: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/pubsub-0.22.0/pubsub/google/cloud/pubsub/subscription.py#L512

Currently the way a deal with this is to wrap the context manager in try/except or override __exit__ to check for an empty list. The former is messy and the latter can just be fixed at source.

quom:
Closes #2878

### DIFF
--- a/pubsub/google/cloud/pubsub/subscription.py
+++ b/pubsub/google/cloud/pubsub/subscription.py
@@ -509,4 +509,5 @@ class AutoAck(dict):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self._subscription.acknowledge(list(self), self._client)
+        if self:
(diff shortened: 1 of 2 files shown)


quom:
I signed it!

daspecster:
@quom I tried to change the title of the PR to be more descriptive but I may have missed the mark. If you have better wording feel free to update!

quom:
Updated it a bit - should have done something similar first time

tseaver:
@quom Thanks very much for the patch!

Issue #2882, opened 2016-12-18
BigQuery: list_jobs() iterator encounters error when calling classmethod from_api_repr()


sean185:
**Disclaimer:** I am new to opening issues so please guide me along on what I can do to improve, thanks.

1. Ubuntu 14.04

2. Python 2.7.6

3. google-cloud-python version
```
google-api-python-client==1.5.5
google-auth==0.4.0
google-auth-httplib2==0.0.2
google-cloud-bigquery==0.22.0
google-cloud-core==0.22.0
googleapis-common-protos==1.5.0
```

4. Stacktrace if available
```
Traceback (most recent call last):
  File "pybq_test.py", line 103, in <module>
    main()
  File "pybq_test.py", line 69, in main
    for i,job in enumerate(jobs):
  File "/home/dev-sean/BQTesting/local/lib/python2.7/site-packages/google/cloud/iterator.py", line 211, in _items_iter
    for item in page:
  File "/home/dev-sean/BQTesting/local/lib/python2.7/site-packages/google/cloud/iterator.py", line 155, in next
    result = self._item_to_value(self._parent, item)
  File "/home/dev-sean/BQTesting/local/lib/python2.7/site-packages/google/cloud/bigquery/client.py", line 382, in _item_to_job
    return iterator.client.job_from_resource(resource)
  File "/home/dev-sean/BQTesting/local/lib/python2.7/site-packages/google/cloud/bigquery/client.py", line 165, in job_from_resource
    return CopyJob.from_api_repr(resource, self)
  File "/home/dev-sean/BQTesting/local/lib/python2.7/site-packages/google/cloud/bigquery/job.py", line 745, in from_api_repr
    for source_config in config['sourceTables']:
KeyError: 'sourceTables'
```

5. Steps to reproduce
- instantiate the list_jobs iterator with parameter all_users=True 
- iterate through the jobs
- as it iterates the error is produced when converting response from_api_repr

6. Code example
```
jobs = bqClient.list_jobs(all_users=True)
for i,job in enumerate(jobs):
    print job.name
```

tseaver:
@sean185 Thanks for the report!  Based on the traceback, the job being constructed is a "copy" job:  does that seem plausible to you?

One thing you could do to help identify the problem would be to set a breakpoint at the failure, and report the contents of the `resource` dict.  E.g.:

```python
jobs = bqClient.list_jobs(all_users=True)
try:
    jobs = list(jobs)
[...]

PR #2884, opened 2016-12-19
Accept copy jobs with single 'sourceTable' config.

tseaver:
Such jobs would be created via another client:  we map that configuration onto a sequence of tables containing only the one item.

Closes: #2882.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -744,7 +744,11 @@ class CopyJob(_AsyncJob):
         sources = []
         source_configs = config.get('sourceTables')
         if source_configs is None:
-            source_configs = [config['sourceTable']]
+            single = config.get('sourceTable')
+            if single is None:
+                raise KeyError(
+                    "Resource missing 'sourceTables' / 'sourceTable'")
+            source_configs = [single]
         for source_config in source_configs:
             dataset = Dataset(source_config['datasetId'], client)
             sources.append(Table(source_config['tableId'], dataset))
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -742,7 +742,10 @@ def from_api_repr(cls, resource, client):
         dataset = Dataset(dest_config['datasetId'], client)
         destination = Table(dest_config['tableId'], dataset)
         sources = []
-        for source_config in config['sourceTables']:
+        source_configs = config.get('sourceTables')
+        if source_configs is None:
+            source_configs = [config['sourceTable']]
```
dhermes: Aren't we trading one accidental `KeyError` for another here?

tseaver: Server docs say "pick one" between `sourceTable` and `sourceTables`:  one cannot create a "copy" job without specifying one of them.

dhermes: Our `KeyError` before was due to misunderstanding this? What's to say we don't misunderstand here as well?

dhermes: That is to say, why don't we throw our own error rather than an opaque `KeyError`, with an explanation of why our interpretation of the docs say the server should never return values in that state

daspecster: Could re-raise `KeyError` but add a message about "No source tables provided.".

daspecster: Oh I think I see now. We default to [`sourceTables`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/a8fb99736bee614ad6d646853c003db15b941d50/bigquery/google/cloud/bigquery/job.py#L707) but if someone creates the job some other way then it *could* be `sourceTable`.

And yeah there shouldn't be any jobs coming back that are broken with neither set.

tseaver: @dhermes f003e0f adds an explicit error message to the `KeyError` being raised.

dhermes:
LGTM

Issue #2885, opened 2016-12-19
Add deprecation notice to old pypi page for gcloud


kir-titievsky:
Would you be so kind as to add a deprecation notice on the gcloud homepage on PyPi: https://pypi.python.org/pypi/gcloud/0.8.0, pointing to the google-cloud-libs page?  The page currently offers no clues about how to upgrade to the next version of the code  since the package name changed,  I had a couple of fairly sophisticated user stuck at a year old version of gcloud, running into problems, with no evident path forward.

daspecster:
This kind of ties in with #2748 as well.

daspecster:
Sounds like this is actually a duplicate.

dhermes:
@kir-titievsky Though a duplicate, it's definitely important to resolve. Thanks for poking us

PR #2889, opened 2016-12-20
Prep gcloud-0.18.4 release.

tseaver:
Identical to 0.18.3, except adding warnings to the PyPI page and the docs that the package has been renamed.

Closes: #2748, #2885.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,12 @@
 Google Cloud Python Client
 ==========================
 
-    Python idiomatic client for `Google Cloud Platform`_ services.
+.. warning::
+
+   This library has been renamed ``google-cloud``.  Please update
+   to use the new version.  https://pypi.python.org/pypi/google-cloud
+
+Python idiomatic client for `Google Cloud Platform`_ services.
 
 .. _Google Cloud Platform: https://cloud.google.com/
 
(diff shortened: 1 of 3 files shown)


daspecster:
Should the `classifier` change to `Development Status :: 7 - Inactive`?

Also, should there be a warning printed on install like @jonparrott was [talking](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2748#issuecomment-268077470) about?

dhermes:
@tseaver I agree with @daspecster about the classifier. See my [comment][1], i.e. we don't need to make a new release just to update the `PKG-INFO`. That's not to say we shouldn't, but I don't think a new release is necessary.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2748#issuecomment-268398162

tseaver:
Emitting a `DeprecationWarning` isn't likely to be much help:  they are not displayed by default.

dhermes:
I've started toying with the PKG-INFO for `0.18.3`. So far I added a `warning` admonition (supported as part of ReST but apparently not by PyPI) and removed the badges that don't point to anything anymore (since our build URLs changed too)

Issue #2886, opened 2016-12-19
BigQuery: timestamps not populated correctly for parameterized queries


tswast:
-   Stacktrace

```
$ python sync_query_params.py timestamp
Traceback (most recent call last):
  File "sync_query_params.py", line 176, in <module>
    sync_query_timestamp_params()
  File "sync_query_params.py", line 124, in sync_query_timestamp_params
    query_results.run()
  File "/Users/swast/src/google-cloud-python/bigquery/google/cloud/bigquery/query.py", line 360, in run
    method='POST', path=path, data=self._build_resource())
  File "/Users/swast/venvs/bq-dml/lib/python3.5/site-packages/google/cloud/_http.py", line 335, in api_request
    error_info=method + ' ' + url)
google.cloud.exceptions.BadRequest: 400 Unparseable query parameter `ts_value` in type `TYPE_TIMESTAMP`, Invalid timestamp: '1.4810976E9' value: '1.4810976E9' (POST https://www.googleapis.com/bigquery/v2/projects/cloud-samples-tests/queries)
```

-   Steps to reproduce

Attempt to use `TIMESTAMP` type in a scalar query parameter.

-   Code example

https://github.com/GoogleCloudPlatform/python-docs-samples/commit/a788ea1829949a4f90d3ce42aa188ce4087f4906

The timestamp type should not be sent as a floating point value. It should take this format: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp-type Note that this is different from RFC 3339. There is an issue open at https://code.google.com/p/google-bigquery/issues/detail?id=842 for accepting RFC 3339 in TIMESTAMP query parameters.

tseaver:
@tswast I have passing system tests which show the back-end sending `TIMESTAMP` values back as floating-point values.  Are you saying that the back-end expects those values to be sent as string timestamps, in spite of how it returns them?  For instance, here is the result from a query including a timestamp value:

```python
{'_client': <google.cloud.bigquery.client.Client object at 0x7f399dfb33d0>,
 '_configuration': <google.cloud.bigquery.query._SyncQueryConfiguration object at 0x7f399e250090>,
 '_job': None,
 '_properties': {u'cacheHit': True,
                 u'jobComplete': True,
                 u'jobReference': {u'jobId': u'job_N75jt8A_drBKYhdW9gbvuImxAWQ',
                                   u'projectId': u'<MY-PROJECT-ID>'},
[...]

PR #2891, opened 2016-12-20
Send TIMESTAMP query parameters as strings.

tseaver:
- *Not* the float-time-since-epoch-in-seconds which Bigquery uses for all other `TIMESTAMP` values. :(
- *Not* RFC3339, but the SQL-mandated format with an embedded space. :(

Closes: #2886.

/cc @tswast

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -18,10 +18,10 @@ import base64
 from collections import OrderedDict
 import datetime
 
+from google.cloud._helpers import UTC
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -150,7 +150,11 @@ def _bytes_to_json(value):
 def _timestamp_to_json(value):
```
daspecster: I know this is only used in BigQuery right now, but since it's kind of BigQuery specific, should the method name be something like `_bq_timestamp_to_json`?

tseaver: If we promote it to `core`, maybe it needs a more descriptive name (probably something about the actual format, rather than the API).  As it is, the fact that it is in `bigquery/_helpers.py` should be enough.

Issue #2887, opened 2016-12-20
BigQuery: structs not populated correctly in query parameters


tswast:
See documentation for structs in query parameters:
https://cloud.google.com/bigquery/querying-data#using_structs_in_parameterized_queries

- Stacktrace

```
$ python sync_query_params.py struct 7 hello
Traceback (most recent call last):
  File "sync_query_params.py", line 204, in <module>
    sync_query_struct_params(args.x, args.y)
  File "sync_query_params.py", line 139, in sync_query_struct_params
    query_results.run()
  File "/Users/swast/src/google-cloud-python/bigquery/google/cloud/bigquery/query.py", line 360, in run
    method='POST', path=path, data=self._build_resource())
  File "/Users/swast/venvs/bq-dml/lib/python3.5/site-packages/google/cloud/_http.py", line 335, in api_request
    error_info=method + ' ' + url)
google.cloud.exceptions.BadRequest: 400 Invalid query parameter type (POST https://www.googleapis.com/bigquery/v2/projects/cloud-samples-tests/queries)
```

- Code sample

https://github.com/GoogleCloudPlatform/python-docs-samples/blob/tswast-bq/bigquery/cloud-client/sync_query_params.py#L129

The google-cloud-bigquery library produces a resource that looks like:

```
{
  'name': 'struct_value',
  'parameterValue': {
    'structValues': {'y': 'hello', 'x': '7'}
  },
  'parameterType': {
    'structTypes': [{'type': 'INT64', 'name': 'x'}, {'type': 'STRING', 'name': 'y'}]
  }
}
```

The API expects a resource that looks like:

```
{
      "name": "struct_value",
      "parameterType": {
        "type": "STRUCT",
        "structTypes": [
          {
            "name": "x",
            "type": {
              "type": "INT64"
            }
          },
          {
            "name": "y",
            "type": {
              "type": "STRING"
            }
          }
        ]
      },
      "parameterValue": {
        "structValues": {
          "x": {
            "value": "1"
          },
          "y": {
            "value": "foo"
          }
        }
      }
    }
```

Notice the missing `{ "value": "my-value" }` in the `structValues` in the request object produced by this library.

PR #2892, opened 2016-12-20
Conform to expected wire format for struct query parameters.

tseaver:
Closes: #2887.

/cc @tswast

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -501,6 +501,7 @@ class ArrayQueryParameter(AbstractQueryParameter):
             values = [converter(value) for value in values]
         resource = {
             'parameterType': {
+                'type': 'ARRAY',
                 'arrayType': self.array_type,
             },
             'parameterValue': {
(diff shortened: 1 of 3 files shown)


bigquery/unit_tests/test__helpers.py
```diff
@@ -1188,13 +1196,17 @@ def test_from_api_repr_w_name(self):
     def test_from_api_repr_wo_name(self):
         RESOURCE = {
             'parameterType': {
+                'type': 'STRUTCT',
```
tswast: Typo: STRUCT not STRUTCT

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -554,10 +555,12 @@ def from_api_repr(cls, resource):
         instance = cls(name)
         types = instance.struct_types
         for item in resource['parameterType']['structTypes']:
-            types[item['name']] = item['type']
+            types[item['name']] = item['type']['type']
```
dhermes: `type`->`type` for realz?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -568,18 +571,19 @@ def to_api_repr(self):
         :returns: JSON mapping
         """
         types = [
-            {'name': key, 'type': value}
+            {'name': key, 'type': {'type': value}}
```
dhermes: Ditto on `type`->`type`

bigquery/google/cloud/bigquery/_helpers.py
tswast: Yeah, for realz. https://cloud.google.com/bigquery/querying-data#bigquery-query-params-structs-api

It's because the type could also be an array or another struct, in which case it would need an `arrayType` or `structTypes`, too.

tswast: I suspect this client library might have problems if someone tries to put an array or struct into a struct, but I won't block this PR for that. (I didn't write a sample for that use case.)

Issue #2687, opened 2016-11-04
Add Error Reporting gRPC


waprin:
This is a tracking work item to get the Error Reporting GAPIC into this library:

https://pypi.python.org/pypi/gapic-google-cloud-error-reporting-v1beta1

@steren and I agree that the existing (small) `report` and `report_exception` is what we want to keep, along with possible more framework integrations in the future. So I think what we primarily want is the same thing as with done with Logging where we add the GAPIC as a dependency, and then fallback to the existing code if gRPC is disabled.

One small question is if and to what extent we want to expose the rest of the GAPIC besides events.report via this library. There are a few API calls that will likely only be used by power users, but it might be worth having some sort of pass-through mechanism. Unless I'm mistake, it looks like so far we either wrap all the GAPIC calls or pass-through completely, but we haven't done both, so not sure exactly how this would look.

cc @bjwatson @jgeewax @omaray in case you have comments.

bjwatson:
@jskeet IIRC, you have experience doing thin and/or partial hand-written layers on the C# GAPIC code. To date, we have done thick hand-written layers for Python. Do you have any advice for how best to manage thinner layers?

cc @anthmgoogle

jskeet:
Hmm... it's likely to be language-specific. In C#, we can add extra methods/properties/etc to existing classes using `partial`... I don't know what the equivalent is in Python. But aside from anything else, we've already decided that all the GAPIC code *is* exposed in C#; is that true in Python?

If you're happy to expose the GAPIC code for power users, you could have a very simple wrapper that exposes simple operations but also exposes a property to access all the GAPIC methods if necessary.

bjwatson:
Thanks @jskeet. Makes sense how you're leveraging the C# `partial` feature.

Since Python is package delivery (i.e. the GAPIC code has its own package on https://pypi.python.org/pypi), the GAPIC code is always exposed to our users.

waprin:
Ok, now that I am thinking about it, if it's a dependency of this library it will be exposed after installing this library, that makes sense.

I am leaving for Brazil tomorrow and will be busy with GCP Next Sao Paulo but I can loop back on this later in the month.

bjwatson:
Yep, the other google-cloud-python libraries that depend on GAPIC call it out as a dependency in their `setup.py`.

steren:
[...]

PR #2894, opened 2016-12-21
Add Error Reporting GAPIC as dependency

waprin:
Step 1 to resolving #2687. See that discussion thread for the other followups.

@bjwatson @dhermes

### DIFF
--- a/error_reporting/google/cloud/error_reporting/_gax.py
+++ b/error_reporting/google/cloud/error_reporting/_gax.py
@@ -1,0 +1,56 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GAX wrapper for Error Reporting API requests."""
+
+from google.cloud.gapic.errorreporting.v1beta1 import report_errors_service_api
+from google.devtools.clouderrorreporting.v1beta1 import report_errors_service_pb2
+from google.protobuf.json_format import ParseDict
+
+
+def make_report_error_api(project):
+    """Create an instance of the GAX Logging API."""
+    api = report_errors_service_api.ReportErrorsServiceApi()
+    return _ErrorReportingGaxApi(api, project)
+
+
+class _ErrorReportingGaxApi(object):
+    """Helper mapping Error Reporting-related APIs
+
+    :type gax_api:
+        :class:`google.cloud.gapic.errorreporting.v1beta1
+        .report_errors_service_api.report_errors_service_api`
+    :param gax_api: API object used to make GAX requests.
+    """
+    def __init__(self, gax_api, project):
+        self._gax_api = gax_api
+        self._project = project
+
+    def report_error_event(self, error_report):
+        """Uses the GAX client to report the error.
+
+        :type project: str
+        :param: project: Project ID to report the error to
+
+        :type error: dict:
+        :param: error: dict payload of the error report formatted
+                       according to
+                       https://cloud.google.com/error-reporting/docs/formatting-error-messages
+                       This object should be built using
+                       Use :meth:~`google.cloud.error_reporting.client._build_error_report`
+        """
+        project_name = self._gax_api.project_path(self._project)
+        error_report_payload = report_errors_service_pb2.ReportedErrorEvent()
+        ParseDict(error_report, error_report_payload)
+        self._gax_api.report_error_event(project_name, error_report_payload)
--- a/error_reporting/google/cloud/error_reporting/_gax.py
+++ b/error_reporting/google/cloud/error_reporting/_gax.py
@@ -28,10 +28,10 @@ def make_report_error_api(client):
     """Create an instance of the GAX Logging API.
 
     :type client::class:`google.cloud.error_reporting.Client`
-    :param client: Error Reporting client
+    :param client: Error Reporting client.
 
     :rtype: :class:_ErrorReportingGaxApi
-    :returns: An Error Reporting API instane
+    :returns: An Error Reporting API instance.
     """
     channel = make_secure_channel(
         client._connection.credentials,
--- a/error_reporting/google/cloud/error_reporting/_gax.py
+++ b/error_reporting/google/cloud/error_reporting/_gax.py
@@ -14,6 +14,9 @@
 
 """GAX wrapper for Error Reporting API requests."""
 
+from google.cloud._helpers import make_secure_channel
+from google.cloud._http import DEFAULT_USER_AGENT
+
 from google.cloud.gapic.errorreporting.v1beta1 import (
     report_errors_service_client)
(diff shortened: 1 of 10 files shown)


daspecster:
@waprin I forgot, can you assign people and add labels to PRs?

waprin:
I can't. Though my teammates could change that for me, up to you guys if you want us to.

bjwatson:
@waprin I gave you one-off write access to this repo. We can refactor this as a Github team if there are other Stackdriver devs we should include (especially if you need access to multiple google-cloud-* repos).

FYI @jgeewax @daspecster @dhermes @tseaver @omaray @danoscarmike

bjwatson:
LGTM, but I'll let @tseaver or @dhermes sign off.

dhermes:
This PR is pointless unless the package actually uses the depended upon package?

waprin:
@dhermes the next incoming PR will actually use this client, but it's still not pointless, the idea is that we keep the `pip install` experience consistent whether it's handwritten or a GAPIC. You install google-cloud-* to get the clients, regardless of if the are hand-written, GAPICs, or a mix. So there will be a bunch of APIs that just have a GAPIC we just install the dependency here for. Or in cases like this, a GAPIC and then maybe some hand-written code on top. At least that's how it's been explained to me.

dhermes:
Yes, that's how it's been explained to me as well, but AFAIK the generated libraries are not in the "ready to go" state yet (e.g. no docs).

bjwatson:
@dhermes We'll get on top of publishing GAPIC docs for all APIs in January. We also need to discuss how to integrate them with google-cloud-python docs as discussed in https://github.com/GoogleCloudPlatform/gcloud-common/issues/207.

waprin:
@dhermes @bjwatson @steren this is not remotely ready for thorough review (missing tests, docs, lots of stye issues) but I wanted to throw up the commits to give the rough sketch of adding the GAPIC as a transport for the helper API, make sure it's all LGTM to you guys and then I'll clean up the PR to merge it.

This is basically what Steren and I had discussed a few weeks ago.

* Copies the basic GAX/non-GAX structure of the logging client
* Unlike the logging client, instead of falling back to an HTTP client, it falls back to the logging client itself, now in _logging
* Basic instrumentation/helper library API remains the same, code to build the error report payload is shared

If you all feel good about this I'll finish up the tests/docs/style issues.

Obviously not expecting any response on this until 2017, happy new year! 🎉 🎉 🎈 🎈

waprin:
@dhermes @bjwatson @steren ping for the new year. again, just looking for an initial +1 to basic direction here before I finish tests/docs.

bjwatson:
Thanks @waprin. Looks reasonable to me based on a superficial review.

@lukesneeringer Can you take a more detailed look? This should have a similar structure to the Logging code that you've looked at already.

waprin:
Cool ty, will finish it up then.

steren:
The direction LGTM.

Will this also allow me to call the

waprin:
@bjwatson  @lukesneeringer cool ty, I wanted to ask, do we need the secure channel stuff? Is there a high level overview of that?

One other problem I've run into is that Error Reporting depends on gax 0.14 but logging (which Error Reporting uses as the fallback) depends on 0.15, and `ssl_creds` changed to `ssl_credentials.` Wonder if we can just update the error reporting gax dependency?

@steren will try to clarify the gRPC stuff.

Since we are adding the GAPIC to the module, all the autogenned VTK calls will be included as a dependency. Documentation of all those calls is still a TODO as noted above. For now this just adds the GAPIC library to this one and uses it for the handwritten helper functions.

bjwatson:
@waprin I believe we'll need to regenerate the GAPIC code to work with GAX `0.15`. Is that correct @geigerj?

The changes allow us to use @jonparrott's new [google-auth](https://pypi.python.org/pypi/google-auth) library, rather than requiring the broken old oauth2client.

geigerj:
@bjwatson Yes, we should regenerate the GAPIC library to use GAX 0.15.0.

waprin:
@dhermes this is ready for review

error_reporting/google/cloud/error_reporting/client.py
```diff
@@ -12,13 +12,27 @@
...
 # limitations under the License.
 
-"""Client for interacting with the Stackdriver Logging API"""
+"""Client for interacting with the Stackdriver Error Reporting API"""
 
+import os
 import traceback
 
-import google.cloud.logging.client
+try:
```
lukesneeringer: This is slightly nitpicky, but rewrite this to:

```python
try:
    from google.cloud.error_reporting._gax import make_report_error_api
    _HAVE_GAX = True
except ImportError:
    _HAVE_GAX = False
```

error_reporting/google/cloud/error_reporting/_gax.py
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+
+"""GAX wrapper for Error Reporting API requests."""
+
+from google.cloud._helpers import make_secure_channel
+from google.cloud._http import DEFAULT_USER_AGENT
+
+from google.cloud.gapic.errorreporting.v1beta1 import (
+    report_errors_service_client)
+from google.cloud.grpc.devtools.clouderrorreporting.v1beta1 import (
```
lukesneeringer: This is `google.cloud.proto` now.

error_reporting/google/cloud/error_reporting/client.py
lukesneeringer: It also is not entirely clear to me why this will raise ImportError. The `_gax_` module is present, so it has to raise `ImportError` somehow on its import. Which one of its imports is going to raise, and in what situation will it do so?

error_reporting/google/cloud/error_reporting/_gax.py
waprin: When I `pip install gapic-google-cloud-error-reporting-v1beta1` I don't get that `proto` package anywhere. Maybe something hasn't been released or there's another dependency needed?

error_reporting/google/cloud/error_reporting/client.py
waprin: I mostly just copied this structure from `logging/client.py`. I can look into a bit more where exactly the import fails, I assume on one of the grpc imports in `_gax.py`.

waprin: Having trouble finding a good example of an environment besides GAE standard where we have this problem. On GAE standard I get the import error `_args_from_interpreter_flags ` but I think that has to do general packaging weirdness on standard. Not sure best way to get development environment besides standard that I can't install gRPC on. Maybe we can get in touch with original author?

dhermes: @lukesneeringer Though it's unclear how many layers deep one needs to go to find it, the assumption is that `grpcio` is included transitively. This **can't** work in pure Python only environments (like GAE).

No need to explain beyond that?

error_reporting/google/cloud/error_reporting/_gax.py
lukesneeringer: The former.

We should have the new package out early next week; alternatively, just be careful to version pin. :-)

error_reporting/google/cloud/error_reporting/client.py
lukesneeringer: That is fine -- thanks.

error_reporting/google/cloud/error_reporting/_gax.py
lukesneeringer: This is released now.

waprin: Wow 1 minute too late, I'll do new PR to update.

lukesneeringer: Hahaha, no big deal. :-p

PR #2898, opened 2016-12-22
Adding support for Standard SQL views

thejens:
Hi, I need to be able to create view using Standard SQL. 

I added it as a separate attribute on Table, but I guess the Table could also be restructured so that view_query and view_use_standard_sql are broken out to a separate "View Definition" class or similar.

If you don't agree with this solution, please help me out as this is becoming critical functionality as more and more use moves to the new dialect.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -363,7 +363,43 @@ class Table(object):
         """
         if not isinstance(value, six.string_types):
             raise ValueError("Pass a string")
-        self._properties['view'] = {'query': value}
+        if 'view' in self._properties:
+            self._properties['view']['query'] = value
+        else:
+            self._properties['view'] = {
+                'query': value,
+                'useLegacySql': True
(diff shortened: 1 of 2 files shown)


dhermes:
@tseaver I don't quite feel comfortable enough here to assess the feature added

tseaver:
@dhermes we already support defining a table as a view by setting its `view_query` property.  Options for exposing the `useLegacySql` flag for that query seem to be:

1. Add a new property only for the flag (this PR implements this option).
2. Add a new property (maybe `standard_view_query`) which takes the same query string but updates the underlying flag to `False`.  Its getter would either raise or return `None` if the view was assigned as legacy.
3. Adapt the current property setter to accept a tuple `(query, use_legacy)`, as well as the bare string (ugly).  I'm not sure what the getter should do.
4. Use a method rather than an property to assign the query and flag (breaks user code).

I think I favor #2.

dhermes:
Option 2 SGTM

PR #2900, opened 2016-12-27
Add a starter class for _gax.

daspecster:
Towards #2753

TL;DR This PR just adds the beginning of the `_GAPICVisionAPI` class and `Client` branching.

I started just doing the straight up conversion but it got huge.
Then I tried just converting face annotations and that got huge.

So this is part 1 of many, hopefully small PRs to add this in.

There are a lot of classes that will need the `from_pb` added to them and also test splitting to support both since the JSON responses vary in the value names a lot from the proto responses.

So for starters this was the smallest chunk I could come up with. Next up I will be adding the `annotate` method which requires some interesting helpers to translate `Feature` and `Image`s.

### DIFF
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -1,0 +1,25 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


vision/google/cloud/vision/_gax.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class _GAPICVisionAPI(object):
+    """Vision API for interacting with the gRPC version of Vision.
+
+    :type client: :class:`~google.cloud.core.client.Client`
+    :param client: Instance of ``Client`` object.
+    """
+    def __init__(self, client=None):
+        raise NotImplementedError
```
dhermes: This is OK to check in, but a do-nothing constructor would've been fine too (i.e. the `NotImplementedError` not needed as much for the constructor as for instance methods)

PR #2901, opened 2016-12-27
Fix array handling in parameterized queries

tswast:
Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2862


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -483,8 +483,11 @@ def from_api_repr(cls, resource):
         :returns: instance
         """
         name = resource.get('name')
-        array_type = resource['parameterType']['arrayType']
-        values = resource['parameterValue']['arrayValues']
+        array_type = resource['parameterType']['arrayType']['type']
```
dhermes: The triply-nested access is very scary

tswast: Is there a better way to do it? Technically it should be recursive since an array could contain a struct, but I didn't want to modify the design too much in this PR. (Structs have a similar problem since they can contain arrays or more structs, but we don't handle that case yet)

tswast:
Yeah, +1 to defining a high-level before jumping into the implementation. Sorry about that! That's on me and the BigQuery eng team to do, and we did a poor job at it.

dhermes:
> That's on me and the BigQuery eng team to do, and we did a poor job at it.

It's not unique to BigQuery. All services could benefit from a simple set of "must pass" requests and DevRel could play a role in defining those. I suppose having an emulator helps that but some things (i.e. a datastore) are easier to emulate than others.

bigquery/google/cloud/bigquery/_helpers.py
dhermes: No better way to do it, I just worry about missing keys and opaque error messages to users:

```python
>>> d = {}
>>> d['a']['b']['c']
KeyError: 'a'
>>> d['a'] = da = {}
>>> d['a']['b']['c']
KeyError: 'b'
>>> da['b'] = db = {}
>>> d['a']['b']['c']
KeyError: 'c'
>>> db['c'] = 'sheesh'
>>> d['a']['b']['c']
'sheesh'
```

tseaver:
@tswast Thanks for digging in.

PR #2909, opened 2016-12-29
Fix for pubsub scope error.

### LINKED ISSUE #2908: gRPC pubsub missing scope error
It appears that something changed in gRPC.

```
Exception google.auth.exceptions.RefreshError: RefreshError('invalid_scope: Empty or missing scope not allowed.', u'{\n  "error" : "invalid_scope",\n  "error_description" : "Empty or missing scope not allowed."\n}') in 'grpc._cython.cygrpc.plugin_get_metadata' ignored
```

Reference: https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/187483537#L2101

Running `$ tox -e system-tests -- pubsub` yields the same results.

However running `$ GOOGLE_CLOUD_DISABLE_GRPC=true tox -e system-tests -- pubsub` allows the tests to pass.

daspecster:
Closes #2908

I'm not sure this is the right way to go about fixing this. But it works!

### DIFF
--- a/pubsub/google/cloud/pubsub/client.py
+++ b/pubsub/google/cloud/pubsub/client.py
@@ -96,7 +96,7 @@ class Client(JSONClient):
                         host=self._connection.host)
                 else:
                     generated = make_gax_publisher_api(
-                        credentials=self._credentials)
+                        credentials=self._connection._credentials)
                 self._publisher_api = GAXPublisherAPI(generated, self)
             else:
                 self._publisher_api = JSONPublisherAPI(self)
@@ -112,7 +112,7 @@ class Client(JSONClient):
                         host=self._connection.host)
                 else:
                     generated = make_gax_subscriber_api(
-                        credentials=self._credentials)
+                        credentials=self._connection._credentials)
                 self._subscriber_api = GAXSubscriberAPI(generated, self)
             else:
                 self._subscriber_api = JSONSubscriberAPI(self)


dhermes:
Indeed. DERP!

I was [assuming][2] that (also [here][3])

```python
client._credentials is client._connection._credentials
```

in #2874 (and maybe #2875). But that isn't true: the `Connection` constructor [calls][1] `google.auth.credentials.with_scopes_if_required`.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8e499ae3d9d38c571a2788eb23681878cee17d5/core/google/cloud/_http.py#L82-L83
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8e499ae3d9d38c571a2788eb23681878cee17d5/pubsub/google/cloud/pubsub/client.py#L99
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8e499ae3d9d38c571a2788eb23681878cee17d5/pubsub/google/cloud/pubsub/client.py#L115

dhermes:
@daspecster I'm not really sure what to do here. Pub/Sub and Logging are "half-way" to just removing `Connection` altogether (or having a very different / stripped down version of the `Connection`). So this change needed to be made as part of that effort, the effort just got halted by Santa.

daspecster:
🎅 💥 

Ok, I can try and continue with the "extraction".  But if you want too, I'll close this and keep rolling with other things.

dhermes:
@daspecster I think the right fix (for now) is to just use `client._connection._credentials` ([here][1] and [here][2]).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8e499ae3d9d38c571a2788eb23681878cee17d5/pubsub/google/cloud/pubsub/client.py#L99
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8e499ae3d9d38c571a2788eb23681878cee17d5/pubsub/google/cloud/pubsub/client.py#L115

daspecster:
Update pushed! I ran the system tests and they passed so that should do the trick!

dhermes:
:+1: LGTM to merge (Travis is green already)

PR #2917, opened 2017-01-04
Only running 3.4 tests on PR builds.

dhermes:
It's worth noting that we are "one foot out the door" with Travis, so this change is just a temporary kludge to keep push builds from erroring out (due to timeout).

The "correct" long-term solution is:

-   run a nightly cron job which runs everything
-   run **everything** on "tag" builds
-   on all other non-PR builds, run a restricted set of tests based on the new changes added since the last build. It's worth noting that CircleCI makes these commits easy to access in the UI (unclear if easy in the VMs):

    screenshot from 2017-01-04 11-27-01

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
 
 script:
   - python2.7 scripts/run_unit_tests.py
-  - python3.4 scripts/run_unit_tests.py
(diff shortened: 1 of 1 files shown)


dhermes:
@tseaver Any objections here?

tseaver:
LGTM

PR #2918, opened 2017-01-05
Add gax support for entity annotations.

daspecster:
Towards #2753 

This adds support for `EntityAnnotation` types of detection.

Face, Safe Search and Image Properties is what's left.

Also, system tests have `use_gax` set for detection types that don't yet have `from_pb` defined.

### DIFF
--- a/system_tests/vision.py
+++ b/system_tests/vision.py
@@ -73,6 +73,10 @@ class BaseVisionTestCase(unittest.TestCase):
                   Likelihood.VERY_UNLIKELY]
         self.assertIn(likelihood, levels)
 
+    def _maybe_http_skip(self, message):
+        if not Config.CLIENT._use_gax:
+            self.skipTest(message)
+
 
 class TestVisionClientLogo(unittest.TestCase):
     def setUp(self):
@@ -190,6 +194,7 @@ class TestVisionClientFace(BaseVisionTestCase):
 
     def test_detect_faces_content(self):
         client = Config.CLIENT
+        self._maybe_http_skip('gRPC is required for face detection.')
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
         faces = image.detect_faces()
@@ -198,6 +203,7 @@ class TestVisionClientFace(BaseVisionTestCase):
             self._assert_face(face)
 
     def test_detect_faces_gcs(self):
+        self._maybe_http_skip('gRPC is required for face detection.')
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
         blob = Config.TEST_BUCKET.blob(blob_name)
@@ -206,7 +212,6 @@ class TestVisionClientFace(BaseVisionTestCase):
             blob.upload_from_file(file_obj)
 
         source_uri = 'gs://%s/%s' % (bucket_name, blob_name)
-
         client = Config.CLIENT
         image = client.image(source_uri=source_uri)
         faces = image.detect_faces()
@@ -215,6 +220,7 @@ class TestVisionClientFace(BaseVisionTestCase):
             self._assert_face(face)
 
     def test_detect_faces_filename(self):
+        self._maybe_http_skip('gRPC is required for face detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         faces = image.detect_faces()
@@ -361,6 +367,7 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_likelihood(safe_search.violence)
 
     def test_detect_safe_search_content(self):
+        self._maybe_http_skip('gRPC is required for safe search detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
@@ -370,6 +377,7 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_safe_search(safe_search)
 
     def test_detect_safe_search_gcs(self):
+        self._maybe_http_skip('gRPC is required for safe search detection.')
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
         blob = Config.TEST_BUCKET.blob(blob_name)
@@ -387,6 +395,7 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_safe_search(safe_search)
 
     def test_detect_safe_search_filename(self):
+        self._maybe_http_skip('gRPC is required for safe search detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         safe_searches = image.detect_safe_search()
@@ -484,6 +493,7 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
             self.assertNotEqual(color_info.score, 0.0)
 
     def test_detect_properties_content(self):
+        self._maybe_http_skip('gRPC is required for text detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
@@ -493,6 +503,8 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
         self._assert_properties(image_property)
 
     def test_detect_properties_gcs(self):
+        self._maybe_http_skip('gRPC is required for text detection.')
+        client = Config.CLIENT
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
         blob = Config.TEST_BUCKET.blob(blob_name)
@@ -502,7 +514,6 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
 
         source_uri = 'gs://%s/%s' % (bucket_name, blob_name)
 
-        client = Config.CLIENT
         image = client.image(source_uri=source_uri)
         properties = image.detect_properties()
         self.assertEqual(len(properties), 1)
@@ -510,6 +521,7 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
         self._assert_properties(image_property)
 
     def test_detect_properties_filename(self):
+        self._maybe_http_skip('gRPC is required for text detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         properties = image.detect_properties()
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -19,6 +19,8 @@ from google.cloud.grpc.vision.v1 import image_annotator_pb2
 
 from google.cloud._helpers import _to_bytes
 
+from google.cloud.vision.annotations import Annotations
+
 
 class _GAPICVisionAPI(object):
     """Vision API for interacting with the gRPC version of Vision.
@@ -28,7 +30,32 @@ class _GAPICVisionAPI(object):
     """
     def __init__(self, client=None):
         self._client = client
-        self._api = image_annotator_client.ImageAnnotatorClient()
+        self._annotator_client = image_annotator_client.ImageAnnotatorClient()
+
+    def annotate(self, image, features):
+        """Annotate images through GAX.
+
+        :type image: :class:`~google.cloud.vision.image.Image`
+        :param image: Instance of ``Image``.
+
+        :type features: list
+        :param features: List of :class:`~google.cloud.vision.feature.Feature`.
+
+        :rtype: :class:`~google.cloud.vision.annotations.Annotations`
+        :returns: Instance of ``Annotations`` with results or ``None``.
+        """
+        gapic_features = [_to_gapic_feature(feature) for feature in features]
+        gapic_image = _to_gapic_image(image)
+        request = image_annotator_pb2.AnnotateImageRequest(
+            image=gapic_image, features=gapic_features)
+        requests = [request]
+        annotator_client = self._annotator_client
+        images = annotator_client.batch_annotate_images(requests)
+        if len(images.responses) == 1:
+            return Annotations.from_pb(images.responses[0])
+        elif len(images.responses) > 1:
+            raise NotImplementedError(
+                'Multiple image processing is not yet supported.')
 
 
 def _to_gapic_feature(feature):
--- a/vision/google/cloud/vision/_http.py
+++ b/vision/google/cloud/vision/_http.py
@@ -14,6 +14,7 @@
 
 """HTTP Client for interacting with the Google Cloud Vision API."""
 
+from google.cloud.vision.annotations import Annotations
 from google.cloud.vision.feature import Feature
 
 
@@ -48,8 +49,12 @@ class _HTTPVisionAPI(object):
         data = {'requests': [request]}
         api_response = self._connection.api_request(
             method='POST', path='/images:annotate', data=data)
-        responses = api_response.get('responses')
-        return responses[0]
+        images = api_response.get('responses')
+        if len(images) == 1:
+            return Annotations.from_api_repr(images[0])
+        elif len(images) > 1:
+            raise NotImplementedError(
+                'Multiple image processing is not yet supported.')
 
 
 def _make_request(image, features):
--- a/vision/google/cloud/vision/annotations.py
+++ b/vision/google/cloud/vision/annotations.py
@@ -93,6 +93,51 @@ class Annotations(object):
                 _entity_from_response_type(feature_type, annotation))
         return cls(**annotations)
 
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct an instance of ``Annotations`` from protobuf.
+
+        :type response: :class:`~google.cloud.grpc.vision.v1.\
+                        image_annotator_pb2.AnnotateImageResponse`
+        :param response: ``AnnotateImageResponse`` from protobuf call.
+
+        :rtype: :class:`~google.cloud.vision.annotations.Annotations`
+        :returns: ``Annotations`` instance populated from gRPC response.
+        """
+        annotations = _process_image_annotations(response)
+        return cls(**annotations)
+
+
+def _process_image_annotations(image):
+    """Helper for processing annotation types from protobuf.
+
+    :type image: :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.\
+                 AnnotateImageResponse`
+    :param image: ``AnnotateImageResponse`` from protobuf.
+
+    :rtype: dict
+    :returns: Dictionary populated with entities from response.
+    """
+    return {
+        'labels': _make_entity_from_pb(image.label_annotations),
+        'landmarks': _make_entity_from_pb(image.landmark_annotations),
+        'logos': _make_entity_from_pb(image.logo_annotations),
+        'texts': _make_entity_from_pb(image.text_annotations),
+    }
+
+
+def _make_entity_from_pb(annotations):
+    """Create an entity from a gRPC response.
+
+    :type annotations:
+    :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.EntityAnnotation`
+    :param annotations: protobuf instance of ``EntityAnnotation``.
+
+    :rtype: list
+    :returns: List of ``EntityAnnotation``.
+    """
+    return [EntityAnnotation.from_pb(annotation) for annotation in annotations]
+
 
 def _entity_from_response_type(feature_type, results):
     """Convert a JSON result to an entity type based on the feature.
--- a/vision/google/cloud/vision/entity.py
+++ b/vision/google/cloud/vision/entity.py
@@ -64,10 +64,30 @@ class EntityAnnotation(object):
         description = response['description']
         locale = response.get('locale', None)
         locations = [LocationInformation.from_api_repr(location)
-                     for location in response.get('locations', [])]
+                     for location in response.get('locations', ())]
         mid = response.get('mid', None)
         score = response.get('score', None)
 
+        return cls(bounds, description, locale, locations, mid, score)
+
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct entity from Vision gRPC response.
+
+        :type response: :class:`~google.cloud.grpc.vision.v1.\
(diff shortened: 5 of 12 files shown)


vision/google/cloud/vision/_gax.py
```diff
@@ -30,6 +32,27 @@ def __init__(self, client=None):
...
+        :returns: Instance of ``Annotations`` with results.
+        """
+        gapic_features = [_to_gapic_feature(feature) for feature in features]
+        gapic_image = _to_gapic_image(image)
+        request = image_annotator_pb2.AnnotateImageRequest(
+            image=gapic_image, features=gapic_features)
+        requests = [request]
+        api = self._api
+        responses = api.batch_annotate_images(requests)
+        return Annotations.from_pb(responses.responses[0])
```
dhermes: My typical complaint applies here: this assumes `len(responses.responses) == 1` but doesn't check it

Also, the access `responses.responses` indicates that `responses` isn't named correctly? What type does `api.batch_annotate_images` return?

vision/google/cloud/vision/_http.py
```diff
@@ -49,7 +50,7 @@ def annotate(self, image, features):
         api_response = self._connection.api_request(
             method='POST', path='/images:annotate', data=data)
         responses = api_response.get('responses')
-        return responses[0]
+        return Annotations.from_api_repr(responses[0])
```
dhermes: Ditto here on `len(responses) == 1`. (I realize that is unrelated to this actual change)

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
...
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct an instance of ``Annotations`` from gRPC response.
+
+        :type response: :class:`~google.cloud.grpc.vision.v1.\
+                        image_annotator_pb2.AnnotateImageResponse`
+        :param response: ``AnnotateImageResponse`` from gRPC call.
+
+        :rtype: :class:`~google.cloud.vision.annotations.Annotations`
+        :returns: ``Annotations`` instance populated from gRPC response.
```
dhermes: s/gRPC response/protobuf

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
                 _entity_from_response_type(feature_type, annotation))
         return cls(**annotations)
 
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct an instance of ``Annotations`` from gRPC response.
```
dhermes: s/gRPC response/protobuf

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
...
+
+    :rtype: dict
+    :returns: Dictionary populated with entities from response.
+    """
+    annotations = {}
+    annotations['labels'] = _make_entity_from_pb(image.label_annotations)
+    annotations['landmarks'] = _make_entity_from_pb(image.landmark_annotations)
+    annotations['logos'] = _make_entity_from_pb(image.logo_annotations)
+    annotations['texts'] = _make_entity_from_pb(image.text_annotations)
+    return annotations
```
dhermes: Why not?

```python
    return {
        'labels': _make_entity_from_pb(image.label_annotations),
        'landmarks': _make_entity_from_pb(image.landmark_annotations),
        'logos': _make_entity_from_pb(image.logo_annotations),
        'texts': _make_entity_from_pb(image.text_annotations),
    }
```

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
...
+    """Create an entity from a gRPC response.
+
+    :type annotations:
+    :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.EntityAnnotation`
+    :param annotations: gRPC instance of ``EntityAnnotation``.
+
+    :rtype: list
+    :returns: List of ``EntityAnnotation``.
+    """
+
```
dhermes: Empty line for fun?

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
...
+    annotations['texts'] = _make_entity_from_pb(image.text_annotations)
+    return annotations
+
+
+def _make_entity_from_pb(annotations):
+    """Create an entity from a gRPC response.
+
+    :type annotations:
+    :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.EntityAnnotation`
+    :param annotations: gRPC instance of ``EntityAnnotation``.
```
dhermes: gRPC is the transport. protobuf is the object serialization format

vision/google/cloud/vision/annotations.py
```diff
@@ -93,6 +93,55 @@ def from_api_repr(cls, response):
...
+
+    :rtype: list
+    :returns: List of ``EntityAnnotation``.
+    """
+
+    entities = []
+    for annotation in annotations:
+        entities.append(EntityAnnotation.from_pb(annotation))
+    return entities
```
dhermes: Why not:

```python
    return [EntityAnnotation.from_pb(annotation)
            for annotation in annotations]
```

vision/google/cloud/vision/geometry.py
```diff
@@ -87,6 +101,19 @@ def from_api_repr(cls, response):
...
+        """Factory: construct location information from Vision gRPC response.
+
+        :type response: :class:`~google.cloud.vision.v1.LocationInfo`
+        :param response: gRPC response of ``LocationInfo``.
+
+        :rtype: :class:`~google.cloud.vision.geometry.LocationInformation`
+        :returns: ``LocationInformation`` with populated latitude and
+                  longitude.
+        """
+        return cls(response.lat_lng.latitude, response.lat_lng.longitude)
```
dhermes: Is there anyone on `response` other than `lat_lng`?

vision/google/cloud/vision/geometry.py
```diff
@@ -87,6 +101,19 @@ def from_api_repr(cls, response):
         longitude = response['latLng']['longitude']
         return cls(latitude, longitude)
 
+    @classmethod
+    def from_pb(cls, response):
```
dhermes: Shouldn't the argument be named `location_info`? The fact that it came as a response is less relevant to what it is?

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,28 @@ def test_ctor(self):
             api = self._make_one(client)
         self.assertIs(api._client, client)
 
+    def test_annotation(self):
+        from google.cloud.vision.feature import Feature
+        from google.cloud.vision.feature import FeatureTypes
+        from google.cloud.vision.image import Image
+
+        client = mock.Mock()
```
dhermes: Can you add `spec` to limit the "oops" effect? You can start with

```python
client = mock.Mock(spec=[])
```

and then add strings to the `spec` list as you realize you need missing attributes

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,28 @@ def test_ctor(self):
...
+        feature = Feature(FeatureTypes.LABEL_DETECTION, 5)
+        image_content = b'abc 1 2 3'
+        image = Image(client, content=image_content)
+        with mock.patch('google.cloud.vision._gax.image_annotator_client.'
+                        'ImageAnnotatorClient'):
+            api = self._make_one(client)
+
+        api._api = mock.Mock()
```
dhermes: Why `api._api`? Bad naming?

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        return Annotations
+
+    def _make_one(self, *args, **kw):
+        return self._get_target_class()(*args, **kw)
+
+    def test_ctor(self):
+        annotations = self._make_one(
+            faces=[True], properties=[True], labels=[True], landmarks=[True],
+            logos=[True], safe_searches=[True], texts=[True])
+        self.assertTrue(annotations.faces[0])
```
dhermes: This misses the length. You could just as easily do:

```python
self.assertEqual(annotations.faces. [True])
```

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertTrue(annotations.logos[0])
+        self.assertTrue(annotations.safe_searches[0])
+        self.assertTrue(annotations.texts[0])
+
+    def test_from_pb(self):
+        from google.cloud.grpc.vision.v1 import image_annotator_pb2
+
+        image_response = image_annotator_pb2.AnnotateImageResponse()
+        annotations = self._make_one().from_pb(image_response)
+        self.assertEquals(annotations.labels, [])
```
dhermes: `assertEquals` is a [deprecated alias][1] for `assertEqual`

[1]: https://docs.python.org/3.6/library/unittest.html#deprecated-aliases

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEquals(annotations.logos, [])
+        self.assertEquals(annotations.faces, ())
+        self.assertEquals(annotations.landmarks, [])
+        self.assertEquals(annotations.texts, [])
+        self.assertEquals(annotations.safe_searches, ())
+        self.assertEquals(annotations.properties, ())
+
+
+class TestMakeEntityFromPB(unittest.TestCase):
```
dhermes: Please follow our testing style: `Test__make_entity_from_pb`

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+    def _call_fut(self, annotations):
+        from google.cloud.vision.annotations import _make_entity_from_pb
+        return _make_entity_from_pb(annotations)
```
dhermes: Please be self-consistent. You have a blank line between your import in `_get_target_class()` above, but here you have no space between imports and code

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        entity_annotation.mid = mid
+        entity_annotation.locale = locale
+        entity_annotation.description = description
+        entity_annotation.score = score
+        entity_annotation.bounding_poly.vertices.add()
+        entity_annotation.bounding_poly.vertices[0].x = 1
+        entity_annotation.bounding_poly.vertices[0].y = 2
+        entity_annotation.locations.add()
+        entity_annotation.locations[0].lat_lng.latitude = 1.0
+        entity_annotation.locations[0].lat_lng.longitude = 2.0
```
dhermes: WDYT of this (takes 5 more lines, but maybe is "clearer"):

```python
from google.cloud.grpc.vision.v1 import geometry_pb2
from google.cloud.grpc.vision.v1 import image_annotator_pb2
from google.type import latlng_pb2

entity_annotation = image_annotator_pb2.EntityAnnotation(
    mid=mid,
    locale=locale,
    description=description,
    score=score,
    bounding_poly=geometry_pb2.BoundingPoly(
        vertices=[
            geometry_pb2.Vertex(x=1, y=2),
        ],
    ),
    locations=[
        image_annotator_pb2.LocationInfo(
            lat_lng=latlng_pb2.LatLng(latitude=1.0, longitude=2.0),
        ),
    ],
)
```

You could put it in a standalone function or staticmethod?

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEquals(annotations.safe_searches, ())
+        self.assertEquals(annotations.properties, ())
+
+
+class TestMakeEntityFromPB(unittest.TestCase):
+    def _call_fut(self, annotations):
+        from google.cloud.vision.annotations import _make_entity_from_pb
+        return _make_entity_from_pb(annotations)
+
+    def test_make_entity_from_pb(self):
```
dhermes: `test_it`?

vision/unit_tests/test_annotations.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEqual(entity.mid, mid)
+        self.assertEqual(entity.locale, locale)
+        self.assertEqual(entity.score, score)
+        self.assertEqual(entity.bounds.vertices[0].x_coordinate, 1)
+        self.assertEqual(entity.bounds.vertices[0].y_coordinate, 2)
+        self.assertEqual(entity.locations[0].latitude, 1.0)
+        self.assertEqual(entity.locations[0].longitude, 2.0)
+
+
+class TestProcessImageAnnotations(unittest.TestCase):
```
dhermes: Ditto here on naming

vision/google/cloud/vision/geometry.py
daspecster: Doesn't appear to be...
https://github.com/googleapis/api-client-staging/blob/22a53426ff8308f4bc227fa3aa529b63ff3e63cf/generated/python/grpc-google-cloud-vision-v1/google/cloud/grpc/vision/v1/image_annotator_pb2.py#L564

system_tests/vision.py
```diff
@@ -190,6 +190,7 @@ def _assert_face(self, face):
 
     def test_detect_faces_content(self):
         client = Config.CLIENT
+        client._use_gax = False
```
daspecster: These are temporary until all the annotation types are covered in both JSON and protobuf.
This way the system tests pass when you run with gRPC enabled and disabled.
This just essentially forces the tests that don't support protobuf yet to always run in HTTP mode.

dhermes: I realize **why** you don't it, I'm saying don't do it in this way. Just skip the test if you require gRPC or HTTP. See an [example][1]. Note that the skip can't be at import time since it depends on `Config.CLIENT`, which is set **at runtime**, so using `@unittest.skipIf` is a no-go.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/e7b1414a0b3d1c3d515de1117a18d38d08d55c09/system_tests/bigtable.py#L313-L318

vision/google/cloud/vision/geometry.py
```diff
@@ -73,20 +87,34 @@ def __init__(self, latitude, longitude):
...
+
+        :type location_info: :class:`~google.cloud.vision.v1.LocationInfo`
+        :param location_info: gRPC response of ``LocationInfo``.
+
+        :rtype: :class:`~google.cloud.vision.geometry.LocationInformation`
+        :returns: ``LocationInformation`` with populated latitude and
+                  longitude.
+        """
+        return cls(location_info.lat_lng.latitude,
+                   location_info.lat_lng.longitude)
```
dhermes: Also confirmed from the proto:
https://github.com/googleapis/googleapis/blob/7e594903ac603ff3b20976163613123be22d8f1c/google/cloud/vision/v1/image_annotator.proto#L293-L296

Thansk for the link to the `pb2` file, though I find the "true source" (i.e. the proto def) to be easier to parse.

vision/google/cloud/vision/_http.py
```diff
@@ -48,8 +49,9 @@ def annotate(self, image, features):
         data = {'requests': [request]}
         api_response = self._connection.api_request(
             method='POST', path='/images:annotate', data=data)
-        responses = api_response.get('responses')
-        return responses[0]
+        images = api_response.get('responses')
+        if len(images) == 1:
+            return Annotations.from_api_repr(images[0])
```
dhermes: What if `len(images) > 1`?

vision/google/cloud/vision/entity.py
```diff
@@ -63,13 +63,33 @@ def from_api_repr(cls, response):
         bounds = Bounds.from_api_repr(response.get('boundingPoly'))
         description = response['description']
         locale = response.get('locale', None)
-        locations = [LocationInformation.from_api_repr(location)
+        locations = [LocationInformation.from_api_repr(location['latLng'])
                      for location in response.get('locations', [])]
```
dhermes: I realize it's not in this PR, but it's good form to use `()` as your default empty container (since it takes up less space than `[]` and is just being thrown away anyhow)

vision/google/cloud/vision/geometry.py
```diff
@@ -73,20 +87,34 @@ def __init__(self, latitude, longitude):
...
 
     @classmethod
-    def from_api_repr(cls, response):
+    def from_api_repr(cls, location_info):
         """Factory: construct location information from Vision API response.
 
-        :type response: dict
-        :param response: Dictionary response of locations.
+        :type location_info: dict
+        :param location_info: Dictionary response of locations.
```
dhermes: Seems now to be just one location? Also, since you use the `latLng` key in your caller, this is now a `lat_lng` dict **not** a `location_info` dict.

vision/google/cloud/vision/_gax.py
```diff
@@ -28,7 +30,29 @@ class _GAPICVisionAPI(object):
...
+        """Annotate images through GAX.
+
+        :type image: :class:`~google.cloud.vision.image.Image`
+        :param image: Instance of ``Image``.
+
+        :type features: list
+        :param features: List of :class:`~google.cloud.vision.feature.Feature`.
+
+        :rtype: :class:`~google.cloud.vision.annotations.Annotations`
+        :returns: Instance of ``Annotations`` with results.
```
dhermes: This is now "optional" per your `if`

vision/google/cloud/vision/_http.py
daspecster: Then it returns `None`, but maybe it should return `[]` or `()`?

vision/google/cloud/vision/geometry.py
daspecster: I moved `latLng` back from the caller to the class.
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2918/commits/9287850457d7f8e04db2af5e7ac597eb0a8a0896

daspecster:
@dhermes I updated the system tests and I think I addressed most of your feedback.

LMKWYT

vision/google/cloud/vision/_http.py
daspecster: Also, once I get this GAPIC support added, I have two other PRs. One for proper handling of batch images and another for handling errors received from the service.

daspecster: Yeah, so it shouldn't be an error actually. I think I should return a list of the images. I was planning on doing that in a soon following PR.

I'll have this raise an exception for now if you like and then I'll take it back out with my image batching PR.

vision/google/cloud/vision/entity.py
```diff
@@ -64,12 +64,32 @@ def from_api_repr(cls, response):
...
         locations = [LocationInformation.from_api_repr(location)
-                     for location in response.get('locations', [])]
+                     for location in response.get('locations', ())]
         mid = response.get('mid', None)
         score = response.get('score', None)
 
         return cls(bounds, description, locale, locations, mid, score)
 
+    @classmethod
+    def from_pb(cls, response):
```
daspecster: So this one is actually a `AnnotateImageResponse` object. So I think it's right already?

I'll double check but I think I hunted around the others in this PR to make sure they were accurately named.

vision/google/cloud/vision/geometry.py
```diff
@@ -73,20 +87,35 @@ def __init__(self, latitude, longitude):
...
 
         :rtype: :class:`~google.cloud.vision.geometry.LocationInformation`
         :returns: ``LocationInformation`` with populated latitude and
                   longitude.
         """
-        latitude = response['latLng']['latitude']
-        longitude = response['latLng']['longitude']
+        lat_long = location_info.get('latLng')
+        latitude = lat_long.get('latitude')
+        longitude = lat_long.get('longitude')
```
daspecster: The  constructor looks like this, so it will make the class where either `LocationInformation.latitude` or `LocationInformation.longitude` is `None.

```python
def __init__(self, latitude, longitude):
    self._latitude = latitude
    self._longitude = longitude
```

Should I add `None` as a possible value to the docstrings?

daspecster:
Re:
>But len(images) > 1 is an error. I realize it returns None, I'm saying that we should be more defensive / have clearer assumptions.

How about this?

```python
if len(images) == 1:
    return Annotations.from_api_repr(images[0])
elif len(images) > 1:
    raise ValueError('Multiple image processing is not yet supported.')
```

vision/google/cloud/vision/_http.py
```diff
@@ -52,6 +52,8 @@ def annotate(self, image, features):
         images = api_response.get('responses')
         if len(images) == 1:
             return Annotations.from_api_repr(images[0])
+        elif len(images) > 1:
+            raise ValueError('Multiple image processing is not yet supported.')
```
dhermes: 1. Based on the message it seems like `NotImplementedError` could be just as appropriate
1. And `if len(images) == 0`, that's OK?

vision/google/cloud/vision/geometry.py
```diff
@@ -25,35 +25,33 @@ def __init__(self, vertices):
...
     @classmethod
-    def from_pb(cls, response_vertices):
+    def from_pb(cls, vertices):
         """Factory: construct BoundsBase instance from Vision gRPC response.
 
-        :type response_vertices: :class:`~google.cloud.grpc.vision.v1.\
+        :type vertices: :class:`~google.cloud.grpc.vision.v1.\
                                  geometry_pb2.BoundingPoly`
-        :param response_vertices: List of vertices.
+        :param vertices: List of vertices.
```
dhermes: Is the `:type:` correct? It doesn't seem like it's a list?

vision/google/cloud/vision/_http.py
daspecster: `if len(images) == 0` then that means the service didn't detect anything it was asked too...most of the time.

There's actually something that I missed implementing originally. There's an `error` value in the response that I don't capture. That's in a PR to come shortly though.

I'll update to `NotImplementedError` real quick.

daspecster:
Squashed with `NotImplementedError` update.

daspecster:
Travis is green, merging.

PR #2919, opened 2017-01-05
Switch to SVG for coverage badge.

craigcitro:
(It felt left out, since all its friends were clearly upgraded.)


dhermes:
Thanks Craig! Long time no see!

PR #2922, opened 2017-01-06
Vision: Add gRPC support for face detection.

daspecster:
~Based from #2918~

This adds gRPC support for `face_detect()`.

### DIFF
--- a/system_tests/vision.py
+++ b/system_tests/vision.py
@@ -62,8 +62,8 @@ class BaseVisionTestCase(unittest.TestCase):
     def _assert_coordinate(self, coordinate):
         if coordinate is None:
             return
+        self.assertIsNotNone(coordinate)
         self.assertIsInstance(coordinate, (int, float))
-        self.assertNotEqual(coordinate, 0.0)
 
     def _assert_likelihood(self, likelihood):
         from google.cloud.vision.likelihood import Likelihood
@@ -73,8 +73,8 @@ class BaseVisionTestCase(unittest.TestCase):
                   Likelihood.VERY_UNLIKELY]
         self.assertIn(likelihood, levels)
 
-    def _maybe_http_skip(self, message):
-        if not Config.CLIENT._use_gax:
+    def _pb_not_implemented_skip(self, message):
+        if Config.CLIENT._use_gax:
             self.skipTest(message)
 
 
@@ -150,7 +150,7 @@ class TestVisionClientFace(BaseVisionTestCase):
 
         for landmark in LandmarkTypes:
             if landmark is not LandmarkTypes.UNKNOWN_LANDMARK:
-                feature = getattr(landmarks, landmark.value.lower())
+                feature = getattr(landmarks, landmark.name.lower())
                 self.assertIsInstance(feature, Landmark)
                 self.assertIsInstance(feature.position, Position)
                 self._assert_coordinate(feature.position.x_coordinate)
@@ -194,7 +194,6 @@ class TestVisionClientFace(BaseVisionTestCase):
 
     def test_detect_faces_content(self):
         client = Config.CLIENT
-        self._maybe_http_skip('gRPC is required for face detection.')
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
         faces = image.detect_faces()
@@ -203,7 +202,6 @@ class TestVisionClientFace(BaseVisionTestCase):
             self._assert_face(face)
 
     def test_detect_faces_gcs(self):
-        self._maybe_http_skip('gRPC is required for face detection.')
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
         blob = Config.TEST_BUCKET.blob(blob_name)
@@ -220,7 +218,6 @@ class TestVisionClientFace(BaseVisionTestCase):
             self._assert_face(face)
 
     def test_detect_faces_filename(self):
-        self._maybe_http_skip('gRPC is required for face detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         faces = image.detect_faces()
@@ -367,7 +364,8 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_likelihood(safe_search.violence)
 
     def test_detect_safe_search_content(self):
-        self._maybe_http_skip('gRPC is required for safe search detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for safe search detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
@@ -377,7 +375,8 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_safe_search(safe_search)
 
     def test_detect_safe_search_gcs(self):
-        self._maybe_http_skip('gRPC is required for safe search detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for safe search detection.')
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
         blob = Config.TEST_BUCKET.blob(blob_name)
@@ -395,7 +394,8 @@ class TestVisionClientSafeSearch(BaseVisionTestCase):
         self._assert_safe_search(safe_search)
 
     def test_detect_safe_search_filename(self):
-        self._maybe_http_skip('gRPC is required for safe search detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for safe search detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         safe_searches = image.detect_safe_search()
@@ -493,7 +493,8 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
             self.assertNotEqual(color_info.score, 0.0)
 
     def test_detect_properties_content(self):
-        self._maybe_http_skip('gRPC is required for text detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
@@ -503,7 +504,8 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
         self._assert_properties(image_property)
 
     def test_detect_properties_gcs(self):
-        self._maybe_http_skip('gRPC is required for text detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
@@ -521,7 +523,8 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
         self._assert_properties(image_property)
 
     def test_detect_properties_filename(self):
-        self._maybe_http_skip('gRPC is required for text detection.')
+        self._pb_not_implemented_skip(
+            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         properties = image.detect_properties()
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -17,8 +17,6 @@
 from google.cloud.gapic.vision.v1 import image_annotator_client
 from google.cloud.grpc.vision.v1 import image_annotator_pb2
 
-from google.cloud._helpers import _to_bytes
-
 from google.cloud.vision.annotations import Annotations
 
 
@@ -85,7 +83,7 @@ def _to_gapic_image(image):
               :class:`~google.cloud.vision.image.Image`.
     """
     if image.content is not None:
-        return image_annotator_pb2.Image(content=_to_bytes(image.content))
+        return image_annotator_pb2.Image(content=image.content)
     if image.source is not None:
         return image_annotator_pb2.Image(
             source=image_annotator_pb2.ImageSource(
--- a/vision/google/cloud/vision/client.py
+++ b/vision/google/cloud/vision/client.py
@@ -58,7 +58,7 @@ class Client(JSONClient):
     _vision_api_internal = None
 
     def __init__(self, project=None, credentials=None, http=None,
-                 use_gax=False):
+                 use_gax=None):
         super(Client, self).__init__(
             project=project, credentials=credentials, http=http)
         self._connection = Connection(
--- a/vision/google/cloud/vision/face.py
+++ b/vision/google/cloud/vision/face.py
@@ -17,9 +17,24 @@
 
 from enum import Enum
 
+from google.cloud.grpc.vision.v1 import image_annotator_pb2
+
 from google.cloud.vision.geometry import BoundsBase
 from google.cloud.vision.likelihood import Likelihood
 from google.cloud.vision.geometry import Position
+
+
+def _get_pb_likelihood(likelihood):
+    """Convert protobuf Likelihood integer value to Likelihood instance.
+
+    :type likelihood: int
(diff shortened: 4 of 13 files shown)


system_tests/vision.py
```diff
@@ -63,7 +63,6 @@ def _assert_coordinate(self, coordinate):
         if coordinate is None:
             return
         self.assertIsInstance(coordinate, (int, float))
-        self.assertNotEqual(coordinate, 0.0)
```
dhermes: Why not?

system_tests/vision.py
```diff
@@ -133,6 +132,7 @@ def test_detect_logos_gcs(self):
 
 class TestVisionClientFace(BaseVisionTestCase):
     def setUp(self):
+        Config.CLIENT = vision.Client(use_gax=True)
```
dhermes: Why?

system_tests/vision.py
```diff
@@ -423,6 +425,7 @@ def _assert_text(self, text):
 
     def test_detect_text_content(self):
         client = Config.CLIENT
+        client._use_gax = True
```
dhermes: I am :-1: on this. See my comments on the other PR

vision/google/cloud/vision/face.py
```diff
@@ -22,6 +22,29 @@
...
+def _get_pb_likelihood(likelihood):
+    """Convert gRPC Likelihood integer value to Likelihood instance.
+
+    :type likelihood: int
+    :param likelihood: Protobuf integer representing ``Likelihood``.
+
+    :rtype: :class:`~google.cloud.vision.likelihood.Likelihood`
+    :returns: Instance of ``Likelihood`` converted from gRPC value.
+    """
+    likelihood_pb = LIKELIHOOD_PB.get(likelihood)
```
dhermes: Where does `likelihood` come from the `pb2` module should have a way to get the enum

vision/google/cloud/vision/face.py
```diff
@@ -42,6 +65,23 @@ def from_api_repr(cls, response):
 
         return cls(roll, pan, tilt)
 
+    @classmethod
+    def from_pb(cls, response):
```
dhermes: Similar comment here as in #2918, the fact that the argument is a `response` is less important than the fact this it's an `Angle`.

vision/google/cloud/vision/face.py
```diff
@@ -42,6 +65,23 @@ def from_api_repr(cls, response):
...
+        """Factory: convert gRPC Angle object to local Angle object.
+
+        :type response: :class:`~google.cloud.grpc.vision.v1.\
+                        image_annotator_pb2.FaceAnnotation`
```
dhermes: Copy-pasta?

vision/google/cloud/vision/face.py
```diff
@@ -101,6 +141,24 @@ def from_api_repr(cls, response):
         return cls(joy_likelihood, sorrow_likelihood, surprise_likelihood,
                    anger_likelihood)
 
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct `Emotions` from Vision API response.
```
dhermes: Single-backtick: Markdown
Double-backtick: RST

vision/google/cloud/vision/face.py
```diff
@@ -101,6 +141,24 @@ def from_api_repr(cls, response):
...
+
+        :type response: dict
```
dhermes: ISTM that `response` isn't a dict?

vision/google/cloud/vision/face.py
```diff
@@ -175,13 +233,39 @@ def from_api_repr(cls, response):
...
+        fd_bounds = FDBounds.from_pb(response.fd_bounding_poly)
+        headwear_likelihood = Likelihood[
+            LIKELIHOOD_PB[response.headwear_likelihood]]
+        image_properties = FaceImageProperties.from_pb(response)
+        landmarks = Landmarks.from_pb(response.landmarks)
+        landmarking_confidence = response.landmarking_confidence
+
+        return cls(angles, bounds, detection_confidence, emotions, fd_bounds,
+                   headwear_likelihood, image_properties, landmarks,
+                   landmarking_confidence)
```
dhermes: Whoa! A mouthful here.

vision/google/cloud/vision/face.py
```diff
@@ -326,6 +410,18 @@ def from_api_repr(cls, response):
 
         return cls(blurred, underexposed)
 
+    @classmethod
+    def from_pb(cls, response):
+        """Factory: construct image properties from image.
+
```
dhermes: Need args description

vision/google/cloud/vision/face.py
```diff
@@ -353,41 +449,110 @@ class LandmarkTypes(Enum):
...
-    LEFT_EAR_TRAGION = 'LEFT_EAR_TRAGION'
-    RIGHT_EAR_TRAGION = 'RIGHT_EAR_TRAGION'
-    LEFT_EYE_PUPIL = 'LEFT_EYE_PUPIL'
-    RIGHT_EYE_PUPIL = 'RIGHT_EYE_PUPIL'
-    FOREHEAD_GLABELLA = 'FOREHEAD_GLABELLA'
-    CHIN_GNATHION = 'CHIN_GNATHION'
-    CHIN_LEFT_GONION = 'CHIN_LEFT_GONION'
-    CHIN_RIGHT_GONION = 'CHIN_RIGHT_GONION'
+    # Unknown face landmark detected. Should not be filled.
+    UNKNOWN_LANDMARK = 0
```
dhermes: Don't hardcode this, it's in the generated `pb2` module

vision/google/cloud/vision/face.py
```diff
@@ -414,6 +579,20 @@ def from_api_repr(cls, response_landmark):
         landmark_type = LandmarkTypes[response_landmark['type']]
         return cls(position, landmark_type)
 
+    @classmethod
+    def from_pb(cls, response_landmark):
+        """Factory: construct an instance of a Landmark from a response.
+
+        :type response_landmark: dict
```
dhermes: Not a dict?

vision/google/cloud/vision/face.py
```diff
@@ -434,8 +613,42 @@ def landmark_type(self):
...
+
+        :type response: dict
+        :param response: JSON face annotation.
+
+        :rtype: :class:`~google.cloud.vision.face.Landmarks`
+        :returns: Instance of ``Landmarks`` populated with facial landmarks.
+        """
+        landmarks = []
+        for landmark in response:
+            landmarks.append(Landmark.from_api_repr(landmark))
```
dhermes: Why not a list comprehension?

vision/google/cloud/vision/image.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, client, content=None, filename=None, source_uri=None):
                 content = file_obj.read()
 
         if content is not None:
-            content = _bytes_to_unicode(b64encode(_to_bytes(content)))
+            content = _to_bytes(content)
```
dhermes: Why?

vision/google/cloud/vision/image.py
```diff
@@ -67,7 +67,7 @@ def as_dict(self):
         """
         if self.content:
             return {
-                'content': self.content
+                'content': _bytes_to_unicode(b64encode(self.content))
```
dhermes: Why? (Same as above, I assume?) Also, trailing comma

vision/unit_tests/test_annotations.py
```diff
@@ -88,6 +88,21 @@ def test_make_entity_from_pb(self):
         self.assertEqual(entity.locations[0].longitude, 2.0)
 
 
+class TestMakeFacesFromPB(unittest.TestCase):
```
dhermes: Please follow our style: `Test__make_faces_from_pb`

system_tests/vision.py
daspecster: So with a face, there could be a facial feature that starts at coordinate 0,0 of a facial bounds box which would be 0,0.

vision/google/cloud/vision/face.py
daspecster: Wouldn't I need to make a `face.py` and a `face_pb.py` file in that case? So that the imports don't crash if protobuf is disabled/not available?

system_tests/vision.py
dhermes: So why was it `0.0` before?

vision/google/cloud/vision/face.py
dhermes: I think the generated stuff `try/except`-s the `grpc` imports now. Double check?

system_tests/vision.py
daspecster: Sorry wrong reason. I forgot, protobuf was returning an `int` there so I was getting `AssertionError: 0 == 0.0`.
I could do `self.assertNotEqual(0,[0, 0.0])` I suppose.

dhermes: In which version of Python? They are meant to be equal.

daspecster:
Rebased on master.

vision/google/cloud/vision/image.py
daspecster: So the PB version of base64 encodes the data for us.

So for the PB version I need the content unencoded and for the HTTP version I need the data encoded.

vision/google/cloud/vision/face.py
daspecster: So I converted the HTTP version to handle it with integers so that both PB and HTTP versions handle the types the same way.

I might be able to change the HTTP version so it just uses the values that are returned but that seems kind of like the wild west.

vision/google/cloud/vision/image.py
daspecster: Yup same as above. HTTP needs the content to be base64 encoded and the PB version needs the content not to be.

system_tests/vision.py
```diff
@@ -137,6 +137,7 @@ def test_detect_logos_gcs(self):
 
 class TestVisionClientFace(BaseVisionTestCase):
     def setUp(self):
+        Config.CLIENT = vision.Client()
```
dhermes: Do this in `setupModule` and nowhere else.

vision/google/cloud/vision/_gax.py
```diff
@@ -85,7 +83,7 @@ def _to_gapic_image(image):
               :class:`~google.cloud.vision.image.Image`.
     """
     if image.content is not None:
-        return image_annotator_pb2.Image(content=_to_bytes(image.content))
+        return image_annotator_pb2.Image(content=image.content)
```
dhermes: Why did this drop?

vision/google/cloud/vision/client.py
```diff
@@ -58,7 +58,7 @@ class Client(JSONClient):
     _vision_api_internal = None
 
     def __init__(self, project=None, credentials=None, http=None,
-                 use_gax=False):
+                 use_gax=None):
```
dhermes: Why this change?

vision/google/cloud/vision/face.py
```diff
@@ -84,19 +119,38 @@ def __init__(self, joy_likelihood, sorrow_likelihood,
...
+        :type emotions: dict
+        :param emotions: Response dictionary representing a face.
+
+        :rtype: :class:`~google.cloud.vision.face.Emotions`
+        :returns: Populated instance of ``Emotions``.
+        """
+        joy_likelihood = Likelihood[emotions['joyLikelihood']]
+        sorrow_likelihood = Likelihood[emotions['sorrowLikelihood']]
+        surprise_likelihood = Likelihood[emotions['surpriseLikelihood']]
+        anger_likelihood = Likelihood[emotions['angerLikelihood']]
```
dhermes: `Likelihood` is an `enum.Enum`?

vision/google/cloud/vision/face.py
dhermes: I'm not following. Are you saying it **needs** to be hard-coded?

vision/google/cloud/vision/geometry.py
```diff
@@ -219,7 +231,8 @@ def x_coordinate(self):
         :rtype: float
         :returns: X position coordinate.
         """
-        return self._x_coordinate
+        if self._x_coordinate != 0:
+            return self._x_coordinate
```
dhermes: No need for this branch if you don't store `0` in the constructor (i.e. just use `None`)

vision/google/cloud/vision/image.py
dhermes: WAT? It's an image, it should always be bytes? If `base64.b64encode` requires a `str` in Python 3, just convert at the call site.

vision/google/cloud/vision/image.py
dhermes: See my comment above

(Also, trailing comma)

vision/unit_tests/test_annotations.py
```diff
@@ -73,7 +73,7 @@ def test_from_pb(self):
         annotations = self._make_one().from_pb(image_response)
         self.assertEqual(annotations.labels, [])
         self.assertEqual(annotations.logos, [])
-        self.assertEqual(annotations.faces, ())
+        self.assertEqual(annotations.faces, [])
```
dhermes: ??

vision/unit_tests/test_annotations.py
```diff
@@ -107,6 +107,21 @@ def test_it(self):
         self.assertEqual(entity.locations[0].longitude, 2.0)
 
 
+class Test__make_faces_from_pb(unittest.TestCase):
+    def _call_fut(self, annotations):
+        from google.cloud.vision.annotations import _make_faces_from_pb
+        return _make_faces_from_pb(annotations)
+
+    def test_make_face_from_pb(self):
```
dhermes: No need for a change, but this name is repetetive. Could just go with `test_it`

vision/google/cloud/vision/face.py
daspecster: TL:DR;  For now I think it's needed. But depending on how we want to handle HTTP responses I may drop it in a future PR. 

2 days ago I would have said "Yeah I believe so".

But after the Speech conversation today...I could just use the value that HTTP sends back without verifying it on our side.

i.e.

HTTP
```
"landmark_type": "LEFT_EYE"
```

Protobuf
```
landmark_type: 1
```

But probably should do that in a different PR anyway.

vision/google/cloud/vision/image.py
daspecster: Sorry I was vague on the kind of encoding.

The pb code base64 encodes the bytes for us. But yeah they're always bytes.

The only place it gets converted to `str` is in the comment below which is in the `to_dict` method.

vision/google/cloud/vision/_gax.py
daspecster: Because it's handled on the constructor of [`Image`](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2922/files#diff-fd4c8b0a5b301b229c829279de2c008bR57).

vision/google/cloud/vision/face.py
daspecster: Yup. [Can be seen here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4f5f782b0cc64920cf63a185b49a430b463d3199/vision/google/cloud/vision/likelihood.py#L21).

vision/unit_tests/test_annotations.py
daspecster: It returns a list from the [list comprehension](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2922/files#diff-764c6fd4c7ba386b0fd21683de027e6cR153) now, instead of the signature default which is `()`.

vision/google/cloud/vision/client.py
daspecster: With the `_maybe_http` addition, it can go back to `None` like it should be.

vision/google/cloud/vision/image.py
dhermes: Ah this is the constructor. :+1: to this change

daspecster:
Merging after chat LGTM.

PR #2933, opened 2017-01-12
Add storage.blob property time_created (timeCreated)

bivald:
This is a super small PR which adds storage.blob property time_created (timeCreated) which I found myself needing recently. It's basically a copy paste of time_deleted. Unit test added.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -933,6 +933,21 @@ class Blob(_PropertyMixin):
             return _rfc3339_to_datetime(value)
 
     @property
+    def time_created(self):
+        """Retrieve the timestamp at which the object was created.
+
+        See: https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :rtype: :class:`datetime.datetime` or ``NoneType``
+        :returns: Datetime object parsed from RFC3339 valid timestamp, or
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -933,6 +933,21 @@ def time_deleted(self):
...
+        See: https://cloud.google.com/storage/docs/json_api/v1/objects
+
+        :rtype: :class:`datetime.datetime` or ``NoneType``
+        :returns: Datetime object parsed from RFC3339 valid timestamp, or
+                  ``None`` if the property is not set locally.
+        """
+        value = self._properties.get('timeCreated')
+        if value is not None:
+            return _rfc3339_to_datetime(value)
+
```
daspecster: One too many spaces here for lint I think.

daspecster:
@bivald Thanks!

Looks like the test doesn't cover the `None` case.
https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/191366039#L308

bivald:
@daspecster I've added the missing unit test as well as fixed the lint-error, I realised that I've put the unit test in the wrong order so I've fixed that as well.

daspecster:
Thanks @bivald!

bivald:
Thank you :)

PR #2934, opened 2017-01-12
Fixing credentials without scopes issue in logging.

dhermes:
See:
https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/191433056#L2195

This is a band-aid for now. I caused it in #2875 and will roll-back this roll-back at a later time.

@daspecster fixed a similar breakage in #2909.


daspecster:
Travis is just finishing up, but it LGTM.

Issue #2583, opened 2016-10-21
Datastore: _Rendezvous of RPC that terminated with StatusCode.UNAVAILABLE


Bogdanp:
We see this fairly often on `commit` with google-cloud-datastore version `0.20`. I believe these should either be retried with exponential backoff automatically by the library (according to [this](https://cloud.google.com/datastore/docs/concepts/errors)) or a more specific error should be raised so user code can deal w/ it (preferably one exception for every one of the cases listed on that doc).

```
_Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, {"created":"@1476898717.596308747","description":"Secure read failed","file":"src/core/lib/security/transport/secure_endpoint.c","file_line":157,"grpc_status":14,"referenced_errors":[{"created":"@1476898717.596257572","description":"EOF","file":"src/core/lib/iomgr/tcp_posix.c","file_line":235}]})>
```

Edit:

Here's our current (somewhat tested) workaround for this issue in our internal ORM:

``` python


# The maximum number of retries that should be done per Datastore
# error code.
_MAX_RETRIES_BY_CODE = {
    grpc.StatusCode.INTERNAL: 1,
    grpc.StatusCode.ABORTED: 5,  # Only retried for non-transactional commits
    grpc.StatusCode.UNAVAILABLE: 5,
    grpc.StatusCode.DEADLINE_EXCEEDED: 5,
}


def _handle_errors(f, transactional=False):
    @functools.wraps(f)
    def handler(*args, **kwargs):
        retries = 0
        while True:
            try:
                return f(*args, **kwargs)
            # TODO: Replace w/ concrete error types once/if they are
            # added to gcloud.  See: google-cloud-python/issues/2583
            except google.cloud.exceptions._Rendezvous as e:
                code = e.code()
                max_retries = _MAX_RETRIES_BY_CODE.get(code)
                if max_retries is None or transactional and code == grpc.StatusCode.ABORTED:
                    raise

                if retries > max_retries:
                    raise RetriesExceeded(e)

                backoff = min(0.0625 * 2 ** retries, 1.0)
                bread.get_logger().debug("Sleeping for %r before retrying failed request...", backoff)

                retries += 1
                time.sleep(backoff)

    return handler


class Client(datastore.Client):
    def __init__(self, *args, **kwargs):
        super(Client, self).__init__(*args, **kwargs)

        self.delete_multi = _handle_errors(self.delete_multi)
[...]

PR #2936, opened 2017-01-17
Create exceptions (and bases) for RPC-only errors in core

bendemaree:
Several RPC exceptions that can be raised from a protobuf response map to the same HTTP status code, which creates some semantic overlap, as the [current strategy](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/datastore/google/cloud/datastore/_http.py#L39) is to always map to an HTTP status code. An example of this is `DATA_LOSS` and `INTERNAL`, which are two different conditions as reported by the backend but are mapped to HTTP 500 (as per the [comments in the proto definition](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L178)), but may have different operational semantics.

My intent in doing this is to address the issue in #2583 which is thus far unresolved and seems to be dead in the water. After establishing these new exception classes, it's a simple matter to [raise them when relevant](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/datastore/google/cloud/datastore/_http.py#L39) rather than mashing multiple backend error codes onto a single exception as is currently implemented. As noted in #2583 this prevents the library user from appropriately handling some error conditions with retry logic.

Pending this, I'll follow up with a simple PR that raises the RPC-specific exceptions.

Note: I hemmed and hawed about re-using the `code` class attribute on these exceptions, because it is a reach IMO to extend the semantics of that from "HTTP status code" to "some protocol that has numeric status codes" but I decided it's alright in this case primarily because it keeps it simple and there's little danger of an additional 84 proto error codes being gobbled up (at which point you'd lose the invariant that `status code < 100 => RPC status code, not HTTP`).

### DIFF
--- a/core/unit_tests/test_exceptions.py
+++ b/core/unit_tests/test_exceptions.py
@@ -46,6 +46,13 @@ class Test_GoogleCloudError(unittest.TestCase):
         self.assertEqual(e.message, 'Testing')
         self.assertEqual(list(e.errors), [ERROR])
 
+    def test_rpc_only_errors_not_exported(self):
+        from google.cloud.exceptions import GoogleCloudRPCError
+        from google.cloud.exceptions import _HTTP_CODE_TO_EXCEPTION
+        http_exceptions = _HTTP_CODE_TO_EXCEPTION.values()
+        self.assertFalse(
+            any([issubclass(e, GoogleCloudRPCError) for e in http_exceptions]))
+
 
 class Test_make_exception(unittest.TestCase):
 
--- a/core/unit_tests/test_exceptions.py
+++ b/core/unit_tests/test_exceptions.py
@@ -49,6 +49,7 @@ class Test_GoogleCloudError(unittest.TestCase):
     def test_rpc_only_errors_not_exported(self):
         from google.cloud.exceptions import GoogleCloudRPCError
         from google.cloud.exceptions import _HTTP_CODE_TO_EXCEPTION
+
         http_exceptions = _HTTP_CODE_TO_EXCEPTION.values()
         self.assertFalse(
             any([issubclass(e, GoogleCloudRPCError) for e in http_exceptions]))
--- a/core/google/cloud/exceptions.py
+++ b/core/google/cloud/exceptions.py
@@ -15,6 +15,7 @@
 """Custom exceptions for :mod:`google.cloud` package.
 
 See: https://cloud.google.com/storage/docs/json_api/v1/status-codes
+See: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
 """
 
 # Avoid the grpc and google.cloud.grpc collision.
@@ -70,6 +71,11 @@ class GoogleCloudError(Exception):
         :returns: a list of mappings describing each error.
         """
         return [copy.deepcopy(error) for error in self._errors]
+
+
+class GoogleCloudRPCError(GoogleCloudError):
+    """Base error class for RPC errors that do not map directly to HTTP errors.
+    """
 
 
 class Redirection(GoogleCloudError):
@@ -181,8 +187,21 @@ class ServiceUnavailable(ServerError):
 
 
 class GatewayTimeout(ServerError):
-    """Excepption mapping a `504 Gateway Timeout'` response."""
+    """Exception mapping a '504 Gateway Timeout' response."""
     code = 504
+
+
+class RPCServerError(GoogleCloudRPCError):
+    """Base for 5xx-like RPC errors:  (abstract)"""
+
+
+class Unknown(RPCServerError):
+    """Exception mapping a '2 UNKNOWN' RPC error."""
+    code = 2
+
+
+class DataLoss(RPCServerError):
+    """Exception mapping a '15 DATA LOSS' RPC error."""
 
 
 def make_exception(response, content, error_info=None, use_json=True):
@@ -245,8 +264,11 @@ def _walk_subclasses(klass):
             yield subsub
 
 
-# Build the code->exception class mapping.
+# Build the HTTP code->exception class mapping, excluding RPC-only exceptions.
 for _eklass in _walk_subclasses(GoogleCloudError):
+    if issubclass(_eklass, GoogleCloudRPCError):
+        continue
+
     code = getattr(_eklass, 'code', None)
     if code is not None:
         _HTTP_CODE_TO_EXCEPTION[code] = _eklass
--- a/core/google/cloud/exceptions.py
+++ b/core/google/cloud/exceptions.py
@@ -202,6 +202,7 @@ class Unknown(RPCServerError):
 
 class DataLoss(RPCServerError):
     """Exception mapping a '15 DATA LOSS' RPC error."""
+    code = 15
 
 
 def make_exception(response, content, error_info=None, use_json=True):


core/unit_tests/test_exceptions.py
```diff
@@ -46,6 +46,13 @@ def test_ctor_explicit(self):
         self.assertEqual(e.message, 'Testing')
         self.assertEqual(list(e.errors), [ERROR])
 
+    def test_rpc_only_errors_not_exported(self):
+        from google.cloud.exceptions import GoogleCloudRPCError
+        from google.cloud.exceptions import _HTTP_CODE_TO_EXCEPTION
+        http_exceptions = _HTTP_CODE_TO_EXCEPTION.values()
```
daspecster: I know the other examples don't have this but could you add a space between the imports and the test code?

daspecster:
Thanks @bendemaree!
This looks pretty good to me besides my one comment.

WDYT @dhermes & @tseaver?

dhermes:
I am :-1: to this approach. For now, we support both JSON-over-HTTP (or protobuf-over-HTTP in the case of `datastore`) and protobuf-over-gRPC APIs. As a result, the same calls, e.g. `Client.put()` can result in very different (backend exceptions). Our goal is to raise the same exception for either backend.

So with that goal in mind, I imagined a fix where we just allow more optional metadata to be passed during the construction of our current exceptions. For example [`UNKNOWN`][1], [`INTERNAL`][2] and [`DATA_LOSS`][3] all map to 500 (`InternalServerError`), which loses information. But if we code get more info, the data could be preserved, e.g.

```python
>>> try:
...     do_something()
... except InternalServerError as exc:
...     print(exc.grpc_code)
...
<StatusCode.UNAVAILABLE: (14, 'unavailable')>
```

[1]: https://github.com/googleapis/googleapis/blob/f5003093e203f24dcd72769cd638b20110a433f1/google/rpc/code.proto#L50
[2]: https://github.com/googleapis/googleapis/blob/f5003093e203f24dcd72769cd638b20110a433f1/google/rpc/code.proto#L163
[3]: https://github.com/googleapis/googleapis/blob/f5003093e203f24dcd72769cd638b20110a433f1/google/rpc/code.proto#L178

bendemaree:
> Our goal is to catch/wrap and raise the same exception for either backend.

This goal seems very odd to me. You have two different _transports_, not two different backends. Is it really true that the same operation causes different errors to be raised, assuming the error is bubbled up completely? One transport's status code semantics (HTTP) constrain you in a way that causes ambiguity, which I understand, but as the client library, you are on the hook to disambiguate that in a way that is idiomatic in the language at hand. An exception that has different meaning based on its state _post-instantiation_ is not Pythonic, no?

bendemaree:
Put another way: using a Python client for a service should not make me, a library user, be concerned with whether I'm using HTTP or gRPC to talk to the backend, if they two protocols are speaking to the same backend...and I understand them to be the same server on the other side, despite differing serving stacks.

dhermes:
> a library user, be concerned with whether I'm using HTTP or gRPC to talk to the backend

You're totally right.

The immediate question that comes to my mind is: how much of the gRPC type information can we get from the equivalent HTTP errors (either as headers or from the payload)? Do you have an example failure in mind? (If we stick to `datastore`, then the response payload on error is literally a protobuf, so it's not really apples-apples.)

bendemaree:
I thought even the HTTP responses, despite having an HTTP status code, had the RCP-like exception available. In [these docs](https://cloud.google.com/datastore/docs/concepts/errors) under "Here's an example of an error response for a JSON request" it shows `"status": "INVALID_ARGUMENT"` which should correspond to one of the error constants from the protobuf linked earlier...unless I'm misunderstanding your question. 😄

bendemaree:
@dhermes just following up on this. To answer your question more directly, the docs are actually clearer than the snippet I grabbed:

> The response object contains a single field error whose value contains the following elements:

> | Element	| Description |
> | - | - |
> | code	| An HTTP status code that generically identifies the request failure. |
> | message	| Specific information about the request failure.
> | status	| The canonical error code (`google.rpc.Code`) for Google APIs. Codes that may be returned by the Cloud Datastore API are listed in Error Codes. |

Essentially, the current implementation only maps exceptions from `code` where it ought to use `status`. This format looks like it carries over to other Cloud* APIs as well.

lukesneeringer:
Just adding a general comment on this:

> Put another way: using a Python client for a service should not make me, a library user, be concerned with whether I'm using HTTP or gRPC to talk to the backend, if the two protocols are speaking to the same backend...and I understand them to be the same server on the other side, despite differing serving stacks.

I think this is the major thrust of @dhermes' point, but I think there are two ways to resolve it:

Here is a thought experiment: Say I wrote some Python code to use datastore, and am doing so in a hosting environment where (for whatever reason) GRPC is unavailable. So, I am using the HTTP backend. I get an exception for `DATA_LOSS`, and I write code in my application to catch the exception and take some appropriate action.

Now, say I migrate that code to a new hosting environment where GRPC is available. Installing the same packages successfully installs GRPC. We want that code to continue to work, as-is. This is why it matters that the exceptions we raise be the same between transports, precisely so the application developer does not need to be concerned with it and the code is portable.

That said, it does look like the HTTP errors provide sufficient information to disambiguate, so it _is_ possible to do what you are suggesting with both transports.

bendemaree:
@lukesneeringer Yes, I fully understand; proper exceptions keep things portable and abstract over backend, transport, environment, etc. The problem is that the exceptions are ambiguous, not that they're environment-specific. My frustration with discussing this issue is that it has nothing to do with whether we're using the APIs over HTTP or gRPC, both because "it shouldn't" (that is, it is proper for it not to matter) and because the RPC status codes are present in both responses...the transport is a non-issue IMO.

What is an issue is that `UNKNOWN`, `INTERNAL`, and `DATA_LOSS` _still_ are all mashed into `InternalServerError`, and `ALREADY_EXISTS` and `ABORTED` are still mashed into `Conflict`, an issue we first raised in mid-October. We are forced to remain on `gcloud>=0.18,<0.19` where we can properly handle the different error cases. If you are wondering why it's so important (why we don't just retry things that aren't supposed to be retried, say), try opening a ticket regarding an elevated Cloud Datastore error rate; the first thing we are asked has been to verify our retry behavior. If it's incorrect in any way, we have no case.

We are on the verge of having to maintain a fork that disambiguates the error cases due to dead-ends and non-responsiveness on the multiple issues that are open. Is there any path forward with this code, or should we abandon hope and consider this closed?

dhermes:
@bendemaree You converted me to "Team Ben" many weeks ago, just haven't had the cycles to get these changes in. Sorry for the frustration.

daspecster:
@dhermes what all is left for do for this?

dhermes:
@daspecster Potentially a lot. The exceptions defined in this PR are a teensy bit too specific to the transport. What we really need is a ground-up rethinking of the exceptions that can be transport agnostic, but won't be too novel that it confuses people used to HTTP error codes.

bendemaree:
@dhermes That's alright, thanks for circling back. 😄 I am more than happy to help wherever I can be useful; just say the word.

lukesneeringer:
This is superceded by #3738. Closing for now, but reopen if there is anything not covered there.

PR #2937, opened 2017-01-17
Vision: Add gRPC support for image properties.

daspecster:
~~Based from #2922~~

Adds gRPC support for image properties detection.

### DIFF
--- a/system_tests/vision.py
+++ b/system_tests/vision.py
@@ -474,13 +474,13 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
             value.delete()
 
     def _assert_color(self, color):
-        self.assertIsInstance(color.red, int)
-        self.assertIsInstance(color.green, int)
-        self.assertIsInstance(color.blue, int)
+        self.assertIsInstance(color.red, float)
+        self.assertIsInstance(color.green, float)
+        self.assertIsInstance(color.blue, float)
+        self.assertIsInstance(color.alpha, float)
         self.assertNotEqual(color.red, 0.0)
         self.assertNotEqual(color.green, 0.0)
         self.assertNotEqual(color.blue, 0.0)
-        self.assertIsInstance(color.alpha, float)
 
     def _assert_properties(self, image_property):
         from google.cloud.vision.color import ImagePropertiesAnnotation
@@ -493,19 +493,13 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
             self.assertNotEqual(color_info.score, 0.0)
 
     def test_detect_properties_content(self):
-        self._pb_not_implemented_skip(
-            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
         properties = image.detect_properties()
-        self.assertEqual(len(properties), 1)
-        image_property = properties[0]
-        self._assert_properties(image_property)
+        self._assert_properties(properties)
 
     def test_detect_properties_gcs(self):
-        self._pb_not_implemented_skip(
-            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         bucket_name = Config.TEST_BUCKET.name
         blob_name = 'faces.jpg'
@@ -518,16 +512,10 @@ class TestVisionClientImageProperties(BaseVisionTestCase):
 
         image = client.image(source_uri=source_uri)
         properties = image.detect_properties()
-        self.assertEqual(len(properties), 1)
-        image_property = properties[0]
-        self._assert_properties(image_property)
+        self._assert_properties(properties)
 
     def test_detect_properties_filename(self):
-        self._pb_not_implemented_skip(
-            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         image = client.image(filename=FACE_FILE)
         properties = image.detect_properties()
-        self.assertEqual(len(properties), 1)
-        image_property = properties[0]
-        self._assert_properties(image_property)
+        self._assert_properties(properties)
--- a/vision/google/cloud/vision/annotations.py
+++ b/vision/google/cloud/vision/annotations.py
@@ -14,6 +14,7 @@
 
 """Annotations management for Vision API responses."""
 
+import six
 
 from google.cloud.vision.color import ImagePropertiesAnnotation
 from google.cloud.vision.entity import EntityAnnotation
@@ -86,11 +87,11 @@ class Annotations(object):
         :rtype: :class:`~google.cloud.vision.annotations.Annotations`
         :returns: An instance of ``Annotations`` with detection types loaded.
         """
-        annotations = {}
-        for feature_type, annotation in response.items():
-            curr_feature = annotations.setdefault(_KEY_MAP[feature_type], [])
-            curr_feature.extend(
-                _entity_from_response_type(feature_type, annotation))
+        annotations = {
+            _KEY_MAP[feature_type]: _entity_from_response_type(
+                feature_type, annotation)
+            for feature_type, annotation in six.iteritems(response)
+        }
         return cls(**annotations)
 
     @classmethod
@@ -123,12 +124,14 @@ def _process_image_annotations(image):
         'labels': _make_entity_from_pb(image.label_annotations),
         'landmarks': _make_entity_from_pb(image.landmark_annotations),
         'logos': _make_entity_from_pb(image.logo_annotations),
+        'properties': _make_image_properties_from_pb(
+            image.image_properties_annotation),
         'texts': _make_entity_from_pb(image.text_annotations),
     }
 
 
 def _make_entity_from_pb(annotations):
-    """Create an entity from a gRPC response.
+    """Create an entity from a protobuf response.
 
     :type annotations:
     :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.EntityAnnotation`
@@ -141,7 +144,7 @@ def _make_entity_from_pb(annotations):
 
 
 def _make_faces_from_pb(faces):
-    """Create face objects from a gRPC response.
+    """Create face objects from a protobuf response.
 
     :type faces:
     :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.FaceAnnotation`
@@ -151,6 +154,20 @@ def _make_faces_from_pb(faces):
     :returns: List of ``Face``.
     """
     return [Face.from_pb(face) for face in faces]
+
+
+def _make_image_properties_from_pb(image_properties):
+    """Create ``ImageProperties`` object from a protobuf response.
+
+    :type image_properties: :class:`~google.cloud.grpc.vision.v1.\
+                            image_annotator_pb2.ImagePropertiesAnnotation`
+    :param image_properties: Protobuf instance of
+                             ``ImagePropertiesAnnotation``.
+
+    :rtype: list or ``None``
+    :returns: List of ``ImageProperties`` or ``None``.
+    """
+    return ImagePropertiesAnnotation.from_pb(image_properties)
 
 
 def _entity_from_response_type(feature_type, results):
@@ -168,8 +185,7 @@ def _entity_from_response_type(feature_type, results):
         detected_objects.extend(
             Face.from_api_repr(face) for face in results)
     elif feature_type == _IMAGE_PROPERTIES_ANNOTATION:
-        detected_objects.append(
-            ImagePropertiesAnnotation.from_api_repr(results))
+        return ImagePropertiesAnnotation.from_api_repr(results)
     elif feature_type == _SAFE_SEARCH_ANNOTATION:
         detected_objects.append(SafeSearchAnnotation.from_api_repr(results))
     else:
--- a/vision/google/cloud/vision/color.py
+++ b/vision/google/cloud/vision/color.py
@@ -26,20 +26,37 @@ class ImagePropertiesAnnotation(object):
         self._colors = colors
 
     @classmethod
-    def from_api_repr(cls, response):
+    def from_api_repr(cls, image_properties):
         """Factory: construct ``ImagePropertiesAnnotation`` from a response.
 
-        :type response: dict
-        :param response: Dictionary response from Vision API with image
-                         properties data.
+        :type image_properties: dict
+        :param image_properties: Dictionary response from Vision API with image
+                                 properties data.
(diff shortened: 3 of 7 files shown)


system_tests/vision.py
```diff
@@ -478,13 +478,16 @@ def tearDown(self):
...
-        self.assertIsInstance(color.red, int)
-        self.assertIsInstance(color.green, int)
-        self.assertIsInstance(color.blue, int)
+        self.assertIsInstance(color.red, (int, float))
+        self.assertIsInstance(color.green, (int, float))
+        self.assertIsInstance(color.blue, (int, float))
+        if not isinstance(color.alpha, float):
+            self.assertIsInstance(color.alpha.value, float)
+        else:
+            self.assertIsInstance(color.alpha, float)
```
tseaver: Grrr:  I wish the back-ends would just be consistent.  In what world are RGB values floats?

daspecster: Right?...and why is it that just `alpha` is a special `FloatValue` instead of a native `float`.

vision/google/cloud/vision/color.py
```diff
@@ -26,20 +26,38 @@ def __init__(self, colors):
...
+        :param image_properties: Protobuf response from Vision API with image
+                                 properties data.
+
+        :rtype: list of
+                :class:`~google.cloud.vision.color.ImagePropertiesAnnotation`
+        :returns: List of ``ImagePropertiesAnnotation``.
+        """
+        colors = getattr(image_properties.dominant_colors, 'colors', ())
+        if len(colors) == 0:
+            return ()
```
daspecster: If we want a tuple returned instead of an empty list I can do this.

system_tests/vision.py
dhermes: @daspecster Which backend is returned which? ISTM you should be cleaning this before / while creating a color instance?

vision/google/cloud/vision/color.py
dhermes: This breaks "principle" of least surprise. You should either return `None` or an instance of `cls`

vision/google/cloud/vision/color.py
```diff
@@ -26,20 +26,38 @@ def __init__(self, colors):
...
+        return [cls([ColorInformation.from_pb(color) for color in colors])]
```
dhermes: Why a list? Should it just be an instance of `cls`?

vision/google/cloud/vision/color.py
```diff
@@ -93,6 +111,22 @@ def from_api_repr(cls, response):
 
         return cls(red, green, blue, alpha)
 
+    @classmethod
+    def from_pb(cls, color):
+        """Factory: construct a ``Color`` from a protobuf response.
+
+        :type color: :class:`~google.type.color_pb2`
```
dhermes: This is a module, not a class (not that it will show up via Intersphinx either way)

vision/google/cloud/vision/color.py
```diff
@@ -93,6 +111,22 @@ def from_api_repr(cls, response):
...
+        :param color: ``Color`` from API Response.
+
+        :rtype: :class:`~google.cloud.vision.color.Color`
+        :returns: Instance of :class:`~google.cloud.vision.color.Color`.
+        """
+        red = getattr(color, 'red', 0.0)
+        green = getattr(color, 'green', 0.0)
+        blue = getattr(color, 'blue', 0.0)
+        alpha = getattr(color.alpha, 'value', 0.0)
```
dhermes: Why are all of these `getattrs` necessary?

```python
>>> import google.type.color_pb2
>>> C1 = google.type.color_pb2.Color()
>>> C2 = google.type.color_pb2.Color(red=0.5)
>>> import google.protobuf.wrappers_pb2
>>> alpha = google.protobuf.wrappers_pb2.FloatValue(value=0.5)
>>> C3 = google.type.color_pb2.Color(
...     red=1.0, green=0.25, blue=1.0, alpha=alpha)
>>> (C1.red, C2.red, C3.red)
(0.0, 0.5, 1.0)
>>> (C1.green, C2.green, C3.green)
(0.0, 0.0, 0.25)
>>> (C1.blue, C2.blue, C3.blue)
(0.0, 0.0, 1.0)
>>> (C1.alpha.value, C2.alpha.value, C3.alpha.value)
(0.0, 0.0, 0.5)
```

vision/google/cloud/vision/color.py
```diff
@@ -149,19 +183,34 @@ def __init__(self, color, score, pixel_fraction):
...
+        :type color_information: dict
+        :param color_information: Color data with extra meta information.
 
         :rtype: :class:`~google.cloud.vision.color.ColorInformation`
         :returns: Instance of ``ColorInformation``.
         """
-        color = Color.from_api_repr(response.get('color'))
-        score = response.get('score')
-        pixel_fraction = response.get('pixelFraction')
+        color = Color.from_api_repr(color_information.get('color'))
```
dhermes: What happens if `Color.from_api_repr` gets `None`? Don't you want to fall back to a dict?

vision/google/cloud/vision/color.py
```diff
@@ -149,19 +183,34 @@ def __init__(self, color, score, pixel_fraction):
...
+        score = color_information.get('score')
+        pixel_fraction = color_information.get('pixelFraction')
```
dhermes: Is `None` the right default value for these two?

system_tests/vision.py
daspecster: Just the protobuf/gRPC backend. I can't find it now, but there was a little comment block in either the protobuf definition or the generated python code about why it's there. Basically to tell if it's been set or not...if I remember right.

dhermes: But isn't `color` here one of your instances?

daspecster: Ha! Yes it is. #fixed in next commit.

vision/google/cloud/vision/color.py
daspecster: I *think* so because `None` would mean that they were not set.

vision/google/cloud/vision/color.py
daspecster: I can change it, but I would want to change the HTTP version to match at the same time. Should that be a different PR?

vision/google/cloud/vision/color.py
```diff
@@ -26,20 +26,37 @@ def __init__(self, colors):
...
+                :class:`~google.cloud.vision.color.ImagePropertiesAnnotation`
+        :returns: List of ``ImagePropertiesAnnotation``.
         """
-        raw_colors = response.get('dominantColors', {}).get('colors', ())
-        colors = [ColorInformation.from_api_repr(color)
-                  for color in raw_colors]
-        return cls(colors)
+        colors = getattr(image_properties.dominant_colors, 'colors', ())
+        if len(colors) > 0:
+            return [cls([ColorInformation.from_pb(color) for color in colors])]
```
dhermes: > I can change it, but I would want to change the HTTP version to match at the same time. Should that be a different PR?

I am looking at the HTTP version on line 41-42:

```python
        return cls([ColorInformation.from_api_repr(color)
                    for color in colors])
```

It returns an instance, not a list.

daspecster: It turns into a list later on. Good catch though! I updated it in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2937/commits/6e39ce0cbf9bf090d47ae515a3f66d1158dc6913.

system_tests/vision.py
```diff
@@ -474,13 +474,13 @@ def tearDown(self):
             value.delete()
 
     def _assert_color(self, color):
-        self.assertIsInstance(color.red, int)
-        self.assertIsInstance(color.green, int)
-        self.assertIsInstance(color.blue, int)
+        self.assertIsInstance(color.red, (int, float))
+        self.assertIsInstance(color.green, (int, float))
+        self.assertIsInstance(color.blue, (int, float))
```
dhermes: Is `int` still possible?

system_tests/vision.py
```diff
@@ -493,19 +493,14 @@ def _assert_properties(self, image_property):
...
-        self._pb_not_implemented_skip(
-            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
         properties = image.detect_properties()
-        self.assertEqual(len(properties), 1)
-        image_property = properties[0]
-        self._assert_properties(image_property)
+        self.assertEqual(len(properties[0]))
```
dhermes: Did you mean to change the actual output type? This is a "breaking API change" that is not at all related to "Add gRPC support for image properties."

daspecster: I don't think it's a breaking change.
It currently returns a [`list`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/309fa8d6a9268875eb9d7f141342f583b75c5d23/system_tests/vision.py#L503).

Is that what you mean?

daspecster: Never mind I see what you mean. My typo.

system_tests/vision.py
daspecster: Yeah HTTP version returns ints. I could cast it in `Color` though. WDYT?

dhermes: The HTTP version return ints just an artifact of JSON encoding stripping extra data?

```node
> JSON.stringify({a: 1.0})
"{"a":1}"
```

to

```python
>>> import json
>>> json.loads('{"a": 1}')
{u'a': 1}
>>> # vs.
>>> json.loads('{"a": 1.0}')
{u'a': 1.0}
```

Should we cast them to floats?

system_tests/vision.py
```diff
@@ -493,19 +493,14 @@ def _assert_properties(self, image_property):
...
-        self._pb_not_implemented_skip(
-            'gRPC not implemented for image properties detection.')
         client = Config.CLIENT
         with open(FACE_FILE, 'rb') as image_file:
             image = client.image(content=image_file.read())
         properties = image.detect_properties()
         self.assertEqual(len(properties), 1)
-        image_property = properties[0]
-        self._assert_properties(image_property)
+        self._assert_properties(properties[0])
```
dhermes: Not that it matters, but this doesn't need to be in the diff (i.e. the changed code did the exact same thing)

vision/unit_tests/test_color.py
```diff
@@ -45,3 +71,36 @@ def test_missing_rgb_values(self):
...
         self.assertEqual(colors.alpha, 0.0)
+
+
+class TestImagePropertiesAnnotation(unittest.TestCase):
+    @staticmethod
+    def _get_target_class():
+        from google.cloud.vision.color import ImagePropertiesAnnotation
+        return ImagePropertiesAnnotation
+
+    def test_color_annotation_from_pb(self):
```
dhermes: You need a test that covers the case when `None` is returned in `from_pb`

vision/google/cloud/vision/annotations.py
```diff
@@ -89,8 +89,11 @@ def from_api_repr(cls, response):
         annotations = {}
         for feature_type, annotation in response.items():
             curr_feature = annotations.setdefault(_KEY_MAP[feature_type], [])
-            curr_feature.extend(
-                _entity_from_response_type(feature_type, annotation))
+            entity = _entity_from_response_type(feature_type, annotation)
+            if isinstance(entity, list):
+                curr_feature.extend(entity)
+            else:
+                annotations[_KEY_MAP[feature_type]] = entity
```
dhermes: Why was this necessary? I would say:

- Just keep returning a list in `_entity_from_response_type`
- Why did we suddenly grow this feature where `annotations` gains an entry?

daspecster: It didn't gain an entry, it's just handling for setting 'properties' to not be a list like the others so it matches the protobuf version.

dhermes: OK. We should try to clean it up later?

daspecster: Yeah, I'm thinking of handling it more like how I'm doing the pb versions with the helper methods that do the conversions.

WDYT?

system_tests/vision.py
daspecster: Cast to floats https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2937/commits/a39be5fe5a9be88de9cf501856589d1965a3e409

daspecster:
I opend https://github.com/googleapis/googleapis/issues/242 for the RGB values.

vision/unit_tests/test_color.py
daspecster: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2937/commits/8809f2ffa72a3f664ae46418e59f5e12940b8998

dhermes: @daspecster That test doesn't call `ImagePropertiesAnnotation.from_pb`

daspecster: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2937/commits/9d40edbb4daa9ff21451234e08042284e65275fe

daspecster: I updated the names as well to be a little more clear.

vision/google/cloud/vision/annotations.py
dhermes: OK, re-reading this, it seems it could benefit from a helper function and then just be:

```python
annotations = {
    _KEY_MAP[feature_type]: _toms_helper(feature_type, annotation),
    for feature_type, annotation in six.iteritems(response)
}
return cls(**annotations)
```

I **WAS** worried about key collision, but after examining the structure it's clear there wouldn't be one (provided the values in `_KEY_MAP` are unique).

I was also thrown off by the fact that you grab

```python
curr_feature = annotations.setdefault(_KEY_MAP[feature_type], [])
```

independent of the value of `isinstance(entity, list)` (you only need it in the `True` branch).

daspecster: Unless you want me to add to this PR, I'll follow up after this and #2938.

dhermes: Please add to this PR

daspecster: Updated in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2937/commits/bd7fd857439c66e2d01c966c6d0c45d0c6ad99dd.

dhermes:
LGTM

PR #2938, opened 2017-01-17
Vision: Add gRPC support for safe search.

daspecster:
~~Based from #2937~~

There's a few things I don't like about this that I'll comment on. I'm not sure how to solve them right now. Maybe you guys have some ideas?

### DIFF
--- a/docs/vision-usage.rst
+++ b/docs/vision-usage.rst
@@ -224,8 +224,7 @@ categorize the entire contents of the image under four categories.
     >>> client = vision.Client()
     >>> with open('./image.jpg', 'rb') as image_file:
     ...     image = client.image(content=image_file.read())
-    >>> safe_search_results = image.detect_safe_search()
-    >>> safe_search = safe_search_results[0]
+    >>> safe_search = image.detect_safe_search()
     >>> safe_search.adult
     <Likelihood.VERY_UNLIKELY: 'VERY_UNLIKELY'>
     >>> safe_search.spoof
--- a/vision/unit_tests/test_annotations.py
+++ b/vision/unit_tests/test_annotations.py
@@ -67,6 +67,8 @@ class TestAnnotations(unittest.TestCase):
         self.assertEqual(annotations.texts, [True])
 
     def test_from_pb(self):
+        from google.cloud.vision.likelihood import Likelihood
+        from google.cloud.vision.safe_search import SafeSearchAnnotation
         from google.cloud.grpc.vision.v1 import image_annotator_pb2
 
         image_response = image_annotator_pb2.AnnotateImageResponse()
@@ -76,8 +78,15 @@ class TestAnnotations(unittest.TestCase):
         self.assertEqual(annotations.faces, [])
         self.assertEqual(annotations.landmarks, [])
         self.assertEqual(annotations.texts, [])
-        self.assertEqual(annotations.safe_searches, ())
         self.assertIsNone(annotations.properties)
+
+        self.assertIsInstance(annotations.safe_searches, SafeSearchAnnotation)
+        safe_search = annotations.safe_searches
+        unknown = Likelihood.UNKNOWN
+        self.assertIs(safe_search.adult, unknown)
+        self.assertIs(safe_search.spoof, unknown)
+        self.assertIs(safe_search.medical, unknown)
+        self.assertIs(safe_search.violence, unknown)
 
 
 class Test__make_entity_from_pb(unittest.TestCase):
--- a/vision/unit_tests/test_safe_search.py
+++ b/vision/unit_tests/test_safe_search.py
@@ -1,0 +1,70 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+
+class TestSafeSearchAnnotation(unittest.TestCase):
+    @staticmethod
+    def _get_target_class():
+        from google.cloud.vision.safe_search import SafeSearchAnnotation
+        return SafeSearchAnnotation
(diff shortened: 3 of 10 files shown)


vision/unit_tests/test_annotations.py
```diff
@@ -67,17 +67,26 @@ def test_ctor(self):
...
         annotations = self._make_one().from_pb(image_response)
         self.assertEqual(annotations.labels, [])
         self.assertEqual(annotations.logos, [])
-        self.assertEqual(annotations.faces, ())
+        self.assertEqual(annotations.faces, [])
         self.assertEqual(annotations.landmarks, [])
         self.assertEqual(annotations.texts, [])
-        self.assertEqual(annotations.safe_searches, ())
-        self.assertEqual(annotations.properties, ())
+        self.assertEqual(len(annotations.safe_searches), 1)
```
daspecster: On the other protobuf objects, there has been a way for me to tell if there are results or not. In this case, the values for `adult`, `spoof` etc... are defaulting to`0` which is the valid `Likelihood.UNKNOWN` type.

I would have to track the feature type's that are passed in from the top in order to determine if this was a requested feature type or not. There has to be a better way.

daspecster:
@dhermes this should be ready to go. Let me know if you have any feedback.

vision/google/cloud/vision/safe.py
```diff
@@ -54,14 +54,30 @@ def from_api_repr(cls, response):
...
+        :type image: :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.\
+                      SafeSearchAnnotation`
+        :param image: Protobuf response from Vision API with safe search data.
+
+        :rtype: :class:`~google.cloud.vision.safe.SafeSearchAnnotation`
+        :returns: Instance of ``SafeSearchAnnotation``.
+        """
+        classifications = map(_get_pb_likelihood, [image.adult, image.spoof,
+                                                   image.medical,
+                                                   image.violence])
```
dhermes: Style nit: Suggest instead

```python
values = [image.adult, image.spoof, image.medical, image.violence]
classifications = map(_get_pb_likelihood, values)
```

vision/unit_tests/test_annotations.py
```diff
@@ -76,9 +78,16 @@ def test_from_pb(self):
         self.assertEqual(annotations.faces, [])
         self.assertEqual(annotations.landmarks, [])
         self.assertEqual(annotations.texts, [])
-        self.assertEqual(annotations.safe_searches, ())
         self.assertIsNone(annotations.properties)
 
+        self.assertIsInstance(annotations.safe_searches, SafeSearchAnnotation)
+        safe_search = annotations.safe_searches
+        unknown = Likelihood.UNKNOWN
+        self.assertEqual(safe_search.adult, unknown)
```
dhermes: Wouldn't `assertIs` be better for an enum?

vision/unit_tests/test_safe.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2016 Google Inc.
```
dhermes: High-level

1. If the `safe` module already existed, this test should have also already existed. Transitive test coverage is not the same as writing actual tests
1. If `safe` was small enough to get transitive coverage without every directly testing anything in it, you probably didn't need to give it it's on module
1. :-1: to the module name `safe`, it is unclear

vision/unit_tests/test_safe.py
```diff
@@ -0,0 +1,70 @@
...
+
+class TestSafeSearchAnnotation(unittest.TestCase):
+    @staticmethod
+    def _get_target_class():
+        from google.cloud.vision.safe import SafeSearchAnnotation
+        return SafeSearchAnnotation
+
+    def test_safe_search_annotation(self):
+        from google.cloud.vision.likelihood import Likelihood
+        from unit_tests._fixtures import SAFE_SEARCH_DETECTION_RESPONSE
```
dhermes: Do you really need `_fixtures`?

docs/vision-usage.rst
```diff
@@ -224,8 +224,7 @@ categorize the entire contents of the image under four categories.
     >>> client = vision.Client()
     >>> with open('./image.jpg', 'rb') as image_file:
     ...     image = client.image(content=image_file.read())
-    >>> safe_search_results = image.detect_safe_search()
-    >>> safe_search = safe_search_results[0]
+    >>> safe_search = image.detect_safe_search()
```
dhermes: API didn't change, these docs were just stale?

vision/unit_tests/test_safe.py
daspecster: For safe search, it could probably be a local variable. However I like having it all in the same place. Some of the other responses are fairly large. I guess it just depends if you want the tests or not. If so, then we may have to bump the max module lines value again.

vision/unit_tests/test_safe.py
daspecster: I can rename to `safe_search`.

dhermes: Yes please do

vision/unit_tests/test_safe.py
dhermes: Bumping that value doesn't bother me.

> I like having it all in the same place

It may just be a preference difference between us? ISTM that it can be said that individual tests are hard to reason about if the data is not in the test (independent of preference), but maybe I'm wrong?

docs/vision-usage.rst
daspecster: Yup, stale.

vision/unit_tests/test_safe.py
daspecster: Sure, I agree with that. However in this case I don't really like having [this](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/422378126b465ce26795fd159eed625f86d996a0/vision/unit_tests/_fixtures.py#L278-L1688) much in a test case.

I guess I could test it in parts instead of a full response. Then we would be relying more on the system test to make sure that it works in its entirety.

daspecster: @dhermes did you want me to bring the safe search response into the tests in this PR or do them all together in another PR?

dhermes: Whoa that's huge. Don't worry about that change

vision/unit_tests/test_safe_search.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2016 Google Inc.
```
dhermes: s/2016/2017

daspecster:
Updated copyright and squashed.

PR #2942, opened 2017-01-18
Update import spacing in tests.

daspecster:
Updating the import spacing to include one newline between imports and code.

This way when contributors look at other tests they aren't confused with the style.

I'll make another PR with the other half of the services after this one.


dhermes:
Cursory LGTM

PR #2944, opened 2017-01-18
Make note of num_results property on iterators

bendemaree:
There's no mention of this handy property in the core docs, so I thought I'd add an example of its usage.

### DIFF
--- a/core/google/cloud/iterator.py
+++ b/core/google/cloud/iterator.py
@@ -46,7 +46,7 @@ At any point, you may check the number of items consumed by referencing the
 ``num_results`` property of the iterator::
 
     >>> my_iterator = Iterator(...)
-    ... for my_item in my_iterator:
+    >>> for my_item in my_iterator:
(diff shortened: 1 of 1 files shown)


core/google/cloud/iterator.py
```diff
@@ -42,6 +42,14 @@
     ...     if not my_item.is_valid:
     ...         break
 
+At any point, you may check the number of items consumed by referencing the
+``num_results`` property of the iterator::
+
+    >>> my_iterator = Iterator(...)
+    ... for my_item in my_iterator:
```
daspecster: I think this would actually be a `>>>`.

daspecster:
Thanks @bendemaree!

core/google/cloud/iterator.py
bendemaree: Serves me right for eyeballing it. 😉 Thanks!

PR #2950, opened 2017-01-19
Assert type check instead of value for Speech.

daspecster:
The score dropped for Speech and so the assertions were failing. This switches that to just type checking.

See: https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/193242535#L2056


dhermes:
LGTM (sadly)

daspecster:
😢 
I know right...

PR #2952, opened 2017-01-20
Moving Bigtable helpers for duration protobufs into core.

dhermes:
~~Note this is from a private branch of @tseaver. Though it largely copies a function [from Bigtable][1] so I'm not sure how to proceed here.~~

@tseaver Can you weigh in?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/5a42aca77f0400ce2ba89fa70b0b039f913f8af0/bigtable/google/cloud/bigtable/column_family.py#L28

/cc @geigerj


tseaver:
@dhermes It looks like the two differ for negative durations, which means the version in Bigtable is more likely correct.  Why don't we promote that function (and the `duration_pb_to_timedelta` companion) up from Bigtable to `google.cloud._helpers` instead?

tseaver:
> I just went with the existing name: _timedelta_to_duration_pb (vs. _timedelta_to_pb_duration in your impl.). WDYT?

Works for me:  I will adjust the private branch accordingly after merging from master.

PR #2959, opened 2017-01-23
Loosen system tests for Vision Text detection.

daspecster:
This resolves the `text_detect` system test errors in Travis.

See: https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/193935352#L2071-L2073

### DIFF
--- a/system_tests/vision.py
+++ b/system_tests/vision.py
@@ -427,8 +427,8 @@ class TestVisionClientText(unittest.TestCase):
     def _assert_text(self, text):
         self.assertIsInstance(text, EntityAnnotation)
         self.assertIn(text.description, self.DESCRIPTIONS)
-        self.assertIn(text.locale, (None, 'en'))
-        self.assertNotEqual(text.score, 0.0)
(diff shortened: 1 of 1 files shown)


system_tests/vision.py
```diff
@@ -427,8 +427,8 @@ def tearDown(self):
     def _assert_text(self, text):
         self.assertIsInstance(text, EntityAnnotation)
         self.assertIn(text.description, self.DESCRIPTIONS)
-        self.assertIn(text.locale, (None, 'en'))
-        self.assertNotEqual(text.score, 0.0)
+        self.assertIn(text.locale, (None, '', 'en'))
+        self.assertIsInstance(text.score, (type(None), float))
```
tseaver: Are there back-end bugs for these inconsitencies?

daspecster: I'm guessing that's what's going on. It happened to Speech a while ago too.
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2935

It seems like it's always `score`.

PR #2961, opened 2017-01-24
Fix KeyError for unsupported annotation types.

### LINKED ISSUE #2930: Client library should not fail when encountering unexpected response bodies
1. OS type and version
OSX 10.11

2. Python version and virtual environment information `python --version`
Python 2.7.12

3. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`
google-cloud-vision==0.22.0

4. Stacktrace if available
```

cls = <class 'google.cloud.vision.annotations.Annotations'>
response = {'fullTextAnnotation': {'pages': [{'blocks': [{'blockType': 'TEXT', 'boundingBox': {...}, 'paragraphs': [...], 'proper... [{'x': 317, 'y': 506}, {'x': 337, 'y': 507}, {'x': 335, 'y': 536}, {'x': 315, 'y': 535}]}, 'description': 'to'}, ...]}

    @classmethod
    def from_api_repr(cls, response):
        """Factory: construct an instance of ``Annotations`` from a response.

            :type response: dict
            :param response: Vision API response object.

            :rtype: :class:`~google.cloud.vision.annotations.Annotations`
            :returns: An instance of ``Annotations`` with detection types loaded.
            """
        annotations = {}
        for feature_type, annotation in response.items():
>           curr_feature = annotations.setdefault(_KEY_MAP[feature_type], [])
E           KeyError: u'fullTextAnnotation'

env/lib/python2.7/site-packages/google/cloud/vision/annotations.py:91: KeyError
```

5. Steps to reproduce
* Enable API backend with extra data in response to detect_text
* Call detect_text
* Extra content in response body crashes the client library

6. Code example

The following code exercises the crash:
```
def detect_text(path):
    """Detects text in the file."""
    vision_client = vision.Client()

    with io.open(path, 'rb') as image_file:
        content = image_file.read()

    image = vision_client.image(content=content)

    texts = image.detect_text()
    print('Texts:')
    for text in texts:
        print(text.description)
```

daspecster:
Fixes #2930.

### DIFF
--- a/vision/google/cloud/vision/annotations.py
+++ b/vision/google/cloud/vision/annotations.py
@@ -87,13 +87,11 @@ class Annotations(object):
         :rtype: :class:`~google.cloud.vision.annotations.Annotations`
         :returns: An instance of ``Annotations`` with detection types loaded.
         """
-        for feature_type in response.keys():
-            if feature_type not in _KEY_MAP:
-                del response[feature_type]
         annotations = {
             _KEY_MAP[feature_type]: _entity_from_response_type(
                 feature_type, annotation)
             for feature_type, annotation in six.iteritems(response)
+            if feature_type in _KEY_MAP
         }
         return cls(**annotations)
 
--- a/vision/google/cloud/vision/annotations.py
+++ b/vision/google/cloud/vision/annotations.py
@@ -87,6 +87,9 @@ class Annotations(object):
         :rtype: :class:`~google.cloud.vision.annotations.Annotations`
         :returns: An instance of ``Annotations`` with detection types loaded.
         """
+        for feature_type in response.keys():
+            if feature_type not in _KEY_MAP:
+                del response[feature_type]
         annotations = {
             _KEY_MAP[feature_type]: _entity_from_response_type(
                 feature_type, annotation)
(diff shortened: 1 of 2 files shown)


vision/google/cloud/vision/annotations.py
```diff
@@ -87,6 +87,9 @@ def from_api_repr(cls, response):
         :rtype: :class:`~google.cloud.vision.annotations.Annotations`
         :returns: An instance of ``Annotations`` with detection types loaded.
         """
+        for feature_type in response.keys():
```
daspecster: Not sure I really like this. But it seemed like the most direct strategy. Let me know if you have a preferred way to do this.

tseaver: You could add the filtering in the dictionary comprehension, which would avoid destroying information in the response.  E.g.:

```python
         annotations = {
             _KEY_MAP[feature_type]: _entity_from_response_type(
                 feature_type, annotation)
             for feature_type, annotation in six.iteritems(response)
            
}
```

dhermes:
@jonparrott @gguuss PTAL. (I can already image the bug reports: "where is my data")

lukesneeringer:
I am all kinds of nervous about just dropping extra fields, but it looks like #2930 has already covered this ground. So, okay.

daspecster:
Bump!

jonparrott:
This is fine with me. I believe this is what @gguuss had in mind - the data isn't there, but the library doesn't explode.

daspecster:
Thanks @jonparrott!

PR #2962, opened 2017-01-25
Cloud Speech Veneer does not raise for > 1 result.

### LINKED ISSUE #2947: Cloud Speech veneer should not throw an exception if there is more than one result.
I believe this code in the veneer is incorrect because it throws if there is more than one result.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/speech/google/cloud/speech/_gax.py#L263

As documented in the proto, the results contain a sequential list of transcription results corresponding to sequential portions of audio.

lukesneeringer:
This commit also:

  * adds `transcript` and `confidence` convenience methods
    on result objects
  * updates unit tests appropriately
  * moves from google.cloud.grpc.* to google.cloud.proto.*
  * updates dependencies accordingly

Fixes #2947.

### DIFF
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -16,13 +16,9 @@
 
 
 from google.cloud.gapic.speech.v1beta1.speech_client import SpeechClient
-from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import RecognitionAudio
-from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import RecognitionConfig
-from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import SpeechContext
 from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
-    StreamingRecognitionConfig)
-from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
-    StreamingRecognizeRequest)
+    RecognitionAudio, RecognitionConfig, SpeechContext,
+    StreamingRecognitionConfig, StreamingRecognizeRequest)
 from google.longrunning import operations_grpc
 
 from google.cloud._helpers import make_secure_channel
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -25,7 +25,6 @@ from google.cloud._helpers import make_secure_channel
 from google.cloud._helpers import make_secure_stub
 from google.cloud._http import DEFAULT_USER_AGENT
 
-from google.cloud.speech.alternative import Alternative
 from google.cloud.speech.operation import Operation
 from google.cloud.speech.result import Result
 
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -230,7 +230,8 @@ class GAPICSpeechAPI(object):
                                and phrases. This can also be used to add new
                                words to the vocabulary of the recognizer.
 
-        :yields: :class:`google.cloud.speech.result.Result` objects.
+        :rtype: list
+        :returns: List of :class:`google.cloud.speech.result.Result` objects.
 
         :raises: ValueError if there are no results.
         """
@@ -249,8 +250,7 @@ class GAPICSpeechAPI(object):
             raise ValueError('No results returned from the Speech API.')
 
         # Iterate over any results that came back.
-        for result in api_response.results:
-            yield Result.from_pb(result)
+        return [Result.from_pb(i) for i in api_response.results]
 
 
 def _stream_requests(sample, language_code=None, max_alternatives=None,
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -250,7 +250,7 @@ class GAPICSpeechAPI(object):
             raise ValueError('No results returned from the Speech API.')
 
         # Iterate over any results that came back.
-        return [Result.from_pb(i) for i in api_response.results]
+        return [Result.from_pb(result) for result in api_response.results]
 
 
 def _stream_requests(sample, language_code=None, max_alternatives=None,
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -16,9 +16,15 @@
 
 
 from google.cloud.gapic.speech.v1beta1.speech_client import SpeechClient
+from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import RecognitionAudio
 from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
-    RecognitionAudio, RecognitionConfig, SpeechContext,
-    StreamingRecognitionConfig, StreamingRecognizeRequest)
+    RecognitionConfig)
+from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
+    SpeechContext)
+from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
+    StreamingRecognitionConfig)
+from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
+    StreamingRecognizeRequest)
 from google.longrunning import operations_grpc
 
 from google.cloud._helpers import make_secure_channel
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -230,16 +230,9 @@ class GAPICSpeechAPI(object):
                                and phrases. This can also be used to add new
                                words to the vocabulary of the recognizer.
 
-        :rtype: list
-        :returns: A list of dictionaries. One dict for each alternative. Each
-                  dictionary typically contains two keys (though not
-                  all will be present in all cases)
+        :yields: :class:`google.cloud.speech.result.Result` objects.
 
-                  * ``transcript``: The detected text from the audio recording.
-                  * ``confidence``: The confidence in language detection, float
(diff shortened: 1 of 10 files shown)


speech/google/cloud/speech/operation.py
```diff
@@ -58,11 +58,13 @@ def _update_state(self, operation_pb):
...
-            raise ValueError('Expected exactly one result, found:',
-                             pb_results)
+        if len(pb_results) == 0:
+            raise ValueError('Speech API returned no results.')
 
-        result = pb_results[0]
-        self.results = [Alternative.from_pb(alternative)
-                        for alternative in result.alternatives]
+        # Save the results to the Operation object.
+        self.results = []
```
daspecster: Since this is getting set to `[]` at the get go anyway, his could just be...
```
self.results = [Result.from_pb(pb_result) for pb_result in pb_results]
```

speech/google/cloud/speech/result.py
```diff
@@ -17,6 +17,53 @@
 from google.cloud.speech.alternative import Alternative
 
 
+class Result(object):
+    """Speech recognition result representation.
+    This is the object that comes back on sync or async requests
```
daspecster: Can you add a space between the opening line and the continuation?

```diff
-    """Speech recognition result representation.
-    This is the object that comes back on sync or async requests
+    """Speech recognition result representation.
+
+    This is the object that comes back on sync or async requests
```

speech/google/cloud/speech/result.py
```diff
@@ -17,6 +17,53 @@
...
+        """Factory: construct instance of ``SpeechRecognitionResult``.
+
+        :type result: :class:`~google.cloud.grpc.speech.v1beta1\
+                               .cloud_speech_pb2.StreamingRecognizeResult`
+        :param result: Instance of ``StreamingRecognizeResult`` protobuf.
+
+        :rtype: :class:`~google.cloud.speech.result.SpeechRecognitionResult`
+        :returns: Instance of ``SpeechRecognitionResult``.
+        """
+        alternatives = []
```
daspecster: Ditto on list comprehension.

Also could you change `alt` to `alternative`.

speech/google/cloud/speech/result.py
```diff
@@ -17,6 +17,53 @@
...
+        return cls(alternatives=alternatives)
+
+    @property
+    def confidence(self):
+        """Return the confidence for the most probable alternative.
+
+        :rtype: float
+        :returns: Confidence value, between 0 and 1.
+        """
+        return self.alternatives[0].confidence
```
daspecster: This would return an `IndexError` if there were not results right?

We should probably handle that more gracefully if possible.

speech/google/cloud/speech/result.py
```diff
@@ -46,9 +93,28 @@ def from_pb(cls, response):
         :rtype: :class:`~google.cloud.speech.result.StreamingSpeechResult`
         :returns: Instance of ``StreamingSpeechResult``.
         """
-        alternatives = [Alternative.from_pb(alternative)
-                        for alternative in response.alternatives]
+        alternatives = []
```
daspecster: Ah, I see here you changed it away from a list comprehension. Again just curious why.
Also can we change `/alt/alternative/`?

speech/unit_tests/test_client.py
```diff
@@ -326,16 +326,19 @@ def speech_api(channel=None):
...
+        self.assertEqual(len(results), 1)
+        self.assertEqual(len(results[0].alternatives), 2)
         self.assertEqual(results[0].transcript,
+                         results[0].alternatives[0].transcript,
                          alternatives[0]['transcript'])
         self.assertEqual(results[0].confidence,
+                         results[0].alternatives[0].confidence,
                          alternatives[0]['confidence'])
-        self.assertEqual(results[1].transcript,
+        self.assertEqual(results[0].alternatives[1].transcript,
```
daspecster: This implies an update to `vision-usage.rst` right?

speech/google/cloud/speech/operation.py
lukesneeringer: There is a bug in coverage/py.test that makes it think that the list comprehension is not actually running. :-/

daspecster:
@lukesneeringer you can ping me on hangouts if you want and we can discuss.

speech/google/cloud/speech/result.py
lukesneeringer: I did that to get around [this bug](https://bitbucket.org/ned/coveragepy/issues/515/coverage-of-multiline-generator-set-dict).

speech/google/cloud/speech/operation.py
daspecster: Oh :'(.

daspecster: /cc @dhermes

speech/google/cloud/speech/_gax.py
```diff
@@ -254,13 +250,14 @@ def sync_recognize(self, sample, language_code=None, max_alternatives=None,
...
-        else:
-            raise ValueError('More than one result or none returned from API.')
+
+        # Sanity check: If we got no results back, raise an error.
+        if len(api_response.results) == 0:
+            raise ValueError('No results returned from the Speecn API.')
+
+        # Iterate over any results that came back.
+        for result in api_response.results:
+            yield Result.from_pb(result)
```
lukesneeringer: It was returning a single, flattened result, which was the major bug this PR fixes. I could return a flat list instead, I suppose.

speech/google/cloud/speech/result.py
lukesneeringer: We already raise `ValueError` before you can get to that point...but this could/should still be better.

speech/google/cloud/speech/result.py
lukesneeringer: :+1:

speech/google/cloud/speech/_gax.py
```diff
@@ -254,13 +250,14 @@ def sync_recognize(self, sample, language_code=None, max_alternatives=None,
...
-            results = api_response.results.pop()
-            alternatives = results.alternatives
-            return [Alternative.from_pb(alternative)
-                    for alternative in alternatives]
-        else:
-            raise ValueError('More than one result or none returned from API.')
+
+        # Sanity check: If we got no results back, raise an error.
+        if len(api_response.results) == 0:
+            raise ValueError('No results returned from the Speecn API.')
```
lukesneeringer: :+1:

speech/google/cloud/speech/operation.py
daspecster: @lukesneeringer it looks like that's only for multiline list comprehensions. Also, were they saying that `--assert=plain` resolves the issue for multiline list comprehensions for now?

speech/unit_tests/test__gax.py
```diff
@@ -36,13 +36,8 @@ def _call_fut(self, sample, language_code, max_alternatives,
...
         from google.cloud.speech.sample import Sample
-        from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
-            SpeechContext)
-        from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
-            RecognitionConfig)
-        from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
-            StreamingRecognitionConfig)
-        from google.cloud.grpc.speech.v1beta1.cloud_speech_pb2 import (
+        from google.cloud.proto.speech.v1beta1.cloud_speech_pb2 import (
+            RecognitionConfig, SpeechContext, StreamingRecognitionConfig,
```
daspecster: I think it's a Google style thing to have separated imports.

lukesneeringer: I checked the internal style guide, and it does allow this (but prefers importing `cloud_speech_pb2` and using everything off of that, which I am not going to change because of the unwieldily-long names).

speech/unit_tests/test_client.py
lukesneeringer: This is Speech, not Vision. Also, what RST file are you referring to?

daspecster: Sorry!  I'm working on vision today and my brain didn't want to flip over to speech apparently.

speech/unit_tests/test__gax.py
daspecster: @dhermes do you remember where this style decision stems from?

dhermes: https://google.github.io/styleguide/pyguide.html#Imports_formatting

daspecster: Thanks @dhermes! I always end up looking in https://google.github.io/styleguide/pyguide.html#Imports instead.

speech/google/cloud/speech/result.py
daspecster: You could check for the `isFinal` alternative and return that confidence otherwise return the top result's confidence.

See: https://cloud.google.com/speech/reference/rest/v1beta1/speech/syncrecognize#SpeechRecognitionAlternative

speech/google/cloud/speech/_gax.py
```diff
@@ -249,8 +250,7 @@ def sync_recognize(self, sample, language_code=None, max_alternatives=None,
             raise ValueError('No results returned from the Speech API.')
 
         # Iterate over any results that came back.
-        for result in api_response.results:
-            yield Result.from_pb(result)
+        return [Result.from_pb(i) for i in api_response.results]
```
daspecster: We try to stay away from one letter vars even in list comprehensions.
`/i/result/`

PR #2963, opened 2017-01-25
Remove default sample_rate.

### LINKED ISSUE #2948: Cloud speech veneer should not provide a default sample rate
Default sample rates should be removed.  For audio types with a header the server can automatically determine the sample rate. (We are implementing this for V1.)  For audio types without a header the default causes more harm than good because if it is incorrect the results could be empty or wildly inaccurate, which causes user confusion.

lukesneeringer:
Fixes #2948.

### DIFF
--- a/speech/google/cloud/speech/sample.py
+++ b/speech/google/cloud/speech/sample.py
@@ -53,7 +53,6 @@ class Sample(object):
     :param client: (Optional) The client that owns this instance of sample.
     """
     default_encoding = Encoding.FLAC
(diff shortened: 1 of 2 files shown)


daspecster:
@lukesneeringer do the systems tests pass for this? I believe they should but I just wanted to double check.

lukesneeringer:
Yes.

daspecster:
Merging on behalf of @lukesneeringer.

PR #2965, opened 2017-01-26
Fix Speech LRO handling and add HTTP side for multiple results. 

daspecster:
This is an update to fill out the HTTP side of #2962 and I have a hacky but effective way to handle the new LRO.

I tried using the `OperationsClient` but for some reason it just kept 404ing. I'm not sure what I was doing wrong.

/cc @lukesneeringer

### DIFF
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -106,7 +106,7 @@ class GAPICSpeechAPI(object):
         api = self._gapic_api
         operation_future = api.async_recognize(config=config, audio=audio)
 
-        return Operation.from_pb(operation_future._operation, self)
+        return Operation.from_pb(operation_future.last_operation_data(), self)
 
     def streaming_recognize(self, sample, language_code=None,
                             max_alternatives=None, profanity_filter=None,
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -104,9 +104,9 @@ class GAPICSpeechAPI(object):
         audio = RecognitionAudio(content=sample.content,
                                  uri=sample.source_uri)
         api = self._gapic_api
-        response = api.async_recognize(config=config, audio=audio)
+        operation_future = api.async_recognize(config=config, audio=audio)
 
-        return Operation.from_pb(response, self)
+        return Operation.from_pb(operation_future._operation, self)
 
     def streaming_recognize(self, sample, language_code=None,
                             max_alternatives=None, profanity_filter=None,
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -64,9 +64,10 @@ See: `Speech Asynchronous Recognize`_
     >>> operation.complete
     True
     >>> for result in operation.results:
-    ...     print('=' * 20)
-    ...     print(result.transcript)
-    ...     print(result.confidence)
+    ...     for alternative in result.alternatives:
+    ...         print('=' * 20)
+    ...         print(alternative.transcript)
+    ...         print(alternative.confidence)
     ====================
     'how old is the Brooklyn Bridge'
     0.98267895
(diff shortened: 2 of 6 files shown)


speech/google/cloud/speech/_gax.py
```diff
@@ -104,9 +104,9 @@ def async_recognize(self, sample, language_code=None,
         audio = RecognitionAudio(content=sample.content,
                                  uri=sample.source_uri)
         api = self._gapic_api
-        response = api.async_recognize(config=config, audio=audio)
+        operation_future = api.async_recognize(config=config, audio=audio)
 
-        return Operation.from_pb(response, self)
+        return Operation.from_pb(operation_future._operation, self)
```
daspecster: The hacky bit...

lukesneeringer:
Ideally, I would prefer the user get the `OperationFuture` object from GAX, and would like to look into why @daspecster is having trouble making it work. I will try to do that in the next day or so.

That said, merging this is almost certainly preferable to (1) reverting #2962 (which fixes an item the Speech team identified as critical or (2) leaving async broken. :-)

daspecster:
@eoogbe I was getting a Rendezvous 404 when I tried to use `OperationsClient.get_operation()`.
I pretty much followed the example, but I had to add the scope.

In this PR I'm just pulling [`_OperationFuture._operation`](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2965#discussion-diff-98087698R109) but I'm sure that's not the intended way it should be used. 

Any help would be greatly appreciated!

speech/google/cloud/speech/_gax.py
eoogbe: `_OperationFuture` has a `last_operation_data` method. I think that's what you're looking for.

daspecster: Ok thanks!
I was thrown off a little bit because of the underscore prefix of `_OperationFuture`.  So that is meant to be used directly and not through `OperationsClient`?

eoogbe: Yes, it's meant to be used directly, but not instantiated directly. The underscore prefix is supposed to indicate that the constructor is private.

daspecster:
@lukesneeringer the `_OperationFuture` usage is updated. PTAL

speech/google/cloud/speech/_gax.py
```diff
@@ -104,9 +104,9 @@ def async_recognize(self, sample, language_code=None,
         audio = RecognitionAudio(content=sample.content,
                                  uri=sample.source_uri)
         api = self._gapic_api
-        response = api.async_recognize(config=config, audio=audio)
+        operation_future = api.async_recognize(config=config, audio=audio)
 
-        return Operation.from_pb(response, self)
+        return Operation.from_pb(operation_future.last_operation_data(), self)
```
lukesneeringer: Why is this returning `Operation.from_pb(operation_future.last_operation_data())` and not just `operation_future`? The _entire point_ of the `_OperationFuture` class is that it has a `.result()` method and such to get the result (while `Operation` objects do not and have to be fed to a client.)

daspecster: To keep the library's surface the same for HTTP and GAX.

lukesneeringer: Okay, I see the problem. I will talk to @bjwatson about this tomorrow. I _really_ want this to return an `_OperationFuture`, which means the HTTP surface needs to return something with a similar interface.

daspecster: Ok, yeah there's an [issue](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2694) about adding retry as a whole.

I'm not sure if having it in GAPIC makes this easier or harder to do.

bjwatson: Hmm... interesting point @daspecster. We might want to leave this as-is for now, until we can provide an `_OperationFuture` wrapper for HTTP/1.1 in `google-gax`, unless you can @lukesneeringer can think of a better solution. We don't want to break the code re-use you have between the two transports.

daspecster:
@lukesneeringer do you have anything else for this?

lukesneeringer:
Yeah, I am writing an `_OperationFuture` clone to go into `google.cloud.core` this morning.

lukesneeringer:
Okay, I am aborting that for now. The rabbit hole is going too far. Yes, you can merge this.

PR #2966, opened 2017-01-26
Pubsub batch autocommitting.

lukesneeringer:
This PR adds some functionality to the Batch object:

  * The ability to specify `max_messages` and have the batch
    automatically call `commit` when the number of messages
    gets that high.
  * The ability to specify `max_interval` and have the batch
    automatically commit when a publish occurs and the batch
    is at least as old as the specified interval.

This is one of two changes requested by the PubSub team.

### DIFF
--- a/pubsub/google/cloud/pubsub/topic.py
+++ b/pubsub/google/cloud/pubsub/topic.py
@@ -14,6 +14,8 @@
 
 """Define API Topics."""
 
+import time
+
 from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud._helpers import _NOW
 from google.cloud.exceptions import NotFound
@@ -408,14 +410,39 @@ class Batch(object):
     :type topic: :class:`google.cloud.pubsub.topic.Topic`
     :param topic: the topic being published
 
-    :type client: :class:`google.cloud.pubsub.client.Client`
     :param client: The client to use.
+    :type client: :class:`google.cloud.pubsub.client.Client`
+
+    :param max_interval: The maximum interval, in seconds, before the batch
+                         will automatically commit. Note that this does not
+                         run a background loop; it just checks when each
+                         message is published. Therefore, this is intended
+                         for situations where messages are published at
+                         reasonably regular intervals. Defaults to infinity
+                         (off).
+    :type max_interval: float
+
+    :param max_messages: The maximum number of messages to hold in the batch
+                         before automatically commiting. Defaults to infinity
+                         (off).
+    :type max_messages: float
     """
-    def __init__(self, topic, client):
+    INFINITY = float('inf')
+
+    def __init__(self, topic, client, max_interval=INFINITY,
+                 max_messages=INFINITY):
         self.topic = topic
         self.messages = []
         self.message_ids = []
         self.client = client
+
+        # Set the autocommit rules. If the interval or number of messages
+        # is exceeded, then the .publish() method will imply a commit.
+        self._max_interval = max_interval
+        self._max_messages = max_messages
+
+        # Set the initial starting timestamp (used against the interval).
+        self._start_timestamp = float(time.time())
 
     def __enter__(self):
         return self
@@ -435,11 +462,26 @@ class Batch(object):
 
         :type attrs: dict (string -> string)
         :param attrs: key-value pairs to send as message attributes
+
+        :rtype: None
+        :returns: None
         """
         self.topic._timestamp_message(attrs)
         self.messages.append(
             {'data': message,
              'attributes': attrs})
+
+        # If too much time has elapsed since the first message
+        # was added, autocommit.
+        if self._max_interval < self.INFINITY:
+            if float(time.time()) - self._start_timestamp > self._max_interval:
+                self._start_timestamp = float(time.time())
+                return self.commit()
+
+        # If the number of messages on the list is greater than the
+        # maximum allowed, autocommit (with the batch's client).
+        if len(self.messages) >= self._max_messages:
+            return self.commit()
 
     def commit(self, client=None):
         """Send saved messages as a single API call.
--- a/pubsub/google/cloud/pubsub/topic.py
+++ b/pubsub/google/cloud/pubsub/topic.py
@@ -427,10 +427,10 @@ class Batch(object):
                          (off).
     :type max_messages: float
     """
-    INFINITY = float('inf')
+    _INFINITY = float('inf')
 
-    def __init__(self, topic, client, max_interval=INFINITY,
-                 max_messages=INFINITY):
+    def __init__(self, topic, client, max_interval=_INFINITY,
+                 max_messages=_INFINITY):
         self.topic = topic
         self.messages = []
         self.message_ids = []
--- a/pubsub/google/cloud/pubsub/topic.py
+++ b/pubsub/google/cloud/pubsub/topic.py
@@ -470,12 +470,11 @@ class Batch(object):
 
         # If too much time has elapsed since the first message
         # was added, autocommit.
-        if self._max_interval < self.INFINITY:
-            now = time.time()
-            if now - self._start_timestamp > self._max_interval:
-                self.commit()
-                self._start_timestamp = now
-                return
+        now = time.time()
+        if now - self._start_timestamp > self._max_interval:
+            self.commit()
+            self._start_timestamp = now
+            return
 
         # If the number of messages on the list is greater than the
         # maximum allowed, autocommit (with the batch's client).
--- a/pubsub/google/cloud/pubsub/topic.py
+++ b/pubsub/google/cloud/pubsub/topic.py
@@ -442,7 +442,7 @@ class Batch(object):
         self._max_messages = max_messages
 
         # Set the initial starting timestamp (used against the interval).
-        self._start_timestamp = float(time.time())
+        self._start_timestamp = time.time()
 
     def __enter__(self):
         return self
@@ -462,9 +462,6 @@ class Batch(object):
 
         :type attrs: dict (string -> string)
         :param attrs: key-value pairs to send as message attributes
-
-        :rtype: None
-        :returns: None
         """
         self.topic._timestamp_message(attrs)
         self.messages.append(
@@ -474,14 +471,17 @@ class Batch(object):
         # If too much time has elapsed since the first message
         # was added, autocommit.
         if self._max_interval < self.INFINITY:
-            if float(time.time()) - self._start_timestamp > self._max_interval:
-                self._start_timestamp = float(time.time())
-                return self.commit()
+            now = time.time()
+            if now - self._start_timestamp > self._max_interval:
+                self.commit()
+                self._start_timestamp = now
+                return
 
         # If the number of messages on the list is greater than the
         # maximum allowed, autocommit (with the batch's client).
         if len(self.messages) >= self._max_messages:
-            return self.commit()
+            self.commit()
+            return
 
     def commit(self, client=None):
         """Send saved messages as a single API call.
--- a/pubsub/unit_tests/test_topic.py
+++ b/pubsub/unit_tests/test_topic.py
@@ -779,6 +779,77 @@ class TestBatch(unittest.TestCase):
         self.assertEqual(list(batch.messages), [MESSAGE1, MESSAGE2])
         self.assertEqual(getattr(api, '_topic_published', self), self)
 
+    def test_message_count_autocommit(self):
+        """Establish that if the batch is assigned to take a maximum
+        number of messages, that it commits when it reaches that maximum.
+        """
+        client = _Client(project='PROJECT')
+        topic = _Topic(name='TOPIC')
+
+        # Track commits, but do not perform them.
+        Batch = self._get_target_class()
+        with mock.patch.object(Batch, 'commit') as commit:
+            with self._make_one(topic, client=client, max_messages=5) as batch:
+                self.assertIsInstance(batch, self._get_target_class())
+
+                # Publish four messages and establish that the batch does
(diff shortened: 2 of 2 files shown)


pubsub/google/cloud/pubsub/topic.py
```diff
@@ -408,15 +410,40 @@ class Batch(object):
...
+
+    :param max_messages: The maximum number of messages to hold in the batch
+                         before automatically commiting. Defaults to infinity
+                         (off).
+    :type max_messages: float
     """
-    def __init__(self, topic, client):
+    INFINITY = float('inf')
+
+    def __init__(self, topic, client, max_interval=INFINITY,
```
jonparrott: Why not just `None` instead of `INFINITY`?

lukesneeringer: Because `inf` handles the `<` and `>` checks correctly without `if foo and foo < bar` logic.

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -408,15 +410,40 @@ class Batch(object):
...
+                         (off).
+    :type max_interval: float
+
+    :param max_messages: The maximum number of messages to hold in the batch
+                         before automatically commiting. Defaults to infinity
+                         (off).
+    :type max_messages: float
     """
-    def __init__(self, topic, client):
+    INFINITY = float('inf')
```
dhermes: Why not a module-level constant?

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -408,15 +410,40 @@ class Batch(object):
...
         self.message_ids = []
         self.client = client
 
+        # Set the autocommit rules. If the interval or number of messages
+        # is exceeded, then the .publish() method will imply a commit.
+        self._max_interval = max_interval
+        self._max_messages = max_messages
+
+        # Set the initial starting timestamp (used against the interval).
+        self._start_timestamp = float(time.time())
```
dhermes: Isn't it already a float?

---

`time.time()` isn't guaranteed to be monotonic, e.g. if there is a leap second. Python 3 does provide a monotonic clock

```python
$ python3.5 -c 'import time; print(time.time()); print(time.monotonic())'
1485491671.1016715
569331.546634608
```

but Python 2 does not

```python
$ python2.7 -c 'import time; print(time.time()); print(time.monotonic())'
1485491665.14
AttributeError: 'module' object has no attribute 'monotonic'
```

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -435,12 +462,27 @@ def publish(self, message, **attrs):
 
         :type attrs: dict (string -> string)
         :param attrs: key-value pairs to send as message attributes
+
+        :rtype: None
```
dhermes: 1. `None` is not a type. (`NoneType` is the type)
1. If it doesn't return anything you don't need a returns/rtype block

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -435,12 +462,27 @@ def publish(self, message, **attrs):
...
         self.messages.append(
             {'data': message,
              'attributes': attrs})
 
+        # If too much time has elapsed since the first message
+        # was added, autocommit.
+        if self._max_interval < self.INFINITY:
+            if float(time.time()) - self._start_timestamp > self._max_interval:
+                self._start_timestamp = float(time.time())
+                return self.commit()
```
dhermes: Since this doesn't return I'd much prefer

```python
self.commit()
return
```

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -435,12 +462,27 @@ def publish(self, message, **attrs):
...
+
+        # If the number of messages on the list is greater than the
+        # maximum allowed, autocommit (with the batch's client).
+        if len(self.messages) >= self._max_messages:
+            return self.commit()
```
dhermes: Ditto

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -435,12 +462,27 @@ def publish(self, message, **attrs):
...
         self.topic._timestamp_message(attrs)
         self.messages.append(
             {'data': message,
              'attributes': attrs})
 
+        # If too much time has elapsed since the first message
+        # was added, autocommit.
+        if self._max_interval < self.INFINITY:
+            if float(time.time()) - self._start_timestamp > self._max_interval:
+                self._start_timestamp = float(time.time())
```
dhermes: 1.  Why cast to `float()`? (as above)
1.  Don't grab `time.time()` twice like this? Instead

    ```python
    if self._max_interval < INFINITY:
        now = time.time()
        if now - self._start_timestamp > self._max_interval:
            self.commit()
            self._start_timestamp = now
            return
    ```

1. Are you worried about setting `self._start_timestamp` and then `self.commit()` failing? i.e. does that put the instance in a bad state?

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -435,12 +462,27 @@ def publish(self, message, **attrs):
...
+        :returns: None
         """
         self.topic._timestamp_message(attrs)
         self.messages.append(
             {'data': message,
              'attributes': attrs})
 
+        # If too much time has elapsed since the first message
+        # was added, autocommit.
+        if self._max_interval < self.INFINITY:
```
dhermes: Why is this check needed? No `float` will ever be `> infinity` in the next check

pubsub/unit_tests/test_topic.py
```diff
@@ -779,6 +779,77 @@ def test_context_mgr_failure(self):
         self.assertEqual(list(batch.messages), [MESSAGE1, MESSAGE2])
         self.assertEqual(getattr(api, '_topic_published', self), self)
 
+    def test_message_count_autocommit(self):
+        """Establish that if the batch is assigned to take a maximum
+        number of messages, that it commits when it reaches that maximum.
+        """
```
dhermes: :-1: on docstring in a unit test. It mucks with test runner output and isn't worth it. (But :+1: on trying to be a good communicator)

pubsub/unit_tests/test_topic.py
```diff
@@ -779,6 +779,77 @@ def test_context_mgr_failure(self):
...
+        client = _Client(project='PROJECT')
+        topic = _Topic(name='TOPIC')
+
+        # Track commits, but do not perform them.
+        Batch = self._get_target_class()
```
dhermes: Style nit: s/`Batch`/`klass` (or your favorite mis-spelling)

pubsub/unit_tests/test_topic.py
```diff
@@ -779,6 +779,77 @@ def test_context_mgr_failure(self):
...
+        with mock.patch.object(Batch, 'commit') as commit:
+            with self._make_one(topic, client=client, max_messages=5) as batch:
+                self.assertIsInstance(batch, self._get_target_class())
+
+                # Publish four messages and establish that the batch does
+                # not commit.
+                for i in range(0, 4):
+                    batch.publish({
+                        'attributes': {},
+                        'data': 'Batch message %d.' % i,
```
dhermes: Style nit: always pass a tuple to `%`, e.g. `'Batch message %d.' % (i,)`

pubsub/unit_tests/test_topic.py
```diff
@@ -779,6 +779,77 @@ def test_context_mgr_failure(self):
...
+        number of messages, that it commits when it reaches that maximum.
+        """
+        client = _Client(project='PROJECT')
+        topic = _Topic(name='TOPIC')
+
+        # Track commits, but do not perform them.
+        Batch = self._get_target_class()
+        with mock.patch.object(Batch, 'commit') as commit:
+            with self._make_one(topic, client=client, max_messages=5) as batch:
+                self.assertIsInstance(batch, self._get_target_class())
```
dhermes: You already have the class so you don't need to call `_get_target_class()` again (though maybe the mocking breaks something?)

pubsub/unit_tests/test_topic.py
```diff
@@ -779,6 +779,77 @@ def test_context_mgr_failure(self):
...
+        """Establish that if the batch is sufficiently old, that it commits
+        the next time it receives a publish.
+        """
+        client = _Client(project='PROJECT')
+        topic = _Topic(name='TOPIC')
+
+        # Track commits, but do not perform them.
+        Batch = self._get_target_class()
+        with mock.patch.object(Batch, 'commit') as commit:
+            mock_time.return_value = 0.0
```
dhermes: Move to very beginning of the context-managed block (to put it on the same scope as the other time you manually set the return value)?

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: There are several places where the function _does_ return in order to stop control flow, which makes the linter require `:rtype:`.

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: > `time.time()` isn't guaranteed to be monotonic

I do not really think that particularly matters for this purpose.

pubsub/google/cloud/pubsub/topic.py
dhermes: Very annoying of Pylint to do that. If you change the `return self.commit` to two lines (per my feedback), does Pylint still complain?

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: My thinking at the time was that `time.time()` was more expensive than the numeric comparison, and `self._max_interval` being set to `inf` is the default (and most common) case.

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: Concur on line 445; fixing.

pubsub/unit_tests/test_topic.py
lukesneeringer: I am not tracking you here. The `return_value` is being set on line 826; this first becomes relevant on line 827. So, I cannot move it lower or into the same scope as line 840. I can move it _higher_, but I do not think that is what you are asking for?

pubsub/unit_tests/test_topic.py
lukesneeringer: Nope, that's just a cleanup fail.

pubsub/unit_tests/test_topic.py
lukesneeringer: Is there a rule for that? `Batch` is clearer. (_Which_ class is it?)

pubsub/unit_tests/test_topic.py
lukesneeringer: How does it muck with the test runner?

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: Agreed.

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: Agreed.

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: Agreed on all counts.

pubsub/unit_tests/test_topic.py
dhermes: e.g. https://gist.github.com/dhermes/e8059e8df6959b58651ca80d3b15d584

pubsub/google/cloud/pubsub/topic.py
dhermes: You are probably right in that it adds some speed. But in that case you might as well use `None` for the default of `_max_interval`

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -441,6 +468,21 @@ def publish(self, message, **attrs):
...
+            now = time.time()
+            if now - self._start_timestamp > self._max_interval:
+                self.commit()
+                self._start_timestamp = now
+                return
+
+        # If the number of messages on the list is greater than the
+        # maximum allowed, autocommit (with the batch's client).
+        if len(self.messages) >= self._max_messages:
```
dhermes: Sorry I didn't notice this on the first pass, but this should be a strict inequality (`>`)

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -441,6 +468,21 @@ def publish(self, message, **attrs):
...
+            self.commit()
+            return
```
dhermes: This return unnecessary since it's the last line in the method. But maybe it's good to keep it to future proof?

pubsub/unit_tests/test_topic.py
dhermes: Nope no rule. Just my "opinion" (which isn't the same as a universally agreed-upon practice)

pubsub/unit_tests/test_topic.py
dhermes: Yeah that's my bad. But I think my lack of clarity is indicative of something. Suggest instead:

```python
# The time() that will be set in Batch() constructor
mock_time.return_value = 0.0
batch = self._make_one(topic, client=client, max_interval=5)
with batch:
    # do some stuff
    # ...
    # Move time ahead so that this batch is too old.
    mock_time.return_value = 10.0
```

You may want to `with batch as batch_ctx` but there is (currently) no reason to since `Batch.__enter__` returns `self`

PR #2967, opened 2017-01-26
Vision support for batch processing part one.

daspecster:
This just get's the under the hood changes ready for the most part.

### DIFF
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -39,8 +39,8 @@ class _GAPICVisionAPI(object):
         :type features: list
         :param features: List of :class:`~google.cloud.vision.feature.Feature`.
 
-        :rtype: :class:`~google.cloud.vision.annotations.Annotations`
-        :returns: Instance of ``Annotations`` with results or ``None``.
+        :rtype: :class:`~google.cloud.vision.annotations.Annotations` or list
+        :returns: Instance of ``Annotations`` or list of ``Annotations``.
         """
         gapic_features = [_to_gapic_feature(feature) for feature in features]
         gapic_image = _to_gapic_image(image)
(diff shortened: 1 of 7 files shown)


vision/google/cloud/vision/_gax.py
```diff
@@ -39,21 +39,21 @@ def annotate(self, image, features):
...
         requests = [request]
         annotator_client = self._annotator_client
-        images = annotator_client.batch_annotate_images(requests)
-        if len(images.responses) == 1:
-            return Annotations.from_pb(images.responses[0])
-        elif len(images.responses) > 1:
-            raise NotImplementedError(
-                'Multiple image processing is not yet supported.')
+        responses = annotator_client.batch_annotate_images(requests).responses
+        if len(responses) == 1:
```
lukesneeringer: Please either return a list (or other iterable) all the time or return the object all the time. Do not magically flatten if there is one and exactly one result, as it is entirely common for the developer to not know how many results to expect back.

(If the Vision API returns results with an excessive amount of nesting, we can abridge that for developers by adding appropriate helper methods on `Annotations` objects.)

vision/google/cloud/vision/_http.py
```diff
@@ -41,20 +41,20 @@ def annotate(self, image, features):
...
+        :rtype: list or class:`~googe.cloud.vision.annotations.Annotations`
+        :returns: Instance of ``Annotations`` or List of ``Annotations``.
         """
         request = _make_request(image, features)
 
         data = {'requests': [request]}
         api_response = self._connection.api_request(
             method='POST', path='/images:annotate', data=data)
-        images = api_response.get('responses')
-        if len(images) == 1:
```
lukesneeringer: Same comment as above. Return a list consistently or object consistently, but do not magic-flatten for a single result.

vision/google/cloud/vision/_gax.py
dhermes: Ditto. Polymorphic inputs / outputs are possible in Python, but still jarring for users

vision/unit_tests/test__gax.py
```diff
@@ -102,9 +102,9 @@ def test_annotate_multiple_results(self):
         gax_api._annotator_client = mock.Mock(
             spec_set=['batch_annotate_images'], **mock_response)
         with mock.patch('google.cloud.vision._gax.Annotations'):
-            with self.assertRaises(NotImplementedError):
-                gax_api.annotate(image, [feature])
+            responses = gax_api.annotate(image, [feature])
 
+        self.assertEqual(len(responses), 2)
```
dhermes: Please verify some/all of the content of each response

PR #2968, opened 2017-01-26
Renaming JSONClient -> ClientWithProject.

dhermes:
Also using empty-tuple default as `extra_options` (avoids a branch).


lukesneeringer:
Looks fine to me, pending tests.

PR #2970, opened 2017-01-27
Making Connection()-s act as proxies for data stored in Client()-s

dhermes:
This is part of a re-factoring effort to remove / reduce the role of `Connection` classes (since they are outdated): #2606 

I ran the system tests and `speech` is failing but I think it's unrelated to this? Also, the umbrella coverage is failing, so I sent #2969 to fix (also unrelated to this PR).

### DIFF
--- a/bigquery/google/cloud/bigquery/_http.py
+++ b/bigquery/google/cloud/bigquery/_http.py
@@ -18,7 +18,11 @@ from google.cloud import _http
 
 
 class Connection(_http.JSONConnection):
-    """A connection to Google BigQuery via the JSON REST API."""
+    """A connection to Google BigQuery via the JSON REST API.
+
+    :type client: :class:`~google.cloud.bigquery.client.Client`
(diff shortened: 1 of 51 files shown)


daspecster:
@dhermes #2965 should get the speech system tests working again.

dhermes:
@lukesneeringer @tseaver PTAL. This is (most likely) blocking a `google-cloud-core` release, which is blocked all other pending releases.

tseaver:
LGTM.

@lukesneeringer Are you from Texas?  That avatar image looks like the "Come and Take It!" flag.

lukesneeringer:
Yeah; I just moved to California in December after living in Texas for most of the last 3.5 decades. :-)

tseaver:
We've been in Virginia for the last 17 years, but I'm still a Texan (I taught Texas history to seventh graders long ago in a galaxy far away).

PR #2973, opened 2017-01-30
Updates for pycodestyle.

### LINKED ISSUE #2974: Pycodestyle update causing our lint to fail.
From what I can tell, there are three new errors that pycodestyle is finding.

E306: expected 1 blank line before a nested definition, found 0
~E123: closing bracket does not match indentation of opening bracket's line~
~E722: do not use bare except~

See: #2973

daspecster:
These changes allow pycodestyle to pass.

Closes #2974

### DIFF
--- a/bigquery/unit_tests/test_dataset.py
+++ b/bigquery/unit_tests/test_dataset.py
@@ -838,7 +838,7 @@ class _Connection(object):
 
         try:
             response, self._responses = self._responses[0], self._responses[1:]
-        except:
+        except IndexError:
             raise NotFound('miss')
         else:
             return response
--- a/datastore/google/cloud/datastore/transaction.py
+++ b/datastore/google/cloud/datastore/transaction.py
@@ -186,7 +186,7 @@ class Transaction(Batch):
         try:
             self._id = self._client._connection.begin_transaction(
                 self.project)
-        except:  # noqa: E722
+        except:  # noqa: E722 do not use bare except, specify exception instead
             self._status = self._ABORTED
             raise
 
--- a/datastore/google/cloud/datastore/transaction.py
+++ b/datastore/google/cloud/datastore/transaction.py
@@ -186,7 +186,7 @@ class Transaction(Batch):
         try:
             self._id = self._client._connection.begin_transaction(
                 self.project)
-        except:
+        except:  # noqa: E722
             self._status = self._ABORTED
             raise
 
--- a/bigquery/unit_tests/test_job.py
+++ b/bigquery/unit_tests/test_job.py
@@ -1904,7 +1904,7 @@ class _Connection(object):
 
         try:
             response, self._responses = self._responses[0], self._responses[1:]
-        except:
+        except IndexError:
             raise NotFound('miss')
         else:
             return response
--- a/bigquery/unit_tests/test_table.py
+++ b/bigquery/unit_tests/test_table.py
@@ -1989,7 +1989,7 @@ class _Connection(_Responder):
 
         try:
             response, self._responses = self._responses[0], self._responses[1:]
-        except:
+        except IndexError:
             raise NotFound('miss')
         else:
             return response
--- a/bigtable/unit_tests/test_instance.py
+++ b/bigtable/unit_tests/test_instance.py
@@ -254,8 +254,8 @@ class TestInstance(unittest.TestCase):
             metadata=Any(
                 type_url=type_url,
                 value=metadata.SerializeToString(),
-                )
             )
+        )
 
         # Patch the stub used by the API method.
         client._instance_stub = stub = _FakeStub(response_pb)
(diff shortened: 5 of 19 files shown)


bigquery/unit_tests/test_dataset.py
```diff
@@ -838,7 +838,7 @@ def api_request(self, **kw):
 
         try:
             response, self._responses = self._responses[0], self._responses[1:]
-        except:
+        except Exception:
```
dhermes: :-1: Never catch `Exception`. Here the exception is an `IndexError`:

```python
>>> a = []
>>> a[0]
IndexError: list index out of range
```

core/unit_tests/test__helpers.py
```diff
@@ -138,12 +138,14 @@ def _call_fut(self, project=None):
         return _determine_default_project(project=project)
 
     def test_it(self):
+
```
dhermes: `pycodestyle` complained about this?

tseaver:
Are the whitespace changes actually required by the tool, or did you just add them because they seemed needed?  I know of no justification in PEP8 etc. for them.

core/unit_tests/test__helpers.py
```diff
@@ -138,12 +138,14 @@ def _call_fut(self, project=None):
...
         with mock.patch('google.auth.default', autospec=True) as default:
             default.return_value = (
                 mock.sentinel.credentials, mock.sentinel.project)
             project = self._call_fut()
 
         self.assertEqual(project, mock.sentinel.project)
+
```
dhermes: Also surprised `pycodestyle` said anything here

dhermes:
> Are the whitespace changes actually required by the tool, or did you just add them because they seemed needed? I know of no justification in PEP8 etc. for them.

Jinx, bigly

daspecster:
@tseaver I'm wondering if it's a bug in pycodestyle. I moved them around a little bit to try and see what it was looking for and that was the way it had to be to pass. We could probably disable that error.

dhermes:
@daspecster Please provide the errors so we can try to help

daspecster:
There were only two types of errors. One was the bare except and this was the other one.

```
core/unit_tests/test__helpers.py:147:9: E306 expected 1 blank line before a nested definition, found 0
```

tseaver:
I vote to suppress that error with prejudice.

dhermes:
I'd agree, we should suppress, but not before filing an issue with the `pycodestyle` folks

daspecster:
SGTM, I'll update and I can file the issue on pycodestyle as well.

daspecster:
There's a bare except in [transaction.py](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/3584917bc96ce2574d700d40b66eeacb3a949043/datastore/google/cloud/datastore/transaction.py#L189) that I'm not sure what the possible exception types could be. I think it could be a few things. If there is a known exception type for this then I can stop ignoring E306.

dhermes:
The except in `transaction.py` is meant to be bare (it's somewhat of a special case). You can locally disable `pycodestyle` checks via `noqa` ([e.g.][1])

[1]: https://github.com/dhermes/bezier/blob/6e44c92f0afdcb14e729dccdf40a34aaada7b9d4/src/bezier/_curve_helpers.py#L51

datastore/google/cloud/datastore/transaction.py
```diff
@@ -186,7 +186,7 @@ def begin(self):
         try:
             self._id = self._client._connection.begin_transaction(
                 self.project)
-        except:
+        except:  # noqa: E722
```
dhermes: @daspecster Still applies

daspecster: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2973/commits/aeeead940047fd64c31f34b9b8c2ce0be36f83bf

datastore/google/cloud/datastore/transaction.py
```diff
@@ -186,7 +186,7 @@ def begin(self):
         try:
             self._id = self._client._connection.begin_transaction(
                 self.project)
-        except:
+        except:  # noqa: E722 do not use bare except, specify exception instead
```
dhermes: FWIW I assumed you'd summarize in your own words, but this fits on one line, so it's fine

daspecster: Yeah, but that text is searchable.  e.g. if you copy and paste that into google you'll get more information and the source of the definition.

lukesneeringer:
> The except in transaction.py is meant to be bare (it's somewhat of a special case). You can locally disable pycodestyle checks via noqa (e.g.)

Or `except Exception`, if it passes the linter.

dhermes:
@lukesneeringer I am allergic to `except Exception` **and** that misses nasty exception types that only inherit from `BaseException` **and** `pylint` will complain about `Exception` being too broad

lukesneeringer:
Fair on all counts. :-)

Issue #2673, opened 2016-11-03
Python Logging API always defaults to Custom logs


HiroLord:
When using the **Python Stackdriver Logging API**, there is not a well-defined way to specify where you want your logs to land. All of them will go into Custom Logs. The way around this is to use a Logging Handler, but this is not well written anywhere except for this one particular blog post:

https://medium.com/google-cloud/cloud-logging-though-a-python-log-handler-a3fbeaf14704#.39a97qpib

The reason it cannot be done through the standard API is because this line 

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/logger.py#L127

overwrites whatever resource you would rather write to. No parameter I could find (or the Google engineer I asked) would set the resource in order to change the destination of the logs. The closest we could find is here:

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/connection.py#L131

but this is set by the previous line stated above. Changing this resource manually is a pretty important feature to us that we are surprised is not already there.

tseaver:
We would need to add a `MonitoredResource` type (including mapping the [`list` API method](https://cloud.google.com/logging/docs/api/reference/rest/v2/monitoredResourceDescriptors/list) for them), and then allow passing an instance of it through to the logger, either as an optional constructor paramater or as a keyword parameter for `log_text` and `log_struct`.

@waprin Any thoughts?

waprin:
yes @HiroLord, you are definitely right, you should be able to specify a monitored resource. 

As a stopgap to at least get the resources types right on GAE/GKE I have #2668 which uses the fluentd handlers. However you are right that in general the client library should allow you to specify a monitored resource of your choice. 

I played around with it a bit and found that besides the resource type, I also needed to set the right labels in order for the logs to show up in places where I expect. But there are different labels for different monitored resource types. So I'm not totally sure about the scope of the change necessary.

Out of curiosity, what types of resources are you trying to write to?

waprin:
Also @HiroLord just a small correction - the logging handler is totally orthogonal to this issue. Handlers are just a way of mapping Python `logging` calls to API calls, but we have logging handlers in this repo as well, though at the moment they only write to custom logs.
 
[...]

Issue #2819, opened 2016-12-06
Error reporting instrumentation and SIGTERM handling


karan:
Dependent on #1921, we need to add a SIGTERM handler that can publish thread trace and heap info to Error Reporting.

Example interface:

```
from google.cloud import error_reporting
client = error_reporting.Client(project='my-project',
                                               credentials=creds,
                                               shutdownLogs={threadDump: True, heapInfo: False})
```

dhermes:
@waprin thoughts?

waprin:
yes @karan has already asked for it, I said I would plan to get it done this quarter, I still have a month left!! Was going to start this week, already did some preliminary research, but need to do a bit more to totally understand what we want and what we can get.

karan:
Thanks Bill. Just creating this as an external tracker! Let me know what I can do to help. Really excited for this work.

lukesneeringer:
@waprin I think we have this now, correct? (If so, let me know so I can close this; if not, what is the priority on it?)

waprin:
Yes, I will still need to add some memory stats but that will also be a small change.

waprin:
I think priority is probably P2 or P3 hence the chipping away in free cycles but it's close to done, will close after.

duggelz:
Can we just use the standard [`faulthandler`](https://docs.python.org/3/library/faulthandler.html) module from Python 3 (and backported to Python 2 on PyPI)?

I worked a lot on the C++ crash handling inside Google, and it's tremendously difficult to get everything right in a robust and portable way.  And there is a direct correlation where the more elaborate you make the crash handler, the less reliable it is.  Writing to files other than stdout/stderr is iffy, let alone making RPC calls.

lukesneeringer:
Hello,
One of the challenges of maintaining a large open source project is that sometimes, you can bite off more than you can chew. As the lead maintainer of `google-cloud-python`, I can definitely say that I have let the issues here pile up.

As part of trying to get things under control (as well as to empower us to provide better customer service in the future), I am declaring a "bankruptcy" of sorts on many of the old issues, especially those likely to have been addressed or made obsolete by more recent updates.

My goal is to close stale issues whose relevance or solution is no longer immediately evident, and which appear to be of lower importance. I believe in good faith that this is one of those issues, but I am scanning quickly and may occasionally be wrong. If this is an issue of high importance, please comment here and we will reconsider. If this is an issue whose solution is trivial, please consider providing a pull request.

Thank you!

PR #2976, opened 2017-01-31
Add GAE Shutdown Handler

waprin:
Very rough draft (no tests/docs, it's messy) etc, to get feedback on overall direction from @karan and whoever will review for this repo.

This is a feature for App Engine Flex customers to get more post-mortem logs on VM shutdown. In theory it could be generalized but it's been spec-ed out to be very GAE specific so currently code just checks for that.

Due to lack of MonitoredResource support (see #2673) this currently uses the `client.logging_api` directly. This could be changed if the ability to specific monitoredresource gets added to the higher level API.


logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,112 @@
...
+def _fetch_metadata(metdata_path):
+    url = 'http://{}/computeMetadata/v1/{}'.format(_METADATA_SERVICE_ADDR,
+                                                   metdata_path)
+    h = httplib2.Http()
+    resp, content = h.request(url, 'GET')
+    return content
+
+
+def _get_gae_instance():
+    return _fetch_metadata('instance/attributes/gae_backend_instance')
```
karan: You can use `GAE_INSTANCE` env variable here and fall back to http request.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def _get_gae_backend():
+    return _fetch_metadata('instance/attributes/gae_backend_name')
```
karan: You can use `GAE_SERVICE` env variable here and fall back to http request.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def _get_gae_version():
+    return _fetch_metadata('instance/attributes/gae_backend_version')
```
karan: You can use `GAE_VERSION` env variable here and fall back to http request.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,112 @@
...
+
+    labels = {'appengine.googleapis.com/version_id': gae_version,
+              'compute.googleapis.com/resource_type': 'instance',
+              'appengine.googleapis.com/instance_name': gae_instance,
+              'appengine.googleapis.com / module_id': gae_backend, }
+    entry = {'text_payload': text_payload}
+
+    entries = [entry]
+
+    client.logging_api.write_entries(
```
karan: This will fail if `entry` size is more than 100Kb. You could split entry into multiple messages (each being at most 100Kb).

logging/google/cloud/logging/client.py
```diff
@@ -327,3 +323,19 @@ def setup_logging(self, log_level=logging.INFO,
...
+        This method installs a SIGTERM handler that will report various
+        application metrics to Stackdriver Logging.
+
+        Currently the only supported option is stacktrace logging.
+
+        :type log_level: bool
+        :param log_level: (Optional) When true, on SIGTERM the application
+                          will log the stacktrace of all running threads.
+        """
+        if thread_dump:
```
lukesneeringer: I am confused by the `thread_dump` argument. If it is `False`, the method no-ops? Why have the argument at all?

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def setup_shutdown_stacktrace_reporting(client):
+    """Installs a SIGTERM handler to log stack traces to  Stackdriver.
+
+    :type client: `google.cloud.logging.Client`
+    :param client: Stackdriver logging client.
+    """
+    if not _is_on_appengine():
+        raise Exception('Shutdown reporting is only supported on App Engine '
```
lukesneeringer: `RuntimeError`, please. (I think.)

waprin:
@dhermes @karan ready for review. I didn't put anything in the usage docs yet, felt a little too niche.

lukesneeringer:
It looks fine to me,  but hoping @dhermes will take a pass also before we merge it.

waprin:
Addresses #2819

lukesneeringer:
Officially assigning @dhermes for review; I forgot to the first time.

waprin:
@dhermes ping

logging/google/cloud/logging/_environment_vars.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2017 Google Inc.
```
dhermes: @lukesneeringer We should probably move most env. vars. out of `google-cloud-core`. `gcloud.environment_vars` made sense for DRY and documentation purposes, but that's not really true anymore after the subpackage split.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+import signal
+import sys
+import traceback
+
+from google.cloud.logging._environment_vars import _APPENGINE_FLEXIBLE_ENV_VM
+from google.cloud.logging._environment_vars import _APPENGINE_FLEXIBLE_ENV_FLEX
+
+
+# Maximum size in bytes to send to Stackdriver Logging in one entry
+MAX_PAYLOAD_SIZE = 1024 * 100
```
dhermes: What is the point of `* 100`? Why not just use the literal `102400`?

Also, name it `_MAX_PAYLOAD_SIZE`

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _get_gae_instance():
+    """Returns the App Engine Flexible instance."""
+    return os.getenv('GAE_INSTANCE')
```
dhermes: This is in your env. vars. module

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _get_gae_service():
+    """Returns the App Engine Flexible service."""
+    return os.getenv('GAE_SERVICE')
```
dhermes: Should `GAE_SERVICE` go in env. vars.?

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _get_gae_version():
+    """Returns the App Engine Flexible version."""
+    return os.getenv('GAE_VERSION')
```
dhermes: Should `GAE_VERSION` go in env. vars.?

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _split_entry(payload):
+    """Splits payload into lists of maximum 100Kb.
+
+    Stackdriver Logging payloads are a maximum of 100Kb.
+    """
+    return [payload[i:i + MAX_PAYLOAD_SIZE]
+            for i in range(0, len(payload), MAX_PAYLOAD_SIZE)]
```
dhermes: s/`range`/`six.moves.xrange`

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _write_stacktrace_log(client, traces):
+    """Writes the trace logs to the appropriate GAE resource in Stackdriver.
+
+    :type client: `google.cloud.logging.Client`
```
dhermes: `:class:` missing here

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+                   Stackdriver logging.
+    """
+    gae_version = _get_gae_version()
+    gae_service = _get_gae_service()
+    gae_instance = _get_gae_instance()
+
+    text_payload = '{}\nThread traces\n{}'.format(gae_instance, traces)
+
+    logger_name = 'projects/{}/logs/{}'.format(
+        client.project, 'appengine.googleapis.com%2Fapp.shutdown')
```
dhermes: Make the `appengine.googleapis.com...` string a non-public module constant? Or better yet, hardcode the template (with that value filled in) as a non-public module constant?

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+    resource = {'type': 'gae_app', 'labels': {'project_id': client.project,
+                                              'version_id': gae_version,
+                                              'module_id': gae_service}}
```
dhermes: "Better" as:

```python
resource = {
    'type': 'gae_app',
    'labels': {
        'project_id': client.project,
        'version_id': gae_version,
        'module_id': gae_service,
    },
}
```

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _is_on_appengine():
+    """Returns True if the environment is detected as App Engine flexible."""
+    return (os.getenv(_APPENGINE_FLEXIBLE_ENV_VM) or os.getenv(
+        _APPENGINE_FLEXIBLE_ENV_FLEX))
+
+
+def _report_stacktraces(
+        client, signal_, frame):  # pylint: disable=unused-argument
```
dhermes: This **only** uses `client`, ostensibly so that using `functools.partial` with `client` allows it to be used with `signal.signal` ([docs][1]). Why don't you want to use that information?

[1]: https://docs.python.org/3.5/library/signal.html#signal.signal

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+    :param client: Stackdriver logging client.
+
+    :type signal_: int
+    :param signal_: Signal number.
+
+    :type frame: frame object
+    :param frame: The current stack frame.
+    """
+    traces = ''
+    for thread_id, stack in sys._current_frames().items():
```
dhermes: Why is it necessary to use a non-public API? That seems like a really bad idea.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+                                              'module_id': gae_service}}
+
+    labels = {
+        'appengine.googleapis.com/version_id': gae_version,
+        'compute.googleapis.com/resource_type': 'instance',
+        'appengine.googleapis.com/instance_name': gae_instance,
+        'appengine.googleapis.com/module_id': gae_service,
+    }
+
+    split_payloads = _split_entry(text_payload.encode('utf-8'))
```
dhermes: Unless you know that `text_payload` is **always** `unicode` (Py2) / `str` (Py3) this will break in Python 3. You're better off using `google.cloud._helpers._to_bytes`.

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+    :type client: `google.cloud.logging.Client`
+    :param client: Stackdriver logging client.
+
+    :type signal_: int
+    :param signal_: Signal number.
+
+    :type frame: frame object
+    :param frame: The current stack frame.
+    """
+    traces = ''
```
dhermes: Typically the recommendation is to create a list and then join the list at the end. The old adage **WAS** that creating a new string every time was memory intensive, whereas appending to a list was not. (I don't think it's true anymore, but still is "good form".)

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,131 @@
...
+    """Installs a SIGTERM handler to log stack traces to Stackdriver.
+
+    :type client: `google.cloud.logging.Client`
+    :param client: Stackdriver logging client.
+    """
+    if not _is_on_appengine():
+        raise RuntimeError('Shutdown reporting is only supported on App '
+                           'Engine flexible environment.')
+    signal.signal(
+        signal.SIGTERM, functools.partial(_report_stacktraces, client))
```
dhermes: I'm not sure `SIGTERM` is sufficient. Do you have any direct experience with this? e.g. it won't catch any `KeyboardInterrupt` (which corresponds to `SIGINT`). Also, if the shell that is running the process terminates it, `SIG_IGN` will be the typical signal.

Big picture, if you want to log on **ANY** program crash this is far from sufficient.

dhermes: Here is a "handy" way you can see which kind of exits correspond to which signals: https://gist.github.com/dhermes/c735287ef6da5a369445b13c2ec1a62b

logging/google/cloud/logging/_shutdown.py
waprin: Renamed, and switched order.

It's more readable to me because in my head it reads as 100KB and shows people how the eventual number is built. Keeping unless you feel strongly.

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: done and preemptively added trailing commas :)

logging/google/cloud/logging/_shutdown.py
waprin: Hmm I think I do know because I can't see how anything else would sneak into a stacktrace *but* seems reasonable to switch to that anyway.

logging/google/cloud/logging/_shutdown.py
waprin: Not sure what you're asking for here. Your assessment of the function's usage is correct. It's a private function that's only ever expected to be called from the signal SIGTERM handler setup a few lines below. Since it's only ever supposed to be called with SIGTERM, I don't need to know the signal/frame since it's only called in that context, it's just here to match the required signature. I could exercise an abundance of caution and assert that it is in fact a SIGTERM? Not sure what else to do with it.

logging/google/cloud/logging/_shutdown.py
waprin: done

logging/google/cloud/logging/_shutdown.py
waprin: I totally agree but so far don't see a better option, virtually every search result is recommending this approach, and I think the whole concept of this PR is slightly exotic enough that maybe sys.py authors thought it wasn't necessary to make public. I can't find any obvious alternative...if it bothers you a lot maybe I could shoot out an email to various mailing lists?

logging/google/cloud/logging/_shutdown.py
waprin: This PR is for a very specific use case on GAE Flex at the request of their engineering team, that's why it raises an exception when not in that environment. They specifically send a SIGTERM on VM shutdown and want this info logged (along with some other stats in future PRs). For a bunch of reasons, I'm hesitant to generalize it beyond that purpose.

waprin:
wow good to know difference between += and append there. Others comments were just me forgetting to run lint all done.

waprin:
Agreed I can add a test for that but have to head to GDG at the moment.

waprin:
@dhermes sorry for delay got pulled into that conference rest of yesterday.

So it turns out that my tests weren't *that* bad because the function behavior was still correct, since the last line of the function joins the entire list into one big string  at the end, it's irrelevant whether the strings get broken up in the interim. Obviously sloppy to mix += and append like that but just noting end result was the same.

Since I took a look at the test anyway I did try to make it check a bit more with a regex rather than just looking for a function name.

waprin:
@dhermes ping

logging/google/cloud/logging/_environment_vars.py
```diff
@@ -0,0 +1,32 @@
...
+
+_APPENGINE_FLEXIBLE_ENV_VM = 'GAE_APPENGINE_HOSTNAME'
+"""Environment variable set in App Engine when vm:true is set."""
+
+
+_APPENGINE_FLEXIBLE_ENV_FLEX = 'GAE_INSTANCE'
+"""Environment variable set in App Engine when env:flex is set."""
+
+
+_CONTAINER_ENGINE_ENV = 'KUBERNETES_SERVICE'
```
dhermes: 1. Since the module is non-public, the variables can be public within the module
1. Since you are "exporting" this globals, you actually **want** them to be public
1. You seem to be unclear on which you want here: two or one empty line between defined constants (you want one line)

logging/google/cloud/logging/_shutdown.py
dhermes: I was just trying to wrap my head around it. Just add a note in the docstring **that** they are unused and note **why** they are unused.

logging/google/cloud/logging/_shutdown.py
dhermes: I'm on the fence on how to feel about a "very specific use case" as it pertains to this library. Would be interested in what @jonparrott and @lukesneeringer think.

:+1: on the `SIGTERM` explanation, though if it were my application I'd feel uneasy about the reliability of this (i.e. it may not always trigger for various reasons)

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import mock
+
+
+class Test_setup_shutdown_stracktrace_reporting(unittest.TestCase):
```
dhermes: Please follow our naming conventions: `Test_setup_stacktrace_crash_report`

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _call_fut(self, request):
+        from google.cloud.logging._shutdown import (
+            setup_stacktrace_crash_report)
+        return setup_stacktrace_crash_report(request)
+
+    def test_setup_shutdown_stacktrace_reporting_no_gae(self):
```
dhermes: 1. You could simply name this `test_outside_of_gae`
1. You should actually mock this method to **make sure** that `_is_on_appengine()` returns `False`

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+        with self.assertRaises(RuntimeError):
+            self._call_fut(mock.Mock())
+
+    def test_setup_shutdown_stacktrace_reporting(self):
```
dhermes: Better name: `test_success`

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+        from google.cloud.logging._environment_vars import (
+            _APPENGINE_FLEXIBLE_ENV_VM)
+        from google.cloud._testing import _Monkey
+        import os
+
+        signal_patch = mock.patch('google.cloud.logging._shutdown.signal')
+
+        with _Monkey(os, environ={_APPENGINE_FLEXIBLE_ENV_VM: 'True'}):
```
dhermes: Just forget about `_Monkey` since we have `mock`: `mock.patch('os.environ', new={...})`

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_setup_shutdown_stacktrace_reporting_no_gae(self):
+        with self.assertRaises(RuntimeError):
+            self._call_fut(mock.Mock())
+
+    def test_setup_shutdown_stacktrace_reporting(self):
+        from google.cloud.logging._environment_vars import (
+            _APPENGINE_FLEXIBLE_ENV_VM)
+        from google.cloud._testing import _Monkey
+        import os
```
dhermes: [Import][1] 1. stdlib, 2. 3rd party 3. local

[1]: https://www.python.org/dev/peps/pep-0008/#imports

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+        signal_patch = mock.patch('google.cloud.logging._shutdown.signal')
+
+        with _Monkey(os, environ={_APPENGINE_FLEXIBLE_ENV_VM: 'True'}):
+            with signal_patch as signal_mock:
+                self._call_fut(mock.Mock())
+                self.assertTrue(signal_mock.signal.called)
```
dhermes: You should assert more here, i.e. what do you expect the arguments to be / to look like

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class Test_write_stackrace_log(unittest.TestCase):
+
+    def _call_fut(self, client, traces):
+        from google.cloud.logging._shutdown import (
+            _write_stacktrace_log
+        )
```
dhermes: This fits easily in 80:

```python
    def _call_fut(self, client, traces):
        from google.cloud.logging._shutdown import _write_stacktrace_log
```

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+        return _write_stacktrace_log(client, traces)
+
+    def test_write_stracktrace_log(self):
+        from google.cloud._testing import _Monkey
+        import os
```
dhermes: Import `os` first

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+        mock_client = mock.Mock()
```
dhermes: Please **always** set a strict `spec` on your mocks. Start with `mock.Mock(spec=[])` and build up as needed. In this case, you also want to mock `client.logging_api` and `client.logging_api.write_entries` (being deliberate is important to **some degree**)

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+        trace = 'a simple stack trace'
+        with _Monkey(os, environ={'GAE_INSTANCE': 'myversion'}):
+            self._call_fut(mock_client, trace)
+            called = mock_client.logging_api.write_entries.call_args
+
+        expected_payload = 'myversion\nThread traces\n{}'.format(
+            trace).encode('utf-8')
+        self.assertEqual(called[0][0], [{'text_payload': expected_payload}])
```
dhermes: This leaves a lot of "what if?". Something a bit cleaner:

```python
self.assertEqual(len(mock_client.logging_api.write_entries.mock_calls), 1))
call = mock_client.logging_api.write_entries.mock_calls[0]
name, positional, keyword = call
# Do stuff with the call arguments (positional and keyword)
# You probably don't care about `name` so it'll probably be `_`
```

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class Test_report_stacktraces(unittest.TestCase):
```
dhermes: Naming convention: `Test__report_stacktraces`

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+            _report_stacktraces
+        )
+        return _report_stacktraces(client, signal, frame)
+
+    def test_report_stacktraces(self):
+        import re
+        patch = mock.patch(
+            'google.cloud.logging._shutdown._write_stacktrace_log')
+        with patch as write_log_mock:
+            self._call_fut(mock.Mock(), mock.Mock(), mock.Mock())
```
dhermes: Instead, use `mock.sentintel.client, None, None` as your arguments

logging/unit_tests/test__shutdown.py
```diff
@@ -0,0 +1,90 @@
...
+
+class Test_report_stacktraces(unittest.TestCase):
+
+    def _call_fut(self, client, signal, frame):
+        from google.cloud.logging._shutdown import (
+            _report_stacktraces
+        )
+        return _report_stacktraces(client, signal, frame)
+
+    def test_report_stacktraces(self):
```
dhermes: 1. `test_it` is a fine name
1. Is there a way to mock the output of `sys._current_frames().items()`?

logging/unit_tests/test_client.py
```diff
@@ -632,6 +632,20 @@ def test_setup_logging(self):
 
         setup_logging.assert_called()
 
+    def test_enable_shutdown_logging(self):
```
dhermes: Good test!

logging/google/cloud/logging/_environment_vars.py
waprin: dpme

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: done

logging/unit_tests/test__shutdown.py
waprin: 1. went with test success for consistency (and to not confuse with integration_test)
2. we could but it doesn't seem especially useful given we can just assert some things based on the actual stack trace

logging/unit_tests/test__shutdown.py
waprin: done

waprin:
@dhermes comments addressed and rebased

re: it belonging in the library, I could see it going either way, hopefully in future we can figure those things out in issue tracker ( #2819  ) before PR.

lukesneeringer:
> I'm on the fence on how to feel about a "very specific use case" as it pertains to this library. Would be interested in what @jonparrott and @lukesneeringer think.

My feeling on monitoring is basically "defer to @waprin when it comes to what he thinks he needs".

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,145 @@
...
+import sys
+import traceback
+
+from google.cloud.logging._environment_vars import APPENGINE_FLEXIBLE_ENV_VM
+from google.cloud.logging._environment_vars import APPENGINE_FLEXIBLE_ENV_FLEX
+from google.cloud.logging._environment_vars import GAE_SERVICE
+from google.cloud.logging._environment_vars import GAE_VERSION
+from google.cloud import _helpers
+
+import six
```
dhermes: Just noticed the location of this. It's 3rd party, so it should go above the `google.cloud` imports

logging/google/cloud/logging/client.py
```diff
@@ -327,3 +322,21 @@ def setup_logging(self, log_level=logging.INFO,
...
+
+        This method installs a SIGTERM handler that will report various
+        application metrics to Stackdriver Logging.
+
+        Currently the only supported option is thread stacktrace logging. This
+        will log all the stacktraces of all active threads.
+
+        :type thread_dump: bool
+        :param thread_dump: (Optional) When true, on SIGTERM the application
+                          will log the stacktrace of all running threads.
```
dhermes: Bad indent

logging/google/cloud/logging/client.py
```diff
@@ -327,3 +322,21 @@ def setup_logging(self, log_level=logging.INFO,
...
         setup_logging(handler, log_level=log_level,
                       excluded_loggers=excluded_loggers)
+
+    def enable_shutdown_logging(self, thread_dump=True):
+        """Enable shutdown report logging.
+
+        .. note::  This method is only supported on App Engine Flexible.
+
+        This method installs a SIGTERM handler that will report various
+        application metrics to Stackdriver Logging.
```
dhermes: Isn't saying "will report various application metrics" a bit of an overstatement? Don't we just report the full stack?

logging/google/cloud/logging/_shutdown.py
```diff
@@ -0,0 +1,145 @@
...
+                    expects SIGTERM, but the parameter is required to match
+                    the function signature for a signal handler.
+
+    :type frame: frame object
+    :param frame: The current stack frame. Unused parameter, this parameter
+                  is required to match the function signature for a signal
+                  handler.
+    """
+    traces = []
+    for thread_id, stack in sys._current_frames().items():
```
dhermes: @lukesneeringer @jonparrott I'm curious what your feelings are on using a non-public interface from the standard library

lukesneeringer: Biggest question: Does it work in pypy/Jython/IronPython?

dhermes: One data point:

```python
$ pypy
Python 2.7.12 (aff251e543859ce4508159dd9f1a82a2f553de00, Nov 12 2016, 08:50:18)
[PyPy 5.6.0 with GCC 6.2.0] on linux2
>>>> import sys
>>>> sys._current_frames()
{140637859424128: <__builtin__.fake_frame object at 0x00007fe8c63f3e18>}
>>>> 
```

waprin:
Based on offline convo going to close this PR and relocate it to https://github.com/googlecloudplatform/stackdriver-python (OOO this week but starting that migration when I return).

PR #2978, opened 2017-01-31
Vision: Add batch processing

### LINKED ISSUE #2897: Adding batch image processing to Vision API
The amount of data that Vision can return can be pretty deep and processing more than one image at a time adds another level to that.

Here's my proposed approach to work with multiple images with multiple feature types.

```python
        >>> images = client.batch()
        >>> for i in range(10):
        ...     images.add_image(client.image(source_uri='gs://%s.jpg' % (i,)))
        >>> images.detect_faces()
        {'0.jpg': [], '1.jpg': [],}
        >>>
        >>> # With multiple feature types.
        >>> features = [Feature(FeatureTypes.FACE_DETECTION, 5),
        ...             Feature(FeatureTypes.LOGO_DETECTION, 3)]
        >>> for i in range(10):
        ...     image = client.image(source_uri='gs://%s.jpg' % (i,))
        ...     images.add_image(image=image, features=features)
        >>> images.detect()
```

With a context manager it could look something like this...

```python
        >>> features = [Feature(FeatureTypes.FACE_DETECTION, 5),
        ...             Feature(FeatureTypes.LOGO_DETECTION, 3)]
        >>> with client.images() as images:
        ...     for i in range(10):
        ...         image = client.image(source_uri='gs://%s.jpg' % (i,))
        ...         images.add_image(image=image, features=features)
        ...     results = images.detect()
```

daspecster:
I wasn't sure the best way to add the context manager to `Batch` since there needs to be a result returned.  I'm not sure of a good way to do that from a context manager's `__exit__` without attaching the results to something.

Closes #2897

### DIFF
--- a/docs/vision-usage.rst
+++ b/docs/vision-usage.rst
@@ -287,8 +287,9 @@ image and determine the dominant colors in the image.
 Batch image detection
 *********************
 
-Multiple images can be processed with a single request by passing ``Image``s to
-``Client.batch(images)``.
+Multiple images can be processed with a single request by passing
+:class:`~google.cloud.vision.image.Image` to
+:meth:`~google.cloud.vision.client.Client.batch()`.
 
 .. code-block:: python
 
@@ -306,9 +307,10 @@ Multiple images can be processed with a single request by passing ``Image``s to
     >>> batch.add_image(image_one, [face_feature, logo_feature])
     >>> batch.add_image(image_two, [logo_feature])
     >>> results = batch.detect()
-    >>> for face in results[0].faces:
-    ...     print('=' * 40)
-    ...     print(face.joy)
+    >>> for image in results:
+    ...     for face in image.faces:
+    ...         print('=' * 40)
+    ...         print(face.joy)
     ========================================
     <Likelihood.VERY_LIKELY: 'VERY_LIKELY'>
     ========================================
--- a/docs/vision-usage.rst
+++ b/docs/vision-usage.rst
@@ -283,6 +283,38 @@ image and determine the dominant colors in the image.
     0.758658
 
 
+*********************
+Batch image detection
+*********************
+
+Multiple images can be processed with a single request by passing ``Image``s to
+``Client.batch(images)``.
+
+.. code-block:: python
+
+    >>> from google.cloud import vision
+    >>> from google.cloud.vision.feature import Feature
+    >>> from google.cloud.vision.feature import FeatureTypes
+    >>>
+    >>> client = vision.Client()
+    >>> batch = client.batch()
+    >>>
+    >>> image_one = client.image(source_uri='gs://my-test-bucket/image1.jpg')
+    >>> image_two = client.image(source_uri='gs://my-test-bucket/image2.jpg')
+    >>> face_feature = Feature(FeatureTypes.FACE_DETECTION, 2)
+    >>> logo_feature = Feature(FeatureTypes.LOGO_DETECTION, 2)
+    >>> batch.add_image(image_one, [face_feature, logo_feature])
+    >>> batch.add_image(image_two, [logo_feature])
+    >>> results = batch.detect()
+    >>> for face in results[0].faces:
+    ...     print('=' * 40)
+    ...     print(face.joy)
+    ========================================
+    <Likelihood.VERY_LIKELY: 'VERY_LIKELY'>
+    ========================================
+    <Likelihood.VERY_LIKELY: 'POSSIBLE'>
+
+
 ****************
 No results found
 ****************
--- a/system_tests/vision.py
+++ b/system_tests/vision.py
@@ -21,6 +21,8 @@ from google.cloud import exceptions
 from google.cloud import storage
 from google.cloud import vision
 from google.cloud.vision.entity import EntityAnnotation
+from google.cloud.vision.feature import Feature
+from google.cloud.vision.feature import FeatureTypes
 
 from system_test_utils import unique_resource_id
 from retry import RetryErrors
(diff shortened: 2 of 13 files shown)


vision/google/cloud/vision/batch.py
```diff
@@ -0,0 +1,55 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Batch multiple images into one request."""
+
+
+class Batch(object):
```
lukesneeringer: I am surprised that `Batch` is not usable as a context manager (as it is in, for example, pubsub). I would expect us to be consistent about that from library to library.

vision/google/cloud/vision/batch.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    def images(self):
+        """List of images to process.
+
+        :rtype: list
+        :returns: List of :class:`~google.cloud.vision.image.Image`.
+        """
+        return self._images
+
+    def detect(self):
```
lukesneeringer: Why does this not actually remove images from the `batch`?

```batch = Batch(client)
batch.add_image('/path/to/img1.jpg', [...])
batch.add_image('/path/to/img2.jpg', [...])
batch.detect()  # returns two annotations
batch.add_image('/path/to/img3.jpg', [...])
batch.detect()  # returns three annotations; I expected one
```

vision/google/cloud/vision/batch.py
lukesneeringer: I am realizing on further review that it appears to be a return value issue -- it is not just that a `commit` happens. There is no obvious way to get the _result_ at that point.

vision/google/cloud/vision/_gax.py
```diff
@@ -30,24 +30,25 @@ def __init__(self, client=None):
         self._client = client
         self._annotator_client = image_annotator_client.ImageAnnotatorClient()
 
-    def annotate(self, image, features):
+    def annotate(self, images):
```
lukesneeringer: Wait, is `features` not necessary here? Why?

daspecster: In this case `images` is more like a tuple of pairs.  e.g. `((image, [features]),)`

I'm thinking of following this up with a PR that attaches the feature types to the `Image` instance itself.

vision/google/cloud/vision/batch.py
daspecster: Good thought. Fixed in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2978/commits/39c7778709366bef4ed9b01fcfc71530a38c3d6a

vision/google/cloud/vision/batch.py
daspecster: Ok thanks! LMK if there's another strategy I should look at.

docs/vision-usage.rst
```diff
@@ -283,6 +283,38 @@ image and determine the dominant colors in the image.
     0.758658
 
 
+*********************
+Batch image detection
+*********************
+
+Multiple images can be processed with a single request by passing ``Image``s to
+``Client.batch(images)``.
```
dhermes: You can / should link to the actual `Image` class and the actual `Client.batch` method

docs/vision-usage.rst
```diff
@@ -283,6 +283,38 @@ image and determine the dominant colors in the image.
...
+    >>> batch = client.batch()
+    >>>
+    >>> image_one = client.image(source_uri='gs://my-test-bucket/image1.jpg')
+    >>> image_two = client.image(source_uri='gs://my-test-bucket/image2.jpg')
+    >>> face_feature = Feature(FeatureTypes.FACE_DETECTION, 2)
+    >>> logo_feature = Feature(FeatureTypes.LOGO_DETECTION, 2)
+    >>> batch.add_image(image_one, [face_feature, logo_feature])
+    >>> batch.add_image(image_two, [logo_feature])
+    >>> results = batch.detect()
+    >>> for face in results[0].faces:
```
dhermes: As I've continued to mention elsewhere, I am very :-1: on accessing `results[0]` without explaining what else is in `results`

system_tests/vision.py
```diff
@@ -507,3 +509,53 @@ def test_detect_properties_filename(self):
         image = client.image(filename=FACE_FILE)
         properties = image.detect_properties()
         self._assert_properties(properties)
+
+
+class TestVisionBatchProcessing(BaseVisionTestCase):
```
dhermes: Does this need it's own `TestCase` class?

vision/google/cloud/vision/_gax.py
dhermes: @daspecster 

1. Your docstring needs to make this clear.
1. You should not make breaking interface changes on unrelated PRs

vision/google/cloud/vision/batch.py
dhermes: @lukesneeringer @daspecster You can add a state to the batch and "tombstone" it once committed. Then the return value can be accessed as an attribute of the batch. The state makes it so that we don't accidentally over-ride the stored return value.

(We tombstone datastore transactions / maybe batches too in datastore.)

vision/google/cloud/vision/client.py
```diff
@@ -71,6 +72,14 @@ def __init__(self, project=None, credentials=None, http=None,
         else:
             self._use_gax = use_gax
 
+    def batch(self):
+        """Batch image processing.
```
dhermes: This is not a very useful sentence. Please at least make it a full sentence

vision/google/cloud/vision/_gax.py
daspecster: 1. Sure.
2. Not sure how this is a breaking change. A user should not call `annotate` directly. Usage of `Image.detect()` has not changed.

system_tests/vision.py
daspecster: I'm kind of following a pattern there. Which previous `TestCase` would you suggest putting it in?

vision/google/cloud/vision/_gax.py
dhermes: It's not breaking in the sense of `_gax.py` being public, but it is breaking in the sense of arbitrarily changing the architecture

system_tests/vision.py
dhermes: I just looked at the file and none of them really seem appropriate. Fine as-is

vision/google/cloud/vision/batch.py
daspecster: I see where datastore [attaches](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/ed0698fe20dee7e0590fbb4554ef77bd9c7291ea/datastore/google/cloud/datastore/batch.py#L287) the batch to the `_client`. I can do that if we want, but it's a fairly different usage pattern from the other image detections.

daspecster:
I just wanted to check. I think everything was addressed here?

lukesneeringer:
I would still like to understand why `annotate`'s signature changed, but everything else seems okay.

daspecster:
@lukesneeringer `annotate` isn't meant to be used directly. It should probably be named `_annotate` in hindsight. `detect()` is the documented method that calls `annotate()`.

That said, the signature change is required because multiple images need to be passed to `_vision_api.annotate()` in order to build the requests.

You build multiple `Request` objects and send them as a list and sending them as one request rather than making a request for each image.

Alternatively I could have duplicated the `annotate()` methods in `_http.py` and `_gax.py` and made the changes to handle multiple requests there, but there would have been a decent amount of duplicated code.

lukesneeringer:
Okay, that is fine.

daspecster:
Ok, I'll merge this in an hour or two. There's some other vision work that I want to get closer to finished before I rebase this into it.

PR #2984, opened 2017-02-06
Move dryRun property to from query to configuration level.

### LINKED ISSUE #2981: BigQuery: jobs created with run_async_query() and run with begin() ignores flag dry_run=True
The aim is to be able to dry run the queries to check how much it would cost us before we actually proceed with it. Hopefully we will be able to use it as a safety mechanism. 

I hope I am not misunderstanding the availability of the `dry_run` parameter as [listed on the docs](https://googlecloudplatform.github.io/google-cloud-python/stable/bigquery-job.html#google.cloud.bigquery.job.QueryJob.dry_run). I have set it, and it appears on the configs, **but it runs anyway**. 

1. Ubuntu 14.04

2. Python 2.7.6

3. google-cloud-python version
```
google-api-python-client==1.5.5
google-auth==0.5.0
google-auth-httplib2==0.0.2
google-cloud-bigquery==0.22.0
google-cloud-core==0.22.1
google-cloud-storage==0.22.0
googleapis-common-protos==1.5.0
```

4. Stacktrace if available
NIL

5. Steps to reproduce
- create job with run_async_query()
- set job.dry_run = True
- run it with job.begin()
- examine job when it is done

6. Code example
```
job = client.run_async_query(JOB_NAME, QUERY_STRING)
job.dry_run = True
job.begin()
print job.__dict__
```

daspecster:
Fixes #2981

Sets `dryRun` at `['configuration']['dryRun']` rather than `['configuration']['query']['dryRun]`.

### DIFF
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -31,6 +31,7 @@ class _SyncQueryConfiguration(object):
     Values which are ``None`` -> server defaults.
     """
     _default_dataset = None
+    _dry_run = None
     _max_results = None
     _timeout_ms = None
     _preserve_nulls = None
@@ -250,6 +251,11 @@ class QueryResults(object):
     https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#defaultDataset
     """
 
+    dry_run = _TypedProperty('dry_run', bool)
+    """See:
+    https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#dryRun
+    """
+
     max_results = _TypedProperty('max_results', six.integer_types)
     """See:
     https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#maxResults
@@ -313,6 +319,9 @@ class QueryResults(object):
 
         if self.use_legacy_sql is not None:
             resource['useLegacySql'] = self.use_legacy_sql
+
+        if self.dry_run is not None:
+            resource['dryRun'] = self.dry_run
 
         if len(self._udf_resources) > 0:
             resource[self._UDF_KEY] = [
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -31,7 +31,6 @@ class _SyncQueryConfiguration(object):
     Values which are ``None`` -> server defaults.
     """
     _default_dataset = None
-    _dry_run = None
     _max_results = None
     _timeout_ms = None
     _preserve_nulls = None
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/query.py
```diff
@@ -31,7 +31,6 @@ class _SyncQueryConfiguration(object):
     Values which are ``None`` -> server defaults.
     """
     _default_dataset = None
-    _dry_run = None
```
tseaver: Please back out changes to `Query` (which is for synchronous aka "non-job" queries):  the field is correctly handled here.  See https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#dryRun

dhermes:
I'll defer to @tseaver here

bigquery/unit_tests/test_query.py
```diff
@@ -307,7 +307,6 @@ def test_run_w_alternate_client(self):
         query.timeout_ms = 20000
         query.use_query_cache = False
         query.use_legacy_sql = True
-        query.dry_run = True
```
tseaver: The tests back out, too.

bigquery/google/cloud/bigquery/query.py
daspecster: So there's actually no property `dryRun` on the Query object.

I looked through the docs you linked pretty hard before I figured this out.

Here's a screenshot of the JSON from that doc page, but showing `dryRun` at the `configuration` level.

screen shot 2017-02-06 at 12 02 18 pm


I could be mistake, I'm not super familiar with BigQuery but I can't find anything about this option being on a specific Query.

tseaver: The `bigquery#queryRequest`resource (used for synchronous queries) is a completely different kind of resource than the `bigquery#job` resource used for async queries (which has a `configuration` section with a `query` subkey).  The `dryRun` key for `bigquery#queryRequest` is at the top-level, not nested at all.

daspecster: OK, I put that back(https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2984/commits/63044573b771ab751302b934dae5d577c285786e).

The example request JSON in the docs you linked doesn't show `dryRun` on the query object AFAICT.

tseaver: I'm not sure what you mean: it is right there on line 10 of your screenshot.  The `query` field in that resource is just the `SELECT` statement (a string).

daspecster: I thought that entire ["request body"](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#request-body) was the `['configuration']` object. I think I'm lost in which levels are where entirely.

tseaver: Your mistake is in assuming there is any logical consistency between the synchronous query request and the asynchronous query job. :)

PR #2986, opened 2017-02-06
Rename `document_from_url` to `document_from_gcs_url`.

lukesneeringer:
This change was requested by the NL team, since we only accept GCS URLs, and `document_from_url` implies anything.

Rather than a flat rename, I am having the original method warn a `DeprecationWarning`. We do not really have a process for deprecating or keeping up with them, so for the time being I am noting the date in the docstring.

### DIFF
--- a/language/google/cloud/language/client.py
+++ b/language/google/cloud/language/client.py
@@ -14,6 +14,8 @@
 
 """Basic client for Google Cloud Natural Language API."""
 
+import functools
+import warnings
 
 from google.cloud import client as client_module
 from google.cloud.language.connection import Connection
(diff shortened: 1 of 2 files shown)


language/google/cloud/language/client.py
```diff
@@ -110,3 +112,13 @@ def document_from_url(self, gcs_url,
         :returns: A document bound to this client.
         """
         return Document(self, gcs_url=gcs_url, doc_type=doc_type, **kwargs)
+
+    @functools.wraps(document_from_gcs_url)
+    def document_from_url(self, *args, **kwargs):
+        """Deprecated equivalent to document_from_gcs_url.
+        (Deprecated on 2017-02-06.)
```
dhermes: Can you add an empty line in between?

lukesneeringer: Sure.

language/unit_tests/test_client.py
```diff
@@ -115,11 +116,31 @@ def test_document_from_url_factory_explicit(self):
...
         self.assertEqual(document.encoding, encoding)
+
+    def test_document_from_url_deprecation(self):
+        import warnings
+
+        creds = make_mock_credentials()
+        client = self._make_one(credentials=creds, http=object())
+
+        Client = self._get_target_class()
+        with mock.patch.object(Client, 'document_from_gcs_url') as dfgu:
```
lukesneeringer: Yes, I know it is a short variable name. I would prefer to keep it because I find it easier to read than a line wrap.

PR #2989, opened 2017-02-09
Add create_resumable_upload_session

### LINKED ISSUE #1240: Start only resumable sessions to do browser-based uploading
Start a resumable upload in the server and then send the URL to the client to finish the upload.

Documentation: https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#resumable

Related issue in gcloud-node: https://github.com/GoogleCloudPlatform/gcloud-node/issues/641

jonparrott:
Resolves #1240

(Aside: wow this code is complicated)

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -403,7 +403,7 @@ class Blob(_PropertyMixin):
         :param content_type: Optional type of content being uploaded.
 
         :type chunk_size: int
-        :param size: The size of each chunk when doing resumable and media
+        :param chunk_size: The size of each chunk when doing resumable and media
             uploads.
 
         :type strategy: str
@@ -549,6 +549,10 @@ class Blob(_PropertyMixin):
                       ``NoneType``
         :param client: Optional. The client to use.  If not passed, falls back
                        to the ``client`` stored on the blob's bucket.
+
+        :rtype: str
+        :returns: The resumable upload session URL. The upload can be
+            completed by making an HTTP PUT request with the file's contents.
 
         :raises: :class:`ValueError` if size is not passed in and can not be
                  determined; :class:`google.cloud.exceptions.GoogleCloudError`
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -380,6 +380,110 @@ def download_as_string(self, client=None):
...
+        # API_BASE_URL and build_api_url).
+        connection = client._base_connection
+
+        content_type = (content_type or self._properties.get('contentType') or
+                        'application/octet-stream')
+
+        headers = {
+            'Accept': 'application/json',
+            'Accept-Encoding': 'gzip, deflate',
+            'User-Agent': connection.USER_AGENT,
```
lukesneeringer: Should this dictionary also add the `Content-Type` header (set to the value of `content_type` immediately above)?

jonparrott: Nope, these headers are for the initial request to start the upload which contains the blob metadata. The content type is `application/json` for this request. The `Upload` object takes care of sending the right content type when uploading the actual file.

lukesneeringer: That makes sense, thanks.

jonparrott:
@dhermes or @tseaver do either of you want to take a look at this before I merge?

dhermes:
I have lots of faith in @lukesneeringer

jonparrott:
@lukesneeringer tests are failing because of the operations/protobuf issue? any recommendations?

lukesneeringer:
I know what needs to happen here; it is related to googleapis/api-client-staging#159.
I will have an update in half an hour.

lukesneeringer:
I assert that [this commit][1] fixes the issue; attempting tests locally to prove it before cutting a release to PyPI.

lukesneeringer:
@jonparrott I merged that other PR and pushed it to PyPI; your tests should pass now (or if they fail, it will be for other reasons ;-)).

daspecster:
I restarted the build.

daspecster:
Seems to have done the trick!

PR #3002, opened 2017-02-13
Fix build failing with float precision mismatch

daspecster:
Build is failing due to float precision.
See: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/876

Clarification: `float` (i.e. single) vs. `double`. Double values are being cast down into `float` (which is correct, but is also a breaking change). For example

```python
>>> (0.9224355).hex()
'0x1.d849774256bcap-1'
>>> from google.cloud.grpc.vision.v1 import image_annotator_pb2
>>> e = image_annotator_pb2.EntityAnnotation(score=0.9224355)
>>> e.score
0.9224355220794678
>>> e.score.hex()
'0x1.d849780000000p-1'
```

### DIFF
--- a/speech/unit_tests/test_client.py
+++ b/speech/unit_tests/test_client.py
@@ -297,10 +297,10 @@ class TestClient(unittest.TestCase):
 
         alternatives = [{
             'transcript': 'testing 1 2 3',
-            'confidence': 0.125,
+            'confidence': 0.740234375,
         }, {
             'transcript': 'testing 4 5 6',
-            'confidence': 0.125,
+            'confidence': 0.6396484375,
         }]
         result = _make_result(alternatives)
 
@@ -500,16 +500,16 @@ class TestClient(unittest.TestCase):
 
         alternatives = [{
             'transcript': 'testing streaming 1 2 3',
-            'confidence': 0.125,
+            'confidence': 0.5888671875,
         }, {
             'transcript': 'testing streaming 4 5 6',
-            'confidence': 0.125,
+            'confidence': 0.28125,
         }]
         first_response = _make_streaming_response(
-            _make_streaming_result([], is_final=False, stability=0.125))
+            _make_streaming_result([], is_final=False, stability=0.314453125))
         second_response = _make_streaming_response(
             _make_streaming_result(alternatives, is_final=False,
-                                   stability=0.125))
+                                   stability=0.28125))
         last_response = _make_streaming_response(
             _make_streaming_result(alternatives, is_final=True,
                                    stability=0.9834534))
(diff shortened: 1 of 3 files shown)


speech/unit_tests/test_client.py
```diff
@@ -297,10 +297,10 @@ def test_sync_recognize_with_gax(self):
 
         alternatives = [{
             'transcript': 'testing 1 2 3',
-            'confidence': 0.9224355,
+            'confidence': 0.125,
         }, {
             'transcript': 'testing 4 5 6',
-            'confidence': 0.0123456,
+            'confidence': 0.125,
```
dhermes: Make this a little interesting? Suggest:  `9 / 32 == 0.28125`

Generated via:

```python
def f():
    import fractions
    import random
    exponent = random.randint(1, 10)
    numerator = 2 * random.randint(1, 2**(exponent - 1)) - 1
    frac = fractions.Fraction(numerator, 2**exponent)
    return frac, float(frac)
```

speech/unit_tests/test_client.py
```diff
@@ -500,16 +500,16 @@ def test_stream_recognize_interim_results(self):
 
         alternatives = [{
             'transcript': 'testing streaming 1 2 3',
-            'confidence': 0.9224355,
+            'confidence': 0.125,
         }, {
             'transcript': 'testing streaming 4 5 6',
-            'confidence': 0.0123456,
+            'confidence': 0.125,
```
dhermes: For all your other random values, I generated 20 more fractions that can be represented exactly in both `float` and `double` values

```
   0.1796875  # 23 / 128
   0.0546875  # 7 / 128
    0.390625  # 25 / 64
         0.5  # 1 / 2
0.5888671875  # 603 / 1024
         0.5  # 1 / 2
       0.875  # 7 / 8
  0.19140625  # 49 / 256
     0.84375  # 27 / 32
         0.5  # 1 / 2
      0.4375  # 7 / 16
         0.5  # 1 / 2
        0.75  # 3 / 4
        0.75  # 3 / 4
      0.4375  # 7 / 16
 0.314453125  # 161 / 512
    0.203125  # 13 / 64
0.6396484375  # 655 / 1024
 0.740234375  # 379 / 512
         0.5  # 1 / 2
```

dhermes:
@lukesneeringer @geigerj Are you aware of any code paths in `google.protobuf` that are casting these values from `double` to `float`?

geigerj:
@dhermes I'm not particularly familiar with the inner workings of `google.protobuf`, but that seems probable given that the source proto [declares `score` as a `float`](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto#L322), which in the proto wire format is 32-bit (https://developers.google.com/protocol-buffers/docs/encoding).

dhermes:
@daspecster LGTM. Let's get this merged ASAP. Squash if you'd like

dhermes:
Whatever issue caused this seems to have been a regression? I can't reproduce on Ubuntu 16.04 or OS X 10.11.

PR #3005, opened 2017-02-13
Check running status via 'job' property.

### LINKED ISSUE #3003: Check query's run status using property, not attribute
Split out from #3001, where @amirshavit reported:

> The run method checks if `self._job is not None`: to prevent a double invocation. It should however run that check with `self.job` instead (no underscore), as `self._job` is only set if you previously used `self.job`.

tseaver:
If the `_job` attribute is not set, but we have a `jobId` in our properties, we still don't want to re-run the query.

Closes #3003.

### DIFF
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -351,7 +351,7 @@ class QueryResults(object):
         :param client: the client to use.  If not passed, falls back to the
                        ``client`` stored on the current dataset.
         """
-        if self._job is not None:
+        if self.job is not None:
             raise ValueError("Query job is already running.")
 
         client = self._require_client(client)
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/query.py
```diff
@@ -351,7 +351,7 @@ def run(self, client=None):
         :param client: the client to use.  If not passed, falls back to the
                        ``client`` stored on the current dataset.
         """
-        if self._job is not None:
+        if self.job is not None:
```
dhermes: What's the distinction here?

tseaver: `job` is a property which lazily initializes `_job` from the server-supplied properties.

dhermes: Ah. That's important

dhermes:
LGTM

PR #3007, opened 2017-02-13
 Return int from 'total_rows'/'total_bytes_processed', if present.

### LINKED ISSUE #3004: Return integer or None from 'Query.total_rows'
Split out from #3001, where @amirshavit reported:

> It would be 'nicer' if the `total_rows property` return an integer (or None) rather than a string.

tseaver:
The first commit adds explicit unit test coverage for the scalar properties of `Query` (previously covered implicitly via tests of other methods).

The second commit adds the convert-to-int logic, because the back-end is marshalling integers as JSON-be-damned strings.

Closes #3004.

### DIFF
--- a/bigquery/unit_tests/test_query.py
+++ b/bigquery/unit_tests/test_query.py
@@ -124,9 +124,15 @@ class TestQueryResults(unittest.TestCase):
         self.assertEqual(query.complete, resource.get('jobComplete'))
         self.assertEqual(query.errors, resource.get('errors'))
         self.assertEqual(query.page_token, resource.get('pageToken'))
-        self.assertEqual(query.total_rows, resource.get('totalRows'))
-        self.assertEqual(query.total_bytes_processed,
-                         resource.get('totalBytesProcessed'))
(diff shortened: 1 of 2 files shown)


tseaver:
Crap, that was the wrong PR.  Can I get a post-facto review, or should I revert?

bigquery/unit_tests/test_query.py
```diff
@@ -251,6 +257,120 @@ def test_job_w_jobid(self):
...
+        resource = {'errors': ERRORS}
+        query._set_properties(resource)
+        self.assertEqual(query.errors, ERRORS)
+
+    def test_name_missing(self):
+        client = _Client(self.PROJECT)
+        query = self._make_one(self.QUERY, client)
+        self.assertIsNone(query.name)
+
+    def test_name_broken_job_reference(self):
```
dhermes: These are just for the sake of being thorough?

tseaver: Yep.

PR #3033, opened 2017-02-17
Renaming all remaining connection.py modules as _http.py.

dhermes:


dhermes:
DERP I didn't remove docs RST reference to the modules. Just pushed a fix. Will merge (now) once Travis goes green.

PR #3035, opened 2017-02-17
Adding ability to send version info header on HTTP requests.

dhermes:
Added an "extra headers" feature to enable this. I am not a fan
of changing `Connection()` so haphazardly, but I hope to
completely re-factor / destory `Connection()` in the near-term
so I am less worried.

This only adds the storage and datastore header info, for the
purposes of a simple review. Once we agree on the approach,
I can add support in the other subpackages.

### DIFF
--- a/datastore/google/cloud/datastore/_http.py
+++ b/datastore/google/cloud/datastore/_http.py
@@ -16,6 +16,7 @@
 
 import contextlib
 import os
+from pkg_resources import get_distribution
 
 from google.rpc import status_pb2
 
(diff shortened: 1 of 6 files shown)


datastore/google/cloud/datastore/_http.py
```diff
@@ -61,6 +62,9 @@
 
 _DISABLE_GRPC = os.getenv(DISABLE_GRPC, False)
 _USE_GRPC = _HAVE_GRPC and not _DISABLE_GRPC
+_DATASTORE_DIST = get_distribution('google-cloud-datastore')
+_CLIENT_INFO = connection_module.CLIENT_INFO_TEMPLATE.format(
+    _DATASTORE_DIST.version)
```
tseaver: The bare version number seems odd:  I guess the fact that `gl-python` is there plus the fact that we are talking to the Datastore API is enough information?

dhermes: You've succinctly summarized the exact conversation that @lukesneeringer and I had yesterday :grinning: 

The "right" thing to do would be to say something like `core-version/X.Y.Z FOO-version/a.b.c` but that decision is out of our hands. So we are using `gccl/a.b.c` (i.e. `FOO`'s version) and relying on the actual backend service to determine which subpackage we meant.

dhermes:
Thanks @tseaver. Will hold off for @lukesneeringer review before merging

dhermes:
Will follow up with a PR for all other HTTP impls and then a separate PR getting this to work with GAPIC generated surfaces

PR #3036, opened 2017-02-17
Moving datastore GAX/gRPC helpers out of _http and into dedicated module.

dhermes:
This leaves `_http.py` closer to what it should look like, but not all the way there (it is doing some of the transport choosing that should be done in `client.py`)


datastore/google/cloud/datastore/_gax.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2014 Google Inc.
```
daspecster: /2014/2017/

datastore/google/cloud/datastore/_gax.py
```diff
@@ -0,0 +1,197 @@
...
+    StatusCode.UNIMPLEMENTED: exceptions.MethodNotImplemented,
+    StatusCode.INTERNAL: exceptions.InternalServerError,
+    StatusCode.UNAVAILABLE: exceptions.ServiceUnavailable,
+    StatusCode.DATA_LOSS: exceptions.InternalServerError,
+}
+
+
+@contextlib.contextmanager
+def _grpc_catch_rendezvous():
+    """Re-map gRPC exceptions that happen in context.
```
daspecster: /Re-map/Remap/

datastore/google/cloud/datastore/_gax.py
dhermes: How did I miss that? DERP

datastore/google/cloud/datastore/_gax.py
dhermes: Was copy-pasted, but you got it.

dhermes:
@daspecster I folded those changes into the original commit (since they were super tiny)

daspecster:
SGTM! Thanks! I know they were nitpicky.

The rest LGTM but I feel like someone else should take a crack at it.

datastore/google/cloud/datastore/_http.py
```diff
@@ -14,46 +14,21 @@
...
-from google.cloud._helpers import make_secure_stub
 from google.cloud import _http as connection_module
 from google.cloud.environment_vars import DISABLE_GRPC
 from google.cloud.environment_vars import GCD_HOST
 from google.cloud import exceptions
 from google.cloud.grpc.datastore.v1 import datastore_pb2 as _datastore_pb2
 try:
-    from grpc import StatusCode
-    from google.cloud.grpc.datastore.v1 import datastore_pb2_grpc
+    from google.cloud.datastore._gax import _DatastoreAPIOverGRPC
```
tseaver: Isn't this an unused import?

datastore/google/cloud/datastore/_http.py
```diff
@@ -14,46 +14,21 @@
...
+    _HAVE_GRPC = True
```
tseaver: Why are we computing `_HAVE_GRPC` in the `_http` module?  Wouldn't it be more natural to move it to `_gax`?

dhermes: It seems natural (to me) to have imports "be free" in `_gax.py` and then let the potential users deal with the import error. Though I think that should happen in `client.py`, not `_http.py` (see my comment above)

datastore/google/cloud/datastore/_http.py
dhermes: No. This module is doing the work of `client.py`, so it actually creates one of these objects

datastore/google/cloud/datastore/_gax.py
```diff
@@ -0,0 +1,197 @@
...
+
+    Remaps gRPC exceptions to the classes defined in
+    :mod:`~google.cloud.exceptions` (according to the description
+    in `code.proto`_).
+    """
+    try:
+        yield
+    except exceptions.GrpcRendezvous as exc:
+        error_code = exc.code()
+        error_class = _GRPC_ERROR_MAPPING.get(error_code)
```
lukesneeringer: Is this intended to be an alternative implementation to #2936?

datastore/google/cloud/datastore/_http.py
lukesneeringer: So, the ultimate idea is that:

  - Developer imports `from x.y.z.v1.client import Client`
  - The `client` module tries to import `_gax`, wrapped in a try-except, and multiplexes based on whether GRPC is available?

datastore/google/cloud/datastore/_gax.py
dhermes: No, this impl. predates that discussion. (The code in `_gax.py` in this PR is just being cut-and-pasted from `_http.py`.) However, it solves essentially the same problem and has the same flaws discussed there: the mapping to HTTP codes is imperfect / loses some data.

datastore/google/cloud/datastore/_http.py
dhermes: Indeed.

PR #3039, opened 2017-02-17
Spanner, BigQuery, Core, Error Reporting, Logging formatting style updates.

daspecster:
Mostly just changing /_makeOne/_make_one`/ and adding a space after the imports of unit_tests.

### DIFF
--- a/error_reporting/unit_tests/test_client.py
+++ b/error_reporting/unit_tests/test_client.py
@@ -107,8 +107,8 @@ class TestClient(unittest.TestCase):
         self.assertIsNotNone(target.report_errors_api)
 
     @mock.patch('google.cloud.error_reporting.client.make_report_error_api')
-    def test_report_exception_with_service_version_in_constructor(self,
-                                                                  make_client):
+    def test_report_exception_with_service_version_in_constructor(
+            self, make_client):
         CREDENTIALS = _make_credentials()
         SERVICE = "notdefault"
         VERSION = "notdefaultversion"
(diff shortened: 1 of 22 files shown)


error_reporting/unit_tests/test_client.py
```diff
@@ -107,9 +107,8 @@ def test_report_exception_wo_gax(self):
         self.assertIsNotNone(target.report_errors_api)
 
     @mock.patch('google.cloud.error_reporting.client.make_report_error_api')
-    def test_report_exception_with_service_version_in_constructor(
-            self,
-            make_client):
+    def test_report_exception_with_service_version_in_constructor(self,
+                                                                  make_client):
```
dhermes: Instead:

```python
    def test_report_exception_with_service_version_in_constructor(
        self, make_client):
```

daspecster: So I tried that but my local lint choked on that(atom.io, linter). I'll play with it some more though.

daspecster: So interestingly this makes everything happy.

```
def test_report_exception_with_service_version_in_constructor(
        self, make_client):
```

8 spaces instead of 4.

daspecster:
This looks interesting but it's a flake8 plugin.

https://github.com/PyCQA/flake8-import-order

daspecster:
Ah, I forgot to rebase that last commit out. Crap.

PR #3044, opened 2017-02-20
Document correct method name: 'list_instance_configs'

supertom:
In the docs, the method is listed as `list_instance_configurations`, but in the [code](https://github.com/supertom/google-cloud-python/blob/d80953d520a51a91fbc43c22fc2821a01900e5e6/spanner/google/cloud/spanner/client.py#L182) (and as in version 0.23.0 installed by pip) it is `list_instance_configs`


tseaver:
@supertom Thanks for the patch!

PR #3045, opened 2017-02-21
Add gccl metrics header to Spanner.

lukesneeringer:
This PR adds the `gccl/x.y.z` portion of the metrics header in the Spanner client library.

### DIFF
--- a/spanner/google/cloud/spanner/__init__.py
+++ b/spanner/google/cloud/spanner/__init__.py
@@ -15,6 +15,10 @@
 """Cloud Spanner API package."""
 
 
+import pkg_resources
+__version__ = pkg_resources.get_distribution('google-cloud-spanner').version
+
+
 from google.cloud.spanner.client import Client
 
 from google.cloud.spanner.keyset import KeyRange
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -30,6 +30,7 @@ import six
 from google.cloud.exceptions import Conflict
 from google.cloud.exceptions import NotFound
 from google.cloud.operation import register_type
+from google.cloud.spanner import __version__
 from google.cloud.spanner._helpers import _options_with_prefix
 from google.cloud.spanner.batch import Batch
 from google.cloud.spanner.session import Session
@@ -179,7 +180,8 @@ class Database(object):
     def spanner_api(self):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
-            self._spanner_api = SpannerClient()
+            self._spanner_api = SpannerClient(lib_name='gccl',
+                                              lib_version=__version__)
         return self._spanner_api
 
     def __eq__(self, other):
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -180,8 +180,8 @@ class Database(object):
     def spanner_api(self):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
-            self._spanner_api = SpannerClient(lib_name='gccl',
-                                              lib_version=__version__)
+            self._spanner_api = SpannerClient(
+                lib_name='gccl', lib_version=__version__)
         return self._spanner_api
 
     def __eq__(self, other):
--- a/spanner/setup.py
+++ b/spanner/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-spanner',
-    version='0.23.1',
+    version='0.23.2',
     description='Python Client for Cloud Spanner',
     long_description=README,
     namespace_packages=[
--- a/spanner/unit_tests/test_client.py
+++ b/spanner/unit_tests/test_client.py
@@ -107,9 +107,6 @@ class TestClient(unittest.TestCase):
         self._constructor_test_helper(expected_scopes, creds)
 
     def test_admin_api_lib_name(self):
-        """Establish that the lib_name and lib_version are passed to
-        the database and instance API objects in the GAPIC.
-        """
         from google.cloud.spanner import __version__
         from google.cloud.gapic.spanner_admin_database import v1 as db
         from google.cloud.gapic.spanner_admin_instance import v1 as inst
--- a/spanner/unit_tests/test_client.py
+++ b/spanner/unit_tests/test_client.py
@@ -111,7 +111,6 @@ class TestClient(unittest.TestCase):
         the database and instance API objects in the GAPIC.
         """
         from google.cloud.spanner import __version__
-        from google.cloud.spanner import client as MUT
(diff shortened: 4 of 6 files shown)


spanner/unit_tests/test_client.py
```diff
@@ -114,7 +151,8 @@ def test_instance_admin_api(self):
         client = self._make_one(project=self.PROJECT, credentials=creds)
 
         class _Client(object):
-            pass
+            def __init__(self, *args, **kwargs):
+                pass
```
tseaver: This new constructor should record the values passed in, so that we can make assertions about them.

spanner/unit_tests/test_client.py
```diff
@@ -131,7 +169,8 @@ def test_database_admin_api(self):
...
```
tseaver: Ditto.

tseaver:
The PR should also be rebased against current `master` to get rid of the spurious commits.

spanner/unit_tests/test_client.py
lukesneeringer: Is there a reason this is not a `mock.MagicMock()`?

spanner/google/cloud/spanner/__init__.py
```diff
@@ -15,6 +15,10 @@
 """Cloud Spanner API package."""
 
 
+import pkg_resources
+__version__ = pkg_resources.get_distribution('google-cloud-spanner').version
```
dhermes: @lukesneeringer I considered doing this for all sub-packages. Seems you are :+1: on this approach?

Also: defining `__version__` should come **after** all the imports?

spanner/google/cloud/spanner/database.py
```diff
@@ -179,7 +180,8 @@ def ddl_statements(self):
     def spanner_api(self):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
-            self._spanner_api = SpannerClient()
+            self._spanner_api = SpannerClient(lib_name='gccl',
+                                              lib_version=__version__)
```
dhermes: @jonparrott Has convinced me of the merits of "one continuation, all continuation", so here

```python
            self._spanner_api = SpannerClient(
                lib_name='gccl', lib_version=__version__)
```

spanner/setup.py
```diff
@@ -59,7 +59,7 @@
 
 setup(
     name='google-cloud-spanner',
-    version='0.23.1',
+    version='0.23.2',
```
dhermes: Do the GAPIC deps not also need to be revved to include `lib_name`/`lib_version`?

spanner/unit_tests/test_client.py
```diff
@@ -106,6 +106,43 @@ def test_constructor_credentials_wo_create_scoped(self):
         expected_scopes = None
         self._constructor_test_helper(expected_scopes, creds)
 
+    def test_admin_api_lib_name(self):
+        """Establish that the lib_name and lib_version are passed to
+        the database and instance API objects in the GAPIC.
+        """
```
dhermes: :-1: on docstring for unit test (and on multi-line single-line docstring)

Can't recall how we resolved the discussion before

spanner/unit_tests/test_client.py
```diff
@@ -114,14 +151,17 @@ def test_instance_admin_api(self):
         client = self._make_one(project=self.PROJECT, credentials=creds)
 
         class _Client(object):
-            pass
+            def __init__(self, *args, **kwargs):
+                self.args = args
+                self.kwargs = kwargs
```
dhermes: You could (if you like) just dump `_Client` and use a mock?

spanner/unit_tests/test_client.py
```diff
@@ -106,6 +106,43 @@ def test_constructor_credentials_wo_create_scoped(self):
         expected_scopes = None
         self._constructor_test_helper(expected_scopes, creds)
 
+    def test_admin_api_lib_name(self):
```
dhermes: Not worth just checking lib name on one of the existing (e.g. constructor) unit test cases?

spanner/google/cloud/spanner/__init__.py
lukesneeringer: Yeah, that was my plan.

`__version__ =` has to come where it does or the other modules are unable to `from google.cloud.spanner import __version__` without a circular import.

spanner/setup.py
lukesneeringer: They already have been.

spanner/unit_tests/test_client.py
lukesneeringer: I did that also, but I prefer having an explicit test for new work.

spanner/unit_tests/test_client.py
lukesneeringer: We agreed to disagree.

spanner/unit_tests/test_client.py
lukesneeringer: I asked the same question. I _think_ the reason @tseaver did what he did was for the instance check. I am doing path of least resistance.

spanner/google/cloud/spanner/__init__.py
dhermes: Ah good call

spanner/google/cloud/spanner/database.py
lukesneeringer: I hate both equally, so sure. :-)

spanner/unit_tests/test_client.py
tseaver: Docstrings are about runtime discoverability, which is an actual downside for testcase methods:  they have caused testrunners to screw up display of failed tests in the past.  I have no idea whether current testrunners have this bug, because I resolved to avoid them forevermore after that, and use more descriptive method names and/or comments instead.

lukesneeringer:
Travis seems to be hung? Anyone know if I can do anything about this?

daspecster:
@lukesneeringer there isn't anything you can do about Travis AFAICT. CircleCI should have covered it though so you should be clear to merge anyway in this case.

lukesneeringer:
Okay, merging once @tseaver gives assent.

spanner/unit_tests/test_database.py
```diff
@@ -188,7 +190,10 @@ def test_spanner_api_property(self):
         _client = object()
         _clients = [_client]
 
-        def _mock_spanner_client():
+        def _mock_spanner_client(*args, **kwargs):
+            self.assertIsInstance(args, (list, tuple))
```
dhermes: Isn't this always a tuple, per the spec?

dhermes:
LGTM

tseaver:
@lukesneeringer FWIW, your docstring-in-a-testcase here will be literally the only one in the whole codebase.  If you feel that strongly about it, go ahead and merge.

lukesneeringer:
I went ahead and removed it.

PR #3048, opened 2017-02-22
Speech Client Library Header

lukesneeringer:


speech/google/cloud/speech/_gax.py
```diff
@@ -45,12 +46,17 @@ def __init__(self, client=None):
...
+            channel=channel,
+            lib_name='gccl',
+            lib_version=__version__,
+        )
         self._operations_stub = make_secure_stub(
             credentials,
             DEFAULT_USER_AGENT,
             operations_grpc.OperationsStub,
-            OPERATIONS_API_HOST)
+            OPERATIONS_API_HOST,
```
dhermes: Editorializing in a PR, how dare ye!

speech/setup.py
```diff
@@ -51,12 +51,12 @@
...
     'google-cloud-core >= 0.23.0, < 0.24dev',
     'grpcio >= 1.0.2, < 2.0dev',
-    'gapic-google-cloud-speech-v1beta1 >= 0.15.0, < 0.16dev',
+    'gapic-google-cloud-speech-v1beta1 >= 0.15.1, < 0.16dev',
 ]
 
 setup(
     name='google-cloud-speech',
-    version='0.22.0',
+    version='0.23.0',
```
dhermes: Hold off on this until the actual version bump / release PR goes out?

speech/unit_tests/test_client.py
```diff
@@ -258,9 +258,9 @@ def make_channel(*args):
             channel_args.append(args)
             return channel_obj
 
-        def speech_api(channel=None):
+        def speech_api(channel=None, **kwargs):
             return _MockGAPICSpeechAPI(response=_make_sync_response(),
-                                       channel=channel)
+                                       channel=channel, **kwargs)
```
dhermes: All of these changes are probably a sign of bad mocks?

lukesneeringer: They totally are, but I am not trying to re-engineer that right now.

speech/setup.py
lukesneeringer: I was just going to bump each one as I hit it. I could go either way.

speech/unit_tests/test_client.py
tseaver: Why not just add the "correct" parameters to `_MockGAPICSpeechAPI.__init__`?

lukesneeringer: Because we accept something like 15 parameters, and more may be added in the future, and that should be able to be done without erroring out unrelated tests.

PR #3050, opened 2017-02-22
GAPIC Header Consistency: Vision

lukesneeringer:
@daspecster Would you please take a longer look at this one? I found the "two clients" thing really confusing and am not sure I did this correctly.

In particular, I am pretty sure the existing code threw away the `credentials` argument when being used with GAX. (Not sure how anyone would successfully set any of the _other_ arguments, either.) I updated this and added a test for it.

### DIFF
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -85,7 +85,7 @@ def _to_gapic_image(image):
     :type image: :class:`~google.cloud.vision.image.Image`
     :param image: Local ``Image`` class to be converted to gRPC ``Image``.
 
-    :rtype: :class:`~google.cloudvision.v1.image_annotator_pb2.Image`
+    :rtype: :class:`~google.cloud.proto.vision.v1.image_annotator_pb2.Image`
     :returns: gRPC ``Image`` converted from
               :class:`~google.cloud.vision.image.Image`.
     """
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -17,6 +17,7 @@
 from google.cloud.gapic.vision.v1 import image_annotator_client
 from google.cloud.grpc.vision.v1 import image_annotator_pb2
 
+from google.cloud.vision import __version__
 from google.cloud.vision.annotations import Annotations
 
 
@@ -25,10 +26,14 @@ class _GAPICVisionAPI(object):
 
     :type client: :class:`~google.cloud.vision.client.Client`
     :param client: Instance of ``Client`` object.
+
+    :type kwargs: dict
+    :param kwargs: Additional keyword arguments are sent to the GAPIC client.
     """
-    def __init__(self, client=None):
+    def __init__(self, client=None, credentials=None):
         self._client = client
-        self._annotator_client = image_annotator_client.ImageAnnotatorClient()
+        self._annotator_client = image_annotator_client.ImageAnnotatorClient(
+            credentials=credentials, lib_name='gccl', lib_version=__version__)
 
     def annotate(self, images):
         """Annotate images through GAX.
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -30,10 +30,11 @@ class _GAPICVisionAPI(object):
     :type kwargs: dict
     :param kwargs: Additional keyword arguments are sent to the GAPIC client.
     """
-    def __init__(self, client=None, credentials=None):
+    def __init__(self, client=None):
         self._client = client
         self._annotator_client = image_annotator_client.ImageAnnotatorClient(
-            credentials=credentials, lib_name='gccl', lib_version=__version__)
+            credentials=client._credentials, lib_name='gccl',
+            lib_version=__version__)
 
     def annotate(self, images):
         """Annotate images through GAX.
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -15,7 +15,7 @@
 """GAX Client for interacting with the Google Cloud Vision API."""
 
 from google.cloud.gapic.vision.v1 import image_annotator_client
-from google.cloud.grpc.vision.v1 import image_annotator_pb2
+from google.cloud.proto.vision.v1 import image_annotator_pb2
 
 from google.cloud.vision import __version__
 from google.cloud.vision.annotations import Annotations
@@ -70,7 +70,7 @@ def _to_gapic_feature(feature):
     :param feature: Local ``Feature`` class to be converted to gRPC ``Feature``
                     instance.
 
-    :rtype: :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.Feature`
+    :rtype: :class:`~google.cloudvision.v1.image_annotator_pb2.Feature`
     :returns: gRPC ``Feature`` converted from
               :class:`~google.cloud.vision.feature.Feature`.
     """
@@ -85,7 +85,7 @@ def _to_gapic_image(image):
     :type image: :class:`~google.cloud.vision.image.Image`
     :param image: Local ``Image`` class to be converted to gRPC ``Image``.
 
-    :rtype: :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.Image`
+    :rtype: :class:`~google.cloudvision.v1.image_annotator_pb2.Image`
     :returns: gRPC ``Image`` converted from
               :class:`~google.cloud.vision.image.Image`.
     """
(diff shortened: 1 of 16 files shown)


vision/google/cloud/vision/_gax.py
```diff
@@ -25,10 +26,14 @@ class _GAPICVisionAPI(object):
 
     :type client: :class:`~google.cloud.vision.client.Client`
     :param client: Instance of ``Client`` object.
+
+    :type kwargs: dict
+    :param kwargs: Additional keyword arguments are sent to the GAPIC client.
```
dhermes: Copy-pasta? Meant to be `credentials`?

vision/google/cloud/vision/client.py
```diff
@@ -109,7 +109,8 @@ def _vision_api(self):
         """
         if self._vision_api_internal is None:
             if self._use_gax:
-                self._vision_api_internal = _GAPICVisionAPI(self)
+                self._vision_api_internal = _GAPICVisionAPI(self,
+                    credentials=None)
```
dhermes: Instead:

```python
                self._vision_api_internal = _GAPICVisionAPI(
                    self, credentials=None)
```

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,47 @@ def test_ctor(self):
             api = self._make_one(client)
         self.assertIs(api._client, client)
 
+    def test_gapic_credentials(self):
+        from google.cloud.gapic.vision.v1.image_annotator_client import (
+            ImageAnnotatorClient)
+
+        from .test_client import _make_credentials
```
dhermes: I am :-1: on unit test modules being imported. Not so bad to just copy-pasta the `_make_credentials` method?

We could (long-term) have a unit test helpers package that we include?

daspecster:
Taking a look now.

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,47 @@ def test_ctor(self):
...
+
+        # Mock the GAPIC ImageAnnotatorClient, whose arguments we
+        # want to check.
+        with mock.patch.object(ImageAnnotatorClient, '__init__') as iac:
+            iac.return_value = None
+
+            # Create the GAX client.
+            credentials = _make_credentials()
+            self._make_one(client=object(), credentials=credentials)
```
dhermes: Need to check / use the return value?

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,47 @@ def test_ctor(self):
...
+
+            # Assert that the GAPIC constructor was called once, and
+            # that the credentials were sent.
+            iac.assert_called_once()
+            self.assertIs(iac.mock_calls[0][2]['credentials'], credentials)
```
dhermes: Can you 

- make an assertion on `len(iac.mock_calls)`
- Unpack the call as `name, pos_args, kwargs = iac.mock_calls[0]` (throwaway things as you see fit)

vision/unit_tests/test__gax.py
```diff
@@ -32,6 +32,47 @@ def test_ctor(self):
...
+            iac.return_value = None
+
+            # Create the GAX client.
+            self._make_one(client=object(), credentials=_make_credentials())
+
+            # Assert that the GAPIC constructor was called once, and
+            # that lib_name and lib_version were sent.
+            iac.assert_called_once()
+            self.assertEqual(iac.mock_calls[0][2]['lib_name'], 'gccl')
+            self.assertEqual(iac.mock_calls[0][2]['lib_version'], __version__)
```
dhermes: Ditto here on being more explicit about what's in `mock_calls`

vision/unit_tests/test__gax.py
lukesneeringer: > make an assertion on len(iac.mock_calls)

  * `iac.assert_called_once()` is equivalent to `self.assertEqual(len(iac.mock_calls), 1)`
  * Sure. That would be clearer.

vision/google/cloud/vision/_gax.py
```diff
@@ -25,10 +26,14 @@ class _GAPICVisionAPI(object):
...
     """
-    def __init__(self, client=None):
+    def __init__(self, client=None, credentials=None):
         self._client = client
-        self._annotator_client = image_annotator_client.ImageAnnotatorClient()
+        self._annotator_client = image_annotator_client.ImageAnnotatorClient(
+            credentials=credentials, lib_name='gccl', lib_version=__version__)
```
daspecster: I think this is where things should change.

I think you want `client.credentials` instead of the passed in credentials that you added.

You can see [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/2c608f95d7d5db2cbfdce77770f523a489b513f4/vision/google/cloud/vision/client.py#L65) that the `credentials` can be passed into the client on initialization which is the pattern the other services use.

vision/google/cloud/vision/_gax.py
lukesneeringer: Updating argument and not docstring fail. I considered kwargs and then decided against it.

vision/unit_tests/test__gax.py
lukesneeringer: Yeah, I was thinking a `utils` module in the future might be better.
In principle, I am probably more allergic to copy-paste than the import (they both stink), but since it is a two-line function, I will copy-paste.

lukesneeringer:
@daspecster Please verify now. (Also, obviously, waiting on tests.)

vision/google/cloud/vision/_gax.py
```diff
@@ -64,7 +70,7 @@ def _to_gapic_feature(feature):
     :param feature: Local ``Feature`` class to be converted to gRPC ``Feature``
                     instance.
 
-    :rtype: :class:`~google.cloud.grpc.vision.v1.image_annotator_pb2.Feature`
+    :rtype: :class:`~google.cloudvision.v1.image_annotator_pb2.Feature`
```
daspecster: `google.cloudvision`/`google.cloud.vision`

vision/google/cloud/vision/client.py
```diff
@@ -109,7 +109,8 @@ def _vision_api(self):
         """
         if self._vision_api_internal is None:
             if self._use_gax:
-                self._vision_api_internal = _GAPICVisionAPI(self)
+                self._vision_api_internal = _GAPICVisionAPI(
+                    self, credentials=None)
```
daspecster: I think you would need to remove `credentials` here as well right?

lukesneeringer: Yep. Thanks.

vision/google/cloud/vision/annotations.py
```diff
@@ -99,7 +99,7 @@ def from_api_repr(cls, response):
     def from_pb(cls, response):
         """Factory: construct an instance of ``Annotations`` from protobuf.
 
-        :type response: :class:`~google.cloud.grpc.vision.v1.\
+        :type response: :class:`~google.cloudvision.v1.\
```
daspecster: Sorry I thought were doing a blanket find/replace. There are a bunch of these.

lukesneeringer: Yep, I did a blanket find/replace incorrectly. Just caught it in tox and fixed it.

daspecster:
Pending travis/circle this LGTM.  I would recommend running the system tests as well if you haven't yet.

lukesneeringer:
The Travis errors suggest the credentials are not being passed correctly.

vision/unit_tests/test__gax.py
```diff
@@ -32,12 +37,55 @@ def test_ctor(self):
...
+        from google.cloud.vision import Client
+
+        # Mock the GAPIC ImageAnnotatorClient, whose arguments we
+        # want to check.
+        with mock.patch.object(ImageAnnotatorClient, '__init__') as iac:
+            iac.return_value = None
+
+            # Create the GAX client.
+            credentials = _make_credentials()
+            client = Client(credentials=credentials)
```
daspecster: The build is failing when trying to [determine the project](https://travis-ci.org/GoogleCloudPlatform/google-cloud-python/builds/204324421#L870).

I think you'll have to set `project` when you create the client. `client = Client(project='my-project', credentials=credentials)`

daspecster: Really though that's a great error message in this case.

```pyton
DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or
explicitly create credential and re-run the application. For more
information, please see
https://developers.google.com/accounts/docs/application-default-credentials.
```

daspecster:
@lukesneeringer cool looks like that worked. Just a lint issue now.

vision/google/cloud/vision/_gax.py
dhermes: Can haz fix?

Issue #1660, opened 2016-03-25
support for predefined ACL to be set directly on the upload request


pdknsk:
This is a follow-up to #651.

This request may be a bit too early, as it will only really matter when #1185 is fixed. Right now with `blob.acl.save_predefined` the object ACL can be set to a predefined value separately, like the _metadata_ can be with `blob.patch`. The idea is to set everything with the upload request, to not have separate request.

pdknsk:
I've made a simple patch for common use cases, on top of [another patch](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1185#issuecomment-287584661). It does not check the name of the predefined ACL, nor does it [translate](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/b035bf78d726ddccc821e8884e6a4eda48e93e56/storage/google/cloud/storage/acl.py#L179) from XML to JSON names.

```Diff
--- a/blob.py
+++ b/blob.py
@@ -407,7 +407,8 @@
 
     def _create_upload(
             self, client, file_obj=None, size=None, content_type=None,
-            chunk_size=None, strategy=None, extra_headers=None):
+            chunk_size=None, strategy=None, extra_headers=None,
+            predefined_acl=None):
         """Helper for upload methods.
 
         Creates a :class:`google.cloud.core.streaming.Upload` object to handle
@@ -482,6 +483,7 @@
             upload.strategy = RESUMABLE_UPLOAD
 
         url_builder = _UrlBuilder(
+            predefined_acl=predefined_acl,
             bucket_name=self.bucket.name,
             object_name=self.name)
         upload_config = _UploadConfig()
@@ -497,6 +499,9 @@
 
         if self._properties:
             headers['content-type'] = 'application/json'
+            # 409 Cannot provide both a predefinedAcl and access controls
+            if predefined_acl and 'acl' in self._properties:
+                del self._properties['acl']
             request.body = json.dumps(self._properties)
 
         upload.configure_request(upload_config, request, url_builder)
@@ -525,7 +530,8 @@
                                  error_info=request.url)
 
     def upload_from_file(self, file_obj, rewind=False, size=None,
-                         content_type=None, num_retries=6, client=None):
+                         content_type=None, num_retries=6, client=None,
+                         predefined_acl=None):
         """Upload the contents of this blob from a file-like object.
 
         The content type of the upload will either be
@@ -621,7 +627,7 @@
         upload, request, _ = self._create_upload(
[...]

Issue #754, opened 2015-03-24
STORAGE: Send blob metadata in upload requests


dhermes:
When `self._changes` is not empty, we should upgrade a `media` request to a [multipart upload](https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#multipart):

> If you have metadata that you want to send along with the data to upload, you can make a single `multipart/related` request. This is a good choice if the data you are sending is small enough to upload again in its entirety if the connection fails.

whereas for resumable uploads, we should

> Make an initial request to the upload URI that includes the metadata, if any.

@craigcitro Is there a way to set the body of the initial request in a resumable upload?

H/T to @pdknsk for #536

craigcitro:
yes -- if you call `upload.ConfigureRequest` with an http request that has a body, it'll go ahead and set up the multipart upload for you:
  https://github.com/google/apitools/blob/master/apitools/base/py/transfer.py#L580

so i remember there was a reason that you guys didn't just want to use a generated storage client (which would handle these sorts of details for you); what was it?

dhermes:
Thanks for the pointer for sending a body.

As for using  generated client, I don't remember there being a reason, but I'd guess it was Python 3 support and `protorpc`.

dhermes:
I expected

```diff
diff --git a/gcloud/storage/blob.py b/gcloud/storage/blob.py
index ac3f9ce..c7e1336 100644
--- a/gcloud/storage/blob.py
+++ b/gcloud/storage/blob.py
@@ -340,7 +340,9 @@ class Blob(_PropertyMixin):
                                         path=self.bucket.path + '/o')
 
         # Use apitools 'Upload' facility.
-        request = http_wrapper.Request(upload_url, 'POST', headers)
+        body = '' if not self._changes else json.dumps(self._properties)
+        request = http_wrapper.Request(url=upload_url, http_method='POST',
+                                       headers=headers, body=body)
 
         upload.ConfigureRequest(upload_config, request, url_builder)
         query_params = url_builder.query_params
```

to handle it, but am getting

```json
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "badContent",
    "message": "Unsupported content with type: application/octet-stream"
   }
  ],
  "code": 400,
  "message": "Unsupported content with type: application/octet-stream"
 }
}
```

when setting `content_type` in the upload. This seems to be a [problem][1] with the API, not the code. Am I missing something?

[...]

PR #3051, opened 2017-02-22
Add 'Blob.update_storage_class' API method.

### LINKED ISSUE #2991: Storage: add support for Per-Object Storage Class
We need to add support for Per-Object Storage Class (POSC).

Here some extra information to help:

- [POSC info from Cloud Storage documentation](https://cloud.google.com/storage/docs/per-object-storage-class)
- [POSC example with JSON API](https://cloud.google.com/storage/docs/changing-storage-classes)
- [Storage Class info from Cloud Storage documentation](https://cloud.google.com/storage/docs/storage-classes)

On top of that, google-cloud-dotnet has made an implementation which can be used as reference. 

- [Please view the Pull Request to understand the changes](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/766)

If you have any implementation questions, please direct them at jskeet

/cc @jskeet @danoscarmike

tseaver:
Closes #2991.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: disable=too-many-lines
+
 """Create / interact with Google Cloud Storage blobs."""
 
 import base64
@@ -72,6 +74,9 @@ class Blob(_PropertyMixin):
 
     _CHUNK_SIZE_MULTIPLE = 256 * 1024
     """Number (256 KB, in bytes) that must divide the chunk size."""
+
+    _STORAGE_CLASSES = (
+        'STANDARD', 'NEARLINE', 'MULTI_REGIONAL', 'REGIONAL', 'COLDLINE')
 
     def __init__(self, name, bucket, chunk_size=None, encryption_key=None):
         super(Blob, self).__init__(name=name)
@@ -851,6 +856,33 @@ class Blob(_PropertyMixin):
             return None, rewritten, size
 
         return api_response['rewriteToken'], rewritten, size
+
+    def update_storage_class(self, new_class, client=None):
+        """Update blob's storage class via a rewrite-in-place.
+
+        See:
+        https://cloud.google.com/storage/docs/per-object-storage-class
+
+        :type new_class: str
+        :param new_class: new storage class for the object
+
+        :type client: :class:`~google.cloud.storage.client.Client`
+        :param client: Optional. The client to use.  If not passed, falls back
+                       to the ``client`` stored on the blob's bucket.
+        """
+        if new_class not in self._STORAGE_CLASSES:
+            raise ValueError("Invalid storage class: %s" % (new_class,))
+
+        client = self._require_client(client)
+        headers = _get_encryption_headers(self._encryption_key)
+        headers.update(_get_encryption_headers(
+            self._encryption_key, source=True))
+
+        api_response = client._connection.api_request(
+            method='POST', path=self.path + '/rewriteTo' + self.path,
+            data={'storageClass': new_class}, headers=headers,
+            _target_object=self)
+        self._set_properties(api_response['resource'])
 
     cache_control = _scalar_property('cacheControl')
     """HTTP 'Cache-Control' header for this object.
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -76,7 +76,27 @@ class Blob(_PropertyMixin):
     """Number (256 KB, in bytes) that must divide the chunk size."""
 
     _STORAGE_CLASSES = (
-        'STANDARD', 'NEARLINE', 'MULTI_REGIONAL', 'REGIONAL', 'COLDLINE')
+        'NEARLINE',
+        'MULTI_REGIONAL',
+        'REGIONAL',
+        'COLDLINE',
+        'STANDARD',  # alias for MULTI_REGIONAL/REGIONAL, based on location
+    )
+    """Allowed values for :attr:`storage_class`.
+
+    See:
+    https://cloud.google.com/storage/docs/json_api/v1/objects#storageClass
+    https://cloud.google.com/storage/docs/per-object-storage-class
+
+    .. note::
+       This list does not include 'DURABLE_REDUCED_AVAILABILITY', which
+       is only documented for buckets (and deprectated.
+
+    .. note::
+       The documentation does *not* mention 'STANDARD', but it is the value
+       assigned by the back-end for objects created in buckets with 'STANDARD'
+       set as their 'storage_class'.
+    """
 
     def __init__(self, name, bucket, chunk_size=None, encryption_key=None):
         super(Blob, self).__init__(name=name)
--- a/storage/unit_tests/test_blob.py
+++ b/storage/unit_tests/test_blob.py
@@ -1453,7 +1453,7 @@ class Test_Blob(unittest.TestCase):
         with self.assertRaises(ValueError):
             blob.update_storage_class(u'BOGUS')
 
-    def test_update_storage_class_setter_valid(self):
+    def test_update_storage_class_wo_encryption_key(self):
         from six.moves.http_client import OK
         BLOB_NAME = 'blob-name'
         STORAGE_CLASS = u'NEARLINE'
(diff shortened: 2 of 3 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: disable=too-many-lines
```
dhermes: :sob:

storage/google/cloud/storage/blob.py
```diff
@@ -73,6 +75,9 @@ class Blob(_PropertyMixin):
     _CHUNK_SIZE_MULTIPLE = 256 * 1024
     """Number (256 KB, in bytes) that must divide the chunk size."""
 
+    _STORAGE_CLASSES = (
+        'STANDARD', 'NEARLINE', 'MULTI_REGIONAL', 'REGIONAL', 'COLDLINE')
```
dhermes: Don't we already have this [**somewhere**][1]?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/2c608f95d7d5db2cbfdce77770f523a489b513f4/storage/google/cloud/storage/bucket.py#L93-L94

storage/google/cloud/storage/blob.py
```diff
@@ -852,6 +857,31 @@ def rewrite(self, source, token=None, client=None):
 
         return api_response['rewriteToken'], rewritten, size
 
+    def update_storage_class(self, new_class, client=None):
```
dhermes: Shouldn't the constructor also get to set this?

storage/google/cloud/storage/blob.py
```diff
@@ -852,6 +857,31 @@ def rewrite(self, source, token=None, client=None):
...
+        """Rewrite source blob into this one.
```
dhermes: This isn't a very helpful one-liner

storage/google/cloud/storage/blob.py
```diff
@@ -852,6 +857,31 @@ def rewrite(self, source, token=None, client=None):
...
+
+        :type new_class: str
+        :param new_class: new storage class for the object
+
+        :type client: :class:`~google.cloud.storage.client.Client` or
+                      ``NoneType``
```
dhermes: The `or NoneType` isn't really needed since "Optional" conveys the same.

storage/google/cloud/storage/blob.py
daspecster: Should we make an issue/reminder to break this file up down the road?

I can make the issue if you think that's an ok plan.

storage/google/cloud/storage/blob.py
```diff
@@ -852,6 +857,31 @@ def rewrite(self, source, token=None, client=None):
...
+        if new_class not in self._STORAGE_CLASSES:
+            raise ValueError("Invalid storage class: %s" % (new_class,))
+
+        client = self._require_client(client)
+        headers = _get_encryption_headers(self._encryption_key)
+        headers.update(_get_encryption_headers(
+            self._encryption_key, source=True))
+
+        api_response = client._connection.api_request(
+            method='POST', path=self.path + '/rewriteTo' + self.path,
```
dhermes: Don't we need to also implement `rewrite` without the fancy storage class helper?

storage/google/cloud/storage/blob.py
```diff
@@ -852,6 +857,31 @@ def rewrite(self, source, token=None, client=None):
...
+            data={'storageClass': new_class}, headers=headers,
```
dhermes: No other data to send?

storage/unit_tests/test_blob.py
```diff
@@ -1446,6 +1446,48 @@ def test_rewrite_same_name_no_key_new_key_w_token(self):
...
+        self.assertEqual(kw[0]['data'], SENT)
+
+        headers = {
+            key.title(): str(value) for key, value in kw[0]['headers'].items()}
+        self.assertNotIn('X-Goog-Copy-Source-Encryption-Algorithm', headers)
+        self.assertNotIn('X-Goog-Copy-Source-Encryption-Key', headers)
+        self.assertNotIn('X-Goog-Copy-Source-Encryption-Key-Sha256', headers)
+        self.assertNotIn('X-Goog-Encryption-Algorithm', headers)
+        self.assertNotIn('X-Goog-Encryption-Key', headers)
+        self.assertNotIn('X-Goog-Encryption-Key-Sha256', headers)
```
dhermes: 1. Why are these 6 keys hardcoded here?
1. All we know is what is **not** in the headers. What **is** in the headers?

storage/google/cloud/storage/blob.py
tseaver: The allowed set for blobs is smaller than for buckets.

storage/google/cloud/storage/blob.py
tseaver: Doesn't matter until we fix the "send metadata with upload" issues.

storage/google/cloud/storage/blob.py
tseaver: There is a separate `rewrite()` method already (added initially to support rotating encryption keys).

storage/google/cloud/storage/blob.py
tseaver: Nope. See the "REST APIs" section of https://cloud.google.com/storage/docs/changing-storage-classes

storage/unit_tests/test_blob.py
tseaver: The blob has no encryption key set, and so we don't send any information to decrypt/re-encrypt it (check the similar tests for `rewrite()`).  The `headers` *might* be empty here, or they might have gotten injected with extra low-level stuff (like the GCCL stuff).

storage/google/cloud/storage/blob.py
tseaver: Yup, it is copy-pasta from the `rewrite` docstring.  Fixed in 087a7fb.

storage/google/cloud/storage/blob.py
tseaver: Fixed in 087a7fb.

storage/google/cloud/storage/blob.py
tseaver: Not specifically:  we hit an arbitrary, self-imposed limit.  Breaking the file up isn't necessarily a good thing.  You might cast the issue as "review those limits and the files where we disable them."

storage/google/cloud/storage/blob.py
dhermes: Ah. Blerg. We definitely need a tracking issue for that (kicking the can down the road is better than deleting it from our memory)

storage/google/cloud/storage/blob.py
dhermes: :+1:

storage/google/cloud/storage/blob.py
dhermes: Ack

storage/unit_tests/test_blob.py
dhermes: Since we run the unit tests always, we can assert what is in the headers and if someone breaks this test by accident, then they can just fix it on the spot? (This is germane to a conversation @lukesneeringer and I were having earlier today.)

lukesneeringer: What is the test trying to establish? My personal feeling is that other, unrelated changes might alter headers, and asserting empty set seems like it is likely to end up breaking the test on unrelated changes.

If the only thing we need to prove is that these headers are absent, then what @tseaver has seems fine to me.

storage/google/cloud/storage/blob.py
tseaver: #754 and #1660 are open.

storage/unit_tests/test_blob.py
tseaver: If it would make it clearer, I could add another test which showed updating the storage class for a blob which *does* have an encryption key:  in that case, those headers will be populated (with the same values for source and destination).

lukesneeringer: That sounds fine. (I am honestly fine with whatever in this case.)

tseaver: f1a03ea adds such a test, along with a comment in the no-encryption-key test.

storage/google/cloud/storage/blob.py
dhermes: I meant an issue to track adding the storage class to the constructor.

storage/google/cloud/storage/blob.py
dhermes: Are we sure that's true? Or has that list just been upgraded under our nose?

dhermes:
Two remaining sticking points:

- A tracking issue for obj. storage class in the Blob constructor
- Making sure we don't repeat ourselves on the list of storage classes in blob and bucket modules

tseaver:
@dhermes 

>  A tracking issue for obj. storage class in the Blob constructor

#3062.

>  Making sure we don't repeat ourselves on the list of storage classes in blob and bucket modules

The two sets aren't identical:  in particular, the DRA class is only documented for Buckets, although it is deprecated, while the 'STANDARD' class isn't documented for POSC at all (although it happens to work).

lukesneeringer:
> The two sets aren't identical: in particular, the DRA class is only documented for Buckets, although it is deprecated, while the 'STANDARD' class isn't documented for POSC at all (although it happens to work).

Is one a strict superset of the other? If so, the file with the larger set could import the smaller one and `.union` on it. (You'd also have to change to sets rather than tuples, which would be fine.)

tseaver:
> Is one a strict superset of the other? 

@lukesneeringer By inspection of the documented values at point-of-last-reading, yes:  however, the docs don't *specify* that:  they just describe the two sets which happen to overlap:

- The [`Bucket` resource spec](https://cloud.google.com/storage/docs/json_api/v1/buckets#storageClass) lists all the values currently laid out in our `Bucket._STORAGE_CLASSES` member.
- The [`Object` resource spec](https://cloud.google.com/storage/docs/json_api/v1/objects#storageClass) doesn't list any values at all (boo, hiss!)
- The [Storage Classes concept doc](https://cloud.google.com/storage/docs/storage-classes) lists only four (multi-regional, regional, nearline, and coldline), and [defines `STANDARD`](https://cloud.google.com/storage/docs/storage-classes#standard) as an alias based on the bucket's location type.  It [documents DRA](https://cloud.google.com/storage/docs/storage-classes#dra) as a deprecated version, pushing use of multi-regional or regional instead.
- The [Per-Object Storage Class concept docs](https://cloud.google.com/storage/docs/per-object-storage-class) *only* allow for the four types: I found by inspection that `STANDARD` is settable, and is the value which a new object shows in its resource if the bucket specified no storage class.

We could perhaps coalesce the two lists if we dropped support for setting the DRA type on buckets, or if we could get the storage back-end folks to document the possible values definitively for both buckets and objects.

lukesneeringer:
Okay, got it. Based on that, I think I am going to say that we stick with @tseaver's approach as implemented for now (while acknowledging that it is a close call).

dhermes:
@lukesneeringer @tseaver We should

- include a comment in code explaining why the duplication
- give feedback to the backend folks that we had to duplicate / the choices don't seem clear from the docs (/cc @jonparrott)

tseaver:
@dhermes c34cd86 adds docstrings to both class attributes, including links to the relevant back-end documentation.

tseaver:
@dhermes, @lukesneeringer Any further issues?

lukesneeringer:
This looks good. For some reason I cannot find the button to change my review status to "approved", so consider this comment to be tacit approval. :-p

Issue #2746, opened 2016-11-17
Port Datastore onto GAPIC


bjwatson:
### What
Port the gRPC code path for Datastore onto GAPIC. The package is available [here](https://pypi.python.org/pypi/gapic-google-cloud-datastore-v1).

### Why
This is not a beta-blocker, since there already is gRPC code for Datastore, but it is preferred.

dhermes:
@bjwatson Two questions:

1. Why doesn't the new package depend on `google-gax>=0.15` (which has the new auth helpers)?
1. Are we in a situation where some of the GAPIC packages depend on `0.13 <= google-gax < 0.14` while others depend on `0.14 <= google-gax < 0.15` (as this one does)?

bjwatson:
@dhermes Only because @geigerj says that he needs to make some small updates to the GAPIC generator for it to be compatible with `google-gax>=0.15`. He might be able to get that done today, and then we can build all packages with the newer `google-gax` dependency.

I thought I would get this in your hands right away, rather than waiting for the necessary fixes to support the newer auth helpers. However, if you think it's a more efficient use of your time to wait, then we can do that.

dhermes:
@bjwatson Sorry I didn't notice this until now, but the [current form][1] of `grpc-google-cloud-datastore-v1` can't run without gRPC. This means we can't get access to the protobufs without forcing `grpc` to be installed.

Right now we support Protobuf-over-HTTP and Protobuf-over-gRPC, so we need access to the protobuf classes in use cases where gRPC isn't available (i.e. App Engine).

The generated protobuf classes currently live in [`datastore._generated`][2], which is how we import them. In order to support gRPC we have `datastore_pb2.py` and `datastore_grpc_pb2.py` in the `_generated` module

/cc @jonparrott 

[1]: https://github.com/googleapis/api-client-staging/blob/f9305a6602771140c954939cf00b8ab09101d0c3/generated/python/grpc-google-cloud-datastore-v1/google/datastore/v1/datastore_pb2.py#L891-L895
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/tree/a10cf56445b51750c4d9bc1f3b5bde2522f92caf/datastore/google/cloud/datastore/_generated

bjwatson:
[...]

PR #3057, opened 2017-02-23
Isolating Connection objects in _http modules (where relevant).

dhermes:
This is a work-in-progress for now, but `vision` and `speech` are done. Will also be doing `logging`, `pubsub` and **maybe** `error_reporting`.

FWIW, this is a result of me working on #2746.

### DIFF
--- a/error_reporting/google/cloud/error_reporting/_gax.py
+++ b/error_reporting/google/cloud/error_reporting/_gax.py
@@ -36,7 +36,7 @@ def make_report_error_api(client):
     :returns: An Error Reporting API instance.
     """
     channel = make_secure_channel(
-        client._connection.credentials,
+        client._credentials,
         DEFAULT_USER_AGENT,
         report_errors_service_client.ReportErrorsServiceClient.SERVICE_ADDRESS)
     gax_client = report_errors_service_client.ReportErrorsServiceClient(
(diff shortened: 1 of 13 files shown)


dhermes:
@daspecster Thanks but you are bit quick on the draw. The work-in-progress bit means more changes will be coming.

daspecster:
Sorry, I thought that meant this was one PR of n-PRs.

dhermes:
Holding off on `logging` and `pubsub` in this PR since it would un-necessarily create 3 connections (since there are 3 services created in either case). What we really care about is `connection.api_request` ([here][1] and [here][2]), which just relies on **CLASS** attributes, the credentials (known on the client) and the inputs. So it can wait until a larger refactor.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/d4d0abcab27b3f1aba9b56db4d643d5692230bd5/logging/google/cloud/logging/_http.py#L65
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/d4d0abcab27b3f1aba9b56db4d643d5692230bd5/pubsub/google/cloud/pubsub/_http.py#L112

dhermes:
@daspecster @lukesneeringer PTAL this is ready for review

daspecster:
This LGTM.

PR #3059, opened 2017-02-23
Send back the complete API responses.

lukesneeringer:
This pull request makes it so that we no longer drop large portions of the API responses from Natural Language.

(It also changes several copyright years.)


daspecster:
@lukesneeringer why the copyright year changes? Those aren't new files this year?

lukesneeringer:
I changed the ones I edited. (Something I should be better at doing generally.)

dhermes:
> I changed the ones I edited. (Something I should be better at doing generally.)

I don't know what a lawyer-cat says but @jonparrott does know (sometimes)

daspecster:
@lukesneeringer I thought the copyright was only for when the file was created but I could be wrong. I'm no copyright lawyer.

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+            language=payload.get('language', None),
+            sentences=[Sentence.from_api_repr(i) for i
+                       in payload.get('sentences', ())],
+            sentiment=Sentiment.from_api_repr(payload['documentSentiment']),
+        )
+
+
+class SyntaxResponse(object):
+    """A representation of a response to an ``analyzeSyntax`` request
+    to the Google Natural Language API.
```
dhermes: Ditto ditto ditto

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @classmethod
+    def from_api_repr(cls, payload):
+        """Return an syntax response from a JSON representation.
+
+        :type payload: dict
+        :param payload: A dictionary representing the response.
+        """
+        return cls(
+            language=payload.get('language', None),
```
dhermes: Drop `None`?

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+            sentences=[Sentence.from_api_repr(i) for i in
```
dhermes: s/`i`/`sentence`

language/google/cloud/language/client.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2016-2017 Google Inc.
```
dhermes: No edits?

daspecster:
@lukesneeringer hyphenating makes a lot more sense. 👍

language/google/cloud/language/document.py
```diff
@@ -27,7 +28,7 @@
 
 Annotations = collections.namedtuple(
     'Annotations',
-    'sentences tokens sentiment entities')
+    ['sentences', 'tokens', 'sentiment', 'entities', 'language'])
```
dhermes: Personal preference (for the list)?

language/google/cloud/language/document.py
```diff
@@ -271,9 +272,10 @@ def annotate_text(self, include_syntax=True, include_entities=True,
         entities = [Entity.from_api_repr(entity)
                     for entity in api_response['entities']]
         annotations = Annotations(
+            entities=entities,
+            language=api_response.get('language', None),
```
dhermes: `None` not needed?

language/google/cloud/language/entity.py
```diff
@@ -84,7 +80,7 @@ class Entity(object):
     def __init__(self, name, entity_type, metadata, salience, mentions):
         self.name = name
         self.entity_type = entity_type
-        self.wikipedia_url = metadata.pop('wikipedia_url', None)
+        self.wikipedia_url = metadata.get('wikipedia_url', None)
```
dhermes: `None` no longer needed. Why not pop?

language/google/cloud/language/sentence.py
```diff
@@ -0,0 +1,67 @@
...
+    def from_api_repr(cls, payload):
+        """Convert a sentence from the JSON API into a :class:`Sentence`.
+
+        :param payload: dict
+        :type payload: The value from the backend.
+
+        :rtype: :class:`Sentence`
+        :returns: The sentence parsed from the API representation.
+        """
+        text_span = payload['text']
```
dhermes: Not worried about a `KeyError`?

language/google/cloud/language/sentiment.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2016-2017 Google Inc.
```
dhermes: No edit?

language/google/cloud/language/sentence.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # The sentence may or may not have a sentiment; only attempt the
+        # typecast if one is present.
+        sentiment = None
+        if payload.get('sentiment', None):
```
dhermes: Recommendation (I think in Google style guide, maybe in Python recs somewhere) is to check the key:

```python
if 'sentiment' in payload:
    ....
```

language/google/cloud/language/syntax.py
```diff
@@ -168,53 +169,3 @@ def from_api_repr(cls, payload):
         lemma = payload['lemma']
         return cls(text_content, text_begin, part_of_speech,
                    edge_index, edge_label, lemma)
-
-
-class Sentence(object):
```
dhermes: The move is just hygiene?

jonparrott:
Copyright is for when the file was first published, not when it was last edited.

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+        """Return an sentiment response from a JSON representation.
+
+        :type payload: dict
+        :param payload: A dictionary representing the response.
+        """
+        return cls(
+            language=payload.get('language', None),
+            sentences=[Sentence.from_api_repr(i) for i
+                       in payload.get('sentences', ())],
+            sentiment=Sentiment.from_api_repr(payload['documentSentiment']),
```
lukesneeringer: I actually got one! :-O

language/google/cloud/language/api_responses.py
lukesneeringer: I believe they only guarantee the return of `language` if you do not specify it.

language/google/cloud/language/document.py
lukesneeringer: Yeah. (I was actually shocked that a space separated string worked.)

language/google/cloud/language/document.py
lukesneeringer: See above.

language/google/cloud/language/entity.py
lukesneeringer: Are we guaranteed to always get a `wikipedia_url`? It seems unlikely.

language/google/cloud/language/sentence.py
lukesneeringer: Those are not the same. If `payload['sentiment']` is set to `False` or `None`, my check will be `False` and `'sentiment' in payload` will be `True`.

Passing a payload of `None` to the `from_api_repr` method raises `AttributeError`, so the difference matters.

I could do `if 'sentiment' in payload and payload['sentiment']`, but seriously? ;-)

language/google/cloud/language/syntax.py
lukesneeringer: Circular import.

language/google/cloud/language/sentence.py
lukesneeringer: This part is moved verbatim.

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @classmethod
+    def from_api_repr(cls, payload):
+        """Return an sentiment response from a JSON representation.
+
+        :type payload: dict
+        :param payload: A dictionary representing the response.
+        """
+        return cls(
+            language=payload.get('language', None),
```
lukesneeringer: I _think_ `language` is not guaranteed to be present.

language/google/cloud/language/api_responses.py
dhermes: I'm just commenting on the behavior of `dict.get`.

language/google/cloud/language/entity.py
dhermes: That's why `pop` falls back to `None`, in case it isn't there. I made it a custom attribute since `wikipedia_url` was the only value empirically observed and since it was there almost all of (all of?) the time.

language/google/cloud/language/sentence.py
dhermes: But we know that `sentiment` will be a float? Why are you checking if it's truth-y (that never occurred to me)?

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+
+from google.cloud.language.entity import Entity
+from google.cloud.language.sentence import Sentence
+from google.cloud.language.sentiment import Sentiment
+from google.cloud.language.syntax import Token
+
+
+class EntityResponse(object):
+    """A representation of a response sent back from the
+    ``analyzeEntites`` request to the Google Natural language API.
```
lukesneeringer: Not altering.

language/google/cloud/language/api_responses.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @classmethod
+    def from_api_repr(cls, payload):
+        """Return an entity response from a JSON representation.
+
+        :type payload: dict
+        :param payload: A dictionary representing the response.
+        """
+        return cls(
+            entities=[Entity.from_api_repr(i) for i in payload['entities']],
```
lukesneeringer: What happened to one-liners spilling over into multiple lines? ;-)

language/google/cloud/language/api_responses.py
lukesneeringer: Oh hey! Today I learned. I thought that `{}.get('foo')` was equivalent to `{}['foo']`, not `{}.get('foo', None)`. I _think_ I prefer the more explicit format though (but I might be biased by not knowing what the shorter form did).

language/google/cloud/language/entity.py
lukesneeringer: The problem is that they have since added `mid` and they may add more items in the future.

language/google/cloud/language/sentence.py
lukesneeringer: Oooh I see. So I really need `if payload.get('sentiment', None) is not None:`.

lukesneeringer:
I also failed at updating the system tests -- working on that now.

language/google/cloud/language/api_responses.py
tseaver: Huh?  `None` is the default value of the default parameter to `get`, so it never raises a `KeyError`:

```python
>>> {}.get('Nonesuch') is None
True
>>> {}['Nonesuch']
KeyError: 'Nonesuch'
```

dhermes: I am somewhat strongly for the simpler version: `dict.get('foo')`

language/google/cloud/language/entity.py
dhermes: Right. But if that's the case, we should just get rid of `self.wikipedia_url`. The `pop` was there so it wasn't duplicated in the (always empty) `self.metadata` and in the `self.wikipedia_url`.

language/google/cloud/language/api_responses.py
lukesneeringer: Okay, I will do the implicit version.

language/google/cloud/language/api_responses.py
dhermes: The "no docstring in unit test" argument applies here as well: until now, we have totally avoided one-line docstring summaries that spill onto multiple lines. It seems a pity to introduce that phenomenon for this docstring, which can be easily rewritten without losing meaning.

language/google/cloud/language/entity.py
lukesneeringer: I suppose this breaking change is breaking enough that this will not hurt. :-)

language/unit_tests/test_sentence.py
```diff
@@ -0,0 +1,74 @@
...
+        payload = {
+            'text': {
+                'content': content,
+                'beginOffset': begin,
+            },
+        }
+        sentence = klass.from_api_repr(payload)
+        self.assertEqual(sentence.content, content)
+        self.assertEqual(sentence.begin, begin)
+        self.assertEqual(sentence.sentiment, None)
```
dhermes: NBD but it's "better" to use `assertIsNone` (here and in some other places)

language/google/cloud/language/entity.py
```diff
@@ -84,7 +80,7 @@ class Entity(object):
     def __init__(self, name, entity_type, metadata, salience, mentions):
         self.name = name
         self.entity_type = entity_type
-        self.wikipedia_url = metadata.pop('wikipedia_url', None)
+        self.wikipedia_url = metadata.get('wikipedia_url')
```
dhermes: Since you aren't `pop`-ing, ISTM that `self.wikipedia_url` should just be removed

language/google/cloud/language/api_responses.py
dhermes: Discussion Bump

language/google/cloud/language/document.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2016-2017 Google Inc.
```
dhermes: ISTM @jonparrott's comment indicates this change is not needed / not OK

language/google/cloud/language/entity.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2016-2017 Google Inc.
```
dhermes: Ditto on no-mo-license-change

language/google/cloud/language/sentence.py
```diff
@@ -0,0 +1,68 @@
...
+
+        :rtype: :class:`Sentence`
+        :returns: The sentence parsed from the API representation.
+        """
+        text_span = payload['text']
+
+        # The sentence may or may not have a sentiment; only attempt the
+        # typecast if one is present.
+        sentiment = None
+        if payload.get('sentiment') is not None:
```
dhermes: So this is still the recommendation to use `'sentiment' in payload`.

lukesneeringer: That is not equivalent, and the difference matters. `payload['sentiment']` can be None.

You wrote similar code in `document.py` to deal with the same issue:

```python
        sentiment_info = api_response.get('documentSentiment')
        if sentiment_info is None:
            sentiment = None
        else:
            sentiment = Sentiment.from_api_repr(sentiment_info)
```

lukesneeringer:
@googlebot Why is that a problem, exactly? :confused: Just saying...

PR #3063, opened 2017-02-23
Drop usage of overly broad mock.Mock() in tests.

dhermes:
@lukesneeringer @tseaver any ideas on an easy "lint"-like check that will scream if people create a `Mock()` without a `spec`?

SIDEBAR: The error reporting unit tests are in a **very bad state**. This change opened my eyes to that fact, but I fear there is more hiding under the surface.

### DIFF
--- a/error_reporting/unit_tests/test__gax.py
+++ b/error_reporting/unit_tests/test__gax.py
@@ -59,24 +59,34 @@ class Test_ErrorReportingGaxApi(unittest.TestCase):
 
     PROJECT = 'PROJECT'
 
-    def _call_fut(self, gax_api, project):
+    def _make_one(self, gax_api, project):
         from google.cloud.error_reporting._gax import _ErrorReportingGaxApi
 
         return _ErrorReportingGaxApi(gax_api, project)
 
     def test_constructor(self):
-        gax_api = mock.Mock()
-        gax_client_wrapper = self._call_fut(gax_api, self.PROJECT)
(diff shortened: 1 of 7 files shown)


error_reporting/google/cloud/error_reporting/_gax.py
```diff
@@ -36,7 +36,7 @@ def make_report_error_api(client):
     :returns: An Error Reporting API instance.
     """
     channel = make_secure_channel(
-        client._connection.credentials,
+        client._credentials,
```
dhermes: This part also in #3057

daspecster:
Should mocks like `mock.Mock(return_value=True)` also pass in `spec`?

dhermes:
@daspecster All mocks should state their spec. Maybe we should even extend this to `mock.patch`?

waprin:
Apologies if you don't think unit tests are up to snuff, thank you for improving. 

I think a system test will inspire the most confidence in the functionality. When I first contributed the PR the API wasn't full fleshed out, but now that it is I can write one.

dhermes:
Travis seems to be having issues, merging with CircleCI already green.

screenshot from 2017-03-01 15-56-04

dhermes:
Ahhh sorry I thought @daspecster had LGTM-ed! Eeek.

Worth reverting or someone just take a look ex-post-merge-o?

lukesneeringer:
It's fine. :-)

PR #3064, opened 2017-02-23
Making datastore Connection.commit() return low-level protobuf.

dhermes:
Towards #2746. This approach is to slowly transition from our current approach to use the GAPIC generated surface.

It is unfortunately tangled quite a bit (partly because we may have too much mocked in the tests).

### DIFF
--- a/datastore/google/cloud/datastore/batch.py
+++ b/datastore/google/cloud/datastore/batch.py
@@ -238,8 +238,9 @@ class Batch(object):
         This is called by :meth:`commit`.
         """
         # NOTE: ``self._commit_request`` will be modified.
-        _, updated_keys = self._client._connection.commit(
+        commit_response_pb = self._client._connection.commit(
             self.project, self._commit_request, self._id)
+        _, updated_keys = _parse_commit_response(commit_response_pb)
         # If the back-end returns without error, we are guaranteed that
         # :meth:`Connection.commit` will return keys that match (length and
         # order) directly ``_partial_key_entities``.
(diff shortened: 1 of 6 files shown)


tseaver:
@dhermes so this is preparatory to adding `google.cloud.datastore._gax`, with its own implemenation of the API via the GAPIC client?

lukesneeringer:
@dhermes Is your goal to get this in with the umbrella releaase, or after?

datastore/google/cloud/datastore/batch.py
```diff
@@ -311,3 +312,21 @@ def _assign_entity_to_pb(entity_pb, entity):
...
+
+    :type commit_response_pb: :class:`.datastore_pb2.CommitResponse`
+    :param commit_response_pb: The protobuf response from a commit request.
+
+    :rtype: tuple
+    :returns: The pair of the number of index updates and a list of
+              :class:`.entity_pb2.Key` for each incomplete key
+              that was completed in the commit.
+    """
+    mut_results = commit_response_pb.mutation_results
```
daspecster: Could just be me, but `mut_results` feels like `MUT` from the unite tests.

datastore/unit_tests/test_batch.py
```diff
@@ -369,30 +369,62 @@ def begin(self):
...
+                    entity_pb2.Key.PathElement(
+                        kind='Bar',
+                        name='baz',
+                    ),
+                ],
+            ),
+        ]
+        response = datastore_pb2.CommitResponse(
+            mutation_results=[
+                datastore_pb2.MutationResult(key=key) for key in keys
```
daspecster: I'm surprised key/keys aren't reserved words, but it doesn't look like they are.
https://docs.python.org/2.5/ref/keywords.html

lukesneeringer: They are not reserved words, and `key` is an extremely common variable name.

datastore/google/cloud/datastore/batch.py
lukesneeringer: It is a four line function. It is fine.

lukesneeringer:
Approved, but holding off on merge until the umbrella release question is answered.

datastore/google/cloud/datastore/batch.py
dhermes: Also the `mutation_results` on the RHS of the line should make it clear

dhermes:
@lukesneeringer A merge is just fine.

@tseaver `google.cloud.datastore._gax` already exists, but yes this is part of the transition to switch to using the GAPIC layer in `google.cloud.datastore._gax`.

dhermes:
Passing on Travis, not done yet on CircleCI (since it doesn't **yet** do a restricted set of tests). Merging before CircleCI is done.

PR #3066, opened 2017-02-24
Making datastore Connection.run_query() return low-level protobuf.

dhermes:
Towards #2746 (as is #3064). This approach is to slowly transition from our current approach to use the GAPIC generated surface.

These unit tests weren't so bad to update. I did "editorialize" as I went, downgrading constants to just variables, renamining for PEP8, etc.

### DIFF
--- a/datastore/google/cloud/datastore/query.py
+++ b/datastore/google/cloud/datastore/query.py
@@ -441,43 +441,33 @@ class Iterator(BaseIterator):
 
         return pb
 
-    def _process_query_results(self, entity_pbs, cursor_as_bytes,
-                               more_results_enum, skipped_results):
+    def _process_query_results(self, response_pb):
         """Process the response from a datastore query.
 
-        :type entity_pbs: iterable
-        :param entity_pbs: The entities returned in the current page.
-
-        :type cursor_as_bytes: bytes
-        :param cursor_as_bytes: The end cursor of the query.
-
-        :type more_results_enum:
-            :class:`.query_pb2.QueryResultBatch.MoreResultsType`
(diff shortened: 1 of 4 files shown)


lukesneeringer:
@dhermes Is your goal to get this into the umbrella release, or put it off until after? (I am okay with either as long as it does not _block_ it.)

datastore/google/cloud/datastore/query.py
```diff
@@ -441,43 +441,33 @@ def _build_protobuf(self):
...
 
         :rtype: iterable
         :returns: The next page of entity results.
         :raises ValueError: If ``more_results`` is an unexpected value.
         """
-        self._skipped_results = skipped_results
+        self._skipped_results = response_pb.batch.skipped_results
 
-        if cursor_as_bytes == b'':  # Empty-value for bytes.
+        if response_pb.batch.end_cursor == b'':  # Empty-value for bytes.
```
daspecster: Clearly this is empty bytes, but maybe a better comment could explain why empty bytes are needed?

lukesneeringer:
Approved, but holding off on merge until the umbrella release question is answered.

dhermes:
@lukesneeringer This isn't release blocking. Only the guts have changed, so no user-facing implications.

datastore/google/cloud/datastore/query.py
dhermes: @daspecster 

1. The comment is intended to explain why the check equal to `b''` rather than using `HasField`. This is because for "scalar types", `HasField` doesn't work in pb objects (it throws an exception). Instead we equate being set to the "empty-value" to be unset.
1. The comment was pre-existing

PR #3074, opened 2017-02-26
for streaming sample should have "stream=" param

yogeshg:
Sample class can take in content, source_uri or stream, should be stream here.
```
    :type content: bytes
    :param content: (Optional) Bytes containing audio data.

    :type stream: file
    :param stream: (Optional) File like object to stream.
```
the type is deduced using:
```sources = [content is not None, source_uri is not None, stream is not None]```

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -207,7 +207,7 @@ If ``interim_results`` is set to :data:`True`, interim results
     >>> from google.cloud import speech
     >>> client = speech.Client()
     >>> with open('./hello.wav', 'rb') as stream:
-    ...     sample = client.sample(content=stream,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by yogeshg** [comment]
I signed it!
```
I signed it!
```

yogeshg:
I signed it!

dhermes:
LGTM. @daspecster LGTY?

daspecster:
@dhermes yup!
Thanks @yogeshg!

yogeshg:
@dhermes, @daspecster  my pleasure!
I also added an example for this use case in the samples repo. I was not sure how to write the test script for it though. You can help/advise me here:
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/826

PR #3084, opened 2017-02-28
Add test_result.py for speech, update GAPICSpeechAPI docstring and test formatting.

daspecster:


speech/unit_tests/test__gax.py
```diff
@@ -50,39 +50,24 @@ def test_constructor(self, mocked_stub, mocked_cls, mocked_channel):
...
+        self.assertIs(speech_api._gapic_api, mocked_cls.return_value)
 
         mocked_stub.assert_called_once_with(
-            mock_cnxn.credentials,
-            DEFAULT_USER_AGENT,
-            operations_grpc.OperationsStub,
-            OPERATIONS_API_HOST,
-        )
+            mock_cnxn.credentials, DEFAULT_USER_AGENT,
+            operations_grpc.OperationsStub, OPERATIONS_API_HOST)
```
dhermes: You really went wild with these style changes? One might argue in this case you've made the code **less readable**.

No need to change, just a remark.

daspecster: Yeah, this is kind of a toss up for me. I wanted the test formatting to conform to the rules of the rest of the library and pep8. That's the goal anyway.

dhermes: PEP8 doesn't say anything about the previous state / current state of these continuations that were changed. And neither does are tooling. So it's just our personal opinions without a tool to enforce it.

Issue #3083, opened 2017-02-28
RetryErrors obscure actual errors.


lukesneeringer:
@daspecster Please take a look at [this issue][1].

Instead of raising a retry error saying that we got an error that was not classified as transient, can we just re-raise the original error? What is there now is a nightmare for debugging.

  [1]: https://github.com/GoogleCloudPlatform/python-docs-samples/issues/833

daspecster:
Interesting, I'll see if I can track it down.

lukesneeringer:
This never happened, but Vision has been moved to autogen, so this issue is more or less obselete.

mr-ubik:
I am having the exact issue but with the google-cloud-language package. I am streaming tweets (working with Apache Beam) to the Language Sentiment Analysis API but some of them seem to break it.
```python
class SentimentAnalyzerDoFn(beam.DoFn):
    def process(self, element):
        """Takes a parsed tweet as input and add to it its sentiment score and values"""
        from google.cloud import language
        client = language.LanguageServiceClient()
        document = language.types.Document(
            content=element['tweet'].encode('utf-8'),
            type='PLAIN_TEXT')
        response = client.analyze_sentiment(
            document=document,
            encoding_type='UTF8')
        sentiment = response.document_sentiment
        element['sentiment_score'] = sentiment.score
        element['sentiment_magnitude'] = sentiment.magnitude
        yield element
```

The error I am getting is this: 
```python
RetryError: Exception occurred in retry method that was not classified as transient [while running 'SentimentAnalasys']
```
Is there a way to investigate it?

ensonic:
@lukesneeringer this is not specific to the API, it is a general problem

PR #3087, opened 2017-03-01
Expose RetryError exceptions.

daspecster:
Closes: #3083

### DIFF
--- a/vision/google/cloud/vision/_gax.py
+++ b/vision/google/cloud/vision/_gax.py
@@ -14,6 +14,8 @@
 
 """GAX Client for interacting with the Google Cloud Vision API."""
 
+from google.gax.errors import RetryError
+
 from google.cloud.gapic.vision.v1 import image_annotator_client
 from google.cloud.proto.vision.v1 import image_annotator_pb2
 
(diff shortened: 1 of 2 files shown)


vision/google/cloud/vision/_gax.py
```diff
@@ -66,7 +68,12 @@ def annotate(self, images=None, requests_pb=None):
             requests = requests_pb
 
         annotator_client = self._annotator_client
-        responses = annotator_client.batch_annotate_images(requests).responses
+        try:
+            api_result = annotator_client.batch_annotate_images(requests)
+            responses = api_result.responses
+        except RetryError as exception:
```
lukesneeringer: I guess what I had in mind was not necessarily catching this one instance of a `RetryError`. I either want to do that _all_ the time, or I want to clearly document to people what they should do if they get one. As a user, I am not sure how I would have been expected to know that `.cause.exception()` gets me the original exception.

daspecster: So the `RetryError` is pretty [specific](https://github.com/googleapis/gax-python/blob/b938df8b343d793f524b22d4b0cbe86541ba42b0/google/gax/errors.py#L77), and I'm not 100% sure that this will work in a general sense.

Let me know if you have a strategy that would work.

daspecster: If you notice how I access it on the line below, it's pretty custom to the `GaxError` class. `exception.cause.exception()`.

lukesneeringer: Ah, okay, I see the problem.

I am opening a [GAX PR](https://github.com/googleapis/gax-python/pull/168) to fix this the right way, which will render this unnecessary.

daspecster: Ah, good idea. Not sure why I didn't think of that.

lukesneeringer:
Closing this PR. The GAX PR is not merged yet, but it is far enough along that it will certainly end up being the route we take.

PR #3089, opened 2017-03-03
Sending x-goog-api-client header in Bigtable.

dhermes:
We missed this because it does not use GAPIC (the most recent generated Bigtable GAPIC surface is 5+ months old).

----

**NOTE**: This will fail all CI since `google-gax==0.15.7` isn't out yet, but I wanted to get reviewers a chance to see it before `0.15.7` gets released tomorrow. (I have personally run the tests locally by installed `google-gax` from `HEAD`.)

### DIFF
--- a/bigtable/google/cloud/bigtable/__init__.py
+++ b/bigtable/google/cloud/bigtable/__init__.py
@@ -15,4 +15,7 @@
 """Google Cloud Bigtable API package."""
 
 
+from pkg_resources import get_distribution
+__version__ = get_distribution('google-cloud-bigtable').version
+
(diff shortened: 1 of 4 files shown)


dhermes:
@lukesneeringer Builds are green! Thanks for releasing `0.15.7`. PTAL, as the gRPC layer for datastore needs the same treatment that this PR gives to Bigtable. (As an easy-to-implement stopgap while #2746 is in progress.)

dhermes:
@lukesneeringer Should I cut `bigtable-0.23.1`?

lukesneeringer:
Yes, please.

PR #3091, opened 2017-03-03
Making datastore Connection.begin_transaction() return low-level protobuf.

dhermes:
Towards #2746 (as were #3064 and #3066). This approach is to slowly transition from our current approach to use the GAPIC generated surface.

### DIFF
--- a/datastore/unit_tests/test_transaction.py
+++ b/datastore/unit_tests/test_transaction.py
@@ -206,9 +206,12 @@ class _Connection(object):
             mutation_results=mutation_results)
 
     def begin_transaction(self, project):
+        import mock
+
         self._begun = project
         if self._side_effect is None:
-            return self._xact_id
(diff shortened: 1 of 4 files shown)


datastore/unit_tests/test_transaction.py
```diff
@@ -206,9 +206,12 @@ def __init__(self, xact_id=123, keys=()):
             mutation_results=mutation_results)
 
     def begin_transaction(self, project):
+        import mock
```
daspecster: This is one of the few imports that we have been putting at the top of the module because it ends up getting used a lot. I'm not sure if it matters in this case, but someone adding a test using a mock might add the import at the top of the module and then it's imported twice anyway.

dhermes: If that occurred then Pylint would give a `redefined-outer-name` error.

PR #3092, opened 2017-03-03
Making datastore Connection.rollback() return low-level protobuf.

dhermes:
Towards #2746 (as were #3064 and #3066 and #3091). This approach is to slowly transition from our current approach to use the GAPIC generated surface.

### DIFF
--- a/datastore/google/cloud/datastore/_http.py
+++ b/datastore/google/cloud/datastore/_http.py
@@ -448,11 +448,14 @@ class Connection(connection_module.Connection):
         :type transaction_id: str
         :param transaction_id: The transaction ID returned from
                                :meth:`begin_transaction`.
+
+        :rtype: :class:`.datastore_pb2.RollbackResponse`
+        :returns: The returned protobuf response object.
         """
         request = _datastore_pb2.RollbackRequest()
         request.transaction = transaction_id
-        # Nothing to do with this response, so just execute the method.
-        self._datastore_api.rollback(project, request)
+        # Response is empty (i.e. no fields) but we return it anyway.
+        return self._datastore_api.rollback(project, request)
 
     def allocate_ids(self, project, key_pbs):
         """Obtain backend-generated IDs for a set of keys.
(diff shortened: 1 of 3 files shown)


datastore/google/cloud/datastore/_http.py
```diff
@@ -448,11 +448,14 @@ def rollback(self, project, transaction_id):
...
                                :meth:`begin_transaction`.
+
+        :rtype: :class:`.datastore_pb2.RollbackResponse`
+        :returns: The returned protobuf response object.
         """
         request = _datastore_pb2.RollbackRequest()
         request.transaction = transaction_id
-        # Nothing to do with this response, so just execute the method.
-        self._datastore_api.rollback(project, request)
+        # Response is empty (i.e. no fields) but we return it anyway.
```
daspecster: Kind of a strange comment. "we return it anyway", leaves me asking why? Why not return `None` or something.

lukesneeringer: Because the response still has the expected structure, I would assume.

daspecster: Yeah, the other comment alluded to that. I think it could be useful to carry that into the new comment.

dhermes: Because the point of this PR is to honor the protobuf response. 

Also, it's not "nothing", just a [message][1] with no fields. Not sure why the datastore folks defined their own version of [`Empty`][2].

[1]: https://github.com/googleapis/googleapis/blob/8c5b4c491603e0b23c7d013941b1354d0b849caf/google/datastore/v1/datastore.proto#L158
[2]: https://github.com/google/protobuf/blob/008ff033ef3308512c1da2b7cdadf50a6296251a/src/google/protobuf/empty.proto#L52

dhermes: I **can** update the comment, but I'd guess it'll be deleted anyway when #2746 is all done, so it's a little pointless.

dhermes:
I'm going to go ahead and merge.

datastore/google/cloud/datastore/_http.py
daspecster: @dhermes sure, your discretion. Just something that popped out at me in reading it.

PR #3094, opened 2017-03-03
Making datastore Connection.lookup() return low-level protobuf.

dhermes:
Towards #2746 (as were #3064, #3066, #3091, #3092 and #3093). This approach is to slowly transition from our current approach to use the GAPIC generated surface. This is the sixth and final API method to be converted in this fashion.

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -116,28 +116,29 @@ def _extended_lookup(connection, project, key_pbs,
     while loop_num < _MAX_LOOPS:  # loop against possible deferred.
         loop_num += 1
 
-        results_found, missing_found, deferred_found = connection.lookup(
+        lookup_response = connection.lookup(
             project=project,
             key_pbs=key_pbs,
             eventual=eventual,
             transaction_id=transaction_id,
         )
 
-        results.extend(results_found)
+        # Accumulate the new results.
+        results.extend(result.entity for result in lookup_response.found)
 
         if missing is not None:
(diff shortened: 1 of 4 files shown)


datastore/google/cloud/datastore/client.py
```diff
@@ -116,28 +116,29 @@ def _extended_lookup(connection, project, key_pbs,
...
             transaction_id=transaction_id,
         )
 
-        results.extend(results_found)
+        # Accumulate the new results.
+        results.extend(result.entity for result in lookup_response.found)
 
         if missing is not None:
-            missing.extend(missing_found)
+            missing.extend(result.entity for result in lookup_response.missing)
```
daspecster: Is there any chance that `lookup_response.missing` could be undefined?

daspecster: Nevermind, I see it's in the [proto](https://github.com/googleapis/api-client-staging/blob/4ab660906bca3bae2620e1814c58b4368781a735/generated/python/proto-google-cloud-datastore-v1/google/cloud/proto/datastore/v1/datastore_pb2.py#L130) with a default of `[]`.

dhermes: Yeah, source diving in the `pb2.py` generated files is seldom (never?) a good idea. The underlying proto [definition][1] would have given you the same answer. Proto definitions should always be the source of truth.

[1]: https://github.com/googleapis/googleapis/blob/1068554917fcda5facfee95c19e2aa879665316b/google/datastore/v1/datastore.proto#L93

PR #3096, opened 2017-03-04
Sending x-goog-api-client header for Datastore gRPC implementation.

dhermes:


dhermes:
@lukesneeringer Does this warrant a release?

lukesneeringer:
Yes, but it can wait until Monday. :-)

PR #3101, opened 2017-03-06
Remove _BrokenResultFuture

### LINKED ISSUE #3020: Remove '_BrokenOperationFuture' workaround
`google.cloud.spanner.databse._BrokenOperationFuture` is a subclass of `google.gax._OperationFuture` which attempts to work around the fact that the spanner back-end sometimes returns metadata instead of the expected empty response, which causes the normal future to blow up.

Once the back-end is updated to fix this, remove `_BrokenOperationFuture`.

lukesneeringer:
Fixes #3020.

Note: I want to make sure the Spanner system tests run, and it is hard to run them locally. System tests' passing is a prerequisite for merging this.

### DIFF
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -51,22 +51,6 @@ register_type(admin_v1_pb2.CreateDatabaseMetadata)
 register_type(admin_v1_pb2.UpdateDatabaseDdlMetadata)
 
 
-class _BrokenResultFuture(_OperationFuture):
-    """An _OperationFuture subclass that is permissive about type mismatches
-    in results, and simply returns an empty-ish object if they happen.
-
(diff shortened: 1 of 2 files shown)


daspecster:
@lukesneeringer I don't think the CI runs the systems tests until after it's merged to master.

@tseaver what all is needed to run the Spanner system tests now? I have an instance and db, but I don't have the schema setup. I'm missing something simple I'm sure.

lukesneeringer:
@daspecster It needs an environment variable, if I remember correctly. (An unfortunate side effect of the gunk that the Spanner folks had us to do avoid production charges on tests.)

dhermes:
@lukesneeringer Changes LGTM. I can confirm the system tests **DO NOT** run on PRs.

PR #3102, opened 2017-03-06
Provide correct default encoding.

lukesneeringer:
It turns out that Python actually does not encode strings in UTF-8 by default, but actually in UTF-16 or UTF-32, depending on the system. Since Unicode is a variable length format, this distinction does not matter unless you are using characters outside of UTF-8's range.

However, we want our character offsets that we send back from the language API to be correct, so we should provide the correct default encoding request.

### DIFF
--- a/system_tests/language.py
+++ b/system_tests/language.py
@@ -76,7 +76,7 @@ class TestLanguage(unittest.TestCase):
         self.assertGreater(entity1.salience, 0.0)
         # Other mentions may occur, e.g. "painter".
         self.assertIn(entity1.name, entity1.mentions)
-        self.assertEqual(entity1.wikipedia_url,
+        self.assertEqual(entity1.metadata['wikipedia_url'],
                          'http://en.wikipedia.org/wiki/Caravaggio')
         self.assertIsInstance(entity1.metadata, dict)
         # Verify entity 2.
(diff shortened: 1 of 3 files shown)


system_tests/language.py
```diff
@@ -76,15 +76,15 @@ def _check_analyze_entities_result(self, entities):
         self.assertGreater(entity1.salience, 0.0)
         # Other mentions may occur, e.g. "painter".
         self.assertIn(entity1.name, entity1.mentions)
-        self.assertEqual(entity1.wikipedia_url,
+        self.assertEqual(entity1.metadata['wikipedia_url'],
```
dhermes: Sneaking in an orthogonal change?

lukesneeringer: No actually, that change is already in, but the system tests did not reflect it (and fail on master).

lukesneeringer:
> Changes look good though your system test change is unreleated to the stated PR subject

Yeah, system tests fail on master currently. I could have opened a separate PR, merged it, and let this one fail until it was merged. I figured it was easier to just fix it and be done.

system_tests/language.py
dhermes: I mean that the fix is the orthogonal change

lukesneeringer: Fair. Do you want me to split it out?

dhermes: No it's fine. Just slap on the wrist

lukesneeringer: Slap accepted. Can I merge? :-p

dhermes: Sure thing

PR #3104, opened 2017-03-06
Vision Beta

lukesneeringer:
This PR officially makes `google-cloud-vision` a beta library.

@daspecster What else (besides a release, of course) is needed to accomplish this? Do the docs proclaim Vision as alpha right now? If so, we need to update that. (Also, we will need a manual docs trigger once this is pushed out.)

### DIFF
--- a/vision/setup.py
+++ b/vision/setup.py
@@ -57,7 +57,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-vision',
-    version='0.90.0',
+    version='0.23.3',
     description='Python Client for Google Cloud Vision',
     long_description=README,
     namespace_packages=[
--- a/vision/setup.py
+++ b/vision/setup.py
@@ -35,7 +35,7 @@ SETUP_BASE = {
     'include_package_data': True,
     'zip_safe': False,
     'classifiers': [
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Operating System :: OS Independent',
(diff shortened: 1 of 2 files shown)


daspecster:
In [this](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2753), there's mention of changing `Landmark` to `FacialFeature`. I think that might be good to do before beta. WDYT?

vision/setup.py
```diff
@@ -52,12 +52,12 @@
...
     'enum34',
     'google-cloud-core >= 0.23.1, < 0.24dev',
-    'gapic-google-cloud-vision-v1 >= 0.15.2, < 0.16dev',
+    'gapic-google-cloud-vision-v1 >= 0.90.3, < 0.91dev',
 ]
 
 setup(
     name='google-cloud-vision',
-    version='0.23.2',
+    version='0.23.3',
```
dhermes: Don't you want to bump to `0.24.0`?

lukesneeringer: That's a good question. Let's discuss at the end of the standup.

lukesneeringer:
@dhermes I just remembered I forgot to bring this up at the standup.

On the GAPIC side, we bump versions to 0.90 when they hit beta. However:

  (1) We did not do that for other libraries here, and
  (2) That would be inconsistent with how we track everything to core 0.23.x

The beta designation here entails an expectation change, but not so much an actually significant code change. I am willing to bump to 0.24.0 in principle but it would entail bumping _everything_ (which is kind of time consuming). Also fine with just bumping to 0.23.3, and what we changed is "hey, beta status".

Thoughts?

lukesneeringer:
As a note, when we go GA, we _will_ need to bump the version of the GA libraries to 1.x

dhermes:
@lukesneeringer That expectation works for me. Also, keep in mind we can do an umbrella release without "bumping everything": simple change the API at hand and rev the version (we don't need 23<-->23 to match)

lukesneeringer:
Concur, although I think the consistency is valuable in the immediate term. I am also hoping `google-cloud-core` is much more stable now than it was a couple months ago.

So, am I good to merge this?

dhermes:
Sure thing

lukesneeringer:
Merged. @daspecster Please cut a release and remember to manually bump docs so we see the beta designation on the index page.

daspecster:
@lukesneeringer nothing was changed in the "docs", just the README here just FYI.

PR #3105, opened 2017-03-06
Add basic helpers needed for GAPIC client in datastore.

dhermes:
Includes

- ``use_gax`` argument in Client constructor
- ``make_datastore_api`` helper to make a gRPC channel with
  the correct credentials
- A basic ``HTTPDatastoreAPI`` class to act as the equivalent
  to the GAPIC ``DatastoreClient``
- A lazy property on ``Client`` that will hold the API
  object.

Towards #2746.

### DIFF
--- a/datastore/google/cloud/datastore/_http.py
+++ b/datastore/google/cloud/datastore/_http.py
@@ -467,6 +467,19 @@ class Connection(connection_module.Connection):
         return self._datastore_api.allocate_ids(project, request)
 
 
+class HTTPDatastoreAPI(object):
+    """An API object that sends proto-over-HTTP requests.
+
+    Intended to provide the same methods as the GAPIC ``DatastoreClient``.
+
+    :type client: :class:`~google.cloud.datastore.client.Client`
+    :param client: The client that provides configuration.
+    """
+
+    def __init__(self, client):
+        self.client = client
+
+
 def _set_read_options(request, eventual, transaction_id):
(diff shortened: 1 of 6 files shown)


datastore/google/cloud/datastore/client.py
```diff
@@ -169,24 +180,45 @@ class Client(ClientWithProject):
...
+        else:
+            self._use_gax = use_gax
 
     @staticmethod
     def _determine_default(project):
         """Helper:  override default project detection."""
         return _determine_default_project(project)
 
+    @property
+    def _datastore_api(self):
```
dhermes: @lukesneeringer I can make this public if you think it's important. I am allergic to "accidentally public interfaces".

datastore/google/cloud/datastore/_http.py
```diff
@@ -467,6 +467,19 @@ def allocate_ids(self, project, key_pbs):
         return self._datastore_api.allocate_ids(project, request)
 
 
+class HTTPDatastoreAPI(object):
```
daspecster: Shouldn't this be non-public? `_HTTPDatastoreAPI`?

dhermes: The module is non-public

datastore/google/cloud/datastore/client.py
lukesneeringer: It is fine as-is, at least for now. I am not sure we made an explicit decision.

PR #3108, opened 2017-03-06
Moving build_api_url() off of datastore Connection.

dhermes:
Also sending DEFAULT_USER_AGENT header instead of relying on the Connection.USER_AGENT property.


dhermes:
I can haz code review?

PR #3111, opened 2017-03-06
Storage release 0.23.1

daspecster:
# Storage 0.23.1

- #2991 (de5d8e1) Add 'Blob.update_storage_class' API method.

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -55,7 +55,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-storage',
-    version='0.23.0',
+    version='0.23.1',
     description='Python Client for Google Cloud Storage',
     long_description=README,
     namespace_packages=[


daspecster:
Release tag: https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/storage-0.23.1

tseaver:
I'm still sad about releasing features in a third-dot.

daspecster:
So this didn't really qualify as a patch?

tseaver:
It adds a method / exposes a new API feature (assigning a storage class to a blob).  The issue is that cod written to use the feature will break if used with an earlier version of the same 0.23 release line.

daspecster:
So you're saying it should have been 0.24?

tseaver:
That would've been my preference.  @lukesneeringer has a different take on it, though.

PR #3110, opened 2017-03-06
Only requiring HTTP object to make datastore RPC-over-HTTP.

dhermes:
**NOTE**: Has #3108 as diffbase.

PR #3112, opened 2017-03-06
Moving datastore host configuration into Client.

dhermes:
This is progress towards #2746 and makes Connection() act as a proxy for the contents of Client (as intended).

### DIFF
--- a/datastore/google/cloud/datastore/_gax.py
+++ b/datastore/google/cloud/datastore/_gax.py
@@ -21,6 +21,7 @@ from google.cloud.gapic.datastore.v1 import datastore_client
 from google.cloud.proto.datastore.v1 import datastore_pb2_grpc
 from google.gax.utils import metrics
 from grpc import StatusCode
+import six
 
 from google.cloud._helpers import make_insecure_stub
 from google.cloud._helpers import make_secure_channel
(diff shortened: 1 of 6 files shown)


datastore/google/cloud/datastore/_gax.py
```diff
@@ -90,20 +91,20 @@ class _DatastoreAPIOverGRPC(object):
...
-    :type secure: bool
-    :param secure: Flag indicating if a secure stub connection is needed.
     """
 
-    def __init__(self, connection, secure):
-        if secure:
+    def __init__(self, connection):
+        parse_result = six.moves.urllib_parse.urlparse(
+            connection.api_base_url)
+        host = parse_result.hostname
```
dhermes: I don't feel great about replacing a simple argument with parsing here, and am open to suggestions

dhermes:
Rebased after merging #3110

datastore/google/cloud/datastore/_gax.py
daspecster: This is more concise than it was previously, but why move it?

What about leaving the parsing at the `Connection.__init__` and then just assigning `secure` as a property on `Connection`?

dhermes: The parsing would move to `Client.__init__` if it moved (I want `Connection` to be a dumb proxy). I just thought it was overkill to carry around two values (the host / URL and the `secure` flag)

lukesneeringer: What is the problem you are trying to solve?

dhermes: Highest of high-level: trying to determine if the live backend should be used or the local datastore emulator. If the protocol is `https`, we assume the host is the live backend and if not, we assume the host it the local emulator.

lukesneeringer: That makes sense, thanks.

dhermes: But if you look at [where][1] it is detected, I could easily set:

```
try:
    host = os.environ[GCD_HOST]
    self._base_url = 'http://' + host
    self._secure = False
except KeyError:
    self._base_url = _DATASTORE_BASE_URL
    self._secure = True
```

I just thought it might be overkill to carry around **two** values.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/8b39bb002cf79e6805eae71e2957f950bba8225f/datastore/google/cloud/datastore/client.py#L208-L212

Issue #3029, opened 2017-02-16
Write more integration tests for BigQuery


omaray:
Please see email I sent about the details and test cases from @tswast . 

Also, C# has [some examples](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/ExhaustiveTypesTest.cs) that might be worth looking at.

/cc @jskeet @jba @tswast @danoscarmike

tseaver:
Specific tests needed:

- [ ] Test queries w/ legacy scalar types, including timestamp.
- [ ] Test queries w/ record nested inside record
- [ ] Test `Table.fetch_data` w/ repeated scalars, records, repeated records, nested records, repeated scalars inside records, standard SQL types.
- [ ] Test `Table.insert_data` w/ repeated  scalars, records, repeated records, nested records, repeated scalars inside records, standard SQL types.
- [ ] Test `Table.create` w/  repeated  scalars, records, repeated records, nested records, repeated scalars inside records, standard SQL types.
- [ ] Test query parameters w/ all standard SQL scalar types, repeated records, nested records, repeated scalars inside records.

dhermes:
@tswast Don't you have some fancy spreadsheet defining which tests are worthwhile?

tswast:
@dhermes Yep, [this spreadsheet](https://docs.google.com/spreadsheets/d/1bbYm2Eav9T9zR1y2VkrDcmB2xY3v7xBZSz9VfU3ipq0/edit#gid=1154704566) has the list. Rows are the data types and columns are the features to be tested.

Note: I can't share this publicly since I created it with my corp account, but I can add people individually as needed.

tseaver:
@dhermes, @tswast The checklist above is based on that spreadsheet.

tseaver:
@tswast If you update your sample data / schema, please reopen and post them here.

PR #3117, opened 2017-03-07
Add an explicit test for records nested in records.

tseaver:
Toward #3029.


daspecster:
This LGTM, but it doesn't really cover `n` levels case. Just one level of nesting.

PR #3118, opened 2017-03-08
Add system test for 'Table.create' using nested schema

tseaver:
Derived from schema from @tswast.

Toward #3029.

### DIFF
--- a/system_tests/bigquery.py
+++ b/system_tests/bigquery.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import operator
+import os
 
 import unittest
 
@@ -25,12 +27,23 @@ from retry import RetryResult
 from system_test_utils import unique_resource_id
 
 
+WHERE = os.path.abspath(os.path.dirname(__file__))
+
+
 def _has_rows(result):
     return len(result) > 0
 
 
 def _make_dataset_name(prefix):
     return '%s%s' % (prefix, unique_resource_id())
+
+
+def _load_json_schema(filename='bigquery_test_schema.json'):
+    from google.cloud.bigquery.table import _parse_schema_resource
+    json_filename = os.path.join(WHERE, filename)
+
+    with open(json_filename, 'r') as schema_file:
+        return _parse_schema_resource(json.load(schema_file))
 
 
 def _rate_limit_exceeded(forbidden):
@@ -650,3 +663,18 @@ class TestBigQuery(unittest.TestCase):
         dataset = Config.CLIENT.dataset(DATASET_NAME, project=PUBLIC)
         table = dataset.table(TABLE_NAME)
         self._fetch_single_page(table)
+
+    def test_create_table_nested_schema(self):
+        table_name = 'test_table'
+        dataset = Config.CLIENT.dataset(
+            _make_dataset_name('create_table_nested_schema'))
+        self.assertFalse(dataset.exists())
+
+        retry_403(dataset.create)()
+        self.to_delete.append(dataset)
+
+        table = dataset.table(table_name, schema=_load_json_schema())
+        table.create()
+        self.to_delete.insert(0, table)
+        self.assertTrue(table.exists())
+        self.assertEqual(table.name, table_name)
--- a/system_tests/bigquery_test_schema.json
+++ b/system_tests/bigquery_test_schema.json
@@ -1,0 +1,83 @@
+{
+   "fields" : [
+      {
+         "type" : "STRING",
+         "name" : "Name",
+         "mode" : "NULLABLE"
+      },
+      {
+         "name" : "Age",
+         "mode" : "NULLABLE",
+         "type" : "INTEGER"
+      },
+      {
+         "type" : "FLOAT",
+         "name" : "Weight",
+         "mode" : "NULLABLE"
+      },
+      {
+         "mode" : "NULLABLE",
+         "name" : "IsMagic",
+         "type" : "BOOLEAN"
+      },
+      {
+         "name" : "Spells",
+         "fields" : [
+            {
+               "mode" : "NULLABLE",
+               "name" : "Name",
+               "type" : "STRING"
+            },
+            {
+               "mode" : "NULLABLE",
+               "name" : "LastUsed",
+               "type" : "TIMESTAMP"
+            },
+            {
+               "type" : "STRING",
+               "mode" : "NULLABLE",
+               "name" : "DiscoveredBy"
+            },
+            {
+               "name" : "Properties",
+               "fields" : [
+                  {
+                     "name" : "Name",
+                     "mode" : "NULLABLE",
+                     "type" : "STRING"
+                  },
+                  {
+                     "type" : "FLOAT",
+                     "name" : "Power",
+                     "mode" : "NULLABLE"
+                  }
+               ],
+               "mode" : "REPEATED",
+               "type" : "RECORD"
+            },
+            {
+               "mode" : "NULLABLE",
+               "name" : "Icon",
+               "type" : "BYTES"
+            }
+         ],
+         "mode" : "REPEATED",
+         "type" : "RECORD"
+      },
+      {
+         "type" : "TIME",
+         "mode" : "NULLABLE",
+         "name" : "TeaTime"
+      },
+      {
+         "type" : "DATE",
+         "name" : "NextVacation",
+         "mode" : "NULLABLE"
+      },
+      {
+         "mode" : "NULLABLE",
+         "name" : "FavoriteTime",
+         "type" : "DATETIME"
+      }
+   ]
+}


bigquery/google/cloud/bigquery/table.py
```diff
@@ -1026,7 +1026,7 @@ def _configure_job_metadata(metadata,  # pylint: disable=too-many-arguments
         load_config['writeDisposition'] = write_disposition
 
 
-def _parse_schema_resource(info):
+def parse_schema_resource(info):
```
dhermes: This is an interface change. Not to be taken lightly?

system_tests/bigquery.py
```diff
@@ -33,6 +35,15 @@ def _make_dataset_name(prefix):
     return '%s%s' % (prefix, unique_resource_id())
 
 
+def _load_json_schema(filename='bigquery_test_schema.json'):
+    from google.cloud.bigquery.table import parse_schema_resource
+    where = os.path.dirname(__file__)
+    json_filename = os.path.join(where, filename)
+
+    with open(json_filename) as schema_file:
```
dhermes: I **always** include a mode by habit. Opposed?

system_tests/bigquery.py
```diff
@@ -33,6 +35,15 @@ def _make_dataset_name(prefix):
     return '%s%s' % (prefix, unique_resource_id())
 
 
+def _load_json_schema(filename='bigquery_test_schema.json'):
+    from google.cloud.bigquery.table import parse_schema_resource
+    where = os.path.dirname(__file__)
```
dhermes: 1. I typically use `os.path.abspath` to be "defensive"
1. Why not make a constant?

system_tests/bigquery.py
```diff
@@ -602,3 +613,18 @@ def test_dump_table_w_public_data(self):
         dataset = Config.CLIENT.dataset(DATASET_NAME, project=PUBLIC)
         table = dataset.table(TABLE_NAME)
         self._fetch_single_page(table)
+
+    def test_create_table_nested_schema(self):
+        TABLE_NAME = 'test_table'
```
dhermes: 1. Not a constant
1. Worth making more unique via `unique_resource_id`?

bigquery/google/cloud/bigquery/table.py
tseaver: The other choices are a) have the system test depend on a private helper, or b) fork the recursive code into the system test module.  Making a previously private helper public seemed the best choice.

dhermes: In the short term, making the system test depend on the private helper is a much better fix than (somewhat arbitrarily) adding to the publicly supported feature set.

system_tests/bigquery.py
tseaver: > 1. I typically use os.path.abspath to be "defensive".

OK.  The interpreter is the thing generating it, not a user, but it can't hurt.

> 2.  Why not make a constant?

It is only used by this function.

system_tests/bigquery.py
tseaver: Nope.

system_tests/bigquery.py
tseaver: > 1.  Not a constant

Huh?  It is as constant as any other name bound to an immutable value.

> 2.    Worth making more unique via unique_resource_id?

The dataset which will contain (only) this table already gets a per-run ID.

system_tests/bigquery.py
tseaver: 2ea2b04df389b595539b3b660e409419170e7228

system_tests/bigquery.py
tseaver: 2ea2b04df389b595539b3b660e409419170e7228

system_tests/bigquery.py
dhermes: I mean "constant" in the way that it is typically used in programming contexts. By the classification you've used, any immutable **literal** should have a variable name in all-caps.

bigquery/google/cloud/bigquery/table.py
tseaver: We've had [hygeine issues](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1059) in the past where we expended effort to *remove* such usage.

dhermes: Yes but the solution is not to add to the public interface, especially if we didn't think it was useful before the test-only change.

tseaver: This is a function which converts the REST API representation of a list of schema fields into a list of our own `SchemaField` items.  Doesn't its use in a system test constitute prima-facie evidence that it is "externally useful"?  If not, how would you suggest that we implement it?  The JSON file is a pretty-printed form of the JSON example provided by the back-end team:  I guess we could construct the list of `SchemaField` instances manually in Python, but then we are stuck with updating it manually when the back-end updates their sample.

system_tests/bigquery.py
tseaver: I don't understand how this is "not a constant" -- it is a semantically-named label for what would otherwise be a "magic" literal.  Is it because we aren't defining it at module scope?  That would be a silly requirement to start enforcing:  we have literally thousands of examples of non-module-scope, all-caps variable assignments in 164 modules scattered throughout our code.

tseaver: E.g.:

```bash
$ git grep " [A-Z][A-Z]* =" | wc -l
2218
$ git grep -l " [A-Z][A-Z]* ="  | wc -l
164
$ git grep -l " [A-Z][A-Z]* =" 
bigquery/README.rst
bigquery/google/cloud/bigquery/client.py
bigquery/google/cloud/bigquery/job.py
bigquery/setup.py
bigquery/unit_tests/test__helpers.py
bigquery/unit_tests/test__http.py
bigquery/unit_tests/test_client.py
bigquery/unit_tests/test_dataset.py
bigquery/unit_tests/test_job.py
bigquery/unit_tests/test_query.py
bigquery/unit_tests/test_table.py
bigtable/google/cloud/bigtable/_generated/_common.proto
bigtable/google/cloud/bigtable/_generated/_instance.proto
bigtable/google/cloud/bigtable/_generated/_table.proto
bigtable/google/cloud/bigtable/_generated/bigtable_instance_admin_pb2.py
bigtable/google/cloud/bigtable/_generated/bigtable_pb2.py
bigtable/google/cloud/bigtable/_generated/bigtable_table_admin_pb2.py
bigtable/google/cloud/bigtable/_generated/data_pb2.py
bigtable/google/cloud/bigtable/_generated/instance_pb2.py
bigtable/google/cloud/bigtable/_generated/table_pb2.py
bigtable/google/cloud/bigtable/row_data.py
bigtable/setup.py
bigtable/unit_tests/test_client.py
bigtable/unit_tests/test_cluster.py
bigtable/unit_tests/test_instance.py
bigtable/unit_tests/test_row_data.py
bigtable/unit_tests/test_table.py
core/google/cloud/_helpers.py
core/google/cloud/client.py
core/setup.py
core/unit_tests/streaming/test_buffered_stream.py
core/unit_tests/streaming/test_exceptions.py
core/unit_tests/streaming/test_http_wrapper.py
core/unit_tests/streaming/test_stream_slice.py
core/unit_tests/streaming/test_transfer.py
core/unit_tests/test__helpers.py
core/unit_tests/test__http.py
core/unit_tests/test_client.py
core/unit_tests/test_credentials.py
core/unit_tests/test_exceptions.py
core/unit_tests/test_iterator.py
core/unit_tests/test_operation.py
... 
system_tests/bigquery.py
system_tests/bigtable.py
system_tests/datastore.py
system_tests/language.py
system_tests/logging_.py
system_tests/monitoring.py
system_tests/pubsub.py
system_tests/spanner.py
system_tests/speech.py
system_tests/storage.py
system_tests/translate.py
system_tests/vision.py
translate/google/cloud/translate/client.py
translate/setup.py
translate/unit_tests/test__http.py
vision/google/cloud/vision/client.py
vision/google/cloud/vision/likelihood.py
vision/setup.py
```

dhermes: I am aware of the copious usage throughout our codebase, but I'm saying that use is inconsistent with the actual meaning of "constant" and that we shouldn't make it worse here.

bigquery/google/cloud/bigquery/table.py
dhermes: Can we punt on the "make this public" discussion? Public interfaces aren't easy to walk away from and it's certainly not the point of this PR.

system_tests/bigquery.py
jonparrott: I agree with @dhermes here - "constants" in Python use are module-level or class-level things that are *intended* to be immutable. If we start making stuff within a function scope a constant we might as well just turn on caps-lock and be done with it.

bigquery/google/cloud/bigquery/table.py
tseaver: By "punt"  do you mean just use the private helper function?

system_tests/bigquery.py
tseaver: This seems like another "opinion without lint support" issue to me, but I've change this case.

bigquery/google/cloud/bigquery/table.py
dhermes: Yes. By punt I meant "leave as-is for now, can discuss the change in a follow-up"

system_tests/bigquery.py
dhermes: But that's because we've [disabled `invalid-name`][1] for tests. We can / should re-enable it? Though @lukesneeringer seems to be dumping Pylint for `flake8`

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/cba8b9f18dda1e4510e5c3990cf84987f0e4b6ff/scripts/run_pylint.py#L57

system_tests/bigquery.py
```diff
@@ -33,6 +38,14 @@ def _make_dataset_name(prefix):
     return '%s%s' % (prefix, unique_resource_id())
 
 
+def _load_json_schema(filename='bigquery_test_schema.json'):
+    from google.cloud.bigquery.table import _parse_schema_resource
```
daspecster: I would add a newline after the import.

I was looking at adding this to lint through pycodestyle but I haven't proposed the change yet.

tseaver:

system_tests/bigquery.py
tseaver: I will tweak that in #3140

PR #3122, opened 2017-03-08
'gcloud preview' command no longer supported.

daspecster:
>141.0.0 (2017-01-25) (Google Cloud SDK) The deprecated gcloud preview app and gcloud preview datastore have been removed. Commands are available under the GA release track.

See: https://cloud.google.com/sdk/docs/release-notes


daspecster:
I cancelled the builds since this is an RST change.

PR #3126, opened 2017-03-09
Using GAPIC datastore object (and an HTTP equivalent) for rollback.

dhermes:
Towards #2746.


dhermes:
Merging on the basis of Travis' green-ness (will be nice when CircleCI builds also limit based on the changed packages)

PR #3127, opened 2017-03-09
Using GAPIC datastore object (and an HTTP equivalent) for commit.

dhermes:
This is a bit more involved since

- The signature requires the **caller** to determine the mode
- The signature receives a list of mutations, so we no longer collect mutations within a `CommitRequest` object
- The exception re-mapping must be done, so we can't use the base `datastore_client.DatastoreClient` GAPIC object

----

~~**NOTE**: I have yet to write unit tests, but have the system tests passing and wanted to get some eyeballs on this before I dive into unit test hell.~~

### DIFF
--- a/datastore/google/cloud/datastore/_gax.py
+++ b/datastore/google/cloud/datastore/_gax.py
@@ -16,6 +16,7 @@
 
 
 import contextlib
+import sys
 
 from google.cloud.gapic.datastore.v1 import datastore_client
 from google.cloud.proto.datastore.v1 import datastore_pb2_grpc
@@ -79,14 +80,16 @@ def _grpc_catch_rendezvous():
         if error_class is None:
             raise
         else:
-            raise error_class(exc.cause.details())
+            new_exc = error_class(exc.cause.details())
+            six.reraise(error_class, new_exc, sys.exc_info()[2])
     except exceptions.GrpcRendezvous as exc:
         error_code = exc.code()
         error_class = _GRPC_ERROR_MAPPING.get(error_code)
         if error_class is None:
             raise
         else:
-            raise error_class(exc.details())
+            new_exc = error_class(exc.details())
+            six.reraise(error_class, new_exc, sys.exc_info()[2])
 
 
 class _DatastoreAPIOverGRPC(object):
--- a/datastore/google/cloud/datastore/_gax.py
+++ b/datastore/google/cloud/datastore/_gax.py
@@ -19,6 +19,8 @@ import contextlib
 
 from google.cloud.gapic.datastore.v1 import datastore_client
 from google.cloud.proto.datastore.v1 import datastore_pb2_grpc
+from google.gax.errors import GaxError
+from google.gax.grpc import exc_to_code
 from google.gax.utils import metrics
 from grpc import StatusCode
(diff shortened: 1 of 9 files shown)


datastore/google/cloud/datastore/_gax.py
```diff
@@ -71,6 +73,13 @@ def _grpc_catch_rendezvous():
     """
     try:
         yield
+    except GaxError as exc:
+        error_code = exc_to_code(exc.cause)
+        error_class = _GRPC_ERROR_MAPPING.get(error_code)
+        if error_class is None:
+            raise
+        else:
+            raise error_class(exc.cause.details())
```
dhermes: Note this violates DRY but once all 6 datastore methods are on GAPIC we will just delete the `exceptions.GrpcRendezvous` code

lukesneeringer: Assuming that it works like I claim it does, can this be changed to use `six.reraise` and maintain the _original_ traceback, please? I have done the same in [this PR][1] on GAX, pending some real-world testing. Assuming I am right about exactly how this works, it should make debugging substantially easier.

  [1]: https://github.com/googleapis/gax-python/pull/168/files#diff-5ce6a350c255795b7c02e598ab053273R149

datastore/google/cloud/datastore/_http.py
```diff
@@ -410,6 +362,39 @@ class HTTPDatastoreAPI(object):
...
+        :rtype: :class:`.datastore_pb2.CommitResponse`
+        :returns: The returned protobuf response object.
+        """
+        request_pb = _datastore_pb2.CommitRequest(
+            project_id=project,
+            mode=mode,
+            transaction=transaction,
+            mutations=mutations,
+        )
+        return _rpc(self.client._http, project, 'commit',
```
lukesneeringer: Is this an RPC over HTTP call, I take it?

dhermes: Yeah the HTTP API is RPC-like.

dhermes:
@lukesneeringer @tseaver @daspecster PTAL have all unit tests (and system tests) passing

datastore/google/cloud/datastore/batch.py
```diff
@@ -133,7 +135,8 @@ def _add_delete_key_pb(self):
         :returns: The newly created key protobuf that will be
                   deleted when sent with a commit.
         """
-        new_mutation = self.mutations.add()
+        new_mutation = _datastore_pb2.Mutation()
+        self._mutations.append(new_mutation)
         return new_mutation.delete
 
     @property
```
daspecster: Since this is just a 'getter' and it's *just* returning `self._mutations`, could this property be removed?

I'm not sure why someone would need to use it practically?

dhermes: @daspecster I am :+1: to that but this interface is "beta" so I didn't want to break it.

I'm 99% sure that the `mutations` property was added by me to aid in the transition from `v1beta3` to `v1`.

@lukesneeringer WDYT?

lukesneeringer: Leave as-is.

PR #3135, opened 2017-03-10
CI rehash

lukesneeringer:
I am not yet specifically asking for review on this (although @dhermes has already been providing feedback "as I go"). This is a PR at this point because I need to be able to thrash at CircleCI.

-------------------------------------------------

This is an attempt to fix the CI situation on this repository. In particular, this will hit the following goals:

  * Move to parallel builds on a per-API basis.
  * Move system tests into the individual API packages.
  * Eliminate Travis, move CircleCI to 2.0

In doing this, I made several changes:

  * Move from tox to nox (Python-based configuration is really useful for the multi-project situation)
  * Move from pylint to flake8 (pylint does not work on Python 3.6, has multiple difficult-to-silence false positives)

**This is not done.** In particular, several things need to be put back into place:

  * [x] Automatic docs build.
  * [ ] Automatic PyPI releasing on appropriate tags.
  * [x] Only testing APIs that have changed files.


bigtable/nox.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Install all test dependencies, then install this package in-place.
+    session.chdir(os.path.dirname(__file__))
+    session.install('mock', 'pytest', 'pytest-cov', '../core/')
+    session.install('-e', '.')
+
+    # Run py.test against the unit tests.
+    session.run('py.test', '--quiet',
+        '--cov=google.cloud.bigtable', '--cov=tests.unit', '--cov-append',
+        '--cov-config=.coveragerc', '--cov-report=', '--cov-fail-under=97',
```
daspecster: I think we've held 100% so far right? `--cov-fail-under=97`?

logging/google/cloud/logging/handlers/__init__.py
```diff
@@ -19,3 +19,6 @@
     ContainerEngineHandler)
 from google.cloud.logging.handlers.handlers import CloudLoggingHandler
 from google.cloud.logging.handlers.handlers import setup_logging
+
+__all__ = ['AppEngineHandler', 'CloudLoggingHandler', 'ContainerEngineHandler',
```
daspecster: What is this required for? Seems like it could/should be a different PR?

bigquery/nox.py
```diff
@@ -0,0 +1,86 @@
...
+@nox.session
+@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
+def unit_tests(session, python_version):
+    """Run the unit test suite."""
+
+    # Run unit tests against all supported versions of Python.
+    session.interpreter = 'python%s' % python_version
+
+    # Install all test dependencies, then install this package in-place.
+    session.chdir(os.path.dirname(__file__))
```
jonparrott: Should Nox always chdir to `dirname(noxfile)`? I think that's a reasonable behavior change.

lukesneeringer: I was planning on adding that soon. :-)

bigtable/nox.py
lukesneeringer: Some of the packages do not get to 100% on each distinct version of Python (but the union of all of them is 100%). Note the `cover` session, which requires 100%.

logging/google/cloud/logging/handlers/__init__.py
lukesneeringer: It is required for flake8 not to complain about an unused import.

bigtable/nox.py
tseaver: I'm -1 on asserting anything less than 100% (97% is just arbitrary).  If we have an omnibus assertion for "across all languages == 100%", that should be good enough.

logging/google/cloud/logging/handlers/__init__.py
tseaver: ISTM we should be disabling that check for package-level exports, rather than adding non-DRY `_all_` which we would otherwise avoid.

jonparrott: I'm very +1 on using `__all__`, as it tells users, sphinx, and pylint what the intended exports are for a package.

bigtable/nox.py
lukesneeringer: Agreed in principle, but it defaults to 100, so I needed _something_. I suppose I could make it zero. On the other hand, if excessively little is being covered in any given version, that is probably an indicator of a problem, too.

daspecster: 💯 😃

PR #3138, opened 2017-03-13
Using GAPIC datastore object (and an HTTP equivalent) for lookup.

dhermes:
This is the last method to be ported over so the `Connection()` base class as well as the `_DatastoreAPIOverGRPC` and `_DatastoreAPIOverHttp` helper classes have been totally removed.

**NOTE**: This temporarily breaks our emulator support, but I plan on restoring that support in a follow-up PR.

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -15,6 +15,8 @@
 
 import os
 
+from google.cloud.proto.datastore.v1 import datastore_pb2 as _datastore_pb2
+
 from google.cloud._helpers import _LocalStack
 from google.cloud._helpers import (
     _determine_default_project as _base_default_project)
@@ -23,7 +25,6 @@ from google.cloud.environment_vars import DISABLE_GRPC
 from google.cloud.environment_vars import GCD_DATASET
 from google.cloud.environment_vars import GCD_HOST
 
-from google.cloud.datastore._http import Connection
 from google.cloud.datastore._http import HTTPDatastoreAPI
 from google.cloud.datastore import helpers
 from google.cloud.datastore.batch import Batch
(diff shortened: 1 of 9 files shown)


daspecster:
Is there any reason we couldn't wait until the PR to restore  emulator support is up and then we can review these at the same time and minimize the emulator breakage?

dhermes:
I don't think it's worth the time or effort to keep it working. I am going to send a follow up PR the moment this is merged, so it will be broken for at most 1 day.

datastore/google/cloud/datastore/client.py
```diff
@@ -569,3 +566,34 @@ def do_something(entity):
...
+    """Validate rules for read options, and assign to the request.
+
+    Helper method for ``lookup()`` and ``run_query``.
+
+    :type eventual: bool
+    :param eventual: Flag indicating if ``EVENTUAL`` or ``STRONG``
+                     consistency should be used.
+
+    :type transaction_id: bytes
+    :param transaction_id: A transaction identifier (may be null).
```
daspecster: Nit but mabe 'null' should be `None`?

dhermes: I'll pull it into the next PR?

dhermes:
Turns out emulator support has been b0rken for awhile:

```
======================================================================
ERROR: setUpModule (datastore)
  File ".../google-cloud-python/system_tests/datastore.py", line 82, in setUpModule
    http=http)
  File ".../google-cloud-python/.tox/datastore-emulator/lib/python2.7/site-packages/google/cloud/datastore/client.py", line 200, in __init__
    project=project, credentials=credentials, http=http)
  File ".../google-cloud-python/.tox/datastore-emulator/lib/python2.7/site-packages/google/cloud/client.py", line 199, in __init__
    Client.__init__(self, credentials=credentials, http=http)
  File ".../google-cloud-python/.tox/datastore-emulator/lib/python2.7/site-packages/google/cloud/client.py", line 122, in __init__
ValueError: This library only supports credentials from google-auth-library-python. See https://google-cloud-python.readthedocs.io/en/latest/google-cloud-auth.html for help on authentication with this library.
```

Fix coming soon.

PR #3140, opened 2017-03-14
Add system test for insert/fetch of rows w/ nested schema.

tseaver:
~~Uses #3118 as a base.~~

Derived from schema / data from @tswast.

Toward #3029.


daspecster:
@tseaver should be able to merge #3118 and rebase this off master.

Also, there's a lint error in this branch just FYI.

system_tests/bigquery.py
```diff
@@ -673,8 +670,58 @@ def test_create_table_nested_schema(self):
...
+
+        errors = table.insert_data(to_insert)
+        self.assertEqual(len(errors), 0)
+
+        retry = RetryResult(_has_rows, max_tries=8)
+        fetched = retry(self._fetch_single_page)(table)
+        self.assertEqual(len(fetched), len(to_insert))
+
+        for found, expected in zip(sorted(fetched), sorted(to_insert)):
+            self.assertEqual(found[0], expected[0])            # Name
```
dhermes: Can you make an assertion on `len(found) == len(expected) == 8`?

system_tests/bigquery.py
```diff
@@ -673,8 +670,58 @@ def test_create_table_nested_schema(self):
...
+            self.assertEqual(found[1], int(expected[1]))       # Age
```
dhermes: Do we control `expected`? If yes, why can't we just store an `int`?

system_tests/bigquery.py
```diff
@@ -673,8 +670,58 @@ def test_create_table_nested_schema(self):
...
+                    e_spell['LastUsed'], '%Y-%m-%d %H:%M:%S UTC')
+                e_used = datetime.datetime(*parts[0:6], tzinfo=UTC)
+                self.assertEqual(f_spell['LastUsed'], e_used)
+                self.assertEqual(f_spell['DiscoveredBy'],
+                                 e_spell['DiscoveredBy'])
+                self.assertEqual(f_spell['Properties'], e_spell['Properties'])
+
+                e_icon = base64.standard_b64decode(
+                    e_spell['Icon'].encode('ascii'))
+                self.assertEqual(f_spell['Icon'], e_icon)
```
dhermes: Why does this have to be so involved? Why can't we just say `found[4] == expected[4]`?

system_tests/bigquery_test_data.json
```diff
@@ -0,0 +1,66 @@
+[
```
dhermes: Where does this data come from?

system_tests/bigquery.py
tseaver: I asserted that the lengths were equal above (line 694):  why would I assert a specific length, when that is driven by the data file, and might change if @tswast provides updates?

system_tests/bigquery.py
tseaver: `expected` is read from disk (data provided by @tswast).  `found` is generated from the resource from the server, using the schema to convert the JSON repr into native Python types.

system_tests/bigquery.py
tseaver: `found[4]` is the *actual* bytes value, already converted from the JSON-mandated base64-ified text.

system_tests/bigquery_test_data.json
tseaver: @tswast's examples, derived from the bigquery doc snippets (not ours, the ones used on cloud.google.com).

tseaver:
@dhermes I think I answered your questions.

tseaver:
@dhermes, @lukesneeringer, @daspecster Any remaining issues?

lukesneeringer:
None from me. :-)

Issue #3129, opened 2017-03-10
Bug, double conversion.


sactre:
in #2860 line: 163, file: logging/google/cloud/logging/logger.py

the function _make_entry_resource return a resource(dict), but de timestamp element has to be a datetime not a string. Because in the next step the timestamp has a conversion from datetime to string (file: _gax.py, function: _log_entry_mapping_to_pb).

Fkawala:
@sactre that's wired ! from the source I get the same understanding than you, it should fail. However, I currently use the timestamp parameter in the log_struct function and it works like a charm. 

Do you have a stack trace to investigate on ?

plowman:
@Fkawala I'm seeing the same thing that @sactre sees. Specifically, if I specify a timestamp, it fails.

**Using a string timestamp**
```python
    from google.cloud import logging as google_logging
    one_week_ago = datetime.utcnow() - timedelta(days=7)
    one_week_ago = one_week_ago.strftime("%Y-%m-%dT%H:%M:%S.%fZ")

    logging_client = google_logging.Client(project="my-project")
    logger = logging_client.logger("mobile-logger")

    logger.log_text("My Logs Here", timestamp=one_week_ago)
```

**String timestamp stack trace:**
```
  File "/project/project3/project/servers/apis/mobile/api_helpers.py", line 261, in _inner_method
    return api_method(api_instance, request, *args, **kwargs)
  File "/project/project3/project/servers/apis/mobile/resources/logs.py", line 144, in write_bulk
    logger.log_text("My Logs Here", timestamp=one_week_ago)
  File "/home/vagrant/.virtualenvs/project3/local/lib/python2.7/site-packages/google/cloud/logging/logger.py", line 201, in log_text
    http_request=http_request, timestamp=timestamp)
  File "/home/vagrant/.virtualenvs/project3/local/lib/python2.7/site-packages/google/cloud/logging/logger.py", line 163, in _make_entry_resource
    resource['timestamp'] = _datetime_to_rfc3339(timestamp)
  File "/home/vagrant/.virtualenvs/project3/local/lib/python2.7/site-packages/google/cloud/_helpers.py", line 329, in _datetime_to_rfc3339
    return value.strftime(_RFC3339_MICROS)
AttributeError: 'str' object has no attribute 'strftime'
```

**Using a datetime timestamp**
```python
    from google.cloud import logging as google_logging
    one_week_ago = datetime.utcnow() - timedelta(days=7)

[...]

PR #3144, opened 2017-03-15
Fix double conversion of datetime for log entries.

daspecster:
Fixes: #3129

This should fix the double conversion issue.

/cc @waprin

### DIFF
--- a/logging/unit_tests/test__gax.py
+++ b/logging/unit_tests/test__gax.py
@@ -369,6 +369,7 @@ class Test_LoggingAPI(_Base, unittest.TestCase):
         from datetime import datetime
         from google.logging.type.log_severity_pb2 import WARNING
         from google.cloud.proto.logging.v2.log_entry_pb2 import LogEntry
+        from google.cloud._helpers import _datetime_to_rfc3339
         from google.cloud._helpers import UTC, _pb_timestamp_to_datetime
 
         NOW = datetime.utcnow().replace(tzinfo=UTC)
@@ -412,7 +413,7 @@ class Test_LoggingAPI(_Base, unittest.TestCase):
             'severity': SEVERITY,
             'labels': LABELS,
             'insertId': IID,
-            'timestamp': NOW.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
+            'timestamp': _datetime_to_rfc3339(NOW),
             'httpRequest': REQUEST,
             'operation': OPERATION,
         }
--- a/logging/unit_tests/test__gax.py
+++ b/logging/unit_tests/test__gax.py
@@ -412,7 +412,7 @@ class Test_LoggingAPI(_Base, unittest.TestCase):
             'severity': SEVERITY,
             'labels': LABELS,
             'insertId': IID,
-            'timestamp': NOW,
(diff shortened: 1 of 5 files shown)


logging/unit_tests/test__gax.py
```diff
@@ -412,7 +412,7 @@ def test_write_entries_w_extra_properties(self):
             'severity': SEVERITY,
             'labels': LABELS,
             'insertId': IID,
-            'timestamp': NOW,
+            'timestamp': NOW.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
```
daspecster: This is added because this would be a string at this point but this test skips the further up building of the request.

Fkawala: I think you'll have to add a conversion from `datetime` to `string` in the `commit` method from the `Batch` class ([link to code](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/logger.py#L479)).

daspecster: Ah good call! Updating.

daspecster: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3144/commits/7d9bce00e94cd31df812519d0af09d4d8dc9fe88

tseaver: Please use `google.cloud._helpersRFC3339_MICROS`, rather than the literal format string.  Even better would be to call `_datetime_to_rfc3339`.

logging/unit_tests/test_logger.py
```diff
@@ -667,10 +667,13 @@ def test_commit_w_bound_client(self):
             'type': 'global',
         }
         ENTRIES = [
-            {'textPayload': TEXT, 'insertId': IID1, 'timestamp': TIMESTAMP1},
-            {'jsonPayload': STRUCT, 'insertId': IID2, 'timestamp': TIMESTAMP2},
+            {'textPayload': TEXT, 'insertId': IID1,
+             'timestamp': TIMESTAMP1.strftime("%Y-%m-%dT%H:%M:%S.%fZ")},
```
tseaver: Likewise here (and throughout).

system_tests/logging_.py
```diff
@@ -130,6 +130,21 @@ def test_log_text(self):
...
+        logger = Config.CLIENT.logger(self._logger_name())
+        now = datetime.datetime.utcnow()
+
+        self.to_delete.append(logger)
+
+        logger.log_text(text_payload, timestamp=now)
+        entries = _list_entries(logger)
+        self.assertEqual(len(entries), 1)
+        self.assertEqual(entries[0].payload, text_payload)
+        # self.assertEqual(entries[0].timestamp, now)
```
tseaver: Shouldn't be commented-out:  make a real assertion, or remove it.

daspecster:
@tseaver updated!

daspecster:
Gah..forgot to squash...apologies.

Fkawala:
@daspecster thanks for the swift fix !

daspecster:
@Fkawala thanks for tracking it down!

PR #3148, opened 2017-03-15
Adding (back) support for datastore emulator.

dhermes:
Sorry for the delay. I made a few wrong turns (e.g. using `hostname` instead of `netloc`) and debugging with gRPC is **not** fun.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -300,7 +300,9 @@ commands =
     {[emulator]emulatorcmd} --package=datastore
 setenv = {[emulator]setenv}
 passenv = {[testing]passenv}
-deps = {[emulator]deps}
+deps =
+    {[emulator]deps}
+    Sphinx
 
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -300,7 +300,9 @@ commands =
     {[emulator]emulatorcmd} --package=datastore
 setenv = {[emulator]setenv}
 passenv = {[testing]passenv}
-deps = {[emulator]deps}
+deps =
+    {[emulator]deps}
+    Sphinx
```
tseaver: Why is `Sphinx` a dependency for this part?

dhermes: Because it runs the `datastore` system tests, which happen to run the doctests.

tseaver: OK, LGTM then.

PR #3156, opened 2017-03-16
Provide full Mention objects, not just the content.

### LINKED ISSUE #3149: Language: support mention type in Entity.mentions.
[Currently](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/language/google/cloud/language/entity.py#L79) the mentions property of an entity is only a list of strings whereas it should be a list of objects containing the mention text and mention type.

Furthermore, this change should add mention_type information to the mention documentation.

lukesneeringer:
Fixes #3149.

### DIFF
--- a/language/google/cloud/language/entity.py
+++ b/language/google/cloud/language/entity.py
@@ -46,6 +46,80 @@ class EntityType(object):
     """Other entity type (i.e. known but not classified)."""
 
 
+class MentionType(object):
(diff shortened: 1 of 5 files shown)


lukesneeringer:
Note: This represents a backwards-incompatible change, so we need to bump language (at minimum) to 0.24.0.

PR #3157, opened 2017-03-16
Add a `max_size` argument to Pub / Sub Batch.

### LINKED ISSUE #3154: Batch upload fails when messages exceed API limit
1. OS type and version
All OS and all versions supporting topic.Batch
2. Steps to reproduce
Create a batch context & publish messages to it
On batch commit an API failure occurs due to batch size being too large

Suggestion:
In the publish function check the serialised size of the message does not exceed 10MB and that the number of messages doesn't exceed 1000. If it does then commit the batch prematurely and clear the batch queue.

I've crudely implemented it in my own project (just for size, not quantity):
```
class SizedBatch(topic.Batch):
    def __init__(self, topic, client):
        super(SizedBatch, self).__init__(topic, client)
        self.messages_size = 0

    def publish(self, message, **attrs):
        self.topic._timestamp_message(attrs)
        item = {'data': message, 'attributes': attrs}
        self.messages.append(item)
        self.messages_size += len(base64.b64encode(json.dumps(item, ensure_ascii=False).encode('utf-8')))
        if self.messages_size > 95e5:
            self.commit()
            self.messages_size = 0
```
Of course this adds the published message to the queue before checking, allowing a case in which the message creates the API violation but from my own observations the API accepts messages of 15MB and sometimes of 20MB. This also does not take into account the other aspects of the message (which are largely insignificant compared to the message content). Another issue with this snippet is that self.messages_size is not reset if performing a manual commit() on the batch.

lukesneeringer:
Fixes #3154

### DIFF
--- a/pubsub/google/cloud/pubsub/topic.py
+++ b/pubsub/google/cloud/pubsub/topic.py
@@ -14,7 +14,11 @@
 
 """Define API Topics."""
 
+import base64
+import json
 import time
+
+import six
 
 from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud._helpers import _NOW
@@ -426,11 +430,16 @@ class Batch(object):
                          before automatically commiting. Defaults to infinity
                          (off).
     :type max_messages: float
+
+    :param max_size: The maximum size that the serialized messages can be
+                     before automatically commiting. Defaults to just under
+                     10 MB.
+    :type max_size: int
     """
     _INFINITY = float('inf')
 
     def __init__(self, topic, client, max_interval=_INFINITY,
-                 max_messages=_INFINITY):
+                 max_messages=_INFINITY, max_size=10e6):
         self.topic = topic
         self.messages = []
         self.message_ids = []
@@ -440,9 +449,12 @@ class Batch(object):
         # is exceeded, then the .publish() method will imply a commit.
         self._max_interval = max_interval
         self._max_messages = max_messages
+        self._max_size = max_size
 
-        # Set the initial starting timestamp (used against the interval).
+        # Set the initial starting timestamp (used against the interval)
+        # and initial size.
         self._start_timestamp = time.time()
+        self._current_size = 0
 
     def __enter__(self):
         return self
@@ -464,21 +476,38 @@ class Batch(object):
         :param attrs: key-value pairs to send as message attributes
         """
         self.topic._timestamp_message(attrs)
-        self.messages.append(
-            {'data': message,
-             'attributes': attrs})
+
+        # Append the message to the list of messages..
+        item = {'attributes': attrs, 'data': message}
+        self.messages.append(item)
+
+        # Determine the approximate size of the message, and increment
+        # the current batch size appropriately.
+        encoded = b''
+        if isinstance(message, six.text_type):
+            encoded += base64.b64encode(message.encode('utf8'))
+        else:
+            encoded += base64.b64encode(message)
+        encoded += base64.b64encode(
+            json.dumps(attrs, ensure_ascii=False).encode('utf8'),
+        )
+        self._current_size += len(encoded)
 
         # If too much time has elapsed since the first message
         # was added, autocommit.
         now = time.time()
         if now - self._start_timestamp > self._max_interval:
             self.commit()
-            self._start_timestamp = now
             return
 
         # If the number of messages on the list is greater than the
         # maximum allowed, autocommit (with the batch's client).
         if len(self.messages) >= self._max_messages:
+            self.commit()
+            return
+
+        # If we have reached the max size, autocommit.
+        if self._current_size >= self._max_size:
             self.commit()
             return
 
@@ -499,3 +528,5 @@ class Batch(object):
         message_ids = api.topic_publish(self.topic.full_name, self.messages[:])
         self.message_ids.extend(message_ids)
         del self.messages[:]
+        self._start_timestamp = time.time()
+        self._current_size = 0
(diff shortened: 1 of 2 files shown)


pubsub/unit_tests/test_topic.py
```diff
@@ -779,10 +779,35 @@ def test_context_mgr_failure(self):
         self.assertEqual(list(batch.messages), [MESSAGE1, MESSAGE2])
         self.assertEqual(getattr(api, '_topic_published', self), self)
 
+    def test_batch_messages(self):
```
lukesneeringer: This unit test was added because my _previous_ unit tests (now corrected) were accepting invalid data but still passing.

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -426,11 +434,16 @@ class Batch(object):
...
+    :param max_size: The maximum size that the serialized messages can be
+                     before automatically commiting. Defaults to just under
+                     10 MB.
+    :type max_size: int
     """
     _INFINITY = float('inf')
 
     def __init__(self, topic, client, max_interval=_INFINITY,
-                 max_messages=_INFINITY):
+                 max_messages=_INFINITY, max_size=10e6):
```
dhermes: 1. You want `1e6`
1. `1e6` will be a float, not an integer, you almost certainly want an integer?

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -464,16 +480,28 @@ def publish(self, message, **attrs):
...
+        item = {'attributes': attrs, 'data': message}
+        self.messages.append(item)
+
+        # Determine the approximate size of the message, and increment
+        # the current batch size appropriately.
+        encoded = b''
+        if isinstance(message, six.text_type):
+            encoded += base64.b64encode(message.encode('utf8'))
+        else:
+            encoded += base64.b64encode(message)
```
dhermes: We have `google.cloud._helpers._to_bytes` for this

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -464,16 +480,28 @@ def publish(self, message, **attrs):
...
+        encoded += base64.b64encode(
+            json.dumps(attrs, ensure_ascii=False).encode('utf8'),
```
dhermes: Won't `ensure_ascii` dump characters?

pubsub/google/cloud/pubsub/topic.py
```diff
@@ -499,3 +532,5 @@ def commit(self, client=None):
         message_ids = api.topic_publish(self.topic.full_name, self.messages[:])
         self.message_ids.extend(message_ids)
         del self.messages[:]
+        self._start_timestamp = time.time()
+        self._current_size = 0
```
dhermes: Seems a `reset_state` helper (or similar) would make sense here and in the constructor

pubsub/google/cloud/pubsub/topic.py
tseaver: The [pubsub docs for publishing](https://cloud.google.com/pubsub/docs/publisher#publish) state that the limit is 10MB after encoding (but would seem to imply *per message*).

pubsub/google/cloud/pubsub/topic.py
tseaver: The Python docs for `json.dumps` say:

> If `ensure_ascii` is false, all non-ASCII characters are not escaped,
> and  the return value may be a `unicode` instance. See `dump` for details.

And for `json.dump`:

> If ``ensure_ascii`` is true (the default), all non-ASCII characters in the
> output are escaped with ``\uXXXX`` sequences, and the result is a ``str``
> instance consisting of ASCII characters only.  If ``ensure_ascii`` is
> ``False``, some chunks written to ``fp`` may be ``unicode`` instances.
> This usually happens because the input contains unicode strings or the
> ``encoding`` parameter is used. Unless ``fp.write()`` explicitly
> understands ``unicode`` (as in ``codecs.getwriter``) this is likely to
> cause an error.

dhermes: :+1:

lukesneeringer: Additionally, it _probably_ does not matter if we lost some characters at this point. What matters is that the size is "close enough".

pubsub/google/cloud/pubsub/topic.py
lukesneeringer: If the published limit is 10 MB, I wonder if it might make more sense to have the batching limit be a little smaller (e.g. 9 MB). There is nothing here that errors out if you send a single message above `max_size`, and for the common use case (sending a bunch of messages), having auto-commit tack in at 90% of the limit is probably actually the most useful behavior.

lukesneeringer: > You want `1e6`

No, I wanted 10 MB, not 1 MB.

> 1e6 will be a float, not an integer, you almost certainly want an integer?

Anything comparable to `int` or `float` is honestly fine here. I consider `10e6` to be more readable than `10000000` or `1024 * 1024 * 10` (the latter would be more accurate though). I could be persuaded to something else.

dhermes: I would prefer `1024 * 1024 * 10`

lukesneeringer: Changed to that, and also did `1024 * 1024 * 9` (see musing above). Documentation updated accordingly.

lukesneeringer:
Approved by @dhermes in chat.

PR #3160, opened 2017-03-16
Raise ValueError exception if bucket name is invalid.

### LINKED ISSUE #2956: Raise warning/exception when user tries to `get_bucket(x)` if x includes forward slash
Hi there.

I made a silly mistake when trying to list blobs in a bucket, and have a proposal that could prevent users from repeating said mistake.

When initializing a `Bucket` object, I did so with a forward slash ("/") in the bucket name string. This produced a supposedly valid bucket, as seen here (actual bucket name changed for sake of demonstration):

```pycon
>>> from gcloud import storage
>>> c = storage.Client()
>>> b0 = c.get_bucket('valid-bucket')
>>> b0.exists()
True
>>> b1 = c.get_bucket('valid-bucket/')  # <-- A silly mistake
>>> b1.exists()
True
```

While both methods of initialization produce seemingly valid `Bucket` objects, initializing with a forward slash breaks further on, i.e:

```pycon
>>> len([x for x in b0.list_blobs()])
8112
>>> len([x for x in b1.list_blobs()])
  File "<stdin>", line 1, in <listcomp>
  File "/usr/local/lib/python3.5/dist-packages/gcloud/iterator.py", line 79, in __iter__
    response = self.get_next_page_response()
  File "/usr/local/lib/python3.5/dist-packages/gcloud/iterator.py", line 115, in get_next_page_response
    method='GET', path=self.path, query_params=self.get_query_params())
  File "/usr/local/lib/python3.5/dist-packages/gcloud/connection.py", line 347, in api_request
    error_info=method + ' ' + url)
gcloud.exceptions.NotFound: 404 Not Found (GET https://www.googleapis.com/storage/v1/b/valid-bucket//o?projection=noAcl)
```

`gcloud` adds a second forward slash when constructing URLs internally, and this obviously breaks. I'd suggest raising a warning or exception when users try to initialize a bucket with a forward slash, or alternatively fixing it behind-the-scenes (simply dropping the slash).

Cheers,
Eilam

daspecster:
Fixes #2956.

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -30,7 +30,8 @@ class _PropertyMixin(object):
       - path
 
     :type name: str
-    :param name: The name of the object.
+    :param name: The name of the object. Bucket names must start and end with a
+                 number or letter.
     """
 
     def __init__(self, name=None):
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -18,8 +18,26 @@ These are *not* part of the API.
 """
 
 import base64
-import re
 from hashlib import md5
+
+
+def _validate_name(name):
+    """Pre-flight ``Bucket`` name validation.
+
+    :type name: str or :data:`NoneType`
+    :param name: Proposed bucket name.
+
+    :rtype: str or :data:`NoneType`
+    :returns: ``name`` if valid.
+    """
+    if name is None:
+        return
+
+    # The first and las characters must be alphanumeric.
+    if not all([name[0].isalnum(), name[-1].isalnum()]):
+        raise ValueError(
+            'Bucket names must start and end with a number or letter.')
+    return name
 
 
 class _PropertyMixin(object):
@@ -35,12 +53,7 @@ class _PropertyMixin(object):
     """
 
     def __init__(self, name=None):
-        if name is None or (re.match(r'\w', name[0]) and
-                            re.match(r'\w', name[-1])):
-            self.name = name
-        else:
-            raise ValueError(
-                'Bucket names must start and end with a number or letter.')
+        self.name = _validate_name(name)
         self._properties = {}
         self._changes = set()
 
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -18,6 +18,7 @@ These are *not* part of the API.
 """
 
 import base64
+import re
 from hashlib import md5
 
 
@@ -33,7 +34,12 @@ class _PropertyMixin(object):
     """
 
     def __init__(self, name=None):
-        self.name = name
+        if name is None or (re.match(r'\w', name[0]) and
+                            re.match(r'\w', name[-1])):
+            self.name = name
+        else:
+            raise ValueError(
+                'Bucket names must start and end with a number or letter.')
         self._properties = {}
         self._changes = set()
 
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -81,7 +81,8 @@ class Bucket(_PropertyMixin):
                    for the bucket (which requires a project).
 
     :type name: str
-    :param name: The name of the bucket.
+    :param name: The name of the bucket. Bucket names must start and end with a
+                 number or letter.
     """
 
     _MAX_OBJECTS_FOR_ITERATION = 256
--- a/storage/unit_tests/test__helpers.py
+++ b/storage/unit_tests/test__helpers.py
@@ -46,6 +46,19 @@ class Test_PropertyMixin(unittest.TestCase):
         mixin = self._make_one()
         self.assertRaises(NotImplementedError, lambda: mixin.client)
 
+    def test_bucket_name_value(self):
+        bucket_name = 'testing123'
+        mixin = self._make_one(name=bucket_name)
+        self.assertEqual(mixin.name, bucket_name)
+
+        bad_start_bucket_name = '/testing123'
+        with self.assertRaises(ValueError):
+            self._make_one(name=bad_start_bucket_name)
+
+        bad_end_bucket_name = 'testing123/'
+        with self.assertRaises(ValueError):
+            self._make_one(name=bad_end_bucket_name)
+
     def test_reload(self):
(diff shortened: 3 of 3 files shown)


dhermes:
@daspecster I am :-1: on this change, not really sure what the right approach is but adding this check in the constructor makes EVERY constructed bucket pay the price

storage/google/cloud/storage/_helpers.py
```diff
@@ -33,7 +34,12 @@ class _PropertyMixin(object):
     """
 
     def __init__(self, name=None):
-        self.name = name
+        if name is None or (re.match(r'\w', name[0]) and
```
daspecster: I'm 100% open to alternative methods of checking each end of the name.
My regex foo is weak.

daspecster:
@dhermes you mean every existing bucket? I don't believe there could be any existing buckets that violate this rule? I think the API would have complained.

https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2956#issuecomment-274531412

dhermes:
@daspecster No I mean the local `Bucket` objects created by users of this library

tseaver:
Quoting my follow up to #2956:

> If the API allows for `/` in the bucket name, then we shouldn't be prohibiting such names:  instead, we should be URL-escaping the bucket name when constructing the API request URL.

daspecster:
@lukesneeringer @tseaver any alternative ideas?

ISTM that object creation would be the place to check this? Maybe there's a more efficient way to do it though?

dhermes:
I think we can "fix" by just more clearly documenting the acceptable values.

tseaver:
@dhermes If the API allows embedding `/` in a bucket name, then we have to honor that, which means that we need to be URL-escaping the bucket name when creating API URLs

daspecster:
@tseaver a user might be confused when then see their bucket names listed from some other library that may/maynot handle the urlencoding/urldecoding right?

@dhermes sure, but the error that is returned is a 404 which doesn't explain what the allowed values are.
I'll update the docs to be more clear about the allowable name values since that should be in here anyway.

lukesneeringer:
@tseaver, you are making an orthogonal point; in actuality, slashes _are not_ permissible at the start or end of a bucket name (by the API), but the client library allows them to pass through.

tseaver:
@lukesneeringer the OP in #2956 reported having created a bucket with a *trailing* slash.

storage/google/cloud/storage/_helpers.py
lukesneeringer: The rules that are worth checking are probably these (being looser than the API is on the last one):

> Bucket names must contain only lowercase letters, numbers, dashes (-), underscores (_), and dots (.). Names containing dots require verification.
> Bucket names must start and end with a number or letter.
> Bucket names must contain 3 to 63 characters. Names containing dots can contain up to 222 characters, but each dot-separated component can be no longer than 63 characters.

A regex that will get close enough on this (e.g. catch common issues but not fail on every single invalid name) would be:

```python
re.match(r'^[a-zA-Z0-9][a-zA-Z0-9_.-]{1,220}[a-zA-Z0-9]$', name)
```

lukesneeringer:
@tseaver Yes, _in our library_, and then the API gave him an error.

lukesneeringer:
> @daspecster I am :-1: on this change, not really sure what the right approach is but adding this check in the constructor makes EVERY constructed bucket pay the price

That seems fine to me. It is not terribly expensive.

dhermes:
My issue is just that RegEx is expensive. In terms of this **specific** approach, we don't even need a regex:

```python
name[0].isalnum()
name[-1].isalnum()
```

Though this is a little strange with `unicode` (2) / `str` (3)

```python
>>> u'\xff'.isalnum()
True
```

lukesneeringer:
Sold. Change to that, and add `len(name) <= 222`.

lukesneeringer:
@daspecter Suggestion:

```python
class _PropertyMixin(object):
   def __init__(self, name=None):
        if name:
            self._validate_name(name)
        [...]

    def _validate_name(self, name):
        # Names must start and end with a letter or number.
        if not name[0].isalnum() or not name[-1].isalnum():
            raise ValueError('Bucket names must start and end with an alphanumeric character.')

        # Names must be between 3 and 222 total characters in length.
        if len(name) < 3:
            raise ValueError('Bucket names must be at least 3 characters.')
        if len(name) > 222:
            raise ValueError('Bucket names can not exceed 222 characters.')

        # Each bucket name component can not exceed 63 characters.
        if any([len(i) > 63 for i in name.split('.')]):
            raise ValueError('Each dot-separated component in a bucket name '
                             'can not exceed 63 characters.')
```

daspecster:
@lukesneeringer where are you seeing the 63 char limit in the docs?

lukesneeringer:
> Bucket names must contain 3 to 63 characters. Names containing dots can contain up to 222 characters, but each dot-separated component can be no longer than 63 characters.

daspecster:
@lukesneeringer found the link. Thanks!

dhermes:
There is just so much complexity here. Why can't we just punt to the server validation and document the rules in our docstring?

daspecster:
To echo @dhermes, there are more requirements for the names as well and I assume they're subject to change(with or without bumping the API version).

```
Your bucket names must meet the following requirements:

Bucket names must contain only lowercase letters, numbers, dashes (-), underscores (_), and dots (.). Names containing dots require verification.
Bucket names must start and end with a number or letter.
Bucket names must contain 3 to 63 characters. Names containing dots can contain up to 222 characters, but each dot-separated component can be no longer than 63 characters.
Bucket names cannot be represented as an IP address in dotted-decimal notation (for example, 192.168.5.4).
Bucket names cannot begin with the "goog" prefix.
Bucket names cannot contain "google" or close misspellings of "google".
Also, for DNS compliance and future compatibility, you should not use underscores (_) or have a period adjacent to another period or dash. For example, ".." or "-." or ".-" are not valid in DNS names.
```

lukesneeringer:
I guess what I was thinking was, "catch the easy ones, let the API fail on the rest".
The issue with the forward slash in particular is that it gets _stripped_ by the API, which leads to non-obvious errors later. So, at minimum, we should catch the alphanumeric `0` and `-1` indices.

I could go either way on the rest.

daspecster:
@lukesneeringer did you have anything else for this?

PR #3146, opened 2017-03-15
CI Rehash

lukesneeringer:
This is a continuation of #3135. It is moved to `google-cloud-python:ci` (from my personal branch) because system tests do not run from personal branches.

-------------------------------------------------

This is an attempt to fix the CI situation on this repository. In particular, this will hit the following goals:

  * ~Move to parallel builds on a per-API basis.~ Superceded by not running unnecessary APIs.
  * Move system tests into the individual API packages.
  * Eliminate Travis, move CircleCI to 2.0

In doing this, I made several changes:

  * Move from tox to nox (Python-based configuration is really useful for the multi-project situation)
  * Move from pylint to flake8 (pylint does not work on Python 3.6, has multiple difficult-to-silence false positives)

**This is not done.** In particular, several things need to be put back into place:

  * [x] Automatic docs build.
      * Note to self: Move docs into individual packages (not part of this PR).
  * [x] Automatic PyPI releasing on appropriate tags.
  * [x] Only testing APIs that have changed files.

PR #3163, opened 2017-03-17
Restoring datastore Gax exception re-mapping.

### LINKED ISSUE #2746: Port Datastore onto GAPIC

dhermes:
Fixes #2746.

See [previous source][1] for reference.

@lukesneeringer I don't really like doing this, but I'm not sure if there's a better way to remap the `GaxError`-s to our `google.cloud` exceptions.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/datastore-0.23.0/datastore/google/cloud/datastore/_gax.py#L70

### DIFF
--- a/datastore/unit_tests/test__gax.py
+++ b/datastore/unit_tests/test__gax.py
@@ -124,6 +124,69 @@ class TestGAPICDatastoreAPI(unittest.TestCase):
     def _make_one(self, *args, **kwargs):
         return self._get_target_class()(*args, **kwargs)
 
+    def test_lookup(self):
+        from google.cloud.gapic.datastore.v1 import datastore_client
+
+        patch1 = mock.patch.object(
+            datastore_client.DatastoreClient, '__init__',
+            return_value=None)
+        patch2 = mock.patch.object(datastore_client.DatastoreClient, 'lookup')
+        patch3 = mock.patch(
+            'google.cloud.datastore._gax._grpc_catch_rendezvous')
+
+        with patch1 as mock_constructor:
+            ds_api = self._make_one()
+            mock_constructor.assert_called_once_with()
+            with patch2 as mock_lookup:
+                with patch3 as mock_catch_rendezvous:
+                    mock_catch_rendezvous.assert_not_called()
+                    ds_api.lookup(None, True, bb='cc')
+                    mock_lookup.assert_called_once_with(None, True, bb='cc')
+                    mock_catch_rendezvous.assert_called_once_with()
+
+    def test_run_query(self):
+        from google.cloud.gapic.datastore.v1 import datastore_client
+
+        patch1 = mock.patch.object(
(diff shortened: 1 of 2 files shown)


lukesneeringer:
> @lukesneeringer I don't really like doing this, but I'm not sure if there's a better way to remap the GaxError-s to our google.cloud exceptions.

It is fine for now. I would personally prefer rethink `GaxError` more generally, but that is a bigger project.

dhermes:
:+1: Worth filing an issue to track this? DERP on not running lint locally

datastore/unit_tests/test__gax.py
```diff
@@ -124,6 +124,69 @@ def _get_target_class():
...
+            datastore_client.DatastoreClient, '__init__',
+            return_value=None)
+        patch2 = mock.patch.object(datastore_client.DatastoreClient, 'lookup')
+        patch3 = mock.patch(
+            'google.cloud.datastore._gax._grpc_catch_rendezvous')
+
+        with patch1 as mock_constructor:
+            ds_api = self._make_one()
+            mock_constructor.assert_called_once_with()
+            with patch2 as mock_lookup:
```
daspecster: you could move this a level up in indentation right? Unless I'm missing something.

daspecster: Or maybe the decorators on the test methods would be clearer in this case?

dhermes: Not sure what you mean. If I dedent it's invalid Python. If you meant that I should start the `with` block **before** called `_make_one()`, that is also a no-go. I did this on purpose, to have "minimal mocks" at that call site.

dhermes: @daspecster Using decorators won't work since I'm mock patching an object that gets imported at run time. `multiple` **kind of** works (maybe?) but that's really just for cases where you would use `patch(..., new='FOO')` and I don't have any `new` object to patch there (i.e. I just want a mock).

lukesneeringer:
> :+1: Worth filing an issue to track this?
Nah.

>  DERP on not running lint locally

My changes in #3146 should make that much easier in the workflow, since lint is moved into the individual packages.

dhermes:
@lukesneeringer Naw it was just a DERP by me. `tox -e lint` is plenty fast when you set `GOOGLE_CLOUD_TESTING_BRANCH` and `GOOGLE_CLOUD_TESTING_REMOTE` (this PR only changed 2 files so it was very quick)

lukesneeringer:
Sure, but because I work inside packages, I often forget to `cd ..` and run lint and system tests separately.

PR #3164, opened 2017-03-17
Removing unused code-path for remapping raw gRPC exceptions in datastore.

dhermes:
Follow-up to #2746.


tseaver:
@dhermes Do we know for certain that the GAPIC layer can no longer raise `GrpcRendezvous` errors?

lukesneeringer:
Related: https://github.com/googleapis/gax-python/pull/168

lukesneeringer:
> @dhermes Do we know for certain that the GAPIC layer can no longer raise GrpcRendezvous errors?

If we find a spot where it still does, file a bug in GAX and I will promptly fix it.

Issue #3124, opened 2017-03-09
the api of detect_full_text() return incorrect results when the texts is chinese characters 


HopLiu:
now  i use   detect_full_text() to detect words in sdk way.  there are two issues:
1.  the detect_full_text cannot return correct language code and words info when the texts in picture is chinese. 
2. meanwhile, the detect_text() can return correct info(words and bounds info).

since i need the info of each words and bounds, i have to use the first api. Any ideas or  is this an exist issue of this api?

lukesneeringer:
Hi @HopLiu,
Thank you for reporting this issue.

I do not _think_ the client library does anything differently with regard to character encodings between `detect_text` and `detect_full_text`, although I intend to double check.

Could you provide me with a reproduction case? (Basically, what is an image I can use to observe this behavior?) I expect it is a problem on the vision backend, but I would like to confirm that before passing the report on.

HopLiu:
use this  attached jpg file(8E27CE646909307503C8C5D16.jpg) to test. 
1. "python detect_chn.py fulltext resources/8E27CE646909307503C8C5D16.jpg " 
2."python detect_chn.py text resources/8E27CE646909307503C8C5D16.jpg" 
the detect_chn.py is the same as detect.py in "python-docs-samples-master/vision/cloud-client"  except i add a little encodings in order to print  on the console.
![8e27ce646909307503c8c5d16](https://cloud.githubusercontent.com/assets/653755/23780081/95e4f69e-057f-11e7-8350-d34184066831.jpg)
<img width="584" alt="qq20170310-diff" src="https://cloud.githubusercontent.com/assets/653755/23780106/c61ea170-057f-11e7-9789-ea6bcb020083.png">

daspecster:
Thanks @HopLiu! I can test this out real quick @lukesneeringer if you want.

lukesneeringer:
@daspecster Sure. I have faith in the reproduction case, what I really want to know is whether it is our bug or the backend API's bug.

daspecster:
@lukesneeringer so I'm still looking into this but for `detect_full_text`, under `pages`, I'm getting results like the following, but there are no properly encoded Chinese characters in the results AFAICT.

```python
symbols {
    property {
      detected_languages {
        language_code: "cy"
      }
    }
    bounding_box {
      vertices {
        x: 141
        y: 442
      }
      vertices {
        x: 156
        y: 442
      }
      vertices {
        x: 156
        y: 462
[...]

Issue #3132, opened 2017-03-10
Vision support ImageContext throughout all detection types.


daspecster:
I think the reason this wasn't supported before was because it wasn't supported on all the detection types.

Right now it's only really in [`Crop Hints`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/7ba4d7a211e19d2a3fe8641ce3443e30a34ae4da/vision/google/cloud/vision/image.py#L164).

If someone can confirm that this is supported on everything else, I can start adding it.

lukesneeringer:
Huh?

daspecster:
[This](https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.ImageContext) might explain a bit better.

I've abstracted the usage away from the user and just allowed them to pass in values that coincide with the API they're using, but my question is maybe I should expose that for all the detection types if it's used by all the detection types? But I don't know if they all use it or not?

lukesneeringer:
All this code is gone. Closing.

PR #3170, opened 2017-03-20
Add language_hints support to detect_full_text.

daspecster:
Fixes: #3124

Towards:  #3132.

Support `language_hints` argument for `detect_full_text()`.

### DIFF
--- a/vision/google/cloud/vision/image.py
+++ b/vision/google/cloud/vision/image.py
@@ -182,17 +182,28 @@ class Image(object):
         annotations = self.detect(features)
         return annotations[0].faces
 
-    def detect_full_text(self, limit=10):
+    def detect_full_text(self, language_hints=None, limit=10):
         """Detect a full document's text.
 
+        :type language_hints: list
+        :param language_hints: (Optional) A list of BCP-47 language codes. See:
+                               https://cloud.google.com/vision/docs/languages
+
         :type limit: int
-        :param limit: The number of documents to detect.
+        :param limit: (Optional) The number of documents to detect.
 
         :rtype: list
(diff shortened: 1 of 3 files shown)


vision/google/cloud/vision/image.py
```diff
@@ -182,17 +182,28 @@ def detect_faces(self, limit=10):
...
         :type limit: int
-        :param limit: The number of documents to detect.
+        :param limit: (Optional) The number of documents to detect.
 
         :rtype: list
         :returns: List of :class:`~google.cloud.vision.text.TextAnnotation`.
         """
-        features = [Feature(FeatureTypes.DOCUMENT_TEXT_DETECTION, limit)]
-        annotations = self.detect(features)
+        feature_type = image_annotator_pb2.Feature.DOCUMENT_TEXT_DETECTION
```
lukesneeringer: Why did this segment change so drastically?

daspecster: I was using `google.cloud.vision.feature.Feature` and `image_annotator_pb2.AnnotateImageRequest` requires `image_annotator_pb2.Feature`. Moving forward I think we should use the protos more.

lukesneeringer: Okay.

dhermes:
It appears this has broken [the build][1], but it ACTUALLY appears that there is a broken version of [`grpcio==1.20`][2] (published today)?

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1291
[2]: https://pypi.python.org/pypi/grpcio/1.2.0

daspecster:
:(

PR #3171, opened 2017-03-20
Add BigQuery system test for nested-nested records.

### LINKED ISSUE #2951: BigQuery `insert_data()` fails with record type columns (nested objects)
BigQuery [tabledata.insertAll](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll) supports inserting rows with columns that take record types (nested objects).

A customer reports that

> My table structure has nested schemas. The [[google-cloud-python] docs](https://googlecloudplatform.github.io/google-cloud-python/stable/bigquery-table.html#google.cloud.bigquery.table.Table.insert_data) don't say how to handle nested structures. All we get is, "rows (list of tuples) – Row data to be inserted. Each tuple should contain data for each schema field on the current table and in the same order as the schema fields." A list of tuples. That's it. If I have nested data do I make sub-tuples? Is it simply not supported at this time? 
> 
> I dug into the BQ code to find out what was going on. BQ does, 
> 
> zip(table._schema, rows) 
> 
> which looks like it won't support nested inserts (or could -- with sub-tuples, but who knows). 
> 

Example in the API explorer, with the following table schema:

image

Try it out in the [API Explorer for bigquery.tabledata.insertAll](https://developers.google.com/apis-explorer/#p/bigquery/v2/bigquery.tabledata.insertAll). Enter the following for the request body (after creating a table with the same schema). You need to switch the editor to free-form mode

```
{
  "rows": 
  [
    {
      "insert_id": "12345",
      "json": 
      {
        "string_col": "Some value",
        "record_col": {
          "nested_string": "another string value",
          "nested_repeated": ["0", "1", "2"],
          "nested_record": {
            "nested_nested_string": "some deep insight"
          }
        }
      }
    }
  ]
}
```

I can verify that this works with a SQL query.

```
#standardSql
SELECT * FROM `swast-scratch.hello_world.nested` LIMIT 1000
```

image

It is not clear from the docs how to do a similar call from our client libraries (and it likely isn't possible).

tseaver:
Closes #2951.

### DIFF
--- a/system_tests/bigquery.py
+++ b/system_tests/bigquery.py
@@ -664,6 +664,45 @@ class TestBigQuery(unittest.TestCase):
         table = dataset.table(TABLE_NAME)
         self._fetch_single_page(table)
 
+    def test_insert_nested_nested(self):
+        # See #2951
+        SF = bigquery.SchemaField
+        schema = [
+            SF('string_col', 'STRING', mode='NULLABLE'),
+            SF('record_col', 'RECORD', mode='NULLABLE', fields=[
+                SF('nested_string', 'STRING', mode='NULLABLE'),
+                SF('nested_repeated', 'INTEGER', mode='REPEATED'),
+                SF('nested_record', 'RECORD', mode='NULLABLE', fields=[
+                    SF('nested_nested_string', 'STRING', mode='NULLABLE'),
+                ]),
+            ]),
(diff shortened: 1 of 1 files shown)


tseaver:
Lint fixed locally, merging.

PR #3173, opened 2017-03-20
Add system tests for all scalar query parameter types. 

tseaver:
Toward #3029.

Fix two issues found by the new tests:

- `BYTES` values must *not* have the trailing newline introduced by the stdlib's "legacy" `base64.encodestring` API.
- `DATETIME` values must *not* have a trailing `Z`:  they are conceptually "naive" / zoneless.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -21,10 +21,11 @@ import datetime
 from google.cloud._helpers import UTC
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
-from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud._helpers import _RFC3339_NO_FRACTION
 from google.cloud._helpers import _time_from_iso8601_time_naive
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -21,11 +21,12 @@
 from google.cloud._helpers import UTC
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
-from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud._helpers import _RFC3339_NO_FRACTION
 from google.cloud._helpers import _time_from_iso8601_time_naive
 from google.cloud._helpers import _to_bytes
 
+_RFC3339_MICROS_NO_ZULU = '%Y-%m-%dT%H:%M:%S.%f'
```
daspecster: Should this go in `google.cloud._helpers` next to `_RFC3339_NO_FRACTION`?

tseaver: For now, only bigquery uses this format, and we've gotten a lot more conservative about changes to core, so I think I'll hold off.

PR #3177, opened 2017-03-21
Fixes #3150 - Ignores resource key when it isn't present

### LINKED ISSUE #3150: blob.py is unable to handle a partial rewrite and fails with KeyError on lack of resource
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py#L871
```
871:        self._set_properties(api_response['resource'])
```

This relies on the API to always return 'resource'.  However, it is optional as per the doc.
https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite
"""
This property is present in the response only when copying completes.
"""

As such, when the write was partial, done: false is returned and resource not present.  Thus failing with the following KeyError.
```
File "xxx.py", line 304, in test_011_CopyFileFromBucketToBucketNewName 
new_destination_file_name=new_file_name) 
File "xxx.py, line 276, in copyFileFromBucketToBucket 
token, rewritten, total = destination_blob.rewrite(source_blob, token=token) 
File "/home/jenkins/venv/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 681, in rewrite 
self._set_properties(api_response['resource']) 
KeyError: 'resource' 
```

The most easy fix would be to push _set_properties down under after done: false check:

```
E.g.;

       api_response = client._connection.api_request(
            method='POST', path=source.path + '/rewriteTo' + self.path,
            query_params=query_params, data=self._properties, headers=headers,
            _target_object=self)
        # self._set_properties(api_response['resource'])  ## Comment out this line
        rewritten = int(api_response['totalBytesRewritten'])
        size = int(api_response['objectSize'])

        if api_response['done']:
            return None, rewritten, size

        self._set_properties(api_response['resource']) # and move here 
        return api_response['rewriteToken'], rewritten, size
```

lucmult:
- The new test fails with the KeyError experienced by customer.
- Tests passing on Python 2.7 and 3.4 (I don't have 3.5 on my env).

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -870,13 +870,12 @@ class Blob(_PropertyMixin):
             _target_object=self)
         rewritten = int(api_response['totalBytesRewritten'])
         size = int(api_response['objectSize'])
-        try:
-            self._set_properties(api_response['resource'])
-        except KeyError:
-            # API might not return resource key
-            pass
 
+        # The resource key is set if and only if the API response is
+        # completely done. Additionally, there is no rewrite token to return
+        # in this case.
         if api_response['done']:
+            self._set_properties(api_response['resource'])
             return None, rewritten, size
 
         return api_response['rewriteToken'], rewritten, size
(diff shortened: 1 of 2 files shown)


lukesneeringer:
I just confirmed that and made the change I suggested, now tossing to @dhermes for review.

CI will still fail until #3178 is merged, so it will have to go in before this can.

lukesneeringer:
I merged in master (which includes #3178), merging once CI passes.

PR #3183, opened 2017-03-22
Make ACL methods chainable.

### LINKED ISSUE #3181: Chaining ACL grants on Storage does not work.
So from what I can tell [this](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/ce6756fbe3633c74fd742567654565147628f4ba/storage/google/cloud/storage/acl.py#L56) doesn't work.

Example:
```python
acl.all().grant_read().revoke_write()
```

None of the grant methods return an object to chain to.
See:  https://github.com/GoogleCloudPlatform/google-cloud-python/blob/486163e5378897961bc3cf367062c8b10cf7ce3e/storage/google/cloud/storage/acl.py#L148-L170

Should I just take this out of the docs?

lukesneeringer:
Fixes #3181

### DIFF
--- a/storage/google/cloud/storage/acl.py
+++ b/storage/google/cloud/storage/acl.py
@@ -133,6 +133,9 @@ class _ACLEntity(object):
 
         :type role: str
         :param role: The role to add to the entity.
+
+        :rtype: ~`google.cloud.storage.acl._ACLEntity`
+        :returns: self
         """
         self.roles.add(role)
         return self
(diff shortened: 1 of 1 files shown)


daspecster:
@lukesneeringer I think we should have tests for this.

daspecster:
@lukesneeringer also note @tseaver's [comment](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3181#issuecomment-288461359).

lukesneeringer:
Agreed on testing. Indifferent on whether we do it.

dhermes:
I am :-1: on method chaining. So is @tseaver and so is @jonparrott (I think)

jonparrott:
I'm generally against method chaining as well, as the return type of a
function should be meaningful and not just self.

lukesneeringer:
Okay, in that case, closing this, and @daspecster please correct the errant docs.

PR #3182, opened 2017-03-22
Fix packaging of resources and tests

jonparrott:
* Don't include tests in the binary (wheel) distribution.
* Include tests in the source distribution.
* Include only .json and .proto package resources.
* Include a LICENSE with every package (extremely important for downstream packagers).

PR #3184, opened 2017-03-22
Fix packaging of resources and tests

jonparrott:
* Don't include tests in the binary (wheel) distribution.
* Include tests in the source distribution.
* Include only .json and .proto package resources.
* Include a LICENSE with every package (extremely important for downstream packagers).

(From #3182)

### DIFF
--- a/bigquery/LICENSE
+++ b/bigquery/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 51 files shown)


jonparrott:
Broken into multiple commits for the sake of review, but should be squashed during merge.

All "Fix * package" commits are identical, just targeting a different package.

lukesneeringer:
This is still against master, as far as I can tell.

jonparrott:
Sigh.

PR #3187, opened 2017-03-22
Logging docs, update references that were 301 redirecting.

daspecster:
These links were 301 redirecting.

/cc @waprin (I know you're OOO, just wanted this in your inbox for later.)


daspecster:
@dhermes yeah I copy and paste the old ones and see where they redirect too.

Issue #1679, opened 2016-03-30
Add IAM support to APIs which support it


tseaver:
(copied from https://github.com/GoogleCloudPlatform/gcloud-node/issues/1192, which is funnily-enough exactly true for `gcloud-python`, too).

[IAM](https://cloud.google.com/iam/docs) is currently buried inside of our Pub/Sub code, but should be abstracted to be used with any API that supports it:

- [ ] BigQuery
- [ ] [Compute](https://cloud.google.com/compute/docs/access/iam)
- [ ] [Logging](https://cloud.google.com/logging/docs/access-control)
- [ ] [Pub/Sub](https://cloud.google.com/pubsub/access_control)
- [ ] [Resource Manager](https://cloud.google.com/resource-manager/docs/access-control-proj)
- [ ] [Storage](https://cloud.google.com/storage/docs/iam)

kir-titievsky:
A gentle ping on this, given that IAM is now GA: is there an AIM API coming?  Any chance the Pub/Sub AIM API might be brought up to date.

ncouture:
@kir-titievsky PR #2031 seems to be related but has not seen updates for a few months.

tseaver:
@lukesneeringer How would you prefer that I factor out the existing `google.cloud.pubsub` IAM support?

- Shared bits move into a new `google.clooud.iam` module inside the `core/` subproject?
- Shared bits move into a new `iam/` subproject?

lukesneeringer:
Good question. My instinct is that I would prefer not have an excessive number of sub-projects, and to put it in a `google.cloud.iam` module in core. My feeling is that if we later have six or seven projects like that, it will become hard to know what is an API and what is a shared module.

That is a loose preference though and I am happy to be persuaded otherwise.

(If this were day 1, I would have preferred it have been `google.cloud.core.*` namespaced.)

tseaver:
I'm find with putting it in `core/`:  I was just worried that we wanted to be more conservative about such changes.

dhermes:
What about putting it all in `google-auth-library-python`? @jonparrott WDYT?

jonparrott:
@dhermes I'd prefer to keep code that hits any non-auth API endpoint outside of google-auth (I was actually *really* wary of adding `IAMSigner`). But I may be able to be convinced if you think it's the best place for this.

jonparrott:
gax/core seems like the right spot for this IMO.

tseaver:
[...]

PR #2031, opened 2016-07-27
Proposal: IAM Meta Interface

elibixby:
Definition of an IAM meta interface to be used for resources which have implemented the IAM meta api

PR #3188, opened 2017-03-22
Factor common IAM policy bits into 'google.cloud.iam'.

tseaver:
Toward #1679.

Pubsub-specific roles, permissions left behind in `google.cloud.pubsub.iam`.

`google.cloud.pubsub.iam.Policy` subclasses the core one, extending it to deal with the pubsub-specific roles.


core/google/cloud/iam.py
```diff
@@ -0,0 +1,217 @@
+# Copyright 2016 Google Inc.
```
dhermes: 2017

elibixby:
I had a previous proposal to do this which got abandoned due to slow review cycle, but it's here: See #2031  

The key takeaway: the API representation of bindings is really gross to interact with. Client side it would be much nicer if bindings were a dict with roles as keys, and members as values. So instead of code like:

```
roles = [binding['role'] for binding in policy.bindings]
if my_role in roles:
   members = policy.bindings[roles.index(my_role)]
   members.append(my_member)
else:
   policy.bindings.append({'role': my_role, 'members': [my_member]})
```

We had code like:
```
members = policy.bindings.setdefault(role, []).append(my_member)
```

core/google/cloud/iam.py
```diff
@@ -0,0 +1,217 @@
...
+    @staticmethod
+    def authenticated_users():
+        """Factory method for a member representing all authenticated users.
+
+        :rtype: str
+        :returns: A member string representing all authenticated users.
+        """
+        return 'allAuthenticatedUsers'
+
+    def _bind_custom_role(self, role, members):
```
elibixby: This can't be private... user (not just API) defined custom roles are coming VERY soon, which will mean all subsclasses need to support allowing ANY rolestring.

core/google/cloud/iam.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, etag=None, version=None):
+        self.etag = etag
+        self.version = version
+        self.owners = set()
+        self.editors = set()
+        self.viewers = set()
+
+    @staticmethod
+    def user(email):
```
elibixby: These don't belong on the policy object. They are common to every IAM object.

tseaver: @elibixby You only need them within the context of defining a policy and its bindins.

core/google/cloud/iam.py
```diff
@@ -0,0 +1,217 @@
...
+    """Roles mapped onto our ``editors`` attribute."""
+
+    _VIEWER_ROLES = (VIEWER_ROLE,)
+    """Roles mapped onto our ``viewers`` attribute."""
+
+
+    def __init__(self, etag=None, version=None):
+        self.etag = etag
+        self.version = version
+        self.owners = set()
```
elibixby: Explicitly calling out individual roles on a policy doesn't make sense. These are "legacy roles" instead the policy should have a `policy.bindings` property which should be a dict from `role` to  member list. See my previous proposal.

elibixby:
Additionally note that Roles are an API resource, and should be represented with an object. What does the `resource.queryGrantableRoles` method return? Raw JSON? Role resources will be user managed with the additoin of custom roles, and will have a list of iam permissions that they grant, which will be important for user protection of custom methods, and for identifying the necessary roles that a member needs to be granted.  (combinding resource.testIAMPermissions and resource.queryGrantableRoles)

Please thoroughly review the design choices made in #2031

core/google/cloud/iam.py
elibixby: I disagree that this justifying putting them as properties on the object, but it's a style nit and inconsequential compared to my other suggestions...  ¯\_(ツ)_/¯

core/google/cloud/iam.py
elibixby: A suggestion that maintains backwards compatibility would be do add properties:

```
@property
def owners(self):
   return self.bindings.get(self._OWNERS_ROLE, set())

@owners.setter
def set_owners(self, values):
   return self.bindings.setdefault(self._OWNERS_ROLE, set()).add(values)
```

pubsub/google/cloud/pubsub/iam.py
```diff
@@ -107,125 +101,44 @@ class Policy(object):
...
-        self.editors = set()
-        self.viewers = set()
+        super(Policy, self).__init__(etag, version)
         self.publishers = set()
         self.subscribers = set()
 
-    @staticmethod
-    def user(email):
-        """Factory method for a user member.
+    def _bind_custom_role(self, role, members):
```
elibixby: This is going to break when custom roles role out, and is already broken for the alpha customers of custom roles. We shouldn't try to keep an accounting of role strings, they can be user defined and change at runtime. If we want to provide shortcuts for "legacy" roles like `roles/owners` then we should implement those as I suggest above

tseaver:
@elibixby I think I have addressed your concerns about supporting user-defined roles in b625017.  

re: `queryGrantableRoles`:  We can't propely target implementing support for not-yet-released APIs in the public repository:  my current writ is to get `storage` up to the same level of IAM support that `pubsub` currently has.  I will be adding `storage`-specific implementations of `get_iam_policy`, `set_iam_policy`, and `test_iam_permissions` in a separate PR.

elibixby:
@tseaver `queryGrantableRoles` isn't unreleased...https://cloud.google.com/iam/reference/rest/v1/roles/queryGrantableRoles

core/google/cloud/iam.py
```diff
@@ -154,47 +178,10 @@ def from_api_repr(cls, resource):
...
-            members = set(binding['members'])
-            if role in cls._OWNER_ROLES:
-                policy.owners |= members
-            elif role in cls._EDITOR_ROLES:
-                policy.editors |= members
-            elif role in cls._VIEWER_ROLES:
-                policy.viewers |= members
-            else:
-                policy._bind_custom_role(role, members)
+            members = sorted(binding['members'])
```
elibixby: Is there a reason you're not continuing to use `set` here? It will be an error to specify the same member twice on the API side, so a set is very helpful.

core/google/cloud/iam.py
```diff
@@ -209,9 +196,9 @@ def to_api_repr(self):
         if self.version is not None:
             resource['version'] = self.version
 
-        bindings = self._role_bindings()
-
-        if bindings:
-            resource['bindings'] = bindings
+        if len(self.bindings) > 0:
+            resource['bindings'] = [
+                {'role': role, 'members': members}
```
elibixby: Might as well check for empty members here since that is also an API error. It's not likely but users accidentally end up with empty members via use of `remove`.

core/google/cloud/iam.py
```diff
@@ -0,0 +1,204 @@
...
+        self.bindings = {}
+
+    @property
+    def owners(self):
+        """Legacy access to owner role."""
+        result = set()
+        for role in self._OWNER_ROLES:
+            for member in self.bindings.get(role, ()):
+                result.add(member)
+        return result
```
elibixby: Didn't the old implementation support syntax like:
```
policy.owners.add(member)
```
This won't work in this case. We should instead be returning an object which can be modified and is being tracked by `bindings` e.g.:
```
@property
def owners(self):
   return self.bindings.setdefault(self._OWNER_ROLE, set())
```
As a side note. What's the use case for making _OWNER_ROLE a list? Seems to me each individual role should have it's own property?

tseaver: Backward compatibility:  the current pubsub implementation accumulated both users with `roles/owner` and those with `roles/pubsub.admin` into its `owners`.

Note that it is `_OWNER_ROLES`, i.e., the roles used to populate `owners`.

elibixby: Ahhh yuk. Ok. How can the `owners` property support set modification syntax like it did previously? In fact, how did the previous version support set modification? How did it know which role to attribute an addition to 'owners' to? This seems really bad as it makes policies opaque to client library users.

As an aside. Set modification is much more idiomatic than reassignment since this API follows a 
read-modify-write pattern with the `etag`.

jonparrott: @lukesneeringer should be the deciding factor for breaking compatibility. IMO, If we're gonna do it, let's do it now before 1.0.

elibixby: Reviewing the old implementation it looks like `roles/owner` and `roles/pubsub.admin` were collated into on set `policy.owners`, but on pushing the new policy out, all of the members of `policy.owners` were added to `roles/pubsub.admin`. This is **really, really** bad. Basically this pub/sub implementation had unintended side-effects even with *no* user modification. i.e. 

```
pubsub.set_iam_policy(pubsub.get_iam_policy())
```

was not guaranteed to be idempotent!!!!!

lukesneeringer: > Reviewing the old implementation it looks like `roles/owner` and `roles/pubsub.admin` were collated into the set `policy.owners`, but on pushing the new policy out, all of the members of `policy.owners` were added to `roles/pubsub.admin`. This is **really, really** bad. Basically this pub/sub implementation had unintended side-effects even with *no* user modification. i.e. 

Could we make `policy.owners` a special `frozenset` computed by the union of `roles/owner` and `roles/pubsub.admin`?

That would maintain back-compat on _reading_ but break it (with an explicit exception) on _writing_.

elibixby: `roles/owner` and `roles/pubsub.admin` are not equivalent, and I don't see why they should be bundled into one set. I think this is a bug not a feature. `roles/owner` applies only on projects, and simply cascades down to resources in the project namespace. It can't actually be set on a pubsub policy. It's a legacy role with a much too wide set of permissions, and should be avoided if at all possible. It's much better to be explicit, and only interact with `policy.admins`. Then `owners` can be a frozenset to avoid accidental modification, which will result in an API error.

elibixby: Note `owners` will have to be overridden in the resource_manager API to be not a `frozenset` as that's the only API under which such a modification can occur.

lukesneeringer: > I think this is a bug not a feature.

That is not under debate; the question is, how do we cause the least disruption for our existing users.

> It's a legacy role with a much too wide set of permissions, and should be avoided if at all possible.

That may be, but lots of people are using it, now. We need to try to avoid or at least minimize how badly we break these people.

elibixby: >>It's a legacy role with a much too wide set of permissions, and should be avoided if at all possible.
>
>That may be, but lots of people are using it, now. We need to try to avoid or at least minimize how badly we break these people.

If we're breaking people why not break them in a way that encourages best practices? I suppose we could do something like:

```
# in iam.Policy
@property
def legacy_owners(self):
   return self.bindings.setdefault('roles/owner', set())

@legacy_owners.setter
def legacy_owners(self, values):
   assert isinstance(values, set)
   self.bindings['roles/owner'] = values

@property
def owners(self):
   return frozenset(self.legacy_owners | self.bindings.get(self._ADMIN_ROLE, set()))

@owners.setter
def owners(self, values):
  raise ValuesError("Nope")

# in pubsub.iam.Policy

_ADMIN_ROLE = 'roles/pubsub.admin'

@property
def admins(self):
  return self.bindings.set_default(self._ADMIN_ROLE, set())
```

lukesneeringer: The point I am making is: Ideally, I would prefer not break people at all. If we must break them, then I want to determine the least intrusive path.

If we must do something sweeping, then yes, we should encourage best practices. But encouraging best practices is not a sufficient reason to do something sweeping.

Personally, I am still trying to get a handle on exactly what the issue is. What I am hearing now is that Pub/Sub conflates two IAM buckets on both read and write. On read it does it by taking the union of two buckets, and on write it does it by writing to one of them.

Both of those things seem like problems. The former seems like a problem we can solve in a backwards-compatible way (which might be really important to, say, not break publishing or subscribing apps). The latter seems like a problem we have to solve by raising an error.

tseaver: Just for clarity, let me summarize the current implementation:

- When creating a `Policy` instance from the resource returned by the `getIAMPolicy` API, it conflates uses with either the "legacy" `roles/owner` or and the "real" `roles/pubsub.admin` into its `owners` set.   Likewise for editors and viewers.  The reason for this implementation is that early on (two years back) there were users who had some topics (maybe subscriptions) with the legacy roles set, but the newer ones were entirely equivalent (i.e., mapped onto the same permissions relevant for topics/subscriptioions).

- When sending the resource *back* to the `setIAMPolicy` API, the policy discards the legacy roles because,  as @elibixby points out, they are not allowed to be set for pubsub topics or subscriptions (they were not allowed to be set even then).

So, nobody has been relying on setting legacy roles since before we added IAM support to pubsub:  it was never possible at all.  Silently converting existing legacy roles to the acceptable equivalents seemed like a good tradeoff at the time.  The other choice would have been to let the `setIAMPolicy` blow up if the user didn't clean out / replace the legacy roles.

elibixby: On reviewing internal docs the Google3 source. I'm fairly confident we can simply move `roles/pubsub.admin` from the `owners` set into it's own set, and it will be backwards compatible-ish.

Right now permissions on `roles/owners` on any pubsub resource is equivalent to `roles/pubsub.admin`, though there's no guarantee it will stay this way. So if we have users who are using `owners` to add `roles/pubsub.admin` roles, they won't be broken by instead adding `roles/owners`. However we should maybe throw a warning telling them to use non-legacy roles?

This result in lots of additional IAM calls when users upgrade versions, as they readd the users they need to the `roles/owner`, (since they were previously added to `roles/pubsub.admin`).

lukesneeringer: Okay, so let me repeat this potential solution in my own words to verify that I understand it:

  * When creating an instance for `getIAMPolicy`, we should:
      * Use `roles/owner ∪ roles/pubsub.admin`
      * Issue a warning if `(roles/owner ∩ roles/pubsub.admin) != roles/owner` saying that this conflation may not be honored by the API in the future. (Or, maybe just saying "we are upgrading your data to the new format" if the below bullet is a thing.)
      * Automatically do a behind the scenes call setting everyone in `roles/owner` into `roles/pubsub.admin`.
  * `setIAMPolicy` should then change to not do any conflation at all, and only write what you tell it to write.
      * Attempting to write to `roles/owner` would be an error.

elibixby: Hmm I don't think that's what I was suggesting =/

* base `iam.Policy` object has a `owners` attribute which *only* tracks `roles/owner`.  This can be readwrite. It turns out that `roles/owner` *is* actually grantable on pubsub resources. We just recommend against it so maybe we an add a warning, but it won't break AFAICT
* `pubsub.iam.Policy` has an `admins` attribute which *only* tracks `roles/pubsub.admin`, warning on `policy.owners` suggests converting calls to `policy.owner` to calls to `policy.admins`. This is also readwrite

Under this solution, what the library returns for a given call, and does for a given call will change, but all calls will still work. The change is just accurately reflecting the API instead.

Alternatively we could do as @tseaver and simply silently convert `policy.owners` to refer exclusively to `roles/pubsub.admin`, and remove all ability to address legacy roles, which I'm also fine with.

lukesneeringer: Okay, so the advantages of @elibixby's approach is that it is a more accurate reflection of the API, but @tseaver's approach preserves back-compat.

Can we do @tseaver's approach but be loud about the conversion? In other words, if we do the conversion of `policy.owners`, also issue a `DeprecationWarning`, and then convert to what @elibixby wants immediately before GA?

elibixby: @lukesneeringer I think my approach also preserves backwards compat, in that the same client library call will result in the same *permissions* but not the same *roles*. And nothing will explode with an error. 

The advantage of @tseaver 's approach is that it automatically upgrades users to the recommended behavior, the disadvantage is it prevents manipulating legacy roles (including removing) legacy roles. I think this is actually a big problem now that I think about it. Suppose you wanted to lock someone out of a pubsub resource, but they had a `roles/owners` membership.

```
policy = topic.get_iam_policy()
policy.owners.remove(member)
topic.set_iam_policy()
```

Doesn't actually do it! That said, neither does my approach, because it has the same problem with `roles/pubsub.admin`. However, my approach does allow the possibility of correcting this problem, via:

```
policy.owners.remove(member)
policy.admins.remove(member)
```

Thoughts all?

elibixby: The only way to maintain true backwards compat *and* prevent conflating these two sets, is by providing a custom set, which `owners` and `admins` share, where adding to one or the other adds to the specific underlying role, but removing from either removes from both.

elibixby: Ok correction: I got myself confused. @tseaver approach is more backwards compat, simply because the current pubsub implementaiton doesn't actually allow removing legacy role members either. But I think this is a thing that needs to be fixed.

lukesneeringer: > Suppose you wanted to lock someone out of a pubsub resource, but they had a roles/owners membership.

I buy that this is an issue. @tseaver Thoughts?

lukesneeringer: > Ok correction: I got myself confused. @tseaver approach is more backwards compat, simply because the current pubsub implementaiton doesn't actually allow removing legacy role members either. But I think this is a thing that needs to be fixed.
  
But that is the API's restriction.

Okay, let's do what @tseaver suggested, with a honking loud warning if `roles/owners` is a non-empty set. We can additionally state that we will stop automatically pulling from `roles/owners` before GA (basically right before GA we will shift to @elibixby's approach).

elibixby: > >Ok correction: I got myself confused. @tseaver approach is more backwards compat, simply because the current pubsub implementaiton doesn't actually allow removing legacy role members either. But I think this is a thing that needs to be fixed.
>
>But that is the API's restriction.

I don't actually think this is. Legacy roles can be added and removed on pubsub resources. It's just recommended against. That said, I don't think this changes the merit of the approach suggested above.

lukesneeringer:
> EDIT @lukesneeringer I think doing @tseaver approach and then moving to my approach immediately before GA is a reasonable approach, thoughts @tseaver?

This is fine with me also.

lukesneeringer:
@tseaver Checking on the status of this. :-)

tseaver:
@lukesneeringer I'm back at this, but can't figure out what the desired resolution would be.  Here is a sketch at a way forward:

- Return `frozenset` instances from the getters for `owners`, `editors`, and `viewers` (breaking anybody who expects to mutate them in place, rather than re-assigning them).
- Emit `DeprecationWarning` from the setters for those properties, pointing people to the desired surface.

Then, the question is what the desired surface would be:  I'm really leery of exposing the mutable `bindings` attribute as a dict:  it allows us no way to do any checking at all.  We could:

- Make the `Policy` class dict-like, with roles as keys and lists of principals as values.  The `__setitem__` implementation would give us the hook to do any required checking.

- Make `Policy.bindings` a dict-like object with a similar `__setitem__`.

- Add a `Policy.bind(role, *principals)` method, and use whatever backing store we like.

lukesneeringer:
My personal preference is the first option (making `Policy` dict-like and defining `__setitem__` to include appropriate validation).

tseaver:
@lukesneeringer That is my preference as well.  Update on the way.

tseaver:
@lukesneeringer OK, I've updated `Policy` to be dict-like, and added deprecations for assignment to the "named" role attributes.

lukesneeringer:
Reviewing now.

core/google/cloud/iam.py
```diff
@@ -53,49 +55,64 @@ class Policy(object):
...
+        return iter(self._bindings)
+
+    def __len__(self):
+        return len(self._bindings)
+
+    def __getitem__(self, key):
+        return self._bindings[key]
+
+    def __setitem__(self, key, value):
+        self._bindings[key] = value
```
elibixby: Should this validate that `value` is a `set`?

lukesneeringer:
Merging this when CI goes green.

PR #3190, opened 2017-03-23
Delete duplicated constant

nownabe:
Hi! ☀️ 

I deleted duplicated constant in VisionAPI FeatureTypes.


vision/google/cloud/vision/feature.py
```diff
@@ -30,7 +30,6 @@ class FeatureTypes(object):
     LABEL_DETECTION = 'LABEL_DETECTION'
     TEXT_DETECTION = 'TEXT_DETECTION'
     SAFE_SEARCH_DETECTION = 'SAFE_SEARCH_DETECTION'
-    IMAGE_PROPERTIES = 'IMAGE_PROPERTIES'
```
dhermes: Can you also move `LABEL_DETECTION` so it is in alphabetical order?

nownabe:
ordered :+1:

dhermes:
Thanks @nownabe!

This LGTM. @daspecster do we need these hard-coded values or can you lean on protobufs more? (Not part of this PR, but this PR clearly indicates that writing this by hand is problematic.)

daspecster:
@dhermes the goal down the road is to really use the protobufs almost entirely. So that's definitely a good direction to head IMHO.

PR #3192, opened 2017-03-23
Update 301 redirects for storage documentation.

daspecster:
Documentation link updates. Fixed 404s and 301 redirects.

### DIFF
--- a/storage/google/cloud/storage/acl.py
+++ b/storage/google/cloud/storage/acl.py
@@ -196,7 +196,7 @@ class ACL(object):
         'bucketOwnerFullControl',
     ])
     """See:
-    https://cloud.google.com/storage/docs/access-control/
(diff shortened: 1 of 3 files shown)


storage/google/cloud/storage/acl.py
```diff
@@ -196,7 +196,7 @@ class ACL(object):
         'bucketOwnerFullControl',
     ])
     """See:
-    https://cloud.google.com/storage/docs/access-control#predefined-acl
+    https://cloud.google.com/storage/docs/access-control/
```
dhermes: 1. This would not be a 301?
1. It seems like this content has moved: https://cloud.google.com/storage/docs/access-control/lists#predefined-acl

daspecster: Ah thanks! The tool I'm using shows broken anchors as well so that's why it came up. My fault for not finding the alternative link.

PR #3194, opened 2017-03-23
Reduce AppVeyor testing matrix.

lukesneeringer:
Right now, we run the unit tests on AppVeyor six times:

  * Python 2.7, 3.4, and 3.5
  * 32-bit and 64-bit.

This takes well over **90 minutes** to complete. The results:

  * We tend not to wait on AppVeyor to actually merge (meaning it loses its value).
  * Extremely obnoxious queue times.

Therefore, let's test on 32-bit 2.7 and 64-bit 3.5 (which will generally take about 30 minutes) and call it soup.

I am making this change on the `ci` branch also, for the same reasons, although the `ci` branch is still in mysterious AppVeyor hell.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,10 @@ environment:
     # a later point release.
     # See: http://www.appveyor.com/docs/installed-software#python
 
+    # We test on 32-bit Python 2.7, and 64-bit Python 3.x.
+    # This is because it takes too long to serially go through six to eight
+    # iterations.
+
     # Python 2.7.11 is the latest Python 2.7 with a Windows installer
     # Python 2.7.11 is the overall latest
     # https://www.python.org/ftp/python/2.7.11/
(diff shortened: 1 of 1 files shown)


dhermes:
@lukesneeringer AppVeyor doesn't build on PRs. It only builds for commits to branches, so-called "push builds". It's just there in the background and I only ever think about it if I get an email that says "build failed".

Seems to be "working as intended"?

lukesneeringer:
That does not really change the principle. It is excessively obnoxious right now because every merge to master means a two hour wait before I can thrash CI against AppVeyor _at all_, and there is no way to cancel a build that I can see.

But really, running _unit_ (not system) tests against every possible combination in an OS-specific environment is pretty needless.

We should either run a long gamut nightly or a sane gamut per-push. This moves toward the latter.

dhermes:
@lukesneeringer OK, LGTM.

Issue #3165, opened 2017-03-18
"Client GRPC deadline too short" for speech streaming_recognize + cannot override defaults


cgratie:
There is currently no way to override default configuration parameter values when using **google.cloud.speech.Client**. Support is available in **google.cloud.gapic.speech.v1beta1.speech_client.SpeechClient** via the **client_config** constructor argument. However, this is not used in **google.cloud.speech._gax.GAPICSpeechAPI**, where a **SpeechClient** instance is constructed without making use of **client_config**.

This means, among other things, that I cannot set the GRPC deadline myself. On the other hand, the default value for timeout present in the configuration file as **"timeout_millis"** under **"StreamingRecognize"** is **190000**, which should be perfect for a call that needs to be shorter than 1 minute anyway (required deadline is 3 * duration + 5, so 185 should already be enough). However, it looks like the actual value that is sent as a deadline comes from **"initial_rpc_timeout_millis"** field from **"retry_params"** in the configuration file (I have actually changed that value to 185000 and the error disappeared). I'm not sure why this is the case and I could not figure out where it happens in the code.

I am experiencing the issue while using the following versions of Python packages:
* gapic-google-cloud-speech-v1beta1 (0.15.3)
* google-auth (0.8.0)
* google-auth-httplib2 (0.0.2)
* google-cloud-core (0.23.1)
* google-cloud-speech (0.23.0)
* google-gax (0.15.8)
* googleapis-common-protos (1.5.2)
* grpcio (1.1.3)

daspecster:
@cgratie thanks for reporting!

The limit for streaming detection [is 1 minute](https://cloud.google.com/speech/limits).
However if the GAPIC code isn't using the `timeout_millis` argument there may be a bug there.

I will look into that and also see about passing that config to the GAPIC API constructor.

cgratie:
So the code I am running it the following.
```
    client = speech.Client(use_gax=True)
    stream = open("sample.raw", "rb")
    sample = client.sample(
        stream=stream, encoding=speech.Encoding.LINEAR16, sample_rate=16000)
    responses = client.speech_api.streaming_recognize(
        sample, language_code="en-GB", max_alternatives=30, interim_results=True)
```
The audio "sample.raw" is 25 seconds long. The error I get when running without changing defaults is:
```
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 367, in next
    return self._next()
[...]

PR #3204, opened 2017-03-24
Speech: Expose the GAPIC client config on the manual layer.

lukesneeringer:
This PR is a response to #3165 but I somewhat doubt we should take it exactly as-is. This is meant to be a discussion starter.

#### Context

Definitely go read #3165 to understand the whole thing, but the basic situation is that the OP had a problem with the Speech API that he got around by overriding the GAPIC `client_config`. He did this by modifying the actual JSON file in the installed GAPIC, which probably is something we should not force our users to do.

#### Solution

The first and most obvious thing we should do is set a more sane default for this one configuration option in Speech. I am going to do that momentarily.

Additionally, I assert that we should expose the `client_config` as an option in our `Client` classes, so that our users _do_ have the option to override these things.

This PR makes this change only for Speech, which is almost certainly the wrong final solution. Either we should do this in every API that depends on a GAPIC or we should do it in none of them. I wanted to put this in place so we can see what the solution looks like and discuss whether it should be done everywhere or nowhere. (If we do it everywhere, then the `self._client_config` part in Speech's `Client` class would migrate into core.)

#### Discussion

Should we do this?

### DIFF
--- a/speech/google/cloud/speech/_gax.py
+++ b/speech/google/cloud/speech/_gax.py
@@ -43,8 +43,14 @@ class GAPICSpeechAPI(object):
 
     :type client: `~google.cloud.core.client.Client`
     :param client: Instance of ``Client`.
+
+    :type client_config: dict
+    :param client_config: A dictionary for call options for each method. See
+                :func:`google.gax.construct_settings` for the structure of
+                this data. Falls back to the default config if not specified
+                or the specified config is missing data points.
     """
-    def __init__(self, client=None):
+    def __init__(self, client=None, client_config=None):
         self._client = client
         credentials = self._client._credentials
         channel = make_secure_channel(
@@ -52,6 +58,7 @@ class GAPICSpeechAPI(object):
             SpeechClient.SERVICE_ADDRESS)
         self._gapic_api = SpeechClient(
             channel=channel,
+            client_config=client_config,
             lib_name='gccl',
             lib_version=__version__,
         )
--- a/speech/google/cloud/speech/client.py
+++ b/speech/google/cloud/speech/client.py
@@ -47,7 +47,14 @@ class Client(BaseClient):
     :param use_gax: (Optional) Explicitly specifies whether
                     to use the gRPC transport (via GAX) or HTTP. If unset,
                     falls back to the ``GOOGLE_CLOUD_DISABLE_GRPC`` environment
-                    variable
+                    variable.
+
+    :type client_config: dict
+    :param client_config: A dictionary for call options for each method. See
+                :func:`google.gax.construct_settings` for the structure of
+                this data. Falls back to the default config if not specified
+                or the specified config is missing data points.
+                Only used with the gRPC transport.
     """
 
     SCOPE = ('https://www.googleapis.com/auth/cloud-platform',)
@@ -55,8 +62,11 @@ class Client(BaseClient):
 
     _speech_api = None
 
-    def __init__(self, credentials=None, http=None, use_gax=None):
+    def __init__(self, credentials=None, http=None, use_gax=None,
+                 client_config=None):
         super(Client, self).__init__(credentials=credentials, http=http)
+        self._client_config = client_config
+
         # Save on the actual client class whether we use GAX or not.
         if use_gax is None:
             self._use_gax = _USE_GAX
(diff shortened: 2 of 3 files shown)


speech/google/cloud/speech/_gax.py
```diff
@@ -43,15 +43,22 @@ class GAPICSpeechAPI(object):
 
     :type client: `~google.cloud.core.client.Client`
     :param client: Instance of ``Client`.
+
+    :type client_config: dict
+    :param client_config: A dictionary for call options for each method. See
+                :func:`google.gax.construct_settings` for the structure of
+                this data. Falls back to the default config if not specified
+                or the specified config is missing data points.
```
dhermes: @lukesneeringer It's been our practice to either align text or indent all of it. So

```python
"""
    :param client_config: A dictionary for call options for each method. See
                          :func:`google.gax.construct_settings` for the
                          structure of this data. Falls back to the default
                          config if not specified or the specified config is
                          missing data points.
"""
```

or

```python
"""
    :param client_config:
        A dictionary for call options for each method. See
        :func:`google.gax.construct_settings` for the structure of
        this data. Falls back to the default config if not specified
        or the specified config is missing data points.
"""
```

However we don't have any lint rule enforcing it, so anything goes?

speech/google/cloud/speech/_gax.py
```diff
@@ -43,15 +43,22 @@ class GAPICSpeechAPI(object):
 
     :type client: `~google.cloud.core.client.Client`
     :param client: Instance of ``Client`.
+
+    :type client_config: dict
+    :param client_config: A dictionary for call options for each method. See
```
dhermes: Note that it's optional? `(Optional) A dictionary...`

speech/google/cloud/speech/client.py
```diff
@@ -47,16 +47,26 @@ class Client(BaseClient):
     :param use_gax: (Optional) Explicitly specifies whether
                     to use the gRPC transport (via GAX) or HTTP. If unset,
                     falls back to the ``GOOGLE_CLOUD_DISABLE_GRPC`` environment
-                    variable
+                    variable.
+
+    :type client_config: dict
+    :param client_config: A dictionary for call options for each method. See
```
dhermes: Note that it's optional? `(Optional) A dictionary...`

lukesneeringer:
@dhermes Can I get some feedback on the _substance_ of the change? :-) As noted in the text, I am not sure whether we should actually do this.

dhermes:
@lukesneeringer Sorry I didn't notice the text. (I hate it when people ignore my carefully crafted text, so big time hypocrite there.)

ISTM a better option would be to take the entire GAPIC client as an optional `Client` argument? It's strange having constructor args that only apply to one of the transports.

daspecster:
@lukesneeringer I 2nd @dhermes. We should also test to see if this is happening on the HTTP side as well.  If it is, I'm not sure how to pass the config parameters?

lukesneeringer:
@dhermes I like that idea. Sold.

lukesneeringer:
So, the API would look something like...

```python
class Client(BaseClient):
    def __init__(self, credentials=None, http=None, gapic=None, use_gax=None):
```

I have to say I am kind of not a fan of a _separate_ `http` and `gapic` argument, only one of which makes sense based on the ultimate `_use_gax` value. But, I hate backwards-incompatible changes more, so I think that ship has sailed.

(Also, it is a crying shame that we use the terms "gapic" and "gax" nearly interchangably, but that ship has also sailed.)

Should it be spelled `gapic` or `gapic_client`? The former would match `http` but the latter would be clearer.

dhermes:
@lukesneeringer Why don't we bring the ship back to harbor? I'm fine dumping usage of GAX (GAPIC was not a term in use when the first `use_gax` argument was added).

I'm also fine overloading what WAS `http` as `transport_object` and then verifying that object based on which transport has been chosen.

lukesneeringer:
Editor's Note: Pulling this discussion into a chat for expediency.

lukesneeringer:
Discussed with @dhermes offline and we decided to punt on this question.

Summary:

  * We should fix the default client config for Speech. (googleapis/googleapis#297)
  * We should take a custom GAPIC, but we are not quite sure yet how we want that to look, based on possible features that we hope the GAPICs will get in Q2.

PR #3213, opened 2017-03-27
Fixing up some format strings in nox configs.

dhermes:
Using `STRING_TEMPLATE % VARIABLE` can introduce hard-to-find bugs if `VARIABLE` is expected to be a string but ends up being a tuple. Instead of using percent formatting, just using `.format`.

Also making tweaks to `get_target_packages` to make some path manipulation / checks OS-independent.

### DIFF
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -25,7 +25,7 @@ def unit_tests(session, python_version):
     """Run the unit test suite."""
 
     # Run unit tests against all supported versions of Python.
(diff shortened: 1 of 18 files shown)


dhermes:
FWIW `git grep ' % [^(]'` turns up quite a lot of cases I've just referred to with "hard-to-track-down" formatting bugs.

jonparrott:
This is one of many reasons I hate % formatting.

PR #3215, opened 2017-03-27
Add integration test for Unicode filenames.

lukesneeringer:
This is a port of GoogleCloudPlatform/google-cloud-dotnet#933.

This also exposes the fact that Python 2 cannot handle some Unicode in URLs at all, because `urllib.quote` raises `KeyError` (wha....?) if you use Unicode. That is an issue with the Python standard library, however.

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -245,6 +245,27 @@ class TestStorageWriteFiles(TestStorageFiles):
         self.assertEqual(base_contents, copied_contents)
 
 
+class TestUnicode(unittest.TestCase):
+    @unittest.skipIf(six.PY2, 'Unicode URLs only work on Python 3.')
+    def test_fetch_object_and_check_content(self):
+        client = storage.Client()
+        bucket = client.bucket('storage-library-test-bucket')
+
+        # Note: These files are public.
+        # Normalization form C: a single character for e-acute;
+        # URL should end with Cafe%CC%81
+        # Normalization Form D: an ASCII e followed by U+0301 combining
+        # character; URL should end with Caf%C3%A9
+        test_data = {
+            u'Caf\u00e9': b'Normalization Form C',
+            u'Cafe\u0301': b'Normalization Form D',
+        }
+        for blob_name, file_contents in test_data.items():
+            blob = bucket.blob(blob_name)
+            self.assertEqual(blob.name, blob_name)
+            self.assertEqual(blob.download_as_string(), file_contents)
+
+
 class TestStorageListFiles(TestStorageFiles):
 
     FILENAMES = ('CloudLogo1', 'CloudLogo2', 'CloudLogo3')
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -246,7 +246,6 @@ class TestStorageWriteFiles(TestStorageFiles):
 
 
 class TestUnicode(unittest.TestCase):
-    @unittest.skipIf(six.PY2, 'Unicode URLs only work on Python 3.')
     def test_fetch_object_and_check_content(self):
         client = storage.Client()
         bucket = client.bucket('storage-library-test-bucket')
@@ -257,8 +256,8 @@ class TestUnicode(unittest.TestCase):
         # Normalization Form D: an ASCII e followed by U+0301 combining
         # character; URL should end with Caf%C3%A9
         test_data = {
-            u'Caf\u00e9': b'Normalization Form C',
-            u'Cafe\u0301': b'Normalization Form D',
+            u'Caf\u00e9'.encode('utf-8'): b'Normalization Form C',
+            u'Cafe\u0301'.encode('utf-8'): b'Normalization Form D',
         }
         for blob_name, file_contents in test_data.items():
             blob = bucket.blob(blob_name)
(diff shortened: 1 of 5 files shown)


storage/tests/system.py
```diff
@@ -245,6 +245,27 @@ def test_copy_existing_file(self):
         self.assertEqual(base_contents, copied_contents)
 
 
+class TestUnicode(unittest.TestCase):
+    @unittest.skipIf(six.PY2, 'Unicode URLs only work on Python 3.')
```
dhermes: @lukesneeringer Can you print the Python 2 error message?

lukesneeringer:
```
=============================================== FAILURES ===============================================
___________________________ TestUnicode.test_fetch_object_and_check_content ____________________________

self = <tests.system.TestUnicode testMethod=test_fetch_object_and_check_content>

    def test_fetch_object_and_check_content(self):
        client = storage.Client()
        bucket = client.bucket('storage-library-test-bucket')
    
        # Note: These files are public.
        # Normalization form C: a single character for e-acute;
        # URL should end with Cafe%CC%81
        # Normalization Form D: an ASCII e followed by U+0301 combining
        # character; URL should end with Caf%C3%A9
        test_data = {
            u'Caf\u00e9': b'Normalization Form C',
            u'Cafe\u0301': b'Normalization Form D',
        }
        for blob_name, file_contents in test_data.items():
            blob = bucket.blob(blob_name)
            self.assertEqual(blob.name, blob_name)
>           self.assertEqual(blob.download_as_string(), file_contents)

tests/system.py:265: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../.nox/c749df86/local/lib/python2.7/site-packages/google/cloud/storage/blob.py:405: in download_as_string
    self.download_to_file(string_buffer, client=client)
../.nox/c749df86/local/lib/python2.7/site-packages/google/cloud/storage/blob.py:351: in download_to_file
    self.reload()
../.nox/c749df86/local/lib/python2.7/site-packages/google/cloud/storage/_helpers.py:98: in reload
    method='GET', path=self.path, query_params=query_params,
../.nox/c749df86/local/lib/python2.7/site-packages/google/cloud/storage/blob.py:171: in path
    return self.path_helper(self.bucket.path, self.name)
../.nox/c749df86/local/lib/python2.7/site-packages/google/cloud/storage/blob.py:146: in path_helper
    return bucket_path + '/o/' + quote(blob_name, safe='')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

s = 'Café', safe = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_.-'

    def quote(s, safe='/'):
        """quote('abc def') -> 'abc%20def'
    
        Each part of a URL, e.g. the path info, the query, etc., has a
        different set of reserved characters that must be quoted.
    
        RFC 2396 Uniform Resource Identifiers (URI): Generic Syntax lists
        the following reserved characters.
    
        reserved    = ";" | "/" | "?" | ":" | "@" | "&" | "=" | "+" |
                      "$" | ","
    
        Each of these characters is reserved in some component of a URL,
        but not necessarily in all of them.
    
        By default, the quote function is intended for quoting the path
        section of a URL.  Thus, it will not encode '/'.  This character
        is reserved, but in typical usage the quote function is being
        called on a path where the existing slash characters are used as
        reserved characters.
        """
        # fastpath
        if not s:
            if s is None:
                raise TypeError('None object cannot be quoted')
            return s
        cachekey = (safe, always_safe)
        try:
            (quoter, safe) = _safe_quoters[cachekey]
        except KeyError:
            safe_map = _safe_map.copy()
            safe_map.update([(c, c) for c in safe])
            quoter = safe_map.__getitem__
            safe = always_safe + safe
            _safe_quoters[cachekey] = (quoter, safe)
        if not s.rstrip(safe):
            return s
>       return ''.join(map(quoter, s))
E       KeyError: u'\u0301'

/usr/lib/python2.7/urllib.py:1288: KeyError
1 failed, 20 passed in 28.73 seconds
```

lukesneeringer: @dhermes Is this PR good to go, or do you want me to change something?

dhermes: Change something. See https://gist.github.com/dhermes/cc2d8e42ccd8825efd547152816da899

It seems if we convert to `bytes` first than all will be well?

dhermes:
LGTM

@lukesneeringer What is your feeling on usage of `encode` / `decode` instead of using `google.cloud._helpers._to_bytes` / `_bytes_to_unicode`?

lukesneeringer:
I would prefer use `.encode` and `.decode` when you know what you have (as is the case here), and helper functions when you do not know what you have. When possible, I prefer using helper functions in six versus custom baked ones. (I am pretty sure that `._to_bytes` covers a case that six does not.)

dhermes:
:+1: That is @jonparrott's take as well (so I suppose I should warm to that approach). I wasn't aware that `six` provided **any** support resembling those two helpers, what functions did you have in mind (provided by `six`)?

lukesneeringer:
https://pythonhosted.org/six/#binary-and-text-data

PR #3219, opened 2017-03-28
Fix untested logging snippet examples.

daspecster:
These snippets aren't covered in `logging_snippets.py` and they had a couple small syntax issues.

### DIFF
--- a/docs/logging-stdlib-usage.rst
+++ b/docs/logging-stdlib-usage.rst
@@ -53,7 +53,7 @@ You can also exclude certain loggers:
 
 .. code-block:: python
 
-   >>> setup_logging(handler, excluded_loggers=('werkzeug',)))
+   >>> setup_logging(handler, excluded_loggers=('werkzeug',))
 
 
 
@@ -68,4 +68,3 @@ The Python logging handler can use different transports. The default is
 
  1. :class:`google.cloud.logging.handlers.SyncTransport` this handler does a direct API call on each
  logging statement to write the entry.
-
--- a/logging/google/cloud/logging/handlers/handlers.py
+++ b/logging/google/cloud/logging/handlers/handlers.py
@@ -52,6 +52,7 @@ class CloudLoggingHandler(logging.StreamHandler):
 
     .. code-block:: python
 
+        import logging
         import google.cloud.logging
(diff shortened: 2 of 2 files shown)


docs/logging-stdlib-usage.rst
```diff
@@ -53,7 +53,7 @@ You can also exclude certain loggers:
 
 .. code-block:: python
 
-   >>> setup_logging(handler, excluded_loggers=('werkzeug',)))
+   >>> setup_logging(handler, excluded_loggers=('werkzeug',))
```
dhermes: @daspecster Why don't you move this into `logging_snippets.py`?

logging/google/cloud/logging/handlers/handlers.py
```diff
@@ -117,7 +118,7 @@ def setup_logging(handler, excluded_loggers=EXCLUDED_LOGGER_DEFAULTS,
 
         client = google.cloud.logging.Client()
         handler = CloudLoggingHandler(client)
-        google.cloud.logging.setup_logging(handler)
+        google.cloud.logging.handlers.setup_logging(handler)
```
dhermes: Maybe @waprin inteded `setup_logging` to be present in `__init__.py`?

lukesneeringer:
> Do you have a plan to stabilize them further?

I would love that to be a project soon, but it is not the project I gave him. :-)

logging/google/cloud/logging/handlers/handlers.py
daspecster: Very possible.

docs/logging-stdlib-usage.rst
daspecster: 👍  Future PR I presume though.

logging/google/cloud/logging/handlers/handlers.py
waprin: I originally had it in `__init__.py` before in review we moved it to be a client method, forgot to update snippet here.

lukesneeringer: This seems like it is worth fixing really quickly.

waprin: I believe this is the fix

PR #3221, opened 2017-03-28
Fix storage encryption example, remove policy document example.

daspecster:
This removes the policy document example. I'm not sure why it's failing at the moment.

Also adds a line to the encryption key example to make it work out of the box.

### DIFF
--- a/docs/storage_snippets.py
+++ b/docs/storage_snippets.py
@@ -94,6 +94,7 @@ def download_to_file(client, to_delete):
     bucket = client.get_bucket('my-bucket')
     encryption_key = 'c7f32af42e45e85b9848a6a14dd2a8f6'
     blob = Blob('secure-data', bucket, encryption_key=encryption_key)
+    blob.upload_from_string('my secret message.')
     with open('/tmp/my-secure-file', 'wb') as file_obj:
         blob.download_to_file(file_obj)
     # [END download_to_file]
@@ -138,7 +139,8 @@ def delete_blob(client, to_delete):
     from google.cloud.exceptions import NotFound
     client = storage.Client()
     bucket = client.get_bucket('my-bucket')
-    assert isinstance(bucket.list_blobs(), list)
+    blobs = list(bucket.list_blobs())
+    assert len(blobs) > 0
     # [<Blob: my-bucket, my-file.txt>]
     bucket.delete_blob('my-file.txt')
     try:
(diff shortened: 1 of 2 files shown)


docs/storage_snippets.py
```diff
@@ -138,7 +139,8 @@ def delete_blob(client, to_delete):
     from google.cloud.exceptions import NotFound
     client = storage.Client()
     bucket = client.get_bucket('my-bucket')
-    assert isinstance(bucket.list_blobs(), list)
+    blobs = list(bucket.list_blobs())
+    assert isinstance(blobs, list)
```
dhermes: This assertion is a ~~little~~ lot pointless after manually casting to a `list`

daspecster:
@dhermes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3172#issuecomment-289529729

docs/storage_snippets.py
daspecster: How about
```
assert len(blobs) > 0
```

dhermes:
@daspecster I saw his comment, but didn't interpret it as "I give up forever"

docs/storage_snippets.py
dhermes: The comment below seems to indicate you could even saw `len(blobs) == 1`? Though probably not with a bucket called `my-bucket`?

daspecster:
@dhermes sorry, I'm just trying to get through all this before the GA so there's at least a record of things that may be broken.

I may have taken @lukesneeringer's comment in the meeting about removing examples to far in this case?

daspecster:
Added "do not merge" label until https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3222 is figured out.

dhermes:
@daspecster You can just restore that example so you don't need to block this PR, then settle what is to be settled in #3222

docs/storage_snippets.py
daspecster: yeah `my-bucket` has it's account locked. Or so the error tells me.

Also, other examples above could add blobs in the future so at this point I just assume you want more than 1.

daspecster:
@dhermes restored!

docs/storage_snippets.py
dhermes: Not sure what makes sense. Probably more changes than are worth it in this PR? Definitely need a "local" bucket name so hard-coding a string in the doc is bad.

I would say don't sweat it in this PR?

lukesneeringer:
> I may have taken @lukesneeringer's comment in the meeting about removing examples to far in this case?

No, you were right. If we cannot fix the code in time, I would prefer remove the example. No example (and an issue to fix the bug) is better than an example that fails.

daspecster:
@lukesneeringer ok I'll rebase the last commit out and then we should be good on this one?

docs/storage_snippets.py
tseaver: FWIW, bucket names are in a global namespace:  we should be prefixing them with something like `gcp-storage-` and then a timestamp / random suffix.

PR #3223, opened 2017-03-28
Check out docs submodule with SSH.

lukesneeringer:
This commit moves docs publishing to use SSH instead of an OAuth key
over HTTP. The SSH key is attached to a robot user configured in
CI.

@dhermes: Please review, but there might be some unavoidable post-merge
thrashing if problems arise.

### DIFF
--- a/test_utils/scripts/update_docs.sh
+++ b/test_utils/scripts/update_docs.sh
@@ -92,4 +92,4 @@ git commit -m "Update docs after merge to master."
 
 # NOTE: This may fail if two docs updates (on merges to master)
 #       happen in close proximity.
-git push -q git@github.com:${GH_OWNER}/${GH_PROJECT_NAME} HEAD:gh-pages
+git push -q origin HEAD:gh-pages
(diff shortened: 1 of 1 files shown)


test_utils/scripts/update_docs.sh
```diff
@@ -87,12 +86,10 @@ if [[ -z "$(git status --porcelain)" ]]; then
...
+git config --global user.email "googleapis-publisher@google.com"
+git config --global user.name "Google APIs Publisher"
 git commit -m "Update docs after merge to master."
 
 # NOTE: This may fail if two docs updates (on merges to master)
 #       happen in close proximity.
-git push -q \
-    "https://${GH_OAUTH_TOKEN}@github.com/${GH_OWNER}/${GH_PROJECT_NAME}" \
-    HEAD:gh-pages
+git push -q git@github.com:${GH_OWNER}/${GH_PROJECT_NAME} HEAD:gh-pages
```
dhermes: Can probably just `git push -q origin HEAD:gh-pages`?

lukesneeringer:
> WDYT of git fetch origin gh-pages; git checkout gh-pages instead of creating a submodule?

That could have side effects if we put a new deployment task after it.

dhermes:
LGTM

PR #3226, opened 2017-03-28
Fix the docs.

lukesneeringer:


test_utils/scripts/get_target_packages.py
```diff
@@ -143,12 +143,7 @@ def get_changed_packages(file_list):
...
-    """Figure out what packages have changed."""
+if __name__ == '__main__':
     file_list = get_changed_files()
     for package in sorted(get_changed_packages(file_list)):
-        print(package)
-
-
-if __name__ == '__main__':
-    main()
+       print(package)
```
dhermes: I assume you didn't mean to revert this change?

lukesneeringer:
Fixes a bug in the `update_docs.sh` script, which I needed to thrash a couple times (including with some now-undone scaffolding changes) to get right.

lukesneeringer:
> Also, maybe (in future changes) we just drop get_version? The umbrella docs can't really have a single reliable version attached to them?

I was going to do that when I move the docs around soon.

test_utils/scripts/get_target_packages.py
lukesneeringer: I did. I am opposed to using `print` in functions.

dhermes: Gotcha. SGTM

PR #3230, opened 2017-03-28
Make clients explicitly unpickleable.

### LINKED ISSUE #3211: Raise exception with helpful error on Client pickle.
Based on the discussion in #3191, raise an exception with a helpful error stating that `Client` classes are not pickleable.

Note to self: We probably want this in GAX also.

lukesneeringer:
Closes #3211.

### DIFF
--- a/core/google/cloud/client.py
+++ b/core/google/cloud/client.py
@@ -14,6 +14,8 @@
 
 """Base classes for client used to interact with Google Cloud APIs."""
 
+from pickle import PicklingError
+
 import google.auth.credentials
 from google.oauth2 import service_account
 import google_auth_httplib2
@@ -125,6 +127,14 @@ class Client(_ClientFactoryMixin):
         self._credentials = google.auth.credentials.with_scopes_if_required(
             credentials, self.SCOPE)
         self._http_internal = http
+
+    def __getstate__(self):
+        """Explicitly state that clients are not pickleable."""
+
+        raise PicklingError('\n'.join([
+            'Pickling client objects is explicitly not supported.',
+            'Clients have non-trivial state that is local and unpickleable.',
+        ]))
 
     @property
     def _http(self):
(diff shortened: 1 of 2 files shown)


core/google/cloud/client.py
```diff
@@ -14,6 +14,8 @@
 
 """Base classes for client used to interact with Google Cloud APIs."""
 
+from pickle import PicklingError
```
lukesneeringer: I explicitly went with a single-member import rather than a more general `import pickle` because I wanted it to be immediately clear when this is read months from now that the _only_ thing pickle is being used for is the exception class.

dhermes: Very thoughtful of you :grin: :+1:

core/google/cloud/client.py
```diff
@@ -126,6 +128,14 @@ def __init__(self, credentials=None, http=None):
             credentials, self.SCOPE)
         self._http_internal = http
 
+    def __getstate__(self):
+        """Explicitly state that clients are not pickleable."""
+
```
dhermes: Empty line just for fun?

core/google/cloud/client.py
```diff
@@ -126,6 +128,14 @@ def __init__(self, credentials=None, http=None):
...
+        raise PicklingError('\n'.join([
+            'Pickling client objects is explicitly not supported.',
+            'Clients have non-trivial state that is local and unpickleable.',
+        ]))
```
dhermes: Why not make this error a constant at module-level?

```python
_PICKLE_ERR = """\
FOO
"""
```

lukesneeringer: I would ask the converse question.

core/google/cloud/client.py
lukesneeringer: That used to be required by PEP 8, and I got used to that before PEP 257 became a thing and forbade them. (I actually had to check PEP 257 now to make sure.)

lukesneeringer:
> I would say it's a "teeny performance hit" to create a new string at every use, but that would be wrong since the string is created right before program exit.

The teeny performance hit is to create a new string at module import for everyone else.

dhermes:
Feel free to merge once you get CI green

PR #3235, opened 2017-03-29
Renaming http argument(s) as _http.

dhermes:
Will also shortly be sending a commit to this PR which renames `use_gax` and `_use_grpc`.

`grep`-ing for `http` and `http=` was pretty ad-hoc, so I'm not sure if there are lingering things needing to be changed (e.g. docs which refer to `http` instead of `_http`)

Note in the commit about `use_gax` the following commands were used:

```
$ git grep -l use_gax | xargs sed -i s/use_gax/_use_grpc/g
$ git grep -l __use_grpc | xargs sed -i s/__use_grpc/_use_grpc/g
$ git grep -l USE_GAX | xargs sed -i s/USE_GAX/USE_GRPC/g
$ git grep -l HAVE_GAX | xargs sed -i s/HAVE_GAX/HAVE_GRPC/g
$ git grep -l DISABLE_GAX | xargs sed -i s/DISABLE_GAX/DISABLE_GRPC/g
```

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -70,6 +70,8 @@ class Client(ClientWithProject):
                   :meth:`~httplib2.Http.request`. If not passed, an
                   ``_http`` object is created that is bound to the
                   ``credentials`` for the current object.
+                  This parameter should be considered private, and could
+                  change in the future.
     """
 
     SCOPE = ('https://www.googleapis.com/auth/bigquery',
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -60,25 +60,25 @@ class Client(ClientWithProject):
 
     :type credentials: :class:`~google.auth.credentials.Credentials`
     :param credentials: (Optional) The OAuth2 Credentials to use for this
-                        client. If not passed (and if no ``http`` object is
+                        client. If not passed (and if no ``_http`` object is
                         passed), falls back to the default inferred from the
                         environment.
 
-    :type http: :class:`~httplib2.Http`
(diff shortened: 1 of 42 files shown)


dhermes:
@lukesneeringer :+1: on your commit

dhermes:
@lukesneeringer We all good here?

surjikal:
@dhermes I just started getting errors that might have to do with this change, though I'm not sure. It's happening in my dataproc jobs, but I was able to reproduce using a plain VM instance:

```bash
$ sudo easy_install pip
$ sudo pip install google-cloud google-cloud-storage
$ python -c "from google.cloud import storage; storage.Client(project="<project>")"

  File "/usr/local/lib/python2.7/dist-packages/google/cloud/storage/client.py", line 59, in 
__init__
    _http=_http)
TypeError: __init__() got an unexpected keyword argument '_http'
```

I'm going to poke around the code in the VM to see what's up.

dhermes:
I'd guess you have versions of `google-cloud-core` and `google-cloud-storage` that conflict with one another.

dhermes:
Thanks for letting us know though!

surjikal:
This was on a clean, new VM though.

surjikal:
But you are correct, I added `pip install google-cloud-core` and now all is good. Thanks!

supertom:
@lukesneeringer @dhermes Is there background as to why the http object was made private?

I haven't looked at this PR in detail yet, but has this change been made across all current clients?  Future clients will adhere to the same contract?

dhermes:
@supertom Yes this was made across all clients. It was done because we will be changing the underlying HTTP transport implementation, so making it part of the public interface was a bad idea (i.e. since we'd be breaking that interface).

PR #3241, opened 2017-03-30
Add docs for 'result_index'  usage and a system test.

### LINKED ISSUE #3166: Result returned from google.cloud.speech.sample.Sample.streaming_recognize
The **streaming_recognize** method of **Sample** computes its output as follows:

```
        for response in responses:
            for result in response.results:
                if result.is_final or interim_results:
                    yield StreamingSpeechResult.from_pb(result)
```

Thus, it returns a generator where results are no longer grouped by response. I feel that this makes it more difficult for the end user to use the results, since results from the same response should be concatenated in order to obtain the complete transcript of the fragment transcribed so far.

daspecster:
Closes #3166.

Updated the docs and added a system test for using `result_index`.

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -31,8 +31,8 @@ create an instance of :class:`~google.cloud.speech.client.Client`.
     >>> client = speech.Client()
 
 
-Asychronous Recognition
------------------------
+Asynchronous Recognition
+------------------------
 
 The :meth:`~google.cloud.speech.Client.async_recognize` sends audio data to the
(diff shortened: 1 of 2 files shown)


lukesneeringer:
No code change required?

daspecster:
@lukesneeringer turns out there isn't!
Since streaming is just returning the proto for now.

dhermes:
It seems that is not correct: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1455

PR #3244, opened 2017-03-30
Make logging system tests verbose.

lukesneeringer:
@dhermes This is a more permanent change we can make for the logging system tests while we are watching them over the next few days.

### DIFF
--- a/logging/nox.py
+++ b/logging/nox.py
@@ -59,7 +59,7 @@ def system_tests(session, python_version):
     session.install('.')
 
     # Run py.test against the system tests.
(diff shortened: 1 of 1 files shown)


dhermes:
@lukesneeringer Are we worried that `-vvv` will leak secrets from env. vars?

lukesneeringer:
Wow, their docs are terrible on this. "Verbose" should not be a hard thing to search for.
I have no idea. Feel free to change it to `-vv` if you want.

dhermes:
No it's fine. Feel free to merge.

PR #3249, opened 2017-03-30
Revert "Add docs for 'result_index'  usage and a system test."

daspecster:
Reverts GoogleCloudPlatform/google-cloud-python#3241

### DIFF
--- a/docs/speech-usage.rst
+++ b/docs/speech-usage.rst
@@ -31,8 +31,8 @@ create an instance of :class:`~google.cloud.speech.client.Client`.
     >>> client = speech.Client()
 
 
-Asychronous Recognition
------------------------
+Asynchronous Recognition
+------------------------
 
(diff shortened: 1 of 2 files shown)


daspecster:
I thought the system tests passed locally but it turns out they didn't run but happily passed anyway.

So as not to interfere with the GA let's just revert this and I can fix it for later.

docs/speech-usage.rst
```diff
@@ -31,8 +31,8 @@ create an instance of :class:`~google.cloud.speech.client.Client`.
     >>> client = speech.Client()
 
 
-Asynchronous Recognition
-------------------------
+Asychronous Recognition
```
dhermes: Can you leave this typo fix in (manually edit this PR)

daspecster: got it!

PR #3270, opened 2017-04-03
Gracefully continue if LogEntry.proto_payload type URL is not in registry.

### LINKED ISSUE #2674: New 'MessageToDict' implementation breaks with unknown entry types
```python
>>> from google.cloud.logging import Client
>>> client = Client()
>>> def do_something_with(entry): pass
... 
>>> for entry in client.list_entries():
...     do_something_with(entry)
... 
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/cloud/iterator.py", line 211, in _items_iter
    for item in page:
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/cloud/iterator.py", line 155, in next
    result = self._item_to_value(self._parent, item)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/cloud/logging/_gax.py", line 478, in _item_to_entry
    resource = MessageToDict(entry_pb)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 133, in MessageToDict
    return printer._MessageToJsonObject(message)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 164, in _MessageToJsonObject
    return self._RegularMessageToJsonObject(message, js)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 196, in _RegularMessageToJsonObject
    js[name] = self._FieldToJsonObject(field, value)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 230, in _FieldToJsonObject
    return self._MessageToJsonObject(value)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 162, in _MessageToJsonObject
    return methodcaller(_WKTJSONMETHODS[full_name][0], message)(self)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 266, in _AnyMessageToJsonObject
    sub_message = _CreateMessageFromTypeUrl(type_url)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/.tox/docs/lib/python2.7/site-packages/google/protobuf/json_format.py", line 341, in _CreateMessageFromTypeUrl
    'Can not find message descriptor by type_url: {0}.'.format(type_url))
TypeError: Can not find message descriptor by type_url: type.googleapis.com/google.cloud.audit.AuditLog.
```

dhermes:
Fixes #2674.

@lukesneeringer @tseaver I am unfortunately "out of time" until Wednesday so if either of you want to take this over the finish line (e.g. unit tests and maybe a system test with an unregistered type URL?) I would not protest.

### DIFF
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -243,6 +243,8 @@ class _SinksAPI(object):
             if exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
                 raise NotFound(path)
             raise
+        # NOTE: LogSink message type does not have an ``Any`` field
+        #       so `MessageToDict`` can safely be used.
         return MessageToDict(sink_pb)
 
     def sink_update(self, project, sink_name, filter_, destination):
@@ -270,11 +272,13 @@ class _SinksAPI(object):
         path = 'projects/%s/sinks/%s' % (project, sink_name)
         sink_pb = LogSink(name=path, filter=filter_, destination=destination)
         try:
-            self._gax_api.update_sink(path, sink_pb, options=options)
+            sink_pb = self._gax_api.update_sink(path, sink_pb, options=options)
         except GaxError as exc:
             if exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
                 raise NotFound(path)
             raise
+        # NOTE: LogSink message type does not have an ``Any`` field
+        #       so `MessageToDict`` can safely be used.
         return MessageToDict(sink_pb)
 
     def sink_delete(self, project, sink_name):
@@ -391,6 +395,8 @@ class _MetricsAPI(object):
             if exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
                 raise NotFound(path)
             raise
+        # NOTE: LogMetric message type does not have an ``Any`` field
+        #       so `MessageToDict`` can safely be used.
         return MessageToDict(metric_pb)
 
     def metric_update(self, project, metric_name, filter_, description):
@@ -418,11 +424,14 @@ class _MetricsAPI(object):
         metric_pb = LogMetric(name=path, filter=filter_,
                               description=description)
         try:
-            self._gax_api.update_log_metric(path, metric_pb, options=options)
+            metric_pb = self._gax_api.update_log_metric(
+                path, metric_pb, options=options)
         except GaxError as exc:
             if exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
                 raise NotFound(path)
             raise
+        # NOTE: LogMetric message type does not have an ``Any`` field
+        #       so `MessageToDict`` can safely be used.
         return MessageToDict(metric_pb)
 
     def metric_delete(self, project, metric_name):
@@ -444,6 +453,35 @@ class _MetricsAPI(object):
             raise
 
 
+def _parse_log_entry(entry_pb):
+    """Special helper to parse ``LogEntry`` protobuf into a dictionary.
+
+    The ``proto_payload`` field in ``LogEntry`` is of type ``Any``. This
+    can be problematic if the type URL in the payload isn't in the
+    ``google.protobuf`` registry. To help with parsing unregistered types,
+    this function will remove ``proto_payload`` before parsing.
+
+    :type entry_pb: :class:`.log_entry_pb2.LogEntry`
+    :param entry_pb: Log entry protobuf.
+
+    :rtype: dict
+    :returns: The parsed log entry. The ``protoPayload`` key may contain
+              the raw ``Any`` protobuf from ``entry_pb.proto_payload`` if
+              it could not be parsed.
+    """
+    try:
+        return MessageToDict(entry_pb)
+    except TypeError:
+        if entry_pb.HasField('proto_payload'):
+            proto_payload = entry_pb.proto_payload
+            entry_pb.ClearField('proto_payload')
+            entry_mapping = MessageToDict(entry_pb)
+            entry_mapping['protoPayload'] = proto_payload
+            return entry_mapping
+        else:
+            raise
+
+
 def _log_entry_mapping_to_pb(mapping):
     """Helper for :meth:`write_entries`, et aliae
 
@@ -451,6 +489,13 @@ def _log_entry_mapping_to_pb(mapping):
     the keys expected in the JSON API.
     """
     entry_pb = LogEntry()
+    # NOTE: We assume ``mapping`` was created in ``Batch.commit``
+    #       or ``Logger._make_entry_resource``. In either case, if
+    #       the ``protoPayload`` key is present, we assume that the
+    #       type URL is registered with ``google.protobuf`` and will
+    #       not cause any issues in the JSON->protobuf conversion
+    #       of the corresponding ``proto_payload`` in the log entry
+    #       (it is an ``Any`` field).
     ParseDict(mapping, entry_pb)
     return entry_pb
 
@@ -482,7 +527,7 @@ def _item_to_entry(iterator, entry_pb, loggers):
     :rtype: :class:`~google.cloud.logging.entries._BaseEntry`
     :returns: The next log entry in the page.
     """
-    resource = MessageToDict(entry_pb)
+    resource = _parse_log_entry(entry_pb)
     return entry_from_resource(resource, iterator.client, loggers)
 
 
@@ -499,6 +544,8 @@ def _item_to_sink(iterator, log_sink_pb):
     :rtype: :class:`~google.cloud.logging.sink.Sink`
     :returns: The next sink in the page.
     """
+    # NOTE: LogSink message type does not have an ``Any`` field
+    #       so `MessageToDict`` can safely be used.
     resource = MessageToDict(log_sink_pb)
     return Sink.from_api_repr(resource, iterator.client)
 
@@ -516,6 +563,8 @@ def _item_to_metric(iterator, log_metric_pb):
     :rtype: :class:`~google.cloud.logging.metric.Metric`
     :returns: The next metric in the page.
     """
+    # NOTE: LogMetric message type does not have an ``Any`` field
+    #       so `MessageToDict`` can safely be used.
     resource = MessageToDict(log_metric_pb)
     return Metric.from_api_repr(resource, iterator.client)
 
--- a/logging/google/cloud/logging/_http.py
+++ b/logging/google/cloud/logging/_http.py
@@ -286,6 +286,9 @@ class _SinksAPI(object):
         :type destination: str
         :param destination: destination URI for the entries exported by
                             the sink.
+
+        :rtype: dict
+        :returns: The returned (updated) resource.
         """
         target = '/projects/%s/sinks/%s' % (project, sink_name)
         data = {
@@ -421,6 +424,9 @@ class _MetricsAPI(object):
 
         :type description: str
         :param description: description of the metric.
+
+        :rtype: dict
+        :returns: The returned (updated) resource.
         """
         target = '/projects/%s/metrics/%s' % (project, metric_name)
         data = {
--- a/logging/google/cloud/logging/_http.py
+++ b/logging/google/cloud/logging/_http.py
@@ -293,7 +293,7 @@ class _SinksAPI(object):
             'filter': filter_,
             'destination': destination,
         }
-        self.api_request(method='PUT', path=target, data=data)
+        return self.api_request(method='PUT', path=target, data=data)
 
     def sink_delete(self, project, sink_name):
         """API call:  delete a sink resource.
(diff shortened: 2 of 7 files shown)


logging/google/cloud/logging/_gax.py
```diff
@@ -444,13 +453,49 @@ def metric_delete(self, project, metric_name):
...
+    :param entry_pb: Log entry protobuf.
+
+    :rtype: dict
+    :returns: The parsed log entry. The ``protoPayload`` key may contain
+              the raw ``Any`` protobuf from ``entry_pb.proto_payload`` if
+              it could not be parsed.
+    """
+    try:
+        return MessageToDict(entry_pb)
+    except TypeError:
```
dhermes: @lukesneeringer @tseaver I had mentioned that I wanted to avoid flow control with exceptions here, but there is no real path offered by `google.protobuf`.

I started digging in with the debugger at:

```python
>>> from google.protobuf import descriptor_pool
>>> pool = descriptor_pool.Default()
>>> type_name = 'google.bigtable.admin.v2.UpdateClusterMetadata'
>>> pool.FindMessageTypeByName(type_name)
```

but it turns out `google.protobuf` also does flow control with exceptions (or it dives into C extensions)

logging/google/cloud/logging/_gax.py
```diff
@@ -270,11 +272,13 @@ def sink_update(self, project, sink_name, filter_, destination):
         path = 'projects/%s/sinks/%s' % (project, sink_name)
         sink_pb = LogSink(name=path, filter=filter_, destination=destination)
         try:
-            self._gax_api.update_sink(path, sink_pb, options=options)
+            sink_pb = self._gax_api.update_sink(path, sink_pb, options=options)
```
dhermes: @lukesneeringer @tseaver This is a change I'm not sure I should make here. But I noticed that we are throwing away the "updated" protobuf and there may be small changes (e.g. timestamps) between the local one and the response.

logging/google/cloud/logging/_gax.py
```diff
@@ -418,11 +424,14 @@ def metric_update(self, project, metric_name, filter_, description):
         metric_pb = LogMetric(name=path, filter=filter_,
                               description=description)
         try:
-            self._gax_api.update_log_metric(path, metric_pb, options=options)
+            metric_pb = self._gax_api.update_log_metric(
```
dhermes: Ditto here on an update change I'm not sure I should make / probably shouldn't be in this PR

logging/google/cloud/logging/_http.py
```diff
@@ -293,7 +293,7 @@ def sink_update(self, project, sink_name, filter_, destination):
             'filter': filter_,
             'destination': destination,
         }
-        self.api_request(method='PUT', path=target, data=data)
+        return self.api_request(method='PUT', path=target, data=data)
```
dhermes: Ditto here on an update change I'm not sure I should make / probably shouldn't be in this PR

logging/google/cloud/logging/_http.py
```diff
@@ -428,7 +428,7 @@ def metric_update(self, project, metric_name, filter_, description):
             'filter': filter_,
             'description': description,
         }
-        self.api_request(method='PUT', path=target, data=data)
+        return self.api_request(method='PUT', path=target, data=data)
```
dhermes: Ditto here on an update change I'm not sure I should make / probably shouldn't be in this PR

logging/google/cloud/logging/entries.py
```diff
@@ -74,7 +77,13 @@ class _BaseEntry(object):
     """
     def __init__(self, payload, logger, insert_id=None, timestamp=None,
                  labels=None, severity=None, http_request=None):
-        self.payload = payload
+        if isinstance(payload, any_pb2.Any):
```
dhermes: @lukesneeringer @tseaver I am "sad" about this change in that it incurs an `isinstance` penalty for every single user of this base class.

logging/google/cloud/logging/_gax.py
lukesneeringer: Bah. Oh well.

lukesneeringer:
I will try to get this over the line today if I can.

logging/google/cloud/logging/_gax.py
dhermes: @lukesneeringer Can you weigh in on this particular change pretty please?

logging/google/cloud/logging/entries.py
```diff
@@ -144,9 +145,45 @@ class ProtobufEntry(_BaseEntry):
...
+    def __init__(self, payload, logger, insert_id=None, timestamp=None,
+                 labels=None, severity=None, http_request=None):
+        super(ProtobufEntry, self).__init__(
+            payload, logger, insert_id=insert_id, timestamp=timestamp,
+            labels=labels, severity=severity, http_request=http_request)
+        if isinstance(self.payload, any_pb2.Any):
+            self.payload_pb = self.payload
+            self.payload = None
+        else:
+            self.payload_pb = None
```
dhermes: So I moved this into `ProtobufEntry` alone but now all of the `ProtobufEntry`-s pay the cost of this branch (@jonparrott and @tseaver always scold me when I worry about branches)

logging/tests/system.py
```diff
@@ -123,6 +124,31 @@ def tearDown(self):
...
+        filter_ = self.TYPE_FILTER.format(type_url)
+        entry_iter = iter(
+            Config.CLIENT.list_entries(page_size=1, filter_=filter_))
+        protobuf_entry = next(entry_iter)
+        self.assertIsInstance(protobuf_entry, entries.ProtobufEntry)
+        if Config.CLIENT._use_grpc:
+            self.assertIsNone(protobuf_entry.payload)
+            self.assertIsInstance(protobuf_entry.payload_pb, any_pb2.Any)
+            self.assertEqual(protobuf_entry.payload_pb.type_url, type_url)
+        else:
```
dhermes: Do I need a `# pragma: NO COVER` for system tests branches which are untested? The `else:` branch only gets hit if the `GOOGLE_CLOUD_DISABLE_GRPC` env. is set

logging/tests/system.py
```diff
@@ -108,6 +108,7 @@ class TestLogging(unittest.TestCase):
             'precipitation': False,
         },
     }
+    TYPE_FILTER = 'protoPayload.@type = "{}"'
```
dhermes: @waprin I already ran this so I know it "works on my machine". But is this the correct syntax to be using for a filter?

logging/tests/system.py
```diff
@@ -123,6 +124,31 @@ def tearDown(self):
...
+        type_name = 'google.cloud.audit.AuditLog'
+        # Make sure the descriptor is not known in the registry.
+        with self.assertRaises(KeyError):
+            pool.FindMessageTypeByName(type_name)
+
+        type_url = 'type.googleapis.com/' + type_name
+        filter_ = self.TYPE_FILTER.format(type_url)
+        entry_iter = iter(
+            Config.CLIENT.list_entries(page_size=1, filter_=filter_))
+        protobuf_entry = next(entry_iter)
```
dhermes: @lukesneeringer @tseaver 

1. What sort of `@Retry` should I bake into this `next()` call on the iterator?
1. I realized that **after** this `next(entry_iter)` keeps going, which surprised me. We should probably have a `max_results` keyword as well?

logging/tests/unit/test__gax.py
```diff
@@ -1101,6 +1101,172 @@ def test_metric_delete_hit(self):
...
+        json_mapping = {
+            'protoPayload': {
+                '@type': type_url,
+                'originalRequest': {
+                    'name': 'foo',
+                    'location': 'bar',
+                },
+                'requestTime': {
+                    'seconds': 1491000125,
+                }
```
dhermes: DERP I missed a trailing comma

logging/tests/unit/test__gax.py
```diff
@@ -1101,6 +1101,172 @@ def test_metric_delete_hit(self):
...
+            }
```
dhermes: DERP I missed a trailing comma

logging/google/cloud/logging/_gax.py
lukesneeringer: So, the difference is, now we save the _new_, updated `sink_pb`. That _seems_ right to me...

logging/google/cloud/logging/entries.py
lukesneeringer: I think I am fine with the cost in this case.

logging/tests/system.py
lukesneeringer: 1. I am not sure.
2. `max_results` would be a good way to keep the system test small / testing a particular thing, also.

logging/tests/system.py
lukesneeringer: No, we do not measure coverage of system tests.

logging/google/cloud/logging/_gax.py
tseaver: There are no "output-only" fields in the `LogSink` resource / proto, so it wouldn't matter either way (`startTime`/`endTime` are user-settable policy knobs).

logging/google/cloud/logging/_gax.py
tseaver: Likewise, `LogMetric` has no output-only fields.  It can't *hurt* to capture it, but it shouldn't matter, either.

logging/google/cloud/logging/_gax.py
tseaver: As an alternative, we could just crack the `proto_payload` bit out unconditionally (so the user would have to parse it no matter what).

logging/google/cloud/logging/_http.py
tseaver: If you're gonna do it (I'm -1, personally), you need to add `:rtype:` and `:returns:` to the docstring.

logging/google/cloud/logging/_http.py
tseaver: Ditto as well.

logging/tests/system.py
tseaver: The goal is to screen out log entries which have non-empty `protoPayload`?

logging/tests/system.py
tseaver: - Do we need a retry at all?
- +1 for `max_results`.

dhermes: 1. I am leaning towards no retry until the first flaky test failure?
1. I filed #3277

logging/google/cloud/logging/_gax.py
dhermes: @tseaver The goal was to get as close to "best experience" as possible. Since the JSON API deserializes them for free, it'd be nice to do this as often as possible.

@lukesneeringer WDYT?

dhermes:
@tseaver @lukesneeringer Do you have any issues?

@waprin Can you vet my newly added system test?

lukesneeringer:
No issues. Approved.

logging/tests/system.py
waprin: LGTM

waprin:
@dhermes tied up today but can review your tests and try to make mine less flaky (and loop back to any other pending issues for me in this repo) tomorrow.

dhermes:
Thanks @waprin. AFAIK the other ones aren't very flaky any more (or not at all?).

Let's get this merged and then @waprin you can "verify" the new system test after the fact?

logging/tests/system.py
lukesneeringer: > I am leaning towards no retry until the first flaky test failure?

I am fine with this rule of thumb.

waprin:
@dhermes sorry for delay, nice PR. One quick note about the system test - I think it's fine but in general the gotcha to be aware of with Logging is that you can get an empty page followed by a page with results. The nextPageToken is used as a sort of "pending" sign (not a fan of this but that's how it works). So in theory `entry = next(iter)` could fail if the query is too slow, really you always want to wrap it it in a loop (not sure if there is an elegant way to have client library do this for you). Now usually you only get slow queries when looking through huge amounts of logs like all of the ones for a project but it's something to be aware of. Again not really an issue just giving you a heads up on something that's burned me and others.

dhermes:
@waprin I'll be on the lookout for that if there is a flaky failure.

PR #3271, opened 2017-04-04
Fix missing BYTES merger

samizuh:
Client needs to know which merger to use when merging column type BYTES that is consumed in chunks as part of a read. Without this fix, client gives a traceback:
```
.../venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 262, in _merge_by_type
    merger = _MERGE_BY_TYPE[type_.code]
KeyError: 7
```
Type 7 is BYTES from the [proto](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto) definition 
The error condition will arise if you write an image (a few MB in size) as base64 encoded in a bytes column. When trying to read the column back using the client, the above traceback will be given. With this fix, the client will use the string merger (treating bytes as a string) and allow the row to be consumed. The test is to read the entire column (with this fix) and write the bytes back to a file (base64 decoded).

### DIFF
--- a/spanner/google/cloud/spanner/streamed.py
+++ b/spanner/google/cloud/spanner/streamed.py
@@ -253,6 +253,7 @@ _MERGE_BY_TYPE = {
     type_pb2.STRING: _merge_string,
     type_pb2.ARRAY: _merge_array,
     type_pb2.STRUCT: _merge_struct,
+    type_pb2.BYTES: _merge_string,
 }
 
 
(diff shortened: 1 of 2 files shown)


samizuh:
Added test case, please let me know if that's sufficient coverage.

tseaver:
Thank you for the patch!

samizuh:
Awesome, thanks!

PR #3279, opened 2017-04-06
Retry test

calpeyser:


bigtable/unit_tests/retry_server.go
```diff
@@ -0,0 +1,295 @@
+package main
```
lukesneeringer: Did you mean for this file to go somewhere else? (This is a Python repository.)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -27,7 +29,29 @@
 from google.cloud.bigtable.row import ConditionalRow
 from google.cloud.bigtable.row import DirectRow
 from google.cloud.bigtable.row_data import PartialRowsData
-
+from google.gax import RetryOptions, BackoffSettings
+from google.cloud.bigtable.retry import ReadRowsIterator, _create_row_request
+from grpc import StatusCode
+
+BACKOFF_SETTINGS = BackoffSettings(
```
lukesneeringer: Are these taken from the [GAPIC settings][]? If so, we should load from there; if not, we should alter those. They look to be pretty different; we should figure out what the correct values are.

  [GAPIC settings]: https://github.com/googleapis/api-client-staging/blob/master/generated/python/gapic-google-bigtable-v2/google/cloud/gapic/bigtable/v2/bigtable_client_config.json

bigtable/google/cloud/bigtable/table.py
```diff
@@ -13,8 +13,10 @@
 # limitations under the License.
 
 """User friendly container for Google Cloud Bigtable Table."""
+from __future__ import absolute_import, division
```
lukesneeringer: Is there division? A quick eyeball grep did not notice any.

bigtable/unit_tests/retry_server.go
lukesneeringer: Also, what are the binary files for?

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,169 @@
...
+        self.max_delay_millis = retry_options.backoff_settings.max_retry_delay_millis
+        self.has_timeout_settings = _has_timeout_settings(retry_options.backoff_settings)
+
+        if self.has_timeout_settings:
+            self.timeout_mult = retry_options.backoff_settings.rpc_timeout_multiplier
+            self.max_timeout = (retry_options.backoff_settings.max_rpc_timeout_millis / _MILLIS_PER_SECOND)
+            self.total_timeout = (retry_options.backoff_settings.total_timeout_millis / _MILLIS_PER_SECOND)
+            self.set_stream()
+
+    def set_start_key(self, start_key):
```
lukesneeringer: Would it be possible to put docstrings on these (this request applies to every public method)?

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,169 @@
...
+            now = time.time()
+            deadline = now + self.total_timeout
+        else:
+            timeout = None
+            deadline = None
+
+        while deadline is None or now < deadline:
+            try:
+                return six.next(self.stream)
+            except StopIteration as stop:
```
lukesneeringer: ~These two lines (74-75) are superfluous.~ Nevermind, they are not -- but they belie a problem below.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,169 @@
...
+                raise stop
+            except Exception as exception:  # pylint: disable=broad-except
```
lukesneeringer: Human-lint would like to know why we are catching every exception. :-)
Is it only possible to get GRPC exceptions here? If so, can we catch just those? If not, then this still will not work because line 77 expects a GRPC exception code.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,169 @@
...
+                code = config.exc_to_code(exception)
+                if code not in self.retry_options.retry_codes:
+                    raise errors.RetryError(
```
lukesneeringer: I am assuming you copied this from us, so I feel bad remarking on it, but I am trying to move us to using `six.reraise` for this and preserving the original traceback.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,169 @@
...
+                        'Exception occurred in retry method that was not'
+                        ' classified as transient', exception)
+
+                # pylint: disable=redefined-variable-type
+                exc = errors.RetryError(
```
lukesneeringer: Same as above.

bigtable/unit_tests/test_table.py
```diff
@@ -1,599 +0,0 @@
-# Copyright 2015 Google Inc.
```
garye: Did you mean to delete this test?

bigtable/unit_tests/test_retry.py
```diff
@@ -0,0 +1,54 @@
...
+        elif (op == "SCAN"):
+            self.process_scan(table, chunks[2], chunks[3])
+
+    def process_read(self, table, payload):
+        pass
+
+    def process_write(self, table, payload):
+        pass
+
+    def process_scan(self, table, range, ids):
```
garye: Looks like this test is a work in progress, but 2 things that jumped out as TODOs are to  process the stdout of the server (looking for PASS or FAIL) and to validate that the rows returned are correct according to the test script.

bigtable/unit_tests/test_retry.py
```diff
@@ -0,0 +1,54 @@
+import unittest
```
garye: I think this should be a system test, not a unit test, since there's an outside dependency.

bigtable/unit_tests/retry_server.go
garye: The go file doesn't have to be here, but the compiled go code is a server that we want to run system/integration tests against. Sorta like testing against the emulator. I was thinking we could have a public GCS bucket that contained the binaries for each platform and pull the appropriate one down during systest to run. Open to other ideas!

bigtable/google/cloud/bigtable/table.py
garye: In particular we don't want a total timeout at all, really, since bigtable scans can take many minutes. I don't think the python library supports user-supplied deadlines but that's the only reason we'd want to stop a read.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -274,6 +274,9 @@ def consume_next(self):
 
             self._validate_chunk(chunk)
 
+            if ("ReadRowsIterator" in self._response_iterator.__class__.__name__):
+                self._response_iterator.set_start_key(chunk.row_key)
```
garye: I think this can only happen once the row has been fully processed, so after `_save_current_row`?

tseaver:
Superseded by #3324

PR #3266, opened 2017-04-03
Speech v1

lukesneeringer:
This updates our manual client library to the Speech v1 API.

This entails several **backwards incompatible changes**:

  * The `language_code` parameter is no longer optional anywhere. It must be explicitly specified, and does _not_ default to `'en-US'`.
  * The `sync_recognize` method has been renamed to `recognize` on every class where it appears.
  * The `async_recognize` method has been renamed to `long_running_recognize` on every class where it appears.
  * The `sample_rate` parameter and property has been renamed to `sample_rate_hertz` everywhere it appears.

Additionally, the backend API contains a backwards incompatible change which does not require a code change in the client library, but will likely require one downstream: The `START_OF_SPEECH`, `END_OF_SPEECH`, and `END_OF_AUDIO` events have been removed.

PR #3280, opened 2017-04-07
Bump Speech to 0.25.0.

lukesneeringer:
I will release once this is merged. Following up momentarily with release notes.

### DIFF
--- a/speech/setup.py
+++ b/speech/setup.py
@@ -57,7 +57,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-speech',
-    version='0.24.0',
+    version='0.25.0',
     description='Python Client for Google Cloud Speech',
     long_description=README,
     namespace_packages=[


lukesneeringer:
### Release Notes

    This updates our manual client library to the Speech v1 API. (#3266)
    
    This entails several **backwards incompatible changes**:
    
      * The `language_code` parameter is no longer optional anywhere. It must be explicitly specified, and does _not_ default to `'en-US'`.
      * The `sync_recognize` method has been renamed to `recognize` on every class where it appears.
      * The `async_recognize` method has been renamed to `long_running_recognize` on every class where it appears.
      * The `sample_rate` parameter and property has been renamed to `sample_rate_hertz` everywhere it appears.
    
    Additionally, the backend API contains a backwards incompatible change which does not require a code change in the client library, but will likely require one downstream: The `START_OF_SPEECH`, `END_OF_SPEECH`, and `END_OF_AUDIO` events have been removed.

lukesneeringer:
This is needed quickly and all the code is approved. Sending it through.

PR #3283, opened 2017-04-11
Language: Expose an api_version parameter on the Client.

lukesneeringer:
This commit adds an `api_version` parameter to the `Client` constructor to provide access to the v1beta2 API methods.

Instantiation looks like:

```python
client = Client(api_version='v1beta2')
```

The default value is `'v1'`, and anything other than `'v1'` or `'v1beta2'` is an error.
Note that using the `'v1beta2'` version will route _all_ calls made to the object through that endpoint, and so existing methods will also get `v1beta2` behavior.

Note that neither `v1` nor `v1beta2` actually use the GAPIC.

### DIFF
--- a/language/google/cloud/language/_http.py
+++ b/language/google/cloud/language/_http.py
@@ -41,3 +41,14 @@ class Connection(_http.JSONConnection):
     _EXTRA_HEADERS = {
         _http.CLIENT_INFO_HEADER: _CLIENT_INFO,
     }
+
+
+class V1Beta2Connection(Connection):
+    """A connection to Google Cloud Natural Language JSON 1.1 REST API.
+
+    :type client: :class:`~google.cloud.language.client.Client`
+    :param client: The client that owns the current connection.
+    """
+
+    API_VERSION = 'v1beta2'
+    """The version of the API, used in building the API call's URL."""
--- a/language/google/cloud/language/client.py
+++ b/language/google/cloud/language/client.py
@@ -20,6 +20,7 @@ import warnings
 from google.cloud import client as client_module
 
 from google.cloud.language._http import Connection
+from google.cloud.language._http import V1Beta2Connection
 from google.cloud.language.document import Document
 
 
@@ -45,10 +46,16 @@ class Client(client_module.Client):
     SCOPE = ('https://www.googleapis.com/auth/cloud-platform',)
     """The scopes required for authenticating as an API consumer."""
 
-    def __init__(self, credentials=None, _http=None):
+    _CONNECTION_CLASSES = {
+        'v1': Connection,
+        'v1beta2': V1Beta2Connection,
+    }
+
+    def __init__(self, credentials=None, api_version='v1', _http=None):
         super(Client, self).__init__(
             credentials=credentials, _http=_http)
-        self._connection = Connection(self)
+        ConnectionClass = self._CONNECTION_CLASSES[api_version]
+        self._connection = ConnectionClass(self)
 
     def document_from_text(self, content, **kwargs):
         """Create a plain text document bound to this client.
--- a/language/google/cloud/language/document.py
+++ b/language/google/cloud/language/document.py
@@ -188,6 +188,32 @@ class Document(object):
             method='POST', path='analyzeEntities', data=data)
         return api_responses.EntityResponse.from_api_repr(api_response)
 
+    def analyze_entity_sentiment(self):
+        """Analyze the entity sentiment.
+
+        Finds entities, similar to `AnalyzeEntities` in the text and
+        analyzes sentiment associated with each entity and its mentions.
+
+        :rtype: :class:`~language.entity.EntitySentimentResponse`
+        :returns: A representation of the entity sentiment response.
+        """
+        # Sanity check: Not available on v1.
+        if self.client._connection.API_VERSION == 'v1':
+            raise NotImplementedError(
+                'The `analyze_entity_sentiment` method is only available '
+                'on the Natural Language 1.1 beta. Use version="v1beta2" '
+                'as a keyword argument to the constructor.',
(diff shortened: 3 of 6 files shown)


language/google/cloud/language/_http.py
```diff
@@ -41,3 +41,14 @@ class Connection(_http.JSONConnection):
     _EXTRA_HEADERS = {
         _http.CLIENT_INFO_HEADER: _CLIENT_INFO,
     }
+
+
+class V1Beta2Connection(Connection):
```
dhermes: @lukesneeringer Do we need to support both versions at the same time?

language/google/cloud/language/client.py
```diff
@@ -45,10 +46,16 @@ class Client(client_module.Client):
...
+    _CONNECTION_CLASSES = {
+        'v1': Connection,
+        'v1beta2': V1Beta2Connection,
+    }
+
+    def __init__(self, credentials=None, api_version='v1', _http=None):
         super(Client, self).__init__(
             credentials=credentials, _http=_http)
-        self._connection = Connection(self)
+        ConnectionClass = self._CONNECTION_CLASSES[api_version]
```
dhermes: 1. Are we worried about an unwrapped `KeyError` (if the wrong `api_version` is used)?
1. Human lint: shouldn't it be `connection_class`?

language/google/cloud/language/document.py
```diff
@@ -188,6 +188,32 @@ def analyze_entities(self):
...
+        """Analyze the entity sentiment.
+
+        Finds entities, similar to `AnalyzeEntities` in the text and
+        analyzes sentiment associated with each entity and its mentions.
+
+        :rtype: :class:`~language.entity.EntitySentimentResponse`
+        :returns: A representation of the entity sentiment response.
+        """
+        # Sanity check: Not available on v1.
+        if self.client._connection.API_VERSION == 'v1':
```
dhermes: Would you rather `not isinstance(self.client._connection, V1Beta2Connection)`?

language/google/cloud/language/entity.py
```diff
@@ -176,4 +184,7 @@ def from_api_repr(cls, payload):
         metadata = payload['metadata']
         salience = payload['salience']
         mentions = [Mention.from_api_repr(val) for val in payload['mentions']]
-        return cls(name, entity_type, metadata, salience, mentions)
+        sentiment = None
+        if payload.get('sentiment'):
```
dhermes: Should be `if payload.get('sentiment') is not None` but you probably just want `if 'sentiment' in payload`

language/tests/unit/test_document.py
```diff
@@ -96,12 +96,12 @@ def _get_entities(include_entities):
...
 
 
-def make_mock_client(response):
+def make_mock_client(response, api_version='v1'):
     import mock
-    from google.cloud.language._http import Connection
     from google.cloud.language.client import Client
 
-    connection = mock.Mock(spec=Connection)
+    connection = mock.Mock(spec=Client._CONNECTION_CLASSES[api_version])
```
dhermes: FWIW they both have the same spec (but not the same `__name__`)

language/tests/unit/test_document.py
```diff
@@ -218,6 +219,10 @@ def _verify_entity(self, entity, name, entity_type, wiki_url, salience):
         self.assertEqual(entity.salience, salience)
         self.assertEqual(len(entity.mentions), 1)
         self.assertEqual(entity.mentions[0].text.content, name)
+        if sentiment:
```
dhermes: `if sentiment is not None`

language/tests/unit/test_entity.py
```diff
@@ -101,6 +103,20 @@ def test_from_api_repr(self):
         for mention in entity.mentions:
             self.assertEqual(mention.mention_type, MentionType.PROPER)
 
+        # Assert that there is no sentiment.
+        self.assertIs(entity.sentiment, None)
```
dhermes: You could also use `assertIsNone` if you feel like it

language/tests/unit/test_entity.py
```diff
@@ -101,6 +103,20 @@ def test_from_api_repr(self):
...
+
+        # Assert that there *is* sentiment if it is provided.
+        payload_with_sentiment = dict(payload, sentiment={
```
dhermes: Never have I ever seen a `dict` constructed this way.

Since you never use the `payload` after this line, why not `payload['sentiment'] = ...`?

language/google/cloud/language/_http.py
lukesneeringer: Yes

language/google/cloud/language/document.py
lukesneeringer: I expect newer items to support the endpoint. (v1beta3 seems likely.)

language/tests/unit/test_entity.py
lukesneeringer: Personal preference. :-)

language/google/cloud/language/entity.py
lukesneeringer: I would want an empty dictionary to fail the check.

language/tests/unit/test_document.py
lukesneeringer: See above.

language/tests/unit/test_entity.py
lukesneeringer: True, but I do not want it enough to make a new commit. :-)

language/google/cloud/language/entity.py
dhermes: Why would there be an empty dictionary in there? The JSON payload dumps any empty fields (as does the protobuf payload)

PR #3289, opened 2017-04-11
Version number bump. [ci skip]

lukesneeringer:


dhermes:
@lukesneeringer Do you prefer using `[ci skip]` to just canceling the jobs?

lukesneeringer:
If I remember. :-)

PR #3291, opened 2017-04-12
Do not force LRO to be FLAC only in Speech.

### LINKED ISSUE #3285: Client library incorrectly returns error when FLAC is used with long-running-recognize
The client library incorrectly returns an error when anything other than LINEAR16 is used for a long-running recognize request:

[https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/speech/google/cloud/speech/sample.py#L173][1]

The v1 server support all encodings so this breaks functionality.

In general, the client library should almost never do it's own validation; it should always delegate to the server to be sure that that behavior and error messages are consistent across APIs.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/46e417e838001f4d1eb13f5e3fc2a54d390dd5db/speech/google/cloud/speech/sample.py#L173

lukesneeringer:
Fixes #3285

### DIFF
--- a/speech/google/cloud/speech/sample.py
+++ b/speech/google/cloud/speech/sample.py
@@ -170,9 +170,6 @@ class Sample(object):
         :rtype: :class:`~google.cloud.speech.operation.Operation`
         :returns: Operation for asynchronous request to Google Speech API.
         """
-        if self.encoding is not Encoding.LINEAR16:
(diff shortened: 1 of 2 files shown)


lukesneeringer:
Fixed in the squash commit.

PR #3295, opened 2017-04-13
Fail fast if setuptools is too old on Python 3.6

HaraldNordgren:
Not 100% sure about this, but worth an attempt I think.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import sys
 
 from setuptools import setup
 
@@ -68,6 +69,10 @@ REQUIREMENTS = [
     'google-cloud-vision >= 0.24.0, < 0.25dev',
     'google-cloud-runtimeconfig >= 0.24.0, < 0.25dev',
 ]
+
+if sys.version_info >= (3, 6):
+    REQUIREMENTS.insert(0, 'setuptools >= 34.0.0')
+
 
 setup(
     name='google-cloud',


HaraldNordgren:
The installation works badly on Python 3.6 when the `setuptools` version is old. It doesn't always fail, but often does.

This should be able to reproduce it. Versions are locked to the current ones on that docker image, for reproducibility in the future:

```docker run -ti python:3.6 bash -c "pip install pip==9.0.1 setuptools==28.8.0 && pip install google-cloud-storage==0.23.1"```

**Comment by jonparrott** [comment]
No conditional logic in `setup.py`. We distribute as universal wheels so in almost every case this logic will never be executed. I'd prefer we just depend on a minimum version of setuptools all the time.

**Comment by dhermes** [comment]
> I'd prefer we just depend on a minimum version of `setuptools` all the time.

Is that common? Also, `setup.py` will fail all together if there is **no** `setuptools` installed.

**Comment by jonparrott** [comment]
It's not unheard of. And yeah, setup.py will fail if no setuptools is installed, but I think this particular failure is actually due to pip's bundled setuptools so even this it might not fix it. @HaraldNordgren can you tell us which particular failure you're trying to fix? Is it the `no attribute sort` issue?

**Comment by HaraldNordgren** [comment]
@jonparrott Yes, it's the Attribute error that happens only setup.

```
  ...
  Running setup.py install for googleapis-common-protos: started
    Running setup.py install for googleapis-common-protos: finished with status 'error'
    Complete output from command /usr/local/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-ctc61993/googleapis-common-protos/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-ubmbxcdd-record/install-record.txt --single-version-externally-managed --compile:
      File "/usr/local/lib/python3.6/site-packages/setuptools/__init__.py", line 10, in <module>
        from setuptools.extern.six.moves import filter, filterfalse, map
      File "/usr/local/lib/python3.6/site-packages/setuptools/extern/__init__.py", line 1, in <module>
        from pkg_resources.extern import VendorImporter
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3017, in <module>
        @_call_aside
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3003, in _call_aside
        f(*args, **kwargs)
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3045, in _initialize_master_working_set
        dist.activate(replace=False)
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2577, in activate
        declare_namespace(pkg)
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2151, in declare_namespace
        _handle_ns(packageName, path_item)
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2091, in _handle_ns
        _rebuild_mod_path(path, packageName, module)
      File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2120, in _rebuild_mod_path
        orig_path.sort(key=position_in_sys_path)
    AttributeError: '_NamespacePath' object has no attribute 'sort'
    
    ----------------------------------------
Command "/usr/local/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-ctc61993/googleapis-common-protos/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-ubmbxcdd-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-ctc61993/googleapis-common-protos/
```

dhermes:
@HaraldNordgren Sorry for the delay on the review. I'm not sure a change like this is the right fix for this problem. I (personally) would prefer the current process: address bugs as filed and try to document clearly that people should have up to date tooling (which should come as no surprise).

Working around people having out of date tooling with **explicit** hacks seems very wrong to me.

@lukesneeringer @jonparrott Are we going to try to get this (or something like this) through? If not we should close this PR.

lukesneeringer:
We definitely are not going to do the conditional logic (which will not actually work because eventually wheels).

I am open to depending on setuptools >= 34 always.

dhermes:
@lukesneeringer So it sounds like we should close this PR then? (Sorry @HaraldNordgren but we appreciate the effort.)

lukesneeringer:
Yeah, we should close this PR, and make a new issue to explicitly depend on a version of setuptools.

PR #3292, opened 2017-04-12
Add new encodings.

lukesneeringer:
Fixes #3288.

PR #3297, opened 2017-04-13
Bump Speech to 0.25.1 [ci skip]

lukesneeringer:
Release notes coming momentarily.


dhermes:
LGTM

lukesneeringer:
* Do not force long-running requests to be `LINEAR16` only in Speech. (#3291)
  * Add new encodings. (#3292)
  * Do not validate sample rate. (#3290)

PR #3303, opened 2017-04-17
Add Pub/Sub snapshot and seek functionality

geigerj:
Notes:
- cloud.google.com links do not yet work
- SeekResponse exists but is empty; the corresponding methods return
  None instead
- It's not documented whether the deleted topic path applies to
  snapshots analogous to how it applies to subscriptions, but it seems
  logical to assume so.
- The new Subscription fields are part of this release but will come
  shortly as a separate PR. (They can be done independently of the
  snap-seek work.)

### DIFF
--- a/pubsub/google/cloud/pubsub/_gax.py
+++ b/pubsub/google/cloud/pubsub/_gax.py
@@ -37,6 +37,7 @@ from google.cloud.exceptions import NotFound
 from google.cloud.iterator import GAXIterator
 from google.cloud.pubsub import __version__
 from google.cloud.pubsub._helpers import subscription_name_from_path
+from google.cloud.pubsub.snapshot import Snapshot
 from google.cloud.pubsub.subscription import Subscription
 from google.cloud.pubsub.topic import Topic
 
@@ -135,10 +136,10 @@ class _PublisherAPI(object):
         """API call:  delete a topic
 
         See:
-        https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/create
+        https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/delete
 
         :type topic_path: str
-        :param topic_path: fully-qualified path of the new topic, in format
+        :param topic_path: fully-qualified path of the topic, in format
                             ``projects/<PROJECT>/topics/<TOPIC_NAME>``.
         """
         try:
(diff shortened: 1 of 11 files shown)


geigerj:
@lukesneeringer CI is failing because it thinks [this branch doesn't have test coverage](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3303/files#diff-c15e14c96ae7c1e14ebdff735c3080e6R492). There is a actually [a test intended to exercise this branch](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3303/files#diff-1c9cda78cbe922973ded9e01cc466698R866). I checked raising an exception in the allegedly uncovered lines of code causes the test to fail, so not sure exactly why codecov doesn't like this.

lukesneeringer:
This looks fine other than figuring out the CI failure. I can look at that first thing tomorrow morning.

lukesneeringer:
The reason you are not getting full coverage is because of the `if`/`elif`:

```python
if time is not None:
    data['time'] = time
elif snapshot is not None:
    data['snapshot'] = snapshot
```

You have a test for `time` being set, and one for `snapshot` being set, but the only way that the `elif` will evaluate to `False` is if _neither_ is set.

I am assuming that setting neither option is an error case. (I assume this is also true if you set both.)

Your current logic basically makes `time` win over `snapshot` if both are set. This is probably arbitrary. I suggest making both clauses simple `if` statements and letting the API itself handle the validation if both (or neither) is set. That will also cause all branches to be covered with your current tests.

geigerj:
Thanks for figuring this out, @lukesneeringer. That makes sense, and I agree with your changes.

tseaver:
@geigerj, @lukesneeringer  I'm curious:  why did we rush in support for a feature which is not yet public (i.e., it requires whitelisting, has no public docs, etc., even a month later).

lukesneeringer:
Because it was pushed for by the Pub/Sub team. They claimed at the time that support was imminent. (I raised the non-public issue and was told we were making an exception in this case.)

tseaver:
@lukesneeringer Seems like your concern was well-founded.  I think our normal protocol (develop and review the client-side support in a private fork, push to public at launch) is more sensible.

lukesneeringer:
For features like this, I would prefer not develop them at all until they go public. :-)

PR #3305, opened 2017-04-17
Update speech docs to remove incorrect warnings.

lukesneeringer:


lukesneeringer:
@dhermes Once this is merged, is it possible to _only_ regenerate the docs? (There is no code change so I hate to do a release just for this.)

dhermes:
@lukesneeringer Which docs? **Some** docs get built on every "push" build.

PR #3308, opened 2017-04-18
Store policy bindings as sets, not frozensets.

tseaver:
The legacy accessors still return frozensets, as they cannot safely be mutated in plcae.


lukesneeringer:
Merging this (and #3309) so I can more easily read #3311.

PR #3309, opened 2017-04-18
Add IAM methods for buckets.

tseaver:
Uses #3308 as a base.

Toward #1679.


lukesneeringer:
Merging this so I can more easily read #3311.

PR #3313, opened 2017-04-19
Set `retain_acked_messages` in seek system test

geigerj:
This should prevent the (as of yet, unexperienced) error case where
the API deletes the acked messages before we have a chance to seek
to them.


dhermes:
LGTM though I don't have great context for this.

@geigerj Can we push this branch to the https://github.com/GoogleCloudPlatform/google-cloud-python remote so that system tests get run?

geigerj:
@dhermes Done as #3314

PR #3316, opened 2017-04-19
Adding check (in datastore) that setup.py README is valid RST.

dhermes:
/cc @jonparrott 

@lukesneeringer You mentioned that you'd prefer a standalone session for this. I am unclear why you prefer this. To me, checking that the `README` isn't invalid in `setup.py` **is** linting.

### DIFF
--- a/datastore/nox.py
+++ b/datastore/nox.py
@@ -102,6 +102,15 @@ def lint(session):
 
 
 @nox.session
+def lint_setup_py(session):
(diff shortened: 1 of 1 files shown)


lukesneeringer:
My rationale is:

  * These are separate and orthogonal checks. If one fails and one succeeds, I want to know that.
  * Different virtualenvs ensures atomicity.

dhermes:
OK @lukesneeringer PTAL

PR #3320, opened 2017-04-21
Update BQ Query parameter reference

roscoejp:
Update BigQuery Query parameter reference to indicate that timeout_ms sets timeout on the request, not on the query. Should address https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3317

### DIFF
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -383,7 +383,7 @@ class QueryResults(object):
 
         :type timeout_ms: int
         :param timeout_ms:
-            (Optional) timeout, in milliseconds, to wait for query to complete
+            (Optional) timeout, in milliseconds, to wait for request to complete
 
         :type client: :class:`~google.cloud.bigquery.client.Client` or
                       ``NoneType``
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -386,10 +386,11 @@ class QueryResults(object):
             (Optional) How long to wait for the query to complete, in
             milliseconds, before the request times out and returns. Note that
             this is only a timeout for the request, not the query. If the query
-            takes longer to run than the timeout value, the call returns without
-            any results and with the 'jobComplete' flag set to false. You can
-            call GetQueryResults() to wait for the query to complete and read
-            the results. The default value is 10000 milliseconds (10 seconds).
+            takes longer to run than the timeout value, the call returns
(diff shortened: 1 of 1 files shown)


roscoejp:
I signed it!

tseaver:
The [docs for `timeoutMs`](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#timeoutMs) actually say:

> [Optional] How long to wait for the query to complete, in milliseconds, before the request times out and returns. Note that this is only a timeout for the request, not the query. If the query takes longer to run than the timeout value, the call returns without any results and with the 'jobComplete' flag set to false. You can call GetQueryResults() to wait for the query to complete and read the results. The default value is 10000 milliseconds (10 seconds).

Which makes your change incorrect:  the timeout occurs on the *server* side, not in the client.

lukesneeringer:
@tseaver I do not think so. The timeout is the _request_ timeout. The query continues to run. Note the penultimate sentence:

> You can call `GetQueryResults() to wait for the query to complete and read the results.

That is very much a request timeout, as opposed to a query timeout.

It is true that it is the _server_ that drops the connection, but that distinction is not relevant for most users. They do not care which side severed the connection.

tseaver:
@lukesneeringer Wouldn't it be better to include more of the semantics from the API docs than go with the language in this PR?  Otherwise we end up introducing potential confusion with other client-driven timeouts.

roscoejp:
I was originally trying to keep this short to keep it in line with the other param references. I changed it from 'query timeout' to 'request timeout'  in order to bring it _more_ in line with the API docs - specifically:

> Note that this is only a timeout for the request, not the query

The original docs are actually more confusing than this change because it specifically contradicts the API docs. It also seems unnecessary to make the param reference more verbose since there's already [a link to the very detailed API doc in the code](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/query.py#L275).

lukesneeringer:
How about we include the exact language that the API does? That way we can wash our hands of what the exact language should be.

roscoejp:
Pushed the whole timeout_ms reference. Lemme know if there are any other changes you'd like made to this reference.

tseaver:
@roscoejp Thanks for the fix, and your willingness to follow up!

roscoejp:
No problem. Committing to the repos isn't part of my normal routine - but we've had a few cases come in with customers getting confused about this (particular) issue so this seemed like the fastest way to get it fixed. 

Thanks for your time.

PR #3323, opened 2017-04-24
Swapping out google.streaming for google-resumable-media in GCS downloads.

dhermes:
Check out https://googlecloudplatform.github.io/google-resumable-media-python/0.0.2/ for details on the new library.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -30,6 +30,9 @@ import httplib2
 import six
 from six.moves.urllib.parse import quote
 
+import google.auth.transport.requests as tr_requests
+from google import resumable_media
+
 from google.cloud._helpers import _rfc3339_to_datetime
 from google.cloud._helpers import _to_bytes
 from google.cloud._helpers import _bytes_to_unicode
@@ -42,12 +45,13 @@ from google.cloud.storage._helpers import _scalar_property
 from google.cloud.storage.acl import ObjectACL
 from google.cloud.streaming.http_wrapper import Request
 from google.cloud.streaming.http_wrapper import make_api_request
-from google.cloud.streaming.transfer import Download
 from google.cloud.streaming.transfer import RESUMABLE_UPLOAD
 from google.cloud.streaming.transfer import Upload
 
 
 _API_ACCESS_ENDPOINT = 'https://storage.googleapis.com'
+_DOWNLOAD_URL_TEMPLATE = (
+    u'https://www.googleapis.com/download/storage/v1{path}?alt=media')
 
 
 class Blob(_PropertyMixin):
@@ -316,6 +320,24 @@ class Blob(_PropertyMixin):
         """
         return self.bucket.delete_blob(self.name, client=client)
 
+    def _get_download_url(self):
+        """Get the download URL for the current blob.
+
+        If the ``media_link`` has been loaded, it will be used, otherwise
+        the URL will be constructed from the current blob's path (and possibly
+        generation) to avoid a round trip.
+
+        :rtype: str
+        :returns: The download URL for the current blob.
+        """
+        if self.media_link is None:
+            download_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
+            if self.generation is not None:
+                download_url += u'&generation={:d}'.format(self.generation)
+            return download_url
+        else:
+            return self.media_link
+
     def download_to_file(self, file_obj, client=None):
         """Download the contents of this blob into a file-like object.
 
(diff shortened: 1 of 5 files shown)


storage/setup.py
```diff
@@ -52,6 +52,9 @@
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.24.0, < 0.25dev',
+    'google-auth',
```
jonparrott: `>=1.0.0`.

storage/setup.py
```diff
@@ -52,6 +52,9 @@
...
+    'google-resumable-media == 0.0.2',
+    'requests',
```
jonparrott: You should specify some minimum version here.

storage/google/cloud/storage/blob.py
```diff
@@ -30,6 +30,9 @@
 import six
 from six.moves.urllib.parse import quote
 
+import google.auth.transport.requests as tr_requests
```
jonparrott: I'd prefer if you didn't alias, maybe just `import google.auth.transport.requests` and then just use the full `google.auth.tranport.request.AuthorizedSession`?

storage/setup.py
dhermes: I would like to, just wasn't sure where to start. Any suggestions?

storage/google/cloud/storage/blob.py
dhermes: I say ick because

```python
>>> len('google.auth.tranport.request.AuthorizedSession')
46
```

storage/setup.py
dhermes: :+1:

storage/google/cloud/storage/blob.py
tseaver: Why not:

```python
from google.auth.transport.reqeusts import AuthorizedSession
```

storage/setup.py
jonparrott: `>=2.0.0` seems safe.

storage/google/cloud/storage/blob.py
```diff
@@ -348,28 +370,23 @@ def download_to_file(self, file_obj, client=None):
...
-        # API_BASE_URL and build_api_url).
-        download.initialize_download(request, client._base_connection.http)
+        # Create a ``requests`` transport with the client's credentials.
+        transport = tr_requests.AuthorizedSession(client._credentials)
+
+        # Download the content.
+        if self.chunk_size is None:
+            download = resumable_media.Download(download_url, headers=headers)
+            response = download.consume(transport)
+            file_obj.write(response.content)
```
tseaver: Does this slurp `response.content` into RAM?

dhermes: That's a good question. @jonparrott Do you know if `requests.Response.content` is lazy? (I'm also source-diving right now @tseaver)

storage/google/cloud/storage/blob.py
dhermes: I **try** to stick to the [Google Python style guide][1]:

> Use imports for packages and modules only.

[1]: https://google.github.io/styleguide/pyguide.html#Imports

storage/google/cloud/storage/blob.py
```diff
@@ -348,28 +370,23 @@ def download_to_file(self, file_obj, client=None):
...
-        download_url = self.media_link
-
-        # Use apitools 'Download' facility.
-        download = Download.from_stream(file_obj)
-
-        if self.chunk_size is not None:
-            download.chunksize = self.chunk_size
-
+        # Get the download URL.
+        download_url = self._get_download_url()
```
tseaver: How are you handling the properties-not-yet-populated-from-backend case?

storage/google/cloud/storage/blob.py
jonparrott: `response.content` is *not* lazy. You should use iter_content: http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow

storage/google/cloud/storage/blob.py
dhermes: See the implementation of `_get_download_url()`. Or am I misunderstanding the question?

storage/google/cloud/storage/blob.py
dhermes: @jonparrott and I just discussed this via chat and here are some conclusions:

- Using `transport.request('GET', ..., stream=True)` would make the `google-resumable-media` implementation overly specific to `requests`
- If a user is worried about how much of the download lives in RAM, they can use a `ChunkedDownload`
- People who want a `stream` for a "one-shot" download can still use a `ChunkedDownload` with the chunk size equal to "all the bytes"

storage/google/cloud/storage/blob.py
tseaver: The old code called `self.reload()` to get the back-end-supplied properties populated.  This version doesn't do that, AFAICT.

storage/google/cloud/storage/blob.py
tseaver: Could we update `resumable_media.Download(download_url, headers=headers)` to take an optional `file_obj` paramter?  If passed, the return value would be `None`, maybe?

storage/google/cloud/storage/blob.py
dhermes: That's correct, the docstring of `_get_download_url()` explains why not:

>  If the ``media_link`` has been loaded, it will be used, otherwise the URL will be constructed from the current blob's path (and possibly generation) to avoid a round trip.

storage/google/cloud/storage/blob.py
dhermes: @tseaver The interface for `Download` intentionally **does not** take a `stream` / `file`-like input. And even if it did, it would use `stream.write(response.content)` internally, so it doesn't save any RAM.

storage/google/cloud/storage/blob.py
tseaver: Seems a little straining-at-gnats-swallowing-camels, given the imports just below.

storage/google/cloud/storage/blob.py
tseaver: How do we know that the URL we are synthesizing is correct, given that the back-end has an output-only `mediaURL` property for the Object resource?

storage/google/cloud/storage/blob.py
tseaver: Seems like a footgun to me.  What about @jonparrott's comment to use `response.iter_content` (maybe with `stream.writelines()`?

storage/google/cloud/storage/blob.py
dhermes: I'm taking the [documentation][1] and the observed value of `mediaLink` for actual objects

[1]: https://cloud.google.com/storage/docs/request-endpoints

storage/google/cloud/storage/blob.py
dhermes: @tseaver Using `response.iter_content` requires using `stream=True`, to which I replied

> Using `transport.request('GET', ..., stream=True)` would make the `google-resumable-media` implementation overly specific to `requests`

dhermes:
I filed https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/5 to carry on the discussion about streaming the contents in a more memory-efficient manner (when possible).

For now, I am going to merge this based on an LGTM from @jonparrott (given in chat)

PR #3324, opened 2017-04-24
RPC retries (second PR)

calpeyser:
Since the repo's layout has changed significantly since the last PR, I've opened a new PR.  I've attempted to address the comments from the last go-around - please have a look.


dhermes:
@calpeyser The `cover` environment is also failing because several lines do not have tests exercising them / some branches may not have all paths covered:

```

google/cloud/bigtable/retry.py:    60-61, 78-120, 123, 126, 175, 45->exit, 172->175
google/cloud/bigtable/row_data.py: 279, 277->279
```

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+        )
+
+        (endpoint, port) = server.stdout.readline().rstrip("\n").split(":")
+        os.environ["BIGTABLE_EMULATOR_HOST"] = endpoint + ":" + port
+        client = Client(project="client", admin=True)
+        instance = Instance("instance", client)
+        table = instance.table("table")
+
+        # Run test, line by line
+        script = open(TEST_SCRIPT, 'r')
```
garye: This isn't looking for PASS or FAIL on stdout to see the result of the test. Let me know if I can point to an example to help with that.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,191 @@
...
+                (retry_options.backoff_settings.total_timeout_millis /
+                 _MILLIS_PER_SECOND)
+            self.set_stream()
+
+    def set_start_key(self, start_key):
+        """
+        Sets the row key at which this iterator will begin reading.
+        """
+        self.start_key = start_key
+        self.start_key_closed = False
```
garye: This works for the retry case but is not what we want generally. Is this only used for retries? If not it should default to `start_key_closed = True` and allow an open interval to be used as an override.

calpeyser: Can you clarify?  The current API to read_rows is a closed interval - there is no override for open interval.  This method handles the case of a partial response from the server, in which case the query needs to change to to an open interval starting at the last key read.

garye: I think you answered my question - I just wasn't totally sure that this method was only used in the partial response/retry case but it sounds like it is.

tseaver:
Supersedes #3279.

garye:
Hi @dhermes, can you help take a look at this?

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,168 @@
...
+from google.cloud.bigtable._generated import (
+    bigtable_pb2 as data_messages_v2_pb2)
+from google.gax import config, errors
+from grpc import RpcError
+
+
+_MILLIS_PER_SECOND = 1000
+
+
+class ReadRowsIterator():
```
lukesneeringer: Please inherit classes from `object`. (We do support Python 2.7 still, so this difference does matter.)

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,168 @@
...
+            try:
+                return six.next(self.stream)
+            except StopIteration as stop:
+                raise stop
+            except RpcError as error:  # pylint: disable=broad-except
+                code = config.exc_to_code(error)
+                if code not in self.retry_options.retry_codes:
+                    six.reraise(type(error), error)
+
+                # pylint: disable=redefined-variable-type
```
lukesneeringer: I am curious why pylint gave you trouble here; you consistently define `exc` as an `errors.RetryError`.
Regardless, this needs to stay if it was complaining. :-)

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,168 @@
...
+                to_sleep = random.uniform(0, delay * 2)
+                time.sleep(to_sleep / _MILLIS_PER_SECOND)
+                delay = min(delay * self.delay_mult, self.max_delay_millis)
+                now = time.time()
+                timeout = min(
+                    timeout * self.timeout_mult, self.max_timeout,
+                    deadline - now)
+                self.set_stream()
+
+        six.reraise(errors.RetryError, exc)
```
lukesneeringer: This will lose the traceback context. Can the third argument be set to `sys.exc_info()[2]`, please?

```python
six.reraise(errors.RetryError, exc, traceback=sys.exc_info()[2])
```

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -274,6 +274,10 @@ def consume_next(self):
 
             self._validate_chunk(chunk)
 
+            if ("ReadRowsIterator" in
```
lukesneeringer: This is kind of a bizarre check. If an unrelated library defined a different class with the same name, it would pass. Likewise, subclasses of `ReadRowsIterator` that do not include that exact string in the class name would fail.
Please either check for the presence of the method, or check the actual class.

Option 1:

```python
if hasattr(self._response_iterator, 'set_start_key'):
```

Option 2:

```python
if isinstance(self._response_iterator, ReadRowsIterator):
```

If there is some reason that you really need the check in this specific form that is not covered by either of the two options above, then please elaborate. :-)

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+
+        # Download server
+        MOCK_SERVER_URLS = {
+            'Linux': 'https://storage.googleapis.com/cloud-bigtable-test/retries/retry_server_linux.tar.gz',
+            'Darwin': 'https://storage.googleapis.com/cloud-bigtable-test/retries/retry_server_mac.tar.gz',
+        }
+
+        test_platform = platform.system()
+        if (test_platform not in MOCK_SERVER_URLS):
+            self.fail("Retry server not available for platform " + test_platform)
```
lukesneeringer: This will make the tests flatly fail on Windows (and therefore AppVeyor), or any other unrecognized platform. This seems harsh. Can we just _skip_ the test instead?

```python
if test_platform not in MOCK_SERVER_URLS:
    self.skip('Retry server not available for platform {0}.'.format(test_platform))
```

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+
+        mock_server_download = urlopen(MOCK_SERVER_URLS[test_platform]).read()
```
lukesneeringer: I am not a big fan of downloading an emulator on the fly, but maybe it is better than any other option?

bigtable/tests/system.py
lukesneeringer: Use the context manager form, please:

```python
with io.open(TEST_SCRIPT, 'r') as script:
```

This ensures the file is necessarily closed.

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+        instance = Instance("instance", client)
+        table = instance.table("table")
+
+        # Run test, line by line
+        script = open(TEST_SCRIPT, 'r')
+        for line in script.readlines():
+            if line.startswith("CLIENT:"):
+                chunks = line.split(" ")
+                op = chunks[1]
+                if (op != "SCAN"):
```
lukesneeringer: Why is this necessary? The test script is checked into the repository; it is static. There is no reason for this check to ever fail.

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+                    self.fail("Script contained " + op + " operation.  Only \'SCAN\' is supported.")
+                else:
+                    process_scan(table, chunks[2], chunks[3])
+
+        # Clean up
+        server.kill()
```
lukesneeringer: These three lines should be in a `finally` clause.

bigtable/tests/system.py
```diff
@@ -295,6 +295,76 @@ def test_delete_column_family(self):
...
+        # Run test, line by line
+        script = open(TEST_SCRIPT, 'r')
+        for line in script.readlines():
+            if line.startswith("CLIENT:"):
+                chunks = line.split(" ")
+                op = chunks[1]
+                if (op != "SCAN"):
+                    self.fail("Script contained " + op + " operation.  Only \'SCAN\' is supported.")
+                else:
+                    process_scan(table, chunks[2], chunks[3])
```
lukesneeringer: At what point do you check the results to ensure you get what you expect?

bigtable/tests/unit/test_table.py
```diff
@@ -396,9 +395,166 @@ def mock_create_row_request(table_name, **kwargs):
...
+            mock_created.append((table_name, kwargs))
+            return request_pb
+
+        # Create a slow response iterator to cause a timeout
+        class MockTimeoutError(RpcError):
+            def code(self):
+                return StatusCode.DEADLINE_EXCEEDED
+
+        def _wait_then_raise():
+            time.sleep(0.5)
```
lukesneeringer: Why is there a `sleep` in a unit test? This makes the test take half a second, which is actually a really long time.

bigtable/tests/system.py
garye: This was my suggestion and I agree it's not great. Could it instead be something that people who want to run the tests have to install the fake server and then run the test with a specific option to use it? Maybe that doesn't work with CI?

lukesneeringer: How big are these? Could we just include the tarball in source? (There are reasons not to do _that_, too...)

garye: A bit less than 4MB each. No reason we couldn't include them that I can see, but that's a fair amount of bloat added to the source and I wasn't sure that was desirable.

bigtable/tests/system.py
calpeyser: I'm struggling to implement this - can you send me an example?

garye: Sure, I wish I could help more with the python-specific stuff but here's a java example:

https://gist.github.com/garye/68c2d373bc670a8c24bdf49f7f53e282

Basically there's a thread that reads from STDOUT until it sees PASS OR FAIL.
The other thread runs through the test case and then waits for the notification that pass/fail has been read and fails the test if it was FAIL.

bigtable/tests/system.py
calpeyser: should the whole test be in a try clause?

bigtable/tests/unit/test_table.py
calpeyser: It must sleep so that the request times out, as the test requires.  I've trimmed it down to 0.1 seconds - is that better?

calpeyser:
@lukesneeringer @garye - I've attempted to address the comments above.  Thanks for bearing with me!

mbrukman:
@calpeyser -- looks like you have some conflicts with the base branch, please resolve them.

@lukesneeringer, @garye -- please take a look at the PR, it would be great to merge this functionality soon.

Thanks everyone!

bigtable/google/cloud/bigtable/retry.py
calpeyser: it was indeed complaining :)

garye:
LGTM from a bigtable perspective

calpeyser:
Conflict resolved, tests are passing.  @lukesneeringer, what do you think?

lukesneeringer:
Reading now

calpeyser:
@lukesneeringer - can you merge it? Neither I nor @garye have write permissions.

lukesneeringer:
:tada:

mbrukman:
Thank you @calpeyser for your work on this and thanks to @garye and @lukesneeringer for the reviews!

dhermes:
This caused a system test breakage, I'm very tempted to roll it back and iterate in a PR / branch until the system tests are passing

PR #3336, opened 2017-04-27
Add basic system test for error reporting

### LINKED ISSUE #3335: Error reporting system tests needed
Follow up to #3263.

waprin:
This is the bare minimal system test that just makes a client call, it did go red before the scope fix at least.

Can add more asserts but need to add more code to authenticate a separate gax client to list errors, can add that to this PR or a new one.

Fixes #3335

### DIFF
--- a/error_reporting/nox.py
+++ b/error_reporting/nox.py
@@ -63,6 +63,25 @@ def lint_setup_py(session):
     session.run(
         'python', 'setup.py', 'check', '--restructuredtext', '--strict')
 
+@nox.session
+@nox.parametrize('python_version', ['2.7', '3.6'])
+def system_tests(session, python_version):
+    """Run the system test suite."""
+
+    # Sanity check: Only run system tests if the environment variable is set.
+    if not os.environ.get('GOOGLE_APPLICATION_CREDENTIALS', ''):
+        return
+
+    # Run the system tests against latest Python 2 and Python 3 only.
+    session.interpreter = 'python{}'.format(python_version)
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install('.')
+
+    # Run py.test against the system tests.
+    session.run('py.test', '-vvv', 'tests/system.py')
+
 
 @nox.session
 def cover(session):
--- a/error_reporting/tests/system.py
+++ b/error_reporting/tests/system.py
@@ -1,0 +1,27 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


error_reporting/nox.py
```diff
@@ -63,6 +63,25 @@ def lint_setup_py(session):
...
+
+    # Run the system tests against latest Python 2 and Python 3 only.
+    session.interpreter = 'python{}'.format(python_version)
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install('.')
+
+    # Run py.test against the system tests.
+    session.run('py.test', '-vvv', 'tests/system.py')
```
dhermes: `-vvv` just for fun? (We have it **only** in logging because there was a strange hanging error that we were trying to diagnose.)

error_reporting/nox.py
```diff
@@ -63,6 +63,25 @@ def lint_setup_py(session):
...
+    # Sanity check: Only run system tests if the environment variable is set.
+    if not os.environ.get('GOOGLE_APPLICATION_CREDENTIALS', ''):
+        return
+
+    # Run the system tests against latest Python 2 and Python 3 only.
+    session.interpreter = 'python{}'.format(python_version)
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install('.')
```
dhermes: @lukesneeringer Can you confirm that we **don't** want editable installs for system tests?

error_reporting/tests/system.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestErrorReporting(unittest.TestCase):
+
+    def test_report_exception(self):
+        client = error_reporting.Client()
+        try:
+            raise NameError
+        except NameError:
+            client.report_exception()
```
dhermes: @waprin This is insufficient.

You need to make a query to retrieve the reported error and then verify chunks of it (I don't know how difficult that is).

Also, let's push this branch to the https://github.com/GoogleCloudPlatform/google-cloud-python/ remote so that the system tests run (you can do that without closing this PR, but if you'd like to close it and open a new one from the branch in this remote, feel free)

error_reporting/nox.py
waprin: Ah ok.

error_reporting/tests/system.py
waprin: Yes, I can use the gax client to do it, though that makes it so the test will only work with gRPC (unless I added an HTTP client only for the purpose of a system test, the current non-gRPC approach uses Logging which has no means of getting the errors).

I didn't add it yet in the spirit of small PRs, can add it on this one or a new one (sounds like you prefer this one).

dhermes: Yes I prefer this one. A single test case is small enough to fit in the spirit of small PRs. You can add a skip ([e.g.][1]) so that the test only runs for gRPC.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/df30d1aa9fddacdc14f6c2412feb445e9d459e95/speech/tests/system.py#L212-L213

dhermes:
Closing this in light of #3348

PR #3337, opened 2017-04-27
Allowing "dank" or "danke" when "dankon" translated from Esperanto to German.

dhermes:
This test has been non-deterministic for some reason, e.g.:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1682


dhermes:
Oops I merged before an explicit LGTM, but I **swear** I got one from @lukesneeringer in a Google Hangout

PR #3339, opened 2017-04-27
Adding retry to test_detect_web_images_from_gcs() system test.

dhermes:
Flaky in:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1694

### DIFF
--- a/vision/tests/system.py
+++ b/vision/tests/system.py
@@ -14,6 +14,7 @@
 
 """System tests for Vision API."""
 
+import functools
 import os
 import unittest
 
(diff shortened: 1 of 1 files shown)


vision/tests/system.py
```diff
@@ -720,6 +720,7 @@ def test_detect_web_images_from_content(self):
         web_images = image.detect_web(limit=limit)
         self._assert_web_images(web_images, limit)
 
+    @RetryErrors(unittest.TestCase.failureException)
```
dhermes: Note this approach is already used on line 714 above

tseaver: I can't say I care much for the approach:  it smells like papering over a set of bad assertions.

dhermes:
@tseaver PTAL. I made the retry more focused (rather than retrying the entire test case).

PR #3348, opened 2017-05-01
Add error reporting system test

### LINKED ISSUE #3335: Error reporting system tests needed

waprin:
This actually uses the gapic client to verify counts.

It doesn't expose the gapic client outside of the system methods since unlikely they will need to be used (and we can wait on the general gapic doc improvements for those who do want it).

Fixes #3335.

### DIFF
--- a/error_reporting/nox.py
+++ b/error_reporting/nox.py
@@ -78,7 +78,9 @@ def system_tests(session, python_version):
 
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
-    session.install('flake8', *LOCAL_DEPS)
+    session.install('mock', 'pytest', *LOCAL_DEPS)
+    session.install('../test_utils/', '../bigquery/', '../pubsub/',
+                '../storage/')
     session.install('.')
 
     # Run py.test against the system tests.
--- a/error_reporting/nox.py
+++ b/error_reporting/nox.py
@@ -63,6 +63,7 @@ def lint_setup_py(session):
     session.run(
         'python', 'setup.py', 'check', '--restructuredtext', '--strict')
 
+
 @nox.session
 @nox.parametrize('python_version', ['2.7', '3.6'])
 def system_tests(session, python_version):
@@ -77,6 +78,7 @@ def system_tests(session, python_version):
 
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
+    session.install('flake8', *LOCAL_DEPS)
     session.install('.')
 
     # Run py.test against the system tests.
--- a/error_reporting/nox.py
+++ b/error_reporting/nox.py
@@ -63,6 +63,25 @@ def lint_setup_py(session):
     session.run(
         'python', 'setup.py', 'check', '--restructuredtext', '--strict')
 
+@nox.session
+@nox.parametrize('python_version', ['2.7', '3.6'])
+def system_tests(session, python_version):
+    """Run the system test suite."""
+
+    # Sanity check: Only run system tests if the environment variable is set.
+    if not os.environ.get('GOOGLE_APPLICATION_CREDENTIALS', ''):
+        return
+
+    # Run the system tests against latest Python 2 and Python 3 only.
+    session.interpreter = 'python{}'.format(python_version)
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install('.')
+
+    # Run py.test against the system tests.
+    session.run('py.test', '-vvv', 'tests/system.py')
+
 
 @nox.session
 def cover(session):
--- a/error_reporting/nox.py
+++ b/error_reporting/nox.py
@@ -79,8 +79,7 @@ def system_tests(session, python_version):
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
     session.install('mock', 'pytest', *LOCAL_DEPS)
-    session.install('../test_utils/', '../bigquery/', '../pubsub/',
-                '../storage/')
+    session.install('../test_utils/')
     session.install('.')
 
     # Run py.test against the system tests.
--- a/error_reporting/tests/system.py
+++ b/error_reporting/tests/system.py
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import time
+import functools
+import operator
 import unittest
 
 from google.cloud import error_reporting
@@ -22,8 +23,12 @@ from google.cloud.proto.devtools.clouderrorreporting.v1beta1 import (
     error_stats_service_pb2)
 from google.protobuf.duration_pb2 import Duration
 
+from test_utils.retry import RetryResult
+from test_utils.system import unique_resource_id
 
-ERROR_NAME = 'Stackdriver Error Reporting System Test'
+
+ERROR_MSG = 'Stackdriver Error Reporting System Test'
+
 
 def setUpModule():
     Config.CLIENT = error_reporting.Client()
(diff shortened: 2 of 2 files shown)


error_reporting/tests/system.py
```diff
@@ -0,0 +1,96 @@
...
+def _is_incremented(initial, new):
+    """Helper to retry until new error is counted."""
+    return new == initial + 1
+
+
+class TestErrorReporting(unittest.TestCase):
+
+    def setUp(self):
+        self._client = error_reporting.Client()
+        self._error_name = 'Stackdriver Error Reporting System Test'
```
dhermes: @waprin We don't have much (any?) use of `setUp` in this library and it's intentional: we try to avoid accidentally shared state across tests.

You can make `ERROR_NAME` a global and can follow our typical pattern with `Config.CLIENT` for the client?

error_reporting/tests/system.py
```diff
@@ -0,0 +1,96 @@
...
+            if self._error_name in group.representative.message:
+                return group.count
+
+    def test_report_exception(self):
+        """Verifies the exception reported increases the group count by one."""
+        # If test has never run, group won't exist until we report first
+        # exception, so first simulate it just to create the group
+        self._simulate_exception()
+
+        initial_count = self._get_error_count()
```
dhermes: ISTM that we need to retry `_get_error_count` below because the query is eventually consistent, so why do we expect it to work just fine right here?

error_reporting/tests/system.py
```diff
@@ -0,0 +1,96 @@
...
+    def _get_error_count(self):
+        """Counts the number of errors in the group of the test exception."""
+        error_stats_api = _ErrorStatsGaxApi(self._client.project)
+        groups = error_stats_api.list_groups()
+        for group in groups:
+            if self._error_name in group.representative.message:
+                return group.count
+
+    def test_report_exception(self):
+        """Verifies the exception reported increases the group count by one."""
```
dhermes: Can you remove this docstring? It mucks with the `nose` output (though we use `py.test` as test runner)

error_reporting/tests/system.py
```diff
@@ -0,0 +1,96 @@
...
+from google.cloud.gapic.errorreporting.v1beta1 import (
+    error_stats_service_client)
+from google.cloud.proto.devtools.clouderrorreporting.v1beta1 import (
+    error_stats_service_pb2)
+from google.protobuf.duration_pb2 import Duration
+
+from test_utils.retry import RetryResult
+
+
+class _ErrorStatsGaxApi(object):
```
dhermes: Why not just add this to the module? Or just write a **function** (not a class) to make the request we want?

error_reporting/tests/system.py
```diff
@@ -0,0 +1,96 @@
...
+    This class provides a small wrapper around making calls to the GAX
+    API. It's used by the system tests to find the appropriate error group
+    to verify the error was successfully reported.
+
+    :type project: str
+    :param project: Google Cloud Project ID
+    """
+    def __init__(self, project):
+        self._project = project
+        self._gax_api = error_stats_service_client.ErrorStatsServiceClient()
```
dhermes: This should also be passed our client.

error_reporting/tests/system.py
waprin: Switched it to a function. I'm hesitant to add it to the module because it's not really a core use of the API, just a helper for a system test, and so I don't want to bloat the module. As GAPIC docs improve people can just rely on the GAPIC itself it they really want to do this stuff.

error_reporting/tests/system.py
waprin: Not sure exactly what you mean.

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: Wow, really good point I totally missed.

In general the eventual consistency makes things confusing, but I think in this case, I only saw it flake a few times and a small sleep fixes it. So I just do a single time.sleep(2) after each report exception. Think that will be fine and not flakey.

error_reporting/google/cloud/error_reporting/_gax.py
```diff
@@ -19,6 +19,7 @@
 
 from google.cloud.gapic.errorreporting.v1beta1 import (
     report_errors_service_client)
+
```
dhermes: Editorializing?

error_reporting/nox.py
```diff
@@ -77,6 +78,7 @@ def system_tests(session, python_version):
 
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
+    session.install('flake8', *LOCAL_DEPS)
```
dhermes: This seems "correct' though you should dump `flake8`?

waprin:
@dhermes still no hints as to why system tests are failing. I don't see the SSH option, wonder if you have any ideas why.

waprin:
@dhermes mergeable?

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+        groups = _list_groups(Config.CLIENT.project)
+        for group in groups:
+            if ERROR_NAME in group.representative.message:
+                return group.count
+
+    def test_report_exception(self):
+        # If test has never run, group won't exist until we report first
+        # exception, so first simulate it just to create the group
+        self._simulate_exception()
+        time.sleep(2)
```
dhermes: :-1: on the `sleep`, please use the `retry` helpers we have for system tests.

(A `sleep` is just a recipe for a flaky system test.)

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+class TestErrorReporting(unittest.TestCase):
+
+    def _simulate_exception(self):
+        """Simulates an exception to verify it was reported."""
+        try:
+            raise RuntimeError(ERROR_NAME)
+        except RuntimeError:
+            Config.CLIENT.report_exception()
+
+    def _get_error_count(self):
```
dhermes: This never uses `self`, it could be a standalone function or a `@staticmethod`

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+    """List Error Groups from the last 60 seconds.
+
+    This class provides a wrapper around making calls to the GAX
+    API. It's used by the system tests to find the appropriate error group
+    to verify the error was successfully reported.
+
+    :type project: str
+    :param project: Google Cloud Project ID
+    """
+    gax_api = error_stats_service_client.ErrorStatsServiceClient()
```
dhermes: You need to pass `credentials=Config.CLIENT._credentials` here

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+    project_name = gax_api.project_path(project)
+
+    time_range = error_stats_service_pb2.QueryTimeRange()
+    time_range.period = (
```
dhermes: FYI

```python
    time_range.period = error_stats_service_pb2.QueryTimeRange.PERIOD_1_HOUR
```

is only 76 chars

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+        error_stats_service_pb2.QueryTimeRange.PERIOD_1_HOUR
+    )
+
+    duration = Duration()
+    duration.seconds = 60 * 60
```
dhermes: Recommend:

```python
duration = Duration(seconds=3600)
```

(you COULD do `60 * 60` if you like, but it's not necessary)

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+
+    return gax_api.list_group_stats(
+        project_name, time_range, timed_count_duration=duration)
+
+ERROR_NAME = 'Stackdriver Error Reporting System Test'
```
dhermes: Plus put this at the top with all globals

error_reporting/tests/system.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class TestErrorReporting(unittest.TestCase):
+
+    def _simulate_exception(self):
```
dhermes: This never uses `self`, it could be a standalone function or a `@staticmethod`

error_reporting/tests/system.py
waprin: I understand the concern but in this case I don't think the test will be flaky, the error reporting system is a bit quicker, without the sleep I only saw it fail once. The problem with the retry helpers is there's no reasonable condition to retry on. You don't know what the initial count is supposed to be.

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: done

error_reporting/tests/system.py
waprin: done

waprin:
lgtm

PR #3352, opened 2017-05-02
Only using Blob.updated if it is set in download_to_filename().

### LINKED ISSUE #3342: Error when downloading blobs with storage v1.1.0
See build log: https://www.google.com/url?q=https%3A%2F%2Ftravis-ci.org%2FGoogleCloudPlatform%2Fpython-docs-samples%2Fbuilds%2F226935178%23L786&sa=D&sntz=1&usg=AFQjCNG3mWBi07AXjvj70yKUdQ0pI4hJNw

dhermes:
Fixes #3342.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -432,8 +432,10 @@ class Blob(_PropertyMixin):
         with open(filename, 'wb') as file_obj:
             self.download_to_file(file_obj, client=client)
 
-        mtime = time.mktime(self.updated.timetuple())
-        os.utime(file_obj.name, (mtime, mtime))
+        updated = self.updated
+        if updated is not None:
+            mtime = time.mktime(updated.timetuple())
+            os.utime(file_obj.name, (mtime, mtime))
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -432,8 +432,10 @@ def download_to_filename(self, filename, client=None):
         with open(filename, 'wb') as file_obj:
             self.download_to_file(file_obj, client=client)
 
-        mtime = time.mktime(self.updated.timetuple())
-        os.utime(file_obj.name, (mtime, mtime))
+        updated = self.updated
```
jonparrott: No change needed, but wondering why you've assigned this to a temporary variable instead of just referencing `self.updated`?

dhermes: Because `self.updated` is a `@property` does a dictionary lookup and a string->native datetime conversion on every call. So calling it twice would be "double computation". (I figured someone would ask.)

jonparrott:
gotcha.

PR #3353, opened 2017-05-02
Bugfix release of Storage.

dhermes:
Release notes will be simple

## google-cloud-storage 1.1.1

- Bug fix: Only using `Blob.updated` if it is set in `download_to_filename()`. (#3352)

/cc @danigosa

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-storage',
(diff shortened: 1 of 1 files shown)


dhermes:
Here is the tag build

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1713

which will push to 

https://pypi.python.org/pypi/google-cloud-storage/1.1.1

Issue #2997, opened 2017-02-10
Stop writing to `/var/log/app_engine/`


JustinBeckwith:
Currently the logging library is writing files to a magical path:  `/var/log/app_engine/`.  This is not an official supported path, and would only work for App Engine.  At any point -that path could stop working. 

The only officially supported ways to write logs are to write to stdout or stderr, or to use the cloud logging APIs.  Come chat with me if you have questions!

[https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/handlers/app_engine.py#L34][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9619004959def22ad2084691f382dc1cada46afd/logging/google/cloud/logging/handlers/app_engine.py#L34

jonparrott:
@waprin can you take this?

waprin:
@JustinBeckwith it's ok if it only works on App Engine because it's only supposed to be used on App Engine. 

stdout/stderr don't capture severity which this was supposed to address.

We can switch it to use the API directly, just need to make sure correct monitored resource gets set.

lukesneeringer:
@waprin Can you provide me with some clarity on this? Does this need to be fixed, or is it working as intended?

waprin:
@lukesneeringer 

tl;dr: It's working, but via an unofficially supported mechanism, and so it needs to be fixed, which I can do.

Longer story:

These are for the Python standard logging handlers, which is not the main API interface we document, it's just a helper that I added. I documented  it under "Additional tools" in the Cloud docs and a section at the bottom of these Github docs.

As mentioned, on App Engine Flex, stdout/stderr ends up in Stackdriver Logging, but the reason I just don't like logging to that (which is where Python standard logging goes by default), is because the fluentd logger doesn't capture error severity levels. So if you try to search for messages logged at severity `error` the best you can do is a text search for the word `error`, and you will get all message that have the word error in them regardless of their log level. It's a small detail but I think a nice one to get right. So the current handler tries to format the messages so the fluentd agent parses the severity correctly. Justin is telling me that I shouldn't rely on the fluentd agent at all and should just write directly to the API.

The only small details to doing that right is to correctly set the monitored resource so the logs show up in the correct App Engine service/version logs rather than in custom global logs. 

[...]

PR #3355, opened 2017-05-02
Add trace_id to log payload to enable log aggregation in GAE Flex

liyanhui1228:
Extract trace_id from HTTP request header and add it to application log. The pantheon UI will show the app logs grouped by trace_id and show them aggregated in one request.

Below is the screenshot of the aggregated log:
screenshot from 2017-05-01 17 37 44

### DIFF
--- a/logging/google/cloud/logging/handlers/_helpers.py
+++ b/logging/google/cloud/logging/handlers/_helpers.py
@@ -16,6 +16,7 @@
 
 import math
 import json
+import os
 
 
 def format_stackdriver_json(record, message):
@@ -26,6 +27,7 @@ def format_stackdriver_json(record, message):
     """
     subsecond, second = math.modf(record.created)
 
+    # Assemble a dictionary with the log string as the message.
     payload = {
         'message': message,
         'timestamp': {
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
         'severity': record.levelname,
     }
 
+    # Make a best effort to add the trace id.
+    # First try to extract the trace_id from HTTP header, only support flask framework for now.
+    trace_id = parse_flask_request_header()
+
+    # If that didn't work, try to get the trace_id from the 'extras' of the record.
+    if not trace_id:
+        trace_id = getattr(record, 'trace_id', None)
+
+    # If still cannot get the trace_id, try to get it from environment.
+    if not trace_id:
+        trace_id = os.getenv('HTTP_X_CLOUD_TRACE_CONTEXT', '').split('/')[0]
+
+    # Add trace_id to payload if it was found.
+    if trace_id:
+        payload['traceId'] = trace_id
+
     return json.dumps(payload)
+
+
+def parse_flask_request_header():
+    """Parse flask request header to get trace_id.
+    
+        :rtype: str
+        :return: trace_id get from flask header
+    """
+    # Import the web framework to parse HTTP header.
+    try:
+        import flask
+    except ImportError:
+        flask = None
+
+    # Extract the trace_id
+    if flask:
+        try:
+            trace_id = flask.request.headers['X-Cloud-Trace-Context'].split('/')[0]
+        except KeyError:
+            trace_id = None
+    else:
+        trace_id = None
+
+    return trace_id


logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
...
+    # Extract the trace_id
+    if flask:
+        try:
+            trace_id = flask.request.headers['X-Cloud-Trace-Context'].split('/')[0]
+        except KeyError:
+            trace_id = None
+    else:
+        trace_id = None
+
+    return trace_id
```
dhermes: Please end your file with a newline

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
...
+        :return: trace_id get from flask header
+    """
+    # Import the web framework to parse HTTP header.
+    try:
+        import flask
+    except ImportError:
+        flask = None
+
+    # Extract the trace_id
+    if flask:
```
dhermes: `if flask is not None`

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
...
+    """Parse flask request header to get trace_id.
+    
+        :rtype: str
+        :return: trace_id get from flask header
+    """
+    # Import the web framework to parse HTTP header.
+    try:
+        import flask
+    except ImportError:
+        flask = None
```
dhermes: Please do this at module-scope

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
...
+        payload['traceId'] = trace_id
+
     return json.dumps(payload)
+
+
+def parse_flask_request_header():
+    """Parse flask request header to get trace_id.
+    
+        :rtype: str
+        :return: trace_id get from flask header
```
dhermes: Should de-dent these to be at the same indent level as everything else

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -36,4 +38,44 @@ def format_stackdriver_json(record, message):
         'severity': record.levelname,
     }
 
+    # Make a best effort to add the trace id.
+    # First try to extract the trace_id from HTTP header, only support flask framework for now.
+    trace_id = parse_flask_request_header()
```
dhermes: I am big time :-1: on putting a **flask-specific** check in a generic method.

A better fix would probably involve allowing a way to register extra checkers, either through a registry in this module or through a registry on a `Client()` object (or a similar type).

waprin:
This whole PR should get closed, we are just supposed to use the generic log handler according to @justinbeckwith in #2997

waprin:
Maybe we can still get the app ID in there but it shouldn't be via the stackdriver helper that currently exists.

dhermes:
@waprin Can we make a decision quickly and either move forward or close this thing? (i.e. just don't want it to come off our plates and sit open in limbo)

waprin:
I think we should close and @liyanhui1228 and @duggelz should have a quick chat about best way to do this.

logging/google/cloud/logging/handlers/_helpers.py
duggelz: Adding Flask-specific, Django-specific, webapp2-specific, pycharm-specific, etc code to support the silver language contract has always been my understanding of the plan.  The exact implementation undetermined, and so now is a good time to talk about it.  Do you have access to the "Stackdriver: Silver Language Requirements" document?  Specifically the "Runtimes: Logging" section.  Shall I setup a meeting to discuss, or take it to email?

dhermes: @duggelz The issue is not **THAT** we provide purpose-built helpers, just **WHERE / HOW**. (Hence my suggestion for a better fix.)

dhermes:
OK, closing now. We can always re-open.

duggelz:
This has nothing to do with #2997

jonparrott:
Okay - deep breath - we failed to properly support a new contributor here. We need to do better.

It's clear that while we *want* this feature, we're not ready to implement it. So, let's create a bug/feature request for this feature and mark all blocking bugs. If the owners of those bugs can't resolve them in a timely manner, let's see what we can do to enable @liyanhui1228 or @duggelz to address them so that this feature can be implemented.

@dhermes or @lukesneeringer, can you own this? If not, please let me know and I will do this personally.

lukesneeringer:
I am happy to own it if needed ("buck stops here" kind of thing), but I am not sure I am the right person. I have read over this ticket and I do not intuitively grok the issue in question.

I _think_ the issue is that if there is an HTTP header that we can easily get at, we want to send it to Stackdriver in some way?

Best case for me is if @jonparrott owns this, but I am willing to if needed. :-)

jonparrott:
@lukesneeringer I'm on it.

PR #3357, opened 2017-05-02
Factoring out some Blob helpers.

dhermes:
This is prep work for swapping out the upload implementation to use `google-resumable-media`.

I just wanted to get "unrelated" and non-controversial changes out of the way before a big PR.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -27,7 +27,6 @@ import os
 import time
 
 import httplib2
-import six
 from six.moves.urllib.parse import quote
 
 import google.auth.transport.requests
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -324,6 +325,23 @@ def delete(self, client=None):
         """
         return self.bucket.delete_blob(self.name, client=client)
 
+    def _make_transport(self, client):
```
jonparrott: Can this go away when we drop httplib2?

storage/google/cloud/storage/blob.py
```diff
@@ -50,8 +49,10 @@
 
 
 _API_ACCESS_ENDPOINT = 'https://storage.googleapis.com'
+_DEFAULT_CONTENT_TYPE = u'application/octet-stream'
 _DOWNLOAD_URL_TEMPLATE = (
     u'https://www.googleapis.com/download/storage/v1{path}?alt=media')
+_CONTENT_TYPE = 'contentType'
```
jonparrott: I'd prefer this name be more descriptive, or at least have a comment.

storage/google/cloud/storage/blob.py
dhermes: Yes it can!

storage/google/cloud/storage/blob.py
dhermes: I promise to put it in a follow-up PR. Pinky swear!

PR #3362, opened 2017-05-03
Changing over Blob.upload*() methods to use google-resumable-media.

dhermes:
Some notes:

- ~~For now, the [build][1] is failing unit tests, because I haven't done those (what I'm sure will be very time-consuming) updates~~
- The **ACTUAL IMPORTANT** thing to note is that the system tests in the [build][1] **are passing**
- ~~I haven't yet updated `create_resumable_upload_session()`, but I can (it should be fairly straightforward, but I'd rather do it in a separate PR)~~
- I would **really** like some suggestions about what to do about the `num_retries` argument. IMO supporting it is a "bad idea™"
- I have ever-so-slightly changed the behavior of `upload_from_file()` when `size` is passed in. This is mostly because the existing implementation is **confused** about what the size / `chunk_size` combination should mean. The implementation now does the "sane" thing: use a resumable upload **IF AND ONLY IF** there is a chunk size specified on the blob
- I have also dropped the ["extra" size check][2] from the implementation and moved an actual `fstat` [into `upload_from_filename()`][3]. This size check never really made sense in the generic "give me an `IO[bytes]` and I'll stream it" method. What's more, a resumable upload works perfectly fine if the size isn't known, so there is no good reason to add in that extra check.
- Previously, if the `chunk_size` was unset (i.e. a simple upload), then `blob.upload_from_file` would completely fail if `size` was not passed in. This is not necessary, since we could just do `file_obj.read()` when there is no size specified
- I would really like to deprecate the `rewind` keyword argument (related #3365)

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1728
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4e6e2b13e7f4fa092c772e1dd4a1a976f4280713/storage/google/cloud/storage/blob.py#L690-L695
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/8b1ae6dd9ba59fc0568e370cc6bc64a562855818/storage/google/cloud/storage/blob.py#L953

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -20,17 +20,20 @@ import base64
 import copy
 import hashlib
 from io import BytesIO
-from io import UnsupportedOperation
-import json
 import mimetypes
 import os
 import time
+import warnings
 
 import httplib2
 from six.moves.urllib.parse import quote
 
 import google.auth.transport.requests
 from google import resumable_media
+from google.resumable_media.requests import ChunkedDownload
+from google.resumable_media.requests import Download
+from google.resumable_media.requests import MultipartUpload
+from google.resumable_media.requests import ResumableUpload
 
 from google.cloud._helpers import _rfc3339_to_datetime
 from google.cloud._helpers import _to_bytes
@@ -43,7 +46,6 @@ from google.cloud.storage._helpers import _PropertyMixin
 from google.cloud.storage._helpers import _scalar_property
 from google.cloud.storage.acl import ObjectACL
 from google.cloud.streaming.http_wrapper import Request
-from google.cloud.streaming.http_wrapper import make_api_request
 from google.cloud.streaming.transfer import RESUMABLE_UPLOAD
 from google.cloud.streaming.transfer import Upload
 
@@ -52,7 +54,35 @@ _API_ACCESS_ENDPOINT = 'https://storage.googleapis.com'
 _DEFAULT_CONTENT_TYPE = u'application/octet-stream'
 _DOWNLOAD_URL_TEMPLATE = (
     u'https://www.googleapis.com/download/storage/v1{path}?alt=media')
-_CONTENT_TYPE = 'contentType'
+_BASE_UPLOAD_TEMPLATE = (
+    u'https://www.googleapis.com/upload/storage/v1{bucket_path}/o?uploadType=')
+_MULTIPART_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u'multipart'
+_RESUMABLE_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u'resumable'
+# NOTE: "acl" is also writeable but we defer ACL management to
+#       the classes in the google.cloud.storage.acl module.
+_CONTENT_TYPE_FIELD = 'contentType'
+_WRITABLE_FIELDS = (
+    'cacheControl',
+    'contentDisposition',
+    'contentEncoding',
+    'contentLanguage',
+    _CONTENT_TYPE_FIELD,
+    'crc32c',
+    'md5Hash',
+    'metadata',
+    'name',
+    'storageClass',
+)
+_NUM_RETRIES_MESSAGE = (
+    'num_retries is no longer supported. When a transient error occurs, '
+    'such as a 429 Too Many Requests or 500 Internal Server Error, upload '
+    'requests will be automatically retried. Subsequent retries will be '
+    'done after waiting 1, 2, 4, 8, etc. seconds (exponential backoff) until '
+    '10 minutes of wait time have elapsed. At that point, there will be no '
+    'more attempts to retry.')
+_READ_LESS_THAN_SIZE = (
+    'Size {:d} was specified but the file-like object only had '
+    '{:d} bytes remaining.')
 
 
 class Blob(_PropertyMixin):
@@ -381,11 +411,11 @@ class Blob(_PropertyMixin):
         :param headers: Optional headers to be sent with the request(s).
         """
         if self.chunk_size is None:
-            download = resumable_media.Download(download_url, headers=headers)
+            download = Download(download_url, headers=headers)
             response = download.consume(transport)
             file_obj.write(response.content)
         else:
-            download = resumable_media.ChunkedDownload(
+            download = ChunkedDownload(
                 download_url, self.chunk_size, file_obj, headers=headers)
             while not download.finished:
                 download.consume_next_chunk(transport)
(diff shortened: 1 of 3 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -613,8 +640,213 @@ def _check_response_error(request, http_response):
...
+        * ``name``
+        * ``storageClass``
+
+        For now, we don't support ``acl``, access control lists should be
+        managed directly through :class:`ObjectACL` methods.
+
+        .. _API reference: https://cloud.google.com/storage/\
+                           docs/json_api/v1/objects
+        """
+        # NOTE: This assumes `self.name` is unicode.
```
jonparrott: Is this a safe assumption?

storage/google/cloud/storage/blob.py
```diff
@@ -613,8 +640,213 @@ def _check_response_error(request, http_response):
...
+        :rtype: tuple
+        :returns: A quadruple of
+
+                  * An
+                    :class:`~google.auth.transport.requests.AuthorizedSession`
+                  * A header dictionary
+                  * An object metadata dictionary
+                  * The ``content_type`` as a string (according to precedence)
+        """
+        transport = self._make_transport(client)
```
jonparrott: having this method create the transport seems over-reaching.

storage/google/cloud/storage/blob.py
```diff
@@ -613,8 +640,213 @@ def _check_response_error(request, http_response):
...
+        headers = _get_encryption_headers(self._encryption_key)
+        object_metadata = self._get_writable_metadata()
+        content_type = self._get_content_type(content_type)
+        return transport, headers, object_metadata, content_type
+
+    def _do_multipart_upload(self, client, stream, content_type, size):
+        """Perform a multipart upload.
+
+        Assumes ``chunk_size`` is :data:`None` on the current blob.
```
jonparrott: Is this a safe assumption? Should it be an assertion?

storage/google/cloud/storage/blob.py
```diff
@@ -613,8 +640,213 @@ def _check_response_error(request, http_response):
...
+        info = self._get_upload_arguments(client, content_type)
+        transport, headers, object_metadata, content_type = info
+
+        upload_url = _RESUMABLE_URL_TEMPLATE.format(
+            bucket_path=self.bucket.path)
+        upload = ResumableUpload(upload_url, self.chunk_size, headers=headers)
+        upload.initiate(
+            transport, stream, object_metadata, content_type,
+            total_bytes=size, stream_final=False)
+        while not upload.finished:
```
jonparrott: super nit: I prefer one blank line above independent control statements.

storage/google/cloud/storage/blob.py
```diff
@@ -751,16 +943,17 @@ def upload_from_filename(self, filename, content_type=None, client=None):
...
-                      ``NoneType``
-        :param client: Optional. The client to use.  If not passed, falls back
+        :type client: :class:`~google.cloud.storage.client.Client`
+        :param client: (Optional) The client to use.  If not passed, falls back
                        to the ``client`` stored on the blob's bucket.
         """
         content_type = self._get_content_type(content_type, filename=filename)
 
         with open(filename, 'rb') as file_obj:
+            total_bytes = os.fstat(file_obj.fileno()).st_size
```
jonparrott: I am not an expert, but is this portable? Would fstat fail on weird filesystems like App Engine?

jonparrott:
Overall this looks fine, just some small concerns.

> For now, the build is failing unit tests, because I haven't done those (what I'm sure will be very time-consuming) updates

Do you plan to address those in this PR?

> I haven't yet updated create_resumable_upload_session(), but I can (it should be fairly straightforward, but I'd rather do it in a separate PR)

Will you file a bug to track that, or do you have confidence you won't forget?

>  have ever-so-slightly changed the behavior of upload_from_file() when size is passed in. This is mostly because the existing implementation is confused about what the size / chunk_size combination should mean. The implementation now does the "sane" thing: use a resumable upload IF AND ONLY IF there is a chunk size specified on the blob

I'm okay with this.

> I would really like to deprecate the rewind keyword argument

Do it. File a bug if needed to track.

storage/google/cloud/storage/blob.py
dhermes: This is a private method and the only caller invokes this method from an `if self.chunk_size is None` block

storage/google/cloud/storage/blob.py
dhermes: I don't follow.

storage/google/cloud/storage/blob.py
dhermes: As of #3354, yes it is.

storage/google/cloud/storage/blob.py
dhermes: Not sure. I can factor this out as a `_get_size_from_fileno` helper and then just return `None` on `OSError` / `UnsupportedOperation`? (Seems YAGNI, but I would do it.)

storage/google/cloud/storage/blob.py
jonparrott: This seems to prepare data for the upload, but also prepares a transport for the upload. I'd prefer the upload functions themselves call `_make_transport`.

dhermes:
> Do you plan to address those (i.e. unit tests) in this PR?

Absolutely.

> Will you file a bug to track that (that `==` fixing the impl. of `create_resumable_upload_session`) , or do you have confidence you won't forget?

I realized that I would go below 100% line coverage in the `_create_upload` function if I left that alone. Since `_create_upload` needs to go anyways, I will just make the `create_resumable_upload_session` in this PR (rather than fighting coverage).

> Do it. File a bug if needed to track.

You mean like do it in this PR?

storage/google/cloud/storage/blob.py
jonparrott: Nah, that's fine.

jonparrott:
> You mean like do it in this PR?

Your call.

storage/google/cloud/storage/blob.py
dhermes: I **can** make that change, but keep in mind this is temporary until the `httplib2` swap is done (which I want to finish ASAP).

storage/google/cloud/storage/blob.py
```diff
@@ -840,52 +966,54 @@ def create_resumable_upload_session(
...
 
-        _, _, start_response = self._create_upload(
-            client,
-            size=size,
-            content_type=content_type,
-            strategy=RESUMABLE_UPLOAD,
-            extra_headers=extra_headers)
+        curr_chunk_size = self.chunk_size
+        try:
+            # Temporarily patch the chunk size. A user should still be able
```
jonparrott: No action needed, but this doesn't sit easy with me. It seems we're abusing the behavior of a non-pure function in order to get it to do something it's not intended to do. It's up to you if you want to do anything about it, I won't block.

dhermes:
Merging this now after discussions with @lukesneeringer and @jonparrott.

This needs a follow-up PR ASAP that supports `num_retries`, so I will be working on that.

Issue #3361, opened 2017-05-03
How put an array of objects at Google Cloud Datastore?


GustavoRPS:
**OS** : Ubuntu 14.04.5
**Python**: 3.4.3
**google-cloud-datastore**: 0.24.0

**Code example**:
```python
from google.cloud import datastore

class GoogleDatastorePipeline(object):
    
    def __init__(self, settings, stats):
        self.client = datastore.Client()
        self.batch = self.client.batch()

    def open(self):
            self.batch.begin()

    def process(self):
        key = self.client.key('Article')
        entity = datastore.Entity(key)
        entity['name'] = 'A post'
        entity['content'] = '<html></html>'
        # How I insert this as array of objects? Can be indexed, if no, no problem.
        entity['authors'] = [{
            'name': 'Author 1', 
            'type': 'person', 
        },{
            'name': 'Author 2', 
            'type': 'organization', 
        }]

        self.batch.put(entity)

    def close(self):
            self.batch.commit()
```
**Stacktrace**:
```python
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/ubuntu/workspace/ze/pipelines/__init__.py", line 215, in process_item
    self.batch.put(entity)
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/google/cloud/datastore/batch.py", line 199, in put
    _assign_entity_to_pb(entity_pb, entity)
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/google/cloud/datastore/batch.py", line 319, in _assign_entity_to_pb
    bare_entity_pb = helpers.entity_to_protobuf(entity)
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/google/cloud/datastore/helpers.py", line 219, in entity_to_protobuf
    _set_protobuf_value(value_pb, value)
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/google/cloud/datastore/helpers.py", line 408, in _set_protobuf_value
    attr, val = _pb_attr_value(val)
  File "/home/ubuntu/workspace/env/lib/python3.4/site-packages/google/cloud/datastore/helpers.py", line 325, in _pb_attr_value
    raise ValueError("Unknown protobuf attr type %s" % type(val))
ValueError: Unknown protobuf attr type <class 'dict'>
[...]

PR #3363, opened 2017-05-03
Add better documentation for support types in datastore Entity.

dhermes:
H/T to @GustavoRPS for bringing this up in #3361.

Also snuck in a change in `google.cloud.datastore.helpers` to use `six.binary_type` in place of `(str, bytes)`. (It wasn't a Py3 error before because that check came **after** a `six.text_type` check.)

### DIFF
--- a/datastore/google/cloud/datastore/entity.py
+++ b/datastore/google/cloud/datastore/entity.py
@@ -24,9 +24,9 @@ class Entity(dict):
     An entity storing the actual instance of data.
 
     Each entity is officially represented with a
-    :class:`google.cloud.datastore.key.Key` class, however it is possible that
-    you might create an Entity with only a partial Key (that is, a Key
-    with a Kind, and possibly a parent, but without an ID).  In such a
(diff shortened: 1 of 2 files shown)


datastore/google/cloud/datastore/entity.py
```diff
@@ -24,9 +24,9 @@ class Entity(dict):
     An entity storing the actual instance of data.
 
     Each entity is officially represented with a
-    :class:`google.cloud.datastore.key.Key` class, however it is possible that
-    you might create an Entity with only a partial Key (that is, a Key
-    with a Kind, and possibly a parent, but without an ID).  In such a
+    :class:`~google.cloud.datastore.key.Key`, however it is possible that
+    you might create an entity with only a partial key (that is, a key
+    with a kind, and possibly a parent, but without an ID).  In such a
```
dhermes: These changes were some "editorializing" by me, happy to remove, though I think they are good changes.

dhermes:
@tseaver @lukesneeringer Can I get a review here?

PR #3365, opened 2017-05-03
Dropping internal usage of rewind in Blob.upload_from_string().

dhermes:


dhermes:
I'm going to hold off on this merge since it will have a conflict with #3362.

PR #3375, opened 2017-05-05
Add back support for retries in storage uploads.

dhermes:
Also clean up a lot of the documentation for the storage blob module.

PR #3378, opened 2017-05-05
Add back support for retries in storage uploads.

dhermes:


dhermes:
Was #3375

dhermes:
@tseaver @lukesneeringer Can I get a review here?

storage/google/cloud/storage/blob.py
```diff
@@ -631,13 +637,19 @@ def _do_multipart_upload(self, client, stream, content_type, size):
         upload_url = _MULTIPART_URL_TEMPLATE.format(
             bucket_path=self.bucket.path)
         upload = MultipartUpload(upload_url, headers=headers)
+
+        if num_retries is not None:
```
tseaver: Should we emit a `DeprecationWarning` for explictly-passed `num_retries`?

storage/google/cloud/storage/blob.py
```diff
@@ -693,13 +709,19 @@ def _initiate_resumable_upload(self, client, stream, content_type,
         upload_url = _RESUMABLE_URL_TEMPLATE.format(
             bucket_path=self.bucket.path)
         upload = ResumableUpload(upload_url, chunk_size, headers=headers)
+
+        if num_retries is not None:
```
tseaver: Should we emit a `DeprecationWarning` for explictly-passed `num_retries`?

storage/tests/unit/test_blob.py
```diff
@@ -817,6 +818,11 @@ def test__do_multipart_upload_no_size(self, mock_get_boundary):
     def test__do_multipart_upload_with_size(self, mock_get_boundary):
         self._do_multipart_success(mock_get_boundary, size=10)
 
+    @mock.patch(u'google.resumable_media._upload.get_boundary',
+                return_value=b'==0==')
+    def test__do_multipart_upload_with_retry(self, mock_get_boundary):
```
tseaver: Would `...with_num_retries` be more descriptive?

storage/google/cloud/storage/blob.py
dhermes: Not in the private helper. The main entry point is `upload_from_file` and that does emit a warning

storage/google/cloud/storage/blob.py
dhermes: Ditto here. This is private and the public entry point will emit.

storage/tests/unit/test_blob.py
dhermes: I'm happy to change it if you like.

PR #3379, opened 2017-05-05
Correct speech Operation docstring

cfperez:
ValueError is raised when there are no results, not when there is more than one result, which is what the docstring says.


dhermes:
LGTM

dhermes:
Thanks @cfperez. No need to wait on CI since your change is docs only (and even more, these docs don't get built since a non-public method)

PR #3366, opened 2017-05-04
Incorrect error creating empty partitioned table

richkadel:
I moved the partitioning_type check into the if/elif block to avoid a ValueError that should not have been thrown.

See: https://cloud.google.com/bigquery/docs/creating-partitioned-tables#creating_a_partitioned_table and the "API" example.

Before my change, the following code failed with the ValueError thrown by this function, but with my change, I was able to successfully create the partitioned table:

    destination_table = dataset.table(table_shortname)
    destination_partition = dataset.table(table_shortname+'$'+partition_date)
    
    if not destination_table.exists():
      destination_table.partitioning_type = 'DAY'
      destination_table.create()

PR #3381, opened 2017-05-06
Fix new partitioned table

richkadel:
[This is a re-do of the pull request #3366 to fix the author email address to use my `google.com` email]

I moved the `partitioning_type` check into the `if` / `elif` block to avoid a `ValueError` that should not have been thrown.

See:
https://cloud.google.com/bigquery/docs/creating-partitioned-tables#creating_a_partitioned_table and the "API" example.

Before my change, the following code failed with the `ValueError` thrown by this function, but with my change, I was able to successfully create the partitioned table:

```python
destination_table = dataset.table(table_shortname)

if not destination_table.exists(): 
    destination_table.partitioning_type = 'DAY'
    destination_table.create()

# Use a table partition
destination_partition = dataset.table(table_shortname + '$' + partition_date)
```

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -473,8 +473,9 @@ class Table(object):
             resource['schema'] = {
                 'fields': _build_schema_resource(self._schema)
             }
-        else:
-            raise ValueError("Set either 'view_query' or 'schema'.")
+        elif self.partitioning_type is None:
+            raise ValueError(
+                "Set either 'view_query' or 'schema' or 'partitioning_type'.")
 
         return resource
 
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -469,13 +469,11 @@ class Table(object):
         if self.view_query is not None:
             view = resource['view'] = {}
             view['query'] = self.view_query
-        elif self._schema:
+
+        if self._schema:
             resource['schema'] = {
                 'fields': _build_schema_resource(self._schema)
             }
-        elif self.partitioning_type is None:
-            raise ValueError(
-                "Set either 'view_query' or 'schema' or 'partitioning_type'.")
 
         return resource
 
--- a/bigquery/tests/unit/test_table.py
+++ b/bigquery/tests/unit/test_table.py
@@ -395,7 +395,7 @@ class TestTable(unittest.TestCase, _SchemaBase):
         self.assertIs(table._dataset._client, client)
         self._verifyResourceProperties(table, RESOURCE)
 
-    def test_create_no_view_query_no_schema(self):
+    def test_create_no_view_query_no_schema_no_partitioning(self):
         conn = _Connection()
         client = _Client(project=self.PROJECT, connection=conn)
         dataset = _Dataset(client)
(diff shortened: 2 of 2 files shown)


richkadel:
@dhermes @jonparrott 
I resubmitted my changes in a new pull request to make sure it has my @google.com email, and withdrew pull request #3366

dhermes:
Thanks @richkadel. This LGTM.

@tswast Any issues?

dhermes:
@richkadel Your code snippet above never **uses** `destination_partition`. Is there a typo somewhere?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -473,8 +473,9 @@ def _build_resource(self):
             resource['schema'] = {
                 'fields': _build_schema_resource(self._schema)
             }
-        else:
-            raise ValueError("Set either 'view_query' or 'schema'.")
+        elif self.partitioning_type is None:
+            raise ValueError(
+                "Set either 'view_query' or 'schema' or 'partitioning_type'.")
```
tswast: This error message is getting a little confusing. Maybe the client shouldn't be doing this kind of validation?

I think schema isn't required if autodetect is set to true, for example.

dhermes: I would be OK with just removing the exception all together

richkadel:
@dhermes re: @richkadel Your code snippet above never uses destination_partition. Is there a typo somewhere?

No typo. The code that uses the partition itself is not shown. I left the line in to show that the table name used in the "create" call is the table name _without_ the partition suffix. Once the table is created then you can use destination_partition value as the "job_destination" setting to materialize data into just that partition.

dhermes:
@richkadel Gotcha, I just made an edit to make that clearer.

bigquery/google/cloud/bigquery/table.py
richkadel: The original logic was not mine, so I didn't want to presume to change how it worked or what it checked (other than what was broken here that prevented create a partitioned table). Do you want me to remove these checks, and the tests that depend on them? (In my opinion, the ValueError does ensure the expected values are testable, so I'd argue against removing them completely.)

Re: "autodetect", I'm not sure that's true (though I'm not an expert here). This method creates an empty table (or view), so since there is no data from which to detect a schema, the schema is required. I believe (correct me if I'm wrong) that autodetect is used with a query that materializes the result to a job_destination, which is a different method.

richkadel: Please advise... Thanks!

dhermes: @lukesneeringer @tseaver @jonparrott WDYT here?

tswast: The autodetect parameter in table insert is useful for external tables (Google Drive, Sheets, Bigtable) where the data is stored outside of BigQuery.

I don't actually know what will happen when we try to create an autodetect external table with this client library. Looks to me like this exception will get thrown.

I understand the desire for easy unit testing, but this client library is not changing in exact sync with the backend API, so I'd prefer if this validation was done server-side.

richkadel: That makes sense. So if I remove the checks and their tests, is this all you need to approve the change? 

These three lines would be removed from table.py:

```python
        elif self.partitioning_type is None:
            raise ValueError(
                "Set either 'view_query' or 'schema' or 'partitioning_type'.")

```

tswast: Sounds good to me.

richkadel: Oh, and I will change `elif self._schema:` to `if self._schema:` for the same reason: Not making assumptions the server side will never allow both a view query and a schema; leave the validation to the server side. OK?

tswast: :+1:

richkadel: @tswast During local nox test, I found that `test_create_w_alternate_client` is creating a create table request with both a schema and a view query. This used to succeed because the `elif` was there and the schema is not copied to into the request. I have a few options: 

1. Add the schema to the expected value in the request (because the request will now have both the view query and schema)
1. "fix" the test so it doesn't create the create table request with a schema. 
1. Change the test to remove the view query and add the schema to the expected value instead

Since the point of this test seems to be independent of the content of the request, the simplest solution is to remove the schema from the input (option 2).

richkadel: Updates have been pushed and tests succeed.

richkadel:
Updates have been pushed and all tests pass.

PR #3384, opened 2017-05-08
    Adding system test to exercise all possible Spanner types.

### LINKED ISSUE #3016: Add system tests for spanner datatypes
Show inserting / reading all the datatypes supported by spanner, particularly:

- Special-case floats (`nan`, `+inf`, `-inf`).
- Arrays of all scalar types.

dhermes:
In the process, changing how +/- Infinity are sent to the backend (this was a bug).

Fixes #3016.

/cc @vkedia It's worth noting that `Value(string_value='inf')` / `Value(string_value='+inf')` / `Value(string_value='-inf')` are being rejected, but should probably be accepted when `Type.code == FLOAT64`.


dhermes:
Just created this branch on the GCP remote so that we'd get a [build][1] with system tests

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1771

vkedia:
@dhermes It has to be `"Infinity"` and `"-Infinity"`

spanner/tests/system.py
```diff
@@ -75,9 +85,6 @@ def setUpModule():
     Config.INSTANCE_CONFIG = configs[0]
     config_name = configs[0].name
 
-    def _list_instances():
-        return list(Config.CLIENT.list_instances())
-
     instances = retry(_list_instances)()
```
tseaver: -0.  I don't really see the benefit in moving this function out to module scope.

spanner/tests/system.py
```diff
@@ -258,12 +266,37 @@ def tearDown(self):
         for doomed in self.to_delete:
             doomed.delete()
 
-    def _check_row_data(self, row_data):
-        self.assertEqual(len(row_data), len(self.ROW_DATA))
-        for found, expected in zip(row_data, self.ROW_DATA):
+    def _assert_timestamp(self, value, nano_value):
+        self.assertIsInstance(value, datetime.datetime)
+        self.assertIn(value.tzinfo, (None, UTC))
```
tseaver: Why the ambiguity?  It seems like the caller should specify whether naive or UTC.

spanner/tests/system.py
```diff
@@ -258,12 +266,37 @@ def tearDown(self):
...
+        self.assertIs(nano_value.tzinfo, UTC)
+
+        self.assertEqual(value.year, nano_value.year)
+        self.assertEqual(value.month, nano_value.month)
+        self.assertEqual(value.day, nano_value.day)
+        self.assertEqual(value.hour, nano_value.hour)
+        self.assertEqual(value.minute, nano_value.minute)
+        self.assertEqual(value.second, nano_value.second)
+        self.assertEqual(value.microsecond, nano_value.microsecond)
+        if isinstance(value, TimestampWithNanoseconds):
```
tseaver: Shouldn't we know which type value should be?

spanner/tests/system.py
```diff
@@ -371,14 +445,19 @@ def test_transaction_read_and_insert_or_update_then_commit(self):
...
+
     def _set_up_table(self, row_count):
         from google.cloud.spanner import KeySet
 
-        def _row_data(max_index):
-            for index in range(max_index):
-                yield [index, 'First%09d' % (index,), 'Last09%d' % (index),
-                       'test-%09d@example.com' % (index,)]
-
         keyset = KeySet(all_=True)
```
tseaver: It isn't called elsewhere:  why move it out to class scope?

dhermes:
> It has to be `"Infinity"` and `"-Infinity"`

@vkedia I realize that now (made the necessary change in this PR). I'm suggesting that you **also** accept `inf`/ `+inf` / `-inf`.

tseaver:
@dhermes The principle you were trying to name in today's call is [Postel's Law](https://en.wikipedia.org/wiki/Jon_Postel#Postel.27s_law):

> An implementation should be conservative in its sending behavior, and liberal in its receiving behavior

spanner/tests/system.py
dhermes: That's fine with me. I'm pretty sure all of the calls to this will be naive timestamps so I'll change it to `assertIsNone`

spanner/tests/system.py
dhermes: I have an allergy to dynamically defined functions and classes. If nothing else, they make the flow of code harder to reason about (for both scope and indentation reasons).

spanner/tests/system.py
dhermes: I want it to allow both `datetime.datetime` and `TimestampWithNanoseconds`. My caller will use both types to exercise the branches [in `_make_value_pb`]1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9870057d5e6d8d58b0ab0e854b03fe0d15543877/spanner/google/cloud/spanner/_helpers.py#L138-L141

spanner/tests/system.py
dhermes: See my allergy to dynamically defined functions [above][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3384#discussion_r115375627

spanner/tests/system.py
tseaver: I don't see how a one-liner defined immediately above its only point-of-use is harder to reason about than the same function moved further away, and made (pointlessly, otherwise) a method.

Go ahead and leave it as is.  However, if we are adopting that as a standard (e.g., you ask for me to change it on my PRs every time you see it), then we should be linting for it:  otherwise it is a pure *de gustibus* quibble.

dhermes:
@tseaver We good to merge this?

tseaver:
Ugh, I screwed up merging mine first.

dhermes:
I'll fix it.

tseaver:
I just tweaked it here.  Feel free to back out my change.

tseaver:
And merge when green.

dhermes:
Ah sorry, just did a force push before seeing your comment.

tseaver:
No worries.

dhermes:
Will wait on build on the GCP remote: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1785

tseaver:
Green on CircleCI.

PR #3385, opened 2017-05-08
Add system test for Database convenience methods:

### LINKED ISSUE #3015: Add system tests for database convenience wrappers.
Add tests which exercise the higher-level `Database` convenience APIs (`read`, `execute_sql`, `run_in_transaction`, `batch`, and `snapshot`).

tseaver:
- `batch`
- `snapshot`
- `read`
- `run_with_transaction`
- `execute_sql`

Closes #3015.

### DIFF
--- a/spanner/tests/system.py
+++ b/spanner/tests/system.py
@@ -19,6 +19,7 @@ import unittest
 from google.cloud.proto.spanner.v1.type_pb2 import STRING
 from google.cloud.proto.spanner.v1.type_pb2 import Type
 from google.cloud.spanner.client import Client
+from google.cloud.spanner.keyset import KeySet
 from google.cloud.spanner.pool import BurstyPool
 from google.cloud.spanner._fixtures import DDL_STATEMENTS
 
@@ -167,12 +168,32 @@ class TestInstanceAdminAPI(unittest.TestCase):
         Config.INSTANCE.update()
 
 
-class TestDatabaseAdminAPI(unittest.TestCase):
+class _TestData(object):
+    TABLE = 'contacts'
+    COLUMNS = ('contact_id', 'first_name', 'last_name', 'email')
+    ROW_DATA = (
+        (1, u'Phred', u'Phlyntstone', u'phred@example.com'),
+        (2, u'Bharney', u'Rhubble', u'bharney@example.com'),
+        (3, u'Wylma', u'Phlyntstone', u'wylma@example.com'),
+    )
+    ALL = KeySet(all_=True)
+    SQL = 'SELECT * FROM contacts ORDER BY contact_id'
+
+    def _check_row_data(self, row_data):
+        self.assertEqual(len(row_data), len(self.ROW_DATA))
+        for found, expected in zip(row_data, self.ROW_DATA):
+            self.assertEqual(len(found), len(expected))
+            for f_cell, e_cell in zip(found, expected):
+                self.assertEqual(f_cell, e_cell)
+
+
+class TestDatabaseAPI(unittest.TestCase, _TestData):
 
     @classmethod
     def setUpClass(cls):
         pool = BurstyPool()
-        cls._db = Config.INSTANCE.database(DATABASE_ID, pool=pool)
+        cls._db = Config.INSTANCE.database(
+            DATABASE_ID, ddl_statements=DDL_STATEMENTS, pool=pool)
         cls._db.create()
 
     @classmethod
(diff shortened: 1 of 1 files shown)


spanner/tests/system.py
```diff
@@ -228,16 +247,49 @@ def test_update_database_ddl(self):
 
         self.assertEqual(len(temp_db.ddl_statements), len(DDL_STATEMENTS))
 
+    def test_db_batch_insert_then_db_snapshot_read_and_db_read(self):
+        from google.cloud.spanner import KeySet
```
dhermes: Just import globally?

spanner/tests/system.py
```diff
@@ -228,16 +247,49 @@ def test_update_database_ddl(self):
...
+        keyset = KeySet(all_=True)
```
dhermes: I noticed we make a lot of these in the system tests, could `ALL_KEYS = KeySet(all_=True)` be a global or does it require some context?

spanner/tests/system.py
```diff
@@ -228,16 +247,49 @@ def test_update_database_ddl(self):
...
-        (1, u'Phred', u'Phlyntstone', u'phred@example.com'),
-        (2, u'Bharney', u'Rhubble', u'bharney@example.com'),
-        (3, u'Wylma', u'Phlyntstone', u'wylma@example.com'),
-    )
-    SQL = 'SELECT * FROM contacts ORDER BY contact_id'
+        retry = RetryInstanceState(_has_all_ddl)
+        retry(self._db.reload)()
+
+        with self._db.batch() as batch:
+            batch.delete(self.TABLE, keyset)
```
dhermes: Why is this necessary? (i.e. is there a way to start with an empty table)

spanner/tests/system.py
```diff
@@ -228,16 +247,49 @@ def test_update_database_ddl(self):
...
+            batch.insert(self.TABLE, self.COLUMNS, self.ROW_DATA)
+
+        with self._db.snapshot(read_timestamp=batch.committed) as snapshot:
+            from_snap = list(snapshot.read( self.TABLE, self.COLUMNS, keyset))
```
dhermes: `snapshot.read( self.TABLE, ...` has an extra space after the parenthesis

spanner/tests/system.py
```diff
@@ -228,16 +247,49 @@ def test_update_database_ddl(self):
...
+        from google.cloud.spanner import KeySet
+        keyset = KeySet(all_=True)
+
+        retry = RetryInstanceState(_has_all_ddl)
+        retry(self._db.reload)()
+
+        with self._db.batch() as batch:
+            batch.delete(self.TABLE, keyset)
+
+        def _unit_of_work(transaction, test):
```
dhermes: Can you make this a `@staticmethod` on the `TestCase`?

spanner/tests/system.py
tseaver: OK,

spanner/tests/system.py
tseaver: I'll make it a class attribute of the base class which has the other constants used to set up the tables.

spanner/tests/system.py
tseaver: Races about previous uses of the same table (we can't create new tables cheaply on the fly).

spanner/tests/system.py
tseaver: Why?  The "normal" use of  `run_in_transaction` is to pass in a function with additional varargs / kwargs as needed, which is exactly what we are doing here.

tseaver:
Should I [follow suit](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3384#issuecomment-300004596) and push the updated branch to the origin to get a build with system tests?  We could close this PR and re-open there.

dhermes:
@tseaver Yes I think you should push this branch to the GCP origin to get system tests running. You don't need to close this PR though, the tests will "just work" (i.e. be associated here).

spanner/tests/system.py
dhermes: More [dynamically created function allergy][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3384#discussion_r115375627

tseaver:
Pushed to `upstream`.

spanner/tests/system.py
tseaver: *De gustibus*

tseaver:
@dhermes New system tests pass on CI.  Any further issues?

dhermes:
LGTM

PR #3386, opened 2017-05-08
Add MonitoredResource to interface

### LINKED ISSUE #3377: Implement monitored resource for logging client
Log entries can have a monitored resource associated with them as of v2 of the logging API.

* entries.write documentation: https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write
* MonitoredResource reference: https://cloud.google.com/logging/docs/reference/v2/rest/v2/MonitoredResource
* Existing tiny resource class in monitoring: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/monitoring/google/cloud/monitoring/resource.py#L159

Essentially, the `_BaseEntry` class should be extend to accept a `resource`. Helper methods such as `log_text` and `log_struct` should likely be extended to accept the resource as well.

This'll need to work with http and gax.

waprin:
This adds MonitoredResource as an option to higher level interface.

Fixes #3377 and #2673.

I also changed the variable name from `resource` to `entry` in `make_entry_resource` because its' confusing to call it a resource when resource is overloaded.

Tests/docs are TBD, wanted +1 on overall direction then I'll write them.

Some notes:

1) This is intended to be a non-breaking change to existing users. If you use `log_text`, `log_struct` etc the default is still the global resource. However, within a `Batch` you could also explicitly set it to None and wait to set it until the final `commit` call. 

2) Copied the Resource class from Monitoring. They are the same proto so we could move it to a common package . We could also use the proto but unclear if that will work with the _gax. This is what @jonparrott suggested and I think is pretty.

cc @liyanhui1228 @duggelz


logging/google/cloud/logging/logger.py
```diff
@@ -123,19 +128,24 @@ def _make_entry_resource(self, text=None, info=None, message=None,
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (Optional) timestamp of event being logged.
 
+        :type resource :class:``google.cloud.logging.resource`
```
jonparrott: This reference should be ":class:\`google.cloud.logging.resource.Resource\`". (seems like this needs to be fixed in a few places)

logging/google/cloud/logging/logger.py
```diff
@@ -436,29 +462,34 @@ def log_proto(self, message, labels=None, insert_id=None, severity=None,
         """
         self.entries.append(
             ('proto', message, labels, insert_id, severity, http_request,
-             timestamp))
+             timestamp, resource))
 
-    def commit(self, client=None):
+    def commit(self, client=None, resource=None):
```
jonparrott: Does a resource param make sense on this method? Wouldn't the resource be populated in all the entries we're committing?

logging/google/cloud/logging/resource.py
```diff
@@ -0,0 +1,275 @@
...
+            '<ResourceDescriptor:\n'
+            ' name={name!r},\n'
+            ' type={type!r},\n'
+            ' labels={labels!r},\n'
+            ' display_name={display_name!r},\n'
+            ' description={description!r}>'
+        ).format(**self.__dict__)
+
+
+class Resource(collections.namedtuple('Resource', 'type labels')):
```
jonparrott: Isn't this the only class used in Logging? Can we drop everything above this?

waprin: Ya good point

logging/google/cloud/logging/logger.py
waprin: So the reason to do it this way is because when you write entries you can leave some entries without resources and override them at the top level. I just wanted to expose that level of the API, maybe not strictly necessary.

waprin:
Ya like I said just wanted +1 on overall direction before clean up tests docs since I hate writing all the tests/docs _then_ being asked to totally change everything.

logging/google/cloud/logging/logger.py
jonparrott: I'd say let's leave it off for now and just have users specify it on the entry level. If there were any other properties that we did this way, I'd be for it, but I'd say let's leave it out until someone asks for it. :)

jonparrott:
@waprin direction looks good. Let me know when you're ready for another review.

logging/google/cloud/logging/logger.py
waprin: I moved it to the constructor which I think makes more sense for a ContextManager.

logging/google/cloud/logging/logger.py
waprin: done

waprin:
@jonparrott ready for review, looks like we have an unrelated BigQuery system test failure.

logging/nox.py
```diff
@@ -73,7 +73,7 @@ def lint(session):
     Returns a failure if flake8 finds linting errors or sufficiently
     serious code quality issues.
     """
-    session.interpreter = 'python3.6'
+    session.interpreter = 'python3.5'
```
jonparrott: why'd we drop this to 3.5?

logging/tests/system.py
```diff
@@ -171,6 +171,24 @@ def test_log_text_with_timestamp(self):
         self.assertEqual(entries[0].payload, text_payload)
         self.assertEqual(entries[0].timestamp, now.replace(tzinfo=UTC))
 
+    def test_log_text_with_resource(self):
```
jonparrott: Nice! Is this passing? With both http and gax?

logging/nox.py
dhermes: Yes. Please keep it at 3.6 (it concerns me if there is a failure in 3.6 that passes in 3.5. If the answer is because you don't have 3.6 installed locally ping me and I can walk you through some "recommended" ways to install)

logging/google/cloud/logging/logger.py
```diff
@@ -16,6 +16,10 @@
 
 from google.protobuf.json_format import MessageToDict
 from google.cloud._helpers import _datetime_to_rfc3339
+from google.cloud.logging.resource import Resource
+
+
+_GLOBAL_RESOURCE = Resource(type='global', labels={})
```
dhermes: Does `labels` not have a sensible default?

logging/google/cloud/logging/logger.py
```diff
@@ -123,19 +128,24 @@ def _make_entry_resource(self, text=None, info=None, message=None,
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (Optional) timestamp of event being logged.
 
+        :type resource :class:`google.cloud.logging.resource.Resource`
```
dhermes: Can you use a tilde before the `google`:

```
:class:`~google.cloud.logging.resource.Resource`
```

This makes the links look much better (would be `Resource` instead of the full path)

logging/google/cloud/logging/logger.py
```diff
@@ -194,17 +205,23 @@ def log_text(self, text, client=None, labels=None, insert_id=None,
         :param http_request: (optional) info about HTTP request associated with
                              the entry
 
+        :type resource :class:`google.cloud.logging.resource.Resource`
+        :param resource: (Optional) Monitored resource of the entry, defaults
+                         to the global resource type.
+
+
```
dhermes: Too much whitespace. Just one empty line between.

logging/google/cloud/logging/logger.py
```diff
@@ -231,17 +248,23 @@ def log_struct(self, info, client=None, labels=None, insert_id=None,
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
+        :type resource :class:``google.cloud.logging.resource.Resource`
+        :param resource: (Optional) Monitored resource of the entry, defaults
+                         to the global resource type.
+
+
```
dhermes: Ditto on whitespace

logging/google/cloud/logging/logger.py
```diff
@@ -268,13 +291,17 @@ def log_proto(self, message, client=None, labels=None, insert_id=None,
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
+        :type resource :class:``google.cloud.logging.resource.Resource`
```
dhermes: You have an extra backtick here (two backticks before `google`)

logging/google/cloud/logging/logger.py
```diff
@@ -231,17 +248,23 @@ def log_struct(self, info, client=None, labels=None, insert_id=None,
...
```
dhermes: You have two backticks before `google` (also drop in a tilde `~` before `google`)

logging/tests/system.py
waprin: It works for me on both http and gax but I don't believe system tests run on both. Seems parametizing system tests for both is higher level issue?

logging/nox.py
waprin: Yes sorry didn't mean to include this. Yeah was just being lazy cause don't have 3.6 installed.

logging/google/cloud/logging/logger.py
```diff
@@ -194,17 +205,23 @@ def log_text(self, text, client=None, labels=None, insert_id=None,
         :param http_request: (optional) info about HTTP request associated with
                              the entry
 
+        :type resource :class:`google.cloud.logging.resource.Resource`
```
dhermes: You are missing a colon after `resource` (here and most of the other places you've added a parameter). Should be:

```
        :type resource: :class:`~google.cloud.logging.resource.Resource`
        :param resource: (Optional) Monitored resource of the entry, defaults
                         to the global resource type.
```

I'm fairly certain this will break the docs build?

logging/google/cloud/logging/logger.py
```diff
@@ -379,13 +410,17 @@ def log_text(self, text, labels=None, insert_id=None, severity=None,
 
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
+
+        :type resource :class:``google.cloud.logging.resource`
```
dhermes: This

- is missing a colon after `resource`
- has an extra backtick before `google` (two instead of one)
- refers to a module not a class
- could benefit from a `~` before `google`

It seems you've got a few other references to `google.cloud.logging.resource` instead of `google.cloud.logging.resource.Resource`. Can you do another pass and find-replace them?

logging/google/cloud/logging/resource.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2016 Google Inc.
```
dhermes: 2017

logging/google/cloud/logging/resource.py
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Monitored Resource for the Google Logging API V2."""
+
+import collections
+
+
+class Resource(collections.namedtuple('Resource', 'type labels')):
```
dhermes: I don't mind but I think someone (maybe @lukesneeringer or @jonparrott) prefers

```
['type', 'labels']
```

Maybe not?

logging/tests/system.py
```diff
@@ -171,6 +171,24 @@ def test_log_text_with_timestamp(self):
...
+        from google.cloud.logging.resource import Resource
```
dhermes: Can you just import this globally (system test style slightly different than unit test style, thanks for trying to match though)

waprin:
System test snafu, forgot they say success without running without service account locally, fixing now.

logging/tests/system.py
```diff
@@ -171,6 +171,24 @@ def test_log_text_with_timestamp(self):
...
+        text_payload = 'System test: test_log_text_with_timestamp'
+
+        logger = Config.CLIENT.logger(self._logger_name())
+        now = datetime.datetime.utcnow()
+        RESOURCE = Resource(
+            type='gae_app', labels={'module_id': 'default',
+                                    'version_id': 'test'})
```
dhermes: This is a style nit, so feel free to ignore, but I would prefer:

```python
resource = Resource(
    type='gae_app',
    labels={
        'module_id': 'default',
        'version_id': 'test',
    },
)
```

(Also you don't need to / shouldn't treat `RESOURCE` like a constant. I realize our unit tests do that, but it's "not a thing".)

logging/tests/unit/test_logger.py
```diff
@@ -89,13 +89,15 @@ def test_batch_w_alternate_client(self):
         self.assertIs(batch.client, client2)
 
     def test_log_text_w_str_implicit_client(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -89,13 +89,15 @@ def test_batch_w_alternate_client(self):
...
         TEXT = 'TEXT'
         ENTRIES = [{
             'logName': 'projects/%s/logs/%s' % (
                 self.PROJECT, self.LOGGER_NAME),
             'textPayload': TEXT,
             'resource': {
                 'type': 'global',
+                'labels': {}
```
dhermes: Trailing comma just for fun?

logging/tests/unit/test_logger.py
```diff
@@ -108,6 +110,7 @@ def test_log_text_w_str_implicit_client(self):
                          (ENTRIES, None, None, None))
 
     def test_log_text_w_default_labels(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -131,6 +135,7 @@ def test_log_text_w_default_labels(self):
 
     def test_log_text_w_timestamp(self):
         import datetime
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -141,6 +146,7 @@ def test_log_text_w_timestamp(self):
             'timestamp': '2016-12-31T00:01:02.999999Z',
             'resource': {
                 'type': 'global',
+                'labels': {}
```
dhermes: trailing comma?

logging/tests/unit/test_logger.py
```diff
@@ -153,6 +159,7 @@ def test_log_text_w_timestamp(self):
                          (ENTRIES, None, None, None))
 
     def test_log_text_w_unicode_explicit_client_labels_severity_httpreq(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -172,6 +179,7 @@ def test_log_text_w_unicode_explicit_client_labels_severity_httpreq(self):
             'textPayload': TEXT,
             'resource': {
                 'type': 'global',
+                'labels': {}
```
dhermes: Trailing comma?

logging/tests/unit/test_logger.py
```diff
@@ -191,13 +199,15 @@ def test_log_text_w_unicode_explicit_client_labels_severity_httpreq(self):
                          (ENTRIES, None, None, None))
 
     def test_log_struct_w_implicit_client(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -191,13 +199,15 @@ def test_log_text_w_unicode_explicit_client_labels_severity_httpreq(self):
...
         STRUCT = {'message': 'MESSAGE', 'weather': 'cloudy'}
         ENTRIES = [{
             'logName': 'projects/%s/logs/%s' % (
                 self.PROJECT, self.LOGGER_NAME),
             'jsonPayload': STRUCT,
             'resource': {
                 'type': 'global',
+                'labels': {}
```
dhermes: trailing comma?

logging/tests/unit/test_logger.py
```diff
@@ -210,6 +220,7 @@ def test_log_struct_w_implicit_client(self):
                          (ENTRIES, None, None, None))
 
     def test_log_struct_w_default_labels(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -218,6 +229,7 @@ def test_log_struct_w_default_labels(self):
             'jsonPayload': STRUCT,
             'resource': {
                 'type': 'global',
+                'labels': {}
```
dhermes: trailing comma?

logging/tests/unit/test_logger.py
```diff
@@ -232,6 +244,7 @@ def test_log_struct_w_default_labels(self):
                          (ENTRIES, None, None, None))
 
     def test_log_struct_w_explicit_client_labels_severity_httpreq(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -272,6 +286,7 @@ def test_log_struct_w_explicit_client_labels_severity_httpreq(self):
 
     def test_log_struct_w_timestamp(self):
         import datetime
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this but never use it

logging/tests/unit/test_logger.py
```diff
@@ -530,16 +551,19 @@ def test_ctor_defaults(self):
         self.assertEqual(len(batch.entries), 0)
 
     def test_log_text_defaults(self):
+        from google.cloud.logging.logger import _GLOBAL_RESOURCE
```
dhermes: You import this and actually use it here!

logging/google/cloud/logging/logger.py
```diff
@@ -379,13 +410,17 @@ def log_text(self, text, labels=None, insert_id=None, severity=None,
...
+        :param resource: (Optional) Monitored resource of the entry
         """
         self.entries.append(
             ('text', text, labels, insert_id, severity, http_request,
-             timestamp))
+             timestamp, resource))
```
dhermes: ISTM that this is incorrect.

You should be allowing `self.resource` to matter if the caller never sends a `resource`

logging/google/cloud/logging/logger.py
```diff
@@ -406,13 +441,17 @@ def log_struct(self, info, labels=None, insert_id=None, severity=None,
...
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
+
+        :type resource :class:``google.cloud.logging.resource`
+        :param resource: (Optional) Monitored resource of the entry
         """
         self.entries.append(
             ('struct', info, labels, insert_id, severity, http_request,
-             timestamp))
+             timestamp, resource))
```
dhermes: Ditto here: should allow `self.resource` as fallback

logging/google/cloud/logging/logger.py
```diff
@@ -433,10 +472,13 @@ def log_proto(self, message, labels=None, insert_id=None, severity=None,
...
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
+
+        :type resource :class:``google.cloud.logging.resource.Resource`
+        :param resource: (Optional) Monitored resource of the entry
         """
         self.entries.append(
             ('proto', message, labels, insert_id, severity, http_request,
-             timestamp))
+             timestamp, resource))
```
dhermes: Ditto here: should allow `self.resource` as fallback

logging/google/cloud/logging/logger.py
```diff
@@ -473,6 +517,8 @@ def commit(self, client=None):
                 info = {'protoPayload': MessageToDict(entry)}
             else:
                 raise ValueError('Unknown entry type: %s' % (entry_type,))
+            if resource is not None:
+                info['resource'] = resource._to_dict()
```
dhermes: OK maybe I don't understand?

logging/google/cloud/logging/logger.py
waprin: I copied it from Monitoring and it's a namedtuple. I could make it not a named tuple or do Resource._from_dict({'type': 'global})

logging/google/cloud/logging/logger.py
waprin: Will leave another comment, but basically each entry has a resource and a batch a resource that is used for any entries that don't have a resource. self.resource is batch resource, here resource is the entry resource.

logging/google/cloud/logging/resource.py
waprin: Copied from Monitoring, I don't care so whatever you want.

logging/tests/unit/test_logger.py
waprin: :-P

waprin:
@dhermes as far as Batch resource goes...

If you look at the API, when you write multiple entries, you can give each individual entry a resource but also the higher level write call (which we encapsulate as a Batch) a resource. The Batch resource (`self.resource`) is used wherever entries are missing one. So a few places you suggested we should fall back to self.resource, but really what we want there is just for the entry resource to be missing so that Batch resource overrides it on server side.

Also since all the `log_*` calls default to the global resource, callers would have to explicitly specify None for the individual resources then set the Batch resource. Debatable whether this is a use case people actually want, @jonparrott seemed skeptical of whether it should go in there but I wanted to keep the client library somewhat aligned with the API. I suppose I could rip it out and just ignore the whole Batch resource aspect altogether and make Resource required on all the entry calls.

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/system.py
waprin: agreed, done

logging/tests/system.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/tests/unit/test_logger.py
waprin: done

logging/google/cloud/logging/logger.py
waprin: done

logging/google/cloud/logging/logger.py
waprin: done

logging/google/cloud/logging/logger.py
waprin: done

logging/google/cloud/logging/logger.py
waprin: done

logging/google/cloud/logging/logger.py
jonparrott: Oh I see https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write there's batch-level arguments for resource, log name, and labels. We should make sure this reflects that.

logging/google/cloud/logging/resource.py
jonparrott: Let's keep it consistent with monitoring.

jonparrott:
@waprin I was skeptical because I didn't see that [the api](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write) has that functionality. We should match the API. We shouldn't go through and fill in the resource for each entry on our end, we should just submit it as part of the request to the server (in both gax and http).

waprin:
Right, I believe the way I did it best matches the API semantics. Only setting it at the Batch level is also not right because your Batch might contain entries of different resource type. 

Fixing coverage now...

logging/google/cloud/logging/logger.py
dhermes: No this is fine

logging/google/cloud/logging/entries.py
```diff
@@ -170,14 +181,18 @@ class ProtobufEntry(_BaseEntry):
     :type http_request: dict
     :param http_request: (optional) info about HTTP request associated with
                          the entry
+
+    :type resource :class:`google.cloud.logging.resource.Resource`
```
dhermes: 1. Please add a `~` before `google`
1. You are missing a colon after `resource`

logging/google/cloud/logging/logger.py
```diff
@@ -344,11 +367,15 @@ class Batch(object):
 
     :type client: :class:`google.cloud.logging.client.Client`
     :param client: The client to use.
+
+    :type resource: :class:`google.cloud.logging.resource.Resource`
```
dhermes: Please add a `~` before `google`

logging/google/cloud/logging/logger.py
```diff
@@ -123,19 +128,22 @@ def _make_entry_resource(self, text=None, info=None, message=None,
...
+        :param resource: (Optional) Monitored resource of the entry
+
         :rtype: dict
         :returns: The JSON resource created.
         """
-        resource = {
+        entry = {
             'logName': self.full_name,
-            'resource': {'type': 'global'},
+            'resource': resource._to_dict()
```
dhermes: You lost your trailing comma :cry:

logging/google/cloud/logging/logger.py
dhermes: But if this is the case, should `None` be the default rather than `_GLOBAL_RESOURCE`?

logging/google/cloud/logging/logger.py
dhermes: ISTM that unless a user **explicitly** sends `None`, this branch will always execute (since the default is `_GLOBAL_RESOURCE`)

logging/nox.py
```diff
@@ -95,7 +95,7 @@ def cover(session):
     This outputs the coverage report aggregating coverage from the unit
     test runs (not system test runs), and then erases coverage data.
     """
-    session.interpreter = 'python3.6'
+    session.interpreter = 'python3.5'
```
dhermes: Please put back to 3.6

logging/tests/system.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2017 Google Inc.
```
dhermes: This should not be edited

logging/tests/system.py
```diff
@@ -171,6 +172,28 @@ def test_log_text_with_timestamp(self):
...
+        text_payload = 'System test: test_log_text_with_timestamp'
+
+        logger = Config.CLIENT.logger(self._logger_name())
+        now = datetime.datetime.utcnow()
+        resource = Resource(
+            type='gae_app',
+            labels={
+                'module_id': 'default',
+                'version_id': 'test'
+        })
```
dhermes: Prefer the `}` dedent to align with it's start:

```python
        resource = Resource(
            type='gae_app',
            labels={
                'module_id': 'default',
                'version_id': 'test',
            },
        )
```

logging/tests/unit/test_entries.py
```diff
@@ -136,6 +143,14 @@ def test_from_api_repr_w_loggers_no_logger_match(self):
...
         METHOD = 'POST'
         URI = 'https://api.example.com/endpoint'
+        RESOURCE = Resource(
+            type='gae_app',
+            labels={
+                'type': 'gae_app',
+                'labels': {
+                    'module_id':'default',
+                    'version': 'test'
+        }})
```
dhermes: Ditto here on the dedent:

```python
        RESOURCE = Resource(
            type='gae_app',
            labels={
                'type': 'gae_app',
                'labels': {
                    'module_id':'default',
                    'version': 'test',
                },
            },
        )
```

logging/tests/unit/test_logger.py
```diff
@@ -329,6 +338,8 @@ def test_log_proto_w_default_labels(self):
             'protoPayload': json.loads(MessageToJson(message)),
             'resource': {
                 'type': 'global',
+                'labels': {},
+
```
dhermes: Nuke this empty line

logging/google/cloud/logging/logger.py
waprin: It could be either...few reasons I think `_GLOBAL_RESOURCE` is best. The best reason to keep it `_GLOBAL_RESOURCE` is because it's backwards compatible, previously every entry was set to global. Setting to `_GLOBAL_RESOURCE` or `None` seems reasonable, `_GLOBAL_RESOURCE` seems a bit easier since you can't accidentally leave off the resource altogehter (by forgetting in in Batch). My thinking looking at the API was that both Entry and Batch have a Resource and Batch overrides Resource is a bit weird (as evidenced by all discussion here), I wanted to expose ability to set it in either places, but just setting all the Entry resources to global seems like a pretty reasonable default (as evidenced by it having been the default until now).

logging/google/cloud/logging/logger.py
waprin: Yeah that's the idea, explained my thought process above. GLOBAL_RESOURCE is a reasonable default (and default until now). Explicitly setting None is basically saying 'let this entry use the resource of the batch as a whole'. Pretty much what it means in the API.

logging/google/cloud/logging/logger.py
waprin: done

logging/google/cloud/logging/entries.py
waprin: done

waprin: running docs now..

waprin:
@jonparrott @dhermes docs built, just looked carefully at all the styles/docs issue and all lgtm so  this is ready for another pass

logging/google/cloud/logging/entries.py
```diff
@@ -72,16 +73,19 @@ class _BaseEntry(object):
     :type http_request: dict
     :param http_request: (optional) info about HTTP request associated with
                          the entry
+    :type resource: :class:`~google.cloud.logging.resource.Resource`
```
dhermes: See my [previous comment][1]

> Please add an emtpy line in between parameters

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3386#discussion_r115809247

logging/google/cloud/logging/logger.py
```diff
@@ -268,13 +287,17 @@ def log_proto(self, message, client=None, labels=None, insert_id=None,
         :param http_request: (optional) info about HTTP request associated with
                              the entry.
 
+        :type resource: :class:`~google.cloud.logging.resource.Resource`
+        :param resource: Monitored resource of the entry
```
dhermes: Throw in your "defaults to the..." text used for the other `log_FOO()` methods

logging/google/cloud/logging/logger.py
```diff
@@ -379,13 +406,17 @@ def log_text(self, text, labels=None, insert_id=None, severity=None,
 
         :type timestamp: :class:`datetime.datetime`
         :param timestamp: (optional) timestamp of event being logged.
+
+        :type resource: :class:`~google.cloud.logging.resource.Resource`
+        :param resource: (Optional) Monitored resource of the entry
```
dhermes: Ditto here on the "defaults to the..." language. It's **extra** important since we want users to be clear on the fact that the batch's resource will not be used by default.

waprin:
@dhermes ready for another pass/merge

logging/google/cloud/logging/logger.py
```diff
@@ -91,7 +95,8 @@ def batch(self, client=None):
 
     def _make_entry_resource(self, text=None, info=None, message=None,
                              labels=None, insert_id=None, severity=None,
-                             http_request=None, timestamp=None):
+                             http_request=None, timestamp=None,
+                             resource=_GLOBAL_RESOURCE):
```
jonparrott: Does it make sense (in the context of the API) to make the resource default to `_GLOBAL_RESOURCE` instead of leaving it empty and letting the server assign the default resource? Does leaving it empty still make the server use the global resource or is it a required parameter?

If the server allows it to be optional, I strongly feel that we should not be baking / duplicated server-side behavior/defaults into the client.

logging/tests/system.py
jonparrott: Does the system test default to grpc?

logging/google/cloud/logging/logger.py
waprin: The server doesn't assign a default resource, it's required that it's in either the entry or the batch.

waprin:
System tests default to whatever the Client defaults to which is usually gRPC unless it's not supported. It seems a reasonable idea to parametrize the system tests to run with gRPC on and off, but don't think that's particular to these tests. Will create separate issue.

logging/google/cloud/logging/logger.py
jonparrott: Cool, confirmed:

```
POST https://logging.googleapis.com/v2/entries:write
{
  "entries": [
    {
      "textPayload": "Test test"
    }
  ],
  "logName": "projects/python-docs-samples-tests/logs/test"
}

Response 400:
{
  "error": {
    "code": 400,
    "message": "A monitored resource must be specified for each log entry",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.DebugInfo",
        "detail": "[ORIGINAL ERROR] generic::invalid_argument: A monitored resource must be specified for each log entry [google.rpc.error_details_ext] { message: \"A monitored resource must be specified for each log entry\" }"
      }
    ]
  }
}
```

jaitaiwan:
Hey guys, when will this be released in the google-cloud-logging package?

dhermes:
@lukesneeringer I can do a release today. WDYT?

bjwatson:
@tseaver I've had an internal request to release this Logging fix ASAP. Since you're planning to do new releases this week, this should be covered, but I just wanted to make sure.

tseaver:
@bjwatson Still on my radar:  at this point I will be making releases over the weekend, to catch up with some unplanned sitting-in-waiting-room time this week.

bjwatson:
Thanks @tseaver! Sorry to hear about excessive waiting room time this week.

Issue #3017, opened 2017-02-15
Add system tests for variations on spanner read operations


tseaver:
- Read with indexes
- Read with limits
- Different concurrency options
- Range reads

tseaver:
#3399 and #3404 should close this issue: we are already testing both transactional reads and snapshots with time boundaries.

PR #3399, opened 2017-05-11
Add system test for read-with-index case.

tseaver:
Toward #3017.


tseaver:
Review as part of #3404

PR #3401, opened 2017-05-11
Bigtable records bulk update (MutateRows RPC call)

skyfox:
Implementation of bulk updates for DirectRow (MutateRows RPC call)


bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
         message.rows.row_ranges.add(**range_kwargs)
 
     return message
+
+
+def _mutate_rows_request(table_name, rows):
+    """Creates a request to read rows in a table.
```
garye: "read rows" -> "mutate rows"

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+
+    :type table_name: str
+    :param table_name: The name of the table to read from.
```
garye: "read from" -> "write to"

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    request_pb = data_messages_v2_pb2.MutateRowsRequest(table_name=table_name)
+    mutations_count = 0
+    for row in rows:
+        entry = request_pb.entries.add()
+        entry.row_key = row.row_key
+        for mutation in row._get_mutations(None):
+            mutations_count += 1
+            entry.mutations.add().CopyFrom(mutation)
+    if mutations_count > 100000:
+        raise TooManyMutationsError('Maximum number of the entries mutations '
```
garye: "Maximum number of the entries mutations" -> "Maximum number of mutations"

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+                                    'is 100000')
+    return request_pb
+
+
+def _check_rows(table_name, rows):
+    """Checks that all rows belong to the table.
```
garye: This isn't quite accurate since the function also checks mutation type.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -29,6 +29,14 @@
 from google.cloud.bigtable.row_data import PartialRowsData
 
 
+class RowBelongingError(Exception):
```
garye: I think the name could be clearer. "TableMismatchError" perhaps?

bigtable/google/cloud/bigtable/row.py
```diff
@@ -55,6 +55,13 @@ def __init__(self, row_key, table):
         self._row_key = _to_bytes(row_key)
         self._table = table
 
+    @property
+    def row_key(self):
+        return self._row_key
+
+    @property
+    def table(self):
+        return self._table
```
dhermes: @skyfox We could probably pull these two `@property`-s out into a separate PR since they are not controversial or complex.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -29,6 +29,14 @@
...
+    """Row from another table."""
+
+
+class TooManyMutationsError(Exception):
+    """The number of mutations for bulk request is too big."""
```
dhermes: In general I (and this library) is/are :-1: on custom exceptions. For example, this one would be just fine as a `ValueError`.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +284,31 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
 
+    def mutate_rows(self, rows):
+        """Mutates multiple rows in bulk.
+    
+        :type rows: list
+        :param rows: List or other iterable of :class:`.DirectRow` instances.
+        
+        :rtype: list
+        :returns: A list of tuples (``MutateRowsResponse.Entry`` protobuf 
+                  corresponding to the errors, :class:`.DirectRow`)
```
dhermes: I'm not sure what this phrase in parentheses is trying to indicate.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +284,31 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        _check_rows(self.name, rows)
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
+
+        unsuccessfully_mutated_rows = []
+        client = self._instance._client
+        responses = client._data_stub.MutateRows(mutate_rows_request)
+        for response in responses:
+            for entry in response.entries:
+                if not entry.status.code:
+                    rows[entry.index].clear()
```
dhermes: Why is it necessary to clear the successfully mutated `row`? If it **is** necessary, you should definitely mention it in the docstring.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +284,31 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        """
+        _check_rows(self.name, rows)
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
+
+        unsuccessfully_mutated_rows = []
+        client = self._instance._client
+        responses = client._data_stub.MutateRows(mutate_rows_request)
+        for response in responses:
+            for entry in response.entries:
+                if not entry.status.code:
```
dhermes: So each [`Entry`][1] has a `Status` and [`Status.code`][2] is an `int32`. Can you change this check instead to:

```python
# 0 is empty value for int32, indicates code is unset
if entry.status.code == 0:
    ...
```

[1]: https://github.com/googleapis/googleapis/blob/e0922c4ea8e2f03c11f5dce405b79ecdf110792b/google/bigtable/v2/bigtable.proto#L255
[2]: https://github.com/googleapis/googleapis/blob/e0922c4ea8e2f03c11f5dce405b79ecdf110792b/google/rpc/status.proto#L82

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+
+    :type table_name: str
+    :param table_name: The name of the table to read from.
+
+    :type rows: list
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+    
+    :rtype: :class:`data_messages_v2_pb2.MutateRowsRequest`
+    :returns: The ``MutateRowsRequest`` protobuf corresponding to the inputs.
+    :raises: :class:`TooManyMutationsError <exceptions.TooManyMutationsError>`
```
dhermes: `TooManyMutationsError` is defined in `table`, instead you want:

```
:exc:`~.table.TooManyMutationsError`
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    :type table_name: str
+    :param table_name: The name of the table to read from.
+
+    :type rows: list
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+    
+    :rtype: :class:`data_messages_v2_pb2.MutateRowsRequest`
+    :returns: The ``MutateRowsRequest`` protobuf corresponding to the inputs.
+    :raises: :class:`TooManyMutationsError <exceptions.TooManyMutationsError>` 
+             if the number of mutations is grater than 100.000
```
dhermes: - s/grater/greater
- 100.000 here means 100,000 (different "thousands" separator)?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    """
+    request_pb = data_messages_v2_pb2.MutateRowsRequest(table_name=table_name)
+    mutations_count = 0
+    for row in rows:
+        entry = request_pb.entries.add()
+        entry.row_key = row.row_key
+        for mutation in row._get_mutations(None):
+            mutations_count += 1
+            entry.mutations.add().CopyFrom(mutation)
+    if mutations_count > 100000:
```
dhermes: Can you make `100000` a non-public module level constant? (Feel free to make it public if you think it's worth documenting.) If nothing else, this will make testing easier, since you could swap that constant for a smaller number.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    :returns: The ``MutateRowsRequest`` protobuf corresponding to the inputs.
+    :raises: :class:`TooManyMutationsError <exceptions.TooManyMutationsError>` 
+             if the number of mutations is grater than 100.000
+    """
+    request_pb = data_messages_v2_pb2.MutateRowsRequest(table_name=table_name)
+    mutations_count = 0
+    for row in rows:
+        entry = request_pb.entries.add()
+        entry.row_key = row.row_key
+        for mutation in row._get_mutations(None):
```
dhermes: Using `row._get_mutations(None)` here will be "bad" if `row` were accidentally a `ConditionalRow` (see my suggestion about making `_check_rows` into a local `_check_row` in this loop)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+
+def _check_rows(table_name, rows):
+    """Checks that all rows belong to the table.
+
+    :type table_name: str
+    :param table_name: The name of the table to read from.
+
+    :type rows: list
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+    """
```
dhermes: Please add a `raises` section?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +284,31 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+    
+        :type rows: list
+        :param rows: List or other iterable of :class:`.DirectRow` instances.
+        
+        :rtype: list
+        :returns: A list of tuples (``MutateRowsResponse.Entry`` protobuf 
+                  corresponding to the errors, :class:`.DirectRow`) 
+        """
+        _check_rows(self.name, rows)
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
```
dhermes: This is problematic because `_check_rows` will **exhaust** the passed in iterator `rows` before you iterate through it again in `_mutate_rows_request`. So this would **only** work if `rows` was something like a `list` / `tuple` / etc. that can be iterated over many times.

Suggested fix: convert `_check_rows` into a `_check_row` helper and then apply it within `_mutate_rows_request` as you iterate through `rows`.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+    """
+    for row in rows:
+        if not isinstance(row, DirectRow):
+            raise TypeError('Bulk processing can not be applied for '
+                            'conditional or append mutations.')
+        if row.table.name != table_name:
+            raise RowBelongingError(
+                'Row %s is a part of %s table. Current table: %s' %
+                (row.row_key, row.table.name, table_name))
```
dhermes: Please end the file with a newline (it's worth configuring your editor to do this by default)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +406,50 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+
+    :type rows: list
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+    """
+    for row in rows:
+        if not isinstance(row, DirectRow):
+            raise TypeError('Bulk processing can not be applied for '
+                            'conditional or append mutations.')
+        if row.table.name != table_name:
+            raise RowBelongingError(
```
dhermes: This could simply be a `ValueError`. I suggest you remove the custom exception.

bigtable/tests/unit/test_row.py
```diff
@@ -16,6 +16,26 @@
 import unittest
 
 
+class TestRow(unittest.TestCase):
```
dhermes: Thanks for adding this. Slap on the wrist (to me) that `TestRow` didn't already exist.

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,53 @@
...
+
+    def _call_fut(self, table_name, rows):
+        from google.cloud.bigtable.table import _mutate_rows_request
+
+        return _mutate_rows_request(table_name, rows)
+
+    def test__mutate_rows_too_many_mutations(self):
+        from google.cloud.bigtable.table import TooManyMutationsError
+        from google.cloud.bigtable.row import DirectRow
+        from google.cloud.bigtable._generated.data_pb2 import Mutation
```
dhermes: Small nit: Imports "should" be in alphabetical order

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,53 @@
...
+
+        mutation = Mutation()
+        rows = [DirectRow(row_key=b'row_key', table='table'),
+                DirectRow(row_key=b'row_key_2', table='table')]
+        rows[0]._pb_mutations = [mutation for _ in range(0, 50000)]
+        rows[1]._pb_mutations = [mutation for _ in range(0, 50001)]
```
dhermes: Here is where I was suggesting just using fewer mutations by mocking out the threshold.

bigtable/tests/unit/test_table.py
```diff
@@ -569,6 +616,11 @@ def _SampleRowKeysRequestPB(*args, **kw):
 
     return messages_v2_pb2.SampleRowKeysRequest(*args, **kw)
 
+def _MutateRowsRequestPB(*args, **kw):
```
dhermes: - Please us PEP8 function names (`_mutate_rows_request_pb`) (I realize there is already a violation of that in this file but let's not pile on)
- Two empty lines between top-level objects (so you need an extra one before this function and after it)

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,53 @@
 import unittest
 
 
+class Test___mutate_rows_request(unittest.TestCase):
```
dhermes: This test module should also have a test for the newly added `Table.mutate_rows`.

bigtable/google/cloud/bigtable/row.py
skyfox: Done #3408

bigtable/google/cloud/bigtable/table.py
lukesneeringer: I disagree; I prefer custom exceptions so the user can be specific about what they want to catch. That said, please subclass `ValueError` rather than `Exception`.

bigtable/google/cloud/bigtable/table.py
skyfox: It tries to indicate that each element of the list is a tuple with 2 values: the first one is an instance of ``MutateRowsResponse.Entry`` and the second one is DirectRow instance itself.
Should I change description or return type in some way?

bigtable/google/cloud/bigtable/table.py
skyfox: Done for _check_row. Keep None here, because it is unused in DirectRow but still required bool argument.
_get_mutations is a protected member. Maybe it is better to add a property (getter only) with public access?

bigtable/google/cloud/bigtable/table.py
skyfox: I would prefer to inherit from ValueError, if it is OK in a scope of the project.

skyfox:
Hey! Thank you for your comments, guys.

I fixed almost all of your comments and leave some comments for other. @garye @dhermes could you please take another look?

skyfox:
Also, could you please take a look why cover failed? Strings it points to are covered... Can't figure out.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +413,65 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+    :type table_name: str
+    :param table_name: The name of the table to write to.
+
+    :type rows: list
+    :param rows: List or other iterable of :class:`.DirectRow` instances.
+
+    :rtype: :class:`data_messages_v2_pb2.MutateRowsRequest`
+    :returns: The ``MutateRowsRequest`` protobuf corresponding to the inputs.
+    :raises: :exc:`~.table.TooManyMutationsError` if the number of mutations is
+             grater than 100,000
```
garye: nit: typo "grater"

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,34 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
         # We expect an iterator of `data_messages_v2_pb2.ReadRowsResponse`
         return PartialRowsData(response_iterator)
 
+    def mutate_rows(self, rows):
```
garye: I'm not sure I like that the input rows are modified. Can we leave those untouched, and return a list that is the same size as the input rows, and each element will contain a status of the corresponding mutation? So the logic could be something simple like:

```
response = []
for each entry:
  response.append(entry.status)
return response
```

This is consistent with the Go client, for example.

bigtable/tests/unit/test_table.py
skyfox: Done.

bigtable/google/cloud/bigtable/table.py
skyfox: Fixed.

bigtable/google/cloud/bigtable/table.py
skyfox: Done.

skyfox: I'm not sure only about the docstring. Could you please take a look?

skyfox:
Fixed. Could you please take another look?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -300,21 +300,19 @@ def mutate_rows(self, rows):
         :param rows: List or other iterable of :class:`.DirectRow` instances.
 
         :rtype: list
-        :returns: A list of tuples (``MutateRowsResponse.Entry`` protobuf
-                  corresponding to the errors, :class:`.DirectRow`)
+        :returns: A list of corresponding to each row statuses.
```
garye: "to each row status"

dhermes:
@garye @skyfox Will do my best to review today.

garye:
Thanks a lot @dhermes. We appreciate all the effort.

dhermes:
Sorry I didn't have a chance today guys. I'll try to take a look this weekend / ASAP Tuesday.

garye:
No worries @dhermes, next week is fine!

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,32 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
+        client = self._instance._client
+        responses = client._data_stub.MutateRows(mutate_rows_request)
+
+        responses_statuses = [None for _ in range(len(rows))]
+        for response in responses:
+            for entry in response.entries:
+                responses_statuses[entry.index] = entry.status
+                if entry.status.code == 0:
+                    rows[entry.index].clear()
```
dhermes: This mutates `rows` but @garye comment above seems to indicate this is not the desired behavior? Also, if `rows` is in iterable, you may not be able to call `__getitem__` (i.e. `rows[entry.index]`)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +411,65 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+        _check_row_table_name(table_name, row)
+        _check_row_type(row)
+        entry = request_pb.entries.add()
+        entry.row_key = row.row_key
+        for mutation in row._get_mutations(None):
+            mutations_count += 1
+            entry.mutations.add().CopyFrom(mutation)
+    if mutations_count > _MAX_BULK_MUTATIONS:
+        raise TooManyMutationsError('Maximum number of mutations is %s' %
+                                    _MAX_BULK_MUTATIONS)
```
dhermes: Always pass a tuple to a format string. So it'd be

```python
    if mutations_count > _MAX_BULK_MUTATIONS:
        raise TooManyMutationsError(
            'Maximum number of mutations is %s' % (_MAX_BULK_MUTATIONS,))
```

or alternatively you could just forego the format string

```python
    if mutations_count > _MAX_BULK_MUTATIONS:
        raise TooManyMutationsError(
            'Maximum number of mutations', _MAX_BULK_MUTATIONS)
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +411,65 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+             greater than 100,000
+    """
+    request_pb = data_messages_v2_pb2.MutateRowsRequest(table_name=table_name)
+    mutations_count = 0
+    for row in rows:
+        _check_row_table_name(table_name, row)
+        _check_row_type(row)
+        entry = request_pb.entries.add()
+        entry.row_key = row.row_key
+        for mutation in row._get_mutations(None):
```
dhermes: Can you add a note as to why `state=None` is correct here? Something like:

```python
# NOTE: Since `_check_row_type` has verified `row` is a `DirectRow`, the
#       mutations have no state.
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +411,65 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+            mutations_count += 1
+            entry.mutations.add().CopyFrom(mutation)
+    if mutations_count > _MAX_BULK_MUTATIONS:
+        raise TooManyMutationsError('Maximum number of mutations is %s' %
+                                    _MAX_BULK_MUTATIONS)
+    return request_pb
+
+
+def _check_row_table_name(table_name, row):
+    """Checks that a row belong to the table.
```
dhermes: Instead:

```
Checks that a row belongs to a table.
```

bigtable/tests/system.py
```diff
@@ -356,6 +356,32 @@ def _write_to_row(self, row1=None, row2=None, row3=None, row4=None):
...
 
+    def test_mutate_rows(self):
+        row_1 = self._table.row(ROW_KEY)
+        row_1.set_cell(COLUMN_FAMILY_ID1, COL_NAME1, CELL_VAL1)
+        row_1.commit()
+        row_2 = self._table.row(ROW_KEY_ALT)
+        row_2.set_cell(COLUMN_FAMILY_ID1, COL_NAME1, CELL_VAL2)
+        row_2.commit()
+        rows = [row_1, row_2]
+        self.rows_to_delete.extend(rows)
```
dhermes: Please add to `rows_to_delete` **before** calling `commit`, since `row_2.commit()` could fail before `row_1` was adding to `rows_to_delete`.

Also, you can just use `row1` and `row2` (i.e. no underscore needed)

bigtable/tests/system.py
```diff
@@ -356,6 +356,32 @@ def _write_to_row(self, row1=None, row2=None, row3=None, row4=None):
...
+        row_2.set_cell(COLUMN_FAMILY_ID1, COL_NAME1, CELL_VAL4)
+        statuses = self._table.mutate_rows(rows)
+        result = [status.code for status in statuses]
+        expected_result = [0, 0]
+        self.assertEqual(result, expected_result)
+
+        # Check the contents
+        row_1_data = self._table.read_row(ROW_KEY)
+        self.assertEqual(
+            row_1_data.cells[COLUMN_FAMILY_ID1][COL_NAME1][0].value, CELL_VAL3)
```
dhermes: Between `getattr` and `__getitem__`, there are 5 layers deep of accesses in:

```
row_1_data.cells[COLUMN_FAMILY_ID1][COL_NAME1][0].value
```

Can we not just construct what `cells` should be a do a full comparison?

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        return _mutate_rows_request(table_name, rows)
+
+    def test__mutate_rows_too_many_mutations(self):
+        from collections import namedtuple
+        from google.cloud.bigtable._generated.data_pb2 import Mutation
+        from google.cloud.bigtable.row import DirectRow
+        import google.cloud.bigtable.table as table_module
+        from google.cloud.bigtable.table import TooManyMutationsError
+
+        table = namedtuple('Table', ['name'])
```
dhermes: Instead, just `import mock` at module level and then:

```python
table = mock.Mock(name='table', spec=['name'])
```

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        table.name = 'table'
+        table_module._MAX_BULK_MUTATIONS = 3
```
dhermes: This is not a good idea, because `table_module` will continue to exist after this test and you have permanently modified it. See my suggestion above for what to do instead.

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        mutation = Mutation()
+        rows = [DirectRow(row_key=b'row_key', table=table),
+                DirectRow(row_key=b'row_key_2', table=table)]
+        rows[0]._pb_mutations = [mutation, mutation]
+        rows[1]._pb_mutations = [mutation, mutation]
```
dhermes: Since there are only 4 mutations, why not just call `set_cell` 4 times rather than mucking with internals?

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
 
 
+class Test___mutate_rows_request(unittest.TestCase):
+
+    def _call_fut(self, table_name, rows):
+        from google.cloud.bigtable.table import _mutate_rows_request
+
+        return _mutate_rows_request(table_name, rows)
+
+    def test__mutate_rows_too_many_mutations(self):
```
dhermes: Decorate this method as follows:

```python
    @mock.patch('google.cloud.bigtable.table._MAX_BULK_MUTATIONS',
                new=3)
    def test__mutate_rows_too_many_mutations(self):
        ...
```

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+
+
+class Test__check_row_table_name(unittest.TestCase):
+
+    def _call_fut(self, table_name, row):
+        from google.cloud.bigtable.table import _check_row_table_name
+
+        return _check_row_table_name(table_name, row)
+
+    def test__check_rows_wrong_table_name(self):
```
dhermes: You need to also add a test case that doesn't fail.

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        table = namedtuple('Table', ['name'])
+        table.name = 'table'
+        table_module._MAX_BULK_MUTATIONS = 3
+        mutation = Mutation()
+        rows = [DirectRow(row_key=b'row_key', table=table),
+                DirectRow(row_key=b'row_key_2', table=table)]
+        rows[0]._pb_mutations = [mutation, mutation]
+        rows[1]._pb_mutations = [mutation, mutation]
+        with self.assertRaises(TooManyMutationsError):
+            self._call_fut('table', rows)
```
dhermes: You need to also add a test case that doesn't fail

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        from google.cloud.bigtable.table import _check_row_table_name
+
+        return _check_row_table_name(table_name, row)
+
+    def test__check_rows_wrong_table_name(self):
+        from collections import namedtuple
+        from google.cloud.bigtable.table import TableMismatchError
+        from google.cloud.bigtable.row import DirectRow
+
+        table = namedtuple('Table', ['name'])
```
dhermes: Use a `mock.Mock` instead

bigtable/tests/unit/test_table.py
```diff
@@ -570,6 +661,12 @@ def _SampleRowKeysRequestPB(*args, **kw):
     return messages_v2_pb2.SampleRowKeysRequest(*args, **kw)
 
 
+def _mutate_rows_request_pb(*args, **kw):
+    from google.cloud.bigtable._generated import (
+        bigtable_pb2 as data_messages_v2_pb2)
+
+    return data_messages_v2_pb2.MutateRowsRequest(*args, **kw)
+
```
dhermes: Two empty lines between top-level definitions please

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+        from google.cloud.bigtable.table import _check_row_type
+
+        return _check_row_type(table_name, row)
+
+    def test__check_rows_wrong_row_type(self):
+        from google.cloud.bigtable.row import ConditionalRow
+
+        row = ConditionalRow(row_key=b'row_key', table='table', filter_=None)
+        with self.assertRaises(TypeError):
+            self._call_fut('table', row)
```
dhermes: You also need a test case that doesn't fail

bigtable/tests/unit/test_table.py
```diff
@@ -16,6 +16,65 @@
...
+            self._call_fut('other_table', row)
+
+
+class Test__check_row_type(unittest.TestCase):
+    def _call_fut(self, table_name, row):
+        from google.cloud.bigtable.table import _check_row_type
+
+        return _check_row_type(table_name, row)
+
+    def test__check_rows_wrong_row_type(self):
```
dhermes: You can just call this: `test_wrong_row_type` (the class name already tells us that we are `Test__check_row_type`)

bigtable/tests/unit/test_table.py
```diff
@@ -348,6 +407,38 @@ def test_read_row_still_partial(self):
...
+        client = _Client()
+        instance = _Instance(self.INSTANCE_NAME, client=client)
+        table = self._make_one(self.TABLE_ID, instance)
+
+        row_1 = DirectRow(row_key=b'row_key', table=table)
+        row_1.set_cell('cf', b'col', b'value1')
+        row_2 = DirectRow(row_key=b'row_key_2', table=table)
+        row_2.set_cell('cf', b'col', b'value2')
+
+        response = MutateRowsResponse()
```
dhermes: I would prefer:

```python
from google.rpc.status_pb2 import Status  # up with imports

...

response = MutateRowsResponse(
    entries=[
        MutateRowsResponse.Entry(
            index=0,
            status=Status(code=0),
        ),
        MutateRowsResponse.Entry(
            index=1,
            status=Status(code=1),
        ),
    ],
)
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,32 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        :param rows: List or other iterable of :class:`.DirectRow` instances.
+
+        :rtype: list
+        :returns: A list of corresponding to each row statuses.
+        """
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
+        client = self._instance._client
+        responses = client._data_stub.MutateRows(mutate_rows_request)
+
+        responses_statuses = [None for _ in range(len(rows))]
```
dhermes: - If `rows` is an iterator, you may not be able to call `len` (you could address this by getting `num_rows` as a return value from `_mutate_rows_request` or using `len(mutate_rows_request.entries)`)
- If `rows` is very large you'd much rather use `six.moves.xrange` than `range`

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,32 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        for response in responses:
+            for entry in response.entries:
+                responses_statuses[entry.index] = entry.status
+                if entry.status.code == 0:
+                    rows[entry.index].clear()
+        return responses_statuses
```
dhermes: Do you want it to be possible for `None` to be one of the entries in the return value? Right now, it is possible if the server doesn't provide a one-for-one mapping.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,32 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        :returns: A list of corresponding to each row statuses.
+        """
+        mutate_rows_request = _mutate_rows_request(self.name, rows)
+        client = self._instance._client
+        responses = client._data_stub.MutateRows(mutate_rows_request)
+
+        responses_statuses = [None for _ in range(len(rows))]
+        for response in responses:
+            for entry in response.entries:
+                responses_statuses[entry.index] = entry.status
```
dhermes: What if this status has **already** been set? (i.e. how much do we want to trust the server not to overwrite?)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -276,6 +288,32 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
...
+        The method tries to update all specified rows.
+        If some of the rows weren't updated, it would not remove mutations.
+        They can be applied to the row separately.
+        If row mutations finished successfully, they would be cleaned up.
+
+        :type rows: list
+        :param rows: List or other iterable of :class:`.DirectRow` instances.
+
+        :rtype: list
+        :returns: A list of corresponding to each row statuses.
```
dhermes: Instead:

```
A list of response statuses (`google.rpc.status_pb2.Status`)
corresponding to success or failure of each row mutation sent. These will
be in the same order as the `rows`.
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -373,3 +411,65 @@ def _create_row_request(table_name, row_key=None, start_key=None, end_key=None,
...
+
+    :type row: :class:`.Row`
+    :param row: An instance of :class:`.Row` subclasses.
+
+    :raises: :class:`TypeError <exceptions.TypeError>` if the row is not an
+             instance of DirectRow.
+    """
+    if not isinstance(row, DirectRow):
+        raise TypeError('Bulk processing can not be applied for '
+                        'conditional or append mutations.')
```
tseaver: Line 474 (and the branch to it) are missing test coverage.

bigtable/tests/unit/test_table.py
tseaver: Or just delete it (never triggered during the test run).

bigtable/google/cloud/bigtable/table.py
```diff
@@ -29,6 +29,18 @@
 from google.cloud.bigtable.row_data import PartialRowsData
 
 
+# Maximum number of mutations in bulk
+_MAX_BULK_MUTATIONS = 100000
```
tseaver: Is this a documented back-end limit?  Or just a SWAG?

skyfox: @tseaver It is documented back-end limit: https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#mutaterowsrequest

tseaver: Then the comment should reference that URL (e.g., for verification, or if the back-end should change the limit in the future).

bigtable/google/cloud/bigtable/table.py
skyfox: We can trust the server: https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#mutaterowsresponse
Server has to return only one record per row:
"The index into the original request's entries list of the Entry for which a result is being reported."

bigtable/google/cloud/bigtable/table.py
skyfox: 'rows' has to be iterable. It is specified in a docstring.
Should I add an explicit data type check? (rows is iterable)

bigtable/google/cloud/bigtable/table.py
skyfox: Based on the RPC 
 interface (https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2) it is impossible. Server always responses the same number of records with their statuses.

bigtable/google/cloud/bigtable/table.py
skyfox: I see the same picture. But the line is covered by test_table.py/Test__check_row_type.test__check_rows_wrong_row_type
Can figure out, why linter thinks it is not.

bigtable/tests/system.py
skyfox: I would prefer to keep the current version. 
When I construct cells, they are the different instances of Cell class, because the system test return true value from test BigTable server.
I can either mock the response from BT or compare exact values of Cell instances.
In the first case the test will became not system anymore. It the second: I won't avoid `getattr`/`__getitem__`.

bigtable/google/cloud/bigtable/table.py
skyfox: Never-mind. Done.

bigtable/google/cloud/bigtable/table.py
skyfox: Done.

bigtable/tests/unit/test_table.py
skyfox: Reused for a new test.

skyfox:
Hey! Could you please take another look?

@dhermes asked: "Why are only DirectRow instances allowed? Is that part of the method or just part of the implementation?"
Only DirectRows (primitive mutations-based) can be processed in bulk.

PR #3404, opened 2017-05-11
Add tests for range reads and reads with limits. 

### LINKED ISSUE #3017: Add system tests for variations on spanner read operations

tseaver:
~~Uses #3399 as a base.~~

Closes #3017.

### DIFF
--- a/spanner/tests/system.py
+++ b/spanner/tests/system.py
@@ -563,6 +563,36 @@ class TestSessionAPI(unittest.TestCase, _TestData):
             [(row[0], row[2]) for row in self._row_data(ROW_COUNT)]))
         self._check_row_data(rows, expected)
 
+    def test_read_w_limit(self):
+        ROW_COUNT = 4000
+        LIMIT = 100
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, self.ALL, limit=LIMIT))
+
+        all_data_rows = list(self._row_data(ROW_COUNT))
+        expected = all_data_rows[:LIMIT]
+        self._check_row_data(rows, expected)
+
+    def test_read_w_range(self):
+        from google.cloud.spanner.keyset import KeyRange
+        ROW_COUNT = 4000
+        START_CLOSED = 1000
+        END_OPEN = 2000
+        session, committed = self._set_up_table(ROW_COUNT)
+        key_range = KeyRange(start_closed=[START_CLOSED], end_open=[END_OPEN])
+        keyset = KeySet(ranges=(key_range,))
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, keyset))
+
+        all_data_rows = list(self._row_data(ROW_COUNT))
+        expected = all_data_rows[START_CLOSED:END_OPEN]
+        self._check_row_data(rows, expected)
+
     def test_execute_sql_w_manual_consume(self):
         ROW_COUNT = 4000
         session, committed = self._set_up_table(ROW_COUNT)
--- a/spanner/tests/system.py
+++ b/spanner/tests/system.py
@@ -492,24 +492,29 @@ class TestSessionAPI(unittest.TestCase, _TestData):
             yield [
                 index,
                 'First%09d' % (index,),
-                'Last09%d' % (index),
+                'Last%09d' % (max_index - index),
                 'test-%09d@example.com' % (index,),
             ]
 
-    def _set_up_table(self, row_count):
-        retry = RetryInstanceState(_has_all_ddl)
-        retry(self._db.reload)()
+    def _set_up_table(self, row_count, db=None):
 
-        session = self._db.session()
+        if db is None:
+            db = self._db
+            retry = RetryInstanceState(_has_all_ddl)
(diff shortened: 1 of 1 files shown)


spanner/tests/system.py
```diff
@@ -492,24 +492,29 @@ def _row_data(max_index):
...
 
-        with session.transaction() as transaction:
-            transaction.delete(self.TABLE, self.ALL)
+        def _unit_of_work(transaction, test):
+            transaction.delete(test.TABLE, test.ALL)
             transaction.insert(
-                self.TABLE, self.COLUMNS, self._row_data(row_count))
+                test.TABLE, test.COLUMNS, test._row_data(row_count))
+
+        committed = session.run_in_transaction(_unit_of_work, test=self)
```
dhermes: Why the switch from the context manager to `run_in_transaction`?

spanner/tests/system.py
```diff
@@ -531,6 +536,60 @@ def test_read_w_manual_consume(self):
         self.assertEqual(streamed._current_row, [])
         self.assertEqual(streamed._pending_chunk, None)
 
+    def test_read_w_index(self):
+        ROW_COUNT = 2000
```
dhermes: I will say this, even though I know you won't like it. `ROW_COUNT` is not a constant or a global, it's a local variable.

spanner/tests/system.py
```diff
@@ -531,6 +536,60 @@ def test_read_w_manual_consume(self):
...
+    def test_read_w_limit(self):
+        ROW_COUNT = 4000
+        LIMIT = 100
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, self.ALL, limit=LIMIT))
+
+        self._check_row_data(rows, list(self._row_data(ROW_COUNT))[:LIMIT])
```
dhermes: WDYT of instead:

```python
row_data = list(self._row_data(ROW_COUNT))
self.assertEqual(len(row_data), ROW_COUNT)
self._check_row_data(rows, row_data[:LIMIT])
```

spanner/tests/system.py
```diff
@@ -531,6 +536,60 @@ def test_read_w_manual_consume(self):
...
+
+    def test_read_w_range(self):
+        from google.cloud.spanner.keyset import KeyRange
+        ROW_COUNT = 4000
+        START_CLOSED = 1000
+        END_OPEN = 2000
+        session, committed = self._set_up_table(ROW_COUNT)
+        key_range = KeyRange(start_closed=[START_CLOSED], end_open=[END_OPEN])
```
dhermes: `start_closed` and `end_open` are the field names? WAT

spanner/tests/system.py
```diff
@@ -531,6 +536,60 @@ def test_read_w_manual_consume(self):
...
+        keyset = KeySet(ranges=(key_range,))
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, keyset))
+
+        self._check_row_data(
+            rows, list(self._row_data(ROW_COUNT))[START_CLOSED:END_OPEN])
```
dhermes: Ditto here on re-factoring this. This line is "doin' too much".

spanner/tests/system.py
```diff
@@ -662,3 +721,13 @@ def test_execute_sql_w_query_param(self):
             param_types={'my_list': array_type},
             expected=[(u'dog',), (u'cat',)],
         )
+
+
+class _DatabaseDropper(object):
```
dhermes: Ha this is a funny little invention.

spanner/tests/system.py
tseaver: - At the implementation level, Python doesn't have *any* [constants](https://en.wikipedia.org/wiki/Constant_(computer_programming)):  any name can be rebound at any time.
- This is a symbolic name for a literal value, which is not intended to be changed during its lifetime, which is the semantic of "constantness".
- The notion that constants can only be declared at module scope is not backed by PEP8, which says, "Constants are usually defined on a module level "  That is not a prescription that they may not be defined at other scopes.

spanner/tests/system.py
tseaver: Because `run_in_transaction` gets us [retry-on-abort](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.Commit) semantics for `Commit`.

spanner/tests/system.py
tseaver: See [`KeyRange](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.KeyRange).

spanner/tests/system.py
tseaver: Done.

spanner/tests/system.py
tseaver: OK

spanner/tests/system.py
dhermes: I get it, I'm not trying to be pedantic, I'm trying to be practical. This is **not** how people use `ALL_CAPS`.

spanner/tests/system.py
dhermes: :+1:

dhermes:
LGTM

Issue #3019, opened 2017-02-15
Add system tests which exercise stream chunking / resume


tseaver:
The Java system tests do this by populating the database with:

- 1000 rows of size 4 KB
- 100 rows of size 40 KB
- 25 rows of size 400KB
- 10 rows of size 4 MB

And then reading all this data back.

tseaver:
Implementation note:  the `setupClass` for this test should check for the existence of a specially-named database within its instance, and [raise a `unittest.SkipTest` exception](https://docs.python.org/2/library/unittest.html#setupclass-and-teardownclass) if it is not there.  This will allow developers to run the system tests without needing an expensive-to-build-and-maintain pre-populated database.

dhermes:
:+1:

tseaver:
@dhermes BTW, where did the script go which pre-populates the datastore indexes for system testing?

PR #3405, opened 2017-05-11
Scripts populating / cleaning DB for streaming / chunking systests.

tseaver:
Add scripts for pre-populating / cleaning the database used for testing streaming / chunking.

Toward #3019.


spanner/tests/system/utils/clear_streaming.py
```diff
@@ -0,0 +1,55 @@
...
+from google.cloud.spanner.keyset import KeySet
+from google.cloud.spanner.pool import BurstyPool
+
+
+INSTANCE_NAME = 'gcp-streaming-systests'
+DATABASE_NAME = 'testing'
+
+
+def print_func(message):
+    if os.getenv('GOOGLE_CLOUD_NO_PRINT') != 'true':
```
dhermes: Why not make this into a boolean at import time?

spanner/tests/system/utils/populate_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+    chunk_me STRING(409600) )
+    PRIMARY KEY (pkey);
+CREATE TABLE four_meg (
+    pkey INT64,
+    chunk_me STRING(2097152),
+    chunk_me_2 STRING(2097152) )
+    PRIMARY KEY (pkey);
+"""
+
+DDL_STATEMENTS = [stmt.strip() for stmt in DDL.split(';') if stmt.strip()]
```
dhermes: What's the point of doing the split if `DDL` is hardcoded anyway? Just hardcode this `DDL_STATEMENTS` variable instead?

spanner/tests/system/utils/populate_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+
+import os
+
+from google.cloud.spanner import Client
+from google.cloud.spanner.keyset import KeySet
+from google.cloud.spanner.pool import BurstyPool
+
+
+INSTANCE_NAME = 'gcp-streaming-systests'
+DATABASE_NAME = 'testing'
```
dhermes: This name is repeated from the other module. This is actually important to DRY.

spanner/tests/system/utils/populate_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+    pkey INT64,
+    chunk_me STRING(2097152),
+    chunk_me_2 STRING(2097152) )
+    PRIMARY KEY (pkey);
+"""
+
+DDL_STATEMENTS = [stmt.strip() for stmt in DDL.split(';') if stmt.strip()]
+
+
+def print_func(message):
```
dhermes: This is copy-pasta. Maybe a `utils` module for shared code and constants?

spanner/tests/system/utils/populate_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+        chunk_me = 'X' * val_size
+        row_data = [(index, chunk_me) for index in range(row_count)]
+        with database.batch() as batch:
+            batch.delete(table_name, all_)
+            batch.insert(table_name, columns, row_data)
+    else:
+        print_func("Leaving table: {}".format(table_name))
+
+
+def populate_table_2(database, table_name, row_count, val_size):
```
dhermes: Can you give a better name than `populate_table_2` here?

spanner/tests/system/utils/populate_streaming.py
tseaver: The SQL is easier to read / write as a chunk of text, vs. a list of lines.

dhermes: How? It's the same text:

```python
DDL_STATEMENTS = [
    ('CREATE TABLE four_kay (\n'
     '    pkey INT64,\n'
     '    chunk_me STRING(4096) )\n'
     '    PRIMARY KEY (pkey)'),
    ('CREATE TABLE forty_kay (\n'
     '    pkey INT64,\n'
     '    chunk_me STRING(40960) )\n'
     '    PRIMARY KEY (pkey)'),
    ('CREATE TABLE four_hundred_kay (\n'
     '    pkey INT64,\n'
     '    chunk_me STRING(409600) )\n'
     '    PRIMARY KEY (pkey)'),
    ('CREATE TABLE four_meg (\n'
     '    pkey INT64,\n'
     '    chunk_me STRING(2097152),\n'
     '    chunk_me_2 STRING(2097152) )\n'
     '    PRIMARY KEY (pkey)'),
]
```

spanner/tests/system/utils/populate_streaming.py
tseaver: Running scripts directly from deep inside the `tests/` tree does not end up with the path set to make anything under `tests/` importable.  We would need to run them as:

```bash
$ /path/to/python -m tests.system.utils.clean_streaming
$ /path/to/python -m tests.system.utils.populate_streaming
```

or else jam `os.getcwd()` into `sys.path`.

spanner/tests/system/utils/populate_streaming.py
tseaver: See above:  the `tests/` tree is not importable when running a script.  I have changes which jam it into `sys.path` in my local checkout, if that seems best.

spanner/tests/system/utils/populate_streaming.py
tseaver: 1b2b2e0

spanner/tests/system/utils/clear_streaming.py
tseaver: Don't look at me, I just cloned it from the version in `datastore`.

spanner/tests/system/utils/populate_streaming.py
tseaver: Do you see all the extra punctuation?  I can't take that and paste it anywhere else as SQL (e.g., into the console) without editing it.

spanner/tests/system/utils/populate_streaming.py
tseaver: ISTM that the least-worst variant is to just import the shared module as a top-level entity (relative to the script), rather than trying to get `tests` to be importable.  I've updated the branch accordingly in 5e822c0.

dhermes: @tseaver I don't understand the `PYTHONPATH` / import problem. We've had no issue else where:

```
$ git grep 'import' -- '*system*' | grep 'tests\.'
datastore/tests/system/test_system.py:from tests.system.utils import clear_datastore
datastore/tests/system/test_system.py:from tests.system.utils import populate_datastore
spanner/tests/system.py:from tests._fixtures import DDL_STATEMENTS
```

spanner/tests/system/utils/populate_streaming.py
dhermes: OK that I can agree with. Fine to leave as-is.

spanner/tests/system/utils/clear_streaming.py
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+
+"""Depopulate spanner databases with data for streaming system tests."""
+
+from google.cloud.spanner import Client
+from google.cloud.spanner.pool import BurstyPool
+
+# Import relative to the script's directory
+from streaming_utils import INSTANCE_NAME
+from streaming_utils import DATABASE_NAME
```
dhermes: Teeny nit: Switch these for alphabetical?

spanner/tests/system/utils/populate_streaming.py
tseaver: When running a script from another directory, `sys.path[0]` is *that* directory, rather than the CWD (which is not on the path anywhere in that case, unless one sets `PYTHONPATH`).  We therefore cannot import `tests.system.utils.foo` within such scripts, unless we set `sys.path` ourselves.

tseaver:
Test failures were due to the logging systest hang.

spanner/tests/system/utils/populate_streaming.py
dhermes: Well we can control (at least in `nox.py` and in CI) where they get run from (and make sure it's `${REPO_ROOT}/spanner`).

"Other usage beware"?

tseaver: One doesn't run the `populate_streaming.py` and `clean_streaming.py` scripts under `nox`:  they are designed to be run manually, out-of-band of normal test execution.

dhermes: Gotcha. I don't think that changes the idea that we can play a little "fast and loose" with them, i.e. make sure users know **where** to run them from.

@jonparrott WDYT?

jonparrott: I think it's totally fine to require nox to run the system tests.

dhermes: Unfortunately, that's not what we're discussing. These scripts are set-up for system tests, but they will create long-lived data.

So when "human A" runs them to set up a project, the imports might fail depending on (a) how we write the imports and (b) which directory "human A" runs the scripts from.

jonparrott: I think it's totally fine to require nox to run the system test setup scripts.

PR #3407, opened 2017-05-12
Overhaul logging background thread transport

jonparrott:


logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+    Gets at least one (blocking) and at most :param:`max` items (non-blocking)
+    from a given Queue. Does not mark the items as done.
 
-    Writes entries to the logger API.
+    :type q: ~queue.Queue
+    :param q: The Queue to get items from.
 
-    This class reuses a single :class:`Batch` method to write successive
-    entries.
+    :type max: int
```
duggelz: This is a real nit, but  hesitate using max as a variable name it shadows a builtin.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
 import copy
+import logging
 import threading
 
+from six.moves import queue
+
 from google.cloud.logging.handlers.transports.base import Transport
 
-_WORKER_THREAD_NAME = 'google.cloud.logging.handlers.transport.Worker'
+_DEFAULT_GRACE_PERIOD = 5.0
```
duggelz: Can you document the units here (seconds I assume)?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
 
-        # Number in seconds of  how long to wait for worker to send remaining
-        self._stop_timeout = 5
+        :type batch_size: int
+        :param batch_size: The maximum number of items to send at a time in the
+            background thread.
+        """
+        self._cloud_logger = cloud_logger
+        self._grace_period = grace_period
+        self._batch_size = batch_size
```
duggelz: Maybe this should be _max_batch_size?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        Additionally, this registers a handler for process exit to attempt
+        to send any pending log entries before shutdown.
         """
-        try:
-            self._entries_condition.acquire()
+        with self._operational_lock:
+            if self.is_alive:
+                return
+
             self._thread = threading.Thread(
```
duggelz: Consider an `assert not self._thread` since start() is a public method?

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
-        # Number in seconds of  how long to wait for worker to send remaining
-        self._stop_timeout = 5
+        :type max_batch_size: int
+        :param max_batch_size: The maximum number of items to send at a time
+            in the background thread.
+        """
+        self._cloud_logger = cloud_logger
+        self._grace_period = grace_period
+        self._max_batch_size = max_batch_size
+        self._queue = queue.Queue(-1)
```
dhermes: `0` will accomplish the same goal (and `0` is the default)

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        self._operational_lock = threading.Lock()
+        self._thread = None
+        self.start()
 
-        self._start()
+    @property
+    def is_alive(self):
+        """Returns True is the background thread is running."""
+        return self._thread and self._thread.is_alive()
```
dhermes: Can you make this an explicit `None` check (nunchuk)

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        :param max_batch_size: The maximum number of items to send at a time
+            in the background thread.
+        """
+        self._cloud_logger = cloud_logger
+        self._grace_period = grace_period
+        self._max_batch_size = max_batch_size
+        self._queue = queue.Queue(-1)
+        self._operational_lock = threading.Lock()
+        self._thread = None
+        self.start()
```
dhermes: :-1: on this side effect in the constructor. Why can't the `__init__` caller also `start`? Or invoke a factory constructor that also starts?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
 
-        self._start()
+    @property
+    def is_alive(self):
+        """Returns True is the background thread is running."""
+        return self._thread and self._thread.is_alive()
 
-    def _run(self):
+    def _thread_main(self):
```
dhermes: `_thread_main` -> `_thread_entry` / `_thread_main_entry`?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
-        self._stop_condition.release()
-
-    def _start(self):
-        """Called by this class's constructor
-
-        This method is responsible for starting the thread and registering
-        the exit handlers.
+        _LOGGER.debug('Background thread started.')
+
+        quit = False
```
dhermes: `quit` -> `quit_`

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        while True:
+            batch = self._cloud_logger.batch()
+            items = _get_many(self._queue, max_items=self._max_batch_size)
+
+            for item in items:
+                if item is _WORKER_TERMINATOR:
+                    quit = True
+                    # Continue, don't break, try to process all items we got
+                    # back before quitting.
+                    continue
```
dhermes: Why not just do an `else` block? The `continue` made me do a double take that the code below was OK.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+            try:
+                if total_logs:
+                    batch.commit()
+                    _LOGGER.debug('Submitted %d logs', total_logs)
+            except Exception:
+                _LOGGER.error(
+                    'Failed to submit %d logs.', total_logs, exc_info=True)
+            finally:
+                # Mark all collected tasks as done.
+                for n in range(len(items)):
```
dhermes: `for _ in ...`. Also, can you use `six.moves.range` (or are you philosophically opposed)?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+                    # Continue, don't break, try to process all items we got
+                    # back before quitting.
+                    continue
+
+                batch.log_struct(**item)
+
+            total_logs = len(batch.entries)
+
+            try:
+                if total_logs:
```
dhermes: `if total_logs > 0`

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+                    batch.commit()
+                    _LOGGER.debug('Submitted %d logs', total_logs)
+            except Exception:
```
dhermes: You don't actually need a catch-all, we know what `batch.commit()` can raise don't we?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+                    quit = True
+                    # Continue, don't break, try to process all items we got
+                    # back before quitting.
+                    continue
+
+                batch.log_struct(**item)
+
+            total_logs = len(batch.entries)
+
+            try:
```
dhermes: Can you put this `try/except/finally` block in a helper (for "read-a-billy")?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+                    'Failed to submit %d logs.', total_logs, exc_info=True)
+            finally:
+                # Mark all collected tasks as done.
+                for n in range(len(items)):
+                    self._queue.task_done()
+
+            if quit:
+                break
+
+        _LOGGER.debug('Background thread exited gracefully.')
```
dhermes: Who are you to say it was graceful?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
-            self._entries_condition.acquire()
+        with self._operational_lock:
+            if self.is_alive:
+                return
+
             self._thread = threading.Thread(
-                target=self._run, name=_WORKER_THREAD_NAME)
+                target=self._thread_main,
+                name=_WORKER_THREAD_NAME)
             self._thread.setDaemon(True)
```
dhermes: I believe this is an "old" API and should instead be:

```python
self._thread.daemon = True
```

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
-        if not self.started or self.stopping:
-            return
+        if not self.is_alive:
+            return True
+
+        with self._operational_lock:
+            self._queue.put_nowait(_WORKER_TERMINATOR)
+
+            if grace_period is not None:
+                print('Waiting up to %d seconds.' % grace_period)
```
dhermes: - Did you mean to log not print?
- Please pass `(grace_period,)` (explicit tuple is best, or just use `.format()`)

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+
+            self._thread.join(timeout=grace_period)
 
-        # lock the stop condition first so that the worker
-        # thread can't notify it's finished before we wait
-        self._stop_condition.acquire()
+            success = not self.is_alive
```
dhermes: Can you add a note that you want to check this **before** `self._thread is None`? (i.e. just returning `return not self.is_alive` **seems** the same, but it isn't)

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
-        self.stopped = True
+    def _main_thread_terminated(self):
+        """Callback that attempts to send pending logs before termination."""
+        if not self.is_alive:
+            return
+
+        if not self._queue.empty():
+            print(
+                'Program shutting down, attempting to send %d queued log '
+                'entries to Stackdriver Logging...' % self._queue.qsize())
```
dhermes: Ditto here:

- Did you want `print` not `log`?
- Please pass an explicit `tuple` to a format string

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+
+        if self.stop(self._grace_period):
+            print('Sent all pending logs.')
+        else:
+            print('Failed to send %d pending logs.' % self._queue.qsize())
```
dhermes: Ditto here:

- Did you want `print` not `log`?
- Please pass an explicit `tuple` to a format string

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
 
 import atexit
```
dhermes: Definitely want a `from __future__ import print_statement` if you actually want all of these `print`-s.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        :param record: Python log record that the handler was called with.
+
+        :type message: str
+        :param message: The message from the ``LogRecord`` after being
+                        formatted by the associated log formatters.
+        """
+        self._queue.put_nowait({
+            'info': {
+                'message': message,
+                'python_logger': record.name
```
dhermes: trailing comma?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+            },
+            'severity': record.levelname
```
dhermes: trailing comma?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        """Queues a log entry to be written by the background thread.
+
+        :type record: :class:`logging.LogRecord`
+        :param record: Python log record that the handler was called with.
+
+        :type message: str
+        :param message: The message from the ``LogRecord`` after being
+                        formatted by the associated log formatters.
+        """
+        self._queue.put_nowait({
```
dhermes: Are we worried about `is_alive` or any other state before adding to the queue?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+                'message': message,
+                'python_logger': record.name
+            },
+            'severity': record.levelname
+        })
 
 
 class BackgroundThreadTransport(Transport):
-    """Aysnchronous transport that uses a background thread.
+    """Asynchronous transport that uses a background thread."""
```
dhermes: I like "Aysnchronous" better. It's an asynchronous spelling of the same word.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
 
-    Writes logging entries as a batch process.
-    """
+    def __init__(self, client, name, grace_period=_DEFAULT_GRACE_PERIOD,
+                 batch_size=_DEFAULT_MAX_BATCH_SIZE):
+        """
+        :type client: ~google.cloud.logging.client.Client
```
dhermes: Ditto, please put in the class docstring.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+        :param name: the name of the logger.
 
-    def __init__(self, client, name):
+        :type grace_period: float
+        :param grace_period: The amount of time to wait for pending logs to
+            be submitted when the process is shutting down.
+
+        :type batch_size: int
+        :param batch_size: The maximum number of items to send at a time in the
+            background thread.
```
dhermes: Not that you need to do it, but we've established one of the two approaches for indents:

```
        :param batch_size: The maximum number of items to send at a time in the
                           background thread.
```

or

```
        :param batch_size:
            The maximum number of items to send at a time in the
            background thread.
```

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -29,36 +29,41 @@ def _get_target_class():
...
     def test_ctor(self):
         client = _Client(self.PROJECT)
-        NAME = 'python_logger'
-        transport = self._make_one(client, NAME)
-        self.assertEqual(transport.worker.logger.name, NAME)
+        name = 'python_logger'
+
+        transport, worker = self._make_one(client, name)
+
+        logger = worker.call_args[0][0]
```
dhermes: Can you be more explicit? Like check `len(worker.call_args)` and `len(worker.call_args[0])`? (i.e. on reading this I have no idea what you **aren't** checking)

ALSO: Feel free to rename this `test_constructor`

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -29,36 +29,41 @@ def _get_target_class():
...
-            'python_logger': python_logger_name
-        }
-        EXPECTED_SENT = (EXPECTED_STRUCT, 'INFO')
-        self.assertEqual(transport.worker.batch.log_struct_called_with,
-                         EXPECTED_SENT)
+        transport.worker.enqueue.assert_called_once_with(record, message)
 
 
 class TestWorker(unittest.TestCase):
+    NAME = 'python_logger'
```
dhermes: Rename `Test_Worker` pretty please?

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -69,103 +74,217 @@ def _get_target_class():
     def _make_one(self, *args, **kw):
         return self._get_target_class()(*args, **kw)
 
+    def _make_one_with_mock_thread(self, *args, **kw):
+        with mock.patch('threading.Thread', new=_Thread):
+            with mock.patch('atexit.register') as atexit_mock:
+                self.atexit_mock = atexit_mock
```
dhermes: :-1: on the `self.atexit_mock` side effect. Why is this necessary?

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -69,103 +74,217 @@ def _get_target_class():
...
+        # Start is implicitly called by worker's ctor
 
-    def test_run_after_stopped(self):
-        # No-op
-        name = 'python_logger'
-        logger = _Logger(name)
-        worker = self._make_one(logger)
+        self.assertTrue(worker.is_alive)
+        self.assertIsNotNone(worker._thread)
+        self.assertTrue(worker._thread._daemon)
```
dhermes: I think the public `daemon` property works fine too?

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -69,103 +74,217 @@ def _get_target_class():
...
+    def test__main_thread_terminated_did_not_join(self):
+        worker = self._make_one_with_mock_thread(_Logger(self.NAME))
+
+        worker._thread._terminate_on_join = False
+        worker.enqueue(mock.Mock(), '')
+        worker._main_thread_terminated()
+
+        self.assertFalse(worker.is_alive)
+
+    def _enqueue_record(self, worker, message):
```
dhermes: `no-self-use` (make it a `@staticmethod`)

logging/google/cloud/logging/handlers/handlers.py
```diff
@@ -20,7 +20,11 @@
 
 DEFAULT_LOGGER_NAME = 'python'
 
-EXCLUDED_LOGGER_DEFAULTS = ('google.cloud', 'oauth2client')
+EXCLUDED_LOGGER_DEFAULTS = (
```
jonparrott: Nope, if they are, they aren't using it as part of this library and we shouldn't exclude it.

Actually, I'm really curious about this whitelist - I feel like it should be pretty narrow (`google.cloud.logging.handlers`) instead of being so broad. I'm wondering why @waprin picked `google.cloud`.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+_DEFAULT_GRACE_PERIOD = 5.0  # Seconds
+_DEFAULT_MAX_BATCH_SIZE = 10
+_WORKER_THREAD_NAME = 'google.cloud.logging.Worker'
+_WORKER_TERMINATOR = object()
+_LOGGER = logging.getLogger(__name__)
 
 
-class _Worker(object):
-    """A threaded worker that writes batches of log entries
+def _get_many(q, max_items=None):
```
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+    """Get multiple items from a Queue.
+
+    Gets at least one (blocking) and at most :param:`max_items` items
```
jonparrott: Actually it can't, I lied. Fixed.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+    (non-blocking) from a given Queue. Does not mark the items as done.
 
-    Writes entries to the logger API.
+    :type q: ~queue.Queue
```
jonparrott: Fixed. God I hate these docstring styles.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+    :param max_items: The maximum number of items to get. If ``None``, then all
+        available items in the queue are returned.
 
-    Currently, the only public methods are constructing it (which also starts
-    it) and enqueuing :class:`Logger` (record, message) pairs.
+    :rtype: Sequence
+    :returns: A sequence of items retrieved from the queue.
     """
+    # Always return at least one item.
+    items = [q.get()]
```
jonparrott: Yep, as documented.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+    while max_items is None or len(items) < max_items:
```
jonparrott: I'd really rather not. This better communicate intent - unbounded vs bounded but at infinity.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -19,137 +19,224 @@
...
+class _Worker(object):
+    """A background thread that writes batches of log entries."""
 
-        # _stop_condition is used to signal from the worker thread to the
-        # main thread that it's finished its last entries
-        self._stop_condition = threading.Condition()
+    def __init__(self, cloud_logger, grace_period=_DEFAULT_GRACE_PERIOD,
+                 max_batch_size=_DEFAULT_MAX_BATCH_SIZE):
+        """
+        The background thread is started automatically.
```
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Fixed via explicit call to `start()`.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done

nunchucks-nuns-chuck-norris-nun-12593488599

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I prefer `_thread_main`, not sure what other information you're trying to communicate with `entry`.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done. Good call. Your boy forgot basic control flow for a second.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I know what batch.commit can raise under usual circumstances, but I do prefer a broad except here to prevent transient unexpected errors from completely murdering the reporting thread.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done and done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: The author!

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Good call, done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I meant to print.

I'd love to use format, but I'm using `%` to be consistent with the rest of the code.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Yes. Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Yes, print is intentional. tuple added.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Tuple added.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Nope. It's valid to add work items to the queue without the thread running. The queue won't be processed, but it's definitely possible and won't have negative effects if you start the thread after.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I swear to god I'm this close to writing a script to translate all of these docstrings to napoleon.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: Done-ish.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: Done.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: To prevent unnecessary atexit handlers holding on to references to objects created in tests.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: Done.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: Done. arg, if this were pytest I could just make this a plain-o-function.

waprin:
So the excluded loggers change, particularly oath2->google.auth looks very important (maybe the simple explanation for the deadlock). In fact now I'm highly highly *highly* suspicious that's the root cause of these deadlocks popping up so that should get merged asap.

Just as a tangent, I think the most useful change to make would be to automatically detect recursive logging traces so that type of breakage is impossible. I think Jon gave me some ideas on how that might be possible with logging filters but I haven't dug deep on it yet.

As for switching from the conditions to the queue, I am 99% indifferent but 1% 👎 , I thought about doing it that way but removing all the items from the `Queue` at once seemed like you might as well just use the `Batch` entries list, instead of having to basically copy the whole list over every time you commit which seems awkward to me. But obviously you found the existing code ugly and it's mostly a stylistic judgement call so it's up to you and other maintainers, like I said largely indifferent one way or the other since it's more or less accomplishing the same thing.

jonparrott:
> As for switching from the conditions to the queue, I am 99% indifferent but 1% 👎  ... But obviously you found the existing code ugly and it's mostly a stylistic judgement call so it's up to you and other maintainers, like I said largely indifferent one way or the other since it's more or less accomplishing the same thing.

My reasoning for doing this wasn't that the code was ugly. I switched to the queue pattern because it's widely deployment, widely tested, well-understood, and easy to reason about. It gets us closer to shared-nothing which is ideal when working with concurrency. I did not like the idea of using locks/conditions/mutexes to patch around a non-thread-safe data structure.

> I thought about doing it that way but removing all the items from the Queue at once seemed like you might as well just use the Batch entries list, instead of having to basically copy the whole list over every time you commit which seems awkward to me

We're not draining the entire queue each time, just a subset. And I'm less worried about copying the list because Python essentially just passes around lists of pointers in this case.

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -50,46 +53,54 @@ def _get_many(q, max_items=None):
...
 
 class _Worker(object):
-    """A background thread that writes batches of log entries."""
+    """A background thread that writes batches of log entries.
 
-    def __init__(self, cloud_logger, grace_period=_DEFAULT_GRACE_PERIOD,
-                 max_batch_size=_DEFAULT_MAX_BATCH_SIZE):
-        """
-        The background thread is started automatically.
+    :type cloud_logger: ~google.cloud.logging.logger.Logger
```
dhermes:
```
:class:`~google.cloud.logging.logger.Logger`
```

Sorry I missed it on first pass

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -50,46 +53,54 @@ def _get_many(q, max_items=None):
...
-        :param max_batch_size: The maximum number of items to send at a time
-            in the background thread.
-        """
+    def __init__(self, cloud_logger, grace_period=_DEFAULT_GRACE_PERIOD,
+                 max_batch_size=_DEFAULT_MAX_BATCH_SIZE):
         self._cloud_logger = cloud_logger
         self._grace_period = grace_period
         self._max_batch_size = max_batch_size
-        self._queue = queue.Queue(-1)
+        self._queue = queue.Queue(0)
```
dhermes: Or `Queue()` (that's what I was getting at)

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -99,35 +110,25 @@ def _thread_main(self):
...
-            if quit:
+                    quit_ = True
+                    # Continue processing items, don't break, try to process
+                    # all items we got back before quitting.
+                else:
+                    batch.log_struct(**item)
+
+            self._safely_commit_batch(batch)
+
+            for _ in range(len(items)):
```
dhermes: Or you pro or con or neither to using `six.moves.range`?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -211,31 +216,33 @@ def enqueue(self, record, message):
...
 class BackgroundThreadTransport(Transport):
-    """Asynchronous transport that uses a background thread."""
+    """Asynchronous transport that uses a background thread.
 
-    def __init__(self, client, name, grace_period=_DEFAULT_GRACE_PERIOD,
-                 batch_size=_DEFAULT_MAX_BATCH_SIZE):
-        """
-        :type client: ~google.cloud.logging.client.Client
-        :param client: The Logging client.
+    :type client: ~google.cloud.logging.client.Client
```
dhermes:
```
:class:`~google.cloud.logging.client.Client`
```

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -211,31 +216,33 @@ def enqueue(self, record, message):
...
+    :param batch_size: The maximum number of items to send at a time in the
+        background thread.
+    """
 
-        :type batch_size: int
-        :param batch_size: The maximum number of items to send at a time in the
-            background thread.
+    def __init__(self, client, name, grace_period=_DEFAULT_GRACE_PERIOD,
+                 batch_size=_DEFAULT_MAX_BATCH_SIZE):
+        """
```
dhermes: Kill these two lines of `"""` (not doing anything)

logging/tests/unit/handlers/transports/test_background_thread.py
```diff
@@ -74,50 +74,52 @@ def _get_target_class():
     def _make_one(self, *args, **kw):
         return self._get_target_class()(*args, **kw)
 
-    def _make_one_with_mock_thread(self, *args, **kw):
+    def _start_with_thread_patch(self, worker):
         with mock.patch('threading.Thread', new=_Thread):
             with mock.patch('atexit.register') as atexit_mock:
                 self.atexit_mock = atexit_mock
-                return self._make_one(*args, **kw)
+                return worker.start()
```
dhermes: `worker.start()` has no return value?

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: pro. I'm using it, see `from six.moves import range`.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I'd like to be explicit about the size. :)

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: I'm using it, see `from six.moves import range`.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: done.

logging/google/cloud/logging/handlers/transports/background_thread.py
jonparrott: Done.

logging/tests/unit/handlers/transports/test_background_thread.py
jonparrott: Right. Fixed.

logging/google/cloud/logging/handlers/transports/background_thread.py
dhermes: Haha DERP

dhermes:
LGTM

PR #3412, opened 2017-05-12
Call start when creating a background thread transport

jonparrott:


jonparrott:
(merged based on offline LGTM from @dhermes)

PR #3414, opened 2017-05-12
Test streaming of various payload row sizes.

### LINKED ISSUE #3019: Add system tests which exercise stream chunking / resume

tseaver:
Closes #3019

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -726,6 +726,59 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         )
 
 
+class TestStreamingChunking(unittest.TestCase, _TestData):
+
+    @classmethod
+    def setUpClass(cls):
+        from tests.system.utils.streaming_utils import INSTANCE_NAME
+        from tests.system.utils.streaming_utils import DATABASE_NAME
+
+        instance = Config.CLIENT.instance(INSTANCE_NAME)
+        if not instance.exists():
+            raise unittest.SkipTest(
+                "Run 'tests/system/utils/populate_streaming.py' to enable.")
+
+        database = instance.database(DATABASE_NAME)
+        if not instance.exists():
+            raise unittest.SkipTest(
+                "Run 'tests/system/utils/populate_streaming.py' to enable.")
+
+        cls._db = database
+
+    def _verify_one_column(self, table_desc):
+        sql = 'SELECT chunk_me FROM {}'.format(table_desc.table)
+        rows = list(self._db.execute_sql(sql))
+        self.assertEqual(len(rows), table_desc.row_count)
+        expected = table_desc.value()
+        for row in rows:
+            self.assertEqual(row[0], expected)
+
+    def _verify_two_columns(self, table_desc):
+        sql = 'SELECT chunk_me, chunk_me_2 FROM {}'.format(table_desc.table)
+        rows = list(self._db.execute_sql(sql))
+        self.assertEqual(len(rows), table_desc.row_count)
+        expected = table_desc.value()
+        for row in rows:
+            self.assertEqual(row[0], expected)
+            self.assertEqual(row[1], expected)
+
+    def test_four_kay(self):
+        from tests.system.utils.streaming_utils import FOUR_KAY
+        self._verify_one_column(FOUR_KAY)
+
+    def test_forty_kay(self):
+        from tests.system.utils.streaming_utils import FOUR_KAY
+        self._verify_one_column(FOUR_KAY)
+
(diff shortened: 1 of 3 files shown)


spanner/tests/system/test_system.py
```diff
@@ -726,6 +726,59 @@ def test_execute_sql_w_query_param(self):
...
+
+        instance = Config.CLIENT.instance(INSTANCE_NAME)
+        if not instance.exists():
+            raise unittest.SkipTest(
+                "Run 'tests/system/utils/populate_streaming.py' to enable.")
+
+        database = instance.database(DATABASE_NAME)
+        if not instance.exists():
+            raise unittest.SkipTest(
+                "Run 'tests/system/utils/populate_streaming.py' to enable.")
```
dhermes: We've used `self.skipTest` [elsewhere][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/2d63cda6bcb3a96230d9ec11c59e243b96cc5c20/bigtable/tests/system.py#L133

spanner/tests/system/test_system.py
```diff
@@ -726,6 +726,59 @@ def test_execute_sql_w_query_param(self):
...
+
+        cls._db = database
+
+    def _verify_one_column(self, table_desc):
+        sql = 'SELECT chunk_me FROM {}'.format(table_desc.table)
+        rows = list(self._db.execute_sql(sql))
+        self.assertEqual(len(rows), table_desc.row_count)
+        expected = table_desc.value()
+        for row in rows:
+            self.assertEqual(rows[0][0], expected)
```
dhermes: You are looping over `rows` but you never use `row`.

Also, what else is in `rows` other than `rows[0]`? What else is in `rows[0]` other than `rows[0][0]`?

spanner/tests/system/test_system.py
```diff
@@ -726,6 +726,59 @@ def test_execute_sql_w_query_param(self):
...
+
+    def _verify_two_columns(self, table_desc):
+        sql = 'SELECT chunk_me, chunk_me_2 FROM {}'.format(table_desc.table)
+        rows = list(self._db.execute_sql(sql))
+        self.assertEqual(len(rows), table_desc.row_count)
+        expected = table_desc.value()
+        for row in rows:
+            self.assertEqual(rows[0][0], expected)
+            self.assertEqual(rows[0][1], expected)
```
dhermes: You are checking `rows`, not `row`

spanner/tests/system/test_system.py
```diff
@@ -726,6 +726,59 @@ def test_execute_sql_w_query_param(self):
...
+        from tests.system.utils.streaming_utils import FOUR_KAY
+        self._verify_one_column(FOUR_KAY)
+
+    def test_four_hundred_kay(self):
+        from tests.system.utils.streaming_utils import FOUR_HUNDRED_KAY
+        self._verify_one_column(FOUR_HUNDRED_KAY)
+
+    def test_four_meg(self):
+        from tests.system.utils.streaming_utils import FOUR_MEG
+        self._verify_two_columns(FOUR_MEG)
```
dhermes: Why did you only do two columns for this test case?

spanner/tests/system/test_system.py
tseaver: "Puh-tayto - poh-tah-to", really.  In this case we are in a class method, which would make it `cls.skipTest` (we are skipping the whole class).

spanner/tests/system/test_system.py
tseaver: Good catch:  I actually patched in the leading `[0]` when the test failed, instead of looking up a line.

spanner/tests/system/test_system.py
tseaver: The first three cases test values in a single column of the size mandated by #3019.  For the 4 Mb case, we need two columns (each 2 Mb in size), as the maximum allowed size for a `STRING` column is slightly larger than 2 Mb.

tseaver:
The logging hang again -- I restarted the build.

jonparrott:
Rebasing *should* fix the hanging.

tseaver:
@jonparrott "This time for *sure*, Rocky!"

jonparrott:
Dude I really hope so. Fingers crossed.

tseaver:
@jonparrott Woohoo!

tseaver:

PR #3423, opened 2017-05-16
Use -s (not -e) when calling nox.

dhermes:
(Sorry @jonparrott for holding it wrong.)


jonparrott:
hahaha, you didn't have to change it. I have no plans to remove the `-e` to `-s` alias for all your toxians. ;)

PR #3424, opened 2017-05-16
iam.Policy: return empty set on missing key.

### LINKED ISSUE #3346: IAM policy should return an empty set of the bindings do not exist
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/bd9fdfba07fac63a91847628613928a569250c0f/core/google/cloud/iam.py#L71

Should return an empty set if no bindings are set for that role.

tseaver:
Closes #3346.


dhermes:
Do we really need a `defaultdict` to solve #3346?

dhermes:
Nah if it's OK with @jonparrott it's OK with me

PR #3425, opened 2017-05-16
Add 'Table.row_from_mapping' helper.

### LINKED ISSUE #3396: table.insert_data should allow for dictionaries
[`table.insert_data`][2] currently only accepts tuples/lists in the order of the schema. It'd be a great feature allow lists of dictionaries to be passed.

Would expect cases where if the keys in the dictionary do not match the schema, it would return back an error.

FWIW, here's an example of how another python library handles it for a typical database connection:

[https://github.com/pudo/dataset/blob/master/dataset/persistence/table.py#L62][1]

[1]: https://github.com/pudo/dataset/blob/5fd8b3ef3ebc3ed143be8f435d917746d10b0c51/dataset/persistence/table.py#L62
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/ff4f492f04c9b4f5511f8b0b7d2a32df86328306/bigquery/google/cloud/bigquery/table.py#L688-L694

tseaver:
Closes #3396.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -684,6 +684,35 @@ class Table(object):
         iterator._NEXT_TOKEN = 'pageToken'
         return iterator
 
+    def row_from_mapping(self, mapping):
+        """Convert a mapping to a row tuple using the schema.
+
+        :type mapping: dict
+        :param mapping: Mapping of row data: must contain keys for all
+               required fields in the schema.  Keys which do not correspond
+               to a field in the schema are ignored.
+
+        :rtype: tuple
+        :returns: Tuple whose elements are ordered according to the table's
+                  schema.
+        :raises: ValueError if table's schema is not set
+        """
+        if len(self._schema) == 0:
+            raise ValueError(_TABLE_HAS_NO_SCHEMA)
+
+        row = []
+        for field in self.schema:
+            if field.mode == 'REQUIRED':
+                row.append(mapping[field.name])
+            elif field.mode == 'REPEATED':
+                row.append(mapping.get(field.name, ()))
+            elif field.mode == 'NULLABLE':
+                row.append(mapping.get(field.name))
+            else:
+                raise ValueError(
+                    "Unknown field mode: {}".format(field.mode))
+        return tuple(row)
+
     def insert_data(self,
                     rows,
                     row_ids=None,
--- a/bigquery/tests/unit/test_table.py
+++ b/bigquery/tests/unit/test_table.py
@@ -1305,6 +1305,61 @@ class TestTable(unittest.TestCase, _SchemaBase):
         self.assertEqual(req['method'], 'GET')
         self.assertEqual(req['path'], '/%s' % PATH)
 
+    def test_row_from_mapping_wo_schema(self):
+        from google.cloud.bigquery.table import _TABLE_HAS_NO_SCHEMA
+        MAPPING = {'full_name': 'Phred Phlyntstone', 'age': 32}
+        client = _Client(project=self.PROJECT)
+        dataset = _Dataset(client)
+        table = self._make_one(self.TABLE_NAME, dataset=dataset)
+
+        with self.assertRaises(ValueError) as exc:
+            table.row_from_mapping(MAPPING)
+
+        self.assertEqual(exc.exception.args, (_TABLE_HAS_NO_SCHEMA,))
+
+    def test_row_from_mapping_w_invalid_schema(self):
+        from google.cloud.bigquery.table import SchemaField
+        MAPPING = {
+            'full_name': 'Phred Phlyntstone',
+            'age': 32,
+            'colors': ['red', 'green'],
+            'bogus': 'WHATEVER',
+        }
+        client = _Client(project=self.PROJECT)
+        dataset = _Dataset(client)
+        full_name = SchemaField('full_name', 'STRING', mode='REQUIRED')
+        age = SchemaField('age', 'INTEGER', mode='REQUIRED')
+        colors = SchemaField('colors', 'DATETIME', mode='REPEATED')
+        bogus = SchemaField('joined', 'STRING', mode='BOGUS')
+        table = self._make_one(self.TABLE_NAME, dataset=dataset,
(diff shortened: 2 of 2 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -684,6 +684,29 @@ def fetch_data(self, max_results=None, page_token=None, client=None):
...
 
+    def row_from_mapping(self, mapping):
+        """Convert a mapping to a row tuple using the schema.
+
+        :type mapping: dict
+        :param mapping: Mapping of row data.
+
+        :rtype: tuple
+        :returns: Tuple matching the table's schema
+        :raises: ValueError if table's schema is not set
```
dhermes: Can a schema be empty?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -684,6 +684,29 @@ def fetch_data(self, max_results=None, page_token=None, client=None):
...
         return iterator
 
+    def row_from_mapping(self, mapping):
+        """Convert a mapping to a row tuple using the schema.
+
+        :type mapping: dict
+        :param mapping: Mapping of row data.
+
+        :rtype: tuple
+        :returns: Tuple matching the table's schema
```
dhermes: Tuple matching the **order** of the table's schema?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -684,6 +684,29 @@ def fetch_data(self, max_results=None, page_token=None, client=None):
...
+        if len(self._schema) == 0:
+            raise ValueError(_TABLE_HAS_NO_SCHEMA)
+
+        row = []
+        for field in self.schema:
+            if field.mode == 'REQUIRED':
+                row.append(mapping[field.name])
+            elif field.mode == 'REPEATED':
+                row.append(mapping.get(field.name, ()))
+            else:  # NULLABLE
```
dhermes: Could make an explicit `field.mode == 'NULLABLE'` and then have a 4th branch that just barfs?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -684,6 +684,29 @@ def fetch_data(self, max_results=None, page_token=None, client=None):
...
+                row.append(mapping.get(field.name))
+        return tuple(row)
```
dhermes: Worth checking that we've hit all the keys in `mapping`? Could raise or warn with a "did you mean..."? Clippy!

bigquery/tests/unit/test_table.py
```diff
@@ -1305,6 +1305,39 @@ def test_fetch_data_w_record_schema(self):
...
+        from google.cloud.bigquery.table import _TABLE_HAS_NO_SCHEMA
+        MAPPING = {'full_name': 'Phred Phlyntstone', 'age': 32}
+        client = _Client(project=self.PROJECT)
+        dataset = _Dataset(client)
+        table = self._make_one(self.TABLE_NAME, dataset=dataset)
+
+        with self.assertRaises(ValueError) as exc:
+            table.row_from_mapping(MAPPING)
+
+        self.assertEqual(exc.exception.args, (_TABLE_HAS_NO_SCHEMA,))
```
dhermes: There's also a `assertRaisesRegexp` if you're so inclined, though I don't like it.

tseaver:
Given a `row_from_mapping` helper method, the `insert_data_from_mappings` one is redundant:  it is the equivalent of:

```python
table.insert_data([table.row_from_mapping(mapping) for mapping in mapping_list])
```

I'm proposing closing #3396 with a note to that effect.

bigquery/google/cloud/bigquery/table.py
tseaver: No:  you cannot perform `Table.create` without supplying a non-empty schema.  If it is empty, it means the user needs to call `Table.reload()` to fetch it, or else configure it manually.

bigquery/google/cloud/bigquery/table.py
tseaver: It would be YAGNI / future-proofing (the only three allowed values for `field.mode` are `REQUIRED`, `REPEATED`, and `NULLABLE`).

bigquery/google/cloud/bigquery/table.py
tseaver: No way do to that without either a) mutating the mapping, b) copying it and mutating the copy, or c) explicitly constructing a set of the keys and mutating that.  I don't see the win, given that it might even be a valid usecase to load a table with row data which held a superset of its columns.

bigquery/tests/unit/test_table.py
tseaver: I don't like it either.  In this case, we know that the error message will be a static string.

bigquery/google/cloud/bigquery/table.py
tseaver: Meh, ok.

bigquery/google/cloud/bigquery/table.py
dhermes: On multiple occasions the BigQuery API has added new acceptable constants / states, so isn't future-proofing a good idea?

bigquery/google/cloud/bigquery/table.py
dhermes: Yeah I suppose that's true. We should definitely document this behavior in the docstring.

bigquery/google/cloud/bigquery/table.py
tseaver: 5ebe002

bigquery/google/cloud/bigquery/table.py
tseaver: 5ebe002

tseaver: b2d11095d34b194b16e1b1ec7329b34dfbdc1f5c removes the stray comment.

lukesneeringer:
@dhermes @tseaver This still seems to be in good shape. Can we merge it?

dhermes:
If LGTY @lukesneeringer then I'm OK with it

PR #3426, opened 2017-05-16
Marshal row data correctly in 'Table.insert_data()'

### LINKED ISSUE #2957: 'Table.insert_data' does not support Datetime values
Seems like the python api library does not support native python datetime objects for inserting BigQuery DateTime-values.

```
INFO:root:Sending load request
[(1L, 'as', datetime.datetime(2014, 2, 1, 0, 0), 'asdasd'),
 (2L, 'ds', datetime.datetime(2014, 2, 2, 0, 0), 'asdsad')]
...
  File "/home/chris/.local/lib/python2.7/site-packages/google/cloud/bigquery/table.py", line 770, in insert_data
    data=data)
  File "/home/chris/.local/lib/python2.7/site-packages/google/cloud/_http.py", line 326, in api_request
    data = json.dumps(data)
  File "/usr/lib/python2.7/json/__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: datetime.datetime(2014, 2, 1, 0, 0) is not JSON serializable
```

tseaver:
Closes #2957.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -126,6 +126,8 @@ _CELLDATA_FROM_JSON = {
 def _row_from_json(row, schema):
     """Convert JSON row data to row with appropriate types.
 
+    Note:  ``row['f']`` and ``schema`` are presumed to be of the same length.
+
     :type row: dict
     :param row: A JSON response row to be converted.
 
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -21,6 +21,7 @@ import datetime
 from google.cloud._helpers import UTC
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
+from google.cloud._helpers import _microseconds_from_datetime
 from google.cloud._helpers import _RFC3339_NO_FRACTION
 from google.cloud._helpers import _time_from_iso8601_time_naive
 from google.cloud._helpers import _to_bytes
@@ -178,14 +179,27 @@ def _bytes_to_json(value):
     return value
 
 
-def _timestamp_to_json(value):
-    """Coerce 'value' to an JSON-compatible representation."""
+def _timestamp_to_json_parameter(value):
+    """Coerce 'value' to an JSON-compatible representation.
+
+    This version returns the string representation used in query parameters.
+    """
     if isinstance(value, datetime.datetime):
         if value.tzinfo not in (None, UTC):
             # Convert to UTC and remove the time zone info.
             value = value.replace(tzinfo=None) - value.utcoffset()
         value = '%s %s+00:00' % (
             value.date().isoformat(), value.time().isoformat())
+    return value
+
+
+def _timestamp_to_json_row(value):
+    """Coerce 'value' to an JSON-compatible representation.
+
+    This version returns floating-point seconds value used in row data.
+    """
+    if isinstance(value, datetime.datetime):
+        value = _microseconds_from_datetime(value) * 1e-6
     return value
 
 
@@ -210,7 +224,8 @@ def _time_to_json(value):
     return value
 
 
-_SCALAR_VALUE_TO_JSON = {
+# Converters used for scalar values marshalled as row data.
+_SCALAR_VALUE_TO_JSON_ROW = {
     'INTEGER': _int_to_json,
     'INT64': _int_to_json,
     'FLOAT': _float_to_json,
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -122,6 +123,36 @@ def _record_from_json(value, field):
...
+
+    :type schema: tuple
+    :param schema: A tuple of
+                   :class:`~google.cloud.bigquery.schema.SchemaField`.
+
+    :rtype: tuple
+    :returns: A tuple of data converted to native types.
+    """
+    row_data = []
+    for field, cell in zip(schema, row['f']):
```
dhermes: This doesn't check that `len(schema) == len(row['f'])` (I realize this has just been moved as-is from below)

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -180,49 +224,25 @@ def _time_to_json(value):
...
-            row_data.append(converter(cell['v'], field))
-
-    return tuple(row_data)
-
-
-def _rows_from_json(rows, schema):
-    """Convert JSON row data to rows with appropriate types."""
-    return [_row_from_json(row, schema) for row in rows]
+# Converters used for scalar values marshalled as query parameters.
+_SCALAR_VALUE_TO_JSON_PARM = _SCALAR_VALUE_TO_JSON_ROW.copy()
```
dhermes: s/`PARM`/`PARAM`

bigquery/google/cloud/bigquery/table.py
```diff
@@ -741,11 +741,9 @@ def insert_data(self,
             row_info = {}
 
             for field, value in zip(self._schema, row):
-                if field.field_type == 'TIMESTAMP':
-                    # BigQuery stores TIMESTAMP data internally as a
-                    # UNIX timestamp with microsecond precision.
-                    # Specifies the number of seconds since the epoch.
-                    value = _convert_timestamp(value)
+                converter = _SCALAR_VALUE_TO_JSON_ROW.get(field.field_type)
+                if converter is not None:
```
dhermes: When will `converter` be `None`?

tseaver: For `STRING` values.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: That would be guarding against either a) a back-end bug (server returns rows which don't match the table's schema), or b) the user manually misconfiguring the schema on the client side, or failing to set it / load it all all (in which case they would get back "short" or empty record tuples).  I can see guarding against the "schema never set" case, but in the caller to this helper (`Table.fetch_data`) rather than here.

bigquery/google/cloud/bigquery/table.py
dhermes: Why not just add a do-nothing converter for `STRING` values?

bigquery/google/cloud/bigquery/_helpers.py
dhermes: OK. Can you add a note about the assumption being made?

tseaver: Note:  `Table.insert_data` already raises a ValueError if the schema is empty.

bigquery/google/cloud/bigquery/table.py
tseaver: Is your concern about performance?  If so, the branch test is going to be orders of magnitude cheaper than a pointless function call.

dhermes: A little bit about performance, but more just code readability. Removing the branch makes it simpler (i.e. no one will worry "When will `converter` be `None`?").

tseaver: I've added a comment.  As I said earlier, the `if` is waaaaaay cheaper than a no-op Python function call.

dhermes: Sorry I didn't acknowledge that before. I understand it'll be faster, e.g.

```
In [1]: def f(x):
   ...:     return x
   ...: 

In [2]: %timeit f(10)
The slowest run took 29.10 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 65.5 ns per loop

In [3]: def g1(x):
   ...:     return f(x)
   ...: 

In [4]: def g2(x):
   ...:     if x:
   ...:         return x
   ...: 

In [5]: %timeit g1(10)
The slowest run took 114.21 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 123 ns per loop

In [6]: %timeit g2(10)
The slowest run took 26.05 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 73.2 ns per loop

In [7]: %timeit g1(10)
The slowest run took 40.91 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 122 ns per loop

In [8]: %timeit g2(10)
The slowest run took 25.87 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 73.7 ns per loop
```

However, I don't think the 60ns speedup is worth the "huh?" reaction that the code invokes on reading.

@jonparrott @lukesneeringer WDYT?

lukesneeringer: I am indifferent.

jonparrott: I have a mild preference to not have exceptional cases in calling code but instead to enforce the assumption that every type has a converter, even if it does nothing.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: 455dc27 adds the comment.

PR #3436, opened 2017-05-17
Adding optional switch to capture project ID in from_service_account_json().

### LINKED ISSUE #1883: set correct project when using from_service_account_json
When instantiating with `bigquery.Client.from_service_account_json('...')` the `project` field in the client should be set to the `project_id` defined in the service account JSON.

Right now it is still the one from env.

dhermes:
Fixes #1883.

@lukesneeringer This is a long-standing bug but should really make it into a `google-cloud-core` release. I know you hate releasing `core`.

### DIFF
--- a/core/tests/unit/test_client.py
+++ b/core/tests/unit/test_client.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
+import json
 import unittest
 
 import mock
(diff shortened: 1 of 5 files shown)


jonparrott:
It would be great to solve this by just consolidating ClientWithProject and ClientWithCredentials.

dhermes:
@jonparrott See the [table][1] I made about usage. The only issue with that is some APIs don't use a project (and we don't want to fail their constructors over a missing project).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1883#issuecomment-282077499

jonparrott:
Does it help that google auth doesn't error if it can't determine a project?

dhermes:
@jonparrott Mostly it doesn't help because it [isn't true][1]? :trollface: 

Or do you mean something else?

[1]: https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/9bea1b15105d6851b315bd622702e0b201e4e874/google/auth/_default.py#L282

jonparrott:
That's if it doesn't find credentials, not if it doesn't find the project.

dhermes:
But there is no `google.auth.default_project`?

jonparrott:
Right. This is part of the constraint. You should never get a project ID
that didn't come from the same source as the credentials.

dhermes:
Makes sense.

@lukesneeringer WDYT of us just using `project` on every `Client` instance? (This would just bring `Client` and `ClientWithProject` together, `ClientWithProject` is already a tiny impl.)

tseaver:
FTR, `ClientWithProject` is skipped by the following:

```bash
$ for f in $(ls */google/cloud/*/client.py); do
       if ! grep -q ClientWithProject $f; then
           echo $f
       fi
  done
bigtable/google/cloud/bigtable/client.py
language/google/cloud/language/client.py
resource_manager/google/cloud/resource_manager/client.py
spanner/google/cloud/spanner/client.py
speech/google/cloud/speech/client.py
translate/google/cloud/translate/client.py
```

Of those, the following have `_ClientProjectMixin` included:

```
$ for f in $(ls */google/cloud/*/client.py); do
       if ! grep -q ClientWithProject $f; then
           echo $f
       fi
  done | xargs grep -l _ClientProjectMixin
bigtable/google/cloud/bigtable/client.py
spanner/google/cloud/spanner/client.py
```

dhermes:
@tseaver This is (intended to be) captured in my [table][1]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/1883#issuecomment-282077499

core/tests/unit/test_client.py
```diff
@@ -221,3 +234,47 @@ def test_ctor_explicit_bytes(self):
...
+                    mock.sentinel.filename, **kwargs)
+
+        self.assertIs(
+            client_obj._credentials, constructor.return_value)
+        self.assertIsNone(client_obj._http_internal)
+        self.assertEqual(client_obj.project, expected_project)
+        # Check that mocks were called as expected.
+        file_open.assert_called_once_with(
+            mock.sentinel.filename, 'r', encoding='utf-8')
+        constructor.assert_called_once_with(info)
```
tseaver: Feels awfully redundant with the version in `TestClient` above.  Not a blocker to merge, though.

dhermes: I can factor out the common parts? The only difference here is the `project` / `project_id`

dhermes:
@tseaver Sorry I missed your LGTM 18 days ago.

It's good I got this in, in case we have a `google-cloud-core` release for your gRPC remap PR, this can go in too.

PR #3443, opened 2017-05-19
Remap new Gax conflict error code

tseaver:
Closes #3175.

Note:  this is a pubsub-only fix for the `ALREADY_EXISTS` error code change.

Note:  also updates `core._testing` to ease generation of errors with that code.

Issue #3175, opened 2017-03-21
PubSub topic.create method should raise a Conflict error instead of RetryError/GaxError when topic exists


lucmult:
Following the [documentation from PubSub site](https://cloud.google.com/pubsub/docs/reference/libraries#client-libraries-install-python) if you run the example 2x it will generate the error:
```
GaxError(Exception occurred in retry method that was not classified as transient, caused by <_Rendezvous of RPC that terminated with (StatusCode.ALREADY_EXISTS, Resource already exists in the project (resource=my-new-topic).)>) <class 'google.gax.errors.RetryError'>
```

However it should raise Conflict Error (google.cloud.exceptions.Conflict), because the API returns a 409 response.
You can see the status code using API Explorer:
https://developers.google.com/apis-explorer/#search/pubsub/pubsub/v1/pubsub.projects.topics.create
```
409  
{
 "error": {
  "code": 409,
  "message": "Resource already exists in the project (resource=my-new-topic).",
  "status": "ALREADY_EXISTS",
...
  }
}
```
Full traceback on Python:
```
>>> topic.create()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/google/home/lucmult/.virtualenvs/pubsub/local/lib/python2.7/site-packages/google/cloud/pubsub/topic.py", line 155, in create
    api.topic_create(topic_path=self.full_name)
  File "/usr/local/google/home/lucmult/.virtualenvs/pubsub/local/lib/python2.7/site-packages/google/cloud/pubsub/_gax.py", line 104, in topic_create
    topic_pb = self._gax_api.create_topic(topic_path)
  File "/usr/local/google/home/lucmult/.virtualenvs/pubsub/local/lib/python2.7/site-packages/google/cloud/gapic/pubsub/v1/publisher_client.py", line 278, in create_topic
    return self._create_topic(request, options)
  File "/usr/local/google/home/lucmult/.virtualenvs/pubsub/local/lib/python2.7/site-packages/google/gax/api_callable.py", line 419, in inner
    return api_caller(api_call, this_settings, request)
  File "/usr/local/google/home/lucmult/.virtualenvs/pubsub/local/lib/python2.7/site-packages/google/gax/api_callable.py", line 407, in base_caller
    return api_call(*args)
[...]

PR #3444, opened 2017-05-19
Move '_catch_remap_gax_error' to 'core.exceptions'.

tseaver:
Toward #3175.

This is the first step toward applying a more uniform mapping of GAX error codes to our exception hierarchy.  For a narrower, pubsub-only fix for #3175, see #3443.

### DIFF
--- a/core/google/cloud/exceptions.py
+++ b/core/google/cloud/exceptions.py
@@ -20,18 +20,27 @@ See: https://cloud.google.com/storage/docs/json_api/v1/status-codes
 # Avoid the grpc and google.cloud.grpc collision.
 from __future__ import absolute_import
 
+import contextlib
 import copy
 import json
+import sys
+
 import six
+
+try:
+    from google.gax.errors import GaxError
+    from google.gax.grpc import exc_to_code
+    from grpc import StatusCode
+    from grpc._channel import _Rendezvous
+except ImportError:  # pragma: NO COVER
+    _HAVE_GRPC = False
+    _Rendezvous = None
+else:
+    _HAVE_GRPC = True
 
 from google.cloud._helpers import _to_bytes
 
 _HTTP_CODE_TO_EXCEPTION = {}  # populated at end of module
-
-try:
-    from grpc._channel import _Rendezvous
-except ImportError:  # pragma: NO COVER
-    _Rendezvous = None
 
 
 # pylint: disable=invalid-name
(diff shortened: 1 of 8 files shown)


core/google/cloud/exceptions.py
```diff
@@ -20,18 +20,27 @@
...
 try:
+    from google.gax.errors import GaxError
+    from google.gax.grpc import exc_to_code
+    from grpc import StatusCode
     from grpc._channel import _Rendezvous
 except ImportError:  # pragma: NO COVER
+    _HAVE_GRPC = False
     _Rendezvous = None
+else:
+    _HAVE_GRPC = True
```
dhermes: You don't **need** this `else` branch, you can set `_HAVE_GRPC = True` in the `try` block

core/google/cloud/exceptions.py
```diff
@@ -250,3 +259,44 @@ def _walk_subclasses(klass):
     code = getattr(_eklass, 'code', None)
     if code is not None:
         _HTTP_CODE_TO_EXCEPTION[code] = _eklass
+
+_GRPC_ERROR_MAPPING = {
```
dhermes: 1. You want two empty lines between top-level definitions
1. This will fail with a `NameError` if `StatusCode` fails import

core/google/cloud/exceptions.py
```diff
@@ -250,3 +259,44 @@ def _walk_subclasses(klass):
...
+    StatusCode.OUT_OF_RANGE: BadRequest,
+    StatusCode.UNIMPLEMENTED: MethodNotImplemented,
+    StatusCode.INTERNAL: InternalServerError,
+    StatusCode.UNAVAILABLE: ServiceUnavailable,
+    StatusCode.DATA_LOSS: InternalServerError,
+}
+
+
+@contextlib.contextmanager
+def _catch_remap_gax_error():
```
dhermes: You can / should define this for the "end-of-module constants"

core/setup.py
```diff
@@ -56,6 +56,7 @@
     'protobuf >= 3.0.0',
     'google-auth >= 0.4.0, < 2.0.0dev',
     'google-auth-httplib2',
+    'google-gax>=0.15.7, <0.16dev',
```
dhermes: 1.  Do we want to do this? (probably yes)
1.  If we do this

    1. You should remove the dep from any subpackages that depend on `google-gax` (if they exist)
    1. Why do we `try/except` the imports?

core/tests/unit/test_exceptions.py
```diff
@@ -14,6 +14,8 @@
 
 import unittest
 
+from google.cloud.exceptions import _HAVE_GRPC
```
dhermes: If this is the only use of `_HAVE_GRPC`, define it in the test module rather than the module we export to users

core/google/cloud/exceptions.py
tseaver: This is what `else:` is for:  assigning after the failing import in the `try:` is much uglier.

dhermes: de gustibus?

core/google/cloud/exceptions.py
tseaver: I'm not sure I understand your comment: `_GRPC_ERROR_MAPPING` *is* and "end-of-module" constant, defined only FBO `_catch_remap_gax_error()`.

dhermes: - I meant to say "before", not "for" (sorry, big time DERP there)
- `_GRPC_ERROR_MAPPING` is definitely "end-of-module" but it's used here at call time, not definition time, so that's not an out-of-order issue

tseaver: 457c207

core/setup.py
tseaver: 2eb6fcc and  	69a8fb7

dhermes:
@lukesneeringer Are you good with the dependency changes?

tseaver:
Note that this change will require releasing a new version of `google-cloud-core` before / with updated API libraries which use the new `google.cloud.exceptions._catch_remap_gax_error`.  It would obsolete #3443 (which is fine by me).

dhermes:
:+1: on the release (I think @lukesneeringer is somewhat anti to the idea of `google-cloud-core` churn)

tseaver:
@dhermes was there another new `core` feature / fix you thought would land soon?  Right now, I see three other PRs touching `core` since the `core-0.24.1` release:

- #3436 (forward-compatible feature)
- #3373 (docs only)
- #3424 (forward-compatible default change)

dhermes:
#3436 was what I was talking about. Looks like we've hit 24 hours, so LGTM?

tseaver:
@dhermes this PR wasn't in the penalty box :).  I'm fine to merge today unless @lukesneeringer has issues with us pushing a new `core` release to support the subsequent API-specific remapping work.

lukesneeringer:
Is the core change backwards-compatible (from an API surface standpoint), or is it something where every single subpackage has to be updated?

I do not mind churn on core; back-incompat churn where everything has to move along with it makes me nervous.

tseaver:
@lukesneeringer the API surface changes are backward-compatible (the add a new, non-API helper function), but the dependency changes are maybe problematic (`google-gax` becomes an unconditional requirement for `core`).

lukesneeringer:
That is a smidge concerning. It would mean that APIs based on HTTP only would now require the _installation of_ (but not use of) GRPC.

Would it be possible to only do the mapping if GAX was present, but not require it? (e.g. `try / except ImportError`)

tseaver:
> That is a smidge concerning. It would mean that APIs based on HTTP only would now require the installation of (but not use of) GRPC.

We already have this for "dual mode" APIs:  even on "classic" GAE they install grpc / gax.

> Would it be possible to only do the mapping if GAX was present, but not require it? (e.g. try / except ImportError)

We can indeed revert to doing that:  this PR actually [backs that stuff out](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3444/files#diff-c90402e5267bb8a7c9369da5b70d9dc6L31).

jonparrott:
I vote with Luke - can we do this in a way that doesn't require gRPC in core (for now)? I would really like to have this so I can move #3617 forward.

tseaver:
Redoing this on a new branch (recent `core` changes make me leery of messing up the original one).

tseaver:
Reworked in #3707.

PR #3446, opened 2017-05-21
DLP Client (GAPIC-only)

lukesneeringer:
@ace-n This should be what you need for samples.


dlp/docs/conf.py
```diff
@@ -0,0 +1,311 @@
+# -*- coding: utf-8 -*-
```
jonparrott: We should separately discuss these generated docs configs, unless you want me to get into it in this PR.

dlp/docs/starting.rst
```diff
@@ -0,0 +1,78 @@
...
+
+
+Using the API
+-------------
+
+
+Authentication
+~~~~~~~~~~~~~~
+
+To authenticate all your API calls, first install and setup the `Google Cloud SDK`_.
```
jonparrott: We gotta get rid of this and just link to https://cloud.google.com/auth

dlp/google/cloud/dlp.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+from google.cloud.dlp_v2beta1 import enums
+from google.cloud.dlp_v2beta1 import types
+from google.cloud.dlp_v2beta1 import DlpServiceClient
```
jonparrott: Was the vision one `VisionServiceClient` or just `VisionClient`?

lukesneeringer: The Vision one is spelled `ImageAnnotatorClient`. It seems to use the name in the proto with `Client` appended.

dlp/docs/starting.rst
lukesneeringer: What is the actual link? That one 404s.

lukesneeringer: (I found this out when I went to go make that change. ;-))

PR #3452, opened 2017-05-25
corrected package name

merla18:
reported by a customer. the package name in public python packager is google-cloud-videointelligence

tested

### DIFF
--- a/videointelligence/README.rst
+++ b/videointelligence/README.rst
@@ -16,7 +16,7 @@ Quick Start
 
 .. code-block:: console
 
-    $ pip install --upgrade google-cloud-video-intelligence
+    $ pip install --upgrade google-cloud-videointelligence
 
 Authentication
 --------------


dhermes:
@lukesneeringer I can confirm on [PyPI][1] and in the [`setup.py`][2] config.

It's also worth noting that the README isn't valid RST for PyPI (see the content on PyPI). This is why we use

```
python setup.py check --restructuredtext --strict
```

It is in the [`nox.py`][4] for the sub-package but we don't run it in our [CircleCI config][3].

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9c48ace907bf60f581352ed66e570ba8dbce849e/.circleci/config.yml

[1]: https://pypi.python.org/pypi/google-cloud-videointelligence
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9c48ace907bf60f581352ed66e570ba8dbce849e/videointelligence/setup.py#L31
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9c48ace907bf60f581352ed66e570ba8dbce849e/.circleci/config.yml
[4]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9c48ace907bf60f581352ed66e570ba8dbce849e/videointelligence/nox.py#L35-L41

dhermes:
@lukesneeringer What say you about my comment?

PR #3469, opened 2017-06-02
Reverse direction of dependencies when determining target packages.

### LINKED ISSUE #3466: get_target_packages has dependencies in the wrong direction
We are [adding dependencies][1] during changes:

```python
answer = answer.union(PKG_DEPENDENCIES.get(package, set()))
```

but this makes a [build][3] with `logging` only changes [trigger][2] BigQuery, Pub / Sub and Storage:

```
$ #!/bin/bash -eo pipefail
  python3.6 test_utils/scripts/get_target_packages.py > ~/target_packages
  cat ~/target_packages

bigquery
logging
pubsub
storage
```

- `logging` **depends** on `bigquery`, `pubsub`, `storage`
- `error_reporting` **depends** on `logging`

So the real output should be a reverse of the dependency map:

```python
{'logging', 'error_reporting'}
```
[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/85fd45a9558f9347dd87abb40a4ce1872b730aec/test_utils/scripts/get_target_packages.py#L150
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/85fd45a9558f9347dd87abb40a4ce1872b730aec/test_utils/scripts/get_target_packages.py#L43
[3]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2063

dhermes:
Fixes #3466.

@lukesneeringer PTAL, I **think** this is the right move, but I may have a case of temporary ~~insaninty~~ insanity misunderstanding the existing behavior.

### DIFF
--- a/test_utils/scripts/get_target_packages.py
+++ b/test_utils/scripts/get_target_packages.py
@@ -109,6 +109,37 @@ def get_changed_files():
         return None
 
 
+def reverse_map(dict_of_sets):
+    """Reverse a map of one-to-many.
+
+    So the map::
+
+      {
+        'A': {'B', 'C'},
+        'B': {'C'},
+      }
+
+    becomes
+
+      {
+        'B': {'A'},
+        'C': {'A', 'B'},
+      }
+
+    Args:
+        dict_of_sets (dict[set]): A dictionary of sets, mapping
+            one value to many.
+
+    Returns:
+        dict[set]: The reversed map.
+    """
+    result = {}
+    for key, values in dict_of_sets.items():
+        for value in values:
+            result.setdefault(value, set()).add(key)
+
+    return result
+
 def get_changed_packages(file_list):
     """Return a list of changed packages based on the provided file list.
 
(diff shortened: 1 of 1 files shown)


lukesneeringer:
That *seems* right to me.

dhermes:
@lukesneeringer Does it seem right on the level of "LGTM"?

test_utils/scripts/get_target_packages.py
```diff
@@ -44,6 +44,11 @@
     'speech': {'storage'},
     'vision': {'storage'},
 }
+REVERSE_DEPS = {}
+for pkg, deps in PKG_DEPENDENCIES.items():
+    for dep in deps:
+        REVERSE_DEPS.setdefault(dep, set()).add(pkg)
+del pkg, deps
```
tseaver: Dunno why you bother with the `del`.  Otherwise, LGTM.

dhermes: To unclutter the global namespace (it's just a reflex)

tseaver: To push that further:  ISTM that you could just as well move the whole kit-and-caboodle (`PKG_DEPENDENCIES`, `REVERSE_DEPS`) inside `get_changed_packages()`, which is their only customer.

dhermes: I am OK with that, at least for doing the reversal. @lukesneeringer WDYT?

lukesneeringer: That sounds right to me.

lukesneeringer:
@dhermes Yes, that was intended to be an approval.

PR #3472, opened 2017-06-05
Expose that settable properties are so.

### LINKED ISSUE #2610: Setting lifecycle rules totally undocumented

tseaver:
Closes #2610.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -578,10 +578,13 @@ class Bucket(_PropertyMixin):
 
     @property
     def lifecycle_rules(self):
-        """Lifecycle rules configured for this bucket.
+        """Retrieve or set lifecycle rules configured for this bucket.
 
         See https://cloud.google.com/storage/docs/lifecycle and
              https://cloud.google.com/storage/docs/json_api/v1/buckets
+
+        :setter: Set lifestyle rules for this bucket.
+        :getter: Gets the lifestyle rules for this bucket.
 
         :rtype: list(dict)
         :returns: A sequence of mappings describing each lifecycle rule.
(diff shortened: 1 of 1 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -591,6 +594,14 @@ def lifecycle_rules(self):
 
     @lifecycle_rules.setter
     def lifecycle_rules(self, rules):
+        """Set lifestyle rules configured for this bucket.
+
+        See https://cloud.google.com/storage/docs/lifecycle and
+             https://cloud.google.com/storage/docs/json_api/v1/buckets
+
+        :type entries: list of dictionaries
+        :param entries: A sequence of mappings describing each CORS policy.
```
dhermes: Copy-pasta?

storage/google/cloud/storage/bucket.py
```diff
@@ -578,11 +578,14 @@ def id(self):
...
     @property
     def lifecycle_rules(self):
-        """Lifecycle rules configured for this bucket.
+        """Retrieve or set lifecycle rules configured for this bucket.
 
         See https://cloud.google.com/storage/docs/lifecycle and
              https://cloud.google.com/storage/docs/json_api/v1/buckets
 
+        :setter: Set lifestyle rules for this bucket.
+        :getter: Gets the lifestyle rules for this bucket.
```
dhermes: :+1: I didn't realize this was part of Sphinx. Win.

@tseaver Have you checked that this docstring (vs. the one from the setter) "wins"?

storage/google/cloud/storage/bucket.py
tseaver: 840dfd2

PR #3475, opened 2017-06-05
Skipping system tests when credentials env. var is unset.

dhermes:
This is labeled "don't merge" because it is **wrong** in it's current state.

So @lukesneeringer @tseaver @jonparrott WDYT the "correct" form should be? We don't want it to raise on PRs, we want it to gracefully do nothing. Otherwise, this is a helpful message and status for people running tests locally with a "bad" setup.

### DIFF
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -49,7 +49,7 @@ def system_tests(session, python_version):
 
     # Sanity check: Only run system tests if the environment variable is set.
     if not os.environ.get('GOOGLE_APPLICATION_CREDENTIALS', ''):
-        return
+        session.skip('Credentials must be set via environment variable.')
 
     # Run the system tests against latest Python 2 and Python 3 only.
     session.interpreter = 'python{}'.format(python_version)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: googleapis/nox:0.11.2
+      - image: googleapis/nox:0.17.0
     steps:
       - checkout
       - run:
--- a/appveyor/requirements.txt
+++ b/appveyor/requirements.txt
@@ -3,4 +3,4 @@
 # pip will build them from source using the MSVC compiler matching the
 # target Python version and architecture
 wheel
-nox-automation==0.11.2
+nox-automation>=0.17.0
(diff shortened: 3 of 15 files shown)


bigquery/nox.py
```diff
@@ -49,7 +50,8 @@ def system_tests(session, python_version):
 
     # Sanity check: Only run system tests if the environment variable is set.
     if not os.environ.get('GOOGLE_APPLICATION_CREDENTIALS', ''):
-        return
+        raise nox.command.CommandFailed(
+            reason='Credentials must be set via environment variable.')
```
duggelz: This message would be even more helpful if it was a link to the "System Test" section of https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/CONTRIBUTING.rst which describes that variable.

dhermes: :+1: Good idea @duggelz

lukesneeringer:
Thoughts on this (sorry for the delay):

  * It is totally possible to read one of the CircleCI / AppVeyor pull request variables and using those as the basis to skip.  (I am less convinced it is a good idea.)
  * I am not as convinced as @dhermes that you always want to run system tests in local testing. In particular, I would expect external contributors generally do not want to do so, or want to run them for only the package they are working on.
  * We could add some kind of `skipped` state to nox. It would be far more valuable for us to report "5 sessions passed, 2 skipped" than "7 sessions passed". I think that would be my preference.

jonparrott:
I'm totally happy to accept a PR to nox to add "session.skip(reason)".

lukesneeringer:
@dhermes Does that work for you?

dhermes:
Yes I am OK with that. I don't have bandwidth to add that feature to `nox` though

jonparrott:
Y'all can file a bug and I'll try to get to it while I'm bored at the hotel
this week.

dhermes:
Donezors: https://github.com/jonparrott/nox/issues/34

dhermes:
I just removed the "do not merge" label.

@jonparrott I think this is read to go. PTAL?

@lukesneeringer ISTM that we'll need our CircleCI image to have the latest version of `nox-automation`, what's the best way to go about that?

@duggelz I have intentionally avoided linked to `CONTRIBUTING.rst` for the time being (though I think it's a great idea). This is because the doc is **months** out-of-date.

dhermes:
Update: currently waiting on @lukesneeringer to rebuild our Docker image, so we can ditch the `pip install --upgrade nox-automation` that I added.

(It's not a big hurry to merge this.)

dhermes:
OK @lukesneeringer got us `googleapis/nox:0.17.0` so this thing is ready to go once green.

PR #3480, opened 2017-06-07
Fix deps

lukesneeringer:


lukesneeringer:
It is what opts-in the `extras_require={'grpc': [...]}` dependencies (which are required here).

PR #3483, opened 2017-06-07
Speech Partial Veneer

lukesneeringer:
This is a partial Veneer for Speech, following the same pattern as Vision.

The base branch for this PR does nothing but add the vendored GAPIC; I did it this way to avoid the clutter in review.


speech/google/cloud/speech/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+# limitations under the License.
+
+from __future__ import absolute_import
+
+
+class SpeechHelpers(object):
+    """A set of convenience methods to make the Speech GAPIC easier to use.
+
+    This class should be considered abstract; it is used as a superclass
+    in a multiple-inheritance construction alongside the applicable GAPIC.
```
jonparrott: Can we avoid the term `gapic` in the docstring?

speech/google/cloud/speech/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+            requests (Iterable[~.speech_v1.types.StreamingRecognizeRequest]):
+                The input objects.
+
+        Returns:
+            Iterable[~.speech_v1.types.StreamingRecognizeRequest]): The
+                correctly formatted input for
+                :meth:`~.speech_v1.SpeechClient.streaming_recognize`.
+        """
+        yield self.types.StreamingRecognizeRequest(streaming_config=config)
+        for r in requests:
```
jonparrott: nit: use `request` instead of `r`, we're not silly Go programmers. :P

speech/google/cloud/speech/helpers.py
lukesneeringer: Yeah, I can change that.

lukesneeringer:
I am letting this go in without waiting on the system tests of literally every other package. I have run them locally (unit, system, lint) repeatedly as well as built docs.

PR #3484, opened 2017-06-07
Make 'QueryResponse.fetch_data' return an iterator.

### LINKED ISSUE #2840: BigQuery: fetch_data() should return a core iterator
See: [https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/query.py#L363][1]

```python
def fetch_data(...):
    """...
     :returns: ``(row_data, total_rows, page_token)``
    ...
    """
```

I would expect it to return some kind of [Iterator](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/cloud/iterator.py) to be consistent with the other cloud APIs.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/2d9c109211331c6cfc59a2a9acbda4c4e3609509/bigquery/google/cloud/bigquery/query.py#L363

tseaver:
Add a system test which exercises it.

Update snippets to match the new usage.

Closes #2840.

### DIFF
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -16,6 +16,7 @@
 
 import six
 
+from google.cloud.iterator import HTTPIterator
 from google.cloud.bigquery._helpers import _TypedProperty
 from google.cloud.bigquery._helpers import _rows_from_json
 from google.cloud.bigquery.dataset import Dataset
@@ -23,6 +24,8 @@ from google.cloud.bigquery.job import QueryJob
 from google.cloud.bigquery.table import _parse_schema_resource
 from google.cloud.bigquery._helpers import QueryParametersProperty
 from google.cloud.bigquery._helpers import UDFResourcesProperty
+from google.cloud.bigquery._helpers import _item_to_row
+from google.cloud.bigquery._helpers import _rows_page_start
 
 
 class _SyncQueryConfiguration(object):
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/query.py
```diff
@@ -426,28 +429,44 @@ def fetch_data(self, max_results=None, page_token=None, start_index=None,
...
+    :type iterator: :class:`~google.cloud.iterator.Iterator`
+    :param iterator: The iterator that is currently in use.
+
+    :type page: :class:`~google.cloud.iterator.Page`
+    :param page: The page that was just created.
+
+    :type response: dict
+    :param response: The JSON API response for a page of rows in a table.
+    """
+    iterator.query_result._set_properties(response)
```
tswast: It's nice that we do this here, but I'd like an explicit reload function, too. https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3506

dstandish:
It appears that documentation has not been updated to reflect this change, e.g. [here](https://googlecloudplatform.github.io/google-cloud-python/stable/bigquery-client.html) and [here](https://cloud.google.com/bigquery/querying-data#bigquery-query-python).

tswast:
The code sample on the documentation has been updated in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/async_query.py I've sent a request to republish the cloud.google.com docs page.

Issue #3474, opened 2017-06-05
Add support for "requester pays" operations


tseaver:
The storage docs don't really explain this, but the JSON API methods have begun to accept a `userProject' query parameter (or is it a header?  only its yak-shaver knows for sure). The point of it looks to be to bill any charges associated with the request to a different project than the one which owns the bucket in which it takes place.  I can't tell how a bucket owner is supposed to configure things such that foreign projects can *only* access objects by passing this parameter/header.

See the [g-c-dotnet PR](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1085) implementing preliminary support.

And hre is the [g-c-php PR](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/527).

jdpedrie:
@tseaver, the bucket owner will enable requester pays by setting `billing.requesterPays` to `true` in the bucket settings.

https://github.com/GoogleCloudPlatform/google-cloud-php/pull/527/files#diff-703c9fa16eb822e79062a7936fceffe5R103

tseaver:
@jdpedrie Thanks.  That seems less elegant than I had imaginged (e.g., there would be some ACL / IAM tweak which would assign permissions to a "user-is-paying" role).

The whole feature needs both narrative and API docs.

tseaver:
@jdpedrie Is there anywhere in the API docs which defines the `billing` field for buckets?  Or how / when to populate the `userProject` header / query parameter / body field?

jdpedrie:
Not that I've seen in the public docs. I shared a google doc with more information. (Check your email).

The [REST discovery document](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest) will show which resources accept the `userProject` parameter (it's in the query string). Our implementation simply adds this field in to each request, if the bucket is constructed with `requesterPays` toggled on.

tseaver:
@jdpedrie Thanks muchly!

tseaver:
Implementation plan:

- [ ] Add a `Bucket.requester_pays` boolean property, mapped onto the `billing/requesterPays` subfield defined in the [REST discovery document](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest).
- [ ] Add an optional `user_project` constructor parameter to `Bucket.__init__`:  we cannot know whether it is valid / appropriate to be set during the constructor, so we just trust the user.
[...]

PR #3488, opened 2017-06-08
Add 'Bucket.requester_pays' property.

tseaver:
Toward #3474.

Note that the new system test is skipped, because `Buckets.insert` fails with the `billing/requesterPays` field set, both in our system tests and in the 'Try It!' form in the docs.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -806,8 +806,9 @@ class Bucket(_PropertyMixin):
     def requester_pays(self):
         """Does the requester pay for API requests for this bucket?
 
-        See  https://cloud.google.com/storage/docs/<DOCS-MISSING> for
-        details.
+        .. note::
+
+           No public docs exist yet for the "requester pays" feature.
 
         :setter: Update whether requester pays for this bucket.
         :getter: Query whether requester pays for this bucket.
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -798,9 +798,38 @@ class Bucket(_PropertyMixin):
         details.
 
         :type value: convertible to boolean
-        :param value: should versioning be anabled for the bucket?
+        :param value: should versioning be enabled for the bucket?
         """
         self._patch_property('versioning', {'enabled': bool(value)})
+
+    @property
+    def requester_pays(self):
+        """Does the requester pay for API requests for this bucket?
+
+        See  https://cloud.google.com/storage/docs/<DOCS-MISSING> for
+        details.
+
+        :setter: Update whether requester pays for this bucket.
+        :getter: Query whether requester pays for this bucket.
+
+        :rtype: bool
+        :returns: True if requester pays for API requests for the bucket,
+                  else False.
+        """
+        versioning = self._properties.get('billing', {})
+        return versioning.get('requesterPays', False)
+
+    @requester_pays.setter
+    def requester_pays(self, value):
+        """Update whether requester pays for API requests for this bucket.
+
+        See  https://cloud.google.com/storage/docs/<DOCS-MISSING> for
+        details.
+
+        :type value: convertible to boolean
+        :param value: should requester pay for API requests for the bucket?
+        """
+        self._patch_property('billing', {'requesterPays': bool(value)})
 
     def configure_website(self, main_page_suffix=None, not_found_page=None):
         """Configure website-related properties.
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -194,7 +194,7 @@ class Client(ClientWithProject):
         except NotFound:
             return None
 
-    def create_bucket(self, bucket_name):
+    def create_bucket(self, bucket_name, requester_pays=None):
         """Create a new bucket.
 
         For example:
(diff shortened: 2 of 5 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -798,10 +798,39 @@ def versioning_enabled(self, value):
...
-        :param value: should versioning be anabled for the bucket?
+        :param value: should versioning be enabled for the bucket?
         """
         self._patch_property('versioning', {'enabled': bool(value)})
 
+    @property
+    def requester_pays(self):
+        """Does the requester pay for API requests for this bucket?
+
+        See  https://cloud.google.com/storage/docs/<DOCS-MISSING> for
```
dhermes: DOCS-MISSING probably not good to check into the source tree

storage/google/cloud/storage/client.py
```diff
@@ -211,10 +211,16 @@ def create_bucket(self, bucket_name):
         :type bucket_name: str
         :param bucket_name: The bucket name to create.
 
+        :type requester_pays: bool
+        :param requester_pays: (Optional) Whether requester pays for
+            API requests for this bucket and its blobs.
```
dhermes: We've typically tried to either indent 4 spaces for **all content** or to line up the continuation lines with the previous content.

storage/google/cloud/storage/client.py
```diff
@@ -194,7 +194,7 @@ def lookup_bucket(self, bucket_name):
         except NotFound:
             return None
 
-    def create_bucket(self, bucket_name):
+    def create_bucket(self, bucket_name, requester_pays=None):
```
dhermes: If we put this here, then why not also put it in the `Client.bucket()` factory?

storage/tests/system.py
```diff
@@ -99,6 +99,15 @@ def test_create_bucket(self):
         self.case_buckets_to_delete.append(new_bucket_name)
         self.assertEqual(created.name, new_bucket_name)
 
+    @unittest.skipIf(True, "requesterPays needs whitelisting?")
```
dhermes: Ha, `True` ouch

storage/google/cloud/storage/bucket.py
tseaver: I don't object in principle, but in practice we are working on wrapping an API feature with no currently-public docs (like the `pubsub.snapshot` bits).  How would you suggest we address that problem?

storage/google/cloud/storage/client.py
tseaver: The feature is especially convenient for this API, which actually creates the bucket on the server side.  Unlike the many other properties, `requester_pays` has an impact on *every* subsequent use of the bucket after the initial `create` call.  Apps which construct buckets via `Client.bucket` or `Bucket.__init__` can set the property before calling `bucket.create`.

storage/google/cloud/storage/bucket.py
dhermes: With clear prose explaining that the docs are missing, rather than a link that looks like it might've been copy-pasta

storage/google/cloud/storage/client.py
dhermes: Yes I understood that as the motivation, though it feels ad-hoc somehow to have it in one place but not the other.

FWIW I realize it's almost "impossible" to allow the `Bucket` constructor to support every single settable attribute/property.

storage/google/cloud/storage/bucket.py
tseaver: 8984768

storage/google/cloud/storage/client.py
tseaver: e5274d2

storage/google/cloud/storage/client.py
tseaver: I think it makes sense to expose this particular property in `Client.create_bucket`:  it affects all subsequent API requests for the bucket.

dhermes: @tseaver I don't disagree, the question is:

- Should it also be in `Client.bucket()`?
- Should we just accept `**kwargs` in the `Bucket` constructor and then try to match them with properties

tseaver:
@dhermes Good to merge when CI is green?

dhermes:
Kinda yes, kinda no. I still feel "uneasy" about the signature delta among:

- `Client.bucket`
- `Client.create_bucket`
- `Bucket`

storage/google/cloud/storage/client.py
tseaver: I would say "no" to both:

- `Client.create_bucket` actually gets the thing constructed on the server side, such that the `billing/requesterPays` setting affects all subsequent API requests for the bucket.
- ` Client.bucket()` / `Bucket.__init__` allow the user to configure the bucket arbitrarily *before* the call to `Bucket.create()`.  I don't see a value in providing noisy alternatives to using the properties.

dhermes: > I don't see a value in providing noisy alternatives to using the properties.

I disagree, but not by much. I don't see **much** value, but I don't think it'd be noisy for users, it'd just be noisy for us to have to document.

But, let's end this discussion and just merge as-is?

tseaver: Note that I think the `user_project` flag (added in #3490) is right for setting in `Client.bucket()` and `Bucket.__init__()`, because it is entirely about how the new client-side bucket instance is used (rather than the "real" server-side bucket).

tseaver:
@dhermes I'm actually opposed on principle to the `Client.{get,lookup,create}_bucket` "convenience" methods:  they violate the Zen, "There should be one-- and preferably only one --obvious way to do it.".

dhermes:
> they violate the Zen

I (mostly) agree with you here.

PR #3489, opened 2017-06-08
Support 'user_project' in '_PropertyMixin.{reload,patch}'.

tseaver:
Uses #3488 as a base.


tseaver:
Oops, turns out this one needs to be folded in with #3490.

dhermes:
:+1:

PR #3490, opened 2017-06-08
Add '{Bucket,Blob}.user_project' properties.

tseaver:
~~Uses #3489 as a base.~~

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -99,6 +99,8 @@ class _PropertyMixin(object):
         # Pass only '?projection=noAcl' here because 'acl' and related
         # are handled via custom endpoints.
         query_params = {'projection': 'noAcl'}
+        if self.user_project is not None:
+            query_params['userProject'] = self.user_project
         api_response = client._connection.api_request(
             method='GET', path=self.path, query_params=query_params,
             _target_object=self)
@@ -145,11 +147,14 @@ class _PropertyMixin(object):
         client = self._require_client(client)
         # Pass '?projection=full' here because 'PATCH' documented not
         # to work properly w/ 'noAcl'.
+        query_params = {'projection': 'full'}
+        if self.user_project is not None:
+            query_params['userProject'] = self.user_project
         update_properties = {key: self._properties[key]
                              for key in self._changes}
         api_response = client._connection.api_request(
             method='PATCH', path=self.path, data=update_properties,
-            query_params={'projection': 'full'}, _target_object=self)
+            query_params=query_params, _target_object=self)
         self._set_properties(api_response)
 
 
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -67,6 +67,11 @@ class _PropertyMixin(object):
         """Abstract getter for the object client."""
         raise NotImplementedError
 
+    @property
+    def user_project(self):
+        """Abstract getter for the object user_project."""
+        raise NotImplementedError
+
     def _require_client(self, client):
         """Check client or verify over-ride.
 
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -227,6 +227,8 @@ class Blob(_PropertyMixin):
         """Project ID used for API requests made via this blob.
 
         Derived from bucket's value.
+
+        :rtype: str
         """
         return self.bucket.user_project
 
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -223,6 +223,14 @@ class Blob(_PropertyMixin):
         return self.bucket.client
 
     @property
+    def user_project(self):
+        """Project ID used for API requests made via this blob.
+
+        Derived from bucket's value.
+        """
+        return self.bucket.user_project
+
+    @property
     def public_url(self):
         """The public URL for this blob's object.
 
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -85,6 +85,10 @@ class Bucket(_PropertyMixin):
     :type name: str
     :param name: The name of the bucket. Bucket names must start and end with a
                  number or letter.
+
+    :type user_project: str
+    :param user_project: (Optional) the project ID to be billed for API
+        requests made via this instance.
     """
 
     _MAX_OBJECTS_FOR_ITERATION = 256
@@ -108,12 +112,13 @@ class Bucket(_PropertyMixin):
     https://cloud.google.com/storage/docs/storage-classes
     """
 
-    def __init__(self, client, name=None):
+    def __init__(self, client, name=None, user_project=None):
         name = _validate_name(name)
         super(Bucket, self).__init__(name=name)
         self._client = client
         self._acl = BucketACL(self)
         self._default_object_acl = DefaultObjectACL(self)
+        self._user_project = user_project
 
     def __repr__(self):
         return '<Bucket: %s>' % (self.name,)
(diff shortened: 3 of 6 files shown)


tseaver:
Rebased against master after the merge of #3488.

storage/google/cloud/storage/_helpers.py
```diff
@@ -94,6 +99,8 @@ def reload(self, client=None):
         # Pass only '?projection=noAcl' here because 'acl' and related
         # are handled via custom endpoints.
         query_params = {'projection': 'noAcl'}
+        if self.user_project is not None:
```
dhermes: We're not worried about a `NotImplementedError` here (i.e. all implementers of this base class will implement this property)?

storage/google/cloud/storage/blob.py
```diff
@@ -223,6 +223,14 @@ def client(self):
         return self.bucket.client
 
     @property
+    def user_project(self):
+        """Project ID used for API requests made via this blob.
+
+        Derived from bucket's value.
```
dhermes: 1. What if the `bucket` is `None`?
1. Put an rtype in there?

storage/google/cloud/storage/bucket.py
```diff
@@ -85,6 +85,10 @@ class Bucket(_PropertyMixin):
     :type name: str
     :param name: The name of the bucket. Bucket names must start and end with a
                  number or letter.
+
+    :type user_project: str
+    :param user_project: (Optional) the project ID to be billed for API
+        requests made via this instance.
```
dhermes: Indent fix? (All lines 4 spaces after or align content)

storage/google/cloud/storage/bucket.py
```diff
@@ -123,6 +128,14 @@ def client(self):
         """The client bound to this bucket."""
         return self._client
 
+    @property
+    def user_project(self):
+        """Project ID to be billed for API requests made via this bucket.
+
+        If unset, API requests are billed to the bucket owner.
```
dhermes: rtype?

storage/tests/unit/test__helpers.py
```diff
@@ -36,15 +36,26 @@ class Derived(self._get_target_class()):
...
+            def user_project(self):
+                return user_project
+
         return Derived
 
     def test_path_is_abstract(self):
         mixin = self._make_one()
-        self.assertRaises(NotImplementedError, lambda: mixin.path)
+        with self.assertRaises(NotImplementedError):
+            mixin.path
```
dhermes: I think `pylint` or somebody might complain? Instead `getattr(mixin, 'path')`?

storage/tests/unit/test__helpers.py
```diff
@@ -62,6 +73,25 @@ def test_reload(self):
         # Make sure changes get reset by reload.
         self.assertEqual(derived._changes, set())
 
+    def test_reload_w_user_project(self):
+        user_project = 'user-project-123'
+        connection = _Connection({'foo': 'Foo'})
+        client = _Client(connection)
+        derived = self._derivedClass('/path', user_project)()
+        # Make sure changes is not a set, so we can observe a change.
```
dhermes: How do you mean "so we can observe a change"? A change would result in an `AttributeError`?

storage/tests/unit/test__helpers.py
```diff
@@ -62,6 +73,25 @@ def test_reload(self):
...
+        derived._changes = object()
+        derived.reload(client=client)
+        self.assertEqual(derived._properties, {'foo': 'Foo'})
+        kw = connection._requested
+        self.assertEqual(len(kw), 1)
+        self.assertEqual(kw[0]['method'], 'GET')
+        self.assertEqual(kw[0]['path'], '/path')
+        self.assertEqual(
+            kw[0]['query_params'],
+            {'projection': 'noAcl', 'userProject': user_project})
```
dhermes: Why don't you just compare `kw[0]` to a dict? This just leaves me (as the reader) wondering if there are unconsidered fields in `kw[0]`.

storage/tests/unit/test_bucket.py
```diff
@@ -54,6 +57,21 @@ def test_ctor(self):
...
+    def test_ctor_w_user_project(self):
+        NAME = 'name'
+        USER_PROJECT = 'user-project-123'
+        connection = _Connection()
+        client = _Client(connection)
+        klass = self._get_target_class()
+        bucket = klass(client, name=NAME, user_project=USER_PROJECT)
+        self.assertEqual(bucket.name, NAME)
+        self.assertEqual(bucket._properties, {})
+        self.assertEqual(bucket.user_project, USER_PROJECT)
```
dhermes: Can you also add a stand-alone test for the newly added property

storage/tests/unit/test__helpers.py
```diff
@@ -95,6 +125,30 @@ def test_patch(self):
...
+        self.assertEqual(derived._properties, {'foo': 'Foo'})
+        kw = connection._requested
+        self.assertEqual(len(kw), 1)
+        self.assertEqual(kw[0]['method'], 'PATCH')
+        self.assertEqual(kw[0]['path'], '/path')
+        self.assertEqual(
+            kw[0]['query_params'],
+            {'projection': 'full', 'userProject': user_project})
+        # Since changes does not include `baz`, we don't see it sent.
+        self.assertEqual(kw[0]['data'], {'bar': BAR})
```
dhermes: Ditto here, why not compare `kw[0]` to a dict?

storage/google/cloud/storage/_helpers.py
tseaver: Yes, both `Bucket` and `Blob` implement it.

storage/google/cloud/storage/blob.py
tseaver: `Blob` instances *must* have a valid bucket (we have lots of other properties which rely on that).

storage/tests/unit/test__helpers.py
tseaver: `nox -e lint` doesn't barf there.  It seems clear to me what is happening, and adding the `getattr` doesn't make it any clearer (lint might barf at not using the return value ;)

storage/tests/unit/test__helpers.py
tseaver: The comment is "not a set", i.e., not an instance of `set`.  It gets replaced during `reload` with an empty `set` instance.

storage/tests/unit/test_bucket.py
tseaver: 4d035d76dd1ba23ca78af642ce37c6facf6174d3

tseaver: I added an assertion for the ctor test which doesn't pass `user_property`.  The value is set in the constructor, and is read-only.

storage/google/cloud/storage/blob.py
tseaver: c2ac0b900238f57342d865e679585ed521533b1d

storage/google/cloud/storage/bucket.py
tseaver: c2ac0b900238f57342d865e679585ed521533b1d

storage/google/cloud/storage/bucket.py
tseaver: c2ac0b900238f57342d865e679585ed521533b1d

storage/tests/unit/test__helpers.py
tseaver: fec9e7c74080dc400e76e24e826e59d451bf9fc1

storage/tests/unit/test__helpers.py
dhermes: Our `nox` config isn't currently using `pylint`, but we'll probably add it back soon

storage/tests/unit/test__helpers.py
dhermes: Gotcha. Can you make the comment clearer, like what you just said?

storage/tests/unit/test_bucket.py
dhermes: Can you add a stand-alone test though? So that all of our unit tests are "explicit" rather than "implicit"? (i.e. each function/method has at least one test just for it)

storage/google/cloud/storage/blob.py
dhermes: OK

dhermes:
LGTM after your comment update

PR #3491, opened 2017-06-08
Adding datastore Key.(to|from)_legacy_urlsafe.

### LINKED ISSUE #3293: Provide a method to convert ndb Key's urlsafe format to/from datastore.Key
In GAE standard, ndb has the concept of the urlsafe string version of a Key. 
e.g. `url_string = sandy_key.urlsafe()` produces a result like `agVoZWxsb3IPCxIHQWNjb3VudBiZiwIM`.

However, I cannot find a way to use this format with the google.cloud.datastore library for code running outside GAE standard. Please add a method to the Key class to create urlsafe strings and a method to generate Keys from urlsafe strings. Alternatively, provide some documentation on how to decode/encode them with the existing library.

NB, the [console support page](https://support.google.com/cloud/answer/6365503?hl=en) says "The encoding method is available in any Cloud Datastore client library."

dhermes:
Fixes #3293.

@lukesneeringer, I'll let @jonparrott describe why this is useful to add.

/cc @pcostell


datastore/google/cloud/datastore/key.py
```diff
@@ -281,6 +298,47 @@ def to_protobuf(self):
 
         return key
 
+    def to_legacy_urlsafe(self):
+        """Convert to a base64 encode urlsafe string for App Engine.
+
+        This is intended to work with the "legacy" representation of a datastore
+        "Key" used within Google App Engine (a so-called "Reference"). This is
+        intended as a drop in for the value returned when using
```
jonparrott: I would word this second sentence as "The returned string can be used as the ``urlsafe`` argument to ``ndb.Key(urlsafe=...)``."

datastore/google/cloud/datastore/key.py
```diff
@@ -281,6 +298,47 @@ def to_protobuf(self):
...
+        ``ndb.Key(...).urlsafe()``.
+
+        :rtype: bytes
+        :returns: ASCII bytes contain the key encoded as URL-safe base64.
```
jonparrott: This reads funny to me, why not just `A bytestring containing the key encoded as URL-safe base64`

datastore/google/cloud/datastore/key.py
```diff
@@ -427,3 +485,153 @@ def _validate_project(project, parent):
...
+        This is 3-4x faster than urlsafe_b64decode()
+
+    :type urlsafe: bytes
+    :param urlsafe: The encoded ASCII string to be decoded
+
+    :rtype: bytes
+    :returns: The value that was decoded.
+    """
+    # This is 3-4x faster than urlsafe_b64decode()
+    return base64.b64decode(
```
jonparrott: ew micro-optimizations. Does this actually differ in behavior at all? If not, can we just use the normal `urlsafe_b64decode` and `encode`?

datastore/google/cloud/datastore/_onestore_v3.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto2";
```
pcostell: To be very clear: This is not a public proto and has no guarantees about forward/backward compatibility from Cloud Datastore. Use at your own risk.

datastore/google/cloud/datastore/key.py
```diff
@@ -427,3 +485,153 @@ def _validate_project(project, parent):
...
+def _clean_app(app_str):
+    """Clean a legacy (i.e. from App Engine) app string.
+
+    :type app_str: str
+    :param app_str: The ``app`` value stored in a "Reference" pb.
+
+    :rtype: str
+    :returns: The cleaned value.
+    """
+    if app_str.startswith('s~') or app_str.startswith('e~'):
```
jonparrott: since project IDs can not contain `~`, why not just `split('~', 1)[-1]`?

datastore/google/cloud/datastore/_onestore_v3.proto
dhermes: @pcostell Thanks for the pointer (I wrote this by hand based on inspecting the raw bytes from a reference)

jonparrott: It's mostly based on GAE's actual usage of the onestore proto, the name is a bit misleading. We can rename this as `gae_key.proto`.

Also, It's *very* unlikely that GAE will ever produce a change that breaks all previously encoded keys.

dhermes: @pcostell Do you want me to rename the file?

dhermes: (UPDATE: I have renamed to `_app_engine_key.proto`)

dhermes:
I've got LGTM here from @jonparrott and from @lukesneeringer over Hangouts, so I'm going to merge.

@pcostell I'm happy to address any other concerns in a follow up PR. In particular, maybe you want me to add some kind of warning or note to the docstrings of these methods?

PR #3492, opened 2017-06-09
Pass 'user_project' if set for Bucket API requests

tseaver:
~~Uses #3490 as a base.~~

This PR might best be reviewed commit-by-commit (one API call at a time).

Toward #3474.


tseaver:
@dhermes rebased after merging #3490 to master.

PR #3495, opened 2017-06-12
Pass 'user_project' if set for Blob API requests

tseaver:
~~Uses #3492 as a base.~~

Toward #3474.


tseaver:
Rebased after merging #3492 to master.

PR #3496, opened 2017-06-12
Use alabaster theme

jonparrott:
Use the same theme as google-gax and google-auth.

Preview here: http://temp.theadora.io/google-cloud-python/index.html


docs/conf.py
```diff
@@ -119,15 +112,21 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-
-if not ON_READ_THE_DOCS:
-    html_theme = 'sphinx_rtd_theme'
-    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+html_theme = 'alabaster'
```
tseaver: Do we want to force the theme even on RTD?

jonparrott: Yes.

dhermes:
Aw man :cry:

jonparrott:
@dhermes why so sad?

dhermes:
> why so sad?

I like the blue theme

lukesneeringer:
I prefer it also, for what it is worth. (I see this as @jonparrott's call though.)

jonparrott:
I'm going to discuss a bit with the tech writing team about these docs, so the theme may change again. Nothing is set in stone. In the meantime, Alabaster is a good choice for several reasons - it's the new default, it's high-contrast, and it's minimal.

PR #3499, opened 2017-06-13
Pass 'user_project' if set for BucketACL/BlobACL API requests

tseaver:
Toward #3474.


tseaver:
Somebody still gets charged, even for updating / querying ACLs. :)

PR #3500, opened 2017-06-13
Pass 'user_project' if set for blob downloads w/ 'mediaLink' set

tseaver:
Fixes an oversight in #3495:  server-set `mediaLink` triggers a separate branch when computing download URL.

Toward #3474.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -407,23 +407,19 @@ class Blob(_PropertyMixin):
         :rtype: str
         :returns: The download URL for the current blob.
         """
+        name_value_pairs = []
         if self.media_link is None:
             base_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
-            scheme, netloc, path, query, frag = urlsplit(base_url)
-            query = parse_qsl(query)
-            # Only add `generation' for synthesized URLs: 'mediaLink' from
-            # server presumably already encodes it.
             if self.generation is not None:
-                query.append(('generation', '{:d}'.format(self.generation)))
+                name_value_pairs.append(
+                    ('generation', '{:d}'.format(self.generation)))
         else:
-            scheme, netloc, path, query, frag = urlsplit(self.media_link)
-            query = parse_qsl(query)
+            base_url = self.media_link
 
         if self.user_project is not None:
-            query.append(('userProject', self.user_project))
+            name_value_pairs.append(('userProject', self.user_project))
 
-        query = urlencode(query)
-        return urlunsplit((scheme, netloc, path, query, frag))
+        return _add_query_parameters(base_url, name_value_pairs)
 
     def _do_download(self, transport, file_obj, download_url, headers):
         """Perform a download without any error handling.
@@ -670,12 +666,14 @@ class Blob(_PropertyMixin):
         info = self._get_upload_arguments(content_type)
         headers, object_metadata, content_type = info
 
-        upload_url = _MULTIPART_URL_TEMPLATE.format(
+        base_url = _MULTIPART_URL_TEMPLATE.format(
             bucket_path=self.bucket.path)
+        name_value_pairs = []
 
         if self.user_project is not None:
-            upload_url += '&userProject={}'.format(self.user_project)
+            name_value_pairs.append(('userProject', self.user_project))
 
+        upload_url = _add_query_parameters(base_url, name_value_pairs)
         upload = MultipartUpload(upload_url, headers=headers)
 
         if num_retries is not None:
@@ -746,12 +744,14 @@ class Blob(_PropertyMixin):
         if extra_headers is not None:
             headers.update(extra_headers)
 
-        upload_url = _RESUMABLE_URL_TEMPLATE.format(
+        base_url = _RESUMABLE_URL_TEMPLATE.format(
             bucket_path=self.bucket.path)
+        name_value_pairs = []
 
         if self.user_project is not None:
-            upload_url += '&userProject={}'.format(self.user_project)
+            name_value_pairs.append(('userProject', self.user_project))
 
+        upload_url = _add_query_parameters(base_url, name_value_pairs)
         upload = ResumableUpload(upload_url, chunk_size, headers=headers)
 
         if num_retries is not None:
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -404,14 +408,22 @@ def _get_download_url(self):
         :returns: The download URL for the current blob.
         """
         if self.media_link is None:
-            download_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
+            base_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
+            scheme, netloc, path, query, frag = urlsplit(base_url)
```
dhermes: In this branch there is no good reason to `urlsplit`. Just update the template so we can avoid text parsing?

storage/google/cloud/storage/blob.py
```diff
@@ -404,14 +408,22 @@ def _get_download_url(self):
...
+            # server presumably already encodes it.
             if self.generation is not None:
-                download_url += u'&generation={:d}'.format(self.generation)
-            if self.user_project is not None:
-                download_url += u'&userProject={}'.format(self.user_project)
-            return download_url
+                query.append(('generation', '{:d}'.format(self.generation)))
         else:
-            return self.media_link
+            scheme, netloc, path, query, frag = urlsplit(self.media_link)
```
dhermes: Can you confirm that the `media_link` won't already contain the `userProject`?

storage/google/cloud/storage/blob.py
```diff
@@ -404,14 +408,22 @@ def _get_download_url(self):
...
+            query = parse_qsl(query)
+
+        if self.user_project is not None:
+            query.append(('userProject', self.user_project))
```
dhermes: If this branch is never exercised it's a wasted parse-unparse.

storage/google/cloud/storage/blob.py
```diff
@@ -404,14 +408,22 @@ def _get_download_url(self):
         :returns: The download URL for the current blob.
         """
         if self.media_link is None:
-            download_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
+            base_url = _DOWNLOAD_URL_TEMPLATE.format(path=self.path)
+            scheme, netloc, path, query, frag = urlsplit(base_url)
+            query = parse_qsl(query)
+            # Only add `generation' for synthesized URLs: 'mediaLink' from
+            # server presumably already encodes it.
```
dhermes: I don't understand this comment, this is the branch where `self.media_link is None`

storage/tests/unit/test_blob.py
```diff
@@ -366,7 +366,7 @@ def test__make_transport(self, fake_session_factory):
 
     def test__get_download_url_with_media_link(self):
         blob_name = 'something.txt'
-        bucket = mock.Mock(spec=[])
+        bucket = _Bucket(name='IRRELEVANT')
```
dhermes: Please use a `mock.Mock`. We don't **need** special classes for just a single property. Instead:

```python
bucket = mock.Mock(name='IRRELEVANT', spec=['name'])
```

storage/google/cloud/storage/blob.py
tseaver: Nope:  I'm explaining why I am adding `generation` in the branch which *doesn't* have `mediaUrl` set by the server.

storage/google/cloud/storage/blob.py
tseaver: It cannot:  `userProject` is not a server-settable/storable field:  it is a per-request query parameter.

storage/google/cloud/storage/blob.py
dhermes: Ah. DERP by me

storage/google/cloud/storage/blob.py
tseaver: Whether or not the server supplies us a `mediaLink` property, any given `Blob` instance (client-side) can have `user_project` set on it when instantiated:  if it does, then we have to pass the equivalent `userProject` query string parameter for all API requests (except in `create`, where it is invalid).

I'd rather use `parse_qsl' and `urlencode` to normalize the process of adding the needed query string parameters:  accreting them by hand is just asking for bugs (e.g., if / when the template changes).

storage/tests/unit/test_blob.py
tseaver: I dissagree, pretty strongly:  having `user_project` on the bucket instances (defaulted to None) is much clearer than having to tweak each `Mock` call site to set it (and add it to the spec).

storage/google/cloud/storage/blob.py
dhermes: We can cross that bridge when we get there, but right now it's a weak argument. It's quite simple to just write code that does:

```python
    download_url = _add_query_param(
        self.media_link, {'userProject': self.user_project})
else:
    download_url = self.media_link
```

storage/tests/unit/test_blob.py
lukesneeringer: I would greatly prefer `mock.Mock`.  It confines to the test what the expectation is, using a format that contributors are going to be familiar with. Having to look up `_Bucket` should not be necessary.

lukesneeringer: As a more general rule: If we are mocking something, we should use `mock`.

tseaver: OK.  Please note that the "mock everything explicitly in each test" pattern makes for a lot of churn:  rather than adding a single line to the (already existing) `_Bucket` class, I have to add one or more lines to basically every test in the module which touches an API call (I switched them to re-using the existing `_Bucket` class in #3495).

lukesneeringer: It is possible to set up a mock once and use it in multiple tests if that is really appropriate:

Set it up in `setUp` and assign it as a class instance, either restoring things in `tearDown` or just passing the mock and `with`-ing it in the tests.

tseaver: Please look at the `test_blob` module:  it *already* uses a `_Bucket` class almost everywhere (84 places currently), and replacing that with a `_make_blob` helper returning a `mock.Mock` instance would be pointless churn in the context of this PR.

I'm not convinced the "contributors won't understand", given that the `_Bucket` class is used everywhere in the module:  how hard is it to do the "search for this name in this module" bit in any editor since even the original `vi` ('*' or '#' WFM in `vim`).  Centralizing the construction of a `mock.Mock` just means they have to look it up somewhere, anyway.

tseaver: I've opened #3509 for this discussion.

storage/google/cloud/storage/blob.py
tseaver: bf94df8 factors out a new `_add_query_parameters` helper, and uses it to ensure that both upload and download URLs are extended cleanly.

tseaver:
@lukesneeringer, @dhermes OK to merge?

PR #3504, opened 2017-06-15
Add sentence about row ordering

garye:
I'm not sure if this is the best place for this, but we want to make sure it's documented that rows are returned in row key order.


dhermes:
@garye Let us know if you want this merged, otherwise we'll look forward to any other docs changes you send

garye:
`consume_all` puts everything into a map keyed by row key so we'd lose the order, right? If that's true then go ahead and merge this one...

Thanks!

dhermes:
@garye `consume_all` just calls `consume_next` until the stream is exhausted

PR #3511, opened 2017-06-19
Implemented trace library based on autogen, gax only.

liyanhui1228:
Implemented the veneer layer of the trace library based on the auto-generated code. This will provide the trace and span context managers and simplify the API for users who want to manually instrument their code. Sample usage in the gist: https://gist.github.com/liyanhui1228/20ac515d74e134ea39b0e63a9d0524bf

Wait for Luke to regenerate the auto-gen layer, which will be a minor change.

Unit tests will be added in another PR.


dhermes:
@liyanhui1228 What sort of review are you looking for here?

/cc @jonparrott

duggelz:
Just look at the non-autogenerated code.

It briefly appeared that the internal trace teams might want us to put the stackdriver trace client code in a different repository and/or do everything completely differently.  However, after some debate, the autogenerated trace client library will definitely go here as originally planned (e.g. like every other cloud product).  The layers on top of that which Angela is working on may eventually move to another repository, or not (I vote for not).

liyanhui1228:
@dhermes Thanks! I just want to get some feedback to ensure I'm on the right track. And in the meantime I'll work on the http part and unit tests.

jonparrott:
Hey @liyanhui1228, it would be really awesome if you could break this in to two PRs: one for just the autogenerated stuff (which we can merge immediately) and one for the handwritten parts.

It also might make things easier to target a non-master branch.

liyanhui1228:
@jonparrott OK! I'll close this one and submit two PRs separately.

PR #3514, opened 2017-06-20
added support for creating a view with useLegacySql = False

### LINKED ISSUE #3388: The Python BigQuery client fails on creating a view in standardSQL
```python

bigquery_client = bigquery.Client(project="my-project")
dataset = bigquery_client.dataset("my-dataset")

table = dataset.table("v_test")

#Fails on Standard SQL
#table._properties['view'] = {'query': "select * from `my-project.my.dataset.test_table`",'useLegacySql':False}  

#Works on Legacy SQL
table._properties['view'] = {'query': "select * from [my-project:my-dataset.test_table]",'useLegacySql':True}  

try:
    table.create()
except Exception as err:
    print(err)
```
The https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/table.py#L355 should also take in another property for useLegacySql

Example  payload at http://stackoverflow.com/questions/33243427/how-can-i-create-a-new-view-in-bigquery-using-the-python-api

richkadel:
Plus a couple of small doc fixes.

Passes tests and coverage.

Closes #3388.


bigquery/tests/unit/test_table.py
```diff
@@ -675,6 +694,7 @@ def test_create_w_alternate_client(self):
         RESOURCE['expirationTime'] = _millis(self.EXP_TIME)
         RESOURCE['view'] = {}
         RESOURCE['view']['query'] = QUERY
+#        RESOURCE['view']['useLegacySql'] = True
```
tseaver: Why is this line commented out?  (and likewise, the ones below for the same testcase).  If we really want this testcase *not* to pass the value, then they should just be removed.

richkadel:
My mistake. These commented out lines are now fully removed.

tseaver:
@richkadel Thank you very much for the patch!

PR #3515, opened 2017-06-20
Add back pylint as info-only for core

jonparrott:


core/nox.py
```diff
@@ -43,15 +41,21 @@ def unit_tests(session, python_version):
...
-    session.install('flake8')
+    session.install('flake8', 'pylint', 'gcp-devrel-py-tools')
     session.install('.')
     session.run('flake8', 'google/cloud/core')
+    session.run(
+        'gcp-devrel-py-tools', 'run-pylint',
+        '--library-filesets', 'google',
+        '--test-filesets', 'tests',
+        # Temporarily allow this to fail.
+        success_codes=range(0, 100))
```
dhermes: You probably want `list(range(0, 100))`? If the script is invoked by Python 3, this will be an iterator.

core/nox.py
```diff
@@ -43,15 +41,21 @@ def unit_tests(session, python_version):
...
+    Returns a failure if the linters find linting errors or sufficiently
     serious code quality issues.
     """
     session.interpreter = 'python3.6'
-    session.install('flake8')
+    session.install('flake8', 'pylint', 'gcp-devrel-py-tools')
     session.install('.')
     session.run('flake8', 'google/cloud/core')
+    session.run(
+        'gcp-devrel-py-tools', 'run-pylint',
```
dhermes: Can it somehow be less opaque how `pylintrc` and `pylintrc.test` are generated by `gcp-devrel-py-tools`?

jonparrott: What do you recommend?

core/nox.py
jonparrott: It works fine in python 3.

core/nox.py
dhermes: Well first, could you explain how it works, then we can go from there?

core/nox.py
dhermes: OK cool. TIL: `__contains__` works as one might expect on a `range` object Python 3

core/nox.py
jonparrott: It's basically what your old "run_pylint" script does, however, the key difference here is that we could specify a `pylintrc.py` file here and use it to programmatically override the defaults.

https://github.com/GoogleCloudPlatform/python-repo-tools/blob/master/gcp_devrel/tools/pylint.py

dhermes: Ohhh that's fancy!

jonparrott: So WDYT, is there a comment or anything I can put here to make things easier?

dhermes: I think we should just have a `pylintrc.py` file, even if it is very/totally bare. Then you can just make a comment referring to it?

jonparrott:
@dhermes added a "empty" config file for you. PTAL?

core/pylint.config.py
```diff
@@ -0,0 +1,25 @@
...
+
+# Library configuration
+
+# library_additions = {}
+# library_replacements = {}
+
+# Test configuration
+
+# test_additions = copy.deepcopy(library_additions)
+# test_replacements = copy.deepcopy(library_replacements)
```
dhermes: Why comment these out? Does it break something to have them empty?

jonparrott: Yep.

dhermes: OK.

dhermes:
This LGTM.

jonparrott:
Cool, I'll send another PR to add this to all the other packages.

dhermes:
SGTM. Do we think @lukesneeringer cares? (Prepare for some lint violations since we've had it turned off.)

jonparrott:
It's in info-only mode so it will not cause any new failures. Luke gave
approval in the email chain you were on.

dhermes:
Boom roasted

PR #3518, opened 2017-06-22
Adding @jonparrott's sphinx-docstring-typing module.

dhermes:
This allows usage of Python 3.5+ typing module, which is very helpful.

Note that the changes to **source** docstrings are likely bugs in `sphinx-docstring-typing`, so that may be addressed on that project.

@jonparrott This is marked "don't merge" right now because I want to check with you what the issues actually are with the source docstrings.

### DIFF
--- a/language/google/cloud/language/document.py
+++ b/language/google/cloud/language/document.py
@@ -266,10 +266,10 @@ class Document(object):
             This API is intended for users who are familiar with machine
             learning and need in-depth text features to build upon.
 
-        .. _annotateText: https://cloud.google.com/natural-language/\
+        .. _annotate-text: https://cloud.google.com/natural-language/\
                           reference/rest/v1/documents/annotateText
 
-        See `annotateText`_.
+        See `annotate-text`_.
 
         :type include_syntax: bool
         :param include_syntax: (Optional) Flag to enable syntax analysis
--- a/vision/google/cloud/proto/vision/v1/text_annotation_pb2.py
+++ b/vision/google/cloud/proto/vision/v1/text_annotation_pb2.py
@@ -558,13 +558,14 @@ TextAnnotation = _reflection.GeneratedProtocolMessageType('TextAnnotation', (_me
   DESCRIPTOR = _TEXTANNOTATION,
   __module__ = 'google.cloud.proto.vision.v1.text_annotation_pb2'
   ,
-  __doc__ = """TextAnnotation contains a structured representation of OCR extracted
-  text. The hierarchy of an OCR extracted text structure is like this:
-  TextAnnotation -> Page -> Block -> Paragraph -> Word -> Symbol Each
+  __doc__ = """``TextAnnotation`` contains a structured representation of OCR extracted text.
+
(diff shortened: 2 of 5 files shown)


language/google/cloud/language/document.py
```diff
@@ -266,10 +266,10 @@ def annotate_text(self, include_syntax=True, include_entities=True,
             This API is intended for users who are familiar with machine
             learning and need in-depth text features to build upon.
 
-        .. _annotateText: https://cloud.google.com/natural-language/\
+        .. _annotate-text: https://cloud.google.com/natural-language/\
```
dhermes: @jonparrott This introduced an error where `sphinx-docstring-typing` lowercased the target:

> Unknown target name: "annotatetext".

vision/google/cloud/proto/vision/v1/text_annotation_pb2.py
```diff
@@ -558,13 +558,14 @@
   DESCRIPTOR = _TEXTANNOTATION,
   __module__ = 'google.cloud.proto.vision.v1.text_annotation_pb2'
   ,
-  __doc__ = """TextAnnotation contains a structured representation of OCR extracted
-  text. The hierarchy of an OCR extracted text structure is like this:
-  TextAnnotation -> Page -> Block -> Paragraph -> Word -> Symbol Each
+  __doc__ = """``TextAnnotation`` contains a structured representation of OCR extracted text.
```
dhermes: @jonparrott The issues here were all variations on

> Inline interpreted text or phrase reference start-string without end-string.

I'm pretty sure it's because `sphinx-docstring-typing` picked up class names and tried to make them pretty?

jonparrott: That would mean that it's outputting bad rst (see [here](https://www.mantidproject.org/Documentation_Guide_For_Devs#Inline_interpreted_text_or_phrase_reference_start-string_without_end-string)). 

I wonder if you would be willing to add both of these cases you ran into into the test case for sphinx-docstring-typing so we can see exactly what's going on.

dhermes: I would be, could you point me to the unit test module / file that needs updating?

jonparrott:
https://github.com/jonparrott/sphinx-docstring-typing/blob/master/tests/test_sphinx_docstring_typing.py

dhermes:
Just sent https://github.com/jonparrott/sphinx-docstring-typing/pull/5

lukesneeringer:
Friendly poke. I would like to either move the needle on this or close it.

dhermes:
I would also like to move the needle on it.

In the current form `sphinx-docstring-typing` probably isn't suitable for our use. It uses no semantic meaning of docstring text when doing replacements, so it can make things like `TextFooBar` turn into very broken Sphinx reST because of the presence of `Text`

jonparrott:
`sphinx-docstring-typing` probably needs some form of real parser instead of using clever regexes.

lukesneeringer:
Okay. Dropping this for now.

PR #3519, opened 2017-06-22
Re-enable pylint in info-only mode for all packages

jonparrott:


dhermes:
A cursory scan looks fine, though ISTM we should have a single `pylint.config.py` at the project root and then every package points at that?

jonparrott:
I'd do that but @lukesneeringer is planning on splitting packages into their own repos.

dhermes:
Yeah in that context it totally makes sense.

tseaver:
@jonparrott I'm missing what makes running pylint this way "advisory only":  can you elaborate?

jonparrott:
@tseaver nox is ignoring its return value, so pylint can fail but the lint session will still pass.

jonparrott:
@dhermes @lukesneeringer y'all got any more of them LGTMs?

dhermes:
LGTM

PR #3520, opened 2017-06-23
BigQuery _EnumProperty ValueError messages are not displayed properly

smasue:
I found this very small bug in the error message, so I fixed it!

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -318,7 +318,7 @@ class _EnumProperty(_ConfigurationProperty):
         :raises: ValueError if value is not allowed.
         """
         if value not in self.ALLOWED:
-            raise ValueError('Pass one of: %s' ', '.join(self.ALLOWED))
+            raise ValueError('Pass one of: %s' % ', '.join(self.ALLOWED))
 
 
 class UDFResource(object):


dhermes:
LGTM. Thanks!

Can you let us know once you've signed the CLA (unless the bot makes an automated reply)?

smasue:
I've just signed the CLA

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -318,7 +318,7 @@ def _validate(self, value):
         :raises: ValueError if value is not allowed.
         """
         if value not in self.ALLOWED:
-            raise ValueError('Pass one of: %s' ', '.join(self.ALLOWED))
```
dhermes: No worries here, but a point of note when using percent (`%`) formatting in Python. To be extra super-duper safe, you should always pass a tuple:

```python
>>> format = 'see %s'
>>> value = 'this'
>>> format % value
'see this'
>>> value = ('a', 'real', 'tuple', 9)
>>> format % value
TypeError: not all arguments converted during string formatting
>>> format % (value,)
"see ('a', 'real', 'tuple', 9)"
```

In this case it's obvious we aren't accidentally passing a tuple as the single argument

smasue: Yes you are right! Do you want me to do another PR for it?

dhermes: Naw no worries, thanks for the contribution!

PR #3542, opened 2017-06-26
Prep error-reporting-0.25.0 release.

tseaver:
Requires merge of #3526 to bump core to 0.25.0.

Draft release notes:

## google-cloud-error-reporting-0.25.0

- Update google-cloud-core dependency to ~= 0.25.

----

Excluded from notes:

- Re-enable pylint in info-only mode for all packages (#3519)
- Vision semi-GAPIC (#3373)
- Increasing retries from 4 to 6 for error reporting system test.
- Add error reporting system test (#3348)

### DIFF
--- a/error_reporting/setup.py
+++ b/error_reporting/setup.py
@@ -51,14 +51,14 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.24.1, < 0.25dev',
+    'google-cloud-core >= 0.25.0, < 0.26dev',
     'google-cloud-logging >= 1.0.0, < 2.0dev',
     'gapic-google-cloud-error-reporting-v1beta1 >= 0.15.0, < 0.16dev'
 ]
 
 setup(
     name='google-cloud-error-reporting',
-    version='0.24.2',
+    version='0.25.0',
     description='Python Client for Stackdriver Error Reporting',
     long_description=README,
     namespace_packages=[


error_reporting/setup.py
```diff
@@ -51,14 +51,14 @@
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.24.1, < 0.25dev',
+    'google-cloud-core >= 0.25.0, < 0.26dev',
     'google-cloud-logging >= 1.0.0, < 2.0dev',
```
dhermes: Actually you should update the `logging` dep here

tseaver: That one is broad enough to let us pick up the new logging version (1.2.0)

dhermes: I'm more worried about the lower bound here

tseaver: Surely we didn't introduce breaking changes in `logging` without bumping its major version. :)

dhermes: I'm not disputing that. It just seems like good hygiene? Maybe not?

tseaver: Looking at the changes, I'm pretty sure we are safe in leaving the dep wide:  `error_reporting` has had no changes which depend on anything newer in `logging`.

dhermes: I'm not saying we are unsafe, just saying it's good hygiene to keep it "fresh".

PR #3318, opened 2017-04-20
Adding check that **all** setup.py README's are valid RST.

dhermes:
Follow up to #3316. Fixes #2446.

PR #3545, opened 2017-06-26
Prep runtimeconfig-0.25.0 release.

tseaver:
Requires merge of #3526 to bump core to 0.25.0.

Draft release notes:

## google-cloud-runtimeconfig-0.25.0

- Update google-cloud-core dependency to ~= 0.25.

----

Excluded from notes:

- Re-enable pylint in info-only mode for all packages (#3519)
- Vision semi-GAPIC (#3373)
- Ignore tests (rather than unit_tests) in setup.py files. (#3319)
- Adding check that **all** setup.py README's are valid RST. (#3318)

### DIFF
--- a/runtimeconfig/setup.py
+++ b/runtimeconfig/setup.py
@@ -51,12 +51,12 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.24.0, < 0.25dev',
+    'google-cloud-core >= 0.25.0, < 0.26dev',
 ]
 
 setup(
     name='google-cloud-runtimeconfig',
-    version='0.24.0',
+    version='0.25.0',
     description='Python Client for Google Cloud RuntimeConfig',
     long_description=README,
     namespace_packages=[


dhermes:
@tseaver I'm not sure about the use of `~` (I presume to mean approximately?) How about instead:

> Update google-cloud-core dependency to 0.25.x series?

tseaver:
I meant to be using `~=`, in the [PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release) sense.

dhermes:
I'm not sure many people know `~=` in the PEP 440 sense. I've read PEP 440 and I don't.

tseaver:
AFAIK, it is what we should be using rather than the incantation we are now.

tseaver:
```bash
$ /opt/Python-2.7.13/bin/virtualenv /tmp/pep440
New python executable in /tmp/pep440/bin/python
Installing setuptools, pip, wheel...done.
$ /tmp/pep440/bin/pip install google-cloud-core~=0.24
Collecting google-cloud-core~=0.24
  Using cached google_cloud_core-0.24.1-py2.py3-none-any.whl
...
```

dhermes:
I'm not sure we're using it anywhere? Or do you mean rather than the lower-upper bound combination?

PR #3554, opened 2017-06-27
Add the ability to specify AnnotateImageRequest items in single-feature methods.

lukesneeringer:
This commit makes it such that any other part of an `AnnotateImageRequest` can be specified on the single-feature methods as a keyword argument. Needed for crop hints.


vision/tests/unit/test_decorators.py
```diff
@@ -55,15 +55,21 @@ class SingleFeatureMethodTests(unittest.TestCase):
     def test_runs_generic_single_image(self, ai):
         ai.return_value = vision.types.AnnotateImageResponse()
 
+        # Prove that other aspects of the AnnotateImageRequest, such as the
+        # image context, will be preserved.
+        SENTINEL = object()
```
dhermes: You could instead use `mock.sentinel.image_context`

PR #3559, opened 2017-06-28
stripping of padding characters from datastore's urlsafe key

michaelenglo:
Hi,

I am proposing a change to the Google Cloud Platform Datastore's Python client library, specifically the Key module. @dhermes and I have discussed about removing the padding characters ("=") from regular b64encoded urlsafe to match the version that is used in Datastore (which does not have the padding character "="). This is done to ensure consistency in using and reproducing Datastore's convention for urlsafe key.

Please let me know if there is any other steps I may not have included and are needed to do to improve the changes.

### DIFF
--- a/datastore/google/cloud/datastore/key.py
+++ b/datastore/google/cloud/datastore/key.py
@@ -304,7 +304,9 @@ class Key(object):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). The returned string can be used as the ``urlsafe``
-        argument to ``ndb.Key(urlsafe=...)``.
+        argument to ``ndb.Key(urlsafe=...)``. Notice that the datastore's
+        version of urlsafe encoding does not have padding characters, unlike
+        the out of the box urlsafe_b64encode's generated urlsafe.
 
         :rtype: bytes
         :returns: A bytestring containing the key encoded as URL-safe base64.
@@ -315,7 +317,7 @@ class Key(object):
             name_space=self.namespace,
         )
         raw_bytes = reference.SerializeToString()
-        return base64.urlsafe_b64encode(raw_bytes)
+        return base64.urlsafe_b64encode(raw_bytes).strip('=')
 
     @classmethod
     def from_legacy_urlsafe(cls, urlsafe):
@@ -324,7 +326,10 @@ class Key(object):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). This assumes that ``urlsafe`` was created within an App
-        Engine app via something like ``ndb.Key(...).urlsafe()``.
+        Engine app via something like ``ndb.Key(...).urlsafe()``.Notice that 
+        the datastore's version of urlsafe encoding does not have padding
+        characters, unlike the out of the box urlsafe_b64encode's generated
+        urlsafe.
 
         :type urlsafe: bytes or unicode
         :param urlsafe: The base64 encoded (ASCII) string corresponding to a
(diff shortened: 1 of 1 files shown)


datastore/google/cloud/datastore/key.py
```diff
@@ -315,7 +317,7 @@ def to_legacy_urlsafe(self):
             name_space=self.namespace,
         )
         raw_bytes = reference.SerializeToString()
-        return base64.urlsafe_b64encode(raw_bytes)
+        return base64.urlsafe_b64encode(raw_bytes).strip('=')
```
dhermes: I'm fairly certain you want to strip `b'='` here (for Py2/Py3) compat

datastore/google/cloud/datastore/key.py
```diff
@@ -304,7 +304,9 @@ def to_legacy_urlsafe(self):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). The returned string can be used as the ``urlsafe``
-        argument to ``ndb.Key(urlsafe=...)``.
+        argument to ``ndb.Key(urlsafe=...)``. Notice that the datastore's
+        version of urlsafe encoding does not have padding characters, unlike
+        the out of the box urlsafe_b64encode's generated urlsafe.
```
dhermes: You can just shorten this newly added content to

> The base64 encoded values will have padding removed.

datastore/google/cloud/datastore/key.py
```diff
@@ -324,7 +326,10 @@ def from_legacy_urlsafe(cls, urlsafe):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). This assumes that ``urlsafe`` was created within an App
-        Engine app via something like ``ndb.Key(...).urlsafe()``.
+        Engine app via something like ``ndb.Key(...).urlsafe()``.Notice that
```
dhermes: See my above about padding and also you are missing a space between the previous sentence and the start of your "Notice..."

datastore/google/cloud/datastore/key.py
```diff
@@ -333,6 +338,15 @@ def from_legacy_urlsafe(cls, urlsafe):
         :rtype: :class:`~google.cloud.datastore.key.Key`.
         :returns: The key corresponding to ``urlsafe``.
         """
+        if not isinstance(urlsafe, basestring):
+            raise TypeError('urlsafe must be a string; received %r' % urlsafe)
+        if isinstance(urlsafe, unicode):
+            urlsafe = urlsafe.encode('utf8')
```
dhermes: These type checks are handled by `_to_bytes` below, please remove them.

datastore/google/cloud/datastore/key.py
```diff
@@ -333,6 +338,15 @@ def from_legacy_urlsafe(cls, urlsafe):
...
+
+        mod = len(urlsafe) % 4
+        if mod:
+            urlsafe += '=' * (4 - mod)
```
dhermes: 1.  Please move this to **after** the `_to_bytes` call
1.  It'll be a bit annoying to have the `if` branch in there for unit tests, so instead I recommend

    ```python
    padding = b'=' * (-len(urlsafe) % 4)
    urlsafe += padding
    ```

----

To sanity check that last bit, observe:

```python
>>> def pad(urlsafe):
...     padding = b'=' * (-len(urlsafe) % 4)
...     urlsafe += padding
...     return urlsafe
... 
>>> for len_ in range(10):
...     value = ''.join(map(str, range(1, len_ + 1)))
...     print(pad(value))
... 

1===
12==
123=
1234
12345===
123456==
1234567=
12345678
123456789===
```

dhermes:
Thanks for sending @michaelenglo. Please be sure to sign the CLA.

PR #3560, opened 2017-06-28
strip padding characters from urlsafe

michaelenglo:
Hi,

I am proposing a change to the Google Cloud Platform Datastore's Python client library, specifically the Key module. @dhermes and I have discussed about removing the padding characters ("=") from regular b64encoded urlsafe to match the version that is used in Datastore (which does not have the padding character "="). This is done to ensure consistency in using and reproducing Datastore's convention for urlsafe key.

Please let me know if there is any other steps I may not have included and are needed to do to improve the changes.

### DIFF
--- a/datastore/google/cloud/datastore/key.py
+++ b/datastore/google/cloud/datastore/key.py
@@ -304,9 +304,8 @@ class Key(object):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). The returned string can be used as the ``urlsafe``
-        argument to ``ndb.Key(urlsafe=...)``. Notice that the datastore's
-        version of urlsafe encoding does not have padding characters, unlike
-        the out of the box urlsafe_b64encode's generated urlsafe.
+        argument to ``ndb.Key(urlsafe=...)``. The base64 encoded values
+        will have padding removed.
 
         :rtype: bytes
         :returns: A bytestring containing the key encoded as URL-safe base64.
@@ -317,7 +316,7 @@ class Key(object):
             name_space=self.namespace,
         )
         raw_bytes = reference.SerializeToString()
-        return base64.urlsafe_b64encode(raw_bytes).strip("=")
+        return base64.urlsafe_b64encode(raw_bytes).strip(b'=')
 
     @classmethod
     def from_legacy_urlsafe(cls, urlsafe):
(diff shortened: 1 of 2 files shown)


michaelenglo:
@dhermes Ignore the last PR. I just made a new exact same commit using my personal account (I was using my work account when committing the earlier one, which is not associated to this Github account and not CLA signed).

dhermes:
Thanks for using your personal account. Did you have a chance to see my comments on #3559?

michaelenglo:
Hi @dhermes , thanks for your helpful comments! I made another commit to my master to follow your comments. Let me know if it's already good to go or not!

datastore/google/cloud/datastore/key.py
```diff
@@ -324,7 +325,8 @@ def from_legacy_urlsafe(cls, urlsafe):
         This is intended to work with the "legacy" representation of a
         datastore "Key" used within Google App Engine (a so-called
         "Reference"). This assumes that ``urlsafe`` was created within an App
-        Engine app via something like ``ndb.Key(...).urlsafe()``.
+        Engine app via something like ``ndb.Key(...).urlsafe()``. The base64
+        encoded values will have padding removed.
```
dhermes: Here the comment is inaccurate, and probably not needed? The base64 encode values are the **input** rather than the output. And this method will work with **both** padded and unpadded values (and even incorrectly padded values, so a certain point).

I think you can probably just remove this comment?

datastore/google/cloud/datastore/key.py
```diff
@@ -333,6 +335,9 @@ def from_legacy_urlsafe(cls, urlsafe):
         :rtype: :class:`~google.cloud.datastore.key.Key`.
         :returns: The key corresponding to ``urlsafe``.
         """
+        padding = b'=' * (-len(urlsafe) % 4)
+        urlsafe += padding
```
dhermes: Please move this to **after** `_to_bytes` is used.

dhermes:
@michaelenglo I've added some new tests to actually make sure the padding gets added/stripped.

This LGTM and I'll merge one CircleCI is green.

Thanks a lot!

michaelenglo:
Thanks a lot @dhermes ! I appreciate your responsiveness and helpfulness since the day one. Hope to get more chance to learn and contribute to the library!

PR #3563, opened 2017-06-28
Making all LICENSE headers "uniform".

dhermes:
This PR is **srs** *bznz*!

(I'm not crazy, I just noticed this when quickly closing one file and the one under it had a different indent level in the same spot.)


dhermes:
Ha, exactly!

PR #3573, opened 2017-06-30
Context tracer for flask

liyanhui1228:
Implemented a context tracer for flask. It can automatically grab the trace_id from the request headers and generate a trace context. Sample usage as below:

```python
from google.cloud.trace.client import Client
from google.cloud.trace.tracer.flask_tracer import FlaskTracer

client = Client('project_id')
tracer = FlaskTracer(client)
tracer.start_trace()

with tracer.span('span1') as span1:
      do_something()
      with span1.span('child1') :
            do_something()

tracer.end_trace()
```


trace/google/cloud/trace/trace.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, client, project_id=None, trace_id=None):
             trace_id = generate_trace_id()
 
         self.trace_id = trace_id
+        self.spans = []
```
duggelz: Add description of .spans to doc comment above

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+# limitations under the License.
+
+"""TraceContext encapsulates the current context within the request's trace."""
+
+from google.cloud.trace.trace import generate_trace_id
+
+import re
+import logging
+
+_TRACE_CONTEXT_HEADER_FORMAT = '([0-9a-f]{32})(\/(\d+))?(;o=(\d+))?'
```
duggelz: Is there a publicly available document that we can link to that describes this format?

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class TraceContext(object):
+    """TraceContext includes 3 fields: traceId, spanId, and an enabled flag 
+    which indicateswhether or not the request is being traced.
```
duggelz: indicateswhether -> indicates whether

TraceContext and ContextTracer and Trace are all very similar names.  Can we make them more distinguishable?  Can we write a sentence describing their relationship to each other?  I think it's like:

Trace: All the spans for a single trace
TraceContext: A textual representation of single span inside a Trace, sent and received in HTTP and RPC messages.  Maybe it should be called SpanSomething instead of TraceSomething? Maybe it should be called TraceHeader?)
ContextTracer: A stack of Spans inside a single Trace.

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+                span_id = int(span_id)
+            except ValueError:
+                logging.warning(
+                    'The type of span_id should be int, got {}.'.format(
+                        span_id.__class__.__name__))
+                span_id = None
+
+        return span_id
+
+    def convert_to_string(self):
```
duggelz: It might be nicer to make this the standard string operator, e.g. `def __str__(self)`, so that `str(a_trace_context)` and `print(a_trace_context)` do the right thing.

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+            self.span_id,
+            self.enabled)
+        return header
+
+
+def generate_context_from_header(header):
+    """Parse a request header and generate a TraceContext object.
+
+    :type header: 
+    :param header: Request header from web application.
```
duggelz: Can you clarify that this is specifically the value of a single HTTP header?

A reader doesn't know whether this is the entire set of HTTP headers ("Foo: bar\nBaz: bazz"), a single HTTP header ("Foo: bar"), or the value of a single header ("bar"), without looking at the regular expression.

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+    :rtype: :class:`~google.cloud.trace.trace_context.TraceContext`
+    :returns: A TraceContext object.
+    """
+    pattern = re.compile(_TRACE_CONTEXT_HEADER_FORMAT)
+    match = re.search(pattern, header)
+
+    if match:
+        trace_id = match.group(1)
+        span_id = match.group(3)
+        enabled = match.group(5)
```
duggelz: You should explicitly turn this into a `bool`, or else update the class doc comment.

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+            enabled = 1
+
+        trace_context = TraceContext(
+            trace_id=trace_id,
+            span_id=span_id,
+            enabled=enabled,
+            from_header=True)
+        return trace_context
+    else:
+        return TraceContext()
```
duggelz: This is somewhat surprising behavior, I would expect an exception to be thrown if the header can't be parsed.

trace/google/cloud/trace/trace_context.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(
+            self,
+            trace_id=None,
+            span_id=None,
+            enabled=True,
+            from_header=False):
+        if trace_id is None:
+            trace_id = generate_trace_id()
+
+        self.trace_id = trace_id
```
duggelz: This should probably check that the trace id is in the expected format if the user passed it in, i.e. 32 hex digits.

trace/google/cloud/trace/tracer/flask_tracer.py
```diff
@@ -0,0 +1,61 @@
...
+
+try:
+    import flask
+except ImportError:  # pragma: NO COVER
+    flask = None
+
+from google.cloud.trace.trace_context import generate_context_from_header
+from google.cloud.trace.tracer.context_tracer import ContextTracer
+
+_FLASK_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
```
duggelz: Should this header name be in trace_context, since it's not specific to Flask?

trace/google/cloud/trace/tracer/flask_tracer.py
```diff
@@ -0,0 +1,61 @@
...
+
+    :rtype: str
+    :returns: Trace context header in HTTP request headers.
+    """
+    if flask is None or not flask.request:
+        return None
+
+    header = flask.request.headers.get(_FLASK_TRACE_HEADER)
+
+    if header is None:
```
duggelz: Can probably just `return flask.request.headers.get(_FLASK_TRACE_HEADER)`

trace/google/cloud/trace/trace.py
liyanhui1228: Done.

trace/google/cloud/trace/trace_context.py
liyanhui1228: I found this in the SD Trace document which describes the format: https://cloud.google.com/trace/docs/faq

trace/google/cloud/trace/trace_context.py
liyanhui1228: Yeah they are confusing. TraceContext encapsulates the current context to be propagate to the child spans. I would just change its name to SpanContext.

trace/google/cloud/trace/trace_context.py
liyanhui1228: Done.

trace/google/cloud/trace/trace_context.py
liyanhui1228: Done.

trace/google/cloud/trace/trace_context.py
liyanhui1228: Done.

trace/google/cloud/trace/trace_context.py
liyanhui1228: Done.

trace/google/cloud/trace/trace_context.py
liyanhui1228: I think it might be possible that the web application is not running on GCP thus the header does not contain the trace_id, but it should still be able to trace the request.

trace/google/cloud/trace/tracer/flask_tracer.py
liyanhui1228: Done.

trace/google/cloud/trace/tracer/flask_tracer.py
liyanhui1228: Done.

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+    current context to be propagate to the child spans.
+    
+    :type trace_id: str
+    :param trace_id: (Optional) Trace_id is a 32 digits uuid for the trace.
+                     If not given, will generate one automatically.
+
+    :type span_id: int
+    :param span_id: (Optional) Identifier for the span, unique within a trace.
+
+    :type enabled: bool
```
bogdandrutu: I would make it a bitmask this, similar with https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/TraceOptions.java

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+class SpanContext(object):
+    """SpanContext includes 3 fields: traceId, spanId, and an enabled flag 
+    which indicates whether or not the request is being traced. It contains the
+    current context to be propagate to the child spans.
+    
+    :type trace_id: str
+    :param trace_id: (Optional) Trace_id is a 32 digits uuid for the trace.
+                     If not given, will generate one automatically.
+
+    :type span_id: int
```
bogdandrutu: To avoid endianness problems I would avoid using ints. Probably better to do something similar with https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/SpanId.java

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+_TRACE_HEADER_KEY = 'X_CLOUD_TRACE_CONTEXT'
+_TRACE_ID_FORMAT = '[0-9a-f]{32}?'
+
+
+class SpanContext(object):
+    """SpanContext includes 3 fields: traceId, spanId, and an enabled flag 
+    which indicates whether or not the request is being traced. It contains the
+    current context to be propagate to the child spans.
+    
+    :type trace_id: str
```
bogdandrutu: I know that we usually use it as hexed (base16), but I would rather keep it as 16-bytes array and do the base16 conversion when needed. Reasons:
* Save space;
* gRPC binary encoding uses the 16-bytes array;

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+    :param trace_id: (Optional) Trace_id is a 32 digits uuid for the trace.
+                     If not given, will generate one automatically.
+
+    :type span_id: int
+    :param span_id: (Optional) Identifier for the span, unique within a trace.
+
+    :type enabled: bool
+    :param enabled: (Optional) Indicates whether the request is traced or not.
+    
+    :type from_header: bool
```
bogdandrutu: Why is this information needed in the SpanContext? I am not arguing that it should not be, but I try to understand the motivation.

trace/google/cloud/trace/trace_span.py
```diff
@@ -105,11 +105,11 @@ def span(self, name='child_span'):
         self._child_spans.append(child_span)
         return child_span
 
-    def set_start_time(self):
+    def start(self):
         """Set the start time for a span."""
         self.start_time = datetime.utcnow().isoformat() + 'Z'
 
-    def set_end_time(self):
+    def finish(self):
```
bogdandrutu: I would call this end.

trace/google/cloud/trace/trace_span.py
```diff
@@ -105,11 +105,11 @@ def span(self, name='child_span'):
         self._child_spans.append(child_span)
         return child_span
 
-    def set_start_time(self):
+    def start(self):
```
bogdandrutu: Is this public API needed? Usually when you create a Span object you also start it.

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+            from_header=False):
+        if trace_id is None:
+            trace_id = generate_trace_id()
+
+        self.trace_id = self.check_trace_id(trace_id)
+        self.span_id = self.check_span_id(span_id)
+        self.enabled = enabled
+        self.from_header = from_header
+
+    def check_span_id(self, span_id):
```
bogdandrutu: trace_id == 0 and span_id == 0 are considered invalid in our data model. so probably you would want to check for that as well.

trace/google/cloud/trace/span_context.py
```diff
@@ -0,0 +1,164 @@
...
+def generate_context_from_header(header):
+    """Parse a request header and generate a SpanContext object.
+
+    :type header: str
+    :param header: Value of a single HTTP request header from web application.
+
+    :rtype: :class:`~google.cloud.trace.trace_context.SpanContext`
+    :returns: A SpanContext object.
+    """
+    pattern = re.compile(_TRACE_CONTEXT_HEADER_FORMAT)
```
bogdandrutu: Using regexp will be very expensive. Also we want to support multiple encoding formats:
* binary
* current google cloud format
* the new trace context format that we try to unify between zipkin, opencensus, etc. (https://github.com/TraceContext/tracecontext-spec/pull/1)

So maybe have a class:
cloud/google/trace/propagation/GoogleCloudFormat (with toHeader/fromHeader) that works with SpanContext.

Later we can add BinaryFormat that is used in gRPC (similar with https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/trace/propagation/BinaryFormatImpl.java)

trace/google/cloud/trace/span_context.py
liyanhui1228: Thanks for this information! I just have a question that what should I return if the span_id or trace_id is not valid? Currently I generate a new one instead to create the SpanContext and output a warning log (which means ignore the current context), but it might also make sense to throw exception and stop creating SpanContext. Which way do you prefer?

trace/google/cloud/trace/trace_span.py
liyanhui1228: This is for usage like:
```
span.start()
do_something()
span.finish()
```
There is another usage like:
```
with span:
      do_something()
```

trace/google/cloud/trace/span_context.py
duggelz: That seems like an unnecessary conversion.  We initially get the value as a string object when we parse it from inbound http headers, and the Python API for outbound grpc requests will take it from us as a string object.  Specifically we'll make grpc calls passing in a dictionary of string to string like `metadata: {'X_CLOUD_TRACE_CONTEXT': whatever.trace_id, ... other metadata ... }`.  Likewise, we'll pass it to the Python protobuf constructor as a string object when we're recording the trace to the trace service.  We don't ever have a use for it as a compact byte array.

trace/google/cloud/trace/span_context.py
duggelz: We use int here to match the Python protobuf API.  The span field is defined as an int, and we assume that protobuf handles all endianness issues.

trace/google/cloud/trace/span_context.py
duggelz: Conceptual, we have "options" which is an int, and "enabled" which is a bool corresponding to a particular bit in options.  Or should that bit be called "sampled", because it's not "yes"/"no", it's like "definitely yes"/"maybe"?

I could see going either way.  Probably that means that "enabled" should be a method or computed property, and the actual implementation as "int + bitmask" vs "bunch of bools" can be a private implementation detail.

trace/google/cloud/trace/span_context.py
liyanhui1228: This is indicating whether this SpanContext is created from request header. Because currently when I check the format of the trace_id and span_id, if the value parsed from header is not valid, I just ignore it and create SpanContext using a newly generated trace_id or span_id. So I think it might be useful to have a variable indicating whether the SpanContext is from a header or not.

trace/google/cloud/trace/trace_span.py
liyanhui1228: Done.

liyanhui1228:
This will be living in opencensus repository, closing it now.

PR #3578, opened 2017-07-05
update doc reference for setting up a service account

SeanBE:
First commit to the repo. The error sends you to a dead page. I updated the reference. I signed a CLA.


tseaver:
Thank you!

PR #3580, opened 2017-07-05
Update 'QueryResponse.fetch_data' docstring to match changes from #3484.

### LINKED ISSUE #3576: Docstrings from fetch_data method needs to be updated
Hello,

Recently a [question](https://stackoverflow.com/questions/44895829/bigquery-async-query-job-the-fetch-results-method-returns-wrong-number-of-va) was posted in SO asking about the return params of the `fetch_data` method of `QueryResults' in BigQuery client.

As it seems, the return is no longer a tuple with 3 values but rather an object of the class `HTTPIterator`. Looks like the docstrings needs to be updated (if it could describe some quick example of how to use the iterator it'd be quite helpful :)!)

Thanks,

tseaver:
Closes #3576.

### DIFF
--- a/bigquery/google/cloud/bigquery/query.py
+++ b/bigquery/google/cloud/bigquery/query.py
@@ -414,13 +414,12 @@ class QueryResults(object):
         :param client: the client to use.  If not passed, falls back to the
                        ``client`` stored on the current dataset.
 
-        :rtype: tuple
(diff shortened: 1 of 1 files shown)


tseaver:
Docstring only fix.  I will merge w/o review.

PR #3585, opened 2017-07-06
Shorten nox virtualenv names to avoid hashing.

tseaver:


dhermes:
w00t! LGTM once CI goes green. (I also want to make sure CircleCI runs **all** the packages, which might keep it from going green, boo.)

/cc @jonparrott

duggelz:
Just curious, since I use nox in another repo, can you explain what "avoid hashing" means here?  What were the old dirnames and why were they bad?

dhermes:
@duggelz It'd be easier to just have @jonparrott explain, but I'll try. AFAIK, on Unix (or maybe just in Bash?) there is a limit on the length of a shebang line. So, when `nox` creates a directory `/my/path/.nox/my-kind-of-long-test-name-param-3.1`, it checks if the total length of that path is "too long". If it is "too long", then `nox` will automatically replace `my-kind-of-long-test-name-param-3.1` with some shortened hash.

jonparrott:
@dhermes is right, see https://github.com/pypa/pip/issues/1773 and https://github.com/jonparrott/nox/commit/b9a6af03f08d88e23466ce474a27cbf9a598da24

PR #3586, opened 2017-07-06
Making requirements tighter for "stable" packages in 0.26.1 release.

dhermes:
See #3579 for context.

Issue #3579, opened 2017-07-05
Conflicting Dependencies


lukesneeringer:
This is reposted from an internal ticket. (Googlers: b/63062329)

-----------------------------

Issue summary:

```
# Install google-cloud 0.25.0 in Ubuntu (16.04, 16.10, 17.04)
sudo apt-get update
sudo apt-get install python-pip
pip install google-cloud==0.25.0

# Its dependencies have conflicts
python -c "__requires__=['google-cloud']; import pkg_resources" <-- will receive error:
```

Here is the traceback:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2927, in <module>
    @_call_aside
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2913, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2940, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 637, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 650, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 834, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (google-cloud-core 0.24.1 (/usr/local/lib/python2.7/dist-packages), Requirement.parse('google-cloud-core<0.26dev,>=0.25.0'), set(['google-cloud-storage', 'google-cloud-logging', 'google-cloud-datastore']))
```

Workaround: Customer has reverted to 0.23.0 to build.

Very similar to #3331.

dhermes:
@lukesneeringer ISTM the issue is that

- They had an existing installation that conflicts with the latest version
- The bug is somewhat in `pip` / `setuptools`?

----

Here is the latest state of things:

-   [`google-cloud==0.26.0`][2]
    -   `google-cloud-core==0.25.*`
    -   `google-cloud-datastore >= 1.1.0, < 2.0dev`
    -   `google-cloud-logging >= 1.1.0, < 2.0dev`
[...]

PR #3588, opened 2017-07-06
Making requirements tighter for "stable" packages in error reporting 0.25.1 release.

dhermes:
See #3579 and #3586 for context.


dhermes:
Merging with permission of @lukesneeringer

dhermes:
New release pushed:
https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/error-reporting-0.25.1

Here is the CircleCI build that **should** push to PyPI:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2329

However, I may end up having to do it by hand due to current build issues (which is "just fine").

PR #3591, opened 2017-07-06
Patch in the version updates made when fixing #3579.

dhermes:
Changes from #3586 and #3588.


error_reporting/setup.py
```diff
@@ -52,13 +52,13 @@
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.25.0, < 0.26dev',
-    'google-cloud-logging >= 1.0.0, < 2.0dev',
+    'google-cloud-logging >= 1.1.0, < 1.2dev',
```
tseaver: I'm not positive that tightening the upper bound is the Right Thing(TM).  ISTM that semver requires that we do not make any breaking changes to `google-cloud-logging` before a 2.0.0 release.

setup.py
```diff
@@ -53,26 +53,26 @@
...
-    'google-cloud-logging >= 1.1.0, < 2.0dev',
+    'google-cloud-logging >= 1.1.0, < 1.2dev',
     'google-cloud-monitoring >= 0.25.0, < 0.26dev',
     'google-cloud-pubsub >= 0.26.0, < 0.27dev',
     'google-cloud-resource-manager >= 0.25.0, < 0.26dev',
     'google-cloud-runtimeconfig >= 0.25.0, < 0.26dev',
     'google-cloud-spanner >= 0.25.0, < 0.26dev',
     'google-cloud-speech >= 0.26.0, < 0.27dev',
-    'google-cloud-storage >= 1.2.0, < 2.0dev',
+    'google-cloud-storage >= 1.2.0, < 1.3dev',
```
tseaver: Likewise for these three "silver" APIs:  anybody should be able to upgrade them to any 1.x release (above the lower bound) without risking a breaking change.

dhermes:
@tseaver I recommend reading #3579 for the "why" here.

The issue isn't API compatibility across versions, it is dependency compatibility.

error_reporting/setup.py
lukesneeringer: I think @tseaver is correct here. Why is `< 2.0dev` wrong?

lukesneeringer: Nevermind, I figured it out.

setup.py
dhermes: The issue is "us". We have made every other package in here depend explicitly on the `0.25.micro` series of `google-cloud-core`. If one of the (possibly many) future releases of `google-cloud-storage`  in the `1.minor.micro` series depends on a **different** segment of `google-cloud-core`, then installing the umbrella package will lead to an incompatible set of version constraints.

tseaver: In retrospect, ISTM that we should've made a `1.0` release of the `core` package at the time we pushed datastore / storage / logging to `1.0`, and we should then be imposing true semver on that package as well.

dhermes: Totally. And we should **continue** that discussion.

But, while we figure out the best way to go forward, this PR should merge? (It is the best way to stay right where we are without being broken.)

lukesneeringer: I assert this PR should merge.

dhermes:
Merged based on @lukesneeringer assertion.

@tseaver would you like to take the lead on deciding "what now" about these versioning issues with `google-cloud-core`?

PR #3593, opened 2017-07-07
Add support for logging the trace-id in webapp2 apps.

csilvers:
Code was added in #3448 that set the trace-id when used in the
flask or django web frameworks.

This extends that to also support webapp2.

webapp2 is not cutting edge technology anymore -- it doesn't even work
on python3 -- but a lot of legacy GAE classic apps use webapp2, so I
think it makes sense to support it.

I tested this by running `nox` in the `logging/` directory, which as far
as I can tell is current state of the art for running tests.

### DIFF
--- a/logging/google/cloud/logging/handlers/_helpers.py
+++ b/logging/google/cloud/logging/handlers/_helpers.py
@@ -22,10 +22,20 @@ try:
 except ImportError:  # pragma: NO COVER
     flask = None
 
+try:
+    import webapp2
+except (ImportError, SyntaxError):  # pragma: NO COVER
+    # If you try to import webapp2 under python3, you'll get a syntax
+    # error (since it hasn't been ported yet).  We just pretend it
+    # doesn't exist.  This is unlikely to hit in real life but does
+    # in the tests.
+    webapp2 = None
+
 from google.cloud.logging.handlers.middleware.request import (
     _get_django_request)
 
 _FLASK_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
+_WEBAPP2_TRACE_HEADER = 'x-cloud-trace-context'
 _DJANGO_TRACE_HEADER = 'HTTP_X_CLOUD_TRACE_CONTEXT'
 
 
@@ -69,6 +79,33 @@ def get_trace_id_from_flask():
     return trace_id
 
 
+def get_trace_id_from_webapp2():
+    """Get trace_id from webapp2 request headers.
+
+    :rtype: str
+    :return: Trace_id in HTTP request headers.
+    """
+    if webapp2 is None:
+        return None
+
+    try:
+        # get_request() succeeds if we're in the middle of a webapp2
+        # request, or raises an assertion error otherwise:
+        # "Request global variable is not set".
+        req = webapp2.get_request()
+    except AssertionError:
+        return None
+
+    header = req.headers.get(_WEBAPP2_TRACE_HEADER)
+
+    if header is None:
+        return None
+
+    trace_id = header.split('/', 1)[0]
+
+    return trace_id
+
+
 def get_trace_id_from_django():
     """Get trace_id from django request headers.
 
@@ -95,7 +132,8 @@ def get_trace_id():
     :rtype: str
     :returns: Trace_id in HTTP request headers.
     """
-    checkers = (get_trace_id_from_django, get_trace_id_from_flask)
+    checkers = (get_trace_id_from_django, get_trace_id_from_flask,
+                get_trace_id_from_webapp2)
 
     for checker in checkers:
         trace_id = checker()
--- a/logging/google/cloud/logging/handlers/_helpers.py
+++ b/logging/google/cloud/logging/handlers/_helpers.py
@@ -34,9 +34,9 @@ except (ImportError, SyntaxError):  # pragma: NO COVER
 from google.cloud.logging.handlers.middleware.request import (
     _get_django_request)
 
-_FLASK_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
-_WEBAPP2_TRACE_HEADER = 'x-cloud-trace-context'
 _DJANGO_TRACE_HEADER = 'HTTP_X_CLOUD_TRACE_CONTEXT'
+_FLASK_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
+_WEBAPP2_TRACE_HEADER = 'X-CLOUD-TRACE-CONTEXT'
 
 
 def format_stackdriver_json(record, message):
@@ -64,7 +64,7 @@ def get_trace_id_from_flask():
     """Get trace_id from flask request headers.
 
     :rtype: str
-    :return: Trace_id in HTTP request headers.
+    :returns: TraceID in HTTP request headers.
     """
     if flask is None or not flask.request:
         return None
@@ -83,7 +83,7 @@ def get_trace_id_from_webapp2():
     """Get trace_id from webapp2 request headers.
 
     :rtype: str
-    :return: Trace_id in HTTP request headers.
+    :returns: TraceID in HTTP request headers.
     """
     if webapp2 is None:
         return None
@@ -110,7 +110,7 @@ def get_trace_id_from_django():
     """Get trace_id from django request headers.
 
     :rtype: str
-    :return: Trace_id in HTTP request headers.
+    :returns: TraceID in HTTP request headers.
     """
     request = _get_django_request()
 
@@ -130,9 +130,10 @@ def get_trace_id():
     """Helper to get trace_id from web application request header.
 
     :rtype: str
-    :returns: Trace_id in HTTP request headers.
+    :returns: TraceID in HTTP request headers.
     """
-    checkers = (get_trace_id_from_django, get_trace_id_from_flask,
+    checkers = (get_trace_id_from_django,
+                get_trace_id_from_flask,
                 get_trace_id_from_webapp2)
 
     for checker in checkers:
--- a/logging/nox.py
+++ b/logging/nox.py
@@ -36,7 +36,7 @@ def unit_tests(session, python_version):
     # Install all test dependencies, then install this package in-place.
     session.install(
         'mock', 'pytest', 'pytest-cov',
-        'flask', 'django', *LOCAL_DEPS)
+        'flask', 'webapp2', 'webob', 'django', *LOCAL_DEPS)
     session.install('-e', '.')
 
     # Run py.test against the unit tests.
--- a/logging/tests/unit/handlers/test__helpers.py
+++ b/logging/tests/unit/handlers/test__helpers.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 import unittest
 
 import mock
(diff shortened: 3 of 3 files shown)


csilvers:
Our company (Khan Academy) has signed the CLA.  I've been added to the group of authorized contributors.  Maybe I need to be added under my github.com email and not my company email.  I've done that and will try again.

csilvers:
(I made a new comment but it doesn't look like the clabot has run again.  I'm not sure what else I need to do.)

duggelz:
@lukesneeringer How do we poke the clabot?  (Craig is/was Google employee `#1` so I'll vouch for him if need be :)

duggelz:
The clabot doesn't think that Khan Academy has signed the corporate CLA (unless it has some wierd legal name that doesn't include the words Khan or Academy), nor does it think that csilvers (GitHub username) or csilvers@khanacademy.org has signed the individual CLA.

The docs says that corporate CLAs might take a few days to process, was it signed recently?

csilvers:
Ah, that sounds like it must be it.  I just signed the corporate CLA earlier today.  I'm happy to wait (though ideally it wouldn't need to block review!)

csilvers:
btw, hi doug!

duggelz:
Hi Craig! :)

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -22,10 +22,20 @@
 except ImportError:  # pragma: NO COVER
     flask = None
 
+try:
+    import webapp2
+except (ImportError, SyntaxError):  # pragma: NO COVER
+    # If you try to import webapp2 under python3, you'll get a syntax
+    # error (since it hasn't been ported yet).  We just pretend it
+    # doesn't exist.  This is unlikely to hit in real life but does
+    # in the tests.
```
dhermes: Ouch

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -69,6 +79,33 @@ def get_trace_id_from_flask():
...
+    :rtype: str
+    :return: Trace_id in HTTP request headers.
+    """
+    if webapp2 is None:
+        return None
+
+    try:
+        # get_request() succeeds if we're in the middle of a webapp2
+        # request, or raises an assertion error otherwise:
+        # "Request global variable is not set".
```
dhermes: `webapp2` has some real warts, eh?

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -69,6 +79,33 @@ def get_trace_id_from_flask():
     return trace_id
 
 
+def get_trace_id_from_webapp2():
+    """Get trace_id from webapp2 request headers.
+
+    :rtype: str
+    :return: Trace_id in HTTP request headers.
```
dhermes: s/Trace_id/Trace ID?

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -95,7 +132,8 @@ def get_trace_id():
     :rtype: str
     :returns: Trace_id in HTTP request headers.
     """
-    checkers = (get_trace_id_from_django, get_trace_id_from_flask)
+    checkers = (get_trace_id_from_django, get_trace_id_from_flask,
+                get_trace_id_from_webapp2)
```
dhermes: Nit: Can we just bite the bullet and make it:

```python
checkers = (
    get_trace_id_from_django,
    get_trace_id_from_flask,
    get_trace_id_from_webapp2,
)
```

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -69,6 +79,33 @@ def get_trace_id_from_flask():
...
+        req = webapp2.get_request()
+    except AssertionError:
+        return None
+
+    header = req.headers.get(_WEBAPP2_TRACE_HEADER)
+
+    if header is None:
+        return None
+
+    trace_id = header.split('/', 1)[0]
```
dhermes: Should there be a helper for splitting the header in this way by now?

logging/nox.py
```diff
@@ -36,7 +36,7 @@ def unit_tests(session, python_version):
     # Install all test dependencies, then install this package in-place.
     session.install(
         'mock', 'pytest', 'pytest-cov',
-        'flask', 'django', *LOCAL_DEPS)
+        'flask', 'webapp2', 'webob', 'django', *LOCAL_DEPS)
```
dhermes: Does installing these fail in Python 3? Worth conditionally adding the deps in Python 2 only?

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
         self.assertEqual(trace_id, expected_trace_id)
 
 
+# webapp2 has not been ported to python3 yet, so do not try to test its
+# functionality there.
+if sys.version_info[0] == 2:
```
dhermes: Instead, use `unittest.skipIf(six.PY3, 'webapp2 is Python 2 only')`

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
...
+    class Test_get_trace_id_from_webapp2(unittest.TestCase):
+
+        @staticmethod
+        def create_app():
```
dhermes: Why not just make this a function at module scope?

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
...
+                    self.response.out.write(trace_id)
+
+            app = webapp2.WSGIApplication([
+                ('/', GetTraceId),
+            ])
+
+            return app
+
+        def setUp(self):
+            self.app = self.create_app()
```
dhermes: Is there a **need** to do this in `setUp`? If not, (strongly) prefer just creating a test local `app` in each test case.

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
...
+# webapp2 has not been ported to python3 yet, so do not try to test its
+# functionality there.
+if sys.version_info[0] == 2:
+    class Test_get_trace_id_from_webapp2(unittest.TestCase):
+
+        @staticmethod
+        def create_app():
+            import webapp2
+
+            class GetTraceId(webapp2.RequestHandler):
```
dhermes: Ditto here, define this handler at module scope?

```python
try:
    from webapp2 import RequestHandler
except (ImportError, SyntaxError):
    RequestHandler = object


class GetTraceId(RequestHandler):
    ...
```

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
...
+                def get(self):
+                    from google.cloud.logging.handlers import _helpers
+
+                    trace_id = _helpers.get_trace_id_from_webapp2()
+                    self.response.content_type = "text/plain"
```
dhermes: Nit: Single quotes instead of double quotes

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -63,6 +64,57 @@ def test_valid_context_header(self):
...
+            self.app = self.create_app()
+
+        def test_no_context_header(self):
+            import webob
+
+            req = webob.BaseRequest.blank('/')
+            response = req.get_response(self.app)
+            trace_id = response.body
+
+            self.assertEquals("None", trace_id)
```
dhermes: 1. Why the string `None` versus the singleton?
1. Single quotes preferred

logging/tests/unit/handlers/test__helpers.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
```
dhermes: This import can go when you `import six` and use `six.PY3`

logging/google/cloud/logging/handlers/_helpers.py
```diff
@@ -22,10 +22,20 @@
...
+    # error (since it hasn't been ported yet).  We just pretend it
+    # doesn't exist.  This is unlikely to hit in real life but does
+    # in the tests.
+    webapp2 = None
+
 from google.cloud.logging.handlers.middleware.request import (
     _get_django_request)
 
 _FLASK_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
+_WEBAPP2_TRACE_HEADER = 'x-cloud-trace-context'
```
csilvers: Sure thing, will do.

logging/google/cloud/logging/handlers/_helpers.py
csilvers: OK, I was just copying from what the other functions in this file did.  I'll fix them all.  (Similarly for trace_id below.)

logging/google/cloud/logging/handlers/_helpers.py
csilvers: Yeah, the AssertionError thing isn't part of the API :-), and I'm kinda worried that one day webapp2 will get updated in a way that raises a different exception.  But I think the tests will notice if that ever happens.

logging/google/cloud/logging/handlers/_helpers.py
csilvers: It's only one line of code, so I was happy copying it, but I can make a helper function too if you think it's clearer.

logging/google/cloud/logging/handlers/_helpers.py
csilvers: Bullet bitten!  (Well, it will be.)

logging/nox.py
csilvers: They install fine in python3, you just can't use them.  I figure that's fine -- it lets us check that SyntaxError thing in the test -- so I'm happy enough to leave it like this.

But if you'd prefer to conditionally install the deps just for python2 -- under the theory nobody is going to be running python3 with webapp installed, so we could take out the SyntaxError handling -- that would make sense to me too.  I just don't know the syntax for it. :-)

logging/tests/unit/handlers/test__helpers.py
csilvers: :+1:

logging/tests/unit/handlers/test__helpers.py
csilvers: Nice!  I figured I wasn't doing it in the best way.  Though I do think that the interpreter still tries to byte-compile the test in this case?  Which means that in python3 it will still try to import webapp2 and break.  I guess if we do your other suggestion we'll catch the problem that way.

logging/tests/unit/handlers/test__helpers.py
csilvers: I was just following what the existing tests here do.  I have no strong feelings myself.  I do feel it's nice to keep all the logic for a test encapsulated in the test class, to make it clear what helper objects go with what test, but I'm happy to change it if you have a preference.

logging/tests/unit/handlers/test__helpers.py
csilvers: That would work, but it seems a bit ugly to have to repeat the exception-list here.  That said, if I go to unittest.skip it's probably going to be necessary.

logging/tests/unit/handlers/test__helpers.py
csilvers: :+1:

logging/tests/unit/handlers/test__helpers.py
csilvers: Again, I was just copying what the other tests did.  I will change them all to create a test-local `app`.

logging/tests/unit/handlers/test__helpers.py
csilvers: (1) is just a side-effect of how I wrote the test: I had the handler return the trace-id as text.  If the trace-id is None, then it gets auto-stringified to `None`.  I don't think webapp2 has something like flask's `test_request_context` so we have to go through a string intermediary at some point.

I could return json instead of just a raw text string, and then the json would convert `null` back to `None` -- would that be better?

liyanhui1228:
@csilvers Yeah there is some inconsistent in the repo for `return` and `returns`, I confirmed with Jon that we should use `returns`. Sorry for the confusion.

Also I have just tested your changes in real life and it works well! I know it will take some time to figure out how to test it in real life because I did the same thing when I worked on this feature. Thank you again for sending this PR :)

csilvers:
Thanks @liyanhui1228 for doing the real-life testing!  And also for your code: I got compliments on my coding style and that's entirely because I cribbed from you. :-)

logging/tests/unit/handlers/test__helpers.py
csilvers: (I'll do it that way.)

csilvers:
My latest commit addresses all the issues but these, which I'm leaving as-is unless people feel strongly:
* I didn't use a helper function for extracting a trace-id from the trace-id/span-id pair.
* I kept the code that installs webapp2 even for python3 tests (which ignore it).

liyanhui1228:
This LGTM, but wait for @dhermes for merging. Thanks!

csilvers:
@liyanhui1228 do you mind approving it?  (However you do that in github.) It still has you marked as requesting changes.

lukesneeringer:
Good to merge. Ignore the CLA bot.

duggelz:
CLA bot now lists Khan Academy as a signer.

PR #3599, opened 2017-07-11
Spanner p0 system tests (batch #1)

tseaver:
Additional systests for `Snapshot.read`:

- Read single key.
- Read multiple keys.
- Read open-closed ranges.
- Read open-open ranges.
- Read closed-open ranges.
- Read closed-closed ranges.

Additional system test for `Snapshot.execute_sql`:

- Query returning `ARRAY<STRUCT>`.
- Bind `INT64` parameter to null.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -786,6 +786,14 @@ class TestSessionAPI(unittest.TestCase, _TestData):
 
         self._check_sql_results(
             snapshot,
+            sql='SELECT description FROM all_types WHERE eye_d = @my_id',
+            params={'my_id': None},
+            param_types={'my_id': Type(code=INT64)},
+            expected=[],
+        )
+
+        self._check_sql_results(
+            snapshot,
             sql='SELECT eye_d FROM all_types WHERE description = @description',
             params={'description': u'dog'},
             param_types={'description': Type(code=STRING)},
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -33,6 +33,7 @@ from google.cloud._helpers import UTC
 from google.cloud.exceptions import GrpcRendezvous
 from google.cloud.spanner._helpers import TimestampWithNanoseconds
 from google.cloud.spanner.client import Client
+from google.cloud.spanner.keyset import KeyRange
 from google.cloud.spanner.keyset import KeySet
 from google.cloud.spanner.pool import BurstyPool
 
@@ -584,6 +585,32 @@ class TestSessionAPI(unittest.TestCase, _TestData):
             [(row[0], row[2]) for row in self._row_data(ROW_COUNT)]))
         self._check_row_data(rows, expected)
 
+    def test_read_w_single_key(self):
+        ROW_COUNT = 40
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, KeySet(keys=[(0,)])))
+
+        all_data_rows = list(self._row_data(ROW_COUNT))
+        expected = [all_data_rows[0]]
+        self._check_row_data(rows, expected)
+
+    def test_read_w_multiple_keys(self):
+        ROW_COUNT = 40
+        indices = [0, 5, 17]
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS,
+            KeySet(keys=[(index,) for index in indices])))
+
+        all_data_rows = list(self._row_data(ROW_COUNT))
+        expected = [row for row in all_data_rows if row[0] in indices]
+        self._check_row_data(rows, expected)
+
     def test_read_w_limit(self):
         ROW_COUNT = 4000
         LIMIT = 100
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -580,6 +581,30 @@ def test_read_w_index(self):
...
+    def test_read_w_single_key(self):
+        ROW_COUNT = 40
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, KeySet(keys=[(0,)])))
+
+        all_data_rows = list(self._row_data(ROW_COUNT))
+        expected = all_data_rows[:1]
```
dhermes: This confused me at first (i.e. I wondered why you had `all_data_rows[:1]`). Will be clearer if you did something like:

```python
indices = [0]
key_set = KeySet([(index,) for index in indices])
...
expected = [all_data_rows[index] for index in indices]
```

spanner/tests/system/test_system.py
```diff
@@ -580,6 +581,30 @@ def test_read_w_index(self):
             [(row[0], row[2]) for row in self._row_data(ROW_COUNT)]))
         self._check_row_data(rows, expected)
 
+    def test_read_w_single_key(self):
```
dhermes: Worth factoring out a helper that takes a list of keys that you just dump into `KeySet`?

spanner/tests/system/test_system.py
```diff
@@ -593,21 +618,40 @@ def test_read_w_limit(self):
...
-        keyset = KeySet(ranges=(key_range,))
-
         snapshot = session.snapshot(read_timestamp=committed)
+        all_data_rows = list(self._row_data(ROW_COUNT))
+
+        closed_closed = KeyRange(start_closed=[START], end_closed=[END])
+        keyset = KeySet(ranges=(closed_closed,))
         rows = list(snapshot.read(
             self.TABLE, self.COLUMNS, keyset))
+        expected = all_data_rows[START:END+1]
```
dhermes: `END + 1` (i.e. spaces around operators)

spanner/tests/system/test_system.py
```diff
@@ -593,21 +618,40 @@ def test_read_w_limit(self):
...
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, keyset))
+        expected = all_data_rows[START:END]
+        self._check_row_data(rows, expected)
+
+        open_open = KeyRange(start_open=[START], end_open=[END])
+        keyset = KeySet(ranges=(open_open,))
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, keyset))
+        expected = all_data_rows[START+1:END]
```
dhermes: `START + 1`

spanner/tests/system/test_system.py
```diff
@@ -593,21 +618,40 @@ def test_read_w_limit(self):
...
+        self._check_row_data(rows, expected)
+
+        open_closed = KeyRange(start_open=[START], end_closed=[END])
+        keyset = KeySet(ranges=(open_closed,))
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, keyset))
+        expected = all_data_rows[START+1:END+1]
```
dhermes: `START + 1` and `END + 1`

spanner/tests/system/test_system.py
```diff
@@ -593,21 +618,40 @@ def test_read_w_limit(self):
...
+        START = 1000
+        END = 2000
         session, committed = self._set_up_table(ROW_COUNT)
-        key_range = KeyRange(start_closed=[START_CLOSED], end_open=[END_OPEN])
-        keyset = KeySet(ranges=(key_range,))
-
         snapshot = session.snapshot(read_timestamp=committed)
+        all_data_rows = list(self._row_data(ROW_COUNT))
+
+        closed_closed = KeyRange(start_closed=[START], end_closed=[END])
```
dhermes: @tseaver WDYT about re-factoring this into a loop over the four cases?

```python
cases = (
    (0, 1),
    (0, 0),
    (1, 0),
    (1, 1),
)
for start_offset, end_offset in cases:
    range_kwargs = {}
    if start_offset == 0:
        range_kwargs['start_closed'] = [START]
    else:
        range_kwargs['start_open'] = [START]
    ...
```

spanner/tests/system/test_system.py
```diff
@@ -580,6 +581,30 @@ def test_read_w_index(self):
...
+        expected = all_data_rows[:1]
+        self._check_row_data(rows, expected)
+
+    def test_read_w_multiple_keys(self):
+        ROW_COUNT = 40
+        session, committed = self._set_up_table(ROW_COUNT)
+
+        snapshot = session.snapshot(read_timestamp=committed)
+        rows = list(snapshot.read(
+            self.TABLE, self.COLUMNS, KeySet(keys=[(0,), (1,), (2,)])))
```
dhermes: How about making them non-contiguous to really flex the muscles? Like `0`, `5` and `17`?

This would require an update of `all_data_rows[:3]` below as well

spanner/tests/system/test_system.py
```diff
@@ -637,6 +681,26 @@ def _check_sql_results(self, snapshot, sql, params, param_types, expected):
             sql, params=params, param_types=param_types))
         self._check_row_data(rows, expected=expected)
 
+    def test_execute_sql_returning_array_of_struct(self):
+        SQL = (
+            "SELECT ARRAY(SELECT AS STRUCT C1, C2 "
+            "FROM (SELECT 'a' AS C1, 1 AS C2 "
+            "UNION ALL SELECT 'b' AS C1, 2 AS C2) "
+            "ORDER BY C1 ASC)"
```
dhermes: Use single-quotes in the first and last line? Leave double quotes in the middle two?

spanner/tests/system/test_system.py
tseaver: I don't think so.  These testcase methods are *short*.

dhermes: Yes they are, but they are like 100% identical (only a few chars difference)

spanner/tests/system/test_system.py
tseaver: I don't like the extra complexity:  as it sits, the names and initializers identify the ranges clearly.

spanner/tests/system/test_system.py
tseaver: PEP8 says [not needed in a slice operation](http://legacy.python.org/dev/peps/pep-0008/#pet-peeves).

spanner/tests/system/test_system.py
tseaver: The double quotes delimit the Python strings:  the single quotes delimit the SQL strings.

spanner/tests/system/test_system.py
tseaver: The point is that it is a *single* key (I don't want to share the implementation with the "mutiple keys" case).  I could maybe see computing `expected` via a list comprehension like:

```python
expected = [row for row in all_data_rows where row[0] == 0]
```

but I doubt it will be any clearer.

spanner/tests/system/test_system.py
dhermes: TIL!

spanner/tests/system/test_system.py
lukesneeringer: That is fine.

spanner/tests/system/test_system.py
tseaver: 707e44e

spanner/tests/system/test_system.py
tseaver: 707e44e

tseaver:
@dhermes, @lukesneeringer I had to cherry-pick 474cb18 into the PR because the back-end suddenly began including instance configs for zones (e.g., `asia-east1`) where my account is not permitted to create instances.

spanner/tests/system/test_system.py
lukesneeringer: Consistency within what is ultimately a single string is fine. (@dhermes, I always struggle with which one to do in this case.)

lukesneeringer:
> @dhermes, @lukesneeringer I had to cherry-pick 474cb18 into the PR because the back-end suddenly began including instance configs for zones (e.g., `asia-east1`) where my account is not permitted to create instances.

Ack. No worries.

tseaver:
Merged via #3604

PR #3601, opened 2017-07-12
Add a hashcode implementation to SchemaField

azymnis:
*Problem:* I need to compare a new schema with an existing table schema. Easiest way would be to use set comparison operators. However this is not possible with `SchemaField` because there is no implementation of `__hash__`.

*Solution:* Add a simple implementation of `__hash__` using tuple keys. This is not the most efficient implementation of a hashcode but should probably be ok for most purposes.

### DIFF
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -32,16 +32,16 @@ class SchemaField(object):
     :type description: str
     :param description: optional description for the field.
 
-    :type fields: list of :class:`SchemaField`, or None
+    :type fields: tuple of :class:`SchemaField`
     :param fields: subfields (requires ``field_type`` of 'RECORD').
     """
     def __init__(self, name, field_type, mode='NULLABLE', description=None,
-                 fields=None):
+                 fields=()):
         self.name = name
         self.field_type = field_type
         self.mode = mode
         self.description = description
-        self.fields = None if(fields is None) else tuple(fields)
+        self.fields = tuple(fields)
 
     def _key(self):
         """
@@ -60,3 +60,6 @@ class SchemaField(object):
 
     def __hash__(self):
         return hash(self._key())
+
+    def __repr__(self):
+        return "SchemaField{}".format(self._key())
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -41,9 +41,13 @@ class SchemaField(object):
         self.field_type = field_type
         self.mode = mode
         self.description = description
-        self.fields = fields
+        self.fields = None if(fields is None) else tuple(fields)
 
-    def __key(self):
+    def _key(self):
+        """
+        A tuple describing the contents of this :class:`SchemaField`.
+        Used to compute this instance's hashcode and evaluate equality.
+        """
         return (
             self.name,
             self.field_type.lower(),
@@ -52,7 +56,7 @@ class SchemaField(object):
             self.fields)
 
     def __eq__(self, other):
-        return self.__key() == other.__key()
+        return self._key() == other._key()
 
     def __hash__(self):
-        return hash(self.__key())
+        return hash(self._key())
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -26,7 +26,7 @@ class SchemaField(object):
                        'FLOAT', 'BOOLEAN', 'TIMESTAMP' or 'RECORD').
 
     :type mode: str
-    :param mode: the type of the field (one of 'NULLABLE', 'REQUIRED',
+    :param mode: the mode of the field (one of 'NULLABLE', 'REQUIRED',
                  or 'REPEATED').
 
     :type description: str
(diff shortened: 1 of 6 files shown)


azymnis:
I signed!

dhermes:
I was a bit confused by what is needed, but this example clarifies for me:

```python
>>> class A(object):
...     def __init__(self, val):
...         self.val = val
...     def __repr__(self):
...         return 'A({!r})'.format(self.val)
...     def __eq__(self, other):
...         if not isinstance(other, A):
...             return NotImplemented
...         return other.val == self.val
...     def __ne__(self, other):
...         if not isinstance(other, A):
...             return NotImplemented
...         return other.val != self.val
...
>>> a1 = A(10)
>>> a2 = A(11)
>>> a3 = A(11)
>>> a1 == a2
False
>>> a1 == a3
False
>>> a2 == a3
True
>>> set([a1, a2, a3])
set([A(10), A(11), A(11)])
```

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -43,10 +43,16 @@ def __init__(self, name, field_type, mode='NULLABLE', description=None,
         self.description = description
         self.fields = fields
 
-    def __eq__(self, other):
+    def __key(self):
```
dhermes: Please just use `_key`

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -43,10 +43,16 @@ def __init__(self, name, field_type, mode='NULLABLE', description=None,
...
-            self.name == other.name and
-            self.field_type.lower() == other.field_type.lower() and
-            self.mode == other.mode and
-            self.description == other.description and
-            self.fields == other.fields)
+            self.name,
+            self.field_type.lower(),
+            self.mode,
+            self.description,
+            self.fields)
```
tseaver: For `RECORD` fields, `self.fields` is a list, which is unhashable:  it needs to be wrapped as a tuple in that case.

dhermes: That **might** be an argument for `SchemaField` to be unhashable?

tseaver: Or else switch to storing `fields` as a tuple for `RECORD` types.

bigquery/google/cloud/bigquery/schema.py
azymnis: Sure, I'll add one. Do you generally have docstrings for private methods (starting with `_`)?

bigquery/google/cloud/bigquery/schema.py
azymnis: Do you want me to switch storing fields as tuples or just for `_key`?

azymnis:
@dhermes @tseaver what do you think of this?

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -41,9 +41,13 @@ def __init__(self, name, field_type, mode='NULLABLE', description=None,
         self.field_type = field_type
         self.mode = mode
         self.description = description
-        self.fields = fields
+        self.fields = None if(fields is None) else tuple(fields)
```
dhermes: You can just use `tuple(fields)` here if you use `fields=()` as the default rather than `fields=None`

----

General Python advice, drop all superfluous parens (here I mean the parens around `(fields is None)`. (See [Google Python Style Guide][1].)

[1]: https://google.github.io/styleguide/pyguide.html#Parentheses

tseaver: I'd say that we should go ahead and make `fields=()` the default.  We would then need to adjust the docstring, and fix any unit tests which rely on having `fields` be `None`.

azymnis:
@dhermes @tseaver i modified the default arg for `fields` to be `()`. The question is what to do with how this resource is represented. Should I use `()` there too? This is what I did, so if you would rather have me use `[]` I can change that.

FYI: this broke a few tests and I'm wondering if it will break any of the actual api calls.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1110,7 +1110,7 @@ def _build_schema_resource(fields):
         if field.description is not None:
             info['description'] = field.description
         if field.fields is not None:
-            info['fields'] = _build_schema_resource(field.fields)
+            info['fields'] = tuple(_build_schema_resource(field.fields))
```
azymnis: @dhermes @tseaver what do you think?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1079,7 +1079,7 @@ def _parse_schema_resource(info):
                 present in ``info``.
     """
     if 'fields' not in info:
-        return None
+        return ()
```
azymnis: @dhermes @tseaver what do you think?

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -32,21 +32,34 @@ class SchemaField(object):
...
+            self.description,
+            self.fields)
+
+    def __eq__(self, other):
+        return self._key() == other._key()
+
+    def __hash__(self):
+        return hash(self._key())
+
+    def __repr__(self):
```
azymnis: @dhermes @tseaver thought that I might as well add this given that I'm touching this class

lukesneeringer: +1

azymnis:
Hmm looks like I broke some tests in circleci, let me go fix that

dhermes:
@azymnis We (@jonparrott, @lukesneeringer and I) just had a discussion about this and determined `SchemaField` should be immutable in order to be hashable.

Rather than put the burden on you, I have sent a commit into this PR to make the class immutable.

bigquery/google/cloud/bigquery/table.py
dhermes: I think this isn't necessary in this PR?

bigquery/tests/unit/test_job.py
```diff
@@ -491,8 +491,15 @@ def test_begin_w_alternate_client(self):
             'sourceFormat': 'CSV',
             'writeDisposition': 'WRITE_TRUNCATE',
             'schema': {'fields': [
-                {'name': 'full_name', 'type': 'STRING', 'mode': 'REQUIRED'},
-                {'name': 'age', 'type': 'INTEGER', 'mode': 'REQUIRED'},
+                {'name':
```
lukesneeringer: Can we leave these as-is unless they are messing up a line break? If they are messing up a line break, please do four-space indenting:

```python
{
    'name': 'full_name',
    'type': 'STRING',
    'mode': 'REQUIRED',
    'fields': (),
},
```

(True for all cases where this is changed.)

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -26,27 +26,84 @@ class SchemaField(object):
...
-            self.description == other.description and
-            self.fields == other.fields)
+            self._name,
+            self._field_type.lower(),
+            self._mode,
+            self._description,
+            self._fields,
+        )
+
+    def __eq__(self, other):
```
lukesneeringer: Do we need to implement `__ne__` also?

bigquery/google/cloud/bigquery/table.py
dhermes: And I was wrong. Cheers!

bigquery/google/cloud/bigquery/schema.py
dhermes: Done

azymnis:
@dhermes makes sense, thanks for the update

azymnis:
Thank you all!

PR #3609, opened 2017-07-14
Fixing merge conflict in `setup.py` for Speech.

dhermes:


speech/setup.py
```diff
@@ -37,7 +37,7 @@
     'include_package_data': True,
     'zip_safe': False,
     'classifiers': [
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
```
dhermes: @lukesneeringer @jonparrott I wasn't actually sure if "Alpha" or "Beta" was right here (it looks like the auto-generated classifiers are "Beta" but we had a human being at some point determine this package was "Alpha")

PR #3615, opened 2017-07-17
Implement multi-use snapshots

tseaver:
Multi-use snapshots trigger an "implicit" server-side transaction, and capture its ID on the first request.  Subsequent requests return that ID, allowing for isolation from other changes.  We default to `multi_use=False` because that mode is much more performant for the simple case.

This feature is one which we originally decided to leave out, but the P0 system test list requires that it be implemented.

I *think* that a commitwise review might be easier than reviewing the whole enchilada.

### DIFF
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -380,8 +380,7 @@ class Database(object):
         """
         return BatchCheckout(self)
 
-    def snapshot(self, read_timestamp=None, min_read_timestamp=None,
-                 max_staleness=None, exact_staleness=None):
+    def snapshot(self, **kw):
         """Return an object which wraps a snapshot.
 
         The wrapper *must* be used as a context manager, with the snapshot
@@ -390,38 +389,15 @@ class Database(object):
         See
         https://cloud.google.com/spanner/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionOptions.ReadOnly
 
-        If no options are passed, reads will use the ``strong`` model, reading
-        at a timestamp where all previously committed transactions are visible.
-
-        :type read_timestamp: :class:`datetime.datetime`
-        :param read_timestamp: Execute all reads at the given timestamp.
-
-        :type min_read_timestamp: :class:`datetime.datetime`
-        :param min_read_timestamp: Execute all reads at a
-                                   timestamp >= ``min_read_timestamp``.
-
-        :type max_staleness: :class:`datetime.timedelta`
-        :param max_staleness: Read data at a
-                              timestamp >= NOW - ``max_staleness`` seconds.
-
-        :type exact_staleness: :class:`datetime.timedelta`
-        :param exact_staleness: Execute all reads at a timestamp that is
-                                ``exact_staleness`` old.
-
-        :rtype: :class:`~google.cloud.spanner.snapshot.Snapshot`
-        :returns: a snapshot bound to this session
-        :raises: :exc:`ValueError` if the session has not yet been created.
+        :type kw: dict
+        :param kw:
+            Passed through to
+            :class:`~google.cloud.spanner.snapshot.Snapshot` constructor.
 
         :rtype: :class:`~google.cloud.spanner.database.SnapshotCheckout`
         :returns: new wrapper
         """
-        return SnapshotCheckout(
-            self,
-            read_timestamp=read_timestamp,
-            min_read_timestamp=min_read_timestamp,
-            max_staleness=max_staleness,
-            exact_staleness=exact_staleness,
-        )
+        return SnapshotCheckout(self, **kw)
 
 
 class BatchCheckout(object):
@@ -467,40 +443,20 @@ class SnapshotCheckout(object):
     :type database: :class:`~google.cloud.spannder.database.Database`
     :param database: database to use
 
-    :type read_timestamp: :class:`datetime.datetime`
-    :param read_timestamp: Execute all reads at the given timestamp.
-
-    :type min_read_timestamp: :class:`datetime.datetime`
-    :param min_read_timestamp: Execute all reads at a
-                               timestamp >= ``min_read_timestamp``.
-
-    :type max_staleness: :class:`datetime.timedelta`
-    :param max_staleness: Read data at a
-                          timestamp >= NOW - ``max_staleness`` seconds.
-
-    :type exact_staleness: :class:`datetime.timedelta`
-    :param exact_staleness: Execute all reads at a timestamp that is
-                            ``exact_staleness`` old.
+    :type kw: dict
+    :param kw:
+        Passed through to
+        :class:`~google.cloud.spanner.snapshot.Snapshot` constructor.
     """
-    def __init__(self, database, read_timestamp=None, min_read_timestamp=None,
-                 max_staleness=None, exact_staleness=None):
+    def __init__(self, database, **kw):
         self._database = database
         self._session = None
-        self._read_timestamp = read_timestamp
-        self._min_read_timestamp = min_read_timestamp
-        self._max_staleness = max_staleness
-        self._exact_staleness = exact_staleness
+        self._kw = kw
 
     def __enter__(self):
         """Begin ``with`` block."""
         session = self._session = self._database._pool.get()
-        return Snapshot(
-            session,
-            read_timestamp=self._read_timestamp,
-            min_read_timestamp=self._min_read_timestamp,
-            max_staleness=self._max_staleness,
-            exact_staleness=self._exact_staleness,
-        )
+        return Snapshot(session, **self._kw)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """End ``with`` block."""
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -184,10 +184,11 @@ class Snapshot(_SnapshotBase):
         if len(flagged) > 1:
             raise ValueError("Supply zero or one options.")
 
-        if multi_use and (min_read_timestamp or max_staleness):
-            raise ValueError(
-                "'multi_use' is incompatile with "
-                "'min_read_timestamp' / 'max_staleness'")
+        if multi_use:
+            if min_read_timestamp is not None or max_staleness is not None:
+                raise ValueError(
+                    "'multi_use' is incompatible with "
+                    "'min_read_timestamp' / 'max_staleness'")
 
         self._strong = len(flagged) == 0
         self._read_timestamp = read_timestamp
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -198,6 +198,9 @@ class Snapshot(_SnapshotBase):
 
     def _make_txn_selector(self):
         """Helper for :meth:`read`."""
+        if self._transaction_id is not None:
+            return TransactionSelector(id=self._transaction_id)
+
         if self._read_timestamp:
             key = 'read_timestamp'
             value = _datetime_to_pb_timestamp(self._read_timestamp)
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -225,3 +225,26 @@ class Snapshot(_SnapshotBase):
             return TransactionSelector(begin=options)
         else:
             return TransactionSelector(single_use=options)
+
+    def begin(self):
+        """Begin a transaction on the database.
+
+        :rtype: bytes
+        :returns: the ID for the newly-begun transaction.
+        :raises: ValueError if the transaction is already begun, committed,
+                 or rolled back.
+        """
+        if not self._multi_use:
+            raise ValueError("Cannot call 'begin' single-use snapshots")
+
+        if self._transaction_id is not None:
+            raise ValueError("Transaction already begun")
+
+        database = self._session._database
+        api = database.spanner_api
+        options = _options_with_prefix(database.name)
+        txn_selector = self._make_txn_selector()
+        response = api.begin_transaction(
+            self._session.name, txn_selector.begin, options=options)
+        self._transaction_id = response.id
+        return self._transaction_id
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -167,9 +167,10 @@ class Snapshot(_SnapshotBase):
                             ``exact_staleness`` old.
 
     :type multi_use: :class:`bool`
-    :param multi_use: If true, the first read operation creates a read-only
-                      transaction, used to ensure isolation / consistency
-                      for subsequent read operations.  Incompatible with
+    :param multi_use: If true, multipl :meth:`read` / :meth:`execute_sql`
+                      calls can be performed with the snapshot in the
+                      context of a read-only transaction, used to ensure
+                      isolation / consistency. Incompatible with
                       ``max_staleness`` and ``min_read_timestamp``.
     """
     _transaction_id = None
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -157,9 +157,15 @@ class Snapshot(_SnapshotBase):
     :type exact_staleness: :class:`datetime.timedelta`
     :param exact_staleness: Execute all reads at a timestamp that is
                             ``exact_staleness`` old.
+
+    :type multi_use: :class:`bool`
+    :param multi_use: If true, the first read operation creates a read-only
+                      transaction, used to ensure isolation / consistency
+                      for subsequent read operations.  Incompatible with
+                      ``max_staleness`` and ``min_read_timestamp``.
     """
     def __init__(self, session, read_timestamp=None, min_read_timestamp=None,
-                 max_staleness=None, exact_staleness=None):
+                 max_staleness=None, exact_staleness=None, multi_use=False):
         super(Snapshot, self).__init__(session)
         opts = [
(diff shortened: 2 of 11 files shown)


spanner/google/cloud/spanner/database.py
```diff
@@ -389,8 +389,7 @@ def batch(self):
         """
         return BatchCheckout(self)
 
-    def snapshot(self, read_timestamp=None, min_read_timestamp=None,
-                 max_staleness=None, exact_staleness=None):
+    def snapshot(self, **kw):
```
dhermes: Once upon a time, Pylint would complain about variables with less than 3 chars.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -168,11 +174,17 @@ def __init__(self, session, read_timestamp=None, min_read_timestamp=None,
         if len(flagged) > 1:
             raise ValueError("Supply zero or one options.")
 
+        if multi_use and (min_read_timestamp or max_staleness):
```
dhermes: Please use an explicit `None` check for `min_read_timestamp` and `max_staleness`.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -168,11 +174,17 @@ def __init__(self, session, read_timestamp=None, min_read_timestamp=None,
...
+            raise ValueError(
+                "'multi_use' is incompatile with "
```
dhermes: s/incompatile/incompatible

spanner/google/cloud/spanner/streamed.py
```diff
@@ -130,6 +134,7 @@ def consume_next(self):
         self._resume_token = response.resume_token
 
         if self._metadata is None:  # first response
+            # XXX: copy implicit txn ID to snapshot, if present.
```
dhermes: Please remove the `XXX` (either delete or file an issue or change it to a `NOTE:`)

spanner/tests/unit/test_snapshot.py
```diff
@@ -274,6 +292,12 @@ def test_execute_sql_normal(self):
         self.assertEqual(options.kwargs['metadata'],
                          [('google-cloud-resource-prefix', database.name)])
 
+    def test_execute_sql_wo_mulit_use(self):
```
dhermes: s/multi/multi

spanner/tests/unit/test_streamed.py
```diff
@@ -99,15 +99,52 @@ def _makeListValue(values=(), value_pbs=None):
             return Value(list_value=ListValue(values=value_pbs))
         return Value(list_value=_make_list_value_pb(values))
 
+    @staticmethod
+    def _makeResultSetMetadata(fields=(), transaction_id=None):
```
dhermes: 1. Please use PEP8 method names: `_make_result_set_metadata`
1. Thank your for making this a `@staticmethod`

spanner/tests/unit/test_streamed.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import mock
```
dhermes: This is not in the stdlib, instead:

```python
import unittest

import mock
```

spanner/google/cloud/spanner/database.py
tseaver: `**kw` is *really* idiomatic Python.

```bash
$ find /opt/Python-2.7.13/lib/python2.7/ -name "*.py" | xargs grep "\*\*kw\>" | wc -l
794
```

spanner/google/cloud/spanner/snapshot.py
tseaver: We just checked above that line that exactly one of the other options is not None.

spanner/google/cloud/spanner/database.py
jonparrott:
```
$ find ~/.pyenv/versions/3.6.0/lib/python3.6/ -name "*.py" | xargs grep "\*\*kwargs\>" | wc -l
1671
```

Plus, pylint probably doesn't like `kw`.

spanner/google/cloud/spanner/snapshot.py
dhermes: :+1:

spanner/google/cloud/spanner/database.py
tseaver: @jonparrott I don't really care if pylint doesn't like it:  it find that tool over-opinionated.  `**kw` is *still* idiomatic Python:

- The 3.6.0 library still uses it in 694 cases,
- Our own code uses it 430 times.  Asking for changes here (rather than making a sweeping change through the codebase) isn't the right way to make such changes.

dhermes: @tseaver It's fine to leave as-is.

FWIW "our own code uses it 430 times" isn't much of an argument if you are the author of all 430 instances.

jonparrott: We also use `kwargs` 1065 times in this codebase. Also, we *do use pylint* (even though it's temporarily in warning-only mode). Our [styleguide](https://google.github.io/styleguide/pyguide.html) doesn't explicit call this out, but the one example that uses key-word args uses `kwargs`. Why add more to the fire? Please just use `kwargs`, It's such a minor thing. We can address the others when we decide to re-enable pylint in error mode.

spanner/google/cloud/spanner/snapshot.py
tseaver: ea6b8562e27779cad27d8228f1390f2ffcfa6bf9

spanner/google/cloud/spanner/snapshot.py
tseaver: ea6b8562e27779cad27d8228f1390f2ffcfa6bf9

spanner/tests/unit/test_streamed.py
tseaver: ea6b8562e27779cad27d8228f1390f2ffcfa6bf9

spanner/tests/unit/test_streamed.py
tseaver: dad60608fee7383404fe8502431adfd7ada323e3

spanner/google/cloud/spanner/database.py
tseaver: @jonparrott I see no value in "human linting" a style point which is completely debatable, where the Python community / stdlib is also divided.  If / when we re-enable `pylint`, and it *does* complain, then we fix *all* points in a single "hygeine" PR.  Until then, it is *de gustibus no disputandum est.*

spanner/tests/system/test_system.py
```diff
@@ -574,6 +574,58 @@ def test_snapshot_read_w_various_staleness(self):
...
+
+        with self._db.batch() as batch:
+            batch.delete(self.TABLE, self.ALL)
+
+        after = list(read_ts.read(self.TABLE, self.COLUMNS, self.ALL))
+        self._check_row_data(after, all_data_rows)
+
+    def test_multiuse_snapshot_read_isolation_exact_staleness(self):
+        import time
+        from datetime import timedelta
```
dhermes: Can you move stdlib imports to module-level?

tseaver:
> Can you stop adding system tests to this PR? They keep coming in "after review"

I haven't added any:  I had to rebase to fix the conflicts with your `assertIs` PR.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -81,7 +83,10 @@ def read(self, table, columns, keyset, index='', limit=0,
...
             resume_token=resume_token, options=options)
 
-        return StreamedResultSet(iterator)
+        if self._multi_use:
+            return StreamedResultSet(iterator, source=self)
+        else:
+            return StreamedResultSet(iterator)
 
     def execute_sql(self, sql, params=None, param_types=None, query_mode=None,
                     resume_token=b''):
```
vkedia: If I call `execute_sql` a second time before the first one has returned the transaction id, I will end up creating two transaction ids which is incorrect. You can fix this by either:
- Blocking sending out the second call till you have the transaction id from the first call. But the issue with that is what if the first call for some reason takes a long time to return while the second call would have returned faster.
- Explicitly calling BeginTransaction and blocking sending out any calls till begin transaction has returned. This is the approach we took in Java client to get around the first issue.
In any case you need to block till you have a transaction id.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -157,9 +165,17 @@ class Snapshot(_SnapshotBase):
     :type exact_staleness: :class:`datetime.timedelta`
     :param exact_staleness: Execute all reads at a timestamp that is
                             ``exact_staleness`` old.
+
+    :type multi_use: :class:`bool`
+    :param multi_use: If true, the first read operation creates a read-only
```
vkedia: This is an implementation detail that its the first read which creates the transaction and need not be mentioned in the doc. You can just say
```
If true, multiple read/queries can be done with this Snapshot. They are all guaranteed to execute at the same timestamp.
```

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -128,7 +133,10 @@ def execute_sql(self, sql, params=None, param_types=None, query_mode=None,
             transaction=transaction, params=params_pb, param_types=param_types,
             query_mode=query_mode, resume_token=resume_token, options=options)
 
-        return StreamedResultSet(iterator)
+        if self._multi_use:
+            return StreamedResultSet(iterator, source=self)
+        else:
+            return StreamedResultSet(iterator)
```
vkedia: I think you should throw an error if read/query is called multiple times on the same Snapshot object which is not multi_use. Otherwise people might mistakenly assume that all those reads are executing at the same snapshot while they are not.

spanner/google/cloud/spanner/snapshot.py
vkedia: Thought about this a bit more. I realized that since python client uses the synchronous grpc api, solution I proposed will not work. But the issue still remains. For example I can do this
```
snapshot = Snapshot(multi_use=true)
iter1 = snapshot.read(....)
iter2 = snapshot.read(....)
rows1 = list(iter1)
rows2 = list(iter2)
```

The two reads above will use different transaction ids. I think the easiest solution will be to add an explicit `begin` method which calls BeginTransaction. Users will need to first call that before they can do any reads/queries in case of multi use.

vkedia:
The documentation on this [page](https://googlecloudplatform.github.io/google-cloud-python/latest/spanner/snapshot-usage.html) suggests that snapshot already allows you to do multiple reads at a consistent snapshot.
This is what it says:
```
A Snapshot represents a read-only transaction: when multiple read operations are peformed via a Snapshot, the results are consistent as of a particular point in time.
```
In light of this PR that is clearly incorrect.

@jonparrott  We also claim the same thing in our [sample code])
https://cloud.google.com/spanner/docs/getting-started/python/#retrieve_data_using_read-only_transactions). That too is incorrect.

@lukesneeringer @bjwatson Can we please get this merged asap. This is a severe bug.

spanner/google/cloud/spanner/database.py
lukesneeringer: Folks, enough with the bikeshedding on this. It is not important enough to justify the amount of ink being spilled on it.

vkedia:
@lukesneeringer I believe my comments need to be addressed before merging this. Specifically
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3615#discussion_r128401471

lukesneeringer:
Ah -- confirmed. Thanks @vkedia. (@tsearver His proposed solution seems reasonable to me. Thoughts?)

vkedia:
@tseaver Any updates on this?

tseaver:
I can see adding an explicit `Snapshot.begin` to capture a transaction ID.

In addition, I had thought of having a `_transaction_pending` flag on the multi-use snapshot, which would cause subsequent `read` / `execute_sql` requests to raise an exception.  This would smooth the "garden path" usage (i.e., reading / querying with fetch, rather than interleaving).

tseaver:
To implement the [single-use requested guard](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3615#discussion_r128397368), the snapshot needs to maintain a request counter, which we can use to detect the ["interleaved" case](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3615#discussion_r128401471).

vkedia:
That sounds good. We can have an explicit `begin` which users can chose to call to get a transaction id and then they can interleave reads. If they do not call `begin` then the first `read` or `query` will create the transaction and if they try to interleave before it has returned they will get back an error.
So essentially users can interleave reads only after there is a transaction id which can be got by either calling `begin` or by the first `read`

vkedia:
Actually why do we need `Snapshot` to support `single_use`? For single use transactions `database.read` and `database.query` method should be sufficient. We should extend those methods to also take timestamp bound specific options.

tseaver:
`Database.read` and `Database.execute_sql` create `Snapshot` instances under-the-covers (via `Session.read` / `Session.execute_sql`).  I just opened #3659 to track adding the read isolation parameters to those methods.

tseaver:
@vkedia I've just updated the system tests to pass `multi_use=True` where needed (I also fixed a new bug where `Transaction` was not implicitly multi-use for reads).  I am, however, seeing system failures on the *third* invocation using multi-use snapshots:

```python
________________ TestSessionAPI.test_execute_sql_w_query_param _________________
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/tests/system/test_system.py", line 960, in test_execute_sql_w_query_param
    expected=[(19,), (99,)],
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/tests/system/test_system.py", line 888, in _check_sql_results
    sql, params=params, param_types=param_types))
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/spanner/streamed.py", line 166, in __iter__
    self.consume_next()  # raises StopIteration
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/spanner/streamed.py", line 132, in consume_next
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/grpc/_channel.py", line 363, in __next__
    return self._next()
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/grpc/_channel.py", line 357, in _next
    raise self
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, Transaction was started in a different session.)>
______________________ TestSessionAPI.test_read_w_ranges _______________________
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/tests/system/test_system.py", line 846, in test_read_w_ranges
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/spanner/streamed.py", line 166, in __iter__
    self.consume_next()  # raises StopIteration
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/spanner/streamed.py", line 132, in consume_next
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/grpc/_channel.py", line 363, in __next__
    return self._next()
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/spanner/.nox/sys-3-6/lib/python3.6/site-packages/grpc/_channel.py", line 357, in _next
    raise self
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, Transaction was started in a different session.)>
```

I tried putting an explicit `snapshot.begin()` before the repeated `read` / `execute_sql` calls, but that didnt' change anything.  Would the back-end be somehow be passing back a new transaction ID for the second `read` / `execute_sql` request (which *does* succeed)?

tseaver:
Looks like the back-end passes back an *empty* transaction ID for read requests after the first one.  0966806 keeps us from clearing it in that case.

vkedia:
Thats right. Transaction field is only set if the read or query started a new transaction.
https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ResultSetMetadata

tseaver:
The new system tests are [passing](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2529).

dhermes:
@tseaver What are you looking for here in terms of review?

tseaver:
@dhermes the commits which are new since @lukesneeringer gave an LGTM on Thursday (the rest are just rebase to fix conflicts).  a5219a58897eff987a70b017efc7cf8f719f3b29 is the hash of that rebased commit, so the diff would be: https://github.com/GoogleCloudPlatform/google-cloud-python/compare/a5219a58897eff987a70b017efc7cf8f719f3b29...spanner-multi_use_snapshot

tseaver:
@vkedia, @dhermes any issues remaining?

dhermes:
LGTM

bjwatson:
@tseaver I just saw this. I think that @vkedia will want to review this more when he returns from vacation on Monday. I guess if he finds anything else, it can be addressed in a separate PR.

Let's remain Alpha until next week. (FYI @lukesneeringer)

bjwatson:
@vkedia Do you plan to finish reviewing this post-merge?

PR #3616, opened 2017-07-17
Add base future package to google.cloud

jonparrott:


core/google/cloud/_helpers.py
```diff
@@ -379,6 +379,29 @@ def _bytes_to_unicode(value):
         raise ValueError('%r could not be converted to unicode' % (value,))
 
 
+def _from_any_pb(pb_type, any_pb):
```
dhermes: [`google.cloud.operation._from_any`][1] already exists, can you "unify"?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/59fd1e4ff50827a868986deee938a26b5d50dc0f/core/google/cloud/operation.py#L64-L81

core/google/cloud/_helpers.py
```diff
@@ -379,6 +379,29 @@ def _bytes_to_unicode(value):
...
+    """Converts an Any protobuf to the specified message type
+
+    Args:
+        pb_type (type): the type of the message that any_pb stores an instance
+            of.
+        any_pb (google.protobuf.any_pb2.Any): the object to be converted.
+
+    Returns:
+        pb_type: An instance of the pb_type message.
```
dhermes: I highly doubt Sphinx would be cool with `pb_type:` here (it's moot because this function will never get rendered in docs, so don't waste your time)

core/google/cloud/future/__init__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2017, Google Inc.
```
dhermes: Do we really need a whole package? Won't a module suffice?

core/google/cloud/future/base.py
```diff
@@ -0,0 +1,173 @@
...
+    # pylint: disable=missing-docstring, invalid-name
+    # We inherit the interfaces here from concurrent.futures.
+
+    """Future interface.
+
+    This interface is based on :class:`concurrent.futures.Future`.
+    """
+
+    @abc.abstractmethod
+    def cancel(self):  # pragma: NO COVER
```
dhermes: Too much to ask that you ditch the `NO COVER` and just write some copy-pasta unit tests (or I'll do it)?

core/google/cloud/future/base.py
```diff
@@ -0,0 +1,173 @@
...
+import abc
+
+import six
+
+from google.cloud.future import _helpers
+
+
+@six.add_metaclass(abc.ABCMeta)
+class Future(object):
+    # pylint: disable=missing-docstring, invalid-name
```
dhermes: 1. There is no corresponding `enable`?
1. What is the `invalid-name`?

core/tests/unit/future/test__helpers.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2017, Google Inc.
```
dhermes: This LICENSE seems different than the one we already have

jonparrott:
Towards #3617

core/google/cloud/_helpers.py
jonparrott: Baby steps, I'll likely remove that whole module.

core/google/cloud/future/__init__.py
jonparrott: Yes, my next PR will drop an `operation.py` module in here for `google.cloud.future.operation` unless you would prefer me to replace the current `google.cloud.operation` module.

core/google/cloud/future/base.py
jonparrott: 1. It's block-level.
2. it's `fn` in `add_done_callback`, I've moved this disable down there to be more specific.

core/google/cloud/future/base.py
jonparrott: Yes. Can you tell me what value those tests add? What benefit do we as maintainers get in the future by adding trivial tests to assert the abstractness of this class? Not trying to be dismissive, I just have yet to hear a good argument for this- if you can come up with anything I'm likely to bend. :)

core/tests/unit/future/test__helpers.py
jonparrott: Fixed.

core/google/cloud/future/base.py
tseaver: I'd rather just have our `.coveragerc` [skip lines  with `raise NotImplementedError`](https://github.com/zopefoundation/zope.interface/blob/master/.coveragerc#L10).

core/google/cloud/future/__init__.py
tseaver: Do you intend to make the `google.cloud.operation` module a backward-compatibilitiy-only shim?

dhermes:
LGTM

core/google/cloud/future/base.py
jonparrott: Good plan, was just discussing that with @dhermes.

core/google/cloud/future/__init__.py
jonparrott: No, the interfaces an behavior are significantly different. Unless @lukesneeringer feels strongly, I will not make a backwards-compatible shim.

core/google/cloud/future/base.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@six.add_metaclass(abc.ABCMeta)
+class Future(object):
+    # pylint: disable=missing-docstring
+    # We inherit the interfaces here from concurrent.futures.
+
+    """Future interface.
+
+    This interface is based on :class:`concurrent.futures.Future`.
```
duggelz: Is Python 2 compatibility the reason you don't directly use concurrent.futures.Future, or are there other reasons?

jonparrott: `concurrent.futures.Future` isn't meant to be subclassed.

duggelz: Huh.  I learned something.  Executor is abstract so I just assumed Future would be too.

GRPC has their own which I assume you already looked at: https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/__init__.py#L35

jonparrott: Yep this is largely based on theirs and asyncio's.

Issue #3548, opened 2017-06-26
BigQuery: add `is_nullable` method to check for `NULLABLE` mode


tswast:
In the DB-API implementation I wrote some code like

    null_ok=field.mode == 'NULLABLE'

Could be nice to have a helper function that does this.

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2921/files/7d59e97172d16bc18601cce0845000b32e6be9a6#r124069049

PR #3620, opened 2017-07-18
BigQuery: Add `is_nullable` method to check for `NULLABLE` mode

sonlac:
Resolves: #3548


tseaver:
@sonlac Thank you very much for the patch!  Could you please add tests for the `is_nullable` property in `tests/unit/test_schema.py`?

sonlac:
@tseaver Done. Thank you for reviewing.

Issue #3617, opened 2017-07-17
Spanner, BigTable, and BigQuery operations should use the Futures interface


jonparrott:

bjwatson:
@jonparrott It sounds like this is Beta blocking for Spanner and the other APIs, correct? IIUC, it affects the surface for LRO calls.

jonparrott:
Yes but this is also fixed.

PR #3626, opened 2017-07-18
Add Future interface to BigQuery jobs

### LINKED ISSUE #3556: Bigquery Jobs should match the Futures interface
Presently, bigquery uses a custom [`Job`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L85) class (and subclasses) to deal with long-running operations such as queries and data import.

The Job class causes the proliferation of code like this around all of our samples and user code:

```python
def wait_for_job(job):
    while True:
        job.reload()  # Refreshes the state via a GET request.
        if job.state == 'DONE':
            if job.error_result:
                raise RuntimeError(job.errors)
            return
        time.sleep(1)
```

Which in terms means our samples require a level on indirection to accomplish simple tasks:

```python
client = bigquery.Client()
query_job = client.run_async_query(str(uuid.uuid4()), query)
query_job.use_legacy_sql = False
query_job.begin()

wait_for_job(query_job)

rows = query_job.results().fetch_data(max_results=10)
for row in rows:
    print(row)
```

However, our gax-based clients share a common "Long Running Operation" strategy that surfaces these types of on-going operations as a class that implements the [`concurrent.futures.Future`](https://docs.python.org/3/library/concurrent.futures.html#future-objects) interface. This is implemented as [`gax._OperationFuture`](https://github.com/googleapis/gax-python/blob/3d48d1fe3cb149f875328630267ce148565cbb35/google/gax/__init__.py#L549). bigquery unfortunately can not use this as it's an http-only API and gax can only currently be used for gRPC APIs.

We should make the `_BaseJob` class and its subclasses closely conform to the `Future` interface so that the usage is similar to other API client and so that we can simplify usage to:

```python
client = bigquery.Client()
query_job = client.run_async_query(str(uuid.uuid4()), query)
query_job.use_legacy_sql = False
query_job.begin()

# Wait for the job to finish
query_result = query.result()

rows = query_result.fetch_data(max_results=10)
for row in rows:
    print(row)

```

jonparrott:
Resolves #3556 
Towards #3617

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -14,8 +14,11 @@
 
 """Define API Jobs."""
 
+import threading
+
 import six
 
+from google.cloud import exceptions
 from google.cloud.exceptions import NotFound
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud.bigquery.dataset import Dataset
@@ -27,6 +30,7 @@ from google.cloud.bigquery._helpers import QueryParametersProperty
 from google.cloud.bigquery._helpers import UDFResourcesProperty
 from google.cloud.bigquery._helpers import _EnumProperty
 from google.cloud.bigquery._helpers import _TypedProperty
+import google.cloud.future.base
 
 
 class Compression(_EnumProperty):
@@ -82,7 +86,7 @@ class WriteDisposition(_EnumProperty):
     ALLOWED = (WRITE_APPEND, WRITE_TRUNCATE, WRITE_EMPTY)
 
 
-class _BaseJob(object):
+class _BaseJob(google.cloud.future.base.PollingFuture):
     """Base class for jobs.
 
     :type client: :class:`google.cloud.bigquery.client.Client`
@@ -90,6 +94,7 @@ class _BaseJob(object):
                    for the dataset (which requires a project).
     """
     def __init__(self, client):
+        super(_BaseJob, self).__init__()
         self._client = client
         self._properties = {}
 
@@ -131,6 +136,8 @@ class _AsyncJob(_BaseJob):
     def __init__(self, name, client):
         super(_AsyncJob, self).__init__(client)
         self.name = name
+        self._result_set = False
+        self._completion_lock = threading.Lock()
 
     @property
     def job_type(self):
@@ -273,6 +280,9 @@ class _AsyncJob(_BaseJob):
         self._properties.clear()
         self._properties.update(cleaned)
 
+        # For Future interface
+        self._set_future_result()
+
     @classmethod
     def _get_resource_config(cls, resource):
         """Helper for :meth:`from_api_repr`
@@ -345,7 +355,7 @@ class _AsyncJob(_BaseJob):
             return True
 
     def reload(self, client=None):
-        """API call:  refresh job properties via a GET request
+        """API call:  refresh job properties via a GET request.
 
         See
         https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get
@@ -371,12 +381,56 @@ class _AsyncJob(_BaseJob):
                       ``NoneType``
         :param client: the client to use.  If not passed, falls back to the
                        ``client`` stored on the current dataset.
+
+        :rtype: bool
+        :returns: Boolean indicating that the cancel request was sent.
         """
         client = self._require_client(client)
 
         api_response = client._connection.api_request(
             method='POST', path='%s/cancel' % (self.path,))
         self._set_properties(api_response['job'])
+        return True
+
+    # The following methods implement the PollingFuture interface. Note that
+    # the methods above are from the pre-Future interface and are left for
+    # compatibility. The only "overloaded" method is :meth:`cancel`, which
+    # satisfies both interfaces.
+
+    def _set_future_result(self):
+        """Set the result or exception from the job if it is complete."""
+        # This must be done in a lock to prevent the polling thread
+        # and main thread from both executing the completion logic
+        # at the same time.
+        with self._completion_lock:
+            # If the operation isn't complete or if the result has already been
+            # set, do not call set_result/set_exception again.
+            # Note: self._result_set is set to True in set_result and
+            # set_exception, in case those methods are invoked directly.
+            if not self.state != 'DONE' or self._result_set:
+                return
+
+            if self.error_result:
+                exception = exceptions.GoogleCloudError(
+                    self.error_result, errors=self.errors)
+                self.set_exception(exception)
+            else:
+                self.set_result(self)
+
+    def done(self):
+        # Do not refresh is the state is already done, as the job will not
+        # change once complete.
+        if self.state != 'DONE':
+            self.reload()
+        return self.state == 'DONE'
+
+    def result(self, timeout=None):
+        if self.state is None:
+            self.begin()
+        return super(self, _AsyncJob).result(timeout=timeout)
+
+    def cancelled(self):
+        return False
 
 
 class _LoadConfiguration(object):
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -15,6 +15,7 @@
 """Define API Jobs."""
 
 import threading
+import warnings
 
 import six
 
@@ -86,17 +87,23 @@ class WriteDisposition(_EnumProperty):
     ALLOWED = (WRITE_APPEND, WRITE_TRUNCATE, WRITE_EMPTY)
 
 
-class _BaseJob(google.cloud.future.base.PollingFuture):
-    """Base class for jobs.
+class _AsyncJob(google.cloud.future.base.PollingFuture):
+    """Base class for asynchronous jobs.
+
+    :type name: str
+    :param name: the name of the job
 
     :type client: :class:`google.cloud.bigquery.client.Client`
     :param client: A client which holds credentials and project configuration
                    for the dataset (which requires a project).
     """
-    def __init__(self, client):
-        super(_BaseJob, self).__init__()
+    def __init__(self, name, client):
+        super(_AsyncJob, self).__init__()
+        self.name = name
         self._client = client
         self._properties = {}
+        self._result_set = False
+        self._completion_lock = threading.Lock()
 
     @property
     def project(self):
@@ -121,23 +128,6 @@ class _BaseJob(google.cloud.future.base.PollingFuture):
         if client is None:
             client = self._client
         return client
-
-
-class _AsyncJob(_BaseJob):
-    """Base class for asynchronous jobs.
-
-    :type name: str
-    :param name: the name of the job
-
-    :type client: :class:`google.cloud.bigquery.client.Client`
-    :param client: A client which holds credentials and project configuration
-                   for the dataset (which requires a project).
-    """
-    def __init__(self, name, client):
-        super(_AsyncJob, self).__init__(client)
-        self.name = name
-        self._result_set = False
-        self._completion_lock = threading.Lock()
 
     @property
     def job_type(self):
@@ -407,10 +397,10 @@ class _AsyncJob(_BaseJob):
             # set, do not call set_result/set_exception again.
             # Note: self._result_set is set to True in set_result and
             # set_exception, in case those methods are invoked directly.
-            if not self.state != 'DONE' or self._result_set:
+            if self.state != 'DONE' or self._result_set:
                 return
 
-            if self.error_result:
+            if self.error_result is not None:
                 exception = exceptions.GoogleCloudError(
                     self.error_result, errors=self.errors)
                 self.set_exception(exception)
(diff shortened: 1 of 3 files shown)


jonparrott:
@tswast this is the usage Im going for: https://gist.github.com/jonparrott/cb0ec50b8c70dfb32693fae021659baa

tswast:
I don't like

```
query_job.result()
rows = query_job.results().fetch_data(max_results=10)
```

I'd prefer if it was

```
rows = query_job.result()
```

Also, the query job needs to be calling [getQueryResults](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) and checking for `jobComplete` in the response rather than polling the job state. The reason for this is that [getQueryResults](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) will wait for the job to complete or timeout (whichever comes first), whereas [get Job](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get) returns immediately.

core/google/cloud/future/base.py
```diff
@@ -89,15 +93,52 @@ def __init__(self):
...
         """
-        # pylint: disable=missing-raises
-        raise NotImplementedError()
+        if self._result_set:
+            return
+
+        retry_on = tenacity.retry_if_result(
+            functools.partial(operator.is_not, True))
+        # Use exponential backoff with jitter.
+        wait_on = (
```
tswast: How long will this wait between calls? I would think that a polling interval should be passed in, in addition to the timeout.

jonparrott: We trying to stick with the futures interface which just allows you to pass in the timeout. Also, if you use `add_done_callback` you will not have the ability to pass in a timeout or retry interval.

jonparrott:
> I'd prefer if it was
> rows = query_job.result()

That seems reasonable as long as we don't lose data, but wouldn't that require switching to core iterator (#2840)? Could we accept the non-breaking behavior of this PR as-is for now and do another PR to switch the return value of `result` after #2840 is implemented?

> Also, the query job needs to be calling getQueryResults and checking for jobComplete in the response rather than polling the job state. The reason for this is that getQueryResults will wait for the job to complete or timeout (whichever comes first), whereas get Job returns immediately.

Why is that preferable? Doing a long-lived HTTP request is quite different from any of the other behavior in this library. Also, would you be okay with addressing this in a follow-up PR rather than in this one?

tswast:
#2840 is already implemented.

`getQueryResults` is preferable because the job polling method adds up to `polling_interval` milliseconds of latency to get the query results compared to the long-lived request which will return results as soon as they are available.

tswast:
I'm okay with doing the `getQueryResults` in a subsequent PR.

jonparrott:
@tswast Cool, are you good with this PR as-is then?

tswast:
Could we make the change where `QueryJob.result()` returns `QueryResults` in this PR?

tswast:
> Could we make the change where QueryJob.result() returns QueryResults in this PR?

If not, I'll propose we make that as a breaking change when changing the polling method.

jonparrott:
> Could we make the change where QueryJob.result() returns QueryResults in this PR?

Yep. Done. Can you try this out and let me know how broken it is?

This should work:

```python
def async_query(query):
    client = bigquery.Client()
    query_job = client.run_async_query(str(uuid.uuid4()), query)
    query_job.use_legacy_sql = False

    rows = query_job.result().fetch_data(max_results=10)
    for row in rows:
        print(row)
```

jonparrott:
@tswast verified that this seems to work:

```python
>>> from google.cloud import bigquery
>>> import uuid
>>>
>>> client = bigquery.Client()
>>> query = 'SELECT 1'
>>> query_job = client.run_async_query(str(uuid.uuid4()), query)
>>> query_job.use_legacy_sql = False
>>>
>>> rows = query_job.result().fetch_data(max_results=10)
>>> for row in rows:
...     print(row)
...
(1,)
```

bigquery/google/cloud/bigquery/job.py
```diff
@@ -27,6 +31,7 @@
 from google.cloud.bigquery._helpers import UDFResourcesProperty
 from google.cloud.bigquery._helpers import _EnumProperty
 from google.cloud.bigquery._helpers import _TypedProperty
+import google.cloud.future.base
```
dhermes: This should go up with the other group (the "1st-party-almost" group)

bigquery/google/cloud/bigquery/job.py
```diff
@@ -371,12 +371,83 @@ def cancel(self, client=None):
...
+
+        :rtype: bool
+        :returns: Boolean indicating that the cancel request was sent.
         """
         client = self._require_client(client)
 
         api_response = client._connection.api_request(
             method='POST', path='%s/cancel' % (self.path,))
         self._set_properties(api_response['job'])
+        return True
```
dhermes: Why?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -371,12 +371,83 @@ def cancel(self, client=None):
...
+        """Set the result or exception from the job if it is complete."""
+        # This must be done in a lock to prevent the polling thread
+        # and main thread from both executing the completion logic
+        # at the same time.
+        with self._completion_lock:
+            # If the operation isn't complete or if the result has already been
+            # set, do not call set_result/set_exception again.
+            # Note: self._result_set is set to True in set_result and
+            # set_exception, in case those methods are invoked directly.
+            if self.state != 'DONE' or self._result_set:
```
dhermes: Where does `DONE` come from? Make it a constant?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -371,12 +371,83 @@ def cancel(self, client=None):
...
+                return
+
+            if self.error_result is not None:
+                exception = exceptions.GoogleCloudError(
+                    self.error_result, errors=self.errors)
```
dhermes: What is `error_result`? We have `google.cloud.exceptions.make_exception`

bigquery/google/cloud/bigquery/job.py
```diff
@@ -371,12 +371,83 @@ def cancel(self, client=None):
...
+
+    def done(self):
+        """Refresh the job and checks if it is complete.
+
+        :rtype: bool
+        :returns: True if the job is complete, False otherwise.
+        """
+        # Do not refresh is the state is already done, as the job will not
+        # change once complete.
+        if self.state != 'DONE':
```
dhermes: OK definitely make `DONE` a constant

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1127,11 +1198,44 @@ def from_api_repr(cls, resource, client):
...
         """Construct a QueryResults instance, bound to this job.
 
         :rtype: :class:`~google.cloud.bigquery.query.QueryResults`
         :returns: results instance
         """
         from google.cloud.bigquery.query import QueryResults
         return QueryResults.from_query_job(self)
+
+    def results(self):
+        """DEPRECATED.
```
dhermes: Can you add a GitHub Issue and a Google Calendar event marking "when" it should be deprecated?

bigquery/google/cloud/bigquery/job.py
jonparrott: see https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs

It's

```python
debugInfo: str
location: str
message: str
reason: str
```

It does not match HTTP error codes.

jonparrott: Seems we can map them using https://cloud.google.com/bigquery/troubleshooting-errors

bigquery/google/cloud/bigquery/job.py
jonparrott: Added a comment explaining (It's to satisfy the future interface)

bigquery/google/cloud/bigquery/job.py
jonparrott: Done. (pun intended)

bigquery/google/cloud/bigquery/job.py
jonparrott: (done)

bigquery/google/cloud/bigquery/job.py
jonparrott: #3652

bigquery/google/cloud/bigquery/job.py
```diff
@@ -33,6 +35,43 @@
...
+    'quotaExceeded': http_client.FORBIDDEN,
+    'rateLimitExceeded': http_client.FORBIDDEN,
+    'resourceInUse': http_client.BAD_REQUEST,
+    'resourcesExceeded': http_client.BAD_REQUEST,
+    'responseTooLarge': http_client.FORBIDDEN,
+    'stopped': http_client.OK,
+    'tableUnavailable': http_client.BAD_REQUEST,
+}
+
+_FakeResponse = collections.namedtuple('_FakeResponse', ['status'])
```
dhermes: You would! :grinning:

bigquery/google/cloud/bigquery/job.py
```diff
@@ -33,6 +35,43 @@
...
+
+
+def _error_result_to_exception(error_result):
+    """"""
```
dhermes: I really love this artisanal hand-crafted docstring

bigquery/google/cloud/bigquery/job.py
```diff
@@ -33,6 +35,43 @@
...
+    reason = error_result.get('reason')
+    status_code = _ERROR_REASON_TO_EXCEPTION.get(
+        reason, http_client.INTERNAL_SERVER_ERROR)
+    # make_exception expects an httplib2 response object.
+    fake_response = _FakeResponse(status=status_code)
+    return exceptions.make_exception(
+        fake_response, b'', error_info=error_result, use_json=False)
```
dhermes: `content` should actually be `error_result` (not `b''`)

bigquery/google/cloud/bigquery/job.py
jonparrott: Done.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -447,7 +487,8 @@ def cancelled(self):
         :rtype: bool
         :returns: False
         """
-        return False
+        return (self.error_result is not None
+                and self.error_result.get('reason') == 'stopped')
```
dhermes: 1. Worth caching this output?
2. Where does `stopped` come from?

bigquery/tests/unit/test_job.py
```diff
@@ -15,9 +15,30 @@
 import copy
 import warnings
 
+from six.moves import http_client
 import unittest
 
 
+class TestErrorResultToException(unittest.TestCase):
```
dhermes: Preferred name: `Test__error_result_to_exception` (your call)

bigquery/google/cloud/bigquery/job.py
jonparrott: Do I need to pass in the `error_info` in that case?

bigquery/google/cloud/bigquery/job.py
jonparrott: Oops. Fixed in incoming commit.

bigquery/google/cloud/bigquery/job.py
jonparrott: I thought you'd love this.

bigquery/google/cloud/bigquery/job.py
jonparrott: 1. No.
2. It's from the list of reasons above. I meant to add a comment there. Done in incoming commit.

bigquery/tests/unit/test_job.py
jonparrott: Will do.

bigquery/google/cloud/bigquery/job.py
dhermes: `error_info` is [like a "post-script"][1] for the actual error message:

```python
    if error_info is not None:
        message += ' (%s)' % (error_info,)
```

i.e. it's not what you think it is

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f8758dfda54c79a7c712e73348fc37a4c349c212/core/google/cloud/exceptions.py#L227-L228

jonparrott: That doesn't really help me understand what I *should* do here. ISTM that according to the test this is doing what I expect it to do (raise an error with the error result as the message).

I'm happy to change it, but I don't yet understand what I should change.

dhermes: Can you give me a REPL example of what it does do?

jonparrott:
Thanks @dhermes and @tswast. This is a huge jump in developer experience for bigquery. 🍰

PR #3631, opened 2017-07-19
Fixing references to "dead" docs links.

dhermes:
Done via:

```
$ git grep -l 'google-cloud-auth.html' | \
> xargs sed -i s/'google-cloud-auth.html'/'core\/auth.html'/g
$ git grep -l 'http\:\/\/google-cloud-python.readthedocs.io' | \
> xargs sed -i s/'http\:\/\/google-cloud-python.readthedocs.io'/\
> 'https\:\/\/google-cloud-python.readthedocs.io'/g
```

Fixes #3531.

PR #3634, opened 2017-07-19
update documentation link

merla18:
due to recent change of doc URL


dhermes:
@merla18 Thanks for sending but this is covered by #3631.

PR #3637, opened 2017-07-20
Pubsub subscriber

lukesneeringer:
Uses #3463 as a base.

This PR implements the Pub/Sub subscriber, which by default uses threading.

The following is yet to be implemented:

   * [x] Automatic pause/resume based on flow control settings.
   * [x] Tests of any kind (working on them in parallel with the Pub/Sub team's review)
   * [x] API Documentation


pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,273 @@
...
+        """
+        return self._subscription
+
+    def ack(self, ack_id):
+        """Acknowledge the message corresponding to the given ack_id.
+
+        Args:
+            ack_id (str): The ack ID.
+        """
+        request = types.StreamingPullRequest(ack_ids=[ack_id])
```
mdietz94: We recommend batching acks/mod ack deadlines together and sending them every second or so. Your batching of lease extensions is right, but I would batch these acks and the nacks below and send them periodically instead of sending them one-at-a-time. If you are concerned about the lease expiring, maybe it would be better to not 'release' them until after the ack returns.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,273 @@
...
+    def call_rpc(self, request_generator):
+        """Invoke the Pub/Sub streaming pull RPC.
+
+        Args:
+            request_generator (Generator): A generator that yields requests,
+                and blocks if there are no outstanding requests (until such
+                time as there are).
+        """
+        return self._client.api.streaming_pull(request_generator,
+            options=gax.CallOptions(timeout=600),
```
mdietz94: Why 600? The StreamingPull connection will be good for a half of an hour, and if this is a timeout in case of an inability to connect this seems too long.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,273 @@
...
+        # We determine the appropriate period of time based on a random
+        # period between 0 seconds and 90% of the lease. This use of
+        # jitter (http://bit.ly/2s2ekL7) helps decrease contention in cases
+        # where there are many clients.
+        snooze = random.uniform(0.0, p99 * 0.9)
+        logger.debug('Snoozing lease management for %f seconds.' % snooze)
+        time.sleep(snooze)
+        self.maintain_leases()
+
+    def modify_ack_deadline(self, ack_id, seconds):
```
mdietz94: I would queue these and batch them with ack, as described above.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: Thanks -- this was some test code that I forgot to get rid of.

pubsub/.flake8
```diff
@@ -4,3 +4,9 @@ exclude =
   .git,
   *.pyc,
   conf.py
+
+ignore =
```
jonparrott: I would *very* strongly prefer that we did not add these exceptions. No other project I maintain has needed to do this.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: The Ruby and Node folks both did this, but I was a little curious as to why. It totally makes sense in REST, but the stream is already open? What value do we gain from doing this?

I do not mind doing it; I just want to understand the reason since this requires an additional worker thread (and eventually a worker thread per open stream), which is somewhat expensive.

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -0,0 +1,100 @@
...
+
+    This creates an object that is capable of subscribing to messages.
+    Generally, you can instantiate this client with no arguments, and you
+    get sensible defaults.
+
+    Args:
+        flow_control (~.pubsub_v1.types.FlowControl): The flow control
+            settings. Use this to prevent situations where you are
+            inundated with too many messages at once.
+        policy_class (class): A class that describes how to handle
```
jonparrott: It may be useful to take a policy instance instead of a class. Depending on the concurrency strategy, the developer may need to pass in initialization options to policy.

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -0,0 +1,100 @@
...
+                argument.
+            flow_control (~.pubsub_v1.types.FlowControl): The flow control
+                settings. Use this to prevent situations where you are
+                inundated with too many messages at once.
+
+        Returns:
+            ~.pubsub_v1.subscriber.consumer.base.BaseConsumer: An instance
+                of the defined ``consumer_class`` on the client.
+        """
+        subscr = self._policy_class(self, subscription)
```
jonparrott: `policy` instead of `subscr`? (you're constructing an instance of the policy class, so it leads that the name should match).

pubsub/google/cloud/pubsub_v1/subscriber/helper_threads.py
```diff
@@ -0,0 +1,121 @@
...
+# limitations under the License.
+
+import collections
+import logging
+import threading
+import uuid
+
+import six
+
+__all__ = (
```
jonparrott: This module doesn't need an `__all__`.

pubsub/google/cloud/pubsub_v1/subscriber/histogram.py
```diff
@@ -0,0 +1,155 @@
...
+    The purpose of this class is to store actual ack timing information
+    in order to predict how long to renew leases.
+
+    The default implementation uses the 99th percentile of previous ack
+    times to implicitly lease messages; however, custom
+    :class:`~.pubsub_v1.subscriber.consumer.base.BaseConsumer` subclasses
+    are free to use a different formula.
+
+    The precision of data stored is to the nearest integer. Additionally,
+    values outside the range of ``10 <= x <= 600`` are stored as ``10`` or
```
jonparrott: Does it make sense to make `10` and `600` parameters to this class?

pubsub/google/cloud/pubsub_v1/subscriber/histogram.py
```diff
@@ -0,0 +1,155 @@
...
+    ``600``, since these are the boundaries of leases in the actual API.
+    """
+    def __init__(self, data=None):
+        """Instantiate the histogram.
+
+        Args:
+            data (dict): The data strucure to be used to store the
```
jonparrott: s/dict/Mapping[str, int]

pubsub/google/cloud/pubsub_v1/subscriber/histogram.py
```diff
@@ -0,0 +1,155 @@
...
+        # Add the value to the histogram's data dictionary.
+        self._data.setdefault(value, 0)
+        self._data[value] += 1
+        self._len += 1
+
+    def percentile(self, percent):
+        """Return the value that is the Nth precentile in the histogram.
+
+        Args:
+            percent (int|float): The precentile being sought. The default
```
jonparrott: s/int|float/Union[int, float]

pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Properties:
+        message_id (str): The message ID. In general, you should not need
+            to use this directly.
+        data (bytes): The data in the message.
+        attributes (dict): The attributes sent along with the message.
+        publish_time (datetime): The time that this message was originally
+            published.
+    """
+    def __init__(self, policy, ack_id, message):
```
jonparrott: You can *not* pass the policy across the concurrency boundary and you can not directly invoke policy method across the concurrency boundary. The pattern I wrote in the proof-of-concept of sending callback messages allows you to invoke actions back on the policy without passing it to the message. The only thing you can share across the concurrency boundary is *plain data* and a *user-specified queue*.

Relevant part of proof-of-concept: https://gist.github.com/jonparrott/55723f4c8c7e5b0027f645c89a42110e#file-subscriber-py-L21

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -0,0 +1,118 @@
...
+        self._callback = callback
+        self._consumer.helper_threads.start('callback requests worker',
+            self._request_queue,
+            self._callback_requests,
+        )
+
+        # Actually start consuming messages.
+        self._consumer.start_consuming()
+
+        # Spawn a helper thread that maintains all of the leases for
```
jonparrott: Use `helper_threads` for this?

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
mdietz94: There is some non-zero serialization of the requests, and if you are receiving 10K QPS, I think you'll find that there is a significant amount of time spent on serialization. Keep in mind that gRPC sends these in sequence, so I think there is enough overhead per message for it to be worth batching, not to mention they can then be dealt with in parallel server-side instead of being received one-at-a-time.

If you think that will not be true, in Python (or at all), we can run some benchmarks.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -0,0 +1,118 @@
...
+
+from google.cloud.pubsub_v1.subscriber import helper_threads
+from google.cloud.pubsub_v1.subscriber.policy import base
+from google.cloud.pubsub_v1.subscriber.message import Message
+
+
+logger = logging.getLogger(__name__)
+
+
+class Policy(base.BasePolicy):
```
jonparrott: The proof of concept allowed users to pass in both a thread and process pool executor. I think its important that we retain that functionality.

Relevant part of proof-of-concept: https://gist.github.com/jonparrott/55723f4c8c7e5b0027f645c89a42110e#file-subscriber-py-L42

pubsub/google/cloud/pubsub_v1/subscriber/helper_threads.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2017, Google Inc. All rights reserved.
```
jonparrott: This module should be private.

pubsub/google/cloud/pubsub_v1/subscriber/histogram.py
```diff
@@ -0,0 +1,155 @@
+# Copyright 2017, Google Inc. All rights reserved.
```
jonparrott: This module should be private

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
jonparrott: The serialization cost will be paid on a thread that's dedicated to sending requests to gRPC. The request generator thread does not handle any other responsibilities. My hunch is that it is probably not worth it to do batching, but I'd be interested in benchmarks - remembering that any performance gains will *only* benefit the gRPC request generator thread (which is IO bound not CPU bound).

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: I actually tried, but it did not really work. The helper threads fundamentally are around burning down a single queue, which is not what is happening here.

jonparrott: Thought so, hence the `?`. 

How do you handle termination of this thread?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: I guess I do not really understand why. There seems to be more or less a one to one mapping with regard to what the right one is, and the policy subclasses seem like the place to set this.

jonparrott: Basically, the consumer does all RPCs and bookkeeping on the main process. Because of this, it will *always* use threads for its internal concurrency (which the user doesn't actually see). However, internal concurrency strategy does not have any bearing on what is used to execute the callbacks:

1. Threads for bookkeeping and a thread pool executor for callbacks.
2. Threads for bookkeeping and a process pool executor for callbacks.
3. Threads for bookkeeping and a magical executor that sends callbacks to another server.
etc.

As long as the user provides an executor and a queue that can be used to communicate between the main process and the process executing the callbacks, everything can work as expected.

pubsub/.flake8
lukesneeringer: That is because you use a different style for multi-line split of function calls than I do. Both are PEP8-compliant, and I find mine easier to read and more cut-and-paste friendly.

PEP 8 _explicitly_ allows the way I do it:

screen shot 2017-07-20 at 12 55 08 pm

jonparrott: flake8 does not complain about that:

```
test = [
    1, 2, 3,
    4, 5, 6
]

result = list([
    1, 2,
    3, 4
])
```

pubsub/google/cloud/pubsub_v1/subscriber/client.py
lukesneeringer: Thanks. :-) I had been using a different naming convention from yours, changed back to yours, and forgot to update this.

pubsub/google/cloud/pubsub_v1/subscriber/histogram.py
lukesneeringer: It could? It seems unlikely to change, but the class could be more generally useful if we need the same concept elsewhere.

jonparrott: It's up to you.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: My intuition more or less matched @jonparrott's. I am definitely open to this in principle, but I am not going to pre-emptively do it.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: The while loop has a check that sees if the connection is still active and it quits if not.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
mdietz94: Fair enough, if it doesn't show up in load tests then I don't feel strongly.

lukesneeringer: And if it does, happy to adjust.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
jonparrott: SG.

pubsub/.flake8
jonparrott: Hrm, I'm not seeing it raise any complains about that.

lukesneeringer: I will re-enable it and go back and check and see what it complained about. If it's a PEP 8 violation, then I will change it and put the lint back. I am 95% sure it was not.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: @jonparrott I have implemented everything you identified except this, which we can discuss tomorrow. Not opposed in principle, but I want to understand rather than take your word for it.

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -0,0 +1,124 @@
...
+    def __call__(self):
+        while True:
+            item = self.queue.get()
+            if item == STOP:
+                break
+
+            # This doesn't presently deal with exceptions that bubble up
+            # through the callback. If there is an error here, the thread will
+            # exit and no further queue items will be processed. We could
+            # potentially capture errors, log them, and then continue on.
```
jonparrott: we should catch and log exceptions here and continue

pubsub/google/cloud/pubsub_v1/subscriber/consumer.py
```diff
@@ -0,0 +1,266 @@
+# Copyright 2017, Google Inc. All rights reserved.
```
jonparrott: make this private for now, we may want to move this into `api.core` at some point.

pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -0,0 +1,203 @@
...
+        self.lease()
+
+    def __repr__(self):
+        # Get an abbreviated version of the data.
+        abbv_data = self._message.data
+        if len(abbv_data) > 50:
+            abbv_data = abbv_data[0:50] + b'...'
+
+        # Return a useful representation.
+        answer = 'Message {\n'
```
jonparrott: nit: I'm not a fan of newlines in reprs.

pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -0,0 +1,203 @@
...
+        *finished* processing them, so that in the event of a failure,
+        you receive the message again.
+
+        .. warning::
+            Acks in Pub/Sub are best effort. You should always
+            ensure that your processing code is idempotent, as you may
+            receive any given message more than once.
+        """
+        time_to_ack = math.ceil(time.time() - self._received_timestamp)
+        self._request_queue.put(('ack', {
```
jonparrott: suggestion: put these "actions" into a enum or make them constants?

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,392 @@
...
+        # Now wait an appropriate period of time and do this again.
+        #
+        # We determine the appropriate period of time based on a random
+        # period between 0 seconds and 90% of the lease. This use of
+        # jitter (http://bit.ly/2s2ekL7) helps decrease contention in cases
+        # where there are many clients.
+        snooze = random.uniform(0.0, p99 * 0.9)
+        logger.debug('Snoozing lease management for %f seconds.' % snooze)
+        time.sleep(snooze)
+        self.maintain_leases()
```
jonparrott: this will eventually hit the recursion limit. Prefer splitting into a `update_leases` single-shot and a `maintain leases` that while loops, or just add a while loop to this one.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,392 @@
...
+        # the size counter.
+        if ack_id not in self.managed_ack_ids:
+            self.managed_ack_ids.add(ack_id)
+            self._bytes += byte_size
+
+        # Sanity check: Do we have too many things in our inventory?
+        # If we do, we need to stop the stream.
+        if self._load >= 1.0:
+            self._paused = True
+            self.close()
```
jonparrott: why stop the stream? Would it not be better to leave it open and just not queue any new requests until the load is lower?

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -0,0 +1,392 @@
...
+            self._bytes -= byte_size
+            self._bytes = min([self._bytes, 0])
+
+        # If we have been paused by flow control, check and see if we are
+        # back within our limits.
+        #
+        # In order to not thrash too much, require us to have passed below
+        # the resume threshold (80% by default) of each flow control setting
+        # before restarting.
+        if self._paused and self._load < self.flow_control.resume_threshold:
```
jonparrott: I feel like this snippet should be move to a separate function? `resume_from_load`?

pubsub/google/cloud/pubsub_v1/subscriber/message.py
lukesneeringer: The proto repr already has them, though.

jonparrott: Fair enough

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: We will still get incoming messages in that case, which we do not want.

pubsub/google/cloud/pubsub_v1/subscriber/consumer.py
lukesneeringer: Done.

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
lukesneeringer: Done (just added).

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
jonparrott: Gotcha.

lukesneeringer:
Merging this is blocked by CircleCI being hosed. Will try tomorrow morning.

lukesneeringer:
Nevermind -- Pub/Sub passed. It failed on Spanner. Merging now.

PR #3648, opened 2017-07-21
Added support for schema auto-detection feature in the job method `Lo…

WillianFuks:
Hello,

I did some implementation adding support for the feature "schema auto-detection" for BigQuery in the job method `LoadTableFromStorageJob`. 

Not sure if got it right, please let me know what you think and I can fix whatever issues you may find.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -83,7 +83,7 @@ def _error_result_to_exception(error_result):
 class AutoDetectSchema(_TypedProperty):
     """Typed Property for ``autodetect`` properties.
 
-       :raises: ValueError on `set` operation if `self.schema`
+       :raises: ValueError on `set` operation if `instance.schema`
                 is already defined.
     """
     def __set__(self, instance, value):
@@ -519,7 +519,7 @@ class _LoadConfiguration(object):
     """
     _allow_jagged_rows = None
     _allow_quoted_newlines = None
-    _autodetect_schema = None
+    _autodetect = None
     _create_disposition = None
     _encoding = None
     _field_delimiter = None
@@ -588,9 +588,9 @@ class LoadTableFromStorageJob(_AsyncJob):
             if not all(isinstance(field, SchemaField) for field in value):
                 raise ValueError('Schema items must be fields')
             if getattr(self, '_configuration', None):
-                if self.autodetect_schema:
+                if self.autodetect:
                     raise ValueError('Schema can not be set if '
-                                     '`autodetect_schema` property is True')
+                                     '`autodetect` property is True')
 
             self._schema = tuple(value)
 
@@ -648,7 +648,7 @@ class LoadTableFromStorageJob(_AsyncJob):
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.allowQuotedNewlines
     """
 
-    autodetect_schema = AutoDetectSchema('autodetect_schema', bool)
+    autodetect = AutoDetectSchema('autodetect', bool)
     """See
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.autodetect
     """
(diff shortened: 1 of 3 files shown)


dhermes:
I'll defer to @tswast or @jonparrott on review here. (In particular, review of the **surface** vs. review of the code itself.)

jonparrott:
@tswast is out-of-office. This surface seems reasonable to me.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -506,6 +529,11 @@ def output_rows(self):
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.allowQuotedNewlines
     """
 
+    autodetect_schema = AutoDetectSchema('autodetect_schema', bool)
```
tswast: Since it's just called `autodetect` in the backend API, I'd prefer that the client library stay consistent with that.

WillianFuks: Hi @tswast, thanks for the review :)! Sure, I can update that (I called it as it is thinking on being explicit for the end user that this property is related to detection of schemas, but agree that it changes overall structure of how the code has been built so far).

Just wondering, how should I proceed on updating the implementations of your reviews? Should I just commit the changes and make the PR again? Is there some required steps you recommend to follow?

Thanks!

tswast: You can add a second commit and push to your branch with fixes. GitHub allows for "Squash and merge" now, so it's no big deal if the git history gets a little messy in the PR.

WillianFuks: Thank you @tswast! Just committed the requested changes.

tswast:
@dhermes could you do a review pass, too, and merge if it looks okay to you?

dhermes:
Working on it now

dhermes:
@tseaver I noticed you assigned yourself, don't want to cross paths here but I've been actually in a local checkout of the code for about 15 minutes.

tseaver:
@dhermes Go ahead!

dhermes:
@WillianFuks I just sent a commit into this PR with a few style tweaks (thanks for adhering to the existing style BTW).

I did have a question about your expectation of the `_configuration` value in the `schema` `@property`, but I'll ask it on the code.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -572,12 +588,20 @@ def schema(self, value):
...
         """
-        if not all(isinstance(field, SchemaField) for field in value):
-            raise ValueError('Schema items must be fields')
-        self._schema = tuple(value)
+        if not value:
+            self._schema = ()
+        else:
+            if not all(isinstance(field, SchemaField) for field in value):
+                raise ValueError('Schema items must be fields')
+            if self._configuration is not None and self.autodetect:
```
dhermes: @WillianFuks This used to say `if getattr(self, '_configuration', None) and self.autodetect`, but that was because in the constructor

```python
self.schema = schema
```

came before

```python
self._configuration = _LoadConfiguration()
```

So `self._configuration` would've been an `AttributeError`.

----

TL;DR: 

- Why did you check for `self._configuration` being not-`None`?
- Will it ever actually be `None`?

WillianFuks: Hi @dhermes, thank you for the review!

Indeed I used the validation of `self._configuration` being already set because otherwise during instantiation of the `LoadTableFromStorageJob` class it would follow straight to the validation of `if self.autodetect: ...` which was not created yet (since `self._configuration = _LoadingConfiguration` was called after `self.schema = shema`).

If `self._configuration` is called before `self.schema` then it's indeed not necessary (I've chosen to do like I did so to not create restrictions on the order the variables are created, maybe it's easier to maintain a code with less restrictions, but as you have documented about `self.schema` having to occur after all variables were loaded then I think it's all fine!).

Leaving `self._configuration` be loaded before schema makes its validation unnecessary indeed, it could be just `if self.autodetect: raise ValueError()`

dhermes: Thanks for the quick response.

> so to not create restrictions on the order the variables are created

IMO, the correct approach here is a "dumb" constructor that just sets values. We are "breaking" this approach by using the fancy `schema` `@property` for validation in the constructor. All methods on a class should be able to freely assume that the attributes set in the constructor will be present.

> it's easier to maintain a code with less restrictions

IMO it's easier to maintain code completely devoid of `getattr` / `setattr`

> Leaving `self._configuration` be loaded before schema makes its validation unnecessary

OK I'll send an update to this PR that removes that check

WillianFuks:
Hi @dhermes , just to confirm, is everything ok in this PR? I saw this message from googlebot saying there's an issue with CLA, I wonder if I need to do something related.

Thanks :)!

dhermes:
@WillianFuks The CLA bot gets confused when there are commits from multiple authors, sorry for that.

As for "is everything OK", I seem to recall that I was waiting on your for response on something, but I can't find any comment to that effect.

dhermes:
I think the issue may just be the failed docs build. I'll rebase and hopefully it'll have been fixed by #3683

PR #3650, opened 2017-07-21
Translate GA

lukesneeringer:
This pull request bumps Translate to GA status, and should be followed by a release.

There are a couple of last minute changes:

  * The actual surface is moved from `translate` to `translate_v2`, with a alias in `translate`.
  * Documentation on models is changed to encourage the use of a string, rather than the constants.


pubsub/nox.py
```diff
@@ -34,14 +34,15 @@ def unit_tests(session, python_version):
     session.virtualenv_dirname = 'unit-' + python_version
 
     # Install all test dependencies, then install this package in-place.
-    session.install('mock', 'pytest', 'pytest-cov', *LOCAL_DEPS)
```
lukesneeringer: Oops, this file's presence is an error. Fixing.

translate/google/cloud/translate_v2/client.py
```diff
@@ -189,8 +189,8 @@ def translate(self, values, target_language=None, format_=None,
                                   in the query.
 
         :type model: str
-        :param model: (Optional) The model used to translate the text. The
-                      only accepted values are :attr:`BASE` and :attr:`NMT`.
+        :param model: (Optional) The model used to translate the text, such
+                      as ``'base'`` or ``'nmt'``.
```
dhermes: @lukesneeringer Are these no longer attributes? Or something else happened?

translate/setup.py
```diff
@@ -56,7 +56,7 @@
 
 setup(
     name='google-cloud-translate',
-    version='0.25.0',
+    version='1.0.0',
```
dhermes: Plan to bump the uber-package?

translate/tests/system.py
```diff
@@ -56,7 +56,7 @@ def test_translate(self):
         values = ['hvala ti', 'dankon',
                   'Me llamo Jeff', 'My name is Jeff']
         translations = Config.CLIENT.translate(
-            values, target_language='de', model=translate.NMT)
+            values, target_language='de', model='nmt')
```
dhermes: Could `translate_v2.NMT` work here or no?

translate/google/cloud/translate_v2/client.py
lukesneeringer: They are specified as a string in the API, and they do not want to update constants in the library whenever they add new ones.

translate/setup.py
lukesneeringer: We should, yes.

translate/tests/system.py
lukesneeringer: It will work, but it is not the format we want to advertise. Specific request from the API team.

PR #3658, opened 2017-07-24
Fix pylint for the main package

jonparrott:
NOTE: Pylint is still failing against tests, I have a longer-term strategy to address this which I'll propose later.

### DIFF
--- a/core/google/cloud/_http.py
+++ b/core/google/cloud/_http.py
@@ -279,7 +279,9 @@ class JSONConnection(Connection):
             can allow custom behavior, for example, to defer an HTTP request
             and complete initialization of the object at a later time.
 
-        :raises: Exception if the response code is not 200 OK.
+        :raises ~google.cloud.exceptions.GoogleCloudError: if the response code
+            is not 200 OK.
+        :raises TypeError: if the response content type is not JSON.
         :rtype: dict or str
         :returns: The API response payload, either as a raw string or
                   a dictionary if the response is valid JSON.
--- a/core/google/cloud/_testing.py
+++ b/core/google/cloud/_testing.py
@@ -14,17 +14,15 @@
 
 """Shared testing utilities."""
 
-
-# Avoid the grpc and google.cloud.grpc collision.
 from __future__ import absolute_import
 
 
 class _Monkey(object):
-    # context-manager for replacing module names in the scope of a test.
+    """Context-manager for replacing module names in the scope of a test."""
 
     def __init__(self, module, **kw):
         self.module = module
-        if len(kw) == 0:  # pragma: NO COVER
+        if not kw:  # pragma: NO COVER
             raise ValueError('_Monkey was used with nothing to monkey-patch')
         self.to_restore = {key: getattr(module, key) for key in kw}
         for key, value in kw.items():
@@ -68,8 +66,12 @@ def _tempdir_maker():
     return _tempdir_mgr
 
 
+# pylint: disable=invalid-name
+# Retain _tempdir as a constant for backwards compatibility despite
+# being an invalid name.
 _tempdir = _tempdir_maker()
 del _tempdir_maker
+# pylint: enable=invalid-name
 
 
 class _GAXBaseAPI(object):
@@ -79,7 +81,8 @@ class _GAXBaseAPI(object):
     def __init__(self, **kw):
         self.__dict__.update(kw)
 
-    def _make_grpc_error(self, status_code, trailing=None):
+    @staticmethod
+    def _make_grpc_error(status_code, trailing=None):
         from grpc._channel import _RPCState
         from google.cloud.exceptions import GrpcRendezvous
 
@@ -111,6 +114,7 @@ class _GAXPageIterator(object):
         self.page_token = kwargs.get('page_token')
 
     def next(self):
+        """Iterate to the next page."""
         import six
         return six.next(self._pages)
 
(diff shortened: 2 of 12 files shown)


core/google/cloud/_http.py
```diff
@@ -279,7 +279,9 @@ def api_request(self, method, path, query_params=None,
             can allow custom behavior, for example, to defer an HTTP request
             and complete initialization of the object at a later time.
 
-        :raises: Exception if the response code is not 200 OK.
+        :raises ~google.cloud.exceptions.GoogleCloudError: if the response code
+            is not 200 OK.
+        :raises TypeError: if the response content type is not JSON.
```
dhermes: @jonparrott I've tried to fight `pylint` on this error before and gave up.

I did because

1. There were too many
1. It seemed I couldn't get them to render the way I hoped

core/google/cloud/_testing.py
```diff
@@ -14,17 +14,15 @@
...
 """Shared testing utilities."""
 
-
-# Avoid the grpc and google.cloud.grpc collision.
 from __future__ import absolute_import
 
 
 class _Monkey(object):
-    # context-manager for replacing module names in the scope of a test.
+    """Context-manager for replacing module names in the scope of a test."""
```
dhermes: This stirs a memory of my previous effort to deprecate `_Monkey`. Are we close?

```
$ git log -1 --pretty=%H
768d667b6c73f099935dd6f93f6596cdcaffecde
$ git grep _Monkey | wc -l
    124
$ git grep _Monkey | cut -f 1 -d '/' | sort | uniq
bigtable
core
datastore
logging
spanner
speech
$ git grep _Monkey -- bigtable/ | wc -l
     28
$ git grep _Monkey -- core/ | wc -l
     42
$ git grep _Monkey -- datastore/ | wc -l
      2
$ git grep _Monkey -- logging/ | wc -l
      4
$ git grep _Monkey -- spanner/ | wc -l
     32
$ git grep _Monkey -- speech/ | wc -l
     16
```

core/google/cloud/_testing.py
```diff
@@ -68,8 +66,12 @@ def _tempdir_mgr():
     return _tempdir_mgr
 
 
+# pylint: disable=invalid-name
+# Retain _tempdir as a constant for backwards compatibility despite
```
dhermes: Backwards compatibility with what? Just get rid of it?

core/google/cloud/client.py
```diff
@@ -64,7 +64,7 @@ def from_service_account_json(cls, json_credentials_path, *args, **kwargs):
 
         :rtype: :class:`_ClientFactoryMixin`
         :returns: The client created with the retrieved JSON credentials.
-        :raises: :class:`TypeError` if there is a conflict with the kwargs
+        :raises TypeError: if there is a conflict with the kwargs
                  and the credentials created by the factory.
```
dhermes: This indent used to be aligned with the `:class:...`. Your call on how to "fix" / if it needs to be "fixed"

core/google/cloud/iterator.py
```diff
@@ -242,7 +242,8 @@ def _page_iter(self, increment):
                           results per page while an items iterator will want
                           to increment per item.
 
-        Yields :class:`Page` instances.
+        :rtype: :class:`Page`
+        :returns: pages
```
dhermes: I'd be happier just converting to a google-style docstring here (since it actually supports `Yields`). WDYT?

core/google/cloud/_http.py
jonparrott: This seems correct?

core/google/cloud/_testing.py
jonparrott: I will as part of the core v1.0.0 refactor I proposed, but I wanted to do this without breaking any interfaces.

core/google/cloud/client.py
jonparrott: I'd prefer not to be picky about hanging indent vs single indent in docstrings.

core/google/cloud/iterator.py
jonparrott: Let's do that as part of the core refactor later. :)

core/google/cloud/_testing.py
dhermes: I'd say we're close.

dhermes: @lukesneeringer @tseaver Can we all "sprint" on this one day this week?

core/google/cloud/iterator.py
dhermes: I have a (maybe not legitimate?) worry here: During a re-factor we may just blindly convert it into a `Returns` section instead of a `Yields` section.

jonparrott: Pylint will catch that. :)

dhermes: I didn't believe. But I just confirmed it:

```python
"""When package begin."""


def yield_a_carrot():
    """Yields two carrots actually.

    Returns:
        str: Two carrots, actually.
    """
    yield 'carrot'
    yield 'carrot'
```

then

```
$ pylint eebowai.py
************* Module eebowai
W:  4, 0: Redundant returns documentation (redundant-returns-doc)
W:  4, 0: Missing yield documentation (missing-yield-doc)
W:  4, 0: Missing yield type documentation (missing-yield-type-doc)
W:  4, 0: Missing yield documentation (missing-yield-doc)
W:  4, 0: Missing yield type documentation (missing-yield-type-doc)

------------------------------------
Your code has been rated at -6.67/10
```

PR #3662, opened 2017-07-24
Split polling future into its own module

jonparrott:


jonparrott:
I did edit bigquery to reference the new name, but yes, polling itself was not modified.

Issue #3562, opened 2017-06-28
gRPC _Rendezvous w/ 'ABORTED' status propagated from 'Session.run_in_transaction'


dhermes:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2281

@tseaver Can you triage / update the issue title to be more specific?

tseaver:
The code for `Session.run_in_transaction` expects that the `ABORTED` bit will be propagated as an instance of `google.gax.errors.GaxError`:  somehow, the underlying GAPIC code is raising a gRCP `_Rendezvous` instead.

tseaver:
@lukesneeringer Do we need to be regenerating the GAPIC code?

vladkluev:
Hi this is very annoying, to have a whole bunch of exceptions thrown in my application whenever a transaction is aborted. Is there any workaround to this issue at the moment?

tseaver:
@vladkluev I'm hopeful we will have a fix pushed out shortly (this bug blocks PR #3663).

tseaver:
So, the issue is that the `streaming_read` and `streaming_execute_sql` calls are propagating the `_Rendezvous` error, rather than wrapping them up in a `GaxError` as the `commit` call does.

bjwatson:
@tseaver I'm discuss this with @lukesneeringer offline.

lukesneeringer:
Repeating this back to ensure I understand: Basically this is a situation where one call wraps the errors while others do not? Is this a configuration difference? Something else?

tseaver:
@lukesneeringer the new system test in #3663 provokes it routinely:  the gRPC `ABORTED` return triggered by `read_streaming` is *not* wrapped, while the other system tests (which mean to trigger it in the `commit` call) do get it wrapped.

lukesneeringer:
Okay, so it is literally non-deterministic then?

tseaver:
It fails repeatably on #3663.  My *suspicion* is that the GAX/GAPIC streaming support is involved (our `read` and `execute_sql` methods actually call the streaming APIs).

tseaver:
@vladkluev The `ABORTED` error is *supposed* to propagate as a `GaxError`, but is somehow escaping as the lower-level `grpc._Rendezvous`.  The `run_in_transaction` code handles retrying correctly when `GaxError` is raised.

vladkluev:
Ah gotcha

bjwatson:
[...]

PR #3663, opened 2017-07-24
 Add a test which provokes abort-during-read during 'run_in_transaction'.

tseaver:
~~Uses #3615 as a base. ca441f7 is the only change from that PR.~~

### DIFF
--- a/spanner/google/cloud/spanner/session.py
+++ b/spanner/google/cloud/spanner/session.py
@@ -24,6 +24,7 @@ from grpc import StatusCode
 
 # pylint: disable=ungrouped-imports
 from google.cloud.exceptions import NotFound
+from google.cloud.exceptions import GrpcRendezvous
 from google.cloud.spanner._helpers import _options_with_prefix
 from google.cloud.spanner.batch import Batch
 from google.cloud.spanner.snapshot import Snapshot
@@ -286,7 +287,7 @@ class Session(object):
                 txn.begin()
             try:
                 return_value = func(txn, *args, **kw)
-            except GaxError as exc:
+            except (GaxError, GrpcRendezvous) as exc:
                 _delay_until_retry(exc, deadline)
                 del self._transaction
                 continue
@@ -318,7 +319,12 @@ def _delay_until_retry(exc, deadline):
     :type deadline: float
     :param deadline: maximum timestamp to continue retrying the transaction.
     """
-    if exc_to_code(exc.cause) != StatusCode.ABORTED:
+    if isinstance(exc, GrpcRendezvous):  # pragma: NO COVER  see #3663
+        cause = exc
+    else:
+        cause = exc.cause
+
+    if exc_to_code(cause) != StatusCode.ABORTED:
         raise
 
     now = time.time()
@@ -326,7 +332,7 @@ def _delay_until_retry(exc, deadline):
     if now >= deadline:
         raise
 
-    delay = _get_retry_delay(exc)
+    delay = _get_retry_delay(cause)
     if delay is not None:
 
         if now + delay > deadline:
@@ -336,7 +342,7 @@ def _delay_until_retry(exc, deadline):
 # pylint: enable=misplaced-bare-raise
 
 
-def _get_retry_delay(exc):
+def _get_retry_delay(cause):
     """Helper for :func:`_delay_until_retry`.
 
     :type exc: :class:`google.gax.errors.GaxError`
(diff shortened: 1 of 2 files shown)


tseaver:
This test can deadlock:  I got interrupted while debugging and blew my stack. :(

tseaver:
@dhermes, @bjwatson, @lukesneeringer OK, I have the test passing, but to do so I have to straddle the fact that the GAPIC / gRPC bits sometimes raise a `_Rendezvous` w/ status code `ABORTED`, and sometimes a `GaxError` with such a `_Rendezvous` as their cause.  I think the APIs should raise one or the other, not a mix-and-match.

Specifics:

- `ABORTED` returned from the server during a `Read` API call is raised as a `_Rendezvous`.
` `ABORTED` returned from the server during a `Commit` API call is raised as a `GaxError`.

My hack to get the read-abort bit passing:

```python
@@ -312,7 +334,12 @@ def _delay_until_retry(exc, deadline):
     :type deadline: float
     :param deadline: maximum timestamp to continue retrying the transaction.
     """
-    if exc_to_code(exc.cause) != StatusCode.ABORTED:
+    if isinstance(exc, GrpcRendezvous):
+        cause = exc
+    else:
+        cause = exc.cause
+
+    if exc_to_code(cause) != StatusCode.ABORTED:
         raise
 
     now = time.time()
@@ -320,7 +347,7 @@ def _delay_until_retry(exc, deadline):
     if now >= deadline:
         raise
 
-    delay = _get_retry_delay(exc)
+    delay = _get_retry_delay(cause)
     if delay is not None:
 
         if now + delay > deadline:
@@ -330,7 +357,7 @@ def _delay_until_retry(exc, deadline):
 # pylint: enable=misplaced-bare-raise
 
 
-def _get_retry_delay(exc):
+def _get_retry_delay(cause):
     """Helper for :func:`_delay_until_retry`.
 
     :type exc: :class:`google.gax.errors.GaxError`
@@ -339,7 +366,7 @@ def _get_retry_delay(exc):
     :rtype: float
     :returns: seconds to wait before retrying the transaction.
     """
-    metadata = dict(exc.cause.trailing_metadata())
+    metadata = dict(cause.trailing_metadata())
     retry_info_pb = metadata.get('google.rpc.retryinfo-bin')
     if retry_info_pb is not None:
         retry_info = RetryInfo()
```

tseaver:
Turns out my issue is the same as #3562.

tseaver:
I've rebased, fixed the deadlock / hang, and added a separate commit (f916edc) with my workaround for #3562 so that I can verify the tests pass on Circle.  I propose waiting for a real fix and then removing that commit.

tseaver:
The [test failure](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2651?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) is for coverage of branches added in the to-be-backed-out commit (f916edc).

jonparrott:
@tseaver once #3738 is in, you can use `google.api.core.exceptions.from_grpc_exception` to map the `grpc.RpcError` to a Google API exception. You shouldn't need to catch `_Rendezvous` directly, instead, catch `grpc.RpcError` or `grpc.Call` (if you need the metadata directly).

bjwatson:
@tseaver I removed the `blocked` label since it looks like @jonparrott unblocked this.

tseaver:
@jonparrott Note that for the `ABORT` error, I'm *not* propagating the error, but using it to trigger a retry (I *do* need access to the trailing metadata to pick out the retry interval).

@bjwatson Am I still supposed to be catching `GaxError` as well as `grpc.Call` here?

jonparrott:
@tseaver yes, in this case catch GaxError and grpc.Call.

tseaver:
@jonparrott  Why are we not fixing the streaming iterator stuff to return only one kind of error?

jonparrott:
@tseaver we are, but it's O(weeks) away, and changing this PR to catch `grpc.Call` instead of `GrpcRendezvous` is a fine compromise for now- in the near term, it'll unblock spanner. Once `google.api.core.grpc` exists and is used by gapic, it will not raise this error which will mean a small change will need to be made here (simplifying this to just catch a `google.api.core.exception` subclass).

tseaver:
@jonparrott Note that one cannot catch `grpc.Call` (it doesn't derive from `BaseException`).

Given that the change is intended to be temporary, I will merge with a `# pragma: NO COVER` on the branch for `if isinstance(exc, GrpcRendezvous)`, and add an issue to remove that once the underling normalization fix has rolled out.

PR #3664, opened 2017-07-24
Re-enable flake8 for core package and tests

jonparrott:
Note to reviewers: this is intentionally 2 commits. The first re-enables linting, the second addresses lint issues.

### DIFF
--- a/core/google/cloud/_helpers.py
+++ b/core/google/cloud/_helpers.py
@@ -25,20 +25,20 @@ import os
 import re
 from threading import local as Local
 
+import google_auth_httplib2
+import httplib2
+import six
+from six.moves import http_client
+
 import google.auth
 from google.protobuf import duration_pb2
 from google.protobuf import timestamp_pb2
-import google_auth_httplib2
 
 try:
     import grpc
     import google.auth.transport.grpc
 except ImportError:  # pragma: NO COVER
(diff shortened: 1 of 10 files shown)


core/google/cloud/_helpers.py
```diff
@@ -25,21 +25,21 @@
...
+from six.moves import http_client
+
 import google.auth
 from google.protobuf import duration_pb2
 from google.protobuf import timestamp_pb2
-import google_auth_httplib2
 
 try:
     import grpc
     import google.auth.transport.grpc
```
dhermes: Does `flake8` not care about these?

core/google/cloud/_http.py
```diff
@@ -16,8 +16,8 @@
 
 import json
 import platform
-from pkg_resources import get_distribution
 
+from pkg_resources import get_distribution
```
dhermes: TIL! Which 3rd party package does this come from? `setuptools`?

jonparrott: yep, setuptools.

core/google/cloud/_helpers.py
jonparrott: It's smart enough to know that try: except imports are their own beasties.

PR #3667, opened 2017-07-25
Removing `get_credentials()` from `core`.

dhermes:
In the process also:

- Slight re-org on `nox.py` config (to pass posargs) for `core` and `datastore`
- Getting rid of last usage of `_Monkey` in datastore

This is part of `@jonparrott`'s effort to slim down / stabilize `core`.

-----

Temporarily labeled "do not merge" so we can discuss "churn" first.


core/google/cloud/client.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, credentials=None, _http=None):
                     credentials, google.auth.credentials.Credentials)):
             raise ValueError(_GOOGLE_AUTH_CREDENTIALS_HELP)
         if credentials is None and _http is None:
-            credentials = get_credentials()
+            credentials, _ = google.auth.default()
```
jonparrott: hrm, this throws away the project. Does this match the behavior we're seeing today?

I would very much like to unify all of this - it seems we could probably fold all of the classes here into one and have a flag on whether a `project` is required or not. But you don't have to do this here, just thinking out loud.

dhermes: > Does this match the behavior we're seeing today?

Yes.

----

FWIW, I noticed this as well when making the change but didn't think "fixing" it made sense in this PR.

dhermes:
Rebased after #3668 was merged (so now `google.cloud.credentials` is gone)

PR #3668, opened 2017-07-25
Moving bytes signing helpers from `core` to `storage`.

dhermes:
This is part of `@jonparrott`'s effort to slim down / stabilize `core`.

-----

Temporarily labeled "do not merge" so we can discuss "churn" first.


tseaver:
Wouldn't it make more sense to base this on #3667?  For instance, at that point, you could entirely remove the `credentials` module from `core`.

dhermes:
@tseaver I'm trying to reduce burden on reviewers by making the PR small.

tseaver:
Yeah, but then we have to review a *third* PR which just removes that module?  The only thing gonna be left in it after these two are merged is the copyright comment.

dhermes:
@tseaver I plan on rebasing the one that gets merged 2nd.

Issue #3364, opened 2017-05-03
Spanner should have helpful aliases for param types


jonparrott:
Having to do:

```python
from google.cloud import spanner
from google.cloud.proto.spanner.v1 import type_pb2

instance = spanner_client.instance(instance_id)
database = instance.database(database_id)

params = {
    'start': start,
    'end': end
}
param_types = {
    'start': type_pb2.Type(code=type_pb2.STRING),
    'end': type_pb2.Type(code=type_pb2.STRING)
}
results = database.execute_sql(
    "SELECT AlbumId, AlbumTitle, MarketingBudget "
    "FROM Albums@{FORCE_INDEX=AlbumsByAlbumTitle} "
    "WHERE AlbumTitle >= @start AND AlbumTitle < @end",
    params=params, param_types=param_types)

```

Is rather awkward, it would be much more friendly to have:

```python

from google.cloud import spanner

instance = spanner_client.instance(instance_id)
database = instance.database(database_id)

params = {
    'start': start,
    'end': end
}
param_types = {
    'start': spanner.ParamTypes.STRING,
    'end': spanner.ParamTypes.STRING
}
results = database.execute_sql(
    "SELECT AlbumId, AlbumTitle, MarketingBudget "
    "FROM Albums@{FORCE_INDEX=AlbumsByAlbumTitle} "
    "WHERE AlbumTitle >= @start AND AlbumTitle < @end",
    params=params, param_types=param_types)
```

or similar

tseaver:
@jonparrott Aliasing the "scalar" types is easy (PR #3670).  What do you envision for array / struct?

jonparrott:
@tseaver would `list` and `dict` work, e.g.:

```python
param_types = {
     tags: list,
     metadata: dict
}
```

jonparrott:
Oh nevermind, ignore my last comment, I'm being thick.

What does it look like today to if you need to specify an array / struct?

tseaver:
@jonparrott For arrays, you have to pass the type of the elements:

```python
array_type = type_pb2.Type(
    code=type_pb2.ARRAY, array_element_type=type_pb2.Type(code=type_pb2.INT64))
```

For structs, you have to construct / pass a `StructType` describing the members:

```python
struct_type_pb = type_pb2.StructType(fields=[
    type_pb2.StructType.Field(name='name', type=type_pb2.Type(code=type_pb2.STRING)),
[...]

PR #3670, opened 2017-07-25
Create parameter type aliases for scalar field types.

tseaver:
See: #3364

### DIFF
--- a/spanner/google/cloud/spanner/__init__.py
+++ b/spanner/google/cloud/spanner/__init__.py
@@ -18,6 +18,13 @@
 import pkg_resources
 __version__ = pkg_resources.get_distribution('google-cloud-spanner').version
 
+from google.cloud.spanner._helpers import BOOL_PARAM_TYPE
+from google.cloud.spanner._helpers import BYTES_PARAM_TYPE
+from google.cloud.spanner._helpers import DATE_PARAM_TYPE
+from google.cloud.spanner._helpers import FLOAT64_PARAM_TYPE
+from google.cloud.spanner._helpers import INT64_PARAM_TYPE
+from google.cloud.spanner._helpers import STRING_PARAM_TYPE
+from google.cloud.spanner._helpers import TIMESTAMP_PARAM_TYPE
 
 from google.cloud.spanner.client import Client
 
--- a/spanner/google/cloud/spanner/__init__.py
+++ b/spanner/google/cloud/spanner/__init__.py
@@ -18,14 +18,6 @@
 import pkg_resources
 __version__ = pkg_resources.get_distribution('google-cloud-spanner').version
 
-from google.cloud.spanner._helpers import BOOL_PARAM_TYPE
-from google.cloud.spanner._helpers import BYTES_PARAM_TYPE
-from google.cloud.spanner._helpers import DATE_PARAM_TYPE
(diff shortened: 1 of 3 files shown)


dhermes:
I'll leave this review to @jonparrott.

Is there any reason to add docstrings?

spanner/google/cloud/spanner/__init__.py
```diff
@@ -18,6 +18,13 @@
 import pkg_resources
 __version__ = pkg_resources.get_distribution('google-cloud-spanner').version
 
+from google.cloud.spanner._helpers import BOOL_PARAM_TYPE
```
jonparrott: I feel weird about plucking members out of a private module to expose. Do we have a `spanner.types` module?

tseaver: Seems like a perfectly reasonable use of https://en.wikipedia.org/wiki/Facade_pattern to me:  the "implementation" is private, but the things themselves are exposed to the public.

tseaver: And no, there is no extant `google.cloud.spanner.types` module.

jonparrott: @tseaver do you think it's worthwhile for us to introduce a `types` module to hold these constants? (like we've done with gapic clients like vision, speech, etc.)

lukesneeringer: I would like to move in the direction of adding a `types` module for everything; I had not done it for Spanner yet.

@tseaver Any objections to my doing that if I can do it today or tomorrow?

tseaver: fa34a34 adds the module here.  I'm not sure what else you have in mind for it.

spanner/google/cloud/spanner/types.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Types exported from this package."""
+
+from google.cloud.proto.spanner.v1 import type_pb2
+
+
+# Scalar paramter types
+STRING_PARAM_TYPE = type_pb2.Type(code=type_pb2.STRING)
```
lukesneeringer: Ideally, I would like to use the identical pattern we are using elsewhere, which has a few nice usability wins:

  1. It uses the identical type names to the `_pb2` modules.
  2. It rewrites `__module__` which helps users when looking at docs.
  3. It grabs all types from the modules on the fly, so if Spanner introduces new types, it does not need manual updating.

[Here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/vision/google/cloud/vision_v1/types.py#L23-L35) is what I did in Vision. Would you have any objection to doing that here?

lukesneeringer: Derp, I just realized that you are _instantiating_ `type_pb2.Type` with constants.

jonparrott:
Note that we don't expose protos directly in spanner. These items are a
special case where we can't map Python types to Spanner types so we need
some constant and using the pb2 Types is very convenient.

lukesneeringer:
Yep, I realized my mistake.

tseaver:
@lukesneeringer Any reason not to merge?

lukesneeringer:
Merge away.

PR #3671, opened 2017-07-25
Simplifying usage of Client classes in Spanner and Bigtable

dhermes:
As a follow-up commit, also simplifying the Client classes.

I am "throwing this up" as a half-finished PR so @jonparrott can take the wheel.

### DIFF
--- a/bigtable/google/cloud/bigtable/client.py
+++ b/bigtable/google/cloud/bigtable/client.py
@@ -31,7 +31,6 @@ In the hierarchy of API concepts
 
 import os
 
-import google.auth
 import google.auth.credentials
 from google.gax.utils import metrics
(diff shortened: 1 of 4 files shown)


jonparrott:
I'm gonna close this for now. This is helpful because it's revealed just how much refactoring will need to happen to use a new client superclass. With the move to google.api.core, I will introduce the new base classes, switch these to that, then drop `core.client`.

dhermes:
@jonparrott At the very least the first commit should go in (just to push all of the warts into one place)

dhermes:
I just force-pushed this branch with one fewer commit.

dhermes:
And the force-push made it impossible to "Reopen and comment"

PR #3512, opened 2017-06-20
Auto-generated trace library (GAPIC only)

liyanhui1228:

PR #3672, opened 2017-07-25
Simplifying Client constructor's for Bigtable and Spanner.

dhermes:
Was part of #3671.

### DIFF
--- a/spanner/google/cloud/spanner/client.py
+++ b/spanner/google/cloud/spanner/client.py
@@ -35,8 +35,7 @@ from google.cloud.gapic.spanner_admin_instance.v1.instance_admin_client import (
 # pylint: enable=line-too-long
 
 from google.cloud._http import DEFAULT_USER_AGENT
-from google.cloud.client import _ClientFactoryMixin
-from google.cloud.client import _ClientProjectMixin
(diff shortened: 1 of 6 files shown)


spanner/google/cloud/spanner/client.py
```diff
@@ -104,21 +103,16 @@ class Client(_ClientFactoryMixin, _ClientProjectMixin):
     _database_admin_api = None
     _SET_PROJECT = True  # Used by from_service_account_json()
 
+    SCOPE = (SPANNER_ADMIN_SCOPE,)
```
dhermes: Worth noting this is a change to the public interface of `spanner.Client` (previously `SCOPE` was the [default][1] value)

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/2083542f3d799a5011fcd4d3092257b958d5511a/core/google/cloud/client.py#L126

dhermes:
I never actually fixed the unit tests, it's open for anyone who has cycles to tackle right meow :smirk_cat:

dhermes:
OK @jonparrott @tseaver @lukesneeringer this is review-able now

dhermes:
Docs breakage in build is caused by #3512

PR #3678, opened 2017-07-26
Python GAPIC: Speech v1

lukesneeringer:
This pull request was generated by artman. Please review it thoroughly before merging.

### DIFF
--- a/speech/google/cloud/gapic/speech/v1/speech_client_config.json
+++ b/speech/google/cloud/gapic/speech/v1/speech_client_config.json
@@ -6,9 +6,7 @@
           "DEADLINE_EXCEEDED",
           "UNAVAILABLE"
         ],
-        "non_idempotent": [
-          "UNAVAILABLE"
-        ]
+        "non_idempotent": []
       },
       "retry_params": {
         "default": {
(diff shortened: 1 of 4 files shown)


speech/google/cloud/gapic/speech/v1/speech_client_config.json
```diff
@@ -6,9 +6,7 @@
           "DEADLINE_EXCEEDED",
           "UNAVAILABLE"
         ],
-        "non_idempotent": [
-          "UNAVAILABLE"
-        ]
+        "non_idempotent": []
```
tseaver: This bit is intentional?  Most of the others of these I have seen allow retries for `UNAVAILABLE`, even for non-idempotent methods.

lukesneeringer: Good question; I will find out. I _expect_ it is, but I am unsure.

lukesneeringer: @shinfan Any idea if this was an intentional change? As in, we _do not_ want to retry `UNAVAILABLE` in Speech?

lukesneeringer: I have not been able to chase this down, and this needs to go out today for a Monday announcement. I can make the change later if it ends up being wrong.

PR #3688, opened 2017-07-27
Fix translate docs build.

lukesneeringer:
Fixes a tiny problem with building the translate docs.

### DIFF
--- a/nox.py
+++ b/nox.py
@@ -30,13 +30,13 @@ def docs(session):
     # Install Sphinx and also all of the google-cloud-* packages.
     session.chdir(os.path.realpath(os.path.dirname(__file__)))
     session.install('Sphinx >= 1.6.2', 'sphinx_rtd_theme')
+    session.install('.')
     session.install(
         'core/', 'bigquery/', 'bigtable/', 'datastore/', 'dns/', 'language/',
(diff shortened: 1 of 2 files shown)


nox.py
```diff
@@ -30,13 +30,13 @@ def docs(session):
     # Install Sphinx and also all of the google-cloud-* packages.
     session.chdir(os.path.realpath(os.path.dirname(__file__)))
     session.install('Sphinx >= 1.6.2', 'sphinx_rtd_theme')
+    session.install('.')
```
tseaver: Can we just kill off installing the umbrella package altogether?  Its only effect is to install the subpackages from PyPI / wheel cache, which we then immediately overwrite with the local versions.

tseaver: I just verified that `nox -e docs` fails without that line (dammit).  OIC, it is because `docs/conf.py` sniffs the version via `pkg_resources`.

lukesneeringer: I went through the _exact_ same process.

Soon.

tseaver: `session.install('--no-deps', '.')` works, but turns into `pip install --upgrade --nodeps .`, which seems to install all the dependencies anyway. :(

Issue #3237, opened 2017-03-29
Logging system tests are hanging on CircleCI


dhermes:
Four builds all hung:

- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1413 (after `.........` == 9)
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1415 (after `.............` == 13)
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1418 (after `...........` == 11)
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/1426 (after `......` == 6)

I have run these tests 8+ times locally since the original failure (and using the same Cloud Console project) and they run without issue.

@waprin Any ideas?

dhermes:
Note this is "blocking" the release of `error_reporting == 0.23.3` /cc @DizzeePascall

dhermes:
Running `py.test tests/system.py -vv` (under the `logging` directory) the (deterministic) order of test cases is:

```
01 tests/system.py::TestLogging::test_create_metric PASSED
02 tests/system.py::TestLogging::test_create_sink_bigquery_dataset PASSED
03 tests/system.py::TestLogging::test_create_sink_pubsub_topic PASSED
04 tests/system.py::TestLogging::test_create_sink_storage_bucket PASSED
05 tests/system.py::TestLogging::test_list_metrics PASSED
06 tests/system.py::TestLogging::test_list_sinks PASSED
07 tests/system.py::TestLogging::test_log_handler_async PASSED
08 tests/system.py::TestLogging::test_log_handler_sync PASSED
09 tests/system.py::TestLogging::test_log_root_handler PASSED
10 tests/system.py::TestLogging::test_log_struct PASSED
11 tests/system.py::TestLogging::test_log_struct_w_metadata PASSED
12 tests/system.py::TestLogging::test_log_text PASSED
13 tests/system.py::TestLogging::test_log_text_w_metadata PASSED
14 tests/system.py::TestLogging::test_log_text_with_timestamp PASSED
15 tests/system.py::TestLogging::test_reload_metric PASSED
16 tests/system.py::TestLogging::test_reload_sink PASSED
17 tests/system.py::TestLogging::test_update_metric PASSED
18 tests/system.py::TestLogging::test_update_sink PASSED
```

(with numbers added by me).

waprin:
[...]

PR #3691, opened 2017-07-27
Allowing logging system tests to fail.

dhermes:
These hose our builds. (See my branch name.)

FWIW Someone should also deal with the eventual-consistency issue with Pub/Sub snapshot system tests.

### DIFF
--- a/logging/nox.py
+++ b/logging/nox.py
@@ -71,7 +71,13 @@ def system_tests(session, python_version):
     session.install('.')
 
     # Run py.test against the system tests.
-    session.run('py.test', '-vvv', 'tests/system.py', *session.posargs)
+    session.run(
+        'py.test',
+        '-vvv',
+        'tests/system.py',
+        *session.posargs,
+        success_codes=range(0, 100),
+    )
 
 
 @nox.session


jonparrott:
This is on @lukesneeringer

dhermes:
Also worth noting that @jonparrott suggested I used `pytest.mark.xfail` but we've got some confusion / disagreement about how it works. As a test (for anyone interested) make `test_it.py` with

```python
import random
import pytest


@pytest.mark.xfail
def test_rand():
    assert random.random() < 0.5
```

and running it

```
$ py.test test_it.py
=============== test session starts ===============
platform linux2 -- Python 2.7.13, pytest-3.1.0, py-1.4.34, pluggy-0.4.0
rootdir: ${CWD}, inifile: tox.ini
plugins: cov-2.4.0
collected 1 items

test_it.py x

=============== 1 xfailed in 0.00 seconds ===============
$ echo ${?}
0
$ py.test test_it.py
=============== test session starts ===============
platform linux2 -- Python 2.7.13, pytest-3.1.0, py-1.4.34, pluggy-0.4.0
rootdir: ${CWD}, inifile: tox.ini
plugins: cov-2.4.0
collected 1 items

test_it.py X

=============== 1 xpassed in 0.00 seconds ===============
$ echo ${?}
0
```

lukesneeringer:
My understanding about `@pytest.mark.xfail` is that it will fail the test if the test _does not_ fail, which is not really what you want for a flaky test.

dhermes:
@lukesneeringer That **was** my own understanding (and @jonparrott's as well). Please run the [above][1] / confirm I'm not insane.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3691#issuecomment-318442077

lukesneeringer:
Yeah, it looks like `@pytest.mark.xfail` is not the semantic equivalent of `@unittest.expectedFailure`.

I have mixed feelings about this. _Both_ of those concepts have value.

tseaver:
@jonparrott Do we need to wait for their feedback to tighten the tourniquet?  They can always open new PRs which fix the logging tests and remove the "allow-them-to-fail" switch.

jonparrott:
@tseaver I'm okay to merge this, I just wanted Doug and Angela to be aware that it's happening.

liyanhui1228:
Thanks for letting us know! This looks good.

dhermes:
Also FYI (whoever is going to fix the b0rken tests): The issue is in clean-up, when some resources seem to 404 (on delete) for 10+ minutes. ISTM that the 404 is "real" and clean-up is doing the wrong thing.

liyanhui1228:
Can I get the link to the CircleCI build which has the error message for the failing system test? I'll try to fix it.

dhermes:
#3237 has a very long history of this failure. If you want a recent failure, just visit https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/tree/master, about 5 out of every 6 builds (I just made up that ratio) fails because of logging.

liyanhui1228:
Thanks for this information, I'll look into it.

dhermes:
So this didn't actually fix the build b0rken-ness because of the CircleCI timeout ([e.g.][1]). I think we should reduce the number of retries / add `-s` to the command so that there won't be 10+ minutes without any output.

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2661

liyanhui1228:
Yeah the CircleCI will kill the test if there are 10+ minutes without any output and won't care the success code we set here. We can change the max retries to 9 and then the time will become 1 + 2 + ... + 256 = 511 seconds which is less than 600 seconds time out threshold. I have prepared a PR for it.

Also question for the clean up issue, is the loggers that should be deleted have already been deleted somewhere other than in the tearDown method and then causing 404?

dhermes:
> Also question for the clean up issue, is the loggers that should be deleted have already been deleted somewhere other than in the tearDown method and then causing 404?

I don't know the system test code well enough to say, so your assessment from reading the code would be the same as mine.

I **assume** the original author's intent was to clean up loggers that need to be deleted, but somehow the test deletes those loggers before finishing (which means there is nothing to cleanup). Either that or the logging API is broken somehow (the test passes **sometimes**).

PR #3692, opened 2017-07-27
Appease current pylint opinions:

tseaver:
- Document missing ':raises:', ':rtype:', and ':returns:'.
- Use ':raises <exception_type>:'.
- Add pro-forma docstrings to namespace package initializers.
- Avoid using 'len(seq) == 0' (or '!=') for boolean tests.

pylint still complains about import cycles:

```
************* Module google.cloud.spanner.streamed
R:  1, 0: Cyclic import (google.cloud.spanner ->
google.cloud.spanner.client -> google.cloud.spanner.instance ->
google.cloud.spanner.database) (cyclic-import)
R:  1, 0: Cyclic import (google.cloud.spanner ->
google.cloud.spanner.client) (cyclic-import)
```

### DIFF
--- a/spanner/google/cloud/spanner/_helpers.py
+++ b/spanner/google/cloud/spanner/_helpers.py
@@ -38,6 +38,7 @@ class TimestampWithNanoseconds(datetime.datetime):
     """
     __slots__ = ('_nanosecond',)
 
+    # pylint: disable=arguments-differ
     def __new__(cls, *args, **kw):
         nanos = kw.pop('nanosecond', 0)
         if nanos > 0:
@@ -48,6 +49,7 @@ class TimestampWithNanoseconds(datetime.datetime):
         inst = datetime.datetime.__new__(cls, *args, **kw)
         inst._nanosecond = nanos or 0
         return inst
+    # pylint: disable=arguments-differ
 
     @property
     def nanosecond(self):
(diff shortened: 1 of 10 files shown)


spanner/google/cloud/spanner/_helpers.py
```diff
@@ -38,6 +38,7 @@ class TimestampWithNanoseconds(datetime.datetime):
     """
     __slots__ = ('_nanosecond',)
 
+    # pylint: disable=arguments-differ
```
dhermes: This is a nasty one. Does this feel like a bug in Pylint? (`disable` is fine as-is)

spanner/google/cloud/spanner/_helpers.py
```diff
@@ -110,7 +113,7 @@ def _try_to_coerce_bytes(bytestring):
                          'base64-encoded bytes.')
 
 
-# pylint: disable=too-many-return-statements
+# pylint: disable=too-many-return-statements,too-many-branches
```
dhermes: It's comforting that Pylint is at least consistent in what it thinks is "too many" branches/returns!

spanner/google/cloud/spanner/_helpers.py
tseaver: I'm not sure:  we are overriding `datetime.datetime.__new__` in Python:  that method is defined in C as a builtin, and cannot be introspected the same way.  I can't think of any other way to acoomplish the goal, either.

PR #3701, opened 2017-07-28
Language 0.26.1

lukesneeringer:
* The `encoding_type` is now optional consistently.


lukesneeringer:
Pushing this without approval because the real thing was approved and the DPE samples are broken.

PR #3702, opened 2017-07-28
Bump Speech to 0.27.1

lukesneeringer:
This release adds `WordTimeOffset`, the ability to have timestamps on the output.


lukesneeringer:
Also merging this unilaterally for the same reason as #3701.

PR #3707, opened 2017-07-31
Move 'datastore.exceptions_catch_remap_gax_error' to 'core.exceptions'.

tseaver:
Toward #3175.

Rework PR #3444 against current version of `core`.

Also, drop the unconditional dependency of `core` on `google-gax`:  we test with it, but omit the feature if it isn't present.

### DIFF
--- a/core/google/cloud/exceptions.py
+++ b/core/google/cloud/exceptions.py
@@ -20,16 +20,26 @@ See https://cloud.google.com/storage/docs/json_api/v1/status-codes
 # Avoid the grpc and google.cloud.grpc collision.
 from __future__ import absolute_import
 
+import contextlib
 import copy
+import sys
 
 import six
 
-from google.cloud._helpers import _to_bytes
-
 try:
+    from google.gax.errors import GaxError
+    from google.gax.grpc import exc_to_code
+    from grpc import StatusCode
     from grpc._channel import _Rendezvous
 except ImportError:  # pragma: NO COVER
+    _HAVE_GRPC = False
     _Rendezvous = None
+else:
+    _HAVE_GRPC = True
+
+from google.cloud._helpers import _to_bytes
+
+_HTTP_CODE_TO_EXCEPTION = {}  # populated at end of module
 
 _HTTP_CODE_TO_EXCEPTION = {}  # populated at end of module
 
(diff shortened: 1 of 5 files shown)


core/google/cloud/exceptions.py
```diff
@@ -244,8 +254,50 @@ def _walk_subclasses(klass):
...
+    try:
+        yield
+    except GaxError as exc:
+        error_code = exc_to_code(exc.cause)
+        error_class = _GRPC_ERROR_MAPPING.get(error_code)
+        if error_class is None:
+            raise
+        else:
+            new_exc = error_class(exc.cause.details())
+            six.reraise(error_class, new_exc, sys.exc_info()[2])
```
jonparrott: Isn't the `exc_info` argument optional?

Also, WDYT about `six.raise_from` here instead?

core/google/cloud/exceptions.py
```diff
@@ -20,16 +20,26 @@
...
 try:
+    from google.gax.errors import GaxError
+    from google.gax.grpc import exc_to_code
+    from grpc import StatusCode
     from grpc._channel import _Rendezvous
 except ImportError:  # pragma: NO COVER
+    _HAVE_GRPC = False
     _Rendezvous = None
+else:
+    _HAVE_GRPC = True
```
jonparrott: nit: I would prefer this to just go at the end of the `try:` block.

core/nox.py
```diff
@@ -35,6 +35,7 @@ def unit_tests(session, python_version):
         'pytest',
         'pytest-cov',
         'grpcio >= 1.0.2',
+        'google-gax >= 0.15, < 0.16.dev',
```
jonparrott: ugh I do not feel good about this, but it's fine for now. I'll remove it once I move gax's error functionality into here.

core/google/cloud/exceptions.py
tseaver: De gustibus?  ISTM that the `else:` block of a `try:` is where everything goes which is *not* going to trigger one of the handled exceptions.

core/google/cloud/exceptions.py
tseaver: > Isn't the `exc_info` argument optional?

The [`six.reraise` docs](https://pythonhosted.org/six/#six.reraise) aren't clear what happens when one passes a new class / instance but no traceback, so I chose to be explicit (it is an error case, so performance isn't really an issue).

> Also, WDYT about six.raise_from here instead?

AFAICT, [`six.raise_from`](https://pythonhosted.org/six/#six.raise_from) *would* discard the existing traceback on Python2.

core/nox.py
tseaver: Right.  We need to test with it, even if we don't make users install it.

core/google/cloud/exceptions.py
jonparrott: that's fine.

core/google/cloud/exceptions.py
jonparrott: Can we confirm?

tseaver:
@jonparrott, @dhermes We seem to have a coverage testing bug,  The [current failure](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2666) is for the code which initializes the mapping from gRPC status code to exception type.  It is always executed (because we install gax via nox), but the "alternate" / "ghost" branch is not.

The [`coverage` docs](http://coverage.readthedocs.io/en/coverage-4.2/excluding.html#branch-coverage) show an example of suppressing coverage on a non-empty branch, but the only way I can see to suppress it for the "little man on the stair" branch is to hang a `# pragma NO COVER` off the main `if:`, or else add a meaningless `else:  # pragma NO COVER` with `pass` as its suite.

tseaver:
6c2f462 gets the incantation right (the `:` after `pragma`).

tseaver:
This PR is invalidated by abfec7008e6040136a16a3a0f7dc72b5579c721e, and should be re-done against the current master (with the new target for `_catch_remap_gax_error` in `google.api.core.exceptions`, I think.  @jonparrott do you concur?

jonparrott:
@tseaver yes - eventually gax will raise the exceptions in `google.api.core` and then eventually gax will cease to exist.

/cc @lukesneeringer

Issue #3711, opened 2017-08-01
Storage: there is no way to remove labels from a bucket


jonparrott:
Attempting to remove a label from a bucket does not appear to work:

```python
    # {'example': 'label'}
    labels = bucket.labels

    if 'example' in labels:
        del labels['example']

    # {}
    bucket.labels = labels

    bucket.patch()

    # Still {'example': 'label'}
    pprint.pprint(bucket.labels)
```

tseaver:
After stopping inside the `patch` call with the empty labels:

```python
(Pdb) pp update_properties
{'labels': {}}
```

stepping over the API call leaves an `api_response`:

```python
(Pdb) pp api_response
{'acl': [{'bucket': 'test-bucket-patch-1501616216661',
          'entity': 'project-owners-455332057385',
          'etag': 'CAM=',
          'id': 'test-bucket-patch-1501616216661/project-owners-455332057385',
          'kind': 'storage#bucketAccessControl',
          'projectTeam': {'projectNumber': '455332057385', 'team': 'owners'},
          'role': 'OWNER',
          'selfLink': 'https://www.googleapis.com/storage/v1/b/test-bucket-patch-1501616216661/acl/project-owners-455332057385'},
         {'bucket': 'test-bucket-patch-1501616216661',
          'entity': 'project-editors-455332057385',
          'etag': 'CAM=',
          'id': 'test-bucket-patch-1501616216661/project-editors-455332057385',
          'kind': 'storage#bucketAccessControl',
          'projectTeam': {'projectNumber': '455332057385', 'team': 'editors'},
          'role': 'OWNER',
          'selfLink': 'https://www.googleapis.com/storage/v1/b/test-bucket-patch-1501616216661/acl/project-editors-455332057385'},
         {'bucket': 'test-bucket-patch-1501616216661',
          'entity': 'project-viewers-455332057385',
          'etag': 'CAM=',
          'id': 'test-bucket-patch-1501616216661/project-viewers-455332057385',
          'kind': 'storage#bucketAccessControl',
          'projectTeam': {'projectNumber': '455332057385', 'team': 'viewers'},
          'role': 'READER',
          'selfLink': 'https://www.googleapis.com/storage/v1/b/test-bucket-patch-1501616216661/acl/project-viewers-455332057385'}],
 'defaultObjectAcl': [{'entity': 'project-owners-455332057385',
                       'etag': 'CAM=',
[...]

Issue #1430, opened 2016-01-29
API unmapped:  storage.buckets.update


tseaver:
Noted in #1420: https://github.com/GoogleCloudPlatform/gcloud-python/pull/1420/files#diff-50dad309c06e8edd83f86b8e40b32a85R287

dhermes:
@tseaver The point of the comment is that we don't need to implement it since `patch` exists.

tseaver:
The `storageClass` property can apparently be set via `update` but not `patch`, but we expose a custom API which does that now via a rewrite.

PR #3714, opened 2017-08-01
Add 'update' API wrapper for buckets/blobs.

tseaver:
Turns out some properties (i.e., `labels`, see #3711) behave differently under [patch semantics](https://cloud.google.com/storage/docs/json_api/v1/how-tos/performance#patch), which makes `update` useful.

See #1430, which was closed as "not needed".


tseaver:
Squash-merged with #3715

PR #3719, opened 2017-08-02
Do not show Trace documentation.

### LINKED ISSUE #3717: Stackdriver Trace API should not yet be surfaced in documentation
https://google-cloud-python.readthedocs.io/en/latest/trace/index.html

Should not yet be there. We are not yet exposing gapic-only clients to users. This library was added to support a separate stackdriver tracing library and not intended for direct consumption by users.

lukesneeringer:
Fixes #3717.

@dhermes, can you merge this and manually update docs?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,6 +33,7 @@ The following client libraries have **beta** support:
 
 -  `Google BigQuery`_ (`BigQuery README`_)
 -  `Google Cloud Natural Language`_ (`Natural Language README`_)
+-  `Google Cloud Speech`_ (`Speech README`_)
 -  `Google Cloud Video Intelligence`_ (`Video Intelligence README`_)
 -  `Google Cloud Vision`_ (`Vision README`_)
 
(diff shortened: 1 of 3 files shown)


tseaver:
@lukesneeringer Just removing it from the TOX breaks the docs build, because we run with `-W`, converting the "not included file" warning to an error.  We need to remove `docs/trace` as well.

jonparrott:
We can either remove `docs/trace` or add the `:orphan:` marker to its index.

lukesneeringer:
Any preference between those?

jonparrott:
preference for orphan, as it still makes sure those pages build without error.

dhermes:
But @jonparrott they will still show up in the docs if someone knows the URL / path

jonparrott:
I'm okay with that for the time being.

dhermes:
OK cool. (I just realized it was you and not @lukesneeringer that filed #3717.)

lukesneeringer:
Okay, this should fix the build. Also noticed Speech was still alpha in `README.rst`, so I sent that up along with it.

PR #3721, opened 2017-08-02
Add properties for new job statistics

### LINKED ISSUE #3706: How to retrieve 'totalBytesBilled'
1. OS type and version: Mac OS Sierra
2. Python version and virtual environment information `python --version`: Python 2.7.13
3. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`: google-cloud-bigquery==0.25.0 and google-cloud-core==0.25.0

Issue: Trying to monitor costs of queries on BigQuery by accessing the 'totalBytesBilled' field which is available through the HTTP API (jobs[].statistics.query.totalBytesBilled, https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/list). Is there a way to get this statistic through the Python Client? I've found you can access this through job._properties and was curious if there is a public API to access this. 

Thank you!

tseaver:
Closes #3706.

Note for reviewers:  commit-by-commit might be an easier path.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -83,6 +83,17 @@ class Client(ClientWithProject):
             project=project, credentials=credentials, _http=_http)
         self._connection = Connection(self)
 
+    def _clone(self, project):
+        """Create a new client for another project.
+
+        Helper for creating dataset / table instances in remote projects.
+
+        :rtype: :class:`Client`
+        :returns: a new instance, bound to the supplied project, using
+                  the same credentials / http object as this instance.
+        """
+        return self.__class__(project, self._credentials, self._http)
+
     def list_projects(self, max_results=None, page_token=None):
         """List projects for the project associated with this client.
 
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -334,6 +334,11 @@ class _AsyncJob(google.cloud.future.polling.PollingFuture):
         # For Future interface
         self._set_future_result()
 
+    def _job_statistics(self):
+        """Helper for properties derived from job statistics."""
+        statistics = self._properties.get('statistics', {})
+        return statistics.get(self._JOB_TYPE, {})
+
     @classmethod
     def _get_resource_config(cls, resource):
         """Helper for :meth:`from_api_repr`
@@ -965,11 +970,6 @@ class ExtractTableToStorageJob(_AsyncJob):
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.extract.printHeader
     """
 
-    def _extract_statistics(self):
-        """Helper for properties derived from extract job statistics."""
-        statistics = self._properties.get('statistics', {})
-        return statistics.get('extract', {})
-
     @property
     def destination_uri_file_counts(self):
         """Return file counts from job statistics, if present.
@@ -981,7 +981,7 @@ class ExtractTableToStorageJob(_AsyncJob):
         :returns: number of DML rows affectd by the job, or None if job is not
                   yet complete.
         """
-        query_stats = self._extract_statistics()
+        query_stats = self._job_statistics()
         return query_stats.get('destinationUriFileCounts')
 
     def _populate_config_resource(self, configuration):
@@ -1297,11 +1297,6 @@ class QueryJob(_AsyncJob):
         job._set_properties(resource)
         return job
 
-    def _query_statistics(self):
-        """Helper for properties derived from query statistics."""
-        statistics = self._properties.get('statistics', {})
-        return statistics.get('query', {})
-
     @property
     def query_plan(self):
         """Return query plan from job statistics, if present.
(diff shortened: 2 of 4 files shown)


jonparrott:
I'm going to defer to @tswast for this.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1277,6 +1278,22 @@ def from_api_repr(cls, resource, client):
...
         return job
 
+    @property
+    def query_plan(self):
+        """Return query plan from job statistics, if present.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.query.queryPlan
+
+        :rtype: list of dict
```
tswast: Feels a bit odd to me to return a dict here rather than a new object, but seems okay for the first pass.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1277,6 +1297,170 @@ def from_api_repr(cls, resource, client):
         job._set_properties(resource)
         return job
 
+    @property
+    def query_plan(self):
```
tswast: Do you think we should create a separate QueryStatistics class to hold these read-only properties to match the REST API more closely?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -83,6 +83,17 @@ def __init__(self, project=None, credentials=None, _http=None):
             project=project, credentials=credentials, _http=_http)
         self._connection = Connection(self)
 
+    def _clone(self, project):
```
tswast: Why do we need this method?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1354,6 +1354,42 @@ def cache_hit(self):
...
+        query_stats = self._query_statistics()
+        clients_by_project = {client.project: client}
+        datasets_by_project_name = {}
+
+        for table in query_stats.get('referencedTables', ()):
+
+            t_project = table['projectId']
+            t_client = clients_by_project.get(t_project)
+            if t_client is None:
+                t_client = client._clone(t_project)
```
tswast: Is there any way we can have the table accept a project ID instead of having to create a new client? It will be useful for end-users to be able to list table data on a different project than the one they use for billing.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1390,6 +1390,20 @@ def referenced_tables(self):
 
         return tables
 
+    @property
+    def num_dml_affected_rows(self):
```
tswast: FYI: `numDmlAffectedRows` is one of few statistics that is also in QueryResults, which you added in https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2920

`totalBytesProcessed` and `cacheHit` also appear in QueryResults, but honestly I'd prefer people look at them on the Job resource.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1391,6 +1391,20 @@ def referenced_tables(self):
         return tables
 
     @property
+    def schema(self):
+        """Return schema from job statistics, if present.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.query.schema
+
+        :rtype: list of :class:`~google.cloud.bigquery.schema.SchemaField
```
tswast: LGTM, but this reminds of an issue that has come up in some questions I've answered in GCP Slack of StackOverflow. Filed https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3723 for a public way to get to/from the raw API representation of a schema.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1418,6 +1418,21 @@ def num_dml_affected_rows(self):
...
         return query_stats.get('numDmlAffectedRows')
 
+    @property
+    def undeclared_query_paramters(self):
+        """Return undeclared query parameters from job statistics, if present.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.query.undeclaredQueryParamters
+
+        :rtype: list of dict
```
tswast: I'd prefer to have this return a list of AbstractQueryParameter https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/_helpers.py#L362

bigquery/google/cloud/bigquery/job.py
tswast: Looks like we should be able to set the project on the dataset without having to clone the client. https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/dataset.html

bigquery/google/cloud/bigquery/client.py
```diff
@@ -83,6 +83,17 @@ def __init__(self, project=None, credentials=None, _http=None):
...
+        """Create a new client for another project.
+
+        Helper for creating dataset / table instances in remote projects.
+
+        :rtype: :class:`Client`
+        :returns: a new instance, bound to the supplied project, using
+                  the same credentials / http object as this instance.
+        """
+        return self.__class__(project, self._credentials, self._http)
```
dhermes: 1. Please use keyword arguments for all keywords (recall the issue we had when a GAPIC surface changed)
1. `self._http` is probably not what you want, rather `self._http_internal`

tseaver:
@tswast I'm going to break this up into separate PRs, targeting the `bigquery-b2` branch.

PR #3722, opened 2017-08-03
Reuse explicit credentials when creating 'database.spanner_api'.

### LINKED ISSUE #3718: Spanner ignores provided credentials when executing a query.
(This bug is re-posted from an e-mail I received yesterday.)

----------------------

We are using python spanner client libraries, and I have a credentials issue in executing a sql query. Seems like executing a sql ignores the provided credentials to the client and just try to use the default credentials.

The code looks like:

```python
spanner_client = spanner.Client(credentials=credentials, project=project_id)
instance = spanner_client.instance(instance_id)
database = instance.database(database_id)
database.execute_sql(stmt)
```

Debugging the code for a bit, I found out that eventually the code [will try to create a session](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/google/cloud/spanner/session.py#L97) and it will [try to get the spanner API client](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/google/cloud//spanner/database.py#L157). When this object created, credentials is not passed and hence it's `None`, and hence the SpannerClient is trying to get the default credentials and fails in my case because I do not have default creds.

tseaver:
- Preserves "custom" credentials (existing code worked only with implicit credentials from the environment).
- Add tests ensuring scopes are set for correctly for all GAX apis (client uses admin scope, which do not grant data access, while database uses data scope, which does not grant admin access).

Closes #3718.

### DIFF
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -158,13 +158,13 @@ class Database(object):
     def spanner_api(self):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
-            base_creds = self._instance._client.credentials
-            scoped = google.auth.credentials.with_scopes_if_required(
-                base_creds, (SPANNER_DATA_SCOPE,))
+            credentials = self._instance._client.credentials
+            if isinstance(credentials, google.auth.credentials.Scoped):
+                credentials = credentials.with_scopes((SPANNER_DATA_SCOPE,))
             self._spanner_api = SpannerClient(
                 lib_name='gccl',
                 lib_version=__version__,
-                credentials=scoped,
+                credentials=credentials,
             )
         return self._spanner_api
 
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -159,7 +159,8 @@ class Database(object):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
             base_creds = self._instance._client.credentials
-            scoped = base_creds.with_scopes((SPANNER_DATA_SCOPE,))
+            scoped = google.auth.credentials.with_scopes_if_required(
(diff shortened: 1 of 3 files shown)


spanner/google/cloud/spanner/database.py
```diff
@@ -154,8 +158,13 @@ def ddl_statements(self):
     def spanner_api(self):
         """Helper for session-related API calls."""
         if self._spanner_api is None:
+            base_creds = self._instance._client.credentials
+            scoped = base_creds.with_scopes((SPANNER_DATA_SCOPE,))
```
jonparrott: you need to check if these are scoped credentials. It's possible to use spanner with JWT credentials which are unscoped:

```python
if instance(base_creds, google.auth.credentials.Scoped):
    scoped = base_creds.with_scopes((SPANNER_DATA_SCOPE,))
```

tseaver: @jonparrott Would using `google.auth.credentials.with_scopes_if_required` work instead?

jonparrott: I checked, and not in this case. the base client would've already scoped the credentials and `requires_scopes` will return false.

jonparrott:
Github buried my comment, re-posting:

>  checked, and not in this case. the base client would've already scoped the credentials and requires_scopes will return false

`with_scopes_if_required` won't work for this case.

tseaver:
@jonparrott Who actually implements `with_scopes`?  I can't find a non-raising concrete implementation except in `google.auth.app_engine`.

jonparrott:
@tseaver  http://google-auth.readthedocs.io/en/latest/_modules/google/oauth2/service_account.html#Credentials.with_scopes

tseaver:
@jonparrott See cf5b2dfa079f51d99d541daac1879ccb5053934f

PR #3724, opened 2017-08-04
Add a function to complain about obselete packages.

lukesneeringer:
This will be used to complain if `gapic-google-cloud-{api}-{version}` or `proto-google-cloud-{api}-{version}` are still installed after they are vendored (for GA libraries only).

### DIFF
--- a/core/google/cloud/obselete.py
+++ b/core/google/cloud/obselete.py
@@ -1,0 +1,38 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/google/cloud/obselete.py
```diff
@@ -0,0 +1,38 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pkg_resources
+import warnings
+
+
+def complain(package_name):
```
jonparrott: s/package/distribution through this.

PR #3735, opened 2017-08-07
BigQuery: Remove client-side enum validation.

### LINKED ISSUE #3725: job.py in bigquery does unnecessary client-side validation of data formats
The BigQuery engineering team may decide to expand the allowed formats for import and export, but the allowed values are hard-coded client side. Imports are limited to CSV, DATASTORE_BACKUP, NEWLINE_DELIMITED_JSON, AVRO, for example.

This means that if the BigQuery starts to allow other formats, users will be unable to use those new formats without updating their client. Moreover, there is no way for BigQuery to roll out new formats to a whitelisted set of users. 
(note this is not mere idle speculation and is currently impacting real users alpha testing a new format)

See code here:
[https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L116][1]
and 
[https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L139][2]

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6d7fae96d25a0a468683a5ab25a2f48014a16378/bigquery/google/cloud/bigquery/job.py#L116
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6d7fae96d25a0a468683a5ab25a2f48014a16378/bigquery/google/cloud/bigquery/job.py#L139

lukesneeringer:
Fixes #3725.

### DIFF
--- a/bigquery/tests/unit/test__helpers.py
+++ b/bigquery/tests/unit/test__helpers.py
@@ -765,7 +765,7 @@ class Test_EnumProperty(unittest.TestCase):
     def test_it(self):
 
         class Sub(self._get_target_class()):
-            ALLOWED = ('FOO', 'BAR', 'BAZ')
+            pass
 
         class Configuration(object):
             _attr = None
@@ -777,9 +777,6 @@ class Test_EnumProperty(unittest.TestCase):
                 self._configuration = Configuration()
 
         wrapper = Wrapper()
-        with self.assertRaises(ValueError):
-            wrapper.attr = 'BOGUS'
-
         wrapper.attr = 'FOO'
         self.assertEqual(wrapper.attr, 'FOO')
         self.assertEqual(wrapper._configuration._attr, 'FOO')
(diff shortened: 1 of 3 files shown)


bigquery/tests/unit/test__helpers.py
```diff
@@ -765,7 +765,7 @@ def _get_target_class():
     def test_it(self):
 
         class Sub(self._get_target_class()):
-            ALLOWED = ('FOO', 'BAR', 'BAZ')
+            pass
```
dhermes: You can probably get rid of `Sub` altogether?

tseaver: Without `_validate`, there really isn't any point in the `_EnumProperty` base class at all:  the concrete classes could just derived from `_ConfigurationProperty, leaving the class-level constants (but then, why, if we aren't going to enforce them?)

PR #3737, opened 2017-08-07
Storage: Track deleted labels; make Bucket.patch() send them.

### LINKED ISSUE #3711: Storage: there is no way to remove labels from a bucket

lukesneeringer:
Uses #3715 as a base.

This PR adds tracking for deleted labels on the `Bucket` object, and causes the appropriate data to be sent when `Bucket.patch` is called.

Fixes #3711.

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -142,6 +142,8 @@ class _PropertyMixin(object):
         # to work properly w/ 'noAcl'.
         update_properties = {key: self._properties[key]
                              for key in self._changes}
+
+        # Make the API call.
         api_response = client._connection.api_request(
             method='PATCH', path=self.path, data=update_properties,
             query_params={'projection': 'full'}, _target_object=self)
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -126,6 +126,8 @@ class _PropertyMixin(object):
         self._properties = value
         # If the values are reset, the changes must as well.
         self._changes = set()
+        if hasattr(self, '_label_removals'):
+            self._label_removals.clear()
 
     def patch(self, client=None):
         """Sends all changed properties in a PATCH request.
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -165,22 +165,6 @@ class _PropertyMixin(object):
             query_params={'projection': 'full'}, _target_object=self)
         self._set_properties(api_response)
 
-    def update(self, client=None):
-        """Sends all properties in a PUT request.
-
-        Updates the ``_properties`` with the response from the backend.
-
-        :type client: :class:`~google.cloud.storage.client.Client` or
-                      ``NoneType``
-        :param client: the client to use.  If not passed, falls back to the
-                       ``client`` stored on the current object.
-        """
-        client = self._require_client(client)
-        api_response = client._connection.api_request(
-            method='PUT', path=self.path, data=self._properties,
-            query_params={'projection': 'full'}, _target_object=self)
-        self._set_properties(api_response)
-
 
 def _scalar_property(fieldname):
     """Create a property descriptor around the :class:`_PropertyMixin` helpers.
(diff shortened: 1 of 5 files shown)


lukesneeringer:
Grumble grumble grumble @googlebot.

tseaver:
#3715 merged, so you can rebase to appease googlebot.

lukesneeringer:
Done, just need a review (and to fix CI).

storage/google/cloud/storage/bucket.py
```diff
@@ -199,6 +200,47 @@ def create(self, client=None):
...
+
+        Updates the ``_properties`` with the response from the backend.
+
+        :type client: :class:`~google.cloud.storage.client.Client` or
+                      ``NoneType``
+        :param client: the client to use.  If not passed, falls back to the
+                       ``client`` stored on the current object.
+        """
+        answer = super(Bucket, self).update(client=client)
+        self._label_removals.clear()
```
tseaver: Clearing `self._label_removals` would be better done inside `_set_properties` (so that it get cleared during a `reload` too.

storage/tests/unit/test_bucket.py
```diff
@@ -710,6 +710,43 @@ def test_labels_setter(self):
...
 
+        # Make sure that a update call correctly adds the labels.
+        client = mock.NonCallableMock(spec=('_connection',))
+        client._connection = mock.NonCallableMock(spec=('api_request',))
+        bucket.update(client=client)
+        client._connection.api_request.assert_called()
+        _, _, kwargs = client._connection.api_request.mock_calls[0]
+        self.assertEqual(len(kwargs['data']['labels']), 2)
+        self.assertEqual(kwargs['data']['labels']['color'], 'red')
+        self.assertEqual(kwargs['data']['labels']['flavor'], 'cherry')
```
tseaver: The extra call to `update` feels out of place for this test:  unit test testcase methods should limit scope to the thing-under-test.

storage/google/cloud/storage/bucket.py
lukesneeringer: I considered that, but it has the problematic side effect that multiple calls to write labels would not do the right thing:

```python
bucket.labels  # {'foo': ..., 'bar': ..., 'baz': ...}
bucket.labels = {'bar': ..., 'baz': ..., 'bacon': ...}
bucket._label_removals  # {'foo'}
bucket.labels = {'baz': ..., 'bacon': ...}
bucket._label_removals  # {'bar'}; should be {'foo', 'bar'}
```

storage/tests/unit/test_bucket.py
lukesneeringer: I could make a separate test to get coverage on `update`; it just feels a bit redundant.

storage/google/cloud/storage/bucket.py
tseaver: Like `reload`, the "super" versions of `patch` and `update` both call `_set_properties`, passing the result of the call (the server-side state).  Preserving `_label_removals` across *any* of those calls is a bug.  That is separate from the work you are doing to preserve it in successive assignements to `labels`.

lukesneeringer: Ahh, I understand now. :-)

lukesneeringer:
Thanks for catching that @tseaver. Updates made.

storage/google/cloud/storage/_helpers.py
```diff
@@ -126,6 +126,8 @@ def _set_properties(self, value):
         self._properties = value
         # If the values are reset, the changes must as well.
         self._changes = set()
+        if hasattr(self, '_label_removals'):
+            self._label_removals.clear()
```
tseaver: Maybe this is over-picky, but ISTM it make more sense to override `_set_properties` and do this work there.

lukesneeringer: I think this is a case where either way is reasonable.

lukesneeringer:
@tseaver @jonparrott Can someone give this a final signoff so I can merge it? :-)

PR #3738, opened 2017-08-07
Add google.api.core.exceptions

jonparrott:


core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    grpc_status_code = (
+        grpc.StatusCode.UNAUTHENTICATED if grpc is not None else None)
+
+
+class Forbidden(ClientError):
+    """Exception mapping a ``403 Forbidden`` response."""
+    code = http_client.FORBIDDEN
+
+
+class PermissionDenied(Forbidden):
```
dhermes: Why the distinction here between `PermissionDenied` and `Forbidden`?

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+class PreconditionFailed(ClientError):
+    """Exception mapping a ``412 Precondition Failed`` response."""
+    code = http_client.PRECONDITION_FAILED
+
+
+class RequestRangeNotSatisfiable(ClientError):
+    """Exception mapping a ``416 Request Range Not Satisfiable`` response."""
+    # There is not a consistent http_client constant for this.
```
dhermes: Go on?

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    code = 416
+
+
+class TooManyRequests(ClientError):
+    """Exception mapping a ``429 Too Many Requests`` response."""
+    # There is not a consistent http_client constant for this.
```
dhermes: i.e. it's present in Python 3.5 but not 3.4 or earlier?

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+        GoogleAPICallError: An instance of the appropriate subclass of
+            :class:`GoogleAPICallError`, with the message and errors populated
+            from the response.
+    """
+    try:
+        payload = response.json()
+    except ValueError:
+        payload = {'error': {'message': response.text or 'unknown error'}}
+
+    error_message = payload.get('error', {}).get('message', 'unknown error')
```
dhermes: You hard-code `unknown error` twice here. Perfect candidate for a local variable or a global (prefer global)

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    errors = payload.get('error', {}).get('errors', ())
+
+    message = '{method} {url}: {error}'.format(
+        method=response.request.method,
+        url=response.request.url,
+        error=error_message)
+
+    exception = from_http_status(
+        response.status_code, message, errors=errors)
+    exception.response = response
```
dhermes: Instead of monkey-patching, why not make `response` part of the constructor?

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    Returns:
+        GoogleAPICallError: An instance of the appropriate subclass of
+            :class:`GoogleAPICallError`.
+    """
+    error_class = exception_class_for_grpc_status(status_code)
+    error = error_class(message, errors)
+
+    if error.grpc_status_code is None:
+        error.grpc_status_code = status_code
+
```
dhermes: It seems you want to keep around the `_Rendezvous` that caused the exception here.

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    """Create a :class:`GoogleAPICallError` from a :class:`grpc.RpcError`.
+
+    Args:
+        rpc_exc (grpc.RpcError): The gRPC error.
+
+    Returns:
+        GoogleAPICallError: An instance of the appropriate subclass of
+            :class:`GoogleAPICallError`.
+    """
+    if isinstance(rpc_exc, grpc.Call):
```
dhermes: Catching `grpc.Call` is preferred over `_Rendezvous` (which we have aliased as `google.cloud.exceptions.GrpcRendezvous`)?

core/tests/unit/api_core/test_exceptions.py
```diff
@@ -0,0 +1,159 @@
+# Copyright 2014 Google Inc.
```
dhermes: Ditto on 2014 vs. 2017 here

core/google/api/core/exceptions.py
lukesneeringer: I would honestly prefer it as-is.

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+
+# Lookup tables for mapping exceptions from HTTP and gRPC transports.
+# Populated by _APICallErrorMeta
+_HTTP_CODE_TO_EXCEPTION = {}
+_GRPC_CODE_TO_EXCEPTION = {}
+
+
+class GoogleAPIError(Exception):
+    """Base class for all exceptions raised by Google API Clients."""
+    pass
```
lukesneeringer: Explicit is better than implicit. :-O

core/google/api/core/exceptions.py
dhermes: ORLY? Why is that? Over time, those lines could drift apart, then the "duh it's the same" will no longer be obvious.

lukesneeringer: I find constants have the opposite problem; it just takes effort to go back and check the value, see if things match, etc.

Where to toe the line is a balancing act.

tseaver:
[Tests fail due to overage gap](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2749).

jonparrott:
I *think* I addressed all comments. Test coverage also fixed, thanks @tseaver.

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+
+class BadRequest(ClientError):
+    """Exception mapping a ``400 Bad Request`` response."""
+    code = http_client.BAD_REQUEST
+
+
+class InvalidArgument(BadRequest):
+    """Exception mapping a :prop:`grpc.StatusCode.INVALID_ARGUMENT` error."""
+    grpc_status_code = (
+        grpc.StatusCode.INVALID_ARGUMENT if grpc is not None else None)
```
lukesneeringer: I do not really care.

core/google/api/core/exceptions.py
```diff
@@ -0,0 +1,399 @@
...
+    grpc_status_code = None
+    """Optional[grpc.StatusCode]: The gRPC status code associated with this
+    error.
+
+    This may be ``None`` if the exception does not match up to a gRPC error.
+    """
+
+    def __init__(self, message, errors=()):
+        """
+        Args:
```
lukesneeringer: What is the rationale for this? Sphinx can document `__init__`.

core/google/api/core/exceptions.py
jonparrott: Using `getattr` would require me to import the class rather than the module, which goes against our written style guide.

core/google/api/core/exceptions.py
jonparrott: It's more or less an unwritten style rule here.

core/google/api/core/exceptions.py
```diff
@@ -87,7 +96,13 @@ def errors(self):
         Returns:
             Sequence[Any]: A list of additional error details.
         """
-        return [copy.deepcopy(error) for error in self._errors]
+        return list(self._errors)
+
+    @property
+    def response(self):
+        """Optional[Union[requests.Request, grpc.Call]]L The response or
```
dhermes: This `]L` seems an accident?

core/google/api/core/exceptions.py
dhermes: Ew, but OK

core/google/api/core/exceptions.py
jonparrott: Good catch. Fixed.

core/google/api/core/exceptions.py
lukesneeringer: I am okay with whatever.

If I were writing it, I would import the class (I agree with preferring module-level imports but I also think that there are exceptions, and this is a good one). That is _not_ intended to mean that you must do this.

If I were writing a style guide, I would have a rule saying to prefer module level imports, a rule forbidding ternary operators, and a guidance saying to have the wisdom to know when you should break the rules. ;-)

jonparrott: Agreed. I'm going to leave as-is for now. Changing this will not impact the public interface, so if we find this offensive later we can change without worry.

jonparrott:
@dhermes @lukesneeringer are we good to merge? (pending CI)?

dhermes:
LGTM

jonparrott:
Thanks all ✨  this is the first step to a unified core. :)

PR #3744, opened 2017-08-07
Allow Table.read_rows to take an inclusive end key.

### LINKED ISSUE #3592: The read_rows method in the bigtable client does not provide an interface to specify open/closed key ranges
The read_rows method of the Table class in the bigtable client is designed such that the results are inclusive of the start key but exclusive of the end key. That seems a reasonable choice, until one considers that keys may be designed to accumulate in reverse order (as recommended in the docs: https://cloud.google.com/bigtable/docs/schema-design-time-series). In that case, it is necessary to flip the natural order of the keys to process a query -i.e. if I query from time t0 to t1, I will create two corresponding keys k0 and k1, but k1 will be the start key and k0 the end key in the call to read_rows.

Currently, the specification of open and closed for the start and end key is hard-coded in the _create_row_request method, and that makes it difficult for an end user of the library to alter the default behavior. Consider adding an option to the read_rows method, or better yet, an attribute of a Table instance, such that queries can implement different behaviors with respect to the start and end keys.

lukesneeringer:
This commit adds the `end_inclusive` keyword argument, which can be explicitly passed to get `[start:end]` rather than `[start:end)`.

Fixes #3592.

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -257,7 +257,7 @@ class Table(object):
         return rows_data.rows[row_key]
 
     def read_rows(self, start_key=None, end_key=None, limit=None,
-                  filter_=None):
+                  filter_=None, end_inclusive=False):
         """Read rows from this table.
 
         :type start_key: bytes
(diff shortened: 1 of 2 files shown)


jdmargulici:
@lukesneeringer thank you for the improvement. Technically this doesn't truly fix #3592 since it is still not possible to query (start:end].

PR #3745, opened 2017-08-07
Explicitly depend on setuptools >= 34.0.0

### LINKED ISSUE #3356: Explicitly depend on setuptools
Context: #3295

We should explicitly depend on a minimum version of setuptools to get around problems installing our packages if setuptools is too old.

lukesneeringer:
Fixes #3356.

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -55,6 +55,7 @@ REQUIREMENTS = [
     'protobuf >= 3.0.0',
     'google-auth >= 0.4.0, < 2.0.0dev',
     'requests >= 2.4.0, < 3.0.0dev',
+    'setuptools >= 34.0.0',
     'six',
     'tenacity >= 4.0.0, <5.0.0dev'
 ]


simonvanderveldt:
@lukesneeringer I know this is a pretty late response :P but was `>= 34.0.0` chosen for a specific reason or was it just the current version at the time of this change?

Asking because installing 34 or later is somewhat annoying in that you need to upgrade pip to install it in many distros that are currently in use. So it would be nice if a slightly lower version could be used if that doesn't cause problems.
See https://github.com/pypa/setuptools/issues/937 for more details about this.

tseaver:
@simonvanderveldt We really can't do the work to keep "old" pip / setuptools working.  Generally, best practice is not to mix-and-match distro-managed libraries with pip-managed ones:  I always create a new virtual environment, upgrade pip and setuptools to the latest, and then install our packages, e.g.:

```bash
$ python3.6 -m venv /tmp/gcp/3745
$ /tmp/gcp/3745/bin/pip install --upgrade setuptools pip
$ /tmp/gcp/3745/bin/pip install google-cloud-<apilib>
```

PR #3748, opened 2017-08-07
Bump requests minimum bound to 2.18.0

### LINKED ISSUE #3747: Google cloud storage needs requests >= 2.18.0 but the setup.py only requires >= 2.4.0
According to the requests [changelog](http://docs.python-requests.org/en/latest/community/updates/) the Response class was turned into a context manager only in version 2.18.0. However in google/resumable_media/requests/download.py line 62 (inside _write_to_stream) the code tries to use "response" as a context manager. Some digging reveals that this response makes its way from Client._http which is defined in google/cloud/client.py and gets the Response object from google/auth/transport/requests.py:AuthorizedSession.

So... it definitely feels like something in here needs to require requests >= 2.18.0 (maybe all of them?). 

Thoughts?

Detailed info:

1. OS type and version: N/A
2. Python version and virtual environment information: 2.7 but I don't think it matters.
3. google-cloud-python version: google-cloud-storage v 1.3.1
4. Stacktrace if available: below.
5. Steps to reproduce: Install a version of the requests library _less than 2.18.0_. Get some blob that exists. Try blob.download_as_string(). I personally re-pro'd this with version 2.10.0 to check. Note that the setup.py only requires 2.0.0 for storage and 2.4.0 for client, etc...

-- PG.

lukesneeringer:
Fixes #3747.

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -54,7 +54,7 @@ REQUIREMENTS = [
     'googleapis-common-protos >= 1.3.4',
     'protobuf >= 3.0.0',
     'google-auth >= 0.4.0, < 2.0.0dev',
-    'requests >= 2.4.0, < 3.0.0dev',
+    'requests >= 2.18.0, < 3.0.0dev',
     'six',
     'tenacity >= 4.0.0, <5.0.0dev'
 ]
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -54,7 +54,7 @@ REQUIREMENTS = [
     'google-cloud-core >= 0.26.0, < 0.27dev',
     'google-auth >= 1.0.0',
     'google-resumable-media >= 0.2.1',
-    'requests >= 2.0.0',
+    'requests >= 2.18.0',
 ]
 
 setup(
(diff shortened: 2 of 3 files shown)


core/setup.py
```diff
@@ -54,7 +54,7 @@
     'googleapis-common-protos >= 1.3.4',
     'protobuf >= 3.0.0',
     'google-auth >= 0.4.0, < 2.0.0dev',
-    'requests >= 2.4.0, < 3.0.0dev',
+    'requests >= 2.18.0, < 3.0.0dev',
```
dhermes: @jonparrott Does this upper bound make sense here?

dhermes:
@lukesneeringer Y U NO `rebase`?

core/setup.py
lukesneeringer: Yes. requests 2.0 was a significant breaking change internally.

lukesneeringer:
> @lukesneeringer Y U NO rebase?

Because squash and merge makes it irrelevant.

lukesneeringer:
Also, merging this without waiting on CI. If it was going to fail for some silly syntax problem, it would have done so already.

dhermes:
@lukesneeringer That also means merging it without dropping `requests` from `storage/setup.py` and `bigquery/setup.py`

PR #3751, opened 2017-08-08
Allow empty indexed arrays.

### LINKED ISSUE #3152: ValueError: For an array_value, subvalues must either all be indexed or all excluded from indexes.

lukesneeringer:
Fixes #3152.


datastore/tests/unit/test_helpers.py
```diff
@@ -135,6 +135,27 @@ def test_mismatched_value_indexed(self):
         with self.assertRaises(ValueError):
             self._call_fut(entity_pb)
 
+    def test_index_mismatch_ignores_empty_list(self):
+        from google.cloud.proto.datastore.v1 import entity_pb2
+        from google.cloud.datastore.helpers import _new_value_pb
+
+        _PROJECT = 'PROJECT'
+        _KIND = 'KIND'
+        _ID = 1234
```
dhermes: Just for fun: can you make this "normal" names for locals, i.e. `project`, `kind` and `id_`?

datastore/tests/unit/test_helpers.py
```diff
@@ -135,6 +135,27 @@ def test_mismatched_value_indexed(self):
...
+        entity_pb = entity_pb2.Entity()
+        entity_pb.key.partition_id.project_id = _PROJECT
+        entity_pb.key.path.add(kind=_KIND, id=_ID)
+
+        array_val_pb = _new_value_pb(entity_pb, 'baz')
+        array_pb = array_val_pb.array_value.values
+
+        # unindexed_value_pb1 = array_pb.add()
+        # unindexed_value_pb1.integer_value = 10
```
dhermes: Please remove commented out code.

datastore/tests/unit/test_helpers.py
```diff
@@ -135,6 +135,27 @@ def test_mismatched_value_indexed(self):
...
+
+        entity = self._call_fut(entity_pb)
+        entity_dict = dict(entity)
+        self.assertIsInstance(entity_dict['baz'], list)
```
dhermes: Why not just check equality for what the value is?

datastore/tests/unit/test_helpers.py
lukesneeringer: Every single other test in this module uses those attribute names.

lukesneeringer:
Closing and re-opening.

PR #3753, opened 2017-08-08
session.run_in_transaction returns the callback's return value.

### LINKED ISSUE #3493: Allow returning values from transactional functions
Somebody asks in the Slack:

> Question about spanner python APIs, specifically transactions. 
> In a read-write transaction, is it possible to return the updated
> value without hitting the database again?
> 
> let's say I have the following transaction (pretty much the example from the docs)
> ```def f(transaction):
>     second_album_result = transaction.read(
>             table='Albums', columns=('MarketingBudget',),
>             keyset=second_album_keyset, limit=1)
>     second_album_row = list(second_album_result)[0]
>     second_album_budget = second_album_row[0]
>     second_album_budget += 10000000
>     transaction.update(table=mytable, columns=('MarketingBudget'),
>                         keyset=second_album_keyset, values=[(second_album_budget)]
>                     )  
>     return second_album_budget
> database.run_in_transaction(f)
> ```
> is the `return` at all possible?

I believe that this is currently not possible in the Python library, because when you pass a closure to `run_in_transaction`, the return value always gets ignored, and it just returns the commit timestamp.

However this _is_ possible in the Cloud Spanner Java library, and being be able to return whatever you want from the body of a transactional function in the case that a transaction successfully commits is extremely useful.

lukesneeringer:
This makes it so `session.run_in_transaction` returns the callback's return value, rather than the commit timestamp.

Fixes #3493.

### DIFF
--- a/spanner/google/cloud/spanner/session.py
+++ b/spanner/google/cloud/spanner/session.py
@@ -268,8 +268,9 @@ class Session(object):
                    If passed, "timeout_secs" will be removed and used to
                    override the default timeout.
 
-        :rtype: :class:`datetime.datetime`
-        :returns: timestamp of committed transaction
+        :rtype: Any
+        :returns: The return value of ``func``.
+
         :raises Exception:
             reraises any non-ABORT execptions raised by ``func``.
         """
@@ -284,7 +285,7 @@ class Session(object):
             if txn._transaction_id is None:
                 txn.begin()
             try:
-                func(txn, *args, **kw)
+                return_value = func(txn, *args, **kw)
             except GaxError as exc:
                 _delay_until_retry(exc, deadline)
                 del self._transaction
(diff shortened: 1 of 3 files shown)


vkedia:
Why can't you return a tuple containing both the commit timestamp and the callback's return value?

lukesneeringer:
I _can_ but I am disinclined to do so. That's pretty ugly. Furthermore, it is available on the transaction object itself.

vkedia:
Sounds good. I guess if someone does need the timestamp they can return the transaction object back from the function or a tuple of (txn, retvalue)

tseaver:
FWIW, note that the transaction object is *not* available to the caller, and the `committed` field will not be set inside the callback.

vkedia:
So the transaction object is available within the callback, right? So they can always return it as the return value of the callback and outside of run_in_transaction they can invoke commit_timestamp on that?

tseaver:
@vkedia Sure, the contract between the "unit of work" callback and the caller can be whatever works.  It still feels like a clunky interface, though.

PR #3759, opened 2017-08-08
Set a 10 MB chunk size default.

### LINKED ISSUE #3494: Blob.download_to_filename eating lots of RAM
I have a large ~4 GB binary file on GCS. When I use blob.download_to_filename, my RAM usage goes increases by 3 GB temporarily. Thus behavior goes away if I explicitly set a chunk size for the blob.

This seems like bad default behavior.

lukesneeringer:
Fixes #3494.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -142,7 +142,8 @@ class Blob(_PropertyMixin):
        set as their 'storage_class'.
     """
 
-    def __init__(self, name, bucket, chunk_size=None, encryption_key=None):
+    def __init__(self, name, bucket, chunk_size=1024 * 1024 * 10,
+                 encryption_key=None):
         name = _bytes_to_unicode(name)
         super(Blob, self).__init__(name=name)
 


dhermes:
@lukesneeringer This is a very non-trivial change. It makes the default `Resumable` vs. `Multipart` **and** makes it unclear how a user might switch to `Multipart` (by explicitly setting `chunk_size = None`)

storage/google/cloud/storage/blob.py
```diff
@@ -142,7 +142,8 @@ class Blob(_PropertyMixin):
        set as their 'storage_class'.
     """
 
-    def __init__(self, name, bucket, chunk_size=None, encryption_key=None):
+    def __init__(self, name, bucket, chunk_size=1024 * 1024 * 10,
```
dhermes: @lukesneeringer I would prefer a named constant:

```python
_TEN_MB = 1024 * 1024 * 10
# or just set it to 10485760
```

lukesneeringer:
Aha. That sounds important. You and Jon both seemed to be fine with it on #3494.
Is the correct answer still to change the default, and just document accordingly?

dhermes:
@lukesneeringer #3703 was the same as #3494 and #3713 fixed it

lukesneeringer:
Ah. Thanks.

PR #3763, opened 2017-08-08
Change docs link from stable/ to latest/

tswast:
Per https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3595#issuecomment-321019971 `latest` is more frequently updated and `stable` is meaningless.

### DIFF
--- a/index.html
+++ b/index.html
@@ -92,7 +92,7 @@
     <div class="container">
       <ul class="featuring-links">
         <li>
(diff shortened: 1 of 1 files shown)


dhermes:
This LGTM though I'd like @jonparrott and @lukesneeringer to weigh in

lukesneeringer:
I also agree.

PR #3765, opened 2017-08-08
Fix __eq__ and __ne__.

### LINKED ISSUE #3455: __eq__ and __ne__ are implemented incorrectly for all classes
`__eq__` and `__ne__` are implemented incorrectly throughout the entire `google-cloud-*` code base. I encountered this issue when trying to compare a `google.cloud.datastore.Key` instance to `unittest.mock.ANY`.

When implementing custom equality in Python (2 or 3), it should be written like this:

```python
class A:
    def __eq__(self, other):
        if not isinstance(other, A):
            # Delegate comparison to the other instance's __eq__.
            return NotImplemented
        # Whatever logic applies to equality of instances of A can be added here.
        ...

    def __ne__(self, other):
        # By using the == operator, the returned NotImplemented is handled correctly.
        return not self == other
```

Throughout the entire code base of this repository, equality is implemented like this:

```python
class A:
    def __eq__(self, other):
        if not isinstance(other, A):
            # Other instances are never equal.
            return False
        # Whatever logic applies to equality of instances of A can be added here.
        ...

    def __ne__(self, other):
        # By negating the return value, NotImplemented is treated as False.
        return not self.__eq__(other)
```

As a result gcloud instances can never equal entities of other classes if they are the first object in the comparison. This is not an issue for most cases, but there are cases where this **is** an issue. Also this behaviour is simply incorrect.

The `__ne__` implementation works, because `__eq__` is implemented incorrectly. The boolean value of `NotImplemented` is `True`.

```python
>>> not NotImplemented
False
```

This is typically an issue when unittesting.

```python
>>> from unittest.mock import ANY
>>> 
>>> from google.cloud.datastore import Client
>>> 
>>> 
>>> client = Client()
>>> key = client.key('foo')
>>> key == ANY  # Expected True
False
>>> ANY == key
True
>>> key != ANY  # Expected False
True
>>> ANY != key
False
>>> 
```

lukesneeringer:
This updates `__eq__` to do type-checking correctly on all classes (except mock classes defined in unit test modules):

```
def __eq__(self, other):
    if not isinstance(other, type(self)):
        return NotImplemented
```

It also updates `__ne__` to be _exactly_:

```
def __ne__(self, other):
    return not self == other
```

Fixes #3455.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -328,9 +328,14 @@ class UDFResource(object):
         self.value = value
 
     def __eq__(self, other):
+        if not isinstance(other, UDFResource):
+            return NotImplemented
         return(
             self.udf_type == other.udf_type and
             self.value == other.value)
+
+    def __ne__(self, other):
+        return not self == other
 
 
 class UDFResourcesProperty(object):
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -101,16 +101,12 @@ class SchemaField(object):
         )
 
     def __eq__(self, other):
-        if isinstance(other, SchemaField):
-            return self._key() == other._key()
-        else:
+        if not isinstance(other, SchemaField):
             return NotImplemented
+        return self._key() == other._key()
 
     def __ne__(self, other):
-        if isinstance(other, SchemaField):
-            return self._key() != other._key()
-        else:
-            return NotImplemented
+        return not self == other
 
     def __hash__(self):
         return hash(self._key())
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -76,10 +76,15 @@ class AccessGrant(object):
         self.entity_id = entity_id
 
     def __eq__(self, other):
+        if not isinstance(other, AccessGrant):
+            return NotImplemented
         return (
             self.role == other.role and
             self.entity_type == other.entity_type and
             self.entity_id == other.entity_id)
+
+    def __ne__(self, other):
+        return not self == other
 
     def __repr__(self):
         return '<AccessGrant: role=%s, %s=%s>' % (
--- a/bigquery/tests/unit/test__helpers.py
+++ b/bigquery/tests/unit/test__helpers.py
@@ -14,6 +14,8 @@
 
 import unittest
 
+import mock
+
 
 class Test_not_null(unittest.TestCase):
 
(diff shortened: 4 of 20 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -328,10 +328,15 @@ def __init__(self, udf_type, value):
...
 
     def __eq__(self, other):
+        if not isinstance(other, UDFResource):
+            return NotImplemented
         return(
             self.udf_type == other.udf_type and
             self.value == other.value)
 
+    def __ne__(self, other):
+        return not self == other
```
dhermes: @lukesneeringer Isn't this wrong? i.e. you still want to return `NotImplemented` if `__eq__` does

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -101,16 +101,12 @@ def _key(self):
...
+        if not isinstance(other, SchemaField):
             return NotImplemented
+        return self._key() == other._key()
 
     def __ne__(self, other):
-        if isinstance(other, SchemaField):
-            return self._key() != other._key()
-        else:
-            return NotImplemented
+        return not self == other
```
dhermes: @lukesneeringer ISTM this refactor has made the impl. **less** correct

bigquery/google/cloud/bigquery/_helpers.py
lukesneeringer: It does the right thing:

```python
Python 3.6.1 (default, May  4 2017, 09:10:34) 
[GCC 4.8.4] on linux
>>> from unittest import mock
>>> 
>>> class Foo:
...     def __eq__(self, other):
...         return NotImplemented
...     def __ne__(self, other):
...         return not self == other
... 
>>> 
>>> foo = Foo()
>>> 
>>> foo == object()
False
>>> foo != object()
True
>>> foo == mock.ANY
True
>>> foo != mock.ANY
False
>>> 
```

bigquery/google/cloud/bigquery/schema.py
lukesneeringer: This spot was repeating (while inverting, of course), the `__eq__` logic.

bigquery/google/cloud/bigquery/_helpers.py
lukesneeringer: The reason why: The entire "right thing" happens with the `==` operator. So, `self.__eq__(other)` gets called, _but_ unlike if we returned `not self.__eq__(other)` directly, if it returns `NotImplemented`, then the other object is asked and it is completely resolved to the boolean, which is then inverted.

Therefore, while `__ne__` never returns `NotImplemented`, it does correctly resolve the one that `__eq__` returns.

dhermes: I am currently poring over an "examination" of how these calls works: https://gist.github.com/dhermes/3e3ad5b6aceabaacc3898fedbe040ff9

I don't fully understand it, so I'm trying.

dhermes: @lukesneeringer Just saw your 2nd comment! Makes sense

lukesneeringer: Is it LGTM then?

dhermes:
LGTM

Issue #3771, opened 2017-08-09
[Vision]  add a new function 'detect_all' in image.py


bryanyang0528:
There is a function `detect` in image.py for detecting multiple feature types. Could we add a function called `detect_all` to detect all feature types at one time? This is my proposal:

```python
    def detect_all(self, limit=10):
        """Detect multiple feature types.

        :type limit: int
        :param limit: The number of faces to try and detect.

        :rtype: list
        :returns: List of
                  :class:`~google.cloud.vision.entity.EntityAnnotation`.
        """

        features = [Feature(x, limit) for x in FeatureTypes.__dict__.keys() if x[:1] != '_']
        annotations = self.detect(features)
        return annotations[0]
```

lukesneeringer:
Hi @bryanyang0528,
Thanks for reporting.

I have good news. :-) This exists (in the newest version), and it is spelled `annotate_image`. Just send a request with `features` unspecified, and we specify the list of "all of them" for you.

Thanks for using the Vision API!

bryanyang0528:
Hi @lukesneeringer  Thank you for quick reply. Looking forward to see the new version :)

PR #3772, opened 2017-08-09
[Vision] add detect_all in image.py

bryanyang0528:
Here is the proposal for the issue #3771


lukesneeringer:
Hi @bryanyang0528,
First, thanks for being interested enough to send a PR. We appreciate it. :-)

The surface that you are playing with is deprecated, and I would prefer not modify it at this point. As I mentioned on #3771 before seeing this, the new surface has the feature you want. :-) I feel a little bad closing this PR since you obviously went to go do work for it, but hopefully the good news of "this exists" is a reasonable consolation prize. :-)

PR #3782, opened 2017-08-09
Wait for load jobs to complete in system tests.

### LINKED ISSUE #3742: test_sync_query_w_dml has num_dml_affected_rows unset
See: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2753

@tswast Any idea what would cause this "flake" to happen? (Note that the 2.7 system tests in the same build did not have this failure, as well as every other time this test case has run on CI.)

tswast:
Specify job ID in BQ DB-API DML tests to allow for further test flake
debugging.

Attempting to debug #3742

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -35,6 +35,7 @@ from test_utils.retry import RetryResult
 from test_utils.system import unique_resource_id
 
 
+JOB_TIMEOUT = 120  # 2 minutes
 WHERE = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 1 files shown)


tswast:
On closer look, the `job.result()` calls might be redundant.

lukesneeringer:
@tswast Do you want this merged?

bigquery/tests/system.py
```diff
@@ -799,7 +802,9 @@ def test_dbapi_w_dml(self):
             WHERE greeting = 'Hello World'
             """
 
-        Config.CURSOR.execute(query_template.format(dataset_name, table_name))
+        Config.CURSOR.execute(
+            query_template.format(dataset_name, table_name),
+            job_id='test_dbapi_w_dml_{}'.format(str(uuid.uuid4())))
```
dhermes: @tswast Prefer using `unique_resource_id`?

lukesneeringer:
That answers that. :-)

bigquery/tests/system.py
tswast: Done.

tswast:
I updated the `.result()` futures API calls to have a 2 minute timeout rather than potentially hang forever. I think the `.result()` was redundant with the retry until job done calls, so I removed those, too.

PR #3785, opened 2017-08-10
Block creation of transaction for session w/ existing txn.

### LINKED ISSUE #3476: Cloud Spanner Client should prevent nested transactions
Cloud Spanner does not support [nested transactions](https://en.wikipedia.org/wiki/Nested_transaction). But the client library does allow users to call `database.run_in_transaction` inside of the callback that is provided to an outer `database.run_in_transaction`. This is misleading since users might believe that this will behave like a nested transaction but in reality these will be two independent transactions. This is confusing and a source of bugs. Specifically the inner transaction might succeed while the outer might ABORT, in which case the callback will be rerun thus rerunning the inner transaction. 
We should prevent this from happening by detecting and raising an error if someone calls `database.run_in_transaction` inside the callback provided to `database.run_in_transaction`.  Note that this needs to be done before Beta since this is a known breaking change.
Also note that if we come across a use case where we do want to allow calling a nested `database.run_in_transaction`, we should be able to enable that in future without making a breaking change. 

cc @jgeewax @bjwatson

tseaver:
Closes #3476.

/cc @vkedia

### DIFF
--- a/spanner/google/cloud/spanner/session.py
+++ b/spanner/google/cloud/spanner/session.py
@@ -249,6 +249,7 @@ class Session(object):
 
         if self._transaction is not None:
             self._transaction._rolled_back = True
+            del self._transaction
 
         txn = self._transaction = Transaction(self)
         return txn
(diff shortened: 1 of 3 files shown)


vkedia:
I don't think this resolves the issue. This only helps if it is the same session. But when user calls database.run_in_transaction inside the callback supplied to database.run_in_transaction, it will check out a new session and thus still proceed.

PR #3787, opened 2017-08-10
Drop 'Database.read' and 'Database.execute_sql' convenience methods.

tseaver:
Because the context managers they use returned the session to the database's pool, application code could not safely iterate over the result sets returned by the methods.

Update docs for `Snapshot.read` and `Snapshot.execute_sql` to emphasize iteration of their results sets before the session is returned to the database pool (i.e., within the `with` block which constructs the snapshot).

Closes #3769.


tseaver:
Not waiting for appveyor.

PR #3789, opened 2017-08-10
Prevent nested transactions.

### LINKED ISSUE #3476: Cloud Spanner Client should prevent nested transactions

lukesneeringer:
Fixes #3476.

/cc @vkedia

Note that there is still some limitation on how much we can stop people from firing a blowgun into their foot. Someone could conceivably have two database objects, for example. But this is well within the limits of what we need to check for, in my opinion, especially given that you have to do pretty unusual stuff to even get _this_ far.

### DIFF
--- a/spanner/google/cloud/spanner/database.py
+++ b/spanner/google/cloud/spanner/database.py
@@ -15,6 +15,7 @@
 """User friendly container for Cloud Spanner Database."""
 
 import re
+import threading
 
 import google.auth.credentials
 from google.gax.errors import GaxError
@@ -79,7 +80,7 @@ class Database(object):
         self.database_id = database_id
         self._instance = instance
         self._ddl_statements = _check_ddl_statements(ddl_statements)
-        self._transaction_running = False
+        self._local = threading.local()
 
         if pool is None:
             pool = BurstyPool()
@@ -336,9 +337,9 @@ class Database(object):
         # Sanity check: Is there a transaction already running?
         # If there is, then raise a red flag. Otherwise, mark that this one
         # is running.
-        if self._transaction_running:
+        if getattr(self._local, 'transaction_running', False):
             raise RuntimeError('Spanner does not support nested transactions.')
-        self._transaction_running = True
+        self._local.transaction_running = True
 
         # Check out a session and run the function in a transaction; once
         # done, flip the sanity check bit back.
@@ -346,7 +347,7 @@ class Database(object):
             with SessionCheckout(self._pool) as session:
                 return session.run_in_transaction(func, *args, **kw)
         finally:
-            self._transaction_running = False
+            self._local.transaction_running = False
 
     def batch(self):
         """Return an object which wraps a batch.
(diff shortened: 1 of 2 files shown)


spanner/google/cloud/spanner/database.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, database_id, instance, ddl_statements=(), pool=None):
         self.database_id = database_id
         self._instance = instance
         self._ddl_statements = _check_ddl_statements(ddl_statements)
+        self._transaction_running = False
```
vkedia: Wouldn't this prevent two concurrent calls to run_in_transaction as well?

lukesneeringer: `Database.run_in_transaction` blocks, does it not?

lukesneeringer: Update: Yes, `run_in_transaction` blocks.

vkedia: Yes but one can use same database object across multiple threads.

lukesneeringer: @tseaver @jonparrott Are `Database` objects thread-safe? That would be surprising to me.

If they are, then Tres is probably correct that there really is not a whole lot we can do (and, considering the contortions required to get into this problem, that does not bother me much).

If they are not, then this solution will work.

lukesneeringer: @jonparrott says they are indeed thread-safe.

vkedia: Can you use `threading.local` to set this state in a thread local?

lukesneeringer: @jonparrott just gave me the exact same idea. Note that it will not work if the user uses some other concurrency strategy though.

lukesneeringer:
@vkedia If this works for you, I will merge it.

tseaver:
FWIW, this seems like a bunch of fragile mechanism to prevent (incompletely) a *very* rare kind of user error.

lukesneeringer:
Yeah, I have a hard time seeing a situation where a user does this by accident, even before this check was in place.

vkedia:
There are databases that support [nested transactions](http://docs.oracle.com/cd/E17076_02/html/gsg_txn/C/nestedtxn.html) so it is not all the inconcievable that some user might expect cloud spanner to behave in the same way. Before this check, they would start using nested transactions and nothing would fail since as far as spanner is concerned these are independent transactions. They might incorrectly assume that this is actually working as a nested transaction. That is why it is worthwhile to have this check even if it does not work in all cases.

Issue #3775, opened 2017-08-09
Cloud spanner client does not seem to be retrying broken streams


vkedia:
[Documentation](http://google-cloud-python.readthedocs.io/en/latest/spanner/snapshot-usage.html) for snapshot transactions mentions:

```
If streaming a chunk fails due to a “resumable” error, Session.read() retries the StreamingRead API reqeust, passing the resume_token from the last partial result streamed.
```

This is the correct thing to do but the client actually does not seem to be doing this. I looked around a little bit in the code and I could find anyplace where it is retrying the read with the resume token. 
This is the place where the stream would throw an error if it broke:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/9d9b6c0708ae6553458ad6107e5ad8efb23762e8/spanner/google/cloud/spanner/streamed.py#L132

 I dont see any exception handling happening here and it looks like that the error would just be propagated back to the user.

vkedia:
cc @bjwatson @tseaver @lukesneeringer

lukesneeringer:
Our retry logic is in GAX, but leaving this open for @tseaver to confirm that it is applicable here.

vkedia:
This needs to be done at a higher level than GAX, since we send resume tokens along with the responses. So if a stream breaks, instead of retrying the whole stream, the library should just pass the resume token back in the request which will cause the stream to be resumed from that point onwards. It is further complicated since cloud spanner might not send resume tokens with all responses so the client library needs to buffer responses till it sees one with resume token and only then yield them to the user. This will allow it to safely resume in case the stream breaks.
This is the Java implementation of the same logic:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L2166

What does GAX do for streaming calls? Does it retry the whole stream if it fails? What if the stream had yielded all but the last 1 response before failing? In that case retrying the whole stream is very wasteful. If GAX does do retries for streaming calls, we need to disable them.

jonparrott:
> What does GAX do for streaming calls? Does it retry the whole stream if it fails? What if the stream had yielded all but the last 1 response before failing? In that case retrying the whole stream is very wasteful. If GAX does do retries for streaming calls, we need to disable them.

gax is pretty naive in what it does for anything other than simple RPCs. I'm writing a doc internally on this now. :)

bjwatson:
[...]

PR #3790, opened 2017-08-11
Document retry of exceptions while streaming 'read'/'execute_sql' results

tseaver:
See: #3775.

This PR does *not* close that issue:  it only documents the actual, current retry semantics, instead of the intended-but-not-yet-implemented semantics in the current docs.

### DIFF
--- a/docs/spanner/snapshot-usage.rst
+++ b/docs/spanner/snapshot-usage.rst
@@ -64,9 +64,24 @@ fails if the result set is too large,
 
 .. note::
 
-   If streaming a chunk fails due to a "resumable" error,
-   :meth:`Session.read` retries the ``StreamingRead`` API reqeust,
-   passing the ``resume_token`` from the last partial result streamed.
+   If streaming a chunk raises an exception, the application can
+   retry the ``read``, passing the ``resume_token`` from ``StreamingResultSet``
+   which raised the error.  E.g.:
+
+   .. code:: python
+
+      result = snapshot.read(table, columns, keys)
+      while True:
+          try:
+              for row in result.rows:
+                  print row
+          except Exception:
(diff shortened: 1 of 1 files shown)


docs/spanner/snapshot-usage.rst
```diff
@@ -64,9 +64,24 @@ fails if the result set is too large,
...
+   .. code:: python
+
+      result = snapshot.read(table, columns, keys)
+      while True:
+          try:
+              for row in result.rows:
+                  print row
+          except Exception:
+               result = snapshot.read(
+                  table, columns, keys, resume_token=result.resume_token)
```
vkedia: This is not completely correct. As I had mentioned resume token might not be present in every single PRS. In that case this field would be empty which mean this code would end reading everything again and thus seeing duplicate rows. Or the other way round where the resume token was there but we did not yet have a complete row. For eg let us say a single row was chunked into 3 chunks PRS1, PRS2, PRS3. PRS1 was received which had a resume token. Stream broke after that. Now if the user resumes we will only fetch PRS2 and PRS3 but PRS1 has been lost since it was in a different iterator. 

 More I think about this, it seems to be really tricky to document exactly what the user needs to do in the case of a broken stream. Its better to just fix this in the code.

tseaver:
@vkedia The `result` in that example is a `StreamedResultSet`, not the raw protobuf response.  We capture the `resume_token` from the response protobuf into the Python  after each successful fetch.

vkedia:
Yes but each successful fetch might not have a resume_token. Also the StreamedResultSet returns the resume_token from the last PRS to the user even though the PRS itself might not have been yielded to the user since it does not form a complete row yet. That will lead to problem 2 I mentioned.

PR #3791, opened 2017-08-11
Make the Spanner README better.

### LINKED ISSUE #3238: Enhance Readme file for spanner
Currently Readme file for spanner does not have much in it. It should be enhanced to include a usage section like the other apis.
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/README.rst

lukesneeringer:
As @vkedia (/cc) pointed out, the existing README is pretty sparse. This improves it.

That said, I have barely _used_ this library and would like to ensure that what I have here is correct and in line with the recommended ways to use the API.

Fixes #3238.

### DIFF
--- a/spanner/README.rst
+++ b/spanner/README.rst
@@ -12,3 +12,132 @@ Quick Start
 .. code-block:: console
 
     $ pip install --upgrade google-cloud-spanner
+
+
+Authentication
+--------------
+
+With ``google-cloud-python`` we try to make authentication as painless as
+possible. Check out the `Authentication section`_ in our documentation to
+learn more. You may also find the `authentication document`_ shared by all
+the ``google-cloud-*`` libraries to be helpful.
+
+.. _Authentication section: https://google-cloud-python.readthedocs.io/en/latest/core/auth.html
+.. _authentication document: https://github.com/GoogleCloudPlatform/gcloud-common/tree/master/authentication
+
+
+Using the API
+-------------
+
+Cloud Spanner is the world’s first fully managed relational database service
+to offer both strong consistency and horizontal scalability for
+mission-critical online transaction processing (OLTP) applications. With Cloud
+Spanner you enjoy all the traditional benefits of a relational database; but
+unlike any other relational database service, Cloud Spanner scales
+horizontally to hundreds or thousands of servers to handle the biggest
+transactional workloads. (`Learn more`_)
+
+.. _Learn more: https://cloud.google.com/spanner/
+
+
+Executing Arbitrary SQL in a Transaction
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Generally, to work with Cloud Spanner, you will want a transaction. The
+preferred mechanism for this is to create a single function, which executes
+as a callback to ``database.run_in_transaction``:
+
+.. code:: python
+
+    # First, define the function that represents a single "unit of work"
+    # that should be run within the transaction.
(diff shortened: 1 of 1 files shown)


spanner/README.rst
```diff
@@ -12,3 +12,132 @@ Quick Start
...
+    # that should be run within the transaction.
+    def update_anniversary(transaction, person_id, unix_timestamp):
+        # The query itself is just a string.
+        #
+        # The use of @parameters is recommended rather than doing your
+        # own string interpolation; this provides protections against
+        # SQL injection attacks.
+        query = """UPDATE people
+            SET anniversary = @uxts
+            WHERE id = @person_id"""
```
vkedia: Cloud spanner does not support DML so this will not work. You will have to use the mutation API to update in spanner.

spanner/README.rst
```diff
@@ -12,3 +12,132 @@ Quick Start
...
+    )
+
+
+Learn More
+----------
+
+See the ``google-cloud-python`` API `Cloud Spanner documentation`_ to learn how
+to connect to Cloud Spanner using this Client Library.
+
+.. _Cloud Spanner documentation: https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/usage.html
```
vkedia: Incorrect link. It points to bigquery.

spanner/README.rst
```diff
@@ -12,3 +12,132 @@ Quick Start
...
+mission-critical online transaction processing (OLTP) applications. With Cloud
+Spanner you enjoy all the traditional benefits of a relational database; but
+unlike any other relational database service, Cloud Spanner scales
+horizontally to hundreds or thousands of servers to handle the biggest
+transactional workloads. (`About Cloud Spanner`_)
+
+.. _About Cloud Spanner: https://cloud.google.com/spanner/
+
+
+Executing Arbitrary SQL in a Transaction
```
vkedia: I am wondering if we need so many examples here. We already have pretty comprehensive set of examples in getting started guide. We can just have a simple quick start which demonstrates how to create a database object and then maybe make a simple sql query using that. We can point to getting started for more examples.

spanner/README.rst
lukesneeringer: Bah. I copied and pasted from them.

spanner/README.rst
lukesneeringer: Sorry, I am confused. I pulled this (not this exact query, but the concept) from [the Spanner docs](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest).

How _should_ this be written?

tseaver: Simplest would be to do a  `SELECT` query, followed by an update using `transaction.update`:

```python
    def update_anniversary(transaction, person_id, years=1):
        # The query itself is just a string.
        #
        # The use of @parameters is recommended rather than doing your
        # own string interpolation; this provides protections against
        # SQL injection attacks.
        query = """SELECT anniversary FROM people
            WHERE id = @person_id"""
        # When executing the SQL statement, the query and parameters are sent
        # as separate arguments. When using parameters, you must specify
        # both the parameters themselves and their types.
        row = transaction.execute_sql(
            query=query,
            params={'person_id': person_id},
            param_types={
                'person_id': types.INT64_PARAM_TYPE,
            },
        ).one()

        old_anniversary = row[0]
        new_anniversary = _compute_anniversary(old_anniversary, years)
        transaction.update('people', ['person_id', 'anniversary'], [(person_id, new_anniversary,)]
```

PR #3792, opened 2017-08-11
Drop 'Session' as a promoted, top-level entity in docs.

### LINKED ISSUE #3740: Spanner docs are outdated
Spanner docs still talk about [checking out a session](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/docs/spanner/snapshot-usage.rst) and then doing operations on that such as `session.snapshot` and `session.batch`. 
We changed the API long time back where such operations can be done directly on the database object and that is the API we want to promote. Please audit all the docs to scrub out all the calls directly made on session object and replace those with calls on Database object. This should be doe before Beta since we expect an uptick in usage once we mark the client as Beta and we want to people to use the preferred API.

tseaver:
Update related usage docs to focus on 'Database' wrappers.

Closes #3740.


docs/spanner/database-usage.rst
```diff
@@ -117,8 +117,141 @@ method:
...
+   retries the "unit of work" function if the read / query operatoins
+   or the commit are aborted due to concurrent updates
+
+See :doc:`transaction-usage` for more complete examples of transaction usage.
+
+Configuring a session pool for a database
+-----------------------------------------
+
+Under the covers, the ``snapshot``, ``batch``, and ``run_in_transaction``
+methods use a pool of :class:`~google.cloud.spanner.session.Session`` objects
```
lukesneeringer: Extra ` here.

PR #3796, opened 2017-08-11
Fix Spanner README.

lukesneeringer:
This fixes the mistakes in my previous README (#3791).

/cc @vkedia


lukesneeringer:
Sanity check: Is it `row[0]` even when `.one()` is used?

tseaver:
> Is it row[0] even when .one() is used?

Yes, a row is a tuple of column values. `row[0]` picks off the `anniversary` column from the one-and-only row.

PR #3799, opened 2017-08-12
Add 'QueryJob.query_plan' property.

tseaver:
Cherry-picked from PR #3721

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -14,7 +14,6 @@
 
 """Define API Jobs."""
 
-import copy
 import threading
 
 import six
@@ -1291,12 +1290,12 @@ class QueryJob(_AsyncJob):
         See:
         https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.query.queryPlan
 
-        :rtype: list of dict
+        :rtype: list of :class:`QueryPlanEntry`
         :returns: mappings describing the query plan, or an empty list
                   if the query has not yet completed.
         """
         plan_entries = self._job_statistics().get('queryPlan', ())
-        return [copy.deepcopy(entry) for entry in plan_entries]
+        return [QueryPlanEntry.from_api_repr(entry) for entry in plan_entries]
 
     def query_results(self):
         """Construct a QueryResults instance, bound to this job.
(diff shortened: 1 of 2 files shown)


tseaver:
@tswast I noted your [question about creating a to-be-written `QueryPlanEntry` class](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3721#discussion_r131236571):  here, because the values are "dumb data" (the class would have no methods except those to marshall instances from the JSON resource), I think it is better to just leave them as dicts.

This is different for `referenced_tables` and `unused_query_parameters`, where we already have domain objects which correspond to those resources.

bigquery/tests/unit/test_job.py
```diff
@@ -1640,6 +1640,39 @@ def test_cancelled(self):
...
+            'id': 1234,
+            'waitRatioAvg': 2.71828,
+            'waitRatioMax': 3.14159,
+            'readRatioAvg': 1.41421,
+            'readRatioMax': 1.73205,
+            'computeRatioAvg': 0.69315,
+            'computeRatioMax': 1.09861,
+            'writeRatioAvg': 3.32193,
+            'writeRatioMax': 2.30258,
+            'recordsRead': 100,
```
tswast: This actually comes back as a string. e.g. "recordsRead": "5552452",

bigquery/tests/unit/test_job.py
```diff
@@ -1640,6 +1640,39 @@ def test_cancelled(self):
...
+            'recordsWritten': 1,
```
tswast: Ditto. This is a string in the API response. e.g. "recordsWritten": "12420",

Reason: to avoid losing precision because JSON number types assume floating point.

tswast:
Note: I still think a class would be useful for string to int conversion, but it was only a P2 requirement from BQ team, so I won't block the PR.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1278,6 +1284,20 @@ def from_api_repr(cls, resource, client):
         job._set_properties(resource)
         return job
 
+    @property
+    def query_plan(self):
```
tswast: I think we should mark as beta until we do have a class to represent these. (Mostly for type conversions of strings in JSON to ints)

tswast: I think we should mark as beta until we do have a class to represent these. (Mostly for type conversions of strings in JSON to ints)

tseaver:
@tswast db24010 should address your requested changes.

Note:  that commit adds 350 lines of code for very little tangible benefit:  I can't imagine anybody writing code which depends on the fact that `records_read` / `records_written` are integers.

PR #3811, opened 2017-08-14
RPC retries (resubmission)

calpeyser:
Re-submission for PR #3324.


calpeyser:
@dhermes please take a look.  Is there anything I should do outside of regular CI to run the tests that failed the last time around?

dhermes:
@calpeyser Sending a PR from your remote will skip the system tests, which is partly why I kept the change around in a branch on the `GoogleCloudPlatform` remote.

So before we merge this, we should push this branch to the `GoogleCloudPlatform` remote so the system tests can be run. (FWIW, I have yet to review, though someone else can sign off, I'm not the only one with authority here.)

bigtable/google/cloud/bigtable/table.py
```diff
@@ -307,9 +307,6 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
         :returns: A :class:`.PartialRowsData` convenience wrapper for consuming
                   the streamed results.
         """
-        request_pb = _create_row_request(
-            self.name, start_key=start_key, end_key=end_key, filter_=filter_,
-            limit=limit, end_inclusive=end_inclusive)
```
dhermes: @calpeyser This is problematic, you've (it seems accidentally) removed the `end_inclusive` feature.

calpeyser: Done

calpeyser:
@dhermes PTAL.  Would it be easier to PR against an empty branch, so as not to deal with the backlog of commits?  I don't seem to have the privileges to rebase this branch from master.

dhermes:
@calpeyser I changed the `base` back to `master`. I'll just manually push your branch to the other remote so CI can run in parallel with this.

dhermes:
@calpeyser Here is the build and the branch:

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2901
https://github.com/GoogleCloudPlatform/google-cloud-python/tree/pr-3811

mbrukman:
@calpeyser – looks like the build/test failed on CircleCI; could you take a look, please?

calpeyser:
@dhermes can you patch the change "encode -> decode for bytestring in bigtable system test" into the branch on the remote? The system tests don't seem to run here.

The current failure is in "Update the docs", and seems unrelated to this change.

calpeyser:
@dhermes ping on pushing the lastest commit to the remote branch so we can run the system tests.

dhermes:
Sorry, doing it now!

dhermes:
(Email is a crappy task list, my bad.)

dhermes:
New build: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2951
Non-system-test red build: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2917

@calpeyser I see the docs breakage in the last build, you should probably rebase against `master` due to some moves / `nox` config changes?

dhermes:
@calpeyser The system tests failed in the build. Do they pass when you run them locally? If you haven't run them locally, is there anything I can do to help you get them running?

```
$ pip install --upgrade nox-automation
$ cd /path/to/your/git-checkout/of/google-cloud-python/
$ cd bigtable/
$ nox -s "system_tests(python_version='2.7')"
$ nox -s "system_tests(python_version='3.6')"
```

calpeyser:
@dhermes sorry for the latency! I've rebased against master and fixed the failing tests.  Looks like CI is passing now.

dhermes:
@calpeyser The green build still doesn't include system tests:

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/tree/pull%2F3811

The branch I was using for system tests didn't get re-run:

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/tree/pr-3811

Should I push this update?

(Our system tests will only run with valid credentials, so for security reasons we limit them to branches from the `GoogleCloudPlatform` remote.)

-----

Also, I'm concerned with how your merge went. 61 commits and 155 changed fails seems to be very bad news bears.

mbrukman:
@calpeyser — are you sure you rebased? It looks like there are a bunch of commits that are not your own showing up; as @dhermes says, too many commits and too many files changed for this PR.

Here's what I typically do for PRs:

```sh
$ git checkout master
$ git pull upstream --ff-only
$ git checkout fix-retry   # your branch in this case
$ git rebase master
$ git push -f
```

Is this what you did?

Sometimes I also squash my branch before pushing:

```
$ git log
# see how many of the most recent commits are mine
$ git rebase -i HEAD~N   # where N = # of most recent commits that are mine
# squash everything
$ git push -f
```

dhermes:
@calpeyser Looks like you got the commits straightened out. Did you push to the `GoogleCloudPlatform` remote as well (so that system tests can run)? If not, would you like me to do that?

mbrukman:
@dhermes — do you mean pushing to the `pr-3811` branch? Would you mind doing that, please?

calpeyser:
@dhermes yes, please do that :)

dhermes:
Done. Here is the build: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3202

bigtable/tests/system.py
```diff
@@ -272,6 +272,85 @@ def test_delete_column_family(self):
...
+        def process_scan(table, range, ids):
+            range_chunks = range.split(",")
+            range_open = range_chunks[0].lstrip("[")
+            range_close = range_chunks[1].rstrip(")")
+            rows = table.read_rows(range_open, range_close)
+            rows.consume_all()
+
+        # Download server
+        MOCK_SERVER_URLS = {
+            'Linux': 'https://storage.googleapis.com/cloud-bigtable-test/retries/retry_server_linux.tar.gz',
```
mbrukman: Where's the source for the mock server? Why is it a raw binary? This is a bit opaque. What if someone wants to update how that server works, how would they do that?

Can it be written in Python and part of this code base that's started up? Or can this be implemented as a feature in the [Bigtable emulator (written on Go)](https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/bigtable/bttest/inmem.go) which can be downloaded via `go get`?

mbrukman:
@calpeyser looks like the error is:

```
nox > Session unit_tests(python_version='3.6') successful. :)
nox > Running session system_tests(python_version='2.7')
nox > virtualenv /var/code/gcp/.nox/sys-2-7 -p python2.7
nox > chdir /var/code/gcp/bigtable
nox > pip install --upgrade mock pytest ../core/
nox > pip install --upgrade ../test_utils/
nox > pip install --upgrade .
nox > py.test --quiet tests/system.py
........F......
_________________________ TestTableAdminAPI.test_retry _________________________
  File "/var/code/gcp/bigtable/tests/system.py", line 338, in test_retry
    process_scan(table, chunks[2], chunks[3])
  File "/var/code/gcp/bigtable/tests/system.py", line 299, in process_scan
    rows.consume_all()
  File "/var/code/gcp/.nox/sys-2-7/lib/python2.7/site-packages/google/cloud/bigtable/row_data.py", line 326, in consume_all
    self.consume_next()
  File "/var/code/gcp/.nox/sys-2-7/lib/python2.7/site-packages/google/cloud/bigtable/row_data.py", line 261, in consume_next
  File "/var/code/gcp/.nox/sys-2-7/lib/python2.7/site-packages/google/cloud/bigtable/retry.py", line 101, in next
    six.reraise(type(error), error)
  File "<string>", line 2, in reraise
_Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.CANCELLED, The test has failed)>
----------------------------- Captured stdout call -----------------------------
retry_server
----------------------------- Captured stderr call -----------------------------

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
1 failed, 14 passed in 28.53 seconds
```

Seems like there's an external server binary that it's attempting to download, but the file is corrupted, or the download did not complete?

bigtable/tests/system.py
dhermes: @mbrukman We actually control the CircleCI environment, so I'd be much happier if we embedded this mock server as part of the Dockerfile. WDYT? /cc @lukesneeringer

mbrukman: I'm fine with that too, I just want to see the code behind this server in source control, so that it's available for inspection to understand what it does, or to debug future client changes, and possible for future developers to update/extend the server, if needed.

calpeyser: The errors in the system tests seem to arise from this network call - the  "tar: Unexpected EOF in archive" failure suggests that the test is getting an empty archive from the network instead of the real server binary.  Futhermore, this tests passes locally for me, with my machine's internet connection.  Perhaps CI is not permitting this network call?

If that's true, it would be easier to just check in the server.  The server is currently written in Go, meaning we'd either have to check in multiple binaries (per platform) or take on a dependency on the go compiler in the system tests.  Alternatively, I could re-implement the server in Python, but that would be some more work.

The server was put at that web endpoint by @garye - Gary WDYT?

garye: It's certainly fine with me if we check in the server. Sounds like that's the easiest way forward right now.

garye: How about we add a wget/download step to the dockerfile for now that pulls the binary down from GCS, and I'll work on open-sourcing the tool itself? When it's out there I can adjust the dockerfile to do a `go get` / `go build` instead?

mbrukman: @dhermes – would modifying [`.circleci/config.yml`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/.circleci/config.yml) to add another `run` step be the right way of doing this?

dhermes: Yes, or just making a shell script that runs for Bigtable?

garye: @dhermes @calpeyser As a concrete proposal, what about incorporating this:
https://github.com/garye/google-cloud-python/commit/5979a5c070915d23e3841938f34364f4b1fae709

And then making the necessary adjustment to the test that uses the retry server?

dhermes: Seems like a fine workaround. I made comments on the commit but it's structurally the right idea.

garye: @dhermes Thanks I made some updates.
@calpeyser Can you update with something like the above commit when you get a chance?

calpeyser:
@dhermes @mbrukman @garye PTAL at the most recent commit.  I've used an environment variable to distinguish between local runs of system_tests (where the server needs to be downloaded) and CI runs (where the server has already been downloaded in test setup).

.circleci/config.yml
```diff
@@ -46,6 +46,8 @@ jobs:
           name: Run tests - google.cloud.bigtable
           command: |
             if [[ -n $(grep bigtable ~/target_packages) ]]; then
+              test_utils/scripts/circleci/prepare_bigtable.sh
+              export DO_NOT_DOWNLOAD_TEST_SERVER=0
```
mbrukman: Rather than making variables negative (which makes DO_NOT_DO_X=0 hard to interpret, as it's a double-negation), how about `export DOWNLOAD_BIGTABLE_TEST_SERVER=0` instead?

bigtable/tests/system.py
```diff
@@ -272,6 +272,88 @@ def test_delete_column_family(self):
...
+            os.remove(SERVER_ZIP)
+
+        def process_scan(table, range, ids):
+            range_chunks = range.split(",")
+            range_open = range_chunks[0].lstrip("[")
+            range_close = range_chunks[1].rstrip(")")
+            rows = table.read_rows(range_open, range_close)
+            rows.consume_all()
+
+        if not os.environ.get("DO_NOT_DOWNLOAD_TEST_SERVER"):
```
mbrukman: How about:

```python
if os.environ.get('DOWNLOAD_BIGTABLE_TEST_SERVER') == '1':
    download_server()
```

bigtable/tests/unit/_testing.py
```diff
@@ -42,5 +51,19 @@ def __call__(self, *args, **kwargs):
         """Sync method meant to mock a gRPC stub request."""
         self._stub.method_calls.append((self._name, args, kwargs))
         curr_result, self._stub.results = (self._stub.results[0],
-                                           self._stub.results[1:])
+                                       self._stub.results[1:])
```
mbrukman: This should probably keep the same indentation as in the earlier change (i.e., the `self`s should line up).

bigtable/tests/system.py
calpeyser: Well, then developers running these tests on their dev machines would have to set an environment variable manually before they ran the system tests - do we want that?

.circleci/config.yml
calpeyser: See comment below.

bigtable/tests/system.py
```diff
@@ -272,6 +272,88 @@ def test_delete_column_family(self):
...
+            from urllib.request import urlopen
+        except ImportError:
+            # python 2
+            from urllib2 import urlopen
+
+        TEST_SCRIPT = 'tests/retry_test_script.txt'
+        SERVER_NAME = 'retry_server'
+        SERVER_ZIP = SERVER_NAME + ".tar.gz"
+
+        def download_server():
```
mbrukman: This method should verify that the downloaded and extracted file doesn't already exist, so it doesn't re-download it every time a user reruns the tests locally.

Right now, it appears that the download will happen unconditionally, every time.

bigtable/tests/system.py
mbrukman: OK, then we can make it default to download it iff:

* the env var is unspecified, OR
* it's specified to `1` (since the Circle CI explicitly defines it to `0`, it will disable it)

```python
should_download = os.environ.get('DOWNLOAD_BIGTABLE_TEST_SERVER')
if should_download is None or should_download == '1':
    download_server()
```

calpeyser: Done

bigtable/tests/system.py
calpeyser: Done

calpeyser:
@mbrukman PTAL

.circleci/config.yml
```diff
@@ -46,6 +46,8 @@ jobs:
           name: Run tests - google.cloud.bigtable
           command: |
             if [[ -n $(grep bigtable ~/target_packages) ]]; then
+              test_utils/scripts/circleci/prepare_bigtable.sh
+              export DOWNLOAD_BIGTABLE_SERVER=0
```
mbrukman: Just so that people don't panic or get too excited that we're downloading the *real* Bigtable server, this should really be `DOWNLOAD_BIGTABLE_TEST_SERVER`. :-)

bigtable/tests/system.py
```diff
@@ -272,6 +272,90 @@ def test_delete_column_family(self):
...
+            # Download server
```
mbrukman: This comment is superfluous: it's the same as the function name itself, and doesn't add any more information. Let's delete it.

bigtable/tests/system.py
```diff
@@ -272,6 +272,90 @@ def test_delete_column_family(self):
...
+            os.remove(SERVER_ZIP)
+
+        def process_scan(table, range, ids):
+            range_chunks = range.split(",")
+            range_open = range_chunks[0].lstrip("[")
+            range_close = range_chunks[1].rstrip(")")
+            rows = table.read_rows(range_open, range_close)
+            rows.consume_all()
+
+        should_download = os.environ.get("DOWNLOAD_BIGTABLE_SERVER")
```
mbrukman: Single quotes for consistency.

bigtable/tests/system.py
```diff
@@ -272,6 +272,90 @@ def test_delete_column_family(self):
...
+            mock_server_download = urlopen(MOCK_SERVER_URLS[test_platform], context=context).read()
+            mock_server_file = open(SERVER_ZIP, 'wb')
+            mock_server_file.write(mock_server_download)
+
+            # Unzip server
+            subprocess.call(['tar', 'zxvf', SERVER_ZIP, '-C', '.'])
+            os.remove(SERVER_ZIP)
+
+        def process_scan(table, range, ids):
+            range_chunks = range.split(",")
```
mbrukman: Single quotes for consistency, here and on the next 2 lines as well.

bigtable/tests/system.py
mbrukman: I'm sorry, I don't see where this function only downloads the first time, and not every time: did you upload your latest changes?

bigtable/tests/system.py
```diff
@@ -272,6 +272,90 @@ def test_delete_column_family(self):
...
+            test_platform = platform.system()
+            if test_platform not in MOCK_SERVER_URLS:
+                self.skip('Retry server not available for platform {0}.'.format(test_platform))
+
+            context = ssl._create_unverified_context()
+            mock_server_download = urlopen(MOCK_SERVER_URLS[test_platform], context=context).read()
+            mock_server_file = open(SERVER_ZIP, 'wb')
+            mock_server_file.write(mock_server_download)
+
+            # Unzip server
```
mbrukman: How about "Extract server binary from archive"?

igorbernstein2:
I think I might be missing something, does this handle mid row failures? for example: if a row is broken up into multiple chunks across 2 responses and there is an io error before receiving the second response. It seems like the `PartialRowsData` will have accumulated the first set of cells, which will re-streamed upon a retry.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,189 @@
...
+        req_pb = _create_row_request(self.name, start_key=self.start_key,
+                                     start_key_closed=self.start_key_closed,
+                                     end_key=self.end_key,
+                                     filter_=self.filter_, limit=self.limit,
+                                     end_inclusive=self.end_inclusive)
+        self.stream = self.client._data_stub.ReadRows(req_pb)
+
+    def next(self, *args, **kwargs):
+        """
+        Read and return the next row from the stream.
```
garye: This is actually the next chunk I think

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -274,6 +274,9 @@ def consume_next(self):
 
             self._validate_chunk(chunk)
 
+            if hasattr(self._response_iterator, 'set_start_key'):
+                self._response_iterator.set_start_key(chunk.row_key)
```
garye: I think this has to happen after `save_current_row()`, otherwise won't this skip a row if the it fails in the middle?

garye:
@igorbernstein2 I think you're right. A test would prove it but there needs to be some way for the retrying mechanism to signal that the cached row data should be cleared.

igorbernstein2:
What is retry_server_linux? is the source available anywhere?

garye:
@igorbernstein2 It's a fake server to help with integration tests. Source is internal but I'll put it in google-cloud-go when I get a chance.

bigtable/google/cloud/bigtable/row_data.py
calpeyser: Can you clarify what you mean here? Don't we need to set the start key on any chunk, even if it is not a "commit_row" chunk?

garye: This happens in `consume_next()`. I'm worried about a case where a partial row gets retrieved and then an error gets thrown. We set the `start_key` to the row key of the partial chunk that we are working on, but make it open/exclusive. So I think the next read would pick up after the row that we never finished (and would never go back to)? 

I believe we need to wipe out any partial row data on error and make sure we set the start key to the last *completed* row key.

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -0,0 +1,205 @@
...
+                exc = errors.RetryError(
+                    'Retry total timeout exceeded with exception', error)
+
+                # Sleep a random number which will, on average, equal the
+                # expected delay.
+                to_sleep = random.uniform(0, delay * 2)
+                time.sleep(to_sleep / _MILLIS_PER_SECOND)
+                delay = min(delay * self.delay_mult, self.max_delay_millis)
+                now = time.time()
+                timeout = min(
```
justinseinlin: Is this 'timeout' used anywhere?

bigtable/google/cloud/bigtable/retry.py
```diff
@@ -113,6 +128,7 @@ def next(self, *args, **kwargs):
                 timeout = min(
                     timeout * self.timeout_mult, self.max_timeout,
                     deadline - now)
+                self._responses_for_row = 0
```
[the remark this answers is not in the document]
calpeyser: Dosen't technically have to happen, since the field is only used to check for recent retries.  Updated anyway for correctness.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -267,6 +267,10 @@ def consume_next(self):
 
         self._last_scanned_row_key = response.last_scanned_row_key
 
+        if hasattr(self._response_iterator, 'responses_for_row'):
+            if (self._response_iterator.responses_for_row == 1):
```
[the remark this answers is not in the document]
calpeyser: No, "1" means that the last read was a retry.

bigtable/google/cloud/bigtable/retry.py
calpeyser: No, removed.

bigtable/google/cloud/bigtable/row_data.py
calpeyser: Does this work? I moved the check to after rows are commited.

calpeyser:
@garye what do you think?

garye:
LGTM! Just need @dhermes, @lukesneeringer or someone from the team to pull the trigger I think.

calpeyser:
@tseaver - I've fixed the issues with the merge above - can this PR be merged?

dhermes:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4249

I think this may need to be reverted

dhermes:
@calpeyser @garye Here is another (albeit less severe) failure:

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4256

There have been no other [builds][1] on `master` since the previous failure.

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/tree/master

----

It's worth noting that we have **other** packages with tests that are flaky (which we tolerate mostly due to lack of developer resources), but this one seems to be a preventable flakiness.

dhermes:
I'm just going to collect a list of broken builds here (for debugging purposes):

- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4249
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4256
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4262

I may edit it as more occur / if more occur.

dhermes:
@calpeyser @garye Please weigh in here. I am leaning towards reverting this PR (we need our builds to not fail).

igorbernstein2:
I can try to fix this tomorrow. After a quick look, I think that the issue is that the test_retry test never cleans up the `BIGTABLE_EMULATOR_HOST` env var, causing the rest of the tests to connect to a dead port

dhermes:
Thanks @igorbernstein2. I'm happy to help (e.g. if you're unfamiliar with `unittest.mock.patch`).

calpeyser:
This breakage certainly looks related to this change.  @igorbernstein2, thanks a ton - I think you're right about the env var.

Issue #3736, opened 2017-08-07
blob.download_to_filename fails with google-cloud-storage==1.3.0


sonlac:
Due to the notable implementation changes in **google-cloud-storage 1.3.0**, the **_blob.download_to_filename_** fails whatever **google-cloud** from **0.24.0** to **0.27.0** (latest)

Our prod server which using **google-cloud==0.24.0** has been broken since last Saturday. I analyzed a bit the error. The problem is come from the upgraded package **google-cloud-storage==1.3.0**. We are using the **google-cloud-storage==1.2.0**.

I did several tests which figured out the following problems:

**Problem 1:** _blob.download_to_filename failed when using google-cloud==0.27.0_ (included **google-cloud-storage==1.3.0** as default, defined in **setup.py**). Here is a part of stacktrace:
```python
blob.download_to_filename(blob_local_path)
  File "/root/.local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 482, in download_to_filename
    self.download_to_file(file_obj, client=client)
  File "/root/.local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 464, in download_to_file
    self._do_download(transport, file_obj, download_url, headers)
  File "/root/.local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 418, in _do_download
    download.consume(transport)
  File "/root/.local/lib/python2.7/site-packages/google/resumable_media/requests/download.py", line 101, in consume
    self._write_to_stream(result)
  File "/root/.local/lib/python2.7/site-packages/google/resumable_media/requests/download.py", line 62, in _write_to_stream
    with response:
AttributeError: __exit__
```

**Problem 2:** _blob.download_to_filename failed when using google-cloud==0.24.0_ (included **google-cloud-storage==1.3.0** as default, defined in **setup.py**). Here is a part of stacktrace:
```python
File "dev_env/python_venv/local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 482, in download_to_filename
    self.download_to_file(file_obj, client=client)
  File "dev_env/python_venv/local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 464, in download_to_file
    self._do_download(transport, file_obj, download_url, headers)
  File "dev_env/python_venv/local/lib/python2.7/site-packages/google/cloud/storage/blob.py", line 418, in _do_download
    download.consume(transport)
[...]

PR #3770, opened 2017-08-09
Move google.cloud.iterator to google.api.core.page_iterator

jonparrott:
* Re-write tests to pytest style.
* Make GAXIterator private- it will soon be removed.
* Pass api_request into HTTPIterator to avoid accessing private members
* Update references across libraries.

PR #3812, opened 2017-08-14
Update storage to 1.3.2.

dhermes:
Did this to update the lower bound on the `requests` dependency (see #3736).

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-storage',
-    version='1.3.1',
+    version='1.3.2',
     description='Python Client for Google Cloud Storage',
     long_description=README,
     namespace_packages=[


dhermes:
Blerg! #3770 means a `core` release would be needed for `storage` and #3737 means this should probably be a `1.4.x` release.

:cry:

jonparrott:
Ugh, I was worried about this.

We have two options:

1. Keep revving core (& all dependent packages) every time we need a release until I'm done. (My ETA is ~1 month)
2. Cherry-pick into a release branch and release from there.

@lukesneeringer WDYT?

jonparrott:
Alternatively I can hold on off making any more breaking changes in core (basically don't remove any google.cloud packages and cut over depend usage until the very end).

dhermes:
> Cherry-pick into a release branch and release from there.

My preference

lukesneeringer:
I have no preference. Whichever you two prefer since you are the ones doing the work.

PR #3814, opened 2017-08-15
Update storage to 1.3.2.

dhermes:
Did this to update the lower bound on the `requests` dependency (see #3736).

This PR is an alternative to #3812. Note the diffbase is `GoogleCloudPlatform:storage-1.3.2`, which is at the same point as the `storage-1.3.1` branch:

https://github.com/GoogleCloudPlatform/google-cloud-python/commits/storage-1.3.1
https://github.com/GoogleCloudPlatform/google-cloud-python/commits/storage-1.3.2

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-storage',
-    version='1.3.1',
+    version='1.3.2',
     description='Python Client for Google Cloud Storage',
     long_description=README,
     namespace_packages=[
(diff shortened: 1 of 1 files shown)


jonparrott:
I see the version bump but not the minimum dep version bump (or am I just missing something)?

dhermes:
MERP! Good call @jonparrott

dhermes:
@jonparrott PTAL

dhermes:
@jonparrott @lukesneeringer The version bump should also be merged into master, so maybe I should re-open #3812 (though I already deleted my branch)

PR #3819, opened 2017-08-15
Add google.api.core.retry with base retry functionality

jonparrott:
Additionally:
* Add google.api.core.exceptions.RetryError
* Add google.api.core.helpers package
* Add google.api.core.helpers.datetime_helpers module

### DIFF
--- a/core/google/api/core/helpers/datetime_helpers.py
+++ b/core/google/api/core/helpers/datetime_helpers.py
@@ -1,0 +1,22 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for :mod:`datetime`."""
+
+import datetime
+
+
+def utcnow():
+    """A :meth:`datetime.datetime.utcnow()` alias to allow mocking in tests."""
+    return datetime.datetime.utcnow()
--- a/core/google/api/core/retry.py
+++ b/core/google/api/core/retry.py
@@ -1,0 +1,139 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for retrying functions with exponential back-off."""
+
+import datetime
+import logging
+import random
+import time
+
+import six
+
+from google.api.core import exceptions
+from google.api.core.helpers import datetime_helpers
+
+_LOGGER = logging.getLogger(__name__)
+_DEFAULT_JITTER_AMOUNT = 0.2
+
+
+def if_exception_type(*exception_types):
+    """Creates a predicate to check if the exception is of a given type.
+
+    Args:
+        exception_types (Sequence[type]): The exception types to check for.
+
+    Returns:
+        Callable[Exception]: A predicate that returns True if the provided
+            exception is of the given type(s).
+    """
+    def inner(exception):
+        """Bound predicate for checking an exception type."""
+        return isinstance(exception, exception_types)
+    return inner
+
+
+# pylint: disable=invalid-name
+# Pylint sees this as a constant, but it is also an alias that should be
+# considered a function.
+if_transient_error = if_exception_type((
+    exceptions.InternalServerError,
+    exceptions.TooManyRequests))
+"""A predicate that checks if an exception is a transient API error."""
+# pylint: enable=invalid-name
+
+
+def exponential_sleep_generator(
+        initial, maximum, multiplier=2, jitter=_DEFAULT_JITTER_AMOUNT):
+    """Generates sleep intervals based on the exponential back-off algorithm.
+
+    This implements the `Truncated Exponential Back-off`_ algorithm.
+
+    .. _Truncated Exponential Back-off:
+        https://cloud.google.com/storage/docs/exponential-backoff
+
+    Args:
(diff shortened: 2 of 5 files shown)


jonparrott:
@dhermes @lukesneeringer please note:

1. This just adds the core retry functionality. I will be sending other PRs to add a general-purpose `Retry` class/decorator and possibly a `RetryIterator` class.
2. I would like feedback on the name `retry_target`. It's a low-level helper, but it will still be public.
3. I would like feedback on the module name `datetime_helper` - it repeats a word in its import path (`google.api.core.helpers.datetime_helper`) but it does so to avoid conflict with the module it helper (e.g., you don't have to `from google.api.core.helpers import datetime as datetime_helper`).

core/google/api/core/retry.py
```diff
@@ -0,0 +1,139 @@
...
+import random
+import time
+
+import six
+
+from google.api.core import exceptions
+from google.api.core.helpers import datetime_helpers
+
+_LOGGER = logging.getLogger(__name__)
+_DEFAULT_JITTER_AMOUNT = 0.2
```
dhermes: It's not jitter if it's deterministic. `_DEFAULT_MAX_JITTER`?

core/google/api/core/retry.py
```diff
@@ -0,0 +1,139 @@
...
+    return inner
+
+
+# pylint: disable=invalid-name
+# Pylint sees this as a constant, but it is also an alias that should be
+# considered a function.
+if_transient_error = if_exception_type((
+    exceptions.InternalServerError,
+    exceptions.TooManyRequests))
+"""A predicate that checks if an exception is a transient API error."""
```
dhermes: Can you add some more content:

```
"""A predicate that checks if an exception is a transient API error.

The following server errors are considered transient:

- ``InternalServerError`` (HTTP 500, gRPC ``INTERNAL(13)``)
- ``TooManyRequests`` (HTTP 429)
- ``ResourceExhausted`` (gRPC ``RESOURCE_EXHAUSTED(8)``)
"""
```

Not sure if you want to leave these out, but they are also ["equivalent"][1] to 500:

```
- ``Unknown`` (gRPC ``UNKNOWN(2)``)
- ``DataLoss`` (gRPC ``DATA_LOSS(15)``)
```

[1]: https://github.com/googleapis/googleapis/blob/660dd12452d36df708172ea01540a83eb2a4362d/google/rpc/code.proto

core/google/api/core/retry.py
```diff
@@ -0,0 +1,139 @@
...
+        jitter (float): The maximum about of randomness to apply to the delay.
+
+    Yields:
+        float: successive sleep intervals.
+    """
+    delay = initial
+    while True:
+        yield delay
+        delay = min(
+            delay * multiplier + random.uniform(0, jitter), maximum)
```
dhermes: Reference: https://github.com/GoogleCloudPlatform/google-resumable-media-python/blob/4eb54dd8652ede0f4e3a9177f2747401c269118c/google/resumable_media/_helpers.py#L97-L121

core/google/api/core/retry.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Returns:
+        Any: the return value of the target function.
+
+    Raises:
+        google.api.core.RetryError: If the deadline is exceeded while retrying.
+        ValueError: If the sleep generator stops yielding values.
+        Exception: If the target raises a method that isn't retryable.
+    """
+    if deadline is not None:
```
dhermes: the `not` is unnecessary since you populate both branches. Drop `not` and switch them?

core/google/api/core/retry.py
```diff
@@ -0,0 +1,139 @@
...
+        except Exception as exc:
+            if not predicate(exc):
+                raise
+            last_exc = exc
+
+        now = datetime_helpers.utcnow()
+        if deadline_datetime is not None and deadline_datetime < now:
+            six.raise_from(
+                exceptions.RetryError(
+                    'Deadline of {:.1f}s exceeded while calling {}'.format(
```
dhermes: Do you really want `.1f`?

core/google/api/core/retry.py
jonparrott: Done.

core/google/api/core/retry.py
jonparrott: Done. Added a comment that all `InternalServerError` subclasses are included.

core/google/api/core/retry.py
jonparrott: Eh.

core/google/api/core/retry.py
jonparrott: yes- gapic can pass silly values into this.

core/google/api/core/helpers/datetime_helpers.py
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for :mod:`datetime`."""
+
+import datetime
+
+
+def utcnow():
```
lukesneeringer: Why is this a thing? `mock.patch.object(datetime.datetime, 'utcnow')` is not difficult.

jonparrott: It *is* difficult:

```python
>>> import datetime
>>> mock.patch.object(datetime.datetime, 'utcnow')
<mock.mock._patch object at 0x1013ca090>
>>> with _:
...     datetime.datetime.utcnow()
...
  File "/Users/jonwayne/.pyenv/versions/2.7.13/lib/python2.7/site-packages/mock/mock.py", line 1460, in __enter__
    setattr(self.target, self.attribute, new_attr)
TypeError: can't set attributes of built-in/extension type 'datetime.datetime'
```

core/google/api/core/retry.py
lukesneeringer: The way it is written now tracks how I would logically think about the branches.

core/google/api/core/retry.py
```diff
@@ -0,0 +1,148 @@
...
+                        deadline, target),
+                    last_exc),
+                last_exc)
+
+        _LOGGER.debug('Retrying due to {}, sleeping {:.1f}s ...'.format(
+            last_exc, sleep))
+        time.sleep(sleep)
+
+    raise ValueError('Sleep generator stopped yielding sleep values.')
```
lukesneeringer: Under what situation do you get here? And should this be `Stopiteration`?

jonparrott: You wouldn't get here under normal circumstances - the sleep generator should be infinite. However, code coverage did catch this case. :)

I went back and forth on `StopIteration`. It's a weird error to raise from a non-iterator and the actual reason for this error is that the *value* of the sleep generator was incorrect.

core/google/api/core/helpers/datetime_helpers.py
lukesneeringer: Nevermind! :-)

core/google/api/core/retry.py
lukesneeringer: Okay, fair enough.

PR #3730, opened 2017-08-04
Updating all affected packages after google-cloud-core update.

dhermes:

PR #3828, opened 2017-08-16
Fix doc build by swapping install order

jonparrott:


dhermes:
This is a lose-lose, sadly. @lukesneeringer fixed this previously, but I undid it in #3730.

The reason: if our source tree depends on packages that are not yet on PyPI, the build won't work.

jonparrott:
I see - ideally we should just stop installing the umbrella package for docs.

dhermes:
Yeah great point. It's not necessary? Or maybe `docs/conf.py` checks the version of `google-cloud` via `pkg_resources`?

jonparrott:
`docs/conf.py` checks the version via pkg_resources. We could use the core version instead.

dhermes:
That seems like the "least worst" strategy. The version used in the user agent comes from `google-cloud-core`.

jonparrott:
I have a day of meetings - if you're willing (and able) to do that today I'll review.

dhermes:
On it

PR #3832, opened 2017-08-17
adding the possibility to get the id from a datastore query 2

alexmnt:
This is very useful when you need to use the entity id, recovered from a query, as a filter in another request.

### DIFF
--- a/datastore/google/cloud/datastore/entity.py
+++ b/datastore/google/cloud/datastore/entity.py
@@ -181,6 +181,19 @@ class Entity(dict):
         if self.key:
             return self.key.kind
 
+    @property
+    def id(self):
+        """Get the id of the current entity.
+
+        .. note::
+          This relies entirely on the :class:`gcloud.datastore.key.Key`
+          set on the entity.  That means that we're not storing the id
+          of the entity at all, just the properties and a pointer to a
+          Key which knows its Id.
+        """
+        if self.key:
+            return self.key.id
+
     def __repr__(self):
         if self.key:
             return '<Entity%s %s>' % (self.key._flat_path,


alexmnt:
I signed it!

datastore/google/cloud/datastore/entity.py
```diff
@@ -181,6 +181,19 @@ def kind(self):
         if self.key:
             return self.key.kind
 
+    @property
+    def id(self):
+        """Get the id of the current entity.
+
+        .. note::
+          This relies entirely on the :class:`gcloud.datastore.key.Key`
```
dhermes: Note that this package has been `google.cloud.datastore` (not `gcloud.datastore`) for over a year

datastore/google/cloud/datastore/entity.py
```diff
@@ -181,6 +181,19 @@ def kind(self):
...
+          set on the entity.  That means that we're not storing the id
+          of the entity at all, just the properties and a pointer to a
+          Key which knows its Id.
```
dhermes: I don't think this last part of the note (about storage) is necessary. It's an implementation detail.

datastore/google/cloud/datastore/entity.py
```diff
@@ -181,6 +181,19 @@ def kind(self):
...
+        """
+        if self.key:
```
dhermes: Please use an explicit `None`-check: `if self.key is not None`

nunchucks-nuns-chuck-norris-nun-12593488599

alexmnt:
@dhermes ok I will write them, maybe tonight 
Thank you for the feedback 👍

tseaver:
@alexmnt I don't see those additional tests.

chemelnucfin:
See #4640

PR #3835, opened 2017-08-17
Add google.api.core.retry.Retry decorator

jonparrott:


core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+    to an RPC call.
+
+    Although the default behavior is to retry transient API errors, a
+    different predicate can be provided to retry other exceptions.
+
+    Args:
+        predicate (Callable[Exception]): A callable that should return ``True``
+            if the given exception is retryable.
+        initial (float): The minimum about of time to delay. This must
+            be greater than 0.
```
dhermes: Add "Defaults to 1.0." (and similar for other default options? or nah?)

core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+        multiplier (float): The multiplier applied to the delay.
+        jitter (float): The maximum about of randomness to apply to the delay.
+        deadline (float): How long to keep retrying.
+    """
+    def __init__(
+            self,
+            predicate=if_transient_error,
+            initial=1,
+            maximum=60,
+            multiplier=2,
```
dhermes: Shouldn't you follow your own documented types here? (i.e. use `1.0`, `60.0` and `2.0`)

core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+
+    Args:
+        predicate (Callable[Exception]): A callable that should return ``True``
+            if the given exception is retryable.
+        initial (float): The minimum about of time to delay. This must
+            be greater than 0.
+        maximum (float): The maximum about of time to delay.
+        multiplier (float): The multiplier applied to the delay.
+        jitter (float): The maximum about of randomness to apply to the delay.
+        deadline (float): How long to keep retrying.
```
dhermes: This is similar to cumulative sleep in `google-resumable-media`?

core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+    different predicate can be provided to retry other exceptions.
+
+    Args:
+        predicate (Callable[Exception]): A callable that should return ``True``
+            if the given exception is retryable.
+        initial (float): The minimum about of time to delay. This must
+            be greater than 0.
+        maximum (float): The maximum about of time to delay.
+        multiplier (float): The multiplier applied to the delay.
+        jitter (float): The maximum about of randomness to apply to the delay.
```
dhermes: Please name this `max_jitter`

core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+            multiplier (float): The multiplier applied to the delay.
+            jitter (float): The maximum about of randomness to apply to the
+                delay.
+
+        Returns:
+            Retry: A new retry instance with the given predicate.
+        """
+        return Retry(
+            predicate=self._predicate,
+            initial=initial if initial is not None else self._initial,
```
dhermes: Prefer: class or module-level defaults instead of ternarys here

Alternatively, use `**kwargs`

core/google/api/core/retry.py
```diff
@@ -146,3 +189,129 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+            multiplier=multiplier if maximum is not None else self._multiplier,
+            jitter=jitter if jitter is not None else self._jitter,
+            deadline=self._deadline)
+
+    def __str__(self):
+        return (
+            '<Retry predicate={}, initial={:.1f}, maximum={:.1f}, '
+            'multiplier={:.1f}, jitter={:.1f}, deadline={:.1f}>'.format(
+                self._predicate, self._initial, self._maximum,
+                self._multiplier, self._jitter, self._deadline))
```
dhermes: So many indented lines! Can you make the template a module-level non-public constant?

core/google/api/core/retry.py
```diff
@@ -12,9 +12,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Helpers for retrying functions with exponential back-off."""
+"""Helpers for retrying functions with exponential back-off.
+
+The :cls`Retry` decorator can be used to retry function that raise exceptions
```
jonparrott: Done.

core/google/api/core/retry.py
jonparrott: Nah. Sphinx will document the defaults in the function signature.

Although I do realize now that all of these defaults should be constants. Done.

core/google/api/core/retry.py
jonparrott: Done.

core/google/api/core/retry.py
jonparrott: I *think* so, but I haven't look very closely at that.

core/google/api/core/retry.py
jonparrott: Done.

core/google/api/core/retry.py
jonparrott: Based on chat, leaving as-is unless @lukesneeringer has feels.

core/google/api/core/retry.py
jonparrott: I'd really rather not introduce unnecessary indirection - leaving as is unless @lukesneeringer has feels.

dhermes:
LGTM

core/google/api/core/retry.py
```diff
@@ -12,9 +12,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Helpers for retrying functions with exponential back-off."""
+"""Helpers for retrying functions with exponential back-off.
+
+The :cls`Retry` decorator can be used to retry functions that raise exceptions
```
lukesneeringer: Typo here, missing `:`.

core/google/api/core/retry.py
```diff
@@ -77,7 +125,8 @@ def exponential_sleep_generator(
             be greater than 0.
         maximum (float): The maximum about of time to delay.
         multiplier (float): The multiplier applied to the delay.
-        jitter (float): The maximum about of randomness to apply to the delay.
+        max_jitter (float): The maximum about of randomness to apply to the
```
lukesneeringer: I have in the past relied on [this article](https://www.awsarchitectureblog.com/2015/03/backoff.html) which, after empirical testing, suggests that the correct way to do jitter is, counter-intuitively, `random.uniform(0, delay)`.

Stopping short of claiming that you must do it this way, but it is what we have done elsewhere.

jonparrott: Ah, I wish you'd brought this up in the last PR. But I'm happy to make this change here.

core/google/api/core/retry.py
```diff
@@ -146,3 +195,132 @@ def retry_target(target, predicate, sleep_generator, deadline):
...
+            return retry_target(
+                target,
+                self._predicate,
+                sleep_generator,
+                self._deadline)
+
+        return retry_wrapped_func
+
+    def with_deadline(self, deadline):
+        """Returns a copy of this retry with the given deadline.
```
lukesneeringer: Epic nitpick: `s/Returns/Return` consistently, because [docstrings should be imperative](https://www.python.org/dev/peps/pep-0257/):

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

Note that I am not asking to block on this item, just mentioning it as an aside. It is worth changing if and only if convenient (e.g. you need to open the file for something else anyway).

core/google/api/core/retry.py
lukesneeringer: I would also prefer class-level defaults, but not enough to ask you to change it.

core/google/api/core/retry.py
lukesneeringer: I am sorry. I even thought about it at the time, but decided then that "he has a good reason". I should have asked then.

jonparrott: Never assume I know what I'm doing. :)

core/google/api/core/retry.py
jonparrott: Done.

Issue #3506, opened 2017-06-15
BigQuery: Add QueryResults.reload() function which calls getQueryResults


tswast:
https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults

One of the things I learned when talking to the BigQuery API folks is that the recommended way to run a query is to create a query job and then call `getQueryResults()` until the `jobComplete` property is `true`.

The reason for this is that a call to get the job resource will return immediately, whereas `getQueryResults` will return after a timeout or the query is complete, whichever comes first. With `getQueryResults` users get their data faster.

This means the code for waiting for a query job to complete would change from

```python
while True:
   query_job.reload()
    if query_job.state == 'DONE':
        return
    time.sleep(1)
```

to something like

```python
query_results = query_job.results()
while not query_results.complete:
    query_results.reload()
    # No need for sleep, since reload should wait for timeout
```

tswast:
I figured out that I can force a reload with

        query_iterator = query_results.fetch_data()
        try:
            return six.next(iter(query_iterator))
        except StopIteration:
            pass

but obviously, that's not ideal for the purpose of waiting for a job to finish.

tseaver:
If you are going to iterate over the rows anyway, then that second example becomes:

```python
for row in query_results.fetch_data():
    do_something_with(row)
```

That certainly seems like "idiomatic" usage.  Does that accomplish what you need?

tswast:
No. The results might not all be there. That's why I need a reload function to check if the query is complete or not. (A bit hard to test, since it requires a query that takes longer than the timeout value, which defaults to 10 seconds).

tswast:
A reload function probably isn't the best solution. After thinking about this, I think `QueryResults` should also implement the futures interface.

jonparrott:
I don't think `QueryResults` should implement the future interface, rather, I think that `QueryJob` (which already has the futures interface) should use the `getQueryResults` API method instead of `getJob`.

tswast:
After I've played around with some proposed designs for GA, I agree with you, Jon. QueryJob's implementation of futures interface will be sufficient.

I've changed this request back to asking for `.reload()` method since there does need to be a more straightforward way to actually call `getQueryResults`.

Closing this issue since I'm tracking internally in GA redesign doc.

PR #3838, opened 2017-08-17
BigQuery - add get_query_results method.

tswast:
This method calls the getQueryResults API directly and returns a
QueryResults object. Note: the response from this API does not include
the query, so I modified the constructor to make query optional in this
case.

Addresses #3506

### DIFF
--- a/bigquery/tests/unit/test_client.py
+++ b/bigquery/tests/unit/test_client.py
@@ -45,6 +45,64 @@ class TestClient(unittest.TestCase):
         self.assertIs(client._connection.credentials, creds)
         self.assertIs(client._connection.http, http)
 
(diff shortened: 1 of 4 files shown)


bigquery/tests/unit/test_client.py
```diff
@@ -607,6 +665,11 @@ def __init__(self, *responses):
         self._requested = []
 
     def api_request(self, **kw):
+        from google.cloud.exceptions import NotFound
         self._requested.append(kw)
+
+        if len(self._responses) == 0:
```
tswast: I grabbed this from https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3804

PR #3843, opened 2017-08-18
Add google.api.page_iterator.GRPCIterator

jonparrott:


core/google/api/core/page_iterator.py
```diff
@@ -423,3 +423,90 @@ def _next_page(self):
...
+        Returns:
+            Page: The next page in the iterator or :data:`None` if there are no
+                pages left.
+        """
+        if not self._has_next_page():
+            return None
+
+        if self.next_page_token is not None:
+            setattr(
+                self._request, self._request_token_field, self.next_page_token)
```
dhermes: Ahhh I am sad we have to use `setattr`

core/google/api/core/page_iterator.py
```diff
@@ -423,3 +423,90 @@ def _next_page(self):
...
+
+        response = self._method(self._request)
+
+        self.next_page_token = getattr(response, self._response_token_field)
+        items = getattr(response, self._items_field)
+        page = Page(self, items, self._item_to_value)
+
+        return page
```
dhermes: This `getattr` / `setattr` business is so icky. How much of it is YAGNI?

core/google/api/core/page_iterator.py
jonparrott: I am, too, but it's entirely possible for a service to use non-standard fields (it's part of the gapic config).

core/google/api/core/page_iterator.py
jonparrott: basically none. It's more or less the same about of configurability as the HTTPIterator.

jonparrott:
> How dare you keep sending such short and well-crafted PRs!

Let's see how long I can keep this streak up. :)

PR #3845, opened 2017-08-19
Allow fetching more than the first page when max_results is set.

### LINKED ISSUE #3840: Fetching next batch of results using BigQuery's dbapi
Currently, the maximum number of rows that you can get when iterating using `fetchmany`, `fetchall` or `fetchone` is limited by `arraysize`.

I see that those methods are using iterator returned by `_query_data.fetch_data()` with `max_results` set to `arraysize`
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/36304f7448677cff2f89558ef87c374e904a0492/bigquery/google/cloud/bigquery/dbapi/cursor.py#L171-L172

However, shouldn't it automatically fetch the next page of results when we are finished iterating over the current batch of results? Otherwise, it makes it impossible to retrieve full results from the query without setting`arraysize` to an arbitrary large number.

Here is what I've tried:

```
cursor.execute('SELECT * FROM dataset.table')

while True:
    rows = cursor.fetchmany(5000)
    if not rows:
        break

    for row in rows:
        print(row)
```

```
cursor.execute('SELECT * FROM dataset.table')

row = cursor.fetchone()
while row:
   print(row)
   row = cursor.fetchone()
```

I am using google-cloud-bigquery 0.26.0

tswast:
Closes #3840

### DIFF
--- a/bigquery/google/cloud/bigquery/dbapi/cursor.py
+++ b/bigquery/google/cloud/bigquery/dbapi/cursor.py
@@ -191,7 +191,6 @@ class Cursor(object):
             self._query_data = iter(
                 self._query_results.fetch_data(max_results=size))
 
-
     def fetchone(self):
         """Fetch a single row from the results of the last ``execute*()`` call.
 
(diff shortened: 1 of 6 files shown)


tswast:
Oh, also fixes a bug where `arraysize` was ignored if set after `execute()` in BigQuery DB-API.

PR #3848, opened 2017-08-21
PubSub Documentation

lukesneeringer:
Uses #3637 as a base.

That means the line is now:

  * master
  * pubsub
  * pubsub-publisher
  * pubsub-subscriber
  * pubsub-docs


lukesneeringer:
I goofed.

Issue #3842, opened 2017-08-18
Move gax functionality into google.api.core


jonparrott:
- [x] [retry](https://github.com/googleapis/gax-python/blob/baa1c13545c642d461c44e6d5f71d6ec446be5f5/google/gax/retry.py)
- [x] [timeout](https://github.com/googleapis/gax-python/blob/baa1c13545c642d461c44e6d5f71d6ec446be5f5/google/gax/retry.py)
- [x] [path template](https://github.com/googleapis/gax-python/blob/baa1c13545c642d461c44e6d5f71d6ec446be5f5/google/gax/path_template.py) [(context)](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L212)
- [x] [api callable](https://github.com/googleapis/gax-python/blob/baa1c13545c642d461c44e6d5f71d6ec446be5f5/google/gax/api_callable.py)
  - [x] loading gapic config
  - [x] grpc error wrapping
  - [x] plain methods
  - [x] paged methods
- [x] LRO client & methods
- [x] [utils](https://github.com/googleapis/gax-python/tree/baa1c13545c642d461c44e6d5f71d6ec446be5f5/google/gax/utils)

PR #3851, opened 2017-08-22
Add google.api.core.path_template

jonparrott:
Towards #3842

### DIFF
--- a/core/google/api/core/path_template.py
+++ b/core/google/api/core/path_template.py
@@ -47,7 +47,7 @@ _VARIABLE_RE = re.compile(r"""
         |
         # Match & capture named variables {name}
         {
-            (?P<name>.+?)
+            (?P<name>[^/]+?)
             # Optionally match and capture the named variable's template.
             (?:=(?P<template>.+?))?
         }
--- a/core/google/api/core/path_template.py
+++ b/core/google/api/core/path_template.py
@@ -69,7 +69,7 @@ def _expand_variable_match(positional_vars, named_vars, match):
         match (re.Match): A regular expression match.
 
     Returns:
-        str: The expaned variable to replace the match.
+        str: The expanded variable to replace the match.
 
     Raises:
         ValueError: If a positional or named variable is required by the
--- a/core/google/api/core/path_template.py
+++ b/core/google/api/core/path_template.py
@@ -55,8 +55,8 @@ _VARIABLE_RE = re.compile(r"""
     """, re.VERBOSE)
 
 # Segment expressions used for validating paths against a template.
-_SINGLE_SEGEMENT_PATTERN = r'([^/]+)'
-_MULTI_SEGEMENT_PATTERN = r'(.+)'
+_SINGLE_SEGMENT_PATTERN = r'([^/]+)'
+_MULTI_SEGMENT_PATTERN = r'(.+)'
 
 
 def _expand_variable_match(positional_vars, named_vars, match):
@@ -143,15 +143,15 @@ def _replace_variable_with_pattern(match):
     template = match.group('template')
     if name is not None:
         if not template:
-            return _SINGLE_SEGEMENT_PATTERN.format(name)
+            return _SINGLE_SEGMENT_PATTERN.format(name)
         elif template == '**':
-            return _MULTI_SEGEMENT_PATTERN.format(name)
+            return _MULTI_SEGMENT_PATTERN.format(name)
         else:
             return _generate_pattern_for_template(template)
     elif positional == '*':
-        return _SINGLE_SEGEMENT_PATTERN
+        return _SINGLE_SEGMENT_PATTERN
     elif positional == '**':
-        return _MULTI_SEGEMENT_PATTERN
+        return _MULTI_SEGMENT_PATTERN
     else:
         raise ValueError(
             'Unknown template expression {}'.format(
@@ -187,7 +187,7 @@ def validate(tmpl, path):
 
     Args:
         tmpl (str): The path template.
-        path (str): The expand path.
+        path (str): The expanded path.
 
     Returns:
         bool: True if the path matches.
--- a/core/google/api/core/path_template.py
+++ b/core/google/api/core/path_template.py
@@ -1,0 +1,200 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Expand and validate URL path templates.
+
+This module provides the :func:`expand` and :func:`validate` functions for
+interacting with Google-style URL `path templates`_ which are commonly used
+in Google APIs for `resource names`_.
+
+.. _path templates: https://github.com/googleapis/googleapis/blob
+    /57e2d376ac7ef48681554204a3ba78a414f2c533/google/api/http.proto#L212
+.. _resource names: https://cloud.google.com/apis/design/resource_names
+"""
+
+from __future__ import unicode_literals
+
+import functools
+import re
+
+import six
+
+# Regular expression for extracting variable parts from a path template.
+# The variables can be expressed as:
+#
+# - "*": a single-segment positional variable, for example: "books/*"
+# - "**": a multi-segment positional variable, for example: "shelf/**/book/*"
+# - "{name}": a single-segment wildcard named variable, for example
+#   "books/{name}"
+# - "{name=*}: same as above.
+# - "{name=**}": a multi-segment wildcard named variable, for example
+#   "shelf/{name=**}"
+# - "{name=/path/*/**}": a multi-segment named variable with a pattern.
+_VARIABLE_RE = re.compile(r"""
+    (  # Capture the entire variable expression
+        (?P<positional>\*\*?)  # Match & capture * and ** positional variables.
+        |
+        # Match & capture named variables {name}
+        {
+            (?P<name>.+?)
+            # Optionally match and capture the named variable's pattern.
+            (?:=(?P<template>.+?))?
+        }
+    )
+    """, re.VERBOSE)
+
+# Segment expressions used for validating paths against a template.
+_SINGLE_SEGEMENT_PATTERN = r'([^/]+)'
+_MULTI_SEGEMENT_PATTERN = r'(.+)'
+
+
+def _expand_variable_match(positional_vars, named_vars, match):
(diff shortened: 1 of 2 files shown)


jonparrott:
@lukesneeringer @dhermes this gave me more trouble than I anticipated, but the tests give me confidence that this works as expected.

lukesneeringer:
Still simpler than a lex-yacc implementation. ;-)

Sanity check: Correct?

jonparrott:
Much simpler

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,196 @@
...
+    positional = match.group('positional')
+    name = match.group('name')
+    if name is not None:
+        try:
+            return six.text_type(named_vars[name])
+        except KeyError:
+            raise ValueError(
+                'Named variable \'{}\' not specified and needed by template '
+                '`{}` at position {}'.format(
+                    name, match.string, match.start()))
```
dhermes: Is it necessary to check the state of `positional` somewhere in this branch?

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,196 @@
...
+            be modified.
+        named_vars (dict): A dictionary of named variables.
+        match (re.Match): A regular expression match.
+
+    Returns:
+        str: The expaned variable to replace the match.
+
+    Raises:
+        ValueError: If a positional or named variable is required by the
+            template but not specified.
```
dhermes: This doesn't seem to accurately describe the error states?

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,196 @@
...
+        ValueError: If an unexpected template expression is encountered.
+    """
+    positional = match.group('positional')
+    name = match.group('name')
+    template = match.group('template')
+    if name is not None:
+        if not template:
+            return _SINGLE_SEGEMENT_PATTERN.format(name)
+        elif template == '**':
+            return _MULTI_SEGEMENT_PATTERN.format(name)
```
dhermes: s/`SEGEMENT`/`SEGMENT` here and everywhere

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,196 @@
...
+        False
+
+    Args:
+        tmpl (str): The path template.
+        path (str): The expand path.
+
+    Returns:
+        bool: True if the path matches.
+    """
+    pattern = _generate_pattern_for_template(tmpl) + '$'
```
dhermes: Why not `'^' + _gen() + '$'`?

core/tests/unit/api_core/test_path_template.py
```diff
@@ -0,0 +1,90 @@
...
+])
+def test_expand_success(tmpl, args, kwargs, expected_result):
+    result = path_template.expand(tmpl, *args, **kwargs)
+    assert result == expected_result
+    assert path_template.validate(tmpl, result)
+
+
+@pytest.mark.parametrize('tmpl, args, kwargs, exc_match', [
+    # Missing positional arg.
+    ['v1/*', [], {}, 'Positional'],
```
dhermes: Would these fail?

```
['/v1/*', ['a/b'], {}, '??'],
['/v1/*', ['a', 'b'], {}, '??'],
['/v1/{name}', [], {'name': 'parent/child'}, '??'],
['/v1/{name=parent/*}', [], {'name': 'doggy/cat'}, '??'],
```

core/tests/unit/api_core/test_path_template.py
```diff
@@ -0,0 +1,90 @@
...
+    ['/v1/{name=**}', [], {'name': 'parent/child'}, '/v1/parent/child'],
+    # Named params with a sub-template
+    ['/v1/{name=parent/*}', [], {'name': 'parent/child'}, '/v1/parent/child'],
+    ['/v1/{name=parent/**}', [], {'name': 'parent/child/object'},
+     '/v1/parent/child/object'],
+    # Combining positional and named params
+    ['/v1/*/{name}', ['a'], {'name': 'parent'}, '/v1/a/parent'],
+    ['/v1/{name}/*', ['a'], {'name': 'parent'}, '/v1/parent/a'],
+    ['/v1/{parent}/*/{child}/*', ['a', 'b'],
+     {'parent': 'parent', 'child': 'child'}, '/v1/parent/a/child/b'],
```
dhermes: Seems that having the template name and value being the same could be a source of bugs. Rename the values? E.g. `{'parent': 'zorp', 'child': 'zap'}`

core/google/api/core/path_template.py
jonparrott: Updated to include this can also happen if an invalid expression is encountered.

core/google/api/core/path_template.py
jonparrott: No, positional is unused in this branch (the group wouldn't match anything)

core/google/api/core/path_template.py
jonparrott: lol done.

core/google/api/core/path_template.py
jonparrott: `^` is implicit when using `match`. 

> If zero or more characters at the beginning of string match the regular expression pattern

`search` is used to match anywhere in the string.

core/tests/unit/api_core/test_path_template.py
jonparrott: Good call, gave them different values.

core/tests/unit/api_core/test_path_template.py
jonparrott: > ['/v1/*', ['a/b'], {}, '??']

This will not fail to expand, but will fail to validate.

I could make this fail during expansion by checking that the value for a `*` positional arg is a single segment.

> ['/v1/*', ['a', 'b'], {}, '??']

This will not fail to expand, the additional positional arg will just be ignored.

I could make this fail during expansion by checking that there are no leftover positional or named args.

> ['/v1/{name}', [], {'name': 'parent/child'}, '??'],

This will expand, but not validate.

I could make this fail during expansion using the same method as the first case.

> ['/v1/{name=parent/*}', [], {'name': 'doggy/cat'}, '??'],

This will expand, but not validate.

I could make this fail during expansion by doing a validation step for the sub template.

All but the second case could also be made to fail by making expand validate before returning the string. I am not sure if it's worth the effort.

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,198 @@
...
+
+from __future__ import unicode_literals
+
+import functools
+import re
+
+import six
+
+# Regular expression for extracting variable parts from a path template.
+# The variables can be expressed as:
```
duggelz: Great Googly-Moogly, now the lexer makes more sense.  Is this Python-specific?  Google-specific?  Documented somewhere?

jonparrott: Google-specific, it's linked above in the docstring: https://github.com/googleapis/googleapis/blob/57e2d376ac7ef48681554204a3ba78a414f2c533/google/api/http.proto#L212

core/tests/unit/api_core/test_path_template.py
```diff
@@ -0,0 +1,90 @@
...
+    ['/v1/*/{name}', ['a'], {'name': 'parent'}, '/v1/a/parent'],
+    ['/v1/{name}/*', ['a'], {'name': 'parent'}, '/v1/parent/a'],
+    ['/v1/{parent}/*/{child}/*', ['a', 'b'],
+     {'parent': 'thor', 'child': 'thorson'}, '/v1/thor/a/thorson/b'],
+    ['/v1/{name}/**', ['a/b'], {'name': 'parent'}, '/v1/parent/a/b'],
+    # Combining positional and named params with sub-templates.
+    ['/v1/{name=parent/*}/*', ['a'], {'name': 'parent/child'},
+     '/v1/parent/child/a'],
+    ['/v1/*/{name=parent/**}', ['a'], {'name': 'parent/child/object'},
+     '/v1/a/parent/child/object'],
```
duggelz: Your testcases need more cruelty.

```
/v1/***
/v1/{{foo
/v1/{}
/v1/}
/v1/{fo/o}
```

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,198 @@
...
+    """Expand a matched variable with its value.
+
+    Args:
+        positional_vars (list): A list of positonal variables. This list will
+            be modified.
+        named_vars (dict): A dictionary of named variables.
+        match (re.Match): A regular expression match.
+
+    Returns:
+        str: The expaned variable to replace the match.
```
duggelz: `expaned` -> `expanded`

core/tests/unit/api_core/test_path_template.py
jonparrott: Possibly, but this module is more intended to detect issues with the arguments rather than the template itself. These templates are more or less coming from proto files which are separately validated. Asserting that a given path template itself is valid would require a lexer/parser which is really what we wanted to avoid.

In a lot of these cases this module won't balk at these strings:

> /v1/***

is recognized as having two positional args, the first one is multi-segment and the second is single-segment.

> /v1/{{foo
> /v1/{}
> /v1/}
> /v1/{fo/o}
> /v1/{name={

Are all considered to not have any variables.

> /v1/{**}

is a fun case because it'll expect a named variable called `**` (which is possible to specify with `expand(tmpl, **{'**': 'value'})`. (I love this one)

core/google/api/core/path_template.py
jonparrott: fixed.

core/google/api/core/path_template.py
duggelz: Do we care about the `Verb = ":" LITERAL` syntax?  Also, I note that `IDENT` is never defined anywhere, but presumably can't contain `}` or `=`.  Too bad it isn't restricted further to alphanumeric or something.

core/google/api/core/path_template.py
```diff
@@ -0,0 +1,198 @@
...
+    Returns:
+        str: The expanded path
+
+    Raises:
+        ValueError: If a positional or named variable is required by the
+            template but not specified or if an unexpected template expression
+            is encountered.
+    """
+    replacer = functools.partial(_expand_variable_match, list(args), kwargs)
+    return _VARIABLE_RE.sub(replacer, tmpl)
```
duggelz: This will allow `/v1/{foo}{bar}` but shouldn't, if I'm reading the grammar right.  Maybe add non-capturing `/` to the start and end of _VARIABLE_RE? (Kind of complicated because we could also be at the start/end of the string).

core/google/api/core/path_template.py
jonparrott: Nah, if it's present it's just kept as-is by this. The purpose of this is to just expand any variables in the string.

core/google/api/core/path_template.py
jonparrott: Yeah, as far as I can tell there isn't a way to do this without full-on parsing, as `{foo}` is technically valid.

core/google/api/core/path_template.py
lukesneeringer: I even made a big callout note about that in my book because it is such a common mistake. :-)

Issue #3622, opened 2017-07-18
Logging in GKE is not using the ContainerEngineHandler


arthurdarcet:
The `KUBERNETES_SERVICE` environment variable is not set in a pod running on GKE, so `ContainerEngineHandler` never actually gets selected as the default handler [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/client.py#L307)

This means all the logs go directly to the "Global" section of the Logging tab in the GCloud console, instead of being parsed by the fluentd pod running on the host, and being routed to the right "GKE > *cluster* > *container name*" section. And of course no pod/host/… metadata is added to the log.

I think the checked variable should read `KUBERNETES_SERVICE_HOST` <del>but i did not open a pull request because `ContainerEngineHandler` does not work either: When I manually add a `ContainerEngineHandler` handler to the logging configuration, the logging interface at console.cloud.google.com does not actually parse the JSON that is written to stderr, so the severity for instance is not set correctly (the JSON produced by `ContainerEngineHandler` is directly displayed as the `textPayload`).</del>

And finally (not really related to this but I'm not sure where this should go): the `fluentd-gcp` pod that is running by default on GKE clusters is configured with this fluentd configuration:
```
<filter …>
  …
  format /^(?<severity>\w)(?<time>\d{4} [^\s]*)\s+(?<pid>\d+)\s+(?<source>[^ \]]+)\] (?<log>.*)/
</filter>
```
which is a very weird regex:
 - it parses only one char for the `severity`
 - the `time` is parsed as `YYYY` followed by a space, followed by anything _but_ a space: this does not correspond to any datetime format i can think of
 - the `source` attribute is followed by a `]`, but no matching `[` is in the regex

The `severity` issue makes it impossible to actually use this filter since stackdriver does not accept logs with a severity outside the usual `INFO|WARN|ERROR|…`

--

All-in-all, this makes it really hard to log data from a python pod: logging with the `CloudLogging` handler goes to "Global" by default (and is missing many kubernetes-related meta), <del>logging to stderr a JSON-formatted object is not parsed by the console.cloud UI,</del> and the regex in the fluentd setup cannot pick up any valid input

```
OS: Kubernetes 1.6.6, running on GKE with cos
Python version: 3.6 or pypy 5.8
google-cloud-python version: 1.1.0
```

duggelz:
Thanks for the bug report!

[...]

PR #3856, opened 2017-08-23
Use metadata server to detect GKE environment

liyanhui1228:
For #3622. This fixes the `get_default_handler()` method to let it return the `ContainerEngineHandler` when in GKE environment using the `cluster_name` attribute from the metadata server. If this attribute exists, it should be in GKE environment.

### DIFF
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -15,6 +15,7 @@
 """Common logging helpers."""
 
 import requests
+from requests import exceptions
 
 from google.cloud.logging.entries import ProtobufEntry
 from google.cloud.logging.entries import StructEntry
@@ -70,11 +71,12 @@ def retrieve_metadata_server(metadata_key):
     try:
         response = requests.get(url, headers=METADATA_HEADERS)
 
-        if response.status_code == requests.codes.ok:
-            return response.text
-        else:
-            return None
-    except requests.exceptions.RequestException:
+        # Exception will raise if the HTTP request returned an unsuccessful
+        # status code.
+        response.raise_for_status()
+
+        return response.text
+    except (exceptions.RequestException, exceptions.HTTPError):
         # Ignore the exception, connection failed means the attribute does not
         # exist in the metadata server.
         pass
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -14,10 +14,16 @@
 
 """Common logging helpers."""
 
+import requests
 
 from google.cloud.logging.entries import ProtobufEntry
 from google.cloud.logging.entries import StructEntry
 from google.cloud.logging.entries import TextEntry
+
+METADATA_URL = 'http://metadata/computeMetadata/v1/'
+METADATA_HEADERS = {
+    'Metadata-Flavor': 'Google'
+}
 
 
 def entry_from_resource(resource, client, loggers):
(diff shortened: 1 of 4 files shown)


logging/google/cloud/logging/_helpers.py
```diff
@@ -46,3 +52,31 @@ def entry_from_resource(resource, client, loggers):
...
+
+    :rtype: str
+    :returns: The value of the metadata key returned by the metadata server.
+    """
+    url = METADATA_URL + metadata_key
+
+    try:
+        response = requests.get(url, headers=METADATA_HEADERS)
+
+        if response.status_code == requests.codes.ok:
```
jonparrott: use `raise_for_status()`?

liyanhui1228: Thanks for the suggestion! Done.

liyanhui1228: I looked at the [raise_for_status](https://github.com/requests/requests/blob/205755834d34a8a6ecf2b0b5b2e9c3e6a7f4e4b6/requests/models.py#L912) method, it only throws exception when status_code is greater than 400. But I think it's safer to only return response when status_code is 200. So I changed it back.

logging/google/cloud/logging/_helpers.py
```diff
@@ -46,3 +52,30 @@ def entry_from_resource(resource, client, loggers):
...
     raise ValueError('Cannot parse log entry resource.')
+
+
+def retrieve_metadata_server(metadata_key):
+    """Retrieve the metadata key in the metadata server.
+
+    See: https://cloud.google.com/compute/docs/storing-retrieving-metadata
+
+    :type metadata_key: str
+    :param metadata_key: Key of the metadata which will form the url.
```
duggelz: Documentation note: This is actually more than the key, at least as implemented, since you can also supply query parameters (E.g. "tags?alt=json" to change the output format).

This seems reasonable, I can't think of any security implications, so we should documented it.

logging/tests/unit/test__helpers.py
```diff
@@ -53,6 +55,69 @@ def test_proto_payload(self):
...
+            requests_mock)
+
+        with patch:
+            metadata = self._call_fut(metadata_key)
+
+        self.assertIsNone(metadata)
+
+    def test_request_exception(self):
+        metadata_key = 'test_url_cannot_connect'
+        metadata_url = 'invalid_url'
```
duggelz: Can you make this explicitly `metadata_url = 'http://metadata.invalid/'`?  (.invalid is the official invalid TLD)  I worry about someone cleverly naming their workstation `invalid_url.corp.google.com` and breaking this.  Actually, that would be kind of awesome.

Also, I'm not sure what the requests library does when you give it URL that doesn't have "http:" at the front.

logging/google/cloud/logging/_helpers.py
liyanhui1228: Yes the metadata key can include the query parameters, I just can't think of a name to represent the whole thing. And in most circumstances the metadata key will just be the metadata key. I added documentation about this parameter.

logging/tests/unit/test__helpers.py
liyanhui1228: Done. The requests library will throw `requests.exceptions.MissingSchema` exception if there is no "http:". And all exceptions that requests explicitly raises inherit from `requests.exceptions.RequestException`.

liyanhui1228:
@dhermes Danny could you please do a final pass for this and merge if looks good?

jonparrott:
LGTM, will merge once CI completes.

dhermes:
Thanks @jonparrott!

PR #3858, opened 2017-08-23
Add google.api.core.timeout

jonparrott:


jonparrott:
Towards #3842

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+automatically determined, for example:
+
+.. code-block:: python
+
+    timeout_ = timeout.ExponentialTimeout()
+    is_thing_ready_with_timeout = timeout_(is_thing_ready)
+
+    for n in range(10):
+        try:
+            is_thing_ready({'example': 'data'})
```
dhermes: `is_thing_ready_with_timeout`?

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+import datetime
+
+import six
+
+from google.api.core.helpers import datetime_helpers
+
+_DEFAULT_INITIAL_TIMEOUT = 5.0  # seconds
+_DEFAULT_MAXIMUM_TIMEOUT = 30.0  # seconds
+_DEFAULT_TIMEOUT_MULTIPLIER = 2.0
+_DEFAULT_DEADLINE = None  # seconds
```
dhermes: `None` seconds is how many?

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+    Args:
+        initial (float): The initial timeout.
+        maximum (float): The maximum timeout.
+        multiplier (float): The multiplier applied to the timeout.
+        deadline (float): The overall deadline across all invocations.
+
+    Yields:
+        float: A timeout value.
+    """
+    if deadline is not None:
```
dhermes: As always, I suggest dropping `not` and swapping the branches

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+        deadline = (
+            datetime_helpers.utcnow() +
+            datetime.timedelta(seconds=deadline))
+    else:
+        deadline = datetime.datetime.max
```
dhermes: You've changed the type of `deadline` from `Optional[int]` to `datetime.datetime`. Even worse, you've changed the **meaning** from a "delta" to an "absolute". Suggest using a two variable names?

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+            # The set maximum timeout.
+            maximum,
+            # The remaining time before the deadline is reached.
+            float((deadline - now).seconds))
+        timeout = timeout * multiplier
+
+
+@six.python_2_unicode_compatible
+class ExponentialTimeout(object):
+    """A decorator that adds a exponentially increasing timeout argument.
```
dhermes: "adds an"

core/google/api/core/timeout.py
```diff
@@ -0,0 +1,210 @@
...
+        Returns:
+            Callable: The wrapped function.
+        """
+        timeouts = _exponential_timeout_generator(
+            self._initial, self._maximum, self._multiplier, self._deadline)
+
+        @six.wraps(func)
+        def func_with_timeout(*args, **kwargs):
+            """Wrapped function that adds timeout."""
+            kwargs['timeout'] = next(timeouts)
```
dhermes: Meaning this signature will know how to handle a generator / iterator?

core/google/api/core/timeout.py
jonparrott: Good catch.

core/google/api/core/timeout.py
jonparrott: Added a better comment.

core/google/api/core/timeout.py
jonparrott: Fair enough, done.

core/google/api/core/timeout.py
jonparrott: Done.

core/google/api/core/timeout.py
jonparrott: I don't know what you mean here.

dhermes: I expect the wrapped function to want `timeout` to be a `float` or coercible to a float. This must assume that the wrapped function will be OK receiving an iterator.

jonparrott: `next(timeouts)` yields a float. The wrapped function get a single float.

dhermes: DERP! I can't read?

jonparrott: I can't spell, so we're even. :)

core/google/api/core/timeout.py
lukesneeringer: As always, I prefer the existing form. :stuck_out_tongue_winking_eye:

PR #3860, opened 2017-08-24
Make system tests not care about case.

lukesneeringer:
I really really want a green build before I merge #3859, and it [failed](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3047?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) on this unrelated system test for Translate.


dhermes:
@lukesneeringer Spanner seems to be the real issue for system tests

PR #3879, opened 2017-08-24
Update README.rst sample.

lukesneeringer:
Ugh. Stupid me missed one other documentation location...the base `README.rst` file on PyPI has an out of date sample. This updates it.

I intend to release an 0.28.2 once this is approved just so PyPI does not show a glaringly wrong sample as the first thing a user sees.


lukesneeringer:
That would be preferable to an 0.28.2 release, but I too do not think it can be done anymore.

PR #3883, opened 2017-08-25
BigQuery: make get_query_results private. Return rows for QueryJob.result()

tswast:
Based on feedback from others in the redesign plan.

The `get_query_results()` function is too low level to expose directly to users.

The result() object should just return the rows as an iterator. This simplifies the proposed query interface to:

    job = client.query('MY QUERY')
    rows = job.result()

No need for separate fetch_data() call after calling result().

### DIFF
--- a/bigquery/google/cloud/bigquery/dbapi/cursor.py
+++ b/bigquery/google/cloud/bigquery/dbapi/cursor.py
@@ -154,7 +154,7 @@ class Cursor(object):
 
         # Wait for the query to finish.
         try:
-            query_job = query_job.result()
+            query_job.result()
         except google.cloud.exceptions.GoogleCloudError:
             raise exceptions.DatabaseError(query_job.errors)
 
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -154,7 +154,7 @@ def execute(self, operation, parameters=None, job_id=None):
 
         # Wait for the query to finish.
         try:
-            query_job = query_job.result()
+            query_job.result()
         except google.cloud.exceptions.GoogleCloudError:
             raise exceptions.DatabaseError(query_job.errors)
```
dhermes: ISTM this depend on `query_job.result()` rather than `query_job` (since in the above the variable was over-written).

tswast: Result was returning self, so nothing has changed, really. The errors property should still get populated okay.

PR #3889, opened 2017-08-28
Retry 'UNAVAILABLE' exceptions during streaming

### LINKED ISSUE #3775: Cloud spanner client does not seem to be retrying broken streams

tseaver:
Closes #3775.

/cc @vkedia

### DIFF
--- a/spanner/google/cloud/spanner/streamed.py
+++ b/spanner/google/cloud/spanner/streamed.py
@@ -158,6 +158,9 @@ class StreamedResultSet(object):
         """Consume the next partial result set from the stream.
 
         Parse the result set into new/existing rows in :attr:`_rows`
+
+        :raises ValueError:
+            if the sleep generator somehow does not yield values.
         """
         sleep_generator = retry.exponential_sleep_generator(
             initial=retry._DEFAULT_INITIAL_DELAY,
--- a/spanner/google/cloud/spanner/streamed.py
+++ b/spanner/google/cloud/spanner/streamed.py
@@ -14,8 +14,6 @@
 
 """Wrapper for streaming results."""
 
-import time
-
 from google.api.core import exceptions
 from google.api.core import retry
 from google.protobuf.struct_pb2 import ListValue
@@ -44,6 +42,9 @@ the following server errors are considered transient:
 - :class:`google.api.core.exceptions.ServiceUnavailable` - HTTP 503, gRPC
     ``UNAVAILABLE``.
 """
+
+retry_unavailable = retry.Retry(predicate=if_unavailable_error)
+"""Used by `StreamedResultSet.consume_next`."""
 # pylint: enable=invalid-name
 
 
@@ -154,6 +155,18 @@ class StreamedResultSet(object):
                 self._rows.append(self._current_row)
                 self._current_row = []
 
+    def _restart_iterator(self, _exc_ignored):
+        """Helper for :meth:`consume_next`."""
+        if self._resume_token in (None, b''):
+            raise
+
+        self._response_iterator = self._restart(
+            resume_token=self._resume_token)
+
+    def _bump_iterator(self):
+        """Helper for :meth:`consume_next`."""
+        return six.next(self._response_iterator)
+
     def consume_next(self):
         """Consume the next partial result set from the stream.
 
@@ -162,34 +175,8 @@ class StreamedResultSet(object):
         :raises ValueError:
             if the sleep generator somehow does not yield values.
         """
-        sleep_generator = retry.exponential_sleep_generator(
-            initial=retry._DEFAULT_INITIAL_DELAY,
-            maximum=retry._DEFAULT_MAXIMUM_DELAY)
-
-        deadline = time.time() + _RESTART_DEADLINE
-
-        response = None
-
-        for sleep in sleep_generator:
-            try:
-                response = six.next(self._response_iterator)
-            except exceptions.ServiceUnavailable:
-
-                if self._resume_token in (None, b''):
-                    raise
-
-                if time.time() > deadline:
-                    raise
-
-                self._response_iterator = self._restart(
-                    resume_token=self._resume_token)
-
-                time.sleep(sleep)
-            else:
-                break
-
-        if response is None:
-            raise ValueError('Sleep generator stopped yielding sleep values.')
+        response = retry_unavailable(
+            self._bump_iterator, on_error=self._restart_iterator)()
 
         self._counter += 1
         self._resume_token = response.resume_token
--- a/spanner/google/cloud/spanner/streamed.py
+++ b/spanner/google/cloud/spanner/streamed.py
@@ -14,15 +14,37 @@
 
 """Wrapper for streaming results."""
 
+import time
+
+from google.api.core import exceptions
+from google.api.core import retry
 from google.protobuf.struct_pb2 import ListValue
 from google.protobuf.struct_pb2 import Value
-from google.cloud import exceptions
 from google.cloud.proto.spanner.v1 import type_pb2
 import six
 
 # pylint: disable=ungrouped-imports
 from google.cloud.spanner._helpers import _parse_value_pb
 # pylint: enable=ungrouped-imports
+
+_RESTART_DEADLINE = 30.0  # seconds
+
+
+# pylint: disable=invalid-name
+# Pylint sees this as a constant, but it is also an alias that should be
+# considered a function.
+if_unavailable_error = retry.if_exception_type((
+    exceptions.ServiceUnavailable,
+))
+"""A predicate that checks if an exception is a transient API error.
+
+For streaming the result of ``read`` / ``execute_sql`` requests, only
+the following server errors are considered transient:
+
+- :class:`google.api.core.exceptions.ServiceUnavailable` - HTTP 503, gRPC
+    ``UNAVAILABLE``.
+"""
+# pylint: enable=invalid-name
 
 
 class StreamedResultSet(object):
@@ -137,7 +159,35 @@ class StreamedResultSet(object):
 
         Parse the result set into new/existing rows in :attr:`_rows`
         """
-        response = six.next(self._response_iterator)
+        sleep_generator = retry.exponential_sleep_generator(
+            initial=retry._DEFAULT_INITIAL_DELAY,
+            maximum=retry._DEFAULT_MAXIMUM_DELAY)
+
+        deadline = time.time() + _RESTART_DEADLINE
+
+        response = None
+
+        for sleep in sleep_generator:
+            try:
+                response = six.next(self._response_iterator)
+            except exceptions.ServiceUnavailable:
+
+                if self._resume_token in (None, b''):
+                    raise
+
+                if time.time() > deadline:
+                    raise
+
+                self._response_iterator = self._restart(
+                    resume_token=self._resume_token)
+
+                time.sleep(sleep)
+            else:
+                break
+
+        if response is None:
+            raise ValueError('Sleep generator stopped yielding sleep values.')
+
         self._counter += 1
         self._resume_token = response.resume_token
 
--- a/spanner/google/cloud/spanner/streamed.py
+++ b/spanner/google/cloud/spanner/streamed.py
@@ -34,18 +34,18 @@ class StreamedResultSet(object):
         :class:`google.cloud.proto.spanner.v1.result_set_pb2.PartialResultSet`
         instances.
 
-    :type retry: callable
-    :param retry:
+    :type restart: callable
+    :param restart:
         Function (typically curried via :func:`functools.partial`) used to
-        retry the initial request if a retriable error is raised during
+        restart the initial request if a retriable error is raised during
         streaming.
 
     :type source: :class:`~google.cloud.spanner.snapshot.Snapshot`
     :param source: Snapshot from which the result set was fetched.
     """
-    def __init__(self, response_iterator, retry, source=None):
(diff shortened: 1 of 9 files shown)


spanner/google/cloud/spanner/streamed.py
```diff
@@ -34,11 +56,18 @@ class StreamedResultSet(object):
...
+    :param restart:
+        Function (typically curried via :func:`functools.partial`) used to
+        restart the initial request if a retriable error is raised during
+        streaming.
+
     :type source: :class:`~google.cloud.spanner.snapshot.Snapshot`
     :param source: Snapshot from which the result set was fetched.
     """
-    def __init__(self, response_iterator, source=None):
+    def __init__(self, response_iterator, restart, source=None):
```
dhermes: You don't want a nice default here? Or you don't expect users to interact with this constructor?

spanner/google/cloud/spanner/streamed.py
```diff
@@ -130,7 +159,35 @@ def consume_next(self):
 
         Parse the result set into new/existing rows in :attr:`_rows`
         """
-        response = six.next(self._response_iterator)
+        sleep_generator = retry.exponential_sleep_generator(
+            initial=retry._DEFAULT_INITIAL_DELAY,
+            maximum=retry._DEFAULT_MAXIMUM_DELAY)
```
dhermes: @jonparrott Should these names be public? Can we somehow avoid accessing the non-public surface in some other way?

spanner/google/cloud/spanner/streamed.py
```diff
@@ -130,7 +159,35 @@ def consume_next(self):
...
+
+        deadline = time.time() + _RESTART_DEADLINE
+
+        response = None
+
+        for sleep in sleep_generator:
+            try:
+                response = six.next(self._response_iterator)
+            except exceptions.ServiceUnavailable:
+
```
dhermes: Empty line just for fun?

spanner/google/cloud/spanner/streamed.py
```diff
@@ -130,7 +159,35 @@ def consume_next(self):
...
+                if self._resume_token in (None, b''):
```
dhermes: Seems strange to have to check both of these, why wouldn't `b''` just be the default?

spanner/google/cloud/spanner/streamed.py
```diff
@@ -130,7 +159,35 @@ def consume_next(self):
...
+
+                self._response_iterator = self._restart(
+                    resume_token=self._resume_token)
+
+                time.sleep(sleep)
+            else:
+                break
+
+        if response is None:
+            raise ValueError('Sleep generator stopped yielding sleep values.')
```
dhermes: 1. You should add a `raises` section?
1. It seems this error message is not comprehensive enough? I.e. `response` could be `None` because `self._response_iterator` broke in some unexpected way

spanner/tests/unit/test_streamed.py
```diff
@@ -25,13 +25,16 @@ def _getTargetClass(self):
...
-    def _make_one(self, *args, **kwargs):
-        return self._getTargetClass()(*args, **kwargs)
+    def _make_one(self, response_iterator, restart=object(), source=None):
+        return self._getTargetClass()(
+            response_iterator, restart, source=source)
 
     def test_ctor_defaults(self):
         iterator = _MockCancellableIterator()
-        streamed = self._make_one(iterator)
+        restart = object()
```
dhermes: `mock.sentinel.restart` may convey slightly more meaning here (not sure how familiar you are with the features of `unittest.mock`)

spanner/google/cloud/spanner/streamed.py
tseaver: The latter, really:  instances of this class are returned from `Snapshot.read` and `Snapshot.execute_sql`.

spanner/google/cloud/spanner/streamed.py
tseaver: Not without another release-the-world danceathon, AFAIK.

spanner/google/cloud/spanner/streamed.py
tseaver: `None` is "never set";  `b''` is "the back-end send an empty token."

dhermes: Is there a place where the distinction matters? (It doesn't matter here.)

spanner/tests/unit/test_streamed.py
tseaver: I'll pass here:  identity is the only point.

spanner/google/cloud/spanner/streamed.py
tseaver: > You should add a raises section?

9ae4121

> I.e. response could be None because self._response_iterator broke in some unexpected way.

`self._response_iterator` is our code (not user-supplied), and is mandated to return a protobuf.

tseaver:
@dhermes 

> The "drop resume_token from interfaces" change seems pretty large (in terms of LOC) and only partially related to this change?

The point of this change is to have the library do all the `resume_token` handling, rather than making / allowing user code to mess with it (it is a foot gun).

spanner/google/cloud/spanner/streamed.py
dhermes: > `self._response_iterator` is our code (not user-supplied)

I realize. It comes from gRPC or hand-rolled here?

spanner/google/cloud/spanner/streamed.py
jonparrott: With #3891 this won't be necessary as we can just use a `Retry` object.

tseaver: @jonparrott I don't think we are going to want to do another release-the-world to get this spanner fix pushed out.

spanner/google/cloud/spanner/streamed.py
tseaver: it is the GAPIC streaming iterator returned from the call to `SpannerClient.{read_streaming,execute_streaming_sql}`.

spanner/google/cloud/spanner/streamed.py
tseaver: @lukesneeringer Can you comment on whether we have to release the world to get a new version of `google-cloud-core` out (one which adds a forward-compatible feature to `retry.Retry`?)

lukesneeringer: What do you mean "forward-compatible" in this case? If it is a non-breaking change to core, then we could just release 0.27.1 and no release-the-world would be necessary. If it is a breaking change, then yeah, we need to release the world.

lukesneeringer: Dumb question, why not just send _nothing_? `initial` and `maximum` both have sensible defaults in `google.api.core` (those values).

tseaver: #3891 adds an optional `on_error` argument to `google.api.core.retry.retry_target` (and the `Retry` wrapper's `__call__`).  It is a semver break to call it `0.27.1` (a new feature, which means that code which uses it cannot revert to `0.27.0` if the change introduces a bug), but could be the pragmatic way to get the change deployed.

tseaver: @jonparrott 555bffa updates `consume_next` to use the `on_error` handling added in `core-0.27.1`.  PTAL

spanner/google/cloud/spanner/streamed.py
```diff
@@ -14,16 +14,39 @@
...
+"""A predicate that checks if an exception is a transient API error.
+
+For streaming the result of ``read`` / ``execute_sql`` requests, only
+the following server errors are considered transient:
+
+- :class:`google.api.core.exceptions.ServiceUnavailable` - HTTP 503, gRPC
+    ``UNAVAILABLE``.
+"""
+
+retry_unavailable = retry.Retry(predicate=if_unavailable_error)
```
jonparrott: Does all of this need to be public?

spanner/google/cloud/spanner/streamed.py
```diff
@@ -125,12 +155,29 @@ def _merge_values(self, values):
...
     def consume_next(self):
         """Consume the next partial result set from the stream.
 
         Parse the result set into new/existing rows in :attr:`_rows`
+
+        :raises ValueError:
+            if the sleep generator somehow does not yield values.
         """
-        response = six.next(self._response_iterator)
+        response = retry_unavailable(
```
jonparrott: I was on the fence about adding more stuff to `retry_target`, but this is just beautiful. Thanks for doing this.

spanner/google/cloud/spanner/streamed.py
```diff
@@ -14,16 +14,38 @@
...
 from google.protobuf.struct_pb2 import Value
-from google.cloud import exceptions
 from google.cloud.proto.spanner.v1 import type_pb2
 import six
 
 # pylint: disable=ungrouped-imports
 from google.cloud.spanner._helpers import _parse_value_pb
 # pylint: enable=ungrouped-imports
 
+_RESTART_DEADLINE = 30.0  # seconds
```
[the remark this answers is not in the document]
jonparrott: This is *not* the overall deadline for the stream, it's the deadline for opening the stream.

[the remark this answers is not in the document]
vkedia: Why do you need a deadline for that?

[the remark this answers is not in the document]
jonparrott: Because we do not want an infinite loop in a degenerate case.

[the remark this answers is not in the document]
vkedia: Actually this is no longer being used anywhere it seems.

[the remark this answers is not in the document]
jonparrott: The `Retry` object that replaced this still has a deadline (the default is two minutes).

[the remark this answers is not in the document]
vkedia: Why not just let user specify the overall deadline for the whole stream and don't retry once that has elapsed? 
I still don't understand how this deadline is used? Is it enforced by gRPC? Is it the deadline to receive the first packet on the stream?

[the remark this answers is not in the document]
lukesneeringer: You want a lower deadline for if you can not establish an initial connection at all. If your router is on fire, you should not be sitting there waiting for the total stream timeout.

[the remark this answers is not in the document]
jonparrott: There is an overall deadline for the stream but keep in mind that if the stream gets re-started then so does the deadline calculation - that part is separate from the deadline we're talking about here.

There's actually (lol) three separate retries at play here:

1. There's the retry-wrapped method to *establish* the stream. This has a deadline set by gapic.
2. There's the gRPC timeout which is set when the stream is established. This is the "overall stream deadline" you're thinking of. This is also set by gapic.
3. There's now this retry for re-establishing the stream. You can see this as sitting at a level above the other two things- it doesn't care or know about those settings. It just tries to re-open the stream if the iterator breaks. For sanity if has a deadline on retrying the stream on a single failure- if it succeeds in re-establishing the stream the deadline is no longer in play and is reset if the iterator breaks again after being re-established. In all likelyhood the deadline will never even be a factor as the retry in (1) should be able to establish a connection. The deadline is just there to keep a degenerate case from causing the app to block indefinitely.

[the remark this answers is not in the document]
vkedia: So in the case deadline in 3 elapses, what is the error code that user sees?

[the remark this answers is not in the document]
jonparrott: The user will see a `RetryError` with an `Unavailable` error as the cause.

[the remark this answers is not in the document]
vkedia: Ok. I guess this is fine for now. If we see users running into any issues due to this we can revisit.

lukesneeringer:
> One concern with this implementation is that if the stream breaks and the last packet yielded did not have a resume token, it cannot be retried. To guard against that in other languages (atleast in Java and Go), we buffer the packets till we see one with resume token and only then yield them to the user. 

Did we address this? (Also, Spanner should really be well behaved and always send a resume token, but whatever.)

tseaver:
> Did we address this?

Nope. If we're going to have to buffer rows until we get the *next* token, or the last partial response, then I think we need to shift the "restart" logic out from `consume_next` to `__iter__` (which already handles buffering rows).

> (Also, Spanner should really be well behaved and always send a resume token, but whatever.)

I still have no idea what engineering goal is being satisfied by that design.

vkedia:
Currently what happens if the stream breaks at a point where there is no resume token? Do you raise the error to the user or do you retry from some older resume token?

tseaver:
Because the back-end does not guarantee a resume token will be present in every partial result set, we have to buffer rows until either a) a new resume token arrives, or b) we hit the last partial set.

Therefore, I believe we cannot reasonably catch / handle the `UNAVAILABLE` inside `consume_next`:  instead, we should handle it inside `__iter__` (which already does row buffering).  I will re-work in that direction.

tseaver:
Because the back-end does not guarantee a resume token will be present in every partial result set, we have to buffer rows until either a) a new resume token arrives, or b) we hit the last partial set.

Therefore, I believe we cannot reasonably catch / handle the `UNAVAILABLE` inside `consume_next`:  instead, we should handle it inside `__iter__` (which already does row buffering).  I will re-work in that direction.

PR #3891, opened 2017-08-28
Add 'on_error' argument to 'retry.retry_target' and 'Retry'.

tseaver:
Permit application code to reset / fix-up state before retrying.

See: #3889

### DIFF
--- a/core/google/api/core/retry.py
+++ b/core/google/api/core/retry.py
@@ -150,7 +150,7 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
         sleep_generator (Iterator[float]): An infinite iterator that determines
             how long to sleep between retries.
         deadline (float): How long to keep retrying the target.
-        on_error(Callable): A function to call while processing a retryable
+        on_error (Callable): A function to call while processing a retryable
             exception.  Any error raised by this function will *not* be
             caught.
 
@@ -236,7 +236,7 @@ class Retry(object):
 
         Args:
             func (Callable): The callable to add retry behavior to.
-            on_error(Callable): A function to call while processing a
+            on_error (Callable): A function to call while processing a
                 retryable exception.  Any error raised by this function will
                 *not* be caught.
 
--- a/core/google/api/core/retry.py
+++ b/core/google/api/core/retry.py
@@ -135,7 +135,7 @@ def exponential_sleep_generator(
         delay = delay * multiplier
 
 
-def retry_target(target, predicate, sleep_generator, deadline):
+def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
     """Call a function and retry if it fails.
 
     This is the lowest-level retry helper. Generally, you'll use the
@@ -150,6 +150,9 @@ def retry_target(target, predicate, sleep_generator, deadline):
         sleep_generator (Iterator[float]): An infinite iterator that determines
             how long to sleep between retries.
         deadline (float): How long to keep retrying the target.
+        on_error(Callable): A function to call while processing a retryable
+            exception.  Any error raised by this function will *not* be
+            caught.
 
     Returns:
         Any: the return value of the target function.
@@ -177,6 +180,8 @@ def retry_target(target, predicate, sleep_generator, deadline):
             if not predicate(exc):
                 raise
             last_exc = exc
+            if on_error is not None:
+                on_error()
 
         now = datetime_helpers.utcnow()
         if deadline_datetime is not None and deadline_datetime < now:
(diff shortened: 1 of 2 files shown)


core/google/api/core/retry.py
```diff
@@ -150,6 +150,9 @@ def retry_target(target, predicate, sleep_generator, deadline):
         sleep_generator (Iterator[float]): An infinite iterator that determines
             how long to sleep between retries.
         deadline (float): How long to keep retrying the target.
+        on_error(Callable): A function to call while processing a retryable
```
dhermes: Want a space after `on_error`?

core/google/api/core/retry.py
```diff
@@ -226,11 +231,14 @@ def __init__(
         self._maximum = maximum
         self._deadline = deadline
 
-    def __call__(self, func):
+    def __call__(self, func, on_error=None):
         """Wrap a callable with retry behavior.
 
         Args:
             func (Callable): The callable to add retry behavior to.
+            on_error(Callable): A function to call while processing a
```
dhermes: Ditto here on space

core/tests/unit/api_core/test_retry.py
```diff
@@ -82,6 +82,34 @@ def target():
...
+        if call_count['target'] < 3:
+            raise ValueError()
+        return 42
+
+    result = retry.retry_target(
+        target, predicate, range(10), None, on_error=on_error)
+
+    assert result == 42
+    assert call_count['target'] == 3
+    assert call_count['on_error'] == 2
```
dhermes: You could just do an equality check for all of `call_count`

core/tests/unit/api_core/test_retry.py
```diff
@@ -82,6 +82,34 @@ def target():
...
+
+    sleep.assert_has_calls([mock.call(0), mock.call(1)])
```
dhermes: Also check the `call_count == 2`?

core/google/api/core/retry.py
```diff
@@ -177,6 +180,8 @@ def retry_target(target, predicate, sleep_generator, deadline):
             if not predicate(exc):
                 raise
             last_exc = exc
+            if on_error is not None:
+                on_error()
```
jonparrott: Should we pass in the `exc` to `on_error`?

core/google/api/core/retry.py
jonparrott: nit: Needs a space between `on_error` and `(Callable)`.

core/google/api/core/retry.py
```diff
@@ -226,11 +231,14 @@ def __init__(
         self._maximum = maximum
         self._deadline = deadline
 
-    def __call__(self, func):
+    def __call__(self, func, on_error=None):
```
jonparrott: Does it make sense to pass it in here or in the constructor? I can see the argument for it being here because it (so far) only makes sense for the function that actually uses the retry to specify this.

core/tests/unit/api_core/test_retry.py
tseaver: That assertion has so little obvious value that I just ignored it.

core/google/api/core/retry.py
tseaver: I tried it that way first, but then realized one might instantiate the `Retry` instance globally, and then use it in places where one wanted to use local-scope objects.

core/google/api/core/retry.py
tseaver: Good question:  without arguments, it is a "before retry" hook, rather than an "on error" hook.  I don't know if there would be something in the exception which would influence how it executed (e.g., trailing metadata?).

The motivating usecase (#3775) doesn't care about the exception, which is why I wrote it that way.

core/google/api/core/retry.py
jonparrott: Cool - makes sense.

core/google/api/core/retry.py
tseaver: e339f94

core/google/api/core/retry.py
tseaver: e339f94

core/google/api/core/retry.py
tseaver: e339f94

core/google/api/core/retry.py
jonparrott: Let's pass the exception unless you feel strongly we shouldn't?

jonparrott:
This is fine to merge once `on_error` is passed the `exc`. Then, @tseaver can use it to make #3889 much simpler.

core/google/api/core/retry.py
tseaver: ddcd917

PR #3904, opened 2017-08-30
Spanner GAPIC update.

lukesneeringer:
This is a regen of the GAPIC for spanner. Marking as `do not merge` because I need to go through and change the GAPIC import paths in the manual layer, and tweak the docs to provide the API reference. I can not easily do that from this particularly bad airplane wi-fi.


lukesneeringer:
Okay, this is now where I want it. Review at your leisure, @tseaver. :-)

spanner/google/cloud/spanner/streamed.py
```diff
@@ -61,7 +59,7 @@ def rows(self):
     def fields(self):
         """Field descriptors for result set columns.
 
-        :rtype: list of :class:`~google.cloud.proto.spanner.v1.type_pb2.Field`
+        :rtype: list of :class:`~google.cloud.proto.spanner.v1.types.Field`
```
tseaver: This seems inconsistent with the changes above, which use `google.cloud.spanner_v1.types`.

spanner/google/cloud/spanner/_helpers.py
```diff
@@ -20,9 +20,8 @@
 import six
 
 from google.gax import CallOptions
-from google.protobuf.struct_pb2 import ListValue
-from google.protobuf.struct_pb2 import Value
 from google.cloud.proto.spanner.v1 import type_pb2
+from google.cloud.spanner_v1 import types
```
tseaver: Is this a documented change to the Spanner API?  System tests have been passing all along with the non-Spanner-specific types protos, which *were* the documented forms previously.

spanner/google/cloud/spanner/streamed.py
lukesneeringer: Yep, this is wrong. Thanks.

spanner/google/cloud/spanner/_helpers.py
lukesneeringer: Both paths work. We updated the autogen to alias to a more user-friendly path for cases where we actually expose this directly to users.

spanner/google/cloud/spanner_v1/types.py
```diff
@@ -0,0 +1,45 @@
...
+
+names = []
+for module in (auth_pb2, http_pb2, keys_pb2, mutation_pb2, query_plan_pb2,
+               result_set_pb2, spanner_pb2, transaction_pb2, type_pb2,
+               descriptor_pb2, duration_pb2, empty_pb2, struct_pb2,
+               timestamp_pb2, ):
+    for name, message in get_messages(module).items():
+        message.__module__ = 'google.cloud.spanner_v1.types'
+        setattr(sys.modules[__name__], name, message)
+        names.append(name)
```
tseaver: I can't say I like this monkeying with the original module of the protos.  For one thing, we aren't documenting them here, and so hiding the real source makes it harder for users to find the original docs.

lukesneeringer:
Closing as we did a bunch of autogen changes that supercede this. Will open a similar one soon.

spanner/google/cloud/spanner_v1/types.py
lukesneeringer: > I can't say I like this monkeying with the original module of the protos. For one thing, we aren't documenting them here, and so hiding the real source makes it harder for users to find the original docs.

I am definitely trying to move toward documenting `types` consistently as `types`. It is much more usable than `random_thing_pb2`.

PR #3901, opened 2017-08-30
BigQuery Support for Sub Second Datetimes

page1:
I'm running into the issue mentioned here [#3870](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3870)

This patch seems to be working for me.

PR #3915, opened 2017-09-05
Use boolean instead of exception for BigQuery flow control.

dhermes:
This is a follow-up to #3901.

Also made sure to return an explicit None in the "field is null" branch. This follows the pattern of "all returns should be explicit or implicit, but don't mix".


dhermes:
This PR revealed a coverage miss: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3155

So that is :+1:, and I added a commit to fix it. :crossed_fingers:

PR #3924, opened 2017-09-06
Use queue if provided in kwargs

garykrige:
Bugfix

Fix for a new Queue being instantiated regardless of whether a queue is passed in or not.


garykrige:
I signed it!

lukesneeringer:
Good catch @garykrige! Thanks!

tseaver:
We really should add a test assertion that the passed-in queue is used.

PR #3927, opened 2017-09-06
Allowing `dict` (as an `Entity`) for property values.

### LINKED ISSUE #3923: Automagically convert dicts to datastore entities.
I just ran into the following error for the X'th time:

```python
    client.put(entity)
  File "lib/python3.6/site-packages/google/cloud/datastore/client.py", line 384, in put
    self.put_multi(entities=[entity])
  File "lib/python3.6/site-packages/google/cloud/datastore/client.py", line 408, in put_multi
    current.put(entity)
  File "lib/python3.6/site-packages/google/cloud/datastore/batch.py", line 199, in put
    _assign_entity_to_pb(entity_pb, entity)
  File "lib/python3.6/site-packages/google/cloud/datastore/batch.py", line 319, in _assign_entity_to_pb
    bare_entity_pb = helpers.entity_to_protobuf(entity)
  File "lib/python3.6/site-packages/google/cloud/datastore/helpers.py", line 219, in entity_to_protobuf
    _set_protobuf_value(value_pb, value)
  File "lib/python3.6/site-packages/google/cloud/datastore/helpers.py", line 420, in _set_protobuf_value
    _set_protobuf_value(i_pb, item)
  File "lib/python3.6/site-packages/google/cloud/datastore/helpers.py", line 408, in _set_protobuf_value
    attr, val = _pb_attr_value(val)
  File "lib/python3.6/site-packages/google/cloud/datastore/helpers.py", line 325, in _pb_attr_value
    raise ValueError("Unknown protobuf attr type %s" % type(val))
ValueError: Unknown protobuf attr type <class 'dict'>
```

The cause is pretty obvious, the following code snippet triggers this error:

```python
entity = datastore.Entity(client.key('Resource'))
entity.update(foo={'bar': 'baz'})
client.put(entity)
```

The code crashes, because `dict` type properties are not supported. Instead they should be converted to an `Entity`.

I think it's cumbersome having to convert `dict` types to `Entity` types every time. I've written a function that does the conversion recursively for production code, but for simple scripts this is still an issue.

I think it would be nice if this library would just do the conversion from `dict` to `Entity` automagically.

dhermes:
Fixes #3923.

### DIFF
--- a/datastore/google/cloud/datastore/entity.py
+++ b/datastore/google/cloud/datastore/entity.py
@@ -83,10 +83,12 @@ class Entity(dict):
     * :class:`~google.cloud.datastore.helpers.GeoPoint`
     * :data:`None`
 
-    In addition, two container types are supported:
+    In addition, three container types are supported:
 
     * :class:`list`
     * :class:`~google.cloud.datastore.entity.Entity`
+    * :class:`dict` (will just be treated like an ``Entity`` without
+      a key or ``exclude_from_indexes``)
 
     Each entry in a list must be one of the value types (basic or
     container) and each value in an
--- a/datastore/google/cloud/datastore/helpers.py
+++ b/datastore/google/cloud/datastore/helpers.py
@@ -290,8 +290,11 @@ def _pb_attr_value(val):
     >>> _pb_attr_value('my_string')
     ('string_value', 'my_string')
 
-    :type val: `datetime.datetime`, :class:`google.cloud.datastore.key.Key`,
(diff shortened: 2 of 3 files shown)


dhermes:
Man, Python 2 I am ready for you to disappear: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3177

dhermes:
@tseaver thanks for the review. @lukesneeringer Can you also take a look? I'm a bit worried I've over-simplified this fix.

PR #3930, opened 2017-09-07
Retry streaming exceptions (redux)

### LINKED ISSUE #3775: Cloud spanner client does not seem to be retrying broken streams

tseaver:
Handle double-buffering required by possibly-missing `resumeToken`.

Closes #3775.

### DIFF
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -89,14 +89,15 @@ class _SnapshotBase(_SessionWrapper):
 
         iterator = api.streaming_read(
             self._session.name, table, columns, keyset.to_pb(),
-            transaction=transaction, index=index, limit=limit,
-            options=options)
+            index=index, limit=limit,
+            transaction=transaction, options=options)
 
         self._read_request_count += 1
 
         restart = functools.partial(
             api.streaming_read, self._session.name, table, columns, keyset,
-            index=index, limit=limit)
+            index=index, limit=limit,
+            transaction=transaction, options=options)
 
         if self._multi_use:
             return StreamedResultSet(iterator, restart, source=self)
@@ -150,14 +151,15 @@ class _SnapshotBase(_SessionWrapper):
         api = database.spanner_api
         iterator = api.execute_streaming_sql(
             self._session.name, sql,
-            transaction=transaction, params=params_pb, param_types=param_types,
-            query_mode=query_mode, options=options)
+            params=params_pb, param_types=param_types, query_mode=query_mode,
+            transaction=transaction, options=options)
 
         self._read_request_count += 1
 
         restart = functools.partial(
             api.execute_streaming_sql, self._session.name, sql,
-            params=params, param_types=param_types, query_mode=query_mode)
+            params=params, param_types=param_types, query_mode=query_mode,
+            transaction=transaction, options=options)
 
         if self._multi_use:
             return StreamedResultSet(iterator, restart, source=self)
--- a/spanner/google/cloud/spanner/snapshot.py
+++ b/spanner/google/cloud/spanner/snapshot.py
@@ -51,8 +51,7 @@ class _SnapshotBase(_SessionWrapper):
         """
         raise NotImplementedError
 
-    def read(self, table, columns, keyset, index='', limit=0,
-             resume_token=b''):
+    def read(self, table, columns, keyset, index='', limit=0):
         """Perform a ``StreamingRead`` API request for rows in a table.
 
         :type table: str
@@ -70,9 +69,6 @@ class _SnapshotBase(_SessionWrapper):
 
         :type limit: int
         :param limit: (Optional) maxiumn number of rows to return
-
-        :type resume_token: bytes
-        :param resume_token: token for resuming previously-interrupted read
 
         :rtype: :class:`~google.cloud.spanner.streamed.StreamedResultSet`
         :returns: a result set instance which can be used to consume rows.
@@ -94,21 +90,20 @@ class _SnapshotBase(_SessionWrapper):
         iterator = api.streaming_read(
             self._session.name, table, columns, keyset.to_pb(),
             transaction=transaction, index=index, limit=limit,
-            resume_token=resume_token, options=options)
+            options=options)
 
         self._read_request_count += 1
 
         retry = functools.partial(
             api.streaming_read, self._session.name, table, columns, keyset,
-            index=index, limit=limit, resume_token=resume_token)
+            index=index, limit=limit)
 
         if self._multi_use:
             return StreamedResultSet(iterator, retry, source=self)
         else:
             return StreamedResultSet(iterator, retry)
 
-    def execute_sql(self, sql, params=None, param_types=None, query_mode=None,
-                    resume_token=b''):
+    def execute_sql(self, sql, params=None, param_types=None, query_mode=None):
         """Perform an ``ExecuteStreamingSql`` API request for rows in a table.
 
         :type sql: str
@@ -127,9 +122,6 @@ class _SnapshotBase(_SessionWrapper):
             :class:`google.cloud.proto.spanner.v1.ExecuteSqlRequest.QueryMode`
         :param query_mode: Mode governing return of results / query plan. See
             https://cloud.google.com/spanner/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest.QueryMode1
-
-        :type resume_token: bytes
-        :param resume_token: token for resuming previously-interrupted query
 
         :rtype: :class:`~google.cloud.spanner.streamed.StreamedResultSet`
         :returns: a result set instance which can be used to consume rows.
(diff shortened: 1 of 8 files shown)


spanner/google/cloud/spanner/streamed.py
```diff
@@ -122,17 +130,49 @@ def _merge_values(self, values):
...
 
+    def _flush_pending_rows(self):
+        """Helper for :meth:`consume_next`."""
+        flushed = self._pending_rows[:]
+        self._pending_rows[:] = ()
+        self._complete_rows.extend(flushed)
+
+    def _do_restart(self):
+        """Helper for :meth:`consume_next`."""
+        if not self._resume_token:
```
vkedia: _resume_token might be absent because the stream broke before yielding the first response or it yielded a few responses none of which had resume token. In either case you can just restart the stream from beginning instead of raising error.

spanner/google/cloud/spanner/streamed.py
```diff
@@ -122,17 +130,49 @@ def _merge_values(self, values):
...
+            raise ValueError("No resume token")
+
+        self._pending_chunk = None
+        self._pending_rows[:] = ()
+        self._current_row[:] = ()
+        self._response_iterator = self._restart(self._resume_token)
+
     def consume_next(self):
```
vkedia: Can you add some unit tests for this class for the following cases:

1. Stream does not break.
2. Stream breaks before yielding any results.
3. Stream break after yielding a few results none of which had resume token.
4. Stream breaks when the last result had a resume token.
5. Stream breaks after a few rows have been yielded to the user and when the last result did not have a resume toekn

Any other cases you can think of.

spanner/google/cloud/spanner/streamed.py
lukesneeringer: > or it yielded a few responses none of which had resume token

Seriously? Ouch.

spanner/google/cloud/spanner/streamed.py
tseaver: @vkedia 7d83816 updates the tests to assert that the `_do_restart` helper calls the `_restart` partial function without any arguments when the instance has no token yet.  The other cases are covered by existing unit tests.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -153,14 +151,18 @@ def execute_sql(self, sql, params=None, param_types=None, query_mode=None,
...
             self._session.name, sql,
             transaction=transaction, params=params_pb, param_types=param_types,
-            query_mode=query_mode, resume_token=resume_token, options=options)
+            query_mode=query_mode, options=options)
 
         self._read_request_count += 1
 
+        restart = functools.partial(
+            api.execute_streaming_sql, self._session.name, sql,
+            params=params, param_types=param_types, query_mode=query_mode)
```
vkedia: you need to also pass in the transaction here.

spanner/google/cloud/spanner/snapshot.py
```diff
@@ -92,17 +90,20 @@ def read(self, table, columns, keyset, index='', limit=0,
         iterator = api.streaming_read(
             self._session.name, table, columns, keyset.to_pb(),
             transaction=transaction, index=index, limit=limit,
-            resume_token=resume_token, options=options)
+            options=options)
 
         self._read_request_count += 1
 
+        restart = functools.partial(
+            api.streaming_read, self._session.name, table, columns, keyset,
```
vkedia: you need to also pass in the transaction here.

spanner/google/cloud/spanner/streamed.py
```diff
@@ -122,17 +130,50 @@ def _merge_values(self, values):
...
+    def _do_restart(self):
+        """Helper for :meth:`consume_next`."""
+        self._pending_chunk = None
+        self._pending_rows[:] = ()
+        self._current_row[:] = ()
+
+        if self._resume_token:
+            self._response_iterator = self._restart(self._resume_token)
+        else:
+            self._response_iterator = self._restart()
```
vkedia: What if you have already yielded some rows to the user but there is no resume token. In that case you can't just do a restart from beginning. You need to differentiate between 3 cases here:
1. There is a resume token.
2. There is no resume token but you have not yielded anything to the user.
3. There is no resume token but you have yielded one or more rows to the user.

In 1 and 2 it is safe to resume. In 3 you need to just raise an error to the user.

spanner/google/cloud/spanner/streamed.py
```diff
@@ -122,17 +130,50 @@ def _merge_values(self, values):
...
+        """Helper for :meth:`consume_next`."""
+        flushed = self._pending_rows[:]
+        self._pending_rows[:] = ()
+        self._complete_rows.extend(flushed)
+
+    def _do_restart(self):
+        """Helper for :meth:`consume_next`."""
+        self._pending_chunk = None
+        self._pending_rows[:] = ()
+        self._current_row[:] = ()
```
vkedia: It is not safe to reset all these under all conditions. It depends on where you got the resume token. Essentially you need to clear out everything you got after the resume token. But some of this might have come before the resume token. For eg:
Row just has a single string column and these are the partial result sets you get

PRS1: value: "h", chunked: true, resume_token: "zjgfdkjf"
PRS2: value: "e", chunked: true, resume_token: Null

Say when you try to read next PRS3 the stream breaks. So you resume using the token yielded by PRS1. The resumed stream will yield PRS2 onwards. But since you cleared pending_chunk, you lost PRS1.

spanner/google/cloud/spanner/streamed.py
tseaver: We no longer yield rows to the user until we receive either a token or the last partial result set (the one with stats).  The not-yet-yielded rows are in the `self._pending_rows` attribute, cleared in line 145.  Those rows get copied into `self._completed_rows` via `self._flush_pending_rows()`, which is invoked only in those two cases (and the oddball edge case that the streaming iterator raises `StopIteration` without sending the last PRS).

spanner/google/cloud/spanner/streamed.py
vkedia: I think this will be easier to implement if you keep on buffering PRS till you see a resume token and only after that merge them/parse them.

spanner/google/cloud/spanner/snapshot.py
tseaver: Should it pass `options` too?

spanner/google/cloud/spanner/streamed.py
vkedia: But even after you have yielded rows to the user, you might get more PRS's without resume tokens.
For eg say for a query with single string column, you get
PRS1: value: "hello", chunked: false, resume_token:"blah".
PRS2: value: "world", chunked: false, resume_token:Null

You will yield row2 to user after seeing PRS1. But PRS2 does not have resume token and if stream breaks after that you will again resume from beginning, thus yielding row1 to the user again.

spanner/google/cloud/spanner/snapshot.py
tseaver: I'm assuming likewise for `options` here.

spanner/google/cloud/spanner/streamed.py
vkedia: :) How do you strikethrough text here?

spanner/google/cloud/spanner/snapshot.py
vkedia: I guess so. What does `options` have?

tseaver: 8f47fe2

spanner/google/cloud/spanner/snapshot.py
tseaver: 8f47fe2

spanner/google/cloud/spanner/snapshot.py
tseaver: Options contains the metadata prefix (with the database name).

spanner/google/cloud/spanner/streamed.py
lukesneeringer: `~` is the strikethrough character in Markdown.

tseaver: > How do you strikethrough text here?

Use "~~" on either side.

tseaver:
@vkedia Can you please point me to docs which specify the semantics of `resume_token`?  The API docs for `PartialResultSet` are ambiguous as to its meaning.

vkedia:
What exactly is ambiguous? I think it can be improved to mention that it is not guaranteed that `resume_token` will be present in every `PartialResultSet`. Other than that, is there something else that is not clear?

vkedia:
@tseaver  Any updates here?

lukesneeringer:
I was just about to pop in and ask the exact same question. :-)

bjwatson:
@tseaver Any updates? This is the primary blocker for us declaring Beta on the Python Spanner client library.

tseaver:
Superseded by #4016.

PR #3938, opened 2017-09-08
bigquery add DatasetReference class and tests

alixhami:


alixhami:
I signed it!

tseaver:
@tswast Is the PR a first step toward something bigger?  In other words, how will a `DatasetReference` be used, and how is it different from a `Dataset`?  Right now, `DatasetReference` is basically just a more verbose "named tuple" (a read only pair of attributes).

tswast:
Yes, it is the first step in something bigger. (I've shared a doc called BigQuery Python Client Libraries Redesign Proposal with you.)

The most important use of this class is that there are places in the returned API resources where we just get a reference to a table or dataset. Rather than have a Table or Dataset class with no properties besides the ID filled in, I propose using reference classes to be more clear that they are just pointers. This would prevent confusion such as https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3754

The reference classes would also be used as the argument to new methods such as `client.get_dataset(dataset_ref)` and `client.get_table(table_ref)`.

PR #3944, opened 2017-09-12
BigQuery: Modify client.dataset() to return DatasetReference instead of Dataset.

alixhami:
The base of this commit is https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3942

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -18,6 +18,7 @@ from google.api.core import page_iterator
 from google.cloud.client import ClientWithProject
 from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery.dataset import Dataset
+from google.cloud.bigquery.dataset import DatasetReference
 from google.cloud.bigquery.job import CopyJob
 from google.cloud.bigquery.job import ExtractJob
 from google.cloud.bigquery.job import LoadJob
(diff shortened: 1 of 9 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -48,6 +48,42 @@
...
+    :type table_id: str
+    :param table_id: the ID of the table
+    """
+
+    def __init__(self, dataset_ref, table_id):
+        self._dataset_ref = dataset_ref
+        self._table_id = table_id
+
+    @property
+    def dataset_ref(self):
```
jba: Should this be called `dataset`, for consistency with `client.dataset`?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -148,7 +149,7 @@ def list_datasets(self, include_all=False, max_results=None,
             extra_params=extra_params)
 
     def dataset(self, dataset_name, project=None):
-        """Construct a dataset bound to this client.
+        """Construct a reference to a dataset bound to this client.
```
jba: It's not really "bound to this client." It doesn't contain the client, and could be used across clients.

bigquery/google/cloud/bigquery/table.py
tswast: We are documenting it as a DatasetReference, so it wouldn't be as likely to be confused with the Dataset class, which was my original fear.

PR #3945, opened 2017-09-13
Parse timestamps in query parameters using canonical format

tswast:
The timestamp format in query parameters follows the [canonical format](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp-type).

This fixes a system test error which was happening in the bigquery-b2 branch.

/cc @alixhami

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -28,10 +28,22 @@ from google.cloud._helpers import _to_bytes
 
 _RFC3339_MICROS_NO_ZULU = '%Y-%m-%dT%H:%M:%S.%f'
 
-# Canonical formats for timestamps in BigQuery
+# Canonical formats for timestamps in BigQuery.
+# Timestamps in an API response will always be in UTC, but they
+# can vary in other ways. The separator between the date and time
+# can be 'T' or ' ', the microseconds may be present or not, and
+# the UTC timezone may be formatted as Z or +00:00. See:
 # g.co/cloud/bigquery/docs/reference/standard-sql/data-types#timestamp-type
-_BIGQUERY_TIMESTAMP = '%Y-%m-%d %H:%M:%S%z'
-_BIGQUERY_TIMESTAMP_MICROS = '%Y-%m-%d %H:%M:%S.%f%z'
+_BIGQUERY_TIMESTAMP_FORMATS = [
+    '%Y-%m-%d %H:%M:%S+00:00',
+    '%Y-%m-%d %H:%M:%SZ',
+    '%Y-%m-%d %H:%M:%S.%f+00:00',
+    '%Y-%m-%d %H:%M:%S.%fZ',
+    '%Y-%m-%dT%H:%M:%S+00:00',
+    '%Y-%m-%dT%H:%M:%SZ',
+    '%Y-%m-%dT%H:%M:%S.%f+00:00',
+    '%Y-%m-%dT%H:%M:%S.%fZ',
+]
 
 
 def _not_null(value, field):
@@ -88,16 +100,16 @@ def _timestamp_query_param_from_json(value, field):
         :data:`None`).
     """
     if _not_null(value, field):
-        # Remove final : from timezone because python represents UTC offset
-        # as +HHMM or -HHMM, whereas BigQuery represents it as +HH:MM or -HH:MM
-        value = ''.join(value.rsplit(':', 1))
-        if '.' in value:
-            # YYYY-MM-DDTHH:MM:SS.ffffff+00:00
-            return datetime.datetime.strptime(
-                value, _BIGQUERY_TIMESTAMP_MICROS)
-        else:
-            # YYYY-MM-DDTHH:MM:SS+00:00
-            return datetime.datetime.strptime(value, _BIGQUERY_TIMESTAMP)
+        error = None
+
+        for timefmt in _BIGQUERY_TIMESTAMP_FORMATS:
+            try:
+                return datetime.datetime.strptime(
+                    value, timefmt).replace(tzinfo=UTC)
+            except ValueError as exc:
+                error = exc
+
+        raise error
     else:
         return None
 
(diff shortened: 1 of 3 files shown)


tswast:
System tests running in branch at https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3223

tswast:
Ugh. Failing in Python 2.7, but I swear I see `%z` in the [datetime docs](https://docs.python.org/2/library/datetime.html)

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -28,6 +28,11 @@
 
 _RFC3339_MICROS_NO_ZULU = '%Y-%m-%dT%H:%M:%S.%f'
 
+# Canonical formats for timestamps in BigQuery
+# g.co/cloud/bigquery/docs/reference/standard-sql/data-types#timestamp-type
+_BIGQUERY_TIMESTAMP = '%Y-%m-%d %H:%M:%S%z'
+_BIGQUERY_TIMESTAMP_MICROS = '%Y-%m-%d %H:%M:%S.%f%z'
```
tseaver: `%z` is one of those supported-by-`strftime`-but-not-`strptime` directives in Python2. :(

If new the zone would always be UTC, we could just strip off the `+00:00` and add the tzinfo ourselves.  As it is, the spec makes it super-complicated:  not only to we have to parse the offset, but we also have to be prepared to deal with zone names.

tseaver: @tswast unless, of course, you can assure us that the back-end will normalize everything to UTC and always send it back as `+00:00`. :)

tswast: True, I'll check with the backend folks to see if this will always be UTC when coming from the API.

tseaver: This passes on 2.7 and 3.x, but we have a coverage gap (any non-null value with non `+00:00` zone):

```python
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -30,8 +30,11 @@ _RFC3339_MICROS_NO_ZULU = '%Y-%m-%dT%H:%M:%S.%f'
 
 # Canonical formats for timestamps in BigQuery
 # g.co/cloud/bigquery/docs/reference/standard-sql/data-types#timestamp-type
+_BIGQUERY_TIMESTAMP_NO_OFFSET = '%Y-%m-%d %H:%M:%S'
 _BIGQUERY_TIMESTAMP = '%Y-%m-%d %H:%M:%S%z'
+_BIGQUERY_TIMESTAMP_MICROS_NO_OFFSET = '%Y-%m-%d %H:%M:%S.%f'
 _BIGQUERY_TIMESTAMP_MICROS = '%Y-%m-%d %H:%M:%S.%f%z'
+_ZULU_NO_OFFSET = '+00:00'
 
 
 def _not_null(value, field):
@@ -88,16 +91,17 @@ def _timestamp_query_param_from_json(value, field):
         :data:`None`).
     """
     if _not_null(value, field):
-        # Remove final : from timezone because python represents UTC offset
-        # as +HHMM or -HHMM, whereas BigQuery represents it as +HH:MM or -HH:MM
-        value = ''.join(value.rsplit(':', 1))
-        if '.' in value:
-            # YYYY-MM-DDTHH:MM:SS.ffffff+00:00
-            return datetime.datetime.strptime(
-                value, _BIGQUERY_TIMESTAMP_MICROS)
-        else:
-            # YYYY-MM-DDTHH:MM:SS+00:00
-            return datetime.datetime.strptime(value, _BIGQUERY_TIMESTAMP)
+
+        if value.endswith(_ZULU_NO_OFFSET):
+            value = value[:-len(_ZULU_NO_OFFSET)]
+            if '.' in value:
+                return datetime.datetime.strptime(
+                    value, _BIGQUERY_TIMESTAMP_MICROS_NO_OFFSET).replace(
+                        tzinfo=UTC)
+            else:
+                return datetime.datetime.strptime(
+                    value, _BIGQUERY_TIMESTAMP_NO_OFFSET).replace(
+                        tzinfo=UTC)
     else:
         return None
 
```

tswast: I got confirmation that the timestamp will always be encoded as UTC.

The internal protos say it'll have trailing Z instead of +00:00, though, so I'll do a little more digging.

tswast: I've pushed a new commit. There are 8 possible formats that the BigQuery API might give us that we'll have to handle.

- 'T' vs ' ' between time and date.
- 'Z' vs '+00:00' for representing UTC time.
- microseconds or no microseconds.

My solution is to just try them all, but maybe we should be constructing the time format string on the fly instead?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -28,10 +28,22 @@
...
+_BIGQUERY_TIMESTAMP_FORMATS = [
+    '%Y-%m-%d %H:%M:%S+00:00',
+    '%Y-%m-%d %H:%M:%SZ',
+    '%Y-%m-%d %H:%M:%S.%f+00:00',
+    '%Y-%m-%d %H:%M:%S.%fZ',
+    '%Y-%m-%dT%H:%M:%S+00:00',
+    '%Y-%m-%dT%H:%M:%SZ',
+    '%Y-%m-%dT%H:%M:%S.%f+00:00',
+    '%Y-%m-%dT%H:%M:%S.%fZ',
+]
```
tseaver: - If we *know for certain* that the backend will send only UTC timestamps, we can just crack off the trailing `Z` / `+00:00`.

- Given that the spec is about the formats the back-end will *accept*, they should be able to promise that they will send *back* the timestamp using exactly *one* format (` ` vs. `T` separator, `Z` vs. `+00:00`).  [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle) applies.

tswast: (1) Okay. I'll do some chopping.

(2) Agreed, but I wasn't able to get the backend team to commit. The protos promise one format, but the actual system behavior is different.

tswast: Chopping done. I did some string replacing too for `' '` vs `'T'`. Wasn't as complicated as I expected.

tswast:
System tests for latest changes running at https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3260

tswast:
I'm having trouble loading the Circle CI logs for the failure. Probably due to the S3 outage today.

PR #3949, opened 2017-09-13
docs/bigquery/usage.rst: Remove incorrect note about UUIDs

evanj:
The following note was listed under "Inserting data (asynchronous)":

* google.cloud.bigquery generates a UUID for each job.

This is incorrect: run_async_query requires a name, which gets passed
as the jobId field to the REST API. For the synchronous case,
run_sync_query calls the query REST API which does not require an ID.

The BigQuery API currently makes no use of UUIDs, so remove the note.

### DIFF
--- a/docs/bigquery/usage.rst
+++ b/docs/bigquery/usage.rst
@@ -273,7 +273,6 @@ Background a query, loading the results into a table:
 
 .. note::
 
-   - ``google.cloud.bigquery`` generates a UUID for each job.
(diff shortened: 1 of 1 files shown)


lukesneeringer:
Hi @evanj! Thanks for the note.
I thought we added the feature you describe recently... :-/

lukesneeringer:
Sure enough, I do not see it. Thanks.

PR #3950, opened 2017-09-13
Pin google-auth >= 1.1.0.

### LINKED ISSUE #3393: Support a query over an external data source / federated data source
Bigquery supports running queries over csv files, but I don't see this option in this bigquery api. 

See https://cloud.google.com/bigquery/external-data-sources
and
https://cloud.google.com/bigquery/external-data-cloud-storage

I see example for reading csv files into a table, but I would rather not have to manage creating and deleting temporary tables.

tseaver:
Closes #3393.

### DIFF
--- a/spanner/setup.py
+++ b/spanner/setup.py
@@ -52,6 +52,7 @@ SETUP_BASE = {
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.27.0, < 0.28dev',
+    'google-auth >= 1.1.0',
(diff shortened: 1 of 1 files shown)


dhermes:
@jonparrott @tseaver Worth also putting this pin in `google-cloud-core`? (I realize we don't want to gate a release on `google-cloud-core`, but it's worth doing it for the next release of `google-cloud-core`?)

jonparrott:
I have no strong feelings either way.

PR #3952, opened 2017-09-13
Vision code re-organization

lukesneeringer:
This PR moves the Vision code around to fit the new auto-gen (which it also updates). Should be a no-op provided tests pass.


dhermes:
Seems sane. Let's see what CI says?

lukesneeringer:
The manual change I had to make in 12e2028 now has an in-flight fix in https://github.com/googleapis/toolkit/pull/1563.

PR #3953, opened 2017-09-13
bigquery: rename TableReference.dataset_ref

jba:
Rename to dataset to be consistent with Client.dataset. Both
methods actually return a DatasetReference.


dhermes:
@jba LGTM modulo the missing test changes (The unit tests for the property also need to be updated, I'm assuming you've seen the CI failure)

tswast:
I imagine there are unit tests to update too?

PR #3955, opened 2017-09-14
bigquery: rename name field of Dataset to dataset_id

jba:
Rename the former dataset_id property to full_dataset_id.

Also rename Table.dataset_name to Table.dataset_id.

Perform other renamings (of various variables and constants).

These names match usage better. The API's Dataset.id field is
"project:dataset_id", which is confusing and basically useless,
so it's a mistake to call that dataset_id.

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -142,8 +142,8 @@ class Dataset(object):
     See
     https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets
 
-    :type name: str
-    :param name: the name of the dataset
+    :type dataset_id: str
+    :param dataset_id: the ID of the dataset
 
     :type client: :class:`google.cloud.bigquery.client.Client`
     :param client: A client which holds credentials and project configuration
@@ -159,8 +159,8 @@ class Dataset(object):
 
     _access_entries = None
 
-    def __init__(self, name, client, access_entries=(), project=None):
-        self.name = name
+    def __init__(self, dataset_id, client, access_entries=(), project=None):
+        self.dataset_id = dataset_id
         self._client = client
         self._properties = {}
         # Let the @property do validation.
@@ -181,9 +181,9 @@ class Dataset(object):
         """URL path for the dataset's APIs.
 
         :rtype: str
-        :returns: the path based on project and dataste name.
+        :returns: the path based on project and dataset ID.
         """
-        return '/projects/%s/datasets/%s' % (self.project, self.name)
+        return '/projects/%s/datasets/%s' % (self.project, self.dataset_id)
 
     @property
     def access_entries(self):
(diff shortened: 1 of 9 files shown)


jba:
Having done this, I'm a bit bothered by the stutter of `dataset.dataset_id`. What about renaming the field to `id`? Is that idiomatic Python?

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -221,8 +221,8 @@ def created(self):
             return _datetime_from_microseconds(1000.0 * creation_time)
 
     @property
-    def dataset_id(self):
-        """ID for the dataset resource.
+    def full_dataset_id(self):
```
lukesneeringer: It is a little whiplash-y to me that what _used_ to be spelled `dataset_id` is now spelled `full_dataset_id`, and what _used_ to be spelled `name` is now spelled `dataset_id`.

Not opposing outright, but is there a just-as-good terminology that would might be a bit easier on folks?

If the answer is no, then charge ahead. :-)

lukesneeringer: Follow-up: I _definitely_ do like that `dataset_id` matches the server name, though, which is (I assume) why it is being used here. Okay, I think I am tracking why this decision is made.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -463,7 +463,7 @@ def list_partitions(self, client=None):
         """
         query = self._require_client(client).run_sync_query(
```
lukesneeringer: Possibly out of scope, but does `run_sync_query` have a DB-API style `params` substitution that we could use in lieu of `__mod__` here? Fine if not.

lukesneeringer:
> Having done this, I'm a bit bothered by the stutter of `dataset.dataset_id`. What about renaming the field to `id`? Is that idiomatic Python?

Yes. I would also like that it neatly solves the problem I mentioned in my previous comment.

tseaver:
> Having done this, I'm a bit bothered by the stutter of dataset.dataset_id. What about renaming the field to id? Is that idiomatic Python?

We have historically avoided using bare `id` as an attribute, both to avoid the (appearance of) shadowing the `id` built-in, and also because it makes the semantics clearer.

Re: stutter:  there is no requirement that instances be assigned to variables named `dataset`.  E.g., consider a notional "copy" operation, with `source` and `destination` as the names for the two dataset instances.

bigquery/google/cloud/bigquery/table.py
jba: Not only out of scope, but I haven't yet learned about DB-API so I don't really understand the question :)

jba:
OK, will leave as-is. Waiting for @tswast to approve.

bigquery/google/cloud/bigquery/table.py
tswast: We can't use query parameters in the the table identifiers anyway.

PR #3960, opened 2017-09-14
updated docstrings for method ``table.insert_data``

WillianFuks:
Hello,

Just recently a [question](https://stackoverflow.com/questions/46222247/should-bigquery-throw-an-error-using-table-insert-data-with-skip-invalid-rows) as asked on Stackoverflow about the method `table.insert_data` and why the docstrings had question marks ("?").

I decided to make this quick PR to update the docstrings so to explain how each parameter works.


dhermes:
@WillianFuks You can ignore the CI failure in this case, it's not from your changes.

@tseaver Worth noting that people are thrown off by descriptions of booleans as questions.

lukesneeringer:
Thanks @WillianFuks!
I merged in master, which fixes CI. I will merge once CI passes.

PR #3963, opened 2017-09-14
Natural Language: Auto-gen update.

lukesneeringer:
Uses the `nl` branch as a base.
A follow-up PR is coming momentarily to add the new GAPICs.

Unit tests _will_ fail on this PR since it is an "in-flight". They will pass once #3964 is merged into it. The PRs are separated to make it (nominally) easier to review.


lukesneeringer:
Not waiting on tests since #3964 succeeded.

Issue #3956, opened 2017-09-14
Add support for storage->pubsub notifications


tseaver:
See the [concepts doc](https://cloud.google.com/storage/docs/pubsub-notifications) and the [registering object changes doc](https://cloud.google.com/storage/docs/reporting-changes).

Pull requests target the [`3956-storage-notifications` branch](https://github.com/GoogleCloudPlatform/google-cloud-python/tree/3956-storage-notifications)

Steps:

- [x] Add a `BucketNotification` class, mapping the [notification resource](https://cloud.google.com/storage/docs/json_api/v1/notifications#resource).  Note that a `BucketNotification` may point to a topic which belongs to a different `project` than the one which owns its bucket. (#3957)
- [x] Add `Bucket.notification` factory. returning new `BucketNotification`. (#3958)
- [x] Add `BucketNotification.create` method, mapping [notifications/insert](https://cloud.google.com/storage/docs/json_api/v1/notifications/insert).  Update the bucket's properties from the response (e.g., to get its server-generated ID).  Should we barf if the ID is already set? (#3966)
- [x] Add `BucketNotification.delete` method, mapping [notifications/delete](https://cloud.google.com/storage/docs/json_api/v1/notifications/delete).  Presumably barf if the ID is not set. (#3983)
- [x] Add `BucketNotification.exists` method, mapping [notifications/get](https://cloud.google.com/storage/docs/json_api/v1/notifications/get) onto boolean (False for `NotFound`).  Presumably barf if the ID is not set. (#3984)
- [x] Add `BucketNotification.reload` method, mapping [notifications/get](https://cloud.google.com/storage/docs/json_api/v1/notifications/get),  Presumably barf if the ID is not set. (#3986)
- [x] Add `Bucket.list_notifications` method, mapping [notifications/list](https://cloud.google.com/storage/docs/json_api/v1/notifications/list).  Return an iterator of `BucketNotification` instances. (#3990)
- [ ] Add a system test which exercises the new API wrapper methods. (#4021)

Questions:

- Do we need to support *de novo* creation of `BucketNotfication` instances w/ a user-supplied ID?  The spec says that the ID is generated during `notifications/insert`, which suggests not.
[...]

PR #3966, opened 2017-09-15
Add 'BucketNotification.create' API wrapper.

tseaver:
Toward #3956.


storage/google/cloud/storage/notification.py
```diff
@@ -22,6 +22,8 @@
 JSON_API_V1_PAYLOAD_FORMAT = 'JSON_API_V1'
 NONE_PAYLOAD_FORMAT = 'NONE'
 
+_TOPIC_REF = '//pubsub.googleapis.com/projects/{}/topics/{}'
```
lukesneeringer: It is fine to not have a protocol? (I know that is valid now, so not complaining; just ensuring it is an intentional decision.)

tseaver: The format is [specified in the docs for `notifications/insert](https://cloud.google.com/storage/docs/json_api/v1/notifications/insert#topic).

PR #3968, opened 2017-09-15
Add AnalyzeEntitySentiment method to language v1.

lukesneeringer:


lukesneeringer:
Bypassing CircleCI since the target branch is `nl`, will run it there.

PR #3969, opened 2017-09-15
Natural Language

lukesneeringer:
This PR is an amalgamation of:

  * #3963 (which shuffles the auto-gen code around)
  * #3968 (which adds `analyze_entity_sentiment` as a method)
  * #3970 (which removes the manual layer)
  * #3972 (which bumps the version number)

PR #3972, opened 2017-09-15
Natural Language 0.29.0

lukesneeringer:
Uses #3969 as a base.

(release notes coming shortly)


lukesneeringer:
Putting release notes in #3969 instead of here.

PR #3973, opened 2017-09-15
BigQuery: Adds client.get_dataset() and removes dataset.reload()

alixhami:


bigquery/google/cloud/bigquery/client.py
```diff
@@ -166,6 +166,20 @@ def dataset(self, dataset_name, project=None):
 
         return DatasetReference(project, dataset_name)
 
+    def get_dataset(self, dataset_ref):
+        """API call:  refresh dataset properties via a GET request.
```
jba: s/refresh/read/?

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -127,6 +127,15 @@ def dataset_id(self):
         """
         return self._dataset_id
 
+    @property
+    def path(self):
```
jba: Huh, this method already exists on the branch...

bigquery/tests/unit/test_client.py
```diff
@@ -272,6 +272,32 @@ def test_dataset_with_default_project(self):
...
         self.assertEqual(dataset.project_id, PROJECT)
 
+    def test_get_dataset(self):
+        project = 'PROJECT'
+        dataset_id = 'dataset_id'
+        path = 'projects/%s/datasets/%s' % (project, dataset_id)
+        creds = _make_credentials()
+        http = object()
+        client = self._make_one(project=project, credentials=creds, _http=http)
+        DS_ID = '%s:%s' % (project, dataset_id)
```
jba: Since you don't use this variable again, maybe just put the % expression in the resource dict?

bigquery/google/cloud/bigquery/client.py
tswast: How about "Fetch the dataset referenced by ``dataset_ref``" to match the docstring for `get_job()`?

bigquery/google/cloud/bigquery/dataset.py
tswast: Ah, I see this appears lower down at line 180. https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3973/files#diff-bafa3d472438f0a46853cc9879265cadR188

This is a duplicate definition and should be removed.

alixhami: One of the definitions is for DatasetReference, and one of them is for Dataset. We added the path method to DatasetReference because client.get_dataset() uses it.

tswast: Oops, you're right. LGTM.

bigquery/tests/unit/test_client.py
alixhami: Made change with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3973/commits/811c9b03d5825d514347e16ff43d8e1948906a2f

bigquery/google/cloud/bigquery/client.py
alixhami: Made change with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3973/commits/811c9b03d5825d514347e16ff43d8e1948906a2f

PR #3974, opened 2017-09-15
Video Intelligence: Remove previous auto-gen.

lukesneeringer:
This removes the previous auto-gen surface from video intelligence. The tests will fail until a follow-up PR that is based on this one is merged in.

Equivalent of #3963, but for video intelligence.


dhermes:
ASIDE: @lukesneeringer Could you have more descriptive PR names? Here really would just be nice to say which package is affected. (The end of the monorepo would make it moot.)

lukesneeringer:
Re-opening a new PR not on my fork.

PR #3983, opened 2017-09-15
Add 'BucketNotification.delete' API wrapper.

tseaver:
Toward #3956.

### DIFF
--- a/storage/google/cloud/storage/notification.py
+++ b/storage/google/cloud/storage/notification.py
@@ -190,6 +190,9 @@ class BucketNotification(object):
     def delete(self, client=None):
         """Delete this notification.
 
+        See:
+        https://cloud.google.com/storage/docs/json_api/v1/notifications/delete
+
         :type client: :class:`~google.cloud.storage.client.Client` or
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/notification.py
```diff
@@ -180,3 +186,23 @@ def create(self, client=None):
...
+        See:
+        https://cloud.google.com/storage/docs/json_api/v1/notifications/delete
+
+        :type client: :class:`~google.cloud.storage.client.Client` or
+                      ``NoneType``
+        :param client: Optional. The client to use.  If not passed, falls back
+                       to the ``client`` stored on the current bucket.
+
+        :raises: :class:`google.api.core.exceptions.NotFound`:
+            if the notification does not exist.
```
dhermes: Shouldn't this be a `ValueError`?

tseaver: Nope, this is a case where we are propagating the 404.  The next PR fixes this docstring, adding the `ValueError` for the "id not set" case.

PR #3985, opened 2017-09-15
Video Intelligence import issue

lukesneeringer:
Auto-generated change that fixes an issue with importing Video Intelligence previously. (I have no idea how the unit tests passed before.)


lukesneeringer:
I know why! Because we do not run video intelligence tests in CI. Good times.

Fixing that first thing Monday morning.

lukesneeringer:
I am wondering the same thing. We _need_ a better process. This is a problem for every language, also.

The actual issue behind this one was maddening -- it turns out that protoc on Mac OS X and protoc on the (Ubuntu) Docker image do not actually output the same thing (!!!), and one of the outputs does not work (!!!).

PR #3990, opened 2017-09-18
Add 'Bucket.list_notifications' API wrapper.

tseaver:
Uses #3983 as a base.

Toward #3956.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -425,7 +425,7 @@ class Bucket(_PropertyMixin):
         return iterator
 
     def list_notifications(self, client=None):
-        """List pubsub notifications for this bucket.
+        """List Pub / Sub notifications for this bucket.
 
         See:
         https://cloud.google.com/storage/docs/json_api/v1/notifications/list
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -77,6 +77,26 @@ def _item_to_blob(iterator, item):
     return blob
 
 
+def _item_to_notification(iterator, item):
+    """Convert a JSON blob to the native object.
+
+    .. note::
+
+        This assumes that the ``bucket`` attribute has been
+        added to the iterator after being created.
+
+    :type iterator: :class:`~google.api.core.page_iterator.Iterator`
+    :param iterator: The iterator that has retrieved the item.
+
+    :type item: dict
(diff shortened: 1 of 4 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -405,6 +424,30 @@ def list_blobs(self, max_results=None, page_token=None, prefix=None,
         iterator.prefixes = set()
         return iterator
 
+    def list_notifications(self, client=None):
+        """List pubsub notifications for this bucket.
```
dhermes: "Pub / Sub"

storage/google/cloud/storage/notification.py
```diff
@@ -82,6 +90,40 @@ def __init__(self, bucket, topic_name,
...
+        :param resource: JSON repr of the notification
+
+        :type bucket: :class:`google.cloud.storage.bucket.Bucket`
+        :param bucket: Bucket to which the notification is bound.
+
+        :rtype: :class:`BucketNotification`
+        :returns: the new notification instance
+        :raises ValueError:
+            if resource is missing 'topic' key, or if it is not formatted
+            per documented spec.
```
dhermes: We should help our users here and explain what our interpretation of "document spec" is.

storage/google/cloud/storage/notification.py
```diff
@@ -82,6 +90,40 @@ def __init__(self, bucket, topic_name,
...
+        """
+        topic_path = resource.get('topic')
+        if topic_path is None:
+            raise ValueError('Resource has no topic')
+
+        match = _TOPIC_REF_RE.match(topic_path)
+        if match is None:
+            raise ValueError(
+                'Resource has invalid topic: {}'.format(topic_path))
```
dhermes: You could include the regex in this error message? (I assume it has a not-so-bad repr?)

storage/google/cloud/storage/bucket.py
tseaver: 998b9f9

storage/google/cloud/storage/notification.py
tseaver: 998b9f9

storage/google/cloud/storage/notification.py
tseaver: 998b9f9 puts the URL of the spec in the message.

tseaver:

PR #3991, opened 2017-09-18
BQ: client.extract_table starts extract job

tswast:
Also, add system tests for extract_table.

I introduce a new property type so that the `_properties` dictionary can be more-or-less unmodified from the actual API resource representation.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -16,6 +16,8 @@
 
 from __future__ import absolute_import
 
+import uuid
+
 from google.api.core import page_iterator
 from google.cloud.client import ClientWithProject
 from google.cloud.bigquery._http import Connection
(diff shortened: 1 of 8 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -360,27 +362,39 @@ def copy_table(self, job_id, destination, *sources):
         """
         return CopyJob(job_id, destination, sources, client=self)
 
-    def extract_table_to_storage(self, job_id, source, *destination_uris):
+    def extract_table(
+            self, source, *destination_uris, job_config=None, job_id=None):
         """Construct a job for extracting a table into Cloud Storage files.
```
jba: s/Construct/Start/

bigquery/google/cloud/bigquery/job.py
```diff
@@ -969,62 +974,127 @@ def from_api_repr(cls, resource, client):
...
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.extract.printHeader
+    """
+
+    def to_api_repr(self):
+        """Build an API representation of the extact job config.
+
+        :rtype: dict
+        :returns: A dictionary in the format used by the BigQuery API.
+        """
+        resource = copy.deepcopy(self._properties)
```
jba: No need to assign to a local variable, is there?

tswast: Removed.

bigquery/google/cloud/bigquery/client.py

tswast:
Just rebased. Full `nox` sessions pass locally. Can use a second pass at a review.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -385,27 +387,49 @@ def copy_table(self, job_id, destination, *sources):
...
+              (Optional) Extra configuration options for the extract job.
+            * *job_id* (``str``) --
+              Additional content
+              (Optional) The ID of the job.
 
         :rtype: :class:`google.cloud.bigquery.job.ExtractJob`
         :returns: a new ``ExtractJob`` instance
         """
-        return ExtractJob(job_id, source, destination_uris, client=self)
+        job_config = None
```
jba: Just curious, couldn't this be written as
`job_config = kwargs.get('job_config')`
?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -973,62 +978,126 @@ def from_api_repr(cls, resource, client):
...
+        config._properties = copy.deepcopy(resource)
+        return config
 
 
 class ExtractJob(_AsyncJob):
     """Asynchronous job: extract data from a table into Cloud Storage.
 
     :type job_id: str
-    :param job_id: the job's ID, within the project belonging to ``client``.
+    :param job_id: the job's ID
```
jba: (Really a comment on the whole class.) 

Users shouldn't be creating this directly; they should always call Client.extract_table, right? Then why do we expose it? I believe @jonparrott said somewhere recently that we shouldn't return instances of private classes to users, but isn't that the right thing here?

bigquery/tests/unit/test_client.py
```diff
@@ -714,26 +715,105 @@ def test_copy_table(self):
...
         self.assertEqual(job.source, source)
         self.assertEqual(list(job.destination_uris), [DESTINATION])
 
+    def test_extract_table_generated_job_id(self):
+        from google.cloud.bigquery.job import ExtractJob
+        from google.cloud.bigquery.job import ExtractJobConfig
+        from google.cloud.bigquery.job import DestinationFormat
+
+        PROJECT = 'PROJECT'
+        JOB = 'job_name'
```
jba: Probably should use *_id instead of *_name, at least for new code.

bigquery/google/cloud/bigquery/job.py
tswast: If this was Java, I'd say we could have a public class but make the constructor private. Not sure how to express that in Python.

bigquery/google/cloud/bigquery/client.py
tswast: Yes, that's equivalent. Updating.

PR #3996, opened 2017-09-18
bigquery: remove dataset.exists

jba:
Dataset won't be able to support this method when we remove its client.  

Don't add `Client.dataset_exists`; the user can use `Client.get_dataset` and catch NotFound.

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -17,7 +17,6 @@ import six
 
 from google.api.core import page_iterator
 from google.cloud._helpers import _datetime_from_microseconds
-from google.cloud.exceptions import NotFound
 from google.cloud.bigquery.table import Table
 from google.cloud.bigquery.table import TableReference
 
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -486,30 +486,6 @@ class Dataset(object):
 
         return resource
 
-    def exists(self, client=None):
-        """API call:  test for the existence of the dataset via a GET request
-
-        See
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -486,30 +485,6 @@ def _build_resource(self):
 
         return resource
 
-    def exists(self, client=None):
```
dhermes: @jba Why don't you just make `client` required?

jba: The idea is to turn `Dataset` into more of a dumb object; all the RPCs will be on `Client`. Passing a client back into `Dataset` so it can do the call seems to be going in the other direction. Put another way, if we had `Dataset.exists(client)`, why wouldn't we have `Dataset.create(client)` and so on?

dhermes: SGTM

dhermes: Actually, I'm not sure it SGTM (sorry for the quick trigger).

This entire (15+ packages) `google.cloud` effort (over 3-ish years) has been pretty consistent in its approach. We have "nouns" for data and any "verbs" that operate on those nouns are methods on the corresponding class.

tswast: Danny, could you review the "BigQuery Python Client Libraries Redesign Proposal"? I've made many similar proposals throughout the client library and would appreciate your feedback.

tswast: The solution I proposed was something closer to what is being done for the newer auto-gen APIs with dumb resource objects and API calls being done from the client.

dhermes: Sure thing (don't hold up this PR on my account).

PR #3997, opened 2017-09-18
BigQuery: Updates Table constructor to use TableReference as parameter

alixhami:


bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -623,7 +623,8 @@ def table(self, name, schema=()):
         :rtype: :class:`google.cloud.bigquery.table.Table`
         :returns: a new ``Table`` instance
         """
-        return Table(name, dataset=self, schema=schema)
+        table_ref = TableReference(self, name)
+        return Table(table_ref, schema=schema, client=self._client)
```
tswast: Is `Dataset.table()` still used anywhere? Might make sense to delete it.

alixhami: It's still being used, but mostly in the tests. It could probably be deleted if you don't think it would be used in practice.

PR #4003, opened 2017-09-19
bigquery: add Client.update_dataset

jba:
Also, remove Dataset.patch and Dataset.update.


bigquery/google/cloud/bigquery/client.py
```diff
@@ -205,6 +205,50 @@ def get_dataset(self, dataset_ref):
...
+        reading a dataset with ``get_dataset``, changing its fields,
+        and then passing it ``update_dataset`` will ensure that the changes
+        will only be saved if no modifications to the dataset occurred
+        since the read.
+
+        :type dataset: :class:`google.cloud.bigquery.dataset.Dataset`
+        :param dataset: the dataset to update.
+
+        :type fields: sequence of string
+        :param fields: the fields of ``dataset`` to change.
```
tswast: We should clarify if these fields correspond to the API representation or the way they are represented here in the Python libraries.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -205,6 +205,50 @@ def get_dataset(self, dataset_ref):
...
+        :raises: ValueError for fields that cannot be updated.
+
+        """
+        if dataset.project is None:
+            dataset._project = self.project
+        path = '/projects/%s/datasets/%s' % (dataset.project,
+                                             dataset.dataset_id)
+        partial = {}
+        for f in fields:
+            if f not in Dataset._updateable_fields:
```
tswast: I might actually prefer a blacklist or no client-side check at all (let the backend handle issues).

I know that there will be fields that don't yet have support in the client libraries (alpha or we just haven't implemented them yet), and it will be nice if developers don't have to revert to full REST calls to use them.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -205,6 +205,50 @@ def get_dataset(self, dataset_ref):
...
+                raise ValueError('Cannot update Dataset field %s' % f)
+            partial[Dataset._updateable_fields[f]] = getattr(dataset, f)
+        if dataset.etag is not None:
+            headers = {'If-Match': dataset.etag}
+        else:
+            headers = None
+        api_response = self._connection.api_request(
+            method='PATCH', path=path, data=partial, headers=headers)
+        ds = Dataset(dataset.dataset_id, project=dataset.project, client=self)
+        ds._set_properties(api_response)
```
tswast: Probably should be a separate PR, but we may wish to provide a `Dataset.from_api_repr()` method and have the Dataset properties map directly to the underlying API resource as I did for the ExtractJobConfig https://github.com/GoogleCloudPlatform/google-cloud-python/blob/3284a0d38c10a4da9d4a34d7be2ecfd2f67ab7a9/bigquery/google/cloud/bigquery/job.py#L1035

(This would also mostly be for the use case of setting fields that the client library doesn't yet support.)

bigquery/tests/system.py
```diff
@@ -132,35 +132,22 @@ def test_get_dataset(self):
...
-            self.assertEqual(found.entity_type, expected.entity_type)
-            self.assertEqual(found.entity_id, expected.entity_id)
+        dataset.friendly_name = 'Friendly'
+        dataset.description = 'Description'
+        ds2 = Config.CLIENT.update_dataset(dataset,
+                                           ['friendly_name', 'description'])
+        self.assertEqual(ds2.friendly_name, 'Friendly')
+        self.assertEqual(ds2.description, 'Description')
+
+    # TODO(jba): test that read-modify-write with ETag works.
```
tswast: :+1:

bigquery/google/cloud/bigquery/client.py
jba: Okay, but I still need a dict from our client spelling (snake case) to the backend spelling (camel case), right? What should I do if something is not in the dict?

Or should I just transform from snake to camel?

bigquery/google/cloud/bigquery/client.py

bigquery/google/cloud/bigquery/client.py
jba: There is a `from_api_repr` method.  I missed it.

Done. I'll leave the direct mapping for another time.

bigquery/google/cloud/bigquery/client.py
tswast: I think we can safely assume that the field names will always be camel case based on https://developers.google.com/protocol-buffers/docs/proto3#json "Message field names are mapped to lowerCamelCase and become JSON object keys."

We might eventually need a dictionary for backend names that are reserved keywords in Python, but I guess we'll address that when we get to it.

jba: Mapped snake to camel case.

PR #4004, opened 2017-09-19
BigQuery: Adds client.get_table() and removes table.reload()

alixhami:


bigquery/google/cloud/bigquery/client.py
```diff
@@ -207,6 +208,20 @@ def get_dataset(self, dataset_ref):
...
+        :type table_ref:
+            :class:`google.cloud.bigquery.table.TableReference`
+        :param table_ref: the table to use.
+
+        :rtype: :class:`google.cloud.bigquery.table.Table`
+        :returns: a ``Table`` instance
+        """
+        api_response = self._connection.api_request(
+            method='GET', path=table_ref.path)
+        return Table.from_api_repr(api_response, self)
```
jba: Maybe out of scope, but why are the `from_api_repr` methods public?

bigquery/tests/unit/test_client.py
```diff
@@ -383,6 +383,35 @@ def test_create_dataset_w_attrs(self):
...
+        resource = {
+            'id': '%s:%s:%s' % (project, dataset_id, table_id),
+            'tableReference': {
+                'projectId': project,
+                'datasetId': dataset_id,
+                'tableId': table_id
+            },
+        }
+        conn = client._connection = _Connection(resource)
+        dataset_ref = client.dataset(dataset_id)
```
jba: dataset_ref = client.dataset(...).table(...)
?

bigquery/tests/system.py
```diff
@@ -402,8 +402,7 @@ def test_load_table_from_local_avro_file_then_dump_table(self):
```
jba: Looks like there are no system tests that specifically target get_table? Maybe worth writing one to confirm that we really get all the fields.

bigquery/google/cloud/bigquery/client.py
tswast: Mostly because it was a convention that already existed in this library. I'm not sure why some classes had it and some don't.

It's relatively common for API representations be serialized to JSON. I know for certain the `bq` command-line tool takes "JSON files" as arguments for things like schemas, which map directly to the API representation, so it's not unheard of that someone would have something they want to parse.

The other thing I might like to use these methods for is to allow setting properties that the client libraries don't yet support (as they are introduced in alpha releases or just have trouble keeping up since this is a hand-written library)

bigquery/tests/system.py
tswast: :+1: We should have at least one system test for each client method. Then again, any such system test will likely have to load a table first.

@lukesneeringer Do you know if  there's a script somewhere to setup test resources that we should update, if we wanted to have system tests for pure GET methods?

bigquery/tests/unit/test_client.py
tswast: I think you mean `table_ref = client.dataset(...).table(...)`, but yeah +1 to that since it looks like dataset_ref is not used for anything else.

bigquery/tests/system.py
jba: I believe the system tests should be self-contained. Ideally each run will create a new uniquely named dataset and tables within it, so multiple runs can occur in parallel even within the same project.

Maybe you're uncomfortable that the get_table test depends on create_table, and the create_table test depends on get_table? Actually, maybe those can be the same test. Anyway, the circularity doesn't really bother me.

tswast: In that case, maybe one of these load tests just needs a new name (and maybe some additional asserts) to explicitly call out that it is testing both load and get?

alixhami: I added a system test using a public dataset

bigquery/tests/unit/test_client.py
alixhami: I removed the unnecessary variable

bigquery/tests/system.py
jba: @alixhami Clever!

PR #4013, opened 2017-09-20
bigquery: add client.list_dataset_tables

jba:
Remove dataset.list_tables


jba:
The base of this PR is #4012.

PR #4014, opened 2017-09-20
BigQuery: updates dataset.table() to return a TableReference instead of a Table

alixhami:


bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -141,6 +141,9 @@ def path(self):
     def table(self, table_id):
         """Constructs a TableReference.
 
+        :type table_id: str
```
tswast: Good catch. Thanks!

PR #4016, opened 2017-09-21
Retry streaming exceptions (this time for sure, Rocky!)

### LINKED ISSUE #3775: Cloud spanner client does not seem to be retrying broken streams

tseaver:
Supersedes #3930.

Closes #3775.

/cc vkedia

### DIFF
--- a/spanner/tests/unit/test_snapshot.py
+++ b/spanner/tests/unit/test_snapshot.py
@@ -226,7 +226,7 @@ class Test_SnapshotBase(unittest.TestCase):
 
         result_set = derived.read(
             TABLE_NAME, COLUMNS, KEYSET,
-            index=INDEX, limit=LIMIT, resume_token=TOKEN)
+            index=INDEX, limit=LIMIT)
 
         self.assertEqual(derived._read_request_count, count + 1)
 
@@ -236,6 +236,7 @@ class Test_SnapshotBase(unittest.TestCase):
             self.assertIsNone(result_set._source)
 
         result_set.consume_all()
+
         self.assertEqual(list(result_set.rows), VALUES)
         self.assertEqual(result_set.metadata, metadata_pb)
         self.assertEqual(result_set.stats, stats_pb)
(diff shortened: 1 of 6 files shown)


tseaver:
@vkedia, @bjwatson PTAL

lukesneeringer:
After @vkedia says yes, I will start the process for a Spanner release.

bjwatson:
Thanks @lukesneeringer, @tseaver, and @jonparrott!

spanner/tests/unit/test_snapshot.py
```diff
@@ -31,6 +33,68 @@
...
+            self._make_item(0),
+            self._make_item(1, resume_token='DEADBEEF'),
+        )
+        SECOND = (  # discarded after 503
+            self._make_item(2),
+        )
+        LAST = (
+            self._make_item(3),
+        )
+        ITEMS = FIRST + SECOND + LAST
```
vkedia: unused

spanner/tests/unit/test_snapshot.py
```diff
@@ -31,6 +33,68 @@
...
+            self._make_item(2),
+            self._make_item(3),
+        )
+        raw = _MockIterator(*ITEMS)
+        restart = mock.Mock(spec=[], return_value=raw)
+        resumable = self._call_fut(restart)
+        self.assertEqual(list(resumable), list(ITEMS))
+        restart.assert_called_once_with()
+
+    def test_iteration_w_raw_raising_unavailable(self):
```
vkedia: can  you also add a test case where it raises unavailable just after getting a resume token.

spanner/tests/unit/test_snapshot.py
tseaver: fa7bf9e

spanner/tests/unit/test_snapshot.py
tseaver: fa7bf9e

lukesneeringer:
Is this approved, @vkedia?

PR #4020, opened 2017-09-21
BigQuery: add to/from API representation for Table & Dataset references.

tswast:
Also, implement equality and hashing for Table & Dataset references.
This will make it easier to use the TableReference and DatasetReference
classes as typed properties in the QueryJob and other job classes.

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -151,6 +151,46 @@ class DatasetReference(object):
         """
         return TableReference(self, table_id)
 
+    @classmethod
+    def from_api_repr(cls, resource):
+        project = resource['projectId']
+        dataset_id = resource['datasetId']
+        return cls(project, dataset_id)
+
+    def to_api_repr(self):
+        return {
+            'projectId': self._project,
+            'datasetId': self._dataset_id,
+        }
+
+    def _key(self):
+        """A tuple key that unique-ly describes this field.
+
+        Used to compute this instance's hashcode and evaluate equality.
+
+        Returns:
+            tuple: The contents of this :class:`DatasetReference`.
+        """
+        return (
+            self._project,
+            self._dataset_id,
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, DatasetReference):
+            return NotImplemented
+        return self._key() == other._key()
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __hash__(self):
+        return hash(self._key())
+
+    def __repr__(self):
+        return 'DatasetReference{}'.format(self._key())
+
+
 
 class Dataset(object):
     """Datasets are containers for tables.
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -191,7 +191,6 @@ class DatasetReference(object):
         return 'DatasetReference{}'.format(self._key())
 
 
-
 class Dataset(object):
     """Datasets are containers for tables.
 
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -147,6 +147,45 @@ def table(self, table_id):
...
+        return cls(project, dataset_id)
+
+    def to_api_repr(self):
+        return {
+            'projectId': self._project,
+            'datasetId': self._dataset_id,
+        }
+
+    def _key(self):
+        """A tuple key that unique-ly describes this field.
```
jba: "uniquely"

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -147,6 +147,45 @@ def table(self, table_id):
...
+            tuple: The contents of this :class:`DatasetReference`.
+        """
+        return (
+            self._project,
+            self._dataset_id,
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, DatasetReference):
+            return NotImplemented
```
jba: s/return/raise/?

jba: And I would say it's TypeError or something like that. But maybe this is a convention?

bigquery/tests/unit/test_table.py
```diff
@@ -58,6 +58,98 @@ def test_ctor_defaults(self):
...
+    def test___eq___wrong_type(self):
+        from google.cloud.bigquery.dataset import DatasetReference
+        dataset_ref = DatasetReference('project_1', 'dataset_1')
+        table = self._make_one(dataset_ref, 'table_1')
+        other = object()
+        self.assertNotEqual(table, other)
+        self.assertEqual(table, mock.ANY)
+
+    def test___eq___project_mismatch(self):
+        from google.cloud.bigquery.dataset import DatasetReference
```
jba: BTW, is there a reason we don't import at the top of the file?

bigquery/google/cloud/bigquery/dataset.py
tswast: Yeah, should be raise or maybe just return False actually. Copy-pasted from SchemaField, so I'll update that too.

bigquery/tests/unit/test_table.py
tswast: Yeah, because we've got some circular references going on, which breaks in Python 2.7.

bigquery/google/cloud/bigquery/dataset.py
tseaver: `return NotImplemented` is [idiomatic Python](https://docs.python.org/2/library/constants.html#NotImplemented) for a binary operator implementation:  it allows the "right-hand-side" version to be invoked, in case the `other` class *does* know how to do the comparison.

bigquery/tests/unit/test_table.py
tseaver: FWIW, https://palladion.com/home/tseaver/obzervationz/2008/unit_testing_notes-20080724 and https://palladion.com/home/tseaver/obzervationz/2009/unit_testing_redux-20090802 have some additional justifications for avoiding imports (particularly of the module/package being tested) at module scope in unit tests.  The point is debated, of course.

bigquery/google/cloud/bigquery/dataset.py
tswast: Done.

bigquery/tests/unit/test_table.py
tswast: Thanks for the background!

Issue #3761, opened 2017-08-08
Storage flaky 429 in system tests


dhermes:
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/2762

lukesneeringer:
I think I am going to leave this one undone. My rationale:

  * It has only happened once, and likely was due to the spurt of system test runs on Tuesday.
  * Properly working around a 429 would be tough. I was originally thinking I would just make it skip the test, but there is the potential that the 429 happens on the cleanup steps (as it did here).

If this recurs, we should reopen and revisit.

tseaver:
[Happened again today](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3430).

I think the fix is to have `tearDown` wrap an entire helper function with `retry_429` (since `retry_429` doesn't make sense inside the batch context manager's suite).

PR #4021, opened 2017-09-21
Add system tests for bucket notifications.

tseaver:
N.B.:  the tests currently fail with a 403:

```python
____________ TestStorageNotificationCRUD.test_notification_explicit ____________
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/tests/system.py", line 628, in test_notification_explicit
    retry_429(notification.create)()
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/.nox/sys-3-6/lib/python3.6/site-packages/test_utils/retry.py", line 95, in wrapped_function
    return to_wrap(*args, **kwargs)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/google/cloud/storage/notification.py", line 243, in create
    data=properties,
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/_http.py", line 293, in api_request
    raise exceptions.from_http_response(response)
google.api.core.exceptions.Forbidden: 403 POST https://www.googleapis.com/storage/v1/b/notification-explicit-1506029348265/notificationConfigs: The service account 'precise-truck-742@gs-project-accounts.iam.gserviceaccount.com' does not have permission to publish messages to to the Cloud Pub/Sub topic '//pubsub.googleapis.com/projects/precise-truck-742/topics/notification-1506029282321', or that topic does not exist.
____________ TestStorageNotificationCRUD.test_notification_minimal _____________
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/tests/system.py", line 607, in test_notification_minimal
    retry_429(notification.create)()
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/.nox/sys-3-6/lib/python3.6/site-packages/test_utils/retry.py", line 95, in wrapped_function
    return to_wrap(*args, **kwargs)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/google/cloud/storage/notification.py", line 243, in create
    data=properties,
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/_http.py", line 293, in api_request
    raise exceptions.from_http_response(response)
google.api.core.exceptions.Forbidden: 403 POST https://www.googleapis.com/storage/v1/b/notification-minimal-1506029351126/notificationConfigs: The service account 'precise-truck-742@gs-project-accounts.iam.gserviceaccount.com' does not have permission to publish messages to to the Cloud Pub/Sub topic '//pubsub.googleapis.com/projects/precise-truck-742/topics/notification-1506029282321', or that topic does not exist.
```

See ~~#3956~~ https://github.com/GoogleCloudPlatform/google-cloud-common/issues/231 where I ask for help on that issue.

### DIFF
--- a/storage/nox.py
+++ b/storage/nox.py
@@ -71,10 +71,11 @@ def system_tests(session, python_version):
     # virutalenv's dist-packages.
     session.install('mock', 'pytest', *LOCAL_DEPS)
     session.install('../test_utils/')
-    session.install('.')
+    session.install('../pubsub')
+    session.install('-e', '.')
 
     # Run py.test against the system tests.
-    session.run('py.test', '--quiet', 'tests/system.py')
+    session.run('py.test', '--quiet', 'tests/system.py', *session.posargs)
 
 
 @nox.session
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -607,7 +607,7 @@ class TestStorageNotificationCRUD(unittest.TestCase):
         retry_429(notification.create)()
         try:
             self.assertTrue(notification.exists())
-            self.assertIsNotNone(notification.id)
+            self.assertIsNotNone(notification.notification_id)
             notifications = list(bucket.list_notifications())
             self.assertEqual(len(notifications), 1)
             self.assertEqual(notifications[0].topic_name, self.TOPIC_NAME)
@@ -628,7 +628,7 @@ class TestStorageNotificationCRUD(unittest.TestCase):
         retry_429(notification.create)()
         try:
             self.assertTrue(notification.exists())
-            self.assertIsNotNone(notification.id)
+            self.assertIsNotNone(notification.notification_id)
             self.assertEqual(
                 notification.custom_attributes, self.CUSTOM_ATTRIBUTES)
             self.assertEqual(notification.event_types, self.event_types())
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -567,14 +567,26 @@ class TestStorageNotificationCRUD(unittest.TestCase):
         return 'projects/{}/topics/{}'.format(
             Config.CLIENT.project, self.TOPIC_NAME)
 
-    def setUp(self):
-        self.case_buckets_to_delete = []
+    def _intialize_topic(self):
         try:
             from google.cloud.pubsub_v1 import PublisherClient
         except ImportError:
             raise unittest.SkipTest("Cannot import pubsub")
         self.publisher_client = PublisherClient()
         retry_429(self.publisher_client.create_topic)(self.topic_path)
+        policy = self.publisher_client.get_iam_policy(self.topic_path)
+        binding = policy.bindings.add()
+        binding.role = 'roles/pubsub.publisher'
+        binding.members.append(
+            'serviceAccount:{}'
+            '@gs-project-accounts.iam.gserviceaccount.com'.format(
+                Config.CLIENT.project))
+        self.publisher_client.set_iam_policy(self.topic_path, policy)
+
+
+    def setUp(self):
+        self.case_buckets_to_delete = []
+        self._intialize_topic()
 
     def tearDown(self):
         retry_429(self.publisher_client.delete_topic)(self.topic_path)
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -550,3 +550,91 @@ class TestStorageRewrite(TestStorageFiles):
         self.assertEqual(total, len(source_data))
 
         self.assertEqual(dest.download_as_string(), source_data)
+
+
+class TestStorageNotificationCRUD(unittest.TestCase):
+
+    topic = None
+    TOPIC_NAME = 'notification' + unique_resource_id('-')
+    CUSTOM_ATTRIBUTES = {
+        'attr1': 'value1',
+        'attr2': 'value2',
+    }
+    BLOB_NAME_PREFIX = 'blob-name-prefix/'
+
+    @property
+    def topic_path(self):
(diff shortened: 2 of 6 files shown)


storage/nox.py
```diff
@@ -71,10 +71,11 @@ def system_tests(session, python_version):
     # virutalenv's dist-packages.
     session.install('mock', 'pytest', *LOCAL_DEPS)
     session.install('../test_utils/')
-    session.install('.')
+    session.install('../pubsub')
+    session.install('-e', '.')
```
dhermes: @tseaver I seem to recall @lukesneeringer had a reason for not making system tests editable.

storage/tests/system.py
```diff
@@ -550,3 +550,91 @@ def test_rewrite_rotate_encryption_key(self):
...
+            raise unittest.SkipTest("Cannot import pubsub")
+        self.publisher_client = PublisherClient()
+        retry_429(self.publisher_client.create_topic)(self.topic_path)
+
+    def tearDown(self):
+        retry_429(self.publisher_client.delete_topic)(self.topic_path)
+        with Config.CLIENT.batch():
+            for bucket_name in self.case_buckets_to_delete:
+                bucket = Config.CLIENT.bucket(bucket_name)
+                retry_429(bucket.delete)()
```
dhermes: Does this retry make sense within the context of a batch?

storage/tests/system.py
```diff
@@ -550,3 +550,91 @@ def test_rewrite_rotate_encryption_key(self):
...
+        notification = bucket.notification(self.TOPIC_NAME)
+        retry_429(notification.create)()
+        try:
+            self.assertTrue(notification.exists())
+            self.assertIsNotNone(notification.id)
+            notifications = list(bucket.list_notifications())
+            self.assertEqual(len(notifications), 1)
+            self.assertEqual(notifications[0].topic_name, self.TOPIC_NAME)
+        finally:
+            notification.delete()
```
dhermes: This `finally` seems icky in the test case, why not register it for a `tearDown` cleanup?

storage/tests/system.py
```diff
@@ -550,3 +550,91 @@ def test_rewrite_rotate_encryption_key(self):
...
+        self.case_buckets_to_delete.append(new_bucket_name)
+        self.assertEqual(list(bucket.list_notifications()), [])
+        notification = bucket.notification(self.TOPIC_NAME)
+        retry_429(notification.create)()
+        try:
+            self.assertTrue(notification.exists())
+            self.assertIsNotNone(notification.id)
+            notifications = list(bucket.list_notifications())
+            self.assertEqual(len(notifications), 1)
+            self.assertEqual(notifications[0].topic_name, self.TOPIC_NAME)
```
dhermes: Anything else from the notification worth checking?

storage/nox.py
tseaver: Having the environments use the checked-out files (rather than the installed ones) is more convenient for debugging, I think, but it isn't essential to me.  I'll back it out if @lukesneeringer says so (I'd like to know why, of course).

storage/tests/system.py
tseaver: Not for the minimal case.  Like `id`, the server sets `etag` and `selfLink` values, but we can't predict their contents.  The explicit case which follows asserts that the values we set in the constructor are preserved when loading the resource returned from `notifications/insert`.

storage/tests/system.py
tseaver: These two methods are really just exercising the CR~~U~~D for notifications, which makes having the `delete` in the body a bit clearer.

tseaver:
7d9ee94 updates the notification system tests, adding the IAM grant to the topic required to allow creation of the notification.  Those tests are now passing (61c5d0f and da6dc1c fix bugs uncovered by running them).

storage/tests/system.py
tseaver: Even better:  the system test [failed at exactly this point today with a 429](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3430), because the `Batch.finish` can't be wrapped with a `retry_429` when invoked via a context manager.

tseaver: #3761

tseaver:
[Branch build with new system tests](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3433) passes the storage system tests.

tseaver:
I rebased the target branch against `master` after the merge of #4050, and rebased this PR's branch against that branch:  here's hoping the "dandruff shampoo" worked.

storage/tests/system.py
```diff
@@ -550,3 +550,103 @@ def test_rewrite_rotate_encryption_key(self):
...
+        try:
+            from google.cloud.pubsub_v1 import PublisherClient
+        except ImportError:
+            raise unittest.SkipTest("Cannot import pubsub")
+        self.publisher_client = PublisherClient()
+        retry_429(self.publisher_client.create_topic)(self.topic_path)
+        policy = self.publisher_client.get_iam_policy(self.topic_path)
+        binding = policy.bindings.add()
+        binding.role = 'roles/pubsub.publisher'
+        binding.members.append(
```
BrandonY: It bothers me that all this needs to be done explicitly to the the client. Will end users need to repeat all this logic?

Could the client provide a one step "subscribe(bucket, topic)" method that adds the appropriate IAM permissions? It would make things much more usable.

PR #4027, opened 2017-09-22
Bump release to 0.27.1.

tseaver:
Draft release notes:

## google-cloud-spanner 0.27.1

- Pin dependency  `google-auth >= 1.1.0` to fix use of `google.cloud.spanner` on GCE. (PR #3950, issue #3933)
- Fix merge of partial timestamp and date field values. (PR #4015, issue #3981, #3998)
- Restart streaming iterators from `read` / `execute_sql` operations after resumable error (PR #4016, issue #3775).

---

Internal housekeeping not included in release notes:

- Suppress instance creation tests by default on CI (PR #3951)

Issue #3998, opened 2017-09-19
Google cloud spanner client fails when reading large columns


paulMissault:
Consider following table:

```
CREATE TABLE big_arrays (
    key INT64 NOT NULL,
    a_big_array ARRAY<TIMESTAMP> NOT NULL,
) PRIMARY KEY (key)
```

Which gets populated by 100 rows each containing a large array of 1000 timestamps.

```
from google.cloud import spanner
import datetime
import random
spanner_client = spanner.Client(project='my-project')
instance = spanner_client.instance('my-instance')
db = instance.database('my-db')

def random_timestamp():
    return datetime.datetime.utcfromtimestamp(random.uniform(1e9, 2e9))
def random_big_array(size=100):
    return [random_timestamp() for i in range(1000)]

batch_size = 10
batches = 10
for batch_index in range(batches):
    with db.batch() as batch:
        batch.insert_or_update(
            table='big_arrays',
            columns=['key', 'a_big_array'],
            values = [(key, random_big_array()) for key in range(batch_index*batch_size, (batch_index+1)*batch_size)]
        )
```
Trying to fetch these rows with `db.execute_sql('select * from big_arrays limit 100').consume_all()` results in following error and stacktrace:

```
Traceback (most recent call last):
  File "Untitled5.py", line 60, in <module>
    db.execute_sql("SELECT * FROM big_arrays limit 100").consume_all()
  File "/Users/myname/Repos/transformations/venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 159, in consume_all
    self.consume_next()
  File "/Users/myname/Repos/transformations/venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 148, in consume_next
    values[0] = self._merge_chunk(values[0])
  File "/Users/myname/Repos/transformations/venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 108, in _merge_chunk
    merged = _merge_by_type(self._pending_chunk, value, field.type)
  File "/Users/myname/Repos/transformations/venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 272, in _merge_by_type
    return merger(lhs, rhs, type_)
  File "/Users/myname/Repos/transformations/venv/lib/python2.7/site-packages/google/cloud/spanner/streamed.py", line 233, in _merge_array
    merged = _merge_by_type(last, first, element_type)
[...]

Issue #3981, opened 2017-09-15
Spanner should know how to merge chunk data of TIMESTAMP and DATE type


yixizhang:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/google/cloud/spanner/streamed.py#L300

The dict doesn't cover `type_pb2.TIMESTAMP` and `type_pb2.DATE`, the fix is actually easy because both type is passed in string_value.

vkedia:
Thanks for filing this. This indeed looks like a bug.
@tseaver @lukesneeringer please take a look.

PR #3951, opened 2017-09-13
Suppress instance creation tests by default on CI

tseaver:
We don't have enough quota to risk orphans.

- Fix `tests/system/utils/populate_streaming` script after PR #3787.

- Add utility for scrubbing orphaned instances.

PR #4028, opened 2017-09-22
Spanner Beta

lukesneeringer:
This shall be the "root" PR for getting Spanner to beta. Multiple PRs are going to be made using it as a base.

The remaining to-do items are:

  * [x] Re-generate the autogen libraries for the data and admin APIs.
      * [x] Spanner Admin Database (#4029)
      * [x] Spanner Admin Instance (#4030)
      * [x] Regular ol' Spanner, yo (#4033)
  * [x] ~Update documentation to include autogen API reference.~ (superceded by #4033)
  * [x] Bit flip `setup.py` and appropriate documentation to claim beta status. (#4034)
  * [x] ~Pin gRPC to < 1.6~ (superceded by #4029)


tseaver:
From #4027, which this PR supersedes:

---

Draft release notes:

## google-cloud-spanner 0.27.1

- Pin dependency  `google-auth >= 1.1.0` to fix use of `google.cloud.spanner` on GCE. (PR #3950, issue #3933)
- Fix merge of partial timestamp and date field values. (PR #4015, issue #3981, #3998)
- Restart streaming iterators from `read` / `execute_sql` operations after resumable error (PR #4016, issue #3775).

---

Internal housekeeping not included in release notes:

- Suppress instance creation tests by default on CI (PR #3951)

bjwatson:
@lukesneeringer @jonparrott We also want a quick review of the API Reference docs to make sure it's reasonably in sync with the client library.

jonparrott:
@lukesneeringer is this ready for me to review the docs?

lukesneeringer:
@jonparrott Yes.

bjwatson:
Thanks @lukesneeringer, @tseaver, and @jonparrott.

jonparrott:
Non-documentation issues/concerns

1. `google.cloud.spanner` is *not* just an alias to `google.cloud.spanner_v1`. This is inconsistent with the approach we took with `pubsub`.

Documentation issues (these may be deferred by moving them into their own bugs):

1. The landing page is... not useful:

screen shot 2017-09-22 at 2 06 58 pm

1. http://temp.theadora.io/google-cloud-spanner/spanner/instance-usage.html#instance-configurations says that `list_instance_configs` and `list_instances` returns a token, but these return iterators.
2. http://temp.theadora.io/google-cloud-spanner/spanner/instance-usage.html#check-on-current-instance-operation references the old operation interface, there is no `finished()` in the new one. It should just reference `result()`.
3. http://temp.theadora.io/google-cloud-spanner/spanner/database-usage.html#create-a-new-database Spanner does not have its own Operation class.
4. http://temp.theadora.io/google-cloud-spanner/spanner/database-usage.html#check-on-current-database-operation similar to (2)
5. http://temp.theadora.io/google-cloud-spanner/spanner/database-usage.html#configuring-a-session-pool-for-a-database the sample here has a typo (and seems to be formatted strangely)
6. The sidebar for spanner is totally whack.  The low-level gapic stuff is mixed in with the high-level wrappers, and one page's sub-titles are treated as separate pages.
7. The page titles mostly end with "API" which is totally weird and overloaded in this context.

bjwatson:
Thanks for your detailed review @jonparrott!

@lukesneeringer is the "Non-documentation issues/concern" a breaking change to fix later. I'm assuming consistency is important here; what do you think?

@tseaver It seems like documentation items 2-6 (shown as 1-5 due to the graphic) should be fixed before the Beta release. The other presentation and organizational concerns can be fixed before GA. Do you agree @jonparrott and @lukesneeringer?

FYI @vkedia

jonparrott:
#6 should really be fixed ASAP.

bjwatson:
@jonparrott I see what you're saying with that. The [Python API Reference](https://googlecloudplatform.github.io/google-cloud-python/latest/spanner/usage.html) combines manual and GAPIC at the same level, whereas [the Ruby docs](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-spanner/v0.23.2/google/cloud/spanner) (and those for other languages) puts it another level below under `v1`.

@lukesneeringer Is this particularly difficult to fix with the current docs setup? I don't think we want to create a lot of work to finalize docs for Beta, but I do see the usability concern that Jon is raising.

lukesneeringer:
Putting in PRs for these.

lukesneeringer:
Okay, I fixed all of these except the `spanner` alias (the item without a PR number in the comment above is waiting on #4054 to be approved, because it needs it as a baseline).

@jonparrott, I am having trouble deciding what to do about the `spanner_v1` thing. I actually considered moving it, and decided against it because doing so would break people in a few cases (e.g. anyone who typed `from google.cloud.spanner.client import Client`, which the documentation previously recommended in one spot, now corrected).

With Pub/Sub, we were breaking everyone _anyway_, so it was an obvious decision. Additionally, since Pub/Sub was a thin client, it was quite necessary to do this.

I am willing to do it here also, but due to the potential customer randomization, it is less clear to me that it is the correct path. Please advise. :-)

jonparrott:
I think we need to make a firm decision on the api -> api_{version} going forwards - either it's a direct alias or it adds functionality. Either spanner or pubsub will need to change. I strongly think we should follow what pubsub did - spanner is transitioning into beta so breaking is allowed (although less than ideal, sure).

jonparrott:
fwiw, the pubsub samples use `google.cloud.pubsub_v1` explicitly.

lukesneeringer:
Okay, I am happy to change it then. :-)

bjwatson:
Thanks, @lukesneeringer for fixing just about everything. I agree that I'm happier with all the `API` suffixes in a post-[4054](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4054) world, too.

What's the PR for the alias fix?

lukesneeringer:
> What's the PR for the alias fix?

There is not one yet. I had wall-to-wall meetings followed by lunch. :-)

bjwatson:
Understood. Must be Monday. Lots of meetings.

lukesneeringer:
@jonparrott @bjwatson Okay, #4064 exists.

This is probably a breaking change at the margins. In other words, most users will not be affected, but some might be. In particular, we did have at least one doc that told users to `from google.cloud.spanner.client import Client`, and that will break.

Do we need any kind of migration guide?

bjwatson:
@lukesneeringer We should at least have release notes that explain this breaking change.

lukesneeringer:
@bjwatson We will have release notes either way. :-)

bjwatson:
Woo hoo! Great job @lukesneeringer, @tseaver, and @jonparrott!

PR #4029, opened 2017-09-22
Move the admin database GAPIC to source delivery.

lukesneeringer:
Uses #4028 as a base.

Commit-by-commit recommended as the review path.


spanner/google/cloud/spanner_admin_database_v1/types.py
```diff
@@ -0,0 +1,52 @@
...
+        policy_pb2,
+        audit_data_pb2,
+        operations_pb2,
+        any_pb2,
+        descriptor_pb2,
+        empty_pb2,
+        timestamp_pb2,
+        status_pb2, ):
+    for name, message in get_messages(module).items():
+        message.__module__ = 'google.cloud.spanner_admin_database_v1.types'
```
tseaver: I don't think we ever resolved my concern about hiding the real source package for the proto-generated message types.

spanner/setup.py
```diff
@@ -51,12 +51,14 @@
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.27.0, < 0.28dev',
-    'google-auth >= 1.1.0',
-    'grpcio >= 1.2.0, < 2.0dev',
     'gapic-google-cloud-spanner-v1 >= 0.15.0, < 0.16dev',
-    'gapic-google-cloud-spanner-admin-database-v1 >= 0.15.0, < 0.16dev',
     'gapic-google-cloud-spanner-admin-instance-v1 >= 0.15.0, < 0.16dev',
```
tseaver: Do I presume there is a similar PR coming for the `admin-instance` GAPIC / proto stuff?

spanner/setup.py
```diff
@@ -51,12 +51,14 @@
 
 
 REQUIREMENTS = [
-    'google-cloud-core >= 0.27.0, < 0.28dev',
-    'google-auth >= 1.1.0',
-    'grpcio >= 1.2.0, < 2.0dev',
```
tseaver: Are we allowing our dependencies to pick the right `grpcio`?

spanner/google/cloud/spanner_admin_database_v1/types.py
lukesneeringer: That is right, we did not.

The goal here is that I want the idiom for users to be `from google.cloud.api_vN import types`, and then we promise that all of the types that they need exist there. I feel like it is a pain for users to have to care about `random_modulename_pb2`, which I believe is surprising (also, if you are a user, what the heck does `_pb2` even mean?).

This approach means that the documentation documents _all_ of these messages under `types`, so that is all a user has to care about, unless he goes and spelunks the code (at which point the import chain is explicit).

spanner/setup.py
lukesneeringer: You assume correctly. Doing them one at a time for review sanity.

spanner/setup.py
lukesneeringer: Yep, `google-gax` (eventually `google-api-core`) will be the point of truth on the correct gRPC. This is increasingly important in a world where things like grpc/grpc#12455 happen.

spanner/google/cloud/spanner_admin_database_v1/types.py
tseaver: My memory is that we were not publishing API docs for the `__init__` module, but this PR does add them for the package and for `types`, which ameliorates the problem.

Have we examined the generated docs for completeness / clarity in the context of the move?

lukesneeringer: They are effectively just API reference and types. No narrative docs. :-(
I would welcome narrative docs but do not have time to write them, and they are not a requirement.

PR #4030, opened 2017-09-22
Move the Spanner Admin Instance GAPIC.

lukesneeringer:
Uses #4028 as a base.

Commit-by-commit recommended as a review path.


spanner/google/cloud/spanner_admin_instance_v1/gapic/instance_admin_client.py
```diff
@@ -437,14 +437,10 @@ def list_instances(self,
                   * NAME:howl --> Equivalent to above.
                   * labels.env:* --> The instance has the label \"env\".
                   * labels.env:dev --> The instance has the label \"env\" and the value of
-                ::
-
-                                       the label contains the string \"dev\".
+                    the label contains the string \"dev\".
```
tseaver: Are there issues open on the tooling project to correct these errors?

lukesneeringer: No, not yet, because I have not determined where the right place is to look. The fault could be with the proto comment, or it could be with pandoc.

lukesneeringer: Just checked the proto; it's pandoc.

lukesneeringer: Nope, I lied, it's a toolkit problem. Toolkit is not using pandoc.

lukesneeringer: https://github.com/googleapis/toolkit/issues/1587

PR #4038, opened 2017-09-22
BigQuery: replaces table.create() with client.create_table()

alixhami:


tswast:
Any idea what the circle failure is? Logs aren't loading for me.

tswast:
LGTM once we figure out Circle.

jba:
The failure is from cover:
```
google/cloud/bigquery/table.py                404      0    171      1    99%   589->592
```
BTW, I find that if I close the wedge icon next to the failing item on the Circle page, then re-open it, the logs load immediately.

tswast:
Re: wedge. Good trick!

Looks like we need to add a "view with legacy SQL" property to a unit test to make coverage happy.

PR #4041, opened 2017-09-22
Add google.api.core.helpers.grpc_helpers

jonparrott:


jonparrott:
Towards #3842

jonparrott:
@dhermes any comments?

jonparrott:
Also, a note on this implementation:

It is intentionally simplistic for now. I've made it possible so that in the future we could return subclasses of the gRPC ABCs (`grpc.UnaryUnaryMultiCallable`, etc.) to handle wrapping errors for *all* of the ways of invoking the rpc (`with_call` and `future` specifically). I don't think this is required for now, but I wanted to make sure we could do it in the future without breaking anything.

core/google/api/core/helpers/grpc_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+# The list of gRPC Callable interfaces that return iterators.
+_STREAM_WRAP_CLASSES = (
+    grpc.UnaryStreamMultiCallable,
+    grpc.StreamStreamMultiCallable,
+)
+
+
+def _patch_callable_name(callable):
```
dhermes: `callable` is in `__builtins__` (see how the code highlighting makes it blue below), please don't shadow it. Suggest `callable_`

jonparrott: Done.

PR #4051, opened 2017-09-23
bigquery: modify CopyJob

jba:
Update CopyJob and CopyJobConfig to conform to the new design for jobs.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import collections
 import uuid
 
 from google.api.core import page_iterator
@@ -492,25 +493,39 @@ class Client(ClientWithProject):
         """
         return LoadJob(job_id, destination, source_uris, client=self)
 
-    def copy_table(self, job_id, destination, *sources):
-        """Construct a job for copying one or more tables into another table.
+    def copy_table(self, sources, destination, job_id=None, job_config=None):
+        """Start a job for copying one or more tables into another table.
 
         See
         https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.copy
 
-        :type job_id: str
-        :param job_id: Name of the job.
+        :type sources: One of:
+                       :class:`~google.cloud.bigquery.table.TableReference`
+                       sequence of
+                       :class:`~google.cloud.bigquery.table.TableReference`
+        :param sources: Table or tables to be copied.
 
-        :type destination: :class:`google.cloud.bigquery.table.Table`
+
+        :type destination: :class:`google.cloud.bigquery.table.TableReference`
         :param destination: Table into which data is to be copied.
 
-        :type sources: sequence of :class:`google.cloud.bigquery.table.Table`
-        :param sources: tables to be copied.
+        :type job_id: str
+        :param job_id: (Optional) The ID of the job.
+
+        :type job_config: :class:`google.cloud.bigquery.job.CopyJobConfig`
+        :param job_config: (Optional) Extra configuration options for the job.
 
         :rtype: :class:`google.cloud.bigquery.job.CopyJob`
         :returns: a new ``CopyJob`` instance
         """
-        return CopyJob(job_id, destination, sources, client=self)
+        job_id = _make_job_id(job_id)
+
+        if not isinstance(sources, collections.Sequence):
+            sources = [sources]
+        job = CopyJob(job_id, sources, destination, client=self,
+                      job_config=job_config)
+        job.begin()
+        return job
 
     def extract_table(self, source, *destination_uris, **kwargs):
         """Start a job to extract a table into Cloud Storage files.
@@ -541,9 +556,7 @@ class Client(ClientWithProject):
         :returns: a new ``ExtractJob`` instance
         """
         job_config = kwargs.get('job_config')
-        job_id = kwargs.get('job_id')
-        if job_id is None:
-            job_id = str(uuid.uuid4())
+        job_id = _make_job_id(kwargs.get('job_id'))
 
         job = ExtractJob(
             job_id, source, list(destination_uris), client=self,
@@ -667,3 +680,17 @@ def _item_to_table(iterator, resource):
     :returns: The next table in the page.
     """
     return Table.from_api_repr(resource, iterator.client)
+
+
+def _make_job_id(job_id):
+    """Construct an ID for a new job.
+
+    :type job_id: str or ``NoneType``
+    :param job_id: the user-provided job ID
+
+    :rtype: str
+    :returns: A job ID
+    """
+    if job_id is None:
+        return str(uuid.uuid4())
+    return job_id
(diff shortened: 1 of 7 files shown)


jba:
Added a system test for copy_table, which fails. I confirmed that it also fails from the command-line, and emailed BQ team about it.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -492,25 +492,45 @@ def load_table_from_storage(self, job_id, destination, *source_uris):
...
+              (Optional) The ID of the job.
 
         :rtype: :class:`google.cloud.bigquery.job.CopyJob`
         :returns: a new ``CopyJob`` instance
         """
-        return CopyJob(job_id, destination, sources, client=self)
+        job_config = kwargs.get('job_config')
+        job_id = _make_job_id(kwargs.get('job_id'))
+
+        if isinstance(sources, TableReference):
```
lukesneeringer: I might be more inclined to do this one the converse way:

```python
if not isinstance(sources, collections.Sequence):
    sources = [sources]
```

My logic is that if at some future point we accept some additional input (say, for example, a dictionary that maps nicely to a `TableReference`, this logic will continue to work.

That may not matter here. I would care a lot if this was mucking with an autogen `_pb2` type, but it is not doing that.

So, essentially, dropping the thought here, but I trust your judgment.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -853,56 +855,98 @@ def from_api_repr(cls, resource, client):
...
+        :returns: Configuration parsed from ``resource``.
+        """
+        config = cls()
+        config._properties = copy.deepcopy(resource)
+        return config
 
 
 class CopyJob(_AsyncJob):
-    """Asynchronous job: copy data into a table from other tables.
+    """Asynchr_onous job: copy data into a table from other tables.
```
lukesneeringer: I assume this is a typo. :-p

bigquery/google/cloud/bigquery/job.py
```diff
@@ -913,31 +957,27 @@ def _build_resource(self):
             'tableId': table.table_id,
         } for table in self.sources]
 
-        resource = {
+        configuration = self._configuration.to_api_repr()
+        configuration['sourceTables'] = source_refs
+        configuration['destinationTable'] = {
```
lukesneeringer: Sanity check: Is there a case where `configuration['destinationTable']` would already be set? (Not saying there is.)

bigquery/tests/unit/test_client.py
```diff
@@ -1113,19 +1113,57 @@ def test_copy_table(self):
...
         destination = dataset.table(DESTINATION)
-        job = client.copy_table(JOB, destination, source)
+
+        job = client.copy_table(source, destination, job_id=JOB)
+
+        # Check that copy_table actually starts the job.
+        self.assertEqual(len(conn._requested), 1)
+        req = conn._requested[0]
+        self.assertEqual(req['method'], 'POST')
+        self.assertEqual(req['path'], '/projects/PROJECT/jobs')
```
lukesneeringer: This seems correct by "accident". It chould be `'/projects/%s/jobs' % PROJECT`, I believe.

bigquery/google/cloud/bigquery/client.py
jba: tbh, I wasn't aware how to test for being a sequence in modern Python. I'll use your code.

bigquery/google/cloud/bigquery/job.py
jba: Done.

bigquery/google/cloud/bigquery/job.py
jba: There certainly is:
```py
job = client.get_job("a-job-id")
```
Will set both job.destination and job._configuration['destinationTable'].  Is that a concern?

bigquery/tests/unit/test_client.py
jba: Good catch. Fixed.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -492,25 +493,45 @@ def load_table_from_storage(self, job_id, destination, *source_uris):
         """
         return LoadJob(job_id, destination, source_uris, client=self)
 
-    def copy_table(self, job_id, destination, *sources):
-        """Construct a job for copying one or more tables into another table.
+    def copy_table(self, sources, destination, **kwargs):
```
tswast: You can use normal keyword arguments in this method.

The only reason I couldn't for extract table is that the destination there is a `*args` (varargs) and you have to do `**kwargs` for keyword arguments after a `*args` in Python 2.7.

tswast: I wonder if I should change `extract_table` to just have the second argument be a single item or a list as you've done here? I kept with what was there, but this way of doing it makes more sense to me.

bigquery/google/cloud/bigquery/job.py
tswast: The current behavior is the one I'd expect as a user since `destination_table` is not defined as a property of the `CopyJobConfig` class. The value set on the `CopyJob` should take precedence.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1033,7 +1069,7 @@ def from_api_repr(cls, resource):
             An extract job configuration in the same representation as is
             returned from the API.
 
-        :rtype: :class:`google.cloud.bigquery.job.ExtractJobConfig`
+        :rtype: :class:`google.cloud.bigquery.job.CopyJobConfig`
```
tswast: Looks like this method belongs to ExtractJobConfig. Probably meant to copy this to the CopyJobConfig class? We may wish to make this part of a shared base class.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1243,7 +1279,8 @@ def __init__(self, job_id, query, client,
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.allowLargeResults
     """
 
-    create_disposition = CreateDisposition('create_disposition')
+    create_disposition = CreateDisposition('create_disposition',
```
tswast: Does this mean the QueryJob needs a `self._properties = {}` in the constructor now?

bigquery/google/cloud/bigquery/client.py
jba: Coming from Go, I like as little magic as possible, so I'd go with having the single-item-or-list.

bigquery/google/cloud/bigquery/job.py
jba: Fixed.

bigquery/google/cloud/bigquery/job.py
jba: I thought we were doing the `ApiResourceProperty` thing for all the job config classes? I was copying `ExtractJobConfig`. I don't have an opinion about this&mdash;we can do whatever makes the most sense to you.

tswast: We are, but I saw this is on QueryJob, which hasn't been fully converted yet. If the system tests seem okay with it, I think it's fine to leave as you have it until we convert QueryJob + QueryJobConfig, too.

PR #4057, opened 2017-09-25
Add google.api.core.gapic_v1.method

jonparrott:
Towards #3842.

### DIFF
--- a/core/google/api/core/gapic_v1/method.py
+++ b/core/google/api/core/gapic_v1/method.py
@@ -22,7 +22,6 @@ import platform
 
 import pkg_resources
 
-from google.api.core import retry
 from google.api.core import timeout
 from google.api.core.helpers import grpc_helpers
 
@@ -42,9 +41,9 @@ def _apply_decorators(func, decorators):
     ``decorators`` may contain items that are ``None`` or ``False`` which will
     be ignored.
     """
-    decorators = list(filter(_is_not_none_or_false, decorators))
+    decorators = filter(_is_not_none_or_false, reversed(decorators))
 
-    for decorator in reversed(decorators):
+    for decorator in decorators:
         func = decorator(func)
 
     return func
@@ -76,23 +75,23 @@ def _prepare_metadata(metadata):
 
 
 class _GapicCallable(object):
-    """Callable that applies retry, timeout, and metadata logic."""
+    """Callable that applies retry, timeout, and metadata logic.
+
+    Args:
+        target (Callable): The low-level RPC method. If timeout is not
+            False, should accept a timeout argument. If metadata is not
+            False, it should accept a metadata argument.
+        retry (google.api.core.retry.Retry): The default retry for the
+            callable. If False, this callable will not retry.
+        timeout (google.api.core.timeout.Timeout): The default timeout
+            for the callable. If ``False``, this callable will not specify
+            a timeout argument to the low-level RPC method.
+        metadata (Sequence[Tuple[str, str]]): gRPC call metadata that's
+            passed to the low-level RPC method. If ``False``, this callable
+            will not specify a metadata argument to the method.
+    """
 
     def __init__(self, target, retry, timeout, metadata):
-        """
-        Args:
-            target (Callable): The low-level RPC method. If timeout is not
-                False, should accept a timeout argument. If metadata is not
-                False, it should accept a metadata argument.
-            retry (google.api.core.retry.Retry): The default retry for the
-                callable. If False, this callable will not retry.
-            timeout (google.api.core.timeout.Timeout): The default timeout
-                for the callable. If False, this callable will not specify
-                a timeout argument to the low-level RPC method.
-            metadata (Sequence[Tuple[str, str]]): gRPC call metadata that's
-                passed to the low-level RPC method. If False, this callable
-                will not specify a metadata argument to the method.
-        """
         self._target = target
         self._retry = retry
         self._timeout = timeout
@@ -120,8 +119,8 @@ class _GapicCallable(object):
         # matches the retry's. This handles the case where the user leaves
         # the timeout default but specifies a lower deadline via the retry.
         if (timeout_ is self._timeout
-                and retry is not False
-                and retry is not self._retry
+                and retry_ is not False
+                and retry_ is not self._retry
                 and isinstance(timeout_, timeout.ExponentialTimeout)):
             timeout_ = timeout_.with_deadline(retry_._deadline)
 
--- a/core/google/api/core/gapic_v1/method.py
+++ b/core/google/api/core/gapic_v1/method.py
@@ -66,12 +66,74 @@ def _prepare_metadata(metadata):
         _API_CORE_VERSION, _PY_VERSION, _API_CORE_VERSION)
 
     # Merge this with any existing metric metadata.
-    client_metadata = ' '.join(filter(None, [
-        client_metadata, metadata.get(METRICS_METADATA_KEY, None)]))
+    if METRICS_METADATA_KEY in metadata:
+        client_metadata = '{} {}'.format(
+            client_metadata, metadata[METRICS_METADATA_KEY])
 
     metadata[METRICS_METADATA_KEY] = client_metadata
 
     return list(metadata.items())
+
+
+class _GapicCallable(object):
+    """Callable that applies retry, timeout, and metadata logic."""
+
+    def __init__(self, target, retry, timeout, metadata):
+        """
+        Args:
+            target (Callable): The low-level RPC method. If timeout is not
+                False, should accept a timeout argument. If metadata is not
+                False, it should accept a metadata argument.
+            retry (google.api.core.retry.Retry): The default retry for the
+                callable. If False, this callable will not retry.
+            timeout (google.api.core.timeout.Timeout): The default timeout
+                for the callable. If False, this callable will not specify
+                a timeout argument to the low-level RPC method.
+            metadata (Sequence[Tuple[str, str]]): gRPC call metadata that's
+                passed to the low-level RPC method. If False, this callable
+                will not specify a metadata argument to the method.
+        """
+        self._target = target
+        self._retry = retry
+        self._timeout = timeout
+        self._metadata = metadata
+
+    def __call__(self, *args, **kwargs):
+        """Invoke the low-level RPC with retry, timeout, and metadata."""
+        # Note: Due to Python 2 lacking keyword-only arguments we use this
+        # pattern to extract the retry and timeout params.
+        retry_ = kwargs.pop('retry', None)
+        if retry_ is None:
+            retry_ = self._retry
+
+        timeout_ = kwargs.pop('timeout', None)
+        if timeout_ is None:
+            timeout_ = self._timeout
+
+        # If timeout is specified as a number instead of a Timeout instance,
+        # convert it to a ConstantTimeout.
+        if timeout_ is not False and isinstance(timeout_, (int, float)):
+            timeout_ = timeout.ConstantTimeout(timeout_)
+
+        # If timeout is the default and the default timeout is exponential and
+        # a non-default retry is specified, make sure the timeout's deadline
+        # matches the retry's. This handles the case where the user leaves
+        # the timeout default but specifies a lower deadline via the retry.
+        if (timeout_ is self._timeout
+                and retry is not False
+                and retry is not self._retry
+                and isinstance(timeout_, timeout.ExponentialTimeout)):
+            timeout_ = timeout_.with_deadline(retry_._deadline)
+
+        # Apply all applicable decorators.
+        wrapped_func = _apply_decorators(self._target, [retry_, timeout_])
+
+        # Set the metadata for the call using the metadata calculated by
+        # _prepare_metadata.
+        if self._metadata is not False:
+            kwargs['metadata'] = self._metadata
+
+        return wrapped_func(*args, **kwargs)
 
 
 def wrap_method(func, default_retry=None, default_timeout=None, metadata=None):
@@ -134,8 +196,9 @@ def wrap_method(func, default_retry=None, default_timeout=None, metadata=None):
     equivalent to just calling ``get_topic`` but with error re-mapping.
 
     Args:
-        func (Callable[Any]): The function to wrap. If timeout is specified,
-            then this function should take an optional timeout parameter.
+        func (Callable[Any]): The function to wrap. If timeout is not
+            False, should accept a timeout argument. If metadata is not False,
+            it should accept a metadata argument.
         default_retry (Optional[google.api.core.Retry]): The default retry
             strategy.
         default_timeout (Optional[google.api.core.Timeout]): The default
@@ -147,47 +210,13 @@ def wrap_method(func, default_retry=None, default_timeout=None, metadata=None):
             at all.
 
     Returns:
-        TODO
+        Callable: A new callable that takes optional ``retry`` and ``timeout``
+            arguments and applies the common error mapping, retry, timeout,
+            and metadata behavior to the low-level RPC method.
     """
     func = grpc_helpers.wrap_errors(func)
+
     if metadata is not False:
         metadata = _prepare_metadata(metadata)
 
-    def method(*args, **kwargs):
-        """RPC method with retry and timeout."""
-        # Note: Due to Python 2 lacking keyword-only arguments we use this
-        # pattern to extract the retry and timeout params.
-        retry_ = kwargs.pop('retry', None)
-        if retry_ is None:
-            retry_ = default_retry
-
-        timeout_ = kwargs.pop('timeout', None)
-        if timeout_ is None:
-            timeout_ = default_timeout
-
-        # If timeout is specified as a number instead of a Timeout instance,
-        # convert it to a ConstantTimeout.
-        if timeout_ is not False and isinstance(timeout_, (int, float)):
-            timeout_ = timeout.ConstantTimeout(timeout_)
-
-        # If timeout is the default and the default timeout is exponential and
-        # a non-default retry is specified, make sure the timeout's deadline
-        # matches the retry's. This handles the case where the user leaves
-        # the timeout default but specifies a lower deadline via the retry.
-        if (timeout_ is default_timeout
-                and retry is not False
-                and retry is not default_retry
-                and isinstance(timeout_, timeout.ExponentialTimeout)):
-            timeout_ = timeout_.with_deadline(retry_._deadline)
-
-        # Apply all applicable decorators.
-        wrapped_func = _apply_decorators(func, [retry_, timeout_])
-
-        # Set the metadata for the call using the metadata calculated by
-        # _prepare_metadata.
-        if metadata is not False:
-            kwargs['metadata'] = metadata
-
-        return wrapped_func(*args, **kwargs)
-
-    return method
+    return _GapicCallable(func, default_retry, default_timeout, metadata)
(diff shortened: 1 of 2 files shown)


core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,193 @@
...
+    return value is not None and value is not False
+
+
+def _apply_decorators(func, decorators):
+    """Apply a list of decorators to a given function.
+
+    ``decorators`` may contain items that are ``None`` or ``False`` which will
+    be ignored.
+    """
+    decorators = list(filter(_is_not_none_or_false, decorators))
```
dhermes: Why cast to a list before passing to reversed? Why not pass a `reversed` iterator on `decorators` to `filter`?

```python
decorators = filter(_is_not_none_or_false, reversed(decorators))
```

FWIW I doubt it matters since it'll be a very small number of values in `decorators`. However the "late wrapping" means this will be called at run time (vs. import time), so perf **might** matter.

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    if metadata is None:
+        metadata = {}
+
+    client_metadata = 'api-core/{} gl-python/{} grpc/{}'.format(
+        _API_CORE_VERSION, _PY_VERSION, _API_CORE_VERSION)
+
+    # Merge this with any existing metric metadata.
+    client_metadata = ' '.join(filter(None, [
+        client_metadata, metadata.get(METRICS_METADATA_KEY, None)]))
```
dhermes: What does `filter(None, ...)` do?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,193 @@
...
+        default_timeout (Optional[google.api.core.Timeout]): The default
+            timeout strategy. If an int or float is specified, the timeout will
+            always be set to that value.
+        metadata (Union(Mapping[str, str], False)): A dict of metadata keys and
+            values. This will be augmented with common ``x-google-api-client``
+            metadata. If False, metadata will not be passed to the function
+            at all.
+
+    Returns:
+        TODO
```
dhermes: Forgot this?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    func = grpc_helpers.wrap_errors(func)
+    if metadata is not False:
```
dhermes: Why are we identity-checking for `False`?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,193 @@
...
+        metadata = _prepare_metadata(metadata)
+
+    def method(*args, **kwargs):
```
dhermes: This closure has **a lot** going on. I'd be much more comfortable with a `__call__`-able object that owns the wrapped values `func`, `default_retry`, `default_timeout`, and `metadata`.

core/google/api/core/gapic_v1/method.py
jonparrott: Without list: `TypeError: 'filter' object is not reversible`.

core/google/api/core/gapic_v1/method.py
jonparrott: The key might not exist in the metadata, so you could end up with:

```' '.join([client_metadata, None])``` which would raise a TypeError.

So the filter just drops the None from the list so you get:

```
' '.join([client_metadata])
```

Which is just equivalent to `client_metadata`.

If you want, I can re-arrange this as:

```python
if METRICS_METADATA_KEY in metadata:
    client_metadata = '{} {}'.format(client_metadata, metadata[METRICS_METADATA_KEY])
```

Actually - I'm gonna do that anyway.

core/google/api/core/gapic_v1/method.py
jonparrott: Yep. Fixed.

core/google/api/core/gapic_v1/method.py
jonparrott: False signals not to specify metadata at all.

core/google/api/core/gapic_v1/method.py
jonparrott: Done.

core/google/api/core/gapic_v1/method.py
dhermes: @jonparrott That does not apply to my suggestion (but it does explain why the current impl. requires `list`).

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,222 @@
...
+
+    return list(metadata.items())
+
+
+class _GapicCallable(object):
+    """Callable that applies retry, timeout, and metadata logic."""
+
+    def __init__(self, target, retry, timeout, metadata):
+        """
+        Args:
```
dhermes: It has been our "tradition" to document `Args` in the class docstring and leave `__init__` without a docstring

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,222 @@
...
+            target (Callable): The low-level RPC method. If timeout is not
+                False, should accept a timeout argument. If metadata is not
+                False, it should accept a metadata argument.
+            retry (google.api.core.retry.Retry): The default retry for the
+                callable. If False, this callable will not retry.
```
dhermes: Use

```
:data:`False`
```

to make it more explicit? Or even more explicit prose? Or should we use our own sentinel value?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,222 @@
...
+        self._target = target
+        self._retry = retry
+        self._timeout = timeout
+        self._metadata = metadata
+
+    def __call__(self, *args, **kwargs):
+        """Invoke the low-level RPC with retry, timeout, and metadata."""
+        # Note: Due to Python 2 lacking keyword-only arguments we use this
+        # pattern to extract the retry and timeout params.
+        retry_ = kwargs.pop('retry', None)
```
dhermes: Trivial question irrelevant to code review: Why the trailing underscore on `retry_` (and `timeout_`)?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,222 @@
...
+        # If timeout is specified as a number instead of a Timeout instance,
+        # convert it to a ConstantTimeout.
+        if timeout_ is not False and isinstance(timeout_, (int, float)):
+            timeout_ = timeout.ConstantTimeout(timeout_)
+
+        # If timeout is the default and the default timeout is exponential and
+        # a non-default retry is specified, make sure the timeout's deadline
+        # matches the retry's. This handles the case where the user leaves
+        # the timeout default but specifies a lower deadline via the retry.
+        if (timeout_ is self._timeout
```
dhermes: I'd be more comfortable with a flag like `original_timeout` or something here rather than the identity check. Maybe I'm being paranoid?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,222 @@
...
+                and retry is not False
```
dhermes: Did you mean `retry` or `retry_`?

core/google/api/core/gapic_v1/method.py
jonparrott: Ah - reading comprehensive is hard. Done.

core/google/api/core/gapic_v1/method.py
jonparrott: Done.

core/google/api/core/gapic_v1/method.py
jonparrott: Done. I don't think we need to use our own sentinel value.

core/google/api/core/gapic_v1/method.py
jonparrott: Because I import `retry` and `timeout`.

core/google/api/core/gapic_v1/method.py
jonparrott: I meant `retry_`, good catch.

jonparrott:
> Why do we need to use False as a sentinel?

At invocation time it indicates the difference between `default` and `don't`:

```python
# Use default retry and timeout
wrapped_method()
# Don't retry
wrapped_method(retry=False)
```

> It seems the __call__ method could benefit from a helper that figures out the right timeout for the caller.

Done.

jonparrott:
> Why do we need to use False as a sentinel?

I suppose I could change

```python
        retry = kwargs.pop('retry', None)
        if retry is None:
            retry = self._retry
```

to

```python
    retry = kwargs.pop('retry', self._retry)
```

which means that `wrapped_method(retry=False)` and `wrapped_method(retry=None)` will be semantically equivalent.

Actually that makes sense, I'm gonna do that.

dhermes:
> At invocation time it indicates the difference between `default` and `don't`

I realize that's the use, I'm saying using `False` as the "don't" sentinel is confusing / problematic.

jonparrott:
``metadata`` is going to keep the ``False`` sentinel, but I'll remove it from retry and timeout.

core/google/api/core/gapic_v1/method.py
```diff
@@ -82,9 +116,9 @@ class _GapicCallable(object):
             False, should accept a timeout argument. If metadata is not
             False, it should accept a metadata argument.
         retry (google.api.core.retry.Retry): The default retry for the
-            callable. If False, this callable will not retry.
+            callable. If falsy, this callable will not retry.
```
dhermes: This is less true than before, it's an explicit `is False` check, not a `if not retry:` check

core/google/api/core/gapic_v1/method.py
```diff
@@ -99,33 +133,22 @@ def __init__(self, target, retry, timeout, metadata):
...
-        # If timeout is the default and the default timeout is exponential and
-        # a non-default retry is specified, make sure the timeout's deadline
-        # matches the retry's. This handles the case where the user leaves
-        # the timeout default but specifies a lower deadline via the retry.
-        if (timeout_ is self._timeout
-                and retry_ is not False
-                and retry_ is not self._retry
-                and isinstance(timeout_, timeout.ExponentialTimeout)):
-            timeout_ = timeout_.with_deadline(retry_._deadline)
+        # Note: Due to Python 2 lacking keyword-only arguments we kwargs to
```
dhermes: "we use kwargs"?

core/google/api/core/gapic_v1/method.py
```diff
@@ -99,33 +133,22 @@ def __init__(self, target, retry, timeout, metadata):
...
+        # to extract the retry and timeout params.
+        timeout_ = _determine_timeout(
+            self._timeout,
+            kwargs.pop('timeout', None),
+            # Use the invocation-specified retry only for this, as we only
+            # want to adjust the timeout deadline if the *user* specified
+            # a different retry.
+            kwargs.get('retry', None))
+
+        retry = kwargs.pop('retry', None)
```
dhermes: 1. Why not pop before setting `timeout_` and then just pass the popped value into `_determine_timeout`?
1. Can we avoid the `timeout` vs. `timeout_` issue just by using `google.api.core.timeout` or `core.timeout`

core/google/api/core/gapic_v1/method.py
```diff
@@ -99,33 +133,22 @@ def __init__(self, target, retry, timeout, metadata):
...
-            timeout_ = timeout_.with_deadline(retry_._deadline)
+        # Note: Due to Python 2 lacking keyword-only arguments we kwargs to
+        # to extract the retry and timeout params.
+        timeout_ = _determine_timeout(
+            self._timeout,
+            kwargs.pop('timeout', None),
+            # Use the invocation-specified retry only for this, as we only
+            # want to adjust the timeout deadline if the *user* specified
+            # a different retry.
+            kwargs.get('retry', None))
```
dhermes: Do you not want to fallback to `self._retry` (just wanted to make sure)

core/google/api/core/gapic_v1/method.py
jonparrott: Should be fixed

core/google/api/core/gapic_v1/method.py
jonparrott: Fixed.

core/google/api/core/gapic_v1/method.py
jonparrott: Yep - hence the comment.

core/google/api/core/gapic_v1/method.py
```diff
@@ -116,13 +114,13 @@ class _GapicCallable(object):
...
-            for the callable. If falsy, this callable will not specify
-            a timeout argument to the low-level RPC method.
-        metadata (Sequence[Tuple[str, str]]): gRPC call metadata that's
-            passed to the low-level RPC method. If ``False``, this callable
-            will not specify a metadata argument to the method.
+            for the callable. If ``None``, this callable will not specify
+            a timeout argument to the low-level RPC method by default.
+        metadata (Optional[Sequence[Tuple[str, str]]]): gRPC call metadata
+            that's passed to the low-level RPC method. If falsy, no metadata
+            will be passed to the low-level RPC method.
```
dhermes: "falsy" or explicitly `None`?

core/google/api/core/gapic_v1/method.py
```diff
@@ -133,32 +131,32 @@ def __init__(self, target, retry, timeout, metadata):
 
     def __call__(self, *args, **kwargs):
         """Invoke the low-level RPC with retry, timeout, and metadata."""
-        # Note: Due to Python 2 lacking keyword-only arguments we kwargs to
+        # Note: Due to Python 2 lacking keyword-only arguments we use kwargs to
```
dhermes: Sorry I missed the "to to"

core/google/api/core/gapic_v1/method.py
```diff
@@ -133,32 +131,32 @@ def __init__(self, target, retry, timeout, metadata):
...
-            retry = self._retry
+        retry = kwargs.pop('retry', self._retry)
 
         # Apply all applicable decorators.
         wrapped_func = _apply_decorators(self._target, [retry, timeout_])
 
         # Set the metadata for the call using the metadata calculated by
         # _prepare_metadata.
-        if self._metadata is not False:
+        if self._metadata:
```
dhermes: ISTM that explicit `None` is a better idea here. What if someone uses `metadata = {}` but they want you to send it?

core/google/api/core/gapic_v1/method.py
```diff
@@ -222,14 +220,15 @@ def get_topic(name, timeout=None):
...
-            always be set to that value.
-        metadata (Union(Mapping[str, str], False)): A dict of metadata keys and
+            timeout strategy. Can also be specified as an int or float. If
+            ``None``, the method will not have timeout specified by default.
+        metadata (Optional(Mapping[str, str])): A dict of metadata keys and
             values. This will be augmented with common ``x-google-api-client``
-            metadata. If ``False``, metadata will not be passed to the function
-            at all.
+            metadata. If ``None``, metadata will not be passed to the function
+            at all, if ``USE_DEFAULT_METADATA`` (the default) then only the
```
dhermes: You **could** use a Sphinx `attr` to refer to the sentinel?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _is_not_none_or_false(value):
+    return value is not None and value is not False
+
+
+def _apply_decorators(func, decorators):
+    """Apply a list of decorators to a given function.
+
+    ``decorators`` may contain items that are ``None`` or ``False`` which will
```
dhermes: Do we care about the `None` or `False` bit anymore? Seems like just `None` now?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,244 @@
...
+            arguments and applies the common error mapping, retry, timeout,
+            and metadata behavior to the low-level RPC method.
+    """
+    func = grpc_helpers.wrap_errors(func)
+
+    if metadata is USE_DEFAULT_METADATA:
+        metadata = {}
+
+    if metadata is not None:
+        metadata = _prepare_metadata(metadata)
```
dhermes: This means that the `None`-check in `_prepare_metadata` is now redundant.

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Returns:
+        Optional[Timeout]: The timeout to apply to the method or ``None``.
+    """
+    if specified_timeout is default_timeout:
+        # If timeout is the default and the default timeout is exponential and
+        # a non-default retry is specified, make sure the timeout's deadline
+        # matches the retry's. This handles the case where the user leaves
+        # the timeout default but specifies a lower deadline via the retry.
+        if retry and isinstance(default_timeout, timeout.ExponentialTimeout):
```
dhermes: Do you want truthy `retry` or just explicitly not `None`?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,244 @@
...
+            return default_timeout.with_deadline(retry._deadline)
```
dhermes: ISTM that `retry._deadline` might want to consider becoming public?

core/google/api/core/gapic_v1/method.py
```diff
@@ -0,0 +1,244 @@
...
+            metadata. If ``None``, metadata will not be passed to the function
+            at all, if ``USE_DEFAULT_METADATA`` (the default) then only the
+            common metadata will be provided.
+
+    Returns:
+        Callable: A new callable that takes optional ``retry`` and ``timeout``
+            arguments and applies the common error mapping, retry, timeout,
+            and metadata behavior to the low-level RPC method.
+    """
+    func = grpc_helpers.wrap_errors(func)
```
dhermes: Not really important, but you **could** just punt on this until the `_GapicCallable` constructor.

core/google/api/core/gapic_v1/method.py
jonparrott: Let's go with None. Fixed.

core/google/api/core/gapic_v1/method.py
jonparrott: Fixed.

core/google/api/core/gapic_v1/method.py
jonparrott: Done.

core/google/api/core/gapic_v1/method.py
jonparrott: Better safe than sorry.

core/google/api/core/gapic_v1/method.py
jonparrott: Done.

core/google/api/core/gapic_v1/method.py
jonparrott: Truthy (it could technically still be `False`).

core/google/api/core/gapic_v1/method.py
jonparrott: Possibly but I don't want to cross that bridge quite yet.

core/google/api/core/gapic_v1/method.py
jonparrott: I'd rather not - I'd prefer the _GapicCallable only deal with late-wrapping logic. (This might be useful later)

core/google/api/core/gapic_v1/method.py
jonparrott: Good catch.

PR #4060, opened 2017-09-25
Improve the Spanner landing page.

lukesneeringer:


docs/conf.py
```diff
@@ -122,7 +122,6 @@
     'github_user': 'GoogleCloudPlatform',
     'github_repo': 'google-cloud-python',
     'github_banner': True,
-    'travis_button': True,
```
dhermes: 1. Why?
1. This is certainly unrelated to Spanner

docs/spanner/usage.rst
```diff
@@ -1,11 +1,21 @@
 Spanner
-=======
+^^^^^^^
+
+`Cloud Spanner`_ is the world’s first fully managed relational database service
```
dhermes: You have a non-ascii single quote here, which I assume is an accident

docs/spanner/usage.rst
```diff
@@ -14,19 +24,77 @@ Spanner
   advanced-session-pool-topics
   api-reference
 
-API requests are sent to the `Cloud Spanner`_ API via RPC over
-HTTP/2.  In order to support this, we'll rely on `gRPC`_.
+Spanner Client
```
dhermes: Do I need to scrutinize this section or is it copy-pasta from the usage doc?

docs/conf.py
lukesneeringer: 1. Because we do not use Travis, at all, so the button is always pegged to our last-ever Travis build.
  2. Yes, it is. The world will go on. :-)

docs/spanner/usage.rst
lukesneeringer: Yep, I copy and pasted that from cloud site.

docs/spanner/usage.rst
lukesneeringer: The latter. :-)

docs/conf.py
dhermes: I agree with "The world will go on", but not "Yes, it is". (Good call on "we do not use Travis". MERP.)

PR #4065, opened 2017-09-26
BigQuery: add client.query_rows(), remove client.run_sync_query().

tswast:
The query_rows() method will be the new way to run a query
synchronously. It starts a query job, then waits for the results,
returning the rows as results.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -29,6 +29,7 @@ from google.cloud.bigquery.job import CopyJob
 from google.cloud.bigquery.job import ExtractJob
 from google.cloud.bigquery.job import LoadJob
 from google.cloud.bigquery.job import QueryJob
+from google.cloud.bigquery.job import QueryJobConfig
 from google.cloud.bigquery.query import QueryResults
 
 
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1218,11 +1218,53 @@ def from_api_repr(cls, resource, client):
...
+    """Configuration options for query jobs.
 
-    Values which are ``None`` -> server defaults.
+    All properties in this class are optional. Values which are ``None`` ->
+    server defaults.
     """
+
+    def __init__(self):
+        self._properties = {}
+        # TODO(swast): Default to standard SQL syntax.
```
jba: You do this in Client.query_rows.  I don't think there's anything to do here.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1218,11 +1218,53 @@ def from_api_repr(cls, resource, client):
...
+    """See
+    https://cloud.google.com/bigquery/docs/\
+    reference/v2/jobs#configuration.query.useLegacySql
+    """
+
+    dry_run = _TypedApiResourceProperty('dry_run', 'dryRun', bool)
+    """See
+    https://g.co/cloud/bigquery/docs/reference/v2/jobs#configuration.dryRun
+    """
+
```
jba: Shouldn't all the fields below become `ApiResourceProperty`s?

bigquery/tests/system.py
```diff
@@ -1082,14 +1090,16 @@ def test_sync_query_w_query_params(self):
...
+            query_job = Config.CLIENT.run_async_query(
+                'test_query_w_query_params{}'.format(unique_resource_id()),
                 example['sql'],
                 query_parameters=example['query_parameters'])
-            query.use_legacy_sql = False
-            query.run()
-            self.assertEqual(len(query.rows), 1)
-            self.assertEqual(len(query.rows[0]), 1)
-            self.assertEqual(query.rows[0][0], example['expected'])
+            query_job.use_legacy_sql = False
```
jba: Shouldn't be necessary: using query params should force this internally.

bigquery/tests/unit/test_client.py
```diff
@@ -1320,58 +1321,167 @@ def test_run_async_w_query_parameters(self):
...
-
-    def test_run_sync_query_w_udf_resources(self):
-        from google.cloud.bigquery._helpers import UDFResource
-        from google.cloud.bigquery.query import QueryResults
+        conn = client._connection = _Connection(
+            RESOURCE, RESULTS_RESOURCE, FIRST_PAGE, LAST_PAGE)
 
-        RESOURCE_URI = 'gs://some-bucket/js/lib.js'
+        rows_iter = client.query_rows(QUERY)
+        rows = [row for row in rows_iter]
```
jba: `rows = list(rows_iter)`?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -595,29 +596,48 @@ def run_async_query(self, job_id, query,
...
+
+        :raises: :class:`~google.cloud.exceptions.GoogleCloudError` if the job
+            failed or  :class:`TimeoutError` if the job did not complete in the
+            given timeout.
         """
-        return QueryResults(query, client=self,
-                            udf_resources=udf_resources,
-                            query_parameters=query_parameters)
+        # TODO(swast): use job_config parameter
+        if job_id is None:
```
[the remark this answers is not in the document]

job-id-is-prefix isn't ideal since I know of some use cases where a user might want to generate a deterministic job ID for a load job to prevent duplicates. Instead, configure a prefix as an option when you create a Client. But, agreed that the logic should be in one place.

bigquery/google/cloud/bigquery/job.py
tswast: Yes, but I only did the minimum for replacing `run_sync_query`. I'll do the rest when I replace `run_async_query`. I started doing both, but that was becoming way too big a change with all the tests that needed updates.

bigquery/tests/system.py
tswast: I tried removing and got "google.api.core.exceptions.BadRequest: 400 POST https://www.googleapis.com/bigquery/v2/projects/swast-scratch/jobs: Cannot use query parameters in a legacy SQL query."

I'll investigate further when I replace `run_async_query`.

bigquery/tests/unit/test_client.py
tswast: Done.

bigquery/google/cloud/bigquery/job.py
tswast: Removed.

bigquery/google/cloud/bigquery/client.py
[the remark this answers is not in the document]
jba: Sorry, I wasn't clear. The logic will be:
- Specify job_id, we use it exactly (that handles the use case you mention).
- Omit job_id, we generate an entirely random one
- Do something else (design TBD; in Go we have the equivalent of an add_job_id_suffix bool keyword arg) to get a random string appended to what you provide.

tswast:
I see the merge system test failed, but it's due to duplicate job IDs. If it happens again, let's increase the length of the random IDs used in system tests.

jba:
Could it be because the random number generator always starts with the same seed?

tswast:
Possibly, though I'd expect we'd be seeing a lot more collisions if that were the case. But it could be the random number generator doesn't have quite as much entropy as we'd expect.

PR #4076, opened 2017-09-27
BigQuery: replaces table.update() and table.patch() with client.update_table()

alixhami:


bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,37 +619,50 @@ def _set_properties(self, api_response):
...
                 'tableId': self.table_id},
         }
-        if self.description is not None:
+        if (
+                self.description is not None and fields == [] or
+                'description' in fields):
             resource['description'] = self.description
 
-        if self.expires is not None:
+        if self.expires is not None and fields == [] or 'expires' in fields:
```
jba: What about doing what Client.update_dataset does instead? It seems simpler.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,37 +619,50 @@ def _set_properties(self, api_response):
...
+        if (
+                self.view_query is not None and fields == []
+                or 'view_query' in fields):
             view = resource['view'] = {}
             view['query'] = self.view_query
-            if self.view_use_legacy_sql is not None:
-                view['useLegacySql'] = self.view_use_legacy_sql
 
-        if self._schema:
+        if (
```
jba: We want standard SQL to be the default, so if self.view_use_legacy_sql is None and fields == [], we want to set it to False in the resource.

bigquery/google/cloud/bigquery/table.py
alixhami: I started by doing what Client.update_dataset does (converting from snake case to camel case), but the property names are not consistent due to naming and nesting. For example, expires => expirationTime, partitioning_type => timePartitioning, view_query => view['query'], etc. The table._build_resource() function is very messy though, and @tswast and I discussed how to reorganize the code to make table resource building easier. One thing that might help is creating classes for these properties that can generate their own pieces of the API resource (like Schema and View).

bigquery/google/cloud/bigquery/table.py
alixhami: I just asked @tswast about this and it sounds like we can fix this when we build a View class

bigquery/google/cloud/bigquery/table.py
jba: I think classes might be overkill. For dataset update I originally had a map from Python settable field names to API field names. Maybe you could have a map from field names to functions that set the field in a resource.

tswast: A map from field names to functions that populate a resource isn't a bad idea. It would certainly be cleaner than these if statements.

bigquery/tests/unit/test_client.py
```diff
@@ -830,7 +830,7 @@ def test_update_table_w_schema_None(self):
                 'projectId': project,
                 'datasetId': dataset_id,
                 'tableId': table_id},
-            'schema': {'fields': []}
+            'schema': []
```
tswast: This line should not have changed. If we are looking at an API response, this should either be None or there needs to be a dictionary with the `fields` key.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -619,54 +619,62 @@ def _set_properties(self, api_response):
...
+            resource['view'] = {}
+        resource['view']['useLegacySql'] = self.view_use_legacy_sql
+
+    def _populate_view_query_resource(self, resource):
+        if 'view' not in resource:
+            resource['view'] = {}
+        resource['view']['query'] = self.view_query
+
+    def _populate_schema_resource(self, resource):
+        if len(self._schema) == 0:
```
tswast: Will `_schema` never be `None`?

bigquery/tests/unit/test_client.py
```diff
@@ -634,6 +634,288 @@ def test_update_dataset(self):
...
+        }
+        schema = [
+            SchemaField('full_name', 'STRING', mode='REQUIRED'),
+            SchemaField('age', 'INTEGER', mode='REQUIRED')
+        ]
+        creds = _make_credentials()
+        client = self._make_one(project=project, credentials=creds)
+        conn = client._connection = _Connection(resource)
+        table_ref = client.dataset(dataset_id).table(table_id)
+        table = Table(table_ref, schema=schema, client=client)
```
tswast: I'm a bit confused as to what `resource` is intended to be doing. I'm thinking that `table = Table()` should be changed to `table = client.get_table(table_ref)` so that this resource is used to simulate read-modify-write.

bigquery/tests/unit/test_client.py
tswast: Actually, I think this line should represent the full "before" schema.

bigquery/google/cloud/bigquery/table.py
alixhami: At this time, the schema will never be None. If it is set to None, the setter function sets it to an empty tuple. A later pull request will add the ability for the schema to be set to None.

bigquery/tests/unit/test_client.py
alixhami: Fixed with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/43611f45f4260ee45709201b0645fbab522f080b

bigquery/tests/unit/test_client.py
alixhami: Fixed with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/43611f45f4260ee45709201b0645fbab522f080b

bigquery/google/cloud/bigquery/table.py
alixhami: Added a map of field names to functions to populate the resource with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/013939afe3eaee6804ac38bd5f693a5d1845e9e1

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,41 +619,58 @@ def _set_properties(self, api_response):
...
+            'view_query': self._populate_view_query_resource,
+            'view_use_legacy_sql': self._populate_view_use_legacy_sql_resource,
+            'schema': self._populate_schema_resource
+        }
+        all_fields = [
+            'description', 'friendly_name', 'expires', 'location',
+            'partitioning_type', 'view_use_legacy_sql', 'view_query', 'schema'
+        ]
+        for f in all_fields:
+            if filter_fields and f not in filter_fields:
```
jba: It looks like you intend that if `filter_fields` is None or [], then all fields should be updated. Two issues I have with that:
1. If the caller of update_table accidentally passes in [], they'll overwrite their entire table. Let's avoid that case.
2. The logic on this line is a little tricky to follow.

One solution: in update_table, check that `properties` is a sequence. Here, make the default None, and near the top say
```
if filter_fields is None:
    filter_fields = all_fields
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,41 +619,58 @@ def _set_properties(self, api_response):
...
+                continue
+            if getattr(self, f) in (None, []) and f not in filter_fields:
+                continue
+            if f in custom_resource_fields:
+                custom_resource_fields[f](resource)
+            else:
+                # snake case to camel case
+                words = f.split('_')
```
jba: Add a TODO to factor out this code so both this and dataset can use it.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,41 +619,58 @@ def _set_properties(self, api_response):
...
-            }
-
+        custom_resource_fields = {
+            'expires': self._populate_expires_resource,
+            'partitioning_type': self._populate_partitioning_type_resource,
+            'view_query': self._populate_view_query_resource,
+            'view_use_legacy_sql': self._populate_view_use_legacy_sql_resource,
+            'schema': self._populate_schema_resource
+        }
+        all_fields = [
```
jba: Ditto.

bigquery/google/cloud/bigquery/table.py
alixhami: If the caller of update_table accidentally passes in [], they'll only update the fields that are not None or empty. Fields that are not filtered in that are None or empty are skipped in the next conditional on line 664. What is the desired behavior when the user accidentally passes in []?

With the current code, if the user wants an unfiltered resource, it will not include any properties with values that are None or empty. If there are filters, the resource will include the property regardless of its value. If I set `filter_fields` equal to `all_fields` when there aren't any filters, there's no longer a way to determine whether a field was specifically meant to be filtered in regardless of value.

jba: > What is the desired behavior when the user accidentally passes in []?

The call is a no-op.

Regarding the second point: I think this function is only called in two places: 

1. `create_table`,  in which case we want to send all the fields (or maybe all the non-empty ones; I'm not sure it matters, but in any case, only one of those two choices)
2. `update_table`, where the user passes in the list of fields and we send exactly those, regardless of their value.

I don't think there is a third case.

alixhami: @tswast had an idea for how to fix this. With https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/19af93abcde93af803b07f6f3ebe468e30c01c11 I updated `create_table()` to send in `all_fields` as the `filter_fields`. Then `_build_resource()` only loops through the filter_fields. This removed the messy logic. After the resource is built, `create_table()` removes any fields that are None before sending the API request. 

As for `update_table()`, with the way it is now set up, nothing will happen if a user passes in [] because the resource builder only loops through the `filter_fields`.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,41 +619,58 @@ def _set_properties(self, api_response):
             cleaned['expirationTime'] = float(cleaned['expirationTime'])
         self._properties.update(cleaned)
 
-    def _build_resource(self):
+    def _populate_expires_resource(self, resource):
+        value = _millis_from_datetime(self.expires)
```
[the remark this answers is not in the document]
alixhami: Fixed with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/19af93abcde93af803b07f6f3ebe468e30c01c11

bigquery/google/cloud/bigquery/table.py
```diff
@@ -616,41 +619,58 @@ def _set_properties(self, api_response):
...
-            view['query'] = self.view_query
-            if self.view_use_legacy_sql is not None:
-                view['useLegacySql'] = self.view_use_legacy_sql
-
-        if self._schema:
-            resource['schema'] = {
-                'fields': _build_schema_resource(self._schema)
-            }
-
+        custom_resource_fields = {
```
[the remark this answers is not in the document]
alixhami: Fixed with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/19af93abcde93af803b07f6f3ebe468e30c01c11

bigquery/google/cloud/bigquery/table.py
alixhami: Added TODO in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/19af93abcde93af803b07f6f3ebe468e30c01c11

bigquery/google/cloud/bigquery/table.py
alixhami: Fixed with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4076/commits/19af93abcde93af803b07f6f3ebe468e30c01c11

bigquery/google/cloud/bigquery/table.py
jba: SGTM.

PR #4079, opened 2017-09-28
Add DEFAULT sentinel for gapic_v1.method

jonparrott:
This is needed because the wrapped gapic methods can't discern between "unspecified" and "None" because the gapic clients will *always* specify retry and timeout, e.g.:

```python
def get_operation(name, retry=None, timeout=None):
     request = operations_pb2.GetOperationRequest(name=name)
     return self._get_operation(request, retry=retry, timeout=timeout)
```

gapic methods will use this sentinel to indicate unspecified:

```python
def get_operation(name, retry=gapic_v1.method.DEFAULT, timeout=gapic_v1.method.DEFAULT):
     request = operations_pb2.GetOperationRequest(name=name)
     return self._get_operation(request, retry=retry, timeout=timeout)
```

### DIFF
--- a/core/google/api/core/gapic_v1/method.py
+++ b/core/google/api/core/gapic_v1/method.py
@@ -34,6 +34,8 @@ _API_CORE_VERSION = pkg_resources.get_distribution('google-cloud-core').version
 METRICS_METADATA_KEY = 'x-goog-api-client'
 USE_DEFAULT_METADATA = object()
 DEFAULT = object()
+"""Sentinel value indicating that a retry or timeout argument was unspecified,
+so the default should be used."""
 
 
 def _is_not_none_or_false(value):
@@ -83,7 +85,8 @@ def _determine_timeout(default_timeout, specified_timeout, retry):
         default_timeout (Optional[Timeout]): The default timeout specified
             at method creation time.
         specified_timeout (Optional[Timeout]): The timeout specified at
-            invocation time.
+            invocation time. If :attr:`DEFAULT`, this will be set to
+            the ``default_timeout``.
         retry (Optional[Retry]): The retry specified at invocation time.
 
     Returns:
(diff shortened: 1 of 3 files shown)


core/google/api/core/gapic_v1/method.py
```diff
@@ -33,6 +33,7 @@
 _API_CORE_VERSION = pkg_resources.get_distribution('google-cloud-core').version
 METRICS_METADATA_KEY = 'x-goog-api-client'
 USE_DEFAULT_METADATA = object()
+DEFAULT = object()
```
dhermes: Want to give this value a docstring so you can refer to it?

core/google/api/core/gapic_v1/method.py
```diff
@@ -88,12 +89,16 @@ def _determine_timeout(default_timeout, specified_timeout, retry):
     Returns:
         Optional[Timeout]: The timeout to apply to the method or ``None``.
     """
+    if specified_timeout is DEFAULT:
+        specified_timeout = default_timeout
```
dhermes: This invalidates the docstring a bit?

core/tests/unit/api_core/gapic/test_method.py
```diff
@@ -97,6 +97,26 @@ def test_wrap_method_with_default_retry_and_timeout(unusued_sleep):
 
 
 @mock.patch('time.sleep')
+def test_wrap_method_with_default_retry_and_timeout_using_sentinel(
+        unusued_sleep):
+    method = mock.Mock(spec=['__call__'], side_effect=[
+        exceptions.InternalServerError(None),
+        42])
```
dhermes: This indenting disturbs my sensibilities for some reason. What about?

```python
method = mock.Mock(
    spec=['__call__'],
    side_effect=[
        exceptions.InternalServerError(None),
        42,
    ],
)
```

core/google/api/core/gapic_v1/method.py
jonparrott: Done.

core/google/api/core/gapic_v1/method.py
jonparrott: Updated it.

core/tests/unit/api_core/gapic/test_method.py
jonparrott: Done.

PR #4081, opened 2017-09-28
Add google.api.core.operations_v1

jonparrott:
Towards #3842

### DIFF
--- a/core/google/api/core/operations_v1/operations_client.py
+++ b/core/google/api/core/operations_v1/operations_client.py
@@ -1,0 +1,274 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A client for the google.longrunning.operations meta-API.
+
+This is a client that deals with long-running operations that follow the
+pattern outlined by the `Google API Style Guide`_.
+
+When an API method normally takes long time to complete, it can be designed to
+return ``Operation`` to the client, and the client can use this interface to
+receive the real response asynchronously by polling the operation resource to
+receive the response.
+
+It is not a separate service, but rather an interface implemented by a larger
+service. The protocol-level definition is available at
+`google/longrunning/operations.proto`_. Typically, this will be constructed
+automatically by another client class to deal with operations.
+
+.. _Google API Style Guide:
+    https://cloud.google.com/apis/design/design_pattern
+    s#long_running_operations
+.. _google/longrunning/operations.proto:
+    https://github.com/googleapis/googleapis/blob/master/google/longrunning
+    /operations.proto
+"""
+
+from google.api.core import gapic_v1
+from google.api.core.operations_v1 import operations_client_config
+from google.longrunning import operations_pb2
+
+
+class OperationsClient(object):
+    """Client for interacting with long-running operations within a service.
+
+    Args:
+        channel (grpc.Channel): The gRPC channel associated with the service
+            that implements the ``google.longrunning.operations`` interface.
+        client_config (dict):
+            A dictionary of call options for each method. If not specified
+            the default configuration is used.
+    """
+
+    def __init__(self, channel, client_config=operations_client_config.config):
+        # Create the gRPC client stub.
+        self.operations_stub = operations_pb2.OperationsStub(channel)
+
+        # Create all wrapped methods using the interface configuration.
+        # The interface config contains all of the default settings for retry
+        # and timeout for each RPC method.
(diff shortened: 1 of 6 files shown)


jonparrott:
Gentle ping on this @lukesneeringer @dhermes.

core/google/api/core/operations_v1/operations_client.py
```diff
@@ -0,0 +1,274 @@
...
+It is not a separate service, but rather an interface implemented by a larger
+service. The protocol-level definition is available at
+`google/longrunning/operations.proto`_. Typically, this will be constructed
+automatically by another client class to deal with operations.
+
+.. _Google API Style Guide:
+    https://cloud.google.com/apis/design/design_pattern
+    s#long_running_operations
+.. _google/longrunning/operations.proto:
+    https://github.com/googleapis/googleapis/blob/master/google/longrunning
```
dhermes: Need backslash continuation?

core/google/api/core/operations_v1/operations_client.py
```diff
@@ -0,0 +1,274 @@
...
+receive the response.
+
+It is not a separate service, but rather an interface implemented by a larger
+service. The protocol-level definition is available at
+`google/longrunning/operations.proto`_. Typically, this will be constructed
+automatically by another client class to deal with operations.
+
+.. _Google API Style Guide:
+    https://cloud.google.com/apis/design/design_pattern
+    s#long_running_operations
```
dhermes: Need backslash continuation?

core/google/api/core/operations_v1/operations_client.py
```diff
@@ -0,0 +1,274 @@
...
+        self.operations_stub = operations_pb2.OperationsStub(channel)
+
+        # Create all wrapped methods using the interface configuration.
+        # The interface config contains all of the default settings for retry
+        # and timeout for each RPC method.
+        interface_config = (
+            client_config
+            ['interfaces']
+            ['google.longrunning.Operations']
+        )
```
dhermes: WAT. Instead:

```python
interfaces = client_config['interfaces']
interface_config = interfaces['google.longrunning.Operations']
```

core/google/api/core/operations_v1/operations_client_config.py
```diff
@@ -0,0 +1,62 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""gapic configuration for the googe.longrunning.operations client."""
+
+config = {
+    "interfaces": {
```
dhermes: Is this autogen? single vs. double quotes?

core/google/api/core/operations_v1/operations_client_config.py
```diff
@@ -0,0 +1,62 @@
...
+                "idempotent": [
+                    "DEADLINE_EXCEEDED",
+                    "UNAVAILABLE"
+                ],
+                "non_idempotent": []
+            },
+            "retry_params": {
+                "default": {
+                    "initial_retry_delay_millis": 100,
+                    "retry_delay_multiplier": 1.3,
```
dhermes: `1.3` is not "exactly representable" in binary64. Why not `1.3125`? Or who cares is also appropriate

core/google/api/core/operations_v1/operations_client.py
jonparrott: I *think* sphinx handles these correctly, but if not, I'll fix them when I enable docs for all of these.

core/google/api/core/operations_v1/operations_client.py
jonparrott: Done.

core/google/api/core/operations_v1/operations_client_config.py
jonparrott: This is autogenerated.

core/google/api/core/operations_v1/operations_client_config.py
jonparrott: I don't know what you mean here.

dhermes: There is no such thing as `1.3` in binary because you can't divide by `5` in binary without having a repeating binary expansion. In binary, it's

```
1.3 = 1.0(1001)(1001)(1001)(1001)(1001)...._2
```

so the number has to be rounded to be put into `binary64`.

```
1.3125 = 1.0101_2
```

so no rounding needed.

dhermes:
LGTM, though can you build the docs locally and make sure those line-continuation links work correctly?

core/google/api/core/operations_v1/operations_client_config.py
jonparrott: I don't think we should worry about that right now. :)

dhermes: Right. Mine was more a warning: there is no such thing as `1.3` on today's computers, so use exactly representable numbers when possible.

lukesneeringer: Because the world will end if it gets rounded to something close enough?

dhermes: No. That's the beauty of our floating point system. I was just saying "don't expect it to be what you asked for". It's worth knowing, no?

PR #4056, opened 2017-09-25
Re-enable 'Bucket.requester_pays' feature

tseaver:
Closes #3474.

Needs extra system tests which pass `user_project` to the various APIs which take it, in the context of a bucket which is created with `requester_pays` enabled.

PR #4084, opened 2017-09-29
Extend test of 'requester_pays' bucket to include RUD w/ 'user_project' set

tseaver:
Skips the new test unless an environment variable, `GOOGLE_CLOUD_TESTS_USER_PROJECT`, is set.

@lukesneeringer What project should that be set to for CI?

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -28,6 +28,9 @@ from test_utils.retry import RetryErrors
 from test_utils.system import unique_resource_id
 
 
+USER_PROJECT = os.environ.get('GOOGLE_CLOUD_TESTS_USER_PROJECT')
+
+
 def _bad_copy(bad_request):
     """Predicate: pass only exceptions for a failed copyTo."""
     err_msg = bad_request.message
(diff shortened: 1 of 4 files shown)


tseaver:
This test, and more to be added shortly, should be in place on the `storage-requester_pays-feature` branch before merging #4056.

dhermes:
@tseaver What is RUD?

storage/tests/system.py
```diff
@@ -142,6 +138,47 @@ def test_bucket_update_labels(self):
         bucket.update()
         self.assertEqual(bucket.labels, {})
 
+    @unittest.skipUnless(USER_PROJECT, 'USER_PROJECT not set in environment.')
```
dhermes: Explicit `None`-check?

tseaver:
`RUD` is `CRUD` without the `C` (because we tested that already.

storage/tests/system.py
tseaver: Don't care:  if it is set to an empty string, I still don't want to run the test.

tseaver:
@lukesneeringer Do you have a sense about the project ID we should use for the `user_project` system tests?

PR #4085, opened 2017-09-29
Test bucket IAM/ACL methods w/ 'user_project' set.

tseaver:
~~Uses #4084 as a base.~~

PR #4086, opened 2017-09-29
Test blob CRUD methods w/ 'user_project' set.

tseaver:
~~Uses #4085 as a base.~~


tseaver:
Since @dhermes LGTM'ed the whole shebang in #4088, I will merge as soon as CI is green.

Issue #4087, opened 2017-09-29
Blob IAM support apparently invalid


tseaver:
While writing system tests for the `requester_pays` feature, I triggered the following:

```python
Traceback (most recent call last):
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/tests/system.py", line 372, in test_blob_acl_iam_w_user_project
    blob.set_iam_policy(policy)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/google/cloud/storage/blob.py", line 1148, in set_iam_policy
    _target_object=None)
  File "/home/tseaver/projects/agendaless/Google/src/google-cloud-python/storage/.nox/sys-3-6/lib/python3.6/site-packages/google/cloud/_http.py", line 293, in api_request
    raise exceptions.from_http_response(response)
google.api.core.exceptions.BadRequest: 400 PUT https://www.googleapis.com/storage/v1/b/new_1506705016579/o/SmallFile/iam?userProject=citric-celerity-697: roles/storage.objectViewer is not a valid role for projects/_/buckets/new_1506705016579/objects/SmallFile#0.
```

Note that the earlier call to `Blob.get_iam_policy` does succeed, returning:

```python
(Pdb) pp policy.to_api_repr()
{'bindings': [{'members': ['projectEditor:some-project-742',
                           'projectOwner:some-project-742',
                           'serviceAccount:1065521786570-19reuv03qbdp37du41inh9gtd1s35g1j@developer.gserviceaccount.com'],
               'role': 'roles/storage.legacyObjectOwner'},
              {'members': ['projectViewer:some-project-742'],
               'role': 'roles/storage.legacyObjectReader'},
              {'members': ['allUsers'], 'role': 'roles/storage.objectViewer'}],
 'etag': 'CAM='}
```

We don't have existing system tests for `Blob.set_iam_policy`, but the [Storage IAM docs](https://cloud.google.com/storage/docs/access-control/using-iam-permissions) don't define any IAM operations for blobs, only for buckets and projects.  Indeed, they say:

> To learn about controlling access to individual objects in your buckets, see [Access Control Lists](https://cloud.google.com/storage/docs/access-control/lists).

The [API documentation for Objects](https://cloud.google.com/storage/docs/json_api/v1/objects) also doesn't (any longer?) show `getIamPolicy`, `setIamPolicy`, or `testIamPermissions`.

[...]

PR #4088, opened 2017-09-29
Test blob ACL methods w/ 'user_project' set.

tseaver:
~~Uses #4086 as a base.~~

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -339,6 +339,27 @@ class TestStorageWriteFiles(TestStorageFiles):
             # Exercise 'objects.delete' (metadata) w/ userProject.
             blob.delete()
 
+    @unittest.skipUnless(USER_PROJECT, 'USER_PROJECT not set in environment.')
+    def test_blob_acl_w_user_project(self):
+        with_user_project = Config.CLIENT.bucket(
+            self.bucket.name, user_project=USER_PROJECT)
+        blob = with_user_project.blob('SmallFile')
+
+        file_data = self.FILES['simple']
+
+        blob.upload_from_filename(file_data['path'])
+        self.case_blobs_to_delete.append(blob)
+
+        # Exercise bucket ACL w/ userProject
+        acl = blob.acl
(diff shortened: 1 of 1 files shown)


tseaver:
Note that this test does not exercise the IAM features, because I now suspect them to be invalid (see #4087).

storage/tests/system.py
```diff
@@ -179,6 +179,44 @@ def test_crud_bucket_with_requester_pays(self):
...
+
+        with_up = Config.CLIENT.bucket(
+            new_bucket_name, user_project=USER_PROJECT)
+
+        # Exercise bucket ACL w/ userProject
+        acl = with_up.acl
+        acl.reload()
+        acl.all().grant_read()
+        acl.save()
+        self.assertTrue('READER' in acl.all().get_roles())
```
dhermes: [`assertIn`][1] exists

[1]: https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.assertIn

storage/tests/system.py
```diff
@@ -179,6 +179,44 @@ def test_crud_bucket_with_requester_pays(self):
...
             with_up.delete()
 
+    @unittest.skipUnless(USER_PROJECT, 'USER_PROJECT not set in environment.')
+    def test_bucket_acls_iam_with_requester_pays(self):
+        new_bucket_name = 'acl-w-user-project' + unique_resource_id('-')
+        created = Config.CLIENT.create_bucket(
+            new_bucket_name, requester_pays=True)
+        self.case_buckets_to_delete.append(new_bucket_name)
+
+        with_up = Config.CLIENT.bucket(
```
dhermes: `up` is ambiguous? (I realize I didn't complain in the previous PR.)

tseaver: I've updated that in #4085, and will fix here and in #4086 when rebasing them.

storage/tests/system.py
tseaver: I've updated that in #4085.

PR #4090, opened 2017-09-29
Test 'Bucket.get_blob' w/ 'user_project' set.

tseaver:


storage/tests/system.py
```diff
@@ -179,6 +179,27 @@ def test_crud_bucket_with_requester_pays(self):
...
+        self.case_buckets_to_delete.append(new_bucket_name)
+        self.assertEqual(created.name, new_bucket_name)
+        self.assertTrue(created.requester_pays)
+
+        with_user_project = Config.CLIENT.bucket(
+            new_bucket_name, user_project=USER_PROJECT)
+
+        self.assertIsNone(with_user_project.get_blob('nonesuch'))
+        to_add = created.blob('blob-name')
+        to_add.upload_from_string(b'DEADBEEF')
```
dhermes: Suggest: `data = b'DEADBEEF'`, then you won't need to copy it below.

As before, `try / finally` breaks from our usual clean-up. (It also breaks from Python's "one way to do it" mantra, since now we cleanup in system tests in two diff. ways.)

tseaver: > Suggest: data = b'DEADBEEF', then you won't need to copy it below.

244ff74

> `try / finally` breaks from our usual clean-up. (It also breaks from Python's "one way to do it" mantra, since now we cleanup in system tests in two diff. ways.)

Here, the 'try: finally:' is to deal with a blob added in a "bucket-centric" test, whose `tearDown` does not know from blobs (the storage system tests are factored somewhat oddly).

dhermes: > the storage system tests are factored somewhat oddly

So change them?

PR #4092, opened 2017-09-29
Test 'Blob.compose' w/ 'user_project' set.

tseaver:


storage/tests/system.py
```diff
@@ -547,6 +547,32 @@ def test_compose_replace_existing_blob(self):
         composed = original.download_as_string()
         self.assertEqual(composed, BEFORE + TO_APPEND)
 
+    @unittest.skipUnless(USER_PROJECT, 'USER_PROJECT not set in environment.')
```
dhermes: I'm beginning to think we should just have a `class TestWithUserProject(unittest.TestCase):` that has a `skipUnless` applied to the whole class.

storage/tests/system.py
```diff
@@ -547,6 +547,32 @@ def test_compose_replace_existing_blob(self):
...
+    def test_compose_with_user_project(self):
+        new_bucket_name = 'compose-user-project' + unique_resource_id('-')
+        created = Config.CLIENT.create_bucket(
+            new_bucket_name, requester_pays=True)
+        try:
+            SOURCE_1 = b'AAA\n'
+            source_1 = created.blob('source-1')
```
dhermes: 1. The underscores seem to be superfluous?
1. Having `SOURCE_1` and `source_1` is a bad idea. `data1` and `blob1` are much better names.

storage/tests/system.py
tseaver: We would have to replicate all the varioius `setUp` / `tearDown` bits I'm currently reusing.

storage/tests/system.py
tseaver: It is a trivial copy of an existing test with the same names. Should I be widening the scope of this PR to address the other one?

dhermes: I'm not sure what you mean. I see one test case with not great variable names.

PR #4093, opened 2017-09-29
Test 'Blob.rewrite' w/ 'user_project' set.

tseaver:
Note that both new tests are failing.

Because they have to be explicitly enabled (via the `GOOGLE_CLOUD_TESTS_USER_PROJECT`` environment variable, which we aren't yet setting for CI), this PR is OK to merge.

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -771,8 +771,62 @@ class TestStorageRewrite(TestStorageFiles):
         # Not adding 'dest' to 'self.case_blobs_to_delete':  it is the
         # same object as 'source'.
 
-        self.assertEqual(token, None)
+        self.assertIsNone(token)
         self.assertEqual(rewritten, len(source_data))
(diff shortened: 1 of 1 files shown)


storage/tests/system.py
```diff
@@ -595,3 +595,57 @@ def test_rewrite_rotate_encryption_key(self):
...
+
+            source = with_user_project.blob('source')
+            source.upload_from_filename(file_data['path'])
+            source_data = source.download_as_string()
+
+            KEY = os.urandom(32)
+            dest = with_user_project.blob('dest', encryption_key=KEY)
+            token, rewritten, total = dest.rewrite(source)
+
+            self.assertEqual(token, None)
```
dhermes: `assertIsNone`

tseaver:
@frankyn notes offline:

> Rewrite has a known issue with userproject and the Storage team has a fix for it, but pending review and release. I'll keep you posted.

PR #4094, opened 2017-09-30
Make QueryResults read-only.

tswast:
Removes any QueryJob-related properties from QueryResults. Now the
QueryResults class more closely reflects the backend resource.


tswast:
This change was reviewed in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4130

PR #4095, opened 2017-10-01
Use POST with Translate API

### LINKED ISSUE #4062: Use POST with Translate API
I try to translate a piece of text that's more than 2000 characters. The limit of a GET request to the translate API is 2000 characters. So my idea was to POST instead of GET, but that does not seem to be configurable, see https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/translate/google/cloud/translate_v2/client.py#L234 .

Is it possible to use POST with Translate?

Maerig:
Fixes #4062


lukesneeringer:
Thanks @Maerig! :-)

PR #4096, opened 2017-10-02
Avoiding `grpcio==1.6.0` in deps.

dhermes:
This is due to `google-gax` doing the same, which has broken
RTD builds:

https://readthedocs.org/projects/google-cloud-python/builds/6063446/

The motivation for avoiding `grpcio==1.6.0` is:

https://github.com/grpc/grpc/issues/12455

----

@lukesneeringer I'm unclear, but it's possible that `googleapis-common-protos` will need to have the strict upper bound set as well (it currently has `>= 1.0.0`).

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -51,7 +51,7 @@ SETUP_BASE = {
 
 
 REQUIREMENTS = [
-    'googleapis-common-protos >= 1.3.4',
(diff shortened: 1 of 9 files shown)


dhermes:
I am going to merge right now since this is just for RTD.

dhermes:
Fingers crossed: https://readthedocs.org/projects/google-cloud-python/builds/6064325/

PR #4098, opened 2017-10-02
Removing `googleapis-common-protos` from deps in non-`core` packages.

dhermes:
Also

- removing `grpcio` from non-`core` packages.
- manually specifying the `grpcio` dep in core (rather than getting it from `googleapis-common-protos[grpc]`)

(**NOTE**: This is a followup to #4096.)

### DIFF
--- a/videointelligence/setup.py
+++ b/videointelligence/setup.py
@@ -38,7 +38,6 @@ setup(
     ],
     packages=find_packages(exclude=('tests*',)),
     install_requires=(
-        'googleapis-common-protos >= 1.5.3, < 2.0dev',
         'google-gax >= 0.15.14, < 0.16dev',
         'six >= 1.10.0',
     ),
--- a/videointelligence/setup.py
+++ b/videointelligence/setup.py
@@ -39,6 +39,7 @@ setup(
     packages=find_packages(exclude=('tests*',)),
     install_requires=(
         'google-gax >= 0.15.14, < 0.16dev',
+        'grpcio >= 1.2.0, < 1.6dev',
         'six >= 1.10.0',
     ),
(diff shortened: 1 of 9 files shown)


dhermes:
@tseaver It already was, this is the line being removed from `core/setup.py`:

```
'googleapis-common-protos[grpc] >= 1.5.3, < 2.0dev',		
```

tseaver:
@dhermes Yeah, but literally just now in #4096, right?

dhermes:
Ahh good call. That was some extra fail copy-pasta by me in #4096.

tseaver:
ISTM that adding unconditional dependencies to an already-GA API should be worth some debate.

dhermes:
@tseaver Yes I did it by mistake in #4096, will undo here.

dhermes:
@tseaver PTAL, I made `grpc` an extra in `core`.

@jonparrott Does this look like the right approach for packaging?

videointelligence/setup.py
```diff
@@ -40,6 +40,7 @@
     install_requires=(
         'googleapis-common-protos >= 1.5.3, < 2.0dev',
         'google-gax >= 0.15.14, < 0.16dev',
+        'grpcio >= 1.2.0, < 1.6dev',
```
dhermes: @lukesneeringer Can you help me understand two things here (this is one of those auto-gen, flew by too fast, packages):

1. Does this package need `grpcio`? (If yes, I think we should explicitly pin here)
1. Why doesn't this package depend on `core`?

dhermes:
@lukesneeringer Can you follow-up on my question (https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4098#discussion_r142552072) when you get a chance?

PR #4103, opened 2017-10-02
bigquery: modify LoadJob

jba:
This PR handles loading from GCS. Loading from a local file will be done separately.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import collections
+import six
 import uuid
 
 from google.api.core import page_iterator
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -542,35 +540,153 @@ def cancelled(self):
...
-    _max_bad_records = None
-    _null_marker = None
-    _quote_character = None
-    _skip_leading_rows = None
-    _source_format = None
-    _write_disposition = None
+
+    def __init__(self):
+        self._properties = {}
+        self._schema = ()
```
tswast: I did something different for QueryJobConfig (not quite ready to mail yet, unfortunately)

https://github.com/tswast/google-cloud-python/blob/da1469f9d421df559d27150b39ce6e9932b65c6b/bigquery/google/cloud/bigquery/job.py#L1355

Basically, I stored the nested properties as the native objects in `_properties` but then do some conversion in the to/from API repr methods.

You might want to do something similar. Either way, you're right that we do need to treat the schema property differently.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -542,35 +540,153 @@ def cancelled(self):
...
+    def to_api_repr(self):
+        """Build an API representation of the load job config.
+
+        :rtype: dict
+        :returns: A dictionary in the format used by the BigQuery API.
+        """
+        config = copy.deepcopy(self._properties)
+        if len(self.schema) > 0:
+            config['schema'] = {'fields': _build_schema_resource(self.schema)}
+        # TODO(jba): add a comment explaining why skipLeadingRows is a string.
```
tswast: I bet the reason is that the proto file in the backend accepts int64, which isn't representable as a JSON number.

jba: Ack.

bigquery/google/cloud/bigquery/job.py
jba: Given that I have to treat setting a schema specially anyway, and there's only that and skipLeadingRows as special cases, I'm not sure that the extra machinery in that CL is worth it. Can we leave it like this for now?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -490,26 +491,37 @@ def list_jobs(self, max_results=None, page_token=None, all_users=None,
             max_results=max_results,
             extra_params=extra_params)
 
-    def load_table_from_storage(self, job_id, destination, *source_uris):
+    def load_table_from_storage(self, destination, source_uris,
```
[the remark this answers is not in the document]

bigquery/google/cloud/bigquery/client.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import collections
+import six
```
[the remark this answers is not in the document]

bigquery/google/cloud/bigquery/job.py
dhermes: Indeed: https://developers.google.com/discovery/v1/type-format

bigquery/google/cloud/bigquery/client.py
```diff
@@ -490,26 +492,37 @@ def list_jobs(self, max_results=None, page_token=None, all_users=None,
             max_results=max_results,
             extra_params=extra_params)
 
-    def load_table_from_storage(self, job_id, destination, *source_uris):
+    def load_table_from_storage(self, source_uris, destination,
+                                job_id=None, job_config=None):
         """Construct a job for loading data into a table from CloudStorage.
```
tswast: s/Construct/Starts

PR #4107, opened 2017-10-03
Do not use easily-misread glyphs in Firestore auto-IDs.

dhermes:
This PR removes `{'i', 'I', 'l', 'o', 'O', '0', '1'}` from the set of characters used in automatically-generated IDs, because it is easy to mis-read them if you are in a situation where you need to do that.

Exact set of blacklisted characters is loaned from Django's [`make_random_password`](https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#django.contrib.auth.models.BaseUserManager.make_random_password).

/cc @schmidt-sebastian

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/collection.py
+++ b/firestore/google/cloud/firestore_v1beta1/collection.py
@@ -24,8 +24,7 @@ from google.cloud.firestore_v1beta1 import query as query_mod
 from google.cloud.firestore_v1beta1.proto import document_pb2
 
 
-_AUTO_ID_CHARS = (
-    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789')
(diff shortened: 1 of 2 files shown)


tseaver:
I can't see how the test failure has anything to do with this change:

```python
_______________ test_wrap_method_with_overriding_retry_deadline _______________
  File "C:\projects\google-cloud-python\core\tests\unit\api_core\gapic\test_method.py", line 159, in test_wrap_method_with_overriding_retry_deadline
    assert timeout_args == [5, 10, 20, 29]
AssertionError: assert [5.0, 10.0, 20.0, 30.0] == [5, 10, 20, 29]
  At index 3 diff: 30.0 != 29
  Full diff:
  - [5.0, 10.0, 20.0, 30.0]
  + [5, 10, 20, 29]
============================== warnings summary ===============================
core/tests/unit/test_iam.py::TestPolicy::test_editors_setter
```

chemelnucfin:
Likewise going to pull the trigger here.  Please let me know if there are objections.

dhermes:
@chemelnucfin Can you revert this? We never got sign off from the Firestore team for this.

PR #4116, opened 2017-10-03
Retry for rows that return retryable error codes.

justinseinlin:
Retry some mutations that are retryable using similar mechanism as google.gax.retry.retryable.
However, we can't use retryable since it raises a RetryError upon final timeout. We want to return responses list with corresponding status codes for the given rows to be mutated.

PS: Creating a PR first to see if we agree on the implementation and how we want to configure the retry parameters (which are currently set locally since no other methods are using it). Will add test cases in another commit.


bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +374,64 @@ def sample_row_keys(self):
...
+        self.client = client
+        self.table_name = table_name
+        self.rows = rows
+        self.responses_statuses = [None for _ in six.moves.xrange(len(self.rows))]
+
+    def __call__(self):
+        return self._do_mutate_retryable_rows()
+
+    def _is_retryable(self, status):
+        return status is None or status.code in _RetryableMutateRowsWorker.RETRY_CODES
```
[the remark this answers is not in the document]
garye: Oh I see, the status get initialized to `None` for the first run. This is a bit confusing.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +374,69 @@ def sample_row_keys(self):
...
+        mutate_rows_request = _mutate_rows_request(self.table_name, retryable_rows)
+        responses = self.client._data_stub.MutateRows(mutate_rows_request)
+
+        has_retryable_responses = False
+        for response in responses:
+            index = 0
+            for entry in response.entries:
+                index = self._next_retryable_row_index(index)
+                self.responses_statuses[index] = entry.status
+                has_retryable_responses = self._is_retryable(entry.status)
```
garye: could `has_retryable_responses` get overwritten by a non-retryable response?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +374,64 @@ def sample_row_keys(self):
...
+class _MutateRowsRetryableError(Exception):
+    """A retryable error in Mutate Rows response."""
+    pass
+
+
+class _RetryableMutateRowsWorker(object):
+    RETRY_CODES = (
+        StatusCode.DEADLINE_EXCEEDED.value[0],
+        StatusCode.ABORTED.value[0],
+        StatusCode.INTERNAL.value[0],
```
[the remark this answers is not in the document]
justinseinlin: Done

bigtable/google/cloud/bigtable/table.py
justinseinlin: Fixed

justinseinlin:
I pushed a commit to just fix the lint errors first. But linter output is now unmanageably huge since it's checking inside google.cloud.bigtable._generated. Isn't that blocked by .flake8?

justinseinlin:
```
nox > Ran multiple sessions:
nox > * unit_tests(python_version='2.7'): success
nox > * unit_tests(python_version='3.4'): success
nox > * unit_tests(python_version='3.5'): success
nox > * unit_tests(python_version='3.6'): success
nox > * cover: success
```

justinseinlin:
Test for google.cloud.bigtable:

```
nox > * unit_tests(python_version='2.7'): success
nox > * unit_tests(python_version='3.4'): success
nox > * unit_tests(python_version='3.5'): success
nox > * unit_tests(python_version='3.6'): success
nox > * system_tests(python_version='2.7'): success
nox > * system_tests(python_version='3.6'): success
nox > * lint: success
nox > * lint_setup_py: success
nox > * cover: success
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -312,18 +317,25 @@ def mutate_rows(self, rows):
...
-        for response in responses:
-            for entry in response.entries:
-                responses_statuses[entry.index] = entry.status
-                if entry.status.code == 0:
-                    rows[entry.index].clear()
-        return responses_statuses
+        delay_millis = 1000
+        delay_mult = 1.5
+        max_delay_millis = 15 * 1000
+        total_timeout_millis = 5 * 60 * 1000
```
igorbernstein2: should these be configureable?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -312,18 +317,25 @@ def mutate_rows(self, rows):
...
+
+        now = time.time()
+        deadline = now + total_timeout_millis / _MILLIS_PER_SECOND
```
igorbernstein2: Currently, I don't think it matters, but if the retry params become user configureable, sub-millis precision will be lost here due to integer division.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +375,74 @@ def sample_row_keys(self):
...
+                index = self._next_retryable_row_index(index)
+                self.responses_statuses[index] = entry.status
+                if self._is_retryable(entry.status):
+                    num_retryable_responses += 1
+                if entry.status.code == 0:
+                    self.rows[index].clear()
+                index += 1
+
+        if num_retryable_responses:
+            raise _MutateRowsRetryableError()
```
igorbernstein2: Would be good to include a list of rows & their failures in the exception

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +375,74 @@ def sample_row_keys(self):
...
+        if not retryable_rows:
+            # All mutations are either successful or non-retryable now.
+            return self.responses_statuses
+
+        mutate_rows_request = _mutate_rows_request(
+            self.table_name, retryable_rows)
+        responses = self.client._data_stub.MutateRows(
+            mutate_rows_request)
+
+        num_retryable_responses = 0
```
igorbernstein2: Would be good to do a sanity check that # rows == # responses

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +375,74 @@ def sample_row_keys(self):
...
+        for response in responses:
+            index = 0
+            for entry in response.entries:
+                index = self._next_retryable_row_index(index)
```
igorbernstein2: This assumes that the responses and the entries will be in order (which is currently how the server is implemented: there is a single response and all entries are sorted by index). However, the protocol allows for async streaming responses from tablets that can be out of order. It would be better to use response.entries[].index for correlating the response to original request.

justinseinlin: Thanks. I made an assumption that wasn't stated. Fixed in the next commit.

bigtable/google/cloud/bigtable/table.py
justinseinlin: Do you mean `assert`? 
I thought of a `raise` as well but that means we have to update the `table.mutate_rows` API doc.

bigtable/google/cloud/bigtable/table.py
justinseinlin: I will include a list of `MutateRowsResponse.Entry` (with actual index to all rows). The caller already has a reference to the rows, so it can do any inspection.

justinseinlin: Done

bigtable/google/cloud/bigtable/table.py
```diff
@@ -363,6 +375,69 @@ def sample_row_keys(self):
...
+                index = index_into_all_rows[entry.index]
+                self.responses_statuses[index] = entry.status
+                if self._is_retryable(entry.status):
+                    entry.index = index  # Override with index into all rows
+                    retryable_responses.append(entry)
+                if entry.status.code == 0:
+                    self.rows[index].clear()
+
+        if retryable_responses:
+            raise _MutateRowsRetryableError(retryable_responses)
```
igorbernstein2: I think this will lose errors that are not retryable

justinseinlin: This exception is raised when there are retryable errors, so the caller can call again if they want to attempt retry. 
Otherwise, the callable returns a complete `responses_statuses` that contains the success, retryable and non-retryable statuses for corresponding rows.

igorbernstein2: Oh, I see, I didn't realize that this error was only used for flow control and wasn't bubbled up to the user.

bigtable/google/cloud/bigtable/table.py
igorbernstein2: I think assert would do.

jonparrott:
o/ I just wanted to make y'all aware of these features in core:

* https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/api/core/retry.py
* https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/api/core/exceptions.py#L425

bigtable/google/cloud/bigtable/table.py
justinseinlin: Changed to a float division in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4116/commits/28366582be449fe77115344abf10c81a2aa21ea1

bigtable/google/cloud/bigtable/table.py
justinseinlin: I intend them to be configurable, but wanted to check if we want it to be user-configurable or pre-configured at API level. 
- For the former, we may end up taking in an optional backoff/retry option at `table.mutate_rows`. 
- For the latter, a module tuple would be good if we see other (retryable) requests to share the same values for retry.

justinseinlin:
Removed custom retry logic and use `google.api.core.retry.Retry` in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4116/commits/8ab4d60ad6f729c8116d75ebfdb7c2a02b62c623

I added `RetryOptions` in `google.api.core.retry` as it looks like the more appropriate place than `google.cloud.bigtable.table`.  It is used as an argument to allow user-configured retry parameters. 
Please let me know if that's not the right place.

justinseinlin:
I may have broken the `cover` test by the introduction of `RetryOptions`.
Will fix that in the next commit.

core/google/api/core/retry.py
```diff
@@ -199,6 +200,34 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
     raise ValueError('Sleep generator stopped yielding sleep values.')
 
 
+class RetryOptions(
```
jonparrott: We intentionally do not use a separate retry options class, instead, preferring to just allow the user to pass in a `Retry` instance.

See the discussion here: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4148#issuecomment-335523296

justinseinlin: Thanks. I created the `RetryOptions` because to the user of `mutate_rows`, requiring the `Retry` instance would be impractical since they don't know what the `predicate` should be. The API method doesn't raise but return an error code for a corresponding row in case of an error. For them, only the retry parameters are meaningful.

Having said that, I understand the discussion. Will revert the `retry.py` in the next commit.

jonparrott: See the lower conversation in that same PR - BigQuery had a similar need to have its own predicate. I would suggest making the default retry a global constant so users can do:

```
retry = bigtable.DEFAULT_RETRY.with_deadline(60)
```

justinseinlin: Yes. I plan to do a `MUTATE_ROWS_DEFAULT_RETRY` which the user can customize with a desired `deadline`. Other retry params are more like impl detail for retrying, so less important to the user I guess.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -45,6 +49,22 @@ class TooManyMutationsError(ValueError):
...
 
+class _MutateRowsRetryableError(Exception):
+    """A retryable error in Mutate Rows response."""
+
+    def __init__(self, retryable_responses=None):
+        super(_MutateRowsRetryableError, self).__init__()
+        self.retryable_responses = retryable_responses or []
+
+
+MUTATE_ROWS_DEFAULT_RETRY = Retry(
```
justinseinlin: I guess we wanna place the constants at the top of the file as a convention, so this is not ideal 'cos it needs to have seen `_MutateRowsRetryableError`. I welcome suggestion for this please.

I'm also thinking if it's a good idea to rename `_MutateRowsRetryableError` into just a generic `_RetryableError` and remove the `retryable_responses` (since it's not used by the `google.core.api.Retry` anyway.

jonparrott: > I guess we wanna place the constants at the top of the file as a convention, so this is not ideal 'cos it needs to have seen _MutateRowsRetryableError. I welcome suggestion for this please.

As long as the linter is happy it's fine.

> I'm also thinking if it's a good idea to rename _MutateRowsRetryableError into just a generic _RetryableError and remove the retryable_responses (since it's not used by the google.core.api.Retry anyway.

If you do that then I would just use one of the many existing exception types in `google.api.core.exceptions`.

tseaver:
@justinseinlin 

> But linter output is now unmanageably huge since it's checking inside google.cloud.bigtable._generated. Isn't that blocked by .flake8?

Nope, we actually run both `flake8` and `pylint` (if `flake8` passes).  I just added PR #4195 to skip running `pylint` on the `_generated` directory.  In the meanwhile, your changes to appease `flake8` and add coverage seem good to me.

bigtable/google/cloud/bigtable/table.py
justinseinlin: Just saw your reply. Done in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4116/commits/184efbeebd1acd1ae57eac7ecfd372235fac3b32

justinseinlin:
Trying to resolve to remove the other commits that are not relevant.

justinseinlin:
Going to close this PR since I made a mistake in doing a rebase with commits already on the remote. 
A new PR is created with just the relevant commits cherry-picked on top of master HEAD for api.core -> api_core refactoring: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4256

PR #4119, opened 2017-10-04
bigquery: add Client.list_rows, remove Table.fetch_data

jba:


bigquery/google/cloud/bigquery/client.py
```diff
@@ -667,6 +668,79 @@ def query_rows(self, query, job_config=None, job_id=None, timeout=None):
...
+                     :class:`~google.cloud.bigquery.table.Table`
+                     :class:`~google.cloud.bigquery.table.TableReference`
+        :param table: the table to list, or a reference to it.
+
+        :type selected_fields: list of :class:`SchemaField`
+        :param selected_fields:
+            The fields to return. Required if ``table`` is a
+            :class:`~google.cloud.bigquery.table.TableReference`.
+
+        :type max_results: int
```
tswast: I'm not seeing the max_results parameter getting passed to the iterator / extra params.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -667,6 +668,79 @@ def query_rows(self, query, job_config=None, job_id=None, timeout=None):
...
+        :param max_results: (Optional) Maximum number of rows to return.
```
tswast: Since this returns an iterator which automatically fetches the next page, this is more accurately described as the batch size.

bigquery/tests/unit/test_client.py
```diff
@@ -1546,6 +1546,242 @@ def test_query_rows_w_job_config(self):
...
+        http = object()
+        client = self._make_one(project=PROJECT, credentials=creds, _http=http)
+        conn = client._connection = _Connection(DATA)
+        table_ref = DatasetReference(PROJECT, DS_ID).table(TABLE_ID)
+        full_name = SchemaField('full_name', 'STRING', mode='REQUIRED')
+        area_code = SchemaField('area_code', 'STRING', 'REQUIRED')
+        local_number = SchemaField('local_number', 'STRING', 'REQUIRED')
+        rank = SchemaField('rank', 'INTEGER', 'REQUIRED')
+        phone = SchemaField('phone', 'RECORD', mode='NULLABLE',
+                            fields=[area_code, local_number, rank])
```
tswast: I'd love to see a repeated field inside a record field.

bigquery/google/cloud/bigquery/client.py
jba: Good catch, fixed.

bigquery/google/cloud/bigquery/client.py
jba: I rewrote the doc, but mentioned pages instead of saying "batch size".

bigquery/tests/unit/test_client.py
jba: That is in test_list_rows_repeated_fields. See the definition of `struct`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -667,6 +668,81 @@ def query_rows(self, query, job_config=None, job_id=None, timeout=None):
...
+
+        iterator = page_iterator.HTTPIterator(
+            client=self,
+            api_request=self._connection.api_request,
+            path='%s/data' % (table.path,),
+            item_to_value=_item_to_row,
+            items_key='rows',
+            page_token=page_token,
+            next_token='pageToken',
+            max_results=max_results,
```
tswast: I think HTTPIterator actually does limit to one page of results when `max_results` is passed in as a keyword argument. That's why I was passing it in via the `extra_params` in the implementation of `fetch_data`. See: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3840

@craigcitro @jonparrott @tseaver @lukesneeringer @dhermes Do you know why the logic to cap at one page was added to the HTTPIterator? It seems to me that if someone wants to adjust the page size for performance reasons, it's pretty sucky to only get one page back.

bigquery/tests/unit/test_client.py
tswast: 👍

bigquery/google/cloud/bigquery/client.py
tswast: Thanks. I think there's a problem with this, though. Commented below. To get the correct behavior we need to either pass it in as `extra_params` or maybe remove some logic to cap results at one page from `HTTPIterator`.

bigquery/google/cloud/bigquery/client.py
jba: OK, I see what's going on here. To `HTTPIterator` and all the Client.list_XXX methods, `max_results` means exactly that: the maximum number of results you'll ever get. To the API, it means the maximum results _per call_, i.e. the page size.

At this point the minimum change would be to leave `max_results` as it is, and if we want, introduce `page_size`. That is the name that the [API Style Guide](https://cloud.google.com/apis/design/design_patterns#list_pagination) gives to the parameter, and a future version of BigQuery will likely conform to that guide, so eventually the client and backend will align.

jba: Never mind; page size is hard to get working, because the iterator does set the maxResults query param (correctly, in order to avoid reading too much).

We should probably leave the library as is, but fix the dbapi code so it doesn't set `max_results = self.array_size`; you could argue that that is the only thing that is actually broken.

tswast: I don't understand your second comment.

If you look at https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3845 it wasn't too tricky to convince it to act as a page size. I just had to pass it in as an "extra param".

:+1: to renaming it to `page_size` here to align with the API style guide.

tswast: Removing it from DBAPI would remove the whole point of the array_size parameter, which is used to control the batch size for performance reasons.

jba: If you change the meaning of `max_results` to mean the page size, then you'll break everyone who uses list_datasets, list_jobs, etc. who depends on its being the maximum number of values returned. You'll also break any other method in any client that uses `HTTPIterator` "as intended.".

You could argue that `max_results` is actually a more interesting arg than `page_size`, which is an implementation detail. (Most people probably just write a `for` loop over the iterator.) 
I guess you could also argue that if you only want `n` values, just stop looping when a counter reaches `n`. There is also a performance cost to that, though, when the batch size is larger than the remaining number of items you want to fetch.

So ideally both would be nice to have, but that's tricky given the current implementation of `HTTPIterator`, which effectively co-ops the API `max_results` query param to implement its meaning of `max_results`.

FWIW, in Go we don't have a `max_results` equivalent, only `page_size`.

tswast: `max_results` certainly is more interesting than `page_size`, but I'd rather have `page_size` than `max_results`, honestly. I guess we could support both.

If we do just do `max_results`, you're right that the DB-API implementation needs to be fixed.

jba: I'm fine with any of the above, but since it involves consistency across many (all?) clients, we should ask @lukesneeringer and @jonparrott what they think.

But in parallel, to make progress, I suggest that we have `list_rows` handle `max_results` exactly as every other `Clist_XXX` method here does, and leave out `page_size` for now. If we do change the meaning of `max_results`, we'd want to do it across all `list_XXX` methods in this client, and probably the others as well, so it is certainly out of scope for this PR.

tswast: SGTM

PR #4127, opened 2017-10-05
Add support for bucket notifications

### LINKED ISSUE #3956: Add support for storage->pubsub notifications

tseaver:
Merges already-reviewed work on the `3956-storage-notifications` branch to `master`,

Closes #3956.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -35,6 +35,7 @@ from google.cloud.storage.acl import BucketACL
 from google.cloud.storage.acl import DefaultObjectACL
 from google.cloud.storage.blob import Blob
 from google.cloud.storage.blob import _get_encryption_headers
+from google.cloud.storage.notification import BucketNotification
 
 
 def _blobs_page_start(iterator, page, response):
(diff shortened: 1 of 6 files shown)


dhermes:
Cursory LGTM? (I didn't look at source, you said already reviewed.)

PR #4130, opened 2017-10-05
BQ: remove client.run_async_query, add client.query

tswast:
Base PR https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4094.

- Add a QueryJobConfig class.
- Move configuration properties from QueryJob to QueryJobConfig.
- Make standard SQL dialect the default.
- Add a helper property class `_ListApiResourceProperty` for handling repeated types.


bigquery/google/cloud/bigquery/client.py
```diff
@@ -656,13 +654,6 @@ def query_rows(self, query, job_config=None, job_id=None, timeout=None):
             given timeout.
         """
         job_id = _make_job_id(job_id)
```
jba: It would be even better if query_rows were
```
job = self.query(...)
return job.result(...)
```

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1253,53 +1389,52 @@ class QueryJob(_AsyncJob):
...
 
         self.query = query
-        self.udf_resources = udf_resources
-        self.query_parameters = query_parameters
         self._configuration = job_config
         self._query_results = None
 
     @property
-    def use_legacy_sql(self):
+    def allow_large_results(self):
```
jba: I thought the plan was to remove all these delegations from the job class? Looks like you removed them then added them back.

tswast: I removed the setters but not the getters. We could potentially make `job.configuration` public if we wanted to get rid of the getters, but I wasn't sure it was worth the extra layer of indirection.

jba: SGTM. I will add getters to LoadJob.

bigquery/google/cloud/bigquery/client.py

PR #4133, opened 2017-10-06
Catch checksum validation failure from download, delete corrupt file

mfschwartz:
This change depends on https://github.com/GoogleCloudPlatform/google-resumable-media-python/pull/31

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -492,7 +492,10 @@ class Blob(_PropertyMixin):
         except resumable_media.DataCorruption as exc:
             # Delete the corrupt downloaded file.
             os.unlink(file_obj.name)
-            raise
+            raise exceptions.from_http_status(
+                exc.response.status_code, exc.message +
+                     ".  Deleted corrupt downloaded file %s" % file_obj.name)
+
 
     def download_to_filename(self, filename, client=None):
         """Download the contents of this blob into a named file.
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -489,6 +489,10 @@ class Blob(_PropertyMixin):
             self._do_download(transport, file_obj, download_url, headers)
         except resumable_media.InvalidResponse as exc:
             _raise_from_invalid_response(exc)
+        except resumable_media.DataCorruption as exc:
+            # Delete the corrupt downloaded file.
+            os.unlink(file_obj.name)
+            raise
 
     def download_to_filename(self, filename, client=None):
         """Download the contents of this blob into a named file.
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -489,13 +489,6 @@ class Blob(_PropertyMixin):
             self._do_download(transport, file_obj, download_url, headers)
         except resumable_media.InvalidResponse as exc:
             _raise_from_invalid_response(exc)
-        except resumable_media.DataCorruption as exc:
-            # Delete the corrupt downloaded file.
-            os.unlink(file_obj.name)
-            raise exceptions.from_http_status(
-                exc.response.status_code, exc.message +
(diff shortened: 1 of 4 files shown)


tseaver:
Needs a new unit test for the case where the `download` object raises `resumable_media.DataCorruption`.

mfschwartz:
@tseaver - there's an integration test for this case

tseaver:
@mfschwartz We require 100% unit test coverage:  see the [current CI failure](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3661).

Note the linter failures in that output as well.

dhermes:
@tseaver @mfschwartz I'm happy to take up the testing mantle. Before this can go in I need to get an LGTM on

https://github.com/GoogleCloudPlatform/google-resumable-media-python/pull/33

and cut a release there.

storage/google/cloud/storage/blob.py
```diff
@@ -503,8 +503,13 @@ def download_to_filename(self, filename, client=None):
...
         """
-        with open(filename, 'wb') as file_obj:
-            self.download_to_file(file_obj, client=client)
+        try:
+            with open(filename, 'wb') as file_obj:
+                self.download_to_file(file_obj, client=client)
+        except resumable_media.DataCorruption as exc:
+            # Delete the corrupt downloaded file.
+            os.remove(filename)
+            raise
```
dhermes: @mfschwartz Can you take a peek at this change? It is different in two ways from your original change:

- It is in `download_to_filename` (vs. `download_to_file`). This method is the only method where we **know** we have created the file, so it is safe to delete it. In `download_to_file`, the user could pass in a non-file but file-like object, so there would be nothing to delete. (It should really be named `download_to_stream`, but we are already 1.0 so yay!)
- I am just calling `raise` to re-raise the error. The previous approach `raise exceptions.from_http_status` didn't make much sense because the status code would be 200 or 206 (and I think it would just be a 200). We don't have any corresponding error class, so it'd just [be][1] a generic `GoogleAPICallError`

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/5cc463f934f165fc8685725d9f0b60a1043dd50d/core/google/api/core/exceptions.py#L337

dhermes:
@mfschwartz RE:

> there's an integration test for this case

He (Tres) was referring to tests in this library, not in `google-resumable-media` (which is in some sense an opaque dependency).

storage/google/cloud/storage/blob.py
mfschwartz: LGTM. Thanks for those fixes.

dhermes:
@jonparrott @lukesneeringer @tseaver Can I get an LGTM?

dhermes:
Since @tseaver has made a **lot** of `user_project` changes since `1.4.0` I am toying with doing a release for this with a cherry-pick branch (though I'm not that excited about it).

I made one [branch][1] just by starting from the `storage-1.4.0` tag, cherry-picking #4133 and #4180 and then "fixing" it up ([build][3]).

I made another [branch][2] from `master` and then just did a `git show storage-1.4.0:${NAME} > ${NAME}` for all the changed files, then made sure the changes from #4133 and #4180 stayed in while `git add`-ing ([build][4]).

Both approaches will make figuring out what has changed for the next release difficult. It will at least be difficult via `git log storage-1.4.1..HEAD` since they will no longer share a merge base.

So the question I put to @lukesneeringer and @jonparrott, is it worth it to cherry-pick this?

The question I put to @tseaver, could we just **release** the `user_project` stuff as-is? If yes, please take the lead on writing the release notes.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/commits/storage-cherry-pick
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/commits/storage-cherry-pick-via-delete
[3]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3758
[4]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3759

PR #4136, opened 2017-10-07
bigquery: add Client.load_table_from_file

jba:
Move the method from Table to Client.


bigquery/google/cloud/bigquery/client.py
```diff
@@ -525,6 +544,171 @@ def load_table_from_storage(self, source_uris, destination,
...
+
+        :type file_obj: file
+        :param file_obj: A file handle opened in binary mode for reading.
+
+        :type destination: :class:`google.cloud.bigquery.table.TableReference`
+        :param destination: Table into which data is to be loaded.
+
+        :type rewind: bool
+        :param rewind: If True, seek to the beginning of the file handle before
+                       writing the file.
```
tswast: before reading, right?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -525,6 +544,171 @@ def load_table_from_storage(self, source_uris, destination,
...
+                 a file opened in text mode.
+        """
+        job_id = _make_job_id(job_id)
+        job = LoadJob(job_id, None, destination, self, job_config)
+        job_resource = job._build_resource()
+        if rewind:
+            file_obj.seek(0, os.SEEK_SET)
+        _check_mode(file_obj)
+        try:
+            if size is None:
```
tswast: I find this a bit odd to switch between resumable upload and multipart upload based solely on the size being none. We should also check if the size is not none but too big.

Based on what I know about other Apiary endpoints [multipart upload should only be used for small files](https://cloud.google.com/storage/docs/json_api/v1/how-tos/multipart-upload).

> For larger files (more than 5 MB) or less reliable network connections, use resumable upload instead.

IIRC, we have got feedback from users that uploads with can fail. I think this could be the reason.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -525,6 +544,171 @@ def load_table_from_storage(self, source_uris, destination,
...
+        :type destination: :class:`google.cloud.bigquery.table.TableReference`
+        :param destination: Table into which data is to be loaded.
+
+        :type rewind: bool
+        :param rewind: If True, seek to the beginning of the file handle before
+                       writing the file.
+
+        :type size: int
+        :param size: The number of bytes to read from the file handle.
+                     If not provided, we'll try to guess the size using
```
tswast: I'm not sure that the `os.fstat` comment is correct, since I'm not seeing it here. I think this argument should talk about resumable vs multi-part upload since that's what it's being used for as far as I can tell.

bigquery/google/cloud/bigquery/client.py

bigquery/google/cloud/bigquery/client.py

bigquery/google/cloud/bigquery/client.py

PR #4144, opened 2017-10-09
BQ: remove fetch_data and client references from QueryResults

tswast:
Based on https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4143.

Now QueryResults is just a container for the getQueryResults response.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -432,7 +432,7 @@ class Client(ClientWithProject):
         resource = self._connection.api_request(
             method='GET', path=path, query_params=extra_params)
 
-        return QueryResults.from_api_repr(resource, self)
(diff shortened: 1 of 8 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -739,7 +739,8 @@ def list_rows(self, table, selected_fields=None, max_results=None,
 
         params = {}
         if selected_fields is not None:
-            params['selectedFields'] = [f.name for f in selected_fields]
+            params['selectedFields'] = ','.join(
+                [f.name for f in selected_fields])
```
jba: You don't need the square brackets (i.e. use a generator instead of a list).

tswast:
Coverage failure is for `list_partitions`. I'll double-check the diff to make sure I didn't delete a test.

PR #4145, opened 2017-10-09
bigquery: remove Table.exists

jba:


jo8937:
why remove exists method??

BruceWouaigne:
Yes indeed, WHY??

jba:
You can write it as a call to `get_table` wrapped in a `try`.

PR #4146, opened 2017-10-09
bigquery: add Client.list_partitions

jba:
Remove from Table.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -37,7 +37,7 @@ from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.job import CopyJob
 from google.cloud.bigquery.job import ExtractJob
 from google.cloud.bigquery.job import LoadJob
-from google.cloud.bigquery.job import QueryJob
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -942,6 +942,25 @@ def list_rows(self, table, selected_fields=None, max_results=None,
...
+                     :class:`~google.cloud.bigquery.table.TableReference`
+        :param table: the table to list, or a reference to it.
+
+        :rtype: list
+        :returns: a list of time partitions
+        """
+        config = QueryJobConfig()
+        config.use_legacy_sql = True  # required for '$' syntax
+        rows = self.query_rows(
+            'SELECT partition_id from [%s.%s$__PARTITIONS_SUMMARY__]' %
```
tswast: We should probably add the project ID, too, since they might want to list partitions on any table reference, not just one with this client's project.

`'SELECT partition_id from [%s:%s.%s$__PARTITIONS_SUMMARY__]' % (table.project, table.dataset_id, table.table_id)`

PR #4148, opened 2017-10-10
bigquery: retry idempotent RPCs

jba:
Add retry logic to every RPC for which it makes sense.

Following the BigQuery team, we ignore the error code and
use the "reason" field of the error to determine whether
to retry.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import collections
+import functools
 import os
 import uuid
 
@@ -27,6 +28,7 @@ from google.resumable_media.requests import MultipartUpload
 from google.resumable_media.requests import ResumableUpload
 
 from google.api.core import page_iterator
+
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
 from google.cloud.bigquery._http import Connection
@@ -44,6 +46,7 @@ from google.cloud.bigquery._helpers import _item_to_row
 from google.cloud.bigquery._helpers import _rows_page_start
 from google.cloud.bigquery._helpers import _field_to_index_mapping
 from google.cloud.bigquery._helpers import _SCALAR_VALUE_TO_JSON_ROW
+from google.cloud.bigquery._helpers import DEFAULT_RETRY
 
 
 _DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
@@ -117,7 +120,8 @@ class Client(ClientWithProject):
             project=project, credentials=credentials, _http=_http)
         self._connection = Connection(self)
 
-    def list_projects(self, max_results=None, page_token=None):
+    def list_projects(self, max_results=None, page_token=None,
+                      retry=DEFAULT_RETRY):
         """List projects for the project associated with this client.
 
         See
@@ -132,13 +136,16 @@ class Client(ClientWithProject):
                            not passed, the API will return the first page of
                            projects.
 
+        :type retry: :class:`google.api.core.retry.Retry`
+        :param retry: (Optional) How to retry the RPC.
+
         :rtype: :class:`~google.api.core.page_iterator.Iterator`
         :returns: Iterator of :class:`~google.cloud.bigquery.client.Project`
                   accessible to the current client.
         """
         return page_iterator.HTTPIterator(
             client=self,
-            api_request=self._connection.api_request,
+            api_request=functools.partial(self._call_api, retry),
             path='/projects',
             item_to_value=_item_to_project,
             items_key='projects',
@@ -146,7 +153,7 @@ class Client(ClientWithProject):
             max_results=max_results)
 
     def list_datasets(self, include_all=False, max_results=None,
-                      page_token=None):
+                      page_token=None, retry=DEFAULT_RETRY):
         """List datasets for the project associated with this client.
 
         See
@@ -164,6 +171,9 @@ class Client(ClientWithProject):
                            not passed, the API will return the first page of
                            datasets.
 
+        :type retry: :class:`google.api.core.retry.Retry`
+        :param retry: (Optional) How to retry the RPC.
+
         :rtype: :class:`~google.api.core.page_iterator.Iterator`
         :returns: Iterator of :class:`~google.cloud.bigquery.dataset.Dataset`.
                   accessible to the current client.
@@ -174,7 +184,7 @@ class Client(ClientWithProject):
         path = '/projects/%s/datasets' % (self.project,)
         return page_iterator.HTTPIterator(
             client=self,
-            api_request=self._connection.api_request,
+            api_request=functools.partial(self._call_api, retry),
             path=path,
             item_to_value=_item_to_dataset,
             items_key='datasets',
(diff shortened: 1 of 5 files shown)


jba:
This is a first cut at adding retry to the BigQuery client. My thinking is that most retry parameters are not interesting to users and should not be exposed, but the timeout (a.k.a. `deadline`) is important, and should be an optional parameter.

Another choice would be to allow an optional google.gax.CallOptions, as the generated code does.

jonparrott:
@jba this is a good start, but needs to be slightly different (and I apologize for not documenting what the retry pattern should look like).

1. All methods that retry should take a full `retry` argument that should be a retry instance.
1. A sane default retry should be provided.
1. Retry can be set to `False` or `None` to disable retrying.
1. The connection's api call should handle raising the appropriate `google.api.core.exceptions.GoogleAPICallError` subclass. This way, you don't need custom `should_retry` and users will get the appropriate exception if retrying fails.

So `get_dataset` should look something like this:

```python
def get_dataset(self, dataset_ref, retry=DEFAULT_RETRY):
    api_call = functools.partial(
        self._connection.api_request,
        method='GET',
        path=dataset_ref.path)

    if retry:
        api_call = retry(retry)

    api_response = api_call()

    return Dataset.from_api_repr(api_response)
```

and `self._connection.api_request` should handle raising the appropriate exception on error.

jba:
Thanks for the explanation. The one thing I'm not clear about is the exception subclass. BigQuery retry ignores the error code and is based solely on the `reason` string in the error. Should I define a special `BigQueryTransientError` subclass to capture that, which wraps the original error? If so, I'll still need a custom predicate, won't I? And won't that be worse for users when retry fails, because they'll have to call `RetryError.cause()` and then call `cause()` on that again?

jonparrott:
> BigQuery retry ignores the error code and is based solely on the reason string in the error.

Yikes. Do the error codes at least match up with idempotent http statuses?

> Should I define a special BigQueryTransientError subclass to capture that, which wraps the original error?

Yeah, that could be reasonable, something like:

```
class BigQueryTransientError(exceptions.BackendError):
     ...
```

> If so, I'll still need a custom predicate, won't I?

Possibly, if the default predicate doesn't work (e.g. if not all BackendErrors are retryable, just the subset that ends up being BigQueryTransientErrors). If that's the case, you can make it global:

```python
google.cloud.bigquery.default_retry_predicate
```

That way custom retries are a bit easier:

```python
myretry = retry.Retry(bigquery.default_retry_predicate, deadline=60)
```

You can also make the default retry object itself a global or class constant, so users can do;

```python
myretry = bigquery_client.DEFAULT_RETRY.with_deadline(60)
```

>  And won't that be worse for users when retry fails, because they'll have to call RetryError.cause() and then call cause() on that again?

If you sublcass one of the existing exception classes they should only need to call `cause()` once, unless I'm misunderstanding something.

jba:
> Yikes. Do the error codes at least match up with idempotent http statuses?

The error codes are irrelevant. Only the reason field matters. So I guess the answer is no.

> If so, I'll still need a custom predicate, won't I?
> > Possibly, if the default predicate doesn't work (e.g. if not all BackendErrors are retryable, just the subset that ends up being BigQueryTransientErrors).

The default predicate won't work, for the reason you said.

> You can also make the default retry object itself a global or class constant, so users can do;
> ```
> myretry = bigquery_client.DEFAULT_RETRY.with_deadline(60)
> ```

Done.

jba:
I don't understand the error I'm getting. It happens only under 2.7. Am I holding `functools` wrong?
```
  File "/var/code/gcp/bigquery/tests/unit/test_client.py", line 297, in test_get_dataset
    dataset = client.get_dataset(dataset_ref)
  File "/var/code/gcp/bigquery/google/cloud/bigquery/client.py", line 288, in get_dataset
    api_call = retry(api_call)
  File "/var/code/gcp/.nox/unit-2-7/lib/python2.7/site-packages/google/api/core/retry.py", line 247, in __call__
    @six.wraps(func)
  File "/var/code/gcp/.nox/unit-2-7/lib/python2.7/site-packages/six.py", line 811, in wrapper
    f = functools.wraps(wrapped, assigned, updated)(f)
  File "/usr/local/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'functools.partial' object has no attribute '__module__'
```

bigquery/google/cloud/bigquery/client.py
```diff
@@ -239,18 +267,26 @@ def create_table(self, table):
...
+        :type retry: :class:`google.api.core.retry.Retry`
+        :param retry: (Optional) How to retry the RPC.
+
         :rtype: :class:`google.cloud.bigquery.dataset.Dataset`
         :returns: a ``Dataset`` instance
         """
-        api_response = self._connection.api_request(
-            method='GET', path=dataset_ref.path)
+        api_call = functools.partial(
+            self._connection.api_request,
```
tswast: I think this may be failing because `api_request` is not getting the `self` argument. Try adding `self._connection,` as a parameter.

tswast: Nevermind, looks like it's more subtle than that. https://stackoverflow.com/a/28752007/101923

jba: I don't know Python well enough to translate that SO page into a fix. @lukesneeringer @jonparrott can you help?

jonparrott: Ugh, wraps is such a pain. I will send a fix for this to core, but I don't have time right now as I'm fire drilling. I'll see what I can do by the end of this week.

jonparrott: Created https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4166

jonparrott: Okay, this is now in core, so if you rebase it should work fine.

jba:
PTAL. In the latest push I added retry to nearly all methods that can support it.

core/google/api/core/page_iterator.py
```diff
@@ -291,7 +293,8 @@ class HTTPIterator(Iterator):
     def __init__(self, client, api_request, path, item_to_value,
                  items_key=_DEFAULT_ITEMS_KEY,
                  page_token=None, max_results=None, extra_params=None,
-                 page_start=_do_nothing_page_start, next_token=_NEXT_TOKEN):
+                 page_start=_do_nothing_page_start, next_token=_NEXT_TOKEN,
+                 retry=None):
```
jonparrott: I wouldn't make this change. Since `api_request` is a callable, you can wrap it in retry before passing to `page_iterator`, e.g.:

```python
api_request = functools.partial(self._call_api, retry)
```

lukesneeringer:
@jba

> I don't understand the error I'm getting. It happens only under 2.7. Am I holding `functools` wrong?
> ```
  File "/var/code/gcp/bigquery/tests/unit/test_client.py", line 297, in test_get_dataset
    dataset = client.get_dataset(dataset_ref)
  File "/var/code/gcp/bigquery/google/cloud/bigquery/client.py", line 288, in get_dataset
    api_call = retry(api_call)
  File "/var/code/gcp/.nox/unit-2-7/lib/python2.7/site-packages/google/api/core/retry.py", line 247, in __call__
    @six.wraps(func)
  File "/var/code/gcp/.nox/unit-2-7/lib/python2.7/site-packages/six.py", line 811, in wrapper
    f = functools.wraps(wrapped, assigned, updated)(f)
  File "/usr/local/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'functools.partial' object has no attribute '__module__'
```

When you create a function on the fly with `functools.partial`, it does not assign it a `__module__` property, which is assigned when a function is built with the `def` keyword. `@six.wraps` (which is just an alias to `functools.wraps`) requires a `__module__` property for aliasing the docstring.

You can get around it with:

```python
# Make the partial that you were making before.
partial_func = functools.partial(func, *args, **kwargs)

# Assign it an appropriate __module__ property based on the function it wraps.
partial_func.__module__ = func.__module__
```

After that it should work with `wraps`.

jonparrott:
@lukesneeringer No need, we already fixed that in core. :)

core/google/api/core/page_iterator.py
jba: Done.

jonparrott:
retry stuff LGTM

PR #4149, opened 2017-10-10
bigquery: add Row class

jba:
Rows are represented by a new class, Row, which allows access by position or field name.

### DIFF
--- a/bigquery/tests/unit/test__helpers.py
+++ b/bigquery/tests/unit/test__helpers.py
@@ -377,12 +377,12 @@ class Test_record_from_json(unittest.TestCase):
         self.assertEqual(coerced, expected)
 
 
-class Test_row_from_json(unittest.TestCase):
+class Test_row_tuple_from_json(unittest.TestCase):
 
     def _call_fut(self, row, schema):
-        from google.cloud.bigquery._helpers import _row_from_json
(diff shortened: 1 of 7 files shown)


bigquery/tests/unit/test__helpers.py
```diff
@@ -468,15 +468,36 @@ def test_w_array_of_struct_w_array(self):
...
+
+        VALUES = (1, 2, 3)
+        r = Row(VALUES, {'a': 0, 'b': 1, 'c': 2})
+        self.assertEqual(r.a, 1)
+        self.assertEqual(r[1], 2)
+        self.assertEqual(r['c'], 3)
+        self.assertEqual(len(r), 3)
+        self.assertEqual(r.values(), VALUES)
+        self.assertEqual(repr(r),
+                         "Row((1, 2, 3), {'a': 0, 'b': 1, 'c': 2})")
```
tswast: This test is failing because the order of dictionary elements isn't guaranteed.

It might not be a bad idea to sort them alphabetically in `__repr__`. Or just pass in an [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) in this test.

jba: I sorted them by value (index), since that will be the order they appear in the schema.

PR #4160, opened 2017-10-11
BigQuery: Adds helper function for snake to camel case conversion

alixhami:
New helper function to factor out repetitive code in client.py and table.py


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -366,6 +366,12 @@ def _time_to_json(value):
 _SCALAR_VALUE_TO_JSON_PARAM['TIMESTAMP'] = _timestamp_to_json_parameter
 
 
+def _snake_to_camel_case(value):
```
tswast: Could you add a unit test for this to `test__helpers.py`?

lukesneeringer:
Do we not have this in core somewhere?

alixhami:
Just asked @jonparrott and he said there aren't any helpers that do this.

PR #4174, opened 2017-10-12
Move debug statement with blocking future.result() call to add_done_callback

### LINKED ISSUE #4167: PubSub: Default thread-based subscriber Policy's on_response() implementation degenerates into series of synchronous calls because of debug statement
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4e17158c0f14e0e91cbeb74c3e8483c5a231e9ac/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py#L147

While testing out the new PubSub API (v 0.28.3), I noticed that the default thread-based `Policy`, which uses a `ThreadPoolExecutor(max_workers=10)` seemed to be executing the subscriber callbacks sequentially, instead of asynchronously.  Upon introspection of the source, it looks like the above referenced code:
```
...
future = self._executor.submit(self._callback, message)
logger.debug('Result: %s' % future.result())
```
will block until `self._callback` completes, because of the call to `future.result()`.  Or am I losing my mind?  Thanks!

jeffkpayne:
Closes #4167

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -30,6 +30,11 @@ from google.cloud.pubsub_v1.subscriber.message import Message
 logger = logging.getLogger(__name__)
 
 
+def _callback_completed(future):
+    """Simple callback that just logs a `Future`'s result."""
+    logger.debug('Result: %s', future.result())
+
+
 class Policy(base.BasePolicy):
     """A consumer class based on :class:`threading.Thread`.
 
@@ -144,4 +149,4 @@ class Policy(base.BasePolicy):
             logger.debug(self._callback)
             message = Message(msg.message, msg.ack_id, self._request_queue)
             future = self._executor.submit(self._callback, message)
-            logger.debug('Result: %s' % future.result())
+            future.add_done_callback(_callback_completed)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -139,9 +139,12 @@ def on_response(self, response):
 
         For each message, schedule a callback with the executor.
         """
+        def callback_completed(future):
+            logger.debug('Result: %s' % future.result())
```
dhermes: 1. `logger.debug('Result: %s', future.result())` (i.e. `logger.debug` will format for you)
1. Can you define this at the module level? It looks like a closure but doesn't enclose any scope.
1. Otherwise LGTM, though maybe the tests should be updated? (I'll defer to @lukesneeringer on that).

jeffkpayne: 1. Gah, yeah, will change that.  I simply copied the line of code from the previous location :)
2. Will do
3. Wasn't really sure of the best way to test for a debug statement being output.  If there are constructs/patterns in place that test for logging output (using mocks, whatever), I'll gladly add one for this.

dhermes: No worries about 1, sometimes its "you break it you buy it".

@lukesneeringer Care to chime in on 3?

jeffkpayne: Made the first two fixes, squashed the two commits into one.  Let me know if that's not preferred.  Thanks.

dhermes: @jeffkpayne Thanks for being considerate of our history in `master`. I used to do that as well, but generally it makes review harder since the old and the new become interspersed. On this PR it's irrelevant (because the PR is so small) but it's worth knowing. Also, note that GitHub has a "merge squash" button (which we use), so your need to squash is no longer a thing.

jeffkpayne: Thanks for the feedback.  Noted!

lukesneeringer:
Thanks @jeffkpayne!

jeffkpayne:
@lukesneeringer NP!  Funny, though, I was just working on a unit test to test that the calls actually happen asynchronously...  Should I continue with that and open a new PR?

jeffkpayne:
It would be along the lines of:

```
def test_on_response():
    call_times = []

    def side_effect(*args, **kwargs):
        # list.append() is thread-safe.
        call_times.append(datetime.now())
        time.sleep(2)

    callback = mock.Mock(spec=())
    callback.side_effect = side_effect

    # Then, after the calls happen, test that the times are both more than a second ago, or something like that...
```

lukesneeringer:
I would accept that as a system test. :-) And yes, additional PR is of course welcome.

lukesneeringer:
P. S. pubsub 0.28.4 is going through CI to release now. (This was a serious enough bug that I wanted it out immediately.)

dhermes:
@lukesneeringer The Pub / Sub unit tests seem to be broken (or heisen-broken) on Python 2.7:

- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3738
- https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3741

PR #4175, opened 2017-10-12
Skip user_project-based rewrite tests if envvar is not set.

tseaver:


tseaver:
> (very tough to review this!)

Neither of those tests is sensible if the environment variable is not set.

dhermes:
(sarcasm)

tseaver:
(tone deaf) :)

PR #4178, opened 2017-10-12
bigquery: retry Client.create_rows

jba:


thiago-ms:
@jba I am trying to use BigQuery to save access log, so i do many requests to BigQuery per second and simultaneously, but sometimes i'm losing data. I made an script to test my general application to see the results, i ran the script for many times and noticed that i lost some data, because sometimes the same data are being saved and sometimes not. How can i use your feature to see exactly what is happening, like to raise an exception for example?

jba:
@thiago-ms You should file an issue on this repo. I no longer maintain this code. But my first thought would be for you to modify the source to add debugging or raise an error.

tswast:
Also, be sure to check the return value of `create_rows`, as it returns a list of errors. The reason for this is that there could be partial errors (such as if only a single row of the batch did not match the schema)

PR #4182, opened 2017-10-13
bigquery: classes to support external table config

jba:
These classes will be used to configure external table
definitions for queries (transient external tables)
as well as external data configurations for tables (permanent
external tables).

This PR just establishes the classes. Subsequent PRs will
wire them into query jobs and tables.

### DIFF
--- a/bigquery/google/cloud/bigquery/external_config.py
+++ b/bigquery/google/cloud/bigquery/external_config.py
@@ -1,0 +1,485 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Define classes that describe external data sources.
+
+   These are used for both Table.externalDataConfiguration and
+   Job.configuration.query.tableDefinitions.
+"""
+
+from __future__ import absolute_import
+
+import base64
+import copy
+
+import six
+
+from google.cloud.bigquery._helpers import _to_bytes
+from google.cloud.bigquery._helpers import _bytes_to_json
+from google.cloud.bigquery._helpers import _TypedApiResourceProperty
+from google.cloud.bigquery._helpers import _ListApiResourceProperty
+from google.cloud.bigquery.schema import SchemaField
+from google.cloud.bigquery.table import _build_schema_resource
+from google.cloud.bigquery.table import _parse_schema_resource
+from google.cloud.bigquery.job import _int_or_none
+
+
+class ExternalConfig(object):
+    """Description of an external data source."""
+
+    def __init__(self):
+        self._properties = {}
+        self._options = None
+        self._schema = None
+
+    autodetect = _TypedApiResourceProperty(
+        'autodetect', 'autodetect', bool)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).autodetect
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.autodetect
+    """
+
+    compression = _TypedApiResourceProperty(
+        'compression', 'compression', six.string_types)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).compression
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.compression
+    """
+
+    ignore_unknown_values = _TypedApiResourceProperty(
+        'ignore_unknown_values', 'ignoreUnknownValues', bool)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).ignoreUnknownValues
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.ignoreUnknownValues
+    """
+
+    max_bad_records = _TypedApiResourceProperty(
+        'max_bad_records', 'maxBadRecords', six.integer_types)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).maxBadRecords
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.maxBadRecords
+    """
+
+    source_format = _TypedApiResourceProperty(
+        'source_format', 'sourceFormat', six.string_types)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).sourceFormat
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.sourceFormat
+    """
+
+    source_uris = _ListApiResourceProperty(
+        'source_uris', 'sourceUris', six.string_types)
+    """See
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+
+        :rtype: :class:`google.cloud.bigquery.external_config.CSVOptions`
+        :returns: Configuration parsed from ``resource``.
+        """
+        config = cls()
+        schema = resource.pop('schema', None)
+        if schema:
+            config.schema = _parse_schema_resource(schema)
+        for optcls in (BigtableOptions, CSVOptions, GoogleSheetsOptions):
+            opts = resource.pop(optcls._RESOURCE_NAME, None)
```
tswast: Clever!

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+
+    only_read_latest = _TypedApiResourceProperty(
+        'only_read_latest', 'onlyReadLatest', bool)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).bigtableOptions.columnFamilies.columns.only_read_latest
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.only_read_latest
+    """
+
+    qualifier_encoded = _TypedApiResourceProperty(
+        'qualifier_encoded', 'qualifierEncoded', six.binary_type)
```
tswast: Does `six.binary_type` mean that we are handling the conversion to base-64 for them?

We should clarify in the docstring whether we expect binary or base-64.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).bigtableOptions.columnFamilies.columns.qualifier_encoded
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.qualifier_encoded
+    """
+
+    qualifier_string = _TypedApiResourceProperty(
```
tswast: We could potentially do like you do for options and take care of populating the right field depending on if we are given a binary value or a unicode value.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+        'qualifier_string', 'qualifierString', six.string_types)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).bigtableOptions.columnFamilies.columns.qualifier_string
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.qualifier_string
+    """
+
+    type = _TypedApiResourceProperty('type', 'type', six.string_types)
```
tswast: Since [`type` is a Python built-in](https://docs.python.org/2/library/functions.html#type), the convention is a trailing underscore, like `type_`.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+    """
+
+    only_read_latest = _TypedApiResourceProperty(
+        'only_read_latest', 'onlyReadLatest', bool)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).bigtableOptions.columnFamilies.onlyReadLatest
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.onlyReadLatest
+    """
+
+    type = _TypedApiResourceProperty('type', 'type', six.string_types)
```
tswast: Ditto on using `type_` here.

jba: Done.

bigquery/google/cloud/bigquery/external_config.py
jba: Yes, but it's very subtle because of the different binary/text types in Python 2 vs. 3. For example, if someone passes in `str`, that will be binary in Python 2 but text in Python 3.

Probably safer to have less magic.

bigquery/google/cloud/bigquery/external_config.py
jba: Done.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+        return self._schema
+
+    @schema.setter
+    def schema(self, value):
+        if not all(isinstance(field, SchemaField) for field in value):
+            raise ValueError('Schema items must be fields')
+        self._schema = value
+
+    @property
+    def options(self):
```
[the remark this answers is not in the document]
jba: BQ is still using an HTTP/JSON discovery doc. I don't think there is anything like oneof.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,485 @@
...
+    def schema(self):
+        """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).schema
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.schema
+        """
+        return self._schema
+
+    @schema.setter
+    def schema(self, value):
+        if not all(isinstance(field, SchemaField) for field in value):
```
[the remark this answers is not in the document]
jba: I was copying from LoadJobConfig, which has an additional constraint. Changed.

jba:
In addition to addressing your comments, I made source_format a required arg of `ExternalConfig` (seemed natural to me as I started to use it), and implemented checking of source format vs. options.

jba:
I just had another idea: if you write
```
ec = ExternalConfig('BIGTABLE')
```
we could populate options with an empty `BigtableOptions`, so you could just start setting
```
ec.options.something_for_bigtable = foo
```
That way we're not setting the user up for failure by having to pick the right XXXOptions class.

WDYT?

bigquery/google/cloud/bigquery/external_config.py
tswast: We do a little bit of this kind of magic in the DB-API,

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/dbapi/_helpers.py#L49

But, I'm okay erring on the side of less magic.

tswast:
Your suggestion on populating options in the constructor makes sense to me.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,491 @@
...
+    """
+
+    max_bad_records = _TypedApiResourceProperty(
+        'max_bad_records', 'maxBadRecords', six.integer_types)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).maxBadRecords
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.maxBadRecords
+    """
+
+    source_format = _TypedApiResourceProperty(
```
tswast: Do we want to make this read-only now that we're setting it in the constructor?

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -0,0 +1,491 @@
...
+    """
+
+    schema = _ListApiResourceProperty('schema', 'schema', SchemaField)
+    """See
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).schema
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.schema
+    """
+
+    @property
+    def options(self):
```
tswast: If you do set this in the constructor, we could make this read-only, too, but keep the options mutable.

bigquery/google/cloud/bigquery/external_config.py
jba: Done.

bigquery/google/cloud/bigquery/external_config.py
jba: Will do in next PR.

PR #4188, opened 2017-10-13
BQ: Make snippets.py run as system tests.

tswast:
Converts snippet function names to pytest conventions (shouldn't affect
the docs, as the docs use region tags to grab sub-sections to include.

I had to make some minor changes to ensure that the samples work on both
Python 2 and Python 3, which is why I didn't try to bulk update the
snippets for all the other products.

### DIFF
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -76,7 +76,7 @@ def system_tests(session, python_version):
         os.path.join('..', 'storage'),
         os.path.join('..', 'test_utils'),
     )
-    session.install('.')
+    session.install('-e', '.')
 
     # Run py.test against the system tests.
     session.run(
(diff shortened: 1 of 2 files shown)


tswast:
Test failure is in Pub/Sub.

bigquery/nox.py
```diff
@@ -88,6 +88,39 @@ def system_tests(session, python_version):
...
+    session.virtualenv_dirname = 'snip-' + python_version
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install('mock', 'pytest', *LOCAL_DEPS)
+    session.install(
+        os.path.join('..', 'storage'),
+        os.path.join('..', 'test_utils'),
+    )
+    session.install('.')
```
jba: I noticed that in the system tests, we have this instead of
```
session.install('-e', '.')
```
The advantage of the latter is that you can fix your files and re-run with nox -r, so the virtualbox doesn't need to be re-installed. Is there a reason not to use -e?

tswast: You're right. I don't think there is a good reason to use install `.` instead of `-e .`. I wondered why I'd get out-of-date results with `-r`. This is probably the reason.

PR #4191, opened 2017-10-14
bigquery: support external table definitions for query jobs

jba:
Also, set ExternalConfig.options based on source_format, and
make read-only.

Also, change from_api_repr functions in external_config.py so that
they don't modify their resource argument. This simplifies tests.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -25,6 +25,7 @@ from google.cloud import exceptions
 from google.cloud.exceptions import NotFound
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud.bigquery.dataset import DatasetReference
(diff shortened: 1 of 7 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1469,6 +1463,12 @@ def from_api_repr(cls, resource):
     https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.writeDisposition
     """
 
+    table_definitions = _TypedApiResourceProperty(
+        'table_definitions', 'tableDefinitions', dict)
+    """See
```
tswast: We should document that the dictionary expects its values to be ExternalConfig objects.

jba: Done.

PR #4192, opened 2017-10-15
bigquery: make views default to standard SQL

jba:


bigquery/google/cloud/bigquery/table.py
```diff
@@ -481,9 +481,14 @@ def view_query(self, value):
         """
```
tswast: We might want to update the docstring for query to call out that the default syntax is standard SQL and/or link to `view_use_legacy_sql` since they might not think to look for it if they are used to the server default of legacy SQL.

jba: Done.

PR #4193, opened 2017-10-16
bigquery: support external data definitions for tables

jba:


jba:
The base of this PR is #4191.

bigquery/nox.py
```diff
@@ -76,7 +76,7 @@ def system_tests(session, python_version):
         os.path.join('..', 'storage'),
         os.path.join('..', 'test_utils'),
     )
-    session.install('-e', '.')
```
tswast: Looks like you undid the `session.install` change from the previous PR.

jba: Fixed.

PR #4194, opened 2017-10-16
 biqquery: factor out common values in system.py

jba:


jba:
The base of this PR is #4193.

PR #4198, opened 2017-10-16
BigQuery: option for job ID generation with user-supplied prefix

alixhami:


bigquery/google/cloud/bigquery/client.py
```diff
@@ -611,6 +611,11 @@ def load_table_from_storage(self, source_uris, destination,
         :type job_id: str
         :param job_id: (Optional) Name of the job.
 
+        :type job_id_prefix: str or ``NoneType``
```
jba: remove the first "job ID:"

"the user-provided prefix for a ..."

bigquery/google/cloud/bigquery/client.py
```diff
@@ -670,7 +680,7 @@ def load_table_from_file(self, file_obj, destination,
                  be determined, or if the ``file_obj`` can be detected to be
                  a file opened in text mode.
         """
-        job_id = _make_job_id(job_id)
+        job_id = _make_job_id(job_id, prefix=job_id_prefix)
```
jba: (nit) I don't think you need to call this with keyword-arg syntax.

bigquery/google/cloud/bigquery/client.py
alixhami: Fixed with c0a7120

bigquery/google/cloud/bigquery/client.py
alixhami: Fixed with c0a7120

PR #4199, opened 2017-10-16
Clarify 'user_project' semantics for API wrapper methods.

### LINKED ISSUE #4142: [Request] Verify User Project methods comments are not ambiguous
General Request:

Verify that the comments are clear for methods which accept User Project with and without Requester Pays.

Here's a distinction between Requester Pays and User Project:
User Project: Is a request metadata field that is used to assign Storage operational costs to a specified project.

Requester Pays: Is a bucket-level metadata field that is used to enforce the use of the User Project metadata field to assign operational costs when an operation is made on a Bucket and its objects.

tseaver:
Closes #4142.

[ci skip]

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -90,6 +90,8 @@ class _PropertyMixin(object):
     def reload(self, client=None):
         """Reload properties from Cloud Storage.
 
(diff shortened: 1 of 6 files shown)


tseaver:
Yes, it is defined on the bucket:  blob and notification instances have read-only properties which "borrow" it from the bucket.

PR #4200, opened 2017-10-16
Document that blob-level IAM APIs are deprecated.

tseaver:
Closes #4087.

[ci skip].

PR #4202, opened 2017-10-16
Prep 'storage-1.5.0' release.

tseaver:
Draft release notes:

### Features

- Add header enabling gzip downloads. (#3380)
- `Bucket.user_project` support:  APIs for such buckets and their blobs are billed to the specified project. (#4056) 
- `Bucket.requester_pays` property:  requires that operations in the bucket be made with `user_project` set. (#4056)
-  Add support for publishing notifications of bucket / blob operations to `google.cloud.pubsub` topics. (#4127)
- Validate checksums during download: when a validation error is detected, delete corrupt file. (#4133)

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-storage',
-    version='1.4.0',
+    version='1.5.0',
     description='Python Client for Google Cloud Storage',
     long_description=README,
     namespace_packages=[


dhermes:
@tseaver Can you expand on "Bucket notifications -> `google.cloud.pubsub` topics."

dhermes:
Also I think the checksum validation should be a feature. We are adding something which didn't exist before. (Removing the file was added at the same time checksum validation was enabled.)

tseaver:
Blocked on #4200.

dhermes:
@tseaver This release does not need to be blocked by #4200.

dhermes:
**Especially** considering our docs **only** appear as `latest` (i.e. there is no version of docs pinned to the Storage release).

If we were worried about blocking a release, this would be more concerning:
https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/34

tseaver:
#4200 is awaiting closure from @frankyn, who is also the driver for getting the other storage features out.

lukesneeringer:
I would like @frankyn to make that decision. I have asked him to weigh in here.

Frank, if you think this should be blocked on #4200, could you provide us with an estimated time of resolution for #4200? If you think this can go ahead, then please give @tseaver an all clear.

Thanks!

frankyn:
I reached out to the GCS team about the deprecation question and pending. I cc'd @lukesneeringer on it and will update #4200. 

As for this PR. I don't think this release should be blocked on this open question given the operations still work at this time. Continue with this release but don't include PR #4200.

PR #4203, opened 2017-10-16
Add documentation for google.api.core modules

jonparrott:


core/google/api/core/exceptions.py
```diff
@@ -332,7 +332,7 @@ def exception_class_for_http_status(status_code):
         status_code (int): The HTTP status code.
 
     Returns:
-        type: the appropriate subclass of :class:`GoogleAPICallError`.
+        :func:`type`: the appropriate subclass of :class:`GoogleAPICallError`.
```
dhermes: This doesn't make sense?

jonparrott: It's to help sphinx disambiguate between multiple targets. It considers builtin `type` to be a function.

PR #4205, opened 2017-10-16
bigquery: add filter to list_datasets

jba:
Support filtering datasets by label.


bigquery/google/cloud/bigquery/client.py
```diff
@@ -181,6 +186,10 @@ def list_datasets(self, include_all=False, max_results=None,
         extra_params = {}
         if include_all:
             extra_params['all'] = True
+        if filter:
+            # TODO: consider supporting a dict of label -> value for filter,
```
tswast: :+1: on this TODO, but probably okay to defer.

PR #4209, opened 2017-10-17
BigQuery: populate timeout parameter for getQueryResults

### LINKED ISSUE #4135: Setting a timeout for BigQuery async_query doesn't work
The documentation and code appears to allow for setting a maximum duration before timing out for an async_query by passing `timeout` (an int for the number of milliseconds) to the `result()` function that is called on a query job (https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L476 and https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L1311). Similarly, there also appear to be some references to the `fetch_data()` function of a QueryResults object being able to accept a timeout parameter (https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/query.py#L390).

However, passing a timeout value of 1 or 0 in either fashion with a async_query does nothing to interrupt the query from being completed. Queries complete regardless and output results:

```
import uuid
client = bigquery.Client(project=project)
query_job = client.run_async_query(str(uuid.uuid4()),
                                   "select 'hello' as a, 32423432 as b")
query_job.begin()
query_results = query_job.result(timeout=1)
data = query_results.fetch_data(timeout_ms=1)
rows = list(data)
print(rows)
```
Output:
```
[(u'hello', 32423432)]
```

By contrast, setting a timeout value for a sync_query by setting the `timeout_ms` property of a QueryJob indeed works as expected as raises the appropriate JobComplete = False flag:

```
client = bigquery.Client(project=project)
query_job = client.run_sync_query("select 'hello' as a, 32423432 as b")
query_job.timeout_ms=1
query_job.run()
print(query_job._properties.get("jobComplete"))
rows = list(query_job.rows)
print(rows)
```
Output:
```
False
[]
```

It appears timeout for async_queries is still to be implemented and it is unclear how to access a JobIncomplete property via an async_query, forcing users to continue using sync_queries if they want the ability to time out their queries.

cc: @tswast who requested I raise the issue which I identified in this pandas-gbq PR: https://github.com/pydata/pandas-gbq/pull/25#discussion_r143263462

OSX
Python 2.7.13
google-cloud-python 0.27.0 and google-cloud-bigquery 0.26.0

tswast:
This will allow QueryJob to respect the timeout value for futures.

Closes #4135.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -1522,6 +1522,7 @@ class QueryJob(_AsyncJob):
         self.query = query
         self._configuration = job_config
         self._query_results = None
+        self._done_timeout = None
 
     @property
     def allow_large_results(self):
@@ -1871,7 +1872,7 @@ class QueryJob(_AsyncJob):
         """
         if not self._query_results:
             self._query_results = self._client._get_query_results(
-                self.job_id, retry)
+                self.job_id, retry, project=self.project)
         return self._query_results
 
     def done(self, retry=DEFAULT_RETRY):
@@ -1880,11 +1881,25 @@ class QueryJob(_AsyncJob):
         :rtype: bool
         :returns: True if the job is complete, False otherwise.
         """
+        # Since the API to getQueryResults can hang up to the timeout value
+        # (default of 10 seconds), set the timeout parameter to ensure that
+        # the timeout from the futures API is respected. See:
+        # https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4135
+        timeout_ms = None
+        if self._done_timeout is not None:
+            # Subtract a buffer of a 100 milliseconds for context switching,
+            # network latency, etc.
+            timeout_ms = int(1000 * self._done_timeout) - 100
+            timeout_ms = max(min(timeout_ms, 10000), 0)
+            self._done_timeout -= (timeout_ms / 1000.0)
+            self._done_timeout = max(0, self._done_timeout)
+
         # Do not refresh is the state is already done, as the job will not
         # change once complete.
         if self.state != _DONE_STATE:
             self._query_results = self._client._get_query_results(
-                self.job_id, retry)
+                self.job_id, retry,
+                project=self.project, timeout_ms=timeout_ms)
 
             # Only reload the job once we know the query is complete.
             # This will ensure that fields such as the destination table are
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1880,11 +1881,25 @@ def done(self, retry=DEFAULT_RETRY):
...
         """
+        # Since the API to getQueryResults can hang up to the timeout value
+        # (default of 10 seconds), set the timeout parameter to ensure that
+        # the timeout from the futures API is respected. See:
+        # https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4135
+        timeout_ms = None
+        if self._done_timeout is not None:
+            # Subtract a buffer of a 100 milliseconds for context switching,
+            # network latency, etc.
+            timeout_ms = int(1000 * self._done_timeout) - 100
```
jonparrott: why not just convert to ms at the end?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1880,11 +1881,25 @@ def done(self, retry=DEFAULT_RETRY):
...
+            timeout_ms = max(min(timeout_ms, 10000), 0)
+            self._done_timeout -= (timeout_ms / 1000.0)
+            self._done_timeout = max(0, self._done_timeout)
+
         # Do not refresh is the state is already done, as the job will not
         # change once complete.
         if self.state != _DONE_STATE:
             self._query_results = self._client._get_query_results(
-                self.job_id, retry)
+                self.job_id, retry,
```
jonparrott: I don't see anything here that deals with the inevitable timeouterror. Also, what if the timeouterror isn't due to the job not being done but is instead due to network issues?

tswast: Sorry, I'm not sure to do if I get a timeout error. Return False?

Does the `reload()` call have the same problem (also used in the shared superclass)?

bigquery/google/cloud/bigquery/job.py
tswast: Moved to the end.

tswast:
Circle error is with Pub/Sub. cloud.google.bigquery tests are passing.

Discussed timeout exception in person. The timeout option for getQueryResults is a parameter, not enforced client-side. When the time expires, the API will return a successful response with the incomplete job state.

bigquery/google/cloud/bigquery/job.py
tswast: For future reference, we discussed in person. The timeout variable in this instance is being used as a parameter to the API request, not in the sense that the futures API uses timeout.

PR #4210, opened 2017-10-17
Add api_core package

jonparrott:
Note:

* Added new module `protobuf_helpers` to break co-dependency on `core`. (This was going to be added anyway as I move helpers from gax)
* This isn't in the docs yet but I'll do that a a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,12 @@ jobs:
             python3.6 test_utils/scripts/get_target_packages.py > ~/target_packages
             cat ~/target_packages
       - run:
+          name: Run tests - google.api.core
+          command: |
+            if [[ -n $(grep api_core ~/target_packages) ]]; then
+              nox -f api_core/nox.py
+            fi
+      - run:
           name: Run tests - google.cloud.core
           command: |
             if [[ -n $(grep core ~/target_packages) ]]; then
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -1,0 +1,14 @@
+[run]
+branch = True
+
+[report]
+fail_under = 100
+show_missing = True
+exclude_lines =
+    # Re-enable the standard pragma
+    pragma: NO COVER
+    # Ignore debug-only repr
+    def __repr__
+    # Ignore abstract methods
+    raise NotImplementedError
+    raise NotImplementedError()
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -11,4 +11,3 @@ exclude_lines =
     def __repr__
     # Ignore abstract methods
     raise NotImplementedError
-    raise NotImplementedError()
--- a/api_core/google/api_core/exceptions.py
+++ b/api_core/google/api_core/exceptions.py
@@ -1,0 +1,443 @@
+# Copyright 2014 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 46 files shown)


.circleci/config.yml
```diff
@@ -41,6 +41,12 @@ jobs:
             python3.6 test_utils/scripts/get_target_packages.py > ~/target_packages
             cat ~/target_packages
       - run:
+          name: Run tests - google.api.core
+          command: |
+            if [[ -n $(grep api_core ~/target_packages) ]]; then
+              nox -f api_core/nox.py
+            fi
```
dhermes: Thanks for reminding me to file https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4211

api_core/setup.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2016 Google Inc.
```
dhermes: 2017?

api_core/setup.py
```diff
@@ -0,0 +1,76 @@
...
+        'Programming Language :: Python :: 3.6',
+        'Topic :: Internet',
+    ],
+}
+
+
+REQUIREMENTS = [
+    'googleapis-common-protos >= 1.5.3, < 2.0dev',
+    'protobuf >= 3.0.0',
+    'google-auth >= 0.4.0, < 2.0.0dev',
```
dhermes: Intentional to cover `0.x` and `1.x`?

api_core/setup.py
jonparrott: Since it originated in 2016, it stays.

api_core/setup.py
jonparrott: Same as core, it should be fine as-is.

api_core/.coveragerc
```diff
@@ -0,0 +1,14 @@
...
+fail_under = 100
+show_missing = True
+exclude_lines =
+    # Re-enable the standard pragma
+    pragma: NO COVER
+    # Ignore debug-only repr
+    def __repr__
+    # Ignore abstract methods
+    raise NotImplementedError
+    raise NotImplementedError()
```
lukesneeringer: This is line redundant, is it not? Otherwise line 11 would not work.

api_core/google/api_core/gapic_v1/method.py
```diff
@@ -0,0 +1,292 @@
...
+    """Transforms metadata to gRPC format and adds global metrics.
+
+    Args:
+        metadata (Mapping[str, str]): Any current metadata.
+
+    Returns:
+        Sequence[Tuple(str, str)]: The gRPC-friendly metadata keys and values.
+    """
+    client_metadata = 'api-core/{} gl-python/{} grpc/{}'.format(
+        _API_CORE_VERSION, _PY_VERSION, _API_CORE_VERSION)
```
lukesneeringer: This seems wrong; the third argument should be the gRPC version, yes?

jonparrott: Gooooood catch! Done.

api_core/google/api_core/exceptions.py
```diff
@@ -0,0 +1,443 @@
...
+    @property
+    def cause(self):
+        """The last exception raised when retrying the function."""
+        return self._cause
+
+    def __str__(self):
+        return '{}, last exception: {}'.format(self.message, self.cause)
+
+
+class _GoogleAPICallErrorMeta(type):
```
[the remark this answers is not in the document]
jonparrott: I can always change it later as it's an implementation detail. I'm fine with this particular use of a metaclass.

api_core/.coveragerc
jonparrott: So it is, removed.

PR #4213, opened 2017-10-18
BigQuery: removes LoadJob error for autodetect=True and schema set

alixhami:
Validations in the `LoadJob` class were set to disallow a schema to be set while autodetect is True. This raised an error when running `client.get_job()` and `client.list_jobs()` for LoadJobs that were created with `autodetect=True` because the jobs came back with the schema having been set. The additional code added in system tests shows how to reproduce the bug.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -143,9 +143,6 @@ class AutoDetectSchema(_TypedApiResourceProperty):
     """
     def __set__(self, instance, value):
         self._validate(value)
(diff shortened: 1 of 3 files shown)


tswast:
You're right that it could be possible to bypass the check on load. It would be to change

```
config.schema = _parse_schema_resource(schema)
```

to

```
config._schema = _parse_schema_resource(schema)
```

I think the server-side error message is pretty clear if someone does try to set a schema and use autodetect. I'm okay removing the client-side check.

PR #4220, opened 2017-10-18
Update docs to use api_core instead of api.core

jonparrott:


nox.py
```diff
@@ -54,8 +54,7 @@ def docs(session):
 
     # Install Sphinx and also all of the google-cloud-* packages.
     session.chdir(os.path.realpath(os.path.dirname(__file__)))
-    session.install('Sphinx >= 1.6.3', 'sphinx_rtd_theme')
-    session.install(*ALL_PACKAGES)
+    session.install('-r', os.path.join('docs', 'requirements.txt'))
```
dhermes: @jonparrott ISTM you can get rid of `ALL_PACKAGES` then?

jonparrott: Done.

PR #4221, opened 2017-10-18
Replace usage of google.api.core with google.api_core

jonparrott:
This is the big one- this replaces any usage of `api.core` with `api_core` and makes `google-cloud-core` depend on `google-api-core`.


.github/CODEOWNERS
```diff
@@ -10,10 +10,8 @@
 core/*                    @dhermes
 test_utils/*              @dhermes
 
-# Jon Wayne Parrott is the author of the new google.api.core library.
-# NOTE: This trumps the previous core/* directive, since the last match is
-#       what matters.
-core/google/api/core/*    @jonparrott
+# Jon Wayne Parrott is the author of the new google.api_core library.
```
dhermes: Do we want to call it "new" in here? I assume this doc will exist long after it isn't new.

jonparrott: I expect this doc will cease to exist when we split the repo.

dhermes: DERP

jonparrott:
Good call on updating get_target_packages, done.

PR #4225, opened 2017-10-19
Add google.api_core.gapic_v2.client_info

### LINKED ISSUE #4068: Ensure that gRPC metrics headers are in the order requested by the spec.
The present order in `api.core` is:

```
core/0.27.1 gl-python/3.6.0 grpc/0.27.1 gapic/0.27.2
```

The order per spec should be:

```
gl-python/2.7.12 gccl/0.21.0 gapic/0.1.0 core/0.14.1 grpc/1.0.1
```

jonparrott:
Resolves #4068


jonparrott:
@lukesneeringer @dhermes this is my attempt to fix #4068 without making myself angry. I'm open to feedback and suggestions at alternatives.

api_core/google/api_core/gapic_v1/client_info.py
```diff
@@ -0,0 +1,82 @@
...
+    """Client information used to generate a user-agent for API calls.
+
+    This user-agent information is sent along with API calls to allow the
+    receiving service to do analytics on which versions of Python and Google
+    libraries are being used.
+
+    Args:
+        python_version (str): The Python interpreter version, for example,
+            ``'2.7.13'``.
+        grpc_version (st): The gRPC library version.
```
lukesneeringer: `str`

api_core/google/api_core/gapic_v1/client_info.py
```diff
@@ -0,0 +1,82 @@
...
+        api_core_version (str): The google-api-core library version.
+        gapic_version (Optional[str]): The version of gapic used to generate
```
lukesneeringer: This description is wrong. It is the version of the auto-gen client library itself. The distinction between this and `client_library_version` is that we will only specify the latter if manual code is involved.

jonparrott: Done

api_core/google/api_core/gapic_v1/client_info.py
jonparrott: Done.

PR #4232, opened 2017-10-20
[DO NOT MERGE] Demonstrate changes needed to move video gapic to api_core

jonparrott:
* This change will pass unit tests once #4231 is in and merged.
* LRO is broken right now because the old protos barf when timeout and metadata arguments are passed to RPC methods. Re-generating the copy of `google.longrunning` in `googleapis-common-protos` will fix this (alternatively we could place our own copy in `google.api_core.operations_v`).

### DIFF
--- a/videointelligence/setup.py
+++ b/videointelligence/setup.py
@@ -39,7 +39,9 @@ setup(
     packages=find_packages(exclude=('tests*',)),
     install_requires=(
         'googleapis-common-protos >= 1.5.3, < 2.0dev',
+        # GAX remains for the time being for the v1beta1 interface.
         'google-gax >= 0.15.14, < 0.16dev',
+        'google-api-core >= 0.1.0, < 0.2.0dev',
         'grpcio >= 1.2.0, < 1.6dev',
         'six >= 1.10.0',
     ),
--- a/videointelligence/tests/unit/gapic/v1beta2/test_video_intelligence_service_client_v1beta2.py
+++ b/videointelligence/tests/unit/gapic/v1beta2/test_video_intelligence_service_client_v1beta2.py
@@ -11,12 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Unit tests."""
 
-import mock
-import unittest
-
-from google.gax import errors
 from google.rpc import status_pb2
 
 from google.cloud import videointelligence_v1beta2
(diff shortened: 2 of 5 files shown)


lukesneeringer:
@landrito This is a sample of what will need to be done in the autogen.

videointelligence/setup.py
```diff
@@ -39,7 +39,9 @@
     packages=find_packages(exclude=('tests*',)),
     install_requires=(
         'googleapis-common-protos >= 1.5.3, < 2.0dev',
+        # GAX remains for the time being for the v1beta1 interface.
```
lukesneeringer: We can (and probably should) regen the v1beta1 interface.

jonparrott: It *does* drop `options` in favor of `retry` and `timeout` on every method and changes the constructor significantly.

jonparrott:
@lukesneeringer hmm using your "updated" longrunning doesn't fix the issue.

Summoning @nathanielmanistaatgoogle, do you know why specifying `metadata` or `timeout` to an RPC method would yield this:

```
  File "/Users/jonwayne/workspace/google-cloud-videointelligence/api_core/google/api_core/grpc_helpers.py", line 51, in error_remapped_callable
    return callable_(*args, **kwargs)
  File "/Users/jonwayne/workspace/new-rpc-tools/env/lib/python3.6/site-packages/grpc/_channel.py", line 507, in __call__
    return _end_unary_response_blocking(state, call, False, deadline)
  File "/Users/jonwayne/workspace/new-rpc-tools/env/lib/python3.6/site-packages/grpc/_channel.py", line 455, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, Request contains an invalid argument.)>
```

videointelligence/tests/unit/gapic/v1beta2/test_video_intelligence_service_client_v1beta2.py
```diff
@@ -11,67 +11,71 @@
...
 from google.rpc import status_pb2
 
 from google.cloud import videointelligence_v1beta2
 from google.cloud.videointelligence_v1beta2.proto import video_intelligence_pb2
 from google.longrunning import operations_pb2
 
 
-class CustomException(Exception):
-    pass
+class UnaryUnaryMultiCallableStub(object):
```
jonparrott: @lukesneeringer @dhermes it may make sense to put a full, fleshed out gRPC channel stub (with method stubs) in `api_core` for the benefit of testing all gapics and people who want to test their own gRPC client. WDYT?

nathanielmanistaatgoogle:
@jonparrott: no, not offhand I can't think of any reason why that would happen. What are the values you're passing for those parameters? And where are you passing them? I went looking for them; should I have seen them on the right side of this pull request's diff?

jonparrott:
@nathanielmanistaatgoogle I'd seen this before with older generated code, but I hadn't seen it since (and luke regenerated the operations_pb2 stuff for me to test with).

The parameters are applied by the api_core decorators. Timeout will be a float, metadata is just a list of tuples (as expected). The parameters work fine on the videointelligence stub but not on the operations stub.

jonparrott:
Hrm this could be something else. A minimal case to reproduce works fine even with the older version of operations_pb2.

jonparrott:
Still not sure what's going on:

```
Calling <grpc._channel._UnaryUnaryMultiCallable object at 0x10559a9e8> (input_uri: "gs://demomaker/cat.mp4"
features: LABEL_DETECTION
,) {'metadata': [('x-goog-api-client', 'gl-python/3.6.0 gapic/0.27.2 gax/0.1.0 grpc/1.4.0')], 'timeout': 120.0}
Calling <grpc._channel._UnaryUnaryMultiCallable object at 0x1055ae828> (name: "us-west1.3363818886345698022"
,) {'metadata': [('x-goog-api-client', 'gl-python/3.6.0 gax/0.1.0 grpc/1.4.0')], 'timeout': 20.0}
  File "/Users/jonwayne/workspace/google-cloud-videointelligence/api_core/google/api_core/grpc_helpers.py", line 52, in error_remapped_callable
    return callable_(*args, **kwargs)
  File "/Users/jonwayne/workspace/new-rpc-tools/env/lib/python3.6/site-packages/grpc/_channel.py", line 507, in __call__
    return _end_unary_response_blocking(state, call, False, deadline)
  File "/Users/jonwayne/workspace/new-rpc-tools/env/lib/python3.6/site-packages/grpc/_channel.py", line 455, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, Request contains an invalid argument.)>
```

The first call to the `VideointelligenceStub` passes in a timeout and succeeds, the second call to the `OperationsStub` (using the same channel) raises the exception.

If I remove timeout from the core library everything works fine.

Also, if I manually do this I get no issues:

```python
import google.auth
import google.auth.transport.requests
import google.auth.transport.grpc
from google.longrunning import operations_pb2

import google.cloud.videointelligence_v1beta2


credentials, _ = google.auth.default(
    scopes=['https://www.googleapis.com/auth/cloud-platform'])

channel = google.auth.transport.grpc.secure_authorized_channel(
    credentials,
    google.auth.transport.requests.Request(),
    'videointelligence.googleapis.com:443')


operations_stub = operations_pb2.OperationsStub(channel)

response = operations_stub.GetOperation(
    operations_pb2.GetOperationRequest(name='us-west1.12515936760487769527'),
    timeout=120.0,
    metadata=[('x-goog-api-client', 'gl-python/3.6.0 gapic/0.27.2 gax/0.1.0 grpc/1.4.0')])

print(response)
```

jonparrott:
Hah, okay, wat? If I specify a timeout of `20.0` in the minimal case I get an invalid argument error?

jonparrott:
Okay it seems like `gRPC` barfs if `timeout` is less than `46.0`. Any idea what's going on with that, @nathanielmanistaatgoogle?

jonparrott:
@landrito pease note f5ca60c7

landrito:
Ack.

nathanielmanistaatgoogle:
@jonparrott: no idea; that is profoundly insane and weird. I look forward to hearing more details about it.

jonparrott:
Sent an internal email. The timeout thing is a non-issue for this, but is
an issue specific to the video API.

PR #4235, opened 2017-10-20
BigQuery: Renames client.load_table_from_storage() to client.load_table_from_uri()

alixhami:
Renamed so we can add functionality to use URIs that point to [Bigtable instances](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.sourceUris)


tswast:
Circle is failing on Pub/Sub not BigQuery, so I'll merge.

PR #4237, opened 2017-10-20
BigQuery: Updates snippets for BigQuery Beta 2 changes

alixhami:


tswast:
Thanks! Reviewing now.

Looks like the current failures are for `google-cloud-storage` which don't appear to be caused by this PR.

docs/bigquery/snippets.py
```diff
@@ -89,476 +97,556 @@ def do_something_with(_):
...
-    # [START dataset_create]
-    dataset = client.dataset(DATASET_NAME)
-    dataset.create()              # API request
-    # [END dataset_create]
-
-    to_delete.append(dataset)
+    DATASET_ID = 'create_dataset_%d' % (_millis(),)
 
+    # [START create_dataset]
+    dataset_ref = client.dataset(DATASET_ID)
```
tswast: We might want to put a comment before `dataset_ref` line and after `START` line like

    # DATASET_ID = 'dataset_ids_are_strings'

docs/bigquery/snippets.py
```diff
@@ -89,476 +97,556 @@ def do_something_with(_):
...
-    dataset.patch(
-        description=PATCHED_DESCRIPTION,
-        default_table_expiration_ms=ONE_DAY_MS
-    )      # API request
-    assert dataset.description == PATCHED_DESCRIPTION
-    assert dataset.default_table_expiration_ms == ONE_DAY_MS
-    # [END dataset_patch]
+    dataset.description = UPDATED_DESCRIPTION
+
+    client.update_dataset(dataset, ['description'])  # API request
```
tswast: This line should be `dataset = client.update_dataset(...)` because we want the user to get the updated `etag` value. `update_dataset` does not (or at least should not) mutate the arguments it is given. (Plus the assert below will make more sense.)

docs/bigquery/snippets.py
```diff
@@ -89,476 +97,556 @@ def do_something_with(_):
...
-    table.patch(
-        friendly_name=PATCHED_FRIENDLY_NAME,
-        description=PATCHED_DESCRIPTION,
-    )      # API request
-    assert table.friendly_name == PATCHED_FRIENDLY_NAME
-    assert table.description == PATCHED_DESCRIPTION
-    # [END table_patch]
+    table.description = UPDATED_DESCRIPTION
 
+    client.update_table(table, ['description'])  # API request
```
tswast: This should be `table = client.update_table(...)`

docs/bigquery/snippets.py
```diff
@@ -89,476 +97,556 @@ def do_something_with(_):
...
+
     assert table.friendly_name == UPDATED_FRIENDLY_NAME
     assert table.description == UPDATED_DESCRIPTION
     assert table.schema == NEW_SCHEMA
-    # [END table_update]
+    # [END update_table_multiple_properties]
 
 
-def _warm_up_inserted_table_data(table):
+def _warm_up_inserted_table_data(client, table):
```
tswast: I believe this function is no longer used by the snippets and can be deleted.

docs/bigquery/snippets.py
alixhami: Added in e30c19a

docs/bigquery/snippets.py
alixhami: Fixed in e30c19a

docs/bigquery/snippets.py
alixhami: Fixed in e30c19a

docs/bigquery/snippets.py
alixhami: Removed in e30c19a

tswast:
Now the test error is on `core`. Seems like it might be caused by an update to `flake8`.

PR #4240, opened 2017-10-23
BigQuery: populate exceptions for query_rows with context about the job.

tswast:
This enables one to catch errors and get the job to wait for it or
cancel it.

Since a full job resource is not required to cancel the job, only the
ID, I also provide a Client.cancel_job method.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import collections
+import concurrent.futures
 import functools
 import os
 import uuid
@@ -28,6 +29,7 @@ from google.resumable_media.requests import MultipartUpload
 from google.resumable_media.requests import ResumableUpload
 
 from google.api.core import page_iterator
+from google.api.core.exceptions import GoogleAPICallError
 
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1129,13 +1178,33 @@ def query_rows(self, query, job_config=None, job_id=None, timeout=None,
...
-            query, job_config=job_config, job_id=job_id, retry=retry)
-        return job.result(timeout=timeout)
+        job_id = _make_job_id(job_id, job_id_prefix)
+
+        try:
+            job = self.query(
+                query, job_config=job_config, job_id=job_id, retry=retry)
+            rows_iterator = job.result(timeout=timeout)
+        except (GoogleAPICallError, concurrent.futures.TimeoutError) as exc:
+            exc.project = self.project
```
jonparrott: Can you tell me why this information is needed?

tswast: If a call to `query_rows()` fails and someone wants to investigate further, they will need the full job ID (including the project). This is the first information that support or BigQuery engineers ask for.

If a call to query_rows() times out, the query is still running in the background. In this case, the user will either want to get more information about the job or cancel it.

jonparrott: So I'm kind of -1 on this.

1. It adds new properties to an object, which is generally frowned upon. There's no docstrings for this, there's no declaration of the interface. Finally, if we were to add `__slots___` to our exception classes, this would fail outright.
2. It doesn't cause any behavioral change in the exceptions. It adds the info, sure, but it doesn't print this info out in `str`.

@dhermes what do you think?

tswast: Other options:

- Make new exception types to wrap this just for `query_rows()`? That seems even yuckier than adding properties.
- Change the behavior to call cancel on the job in case of error or timeout. (Noting in the docs that cancel is best effort.)

Calling `cancel` is less flexible. Then again, someone who is using `query_rows()` instead of `query()`is actually asking for less flexibility. It also does not address that fact that they don't have the job ID, but they could potentially call `list_jobs` to figure it out.

tswast: I'm going with "Change the behavior to call cancel on the job in case of error or timeout. (Noting in the docs that cancel is best effort.)" Will open a second PR as bigquery-b2 is now merged to master.

PR #4244, opened 2017-10-23
Updating some storage mocks after `google-resumable-media==0.3.1` release.

### LINKED ISSUE #4243: GCS test failures for 'headers'
Was: https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/41

dhermes:
Fixes #4243.

Also increasing the lower-bound on `google-resumable-media` to `0.3.1`.

### DIFF
--- a/storage/tests/unit/test_blob.py
+++ b/storage/tests/unit/test_blob.py
@@ -441,7 +441,9 @@ class Test_Blob(unittest.TestCase):
         response.status_code = status_code
         response.headers.update(headers)
         if stream:
-            response.raw = io.BytesIO(content)
+            raw = io.BytesIO(content)
+            raw.headers = headers
+            response.raw = raw
             response._content = False
         else:
             response.raw = None
@@ -677,15 +679,18 @@ class Test_Blob(unittest.TestCase):
         empty_hash = base64.b64encode(
             hashlib.md5(b'').digest()).decode(u'utf-8')
         headers = {'x-goog-hash': 'md5=' + empty_hash}
+        mock_raw = mock.Mock(headers=headers, spec=['headers'])
         response = mock.MagicMock(
             headers=headers,
             status_code=http_client.OK,
+            raw=mock_raw,
             spec=[
                 '__enter__',
                 '__exit__',
                 'headers',
                 'iter_content',
                 'status_code',
+                'raw',
             ],
         )
         # i.e. context manager returns ``self``.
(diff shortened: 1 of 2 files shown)


dhermes:
Can I get a review @jonparrott @tseaver @lukesneeringer? Our tests are broken right meow (e.g. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/3882)

storage/tests/unit/test_blob.py
```diff
@@ -441,7 +441,9 @@ def _mock_requests_response(
         response.status_code = status_code
         response.headers.update(headers)
         if stream:
-            response.raw = io.BytesIO(content)
+            raw = io.BytesIO(content)
+            raw.headers = headers
+            response.raw = raw
```
tseaver: Why is `google-resumable-media` looking for headers on the response's stream, rather than the response itself?

dhermes: @tseaver Good point. Let's discuss, but the discussion shouldn't block the PR from un-breaking tests for everyone? (Will follow up with another comment.)

tseaver:
I'm OK with merging now to unblock others, but I'm worried that we are papering over the real bug (either here or in `google-resumable-media`).  So we should either *not* close #4243 (but maybe edit its title / description), or else reopen https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/41.

storage/tests/unit/test_blob.py
dhermes: The "why?" is "because that's what `urllib3` [does][1]". The [change][2] that added it actually monkey-patches the `gzip` decoder attached to `response.raw`, so I used the same logic as `urllib3` uses in `urllib3.response.HTTPResponse._init_decoder`.

But, as you pointed out, `urrlib3` **can't** assume it's being used along-side / used by `requests` but we **can** assume that, so we **could** do the `content-encoding` header check from `requests.models.Response.headers`.

[1]: https://github.com/shazow/urllib3/blob/3a74b5b7862a1b55c3a059e9621aea4322e786f8/urllib3/response.py#L250-L258
[2]: https://github.com/GoogleCloudPlatform/google-resumable-media-python/pull/36

dhermes:
@tseaver There is no bug. This is working as "intended", but the intent can be discussed (see my comment https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4244#discussion_r146901312)

PR #4245, opened 2017-10-24
BigQuery Beta 2 Changes

tswast:
This pull request implements requirements for the beta 2 launch as well as a redesign of the API surface. Changes include:

- Query and view operations default to the Standard SQL dialect.
- Client functions related to [jobs](https://cloud.google.com/bigquery/docs/jobs-overview) like running queries immediately start the job.
  - Use `QueryJobConfig` and analogous classes to configure optional properties about the job before starting it with the client methods.
  - Use the `concurrent.futures` API (e.g. `result()`) on jobs to wait for jobs to complete. In the case of `QueryJob`, the `result()` method returns an iterator over the rows of the destination table.
- Added `TableReference` and `DatasetReference` classes to be more clear about when an object does not contain additional properties of the resource.
- Functions to create/get/update/delete datasets and tables moved from `Table` and `Dataset` to the `Client` class.
- Use `Client.create_rows()` instead of `Table.insert_data()` to [stream rows to a table](https://cloud.google.com/bigquery/streaming-data-into-bigquery).
- Use `Client.list_rows()` instead of `Table.fetch_data()`.
- Row iterators allow access to column values by keyword or attribute, in addition to by integer index.

/cc @alixhami @jba


tseaver:
@tswast We'd have to, unless one of us does the merge at the command line (non-squash PR merge is disabled for the project).

lukesneeringer:
> Question: When we merge this are we going to squash it?

I can enable non-squash merge temporarily if you want.

tseaver:
@lukesneeringer 

> I can enable non-squash merge temporarily if you want.

ISTM we could just do it at the command line:

```bash
$ git checkout master && git fetch --all --prune && git merge upstream master
$ git merge bigquery-b2 
$ git push upstream master
```

PR #4252, opened 2017-10-25
Bigtable ReadRows doesn't handle empty string column qualifier

garye:
This problem was reported by a customer. As background, the empty string is a valid column qualifier. A missing qualifier, as opposed to the empty string, means the previously seen qualifier should be applied instead.

I added a new test case that has a cell with a column qualifier of empty string arriving after a cell with a non-empty qualifier. The code fails to distinguish between a missing qualifier and an empty qualifier and overwrites the empty string with the first qualifier.

With no code changes the new test case fails. I attempted to fix the bug in `row_data.py` but this causes other failures. I was hoping someone could pick this up and investigate.

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -283,10 +283,14 @@ class PartialRowsData(object):
                 row = self._row = PartialRowData(chunk.row_key)
 
             if cell is None:
+                qualifier = None
+                if chunk.HasField('qualifier'):
+                    qualifier = chunk.qualifier.value
+
                 cell = self._cell = PartialCellData(
                     chunk.row_key,
                     chunk.family_name.value,
-                    chunk.qualifier.value,
+                    qualifier,
                     chunk.timestamp_micros,
                     chunk.labels,
                     chunk.value)
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -114,7 +114,7 @@ class PartialRowData(object):
     """Representation of partial row in a Google Cloud Bigtable Table.
 
     These are expected to be updated directly from a
-   :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
+    :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
 
     :type row_key: bytes
     :param row_key: The key for the row holding the (partial) data.
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -421,7 +421,8 @@ class PartialRowsData(object):
                 cell.row_key = previous.row_key
             if not cell.family_name:
                 cell.family_name = previous.family_name
-            if cell.qualifier is None: # Note: qualifier can be empty string
+            # NOTE: ``cell.qualifier`` **can** be empty string.
+            if cell.qualifier is None:
                 cell.qualifier = previous.qualifier
 
     def _save_current_row(self):
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -114,7 +114,7 @@ class PartialRowData(object):
     """Representation of partial row in a Google Cloud Bigtable Table.
 
     These are expected to be updated directly from a
-    :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
+   :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
 
     :type row_key: bytes
     :param row_key: The key for the row holding the (partial) data.
@@ -421,7 +421,7 @@ class PartialRowsData(object):
                 cell.row_key = previous.row_key
             if not cell.family_name:
                 cell.family_name = previous.family_name
-            if not cell.qualifier:
+            if cell.qualifier is None: # Note: qualifier can be empty string
                 cell.qualifier = previous.qualifier
 
     def _save_current_row(self):
--- a/bigtable/tests/unit/test_row_data.py
+++ b/bigtable/tests/unit/test_row_data.py
@@ -334,7 +334,7 @@ class TestPartialRowsData(unittest.TestCase):
         prd._copy_from_previous(cell)
         self.assertEqual(cell.row_key, '')
         self.assertEqual(cell.family_name, u'')
-        self.assertEqual(cell.qualifier, None)
+        self.assertIsNone(cell.qualifier)
         self.assertEqual(cell.timestamp_micros, 0)
         self.assertEqual(cell.labels, [])
 
@@ -530,7 +530,6 @@ class TestPartialRowsData_JSON_acceptance_tests(unittest.TestCase):
     def _match_results(self, testcase_name, expected_result=_marker):
         chunks, results = self._load_json_test(testcase_name)
         response = _ReadRowsResponseV2(chunks)
-
         iterator = _MockCancellableIterator(response)
         prd = self._make_one(iterator)
         prd.consume_next()
--- a/bigtable/tests/unit/test_row_data.py
+++ b/bigtable/tests/unit/test_row_data.py
@@ -334,7 +334,7 @@ class TestPartialRowsData(unittest.TestCase):
         prd._copy_from_previous(cell)
         self.assertEqual(cell.row_key, '')
         self.assertEqual(cell.family_name, u'')
-        self.assertEqual(cell.qualifier, b'')
+        self.assertEqual(cell.qualifier, None)
         self.assertEqual(cell.timestamp_micros, 0)
         self.assertEqual(cell.labels, [])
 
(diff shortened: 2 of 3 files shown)


bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -421,7 +421,7 @@ def _copy_from_previous(self, cell):
                 cell.row_key = previous.row_key
             if not cell.family_name:
                 cell.family_name = previous.family_name
-            if not cell.qualifier:
+            if cell.qualifier is None: # Note: qualifier can be empty string
```
garye: This line fixes the new test but causes failures in others such as `TestPartialRowsData_JSON_acceptance_tests.test_two_unsplit_cells` and `TestPartialRowsData_JSON_acceptance_tests.test_with_labels`.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -114,7 +114,7 @@ class PartialRowData(object):
     """Representation of partial row in a Google Cloud Bigtable Table.
 
     These are expected to be updated directly from a
-    :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
+   :class:`._generated.bigtable_service_messages_pb2.ReadRowsResponse`
```
dhermes: This looks like an accidental character deletion?

bigtable/google/cloud/bigtable/row_data.py
dhermes: Yes the famous [`None`-check][1]!

When would `cell.qualifier` (it's not actually a `google.cloud.bigtable.row_data.Cell`, it's a `google.cloud.bigtable.row_data.PartialCellData` believe) be set to `None`?

bigtable/tests/unit/test_row_data.py
```diff
@@ -530,6 +530,7 @@ def test_invalid_last_row_missing_commit(self):
     def _match_results(self, testcase_name, expected_result=_marker):
         chunks, results = self._load_json_test(testcase_name)
         response = _ReadRowsResponseV2(chunks)
+
```
dhermes: Whitespace just for fun?

bigtable/tests/unit/test_row_data.py
```diff
@@ -334,7 +334,7 @@ def test__copy_from_previous_unset(self):
         prd._copy_from_previous(cell)
         self.assertEqual(cell.row_key, '')
         self.assertEqual(cell.family_name, u'')
-        self.assertEqual(cell.qualifier, b'')
+        self.assertEqual(cell.qualifier, None)
```
dhermes: `assertIsNone` is preferred?

bigtable/tests/unit/test_row_data.py
```diff
@@ -678,7 +681,7 @@ class _PartialCellData(object):
 
     row_key = ''
     family_name = u''
-    qualifier = b''
+    qualifier = None
```
dhermes: This is to match [`PartialCellData`][1], but that class has no defaults (i.e. `None` isn't the default).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6dbb16c02ba4cdf4624a23b88167465133cb45f1/bigtable/google/cloud/bigtable/row_data.py#L95-L102

dhermes:
@garye I think I get it, since [`ReadRowsResponse.CellChunk.qualifier`][1] is a `google.protobuf.BytesValue`, it's a message field rather than a scalar field, so `None` is an acceptable default. (I'm still trying to figure out where a `None` value gets passed into the `PartialCellData` constructor.)

[1]: https://github.com/googleapis/googleapis/blob/85784381208629031174a992db3f20ce08d61c6c/google/bigtable/v2/bigtable.proto#L121

dhermes:
@garye I just noticed CircleCI is trying to run tests for **all** the packages for you. This is probably because your branch isn't pointing at the latest `HEAD` in this repo. Mind if I rebase for you? While I'm doing it I can send in the few cosmetic changes I requested.

bigtable/google/cloud/bigtable/row_data.py
garye: As the server sends chunks of a row, it only sends the row key/column family/column qualifier for a new cell if it is different than the last cell (as an optimization). So if you receive the same column at two different timestamps the second cell won't have a qualifier set and will fall back to the previously read one.

bigtable/google/cloud/bigtable/row_data.py
garye: yes, looks that way

garye:
Sounds great @dhermes, please rebase when you can. Thanks for the help.

bigtable/google/cloud/bigtable/row_data.py
dhermes: @garye Thanks for the info, I meant where in this codebase though :grinning:

garye: haha, I'm afraid I don't have a great answer for that!

dhermes: So I'm pretty sure [this][1] is the only place:

```python
cell = self._cell = PartialCellData(
    chunk.row_key,
    chunk.family_name.value,
    chunk.qualifier.value,
    chunk.timestamp_micros,
    chunk.labels,
    chunk.value)
```

But since we're accessing `chunk.qualifier.value` in on the protobuf I'm fairly certain it will never be `None` (checking now).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/99db799c5ad40c272fb921b9b9ed754958c33d6a/bigtable/google/cloud/bigtable/row_data.py#L286-L292

dhermes: OK I confirmed this value can never be `None` unless we change how a `PartialCellData` instance is constructed:

```python
>>> from google.cloud.bigtable._generated import bigtable_pb2
>>> chunk = bigtable_pb2.ReadRowsResponse.CellChunk()
>>> chunk.qualifier.value == b''
True
>>> chunk.HasField('qualifier')
False
>>> chunk.qualifier is None
False
```

dhermes:
@garye After rebasing so that the `bigtable` tests actually run, there are now 9 failing unit tests. Did you run the tests locally before pushing the change?

Also, should I change the way we construct a `PartialCellData` so that when `chunk.HasField('qualifier')` is `False` we use `None` instead of `chunk.qualifier.value`?

(FWIW, I'd imagine the reason a `google.protobuf.BytesValue` message field is used instead of a scalar `bytes` field is the reason you mentioned. The backend wants a way to distinguish between unset and empty string, which is impossible with a scalar field.)

garye:
Yeah I mentioned in my initial comment that there were failing tests but wanted to get this up for discussion.

Yes, please make that change! You're right about the reason for the `BytesValue` wrapper, AFAIK.

garye:
I think this latest change fixes the test failures and makes the new test pass. @dhermes, PTAL

dhermes:
Thanks @garye. I am rebasing and sorting out the docs issue right now.

----

Funny `git` trivia, you've created I beast I've never seen before with your last commit. It silently "broke" `git rebase` (i.e. the rebase dropped it from history). It breaks `cherry-pick`:

```
$ git cherry-pick d736e7227c58b3769e5ebedda4155700a1354942
error: Commit d736e7227c58b3769e5ebedda4155700a1354942 is a merge but no -m option was given.
fatal: cherry-pick failed
```

and if I do `git show d736e7227c58b3769e5ebedda4155700a1354942`, each of the diff lines has **two** plus/minus signs versus the typical one:

```diff
diff --cc bigtable/google/cloud/bigtable/row_data.py
index ae933a6,8fecb59..13e32c1
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@@ -283,10 -283,10 +283,14 @@@ class PartialRowsData(object)
                  row = self._row = PartialRowData(chunk.row_key)
  
              if cell is None:
++                qual = None
++                if chunk.HasField('qualifier'):
++                    qual = chunk.qualifier.value
++
                  cell = self._cell = PartialCellData(
                      chunk.row_key,
                      chunk.family_name.value,
--                    chunk.qualifier.value,
++                    qual,
                      chunk.timestamp_micros,
                      chunk.labels,
                      chunk.value)
```

dhermes:
I ran `nox -s docs` locally (after rebasing) and it passed, so fingers crossed.

garye:
>> you've created I beast I've never seen before with your last commit.

One for the record books! Sorry about that :/

dhermes:
> Sorry about that

Luckily it was a 4 line diff :grinning:

garye:
Thanks for all the help! Any guidance I can give the customer on when this will be released?

dhermes:
@garye Would you like a release for this? `0.28.1` OK with you (vs. `0.29.0`)?

dhermes:
Jinx! I can do one right now, it should be very quick since we just did one 2 days ago so the release notes will be teeny.

garye:
Perfect! `0.28.1` would be a big help.

PR #4253, opened 2017-10-25
Bucket logging: use bucket name as default prefix

### LINKED ISSUE #4228: bucket.enable_logging should use bucket.name as default prefix
It matches [`gsutil`](https://cloud.google.com/storage/docs/access-logs#delivery) and is probably what most users want by default.
> By default, the object prefix is the name of the bucket for which the logs are enabled.

Currently it sets an [empty prefix](https://googlecloudplatform.github.io/google-cloud-python/latest/_modules/google/cloud/storage/bucket.html#Bucket.enable_logging).

If someone stores logs from multiple buckets in a common *logs-bucket*, without explicitly setting a prefix, they can overwrite each other.

chemelnucfin:
@tseaver My first pull request as a test, so please feel free to look through it and scrutinize it!

Fixes #4228.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -904,7 +904,7 @@ class Bucket(_PropertyMixin):
         info = self._properties.get('logging')
         return copy.deepcopy(info)
 
-    def enable_logging(self, bucket_name, object_prefix=''):
+    def enable_logging(self, bucket_name, object_prefix=None):
         """Enable access logging for this bucket.
 
         See https://cloud.google.com/storage/docs/access-logs
@@ -913,8 +913,13 @@ class Bucket(_PropertyMixin):
         :param bucket_name: name of bucket in which to store access logs
 
         :type object_prefix: str
-        :param object_prefix: prefix for access log filenames
+        :param object_prefix: prefix for access log filenames and defaults
+                              to bucket name.  Use object_prefix='' for no
+                              prefix to reproduce previous default behavior.
         """
+
+        if object_prefix is None:
+            object_prefix = self.name
         info = {'logBucket': bucket_name, 'logObjectPrefix': object_prefix}
         self._patch_property('logging', info)
 
--- a/storage/tests/unit/test_bucket.py
+++ b/storage/tests/unit/test_bucket.py
@@ -939,7 +939,7 @@ class Test_Bucket(unittest.TestCase):
         bucket.enable_logging(LOG_BUCKET)
         info = bucket.get_logging()
         self.assertEqual(info['logBucket'], LOG_BUCKET)
-        self.assertEqual(info['logObjectPrefix'], '')
+        self.assertEqual(info['logObjectPrefix'], bucket.name)
 
     def test_enable_logging(self):
         NAME = 'name'


tseaver:
@chemelnucfin In the future, please use `Closes #4228` in your commit message, so that the issue is closed when the commit is merged.

@lukesneeringer Does this commit suit you?  It doesn't answer my [backward-compatibility question](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4228#issuecomment-338054022) in #4228.

lukesneeringer:
Oh, I see the concern. I _think_ I agree that this is more of a bugfix. Am I correct in understanding that for most users, this change would be invisible, and the caveat would be if anyone was _relying on_ two bucket names to match (which seems unlikely)?

I think I am okay with saying "this is a bugfix" but also documenting that you can send `prefix=''` if you wanted the old behavior. But I am also not passionate about it if you think it is unwise.

tseaver:
The docstring change seems reasonable, given that we want the change (I find the pattern, "let's change a default to suit *my* case, thereby making everybody else change to accommodate it" a bit questionable, but I'm willing to be overruled).

lukesneeringer:
I am pretty hesitant to overrule you here. :-) (As in, I am probably flatly unwilling to do so.)

Is there a more-backwards-compatible change, or is the docstring recommendation the best we can do?

lukesneeringer:
Also, if we make this change, is there concern that _new_ logs would suddenly go to a new bucket? (I think there is that concern.)

tseaver:
@lukesneeringer The prefix is applied to the names of the *blobs* created by logging, not the target bucket.

tseaver:
The choice comes down to deciding how prevalent / important the "store logs from mutiple buckets into one bucket" usecase is, versus leaving existing / working cases alone.  Users who have that usecase can already satisfy it, by supplying an explicit prefix:  so the choice is about convenience for new users who want that, versus backward-compatibilty.

lukesneeringer:
Okay, in that case, back-compat needs to win, unless we can make a compelling case that we need to follow `gsutil`. We should update the documentation to make it clear what to do if one is worried about bucket collision.

Sound right, @tseaver and @chemelnucfin?

chemelnucfin:
For me, it depends on how users use this call.  If it's not a critical
issue for old users, it might be better to make it convenient for future
users and to standardize calls more.

I'd imagine there might be more chances of accidental collision if there
was no prefix for all new users, but if it broke things for old users,
collisions might not be that big a deal (since they might have just
prefixed with the same default name if they didn't want to use the empty
string)

But I don't know enough about the use cases to determine.  I'm only putting
my thoughts out there.  There might be some annoying bugs if they had an
empty string before and now they had a default name.

lukesneeringer:
@tseaver So, if I understand correctly, new blobs will come in with a new pattern. Correct? And, if someone was using the old pattern to read stuff, they might run into issues?

In other words, say someone logs their blobs with `prefix=''`, `name='foo.bar.20171030'`, and we start calling them `bucketname-foo.bar.20171030`, all of the sudden _some other thing_ they wrote to pull out those logs on the other end stops working, and they have to specify `prefix=''` to get the old behavior back.

If that is a correct understanding then yes, that's a back-compat break, and we can not do it on a GA library.

PR #4256, opened 2017-10-25
Retry for 'mutate_rows' that return retryable errors.

justinseinlin:
Trying a new PR in place of (https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4116). 

Reason: A mistake was made in `git rebase` commits that are already pushed to remote, where a `git merge` should have been done.

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -450,15 +450,19 @@ class _RetryableMutateRowsWorker(object):
         responses = self.client._data_stub.MutateRows(
             mutate_rows_request)
 
+        num_responses = 0
         num_retryable_responses = 0
         for response in responses:
             for entry in response.entries:
+                num_responses += 1
                 index = index_into_all_rows[entry.index]
                 self.responses_statuses[index] = entry.status
                 if self._is_retryable(entry.status):
                     num_retryable_responses += 1
                 if entry.status.code == 0:
                     self.rows[index].clear()
+
+        assert len(retryable_rows) == num_responses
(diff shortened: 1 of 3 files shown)


igorbernstein2:
@lukesneeringer Would it be possible to get this reviewed (and possibly merged) before the upcoming repo split?

justinseinlin:
@lukesneeringer do you have any comments that I should address?

lukesneeringer:
Nope, this seems fine to me. :-)

justinseinlin:
I'm good with merge.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -17,6 +17,13 @@
 
 import six
 
+from google.api_core.exceptions import RetryError
+from google.api_core.exceptions import Aborted
+from google.api_core.exceptions import DeadlineExceeded
+from google.api_core.exceptions import ServiceUnavailable
+from google.api_core.exceptions import from_grpc_status
```
dhermes: @justinseinlin @lukesneeringer Future reference: these should be in alphabetical order. (We really need a tool to enforce this, so don't feel bad.)

justinseinlin: :man_facepalming: Thanks for checking and following up with a cleanup PR!

dhermes: Sure thing. (I was a bit intimidated by the size of this PR, so was hoping not to check it. But then I had to write release notes for it, so I had no choice but to read it.)

PR #4257, opened 2017-10-25
Remove gapic_v1.method.wrap_with_paging

### LINKED ISSUE #4233: Drop api_core.gapic_v1.wrap_with_paging
Prefer just inlining the wrapping logic in gapic clients.

/cc @landrito

jonparrott:
Closes #4233


jonparrott:
> Why did this need to happen?

Turns out we didn't need this functionality in the first place. I originally added it for parity with `gax`, as when gax wrapped methods it would also handle wrapping the response with a page iterator. While exploring changing the generator to use api_core, @landrito and I noticed that methods that return operations just construct the operation in the generated method body, so we saw no reason why we couldn't do the same with page iterators.

Issue #4186, opened 2017-10-13
Core "grpc" extra not found by setuptools


markcwhitfield:
`google-cloud-pubsub==0.28.4` requires `google-cloud-core[grpc] > 0.27.1, <0.28dev`. I can see from the core library's [setuptools build](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/setup.py#L62) that the `grpc` extra is declared; however, setuptools does not seem to be able to locate it from the published PyPi package for whatever reason:

```
$ docker run python:3 bash -c "pip install google-cloud-core[grpc]"  # python:2 no different
Collecting google-cloud-core[grpc]
  Downloading google_cloud_core-0.27.1-py2.py3-none-any.whl (50kB)
  google-cloud-core 0.27.1 does not provide the extra 'grpc'
...
Successfully installed cachetools-2.0.1 certifi-2017.7.27.1 chardet-3.0.4 google-auth-1.1.1 google-cloud-core-0.27.1 googleapis-common-protos-1.5.3 idna-2.6 protobuf-3.4.0 pyasn1-0.3.7 pyasn1-modules-0.1.5 requests-2.18.4 rsa-3.4.2 six-1.11.0 urllib3-1.22
```

When installing via `pip` this is not really an issue, so long as `grpcio` is included somewhere in the dependency tree, as `pip` will happily ignore it. Setuptools itself will not, however, and including `google-cloud-pubsub==0.28.4` as a requirement in a setuptools build is currently impossible, so far as I can tell.

dhermes:
@markcwhitfield Sorry about that! The issue is that we haven't released the updated `google-cloud-core` that requires that yet!

markcwhitfield:
Oh, haha. Hadn't even noticed that 0.28.4 was only 3 *hours* old. Just poor timing on my part, I suppose.

dhermes:
And poor something else on our part!

jonparrott:
i'd prefer we hold off on releasing core (it's a big release) and just release a patch of pubsub that fixes this.

dhermes:
SGTM. @lukesneeringer SGTY?

dhermes:
Can I propose we add `.devN` suffixes to all version numbers? This way we can accurately reflect, e.g. that `google-cloud-pubsub` depends on `google-cloud-core >= 0.28.0.dev1`.

lukesneeringer:
[...]

PR #4258, opened 2017-10-25
BigQuery: Increment version to 0.28.0

tswast:
Prepare for beta 2 launch.

Note: this release depends on a google-cloud-core 0.28.0 release happening.


dhermes:
@tswast Does this release depend on un-released features of `google-cloud-core`?

tswast:
Well, it does use the `api_core` package now. I figured that'd be a breaking change, but maybe it doesn't have to be synced with that release?

dhermes:
@tswast Let's just make sure we're all good before pushing to the real PyPI? I can assist you in pushing to TestPyPI if you like.

tswast:
Yeah, I'll probably need some help with this. I'll pick this up again tomorrow when Jon is planning on doing a core release.

dhermes:
Good. We should also make sure to resolve #4186 when `google-cloud-core` gets released.

lukesneeringer:
I want to be _really_ careful about a core release. It almost always involves a release of the world that breaks things.

tswast:
How might I go about patching so that it doesn't use `api_core` so that we aren't synchronized to a core release? Or do we think chances are good it will go smooth today?

dhermes:
@lukesneeringer We have to do at least one more "release the world", so we might as well rip the band-aid.

tswast:
I added a `CHANGELOG.md` file, following Jon's lead in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4266 as a draft of the release notes.

bigquery/setup.py
```diff
@@ -52,14 +52,15 @@
 
 REQUIREMENTS = [
     'google-cloud-core >= 0.28.0, < 0.29dev',
+    'google-api-core >= 0.1.1, < 0.2.0dev',
```
dhermes: @jonparrott This is what you had in mind, adding an explicit dep on `google-api-core`?

jonparrott:
Yes, this is fine. It's not needed right now but since bq explicitly
depends on API core it is cool to go ahead and do it. In the future we can
drop the cloud core dep.

PR #4259, opened 2017-10-25
Add final set of protobuf helpers to api_core

### LINKED ISSUE #3842: Move gax functionality into google.api.core

jonparrott:
Closes #3842 

Note: this is directly ported from gax, barring minor changes, this code has already been reviewed.

### DIFF
--- a/api_core/google/api_core/protobuf_helpers.py
+++ b/api_core/google/api_core/protobuf_helpers.py
@@ -179,15 +179,15 @@ def _set_field_on_message(msg, key, value):
             if isinstance(item, collections.Mapping):
                 getattr(msg, key).add(**item)
             else:
+                # protobuf's RepeatedCompositeContainer doesn't support
+                # append.
                 getattr(msg, key).extend([item])
     elif isinstance(value, collections.Mapping):
         # Assign the dictionary values to the protobuf message.
         for item_key, item_value in value.items():
             set(getattr(msg, key), item_key, item_value)
     elif isinstance(value, Message):
-        # Assign the protobuf message values to the protobuf message.
-        for item_key, item_value in value.ListFields():
-            set(getattr(msg, key), item_key.name, item_value)
+        getattr(msg, key).CopyFrom(value)
     else:
         setattr(msg, key, value)
 
--- a/api_core/google/api_core/protobuf_helpers.py
+++ b/api_core/google/api_core/protobuf_helpers.py
@@ -85,8 +85,20 @@ def get_messages(module):
 
 
 def _resolve_subkeys(key, separator='.'):
-    """Given a key which may actually be a nested key, return the top level
-    key and any nested subkeys as separate values.
+    """Resolve a potentially nested key.
+
+    If the key contains the ``separator`` (e.g. ``.``) then the key will be
+    split on the first instance of the subkey::
+
+       >>> _resolve_subkeys('a.b.c')
+       ('a', 'b.c')
+       >>> _resolve_subkeys('d|e|f', separator='|')
+       ('d', 'e|f')
+
+    If not, the subkey will be :data:`None`::
+
+        >>> _resolve_subkeys('foo')
+        ('foo', None)
 
     Args:
         key (str): A string that may or may not contain the separator.
@@ -95,12 +107,12 @@ def _resolve_subkeys(key, separator='.'):
     Returns:
         Tuple[str, str]: The key and subkey(s).
     """
-    subkey = None
-    if separator in key:
-        index = key.index(separator)
-        subkey = key[index + 1:]
-        key = key[:index]
-    return key, subkey
+    parts = key.split(separator, 1)
+
+    if len(parts) > 1:
+        return parts
+    else:
+        return parts[0], None
 
 
 def get(msg_or_dict, key, default=_SENTINEL):
@@ -115,8 +127,8 @@ def get(msg_or_dict, key, default=_SENTINEL):
             default is generally recommended, as protobuf messages almost
             always have default values for unset values and it is not always
             possible to tell the difference between a falsy value and an
-            unset one. If no default is set then raises :class:`KeyError` will
-            be raised if the key is not present in the object.
+            unset one. If no default is set then :class:`KeyError` will be
+            raised if the key is not present in the object.
 
     Returns:
         Any: The return value from the underlying Message or dict.
@@ -146,10 +158,38 @@ def get(msg_or_dict, key, default=_SENTINEL):
         raise KeyError(key)
 
     # If a subkey exists, call this method recursively against the answer.
-    if subkey and answer is not default:
+    if subkey is not None and answer is not default:
         return get(answer, subkey, default=default)
 
     return answer
+
+
+def _set_field_on_message(msg, key, value):
+    """Set helper for protobuf Messages."""
+    # Attempt to set the value on the types of objects we know how to deal
+    # with.
+    if isinstance(value, (collections.MutableSequence, tuple)):
+        # Clear the existing repeated protobuf message of any elements
+        # currently inside it.
+        while getattr(msg, key):
+            getattr(msg, key).pop()
+
+        # Write our new elements to the repeated field.
+        for item in value:
+            if isinstance(item, collections.Mapping):
+                getattr(msg, key).add(**item)
+            else:
+                getattr(msg, key).extend([item])
+    elif isinstance(value, collections.Mapping):
+        # Assign the dictionary values to the protobuf message.
+        for item_key, item_value in value.items():
+            set(getattr(msg, key), item_key, item_value)
+    elif isinstance(value, Message):
+        # Assign the protobuf message values to the protobuf message.
+        for item_key, item_value in value.ListFields():
+            set(getattr(msg, key), item_key.name, item_value)
+    else:
+        setattr(msg, key, value)
 
 
 def set(msg_or_dict, key, value):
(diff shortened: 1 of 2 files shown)


api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
         if inspect.isclass(candidate) and issubclass(candidate, Message):
             answer[name] = candidate
     return answer
+
+
+def _resolve_subkeys(key, separator='.'):
+    """Given a key which may actually be a nested key, return the top level
+    key and any nested subkeys as separate values.
```
dhermes: PEP257 grumble grumble.

Suggestion:

```python
"""Resolve a potentially nested key.

If the key contains the ``separator`` (e.g. ``.``) then the key will be split on the first
instance of the subkey::

   >>> _resolve_subkeys('a.b.c')
   ('a', 'b.c')
   >>> _resolve_subkeys('d|e|f', separator='|')
   ('d', 'e|f')

If not, the subkey will be :data:`None`::

    >>> _resolve_subkeys('foo')
    ('foo', None)

...
"""
```

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+        separator (str): The namespace separator. Defaults to `.`.
+
+    Returns:
+        Tuple[str, str]: The key and subkey(s).
+    """
+    subkey = None
+    if separator in key:
+        index = key.index(separator)
+        subkey = key[index + 1:]
+        key = key[:index]
```
dhermes: Why not use `key.split(separator, 1)`? You can even use the length of `key.split(separator, 1)` without having to do the `separator in key` check.

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+    Args:
+        mdg_or_dict (Union[~google.protobuf.message.Message, Mapping]): the
+            object.
+        key (str): The key to retrieve from the object.
+        default (Any): If the key is not present on the object, and a default
+            is set, returns that default instead. A type-appropriate falsy
+            default is generally recommended, as protobuf messages almost
+            always have default values for unset values and it is not always
+            possible to tell the difference between a falsy value and an
+            unset one. If no default is set then raises :class:`KeyError` will
```
dhermes: ISTM you want "raises" deleted in this sentence?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+            messages and dictionaries may not have consistent behavior.
+        TypeError: If ``msg_or_dict`` is not a Message or Mapping.
+    """
+    # We may need to get a nested key. Resolve this.
+    key, subkey = _resolve_subkeys(key)
+
+    # Attempt to get the value from the two types of objects we know about.
+    # If we get something else, complain.
+    if isinstance(msg_or_dict, Message):
+        answer = getattr(msg_or_dict, key, default)
```
dhermes: There is no better way to do this with the `google.protobuf` API? Something like `HasField` (not the right choice here, but a relative maybe)?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+            'get() expected a dict or protobuf message, got {!r}.'.format(
+                type(msg_or_dict)))
+
+    # If the object we got back is our sentinel, raise KeyError; this is
+    # a "not found" case.
+    if answer is _SENTINEL:
+        raise KeyError(key)
+
+    # If a subkey exists, call this method recursively against the answer.
+    if subkey and answer is not default:
```
dhermes: Can you add an explicit [`None`-check][1] for `subkey`? Unless you want the empty string to be ignored? Or maybe you want your `_resolve_subkeys` to barf if the 2nd return value is false-y?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+        TypeError: If ``msg_or_dict`` is not a Message or dictionary.
+    """
+    # Sanity check: Is our target object valid?
+    if not isinstance(msg_or_dict, (collections.MutableMapping, Message)):
+        raise TypeError(
+            'set() expected a dict or protobuf message, got {!r}.'.format(
+                type(msg_or_dict)))
+
+    # We may be setting a nested key. Resolve this.
+    key, subkey = _resolve_subkeys(key)
```
dhermes: I had to do a double-take on the code below because you changed the purpose of `key` here. Instead of over-writing, just give it a new name? `base_key`?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+
+    # If a subkey exists, then get that object and call this method
+    # recursively against it using the subkey.
+    if subkey is not None:
+        if isinstance(msg_or_dict, collections.MutableMapping):
```
dhermes: At this point, the function becomes nasty to grok. Why not just do your type-check and then dispatch to an appropriate `dict`-only or `Message`-only helper?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+    elif isinstance(value, Message):
+        # Assign the protobuf message values to the protobuf message.
+        for item_key, item_value in value.ListFields():
+            set(getattr(msg_or_dict, key), item_key.name, item_value)
+    else:
+        setattr(msg_or_dict, key, value)
+
+
+def setdefault(msg_or_dict, key, value):
+    """Set the key on a protobuf Message or dictioanary to a given value if the
```
dhermes: s/dictioanary/dictionary

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -76,3 +82,150 @@ def get_messages(module):
...
+    current value is falsy.
+
+    Because protobuf Messages do not distinguish between unset values and
+    falsy ones particularly well, this method treats any falsy value
```
dhermes: "particularly well (by design), this ..."

api_core/google/api_core/protobuf_helpers.py
lukesneeringer: I was not able to find one when I wrote this (months ago).

api_core/google/api_core/protobuf_helpers.py
lukesneeringer: I just did not think of that at the time.

api_core/google/api_core/protobuf_helpers.py
dhermes: What about something like this:

```python
>>> from google.cloud.bigtable._generated import data_pb2
>>>
>>> t1 = data_pb2.TimestampRange()
>>> t2 = data_pb2.TimestampRange(start_timestamp_micros=100)
>>> t3 = data_pb2.TimestampRange(end_timestamp_micros=200)
>>> t4 = data_pb2.TimestampRange(
...     start_timestamp_micros=1000, end_timestamp_micros=2000)
>>>
>>> d1 = {field.name: value for field, value in t1.ListFields()}
>>> d2 = {field.name: value for field, value in t2.ListFields()}
>>> d3 = {field.name: value for field, value in t3.ListFields()}
>>> d4 = {field.name: value for field, value in t4.ListFields()}
>>>
>>> d1
{}
>>> d2
{'start_timestamp_micros': 100}
>>> d3
{'end_timestamp_micros': 200}
>>> d4
{'end_timestamp_micros': 2000, 'start_timestamp_micros': 1000}
```

(Probably not better.)

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -146,12 +158,40 @@ def get(msg_or_dict, key, default=_SENTINEL):
...
+        # currently inside it.
+        while getattr(msg, key):
+            getattr(msg, key).pop()
+
+        # Write our new elements to the repeated field.
+        for item in value:
+            if isinstance(item, collections.Mapping):
+                getattr(msg, key).add(**item)
+            else:
+                getattr(msg, key).extend([item])
```
dhermes: Worth adding a note that `RepeatedCompositeContainer` doesn't support `append`?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -146,12 +158,40 @@ def get(msg_or_dict, key, default=_SENTINEL):
...
+    elif isinstance(value, collections.Mapping):
+        # Assign the dictionary values to the protobuf message.
+        for item_key, item_value in value.items():
+            set(getattr(msg, key), item_key, item_value)
+    elif isinstance(value, Message):
+        # Assign the protobuf message values to the protobuf message.
+        for item_key, item_value in value.ListFields():
+            set(getattr(msg, key), item_key.name, item_value)
```
dhermes: This is probably not what we want to do. Instead:

```python
getattr(msg, key).CopyFrom(value)
```

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -146,12 +158,40 @@ def get(msg_or_dict, key, default=_SENTINEL):
...
+    if subkey is not None and answer is not default:
         return get(answer, subkey, default=default)
 
     return answer
 
 
+def _set_field_on_message(msg, key, value):
+    """Set helper for protobuf Messages."""
+    # Attempt to set the value on the types of objects we know how to deal
+    # with.
```
dhermes: Can you just make:

```python
existing = getattr(msg, key)
```

and refer to `existing` throughout?

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -146,12 +158,40 @@ def get(msg_or_dict, key, default=_SENTINEL):
...
+        while getattr(msg, key):
+            getattr(msg, key).pop()
+
+        # Write our new elements to the repeated field.
+        for item in value:
+            if isinstance(item, collections.Mapping):
+                getattr(msg, key).add(**item)
+            else:
+                getattr(msg, key).extend([item])
+    elif isinstance(value, collections.Mapping):
```
dhermes: This seems "wrong". I.e. why would a dictionary be used to set a protobuf value? It gets really nasty if `value` contains subdictionaries. If it did not, one could do:

```python
existing = getattr(msg, key)
new_pb = existing.__class__(**value)
existing.CopyFrom(new_pb)
```

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -146,12 +158,40 @@ def get(msg_or_dict, key, default=_SENTINEL):
...
 
+def _set_field_on_message(msg, key, value):
+    """Set helper for protobuf Messages."""
+    # Attempt to set the value on the types of objects we know how to deal
+    # with.
+    if isinstance(value, (collections.MutableSequence, tuple)):
+        # Clear the existing repeated protobuf message of any elements
+        # currently inside it.
+        while getattr(msg, key):
+            getattr(msg, key).pop()
```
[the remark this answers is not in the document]
jonparrott: 1. Okay, noted, any action needed?
2. We don't, we should depend on protobuf to enforce type checking there.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
jonparrott: I mean, there's a test for it and it works.

api_core/google/api_core/protobuf_helpers.py
jonparrott: Done.

api_core/google/api_core/protobuf_helpers.py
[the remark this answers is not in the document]
dhermes: 1. Probably remove `tuple` from the type list above
1. I guess we only rely on `value` being iterable.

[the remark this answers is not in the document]
jonparrott: Discussed over chat, will leave as is because the new value is being checked a tuple but the *existing* value is being popped.

Issue #4234, opened 2017-10-20
Pub/Sub Subscriber does not catch & retry UNAVAILABLE errors


kir-titievsky:
A basic Pub/Sub message consumer stops consuming messages after a retryable error (see stack trace below, but in short `_Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, The service was unable to fulfill your request. Please try again. [code=8a75])>`).  The app does not crash but the stream never recovers and continue to receive messages.  Interesting observations;
- If I simply turn off WiFi on my laptop and run the same code, it keeps retrying until the machine is connected to the network and functions as expected. This tells me that this is a reaction to the specific StatusCode
- The exception sometimes happens on startup sometimes mid-stream.

Expected behavior:
- The application code would continue retrying to build the streamingPull connection and eventually recover and receive messages.
- This would be handled and surfaced as a warning, rather than a thread-killing exception.

This might be the same issue as [2683](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2683).  [This comment](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2683#issuecomment-337876366), in particular, seems like the solution that I would expect the client library to implement.  

Answers to standard questions:
1. OS type and version
MacOS Sierra 10.12.6
2. Python version and virtual environment information `python --version`
Python 2.7.10 (running in virtualenv)
3. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`
<pre>
$ pip show google-cloud
Name: google-cloud
Version: 0.27.0
pip show google-cloud-pubsub
Name: google-cloud-pubsub
Version: 0.28.4
</pre>
4. Stacktrace if available
<pre>Exception in thread Consumer helper: consume bidirectional stream:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/Users/kir/cloud/env/lib/python2.7/site-packages/google/cloud/pubsub_v1/subscriber/_consumer.py", line 248, in _blocking_consume
    self._policy.on_exception(exc)
[...]

PR #4265, opened 2017-10-26
Add futures for subscriptions.

### LINKED ISSUE #3888: Pub/Sub has no way to track errors from the subscriber thread.
Presently if you subscribe and an error occurs in the consumer's thread there's no way to respond to the error on the main thread.

It just logs the exception and hangs:

```python
Exception in thread Consumer helper: consume bidirectional stream:
  File "/Users/jonwayne/.pyenv/versions/3.6.0/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Users/jonwayne/.pyenv/versions/3.6.0/lib/python3.6/threading.py", line 864, in run
  File "/Users/jonwayne/workspace/python-docs-samples/pubsub/cloud-client/env/lib/python3.6/site-packages/google/cloud/pubsub_v1/subscriber/_consumer.py", line 248, in _blocking_consume
  File "/Users/jonwayne/workspace/python-docs-samples/pubsub/cloud-client/env/lib/python3.6/site-packages/google/cloud/pubsub_v1/subscriber/policy/thread.py", line 135, in on_exception
    raise exception
  File "/Users/jonwayne/workspace/python-docs-samples/pubsub/cloud-client/env/lib/python3.6/site-packages/google/cloud/pubsub_v1/subscriber/_consumer.py", line 234, in _blocking_consume
    for response in response_generator:
  File "/Users/jonwayne/workspace/python-docs-samples/pubsub/cloud-client/env/lib/python3.6/site-packages/grpc/_channel.py", line 363, in __next__
    return self._next()
  File "/Users/jonwayne/workspace/python-docs-samples/pubsub/cloud-client/env/lib/python3.6/site-packages/grpc/_channel.py", line 357, in _next
    raise self
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.NOT_FOUND, Resource not found (resource=my-subs).)>
```

It seems that `consumer.open` should return a future. This future could be used to both block the main thread to consume message and pass through exceptions.

So the current sample:

```python
def receive_messages(project, subscription_name):
    """Receives messages from a pull subscription."""
    subscription_path = subscriber.subscription_path(
        project, subscription_name)

    def callback(message):
        print('Received message: {}'.format(message))
        message.ack()

    subscriber.subscribe(subscription_path, callback=callback)

    # The subscriber is non-blocking, so we must keep the main thread from
    # exiting to allow it to process messages in the background.
    print('Listening for messages on {}'.format(subscription_path))
    while True:
        time.sleep(60)
```

Could instead be:

```python
def receive_messages(project, subscription_name):
    """Receives messages from a pull subscription."""
    subscription_path = subscriber.subscription_path(
        project, subscription_name)

    def callback(message):
        print('Received message: {}'.format(message))
        message.ack()

    future = subscriber.subscribe(subscription_path, callback=callback)

    print('Listening for messages on {}'.format(subscription_path))
    # The subscriber is non-blocking but returns a future that allows us
    # to block the main thread and allow messages to process in the
    # background. This effectively blocks forever unless an error
    # occurs in the subscriber thread in which case it'll be
    # re-raised here.
    future.result()
```

### LINKED ISSUE #4019: Publish future blocks indefinitely if API raises exception
1. OS type and version: macOS 10.12.6
2. Python version and virtual environment information `python --version`: Python 3.4.6
3. google-cloud-python version(s):
```
google-cloud-core==0.27.1
google-cloud-pubsub==0.28.3
dill==0.2.7.1
future==0.16.0
google-auth==1.1.0
google-gax==0.15.14
googleapis-common-protos==1.5.2
grpc-google-iam-v1==0.11.3
grpcio==1.6.0
oauth2client==3.0.0
ply==3.8
protobuf==3.4.0
psutil==5.3.1
```
4. Stacktrace if available
```
>>> Exception in thread Thread-7:
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/retry.py", line 121, in inner
    return to_call(*args)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/retry.py", line 68, in inner
    return a_func(*updated_args, **kwargs)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/grpc/_channel.py", line 492, in __call__
    return _end_unary_response_blocking(state, call, False, deadline)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/grpc/_channel.py", line 440, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.NOT_FOUND, Requested project not found or user does not have access to it (project=not-a-cloud-259388BD5DF4). Make sure to specify the unique project identifier and not the Google Cloud Console display name.)>

  File "/Users/schultzm/.pyenv/versions/3.4.6/lib/python3.4/threading.py", line 911, in _bootstrap_inner
    self.run()
  File "/Users/schultzm/.pyenv/versions/3.4.6/lib/python3.4/threading.py", line 859, in run
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/cloud/pubsub_v1/publisher/batch/thread.py", line 212, in monitor
    return self._commit()
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/cloud/pubsub_v1/publisher/batch/thread.py", line 173, in _commit
    self.messages,
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/cloud/gapic/pubsub/v1/publisher_client.py", line 314, in publish
    return self._publish(request, options)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/api_callable.py", line 452, in inner
    return api_caller(api_call, this_settings, request)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/api_callable.py", line 438, in base_caller
    return api_call(*args)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/api_callable.py", line 376, in inner
    return a_func(*args, **kwargs)
  File "/Users/schultzm/.virtualenvs/onsite/lib/python3.4/site-packages/google/gax/retry.py", line 127, in inner
    ' classified as transient', exception)
google.gax.errors.RetryError: RetryError(Exception occurred in retry method that was not classified as transient, caused by <_Rendezvous of RPC that terminated with (StatusCode.NOT_FOUND, Requested project not found or user does not have access to it (project=not-a-cloud-259388BD5DF4). Make sure to specify the unique project identifier and not the Google Cloud Console display name.)>)
```
5. Steps to reproduce

Attempt to publish to a unknown project/topic (or perhaps a topic you don't have permissions for).

6. Code example

```
from json import dumps

topic = 'projects/not-a-cloud-259388BD5DF4/topics/not-a-topic'
data = {'toss': 'exception'}
payload = dumps(data).encode('utf8')
future = publisher.publish(topic, payload)
```

At this point you'll have a future object, but the Batch thread that is going to send the message is already working on it and tries to call the API. It can [throw an exception that is uncaught](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4a8e155/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py#L171-L174) which will cause the thread to exit. The future that is returned will never get updated and the user has no way to access that exception.

I think just throwing that code in a try/except and running `set_exception` on all the `self._futures` would suffice?

lukesneeringer:
This causes the `.open` method on subscriptions to return a future, which can be used to block the main thread or trap exceptions.

Fixes #3888.
Fixes #4019.

@jonparrott A couple items we should nail down and possibly tweak before merging this:

  * The current interface has a three-object pass to start consuming messages: instantiate a client, instantiate a subscription, then call `.open` on the subscription, which now returns a future. However, we currently have some shortcut logic that lets you pass a `callback` on step 2. You still get the subscription back, but `.open` is called implicitly; there's no good place for me to return the future in that case.
      * This is kind of a big deal; this future is basically _the_ way to both block and handle errata.
      * On the other hand, maybe this is fine; `.future` is a public property.
          * At the very minimum, I should update all documentation to show the full baton pass and highlight the future and what it is for.
  * There is not really any good way to stop consuming automatically. I can not do it in `on_exception` because that runs in one of the threads that the `helper_threads.stop_all()` method is trying to burn down, so it raises `RuntimeError`.
      * I could have an `.add_done_callback` call in `Future.__init__` that adds a callback that does the right thing, but this feels like an interface violation to me.
      * Maybe this is fine too; we (a) leave it up to users to decide how to handle errors, albeit (b) also fix bugs like #4234 so they are far less common.

Holding off on updating tests and docs until we make a decision.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/futures.py
+++ b/pubsub/google/cloud/pubsub_v1/futures.py
@@ -1,0 +1,176 @@
+# Copyright 2017, Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import threading
+
+import google.api_core.future
+from google.cloud.pubsub_v1.publisher import exceptions
+
+
+class Future(google.api_core.future.Future):
+    """Encapsulation of the asynchronous execution of an action.
+
+    This object is returned from asychronous Pub/Sub calls, and is the
+    interface to determine the status of those calls.
+
+    This object should not be created directly, but is returned by other
+    methods in this library.
+    """
+    _SENTINEL = object()
+
+    def __init__(self):
+        self._result = self._SENTINEL
+        self._exception = self._SENTINEL
+        self._callbacks = []
+        self._completed = threading.Event()
+
+    def cancel(self):
+        """Actions in Pub/Sub generally may not be canceled.
+
+        This method always returns False.
+        """
+        return False
+
+    def cancelled(self):
+        """Actions in Pub/Sub generally may not be canceled.
+
+        This method always returns False.
+        """
+        return False
+
+    def running(self):
+        """Actions in Pub/Sub generally may not be canceled.
+
+        Returns:
+            bool: ``True`` if this method has not yet completed, or
+                ``False`` if it has completed.
+        """
+        if self.done():
+            return False
+        return True
+
+    def done(self):
+        """Return True if the publish has completed, False otherwise.
+
+        This still returns True in failure cases; checking :meth:`result` or
+        :meth:`exception` is the canonical way to assess success or failure.
+        """
+        return (self._exception is not self._SENTINEL or
+                self._result is not self._SENTINEL)
+
+    def result(self, timeout=None):
+        """Return the message ID, or raise an exception.
+
+        This blocks until the message has successfully been published, and
+        returns the message ID.
+
+        Args:
+            timeout (Union[int, float]): The number of seconds before this call
+                times out and raises TimeoutError.
+
+        Returns:
+            str: The message ID.
+
+        Raises:
+            ~.pubsub_v1.TimeoutError: If the request times out.
+            Exception: For undefined exceptions in the underlying
+                call execution.
+        """
+        # Attempt to get the exception if there is one.
+        # If there is not one, then we know everything worked, and we can
+        # return an appropriate value.
+        err = self.exception(timeout=timeout)
+        if err is None:
+            return self._result
+        raise err
+
+    def exception(self, timeout=None, _wait=1):
+        """Return the exception raised by the call, if any.
+
+        This blocks until the message has successfully been published, and
+        returns the exception. If the call succeeded, return None.
+
+        Args:
+            timeout (Union[int, float]): The number of seconds before this call
+                times out and raises TimeoutError.
+
+        Raises:
+            TimeoutError: If the request times out.
+
+        Returns:
+            Exception: The exception raised by the call, if any.
+        """
+        # Wait until the future is done.
+        if not self._completed.wait(timeout=timeout):
+            raise exceptions.TimeoutError('Timed out waiting for result.')
+
+        # If the batch completed successfully, this should return None.
+        if self._result is not self._SENTINEL:
+            return None
+
+        # Okay, this batch had an error; this should return it.
+        return self._exception
+
+    def add_done_callback(self, fn):
+        """Attach the provided callable to the future.
+
+        The provided function is called, with this future as its only argument,
+        when the future finishes running.
+        """
+        if self.done():
+            fn(self)
+        self._callbacks.append(fn)
+
+    def set_result(self, result):
+        """Set the result of the future to the provided result.
+
+        Args:
+            result (str): The message ID.
+        """
+        # Sanity check: A future can only complete once.
+        if self.done():
+            raise RuntimeError('set_result can only be called once.')
+
+        # Set the result and trigger the future.
+        self._result = result
+        self._trigger()
+
+    def set_exception(self, exception):
+        """Set the result of the future to the given exception.
+
+        Args:
+            exception (:exc:`Exception`): The exception raised.
+        """
+        # Sanity check: A future can only complete once.
+        if self.done():
+            raise RuntimeError('set_exception can only be called once.')
+
+        # Set the exception and trigger the future.
+        self._exception = exception
+        self._trigger()
+
+    def _trigger(self):
+        """Trigger all callbacks registered to this Future.
+
+        This method is called internally by the batch once the batch
+        completes.
+
+        Args:
+            message_id (str): The message ID, as a string.
+        """
+        self._completed.set()
+        for callback in self._callbacks:
+            callback(self)
--- a/pubsub/google/cloud/pubsub_v1/futures.py
+++ b/pubsub/google/cloud/pubsub_v1/futures.py
@@ -63,7 +63,7 @@ class Future(google.api_core.future.Future):
         return True
 
     def done(self):
-        """Return True if the publish has completed, False otherwise.
+        """Return True the future is done, False otherwise.
 
         This still returns True in failure cases; checking :meth:`result` or
         :meth:`exception` is the canonical way to assess success or failure.
@@ -131,14 +131,14 @@ class Future(google.api_core.future.Future):
         when the future finishes running.
         """
         if self.done():
-            fn(self)
+            return fn(self)
         self._callbacks.append(fn)
 
     def set_result(self, result):
         """Set the result of the future to the provided result.
 
         Args:
-            result (str): The message ID.
+            result (Any): The result
         """
         # Sanity check: A future can only complete once.
         if self.done():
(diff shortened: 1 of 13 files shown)


lukesneeringer:
> doesn't `.close()` cover this? Or am I missing something?

It does, but if you call it in `.on_exception`, you get a `RuntimeError` because a thread tries to join itself.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, client, topic, settings, autocommit=True):
 
         # These objects are all communicated between threads; ensure that
         # any writes to them are atomic.
+        # import pdb ; pdb.set_trace()
```
dhermes: Oops?

lukesneeringer: Ha. Thanks. :-)

jonparrott:
Can we have close() send a signal to exit if it's being invoked from the "wrong" thread?

lukesneeringer:
_Maybe._

lukesneeringer:
@jonparrott This is ready for review. I am going to punt on `.close()` for now because I think it is more valuable to push this as is and spend cycles on a couple of the other bugs. I am also not 100% sure that `.close()` is even the right behavior.

jonparrott:
@lukesneeringer totally fair

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns:
+            bool: ``True`` if this method has not yet completed, or
+                ``False`` if it has completed.
+        """
+        if self.done():
+            return False
+        return True
+
+    def done(self):
+        """Return True if the publish has completed, False otherwise.
```
jonparrott: This isn't accurate is it?

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        # Attempt to get the exception if there is one.
+        # If there is not one, then we know everything worked, and we can
+        # return an appropriate value.
+        err = self.exception(timeout=timeout)
+        if err is None:
+            return self._result
+        raise err
+
+    def exception(self, timeout=None, _wait=1):
```
jonparrott: what's the purpose of `_wait` here?

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -0,0 +1,176 @@
...
+        return self._exception
+
+    def add_done_callback(self, fn):
+        """Attach the provided callable to the future.
+
+        The provided function is called, with this future as its only argument,
+        when the future finishes running.
+        """
+        if self.done():
+            fn(self)
```
jonparrott: return so that you don't append this needless to the callbacks?

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -0,0 +1,176 @@
...
+        self._callbacks.append(fn)
+
+    def set_result(self, result):
+        """Set the result of the future to the provided result.
+
+        Args:
+            result (str): The message ID.
```
jonparrott: This isn't accurate, is it? This will usually be called with `None`.

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -0,0 +1,176 @@
+# Copyright 2017, Google Inc. All rights reserved.
```
jonparrott: Note: it's not technically required to make this public.

pubsub/google/cloud/pubsub_v1/publisher/futures.py
```diff
@@ -14,156 +14,27 @@
...
-        Args:
-            message_id (str): The message ID, as a string.
-        """
-        self._completed.set()
-        for callback in self._callbacks:
-            callback(self)
+    # The publishing-side subclass does not need any special behavior
+    # at this time.
+    #
+    # However, there is still a subclass so that if someone attempts
```
jonparrott: I don't think we should bother with this distinction - in fact, I don't think we should bother with the distinction that these are anything other that `google.api_core.future.Future` instances.

pubsub/google/cloud/pubsub_v1/futures.py
lukesneeringer: I tend to oppose returning instances of private classes to users. (I think @dhermes dislikes that also.)

pubsub/google/cloud/pubsub_v1/futures.py
lukesneeringer: Oops, old argument that persisted through a previous refactor (this file is just moved from another location). I will remove it.

pubsub/google/cloud/pubsub_v1/futures.py
dhermes: I **really** dislike that.

lukesneeringer: @jonparrott and I discussed offline. I changed the docs to reference `google.api_core.futures.Future`, otherwise this is staying as-is.

Merging once CI goes green.

PR #4266, opened 2017-10-26
Release google-api-core v0.1.0

jonparrott:


jonparrott:
Waiting on @dhermes' approval as well.

lukesneeringer:
Valid on GitHub is fine. When we repo dissect this will end up in the releases area (hopefully in addition to, not instead of, `CHANGELOG.md`).

PR #4269, opened 2017-10-26
Set gRPC message options and keepalive.

### LINKED ISSUE #3886: Pubsub API redesign broke support for PUBSUB_EMULATOR_HOST
Broke 2 days ago with this diff:
https://github.com/GoogleCloudPlatform/google-cloud-python/commit/4a8e155c0e85bf80edbf2e5f1e9342bc3f571acb#diff-c15e14c96ae7c1e14ebdff735c3080e6L62

### LINKED ISSUE #4083: pubsub: streaming pull needs keepalive
With default configuration, streaming pull will hang if the stream is idle for a few minutes. This usually happens when the subscription has no more messages. In Go, we set gRPC keepalive to 5 minutes [(CL)](https://code-review.googlesource.com/#/c/gocloud/+/17070/). A similar option should be available in Python, though I'm not sure what the code to pass in the option looks like.

lukesneeringer:
Fixes #4083, although the UNAVAILABLE errors still occur sometimes for other reasons.

That will need to be handled with new retry logic, and is being tracked as #4234.

Addendum: Fixes #3886.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -15,11 +15,14 @@
 from __future__ import absolute_import
 
 import copy
+import os
 import pkg_resources
 import threading
 
+import grpc
 import six
 
+from google.api_core import grpc_helpers
 from google.cloud.gapic.pubsub.v1 import publisher_client
 
 from google.cloud.pubsub_v1 import _gapic
@@ -53,6 +56,28 @@ class Client(object):
             Generally, you should not need to set additional keyword arguments.
     """
     def __init__(self, batch_settings=(), batch_class=thread.Batch, **kwargs):
+        # Sanity check: Is our goal to use the emulator?
+        # If so, create a grpc insecure channel with the emulator host
+        # as the target.
+        if os.environ.get('PUBSUB_EMULATOR_HOST'):
+            kwargs['channel'] = grpc.insecure_channel(
+                target=os.environ.get('PUBSUB_EMULATOR_HOST'),
+            )
+
+        # Use a custom channel.
+        # We need this in order to set appropriate default message size and
+        # keepalive options.
+        if 'channel' not in kwargs:
+            kwargs['channel'] = grpc_helpers.create_channel(
+                credentials=kwargs.get('credentials', None),
+                target=self.target,
+                scopes=publisher_client.PublisherClient._ALL_SCOPES,
+                options={
+                    'grpc.max_send_message_length': -1,
+                    'grpc.max_receive_message_length': -1,
+                }.items(),
+            )
+
         # Add the metrics headers, and instantiate the underlying GAPIC
         # client.
         kwargs['lib_name'] = 'gccl'
@@ -65,6 +90,18 @@ class Client(object):
         self._batch_class = batch_class
         self._batch_lock = threading.Lock()
         self._batches = {}
+
+    @property
+    def target(self):
+        """Return the target (where the API is).
+
+        Returns:
+            str: The location of the API.
+        """
+        return '{host}:{port}'.format(
+            host=publisher_client.PublisherClient.SERVICE_ADDRESS,
+            port=publisher_client.PublisherClient.DEFAULT_SERVICE_PORT,
+        )
 
     def batch(self, topic, message, create=True, autocommit=True):
         """Return the current batch for the provided topic.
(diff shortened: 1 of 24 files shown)


lukesneeringer:
Ignore the long list of commits that included #4265; squash will fix it. :-)

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -49,6 +53,23 @@ class in order to define your own consumer. This is primarily
             arguments.
     """
     def __init__(self, policy_class=thread.Policy, **kwargs):
+        # Use a custom channel.
+        # We need this in order to set appropriate default message size and
+        # keepalive options.
+        target = '{host}:{port}'.format(
+            host=subscriber_client.SubscriberClient.SERVICE_ADDRESS,
+            port=subscriber_client.SubscriberClient.DEFAULT_SERVICE_PORT,
+        )
```
dhermes: Why do this at runtime? Instead do it at import time?

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -49,6 +53,23 @@ class in order to define your own consumer. This is primarily
...
+        kwargs.setdefault('channel', grpc_helpers.create_channel(
+            target=target,
+            scopes=subscriber_client.SubscriberClient._ALL_SCOPES,
+            options={
+                'grpc.max_send_message_length': -1,
+                'grpc.max_receive_message_length': -1,
+                'grpc.keepalive_time_ms': 30000,
+            }.items(),
```
dhermes: This (using `.items()`) is a strange indirection, just use a list/tuple or pairs?

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -49,6 +53,23 @@ class in order to define your own consumer. This is primarily
...
     """
     def __init__(self, policy_class=thread.Policy, **kwargs):
+        # Use a custom channel.
+        # We need this in order to set appropriate default message size and
+        # keepalive options.
+        target = '{host}:{port}'.format(
+            host=subscriber_client.SubscriberClient.SERVICE_ADDRESS,
+            port=subscriber_client.SubscriberClient.DEFAULT_SERVICE_PORT,
+        )
+        kwargs.setdefault('channel', grpc_helpers.create_channel(
```
dhermes: Would prefer you "do less". i.e. create a `channel` as a single statement and then set the default as another statement. Easier to grok?

lukesneeringer: I do not think it works that way. You can not generate a channel with no target.

dhermes: @lukesneeringer I'm suggesting you make a global named `_TARGET` with this same value rather than re-computing it every time the constructor is called.

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -16,6 +16,10 @@
 
 import pkg_resources
 
+import grpc
+
+from google import auth
+from google.api_core import grpc_helpers
```
dhermes: @lukesneeringer I'm suggesting that up here, right after importing, you do

```python
_TARGET = '{host}:{port}'.format(
    host=subscriber_client.SubscriberClient.SERVICE_ADDRESS,
    port=subscriber_client.SubscriberClient.DEFAULT_SERVICE_PORT,
)
```

pubsub/google/cloud/pubsub_v1/subscriber/client.py
dhermes: Down here, I'm suggesting you do:

```python
channel = grpc_helpers.create_channel(
    target=_TARGET,
    scopes=subscriber_client.SubscriberClient._ALL_SCOPES,
    options=(
        ('grpc.max_send_message_length', -1),
        ('grpc.max_receive_message_length', -1),
        ('grpc.keepalive_time_ms', 30000),
    ),
)
kwargs.setdefault('channel', channel)
```

Though it just occurs to me that you probably want to avoid creating `channel` if `'channel' in kwargs` (gRPC may do some things with the channel on creation?)

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -66,6 +91,18 @@ def __init__(self, batch_settings=(), batch_class=thread.Batch, **kwargs):
         self._batch_lock = threading.Lock()
         self._batches = {}
 
+    @property
+    def target(self):
```
dhermes: @lukesneeringer This doesn't address the issue I had with it. It will still get re-computed every time this property is accessed (i.e. at run time, rather than import time).

PR #4273, opened 2017-10-27
bigquery: add simple benchmark

pongad:
Runs queries and measure time to first response
and time to iterate all rows

### DIFF
--- a/bigquery/benchmark/benchmark.py
+++ b/bigquery/benchmark/benchmark.py
@@ -1,0 +1,32 @@
+from google.cloud import bigquery
+from datetime import datetime
+import json
+import sys
+
+if len(sys.argv) < 2:
+    raise Exception('need query file, usage: python {0} <queries.json>'.format(sys.argv[0]))
+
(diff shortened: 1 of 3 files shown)


pongad:
@lukesneeringer We'll probably start up something similar for spanner soon. According to current design, the benchmarks will be a little more involved though. As we have more benchmarks, I also hope things will become less ad-hoc.

So the answer to your question is: yes, but the "flavor" of the PRs will probably change.

bigquery/benchmark/benchmark.py
```diff
@@ -0,0 +1,32 @@
+from google.cloud import bigquery
```
dhermes: This file needs the `LICENSE` header

pongad:
@dhermes Sorry about that. https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4284

bigquery/benchmark/benchmark.py
dhermes: Thanks

PR #4277, opened 2017-10-30
Correct speech documentation.

### LINKED ISSUE #4275: Speech: Deprecated doc!
Hello,
I was using cloud speech and therefore reading the docs when I came across a problem:
```python
operation = client.long_running_recognize(audio, config)
```
it returns an _OperationFuture object from python Gax Library. Here is the [source code](http://gax-python.readthedocs.io/en/latest/_modules/google/gax.html?highlight=_operation).
But the attributes the documentation tells this object has are untrue. Indeed operation.complete really is operation.done().
And operation.poll() really is operation._poll() but isn't necessary here because operation.results became the method operation.result() which blocks untils operation.complete() is True.
The documentation that needs an update is here:
https://googlecloudplatform.github.io/google-cloud-python/latest/speech/index.html

Kind regards :dango:

lukesneeringer:
Properly document the `c.f.Future` interface.

Fixes #4275.

### DIFF
--- a/docs/speech/index.rst
+++ b/docs/speech/index.rst
@@ -55,14 +55,7 @@ See: `Speech Asynchronous Recognize`_
     ...         sample_rate_hertz=44100,
     ...     ),
     ... )
-    >>> retry_count = 100
-    >>> while retry_count > 0 and not operation.complete:
-    ...     retry_count -= 1
-    ...     time.sleep(10)
-    ...     operation.poll()  # API call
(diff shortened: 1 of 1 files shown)


chemelnucfin:
I think it should be operation.result?

lukesneeringer:
@chemelnucfin I think you are right. Mind updating this to be accurate?

chemelnucfin:
See #4281

PR #4281, opened 2017-10-30
Speech: long_run_recognize document update

### LINKED ISSUE #4275: Speech: Deprecated doc!

chemelnucfin:
If someone can point me how else to simply grab a really big audio file, that could make the test simpler.

### DIFF
--- a/docs/speech/index.rst
+++ b/docs/speech/index.rst
@@ -54,8 +54,8 @@ See: `Speech Asynchronous Recognize`_
     ...         sample_rate_hertz=44100,
     ...     ),
     ... )
-    >>> opresult = operation.result()
-    >>> for result in opresult.results:
+    >>> op_result = operation.result()
+    >>> for result in op_result.results:
     ...     for alternative in result.alternatives:
     ...         print('=' * 20)
     ...         print(alternative.transcript)
--- a/docs/speech/index.rst
+++ b/docs/speech/index.rst
@@ -42,7 +42,6 @@ See: `Speech Asynchronous Recognize`_
 
 .. code-block:: python
 
-    >>> import time
     >>> from google.cloud import speech
     >>> client = speech.SpeechClient()
     >>> operation = client.long_running_recognize(
@@ -55,14 +54,8 @@ See: `Speech Asynchronous Recognize`_
     ...         sample_rate_hertz=44100,
     ...     ),
     ... )
-    >>> retry_count = 100
-    >>> while retry_count > 0 and not operation.complete:
-    ...     retry_count -= 1
-    ...     time.sleep(10)
-    ...     operation.poll()  # API call
-    >>> operation.complete
-    True
-    >>> for result in operation.results:
+    >>> opresult = operation.result()
+    >>> for result in opresult.results:
     ...     for alternative in result.alternatives:
     ...         print('=' * 20)
     ...         print(alternative.transcript)
--- a/speech/tests/system.py
+++ b/speech/tests/system.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
 import os
-import tempfile
 import time
 import unittest
 import wave
@@ -250,54 +250,36 @@ class TestSpeechClient(unittest.TestCase):
 
     def test_long_running_recognize(self):
 
-        def _build_large_audio_file(wav, reps):
-            ''' From https://stackoverflow.com/questions/46538867/
-                             attributeerror-wave-write-instance-has-no-
-                             attribute-exit
-            '''
-            def __enter__(self):
-                return self
-            def __exit__(self, exc_type, exc_value, traceback):
-                self.close()
+        content = None
+        reps = 5
+        with open(AUDIO_FILE, 'rb') as in_stream:
+            in_wavfile = wave.open(in_stream)
+            params = in_wavfile.getparams()
+            frames = in_wavfile.readframes(in_wavfile.getnframes())
 
-            wave.Wave_read.__exit__ = wave.Wave_write.__exit__ = __exit__
-            wave.Wave_read.__enter__ = wave.Wave_write.__enter__ = __enter__
-            data = []
+        with io.BytesIO() as out_stream:
+            out_wavfile = wave.open(out_stream, 'w')
+            out_wavfile.setparams(params)
+            for rep in range(reps):
+                out_wavfile.writeframes(frames)
+                content = out_stream.getvalue()
 
-            ''' From https://stackoverflow.com/questions/2890703/
-                             how-to-join-two-wav-files-using-python
-            '''
-            for rep in range(0, reps):
-                with wave.open(AUDIO_FILE, 'rb') as wavfile:
-                    data.append([wavfile.getparams(),
-                                 wavfile.readframes(wavfile.getnframes())])
-
-            with wave.open(wav.name, 'wb') as wavfile:
-                wavfile.setparams(data[0][0])
-                for rep in range(0, reps):
-                    wavfile.writeframes(data[rep][1])
-            return wav.name
-        
-        wav = tempfile.NamedTemporaryFile()
-        reps = 15
-        audio_file = _build_large_audio_file(wav, reps)
         client = speech.SpeechClient()
-        with open(audio_file, 'rb') as audio:
-            content = audio.read()
-            operation = client.long_running_recognize(
-                config=speech.types.RecognitionConfig(
-                    encoding='LINEAR16',
-                    language_code='en-US',
-                ),
-                audio=speech.types.RecognitionAudio(content=content)
-            )
-        
+
+        operation = client.long_running_recognize(
+            config=speech.types.RecognitionConfig(
+                encoding='LINEAR16',
+                language_code='en-US',
+            ),
+            audio=speech.types.RecognitionAudio(content=content)
+        )
+
         try:
-            opresult = operation.result()
+            op_result = operation.result()
         except GaxError:
             self.fail()
-        self.assertGreater(len(opresult.results), 0)
-        for result in opresult.results:
+        self.assertGreater(len(op_result.results), 0)
+        for result in op_result.results:
             self.assertGreater(len(result.alternatives), 0)
             for alternative in result.alternatives:
                 self.assertEqual(alternative.transcript.lower(),
--- a/speech/tests/system.py
+++ b/speech/tests/system.py
@@ -249,9 +249,8 @@ class TestSpeechClient(unittest.TestCase):
                 self._check_results(results.alternatives)
 
     def test_long_running_recognize(self):
-
         content = None
-        reps = 5
+        repetitions = 5
         with open(AUDIO_FILE, 'rb') as in_stream:
             in_wavfile = wave.open(in_stream)
             params = in_wavfile.getparams()
@@ -260,7 +259,7 @@ class TestSpeechClient(unittest.TestCase):
         with io.BytesIO() as out_stream:
             out_wavfile = wave.open(out_stream, 'w')
             out_wavfile.setparams(params)
-            for rep in range(reps):
+            for _ in range(repetitions):
                 out_wavfile.writeframes(frames)
                 content = out_stream.getvalue()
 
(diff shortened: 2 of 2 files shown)


speech/tests/system.py
```diff
@@ -245,3 +247,59 @@ def test_stream_recognize_single_utterance(self):
...
+                self.close()
+
+            wave.Wave_read.__exit__ = wave.Wave_write.__exit__ = __exit__
+            wave.Wave_read.__enter__ = wave.Wave_write.__enter__ = __enter__
+            data = []
+
+            ''' From https://stackoverflow.com/questions/2890703/
+                             how-to-join-two-wav-files-using-python
+            '''
+            for rep in range(0, reps):
```
tseaver: There is no need to loop over opening the file:  just open it once and read the values (note that I open the stream explicitly first, to ensure the close semantics you are aiming for above.

```python
        with open(AUDIO_FILE, 'rb') as in_stream:
            with wave.open(in_stream) as in_wavfile:
                params = in_wavfile.getparams()
                frames = in_wavfile.readframes(wavefile.getnframes())
```

For the temporary, you could just use an `io.BytesIO` instance, and then use `params` and `frames` to construct the big wav

```python

        with io.BytesIO() as out_stream:
            with wave.open(out_stream, 'w') as out_wavfile:
                out_wavfile.setparams(params)
                for rep in range(15):
                    out_wavfile.writeframes(frames)
            content = out_stream.getvalue()
```

docs/speech/index.rst
```diff
@@ -55,14 +54,8 @@ See: `Speech Asynchronous Recognize`_
...
     ... )
-    >>> retry_count = 100
-    >>> while retry_count > 0 and not operation.complete:
-    ...     retry_count -= 1
-    ...     time.sleep(10)
-    ...     operation.poll()  # API call
-    >>> operation.complete
-    True
-    >>> for result in operation.results:
+    >>> opresult = operation.result()
```
dhermes: The "word boundary" should be evident in variable names, so here it'd be `op_result`.

speech/tests/system.py
```diff
@@ -245,3 +247,59 @@ def test_stream_recognize_single_utterance(self):
...
+                             attributeerror-wave-write-instance-has-no-
+                             attribute-exit
+            '''
+            def __enter__(self):
+                return self
+            def __exit__(self, exc_type, exc_value, traceback):
+                self.close()
+
+            wave.Wave_read.__exit__ = wave.Wave_write.__exit__ = __exit__
+            wave.Wave_read.__enter__ = wave.Wave_write.__enter__ = __enter__
```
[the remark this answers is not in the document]
dhermes: I agreed. Use` mock.patch`.

speech/tests/system.py
chemelnucfin: so I'm getting a problem that if I mock and patch it I'm not getting a large audio file, I'll look at it again later.

docs/speech/index.rst
chemelnucfin: fixed.

chemelnucfin: i'll upload the fix when i fix the mock issue.

speech/tests/system.py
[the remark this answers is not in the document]
tseaver: I would rather avoid the patching altogether by opening the file separately, as I suggest below.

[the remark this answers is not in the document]
chemelnucfin: Could I subclass it?  The patching was for the content manager to open the wave file.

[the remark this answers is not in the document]
tseaver: @chemelnucfin Do you see my example below?  It doesn't need patching, because it passes an already-opened file object `wave.open()`.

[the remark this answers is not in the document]
chemelnucfin: The context manager doesn't exist in python 2.7, but works in python 3.6

chemelnucfin:
@dhermes @tseaver Ready to review, thanks

For the future, what is the best practice to rebase and push?  It seems like if I rebase against master I'd have new commit ids and I'd have to force push, but if I don't rebase I might get outdated files like when Danny removed an appveyor file and my commits didn't look at it.

dhermes:
> rebase against master

You'd be rebase-ing **on top** of master and the commits in your PR would be based on your `git merge-base` with `master`. If it's still unclear and you want to get a handle on it, you can create a repo of your own and send yourself some PRs.

speech/tests/system.py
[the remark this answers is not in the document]
tseaver: > The context manager doesn't exist in python 2.7, but works in python 3.6

OK.  We *still* don't need to use it to get the file closed if we wrap the `wave.open` call inside the regular `file.open` / 'io.BytesIO` call:

```python
        with io.BytesIO() as out_stream:
            out_wavfile = wave.open(out_stream, 'w')
            out_wavfile.setparams(params)
            for rep in range(15):
                out_wavfile.writeframes(frames)
        content = out_stream.getvalue()
```

tseaver:
> It seems like if I rebase against master I'd have new commit ids and I'd have to force push

Yes, exactly.  We prefer the "rebase + force push" model for PRs over the "merge and normal push" model, because it keeps the commits within the PR focused on its purpose.

speech/tests/system.py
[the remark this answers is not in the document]
chemelnucfin: Would out_wavfile get cleaned up if an exception is thrown though?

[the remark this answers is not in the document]
tseaver: @chemelnucfin That example creates a wrapper around an in-memory `BytesIO` instance:  there is nothing to be cleaned up.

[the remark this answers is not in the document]
chemelnucfin: ok, but is there anything wrong with my approach?  Otherwise would I have to close the wave.open somewhere?  And is that only for the out_stream or both in_stream and out_stream? thanks.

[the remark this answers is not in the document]
tseaver: @chemelnucfin, patching `wave.Wave_{read,write}` is intrinsically poor practice (your code does nothing to restore the original values, for instance), especially if there is a simpler alternative.  My example avoids needing to open (and therefore close / clean up) a tempfile, and is therefore both simpler to understand and less likely to break.

chemelnucfin:
@dhermes @tseaver Ready for review.  Thanks.

dhermes:
> Ready for review. Thanks.

@chemelnucfin We typically say PTAL (== "please take another look"). Message received either way, just a note about "lingo".

speech/tests/system.py
```diff
@@ -245,3 +247,41 @@ def test_stream_recognize_single_utterance(self):
             for results in self._make_streaming_request(
                     file_obj, single_utterance=False):
                 self._check_results(results.alternatives)
+
+    def test_long_running_recognize(self):
+
```
dhermes: @chemelnucfin Please remove this empty line and just "start the code"

speech/tests/system.py
```diff
@@ -245,3 +247,41 @@ def test_stream_recognize_single_utterance(self):
...
+        content = None
+        reps = 5
```
dhermes: Maybe I am n00b but I don't know what `reps` is supposed to mean. In the future, please use "full words" unless it's painfully clear. But, as you'll see below, I recommend dropping this variable all together.

speech/tests/system.py
```diff
@@ -245,3 +247,41 @@ def test_stream_recognize_single_utterance(self):
...
+        with open(AUDIO_FILE, 'rb') as in_stream:
+            in_wavfile = wave.open(in_stream)
+            params = in_wavfile.getparams()
+            frames = in_wavfile.readframes(in_wavfile.getnframes())
+
+        with io.BytesIO() as out_stream:
+            out_wavfile = wave.open(out_stream, 'w')
+            out_wavfile.setparams(params)
+            for rep in range(reps):
```
dhermes: 1. Just hard-code `5` here instead of having `reps`
1. Since `rep` is unused, it's considered "good form" to use `_` as the variable name: `for _ in range(5):`

speech/tests/system.py
```diff
@@ -245,3 +247,41 @@ def test_stream_recognize_single_utterance(self):
...
+                encoding='LINEAR16',
+                language_code='en-US',
+            ),
+            audio=speech.types.RecognitionAudio(content=content)
+        )
+
+        try:
+            op_result = operation.result()
+        except GaxError:
+            self.fail()
```
dhermes: Please remove this `try/except` and just let the test fail with a `GaxError`

speech/tests/system.py
```diff
@@ -245,3 +247,41 @@ def test_stream_recognize_single_utterance(self):
...
+        self.assertGreater(len(op_result.results), 0)
+        for result in op_result.results:
+            self.assertGreater(len(result.alternatives), 0)
+            for alternative in result.alternatives:
+                self.assertEqual(alternative.transcript.lower(),
+                                 ("hello thank you for using google cloud "
+                                  "platform " *reps)[:-1])
```
dhermes: Where does this "hello thank you..." string come from?

speech/tests/system.py
```diff
@@ -245,3 +247,37 @@ def test_stream_recognize_single_utterance(self):
...
+        )
+
+        op_result = operation.result()
+        wav_transcription = "hello thank you for using google cloud platform "
+        self.assertGreater(len(op_result.results), 0)
+        for result in op_result.results:
+            self.assertGreater(len(result.alternatives), 0)
+            for alternative in result.alternatives:
+                self.assertEqual(alternative.transcript.lower(),
+                                 (wav_transcription * repetitions)[:-1])
```
dhermes: @chemelnucfin Why the use of `[:-1]`?

speech/tests/system.py
chemelnucfin: ok.

speech/tests/system.py
chemelnucfin: ok.

speech/tests/system.py
chemelnucfin: done.

speech/tests/system.py
chemelnucfin: ok.

speech/tests/system.py
chemelnucfin: moved it to its own variable describing where the string came from.

speech/tests/system.py
chemelnucfin: remove the final space on the transcription.  Should I comment it or make it more clear somehow?

dhermes: Ah I see. Instead I suggest the following.

```python
wav_text = 'hello thank you for using google cloud platform'
responses = [wav_text] * repetitions
transcript = ' '.join(responses)
...
for result in op_result.results:
    self.assertGreater(len(result.alternatives), 0)
    for alternative in result.alternatives:
        self.assertEqual(
            alternative.transcript.lower(), transcript)
```

----

Also, sorry for not mentioning it before but please use single quotes, not double quotes, for string variables (you used double quotes in `wav_transcription`). @tseaver has a bit of nuance for this rule when it comes to exception text, but let's cross that bridge when we get there.

chemelnucfin: I made it more clear what I was doing in the new commit.  thanks.

chemelnucfin: got it.

chemelnucfin:
I added the test to made sure the doc code worked like it should, so I thought it should have gone together.

dhermes:
> I added the test to made sure the doc code worked like it should, so I thought it should have gone together.

I see. That's reasonable I suppose. Just make it clearer in the "long description" for the PR (the short one should stay short, the so-called commit "subject").

Issue #4208, opened 2017-10-16
Add `.devN` suffixes to all version numbers


dhermes:
I propose we add `.devN` suffixes to all version numbers. This way we can accurately reflect, e.g. that `google-cloud-pubsub` depends on `google-cloud-core >= 0.28.0.dev1` rather than a released version of `google-cloud-core`.

Ref: https://snarky.ca/how-i-manage-package-version-numbers/

----

From https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4186#issuecomment-336531198

lukesneeringer:
I suggested something similar to @jonparrott in chat, although I discussed doing a 1.0 release candidate series.

Let's discuss more tomorrow. I really want to avoid "release the world" pain for customers.

duggelz:
Could we add the blog link and/or explanantion to CONTRIBUTING.rst?  I was looking for this explanation, and it took a bit of googling based on half-remembered commit messages.  CC @liyanhui1228

dhermes:
Sure thing @duggelz. You want to send a PR or you want me, @tseaver, @chemelnucfin or somebody else to send it?

duggelz:
Since I asked for this, I can write something up, sometime next week.

PR #4282, opened 2017-10-30
Explicitly marking `google-cloud-core` and `google-api-core` as "dev".

dhermes:
This is to make it clear the code is between releases. Towards #4208.

See: https://snarky.ca/how-i-manage-package-version-numbers/

### DIFF
--- a/api_core/setup.py
+++ b/api_core/setup.py
@@ -65,7 +65,7 @@ EXTRAS_REQUIREMENTS = {
 
 setup(
     name='google-api-core',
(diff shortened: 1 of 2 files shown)


dhermes:
I would suggest this become our common practice: immediately after pushing a [release][1] ([also][2]) we follow up with a new PR to "un-release" the version.

[1]: https://pypi.org/project/google-api-core/
[2]: https://pypi.org/project/google-cloud-core/

PR #4283, opened 2017-10-30
Language: Add `classify_text` method (v1).

lukesneeringer:
This PR adds the `classify_text` method currently in the `v1beta2` API version to the `v1` version. This should be held until November 15 (the endpoint will not be live on `v1` until then).


lukesneeringer:
Okay, I tested this and am happy with it (but need to see what is up with CI).

@dhermes, I want to do the following:

  * I want to launch a new beta with these changes imminently (today if possible).
  * Next week (probably), I want to promote this to GA.

How does this interact with the releases you just did? I want to make sure I do not step on your toes.

dhermes:
@lukesneeringer I think it'd be fine to do two releases only (since you don't depend on a new version of `core`):

- `google-cloud-language==0.31.0`
- `google-cloud==0.29.0` (which will have as the only change the dependence on the `0.31.x` series of `language`

lukesneeringer:
@dhermes Sounds good.

PR #4291, opened 2017-10-31
BigQuery: moves Row class out of helpers and updates docstrings

alixhami:
- Imports `Row` in `__init__.py`
- Moves `Row` from `_helpers.py` to `table.py` because it is a public class
- Updates incorrect docstrings for functions returning an `Iterator` of `Row` objects

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -16,9 +16,6 @@
 
 import base64
 import datetime
-import operator
-
-import six
 
 from google.api_core import retry
 from google.cloud._helpers import UTC
@@ -178,64 +175,6 @@ _QUERY_PARAMS_FROM_JSON = dict(_CELLDATA_FROM_JSON)
 _QUERY_PARAMS_FROM_JSON['TIMESTAMP'] = _timestamp_query_param_from_json
 
 
-class Row(object):
-    """A BigQuery row.
-
-    Values can be accessed by position (index), by key like a dict,
-    or as properties.
-
-    :type values: tuple
-    :param values:  the row values
-
-    :type field_to_index: dict
-    :param field_to_index:  a mapping from schema field names to indexes
-    """
-
-    # Choose unusual field names to try to avoid conflict with schema fields.
-    __slots__ = ('_xxx_values', '_xxx_field_to_index')
-
-    def __init__(self, values, field_to_index):
-        self._xxx_values = values
-        self._xxx_field_to_index = field_to_index
-
-    def values(self):
-        return self._xxx_values
-
-    def __getattr__(self, name):
-        i = self._xxx_field_to_index.get(name)
-        if i is None:
-            raise AttributeError('no row field "%s"' % name)
-        return self._xxx_values[i]
-
-    def __len__(self):
-        return len(self._xxx_values)
-
-    def __getitem__(self, key):
-        if isinstance(key, six.string_types):
-            i = self._xxx_field_to_index.get(key)
-            if i is None:
-                raise KeyError('no row field "%s"' % key)
-            key = i
-        return self._xxx_values[key]
-
-    def __eq__(self, other):
-        if not isinstance(other, Row):
-            return NotImplemented
-        return(
-            self._xxx_values == other._xxx_values and
-            self._xxx_field_to_index == other._xxx_field_to_index)
-
-    def __ne__(self, other):
-        return not self == other
-
-    def __repr__(self):
-        # sort field dict by value, for determinism
-        items = sorted(self._xxx_field_to_index.items(),
-                       key=operator.itemgetter(1))
-        f2i = '{' + ', '.join('%r: %d' % i for i in items) + '}'
-        return 'Row({}, {})'.format(self._xxx_values, f2i)
-
-
 def _field_to_index_mapping(schema):
     """Create a mapping from schema field name to index of field."""
     return {f.name: i for i, f in enumerate(schema)}
@@ -270,6 +209,8 @@ def _row_tuple_from_json(row, schema):
 
 def _rows_from_json(values, schema):
     """Convert JSON row data to rows with appropriate types."""
+    from google.cloud.bigquery import Row
+
     field_to_index = _field_to_index_mapping(schema)
     return [Row(_row_tuple_from_json(r, schema), field_to_index)
             for r in values]
@@ -505,6 +446,8 @@ def _item_to_row(iterator, resource):
     :rtype: :class:`~google.cloud.bigquery.Row`
     :returns: The next row in the page.
     """
+    from google.cloud.bigquery import Row
+
     return Row(_row_tuple_from_json(resource, iterator.schema),
                iterator._field_to_index)
 
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -209,7 +209,6 @@ def _row_tuple_from_json(row, schema):
 
 def _rows_from_json(values, schema):
     """Convert JSON row data to rows with appropriate types."""
-    # TODO(alixh) remove circular import
     from google.cloud.bigquery import Row
 
     field_to_index = _field_to_index_mapping(schema)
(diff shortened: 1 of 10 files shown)


jonparrott:
@dhermes can you take a look at this as well?

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -270,6 +209,8 @@ def _row_tuple_from_json(row, schema):
 
 def _rows_from_json(values, schema):
     """Convert JSON row data to rows with appropriate types."""
+    from google.cloud.bigquery import Row
```
dhermes: I am :-1: on this circular import, but hesitate to tell you "you must fix it". Though I wish you would.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1172,10 +1172,10 @@ def query_rows(
 
         :rtype: :class:`~google.api_core.page_iterator.Iterator`
         :returns:
-            Iterator of row data :class:`tuple`s. During each page, the
-            iterator will have the ``total_rows`` attribute set, which counts
-            the total number of rows **in the result set** (this is distinct
-            from the total number of rows in the current page:
+            Iterator of row data :class:`~google.cloud.bigquery.Row`s. During
```
dhermes: Didn't @tswast change the documentation so that objects were documented in the module where they are defined? IOW, this should be `~google.cloud.bigquery.table.Row`

Also, I'm fairly certain this won't render correctly with Sphinx unless there is a separator before the `s` (like a hyphen):

```
:class:`~google.cloud.bigquery.Row`-s
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -759,3 +760,61 @@ def __init__(self, resource):
...
+            self._xxx_field_to_index == other._xxx_field_to_index)
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __repr__(self):
+        # sort field dict by value, for determinism
+        items = sorted(self._xxx_field_to_index.items(),
+                       key=operator.itemgetter(1))
+        f2i = '{' + ', '.join('%r: %d' % i for i in items) + '}'
```
dhermes: I realize this is just copy-pasted, but please change the variable name `i` to `item`.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -759,3 +760,61 @@ def __init__(self, resource):
...
+        if i is None:
+            raise AttributeError('no row field "%s"' % name)
+        return self._xxx_values[i]
+
+    def __len__(self):
+        return len(self._xxx_values)
+
+    def __getitem__(self, key):
+        if isinstance(key, six.string_types):
+            i = self._xxx_field_to_index.get(key)
```
dhermes: I realize this is just copy-pasted, but please change the variable name `i` to `value`.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -759,3 +760,61 @@ def __init__(self, resource):
...
+            if i is None:
+                raise KeyError('no row field "%s"' % key)
```
dhermes: This is room for a bug, please pass an explicit tuple:

```python
raise KeyError('no row field "%s"' % (key,))
```

or better yet ditch the percent formatting

```python
raise KeyError('no row field {!r}'.format(key))
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -759,3 +760,61 @@ def __init__(self, resource):
...
+        self._xxx_values = values
+        self._xxx_field_to_index = field_to_index
+
+    def values(self):
+        return self._xxx_values
+
+    def __getattr__(self, name):
+        i = self._xxx_field_to_index.get(name)
+        if i is None:
+            raise AttributeError('no row field "%s"' % name)
```
dhermes: This is room for a bug, please pass an explicit tuple:

```python
raise AttributeError('no row field "%s"' % (key,))
```

or better yet ditch the percent formatting

```python
raise AttributeError('no row field {!r}'.format(key))
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -759,3 +760,61 @@ def __init__(self, resource):
...
+
+    # Choose unusual field names to try to avoid conflict with schema fields.
+    __slots__ = ('_xxx_values', '_xxx_field_to_index')
+
+    def __init__(self, values, field_to_index):
+        self._xxx_values = values
+        self._xxx_field_to_index = field_to_index
+
+    def values(self):
+        return self._xxx_values
```
dhermes: This means the caller can mutate your **internal** representation if it is a mutable type. Do you want that? (I suppose that's the users' "fault", since the docstring says `values` must be a `tuple`.

bigquery/tests/unit/test_client.py
```diff
@@ -2182,8 +2182,7 @@ def _row_data(row):
         self.assertEqual(req['data'], SENT)
 
     def test_create_rows_w_list_of_Rows(self):
-        from google.cloud.bigquery._helpers import Row
-        from google.cloud.bigquery.table import Table, SchemaField
+        from google.cloud.bigquery.table import Table, SchemaField, Row
```
dhermes: This should be 3 imports. From the [Google style guide][1]:

> Imports should be on separate lines.
> E.g.:
> ```
> Yes: import os
>      import sys
> No:  import os, sys
> ```

[1]: https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting

bigquery/tests/unit/test_client.py
```diff
@@ -2720,8 +2719,7 @@ def test_query_rows_w_api_error(self):
     def test_list_rows(self):
         import datetime
         from google.cloud._helpers import UTC
-        from google.cloud.bigquery.table import Table, SchemaField
-        from google.cloud.bigquery._helpers import Row
+        from google.cloud.bigquery.table import Table, SchemaField, Row
```
dhermes: Ditto on imports

bigquery/tests/unit/test_table.py
```diff
@@ -751,3 +751,25 @@ def test__row_from_mapping_w_schema(self):
...
             ('Phred Phlyntstone', 32, ['red', 'green'], None))
+
+
+class TestRow(unittest.TestCase):
+
+    def test_row(self):
+        from google.cloud.bigquery.table import Row
+
+        VALUES = (1, 2, 3)
+        r = Row(VALUES, {'a': 0, 'b': 1, 'c': 2})
```
dhermes: s/`r`/`row`. From [guide][1]

> Names to Avoid
> - single character names except for counters or iterators

[1]: https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -209,6 +209,7 @@ def _row_tuple_from_json(row, schema):
 
 def _rows_from_json(values, schema):
     """Convert JSON row data to rows with appropriate types."""
+    # TODO(alixh) remove circular import
```
dhermes: Please don't use a `TODO`, instead use the issue tracker.

bigquery/google/cloud/bigquery/_helpers.py
alixhami: I created a new issue for this: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4305

bigquery/google/cloud/bigquery/client.py
alixhami: Fixed the references in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4291/commits/0a4fb77d565b3edfdfc76e6041a06dad92d13e9f

bigquery/google/cloud/bigquery/table.py
alixhami: Fixed these in f086329

bigquery/google/cloud/bigquery/table.py
alixhami: Fixed all the variable naming in f086329

bigquery/tests/unit/test_client.py
alixhami: Fixed these issues in 8cf4f53

bigquery/google/cloud/bigquery/_helpers.py
alixhami: Reverted this commit in cac00de and created a issue #4305

bigquery/google/cloud/bigquery/table.py
alixhami: Talked to @jonparrott about this. I think we can let the user accept this risk if they use the class improperly.

PR #4298, opened 2017-10-31
Video Intelligence v1

lukesneeringer:
This PR adds the `v1` endpoint to video intelligence.

This PR should *not* be merged until just before the November 15 rollout.


dhermes:
@lukesneeringer So I am good to cut a `videointelligence` release today with the deps only changes?

lukesneeringer:
Yes please. This is going to be in limbo for a week or so; staged here so DPE can write/update samples.

PR #4299, opened 2017-10-31
Marking all remaining versions as "dev".

### LINKED ISSUE #4208: Add `.devN` suffixes to all version numbers

dhermes:
This is to make it clear the code is between releases. Any code that relies on a **new** feature (e.g. of `google-api-core`) will then be able to **explicitly** make this clear by using the lower bound of the `devN` version.

Fixes #4208.

See: https://snarky.ca/how-i-manage-package-version-numbers/

### DIFF
--- a/bigtable/setup.py
+++ b/bigtable/setup.py
@@ -58,7 +58,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-bigtable',
-    version='0.28.0',
+    version='0.28.1.dev1',
     description='Python Client for Google Cloud Bigtable',
     long_description=README,
     namespace_packages=[
--- a/datastore/setup.py
+++ b/datastore/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-datastore',
-    version='1.4.0',
+    version='1.4.1.dev1',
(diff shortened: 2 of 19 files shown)


duggelz:
Can we document this in a README or something?  Also, is the whole release process documented somewhere (since I'll want to release opencensus-python soon and hope to shamelessly copy the google-cloud-python process)?

dhermes:
> Can we document this in a README or something? 

I think the `CONTRIBUTING.rst` is the right place?

> is the whole release process documented somewhere (since I'll want to release opencensus-python soon and hope to shamelessly copy the google-cloud-python process)?

Not really (but maybe a little bit in `CONTRIBUTING.rst`?). It's mostly done by our scripts, which can detect when a tag has been pushed and manually parse the version from the tag.

I am happy to chat about packaging "things to care about" if you like Doug.

PR #4308, opened 2017-11-01
Trace: #4307 setup config

### LINKED ISSUE #4307: google-cloud-trace wheels are **not** universal
This is because of a missing `trace/setup.cfg`. For example, in the latest [release][1] the file uploaded corresponds to the Python that built the wheel: `google_cloud_trace-0.16.0-py3-none-any.whl` (it should end with `-py2.py3-none-any.whl`.

@lukesneeringer I am assigning to you because I think this is a codegen issue. There is a simple fix to add a two line file `setup.cfg`:

```
[bdist_wheel]
universal = 1
```

[1]: https://pypi.org/project/google-cloud-trace/0.16.0/

chemelnucfin:
Hi @lukesneeringer @dhermes Does this work? #4307 Or will I need to add some more stuff?  I'm not familiar with this setup thing.

### DIFF
--- a/trace/setup.cfg
+++ b/trace/setup.cfg
@@ -1,0 +1,3 @@
+[bdist_wheel]
+universal = 1
+


trace/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
+universal = 1
```
dhermes: You need a newline at the end of this file (I'd recommend setting up your editor so it adds one by default).

Otherwise LGTM.

@chemelnucfin I'd rather not close out the issue, as I believe there is an issue in codegen. But feel free to merge this once you've resolved the newline issue,

dhermes:
@chemelnucfin You added **two** extra newline characters, not one.

```python
>>> with open('trace/setup.cfg', 'r') as file_obj:
...     content = file_obj.read()
... 
>>> content
'[bdist_wheel]\nuniversal = 1\n\n'
```

What you had before was **no** trailing newline at all

```python
'[bdist_wheel]\nuniversal = 1'
```

and what I was asking for was a single trailing newline

```python
'[bdist_wheel]\nuniversal = 1\n'
```

chemelnucfin:
Ok, I think it didn't show with the diff, so I added an extra one.  I'll
fix it right now.

PR #4315, opened 2017-11-01
Enable anonymous access to blobs in public buckets

### LINKED ISSUE #4279: Storage: Provide credential-free access to public-access buckets
Per the [GCS authentication doc](https://cloud.google.com/storage/docs/authentication):

> Objects are anonymously accessible if the `allUsers` group has `READ` permission.

Add support for such access to the storage client.  Two strategies seem possible:

- Add a classmethod factory to the `Client` class which bypasses inferring credentials from the user environment.
- Add a custom `AnonymousCredentials` singleton which can be passed to the `Client` constructor.

We don't need to enforce any limits on the access to buckets / objects retrieved from the "anonymous" client:  the back-end will return `UNAUTHORIZED`.

System tests can use the `gcp-public-data-landsat` bucket, if desired (hopefully CI throttling won't be an issue).

tseaver:
/cc @jskeet 

Closes #4279

### DIFF
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -15,6 +15,8 @@
 """Client for interacting with the Google Cloud Storage API."""
 
 
+from google.auth.credentials import AnonymousCredentials
+
 from google.api_core import page_iterator
 from google.cloud._helpers import _LocalStack
 from google.cloud.client import ClientWithProject
@@ -59,6 +61,22 @@ class Client(ClientWithProject):
                                      _http=_http)
         self._connection = Connection(self)
         self._batch_stack = _LocalStack()
+
+    @classmethod
+    def create_anonymous_client(cls):
+        """Factory: return client with anonymous credentials.
+
+        .. note::
+
+           Such a client has only limited access to "public" buckets:
+           listing their contents and downloading their blobs.
+
+        :rtype: :class:`google.cloud.storage.client.Client`
+        :returns: Instance w/ anonymous credentials and no project.
+        """
+        client = cls(project='<none>', credentials=AnonymousCredentials())
+        client.project = None
+        return client
 
     @property
     def _connection(self):
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -53,7 +53,7 @@ SETUP_BASE = {
 REQUIREMENTS = [
     'google-cloud-core >= 0.28.0, < 0.29dev',
     'google-api-core >= 0.1.1, < 0.2.0dev',
-    'google-auth >= 1.0.0',
+    'google-auth >= 1.2.0',
     'google-resumable-media >= 0.3.1',
     'requests >= 2.18.0',
 ]
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -952,3 +952,15 @@ class TestStorageNotificationCRUD(unittest.TestCase):
             self.assertEqual(notifications[0].topic_name, self.TOPIC_NAME)
         finally:
             notification.delete()
+
+
+class TestAnonymousClient(unittest.TestCase):
+
+    PUBLIC_BUCKET = 'gcp-public-data-landsat'
+
+    def test_access_to_public_bucket(self):
+        anonymous = storage.Client.create_anonymous_client()
+        bucket = anonymous.bucket(self.PUBLIC_BUCKET)
+        blob, = bucket.list_blobs(max_results=1)
+        with tempfile.TemporaryFile() as stream:
+            blob.download_to_file(stream)
--- a/storage/tests/unit/test_client.py
+++ b/storage/tests/unit/test_client.py
@@ -68,11 +68,24 @@ class TestClient(unittest.TestCase):
         CREDENTIALS = _make_credentials()
 
         client = self._make_one(project=PROJECT, credentials=CREDENTIALS)
+
         self.assertEqual(client.project, PROJECT)
         self.assertIsInstance(client._connection, Connection)
         self.assertIs(client._connection.credentials, CREDENTIALS)
         self.assertIsNone(client.current_batch)
         self.assertEqual(list(client._batch_stack), [])
+
+    def test_create_anonymous_client(self):
+        from google.auth.credentials import AnonymousCredentials
+        from google.cloud.storage._http import Connection
+
+        klass = self._get_target_class()
+        client = klass.create_anonymous_client()
+
+        self.assertIsNone(client.project)
+        self.assertIsInstance(client._connection, Connection)
+        self.assertIsInstance(
+            client._connection.credentials, AnonymousCredentials)
 
     def test__push_batch_and__pop_batch(self):
         from google.cloud.storage.batch import Batch


storage/setup.py
```diff
@@ -53,7 +53,7 @@
 REQUIREMENTS = [
     'google-cloud-core >= 0.28.0, < 0.29dev',
     'google-api-core >= 0.1.1, < 0.2.0dev',
-    'google-auth >= 1.0.0',
+    'google-auth >= 1.2.0',
```
dhermes: @jonparrott Should we also update the requirement in `api_core/setup.py`, `bigquery/setup.py` or `spanner/setup.py`?

storage/google/cloud/storage/client.py
```diff
@@ -60,6 +62,22 @@ def __init__(self, project=None, credentials=None, _http=None):
...
+        
+        .. note::
+
+           Such a client has only limited access to "public" buckets:
+           listing their contents and downloading their blobs.
+           
+        :rtype: :class:`google.cloud.storage.client.Client`
+        :returns: Instance w/ anonymous credentials and no project.
+        """
+        client = cls(project='<none>', credentials=AnonymousCredentials())
```
dhermes: @tseaver Is there any reason not to also define `Client.ANONYMOUS = AnonymousCredentials()` so users can re-use?

storage/tests/system.py
```diff
@@ -952,3 +952,17 @@ def test_notification_w_user_project(self):
...
+
+class TestAnonymousClient(unittest.TestCase):
+
+    PUBLIC_BUCKET = 'gcp-public-data-landsat'
+
+    def test_access_to_public_bucket(self):
+        anonymous = storage.Client.create_anonymous_client()
+        bucket = anonymous.bucket(self.PUBLIC_BUCKET)
+        for blob in bucket.list_blobs():
+            print("Downloading blob: {}".format(blob.name))
```
dhermes: We shouldn't have print statements in a test (system or otherwise)

storage/tests/system.py
```diff
@@ -952,3 +952,17 @@ def test_notification_w_user_project(self):
...
+            with tempfile.TemporaryFile() as stream:
+                blob.download_to_file(stream)
+            break  # just use the first one
```
dhermes: Instead, just grab the first from the iterator:

```python
blobs_iter = bucket.list_blobs(max_results=1)
blob = next(blobs_iter)
```

storage/tests/system.py
```diff
@@ -952,3 +952,17 @@ def test_notification_w_user_project(self):
...
+
+    PUBLIC_BUCKET = 'gcp-public-data-landsat'
+
+    def test_access_to_public_bucket(self):
+        anonymous = storage.Client.create_anonymous_client()
+        bucket = anonymous.bucket(self.PUBLIC_BUCKET)
+        for blob in bucket.list_blobs():
+            print("Downloading blob: {}".format(blob.name))
+            with tempfile.TemporaryFile() as stream:
+                blob.download_to_file(stream)
```
dhermes: @tseaver The test should verify **something** other than "we didn't fail"?

storage/setup.py
tseaver: @dhermes This pin is to get the `AnonymousCredentials` feature, which they aren't likely to need.

dhermes: @tseaver I'm aware, thanks. This question doesn't need to block the PR, but is still worth noting.

storage/google/cloud/storage/client.py
tseaver: I figured YAGNI:  originally, I had planned to have the user pass in an `AnonymousCredentials` instance directly, but this method makes that path unimportant.  It also provides a place to document the expected usage pattern / restrictions.

tseaver:
@dhermes 

> Don't you need more tests for methods that do and should fail?

The back-end enforces the policy by raising 401s:  why bother testing them?

storage/google/cloud/storage/client.py
dhermes: Sphinx allows documentation on `attr`s, so we could also document expected usage pattern / restrictions on the class constant:

```python
class Client(...):
    ANONYMOUS = AnonymousCredentials()
    """Unauthenticated credentials to be used for public buckets.

    For example:

    .. code-block:: python

       >>> import foo
    ...
    """
```

dhermes:
> The back-end enforces the policy by raising 401s: why bother testing them?

I sympathize with the sentiment, but am just weary of the cases we haven't thought about.

storage/tests/system.py
tseaver: Oops. ea8244b

storage/tests/system.py
tseaver: The only interesting thing is that we can list objects and download them: there is nothing special about the data to be asserting.

storage/tests/system.py
```diff
@@ -961,8 +961,6 @@ class TestAnonymousClient(unittest.TestCase):
     def test_access_to_public_bucket(self):
         anonymous = storage.Client.create_anonymous_client()
         bucket = anonymous.bucket(self.PUBLIC_BUCKET)
-        for blob in bucket.list_blobs():
-            print("Downloading blob: {}".format(blob.name))
-            with tempfile.TemporaryFile() as stream:
-                blob.download_to_file(stream)
-            break  # just use the first one
+        blob, = list(bucket.list_blobs(max_results=1))
```
dhermes: @tseaver You can tuple-unpack a generator without `list()`. e.g.

```python
>>> gen = (i * i for i in (1, 2, 3))
>>> gen
<generator object <genexpr> at 0x7f1a3ab818e0>
>>> a, b, c = gen
>>> a
1
>>> b
4
>>> c
9
```

storage/tests/system.py
tseaver: That no work:

```python
    blob = next(bucket.list_blobs(max_results=1))
TypeError: 'HTTPIterator' object is not an iterator
```

ea8244b does:

```python
blob, = list(bucket.list_blobs(max_results=1))
```

storage/tests/system.py
dhermes: Well if the data is deterministic, we can just assert the content is what we expect.

storage/google/cloud/storage/client.py
jonparrott: Seems like unnecessary indirection to me.

storage/google/cloud/storage/client.py
```diff
@@ -60,6 +62,22 @@ def __init__(self, project=None, credentials=None, _http=None):
         self._connection = Connection(self)
         self._batch_stack = _LocalStack()
 
+    @classmethod
+    def create_anonymous_client(cls):
+        """Factory: return client w/ anonymous credentials.
```
jonparrott: Please use `with` (throughout). Our (internal) documentation style guide discourages these kinds of abbreviations.

storage/tests/unit/test_client.py
```diff
@@ -68,12 +68,27 @@ def test_ctor_connection_type(self):
...
+        from google.cloud.storage._http import Connection
+
+        klass = self._get_target_class()
+        client = klass.create_anonymous_client()
+
+        self.assertIsNone(client.project)
+        self.assertIsInstance(client._connection, Connection)
+        self.assertIsInstance(
+            client._connection.credentials, AnonymousCredentials)
+        self.assertIsNone(client.current_batch)
```
jonparrott: These last two assertions seem unnecessary in context?

tseaver: 93eec3e

storage/google/cloud/storage/client.py
tseaver: 93eec3e

storage/google/cloud/storage/client.py
dhermes: It may be useful, especially if we want to check for identity:

```python
if self._credentials is Client.ANONYMOUS:
    ...
```

Though I suppose an `isinstance()` check would be fine too.

jonparrott: Yeah it's not a singleton, so I would never want us to rely on a specific anonymouscredentials instance.

lukesneeringer:
@dhermes Apparently...

> ValueError: invalid version number '0.28.1.dev1'
> Command exited with code 1

...but only on Windows.

tseaver:
@dhermes, @lukesneeringer That spelling is correct according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#developmental-releases).  We could try `dev0` (we won't need more than one) or `dev` (an allowed abbreviation).

tseaver:
Rebased / semi-squashed to fix Appveyor, tidy up.

dhermes:
The AppVeyor issue was fixed in #4313 (I just removed `bdist_msi` from our build since we never used the artifacts). @jonparrott and I discussed, there seems to be a bug in `bdist_msi`, it uses [`StrictVersion`][1].

[1]: https://github.com/python/cpython/blob/v2.7.14/Lib/distutils/version.py#L63

tseaver:
@dhermes Yup, that was why I rebased on current master.

tseaver:
AFAIK, this PR is ready to merge once CI is green.

PR #4322, opened 2017-11-01
api_core: Add metadata option to wrap_method

eoogbe:
To update GAPIC to use routing headers, `wrap_method` needs to accept more metadata than just the user agent. Example:
```python
metadata = [
    (b'x-google-request-params',
     b'name={}&book.read={}'.format(name, book.read))]

self._create_book(request, retry=retry, timeout=timeout, metadata=metadata)
```

### DIFF
--- a/api_core/google/api_core/gapic_v1/method.py
+++ b/api_core/google/api_core/gapic_v1/method.py
@@ -94,16 +94,16 @@ class _GapicCallable(object):
         timeout (google.api_core.timeout.Timeout): The default timeout
             for the callable. If ``None``, this callable will not specify
             a timeout argument to the low-level RPC method by default.
-        user_agent_metadata (Tuple[str, str]): The user agent metadata key and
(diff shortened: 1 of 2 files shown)


jonparrott:
@eoogbe not to be obnoxious, but I have the exact same question about this change as I had with the (same? similar?) feature proposed in https://github.com/googleapis/gax-python/pull/203

Why does this need to occur at *wrap* time as opposed to *invocation* time? Doing this at wrap time adds complexity to this layer but adding metadata at invocation time is a feature that already exists. It seems the rationale here is "routing headers" - won't those depend on the RPC request message which isn't known until invocation time?

eoogbe:
@jonparrott I got a little confused at the api change. Can I specify metadata once the method is wrapped method?

jonparrott:
Yep! The wrapper passes through all keyword args to the wrapped method so
you can just pass it in like you would when invoking any old grpc method. I
can throw up a code snippet tomorrow if needed.

PR #4323, opened 2017-11-01
Comparing `datetime.datetime`-s in trace unit tests.

dhermes:
Doing this because [on Windows][1], when parsed into a protobuf and converted back to a string, trailing zeros are truncated.

[1]: https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-python/build/1.0.2506.master/job/xl4wk678mibxw0wa


dhermes:
Thanks for the review Doug!

PR #4329, opened 2017-11-02
Post-release cleanup for language and `google-cloud`.

dhermes:
In particular:

- Marking the package versions as "dev"
- Updating the `releases.rst` for the packages

(Again, this showcases a "need" for the `releases.rst` files to be checked in some automated way.)


dhermes:
@tseaver @lukesneeringer @jonparrott Can I get an LGTM?

PR #4330, opened 2017-11-02
Datastore: test_transaction minor refactoring

chemelnucfin:
Minor refactoring and just to separate the commits from my other code

### DIFF
--- a/datastore/tests/unit/test_transaction.py
+++ b/datastore/tests/unit/test_transaction.py
@@ -19,6 +19,9 @@ import mock
 
 class TestTransaction(unittest.TestCase):
 
+    def setUp(self):
+        self.project = 'PROJECT'
+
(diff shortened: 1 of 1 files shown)


datastore/tests/unit/test_transaction.py
```diff
@@ -19,6 +19,9 @@
 
 class TestTransaction(unittest.TestCase):
 
+    def setUp(self):
+        self.project = 'PROJECT'
```
dhermes: Please just use a class constant instead of `setUp`. (In general we avoid `setUp` and `tearDown` at all costs because we'd prefer state to be "obvious" like a class constant or local to each test case.)

chemelnucfin:
ok.  Can close it.  I'll also change it in my other commit.

PR #4332, opened 2017-11-02
Marking Bigtable as "dev" after release.

dhermes:


dhermes:
MERP. I realized we also released #4256. I am quite intimidated by the size of the PR but will try to update the `CHANGELOG` in this PR and then also update https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/bigtable-0.28.1

PR #4333, opened 2017-11-02
BigTable: Small docs/hygiene tweaks after #4256.

dhermes:
/cc @justinseinlin

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -15,8 +15,6 @@
 """User-friendly container for Google Cloud Bigtable Table."""
 
 
-import six
-
 from google.api_core.exceptions import Aborted
 from google.api_core.exceptions import DeadlineExceeded
 from google.api_core.exceptions import from_grpc_status
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -426,7 +426,7 @@ class _RetryableMutateRowsWorker(object):
                   sent. These will be in the same order as the ``rows``.
         """
         try:
-            retry(self.__class__._do_mutate_retryable_rows)(self)
+            retry(self._do_mutate_retryable_rows)()
         except (RetryError, ValueError) as err:
             # Upon timeout or sleep generator error, return responses_statuses
             pass
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -17,13 +17,13 @@
 
 import six
 
-from google.api_core.exceptions import RetryError
 from google.api_core.exceptions import Aborted
 from google.api_core.exceptions import DeadlineExceeded
-from google.api_core.exceptions import ServiceUnavailable
 from google.api_core.exceptions import from_grpc_status
-from google.api_core.retry import Retry
+from google.api_core.exceptions import RetryError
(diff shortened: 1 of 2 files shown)


bigtable/google/cloud/bigtable/table.py
```diff
@@ -412,13 +426,14 @@ def __call__(self, retry=DEFAULT_RETRY):
                   sent. These will be in the same order as the ``rows``.
         """
         try:
-            retry(self.__class__._do_mutate_retryable_rows)(self)
+            retry(self._do_mutate_retryable_rows)()
```
dhermes: @justinseinlin Did you have a specific reason for doing it in this hacky way? Or just "if it works, it sticks"?

Also, IIUC @jonparrott would deem `retry=None` as an acceptable input here, so ISTM the code should be:

```python
if retry is None:
    self._do_mutate_retryable_rows()
else:
    retry(self._do_mutate_retryable_rows)()
```

justinseinlin: Yes, we should change it into w/ and w/o the retry. Thanks!

bigtable/google/cloud/bigtable/table.py
```diff
@@ -462,11 +483,15 @@ def _do_mutate_retryable_rows(self):
                 if entry.status.code == 0:
                     self.rows[index].clear()
 
-        assert len(retryable_rows) == num_responses
+        if len(retryable_rows) != num_responses:
+            raise RuntimeError(
+                'Unexpected the number of responses', num_responses,
```
justinseinlin: Did you mean to include the `the` here?

dhermes: I did not, good catch.

bigtable/google/cloud/bigtable/table.py
dhermes: @justinseinlin Do you mind making that change in a separate PR. That will need updated logic with new tests (i.e. it's not a trivial change).

In that PR, you can discuss with @jonparrott if raising a "fake" `ServiceUnavailable` is the "right" way to force a retry from within `_do_mutate_retryable_rows()`.

justinseinlin: Sure. I will make a separate PR with that change.

justinseinlin: Since this PR also affects the same line, I'll wait for this PR to land and start from that.

dhermes: SGTM

bigtable/google/cloud/bigtable/table.py
dhermes: @justinseinlin To avoid making CI restart, you can include this change in your PR? Or I can do once I get an LGTM from a reviewer?

justinseinlin: I'll include this in my PR.

bigtable/google/cloud/bigtable/table.py
justinseinlin: I have the first draft of the change ready. Will do rebase after this PR is merged, and then create a new PR. 

I'll discuss with @jonparrott on the fake error thrown in order to trigger retry. Will look into making an internal exception and updating the predicate of `DEFAULT_RETRY`.

dhermes:
Merging with an LGTM from @tseaver in chat.

PR #4335, opened 2017-11-02
Add support for 'TransactionOptions'

### LINKED ISSUE #4278: Datastore: Add support for new 'TransactionOptions' field
## Overview

Datastore has introduced two transaction options that can be passed when creating a transaction (BeginTransaction RPC):
- A transaction can be declared read-only. A read-only transaction cannot modify data.
- A read-write transaction can provide the ID of a previous transaction.

### Transaction Mode

Transactions can optionally be created in read-only or read-write mode. Specifying the mode of the transaction will allow for Datastore to perform future performance optimizations in upcoming changes.

A read-write transaction can read and modify entities. Read-write transactions that operate on the same data can conflict and suffer from reduced throughput due to contention.

A read-only transaction cannot modify entities; in return they do not contend with other read-write or read-only transactions. Using a read-only transaction for transactions that only read data will potentially improve throughput.

### Transaction Retry

Read-write transactions may fail due to contention. A read-write transaction can be retried by specifying `previous_transaction` when creating the new transaction. 

Specifying `previous_transaction` provides more information that can be used to improve throughput. In particular, if transactional operations `A` and `B` conflict, specifying the `previous_transaction` can help to prevent livelock.  

## Requirements

Users should be able to specify that a transaction is read-only when they create it.
When retrying a read-write transaction, the previous transaction’s ID should be included in the request.
If your client has a `RunInTransaction` method that takes a function and automatically creates and retries the transaction, then this method should handle setting the previous transaction ID. There is no need to expose the option to the user.
If your client requires that users retry themselves, then expose the option.

### Unit Tests

- When the read-only option is passed, the BeginTransactionRequest is
        project_id:"project" transaction_options:<read_only:<> > 
- When the previous transaction ID “tid” is passed, the BeginTransactionRequest is
project_id:"project"
    transaction_options:<read_write:<previous_transaction:"tid" > > 
- If you have a `RunInTransaction function, simulate a retryable failure to verify that previous_transaction is passed on retries.

### Integration Tests

- A read-only transaction can perform reads.
- A read-only transaction fails if it attempts to write. This test currently fails against the production endpoint.

## Resources

- [Public protos](https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/datastore.proto#L352)
- [Go implementation](https://code-review.googlesource.com/c/gocloud/+/18350)
- [C# implementation](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1447)

chemelnucfin:
**Uses #4348 as a base.**

----------

Please look it over and feel free to critique.

Also let me know what I'm missing.

Closes #4278.


datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +282,36 @@ def commit(self):
         finally:
             # Clear our own ID in case this gets accidentally reused.
             self._id = None
+
+    def put(self, entity):
+        """Remember an entity's state to be saved during :meth:`commit`.
```
dhermes: Rather than copy-paste it'd be better to just say this is `Batch.put` PLUS some extra features.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +282,36 @@ def commit(self):
...
+
+        .. note::
+           Any existing properties for the entity will be replaced by those
```
dhermes: It's typically "good form" with Sphinx to have an empty line between the `.. note::` directive and the actual note content. (I realize you copy-pasted from somewhere else, just worth pointing out for your future usage. I also realize other contributors don't follow this / aren't aware of this, but still worth sharing.)

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +282,36 @@ def commit(self):
...
+        :raises: :class:`RuntimeError` if the transaction
+                 is marked ReadOnly
+                 :class:`ValueError` if the batch is not in
+                 progress, if entity has no key assigned, or if the key's
+                 ``project`` does not match ours.
+        """
+        if isinstance(self.options.mode, self.options.ReadWrite):
+            super(Transaction, self).put(entity)
+        else:
+            raise RuntimeError("Transaction is read only")
```
dhermes: I'd prefer a `ValueError` here, `RuntimeError` is reserved for "we don't really know what to call it".

datastore/tests/unit/test_transaction.py
```diff
@@ -15,7 +15,30 @@
 import unittest
 
 import mock
+from google.cloud.datastore.transaction import TransactionOptions
 
+class TestTransactionOptions(unittest.TestCase):
```
dhermes: There should be two empty lines between imports and your defined members (like classes and functions). I recommend reading:

- https://google.github.io/styleguide/pyguide.html
- https://www.python.org/dev/peps/pep-0008/

datastore/tests/unit/test_transaction.py
```diff
@@ -15,7 +15,30 @@
 import unittest
 
 import mock
+from google.cloud.datastore.transaction import TransactionOptions
```
dhermes: We follow a rule in our **unit** tests (but not system) that imports of the code being tested happen at run time, not import time. So you'd move this import into `_get_target_class()` so that the code only gets imported when the test is run.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -17,6 +17,48 @@
 from google.cloud.datastore.batch import Batch
 
 
+class TransactionOptions(object):
```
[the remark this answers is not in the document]
chemelnucfin: I thought about that, and then decided to follow protobuf.  This could also be better in the case of a change upstream to mirror what they do.  I'll do whatever you guys decide on though.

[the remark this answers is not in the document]
dhermes: I'm not sure what you mean when you say "follow protobuf". As I see it, there is a class with 30+ lines of code accomplishing what a keyword argument could. It doesn't seem like a win for usability.

As for future-proofing it ("could also be better in the case of a change upstream") that is definitely a possibility. If they added **many** arguments, then having an `option=` keyword (vs. a `read_only=`) keyword would be beneficial. But that's a big if and not one worth taking into this design decision IMO.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -17,6 +17,48 @@
...
+    previous transaction id to be stored.
+
+    :type readonly: `boolean`
+    :param readonly: indicates whether the transaction is readonly (True) or
+                     readwrite (False)
+
+    :type previous_transaction: `bytes`
+    :param previous_transaction: The transaction identifier of the transaction
+                                 being retried.
+
```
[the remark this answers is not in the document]
chemelnucfin: ok.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -17,6 +17,48 @@
...
+    """
+
+    class ReadWrite(object):
```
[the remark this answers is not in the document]
chemelnucfin: I agree and normally don't do that.

datastore/google/cloud/datastore/transaction.py
chemelnucfin: ok.

datastore/google/cloud/datastore/transaction.py
chemelnucfin: ok.

datastore/google/cloud/datastore/transaction.py
chemelnucfin: I thought about ValueError, but it's not that I'm passing a bad value to put.  But I shouldn't be able to use this function no matter what value I give it.  I didn't think it fit any other standard errors, and IOError doesn't seem right, thus RuntimeError was the best I thought.  I don't mind changing it though.

datastore/tests/unit/test_transaction.py
chemelnucfin: ok.

datastore/tests/unit/test_transaction.py
chemelnucfin: ok, I thought nox lint would catch it.

dhermes: Good point, not sure why it didn't. Still worth knowing.

datastore/google/cloud/datastore/transaction.py
dhermes: The "value" in error would be the value of `self.options` (or `self._read_only` if you go that route).

datastore/google/cloud/datastore/transaction.py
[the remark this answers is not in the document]
chemelnucfin: But my thought was that if protobuf just wanted a boolean, it would have specified a boolean.  Doesn't matter though.  I don't mind changing it.

chemelnucfin:
Thank you for your comments.
Should I incorporate these comments now or wait until @lukesneeringer regenerates the library?

dhermes:
> Should I incorporate these comments now or wait until lukesneeringer regenerates the library?

IMO it's not worth working on this until you can actually send the payload to the live server, so I would wait.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -17,6 +17,48 @@
...
 
 
+class TransactionOptions(object):
+    """Options for Transactions
+
+    TransactionOptions stores the mode to indicate whether it is ReadOnly or
+    ReadWrite.  In the case of ReadWrite an optional parameter allows the
+    previous transaction id to be stored.
+
+    :type readonly: `boolean`
```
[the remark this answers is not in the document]
chemelnucfin: ok

lukesneeringer:
> @chemelnucfin Thanks for the work but it seems this effort is futile until support for `TransactionOptions` is in the `gapic-google-cloud-datastore-v1` library (since the underlying GAX / proto-over-HTTP clients will never send the right payload).
> 
> That requires @lukesneeringer (or someone on his team) to regenerate the [library][1].
> 
> [1]: https://github.com/googleapis/api-client-staging/tree/master/generated/python/gapic-google-cloud-datastore-v1

I am happy to do that.

lukesneeringer:
Auto-gen updated. In theory, this should now work against the live service (and we should test to ensure that).

chemelnucfin:
See #4357

PR #4336, opened 2017-11-03
api_core: Add routing header

eoogbe:
Example:
```python
routing_header = google.api_core.gapic_v1.routing_header(
            [('name', name), ('book.read', book.read)])
self._create_book(request, retry=retry, timeout=timeout, metadata=[routing_header])
```

### DIFF
--- a/api_core/google/api_core/gapic_v1/routing_header.py
+++ b/api_core/google/api_core/gapic_v1/routing_header.py
@@ -16,9 +16,11 @@
 
 METADATA_KEY = 'x-goog-header-params'
 
+
 def to_routing_header(params):
     """Returns the routing header string that the params form"""
     return "&".join(["{}={}".format(*pair) for pair in params])
+
 
 def to_grpc_metadata(params):
     """Returns the gRPC metadata that the routing header params form"""
--- a/api_core/google/api_core/gapic_v1/routing_header.py
+++ b/api_core/google/api_core/gapic_v1/routing_header.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from google.api_core.gapic_v1 import config
-from google.api_core.gapic_v1 import method
-from google.api_core.gapic_v1 import routing_header
+"""Helpers for handling routing header params."""
 
-__all__ = [
-    'config',
-    'method',
-    'routing_header',
-]
+METADATA_KEY = 'x-goog-header-params'
+
+def to_routing_header(params):
+    """Returns the routing header string that the params form"""
+    return "&".join(["{}={}".format(*pair) for pair in params])
+
+def to_grpc_metadata(params):
+    """Returns the gRPC metadata that the routing header params form"""
+    return (METADATA_KEY, to_routing_header(params))
--- a/api_core/google/api_core/gapic_v1/routing_header.py
+++ b/api_core/google/api_core/gapic_v1/routing_header.py
@@ -14,14 +14,30 @@
 
 """Helpers for handling routing header params."""
 
-METADATA_KEY = 'x-goog-header-params'
+
+try:
+    from urllib.parse import urlencode
+except ImportError:
+    def urlencode(params):
+        return '&'.join(['{}={}'.format(*pair) for pair in params])
+
+ROUTING_METADATA_KEY = 'x-goog-header-params'
+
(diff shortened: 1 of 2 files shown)


api_core/google/api_core/gapic_v1/routing_header.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for handling routing header params."""
+
+METADATA_KEY = 'x-goog-header-params'
```
jonparrott: For consistency with `client_info`, can you make this `ROUTING_METADATA_KEY`?

api_core/google/api_core/gapic_v1/routing_header.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def to_routing_header(params):
+    """Returns the routing header string that the params form"""
+    return "&".join(["{}={}".format(*pair) for pair in params])
```
jonparrott: This seems really brittle, could we possibly use [urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)?

api_core/tests/unit/gapic/test_routing_header.py
```diff
@@ -0,0 +1,29 @@
...
+    params = [('name', 'meep'), ('book.read', True)]
+    value = routing_header.to_routing_header(params)
+    assert value == "name=meep&book.read=True"
+
+
+def test_to_grpc_metadata():
+    params = [('name', 'meep'), ('book.read', True)]
+    metadata = routing_header.to_grpc_metadata(params)
+    assert metadata == (
+        routing_header.METADATA_KEY, "name=meep&book.read=True")
```
jonparrott: Real urlencoding would use '1' to represent `True`, yeah?

api_core/google/api_core/gapic_v1/routing_header.py
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for handling routing header params."""
+
+
+try:
```
jonparrott: Use `six.urlparse.urlencode`, it'll save you the trouble of needing this import guard.

api_core/google/api_core/gapic_v1/routing_header.py
```diff
@@ -0,0 +1,43 @@
...
+    from urllib.parse import urlencode
+except ImportError:
+    def urlencode(params):
+        return '&'.join(['{}={}'.format(*pair) for pair in params])
+
+ROUTING_METADATA_KEY = 'x-goog-header-params'
+
+
+def _to_url_string(x):
```
jonparrott: This really isn't necessary. My previous comment was just making an observation.

api_core/google/api_core/gapic_v1/routing_header.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+from six.moves.urllib.parse import urlencode
+
+ROUTING_METADATA_KEY = 'x-goog-header-params'
+
+
+def to_routing_header(params):
+    """Returns the routing header string that the params form"""
+    return urlencode(params)
```
jonparrott: So it's at this point I feel that I need to ask: what is the benefit of having this module at all if all it does it directly call `urlencode`. If the answer is "we may change how this works under the covers but keep the same interface" then cool.

/cc @lukesneeringer

eoogbe: You guessed it. The initial implementation is simple, but it will change.

jonparrott: coolbeans

Since this is a public API, I'm gonna make some documentation suggestions here, then we can merge.

PR #4341, opened 2017-11-03
Handle retry=None for mutate_rows, and some refactoring in test

justinseinlin:
/cc @dhermes 

As discussed in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4333, this change addresses the input retry=None for `mutate_rows`. @jonparrott I made an internal error `_BigtableRetryableError` and have it be used to trigger the retry instead of raising a fake `SerivceUnavailable`. Would like to get your opinion regarding this use.

Also factored out test response creation and replaced the magic numbers with constants for RPC Status Code. And removed the extra `the` from the error message.

### DIFF
--- a/bigtable/tests/unit/test_table.py
+++ b/bigtable/tests/unit/test_table.py
@@ -611,10 +611,7 @@ class Test__RetryableMutateRowsWorker(unittest.TestCase):
 
     def test_callable_no_retry_strategy(self):
         from google.api_core.retry import Retry
-        from google.cloud.bigtable._generated.bigtable_pb2 import (
-            MutateRowsResponse)
         from google.cloud.bigtable.row import DirectRow
-        from google.rpc.status_pb2 import Status
 
         # Setup:
         #   - Mutate 3 rows.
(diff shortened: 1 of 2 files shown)


justinseinlin:
This is also a follow up on #4256

bigtable/tests/unit/test_table.py
```diff
@@ -556,6 +556,12 @@ class Test__RetryableMutateRowsWorker(unittest.TestCase):
     INSTANCE_NAME = ('projects/' + PROJECT_ID + '/instances/' + INSTANCE_ID)
     TABLE_ID = 'table-id'
 
+    # RPC Status Codes
+    SUCCESS = 0
+    RETRYABLE_1 = 4
+    RETRYABLE_2 = 10
+    NON_RETRYABLE = 1
```
dhermes: Please don't hard-code these, i.e. read them from the relevant "source of truth" (if there is one).

justinseinlin: I replaced the magic numbers with their source values. I'm still keeping the `SUCCESS`, `RETRYABLE_1`, etc. as they are easier to read in the test code with the corresponding comments there.

justinseinlin: @dhermes @tseaver I noticed that we don't run linter on the test files. Is it just so our CI doesn't take too long (or longer that it already does)?

justinseinlin:
@dhermes Can you help setting the relevant label(s) to the PR? Should we normally set assignee as well? I could update neither though.

PR #4348, opened 2017-11-06
New Datastore auto-gen.

lukesneeringer:
This updates the auto-gen layer for Datastore.


dhermes:
@lukesneeringer The dependency on `gapic-google-cloud-datastore-v1` in `datastore/setup.py` should go away as well.

lukesneeringer:
Pretty sure it did go away. I just double-checked. Did I miss something?

lukesneeringer:
@dhermes I am also a bit baffled by the new system test errors. Any idea what `Value must be an iterable` in `lookup` is all about? (I can investigate if you do not know.)

datastore/setup.py
```diff
@@ -53,8 +53,10 @@
 REQUIREMENTS = [
     'google-cloud-core >= 0.28.0, < 0.29dev',
     'google-api-core >= 0.1.1, < 0.2.0dev',
-    'google-gax >= 0.15.7, < 0.16dev',
-    'gapic-google-cloud-datastore-v1 >= 0.15.0, < 0.16dev',
+    'google-auth >= 1.0.2, < 2.0dev',
+    'google-gax >= 0.15.15, < 0.16dev',
+    'googleapis-common-protos[grpc] >= 1.5.2, < 2.0dev',
+    'requests >= 2.18.4, < 3.0dev',
```
dhermes: @lukesneeringer Why do we "repeat" ourselves here when all of these are in the deps for `google-api-core` / `google-cloud-core`?

dhermes:
> Pretty sure it did go away. I just double-checked. Did I miss something?

No, I was mistaken (I missed `setup.py` in the diff).

> I am also a bit baffled by the new system test errors...

I do not know, but it feels like a mock doesn't have `__iter__` capabilities.

lukesneeringer:
> I do not know, but it feels like a mock doesn't have `__iter__` capabilities.

It is a system test, so it should not be a mock issue. I am asking Alan to look into it for a bit.

chemelnucfin:
So the issue is that in datastore_client.py lookup takes positional arguments, (project_id, keys, read_options=None), and in _extended_lookup from client.py, datastore_api.lookup takes (*args, **kwargs).  if those calls mismatch, we get problems.

Changing the datastore_client.py lookup into def lookup(self,
               project,
               read_options=None,               
               keys=None,
works.  There are other options that can be changed, but the others will break tests.  
Protobuf index order is project, options, and then keys if that makes a difference.

Any comments?

chemelnucfin:
@lukesneeringer

lukesneeringer:
Thanks -- that diagnosis was perfect. Fixed it.

dhermes:
Now there's a linter feature request! Always make sure we pass positional as positional and keyword as keyword.

lukesneeringer:
Unrelated CI failure, and tests are known to work. Merging.

mathewcohle:
Commit 9246de7a7138fabb3f25dceb50715de445684323 introduced following bug:
```
  File "/usr/local/lib/python2.7/dist-packages/google/cloud/datastore/client.py", line 321, in get
    eventual=eventual)
  File "/usr/local/lib/python2.7/dist-packages/google/cloud/datastore/client.py", line 376, in get_multi
    transaction_id=transaction and transaction.id,
  File "/usr/local/lib/python2.7/dist-packages/google/cloud/datastore/client.py", line 138, in _extended_lookup
    keys=key_pbs,
TypeError: lookup() got an unexpected keyword argument 'project_id'
```

Patch
```diff
From 7470329ecf13d9f85bfe8af71ee533856ff9ec4c Mon Sep 17 00:00:00 2001
From: mathewcohle <mathewcohle@gmail.com>
Date: Mon, 13 Nov 2017 11:41:01 +0100
Subject: [PATCH] Fix wrong keyword argument name introduced in #4348

---
 datastore/google/cloud/datastore/client.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/datastore/google/cloud/datastore/client.py b/datastore/google/cloud/datastore/client.py
index a3d1f9d43..c42c75e32 100644
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -133,7 +133,7 @@ def _extended_lookup(datastore_api, project, key_pbs,
     while loop_num < _MAX_LOOPS:  # loop against possible deferred.
         loop_num += 1
         lookup_response = datastore_api.lookup(
-            project_id=project,
+            project=project,
             read_options=read_options,
             keys=key_pbs,
         )
-- 
2.15.0
```

@lukesneeringer

PR #4349, opened 2017-11-06
Set next_page_token to none if there are no more result

### LINKED ISSUE #4347: Datastore: broken paging next_page_token
I want to use `datastore` paging to fetch the query results. Specifically, I upload 11 entities to `datastore` and then want to fetch them (`limit=10`) using following code:
```python
query = <google.cloud.datastore.Client>.query(kind='test_kind')

query_iter = query.fetch(limit=limit, start_cursor=cursor)
page = next(query_iter.pages)
next_cursor = query_iter.next_page_token
entities = list(page)

logging.info('Cursor: %s, Next cursor: %s, Loaded entities: %d.',
              cursor, next_cursor, len(entities))
```
(where initially `cursor = None` and then is set to `next_cursor` value)

Unfortunately, I am getting not `None` value for the `next_cursor` even there are not any more results to fetch:
```
# First call, everything as expected
Cursor: None, Next cursor: foo, Loaded entities: 10.

# Second call, expected next_page_token = None
Cursor: foo, Next cursor: bar, Loaded entities: 1.

# Third call, again expected next_page_token = None
Cursor: bar, Next cursor: bar, Loaded entities: 0.
```

It seems that `next_page_token` is always not `None` and moreover, if there is not any result to fetch, `next_cursor == cursor`.

Setup:
```
Python 2.7.12
google-cloud-datastore==1.3.0
```

Is this expected? If yes, how can I recognise that there are not more results to fetch? 

I've checked the code briefly and it might be related to https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/datastore/google/cloud/datastore/query.py#L457. It seems like `response_pb.batch.end_cursor == b''` does not hold and thus `next_page_token` is not set to `None`, but not sure why this is the case.

mathewcohle:
Depending on empty value of `response_pb.batch.end_cursor` is not reliable as the behaviour is not documented. It's better to check API response field `moreResults`, see
https://cloud.google.com/datastore/docs/reference/rpc/google.datastore.v1#google.datastore.v1.QueryResultBatch.

Closes #4347.

### DIFF
--- a/datastore/google/cloud/datastore/query.py
+++ b/datastore/google/cloud/datastore/query.py
@@ -26,9 +26,10 @@ from google.cloud.datastore.key import Key
 
 
 _NOT_FINISHED = query_pb2.QueryResultBatch.NOT_FINISHED
+_NO_MORE_RESULTS = query_pb2.QueryResultBatch.NO_MORE_RESULTS
 
 _FINISHED = (
-    query_pb2.QueryResultBatch.NO_MORE_RESULTS,
+    _NO_MORE_RESULTS,
     query_pb2.QueryResultBatch.MORE_RESULTS_AFTER_LIMIT,
     query_pb2.QueryResultBatch.MORE_RESULTS_AFTER_CURSOR,
 )
@@ -470,7 +471,7 @@ class Iterator(page_iterator.Iterator):
         """
         self._skipped_results = response_pb.batch.skipped_results
 
-        if response_pb.batch.end_cursor == b'':  # Empty-value for bytes.
+        if response_pb.batch.more_results == _NO_MORE_RESULTS:
             self.next_page_token = None
         else:
             self.next_page_token = base64.urlsafe_b64encode(
--- a/datastore/tests/unit/test_query.py
+++ b/datastore/tests/unit/test_query.py
@@ -467,7 +467,7 @@ class TestIterator(unittest.TestCase):
         entity_pbs = [
             _make_entity('World', 1234, 'PROJECT'),
         ]
-        cursor_as_bytes = b''
+        cursor_as_bytes = b'\x9ai\xe7'
         skipped_results = 44
         more_results_enum = query_pb2.QueryResultBatch.NO_MORE_RESULTS
         response_pb = _make_query_response(


datastore/google/cloud/datastore/query.py
```diff
@@ -454,7 +455,7 @@ def _process_query_results(self, response_pb):
         """
         self._skipped_results = response_pb.batch.skipped_results
 
-        if response_pb.batch.end_cursor == b'':  # Empty-value for bytes.
+        if response_pb.batch.more_results == _NO_MORE_RESULTS:
```
dhermes: I'm fairly certain there **was** a reason we did this, i.e. the server API didn't quite behave the way we expected.

I'd prefer sign off from the backend team before making the change.

mathewcohle:
FYI, I've tested the change against our tests which were previously failing. The patch has fixed the problem.

mathewcohle:
@dhermes how does it look? This is currently blocker for migration to `env: flex` for us.

dhermes:
@mathewcohle RE: "how does it look?"

The code looks fine, but my previous sentiment is unchanged:

> I'd prefer sign off from the backend team before making the change.

datastore/google/cloud/datastore/query.py
wsh: Hi there. :smile: I'm on the Datastore backend team.

This change looks good to me, but I'm thinking the check should actually mirror L487, i.e. `response_pb.batch.more_results in _FINISHED` (see https://cloud.google.com/datastore/docs/reference/rpc/google.datastore.v1#google.datastore.v1.QueryResultBatch.MoreResultsType).

cc @pcostell

wsh: Talked with @pcostell; the original proposal (`NO_MORE_RESULTS`) is correct. :thumbsup:

mathewcohle: Thanks @wsh!

mathewcohle: @dhermes could it be merged now? Also do you know when is the next release of `datastore` submodule planned? Thanks.

tseaver:
@mathewcohle Thank you for the patch!

dhermes:
@tseaver Sorry I didn't chime in. I wanted to actually reproduce #4347 before merging this. The [build][1] right after this broke, so we may have to revert this change.

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4200

tseaver:
Sorry:  everything was green on the last commit in the PR, and you said you were waiting on signoff from the back-end team, which we got.

PR #4350, opened 2017-11-06
BigQuery: Can iterate over QueryJob results without explicitly calling result()

alixhami:


jonparrott:
I'd like @tswast or @jba to take a look from a BigQuery perspective as well.

jba:
LGTM, but I will defer to Tim.

alixhami:
@tswast If the purpose of `Client.query_rows()` was just to get query results without calling `result()`, then it would be unnecessary. Are there other uses for it, or should I remove it in another PR?

tswast:
Talked in person. Let's remove `query_rows` in another PR. I can think of no other uses for it.

Issue #4152, opened 2017-10-11
BigQuery: Client.list_rows should prefetch the first page and populate total_rows and other attributes


bits01:
https://googlecloudplatform.github.io/google-cloud-python/latest/_modules/google/cloud/bigquery/table.html#Table.fetch_data

Return value doc:
```
:returns: Iterator of row data :class:`tuple`s. During each page, the
                  iterator will have the ``total_rows`` attribute set,
                  which counts the total number of rows **in the table**
                  (this is distinct from the total number of rows in the
                  current page: ``iterator.page.num_items``).
```

The iterator does not have the `total_rows` attribute or a `page` attribute either.

Can `fetch_data()` get a count of all rows or the only way is to iterate and count?

bits01:
`query.fetch_data()` too.

lukesneeringer:
Summoning @tswast since he has been in this code a lot recently and probably knows the answer off hand. :-)

tswast:
There's definitely a `pages` property. Not sure about page.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/99c4498339e8cddb628962215e7fa7953a77c957/bigquery/tests/system.py#L296-L299

Looking at https://github.com/GoogleCloudPlatform/google-cloud-python/blob/99c4498339e8cddb628962215e7fa7953a77c957/core/google/api/core/page_iterator.py#L248, I think you're right that the docs need to be updated.

dhermes:
@tswast The iterator docs are great, they just got unpublished when the module moved from `google.cloud.iterator` to `google.api.core.page_iterator`.

/cc @jonparrott

bits01:
#4153 is related.

tswast:
The iterator is using

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/e6cc4b4a6486971f910139d766c7807a588ea53b/bigquery/google/cloud/bigquery/_helpers.py#L507-L522

which appears to be populating `total_rows`. I'll investigate further and make sure our system tests are checking for this property.

bits01:
It works on `QueryResult`, but *not* on the iterator returned by the destination table `fetch_data()`:
`AttributeError: 'HTTPIterator' object has no attribute 'total_rows'`

tswast:
My guess as to what is happening is that total_rows only gets populated after you start iterating through, because it gets that information from the first page of results.

I agree that this is confusing behavior.

[...]

PR #4354, opened 2017-11-07
BigQuery: Get query results as pandas DataFrame

alixhami:
This is a basic implementation of a method converting query results to a pandas DataFrame. I am hoping for comments on how best to implement this. Here is an example of how to use the current implementation:
`df = client.query(QUERY).to_dataframe()`

The intent is that pandas would be an optional dependency, and would not be required unless the DataFrame functionality is used. This is not yet handled in my current implementation, which is why CircleCI will fail. I ran the tests locally with pandas installed and they pass.

~~The base of this PR is #4350~~


jba:
I don't personally like the idea of an optional dependency. I guess I try to think of Python as a typed language as much as possible (and hopefully someday it will be). 

Maybe instead of actually returning a pandas DataFrame, you could (a) expose whatever currently hidden information is needed to efficiently create one, and (b) add a different module that depends on this one and pandas, and creates a DataFrame.

dhermes:
> I don't personally like the idea of an optional dependency. 

I don't particularly like it too much, but in this case, the `pandas` dependency is a bit too heavy to burden all users with. We **could** have a more helpful error message if the import fails. We also **should** provide an "extra" ([e.g.][1]) that allows installing via:

```
pip install google-cloud-bigquery[pandas-extra]
```

(you can name the "extra" anything you like, I just chose `pandas-extra` out of a hat).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1ed56fdfac90b71ad3c8f2fd895f30c281825a82/api_core/setup.py#L61-L64

bigquery/tests/system.py
```diff
@@ -1235,6 +1235,30 @@ def test_query_future(self):
         row_tuples = [r.values() for r in iterator]
         self.assertEqual(row_tuples, [(1,)])
 
+    def test_query_iter(self):
```
dhermes: This doesn't need to be a system test, it's just checking behavior of a magic method. (Unless you don't have an equivalent test that invoked `.result()`.)

bigquery/tests/system.py
```diff
@@ -1235,6 +1235,30 @@ def test_query_future(self):
...
+        import types
+        query_job = Config.CLIENT.query('SELECT 1')
+        self.assertIsInstance(iter(query_job), types.GeneratorType)
+        row_tuples = [r.values() for r in query_job]
+        self.assertEqual(row_tuples, [(1,)])
+
+    def test_query_to_dataframe(self):
+        import pandas as pd
```
dhermes: It doesn't seem like this needs to be a system test, since (as above), it is just wrapping `.result()` which is presumably already system-tested.

----

If you insist, I'd prefer the system tests to be runnable without `pandas`. To make this possible you could:

```python
try:
    import pandas
except ImportError:
    pandas = None
```

and then use

```python
@unittest.skipIf(pandas is None, 'Requires `pandas`')
def test_query_to_dataframe(self):
    ...
```
----

In general, our style is "import at run time in unit tests" and "import at import time" in system tests. The idea being that you will run unit tests **a lot** and don't want a silly error (e.g. a syntax error, which would fail an import) to fail the entire unit test suite at **import time**.

bigquery/tests/unit/test_job.py
```diff
@@ -2720,6 +2720,89 @@ def test_reload_w_alternate_client(self):
         self.assertEqual(req['path'], PATH)
         self._verifyResourceProperties(job, RESOURCE)
 
+    def test_to_dataframe(self):
+        import pandas as pd
+
+        begun_resource = self._makeResource()
```
dhermes: **NOTE**: Method names should still adhere to PEP8 in unit tests. (I'll send a PR to rename `_makeResource` as `_make_resource`, so don't worry about it in this PR.) (Sorry our tooling didn't complain about this.)

bigquery/tests/unit/test_job.py
```diff
@@ -2720,6 +2720,89 @@ def test_reload_w_alternate_client(self):
...
+        done_resource['status'] = {'state': 'DONE'}
+        connection = _Connection(
+            begun_resource, query_resource, done_resource, query_resource)
+        client = _make_client(project=self.PROJECT, connection=connection)
+        job = self._make_one(self.JOB_ID, self.QUERY, client)
+        df = job.to_dataframe()
+
+        self.assertIsInstance(df, pd.DataFrame)
+        self.assertEqual(len(df), 4)
+        self.assertEqual(list(df), ['name', 'age'])
```
dhermes: It's surprising to me that `len(df)` and `len(list(df))` could differ. Why is that?

bigquery/tests/unit/test_job.py
```diff
@@ -2720,6 +2720,89 @@ def test_reload_w_alternate_client(self):
...
+
+    def test_to_dataframe_w_empty_results(self):
```
dhermes: Ditto here using `unittest.skipIf` to avoid this test if `pandas` is missing.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1949,18 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
...
+        import pandas as pd
+
+        iterator = self.result()
+        column_headers = [field.name for field in iterator.schema]
+        rows = [row.values() for row in iterator]
+
+        return pd.DataFrame(rows, columns=column_headers)
+
+    def __iter__(self):
+        return iter(self.result())
```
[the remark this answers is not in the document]
alixhami: I originally based the PR on #4350, but just rebased on master to avoid confusion.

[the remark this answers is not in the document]
dhermes: Which in turn caused confusion :grinning: 

What is holding up #4350?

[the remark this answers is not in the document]
alixhami: @tswast is out of town and hasn't reviewed, but he will be back next week

bigquery/tests/unit/test_job.py
dhermes: @alixhami You'll want to rebase after #4355 was merged

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1949,15 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
         return self._client.list_rows(dest_table, selected_fields=schema,
                                       retry=retry)
 
+    def to_dataframe(self):
+        import pandas as pd
```
tseaver: I would attempt to import it at module scope, and then set a flag to indicate success.  E.g.:

```python
try:
    import pandas
except ImportError:
    _HAVE_PANDAS = False
else:
    _HAVE_PANDAS = True
```

I would then raise a ValueError inside `to_dataframe` if `_HAVE_PANDAS` is False.

bigquery/tests/system.py
```diff
@@ -1235,6 +1235,23 @@ def test_query_future(self):
         row_tuples = [r.values() for r in iterator]
         self.assertEqual(row_tuples, [(1,)])
 
+    def test_query_to_dataframe(self):
```
tseaver: Use `unittest.skipUnless(_HAVE_PANDAS)` here.

bigquery/tests/unit/test_job.py
```diff
@@ -2720,6 +2720,70 @@ def test_reload_w_alternate_client(self):
         self.assertEqual(req['path'], PATH)
         self._verifyResourceProperties(job, RESOURCE)
 
+    def test_to_dataframe(self):
```
tseaver: Again, use `unittest.skipUnless(_HAVE_PANDAS)`.

jonparrott:
+1 to what @dhermes said. We should provide a helpful message if they use a method that requires pandas (I also doubt the folks who want to use this will install bigquery before pandas), and we should provide the extra for pip.

bigquery/google/cloud/bigquery/job.py
jonparrott: +1, you can also do:

```python
try:
    import pandas
except ImportError:
    pandas = None
```

and then check `if pandas is None`.

Also, don't alias imports in this library (I know it's very common to alias pandas and numpy elsewhere).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1949,15 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
...
+
+        iterator = self.result()
```
jonparrott: `results`? (`iterator` is a really generic term, better to name it after *what* it iterates over).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1949,18 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
         return self._client.list_rows(dest_table, selected_fields=schema,
                                       retry=retry)
 
+    def to_dataframe(self):
```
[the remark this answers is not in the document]
alixhami: Added in 49d65ce

bigquery/tests/system.py
alixhami: Removed the system test in 823b686. Fixed the imports and skipped the tests in 70d6bd9

bigquery/tests/unit/test_job.py
alixhami: Rebased and updated the function name in ea0c219

bigquery/tests/unit/test_job.py
alixhami: This is a pretty weird behavior, but `len()` returns the number of rows and `list()` returns the column names. I added comments to these tests to make this more clear in 70d6bd9

bigquery/tests/unit/test_job.py
alixhami: Added skips in 70d6bd9

bigquery/google/cloud/bigquery/job.py
alixhami: Fixed in 4bc85fd

bigquery/tests/system.py
alixhami: Added skips in 70d6bd9

bigquery/tests/unit/test_job.py
alixhami: Added skips in 70d6bd9

bigquery/google/cloud/bigquery/job.py
alixhami: Changed to `query_results` in 4bc85fd

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1953,29 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
...
+        """Create a pandas DataFrame from the query results.
+
+        :rtype: ``pandas.DataFrame``
```
dhermes: This should be:

```
:class:`pandas.DataFrame`
# OR
:class:`~pandas.DataFrame`
```

It will even render as a link to the pandas docs because we have `pandas` in [our intersphinx mapping][1].

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/69a9840955caefbd431ee459ae91632b9cf79a16/docs/conf.py#L302

bigquery/setup.py
```diff
@@ -58,6 +58,10 @@
     'requests >= 2.18.0',
 ]
 
+EXTRAS_REQUIREMENTS = {
+    'pandas': ['pandas >= 0.3.0'],
```
dhermes: You sure you don't want a later release? This one is from 2011 and the lastest one is 0.21.0

dhermes:
So @alixhami, after discussion with @jonparrott we have decided it is "preferred" if all new docstrings are in the Google style (vs. traditional Sphinx).

I'm happy to help out / chat over hangouts / send you some code snippets.

Can you change the newly added docstring to Google style?

bigquery/setup.py
alixhami: This is the earliest version of pandas that will work. Should I use a different metric for choosing a version?

dhermes: @jonparrott WDYT?

dhermes:
LGTM, but you need to resolve the coverage issues

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1965,7 +1965,7 @@ def to_dataframe(self):
             ValueError: If the `pandas` library cannot be imported.
 
         """
-        if pandas is None:
+        if pandas is None:  # pragma: NO COVER
```
dhermes: @alixhami Please just test this branch. You can use `@mock.patch('google.cloud.bigquery.job.pandas', new=None)` to have a unit test which will raise this exception

dhermes:
@alixhami This is good to merge once CI goes green, right?

alixhami:
@dhermes it should be good to merge since all the comments have now been addressed. @tswast what do you think?

dhermes:
`lint` failed :grinning:

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1953,29 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
         return self._client.list_rows(dest_table, selected_fields=schema,
                                       retry=retry)
 
+    def to_dataframe(self):
+        """Create a pandas DataFrame from the query results.
```
tswast: What's the reason this is in the QueryJob and not a subclass of the paged iterator we use to list rows?

I think this method could be handy to use on the result of `client.list_rows(table)`.

alixhami: The Iterator class is in api_core and we only want to call `to_dataframe()` on the iterators of query results, so I thought it would make more sense to have it here. Would it make sense to have a special type of iterator for query results?

tswast: Per https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4152#issuecomment-336952584 @dhermes recommends we create a custom iterator class to address that issue, anyway. I'd prefer we do it now and add the `to_dataframe()` to that class.

Note: there isn't anything special about the query results, if you trace through the code you will see that it calls `list_rows` behind the scenes.

tswast: (I'm not saying fix #4152 in this PR, just add the custom class)

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1949,6 +1953,29 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
...
+        if pandas is None:
+            raise ValueError('The pandas library is not installed, please '
+                             'install pandas to use the to_dataframe() '
+                             'function.')
+
+        query_results = self.result()
+        column_headers = [field.name for field in query_results.schema]
+        rows = [row.values() for row in query_results]
+
+        return pandas.DataFrame(rows, columns=column_headers)
```
tswast: Have you tried this on a query that includes some nulls in results? I'm curious if we have to do anything special to tell Pandas when there might be nulls. Also curious about what types Pandas ends up using. It'd be good to see a unit test with some rows that are timestamp values.

alixhami: I have a bunch of type testing I did in a notebook that I can move into the unit tests. Any other types we should cover in unit tests besides timestamps?

tswast: Based on https://stackoverflow.com/a/21767006/101923

Pandas supports:
datetime64, timedelta64, int64, float64, bool and object

So, we should try to cover at least these.

I don't believe BigQuery has a native timedelta type, since a [timestamp diff comes back as an int64](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#timestamp_diff).

alixhami: I added a test to check handling for null and most of the BigQuery data types. I used the [standard SQL data types](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types) for reference. Should I add tests for array and/or struct types?

tswast: Pandas doesn't really have great support for those data types AFAIK, but I suppose it would be good to verify that Pandas doesn't mess them up (falling back to the "object" data type in Pandas)

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1269,20 +1268,16 @@ def list_rows(self, table, selected_fields=None, max_results=None,
...
-            items_key='rows',
             page_token=page_token,
-            next_token='pageToken',
             max_results=max_results,
-            page_start=_rows_page_start,
             extra_params=params)
-        iterator.schema = schema
-        iterator._field_to_index = _field_to_index_mapping(schema)
-        return iterator
+        row_iterator._schema = schema
```
tswast: We might want to move schema and field_to_index to the RowIterator constructor.

bigquery/tests/unit/test_table.py
```diff
@@ -773,3 +778,189 @@ def test_row(self):
...
+        schema = [
+            SchemaField('start_timestamp', 'TIMESTAMP'),
+            SchemaField('seconds', 'INT64'),
+            SchemaField('miles', 'FLOAT64'),
+            SchemaField('payment_type', 'STRING'),
+            SchemaField('complete', 'BOOL'),
+            SchemaField('date', 'DATE'),
+        ]
+        row_data = [
+            [None, None, None, None, None, None],
```
tswast: Could you also add another test for checking the types where there are no nulls? Pandas internally treats a column that contains nulls as Python objects rather than it's more specific types.

I'd like to see checking the types of the columns (in the new test without nulls, since I'm pretty sure it'll fail here with nulls) as done here:
https://stackoverflow.com/a/45568211/101923

```
assert types.is_datetimetz(df['start_timestamp'])
assert types.is_integer_dtype(df['seconds'])
assert types.is_float_dtype(df['miles'])
...
```

tswast:
In the same spirit as https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4350, I'm thinking it could be handy to have a little helper function in `QueryJob` called `to_dataframe()` that calls `result().to_dataframe()`.

bigquery/google/cloud/bigquery/client.py
alixhami: Refactored in deb3e48

bigquery/tests/unit/test_table.py
alixhami: Added dtype tests in 0cdc967

bigquery/nox.py
```diff
@@ -36,7 +36,7 @@ def default(session):
     """
     # Install all test dependencies, then install this package in-place.
     session.install('mock', 'pytest', 'pytest-cov', *LOCAL_DEPS)
-    session.install('-e', '.')
+    session.install('-e', '.[pandas]')
```
tswast: I'd be happier if there was at least one session that ran the unit tests without Pandas.

Maybe @dhermes @lukesneeringer @jonparrott are more comfortable with the risk that we accidentally convert the optional dependency into a required dependency at some point?

Looks like the snippets session doesn't include Pandas, but maybe that's more accidental than intentional?

dhermes: I'm not worried about converting the optional dependency, but I'm also fine with running a session without pandas.

jonparrott:
This LGTM, @tseaver @dhermes did y'all have any reservations before merging this?

dhermes:
@alixhami This is failing on Python 2.7, distinction being between `str` and `unicode`

1. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4554
1. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4556
1. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4557

PR #4357, opened 2017-11-08
Datastore: Add support for 'TransactionOptions'

chemelnucfin:
Changes reflecting review and Luke's autogen (please see #4348 for issues regarding his autogen).  A manual fix has been included.

Closes #4335.  This includes reviews and changes from there.

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -511,9 +511,13 @@ class Client(ClientWithProject):
         """Proxy to :class:`google.cloud.datastore.batch.Batch`."""
         return Batch(self)
 
-    def transaction(self):
-        """Proxy to :class:`google.cloud.datastore.transaction.Transaction`."""
-        return Transaction(self)
+    def transaction(self, **kwargs):
+        """Proxy to :class:`google.cloud.datastore.transaction.Transaction`.
+
+        :type kwargs: dict
+        :param kwargs: Keyword arguments to be passed in.
+        """
+        return Transaction(self, **kwargs)
 
     def query(self, **kwargs):
         """Proxy to :class:`google.cloud.datastore.query.Query`.
--- a/datastore/google/cloud/datastore/transaction.py
+++ b/datastore/google/cloud/datastore/transaction.py
@@ -15,6 +15,7 @@
 """Create / interact with Google Cloud Datastore transactions."""
 
 from google.cloud.datastore.batch import Batch
+from google.cloud.datastore_v1.types import TransactionOptions
 
 
 class Transaction(Batch):
@@ -152,13 +153,22 @@ class Transaction(Batch):
 
     :type client: :class:`google.cloud.datastore.client.Client`
     :param client: the client used to connect to datastore.
+
+    :type read_only: bool
+    :param read_only: indicates the transaction is read only.
     """
 
     _status = None
 
-    def __init__(self, client):
+    def __init__(self, client, read_only=False):
         super(Transaction, self).__init__(client)
         self._id = None
+        if read_only:
+            options = TransactionOptions(
+                read_only=TransactionOptions.ReadOnly())
+        else:
+            options = TransactionOptions()
+        self._options = options
 
     @property
     def id(self):
@@ -234,3 +244,21 @@ class Transaction(Batch):
         finally:
             # Clear our own ID in case this gets accidentally reused.
             self._id = None
+
+    def put(self, entity):
+        """Adds an entity to be committed.
+
+        Ensures the transaction is not marked readonly.
+        Please see documentation at
+        :meth:`~google.cloud.datastore.batch.Batch.put`
+
+        :type entity: :class:`~google.cloud.datastore.entity.Entity`
+        :param entity: the entity to be saved.
+
+        :raises: :class:`RuntimeError` if the transaction
+                 is marked ReadOnly
+        """
+        if self._options.HasField('read_only'):
+            raise RuntimeError("Transaction is read only")
+        else:
+            super(Transaction, self).put(entity)
--- a/datastore/tests/system/test_system.py
+++ b/datastore/tests/system/test_system.py
@@ -22,6 +22,7 @@ import six
 from google.cloud._helpers import UTC
 from google.cloud import datastore
 from google.cloud.datastore.helpers import GeoPoint
+from google.cloud.datastore_v1 import types
 from google.cloud.environment_vars import GCD_DATASET
 from google.cloud.exceptions import Conflict
 
--- a/datastore/tests/unit/test_client.py
+++ b/datastore/tests/unit/test_client.py
@@ -934,6 +934,22 @@ class TestClient(unittest.TestCase):
             self.assertIs(xact, mock_klass.return_value)
             mock_klass.assert_called_once_with(client)
 
+    def test_read_only_transaction_defaults(self):
+        from google.cloud.datastore.transaction import Transaction
+        from google.cloud.datastore_v1.types import TransactionOptions
+        creds = _make_credentials()
+        client = self._make_one(credentials=creds)
+        xact = client.transaction(read_only=True)
+        self.assertEqual(xact._options,
+                         TransactionOptions(
+                             read_only=TransactionOptions.ReadOnly()
+                         )
+        )
+        self.assertFalse(xact._options.HasField("read_write"))
+        self.assertTrue(xact._options.HasField("read_only"))
+        self.assertEqual(xact._options.read_only,
+                         TransactionOptions.ReadOnly())
+
     def test_query_w_client(self):
         KIND = 'KIND'
 
--- a/datastore/tests/unit/test_transaction.py
+++ b/datastore/tests/unit/test_transaction.py
@@ -22,11 +22,17 @@ class TestTransaction(unittest.TestCase):
     @staticmethod
     def _get_target_class():
         from google.cloud.datastore.transaction import Transaction
-
         return Transaction
+
+    def _get_options_class(self, **kw):
+        from google.cloud.datastore_v1.types import TransactionOptions
+        return TransactionOptions
 
     def _make_one(self, client, **kw):
         return self._get_target_class()(client, **kw)
+
+    def _make_options(self, **kw):
+        return self._get_options_class()(**kw)
 
     def test_ctor_defaults(self):
         project = 'PROJECT'
@@ -211,6 +217,27 @@ class TestTransaction(unittest.TestCase):
         client._datastore_api.commit.assert_not_called()
         self.assertIsNone(xact.id)
         self.assertEqual(ds_api.begin_transaction.call_count, 1)
+
+    def test_constructor_read_only(self):
+        project = 'PROJECT'
+        id_ = 850302
+        ds_api = _make_datastore_api(xact=id_)
+        client = _Client(project, datastore_api=ds_api)
+        read_only = self._get_options_class().ReadOnly()
+        options = self._make_options(read_only=read_only)
+        xact = self._make_one(client, read_only=True)
+        self.assertEqual(xact._options, options)
+
+    def test_put_read_only(self):
+        project = 'PROJECT'
+        id_ = 943243
+        ds_api = _make_datastore_api(xact_id=id_)
+        client = _Client(project, datastore_api=ds_api)
+        entity = _Entity()
+        xact = self._make_one(client, read_only=True)
+        xact.begin()
+        with self.assertRaises(RuntimeError):
+            xact.put(entity)
 
 
 def _make_key(kind, id_, project):


chemelnucfin:
hold on a second.  I am still missing a couple comments.

datastore/google/cloud/datastore_v1/gapic/datastore_client.py
```diff
@@ -151,8 +151,8 @@ def __init__(self,
     # Service calls
     def lookup(self,
                project_id,
-               keys,
                read_options=None,
+               keys=None,
```
dhermes: @chemelnucfin Is this change not in `master` yet? Wasn't that required for #4348?

datastore/google/cloud/datastore/transaction.py
```diff
@@ -152,13 +153,19 @@ def Entity(*args, **kwargs):
...
     :param client: the client used to connect to datastore.
+
+    :type options: `google.cloud.datastore.transaction.TransactionOptions`
+    :param options: the transaction options used for this transaction.
     """
 
     _status = None
 
-    def __init__(self, client):
+    def __init__(self, client, options=None):
```
dhermes: @chemelnucfin What became of our discussion about just using a `bool` here for `read_only`?

datastore/tests/system/test_system.py
```diff
@@ -442,6 +443,25 @@ def test_transaction_via_with_statement(self):
         self.case_entities_to_delete.append(retrieved_entity)
         self.assertEqual(retrieved_entity, entity)
 
+    def test_readonly_put(self):
```
dhermes: @chemelnucfin This **should not** be a system test. It is testing behavior of the client that **does not** depend on a live network.

datastore/tests/system/test_system.py
```diff
@@ -442,6 +443,25 @@ def test_transaction_via_with_statement(self):
...
+        options = types.TransactionOptions(
+                            read_only=types.TransactionOptions.ReadOnly())
+        with self.assertRaises(RuntimeError):
+            with Config.CLIENT.transaction(options) as xact:
+                xact.put(entity)
+
+    def test_readwrite_put(self):
+        entity = datastore.Entity(key=Config.CLIENT.key('Company', 'Google'))
+        options = types.TransactionOptions(
+                            read_write=types.TransactionOptions.ReadWrite())
```
dhermes: Please indent 4 spaces (not sure why linter didn't complain, sorry).

datastore/google/cloud/datastore/transaction.py
```diff
@@ -152,13 +153,19 @@ def Entity(*args, **kwargs):
 
     :type client: :class:`google.cloud.datastore.client.Client`
     :param client: the client used to connect to datastore.
+
+    :type options: `google.cloud.datastore.transaction.TransactionOptions`
```
dhermes: @chemelnucfin 

1. This class doesn't seem to exist?
1. In general we don't ask users to explicitly create protobuf objects, so you'd either need a wrapper class of your own, or you should do as we discussed before (just use `read_only=False` as the argument and create `TransactionOptions` protobuf for the user).
1. You should use `:class:` before a Sphinx type

datastore/google/cloud/datastore/transaction.py
```diff
@@ -152,13 +153,19 @@ def Entity(*args, **kwargs):
...
     """
 
     _status = None
 
-    def __init__(self, client):
+    def __init__(self, client, options=None):
         super(Transaction, self).__init__(client)
         self._id = None
+        if options is None:
+            options = TransactionOptions()
```
dhermes: `None` is an acceptable default for optional protobuf arguments (e.g. in GAPIC methods)

datastore/tests/unit/test_transaction.py
```diff
@@ -12,22 +12,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import mock
 import unittest
 
-import mock
+
+class TestTransactionOptions(unittest.TestCase):
```
dhermes: This test corresponds to an auto-generated class. It should already be tested.

datastore/tests/unit/test_transaction.py
```diff
@@ -12,22 +12,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import mock
```
dhermes: This is not in the standard library, it should be in another imports section:

```python
import unittest

import mock
```

FWIW `unittest.mock` **is** in the standard library, but only in Python 3.

datastore/tests/unit/test_transaction.py
```diff
@@ -12,22 +12,55 @@
...
+    def test_read_write(self):
+        options = self._make_one(read_write=
+                                 self._get_target_class().ReadWrite())
+        self.assertTrue(options.HasField('read_write'))
+
+    def test_previous_transaction(self):
+        previous_transaction = b'234'
+        options = self._make_one(
+                           read_write=self._get_target_class().ReadWrite(
+                               previous_transaction=previous_transaction))
```
dhermes: This method should be deleted? (Per my previous comment, this whole class should.)

----

Just for the sake of explanation: What rules is your editor using to make these indents? They don't seem to make sense? Please indent 4 spaces. I would **much** prefer to see:

```python
klass = self._get_target_class()
options = self._make_one(
    read_write=klass.ReadWrite(
        previous_transaction=previous_transaction,
    ),
)
```

Note also the call to `_get_target_class()` **before** it is used (there is such a thing as doing "too much" in a single expression).

datastore/tests/unit/test_transaction.py
```diff
@@ -12,22 +12,55 @@
...
+        self.assertEqual(options.read_write.previous_transaction,
+                         previous_transaction)
 
 
 class TestTransaction(unittest.TestCase):
 
     @staticmethod
     def _get_target_class():
         from google.cloud.datastore.transaction import Transaction
-
```
dhermes: Please don't make cosmetic changes like this unrelated to the PR

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +245,95 @@ class Foo(Exception):
...
+        entity = _Entity()
+        xact = self._make_one(client, options=options)
+        xact.begin()
+        xact.put(entity)
+        mutated_entity = _mutated_pb(self, xact.mutations, 'insert')
+        self.assertEqual(mutated_entity.key, entity.key.to_protobuf())
+
+
+def _assert_num_mutations(test_case, mutation_pb_list, num_mutations):
+    test_case.assertEqual(len(mutation_pb_list), num_mutations)
```
dhermes: This is a one-liner, ISTM you should just inline this logic wherever `_assert_num_mutations` is used?

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +245,95 @@ class Foo(Exception):
...
+            xact.put(entity)
+
+    def test_put_read_write(self):
+        project = 'PROJECT'
+        id_ = 943243
+        ds_api = _make_datastore_api(xact_id=id_)
+        client = _Client(project, datastore_api=ds_api)
+        options = self._make_options(read_write=
+                                     self._make_options().ReadWrite(
+                                        previous_transaction=b'321'))
```
dhermes: @chemelnucfin Ditto here on the indents. Also **never** indent after an `=` sign.

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +245,95 @@ class Foo(Exception):
...
+        with self.assertRaises(RuntimeError):
+            xact.put(entity)
+
+    def test_put_read_write(self):
+        project = 'PROJECT'
+        id_ = 943243
+        ds_api = _make_datastore_api(xact_id=id_)
+        client = _Client(project, datastore_api=ds_api)
+        options = self._make_options(read_write=
+                                     self._make_options().ReadWrite(
```
dhermes: ISTM you want `_get_options_class()` here, not `_make_options()`.

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +245,95 @@ class Foo(Exception):
...
+        begin = ds_api.begin_transaction
+        begin.assert_called_once()
+        self.assertEqual(begin.call_count, 1)
+
+    def test_put_read_only(self):
+        project = 'PROJECT'
+        id_ = 943243
+        ds_api = _make_datastore_api(xact_id=id_)
+        client = _Client(project, datastore_api=ds_api)
+        options = self._make_options(read_only=self._make_options().ReadOnly())
```
dhermes: ISTM you want `_get_options_class()` here, not `_make_options()`.

datastore/tests/unit/test_transaction.py
```diff
@@ -281,7 +403,6 @@ def _make_commit_response(*keys):
 def _make_datastore_api(*keys, **kwargs):
     commit_method = mock.Mock(
         return_value=_make_commit_response(*keys), spec=[])
-
```
dhermes: Ditto here on cosmetic changes

datastore/google/cloud/datastore/transaction.py
chemelnucfin: @lukesneeringer @dhermes So I just want to make sure what our consensus on this is.  Do we actually allow users to specify the entire TransactionOptions just like what it is in protobuf, or do we just make it easier for the users by just specifying a readonly boolean or something similar?  I originally thought that it would be best to follow the protobuf, but @dhermes mentioned it would be easier for the users if it was simpler.  So now I think I can be convinced either way.

I will shelve this until we reach a consensus on this.

tseaver: As a user, I would prefer to pass "simple" arguments, rather than having to construct a "complicated" options object.  E.g.:

```python
with client.transaction(read_only=True):
    ...
```

is much nicer than:

```python
from google.cloud.datastore import TransactionOptions
txn_options = TransactionOptions(read_only=True)
with client.transaction(options=txn_options):
    ...
```

Likewise for passing in the prior transaction ID (note that the two cases are exclusive).

The tradeoff comes when the number of things to be configured gets larger, or when they interact in subtle ways.

dhermes: @tseaver IMO we "shouldn't" even expose the `previous_transaction` field to users. It should be something we use internally when doing retries **for** users.

chemelnucfin: @lukesneeringer Could you chime in on this issue?  Especially related to (future) upstream developments or YAGNI.

Also regarding the previous transaction (RunInTransaction)

chemelnucfin: @lukesneeringer @dhermes @tseaver 
I thought about this issue some more.  I think I will continue to support that TransactionOptions is a standalone object as opposed to a boolean.  My reasoning is based on the fact that the signature will be hard to change after it sets in stone.  If it was a boolean, it would be hard to change that should TransactionOptions change in the future, which I think is in the plans or they would have used a boolean also for the readonly.

dhermes: So I just discussed with @schmidt-sebastian and have concluded that `TransactionOptions` would be preparing for a future that will never come. Please use a boolean `read_only` keyword argument.

datastore/tests/system/test_system.py
chemelnucfin: @dhermes how is this different than the test above it?

dhermes: The test above it (`test_transaction_via_with_statement`) actually talks to the network.

chemelnucfin:
I mean, wouldn't it be good to have one here to make sure it doesn't reach
the network (like it is doing what it should do)?

dhermes:
No, we can test that **code path** with a unit test.

chemelnucfin:
@dhermes @tseaver PTAL.

datastore/google/cloud/datastore/client.py
```diff
@@ -134,7 +134,7 @@ def _extended_lookup(datastore_api, project, key_pbs,
         loop_num += 1
         lookup_response = datastore_api.lookup(
             project,
-            key_pbs,
+            keys=key_pbs,
```
dhermes: Please don't modify this. I intentionally changed it so that positional arguments would be passed as positional and keyword as keyword. (Also this is unrelated to your PR.)

datastore/google/cloud/datastore/client.py
```diff
@@ -511,6 +511,10 @@ def batch(self):
         """Proxy to :class:`google.cloud.datastore.batch.Batch`."""
         return Batch(self)
 
+    def read_only_transaction(self):
```
dhermes: Please remove this factory and add `**kwargs` to the `Client.transaction()` factory. In addition, you might update that docstring to just say the same arguments accepted by `Transaction()` are accepted as keywords.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -152,13 +153,22 @@ def Entity(*args, **kwargs):
...
-    def __init__(self, client):
+    def __init__(self, client, read_only=False):
         super(Transaction, self).__init__(client)
         self._id = None
+        if not read_only:
+            options = TransactionOptions()
+        else:
+            options = TransactionOptions(
+                read_only=TransactionOptions.ReadOnly())
+        self.options = options
```
dhermes: Please make this non-public (i.e. `self._options`)

datastore/google/cloud/datastore/transaction.py
```diff
@@ -152,13 +153,22 @@ def Entity(*args, **kwargs):
...
+    :param read_only: indicates the transaction is read only.
     """
 
     _status = None
 
-    def __init__(self, client):
+    def __init__(self, client, read_only=False):
         super(Transaction, self).__init__(client)
         self._id = None
+        if not read_only:
```
dhermes: This is definitely a nit-pick, but I would prefer you switch the order of the branches and drop the negation (it is unnecessary "work" to negate the boolean).

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +244,20 @@ def commit(self):
         finally:
             # Clear our own ID in case this gets accidentally reused.
             self._id = None
+
+    def put(self, entity):
+        """Ensures the transaction is not marked readonly and calls Batch.put
```
dhermes: 1. This should not be primary description of what `put()` does. I would suggest "Add an entity to be committed." as the one-liner and then move this "Ensures ..." bit into the long description.
1. "readonly" -> "read only"

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +244,20 @@ def commit(self):
...
+
+        This method calls :meth:`put` in the superclass, `Batch`, as long as
```
dhermes: This is user-facing documentation, so it's "inappropriate" to be discussing implementation details.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +244,20 @@ def commit(self):
...
+        the transaction is not marked read only.
+
+        :type entity: :class:`google.cloud.datastore.entity.Entity`
```
dhermes: I would suggest using a `~` before `google.cloud.datastore.entity.Entity` (it will make Sphinx only print `Entity` in rendered documentation)

datastore/google/cloud/datastore/transaction.py
```diff
@@ -234,3 +244,20 @@ def commit(self):
...
+        :param entity: the entity to be saved.
+
+        :raises: :class:`RuntimeError` if the transaction
+                 is marked ReadOnly
+        """
+        if not self.options.HasField('read_only'):
```
dhermes: Ditto here on my "don't use `not`" nitpick

datastore/google/cloud/datastore_v1/gapic/datastore_client.py
dhermes: @chemelnucfin Why is this change here? The autogen was updated in #4348 so you should not be touching these files.

datastore/google/cloud/datastore_v1/gapic/datastore_client.py
```diff
@@ -194,7 +194,7 @@ def lookup(self,
             ValueError: If the parameters are invalid.
         """
         request = datastore_pb2.LookupRequest(
-            project_id=project_id, keys=keys, read_options=read_options)
+            project_id=project_id, read_options=read_options, keys=keys)
```
dhermes: Ditto here. Please remove edits from autogenerated files.

datastore/tests/system/test_system.py
```diff
@@ -442,6 +443,17 @@ def test_transaction_via_with_statement(self):
         self.case_entities_to_delete.append(retrieved_entity)
         self.assertEqual(retrieved_entity, entity)
 
+    def test_readwrite_put(self):
```
dhermes: IMO this is an unnecessary test. Transactions are read-write by default and we already have several tests which exercise transactions.

datastore/tests/system/test_system.py
```diff
@@ -442,6 +443,17 @@ def test_transaction_via_with_statement(self):
...
+        entity = datastore.Entity(key=Config.CLIENT.key('Company', 'Google'))
+        options = types.TransactionOptions(
+            read_write=types.TransactionOptions.ReadWrite())
+        with Config.CLIENT.transaction() as xact:
+            xact.put(entity)
```
dhermes: You never actually use the `options` variable (PyLint probably complains but maybe `flake8` doesn't).

If you wanted to actually use it you'd need to so `xact.options = options` (after making it non-public `xact._options = options`)

datastore/tests/unit/test_client.py
```diff
@@ -353,7 +353,7 @@ def test_get_multi_miss(self):
         read_options = datastore_pb2.ReadOptions()
         ds_api.lookup.assert_called_once_with(
             self.PROJECT,
-            [key.to_protobuf()],
+            keys=[key.to_protobuf()],
```
dhermes: Ditto here (and throughout this file), please drop the `keys=`.

(Maybe you got it here as an artifact of a bad merge?)

datastore/tests/unit/test_client.py
```diff
@@ -934,6 +934,17 @@ def test_transaction_defaults(self):
             self.assertIs(xact, mock_klass.return_value)
             mock_klass.assert_called_once_with(client)
 
+    def test_read_only_transaction_defaults(self):
```
dhermes: Once you've deleted `Client.read_only_transaction()` and added `**kwargs` to `Client.transaction()` you can convert this test to cover the case where you pass `read_only=True`.

datastore/tests/unit/test_client.py
```diff
@@ -934,6 +934,17 @@ def test_transaction_defaults(self):
...
+        creds = _make_credentials()
+        client = self._make_one(credentials=creds)
+
+        patch = mock.patch(
+            'google.cloud.datastore.client.Transaction', spec=['__call__'])
```
dhermes: The `Transaction` constructor has no network requirements, so why do you mock it?

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
         self.assertIsNone(xact.id)
         self.assertEqual(ds_api.begin_transaction.call_count, 1)
 
+    def test_default(self):
```
dhermes: You don't need to add this test case, `test_ctor_defaults` already exists.

datastore/tests/unit/test_transaction.py
```diff
@@ -22,12 +22,18 @@ class TestTransaction(unittest.TestCase):
     @staticmethod
     def _get_target_class():
         from google.cloud.datastore.transaction import Transaction
-
         return Transaction
 
+    def _get_options_class(self, **kw):
```
dhermes: :+1: This adheres to our "import at run time" style

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
 
+    def test_default(self):
+        project = 'PROJECT'
+        id_ = 850302
+        ds_api = _make_datastore_api(xact=id_)
+        client = _Client(project, datastore_api=ds_api)
+        xact = self._make_one(client)
+        self.assertTrue(isinstance(xact.options, self._get_options_class()))
+
+    def test_read_only(self):
```
dhermes: Name instead `test_constructor_read_only`

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        client = _Client(project, datastore_api=ds_api)
+        read_only = self._get_options_class().ReadOnly()
+        options = self._make_options(read_only=read_only)
+        xact = self._make_one(client, read_only=True)
+        xact.begin()
+        begin = ds_api.begin_transaction
+        begin.assert_called_once()
+        self.assertEqual(begin.call_count, 1)
+
+    def test_read_write(self):
```
dhermes: Delete this test? It doesn't use `options` and isn't supported by the constructor.

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+
+    def test_read_only(self):
+        project = 'PROJECT'
+        id_ = 850302
+        ds_api = _make_datastore_api(xact=id_)
+        client = _Client(project, datastore_api=ds_api)
+        read_only = self._get_options_class().ReadOnly()
+        options = self._make_options(read_only=read_only)
+        xact = self._make_one(client, read_only=True)
+        xact.begin()
```
dhermes: You don't need to test `begin()`

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        self.assertTrue(isinstance(xact.options, self._get_options_class()))
+
+    def test_read_only(self):
+        project = 'PROJECT'
+        id_ = 850302
+        ds_api = _make_datastore_api(xact=id_)
+        client = _Client(project, datastore_api=ds_api)
+        read_only = self._get_options_class().ReadOnly()
+        options = self._make_options(read_only=read_only)
+        xact = self._make_one(client, read_only=True)
```
dhermes: You should check that `xact._options == options`

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        client = _Client(project, datastore_api=ds_api)
+        options = self._make_options(read_write=
+                                     self._get_options_class().ReadWrite())
+        xact = self._make_one(client)
+        xact.begin()
+        begin = ds_api.begin_transaction
+        begin.assert_called_once()
+        self.assertEqual(begin.call_count, 1)
+
+    def test_previous_tid(self):
```
dhermes: This test doesn't correspond to any code changes, please remove it.

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        client = _Client(project, datastore_api=ds_api)
+        options_klass = self._get_options_class()
+        options = self._make_options(read_only=options_klass.ReadOnly())
+        entity = _Entity()
+        xact = self._make_one(client, read_only=True)
+        xact.begin()
+        with self.assertRaises(RuntimeError):
+            xact.put(entity)
+
+    def test_put_read_write(self):
```
dhermes: This test doesn't correspond to any code changes, please remove it.

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        options = self._make_options(read_write=read_write)
+        entity = _Entity()
+        xact = self._make_one(client)
+        xact.begin()
+        xact.put(entity)
+        mutated_entity = _mutated_pb(self, xact.mutations, 'insert')
+        self.assertEqual(mutated_entity.key, entity.key.to_protobuf())
+
+
+def _mutated_pb(test_case, mutation_pb_list, mutation_type):
```
dhermes: Please remove this helper (only used by `test_put_read_write`, which should be removed)

datastore/tests/unit/test_transaction.py
```diff
@@ -212,6 +218,95 @@ class Foo(Exception):
...
+        begin.assert_called_once()
+        self.assertEqual(begin.call_count, 1)
+
+    def test_put_read_only(self):
+        project = 'PROJECT'
+        id_ = 943243
+        ds_api = _make_datastore_api(xact_id=id_)
+        client = _Client(project, datastore_api=ds_api)
+        options_klass = self._get_options_class()
+        options = self._make_options(read_only=options_klass.ReadOnly())
```
dhermes: `options` is unused.

datastore/google/cloud/datastore/client.py
chemelnucfin: I get errors relating to the tests if I don't change it though.  Should I modify the tests instead?

datastore/google/cloud/datastore/transaction.py
chemelnucfin: I was assuming that read_only is False by default, so I made that the first option.  I can change it if you want.

datastore/google/cloud/datastore/client.py
dhermes: That's because you've edited the `gapic` generated file. Please revert that edit (see my comment there).

datastore/google/cloud/datastore/transaction.py
dhermes: I want

datastore/google/cloud/datastore/client.py
chemelnucfin: The errors are this:
https://gist.github.com/chemelnucfin/eee00b41ba2ac8b8f10e1d68bb3047f6
and this:
https://gist.github.com/chemelnucfin/4b010867914d44473b2994b7b8cd6317

dhermes: That's because of your edit in the `gapic` generated file.

tseaver:
@chemelnucfin  [Lint failure](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4460).

chemelnucfin:
@tseaver PTAL, thanks.

datastore/tests/unit/test_client.py
```diff
@@ -934,6 +934,15 @@ def test_transaction_defaults(self):
...
             mock_klass.assert_called_once_with(client)
 
+    def test_read_only_transaction_defaults(self):
+        from google.cloud.datastore.transaction import Transaction
+        from google.cloud.datastore_v1.types import TransactionOptions        
+        creds = _make_credentials()
+        client = self._make_one(credentials=creds)
+        xact = client.transaction(read_only=True)
+        self.assertIsInstance(xact._options.read_only,
+                              TransactionOptions.ReadOnly)
```
dhermes: Just assert equality here. `__eq__` on protobuf objects will be defined as expected.

datastore/tests/unit/test_client.py
```diff
@@ -940,8 +940,8 @@ def test_read_only_transaction_defaults(self):
         creds = _make_credentials()
         client = self._make_one(credentials=creds)
         xact = client.transaction(read_only=True)
-        self.assertIsInstance(xact._options.read_only,
-                              TransactionOptions.ReadOnly)
+        self.assertEqual(xact._options.read_only,
+                         TransactionOptions.ReadOnly())
```
dhermes: @chemelnucfin Assert equality for all of `xact._options`

dhermes: Your latest commit still isn't doing `self.assertEqual(xact._options, foo)`

datastore/google/cloud/datastore/client.py
```diff
@@ -511,9 +511,13 @@ def batch(self):
         """Proxy to :class:`google.cloud.datastore.batch.Batch`."""
         return Batch(self)
 
-    def transaction(self):
-        """Proxy to :class:`google.cloud.datastore.transaction.Transaction`."""
-        return Transaction(self)
+    def transaction(self, **kwargs):
+        """Proxy to :class:`google.cloud.datastore.transaction.Transaction`.
+
+            :type kwargs: dict
```
dhermes: Sorry I just noticed this. It should be dedented to align with the `"""`.

chemelnucfin:
@tseaver any comments?

PR #4363, opened 2017-11-08
firestore: make a class for sentinel values

jba:
This way they'll print nicely.

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/constants.py
+++ b/firestore/google/cloud/firestore_v1beta1/constants.py
@@ -24,6 +24,13 @@ class sentinel(object):
     def __str__(self):
         return self._name
 
+    def __eq__(self, other):
+        raise NotImplementedError('use "is" to compare sentinels')
+
+    def __ne__(self, other):
+        raise NotImplementedError('use "is" to compare sentinels')
+
+
 DELETE_FIELD = sentinel('DELETE_FIELD')  # Sentinel object.
 """Sentinel value used to delete a field in a document."""
 
--- a/firestore/tests/unit/test_constants.py
+++ b/firestore/tests/unit/test_constants.py
@@ -12,20 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Helpful constants to use for Google Cloud Firestore."""
+import unittest
 
 
-class sentinel(object):
-    """Class for sentinel values, so they print nicely."""
+class TestSentinel(unittest.TestCase):
 
-    def __init__(self, name):
-        self._name = name
+    def test_str(self):
+        from google.cloud.firestore_v1beta1 import DELETE_FIELD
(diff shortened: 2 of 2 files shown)


dhermes:
@jba I'm having a hard time understanding the value here. Can you expound a bit on the motivation?

lukesneeringer:
Also, depending on your motivation, the implementation is probably incomplete. At minimum we probably need a `__repr__` and `__eq__` method.

How does this differ from `mock.sentinel` (other than you probably do not want to use a thing in the mock library in production code)?

jba:
Let's say you accidently put `SERVER_TIMESTAMP` somewhere in your data it doesn't belong, like in a list, and then give that data to `create`, `set` or `update`. Currently, this is the error you get:
```
TypeError: ('Cannot convert to a Firestore Value', <object object at 0x7ffbcf5b1b90>, 'Invalid type', <type 'object'>)
```
With this change, you'll see `SERVER_TIMESTAMP` instead of the raw object string.

I'm not familiar with `mock.sentinel`, but if there's an easier way to do this that would be fine with me.

I don't see why we'd need `__repr__`. I don't think we need `__eq__` either, because there are only two instances of this class, and they can be compared for identity with `==`. What I would like to do is prevent the user from ever creating another instance. I don't know how to do that in Python.

dhermes:
> can be compared for identity with `==`

One of the points of a sentinel is that you only use `is` for comparison, so if anything you'd want `__eq__` to error.

jba:
@dhermes Will do. I've obviously forgotten how equality works in Python.

@lukesneeringer Why do we need `__repr__`?

firestore/google/cloud/firestore_v1beta1/constants.py
```diff
@@ -15,8 +15,24 @@
 """Helpful constants to use for Google Cloud Firestore."""
 
 
-DELETE_FIELD = object()  # Sentinel object.
+class sentinel(object):
```
dhermes: @jba Can you name the class `_Sentinel` instead? (Still not sure I like this, but if it were to merge, that'd be a more preferable class name.)

@jonparrott Can you weigh in here?

tseaver:
@jba I'm still not convinced of the utility, but `__repr__` is more useful than `__str__` for debugging, which seems to be the dominant usecase:

```python
>>> class Foo(object):
...     def __str__(self):
...         return 'foo'
... 
>>> Foo()
<__main__.Foo object at 0x7fd2b4f85b90>
>>> class Bar(object):
...     def __repr__(self):
...         return '<bar>'
... 
>>> Bar()
<bar>
```

jba:
I'll hold off on this for now.

firestore/google/cloud/firestore_v1beta1/constants.py
jonparrott: Yeah style guide is `CamelCase` for classes, with a leading underscore for private classes.

Issue #4364, opened 2017-11-08
Additional spanner integration tests.


tseaver:
For spanner GA.  Investigate whether these are already tested, and add tests as appropriate.  Link either PR # or line reference of existing test.

- [x] Read single key
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6c602823e7101b40dfe5bfc614da633beb8accfb/spanner/tests/system/test_system.py#L840-L850
- [x] Read single key w/ index (#4403)
- [x] Range read w/ single key (#4573)
- [x] Range read w/ partial key (#4631)
- [x] Range read over index w/ open/open range (#4411)
- [x] Range read over index w/ open/closed range (#4411)
- [x] Range read over index w/ closed/open range (#4411)
- [x] Range read over index w/ closed/closed range (#4411)
- [x] Bind string to null (#4371)
- [x] Bind timestamp (#4371)
- [x] Bind date https://github.com/GoogleCloudPlatform/google-cloud-python/blob/f69d704841bd30c08af4a522397b3fe95107bdfd/spanner/tests/system/test_system.py#L673-L679
- [x] Bind string array (#4371)
- [x] Bind empty string array (#4371)
- [x] Query returning array of `-Inf`, `+Inf`, `Nan` (#4371)
- [x] Write / read back null string (#4372)
- [x] Write / read back empty array of string (#4372)
- [x] Write / read back null array of string (#4372)
- [x] Write / read back array of string with mixed null / non-null values (#4372)
- [x] Writing incorrect type of value should fail (#4569 )

/cc @lukesneeringer

tseaver:
@chemelnucfin Thanks for the updates!  As we each begin adding one or more related tests, let's adopt the covention of marking the checkbox items with our handles (to avoid duplicating effort).

tseaver:
@chemelnucfin I had to uncheck some of the boxes :( -- the test description was missing "*over index*".

chemelnucfin:
What is range read with partial key?

chemelnucfin:
Never mind, got it.

chemelnucfin:
Closed via PRs from above

PR #4371, opened 2017-11-09
Add spanner system tests:

tseaver:
- Bind timestamp.
- Bind string array.
- Bind empty string array.
- Bind string to null.
- Query returning array of -Inf, +Inf, NaN.

Toward #4364.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -1036,24 +1036,6 @@ class TestSessionAPI(unittest.TestCase, _TestData):
             expected=[(None,), (19,)],
         )
 
-        # Find -inf
-        self._check_sql_results(
-            snapshot,
-            sql='SELECT eye_d FROM all_types WHERE approx_value = @pos_inf',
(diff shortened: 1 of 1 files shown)


tseaver:
@dhermes I split the "transfinite" cases out into a separate method.  Code is otherwise unchanged:  I will merge when Circle is green.

tseaver:
Rebased to pick up hardening from merged #4374.

dhermes:
> Also, may I suggest to use origin (your own) for branching instead of upstream (gcp) to keep branches tidy?

This was intentional by @tseaver. When we push to the upstream remote, the system tests get run (they get skipped on PR builds otherwise).

But good advice otherwise!

chemelnucfin:
@dhermes wait, so if I make a PR the system tests don't run on CI?

dhermes:
@chemelnucfin That's correct. It's a security concern, since **anyone** can send a PR. Check out one of your builds, it'll say "skipped" by the system tests.

PR #4372, opened 2017-11-09
Add spanner system tests:

tseaver:
- Write / read back null string.
- Write / read back empty array of string.
- Write / read back null array of string.
- Write / read back array of string with mixed null / non-null values.

Toward #4364.


tseaver:
Rebased to pick up hardening from merged #4374.

PR #4374, opened 2017-11-09
Harden spanner system tests against stray databases

tseaver:
Because of quota, we are sharing the instance we use to test CRUD of databases: we therefore cannot rely on having only the expected databases present, as other tests may be running simultaneously, or may have failed
in ways which left 'stray' databases in the instance.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -274,12 +274,10 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
         # We want to make sure the operation completes.
         operation.result(30)  # raises on failure / timeout.
 
-        name_attr = operator.attrgetter('name')
-        expected = sorted([temp_db, self._db], key=name_attr)
-
-        databases = list(Config.INSTANCE.list_databases())
-        found = sorted(databases, key=name_attr)
-        self.assertEqual(found, expected)
+        database_ids = [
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -274,12 +274,10 @@ def test_create_database(self):
...
 
-        name_attr = operator.attrgetter('name')
-        expected = sorted([temp_db, self._db], key=name_attr)
-
-        databases = list(Config.INSTANCE.list_databases())
-        found = sorted(databases, key=name_attr)
-        self.assertEqual(found, expected)
+        database_ids = [
+            database.database_id
+                for database in Config.INSTANCE.list_databases()]
```
dhermes: @tseaver This 2nd indent seems "wrong"?

tseaver: I ended up wrapping because I was avoiding using `db` as the loop variable, channellng you. :)  How should it be wrapped?

dhermes: I'm not 100% but I'm pretty sure it should just be:

```python
database_ids = [
    database.database_id
    for database in Config.INSTANCE.list_databases()]
```

tseaver:
#4377 is blocking the run.  The system tests pass locally.

PR #4379, opened 2017-11-09
Auto-generate Trace v2

lukesneeringer:


dhermes:
Anything in particular you'd like reviewed?

liyanhui1228:
Thanks for generating this! Why removing the V1 related code? We probably need to keep both V1 and V2 library as the V1 API is still serving.

lukesneeringer:
Whups, sorry about that. Re-added the v1 guts.

Note that users explicitly wanting to use v1 will have to `from google.cloud import trace_v1`; just importing `trace` will get them v2.

trace/google/cloud/trace_v2/proto/tracing_pb2.py
```diff
@@ -0,0 +1,274 @@
...
+from google.protobuf import reflection as _reflection
+from google.protobuf import symbol_database as _symbol_database
+from google.protobuf import descriptor_pb2
+# @@protoc_insertion_point(imports)
+
+_sym_db = _symbol_database.Default()
+
+
+from google.api import annotations_pb2 as google_dot_api_dot_annotations__pb2
+from google.devtools.cloudtrace_v2.proto import trace_pb2 as google_dot_devtools_dot_cloudtrace__v2_dot_proto_dot_trace__pb2
```
liyanhui1228: Importing from `google.devtools` is causing the tests failure, can we rename it?

liyanhui1228:
CircleCI is blocked by error-reporting system tests. Verified locally that the trace tests passed. LGTM to merge.

PR #4380, opened 2017-11-10
Fix RuntimeError: set_exception can only be called once

PicardParis:


tseaver:
@PicardParis Thanks for the patch!  I'll defer to @lukesneeringer about the details, but we need to add one or more testcases / assertions which verify both new codepaths.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -159,8 +159,12 @@ def on_exception(self, exception):
         if getattr(exception, 'code', lambda: None)() == deadline_exceeded:
             return
 
-        # Set any other exception on the future.
-        self._future.set_exception(exception)
+        if self._future.done():
+            # Simply trigger the future.
+            self._future._trigger()
```
dhermes: @jonparrott Can we discuss?

1. It seems the `self._future.done()` check is necessary (but I'd love to have a case which can reliably cause this branch to occur, I certainly saw it a lot while [reproducing][1] other issues)
1. It seems "wrong" to call `_trigger()` a second/third/fourth/etc. time

jonparrott: Yeah this seems wrong. 

The future should be resolved with one of two situations:

1. The subscriber exits gracefully (`set_result(None)`).
2. The subscriber exits with an error here (`set_exception(exception)`).

I'm not sure why the result is being set twice, but it seems this is papering over that instead of fixing the fundamental issue (& I'm not sure why that's happening).

dhermes: OK cool. I think I may be able to reproduce over in #4444 so I will keep at it.

dhermes:
@PicardParis Thanks a lot for taking the time to write and send this patch. Reading it and writing tests for it certainly helped me understand the implementation as I started to dig in.

After debugging reproducible cases for #4463, it's clear that this approach is not the right one, so we'll be closing this PR.

Thanks!

PR #4381, opened 2017-11-10
Make 'project' optional / overridable for storage client

### LINKED ISSUE #4239: Vision: Google Cloud Storage Client project=None bootstrapping problem
Hello,

In using the Google Cloud API's, I noticed a small discrepancy in the documentation's expected behavior, and the actual behavior. 
The documentation and Google cloud examples code typically assumes that credentials should be created using a JSON key store saved on persistent storage. However, this isn't always an option, as with deployments using ephemeral storage. Luckily, a Credentials object can be created directly from a dictionary, which houses the same mapping that the JSON keystore would hold. This is the procedure that I use in my project (the specifics of which, is shown below).

The [Storage.Client documentation](https://googlecloudplatform.github.io/google-cloud-python/latest/storage/client.html#google.cloud.storage.client.Client) suggests that the `project` parameter is optional, and thus may be None. We found, however, that when it is not specified, an error is thrown, as a result of a sort of bootstrapping problem in which the Google Cloud APIs check to see if the project is None, and if so, attempt to create it (which is impossible, without first having the keys).
To elaborate:

1. When we make a call to Storage.Client(), we step into [storage/google/cloud/storage/client.py](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/). 
2. This class calls `super(Client, self).__init__(project=project, credentials=credentials, _http=_http)`, which takes us to [core/google/cloud/client.py](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/cloud/client.py) .
3. This calls `_ClientProjectMixin.__init__(self, project=project)`, where we see the following call:
```
        project = self._determine_default(project)
```
4. Following this call, we see:
```
    if project is None:
        _, project = google.auth.default()
```

So, there is an attempt to infer the project parameter based on the environment credentials. However, as discussed above, we are using a dictionary-created Credentials object as our credentials,  so no such credentials exist in the environment. **This is a problem because the documentation claims that project is optional, when, in fact, it is not if the project is not stored in the environment.**

Note, however, that this seems to be a problem specific with Storage.Client(), and not other API clients. Trying this procedure (see the code below) for `google.cloud.vision.ImageAnnotatorClient` seems to work perfectly fine with just a credentials object, without the project argument. 

----

1. **OS:** Ubuntu 16.04
2. **Python Version:** Python 2.7.10
3. **Google APIs:**
google-auth==1.1.1
google-cloud-core==0.27.1
google-cloud-storage==1.4.0
google-cloud-vision==0.27.0
google-gax==0.15.15
google-resumable-media==0.2.3
oauth2client==4.1.2

4. **Stacktrace:**

```
  File "storage_client.py", line 23, in <module>
    client = _get_client()
  File "storage_client.py", line 18, in _get_client
    client = storage.Client(credentials=creds)
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/cloud/storage/client.py", line 59, in __init__
    _http=_http)
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/cloud/client.py", line 211, in __init__
    _ClientProjectMixin.__init__(self, project=project)
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/cloud/client.py", line 165, in __init__
    project = self._determine_default(project)
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/cloud/client.py", line 178, in _determine_default
    return _determine_default_project(project)
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/cloud/_helpers.py", line 179, in _determine_default_project
    _, project = google.auth.default()
  File "/Users/admin/testproject/venv/lib/python2.7/site-packages/google/auth/_default.py", line 286, in default
    raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or
explicitly create credential and re-run the application. For more
information, please see
https://developers.google.com/accounts/docs/application-default-credentials.

```

5. **Steps to reproduce:**

- So, create an OAuth2 credentials object from a python dictionary (let's call it `ketfile_dict` using `google.oauth2.service_account.Credentials.from_service_account_info(keyfile_dict)`.

- Create an instance of a storage client using `google.cloud.storage.Client(credentials=creds)`

- Although the documentation states that the "project" argument is optional, the API will in fact throw an error when no project argument is presented.

6. Code example

Assume that in the same directory, you create a file, `constants.py`, with the following fields defined from the appropriate fields of the service account key JSON.

constants.GCP_SCOPE
constants.GCP_TYPE
constants.GCP_CLIENT_EMAIL
constants.GCP_PRIVATE_KEY
constants.GCP_PRIVATE_KEY_ID
constants.GCP_CLIENT_ID
constants.GCP_TOKEN_URI
constants.GCP_PROJECT_ID

```
from google.oauth2 import service_account
import constants


# Return an Oauth2 credentials instance
def get_credentials(with_keyfile=False):
    keyfile_dict = _get_keyfile_dict()
    scope = [constants.GCP_SCOPE]
    creds = service_account.Credentials.from_service_account_info(keyfile_dict).with_scopes(scope)

    if not with_keyfile:
        return creds
    else:
        # Return creds and the keyfile.
        return creds, keyfile_dict


# Construct a dictionary of the required key:value pairs.
def _get_keyfile_dict():
    keyfile_dict = {}
    keyfile_dict['type'] = constants.GCP_TYPE
    keyfile_dict['client_email'] = constants.GCP_CLIENT_EMAIL
    keyfile_dict['private_key'] = constants.GCP_PRIVATE_KEY
    keyfile_dict['private_key_id'] = constants.GCP_PRIVATE_KEY_ID
    keyfile_dict['client_id'] = constants.GCP_CLIENT_ID
    keyfile_dict['token_uri'] = constants.GCP_TOKEN_URI
    keyfile_dict['project_id'] = constants.GCP_PROJECT_ID
    return keyfile_dict


def _get_client():
    creds, keyfile_dict = get_credentials(with_keyfile=True)
    project = keyfile_dict['project_id']
    #client = storage.Client(project=project, credentials=creds) # This WILL work because a project is passed in.
    #client = vision.ImageAnnotatorClient(credentials=creds) #Note that the Vision client will work without a project argument
    client = storage.Client(credentials=creds) # This will throw an error because project=None.
    return client


# Instantiates a client
client = _get_client()
```

Let me know if you have any other questions. :)

tseaver:
Closes #4239.

### DIFF
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -228,7 +228,7 @@ class Client(ClientWithProject):
         except NotFound:
             return None
 
-    def create_bucket(self, bucket_name, requester_pays=None):
+    def create_bucket(self, bucket_name, requester_pays=None, project=None):
         """Create a new bucket.
 
         For example:
@@ -250,13 +250,18 @@ class Client(ClientWithProject):
             (Optional) Whether requester pays for API requests for this
             bucket and its blobs.
 
+        :type project: str
+        :param project: (Optional) the project under which the  bucket is to
+                        be created.  If not passed, uses the project set on
+                        the client.
+
         :rtype: :class:`google.cloud.storage.bucket.Bucket`
         :returns: The newly created bucket.
         """
         bucket = Bucket(self, name=bucket_name)
         if requester_pays is not None:
             bucket.requester_pays = requester_pays
-        bucket.create(client=self)
+        bucket.create(client=self, project=project)
         return bucket
 
     def list_buckets(self, max_results=None, page_token=None, prefix=None,
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -260,7 +260,7 @@ class Client(ClientWithProject):
         return bucket
 
     def list_buckets(self, max_results=None, page_token=None, prefix=None,
-                     projection='noAcl', fields=None):
+                     projection='noAcl', fields=None, project=None):
         """Get all buckets in the project associated to the client.
 
         This will not populate the list of blobs available in each
@@ -296,11 +296,24 @@ class Client(ClientWithProject):
             response with just the next page token and the language of each
             bucket returned: 'items/id,nextPageToken'
 
+        :type project: str
+        :param project: (Optional) the project whose buckets are to be listed.
+                        If not passed, uses the project set on the client.
+
         :rtype: :class:`~google.api_core.page_iterator.Iterator`
+        :raises ValueError: if both ``project`` is ``None`` and the client's
+                            project is also ``None``.
         :returns: Iterator of all :class:`~google.cloud.storage.bucket.Bucket`
                   belonging to this project.
         """
-        extra_params = {'project': self.project}
+        if project is None:
+            project = self.project
+
+        if project is None:
+            raise ValueError(
+                "Client project not set:  pass an explicit project.")
+
+        extra_params = {'project': project}
 
         if prefix is not None:
             extra_params['prefix'] = prefix
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -26,10 +26,13 @@ from google.cloud.storage.batch import Batch
 from google.cloud.storage.bucket import Bucket
 
 
+_marker = object()
+
+
 class Client(ClientWithProject):
     """Client to bundle configuration needed for API requests.
 
-    :type project: str
+    :type project: str or None
     :param project: the project which the client acts on behalf of. Will be
                     passed when creating a topic.  If not passed,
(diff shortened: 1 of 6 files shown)


storage/google/cloud/storage/client.py
```diff
@@ -26,10 +26,13 @@
 from google.cloud.storage.bucket import Bucket
 
 
+_marker = object()
```
jonparrott: Nit (& only change if you end up modifying something else in this PR): We've called this kind of thing `_SENTINEL` elsewhere.

PR #4386, opened 2017-11-13
read multiple keys with index

chemelnucfin:
reads multiple keys using index


dhermes:
@chemelnucfin Don't forget to add relevant labels to PRs as well as issues

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,33 @@ def _unit_of_work(transaction, test):
 
         return session, committed
 
+    def test_read_with_multiple_keys_index(self):
+        ROW_COUNT = 10
+        EXTRA_DDL = ['CREATE INDEX name ON contacts(first_name, last_name)']
+        MY_COLUMNS = self.COLUMNS[1], self.COLUMNS[2]
```
dhermes: `MY_` isn't a great prefix for a variable name. Just `COLUMNS` seems fine

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,33 @@ def _unit_of_work(transaction, test):
...
+        pool = BurstyPool()
+
+        temp_db = Config.INSTANCE.database(
+            'test_read_multi_keys_index', ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
```
dhermes: Please make sure your lines are under 80 chars. I'd also suggest configuring your editor so this is "obvious".

tseaver: Because quotas prevent us from using per-CI-run instances, this test needs to mangle the database ID with `unique_resource_id`.

Another, maybe better strategy might be to just add the `CRETATE INDEX` DDL to the "stock" set in `tests/_fixtures.py`.

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,33 @@ def _unit_of_work(transaction, test):
...
+            pool=pool)
+        operation = temp_db.create()
+        self.to_delete.append(_DatabaseDropper(temp_db))
+        operation.result(90)
+        
+        session, committed = self._set_up_table(ROW_COUNT, db=temp_db)
+        self.to_delete.append(session)
+        expected = list(
+            [[row[1], row[2]] for
+                index, row in enumerate(self._row_data(ROW_COUNT))])
```
tseaver: No need to wrap a list comprehension in `list()` -- it is one already.

Also, no need for `enumerate` (we aren't using `index` anyway).  So, this could just be:

```python
        expected = [(row[1], row[2]) for row in self._row_data(ROW_COUNT)]
```

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,33 @@ def _unit_of_work(transaction, test):
...
+        keyset = [(expected[index][0], expected[index][1]) for
+                      index in range(len(expected))]
```
tseaver: Again, no need for the index.  Just:

```python
        keyset = [(row[0], row[1]) for row in expected]
```

Since `expected` is just a list of the tuples anyway, you could just use it instead (below, passed to the `Keyset` constructor).

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,33 @@ def _unit_of_work(transaction, test):
...
+        rows = list(session.read(self.TABLE,
+                                 MY_COLUMNS,
+                                 KeySet(keys=keyset),
+                                 index='name'))
+        temp_db.drop()
```
tseaver: If we're using a temporary, we need a `try:  ... finally: temp_db.drop()` to ensure it gets dropped even if the intermediate code raises.

tseaver:
Ugh, @chemelnucfin my review comments predated your force push.  They are still probably valid.

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,40 @@ def _unit_of_work(transaction, test):
...
+
+        temp_db = Config.INSTANCE.database(
+            'test_read_multi_keys_index',
+            ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
+            pool=pool
+        )
+        if not temp_db.exists(): # prevent Conflicts 
+            operation = temp_db.create()
+            self.to_delete.append(_DatabaseDropper(temp_db))
+            operation.result(90)
```
tseaver: Because quotas prevent us from using per-CI-run instances, this test needs to mangle the database ID with unique_resource_id.

Another, maybe better strategy might be to just add the CRETATE INDEX DDL to the "stock" set in tests/_fixtures.py.

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,40 @@ def _unit_of_work(transaction, test):
...
+        
+        session, committed = self._set_up_table(ROW_COUNT, db=temp_db)
+        self.to_delete.append(session)
+        expected = list(
+            [[row[1], row[2]] for
+            index, row in enumerate(self._row_data(ROW_COUNT))]
+        )
```
tseaver: No need to wrap a list comprehension in list() -- it is one already.

Also, no need for enumerate (we aren't using index anyway). So, this could just be:

```python
        expected = [(row[1], row[2]) for row in self._row_data(ROW_COUNT)]
```

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,40 @@ def _unit_of_work(transaction, test):
...
+        keyset = [
+            (expected[index][0], expected[index][1]) for
+            index in range(len(expected))
+        ]
```
tseaver: Again, no need for the index. Just:

```python
        keyset = [(row[0], row[1]) for row in expected]
```

Since expected is just a list of the tuples anyway, you could just use it instead (below, passed to the `Keyse`t constructor).

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,40 @@ def _unit_of_work(transaction, test):
...
+        rows = list(session.read(self.TABLE,
+                                 COLUMNS,
+                                 KeySet(keys=keyset),
+                                 index='name')
+        )
+        temp_db.drop()
```
tseaver: If we're using a temporary, we need a `try: ... finally: temp_db.drop()` to ensure it gets dropped even if the intermediate code raises.

spanner/tests/system/test_system.py
chemelnucfin: you are right, I think I was trying to work on a certain set of keys before and I didn't think it through when I was simplifying it.

spanner/tests/system/test_system.py
chemelnucfin: I either get
`=================================== FAILURES ===================================
______________ TestSessionAPI.test_read_with_multiple_keys_index _______________
  File "/home/aw/projects/google-cloud-python/spanner/tests/system/test_system.py", line 738, in test_read_with_multiple_keys_index
    session, committed = self._set_up_table(ROW_COUNT)
  File "/home/aw/projects/google-cloud-python/spanner/tests/system/test_system.py", line 716, in _set_up_table
    retry(db.reload)()
  File "/home/aw/projects/google-cloud-python/spanner/.nox/sys-2-7/local/lib/python2.7/site-packages/test_utils/retry.py", line 205, in wrapped_function
    raise BackoffFailed()
BackoffFailed
----------------------------- Captured stdout call -----------------------------
_has_all_ddl. Trying again in 1 seconds...
_has_all_ddl. Trying again in 2 seconds...
_has_all_ddl. Trying again in 4 seconds...
_has_all_ddl. Trying again in 8 seconds...
============================= 41 tests deselected ==============================`
or
`
______________ TestSessionAPI.test_read_with_multiple_keys_index _______________
  File "/home/aw/projects/google-cloud-python/spanner/tests/system/test_system.py", line 737, in test_read_with_multiple_keys_index
    operation.result(90)
  File "/home/aw/projects/google-cloud-python/spanner/.nox/sys-2-7/local/lib/python2.7/site-packages/google/gax/__init__.py", line 595, in result
    if not self._poll(timeout).HasField('response'):
  File "/home/aw/projects/google-cloud-python/spanner/.nox/sys-2-7/local/lib/python2.7/site-packages/google/gax/__init__.py", line 705, in _poll
    return retryable_done_check()
  File "/home/aw/projects/google-cloud-python/spanner/.nox/sys-2-7/local/lib/python2.7/site-packages/google/gax/retry.py", line 143, in inner
    raise exc
RetryError: RetryError(Retry total timeout exceeded with exception, caused by Deadline Exceeded)
============================= 41 tests deselected ==============================
`

chemelnucfin:
@tseaver PTAL, thanks.

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,28 @@ def _unit_of_work(transaction, test):
 
         return session, committed
 
+    def test_read_with_multiple_keys_index(self):
+        row_count = 10
+        ddl_create = ['CREATE INDEX name ON contacts(first_name, last_name)']
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        try:
+            operation = self._db.update_ddl(ddl_create)
```
tseaver: Is there a specific reason not to just add that index to the default DDL in `tests/_fixtures.py`?  That would make this test much simpler / clearer (it would contain only fourstatements and the final assertion).

spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,28 @@ def _unit_of_work(transaction, test):
...
+            operation.result(120)
+            session, committed = self._set_up_table(row_count, db=self._db)
+            self.to_delete.append(session)
+            expected = [[row[1], row[2]] for row in self._row_data(row_count)]
+            keyset = [(row[0], row[1]) for row in expected]
```
tseaver: This is still redundant:  `keyset` is exactly equal to `expected`.

spanner/tests/system/test_system.py
chemelnucfin: No reason, I just wasn't sure if I should have changed the actual fixture instead of temporary changing it.  
Added it permanently to the fixtures.

spanner/tests/system/test_system.py
chemelnucfin: There was a time where I believe there was a mismatch between tuple and lists on the output side, so I think that was what I changed.  But you are right and I now removed the redundancy.

chemelnucfin:
@tseaver PTAL, thanks.

PR #4389, opened 2017-11-13
Post #4350 cleanup: Missed a rename of a test helper due to concurrent merges

dhermes:
This happened because #4350 was sent before #4355 was merged.

/cc @alixhami This may have been avoidable (e.g. with a rebase) but since there were no changes needed in the review, it's likely that this was just unavoidable (except for superhumans).


dhermes:
I merge before CI essentially to unbreak all downstream merges to master.

Issue #4378, opened 2017-11-09
firestore: add a FieldPath class


jba:
The firestore client conflates dot-separated strings with field paths in the `update` methods. There should be two cases for top-level dict keys passed to `update`:

- If the key is a string, it is split on dots and cannot contain any of the special characters `~*/[]`. There also can't be any empty parts. One reasonable approach is:
    ```
    # validate that there are no special characters in s
    fp = FieldPath(s.split('.'))
    ```
- If the key is a `FieldPath`, it is treated as-is.

Where `FieldPath` is something like:
```
class FieldPath(object):
    def __init__(self, *parts):
        # Verify that each part is a string, and that no part is the empty string.

    def _to_api_repr(self):
        # Apply Firestore escaping rules, using backticks and backslashes for parts
        # that do not match ^[A-Za-z_][A-Za-z_0-9]*$.
        # E.g. FieldPath("a", "~", "`").to_api_repr() => "a.`~`.`\``"

    def __eq__(self): # and __ne__ and whatever else is needed to be a dict key.
```
Since it is now possible to pass a dict where the same effective path occurs twice, that needs to be checked explicitly.
E.g.
```
{"a": 1, FieldPath("a"): 2}
```
should be an error.

@schmidt-sebastian Confirm this is accurate.

schmidt-sebastian:
Yes, this is accurate.

chemelnucfin:
@jba I would like some clarifications on the FieldPath (FP).

1.  It seems like I should be able to take either a string, or an iterable and FP will figure out what to do.  So either FP("a.b.c"), FP("a", "b", "c") and FP(["a", "b", "c"]) should all work and return the same thing, but I assume FP(["a"], "b", "c") will raise ValueError.  Is this correct?

2.  You mentioned above that ~ is an invalid character, but you allowed FieldPath to have "~" in the constructor.  I would like to confirm that that "~" is only invalid inside a string such as FP("a.~.c"), but valid in FP("a", "~", "b")

3.  How do backslashes get handled in api_repr()?  Would "\\" (single slash) return "\\" or "`\\`?  I assume that "\\\\" doesn't happen in paths, so how do you want that to be handled?  An error, or "`\\``\\`" or something else?

4.  Do unicode characters fit into your regex requirements?

5.  Not exactly sure where to check for the same effective path, but I haven't looked at that part of the code yet.

Thanks in advance.

jba:
[...]

PR #4392, opened 2017-11-14
#4378 - Field Path

chemelnucfin:
@jba please see if this is the desired behavior you wanted or feel free to give me any comments.


jba:
@schmidt-sebastian, please take a look.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+class FieldPath(object):
+    """ Field Path object for client use.
+
+    Args:
+        parts (string or iterable): Indicating path of the key to be used.
+    """
+
+    def __init__(self, *parts):
+        invalid_characters = '~*/[]'
+        if len(parts) == 1:
```
jba: I don't think there should be special splitting behavior when len(parts) == 1. When I write `FieldPath("a.b")`, I mean a field path of one component which is `a.b`. The place for splitting is in `update`.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+            parts = parts[0]
+        if isinstance(parts, basestring):
+            parts = tuple(parts.split('.'))
+        for part in parts:
+            if (not part
+                or not isinstance(part, basestring)
+                or part in invalid_characters):
+                raise ValueError
```
jba: Give some help in a string arg to `ValueError`.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+        self.parts = tuple(parts)
+
+    def to_api_repr(self):
+        """ Returns string representation of the FieldPath
+
+        Returns: string representation of the path stored within
+        """
+        pattern = re.compile(r'[A-Za-z_][A-Za-z_0-9]*')
```
jba: Can you do this globally to avoid a regexp compile on each call?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+
+    Args:
+        parts (string or iterable): Indicating path of the key to be used.
+    """
+
+    def __init__(self, *parts):
+        invalid_characters = '~*/[]'
+        if len(parts) == 1:
+            parts = parts[0]
+        if isinstance(parts, basestring):
```
schmidt-sebastian: If this is intended to be part of the public API, then I would like to suggest to separate the two cases here more clearly:

If we pass a string, we split on dots and reject ~*/[]. If we pass a list of strings, we don't.

So, calling FieldPath with (["foo.bar"]) should result in a field path of `foo.bar`. 

The FieldPath constructor should ideally never split on dots.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+            parts = tuple(parts.split('.'))
+        for part in parts:
+            if (not part
+                or not isinstance(part, basestring)
+                or part in invalid_characters):
```
schmidt-sebastian: This limitation only applies when the field path is supplied using a string via any of the API method that consume these paths.

`query.where('~', '==', 'foo')` is invalid, but `query.where(new FieldPath('~'), '==', 'foo')` is not.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+                part_ans = ''
+                for letter in part:
+                    if letter == '\\':
+                        part_ans += '\\'
+                    elif letter == '`':
+                        part_ans += '\`'
+                    else:
+                        part_ans += letter
+                ans.append('`' + part_ans  + '`')
+        return '.'.join(ans)
```
schmidt-sebastian: This looks good.

firestore/tests/unit/test__helpers.py
```diff
@@ -86,6 +86,116 @@ def test___ne__type_differ(self):
...
+        from google.cloud.firestore_v1beta1._helpers import FieldPath
+        return FieldPath
+
+    def _make_one(self, *args, **kwargs):
+        klass = self._get_target_class()
+        return klass(*args, **kwargs)
+
+    def test_empty_string_inside_string_fails(self):
+        with self.assertRaises(ValueError):
+            field_path = self._make_one('a..b')
```
schmidt-sebastian: This is a valid path and should be represented as `a..b`

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,56 @@ def __ne__(self, other):
...
+        Returns: string representation of the path stored within
+        """
+        pattern = re.compile(r'[A-Za-z_][A-Za-z_0-9]*')
+        ans = []
+        for part in self.parts:
+            match = re.match(pattern, part)
+            if match:
+                ans.append(part)
+            else:
+                part_ans = ''
```
jba: This is quadratic. Maybe you can do something with replace. See the [node implementation](https://github.com/googleapis/nodejs-firestore/blob/master/src/path.js#L520).

firestore/tests/unit/test__helpers.py
chemelnucfin: @schmidt-sebastian 
But if we split on dots, wouldn't this return `['a', '', 'b']`?

firestore/google/cloud/firestore_v1beta1/_helpers.py
jba: This confusion goes away if `FieldPath` always constructs a field path from component strings.

jba: I actually missed this in my review. I think what you want is a static method `from_string`, so that `FieldPath.from_string` checks for invalid characters and splits on dots, then constructs a `FieldPath`. Call that in `update`, `query`, etc. after doing an `isinstance` to distingish a string from a `FieldPath`. All the internal logic can then work solely with `FieldPath`s.

firestore/tests/unit/test__helpers.py
jba: Another reason why len(parts)==1 should not be a special case.

schmidt-sebastian: Please take a look at Node's `fromDotSeparatedString`: https://github.com/googleapis/nodejs-firestore/blob/d85f33e9b8538f34ac3f2d9a07495bd32a65c325/src/path.js#L326. This is a special entry point that we use internally. The public entry point to FieldPath does not split on dots.

`query.where('a.b', ... )` calls `fromDotSeparatedString` and represents `{a : b: ... }` . `query.where(new FieldPath('a.b'), ... )` represents `{ a.b : ...}`

chemelnucfin:
@jba @schmidt-sebastian PTAL.  

I used the string replace method like node.  Please check my tests to see that they are still the desired behavior.  

In addition, what about unicode?  

I will work on the update and query methods in a separate PR.

chemelnucfin:
I will wait on updating `update` and `query` methods until I get more confirmation about the unicode issue.

chemelnucfin:
@jba @schmidt-sebastian 
So I'm looking through the different languages, and I seem different things for each language.  For example, the function api_repr that I was implementing is called 
toServiceFieldPathComponent in Go and C#, 
but canonicalString in Java and Node

I understand each language has different quirks and that each user would probably not be using cross language.  However, shouldn't something like this be consistent across all languages?  

Also, according to the docs that was here: 
https://cloud.google.com/firestore/docs/reference/rest/v1beta1/Document
it doesn't seem like unicode is allowed, but is there a reason for that?

Thanks.

jba:
The doc is now at https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1beta1#google.firestore.v1beta1.Document and it explicitly allows unicode keys (it refers to UTF-8). 

It's not important that the names of unexposed functions are the same across languages. We only care about the exposed surface and the behavior.

firestore/tests/unit/test__helpers.py
```diff
@@ -86,6 +86,130 @@ def test___ne__type_differ(self):
...
+        parts = '\\'
+        field_path = self._make_one(parts)
+        self.assertEqual(r'`\\`', field_path.to_api_repr())
+
+    def test_to_api_repr_double_slash(self):
+        parts = r'\\'
+        field_path = self._make_one(parts)
+        self.assertEqual(r'`\\\\`', field_path.to_api_repr())
+
+    def test_to_api_repr_underscore_valid(self):
```
jba: Actually, the [doc](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1beta1#google.firestore.v1beta1.Document) says a key that starts with an underscore is not a simple field name, so it should be quoted. This is a bug in the Go implementation and perhaps others.

firestore/tests/unit/test__helpers.py
```diff
@@ -86,6 +86,130 @@ def test___ne__type_differ(self):
...
+        parts = ['a', 'b', 'c']
+        with self.assertRaises(ValueError):
+            field_path = self._make_one(parts)            
+
+    def test_tuple_fails(self):
+        parts = ('a', 'b', 'c')
+        with self.assertRaises(ValueError):        
+            field_path = self._make_one(parts)
+
+    def test_key(self):
```
jba: This test is fine as it is, but it doesn't handle the case I was worried about. In `update` you'll accept a dict that can have either strings or FieldPaths as keys. You will then loop through the keys and call `FieldPath.from_string` on the strings. My concern is that someone can pass this to update:
```
{"a.b": 1, FieldPath("a", "b"): 2}
```
Those keys are not equal, but after you call `from_string` on the first they will denote the same field. I think you have to explicitly check for that in `update`.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,57 @@ def __ne__(self, other):
...
+                ans.append(part)
+            else:
+                replaced = part.replace('\\', '\\\\').replace('`', '\\`')
+                ans.append('`' + replaced + '`')
+        return '.'.join(ans)
+
+    def __hash__(self):
+        return hash(self.to_api_repr())
+
+    def __eq__(self, other):
```
jba: You need to follow a more complicated protocol to handle cases where `other` is a different type. Take a look at some of the `__eq__` methods in bigquery as examples.

chemelnucfin:
@jba 
Thank you for your comments.  My apologies regarding the unicode issue.  It was not clear from the documentation that it was an alternative representation.  I might suggest a different wording later, but let me work on your comments first.

firestore/tests/unit/test__helpers.py
jba: Belay that: initial underscore are OK, the doc is wrong.

chemelnucfin:
got it.  I'll have a new commit soon.  I'll squash from HEAD to HEAD^^
after I'll getting your review on the diff from HEAD to HEAD^.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -127,17 +127,27 @@ class FieldPath(object):
...
             Indicating path of the key to be used.
     """
-    pattern = re.compile(r'[A-Za-z_][A-Za-z_0-9]*')
+    simple_field_name = re.compile(r'[A-Za-z_][A-Za-z_0-9]*')
 
     def __init__(self, *parts):
         for part in parts:
-            if not isinstance(part, str):
-                raise ValueError("One or more components is not a string")
+            try:
```
jba: @tswast @lukesneeringer Is this kosher, or should the `six` module be involved here somewhere?

firestore/tests/unit/test__helpers.py
chemelnucfin: I will address this in the update code.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -127,17 +127,27 @@ class FieldPath(object):
...
+            if not isinstance(part, string):
+                raise ValueError("One or more components is not a string.")
         self.parts = tuple(parts)
 
-    @classmethod
-    def from_string(cls, string):
-        """ Creates a FieldPath with a string representation.
+    @staticmethod
+    def from_string(string):
+        """ Creates a FieldPath from an unicode string representation.
```
jba: " a unicode", not "an unicode". Stupid English language.

jba: (I think it's because "unicode" is pronounced as if it had an initial consonant: "yoonicode".)

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -127,17 +127,27 @@ class FieldPath(object):
...
+
+        Args:
+            :type string: str
+            :param string: An unicode string which cannot contain
```
jba: Mention the split on dots.

firestore/google/cloud/firestore_v1beta1/_helpers.py
dhermes: Use of `str` should almost never occur (and likely **never** in `google-cloud-python`). Here it seems `six.string_types` should be used? But it's also possible that `six.text_type` is desired?

chemelnucfin: I changed it to six.string_types.  Thought six might have had something like that, but a few days ago I didn't notice it on here: http://python-future.org/compatible_idioms.html and didn't check six and figured it wouldn't have been hard to write it that way.

chemelnucfin:
@dhermes Could you please take a quick look on this?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -119,6 +120,69 @@ def __ne__(self, other):
...
+                           `~*/[]` characters, cannot exceed 1500 bytes,
+                           and cannot be empty.
+
+        Returns:
+            A :class: `FieldPath` instance with the string split on "."
+            as arguments to `FieldPath`.
+        """
+        invalid_characters = '~*/[]'
+        string = string.split('.')
+        for part in string:
```
schmidt-sebastian: Can we move this validation up one line and validate the unsplit input string? Empty string should be caught by line 135 already.

chemelnucfin:
@schmidt-sebastian I moved the validation and did the empty string check in the constructor.  Please take a look at my 3rd review changes (4th commit) and see if that's what you meant.  Line 135 did not check for an empty string before.

schmidt-sebastian:
> @schmidt-sebastian I moved the validation and did the empty string check in the constructor. Please take a look at my 3rd review changes (4th commit) and see if that's what you meant. Line 135 did not check for an empty string before.

Looks good to me. Thanks.

chemelnucfin:
@dhermes?

dhermes:
Sorry I don't have cycles for this right now. @tseaver can you tap in? (In particular, we want to avoid re-inventing the wheel for code that already does this.)

chemelnucfin:
@dhermes it's ok.  I am just trying to clean out some issues/prs.

PR #4393, opened 2017-11-14
BigQuery: added methods for getting keys, items and dict

josefbarta:
This change enables to retrieve the row as a dict and iterate the keys and/or items, like with a normal dict; in other words, making the Row object a dict-like object:

>>> row.dict()
{'name': 'Isabel', 'profession': 'bridge builder'}
>>> for k, v in row.items():
>>> for k in row.keys():


josefbarta:
I signed it!

dhermes:
I'd like to do a thorough review of the code (e.g. `keys()` and `items()` could be generators) but want to wait until @tswast comments on whether or not these features are desired.

josefbarta:
I see, good point about the generators, thanks. I'd fix that but will hold off until we know more about whether these are desirable at all. The thing is that it would be great to be able to either return the row as a dict or convert it to a dict really easily, which currently is a bit of a hacky process. Thanks!

josefbarta:
I fixed it after all; the tests seem to have run successfully :-)

tswast:
I like `keys()` and `items()` to make `Row` act even more like a dictionary. That part LGTM.

What is your desired use for `dict()`? Are you using it to serialize the rows?

josefbarta:
I'm using `dict()` to serialise rows into a standard dict, which I can then convert to json. As long as `keys()` and/or `items()` makes it in, I'm happy not to have `dict()`, simply because I can do it in my own code. Alternatively, I'm happy to write a `get()` method for `Row` instead, so it's possible to get a value in a dict-standard failsafe way. Thanks!

tswast:
> As long as `keys()` and/or `items()` makes it in, I'm happy not to have `dict()`.

Yeah, I'd prefer not to have `dict()` since it's not a standard method and it's easy enough for someone to make their own dictionary now that there is `keys()` and `items()`.

> Alternatively, I'm happy to write a get() method for Row instead, so it's possible to get a value in a dict-standard failsafe way. Thanks!

Ooh, yes please to adding the `get()` method. That sounds great.

josefbarta:
I signed it!

josefbarta:
So `dict()` is now removed and `get()` is in place, including code checks and the tests for the module check out (not yet completed for the whole hog but that should be fine). Please let me know if there are any issues with my code I should fix. Thanks!

bigquery/google/cloud/bigquery/table.py
```diff
@@ -791,30 +791,40 @@ def keys(self):
...
+        'a'
+        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z')
+        None
+        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z', '')
+        ''
+        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z', default = '')
+        ''
+        """
+        index = self._xxx_field_to_index.get(key)
+        if index is None:
```
tswast: I think this might not quite be right in the case where the key exists, but is explicitly set to None.

```
>>> d = {'x': None}
>>> y = d.get('x', 'default')
>>> print(y)
None
```

tswast: Nevermind. I see this is checking if the index is None, not the value.

josefbarta:
Thanks :-)

Issue #4267, opened 2017-10-26
Move google-cloud-core functionality into api_core


jonparrott:
- [ ] [IAM](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/895951a7ae244634f44debee27597ef384b6c138/core/google/cloud/iam.py) (Related #3882)
- [ ] [Helpers](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/895951a7ae244634f44debee27597ef384b6c138/core/google/cloud/_helpers.py) - many of these will just be distributed among the existing helper modules in api_core.
- [ ] [Connection/JSONConnection](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4c24da38f1a8d9a3cc6422cf0c20495880d1fe38/core/google/cloud/_http.py)
- [ ] [Client base](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4c24da38f1a8d9a3cc6422cf0c20495880d1fe38/core/google/cloud/client.py) - (Consider eliminating in favor of factory mixins and helpers?)

jonparrott:
Scattered thoughts on stuff for my own future benefit (feel free to comment, but just know these are raw).

Clients:

1. We should just drop the common base class and depend on google.auth.default() for both default credential and project ID detection.
2. We should add a mixin class for `from_service_account_file()` and `from_service_account_info()`, with a deprecated alias for `from_service_account_json`. This can be used by gapic clients as well.

Connection/JSONConnection:

1. Stop using inheritance for configuration.
1. We should try to just treat this the same we do gRPC channels in gapic clients. The clients should accept a `connection`  argument and construct their own if not provided.
1. Connections should be reduced to just a basic class with an `http` property and a `request` method and made public.
1. The `request` method should handle serialization, deserialization, exceptions, and injecting clientinfo headers.
1. Storage seems to be the only client that does special things with the connection, so the bulk of the existing layers of code should be moved into storage. It's special enough to warrant that.
1. Some classes write "Stub"-like classes (like `_HTTPVisionAPI`). This is fine and aligns with the plans above.

dhermes:
@jonparrott I actually massively reduced the interface of `Connection` with plans to remove it (and I made it a non-public member intentionally). So I am all for it!

jonparrott:
Okay, here's my proposal for removing `google.cloud.core.client` and somewhat addressing #4840.

[...]

PR #4399, opened 2017-11-15
Move datetime helpers from google.cloud._helpers to google.api_core.datetime_helpers

jonparrott:
Towards #4267


jonparrott:
After chatting with @dhermes, we determined that while we could technically remove all of the `if pytz` statements from cloud core, it's best to just not modify that library. So instead I'm going to just `pragma: NO COVER` those lines as a lesser of several evils scenario.

dhermes:
Still LGTM, but let's wait for green?

jonparrott:
Ya

PR #4401, opened 2017-11-16
Spanner: Tests with None

chemelnucfin:
I looked at the INT64 implementation and think I did the right thing that the SQL queries always return nothing.  Please let me know if this is not right.

I'll add more tests to other PRs later.


chemelnucfin:
Was told not to do non-integration 7 tests

PR #4402, opened 2017-11-16
The subscription example in the readme of Pub/Sub has been updated.

mehmetboraezer:
I need to use pub/sub client. But this example out of date, it's not working. I tried to fix it. It only contains minor changes in the readme file.

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -98,7 +98,8 @@ the topic, and subscribe to that.
         project_id=os.getenv('GOOGLE_CLOUD_PROJECT'),
         sub='MY_SUBSCRIPTION_NAME',  # Set this to something appropriate.
     )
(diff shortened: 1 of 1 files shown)


tseaver:
@mehmetboraezer Thanks for the fix!

pubsub/README.rst
```diff
@@ -98,7 +98,8 @@ the topic, and subscribe to that.
         project_id=os.getenv('GOOGLE_CLOUD_PROJECT'),
         sub='MY_SUBSCRIPTION_NAME',  # Set this to something appropriate.
     )
-    subscriber.create_subscription(name=subscription_name, topic=topic_name)
```
mehmetboraezer: Hello @tseaver. I don't understand why we should split this line? It has 72 characters. I checked google style guide and maximum line length is 80. Just I want to understand the reason. :)

PR #4403, opened 2017-11-16
Spanner: Read single key with index

chemelnucfin:
I'm trying for smaller PRs to simplify debugging/code writing for other issues.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -287,12 +287,12 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
         self.to_delete.append(temp_db)
 
         # We want to make sure the operation completes.
-        create_op.result(90)  # raises on failure / timeout.
+        create_op.result(120)  # raises on failure / timeout.
 
         operation = temp_db.update_ddl(DDL_STATEMENTS)
 
         # We want to make sure the operation completes.
-        operation.result(90)  # raises on failure / timeout.
+        operation.result(120)  # raises on failure / timeout.
 
         temp_db.reload()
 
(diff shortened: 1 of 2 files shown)


spanner/tests/system/test_system.py
```diff
@@ -728,6 +728,20 @@ def _unit_of_work(transaction, test):
...
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        session, committed = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        expected = [[row[1], row[2]] for row in self._row_data(row_count)]
+        row = 5
+        rows = list(session.read(self.TABLE,
+                            columns,
+                            KeySet(keys=[[expected[row][0], expected[row][1]]]),
+                            index='name')
+                    )
```
dhermes: There is a lot (too much?) going on here, making it hard to read. Instead I recommend the following:

```python
XYZ = KeySet(keys=[[expected[row][0], expected[row][1]]])
read_iter = session.read(self.TABLE, columns, XYZ, index='name')
rows = list(read_iter)
```

You should replace `XYZ` with something that describes what the `KeySet` is doing

dhermes:
I'll let @tseaver weigh in, but it seems this is lacking an explanation of

- Why the timeout had to go up from 90 to 120
- Why `CREATE INDEX name ON contacts(first_name, last_name);` was added to the `DDL`
- Why `ROW_COUNT` was lowered from `4000` to `3000`

chemelnucfin:
Please see #4386 for the CREATE INDEX reasoning, and ROW_COUNT needs to be
lowered because apparently 20000 mutations is the cap,
and adding an additional INDEX apparently gets to the cap.

Short story is that instead of building a temp and removing it, it was
added to the fixture for easier manipulation.

The timeout at 90 would fail occasionally for me running the tests.  120 so
far has never failed for me.

PR #4405, opened 2017-11-17
BigQuery: add reference property to Table and Dataset

tswast:
This makes it easier to call `get_table`, `get_dataset`, or other
functions that want just a reference, not a full resource.

Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4327

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -34,6 +34,19 @@ _TABLE_HAS_NO_SCHEMA = "Table has no schema:  call 'client.get_table()'"
 _MARKER = object()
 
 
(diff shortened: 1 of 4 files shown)


robfraz:
Thanks!  LGTM (for what it's worth).

bigquery/google/cloud/bigquery/table.py
```diff
@@ -204,6 +204,20 @@ def table_id(self):
         """
         return self._table_id
 
+    @property
+    def reference(self):
```
tswast: I recently added this property to `TableListItem` as well. I'll update this PR to share this property with that implementation before submitting.

PR #4407, opened 2017-11-17
Spanner: Row check allow multiple rows to be unordered in check

chemelnucfin:
To help writing tests with multiple rows.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -219,20 +219,23 @@ class _TestData(object):
         else:
             self.assertEqual(value.microsecond * 1000, nano_value.nanosecond)
 
-    def _check_row_data(self, row_data, expected=None):
+    def _check_rows_data(self, rows_data, expected=None):
         if expected is None:
             expected = self.ROW_DATA
 
+        self.assertEqual(len(rows_data), len(expected))
+        for row, expected in zip(rows_data, expected):
+            self._check_row_data(row, expected)
+
+    def _check_row_data(self, row_data, expected):
         self.assertEqual(len(row_data), len(expected))
-        for found, expected in zip(row_data, expected):
-            self.assertEqual(len(found), len(expected))
-            for found_cell, expected_cell in zip(found, expected):
-                if isinstance(found_cell, TimestampWithNanoseconds):
-                    self._assert_timestamp(expected_cell, found_cell)
-                elif isinstance(found_cell, float) and math.isnan(found_cell):
-                    self.assertTrue(math.isnan(expected_cell))
-                else:
-                    self.assertEqual(found_cell, expected_cell)
+        for found_cell, expected_cell in zip(row_data, expected):
+            if isinstance(found_cell, TimestampWithNanoseconds):
+                self._assert_timestamp(expected_cell, found_cell)
+            elif isinstance(found_cell, float) and math.isnan(found_cell):
+                self.assertTrue(math.isnan(expected_cell))
+            else:
+                self.assertEqual(found_cell, expected_cell)
 
 
 class TestDatabaseAPI(unittest.TestCase, _TestData):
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -220,6 +220,12 @@ def _assert_timestamp(self, value, nano_value):
             self.assertEqual(value.microsecond * 1000, nano_value.nanosecond)
 
     def _check_row_data(self, row_data, expected=None):
+        try:
+            row_data = sorted(row_data)
+            expected = sorted(expected)
+        except TypeError:
```
dhermes: @chemelnucfin When will this fail with a `TypeError`?

I'd be more comfortable having the caller be explicit. So you could have a flag `sort=True` and then the callers which are un-sortable would have to change from the default to `sort=False`

chemelnucfin:
I broke up the function into table and row, so this works better now.

tseaver:
Sorting the rows is actually incorrect:  we expect that the values come back in a particular order in some testcases.

chemelnucfin:
@tseaver PTAL

spanner/tests/system/test_system.py
```diff
@@ -219,20 +219,22 @@ def _assert_timestamp(self, value, nano_value):
         else:
             self.assertEqual(value.microsecond * 1000, nano_value.nanosecond)
 
-    def _check_row_data(self, row_data, expected=None):
+    def _check_table_data(self, table_data, expected=None):
```
tseaver: This name is wrong: we aren't always checking a whole table. if you must split out the checks for a single row (why?), then call this `_check_rows_data`.

spanner/tests/system/test_system.py
```diff
@@ -219,20 +219,22 @@ def _assert_timestamp(self, value, nano_value):
...
-                    self._assert_timestamp(expected_cell, found_cell)
-                elif isinstance(found_cell, float) and math.isnan(found_cell):
-                    self.assertTrue(math.isnan(expected_cell))
-                else:
-                    self.assertEqual(found_cell, expected_cell)
+        self.assertEqual(len(table_data), len(expected))
+        for row, expected in zip(table_data, expected):
+            self._check_row_data(row, expected)
+
+    def _check_row_data(self, row_data, expected):
```
tseaver: We need another length check here for the individual rows:  Python's `zip` builtin truncates at the shortest component.

spanner/tests/system/test_system.py
```diff
@@ -492,7 +494,7 @@ def test_batch_insert_then_read_all_datatypes(self):
         snapshot = session.snapshot(read_timestamp=batch.committed)
         rows = list(snapshot.read(
             self.ALL_TYPES_TABLE, self.ALL_TYPES_COLUMNS, self.ALL))
-        self._check_row_data(rows, expected=self.ALL_TYPES_ROWDATA)
+        self._check_table_data(rows, self.ALL_TYPES_ROWDATA)
```
tseaver: Leave the keyword argument in place here.

spanner/tests/system/test_system.py
```diff
@@ -874,7 +875,7 @@ def test_read_w_index(self):
             self.TABLE, MY_COLUMNS, self.ALL, index='contacts_by_last_name'))
 
         expected = list(reversed(
-            [(row[0], row[2]) for row in self._row_data(ROW_COUNT)]))
+            [[row[0], row[2]] for row in self._row_data(ROW_COUNT)]))
```
tseaver: FWIW, there is no real difference here, given how the check function works.

spanner/tests/system/test_system.py
```diff
@@ -874,7 +875,7 @@ def test_read_w_index(self):
...
         self._check_row_data(rows, expected)
```
tseaver: This needs to be `_check_rows_data`.

PR #4410, opened 2017-11-17
BigQuery: docs for Row.keys, items, values, and get functions.

tswast:
I forgot to ask for these in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4393 before I merged.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -789,7 +789,8 @@ class Row(object):
         Returns:
             Sequence[object]: A sequence of length ``len(row)``.
         """
-        return copy.deepcopy(self._xxx_values)
+        for value in self._xxx_values:
+            yield copy.deepcopy(value)
 
     def keys(self):
         """Return the keys for using a row as a dict.
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import copy
 import datetime
 import operator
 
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -783,22 +783,38 @@ def __init__(self, values, field_to_index):
         self._xxx_field_to_index = field_to_index
 
     def values(self):
+        """Return the values included in this row.
+
```
dhermes: Worth noting that it's not a copy? Also won't it always be a `tuple`? If not, returning it is a "bad idea"

bigquery/google/cloud/bigquery/table.py
```diff
@@ -783,22 +783,38 @@ def __init__(self, values, field_to_index):
...
+        Returns:
+            Sequence[str]: The keys corresponding to the columns of a row
+
+        Examples:
+
+            >>> Row(('a', 'b'), {'x': 0, 'y': 1}).keys()
+            ['x', 'y']
         """
         keys = self._xxx_field_to_index.keys()
         return keys
```
dhermes: Where you're refactoring, just drop the `keys =` and return the output? Also, do you want `six.iterkeys`?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -808,18 +824,38 @@ def items(self):
         return items
```
dhermes: Another possible re-factoring:

```python
for key, index in six.iteritems(self._xxx_field_to_index):
    yield (key, self._xxx_values[index])
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -808,18 +824,38 @@ def items(self):
...
-        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('x')
-        'a'
-        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z')
-        None
-        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z', '')
-        ''
-        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('z', default = '')
-        ''
+        """Return value for ``key`` if ``key`` is in the row, else
+        ``default``.
```
dhermes: Please follow [PEP257][1] and keep the one-line summary to one line.

[1]: https://www.python.org/dev/peps/pep-0257/

bigquery/google/cloud/bigquery/table.py
```diff
@@ -808,18 +824,38 @@ def items(self):
...
+
+        Args:
+            key (str): The key of the column to access
+
```
dhermes: Phantom empty line?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -808,18 +824,38 @@ def items(self):
...
+            object:
+                The value associated with the provided key, or a default value.
+
+        Examples:
+            When the key exists, the value associated with it is returned.
+
+            >>> Row(('a', 'b'), {'x': 0, 'y': 1}).get('x')
+            'a'
+
+            The default value is ``None`` when the key does not exist.
```
dhermes: For your future reference

```
:data:`None`
```

will actually link to the std. library, rather than just code font. But I doubt people need that link.

bigquery/google/cloud/bigquery/table.py
tswast: It's _supposed_ to always be a tuple. Agreed that it's a little scary, though.

The values themselves can sometimes be mutable, such as with RECORD/STRUCT types. I'll do a deep copy.

bigquery/google/cloud/bigquery/table.py
tswast: Good to know, thanks

bigquery/google/cloud/bigquery/table.py
tswast: Done.

Ideally we'd somehow make this a "view" like in Python 3 dictionaries, but that seems like a lot of effort, especially since rows should be immutable.

bigquery/google/cloud/bigquery/table.py
tswast: Done.

dhermes:
@tswast For future PRs, can you squash while merging the PR? It's harder to review a PR when I can't tell what you changed when squashing. (Though I definitely feel the seem way, I squash locally a lot because for some reason I care about commit log clutter, even though we are 5500+ commits in to this project.)

bigquery/google/cloud/bigquery/table.py
```diff
@@ -783,43 +784,72 @@ def __init__(self, values, field_to_index):
...
+        """Return the values included in this row.
 
-    def keys(self):
+        Returns:
+            Sequence[object]: A sequence of length ``len(row)``.
         """
-        Return keys as of a dict:
-        >>> Row(('a', 'b'), {'x': 0, 'y': 1}).keys()
-        ['x', 'y']
+        return copy.deepcopy(self._xxx_values)
```
dhermes: @tswast Sorry, I wasn't trying to suggest you actually do a copy. The constructor "mandates" a `tuple` and it certainly would be more performant to just return the immutable `values`

bigquery/google/cloud/bigquery/table.py
```diff
@@ -783,43 +784,72 @@ def __init__(self, values, field_to_index):
...
-        [('x', 'a'), ('y', 'b')]
-        """
-        items = [
-            (k, self._xxx_values[i])
-            for k, i
-            in self._xxx_field_to_index.items()
-        ]
-        return items
+        for key, index in six.iteritems(self._xxx_field_to_index):
+            yield (key, copy.deepcopy(self._xxx_values[index]))
```
dhermes: Here the copy might make more sense, because even if `self._xxx_values` is a `tuple` (i.e. immutable), it's values might not be.

bigquery/google/cloud/bigquery/table.py
tswast: I've just updated `values()` to return an iterator/generator to be more consistent with [dict.values()](https://docs.python.org/3.6/library/stdtypes.html#dict.values). I'll keep the copy for the reason that the items might be mutable objects.

tswast:
Yeah, it is a bit sad about copies. There's definitely a trade-off between safety and performance. Probably better to err on the side of safety.

dhermes:
@tswast Looks like the generators broke the build: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4288

Shall we revert, then re-submit this PR but push the branch to the GoogleCloudPlatform remote so system tests run?

tswast:
Yeah, let's revert. I didn't think to check the system tests.

PR #4411, opened 2017-11-17
Spanner: Range read with keys and index

chemelnucfin:
For some reason, a similar code wasn't working for me the other day and failed with:
_Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, Endpoint read failed)>
But it now works.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -959,6 +959,122 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         expected = all_data_rows[START+1:END+1]
         self._check_row_data(rows, expected)
 
+    def test_read_with_range_keys_index_closed_closed(self):
+        row_count = 10
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        data = [[row[1], row[2]] for row in self._row_data(row_count)]
+        session, _ = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        start, end = 3, 7
+        krange = KeyRange(start_closed=data[start], end_closed=data[end])
+        keyset = KeySet(ranges=(krange,))
+        rows = list(session.read(self.TABLE, columns, keyset, index='name'))
+        self.assertEqual(rows, data[start : end+1])
+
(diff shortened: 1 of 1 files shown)


spanner/tests/_fixtures.py
```diff
@@ -47,6 +47,7 @@
     name STRING(16),
     tags ARRAY<STRING(16)> )
     PRIMARY KEY (id);
+CREATE INDEX name ON contacts(first_name, last_name);
```
tseaver: This change (and the ones below which accompany it) should be split out into a separate PR:  there is no point in cluttering the rest of them with the noise.

spanner/tests/system/test_system.py
```diff
@@ -936,8 +936,86 @@ def test_read_w_ranges(self):
...
+        for index in range(len(range_type)):
+            keyset = KeySet(ranges=range_type[index])
+            rows = list(session.read(self.TABLE,
+                                     columns,
+                                     keyset,
+                                     index='name',
+                                     limit=limit)
+            )
+            expected = data[indices[index][0]:indices[index][1]]
+            self.assertEqual(rows, expected[:limit])
```
tseaver: Rather than looping, please split these two test cases up into eight (four each, one per type of range).

tseaver: A simpler testcase is a) easier to review, and more importanly b) simpler to diagnose when it fails due to future changes.  Here is what one of them would look like:

```python
    def test_read_with_range_keys_index_closed_closed(self):
        row_count = 10
        columns = self.COLUMNS[1], self.COLUMNS[2]
        data = [[row[1], row[2]] for row in self._row_data(row_count)]
        session, _ = self._set_up_table(row_count)
        self.to_delete.append(session)
        start, end = 3, 7
        range = KeyRange(start_closed=data[start], end_closed=data[end])
        keyset = KeySet(ranges=(range,))
        rows = list(session.read(self.TABLE, columns, keyset, index='name'))
        self.assertEqual(rows, data[start:end + 1])
```

chemelnucfin:
@tseaver PTAL thanks.

PR #4416, opened 2017-11-18
Spanner: Range single keys index

chemelnucfin:
I'm separating the tests into different PRs to minimize the damage.  I plan to do the refactor once all the tests are all in and I'll do it separately so I don't change too much code at once.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -1104,6 +1104,83 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         expected = data[start+1 : end]
         self.assertEqual(rows, expected[:limit])
 
+    def test_read_with_range_keys_and_index_closed_closed(self):
+        row_count = 10
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+
+        session, committed = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        data = [[row[1], row[2]] for row in self._row_data(row_count)]
+        keyrow, start, end = 1, 3, 7
+        closed_closed = KeyRange(start_closed=data[start],
+                                 end_closed=data[end])
+        keys = [data[keyrow],]
+        keyset = KeySet(keys=keys, ranges=(closed_closed,))
+        rows = list(session.read(self.TABLE,
+                                 columns,
+                                 keyset,
+                                 index='name')
+        )
+        expected = ([data[keyrow]] + data[start : end+1])
+        self.assertEqual(rows, expected)
+
+    def test_read_with_range_keys_and_index_closed_open(self):
+        row_count = 10
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        session, committed = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        data = [[row[1], row[2]] for row in self._row_data(row_count)]
+        keyrow, start, end = 1, 3, 7
+        closed_open = KeyRange(start_closed=data[start],
+                               end_open=data[end])
+        keys = [data[keyrow],]
+        keyset = KeySet(keys=keys, ranges=(closed_open,))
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -936,8 +936,49 @@ def test_read_w_ranges(self):
...
+        for index in range(len(range_type)):
+            keyset = KeySet(keys=keys, ranges=range_type[index])
+            rows = list(session.read(self.TABLE,
+                                     columns,
+                                     keyset,
+                                     index='name')
+            )
+            selected = ([expected[keyrow]] +
+                        expected[indices[index][0]:indices[index][1]])
+            self.assertEqual(rows, selected)
```
tseaver: Rather than looping, please split this into four testcases.

chemelnucfin:
@tseaver PTAL, thanks.

spanner/tests/system/test_system.py
```diff
@@ -973,6 +973,90 @@ def test_read_w_ranges(self):
         expected = all_data_rows[START+1:END+1]
         self._check_row_data(rows, expected)
 
+    def test_read_with_range_keys_index_closed_closed(self):
+        row_count = 10
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        session, committed = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        expected = [[row[1], row[2]] for row in self._row_data(row_count)]
```
tseaver: `expected` is not a good name for this variable, as it no longer matches what we expect to be returned from the `read` call.  `data` would maybe be better?  Applies to all four testcases.

spanner/tests/system/test_system.py
```diff
@@ -973,6 +973,90 @@ def test_read_w_ranges(self):
...
+        keyrow, start, end = 1, 3, 7
+        start_index = [expected[start][0], expected[start][1]]
+        end_index = [expected[end][0], expected[end][1]]
```
tseaver: `start_index` and `end_index` are unused (in all these testcases).

spanner/tests/system/test_system.py
```diff
@@ -973,6 +973,90 @@ def test_read_w_ranges(self):
...
+        closed_closed = KeyRange(start_closed=expected[start],
+                                 end_closed=expected[end])
+        keys = [expected[keyrow], ]
+        keyset = KeySet(keys=keys, ranges=(closed_closed, ))
+        rows = list(session.read(self.TABLE,
+                                 columns,
+                                 keyset,
+                                 index='name')
+        )
+        selected = ([expected[keyrow]] + expected[start : end+1])
```
tseaver: *This* is the one which should be called `expected`.

spanner/tests/system/test_system.py
```diff
@@ -988,6 +988,82 @@ def test_read_w_ranges(self):
...
+    def test_read_with_range_keys_index_closed_closed(self):
+        row_count = 10
+        columns = self.COLUMNS[1], self.COLUMNS[2]
+        session, committed = self._set_up_table(row_count)
+        self.to_delete.append(session)
+        data = [[row[1], row[2]] for row in self._row_data(row_count)]
+        keyrow, start, end = 1, 3, 7
+        closed_closed = KeyRange(start_closed=expected[start],
+                                 end_closed=expected[end])
+        keys = [expected[keyrow],]
```
tseaver: `expected` is not yet defined, which means that these three lines must be picking up a global, or else that this testcase name is shadowed by another (neither is good -- we should figure out which is happening).

chemelnucfin: ahh, I forgot that system tests don't get triggered.  I was using these as CI tools sometimes.

chemelnucfin:
I added "and" in the title to separate from another set of tests.  Merging on green.

PR #4417, opened 2017-11-18
Spanner: Prevents Conflict from having the same database name

chemelnucfin:
Somehow this conflict occurs occasionally.  I might have created another temp_db database while working on the tests.  

This PR guards against it.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -845,12 +845,11 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         temp_db = Config.INSTANCE.database(
             'test_read_w_index', ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
             pool=pool)
-        operation = temp_db.create()
+        if not temp_db.exists():
+            operation = temp_db.create()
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -845,12 +845,11 @@ def test_read_w_index(self):
         temp_db = Config.INSTANCE.database(
             'test_read_w_index', ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
             pool=pool)
-        operation = temp_db.create()
+        if not temp_db.exists():
```
dhermes: @chemelnucfin This won't help contention, because if it exists, then (at least) two tests would be trying to modify the same database.

You'd be better of just failing the test if `temp_db.exists()`. (Even this is not quite sufficient due to race conditions.)

chemelnucfin:
You are right, I wasn't thinking about race conditions, but only locally
where I was reusing the same name.  Should I be thinking about race
conditions or just close this issue?

Thanks.

chemelnucfin:
Closing in favor of #4435

PR #4422, opened 2017-11-20
Spanner: Query NaN

chemelnucfin:


spanner/tests/system/test_system.py
```diff
@@ -1165,6 +1165,15 @@ def test_execute_sql_w_query_param_transfinite(self):
             expected=[(107,)],
         )
 
+        # Find NaN
+        self._check_sql_results(
+            snapshot,
+            sql='SELECT eye_d FROM all_types WHERE IS_NAN(approx_value)',
```
tseaver: This query does not use the parameter, which means it does not belong in this testcase.

chemelnucfin:
Not an integration 7 test.

PR #4423, opened 2017-11-20
Spanner: Select one and invalid sql queries

chemelnucfin:


tseaver:
Neither of these testcases is supposed to be added to our system tests: "select 1" is marked "remove,", and "invalid query" is marked "unit."

Issue #4373, opened 2017-11-09
BigQuery Python Client v0.28.0: num_rows, schema property of a Table returning None and empty list


pramodvspk:
Hi,

I am using the BigQuery Python client v0.28.0

Using the bigquery client to list all tables in a dataset using method `list_dataset_tables`, I get an iterator of table references in the dataset. When I try to access the `num_rows` and `schema` properties of the Table object it returns `None` and `[]`.

But when I access the table using `client.get_table()` I am able to access both the num_rows and schema properties.

I am confused since both the methods return a `google.cloud.bigquery.table.Table` object but are inconsistent.

```
for table in bqclient.list_dataset_tables(dataset_ref):
	print(table.num_rows)
	print(table.schema)
```
Returns None and []
```
table = bqclient.get_table(table)
	print(table.num_rows)
	print(table.schema)
```
Returns the number of rows and list of schemaFields

Thanks

tswast:
Thanks for the report.

It appears as though the [BigQuery tables.list](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/list) method only returns a subset of the fields of a table.

I see a few potentially sane fixes:
1. (most desirable) See if we can get the backend team to expose a full Table resource in list operations (maybe with an optional parameter).
2. (breaking, but maybe our best option) Make `list_dataset_tables` return a list of table references and discard the few extra table fields that the list call returns.
3. (inefficient) Make a call to get the table resource for each item in the returned list.

I'll send a note to my contacts on the BigQuery backend team about doing (1), but the most likely fix will be (2).

tswast:
I guess there's another option (4) document that only a subset of fields are exposed in the table response objects from the list request.

I dislike that option the most, as it leaves a lot of room for confusion.

tswast:
I got confirmation from the backend team that (1) is infeasible. Properties like the schema and total rows of the table take much longer to fetch in the backend. The list API call includes only properties which are fast for the backend to fetch.

I think I'd like to propose another option (5) Introduce a new type containing this subset of properties. Maybe `PartialTable` or `TableListItem` which includes only those properties present in the list API response. That way it will be much clearer from the documentation which properties are included.

tseaver:
I had (naively, I guess) assumed that this issue was the exact reason `TableReference` even exists.

tswast:
[...]

PR #4427, opened 2017-11-20
BigQuery: Use TableListItem for table listing.

tswast:
The table list response only includes a subset of all table properties.
This commit adds a new type to document explicitly which properties are
included, but also make it clear that this object should not be used in
place of a full Table object.

Towards #4373 (needs a similar change for dataset listing before closing that issue)

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -31,27 +31,28 @@ from google.resumable_media.requests import ResumableUpload
 from google.api_core import page_iterator
 from google.api_core.exceptions import GoogleAPICallError
 from google.api_core.exceptions import NotFound
-
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
+
+from google.cloud.bigquery._helpers import DEFAULT_RETRY
+from google.cloud.bigquery._helpers import _SCALAR_VALUE_TO_JSON_ROW
+from google.cloud.bigquery._helpers import _field_to_index_mapping
+from google.cloud.bigquery._helpers import _item_to_row
+from google.cloud.bigquery._helpers import _rows_page_start
+from google.cloud.bigquery._helpers import _snake_to_camel_case
 from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetReference
-from google.cloud.bigquery.table import Table, _TABLE_HAS_NO_SCHEMA
-from google.cloud.bigquery.table import TableListItem
-from google.cloud.bigquery.table import TableReference
-from google.cloud.bigquery.table import _row_from_mapping
 from google.cloud.bigquery.job import CopyJob
 from google.cloud.bigquery.job import ExtractJob
 from google.cloud.bigquery.job import LoadJob
 from google.cloud.bigquery.job import QueryJob, QueryJobConfig
 from google.cloud.bigquery.query import QueryResults
-from google.cloud.bigquery._helpers import _item_to_row
-from google.cloud.bigquery._helpers import _rows_page_start
-from google.cloud.bigquery._helpers import _field_to_index_mapping
-from google.cloud.bigquery._helpers import _SCALAR_VALUE_TO_JSON_ROW
-from google.cloud.bigquery._helpers import DEFAULT_RETRY
-from google.cloud.bigquery._helpers import _snake_to_camel_case
+from google.cloud.bigquery.table import Table
+from google.cloud.bigquery.table import TableListItem
+from google.cloud.bigquery.table import TableReference
+from google.cloud.bigquery.table import _TABLE_HAS_NO_SCHEMA
+from google.cloud.bigquery.table import _row_from_mapping
 
 
 _DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -38,6 +38,7 @@ from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetReference
 from google.cloud.bigquery.table import Table, _TABLE_HAS_NO_SCHEMA
+from google.cloud.bigquery.table import TableListItem
 from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.table import _row_from_mapping
 from google.cloud.bigquery.job import CopyJob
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -38,6 +38,7 @@
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetReference
 from google.cloud.bigquery.table import Table, _TABLE_HAS_NO_SCHEMA
+from google.cloud.bigquery.table import TableListItem
```
dhermes: While you're here, would you mind adding a "fix" (per the Google Python style guide) for the import above:

```python
from google.cloud.bigquery.table import Table
from google.cloud.bigquery.table import _TABLE_HAS_NO_SCHEMA
```

(Probably lexical order too?)

tswast:
> 1. How many of these @property-s are copy-pasta?

Basically all of them. I made a slight change to `view_use_legacy_sql` to use the `type` property to check if a table is a view. This is because the `query` property of the view is not included in the table list response, so the type property is the only way to check if the table is a view when `useLegacySql` is left at the default value.

> 2. I don't think the class name really conveys what it is (a read-only version of Table).

It being read-only is more of a side-effect than the primary purpose of the class.

> 3. Is there a reason you might be opposed to just subclassing Table or just an arbitrary decision to make a new class?

The primary purpose of `TableListItem` is to indicate that there is a different set of properties available on the list response than a full `Table` object. Notable properties that aren't there include schema and total rows (#4373).

A sub-class would allow this object to be used in any place a table is allowed, such as update or create requests, which is not something I want people to be doing (because not all properties are present).

> 4. Are you opposed to adding a read_only attribute / member to Table?

I hadn't really considered it, because it wouldn't solve the problem of `schema` or `num_rows` not being populated from a list operation.

dhermes:
@tswast I'm just trying to avoid code duplication, because the copy-pasta will drift over time.

If you did a subclass, you could over-ride and raise an `AttributeError` for properties not contained in the response (or you could do it in the `Table` class by using `self._read_only` to check).

The same applies to methods like `Table.update()` and `Table.create()`.

I think having a **different** class is probably the right move as far as "principle of least astonishment" is concerned. (Or rather, if you returned a `Table` that was limited, users might be surprised/astonished if they never checked `table.read_only`.) But I'd **strongly** suggest making the other class have much less code defined in it / leverage the existing stuff.

tswast:
There's so little code in this properties, I'm not super concerned about drift.

bigquery/google/cloud/bigquery/client.py

tswast: Looks like we'll have a bit of work cut out for us to get `flake8-import-order` happy once it works well for us.

dhermes:
> There's so little code in these properties, I'm not super concerned about drift.

But I **am** concerned. Care to sell me a bit on your position?

tswast:
Most of these properties have implementations like:

```
return self._properties.get('tableReference', {}).get('datasetId')
```

These properties directly reflect properties on the resource, with minimal modification.

The `reference` does have code which could drift.

    from google.cloud.bigquery import dataset

    dataset_ref = dataset.DatasetReference(self.project, self.dataset_id)
    return TableReference(dataset_ref, self.table_id)

I agree this could be shared with the `Table` class, but the `Table` class doesn't have this property yet. I'm adding it in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4405. I could update that PR to share the code with the `TableListItem` class once this one goes in.

You're right that `view_use_legacy_sql` is code that could drift, too. I think there are probably better ways to share that code than a subclass. I'll make an update to the PR.

dhermes:
@tswast Though these are simple "schema->data->Python" mappings, some APIs (BQ included) have changed the underlying schema and caught us off guard. With that in mind, "so little code" doesn't inspire much confidence?

Also, I don't want to get lost in this discussion that I think `TableListItem` is a not great name. Other candidates? `ReadOnlyTable`?

tswast:
Discussed in chat. Since `Table` and "whatever we call the resource from table list" are actually decoupled from each other in the backend, it doesn't make sense to couple them here.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -713,6 +713,137 @@ def _build_resource(self, filter_fields):
         return resource
 
 
+class TableListItem(object):
+    """Read-only table resource object with a subset of table properties.
```
dhermes: @tswast Can you add another sentence or two before "See..." explaining why this type exists and where the data comes from?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -531,23 +550,7 @@ def view_query(self):
...
-        :returns: The boolean for view.useLegacySql, or None if not a view.
-        """
-        view = self._properties.get('view')
-        if view is not None:
-            # useLegacySql is never missing from the view dict if this table
-            # was created client-side, because the view_query setter populates
-            # it. So a missing or None can only come from the server, whose
-            # default is True.
-            return view.get('useLegacySql', True)
+    view_use_legacy_sql = property(_view_use_legacy_sql_getter)
```
dhermes: You can keep the docstring, just dedent it 4 spaces?

bigquery/google/cloud/bigquery/table.py
tswast: Added.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -717,13 +717,21 @@ def _build_resource(self, filter_fields):
 
 
 class TableListItem(object):
-    """Read-only table resource object with a subset of table properties.
+    """A read-only table resource from a list operation.
+
+    For performance reasons, the BigQuery API only includes some of the table
+    properties when listing tables. Notably,
+    :attr:`google.cloud.bigquery.table.Table.schema` and
+    :attr:`google.cloud.bigquery.table.Table.num_rows` are missing.
```
dhermes: Probably want to use a `~` before these super long attribute names?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -717,13 +717,21 @@ def _build_resource(self, filter_fields):
...
+
+    For a full list of the properties that the BigQuery API returns, see the
+    `REST documentation for tables.list
+    <https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/list>`.
```
dhermes: How does this multiline anchor render?

bigquery/google/cloud/bigquery/table.py
tswast: The docs render fine when I test it out. It grabs them off the `_view_use_legacy_sql_getter`.

tswast: image

dhermes: Ah OK. I wonder why that is? (I hope it's not a strange Sphinx cache "feature" making it seem like that.)

bigquery/google/cloud/bigquery/table.py
tswast: Looks fine to me.

image

dhermes:
@tswast OK Do you feel this is merge-able?

tswast:
> Do you feel this is merge-able?

Yeah. Just ran the full nox suite locally to double-check.

dhermes:
Cool LGTM

PR #4435, opened 2017-11-22
Spanner: Unique resource for tempdbs

chemelnucfin:


spanner/tests/system/test_system.py
```diff
@@ -281,7 +281,7 @@ def test_create_database(self):
 
     def test_update_database_ddl(self):
         pool = BurstyPool()
-        temp_db_id = 'temp_db'
+        temp_db_id = 'temp_db' + unique_resource_id('_')
```
tseaver: This part is covered by #4489.

spanner/tests/system/test_system.py
```diff
@@ -843,7 +843,8 @@ def test_read_w_index(self):
         ]
         pool = BurstyPool()
         temp_db = Config.INSTANCE.database(
-            'test_read_w_index', ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
+            'temp_db' + unique_resource_id('_'),
+            ddl_statements=DDL_STATEMENTS + EXTRA_DDL,
```
tseaver: It would be better to redo this part after landing the separate PR which adds the index to the main fixture.

Issue #4415, opened 2017-11-17
Trace: Implement V2 API client


liyanhui1228:
As the [Stackdriver Trace V2 API](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2) is launched, we need to implement client library to support using it.

Checklist for this issue:

- [x] Merge the Autogen Trace V2 library (PR #4379)
- [x] Implement Stackdriver Trace V2 API client

After the Trace V2 API client library is available, calling `from google.cloud import trace` will by default using the V2 API. To use V1 API, one will need to explicitly import like this: `from google.cloud.trace.v1 import Client`.

PR #4437, opened 2017-11-22
Add Trace V2 client and _gax

liyanhui1228:
For #4415


liyanhui1228:
Friendly ping.

dhermes:
@liyanhui1228 Who is the ping intended for?

liyanhui1228:
Would like @lukesneeringer for initial review to make sure we agree on making Trace V2 API as the default.

trace/google/cloud/trace/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Google LLC
+# Copyright 2017, Google LLC All rights reserved.
```
duggelz: From my reading of the OSPO guidelines, this was correct as it was, e.g.

`# Copyright 2017 Google LLC`

Unless you found another source that says something differently (very possible).

trace/google/cloud/trace/v1/_gax.py
```diff
@@ -0,0 +1,206 @@
...
+    :param client: The client that owns this API object.
+    """
+    def __init__(self, gax_api, client):
+        self._gax_api = gax_api
+        self.client = client
+
+    def patch_traces(self, project_id, traces, options=None):
+        """Sends new traces to Stackdriver Trace or updates existing traces.
+
+        :type project_id: str
```
duggelz: Just to check, is there any reason we need to leave these with the old-style docstrings?  I.e. we have napoleon above like:

```
Args:
    start_time (:class:`~datetime.datetime`):
               The start time of the span. 
```

vs the `:type`, `:param` style here.

If there isn't a reason to keep them different, then we should file an issue against ourselves to go clean everything up to a single consistent style (some day).

dhermes: @duggelz We've been "fine with both" for some time now. However we've privately discussed being OK with badgering Googlers to exclusively use Google-style docstrings.

liyanhui1228: Changed to use napoleon style.

liyanhui1228:
`from google.cloud import trace` will be using V2 as the default, and using V1 will be like `from google.cloud import trace_v1`. The V1 API includes both read and write methods, but V2 API only contains write methods.

PR #4439, opened 2017-11-22
BigQuery: Use DatasetListItem for client.list_datasets

### LINKED ISSUE #4373: BigQuery Python Client v0.28.0: num_rows, schema property of a Table returning None and empty list

tswast:
Listing datasets only includes a subset of the properties available on a
dataset. The DatasetListItem class is used to explicitly document which
features are available and to prevent confusion from trying to use the
resulting object in other contexts, like updating.

Closes #4373.


tswast:
Ping. Please review.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -546,4 +545,91 @@ def table(self, table_id):
...
+        self._properties = resource
+
+    @property
+    def project(self):
+        """Project bound to the dataset.
+
+        :rtype: str
+        :returns: the project.
+        """
+        return self._properties.get('datasetReference').get('projectId')
```
tseaver: Presumes that the `datasetReference` key is always populated appropriately (with at least an empty dict).

We've not been perfectly consistent about defending against "perverse" back-end responses, so maybe this one can go by.  In that case, is `self._properties['datasetReference']['projectId']` clearer?

If we *do* leave the `.get()` bits in place, then we need to exercise the implicit branches for missing `datasetReference` in the unit tests.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -546,4 +545,91 @@ def table(self, table_id):
...
+
+    @property
+    def dataset_id(self):
+        """Dataset ID.
+
+        :rtype: str
+        :returns: the dataset ID.
+        """
+        return self._properties.get('datasetReference').get('datasetId')
```
tseaver: Ditto here.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -546,4 +545,91 @@ def table(self, table_id):
...
+        return self._properties.get('friendlyName')
+
+    @property
+    def labels(self):
+        """Labels for the dataset.
+
+        :rtype: dict, {str -> str}
+        :returns: A dict of the the dataset's labels.
+        """
+        return self._properties['labels']
```
tseaver: This one is *not* defensive against missing labels:  I would say that one is *more* likely to be missing than `datasetReference`, based on past experience.

If we mean to raise `KeyError` from the property, then we should test that as well.

tswast: Good point. Agreed that it's more likely to be missing.

bigquery/google/cloud/bigquery/dataset.py
tswast: I agree that `self._properties['datasetReference']['projectId']` is clearer in this case. I've added an explicit check for these properties in the constructor.

PR #4442, opened 2017-11-22
Spanner: Tests for empty array lists and array lists of various types such as BOOL

chemelnucfin:


tseaver:
No system test required ("empty read" tests are marked "unit").

PR #4444, opened 2017-11-23
PubSub: Making `thread.Policy.on_exception` more robust.

dhermes:
**NOTE**: I am **NOT** comfortable with this as a fix, it seems **WAAAAAY** too bolted on.

- ~~I'm surprised the retry strategy doesn't come into play here~~
- ~~There should a less hard-coded way to map to the idempotent retry codes~~
- ~~The method should have a straightforward way to determine the gRPC status code from the exception~~
- `RuntimeError: set_exception can only be called once` occurs if I don't filter out the `UNAVAILABLE` ([e.g.][1]). This seems like a bug in a different code path than is covered here.

----

- Adding special handling for API core exceptions
- Retrying on both types of idempotent error
- Also doing a "drive-by" hygiene fix changing a global from `logger` to `_LOGGER`

Towards #4234.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
@@ -66,7 +66,7 @@ class BasePolicy(object):
     """
 
     _managed_ack_ids = None
-    _IDEMPOTENT_RETRIES = (
+    _RETRYABLE_STREAM_ERRORS = (
         exceptions.DeadlineExceeded,
         exceptions.ServiceUnavailable,
     )
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
@@ -21,6 +21,7 @@ import logging
 import random
 import time
 
+from google.api_core import exceptions
 import six
 
 from google.cloud.pubsub_v1 import types
@@ -65,6 +66,10 @@ class BasePolicy(object):
     """
 
     _managed_ack_ids = None
+    _IDEMPOTENT_RETRIES = (
+        exceptions.DeadlineExceeded,
+        exceptions.ServiceUnavailable,
+    )
 
     def __init__(self, client, subscription,
                  flow_control=types.FlowControl(), histogram_data=None):
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -155,7 +155,7 @@ class Policy(base.BasePolicy):
         """
         # If this is in the list of idempotent exceptions, then we want to
         # retry. That entails just returning None.
-        if isinstance(exception, self._IDEMPOTENT_RETRIES):
+        if isinstance(exception, self._RETRYABLE_STREAM_ERRORS):
             return
 
         # Set any other exception on the future.
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -21,6 +21,7 @@ import threading
 import grpc
 from six.moves import queue as queue_mod
 
+from google.api_core import exceptions
 from google.cloud.pubsub_v1 import types
 from google.cloud.pubsub_v1.subscriber import _helper_threads
 from google.cloud.pubsub_v1.subscriber.futures import Future
@@ -28,12 +29,16 @@ from google.cloud.pubsub_v1.subscriber.policy import base
 from google.cloud.pubsub_v1.subscriber.message import Message
 
 
-logger = logging.getLogger(__name__)
+_LOGGER = logging.getLogger(__name__)
+_IDEMPOTENT_RETRY_CODES = (
+    grpc.StatusCode.DEADLINE_EXCEEDED,
+    grpc.StatusCode.UNAVAILABLE,
+)
 
 
 def _callback_completed(future):
     """Simple callback that just logs a `Future`'s result."""
-    logger.debug('Result: %s', future.result())
+    _LOGGER.debug('Result: %s', future.result())
 
 
 class Policy(base.BasePolicy):
@@ -80,7 +85,7 @@ class Policy(base.BasePolicy):
         )
 
         # Also maintain a request queue and an executor.
-        logger.debug('Creating callback requests thread (not starting).')
+        _LOGGER.debug('Creating callback requests thread (not starting).')
         if executor is None:
             executor = futures.ThreadPoolExecutor(max_workers=10)
         self._executor = executor
@@ -122,7 +127,7 @@ class Policy(base.BasePolicy):
         self._future = Future(policy=self)
 
         # Start the thread to pass the requests.
-        logger.debug('Starting callback requests worker.')
+        _LOGGER.debug('Starting callback requests worker.')
         self._callback = callback
         self._consumer.helper_threads.start(
             'callback requests worker',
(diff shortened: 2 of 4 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -153,10 +158,16 @@ def on_exception(self, exception):
...
         """
-        # If this is DEADLINE_EXCEEDED, then we want to retry.
-        # That entails just returning None.
-        deadline_exceeded = grpc.StatusCode.DEADLINE_EXCEEDED
-        if getattr(exception, 'code', lambda: None)() == deadline_exceeded:
+        if isinstance(exception, exceptions.GoogleAPICallError):
+            code = exception.grpc_status_code
+        else:
+            code = getattr(exception, 'code', None)
+            if callable(code):
```
jonparrott: ???

pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
```diff
@@ -89,9 +91,26 @@ def test_on_callback_request():
...
     policy = create_policy()
-    exc = mock.Mock(spec=('code',))
-    exc.code.return_value = grpc.StatusCode.DEADLINE_EXCEEDED
+
+    trailing = None
+    status_code = grpc.StatusCode.DEADLINE_EXCEEDED
+    details = 'Bad thing happened. Time out, go sit in the corner.'
+    exc_state = grpc._channel._RPCState(
+        (), None, trailing, status_code, details)
+    exc = grpc._channel._Rendezvous(exc_state, None, None, None)
```
jonparrott: Yesh, just subclass `grpc.Call` and `grpc.RpcError`.

jonparrott:
> I'm surprised the retry strategy doesn't come into play here

It's because this happens during streaming. `api_core` can't know the semantics for restarting the stream. In this case, it's simple, in others (such as spanner) it's not.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -153,10 +158,16 @@ def on_exception(self, exception):
 
         This will cause the stream to exit loudly.
         """
-        # If this is DEADLINE_EXCEEDED, then we want to retry.
-        # That entails just returning None.
-        deadline_exceeded = grpc.StatusCode.DEADLINE_EXCEEDED
-        if getattr(exception, 'code', lambda: None)() == deadline_exceeded:
+        if isinstance(exception, exceptions.GoogleAPICallError):
+            code = exception.grpc_status_code
+        else:
```
jonparrott: When is this ever not a `GoogleAPICallError`?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -153,10 +158,16 @@ def on_exception(self, exception):
...
+            code = getattr(exception, 'code', None)
+            if callable(code):
+                code = code()
+
+        # If this is in the list of idempotent exceptions DEADLINE_EXCEEDED,
+        # then we want to retry. That entails just returning None.
+        if code in _IDEMPOTENT_RETRY_CODES:
```
jonparrott: Couldn't you just do `if isinstance(exception, _IDEMPOTENT_STREAM_ERRORS):`

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: `_Rendevous.code` is callable, which is why he did that. In theory we should catch all of them, but if we miss one...

jonparrott: But it's is a `grpc.Call` then `code` will *always* be callable, what's the case where there's a `code` attribute?

dhermes: I'm just going to do an `isinstance` check for `_Rendezvous`.

dhermes:
So I have updated and have a better understanding thanks to chats with @jonparrott and @lukesneeringer.

1. We **only** expect exceptions from `api_core` here. This is even [true for iterators][1].
1. I added `BasePolicy._IDEMPOTENT_RETRIES` and just added a unit test to sanity check that it matches the config. (I think this is preferred to an import time penalty for every user ever.)
1. Using `isinstance(exception, _IDEMPOTENT_RETRIES)`, all of the "branch to check for Rendezvous? is `code` callable? huh?" issues are gone.

I'm still not sure why `set_exception` gets called twice (crossref #4380). I'm going to try to reproduce this by removing `exceptions.ServiceUnavailable` from `_IDEMPOTENT_RETRIES` and then running the code in my [gist][2].

I would like to continue that exploration but still merge this PR. SGTY @jonparrott and @lukesneeringer?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/68cedf952885847fa24f96487863b445ee821d9d/api_core/google/api_core/grpc_helpers.py#L58
[2]: https://gist.github.com/dhermes/488bf528e10456d3e2266e163ae1a72d

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -65,6 +66,10 @@ class BasePolicy(object):
     """
 
     _managed_ack_ids = None
+    _IDEMPOTENT_RETRIES = (
```
jonparrott: I hate to be *that guy*, but I'm not a fan of this name, how about `_RETRYABLE_STREAM_ERRORS`?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -153,10 +153,9 @@ def on_exception(self, exception):
 
         This will cause the stream to exit loudly.
         """
-        # If this is DEADLINE_EXCEEDED, then we want to retry.
-        # That entails just returning None.
-        deadline_exceeded = grpc.StatusCode.DEADLINE_EXCEEDED
-        if getattr(exception, 'code', lambda: None)() == deadline_exceeded:
+        # If this is in the list of idempotent exceptions, then we want to
+        # retry. That entails just returning None.
+        if isinstance(exception, self._IDEMPOTENT_RETRIES):
```
jonparrott: I hate to be *that girl*, but should you be accessing a private member of the superclass? (no need to do anything here if you think it's fine because it probably is fine)

jonparrott:
> I would like to continue that exploration but still merge this PR.

its-the-90s-go-for-it-girl-vhs-yospos-1425428235z

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
dhermes: I think it's fine. Internal to the "implementation", not internal to the particular implementation (i.e. class)

dhermes:
Well I was able to reproduce

```python
  File "${HOME}/.pyenv/versions/3.6.3/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "${HOME}/.pyenv/versions/3.6.3/lib/python3.6/threading.py", line 864, in run
  File ".../google/cloud/pubsub_v1/subscriber/_consumer.py", line 249, in _blocking_consume
  File ".../google/cloud/pubsub_v1/subscriber/policy/thread.py", line 162, in on_exception
    self._future.set_exception(exception)
  File ".../google/cloud/pubsub_v1/futures.py", line 159, in set_exception
    raise RuntimeError('set_exception can only be called once.')
RuntimeError: set_exception can only be called once.
```

on my very first run with the hack

```python
from google.cloud.pubsub_v1.subscriber.policy import base
# Monkey-patch to intentionally miss errors.
base.BasePolicy._RETRYABLE_STREAM_ERRORS = ()
```

PR #4445, opened 2017-11-23
PubSub: Don't throw exception to python interpreter.

shileiyu:
Avoiding annoying traceback.


dhermes:
@shileiyu Thanks for sending this PR, could you expound a little bit on what problem this solves? I.e. could you provide a code snippet that causes the "annoying traceback" you refer to and then show the equivalent code snippet **after** this change?

dhermes:
@shileiyu Thanks for taking the time to write and send this PR.

However, this issue is being resolve in a slightly different manner in #4472. Cheers!

shileiyu:
@googlebot I signed it!

PR #4446, opened 2017-11-23
Spanner: Read invalid column and read invalid table tests

chemelnucfin:


tseaver:
No system test required.

PR #4451, opened 2017-11-27
Spanner: Max staleness and min read timestamp multiuse tests

chemelnucfin:


tseaver:
No system test needed here:  we cover these cases in the unit tests.

PR #4453, opened 2017-11-27
BigQuery: Rename `properties` arg to `fields` in `update_table`

### LINKED ISSUE #4452: BigQuery: 'properties' parameter in google.cloud.bigquery.client.Client.update_table is not documented
'properties' parameter in google.cloud.bigquery.client.Client.update_table is not documented.
The body of the description is also inaccurate as it always does a PATCH request.
See: https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/reference.html#google.cloud.bigquery.client.Client.update_table

tswast:
This makes `update_table` consistent with `update_dataset`. Also,
documents the arguments to `update_table`.

Closes #4452.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -20,7 +20,6 @@ import collections
 import functools
 import os
 import uuid
-import warnings
 
 import six
 
@@ -314,7 +313,7 @@ class Client(ClientWithProject):
         If ``dataset.etag`` is not ``None``, the update will only
         succeed if the dataset on the server has the same ETag. Thus
         reading a dataset with ``get_dataset``, changing its fields,
-        and then passing it ``update_dataset`` will ensure that the changes
+        and then passing it to ``update_dataset`` will ensure that the changes
         will only be saved if no modifications to the dataset occurred
         since the read.
 
@@ -363,7 +362,7 @@ class Client(ClientWithProject):
 
         If ``table.etag`` is not ``None``, the update will only succeed if
         the table on the server has the same ETag. Thus reading a table with
-        ``get_table``, changing its fields, and then passing it
+        ``get_table``, changing its fields, and then passing it to
         ``update_table`` will ensure that the changes will only be saved if
         no modifications to the table occurred since the read.
 
@@ -377,8 +376,7 @@ class Client(ClientWithProject):
 
         Returns:
             google.cloud.bigquery.table.Table:
-                The :class:``~google.cloud.bigquery.table.Table`` returned
-                from the API call to update the table.
+                The table resource returned from the API call.
         """
         partial = table._build_resource(fields)
         if table.etag is not None:
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -354,8 +354,7 @@ class Client(ClientWithProject):
             retry, method='PATCH', path=path, data=partial, headers=headers)
         return Dataset.from_api_repr(api_response)
 
-    def update_table(
-            self, table, fields=None, properties=None, retry=DEFAULT_RETRY):
+    def update_table(self, table, fields, retry=DEFAULT_RETRY):
         """Change some fields of a table.
 
         Use ``fields`` to specify which fields to update. At least one field
@@ -373,8 +372,6 @@ class Client(ClientWithProject):
             fields (Sequence[str]):
                 The fields of ``table`` to change, spelled as the Table
                 properties (e.g. "friendly_name").
-            properties (Sequence[str]):
-                Deprecated alias for ``fields`` argument.
             retry (google.api_core.retry.Retry):
                 (Optional) A description of how to retry the API call.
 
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -353,23 +354,43 @@ def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY):
...
-
-        :rtype: :class:`google.cloud.bigquery.table.Table`
-        :returns: a ``Table`` instance
+        Use ``fields`` to specify which fields to update. At least one field
+        must be provided. If a field is listed in ``fields`` and is ``None``
+        in ``table``, it will be deleted.
+
+        If ``table.etag`` is not ``None``, the update will only succeed if
+        the table on the server has the same ETag. Thus reading a table with
+        ``get_table``, changing its fields, and then passing it
```
dhermes: Did you mean "passing it **to**"?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -353,23 +354,43 @@ def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY):
...
+                The fields of ``table`` to change, spelled as the Table
+                properties (e.g. "friendly_name").
+            properties (Sequence[str]):
+                Deprecated alias for ``fields`` argument.
+            retry (google.api_core.retry.Retry):
+                (Optional) A description of how to retry the API call.
+
+        Returns:
+            google.cloud.bigquery.table.Table:
+                The :class:``~google.cloud.bigquery.table.Table`` returned
```
dhermes: Only single backticks after `:class:`, also you probably don't need this because you've already stated the type

bigquery/google/cloud/bigquery/client.py
```diff
@@ -353,23 +354,43 @@ def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY):
...
+                from the API call to update the table.
         """
-        partial = table._build_resource(properties)
+        if properties is not None:
+            warnings.warn(
+                ('properties argument to update_table is deprecated. '
+                 'Use fields instead.'),
+                DeprecationWarning)
+            fields = properties
```
dhermes: This will over-write `fields` if both are not `None`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -353,23 +354,43 @@ def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY):
             retry, method='PATCH', path=path, data=partial, headers=headers)
         return Dataset.from_api_repr(api_response)
 
-    def update_table(self, table, properties, retry=DEFAULT_RETRY):
-        """API call:  update table properties via a PUT request
+    def update_table(
+            self, table, fields=None, properties=None, retry=DEFAULT_RETRY):
```
dhermes: This is already a breaking change because you're

1. Changing `properties` from a positional to a keyword argument
1. Changing `properties` in the "positional order" from 2nd to 3rd (the term "positional order" is not one that is used, I'm just using it as shorthand)

tswast:
I deprecated since `properties` is actually a [positional-or-keyword](https://docs.python.org/2/glossary.html#term-parameter), but I think you're right that almost no one will be using it as a keyword parameter.

I'll update the PR to replace rather than deprecate.

bigquery/google/cloud/bigquery/client.py
tswast: Acknowledged, moot since I just renamed the field in my latest commit.

It's a breaking change for a different reason than the ones you provide, though.

(1 & 2).  Actually, the `fields` argument can still be used as a positional parameter, identically how `properties` argument was before. It remains a a [positional-or-keyword parameter](https://docs.python.org/2/glossary.html#term-parameter) after adding a default value. If they used the `properties` argument positionally it works the same and if they used it as a keyword argument it works the same.

(3). The actual breaking change here is with the retry argument. Someone could have used the `retry` argument positionally, which now would get passed to `properties` instead. If we kept with the strategy of deprecating `properties`, I should have added it as a keyword argument to the end or have used `**kwargs` to fetch it.

bigquery/google/cloud/bigquery/client.py
tswast: Fixed.

bigquery/google/cloud/bigquery/client.py

bigquery/google/cloud/bigquery/client.py
dhermes: I realize you **can** use a positional as a keyword and pass a keyword as positional, but that opens the door for more states than one could possibly hope to wrangle.

The most relevant surface is what we **provide**, i.e. what is defined in our code and indicated in our docs. I have seen people who consider all arguments to be allowed as positional and this makes updating APIs very difficult.

@jonparrott and I have discussed before and determined that the order of keywords should not be a deciding factor in API changes.

rishibarve:
@tswast how can this be used to add a new column to the table is still not mentioned in the docs

tswast:
@rishibarve There is a sample of updating the schema property on https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/usage.html

Search for "Update multiple properties in a table".

You're right that an example of just adding a column to a table would be useful, though.

PR #4454, opened 2017-11-27
Making Pub/Sub unit tests use mock credentials.

dhermes:

PR #4464, opened 2017-11-27
Firestore: Import column lengths pass 79

chemelnucfin:
Saw this as I was adding in functionality.  Making separate PR to clean up.


dhermes:
@chemelnucfin The Pub / Sub failure was fixed in #4454, please rebase against the latest `master`

PR #4466, opened 2017-11-28
Distinguish FieldPath classes from field path strings

chemelnucfin:
~~Please wait until #4392 and #4464 is merged first.~~ 
Merged

### DIFF
--- a/firestore/tests/unit/test__helpers.py
+++ b/firestore/tests/unit/test__helpers.py
@@ -135,12 +135,17 @@ class TestFieldPath(unittest.TestCase):
     def test_to_api_repr_a(self):
         parts = 'a'
         field_path = self._make_one(parts)
-        self.assertEqual('a', field_path.to_api_repr())
+        self.assertEqual(field_path.to_api_repr(), 'a')
 
     def test_to_api_repr_backtick(self):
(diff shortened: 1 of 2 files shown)


firestore/tests/unit/test__helpers.py
```diff
@@ -456,13 +456,16 @@ def test_add_value_at_field_path_failure_adding_more_generic_path(self):
...
 
     def test_parse(self):
+        from google.cloud.firestore_v1beta1._helpers import FieldPath
         # "Cheat" and use OrderedDict-s so that iteritems() is deterministic.
         field_updates = collections.OrderedDict((
             ('a.b.c', 10),
             ('d', None),
             ('e.f1', [u'no', b'yes']),
             ('e.f2', 4.5),
+            (FieldPath.from_string('e.f3').to_api_repr(), (3, 1)),
```
schmidt-sebastian: Can you add a test case that verifies field path escaping? `FieldPath('h.i')` should do the trick.

chemelnucfin:
@schmidt-sebastian I fixed a bug where the regex did not match the entire group, and added some new tests to ensure the behavior, PTAL.  Thanks.

PR #4467, opened 2017-11-28
Storage: Relax project id regex due to issue

### LINKED ISSUE #4447: Storage: This doesn't match our existing project ID: hello-rightscale002
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/4b107613b0fcf139569504e9d44fa4a0793b4fd0/storage/google/cloud/storage/notification.py#L31

chemelnucfin:
Closes #4447

### DIFF
--- a/storage/google/cloud/storage/notification.py
+++ b/storage/google/cloud/storage/notification.py
@@ -28,7 +28,7 @@ JSON_API_V1_PAYLOAD_FORMAT = 'JSON_API_V1'
 NONE_PAYLOAD_FORMAT = 'NONE'
 
 _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
-_PROJECT_PATTERN = r'(?P<project>[a-z]+-[a-z]+-\d+)'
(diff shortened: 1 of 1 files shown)


dhermes:
@chemelnucfin Can you provide a link to documentation which specifies what an ID should look like? This doesn't seem generic enough.

chemelnucfin:
I haven't been able to find it yet.  This was just to solve the issue for
the user instead of having them wait.  Maybe I can open another more
generic issue while I look.

dhermes:
> just to solve the issue for the user instead of having them wait

It doesn't make sense to do that without understanding the fix. For that particular user, we can just recommend a monkey patch:

```python
import re

from google.cloud.storage import notification


notification._PROJECT_PATTERN = r'(?P<project>[a-z]+-[a-z]+\d+)'
notification._TOPIC_REF_PATTERN = notification._TOPIC_REF_FMT.format(
    notification._PROJECT_PATTERN,
    notification._TOPIC_NAME_PATTERN)
notification._TOPIC_REF_RE = re.compile(notification._TOPIC_REF_PATTERN)
```

chemelnucfin:
got it.  I'll see if I can find the project-id somewhere.

PR #4469, opened 2017-11-28
Changing `dev` to `dev1` in Pub / Sub version.

dhermes:
This is a follow-up to #4462, where I forgot to add it. In
the current form, `pip` adds a 0 at the end for us:

```
$ virtualenv venv
$ venv/bin/pip install pubsub/
$ venv/bin/pip show google-cloud-pubsub
Name: google-cloud-pubsub
Version: 0.29.2.dev0
...
```

### DIFF
--- a/pubsub/setup.py
+++ b/pubsub/setup.py
@@ -59,7 +59,7 @@ REQUIREMENTS = [
 
 setup(
     name='google-cloud-pubsub',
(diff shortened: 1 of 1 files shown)


dhermes:
I am "auto-merging" without review since this is so straightforward.

PR #4470, opened 2017-11-28
BigQuery docs: adds headers to API reference page

alixhami:
This PR adds headers to visually separate the sections in the API reference page of the BigQuery client library docs.


alixhami:
@tswast It didn't change the TOC.

Issue #4463, opened 2017-11-27
PubSub: `set_exception can only be called once.` can still occur "erroneously"


dhermes:
For example, If I add

```python
_LOGGER.debug('thread.Policy.on_exception: %s', exception)
```

just [before `future.set_exception()`][1], then, when forcing a failure by:

```python
from google.cloud.pubsub_v1.subscriber.policy import base
# Monkey-patch to intentionally miss errors.
base.BasePolicy._RETRYABLE_STREAM_ERRORS = ()
```

I see (focus on **00117233** and **00185066**):

```
...
00116110:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :The current p99 value is 10 seconds.
00116111:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Renewing lease for 0 ack IDs.
00116111:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Snoozing lease management for 2.555879 seconds.
00117233:DEBUG:google.cloud.pubsub_v1.subscriber.policy.thread:Consumer helper: consume bidirectional stream:thread.Policy.on_exception: 503 The service was unable to fulfill your request. Please try agai
n. [code=8a75]
00117236:DEBUG:google.cloud.pubsub_v1.subscriber._consumer:Thread-23             :Sending initial request: subscription: "projects/precise-truck-742/subscriptions/s-djh-local-1511822800630"
stream_ack_deadline_seconds: 10

00118670:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :The current p99 value is 10 seconds.
00118670:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Renewing lease for 0 ack IDs.
00118670:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Snoozing lease management for 1.883336 seconds.
...
00181929:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :The current p99 value is 10 seconds.
00181929:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Renewing lease for 0 ack IDs.
00181929:DEBUG:google.cloud.pubsub_v1.subscriber.policy.base:Thread-8              :Snoozing lease management for 7.864327 seconds.
[...]

PR #4471, opened 2017-11-28
Pub/Sub logging hygiene.

dhermes:
Uses newline character when logging protobufs that may end up having newlines in them. These are harder to grok in the log output since it "blends" the protobuf and the description. For example.

```
Received response: received_messages {
  ack_id: "Pn41MEV..."
  message {
    data: ...
```

vs.

```
Received response:
received_messages {
  ack_id: "Pn41MEV..."
  message {
    data: ...
```

Also

- replaced a `lambda` with a simple function
- added docstrings to two default callbacks used by the `policy.thread.Policy` implementation.
- updated grammatical error in a comment
- replaced a bare `if future` with `if future is not None`

----

Context: I have discovered each of these "issues" while dealing with #4463 and #4234.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -32,8 +32,32 @@ _LOGGER = logging.getLogger(__name__)
 
 
 def _callback_completed(future):
-    """Simple callback that just logs a `Future`'s result."""
+    """Simple callback that just logs a future's result.
+
+    Used on completion of processing a message received by a
+    subscriber.
+
+    Args:
+        future (concurrent.futures.Future): A future returned
+            from :meth:`~concurrent.futures.Executor.submit`.
+    """
     _LOGGER.debug('Result: %s', future.result())
+
+
+def _do_nothing_callback(message):
+    """Default callback for messages received by subscriber.
+
+    Does nothing with the message and returns :data:`None`.
+
+    Args:
+        message (~google.cloud.pubsub_v1.subscriber.message.Message): A
+            protobuf message returned by the backend and parsed into
+            our high level message type.
+
+    Returns:
+        NoneType: Always.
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -167,7 +191,7 @@ def on_response(self, response):
         For each message, schedule a callback with the executor.
         """
         for msg in response.received_messages:
-            _LOGGER.debug('New message received from Pub/Sub: %r', msg)
+            _LOGGER.debug('New message received from Pub/Sub:\n%r', msg)
```
dhermes: @lukesneeringer Is there a reason you used `%r` instead of `%s` here? There is actually no distinction for protobuf types:

```python
>>> from google.cloud.pubsub_v1 import types
>>> msg = types.PubsubMessage(data=b'a', message_id=b'123f')
>>> str(msg)
'data: "a"\nmessage_id: "123f"\n'
>>> repr(msg)
'data: "a"\nmessage_id: "123f"\n'
>>> msg
data: "a"
message_id: "123f"

>>>
```

lukesneeringer: Because in principle, `%r` is what I would want if there _was_ a distinction.

dhermes: @lukesneeringer So then I should change every other line to do the same? (i.e. `%r` rather than  `%s`, `{!r}` rather than `{}`)

dhermes: I went ahead and did it.

dhermes:
Changed 3 instances of `%s` to `%r` (channeling Luke). Will merge without waiting for CI to finish as this change doesn't impact any of the tests (I checked locally).

PR #4472, opened 2017-11-28
PubSub: `Policy.on_exception` actually used to make consumer go inactive.

### LINKED ISSUE #4463: PubSub: `set_exception can only be called once.` can still occur "erroneously"

dhermes:
Towards #4463.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -244,10 +244,9 @@ class Consumer(object):
                 self.stop_consuming()
             except Exception as exc:
                 try:
-                    self._policy.on_exception(exc)
+                    self.active = self._policy.on_exception(exc)
                 except:
                     self.active = False
-                    raise
 
     def start_consuming(self):
         """Start consuming the stream."""
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -243,10 +243,7 @@ class Consumer(object):
             except KeyboardInterrupt:
                 self.stop_consuming()
             except Exception as exc:
-                try:
-                    self.active = self._policy.on_exception(exc)
-                except:
-                    self.active = False
+                self.active = self._policy.on_exception(exc)
 
     def start_consuming(self):
         """Start consuming the stream."""
--- a/pubsub/google/cloud/pubsub_v1/subscriber/futures.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/futures.py
@@ -39,16 +39,19 @@ class Future(futures.Future):
 
         .. note::
 
-            A ``False`` value here does not necessarily mean that the
+            A :data:`False` value here does not necessarily mean that the
             subscription is closed; it merely means that _this_ future is
             not the future applicable to it.
 
             Since futures have a single result (or exception) and there is
-            not a concept of resetting them, a closing re-opening of a
+            not a concept of resetting them, a closing / re-opening of a
             subscription will therefore return a new future.
 
         Returns:
-            bool: ``True`` if this subscription is opened with this future,
-                ``False`` otherwise.
+            bool: :data:`True` if this subscription is opened with this
+            future, :data:`False` otherwise.
         """
-        return self._policy.future is self
+        if self._policy.future is not self:
+            return False
+
+        return super(Future, self).running()
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
@@ -91,12 +91,16 @@ def test_blocking_consume_keyboard_interrupt():
 
 class OnException(object):
 
-    def __init__(self, exiting_event):
+    def __init__(self, exiting_event, acceptable=None):
         self.exiting_event = exiting_event
+        self.acceptable = acceptable
 
     def __call__(self, exception):
-        self.exiting_event.set()
-        return False
+        if exception is self.acceptable:
+            return True
+        else:
+            self.exiting_event.set()
+            return False
 
 
 def test_blocking_consume_on_exception():
@@ -115,6 +119,31 @@ def test_blocking_consume_on_exception():
     policy.call_rpc.assert_called_once()
     policy.on_response.assert_called_once_with(mock.sentinel.A)
     policy.on_exception.assert_called_once_with(exc)
+
+
+def test_blocking_consume_two_exceptions():
+    policy = mock.Mock(spec=('call_rpc', 'on_response', 'on_exception'))
+    policy.call_rpc.side_effect = (
+        (mock.sentinel.A,),
+        (mock.sentinel.B,),
+    )
+    exc1 = NameError('Oh noes.')
+    exc2 = ValueError('Something grumble.')
+    policy.on_response.side_effect = (exc1, exc2)
+
+    consumer = _consumer.Consumer(policy=policy)
+    policy.on_exception.side_effect = OnException(
+        consumer._exiting, acceptable=exc1)
+
+    # Establish that we get responses until we are sent the exiting event.
+    consumer._blocking_consume()
+
+    # Check mocks.
+    assert policy.call_rpc.call_count == 2
+    policy.on_response.assert_has_calls(
+        [mock.call(mock.sentinel.A), mock.call(mock.sentinel.B)])
+    policy.on_exception.assert_has_calls(
+        [mock.call(exc1), mock.call(exc2)])
 
 
 def test_start_consuming():
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import types as base_types
+
 from google.auth import credentials
 import mock
 import pytest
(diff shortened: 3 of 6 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/futures.py
```diff
@@ -39,16 +39,19 @@ def running(self):
...
             Since futures have a single result (or exception) and there is
-            not a concept of resetting them, a closing re-opening of a
+            not a concept of resetting them, a closing / re-opening of a
             subscription will therefore return a new future.
 
         Returns:
-            bool: ``True`` if this subscription is opened with this future,
-                ``False`` otherwise.
+            bool: :data:`True` if this subscription is opened with this
+            future, :data:`False` otherwise.
```
dhermes: @lukesneeringer Can you verify this **note** above?

pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
```diff
@@ -87,18 +89,55 @@ def test_blocking_consume_keyboard_interrupt():
...
+
+
+def test_blocking_consume_exception_normal_handling():
+    policy = mock.Mock(spec=('call_rpc', 'on_response', 'on_exception'))
+    policy.call_rpc.return_value = (mock.sentinel.A, mock.sentinel.B)
+    exc = TypeError('Bad things!')
+    policy.on_response.side_effect = exc
+
+    consumer = _consumer.Consumer(policy=policy)
+    policy.on_exception.side_effect = OnException(consumer._exiting)
```
dhermes: @lukesneeringer This is a nasty hack that doesn't **really** test what's going on. AFAICT setting `Consumer().active = False` stops the thread in a very indirect [place][1].

1. Is this correct?
1. Any better ideas for testing it?

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/56fc8843728e606523f035763f93d8d7bd4aeb90/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py#L289-L291

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -244,10 +244,9 @@ def _blocking_consume(self):
                 self.stop_consuming()
             except Exception as exc:
                 try:
-                    self._policy.on_exception(exc)
+                    self.active = self._policy.on_exception(exc)
                 except:
                     self.active = False
```
jonparrott: Is this branch needed any longer?

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -244,10 +244,9 @@ def _blocking_consume(self):
                 self.stop_consuming()
             except Exception as exc:
                 try:
-                    self._policy.on_exception(exc)
+                    self.active = self._policy.on_exception(exc)
```
jonparrott: Active gets set to `False` here, but what actually makes the consumer exit? From what I can tell, active will be set to false but the next loop here will just re-open the stream.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
dhermes: It depends. Do we want to prepare for the case that implementations of `on_exception` can fail?

jonparrott: I think it's less useful to swallow the exception, as it would be truly exception if `on_exception` exceptioned.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
dhermes: I am also hazy on this. See my [note][1], AFAICT `.close()` (which happens [here][2]) is what cleans up (it's the only thing that can stop `Consumer._blocking_consume` since `Consumer.stop_consuming` is the only thing that calls `self._exiting.set()`).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4472/files#r153624395
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/56fc8843728e606523f035763f93d8d7bd4aeb90/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py#L289-L291

jonparrott: Yeah, it seems to me that `active` is just an external signal and doesn't impact flow at all within this class (maybe that's worth adding a comment).

ISTM that an unrecoverable error here should have the same impact as `stop_consuming` below.

dhermes:
OK I have updated the [gist][1] to be much clearer as to how things **were** behaving and how they behave now. Most notable is:

- Heartbeats where `running=True, done=True` in the `logs_pypi.txt` (happens immediately after `thread.Policy.on_exception` is called)
- Lease management (e.g. "Snoozing lease management...") ceases in `logs_local.txt` immediately after the heartbeats change to `running=False, done=True`. Just to be sure, we have 40 seconds worth of heartbeats before stopping the main thread and there is still no lease management happening

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
lukesneeringer: Agreed.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
lukesneeringer: I _think_ that is the best answer.

pubsub/google/cloud/pubsub_v1/subscriber/futures.py
lukesneeringer: That note is still correct. However, with your change, a `True` value _should_ necessarily mean that the subscription is open.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
dhermes: I dropped the `try/except` and just always set `active`. I am going to add another unit test which is OK with one exception but not with another.

dhermes:
Remaining point of discussion (got [buried][3] below the fold):

----

I am also hazy on this. See my [note][1], AFAICT `thread.Policy.close()` (which gets called by [`BasePolicy.lease()`][2]) is what cleans up (it's the only thing that can stop `Consumer._blocking_consume` since `Consumer.stop_consuming` is the only thing that calls `self._exiting.set()`).

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4472/files#r153624395
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/56fc8843728e606523f035763f93d8d7bd4aeb90/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py#L289-L291
[3]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4472#discussion_r153630816

dhermes:
OK my [previous remark][1] seems to be wrong, instead it seems that `policy.base.BasePolicy.maintain_leases` is [where the action is][2].

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4472#issuecomment-347712249
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/6120936a5e98541f6bfe3a1a6a1d00054c088487/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py#L311-L312

dhermes:
I am going to merge this since it "moves the bar" towards correctness even though I know it isn't 100% correct. E.g. I can still trigger "set_exception can only be called once." by using a subscription that doesn't exist.

PR #4474, opened 2017-11-29
PubSub: Naming the explicitly spawned threads

dhermes:


pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -64,7 +64,8 @@ def start(self, name, queue, target, *args, **kwargs):
         thread = threading.Thread(
             name='Consumer helper: {}'.format(name),
             target=target,
-            *args, **kwargs
+            *args,
```
dhermes: @lukesneeringer @jonparrott Any ideas for a better name here? We only have two uses ([here][1] and [here][1]) which result in

```
Consumer helper: callback requests worker
Consumer helper: consume bidirectional stream
```

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/10e6f998be063854112ddf38241ceb6e213f5aef/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py#L151-L155
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/10e6f998be063854112ddf38241ceb6e213f5aef/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py#L256-L260

PR #4476, opened 2017-11-29
Renaming "Consumer helper" threads in Pub/Sub.

dhermes:
Now names will show up in logs as

```
Thread-ConsumerHelper-CallbackRequestsWorker
Thread-ConsumerHelper-ConsumeBidirectionalStream
```

instead of as

```
Consumer helper: callback requests worker
Consumer helper: consume bidirectional stream
```

This is a follow up to #4474.


dhermes:
Just got LGTM from @lukesneeringer via chat.

PR #4480, opened 2017-11-29
Adding thread prefix for ThreadPoolExecutor in Pub/Sub.

dhermes:
This is a follow-on to  #4474 and #4476.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -106,7 +106,10 @@ class Policy(base.BasePolicy):
 
         # Also maintain a request queue and an executor.
         if executor is None:
-            executor = futures.ThreadPoolExecutor(max_workers=10)
+            executor = futures.ThreadPoolExecutor(
+                max_workers=10,
+                thread_name_prefix='ThreadPoolExecutor-SubscriberPolicy',
+            )
         self._executor = executor
         _LOGGER.debug('Creating callback requests thread (not starting).')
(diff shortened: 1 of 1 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -106,7 +106,10 @@ def __init__(self, client, subscription, flow_control=types.FlowControl(),
         # Also maintain a request queue and an executor.
         _LOGGER.debug('Creating callback requests thread (not starting).')
         if executor is None:
-            executor = futures.ThreadPoolExecutor(max_workers=10)
+            executor = futures.ThreadPoolExecutor(
+                max_workers=10,
+                thread_name_prefix='ThreadPoolExecutor-SubscriberPolicy',
```
dhermes: Original default is `''` which results in thread names like:

```
ThreadPoolExecutor-0_0
ThreadPoolExecutor-0_1
ThreadPoolExecutor-0_2
ThreadPoolExecutor-0_3
ThreadPoolExecutor-0_4
```

which AFAICT correspond to the prefix `ThreadPoolExecutor-0`.

dhermes:
This feature is not present in the `concurrent.futures` backport, but has been checked into their source tree (https://github.com/agronholm/pythonfutures/issues/63).

jonparrott:
> This feature is not present in the concurrent.futures backport, but has been checked into their source tree (agronholm/pythonfutures#63).

great. Can we turn it on for Python 3 for now?

dhermes:
@jonparrott We can, but the maintainer of the backport indicated he may do a release soon, so it might be worth waiting (if it's on the order of a day).

PR #4483, opened 2017-11-29
Removing unused "extra" arguments in Pub/Sub helper thread start.

dhermes:


dhermes:
Merging this based on discussion in chat with @lukesneeringer

PR #4497, opened 2017-11-29
 PubSub: Raises TypeError on noncallable callback in subscriber client

### LINKED ISSUE #4487: Pub/Sub Subscriber Client.subscribe() may drop callback on the floor
From [source][1]

```python
    def subscribe(self, subscription, callback=None, flow_control=()):
        ...
        if callable(callback):
            subscr.open(callback)
```

Instead it should be something like:

```python
if callback is not None:
    if callable(callback):
        subscr.open(callback)
    else:
        raise TypeError('Hey man, that is not, like, callable, man.')
```

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/cc7da65b4a31d7a33055afe0bd6dc9855c3d6493/pubsub/google/cloud/pubsub_v1/subscriber/client.py#L130-L131

chemelnucfin:
Closes #4487

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/client.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/client.py
@@ -124,9 +124,15 @@ class Client(object):
         Returns:
             ~.pubsub_v1.subscriber.consumer.base.BaseConsumer: An instance
                 of the defined ``consumer_class`` on the client.
+
+        Raises:
+            TypeError: If ``callback`` is not callable.
         """
         flow_control = types.FlowControl(*flow_control)
         subscr = self._policy_class(self, subscription, flow_control)
         if callable(callback):
             subscr.open(callback)
+        elif callback is not None:
+            error = '{!r} is not callable, please check input'.format(callback)
+            raise TypeError(error)
         return subscr
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
@@ -14,6 +14,7 @@
 
 from google.auth import credentials
 import mock
+import pytest
 
 from google.cloud.pubsub_v1 import subscriber
 from google.cloud.pubsub_v1.subscriber.policy import thread
@@ -54,3 +55,12 @@ def test_subscribe_with_callback():
         subscription = client.subscribe('sub_name_b', callback)
         open_.assert_called_once_with(callback)
     assert isinstance(subscription, thread.Policy)
+
+
+def test_subscribe_with_failed_callback():
+    creds = mock.Mock(spec=credentials.Credentials)
+    client = subscriber.Client(credentials=creds)
+    callback = 'abcdefg'
+    with pytest.raises(TypeError) as exc_info:
+        subscription = client.subscribe('sub_name_b', callback)
+    assert callback in str(exc_info.value)


pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -129,4 +129,9 @@ def subscribe(self, subscription, callback=None, flow_control=()):
         subscr = self._policy_class(self, subscription, flow_control)
         if callable(callback):
             subscr.open(callback)
+        else:
+            if callback is not None:
```
dhermes: Nit: collapse one indent level and just us `elif callback is not None`

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -129,4 +129,9 @@ def subscribe(self, subscription, callback=None, flow_control=()):
...
+                string = "{!r} is not callable, please check input"
```
dhermes: 1. Never use `string` as a variable name. Instead here you could say `template` or just avoid the template all together and just do `error = '{!r} ...'.format(callback)`
1. Please use single quotes in your string

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
```diff
@@ -38,19 +40,26 @@ def test_init_emulator(monkeypatch):
     channel = client.api.subscriber_stub.Pull._channel
     assert channel.target().decode('utf8') == '/baz/bacon/'
 
+class TestSubscribe(unittest.TestCase):
+    def test_subscribe(self):
```
dhermes: Please remove this class. This module was written in the `pytest` style and is fine in it's current format.

Also, if you **were** to leave it as a class, put an empty line between the class declaration and the first method. (But just delete the class, this is for next time.)

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
```diff
@@ -38,19 +40,26 @@ def test_init_emulator(monkeypatch):
...
-    with mock.patch.object(thread.Policy, 'open') as open_:
-        subscription = client.subscribe('sub_name_b', callback)
-        open_.assert_called_once_with(callback)
-    assert isinstance(subscription, thread.Policy)
+    def test_subscribe_with_failed_callback(self):
+        creds = mock.Mock(spec=credentials.Credentials)
+        client = subscriber.Client(credentials=creds)
+        callback = 'abcdefg'
+        with self.assertRaisesRegexp(TypeError, "'abcdefg' is not callable"):
+            with mock.patch.object(thread.Policy, 'open') as open_:
```
dhermes: You don't need to patch this since you'll fail before you call `open()`.

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
```diff
@@ -38,19 +40,26 @@ def test_init_emulator(monkeypatch):
...
-    callback = mock.Mock()
-    with mock.patch.object(thread.Policy, 'open') as open_:
-        subscription = client.subscribe('sub_name_b', callback)
-        open_.assert_called_once_with(callback)
-    assert isinstance(subscription, thread.Policy)
+    def test_subscribe_with_failed_callback(self):
+        creds = mock.Mock(spec=credentials.Credentials)
+        client = subscriber.Client(credentials=creds)
+        callback = 'abcdefg'
+        with self.assertRaisesRegexp(TypeError, "'abcdefg' is not callable"):
```
dhermes: I am :-1: on hardcoding the message in the `Regexp`, but feel free to leave it.

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
chemelnucfin: So do I import pytest for pytest.raises?

dhermes: Yes.

chemelnucfin:
@dhermes PTAL, thanks.

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -129,4 +129,7 @@ def subscribe(self, subscription, callback=None, flow_control=()):
         subscr = self._policy_class(self, subscription, flow_control)
         if callable(callback):
             subscr.open(callback)
+        elif callback is not None:
+            error = "{!r} is not callable, please check input".format(callback)
```
dhermes: @chemelnucfin Single quotes

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
```diff
@@ -54,3 +54,13 @@ def test_subscribe_with_callback():
         subscription = client.subscribe('sub_name_b', callback)
         open_.assert_called_once_with(callback)
     assert isinstance(subscription, thread.Policy)
+
+
+def test_subscribe_with_failed_callback():
+    import pytest
```
dhermes: @chemelnucfin Please import this at module level (but thanks for trying to adhere to some semblance of a rule)

pubsub/tests/unit/pubsub_v1/subscriber/test_subscriber_client.py
```diff
@@ -54,3 +54,13 @@ def test_subscribe_with_callback():
...
+    creds = mock.Mock(spec=credentials.Credentials)
+    client = subscriber.Client(credentials=creds)
+    callback = 'abcdefg'
+    with pytest.raises(TypeError) as exception:
+        subscription = client.subscribe('sub_name_b', callback)
+    assert callback in str(exception.value)
```
dhermes: :+1: For this check. I'd suggest renaming `exception` to `exc_info` (it's not actually an exception, it's some context manager provided by `pytest`)

chemelnucfin:
@dhermes should be ready.

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -129,4 +129,7 @@ def subscribe(self, subscription, callback=None, flow_control=()):
         subscr = self._policy_class(self, subscription, flow_control)
         if callable(callback):
             subscr.open(callback)
+        elif callback is not None:
```
dhermes: Sorry I missed this, but you need to add a `Raises:` section to this docstring

chemelnucfin: My fault for forgetting.  :)
Added.  Squash and merge on green?

pubsub/google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -124,6 +124,9 @@ def subscribe(self, subscription, callback=None, flow_control=()):
         Returns:
             ~.pubsub_v1.subscriber.consumer.base.BaseConsumer: An instance
                 of the defined ``consumer_class`` on the client.
+
+        Raises:
+            :exc:`TypeError` if callback is not callable.
```
dhermes: I believe the "correct" way to do this with a Google-style docstring is:

```
Raises:
    TypeError: If ``callback`` is not callable.
```

(I added the backticks around `callback` so that it would be in code font, that's just a personal preference.)

PR #4502, opened 2017-11-30
WIP: Spanner: Bad constructor behavior

chemelnucfin:
According to the documents, https://cloud.google.com/spanner/docs/reference/rest/v1/KeySet,
these should be errors.  This came up when I was looking at partial ranges.
The to_pb method does not work if the keys aren't specified.


chemelnucfin:
On second thought, let me think about this further.

chemelnucfin:
Something is wrong with this, I just haven't figured out how to fix it yet.  WIP.

chemelnucfin:
See #4515

PR #4503, opened 2017-11-30
Make sure inactive request generator is stopped before spawning new one in Pub / Sub consumer.

dhermes:
~~This makes the `request_queue` used by `Consumer._request_generator_thread()` local to the caller. This way on recovery a new queue can be used for the new thread.~~

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -249,6 +249,15 @@ class Consumer(object):
             except ValueError:
                 # Should be ``ValueError('generator already executing')``
                 if not self._request_queue.empty():
+                    # This case may be a false negative in **very** rare
+                    # cases. We **assume** that the generator can't be
+                    # ``close()``-ed because it is blocking on ``get()``.
+                    # It's **very unlikely** that the generator was not
+                    # blocking, but instead **in between** the blocking
+                    # ``get()`` and the next ``yield`` / ``break``. However,
+                    # for practical purposes, we only need to stop the request
+                    # generator if the connection has timed out due to
+                    # inactivity, which indicates an empty queue.
                     _LOGGER.debug(
                         'Request generator could not be closed but '
                         'request queue is not empty.')
@@ -263,7 +272,15 @@ class Consumer(object):
                     'Waiting for active request generator to receive STOP')
                 while not self._request_queue.empty():
                     pass
-                request_generator.close()
+                # We would **like** to call ``request_generator.close()`` here
+                # but can't guarantee that the generator is paused, since it
+                # has a few instructions to complete between the ``get()``
+                # and the ``break``. However, we are confident that
+                #   1. The queue was empty and we hold the ``put()`` lock
+                #   2. We added ``STOP``
+                #   3. We waited until the request generator consumed ``STOP``
+                # so we know the request generator **will** stop within a
+                # few cycles.
             except Exception as exc:
                 _LOGGER.error('Failed to close request generator: %r', exc)
                 return False
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -180,6 +180,7 @@ class Consumer(object):
         self._policy = policy
         self._request_queue = queue.Queue()
         self._exiting = threading.Event()
+        self._put_lock = threading.Lock()
 
         self.active = False
         self.helper_threads = _helper_threads.HelperThreadRegistry()
@@ -193,20 +194,17 @@ class Consumer(object):
         Args:
             request (Any): The request protobuf.
         """
-        self._request_queue.put(request)
+        with self._put_lock:
+            self._request_queue.put(request)
 
-    def _request_generator_thread(self, request_queue):
+    def _request_generator_thread(self):
         """Generate requests for the stream.
 
         This blocks for new requests on the request queue and yields them to
         gRPC.
 
-        Args:
-            request_queue (~queue.Queue): A queue where subscribers will add
-                requests to be sent.
-
         Yields:
-            google.protobuf.message.Message: Protobuf requests.
+            google.cloud.pubsub_v1.types.StreamingPullRequest: Requests
         """
         # First, yield the initial request. This occurs on every new
         # connection, fundamentally including a resumed connection.
@@ -217,13 +215,52 @@ class Consumer(object):
         # Now yield each of the items on the request queue, and block if there
         # are none. This can and must block to keep the stream open.
         while True:
-            request = request_queue.get()
+            request = self._request_queue.get()
             if request == _helper_threads.STOP:
                 _LOGGER.debug('Request generator signaled to stop.')
                 break
 
             _LOGGER.debug('Sending request:\n%r', request)
             yield request
+
+    def _stop_request_generator(self, request_generator):
+        """Ensure a request generator is closed.
+
+        This **must** be done when recovering from a retry-able exception.
+        If not, then an inactive request generator (i.e. not attached to any
+        actual RPC) will be trying to access the same request queue as the
+        active request generator.
+
+        In addition, we want the gRPC thread consuming to cleanly exit so
+        that system resources are not wasted.
+
+        Args:
+            request_generator (Generator): A streaming pull request generator
+                returned from :meth:`_request_generator_thread`.
+
+        Raises:
+            ValueError: If ``.close()`` fails for any reason other than
+                "generator already executing".
+            ValueError: If the queue is not empty and the generator is
+                running.
+        """
+        with self._put_lock:
+            try:
+                request_generator.close()
+            except ValueError as exc:
+                if exc.args != ('generator already executing',):
+                    raise
+                if not self._request_queue.empty():
+                    raise ValueError('Queue expected to be empty.')
+                # If we **cannot** close the request generator,
+                # then there is no blocking get on the queue. Since
+                # we have locked ``.put()`` this means that the
+                # queue **was** and remains empty.
+                self._request_queue.put(_helper_threads.STOP)
+                # Wait for the request generator to ``.get()`` the ``STOP``.
+                while not self._request_queue.empty():
+                    pass
+                request_generator.close()
 
     def _blocking_consume(self):
         """Consume the stream indefinitely."""
@@ -236,8 +273,7 @@ class Consumer(object):
                 _LOGGER.debug('Event signalled consumer exit.')
                 break
 
-            request_generator = self._request_generator_thread(
-                self._request_queue)
+            request_generator = self._request_generator_thread()
             response_generator = self._policy.call_rpc(request_generator)
             try:
                 for response in response_generator:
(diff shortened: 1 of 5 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -244,7 +252,19 @@ def _blocking_consume(self):
                 break
             except Exception as exc:
                 recover = self._policy.on_exception(exc)
-                if not recover:
+                if recover:
+                    with threading.Lock():
```
lukesneeringer: I am fairly certain that `with threading.Lock():` used in this way does not work, because it is always using a different lock. You need to save a single lock object to the consumer instance so you are calling the `__enter__` method of the same lock.

Basically, `self._recovery_lock = threading.Lock()` in the `__init__.py` file, and then `with self._recovery_lock:` here.

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -60,12 +60,13 @@ def __init__(self):
     def __contains__(self, needle):
         return needle in self._helper_threads
 
-    def start(self, name, queue, target):
+    def start(self, name, queue_put, target):
```
lukesneeringer: What is the rationale behind this change?

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -244,7 +252,19 @@ def _blocking_consume(self):
...
+                        # Must lock so that ``self.send_request()`` cannot add
+                        # more items to the queue while it is being replaced.
+                        previous_queue = self._request_queue
```
dhermes: @lukesneeringer After chatting with @jonparrott it seems there is another option here.

If we could guarantee that `self._request_queue.empty()` is `True`, then we could just send `STOP` into the queue and wait until the `STOP` gets consumed, then we don't have to swap the queue.

(As you can see from the coverage failure, the queue will always be empty.)

lukesneeringer: That makes sense to me.

dhermes: #4504 is an implementation of that other option

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
dhermes: See where I am sending in `Consumer.send_request()` rather than passing in `Consumer._request_queue`. The problem is that if you send in a queue and then swap it off of the `Consumer`, then the helper thread won't be able to talk to the active request generator.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
dhermes: What is the issue with always using a different lock?

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
lukesneeringer: Understood.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
lukesneeringer: It will never actually lock anything.

dhermes: BLERG

lukesneeringer: Longer explanation: `threading.Lock()` creates a single, discrete lock. To illustrate, imagine a world where we have two separate locks that lock two separate things:

To illustrate, consider a kitchen with two cooks. Both cooks want to make scrambled eggs on the stove top, and they want to put bread in the toaster. Since there is only one stove top and only one toaster, the cooks can not use them at the same time, but it would be fine for one cook to use the stove top while the other one uses the toaster.

These are different lock instances:

```python
stove_lock = threading.Lock()
toaster_lock = threading.Lock()

with stove_lock:
    scramble_eggs()

with toaster_lock:
    toast_bread()
```

When you call `with toaster_lock:`, if another thread is currently executing code within a separate `with toaster_lock:`, then your thread will block until the other thread calls `toaster_lock.__exit__`. However, it does not matter if another thread is currently using `stove_lock`.

But when you just do `with threading.Lock():` then you are creating a _different_ lock instance every time, and it is not actually possible for anything else to be checking against the same lock. The other threads are just blindly using their own, distinct locks and not checking in with yours.

dhermes:
@jonparrott @lukesneeringer PTAL

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,45 @@ def _request_generator_thread(self):
...
+            ValueError: If ``.close()`` fails for any reason other than
+                "generator already executing".
+            ValueError: If the queue is not empty and the generator is
+                running.
+        """
+        with self._put_lock:
+            try:
+                request_generator.close()
+            except ValueError as exc:
+                if exc.args != ('generator already executing',):
```
dhermes: This is not so great to have this be "stringly-typed", I could just assume that a `ValueError` means it's already executing.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,45 @@ def _request_generator_thread(self):
...
+
+        Raises:
+            ValueError: If ``.close()`` fails for any reason other than
+                "generator already executing".
+            ValueError: If the queue is not empty and the generator is
+                running.
+        """
+        with self._put_lock:
+            try:
+                request_generator.close()
```
dhermes: There **might** is a seemingly impossible-to-trigger corner case where the generator is **still running** for the few nanoseconds **after** the blocking `.get()` returns. In that case, we'd either need to

- Somehow force the gRPC thread that is consuming the request generator to knock it off
- Just try to `.close()` one more time if we detect the queue isn't empty

----

FWIW I don't think it's worth accounting for that impossible-to-trigger corner case.

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
dhermes: @lukesneeringer The rationale has change slightly, since I no longer swap out the request queue. However, due to the `_put_lock` in `send_request()`, it still makes sense to avoid passing the queue itself.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
lukesneeringer: I trust your judgment on this.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,45 @@ def _request_generator_thread(self):
...
+                    raise
+                if not self._request_queue.empty():
+                    raise ValueError('Queue expected to be empty.')
+                # If we **cannot** close the request generator,
+                # then there is no blocking get on the queue. Since
+                # we have locked ``.put()`` this means that the
+                # queue **was** and remains empty.
+                self._request_queue.put(_helper_threads.STOP)
+                # Wait for the request generator to ``.get()`` the ``STOP``.
+                while not self._request_queue.empty():
```
jonparrott: if we use `task_done` we can use [queue.join](https://docs.python.org/2/library/queue.html#Queue.Queue.join).

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,45 @@ def _request_generator_thread(self):
...
+            ValueError: If the queue is not empty and the generator is
+                running.
+        """
+        with self._put_lock:
+            try:
+                request_generator.close()
+            except ValueError as exc:
+                if exc.args != ('generator already executing',):
+                    raise
+                if not self._request_queue.empty():
```
jonparrott: I thought we discussed that it's perfectly valid for the queue to have some pending items in it?

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,54 @@ def _request_generator_thread(self):
...
+                # then there is no blocking get on the queue. Since
+                # we have locked ``.put()`` this means that the
+                # queue **was** and remains empty.
+                self._request_queue.put(_helper_threads.STOP)
+                # Wait for the request generator to ``.get()`` the ``STOP``.
+                _LOGGER.debug(
+                    'Waiting for active request generator to receive STOP')
+                while not self._request_queue.empty():
+                    pass
+                request_generator.close()
```
dhermes: Luckily I re-factored the unit tests, because it made me realize that this could fail, because the generator still has to do:

```python
if request == _helper_threads.STOP:
    logger.debug()
    break
```

after the blocking `get()` completes, so this call to `close()` might happen **before** the generator gets those instructions.

I'm inclined to say we should either:

- Drop this statement completely (we are "safe" since we know that `STOP` has been pulled off the queue)
- Do a `while True` loop that tries to `close()` and `break`-s if it succeeds
- Do a `while request_generator.gi_running: pass` loop to block until it stops running and then call `close()`

jonparrott: I am in favor of dropping `close()` as we know that as long as `STOP` (or *any* item in the queue) has been consumed post-RPC-cancellation that gRPC will exit the thread.

dhermes: Me too. Jinx.

dhermes:
@lukesneeringer @jonparrott PTAL. Phew I think it's finally merge-able.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -218,6 +223,71 @@ def _request_generator_thread(self):
...
+                    # ``get()`` and the next ``yield`` / ``break``. However,
+                    # for practical purposes, we only need to stop the request
+                    # generator if the connection has timed out due to
+                    # inactivity, which indicates an empty queue.
+                    _LOGGER.debug(
+                        'Request generator could not be closed but '
+                        'request queue is not empty.')
+                    return False
+                # If we **cannot** close the request generator,
+                # then there is no blocking get on the queue. Since
```
jonparrott: Do you mean that there *is* a blocking get?

pubsub/tests/unit/pubsub_v1/subscriber/test_consumer.py
```diff
@@ -142,6 +140,106 @@ def test_start_consuming():
...
+
+
+def basic_queue_generator(queue, received):
+    while True:
+        value = queue.get()
+        received.put(value)
+        yield value
+
+
+def test_stop_request_generator_not_running():
```
jonparrott: I would love some comments explaining the scenarios that each test exercise.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
dhermes: Yes, thanks.

dhermes: Updated

dhermes:
I added explanations for each unit test @jonparrott. Also updated one of them to use a real generator instead of a `mock`. Will merge once CI is green.

This code is no fun.

PR #4504, opened 2017-11-30
Send STOP to defunct consumer threads without swapping request queue.

dhermes:
This is a simpler alternative to #4503.


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -244,6 +244,19 @@ def _blocking_consume(self):
                 break
             except Exception as exc:
                 recover = self._policy.on_exception(exc)
+                if recover:
+                    with threading.Lock():
```
lukesneeringer: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4503#discussion_r154212590

PR #4515, opened 2017-12-02
Spanner: Partial range test and bug fix

chemelnucfin:
I need to clean this up a bit, but feel free to have a brief look in case I'm doing something atrocious.

I changed the behavior of a default constructor to include all instead of raising ValueError, which is the same behavior exhibited if you don't specify a end_key.

Also fixed a bug where if an end_key doesn't get specified, it adds the required field before it passes into pb.

### DIFF
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -38,6 +38,9 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
+        if not any([start_open, start_closed, end_open, end_closed]):
+            raise ValueError("Must specify at least a start or end row.")
+
         if start_open is not None and start_closed is not None:
             raise ValueError("Specify one of 'start_open' / 'start_closed'.")
 
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -38,13 +38,10 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
-            raise ValueError("Must specify at least a start or end row.")
-
-        if start_open and start_closed:
+        if start_open is not None and start_closed is not None:
             raise ValueError("Specify one of 'start_open' / 'start_closed'.")
 
-        if end_open and end_closed:
+        if end_open is not None and end_closed is not None:
             raise ValueError("Specify one of 'end_open' / 'end_closed'.")
 
         self.start_open = start_open
@@ -60,17 +57,19 @@ class KeyRange(object):
         """
         kwargs = {}
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-
-        if self.start_closed:
+        elif self.start_closed is not None:
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
+        else:
+            kwargs['start_closed'] = _make_list_value_pb([])
 
-        if self.end_open:
+        if self.end_open is not None:
             kwargs['end_open'] = _make_list_value_pb(self.end_open)
-
-        if self.end_closed:
+        elif self.end_closed is not None:
             kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
+        else:
+            kwargs['end_closed'] = _make_list_value_pb([])
 
         return KeyRangePB(**kwargs)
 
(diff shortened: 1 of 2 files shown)


tseaver:
@chemelnucfin I'm not in favor of `KeyRange()` as a valid way to spell "include all rows" -- if that was what I wanted, why would I use a range at all?

chemelnucfin:
@tseaver 
Maybe someone needs to pass in a KeyRange into a function expecting that object which selects all rows?  Also the default behavior in protobuf is like that, isn't it?

chemelnucfin:
@tseaver PTAL, thanks.

tseaver:
@chemelnucfin I'm still opposed to dropping the ValueError in the constructor:  application code is responsible for constructing the `KeySet` instance which contains the `KeyRange`:  if it wants to include all values, there are easier ways to spell that, e.g. `KeySet(all_=True)`.

chemelnucfin:
@tseaver My main concern is that there is a difference in behavior if you don't specify either a start or an end key, it gets filled out automatically, and if you don't specify both start and end key it raises ValueError.  I think either both cases should raise ValueError, or both should get them filled out automatically.

tseaver:
Specifying a range with only one of `start_*` or `end_*` has well-defined semantics, which we have no other way to spell.  Specifying a range with not bounds at all is just a confusing / potentially expensive way to as for the same rows as `KeySet(all_=True)1`.

@lukesneeringer, @dhermes  Can one of you chime in?

dhermes:
ISTM @tseaver is right on here.

chemelnucfin:
I still somewhat disagree, since a `KeyRange` object shouldn't care what a `KeySet` does with it.  If anything, the check should be in `KeySet` to convert an empty `KeyRange` into `all_=True`.  The behavior within `KeyRange` is still inconsistent with not supplying missing keys.

However, I can understand your point that you shouldn't construct a plain `KeyRange` object.  

Anyway, my suggestion would be, either to document this behavior, or convert an empty `KeyRange` object into `all_=True` in `KeySet`.

chemelnucfin:
BTW, I did change the empty constructor back to the way it was.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -60,17 +60,19 @@ def to_pb(self):
         """
         kwargs = {}
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-
-        if self.start_closed:
+        elif self.start_closed is not None:
```
dhermes: I don't think `elif` is what you want here?

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -60,17 +60,19 @@ def to_pb(self):
...
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
+        else:
+            kwargs['start_closed'] = _make_list_value_pb([])
```
dhermes: It's unclear to me why'd you prefer an **explicitly** empty value over an implicitly empty one.

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: it's been so long, but I think elif is right.  I think I was guarding against the case for [] where neither self.start_open and self.start_closed would not be hit, and therefore passing to pb will complain that there was a missing key.

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: iirc, to_pb complained because I didn't supply with start keys if I only specified end_*

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: I don't think it matters in this case whether it's an if or an elif (since ValueError gets raised earlier), but elif makes it clear that it should either be a start_open, or a start_closed.

dhermes: There is a slight difference when using an explicitly empty **message field**:

```python
>>> from google.protobuf.struct_pb2 import ListValue
>>> from google.cloud.spanner_v1.proto.keys_pb2 import KeyRange
>>> k1 = KeyRange()
>>> k2 = KeyRange(
...     start_closed=ListValue(values=[]),
...     end_closed=ListValue(values=[]),
... )
>>> k1

>>> k2
start_closed {
}
end_closed {
}

>>>
```

As opposed to a simple field:

```python
>>> from google.cloud.spanner_v1.proto.keys_pb2 import KeySet
>>> k1 = KeySet()
>>> k2 = KeySet(all=False)
>>> k3 = KeySet(all=True)
>>> k1

>>> k2

>>> k3
all: true

>>>
```

spanner/google/cloud/spanner_v1/keyset.py
dhermes: Can you show me the complaint?

chemelnucfin:
Closing in favor of #4618 and #4623.  Maybe revisit later.

PR #4519, opened 2017-12-04
Pinning `django` test dependency to < 2.0 in Python 2.7.

dhermes:
See:
https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/

Python 2.7 support is explicitly removed.

### DIFF
--- a/logging/nox.py
+++ b/logging/nox.py
@@ -23,6 +23,14 @@ LOCAL_DEPS = (
     os.path.join('..', 'api_core'),
     os.path.join('..', 'core'),
 )
+UNIT_TEST_DEPS = (
+    'mock',
+    'pytest',
+    'pytest-cov',
+    'flask',
+    'webapp2',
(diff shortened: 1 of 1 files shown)


logging/nox.py
```diff
@@ -23,6 +23,14 @@
...
     os.path.join('..', 'core'),
 )
+UNIT_TEST_DEPS = (
+    'mock',
+    'pytest',
+    'pytest-cov',
+    'flask',
+    'webapp2',
+    'webob',
+)
```
tseaver: Ugh. FWIW, these can hardly be considered "unit" test dependenies.  Ideally, the code which provides integration for the various web frameworks wouldn't even be packaged together with `google.cloud.logging`.

dhermes: @tseaver I agree, though it's "optional" (but not listed as an extra)

PR #4521, opened 2017-12-04
Requiring latest release of concurrent futures backport.

dhermes:
See: https://github.com/agronholm/pythonfutures/issues/63

----

ASIDE: For the first time, the `.dev1` versions come in handy to specify the actual ways a package depends on **in development** vs. published.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -108,7 +108,7 @@ class Policy(base.BasePolicy):
         # Also maintain a request queue and an executor.
         if executor is None:
             executor_kwargs = {}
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, client, subscription, flow_control=types.FlowControl(),
         # Also maintain a request queue and an executor.
         if executor is None:
             executor_kwargs = {}
-            if sys.version_info >= (3, 6):
+            if sys.version_info[:2] == (2, 7) or sys.version_info >= (3, 6):
```
dhermes: I could also just use `six.PY2` since we **only** support Python 2.7.

PR #4522, opened 2017-12-04
Shutdown executor when closing Pub/Sub subscriber policy.

dhermes:


dhermes:
I have a feeling that the executor may be a source of a spike in CPU and/or memory usage, but I'm not entirely sure how.

PR #4525, opened 2017-12-04
WIP: Spanner: Read with range keys index tests

chemelnucfin:
@tseaver I actually prefer the format in #4411, but maybe this is what you want?  

Using CI as tester.


tseaver:
Check out the [short form example[(https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4411#discussion_r154791057) I posted in #4411.  I'd rather we just fix that PR there.

Issue #4530, opened 2017-12-05
Firestore DocumentReference.get NotFound error should be configurable


sourcec0de:
### Firestore 

`DocumentReference.get` throws an error instead of returning a `DocumentSnapshot` that has a property of `exists`.

This seems inconsistent with the other firebase firestore clients in that if you call `.get` it returns a snapshot where you can check the `exists` property rather than having to wrap it in a `try:catch`.

Is this something the library plans to support?

chemelnucfin:
Hi, I'm going to take a stab at this question.  I think that using `try except` is just as preferable or even more so than using `if else`.  From the standard docs 
https://docs.python.org/3.6/glossary.html#term-eafp

> EAFP
> Easier to ask for forgiveness than permission. This common Python coding style assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false. This clean and fast style is characterized by the presence of many try and except statements. The technique contrasts with the LBYL style common to many other languages such as C..
> 

Please feel free to reopen if you'd like to discuss further.

dhermes:
@sourcec0de We are not opposed to it, but as @chemelnucfin notes we felt the extra wrapping was unnecessary / not Pythonic. The current interface isn't something that we refuse to change though.

sourcec0de:
I want to say great work first and foremost.
Thank you for giving me an opportunity to just raise some questions. 

I realize the firebase client libraries are really designed to be idiomatic. However, after using some of the other SDK's with similar APIs it was a little unexpected to see an exception here.

I would agree that in python raising an exception is the standard idiom and `EAFP` is a good method to live by if the code you're executing is prone to `exceptions`. But if we know there is only a single possible error `NotFound` then it feels a little heavy handed to wrap in a `try:catch`.

`LBYL` feels like the better option here.

But I'm not proposing you change the default behavior, I'm simply asking if it makes sense to add a keyword argument to `ignore_notfound`

I would expect an example like this to work well if we simply returned a `document snapshot` that contained an empty dictionary and `exists: False`. Pythons `dictionary.get` allows me to set the default value without throwing an exception.

Having to introduce a `try:catch` also introduces a write really early on in a transaction that doesn't support read after write.

```python
@firestore.transactional
def debit_account(tx, account, amount):
    walletRef = wallets.document(account)
    walletSnap = walletRef.get(transaction=tx, ignore_notfound=True)
    wallet = walletSnap.to_dict()
[...]

PR #4531, opened 2017-12-05
Firestore: Allow document get to be error free

chemelnucfin:
@dhermes ?

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/document.py
+++ b/firestore/google/cloud/firestore_v1beta1/document.py
@@ -399,7 +399,7 @@ class DocumentReference(object):
 
         return commit_response.commit_time
 
-    def get(self, field_paths=None, transaction=None):
+    def get(self, field_paths=None, transaction=None, raise_error=True):
         """Retrieve a snapshot of the current document.
 
         See :meth:`~.firestore_v1beta1.client.Client.field_path` for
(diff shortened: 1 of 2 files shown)


dhermes:
@chemelnucfin I added a "do not merge" label since this is for discussion.

PS Don't forget to label PRs (i.e. "Firestore" here).

dhermes:
A few very high-level comments:

- I **very much** don't like the `raise_error` keyword. I would prefer always raise or never raise, not this "both are OK" approach
- It didn't occur to me how easy it'd be to return a non-existent `DocumentSnapshot`
- I'm still not sure it's broke, so "if it ain't broke, don't fix it", might still apply

I'd like to see what @jonparrott thinks as far as "idiomatic"-ness

jonparrott:
Agreed that it should be either or and not both, from zen of Python:

> There should be one-- and preferably only one --obvious way to do it.

This really just boils down to one question:

It is a common thing in Firestore to get an empty/nonexistent document and not care? If yes, then we should return a empty `DocumentSnapshot`, if no, then we should raise an exception.

dhermes:
@jonparrott For context #4530 is the issue where this is being discussed (see @chemelnucfin's branch name).

dhermes:
@chemelnucfin Just noticed "Declare target packages" in your build picked up all the packages. This means your PR is behind `master`. Builds will be much faster if you stay rebased against the latest version of `master`.

chemelnucfin:
Seems like there's not much interest in carrying this forward.

PR #4543, opened 2017-12-07
Storage: ProjectID Regex

chemelnucfin:
ProjectID description from https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects#Project.FIELDS.project_id

#4447

### DIFF
--- a/storage/google/cloud/storage/notification.py
+++ b/storage/google/cloud/storage/notification.py
@@ -31,10 +31,12 @@ _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
 _PROJECT_PATTERN = r'(?P<project>[a-z][a-z0-9-]{4,28}[a-z0-9])'
 _TOPIC_NAME_PATTERN = r'(?P<name>[A-Za-z](\w|[-_.~+%])+)'
 _TOPIC_REF_PATTERN = _TOPIC_REF_FMT.format(
-    _PROJECT_PATTERN.replace(r'\A', '').replace(r'\Z', ''),
-    _TOPIC_NAME_PATTERN
-)
+    _PROJECT_PATTERN, _TOPIC_NAME_PATTERN)
 _TOPIC_REF_RE = re.compile(_TOPIC_REF_PATTERN)
+_BAD_TOPIC = (
+    'Resource has invalid topic: {}; see '
+    'https://cloud.google.com/storage/docs/json_api/v1/'
+    'notifications/insert#topic')
 
 
 class BucketNotification(object):
@@ -112,25 +114,12 @@ class BucketNotification(object):
 
         :rtype: :class:`BucketNotification`
         :returns: the new notification instance
-        :raises ValueError:
-            if resource is missing 'topic' key, or if it is not formatted
-            per the spec documented in
-            https://cloud.google.com/storage/docs/json_api/v1/notifications/insert#topic
         """
         topic_path = resource.get('topic')
         if topic_path is None:
             raise ValueError('Resource has no topic')
 
-        match = _TOPIC_REF_RE.match(topic_path)
-        if match is None:
-            raise ValueError(
-                'Resource has invalid topic: {}; see {}'.format(
-                    topic_path,
-                    'https://cloud.google.com/storage/docs/json_api/v1/'
-                    'notifications/insert#topic'))
-
-        name = match.group('name')
-        project = match.group('project')
+        name, project = _parse_topic_path(topic_path)
         instance = cls(bucket, name, topic_project=project)
         instance._properties = resource
 
@@ -362,3 +351,40 @@ class BucketNotification(object):
             path=self.path,
             query_params=query_params,
         )
+
+
+def _parse_topic_path(topic_path):
+    """Verify that a topic path is in the correct format.
+
+    .. _resource manager docs: https://cloud.google.com/resource-manager/\
+                               reference/rest/v1beta1/projects#\
+                               Project.FIELDS.project_id
+    .. _topic spec: https://cloud.google.com/storage/docs/json_api/v1/\
+                    notifications/insert#topic
+
+    Expected to be of the form:
+
+        //pubsub.googleapis.com/projects/{project}/topics/{topic}
+
+    where the ``project`` value must be "6 to 30 lowercase letters, digits,
+    or hyphens. It must start with a letter. Trailing hyphens are prohibited."
+    (see `resource manager docs`_) and ``topic`` must have length at least two,
+    must start with a letter and may only contain alphanumeric characters or
+    ``-``, ``_``, ``.``, ``~``, ``+`` or ``%`` (i.e characters used for URL
+    encoding, see `topic spec`_).
+
+    Args:
+        topic_path (str): The topic path to be verified.
+
+    Returns:
+        Tuple[str, str]: The ``project`` and ``topic`` parsed from the
+        ``topic_path``.
+
+    Raises:
+        ValueError: If the topic path is invalid.
+    """
+    match = _TOPIC_REF_RE.match(topic_path)
+    if match is None:
+        raise ValueError(_BAD_TOPIC.format(topic_path))
+
+    return match.group('name'), match.group('project')
--- a/storage/google/cloud/storage/notification.py
+++ b/storage/google/cloud/storage/notification.py
@@ -28,10 +28,12 @@ JSON_API_V1_PAYLOAD_FORMAT = 'JSON_API_V1'
 NONE_PAYLOAD_FORMAT = 'NONE'
 
 _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
-_PROJECT_PATTERN = r'(?P<project>[a-z]+-[a-z]+-\d+)'
+_PROJECT_PATTERN = r'(?P<project>[a-z][a-z0-9-]{4,28}[a-z0-9])'
 _TOPIC_NAME_PATTERN = r'(?P<name>[A-Za-z](\w|[-_.~+%])+)'
 _TOPIC_REF_PATTERN = _TOPIC_REF_FMT.format(
-    _PROJECT_PATTERN, _TOPIC_NAME_PATTERN)
+    _PROJECT_PATTERN.replace(r'\A', '').replace(r'\Z', ''),
+    _TOPIC_NAME_PATTERN
+)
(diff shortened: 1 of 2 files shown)


chemelnucfin:
Hmm.  it is locally 100%.  Investigating

chemelnucfin:
Strange.  Let's see if #4545 fixes it.

storage/google/cloud/storage/notification.py
```diff
@@ -129,6 +129,8 @@ def from_api_repr(cls, resource, bucket):
 
         name = match.group('name')
         project = match.group('project')
+        if project and len(project) > 30:
```
dhermes: @chemelnucfin The length is **part** of the regex, why are you checking it?

storage/tests/unit/test_notification.py
```diff
@@ -16,6 +16,72 @@
 
 import mock
 
+class TestProjectID(unittest.TestCase):
+
+    from google.cloud.storage import notification
```
dhermes: 1. Please have **two** empty lines between top-level definitions (i.e. after `import mock` and before `class TestProjectID`)
1. This import will still happen at the time this module is imported, not at runtime, so please move it.

storage/tests/unit/test_notification.py
```diff
@@ -16,6 +16,72 @@
...
+    PROJECT = notification._PROJECT_PATTERN
+
+    def test_min_length_fail(self):
+        import re
```
dhermes: Imports from the standard library are OK to do at module-level (i.e. they don't conflict with our testing style, though you'll find cases where standard library modules **are** imported at runtime and you'll find cases where they aren't)

storage/tests/unit/test_notification.py
```diff
@@ -16,6 +16,72 @@
...
+        self.assertIsNotNone(project)
+
+    def test_capital_letters_fail(self):
+        import re
+        project = re.match(self.PROJECT, 'aaaAaa')
+        self.assertIsNone(project)
+
+    def test_leading_digit_fail(self):
+        import re
+        project = re.match(self.PROJECT, '1''a' * 5)
```
dhermes: 1. The string concat makes me do a double take, I'd rather see a `+` in between?
1. Why do the concat at all, it's short enough that you could use the literal `1aaaaa`

storage/google/cloud/storage/notification.py
chemelnucfin: I had a failing contrived test case:
'a' * 31, matched, but an invalid project id because it was too long.

chemelnucfin: I also tried word boundaries and such, but did not work combining with the topics.

storage/tests/unit/test_notification.py
chemelnucfin: you are right, I missed a '+'.
I guess just to be consistent with the other ones.  Either works, I guess.

storage/google/cloud/storage/notification.py
```diff
@@ -28,10 +28,10 @@
 NONE_PAYLOAD_FORMAT = 'NONE'
 
 _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
-_PROJECT_PATTERN = r'(?P<project>[a-z][a-z0-9-]{4,28}[a-z0-9])'
+_PROJECT_PATTERN = r'(?P<project>\A[a-z][a-z0-9-]{4,28}[a-z0-9]\Z)'
```
dhermes: @chemelnucfin Why not use `^` and `$`?

chemelnucfin: probably better in this case so it doesn't match \n?

dhermes: Which is probably better? Newlines won't get matched by default

chemelnucfin: but if someone accidentally wrote one?

chemelnucfin: I guess it's not that big a deal for me to change it.  But just putting my thoughts out there.

chemelnucfin: changed

storage/google/cloud/storage/notification.py
```diff
@@ -28,10 +28,12 @@
 NONE_PAYLOAD_FORMAT = 'NONE'
 
 _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
-_PROJECT_PATTERN = r'(?P<project>[a-z]+-[a-z]+-\d+)'
+_PROJECT_PATTERN = r'(?P<project>^[a-z][a-z0-9-]{4,28}[a-z0-9]$)'
 _TOPIC_NAME_PATTERN = r'(?P<name>[A-Za-z](\w|[-_.~+%])+)'
 _TOPIC_REF_PATTERN = _TOPIC_REF_FMT.format(
-    _PROJECT_PATTERN, _TOPIC_NAME_PATTERN)
+    _PROJECT_PATTERN.replace(r'^', '').replace(r'$', ''),
```
dhermes: @chemelnucfin These `replace` salls defeat the purpose. I'm confused as to why this is happening.

chemelnucfin: My reasoning was that PROJECT_PATTERN deals with the project only (which is why I chose \A\Z), which should be standalone.  TOPIC_REF_FMT should deal with something else differently, but PROJECT_PATTERN is inside the _TOPIC_REF_FMT.

iirc PROJECT_PATTERN is not used elsewhere now, but it could be in the future.

dhermes: Can you provide a concrete example of what this fixes / what would be broken without it?

chemelnucfin: I changed it back to \A \Z.  
The project pattern would match for test_max_length_fail without any line end delimiters (^$ or \A\Z)
The project pattern would match for test_newline_fail with ^$

storage/google/cloud/storage/notification.py
```diff
@@ -28,10 +28,12 @@
 NONE_PAYLOAD_FORMAT = 'NONE'
 
 _TOPIC_REF_FMT = '//pubsub.googleapis.com/projects/{}/topics/{}'
-_PROJECT_PATTERN = r'(?P<project>[a-z]+-[a-z]+-\d+)'
+_PROJECT_PATTERN = r'(?P<project>\A[a-z][a-z0-9-]{4,28}[a-z0-9]\Z)'
 _TOPIC_NAME_PATTERN = r'(?P<name>[A-Za-z](\w|[-_.~+%])+)'
 _TOPIC_REF_PATTERN = _TOPIC_REF_FMT.format(
-    _PROJECT_PATTERN, _TOPIC_NAME_PATTERN)
+    _PROJECT_PATTERN.replace(r'\A', '').replace(r'\Z', ''),
```
dhermes: OK @chemelnucfin so I've actually pulled out my editor and looked at this module wholesale. Final answer: just remove all mention of `\A` and `\Z` (they don't need to be replaced with `^` or `$` either).

Reasoning:

- `_PROJECT_PATTERN` is never a full reg-ex, it's a fragment intended to be used with `_TOPIC_REF_PATTERN`
- It "makes" sense to re-use `_TOPIC_REF_FMT` in this way because `_TOPIC_REF_FMT` is needed in `BucketNotification.create`

I have never seen any code that actually uses `\A` and `\Z`, likely because `\A` and `\Z` are only relevant (as alternatives to `^` and `$`) in multiline mode, and when in multiline mode people don't usually think of "beginning" and "end".

storage/tests/unit/test_notification.py
```diff
@@ -12,11 +12,92 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 import unittest
 
 import mock
 
 
+class TestProjectID(unittest.TestCase):
```
dhermes: This class name should correspond to the thing being tested (this is why we have the `_call_function_under_test` pattern, etc.)

So I don't in principle have any issue with these tests, but they should either

- All be tests for `BucketNotification.from_api_repr`
- Be tests for an as-yet-undefined helper function which verifies a `topic_path`

----

Note that I'm also suggesting here that we **should not** be testing a private constant that is only used as a helper for another private constant. We should at most be testing for the **behavior** that we want those constants to exhibit in our actual code.

Grokking these tests, I understand why you worried about having `\A` and `\Z` on the constant, but that is the tail wagging the dog (i.e. having code grow an "appendage" that is unused, just for the sake of testing).

storage/tests/unit/test_notification.py
```diff
@@ -12,11 +12,92 @@
...
+        pattern = notification._PROJECT_PATTERN
+        project = re.match(pattern, 'a-bcdefghijklmn12opqrstuv0wxyz')
+        self.assertIsNotNone(project)
+
+    def test_newline_fail(self):
+        from google.cloud.storage import notification
+        pattern = notification._PROJECT_PATTERN
+        project = re.match(pattern, 'z0123456789\nmalicious', re.M)
+        self.assertIsNone(project)
+
```
dhermes: Style note, from [PEP8][1]:

> Surround top-level function and class definitions with two blank lines.

[1]: https://www.python.org/dev/peps/pep-0008/#blank-lines

storage/google/cloud/storage/notification.py
chemelnucfin: @dhermes I would suggest again to leave the code as is if there is nothing wrong with it.  In this case, it might be better safe than sorry rather than having "extra" code, especially since someone may accidentally use PROJECT_PATTERN.

dhermes:
@chemelnucfin I did a small re-factor to push this over the finish line.

Some things worth noting:

- Setting `self.TOPIC_REF` in those test cases will leave lingering shared state, which we avoid as much as we can (at least in unit tests)
- Since the number of `from_api_repr` tests overwhelmed the rest, I feel it is much easier to read by putting it in a standalone helper (note how much shorter each test case became)
- We should probably use `@pytest.mark.parametrize` here, but that's for another day, another PR (I collapsed all of the "success" cases into one)

chemelnucfin:
@dhermes 
1.  Hmm. I guess I was just using what was there before.  My bad for not paying attention to assigning to self.TOPIC_REF

2, 3.  ok.  Sometimes I'm hesitant to refactor for fear of breaking things.  

So is it ok to merge?

dhermes:
@chemelnucfin Yes go ahead and merge (make sure to squash).

PR #4544, opened 2017-12-07
Storage: Fix coverage issue

chemelnucfin:
Not sure if this is "desired", but it does make the coverage to be 100%.

### DIFF
--- a/storage/tests/unit/test__signing.py
+++ b/storage/tests/unit/test__signing.py
@@ -43,10 +43,12 @@ class Test_get_expiration_seconds(unittest.TestCase):
         self.assertEqual(self._call_fut(123), 123)
 
     def test_w_long(self):
-        if six.PY3:
-            raise unittest.SkipTest('No long on Python 3')
-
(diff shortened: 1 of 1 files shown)


dhermes:
@chemelnucfin This is not needed. `nox -s cover` combines coverage from the `nox -s unit` runs for **ALL** versions, so it will hit both branches.

chemelnucfin:
But when you test just a single version, you will wonder what you are missing.

dhermes:
That's not the point?

chemelnucfin:
But it does require extra time to confirm that it wasn't what you did that lowered the coverage, right?

dhermes:
@chemelnucfin I don't follow. Run `nox -s unit cover`?

chemelnucfin:
Is there something wrong with the test though?  It does fix the coverage issue.

dhermes:
> Is there something wrong with the test though?

It's the tail wagging the dog. We (our code) shouldn't work around our tools / language versions, they should work around us.

> It does fix the coverage issue.

There is no coverage issue to fix.

PR #4546, opened 2017-12-07
Storage: support 'sourceGeneration' in 'copy_blob'

### LINKED ISSUE #4533: Storage: sourceGeneration in copy_blob
Hi,
I need that the copy_blob function of Bucket object in Cloud Storage library manage the "sourceGeneration" parameter to copy the archived version of an object.

Thanks

davidebelloni:
Closes #4533.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -678,6 +678,10 @@ class Bucket(_PropertyMixin):
         :param preserve_acl: Optional. Copies ACL from old blob to new blob.
                              Default: True.
 
+        :type source_generation: long
+        :param source_generation: Optional. The generation of the blob to be copied.
+                                    Default: None.
+
         :rtype: :class:`google.cloud.storage.blob.Blob`
         :returns: The new Blob.
         """
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -679,8 +679,8 @@ class Bucket(_PropertyMixin):
                              Default: True.
 
         :type source_generation: long
-        :param source_generation: Optional. The generation of the blob to be copied.
-                                    Default: None.
+        :param source_generation: Optional. The generation of the blob to be
+                                  copied.
 
         :rtype: :class:`google.cloud.storage.blob.Blob`
         :returns: The new Blob.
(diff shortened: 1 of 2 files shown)


chemelnucfin:
Hello!  Thank you for your contribution.  Tests are also needed when you change source code.

Also, I thought about this problem.  If source_generation was to be put into copy_blob, would other params also go into copy_blob?  Also, the other functions such as delete_blob, etc. might need an update as well.

I will also continue to think about the problem, but feel free to finish this PR by adding tests.

Any additional comments @dhermes @tseaver ?

chemelnucfin:
I also just noticed that you made the PR using the master branch.  A suggestion of using git is to make PRs on a separate branch, so that you can continue to do work and not be worried about automatic updates.  If you commit and push on the master branch while doing some other work, this PR will automatically update with this other work.

You can use `git checkout -b newbranchname` to make a new branch.

davidebelloni:
Hi,
I think to have added tests. Your question about others params and functions are correct, but for my goal should be enough this commits. I'm trying to "restore" all the bucket's objects, live and versioned, with a specific metadata.

I'm not so expert about git.  About PR, what you mean is to make a branch on my repo fork about this feature and do the PR relative to this branch? I'm not an active developer on this repo.

Thanks

chemelnucfin:
@dhermes @tseaver Any comments regarding the other possible kwargs or other functions that could use these possible kwargs?

@davidebelloni Yes, git is confusing.  I understand that this might be your single pull request, but I just wanted to let you know in case you push to the master branch again, this PR would have automatically update with that commit, which might not be what you want.  It's not a big deal for us, but it might make future PRs easier for you to work with.

davidebelloni:
Hi @chemelnucfin ,
you are right about git, but in this case there's no problem.

Are you waiting for something from me to proceed with this PR?

Thanks

chemelnucfin:
@davidebelloni Hi, I don't see anything that immediately jumps out at me, but since I'm relatively new here I would love to hear @dhermes or @tseaver comments also.  Thanks for the PR.

tseaver:
@davidebelloni Thank you for the patch!

davidebelloni:
Hi @tseaver ,
when you think that this code will be public in 1.7.0 google-cloud-storage version?

tseaver:
@jonparrott, @frankyn Is there any reason we need to delay a 1.7.0 release?

jonparrott:
go for it

PR #4554, opened 2017-12-07
Checking _Rendezvous.done() when stopping Pub / Sub request generator.

dhermes:


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -237,12 +237,23 @@ def _stop_request_generator(self, request_generator):
         Args:
             request_generator (Generator): A streaming pull request generator
                 returned from :meth:`_request_generator_thread`.
+            response_generator (grpc._channel._Rendezvous): The gRPC
```
jonparrott: The type you're looking for is `grpc.Call`.

dhermes: I can change it if you want, but I feel like that's me yielding to the insanity that is `_Rendezvous`, so I am not keen on it.

jonparrott: I'd really prefer it, and actually, to add to the madness this is actually [`grpc.Future`](https://grpc.io/grpc/python/grpc.html#grpc.Future).

dhermes:
Updated docstring (no code changes). Merging without CI (since no code changes).

PR #4556, opened 2017-12-08
Spanner: clean up databases after use

chemelnucfin:


tseaver:
This change is not correct.  We are forced to re-use / share instances across test runs (and for other purposes), which means that we should only ever drop databases which are created in the *current* test run.

chemelnucfin:
But the way the tests are set up, it doesn't seem like all the tests get
cleaned up, especially if I interrupt in the middle?

dhermes:
> especially if I interrupt in the middle

Our cleanup has been tailored to a "graceful" exit (i.e. tests may error, but user doesn't kill process).

We could consider a more resilient approach but I don't think it's worth the effort.

tseaver:
Until we go back to using a created-from-scratch instance, we *cannot* drop all databases, because the CI may be running more than one job at once in the same instance.  There may also be other databases pre-populated (e.g., to run the "large chunk" tests) which should never be dropped.

PR #4557, opened 2017-12-08
Spanner: Database not found

### LINKED ISSUE #4071: Cloud Spanner client incorrectly handles NOT_FOUND error returned by create database
While creating a database, is the backend returns NOT_FOUND error, the client assumes it to indicate that instance is not found and thus returns "Instance not found" error message:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/google/cloud/spanner_v1/database.py#L213

This is not always the case as cloud spanner can return NOT_FOUND in other scenarios. For eg in the "CREATE INDEX ... " ddl statement that table that the index is being created on does not exist it would return a NOT_FOUND.  I think the fix would be to just pass the error message returned by the rpc to the user.

This can be reproduced using:
```python
from google.cloud import spanner

"""Creates a database and tables for pattern 1 sample data."""

spanner_client = spanner.Client()

instance = spanner_client.instance("test-instance")



database = instance.database("bitemporal_pattern1", ddl_statements=[

    """CREATE TABLE MyTable (

        Id        String(36) NOT NULL,

        Field1  STRING(36) NOT NULL

) PRIMARY KEY (Id)"""

,

"""CREATE INDEX IDX ON yTable (Field1)"""

])



operation = database.create()



print('Waiting for operation to complete...')

operation.result()
```

chemelnucfin:
Closes #4071

### DIFF
--- a/spanner/google/cloud/spanner_v1/database.py
+++ b/spanner/google/cloud/spanner_v1/database.py
@@ -17,20 +17,20 @@
 import re
 import threading
 
-import google.auth.credentials
 from google.api_core import exceptions
+import google.auth.credentials
+from google.cloud.exceptions import Conflict
+from google.cloud.exceptions import NotFound
 from google.gax.errors import GaxError
 from google.gax.grpc import exc_to_code
-from google.cloud.spanner_v1.gapic.spanner_client import SpannerClient
 from grpc import StatusCode
 import six
 
 # pylint: disable=ungrouped-imports
-from google.cloud.exceptions import Conflict
-from google.cloud.exceptions import NotFound
 from google.cloud.spanner_v1 import __version__
 from google.cloud.spanner_v1._helpers import _options_with_prefix
 from google.cloud.spanner_v1.batch import Batch
+from google.cloud.spanner_v1.gapic.spanner_client import SpannerClient
 from google.cloud.spanner_v1.pool import BurstyPool
 from google.cloud.spanner_v1.pool import SessionCheckout
 from google.cloud.spanner_v1.session import Session
--- a/spanner/google/cloud/spanner_v1/database.py
+++ b/spanner/google/cloud/spanner_v1/database.py
@@ -18,6 +18,7 @@ import re
 import threading
 
 import google.auth.credentials
+from google.api_core import exceptions
 from google.gax.errors import GaxError
 from google.gax.grpc import exc_to_code
 from google.cloud.spanner_v1.gapic.spanner_client import SpannerClient
@@ -208,14 +209,7 @@ class Database(object):
                 options=options,
             )
         except GaxError as exc:
-            if exc_to_code(exc.cause) == StatusCode.ALREADY_EXISTS:
-                raise Conflict(self.name)
-            elif exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
-                raise NotFound('Instance not found: {name}'.format(
-                    name=self._instance.name,
-                ))
-            raise
-
+            raise exceptions.from_grpc_error(exc.cause)
         return future
 
     def exists(self):
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -36,6 +36,7 @@ from grpc import StatusCode
 
 from google.cloud._helpers import UTC
 from google.cloud.exceptions import GrpcRendezvous
+from google.cloud.exceptions import NotFound
 from google.cloud.spanner_v1._helpers import TimestampWithNanoseconds
 from google.cloud.spanner import Client
 from google.cloud.spanner import KeyRange
@@ -281,6 +282,35 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
             database.database_id
             for database in Config.INSTANCE.list_databases()]
         self.assertIn(temp_db_id, database_ids)
+
+    def test_table_not_found(self):
+        temp_db_id = 'temp_db' + unique_resource_id('_')
+
+        correct_table = 'MyTable'
+        incorrect_table = 'NotMyTable'
+        self.assertNotEqual(correct_table, incorrect_table)
+
+        create_table = (
+            'CREATE TABLE {} (\n'
+            '    Id      STRING(36) NOT NULL,\n'
+            '    Field1  STRING(36) NOT NULL\n'
+            ') PRIMARY KEY (Id)').format(correct_table)
+        index = 'CREATE INDEX IDX ON {} (Field1)'.format(incorrect_table)
+
+        temp_db = Config.INSTANCE.database(
+            temp_db_id,
+            ddl_statements=[
+                create_table,
+                index,
+            ],
+        )
+        try:
+            temp_db.create()
+        except NotFound as exc:
+            self.assertEqual(exc.args[0],
+                             'Table not found: {0}'.format(incorrect_table))
+        except:
+            self.fail()
 
     def test_update_database_ddl(self):
         pool = BurstyPool()
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -304,13 +304,11 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
                 index,
             ],
         )
-        try:
+        with self.assertRaises(NotFound) as exc_info:
             temp_db.create()
-        except NotFound as exc:
-            self.assertEqual(exc.args[0],
-                             'Table not found: {0}'.format(incorrect_table))
-        except:
-            self.fail()
+
+        expected = 'Table not found: {0}'.format(incorrect_table)
+        self.assertEqual(exc_info.exception.args, (expected,))
 
     def test_update_database_ddl(self):
         pool = BurstyPool()
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -304,6 +304,7 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
                 index,
             ],
         )
+        self.to_delete.append(temp_db)
         with self.assertRaises(NotFound) as exc_info:
             temp_db.create()
 
--- a/spanner/tests/unit/test_database.py
+++ b/spanner/tests/unit/test_database.py
@@ -284,7 +284,7 @@ class TestDatabase(_BaseTest):
         self.assertNotEqual(database1, database2)
 
     def test_create_grpc_error(self):
-        from google.gax.errors import GaxError
+        from google.api_core.exceptions import GoogleAPICallError
 
         client = _Client()
         api = client.database_admin_api = _FauxDatabaseAdminAPI(
@@ -293,7 +293,7 @@ class TestDatabase(_BaseTest):
         pool = _Pool()
         database = self._make_one(self.DATABASE_ID, instance, pool=pool)
 
-        with self.assertRaises(GaxError):
+        with self.assertRaises(GoogleAPICallError):
             database.create()
 
         (parent, create_statement, extra_statements,


chemelnucfin:
@dhermes PTAL on this, it closes one of vikas' issues.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -211,8 +211,8 @@ def create(self):
             if exc_to_code(exc.cause) == StatusCode.ALREADY_EXISTS:
                 raise Conflict(self.name)
             elif exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
-                raise NotFound('Instance not found: {name}'.format(
-                    name=self._instance.name,
+                raise NotFound('Database not found: {name}'.format(
```
dhermes: @chemelnucfin 

1. "Instance not found" is the correct error message here (the database is what is being created, so of course it wouldn't be found, it's the parent which might be "not found") You could add more detail like, "Instance {} not found when trying to create database {}".
1. Any reason not to use `exc.cause.details()`?

spanner/tests/system/test_system.py
```diff
@@ -282,6 +283,18 @@ def test_create_database(self):
...
         self.assertIn(temp_db_id, database_ids)
 
+    def test_create_database_not_found(self):
+        temp_db_id = 'temp_db' + unique_resource_id('_')
+        temp_db = Config.INSTANCE.database("bitemporal_pattern1", ddl_statements=[
+            """CREATE TABLE MyTable (
+                Id        String(36) NOT NULL,
+                Field1  STRING(36) NOT NULL
+            ) PRIMARY KEY (Id)""",
+            """CREATE INDEX IDX ON yTable (Field1)"""
```
dhermes: 1. Did you mean `MyTable` rather than `yTable`
1. Please only create triple-quote strings at module level (you end up with unintended indents)
1. Please use single quotes for `bitemporal_pattern1`
1. What is a `bitemporal_pattern1`? Maybe use a "normal" name?

spanner/tests/system/test_system.py
```diff
@@ -282,6 +283,18 @@ def test_create_database(self):
...
+        ])
+        with self.assertRaisesRegexp(NotFound, "Database not found"):
```
dhermes: 1. Please use single quotes
1. Any reason not to just check the entire message?

spanner/google/cloud/spanner_v1/database.py
dhermes: **UPDATE**: I see the example in #4071.

The **issue** is that the **second** DDL statement refers to `yTable` rather than `MyTable` (this seems intentional)?

Thus, returning `exc.cause.details()` seems to be the only right answer here.

In fact `google.api_core.exceptions.from_grpc_error` is made for exactly this case.

spanner/tests/system/test_system.py
chemelnucfin: ok, I just grabbed the case from vikas' example.

spanner/google/cloud/spanner_v1/database.py
chemelnucfin: Everywhere else in the code uses exc_to_code(exc.cause).  I think I will leave it as is.

spanner/tests/system/test_system.py
chemelnucfin: My reasoning was that "Database not found" was the crucial part.  The message might fluctuate a little?

spanner/tests/system/test_system.py
dhermes: @chemelnucfin I see that now. It's up to us to translate feedback from issues into the codebase (and that usually means changing some parts).

spanner/google/cloud/spanner_v1/database.py
dhermes: @chemelnucfin That's because this code pre-dates the `api_core` package. Please change it.

chemelnucfin: Can I do it in a separate PR?  There's a lot of other code.

dhermes: The whole point of this PR is to make the exception correct. The correct way to do it is to use the facility provided by `api_core`

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -208,14 +209,12 @@ def create(self):
                 options=options,
             )
         except GaxError as exc:
-            if exc_to_code(exc.cause) == StatusCode.ALREADY_EXISTS:
+            exception = exceptions.from_grpc_error(exc.cause)
+            if exception.grpc_status_code == StatusCode.ALREADY_EXISTS:
```
dhermes: @chemelnucfin Just `raise exception` here

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
...
+        table = 'yTable'
+        temp_db_id = 'temp_db' + unique_resource_id('_')        
+        temp_db = Config.INSTANCE.database(temp_db_id, ddl_statements=[
+            'CREATE TABLE MyTable ('
+            '    Id      STRING(36) NOT NULL,'
+            '    Field1  STRING(36) NOT NULL'
+            ') PRIMARY KEY (Id)',
+            'CREATE INDEX IDX ON {table} (Field1)'.format(table)
+        ])
+        with self.assertRaisesRegexp(GrpcRendezvous,
```
dhermes: This won't be a `GrpcRendezvous`, it should be one of the exceptions from `api_core.exception`. Maybe I'm confused?

Please push your branch to the `GoogleCloudPlatform` remote so the system tests actually run.

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
             for database in Config.INSTANCE.list_databases()]
         self.assertIn(temp_db_id, database_ids)
 
+    def test_table_not_found(self):
+        table = 'yTable'
+        temp_db_id = 'temp_db' + unique_resource_id('_')        
+        temp_db = Config.INSTANCE.database(temp_db_id, ddl_statements=[
```
dhermes: Suggest:

```python
    temp_db_id = 'temp_db' + unique_resource_id('_')

    table_name1 = 'MyTable'
    table_name2 = 'NotMyTable'
    self.assertNotEqual(table_name1, table_name2)

    create_table = (
        'CREATE TABLE {} (\n'
        '    Id      STRING(36) NOT NULL,\n'
        '    Field1  STRING(36) NOT NULL\n'
        ') PRIMARY KEY (Id)')
    create_table = create_table.format(table_name1)
    create_index = 'CREATE INDEX IDX ON {} (Field1)'.format(table_name2)

        temp_db_id,
        ddl_statements=[
            create_table,
            create_index,
        ],
    )
```

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
...
+            'CREATE TABLE MyTable ('
+            '    Id      STRING(36) NOT NULL,'
+            '    Field1  STRING(36) NOT NULL'
+            ') PRIMARY KEY (Id)',
+            'CREATE INDEX IDX ON {table} (Field1)'.format(table)
+        ])
+        with self.assertRaisesRegexp(GrpcRendezvous,
+                                     'Table not found: {table}'.format(table)):
+            operation = temp_db.create()
```
dhermes: As in the other PR, assigning to `operation` is "bad form" (since that variable can't actually receive the value due to the exception)

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
             for database in Config.INSTANCE.list_databases()]
         self.assertIn(temp_db_id, database_ids)
 
+    def test_table_not_found(self):
```
dhermes: It seems that the first DDL ("CREATE TABLE") will succeed, so this test should have a "clean up" that deletes the table.

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
...
+        table = 'yTable'
+        temp_db_id = 'temp_db' + unique_resource_id('_')        
+        temp_db = Config.INSTANCE.database(temp_db_id, ddl_statements=[
+            'CREATE TABLE MyTable ('
+            '    Id      STRING(36) NOT NULL,'
+            '    Field1  STRING(36) NOT NULL'
+            ') PRIMARY KEY (Id)',
+            'CREATE INDEX IDX ON {table} (Field1)'.format(table)
```
dhermes: This is invalid:

```python
>>> table = 'yTable'
>>> 'CREATE INDEX IDX ON {table} (Field1)'.format(table)
KeyError: 'table'
```

Seems maybe you didn't have a chance to run this code?

spanner/tests/system/test_system.py
```diff
@@ -282,6 +282,20 @@ def test_create_database(self):
...
+        ])
+        with self.assertRaisesRegexp(GrpcRendezvous,
+                                     'Table not found: {table}'.format(table)):
```
dhermes: Rather than use a regexp, can we just assert the entire error message? (Unless it's got "temporary" data in there like some opaque ID.)

spanner/tests/system/test_system.py
chemelnucfin: that is weird, I really don't know how that (Field1) got in there...

chemelnucfin: I actually did run the system test locally, but somehow I guess I accidentally changed the code...

chemelnucfin: never mind.  wrong error.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -208,14 +209,11 @@ def create(self):
...
-            if exc_to_code(exc.cause) == StatusCode.ALREADY_EXISTS:
-                raise Conflict(self.name)
-            elif exc_to_code(exc.cause) == StatusCode.NOT_FOUND:
-                raise NotFound('Instance not found: {name}'.format(
-                    name=self._instance.name,
-                ))
+            exception = exceptions.from_grpc_error(exc.cause)
+            if (exception.grpc_status_code == StatusCode.ALREADY_EXISTS
+                    or exception.grpc_status_code == StatusCode.NOT_FOUND):
+                raise exception
```
dhermes: 1. I don't see any issue just raising it always (I realize the unit tests will have a branch for that solo `raise` below, but we can just update the unit tests).
1. You should use an `isinstance()` check there (rather than equality against `exception.grpc_status_code`)

spanner/tests/system/test_system.py
```diff
@@ -282,6 +283,33 @@ def test_create_database(self):
...
+        temp_db = Config.INSTANCE.database(
+            temp_db_id,
+            ddl_statements=[
+                create_table,
+                create_index,
+            ],
+        )
+        with self.assertRaisesRegexp(NotFound,
+                                     '404 Table not found: '
+                                     + table_name2):
```
dhermes: @chemelnucfin Please ditch the `assertRaisesRegexp` and just check the entire `args` of the caught exception (unless there is some strange opaque ID).

(See my previous [comment][1])

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4557#discussion_r157915661

spanner/tests/system/test_system.py
dhermes: screenshot from 2017-12-20 09-18-20

I'm going to send a commit that does this cleanup ASAP.

PR #4558, opened 2017-12-08
Adding pause/resume to Pub / Sub consumer.

dhermes:
Using these (rather then open/close on the subscription Policy) when the flow control signals the message load is too great.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -179,6 +179,7 @@ class Consumer(object):
     def __init__(self):
         self._request_queue = queue.Queue()
         self.stopped = threading.Event()
+        self._can_consume = threading.Event()
         self._put_lock = threading.Lock()
         self._consumer_thread = None
 
@@ -319,8 +320,10 @@ class Consumer(object):
 
             request_generator = self._request_generator_thread(policy)
             response_generator = policy.call_rpc(request_generator)
+            responses = _pausable_iterator(
+                response_generator, self._can_consume)
             try:
-                for response in response_generator:
+                for response in responses:
                     _LOGGER.debug('Received response:\n%r', response)
                     policy.on_response(response)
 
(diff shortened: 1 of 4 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -392,3 +425,27 @@ def stop_consuming(self):
         """
         thread = self._stop_no_join()
         thread.join()
+
+
+def _pausable_iterator(iterator, can_continue):
```
dhermes: I toyed with a [`Pausable`][1] class but I think it makes more sense to store the `threading.Event` for `can_continue` / `can_consume` on the `Consumer` instance.

[1]: https://gist.github.com/dhermes/7aa696229833ead1a414f87697d82b46

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -291,7 +291,7 @@ def lease(self, ack_id, byte_size):
         # If we do, we need to stop the stream.
         if self._load >= 1.0:
             self._paused = True
-            self.close()
+            self._consumer.pause()
```
dhermes: @lukesneeringer Should this also halt the thread running `maintain_leases`?

The whole point of `pause` is that we want to stop our `QueueCallbackWorker` from receiving new "dispatch" instructions because it will convert those into more requests to send (and we need to clear the unsent).

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
jonparrott: I'm okay with less classes.

dhermes: Yeah, objects seem super cumbersome when we have so many related but independent workers.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: No, it is important that leases continue to be maintained.

PR #4564, opened 2017-12-11
Add get_cell() and get_cells() methods to PartialRowData

zakons:
Simple method to get a cell value. If no index supplied, get the most recent value.

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -181,12 +181,43 @@ class PartialRowData(object):
         """
         return self._row_key
 
+    def _get_cells_no_copy(self, column_family_id, column):
+        """Get a time series of cells stored on this instance.
+
+        Args:
+            column_family_id (str): The ID of the column family. Must be of the
+                form ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
+            column (bytes): The column within the column family where the cells
+                are located.
+
+        Returns:
+            List[~google.cloud.bigtable.row_data.Cell]: The cells stored in the
+            specified column.
+
+        Raises:
+            KeyError: If ``column_family_id`` is not among the cells stored
+                in this row.
+            KeyError: If ``column`` is not among the cells stored in this row
+                for the given ``column_family_id``.
+        """
+        try:
+            column_family = self._cells[column_family_id]
+        except KeyError:
+            raise KeyError(_MISSING_COLUMN_FAMILY.format(column_family_id))
+
+        try:
+            cells = column_family[column]
+        except KeyError:
+            raise KeyError(_MISSING_COLUMN.format(column, column_family_id))
+
+        return cells
+
     def get_cell(self, column_family_id, column, index=0):
         """Get a single cell stored on this instance.
 
         .. note::
 
-            This returns a copy of the actual ``Cell`` (so that the
+            This returns a copy of the actual cell (so that the
             caller cannot mutate internal state).
 
         Args:
@@ -210,15 +241,7 @@ class PartialRowData(object):
                 in this row for the given ``column_family_id``, ``column``
                 pair.
         """
-        try:
-            column_family = self._cells[column_family_id]
-        except KeyError:
-            raise KeyError(_MISSING_COLUMN_FAMILY.format(column_family_id))
-
-        try:
-            cells = column_family[column]
-        except KeyError:
-            raise KeyError(_MISSING_COLUMN.format(column, column_family_id))
+        cells = self._get_cells_no_copy(column_family_id, column)
 
         try:
             cell = cells[index]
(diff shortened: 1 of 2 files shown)


zakons:
Signed the CLA!

zakons:
Added an additional email address for CLA.

chemelnucfin:
Thank you for your contribution.  We require corresponding tests when you change source code.  I was working on something similar and was trying to figure out the unit test but I had the system test working if you want to compare.

Also, @dhermes maybe value would be good enough for the function name?

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -171,6 +171,24 @@ def row_key(self):
...
+        :param column_family_id: The ID of the column family. Must be of the
+                                 form ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
+        :type column: bytes
+        :param column: The column within the column family where the cell
+                       is located.
+
+        :type index: int
+        :param index: The offset in the series of values, default = 0
+
+        :rtype:  bytes or :class:`int`
```
chemelnucfin: should this be bytes only?

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -171,6 +171,24 @@ def row_key(self):
         """
         return self._row_key
 
+    def cell_value(self, column_family_id, column, index=0):
+        """Get a cell value
+
+        :type column_family_id: str
+        :param column_family_id: The ID of the column family. Must be of the
+                                 form ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
+        :type column: bytes
```
chemelnucfin: \n here?

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -171,6 +171,24 @@ def row_key(self):
         """
         return self._row_key
 
+    def cell_value(self, column_family_id, column, index=0):
+        """Get a cell value
+
+        :type column_family_id: str
+        :param column_family_id: The ID of the column family. Must be of the
+                                 form ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
```
chemelnucfin: \n here?

chemelnucfin:
@dhermes I think it LGTM to me after my comments, but I'd like you to look it over.

bigtable/google/cloud/bigtable/row_data.py
zakons: This documentation was copied verbatim from method set_cell on the different row classes in row.py.  Any changes here should be reviewed in the context of all the documentation.

Thanks!

chemelnucfin: Hello @zakons, my apologies.  I'm relatively new here and I didn't mean to be offensive.  I wasn't sure of the answers myself and therefore phrased it as a question.  Sorry if it seemed offensive.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -171,6 +181,105 @@ def row_key(self):
...
+
+        try:
+            cell = cells[index]
+        except (TypeError, IndexError):
+            num_cells = len(cells)
+            msg = _MISSING_INDEX.format(
+                index, column, column_family_id, num_cells)
+            raise IndexError(msg)
+
+        return copy.deepcopy(cell)
```
dhermes: @lukesneeringer Can you weigh in here on the `deepcopy`? I think I'd rather just return the `Cell` or at least provide a keyword `copy=True` that users can disable?

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -171,6 +181,105 @@ def row_key(self):
...
+            specified column.
+
+        Raises:
+            KeyError: If ``column_family_id`` is not among the cells stored
+                in this row.
+            KeyError: If ``column`` is not among the cells stored in this row
+                for the given ``column_family_id``.
+        """
+        cells = self._get_cells_no_copy(column_family_id, column)
+        return copy.deepcopy(cells)
```
dhermes: @lukesneeringer Ditto here, I think I'd like to copy the list containing the cells, but not copy the `Cell`-s themselves. WDYT?

bigtable/google/cloud/bigtable/row_data.py
chemelnucfin: I thought of having another function to return just the value.

dhermes: The user can just call `cell.value` and I think having two nearly identical methods would be overly "wordy" for the API. (Happy to be overruled though.)

bigtable/tests/unit/test_row_data.py
```diff
@@ -174,6 +177,138 @@ def test_to_dict(self):
...
+        partial_row_data = self._make_one(None)
+        self.assertEqual(partial_row_data._cells, {})
+
+        with self.assertRaises(KeyError) as exc_info:
+            partial_row_data.get_cells(family_name, None)
+
+        expected_arg = row_data._MISSING_COLUMN_FAMILY.format(family_name)
+        self.assertEqual(exc_info.exception.args, (expected_arg,))
+
+    def test_get_cell_bad_column(self):
```
dhermes: Oops, this was a copy-pasta error by me. Needs to be `test_get_cells_bad_column`

bigtable/google/cloud/bigtable/row_data.py
zakons: Actually, I agree it should be bytes only, but as I said, I copied the documentation for consistency.  I would encourage it be corrected and made consistent if this is correct. I will follow up on this.

dhermes: @zakons The distinction as that `set_cell()` will accept an integer and then encode the integer as a big-endian `long long` (`>q`) using the [`struct` library][1].

[1]: https://docs.python.org/3/library/struct.html

sduskis:
What do we need to move this forward?

zakons:
I will take some steps over the weekend to move this forward, now that the row iterator is merged.
    On Friday, January 26, 2018, 8:55:03 AM EST, Solomon Duskis <notifications@github.com> wrote:  
 
 
What do we need to move this forward?

zakons:
Superseded by PR #4908.

PR #4434, opened 2017-11-22
Spanner: Tests read and write back arrays with None and non None values

chemelnucfin:

PR #4569, opened 2017-12-11
Spanner: invalid type test

chemelnucfin:
This got closed inadvertently in #4434 
Breaking this up for integration 7 test.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -1235,6 +1235,31 @@ class TestSessionAPI(unittest.TestCase, _TestData):
                 [[['a', 1], ['b', 2]]],
             ])
 
+    def test_invalid_type(self):
+        table = 'counters'
+        columns = ('name', 'value')
+        session = self._db.session()
+        session.create()
+        self.to_delete.append(session)
+
+        valid_input = (('', 0),)
+        with session.batch() as batch:
+            batch.delete(table, self.ALL)
+            batch.insert(table, columns, valid_input)
+
+        invalid_input = ((0, ''),)
+        with self.assertRaises(errors.RetryError) as exc_info:
+            with session.batch() as batch:
+                batch.delete(table, self.ALL)
+                batch.insert(table, columns, invalid_input)
+
+        cause = exc_info.exception.cause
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -1052,6 +1052,23 @@ def test_execute_sql_returning_array_of_struct(self):
...
+        session = self._db.session()
+        session.create()
+        self.to_delete.append(session)
+        wrong = ((0, ''),)
+        right = (('', 0),)
+        error = 'StatusCode.FAILED_PRECONDITION, Invalid value for column'
+        with session.batch() as batch:
+            batch.delete(table, self.ALL)
+            batch.insert(table, columns, right)
+        with self.assertRaisesRegexp(errors.RetryError, error):
```
dhermes: Suggest instead:

```python
    def test_invalid_type(self):
        table = 'counters'
        columns = ('name', 'value')
        session = self._db.session()
        session.create()
        self.to_delete.append(session)

        valid_input = (('', 0),)
        with session.batch() as batch:
            batch.delete(table, self.ALL)
            batch.insert(table, columns, valid_input)

        invalid_input = ((0, ''),)
        with self.assertRaises(errors.RetryError) as exc_info:
            with session.batch() as batch:
                batch.delete(table, self.ALL)
                batch.insert(table, columns, invalid_input)

        cause = exc_info.exception.cause
        self.assertEqual(cause.code(), grpc.StatusCode.FAILED_PRECONDITION)
        error_msg = (
            'Invalid value for column value in table '
            'counters: Expected INT64.')
        self.assertEqual(cause.details(), error_msg)
```

Reasons:

- Would be nice to get the **actual** error message (vs. just some "stringly typed" checks)
- The `Expected INT64.` piece in the error message is quite useful

chemelnucfin: My concern with having `INT64` in the error message was that it ties it down to a specific type?  
Obviously that's the type used...

dhermes: Right, which is why it's OK to tie down to that type. Because we know what type we sent. I can't tell if you're agreeing or disagreeing.

PR #4572, opened 2017-12-11
Spanner: Range read with single key

chemelnucfin:


spanner/tests/system/test_system.py
```diff
@@ -960,6 +960,12 @@ def test_read_w_ranges(self):
         snapshot = session.snapshot(read_timestamp=committed, multi_use=True)
         all_data_rows = list(self._row_data(ROW_COUNT))
 
+        single_key = KeyRange(start_closed=[START], end_open=[START+1])
```
dhermes: `START+1` -> `START + 1` here and below (sorry for being a human linter)

chemelnucfin: I'll change it like so to comply with PEP8.  Making the spaces inside slices makes it ugly.

spanner/tests/system/test_system.py
```diff
@@ -960,32 +960,38 @@ def test_read_w_ranges(self):
...
+        rows = list(snapshot.read(self.TABLE, self.COLUMNS, keyset))
+        expected = all_data_rows[START : START+1]
+        self._check_rows_data(rows, expected)
+
         closed_closed = KeyRange(start_closed=[START], end_closed=[END])
         keyset = KeySet(ranges=(closed_closed,))
         rows = list(snapshot.read(
             self.TABLE, self.COLUMNS, keyset))
-        expected = all_data_rows[START:END+1]
+        expected = all_data_rows[START : END+1]
```
dhermes: Similar comment as in #4573, I'm fairly certain `[START:END + 1]` is the desired spacing, but I'm open to being wrong (e.g. with a link to [PEP8][1] and/or the [Google Python Style Guide][2])

[1]: https://www.python.org/dev/peps/pep-0008/
[2]: https://google.github.io/styleguide/pyguide.html

chemelnucfin: From PEP8.  I think it is because the + binds more closely.
Yes:

ham[1:9], ham[1:9:3], ham[:9:3], ham[1::3], ham[1:9:]
ham[lower:upper], ham[lower:upper:], ham[lower::step]
ham[lower+offset : upper+offset]
ham[: upper_fn(x) : step_fn(x)], ham[:: step_fn(x)]
ham[lower + offset : upper + offset]

dhermes: Cool :+1: thanks for the lesson!

chemelnucfin: I would usually just do [a+b:c+d] anyway, but in this case [START + 1] was recommended, so just trying to be more consistent.

Anyway, merging.

PR #4574, opened 2017-12-11
Combine two logging statements into one.

dhermes:


dhermes:
Oy. Sorry for the terrible PR subject.

PR #4576, opened 2017-12-12
Retry transaction for read write keeping same transaction ID over retries

chemelnucfin:


chemelnucfin:
Somehow the tests are different locally and on circle ci.  Will look into it.

datastore/google/cloud/datastore/transaction.py
```diff
@@ -262,3 +279,28 @@ def put(self, entity):
             raise RuntimeError("Transaction is read only")
         else:
             super(Transaction, self).put(entity)
+
+
+def _sleep(current_sleep, max_sleep=30.0, multiplier=2.0):
```
dhermes: You shouldn't need to implement a retry here, it should be in `api_core`

datastore/google/cloud/datastore/batch.py
```diff
@@ -270,9 +270,10 @@ def commit(self):
             raise ValueError('Batch must be in progress to commit()')
 
         try:
-            self._commit()
+            commit_response_pb = self._commit()
         finally:
             self._status = self._FINISHED
+        return commit_response_pb
```
dhermes: Why did you add this? It doesn't seem like it's used in `transaction.py`?

dhermes:
@chemelnucfin The code in this PR doesn't seem to match the description. Can you clarify?

chemelnucfin:
I don't think this is needed any more.

PR #4585, opened 2017-12-13
BigQuery Data Transfer 

landrito:


lukesneeringer:
@jonparrott This API has a problem. It is generated into the `google.cloud.bigquery.datatransfer_v1` namespace, which obviously competes with `google.cloud.bigquery`. This is not just a matter of monorepo things; the BigQuery package does not declare `google.cloud.bigquery` as a namespace package.

Ernest, you are the engineering point of contact listed on the spreadsheet. We made an explicit decision that this is a separate package, yes? If so, we want it to be a separate package for all languages.

I think that means we need to make this be `google.cloud.bigquerydatatransfer_v1` (gross; I wish I had used `_` for video intelligence and error reporting).

The other option is to include `datatransfer` in the `google-cloud-bigquery` API package even if other languages do not.

Thoughts?

lukesneeringer:
The directory name is also so gross that I am tempted to say that using an `_` word separator is the rule going forward, with error reporting and video intelligence as regrettable exceptions.

jonparrott:
I agree with luke's assessment, we either need to:

1. Place datatransfer into the bigquery package and tie its version together with bigquery. (There is precedence for this, both Spanner and BigTable have separate Admin APIs that are distributed along with their database APIs)
2. Create a new top-level `google.cloud.bigquery_datatransfer` package.

> he directory name is also so gross that I am tempted to say that using an _ word separator is the rule going forward, with error reporting and video intelligence as regrettable exceptions. Although if we house it within BigQuery I do not care nearly as much (google.cloud.bigquery.datatransfer_v1 is fine).

Neither are 1.0.0, so we can still fix this and use an alias for a few releases.

lukesneeringer:
> Neither are 1.0.0, so we can still fix this and use an alias for a few releases.

Video is 1.0.

lukesneeringer:
> Place datatransfer into the bigquery package and tie its version together with bigquery. (There is precedence for this, both Spanner and BigTable have separate Admin APIs that are distributed along with their database APIs)

This is what I would prefer to do, but I am nervous about being different from the other six languages. Thoughts?

jonparrott:
> This is what I would prefer to do, but I am nervous about being different from the other six languages. Thoughts?

I'm not overly concerned, as the only noticeable difference will be the package name used for installation.

lukesneeringer:
@tswast @jonparrott Okay, this is ready. Can I get an LGTM?

Issue #4587, opened 2017-12-13
BigQuery tests should avoid `pandas` in Python 3.4


dhermes:
At the end of the `0.20.x` series `3.4` was supported:

![screenshot from 2017-12-13 14-19-27](https://user-images.githubusercontent.com/520669/33965439-b4e8de7c-e010-11e7-9263-23279802f515.png)

but at the beginning of the `0.21.x` series it is not:

![screenshot from 2017-12-13 14-19-16](https://user-images.githubusercontent.com/520669/33965446-b8e9126c-e010-11e7-93ee-9428f72d544b.png)

There didn't seem to be an issue with `0.21.0` but `0.21.1` just came out and broke the [first build][3] it ran in.

[1]: https://pypi.org/project/pandas/0.20.3/
[2]: https://pypi.org/project/pandas/0.21.0/
[3]: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/4734

jonparrott:
I guess this begs the question of if we should drop 3.4 support.

dhermes:
Heck yeah we should. Hasn't it already been EOL-ed?

chemelnucfin:
https://devguide.python.org/#status-of-python-branches

Apparently not until 2019-03-16.

tseaver:
3.3 was EOLed in September.  FWIW, 3.4 is still the default Python 3 on LTS Linux versions.

tswast:
Sounds like they dropped it because Conda doesn't support it. https://github.com/pandas-dev/pandas/issues/15251 That doesn't really apply to us here.

PR #4588, opened 2017-12-13
Removing redundant constant.

dhermes:
It looks like it was moved to `google.api_core.gapic_v1.client_info` and during the move the delete was forgotten.


dhermes:
Merging, as the BigQuery failure is #4587.

PR #4590, opened 2017-12-13
Do not use easily-misread glyphs in Firestore auto-IDs.

dhermes:
This PR removes `{'i', 'I', 'l', 'o', 'O', '0', '1'}` from the set of characters used in automatically-generated IDs, because it is easy to mis-read them if you are in a situation where you need to do that.

Exact set of blacklisted characters is loaned from Django's [`make_random_password`](https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#django.contrib.auth.models.BaseUserManager.make_random_password).

/cc @schmidt-sebastian

----

Was #4107

### DIFF
--- a/firestore/tests/unit/test_collection.py
+++ b/firestore/tests/unit/test_collection.py
@@ -427,12 +427,12 @@ class Test__auto_id(unittest.TestCase):
     def test_it(self, mock_rand_choice):
         from google.cloud.firestore_v1beta1.collection import _AUTO_ID_CHARS
 
-        mock_result = '0123456789abcdefghij'
+        mock_result = '23456789abcdefghjkmnp'
         mock_rand_choice.side_effect = list(mock_result)
         result = self._call_fut()
         self.assertEqual(result, mock_result)
 
-        mock_calls = [mock.call(_AUTO_ID_CHARS)] * 20
+        mock_calls = [mock.call(_AUTO_ID_CHARS)] * 21
         self.assertEqual(mock_rand_choice.mock_calls, mock_calls)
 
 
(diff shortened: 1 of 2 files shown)


dhermes:
@lukesneeringer This is the 3rd time you or I have sent this PR (once was in a private repo) :grinning: 

The previous two were a no-go because we were waiting for a thumbs up from the Firestore team.

chemelnucfin:
@lukesneeringer @dhermes My fault.  I just thought it has been approved and was just sitting, so I merged it.

lukesneeringer:
@dhermes This is a pretty minor change. Either merge it or close it and let's move on with our lives. :-)

firestore/tests/unit/test_collection.py
```diff
@@ -427,12 +427,12 @@ def _call_fut():
...
         from google.cloud.firestore_v1beta1.collection import _AUTO_ID_CHARS
 
-        mock_result = '0123456789abcdefghij'
+        mock_result = '23456789abcdefghjkmnp'
         mock_rand_choice.side_effect = list(mock_result)
         result = self._call_fut()
         self.assertEqual(result, mock_result)
 
-        mock_calls = [mock.call(_AUTO_ID_CHARS)] * 20
+        mock_calls = [mock.call(_AUTO_ID_CHARS)] * 21
```
schmidt-sebastian: It should be 20 characters in length.

dhermes: @schmidt-sebastian 21 is correct (the change in this PR is dropping set of possible characters, so we add an extra character to keep the entropy up)

schmidt-sebastian: The decreased level of entropy seems more acceptable than having auto-ids that differ in length between implementations. If you are worried about the entropy, I would much rather suggest dropping this PR :)

dhermes: That works for me. Dropping it.

Issue #4578, opened 2017-12-13
Logging: expose "uniqueWriterIdentity" for creating sinks


polleyg:
Derived from SO question [here](https://stackoverflow.com/questions/47769820/how-to-set-uniquewriteridentity-true-on-sink-create).

To create sinks cross-project, `uniqueWriterIdentity` (see [here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.sinks/create)) needs to be set to `true`, otherwise the following error is thrown:

`google.gax.errors.RetryError: RetryError(Exception occurred in retry method that was not classified as transient, caused by <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, The destination project must be the same as the sink's project. Use uniqueWriterIdentity=true to allow cross project exports.)>)`

Currently this is not supported by the SDK. Coalescing logs/sinks to a single (admin) project is common pattern, so it should be available.

Until it's supported, it should also be clearly documented that it's not possible to use this SDK to create sinks cross-project.

dhermes:
Closed by #4595.

PR #4595, opened 2017-12-15
Logging: Unique Writer Identity

chemelnucfin:
#4578 Not sure if this is right, but an attempt.

### DIFF
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -222,12 +222,12 @@ class _SinksAPI(object):
         parent = 'projects/%s' % (project,)
         sink_pb = LogSink(name=sink_name, filter=filter_,
                           destination=destination)
+        uwi = unique_writer_identity
         try:
             self._gax_api.create_sink(parent,
-                sink_pb,
-                unique_writer_identity=unique_writer_identity,
-                options=options,
-            )
+                                      sink_pb,
+                                      unique_writer_identity=uwi,
+                                      options=options)
         except GaxError as exc:
             if exc_to_code(exc.cause) == StatusCode.FAILED_PRECONDITION:
                 path = 'projects/%s/sinks/%s' % (project, sink_name)
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -222,12 +222,12 @@ class _SinksAPI(object):
         parent = 'projects/%s' % (project,)
         sink_pb = LogSink(name=sink_name, filter=filter_,
                           destination=destination)
-        uwi = unique_writer_identity
         try:
             self._gax_api.create_sink(parent,
-                                      sink_pb,
-                                      unique_writer_identity=uwi,
-                                      options=options)
+                sink_pb,
+                unique_writer_identity=unique_writer_identity,
+                options=options,
+            )
         except GaxError as exc:
             if exc_to_code(exc.cause) == StatusCode.FAILED_PRECONDITION:
                 path = 'projects/%s/sinks/%s' % (project, sink_name)
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -222,12 +222,13 @@ class _SinksAPI(object):
         parent = 'projects/%s' % (project,)
         sink_pb = LogSink(name=sink_name, filter=filter_,
                           destination=destination)
-        uwi = unique_writer_identity
         try:
-            self._gax_api.create_sink(parent,
-                                      sink_pb,
-                                      unique_writer_identity=uwi,
-                                      options=options)
+            self._gax_api.create_sink(
+                parent,
+                sink_pb,
+                unique_writer_identity=unique_writer_identity,
+                options=options,
+            )
         except GaxError as exc:
             if exc_to_code(exc.cause) == StatusCode.FAILED_PRECONDITION:
                 path = 'projects/%s/sinks/%s' % (project, sink_name)
--- a/logging/google/cloud/logging/_gax.py
+++ b/logging/google/cloud/logging/_gax.py
@@ -192,7 +192,8 @@ class _SinksAPI(object):
         return page_iterator._GAXIterator(
             self._client, page_iter, _item_to_sink)
 
-    def sink_create(self, project, sink_name, filter_, destination):
+    def sink_create(self, project, sink_name, filter_, destination,
+                    unique_writer_identity=False):
         """API call:  create a sink resource.
 
         See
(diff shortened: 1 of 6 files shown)


logging/google/cloud/logging/_gax.py
```diff
@@ -211,13 +212,22 @@ def sink_create(self, project, sink_name, filter_, destination):
         :type destination: str
         :param destination: destination URI for the entries exported by
                             the sink.
+
+        :type unique_writer_identity: bool
```
duggelz: Just a general question whenever I see a 'bool' parameter: Is there any scenario you can image where you might want a third choice?  In which case, make this an enum instead of bool?

dhermes: @duggelz In this case, the parameter corresponds to [a protobuf field][1] so the mapping is as expected. (Worth noting that `proto3` has no distinction between "unset" and "zero value" for simple fields like `bool`, `int`, etc.)

[1]: https://github.com/googleapis/googleapis/blob/c44c8d7edcb75dcafdf75d0cdfa7f98747982266/google/logging/v2/logging_config.proto#L258

logging/google/cloud/logging/_gax.py
```diff
@@ -211,13 +212,22 @@ def sink_create(self, project, sink_name, filter_, destination):
...
         parent = 'projects/%s' % (project,)
         sink_pb = LogSink(name=sink_name, filter=filter_,
                           destination=destination)
+        uwi = unique_writer_identity
         try:
-            self._gax_api.create_sink(parent, sink_pb, options=options)
+            self._gax_api.create_sink(parent,
+                                      sink_pb,
+                                      unique_writer_identity=uwi,
+                                      options=options)
```
dhermes: If line length issues are the cause of your `uwi` variable, I suggest instead:

```python
            self._gax_api.create_sink(
                parent,
                sink_pb,
                unique_writer_identity=unique_writer_identity,
                options=options,
            )
```

(This way of indenting is probably the "best" in the sense that a machine could easily follow this strategy, though PEP8 accepts many different ways to do line continuation.)

logging/google/cloud/logging/sink.py
```diff
@@ -40,11 +40,13 @@ class Sink(object):
     :param client: A client which holds credentials and project configuration
                    for the sink (which requires a project).
     """
-    def __init__(self, name, filter_=None, destination=None, client=None):
+    def __init__(self, name, filter_=None, destination=None, client=None,
+                 unique_writer_identity=False):
```
dhermes: Please update the class docstring with the newly added argument

logging/google/cloud/logging/sink.py
```diff
@@ -116,7 +118,9 @@ def create(self, client=None):
         """
         client = self._require_client(client)
         client.sinks_api.sink_create(
-            self.project, self.name, self.filter_, self.destination)
+            self.project, self.name, self.filter_, self.destination,
+            self._unique_writer_identity
```
dhermes: 1. Please use a trailing comma after the last argument
1. Please pass the keyword argument as a keyword (not positional)

logging/tests/unit/test__gax.py
```diff
@@ -40,6 +40,7 @@ class _Base(object):
     PROJECT = 'PROJECT'
     PROJECT_PATH = 'projects/%s' % (PROJECT,)
     FILTER = 'logName:syslog AND severity>=ERROR'
+    UNIQUE_WRITER_IDENTITY = True
```
dhermes: How necessary is it to make this a constant? It's easy enough to just type the literals `True` and `False`?

logging/tests/unit/test__gax.py
```diff
@@ -737,7 +738,7 @@ def test_sink_create_ok(self):
         api.sink_create(
             self.PROJECT, self.SINK_NAME, self.FILTER, self.DESTINATION_URI)
 
-        parent, sink, options = (
+        parent, sink, options, unique_writer_identity = (
```
dhermes: Please use this variable (lint should complain that it's an unused variable).

```python
self.assertFalse(unique_writer_identity)
```

logging/tests/unit/test__gax.py
```diff
@@ -746,6 +747,25 @@ def test_sink_create_ok(self):
...
         self.assertIsNone(options)
 
+    def test_sink_create_with_unique_writer_identity(self):
+        from google.cloud.proto.logging.v2.logging_config_pb2 import LogSink
+
+        gax_api = _GAXSinksAPI()
+        api = self._make_one(gax_api, None)
+        api.sink_create(
+            self.PROJECT, self.SINK_NAME, self.FILTER, self.DESTINATION_URI,
+            self.UNIQUE_WRITER_IDENTITY
```
dhermes: 1. Please use keyword here
1. Just use a literal `True` or `False`?

logging/tests/unit/test__gax.py
```diff
@@ -746,6 +747,25 @@ def test_sink_create_ok(self):
...
+        )
+        parent, sink, options, unique_writer_identity = (
+            gax_api._create_sink_called_with)
+        self.assertEqual(parent, self.PROJECT_PATH)
+        self.assertIsInstance(sink, LogSink)
+        self.assertEqual(sink.name, self.SINK_NAME)
+        self.assertEqual(sink.filter, self.FILTER)
+        self.assertEqual(sink.destination, self.DESTINATION_URI)
+        self.assertEqual(unique_writer_identity, self.UNIQUE_WRITER_IDENTITY)
```
dhermes: May make more sense to use `assertTrue/assertFalse` here

logging/tests/unit/test__http.py
```diff
@@ -476,6 +476,26 @@ def test_sink_create_ok(self):
...
+            'filter': self.FILTER,
+            'destination': self.DESTINATION_URI,
+        }
+        
+        conn = _Connection({})
+        client = _Client(conn)
+        api = self._make_one(client)
+
+        api.sink_create(
+            self.PROJECT, self.SINK_NAME, self.FILTER, self.DESTINATION_URI, True)
```
dhermes: Please pass as a keyword argument here (vs. positional)

logging/tests/unit/test__http.py
```diff
@@ -476,6 +476,26 @@ def test_sink_create_ok(self):
...
+
+        self.assertEqual(conn._called_with['method'], 'POST')
+        path = '/projects/%s/sinks' % (self.PROJECT,)
+        self.assertEqual(conn._called_with['path'], path)
+        self.assertEqual(conn._called_with['data'], SENT)
+        self.assertEqual(conn._called_with['query_params'], {'uniqueWriterIdentity': True})
```
dhermes: If `conn._called_with` is a dictionary, why not just construct the expected dictionary and assert equality? In the current approach, `_called_with` could grow new keys and your test wouldn't detect that.

logging/google/cloud/logging/sink.py
chemelnucfin: trailing comma?

logging/google/cloud/logging/_gax.py
chemelnucfin: @duggelz Thank you for reviewing my code.  As @dhermes mentioned, I grabbed the type from the protobuf field.

logging/tests/unit/test__gax.py
chemelnucfin: I guess just to be consistent with the other parameters, I'll keep it like that.

dhermes: The other parameters have meaning, `True` is a literal with only 1 bit of "entropy".

PR #4597, opened 2017-12-15
BigQuery: removes support for python 3.4

alixhami:
Addresses https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4587

I queried the [pypi dataset](https://bigquery.cloud.google.com/dataset/the-psf:pypi) to see which versions of Python have been used with google-cloud-bigquery for the last 6 months.

python version | % of downloads
--- | ---
2.7 | 91.3%
3.6 | 4.1%
3.5 | 2.8%
3.4 | 1.5%
null | 0.3%
other | 0.01%

### DIFF
--- a/bigquery/CHANGELOG.md
+++ b/bigquery/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 [1]: https://pypi.org/project/google-cloud-bigquery/#history
 
+## 0.x.y
+
+- Remove support for Python version 3.4. (#4597)
+
+
 ## 0.28.0
 
 **0.28.0 significantly changes the interface for this package.** For examples
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -54,7 +54,7 @@ def default(session):
 
 
 @nox.session
-@nox.parametrize('py', ['2.7', '3.4', '3.5', '3.6'])
+@nox.parametrize('py', ['2.7', '3.5', '3.6'])
 def unit(session, py):
     """Run the unit test suite."""
 
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -42,7 +42,6 @@ SETUP_BASE = {
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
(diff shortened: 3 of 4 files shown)


dhermes:
@alixhami It's probably worth 

- Updating the docs immediately
- Making a note in `CHANGELOG.md` in this PR (so that it's clear in the next release, but worded now)

alixhami:
@dhermes What should be changed in the docs immediately? I thought we generated the changelog in the release PR with the names of PRs since the last release.

dhermes:
> What should be changed in the docs immediately?

Just add a note that `3.4` is no longer supported.

> I thought we generated the changelog in the release PR with the names of PRs since the last release.

`CHANGELOG.md` is a living file, you can edit it whenever you like. We have not yet edited them outside of release PRs, but there's no hard and fast rule. You could just add a section with an ambiguous version (e.g. `0.x.y`) and then put one bullet in there mentioning this PR

lukesneeringer:
It makes me sad that Python 2.7 is still _91%_ of our usage.

I am not sure I am comfortable with this change; can we discuss it? In particular:

  * I think we should have consistent support across APIs. So, if we are dropping Python 3.4 in BigQuery, we should drop it everywhere.
  * Even with low usage, dropping a Python version is probably a semver-major change. (BigQuery is not 1.0 yet, but see above.)
  * Ideally I would like to tie version support to the EOL dates on the Python versions. This is _especially_ important for when it _does_ come time to drop Python 2.7. I am going to want to point to that bright, shiny, "Python 2.7 is EOL in 2020" sign, and that case is weaker if we are willy nilly about other versions.

jonparrott:
It was my idea. ;)

tseaver:
@alixhami I don't actually think those stats mean what you think they mean:  it would be more in line with them to drop all Python3 support.

Is Pandas "core" to BigQuery's mission?  If not, let's not have the tail wag the dog here.

alixhami:
It sounds like the Python 3.4 support decision should be made at a higher level than the BigQuery client library, so I created PR https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4681 to address just the BigQuery/Pandas side of things

PR #4606, opened 2017-12-16
Making it impossible to call `Policy.open()` on an already opened policy.

### LINKED ISSUE #4488: Multiple calls to Pub/Sub Policy.open() can "interact"
[E.g.][1] if you call `Policy.open()` a second time then it will overwrite the callback attached to the `Policy`.

ISTM that we should "tombstone" a `Policy` so that it can't be `open()`-ed more than once.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/cc7da65b4a31d7a33055afe0bd6dc9855c3d6493/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py#L150

dhermes:
Similar with `Policy.close()`.

Fixes #4488.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -147,7 +147,22 @@ class Policy(base.BasePolicy):
             return executor
 
     def close(self):
-        """Close the existing connection."""
+        """Close the existing connection.
+
+        .. warning::
+
+            This method is not thread-safe. For example, if this method is
+            called while another thread is executing :meth:`open`, then the
+            policy could end up in an undefined state. The **same** policy
+            instance is not intended to be used by multiple workers (though
+            each policy instance **does** have a threadsafe private queue).
+
+        Raises:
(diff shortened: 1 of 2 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -159,7 +174,7 @@ def close(self):
 
         # The subscription is closing cleanly; resolve the future if it is not
         # resolved already.
-        if self._future is not None and not self._future.done():
+        if not self._future.done():
             self._future.set_result(None)
         self._future = None
```
dhermes: @lukesneeringer WDYT of returning the future here?

```python
future = self._future
self._future = None
return future
```

dhermes:
I'll hold off on merging until we can chat about the possibility of returning the "removed" future in `Policy.close()`.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: That is fine with me.

PR #4610, opened 2017-12-18
Manual update of docs.

### LINKED ISSUE #3740: Spanner docs are outdated

dhermes:
Done at commit 1a37fce18dd2fdbbcde1af61e7fa672c427233ad.

Fixes #3740.


dhermes:
Done via:

```
$ git worktree add ../temp
$ cd ../temp
$ nox -s docs
$ mv docs/_build/html/ ..
$ git clean --force -x -d
$ git branch gh-pages-manual upstream/gh-pages
$ rm -fr latest/
$ mv ../html/ latest/
$ git add latest/
$ git commit
$ git push official gh-pages-manual
$ cd -
$ rm -fr ../temp
$ git branch -D temp
$ git worktree prune
```

dhermes:
I am merging this (without LGTM) as there is no code written by a human, so it's just a matter of process while the broken build gets sorted out.

PR #4612, opened 2017-12-18
Adding extra check in `Batch.will_accept` for the number of messages.

dhermes:
Also adding a note about the fleeting nature of the "will accept?" check.


dhermes:
FYI this is towards #4575

PR #4613, opened 2017-12-18
Moving `will_accept()` check out of `PublisherClient.batch()` factory.

dhermes:
This is the first stage of moving that check directly into `Batch.publish()`. This check **must** occur when the message is being published because concurrent access to the messages and futures on a `Batch` make it impossible to reliably do an LBYL "will accept?" check.

Towards #4575.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -98,19 +98,21 @@ class Client(object):
         """
         return publisher_client.PublisherClient.SERVICE_ADDRESS
 
-    def batch(self, topic, message, create=True, autocommit=True):
+    def batch(self, topic, create=False, autocommit=True):
         """Return the current batch for the provided topic.
 
-        This will create a new batch only if no batch currently exists.
+        This will create a new batch if ``create=True`` or if no batch
+        currently exists.
 
         Args:
             topic (str): A string representing the topic.
-            message (~google.cloud.pubsub_v1.types.PubsubMessage): The message
-                that will be committed.
-            create (bool): Whether to create a new batch if no batch is
-                found. Defaults to True.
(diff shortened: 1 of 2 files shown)


pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -190,4 +193,12 @@ def publish(self, topic, data, **attrs):
...
 
         # Delegate the publishing to the batch.
-        return self.batch(topic, message=message).publish(message)
+        batch = self.batch(topic)
+        future = None
+        while future is None:
+            if batch.will_accept(message):
+                future = batch.publish(message)
+            else:
+                batch = self.batch(topic, create=True)
```
dhermes: @lukesneeringer WDYT of possibly logging in this "failure" branch?

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -98,30 +98,34 @@ def target(self):
...
         # and place it on the batches dictionary.
         with self._batch_lock:
-            batch = self._batches.get(topic, None)
-            if batch is None or not batch.will_accept(message):
-                if not create:
-                    return None
+            if not create:
+                batch = self._batches.get(topic)
+                if batch is None:
+                    create = True
```
dhermes: @lukesneeringer WDYT of possibly logging in this "doesn't exist" branch? I suppose users could get the logging they desire by using a custom `_batch_class` and logging in the constructor.

lukesneeringer: I think the bar to add DEBUG or INFO logs is low. If you think it has value, add it.

dhermes: May do in a follow-up PR

PR #4614, opened 2017-12-18
Adding a "STARTING" status for Pub / Sub `Batch.commit()`.

dhermes:
Towards #4575.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
@@ -18,6 +18,8 @@ import logging
 import threading
 import time
 
+import six
+
 from google.cloud.pubsub_v1 import types
 from google.cloud.pubsub_v1.publisher import exceptions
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -162,48 +165,47 @@ def _commit(self):
             version, which calls this one.
         """
         with self._commit_lock:
-            # If, in the intervening period, the batch started to be committed,
-            # or completed a commit, then no-op at this point.
-            if self._status != base.BatchStatus.ACCEPTING_MESSAGES:
```
dhermes: @lukesneeringer Note that by **asserting** that only `ACCEPTING_MESSAGES` could advance, this meant that **any** commit initiated via `Batch.commit()` would bail out at this point. This is because `Batch.commit()` sets `status` to `in-flight` (which is not equal to "accepting messages").

This is the **root** cause of #4575, since `Batch.publish()` calls `Batch.commit()` when it hits the `max_messages` threshold.

PR #4616, opened 2017-12-18
Checking Pub / Sub `Batch.will_accept` in thread-safe way.

### LINKED ISSUE #4575: Pub / Sub Publisher: Async publish breaks around 1000 messages
Am unsure if this is a bug (can't find documentation explaining this behavior).
I can send up to 999 pubsub messages and wait for results before continuing, but if I try to send 1000, none of them complete. However, if I run in debug mode, they will complete (because I have a breakpoint in the code allowing more execution time).

It exhibits this failure both in the main thread, or calling Publisher.publish from a thread. 

2. OS type and version
MacOS High Sierra

3. Python version and virtual environment information `python --version`
Python 3.6.3

4. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`

google-api-core==0.1.2
google-auth==1.2.1
google-cloud-core==0.28.0
google-cloud-pubsub==0.29.2
google-gax==0.15.16
grpc-google-iam-v1==0.11.4

5. Stacktrace if available

None, hanging, no exception.

6. Steps to reproduce

Publish 1000 messages storing their futures. None of the futures will complete.

7. Code example
Goal here is to send specified number of messages per second and sleep.

```python
    def publish_function(self):
        # save all futures so we can wait for them to complete
        # Callbacks receive the future as their only argument, as defined in
        # the Future interface.
        def callback(future):
            message_id = future.result()
            self.logger.debug(f"Sent message: {message_id}")

        # The callback is added once you get the future. If you add a callback
        # and the future is already done, it will simply be executed immediately.
        future = self.publisher.publish(self.topic_path, data=self.message)
        print(future)
        future.add_done_callback(callback)
        self.results.append(future)

    def publish(self, total_messages, messages_per_second):
        i = 0
        sleep_time = 1
        if messages_per_second < 1:
            sleep_time = float(sleep_time) / float(messages_per_second)
        logger.debug(f"Sleep time is: {sleep_time}")
        while i < total_messages and not self.exit:
            start = time.time()
            # number of times to send per second
            # if less than one, then the sleep timer handles the wait. We just send one message here
            for j in range(int(math.ceil(messages_per_second))):
                self.publish_function()
                i += 1
                self.next_id()
                # exit inner loop if we hit the total to send
                if i >= total_messages or self.exit:
                    self.stop()
                    break

            # apply result to all futures in self.results. This blocks until the future it's called on completes
            # list forces the map to execute since map returns an iterable


            list(map(futures.Future.result, self.results)) <<<<<<<<<<<--------- ***HANGS HERE***

            end = time.time()
            logger.info(
                f"Took {end - start} to send {messages_per_second} messages in thread {current_thread().getName()}")
            logger.debug(f"If greater than 0 seconds, sleeping for {sleep_time - (end - start)}")
            if sleep_time - (end - start) > 0:
                time.sleep(sleep_time - (end - start))
```

dhermes:
Unit tests will be coming in the next commit.

Fixes #4575.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
@@ -114,7 +114,7 @@ class Batch(object):
 
         # If this message will make the batch exceed the ``max_bytes``
         # setting, return False.
-        if self.size + message.ByteSize() > self.settings.max_bytes:
+        if self.size + message.ByteSize() >= self.settings.max_bytes:
             return False
 
         # If this message will make the batch exceed the ``max_messages``
--- a/pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
@@ -114,7 +114,7 @@ class Batch(object):
 
         # If this message will make the batch exceed the ``max_bytes``
         # setting, return False.
-        if self.size + message.ByteSize() >= self.settings.max_bytes:
+        if self.size + message.ByteSize() > self.settings.max_bytes:
             return False
 
         # If this message will make the batch exceed the ``max_messages``
--- a/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
@@ -37,7 +37,7 @@ class Batch(base.Batch):
     """A batch of messages.
 
     The batch is the internal group of messages which are either awaiting
-    publication or currently in-flight.
+    publication or currently in progress.
 
     A batch is automatically created by the PublisherClient when the first
     message to be published is received; subsequent messages are added to
@@ -72,9 +72,9 @@ class Batch(base.Batch):
         self._topic = topic
         self._settings = settings
 
-        self._commit_lock = threading.Lock()
-        # These objects are all communicated between threads; ensure that
-        # any writes to them are atomic.
+        self._state_lock = threading.Lock()
+        # These members are all communicated between threads; ensure that
+        # any writes to them use the "state lock" to remain atomic.
         self._futures = []
         self._messages = []
         self._size = 0
@@ -133,17 +133,24 @@ class Batch(base.Batch):
     def commit(self):
         """Actually publish all of the messages on the active batch.
 
-        This synchronously sets the batch status to in-flight, and then opens
-        a new thread, which handles actually sending the messages to Pub/Sub.
-
         .. note::
 
             This method is non-blocking. It opens a new thread, which calls
             :meth:`_commit`, which does block.
+
+        This synchronously sets the batch status to "starting", and then opens
+        a new thread, which handles actually sending the messages to Pub/Sub.
+
+        If the current batch is **not** accepting messages, this method
+        does nothing.
         """
         # Set the status to "starting" synchronously, to ensure that
         # this batch will necessarily not accept new messages.
-        self._status = base.BatchStatus.STARTING
+        with self._state_lock:
+            if self._status == base.BatchStatus.ACCEPTING_MESSAGES:
+                self._status = base.BatchStatus.STARTING
+            else:
+                return
 
         # Start a new thread to actually handle the commit.
         commit_thread = threading.Thread(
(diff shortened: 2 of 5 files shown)


pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -213,13 +220,13 @@ def monitor(self):
...
         and then calls commit unless the batch has already been committed.
         """
-        # Note: This thread blocks; it is up to the calling code to call it
-        # in a separate thread.
-        #
+        # NOTE: This blocks; it is up to the calling code to call it
+        #       in a separate thread.
+
         # Sleep for however long we should be waiting.
         time.sleep(self._settings.max_latency)
```
dhermes: @lukesneeringer This could be a bug, e.g.

```python
>>> import time
>>> time.sleep(-1.0)
ValueError: sleep length must be non-negative
```

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -217,21 +230,35 @@ def test_publish():
...
     # messages.
-    assert batch.size == sum([m.ByteSize() for m in messages])
+    expected_size = sum([message_pb.ByteSize() for message_pb in messages])
+    assert batch.size == expected_size
     assert batch.size > 0  # I do not always trust protobuf.
 
 
-def test_publish_max_messages():
-    batch = create_batch(max_messages=4)
+def test_publish_not_will_accept():
```
dhermes: @lukesneeringer Note that I went with `not_will_accept` over `will_not_accept` here to make `git grep will_accept` slightly more useful. I can change to better grammar if you see fit.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
lukesneeringer: Happy to add `max(..., 0)`, but also happy to leave this be. :-)

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
lukesneeringer: That seems like a reasonable decision.

pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
```diff
@@ -114,7 +114,7 @@ def will_accept(self, message):
 
         # If this message will make the batch exceed the ``max_bytes``
         # setting, return False.
-        if self.size + message.ByteSize() > self.settings.max_bytes:
+        if self.size + message.ByteSize() >= self.settings.max_bytes:
```
lukesneeringer: `>` was correct.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -235,24 +242,34 @@ def publish(self, message):
...
+        # ``commit()`` will try to obtain the lock.
+        if num_messages >= self._settings.max_messages:
             self.commit()
 
-        # Return a Future. That future needs to be aware of the status
-        # of this batch.
-        f = futures.Future()
-        self._futures.append(f)
-        return f
+        return future
```
lukesneeringer: Where did the `future` variable come from?

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -84,19 +84,32 @@ def test_commit():
...
-            name='Thread-CommitBatchPublisher',
-            target=batch._commit,
-        )
-        Thread.return_value.start.assert_called_once_with()
+    # A thread should have been created to do the actual commit.
+    Thread.assert_called_once_with(
+        name='Thread-CommitBatchPublisher',
+        target=batch._commit,
+    )
+    Thread.return_value.start.assert_called_once_with()
```
lukesneeringer: My rationale for this being within the `with` block is that it makes more intuitive and logical sense to me to only use variables declared by context managers within the context manager.

pubsub/google/cloud/pubsub_v1/publisher/batch/base.py
dhermes: Ah yes. In my brain, the check was

```python
self.size > self.settings.max_bytes
```

DERP

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
dhermes: I'll put it back.

My rationale: do **as little as possible** while the context manager is active

PR #4617, opened 2017-12-18
Correct BigQuery job parameter annotations.

weslem:


dhermes:
Thanks for the PR @weslem!

PR #4618, opened 2017-12-18
Spanner: partial key range and bug fix with single key

chemelnucfin:
breaking up #4515

### DIFF
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -41,10 +41,10 @@ class KeyRange(object):
         if not any([start_open, start_closed, end_open, end_closed]):
             raise ValueError("Must specify at least a start or end row.")
 
-        if start_open and start_closed:
+        if start_open is not None and start_closed is not None:
             raise ValueError("Specify one of 'start_open' / 'start_closed'.")
 
-        if end_open and end_closed:
+        if end_open is not None and end_closed is not None:
             raise ValueError("Specify one of 'end_open' / 'end_closed'.")
 
         self.start_open = start_open
@@ -60,17 +60,19 @@ class KeyRange(object):
         """
         kwargs = {}
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-
-        if self.start_closed:
+        elif self.start_closed is not None:
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
+        else:
+            kwargs['start_closed'] = _make_list_value_pb([])
 
-        if self.end_open:
+        if self.end_open is not None:
             kwargs['end_open'] = _make_list_value_pb(self.end_open)
-
-        if self.end_closed:
+        elif self.end_closed is not None:
             kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
+        else:
+            kwargs['end_closed'] = _make_list_value_pb([])
 
         return KeyRangePB(**kwargs)
 
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -40,11 +40,11 @@ class KeyRange(object):
                  end_open=None, end_closed=None):
         if ((start_open is None and start_closed is None)
             or (start_open is not None and start_closed is not None)):
-            raise ValueError("Specify exactly one start key type")
+            raise ValueError('Specify exactly one of start_closed or start_open')
 
         if ((end_open is None and end_closed is None)
             or (end_open is not None and end_closed is not None)):
-            raise ValueError("Specify exactly one end key type")
+            raise ValueError('Specify exactly one of end_closed or end_open')
 
         self.start_open = start_open
         self.start_closed = start_closed
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -38,19 +38,13 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        keys = (start_open, start_closed, end_open, end_closed)
-        if all(key is None for key in keys):
-            raise ValueError("Must specify at least a start or end row.")
+        if ((start_open is None and start_closed is None)
+            or (start_open is not None and start_closed is not None)):
+            raise ValueError("Specify exactly one start key type")
 
-        if start_open is not None and start_closed is not None:
-            raise ValueError("Specify one of 'start_open' / 'start_closed'.")
-        elif start_open is None and start_closed is None:
-            start_closed = []
-
-        if end_open is not None and end_closed is not None:
-            raise ValueError("Specify one of 'end_open' / 'end_closed'.")
-        elif end_open is None and end_closed is None:
-            end_closed = []
+        if ((end_open is None and end_closed is None)
+            or (end_open is not None and end_closed is not None)):
+            raise ValueError("Specify exactly one end key type")
 
         self.start_open = start_open
         self.start_closed = start_closed
@@ -65,14 +59,16 @@ class KeyRange(object):
         """
         kwargs = {}
 
-        if self.start_open is not None:
+        if self.start_open:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-        else:
+
+        if self.start_closed:
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
 
-        if self.end_open is not None:
+        if self.end_open:
             kwargs['end_open'] = _make_list_value_pb(self.end_open)
-        else:
+
+        if self.end_closed:
             kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
 
         return KeyRangePB(**kwargs)
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -38,14 +38,19 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
+        keys = (start_open, start_closed, end_open, end_closed)
+        if all(key is None for key in keys):
             raise ValueError("Must specify at least a start or end row.")
 
         if start_open is not None and start_closed is not None:
             raise ValueError("Specify one of 'start_open' / 'start_closed'.")
+        elif start_open is None and start_closed is None:
+            start_closed = []
 
         if end_open is not None and end_closed is not None:
             raise ValueError("Specify one of 'end_open' / 'end_closed'.")
+        elif end_open is None and end_closed is None:
+            end_closed = []
 
         self.start_open = start_open
         self.start_closed = start_closed
@@ -62,17 +67,13 @@ class KeyRange(object):
 
         if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-        elif self.start_closed is not None:
-            kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
         else:
-            kwargs['start_closed'] = _make_list_value_pb([])
+            kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
 
         if self.end_open is not None:
             kwargs['end_open'] = _make_list_value_pb(self.end_open)
-        elif self.end_closed is not None:
-            kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
         else:
-            kwargs['end_closed'] = _make_list_value_pb([])
+            kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
 
         return KeyRangePB(**kwargs)
 
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -39,11 +39,11 @@ class KeyRange(object):
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
         if ((start_open is None and start_closed is None)
-            or (start_open is not None and start_closed is not None)):
+                or (start_open is not None and start_closed is not None)):
             raise ValueError('Specify exactly one of start_closed or start_open')
 
         if ((end_open is None and end_closed is None)
-            or (end_open is not None and end_closed is not None)):
+                or (end_open is not None and end_closed is not None)):
             raise ValueError('Specify exactly one of end_closed or end_open')
 
         self.start_open = start_open
--- a/spanner/google/cloud/spanner_v1/keyset.py
+++ b/spanner/google/cloud/spanner_v1/keyset.py
@@ -72,16 +72,16 @@ class KeyRange(object):
         """
         kwargs = {}
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
 
-        if self.start_closed:
+        if self.start_closed is not None:
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
 
-        if self.end_open:
+        if self.end_open is not None:
             kwargs['end_open'] = _make_list_value_pb(self.end_open)
 
-        if self.end_closed:
+        if self.end_closed is not None:
             kwargs['end_closed'] = _make_list_value_pb(self.end_closed)
 
         return KeyRangePB(**kwargs)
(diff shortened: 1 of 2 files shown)


spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -60,17 +60,19 @@ def to_pb(self):
...
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-
-        if self.start_closed:
+        elif self.start_closed is not None:
             kwargs['start_closed'] = _make_list_value_pb(self.start_closed)
+        else:
+            kwargs['start_closed'] = _make_list_value_pb([])
```
dhermes: As in #4515, I don't think the `if -> elif` change or the `else:` changes are warranted. (FWIW the `elif` is effectively re-doing verification from the constructor, so it's just a DRY to avoid it, the `else:` is the real problem).

spanner/tests/unit/test_keyset.py
```diff
@@ -15,6 +15,7 @@
 
 import unittest
 
+import six
```
dhermes: There should be two empty lines between top-level definitions (i.e. this import and the class the follows)

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
 
-    def test_ctor_w_end_open_and_end_closed(self):
+    def test_ctor_end_open_and_end_closed(self):
         KEY_1 = [u'key_1']
         KEY_2 = [u'key_2']
         with self.assertRaises(ValueError):
             self._make_one(end_open=KEY_1, end_closed=KEY_2)
 
-    def test_ctor_w_only_start_open(self):
+    def test_ctor_start_open_and_start_closed_empty_lists_and_end_closed(self):
```
dhermes: Can you shorten this name? `test_ctor_conflicting_empty_start`?

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
+        with self.assertRaises(ValueError):
+            self._make_one(start_open=[], start_closed=[], end_closed=KEY_1)
 
-    def test_ctor_w_only_start_closed(self):
+    def test_ctor_start_open_and_end_open_and_end_closed_empty_lists(self):
+        KEY_1 = [u'key_1']
+        with self.assertRaises(ValueError):
+            self._make_one(start_open=KEY_1, end_open=[], end_closed=[])
+
+    def _check_unused_keys(self, krange, used_keys):
```
dhermes: s/`krange`/`key_range` please

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
+        key_types = ('start_closed',
+                     'start_open',
+                     'end_closed',
+                     'end_open')
+        for key_type in key_types:
+            if key_type not in used_keys:
+                self.assertFalse(krange.to_pb().HasField(key_type))
```
dhermes: Is this not equivalent to `krange.to_pb() == KeyRangePB()`?

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
+    def test_ctor_start_open_and_start_closed_empty_lists_and_end_closed(self):
         KEY_1 = [u'key_1']
-        krange = self._make_one(start_open=KEY_1)
-        self.assertEqual(krange.start_open, KEY_1)
-        self.assertEqual(krange.start_closed, None)
-        self.assertEqual(krange.end_open, None)
-        self.assertEqual(krange.end_closed, None)
+        KEY_2 = [u'key_2']
+        with self.assertRaises(ValueError):
+            self._make_one(start_open=[], start_closed=[], end_closed=KEY_1)
```
dhermes: The `end_closed` must be here because the way the `any()` works in the constructor, but that smells a bit to me.

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
 
-    def test_ctor_w_only_start_closed(self):
+    def test_ctor_start_open_and_end_open_and_end_closed_empty_lists(self):
```
dhermes: `test_ctor_conflicting_empty_end`?

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +31,69 @@ def test_ctor_no_start_no_end(self):
...
         krange = self._make_one(end_closed=KEY_1)
-        self.assertEqual(krange.start_open, None)
-        self.assertEqual(krange.start_closed, None)
-        self.assertEqual(krange.end_open, None)
         self.assertEqual(krange.end_closed, KEY_1)
+        self.assertIsNone(krange.start_open)
+        self.assertIsNone(krange.start_closed)
+        self.assertIsNone(krange.end_open)
+
+    def test_ctor_single_key_end_open(self):
```
dhermes: Are these 4 single key tests really necessary?

spanner/tests/unit/test_keyset.py
```diff
@@ -119,6 +140,73 @@ def test_to_pb_w_start_open_and_end_closed(self):
...
+        used_keys.append('end_closed')
+        used_keys.append('start_closed')
+        self.assertTrue(krange_pb.HasField('start_closed'))
+        self.assertTrue((len(krange_pb.start_closed) == 0))
+        pb_key_type = krange_pb.end_closed
+        self.assertEqual(len(pb_key_type), 1)
+        self.assertEqual(pb_key_type.values[0].string_value, KEY_1[0])
+        self._check_unused_keys(krange, used_keys)
+
+    def test_to_pb_single_key_end_open(self):
```
dhermes: Ditto, are these single key tests really necessary?

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -41,10 +41,10 @@ def __init__(self, start_open=None, start_closed=None,
         if not any([start_open, start_closed, end_open, end_closed]):
```
dhermes: Do you not want to check `is None` in this `any()` to conform to the way it's handled below?

spanner/google/cloud/spanner_v1/keyset.py
dhermes: A summary of ongoing discussion in chat.

These hacks **seem** necessary because the **server** [requires][1] the start key

> The start key must be provided. It can be either closed or open.

and [requires][2] the end key

> The end key must be provided. It can be either closed or open.

I propose that we instead move that validation into the constructor, rather than do it here. This method should just be a "dumb" marshaling of already verified data.

[1]: https://github.com/googleapis/googleapis/blob/c5d0b08ecd3049c378972d833a49df99f779f97f/google/spanner/v1/keys.proto#L118
[2]: https://github.com/googleapis/googleapis/blob/c5d0b08ecd3049c378972d833a49df99f779f97f/google/spanner/v1/keys.proto#L129

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -38,14 +38,15 @@ class KeyRange(object):
...
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
-            raise ValueError("Must specify at least a start or end row.")
-
-        if start_open and start_closed:
+        if start_open is not None and start_closed is not None:
             raise ValueError("Specify one of 'start_open' / 'start_closed'.")
+        elif start_open is None and start_closed is None:
+            start_closed = []
```
dhermes: I think this should be an exception. If not, you need to update the docstring.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -38,14 +38,15 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
-            raise ValueError("Must specify at least a start or end row.")
-
-        if start_open and start_closed:
+        if start_open is not None and start_closed is not None:
```
dhermes: It seems "wasteful" to do this `start_foo is not None` check in both `if` statements. What about:

```python
num_start = int(start_open is not None) + int(start_closed is not None)
```

(I had to write it down to see it's not any better than the current approach.)

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: this has to be there or you can't do 
KeyRange(end_open = x)

chemelnucfin: But this also makes the empty constructor allowed.  I can add back in the (if not any...)

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: I don't really see how it's better.

chemelnucfin:
maybe breaking this up and changing the code all at once isn't such a great idea.....

chemelnucfin:
working on it...

spanner/tests/unit/test_keyset.py
chemelnucfin: I have to specify the used keys, so I don't think so?

spanner/tests/unit/test_keyset.py
chemelnucfin: Probably not really "necessary", but it does test different arguments.

Ditto also to below.

chemelnucfin: Also, there was an error that I caught because I had all different combinations due to a copy paste error.  So, I guess it's also helpful.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -38,14 +38,19 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
+        keys = (start_open, start_closed, end_open, end_closed)
+        if all(key is None for key in keys):
```
dhermes: The presence / necessity of this check indicates the inherent "problem" with the approach of converting `start_open=None, start_closed=None` to an implicit `start_closed=[]` (similar for `end_*`).

Seeing this / realizing that only **one** of `start_` and `end_` can fall back to an empty value, it seems that the lesser of two evils is requiring the **user** to provide the `[]`.

spanner/google/cloud/spanner_v1/keyset.py
dhermes: So here is what I'm saying, we have two options:

1. Instead of setting `start_closed = []` here, we raise an exception stating that exactly one of `start_open` and `start_closed` must be provided
1. We must update the `start_open /  start_closed` docstrings to explain that if **neither** is provided, then we implicitly take it to mean `start_closed=[]`

Similar applies to `end_*`. My comment above (about the need for the `all()` check) indicates my preference (which is getting stronger by the minute) for the "raise an exception" choice.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -60,16 +65,14 @@ def to_pb(self):
         """
         kwargs = {}
 
-        if self.start_open:
+        if self.start_open is not None:
             kwargs['start_open'] = _make_list_value_pb(self.start_open)
-
-        if self.start_closed:
+        else:
```
dhermes: @chemelnucfin As I mentioned before, anything other than the state it was in before is just re-validating the value. The only change this method needed was converting the `if self.foo:` checks into `if self.foo is not None:` checks.

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +33,68 @@ def test_ctor_no_start_no_end(self):
         with self.assertRaises(ValueError):
             self._make_one()
 
-    def test_ctor_w_start_open_and_start_closed(self):
+    def test_ctor_start_open_and_start_closed(self):
```
dhermes: FWIW, your editorial removal of the `w_` (short for "with") hasn't made these names easier to read (it hasn't made them harder to read either, but why bother with the change?)

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +33,68 @@ def test_ctor_no_start_no_end(self):
...
+        with self.assertRaises(ValueError):
+            self._make_one(start_open=[], start_closed=[], end_closed=KEY_1)
 
-    def test_ctor_w_only_start_closed(self):
+    def test_ctor_conflicting_end(self):
+        KEY_1 = [u'key_1']
+        with self.assertRaises(ValueError):
+            self._make_one(start_open=KEY_1, end_open=[], end_closed=[])
+
+    def _check_unused_keys(self, key_range, used_keys):
```
dhermes: As mentioned in #4623, I would prefer this helper be removed and you just construct the raw protobufs and use `self.assertEqual`. If you need help, I can provide it.

spanner/tests/unit/test_keyset.py
```diff
@@ -30,49 +33,68 @@ def test_ctor_no_start_no_end(self):
...
         krange = self._make_one(end_closed=KEY_1)
-        self.assertEqual(krange.start_open, None)
-        self.assertEqual(krange.start_closed, None)
-        self.assertEqual(krange.end_open, None)
         self.assertEqual(krange.end_closed, KEY_1)
+        self.assertEqual(krange.start_closed, [])
+        self.assertIsNone(krange.start_open)
+        self.assertIsNone(krange.end_open)
+
+    def test_ctor_single_key_end_open(self):
```
dhermes: 1. I don't think it's necessary to test **all four** possible "single input" cases
1. All four of these `single_key` examples will be exceptions under my new proposal

spanner/google/cloud/spanner_v1/keyset.py
chemelnucfin: 1 does make it more consistent with the behavior of the empty constructor.

spanner/tests/unit/test_keyset.py
chemelnucfin: it shortened some tests just enough to fit the name I think was my reasoning.  I wanted to make it consistent across the file.

spanner/google/cloud/spanner_v1/keyset.py
dhermes: On this we agree

chemelnucfin:
@dhermes PTAL, thanks.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -38,14 +38,13 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
-            raise ValueError("Must specify at least a start or end row.")
+        if ((start_open is None and start_closed is None)
+            or (start_open is not None and start_closed is not None)):
+            raise ValueError("Specify exactly one start key type")
```
dhermes: Any opposition to just saying "Specify exactly one of start_open and start_closed"? (Ditto for `end_*`)

chemelnucfin:
no opposition, changed.

spanner/google/cloud/spanner_v1/keyset.py
```diff
@@ -38,14 +38,13 @@ class KeyRange(object):
     """
     def __init__(self, start_open=None, start_closed=None,
                  end_open=None, end_closed=None):
-        if not any([start_open, start_closed, end_open, end_closed]):
-            raise ValueError("Must specify at least a start or end row.")
+        if ((start_open is None and start_closed is None)
+            or (start_open is not None and start_closed is not None)):
```
dhermes: Sorry I didn't mention this before, but continuations like this should be indented extra so it's not confused with the line below.

```python
if ((start_open is None and start_closed is None)
        or (start_open is not None and start_closed is not None)):
    raise ValueError('Specify exactly one of start_closed or start_open')
```

chemelnucfin: no problem.  
changed.

spanner/tests/unit/test_keyset.py
```diff
@@ -119,6 +117,33 @@ def test_to_pb_w_start_open_and_end_closed(self):
         self.assertEqual(len(krange_pb.end_closed), 1)
         self.assertEqual(krange_pb.end_closed.values[0].string_value, KEY_2[0])
 
+    def test_to_pb_single_key_start_closed(self):
+        from google.cloud.spanner_v1.proto.keys_pb2 import KeyRange
+
+        KEY_1 = [u'key_1']
+        with self.assertRaises(ValueError):
+            krange = self._make_one(start_closed=KEY_1)
```
dhermes: 1. You don't need to assign to `krange` here (Pylint would even complain that it is assigned but never used)
1. Please stop using `krange` as a variable name

chemelnucfin: removed.

PR #4623, opened 2017-12-19
Spanner: More keyset tests

chemelnucfin:
breaking up #4515


spanner/tests/unit/test_keyset.py
```diff
@@ -26,6 +29,15 @@ def _getTargetClass(self):
     def _make_one(self, *args, **kwargs):
         return self._getTargetClass()(*args, **kwargs)
 
+    def _check_unused_keys(self, krange, used_keys):
```
dhermes: 1. s/`krange`/`key_range` please
1. You should ditch this helper and just create the protobufs to compare against and really on the `__eq__` provided by the base `google.protobuf.Message` type

spanner/tests/unit/test_keyset.py
```diff
@@ -119,6 +131,366 @@ def test_to_pb_w_start_open_and_end_closed(self):
         self.assertEqual(len(krange_pb.end_closed), 1)
         self.assertEqual(krange_pb.end_closed.values[0].string_value, KEY_2[0])
 
+    def test_ctor_empty_list_start_closed_and_key_end_closed(self):
```
dhermes: @chemelnucfin Can you "port over" my previous comments? It seems I've seen this code 3 (?) times now and I'm having a deja vu about my comments.

chemelnucfin: Sorry, I think in the process of breaking up the PR some old code was retained.

chemelnucfin: Also, I think this test is different than the previous you commented.

Issue #3890, opened 2017-08-28
Pub / Sub doesn't work with multiprocessing


jonparrott:
This code fails with a pickle error:

```python
def multiprocessing_callback(message):
    print('Received message: {}'.format(message))
    message.ack()


def receive_messages_with_multiprocessing(
        project, topic_name, subscription_name):
    """Receives messages from a pull subscription using multiprocessing."""
    import concurrent.futures
    import functools
    import multiprocessing

    import google.cloud.pubsub_v1.subscriber.policy.thread

    # Create a process pool and a queue for sending messages. This will
    # be used by the subscriber policy to execute callbacks.
    executor = concurrent.futures.ProcessPoolExecutor()
    manager = multiprocessing.Manager()
    queue = manager.Queue()

    policy_factory = functools.partial(
        google.cloud.pubsub_v1.subscriber.policy.thread.Policy,
        executor=executor,
        queue=queue)
    subscriber = pubsub_v1.SubscriberClient(policy_class=policy_factory)
    subscription_path = subscriber.subscription_path(
        project, subscription_name)

    subscriber.subscribe(subscription_path, callback=multiprocessing_callback)

    # The subscriber is non-blocking, so we must keep the main thread from
    # exiting to allow it to process messages in the background.
    print('Listening for messages on {}'.format(subscription_path))
    while True:
        time.sleep(60)
```

lukesneeringer:
Can you provide the traceback, too? ;-)

jonparrott:
It's pretty unhelpful:

```
Traceback (most recent call last):
  File "/Users/jonwayne/.pyenv/versions/3.6.0/lib/python3.6/multiprocessing/queues.py", line 241, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/Users/jonwayne/.pyenv/versions/3.6.0/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: can't pickle _thread.lock objects
```

danoscarmike:
Hi @lukesneeringer what's the next step here?

michaelplaing:
This occurs when using the "spawn" context of multiprocessing.
However the "fork" context will also fail (but silently) when using grpcio==0.1.6.
A workaround for "fork" is to specify grpcio==0.1.4.

dhermes:
[...]

PR #4628, opened 2017-12-19
Allowing over-ride for lock used in publisher client.

dhermes:
Towards #3890.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -17,7 +17,6 @@ from __future__ import absolute_import
 import copy
 import os
 import pkg_resources
-import threading
 
 import grpc
 import six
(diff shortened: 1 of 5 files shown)


pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -44,18 +44,25 @@ class Client(object):
...
         batch_settings (~google.cloud.pubsub_v1.types.BatchSettings): The
             settings for batch publishing.
-        batch_class (class): A class that describes how to handle
+        batch_class (Optional[type]): A class that describes how to handle
             batches. You may subclass the
             :class:`.pubsub_v1.publisher.batch.base.BaseBatch` class in
             order to define your own batcher. This is primarily provided to
             allow use of different concurrency models; the default
             is based on :class:`threading.Thread`.
+        batch_lock (Optional[Any]): A lock used to ensure atomic modifications
```
lukesneeringer: Hmm. I might prefer an interface where `batch_class` defined its lock class. Thoughts?

dhermes: I started out that way, but `threading.Lock` **is not** a class. Then I transitioned to a `Callable[[], Any]`. But having a class or callable is overkill: there is only one lock needed and it gets created when the object is constructed.

lukesneeringer: Hmm. I am unconvinced, but I do not feel strongly about it either. :-)

dhermes: I misread Luke's initial comment. It turns out I agree with him.

PR #4629, opened 2017-12-19
Spanner: minor issues

chemelnucfin:


chemelnucfin:
really is minor though...  not even worth mentioning.  :)

PR #4631, opened 2017-12-19
Spanner: System test partial range

chemelnucfin:
Merge #4618 first.  I still need to test this.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -1003,77 +1003,59 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         expected = all_data_rows[START+1 : END+1]
         self._check_row_data(rows, expected)
 
-    def test_read_partial_range_start(self):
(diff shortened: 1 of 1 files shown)


chemelnucfin:
Ugh, I wrote tests in a loop style before and it was not allowed.

dhermes:
> Ugh, I wrote tests in a loop style before and it was not allowed.

Sorry to hear that. I am OK with it.

tseaver:
FWIW, the reason I asked @chemelnucfin to remove them elsewhere is that the loops introduce a lot of noise, and make debugging failures harder:  those tests are "speed bumps" for readers.

chemelnucfin:
I don't have a preference for either style.  I can see merits and drawbacks in both.

dhermes:
@tseaver I somewhat agree with you, but the copy-pasta creates other issues.

I am 100% for switching to `pytest` style, then we can use `pytest.mark.parameterize` to have one test but multiple data.

PR #4633, opened 2017-12-20
Fix build issue with dataproc docs.

dhermes:
Also:

- Fix an ambiguous name (`type`) in Pub / Sub (was introduced
  because I merged with the docs build failing)
- Marking the `dataproc` version as `.dev1`

### DIFF
--- a/dataproc/google/cloud/dataproc_v1/gapic/cluster_controller_client.py
+++ b/dataproc/google/cloud/dataproc_v1/gapic/cluster_controller_client.py
@@ -276,6 +276,7 @@ class ClusterControllerClient(object):
                         }
                       }
                     }
+
                 Similarly, to change the number of preemptible workers in a cluster to 5,
                 the ``update_mask`` parameter would be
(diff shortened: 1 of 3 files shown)


dhermes:
I am going to merge this pre-review since it fixes the docs build which is currently breaking every build.

lukesneeringer:
Thanks! We merged prior to docs because our build was failing on Spanner. _Mea culpa._

dhermes:
@lukesneeringer Totally get it (but the Spanner issue had been fixed in `master`, that PR was just behind `master`).

chemelnucfin:
Wait, the #4853 on master that failed which said Spanner failed on pubsub.

The previous build was fine, and the subsequent one failed on docs...  Which one is failing on spanner?

dhermes:
@chemelnucfin See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4626#issuecomment-352920294. All is well in GitHub-town.

PR #4635, opened 2017-12-20
Adds optional location_prefix kwarg in to_legacy_urlsafe

### LINKED ISSUE #3597: Datastore Key's reference is missing location identifier in app argument in to_legacy_urlsafe()
@dhermes I put this as a new issue because I thought people might have a problem with this too. Please feel free to edit/remove this issue if it is not appropriate to put it here.

I was running `pip install --upgrade google-cloud-datastore` to get latest version that has the `key.to_legacy_urlsafe()` implemented in [#3491](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3491). I tested the method and I found out that the urlsafe generated from the method is still not the same as datastore's urlsafe.

I have implemented my own function to generate the urlsafe key before the update was released and so far it worked properly for me (I can share my code if anybody wants). So I compared my function with the to_legacy_urlsafe and I found the problem:

```
reference = _app_engine_key_pb2.Reference(
        app=self.project,
        path=_to_legacy_path(self._path),  # Avoid the copy.
        name_space=self.namespace,
        )
```

In the app argument, the location identifier is missing unlike in my function where I concatenated the "s~" at the beginning of the string. I think this is the cause of the problem why it was not getting the urlsafe correct, although I don't understand why it needs the location identifier in front of the project name in the first place.

Do you think I should send a PR to fix this?

egalpin:
Fixes #3597

I encountered an issue with respect to urlsafe keys in a system that is using _both_ `google-cloud-datastore` and `ndb` in different services.  I was receiving incompatible urlsafe keys when using the _output_ of `to_legacy_urlsafe` in things like `ndb.Key(urlsafe=<output_from_legacy_urlsafe>)` running under App Engine. I received the following error:

```
BadRequestError: app s~your-app-id cannot access app your-app-id's data
```

 I then tried the suggestion as per #3597 of adding the location prefix to the project under cloud-datastore like so:

`datastore.Client(project='s~your-app-id')`

Unfortunately, I was met with this error (where `your-app-id` is a real project, with active datastore db which I can access via the Google Cloud Console):

```
The project s~your-app-id does not exist or it does not contain an active Cloud Datastore database.
```

I was able to produce compatible urlsafe keys by specifying the location prefix _only_ when creating the urlsafe keys. This PR contains that small change.

### DIFF
--- a/datastore/google/cloud/datastore/key.py
+++ b/datastore/google/cloud/datastore/key.py
@@ -298,7 +298,7 @@ class Key(object):
 
         return key
 
-    def to_legacy_urlsafe(self):
+    def to_legacy_urlsafe(self, location_prefix=None):
         """Convert to a base64 encode urlsafe string for App Engine.
 
         This is intended to work with the "legacy" representation of a
@@ -310,13 +310,25 @@ class Key(object):
         .. note::
 
             The string returned by ``to_legacy_urlsafe`` is equivalent, but
-            not identical, to the string returned by ``ndb``.
+            not identical, to the string returned by ``ndb``. The location
+            prefix may need to be specified to obtain idendentical urlsafe
+            keys.
+
+        :type location_prefix: str
+        :param location_prefix: The location prefix of ones project. Often
+                                this value is 's~', but may be 'dev~', 'e~', or
+                                other location prefixes currently unknown.
 
         :rtype: bytes
         :returns: A bytestring containing the key encoded as URL-safe base64.
         """
+        if location_prefix:
+            project_id = location_prefix + self.project
+        else:
+            project_id = self.project
+
         reference = _app_engine_key_pb2.Reference(
-            app=self.project,
+            app=project_id,
             path=_to_legacy_path(self._path),  # Avoid the copy.
             name_space=self.namespace,
         )
--- a/datastore/google/cloud/datastore/key.py
+++ b/datastore/google/cloud/datastore/key.py
@@ -311,21 +311,22 @@ class Key(object):
 
             The string returned by ``to_legacy_urlsafe`` is equivalent, but
             not identical, to the string returned by ``ndb``. The location
-            prefix may need to be specified to obtain idendentical urlsafe
+            prefix may need to be specified to obtain identical urlsafe
             keys.
 
         :type location_prefix: str
-        :param location_prefix: The location prefix of ones project. Often
-                                this value is 's~', but may be 'dev~', 'e~', or
-                                other location prefixes currently unknown.
+        :param location_prefix: The location prefix of an App Engine project
+                                ID. Often this value is 's~', but may also be
+                                'e~', or other location prefixes currently
+                                unknown.
 
         :rtype: bytes
         :returns: A bytestring containing the key encoded as URL-safe base64.
         """
-        if location_prefix:
-            project_id = location_prefix + self.project
-        else:
+        if location_prefix is None:
             project_id = self.project
+        else:
+            project_id = location_prefix + self.project
 
         reference = _app_engine_key_pb2.Reference(
             app=project_id,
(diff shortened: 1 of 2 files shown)


egalpin:
I've signed the CLA

egalpin:
If this isn't acceptable as a change in the repo, anyone wanting to do the same should be able to use this snippet in their code:

```python
import base64

from google.cloud.datastore import _app_engine_key_pb2
from google.cloud.datastore.key import _to_legacy_path


def to_proper_urlsafe(key, location_prefix=None):
    """Convert to a base64 encode urlsafe string for App Engine.

    "Reference"). The returned string can be used as the ``urlsafe``
    argument to ``ndb.Key(urlsafe=...)``. The base64 encoded values
    will have padding removed.
    .. note::
            The string returned by ``to_legacy_urlsafe`` is equivalent, but
            not identical, to the string returned by ``ndb``. The location
            prefix may need to be specified to obtain idendentical urlsafe
            keys.

    :type key: google.cloud.datastore.key.Key
    :param key: The key that you wish to convert to urlsafe

    :type location_prefix: str
    :param location_prefix: The location prefix of ones project. Often
                            this value is 's~', but may be 'dev~', 'e~', or
                            other location prefixes currently unknown.
    :rtype: bytes
    :returns: A bytestring containing the key encoded as URL-safe base64.
    """
    if location_prefix:
        project_id = location_prefix + key._project
    else:
        project_id = key._project

    reference = _app_engine_key_pb2.Reference(
        app=project_id,
        path=_to_legacy_path(key._path),  # Avoid the copy.
        name_space=key.namespace,
    )
    raw_bytes = reference.SerializeToString()
    return base64.urlsafe_b64encode(raw_bytes)
```

datastore/google/cloud/datastore/key.py
```diff
@@ -310,13 +310,25 @@ def to_legacy_urlsafe(self):
         .. note::
 
             The string returned by ``to_legacy_urlsafe`` is equivalent, but
-            not identical, to the string returned by ``ndb``.
+            not identical, to the string returned by ``ndb``. The location
+            prefix may need to be specified to obtain idendentical urlsafe
```
dhermes: s/idendentical/identical

datastore/google/cloud/datastore/key.py
```diff
@@ -310,13 +310,25 @@ def to_legacy_urlsafe(self):
...
+            keys.
+
+        :type location_prefix: str
+        :param location_prefix: The location prefix of ones project. Often
```
dhermes: s/ones project/an App Engine project ID

datastore/google/cloud/datastore/key.py
```diff
@@ -310,13 +310,25 @@ def to_legacy_urlsafe(self):
...
+                                this value is 's~', but may be 'dev~', 'e~', or
```
dhermes: `dev~` is only from the `dev_appserver`

datastore/google/cloud/datastore/key.py
```diff
@@ -310,13 +310,25 @@ def to_legacy_urlsafe(self):
...
+                                other location prefixes currently unknown.
 
         :rtype: bytes
         :returns: A bytestring containing the key encoded as URL-safe base64.
         """
+        if location_prefix:
```
dhermes: Can you please make this an explicit [`None`-check][1]:

```python
if location_prefix is None:
    project_id = self.project
else:
    project_id = location_prefix + self.project
```

dhermes:
@egalpin Thanks for the update, ~~are you planning on sending unit test updates or shall I?~~

I had some strange GitHub refresh state and never saw them. My bad.

datastore/tests/unit/test_key.py
```diff
@@ -430,6 +435,15 @@ def test_from_legacy_urlsafe_needs_padding(self):
         self.assertIsNone(key.namespace)
         self.assertEqual(key.flat_path, self._URLSAFE_FLAT_PATH2)
 
+    def test_from_legacy_urlsafe_with_location_prefix(self):
```
dhermes: @egalpin This test thinks it's a `from_legacy_urlsafe` test, but it is testing `to_legacy_urlsafe` (as it should be).

Do you mind changing the name and moving it up into the section just before `def test_from_legacy_urlsafe(self):` (line 416)?

If not, I can take the wheel, I really appreciate you taking the time to write the test.

egalpin:
I need update author details on some commits (working with multiple GH accounts on this machine)

dhermes:
@egalpin The CLA issue was because I pushed a commit to the PR and the CLA-bot gets confused by multi-author PRs. No need to do anything. Cheers!

PR #4640, opened 2017-12-20
Datastore: id from #3832 pull request with unit test

chemelnucfin:
See #3832.  Just following up with it.

I'm going to leave the document in place as it is the same as the other function.  I can make a separate pull request to remove implementation details.

### DIFF
--- a/datastore/google/cloud/datastore/entity.py
+++ b/datastore/google/cloud/datastore/entity.py
@@ -183,6 +183,22 @@ class Entity(dict):
         if self.key:
             return self.key.kind
 
+    @property
+    def id(self):
+        """Get the ID of the current entity.
+
+        .. note::
+
+            This relies entirely on the :class:`google.cloud.datastore.key.Key`
+            set on the entity.  That means that we're not storing the ID
+            of the entity at all, just the properties and a pointer to a
+            Key which knows its ID.
+        """
+        if self.key is None:
+            return None
+        else:
+            return self.key.id
+
     def __repr__(self):
         if self.key:
             return '<Entity%s %s>' % (self.key._flat_path,
--- a/datastore/tests/unit/test_entity.py
+++ b/datastore/tests/unit/test_entity.py
@@ -192,6 +192,19 @@ class TestEntity(unittest.TestCase):
 
         self.assertFalse(entity1 == entity2)
 
+    def test_id(self):
+        from google.cloud.datastore.key import Key
+
+        key = Key(_KIND, _ID, project=_PROJECT)
+        entity = self._make_one(key=key)
+        self.assertEqual(entity.id,  _ID)
+
+    def test_id_none(self):
+        from google.cloud.datastore.key import Key
+
+        entity = self._make_one(key=None)
+        self.assertEqual(entity.id,  None)
+
     def test___repr___no_key_empty(self):
         entity = self._make_one()
         self.assertEqual(repr(entity), '<Entity {}>')


datastore/tests/unit/test_entity.py
```diff
@@ -36,6 +36,7 @@ def test_ctor_defaults(self):
         entity = klass()
         self.assertIsNone(entity.key)
         self.assertIsNone(entity.kind)
+        self.assertIsNone(entity.id)
```
dhermes: @chemelnucfin IMO it's "bad form" to check a `@property` in the constructor test. I view these assertions as essentially checking whats in `entity.__dict__`

datastore/tests/unit/test_entity.py
```diff
@@ -192,6 +193,13 @@ def test__eq__same_value_different_meanings(self):
 
         self.assertFalse(entity1 == entity2)
 
+    def test_id(self):
+        from google.cloud.datastore.key import Key
+
+        key = Key(_KIND, _ID, project=_PROJECT)
+        entity = self._make_one(key=key)
+        self.assertTrue(entity.id == _ID)
```
dhermes: Please use `assertEqual`

datastore/tests/unit/test_entity.py
```diff
@@ -192,6 +193,13 @@ def test__eq__same_value_different_meanings(self):
 
         self.assertFalse(entity1 == entity2)
 
+    def test_id(self):
```
dhermes: You also need a test for the case when it is `None`

datastore/google/cloud/datastore/entity.py
```diff
@@ -183,6 +183,19 @@ def kind(self):
...
+    def id(self):
+        """Get the id of the current entity.
+
+        .. note::
+          This relies entirely on the :class:`gcloud.datastore.key.Key`
+          set on the entity.  That means that we're not storing the id
+          of the entity at all, just the properties and a pointer to a
+          Key which knows its Id.
+        """
+        if self.key is not None:
```
dhermes: From [PEP8][1]

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should.

So this should be:

```python
if self.key is None:
    return None
else:
    return self.key.id
```

[1]: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

datastore/google/cloud/datastore/entity.py
```diff
@@ -183,6 +183,19 @@ def kind(self):
         if self.key:
             return self.key.kind
 
+    @property
+    def id(self):
+        """Get the id of the current entity.
+
+        .. note::
+          This relies entirely on the :class:`gcloud.datastore.key.Key`
```
dhermes: 1. There is no "Sphinx / rST lint" but I've seen @lukesneeringer use 4 space indents after a `.. note::` and IMO it's good form to leave an empty line between the ``.. note::` directive and the actual content.
1. There is no such thing a `gcloud` any more, not sure where this text comes from.
1. Please use ID (not "id")

datastore/tests/unit/test_entity.py
chemelnucfin: I was just following what was there previously.

datastore/tests/unit/test_entity.py
chemelnucfin: It was in the constructor test.  I can add one I guess.

datastore/tests/unit/test_entity.py
dhermes: Great. Does my approach seem logical?

datastore/tests/unit/test_entity.py
dhermes: Please do.

datastore/google/cloud/datastore/entity.py
chemelnucfin: I will follow up with the documentation changes for both on a separate PR.  It would be inconsistent to change one and not the other and it would detract from this PR if I sent it together.

dhermes: Just change this part and feel free to send a follow-up PR to update the other part.

chemelnucfin: See #4641.  I changed them both there.

PR #4642, opened 2017-12-20
Requiring 'grpcio >= 1.8.2'.

dhermes:
This is due to a nasty spinlock [bug][1] that has been partially [fixed][2] in `1.8.2`.

[1]: https://github.com/grpc/grpc/issues/9688
[2]: https://github.com/grpc/grpc/pull/13665

### DIFF
--- a/pubsub/setup.py
+++ b/pubsub/setup.py
@@ -52,6 +52,7 @@ SETUP_BASE = {
 
 REQUIREMENTS = [
     'google-api-core[grpc] >= 0.1.2, < 0.2.0dev',
+    'grpcio >= 1.8.2',
     'google-auth >= 1.0.2, < 2.0dev',
     'grpc-google-iam-v1 >= 0.11.1, < 0.12dev',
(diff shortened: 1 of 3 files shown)


pubsub/setup.py
```diff
@@ -52,6 +52,7 @@
 
 REQUIREMENTS = [
     'google-api-core[grpc] >= 0.1.2, < 0.2.0dev',
+    'grpcio >= 1.8.2',
```
dhermes: @lukesneeringer How bad an idea is it to put this requirement here? Should I just cut a new release of `google-api-core` rather than depending on "future me" to remove this?

lukesneeringer: If the pub/sub library explicitly imports `grpcio`, I would actually prefer the requirement be explicitly stated.

dhermes:
I'm going to merge this, but hope to remove the explicit `grpcio` requirement in `pubsub` eventually.

PR #4643, opened 2017-12-20
Allowed a custom `Event` type in Pub / Sub futures.

dhermes:


pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -29,18 +29,28 @@ class Future(google.api_core.future.Future):
...
+            :class:`threading.Event`.
     """
 
     # This could be a sentinel object or None, but the sentinel object's ID
     # can change if the process is forked, and None has the possibility of
     # actually being a result.
     _SENTINEL = uuid.uuid4()
 
-    def __init__(self):
+    def __init__(self, event_factory=threading.Event):
```
dhermes: Worth noting that doing it like this (vs. `event_factory=None` and then doing an `if event_factor is None` check) makes it much more difficult to mock `threading.Event`.

pubsub/google/cloud/pubsub_v1/subscriber/futures.py
```diff
@@ -29,10 +29,18 @@ class Future(futures.Future):
...
+            to take no arguments and return an event with the same interface as
+            :class:`threading.Event`. This is provided so that callers
+            with different concurrency models (e.g. ``threading`` or
+            ``multiprocessing``) can wait on an event that is compatible
+            with that model. The ``wait()`` and ``set()`` methods will be
+            used on the returned event. The default is
+            :class:`threading.Event`.
     """
-    def __init__(self, policy):
+    def __init__(self, policy, *args, **kwargs):
```
dhermes: @lukesneeringer Would you rather I use `*a, **k` here or just be explicit with the `event_factory` argument? It may make more sense to make `event_factory=None` the default in the latter case.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -263,7 +263,7 @@ def open(self, callback):
         # Create the Future that this method will return.
         # This future is the main thread's interface to handle exceptions,
         # block on the subscription, etc.
-        self._future = Future(policy=self)
+        self._future = Future(policy=self, event_factory=threading.Event)
```
dhermes: @lukesneeringer I was being **intentionally** explicit here, i.e. I want the `threading` implementation to be clear it is using a `threading.Event`

pubsub/google/cloud/pubsub_v1/subscriber/futures.py
lukesneeringer: Will every subclass need `event_factory`?

dhermes: Well the subclass needs to be able to pass it through or it's not a useful argument. The point is that one caller could create a `Future(threading.Event)` and another could pass a `Future(multiprocessing.Event)` (or likely `manager.Event`).

(In other words, this isn't publisher or subscriber specific, it's concurrency model specific.)

dhermes: FWIW I struggled with three choices for that input:

- `completed` (`None` or an already instantiated `Event`)
- `event_class` (a class with a constructor that takes no arguments)
- `event_factory` (alternate name: `make_event`)

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -29,18 +29,29 @@ class Future(google.api_core.future.Future):
...
+            :class:`threading.Event` will be created and used.
     """
 
     # This could be a sentinel object or None, but the sentinel object's ID
     # can change if the process is forked, and None has the possibility of
     # actually being a result.
     _SENTINEL = uuid.uuid4()
 
-    def __init__(self):
+    def __init__(self, completed=None):
```
dhermes: @lukesneeringer I'm tempted to just make `completed` a required argument. WDYT?

lukesneeringer: I think I prefer it as-is.

PR #4646, opened 2017-12-20
nox unittest updates

chemelnucfin:
Manually changed.  Where do I change the generator?
@dhermes @lukesneeringer

### DIFF
--- a/bigquery_datatransfer/nox.py
+++ b/bigquery_datatransfer/nox.py
@@ -13,24 +13,53 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+
 import os
 
 import nox
 
 
+LOCAL_DEPS = (
+    os.path.join('..', 'api_core'),
+)
+
+
 @nox.session
-@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
-def unit_tests(session, python_version):
-    """Run the unit test suite."""
+def default(session):
+    """Default unit test session.
 
-    session.interpreter = 'python{}'.format(python_version)
-
-    session.virtualenv_dirname = 'unit-' + python_version
-
-    session.install('pytest')
+    This is intended to be run **without** an interpreter set, so
+    that the current ``python`` (on the ``PATH``) or the version of
+    Python corresponding to the ``nox`` binary on the ``PATH`` can
+    run the tests.
+    """
+    # Install all test dependencies, then install this package in-place.
+    session.install('mock', 'pytest', 'pytest-cov', *LOCAL_DEPS)
     session.install('-e', '.')
 
-    session.run('py.test', '--quiet', os.path.join('tests', 'unit'))
+    # Run py.test against the unit tests.
+    session.run(
+        'py.test',
+        '--quiet',
+        '--cov=google.cloud.bigquery_datatransfer',
+        '--cov=google.cloud.bigquery_datatransfer_v1',        
+        '--cov=tests.unit',
+        '--cov-append',
+        '--cov-config=.coveragerc',
+        '--cov-report=',
+        os.path.join('tests', 'unit', 'gapic', 'v1'),
+        *session.posargs
+    )
+
+
+@nox.session
+@nox.parametrize('py', ['2.7', '3.4', '3.5', '3.6'])
+def unit(session, py):
+    """Run the unit test suite."""
+
+    session.interpreter = 'python{}'.format(py)
+    session.virtualenv_dirname = 'unit-' + py
+    default(session)
 
 
 @nox.session
--- a/container/nox.py
+++ b/container/nox.py
@@ -13,29 +13,58 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+
 import os
 
 import nox
 
 
+LOCAL_DEPS = (
+    os.path.join('..', 'api_core'),
+)
+
+
 @nox.session
-@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
-def unit_tests(session, python_version):
-    """Run the unit test suite."""
+def default(session):
+    """Default unit test session.
 
-    session.interpreter = 'python{}'.format(python_version)
-
-    session.virtualenv_dirname = 'unit-' + python_version
-
-    session.install('pytest')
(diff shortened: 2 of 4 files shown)


bigquery_datatransfer/nox.py
```diff
@@ -19,8 +19,36 @@
 
 
 @nox.session
-@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
-def unit_tests(session, python_version):
+def default(session):
```
dhermes: @chemelnucfin Can you reduce the copy-paste and make `unit` call `default()`?

bigquery_datatransfer/nox.py
```diff
@@ -13,24 +13,54 @@
...
-@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
-def unit_tests(session, python_version):
-    """Run the unit test suite."""
+def default(session):
+    """Default unit test session.
+
+    This is intended to be run **without** an interpreter set, so
+    that the current ``python`` (on the ``PATH``) or the version of
+    Python corresponding to the ``nox`` binary the ``PATH`` can
+    run the tests.
```
dhermes: This docstring reminds of why `default` exists to begin with: for running on AppVeyor.

@chemelnucfin Can you add the 3 new APIs to `appveyor.yml` as well?

bigquery_datatransfer/nox.py
```diff
@@ -65,3 +95,18 @@ def system(session, py):
 
     # Run py.test against the system tests.
     session.run('py.test', '--quiet', 'tests/system/')
+
+
+@nox.session
+def cover(session):
```
dhermes: @lukesneeringer Is there a reason `cover` commands aren't generated? Is this intentional or an oversight or some 3rd option?

bigquery_datatransfer/nox.py
```diff
@@ -13,24 +13,54 @@
...
 from __future__ import absolute_import
+
 import os
 
 import nox
 
 
+LOCAL_DEPS = (
+    os.path.join('..', 'api_core'),
+    os.path.join('..', 'core'),
```
dhermes: @chemelnucfin If you check out `bigquery_datatransfer/setup.py`, you'll see only `google-api-core` as a dependency, so `../core` doesn't belong in this list.

container/nox.py
```diff
@@ -13,40 +13,70 @@
...
 from __future__ import absolute_import
+
 import os
 
 import nox
 
 
+LOCAL_DEPS = (
+    os.path.join('..', 'api_core'),
+    os.path.join('..', 'core'),
```
dhermes: @chemelnucfin If you check out `container/setup.py`, you'll see only `google-api-core` as a dependency, so `../core` doesn't belong in this list.

dataproc/nox.py
```diff
@@ -13,37 +13,67 @@
...
 from __future__ import absolute_import
+
 import os
 
 import nox
 
 
+LOCAL_DEPS = (
+    os.path.join('..', 'api_core'),
+    os.path.join('..', 'core'),
```
dhermes: @chemelnucfin If you check out `dataproc/setup.py`, you'll see only `google-api-core` as a dependency, so `../core` doesn't belong in this list.

bigquery_datatransfer/nox.py
```diff
@@ -65,3 +95,18 @@ def system(session, py):
...
+    """Run the final coverage report.
+
+    This outputs the coverage report aggregating coverage from the unit
+    test runs (not system test runs), and then erases coverage data.
+    """
+    session.interpreter = 'python3.6'
+    session.install('coverage', 'pytest-cov')
+    session.run('coverage', 'report', '--show-missing', '--fail-under=70')
+    # uncomment when coverage is met
+    # session.run('coverage', 'report', '--show-missing', '--fail-under=100')
```
dhermes: @chemelnucfin I'm inclined to just check in a failing `nox -s cover` since there is no "automatic" way for us to follow-up on this when / if coverage gets higher. But maybe @lukesneeringer has other ideas.

chemelnucfin: It would stall the builds, wouldn't it?

I think eventually someone will see that it's not high enough?

dhermes: How do you mean stall? It will fail the builds.

> eventually someone will see that it's not high enough

Can you elaborate? These config files are essentially set it and forget it (i.e. no one may ever look at that line again).

chemelnucfin: That's what I mean, it will fail the builds until coverage gets to 100.  Which means we can't tell if other tests are successful or not.

It's not any worse than it is now.  If someone increases coverage, great.  If not, it's what we have.

bigquery_datatransfer/nox.py
lukesneeringer: Mostly oversight.

bigquery_datatransfer/nox.py
lukesneeringer: Can we leave a `cover` off for now and I will open a bug to correct in autogen (both the coverage itself and the nox item)?

lukesneeringer: https://github.com/googleapis/toolkit/issues/1768

bigquery_datatransfer/nox.py
lukesneeringer: Fixed this in autogen: googleapis/toolkit#1770.

bigquery_datatransfer/nox.py
chemelnucfin: removed.

container/nox.py
```diff
@@ -13,40 +13,69 @@
...
 @nox.session
-@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
-def unit_tests(session, python_version):
-    """Run the unit test suite."""
+def default(session):
+    """Default unit test session.
+
+    This is intended to be run **without** an interpreter set, so
+    that the current ``python`` (on the ``PATH``) or the version of
+    Python corresponding to the ``nox`` binary the ``PATH`` can
```
dhermes: Should this say "on the ``PATH``"?

PR #4653, opened 2017-12-22
BigQuery: deprecate list_dataset_tables in favor of list_tables

tswast:
`list_dataset_tables` was needlessly verbose. Feedback from BQ GA review.

Even though we aren't GA yet, I deprecate rather than remove because we've just done a ton of breaking changes with 0.28.0 release and I'd rather not increase the fatigue of upgrading versions needlessly.

### DIFF
--- a/bigquery/CHANGELOG.md
+++ b/bigquery/CHANGELOG.md
@@ -4,6 +4,39 @@
 
 [1]: https://pypi.org/project/google-cloud-bigquery/#history
 
+## 0.29.0 (unreleased)
+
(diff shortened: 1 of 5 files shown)


tswast:
Note to self: wait to merge until I've double checked that the snippets at the migration guide point to a commit hash.

tswast:
Ha. Well, I didn't actually add a sample for listing tables at https://cloud.google.com/bigquery/docs/python-client-migration#table_code_changes We can merge this to our heart's content.

Issue #4111, opened 2017-10-03
firestore.CreateIfMissingOption should be replaced


dhermes:
Paraphrased from @samtstern:

It's still possible to do

```python
    city_ref.update({
        u'capital': True
    }, firestore.CreateIfMissingOption(True))
```

The other Firestore SDKs have removed this option and moved to an option on the `set()` operation called, `merge()`.  For example, in the Android SDK:

```java
db.collection("cities").document("BJ")
        .set(data, SetOptions.merge());
```

The new semantics make it so that the `set()` operation (and `create()` which some SDKs offer) is the only operation which can make a new document.  The `merge()` option means that the data will be deep merged into any existing document, rather than just obliterating and replacing.

samtstern:
A list of things that need to change:

  * Total removal of `firestore.CreateIfMissingOption()`, which will require changing the signature of the `update()` methods for document references, batched writes, and transactions.
  * Addition of a new option for set operations called `MergeOption` or similar, which will require changing the signature of the `set()` methods for document references, batched writes and transactions.

In spirit, users will be able to use the "set with merge" operation wherever they previously used the "update and create if missing", however there are a few key differences.

Update operations on `key:value` data where the `key` is a fieldPath.  Set operates on object-like data.  So take the following document:

```
{
  name: Sam
  address: {
    city: 'SF',
    state: 'CA'
  }
}
```

If I want to update only my city to `LA` and create the document if it does not exist, I would previously have done:

```python
    ref.update({
        u'address.city': u'LA'
    }, firestore.CreateIfMissingOption(True))
```

Now I would do:

```python
    ref.set({
        u'address': {
            u'city': u'LA'
        }
    }, firestore.MergeOption())
```

Notice that this is a **deep merge** meaning that if the `address.state` field was previously set it is left intact.  The incorrect behavior would be replacing the entire `address` field with `{ city: 'LA' }`.

dhermes:
@jba FWIW @samtstern and I had previously discussed a way to map out what features are missing before starting the work on updating the API surface. As of now, this is the only "known" work to be done.

schmidt-sebastian:
@dhermes Can you give us an estimate for the completion of this change?

jba:
Does Python handle set/merge with deletes?

There are also the conformance tests. I started work on this (https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4359) but can't get anyone to look at it.

[...]

PR #4654, opened 2017-12-22
Remove CreateIfMissing option

chemelnucfin:
Partial #4111

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1beta1/_helpers.py
@@ -47,10 +47,6 @@ FIELD_PATH_WRONG_TYPE = (
     'The data at {!r} is not a dictionary, so it cannot contain the key {!r}')
 FIELD_PATH_DELIMITER = '.'
 DOCUMENT_PATH_DELIMITER = '/'
-_NO_CREATE_TEMPLATE = (
-    'The ``create_if_missing`` option cannot be used '
-    'on ``{}()`` requests.')
-NO_CREATE_ON_DELETE = _NO_CREATE_TEMPLATE.format('delete')
 INACTIVE_TXN = (
     'Transaction not in progress, cannot be used in API requests.')
 READ_AFTER_WRITE_ERROR = 'Attempted read after write in a transaction.'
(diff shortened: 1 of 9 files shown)


firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -897,7 +893,7 @@ def pb_for_delete(document_path, option):
     """
     write_pb = write_pb2.Write(delete=document_path)
     if option is not None:
-        option.modify_write(write_pb, no_create_msg=NO_CREATE_ON_DELETE)
+        option.modify_write(write_pb)
```
dhermes: @chemelnucfin Why was this removed but not replaced with something equivalent?

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -252,10 +251,8 @@ def write_option(**kwargs):
         :meth:`~.DocumentReference.update` and
         :meth:`~.DocumentReference.delete`.
 
-        Exactly one of three keyword arguments must be provided:
+        One of the following two keyword arguments must be provided:
```
dhermes: Can you just remove `two` here? (`three` should've been removed from the previous description)

Reasoning: the count is clear from the bullets but is just one more thing we have to "maintain"

firestore/tests/unit/test__helpers.py
```diff
@@ -1216,6 +1217,9 @@ def _helper(self, option=None, do_transform=False, **write_kwargs):
         map_pb = document_pb2.MapValue(fields={
             'yum': _value_pb(bytes_value=value),
         })
+        if isinstance(option, ExistsOption):
+            write_kwargs = {'current_document' : {'exists': option._exists}}
```
dhermes: @chemelnucfin 

1. This should update ``write_kwargs`` but should not discard the previous values.
1. The value should be a protobuf, not a dictionary

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: That message is only good for create_if_missing.

firestore/tests/unit/test__helpers.py
chemelnucfin: would an update work on write_args?

the value gets used in the protobuf below I think?

chemelnucfin:
Closing this pull request

The commits are in #4851 and can be reviewed commit by commit.

PR #4657, opened 2017-12-22
BigQuery: Add Client.insert_rows, deprecate Client.create_rows

tswast:
`insert_rows` aligns better with API request (Tabledata.insertAll). Feedback from BQ GA review.

Note that the [`create_rows()` sample in the Python migration guide](https://cloud.google.com/bigquery/docs/python-client-migration#streaming_data_into_a_table) already points to a commit hash, so this should be safe to merge whenever.

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -124,9 +124,9 @@ class TestBigQuery(unittest.TestCase):
         for doomed in self.to_delete:
             if isinstance(doomed, Bucket):
                 retry_409(doomed.delete)(force=True)
-            elif isinstance(doomed, Dataset):
+            elif isinstance(doomed, (Dataset, bigquery.DatasetReference)):
                 retry_in_use(Config.CLIENT.delete_dataset)(doomed)
-            elif isinstance(doomed, Table):
+            elif isinstance(doomed, (Table, bigquery.TableReference)):
                 retry_in_use(Config.CLIENT.delete_table)(doomed)
             else:
                 doomed.delete()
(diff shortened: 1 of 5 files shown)


bigquery/tests/system.py
```diff
@@ -1467,19 +1467,24 @@ def test_nested_table_to_dataframe(self):
...
+        job_config = bigquery.LoadJobConfig()
+        job_config.write_disposition = 'WRITE_TRUNCATE'
+        job_config.source_format = 'NEWLINE_DELIMITED_JSON'
+        job_config.schema = schema
+        # Load a table using a local JSON file from memory.
+        Config.CLIENT.load_table_from_file(
+            body, table, job_config=job_config).result()
+
+        df = Config.CLIENT.list_rows(
+            table, selected_fields=schema).to_dataframe()
```
tseaver: What drives such a big change to this testcase?  Wouldn't replacing `create_rows` with `insert_rows` just work?

docs/bigquery/snippets.py
```diff
@@ -600,9 +601,20 @@ def test_extract_table(client, to_delete):
...
+    rows = [json.dumps(row) for row in to_insert]
+    body = six.StringIO('{}\n'.format('\n'.join(rows)))
+    job_config = bigquery.LoadJobConfig()
+    job_config.write_disposition = 'WRITE_TRUNCATE'
+    job_config.source_format = 'NEWLINE_DELIMITED_JSON'
+    job_config.schema = SCHEMA
+    to_delete.insert(0, table_ref)
+    # Load a table using a local JSON file from memory.
+    client.load_table_from_file(
+        body, table_ref, job_config=job_config).result()
```
tseaver: Similar concern here.

bigquery/tests/system.py
tswast: In short, the reason is to reduce unnecessary test flakiness.

It would have continued to work as well as it had been, but `create_rows` / `insert_rows` is not intended for this use case (writing rows to a just-created empty table). After you create a table there is a time period (up to 2 minutes) where the streaming API might not be able to write to it due to some internal eventual consistency / cacheing. A load job does not have these same concerns.

tseaver: OK, thanks.  In a future PR, it might be good to note the "I found this other thing while renaming" bit to the description.

Issue #4656, opened 2017-12-22
BigTable: row.commit throwing _Rendezvous exception instead of ServiceUnavailable


zakons:
We are using retry.py and creating a retry exception predicate as follows:

predicate=retry.if_exception_type(exceptions.ServiceUnavailable)

However, since we are only getting _Rendezvous exceptions with StatusCode.UNAVAILABLE (see below stack trace), we had to change this to:

predicate=retry.if_exception_type(exceptions.GrpcRendezvous)

I see a different issue in cloud datastore for mapping gRPC codes to service level exceptions: 

https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2590

Will this also be done for cloud bigtable?

Stack Trace:

  File "single_key_read_write.py", line 135, in write
    row.commit()
  File "/home/jenkins/.local/lib/python3.5/site-packages/google/cloud/bigtable/row.py", line 445, in commit
    retry_(retry_commit)()
  File "/home/jenkins/.local/lib/python3.5/site-packages/google/api_core/retry.py", line 261, in retry_wrapped_func
    on_error=on_error,
  File "/home/jenkins/.local/lib/python3.5/site-packages/google/api_core/retry.py", line 178, in retry_target
    return target()
  File "/home/jenkins/.local/lib/python3.5/site-packages/google/cloud/bigtable/row.py", line 256, in __call__
    client._data_stub.MutateRow(self.request_pb)
  File "/home/jenkins/.local/lib/python3.5/site-packages/grpc/_channel.py", line 492, in __call__
    return _end_unary_response_blocking(state, call, False, deadline)
  File "/home/jenkins/.local/lib/python3.5/site-packages/grpc/_channel.py", line 440, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, Transport closed)>

# Thank you for reporting an issue to google-cloud-python!

If you are reporting an issue or requesting a feature, please search the
existing open and closed issues to see if there is already work being done.

- https://github.com/GoogleCloudPlatform/google-cloud-python/issues
- http://stackoverflow.com/questions/tagged/google-cloud-python

If you can provide us with as much of the following information as possible it
will help us identify the cause of your issue more quickly.

1. Specify the API at the beginning of the title (for example, "BigQuery: ...")
   General, Core, and Other are also allowed as types
[...]

PR #4665, opened 2017-12-27
refactor retry logic on mutate row

aneepct:
Added retry logic for commit on a DirectRow. The retry uses the retry.py module and performs retry when a Rendezvous exception is thrown due to a disconnect on the channel. The retry will call MutateRow on the original protobuf request with a timeout backoff algorithm provided in retry.py. This has been tested using an endurance test framework that runs for 23 hours and was receiving 8+ exceptions previously. Now all the retries are successful.

### DIFF
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -19,8 +19,6 @@ import struct
 
 import six
 
-from concurrent import futures
-
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _microseconds_from_datetime
 from google.cloud._helpers import _to_bytes
@@ -437,13 +435,7 @@ class DirectRow(_SetDeleteRow):
         retry_ = retry.Retry(
             predicate=retry.if_exception_type(exceptions.GrpcRendezvous),
             deadline=30)
-
-        try:
-            retry_(retry_commit)()
-        except exceptions.RetryError:
-            raise futures.\
-                TimeoutError('Operation did not complete '
-                             'within the designated timeout.')
+        retry_(retry_commit)()
 
         self.clear()
 
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -239,13 +239,6 @@ class _SetDeleteRow(Row):
             mutations_list.extend(to_append)
 
 
-def wrapped_partial(func, *args, **kwargs):
-    """Create partial wrapper for a function"""
-    partial_func = functools.partial(func, *args, **kwargs)
-    functools.update_wrapper(partial_func, func)
-    return partial_func
-
-
 def call_mutate_row(table, request_pb):
     """Call the MutateRow"""
     client = table._instance._client
@@ -429,9 +422,10 @@ class DirectRow(_SetDeleteRow):
             mutations=mutations_list,
         )
 
-        retry_commit = wrapped_partial(call_mutate_row,
-                                       self._table,
-                                       request_pb)
+        retry_commit = functools.partial(
+            call_mutate_row,
+            table=self._table,
+            request_pb=request_pb)
         retry_ = retry.Retry(
             predicate=retry.if_exception_type(exceptions.GrpcRendezvous),
             deadline=30)
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -431,7 +431,9 @@ class DirectRow(_SetDeleteRow):
             mutations=mutations_list,
         )
 
-        retry_commit = wrapped_partial(call_mutate_row, self._table, request_pb)
+        retry_commit = wrapped_partial(call_mutate_row,
+                                       self._table,
+                                       request_pb)
         retry_ = retry.Retry(
             predicate=retry.if_exception_type(exceptions.GrpcRendezvous),
             deadline=30)
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -17,8 +17,8 @@
 
 import struct
 
-import six
 import functools
+import six
 import grpc
 
 from google.cloud._helpers import _datetime_from_microseconds
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -18,6 +18,7 @@
 import struct
 
 import six
+import functools
 
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _microseconds_from_datetime
@@ -28,20 +29,12 @@ from google.cloud.bigtable._generated import (
     bigtable_pb2 as messages_v2_pb2)
 from google.api_core import retry
 from google.cloud import exceptions
-from functools import partial, update_wrapper
 
 
 _PACK_I64 = struct.Struct('>q').pack
 
 MAX_MUTATIONS = 100000
 """The maximum number of mutations that a row can accumulate."""
-
-
-def wrapped_partial(func, *args, **kwargs):
-    """Create partial wrapper for a function"""
-    partial_func = partial(func, *args, **kwargs)
-    update_wrapper(partial_func, func)
-    return partial_func
 
 
 class Row(object):
@@ -244,6 +237,13 @@ class _SetDeleteRow(Row):
             # We don't add the mutations until all columns have been
             # processed without error.
             mutations_list.extend(to_append)
+
+
+def wrapped_partial(func, *args, **kwargs):
+    """Create partial wrapper for a function"""
+    partial_func = functools.partial(func, *args, **kwargs)
+    functools.update_wrapper(partial_func, func)
+    return partial_func
 
 
 def call_mutate_row(table, request_pb):
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -439,8 +439,9 @@ class DirectRow(_SetDeleteRow):
         try:
             retry_(retry_commit)()
         except exceptions.RetryError:
-            raise futures.TimeoutError('Operation did not complete '
-                                       'within the designated timeout.')
+            raise futures.\
+                TimeoutError('Operation did not complete '
+                             'within the designated timeout.')
 
         self.clear()
 
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -19,6 +19,7 @@ import struct
 
 import six
 import functools
+import grpc
 
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _microseconds_from_datetime
@@ -28,7 +29,7 @@ from google.cloud.bigtable._generated import (
 from google.cloud.bigtable._generated import (
     bigtable_pb2 as messages_v2_pb2)
 from google.api_core import retry
-from google.cloud import exceptions
+from google.api_core import exceptions
 
 
 _PACK_I64 = struct.Struct('>q').pack
@@ -239,8 +240,13 @@ class _SetDeleteRow(Row):
             mutations_list.extend(to_append)
 
 
-def call_mutate_row(table, request_pb):
-    """Call the MutateRow"""
+def _retry_commit_exception(exc):
+    if isinstance(exc, grpc.RpcError):
+        exc = exceptions.from_grpc_error(exc)
+    return isinstance(exc, exceptions.ServiceUnavailable)
+
+
+def _call_mutate_row(table, request_pb):
     client = table._instance._client
     client._data_stub.MutateRow(request_pb)
 
@@ -423,11 +429,11 @@ class DirectRow(_SetDeleteRow):
         )
 
         retry_commit = functools.partial(
-            call_mutate_row,
+            _call_mutate_row,
             table=self._table,
             request_pb=request_pb)
         retry_ = retry.Retry(
-            predicate=retry.if_exception_type(exceptions.GrpcRendezvous),
+            predicate=_retry_commit_exception,
             deadline=30)
         retry_(retry_commit)()
 
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -15,12 +15,14 @@
 """User-friendly container for Google Cloud Bigtable Row."""
 
 
+import functools
 import struct
 
-import functools
-import six
 import grpc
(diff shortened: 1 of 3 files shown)


sduskis:
Can you please add a description for the pull request?

bigtable/google/cloud/bigtable/row.py
```diff
@@ -26,6 +26,9 @@
     data_pb2 as data_v2_pb2)
 from google.cloud.bigtable._generated import (
     bigtable_pb2 as messages_v2_pb2)
+from google.api_core import retry
+from google.cloud import exceptions
+from functools import partial, update_wrapper
```
jonparrott: please just import `functools`.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -34,6 +37,13 @@
 """The maximum number of mutations that a row can accumulate."""
 
 
+def wrapped_partial(func, *args, **kwargs):
```
jonparrott: Can you explain why this is needed?

bigtable/google/cloud/bigtable/row.py
```diff
@@ -412,9 +428,15 @@ def commit(self):
...
         )
-        # We expect a `google.protobuf.empty_pb2.Empty`
-        client = self._table._instance._client
-        client._data_stub.MutateRow(request_pb)
+
+        retry_commit = wrapped_partial(call_mutate_row,
+                                       self._table,
+                                       request_pb)
+        retry_ = retry.Retry(
+            predicate=retry.if_exception_type(exceptions.GrpcRendezvous),
```
jonparrott: We will at some point remove the `GrpcRendezvous` alias, better to catch `grpc.RpcError`.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -236,6 +246,12 @@ def _delete_cells(self, column_family_id, columns, time_range=None,
             mutations_list.extend(to_append)
 
 
+def call_mutate_row(table, request_pb):
```
jonparrott: Couldn't you just use `self._table._instance._client._data_stub.MutateRow` directly below?

jonparrott:
@aneepct this PR has open comments and little activity, do you still intend to merge this?

bigtable/google/cloud/bigtable/row.py
zakons: Please note that exceptions.GrpcRendezvous is the actual exception we are receiving, so we cannot change this and effectively test.

Also, please see the original issue #4656, where the goal was to map underlying Rendezvous exceptions to exceptions.ServiceUnavailable.  So far, we have been able to catch the following exception with our retry predicate:

```
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, Transport closed)
```
We thought there was a reorganization of the error handling code in progress and were waiting for that.  Has this been done in the following pull request: #4743? If so, where would we find the new exceptions that would include the proper mapping to exceptions.ServiceUnavailable?  Your help here would be much appreciated, since it seems like the right exception for our predicate.

aneepct:
I am actively resolving all the review questions now and will advance to the state needed for merging.

bigtable/google/cloud/bigtable/row.py
aneepct: The Retry logic requires a callable method. By using a wrapped_partial we also get the encapsulation of the arguments for the retry method, call_mutate_row. This functionality is provided by functools.

bigtable/google/cloud/bigtable/row.py
aneepct: Change committed

bigtable/google/cloud/bigtable/row.py
aneepct: Please see the reply above about the use of a callable wrapped partial.

zakons:
All questions / requests have been responded to.  If it is not possible in the current release to get the mapping to exceptions.ServiceUnavailable from the _Rendezvous exception (see comment above), this could be done later when that mapping is available. Can we move forward with the merge?

bigtable/google/cloud/bigtable/row.py
jonparrott: Let me be more clear: Yes, the concrete type you are catching is `grpc._channel._Rendezvous` which out of misunderstanding we previously aliased to `exceptions.GrpcRendezvous`, however, that is a private concrete type and we should not refer to it in our code. That class is an concrete implementation of `grpc.RpcError` (a public abstract type) and *that* is what you should be catching.

Also you should not indiscriminately retry on *every* RPC error, some of them are *not* idempotent.  Your predicate should invoke `google.api_core.exceptions.from_grpc_error` and ensure the exception is `ServiceUnavailable`.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -236,6 +239,19 @@ def _delete_cells(self, column_family_id, columns, time_range=None,
...
 
 
+def wrapped_partial(func, *args, **kwargs):
+    """Create partial wrapper for a function"""
+    partial_func = functools.partial(func, *args, **kwargs)
+    functools.update_wrapper(partial_func, func)
+    return partial_func
+
+
+def call_mutate_row(table, request_pb):
```
jonparrott: please make this private.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -236,6 +239,19 @@ def _delete_cells(self, column_family_id, columns, time_range=None,
             mutations_list.extend(to_append)
 
 
+def wrapped_partial(func, *args, **kwargs):
```
jonparrott: Other APIs have been happy with just using `functools.partial` and not needing to also use `update_wrapper`. Can you justify why this is needed here over just using `partial`?

core/google/cloud/exceptions.py
```diff
@@ -55,5 +55,6 @@
 BadGateway = exceptions.BadGateway
 ServiceUnavailable = exceptions.ServiceUnavailable
 GatewayTimeout = exceptions.GatewayTimeout
+RetryError = exceptions.RetryError
```
jonparrott: Please do not add new aliases to this file (I am eventually going to deprecate this)

zakons: Thanks. Almost there.  Just not sure how to do the checking for ServiceUnavailable (the exception that we want).  Here is the retry logic with the predicate you are recommending:

```
retry_commit = functools.partial(
    _call_mutate_row, 
     table=self._table, 
     request_pb=request_pb)

retry_ = retry.Retry(
    predicate=retry.if_exception_type(exceptions.from_grpc_error),
    deadline=30)

retry_(retry_commit)()
```
Where does the checking for exceptions.ServiceUnavailable go?

zakons:
Thanks. Almost there.  Just not sure how to do the checking for ServiceUnavailable (the exception that we want).  Here is the retry logic with the predicate you are recommending:

```
retry_commit = functools.partial(
    _call_mutate_row, 
     table=self._table, 
     request_pb=request_pb)

retry_ = retry.Retry(
    predicate=retry.if_exception_type(exceptions.from_grpc_error),
    deadline=30)

retry_(retry_commit)()
```
Where does the checking for exceptions.ServiceUnavailable go?

jonparrott:
@zakons you'll need to basically write a custom predicate function:

```python
def _should_retry_commit_exception(exc):
    if isinstance(exc, grpc.RpcError):
        exc = exceptions.from_grpc_error(exc)
    return isinstance(exc, exceptions.ServiceUnavailable)

...

retry_commit = functools.partial(
    _call_mutate_row,
    table=self._table,
    request_pb=request_pb)

retry_ = retry.Retry(
    predicate=_should_retry_commit_exception,
    deadline=30)

retry_(retry_commit)()
```

core/google/cloud/exceptions.py
zakons: @jonparrott @sduskis The requested changes were made but there were two issues: 1) call_mutate_row cannot be made private since it is then not visible in the namespace needed by the functools.partial code, 2) there is NO mapping from exceptions.GrpcRendezvous to exceptions.ServiceUnavailable, which is the original problem we identified, and therefore exceptions.from_grpc_error does not create exceptions.ServiceUnavailable, so we have had to revert to retry on Rendezvous exceptions. Please see last commit.

It does seem too complex to have to add a custom predicate and also seems like too much deep knowledge required to know we have to remap the Rendezvous exception to ServiceUnavailable ourselves.  That is why we were really hoping from the beginning that this mapping would have been done by the layers below us.  Even if this were done in those layers manually for this one requested mapping, that would be a big help.

jonparrott: > 1) call_mutate_row cannot be made private since it is then not visible in the namespace needed by the functools.partial code

I don't understand, you can absolutely bind partial to a private method, e.g.:

```python
retry_commit = functools.partial(
   _call_mutate_row,
   ...)
```

> 2) there is NO mapping from exceptions.GrpcRendezvous to exceptions.ServiceUnavailable, which is the original problem we identified, and therefore exceptions.from_grpc_error does not create exceptions.ServiceUnavailable, so we have had to revert to retry on Rendezvous exceptions. Please see last commit.

Yes there is, it's right here: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/7714d0cfbbf2da6aff45daef3bba8443927afd6f/api_core/google/api_core/exceptions.py#L313

> It does seem too complex to have to add a custom predicate and also seems like too much deep knowledge required to know we have to remap the Rendezvous exception to ServiceUnavailable ourselves.

It's really not, other clients do this as well.

> That is why we were really hoping from the beginning that this mapping would have been done by the layers below us. Even if this were done in those layers manually for this one requested mapping, that would be a big help.

Once we get the autogen update for bigtable in, there will be automatic mapping to exception classes and automatic retries. This PR may not even been needed in that case.

zakons: @jonparrott We made a unit test case to step through the exception mapping code and it now seems to work.  Your help is very much appreciated.  Running our endurance tests now. If they handle the predicate, then I think the latest commit works and satisfies your review. Thanks!

aneepct:
@tseaver @jonparrot @sduskis The circleci coverage is now at 100% and our endurance test looks good, can we please have this PR merged?  Many thanks for your help.

tseaver:
LGTM if @jonparrott says go.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -17,7 +17,9 @@
 
 import struct
 
+import functools
```
jonparrott: functools goes in the first group of imports

bigtable/google/cloud/bigtable/row.py
```diff
@@ -17,7 +17,9 @@
...
 import six
+import grpc
```
jonparrott: grpc goes before six

imports are in three groups: standard library, third-party, and package-local. Each group should be in alphabetical order.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -26,6 +28,8 @@
     data_pb2 as data_v2_pb2)
 from google.cloud.bigtable._generated import (
     bigtable_pb2 as messages_v2_pb2)
+from google.api_core import retry
```
jonparrott: `google.api_core` should be placed before `google.cloud` imports

bigtable/google/cloud/bigtable/row.py
```diff
@@ -236,6 +240,17 @@ def _delete_cells(self, column_family_id, columns, time_range=None,
             mutations_list.extend(to_append)
 
 
+def _retry_commit_exception(exc):
+    if isinstance(exc, grpc.RpcError):
+        exc = exceptions.from_grpc_error(exc)
+    return isinstance(exc, exceptions.ServiceUnavailable)
+
+
+def _call_mutate_row(table, request_pb):
```
jonparrott: Can you eliminate this and make the call to `functools.partial` below be:

```python
commit = functools.partial(
    self._table._instance._client._data_stub.MutateRow,
    request_pb)
```

bigtable/google/cloud/bigtable/row.py
```diff
@@ -412,9 +427,16 @@ def commit(self):
             row_key=self._row_key,
             mutations=mutations_list,
         )
-        # We expect a `google.protobuf.empty_pb2.Empty`
-        client = self._table._instance._client
-        client._data_stub.MutateRow(request_pb)
+
+        retry_commit = functools.partial(
```
jonparrott: It makes more sense for this to just be called `commit` not `retry_commit` as it doesn't in itself retry. Then, the application of `retry_` below reads `retry_(commit)()`.

bigtable/tests/unit/test_row.py
```diff
@@ -367,6 +367,27 @@ def test_commit(self):
...
+            condition = threading.Condition()
+            code = grpc.StatusCode.UNAVAILABLE
+            details = 'Endpoint read failed'
+
+        def _check_rendezvous_exception(exc):
+            return _retry_commit_exception(exc)
+
+        expected_result = True
+        result = _check_rendezvous_exception(
+            grpc._channel._Rendezvous(State, None, None, 0))
```
jonparrott: At no point should you be using private members in grpc. Create a mock using `grpc.Call/grpc.RpcError` as the spec, see https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/api_core/tests/unit/test_exceptions.py#L188

core/google/cloud/exceptions.py
jonparrott: This still needs to be removed.

aneepct:
@jonparrott How do i use RpcError and set the code to UNAVAILABLE?

aneepct:
@jonparrott  @zakons  OK, we are at 100% coverage and all of your review comments have been incorporated. Please review and if OK, please merge.

zakons:
@jonparrott @tseaver  Can we please merge this PR now that all requested changes have been applied?  Our endurance test results are all fine as well, meaning that the UNAVAILABLE exception is caught and the commit retried until success.

jonparrott:
@zakons @aneepct just a heads up - I'm traveling and will not have a chance to review this until later next week. Apologies for the delay, and thank you for your patience.

jonparrott:
Thanks everyone! Sorry for the protracted review process here, I'm glad we ended up with something short and sweet. :)

zakons:
@jonparrott Thank-you!  We were impressed with your reviews. In fact, I have been looking at your [python testing style guide](http://blog.jonparrott.com/my-python-testing-style-guide/) and am finding it very clear and helpful.

jonparrott:
Awesome, I'm glad to be helpful. :)

PR #4678, opened 2017-12-29
Update Spanner system test to use correct constant (copy-pasta bug).

chemelnucfin:


dhermes:
@chemelnucfin I've updated the PR description. Please take note.

chemelnucfin:
are all bugs worth mentioning?

dhermes:
@chemelnucfin It's always worth explaining what a PR does

Issue #1812, opened 2016-05-20
Bigtable `PartialRowsData` is not an iterable.


tswast:
It appears to me that if I want to scan through all rows of a Bigtable, [`Table.read_rows()`](https://googlecloudplatform.github.io/gcloud-python/stable/bigtable-table.html#gcloud.bigtable.table.Table.read_rows) is the function to use. This returns a [PartialRowsData](https://googlecloudplatform.github.io/gcloud-python/stable/bigtable-row-data.html#gcloud.bigtable.row_data.PartialRowsData).

Why is `PartialRowsData` not an iterable? I'd like to consume all the rows for a "hello bigtable" sample by iterating through them. Something like

```
print('Scan for all greetings:')
for row in table.read_rows():
    print(row.cells[COLUMN_FAMILY_NAME][COLUMN_NAME.encode('UTF-8')][0].decode('UTF-8'))
```

Instead, it appears that I need to use a while loop and watch a dictionary that gets populated. I think this means that even if I don't need to store all the rows in memory at once, they still do. That's both inefficient and not following Python idioms (of generators / iterables).

dhermes:
Taking a [peek at `PartialRowsData.consume_next()`][1], it's because of the nature of the Bigtable API. The actual stream returned from the API may have partial results for a row, hence we need to build it up to get a complete answer. I agree this is memory hungry and this was something I worried about (tried to track down old discussions, can't find any right now).

You don't need to use a while loop, you can just use

```python
partial_rows = table.read_rows()
partial_rows.consume_all()
```

Maybe we should make this clearer in the [Data API doc][2]?

[1]:https://github.com/GoogleCloudPlatform/gcloud-python/blob/8dd7fefeac9ecf08c10b4da1edb37caba16b71c6/gcloud/bigtable/row_data.py#L287-L302
[2]: https://googlecloudplatform.github.io/gcloud-python/stable/bigtable-data-api.html

tswast:
Yeah, it could could definitely be clearer in the docs. I'll see if I can improve them at all as I work on the bigtable "hello world".

I can imagine `consume_all()` being super memory hungry, so I would still like an easy way to consume in a specified batch size.

@dhermes What do you think about adding an `iterrows()` function/generator to `PartialRowsData` that yields a `(key, row)` tuple and deletes that row from `rows` after calling `consume_next()`?

dhermes:
[...]

PR #4679, opened 2017-12-29
BigTable: Adding a row generator on a table.

### LINKED ISSUE #4586: BigTable: Memory usage abnormally increasing when using table.read_rows method
Hi,

While using the python google-cloud-bigtable to read a large collection of data (10M to 90M rows of at least 4MB to 5MB per row), I noticed a strange behavior: The used RAM was at lease twice the size of the data I was reading from the table.

Here is the code I did to reproduce :

```python
from google.cloud import bigtable

instance_name = "the-instance"
table_name = "the-table"

client = bigtable.Client(admin=False)
instance = client.instance(instance_name)
table = instance.table(table_name)

last_readall_key = None
rows = table.read_rows(start_key=last_readall_key)

while True:
    try:
        rows.consume_next()
    except StopIteration as e:
        break
    except Exception as e:
        print 'Got exception of type %s : %s' % (type(e), str(e)) # mostly grpc._channel._Rendezvous exceptions are thrown here

        table = instance.table(table_name)
        rows = table.read_rows(start_key=last_readall_key)

        rows.consume_next()

    for key in rows.rows:
        last_readall_key = key

    rows._rows.clear() #this was an attempt to fix, but it doesn't have any effect
```

When running the above code, my instance RAM fills itself until full.

Another weird thing, is that the problem does not happen on my MacOS Sierra 10.12.6 machine. I only witness the problem on Linux machines.

I saw this problem on Linux Ubuntu 14.04 LTS and Linux Ubuntu 10.04.5 LTS. I did not try with other Linux distributions or others versions of Ubuntu though.

Am I doing something wrong ? Does anyone knows what may cause this problem ?
Would anyone have any trick to avoid this problem ?

Find below the pip requirements to reproduce :

```
cachetools==2.0.1
certifi==2017.11.5
chardet==3.0.4
dill==0.2.7.1
enum34==1.1.6
future==0.16.0
futures==3.2.0
google-api-core==0.1.2
google-auth==1.2.1
google-cloud-bigtable==0.28.1
google-cloud-core==0.28.0
google-gax==0.15.16
grpcio==1.7.3
idna==2.6
ply==3.8
protobuf==3.5.0.post1
pyasn1==0.4.2
pyasn1-modules==0.2.1
pytz==2017.3
requests==2.18.4
rsa==3.4.2
six==1.11.0
urllib3==1.22
```
(these dependancies are installed automatically when using `pip install google-cloud-bigtable` )

Thanks for your time.

aneepct:
Adding a row generator on a table. This will allow iteration over the rows in a table instead of reading the rows into an internal dictionary first. As soon as a row has been validated, it is available in the iterator, by adding a row generator on a table. This will allow iteration over the rows in a table instead of reading the rows into an internal dictionary first. As soon as a row has been validated, it is available in the iterator, by using yield in the generator. The read_rows() method on table will still return PartialRowsData, and consume_all() will now use the row generator to process the ReadRowsResponses to populate the rows or in an internal dictionary. The yield_rows() method on table provides a row iterator and does not store any of the rows internally. The rows are available as soon as they are read from the channel and validated.

Closes #4586.


bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
 
     def __init__(self, response_iterator):
         self._response_iterator = response_iterator
+        self.generator = YieldRowsData(response_iterator)
```
dhermes: Please rename as `self._generator`

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
...
+
         # Fully-processed rows, keyed by `row_key`
         self._rows = {}
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
```
dhermes: Please `return NotImplemented` here. See, e.g. `Entity.__eq__` in datastore

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/0b6df3d8ba76a9ed917ff94357a7fa7eb16a7582/datastore/google/cloud/datastore/entity.py#L154-L155

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
...
+        return other._response_iterator == self._response_iterator
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
dhermes: Please use `not self == other` so the Python `NotImplemented` fallbacks can work properly.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
...
+    @property
+    def rows(self):
+        """Property returning all rows accumulated from the stream.
+
+        :rtype: dict
+        :returns: row_key -> :class:`PartialRowData`.
+        """
+        # NOTE: To avoid duplicating large objects, this is just the
+        #       mutable private data.
+        return self._rows
```
dhermes: There is no point in having this property since `self._rows` is mutable (i.e. just upgrade the atrribute from `_rows` to `rows` if you want to do this)

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
...
+
+
+class YieldRowsData(object):
+    """Convenience wrapper for consuming a ``ReadRows`` streaming response.
+
+    :type response_iterator: :class:`~google.cloud.exceptions.GrpcRendezvous`
+    :param response_iterator: A streaming iterator returned from a
+                              ``ReadRows`` request.
+    """
+    START = "Start"  # No responses yet processed.
```
dhermes: 1. Please use single quotes (i.e. `'Start'`, `'New row'`, etc.)
1. Please leave an empty line between the docstring and the class members / constants

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -194,8 +194,66 @@ class PartialRowsData(object):
...
+        return other._response_iterator == self._response_iterator
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    @property
+    def state(self):
+        """State machine state.
+
+        :rtype: str
```
dhermes: Here and throughout the rest of the PR, feel free to use Google-style docstrings (vs. the traditional Sphinx-style here). This project is 3.5 years old and we've only switched to **allowing** Google-style docstrings for a year or so. In general, we "require" PRs from Googlers to use Google-style docstrings when adding new methods/functions/classes.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -425,18 +451,9 @@ def _copy_from_previous(self, cell):
                 cell.row_key = previous.row_key
             if not cell.family_name:
                 cell.family_name = previous.family_name
-            # NOTE: ``cell.qualifier`` **can** be empty string.
-            if cell.qualifier is None:
+            if not cell.qualifier:
```
dhermes: This is a reversion, see #4252.

Did you revert this on purpose or do you have a 2+ month old checkout of the codebase?

bigtable/tests/unit/test_row_data.py
```diff
@@ -189,117 +189,85 @@ def test_row_key_getter(self):
 
 
 class TestPartialRowsData(unittest.TestCase):
+    @staticmethod
```
dhermes: Please leave an empty line between the `class ...` line and the first method

bigtable/tests/unit/test_row_data.py
```diff
@@ -189,117 +189,85 @@ def test_row_key_getter(self):
...
+    def _get_yield_target_class():
+        from google.cloud.bigtable.row_data import YieldRowsData
+
+        return YieldRowsData
 
     @staticmethod
-    def _get_target_class():
+    def _get_partial_target_class():
```
dhermes: Please revert this name change. The test `TestPartialRowsData` is meant to test the `PartialRowsData` class, hence this method returns the "class under test".

You should add a **new** `TestYieldRowsData` and test the newly factored out class there.

sduskis:
Since `consume_next()` was a public method, does that mean that we need to maintain backwards compatibility?  If so, can we just have it read a single row via the generator?

dhermes:
> does that mean that we need to maintain backwards compatibility

Not until `1.0` (current version is `0.28.1`). We typically try to provide a "warning" period though before breaking an interface, but in a case like this it might not be possible and it'd be fine to make a breaking change in the release (we can just include that in the release notes).

zakons:
All review requests made but not finding a solution for providing unit tests that provide coverage of iterator / generator code.  Please provide any suggestions for this so this can be completed.

chemelnucfin:
This would close #1812 from discussion in #4761

zakons:
@tseaver @sduskis We are getting 100% code coverage in PyCharm, which uses coverage.py:

yield_row_coverage

However, circleci is identifying lines 338-341 as NOT having coverage.  This is the only thing holding up the merge for this PR.  Any suggestions or ideas would be very much appreciated so we can complete this PR.

tseaver:
@zakons The output on CircleCI is `338 -> 341`, which is a *branch* miss -- a normal gap would be denoted `338-341`.  Can you configure the PyCharm -> coverage bit so that it enables branch coverage?

To mitigate:  the branch is the test `if self._cell:` -- that condition is always true, and therefore the call to `self._save_current_cell()` is always executed.  We need a test which gets to that point without having a cell assigned, or else (if that is impossible) we can just take out the `if` altogether.

sduskis:
Thanks!

zunger-humu:
Footnote: You neither made this change backwards-compatible, nor even mentioned it in the release notes, which meant that I just spent quite a while bisecting the git repository to figure out why our code just broke.

Are there any notes on how this new API is even meant to be used?

PR #4686, opened 2018-01-03
Fix inaccurate storage docstring

andrewsg:


jonparrott:
Not waiting for CI, it's a trivial docs-only change.

tseaver:
@andrewsg Thanks for the patch!

PR #4688, opened 2018-01-03
DOCS: DefaultCredentialsError could be raised if credentials not supplied

tswast:
Also, use Napoleon-style docs for the changed docs.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -93,7 +93,7 @@ class Project(object):
 class Client(ClientWithProject):
     """Client to bundle configuration needed for API requests.
 
-    Keyword Args:
+    Args:
         project (str):
             Project ID for the project which the client acts on behalf of.
             Will be passed when creating a dataset / job. If not passed,
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -93,25 +93,28 @@ class Project(object):
 class Client(ClientWithProject):
     """Client to bundle configuration needed for API requests.
 
-    :type project: str
-    :param project: the project which the client acts on behalf of. Will be
-                    passed when creating a dataset / job.  If not passed,
-                    falls back to the default inferred from the environment.
+    Keyword Args:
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -93,25 +93,28 @@ def from_api_repr(cls, resource):
...
-
-    :type _http: :class:`~requests.Session`
-    :param _http: (Optional) HTTP object to make requests. Can be any object
-                  that defines ``request()`` with the same interface as
-                  :meth:`requests.Session.request`. If not passed, an
-                  ``_http`` object is created that is bound to the
-                  ``credentials`` for the current object.
-                  This parameter should be considered private, and could
-                  change in the future.
+    Keyword Args:
```
jonparrott: Just `Args:` is fine.

core/google/cloud/client.py
```diff
@@ -94,20 +94,23 @@ class Client(_ClientFactoryMixin):
...
-
-    :type _http: :class:`~requests.Session`
-    :param _http: (Optional) HTTP object to make requests. Can be any object
-                  that defines ``request()`` with the same interface as
-                  :meth:`requests.Session.request`. If not passed, an
-                  ``_http`` object is created that is bound to the
-                  ``credentials`` for the current object.
-                  This parameter should be considered private, and could
-                  change in the future.
+    Keyword Args:
```
jonparrott: likewise here.

dhermes:
@tswast We've generally followed a "don't document exceptions that are raised transitively" rule, but I'm happy to break it if you have a compelling case.

At the very least we could merge this PR with the Google-style docstring changes.

jonparrott:
> but I'm happy to break it if you have a compelling case.

I'm fine with this as it's one of the primary failure cases here.

tswast:
Understood. The reason I'm documenting this is that the colab team want to know what the exception type is raised so they can catch it and do their own auth flow on errors.

jonparrott:
Will wait on CI to ensure docs still build

jonparrott:
Build exceeded 10m, merging.

PR #4691, opened 2018-01-04
Mark bigquery as back under dev after 0.29.0 release.

tseaver:
[ci skip]


bigquery/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 [1]: https://pypi.org/project/google-cloud-bigquery/#history
 
+## 0.29.1 (unreleased)
```
dhermes: @tseaver IMO adding a section like this isn't necessary (but thanks for thinking of it).

PR #4698, opened 2018-01-04
Harden test for 'retrieve_metadata_server' against transparent DNS proxies

tseaver:
Grrr.

### DIFF
--- a/logging/tests/unit/test__helpers.py
+++ b/logging/tests/unit/test__helpers.py
@@ -110,7 +110,7 @@ class Test_retrieve_metadata_server(unittest.TestCase):
         metadata_key = 'test_url_cannot_connect'
         metadata_url = 'http://metadata.invalid/'
 
-        requests_get_mock = mock.Mock()
+        requests_get_mock = mock.Mock(spec=['__call__'])
         requests_get_mock.side_effect = requests.exceptions.RequestException
 
         requests_get_patch = mock.patch('requests.get', requests_get_mock)
--- a/logging/tests/unit/test__helpers.py
+++ b/logging/tests/unit/test__helpers.py
@@ -105,15 +105,23 @@ class Test_retrieve_metadata_server(unittest.TestCase):
         self.assertIsNone(metadata)
 
     def test_request_exception(self):
+        import requests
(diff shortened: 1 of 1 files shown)


tseaver:
The rationale:  I've seen the logging tests failing from my home machine for months now, due to that fact that my cable company intercepts DNS queries for unknown domains and redirects them to a "helpful" page.

Allowing a web request to propagate from inside a unit test is a Bad Thing(TM), in general.

logging/tests/unit/test__helpers.py
```diff
@@ -105,15 +105,23 @@ def test_metadata_does_not_exist(self):
         self.assertIsNone(metadata)
 
     def test_request_exception(self):
+        import requests
+
         metadata_key = 'test_url_cannot_connect'
         metadata_url = 'http://metadata.invalid/'
 
-        patch = mock.patch(
+        requests_get_mock = mock.Mock()
```
dhermes: Can you set the `spec`?

tseaver: It is being used to mock a single function:  do we set `spec=['__call__']` as a rule in such cases?  I'd originally written it as an inline function:

```python
        def requests_get(*args, **kw):
              raise requests.exceptions.RequestsException('test')
```

which seemed clearer, but you have an allergy for them. :)

tseaver: `git grep` showed a number of such cases, so I followed suit in 52c4b9f.

logging/tests/unit/test__helpers.py
```diff
@@ -105,15 +105,23 @@ def test_metadata_does_not_exist(self):
         self.assertIsNone(metadata)
 
     def test_request_exception(self):
+        import requests
+
         metadata_key = 'test_url_cannot_connect'
         metadata_url = 'http://metadata.invalid/'
 
-        patch = mock.patch(
+        requests_get_mock = mock.Mock(spec=['__call__'])
```
dhermes: @tseaver I think `spec=()` will produce the same results (i.e. `__call__` isn't actually necessary)

PR #4701, opened 2018-01-05
api_core: Fix handling of gapic metadata when specified as 'None'.

jonparrott:


api_core/tests/unit/gapic/test_method.py
```diff
@@ -94,6 +94,19 @@ def test_invoke_wrapped_method_with_metadata():
...
 
 
+def test_invoke_wrapped_method_with_metadata_as_none():
+    method = mock.Mock(spec=['__call__'])
+
+    wrapped_method = google.api_core.gapic_v1.method.wrap_method(method)
+
+    wrapped_method(mock.sentinel.request, metadata=None)
+
+    method.assert_called_once_with(mock.sentinel.request, metadata=mock.ANY)
```
dhermes: Why use `mock.ANY` here?

api_core/tests/unit/gapic/test_method.py
```diff
@@ -94,6 +94,19 @@ def test_invoke_wrapped_method_with_metadata():
...
+    metadata = method.call_args[1]['metadata']
```
dhermes: Why are you checking `call_args`?

api_core/tests/unit/gapic/test_method.py
jonparrott: Because I don't need the tests to re-calculate the metrics header.

dhermes: Ah

PR #4702, opened 2018-01-05
api_core: Make PageIterator.item_to_value public.

jonparrott:
This is necessary because some clients, such as Spanner, will basically hand the user the page iterator returned from gapic, but need to apply their own transformer.

### DIFF
--- a/api_core/google/api_core/page_iterator.py
+++ b/api_core/google/api_core/page_iterator.py
@@ -158,12 +158,25 @@ class Iterator(object):
                  page_token=None, max_results=None):
         self._started = False
         self.client = client
-        self._item_to_value = item_to_value
+        """Optional[Any]: The client that created this iterator."""
+        self.item_to_value = item_to_value
+        """Callable[Iterator, Any]: Callable to convert an item from the type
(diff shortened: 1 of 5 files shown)


chemelnucfin:
Hi Jon, I'm more used to the comments being right under the function instead of under the variable?  Also, it doesn't seem like client is optional or maybe I misunderstood the meaning?

jonparrott:
> Hi Jon, I'm more used to the comments being right under the function instead of under the variable? 

Sphinx uses that format to document member variables.

> Also, it doesn't seem like client is optional or maybe I misunderstood the meaning?

It is. The iterator doesn't actually use the client at all.

> LGTM but why is this good?

Basically (from spanner):

```python
    def list_instances(self, filter_='', page_size=None, page_token=None):
        """List instances for the client's project. """
        metadata = _metadata_with_prefix(self.project_name)
        path = 'projects/%s' % (self.project,)
        page_iter = self.instance_admin_api.list_instances(
            path, page_size=page_size, metadata=metadata)
        page_iter.item_to_value = self._item_to_instance
        page_iter.next_page_token = page_token
        return page_iter
```

PR #4705, opened 2018-01-05
api_core: Add grpc.Channel stub to grpc_helpers

jonparrott:
This is intended to help with testing *all* grpc-based clients. With this, we should be able to mostly avoid using mock/fakes/stubs and instead use this channel-based approach. This means that *every* layer of the client is tested and there is less concern about a mock/stub/fake not matching its upstream target. This is akin to mocking at the requests level for HTTP clients.

@dhermes: I considered putting this into a package like `api_core.testing.grpc_stub`, but I decided against it. I'm happy to reconsider.

### DIFF
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -178,8 +178,13 @@ class _CallableStub(object):
         self.requests.append(request)
 
         response = self.response
-        if response is None and self.responses is not None:
-            response = next(self.responses)
+        if self.responses is not None:
+            if response is None:
+                response = next(self.responses)
+            else:
+                raise ValueError(
+                    '{method}.response and {method}.responses are mutually '
+                    'exclusive.'.format(method=self._method))
 
         if callable(response):
             return response(request)
@@ -250,7 +255,7 @@ class ChannelStub(grpc.Channel):
         # Use a sequence of responses:
         channel_stub.GetFoo.responses = iter([
             foo_pb2.Foo(name='bar'),
-            foo_pb2.Foo(name='baz')
+            foo_pb2.Foo(name='baz'),
         ])
 
         assert client.get_foo().name == 'bar'
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -140,8 +140,11 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
         credentials, request, target, **kwargs)
 
 
-_ChannelCall = collections.namedtuple(
-    '_ChannelCall', ['request', 'timeout', 'metadata', 'credentials'])
+_MethodCall = collections.namedtuple(
+    '_MethodCall', ['request', 'timeout', 'metadata', 'credentials'])
+
+_ChannelRequest = collections.namedtuple(
+    '_ChannelRequest', ('method', 'request'))
 
 
 class _CallableStub(object):
@@ -167,10 +170,11 @@ class _CallableStub(object):
         request, timeout, metadata, and credentials."""
 
     def __call__(self, request, timeout=None, metadata=None, credentials=None):
-        self._channel._requests.append((self._method, request))
-        self.requests.append(request)
+        self._channel._requests.append(
+            _ChannelRequest(self._method, request))
         self.calls.append(
-            _ChannelCall(request, timeout, metadata, credentials))
+            _MethodCall(request, timeout, metadata, credentials))
+        self.requests.append(request)
 
         if self.responses:
             self.response = next(self.responses)
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -141,7 +141,7 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
 
 
 _MethodCall = collections.namedtuple(
-    '_MethodCall', ['request', 'timeout', 'metadata', 'credentials'])
+    '_MethodCall', ('request', 'timeout', 'metadata', 'credentials'))
 
 _ChannelRequest = collections.namedtuple(
     '_ChannelRequest', ('method', 'request'))
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -160,9 +160,10 @@ class _CallableStub(object):
         exception, the exception will be raised when this is invoked.
         """
         self.responses = None
-        """Iterator[protobuf.Message]: A iterator of responses. If specified,
-        self.response will be populated on each invocation by calling
-        ``next(self.responses)``."""
+        """Iterator[
+            Union[protobuf.Message, Callable[protobuf.Message], exception]]:
+        An iterator of responses. If specified, self.response will be populated
+        on each invocation by calling ``next(self.responses)``."""
         self.requests = []
         """List[protobuf.Message]: All requests sent to this callable."""
         self.calls = []
@@ -170,26 +171,27 @@ class _CallableStub(object):
         request, timeout, metadata, and credentials."""
 
     def __call__(self, request, timeout=None, metadata=None, credentials=None):
-        self._channel._requests.append(
+        self._channel.requests.append(
             _ChannelRequest(self._method, request))
         self.calls.append(
             _MethodCall(request, timeout, metadata, credentials))
         self.requests.append(request)
 
-        if self.responses:
-            self.response = next(self.responses)
+        response = self.response
+        if response is None and self.responses is not None:
+            response = next(self.responses)
 
-        if callable(self.response):
-            return self.response(request)
+        if callable(response):
+            return response(request)
 
-        if isinstance(self.response, Exception):
-            raise self.response
+        if isinstance(response, Exception):
+            raise response
 
-        if self.response is not None:
-            return self.response
+        if response is not None:
+            return response
 
         raise ValueError(
-            'Method stub for "{}" has no responses.'.format(self._method))
+            'Method stub for "{}" has no response.'.format(self._method))
 
 
 def _simplify_method_name(method):
@@ -220,9 +222,9 @@ class ChannelStub(grpc.Channel):
     .. code-block:: python
 
         channel_stub = grpc_helpers.ChannelStub()
-        channel_stub.GetFoo.response = foo_pb2.Foo(name='bar')
+        client = FooClient(channel=channel_stub)
 
-        client = FooClient(channel_stub)
+        channel_stub.GetFoo.response = foo_pb2.Foo(name='bar')
 
         foo = client.get_foo(labels=['baz'])
 
@@ -248,7 +250,8 @@ class ChannelStub(grpc.Channel):
         # Use a sequence of responses:
         channel_stub.GetFoo.responses = iter([
             foo_pb2.Foo(name='bar'),
-            foo_pb2.Foo(name='baz')])
+            foo_pb2.Foo(name='baz')
+        ])
 
         assert client.get_foo().name == 'bar'
         assert client.get_foo().name == 'baz'
@@ -264,21 +267,15 @@ class ChannelStub(grpc.Channel):
     """
 
     def __init__(self, responses=[]):
-        self._requests = []
+        self.requests = []
+        """Sequence[Tuple[str, protobuf.Message]]: A list of all requests made
+        on this channel in order. The tuple is of method name, request
+        message."""
         self._method_stubs = {}
-
-    @property
-    def requests(self):
-        """Sequence[Tuple[str, protobuf.Message]]: Returns a list of all
-        requests made on this channel in order. The tuple is of method name,
-        request proto."""
-        return self._requests
 
     def _stub_for_method(self, method):
         method = _simplify_method_name(method)
-        if method not in self._method_stubs:
-            self._method_stubs[method] = _CallableStub(method, self)
-
+        self._method_stubs[method] = _CallableStub(method, self)
         return self._method_stubs[method]
 
     def __getattr__(self, key):
@@ -313,8 +310,8 @@ class ChannelStub(grpc.Channel):
 
     def subscribe(self, callback, try_to_connect=False):
         """grpc.Channel.subscribe implementation."""
-        return
+        pass
 
     def unsubscribe(self, callback):
         """grpc.Channel.unsubscribe implementation."""
-        return
+        pass
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -278,9 +278,10 @@ class ChannelStub(grpc.Channel):
         return self._method_stubs[method]
 
     def __getattr__(self, key):
-        # Ideally this would only return stubs for known methods, however,
-        # grpc doesn't actually create these methods until they're needed.
-        return self._stub_for_method(key)
+        try:
+            return self._method_stubs[key]
+        except KeyError:
+            raise AttributeError
 
     def unary_unary(
             self, method,
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -145,7 +145,7 @@ _ChannelCall = collections.namedtuple(
 
 
 class _CallableStub(object):
-    """Stub for the grpc callable interface."""
+    """Stub for the grpc.*MultiCallable interfaces."""
 
     def __init__(self, method, channel):
         self._method = method
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -14,6 +14,8 @@
 
 """Helpers for :mod:`grpc`."""
 
+import collections
+
 import grpc
 import six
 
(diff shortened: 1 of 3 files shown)


api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        if callable(self.response):
+            return self.response(request)
+
+        if isinstance(self.response, Exception):
+            raise self.response
+
+        if self.response is not None:
+            return self.response
+
+        raise ValueError(
```
lukesneeringer: Is there a valid use case where someone creates a stub with no responses?

When I wrote an equivalent library for HTTP, I required pre-registration of endpoints you planned to use (so I could error if you used something else), but you already have that. That said, I am not sure there _is_ a valid situation in which GRPC succeeds (on a unary-unary request) and provides no response.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    def __init__(self, method, channel):
+        self._method = method
+        self._channel = channel
+        self.response = None
+        """Union[protobuf.Message, Callable[protobuf.Message], exception]:
+        The response to give when invoking this callable. If this is a
+        callable, it will be invoked with the request protobuf. If it's an
+        exception, the exception will be raised when this is invoked.
+        """
+        self.responses = None
```
lukesneeringer: How does this interact with a unary-stream call? Or a stream-stream, for that matter?
("Out of scope" is a valid answer.)

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    '_ChannelCall', ['request', 'timeout', 'metadata', 'credentials'])
+
+
+class _CallableStub(object):
+    """Stub for the grpc.*MultiCallable interfaces."""
+
+    def __init__(self, method, channel):
+        self._method = method
+        self._channel = channel
+        self.response = None
```
lukesneeringer: Should `responses` alone be considered sufficient?
What happens if both are set?

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        assert client.get_foo(id='123').name == 'bar123'
+    """
+
+    def __init__(self, responses=[]):
+        self._requests = []
+        self._method_stubs = {}
+
+    @property
+    def requests(self):
+        """Sequence[Tuple[str, protobuf.Message]]: Returns a list of all
```
lukesneeringer: Can this be a `namedtuple('Request', ('method_name', 'request_proto'))`?

api_core/google/api_core/grpc_helpers.py
jonparrott: No, an RPC *must* return a response (that response can totally be `empty_pb2.Empty()`, see the tests for `operations_client`.) This case is the equivalent of your client-under-test making an RPC that you do not anticipate so it should throw an error.

api_core/google/api_core/grpc_helpers.py
jonparrott: For stream responses, you're expected to set `response` to an iterable:

```python
channel.SomeStreamingThing.response = [resp1, resp2, resp3]
```

For multiple invocations of the same streaming method, you should set `responses` to an iterable of iterables:

```python
channel.SomeStreamingThing.responses = [
    [resp1, resp2],
    [resp3, resp4]
]
```

This is the reason for the `response`/`responses` split.

api_core/google/api_core/grpc_helpers.py
jonparrott: `response` will be ignored as it'll be overwritten with the first item in `responses`.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    def _stub_for_method(self, method):
+        method = _simplify_method_name(method)
+        if method not in self._method_stubs:
+            self._method_stubs[method] = _CallableStub(method, self)
+
+        return self._method_stubs[method]
+
+    def __getattr__(self, key):
+        # Ideally this would only return stubs for known methods, however,
+        # grpc doesn't actually create these methods until they're needed.
```
dhermes: Is that true?

dhermes: Seems not to be in [`proto`][1] or [`gapic`][2].

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/04308483f852d3cf551d2e33f5528bfb3e9650ac/datastore/google/cloud/datastore_v1/proto/datastore_pb2_grpc.py#L17-L57
[2]: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/04308483f852d3cf551d2e33f5528bfb3e9650ac/datastore/google/cloud/datastore_v1/gapic/datastore_client.py#L115-L149

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
     return google.auth.transport.grpc.secure_authorized_channel(
         credentials, request, target, **kwargs)
+
+
+_ChannelCall = collections.namedtuple(
+    '_ChannelCall', ['request', 'timeout', 'metadata', 'credentials'])
+
+
+class _CallableStub(object):
+    """Stub for the grpc callable interface."""
```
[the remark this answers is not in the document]
jonparrott: Fixed

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        self._method = method
+        self._channel = channel
+        self.response = None
+        """Union[protobuf.Message, Callable[protobuf.Message], exception]:
+        The response to give when invoking this callable. If this is a
+        callable, it will be invoked with the request protobuf. If it's an
+        exception, the exception will be raised when this is invoked.
+        """
+        self.responses = None
+        """Iterator[protobuf.Message]: A iterator of responses. If specified,
```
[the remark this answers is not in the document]
jonparrott: Fixed

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        self.response will be populated on each invocation by calling
+        ``next(self.responses)``."""
+        self.requests = []
+        """List[protobuf.Message]: All requests sent to this callable."""
+        self.calls = []
+        """List[Tuple]: All invocations of this callable. Each tuple is the
+        request, timeout, metadata, and credentials."""
+
+    def __call__(self, request, timeout=None, metadata=None, credentials=None):
+        self._channel._requests.append((self._method, request))
```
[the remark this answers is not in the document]
jonparrott: Done.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        self.requests.append(request)
+        self.calls.append(
+            _ChannelCall(request, timeout, metadata, credentials))
```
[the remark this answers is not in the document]
jonparrott: Yeah, considering this may be randomly accessed it's really not worth it to do that.
 
(It's also possible to change it in the future)

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+
+        if self.responses:
```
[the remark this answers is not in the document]
jonparrott: Good point, done.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    method name like "/google.pubsub.v1.Publisher/CreateTopic". This
+    returns just the name of the method, in this case "CreateTopic".
+
+    Args:
+        method (str): The name of the method.
+
+    Returns:
+        str: The simplified name of the method.
+    """
+    return method.rsplit('/', 1).pop()
```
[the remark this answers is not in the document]
jonparrott: They're more or less the same, though you're welcome to do a microbenchmark. :)

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    responses are returned and track which requests are made.
+
+    For example:
+
+    .. code-block:: python
+
+        channel_stub = grpc_helpers.ChannelStub()
+        channel_stub.GetFoo.response = foo_pb2.Foo(name='bar')
+
+        client = FooClient(channel_stub)
```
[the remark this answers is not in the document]
jonparrott: Good catch. Done.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+        # Raise an exception:
+        channel_stub.GetFoo.response = NotFound('...')
+
+        with pytest.raises(NotFound):
+            client.get_foo()
+
+        # Use a sequence of responses:
+        channel_stub.GetFoo.responses = iter([
+            foo_pb2.Foo(name='bar'),
+            foo_pb2.Foo(name='baz')])
```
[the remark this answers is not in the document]
jonparrott: Done the Luke way.

api_core/google/api_core/grpc_helpers.py
jonparrott: Done.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+    def __init__(self, responses=[]):
+        self._requests = []
+        self._method_stubs = {}
+
+    @property
+    def requests(self):
+        """Sequence[Tuple[str, protobuf.Message]]: Returns a list of all
+        requests made on this channel in order. The tuple is of method name,
+        request proto."""
+        return self._requests
```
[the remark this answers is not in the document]
jonparrott: Fixed, it's public meow.

api_core/google/api_core/grpc_helpers.py
jonparrott: Yeah I was doing things wrong before. Fixed.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -136,3 +138,178 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
...
+
+    def stream_stream(
+            self, method,
+            request_serializer=None, response_deserializer=None):
+        """grpc.Channel.stream_stream implementation."""
+        return self._stub_for_method(method)
+
+    def subscribe(self, callback, try_to_connect=False):
+        """grpc.Channel.subscribe implementation."""
+        return
```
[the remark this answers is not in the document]
jonparrott: They have to be implemented to satisfy the interface, so doing nothing is totally valid here.

api_core/tests/unit/operations_v1/test_operations_client.py
```diff
@@ -12,90 +12,60 @@
...
+    channel.GetOperation.response = operations_pb2.Operation(name='meep')
 
     response = client.get_operation('name')
 
-    request = client.operations_stub.GetOperation.call_args[0][0]
-    assert isinstance(request, operations_pb2.GetOperationRequest)
-    assert request.name == 'name'
-
-    assert response == mock.sentinel.operation
+    assert channel.GetOperation.requests[0].name == 'name'
```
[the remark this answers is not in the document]
jonparrott: > Also, I'd prefer you just check the whole protobuf rather than checking .name == (I can write that code if it's cumbersome for you)

So this runs dangerously close to duplicating the same logic in the tests that are in the method. I would prefer we just verify that the data is passed through as expected.

> I am in general uncomfortable with assertions that grab an item from a list but don't check the lists length. 

length check added.

api_core/tests/unit/operations_v1/test_operations_client.py
```diff
@@ -12,90 +12,60 @@
...
+    assert response == channel.GetOperation.response
```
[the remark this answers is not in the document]
jonparrott: equality is enough, it doesn't have to be identity. The implementation reserves the right to return a different, equivalent, object.

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -186,3 +187,131 @@ def test_create_channel_explicit_scoped(unused_secure_authorized_channel):
...
+        channel = grpc_helpers.ChannelStub()
+        stub = operations_pb2.OperationsStub(channel)
+        expected_request = operations_pb2.GetOperationRequest(name='meep')
+        expected_response = operations_pb2.Operation(name='moop')
+
+        channel.GetOperation.response = expected_response
+
+        response = stub.GetOperation(expected_request)
+
+        assert response == expected_response
```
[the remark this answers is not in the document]
jonparrott: See above.

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -186,3 +187,131 @@ def test_create_channel_explicit_scoped(unused_secure_authorized_channel):
...
+
+    def test_no_response(self):
+        channel = grpc_helpers.ChannelStub()
+        stub = operations_pb2.OperationsStub(channel)
+        expected_request = operations_pb2.GetOperationRequest(name='meep')
+
+        with pytest.raises(ValueError) as exc_info:
+            stub.GetOperation(expected_request)
+
+        assert exc_info.match('GetOperation')
```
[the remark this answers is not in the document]
jonparrott: I'm not a huge fan of doing that.

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -186,3 +187,131 @@ def test_create_channel_explicit_scoped(unused_secure_authorized_channel):
...
+            stub.GetOperation(expected_request)
+
+    def test_callable_response(self):
+        channel = grpc_helpers.ChannelStub()
+        stub = operations_pb2.OperationsStub(channel)
+        expected_request = operations_pb2.GetOperationRequest(name='meep')
+        expected_response = operations_pb2.Operation(name='moop')
+
+        def on_get_operation(request):
+            return expected_response
```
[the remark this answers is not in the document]
jonparrott: Sure why not, i can verify the args as well. Done.

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -186,3 +187,131 @@ def test_create_channel_explicit_scoped(unused_secure_authorized_channel):
...
+            operations_pb2.Operation(name='foo'),
+            operations_pb2.Operation(name='bar'),
+            operations_pb2.Operation(name='baz'),
+        ]
+
+        channel.GetOperation.responses = iter(expected_responses)
+
+        response1 = stub.GetOperation(expected_request)
+        response2 = stub.GetOperation(expected_request)
+        response3 = stub.GetOperation(expected_request)
```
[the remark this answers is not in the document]
jonparrott: See the last part of this test.

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -186,3 +187,131 @@ def test_create_channel_explicit_scoped(unused_secure_authorized_channel):
...
+    def test_stream_stream(self):
+        channel = grpc_helpers.ChannelStub()
+        method_name = 'GetOperation'
+        callable_stub = channel.stream_stream(method_name)
+        assert callable_stub._method == method_name
+        assert callable_stub._channel == channel
+
+    def test_subscribe_unsubscribe(self):
+        channel = grpc_helpers.ChannelStub()
+        channel.subscribe(None)
```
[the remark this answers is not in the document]
jonparrott: Done.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -140,8 +140,11 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
         credentials, request, target, **kwargs)
 
 
-_ChannelCall = collections.namedtuple(
-    '_ChannelCall', ['request', 'timeout', 'metadata', 'credentials'])
+_MethodCall = collections.namedtuple(
+    '_MethodCall', ['request', 'timeout', 'metadata', 'credentials'])
```
dhermes: `list` here and `tuple` below. Pick one?

jonparrott: Fixed.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -157,35 +160,38 @@ def __init__(self, method, channel):
...
+            _ChannelRequest(self._method, request))
         self.calls.append(
-            _ChannelCall(request, timeout, metadata, credentials))
+            _MethodCall(request, timeout, metadata, credentials))
+        self.requests.append(request)
 
-        if self.responses:
-            self.response = next(self.responses)
+        response = self.response
+        if response is None and self.responses is not None:
```
dhermes: If you **want** you could do validation here:

```python
if self.responses is not None:
    if response is None:
        response = next(self.responses)
    else:
        raise ValueError('Why you collision')
```

api_core/google/api_core/grpc_helpers.py
```diff
@@ -244,7 +250,8 @@ class ChannelStub(grpc.Channel):
         # Use a sequence of responses:
         channel_stub.GetFoo.responses = iter([
             foo_pb2.Foo(name='bar'),
-            foo_pb2.Foo(name='baz')])
+            foo_pb2.Foo(name='baz')
```
dhermes: Missing the trailing comma (but thanks for yielding to my lame nit)

api_core/google/api_core/grpc_helpers.py
```diff
@@ -260,27 +267,22 @@ def on_get_foo(request):
...
         return self._method_stubs[method]
 
     def __getattr__(self, key):
-        # Ideally this would only return stubs for known methods, however,
-        # grpc doesn't actually create these methods until they're needed.
-        return self._stub_for_method(key)
+        try:
+            return self._method_stubs[key]
+        except KeyError:
+            raise AttributeError
```
dhermes: :+1: to this change

api_core/google/api_core/grpc_helpers.py
```diff
@@ -308,8 +310,8 @@ def stream_stream(
 
     def subscribe(self, callback, try_to_connect=False):
         """grpc.Channel.subscribe implementation."""
-        return
+        pass
```
dhermes: With a docstring you don't need any function body (if it pleases you)

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -231,14 +237,15 @@ def test_callable_response(self):
         expected_request = operations_pb2.GetOperationRequest(name='meep')
         expected_response = operations_pb2.Operation(name='moop')
 
-        def on_get_operation(request):
-            return expected_response
+        on_get_operation = mock.Mock(
+            spec=('__call__'), return_value=expected_response)
```
dhermes: @jonparrott You aren't using a tuple there.

Also, FWIW `spec=('__call__',)` doesn't do anything (it **might** on a `MagicMock`, but not a `Mock`). All `Mock`-s are callable and you can't disable it.

api_core/google/api_core/grpc_helpers.py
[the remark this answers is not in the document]
dhermes: It's not about speed. I had to think when reading this if `.pop()` comes from the end of the beginning.

api_core/tests/unit/test_grpc_helpers.py
jonparrott: Yeah but it's more explicit, I like to say "I'm only using this mock for `__call__`"

api_core/google/api_core/grpc_helpers.py
jonparrott: Noted.

api_core/tests/unit/test_grpc_helpers.py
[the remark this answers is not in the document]
dhermes: DERP

api_core/google/api_core/grpc_helpers.py
jonparrott: Fixed.

api_core/google/api_core/grpc_helpers.py
jonparrott: Donion rings.

jonparrott:
Thanks, @dhermes and @lukesneeringer!

@dhermes, @tseaver, @chemelnucfin, @lukesneeringer: Please keep an eye out in code reviews for excessive mocking of gapic/grpc clients - where possible, this should be used instead. I'm happy to help write tests until we develop a good reference corpus.

PR #4706, opened 2018-01-05
Move 'unique_writer_identity' from LogSink ctor to 'create'

tseaver:
Preps support for passing it to 'update' as well.

See:
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4703#issuecomment-355622741


tseaver:
> Was that change ever released? (I'm fairly certain the answer is no)

Nope, I checked.

chemelnucfin:
I was going to ask the same question...

PR #4707, opened 2018-01-05
Capture server-generated 'writerIdentity' during 'LogSink.create'

tseaver:
~~Uses #4706 as a base.~~

PR #4708, opened 2018-01-05
Support passing 'uniqueWriterIdentity' to 'Sink.update'.

tseaver:
~~Uses #4707 as a base.~~

Also, capture server-set read-only properties returned from `sinks.update`, such as `writerIdentity`.

### DIFF
--- a/logging/google/cloud/logging/_http.py
+++ b/logging/google/cloud/logging/_http.py
@@ -290,7 +290,8 @@ class _SinksAPI(object):
         target = '/projects/%s/sinks/%s' % (project, sink_name)
         return self.api_request(method='GET', path=target)
 
-    def sink_update(self, project, sink_name, filter_, destination):
+    def sink_update(self, project, sink_name, filter_, destination,
+                    unique_writer_identity=False):
         """API call:  update a sink resource.
 
         See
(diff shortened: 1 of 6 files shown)


dhermes:
@tseaver Ping once #4707 is merged?

tseaver:
@dhermes Done.

logging/google/cloud/logging/_http.py
```diff
@@ -319,7 +325,9 @@ def sink_update(self, project, sink_name, filter_, destination):
             'filter': filter_,
             'destination': destination,
         }
-        return self.api_request(method='PUT', path=target, data=data)
+        query_params = {'uniqueWriterIdentity': unique_writer_identity}
```
dhermes: @tseaver Is passing an explicit false somehow different than the "implicit false" of sending nothing at all?

tseaver: For [update](https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.sinks/update#body.QUERY_PARAMETERS.unique_writer_identity), the docs say:

> When updating a sink, the effect of this field on the value of writerIdentity in the updated sink depends on both the old and new values of this field:
>
> - If the old and new values of this field are both false or both true, then there is no change to the sink's writerIdentity.
 > - If the old value is false and the new value is true, then writerIdentity is changed to a unique service account.
 > - It is an error if the old value is true and the new value is set to false or defaulted to false.

tseaver: Which is a bogus API choice, but the answer is "no, no difference."

PR #4716, opened 2018-01-08
Do not quote embedded slashes for public / signed URLs

### LINKED ISSUE #3809: public_url returning quoted_name with '/'
i'm adding my files inside folders, so i give the blob name as `"UUID/file.ext"`, but when i request the [public_url()](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/d0cfeb772fd612179984a191be1d8189b369a2c5/storage/google/cloud/storage/blob.py#L224) i get the fullname as [quoted](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/d0cfeb772fd612179984a191be1d8189b369a2c5/storage/google/cloud/storage/blob.py#L233) example: `".../MYBUCKET/UUID%2Ffile.ext"` 

both `"%2F"` and `"/"` in the link downloads the same file, but with different names

tseaver:
Closes #3809.

As noted on that issue, the `public_url` and `generate_signed_url` methods use the XML endpoints, which allow for / expect the slash not to be quoted.  All other methods use the JSON endpoints, which *require* quoting the embedded slashes.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -244,7 +244,7 @@ class Blob(_PropertyMixin):
         return '{storage_base_url}/{bucket_name}/{quoted_name}'.format(
             storage_base_url=_API_ACCESS_ENDPOINT,
             bucket_name=self.bucket.name,
-            quoted_name=quote(self.name))
+            quoted_name=quote(self.name.encode('utf-8')))
 
     def generate_signed_url(self, expiration, method='GET',
                             content_type=None,
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -244,7 +244,7 @@ class Blob(_PropertyMixin):
         return '{storage_base_url}/{bucket_name}/{quoted_name}'.format(
             storage_base_url=_API_ACCESS_ENDPOINT,
             bucket_name=self.bucket.name,
-            quoted_name=_quote(self.name))
+            quoted_name=quote(self.name))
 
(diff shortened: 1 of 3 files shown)


frankyn:
@tseaver, it looks like it's still failing. Did you update the PR given the last commit was back in Jan 8?

frankyn:
```
Required test coverage of 97% reached. Total coverage: 99.97%
================================== FAILURES ===================================
__________________ Test_Blob.test_public_url_with_non_ascii ___________________
  File "C:\projects\google-cloud-python\storage\tests\unit\test_blob.py", line 182, in test_public_url_with_non_ascii
    self.assertEqual(blob.public_url, expected_url)
  File "C:\projects\google-cloud-python\storage\google\cloud\storage\blob.py", line 247, in public_url
    quoted_name=quote(self.name))
  File "c:\python27\Lib\urllib.py", line 1298, in quote
    return ''.join(map(quoter, s))
KeyError: u'\u2603'
1 failed, 422 passed in 3.61 seconds
```
It's an issue with 2.7.

frankyn:
@tseaver, I think it's a unicode issue given the following [SO question](https://stackoverflow.com/questions/15115588/urllib-quote-throws-keyerror). 

Given that, the line could be updated with:
```
quoted_name=quote(self.name.encode('utf-8')))
```

tseaver:
@frankyn I opened issue #4935  for the unrelated Bigtable system test flakiness, and restarted the build after cleaning out the orphaned instances.  I plan to merge here later today unless you see a problem.

pfk-beta:
When it will be in official release?:>

PR #4718, opened 2018-01-09
fix docs

ofek:


tseaver:
Thank you for the patch, @ofek!

PR #4719, opened 2018-01-09
slo compliance lister

chemelnucfin:
I don't know where best to put this if this should even go here...  Any suggestions?
I also need to double check the exact response days.

Other languages may find it useful also?

### DIFF
--- a/slo_compliance.py
+++ b/slo_compliance.py
@@ -1,0 +1,112 @@
+from datetime import date
+from datetime import timedelta
+from collections import namedtuple
+import requests
+from bs4 import BeautifulSoup
+
+today = date.today()
+language = 'python'
+google_cloud = 'https://www.github.com/GoogleCloudPlatform/google-cloud-{}/'
+github_url = google_cloud.format(language)
+
+no = '-'
+Types = namedtuple('Label',
+                   ['bug', 'feature', 'question', 'cleanup', 'process'])
+types = Types('label:"type: bug" ',
+              'label:"type: feature request" ',
+              'label:"type: question" ',
+              'label:"type: cleanup" ',
+              'label:"type: process" ')
+
+Priority = namedtuple('Priority', ['p0', 'p1', 'p2'])
+priority = Priority('label:"priority: p0" ',
+                    'label:"priority: p1" ',
+                    'label:"priority: p2" ')
+
+
+def updated(days):
+    return 'updated:<={}'.format(today-timedelta(days))
+
+
+def created(days):
+    return 'created:<={}'.format(today-timedelta(days))
+
+
+def query(url, params, name):
+    r = requests.get(url, params=params)
+    soup = BeautifulSoup(r.text, 'html.parser')
+    number = int(soup
+                 .find('a', class_='btn-link selected')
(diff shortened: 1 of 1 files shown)


slo_compliance.py
```diff
@@ -0,0 +1,111 @@
...
+                                          + priority.p2
+                                          + created(12)),
+          'feature_requests_outside_response_slo': (types.feature
+                                                    + updated(120)),
+          'feature_requests_outside_closure_slo': (types.feature
+                                                   + created(120)),
+          'cleanup_outside_response_slo': (types.cleanup
+                                           + updated(120)),
+          'cleanup_outside_closure_slo': (types.cleanup
+                                          + created(120))}
```
tseaver: As far as I know, there is no SLO for closure / response to a `type: cleanup` issue.

slo_compliance.py
```diff
@@ -0,0 +1,111 @@
...
+                                           + updated(5)),
+          'p1_bugs_outside_closure_slo': (types.bug
+                                          + priority.p1
+                                          + created(42)),
+          'p2_bugs_outside_response_slo': (types.bug
+                                           + priority.p2
+                                           + updated(120)),
+          'p2_bugs_outside_closure_slo': (types.bug
+                                          + priority.p2
+                                          + created(12)),
```
tseaver: I don't think this makes sense (12 days closure vs. 120 days response).  Use constants rather than "magic numbers", and the problem would be more obvious.

slo_compliance.py
```diff
@@ -0,0 +1,111 @@
...
+          'cleanup_outside_closure_slo': (types.cleanup
+                                          + created(120))}
+
+pull_requests = {'pull_requests_with_no_type': (no+types.bug
+                                                + no+types.feature
+                                                + no+types.question
+                                                + no+types.cleanup
+                                                + no+types.process),
+                 'pull_requests_outside_response_slo': updated(5),
+                 'pull_requests_outside_closure_slo': created(5)}
```
tseaver: I don't think we have SLO on pull requests, either:  certainly not unless somebody has assigned a priority to them.

slo_compliance.py
chemelnucfin: I think there isn't an official one, but if it's been there for 4 months, maybe we should close it out or do something with it?
ditto for pull requests.

slo_compliance.py
chemelnucfin: I changed the numbers to 42 and 120.

Also, this is just a tool to list issues.  If it isn't official, we can change it to whatever to suit our needs.

slo_compliance.py
chemelnucfin: Changed 12 to 120.

I *think* I will leave the numbers here rather than putting it in constants higher up unless you feel strongly about it?  They are fairly separate issues and it is easier to see that way.  Unless we use these numbers elsewhere, then I will consider refactoring them out.  But otherwise you will just see a variable that you have to scroll above for.

chemelnucfin:
I'm going to close this one.  I wrote a note linking it here in the Issue and Bug Process document.

Feel free to use it.

PR #4724, opened 2018-01-09
BigQuery: expose delete contents

### LINKED ISSUE #4721: BigQuery: add argument to delete_dataset for deleteContents parameter
The [delete verb for datasets](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/delete) has a `deleteContents` parameter.

> If True, delete all the tables in the dataset. If False and the dataset contains tables, the request will fail. Default is False

We do not currently expose this parameter in [delete_dataset()](https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/reference.html#google.cloud.bigquery.client.Client.delete_dataset).

I could see this parameter being very handy in integration and snippets tests, where we want to delete a dataset and all the tables we have created.

chemelnucfin:
Closes #4721

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -446,7 +446,8 @@ class Client(ClientWithProject):
             DeprecationWarning)
         return self.list_tables(*args, **kwargs)
 
-    def delete_dataset(self, dataset, retry=DEFAULT_RETRY):
+    def delete_dataset(self, dataset, delete_contents=False,
+                       retry=DEFAULT_RETRY):
         """Delete a dataset.
 
         See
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -446,7 +446,8 @@ def list_dataset_tables(self, *args, **kwargs):
             DeprecationWarning)
         return self.list_tables(*args, **kwargs)
 
-    def delete_dataset(self, dataset, retry=DEFAULT_RETRY):
+    def delete_dataset(self, dataset, retry=DEFAULT_RETRY,
```
jonparrott: I'd prefer if `delete_contents` appeared before `retry`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -455,15 +456,27 @@ def delete_dataset(self, dataset, retry=DEFAULT_RETRY):
...
+        :type delete_contents: boolean
+        :param delete_contents: (Optional) If True, delete all the tables
+             in the dataset. If False and the dataset contains tables, the
+             request will fail. Default is False
         """
         if not isinstance(dataset, (Dataset, DatasetReference)):
             raise TypeError('dataset must be a Dataset or a DatasetReference')
-        self._call_api(retry, method='DELETE', path=dataset.path)
+
+        params = {'deleteContents': 'false'}
```
jonparrott: I would not specify this param at all in the false case.

PR #4727, opened 2018-01-10
BigQuery: missing return value in LoadJobConfig.from_api_repr and LoadJobConfig Tests

### LINKED ISSUE #4726: BigQuery: LoadJobConfig.from_api_repr does not return config
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/63b7245699510259f584bc3ca328a72af8c2120f/bigquery/google/cloud/bigquery/job.py#L677
misses `return config`. Hence cannot build based on a JSON schema file.

chemelnucfin:
Closes #4726

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -674,6 +674,7 @@ class LoadJobConfig(object):
         config._properties = copy.deepcopy(resource)
         config.schema = _parse_schema_resource(schema)
         config.skip_leading_rows = _int_or_none(slr)
+        return config
 
 
 class LoadJob(_AsyncJob):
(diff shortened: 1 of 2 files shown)


chemelnucfin:
cover said 100%  I'll take a look.

jonparrott:
Sure, but, clearly there was a bug here that no test caught before. Some test, some where, should be updated to make sure that this returns the right object.

dhermes:
Also @chemelnucfin please make your PR descriptions actually describe what they do (this is **much** more important than the `BigQuery: ` prefix, which is largely irrelevant since we have labels).

chemelnucfin:
@jonparrott @dhermes I'm more curious why cover is not catching it.  Seems like the whole class is missing.
Ok to merge it and then open another issue?

jonparrott:
> Ok to merge it and then open another issue?

No, I'd prefer we take care of this here, especially considering how small this patch is.

chemelnucfin:
ok.

chemelnucfin:
@jonparrott PTAL

@dhermes changed title

Any idea why cover wasn't working?

jonparrott:
> Any idea why cover wasn't working?

Because coverage doesn't have a way of knowing whether you checked the return value.

PR #4734, opened 2018-01-11
Revert "Revert "api_core: Make PageIterator.item_to_value public. (#4702)""

jonparrott:
Reverts GoogleCloudPlatform/google-cloud-python#4731

inception-meme

Do not merge until api_core 0.2.0 is being prepped for release.

Context: 

> This is done to prevent a release-the-world scenario in order to release bigquery, spanner, and datastore. It will be re-reverted before the next minor release of api_core.

### DIFF
--- a/api_core/google/api_core/page_iterator.py
+++ b/api_core/google/api_core/page_iterator.py
@@ -179,15 +179,6 @@ class Iterator(object):
         """int: The total number of results fetched so far."""
 
     @property
-    def _item_to_value(self):
-        """DEPRECATED: Backwards-compatible alias to ``item_to_value``."""
-        return self.item_to_value
-
-    @_item_to_value.setter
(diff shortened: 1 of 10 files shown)


dhermes:
@jonparrott After #4750 goes in (and this gets rebased on top of it) I'd like to add another commit to make sure that datastore and spanner depend on `google-api-core >= 0.1.5.dev1` (I'm on the fence on whether BigQuery should depend on this version, since it's a test-only change).

jonparrott:
I have added an alias property for the private method. This allows us to maintain compatibility even for some older clients. We can remove it at a later date.

I will merge this once CI is green.

jonparrott:
nvm, I'm going to release the world anyway, I'm gonna remove the alias.

Issue #4710, opened 2018-01-05
Spanner v1.0.0 release


jonparrott:
This bug tracks the criteria for releasing v1.0.0 of Spanner.

The following items need to be completed:

- [x] Update gapic code. #4695
- [x] Remove dependency on deprecated gax library. #4695
- [x] Release new `api_core`, since it contains changes that Spanner needs. #4730
- [x] Release a release candidate. #4732
- [x] Have a Storage DPE review the API surface & documentation. (@jsimonweb)
- [x] Have a Python DPE review the API surface for overall API consistency. (@jonparrott)
- [x] Have Spanner POC review the API surface & documentation. (@vkedia)
- [x] Wait one week and resolve any user issues.
- [x] Release v1.0.0. ✨

jonparrott:
@jsimonweb @vkedia can you please review v0.30.0 of the client library and it's [documentation](https://google-cloud-python.readthedocs.io/en/latest/spanner/usage.html)?

lukesneeringer:
@jonparrott We are requiring `api_core` to go to 1.0 in order for Spanner to do so, yes? (I assume it is in a similar "wait a week" state at the moment.)

vkedia:
Some corrections need to be done on the documentation:

Code snippet on [this](https://google-cloud-python.readthedocs.io/en/latest/spanner/batch-usage.html) page talks about `session.batch()` instead of `client.batch()`.

On [snapshots](https://google-cloud-python.readthedocs.io/en/latest/spanner/snapshot-usage.html) page we talk about user having to resume the stream but now we do it automatically so it should be removed.

On [snapshots](https://google-cloud-python.readthedocs.io/en/latest/spanner/snapshot-usage.html) page (and probably elsewhere) we show iterator over rows as `for row in result.rows`. But this is incorrect since we now removed the `rows` property.

Read write transactions [https://google-cloud-python.readthedocs.io/en/latest/spanner/transaction-usage.html] page talks about manually managing the transaction. Instead it should talk about `run_in_transaction` method.

vkedia:
Why are the `to_pb` methods part of public api in [KeySet](https://google-cloud-python.readthedocs.io/en/latest/spanner/keyset-api.html#google.cloud.spanner_v1.keyset.KeyRange.to_pb) ?

jonparrott:
> Why are the to_pb methods part of public api in KeySet ?

[...]

PR #4740, opened 2018-01-11
Spanner: make to_pb private in keyset

chemelnucfin:
This might be considered a breaking change?
See #4710


jonparrott:
Let's wait on this until we get more details.

chemelnucfin:
@jonparrott merging because cover is missing on */gapic/* and */proto/* files.  Let me know otherwise.

PR #4741, opened 2018-01-11
Adding a unit test with the empty list case in `KeyRange.to_pb()`.

chemelnucfin:
Also reworked the `to_pb()` tests to just create a protobuf and
just use one assertion.

--------------------------------------------------------------------------------------------------
This test got reverted and I think it should still be there.


chemelnucfin:
@dhermes permission?

chemelnucfin:
Got permission on chat from @dhermes

PR #4744, opened 2018-01-12
Spanner: system tests use database instead of session

chemelnucfin:
There are still a couple of them that use session, I'll look later on them.

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -457,17 +457,12 @@ class TestSessionAPI(unittest.TestCase, _TestData):
         retry = RetryInstanceState(_has_all_ddl)
         retry(self._db.reload)()
 
-        session = self._db.session()
-        session.create()
-        self.to_delete.append(session)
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -728,153 +698,150 @@ def _row_data(max_index):
...
         keyset = [[expected[row][0], expected[row][1]]]
-        results_iter = session.read(self.TABLE,
-                                    columns,
-                                    KeySet(keys=keyset),
-                                    index='name'
-        )
-        rows = list(results_iter)
-        self.assertEqual(rows, [expected[row]])
+        with self._db.snapshot() as snapshot:            
+            results_iter = snapshot.read(self.TABLE,
```
jonparrott: since you're here, fix this hanging indent (line break at the first `(` and just indent once).

chemelnucfin: changed this and other ones like it to 
```
function(
  arg1,
  arg2,
)
```

PR #4754, opened 2018-01-15
Add handling of missing properties in SchemaField.from_api_repr()

### LINKED ISSUE #4456: BigQuery: SchemaField cannot load resource with missing `mode` property
The [mode property](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#schema.fields.mode) is optional in the REST API, but this code fails:

```
SchemaField.from_api_repr({'name': 'colname', 'type': 'INT64'})
```

I believe this is due to the line

```
mode=api_repr['mode'].upper(),
```

which does not properly handle when the `mode` property is missing.

Related to https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3723 as it is often convenient to omit the mode parameter when defining a schema file by hand.

alanbato:
This should fix #4456 

I added safe attribute access to the optional properties `mode` and `fields` in the param dictionary, using the default parameteres from the constructor as fallback values.

I couldn't find the appropiate place to update the docs if it's necessary, so any pointer is greatly appreciated. :)

🐍

### DIFF
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -56,10 +56,13 @@ class SchemaField(object):
             google.cloud.biquery.schema.SchemaField:
                 The ``SchemaField`` object.
         """
+        # Handle optional properties with default values
+        mode = api_repr.get('mode', 'NULLABLE')
+        fields = api_repr.get('fields', ())
         return cls(
             field_type=api_repr['type'].upper(),
-            fields=[cls.from_api_repr(f) for f in api_repr.get('fields', ())],
-            mode=api_repr['mode'].upper(),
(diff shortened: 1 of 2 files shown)


alanbato:
@tseaver What do you mean they should break in the same way? I thought we wanted to avoid #4456, meaning it doesn't break when you don't send those parameters.

Can you explain that part a little bit more, please?

Meanwhile I'll be adding the tests that validate the defaults work

tseaver:
Sorry for the confusion (posted early):  I meant that those tests should fail with the same traceback as #4456 *when your fix is not applied*.

alanbato:
Hi @tseaver, I added a unit test to verify the defaults are working, as you advised. If you could take a look it'd be awesome :)

Issue #4736, opened 2018-01-11
Cloud Spanner with Python multiprocessing hangs


snthibaud:
Inserting rows turned out to be very slow (80 KB/s), so I am trying multiprocessing to parallelize the process.
I am currently using only one node in Cloud Spanner, the rows are sorted by primary key and there are ~30 batches that have a few hundred rows each.
For this I am using the following code:
```
pool = Pool(5)
pool.starmap(_process_batch, batches)


def _process_batch(instance_name, database_name, table_name, column_names, b):
    spanner_client = spanner.Client(credentials=get_credentials())
    instance = spanner_client.instance(instance_name)
    database = instance.database(database_name)
    with database.batch() as batch:
        batch.insert(table=table_name, columns=tuple(column_names), values=b)
```
The code hangs when it tries to enter the batch scope (it seems to never succeed in creating a session).
Any ideas?

vkedia:
cc @jonparrott

jonparrott:
@snthibaud can you let us know what versions you're using? `pip freeze` is helpful.

snthibaud:
These are the version in my virtual environment:
```
Flask | 0.12.2 | 0.12.2
-- | -- | --
Jinja2 | 2.9.6 | 2.10
MarkupSafe | 1.0 | 1.0
PyYAML | 3.12 | 3.12
TA-Lib | 0.4.10 | 0.4.14
Werkzeug | 0.12.2 | 0.14.1
amqp | 2.2.2 | 2.2.2
argparse | 1.4.0 | 1.4.0
billiard | 3.5.0.3 | 3.5.0.3
cachetools | 2.0.1 | 2.0.1
celery | 4.1.0 | 4.1.0
certifi | 2017.11.5 | 2017.11.5
certitude | 1.0.1 | 1.0.1
cffi | 1.11.2 | 1.11.2
chardet | 3.0.4 | 3.0.4
click | 6.7 | 6.7
cycler | 0.10.0 |  
dateutils | 0.6.6 | 0.6.6
dill | 0.2.7.1 | 0.2.7.1
facebookads | 2.11.1 | 2.11.1
future | 0.16.0 | 0.16.0
gapic-google-cloud-datastore-v1 | 0.15.3 | 0.90.4
google-api-core | 0.1.1 | 0.1.3
google-api-python-client | 1.6.4 | 1.6.4
google-auth | 1.2.0 | 1.2.1
[...]

PR #4756, opened 2018-01-16
Test for multiprocessing usage

chemelnucfin:
@kpayson64 @jonparrott For #4736.  This test passes in python 3.6 (starmap doesn't exist in python2.7).  Did I write something wrong?

I'm also wondering if something else changed since I last looked, as `batch.delete` and `batch.insert` also seemed to be working now.

(I also know there is a possible race condition, but i just wanted to see if the process would hang or not)

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -59,6 +59,43 @@ COUNTERS_TABLE = 'counters'
 COUNTERS_COLUMNS = ('name', 'value')
 
 
+def _process(databases):
+    table = 'contacts'
+    columns = ('contact_id', 'first_name', 'last_name', 'email')
+    
+    client = Client()
+    pool = BurstyPool()
+    retry = RetryErrors(exceptions.ServiceUnavailable)    
+    configs = list(retry(client.list_instance_configs)())
+
+    # Defend against back-end returning configs for regions we aren't
+    # actually allowed to use.
+    configs = [config for config in configs if '-us-' in config.name]
+
+    if len(configs) < 1:
+        raise ValueError('List instance configs failed in module set up.')
+
+    config_name = configs[0].name
+    instance = client.instance(INSTANCE_ID, config_name)
+    database = instance.database(
+        databases, ddl_statements=DDL_STATEMENTS, pool=pool)
+
+    #removed a row here
+    row_data = (
+        (1, u'Phred', u'Phlyntstone', u'phred@example.com'),
+        (2, u'Bharney', u'Rhubble', u'bharney@example.com'),
+    )
+
+    with database.batch() as batch:
+        batch.delete(table, KeySet(all_=True))
+        batch.insert(table, columns, row_data)
+        
+    with database.snapshot() as snapshot:
+        rows = list(snapshot.read(table, columns, KeySet(all_=True)))
+
+    return rows
+
+
 class Config(object):
     """Run-time configuration to be modified at set-up.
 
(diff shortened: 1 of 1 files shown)


spanner/tests/system/test_system.py
```diff
@@ -395,6 +432,53 @@ def _unit_of_work(transaction, name):
         self.assertEqual(len(rows), 2)
 
 
+    def test_multi_processing_transaction(self):
+        retry = RetryErrors(exceptions.ServiceUnavailable)
+
+        from multiprocessing import Pool
+        processes = 3
+        pool = Pool(processes)
```
jonparrott: I suspect if you drop this below `client = Client()` in 442 this will fail.

chemelnucfin: still passing after moving the pool below client.

jonparrott: If you make a call before constructing the pool does it still work?

chemelnucfin: I moved the pool construction to after line 463 and it still worked.

jonparrott: hrm

chemelnucfin: Would I need to update some pip files or would nox take care of that?

jonparrott: nox should handle that, but it might be useful to try reproducing this outside of our system testing harness.

vkedia: `_process` creates its own client at 66. So the client being created here should not be relevant.

jonparrott: @vkedia that's what one would think, but according to the gRPC team interacting with gRPC at all in the process starts persistent state (threads) that does not survive `fork`.

vkedia: I see. So even if they create a new Client object in the subprocess (which is what is happening here), it is not supposed to work?
What I am trying to say is, even though a Client object is being created here, it is not being used in the sub processes since they are all creating there own client objects. IIUC you are saying even that should not work?

jonparrott: From my understanding that is correct.

chemelnucfin:
@jonparrott so I'm having problems importing both protobuf or spanner.  The former I can't import google.api, the latter I can't import google.protobuf.  I think somehow the namespaces are clashing.

I have tried to look through python -v, but I haven't been able to find any obvious fixes at the moment.

chemelnucfin:
@jonparrott 
So after feeling like an idiot trying to import google.cloud.spanner from spanner directory,

I used the same code I had in the test in an interpreter setting with no problems.

chemelnucfin:
@jonparrott Actually, is the batch commit atomic, so there won't be a race condition?

chemelnucfin:
@jonparrott So this is now python 2 and python 3 compatible and can be merged.  Is there something I missed regarding the multiprocessing that both of these tests pass locally?

jonparrott:
I don't know why these would work without hanging. Perhaps the gRPC team can shed some light?

kpayson64:
@chemelnucfin 
Are these tests attempting to use the same channel in both the parent and child process?  I would be really surprised if that works.

Also, the guidance we currently provide (don't create any channels before forking) isn't a hard guideline;  some applications that violate this might still work anyways, but some might not.  The guideline is provided because we know that applications that follow it will work.

FWIW, we are taking a stab at a more comprehensive fork support given the pain that users reported in 
https://github.com/grpc/grpc/issues/13873

chemelnucfin:
https://github.com/grpc/grpc/issues/12455#issuecomment-348578950 mentions that this is only reproduced on mac, not linux.  I'm on ubuntu 16.04 right now.

Anyone have a mac to test this on?

chemelnucfin:
@jonparrott If you have some time next week, could you take a look and see if the code would hang for you?  (I believe you have a Mac?)
Thanks.

jonparrott:
@tseaver I think it's probably a good idea to have this in our test suite. Can you review and merge?

spanner/tests/system/test_system.py
```diff
@@ -394,6 +431,94 @@ def _unit_of_work(transaction, name):
...
+
+        config_name = configs[0].name
+        instance = client.instance(INSTANCE_ID, config_name)
+        database_name = 'test_database' + unique_resource_id('_')
+        bursty_pool = BurstyPool()
+        database = instance.database(
+            database_name, ddl_statements=DDL_STATEMENTS, pool=bursty_pool)
+        try:
+            operation = database.create()
+            operation.result(30)  # raises on failure / timeout.
```
tseaver: Why not just use the shared database?  Setting one up is slow, and doesn't exercise anything specific to testing multiprocessing (we just need to pass its name through, same as the one generated here).

spanner/tests/system/test_system.py
```diff
@@ -394,6 +431,94 @@ def _unit_of_work(transaction, name):
...
+            processes = 3
+            pool = Pool(processes)
+            batches = [database_name] * processes
+            rows_list = pool.map(_process, batches)
+            pool.close()
+            pool.join()
+            self.assertEqual(len(rows_list), processes)
+            for rows in rows_list:
+                # removed a row in _process
+                self._check_rows_data(rows, self.ROW_DATA[:-1])
```
tseaver: The comment / slice doesn't make sense -- `_process` deletes all rows *before* the insert, but doesn't delete any after.

spanner/tests/system/test_system.py
```diff
@@ -394,6 +431,94 @@ def _unit_of_work(transaction, name):
...
+
+        config_name = configs[0].name
+        instance = client.instance(INSTANCE_ID, config_name)
+        database_name = 'test_database' + unique_resource_id('_')
+        bursty_pool = BurstyPool()
+        database = instance.database(
+            database_name, ddl_statements=DDL_STATEMENTS, pool=bursty_pool)
+        try:
+            operation = database.create()
+            operation.result(30)  # raises on failure / timeout.
```
tseaver: Again, scrap the database creation -- just use `self._db.name`.

spanner/tests/system/test_system.py
chemelnucfin: @tseaver I removed a row from the `row_data` which populates the table in `_process` to show that I'm getting something different.  I'll address your other comments.

chemelnucfin:
This test is actually hanging for me now with no changes.

chemelnucfin:
It looks like something has changed in the backend and getting segmentation fault, getting a -11 exit code from pytest.  (SIGSEGV)

@jonparrott Maybe need to take this to the gRPC team?

jonparrott:
@mehrdada @nathanielmanistaatgoogle

spanner/tests/system/test_system.py
```diff
@@ -394,6 +431,94 @@ def _unit_of_work(transaction, name):
                 COUNTERS_TABLE, COUNTERS_COLUMNS, self.ALL))
         self.assertEqual(len(rows), 2)
 
+    def test_multi_processing_map_transaction(self):
+        from multiprocessing import Pool
+        client = Client()
```
kpayson64: Why are you creating a client before the fork()?

This is known to cause issues with gRPC, and isn't something we support yet.  If it worked at some point it was purely coincidental.

chemelnucfin: That is what is interesting.  It was working fine before.  But I guess if it is meant to not be supported then I guess this isn't a valid behavior test.

chemelnucfin:
Since this is not actually going to be supported behavior I'm just going to close this.

Maybe at some point it will be good as a behavior test.

PR #4757, opened 2018-01-17
Implement predefined acl

### LINKED ISSUE #1660: support for predefined ACL to be set directly on the upload request

chemelnucfin:
Closes #1660

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -55,6 +55,7 @@ from google.cloud.iam import Policy
 from google.cloud.storage._helpers import _PropertyMixin
 from google.cloud.storage._helpers import _scalar_property
 from google.cloud.storage._signing import generate_signed_url
+from google.cloud.storage.acl import ACL
 from google.cloud.storage.acl import ObjectACL
 
 
@@ -633,7 +634,7 @@ class Blob(_PropertyMixin):
         return headers, object_metadata, content_type
 
     def _do_multipart_upload(self, client, stream, content_type,
-                             size, num_retries):
+                             size, num_retries, predefined_acl):
         """Perform a multipart upload.
 
         Assumes ``chunk_size`` is :data:`None` on the current blob.
@@ -664,6 +665,9 @@ class Blob(_PropertyMixin):
         :param num_retries: Number of upload retries. (Deprecated: This
                             argument will be removed in a future release.)
 
+        :type predefined_acl: str
+        :param predefined_acl: (Optional) predefined access control list
+
         :rtype: :class:`~requests.Response`
         :returns: The "200 OK" response object returned after the multipart
                   upload request.
@@ -688,6 +692,8 @@ class Blob(_PropertyMixin):
 
         if self.user_project is not None:
             name_value_pairs.append(('userProject', self.user_project))
+        if predefined_acl is not None:
+            name_value_pairs.append(('predefinedAcl', predefined_acl))
 
         upload_url = _add_query_parameters(base_url, name_value_pairs)
         upload = MultipartUpload(upload_url, headers=headers)
@@ -702,8 +708,8 @@ class Blob(_PropertyMixin):
         return response
 
     def _initiate_resumable_upload(self, client, stream, content_type,
-                                   size, num_retries, extra_headers=None,
-                                   chunk_size=None):
+                                   size, num_retries, predefined_acl,
+                                   extra_headers=None, chunk_size=None):
         """Initiate a resumable upload.
 
         The content type of the upload will be determined in order
@@ -727,6 +733,9 @@ class Blob(_PropertyMixin):
         :param size: The number of bytes to be uploaded (which will be read
                      from ``stream``). If not provided, the upload will be
                      concluded once ``stream`` is exhausted (or :data:`None`).
+
+        :type predefined_acl: str
+        :param predefined_acl: (Optional) predefined access control list
 
         :type num_retries: int
         :param num_retries: Number of upload retries. (Deprecated: This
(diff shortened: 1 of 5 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -688,6 +691,9 @@ def _do_multipart_upload(self, client, stream, content_type,
 
         if self.user_project is not None:
             name_value_pairs.append(('userProject', self.user_project))
+        if predefined_acl is not None:
+            self._acl.save_predefined(predefined_acl)
+            name_value_pairs.append(('predefinedAcl', predefined_acl))
```
tseaver: I don't think we can use `save_predefined` here -- it forces an API request, which will fail for not-yet-existing blobs.  We likely need to factor out the "name conversion" logic from that method for reuse here.

A system test which exercises the feature would have made this clear.

tseaver: Also, this argument shouldn't need to be passed down from `upload_*` to the helper methods:  instead, the top-level methods should do the work of setting the ACL:  the lower-level methods will generate the right payload from that.

chemelnucfin:
@tseaver PTAL, thanks.  I set the acl after the blob was created as you had suggested.

tseaver:
@chemelnucfin Moving the `save_predefined` call after the actual upload still makes two API calls.  The change I asked for was to avoid using `save_predefined` at all, and instead to get the appropriate query parameter added to the actual upload API request.  This change will require factoring out the XML->JSON parameter name code (currently inside `save_predefined`), and tweaking `_initiate_resumable_upload` and `_do_multipart_upload` to inject the query parameter appropriately (like for `userProject`).

chemelnucfin:
Sorry, I must have misunderstood you.  I will take a look in a bit.  Thanks.

chemelnucfin:
@tseaver PTAL at your convenience.  Thanks.

storage/tests/unit/test_blob.py
```diff
@@ -1009,9 +1013,11 @@ def _initiate_resumable_helper(
         data = b'hello hallo halo hi-low'
         stream = io.BytesIO(data)
         content_type = u'text/plain'
+        predefined_acl = 'private'
```
tseaver: Shouldn't `_initiate_resumable_helper` take `predefined_acl` as an argument, like `_do_multipart_helper`.  We need to continue to exercise the codepath *without* the value passed in, too.

storage/tests/unit/test_blob.py
```diff
@@ -1180,8 +1187,9 @@ def _do_resumable_helper(self, use_size=False, num_retries=None):
         client = mock.Mock(_http=transport, spec=['_http'])
         stream = io.BytesIO(data)
         content_type = u'text/html'
+        predefined_acl = 'private'
```
tseaver: Likewise `_do_resumable_helper`.

storage/tests/unit/test_blob.py
```diff
@@ -1225,20 +1233,21 @@ def _do_upload_helper(self, chunk_size=None, num_retries=None):
         stream = mock.sentinel.stream
         content_type = u'video/mp4'
         size = 12345654321
-
+        predefined_acl = 'private'
```
tseaver: Likewise `_do_upload_helper`

storage/tests/unit/test_blob.py
```diff
@@ -1252,9 +1261,10 @@ def test__do_upload_with_retry(self):
 
     def _upload_from_file_helper(self, side_effect=None, **kwargs):
         from google.cloud._helpers import UTC
+        from google.cloud.storage.acl import ACL
 
         blob = self._make_one('blob-name', bucket=None)
-
+        blob._acl = mock.create_autospec(ACL)
```
tseaver: Mocking the ACL without making assertions about how it was used seems wrong.

storage/tests/unit/test_blob.py
```diff
@@ -1187,7 +1190,6 @@ def _do_resumable_helper(self, use_size=False, num_retries=None):
         client = mock.Mock(_http=transport, spec=['_http'])
         stream = io.BytesIO(data)
         content_type = u'text/html'
-        predefined_acl = 'private'
         response = blob._do_resumable_upload(
             client, stream, content_type, size, num_retries, None)
```
tseaver: This should be passing through the `predefined_acl` argument, and checking that it is passed through in the assertions.

chemelnucfin:
@tseaver Hopefully this time is good.  PTAL at your convenience.

storage/google/cloud/storage/blob.py
```diff
@@ -941,14 +965,18 @@ def upload_from_file(self, file_obj, rewind=False, size=None,
             warnings.warn(_NUM_RETRIES_MESSAGE, DeprecationWarning)
 
         _maybe_rewind(file_obj, rewind=rewind)
+        predefined_acl = ACL.validate_predefined(predefined_acl)
```
tseaver: Hmm, I guess that does the Right Thing(TM) when the value passed in is `None`?

storage/google/cloud/storage/blob.py
```diff
@@ -1109,7 +1145,7 @@ def create_resumable_upload_session(
             # to the `ResumableUpload` constructor. The chunk size only
             # matters when **sending** bytes to an upload.
             upload, _ = self._initiate_resumable_upload(
-                client, dummy_stream, content_type, size, None,
+                client, dummy_stream, content_type, size, None, None,
```
tseaver: For clarity, can we use keyword arguments for the two `None` values here?

chemelnucfin: They are actually positional args.  Let's just fix this in another PR for clarity purposes.

storage/google/cloud/storage/blob.py
chemelnucfin: Seemed like it and shown with the test.  it was code nicely written which I moved.

chemelnucfin:
@tseaver added the unit test for validation.

PR #4760, opened 2018-01-18
Make row assertion in sample tests account for non-deterministic order

jonparrott:
This was causing some fun failures: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/5213?utm_campaign=build-failed&utm_medium=email&utm_source=notification

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -485,15 +485,13 @@ Wylma Phlyntstone,29
     token = iterator.next_page_token
     # [END table_list_rows_iterator_properties]
 
-    expected_rows = [
-        bigquery.Row(('Wylma Phlyntstone', 29), {'full_name': 0, 'age': 1}),
-        bigquery.Row(('Phred Phlyntstone', 32), {'full_name': 0, 'age': 1}),
-    ]
     assert len(rows) == total == 2
     assert token is None
     # Order is not preserved, so compare individually
-    for row in expected_rows:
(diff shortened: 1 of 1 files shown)


docs/bigquery/snippets.py
```diff
@@ -491,7 +491,9 @@ def do_something(row):
     ]
     assert len(rows) == total == 2
     assert token is None
-    assert rows == expected_rows
+    # Order is not preserved, so compare individually
+    for row in expected_rows:
+        assert row in rows
```
dhermes: You also need a length check?

jonparrott: The length check is above.

dhermes: Ah MERP! Since there is only two, it seems a for loop is "overkill"

jonparrott: Eh, yeah, still a useful assertion I think.

dhermes: The `for` loop isn't an assertion? I'm saying, just do:

```python
row1 = bigquery.Row(('Wylma Phlyntstone', 29), {'full_name': 0, 'age': 1})
assert row1 in rows
row2 = bigquery.Row(('Phred Phlyntstone', 32), {'full_name': 0, 'age': 1})
assert row2 in rows
```

jonparrott: Oh I see, easy enough. Done in e07a8620b.

PR #4761, opened 2018-01-19
BigTable: PartialRowsData iterator

chemelnucfin:
#1812

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -221,6 +221,69 @@ class PartialRowsData(object):
     def __ne__(self, other):
         return not self == other
 
+    def __iter__(self):
+        return self._consume_next(True)
+
+    def _consume_next(self, yield_=False):
+        """ Helper for consume_next.
+
+        :type yield_: bool
+        :param yield_: if True, yields rows as they complete,
+                       else finish iteration of the response_iterator
+        """
+        while True:
+            response = six.next(self._response_iterator)
+            self._counter += 1
+
+            if self._last_scanned_row_key is None:  # first response
+                if response.last_scanned_row_key:
+                    raise InvalidReadRowsResponse()
+
+            self._last_scanned_row_key = response.last_scanned_row_key
+
+            row = self._row
+            cell = self._cell
+
+            for chunk in response.chunks:
+
+                self._validate_chunk(chunk)
+
+                if chunk.reset_row:
+                    row = self._row = None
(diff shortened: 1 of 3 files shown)


bigtable/tests/unit/test_row_data.py
```diff
@@ -641,12 +659,28 @@ def _flatten_cells(prd):
...
+        # Flatten PartialRowData
+        for family_name, family in prd.cells.items():
             for qualifier, column in family.items():
                 for cell in column:
                     yield {
-                        u'rk': _bytes_to_unicode(row_key),
+                        u'rk': _bytes_to_unicode(prd.row_key),
                         u'fm': family_name,
                         u'qual': _bytes_to_unicode(qualifier),
                         u'ts': _microseconds_from_datetime(cell.timestamp),
```
chemelnucfin: Haven't thought of a good way to refactor this yet due to the yield statements...  Suggestions would be welcome.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -221,6 +221,69 @@ def __eq__(self, other):
...
+                        chunk.value)
+                    self._copy_from_previous(cell)
+                else:
+                    cell.append_value(chunk.value)
+
+                if chunk.commit_row:
+                    self._save_current_row()
+                    if yield_:
+                        yield self._previous_row
+                    row = cell = None
```
chemelnucfin: The yield_ statements are the only added code from the block below

zakons:
This PR is redundant to PR #4679 where PartialRowsData uses YieldRowsData, which in turn uses a generator, so the client can iterate over the rows on a large scan and not have to store them in a dictionary (taking lots of memory for large scans of 1M rows). If PartialRowsData is used, the generator stores the result in the **rows** dictionary, which already provides a built-in iterator, hence no need to add an iterator:
```
for row_key, row_data in partial_rows_data.rows.iteritems():
    ...
```
@sduskis has reviewed PR #4679 and participated in the design.  There is further refactoring of this  code in progress to improve performance of large scans and, since PartialRowData uses YieldRowsData in this PR, it will also benefit from this ongoing work.

All review comments have been incorporated for this PR and only the coverage for the yield is still in progress but should be done early this week, then ready to merge.

chemelnucfin:
@zakons sounds good.

chemelnucfin:
@zakons would removing consume_next be a breaking change on #4679?

chemelnucfin:
@zakons never mind, Danny addressed that.

PR #4762, opened 2018-01-19
Add max_latency to BackgroundThreadTransport

tcwalther:
This PR adds a new parameter `max_latency` to `BackgroundThreadTransport` in logging. It can be used to enforce batching by asking the transport to wait for new log messages for a specified amount of time. This is useful to avoid hitting the Stackdriver Logging rate limit.

This PR is written on behalf of Spotify, who, as a company, has signed a corporate CLA. The idea for this PR is based on a support case with Google.

### DIFF
--- a/logging/google/cloud/logging/handlers/transports/background_thread.py
+++ b/logging/google/cloud/logging/handlers/transports/background_thread.py
@@ -31,13 +31,13 @@ from google.cloud.logging.handlers.transports.base import Transport
 
 _DEFAULT_GRACE_PERIOD = 5.0  # Seconds
 _DEFAULT_MAX_BATCH_SIZE = 10
-_DEFAULT_MIN_WAIT_TIME = 0  # Seconds
+_DEFAULT_MAX_LATENCY = 0  # Seconds
 _WORKER_THREAD_NAME = 'google.cloud.logging.Worker'
 _WORKER_TERMINATOR = object()
 _LOGGER = logging.getLogger(__name__)
 
 
-def _get_many(queue_, max_items=None, min_wait_time=0):
+def _get_many(queue_, max_items=None, max_latency=0):
     """Get multiple items from a Queue.
 
     Gets at least one (blocking) and at most ``max_items`` items
@@ -50,9 +50,10 @@ def _get_many(queue_, max_items=None, min_wait_time=0):
     :param max_items: The maximum number of items to get. If ``None``, then all
         available items in the queue are returned.
 
-    :type min_wait_time: float
-    :param min_wait_time: The minimum number of seconds to wait for items from
-        a queue.
+    :type max_latency: float
+    :param max_latency: The maximum number of seconds to wait for more than one
+        item from a queue. This number includes the time required to retrieve
+        the first item.
 
     :rtype: Sequence
     :returns: A sequence of items retrieved from the queue.
@@ -63,7 +64,7 @@ def _get_many(queue_, max_items=None, min_wait_time=0):
     while max_items is None or len(items) < max_items:
         try:
             elapsed = time.time() - start
-            timeout = max(0, min_wait_time - elapsed)
+            timeout = max(0, max_latency - elapsed)
             items.append(queue_.get(timeout=timeout))
         except queue.Empty:
             break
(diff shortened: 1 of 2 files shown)


tcwalther:
I signed it!

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -48,14 +50,21 @@ def _get_many(queue_, max_items=None):
...
     """
+    start = time.time()
     # Always return at least one item.
     items = [queue_.get()]
     while max_items is None or len(items) < max_items:
         try:
-            items.append(queue_.get_nowait())
+            elapsed = time.time() - start
+            timeout = max(0, min_wait_time - elapsed)
+            items.append(queue_.get(timeout=timeout))
```
jonparrott: This actually sets the "maximum wait time", if there are items in the queue this will still return immediately until there are none but will also wait an additional amount of time specified by `timeout` to see if any more items come in. If that's the intended behavior, cool. I would name this `max_latency` to match PubSub's [batch semantics](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py#L86).

tcwalther: If `max_items is None`, this function will always take at least `min_wait_time` seconds to run. Example:

* `min_wait_time = 2`
* queue item 1 is there at the start
* queue item 2 comes after 1s
* queue item 3 comes after 3s

The function starts, and immediately `items == [item1]`. It then enters the loop and calls `queue_.get(timeout=2)`. After 1s, `items == [item1, item2]`. It does another iteration and calls `queue_.get(timeout=1)`. This will time out, and the function returns `[item1, item2]`. Only one second after the function returns, queue item 3 will be enqueued.

tcwalther: @jonparrott You are right, however, in that if `max_items` is set, this function will return earlier if `min_wait_time` is set and `max_items` has been reached. I am happy to rename this to `max_latency` if you feel that this is more intuitive.

jonparrott: Yeah, I think `max_latency` is better and fits with the Pub/Sub strategy. Basically pubsub does the "same" thing as this - collects some messages in the background and sends them to the server after a certain amount of time has passed.

tcwalther: I've renamed the variable to `max_latency`, as suggested.

tcwalther:
When reviewing this PR, I would be particularly interested in your opinion on how safe it is when a Python processes terminates due to an exception - is there a possibility for log messages getting lost?

logging/google/cloud/logging/handlers/transports/background_thread.py
```diff
@@ -48,14 +50,22 @@ def _get_many(queue_, max_items=None):
     :param max_items: The maximum number of items to get. If ``None``, then all
         available items in the queue are returned.
 
+    :type max_latency: float
+    :param max_latency: The maximum number of seconds to wait for more than one
+        item from a queue. This number includes the time required to retrieve
+        the first item.
```
jonparrott: Suggestion, add this to the end here:

```This means this is effectively the longest amount of time the background thread will hold onto log entries before sending them to the server.```

File: logging/tests/unit/handlers/transports/test_background_thread.py [hunk 3]
```diff
@@ -264,6 +272,34 @@ def test__thread_main_batches(self):
         self.assertFalse(worker._cloud_logger._batch.commit_called)
         self.assertEqual(worker._queue.qsize(), 0)
 
+    def test__thread_main_max_latency(self):
```
jonparrott: It might be useful to mock `time` here so that these tests are more deterministic.

jonparrott:
> There is a method called _Worker._main_thread_terminated, which claims to send unsent logs. However, it does not seem to do anything other than calling print() a couple of times. It seems to me that this is a bug in the current implementation already.


The call to [`self.stop`](https://github.com/tcwalther/google-cloud-python/blob/2ed09e5aaf2bbf74d5d5ca544d2ad8f7a1de98c7/logging/google/cloud/logging/handlers/transports/background_thread.py#L219) attempts to [block](https://github.com/tcwalther/google-cloud-python/blob/2ed09e5aaf2bbf74d5d5ca544d2ad8f7a1de98c7/logging/google/cloud/logging/handlers/transports/background_thread.py#L198) until the background thread has sent pending logs.

> When reviewing this PR, I would be particularly interested in your opinion on how safe it is when a Python processes terminates due to an exception - is there a possibility for log messages getting lost?

This is always possible, yes, however, an exception on the *main* thread (or any thread other than *this* background thread) will not cause this to lose logs. Lost logs can happen due to one of three reasons:

1. The request to send the logs fails (due to timeout, etc.)
2. The main thread dies for any reason (graceful exit, exception, etc.) and sending the pending logs takes longer than `grace_period` seconds.
3. There is a bug in the background thread that causes it to terminate without sending logs.

logging/google/cloud/logging/handlers/transports/background_thread.py
tcwalther: I've added this line, but to the `_Worker` and `BackgroundThreadTransport` docstrings, since the `_get_many` function's current doc string is not tied to logging but kept quite generic.

logging/tests/unit/handlers/transports/test_background_thread.py
tcwalther: I don't know of an easy way to do this. If I mocked the `time` module, I would also have to mock `worker._queue.get`, which doesn't seem very clean. I'd personally be more comfortable with the test as-is.

jonparrott:
@tcwalther I re-wrote the test to use monotonic time to keep our unit tests deterministic and fast, totally understand not wanting to do it - it's not an obvious way to write the test.

This looks good, I'm merging. Thank you for doing all of this (and enduring my review).

PR #4765, opened 2018-01-19
BigTable: Timestamp system test fix

chemelnucfin:


bigtable/tests/system.py
```diff
@@ -311,8 +311,15 @@ def _write_to_row(self, row1=None, row2=None, row3=None, row4=None):
         timestamp1 = _datetime_from_microseconds(timestamp1_micros)
         # 1000 microseconds is a millisecond
         timestamp2 = timestamp1 + datetime.timedelta(microseconds=1000)
+        timestamp2_micros = _microseconds_from_datetime(timestamp2)
+        timestamp2_micros -= (timestamp1_micros % 1000)
```
jonparrott: Weird, are you truncating the nanosecond part?

chemelnucfin: I was just following what was done before.  IIRC, the tests work truncated or not.

jonparrott: so why is this needed?

chemelnucfin: I see your point.  Removed the truncations since they don't do anything anyway.

PR #4768, opened 2018-01-21
Logging: Fix logging system tests

chemelnucfin:


logging/tests/system/test_system.py
```diff
@@ -103,7 +103,11 @@ def setUp(self):
     def tearDown(self):
         retry = RetryErrors(NotFound, max_tries=9)
         for doomed in self.to_delete:
-            retry(doomed.delete)()
+            try:
```
jonparrott: one of the things you can do is make `to_delete` a list of functions to call, I think @dhermes has done this in other places by calling it `cleanup`, e.g.:

```
self.cleanup.append(thing.delete)
```

chemelnucfin: if that's the case I'll clean it up in another PR since I have to touch a bit more other code.  I'll open up an issue.

PR #4778, opened 2018-01-23
Datastore: Allow empty lists in protobuf array value

### LINKED ISSUE #3152: ValueError: For an array_value, subvalues must either all be indexed or all excluded from indexes.

chemelnucfin:
Fixes #3152.

@lukesneeringer This was your fix, so I believe you should be able to get me an LGTM soon right?  :)

@dhermes I would like to ask that you are ok with it since I mostly referred to your test_system code.

@tseaver Do you mind looking it over just so I didn't miss anything?

I consolidated all of our responses and comments and made this pull request.

I backed out the code change and the errors happen. 

Separately, I have also investigated that
``` 
array_values{
  values{
  }
}
```
would `raise ValueError` when put into an entity.  It might be desired behavior, but we might need to deal with it at some point.  The issue is that `len(array_value.values)` is 1 and the result for this was not `[]` as I had expected in the following code:
```
        result = [_get_value_from_value_pb(value)
                  for value in value_pb.array_value.values]
```

Anyway, shall we get this merged soon please?

### DIFF
--- a/datastore/tests/unit/test_helpers.py
+++ b/datastore/tests/unit/test_helpers.py
@@ -147,10 +147,8 @@ class Test_entity_from_protobuf(unittest.TestCase):
         entity_pb.key.path.add(kind=_KIND, id=_ID)
 
         array_val_pb = _new_value_pb(entity_pb, 'baz')
-        array_pb = array_val_pb.array_value.values
         array_val_pb.array_value.CopyFrom(entity_pb2.ArrayValue(values=[]))
-        
-        unindexed_value_pb1 = array_pb.add()
+
         entity = self._call_fut(entity_pb)
         entity_dict = dict(entity)
         self.assertEqual(entity_dict['baz'], [])
--- a/datastore/tests/unit/test_helpers.py
+++ b/datastore/tests/unit/test_helpers.py
@@ -135,6 +135,26 @@ class Test_entity_from_protobuf(unittest.TestCase):
         with self.assertRaises(ValueError):
             self._call_fut(entity_pb)
 
+    def test_index_mismatch_ignores_empty_list(self):
+        from google.cloud.datastore_v1.proto import entity_pb2
+        from google.cloud.datastore.helpers import _new_value_pb
+
+        _PROJECT = 'PROJECT'
+        _KIND = 'KIND'
+        _ID = 1234
+        entity_pb = entity_pb2.Entity()
+        entity_pb.key.partition_id.project_id = _PROJECT
+        entity_pb.key.path.add(kind=_KIND, id=_ID)
+
+        array_val_pb = _new_value_pb(entity_pb, 'baz')
+        array_pb = array_val_pb.array_value.values
+        array_val_pb.array_value.CopyFrom(entity_pb2.ArrayValue(values=[]))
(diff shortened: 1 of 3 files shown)


datastore/tests/unit/test_helpers.py
```diff
@@ -135,6 +135,26 @@ def test_mismatched_value_indexed(self):
...
+        _ID = 1234
+        entity_pb = entity_pb2.Entity()
+        entity_pb.key.partition_id.project_id = _PROJECT
+        entity_pb.key.path.add(kind=_KIND, id=_ID)
+
+        array_val_pb = _new_value_pb(entity_pb, 'baz')
+        array_pb = array_val_pb.array_value.values
+        array_val_pb.array_value.CopyFrom(entity_pb2.ArrayValue(values=[]))
+        
+        unindexed_value_pb1 = array_pb.add()
```
tseaver: We don't want to assign any value into the `array_pb`:  it won't be empty if we do.

I'm not sure whether this line is doing nothing (because 151 might replace the `array_val_pb.array_value.values`) or whether it is actually making a non-empty array.

chemelnucfin: I think you are right.  I had a previous test that I was using, and I think L150 and L153 are redundant and I have removed them.

chemelnucfin:
This is what `entity_pb` looks like after line 150
```
key {
  partition_id {
    project_id: "PROJECT"
  }
  path {
    kind: "KIND"
    id: 1234
  }
}
properties {
  key: "baz"
  value {
    array_value {
    }
  }
}
```

datastore/tests/unit/test_helpers.py
chemelnucfin: Actually, might as well make a note here, the array_pb.add() is how I got this I mentioned above, which cannot be used to create an entity at this time.  I changed the code in the original post to fix it, but I figured if it isn't broken, don't fix it.  But this can be a note in case we need to in the future.
```
array_values{
  values{
  }
}
```

chemelnucfin:
@lukesneeringer @tseaver Are we concerned about something else breaking if we merge this in?

From what I understand of the code, the bug was only a problem if value has `[]` in it, which is fine.
But the `set` was checking if whether all of the values had the same exclusion, it was comparing it to length 1, which missed the empty array which is always `True`.  

The other solution would be to check to see if `len(set) < 1` instead of `!= 1` and then make sure we catch the pop `KeyError`, but that's similar to just checking to see if the array has something in it.

It seems that the only case that the merged code would be different is that value has an `[]` in it, in which something else would occur, and since the code only appends to `exclude_from_indexes`.  Every other case of `array_value` would remain the same, and an empty array would have no excluded indexes to worry about.

Should we do a 
```
try:
  merge
except:
  fix
``` 
here or is there still too much risk concern?

chemelnucfin:
Or maybe Luke's version was better, since maybe the empty list gets excluded also?

chemelnucfin:
Changed to Luke's version

chemelnucfin:
I'll try again later.

tseaver:
Superseded by #4915.

PR #4241, opened 2017-10-23
Updated CONTRIBUTING.rst to describe setting up dev environment using…

mfschwartz:
… nox instead of virtualenv

PR #4779, opened 2018-01-24
Fix contrib note (based from #4241)

chemelnucfin:
I addressed the comments in #4241, so if I may pull the trigger on that, someone can look at the diff here.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -62,55 +62,6 @@ but you can also use a ``virtualenv`` of your own creation.
 
 .. _repo: https://github.com/GoogleCloudPlatform/google-cloud-python
 
-Use ``nox``, not ``virtualenv``
-===========================
-
-- To install dependencies into a virtual environment, run the unit test via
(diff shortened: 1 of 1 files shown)


mfschwartz:
@chemelnucfin - your updates LGTM, though it looks like @lukesneeringer is the requested reviewer.

CONTRIBUTING.rst
```diff
@@ -62,8 +62,26 @@ but you can also use a ``virtualenv`` of your own creation.
 
 .. _repo: https://github.com/GoogleCloudPlatform/google-cloud-python
 
-Using a custom ``virtualenv``
-=============================
+Use ``nox``, not ``virtualenv``
```
jonparrott: Can someone explain why we need this section at all? There's instructions below on how to use Nox.

mfschwartz: +1
Using virtualenv directly creates problems, so I think we should just say to use nox, and drop this section.

chemelnucfin:
@jonparrott Removed virtualenv section

PR #4785, opened 2018-01-24
Update monitoring gapic client, re-integrate helpers

supriyagarg:
Added back support for the Query class, and for converting timeseries data into a pandas DataFrame object.


jonparrott:
@supriyagarg I don't see the actual addition of the monitoring gapic client here?

monitoring/google/cloud/monitoring/_dataframe.py
```diff
@@ -0,0 +1,143 @@
...
+        Specifying neither ``label`` or ``labels`` results in a dataframe
+        with a multi-level column header including the resource type and
+        all available resource and metric labels.
+
+        Specifying both ``label`` and ``labels`` is an error.
+
+    :rtype: :class:`pandas.DataFrame`
+    :returns: A dataframe where each column represents one time series.
+    """
+    import pandas   # pylint: disable=import-error
```
lukesneeringer: I would prefer have the import at the top of the file:

```python
try:
    import pandas
except ImportError:
    pandas = None
```

And then do tests for pandas in relevant methods:

```python
if pandas is None:
    raise RuntimeError('This method requires `pandas` to be installed.')
```

This is not a mission-critical change, so I am only asking for you to alter it if there are _other_ changes that also need to be made.

monitoring/google/cloud/monitoring/_dataframe.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if labels is not None:
```
lukesneeringer: A better way to go about this might be to coerce this into a sequence, always:

```python
if label is not None:
    if labels:
        raise ValueError('Cannot specify both `label` and `labels`.')
    labels = (label,)
```

monitoring/google/cloud/monitoring/_dataframe.py
```diff
@@ -0,0 +1,143 @@
...
+        pandas_series = pandas.Series(
+            data=[_extract_value(point.value) for point in time_series.points],
+            index=[point.interval.end_time.ToNanoseconds()
+                   for point in time_series.points],
+        )
+        columns.append(pandas_series)
+        headers.append(_extract_header(time_series))
+
+    # Implement a smart default of using all available labels.
+    if label is None and labels is None:
```
lukesneeringer: If you made the change above, this test would change to `if labels is not None:`.

Also, do you want `is None` or just falsity? Do you want an empty sequence to behave differently from `None`? (An alternate way to ask that: is having zero loop iterations in 122-124 valid?)

monitoring/google/cloud/monitoring/_dataframe.py
```diff
@@ -0,0 +1,143 @@
...
+    # Assemble the columns into a DataFrame.
+    dataframe = pandas.DataFrame.from_records(columns).T
+
+    # Convert the timestamp strings into a DatetimeIndex.
+    dataframe.index = pandas.to_datetime(dataframe.index)
+
+    # Build a multi-level stack of column headers. Some labels may
+    # be undefined for some time series.
+    levels = []
+    for key in labels or [label]:
```
lukesneeringer: This just becomes `for key in labels:` once you make the change I suggested above.

monitoring/google/cloud/monitoring/_dataframe.py
```diff
@@ -0,0 +1,143 @@
...
+        level = [_extract_labels(header).get(key, '') for header in headers]
+        levels.append(level)
+
+    # Build a column Index or MultiIndex. Do not include level names
+    # in the column header if the user requested a single-level header
+    # by specifying "label".
+    dataframe.columns = pandas.MultiIndex.from_arrays(
+        levels,
+        names=labels or None)
```
lukesneeringer: Ahhhh, _this_ is why you maintain both `label` and `labels` throughout the function. And you meant to do it (I would have guessed bug, but your comment describes this behavior).

I would still recommend making the change as I suggested, just because it will simplify most issues. Since `label` is still set (it never gets cleared) you can do `labels if not label else None` here, and get the same effect.

monitoring/google/cloud/monitoring/query.py
```diff
@@ -0,0 +1,604 @@
...
+
+    def copy(self):
+        """Copy the query object.
+
+        :rtype: :class:`Query`
+        :returns: The new query object.
+        """
+        # Using copy.deepcopy() would be appropriate, except that we want
+        # to copy self._client only as a reference.
+        new_query = copy.copy(self)
```
lukesneeringer: Why not implement [`__deepcopy__`](https://docs.python.org/3/library/copy.html) and copy everything in the memo dictionary except `_client`? That is what your comment suggests you want.

lukesneeringer:
> @supriyagarg I don't see the actual addition of the monitoring gapic client here?

I added it in the base branch.

monitoring/google/cloud/monitoring/_dataframe.py
supriyagarg: Done

monitoring/google/cloud/monitoring/_dataframe.py
supriyagarg: Done

monitoring/google/cloud/monitoring/_dataframe.py
supriyagarg: Done

monitoring/google/cloud/monitoring/_dataframe.py
supriyagarg: Check for falsity is fine. None or [] would behave the same now. Earlier an error was being thrown when labels == [].

In 122-124, we will always have 1 or more iterations. When both label and labels are not passed in, we return all the available "labels" or headers in the data.

monitoring/google/cloud/monitoring/_dataframe.py
supriyagarg: Done

monitoring/google/cloud/monitoring/query.py
supriyagarg: Thanks for the suggestion. Done.
PTAL, and let me know if it makes sense.

supriyagarg:
@lukesneeringer:  Thanks very much for your comments. I have addressed all of them.
The code is largely copied over from the master branch, including the comments - this was originally written in collaboration with @rimey

chemelnucfin:
@lukesneeringer How does this tie in with autogen?  Is just specifying automodule will autogenerate that class?  And then query and _dataframe are just manual classes?

PR #4786, opened 2018-01-25
Allow listing/getting jobs even when there is an "invalid" job.

alixhami:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4246 by bypassing property validation in `QueryJobConfig.from_api_repr()`. A `QueryJobConfig` object can be created for an invalid job retrieved from the API, but validation will still occur when assigning invalid property values through the setters.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -1468,7 +1468,7 @@ class QueryJobConfig(object):
     @property
     def default_dataset(self):
         """google.cloud.bigquery.dataset.DatasetReference: the default dataset
-        to use for unqualified table names in the query.
+        to use for unqualified table names in the query or ``None`` if not set.
 
         See
         https://g.co/cloud/bigquery/docs/reference/v2/jobs#configuration.query.defaultDataset
@@ -1485,7 +1485,7 @@ class QueryJobConfig(object):
     @property
     def destination(self):
         """google.cloud.bigquery.table.TableReference: table where results are
-        written
+        written or ``None`` if not set.
 
         See
         https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.destinationTable
@@ -1521,7 +1521,7 @@ class QueryJobConfig(object):
 
     @property
     def maximum_bytes_billed(self):
-        """int: Maximum bytes to be billed for this job.
+        """int: Maximum bytes to be billed for this job or ``None`` if not set.
 
         See
         https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.maximumBytesBilled
@@ -1595,7 +1595,7 @@ class QueryJobConfig(object):
     @property
     def table_definitions(self):
         """Dict[str, google.cloud.bigquery.external_config.ExternalConfig]:
-        Definitions for external tables.
+        Definitions for external tables or ``None`` if not set.
 
         See
         https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -26,7 +26,6 @@ from google.cloud.exceptions import NotFound
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud.bigquery.dataset import DatasetReference
 from google.cloud.bigquery.external_config import ExternalConfig
-from google.cloud.bigquery.query import _AbstractQueryParameter
 from google.cloud.bigquery.query import _query_param_from_api_repr
 from google.cloud.bigquery.query import ArrayQueryParameter
 from google.cloud.bigquery.query import ScalarQueryParameter
@@ -1481,8 +1480,6 @@ class QueryJobConfig(object):
 
     @default_dataset.setter
     def default_dataset(self, value):
-        if value is not None and not isinstance(value, DatasetReference):
-            raise ValueError('Required type: DatasetReference')
         self._properties['defaultDataset'] = DatasetReference.to_api_repr(
             value)
 
@@ -1501,8 +1498,6 @@ class QueryJobConfig(object):
 
     @destination.setter
     def destination(self, value):
-        if value is not None and not isinstance(value, TableReference):
-            raise ValueError('Required type: TableReference')
         self._properties['destinationTable'] = TableReference.to_api_repr(
             value)
 
@@ -1540,8 +1535,6 @@ class QueryJobConfig(object):
 
     @maximum_bytes_billed.setter
     def maximum_bytes_billed(self, value):
-        if value is not None and not isinstance(value, int):
-            raise ValueError('Required type: int')
         self._properties['maximumBytesBilled'] = str(value)
 
     priority = QueryPriority('priority', 'priority')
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1419,39 +1428,78 @@ def from_api_repr(cls, resource):
...
-    A list of :class:`google.cloud.bigquery.UDFResource` (empty
-    by default)
+        See:
+        https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.queryParameters
+        """
+        prop = self._properties.get('queryParameters', [])
+        return _from_api_repr_query_parameters(prop)
+
+    @query_parameters.setter
+    def query_parameters(self, value):
```
chemelnucfin: If value is an iterable, would it be better to name it "values"?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1316,58 +1315,43 @@ class QueryJobConfig(object):
...
     def from_api_repr(cls, resource):
         """Factory: construct a job configuration given its API representation
 
-        :type resource: dict
-        :param resource:
-            An extract job configuration in the same representation as is
-            returned from the API.
+        Args:
+            resource (dict): A query job configuration in the same
+                representation as is returned from the API.
```
tswast: @jonparrott Should this be aligned with "A" on the above line, or is indenting only 4 chars okay for Napoleon/Google-style doc-strings?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1316,58 +1315,43 @@ class QueryJobConfig(object):
...
 
-        :rtype: :class:`google.cloud.bigquery.job.QueryJobConfig`
-        :returns: Configuration parsed from ``resource``.
+        Returns:
+            :class:`~google.cloud.bigquery.job.QueryJobConfig`: Configuration
```
tswast: I think the `:class:` piece isn't necessary with Google-style. Just `google.cloud.bigquery.job.QueryJobConfig: Config...`

bigquery/google/cloud/bigquery/job.py
jonparrott: If the question is ever "Should I do an hanging indent"? The answer is always no.

alixhami: I was using the [Sphinx documentation example](http://www.sphinx-doc.org/en/stable/ext/example_google.html) for Google Style Python docstrings, which uses a 4 space indent when continuing to the next line. Is there a better source for docstring style?

tswast: I'd recommend breaking after the `:` and using 4-space indent for all lines so they line up.

tswast:
Yes, we could extend `_TypedApiResourceProperty`, but that class is hard to follow and would get harder if we extended it to do the resource property conversions.

You're right that we probably don't need to do this much validation. What if we skip the type checks but leave the type conversions?

jonparrott:
Why would you still need the type conversions?

If a class is hard to follow by the maintainers, especially a private class, feel free to replace it.

tswast:
Type conversions need to happen at some point, because we want to expose things as nice Python types, whereas the JSON resource type might be a dictionary or a string.

For example, `maximum_bytes_billed` would be nice as a `int` in Python but in the JSON resource it's a string because it could be up to 64-bits and would lose precision if stored as a JSON number.

The problem this PR is solving is that the type conversions were happening when we construct the class, but now they are happening as late as possible (in the property getter/setter). It's possible to send a resource with bad types, such as a timestamp in the wrong format (as is done in Alix's system test for this) and then it becomes impossible to get that resource or even list them. Instead, it would make more sense to fail when trying to fetch the problem property, but leave the others alone.

jonparrott:
Gotcha

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1384,20 +1369,45 @@ def from_api_repr(cls, resource):
...
+        prop = self._properties.get('destinationTable')
+        if prop is not None:
+            prop = TableReference.from_api_repr(prop)
+        return prop
+
+    @destination.setter
+    def destination(self, value):
+        if value is not None and not isinstance(value, TableReference):
+            raise ValueError('Required type: TableReference')
+        self._properties['destinationTable'] = TableReference.to_api_repr(
```
tswast: The `to_api_repr()` function is a regular method, so this could be `value.to_api_repr()`

Ditto for other uses of `to_api_repr()`.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1384,20 +1369,45 @@ def from_api_repr(cls, resource):
...
+        https://g.co/cloud/bigquery/docs/reference/rest/v2/jobs#configuration.query.destinationTable
+        """
+        prop = self._properties.get('destinationTable')
+        if prop is not None:
+            prop = TableReference.from_api_repr(prop)
+        return prop
+
+    @destination.setter
+    def destination(self, value):
+        if value is not None and not isinstance(value, TableReference):
```
tswast: Per @jonparrott's comment, we don't need the `isinstance(...)` check in our setters.

chemelnucfin:
@alixhami @tswast Please feel free to remove me from the assignee list.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1384,20 +1368,39 @@ def from_api_repr(cls, resource):
...
     """
 
-    default_dataset = _TypedApiResourceProperty(
-        'default_dataset', 'defaultDataset', DatasetReference)
-    """See
-    https://g.co/cloud/bigquery/docs/reference/v2/jobs#configuration.query.defaultDataset
-    """
+    @property
+    def default_dataset(self):
+        """google.cloud.bigquery.dataset.DatasetReference: the default dataset
```
tswast: This renders as a link to the property's class?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1384,20 +1368,39 @@ def from_api_repr(cls, resource):
...
 
-    destination = _TypedApiResourceProperty(
-        'destination', 'destinationTable', TableReference)
-    """
-    google.cloud.bigquery.table.TableReference: table where results are written
+        See
+        https://g.co/cloud/bigquery/docs/reference/v2/jobs#configuration.query.defaultDataset
+        """
+        prop = self._properties.get('defaultDataset')
+        if prop is not None:
```
tswast: The docstring should indicate that the value could possibly be None if unset.

Maybe:

```
google.cloud.bigquery.dataset.DatasetReference: a reference to
the default dataset ... or ``None`` if not set.
```

Same for the other properties, like `destination`.

bigquery/google/cloud/bigquery/job.py
alixhami: These don't render as links to the class. That line ends up getting formatted like this:

_google.cloud.bigquery.dataset.DatasetReference_ – the default dataset to use for unqualified table names in the query.

I can make it into a link by adding `:class:` before it, but that would depart from the Sphinx Google style [example](http://www.sphinx-doc.org/en/stable/ext/example_google.html).

bigquery/google/cloud/bigquery/job.py
alixhami: Agreed. Added in 55fad6c

PR #4790, opened 2018-01-26
BigQuery: samples for loading JSON from Cloud Storage.

tswast:
Samples for each API tab in https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json

Also: use `delete_contents=True` on dataset clean-up so that they are still cleaned up if there are errors loading the table.

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -607,7 +607,7 @@ def test_load_table_from_uri_truncate(client, to_delete):
     client.load_table_from_file(
         body, table_ref, job_config=job_config).result()
 
-    # [START bigquery_load_table_gcs_json_append]
+    # [START bigquery_load_table_gcs_json_truncate]
     # table_ref = client.dataset('my_dataset').table('existing_table')
     previous_rows = client.get_table(table_ref).num_rows
     assert previous_rows > 0
@@ -628,7 +628,7 @@ def test_load_table_from_uri_truncate(client, to_delete):
 
     assert load_job.state == 'DONE'
     assert client.get_table(table_ref).num_rows == 50
-    # [END bigquery_load_table_gcs_json_append]
+    # [END bigquery_load_table_gcs_json_truncate]
 
 
 def _write_csv_to_storage(bucket_name, blob_name, header_row, data_rows):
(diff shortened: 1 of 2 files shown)


docs/bigquery/usage.rst
```diff
@@ -185,11 +185,13 @@ Upload table data from a file:
...
-Load table data from Google Cloud Storage:
+Load table data from Google Cloud Storage (See also: `Loading JSON data from
+Google Cloud Storage
+<https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json>`_):
 
 .. literalinclude:: snippets.py
-   :start-after: [START load_table_from_uri]
-   :end-before: [END load_table_from_uri]
+   :start-after: [START bigquery_load_table_gcs_json]
+   :end-before: [END bigquery_load_table_gcs_json]
```
tseaver: We don't want to link in the autodetect / append / truncate cases?

docs/bigquery/snippets.py
```diff
@@ -495,53 +495,140 @@ def do_something(row):
...
     ]
-    HEADER_ROW = ('Full Name', 'Age')
-    bucket_name = 'gs_bq_load_test_{}'.format(_millis())
-    blob_name = 'person_ages.csv'
-    bucket, blob = _write_csv_to_storage(
-        bucket_name, blob_name, HEADER_ROW, ROWS)
-    to_delete.extend((blob, bucket))
-    DATASET_ID = 'delete_table_dataset_{}'.format(_millis())
-    dataset = bigquery.Dataset(client.dataset(DATASET_ID))
+    job_config.source_format = 'NEWLINE_DELIMITED_JSON'
```
jonparrott: Do we not have constants for things like this?

tswast: Not that I could find in https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/reference.html Filed https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4793

docs/bigquery/usage.rst
tswast: Done. Added a section for these.

The usage doc started to get pretty long, so I added a local table-of-contents, too.

Issue #4692, opened 2018-01-04
Drop usage of gax in favor of api_core


jonparrott:
We are trying to deprecate gax in favor of api_core, and to do so, we must remove all usage of gax here.

The following libraries still have references/use gax:

- [x] bigtable: only uses gax.metrics, should be trivial to fix. https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4873
- [x] datastore: has generated code that uses gax, and hand-written adapter over gapic client. Needs to be regenerated and the adapter code likely needs to be updated (one quick glance, most of it can be removed). #4877
- [x] error reporting: depends on the now-legacy separate gapic package. generated code needs to be co-located. Additionally has a small adapter like datastore which can mostly be removed. #4743
- [ ] firestore:  has generated code that uses gax. Needs to be regenerated. Has two small helpers for remapping gax errors (can be deleted) and adding metadata (will need to just be moved to the callsite). @chemelnucfin made a stab at this in the closed-without-merging PR #4880.
- [x] logging: depends on the now-legacy separate gapic package. generated code needs to be co-located. Additionally has an adapter like datastore which can mostly be removed. #4759
- [x] spanner: has generated code that uses gax. Needs to be regenerated. Has one helper that adds metadata which can be moved to the callsite. #4695
- [x] speech: has generated code that uses gax, and hand-written adapter over gapic client. Needs to be regenerated and the adapter code likely needs to be updated (one quick glance, most of it can be removed). This also has an error where `speech` is not a strict alias of `speech_v1`. #4788
- [x] trace: depends on the now-legacy separate gapic package. generated code needs to be co-located. Additionally has an adapter like datastore which can mostly be removed. https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4799

tseaver:
I'm pulling "priority: p2+" of issues labeled "type: cleanup" , because that label is supposed to be used for non-SLO-impacting stuff (normally things we open ourselves to handle at will later).

For this one, I don't know whether we should remove "type: cleanup" or "priority: p1", but one of them should go. :)

jonparrott:
How about we swap process for cleanup? :)

tseaver:
Sure. :)

chemelnucfin:
@lukesneeringer @jonparrott if one of you can kinda walk me through a bit of the process I can try to take some of these out.

I don't understand the autogenerator in tookit enough I think to do this, but I would be interested in understanding it.

[...]

PR #4796, opened 2018-01-26
Regenerate trace gapic, remove references to gax

jonparrott:
Towards #4692


jonparrott:
@chemelnucfin you indicated you were interested in helping with this - a great place to start is by fixing the tests for this library. You can use what I did in [logging](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4759/commits/c0ddd73a3ba316dd8856da0135f342a31712eaec) as a reference. Holler if you need help (but keep in mind I'm traveling over the next week).

chemelnucfin:
@jonparrott So far almost everything seems to be going great.  Seems like I just need to figure out how to replace the _GAXIterator with something else (GPRCIterator, I think).  Anyway, I'll keep looking at how it works, but just wanted to give you a heads up on it.

jonparrott:
I'd suggest you do what I did and just re-write those tests using channelstub. See here: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/8e3bbc11281ffffeb857e91e26be60254627cef7/logging/tests/unit/test__gapic.py#L59

Issue #4295, opened 2017-10-31
Packaging: "Guard" calls to `setuptools.setup()` behind `if __name__ == '__main__'`


dhermes:
This way our `setup.py` files could be imported (for example).

This is low-priority, but an easy change to make.

tseaver:
I'm a pretty big -1 on the idea that `setup.py` should ever be importable.

dhermes:
@tseaver I don't actually mind if it's import-able, but that was just an example. It's more just like good hygiene to have `if __name__ == '__main__'` in all scripts.

duggelz:
+1 on `__main__` hygiene across the board.

jonparrott:
`setup.py`, like `nox.py`, is a configuration file for a specific tool. We should not treat it as a Python module.

PR #4800, opened 2018-01-29
Guard setup with __name__ == '__main__'

chemelnucfin:
Partial #4295
The other part is in [toolkit/1828](https://github.com/googleapis/toolkit/pull/1828)


jonparrott:
We shouldn't do this

(Brevity because I'm traveling and such)

chemelnucfin:
ok

chemelnucfin:
should I close #4295 also then?

Issue #3897, opened 2017-08-29
How to create new Configs & Variables ?


DazWilkin:
The service supports and the documentation suggests that Configs and Variables resources creates methods (should) exist. Reviewing the documentation, I'm unable to find (expected?) create methods for these and reviewing the sources, it's unclear how these could be constructed.

Is it me? Or, are these methods missing?

"Create / interact with ... configs"
https://googlecloudplatform.github.io/google-cloud-python/latest/runtimeconfig/config.html

"Create / interact with ... variables"
https://googlecloudplatform.github.io/google-cloud-python/latest/runtimeconfig/variable.html

### Confirm
```
gcloud beta runtime-config configs list \
--project=${PROJECT}
```
```
NAME      DESCRIPTION
exists    Description
```
```
gcloud beta runtime-config configs variables list \
--config-name=exists \
--project=${PROJECT}
```
```
NAME       UPDATE_TIME
exists     2017-08-29T15:03:26.359064036Z
```

### Getters

```python
from google.cloud import runtimeconfig
client = runtimeconfig.Client()

config = client.config("exists")
print (config.exists())
print (config.get_variable("exists").exists())
```
Results:
```
True
True
```
### Setters?
```python
config = client.config("create")
print (config.exists())
config.create()
print (config.exists())

```
Results:

```
    config.create()
AttributeError: 'Config' object has no attribute 'create'
```
```python
print (config.exists())
variable = config.variable("test")
variable.create()
```
Results:
```
    variable.create()
AttributeError: 'Variable' object has no attribute 'create'
```

chemelnucfin:
@DazWilkin Hello, I apologize for the late response.

I looked over the issue, and the create functions are intended not to be there.  
You can use 
`config = client.config('config_name')`  to create a configuration 
https://googlecloudplatform.github.io/google-cloud-python/latest/runtimeconfig/client.html
and `variable = config.variable('variable_name')` to create a variable
https://googlecloudplatform.github.io/google-cloud-python/latest/runtimeconfig/config.html

Please feel free to reopen if you have any more questions.

kw:
Those will create the local object, not the config/variable in the cloud. The exists() methods of the created objects will return False if they haven't been created using `gcloud` or direct HTTP calls. Likewise, Variable doesn't expose any way to set the variable's value in the cloud.

[...]

PR #4810, opened 2018-01-30
Make Variable value compatible with text and value, returning whatever is set

### LINKED ISSUE #3898: Problems accessing Variables properties
using Cloud SDK, I'm able to construct a Configs resource, a Variables resource in it and set a text value but, when I try to access these properties through Cloud Client Libraries for Python, I'm unable to get the values:

```
gcloud beta runtime-config configs variables get-value my-variable \
--config-name=${CONFIG} \
--project=${PROJECT}
```
Returns:
```
my-value
```
```
gcloud beta runtime-config configs variables describe my-variable  \
--config-name=${CONFIG} \
--project=${PROJECT}
```
Returns:
```
atomicName: projects/$PROJECT/configs/$NAME/variables/my-variable
name: my-variable
text: my-value
updateTime: '2017-08-29T15:03:26.359064036Z'
```
And:
```python
from google.cloud import runtimeconfig

client = runtimeconfig.Client()

config = client.config(CONFIG)
print (config.exists())
print(config.get_variable(VARIABLE).exists())
print(config.get_variable(VARIABLE).full_name)
```
Returns:
```
True
True
projects/$PROJECT/configs/$NAME/variables/$VARIABLE
```
But:
```

print(config.get_variable(VARIABLE).state)
print(config.get_variable(VARIABLE).value)
```
Returns:
```
None
None
```

chemelnucfin:
I have assumed that `value` is always base64 encoded `bytes` and `text` is always `str`.  If that isn't a valid assumption then I may need to address that somehow, but it might have some ambiguous situations in python2.

Closes #3898

### DIFF
--- a/runtimeconfig/google/cloud/runtimeconfig/variable.py
+++ b/runtimeconfig/google/cloud/runtimeconfig/variable.py
@@ -116,19 +116,21 @@ class Variable(object):
 
     @property
     def value(self):
-        """Value of the variable, as bytes.
+        """Value of the variable
 
         See
         https://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/v1beta1/projects.configs.variables
 
-        :rtype: bytes or ``NoneType``
+        :rtype: bytes, str or ``NoneType``
         :returns: The value of the variable or ``None`` if the property
                   is not set locally.
         """
         value = self._properties.get('value')
-        if value is not None:
-            value = base64.b64decode(value)
-        return value
+        if value is None:
+            text = self._properties.get('text')
+            return text
+        else:
+            return base64.b64decode(value)
 
     @property
     def state(self):
--- a/runtimeconfig/tests/unit/test_variable.py
+++ b/runtimeconfig/tests/unit/test_variable.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright 2016 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


chemelnucfin:
@tswast Any pointers on how to use the autogen for runtime config?

It might solve #3897 also?

tswast:
Sorry, I have no idea how the autogen pipeline works. @jonparrott might know?

jonparrott:
Hrm, the protos are [published](https://github.com/googleapis/googleapis/tree/master/google/cloud/runtimeconfig/v1beta1) but it doesn't yet have an autogen configuration. It might be interesting for me to create one as a learning exercise. WDYT @lukesneeringer?

chemelnucfin:
How about putting the learning steps in the wiki?

tswast:
@jonparrott In the meantime, do you have an opinion on this change? I believe it improves usability, but unsure if the client libraries team wants to commit to a hand-written layer for RuntimeConfig.

jonparrott:
I'm -1 on this change. The fields are distinct in the API and have distinct types in Python (bytes/str), so I'd rather not do this.

chemelnucfin:
@jonparrott [3898](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3898#issuecomment-361432534) 

This was the reason I did the change like that.  I thought about having text and value functions to retrieve whatever was present, but that would violate the one way of doing things.  I also thought about having separate text and value functions to retrieve only the respective text or value if it existed, but that would require the user to keep track of text or value instead of just returning what was present, which might be better in terms of design, but at the cost of less convenience.  

So I thought that this was the better choice.

jonparrott:
I'm definitely more in favor of separate `text` and `value` getters than this.

kw:
As long as there's a sensible way to access `text` variables I'm happy.

tswast:
@jonparrott @chemelnucfin Does this mean there was a PR to add a separate getter / setter for `text` and `value` or is there still no way to get `text`?

jonparrott:
The original issue is still open, no PR has been made to add `text`.

chemelnucfin:
At some point I was going to look at how to write that yaml file for runtimeconfig with the protos.  Any pointers will be appreciated, otherwise I'd just try to compare the file with other ones.  But I have some other issues I want to work on first (even though I know this one is also important)

PR #4811, opened 2018-01-30
Vision updates.

lukesneeringer:
The changes are:

  * Updated autogen surface.
  * New web detection features: title, full matching images, partial matching images, visually similar images, best guess labels.

GAX is already excised from Vision.

### DIFF
--- a/vision/setup.py
+++ b/vision/setup.py
@@ -25,10 +25,10 @@ with io.open(os.path.join(PACKAGE_ROOT, 'README.rst'), 'r') as readme_file:
     readme = readme_file.read()
 
 REQUIREMENTS = [
-    'google-api-core>=0.1.0, <0.2.0dev',
-    'google-auth>=1.0.2, <2.0dev',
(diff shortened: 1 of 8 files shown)


vision/setup.py
```diff
@@ -25,10 +25,10 @@
     readme = readme_file.read()
 
 REQUIREMENTS = [
-    'google-api-core>=0.1.0, <0.2.0dev',
-    'google-auth>=1.0.2, <2.0dev',
-    'googleapis-common-protos[grpc]>=1.5.2, <2.0dev',
-    'requests>=2.18.4, <3.0dev',
+    'google-api-core >= 0.1.0, < 0.2.0dev',
+    'google-auth >= 1.0.2, < 2.0dev',
```
jonparrott: We should stop depending on google-auth in individual libraries and just rely on google-api-core to bring in the right version, likewise with request.

(You don't have to do it in this PR, but we should probably file a bug.)

lukesneeringer: https://github.com/googleapis/toolkit/pull/1831

PR #4816, opened 2018-01-31
Update index.rst

VikramTiwari:
fixes link for data transfer service


chemelnucfin:
Thank you for the pull request.

@lukesneeringer My only concern before merging this in is that is this some sort of autogen issue?

lukesneeringer:
Make it correct regardless. I will check on fixing the link in the source material.

PR #4817, opened 2018-01-31
Add max results to feature

### LINKED ISSUE #4272: Vision: Single feature functions generated by cloud vision client library does not support parameter max_results
As specified in [the gRPC reference](https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.AnnotateImageRequest), AnnotateImageRequest message have three fields, _image_, _features[]_ and _image_context_, in which the _features[]_ field determines what feature user is request (_type_) and max number of returned results (_max_results_). The code for generating single-feature functions (for example, _face_detection()_), however, does not allow user to specify _max_results_:

```
feature_value = {'type': enum.__dict__[feature]}
def inner(self, image, options=None, **kwargs):
        request = dict(
            image=image,
            features=[feature_value],
            **kwargs
        )
        return self.annotate_image(request, options=options)
```

Reported in https://github.com/GoogleCloudPlatform/python-docs-samples/issues/1173

chemelnucfin:
Closes #4272

### DIFF
--- a/vision/google/cloud/vision_helpers/decorators.py
+++ b/vision/google/cloud/vision_helpers/decorators.py
@@ -14,6 +14,8 @@
 
 from __future__ import absolute_import
 
+import copy
+
 
 def add_single_feature_methods(cls):
     """Custom decorator intended for :class:`~vision.helpers.VisionHelpers`.
@@ -99,20 +101,23 @@ def _create_single_feature_method(feature, enum):
     feature_value = {'type': enum.__dict__[feature]}
 
     # Define the function to be returned.
-    def inner(self, image, max_results=2**31-1,
+    def inner(self, image, max_results=None,
               retry=None, timeout=None, **kwargs):
         """Return a single feature annotation for the given image.
 
         Intended for use with functools.partial, to create the particular
         single-feature methods.
         """
-        feature_value['max_results'] = max_results
+        copied_features = copy.copy(feature_value)
+        if max_results is not None:
+            copied_features['max_results'] = max_results
         request = dict(
             image=image,
-            features=[feature_value],
+            features=[copied_features],
             **kwargs
         )
-        return self.annotate_image(request, retry=retry, timeout=timeout)
+        response = self.annotate_image(request, retry=retry, timeout=timeout)
+        return response
 
     # Set the appropriate function metadata.
     inner.__name__ = fx_name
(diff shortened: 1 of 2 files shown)


vision/google/cloud/vision_helpers/decorators.py
```diff
@@ -96,12 +99,14 @@ def _create_single_feature_method(feature, enum):
     feature_value = {'type': enum.__dict__[feature]}
 
     # Define the function to be returned.
-    def inner(self, image, retry=None, timeout=None, **kwargs):
+    def inner(self, image, max_results=2**31-1,
+              retry=None, timeout=None, **kwargs):
```
tseaver: It would be better to use `None` as the sentinel value, and just leave it out of the `feature_value` mapping below.

vision/tests/unit/test_decorators.py
```diff
@@ -65,12 +65,15 @@ def test_runs_generic_single_image(self, ai):
...
         assert isinstance(response, vision.types.AnnotateImageResponse)
 
         # Assert that the single-image method was called as expected.
         ai.assert_called_once_with({
-            'features': [{'type': vision.enums.Feature.Type.FACE_DETECTION}],
+            'features': [{'type': vision.enums.Feature.Type.FACE_DETECTION,
+                          'max_results': max_results}],
             'image': image,
             'image_context': SENTINEL,
         }, retry=None, timeout=None)
```
tseaver: Needs another testcase for when the user does not pass `max_results`.

vision/google/cloud/vision_helpers/decorators.py
```diff
@@ -99,20 +101,23 @@ def _create_single_feature_method(feature, enum):
...
-    def inner(self, image, max_results=2**31-1,
+    def inner(self, image, max_results=None,
               retry=None, timeout=None, **kwargs):
         """Return a single feature annotation for the given image.
 
         Intended for use with functools.partial, to create the particular
         single-feature methods.
         """
-        feature_value['max_results'] = max_results
+        copied_features = copy.copy(feature_value)
```
tseaver: I see we need to copy because `feature_value` is a "shared global" (well, in the scope of the closure).  It would be simpler / faster / clearer just to use `dict.copy`, though, e.g.:

```python
    copied_features = feature_value.copy()
```

chemelnucfin: done

chemelnucfin:
@dpebot merge on green

PR #4795, opened 2018-01-26
Allow datastore client to connect to emulator host.

ziplokk1:
Fix for issue #3920 
Updated PR for #4794 

Code Example:

```python
# This assumes you have the datastore emulator running on localhost:8081
from google.cloud import datastore

import os
os.environ['DATASTORE_EMULATOR_HOST'] = 'localhost:8081'

ds_client = datastore.Client()
key = ds_client.key('Testing', 'test')
entity = datastore.Entity(key)
entity.update({
    'hello': 'world'
})

ds_client.put(entity)
print(ds_client.get(key))
```

Issue #3920, opened 2017-09-05
Unable to connect to datastore emulator


patrick91:
Hi, I've been trying to use the datastore emulator without success, here are the steps I've used:

1. Start the datastore

```
$ gcloud beta emulators datastore start
WARNING: Reusing existing data in [/Users/patrick/.config/gcloud/emulators/datastore].
Executing: /Users/patrick/Documents/google-cloud-sdk/platform/cloud-datastore-emulator/cloud_datastore_emulator start --host=::1 --port=8822 --store_on_disk=True --consistency=0.9 --allow_remote_shutdown /Users/patrick/.config/gcloud/emulators/datastore
[datastore] Sep 05, 2017 9:55:20 PM com.google.cloud.datastore.emulator.CloudDatastore$FakeDatastoreAction$7 apply
[datastore] INFO: Provided --allow_remote_shutdown to start command which is no longer necessary.
[datastore] Sep 05, 2017 9:55:20 PM com.google.cloud.datastore.emulator.impl.LocalDatastoreFileStub <init>
[datastore] INFO: Local Datastore initialized:
[datastore] 	Type: High Replication
[datastore] 	Storage: /Users/patrick/.config/gcloud/emulators/datastore/WEB-INF/appengine-generated/local_db.bin
[datastore] Sep 05, 2017 9:55:20 PM io.grpc.internal.ManagedChannelImpl <init>
[datastore] INFO: [ManagedChannelImpl@cb0ed20] Created with target [::1]:8822
[datastore] Sep 05, 2017 9:55:20 PM com.google.cloud.datastore.emulator.impl.LocalDatastoreFileStub load
[datastore] INFO: The backing store, /Users/patrick/.config/gcloud/emulators/datastore/WEB-INF/appengine-generated/local_db.bin, does not exist. It will be created.
[datastore] Sep 05, 2017 9:55:20 PM io.gapi.emulators.netty.NettyUtil applyJava7LongHostnameWorkaround
[datastore] INFO: Unable to apply Java 7 long hostname workaround.
[datastore] API endpoint: http://::1:8822
[datastore] If you are using a library that supports the DATASTORE_EMULATOR_HOST environment variable, run:
[datastore]
[datastore]   export DATASTORE_EMULATOR_HOST=::1:8822
[datastore]
[datastore] Dev App Server is now running.
````
2. setting the environment variables

```
$ gcloud beta emulators datastore env-init
[...]

PR #4818, opened 2018-01-31
Fix for DS client not connecting to DS emulator host via env var.

ziplokk1:
Fix for issue #3920
Updated PR for #4795

Code Example:

```python
# This assumes you have the datastore emulator running on localhost:8081
from google.cloud import datastore

import os
os.environ['DATASTORE_EMULATOR_HOST'] = 'localhost:8081'

ds_client = datastore.Client()
key = ds_client.key('Testing', 'test')
entity = datastore.Entity(key)
entity.update({
    'hello': 'world'
})

ds_client.put(entity)
print(ds_client.get(key))
```

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -163,10 +163,12 @@ class EmulatorCreds(Credentials):
     """
     A mock credential object.
 
-    Used to avoid unnecessary token refreshing or reliance on the network when using an emulator.
+    Used to avoid unnecessary token refreshing or reliance on the network
+    when using an emulator.
     """
     # Credit to https://github.com/patrick91.
-    # Taken from https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3920#issuecomment-327315159
+    # Taken from https://github.com/GoogleCloudPlatform/google-cloud-python
+    # /issues/3920#issuecomment-327315159
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


ziplokk1:
Giving this another go.

I realized after deep diving that the logic was already there to accept the datastore emulator host but wasn't working without the credentials.

Just patched it up a little bit and added a test to make sure that it works without any default credentials.

I think I'm on the right track this time.

Sorry about all the PRs.

jonparrott:
Hi @ziplokk1. Thank you for putting this together, but I don't think we can accept this PR right now. We need to have a more broad conversation about how we want to support emulators. I'm going to keep the referenced issue open for discussion, but I'm going to close this for now.

ziplokk1:
Not a problem! I hope it can help someone out in the future.

drmarshall:
@theacodes -- has the discussion about emulators support taken place or at a clear workaround been documented? This is a pretty major stumbling block for a well-functioning CI and has been visible for 6 months.

theacodes:
Hmmm. Revisting this PR I would be open to merging something like this if it used `google.auth.credentials.AnonymousCredentials` instead of a custom class.

ziplokk1:
The code is merged into master on my fork if someone wants to change that functionality.

Saturn-V:
@theacodes -- Any updates regarding discussion for DS emulator support?

PR #4819, opened 2018-01-31
Fix variable nanosecond timestamp update_time from response

### LINKED ISSUE #4807: Runtimeconfig: Update time fails to parse
OS: Linux (Debian 4.9.65)
Python: 3.5.3 / 2.7.13
Version: google-cloud-runtimeconfig==0.28.0, google-cloud==0.32.0

Steps to reproduce:
1. Using gcloud, set a runtimeconfig variable.
2. In Python, load that variable
3. Try to access its update_time property.

```pycon
>>> from google.cloud import runtimeconfig
>>> client = runtimeconfig.Client()
>>> config = client.config('testconfig')
>>> var = config.get_variable('foo')
>>> var.update_time
  File "/redacted/lib/python3.5/site-packages/google/cloud/runtimeconfig/variable.py", line 160, in update_time
    value = _rfc3339_to_datetime(value)
  File "/redacted/lib/python3.5/site-packages/google/cloud/_helpers.py", line 274, in _rfc3339_to_datetime
    dt_str, _RFC3339_MICROS).replace(tzinfo=UTC)
  File "/usr/lib/python3.5/_strptime.py", line 510, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/usr/lib/python3.5/_strptime.py", line 343, in _strptime
    (data_string, format))
ValueError: time data '2018-01-22T21:39:44.095040522Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'
>>> var._properties
{'text': '43', 'updateTime': '2018-01-22T21:39:44.095040522Z'}
```

Observation: The `%f` format accepts microseconds, not fractions of arbitrary precision. That means six digits at most, but I see nine in the data.

chemelnucfin:
@tswast Thanks for the hint.

Closes #4807

### DIFF
--- a/runtimeconfig/tests/unit/test_config.py
+++ b/runtimeconfig/tests/unit/test_config.py
@@ -145,14 +145,14 @@ class TestConfig(unittest.TestCase):
         self.assertEqual(len(conn._requested), 0)
 
     def test_get_variable_w_bound_client(self):
-        from google.cloud._helpers import _rfc3339_to_datetime
+        from google.api_core import datetime_helpers
 
         VARIABLE_NAME = 'my-variable/abcd'
         VARIABLE_PATH = '%s/variables/%s' % (self.CONFIG_PATH, VARIABLE_NAME)
         RESOURCE = {
             'name': VARIABLE_PATH,
             'value': 'bXktdmFyaWFibGUtdmFsdWU=',  # base64 my-variable-value
-            'updateTime': '2016-04-14T21:21:54.5000Z',
+            'updateTime': '2016-04-14T21:21:54.123456789Z',
             'state': 'VARIABLE_STATE_UNSPECIFIED',
         }
         conn = _Connection(RESOURCE)
@@ -163,9 +163,10 @@ class TestConfig(unittest.TestCase):
 
         self.assertEqual(variable.name, VARIABLE_NAME)
         self.assertEqual(variable.full_name, VARIABLE_PATH)
+        DatetimeWithNanoseconds = datetime_helpers.DatetimeWithNanoseconds
         self.assertEqual(
             variable.update_time,
-            _rfc3339_to_datetime(RESOURCE['updateTime']))
+            DatetimeWithNanoseconds.from_rfc3339(RESOURCE['updateTime']))
         self.assertEqual(variable.state, RESOURCE['state'])
 
         self.assertEqual(len(conn._requested), 1)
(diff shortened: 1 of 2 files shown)


chemelnucfin:
Would a try except work better then or should we just wait?
```
try:
  return ms
except ValueError:
  return ns
```

tswast:
I'm emailing the team to check.

tseaver:
@tswast ping?

tswast:
I never got a response from the folks I've contacted before on the RuntimeConfig team.

@mgannholm Can you comment on this?

tswast:
Okay, I got confirmation from Victor on the RuntimeConfig team that "it might change from platform to platform". We need to be able to handle both.

Yes, try-except would be a good way to do it. Alternatively, I believe we have places (in BigQuery, maybe?) where we try to parse the string a little bit to see how many digits we have before trying to parse.

chemelnucfin:
I'll wait until the `datetime` PRs #4979, #4980, #4981 are merged before finishing this up.

chemelnucfin:
@tswast PTAL at your convenience.

runtimeconfig/tests/unit/test_config.py
```diff
@@ -145,14 +145,14 @@ def test_variable(self):
...
-        from google.cloud._helpers import _rfc3339_to_datetime
+        from google.api_core import datetime_helpers
 
         VARIABLE_NAME = 'my-variable/abcd'
         VARIABLE_PATH = '%s/variables/%s' % (self.CONFIG_PATH, VARIABLE_NAME)
         RESOURCE = {
             'name': VARIABLE_PATH,
             'value': 'bXktdmFyaWFibGUtdmFsdWU=',  # base64 my-variable-value
-            'updateTime': '2016-04-14T21:21:54.5000Z',
+            'updateTime': '2016-04-14T21:21:54.123456789Z',
```
tswast: Could we keep a test around that doesn't have nanosecond precision in case the API changes again?

chemelnucfin: There is one here:
[ms precision](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/runtimeconfig/tests/unit/test_variable.py#L162)

chemelnucfin: well, I'll write another one in variable.

chemelnucfin:
going to merge this and diff from this in the new PR.  (I have it written already)

Issue #3602, opened 2017-07-12
BigQuery data load fails from GCS on the first load.


lukesneeringer:
This issue is re-posted from an internal ticket. (Googlers: b/63330731)

--------------------------------

Using the google-cloud-python BigQuery client, loading data from GCS fails repeatably on first load due to schema issues, then succeeds on retry.  Using _any_ other method of BQ load from GCS, this behavior is not observed.  Occurs on version 0.24 and 0.25 of the BigQuery client.

GCS source objects are AVRO files.

The error indicates an incorrect extra field in the schema:

```python
[{u'reason': u'invalid', u'message': u'Provided Schema does not match Table <redacted>. Cannot add fields (field: com_<redacted>_<redacted>_Content_value).'}]
```

As a workaround, the customer reimplemented the BQ load from GCS process using JSON api, which also works correctly.

dhermes:
@lukesneeringer Any code snippet?

tseaver:
#2926 might be relevant.

lukesneeringer:
@dhermes Sorry, that was all they gave me. I can ask for a snippet, though.

lukesneeringer:
I have a snippet; sending out-of-band for privacy reasons.

lukesneeringer:
@dhermes @tseaver Did either of you have a chance to investigate this?

dhermes:
@lukesneeringer I did not

lukesneeringer:
Is that something you could do today or tomorrow?

dhermes:
I'd prefer to punt to @tseaver

chemelnucfin:
Would this be fixed by now?

chemelnucfin:
@lukesneeringer Would you be able to hand me that snippet so I can take a look?

chemelnucfin:
@lukesneeringer I believe that this has been fixed and confirmed via #4821.  Please feel free to reopen if it's not.  You may also be able to close the internal ticket also.  Thanks.

DrValani:
Is this released in google-cloud-bigquery==0.31.0. If so then, I am still getting this error.

PR #4821, opened 2018-02-01
BigQuery: Confirm loading avro from gcs is ok

chemelnucfin:
I believe this shows that #3602 is not a problem any more.

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -464,6 +464,41 @@ class TestBigQuery(unittest.TestCase):
         self.assertEqual(sorted(row_tuples, key=by_wavelength),
                          sorted(ROWS, key=by_wavelength))
 
+    def test_load_avro_from_uri_then_dump_table(self):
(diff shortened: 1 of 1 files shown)


bigquery/tests/system.py
```diff
@@ -564,7 +603,12 @@ def _write_csv_to_storage(self, bucket_name, blob_name, header_row,
                 blob.upload_from_file(csv_read, content_type='text/csv')
 
         self.to_delete.insert(0, blob)
+        return 'gs://{}/{}'.format(bucket_name, blob_name)
 
+    def _write_avro_to_storage(self, bucket_name, blob_name, avro_file):
```
tswast: There doesn't seem to be anything Avro-specific about this function. Should you be setting a content type as is done in the CSV case?

Per https://issues.apache.org/jira/browse/AVRO-488 the MIME-type `application/x-avro-binary` may be appropriate.

Issue #3359, opened 2017-05-02
Send trace context with logs from web applications


jonparrott:
When running on App Engine (and potentially other platforms in the future), we should send along the trace context when writing logs. This context is available in the HTTP request headers as `X-Cloud-Trace-Context` and should be passed in the structured log as the `traceId`. For more details, read [here](https://gist.github.com/JustinBeckwith/43dd04cfbc4551ddd8996a5b803da0b0).

Since this is a web request header, we will need to make the core of this web application-framework agnostic. We'll then need to write helpers/adaptors for Flask and Django.

To my knowledge, we have one blocking bug (#2997) that needs to be resolved before this can happen. Secondly, we need to agree on where the framework helpers should live. (@dhermes). I'm not aware of any other blockers.

Assigned to initially @waprin, as he currently has the most context and can hopefully address open questions and point out any additional pre-work that needs to be done. @liyanhui1228 and @duggelz are the product owners for this, and have offered to provide the implementation.

jonparrott:
Original context: #3355

waprin:
#2673 is a blocker too because since we're no longer using fluentd we will need to specify the monitoredresource.

jonparrott:
Also blocked by: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3377

jonparrott:
@duggelz @liyanhui1228  #2673 is resolved, you should be good to start on #2997. Holler if you need anything.

liyanhui1228:
OK sure.

jonparrott:
@liyanhui1228 https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3410 is merged, so this work should now be unblocked. Holler if you need anything. :)

liyanhui1228:
I have done this for flask and django. For flask the trace_id is just extracted from request headers, for django I used a middleware to store the request in thread local, and then get the trace_id from it. But the user will need to add the middleware to django settings before deploy the app. If this sounds reasonable, I can post a PR :) Any suggestions? @jonparrott @duggelz @waprin @dhermes

tseaver:
AFAICT, this issue was closed with the merge of #3448.

aatreya:
For flask (at least), this doesn't appear to be working as intended for us.

[...]

PR #4824, opened 2018-02-01
Move the code path of get_gae_labels() to emit()

### LINKED ISSUE #4823: [Logging] Move the code path of calling get_gae_labels() to emit()
Related with #3359. To make it easier to set the labels in each request without explicitly calling `handler.get_gae_labels()` or calling `client.setup_logging()` inside the request which causes a lot overhead, it's better to move the code path of getting the gae labels to the `emit()` function in the logging handler. And change the `AppEngineHandler` to extend the `logging.StreamHandler` directly instead of extending the `CloudLoggingHandler`.

liyanhui1228:
Closes #4823. See discussion in #3359.

### DIFF
--- a/logging/google/cloud/logging/handlers/app_engine.py
+++ b/logging/google/cloud/logging/handlers/app_engine.py
@@ -18,10 +18,10 @@ Sends logs to the Stackdriver Logging API with the appropriate resource
 and labels for App Engine logs.
 """
 
+import logging
 import os
 
 from google.cloud.logging.handlers._helpers import get_trace_id
-from google.cloud.logging.handlers.handlers import CloudLoggingHandler
 from google.cloud.logging.handlers.transports import BackgroundThreadTransport
 from google.cloud.logging.resource import Resource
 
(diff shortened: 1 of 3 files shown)


logging/google/cloud/logging/handlers/app_engine.py
```diff
@@ -34,7 +34,7 @@
 _TRACE_ID_LABEL = 'appengine.googleapis.com/trace_id'
 
 
-class AppEngineHandler(CloudLoggingHandler):
+class AppEngineHandler(logging.StreamHandler):
```
jonparrott: Can you explain why you changed the base class?

liyanhui1228: Because after moving the code path of `get_gae_labels()` to `emit()`, we need to override the `emit()`  method. So it's fine either to extend from `CloudLoggingHandler` or `logging.StreamHandler`， as the `AppEngineHandler` now have all the methods in `CloudLoggingHandler`, I just extend from `logging.StreamHandler` directly.

duggelz: So CloudLoggingHandler is just for GKE now?  And/or GCE?  Can you add a comment to CloudLoggingHandler clarifying this?

liyanhui1228: GKE has its own `ContainerEngineHandler`, in this [`get_default_handler`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/client.py#L295) method, the `CloudLoggingHandler` is returned if GAE and GKE environment are not detected.

jonparrott:
@liyanhui1228 thank you! do you want to cut a new release?

liyanhui1228:
@jonparrott Yeah thx!

Issue #4820, opened 2018-02-01
[Speech] TypeError: __call__() got multiple values for keyword argument 'timeout'


akmalhakimi1991:
1. Specify the API at the beginning of the title (for example, "BigQuery: ...")
   General, Core, and Other are also allowed as types
`google-cloud-speech`
2. OS type and version
`0.31.0`
3. Python version and virtual environment information `python --version`
`2.7.3`
4. google-cloud-python version `pip show google-cloud`, `pip show google-<service>` or `pip freeze`

> altgraph==0.15
> asn1crypto==0.23.0
> aspy.yaml==1.0.0
> backports-abc==0.5
> backports.ssl-match-hostname==3.5.0.1
> cached-property==1.3.1
> cachetools==2.0.1
> certifi==2018.1.18
> cffi==1.11.2
> chardet==3.0.4
> click==6.7
> cryptography==2.1.4
> dill==0.2.7.1
> dis3==0.1.2
> enum34==1.1.6
> future==0.16.0
> futures==3.2.0
> google-api-core==0.1.4
> google-auth==1.3.0
> google-cloud-core==0.28.0
> google-cloud-speech==0.31.0
> google-cloud-translate==1.3.0
> google-gax==0.15.16
> googleapis-common-protos==1.5.3
> grpcio==1.8.6
> identify==1.0.7
> idna==2.6
> ipaddress==1.0.18
> lxml==4.1.1
> macholib==1.9
> monotonic==1.4
> nodeenv==1.2.0
> pefile==2017.11.5
> pluggy==0.6.0
> ply==3.8
> pre-commit==1.4.5
> protobuf==3.5.1
> py==1.5.2
> pyasn1==0.4.2
> pyasn1-modules==0.2.1
> PyAudio==0.2.11
> pycparser==2.18
> PyInstaller==3.3.1
> pyOpenSSL==17.5.0
> pyttsx==1.1
> pytz==2017.3
> PyYAML==3.12
> requests==2.18.4
> rsa==3.4.2
> singledispatch==3.4.0.3
> six==1.11.0
> SpeechRecognition==3.8.1
> tenacity==4.7.1
> tornado==4.5.2
> tox==2.9.1
> translate==3.5.0
> urllib3==1.22
> virtualenv==15.1.0
> ws4py==0.4.2

5. Stacktrace if available

> Traceback (most recent call last):
>   File "transcribe_streaming_mic.py", line 195, in <module>
>     main()
[...]

PR #4828, opened 2018-02-02
Fix speech helpers to properly pass retry and timeout args

jonparrott:
Actual fix for #4820 (caused by regenerating the gapic layer)


chemelnucfin:
It would raise the actual GRPC error if the RPC is aborted?

I figured it was something related to gax/gapic, but I wasn't sure how to handle it without spending a lot of time on it.

Thanks.

jonparrott:
Yeah the problem is that the gapic regen changed the underlying interface
but we didn't catch that we needed to update the helpers as well.

PR #4833, opened 2018-02-03
BigQuery: Use constant strings for job properties

### LINKED ISSUE #4793: BigQuery: Document constants for things like NEWLINE_DELIMITED_JSON and move to bigquery module
See: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4790#discussion_r164172094

I think we used to have constants / enum-like classes for some of these things, but maybe they got lost in the rewrite. I can't find NEWLINE_DELIMITED_JSON in the [reference docs for the BigQuery client](https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/reference.html
).

**Feature request:** Add constants for commonly-used strings such as the source formats.

chemelnucfin:
Towards #4793

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -394,6 +394,9 @@ class TestBigQuery(unittest.TestCase):
 
     def test_load_table_from_local_file_then_dump_table(self):
         from google.cloud._testing import _NamedTemporaryFile
+        from google.cloud.bigquery.job import CreateDisposition
+        from google.cloud.bigquery.job import SourceFormat
+        from google.cloud.bigquery.job import WriteDisposition
 
         TABLE_NAME = 'test_table'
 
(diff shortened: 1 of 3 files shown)


tswast:
Could you also import them into https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/__init__.py ?

tswast:
Also, the reason for #4793 is that we weren't using the constants in https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/docs/bigquery/snippets.py , which needs to be updated, too. For the `snippets.py` usages, it will be best to use the `bigquery` module (my previous comment)

tswast:
Another reason this does not close #4833 

- The docs do not include the possible values. See: https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/reference.html#google.cloud.bigquery.job.WriteDisposition It defines the class, but none of the values are shown.

chemelnucfin:
@tswast This is my fault.  I remembered this PR, but I have been trying to work with pubsub and ended up just not wanting to switch into this and left it hanging too long.

PR #4835, opened 2018-02-04
fix docs

ofek:


chemelnucfin:
Thank you for your patch!  There was a misspelling, but I can fix that for you if you wish.

PR #4839, opened 2018-02-05
Pubsub: ignore creds kwarg if passed in for emulator

econchick:
Hey folks -

Developing on pubsub and running against the emulator, I noticed that pubsub client instantiation complains about both a channel and a credentials object being given. I'd prefer not to have to do the if/then dance within my own application code, so I hope y'all don't mind something like this in here.

LMK your thoughts! Thanks!


econchick:
(ah, hello @lukesneeringer ! Funny seeing you here. See you in a few months in Cleveland! 😄 )

jonparrott:
Thanks for doing this @econchick, but we really need to figure out how we're going to support emulators across the board before we do something like this.

Closing in favor of #4840, where we should hash this out.

PR #4843, opened 2018-02-06
Fix coveragerc to correctly omit generated files

jonparrott:


jonparrott:
@tseaver thanks. I merged because it's just a tooling change. :)

PR #4845, opened 2018-02-06
Add support for EncryptionConfiguration

vitaliili:
As part of a query, load job, or create table request, you can specify the encryption configuration, which just consists of a single string field containing the KMS key to use.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -659,8 +659,10 @@ class LoadJobConfig(object):
 
     @destination_encryption_configuration.setter
     def destination_encryption_configuration(self, value):
-        self._properties[
-            'destinationEncryptionConfiguration'] = value.to_api_repr()
+        api_repr = value
+        if value is not None:
+            api_repr = value.to_api_repr()
+        self._properties['destinationEncryptionConfiguration'] = api_repr
 
     def to_api_repr(self):
         """Build an API representation of the load job config.
@@ -985,8 +987,10 @@ class CopyJobConfig(object):
 
     @destination_encryption_configuration.setter
     def destination_encryption_configuration(self, value):
-        self._properties[
-            'destinationEncryptionConfiguration'] = value.to_api_repr()
+        api_repr = value
+        if value is not None:
+            api_repr = value.to_api_repr()
+        self._properties['destinationEncryptionConfiguration'] = api_repr
 
     def to_api_repr(self):
         """Build an API representation of the copy job config.
@@ -1410,8 +1414,10 @@ class QueryJobConfig(object):
 
     @destination_encryption_configuration.setter
     def destination_encryption_configuration(self, value):
-        self._properties[
-            'destinationEncryptionConfiguration'] = value.to_api_repr()
+        api_repr = value
+        if value is not None:
+            api_repr = value.to_api_repr()
+        self._properties['destinationEncryptionConfiguration'] = api_repr
 
     def to_api_repr(self):
         """Build an API representation of the copy job config.
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -693,10 +693,11 @@ class LoadJobConfig(object):
         config.skip_leading_rows = _int_or_none(slr)
         if config.skip_leading_rows is None:
             del config.skip_leading_rows
-        if 'destinationEncryptionConfiguration' in resource:
-            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
-            if key is not None:
-                config.kms_key_name = str(key)
+        if ('destinationEncryptionConfiguration' in resource
+            and 'kmsKeyName' in resource[
+                'destinationEncryptionConfiguration']):
+            config.kms_key_name = str(
+                resource['destinationEncryptionConfiguration']['kmsKeyName'])
         return config
 
 
@@ -999,10 +1000,11 @@ class CopyJobConfig(object):
         """
         config = cls()
         config._properties = copy.deepcopy(resource)
-        if 'destinationEncryptionConfiguration' in resource:
-            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
-            if key is not None:
-                config.kms_key_name = str(key)
+        if ('destinationEncryptionConfiguration' in resource
+            and 'kmsKeyName' in resource[
+                'destinationEncryptionConfiguration']):
+            config.kms_key_name = str(
+                resource['destinationEncryptionConfiguration']['kmsKeyName'])
         return config
 
 
@@ -1443,10 +1445,11 @@ class QueryJobConfig(object):
             if nested_resource is not None:
                 config._properties[prop] = from_resource(nested_resource)
 
-        if 'destinationEncryptionConfiguration' in resource:
-            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
-            if key is not None:
-                config.kms_key_name = str(key)
+        if ('destinationEncryptionConfiguration' in resource
+            and 'kmsKeyName' in resource[
+                'destinationEncryptionConfiguration']):
+            config.kms_key_name = str(
+                resource['destinationEncryptionConfiguration']['kmsKeyName'])
 
         return config
 
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -645,12 +645,12 @@ class LoadJobConfig(object):
     def kms_key_name(self):
         """ Resource ID of Cloud KMS key
 
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.destinationEncryptionConfiguration.kmsKeyName
+
         :rtype: str or None
         :returns: Resource ID of Cloud KMS key to encrypt destination table
                   or ``None`` if using default encryption.
-
-        See
-        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.destinationEncryptionConfiguration.kmsKeyName
         """
         return self._kms_key_name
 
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -676,6 +693,10 @@ def from_api_repr(cls, resource):
         config.skip_leading_rows = _int_or_none(slr)
         if config.skip_leading_rows is None:
             del config.skip_leading_rows
+        if 'destinationEncryptionConfiguration' in resource:
+            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
```
tswast: `resource['destinationEncryptionConfiguration']['kmsKeyName']` could cause a problem if `resource['destinationEncryptionConfiguration']` is an empty dictionary. Use `resource['destinationEncryptionConfiguration'].get('kmsKeyName')`

bigquery/google/cloud/bigquery/job.py
```diff
@@ -640,6 +641,19 @@ def schema(self, value):
...
+    def kms_key_name(self):
+        """See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.destinationEncryptionConfiguration.kmsKeyName
+        """
+        return self._kms_key_name
+
+    @kms_key_name.setter
+    def kms_key_name(self, value):
+        if not isinstance(value, six.string_types) and value is not None:
+            raise ValueError("Pass a string, or None")
```
tswast: Error message should say what property this is.

You could potentially drop the type check entirely, as in general we prefer to let the server do request validation.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -931,13 +960,30 @@ def __init__(self):
     https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.copy.writeDisposition
     """
 
+    @property
+    def kms_key_name(self):
+        """See
```
tswast: Add the type here and a one sentence description, too.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -953,6 +999,10 @@ def from_api_repr(cls, resource):
         """
         config = cls()
         config._properties = copy.deepcopy(resource)
+        if 'destinationEncryptionConfiguration' in resource:
+            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
```
tswast: Use `get('kmsKeyName)`, too

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1368,6 +1443,11 @@ def from_api_repr(cls, resource):
             if nested_resource is not None:
                 config._properties[prop] = from_resource(nested_resource)
 
+        if 'destinationEncryptionConfiguration' in resource:
+            key = resource['destinationEncryptionConfiguration']['kmsKeyName']
```
tswast: Use get('kmsKeyName), too

bigquery/google/cloud/bigquery/table.py
```diff
@@ -656,6 +677,10 @@ def from_api_repr(cls, resource):
 
         table = cls(dataset_ref.table(table_id))
         table._set_properties(resource)
+        if 'encryptionConfiguration' in resource:
+            key = resource['encryptionConfiguration']['kmsKeyName']
```
tswast: Use `get('kmsKeyName')`, too

bigquery/tests/unit/test_job.py
```diff
@@ -574,6 +585,34 @@ def test_from_api_repr_bare(self):
...
+                    'destinationEncryptionConfiguration': {
+                        'kmsKeyName': self.KMS_KEY_NAME
+                    }
+                }
+            },
+        }
+        klass = self._get_target_class()
+        job = klass.from_api_repr(RESOURCE, client=client)
+        self.assertIs(job._client, client)
+        self._verifyResourceProperties(job, RESOURCE)
```
tswast: I know the other tests are doing it, but I'd prefer if we don't use `_verifyResourceProperties` here. Instead, test that the `kms_key_name` property returns the expected result.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -640,6 +641,19 @@ def schema(self, value):
             raise ValueError('Schema items must be fields')
         self._schema = tuple(value)
 
+    @property
+    def kms_key_name(self):
+        """See
```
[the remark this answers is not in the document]
vitaliili: This is consistent with the rest of the file

bigquery/google/cloud/bigquery/job.py
vitaliili: This is consistent with the rest of the file

bigquery/google/cloud/bigquery/job.py
vitaliili: This is consistent with the rest of the file

bigquery/tests/unit/test_job.py
vitaliili: Adding an explicit check for these tests.

tswast:
@vitaliili the documentation build is failing. There's some info in https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/CONTRIBUTING.rst to get your dev environment set up.

The way to build the docs locally is to run `nox -s docs` from the root directory of the repo.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -676,6 +698,11 @@ def from_api_repr(cls, resource):
         config.skip_leading_rows = _int_or_none(slr)
         if config.skip_leading_rows is None:
             del config.skip_leading_rows
+        if ('destinationEncryptionConfiguration' in resource
```
jonparrott: Is it possible that there will be other `EncryptionConfiguration`s? If so, wouldn't it be unwise for us to reserve `destinationEncryptionConfiguration`'s `kmsKeyName` as the top-level `kms_key_name` property?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -918,6 +958,7 @@ class CopyJobConfig(object):
 
     def __init__(self):
         self._properties = {}
+        self._kms_key_name = None
```
jonparrott: What's special about this that it doesn't go under `properties` like most everything else?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -931,13 +972,35 @@ def __init__(self):
...
+        if using default encryption.
+
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.copy.destinationEncryptionConfiguration.kmsKeyName
+        """
+        return self._kms_key_name
+
+    @kms_key_name.setter
+    def kms_key_name(self, value):
+        if not isinstance(value, six.string_types) and value is not None:
```
jonparrott: @tswast do we do this kind of validation in other BigQuery stuff? We should generally avoid doing excessive type-checking or validation in the clients.

bigquery/google/cloud/bigquery/job.py
tswast: I asked that in the issue, also. https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4842

You're right. I'd wager there's going to be other `EncryptionConfigurations`. We should probably make a separate class to hold the various ones?

bigquery/google/cloud/bigquery/job.py
tswast: We used to do more and have been removing them as they've caused problems. Agreed that this check isn't necessary.

bigquery/google/cloud/bigquery/job.py
vitaliili: Refactored. Added EncryptionConfiguration class and respective properties.

bigquery/google/cloud/bigquery/job.py
vitaliili: Good point. Removed and refactored.

bigquery/google/cloud/bigquery/job.py
vitaliili: Removed and refactored.

tswast:
Note: I've added code sample snippets for all customer-managed encryption key operations in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4852, which should also serve well as system tests. (The tests all pass)

bigquery/google/cloud/bigquery/job.py
```diff
@@ -640,6 +641,27 @@ def schema(self, value):
...
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.destinationEncryptionConfiguration
+        """
+        prop = self._properties.get('destinationEncryptionConfiguration')
+        if prop is not None:
+            prop = EncryptionConfiguration.from_api_repr(prop)
+        return prop
+
+    @destination_encryption_configuration.setter
+    def destination_encryption_configuration(self, value):
+        self._properties[
```
jonparrott: Is it not possible to set this to `None`?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -76,6 +76,56 @@ def _view_use_legacy_sql_getter(table):
...
+class EncryptionConfiguration(object):
+    """Custom encryption configuration (e.g., Cloud KMS keys).
+
+    Args:
+        kms_key_name (str): resource ID of Cloud KMS key used for encryption
+    """
+
+    def __init__(self, **kwargs):
+        self._properties = {}
+        if 'kms_key_name' in kwargs:
```
jonparrott: why not just make this an actual kwarg?

tswast: Because I wanted it to be set to `None` if explicitly passed in as such.

bigquery/google/cloud/bigquery/job.py
tswast: You're right, this needs to be fixed.

bigquery/google/cloud/bigquery/table.py
jonparrott: You can still do that?

tswast: How do you tell the difference between None and unset with actual kwargs?

jonparrott: You don't, my question is why do you need to?

tswast: You're right, it shouldn't matter. It makes a difference in the JSON representation, but when it gets serialized to proto in the backend it doesn't make a difference.

bigquery/google/cloud/bigquery/job.py
tswast: Done.

jonparrott:
lgtm once you appease the ci

PR #4851, opened 2018-02-07
Implement `MergeOption` as an option

chemelnucfin:
Should close #4111
The first and second commits is #4654.

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1beta1/_helpers.py
@@ -763,7 +763,7 @@ def get_doc_id(document_pb, expected_prefix):
     return document_id
 
 
-def get_field_paths(update_data):
+def extract_field_paths(update_data):
     field_paths = []
     for field_name, value in six.iteritems(update_data):
         match = re.match(FieldPath.simple_field_name, field_name)
@@ -771,7 +771,7 @@ def get_field_paths(update_data):
             field_name = field_name.replace('\\', '\\\\').replace('`', '\\`')
             field_name = '`' + field_name + '`'
         if isinstance(value, dict):
-            sub_field_paths = get_field_paths(value)
+            sub_field_paths = extract_field_paths(value)
             field_paths.extend(
                 [field_name + "." + sub_path for sub_path in sub_field_paths])
         else:
@@ -898,7 +898,7 @@ def pbs_for_set(document_path, document_data, option):
         ),
     )
     if option is not None:
-        field_paths = get_field_paths(actual_data)
+        field_paths = extract_field_paths(actual_data)
         option.modify_write(
             update_pb, field_paths=field_paths, path=document_path)
 
--- a/firestore/google/cloud/firestore_v1beta1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1beta1/_helpers.py
@@ -24,10 +24,9 @@ from google.type import latlng_pb2
 import grpc
 import six
 
+from google.cloud import exceptions
 from google.cloud._helpers import _datetime_to_pb_timestamp
 from google.cloud._helpers import _pb_timestamp_to_datetime
-from google.cloud import exceptions
-
 from google.cloud.firestore_v1beta1 import constants
 from google.cloud.firestore_v1beta1.gapic import enums
 from google.cloud.firestore_v1beta1.proto import common_pb2
@@ -764,6 +763,22 @@ def get_doc_id(document_pb, expected_prefix):
     return document_id
 
 
+def get_field_paths(update_data):
+    field_paths = []
+    for field_name, value in six.iteritems(update_data):
+        match = re.match(FieldPath.simple_field_name, field_name)
+        if not (match and match.group(0) == field_name):
+            field_name = field_name.replace('\\', '\\\\').replace('`', '\\`')
+            field_name = '`' + field_name + '`'
+        if isinstance(value, dict):
+            sub_field_paths = get_field_paths(value)
+            field_paths.extend(
+                [field_name + "." + sub_path for sub_path in sub_field_paths])
+        else:
+            field_paths.append(field_name)
+    return field_paths
+
+
 def remove_server_timestamp(document_data):
     """Remove all server timestamp sentinel values from data.
 
@@ -876,7 +891,6 @@ def pbs_for_set(document_path, document_data, option):
         or two ``Write`` protobuf instances for ``set()``.
     """
     transform_paths, actual_data = remove_server_timestamp(document_data)
-
     update_pb = write_pb2.Write(
         update=document_pb2.Document(
             name=document_path,
@@ -884,7 +898,9 @@ def pbs_for_set(document_path, document_data, option):
         ),
     )
     if option is not None:
-        option.modify_write(update_pb)
+        field_paths = get_field_paths(actual_data)
+        option.modify_write(
+            update_pb, field_paths=field_paths, path=document_path)
 
     write_pbs = [update_pb]
     if transform_paths:
@@ -949,7 +965,7 @@ def pbs_for_update(client, document_path, field_updates, option):
         update_mask=common_pb2.DocumentMask(field_paths=sorted(field_paths)),
     )
     # Due to the default, we don't have to check if ``None``.
-    option.modify_write(update_pb)
+    option.modify_write(update_pb, field_paths=field_paths)
     write_pbs = [update_pb]
 
     if transform_paths:
--- a/firestore/google/cloud/firestore_v1beta1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1beta1/_helpers.py
@@ -43,10 +43,6 @@ FIELD_PATH_WRONG_TYPE = (
     'The data at {!r} is not a dictionary, so it cannot contain the key {!r}')
 FIELD_PATH_DELIMITER = '.'
 DOCUMENT_PATH_DELIMITER = '/'
-_NO_CREATE_TEMPLATE = (
-    'The ``create_if_missing`` option cannot be used '
-    'on ``{}()`` requests.')
-NO_CREATE_ON_DELETE = _NO_CREATE_TEMPLATE.format('delete')
 INACTIVE_TXN = (
     'Transaction not in progress, cannot be used in API requests.')
 READ_AFTER_WRITE_ERROR = 'Attempted read after write in a transaction.'
@@ -938,7 +934,7 @@ def pbs_for_update(client, document_path, field_updates, option):
     """
     if option is None:
         # Default uses ``exists=True``.
-        option = client.write_option(create_if_missing=False)
+        option = client.write_option(exists=True)
 
     transform_paths, actual_updates = remove_server_timestamp(field_updates)
     update_values, field_paths = FieldPathHelper.to_field_paths(actual_updates)
@@ -980,7 +976,7 @@ def pb_for_delete(document_path, option):
     """
     write_pb = write_pb2.Write(delete=document_path)
     if option is not None:
-        option.modify_write(write_pb, no_create_msg=NO_CREATE_ON_DELETE)
+        option.modify_write(write_pb)
 
     return write_pb
 
--- a/firestore/google/cloud/firestore_v1beta1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1beta1/_helpers.py
@@ -126,6 +126,13 @@ class FieldPath(object):
                 raise ValueError(error)
         self.parts = tuple(parts)
 
+    def __repr__(self):
+        paths = ""
+        for part in self.parts:
+            paths += "'" + part + "',"
+        paths = paths[:-1]
+        return 'FieldPath({})'.format(paths)
+
     @staticmethod
     def from_string(string):
         """ Creates a FieldPath from a unicode string representation.
@@ -510,6 +517,31 @@ def encode_dict(values_dict):
     }
 
 
+def extract_field_paths(actual_data):
+    """Extract field paths from actual data
+
+    Args:
+       actual_data (dict): The dictionary of the actual set data.
+
+    Returns:
+       List[~.firestore_v1beta1._helpers.FieldPath]:
+           A list of `FieldPath` instances from the actual data.
+    """
+    field_paths = []
+    for field_name, value in six.iteritems(actual_data):
+        if isinstance(value, dict):
+            sub_field_paths = extract_field_paths(value)
+            for sub_path in sub_field_paths:
+                paths = [field_name]
+                paths.extend(sub_path.parts)
+                field_path = FieldPath(*paths)
+                field_paths.append(field_path)
+        else:
+            path = FieldPath(field_name)
+            field_paths.append(path)
+    return field_paths
+
+
 def reference_value_to_document(reference_value, client):
     """Convert a reference value string to a document.
 
@@ -763,23 +795,7 @@ def get_doc_id(document_pb, expected_prefix):
     return document_id
 
 
-def extract_field_paths(update_data):
-    field_paths = []
-    for field_name, value in six.iteritems(update_data):
-        match = re.match(FieldPath.simple_field_name, field_name)
-        if not (match and match.group(0) == field_name):
-            field_name = field_name.replace('\\', '\\\\').replace('`', '\\`')
-            field_name = '`' + field_name + '`'
-        if isinstance(value, dict):
-            sub_field_paths = extract_field_paths(value)
-            field_paths.extend(
-                [field_name + "." + sub_path for sub_path in sub_field_paths])
-        else:
-            field_paths.append(field_name)
-    return field_paths
-
-
-def remove_server_timestamp(document_data):
+def remove_server_timestamp(document_data, top_level=True):
     """Remove all server timestamp sentinel values from data.
 
     If the data is nested, for example:
@@ -827,23 +843,24 @@ def remove_server_timestamp(document_data):
     actual_data = {}
     for field_name, value in six.iteritems(document_data):
         if isinstance(value, dict):
-            sub_field_paths, sub_data = remove_server_timestamp(value)
-            field_paths.extend(
-                get_field_path([field_name, sub_path])
-                for sub_path in sub_field_paths
-            )
+            sub_field_paths, sub_data = remove_server_timestamp(value, False)
+            for sub_path in sub_field_paths:
+                field_path = FieldPath.from_string(field_name)
+                field_path.parts = field_path.parts + sub_path.parts
+                field_paths.extend([field_path])
             if sub_data:
                 # Only add a key to ``actual_data`` if there is data.
                 actual_data[field_name] = sub_data
         elif value is constants.SERVER_TIMESTAMP:
-            field_paths.append(field_name)
+            if top_level:
+                field_paths.append(FieldPath(*field_name.split(".")))
+            else:
+                field_paths.append(FieldPath.from_string(field_name))
         else:
             actual_data[field_name] = value
-
-    if field_paths:
-        return field_paths, actual_data
-    else:
-        return field_paths, document_data
+    if not field_paths:
+        actual_data = document_data
+    return field_paths, actual_data
 
 
 def get_transform_pb(document_path, transform_paths):
@@ -860,6 +877,7 @@ def get_transform_pb(document_path, transform_paths):
         google.cloud.firestore_v1beta1.types.Write: A
         ``Write`` protobuf instance for a document transform.
     """
+    transform_paths = canonicalize_field_paths(transform_paths)
     return write_pb2.Write(
         transform=write_pb2.DocumentTransform(
             document=document_path,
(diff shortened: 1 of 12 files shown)


firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -768,6 +763,22 @@ def get_doc_id(document_pb, expected_prefix):
     return document_id
 
 
+def get_field_paths(update_data):
```
schmidt-sebastian: I named this `extract_field_paths` in the other SDKs. I think that's a much more descriptive name.

Furthermore, `update_data` is something different in the Firestore SDKs. Updata data is 
`{ fieldPath1 : value, fieldPath2: value }`. I think you are just trying to iterate a map of data here `{ field : value, field: value }`. That's `document_data`.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -880,15 +891,16 @@ def pbs_for_set(document_path, document_data, option):
...
     """
     transform_paths, actual_data = remove_server_timestamp(document_data)
-
     update_pb = write_pb2.Write(
         update=document_pb2.Document(
             name=document_path,
             fields=encode_dict(actual_data),
         ),
     )
     if option is not None:
```
schmidt-sebastian: This should only be true for `option.merge = true`

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -34,13 +34,13 @@
...
 from google.cloud.firestore_v1beta1.transaction import Transaction
+from google.cloud.firestore_v1beta1.proto import common_pb2
 
 
 DEFAULT_DATABASE = '(default)'
 """str: The default database used in a :class:`~.firestore.client.Client`."""
 _BAD_OPTION_ERR = (
-    'Exactly one of ``create_if_missing``, ``last_update_time`` '
-    'and ``exists`` must be provided.')
+    'Exactly one of ``last_update_time`` or ``exists`` must be provided.')
```
schmidt-sebastian: Add `merge` here.

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -422,46 +423,29 @@ def modify_write(self, write_pb, **unused_kwargs):
...
+            path (str): A fully-qualified document_path
+            unused_kwargs (Dict[str, Any]): Keyword arguments accepted by
+                other subclasses that are unused here.
         """
-        if no_create_msg is not None:
-            raise ValueError(no_create_msg)
-        elif not self._create_if_missing:
-            current_doc = types.Precondition(exists=True)
-            write_pb.current_document.CopyFrom(current_doc)
+        mask = common_pb2.DocumentMask(field_paths=sorted(field_paths))
```
schmidt-sebastian: Are you sorting for testing purposes? Would be good to point that out in a comment.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I'm actually recursing the `get_field_paths` (see line 774), so I want the actual `fieldPath` to the value as an api_repr.  So I think this is actually an `update_data`?

I think I wrote another method to extract both, but for this PR and what I needed here, the field path itself suffices.  

It seems like there are a few ways to represent the same thing:  In dictionary form `{'a': {'b': {1.2 : value}}}`, in api_repr form `{'a.b.(bt)1.2(bt) : value}`, and in python form `{FieldPath("a", "b", "1.2"): value}`, and I think at some point I will need to clean up all these functions.  

But it will probably be easier to make PRs one at a time for easier reviewing and then clean up later instead of writing them all in one PR.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: `field_paths` is ignored in all other options, and I had merge take and ignore any argument from what we discussed earlier.  Do you only want it to work for `merge=True` and ignore all other values?

firestore/google/cloud/firestore_v1beta1/client.py
chemelnucfin: ok.

firestore/google/cloud/firestore_v1beta1/client.py
chemelnucfin: ok.

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -422,46 +423,29 @@ def modify_write(self, write_pb, **unused_kwargs):
...
         Args:
             write_pb (google.cloud.firestore_v1beta1.types.Write): A
                 ``Write`` protobuf instance to be modified with a precondition
                 determined by the state of this option.
-            no_create_msg (Optional[str]): A message to use to indicate that
-                a create operation is not allowed.
-
-        Raises:
-            ValueError: If ``no_create_msg`` is passed.
+            field_paths (dict):
```
jba: Isn't field_paths a list of strings?

jba:
I'll leave it to Sebastian to approve.

firestore/google/cloud/firestore_v1beta1/client.py
chemelnucfin: right, fixed.

chemelnucfin:
Some of these PRs were made without making sure the conformance tests all pass, so I might have made some mistakes.  

I'm now working on making sure all of the conformance tests pass.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: @schmidt-sebastian I have passed all of the conformance tests locally.  I still have cleanups to do though.

Do you want me to do the entire thing in one new PR or would you like to merge this first and then work on the diffs from this one?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -131,6 +126,13 @@ def __init__(self, *parts):
                 raise ValueError(error)
         self.parts = tuple(parts)
 
+    def __repr__(self):
+        paths = ""
+        for part in self.parts:
+            paths += "'" + part + "',"
```
schmidt-sebastian: I would encourage you to just call `to_api_repr`. Field Paths in the API representation are in the simple case more readable (a.b versus 'a','b') and in the complex case more correct('a'.b versus ''a''.b).

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -515,6 +517,31 @@ def encode_dict(values_dict):
     }
 
 
+def extract_field_paths(actual_data):
```
schmidt-sebastian: `acutal_data` is `document_data` in the Node/Java SDK.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -816,23 +843,24 @@ def remove_server_timestamp(document_data):
...
+            for sub_path in sub_field_paths:
+                field_path = FieldPath.from_string(field_name)
+                field_path.parts = field_path.parts + sub_path.parts
+                field_paths.extend([field_path])
             if sub_data:
                 # Only add a key to ``actual_data`` if there is data.
                 actual_data[field_name] = sub_data
         elif value is constants.SERVER_TIMESTAMP:
-            field_paths.append(field_name)
+            if top_level:
```
schmidt-sebastian: Is `top_level` really `split_on_dots`?

I am a little bit concerned by your usage of strings in this method. `field_name` (which is really a field path) should be a FieldPath class so that this method can be called from an update method that takes FieldPaths.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -768,7 +795,7 @@ def get_doc_id(document_pb, expected_prefix):
     return document_id
 
 
-def remove_server_timestamp(document_data):
+def remove_server_timestamp(document_data, top_level=True):
```
schmidt-sebastian: This is really more of a `process_server_timestamps` that returns a `document_data` without the transforms. We also need a much more specific name than `field_paths` to describe the return type.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -938,9 +970,11 @@ def pbs_for_update(client, document_path, field_updates, option):
     """
     if option is None:
         # Default uses ``exists=True``.
-        option = client.write_option(create_if_missing=False)
+        option = client.write_option(exists=True)
 
     transform_paths, actual_updates = remove_server_timestamp(field_updates)
+    if not actual_updates:
```
schmidt-sebastian: It's fine to only specify server timestamps. There should be either `transform_paths` or `actual_updates`.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -880,21 +909,24 @@ def pbs_for_set(document_path, document_data, option):
...
     update_pb = write_pb2.Write(
         update=document_pb2.Document(
             name=document_path,
             fields=encode_dict(actual_data),
         ),
     )
     if option is not None:
-        option.modify_write(update_pb)
+        field_paths = extract_field_paths(actual_data)
+        option.modify_write(update_pb, field_paths=field_paths)
```
schmidt-sebastian: I still think we should only do this when merge is set.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I thought about doing something similar at first, but ideally __repr__ in python returns the construction of the object in string form.  In other words, instead of duplicating the functionality of `to_api_repr()`,
`repr(FieldPath("a", "b"))` should return a string `FieldPath("a", "b")`

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I'll take a look, but field_paths is ignored for the ExistsOption and LastUpdateOptions.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I think only updates are `split_on_dots` on top level, but sets aren't.  I'll take a look and maybe rename that better.

This was written before the FieldPath class, so I think it does take strings as the default.

schmidt-sebastian: You are correct that 'split_on_dots' only happens at the top level. You can always pass false when you invoke this function recursively.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: Done.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: Done.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: so would you like me to change the name `top_level` to `split_on_dots` or is `top_level` better?

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I have looked and tried to do this for `isinstance(option, MergeOption)` and resulted in no change.  I think I will just duck type this instead of checking for `isinstance`.  If other options will change in the future, we can deal with it when we need it easily.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: Done.  Is this ValueError check needed for pbs_for_set also?

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: Maybe you have some situation that I'm missing?

chemelnucfin: Looking at the api file, maybe I should implement the options differently.  But I think the behavior for this case would be correct.

I'll work on the options in a bit.

firestore/google/cloud/firestore_v1beta1/_helpers.py
schmidt-sebastian: Sounds good.

firestore/google/cloud/firestore_v1beta1/_helpers.py
schmidt-sebastian: I think `split_on_dots` is much clearer in its intend.

firestore/google/cloud/firestore_v1beta1/_helpers.py
schmidt-sebastian: If you update to the latest version of the conformance tests, you will no longer need it for set. Both set({}) and set({}, merge:true) should be allowed.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -802,37 +804,51 @@ def remove_server_timestamp(document_data):
...
-            field_paths.extend(
-                get_field_path([field_name, sub_path])
-                for sub_path in sub_field_paths
-            )
+            sub_transform_paths, sub_data, sub_field_paths = (
+                process_server_timestamp(value, False))
+            for sub_path in sub_transform_paths:
+                field_path = FieldPath.from_string(field_name)
+                field_path.parts = field_path.parts + sub_path.parts
+                transform_paths.extend([field_path])
```
schmidt-sebastian: Nit: Change all the naming in here to use 'transform' somewhere. That way, it is easy to tell the difference between this block and the one just below.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -802,37 +804,51 @@ def remove_server_timestamp(document_data):
...
+                for sub_field_path in sub_field_paths:
+                    field_path = FieldPath.from_string(field_name)
+                    field_path.parts = field_path.parts + sub_field_path.parts
+                    field_paths.append(field_path)
         elif value is constants.SERVER_TIMESTAMP:
-            field_paths.append(field_name)
+            if split_on_dots:
+                transform_paths.append(FieldPath(*field_name.split(".")))
+            else:
+                transform_paths.append(FieldPath.from_string(field_name))
```
schmidt-sebastian: Can you move the FieldPath logic much further up? If you have  transform the `field_name` into a FieldPath right as you enter the loop, then you can re-use it everywhere here.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -802,37 +804,51 @@ def remove_server_timestamp(document_data):
...
         else:
             actual_data[field_name] = value
-
-    if field_paths:
-        return field_paths, actual_data
-    else:
-        return field_paths, document_data
+            field_paths.append(FieldPath(field_name))
+    if not transform_paths:
```
schmidt-sebastian: Is this needed or just an optimization?

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -39,8 +39,8 @@
 DEFAULT_DATABASE = '(default)'
 """str: The default database used in a :class:`~.firestore.client.Client`."""
 _BAD_OPTION_ERR = (
-    'Exactly one of ``create_if_missing``, ``last_update_time`` '
-    'and ``exists`` must be provided.')
+    'Exactly one of ``last_update_time``, ``exists`` '
```
schmidt-sebastian: Nit: add 'or'

firestore/tests/system.py
```diff
@@ -145,18 +139,13 @@ def test_document_set(client, cleanup):
...
-        document.set({'no': 'way'}, option=option0)
-
-    assert exc_info.value.message.startswith(MISSING_DOCUMENT)
-    assert document_id in exc_info.value.message
+    # 0. Make sure the document doesn't exist yet
+    snapshot = document.get()
+    assert snapshot.to_dict() is None
 
-    # 1. Use ``set()`` to create the document (using an option).
+    # 1. Use ``create()`` to create the document (using an option).
```
schmidt-sebastian: Remove "(using an option)"

firestore/tests/system.py
```diff
@@ -211,11 +176,10 @@ def test_document_integer_field(client, cleanup):
             '7g': '8h',
             'cd': '0j'}
     }
-    option1 = client.write_option(exists=False)
-    document.set(data1, option=option1)
+    document.create(data1)
 
     data2 = {'1a.ab': '4d', '6f.7g': '9h'}
-    option2 = client.write_option(create_if_missing=True)
+    option2 = client.write_option(exists=True)
```
schmidt-sebastian: Passing 'exists=true' to create should be disallowed. Can you add a TODO somewhere?

firestore/tests/system.py
```diff
@@ -229,6 +193,39 @@ def test_document_integer_field(client, cleanup):
...
+    document_id = 'for-set' + unique_resource_id('-')
+    document = client.document('i-did-it', document_id)
+    # Add to clean-up before API request (in case ``set()`` fails).
+    cleanup(document)
+
+    # 0. Make sure the document doesn't exist yet
+    snapshot = document.get()
+    assert not snapshot.exists
+
+    # 1. Use ``set()`` to create the document (using an option).
```
schmidt-sebastian: Please update these comments. You are using "create" here with no option.

firestore/tests/unit/test_cross_language.py
```diff
@@ -27,19 +27,25 @@ class TestCrossLanguage(unittest.TestCase):
...
                 os.path.splitext(os.path.basename(test_filename))[0])
-            if test_proto.WhichOneof("test") == "get":
-                pass  # The Get tests assume a call to GetDocument, but Python
-                # calls BatchGetDocuments.
-                # TODO: make this work.
-            else:
+            try:
                 self.run_write_test(test_proto, desc)
+            except (AssertionError, Exception) as error:
+                count += 1
```
schmidt-sebastian: What does 'count' stand for? Can the name be more descriptive?

firestore/tests/unit/test_cross_language.py
```diff
@@ -59,11 +65,23 @@ def run_write_test(self, test_proto, desc):
...
+                tp.is_error
+            except AttributeError:
+                return
         elif kind == "set":
             tp = test_proto.set
             client, doc = self.setup(firestore_api, tp)
             data = convert_data(json.loads(tp.json_data))
-            # TODO: call doc.set.
+            if tp.HasField("option"):
+                option = True
```
schmidt-sebastian: Is option the SetOptions or just "merge" (in which case it should be `mergeOption`)?

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: just a simplification.

firestore/tests/system.py
chemelnucfin: This is an update.  I just checked that `document.create` cannot take options.

mikedh:
Hey, this finally made it to Pypi in a release, and unfortunately broke some things for me. This PR removes `firestore.CreateIfMissingOption`, which is still [recommended in the Firebase docs](https://firebase.google.com/docs/firestore/manage-data/add-data). The docs:

```
If the document does not exist, it will be created. If the document does exist, its contents will be overwritten with the newly provided data, unless you specify that the data should be merged into the existing document, as follows:

# The option to merge data is not yet available for Python. Instead, call the
# update method and pass the option to create the document if it's missing.

city_ref = db.collection(u'cities').document(u'BJ')

city_ref.update({
    u'capital': True
}, firestore.CreateIfMissingOption(True))
```

It looks like the migration here is (?):
```
city_ref.update({
    u'capital': True}, merge=True)
```

Given the arg ordering, an ugly backwards compatibility patch could be:
```
CreateIfMissingOption = lambda x : bool(x)
```

tseaver:
@mikedh The Firestore API team explicitly asked us to remove `CreateIfMissingOption` from the API surface of `google-cloud-firestore`.

@schmidt-sebastian Can you comment here?

samtstern:
That's the correct API change, just looks like it took a bit longer for it to land in Python than the other SDKs and we didn't time the doc updates with the release.  I'll get the docs updated.

samtstern:
Fixing here:
https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1767

mikedh:
Thanks! The API change seems like an improvement.

PR #4848, opened 2018-02-07
Release Speech 0.31.1

jonparrott:

PR #4854, opened 2018-02-08
Fix bad version number for speech

jonparrott:


jonparrott:
I'm going ahead and merging, this fixes a mistake from #4848.

PR #4860, opened 2018-02-09
Fix unit test for default `max_bytes` value

chemelnucfin:
See #4857


jonparrott:
@chemelnucfin note that I re-wrote your PR name.

PR #4870, opened 2018-02-12
WIP: pubsub large message

chemelnucfin:


chemelnucfin:
Closing for now.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -204,6 +204,8 @@ def _commit(self):
                 self._messages,
             )
             end = time.time()
+            self._client._publish_count += len(self._messages)
+            self._client._commit_count += 1
```
tseaver: Seems like this bit should be encapsulated by a method on the client.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -234,9 +236,10 @@ def monitor(self):
 
         # Sleep for however long we should be waiting.
         time.sleep(self._settings.max_latency)
-
         _LOGGER.debug('Monitor is waking up')
-        return self._commit()
+
+        if self._size:
+            return self._commit()
```
tseaver: This is potentially racy (e.g., if another thread calls `_commit()`.  Could we just make `_commit` bail early in this case?  Maybe drop the warning message, too.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -278,7 +281,8 @@ def publish(self, message):
 
         # Try to commit, but it must be **without** the lock held, since
         # ``commit()`` will try to obtain the lock.
-        if num_messages >= self._settings.max_messages:
+        if (num_messages >= self._settings.max_messages
+                or self._size >= self._settings.max_bytes):
```
tseaver: This is also racy.

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -130,16 +135,40 @@ def batch(self, topic, create=False, autocommit=True):
...
+            autocommit (bool): Whether to autocommit this batch. This is
+                primarily useful for debugging and testing, since it allows
+                the caller to avoid some side effects that batch creation
+                might have (e.g. spawning a worker to publish a batch).
+
+        Returns:
+            ~.pubsub_v1.batch.Batch: The batch object.
+        """
+        return self._batch(topic, create, autocommit)
+
```
tseaver: What is the point of the wrapper method here?  ISTM we could have just made `batch` take an optional `batch_settings` argument.

pubsub/nox.py
```diff
@@ -88,7 +88,9 @@ def system(session, py):
     session.install('.')
 
     # Run py.test against the system tests.
-    session.run('py.test', '--quiet', 'tests/system.py')
+    session.run('py.test',
+                '--quiet',
+                'tests/system.py')
```
tseaver: Might as well add in the `*session.posargs` while we're at it.

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -199,19 +228,26 @@ def publish(self, topic, data, **attrs):
...
+        if message_size > batch.settings.max_bytes:
+            with self._batch_lock:
+                backup = self._batches[topic]
+                settings = types.BatchSettings(message_size, 0.05, 1)
+                batch = self._batch(
+                    topic, create=True, batch_settings=settings)
+                future = None
+                while future is None:
+                    future = batch.publish(message)
+                self._batches[topic] = backup
```
tseaver: Rather than save - restore any existing batch for the topic, jjust create a new one, e.g.:

```python
       if message_size > batch.settings.max_bytes:
           with self._batch_lock:
               settings = types.BatchSettings(message_size, 0.05, 1)
               batch = self._batch_class(
                     autocommit=autocommit,
                     client=self,
                     settings=batch_settings,
                     topic=topic,
                 )
               future = None
               while future is None:
                   future = batch.publish(message)
```

We could then remove the whole `_batch()` vs. `batch()` bit altogether.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -72,7 +72,7 @@ def test_init_infinite_latency():
     assert batch._thread is None
 
 
-@mock.patch.object(threading, 'Lock')
+@mock.patch.object(threading, 'RLock')
 def test_make_lock(Lock):
```
tseaver: If we *are* going to use an `RLock` (see above), then change the name of the argument to match.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -203,6 +203,7 @@ def test_blocking__commit_wrong_messageid_length():
 
 def test_monitor():
     batch = create_batch(max_latency=5.0)
+    batch._size = 1
```
tseaver: We need to test the case that `batch._size` is zero, too.

pubsub/google/cloud/pubsub_v1/publisher/client.py
chemelnucfin: Luke mentioned about not allowing users to mess with the batch_settings.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -97,7 +97,7 @@ def make_lock():
         Returns:
             _thread.Lock: A newly created lock.
         """
-        return threading.Lock()
+        return threading.RLock()
```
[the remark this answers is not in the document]
chemelnucfin: I'm not 100% sure it works, but I'm 100% sure a regular lock didn't work.  Any suggestions?

pubsub/google/cloud/pubsub_v1/publisher/client.py
chemelnucfin: makes sense.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
chemelnucfin: you are right.  It actually does bail early already.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
chemelnucfin: got it.

PR #4872, opened 2018-02-12
Pubsub: Raise ValueError for message bytesize too big for batch

chemelnucfin:


pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -195,6 +195,10 @@ def publish(self, topic, data, **attrs):
 
         # Create the Pub/Sub message object.
         message = types.PubsubMessage(data=data, attributes=attrs)
+        if message.ByteSize() > self.batch_settings.max_bytes:
+            raise ValueError('Message being published is too large for the '
+                             'batch settings with max bytes {}.'.
```
jonparrott: Nit: please don't do hanging indents. Please line break after the opening `(` and indent once.

pubsub/tests/system.py
```diff
@@ -80,12 +80,40 @@ def test_publish_messages(publisher, topic_path, cleanup):
                 num=str(index),
             ),
         )
-
     for future in futures:
         result = future.result()
         assert isinstance(result, six.string_types)
 
 
+def test_publish_max_bytes(publisher, topic_path, cleanup):
```
jonparrott: This doesn't need to be a system test, it can likely just be a unit test.

chemelnucfin: you mean just delete it?

pubsub/google/cloud/pubsub_v1/publisher/client.py
chemelnucfin: I'd like to suggest to just leave it since the other two TypeErrors are the same way...

jonparrott: Please don't. :)

pubsub/tests/unit/pubsub_v1/publisher/test_publisher_client.py
```diff
@@ -132,6 +132,15 @@ def test_publish_data_not_bytestring_error():
         client.publish(topic, 42)
 
 
+def test_publish_data_too_large():
```
chemelnucfin: @jonparrott different from this one?

chemelnucfin:
@jonparrott PTAL, thanks.

pubsub/tests/unit/pubsub_v1/publisher/test_publisher_client.py
jonparrott: This is fine.

jonparrott:
@chemelnucfin the CI failure here seems unrelated, so feel free to merge - but, could you send a follow-up PR to fix that?

chemelnucfin:
@dpebot merge on green

PR #4877, opened 2018-02-13
Regenerate gapic for datastore and remove gax.

jonparrott:
Towards https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4692


jonparrott:
This was *too* easy...

PR #4878, opened 2018-02-13
Add support for streaming pull receipts.

### LINKED ISSUE #4325: pubsub: implement streaming pull receipt
StreamingPull currently has a "protocol" level bug affecting all languages.

TLDR: client libs should send modify ack deadline for all messages upon receiving them.

## Problem description
When streaming, `StreamingPullResponse` can be buffered, either by gRPC itself or by proxy/firewall.
There might be significant delay between the server sending a response and the client library reading it.
For this reason, by the time client library decides it should send a modify ack deadline ("modack"), it might be too late. The message might have already expired and resent.

Experimenting with Java client, we have seen many messages delivered 3-5 times because of this problem.

## What server-side is doing
To fix this, pubsub team is implementing the "receipt" feature.
The server will not start the expiration timer until it receives a receipt from the client [1]. A receipt here is similar to an ack. A receipt notifies the server that the client has *seen* the message, while an ack notifies that the client has finished with it.
This should make the client and server start expiration timer closer to each other.

[1] There's actually a limit on how long the server is willing to wait, so that a message isn't lost forever if the client doesn't send a receipt.

## What the client needs to do
The client should be modified to send these receipts.
For compatibility, these "receipts" are just modack requests. The client should modack messages after seeing it, in addition to what it is currently doing. These modacks can be batched for performance.

## Few more details for the curious
These modacks/receipts aren't strictly necessary for avoiding duplication, but it will help pubsub server expire messages on time. Without them, messages might expire too late.

Each `StreamingPullResponse` can contain multiple messages. In an implementation about to roll out, a modack for any message in the response will serve as receipts for all messages in that response. Still, client libs are encouraged to actually modack all messages since this behavior might change in the future.

chemelnucfin:
Closes #4325

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -331,6 +331,11 @@ class Policy(base.BasePolicy):
 
         For each message, schedule a callback with the executor.
         """
+        items = [
+            base.ModAckRequest(message.ack_id, self.histogram.percentile(99))
+            for message in response.received_messages
+        ]
+        self.modify_ack_deadline(items)
         for msg in response.received_messages:
             _LOGGER.debug(
                 'Using %s to process message with ack_id %s.',
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -336,7 +336,7 @@ class Policy(base.BasePolicy):
         redelivered multiple times.
 
         After the messages have all had their ack deadline updated, execute
-        the callback for each messaging using the executor.
+        the callback for each message using the executor.
         """
         items = [
             base.ModAckRequest(message.ack_id, self.histogram.percentile(99))
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -329,7 +329,14 @@ class Policy(base.BasePolicy):
     def on_response(self, response):
         """Process all received Pub/Sub messages.
 
-        For each message, schedule a callback with the executor.
+        For each message, send a modified acknowledgement request to the
+        server to prevent expiration of the message due to buffering by
+        gRPC or proxy/firewall. This allows the server and client
+        expiration timer to be closer to each other thus preventing the
+        message being redelivered multiple times.
+
+        After the `ModAckRequest` is delivered, execute callback with the
+        executor.
         """
         items = [
             base.ModAckRequest(message.ack_id, self.histogram.percentile(99))
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -330,13 +330,13 @@ class Policy(base.BasePolicy):
         """Process all received Pub/Sub messages.
 
         For each message, send a modified acknowledgement request to the
-        server to prevent expiration of the message due to buffering by
-        gRPC or proxy/firewall. This allows the server and client
-        expiration timer to be closer to each other thus preventing the
-        message being redelivered multiple times.
+        server. This prevents expiration of the message due to buffering by
+        gRPC or proxy/firewall. This makes the server and client expiration
+        timer closer to each other thus preventing the message being
+        redelivered multiple times.
 
-        After the `ModAckRequest` is delivered, execute callback with the
-        executor.
+        After the messages have all had their ack deadline updated, execute
+        the callback for each messaging using the executor.
         """
         items = [
             base.ModAckRequest(message.ack_id, self.histogram.percentile(99))
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
@@ -211,19 +211,15 @@ def test_on_response():
 
     # Actually run the method and prove that modack and executor.submit
     # are called in the expected way.
-    with mock.patch.object(thread.Policy, 'modify_ack_deadline') as modack:
+    modack_patch = mock.patch.object(
+        policy, 'modify_ack_deadline', autospec=True)
+    with modack_patch as modack:
         policy.on_response(response)
-        assert modack.called_with(
-            [base.ModAckRequest('fack', 10),
-             base.ModAckRequest('back', 10)]
-        )
 
-    requests = [{'modify_deadline_ack_ids':['fack'],
-                 'modify_deadline_seconds':[10]},
-                {'modify_deadline_ack_ids':['back'],
-                 'modify_deadline_seconds':[10]}]
-    calls = [mock.call(types.StreamingPullRequest(**requests[0])),
-             mock.call(types.StreamingPullRequest(**requests[1]))]
+    modack.assert_called_once_with(
+        [base.ModAckRequest('fack', 10),
+         base.ModAckRequest('back', 10)]
+    )
 
     submit_calls = [m for m in executor.method_calls if m[0] == 'submit']
     assert len(submit_calls) == 2
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
@@ -194,6 +194,7 @@ def test_on_response():
     # Set up the policy.
     policy = create_policy(executor=executor)
     policy._callback = callback
+    policy.modify_ack_deadline = mock.Mock()
 
     # Set up the messages to send.
     messages = (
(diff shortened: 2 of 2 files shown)


chemelnucfin:
test

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -346,13 +346,17 @@ def maintain_leases(self):
             _LOGGER.debug('Snoozing lease management for %f seconds.', snooze)
             time.sleep(snooze)
 
-    def modify_ack_deadline(self, ack_id, seconds):
+    def modify_ack_deadline(self, ack_id, seconds=None):
```
lukesneeringer: I would prefer this be explicitly sent to this method, not have it be a default.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -82,10 +82,14 @@ class Policy(base.BasePolicy):
...
             ``executor``.
+        ack_on_receipt (boolean): 
+            (Optional.) Whether to send `modify_ack_deadlines` upon 
+            receipt first at the cost of increased memory or to send receipts
+            as the messages get processed by callbacks.
     """
 
     def __init__(self, client, subscription, flow_control=types.FlowControl(),
-                 executor=None, queue=None):
+                 executor=None, queue=None, ack_on_receipt=False):
```
lukesneeringer: _Quo ante_, but _if_ this knob is to exist, default this to `True`.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -82,10 +82,14 @@ class Policy(base.BasePolicy):
         queue (~queue.Queue): (Optional.) A Queue instance, appropriate
             for crossing the concurrency boundary implemented by
             ``executor``.
+        ack_on_receipt (boolean): 
+            (Optional.) Whether to send `modify_ack_deadlines` upon 
+            receipt first at the cost of increased memory or to send receipts
```
lukesneeringer: _Quo ante_, but _if_ this knob is to exist:

```diff
-Whether to send `modify_ack_deadlines` upon
-receipt first at the cost of increased memory or to send receipts
-as the messages get processed by callbacks.
+Whether to send a `modify_ack_deadline` upon receipt of each message.
+The server uses this to know when to start the initial re-delivery timer.
```

The memory usage is trivial and the server behavioral difference is critical.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -82,10 +82,14 @@ class Policy(base.BasePolicy):
         queue (~queue.Queue): (Optional.) A Queue instance, appropriate
             for crossing the concurrency boundary implemented by
             ``executor``.
+        ack_on_receipt (boolean):
```
lukesneeringer: I do not think this knob should exist.

If this knob is to exist, it should be spelled `modack_on_receipt`. Ack and modack are _very_ different things.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -330,10 +335,47 @@ def on_response(self, response):
 
         For each message, schedule a callback with the executor.
         """
-        for msg in response.received_messages:
+        if self._ack_on_receipt:
```
lukesneeringer: This two separate functions thing is really confusing. The names mean basically nothing, and the bulk of the body is copied and pasted. Even if we had the knob, why is it not just:

```
if knob:
    knob_only_thing
other_things
```

pubsub/tests/system.py
```diff
@@ -99,35 +100,34 @@ def test_subscribe_to_messages(
...
-        time.sleep(1)
-
-        # The callback should have fired at least fifty times, but it
-        # may take some time.
-        if callback.calls >= 50:
-            return
-
-    # Okay, we took too long; fail out.
-    assert callback.calls >= 50
+    while len(set(callback.data.keys())) < messages:
```
lukesneeringer: Why did you change this? You took a test that had a potential failure case that would fail and made it into a failure case that would loop forever, and for no obvious gain that I can see.

Also, a `while/pass` is not guaranteed to surrender the GIL. (As in, it likely will not.)

pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
```diff
@@ -223,10 +226,32 @@ def test_on_response():
...
+                 'modify_deadline_seconds':[10]}]
+    calls = [mock.call(types.StreamingPullRequest(**requests[0])),
+             mock.call(types.StreamingPullRequest(**requests[1]))]
 
     submit_calls = [m for m in executor.method_calls if m[0] == 'submit']
-    assert len(submit_calls) == 2
-    for call in submit_calls:
+    assert len(submit_calls) == 4
+    if policy._ack_on_receipt:
+        # first submit acks then process
```
lukesneeringer: These _are not acks_.

pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
```diff
@@ -192,7 +192,9 @@ def test_on_exception_other():
         policy.future.result()
 
 
-def test_on_response():
+@pytest.mark.parametrize("ack_on_receipt", [True, False])
+def test_on_response(ack_on_receipt):
+    from google.cloud.pubsub_v1.subscriber import _consumer
```
lukesneeringer: The rest of the file does not do import-within-test (I do not believe).

pubsub/tests/system.py
```diff
@@ -187,12 +187,14 @@ class AckCallback(object):
...
         self.calls = 0
         self.lock = threading.Lock()
+        self.data = {}
 
     def __call__(self, message):
         message.ack()
         # Only increment the number of calls **after** finishing.
         with self.lock:
             self.calls += 1
+            self.data[message.attributes['num']] = message.data
```
lukesneeringer: But why?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
chemelnucfin: As long as it doesn't blow up memory I'm fine not having the knob also.

pubsub/tests/system.py
chemelnucfin: I thought that it was required to deliver the messages at least once and that "taking too long" shouldn't be a failure case. 

I guess I was wrong.  I'm changing the test back.

chemelnucfin:
I didn't intend for the `modack_on_receipt`  functionality to be optional, rather, whether to process them before or along with the callbacks (which I thought could possibly take a long time).  My wording could have been better though.

pubsub/tests/system.py
chemelnucfin: Could it be possible that I didn't get all of the messages at least once but have multiple duplicates?  I wasn't sure.

chemelnucfin:
@lukesneeringer I removed the extra "knob" as you suggested.  PTAL.  thanks.

jonparrott:
@chemelnucfin can you re-base this now that request batching is in?

chemelnucfin:
@jonparrott rebased

tseaver:
Is this intended to close #4325?  It should say so in the description (and the initial commit message).

Also, that issue was marked as `release blocking` but we released `pubsub-0.31.0` without merging this PR (yesterday).  The issue is also marked `triaged for GA`, so maybe this PR just needs to be merged before `pubsub-1.0.0`?  @jonparrott can you clarify?

chemelnucfin:
@tseaver Yes, I believe that it's 1.0.0 release blocking, not 0.31.0 release blocking.

chemelnucfin:
@tseaver I'll change the commit message before I merge it.

jonparrott:
Yes, "release blocking" in this instance means it's blocking GA. I released `0.31.0` yesterday because it contained two significant fixes to memory consumption problems.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -331,7 +331,12 @@ def on_response(self, response):
 
         For each message, schedule a callback with the executor.
         """
+        messages = []
         for msg in response.received_messages:
+            self._executor.submit(self.modify_ack_deadline, msg.ack_id)
+            messages.append(Message(msg.message, msg.ack_id, self._request_queue))
```
jonparrott: Why is this using the executor? It should be able to just call `self.modify_ack_deadline` - this is a very fast non-blocking operation (it just appends a tuple to a queue).

chemelnucfin: @jonparrott you are right.  I completely forgot about the change affects this.  I changed it.  Thanks.

pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
```diff
@@ -194,6 +194,7 @@ def test_on_response():
     # Set up the policy.
     policy = create_policy(executor=executor)
     policy._callback = callback
+    policy.modify_ack_deadline = mock.Mock()
```
jonparrott: I would use `patch.object` with autospeccing.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -331,6 +331,11 @@ def on_response(self, response):
 
         For each message, schedule a callback with the executor.
         """
+        items = [
```
jonparrott: Can you add a comment here that explains why we're doing this?

pubsub/tests/unit/pubsub_v1/subscriber/test_policy_thread.py
chemelnucfin: @jonparrott Does that work better?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
chemelnucfin: done.

lukesneeringer:
Approved _post hoc_. :-)

PR #4879, opened 2018-02-13
DLP

lukesneeringer:
This adds DLP v2beta1 to google-cloud-python.


lukesneeringer:
Waiting on CI; I am sure I will have missed something.

dlp/setup.py
```diff
@@ -0,0 +1,47 @@
...
+https://packaging.python.org/en/latest/distributing.html
+https://github.com/pypa/sampleproject
+"""
+
+from setuptools import setup, find_packages
+import io
+
+install_requires = [
+    'google-api-core>=0.1.0, <0.2.0dev',
+    'google-auth>=1.0.2, <2.0dev',
```
jonparrott: this should *only* depend on `google-api-core[grpc]`.

lukesneeringer:
@jonparrott Baton to you once CI passes.

jonparrott:
Merge away. I'll do the release in a separate PR.

Let's see if our bot will merge this time.

tseaver:
@lukesneeringer Is the cryptic `dlp` really the best name we can use here?  I see that it is used on cloud.google.com, so I guess the question is really moot.

jonparrott:
@tseaver yeah, that's going to be the name. I agree three-letter acronyms aren't the best.

lukesneeringer:
Agree with the complaint, and agree that it is not going to change.

lukesneeringer:
It also does not help that the name itself is not great. It is titled "Data Loss Prevention" and it does a bunch of redaction, PII detection, etc.

PR #4880, opened 2018-02-14
WIP Firestore: replace gax and regenerate gapic

chemelnucfin:
@jonparrott I haven't ran the tests yet, but something like this?

It looks like it's removed some enums that wasn't replaced.  Is that correct behavior?


chemelnucfin:
@jonparrott I have fixed the unit tests, but I haven't figured out what's causing the 504 `deadline_exceeded errors` in the system tests.  There're also some `gax` stuff in `gapic/firestore_admin_client.py` that I had to edit manually, which didn't seem right.

I'll probably head back to `pubsub` again and then come back to this.  If you see something obvious that I'm doing wrong, please let me know.  thanks.

chemelnucfin:
I also looked at the config files, but think that the commits should not be idempotent.  I did also try to retry the nonidempotent ones with `deadline_exceeded` though.

jonparrott:
@chemelnucfin no worries. Thanks for giving this a shot but pubsub is absolutely a higher priority.

chemelnucfin:
Closing for now

PR #4882, opened 2018-02-14
Add retry for yield_rows

aneepct:
Created retry for read rows. Using the retry code in retry.py, created a retry on yield_rows in table.py. If the retry flag is set on this method, create a retry wrapper and start reading from the last successful row key in a scan if an ErrorUnavailable exception is called during a previous scan.


jonparrott:
/cc @lukesneeringer 

@aneepct if we generate a gapic for this client which will make `client._data_stub.ReadRows` do retries on its own, does this PR become irrelevant?

zakons:
@jonparrott There is state information in the retry logic - last successful row key - that is used to start the read scan from just past that row key.  This is specific to the BigTable API, and from our discussions with @sduskis, gapic is operating at a level below the application specific protocol.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -317,7 +319,7 @@ def read_rows(self, start_key=None, end_key=None, limit=None,
         return PartialRowsData(response_iterator)
 
     def yield_rows(self, start_key=None, end_key=None, limit=None,
-                   filter_=None):
+                   filter_=None, retry=False):
```
sduskis: Please remove the option to not retry.  There is no case where a user doesn't want to retry.

jonparrott: The parameter should remain so that users can adjust the retry options, but it should probably have a sane default.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -532,8 +544,48 @@ def _do_mutate_retryable_rows(self):
...
+        self.last_scanned_key = last_scanned_key
+        self.end_key = end_key
+        self.filter_ = filter_
+        self.limit = limit
+
+    def __call__(self):
+        from google.cloud.bigtable.row import _retry_commit_exception
+
+        read_rows = functools.partial(self._do_read_retryable_rows)
+        retry_ = Retry(
```
jonparrott: This class should accept `retry` as a parameter.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -532,8 +544,48 @@ def _do_mutate_retryable_rows(self):
...
+                 end_key, filter_, limit):
+        self.client = client
+        self.table_name = table_name
+        self.last_scanned_key = last_scanned_key
+        self.end_key = end_key
+        self.filter_ = filter_
+        self.limit = limit
+
+    def __call__(self):
+        from google.cloud.bigtable.row import _retry_commit_exception
```
jonparrott: Can you put this import at the top of the file?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -532,8 +544,48 @@ def _do_mutate_retryable_rows(self):
...
+
+        read_rows = functools.partial(self._do_read_retryable_rows)
+        retry_ = Retry(
+            predicate=_retry_commit_exception,
+            deadline=30)
+        return retry_(read_rows)()
+
+    def _do_read_retryable_rows(self):
```
jonparrott: Hrm, after reading this I'm not convinced this class needs to exist as all. Could it just be a function applied with functools.partial?

bigtable/google/cloud/bigtable/table.py
sduskis: Retry should be on by default. This is a streaming response, and no user wants to re-retrieve what they already read. 

If we do need a parameter, then we need to have some sort of ReadRetyOptions class rather than a boolean.  Things like deadlines between streaming responses can be added to that ReadRetyOptions object.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -532,8 +544,48 @@ def _do_mutate_retryable_rows(self):
...
+        self.filter_ = filter_
+        self.limit = limit
+
+    def __call__(self):
+        from google.cloud.bigtable.row import _retry_commit_exception
+
+        read_rows = functools.partial(self._do_read_retryable_rows)
+        retry_ = Retry(
+            predicate=_retry_commit_exception,
+            deadline=30)
```
sduskis: A deadline doesn't make sense for a streaming read.  If we absolutely need one as a default, then perhaps something large like 20 minutes would do. This should be part of any ReadRetryOptions object we create.

Users can stream a lot of data, and might even read for an hour or so.

bigtable/google/cloud/bigtable/table.py
jonparrott: Please do not introduce a new class.

bigtable/google/cloud/bigtable/table.py
jonparrott: For `unary-stream` calls, retry applies to **establishing** the stream. Basically, this deadline says "if I can't start a stream after trying for 30 seconds, give up" not "if the stream is open for 30 seconds, give up".

bigtable/google/cloud/bigtable/table.py
```diff
@@ -340,17 +344,19 @@ def yield_rows(self, start_key=None, end_key=None, limit=None,
...
-        # We expect an iterator of `data_messages_v2_pb2.ReadRowsResponse`
-        generator = YieldRowsData(response_iterator)
-        for row in generator.read_rows():
+
+        retryable_read_rows = _RetryableReadRows(
+            self._instance._client, self.name, start_key,
+            end_key, filter_, limit, retry)
+
+        for row in retryable_read_rows().read_rows():
+            retryable_read_rows.last_scanned_key = row.row_key
```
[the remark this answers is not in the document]
sduskis: I think that`YieldRowData` should keep track of `last_scanned_key` instead of it being tracked in _RetryableReadRows.

`response.last_scanned_row_key` is not generally set.  `YieldRowData` should use `self._row.row_key` in the for loop.  `self._last_scanned_row_key = response.last_scanned_row_key` should only be set after the for loop, only if `response.last_scanned_row_key` is not None or empty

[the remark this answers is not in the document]
zakons: Agreed.  After testing, it appears that @sduskis is correct, the ```response.last_scanned_row_key``` is always None (hence, no idea what this was supposed to do).

In our code, ```self._last_scanned_row_key``` can be set to ```self._row.row_key``` after the yield in the following code:
```
                if chunk.commit_row:
                    self._save_current_cell()
                    yield self._row
```
After the yield, the client has been provided with the successfully read row, so it is safe to read from beyond this row key. Please confirm this makes sense, then we will refactor.

jonparrott:
Seems to make sense. Go for it.

bigtable/google/cloud/bigtable/table.py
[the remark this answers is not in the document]
jonparrott: Sounds reasonable. You should do that.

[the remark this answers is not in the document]
zakons: Thanks @jonparrott. Please note that the change to YieldRowsData will be done on PR #4899 since there is validation involved and the changes are localized to that code.  Based on the above comments, we are covered in this PR with the current implementation, where last_scanned_key is maintained in the retry logic.

bigtable/google/cloud/bigtable/table.py
zakons: The retry for read rows is under test on an endurance test server and we received the following StatusCode.DEADLINE_EXCEEDED exception.  It seems that Retry terminates the connection and causes our test code to abort, so there is more testing to be done.

It is hard to know what conditions cause this exception on the network, and whether this should be recoverable in Retry.  I suspect that the retry deadline of 30 seconds is too short.  We need to go into the Retry code itself to see how it treats this exception.

```
 File "/home/jenkins/.local/lib/python3.5/site-packages/google/cloud/bigtable/row_data.py", line 298, in read_rows
  File "/home/jenkins/.local/lib/python3.5/site-packages/grpc/_channel.py", line 348, in __next__
    return self._next()
  File "/home/jenkins/.local/lib/python3.5/site-packages/grpc/_channel.py", line 342, in _next
    raise self
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.DEADLINE_EXCEEDED, Error while reading table 'projects/grass-clump-479/instances/shared-perf/tables/scantest' : Response was not consumed in time; terminating connection. (Possible causes: slow client data read or network problems))>
```

bigtable/google/cloud/bigtable/table.py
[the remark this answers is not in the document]
jonparrott: I'm not sure what you mean.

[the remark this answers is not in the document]
zakons: Upon re-review with @sduskis, the last_scanned_row_key in a ReadRowsResponse was made clearer:

There are two things in a ReadRowsResponse:
1) a list of chunks
2) a variable literally called last_scanned_row_key

If there are no chunks, then last_scanned_row_key should be set.  See here for the protobuf definition: https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable.proto#L172

Based on this, the code is being refactored according to the comment from @sduskis above.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -243,7 +243,8 @@ def _delete_cells(self, column_family_id, columns, time_range=None,
 def _retry_commit_exception(exc):
     if isinstance(exc, grpc.RpcError):
         exc = exceptions.from_grpc_error(exc)
-    return isinstance(exc, exceptions.ServiceUnavailable)
+    return (isinstance(exc, exceptions.ServiceUnavailable) or
```
jonparrott: You can just do `isinstance(exc, (exceptions.ServiceUnavailable, exceptions.DeadlineExceeded))`.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -532,8 +536,55 @@ def _do_mutate_retryable_rows(self):
...
+            end_key=self.end_key,
+            filter_=self.filter_,
+            limit=self.limit,
+            start_inclusive=False)
+        client = self.client
+        response_iterator = client._data_stub.ReadRows(request_pb)
+        generator = YieldRowsData(response_iterator)
+        rows = []
+        for row in generator.read_rows():
+            self.last_scanned_key = row.row_key
```
jonparrott: I thought we were going to move this into `YieldRowsData`?

aneepct:
The code as committed should be correct. But not able to create at test case that provides 100% coverage. This code is currently running on an endurance test machine. The results of this testing will determine correctness of the implementation.

aneepct:
The test case for the read rows retry that is not working is test_yield_retry_rows.

It is using a _MockRetryableIterator, which should read a row successfully in order to get last_scanned_row_key, then throw an exception for next(). This should trigger retry to run the code again and then next() should return the second row successfully.

What is happening instead is that the exception is NOT being handled by Retry. This seems to be related to the fact that the exception is being thrown when yield is being used. It is peculiar to yield only, since it was tried without yield and the exception was handled properly in the Retry.

In order to reach 100% code coverage, the test case needs Retry to catch the exception and do retry EVEN when yield is used. 

Any suggestions on why yield behaves differently and what can be done to get this to work are welcome.

zakons:
@jonparrott Any suggestions on getting the code coverage to %100 with the issue we are seeing with exceptions being caught when yield is used?

jonparrott:
@zakons this is on my list to review, I just didn't have any time today. I'm going to try to get to it by the end of the week, but I may not be able to.

@tseaver might be able to take a look in my absence.

bigtable/google/cloud/bigtable/table.py
jonparrott: This comment still applies.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -27,18 +32,19 @@
...
 from google.cloud.bigtable.row_data import PartialRowsData
 from google.cloud.bigtable.row_data import YieldRowsData
-from grpc import StatusCode
 
 
 # Maximum number of mutations in bulk (MutateRowsRequest message):
 # (https://cloud.google.com/bigtable/docs/reference/data/rpc/
 #  google.bigtable.v2#google.bigtable.v2.MutateRowRequest)
 _MAX_BULK_MUTATIONS = 100000
+logging.getLogger().setLevel(logging.INFO)
```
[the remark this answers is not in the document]
zakons: This was removed.

bigtable/google/cloud/bigtable/table.py
zakons: To remove the issues with throwing an exception inside a generator, which breaks the generator, retry was moved INSIDE of YieldRowsData.read_rows.  When the iterator on ReadRowsResponses throws an exception, the iterator is recreated using the value of last_scanned_row_key.  See the new commit for details. 

There could be one final refactoring, to use Retry for the following:
```
```
To do so would introduce a method on YieldRowsData, which would have access to the state information needed to recreate the iterator (self.request_pb and self.client).  However, this should be run by @jonparrott first, since his preference is to create a simple functools.partial, as was done for MutateRow.  In our case, since we need to modify the request_pb with the last_scanned_row_key, this extra state information needs to be conveyed. The saved client is also needed to create the new iterator.  Please approve this design first.

zakons: @jonparrott The last refactoring was successfully applied, so Retry is being employed to do the retries on the ReadRowsReponse iterator.  The solution is now fully encapsulated within YieldRowsData, which updates the last_scanned_row_key, which is needed for starting the retry read rows from the last successful scan point.  Ready for full review now.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -286,6 +300,29 @@ def cancel(self):
         """Cancels the iterator, closing the stream."""
         self._response_iterator.cancel()
 
+    def create_retry_request(self):
```
jonparrott: should all these new methods be private?

jonparrott: In either case, they need docstrings.

bigtable/tests/unit/test_table.py
```diff
@@ -1231,6 +1246,24 @@ def __next__(self):  # pragma: NO COVER Py3k
...
+
+            def code(self):
+                return grpc.StatusCode.DEADLINE_EXCEEDED
+
+            def details(self):
+                return "Failed to read from server"
+
+        raise DeadlineExceeded()
+
+    def __next__(self):  # pragma: NO COVER Py3k
```
jonparrott: just do `__next__ = next`

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -286,6 +300,29 @@ def cancel(self):
...
+        row_range = self.request_pb.rows.row_ranges.pop()
+        range_kwargs = {}
+        range_kwargs['start_key_open'] = self.last_scanned_row_key
+        range_kwargs['end_key_open'] = row_range.end_key_open
+        self.request_pb.rows.row_ranges.add(**range_kwargs)
+
+    def retry_read_rows(self):
+        if (self.last_scanned_row_key and self.request_pb is not None):
```
jonparrott: Can you help me understand the logic here?

bigtable/tests/unit/test_table.py
aneepct: @jonparrott This change is not working on python 2.7.

bigtable/google/cloud/bigtable/row_data.py
zakons: Please see the comments added on how last_scanned_row_key is used to restart the scan.  Also, the helper methods were made private and docstrings added.

This latest revision has been running for several days on our endurance machine and we have not seen any of the read_rows failures previously seen, so the testing looks good. If you are OK with the changes, can we merge these?

bigtable/google/cloud/bigtable/row_data.py
zakons: Done.

bigtable/tests/unit/test_table.py
jonparrott: err, rather, the other way around.

bigtable/google/cloud/bigtable/row_data.py
jonparrott: > Please see the comments added on how last_scanned_row_key is used to restart the scan. Also, the helper methods were made private and docstrings added.

You added two new args to the constructor and didn't document them. These argument appear to significantly change the behavior of the class. Can you update the docstring to describe both the arguments and the behavior when those arguments are specified?

> This latest revision has been running for several days on our endurance machine and we have not seen any of the read_rows failures previously seen, so the testing looks good. If you are OK with the changes, can we merge these?

That's great, but I want to understand these changes before we merge.

zakons: Added the additional parameter documentation and usage, sorry for omission. Note that these were added as optional parameters to maintain backward compatibility.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -234,25 +239,42 @@ def consume_all(self, max_loops=None):
...
+                    retry to create a new iterator.
     """
 
     START = 'Start'  # No responses yet processed.
     NEW_ROW = 'New row'  # No cells yet complete for row
     ROW_IN_PROGRESS = 'Row in progress'  # Some cells complete for row
     CELL_IN_PROGRESS = 'Cell in progress'  # Incomplete cell for row
 
-    def __init__(self, response_iterator):
+    def __init__(self, response_iterator, request_pb=None, client=None):
```
jonparrott: A couple of things here:

1. Can you just call `request_pb` `request`. It's *not* a protobuf, it's a message, so if you're going to postfix it should be `request_msg`, but there's no need for all that so just `request` will do.
2. Instead of passing the client, just pass in the bound method to execute the request. Call it `method` and pass it in like `YieldRowsData(..., method=client._data_stub.ReadRows)`. This will make the class more flexible and easier to test.
3. It seems you're going out of your way not to break this classes interface and it's actually making things more complex here. This client is not GA, could we consider breaking this so the interface is simply `YieldRowsData(method, request)`? That way, this class would require far less branches to accomplish its task.

zakons: OK, agreed that this would clear up the interfaces. Will need to publish release notes to make the new interface changes clear to users.  Starting the changes, which will require updates to the unit test cases and will also alter PartialRowsData to be consistent with the new interface.  This will provide the additional benefit of properly supporting retry in PartialRowsData.

zakons: Latest commits use the suggestions above and the code is now consistent for the cases that used an iterator as a parameter previously. In the Table class, read_row, read_rows and yield_rows all use the suggested change to pass the ReadRows method and ReadRowsRequest as input parameters to the classes implementing the requests.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -284,7 +309,36 @@ def state(self):
...
+
+        self.response_iterator = self.read_method(self.request)
+
+    def _read_next(self):
+        """Helper for :meth:`read_rows`."""
+        return six.next(self.response_iterator)
+
+    def _read_next_response(self):
+        """Helper for :meth:`read_rows`."""
+        next_response = functools.partial(self._read_next)
```
jonparrott: I don't think you actually need to use `functools.partial` here anymore.

zakons: Thanks! Your interface recommendations made a big improvement to the design.  Also, I think we finally understand the full range of features in Retry. For this implementation we needed ```on_error```.

We will take a look at the last review comment about partial and see if this can come out.

jonparrott:
@aneepct @zakons are y'all ready for me to merge this?

zakons:
@jonparrott Yes! The code is looking good and our endurance test did not show any uncaught, retryable exceptions.  Thanks again for your valuable review suggestions.

PR #4887, opened 2018-02-15
Add client configuration from dictionary

mdhausman:
I've added this change that we implement in our own code, as it may be more broadly useful. As we don't store our service account JSON in plaintext, we can't use the `from_service_account_json` convenience method.

This PR separates out some of the logic from `from_service_account_json` into a separate `from_dict` convenience method.

- split `_ClientFactoryMixin.from_service_account_json`
    - add `from_dict` to offer users a hook in without needing to load from plaintext JSON
- add new test for completeness, though test coverage would be unaffected

### DIFF
--- a/core/google/cloud/client.py
+++ b/core/google/cloud/client.py
@@ -67,15 +67,40 @@ class _ClientFactoryMixin(object):
         :raises TypeError: if there is a conflict with the kwargs
                  and the credentials created by the factory.
         """
-        if 'credentials' in kwargs:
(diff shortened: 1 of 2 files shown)


jonparrott:
Hi @mdhausman, I super appreciate you doing this, but I don't want to touch anything in `core` until https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4267 is done. As part of #4267 we want to eliminate this class altogether, but I'll be sure to consider including similar functionality when we do that.

PR #4892, opened 2018-02-16
Add pending request backpressure for subscriber.

jonparrott:
This change includes the consumer's pending request backlog in the policy's load calculations. This allows the policy to pause (and resume) the response stream if there are a large number of outstanding requests to be send on the stream. This additionally adds `max_requests` to `FlowControl` to allow tweaking this number.

Resolves: #4792
Related: #4841

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -187,6 +187,10 @@ class Consumer(object):
     def active(self):
         """bool: Indicates if the consumer is active.
 
+        *Active* means that the stream is open and that it is possible to
+        send and receive messages. This is distinct from *pausing* which just
+        pauses *response* consumption.
+
         This is intended to be an implementation independent way of indicating
         that the consumer is stopped. (E.g. so a policy that owns a consumer
         doesn't need to know what a ``threading.Event`` is.)
@@ -201,6 +205,14 @@ class Consumer(object):
         """
         with self._put_lock:
             self._request_queue.put(request)
+
+    @property
+    def pending_requests(self):
+        """int: An approximate count of the outstanding requests.
+
+        This can be used to determine if the consumer should be paused if there
+        are too many outstanding requests."""
+        return self._request_queue.qsize()
 
     def _request_generator_thread(self, policy):
         """Generate requests for the stream.
@@ -231,8 +243,9 @@ class Consumer(object):
                 _LOGGER.debug('Request generator signaled to stop.')
                 break
 
-            _LOGGER.debug('Sending request:\n%r', request)
+            _LOGGER.debug('Sending request on stream')
             yield request
+            policy.on_request(request)
 
     def _stop_request_generator(self, request_generator, response_generator):
         """Ensure a request generator is closed.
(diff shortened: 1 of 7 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -334,14 +347,14 @@ def _blocking_consume(self, policy):
                 response_generator, self._can_consume)
             try:
                 for response in responses:
-                    _LOGGER.debug('Received response:\n%r', response)
```
lukesneeringer: I did actually need these for debugging at a previous point. Perhaps still send them with a envvar or something? (Not critically important, no need to block the PR on them.)

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -332,8 +318,7 @@ def on_response(self, response):
         """
         for msg in response.received_messages:
             _LOGGER.debug(
-                'Using %s to process new message received:\n%r',
-                self._callback, msg)
+                'Using %s to process message with ack_id %s.',
+                self._callback, msg.ack_id)
             message = Message(msg.message, msg.ack_id, self._request_queue)
-            future = self._executor.submit(self._callback, message)
```
lukesneeringer: Can you explain what this change fixes? It looks suspect for unintended consequences.

pubsub/google/cloud/pubsub_v1/types.py
```diff
@@ -53,12 +53,13 @@
 # The defaults should be fine for most use cases.
 FlowControl = collections.namedtuple(
     'FlowControl',
-    ['max_bytes', 'max_messages', 'resume_threshold'],
+    ['max_bytes', 'max_messages', 'resume_threshold', 'max_requests'],
```
lukesneeringer: So, to be clear, this will pause the **incoming** stream if the **outgoing** stream becomes congested?

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
jonparrott: This actually causes unnecessary memory and CPU to be consumed just for a log output, especially with a large amount of pending messages..

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
jonparrott: This can cause a large amount of CPU to be consumed as Python converts the response to a string every time regardless of if logging is turned on.

pubsub/google/cloud/pubsub_v1/types.py
jonparrott: right.

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
lukesneeringer: Ack.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
lukesneeringer: Ack.

PR #4895, opened 2018-02-16
Add ability for subscriber to batch requests.

### LINKED ISSUE #4893: Pub/Sub subscriber should batch requests
Presently, the pub/sub subscriber only batches modacks, other actions are sent in the stream as single requests. This can lead to the request backlog growing (though this is less concerning with #4892).

Other clients (Go, for instance) actually batch the other actions. We should attempt to do so as well.

We have two options that I can see right now:

1. Make `QueueCallbackThread` use a variant of `_get_many` to pull multiple items off the queue and then modify `dispatch_callback` to combine actions. We'd have to make the action functions (`ack`, `nack`, etc.) accept multiple items instead of just a single item.
2. Make `dispatch_callback` keep a list of items and occasionally submit them to the appropriate action functions. We'd still need to modify the action functions to accept multiple items. We'd have to have yet another thread to deal with submitting the items on a cadence or modify `QueueCallbackThread` to do so.

I'm leaning towards (1).

jonparrott:
The size of the batches and the rate at which they are submitted is controlled by two new flow control settings, `max_request_batch_size` and `max_request_batch_latency`.

Closes #4893

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 import logging
+import time
 import uuid
+
+from six.moves import queue
 
 
 __all__ = (
@@ -31,6 +34,36 @@ _LOGGER = logging.getLogger(__name__)
 STOP = uuid.uuid4()
 
 
+def _get_many(queue_, max_items=None, max_latency=0):
+    """Get multiple items from a Queue.
+
+    Gets at least one (blocking) and at most ``max_items`` items
+    (non-blocking) from a given Queue. Does not mark the items as done.
+
+    Args:
+        queue_ (~queue.Queue`): The Queue to get items from.
+        max_items (int): The maximum number of items to get. If ``None``, then
+            all available items in the queue are returned.
+        max_latency (float):  The maximum number of seconds to wait for more
+            than one item from a queue. This number includes the time required
+            to retrieve the first item.
+
+    Returns:
+        Sequence[Any]: A sequence of items retrieved from the queue.
+    """
+    start = time.time()
+    # Always return at least one item.
+    items = [queue_.get()]
+    while max_items is None or len(items) < max_items:
+        try:
+            elapsed = time.time() - start
+            timeout = max(0, max_latency - elapsed)
+            items.append(queue_.get(timeout=timeout))
+        except queue.Empty:
+            break
+    return items
+
+
 class QueueCallbackWorker(object):
     """A helper that executes a callback for every item in the queue.
 
@@ -42,27 +75,42 @@ class QueueCallbackWorker(object):
             concurrency boundary implemented by ``executor``. Items will
             be popped off (with a blocking ``get()``) until :attr:`STOP`
             is encountered.
-        callback (Callable[[str, Dict], Any]): A callback that can process
-            items pulled off of the queue. Items are assumed to be a pair
-            of a method name to be invoked and a dictionary of keyword
-            arguments for that method.
+        callback (Callable[Sequence[Any], Any]): A callback that can process
+            items pulled off of the queue. Multiple items will be passed to
+            the callback in batches.
+        max_items (int): The maximum amount of items that will be passed to the
+            callback at a time.
+        max_latency (float): The maximum amount of time in seconds to wait for
+            additional items before executing the callback.
     """
 
-    def __init__(self, queue, callback):
+    def __init__(self, queue, callback, max_items=100, max_latency=0):
         self.queue = queue
         self._callback = callback
+        self.max_items = max_items
+        self.max_latency = max_latency
 
     def __call__(self):
-        while True:
-            item = self.queue.get()
-            if item == STOP:
-                _LOGGER.debug('Exiting the QueueCallbackWorker.')
-                return
+        continue_ = True
+        while continue_:
+            items = _get_many(
+                self.queue,
+                max_items=self.max_items,
+                max_latency=self.max_latency)
+
+            # If stop is in the items, process all items up to STOP and then
+            # exit.
+            try:
+                items = items[:items.index(STOP)]
+                continue_ = False
+            except ValueError:
+                pass
 
             # Run the callback. If any exceptions occur, log them and
             # continue.
             try:
-                action, kwargs = item
-                self._callback(action, kwargs)
+                self._callback(items)
             except Exception as exc:
                 _LOGGER.error('%s: %s', exc.__class__.__name__, exc)
+
+        _LOGGER.debug('Exiting the QueueCallbackWorker.')
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import, division
 
 import abc
+import collections
 import logging
 import random
 import time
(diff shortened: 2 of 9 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -42,27 +75,43 @@ class QueueCallbackWorker(object):
...
+        self.max_latency = max_latency
 
     def __call__(self):
-        while True:
-            item = self.queue.get()
-            if item == STOP:
-                _LOGGER.debug('Exiting the QueueCallbackWorker.')
-                return
+        should_stop = False
+        while not should_stop:
```
chemelnucfin: `continue = True`
`while continue` ...etc.
reads better?

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -42,27 +75,43 @@ class QueueCallbackWorker(object):
...
+            items = _get_many(
+                self.queue,
+                max_items=self.max_items,
+                max_latency=self.max_latency)
+
+            # If stop is in the items, process all items up to STOP and then
+            # exit.
+            should_stop = False
+            try:
+                items = items[:items.index(STOP)]
```
chemelnucfin: if/else better in this case?

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
jonparrott: Definitely. Done.

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
jonparrott: Eh, it wouldn't be quite as tidy. You could do:

```
if STOP in items:
    cutoff = items.index(STOP)
    items = items[:items.index(STOP)
```

But that incurs two operations on the list (once to find if stop is there, another to find where stop is).

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
```diff
@@ -31,6 +34,36 @@
...
 
 
+def _get_many(queue_, max_items=None, max_latency=0):
+    """Get multiple items from a Queue.
+
+    Gets at least one (blocking) and at most ``max_items`` items
+    (non-blocking) from a given Queue. Does not mark the items as done.
+
+    Args:
+        queue_ (~queue.Queue`): The Queue to get items from.
```
lukesneeringer: Probable RST syntax error. :-)

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -31,6 +32,33 @@
...
 _LOGGER = logging.getLogger(__name__)
 
+# Namedtuples for management requests. Used by the Message class to communicate
+# items of work back to the policy.
+AckRequest = collections.namedtuple(
+    'AckRequest',
+    ['ack_id', 'byte_size', 'time_to_ack'],
+)
+
+DropRequest = collections.namedtuple(
```
lukesneeringer: One quibble I have with this is that `namedtuple` has a somewhat lousy equivalence method. In particular:

```python
DropRequest('ack123', 42) == LeaseRequest('ack123', 42)  # True
```

This is what data classes could solve if you did not need to support <= 3.5. ;-)

These are fine as-is, but be careful that on any tests you are always running `isinstance` checks. :-)

pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -273,21 +283,26 @@ def dispatch_callback(self, action, kwargs):
...
+            batched_commands[item.__class__].append(item)
+
+        _LOGGER.debug('Handling %d batched requests', len(items))
+
+        if batched_commands[base.LeaseRequest]:
+            self.lease(batched_commands.pop(base.LeaseRequest))
+        if batched_commands[base.ModAckRequest]:
+            self.modify_ack_deadline(
+                batched_commands.pop(base.ModAckRequest))
+        # Note: Drop and ack *must* be after lease. It's possible to get both
```
lukesneeringer: Pub/Sub is smart enough that if you ack and modack, it knows what to do (ignore the modack).
Since nacking is really just a modack of 0, though, order does matter (and may actually not be reliable, but that is not the end of the world).

So, leave as is.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
jonparrott: I'm kind of okay with this equivalence - the only time this actually matters is in `dispatch_callback` which does a `type` switch.

pubsub/google/cloud/pubsub_v1/types.py
```diff
@@ -53,13 +53,16 @@
 # The defaults should be fine for most use cases.
 FlowControl = collections.namedtuple(
     'FlowControl',
-    ['max_bytes', 'max_messages', 'resume_threshold', 'max_requests'],
+    ['max_bytes', 'max_messages', 'resume_threshold', 'max_requests',
```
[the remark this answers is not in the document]
jonparrott: Yeah. *Additive* changes are generally fine, but if it makes you more comfortable I can make them private until such a point that someone wants to turn the knobs.

pubsub/tests/unit/pubsub_v1/subscriber/test_message.py
```diff
@@ -53,53 +54,52 @@ def test_publish_time():
...
     with mock.patch.object(msg._request_queue, 'put') as put:
-        with mock.patch.object(message.Message, 'drop') as drop:
-            msg.ack()
-            put.assert_called_once_with(('ack', {
-                'ack_id': 'bogus_ack_id',
-                'byte_size': 25,
-                'time_to_ack': mock.ANY,
-            }))
+        msg.ack()
+        put.assert_called_once_with(base.AckRequest(
```
lukesneeringer: These tests are actually less precise now because of the generous `__eq__` implementation in `namedtuple`.

Recommend changing all of these to:

```python
_, args, _ = put.mock_calls[0]
assert isinstance(args[0], base.AckRequest)  # namedtuple equality is permissive.
assert args[0] == base.AckRequest(ack_id='bogus_ack_id', byte_size=25, time_to_ack=mock.ANY)
```

pubsub/google/cloud/pubsub_v1/subscriber/_helper_threads.py
chemelnucfin: I was thinking more of likely cases and branch prediction issues, but probably microoptimizing (whether STOP is more likely than not to happen)

pubsub/google/cloud/pubsub_v1/types.py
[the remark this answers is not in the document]
lukesneeringer: It is not so much that there are additive changes, it is that I feel like semi-advanced users should be expected to have to turn the first three knobs, and hopefully nobody ever looks at the others. Basically it just looks like an overwhelming number of options now.

Not sure there is a good way around that though. I would NOT want to see two objects being passed.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
lukesneeringer: So long as the tests check `isinstance` I am okay with it.

pubsub/tests/unit/pubsub_v1/subscriber/test_message.py
jonparrott: Done in 33947c285

PR #4907, opened 2018-02-21
Removes unnecessary debug print from test

alixhami:


jonparrott:
Merging despite CI, as it seems like it was just a case of a flaky backend.

PR #4909, opened 2018-02-22
Normalize all setup.py files

jonparrott:
This was done with this script: https://gist.github.com/jonparrott/34fa5e883ddff596a0a4529199022bb2

/cc @lukesneeringer the generator should be updated to output setup.py files in this format.

### DIFF
--- a/api_core/setup.py
+++ b/api_core/setup.py
@@ -1,4 +1,4 @@
-# Copyright 2018, Google LLC
+# Copyright 2018 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


api_core/setup.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google LLC
+# Copyright 2018, Google LLC
```
tswast: This differs slightly from the header in go/releasing/preparing (no comma there)

api_core/setup.py
```diff
@@ -12,68 +12,72 @@
...
-    'url': 'https://github.com/GoogleCloudPlatform/google-cloud-python',
-    'license': 'Apache 2.0',
-    'platforms': 'Posix; MacOS X; Windows',
-    'include_package_data': True,
-    'zip_safe': False,
-    'classifiers': [
-        'Development Status :: 4 - Beta',
+package_root = os.path.abspath(os.path.dirname(__file__))
+
+with io.open(os.path.join(package_root, 'README.rst')) as readme_file:
```
tswast: Do you want to explicitly pass in the encoding in case we ever decide to add some unicode to a README?

api_core/setup.py
```diff
@@ -12,68 +12,72 @@
...
-    version='0.1.5.dev1',
-    description='Core Google API Client Library',
-    long_description=README,
+    platforms='Posix; MacOS X; Windows',
+    zip_safe=False,
+    include_package_data=True,
+    long_description=readme,
+    install_requires=dependencies,
+    extras_require=extras,
+    packages=setuptools.find_packages(exclude=('tests*',)),
```
tswast: Might have to exclude "benchmark" folder too.https://github.com/GoogleCloudPlatform/google-cloud-python/tree/60d4460d13c2ce79ad085368f72cb93e97d7e6f9/bigquery/benchmark

PR #4916, opened 2018-02-22
Regenerate gapic for firestore, remove gax usage.

### LINKED ISSUE #4692: Drop usage of gax in favor of api_core

jonparrott:
Closes #4692, as it's the last API that uses gax.


jonparrott:
Do not merge until @lukesneeringer lets me know why the Admin client was removed.

chemelnucfin:
so what was the reasoning behind removing the admin client?  (low priority question)

jonparrott:
It wasn't ready. :)

PR #4931, opened 2018-02-26
Bigquery: adds `page_size` parameter for `list_rows` and use in DB-API for `arraysize`

### LINKED ISSUE #4260: BigQuery: add page_size parameter for list_rows and use in DB-API for arraysize
With https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4144 the arraysize parameter no longer affects how much data the library actually requests from the API.

We should add a `page_size` parameter to `Client.list_rows` and use that parameter in the DB-API implementation.

See also: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4134

chemelnucfin:
Closes #4260?

@tswast I'm not sure if there's more that needs to be done here.  But here's a stab at it.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1289,7 +1289,8 @@ class Client(ClientWithProject):
                            row to read.
 
         :type page_size: int
-        :param page_size: (Optional) The maximum number of items to return.
+        :param page_size: (Optional) The maximum number of items to return
+                          per page in the iterator.  Defaults to 50.
 
         :type retry: :class:`google.api_core.retry.Retry`
         :param retry: (Optional) How to retry the RPC.
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1247,7 +1247,8 @@ class Client(ClientWithProject):
         return self.insert_rows_json(*args, **kwargs)
 
     def list_rows(self, table, selected_fields=None, max_results=None,
-                  page_token=None, start_index=None, retry=DEFAULT_RETRY):
+                  page_token=None, start_index=None, page_size=None,
+                  retry=DEFAULT_RETRY):
         """List the rows of the table.
 
         See
@@ -1287,6 +1288,9 @@ class Client(ClientWithProject):
         :param start_index: (Optional) The zero-based index of the starting
                            row to read.
 
+        :type page_size: int
+        :param page_size: (Optional) The maximum number of items to return.
+
         :type retry: :class:`google.api_core.retry.Retry`
         :param retry: (Optional) How to retry the RPC.
 
@@ -1314,9 +1318,10 @@ class Client(ClientWithProject):
         if selected_fields is not None:
             params['selectedFields'] = ','.join(
                 field.name for field in selected_fields)
-
         if start_index is not None:
             params['startIndex'] = start_index
+        if page_size is not None:
+            params['pageSize'] = page_size
 
         row_iterator = RowIterator(
             client=self,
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1256,6 +1257,9 @@ def list_rows(self, table, selected_fields=None, max_results=None,
         :param start_index: (Optional) The zero-based index of the starting
                            row to read.
 
+        :type page_size: int
+        :param page_size: (Optional) The maximum number of items to return.
```
tswast: This docstring is very misleading. Unless `page_size` is just an alias for `max_results`?

I would hope that in setting the page size, I could still fetch all pages via the iterator.

chemelnucfin: From this, [paging-results](https://cloud.google.com/bigquery/docs/paging-results)

it looks like the default value is 50.  Please let me know if that's not the case.

chemelnucfin:
@tswast Are there any guidelines on when we do system tests?

I have had some that wanted system tests, I have had some that didn't.

tswast:
The reason I asked for a system test here is that I know there is some tricky API request rewriting that happens behind the scenes, so I'd like to see something end-to-end that verifies that "yes, this does what I expect."

In general, it usually doesn't hurt to add a system test. The exception to that is if it's a very flakey test or takes a very long time to run.

bigquery/tests/system.py
```diff
@@ -1609,6 +1609,31 @@ def test_nested_table_to_dataframe(self):
...
+
+        SF = bigquery.SchemaField
+        schema = [SF('string_col', 'STRING', mode='NULLABLE')]
+        to_insert = [{'string_col': 'item%d' % i} for i in range(num_items)]
+        table_id = 'test_table'
+        dataset = self.temp_dataset(_make_dataset_id('nested_df'))
+        table_arg = Table(dataset.table(table_id), schema=schema)
+        table = Config.CLIENT.create_table(table_arg)
+        self.to_delete.insert(0, table)
+        Config.CLIENT.insert_rows(table, to_insert, selected_fields=schema)
```
tswast: Please don't use `insert_rows()` in this system test. This makes this test very flakey. Instead it's recommended to use `load_table_from_file()` (search for `BytesIO` in this file for an example)

Why? With `insert_rows()`, rows get inserted into the streaming buffer and only asynchronously get added to the real table. `list_rows()` does not return rows from the streaming buffer.

bigquery/tests/system.py
```diff
@@ -1609,6 +1609,42 @@ def test_nested_table_to_dataframe(self):
...
+        
+        num_items = 7
+        page_size = 3
+        num_pages, num_last_page = divmod(num_items, page_size)
+
+        SF = bigquery.SchemaField
+        schema = [SF('string_col', 'STRING', mode='NULLABLE')]
+        to_insert = [{'string_col': 'item%d' % i} for i in range(num_items)]
+        rows = [json.dumps(row) for row in to_insert]
+        body = six.StringIO('{}\n'.format('\n'.join(rows)))
```
tswast: Please use BytesIO instead. StringIO will fail if given non-ASCII characters, so I want to make sure we are showing a good example in case someone stumbles onto this.

chemelnucfin: should the other test be changed also?

tswast: Yes, that would be helpful. If you feel like doing that, let's do it as a separate clean-up PR.

chemelnucfin: might as well since I'm here already.

chemelnucfin:
@dpebot merge

chemelnucfin:
@dpebot green

chemelnucfin:
@dpebot merge green

chemelnucfin:
@dpebot merge green

PR #4936, opened 2018-02-26
Install local dependencies when running lint

jonparrott:


jonparrott:
Nox-only change, merging.

PR #4937, opened 2018-02-26
Don't run tests for every package when api_core is modified

jonparrott:
This is in preparation for both a repository split and for api_core v1.0.0.


jonparrott:
CI-only change, merging.

PR #4938, opened 2018-02-26
Add batch query support

tseaver:
/cc @vkedia

### DIFF
--- a/spanner/google/cloud/spanner_v1/database.py
+++ b/spanner/google/cloud/spanner_v1/database.py
@@ -27,6 +27,7 @@ from google.cloud.spanner_v1 import __version__
 from google.cloud.spanner_v1._helpers import _metadata_with_prefix
 from google.cloud.spanner_v1.batch import Batch
 from google.cloud.spanner_v1.gapic.spanner_client import SpannerClient
+from google.cloud.spanner_v1.keyset import KeySet
 from google.cloud.spanner_v1.pool import BurstyPool
 from google.cloud.spanner_v1.pool import SessionCheckout
 from google.cloud.spanner_v1.session import Session
@@ -308,6 +309,14 @@ class Database(object):
         """
         return BatchCheckout(self)
 
+    def batch_transaction(self):
+        """Return an object which wraps a batch read / query.
+
+        :rtype: :class:`~google.cloud.spanner_v1.database.BatchTransaction`
+        :returns: new wrapper
+        """
+        return BatchTransaction(self)
+
     def run_in_transaction(self, func, *args, **kw):
         """Perform a unit of work in a transaction, retrying on abort.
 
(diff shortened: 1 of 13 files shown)


tseaver:
@dpebot merge on green please

vkedia:
cc @tseaver 
Sorry I have been busy last few days. I will take a look this week.
Can you please me as a reviewer. I want to make sure this is merged after my review.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -308,6 +309,37 @@ def batch(self):
         """
         return BatchCheckout(self)
 
+    def batch_transaction(
+            self, read_timestamp=None, min_read_timestamp=None,
+            max_staleness=None, exact_staleness=None):
```
vkedia: min_read_timestamp and max_staleness are not supported for batch transaction.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -406,6 +438,258 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        :returns: a result set instance which can be used to consume rows.
+        :raises ValueError: if batch does not contain either 'read' or 'query'
+        """
+        if 'query' in batch:
+            return self.process_query_batch(batch)
+        if 'read' in batch:
+            return self.process_read_batch(batch)
+        raise ValueError("Invalid batch")
+
+    def close(self):
```
vkedia: can you add documentation here mentioning something like:
```
If BatchTransaction has been shared across multiple machines, calling this on any machine would invalidate the transaction everywhere. Ideally this would be called when data has been read from all the partitions.
```

We do not want users to call this on every worker.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -406,6 +438,258 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self._database._pool.put(self._session)
 
 
+class BatchTransaction(object):
```
vkedia: Users should be able to do normal read/queries also on BatchTransaction. Can they do that?

spanner/google/cloud/spanner_v1/snapshot.py
```diff
@@ -194,6 +209,141 @@ def execute_sql(self, sql, params=None, param_types=None, query_mode=None):
         else:
             return StreamedResultSet(iterator)
 
+    def partition_read(self, table, columns, keyset, index='',
```
vkedia: why do we need to make these public here as well? If users need to call these methods they should be using BatchTransaction.

spanner/google/cloud/spanner_v1/database.py
tseaver: Removed: e426c59

spanner/google/cloud/spanner_v1/database.py
tseaver: 487dc53

spanner/google/cloud/spanner_v1/database.py
tseaver: Added in 5342ccd

spanner/google/cloud/spanner_v1/snapshot.py
tseaver: They are lower-level, but are API wrappers just like all the other snapshot methods.

vkedia: But we do not want people to be invoking these method directly via a snapshot. They should create a BatchTransaction if they want to invoke these methods. Also these methods are a very niche use case. Why enhance the Snapshot API with methods that hardly anyone will use?

tseaver: As with the `Session` class, or explicit transaction management, we don't expect "normal" users to interact directly with these methods:  we document the recommended patterns instead.

Leaving these methods public, like the other API method wrappers, enables users who have use cases we don't anticpate (e.g. perhaps the kind of fan-out/fan-in I originally sketched) to design new solutions using them as primitives.

tseaver:
For one thing, the contract is "narrower" -- as you had me fix in e426c59, `BatchTransaction` has more stringent requirements than `Snapshot` for the staleness parameters.

In general, we prefer composition over inheritance for this kind of problem space.  Python makes writing the "delegating" methods easy, and we don't rely on `IsA` semantics anywhere.

spanner/google/cloud/spanner_v1/snapshot.py
vkedia: We can always make these public if needed later based on customer feedback. I think in this specific case, making these methods public might cause more people to directly use these rather than batch transaction since they would be familiar with the Snapshot interface and using it everywhere else.

vkedia:
I am not sure the contract is narrower just because while constructing the object, you are giving a narrower set of parameters. Once constructed, a batch transaction supports everything that a normal snapshot does and some more. But I am ok with using composition if that is more idiomatic in Python.

vkedia:
Actually BatchTransaction and the related methods should be called `BatchSnapshot` .

tseaver:
> `BatchTransaction` and the related methods should be called `BatchSnapshot` 

@vkedia 26a21ce renames the class and the associated factory.

spanner/google/cloud/spanner_v1/snapshot.py
jonparrott: I am generally in favor of making things private until we need them to be public.

tseaver: Done in dbabfba

tseaver:
Test failure was in BigQuery system tests (why are we running them for this PR again?)

tseaver:
@jonparrott Do you have a notion why the BigQuery system tests are even being run here?  I just opened #5003 to track the flakiness.

spanner/google/cloud/spanner_v1/snapshot.py
```diff
@@ -105,7 +105,7 @@ def read(self, table, columns, keyset, index='', limit=0, partition=None):
 
         :type partition: bytes
         :param partition: (Optional) one of the partition tokens returned
-                          from :meth:`partition_read`.  Incompatible with
+                          from :meth:`_partition_read`.  Incompatible with
```
vkedia: I think we should remove this parameter from this method and instead create a private overloaded method that takes this parameter as well.

jonparrott:
> Do you have a notion why the BigQuery system tests are even being run here? I just opened #5003 to track the flakiness.

Probably because it computes the changed packages based on both the changes in this branch and the changes in master. I suspect that if you re-base it'll stop doing that.

tseaver:
@jonparrott That seems broken to me.  I rebased just now.

spanner/google/cloud/spanner_v1/snapshot.py
jonparrott: Python doesn't have the concept of "overloads". 

Do we expect users to call `read` with `partition` or are we expecting just our helpers to do so? If the former, we should leave as-is, if the later, we can definitely move this to a private `_read_from_partition` (which is unfortunately named because we also have `_partition_read`.

vkedia: We expect users to use the methods in BatchSnapshot to read/query from partitions and not directly invoke the method in Snapshot. Also note that the way this is documented right now, it tells users that to pass this parameter, they need to invoke a private method `_partition_read`.
I am ok with whatever name you chose for the private method.

jonparrott: Sounds good. There should be preferably one obvious way to do things. @tseaver, can you handle this?

tseaver: @jonparrott  I would rather back out fad5366:  ISTM that the `Snapshot.partition_{read,query}` API wrapper methods are exactly like `Session.{create,delete}`:  they are lower-level methods which most users won't call directly, but we have no reason to prevent them from doing so.

jonparrott: Okay, so help me understand here. We should have one (preferably obvious) way to accomplish starting a partition read/query and then reading data from it.

Can we pick whichever way that is and make everything else either private or remove it? What's stopping us from doing this?

vkedia: I think there is a good reason to prevent people from directly using the snapshot interface for partitioned reads. If they use the snapshot interface how will they make sure that they are using the same session id and transaction id across machines?

tseaver: Here is the "obvious" / documented / expected way:

```python
        batch_txn = self._db.batch_snapshot(read_timestamp=committed)
        for batch in batch_txn.generate_query_batches(sql):
            do_something_with_rows( batch_txn.process(batch))  # might  be on another host
        batch_txn.close()
```

`Database.generate_query_batches` is a convenience wrapper:  it calls the underling `Snapshot.partition_query` method and returns results suitable for serialization / processing by `Database.process()`.

Making the lower level `partition_query` private is a bit BDSM for Python:  it is the wrapper we have for the published `PartitionQueryRequest` API.  In this case, it is like the `Session.create` / `Session.delete` API wrappers:  they are low-level API wrappers used by the higher-level database convenience methods.

jonparrott: Gotcha. Your position makes sense to me. Let's do that and get this merged.

tseaver: Rebased with fad5366 backed out.

vkedia:
@tseaver Please don't merge yet. I want to take another look.

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -406,6 +426,267 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        .. note::
+
+           Caller is responsible for cleaning up the session after
+           all partitions have been processed.
+        """
+        if self._session is None:
+            session = self._session = self._database.session()
+            session.create()
+            txn = session.transaction()
```
vkedia: won't this start a read write transaction on this session?

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -406,6 +426,267 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        Result can be used to serialize the instance and reconstitute
+        it later using :meth:`from_dict`.
+
+        :rtype: dict
+        """
+        session = self._get_session()
+        return {
+            'session_id': session._session_id,
+            'transaction_id': session._transaction._transaction_id,
```
vkedia: is this the transaction id of the snapshot?

spanner/google/cloud/spanner_v1/database.py
```diff
@@ -406,6 +426,267 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        :param mapping: serialized state of the instance
+
+        :rtype: :class:`BatchSnapshot`
+        """
+        instance = cls(database)
+        session = instance._session = database.session()
+        session._session_id = mapping['session_id']
+        txn = session.transaction()
+        txn._transaction_id = mapping['transaction_id']
+        return instance
```
vkedia: you are not initializing snapshot here? That means that on every worker a new snapshot will be created with a different id?

spanner/google/cloud/spanner_v1/database.py
tseaver: @vkedia you are correct.  The session belonging to the `BatchSnapshot` instance should not have a `_transaction` created.

spanner/google/cloud/spanner_v1/database.py
tseaver: 36c4857

spanner/google/cloud/spanner_v1/database.py
tseaver: 36c4857

spanner/google/cloud/spanner_v1/database.py
tseaver: 36c4857

tseaver:
@vkedia PTAL.

tseaver:
Somthing is triggering the BigQuery system tests again (which is broken!), and they are timing out our CircleCI job.  I just rebased here to try to work around that problem.

spanner/google/cloud/spanner_v1/database.py
tseaver: @vkedia ab34ccb removes those two stray lines.

spanner/google/cloud/spanner_v1/snapshot.py
```diff
@@ -99,10 +100,17 @@ def read(self, table, columns, keyset, index='', limit=0):
...
 
         :type limit: int
-        :param limit: (Optional) maximum number of rows to return
+        :param limit: (Optional) maximum number of rows to return.
+                      Incompatible with ``partition``.
+
+        :type partition: bytes
+        :param partition: (Optional) one of the partition tokens returned
+                          from :meth:`partition_read`.  Incompatible with
+                          ``limit``.
```
chemelnucfin: Should ValueError be raised if both are specified?

tseaver: The server enforces that constraint.

PR #4944, opened 2018-02-27
Update dependency range for api-core to include v1.0.0 releases

jonparrott:


jonparrott:
Merging, local tests pass and all of these tests will be run during the release the world. We can catch any issues there.

PR #4979, opened 2018-02-28
Move TimestampNanoSeconds to api_core

### LINKED ISSUE #4829: Promote 'TimestampWithNanos' from 'spanner' to 'api_core'
It would be useful for other services as they begin to render nanosecond-resolution timestamps.

See: #4807

chemelnucfin:
Closes #4829

### DIFF
--- a/api_core/google/api_core/datetime_helpers.py
+++ b/api_core/google/api_core/datetime_helpers.py
@@ -181,7 +181,7 @@ def to_rfc3339(value, ignore_zone=True):
     return value.strftime(_RFC3339_MICROS)
 
 
-class TimestampWithNanoseconds(datetime.datetime):
+class DatetimeWithNanoseconds(datetime.datetime):
     """Track nanosecond in addition to normal datetime attrs.
 
     Nanosecond can be passed only as a keyword argument.
@@ -225,7 +225,7 @@ class TimestampWithNanoseconds(datetime.datetime):
             stamp (str): RFC 3339 stamp, with up to nanosecond precision
 
         Returns:
-            :class:`TimestampWithNanoseconds`:
+            :class:`DatetimeWithNanoseconds`:
                 an instance matching the timestamp string
 
         Raises:
--- a/api_core/google/api_core/datetime_helpers.py
+++ b/api_core/google/api_core/datetime_helpers.py
@@ -179,3 +179,71 @@ def to_rfc3339(value, ignore_zone=True):
         value = value.replace(tzinfo=None) - value.utcoffset()
 
     return value.strftime(_RFC3339_MICROS)
+
+
+class TimestampWithNanoseconds(datetime.datetime):
+    """Track nanosecond in addition to normal datetime attrs.
+
+    Nanosecond can be passed only as a keyword argument.
+    """
+    __slots__ = ('_nanosecond',)
+
+    # pylint: disable=arguments-differ
+    def __new__(cls, *args, **kw):
+        nanos = kw.pop('nanosecond', 0)
+        if nanos > 0:
+            if 'microsecond' in kw:
+                raise TypeError(
+                    "Specify only one of 'microsecond' or 'nanosecond'")
+            kw['microsecond'] = nanos // 1000
+        inst = datetime.datetime.__new__(cls, *args, **kw)
+        inst._nanosecond = nanos or 0
+        return inst
+    # pylint: disable=arguments-differ
+
+    @property
+    def nanosecond(self):
+        """Read-only: nanosecond precision."""
+        return self._nanosecond
+
+    def rfc3339(self):
+        """Return an RFC 3339-compliant timestamp.
+
+        Returns:
+            (str): Timestamp string according to RFC 3339 spec.
+        """
+        if self._nanosecond == 0:
+            return to_rfc3339(self)
+        nanos = str(self._nanosecond).rstrip('0')
+        return '{}.{}Z'.format(self.strftime(_RFC3339_NO_FRACTION), nanos)
+
+    @classmethod
+    def from_rfc3339(cls, stamp):
+        """Parse RFC 3339-compliant timestamp, preserving nanoseconds.
+
+        Args:
+            stamp (str): RFC 3339 stamp, with up to nanosecond precision
+
+        Returns:
+            :class:`TimestampWithNanoseconds`:
+                an instance matching the timestamp string
+
+        Raises:
+            ValueError: if `stamp` does not match the expected format
+        """
+        with_nanos = _RFC3339_NANOS.match(stamp)
+        if with_nanos is None:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+            nanos = 0
+        else:
+            scale = 9 - len(fraction)
+            nanos = int(fraction) * (10 ** scale)
+        return cls(bare.year, bare.month, bare.day,
+                   bare.hour, bare.minute, bare.second,
+                   nanosecond=nanos, tzinfo=UTC)
+
+
+class _UTC(datetime.tzinfo):
```
jonparrott: We depend on `pytz`, so you can remove this.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+    nanosecond can be passed only as a keyword argument.
+    """
+    __slots__ = ('_nanosecond',)
+
+    # pylint: disable=arguments-differ
+    def __new__(cls, *args, **kw):
+        nanos = kw.pop('nanosecond', 0)
+        if nanos > 0:
+            if 'microsecond' in kw:
+                raise TypeError(
```
jonparrott: This is a `ValueError`, not a `TypeError`.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+
+class TimestampWithNanoseconds(datetime.datetime):
+    """Track nanosecond in addition to normal datetime attrs.
+
+    nanosecond can be passed only as a keyword argument.
+    """
+    __slots__ = ('_nanosecond',)
+
+    # pylint: disable=arguments-differ
+    def __new__(cls, *args, **kw):
```
jonparrott: Why are we overriding `__new__` instead of `__init__`? @tseaver?

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+        return inst
+    # pylint: disable=arguments-differ
+
+    @property
+    def nanosecond(self):
+        """Read-only: nanosecond precision."""
+        return self._nanosecond
+
+    def rfc3339(self):
+        """RFC 3339-compliant timestamp.
```
jonparrott: `Return an RFC 3339 string representing this datetime.`

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+
+        :rtype: str
```
jonparrott: please use google-style docstrings throughout.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,109 @@ def to_rfc3339(value, ignore_zone=True):
...
+        :returns: Timestamp string according to RFC 3339 spec.
+        """
+        if self._nanosecond == 0:
+            return to_rfc3339(self)
+        nanos = str(self._nanosecond).rstrip('0')
+        return '%s.%sZ' % (self.strftime(_RFC3339_NO_FRACTION), nanos)
```
jonparrott: please use format throughout.

api_core/tests/unit/test_datetime_helpers.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import unittest
```
jonparrott: please don't use unit test. Convert the tests to pytest style.

api_core/google/api_core/datetime_helpers.py
chemelnucfin: @tseaver might have meant that it is a TypeError to use this class in that case?  I'll wait for his answer.

tseaver: Grey area, actually:  the "allowed arguments" to a constructor are actually part of the objects "type," in a formal sense:  there is nothing wrong with the *value* being passed, it is just invalid to pass both `microsecond` and `nanosecond`.

api_core/google/api_core/datetime_helpers.py
tseaver: @jonparrott Try deriving from `datetime.datetime` and using `__init__`:  the majyk happens inside `__new__`, so we have to override it:

```python
# override_datetime___init__.py
import datetime

class MyDateTime(datetime.datetime):

    __slots__ = ('foo',)

    def __init__(self, *args, **kw):
        foo = kw.pop('foo', None)
        super(MyDateTime, self).__init__(*args, **kw)
        self.foo = foo

mdt = MyDateTime(2018, 2, 28, 23, 11, 57, foo='bar')
```

```bash
$ python override_datetime___init__.py
  File "override_datetime___init__.py", line 13, in <module>
    mdt = MyDateTime(2018, 2, 28, 23, 11, 57, foo='bar')
TypeError: 'foo' is an invalid keyword argument for this function
```

api_core/tests/unit/test_datetime_helpers.py
```diff
@@ -148,3 +148,105 @@ def test_to_rfc3339_with_non_utc_ignore_zone():
...
+
+def test_timestampwithnanos_from_rfc3339_w_invalid():
+    klass = datetime_helpers.TimestampWithNanoseconds
+    STAMP = '2016-12-20T21:13:47'
+    with pytest.raises(ValueError):
+        klass.from_rfc3339(STAMP)
+
+
+def test_timestampwithnanos_from_rfc3339_wo_fraction():
+    from google.api_core import datetime_helpers
```
jonparrott: remove these function-level imports and just import at the top level.

api_core/tests/unit/test_datetime_helpers.py
```diff
@@ -148,3 +148,105 @@ def test_to_rfc3339_with_non_utc_ignore_zone():
...
+
+    klass = datetime_helpers.TimestampWithNanoseconds
```
jonparrott: No need for this alias.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -179,3 +179,71 @@ def to_rfc3339(value, ignore_zone=True):
         value = value.replace(tzinfo=None) - value.utcoffset()
 
     return value.strftime(_RFC3339_MICROS)
+
+
+class TimestampWithNanoseconds(datetime.datetime):
```
jonparrott: Wait - this should be `DatetimeWithNanoseconds`.

PR #4980, opened 2018-03-01
Remove Timestamp from Spanner

chemelnucfin:
~~Diffbase #4979~~

### DIFF
--- a/spanner/google/cloud/spanner_v1/_helpers.py
+++ b/spanner/google/cloud/spanner_v1/_helpers.py
@@ -19,80 +19,12 @@ import math
 
 import six
 
-from google.protobuf.struct_pb2 import ListValue
-from google.protobuf.struct_pb2 import Value
-from google.cloud.spanner_v1.proto import type_pb2
-
+from google.api_core import datetime_helpers
 from google.cloud._helpers import _date_from_iso8601_date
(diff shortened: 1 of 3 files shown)


jonparrott:
@tseaver can you review this? (I don't have the bandwidth)

spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -19,80 +19,12 @@
...
-            nanos = 0
-        else:
-            scale = 9 - len(fraction)
-            nanos = int(fraction) * (10 ** scale)
-        return cls(bare.year, bare.month, bare.day,
-                   bare.hour, bare.minute, bare.second,
-                   nanosecond=nanos, tzinfo=UTC)
+from google.cloud.spanner_v1.proto import type_pb2
+from google.protobuf.struct_pb2 import ListValue
+from google.protobuf.struct_pb2 import Value
```
tseaver: Per [PEP 8](https://www.python.org/dev/peps/pep-0008/#id23), non-local imports (e.g., `google.protobuf`) are supposed to be sorted between standard-library imports and local imports.

jonparrott:
Please note that now we need to cut a release of api_core before we can release any new versions of spanner.

PR #4981, opened 2018-03-01
Remove Timestamp from Core

chemelnucfin:
~~Diffbase #4980~~
Merged.


jonparrott:
@chemelnucfin I don't want to touch `core` at all until we deprecate it.

chemelnucfin:
ok.  closing for now.

PR #4991, opened 2018-03-06
futures now install_requires

david-gang:
According to this [comment ]( https://github.com/pypa/pipenv/issues/1547#issuecomment-370778759) the current extras_require format to specify platform dependencies is deprecated, so moving it upwards
Also look here at the [documentation](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies) and at [SO](https://stackoverflow.com/questions/21082091/install-requires-based-on-python-version)

### DIFF
--- a/api_core/setup.py
+++ b/api_core/setup.py
@@ -36,10 +36,10 @@ dependencies = [
     'setuptools>=34.0.0',
     'six>=1.10.0',
     'pytz',
+    'futures>=3.2.0;python_version<"3.2"'
 ]
 extras = {
-    'grpc': 'grpcio>=1.8.2',
-    ':python_version < "3.2"': 'futures>=3.2.0',
+    'grpc': 'grpcio>=1.8.2'
 }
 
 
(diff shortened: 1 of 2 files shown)


jonparrott:
Interesting. I'll need to do some testing to make sure this doesn't break in some older toolchains, but thank you for bringing this to our attention!

david-gang:
Thanks. 
As i can see in the code you require a recent version of setuptools anyway:
https://github.com/david-gang/google-cloud-python/blob/fa7b4ed8e099bf5ed0606e6f9196b0c0cb07f94f/api_core/setup.py#L36

uranusjr:
Here’s how Pytest makes sure about the environment marker support level, for the most paranoid: https://github.com/pytest-dev/pytest/blob/master/setup.py#L68

In general most packages shouldn’t worry about that unless with good reasons, however. Keeping unnecessary support to the old syntax gives users incentive to use outdated versions of Setuptools, which has security implications.

lukesneeringer:
@jonparrott Any objection to my merging this?

jonparrott:
@lukesneeringer I wanted to give this a test with some older pips.

jonparrott:
@david-gang thanks for doing this!

PR #4998, opened 2018-03-07
Add vision v1p2beta1

lukesneeringer:
Self-explanatory.

I noticed that we lack docs for v1p1beta1, which may have been just an oversight on my part. v1p2beta1 are also excluded for the moment, but I can add them and make them work like video/speech/etc.


jonparrott:
I've been making sure to log all commands used to generate the libraries (see [here](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4877/commits/619cabf19a82b2d094d7d34a1590f9df1f1501c2)) can you do similar here?

lukesneeringer:
> I've been making sure to log all commands used to generate the libraries (see here) can you do similar here?

Sorry, I just noticed this. I will do it in the future.

PR #4999, opened 2018-03-07
BigQuery: Updates table snippets

alixhami:
Updates existing snippets and adds new snippets for inclusion in [/bigquery/docs/tables](https://cloud.google.com/bigquery/docs/tables)

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -39,8 +39,8 @@ UPDATED_FRIENDLY_NAME = 'Updated friendly name'
 UPDATED_DESCRIPTION = 'Updated description'
 
 SCHEMA = [
-    bigquery.SchemaField('full_name', 'STRING', mode='required'),
-    bigquery.SchemaField('age', 'INTEGER', mode='required'),
+    bigquery.SchemaField('full_name', 'STRING', mode='REQUIRED'),
+    bigquery.SchemaField('age', 'INTEGER', mode='REQUIRED'),
 ]
 
 ROWS = [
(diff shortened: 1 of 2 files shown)


tswast:
Circle is failing due to missing region tag reference in the usage docs.

docs/bigquery/snippets.py
```diff
@@ -255,41 +255,66 @@ def test_list_tables(client, to_delete):
...
+    # [START bigquery_list_tables]
     tables = list(client.list_tables(dataset))  # API request(s)
     assert len(tables) == 0
 
     table_ref = dataset.table('my_table')
     table = bigquery.Table(table_ref)
-    table.view_query = QUERY
-    client.create_table(table)                          # API request
+    client.create_table(table)                  # API request
     tables = list(client.list_tables(dataset))  # API request(s)
```
tswast: We should double-check with the BigQuery team to see if table listing is strong consistent or eventually consistent. If eventually consistent, this test / sample could be flakey.

docs/bigquery/snippets.py
```diff
@@ -338,6 +363,40 @@ def test_get_table(client, to_delete):
...
+    dataset = bigquery.Dataset(dataset_ref)
+    client.create_dataset(dataset)
+    to_delete.append(dataset)
+
+    table = bigquery.Table(dataset.table(TABLE_ID), schema=SCHEMA)
+    table.description = ORIGINAL_DESCRIPTION
+    table = client.create_table(table)
+    to_delete.insert(0, table)
+
+    # [START bigquery_show_table]
```
tswast: Show table? Which section in the docs does this correspond to? I'd think that `bigquery_get_table` might better match the convention used for other region tags.

If you think that conflicts too much with the `get_table` above, I think that one this is more like `bigquery_get_table_reload` since it refreshes a table they've already fetched, but I can't imagine where we'd use in the docs (other than in the migration guide for the folks used to "reload")

docs/bigquery/snippets.py
```diff
@@ -338,6 +363,40 @@ def test_get_table(client, to_delete):
     # [END get_table]
 
 
+def test_show_table(client, to_delete):
+    """Show a table's properties"""
+    DATASET_ID = 'show_table_dataset_{}'.format(_millis())
```
tswast: Nit: These probably shouldn't be all caps, since they aren't global constants.

docs/bigquery/snippets.py
alixhami: It's for this [section](https://cloud.google.com/bigquery/docs/tables#getting_table_information).  I can update since adding `bigquery_` to the front will distinguish it. It looks like sample labeled `get_table` is just in the migration guide and `usage.rst`.

PR #5007, opened 2018-03-09
Firestore: return non existent snapshot if document not found instead of raising NotFound exception

### LINKED ISSUE #4809: Firestore: Non-existing Documents
Calls to `DocumentSnapshot.to_dict()` for non-existing documents should return `None` and not an empty dictionary.

chemelnucfin:
closes #4809

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/client.py
+++ b/firestore/google/cloud/firestore_v1beta1/client.py
@@ -582,8 +582,7 @@ def _parse_batch_get(get_doc_response, reference_map, client):
             a document factory.
 
     Returns:
-        Optional[.DocumentSnapshot]: The retrieved snapshot. If the
-        snapshot is :data:`None`, that means the document is ``missing``.
+        Optional[.DocumentSnapshot]: The retrieved snapshot.
 
     Raises:
         ValueError: If the response has a ``result`` field (a oneof) other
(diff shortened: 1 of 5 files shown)


firestore/google/cloud/firestore_v1beta1/document.py
```diff
@@ -420,17 +419,11 @@ def get(self, field_paths=None, transaction=None):
         Returns:
             ~.firestore_v1beta1.document.DocumentSnapshot: A snapshot of
             the current document.
```
tseaver: Needs a note about the state of  a missing document (empty `data`, `create_time` and `update_time` are `None`).

firestore/google/cloud/firestore_v1beta1/client.py
```diff
@@ -582,8 +582,7 @@ def _parse_batch_get(get_doc_response, reference_map, client):
             a document factory.
 
     Returns:
-        Optional[.DocumentSnapshot]: The retrieved snapshot. If the
-        snapshot is :data:`None`, that means the document is ``missing``.
+        Optional[.DocumentSnapshot]: The retrieved snapshot.
```
tseaver: I don't know what `Optional` is supposed to mean for `Returns:`, but if it meant that the method might return `None`, that is no longer treu.

firestore/google/cloud/firestore_v1beta1/document.py
```diff
@@ -420,17 +419,11 @@ def get(self, field_paths=None, transaction=None):
...
-
-        Raises:
-            ~google.cloud.exceptions.NotFound: If the document does not exist.
         """
         snapshot_generator = self._client.get_all(
             [self], field_paths=field_paths, transaction=transaction)
         snapshot = _consume_single_get(snapshot_generator)
```
tseaver: Just return the result of `_consume_single_get` -- we don't need a local copy.

firestore/tests/unit/test_document.py
```diff
@@ -445,10 +445,12 @@ def test_get_with_transaction(self):
             [document], field_paths=None, transaction=transaction)
 
     def test_get_not_found(self):
-        from google.cloud.exceptions import NotFound
+        from google.cloud.firestore_v1beta1.document import DocumentSnapshot
 
         # Create a minimal fake client with a dummy response.
-        response_iterator = iter([None])
+        read_time = 123
+        snapshot = DocumentSnapshot(None, None, False, read_time, None, None)
```
tseaver: Use a different variable name than the one below, e.g. `expected`.

chemelnucfin:
@tseaver addressed your comments.  thanks.  I kept a lot of the unittest to not compare to `expected` since that was the expected behavior.

PR #5009, opened 2018-03-09
BigQuery: Updates table management snippets

alixhami:
Updates table samples for cloud.google.com site How-to Guides sections "Working with Tables" and "Working with Table Schemas"

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -1145,7 +1145,7 @@ def test_extract_table_compressed(client, to_delete):
     # [START bigquery_extract_table_compressed]
     # client = bigquery.Client()
     # bucket_name = 'my-bucket'
-    destination_uri = 'gs://{}/{}'.format(bucket_name, 'shakespeare.gz')
+    destination_uri = 'gs://{}/{}'.format(bucket_name, 'shakespeare.csv.gz')
     dataset_ref = client.dataset('samples', project='bigquery-public-data')
     table_ref = dataset_ref.table('shakespeare')
     job_config = bigquery.job.ExtractJobConfig()
@@ -1156,7 +1156,7 @@ def test_extract_table_compressed(client, to_delete):
     extract_job.result()  # Waits for job to complete.
     # [END bigquery_extract_table_compressed]
 
-    blob = bucket.get_blob('shakespeare.gz')
+    blob = bucket.get_blob('shakespeare.csv.gz')
     assert blob.exists
     assert blob.size > 0
     to_delete.insert(0, blob)
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -23,7 +23,6 @@ a ``to_delete`` list;  the function adds to the list any objects created which
 need to be deleted during teardown.
 """
 
-import json
 import time
 
 import pytest
@@ -300,7 +299,7 @@ def test_create_table(client, to_delete):
     to_delete.insert(0, table)
 
 
-def test_create_table_without_schema(client, to_delete):
+def test_create_table_then_add_schema(client, to_delete):
     """Create a table without specifying a schema"""
     dataset_id = 'create_table_without_schema_dataset_{}'.format(_millis())
     dataset_ref = client.dataset(dataset_id)
@@ -320,6 +319,24 @@ def test_create_table_without_schema(client, to_delete):
     # [END bigquery_create_table_without_schema]
 
     to_delete.insert(0, table)
+    table_id = 'my_table'
+
+    # [START bigquery_add_schema_to_empty]
+    # client = bigquery.Client()
+    # dataset_id = 'my_dataset'
+    # table_id = 'my_table'
+
+    table_ref = client.dataset(dataset_id).table(table_id)
+    schema = [
+        bigquery.SchemaField('full_name', 'STRING', mode='REQUIRED'),
+        bigquery.SchemaField('age', 'INTEGER', mode='REQUIRED'),
+    ]
+    table = bigquery.Table(table_ref, schema=schema)
+
+    table = client.update_table(table, ['schema'])  # API request
+
+    assert table.schema == schema
+    # [START bigquery_add_schema_to_empty]
 
 
 def test_create_table_cmek(client, to_delete):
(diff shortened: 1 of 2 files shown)


docs/bigquery/snippets.py
```diff
@@ -320,6 +319,24 @@ def test_create_table_without_schema(client, to_delete):
     # [END bigquery_create_table_without_schema]
 
     to_delete.insert(0, table)
+    table_id = 'my_table'
```
tseaver: ISTM it would be better to move this assignment to the top of the method and then use it for initializing / comparing the table ref. rather than repeating the literal `'my_table'`.

docs/bigquery/snippets.py
```diff
@@ -440,64 +457,167 @@ def test_table_exists(client, to_delete):
...
+    dataset_id = 'update_table_description_dataset_{}'.format(_millis())
+    table_id = 'update_table_description_table_{}'.format(_millis())
+    dataset = bigquery.Dataset(client.dataset(dataset_id))
     client.create_dataset(dataset)
     to_delete.append(dataset)
 
-    table = bigquery.Table(dataset.table(TABLE_ID), schema=SCHEMA)
-    table.description = ORIGINAL_DESCRIPTION
+    table = bigquery.Table(dataset.table(table_id), schema=SCHEMA)
+    table.description = "Original description."
```
tseaver: Likewise, use an `original_description` variable to avoid repeating the literal.

docs/bigquery/snippets.py
```diff
@@ -440,64 +457,167 @@ def test_table_exists(client, to_delete):
...
     """Update a table's metadata."""
-    DATASET_ID = 'update_table_multiple_properties_dataset_{}'.format(
+    dataset_id = 'update_table_multiple_properties_dataset_{}'.format(
         _millis())
-    TABLE_ID = 'update_table_multiple_properties_table_{}'.format(_millis())
-    dataset = bigquery.Dataset(client.dataset(DATASET_ID))
-    dataset.description = ORIGINAL_DESCRIPTION
+    table_id = 'update_table_multiple_properties_table_{}'.format(_millis())
+    dataset = bigquery.Dataset(client.dataset(dataset_id))
+    dataset.description = "Original description"
```
tseaver: Again, avoid repeating the literals, please.

docs/bigquery/snippets.py
tswast: Yeah, let's be consistent within this method at least.

I'd also be okay with repeating the literal `'my_table'` everywhere, since that makes it 100% clear in the sample that we are talking about string IDs.

docs/bigquery/snippets.py
tswast: I'm okay with this being a literal to make it 100% clear the expected data type in the snippet itself, but it should use single-quotes.

docs/bigquery/snippets.py
```diff
@@ -440,64 +457,167 @@ def test_table_exists(client, to_delete):
...
+    assert table.expires is None
+
+    # set table to expire 5 days from now
+    expiration = datetime.datetime.now(pytz.utc) + datetime.timedelta(days=5)
+    table.expires = expiration
+    table = client.update_table(table, ['expires'])  # API request
+
+    # expiration is stored in milliseconds
+    margin = datetime.timedelta(microseconds=1000)
+    assert expiration - margin <= table.expires <= expiration + margin
```
tswast: :+1: I like how you showed the precision here.

docs/bigquery/snippets.py
```diff
@@ -440,64 +457,167 @@ def test_table_exists(client, to_delete):
...
+        bigquery.SchemaField('full_name', 'STRING', mode='REQUIRED'),
+        bigquery.SchemaField('age', 'INTEGER', mode='REQUIRED'),
+    ]
+    table_ref = client.dataset(dataset_id).table(table_id)
+    table = bigquery.Table(table_ref, schema=original_schema)
+    table = client.create_table(table)
+    assert all(field.mode == 'REQUIRED' for field in table.schema)
+
+    # SchemaField properties cannot be edited after initialization.
+    # To make a changes, new SchemaField objects must be constructed.
```
tswast: Nit: use active voice, not passive voice for instructions.

"To make changes, construct new SchemaField objects."

docs/bigquery/snippets.py
tswast: Single quotes for strings.

docs/bigquery/snippets.py
```diff
@@ -536,6 +656,41 @@ def test_update_table_cmek(client, to_delete):
     # [END bigquery_update_table_cmek]
 
 
+def test_browse_table_data(client, to_delete, capsys):
+    """Retreive selected row data from a table."""
+
+    # [START bigquery_browse_table]
```
tswast: I see that "max_results" is used for all `list_rows` calls in this snippet. Could we add one without that so that we can explain that it's an iterator over all pages of results?

docs/bigquery/snippets.py
```diff
@@ -871,67 +1078,114 @@ def test_copy_table_cmek(client, to_delete):
...
+    bucket_name = 'extract_shakespeare_compress_{}'.format(_millis())
+    storage_client = storage.Client()
+    bucket = storage_client.create_bucket(bucket_name)  # API request
     to_delete.append(bucket)
-    to_delete.insert(0, destination)
+
+    # [START bigquery_extract_table_compressed]
+    # client = bigquery.Client()
+    # bucket_name = 'my-bucket'
+    destination_uri = 'gs://{}/{}'.format(bucket_name, 'shakespeare.zip')
```
tswast: `.gz` is the common extension for gzipped files.

alixhami: I was following the command-line example [here](https://cloud.google.com/bigquery/docs/exporting-data#compression), which is where this snippet will go. Should I update the command-line example to use `.gz` as well when I make the docs update?

docs/bigquery/snippets.py
alixhami: Added in c7d849b

docs/bigquery/snippets.py
tswast: Ah good catch. I've mailed CL 188763757 to update the docs.

PR #5019, opened 2018-03-12
Nack messages when the subscriber callback errors

jonparrott:
Towards #5005

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 
 import collections
 from concurrent import futures
+import functools
 import logging
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/thread.py
```diff
@@ -85,9 +82,8 @@ def __init__(self, client, subscription, flow_control=types.FlowControl(),
             flow_control=flow_control,
             subscription=subscription,
         )
-        # Default the callback to a no-op; the **actual** callback is
-        # provided by ``.open()``.
-        self._callback = _do_nothing_callback
+        # The **actual** callback is provided by ``.open()``.
+        self._callback = None
```
lukesneeringer: Are you testing `callable` before the callback is called?

jonparrott: Not sure what you mean?

PR #5020, opened 2018-03-12
Drop leased messages after flow_control.max_lease_duration has passed.

### LINKED ISSUE #5005: PubSub: Sub client not retrieving message again if it doesn't ack
I'm experiencing an issue where messages are not retrieved again when a message is not ack'ed. 

In the below code, the self.sub is opened and the main thread sleeps. When the process method fails, the message is not ack'ed and I am expecting it to be resent in the ack deadline of the subscription (10 seconds this case). However, the message is never receieved again in this subclient, unless I completely restart the subclient and open a new callback on the same subscription. The subscription continues to listen to new messages though.

In other threads (such as this one https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4615) and in the docs, it mentions using `future.result()` or `future.exception()`, but both of these just block and never show a problem has occured.

What is the best way to fix this sort of problem? Additionally, is there a way to get the error to propagate? 

A snippet:
```
import random

from google import auth
from google.auth import exceptions
from google.auth.credentials import AnonymousCredentials
from google.cloud import pubsub


class TestProcessor:

    def process_message(self, message):
        print('processing...')
        if random.randint(0, 1): # Some unforeseen error
            raise ValueError
        print(message)
        message.ack()


class GoogleSubClient:

    def __init__(self, project_id, topic, subscription, message_processor, max_messages=10):
        try:
            credentials, _ = auth.default()
        except exceptions.DefaultCredentialsError:  # For running on emulator
            credentials = AnonymousCredentials()
        self._subscriber = pubsub.SubscriberClient(credentials=credentials)
        self.sub = None
        self.future = None
        self._flow = pubsub.types.FlowControl(max_messages=max_messages)
        self._topic = self._subscriber.topic_path(project=project_id, topic=topic)
        self._subscription = self._subscriber.subscription_path(project=project_id, subscription=subscription)
        self._message_processor = message_processor

    def run(self):
        self.sub = self._subscriber.subscribe(subscription=self._subscription, flow_control=self._flow)
        self.future = self.sub.open(self._message_processor.process_message)
```

pubsub version:
```
Name: google-cloud-pubsub
Version: 0.32.0
```

jonparrott:
Closes #5005

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
@@ -59,6 +59,10 @@ NackRequest = collections.namedtuple(
     ['ack_id', 'byte_size'],
 )
 
+_LeasedMessage = collections.namedtuple(
+    '_LeasedMessage',
+    ['added_time', 'size'])
+
 
 @six.add_metaclass(abc.ABCMeta)
 class BasePolicy(object):
@@ -92,8 +96,6 @@ class BasePolicy(object):
                 your own dictionary class, ensure this assumption holds
                 or you will get strange behavior.
     """
-
-    _managed_ack_ids = None
     _RETRYABLE_STREAM_ERRORS = (
         exceptions.DeadlineExceeded,
         exceptions.ServiceUnavailable,
@@ -109,6 +111,10 @@ class BasePolicy(object):
         self._future = None
         self.flow_control = flow_control
         self.histogram = _histogram.Histogram(data=histogram_data)
+        """.Histogram: the histogram tracking ack latency."""
+        self.leased_messages = {}
+        """dict[str, float]: A mapping of ack IDs to the local time when the
+            ack ID was initially leased in seconds since the epoch."""
 
         # These are for internal flow control tracking.
         # They should not need to be used by subclasses.
@@ -145,17 +151,6 @@ class BasePolicy(object):
         return self._future
 
     @property
-    def managed_ack_ids(self):
-        """Return the ack IDs currently being managed by the policy.
-
-        Returns:
-            set: The set of ack IDs being managed.
-        """
-        if self._managed_ack_ids is None:
-            self._managed_ack_ids = set()
-        return self._managed_ack_ids
-
-    @property
     def subscription(self):
         """Return the subscription.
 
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -252,10 +247,10 @@ def drop(self, items):
         # Remove the ack ID from lease management, and decrement the
         # byte counter.
         for item in items:
-            if item.ack_id in self.managed_ack_ids:
-                self.managed_ack_ids.remove(item.ack_id)
+            try:
```
lukesneeringer: A more performant version of this would be:

```python
if self.leased_messages.pop(item.ack_id, None) is not None:
    self._bytes -= item.byte_size
else:
    _LOGGER.debug('Item %s was not managed.', item.ack_id)
```

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
```diff
@@ -361,22 +358,36 @@ def maintain_leases(self):
...
-            _LOGGER.debug('Renewing lease for %d ack IDs.', len(ack_ids))
+            ack_ids = list(self.leased_messages.keys())
             if ack_ids:
-                request = types.StreamingPullRequest(
-                    modify_deadline_ack_ids=ack_ids,
-                    modify_deadline_seconds=[p99] * len(ack_ids),
-                )
+                _LOGGER.debug('Renewing lease for %d ack IDs.', len(ack_ids))
+
                 # NOTE: This may not work as expected if ``consumer.active``
```
lukesneeringer: Is this comment still true? It references `send_request`, which is no longer directly called.

pubsub/google/cloud/pubsub_v1/types.py
```diff
@@ -54,7 +54,8 @@
 FlowControl = collections.namedtuple(
     'FlowControl',
     ['max_bytes', 'max_messages', 'resume_threshold', 'max_requests',
-     'max_request_batch_size', 'max_request_batch_latency'],
+     'max_request_batch_size', 'max_request_batch_latency',
+     'max_lease_duration'],
```
lukesneeringer: I worry a little about different languages having diverging flow control tunables.

jonparrott: Me too, will follow-up internally.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
jonparrott: Yep, it is, even though `send_request` isn't called directly.

pubsub/google/cloud/pubsub_v1/subscriber/policy/base.py
jonparrott: Good call, done.

PR #5022, opened 2018-03-12
BigQuery: Updates dataset samples

alixhami:
Updates dataset samples for the following docs pages: 
+ [Creating and Using Datasets](https://cloud.google.com/bigquery/docs/datasets)
+ [Managing Datasets](https://cloud.google.com/bigquery/docs/managing-datasets)

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -107,14 +107,18 @@ def test_create_dataset(client, to_delete):
 
     # [START bigquery_create_dataset]
     # client = bigquery.Client()
-    # dataset_id = 'my_dataset'  # must be unique within your project
+    # dataset_id = 'my_dataset'
 
     # Create a DatasetReference using a chosen dataset ID.
     # The project defaults to the Client's project if not specified.
     dataset_ref = client.dataset(dataset_id)
-    # Construct a full Dataset object to send to the API
+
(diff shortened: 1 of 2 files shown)


docs/bigquery/snippets.py
```diff
@@ -85,49 +85,79 @@ def delete(self):
...
     """Create a dataset."""
-    DATASET_ID = 'create_dataset_{}'.format(_millis())
+    dataset_id = 'create_dataset_{}'.format(_millis())
 
-    # [START create_dataset]
-    # DATASET_ID = 'dataset_ids_are_strings'
-    dataset_ref = client.dataset(DATASET_ID)
+    # [START bigquery_create_dataset]
+    # client = bigquery.Client()
+    # dataset_id = 'my_dataset'  # must be unique within your project
```
tswast: Is the "must be unique" comment necessary? If we do want to indicate this, I would prefer we add a comment to the `create_dataset()` line that is could throw [google.api_core.exceptions.AlreadyExists](https://googlecloudplatform.github.io/google-cloud-python/latest/core/exceptions.html#google.api_core.exceptions.AlreadyExists).

tswast: I see you added the comment. Thanks!

PR #5037, opened 2018-03-14
BigQuery: Fixes update table expiration region tag

alixhami:


tswast:
@dpebot merge when green

PR #5040, opened 2018-03-14
Change GCS batch endpoint from /batch to /batch/storage/v1

janlugt:
This is in line with documentation at https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch.


tseaver:
Thanks for the patch!

PR #5049, opened 2018-03-15
Add text-to-speech beta.

lukesneeringer:


texttospeech/setup.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2018 Google LLC
```
jonparrott: This needs to use our new setup.py format.

texttospeech/nox.py
```diff
@@ -0,0 +1,52 @@
...
+    session.run('py.test', '--quiet', os.path.join('tests', 'unit'))
+
+
+@nox.session
+def lint_setup_py(session):
+    """Verify that setup.py is valid (including RST check)."""
+    session.interpreter = 'python3.6'
+    session.install('docutils', 'pygments')
+    session.run('python', 'setup.py', 'check', '--restructuredtext',
+                '--strict')
```
chemelnucfin: What about the system/cover tests?

texttospeech/setup.py
lukesneeringer: Done. Could you file me a bug to update that in toolkit?

texttospeech/nox.py
lukesneeringer: I am not as worried about coverage for auto-generated APIs. That is something I want to improve, but it is not my immediate priority.

PR #5054, opened 2018-03-16
Use the rpc's status to determine when to exit the request generator thread

jonparrott:


chemelnucfin:
@jonparrott Could the _RequestQueueGenerator be in api_core?

jonparrott:
@chemelnucfin possibly, but we should only promote it there once we need to.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -133,6 +133,85 @@
...
+    Example::
+
+        requests = request_queue_generator(q)
+        rpc = stub.StreamingRequest(iter(requests))
+        requests.rpc = rpc
+
+        for response in rpc:
+            print(response)
+            q.put(...)
+    """
```
tseaver: Document argument types?

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -203,8 +281,7 @@ def send_request(self, request):
         Args:
             request (Any): The request protobuf.
         """
-        with self._put_lock:
-            self._request_queue.put(request)
+        self._request_queue.put(request)
```
tseaver: Yay!  `queue.Queue` FTW

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -341,10 +307,12 @@ def _blocking_consume(self, policy):
...
 
-            request_generator = self._request_generator_thread(policy)
-            response_generator = policy.call_rpc(request_generator)
-            responses = _pausable_iterator(
-                response_generator, self._can_consume)
+            initial_request = policy.get_initial_request()
+            request_generator = _RequestQueueGenerator(
+                self._request_queue, initial_request=initial_request)
+            rpc = policy.call_rpc(request_generator)
+            request_generator.rpc = rpc
```
tseaver: Ugh, that looks racy, at least on the surface.  I guess we can't get around it, though.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -133,6 +133,85 @@
...
+        self.rpc = None
+
+    def _should_exit(self):
+        # Note: there is a possibility that this starts *before* the rpc
+        # property is set. So we have to check if self.rpc is set before seeing
+        # if it's active.
+        if self.rpc is not None and not self.rpc.is_active():
+            return True
+        else:
+            return False
```
tseaver: In the "racy" case indicated by the comment, how do we expect the user to recover from this state?

pubsub/tests/system.py
```diff
@@ -181,6 +183,8 @@ def test_subscribe_to_messages_async_callbacks(
     # Okay, we took too long; fail out.
     assert callback.calls >= 2
 
+    subscription.close()
+
```
tseaver: Should we be exposing this as a context manager?  Seems like a bug magnet to force explicit closure.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
jonparrott: It's okay that it's racy. The resolution is to just keep the generator open. The idea is that there are three states:

1. Generator created, rpc not set.
2. gRPC started a thread to consume the generator, rpc is still not set.
3. rpc is set, now start monitoring it to see when it stops.

So in all of those cases the right thing to do is to just keep the generator open.

pubsub/tests/system.py
jonparrott: This was just a "make sure that close() works". You don't have to call it (all threads are daemon threads).

tseaver:
LGTM if you add the docstrings for the arguments to `_RequestQueueGenerator`.

pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
jonparrott: Done in cc8766783

lukesneeringer:
I ran this for several hours without issue. Merging.

PR #5057, opened 2018-03-16
BigQuery: Adds Parquet SourceFormat and samples

alixhami:


docs/bigquery/snippets.py
```diff
@@ -1037,6 +1103,46 @@ def test_load_table_from_uri_truncate(client, to_delete):
...
+    client.create_dataset(dataset)
+    to_delete.append(dataset)
+
+    job_config = bigquery.LoadJobConfig()
+    job_config.schema = [
+        bigquery.SchemaField('name', 'STRING'),
+        bigquery.SchemaField('post_abbr', 'STRING')
+    ]
+    table_ref = dataset.table('us_states')
+    body = six.StringIO('Washington,WA')
```
tswast: Nit: Per https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4753 we should use BytesIO instead so that users who bump into this in our snippets files do the right thing when they have non-ASCII characters.

alixhami: Updated all `StringIO()` to `BytesIO()` in b3dd926

PR #5065, opened 2018-03-16
Link out to requester pays docs.

frankyn:
Link out to requester pays docs instead of stating there are no existing docs.


frankyn:
Thanks @chemelnucfin!

Issue #4864, opened 2018-02-09
BigQuery: Error 413 (Request Entity Too Large)!!1


VikramTiwari:
Python 3
google-cloud-bigquery==0.29.0

To reproduce:
Function:
```
def get_result_dataframe(query):
    query_job = client.query(query)
    return query_job.to_dataframe()
```
Query: 'SELECT * FROM `project.dataset.table_with_lots_of_columns` LIMIT 1000'
Result: Error 413 (Request Entity Too Large)!!1
More data from log: Seems like it queries the table first to get all the possible columns in the table and then makes an API call to the actual table with every column possible in the query's select statement. The query size then becomes too large.

tswast:
When you fetch the job resource that was created for this query, what is the destination table? Is it `project.dataset.table_with_lots_of_columns`?

Note: you can do the equivalent of

```
SELECT * FROM `project.dataset.table_with_lots_of_columns` LIMIT 1000
```

by using `client.list_rows(max_results=1000)`. Using `list_rows` for this operation will cost you much less money as well, since a `LIMIT` query still has to scan all the rows in the whole table.

VikramTiwari:
Hey, @tswast Sorry for the delayed response.
Destination table is the `project.dataset.table_with_lots_of_columns` and the queries are mostly aggregation but with most of the columns in the original table. I am also not trying to limit anything.

Also, the issue is happening on the request, not the response. Here are raw logs:

```
413 GET https://www.googleapis.com/bigquery/v2/projects/<project_id>/datasets/<alphanumeric_key>/tables/anon<alphanumeric_key>/data?selectedFields=__approximately_150_fields__: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 413 (Request Entity Too Large)!!1</title>
  <style>
[...]

PR #5072, opened 2018-03-20
BigQuery: Don't use `selected_fields` for listing query result rows

VikramTiwari:
fix for #4864


tswast:
@VikramTiwari Thanks for the contribution! I've added a commit to the PR with the system tests I was using to investigate the issue and confirm that this does fix the issue.

tswast:
Looks like a couple unit tests are failing due to some assumptions about where the schema would be available. I'll look into that.

tswast:
@tseaver Since I authored the tests for this change would you or someone else on the Python team mind taking a look before we merge this?

I've verified locally that all system tests do pass with this change.

tswast:
Thanks again @VikramTiwari !

PR #5073, opened 2018-03-20
Add see also for Bucket.create method call for Client.create_bucket() documentation.

frankyn:
Raised by @tswast, users[1] are confused by this method when trying to define storage class and location. There's no clear way to define these properties using this method and adding a link to `Bucket.create` method call.

I added a short `see also` to documentation. I have the following two questions:
1. Is this an expected way to define links out to other methods (trying to understand Python documentation). 
2. How is documentation for this client library updated at https://googlecloudplatform.github.io/google-cloud-python/latest/storage/client.html#google.cloud.storage.client.Client.create_bucket? Is it automatic when merged?

References:
[1]: https://stackoverflow.com/questions/44019273/python-to-create-google-storage-buckets/44035030#44035030

### DIFF
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -242,7 +242,8 @@ class Client(ClientWithProject):
         If the bucket already exists, will raise
         :class:`google.cloud.exceptions.Conflict`.
 
-        See also: :meth:`~.storage.Bucket.create`
+        To set additional properties when creating a bucket such as
+        :attr:`~.storage.Bucket.location`, use :meth:`~.storage.Bucket.create`.
         
         :type bucket_name: str
         :param bucket_name: The bucket name to create.
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -242,6 +242,8 @@ class Client(ClientWithProject):
         If the bucket already exists, will raise
         :class:`google.cloud.exceptions.Conflict`.
 
+        See also: :meth:`~.storage.Bucket.create`
+        
         :type bucket_name: str
         :param bucket_name: The bucket name to create.
 


storage/google/cloud/storage/client.py
```diff
@@ -242,6 +242,8 @@ def create_bucket(self, bucket_name, requester_pays=None, project=None):
         If the bucket already exists, will raise
         :class:`google.cloud.exceptions.Conflict`.
 
+        See also: :meth:`~.storage.Bucket.create`
```
tswast: I think we could be even more explicit about why someone should prefer the Bucket.create function.

Something like: To set additional properties when creating a bucket such as (link to Bucket.location), use (link to Bucket.create).

Did you think more about having this method accept a bucket or bucket name as the first argument?

frankyn: Thanks for the feedback will update documentation.

As for the `create_bucket` method, what if we were to remove this method from documentation as we shouldn't deprecate GA commands and instead surface `Bucket.create()` in this documentation? This also has a draw back this documentation is meant for `Client` and not `Bucket`..

Another approach is to update the `create_bucket(..)` to have an additional parameter `bucket` to that accepts a Bucket object with defined parameters, but I feel that if a user defines a Bucket object with required parameters they are already one step away from `Bucket.create` instead of having to call `Client.create_bucket`.

tswast: Yeah, we shouldn't deprecate methods in a GA library, but we can encourage one method over the other. It's part of the [Python philosophy](https://www.python.org/dev/peps/pep-0020/) that "There should be one-- and preferably only one --obvious way to do it."

I think we are just disagreeing on which method should be the one obvious way.

The generated libraries have all API requests as methods of the `Client` object, and that's the direction we went for the BigQuery library too, so if we were to deprecate/discourage anything, I'd prefer we discourage `Bucket.create()` even if we end up making `create_bucket()` call it under the covers.

Since you need to provide a bucket name to create a `Bucket` object in the first place, I think it would make sense for the first parameter here to accept either a string or a Bucket.

frankyn:
@tswast updated PTAL. I'm not 100% sure about the use of `:attr:` and `:meth:`. I've referenced other parts of documentation in this repository. If someone could share a link to documentation that would be awesome!

frankyn:
You beat me to it lol... I'll open an issue and we can continue conversation there.

tswast:
Since CircleCI built, the docs syntax probably worked. You can verify yourself by running `nox -s docs` from the root of the repo to build all the product docs. Then `open docs/_build/html/index.html` to view them locally. (I usually run them together: `nox -s docs && open docs/_build/html/index.html`)

The docs get republished whenever there is a release of any Google Cloud Python library.

frankyn:
@tswast thanks, I got back to this a little late. The :meth: and :attr: don't link to the reference documentation associated to the attribute or method after running `nox -s docs`. I'm guessing I'm not doing something correctly, as Firestore uses a similar approach and is able to link to related docs.

tswast:
Could be that you have to use the full module name `google.cloud.storage...` instead of a relative path?

PR #5081, opened 2018-03-20
added range downloads

javisantana:
added start and end support for download_as_ methods. ChunckedDownload class works in a different way than Download (it does not accept start to be None, that why the start check is being done.

### DIFF
--- a/storage/tests/unit/test_blob.py
+++ b/storage/tests/unit/test_blob.py
@@ -615,7 +615,7 @@ class Test_Blob(unittest.TestCase):
         bucket = _Bucket(client)
         blob = self._make_one(blob_name, bucket=bucket)
 
-        # Modify the blob so there there will be 2 chunks of size 3.
+        # Modify the blob so there there will be 2 chunks of size 2.
         blob._CHUNK_SIZE_MULTIPLE = 1
         blob.chunk_size = 2
 
--- a/storage/tests/unit/test_blob.py
+++ b/storage/tests/unit/test_blob.py
@@ -536,6 +536,35 @@ class Test_Blob(unittest.TestCase):
         transport.request.assert_called_once_with(
             'GET', download_url, data=None, headers=headers, stream=True)
 
+    def test__do_download_simple_with_range(self):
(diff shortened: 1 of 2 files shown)


storage/tests/unit/test_blob.py
```diff
@@ -536,6 +536,35 @@ def test__do_download_simple(self):
...
+        bucket = _Bucket(client)
+        blob = self._make_one(blob_name, bucket=bucket)
+
+        # Make sure this will not be chunked.
+        self.assertIsNone(blob.chunk_size)
+
+        transport = mock.Mock(spec=['request'])
+        transport.request.return_value = self._mock_requests_response(
+            http_client.OK,
+            {'content-length': '2', 'content-range': 'bytes 1-3'},
```
chemelnucfin: Hi, just curious, should this be 1-2 instead?

storage/tests/unit/test_blob.py
```diff
@@ -564,6 +593,34 @@ def test__do_download_chunked(self):
...
 
+    def test__do_download_chunked_with_range(self):
+        blob_name = 'blob-name'
+        # Create a fake client/bucket and use them in the Blob() constructor.
+        client = mock.Mock(
+            _credentials=_make_credentials(), spec=['_credentials'])
+        bucket = _Bucket(client)
+        blob = self._make_one(blob_name, bucket=bucket)
+
+        # Modify the blob so there there will be 2 chunks of size 3.
```
chemelnucfin: Is this going to be 4 chunks?

javisantana: as far as I understand there will be 2 chunks of size 2 (I changed that), why do you think there are 4 chunks?

chemelnucfin: I thought there would be 4 chunks since it was bytes from 10-20 and the chunks were 3 bytes each.    Either way I think it makes sense now.  Thanks.

chemelnucfin:
Merging. Let me know if there are any problems.

PR #5085, opened 2018-03-21
clouddns - client quota failing, unable to handle list in response.

cycomike:
Hi 
clouddns client started failing yesterday because API quota response now includes a list where a string is expected. 
Added condition to skip value if it's a list.

Original Error:
```
  File "/usr/lib/python2.7/site-packages/google/cloud/dns/client.py", line 60, in quotas
    for key, value in resp['quota'].items()
  File "/usr/lib/python2.7/site-packages/google/cloud/dns/client.py", line 61, in <dictcomp>
    if key != 'kind'}
TypeError: int() argument must be a string or a number, not 'list'
```

Caused By:
Addition of the following in the API quota response:
` (u'whitelistedKeySpecs', [{u'kind': u'dns#dnsKeySpec', u'algorithm': u'ecdsap256sha256'}, {u'kind': u'dns#dnsKeySpec', u'algorithm': u'ecdsap384sha384'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 2048, u'algorithm': u'rsasha256'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 1024, u'keyType': u'zoneSigning', u'algorithm': u'rsasha256'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 2048, u'algorithm': u'rsasha512'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 1024, u'keyType': u'zoneSigning', u'algorithm': u'rsasha512'}])`

Kind regards, Mike


cycomike:
I signed it

tseaver:
@cycomike The [API docs for `projects.get`](https://cloud.google.com/dns/api/v1/projects#resource) don't show any list values:  can you report the keys for which you *are* seeing a list?  Best would be to show the whole response body, with any project-identifying information redacted.

In addition, if the API really has changed to returning lists, then we should likely reflect that as well, which would mean doing a type-sniff to figure out what to do with the `whitelistedKeySpecs` entries.

tseaver:
@lukesneeringer @jonparrott Do we have a PoC for the back-end DNS team?

jonparrott:
Not that I know of, reaching out internally.

cycomike:
@tseaver Here is the full response for the quota information which now has the whitelistedKeySpecs entity. Yes you are right that something probably should be done with it.

`{u'quota': {u'kind': u'dns#quota', u'rrsetAdditionsPerChange': 1000, u'resourceRecordsPerRrset': 100, u'totalRrdataSizePerChange': 100000, u'dnsKeysPerManagedZone': 4, u'rrsetsPerManagedZone': 10000, u'whitelistedKeySpecs': [{u'kind': u'dns#dnsKeySpec', u'algorithm': u'ecdsap256sha256'}, {u'kind': u'dns#dnsKeySpec', u'algorithm': u'ecdsap384sha384'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 2048, u'algorithm': u'rsasha256'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 1024, u'keyType': u'zoneSigning', u'algorithm': u'rsasha256'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 2048, u'algorithm': u'rsasha512'}, {u'kind': u'dns#dnsKeySpec', u'keyLength': 1024, u'keyType': u'zoneSigning', u'algorithm': u'rsasha512'}], u'rrsetDeletionsPerChange': 1000, u'managedZones': 10000}, u'kind': u'dns#project', u'number': u'1234567', u'id': u'XXXXXXX'}
`

seils:
Looks like the whitelistedKeySpecs[] addition is documented as part of https://cloud.google.com/dns/api/v1beta2/projects.

tseaver:
@seils Interesting that the field would be removed in the `v1` [API docs for projects](https://cloud.google.com/dns/api/v1/projects).

@theacodes Any status on a PoC?

tseaver:
@cycomike I'm sorry we've let your patch languish so long.  At this point, it seems likely to me that we won't be able to provide proper support for zones created with the `whitelistedKeySpecs` property under the `v1beta2` API.

cycomike:
> Cloud dns api still throws error on accessing project quotas due to the whitelistedKeySpecs. The error thrown was :-
> 
> ```
>   key: int(value) for key, value in resp["quota"].items() if key != "kind"
>   TypeError:int() argument must be a string, a bytes-like object or a number, not 'list
> ```
> 
> How to handle this ?

Hi

The changes were never merged to fix this issue, the patch I created is still sitting in this pull request.
https://github.com/googleapis/google-cloud-python/pull/5085/commits/4437bc4c6ef58d7a7f68f46fdd5929ddfaf09ec6

NOTE: It does not do anything with the whitelistedkeys, so if you need that data additional changes would be required.

tseaver:
Two points about this PR:

- Adding "correct" support for the `whitelistedKeySpecs` resource field would be far preferable to just skipping over it.  I realize that involves a lot more work.  FWIW, I cannot see any reason why that field is under the top-level `quotas` field:  I think the best choice is to adjust the `quotas` method to skip that field completely (by name, rather than by type), and then add a new `whitelisted_key_specs` method which returns the values from the nested field, if present.
- The code for the DNS library has been moved to a [separate repository](https://github.com/googleapis/python-dns/).  Any follow-ups should happen there.

PR #5098, opened 2018-03-23
Firestore: Removed get_field_path

chemelnucfin:
Function would fail for [a.b, c.d]) and return `"a.b.c.d"`, which should return 
```
`a.b`.`c.d`
```
It has been replaced by the FieldPath class with unique representation.

### DIFF
--- a/firestore/tests/unit/test__helpers.py
+++ b/firestore/tests/unit/test__helpers.py
@@ -323,7 +323,7 @@ class TestFieldPathHelper(unittest.TestCase):
         field_path = _helpers.FieldPath.from_string('start')
         err_val = helper.path_end_conflict(field_path, conflicting_paths)
         self.assertIsInstance(err_val, ValueError)
-        conflict = _helpers.get_field_path([field_path.to_api_repr(), key])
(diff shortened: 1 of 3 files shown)


firestore/tests/unit/test__helpers.py
```diff
@@ -1333,7 +1321,7 @@ def test_field_updates(self):
             ('f.g', SERVER_TIMESTAMP),
         ))
         field_paths, actual_data = self._call_fut(data)
-        self.assertEqual(field_paths, ['c.d.e', 'f.g'])
+        self.assertEqual(field_paths, ['`c.d`.e', 'f.g'])
```
tseaver: Those back-ticks are actually expected here?  Do we have docs which define that?

chemelnucfin: It does seem a little strange to me also, but I think it might have to do with how using 'a.b' is actually discouraged because you can't tell whether it is a nest or a field name.  I'm trying to work on some issues relating to this.  Adding a do not merge tag.

chemelnucfin:
Closing this for now.

PR #5100, opened 2018-03-23
Add monitoring missing dependency: pandas

kabute:
Add missing pandas dependency as per:

google/cloud/monitoring/_dataframe.py
google/cloud/monitoring/client.py
google/cloud/monitoring/query.py

### DIFF
--- a/monitoring/README.rst
+++ b/monitoring/README.rst
@@ -76,3 +76,15 @@ to connect to Stackdriver Monitoring using this Client Library.
    :target: https://pypi.org/project/google-cloud-monitoring/
 .. |versions| image:: https://img.shields.io/pypi/pyversions/google-cloud-monitoring.svg
    :target: https://pypi.org/project/google-cloud-monitoring/
+
+
+Optional Features/Packages
+--------------------------
+
+**Pandas:**
+
+Pandas is an optional dependency that is needed if *query.as_dataframe* is used.
+
+To install pandas package run:
+
+    $ pip install --upgrade .[pandas]
--- a/monitoring/setup.py
+++ b/monitoring/setup.py
@@ -31,9 +31,9 @@ release_status = 'Development Status :: 3 - Alpha'
 dependencies = [
     'google-cloud-core<0.29dev,>=0.28.0',
     'google-api-core<2.0.0dev,>=0.1.1',
-    'pandas>=0.22.0',
 ]
 extras = {
+    'optional': ['pandas>=0.22.0'],
 }
 
 
--- a/monitoring/setup.py
+++ b/monitoring/setup.py
@@ -33,7 +33,7 @@ dependencies = [
     'google-api-core<2.0.0dev,>=0.1.1',
 ]
 extras = {
-    'optional': ['pandas>=0.22.0'],
+    'pandas': ['pandas>=0.22.0'],
 }
 
 
--- a/monitoring/setup.py
+++ b/monitoring/setup.py
@@ -31,6 +31,7 @@ release_status = 'Development Status :: 3 - Alpha'
 dependencies = [
     'google-cloud-core<0.29dev,>=0.28.0',
     'google-api-core<2.0.0dev,>=0.1.1',
+    'pandas>=0.22.0',
 ]
 extras = {
 }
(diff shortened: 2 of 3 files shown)


kabute:
(Fixed unit testing for pandas)

tswast:
Pandas should probably be an "extras" as was done here: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4354

kabute:
Given that client _query_ will break (Pandas seems to be a hard dependency), can I suggest changes to?

- Add the dependency to the README (indicate users this fact).
- Fix unit testing (won't pass, including coverage) and put pandas tests on its own file.

jonparrott:
It's fine with me if this needs to directly depend on Pandas. Is the query method the most commonly used method here?

tseaver:
@jonparrott It looks like `pandas` should only be a conditional dependency:  folks who want to use the rest of the package, but *not* use `query.as_dataframe`, shouldn't need to install that rather large dependency.

jonparrott:
Okay so it's not needed to query, just need to do `query.as_dataframe`. In which case, yeah, this should just be optional.

kabute:
Then _extras_ is probably the best way to go (and add the conditional back on the unit testing). What do you think?

kabute:
Added pandas as optional (and documented how to install it).

monitoring/setup.py
```diff
@@ -33,6 +33,7 @@
     'google-api-core<2.0.0dev,>=0.1.1',
 ]
 extras = {
+    'optional': ['pandas>=0.22.0'],
```
jonparrott: we've generally named our extras after the thing they pull in, so I would prefer this extra be called "pandas"

kabute: Changed (and reflected on the README).

monitoring/README.rst
```diff
@@ -76,3 +76,15 @@ to connect to Stackdriver Monitoring using this Client Library.
...
 .. |versions| image:: https://img.shields.io/pypi/pyversions/google-cloud-monitoring.svg
    :target: https://pypi.org/project/google-cloud-monitoring/
+
+
+Optional Features/Packages
+--------------------------
+
+**Pandas:**
+
+Pandas is an optional dependency that is needed if *query.as_dataframe* is used.
```
jonparrott: Use `` instead of * for query.as_dataframe here.

monitoring/README.rst
```diff
@@ -76,3 +76,15 @@ to connect to Stackdriver Monitoring using this Client Library.
...
+
+To install pandas package run:
+
+    $ pip install --upgrade .[pandas]
```
jonparrott: We should assume people aren't installing from source, but rather, form the package index. This should be ``pip install google-cloud-monitoring[pandas]``.

monitoring/tests/unit/test__dataframe.py
```diff
@@ -14,14 +14,13 @@
 
 try:
     import pandas
-except ImportError:
+except ImportError:  # pragma: NO COVER
```
jonparrott: At least one of these two branches should happen in our normal tests. I would assume that the `HAVE_PANDAS = True` branch is the one we want to ensure is covered.

monitoring/tests/unit/test__dataframe.py
```diff
@@ -14,14 +14,13 @@
...
     HAVE_PANDAS = False
-else:
-    HAVE_PANDAS = True  # pragma: NO COVER
+else:  # pragma: NO COVER
```
jonparrott: This can go in the `try` branch instead of needing an `else` branch.

monitoring/tests/unit/test__dataframe.py
```diff
@@ -202,12 +199,11 @@ def test_empty_table_smart_labels(self):
         dataframe = self._call_fut([])
         self.assertEqual(dataframe.shape, (0, 0))
         self.assertEqual(dataframe.columns.names, NAMES)
-        self.assertEqual(dataframe.columns.name, NAME)
         self.assertIsNone(dataframe.index.name)
         self.assertIsInstance(dataframe.index, pandas.DatetimeIndex)
 
 
-class Test__sorted_resource_labels(unittest.TestCase):
+class Test__sorted_resource_labels(unittest.TestCase):  # pragma: NO COVER
```
jonparrott: why are we skipping coverage on this entire class?

jonparrott:
superseded by #5212

PR #5107, opened 2018-03-23
BigQuery: Removes _ApiResourceProperty class

alixhami:
Refactors external configurations to not use _ApiResourceProperty and related classes. Using standard Python properties will make the following more clear:

+ What an object's properties are
+ How a property is stored in the API (e.g., 'only_read_latest' is stored as 'onlyReadLatest')
+ The type used for python vs. the API

Issue was previously discussed in PR [4786](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4786#issuecomment-360854465)


bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -41,49 +37,100 @@ class BigtableColumn(object):
...
-        'field_name', 'fieldName', six.string_types)
-    """See
-    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.%28key%29.bigtableOptions.columnFamilies.columns.fieldName
-    https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.fieldName
-    """
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.%28key%29.bigtableOptions.columnFamilies.columns.encoding
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.encoding
+        """
+        prop = self._properties.get('encoding')
```
tswast: nit: couldn't this just be `return self._properties.get('encoding')` without the `prop =`?

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -41,49 +37,100 @@ class BigtableColumn(object):
...
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.%28key%29.bigtableOptions.columnFamilies.columns.qualifierEncoded
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.qualifierEncoded
+        """
+        prop = self._properties.get('qualifierEncoded')
+        return prop
+
+    @qualifier_encoded.setter
+    def qualifier_encoded(self, value):
+        self._properties['qualifierEncoded'] = six.binary_type(value)
```
tswast: Do we need to base64 encode here?

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -41,49 +37,100 @@ class BigtableColumn(object):
...
+        """Union[str, bytes]: The qualifier encoded in binary.
+
+        The type is ``str`` (Python 2.x) or ``bytes`` (Python 3.x). The module
+        will handle base64 encoding for you.
+
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.%28key%29.bigtableOptions.columnFamilies.columns.qualifierEncoded
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.bigtableOptions.columnFamilies.columns.qualifierEncoded
+        """
+        prop = self._properties.get('qualifierEncoded')
```
tswast: Do we need to base64 decode here? (Note if we do it here, we'll also have to handle `None`)

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -259,47 +376,99 @@ class CSVOptions(object):
...
+
+    @property
+    def skip_leading_rows(self):
+        """int: The number of rows at the top of a CSV file.
+
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).csvOptions.skipLeadingRows
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.csvOptions.skipLeadingRows
+        """
+        prop = self._properties.get('skipLeadingRows')
```
tswast: We have a function in `_helpers` called `int_or_none` which would be helpful here.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -341,12 +508,23 @@ class GoogleSheetsOptions(object):
...
+    @property
+    def skip_leading_rows(self):
+        """int: The number of rows at the top of a sheet that BigQuery will
+        skip when reading the data.
+
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).googleSheetsOptions.skipLeadingRows
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.googleSheetsOptions.skipLeadingRows
+        """
+        prop = self._properties.get('skipLeadingRows')
```
tswast: Ditto for using `int_or_none` here.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -412,46 +588,99 @@ def options(self):
...
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.tableDefinitions.(key).schema
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.schema
+        """
+        prop = self._properties.get('schema', [])
+        return prop
+
+    @schema.setter
+    def schema(self, value):
+        self._properties['schema'] = value
```
tswast: We might want to go all the way and convert these to API format now. (Check out LoadJobConfig.schema for an example)

bigquery/tests/unit/test_external_config.py
```diff
@@ -56,15 +55,29 @@ def test_api_repr_base(self):
...
+        ec = external_config.ExternalConfig('')
+        ec.source_uris = self.SOURCE_URIS
+        ec.max_bad_records = 17
+        ec.autodetect = True
+        ec.ignore_unknown_values = False
+        ec.compression = 'compression'
+
+        got_resource = ec.to_api_repr()
+
+        self.assertEqual(got_resource, self.BASE_RESOURCE)
```
tswast: Don't use `BASE_RESOURCE` here, use a dictionary with keys and values instead.

Even if it's redundant with a shared value, a test should contain everything you need to understand it's correctness. Your `test_to_api_repr_sheets` is an example of a test that does this right.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -412,46 +542,98 @@ def options(self):
...
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#externalDataConfiguration.schema
+        """
+        prop = self._properties.get('schema', {})
+        return [SchemaField.from_api_repr(field)
+                for field in prop.get('fields', [])]
+
+    @schema.setter
+    def schema(self, value):
+        prop = value
+        if value:
```
tswast: I think we probably want this to explicitly check for `is None`. If an empty iterable (`()` or `[]`) is passed in we don't want it being set as the schema (then we'd miss the `{'fields': ...}` part.

Also, Circle says we should add a unit test for passing in None. (We probably want a unit test for passing in `[]`, too).

bigquery/tests/unit/test_external_config.py
```diff
@@ -56,15 +55,64 @@ def test_api_repr_base(self):
...
+            'sourceUris': self.SOURCE_URIS,
+            'maxBadRecords': 17,
+            'autodetect': True,
+            'ignoreUnknownValues': False,
+            'compression': 'compression',
+            'schema': exp_schema
+        }
+        self.assertEqual(got_resource, exp_resource)
+
+    def test_schema_None(self):
```
tswast: I guess this passes flake8?

alixhami: I'm not sure what you mean by that. Is it a bad name? Yes everything is passing.

tswast: Yeah, it's an odd name, since it wouldn't be valid in a regular function to use a capital letter. But I think I've seen tests use `..._None` before, so I'm fine so long as `flake8` is also fine with it.

PR #5108, opened 2018-03-23
Make the pausable response iterator aware of the RPC state to prevent deadlock

### LINKED ISSUE #5103: Subscriber can effectively deadlock itself when there are pending requests.
Due to emergent behavior between three different systems:

1. Re-establishing the stream on error.
2. Pausing the stream when over load.
3. Factoring request queue size in load.

The subscriber can deadlock itself if it tries to re-establish a stream while the stream is paused due to request pressure, for example:

```
===== Subscriber Monitor =====
Messages processed: 1659
Rate: 0.82 Messages/second
Run time: 0:33:35.001386
Load: 1.67
p99 ack: 299 seconds
Leased Messages: 0
Executor queue size: 0
Callback queue size: 0
Request queue size: 167
```

The load is `1.67` because there are `167` requests which means the stream is *paused*. Because the stream is paused when the consumer tries to restart the stream the RPC is never started because to *start* the RPC the iterator must at least be started and the request iterator is blocked on the load going down. It won't go down because it *can't* send requests until the RPC is started.

jonparrott:
Closes #5103

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
@@ -251,7 +251,7 @@ def _pausable_response_iterator(iterator, can_continue, period=1):
         Any: The items yielded from ``iterator``.
     """
     while True:
-        can_yield = can_continue.wait(timeout=1)
+        can_yield = can_continue.wait(timeout=period)
         # Calling next() on a cancelled RPC will cause it to raise the
         # grpc.RpcError associated with the cancellation.
         if can_yield or not iterator.is_active():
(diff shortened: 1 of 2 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_consumer.py
```diff
@@ -228,6 +228,36 @@ def __iter__(self):
...
+    worker to pause and resume the iteration over ``iterator``.
+
+    Args:
+        iterator (grpc.RpcContext, Iterator[protobuf.Message]): A
+            ``grpc.RpcContext`` instance that is also an iterator of responses.
+            This is a typically returned from grpc's streaming response call
+            types.
+        can_continue (threading.Event): An event which determines if we
+            can advance to the next iteration. Will be ``wait()``-ed on
+            before
```
lukesneeringer: before what? :-)

jonparrott: Good catch, fixed in 895c9b7e2

PR #5109, opened 2018-03-24
Firestore: from_api_repr and parse_field_path methods

chemelnucfin:


tseaver:
@chemelnucfin Although PRs may *introduce* bugs, we don't want to classify them *as* bugs.  If this PR is fixing an existing bug (or even working toward it), please link that issue in the description.

chemelnucfin:
@jba could you take a look at this one first?

Thanks.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -150,10 +154,72 @@ def from_string(string):
...
+            return field_names[0], '.'.join(field_names[1:])
+        else:
+            index = 1
+            while index < len(api_repr):
+                if api_repr[index] == '\\':  # escape character
+                    index += 2
+                    if api_repr[index-1] == '`':  # skips escaped backticks
+                        value = (
+                            api_repr[:index+1],
+                            '.'.join(api_repr[index+1:].split('.')))
```
jba: Since split and join are inverses, isn't this the same as `api_repr[index+1:]`?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -150,10 +154,72 @@ def from_string(string):
...
+        Returns:
+            Tuple[str, str]:
+                A tuple with the first field name and the api_repr
+                of the rest.
+        """
+        if api_repr[0] == '.':
+            api_repr = api_repr[1:]
+        if api_repr[0] != '`':  # first field name is simple
+            field_names = api_repr.split('.')
+            return field_names[0], '.'.join(field_names[1:])
```
jba: This is an expensive way to do this. Couldn't you search for the index of the first dot instead?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -150,10 +154,72 @@ def from_string(string):
...
+        else:
+            index = 1
+            while index < len(api_repr):
```
jba: If given the string
```
a.`
```
won't you return `('a', None)`?

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -150,10 +154,72 @@ def from_string(string):
             if invalid_character in string:
                 raise ValueError('Invalid characters in string.')
         string = string.split('.')
-        return FieldPath(*string)
+        return cls(*string)
+
+    @classmethod
+    def _parse_field_name(cls, api_repr):
```
jba: I think indexing a string `s[i]` is expensive in Python, because it creates a new string, right? So maybe use a regexp to find the parts?

By the way, I never needed this functionality in the Go client. When does it come up?

firestore/tests/unit/test__helpers.py
```diff
@@ -133,64 +133,90 @@ def test_unicode(self):
         self.assertEqual(field_path.parts, ('一', '二', '三'))
 
     def test_to_api_repr_a(self):
```
jba: It will be much easier and shorter to write a single test that loops over a list of (input, expected) pairs.

firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -131,10 +131,16 @@ def __init__(self, *parts):
...
-    @staticmethod
-    def from_string(string):
+    @classmethod
+    def from_string(cls, string):
         """ Creates a FieldPath from a unicode string representation.
 
+        The constructor must be used if the unicode string representation
+        consists of invalid characters `~*/[]`. This method always splits
+        on the character `.` for nesting. If any field names in this field
+        path includes the character `.`, the constructor must also be used
```
jba: You still have the phrase "the constructor must also be used to determine nesting", which I don't understand. I don't think "nesting" is a good word to use here.

Let's try:

"This method splits on the character `.` and disallows the characters `~*/[]`. To create a FieldPath whose components have those characters, call the constructor."

firestore/google/cloud/firestore_v1beta1/_helpers.py
jba: Before you write any more code, let's figure out if you need `from_api_repr`. Where do you use it in the client?

chemelnucfin: @jba I think having it would help me write code for tests that uses the actual api_repr, which is not ambiguous as in the case of using strings.  I know I could use the constructor instead, but it might not have been as clear.

chemelnucfin: And I think it was also the `parse_field_name` function I could also use elsewhere.

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: That sounds to me like you can't use this function to call FieldPath.from_string('a.b'), which creates the api_repr `a.b`.  But you won't be able to use it to do `(bt)a.b(bt)`

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I don't think that is a valid api_repr?

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: yes, I think I just used the first method I thought of...

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: oops.  :)

firestore/google/cloud/firestore_v1beta1/_helpers.py
chemelnucfin: I thought about using regex, but I think it was easier for me to write it this way to get the behavior right.

firestore/google/cloud/firestore_v1beta1/_helpers.py
jba: It's a FieldPath whose *components* have those characters. The components of `FieldPath.from_string('a.b')` are `a` and `b`, which have no invalid characters or dots.

firestore/google/cloud/firestore_v1beta1/_helpers.py
jba: For tests, use the constructor to build the exact FieldPath you need. It doesn't make sense to write this complicated parsing code when you don't really need it.

chemelnucfin: ok, closing this and moved the cleanups to #5121 and #5122.

chemelnucfin:
@jba I think I found the scenario where I needed this from_api_repr.

From this test `(update-10)`
```
description: "prefix #2"
update {
  doc_ref_path: "projects/projectID/databases/(default)/documents/C/d"
  json_data: "{\"a\": 1, \"a.b\": 2}"
  is_error: true
}
```
The only reason I know that that is actually `FieldPath("a", "b")` instead of `FieldPath("(bt)a.b(bt)")` is that since this test is an error I'm assuming this is an `api_repr`.  I can't easily split on dots if it's an `api_repr`.
For example:
```
>>> json.loads("\"a.b\"")
'a.b'
>>> json.loads("\"`a.b`\"")
'`a.b`'
>>> 
```
Maybe I missed something?  How do you handle this in golang?

jba:
The JSON in the tests should correspond exactly to a Python dict that is the argument to the update method. So that JSON should map to the Python `{"a": 1, "a.b": 2}`. If I understand correctly, in dynamic languages the keys of the dict passed to update can either be strings, which are split on dots, or FieldPaths, which are interpreted literally. @schmidt-sebastian can elaborate on that. You should also read the section on update in the DocumentReference section of the API doc. Note the sentence:

   Thus,  update( { ‘foo.bar’ : 42 } ) becomes an object with the nested hierarchy of { foo { bar : 42 }}.

jba:
Just to be clear: nothing in the tests requires you to parse the API format for field paths. You only need to generate that format, not parse it.

chemelnucfin:
Ahh, I think this makes more sense now.

```
Note that the server rejects documents that contain single field names with dots that prefix match other field names. Hence, a document cannot contain both ‘a’ and ‘`a.b`’. The Server SDKs can rely on the backend to perform this validation.
```

chemelnucfin:
Closing this for now.

PR #5114, opened 2018-03-26
BigQuery: Make SchemaField be able to include description via from_api_repr method

mia-0032:
Hi,
I fixed a problem that the description was not reflected in the `SchemaField` via `from_api_repr` method.


tseaver:
@mia-0032 Thank you for the patch!

PR #5118, opened 2018-03-26
Remove extra print statement

chemelnucfin:
extra print statement?


jonparrott:
Good catch, @chemelnucfin!

PR #5119, opened 2018-03-26
BigQuery: Add snippet to run a dry run query.

tswast:
I was wondering if it was possible to run a dry run query. (It is.)

The reason I wondered is that a dry run query does not preserve its job ID. No job is actually created, so calling `query_job.result()` will fail.


tswast:
@alixhami Could you create a sample for this in our sample tracker? I think it'd be a good one for https://cloud.google.com/bigquery/docs/estimate-costs#estimating_query_costs

alixhami:
@tswast Added this to the sample tracker

PR #5123, opened 2018-03-28
Audio Logging and Recognition Metadata.

lukesneeringer:


lukesneeringer:
Do not merge until March 30. Tagging as such.

lukesneeringer:
This has a backwards incompatible change (in Python only), which is that `google.cloud.speech_v1p1beta1.proto.speech_pb2.SpeechStub` is gone. It is in `google.cloud.speech_v1p1beta1.proto.speech_pb2_grpc` now.

This is in a beta module so we are okay; we will need to determine what our path is when it bleeds into GA modules (it is a back-incompat change in protoc itself).

chemelnucfin:
@lukesneeringer ok to merge now?  It's after March 30th.

danoscarmike:
@chemelnucfin please do not merge this one.  I will update by end of week.

chemelnucfin:
@danoscarmike got it.

danoscarmike:
@chemelnucfin @lukesneeringer we're approved to merge.

PR #5124, opened 2018-03-28
Surface publish RPC errors back to the publish futures

### LINKED ISSUE #5017: Every error that is not catchable by the user should be logged via the logging module
Currently, at the very least helper threads for pubsub seem to not and they let the exception bubble up and get printed to stderr. 

I think this is bad form, there's nothing _I_ can do to catch this error, but it'd be nice if I could get it logged in the format I want.

jonparrott:
Closes #5017, Related #4914

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
@@ -210,8 +210,12 @@ class Batch(base.Batch):
                 # We failed to publish, set the exception on all futures and
                 # exit.
                 self._status = base.BatchStatus.ERROR
+
                 for future in self._futures:
                     future.set_exception(exc)
+
+                _LOGGER.exception(
+                    'Failed to publish %s messages.', len(self._futures))
                 return
 
             end = time.time()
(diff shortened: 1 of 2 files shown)


pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -201,6 +202,27 @@ def test_blocking__commit_wrong_messageid_length():
...
+def test_block__commmit_api_error():
+    batch = create_batch()
+    futures = (
+        batch.publish({'data': b'blah blah blah'}),
+        batch.publish({'data': b'blah blah blah blah'}),
+    )
+
+    # Make the API throw an error when publishing.
+    error = google.api_core.exceptions.InternalServerError('uh oh')
+    types.PublishResponse(message_ids=['a'])
```
tseaver: Why are we creating a `PublishResopnse` instance here?

jonparrott: copypasta. Fixed f53ccbff6.

pubsub/google/cloud/pubsub_v1/publisher/batch/thread.py
```diff
@@ -220,9 +235,14 @@ def _commit(self):
                 self._status = base.BatchStatus.ERROR
                 exception = exceptions.PublishError(
                     'Some messages were not successfully published.')
+
                 for future in self._futures:
                     future.set_exception(exception)
 
+                _LOGGER.error(
```
lukesneeringer: So this is _not_ bubbling up? Maybe `warnings.warn` here?

jonparrott: "Bubbling up" is ambiguous: errors in this can only "bubble" up to the main thread via the future. However, we were not *logging* errors at all. With this, even if no future is used as long as logging is enabled these errors will show up *somewhere*.

PR #5125, opened 2018-03-28
BigQuery: Moves all Table property conversion logic into properties

alixhami:
The Table class had several special methods to convert properties to/from API representation. This PR updates the properties so they are all stored in `_properties` in API format. This moves all property conversion logic to the getters and setters, which will be easier to understand and maintain for contributors.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -252,7 +252,7 @@ class Table(object):
     :param schema: The table's schema
     """
 
-    property_to_api_field = {
+    _PROPERTY_TO_API_FIELD = {
         'description': 'description',
         'friendly_name': 'friendlyName',
         'expires': 'expirationTime',
@@ -267,15 +267,13 @@ class Table(object):
     }
 
     def __init__(self, table_ref, schema=None):
-        self._properties = {'labels': {}}
-        self._properties['tableReference'] = {
-            'projectId': table_ref.project,
-            'tableId': table_ref.table_id,
-            'datasetId': table_ref.dataset_id
+        self._properties = {
+            'tableReference': table_ref.to_api_repr(),
+            'labels': {},
         }
         # Let the @property do validation.
-        self.external_data_configuration = None
-        self.schema = schema
+        if schema is not None:
+            self.schema = schema
 
     @property
     def project(self):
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -252,21 +252,29 @@ class Table(object):
...
     """
 
-    _schema = None
-
-    all_fields = [
-        'description', 'friendly_name', 'expires', 'location',
-        'partitioning_type', 'view_use_legacy_sql', 'view_query', 'schema',
-        'external_data_configuration', 'labels', 'encryption_configuration'
-    ]
+    property_to_api_field = {
```
tswast: Let's stick an underscore in front of this to keep it "private"

bigquery/google/cloud/bigquery/table.py
```diff
@@ -252,21 +252,29 @@ class Table(object):
...
+    }
 
-    def __init__(self, table_ref, schema=()):
-        self._project = table_ref.project
-        self._table_id = table_ref.table_id
-        self._dataset_id = table_ref.dataset_id
-        self._external_config = None
+    def __init__(self, table_ref, schema=None):
         self._properties = {'labels': {}}
+        self._properties['tableReference'] = {
```
tswast: Does `table_ref` not have a `to_api_repr()` function?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -690,7 +704,10 @@ def external_data_configuration(self):
         :rtype: :class:`~google.cloud.bigquery.ExternalConfig`, or ``NoneType``
         :returns: The external configuration, or None (the default).
         """
-        return self._external_config
+        prop = self._properties['externalDataConfiguration']
```
tswast: `_properties.get('externalDataConfiguration')`

bigquery/tests/unit/test_job.py
```diff
@@ -3341,8 +3341,8 @@ def test_from_api_repr_normal(self):
 
 class _Table(object):
```
tswast: This class actually seems completely unnecessary now that the real `Table` class never makes an API call. I've filed https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5126 to remove it.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -245,117 +232,97 @@ class Table(object):
...
-
-    all_fields = [
-        'description', 'friendly_name', 'expires', 'location',
-        'partitioning_type', 'view_use_legacy_sql', 'view_query', 'schema',
-        'external_data_configuration', 'labels', 'encryption_configuration'
-    ]
+    _PROPERTY_TO_API_FIELD = {
+        'friendly_name': 'friendlyName',
+        'expires': 'expirationTime',
+        'partitioning_type': 'timePartitioning',
```
tswast: Probably needs a property for `partition_expiration`, too.

tswast: Thanks, I see you added it.

PR #5126, opened 2018-03-28
BigQuery: Remove unnecessary _Table class from test_job.py

tswast:
The way the `_Table` class is used in test_job.py is redundant with the
TABLE_REF property.


tswast:
@dpebot merge when green

PR #5145, opened 2018-04-04
BigQuery: Specifies IPython version when running Python 2.7 tests

alixhami:
IPython no longer supports Python 2.7 after version 5.x. This PR updates the nox sessions to specify IPython version 5.5 for Python 2.7 tests, which is the latest version that passes the tests. Resolves the tests failures in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5143.

### DIFF
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -35,11 +35,19 @@ def default(session):
     run the tests.
     """
     # Install all test dependencies, then install this package in-place.
-    session.install('mock', 'pytest', 'pytest-cov', 'ipython', *LOCAL_DEPS)
+    session.install('mock', 'pytest', 'pytest-cov', *LOCAL_DEPS)
+
+    # Pandas does not support Python 3.4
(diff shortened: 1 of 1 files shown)


bigquery/nox.py
```diff
@@ -35,12 +35,20 @@ def default(session):
...
-    session.install('mock', 'pytest', 'pytest-cov', 'ipython', *LOCAL_DEPS)
+    session.install('mock', 'pytest', 'pytest-cov', *LOCAL_DEPS)
+
+    # Pandas does not support Python 3.4
     if session.interpreter == 'python3.4':
         session.install('-e', '.')
     else:
         session.install('-e', '.[pandas]')
 
+    # IPython does not support Python 2 after version 5.x
```
jonparrott: You should be able to use `ipython; python_version>3` directly to pip for this.

jonparrott: Oh nvm, I see you're testing with an older version. This is fine.

PR #5149, opened 2018-04-04
Firestore: __add__ function for FieldPaths

chemelnucfin:


tseaver:
I'm not sure of the utility here.  Can @jba or @schmidt-sebastian comment on how / when concatenating `FieldPath` instances would be used?

chemelnucfin:
@tseaver For something like this:
[concat](https://github.com/GoogleCloudPlatform/google-cloud-python/commit/dd4b6465af550deb642733c3e52b9ed131141bdb#diff-a51c85bccc259f5cd6e5be357ae159ceR834)

It's just for convenience, not necessity.

schmidt-sebastian:
I don't have any concerns if this internal only.  Note that this is called `append` in the Node/Java clients.

jba:
I have no objections to overloading `+` to mean append. But in Go I decided to make FieldPaths immutable, because I thought the extra cost was worth avoiding aliasing bugs.

schmidt-sebastian:
> I have no objections to overloading + to mean append. But in Go I decided to make FieldPaths immutable, because I thought the extra cost was worth avoiding aliasing bugs.

The Node and Java implementations return new copies, so maybe that's the route we should go down everywhere?

jba:
LGTM.

PR #5177, opened 2018-04-10
Correctly specifies path to WriteDisposition

justinwhite:
Minor doc update

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -1087,7 +1087,7 @@ def test_load_table_from_uri_csv_append(client, to_delete):
     assert previous_rows > 0
 
     job_config = bigquery.LoadJobConfig()
-    job_config.write_disposition = bigquery.WriteDisposition.WRITE_APPEND
+    job_config.write_disposition = bigquery.job.WriteDisposition.WRITE_APPEND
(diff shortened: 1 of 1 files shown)


tseaver:
`from google.cloud.bigquery import WriteDisposition` works of last week's `google-cloud-bigquery 0.32.0` release, and is actually the "preferred" location for the docs (AFAIK).  @tswast can you confirm?

tswast:
@tseaver That's correct. The `job` module isn't quite private, but it's better to use the explicitly public (as in used in all samples) `bigquery` module.

tswast:
And yes, this class is imported correctly into the `bigquery` module in 0.32.0 https://github.com/GoogleCloudPlatform/google-cloud-python/blob/bigquery-0.32.0/bigquery/google/cloud/bigquery/__init__.py#L111

tswast:
@justinwhite [`google-cloud`](https://pypi.org/project/google-cloud/) != [`google-cloud-bigquery`](https://pypi.org/project/google-cloud-bigquery/)

tswast:
The `google-cloud` package includes all Google Cloud packages, but was last updated in 2017.

justinwhite:
Ah, I see. Upgraded and everything looks good now. Thanks.

PR #5178, opened 2018-04-11
BigTable: Integration of GAPIC

aneepct:
Full integration of GAPIC code into BigTable including GAPIC admin api.


sduskis:
@aneepct and @zakons, are we ready to go on this from your perspective?

zakons:
@sduskis This PR has been rebased against master and the GAPIC PR #4693 was merged into our feature branch.  It is functionally complete, including the GAPIC admin API.  The admin functions previously in BigTable were removed from the code and the documentation.

Note that there was a flaw in the GAPIC generated code for the routing headers - see details in my notes on that PR.  @jonparrott has alerted @lukesneeringer that this needs to be fixed in the generation code. But he agreed that we can proceed with the manual code changes that we applied by adding a method call to ```to_grpc_metadata```.

Also note that our goal is to add back some Pythonic wrappers to the admin API for the basic functions of table creation and deletion, as well as column family creation. However, this should NOT hold up the review of this PR since it is now functionally complete.

Attached is documentation on the interface differences.
[GAPIC Client Changes.pdf](https://github.com/GoogleCloudPlatform/google-cloud-python/files/1901165/GAPIC.Client.Changes.pdf)

Please commence review. Thanks!

bigtable/google/__init__.py
```diff
@@ -1,17 +1,3 @@
-# Copyright 2016 Google LLC
```
jonparrott: This license header needs to be preserved (this applies to all files).

jonparrott:
Oh I see, I read the doc you attached. 

This makes sense, but I would not feel comfortable releasing this as-is to users. Do y'all anticipate adding more high-level stuff before cutting the next release?

For example, I would expect y'all to extend `bigtable_v1.BigtableClient` to include a helper method to gets the `Table` instance for you. See [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/client.py#L33) for how we extending pub/sub.

sduskis:
@jonparrott: For this PR, I would really like to have 0 interface changes and no breaking changes.  We ought to consider what our API should look like for Beta, but that should be in another PR.  I had an  @aneepct and @zakons about this change, and I hope that you're ok with it too.

I'd like to defer discussions about extensions/wrapping.  That discussion is probably best handled offline.

zakons:
@jonparrott Following on comment by @sduskis, the next step will be to hide the details of the GAPIC table and table admin API clients under Instance and putting back the create table, create column family and delete table interfaces, but using the new GAPIC API's under the covers.

Our discussion with @sduskis was leaning toward a fully wrapped GAPIC admin client, different from the augmented pub/sub client you shared with us. Given his post on the complexities of generating usable code for some of the protobuf specification details, it would seem to make sense to provide the users with a clean Python interface across the admin API.  As always, your feedback is helpful here as the fully wrapped admin API could kick off immediately after getting this PR satisfied by the steps above.

jonparrott:
>  For this PR, I would really like to have 0 interface changes and no breaking changes.

That's what I wanted to call out - this PR *does* break the interface (significantly), unless I'm reading it wrong.

Let me know if y'all want to merge this as-is or if it needs more work.

zakons:
@jonparrott Please see my above 'next step' details.  This is being done now and only after that would it be ready to merge. Thanks!

zakons:
@jonparrott OK, the latest commits restore the original interface, with the GAPIC clients underneath the covers. This branch is ready to merge with master.  

Please note that a next step is to wrap the GAPIC admin API to make it Pythonic across the full range of methods.  Your feedback on this next step would be useful to avoid taking the wrong steps toward this goal.

zakons:
@jonparrott OK, now this PR is REALLY ready to merge.  Cluster was taken out and now it was put back in, so the original client is now restored, with the added GAPIC API's hidden underneath. Please proceed with the merge.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -187,45 +72,38 @@ class Client(ClientWithProject):
                   interact with the Instance Admin or Table Admin APIs. This
                   requires the :const:`ADMIN_SCOPE`. Defaults to :data:`False`.
 
-    :type user_agent: str
-    :param user_agent: (Optional) The user agent to be used with API request.
-                       Defaults to :const:`DEFAULT_USER_AGENT`.
+    :type project: :instance: grpc.Channel
```
jonparrott: did you mean `channel`?

bigtable/google/cloud/bigtable/client.py
```diff
@@ -187,45 +72,38 @@ class Client(ClientWithProject):
...
         # NOTE: We set the scopes **before** calling the parent constructor.
         #       It **may** use those scopes in ``with_scopes_if_required``.
+        self.project = project
         self._read_only = bool(read_only)
         self._admin = bool(admin)
+        self.channel = channel
+        self._credentials = credentials
         self.SCOPE = self._get_scopes()
 
         # NOTE: This API has no use for the _http argument, but sending it
```
jonparrott: This note probably doesn't apply anymore.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -349,30 +198,15 @@ def instance(self, instance_id, location=_EXISTING_INSTANCE_LOCATION_ID,
...
 
     def list_instances(self):
         """List instances owned by the project.
 
-        :rtype: tuple
-        :returns: A pair of results, the first is a list of
-                  :class:`~google.cloud.bigtable.instance.Instance` objects
-                  returned and the second is a list of strings (the failed
-                  locations in the request).
+        :rtype: :class:`~google.gax.PageIterator`
```
jonparrott: This is actually `google.api_core.page_iterator.Iterator`.

bigtable/google/cloud/bigtable/cluster.py
```diff
@@ -199,24 +108,15 @@ def create(self):
 
             before calling :meth:`create`.
 
-        :rtype: :class:`Operation`
+        :rtype: :class:`~google.cloud.bigtable_admin_v2.types._OperationFuture`
```
jonparrott: should be `google.api_core.operation.Operation`.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -219,20 +156,14 @@ def create(self):
 
             before calling :meth:`create`.
 
-        :rtype: :class:`Operation`
-        :returns: The long-running operation corresponding to the
-                  create operation.
+        :rtype: :class:`~google.cloud.bigtable_admin_v2.types._OperationFuture`
```
jonparrott: same here, should be `google.api_core.operation.Operation`.

bigtable/google/cloud/bigtable/__init__.py
```diff
@@ -18,7 +18,5 @@
 from pkg_resources import get_distribution
 __version__ = get_distribution('google-cloud-bigtable').version
 
-from google.cloud.bigtable.client import Client
 
-
-__all__ = ['__version__', 'Client']
```
jonparrott: Should this be preserved? How will users import the client without it?

bigtable/google/cloud/bigtable/client.py
```diff
@@ -187,45 +72,34 @@ class Client(ClientWithProject):
...
         if read_only and admin:
             raise ValueError('A read-only client cannot also perform'
                              'administrative actions.')
 
         # NOTE: We set the scopes **before** calling the parent constructor.
         #       It **may** use those scopes in ``with_scopes_if_required``.
+        self.project = project
         self._read_only = bool(read_only)
         self._admin = bool(admin)
+        self.channel = channel
```
jonparrott: will you make this private? There's no need for it to be public.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -187,45 +72,34 @@ class Client(ClientWithProject):
...
-        self.user_agent = user_agent
-        self.emulator_host = os.getenv(BIGTABLE_EMULATOR)
-
-        # Create gRPC stubs for making requests.
-        self._data_stub = _make_data_stub(self)
-        if self._admin:
-            self._instance_stub_internal = _make_instance_stub(self)
-            self._operations_stub_internal = _make_operations_stub(self)
-            self._table_stub_internal = _make_table_stub(self)
+        if self.channel is not None:
```
jonparrott: why is this here?

jonparrott:
@zakons I reviewed, PTAL.

crwilcox:
@zakons  and @sduskis 
Currently master is not passing tests for BigTable. They are failing because project isn't being set. It seems in the rework Client stopped inheriting from ClientWithProject and now we aren't auto-magically setting the user's project anymore and instead are assuming the project name is 'None'.

If you make the following modifications to `google/cloud/bigtable/client.py`
```
+from google.cloud.client import ClientWithProject

 from google.cloud import bigtable_v2
 from google.cloud import bigtable_admin_v2
@@ -44,7 +45,7 @@ READ_ONLY_SCOPE = 'https://www.googleapis.com/auth/bigtable.data.readonly'
 """Scope for reading table data."""


-class Client(object):
+class Client(ClientWithProject):
```

You get further, but the next thing you will get is a failure on 43 tests about `ValueError: The `channel` and `credentials` arguments to BigtableClient are mutually exclusive.`

zakons:
@crwilcox These system tests are being fixed right now and will let you and @jonparrott know as soon as a PR is ready with this.

PR #5187, opened 2018-04-11
Allow uploading files larger than 2GB by using Resumable Media Requests

crwilcox:
Change the logic to elect to use multipart or resumable media for upload requests.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -879,9 +879,9 @@ class Blob(_PropertyMixin):
                    size, num_retries, predefined_acl):
         """Determine an upload strategy and then perform the upload.
 
-        If the size of the data to be uploaded exceeds 5 MB, or the size is
-        unknown, a resumable media request will be used, otherwise the content
-        and the metadata will be uploaded in a single multipart upload request.
+        If the size of the data to be uploaded exceeds 5 MB a resumable media 
+        request will be used, otherwise the content and the metadata will be
+        uploaded in a single multipart upload request.
 
         The content type of the upload will be determined in order
         of precedence:
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -879,7 +879,7 @@ class Blob(_PropertyMixin):
                    size, num_retries, predefined_acl):
         """Determine an upload strategy and then perform the upload.
 
-        If the size of the data to be uploaded exceeds 5 MB a resumable media 
+        If the size of the data to be uploaded exceeds 5 MB a resumable media
         request will be used, otherwise the content and the metadata will be
         uploaded in a single multipart upload request.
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -93,6 +93,8 @@
 _READ_LESS_THAN_SIZE = (
     'Size {:d} was specified but the file-like object only had '
     '{:d} bytes remaining.')
+_DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
+_MAX_MULTIPART_SIZE = 5 * 1024 * 1024 # 5 MB
```
frankyn: Provide constant instead of multiplying value.

storage/tests/unit/test_blob.py
```diff
@@ -1346,14 +1346,14 @@ def _do_upload_helper(
...
+        if size is None:
+            size = 12345654321
         # Make the request and check the mocks.
         created_json = blob._do_upload(
             client, stream, content_type, size, num_retries, predefined_acl)
         self.assertIs(created_json, mock.sentinel.json)
         response.json.assert_called_once_with()
-        if chunk_size is None:
-
+        if size is not None and size < 1024*1024*5:
```
frankyn: Swap with _MAX_MULTIPART_SIZE

storage/tests/unit/test_blob.py
```diff
@@ -1364,8 +1364,8 @@ def _do_upload_helper(
                 client, stream, content_type, size, num_retries,
                 predefined_acl)
 
-    def test__do_upload_without_chunk_size(self):
-        self._do_upload_helper()
+    def test__do_upload_small_file(self):
+        self._do_upload_helper(size=100)
 
     def test__do_upload_with_chunk_size(self):
         chunk_size = 1024 * 1024 * 1024  # 1GB
```
frankyn: Should this be 1GB or `_DEFAULT_CHUNKSIZE`?

crwilcox: This could be changed to this for sure. It just needs to be smaller then 5MB so it uses the non-resumable codepath.

storage/tests/unit/test_blob.py
```diff
@@ -1364,8 +1367,9 @@ def _do_upload_helper(
...
                 predefined_acl)
 
-    def test__do_upload_without_chunk_size(self):
-        self._do_upload_helper()
+    def test__do_upload_uses_multipart(self):
+        self._do_upload_helper(
+            size=google.cloud.storage.blob._MAX_MULTIPART_SIZE)
 
     def test__do_upload_with_chunk_size(self):
         chunk_size = 1024 * 1024 * 1024  # 1GB
```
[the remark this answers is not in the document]
crwilcox: Oh, I think I may have been looking at the wrong area. I focused on the test I had altered. I think that makes sense.

PR #5199, opened 2018-04-16
BigQuery: Adds samples from python-docs-samples

alixhami:
There were many redundant samples between google-cloud-python and python-docs-samples. This repo is the primary source of code samples and only had a few missing that were still being sourced from python-docs-samples. This PR adds those few remaining samples so that all samples except quickstart, simple app, and user credentials can be deleted from python-docs-samples.

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -851,7 +851,8 @@ def test_load_table_from_file(client, to_delete):
 
     job.result()  # Waits for table load to complete.
 
-    print('Job {} completed.'.format(job.job_id))
+    print('Loaded {} rows into {}:{}.'.format(
+        job.output_rows, dataset_id, table_id))
     # [END bigquery_load_from_file]
 
     table = client.get_table(table_ref)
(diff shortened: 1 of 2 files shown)


docs/bigquery/snippets.py
```diff
@@ -90,6 +91,40 @@ def delete(self):
...
+
+
+def test_create_client_json_credentials():
+    """Create a BigQuery client with Application Default Credentials"""
+    with open(os.environ['GOOGLE_APPLICATION_CREDENTIALS']) as creds_file:
+        creds_file_data = creds_file.read()
+
+    open_mock = mock.mock_open(read_data=creds_file_data)
+
+    with mock.patch('io.open', open_mock):
```
tswast: Fancy! I like this io mocking magic.

Do these snippets tests pass on Python 2.7? I know some I/O methods got moved around for Python 3.

alixhami: Ripped directly from python-docs-samples. Just checked and it's passing snippets tests for Python 2.7.

PR #5201, opened 2018-04-17
Add subscriber dispatcher

jonparrott:
Dispatcher handles sending requests *back* from callbacks. This is the *second* helper (after `BackgroundStreamConsumer`) that will be used by the `Subscriber` class.

Still trying to be minimally invasive, but I did move the `helper_threads` module to its final place under `_protocol`. 

Don't give the subscriber class much scrutiny right now, I just need *some* interface for the autospec in tests.

### DIFF
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
@@ -79,7 +79,6 @@ def test_start_already_started(thread):
 
 
 def test_stop():
-    subscriber_ = mock.create_autospec(subscriber.Subscriber, instance=True)
     queue_ = queue.Queue()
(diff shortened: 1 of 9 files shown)


pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
```diff
@@ -0,0 +1,99 @@
...
+    dispatcher_._thread = mock.sentinel.thread
+
+    with pytest.raises(ValueError):
+        dispatcher_.start()
+
+    thread.assert_not_called()
+
+
+def test_stop():
+    subscriber_ = mock.create_autospec(subscriber.Subscriber, instance=True)
```
crwilcox: linter isn't happy with this as it isn't ever used.

Issue #5003, opened 2018-03-07
Flaky BigQuery system tests


tseaver:
E.g., see:

https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/5869

tseaver:
@tswast Any clue why the back-end is failing this way?

tswast:
```
ServiceUnavailable: 503 DELETE https://www.googleapis.com/storage/v1/b/bq_load_test_5869_1520454975/o/person_ages.csv: Backend Error
```

Looks like it's having trouble cleaning up some Cloud Storage files it created. @frankyn is it expected that creating a file in GCS and then deleting it shortly after could be flakey?

The `test_load_table_from_uri_w_autodetect_schema_then_get_job` test in question is here:

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/fefac75cf728599b6a8a05812337847ef595b705/bigquery/tests/system.py#L546-L588

I might actually be tempted to delete this test altogether. I think we cover the same code with the `bigquery_load_table_gcs_json_autodetect` snippet.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/fefac75cf728599b6a8a05812337847ef595b705/docs/bigquery/snippets.py#L622-L646

tswast:
Thanks Frank. Yeah, I think it is doing backoff, but perhaps there was just a random failure for this run.

Still, I think the best action here is to delete this particular test since it is redundant with the code snippet test (which also runs as a system test).

chemelnucfin:
Made PR #5004 to remove this test.  Is it likely other tests will act the same way?

tswast:
Thanks @chemelnucfin. I'm slowly adding more "load from GCS" snippets. As I add them, I'll try to remember to remove the corresponding redundant system tests.

tseaver:
@tswast The bigquery system tests are running *very* slowly today (4 - 4.5 minutes each) today, which is causing unrelated PRs to be marked as failing (CI times the jobs out).  Is there a known issue on the back-end we should be concerned about?

tswast:
I'm not seeing anything on https://status.cloud.google.com/ or in the internal outage tracker. That extra slowness does sound like something could be going on, though.

tseaver:
FWIW, I've just cleaned out a bunch of GCS buckets with `bq-` prefixes, presumably orphaned somehow from Bigquery system tests.

tseaver:
[...]

PR #5202, opened 2018-04-17
Don't suppress 'dots' output for snippets tests.

tseaver:
CI aborts the run if the test goes longer than 10 minutes without emitting any output.

This change also eases running with verbose output for debugging long-running tests, e.g., `nox -e snippets -- --verbose`.

See #5003.

### DIFF
--- a/bigquery/nox.py
+++ b/bigquery/nox.py
@@ -144,7 +144,6 @@ def snippets(session, py):
     # Run py.test against the system tests.
     session.run(
         'py.test',
-        '--quiet',
         os.path.join(os.pardir, 'docs', 'bigquery', 'snippets.py'),
         *session.posargs
     )


tseaver:
@dpebot Please merge.

jonparrott:
You gotta tell her when

tseaver:
OK.  Maybe she should respond to non-understood mentions with a help message?

jonparrott:
@tseaver maybe. The source is over here: https://github.com/GoogleCloudPlatform/repository-gardener/tree/master/webhook-app

tseaver:
Hmm, @dpebot doesn't seem to have picked up the baton here.

jonparrott:
let me see why.

tseaver:
Oops, "ships passing".

jonparrott:
It's fine, I still found the issue.

tswast:
Thought: does `--quiet` prevent `pytest` from dumping fixture values on test failures? I recall @jonparrott had some frustration with that leaking things it shouldn't.

jonparrott:
pytest shouldn't leak anything that I'm aware of. Also the real "secret" is the service account keyfile which is a file and thus would never be logged.

PR #5209, opened 2018-04-18
BigQuery: limit list_jobs sample to 10 most recent jobs

tswast:
Listing all jobs can take a very long time, especially on a project used
by CI which starts many jobs every test run.


alixhami:
Looks good, though I think the comment could be more informative. For example:
```
List the 10 most recent jobs.
If the optional max_results parameter is not used, all jobs will be listed.
```

PR #5210, opened 2018-04-18
Add pubsub subscriber leaser

jonparrott:
This is another helper thread that maintains the leases for any messages that the current subscriber is working on. The majority of this code (notably, `maintain_leases`) is copied from `pubsub_v1.subscriber.policy.base.BasePolicy.maintain_leases`.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
@@ -1,0 +1,186 @@
+# Copyright 2017, Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import collections
+import copy
+import logging
+import random
+import threading
+import time
+
+import six
+
+from google.cloud.pubsub_v1.subscriber._protocol import requests
+
+
+_LOGGER = logging.getLogger(__name__)
+_LEASE_WORKER_NAME = 'Thread-LeaseMaintainer'
+
+
+_LeasedMessage = collections.namedtuple(
+    '_LeasedMessage',
+    ['added_time', 'size'])
+
+
+class Leaser(object):
+    def __init__(self, subscriber):
+        self._thread = None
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -0,0 +1,186 @@
...
+
+            if to_drop:
+                _LOGGER.warning(
+                    'Dropping %s items because they were leased too long.',
+                    len(to_drop))
+                self._subscriber.drop(to_drop)
+
+            # Remove dropped items from our copy of the leased messages (they
+            # have already been removed from the real one by self.drop).
+            for item in to_drop:
```
crwilcox: won't fix on this is fine, but are we heavily pro flat nesting. Really this is a second if to_drop check in disguise.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -0,0 +1,186 @@
...
+                leased_messages.pop(item.ack_id)
+
+            # Create a streaming pull request.
+            # We do not actually call `modify_ack_deadline` over and over
+            # because it is more efficient to make a single request.
+            ack_ids = list(leased_messages.keys())
```
crwilcox: as best I can tell this doesn't really need to be a list. Does this have something to do with back-compat?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -0,0 +1,186 @@
...
+                self._subscriber.modify_ack_deadline([
+                    requests.ModAckRequest(ack_id, p99) for ack_id in ack_ids])
+
+            # Now wait an appropriate period of time and do this again.
+            #
+            # We determine the appropriate period of time based on a random
+            # period between 0 seconds and 90% of the lease. This use of
+            # jitter (http://bit.ly/2s2ekL7) helps decrease contention in cases
+            # where there are many clients.
+            snooze = random.uniform(0.0, p99 * 0.9)
```
crwilcox: Randomizing the exact delay makes sense, but going to 0.0 is interesting instead of some small, but non-zero, number.

pubsub/tests/unit/pubsub_v1/subscriber/test_leaser.py
```diff
@@ -0,0 +1,235 @@
...
+    assert 'not managed' in caplog.text
+
+
+def test_remove_negative_bytes(caplog):
+    caplog.set_level(logging.DEBUG)
+
+    leaser_ = leaser.Leaser(mock.sentinel.subscriber)
+
+    leaser_.add([
+        requests.LeaseRequest(ack_id='ack1', byte_size=50)])
```
crwilcox: More general questions:
- Are messages guaranteed to have unique ack ids
- Are messages immutable?

the first is more a general question. the second is about understanding how we end up in negative byte scenarios.

PR #5217, opened 2018-04-20
BigQuery: Add optional project parameter to `list_datasets` and `list_jobs`

### LINKED ISSUE #5216: BigQuery: add project parameter to `list_datasets`, `list_jobs`
I think of a Client as corresponding to the "billing project" for BigQuery, but this should not restrict the client from listing datasets / jobs in other projects if its associated credentials have permissions to do so in those projects.

https://github.com/ibis-project/ibis/pull/1428#discussion_r183066783

tswast:
This allows a Client to list datasets or jobs in projects other than the
default project.

Closes #5216

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -186,51 +186,53 @@ class Client(ClientWithProject):
             page_token=page_token,
             max_results=max_results)
 
-    def list_datasets(self, include_all=False, filter=None, max_results=None,
-                      page_token=None, retry=DEFAULT_RETRY):
+    def list_datasets(
+            self, project=None, include_all=False, filter=None,
+            max_results=None, page_token=None, retry=DEFAULT_RETRY):
         """List datasets for the project associated with this client.
 
         See
         https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/list
 
-        :type include_all: bool
-        :param include_all: True if results include hidden datasets.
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -186,51 +186,53 @@ def list_projects(self, max_results=None, page_token=None,
             page_token=page_token,
             max_results=max_results)
 
-    def list_datasets(self, include_all=False, filter=None, max_results=None,
-                      page_token=None, retry=DEFAULT_RETRY):
+    def list_datasets(
+            self, project=None, include_all=False, filter=None,
+            max_results=None, page_token=None, retry=DEFAULT_RETRY):
```
tseaver: FWIW, putting the new argument last in the list provides somewhat better backward-compatibility for users who may not have used named parameters.  I'm not sure how important that is in this case.

tswast: Yeah, I wasn't sure. Looks like we want `retry` to be the last arg in `Client` methods. Since it's the resource you are listing over, the first place felt the most natural.

Per the last time I tried to preserve positional/kwarg semantics, it sounds like we consider these args just keyword arguments even though they could use either.

tseaver: I'm fine to merge if you think the switch to `1.0` doesn't tighten up that constraint.

PR #5229, opened 2018-04-24
Rename histogram module

jonparrott:


jonparrott:
@crwilcox FYI, I'm merging as this is just a module rename.

PR #5230, opened 2018-04-24
Add scheduler

jonparrott:


pubsub/google/cloud/pubsub_v1/subscriber/scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @abc.abstractmethod
+    def shutdown(self):
+        """Shuts down the scheduler and immediately end all pending callbacks.
+        """
+        raise NotImplementedError
+
+
+def _make_default_thread_pool_executor():
+    # Python 2.7 and 3.6+ have the thead_name_prefix argument, which is useful
```
crwilcox: thead_name_prefix should be thread_name_prefix

pubsub/google/cloud/pubsub_v1/subscriber/scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+    # for debugging.
+    executor_kwargs = {}
+    if sys.version_info[:2] == (2, 7) or sys.version_info >= (3, 6):
+        executor_kwargs['thread_name_prefix'] = (
+            'ThreadPoolExecutor-ThreadScheduler')
+    return concurrent.futures.ThreadPoolExecutor(
+        max_workers=10,
```
crwilcox: Does it make sense to modify this based on the system? why 10?

jonparrott: It's just a sane default. The existence of this class now makes it far easier to customize.

PR #5231, opened 2018-04-24
BigQuery: Reduces repetition in load samples

alixhami:
Reduces repetition in load samples discussed [here](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5003#issuecomment-382127756)

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -1012,9 +1012,12 @@ def test_load_table_from_uri_parquet(client, to_delete):
 
 
 def test_load_table_from_uri_autodetect(client, to_delete):
-    """Different file formats should have their own tested load sample
-    (test_load_table_from_uri_format), but repetitive code should be reduced
-    for autodetect, append, and truncate.
+    """Load table from a GCS URI using various formats and auto-detected schema
+
+    Each file format has its own tested load from URI sample. Because most of
+    the code is common for autodetect, append, and truncate, this sample
+    includes snippets for all supported formats but only calls a single load
+    job.
 
     This code snippet is made up of shared code, then format-specific code,
     followed by more shared code. Note that only the last format in the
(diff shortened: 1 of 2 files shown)


docs/bigquery/snippets.py
```diff
@@ -887,9 +887,10 @@ def test_load_table_from_uri_csv(client, to_delete):
     job_config.skip_leading_rows = 1
     # The source format defaults to CSV, so the line below is optional.
     job_config.source_format = bigquery.SourceFormat.CSV
+    uri = 'gs://cloud-samples-data/bigquery/us-states/us-states.csv'
 
     load_job = client.load_table_from_uri(
-        'gs://cloud-samples-data/bigquery/us-states/us-states.csv',
+        uri,
```
tswast: Why move the URI up in these first few samples?

docs/bigquery/snippets.py
```diff
@@ -1007,71 +1011,75 @@ def test_load_table_from_uri_parquet(client, to_delete):
     # [END bigquery_load_table_gcs_parquet]
 
 
-def test_load_table_from_uri_csv_autodetect(client, to_delete):
+def test_load_table_from_uri_autodetect(client, to_delete):
+    """Different file formats should have their own tested load sample
+    (test_load_table_from_uri_format), but repetitive code should be reduced
```
tswast: Nit: First line should be a one-line description of the sample.

> Snippet to load with autodetected schema (all supported formats).

Also, since you have reduced repetitive code, "should" isn't quite the right word. It reads too passive.

How about:

> Each file format has it's own tested load from URI sample. Since most of the code is common for autodetect, append, and truncate, this sample includes snippets for all supported formats but only calls a single load job.

docs/bigquery/snippets.py
```diff
@@ -1097,100 +1108,37 @@ def test_load_table_from_uri_csv_append(client, to_delete):
...
-    ]
-    table_ref = dataset.table('us_states')
-    body = six.BytesIO(b'Washington,WA')
-    client.load_table_from_file(
-        body,
-        table_ref,
-        location='US',  # Location must match that of the destination dataset.
-        job_config=job_config).result()
+    # unset csv-specific attribute
+    del job_config._properties['load']['skipLeadingRows']
```
tswast: Could we set the public property to `None` to accomplish the same thing?

alixhami: That was the first thing I tried, but it tried to send `None` to the API and it sent back an error. The whole `_properties` dict gets sent to the API, so I had to delete it to prevent that.

tswast: Okay, makes sense. We could potentially add some property deleters, but let's wait until someone has a real usecase for one.

docs/bigquery/snippets.py
alixhami: Updated

PR #5235, opened 2018-04-25
Regenerate speech v1p1beta1 

crwilcox:
Regenerate speech v1p1beta1 to account for the removal of GoogleDataCollectionConfig and google_data_collection_opt_in.


speech/google/cloud/speech_v1p1beta1/__init__.py
```diff
@@ -18,13 +18,10 @@
 from google.cloud.speech_v1p1beta1.gapic import enums
```
jonparrott: this needs to be reverted. We must keep the `SpeechHelpers` mixing.

speech/tests/unit/gapic/v1p1beta1/test_speech_client_v1p1beta1.py
```diff
@@ -197,7 +197,7 @@ def test_streaming_recognize(self):
         request = cloud_speech_pb2.StreamingRecognizeRequest(**request)
         requests = [request]
 
-        response = client._streaming_recognize(requests)
```
jonparrott: this probably also needs to be reverted.

PR #5237, opened 2018-04-26
Add streaming pull manager and integrate new subscriber logic

jonparrott:
Adds a new method, `subscribe_experimental`, to the pubsub subscriber client to use this new functionality. Leaves the older subscription implementation in place for now (will be removed by another PR after this gets more testing.)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -0,0 +1,382 @@
...
+            suitable for being the first request on the stream (and not
+            suitable for any other purpose).
+        """
+        # Any ack IDs that are under lease management need to have their
+        # deadline extended immediately.
+        lease_ids = self._leaser.ack_ids
+
+        # Put the request together.
+        request = types.StreamingPullRequest(
+            modify_deadline_ack_ids=list(lease_ids),
```
crwilcox: If not required to create as list, should remove.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -0,0 +1,382 @@
...
+        for received_message in response.received_messages:
+            message = google.cloud.pubsub_v1.subscriber.message.Message(
+                received_message.message,
+                received_message.ack_id,
+                self._scheduler.queue)
+            # TODO: Immediately lease instead of using the callback queue.
+            self._scheduler.schedule(self._callback, message)
+
+    def _should_recover(self, exception):
+        """Determine if an error on the RPC straem should be recovered.
```
crwilcox: typo

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -27,9 +28,40 @@
 
 
 class Dispatcher(object):
-    def __init__(self, queue, subscriber):
+    def __init__(self, manager, queue):
```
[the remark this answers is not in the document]
jonparrott: Yeah, I noticed while testing that the constructor was inconsistent with the other helper threads, so I went ahead and swapped it.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -55,46 +87,79 @@ def dispatch_callback(self, items):
...
+        Args:
+            items(Sequence[AckRequest]): The items to acknowledge.
         """
-        if self._thread is not None:
-            raise ValueError('Dispatcher is already running.')
+        # If we got timing information, add it to the histogram.
+        for item in items:
+            time_to_ack = item.time_to_ack
+            if time_to_ack is not None:
+                self._manager.ack_histogram.add(int(time_to_ack))
```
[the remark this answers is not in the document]
jonparrott: Yep, histogram is essentially a map of {value: count},  where values are quantized to the nearest integer. However, histogram does this itself in `add`, so I can actually remove this.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -0,0 +1,382 @@
...
+        self._rpc.add_done_callback(self._on_rpc_done)
+        self._consumer = bidi.BackgroundConsumer(
+            self._rpc, self._on_response)
+        self._consumer.start()
+
+        # Start the lease maintainer thread.
+        self._leaser = leaser.Leaser(self)
+        self._leaser.start()
+
+    def close(self, reason=None):
```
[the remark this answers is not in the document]
jonparrott: It can be called from multiple threads (namely, it could be closed by the calling `future.cancel` and `on_rpc_done` at the same time). Thankfully, this whole method is guarded by a lock (`with self._closing:`) which forces all calls to this to be handled serially regardless of concurrency. So the "first" caller will actually close and the second will hit the early out branch.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
jonparrott: We have to materialize the list at some point for protobuf, might as well be explicit here.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
jonparrott: fixed.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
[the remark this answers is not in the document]
crwilcox: I totally missed the lock :|

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -27,9 +28,40 @@
...
+        self._manager = manager
         self._queue = queue
-        self._subscriber = subscriber
+        self._thread = None
+
+    def start(self):
+        """Start a thread to dispatch requests queued up by callbacks.
+        Spawns a thread to run :meth:`dispatch_callback`.
+        """
+        if self._thread is not None:
```
lukesneeringer: Maybe add a lock? I expect the code between lines 40 to 57 is probably more or less instant, but to be safe... :-)

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -0,0 +1,382 @@
...
+        target = min([
+            self._last_histogram_size * 2,
+            self._last_histogram_size + 100,
+        ])
+        if len(self.ack_histogram) > target:
+            self._ack_deadline = self.ack_histogram.percentile(percent=99)
+        return self._ack_deadline
+
+    @property
+    def load(self):
```
lukesneeringer: This was a great idea.

jonparrott: Said the person who wrote it. :P

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
jonparrott: Not a bad idea, I'll add operational locks to all helper threads.

PR #5245, opened 2018-04-26
Fix docstring for 'ExtractJob.destination_uri_file_counts'.

### LINKED ISSUE #5244: 'ExtractJob.destination_uri_file_counts' docstring copy pasta
In #5241, @yiga2 reported:

> BTW, there is also an issue with the documentation - see discrepancy @ 
> https://google-cloud-python.readthedocs.io/en/latest/bigquery/reference.html#google.cloud.bigquery.job.ExtractJob.destination_uri_file_counts

and elaborated later:

> For the doc, it seems `Returns: number of DML rows affected by the job, or None if job is not yet complete.` is from another property.

tseaver:
Closes #5244.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -1541,9 +1541,11 @@ class ExtractJob(_AsyncJob):
         See:
         https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.extract.destinationUriFileCounts
 
-        :rtype: int or None
-        :returns: number of DML rows affectd by the job, or None if job is not
-                  yet complete.
+        :rtype: list of int or None
+        :returns: number of files per destination URI or URI pattern specified
+                  in the extract configuration. These values will be in the
+                  same order as the URIs specified in the 'destinationUris'
+                  field, or None if job is not yet complete.
         """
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1541,9 +1541,11 @@ def destination_uri_file_counts(self):
         See:
         https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.extract.destinationUriFileCounts
 
-        :rtype: int or None
-        :returns: number of DML rows affectd by the job, or None if job is not
-                  yet complete.
+        :rtype: list of int or None
```
tswast: Could we change this to a [Napoleon/Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google) docstring while we're in here?

tseaver:
@tswast In converting to use the nappy docstyle (a3cb9ee), I noticed that the API returns a list of counts, one per destination URL.  e92bc8e changes the implementation to correspond.  It is not a backward-compatibility break, because anybody trying to use the `destination_uri_file_counts` property on a completed job was going to get a traceback.

tswast:
Thanks @tseaver !

PR #5249, opened 2018-04-26
Expose 'Blob.kms_key_name' as read-only property

tseaver:
Backed by `_properties`.


tseaver:
@frankyn PTAL

frankyn:
@tseaver review now.

tseaver:
Creating a destination key w/ `kms_key_name` allows rewrite to set / rotate the key for an existing blob.

PR #5255, opened 2018-04-27
BigQuery: add from_string factory methods to Dataset and Table

tswast:
To make it easier to construct Datasets and Tables when the
full-qualified ID is known ahead of time, this PR adds helper methods
called `from_string` to Dataset, DatasetReference, Table, and
TableReference. In each case, the expected format is a fully-qualified
ID (including the project ID) in standard SQL format.

```
dataset = client.create_dataset(
    bigquery.Dataset(
        client.dataset('mydataset', project='other-project')
    )
)
```

becomes

```
dataset = client.create_dataset(
    bigquery.Dataset.from_string('other-project.mydataset')
)
```

Also, while writing doctests/examples, I noticed `__repr__` was
inconsistent with the actual constructor on some of these classes. I
fixed that while I was here so those string representations don't cause
confusion.

/cc @maxim-lian

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -667,8 +667,7 @@ class Table(object):
                 If ``full_table_id`` is not a fully-qualified table ID in
                 standard SQL format.
         """
-        table_ref = TableReference.from_string(full_table_id)
-        return cls(table_ref)
+        return cls(TableReference.from_string(full_table_id))
 
     @classmethod
     def from_api_repr(cls, resource):
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -167,6 +167,39 @@ class TableReference(object):
             self._project, self._dataset_id, self._table_id)
 
     @classmethod
+    def from_string(cls, full_table_id):
+        """Construct a table reference from fully-qualified table ID.
+
+        Args:
+            full_table_id (str):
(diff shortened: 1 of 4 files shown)


maxim-lian:
Thanks a lot!

bigquery/google/cloud/bigquery/table.py
```diff
@@ -166,6 +166,39 @@ def path(self):
...
+        from google.cloud.bigquery.dataset import DatasetReference
+
+        parts = full_table_id.split('.')
+        if len(parts) != 3:
+            raise ValueError(
+                'full_table_id must be a fully-qualified table ID in '
+                'standard SQL format. e.g. "project.dataset.table", got '
+                '{}'.format(full_table_id))
+
+        return cls(DatasetReference(parts[0], parts[1]), parts[2])
```
maxim-lian: FYI for future you can pass `*parts` if you like

bigquery/google/cloud/bigquery/table.py
```diff
@@ -609,6 +645,31 @@ def external_data_configuration(self, value):
...
+            >>> Table.from_string('my-project.mydataset.mytable')
+            Table(TableRef...(D...('my-project', 'mydataset'), 'mytable'))
+
+        Raises:
+            ValueError:
+                If ``full_table_id`` is not a fully-qualified table ID in
+                standard SQL format.
+        """
+        table_ref = TableReference.from_string(full_table_id)
+        return cls(table_ref)
```
alixhami: Same as above. Can this be one line?

PR #5259, opened 2018-04-27
Storage: add support for KMS keys

tseaver:
Merge is blocked for public beta of the KMS support.  We will need to adjust the system tests at that time to switch from using keys in the `global` location to keys in the same location as the bucket.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -126,7 +126,6 @@ class Blob(_PropertyMixin):
     :param kms_key_name:
         Optional resource name of Cloud KMS key used to encrypt the blob's
         contents.
-        See https://cloud.google.com/storage/docs/encryption#kms.
     """
 
     _chunk_size = None  # Default value for each instance.
@@ -1718,8 +1717,6 @@ class Blob(_PropertyMixin):
     @property
     def kms_key_name(self):
         """Resource name of Cloud KMS key used to encrypt the blob's contents.
-
-        See https://cloud.google.com/storage/docs/encryption#kms.
 
         :rtype: str or ``NoneType``
         :returns:
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -121,6 +121,12 @@ class Blob(_PropertyMixin):
     :param encryption_key:
         Optional 32 byte encryption key for customer-supplied encryption.
         See https://cloud.google.com/storage/docs/encryption#customer-supplied.
+
+    :type kms_key_name: str
+    :param kms_key_name:
+        Optional resource name of Cloud KMS key used to encrypt the blob's
+        contents.
+        See https://cloud.google.com/storage/docs/encryption#kms.
     """
 
     _chunk_size = None  # Default value for each instance.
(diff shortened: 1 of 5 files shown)


storage/tests/system.py
```diff
@@ -990,3 +990,131 @@ def test_access_to_public_bucket(self):
...
+        source.upload_from_filename(file_data['path'])
+        self.case_blobs_to_delete.append(source)
+        source_data = source.download_as_string()
+
+        kms_key_name = self._kms_key_name()
+
+        # We can't verify it, but ideally we would check that the following
+        # URL was resolvable with our credentals
+        # KEY_URL = 'https://cloudkms.googleapis.com/v1/{}'.format(
+        #     kms_key_name)
```
frankyn: What do you mean by you can't verify *it*?

storage/tests/unit/test_blob.py
```diff
@@ -459,6 +588,25 @@ def test__get_download_url_on_the_fly_with_user_project(self):
...
+        blob_name = 'bzzz-fly.txt'
+        bucket = _Bucket(name='buhkit')
+        blob = self._make_one(
+            blob_name, bucket=bucket, kms_key_name=kms_resource)
+
+        self.assertIsNone(blob.media_link)
+        download_url = blob._get_download_url()
+        expected_url = (
+            'https://www.googleapis.com/download/storage/v1/b/'
+            'buhkit/o/bzzz-fly.txt?alt=media')  # kmsKeyName *not* expected
```
frankyn: You don't need this comment here.

storage/tests/system.py
```diff
@@ -990,3 +990,131 @@ def test_access_to_public_bucket(self):
...
+            "keyRings/{}/"
+            "cryptoKeys/{}"
+        ).format(
+            Config.CLIENT.project,
+            'global',  # will be 'self.bucket.location' after launch
+            self.KEYRING_NAME,
+            key_name,
+        )
+
+    def test_blob_w_explicit_kms_key_name(self):
```
frankyn: Add test to verify kms_key_name is available when listing blobs in a bucket.

storage/tests/system.py
tseaver: The system test can't exercise the KMS API endpoint to ensure that the key exists:  ideally we would do that, and then set up they keychain / key if it was missing / invalid.

storage/google/cloud/storage/blob.py
```diff
@@ -120,6 +120,12 @@ class Blob(_PropertyMixin):
     :param encryption_key:
         Optional 32 byte encryption key for customer-supplied encryption.
         See https://cloud.google.com/storage/docs/encryption#customer-supplied.
+
+    :type kms_key_name: str
+    :param kms_key_name:
+        Optional resource name of Cloud KMS key used to encrypt the blob's
+        contents.
+        See https://cloud.google.com/storage/docs/encryption#kms.
```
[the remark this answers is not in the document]
tseaver: 95807a7

storage/tests/unit/test_blob.py
tseaver: 95807a7

storage/tests/system.py
tseaver: ef6c181

storage/tests/system.py
frankyn: Understood, hopefully when KMS autogen is available then this can be resolved.

tseaver:
@frankyn FYI I just ran the KMS system tests without changing to regional keys, and they passed.  When I switch to regional keys, they fail.

frankyn:
Thanks @tseaver, what is the error you're seeing?

tseaver:
@frankyn  `The Cloud KMS key resides in a region that is not allowed.`

frankyn:
Synced with @tseaver and will look into reproducing error and syncing with GCS team.

frankyn:
I made commits and caused the CLA issue. I'm okay with my contributions being added to the repo.

tseaver:
@frankyn We are good to merge this branch today, right?

frankyn:
I saw the failures in tests were related to logging, and tests for storage weren't ran. I ran tests locally and they passed, but I want to make sure they're not broken for repo system tests.

tseaver:
@frankyn The logging system tests shouldn't even *run* against this branch.  I've run them locally using the same credentials we use on CircleCI, and they pass.  I will try rebuilding them.

tseaver:
I've rebased the branch against current `master`, hoping that the fix in #5264 will exclude the irrelevant system tests.

frankyn:
Same state as before, but CLA: yes was removed again..

PR #5265, opened 2018-04-30
Update setuptools, wheel, and twine before building packages

jonparrott:


test_utils/scripts/circleci/twine_upload.sh
```diff
@@ -27,9 +27,10 @@ SCRIPT="$(dirname "${BASH_SOURCE[0]}")/get_tagged_package.py"
 # Determine the package directory being deploying on this tag.
 PKG_DIR="$(python ${SCRIPT})"
 
+# Ensure that we have the latest versions of Twine, Wheel, and Setuptools.
+python3 -m pip install --upgrade twine wheel setuptools
```
tseaver: @jonparrott I think this change broke our auto-release bits:  I just got the following e-mail following @crwilcox's releases:

```bash
# If this is not a CircleCI tag, no-op.
if [[ -z &quot;$CIRCLE_TAG&quot; ]]; then
  echo &quot;This is not a release tag. Doing nothing.&quot;
  exit 0
fi

# H/T: http://stackoverflow.com/a/246128/1068170
SCRIPT=&quot;$(dirname &quot;${BASH_SOURCE[0]}&quot;)/get_tagged_package.py&quot;
dirname &quot;${BASH_SOURCE[0]}&quot;
# Determine the package directory being deploying on this tag.
PKG_DIR=&quot;$(python ${SCRIPT})&quot;
python ${SCRIPT}

# Ensure that we have the latest versions of Twine, Wheel, and Setuptools.
python3 -m pip install --upgrade twine wheel setuptools
/usr/local/bin/python3: No module named pip
Exited with code 1
```

jonparrott: how the hell? @crwilcox can you investigate? Did we build python3.6 without ensurepip?

crwilcox: I am looking at it. the issue is that 3.7 is not python3. So our docker file installed pip on py3.6  and then tried to use it from python3.7. I am working on it right now :)

PR #5267, opened 2018-04-30
BigQuery: Column based time partitioning

alixhami:
Adds client library support for column based time partitioning

Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4658

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -34,6 +34,7 @@ from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import EncryptionConfiguration
 from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.table import Table
+from google.cloud.bigquery.table import TimePartitioning
 from google.cloud.bigquery import _helpers
 from google.cloud.bigquery._helpers import DEFAULT_RETRY
 from google.cloud.bigquery._helpers import _int_or_none
@@ -1910,6 +1911,20 @@ class QueryJobConfig(_JobConfig):
         self._set_sub_prop(
             'tableDefinitions',  _to_api_repr_table_defs(values))
 
+    @property
+    def time_partitioning(self):
+        """google.cloud.bigquery.table.TimePartitioning: Specifies time-based
+        partitioning for the destination table.
+        """
+        prop = self._get_sub_prop('timePartitioning')
+        if prop is not None:
+            prop = TimePartitioning.from_api_repr(prop)
+        return prop
+
+    @time_partitioning.setter
+    def time_partitioning(self, value):
+        self._set_sub_prop('timePartitioning', value.to_api_repr())
+
     def to_api_repr(self):
         """Build an API representation of the query job config.
 
@@ -2080,6 +2095,13 @@ class QueryJob(_AsyncJob):
         :attr:`google.cloud.bigquery.job.QueryJobConfig.table_definitions`.
         """
         return self._configuration.table_definitions
+
+    @property
+    def time_partitioning(self):
+        """google.cloud.bigquery.table.TimePartitioning: Specifies time-based
+        partitioning for the destination table.
+        """
+        return self._configuration.time_partitioning
 
     def _build_resource(self):
         """Generate a resource for :meth:`begin`."""
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -968,6 +968,20 @@ class LoadJobConfig(_JobConfig):
             api_repr = value.to_api_repr()
         self._set_sub_prop('destinationEncryptionConfiguration', api_repr)
 
+    @property
+    def time_partitioning(self):
+        """google.cloud.bigquery.table.TimePartitioning: Specifies time-based
+        partitioning for the destination table.
+        """
+        prop = self._get_sub_prop('timePartitioning')
+        if prop is not None:
+            prop = TimePartitioning.from_api_repr(prop)
+        return prop
+
+    @time_partitioning.setter
+    def time_partitioning(self, value):
+        self._set_sub_prop('timePartitioning', value.to_api_repr())
+
 
 class LoadJob(_AsyncJob):
     """Asynchronous job for loading data into a table.
@@ -1114,6 +1128,13 @@ class LoadJob(_AsyncJob):
         :attr:`google.cloud.bigquery.job.LoadJobConfig.destination_encryption_configuration`.
         """
         return self._configuration.destination_encryption_configuration
+
+    @property
+    def time_partitioning(self):
+        """google.cloud.bigquery.table.TimePartitioning: Specifies time-based
+        partitioning for the destination table.
+        """
+        return self._configuration.time_partitioning
 
     @property
     def input_file_bytes(self):
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -980,7 +980,10 @@ class LoadJobConfig(_JobConfig):
 
     @time_partitioning.setter
     def time_partitioning(self, value):
-        self._set_sub_prop('timePartitioning', value.to_api_repr())
+        api_repr = value
+        if value is not None:
+            api_repr = value.to_api_repr()
+        self._set_sub_prop('timePartitioning', api_repr)
 
 
 class LoadJob(_AsyncJob):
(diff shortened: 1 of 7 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -967,6 +968,20 @@ def destination_encryption_configuration(self, value):
             api_repr = value.to_api_repr()
         self._set_sub_prop('destinationEncryptionConfiguration', api_repr)
 
+    @property
+    def time_partitioning(self):
+        """google.cloud.bigquery.table.TimePartitioning: Specifies time-based
```
tswast: I really wish these would render as links, especially with non-built-in types. I know we are waiting on https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5027 , but maybe until that is resolved we should make these into Sphinx references? @jonparrott, thoughts on that?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -967,6 +968,20 @@ def destination_encryption_configuration(self, value):
...
+        partitioning for the destination table.
+        """
+        prop = self._get_sub_prop('timePartitioning')
+        if prop is not None:
+            prop = TimePartitioning.from_api_repr(prop)
+        return prop
+
+    @time_partitioning.setter
+    def time_partitioning(self, value):
+        self._set_sub_prop('timePartitioning', value.to_api_repr())
```
tswast: Check for `None` before calling `to_api_repr()`?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -456,52 +458,77 @@ def table_type(self):
...
+                If the value is not :class:`TimePartitioning` or :data:`None`.
+        """
+        prop = self._properties.get('timePartitioning')
+        if prop is not None:
+            return TimePartitioning.from_api_repr(prop)
+
+    @time_partitioning.setter
+    def time_partitioning(self, value):
+        if value is not None and isinstance(value, TimePartitioning):
+            self._properties['timePartitioning'] = value.to_api_repr()
```
tswast: Don't we also need to set the property to `None` if it is `None`?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -456,52 +458,77 @@ def table_type(self):
...
+            ValueError:
+                If the value is not :class:`TimePartitioning` or :data:`None`.
+        """
+        prop = self._properties.get('timePartitioning')
+        if prop is not None:
+            return TimePartitioning.from_api_repr(prop)
+
+    @time_partitioning.setter
+    def time_partitioning(self, value):
+        if value is not None and isinstance(value, TimePartitioning):
```
tswast: Is checking `isinstance` necessary? We've been removing a lot of client-side validation like that.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -816,22 +843,41 @@ def table_type(self):
...
+            "TableListItem.time_partitioning.partitioning_type instead.",
+            PendingDeprecationWarning)
+        if self.time_partitioning is not None:
+            return self.time_partitioning.partitioning_type
 
     @property
     def partition_expiration(self):
         """Union[int, None]: Expiration time in milliseconds for a partition.
+
+        If this property is set and :attr:`partitioning_type` is not set,
```
tswast: Looks like this comment about `partitioning_type` migrated to the wrong property.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1099,3 +1145,101 @@ def to_dataframe(self):
...
+            Number of milliseconds for which to keep the storage for a
+            partition.
+    """
+    def __init__(self, partitioning_type, field=None, expiration_ms=None):
+        self._properties = {}
+        self.partitioning_type = partitioning_type
+        if field is not None:
+            self.field = field
+        if expiration_ms is not None:
+            self.expiration_ms = str(expiration_ms)
```
tswast: If you're using the property to set, you don't need to convert to `str` here.

Style-wise, I slightly prefer setting the whole `_properties` dictionary directly in the constructor, but it's fine to use the properties to set.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1099,3 +1145,101 @@ def to_dataframe(self):
...
+
+    def to_api_repr(self):
+        """Return a dictionary representing this object.
+
+        Returns:
+            dict:
+                A dictionary representing the TimePartitioning object in
+                serialized form.
+        """
+        return self._properties
```
tswast: Maybe add a comment why we aren't making a copy?

From chat: "Table already has some top-level properties which modify the time partitioning resource, so return a reference instead of a copy. This way any changes on the related Table properties also appear here."

bigquery/google/cloud/bigquery/table.py
alixhami: I think it is in the appropriate doc string. It's about something that will happen to `partitioning_type` based on how you set `partition_expiration`. The behavior existed previously, so I thought I would call out the unique defaulting that is happening. I will change "this property" to "partition_expiration" to make it more clear.

bigquery/google/cloud/bigquery/table.py
alixhami: Added some text similar to what is in `from_api_repr()` docstring

bigquery/google/cloud/bigquery/table.py
```diff
@@ -508,14 +511,14 @@ def partitioning_type(self, value):
...
-        :attr:`TimePartitioningType.DAY`.
+        If :attr:`partition_expiration` is set and :attr:`partitioning_type` is
+        not set, :attr:`partitioning_type` will default to
+        :attr:`~google.cloud.bigquery.table.TimePartitioningType.DAY`.
         """
         warnings.warn(
             "This method will be deprecated in future versions. Please use "
             "Table.time_partitioning.expiration_ms instead.",
-            PendingDeprecationWarning)
+            UserWarning)
```
tswast: What's the reason to use `UserWarning` instead of `PendingDeprecationWarning` or `DeprecationWarning`?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1099,3 +1148,109 @@ def to_dataframe(self):
...
+
+
+class TimePartitioningType(object):
+    """Specifies the type of time partitioning to perform."""
+
+    DAY = 'DAY'
+    """str: Generates one partition per day."""
+
+
+class TimePartitioning(object):
```
tswast: Could you also add `requirePartitionFilter` while you're here (currently beta feature)? Search for it in https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1099,3 +1148,109 @@ def to_dataframe(self):
...
+    def __init__(self, partitioning_type, field=None, expiration_ms=None):
+        self._properties = {}
+        self.partitioning_type = partitioning_type
+        if field is not None:
+            self.field = field
+        if expiration_ms is not None:
+            self.expiration_ms = expiration_ms
+
+    @property
+    def partitioning_type(self):
```
tswast: Nit: since "partitioning" is redundant with the class name, we could call this `type_` (trailing underscore to prevent conflicts with built-in) See: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4658#issue-284349635

bigquery/google/cloud/bigquery/table.py
alixhami: This was @jonparrott's preference

bigquery/tests/unit/test_client.py
```diff
@@ -530,7 +531,7 @@ def test_create_table_w_day_partition(self):
         }
         conn = client._connection = _make_connection(resource)
         table = Table(self.TABLE_REF)
-        table.partitioning_type = 'DAY'
+        table.time_partitioning = TimePartitioning()
```
tswast: 👍 Thanks for testing with a default value.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -278,7 +279,8 @@ class Table(object):
     _PROPERTY_TO_API_FIELD = {
         'friendly_name': 'friendlyName',
         'expires': 'expirationTime',
-        'partitioning_type': 'timePartitioning',
+        'time_partitioning': 'timePartitioning',
+        'type_': 'timePartitioning',
```
tswast: I think this instance of `partitioning_type` needs to stay `partitioning_type`.

PR #5273, opened 2018-04-30
BigQuery: switch list_partitions helper to a direct metatable read

shollyman:
Reading via a query isn't strictly necessary to get partition information.  Instead of accessing the __PARTITION_SUFFIX__ metatable via a legacy SQL query, we read the results from the metatable directly (via tabledata.list / list_rows).

Open questions:

* Should this be held off after we publish more canonical tables with the properties we want to test?  Right now the system test references a stable, public partitioned table created as part of partitioning work, but its very much a one-off table.  A more extension partition table example should expose interesting partition ids such as the null partition to make sure no one's expecting only numeric partition ids.

* What's the preferred outcome for unit testing?  The unit test is basically just testing list_rows (again), but the underlying schema of the metatable is wider than the default test representations.   I've pulled test_list_partitions out of the client unit test in favor of the system test, since the schema for the metatable is something that may be extended so the unit test is prone to diverge from reality.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1250,7 +1250,7 @@ class Client(ClientWithProject):
                 (Optional) How to retry the RPC.
 
         Returns:
-            [str]:
+            List[str]:
                 A list of the partition ids present in the partitioned table
         """
         meta_table = self.get_table(
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1242,16 +1242,16 @@ class Client(ClientWithProject):
     def list_partitions(self, table, retry=DEFAULT_RETRY):
         """List the partitions in a table.
 
-        :type table: One of:
-                     :class:`~google.cloud.bigquery.table.Table`
-                     :class:`~google.cloud.bigquery.table.TableReference`
-        :param table: the table to list, or a reference to it.
+        Arguments:
+            table (Union[google.cloud.bigquery.table.Table,
+                    google.cloud.bigquery.table.TableReference]):
+                The table or reference from which to get partition info
+            retry (google.api_core.retry.Retry):
+                (Optional) How to retry the RPC.
 
-        :type retry: :class:`google.api_core.retry.Retry`
-        :param retry: (Optional) How to retry the RPC.
-
-        :rtype: list
-        :returns: a list of time partitions
+        Returns:
+            [str]:
+                A list of the partition ids present in the partitioned table
         """
         meta_table = self.get_table(
             TableReference(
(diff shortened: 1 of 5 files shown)


bigquery/tests/unit/test_client.py
```diff
@@ -3111,54 +3111,6 @@ def test_list_rows_errors(self):
         with self.assertRaises(TypeError):
             client.list_rows(1)
 
-    def test_list_partitions(self):
```
tswast: You're probably going to need some sort of unit test for 100% unit test coverage.

tswast: P.S. I agree that a system test is more useful, but they are quite slow to run. The purpose of the 100% coverage by unit tests is so that we can be mostly confident things won't break while developing by just running unit tests.

tswast:
You've got a couple more flake8 errors.

```
nox > flake8 google/cloud/bigquery
google/cloud/bigquery/client.py:1262:13: E128 continuation line under-indented for visual indent
google/cloud/bigquery/client.py:1355:1: E302 expected 2 blank lines, found 1
nox > Command flake8 google/cloud/bigquery failed with exit code 1
nox > Session lint failed. :(
```

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1239,6 +1239,31 @@ def insert_rows_json(self, table, json_rows, row_ids=None,
 
         return errors
 
+    def list_partitions(self, table, retry=DEFAULT_RETRY):
+        """List the partitions in a table.
+
+        :type table: One of:
```
alixhami: Please update to [Google style docstring](http://www.sphinx-doc.org/en/stable/ext/example_google.html)

bigquery/tests/unit/test_client.py
```diff
@@ -2876,6 +2876,49 @@ def test_insert_rows_json(self):
             path='/%s' % PATH,
             data=SENT)
 
+    def test_list_partitions(self):
+        from google.cloud.bigquery.table import Table
+
+        ROWS = 3
```
alixhami: These should not be constants (make lowercase). Many of the other tests do this, but it's something we're slowly fixing throughout the tests.

bigquery/tests/system.py
```diff
@@ -280,6 +280,13 @@ def test_get_table_w_public_dataset(self):
         self.assertEqual(
             schema_names, ['word', 'word_count', 'corpus', 'corpus_date'])
 
+    def test_list_partitions(self):
+        # TODO(shollyman): public canonical static partition tables samples
+        table_ref = DatasetReference('bigquery-partition-samples', 'samples').table('stackoverflow_comments')
+        all_rows = Config.CLIENT.list_partitions(table_ref)
+        self.assertIn("20150508", all_rows)
```
tswast: Nit: Use single-quotes for strings.

PR #5276, opened 2018-05-01
Add gzip support to BigQuery load_table_from_file().

ar-qun:
Gzip uses integers 1 and 2 instead of strings 'r' and 'w'.
This small change adds support for it, at least until it's changed
in python itself.

This issue is reported in this Stack Overflow question:
https://stackoverflow.com/questions/49712542

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1439,11 +1439,12 @@ def _check_mode(stream):
     :param stream: A bytes IO object open for reading.
 
     :raises: :exc:`ValueError` if the ``stream.mode`` is a valid attribute
-             and is not among ``rb``, ``r+b`` or ``rb+``.
+             and is not among ``rb``, ``r+b``, ``rb+``, or  in case of gzip
+             in ``1`` or ``2``.
     """
     mode = getattr(stream, 'mode', None)
 
-    if mode is not None and mode not in ('rb', 'r+b', 'rb+'):
+    if mode is not None and mode not in ('rb', 'r+b', 'rb+', 1, 2):
         raise ValueError(
             "Cannot upload files opened in text mode:  use "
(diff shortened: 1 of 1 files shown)


ar-qun:
I signed it!

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1439,11 +1439,12 @@ def _check_mode(stream):
     :param stream: A bytes IO object open for reading.
 
     :raises: :exc:`ValueError` if the ``stream.mode`` is a valid attribute
-             and is not among ``rb``, ``r+b`` or ``rb+``.
+             and is not among ``rb``, ``r+b``, ``rb+``, or, in case of gzip,
+             not among ``1`` or ``2``.
```
tseaver: Can you please use the named constant, `gzip.READ`, rather than `1`?  Also, I don't believe `2` (aka `gzip.WRITE`) should be an allowable value:  we explicitly want a readable stream.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1439,11 +1439,12 @@ def _check_mode(stream):
...
     """
     mode = getattr(stream, 'mode', None)
```
tseaver: Rather than mixing in the oddball, let's add an `if isinstance(stream, gzip.GzipFile)` test here, e.g.:

```python
import gzip
...

        if isinstance(stream, gzip.GzipFile):
            readable = mode == gzip.READ
        else:
            readable = mode in ('rb', 'r+b', 'rb+')
        if not readable:
            raise ValueError(...)

tseaver:
@ar-qun I forgot to say "Thanks!" for the patch!

tseaver:
@ar-qun I have opened #5711 to add this feature with tests.  Thanks again for your effort!

tseaver:
Superseded by #5711.

PR #5277, opened 2018-05-01
Update docstring to reflect correct units

wyardley:
I am not sure if 1 MB was an old default or the units, but I think it's a typo either way? The current default seems to be `None` (which shows up in the generated docs), and I believe the unit is bytes, correct?

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -112,8 +112,9 @@ class Blob(_PropertyMixin):
     :param bucket: The bucket to which this blob belongs.
 
     :type chunk_size: int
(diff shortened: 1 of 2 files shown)


crwilcox:
Hi @wyardley 

Good catch and thanks for this PR. The old default was 1MB at a point and we altered the defaults a little while ago to be more in line with some of the other GCP tooling. When None on creation, it will be initialized to the default chunk size, at least for resumable uploaded (where we chunk).

I believe there are other places where this string may be wrong, specifically in `bucket.py` for the blob constructor. Could you also fix it there?

Thanks!

wyardley:
@crwilcox Updated there as well, and fixed the flake8 warning about line length, which should resolve the CI failure.

tseaver:
Thank you!

PR #5278, opened 2018-05-01
Generate websecurity v1alpha

crwilcox:


websecurityscanner/setup.py
```diff
@@ -0,0 +1,60 @@
...
+https://packaging.python.org/en/latest/distributing.html
+https://github.com/pypa/sampleproject
+"""
+
+from setuptools import setup, find_packages
+import io
+import sys
+
+install_requires = [
+    'google-api-core >= 0.1.0, < 0.2.0dev',
```
jonparrott: this need to just be `google-api-core[grpc]>=1.0.0,<2.0.0`, no need for the `googleapis-common-protos`.

websecurityscanner/setup.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2018 Google LLC
```
jonparrott: This needs to use our new setup.py template.

crwilcox: I will look at language and port that in.

PR #5280, opened 2018-05-01
Remove unused var (flake8 warning)

wyardley:
This showed up in a flake8 warning when I was looking at #5277 
Let me know if this is right vs. using `_raise_from_invalid_response()`.


tseaver:
Thank you!

PR #5281, opened 2018-05-01
BigQuery: add UnknownJob type for redacted jobs.

tswast:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5220

/cc @shollyman


tseaver:
I can't see any benefit to the caller in getting back an instance of  `UnknownJob`:  I would certainly find it disconcerting and useless.  ISTM that the caller's experience would be better if the backend should skip reporting jobs for which the caller does not access.  As an alternative, `Client.list_jobs()` could just discard entries without any configuration.

That said, I don't see an issue with the code here, so feel free to merge if the consensus is agin' me.

Issue #5282, opened 2018-05-01
Spanner: generator raises StopIteration under Python 3.7 (due to PEP 479)


myelin:
Running the following code (modified slightly from actual example, for clarity) under Python 3.7 results in RuntimeError('generator raised StopIteration').

~~~~~~
from google.cloud import spanner
import traceback

spanner_client = spanner.Client()
instance = spanner_client.instance('my-test-instance')
database = instance.database('spanner')

with database.snapshot() as snapshot:
  results = snapshot.execute_sql(
      "SELECT 'Success from Spanner!' as message")
  for row in results:
    print(row)

database.close()
~~~~~~

Here's the traceback:

~~~~~~
RuntimeError('generator raised StopIteration')
Traceback (most recent call last):
  File "/env/lib/python3.7/site-packages/google/cloud/spanner_v1/streamed.py", line 140, in __iter__
    self._consume_next()  # raises StopIteration
  File "/env/lib/python3.7/site-packages/google/cloud/spanner_v1/streamed.py", line 114, in _consume_next
    response = six.next(self._response_iterator)
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "xxx", line xx, in page_index
    for row in results:
RuntimeError: generator raised StopIteration
~~~~~~

This is the result of [PEP 479](https://www.python.org/dev/peps/pep-0479/), which changes the behaviour of StopIteration exceptions raised in generators: pre-3.7, these cause the generator to exit, but in 3.7 they result in a RuntimeError, so must be caught inside the generator.

tseaver:
@myelin Thanks for the report and for digging up the root cause (and trying it out on Python 3.7)!

crwilcox:
@myelin , Thanks again for reporting this. The fix is in master and the next release of spanner will have this fix.

PR #5288, opened 2018-05-02
spanner fixes to enable 3.7 #5282

crwilcox:


spanner/google/cloud/spanner_v1/streamed.py
```diff
@@ -137,7 +137,10 @@ def __iter__(self):
         iter_rows, self._rows[:] = self._rows[:], ()
         while True:
             if not iter_rows:
-                self._consume_next()  # raises StopIteration
+                try:
+                    self._consume_next()  # raises StopIteration
```
tseaver: Probably don't need the comment, given the `try:...except StopIteration:`  :).

crwilcox: Fair point.

PR #5295, opened 2018-05-03
Add Test runs for Python 3.7 and remove 3.4

crwilcox:
There are *very* few people using 3.4 and 3.7 is coming up. A few of the packages cannot use 3.7 yet

PubSub: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5292
Monitoring: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5293
BigQuery: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5294

### DIFF
--- a/api_core/nox.py
+++ b/api_core/nox.py
@@ -52,7 +52,7 @@ def default(session):
 
 
 @nox.session
-@nox.parametrize('py', ['2.7', '3.5', '3.6'])
+@nox.parametrize('py', ['2.7', '3.5', '3.6', '3.7'])
 def unit(session, py):
     """Run the unit test suite."""
 
--- a/api_core/nox.py
+++ b/api_core/nox.py
@@ -52,7 +52,7 @@ def default(session):
 
 
 @nox.session
-@nox.parametrize('py', ['2.7', '3.4', '3.5', '3.6'])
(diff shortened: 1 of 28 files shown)


tseaver:
Are you *sure* about the 3.4 bit?  It is still the default `python3` for LTS Linux versions, AFAIK.

crwilcox:
@tartavull @jonparrott  looked up the stats and said ~4%. We can leave it if you have strong feelings about it. It seemed a convenient time to not elongate the test runs is all.

jonparrott:
Let me double check that... I am bad at math.

tswast:
Query looks right, though I'd filter out the "bandersnatch" installs with `details.installer.name = 'pip'` if I were you. In my experiments those are a significant percentage of downloads and don't actually count as installs since they mirror all of PyPI.

P.S. I could give you some help updating to standard SQL syntax if you'd like.

jonparrott:
Thanks @tswast. Limiting to just pip installs made the Py 3.4 percentage even lower at 1.43%. ;)

tswast:
As pointed out by @tseaver in #5401, this also requires an update to all `setup.py` files.

PR #5301, opened 2018-05-07
BigTable: add "end_inclusive" flag to yield_rows

tomp-raycatch:
Modify the signature of "yield_rows" to match "read_rows".

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -282,7 +282,7 @@ class Table(object):
                                request_pb)
 
     def yield_rows(self, start_key=None, end_key=None, limit=None,
-                   filter_=None):
+                   filter_=None, end_inclusive=False):
         """Read rows from this table.
 
         :type start_key: bytes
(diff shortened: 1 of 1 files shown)


tomp-raycatch:
I signed it!

tseaver:
@tomp-raycatch Thanks for the patch!  In order to merge, the PR also needs to add / update unit tests in order to exercise passing the `end_inclusive` parameter to `yield_rows`.

tomp-raycatch:
@tseaver 

There is no existing test to update for `yield_rows` (at least not that is mocking out the call to `_create_row_request` in order to assert it's arguments)

I've spent a few hours trying to add one, based on `test_read_rows`. It's not simple, and I think it's outside the scope of this PR.

BTW coverage is still at 100%

sduskis:
Our team is going to expand the capabilities of yield rows as it relates to row key/range selection.  I would like to hold off on merging this PR so that we can develop an API that implements a wider variety of cases.

tomp-raycatch:
Okay.
So I'm closing this pull request.

For reference in case somebody is stuck at my situation - what I've ended up doing is this:
```python
    partial_rows_data = table.read_rows(
            start_key=start_key,
            end_key=end_key,
            limit=limit,
            filter_=filter_,
            end_inclusive = end_inclusive
    )
    for row in partial_rows_data._generator.read_rows():
        yield ( row.row_key, row )
```

`partial_rows_data._generator` is a `YieldRowsData` so besides the access to the 'private' attribute, it's pretty much the same as `yield_rows()`.

PR #5283, opened 2018-05-02
BigTable: Add split keys on create table

aneepct:
Add ability to create initial splits during creation of a table. This feature branch is a fork from the GAPIC integration branch (PR #5178).

PR #5302, opened 2018-05-07
BigTable: Modify system test for new GAPIC code

aneepct:
Modify system test for new GAPIC code


aneepct:
I am getting error on create_cluster on system tests.

`
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, Error in field 'clusters' : Need to have at least one cluster map element in CreateInstanceRequest.)>
`

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -161,9 +162,18 @@ def create(self):
...
         parent = self._client.project_path
+        cluster_path = self._client._instance_admin_client.cluster_path(
+            self._client.project, self.instance_id, 'shared-perf-cluster'
+        )
+        location_path = self._client._instance_admin_client.location_path(
+            self._client.project, 'us-central1-f'
+        )
+        clusters = []
+        clusters.append(instance_pb2.Cluster(name=cluster_path,
+                                             location=location_path))
```
tseaver: It cannot possibly be correct to jam this hardwired cluster configuration into `Instance.create`:  the method should take additional parameters (e.g., `cluster_id` and `location_id`) and use them to construct the two paths, if they are passed.

The *system tests* should then pass in the currently-hardcoded values (maybe with an environment check?) when creating an instance.

bigtable/tests/system.py
```diff
@@ -36,6 +36,7 @@
 from test_utils.system import unique_resource_id
 
 
+PROJECT_ID = 'grass-clump-479'
```
tseaver: We derive the project ID from the environment / credentials -- it shouldn't be hardcoded.

bigtable/tests/system.py
```diff
@@ -79,19 +80,19 @@ def setUpModule():
         credentials = EmulatorCreds()
         Config.CLIENT = Client(admin=True, credentials=credentials)
     else:
-        Config.CLIENT = Client(admin=True)
+        Config.CLIENT = Client(project=PROJECT_ID, admin=True)
```
tseaver: Shouldn't be needed:  by default, the client looks it up.

bigtable/tests/system.py
```diff
@@ -79,19 +80,19 @@ def setUpModule():
...
-        instances, failed_locations = retry(Config.CLIENT.list_instances)()
+        instances_response = retry(Config.CLIENT.list_instances)()
 
-        if len(failed_locations) != 0:
-            raise ValueError('List instances failed in module set up.')
+        # if len(instances_response.instances) != 0:
+        #     raise ValueError('List instances failed in module set up.')
 
-        EXISTING_INSTANCES[:] = instances
+        EXISTING_INSTANCES[:] = instances_response.instances
```
tseaver: These checks should not be removed:  having failed locations returned from `Client.list_instances` indicates that the setup for the tests is broken.

tseaver:
@aneepct Also, can you please rebase against the current `master` to dump all the redundant commits?

bigtable/google/cloud/bigtable/instance.py
aneepct: Fixed

bigtable/tests/system.py
aneepct: Fixed

bigtable/tests/system.py
aneepct: Removed

bigtable/tests/system.py
aneepct: Fixed

zakons:
@jonparrott @tseaver Hoping this closes the loop on a rather large series of merges on master to support GAPIC.  Together with #5275, #5315 and #5283 - which add support for app profile, truncate table and initial splits on top of the GAPIC interface - this should complete the series of changes for the python beta.  All have been updated to resolve the review issues found.  Please let us know if any issues remain so we can turn these around quickly.  Thanks again for the excellent reviews.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -140,8 +140,7 @@ def _table_data_client(self):
         :rtype: :class:`.bigtable_v2.BigtableClient`
         :returns: A BigtableClient object.
         """
-        return bigtable_v2.BigtableClient(channel=self._channel,
-                                          credentials=self._credentials)
+        return bigtable_v2.BigtableClient(channel=self._channel)
```
tseaver: Why are we not forwarding through the credentials here?  See for instance the [Spanner client](https://github.com/zakons/google-cloud-python/blob/35d3199f654da56d0a3d7de1196b7f70eed50460/spanner/google/cloud/spanner_v1/client.py#L148-L155)

bigtable/google/cloud/bigtable/client.py
```diff
@@ -156,7 +155,7 @@ def _table_admin_client(self):
         if not self._admin:
             raise ValueError('Client is not an admin client.')
         return bigtable_admin_v2.BigtableTableAdminClient(
-            channel=self._channel, credentials=self._credentials)
+            channel=self._channel)
```
tseaver: Likewise here?

bigtable/google/cloud/bigtable/client.py
```diff
@@ -171,7 +170,7 @@ def _instance_admin_client(self):
         if not self._admin:
             raise ValueError('Client is not an admin client.')
         return bigtable_admin_v2.BigtableInstanceAdminClient(
-            channel=self._channel, credentials=self._credentials)
+            channel=self._channel)
```
tseaver: And here?

bigtable/tests/system.py
```diff
@@ -378,7 +380,7 @@ def test_read_large_cell_limit(self):
         row = self._table.row(ROW_KEY)
         self.rows_to_delete.append(row)
 
-        number_of_bytes = 10 * 1024 * 1024
+        number_of_bytes = 10 * 1024
         data = b'1' * number_of_bytes  # 10MB of 1's.
```
tseaver: If we really mean to drop the data size from 10MB to 1 MB, then we need also to update the comment on the next line accordingly.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -160,10 +171,22 @@ def create(self):
...
                     operation.
         """
+        clusters = {}
+        cluster_id = '{}-cluster'.format(self.instance_id)
+        cluster_name = self._client._instance_admin_client.cluster_path(
+            self._client.project, self.instance_id, cluster_id)
+        location = self._client._instance_admin_client.location_path(
+            self._client.project, self._cluster_location_id)
+        cluster = instance_pb2.Cluster(name=cluster_name, location=location,
+                                       serve_nodes=DEFAULT_SERVE_NODES)
```
tseaver: Do we need to expose a way for the caller to set `serve_nodes` explicitly?  What about `default_storage_type`?

bigtable/tests/unit/test_client.py
```diff
@@ -93,13 +93,16 @@ def test_credentials_getter(self):
...
+                      credentials.with_scopes.return_value)
 
     def test_project_name_property(self):
         credentials = _make_credentials()
+        channel = _make_channel()
         project = 'PROJECT'
         client = self._make_one(
-            project=project, credentials=credentials, admin=True)
+            project=project, channel=channel, credentials=credentials,
+            admin=True)
```
tseaver: If we are going to pass `channel` here, shouldn't there be an assertion that it was saved correctly?

bigtable/tests/unit/test_client.py
```diff
@@ -141,8 +144,8 @@ def test_instance_factory_w_explicit_serve_nodes(self):
         self.assertIs(instance._client, client)
 
     def test_admin_client_w_value_error(self):
-        channel = _make_channel()
-        client = self._make_one(project=self.PROJECT, channel=channel)
+        credentials = _make_credentials()
+        client = self._make_one(project=self.PROJECT, credentials=credentials)
```
tseaver: Why this change?

bigtable/tests/unit/test_cluster.py
```diff
@@ -18,6 +18,17 @@
...
 
+def _make_credentials():
+    import google.auth.credentials
+
+    class _CredentialsWithScopes(
+            google.auth.credentials.Credentials,
+            google.auth.credentials.Scoped):
+        pass
+
+    return mock.Mock(spec=_CredentialsWithScopes)
```
tseaver: Rather than copying `_make_credentials`, we should be importing it from a shared location (e.g., `from ._utils import _make_credentials`

bigtable/tests/unit/test_client.py
aneepct: Done.

aneepct:
@tseaver Please review the recent changes

aneepct:
@tseaver Thank you for all your help.

crwilcox:
@aneepct the tests are failing after this merge. `nox -s "system(py='3.6')"` run from the bigtable directory will show the failures.

Here is a circle build failing if you want to see what the errors are. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/6646

aneepct:
@crwilcox The system test already fixed on different PR #5315

PR #5304, opened 2018-05-07
Grpcio pre system tests

crwilcox:


crwilcox:
Logging and BigTable will likely fail. This is due to https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5303 (Logging) and https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5302 (BigTable)

bigquery/nox.py
```diff
@@ -93,6 +93,9 @@ def system(session, py):
     # Set the virtualenv dirname.
     session.virtualenv_dirname = 'sys-' + py
 
+    # Use pre-release gRPC for system tests.
+    session.install('--pre', 'grpcio')
```
tswast: BigQuery doesn't have any dependencies on gRPC yet AFAIK. We use the REST API directly with Requests.

bigquery_datatransfer/nox.py
```diff
@@ -86,6 +86,9 @@ def system(session, py):
     # Set the virtualenv dirname.
     session.virtualenv_dirname = 'sys-' + py
 
+    # Use pre-release gRPC for system tests.
+    session.install('--pre', 'grpcio')
```
tswast: 👍

tseaver:
@crwilcox How long do we expect to need the pre-release version of `grpcio`?  Also, it would be useful to link the issue(s) here that the new pre-release solves, to give us scope about whether / when to revert.

tswast:
@tseaver I assumed this was proactive on our part so that we could catch errors in gRPC that cause issues in the client libs before the release. I agree it would be good to get some clarification as to why we are adding this (ideally as comments in the code).

tseaver:
@tswast I'm actually opposed to doing our normal CI against pre-releases of dependencies:  we need *more* stability in our everyday runs, not less.  It would be fine if we had an additional (maybe nightly?) "smoke test" for such things.

tswast:
I'd be on board with splitting risky / flakey / slow & only moderate value system tests into a nightly build.

jonparrott:
That sounds lovely, unfortunately, our current CI is working against us here. Let's defer that until we move to Kokoro.

tseaver:
@crwilcox You merged this one without getting consensus about the tradeoffs of using prerelease dependencies.  I will ask again:  Is there a specifc issue addressed by a particular `grpcio` prerlease that you are trying to fix?  If not, then I think we should reconsider making our everyday CI more fragile.

tseaver:
Note that #5333 (addressing https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/6613) is the kind of flakiness I'm talking about:  we are now in a mode where we *can't run* against both the released and the unreleased `grpcio` libraries, which is not a Good Thing(TM).

PR #5307, opened 2018-05-07
BigQuery:  Add additional statistics to query plan stages.

shollyman:
Part 1 of 2.  This plumbs through the additional statistics that are reported for each stage of query execution.   A followup change will introduce the timeline information.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2483,11 +2483,44 @@ class QueryPlanEntry(object):
     :type entry_id: int
     :param entry_id: ID of the entry
 
+    :type start_ms: int
+    :param start_ms: start time for this stage of the query plan
+
+    :type end_ms: int
+    :param end_ms: end time for this stage of the query plan
+
+    :type input_stages: List(int)
+    :param input_stages: list of stage IDs that are inputs for this stage
+
+    :type parallel_inputs: int
+    :param completed_parallel_inputs: number of parallel input
+            segments in the stage
+
+    :type completed_parallel_inputs: int
+    :param completed_parallel_inputs: number of parallel input
+            segments completed
+
+    :type wait_ms_avg: int
+    :param wait_ms_avg: milliseconds the average shard spent
+            waiting to be scheduled
+
+    :type wait_ms_max: int
+    :param wait_ms_max: milliseconds the slowest shard spent
+            waiting to be scheduled
+
     :type wait_ratio_avg: float
     :param wait_ratio_avg: average wait ratio
 
     :type wait_ratio_max: float
     :param wait_ratio_max: maximum wait ratio
+
+    :type read_ms_avg: int
+    :param read_ms_avg: milliseconds the average shard spent
+            reading input
+
+    :type read_ms_max: int
+    :param read_ms_max: milliseconds the slowest shard spent
+            reading input
 
     :type read_ratio_avg: float
     :param read_ratio_avg: average read ratio
@@ -2495,11 +2528,27 @@ class QueryPlanEntry(object):
     :type read_ratio_max: float
     :param read_ratio_max: maximum read ratio
 
+    :type compute_ms_avg: int
+    :param compute_ms_avg: milliseconds the average shard spent
+            on cpu-bound tasks
+
+    :type compute_ms_max: int
+    :param compute_ms_max: milliseconds the slowest shard spent
+            on cpu-bound tasks
+
     :type compute_ratio_avg: float
     :param compute_ratio_avg: average compute ratio
 
     :type compute_ratio_max: float
     :param compute_ratio_max: maximum compute ratio
+
+    :type write_ms_avg: int
+    :param write_ms_avg: milliseconds the average shard spent
+            writing output
+
+    :type write_ms_max: int
+    :param write_ms_max: milliseconds the slowest shard spent
+            writing output
 
     :type write_ratio_avg: float
     :param write_ratio_avg: average write ratio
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2483,24 +2483,73 @@ class QueryPlanEntry(object):
     :type entry_id: int
     :param entry_id: ID of the entry
 
+    :type start_ms: int
```
tswast: Howdy. Not to dump on you for docstrings again, but any chance we could update these to Google-style, too? Would be good to do now since we are doubling the number of properties.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2516,37 +2565,74 @@ class QueryPlanEntry(object):
...
+    :param shuffle_output_bytes_spilled: number of bytes written
+            to shuffle and spilled to disk
+
     :type steps: List(QueryPlanEntryStep)
     :param steps: steps in the entry
     """
     def __init__(self,
                  name,
                  entry_id,
+                 start_ms,
```
tswast: Rather than pass in a billion things to the constructor, how about we update this to take a resource object, which we save to `_properties` like our other classes?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2570,21 +2656,57 @@ def from_api_repr(cls, resource):
...
+                    'shuffleOutputBytesSpilled'),
             steps=[QueryPlanEntryStep.from_api_repr(step)
                    for step in resource.get('steps', ())],
         )
 
+    @property
+    def started(self):
+        """Datetime at which the stage started
+
+        :rtype: ``datetime.datetime``, or ``NoneType``
```
tswast: This should definitely get updated to Google-style.

Example:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/ef16bdde527fdc0aff908d8aa957482dbea82158/bigquery/google/cloud/bigquery/table.py#L658

bigquery/tests/unit/test_job.py
```diff
@@ -3376,3 +3498,29 @@ def test_from_api_repr_normal(self):
...
+
+    def test_started(self):
+        klass = self._get_target_class()
+
+        entry = klass.from_api_repr({})
+        self.assertEqual(
+            entry.started,
+            None)
+
+        entry.start_ms = self.START_MS
```
tswast: I think I'd prefer if you went from a datetime to ms here and then verify. I don't want to have to trust the `_datetime_from_microseconds` function to know the test is correct.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2475,237 +2475,268 @@ def __eq__(self, other):
...
-    :param start_ms: start time for this stage of the query plan
-
-    :type end_ms: int
-    :param end_ms: end time for this stage of the query plan
+    """
 
-    :type input_stages: List(int)
-    :param input_stages: list of stage IDs that are inputs for this stage
+    _PROPERTY_TO_API_FIELD = {
+        'name': 'name',
```
tswast: Nit. An entry shouldn't be necessary when the key is the same as the value.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2475,237 +2475,268 @@ def __eq__(self, other):
...
 
-    :type wait_ms_max: int
-    :param wait_ms_max: milliseconds the slowest shard spent
-            waiting to be scheduled
+        Returns:
+            google.cloud.bigquery.QueryPlanEntry:
+                Query plan entry parsed from ``resource``
+        """
+        entry = cls()
+        entry._properties = copy.deepcopy(resource)
```
tswast: Probably not worth doing the copy for a read-only class. Plus @alixhami and I are thinking about dropping this copy everywhere so that it's possible to do things like `job_config.destination_time_partitioning.field = 'field_name'` (we already dropped it from `TimePartitioning` for this example).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2475,237 +2475,268 @@ def __eq__(self, other):
...
+        """Union[str, None]: Unique ID for the stage within the plan."""
+        return self._properties.get('id')
 
-    :type read_ms_avg: int
-    :param read_ms_avg: milliseconds the average shard spent
-            reading input
+    @property
+    def start(self):
+        """Union[Datetime, None]: Datetime when the stage started."""
+        if self._properties.get('startMs') is not None:
```
tswast: Nit: I'd prefer for a check for `None` with an explicit `return None`.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2475,237 +2475,268 @@ def __eq__(self, other):
...
+        if self._properties.get('endMs') is not None:
+            return _datetime_from_microseconds(
+                    self._properties.get('endMs') * 1000.0)
 
-    :type read_ratio_avg: float
-    :param read_ratio_avg: average read ratio
+    @property
+    def input_stages(self):
+        """List(int): Entry IDs for stages that were inputs
+                for this stage.
```
tswast: We don't do a hanging indent for property (I think Sphinx won't render it right). Should be

```
"""List(int): Entry IDs for stages that were inputs
for this stage.
```

 See: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2475,115 +2475,268 @@ def __eq__(self, other):
...
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs
+    for the underlying API representation within query statistics.
 
-    :type entry_id: int
-    :param entry_id: ID of the entry
+    """
 
-    :type wait_ratio_avg: float
-    :param wait_ratio_avg: average wait ratio
+    _PROPERTY_TO_API_FIELD = {
```
tswast: Actually, I think we can drop this altogether, right? This sister dictionaries to `_PROPERTY_TO_API_FIELD` are used to convert from the update field mask to the JSON object, which isn't necessary for this read-only class.

shollyman: I'll go ahead and drop it.  I wrote it half to keep track of all the property methods I needed to author, as well as for possible testing.

bigquery/google/cloud/bigquery/job.py
shollyman: Will switch to direct assignment to _properties then.

PR #5312, opened 2018-05-08
Add timeline and top-level slot-millis to query statistics.

shollyman:
Part 2 of 2.  Expose the query timeline as a list of TimelineEntry objects.  Each sample provides progress information for a query query in terms of input units, as well as cumulative attribution of slot-millisecond resources.

Additionally, expose slot-milliseconds usage as a top-level field from the query statistics.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2201,6 +2201,14 @@ class QueryJob(_AsyncJob):
         return [QueryPlanEntry.from_api_repr(entry) for entry in plan_entries]
 
     @property
+    def timeline(self):
+        """List(TimelineEntry): Return the query execution timeline
+        from job statistics.
+        """
+        raw = self._job_statistics().get('timeline', ())
+        return [TimelineEntry.from_api_repr(entry) for entry in raw]
+
+    @property
     def total_bytes_processed(self):
         """Return total bytes processed from job statistics, if present.
 
@@ -2273,6 +2281,11 @@ class QueryJob(_AsyncJob):
         if result is not None:
             result = int(result)
         return result
+
+    @property
+    def slot_millis(self):
+        """Union[int, None]: Slot-milliseconds used by this query job."""
+        return self._job_statistics().get('totalSlotMs')
 
     @property
     def statement_type(self):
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2703,6 +2716,66 @@ def steps(self):
...
+        """
+        entry = cls()
+        entry._properties = resource
+        return entry
+
+    @property
+    def elapsed_ms(self):
+        """Union[int, None]: Milliseconds elapsed since start of query
+        execution."""
+        return self._properties.get('elapsedMs')
```
tswast: Needs `_int_or_none()` call since it's an `int64` in the REST API. Probably ditto for the other ints.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/3582164c39d53bccb6aa794814a913d70adf6df6/bigquery/google/cloud/bigquery/_helpers.py#L474

shollyman: Revisited guard checks through previous PR as well.

bigquery/tests/system.py
```diff
@@ -1087,6 +1087,90 @@ def test_query_w_timeout(self):
...
+            if len(entry.input_stages) > 0:
+                stages_with_inputs = stages_with_inputs + 1
+            if (entry.compute_ratio_max == 1.0 or
+                    entry.read_ratio_max == 1.0 or
+                    entry.wait_ratio_max == 1.0 or
+                    entry.write_ratio_max == 1.0):
+                stages_with_max_ratio = stages_with_max_ratio + 1
+        self.assertGreater(stages_with_inputs, 0)
+        self.assertGreater(len(plan), stages_with_inputs)
+        self.assertGreater(stages_with_max_ratio, 0)
```
tswast: I read your comment, but it's still not 100% clear to me why there will be at least one stage with `X_ratio_max` at exactly 1.0.

shollyman: They're reporting a ratio in relation to the worst accumulation type (wait/read/compute/write) in all stages.  The real measurement (in terms of milliseconds) will be the largest among all accumulation types in all stages, and becomes the denominator for all ratios.  It also means the slowest one has a 1.0 ratio, as numerator=denominator.

This is more validating the expectations of the server response, so I can pull it out as its not really testing the client code if you prefer.

tswast: My main concern is if it starts failing we won't know what to do from client side, but okay to leave it in.

PR #5319, opened 2018-05-10
BigQuery:  improve system test performance

shollyman:
Before, running the system tests verbosely with timing, top three slowest tests running on my local workstation:
90.16s call     bigquery/tests/system.py::TestBigQuery::test_query_many_columns
60.80s call     bigquery/tests/system.py::TestBigQuery::test_load_table_from_file_w_explicit_location
34.53s call     bigquery/tests/system.py::TestBigQuery::test_list_rows_many_columns

With these changes:
30.34s call     bigquery/tests/system.py::TestBigQuery::test_query_many_columns
21.44s call     bigquery/tests/system.py::TestBigQuery::test_load_table_from_file_w_explicit_location
(removed)      bigquery/tests/system.py::TestBigQuery::test_list_rows_many_columns

Changes to test_query_many_columns:

Previously, the wide column tests generated row data for a large (10k) column table locally, sent it via a POST-based load job, and then ran a rowread test and a query test against it.  It also depended on some nondeterministic behaviors such as row ordering remaining stable.

New behavior is to leverage a DDL CTAS query that generates the desired number of rows, with the field values being generated in a deterministic fashion based on value within the first field. Testing list_rows() on the raw table vs a SELECT * isn't a meaningful distinction as list_rows is tested elsewhere, so behavior differences will come out elsewhere.

Changes to  test_load_table_from_file_w_explicit_location:

We know (and test) that jobs work when data and job location are aligned, so save some time on the test by excising the additional verification using explicit location and instead focus on verifying the failures.  We do keep a query test in the correct location, but turn it into a dry run since we can validate location mismatches in dry_run stage rather than needing to invoke a full query.

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -385,58 +385,54 @@ class TestBigQuery(unittest.TestCase):
         page = six.next(iterator.pages)
         return list(page)
 
-    def _create_table_many_columns(self, rows):
-        # Load a table with many columns
-        dataset = self.temp_dataset(_make_dataset_id('list_rows'))
+    def _create_table_many_columns(self, rowcount):
+        # Generate a table of maximum width via CREATE TABLE AS SELECT.
+        # first column is named 'rowval', and has a value from 0..rowcount
+        # Subsequent column is named col_<N> and contains the value N*rowval,
+        # where N is between 1 and 9999 inclusive.
+        dsname = _make_dataset_id('wide_schema')
+        dataset = self.temp_dataset(dsname)
         table_id = 'many_columns'
         table_ref = dataset.table(table_id)
         self.to_delete.insert(0, table_ref)
-        schema = [
-            bigquery.SchemaField(
-                'column_{}_with_long_name'.format(col_i),
-                'INTEGER')
-            for col_i in range(len(rows[0]))]
-        body = ''
-        for row in rows:
-            body += ','.join([str(item) for item in row])
-            body += '\n'
-        config = bigquery.LoadJobConfig()
-        config.schema = schema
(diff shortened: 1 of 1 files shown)


bigquery/tests/system.py
```diff
@@ -385,58 +385,54 @@ def _fetch_single_page(table, selected_fields=None):
...
 
-    def _create_table_many_columns(self, rows):
-        # Load a table with many columns
-        dataset = self.temp_dataset(_make_dataset_id('list_rows'))
+    def _create_table_many_columns(self, rowcount):
+        # Generate a table of maximum width via CREATE TABLE AS SELECT.
+        # first column is named 'rowval', and has a value from 1..rowcount
+        # Subsequent column is named col_<N> and contains the value N*rowval,
+        # where N is between 1 and 9999 inclusive.
+        dsname = _make_dataset_id('wide_schema')
```
tseaver: Any reason not to mangle the dataset name using `_make_dataset_id`?  This test will fail is some future condition causes the bare `wide_dataset` dataset not to be deleted.

tseaver:
@shollyman Thanks muchly for working to reduce the BQ system test times!

bigquery/tests/system.py
tswast: Looks like he is using `_make_dataset_id` to me. Maybe this comment was on an old commit?

shollyman: Yes, _make_dataset_id suffixes the wide_schema prefix to avoid name collisions.

tseaver: Hmm, maybe my eyes are going.

PR #5321, opened 2018-05-11
Begin deprecation process for google-cloud umbrella package

crwilcox:


setup.py
```diff
@@ -76,9 +76,16 @@
...
     name='google-cloud',
-    version='0.32.1.dev1',
+    version='0.32.2',
     description='API Client library for Google Cloud',
     long_description=README,
     install_requires=REQUIREMENTS,
     **SETUP_BASE
 )
+
+print("WARNING: The google-cloud Python package is deprecated. On "
```
theacodes: Use `warnings.warn`?

README.rst
```diff
@@ -121,18 +121,6 @@ If you need support for other Google APIs, check out the
 
 .. _Google APIs Python Client library: https://github.com/google/google-api-python-client
 
-Quick Start
```
theacodes: Might make sense to replace this with instructions on installing specific service packages?

crwilcox: I debated the same thing. In the moment I thought the lower list would be good enough but if you think there is value in adding a list of the GA `pip install _PACKAGE_` I think we should.

theacodes: True, let's leave it out for now.

setup-README.rst
```diff
@@ -0,0 +1,187 @@
+Google Cloud Python Client
```
theacodes: Let's just make this *really* small and only mention the deprecation. Also go ahead and move setup.py/etc into legacy/google-cloud

crwilcox: lmk if it is small enough now. I left some other things. I can drop absolutely everything also but I figured the license text and build state should still be present. Happy to remove more.

legacy/google-cloud/setup.py
```diff
@@ -76,9 +76,17 @@
 
 setup(
     name='google-cloud',
-    version='0.32.1.dev1',
+    version='0.32.2',
```
tseaver: ISTM that jumping into the fully-deprecated state should be a non-patch version bump, i.e., `0.33.0`.

crwilcox: This version doesn't change any behavior so I thought a patch release made sense.

The following release, which will be in mid June, will result in this package being a no-op install that has none of the libraries anymore.

While this doesn't make any code change, I can also see your view that we should rev a minor version for this. 

With this explanation do you still believe it makes more sense to rev to 0.33 now and 0.34 in June? The plan was to do 0.32.2 now and rev to 0.33.0 in June.  I am going to proactively fix this to be 0.33.0.

PR #4874, opened 2018-02-12
Support for BigQuery's NUMERIC type, which is currently in alpha.

ElliottBrossard:

PR #5331, opened 2018-05-15
BigQuery: NUMERIC type support

tswast:
Support for BigQuery's NUMERIC type.

Note that this PR was already reviewed in #4874 for when this feature was still in alpha.

Originally authored by @ElliottBrossard


bigquery/tests/system.py
```diff
@@ -1005,6 +1006,10 @@ def _generate_standard_sql_types_examples(self):
                 'sql': 'SELECT TIME(TIMESTAMP "%s")' % (stamp,),
                 'expected': naive.time(),
             },
+            {
+                'sql': 'SELECT NUMERIC "123456789.123456789"',
+                'expected': decimal.Decimal('123456789.123456789'),
+            },
```
tseaver: Quibble:  it would be better to name the string literal constant, and format it into the query (as with the timestamp above).

tswast: Agreed, done.

tseaver:
@tswast CI fails due to #5304. Can you rebase to pick up the fixes from PR #5333?

PR #5333, opened 2018-05-16
Fix tests after grpcio update

crwilcox:
grpcio updated their package and seem to have added a close method. This PR updates the stub.


theacodes:
We will need to cut a new release of api_core.

PR #5339, opened 2018-05-16
BigQuery: add inherited members to API reference docs

tswast:
Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5328

### DIFF
--- a/docs/bigquery/reference.rst
+++ b/docs/bigquery/reference.rst
@@ -9,6 +9,7 @@ Client
 
 .. automodule:: google.cloud.bigquery.client
   :members:
+  :inherited-members:
   :show-inheritance:
 
 
@@ -17,6 +18,7 @@ Job
 
 .. automodule:: google.cloud.bigquery.job
   :members:
+  :inherited-members:
   :show-inheritance:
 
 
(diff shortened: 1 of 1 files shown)


tswast:
Note: this conflicts with https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5340, but I did this first since I think https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5340 could be controversial.

docs/bigquery/reference.rst
```diff
@@ -9,6 +9,7 @@ Client
 
 .. automodule:: google.cloud.bigquery.client
   :members:
+  :inherited-members:
   :show-inheritance:
```
tseaver: Don't we want to drop the `:show-inheritance:` bit, since the base class is private?

tswast: I think we'd just want to drop it from QueryJob, ExtractJob, etc. I'm not sure how to do that with the current setup.

tseaver: I don't think there is any relevant (i.e., non-public) inheritance to be shown in the `google.cloud.bigquery.job` module.

tswast: You're right. Dropped from all modules except `client` and `table`.

tswast:
Obsolete by https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5340

PR #5340, opened 2018-05-17
BigQuery: use autosummary to split up API reference docs

tswast:
@alixhami and I have heard from several folks that the [BigQuery API Reference docs](https://google-cloud-python.readthedocs.io/en/latest/bigquery/reference.html) are pretty impossible to navigate. This PR uses the [autosummary](http://www.sphinx-doc.org/en/master/ext/autosummary.html) module to split the docs into separate pages with an easier to browse summary table.

### DIFF
--- a/docs/bigquery/reference.rst
+++ b/docs/bigquery/reference.rst
@@ -1,68 +1,133 @@
 API Reference
 ~~~~~~~~~~~~~
 
-.. automodule:: google.cloud.bigquery
-  :no-members:
+.. currentmodule:: google.cloud.bigquery
+
+The main concepts with this API are:
+
+- :class:`~google.cloud.bigquery.client.Client` manages connections to the
+  BigQuery API. Use the client methods to run jobs (such as a
+  :class:`~google.cloud.bigquery.job.QueryJob` via
+  :meth:`~google.cloud.bigquery.client.Client.query`) and manage resources.
+
+- :class:`~google.cloud.bigquery.dataset.Dataset` represents a
+  collection of tables.
+
+- :class:`~google.cloud.bigquery.table.Table` represents a single "relation".
 
 Client
 ======
 
-.. automodule:: google.cloud.bigquery.client
-  :members:
-  :show-inheritance:
+.. autosummary::
+    :toctree: reference
 
+    client.Client
 
 Job
 ===
 
-.. automodule:: google.cloud.bigquery.job
-  :members:
-  :show-inheritance:
+Job Configuration
+-----------------
+
+.. autosummary::
+    :toctree: reference
+
+    job.QueryJobConfig
+    job.CopyJobConfig
+    job.LoadJobConfig
+    job.ExtractJobConfig
+
+Job Classes
+-----------
+
+.. autosummary::
+    :toctree: reference
+
+    job.QueryJob
+    job.CopyJob
+    job.LoadJob
+    job.ExtractJob
+    job.UnknownJob
+
+Job Resources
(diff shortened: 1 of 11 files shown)


docs/bigquery/reference.rst
```diff
@@ -1,68 +1,133 @@
...
+.. autosummary::
+    :toctree: generated
+
+    job.QueryJob
+    job.CopyJob
+    job.LoadJob
+    job.ExtractJob
+    job.UnknownJob
+
+Job Resources
```
tseaver: "Resources" doesn't seem right here:  it suggests things fetchable via an API request.  How about "Job-Related Types" or "Enumerations"?

third_party/sphinx/LICENSE
```diff
@@ -0,0 +1,280 @@
+License for Sphinx
```
tseaver: We ended up splitting out happybase because it was under a different license:  is there any consideration to that here?  I think that `sphinx-ext-autosummary` will use these templates by default from within its own resources:  are we customizing them here?

third_party/sphinx/sphinx/ext/autosummary/templates/autosummary/class.rst
```diff
@@ -0,0 +1,33 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
```
tseaver: If we are customizing, then `:inherited-members:` should go here?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -603,7 +603,7 @@ def get_job(
         Returns:
             Union[google.cloud.bigquery.job.LoadJob, \
                   google.cloud.bigquery.job.CopyJob, \
-                  google.cloud.bigquery.job.ExtractJob \
+                  google.cloud.bigquery.job.ExtractJob, \
```
tseaver: These fixes are uncontroversial, as well as unrelated to the main thrust of this PR.  Maybe they should move to their own PR?

tswast: Good idea. Mailed https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5342

third_party/sphinx/LICENSE
tswast: I believe it is okay in this case since we are putting it in a directory called `third_party`, which is different from the happybase case.

From some internal policies docs (go/thirdparty/non-google3):

> All non-Google code goes into the third_party/ directory of your project
>
> By default this should be a folder called third_party/ in the root directory of your project repository, but the exact name may vary by project.

third_party/sphinx/sphinx/ext/autosummary/templates/autosummary/class.rst
tswast: Yes, we are customizing slightly (the toctree so that sub-pages are generated). I'll add a comment.

Somehow that doesn't seem necessary for autosummary. I see the inherited members at https://storage.googleapis.com/tswast-docs-staging/google-cloud-python/pr-5340/bigquery/generated/google.cloud.bigquery.job.LoadJob.html#google.cloud.bigquery.job.LoadJob

docs/bigquery/reference.rst
tswast: I'll use "Job-Related Types" to keep it open to non-enumerations if we ever add some.

PR #5345, opened 2018-05-18
make dependency on logging less restrictive in error_reporting

tcwalther:
This PR makes the dependency on `google-cloud-python` less restrictive in `google-cloud-error-reporting`. Currently, it is not possible to install `google-cloud-error-reporting` alongside other up-to-date `google-cloud-python` packages.

Of course, there may be great reasons why it was so restrictive in the first place, in which case I'd love for the author to explain the cause.

If this change is fine, it would be great if this was merged swiftly and published to PyPI with an updated version number.

### DIFF
--- a/error_reporting/setup.py
+++ b/error_reporting/setup.py
@@ -29,7 +29,7 @@ version = '0.29.1'
 # 'Development Status :: 5 - Production/Stable'
 release_status = 'Development Status :: 3 - Alpha'
 dependencies = [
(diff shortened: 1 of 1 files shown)


theacodes:
Yeah go for that. I seem to remember us wanting to just absorb error reporting into logging.

tcwalther:
This has already been merged, but just for completeness: I hereby to confirm that I am happy for this altered commit to be merged.

Would be fantastic to see a new version being pushed to PyPI :-).

PR #5352, opened 2018-05-21
Tasks v2beta2

crwilcox:


tasks/setup.py
```diff
@@ -0,0 +1,74 @@
...
+readme_filename = os.path.join(package_root, 'README.rst')
+with io.open(readme_filename, encoding='utf-8') as readme_file:
+    readme = readme_file.read()
+
+packages = [
+    package for package in setuptools.find_packages()
+    if package.startswith('google')
+]
+
+namespaces = ['google']
```
theacodes: Uh-oh, we're missing the `google.cloud` namespace here. Can you make sure we didn't forget this on any other packages?

crwilcox: locally they all seem to use these lines. I searched for `namespaces =` and 30 results came back. All using the following:

```
# Determine which namespaces are needed.
namespaces = ['google']
if 'google.cloud' in packages:
    namespaces.append('google.cloud')
```

PR #5368, opened 2018-05-22
BigQuery: adds load partitioned table sample and reworks create/query samples

alixhami:
This PR adds a load sample for column based time partitioning, and updates the create table and query samples. The create table and query samples were leveraged from the alpha documentation, but they were not as simple as possible, which is preferred for samples. To resolve this, I used data with a simpler schema and ran a full query instead of a dry run.

These samples are meant to be similar to the other load samples, which use US state data. I uploaded a csv file to the shared bucket so that other language samples can use the same data. I also used print statements rather than asserts so that the other languages can more closely follow these samples. I noticed that when the Python sample is canonical and uses asserts, the other language samples tend to print inconsistent messages because they have no example to follow.

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -514,21 +514,15 @@ def test_create_partitioned_table(client, to_delete):
 
     table_ref = dataset_ref.table('my_partitioned_table')
     schema = [
-        bigquery.SchemaField('requested_url', 'STRING', mode='REQUIRED'),
-        bigquery.SchemaField('request_ts', 'TIMESTAMP', mode='REQUIRED'),
-        bigquery.SchemaField(
-            'response_info', 'RECORD', fields=[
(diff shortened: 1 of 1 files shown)


docs/bigquery/snippets.py
```diff
@@ -537,45 +531,90 @@ def test_create_partitioned_table(client, to_delete):
...
         location='US',
         job_config=job_config)  # API request
 
-    # A dry run query completes immediately.
-    assert query_job.state == 'DONE'
-
-    print("This query will process {} bytes.".format(
-        query_job.total_bytes_processed))
+    rows = list(query_job)
+    print("{} states were admitted to the US in the 1800s".format(len(rows)))
```
tswast: I love this.

shollyman: So now we have to choose between test correctness and building a time machine that alters the course of human history?  What a pickle.

Fine.  No flux capacitors for me.

PR #5369, opened 2018-05-23
BigTable: Add data app profile id 

aneepct:
Add ability to specify the app_profile_id to use for the following Table methods:

read_rows
mutate_rows
sample_row_keys

### DIFF
--- a/bigtable/tests/unit/test_table.py
+++ b/bigtable/tests/unit/test_table.py
@@ -342,7 +342,7 @@ class TestTable(unittest.TestCase):
     def test_list_column_families(self):
         self._list_column_families_helper()
 
-    def _read_row_helper(self, chunks, expected_result):
+    def _read_row_helper(self, chunks, expected_result, app_profile_id=None):
         from google.cloud._testing import _Monkey
         from google.cloud.bigtable import table as MUT
         from google.cloud.bigtable_v2.gapic import bigtable_client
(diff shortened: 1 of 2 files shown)


bigtable/tests/unit/test_table.py
```diff
@@ -362,7 +362,8 @@ def _read_row_helper(self, chunks, expected_result):
         request_pb = object()  # Returned by our mock.
         mock_created = []
 
-        def mock_create_row_request(table_name, row_key, filter_):
+        def mock_create_row_request(table_name, row_key, filter_,
+                                    app_profile_id=None):
             mock_created.append((table_name, row_key, filter_))
```
tseaver: This mock should append the `app_profile_id` to the tuple (after `filter_`), and then the code below (line 390) should assert that it is there, and matches the value passed by the caller.  At least one of the testcases (e.g., `test_read_row_complete` at line 399) should pass through a non-None value for the argument.

bigtable/tests/unit/test_table.py
```diff
@@ -502,6 +503,7 @@ def mock_create_row_request(table_name, **kwargs):
             'filter_': filter_obj,
             'limit': limit,
             'end_inclusive': False,
+            'app_profile_id': None
```
tseaver: We should probably pass through a non-None `app_profile_id` here.

aneepct:
@tseaver Fixed please review

bigtable/tests/unit/test_table.py
```diff
@@ -399,20 +402,21 @@ def test_read_row_complete(self):
...
 
+        app_profile_id = 'app-profile-id'
         chunk = _ReadRowsResponseCellChunkPB(
             row_key=self.ROW_KEY,
             family_name=self.FAMILY_NAME,
             qualifier=self.QUALIFIER,
             timestamp_micros=self.TIMESTAMP_MICROS,
             value=self.VALUE,
-            commit_row=True,
+            commit_row=True
```
tseaver: Not a blocker for merge, but trailing commas are, in general, a Good Thing(TM), especially in multi-line scenarios like this one:  they make future editing simpler (e.g., if another paramter is added, one doesn't need to remember to add the comma).

PR #5371, opened 2018-05-23
Updated SubscriberClient docs

andrewmfiorillo:
`subscribe_experimental` was promoted to `subscribe` but the docs for the `SubscriberClient` still suggested using `subscribe_experimental`


tseaver:
@andrewmfiorillo Thanks for the patch!  Looks like @theacodes noticed the error at the same time, and fixed it.

PR #5382, opened 2018-05-23
BigQuery: add ddl-related query stats.

shollyman:
python library already supported statement type, so its possible to determine if a query was a DDL statement based on the classification (e.g. CREATE_TABLE.
This adds ddl target table and operation performed.  The first identifies the resolved table being manipulated via a DDL statement, and the operation performed provides insight into whether that table was created, replaced, etc.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2267,6 +2267,29 @@ class QueryJob(_AsyncJob):
         return self._job_statistics().get('cacheHit')
 
     @property
+    def ddl_operation_performed(self):
+        """Union[str, None]: Return the DDL operation performed.
+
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2266,6 +2266,29 @@ def cache_hit(self):
         """
         return self._job_statistics().get('cacheHit')
 
+    @property
+    def ddl_operation_performed(self):
+        """Union[str, None]: Return the DDL operation performed.
```
tswast: TIL there's an `Optional[X]` https://docs.python.org/3/library/typing.html#typing.Optional which is "equivalent to `Union[X, None]`. Let's use that here.

PR #5387, opened 2018-05-24
BigQuery: Adds load_table_from_dataframe() and snippet

alixhami:
Loads a pandas DataFrame to a BigQuery table by converting to parquet and loading


docs/bigquery/snippets.py
```diff
@@ -2073,5 +2077,44 @@ def test_list_rows_as_dataframe(client):
...
+
+    job.result()  # Waits for table load to complete.
+
+    assert job.state == 'DONE'
+    table = client.get_table(table_ref)
+    assert table.num_rows == 4
+    # [END bigquery_load_table_dataframe]
+    column_names = [field.name for field in table.schema]
+    assert 'title' in column_names
+    assert 'release_year' in column_names
```
tswast: I think checking the columns is a good way to check.

I'm curious what the round-trip looks like if you do a full round trip with `client.list_rows(table).to_dataframe()`. Does Pandas treat the `__index_0__` column as a real index or does it add a new third column?

If it doesn't treat it as an index, I'd consider that a feature request for `to_dataframe()` more so than this method, though.

bigquery/tests/unit/test_client.py
```diff
@@ -3484,6 +3492,68 @@ def test_load_table_from_file_bad_mode(self):
...
+    def test_load_table_from_dataframe(self):
+        from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
+        from google.cloud.bigquery import job
+
+        client = self._make_client()
+        records = [
+            {'name': 'Monty', 'age': 100},
+            {'name': 'Python', 'age': 60},
+        ]
+        dataframe = pandas.DataFrame(records)
```
tswast: Let's use the `index` argument to DataFrame. Hopefully no extra columns beyond the data columns + index column are created. https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.html

tswast: Added in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5387/commits/fe6d81dd431a35824d1a7e6fc7bc7328d2ffb44a

bigquery/tests/unit/test_client.py
```diff
@@ -23,6 +23,14 @@
 import six
 from six.moves import http_client
 import pytest
+try:
+    import pandas
+except (ImportError, AttributeError):  # pragma: NO COVER
```
tswast: A bit strange that AttributeError is required. Is this from debugging the 2.7 issue?

alixhami: I saw it like this in `test_job.py` and copied it. It was added in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4585/commits/c72ba75da1cbd62f89fcf88ef0a36178b13de9d5. Should I add to anywhere we do pandas/pyarrow imports?

PR #4899, opened 2018-02-19
BigTable: improve read rows validation performance

zakons:
Improve performance of reading multiple rows by reducing the number of times values are retrieved from a chunk, as this is very expensive.  Instead, move values from a chunk into a cell data object and access data from here in order to do validation.  includes other performance improvements as well.  Measurement were done with cProfile to find hotspots for tuning.

PR #5390, opened 2018-05-25
BigTable: improve read rows validation performance

### LINKED ISSUE #4755: BigTable: row_data _validateChunk performance
Looking at the performance graph for reading a large number (1000) of rows, each row having 10 cells in one column family, the validation of chunks takes a large percentage of time (>25%), as measured by cProfile.  Areas for tuning are identified below.

**Validation of timestamp_micros ordering** 
While a row is being read, a validation is made to make sure that cells for one qualifier are in timestamp_micros order (most recent first):
```
    def _validate_chunk_row_in_progress(self, chunk):
        ...
        previous = self._previous_cell
        _raise_if(self._same_as_previous(chunk) and
                  chunk.timestamp_micros <= previous.timestamp_micros)

    def _same_as_previous(self, chunk):
        previous = self._previous_cell
        return (chunk.row_key == previous.row_key and
                chunk.family_name == previous.family_name and
                chunk.qualifier == previous.qualifier and
                chunk.labels == previous.labels)
```
This validation may be extraneous because:

1. A RowFilter may have been applied - CellsColumnLimitFilter(1) - which will limit the result to the cell with the most recent timestamp.
2. There may only be one cell value for the qualifier.
3. The BigTable server may ALWAYS be providing the cells in timestamp order.

Two possible tunings:

1. Remove the validation if the condition is unlikely to occur in the ReadRowResponse.
2. If the validation is kept, it could be performed only when qualifier cell lists with length > 1 are detected AFTER the chunks for the qualifier are consumed.

aneepct:
Improve performance of reading multiple rows by reducing the number of times values are retrieved from a chunk, as this is very expensive. Instead, move values from a chunk into a cell data object and access data from here in order to do validation. includes other performance improvements as well. Measurement were done with cProfile to find hotspots for tuning.

Closes #4755


aneepct:
This PR replacing old PR #5390

spanner/benchmark/bin/ycsb
```diff
@@ -17,4 +17,4 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-python ${DIR}/../ycsb.py "${@:1}"
+python ${DIR}/../ycsb.py "${@:1}"
```
tseaver: This change is spurious -- please remove it.

tseaver:
I believe the old version of the PR is #4899.

tseaver:
It also seems to me that this PR should be marked as closing #4755.  If so, @aneepct can you please add the line "Closes #4755" to the PR description?

aneepct:
@tseaver Fixed please review

spanner/benchmark/bin/ycsb
```diff
@@ -16,5 +16,3 @@
 # A YCSB-like executable that can be integrated into PerfKitBenchmarker.
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-
-python ${DIR}/../ycsb.py "${@:1}"
```
tseaver: This PR should *not* touch files inside `spanner/` at all:  the previous version (before 96257f8) had stripped the trailing newline from that line.  *Removing* that line altogether is much worse.

PR #5397, opened 2018-05-27
fix docstring: str -> dict

iddoberger:
Fix error in documentation. In case of single value the function "detect_language" returns a dictionary (not a string).


tseaver:
@iddoberger Thank you for the patch!

PR #5398, opened 2018-05-29
BigTable: Fixing client for system tests

aneepct:
Change the use of property instead of private variables for admin and data clients. System tests are also working fine with this changes.


tseaver:
I'm not sure I understand the rationale here:  the `table_admin_client` and `instance_admin_client` properties already exist, which means that external code (e.g., the system tests) can use them.  Why is this issue a blocker for anything else, if it only changes method implementations to use the properties?

PR #5401, opened 2018-05-29
BigQuery: Adds Python 3.7 and removes Python 3.4

alixhami:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5294


bigquery/nox.py
```diff
@@ -65,7 +65,7 @@ def default(session):
 
 
 @nox.session
-@nox.parametrize('py', ['2.7', '3.5', '3.6'])
+@nox.parametrize('py', ['2.7', '3.5', '3.6', '3.7'])
```
tswast: Add the same to `system`?

tswast: I guess we don't have 3.5 on system, either, nevermind.

tseaver: Right, we need to keep `system` as snappy as possible.  The chance that 3.x will break a system test which 3.y passes is pretty minimal.

crwilcox: System should likely be set to 2.7, 3.7 though.

tswast:
Right, wasn't that what https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5295 was about? Or did I misinterpret the intent?

bigquery/nox.py
tseaver: 3.7 is still beta.

PR #5403, opened 2018-05-29
 Attempt again to reproduce #4264. 

### LINKED ISSUE #4264: Datastore: receiving duplicate entities during query iteration
Hi.

I'm using google-cloud-datastore==1.3.0 from pypi.

When iterating over a query iterator the google-cloud-datastore sometimes returns the same entitiy twice.

I've had a look at the source and it seems that the last entity of a fetched page is the same as the first entity of the next page.

Iterating of pages and printing each enitity id makes the problem obvious.
```python
query_iter = qry.fetch()
pages = query_iter.pages
while True:
    page = next(pages)
    if not page:
        return

    for ent in page:
        print(ent.key.id)

    print("-- new page --")
```

The output looks like this:
```
6438744212111360
5774037759295488
5226622299930624
6390563805855744
5537056630505472
5007404484788224
5289011263307776
5805078561685504
-- new page --
5805078561685504
4785039263924224
4655690686660608
5253444773347328
5481464721309696
```

As you can see the entity 5805078561685504 is returned twice.

Dumping the protocol buffer shows that the 'cursor'-value of the last entity and is identical to the 'end_cursor'-value. Here are the last lines of the dump:```

```
      }
    }
  }
  cursor: "\n\346\002\n\260\002\n\003ka2\022\250\002\032\245\002..."
  version: 1507954620963000
}
end_cursor: "\n\346\002\n\260\002\n\003ka2\022\250\002\032\245\002..."
more_results: NOT_FINISHED
```

Maybe you can help.

tseaver:
This time, use integer timestamps for the entity keys.

The new system test passes.

Closes #4264.

### DIFF
--- a/datastore/tests/system/test_system.py
+++ b/datastore/tests/system/test_system.py
@@ -363,7 +363,7 @@ class TestDatastoreQuery(TestDatastore):
         sansa_dict = dict(sansa_entity)
         self.assertEqual(sansa_dict, {'name': 'Sansa', 'family': 'Stark'})
 
-    def test_query_paginate_simple(self):
+    def test_query_paginate_simple_uuid_keys(self):
 
         # See issue #4264
         page_query = self.CLIENT.query(kind='uuid_key')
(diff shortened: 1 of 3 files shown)


tseaver:
Rebased to pick up the BQ systest fix from #5400.

PR #5404, opened 2018-05-29
Release pubsub 0.36.0

crwilcox:

### DIFF
--- a/pubsub/CHANGELOG.md
+++ b/pubsub/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 [1]: https://pypi.org/project/google-cloud-pubsub/#history
 
+## 0.36.0
+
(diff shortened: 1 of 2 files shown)


tseaver:
Hmm, this might be a case where `0.35.1` would be a better release number:  it is a bugfix-only implementation change, right?

crwilcox:
this is only a bug fix. I defaulted to 'go bigger'. I can modify that though.

crwilcox:
Release is at https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/pubsub-0.36.0

tseaver:
Shouldn't that have been `0.35.1`?

crwilcox:
ignore that comment it is :)

PR #5406, opened 2018-05-29
Don't error out for empty response.

jerjou:
grpc will not set a field that's set to its default value. So, if the
api should be returning an empty string or an empty list, for example,
operation.py currently throws an error because the response isn't set.
In fact, though, the response is just set to the default value, which
doesn't get serialized over the wire.

### DIFF
--- a/api_core/google/api_core/operation.py
+++ b/api_core/google/api_core/operation.py
@@ -107,21 +107,21 @@ class Operation(polling.PollingFuture):
             if not self._operation.done or self._result_set:
                 return
 
-            if self._operation.HasField('response'):
-                response = protobuf_helpers.from_any_pb(
-                    self._result_type, self._operation.response)
-                self.set_result(response)
-            elif self._operation.HasField('error'):
+            if self._operation.HasField('error'):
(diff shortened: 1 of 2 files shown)


tseaver:
@jerjou Thanks for the patch!  We will need to add a unit test which exercises the "empty response" case.

jerjou:
Sure! Updated the existing unit test for the new behavior.
Wasn't sure how to run the tests, but hopefully this'll work?

jerjou:
Actually - I might have my analysis wrong. Let me investigate, and I'll ping this PR again once I'm ready again..

tseaver:
@jerjou Any update?

jerjou:
Sorry - there's always five things to do at once, and five stakeholders wondering why they're not done yet -_-. Preliminary investigation seems to indicate that the server change that I made didn't cause this problem (it occurs even on a server running a version just before the change), but I'd like to investigate more. Unfortunately, there're always other priorities -_-. Let me close this PR for now, and I'll reopen when I'm more confident.

jerjou:
Just to close the loop on this - turns out it was a server error.

PR #5412, opened 2018-05-30
Fix retrying of bidirectional RPCs and closing the streaming pull manager

### LINKED ISSUE #5306: Subscribe experimental doesn't appear to be retrying on retryable exceptions, also doesn't surface the error.
Full context at https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4841#issuecomment-386988007 downwards.

gist with logs: https://gist.github.com/psalaberria002/e91956d94e82d706d65da7111a40ca24

Possible culprits:

1. `send()` and `recv()` expectedly error on `next(self.call)`, but if that was a *retryable* error that somehow re-bubbled, it would not trigger any callbacks.
2. `bidi` uses `rpc._wrapped` because `api_core` doesn't appropriately expose `add_done_callback`. This means that bidi unwraps all of the retry logic.

This doesn't explain *why* retrying a perfectly normal looking error somehow still leads us to a broken stream, unless the newly opened stream also has an error. But it's unclear why that wouldn't happen when calling the initial RPC.

theacodes:
Closes #5306

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -33,6 +33,7 @@ import google.cloud.pubsub_v1.subscriber.message
 import google.cloud.pubsub_v1.subscriber.scheduler
 
 _LOGGER = logging.getLogger(__name__)
+_RPC_ERROR_THREAD_NAME = 'Thread-OnRpcTerminated'
 _RETRYABLE_STREAM_ERRORS = (
     exceptions.DeadlineExceeded,
     exceptions.ServiceUnavailable,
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -414,11 +415,28 @@ def _should_recover(self, exception):
         # If this is in the list of idempotent exceptions, then we want to
         # recover.
         if isinstance(exception, _RETRYABLE_STREAM_ERRORS):
+            logging.info('Observed recoverable stream error %s', exception)
```
sjagoe: Are these log messages intentionally using the root logger?

theacodes:
No, I just noticed this myself. Will fix asap.

PR #5413, opened 2018-05-30
Add heartbeating to the streaming pull manager

theacodes:


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
```diff
@@ -0,0 +1,67 @@
...
+
+from __future__ import absolute_import
+
+import logging
+import threading
+
+
+_LOGGER = logging.getLogger(__name__)
+_HEARTBEAT_WORKER_NAME = 'Thread-Heartbeater'
+_DEFAULT_PERIOD = 30
```
crwilcox: there is no sense of the unit of this from the interface. Not necessary to fix, but a doc comment on the constructor might be helpful.
Also, is there a reason to pick 30 seconds? It is pretty infrequent, but wondering if we can't make it even less frequent?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
```diff
@@ -0,0 +1,67 @@
...
+                name=_HEARTBEAT_WORKER_NAME,
+                target=self.heartbeat)
+            thread.daemon = True
+            thread.start()
+            _LOGGER.debug('Started helper thread %s', thread.name)
+            self._thread = thread
+
+    def stop(self):
+        with self._operational_lock:
+            self._stop_event.set()
```
crwilcox: does the set need to be in the lock?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
theacodes: Done.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
theacodes: not necessarily, but the rest of this should be.

PR #5415, opened 2018-05-30
BigQuery: Adds schema update options and snippets

alixhami:


docs/bigquery/snippets.py
```diff
@@ -1365,6 +1365,126 @@ def test_load_table_from_uri_truncate(client, to_delete):
...
+    filepath = os.path.join(
+        snippets_dir, '..', '..', 'bigquery', 'tests', 'data', 'people.csv')
+
+    # [START bigquery_add_column_load_append]
+    # from google.cloud import bigquery
+    # client = bigquery.Client()
+    # dataset_ref = client.dataset('my_dataset')
+    # filepath = 'path/to/your_file.csv'
+
+    table_ref = dataset_ref.table('my_table')
```
shollyman: Variations of this table setup is present in all the samples, but isn't core to the snippet.  We should either clearly comment that this is setup, or move the table instantiation outside the snippet range and change the example to more clearly highlight the schema adaptation going on.

One possibility: walk the old/new schema and highlight how many fields were relaxed, or how many new fields were added with the print statements or asserts?

docs/bigquery/snippets.py
```diff
@@ -1365,6 +1365,126 @@ def test_load_table_from_uri_truncate(client, to_delete):
...
+    table = client.create_table(bigquery.Table(table_ref, schema=old_schema))
+
+    # Configure the load job to append the data to a destination table,
+    # allowing field relaxation
+    job_config = bigquery.LoadJobConfig()
+    job_config.write_disposition = bigquery.WriteDisposition.WRITE_APPEND
+    job_config.schema_update_options = [
+        bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION,
+    ]
+    # The source file data to load has one less column than the existing table
```
shollyman: Make this comment clearer that the missing column is required without the relaxation?

docs/bigquery/snippets.py
```diff
@@ -1810,6 +1930,100 @@ def test_client_query_destination_table_cmek(client, to_delete):
...
+    # allowing field relaxation
+    job_config = bigquery.QueryJobConfig()
+    job_config.schema_update_options = [
+        bigquery.SchemaUpdateOption.ALLOW_FIELD_RELAXATION,
+    ]
+    job_config.destination = table_ref
+    job_config.write_disposition = bigquery.WriteDisposition.WRITE_APPEND
+
+    query_job = client.query(
+        # Write a query that will omit a required field
```
shollyman: Perhaps we demonstrate with all the columns projected but the value being NULL?  SELECT "Beyonce" as full_name, NULL as age?

docs/bigquery/snippets.py
```diff
@@ -1810,6 +1952,130 @@ def test_client_query_destination_table_cmek(client, to_delete):
...
+    dataset = client.create_dataset(dataset)
+    to_delete.append(dataset)
+
+    dest_table_ref = dataset_ref.table('my_table')
+    schema = [
+        bigquery.SchemaField('full_name', 'STRING', mode='REQUIRED'),
+        bigquery.SchemaField('age', 'INTEGER', mode='REQUIRED'),
+    ]
+    dest_table = client.create_table(
+        bigquery.Table(dest_table_ref, schema=schema))
```
tseaver: This variable is unused, which is what caused the [CI failure for `lint`](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/6764).

PR #5417, opened 2018-05-31
BigTable: Integrate mutations

aneepct:
Separate all mutations into a standalone package which may be used in a Beam environment, without the need to create DirectRow's prior to creating mutations.  Refactor DirectRow to use the new mutation classes.


bigtable/google/cloud/bigtable/row.py
```diff
@@ -86,7 +85,7 @@ class _SetDeleteRow(Row):
     * :meth:`delete_cell`
     * :meth:`delete_cells`
 
-    :type row_key: bytes
+    :type row_key: str
```
tseaver: What is the rationale here?  The [backend docs](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Row) explicitly state that row keys are bytes, not text.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -313,20 +311,20 @@ def set_cell(self, column_family_id, column, value, timestamp=None):
                                  Must be of the form
                                  ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
 
-        :type column: bytes
+        :type column: str
```
tseaver: Column qualifiers are likewise bytes, not text, in the [backend docs](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Column).

bigtable/google/cloud/bigtable/row.py
```diff
@@ -313,20 +311,20 @@ def set_cell(self, column_family_id, column, value, timestamp=None):
...
         :param column: The column within the column family where the cell
                        is located.
 
-        :type value: bytes or :class:`int`
+        :type value: str or :class:`int`
```
tseaver: Cell values are likewise bytes, not text, in the [backend docs](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Cell).

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
+from google.cloud._helpers import _microseconds_from_datetime
```
sduskis: Ee need the standard Copyright header

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+from google.cloud._helpers import _to_bytes
+from google.cloud.bigtable_v2.proto import (
+    data_pb2 as data_v2_pb2)
+from google.cloud.bigtable_v2.proto import (
+    bigtable_pb2 as table_v2_pb2)
+
+
+class RowMutations(object):
+    """Create Entry using list of mutations
```
sduskis: I think that the comments need to be fleshed out a bit more.  Let's discuss this offline.

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Arguments:
+            row_key (str): Key of the Row in string.
```
sduskis: Is str the standard everywhere?  What about byte arrays?

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    ALL_COLUMNS = object()
+    """Sentinel value used to indicate all columns in a column family."""
+
+    def __init__(self, row_key, app_profile_id=None):
```
sduskis: app_profile_id isn't needed here.  It's only needed on `table.py`.

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+                Must match ``[-_.a-zA-Z0-9]+``.
+            columns (list):
+                The columns within the column family that will have cells
+                deleted. If :attr:`ALL_COLUMNS` is used then the entire
+                column family will be deleted from the row.
+            time_range (TimestampRange):
+                (optional) The range of timestamps within which cells should be
+                deleted.
+        """
+        if columns is self.ALL_COLUMNS:
```
sduskis: I don't think that ALL_COLUMNS is a good idea.  Users should just call `delete_from_family` directly

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+                Must match ``[-_.a-zA-Z0-9]+``.
+        """
+        delete_from_family_mutation = data_v2_pb2.Mutation.DeleteFromFamily(
+            family_name=family_name
+        )
+        mutation_message = data_v2_pb2.Mutation(
+            delete_from_family=delete_from_family_mutation)
+        self.mutations.append(mutation_message)
+
+    def delete(self):
```
sduskis: Can you please rename this `delete_row` or something like that?

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+        """Create the mutation request message for DeleteFromRow and add it
+        to the list of mutations"""
+        delete_from_row_mutation = data_v2_pb2.Mutation.DeleteFromRow()
+        mutation_message = data_v2_pb2.Mutation(
+            delete_from_row=delete_from_row_mutation)
+        self.mutations.append(mutation_message)
+
+    def create_entry(self):
```
sduskis: Should this be private?

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,125 @@
...
+        """Create a MutateRowsRequest Entry from the list of mutations
+
+        Returns:
+            `Entry <google.bigtable.v2.MutateRowsRequest.Entry>`
+             An ``Entry`` for a MutateRowsRequest message.
+        """
+        entry = table_v2_pb2.MutateRowsRequest.Entry(row_key=self.row_key)
+        for mutation in self.mutations:
```
sduskis: We can skip this, if we create the `MutateRowsRequest.Entry` in the constructor, and set it to a member variable.  We could update t on each method, rather than doing extra work to copy from `self.mutations`.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -409,28 +407,13 @@ def commit(self):
         :raises: :class:`ValueError <exceptions.ValueError>` if the number of
                  mutations exceeds the :data:`MAX_MUTATIONS`.
         """
-        mutations_list = self._get_mutations(None)
-        num_mutations = len(mutations_list)
+        num_mutations = len(self.row_mutations.mutations)
         if num_mutations == 0:
-            return
+            return {}
```
sduskis: This should be an exception, not just a return.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -409,28 +407,13 @@ def commit(self):
...
-            predicate=_retry_commit_exception,
-            deadline=30)
-        retry_(commit)()
-
-        self.clear()
-
-    def clear(self):
-        """Removes all currently accumulated mutations on the current row."""
-        del self._pb_mutations[:]
+            raise ValueError('%d total append mutations exceed the maximum '
```
sduskis: This error should be thrown from `row_mutations` methods instead of doing the logic here.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -321,7 +321,21 @@ def yield_rows(self, start_key=None, end_key=None, limit=None,
         for row in generator.read_rows():
             yield row
 
-    def mutate_rows(self, rows, retry=DEFAULT_RETRY):
+    def save_mutations(self, row_mutations):
```
sduskis: shouldn't there be a retry parameter here?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -338,22 +352,16 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY):
         :type rows: list
         :param rows: List or other iterable of :class:`.DirectRow` instances.
```
sduskis: We should keep the retry parameter

bigtable/google/cloud/bigtable/table.py
```diff
@@ -338,22 +352,16 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY):
...
         :rtype: list
         :returns: A list of response statuses (`google.rpc.status_pb2.Status`)
                   corresponding to success or failure of each row mutation
                   sent. These will be in the same order as the `rows`.
         """
-        retryable_mutate_rows = _RetryableMutateRowsWorker(
-            self._instance._client, self.name, rows,
-            app_profile_id=self._app_profile_id)
-        return retryable_mutate_rows(retry=retry)
+        mutation_rows = []
```
sduskis: This may look nicer as a lambda.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -447,6 +455,19 @@ def _is_retryable(status):
         return (status is None or
                 status.code in _RetryableMutateRowsWorker.RETRY_CODES)
 
+    def _mutate_rows_entries(self, mutation_rows):
+        entries = []
+        mutations_count = 0
+        for mutation_row in mutation_rows:
+            mutations_count += 1
```
sduskis: this should be:

```
    entry = mutation_row.create_entry()
    mutations_count += len(entry.mutations) // I'm not sure what the right syntax is
    // perform size check here
    entries.append(entry)

File: bigtable/tests/unit/test_mutation.py [hunk 18]
```diff
@@ -0,0 +1,122 @@
+# Copyright 2015 Google LLC
```
sduskis: This should be 2018 instead of 2015

bigtable/google/cloud/bigtable/mutation.py
tseaver: Row keys are bytes, not text, in the API, and we surface that to the library clients.

bigtable/google/cloud/bigtable/table.py
aneepct: Done

bigtable/google/cloud/bigtable/table.py
aneepct: This is changed now. Please review.

tseaver:
CI failures:

### lint
```
nox > flake8 google tests
tests/unit/test_row.py:218:9: F841 local variable 'klass' is assigned to but never used
tests/unit/test_row.py:442:18: E128 continuation line under-indented for visual indent
tests/unit/test_mutation.py:18:1: F401 'google.cloud._helpers._to_bytes' imported but unused
```

### cover
```
tests/unit/test_mutation.py                 45      2      0      0    96%   101-103
```

bigtable/google/cloud/bigtable/row.py
```diff
@@ -245,7 +244,7 @@ def _retry_commit_exception(exc):
                             exceptions.DeadlineExceeded))
 
 
-class DirectRow(_SetDeleteRow):
+class DirectRow(mutation.RowMutations):
```
sduskis: What do you think about extending `_SetDeleteRow` from `mutation.RowMutations` instead of just doing it in `DirectRow`?

aneepct: @sduskis _SetDeleteRow is inherited by other classes in row.py:  ConditionalRow and AppendRow and RowMutations isn't integrated with these classes. Note that these classes use other GAPIC api's, not the GAPIC api used by RowMutations.

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,135 @@
...
+            # Use -1 for current Bigtable server time.
+            timestamp_micros = -1
+        else:
+            timestamp_micros = _microseconds_from_datetime(timestamp)
+            # Truncate to millisecond granularity.
+            timestamp_micros -= (timestamp_micros % 1000)
+
+        set_cell_mutation = data_v2_pb2.Mutation.SetCell(
+            family_name=family_name,
+            column_qualifier=_to_bytes(column_id),
```
tseaver: `column_id` is documented above as `bytes`, so the conversion shouldn't be needed.

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def delete_cells(self, family_name, columns, time_range=None):
+        """Create the mutation request message for DeleteFromColumn and
+            add it to the list of mutations in the MutateRowsRequest.Entry
+
+        Arguments:
+            family_name (str):
+                The name of the family into which new data should be written.
+                Must match ``[-_.a-zA-Z0-9]+``.
+            columns (list):
```
tseaver: Document as `list of bytes`.

bigtable/google/cloud/bigtable/mutation.py
```diff
@@ -0,0 +1,135 @@
...
+                deleted.
+            time_range (TimestampRange):
+                (optional) The range of timestamps within which cells should be
+                deleted.
+        """
+        for column_id in columns:
+            delete_from_column_mutation = (
+                data_v2_pb2.Mutation.DeleteFromColumn(
+                    family_name=family_name,
+                    column_qualifier=_to_bytes(column_id),
```
tseaver: Once documented as `list of bytes`, we don't need the conversion.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -355,16 +354,16 @@ def delete_cell(self, column_family_id, column, time_range=None):
                                  or columns with cells being deleted. Must be
                                  of the form ``[_a-zA-Z0-9][-_.a-zA-Z0-9]*``.
 
-        :type column: bytes
+        :type column: str
```
tseaver: Column IDs are `bytes`, not `str`.

bigtable/google/cloud/bigtable/table.py
tseaver: A list comprehension:

```python
        mutation_rows = [row.row_mutations for row in rows]
```

bigtable/tests/unit/test_row.py
```diff
@@ -78,80 +78,65 @@ def _make_client(self, *args, **kwargs):
...
-                         timestamp_micros=-1):
-        import six
-        import struct
+    def _set_cell_helper(self, column=None, value=b'foobar', timestamp=None):
+
+        from google.cloud.bigtable.mutation import RowMutations
 
         row_key = b'row_key'
-        column_family_id = u'column_family_id'
+        column_family_id = 'column_family_id'
```
tseaver: The `u''` prefix ensures that the value is "text" on Python2, rather than bytes.

bigtable/tests/unit/test_row.py
```diff
@@ -78,80 +78,65 @@ def _make_client(self, *args, **kwargs):
...
 
     def test_set_cell(self):
         self._set_cell_helper()
 
     def test_set_cell_with_string_column(self):
-        column_bytes = b'column'
         column_non_bytes = u'column'
-        self._set_cell_helper(column=column_non_bytes,
-                              column_bytes=column_bytes)
+        self._set_cell_helper(column=column_non_bytes)
```
tseaver: I don't think we want to support passing non-bytes column IDs.

bigtable/tests/unit/test_row.py
```diff
@@ -78,80 +78,65 @@ def _make_client(self, *args, **kwargs):
...
 
     def test_set_cell_with_integer_value(self):
-        value = 1337
+        value = b'1337'
```
tseaver: If we don't support passing integer values, then just delete this test altogether.

bigtable/tests/unit/test_row.py
```diff
@@ -78,80 +78,65 @@ def _make_client(self, *args, **kwargs):
...
         self._set_cell_helper(value=value)
 
     def test_set_cell_with_non_bytes_value(self):
         row_key = b'row_key'
         column = b'column'
-        column_family_id = u'column_family_id'
+        column_family_id = 'column_family_id'
```
tseaver: Preserve the `u''` prefix for correctness on Python2.

bigtable/tests/unit/test_row.py
```diff
@@ -164,23 +149,28 @@ def test_set_cell_with_non_null_timestamp(self):
...
+        self.assertEqual(row.row_mutations.mutations_entry.mutations,
+                         entry.mutations)
         row.delete()
 
         expected_pb = _MutationPB(
             delete_from_row=_MutationDeleteFromRowPB(),
         )
-        self.assertEqual(row._pb_mutations, [expected_pb])
+        self.assertEqual(row.row_mutations.mutations_entry.mutations[0],
+                         expected_pb)
```
tseaver: This assertion allows for (unwanted) additional mutations.  It either needs an additional check of the length, or else to be inverted, e.g.:

```python
        self.assertEqual(row.row_mutations.mutations_entry.mutations,
                         [expected_pb])
```

bigtable/tests/unit/test_row.py
```diff
@@ -190,31 +180,23 @@ def __init__(self, *args, **kwargs):
...
-            # Replace the called method with one that logs arguments.
-            def _delete_cells(self, *args, **kwargs):
-                self._args.append(args)
-                self._kwargs.append(kwargs)
-
         row_key = b'row_key'
-        column = b'column'
-        column_family_id = u'column_family_id'
+        column = 'column'
+        column_family_id = 'column_family_id'
```
tseaver: The previous diff (with `b''` and `u''`) is correct.

bigtable/tests/unit/test_row.py
```diff
@@ -190,31 +180,23 @@ def __init__(self, *args, **kwargs):
                 self._args = []
                 self._kwargs = []
 
-            # Replace the called method with one that logs arguments.
-            def _delete_cells(self, *args, **kwargs):
-                self._args.append(args)
-                self._kwargs.append(kwargs)
-
```
tseaver: Without the `_delete_cells` class, the entire `MockRow` bit is pointless, and should be deleted:  the test should just use an instance of the real class.

bigtable/tests/unit/test_row.py
```diff
@@ -330,35 +335,56 @@ def test_delete_cells_with_string_columns(self):
                 column_qualifier=column2_bytes,
             ),
         )
-        self.assertEqual(row._pb_mutations, [expected_pb1, expected_pb2])
+        self.assertEqual(row.row_mutations.mutations_entry.mutations[0],
+                         expected_pb1)
+        self.assertEqual(row.row_mutations.mutations_entry.mutations[1],
+                         expected_pb2)
```
tseaver: This misses the length check.

```python
        self.assertEqual(row.row_mutations.mutations_entry.mutations
                         [expected_pb1, expected_pb2])
```

bigtable/tests/unit/test_row.py
```diff
@@ -330,35 +335,56 @@ def test_delete_cells_with_string_columns(self):
...
+
+    def _make_responses(self, codes):
+        import six
+        from google.cloud.bigtable_v2.proto.bigtable_pb2 import (
+            MutateRowsResponse)
+        from google.rpc.status_pb2 import Status
+
+        entries = [MutateRowsResponse.Entry(
+            index=i, status=Status(code=codes[i]))
+            for i in six.moves.xrange(len(codes))]
```
tseaver: This would be clearer using `enumerate`, e.g.:

```python
        entries = [MutateRowsResponse.Entry(
            index=index, status=Status(code=code))
            for index, code in enumerate(codes)]
```

bigtable/tests/unit/test_row.py
```diff
@@ -330,35 +335,56 @@ def test_delete_cells_with_string_columns(self):
...
-        api = bigtable_client.BigtableClient(mock.Mock())
+        data_api = bigtable_client.BigtableClient(mock.Mock())
+        table_api = bigtable_table_admin_client.BigtableTableAdminClient(
+            mock.Mock())
         credentials = _make_credentials()
         client = self._make_client(project=project_id,
                                    credentials=credentials, admin=True)
-        table = _Table(table_name, client=client)
+        instance = client.instance(instance_id)
+        table = instance.table(table_id)
```
tseaver: Why are we creating "real" instance / table objects here?  We stub out the `table_data_client` below, which is all that the `row.commit` call needs.

bigtable/tests/unit/test_row.py
```diff
@@ -389,40 +415,39 @@ class ErrorUnavailable(grpc.RpcError, grpc.Call):
...
-        from google.cloud.bigtable import row as MUT
-
-        row_key = b'row_key'
-        table = object()
-        row = self._make_one(row_key, table)
-        row._pb_mutations = [1, 2, 3]
-        num_mutations = len(row._pb_mutations)
-        with _Monkey(MUT, MAX_MUTATIONS=num_mutations - 1):
-            with self.assertRaises(ValueError):
-                row.commit()
```
tseaver: Why is this test gone?  Don't we need to check the case that there are too many mutations?

bigtable/tests/unit/test_row.py
aneepct: @tseaver I am getting error ```TypeError: Can only compare repeated composite fields against other repeated composite fields.```. I think it is because mutations is the protobuf message

bigtable/tests/unit/test_row.py
aneepct: @tseaver I am getting error ```TypeError: Can only compare repeated composite fields against other repeated composite fields.```  I think it is because mutations is the protobuf message

bigtable/tests/unit/test_row.py
aneepct: @tseaver Here I am getting instance / table from mocked client which is on self._make_client().

bigtable/tests/unit/test_row.py
aneepct: @tseaver on our last meeting with Mr. Solomon. We are shifting this Max Mutations check on RowMutation class which is on mutation.py. That is why I removed this from test_row.py

bigtable/tests/unit/test_row.py
sduskis: @aneepct: ping

bigtable/tests/unit/test_row.py
sduskis: @aneepct: ping

bigtable/tests/unit/test_row.py
sduskis: @aneepct ping

bigtable/tests/unit/test_row.py
sduskis: @aneepct: ping.

bigtable/tests/unit/test_row.py
sduskis: @aneepct can you convert `expected_pb` into a MutationEntry?

bigtable/tests/unit/test_row.py
sduskis: @aneepct and @zakons, should we remove MockRow?

bigtable/tests/unit/test_row.py
sduskis: @aneepct: ping.

bigtable/tests/unit/test_row.py
sduskis: @aneepct: should we just create a MutationEntry for `expected_pb`?

bigtable/tests/unit/test_row.py
sduskis: @aneepct: ping.

bigtable/tests/unit/test_row.py
sduskis: @aneepct: this test needs to be somewhere accessible to DirectRow.  Why does it have to be removed from here?

aneepct:
@sduskis  Just finishing changes for clusters branches and these changes are the next top priority.

bigtable/tests/unit/test_row.py
aneepct: @sduskis You just need to change the name? or You want to compare Entry from Mutation? Here I am checking that which mutation is called when row.delete() api called.

aneepct:
@sduskis I am checking the MAX_BULK_MUTATIONS on test_table.py line no. 782 (test_max_bulk_mutations) and line no 31 (test__mutate_rows_too_many_mutations)

PR #5420, opened 2018-05-31
Make 'Client.list_instances' return actual instance objects, not protos.

tseaver:
/cc @aneepct @sduskis


tseaver:
@theacodes The Appveyor failure here is the known Pandas + PyArrow + Python 2.7 problem in BigQuery (#5441, solved via PR #5442).  Rather than rebase and force CI to run again, I'd like to just merge here on the strength of the CircleCI green.

theacodes:
That's fine, but I'd like @sduskis to take a look first.

PR #5421, opened 2018-05-31
Unmark most of the xfailed tests.

tseaver:
Fixed by PR #5398.


tseaver:
The source of the failures when running all system tests was re-using table names across tests:  apparently, a deleted table now "sticks around" a while, and cannot be re-created without risking timeout / deadline exceeded.  1cbdcda ensures that the tests no longer try to re-created a table (by using unique names).

tseaver:
@theacodes This is the same #5441 / #5442 issue.  I'd like to merge here as well based on the CircleCI green.

PR #5429, opened 2018-06-01
BigQuery: Adds time filtering to client.list_jobs()

### LINKED ISSUE #5067: BigQuery - filter job list by time window
Please add `minCreationTime` and `maxCreationTime` optional parameters to `list_jobs`, which are supported by the API.

alixhami:
Resolves #5067

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -33,6 +33,7 @@ from google.cloud.client import ClientWithProject
 
 from google.cloud.bigquery._helpers import DEFAULT_RETRY
 from google.cloud.bigquery._helpers import _SCALAR_VALUE_TO_JSON_ROW
+from google.cloud.bigquery._helpers import _str_or_none
 from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetListItem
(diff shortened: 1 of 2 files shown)


bigquery/tests/unit/test_client.py
```diff
@@ -1753,6 +1753,23 @@ def test_list_jobs_w_project(self):
                 'projection': 'full',
             })
 
+    def test_list_jobs_w_time_filter(self):
+        creds = _make_credentials()
+        client = self._make_one(self.PROJECT, creds)
+        conn = client._connection = _make_connection({})
+
+        list(client.list_jobs(
+            min_creation_time=1, max_creation_time=1527874895820))
```
tseaver: I've got a preference for naming the majyk constants here.  It isn't a blocker for merge, but it does ease maintenance down the road.

alixhami: The value of the inputs is not important here because the function output is not being saved or tested in this test. This is just testing that the parameters exist and get passed to the API properly.

crwilcox: I can agree with Tres, specifically for the max_creation since it does look magical. Does that number have any significance?

alixhami:
@shollyman Per our in person discussion, I will send another PR expanding the `bigquery_list_jobs` snippet to include examples of more of the possible arguments (`all_users`, `min_creation_time`, `max_creation_time`). This will serve as an integration test.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -662,52 +663,62 @@ def cancel_job(
...
+            state_filter (str, optional):
+                If set, include only jobs matching the given state. One of:
                     * ``"done"``
                     * ``"pending"``
                     * ``"running"``
-            retry (google.api_core.retry.Retry):
-                Optional. How to retry the RPC.
+            retry (google.api_core.retry.Retry, optional):
+                How to retry the RPC.
+            min_creation_time (int, optional):
```
tswast: I'd strongly prefer a datetime.datetime here.

bigquery/tests/unit/test_client.py
tswast: I added some constants in the follow-up change https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5431

PR #5436, opened 2018-06-04
BigQuery: Adds examples of optional params for client.list_jobs() snippet

alixhami:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5430

### DIFF
--- a/docs/bigquery/snippets.py
+++ b/docs/bigquery/snippets.py
@@ -2297,10 +2297,30 @@ def test_client_list_jobs(client):
     # [START bigquery_list_jobs]
     # from google.cloud import bigquery
     # client = bigquery.Client(project='my_project')
+    import datetime
 
     # List the 10 most recent jobs in reverse chronological order.
     # Omit the max_results parameter to list jobs from the past 6 months.
+    print("Last 10 jobs:")
     for job in client.list_jobs(max_results=10):  # API request(s)
+        print(job.job_id)
+
+    # The following are examples of additional optional parameters:
+
+    # Use min_creation_time and/or max_creation_time to specify a time window.
+    print("Jobs from the last two days:")
+    two_days_ago = datetime.datetime.now() - datetime.timedelta(days=2)
+    for job in client.list_jobs(min_creation_time=two_days_ago):
+        print(job.job_id)
+
+    # Use all_users to include jobs run by all users in the project.
+    print("Last 10 jobs run by all users:")
+    for job in client.list_jobs(max_results=10, all_users=True):
+        print(job.job_id)
+
+    # Use state_filter to filter by job state.
+    print("Jobs currently running:")
+    for job in client.list_jobs(state_filter='RUNNING'):
         print(job.job_id)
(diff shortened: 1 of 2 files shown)


docs/bigquery/snippets.py
```diff
@@ -2297,11 +2297,31 @@ def test_client_list_jobs(client):
...
     # List the 10 most recent jobs in reverse chronological order.
     # Omit the max_results parameter to list jobs from the past 6 months.
+    print("Last 10 jobs:")
     for job in client.list_jobs(max_results=10):  # API request(s)
         print(job.job_id)
+
+    # The following are examples of additional optional parameters:
+
+    # Use min_creation_time and/or max_creation_time to specify a time window.
+    two_days_ago = datetime.datetime.now() - datetime.timedelta(days=2)
```
tswast: I think we may want to use the `utcnow()` function instead of `now()`. I believe `now()` can return the local time but doesn't always populate the local time zone correctly.

tswast: From https://julien.danjou.info/python-and-timezones/

```
>>> import datetime
>>> datetime.datetime.utcnow()
datetime.datetime(2015, 6, 15, 13, 24, 48, 27631)
>>> datetime.datetime.now()
datetime.datetime(2015, 6, 15, 15, 24, 52, 276161)
```

> As you can notice, none of these results contains timezone information. Indeed, Python datetime API always returns unaware datetime objects,

docs/bigquery/snippets.py
```diff
@@ -2297,11 +2297,31 @@ def test_client_list_jobs(client):
...
     # Omit the max_results parameter to list jobs from the past 6 months.
+    print("Last 10 jobs:")
     for job in client.list_jobs(max_results=10):  # API request(s)
         print(job.job_id)
+
+    # The following are examples of additional optional parameters:
+
+    # Use min_creation_time and/or max_creation_time to specify a time window.
+    print("Jobs from the last two days:")
+    two_days_ago = datetime.datetime.utcnow() - datetime.timedelta(days=2)
```
tswast: Thanks for the update. I wonder if 2 days is too many for the CircleCI project, since it runs a lot of jobs. Maybe cut it to the last 10 minutes?

tseaver: Trimming the number of jobs listed SGTM.

shollyman: Yeah, 10 minutes should be sufficient for this case.

docs/bigquery/snippets.py
```diff
@@ -2297,11 +2297,31 @@ def test_client_list_jobs(client):
...
+    for job in client.list_jobs(min_creation_time=two_days_ago):
+        print(job.job_id)
+
+    # Use all_users to include jobs run by all users in the project.
+    print("Last 10 jobs run by all users:")
+    for job in client.list_jobs(max_results=10, all_users=True):
+        print(job.job_id)
```
shollyman: Consider printing the job owner along with the job id for the allUsers projection?

PR #5438, opened 2018-06-05
Add additional error handling to unary RPCs

theacodes:


tseaver:
Maybe the commit message should also note the switch away from the root logger?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/bidi.py
```diff
@@ -481,6 +481,11 @@ def _thread_main(self):
                 '%s caught unexpected exception %s and will exit.',
                 _BIDIRECTIONAL_CONSUMER_NAME, exc)
 
+        else:
```
crwilcox: Is self._bidi_rpc.is_active being false really unexpected?

theacodes: Yep, it really shouldn't happen. This is to help produce more logs to debug.

crwilcox: SGTM. I just didn't know if 'unexpectedly' was a bit strong or not.

theacodes:
> Maybe the commit message should also note the switch away from the root logger?

Eh, too minor to call out.

Issue #5444, opened 2018-06-06
Pubsub 0.35.3 stop consuming messages - ConsumeBidirectionalStream exited RPC


shomratalon:
After an ~hour, the rpc connection is closed:

```
[2018-06-06 11:02:36,258 - INFO - Thread-14 - streaming_pull_manager] Observed recoverable stream error 504 Deadline Exceeded
[2018-06-06 11:02:36,259 - INFO - Thread-ConsumeBidirectionalStream - streaming_pull_manager] Observed recoverable stream error 504 Deadline Exceeded
[2018-06-06 11:02:36,260 - INFO - Thread-14 - streaming_pull_manager] Observed recoverable stream error 504 Deadline Exceeded
[2018-06-06 11:02:36,260 - **ERROR - Thread-ConsumeBidirectionalStream - bidi] The bidirectional RPC unexpectedly exited. <google.api_core.grpc_helpers._StreamingResponseIterator object at 0x7fb758946090>** 
```

I need to cancel and start the subscriber again.

OS type and version - "Debian GNU/Linux 8 (jessie)"
Version - Python 2.7.9
4. pip freeze | grep pubsub
google-cloud-pubsub==0.35.3

Code example:

```
        def subscriber_handler(message):
            path = message.attributes.get("type")
            self.handler_function(path=path, body=message.data)  
            message.ack()
        self.listening_subscriber = self.subscriber.subscribe(subscription_path, callback=subscriber_handler, flow_control=flow_control)
```

All the other info is the same as in https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5325.

theacodes:
Thanks, @shomratalon! Can you provide me a bit more information?

1. What does your callback do?
2. How many messages per second are you processing?
3. How are you authenticating?

theacodes:
Okay, I think I may have identified one potential cause of this issue. I've yet to be able to reproduce this so if y'all could try this out and let me know that would be really awesome.

You can install the dev version by either:

```
pip install --upgrade http://temp.theadora.io/google_cloud_pubsub-0.35.3-py2.py3-none-any.whl
```

or

```
git clone https://github.com/GoogleCloudPlatform/google-cloud-python.git
cd google-cloud-python
git checkout pubsub-lock-active
pip install --upgrade .
```

I would **highly** recommend doing this in a *fresh* virtualenv.

theacodes:
[...]

PR #5445, opened 2018-06-06
Use operational lock when checking for activity on streams

theacodes:


theacodes:
Possible resolution for #5444

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/bidi.py
```diff
@@ -393,6 +393,14 @@ def recv(self):
...
             super(ResumableBidiRpc, self).recv)
 
+    @property
+    def is_active(self):
+        """bool: True if this stream is currently open and active."""
+        # Use the operational lock. It's entirely possible for something
+        # to check the active state *while* the RPC is being retried.
+        with self._operational_lock:
+            return self.call is not None and self.call.is_active()
+
```
tseaver: Where is the `is_active` property being used?

theacodes: Down in BackgroundStreamConsumer

PR #5446, opened 2018-06-06
Recover streams during the gRPC error callback

### LINKED ISSUE #5444: Pubsub 0.35.3 stop consuming messages - ConsumeBidirectionalStream exited RPC

theacodes:
Resolves #5444

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -402,6 +402,7 @@ class StreamingPullManager(object):
         After the messages have all had their ack deadline updated, execute
         the callback for each message using the executor.
         """
+
         _LOGGER.debug(
             'Scheduling callbacks for %s messages.',
             len(response.received_messages))
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/bidi.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/bidi.py
@@ -235,7 +235,9 @@ class BidiRpc(object):
             raise ValueError(
                 'Can not send() on an RPC that has never been open()ed.')
 
-        if self.is_active:
+        # Don't use self.is_active(), as ResumableBidiRpc will overload it
+        # to mean something semantically different.
+        if self.call.is_active():
             self._request_queue.put(request)
         else:
             # calling next should cause the call to raise.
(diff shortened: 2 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -402,6 +402,7 @@ def _on_response(self, response):
         After the messages have all had their ack deadline updated, execute
         the callback for each message using the executor.
         """
+
```
tseaver: Stray?

theacodes: Yes, but I'm keeping it. :)

theacodes:
From a conversations with me @crwilcox offline for records:

> the root cause is that gRPC can mark the RPC as inactive in another thread, after recv() has been called. So background consumer would recv(), get a (buffered) response, and the gRPC would reap the RPC right after. When background consumer would go to the top of its while loop, the `is_active` condition would be `False`. Had anything called `send()` or if the background consumer called `recv()` somehow before then, the RPC would be re-established.

There are *three* changes to address this:

1. we attempt to recover from a place we didn't previously (the gRPC error callback).

> This is defensive. Why not just go ahead and recover in the error callback? In practice, this preempts `recoverable` (which wraps `send` and `recv`) from ever "winning the race" to recover the stream. The stream will always be recovered by the gRPC error callback.

2. we monitor a different semaphore for if a grpc channel is actually active.

> This alone is probably enough to fix the condition above. By checking if the bidi rpc is *finalized* instead of whether or not the underlying RPC is active, the `while` condition in the background consumer would be `True` right up until finalization.

3.  when we can't recover in recoverable, finalize in addition to close. We use if it was finalized to detect if the stream is active.

> This is a follow-up to (2) that makes `is_active` stay semantically correct. Without it, recoverable would never mark the recoverable bidi rpc as inactive in the case of an unrecoverable error.

PR #5452, opened 2018-06-07
minor fix to datastore example

salrashid123:
update example for strong consistency (i.,e key-get vs query).  

fixes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2023


tseaver:
@salrashid123 Thanks for the patch!

PR #5453, opened 2018-06-07
BigQuery: adds legacy and batch samples

alixhami:


docs/bigquery/snippets.py
```diff
@@ -2000,42 +1998,66 @@ def test_client_query_destination_table(client, to_delete):
...
-    # [END bigquery_query_destination_table]
+def test_client_query_destination_table_legacy(client, to_delete):
+    dataset_id = 'query_destination_table_legacy_{}'.format(_millis())
+    dataset_ref = client.dataset(dataset_id)
+    to_delete.append(dataset_ref)
+    dataset = bigquery.Dataset(dataset_ref)
+    dataset.location = 'US'
+    client.create_dataset(dataset)
+
+    # [START bigquery_query_legacy_large_results]
```
tswast: Looks like we aren't setting the `allow_large_results` property?
https://google-cloud-python.readthedocs.io/en/latest/bigquery/generated/google.cloud.bigquery.job.QueryJobConfig.allow_large_results.html#google.cloud.bigquery.job.QueryJobConfig.allow_large_results

docs/bigquery/snippets.py
```diff
@@ -2000,42 +1998,66 @@ def test_client_query_destination_table(client, to_delete):
...
-
-    # The write_disposition specifies the behavior when writing query results
-    # to a table that already exists. With WRITE_TRUNCATE, any existing rows
-    # in the table are overwritten by the query results.
-    job_config.write_disposition = bigquery.WriteDisposition.WRITE_TRUNCATE
+    sql = """
+        SELECT corpus
+        FROM `bigquery-public-data.samples.shakespeare`
+        GROUP BY corpus;
+    """
```
tseaver: Do we need to be exercising a more expensive query (timewise) in order to show using a destination table?

docs/bigquery/snippets.py
```diff
@@ -2000,42 +1998,66 @@ def test_client_query_destination_table(client, to_delete):
...
+    query_job = client.query(
+        sql,
+        # Location must match that of the dataset(s) referenced in the query
+        # and of the destination table.
+        location='US',
+        job_config=job_config)  # API request - starts the query
+
+    query_job.result()  # Waits for the query to finish
+    print('Query results loaded to table {}'.format(table_ref.path))
+    # [END bigquery_query_legacy_large_results]
```
tseaver: Do we really need to re-do the expensive query using legacy syntax here?  It doesn't really change the destination table usage.

docs/bigquery/snippets.py
```diff
@@ -2082,6 +2104,34 @@ def test_client_query_destination_table_cmek(client, to_delete):
...
+        GROUP BY corpus;
+    """
+    # Location must match that of the dataset(s) referenced in the query.
+    location = 'US'
+
+    # API request - starts the query
+    query_job = client.query(sql, location=location, job_config=job_config)
+
+    # Monitor job progress until it reaches the `DONE` state by polling
+    # periodically.
```
tseaver: The comment seems to be copy-pasta:  we aren't actually polling / waiting for the job to complete here.

docs/bigquery/snippets.py
alixhami: Oops - added in e078566

docs/bigquery/snippets.py
alixhami: You have to set `allow_large_results` (added in e078566), which is why there is a separate sample in the [docs](https://cloud.google.com/bigquery/docs/writing-results#writing_large_results_using_legacy_sql)

shollyman: Behavior of allow_large_results on the backend just controls details of the materialization path.  This query is pretty trivial (2 mb scanned in 1 file, yielding 42 rows), so there's not much speedup to gain from a simpler query as its already a subsecond query in the normal case.

PR #5459, opened 2018-06-08
Use subpackage installation instructions on main index.

tswast:
* Add installation instructions to subpackage pages that were missing them.
* Add deprecation notice to google-cloud releases page.

Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5455

### DIFF
--- a/docs/bigquery/usage.rst
+++ b/docs/bigquery/usage.rst
@@ -10,6 +10,15 @@ BigQuery
 
 .. contents:: :local:
 
+Installation
+------------
+
+Install the ``google-cloud-bigquery`` library using ``pip``:
+
+.. code-block:: console
+
+    $ pip install google-cloud-bigquery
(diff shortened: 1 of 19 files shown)


docs/bigquery/usage.rst
```diff
@@ -10,6 +10,15 @@ BigQuery
 
 .. contents:: :local:
 
+Installation
+------------
+
+Install the ``google-cloud-bigquery`` library using ``pip``:
+
+.. code-block:: console
```
tseaver: I believe the language code you want here is `bash`.

tswast: I copy-pasted from other installation instructions. Seems to render as intended with `console`.

image

crwilcox:
Thanks for doing this Tim!

PR #5460, opened 2018-06-08
BigQuery: Adds nested repeated schema and undelete table samples

alixhami:


docs/bigquery/snippets.py
```diff
@@ -544,6 +544,40 @@ def test_create_table_then_add_schema(client, to_delete):
     # [END bigquery_add_schema_to_empty]
 
 
+def test_create_table_nested_repeated_schema(client, to_delete):
+    dataset_id = 'create_table_dataset_{}'.format(_millis())
```
tseaver: Maybe make the name more unique?  E.g., `create_table_nested_repeated_{}`?

tswast: Yeah, with `_millis()` I've definitely had some flakes due to collisions before.

PR #5467, opened 2018-06-11
Regenerate vision gapics

theacodes:


vision/google/cloud/vision_v1/types.py
```diff
@@ -15,46 +15,40 @@
...
+        operations_pb2,
+        any_pb2,
+        descriptor_pb2,
+        empty_pb2,
+        timestamp_pb2,
+        wrappers_pb2,
+        status_pb2,
+        color_pb2,
+        latlng_pb2,
+):
```
tseaver: The generator must not undo the change to stop overwriting the modules of shared message types (#5364).

crwilcox: just fixed this in the tasks generation if you want to copy/pasta the same transformation (or at least it should be pretty similar)

tseaver:
@theacodes Did you overlook my comment on undoing #5364?

theacodes:
@tseaver saw it, but we won't be fixing it here. We'll wait for the generator fix and re-generate.

PR #5468, opened 2018-06-11
Generate text-to-speech v1 and v1beta1

theacodes:


tseaver:
I guess `0.1.0` is the ideal time to be making such a breaking change, but I thought we were versioning our APIs so that people could migrate their code in place, piecemeal, using the upgraded package (i.e., leaving older API version sub-packages around for at least a couple of releases).

theacodes:
@tseaver indeed, but yeah, this is a super early library so we opted to just s/v1beta/v1.

texttospeech/google/cloud/texttospeech_v1/types.py
```diff
@@ -15,32 +15,20 @@
...
-    for name, message in get_messages(module).items():
-        setattr(sys.modules[__name__], name, message)
-        names.append(name)
-
-for module in _local_modules:
+for module in (
+        http_pb2,
+        cloud_tts_pb2,
+        descriptor_pb2,
+):
```
tseaver: Undoes the fix from #5364.  We need to fix the generator.

PR #5469, opened 2018-06-11
regenerate tasks v2beta2

crwilcox:


tasks/google/cloud/tasks_v2beta2/types.py
```diff
@@ -26,41 +32,23 @@
...
+        iam_policy_pb2,
+        policy_pb2,
+        any_pb2,
+        descriptor_pb2,
+        duration_pb2,
+        empty_pb2,
+        field_mask_pb2,
+        timestamp_pb2,
+        status_pb2,
+):
```
tseaver: The generator needs to be fixed -- it should not be undoing the `_shared_modules` / `_local_modules` split.

crwilcox: @theacodes can you confirm that this is making an incorrect change? I can add a transform to handle it if so.

crwilcox: PS - Vision gen has the same issue if this is a problem

tseaver: See #5364.

crwilcox: addressed.

tasks/synth.py
```diff
@@ -82,3 +82,42 @@
...
+    for name, message in get_messages(module).items():
+        setattr(sys.modules[__name__], name, message)
+        names.append(name)
+
+for module in _local_modules:
+    for name, message in get_messages(module).items():
+        message.__module__ = 'google.cloud.tasks_v2beta2.types'
+        setattr(sys.modules[__name__], name, message)
+        names.append(name)
+""")
```
tseaver: Ugh, can we not just fix the generator?

tseaver: OK, I see the [generator issue](https://github.com/googleapis/gapic-generator/issues/2058).

PR #5481, opened 2018-06-13
regen bigquery_datatransfer with toolkit and googleapis changes (WIP)

andreamlin:
- in googleapis bigquery_datatransfer_gapic.yaml, set python package as com.google.cloud.bigquery.datatransfer_v1
- regen with toolkit local_and_shared branch; change pip egg to use `google-cloud-bigquerydatatransfer` and pip install -e


tseaver:
The main `google-cloud-bigquery` package does not declare `google.cloud.bigquery` as a namespace package, which means that the `google-cloud-bigquery-datatransfer` package cannot install files (the `datatranfer_v1` subdirectory) inside it.  We need to keep the underscore in the package name, `bigquery_datatransfer_v1`.

Issue #5083, opened 2018-03-20
Re-do fix to run only affected systests on CircleCI


theacodes:
It seems #5018 broke our tests, and now circle is skipping stuff.

See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5079

chemelnucfin:
temporary fix in #5092

tseaver:
Tweaked after merge of #5092.

tseaver:
#5487 *seems* to be the right fix:  the clue was to do the `git merge-base --fork-point` test against `baseline/master` (the local master is in some weird state).

PR #5482, opened 2018-06-13
Redo 5018: run only needed systests

tseaver:
This version emits debugging output from `get_target_packages.py`, in "comment" form.

It also updates the grep commands used to detect individual API packages to use whole-line matching only (`grep -x`).

See #5018, #5083

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: googleapis/nox:0.18.2
+      - image: googleapis/nox:0.17.0
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


tseaver:
@theacodes, @crwilcox When we merge this PR, we will need to monitor closely subsequent PRs, looking particularly at the debugging output in the "Declare target packages" build phase.

tseaver:
CI is finally passing:  See the [Declare target packages](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/6879#action-105] section, which shows the changed files and *no* packages (because we didn't touch an API or core subpackaged_.

tseaver:
*Fingers crossed.*

PR #5491, opened 2018-06-14
Adds samples for query external data sources (GCS & Sheets)

alixhami:


docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+    assert len(w_states) == 4
+
+
+def test_query_external_gcs_permanent_table(client, to_delete):
+    dataset_id = 'query_external_gcs_{}'.format(_millis())
+    dataset = bigquery.Dataset(client.dataset(dataset_id))
+    client.create_dataset(dataset)
+    to_delete.append(dataset)
+
+    # [START bigquery_query_external_gcs_perm]
```
tswast: Should creating the permanent table be a separate sample from querying one? I guess the query part is pretty small, so maybe not.

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+    dataset_ref = client.dataset(dataset_id)
+    table_id = 'us_states'
+    schema = [
+        bigquery.SchemaField('name', 'STRING'),
+        bigquery.SchemaField('post_abbr', 'STRING')
+    ]
+    table = bigquery.Table(dataset_ref.table(table_id), schema=schema)
+    external_config = bigquery.ExternalConfig('CSV')
+    external_config.source_uris = [
+        'gs://cloud-samples-data/bigquery/us-states/us-states.csv',
```
tswast: Does the doc use CSV? If we did Avro we could drop the schema definition.

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
     # [END bigquery_query_no_cache]
 
 
+def test_query_external_gcs_temporary_table(client):
+    # [START bigquery_query_external_gcs_temp]
+    # from google.cloud import bigquery
+    # client = bigquery.Client()
+
+    # Configure the external data source and query job
+    external_config = bigquery.ExternalConfig('CSV')
```
tswast: Maybe use [SourceFormat.CSV](https://google-cloud-python.readthedocs.io/en/latest/bigquery/generated/google.cloud.bigquery.job.SourceFormat.CSV.html#google.cloud.bigquery.job.SourceFormat.CSV) instead of a plain string?

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+    # [END bigquery_query_external_gcs_perm]
+    assert len(w_states) == 4
+
+
+def test_query_external_sheets_temporary_table(client):
+    # [START bigquery_query_external_sheets_temp]
+    # [START bigquery_auth_drive_scope]
+    # from google.cloud import bigquery
+
+    import google.auth
```
tswast: Can you move this import above the `# from google.cloud import bigquery` so the imports are alphabetical?

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+
+    # Create credentials with Drive & BigQuery API scopes
+    # Both APIs must be enabled for your project before running this code
+    credentials, project = google.auth.default(scopes=[
+        'https://www.googleapis.com/auth/drive',
+        'https://www.googleapis.com/auth/bigquery',
+    ])
+    client = bigquery.Client(credentials=credentials)
```
tswast: Should probably pass `project` in, too. Otherwise `Client` is going to have to call `google.auth.default` again internally to get the same project out.

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+    # [END bigquery_auth_drive_scope]
+
+    # Configure the external data source and query job
+    external_config = bigquery.ExternalConfig('GOOGLE_SHEETS')
+    # Use a shareable link or grant viewing access to the email address you
+    # used to authenticate with BigQuery (this example Sheet is public)
```
tswast: 👍 Nice sheet.

Should we clarify that it is possible to share sheets with service accounts?

docs/bigquery/snippets.py
```diff
@@ -2579,6 +2579,178 @@ def test_query_no_cache(client):
...
+
+    import google.auth
+
+    # Create credentials with Drive & BigQuery API scopes
+    # Both APIs must be enabled for your project before running this code
+    credentials, project = google.auth.default(scopes=[
+        'https://www.googleapis.com/auth/drive',
+        'https://www.googleapis.com/auth/bigquery',
+    ])
+    client = bigquery.Client(credentials=credentials)
```
tswast: Ditto that we should pass the project in here.

docs/bigquery/snippets.py
alixhami: The docs use a general example as well as CSV. I can change to Avro if you think that would be more common use case though.

docs/bigquery/snippets.py
alixhami: I didn't want to confuse non-service account users by referencing service accounts since those would also be email addresses. Do you think it is necessary for clarity to say that this also applies to service accounts?

tswast: When I search for "google sheets share with service account" I don't see any official docs that verify that you can do it, but you're right that email is the most general.

Probably the most appropriate place for how to share with a service account is in the Sheets API or Drive API docs rather than this sample.

docs/bigquery/snippets.py
tswast: You're right that we should probably create a separate set of constants since the set of supported formats aren't the same. For example, you can't use Parquet in an external config yet.

docs/bigquery/snippets.py
tswast: CSV is probably more common, but Avro would make the sample smaller. CSV is probably the right choice.

docs/bigquery/snippets.py
alixhami: When I was testing out how this worked, I just shared the file with my service account email address and it worked. For example, share with `my-service-account@project.iam.gserviceaccount.com` instead of `actual.person@gmail.com`. The doc that this sample will be included in actually recommends using a shareable link, which I also referenced in the comment, and is probably a better idea anyway. I just thought it would be nice to document the options.

tswast: Shareable link allows service accounts to access for docs owned by gmail accounts but not for docs owned by gsuite customers.

docs/bigquery/snippets.py
alixhami: Created issue https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5499

PR #5498, opened 2018-06-14
Override gRPC max message lengths.

### LINKED ISSUE #5362: Fix Failing BigTable System tests
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/6646

tseaver:
Remove 'xfail' for the large cell test which needs the override.

Closes #5362.

### DIFF
--- a/bigtable/google/cloud/bigtable_v2/gapic/bigtable_client.py
+++ b/bigtable/google/cloud/bigtable_v2/gapic/bigtable_client.py
@@ -97,6 +97,10 @@ class BigtableClient(object):
                 self.SERVICE_ADDRESS,
                 credentials=credentials,
                 scopes=self._DEFAULT_SCOPES,
+                options={
+                    'grpc.max_send_message_length': -1,
+                    'grpc.max_receive_message_length': -1,
+                }.items(),
             )
 
         # Create the gRPC stubs.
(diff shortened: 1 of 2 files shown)


bigtable/google/cloud/bigtable_v2/gapic/bigtable_client.py
```diff
@@ -97,6 +97,10 @@ def __init__(self,
                 self.SERVICE_ADDRESS,
                 credentials=credentials,
                 scopes=self._DEFAULT_SCOPES,
+                options={
```
theacodes: Hrm, I'm a little worried about fragility here. Do we make users construct this directly or do we have a wrapper class around this?

tseaver: The wrapper class is `google.cloud.bigtable.client.Client`, which constructs / holds an instance of this class as its `table_data_client` property.

theacodes: wanna move that there? (like we do in pubsub)?

tseaver: This *is* the equivalent location for `pubsub_v1/publisher/client.py` / `pubsub_v1/subscriber/client.py`:  there is no wrapper client class there.

PR #5502, opened 2018-06-15
Add `pip_install: true` to ReadTheDocs config.

tswast:
This option fixed the docs build for `pandas-gbq`. I suspect it is needed for the docs here too.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,4 +2,5 @@ requirements_file: docs/requirements.txt
 build:
   image: latest
 python:
+  pip_install: true
   version: 3.6


tseaver:
Hmm, that might be problematic, now that I think of it.  Installing the umbrella `google-cloud` package is deprecated, right?

tswast:
I think @crwilcox is working on deprecating it, yeah. It's still used for the docs AFAIK, though.

theacodes:
It's deprecated and we shouldn't merge this.

tseaver:
We are already passing `requirements_file: docs/requirements.txt`, which installs all the API subpackages.

crwilcox:
The google-cloud package will be fully deprecated next week. The countdown started a bit back but the date of deprecation is June 18 (Monday).

PR #5503, opened 2018-06-15
The offset param to the fetch method of a query did not work properly one there was a cursor

### LINKED ISSUE #4675: Datastore query offsets are not working as it supposed to
This is a weird behaviour when setting the offset parameter so I have 9825 entities and 
```python
len(list(client.query(kind="Foo").fetch()))
```
Will yield 9825, but
```python
len(list(client.query(kind="Foo").fetch(offset=1)))
```
will yield 9808

mcdonc:
This fixes the issue at GoogleCloudPlatform/google-cloud-python#4675 where fetch executions that pass the offset parameter would return incorrect result values

Closes #4675.

### DIFF
--- a/datastore/google/cloud/datastore/query.py
+++ b/datastore/google/cloud/datastore/query.py
@@ -452,10 +452,12 @@ class Iterator(page_iterator.Iterator):
         if self.max_results is not None:
             pb.limit.value = self.max_results - self.num_results
 
-        if self._offset is not None:
-            # NOTE: The offset goes down relative to the location
-            #       because we are updating the cursor each time.
-            pb.offset = self._offset - self._skipped_results
+        if start_cursor is None and self._offset is not None:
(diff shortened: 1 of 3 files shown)


mcdonc:
@theacodes apologies, I did not have the email account I am using for Google communications associated with my GitHub account; I *think* that the Agendaless CLA has already been updated to include that account.  I've now done that and to prevent future cla: no issues I've set user.email for my checkout of this repo to that email (chrismgoog@gmail.com).  If I need to do anything else, please let know.  Thanks!

theacodes:
@mcdonc no worries. Before I go and human-approve this, can you try re-writing your commits to update the author with the proper email? Should be able to do `git commit --amend --update-author --no-edit` and then `git push -f`.

mcdonc:
@theacodes done and looks like googlebot cla-approved the result; thanks!

theacodes:
Yay. Thank you!

PR #5509, opened 2018-06-20
Clarify 'None' values for resource-backed properties.

### LINKED ISSUE #3989: Documentation Unclear: "If the property is not set locally, returns None."
The [Google Storage Blob documentation](https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html) says that almost all properties return the value `None` if the "property is not set locally." However, there is no documentation explaining how properties are set/unset locally or when that occurs. 

This is confusing, there should be a preamble explaining when local Blob properties are set locally.

### Motivation

I ran into this because I started seeing some of these values return `None` in my project in unexpected circumstances (there may be another bug there), but when I went to the documentation, it did not shed much light on why or when these properties should be returning `None`.

tseaver:
Closes #3989.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -1501,8 +1501,6 @@ class Blob(_PropertyMixin):
 
     See `RFC 7234`_ and `API reference docs`_.
 
-    If the property is not set locally, returns :data:`None`.
-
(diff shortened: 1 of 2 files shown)


tseaver:
@theacodes PTAL.

PR #5541, opened 2018-06-26
[wip] Introduce to the `Blob` api extra headers as a way to use extensions (custom) http headers

mlhamel:
Fixing https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5538

This is a proposition of feature that is adding support for setting `extra_headers` to `upload_from_file`, `upload_from_filename` and `upload_from_string`.

Those `extra_headers` options are being documented as part of the official GCS documentation at https://cloud.google.com/storage/docs/xml-api/reference-headers and used to be part of the cloudstorage api of Google App Engine.

Please note that we would probably want to also introduce those options for `delete` and functions of the `Blob` Api. We could maybe also think of adding it to the `Bucket` api.

### DIFF
--- a/storage/tests/unit/test_blob.py
+++ b/storage/tests/unit/test_blob.py
@@ -1251,8 +1251,7 @@ class Test_Blob(unittest.TestCase):
         self.assertGreater(size, len(data))
 
         with self.assertRaises(ValueError) as exc_info:
-            blob._do_multipart_upload(None, stream, None, size, None, None,
-                None)
+            blob._do_multipart_upload(None, stream, None, size, None, None, None)
 
         exc_contents = str(exc_info.exception)
         self.assertIn(
@@ -1294,8 +1293,8 @@ class Test_Blob(unittest.TestCase):
         content_type = u'text/plain'
         upload, transport = blob._initiate_resumable_upload(
             client, stream, content_type, size, num_retries,
-            extra_headers=extra_headers,
-            chunk_size=chunk_size, predefined_acl=predefined_acl)
+            chunk_size=chunk_size, predefined_acl=predefined_acl,
+            extra_headers=extra_headers)
 
         # Check the returned values.
         self.assertIsInstance(upload, ResumableUpload)
@@ -1625,14 +1624,14 @@ class Test_Blob(unittest.TestCase):
 
     def test_upload_from_file_success(self):
         stream = self._upload_from_file_helper(
-            predefined_acl='private', extra_headers={})
+            predefined_acl='private', extra_headers={'X-Goog-Testing': 'true'})
         assert stream.tell() == 2
 
     @mock.patch('warnings.warn')
     def test_upload_from_file_with_retries(self, mock_warn):
         from google.cloud.storage import blob as blob_module
 
-        self._upload_from_file_helper(num_retries=20, extra_headers={})
+        self._upload_from_file_helper(num_retries=20)
         mock_warn.assert_called_once_with(
             blob_module._NUM_RETRIES_MESSAGE, DeprecationWarning)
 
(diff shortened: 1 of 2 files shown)


storage/tests/unit/test_blob.py
```diff
@@ -1164,7 +1165,8 @@ def _do_multipart_success(self, mock_get_boundary, size=None,
         stream = io.BytesIO(data)
         content_type = u'application/xml'
         response = blob._do_multipart_upload(
-            client, stream, content_type, size, num_retries, predefined_acl)
+            client, stream, content_type, size, num_retries, predefined_acl,
+            extra_headers)
```
tseaver: We should be exercising the headers in a new testcase, e.g. `test__do_multipart_upload_with_headers`, which passes actual headers to `_do_multipart_success`.  E.g.:

```python
    @mock.patch(u'google.resumable_media._upload.get_boundary',
                 return_value=b'==0==')
    def test__do_multipart_upload_with_headers(self, mock_get_boundary):
        self._do_multipart_success(
            mock_get_boundary,
            headers={'X-Goog-Testing': 'true'})
```

storage/tests/unit/test_blob.py
```diff
@@ -1472,7 +1475,8 @@ def _do_resumable_upload_call2(
             'PUT', resumable_url, data=payload, headers=expected_headers)
 
     def _do_resumable_helper(
-            self, use_size=False, num_retries=None, predefined_acl=None):
+            self, use_size=False, num_retries=None, predefined_acl=None,
+            extra_headers=None):
```
tseaver: We should add a testcase, e.g. `test__do_resumable_upload_with_headers`, passing actual headers.  E.g.:

```python
    def test__do_resumable_upload_with_retry(self):
        self._do_resumable_helper(extra_headers={'X-Goog-Testing': 'true'})
```

storage/tests/unit/test_blob.py
```diff
@@ -1530,7 +1535,7 @@ def test__do_resumable_upload_with_predefined_acl(self):
 
     def _do_upload_helper(
             self, chunk_size=None, num_retries=None, predefined_acl=None,
-            size=None):
+            extra_headers=None, size=None):
```
tseaver: Likewise, needs a testcase passing actual headers.

storage/tests/unit/test_blob.py
```diff
@@ -1612,23 +1619,25 @@ def _upload_from_file_helper(self, side_effect=None, **kwargs):
...
             client, stream, content_type,
-            len(data), num_retries, predefined_acl)
+            len(data), num_retries, predefined_acl,
+            extra_headers)
         return stream
 
     def test_upload_from_file_success(self):
-        stream = self._upload_from_file_helper(predefined_acl='private')
+        stream = self._upload_from_file_helper(
+            predefined_acl='private', extra_headers={})
```
tseaver: For clarity, please pass an actual header here, e.g., `extra_headers={'X-Goog-Testing': 'true'}`.

storage/tests/unit/test_blob.py
```diff
@@ -1612,23 +1619,25 @@ def _upload_from_file_helper(self, side_effect=None, **kwargs):
...
         assert stream.tell() == 2
 
     @mock.patch('warnings.warn')
     def test_upload_from_file_with_retries(self, mock_warn):
         from google.cloud.storage import blob as blob_module
 
-        self._upload_from_file_helper(num_retries=20)
+        self._upload_from_file_helper(num_retries=20, extra_headers={})
```
tseaver: You can leave it off of other methods.

storage/tests/unit/test_blob.py
```diff
@@ -1612,23 +1618,25 @@ def _upload_from_file_helper(self, side_effect=None, **kwargs):
         num_retries = kwargs.get('num_retries')
         blob._do_upload.assert_called_once_with(
             client, stream, content_type,
-            len(data), num_retries, predefined_acl)
+            len(data), num_retries, predefined_acl,
+            extra_headers)
```
tseaver: `_do_upload` gets called with `extra_headers` as a keyword argument, so you need to have it as a keyword argument here, i.e.: `extra_headers=extra_headers`.

sduskis:
@mlhamel, this PR is out of date, and a WIP.  How can we move this PR forward (either resolving conflicts or just closing this issue)?

mlhamel:
@sduskis I'm not gonna be able to invest more time into this. I'm gonna close this PR, do you think it make sense to keep the issue open still?

Issue #5454, opened 2018-06-07
Bigtable: Fully expose Instance / cluster admin features


sduskis:
[Instance.py](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/instance.py) does not fully support instance and cluster management. [Cloud Bigtable replication allow](https://cloud.google.com/bigtable/docs/replication-overview) introduced the ability to create instances with multiple clusters.  Also, there were minor features that were added that also need to be available in the client.  Here's a list of features required:

- [x] Support labels on instance creation and updates (#5614)
- [x] Support the ability to create an instance with multiple clusters (#5622, #5663)
- [x] Users should be able to specify clusterIds and display names per cluster (#5622, #5663)
- [x] The client should use the [partialUpdateInstance](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/86102c959c6852d6b02c2fe3ddeea946a003bee6/bigtable/google/cloud/bigtable_admin_v2/gapic/bigtable_instance_admin_client.py#L591) method and not update instance. (#5643)
- [x] The project that system tests run in now need more nodes.  Increase the quota to 60 nodes so that up to 10 concurrent tests can run.

**NOTE: This is a blocker to achieving Beta status**

sduskis:
It looks like `bigtable/tests/unit/test_instance.py`'s `test_create_w_explicit_serve_nodes` method was never complete.  (see [this comment](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5542#discussion_r198204531))

The method needs to set all parameters, and the input should be checked.

PR #5542, opened 2018-06-26
Bigtable: Instance creation cleanup.

sduskis:
A Cloud Bigtable instance is a collection of clusters.  Each cluster has a node count and a location.  The python Cloud Bigtable client needs a bit of refactoring to better represent the distinction between instances and clusters.

Long term, Instance.create() should take in a complete set of variables required for creating an [instance](https://github.com/googleapis/googleapis/blob/c89b6330b6386298f8ea65e47e0b77b28294e3e7/google/bigtable/admin/v2/instance.proto#L34) and a collection of [cluster](https://github.com/googleapis/googleapis/blob/c89b6330b6386298f8ea65e47e0b77b28294e3e7/google/bigtable/admin/v2/instance.proto#L104)s.

This is PR is related to issue #5454

### DIFF
--- a/bigtable/tests/unit/test_instance.py
+++ b/bigtable/tests/unit/test_instance.py
@@ -94,7 +94,8 @@ class TestInstance(unittest.TestCase):
         display_name = 'display_name'
         client = object()
 
-        instance = self._make_one(self.INSTANCE_ID, client, display_name=display_name)
+        instance = self._make_one(self.INSTANCE_ID, client,
+                                  display_name=display_name)
         self.assertEqual(instance.instance_id, self.INSTANCE_ID)
         self.assertEqual(instance.display_name, display_name)
         self.assertIs(instance._client, client)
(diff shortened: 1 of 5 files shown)


bigtable/tests/unit/test_instance.py
```diff
@@ -330,7 +325,7 @@ def test_create_w_explicit_serve_nodes(self):
         client._instance_admin_client.bigtable_instance_admin_stub = stub
 
         # Perform the method and check the result.
-        result = instance.create()
+        result = instance.create(location_id=self.LOCATION_ID)
```
tseaver: Hmm, the name of the testcase suggests that we should be passing all the non-default arguments to `Instance.create` here (and asserting they are propagated correctly).

sduskis: I definitely agree that this needs to be changed.  I added [this comment](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5454#issuecomment-400371391) to issue #5454 to track this. This PR was intended just to be an initial refactoring, and I hope that it's OK to keep this PR as such.

There's quite a bit more work required to complete issue #5454, so is it ok to improve the test cases as a follow up PR?

sduskis:
I merged from master, so the unit test is now fully testing the expected behavior..

tseaver:
@sduskis 

> I merged from master, so the unit test is now fully testing the expected behavior..

I will merge when CI is green.

PR #5545, opened 2018-06-27
table.exists() implemented. Unit test added for exists()

vikas-jamdar:
table.exist() implemented. Unit test added for exist()

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -184,6 +184,16 @@ class Table(object):
                                                 table_id=self.table_id,
                                                 table={})
 
+    def exist(self):
+        """Checks whether a table with given table ID exist
+        or not for given Bigtable instance.
+        """
+        instance = self._instance
+        for table in instance.list_tables():
+            if self.table_id == table.table_id:
+                return True
+        return False
+
     def delete(self):
         """Delete this table."""
         client = self._instance._client
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -196,8 +196,6 @@ class Table(object):
             table_client.get_table(name=self.name)
         except NotFound:
             return False
-        except StopIteration:
-            return False
         else:
             return True
 
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -18,6 +18,7 @@
 from grpc import StatusCode
 
 from google.api_core.exceptions import RetryError
+from google.api_core.exceptions import NotFound
 from google.api_core.retry import if_exception_type
 from google.api_core.retry import Retry
 from google.cloud._helpers import _to_bytes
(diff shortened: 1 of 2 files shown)


bigtable/google/cloud/bigtable/table.py
```diff
@@ -184,6 +184,16 @@ def create(self):
                                                 table_id=self.table_id,
                                                 table={})
 
+    def exist(self):
```
tseaver: Our convention would be to spell this method name `exists`. The docstring should be in the form:

```python
    def exists(self):
        """Check whether the table exists.

        :rtype: bool
        :returns: True if the table exists, else False.
        """"
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -184,6 +184,16 @@ def create(self):
...
+        """Checks whether a table with given table ID exist
+        or not for given Bigtable instance.
+        """
```
tseaver: A cleaner / faster implementation is to use the table admin API `get_table` method, e.g.:

```python
        table_client = self._instance._client.table_admin_client
        instance_name = self._instance.name
        try:
            table_client.get_table(name=self.name)
        except exceptions.NotFound:
            return False
        else:
            return True
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -184,15 +185,21 @@ def create(self):
...
-            if self.table_id == table.table_id:
-                return True
-        return False
+        table_client = self._instance._client.table_admin_client
+        try:
+            table_client.get_table(name=self.name)
+        except NotFound:
+            return False
+        except StopIteration:
+            return False
```
tseaver: I don't see how `google.cloud.bigtable_admin_v2.gapic.bigtable_table_admin_client.BigtableTableAdminClient.get_table` would ever raise a `StopIteration`:  it is defined to return a single `google.cloud.bigtable_admin_v2.types.Table` message.

bigtable/tests/unit/test_table.py
```diff
@@ -283,6 +283,47 @@ def _create_test_helper(self):
     def test_create(self):
         self._create_test_helper()
 
+    def test_exist(self):
```
tseaver: Please match the method name change above, i.e., `test_exists`.

bigtable/tests/unit/test_table.py
```diff
@@ -283,6 +283,47 @@ def _create_test_helper(self):
...
+                table_data_v2_pb2.Table(name=self.TABLE_NAME),
+            ],
+        )
+
+        # Patch API calls
+        client._table_admin_client = table_api
+        client._instance_admin_client = instance_api
+        bigtable_table_stub = (
+            client._table_admin_client.bigtable_table_admin_stub)
+        bigtable_table_stub.ListTables.side_effect = [response_pb]
```
tseaver: Use a `NotFound` exception instance for the second call, e.g.:

```python
        from google.cloud.api_core import exceptions
        ...
        bigtable_table_stub.ListTables.side_effect = [
            response_pb,
            exceptions.NotFound('testing'),
        ]
```

bigtable/google/cloud/bigtable/table.py
vikas-jamdar: Agree with you. I also worried about that. But get_table gives StopIteration. Can you check my unittest test_exists() populating mock environment correctly.

PR #5547, opened 2018-06-28
Added retry on internal error

paoloburelli:
Added retry on internal error as suggested by the internal error message: '500 Error encountered during execution. Retrying may solve the problem.'. 
Rationalised the conditions structure so it would simplify addition of other retry conditions.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -387,6 +387,7 @@ def _should_retry(exc):
 
     return False
 
+
 def get_sub_prop(container, keys, default=None):
     """Get a nested value from a dictionary.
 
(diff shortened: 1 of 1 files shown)


tseaver:
@paoloburelli Thanks for the patch!  Reviewing it, I was going to ask that you add a testcase for the new exception reason, but found that there were no explicit test cases for `_should_retry`, which led me to add PR #5550, and then in response to @tswast's review comment, #5552, which moves `_should_retry` and `DEFAULT_RETRY` to a public module.  Both of those merged yesterday, but end up conflicting with your patch.

Can I ask that you revise this PR to re-apply your patch to the new `google/cloud/bigquery/retry.py` module, and add a test to `tests/unit/test_retry.py` for the `internalError` reason?  Sorry for the trouble!

tseaver:
@paoloburelli I've added the `'internalError'` reason to the set used in the new `retry` module.  Thanks again for your effort / inspiration!

PR #5550, opened 2018-06-28
BigQuery: add missing explict coverage for '_helpers'

tseaver:


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -388,7 +344,18 @@ def _should_retry(exc):
     return reason == 'backendError' or reason == 'rateLimitExceeded'
 
 
-def get_sub_prop(container, keys, default=None):
+DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
```
tswast: I wonder if we should move this to a public module? Currently these docs won't get rendered.

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -388,7 +344,18 @@ def _should_retry(exc):
...
+
+Any method with a ``retry`` parameter will be retried automatically,
+with reasonable defaults. To disable retry, pass ``retry=None``.
+To modify the default retry behavior, call a ``with_XXX`` method
+on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
+pass ``retry=bigquery.DEFAULT_RETRY.with_deadline(30)``.
+"""
+
+
+def _get_sub_prop(container, keys, default=None):
```
tswast: Do we need a leading underscore for functions in a private module?

tseaver: FWIW, every other function in the module follows that convention.

tswast: Fine with me either way.

bigquery/google/cloud/bigquery/_helpers.py
tseaver: See PR #5552.

PR #5553, opened 2018-06-29
Connection management support (with optional grpc_gcp dependency)

WeiranFang:
Add optional grpc_gcp dependency for api_core and spanner_v1. If user has installed grpcio-gcp, then the connection management feature will be enabled, and with the configuration specified in spanner.grpc.config . Otherwise, grpc_gcp won't be imported and everything works as usual.

### DIFF
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -157,7 +157,7 @@ def wrap_errors(callable_):
 def _create_secure_channel(
         credentials, request, target, ssl_credentials=None, **kwargs):
     """Creates a secure authorized gRPC channel.
-    
+
     This overwrites google.auth.transport.grpc.secure_authorized_channel to
     return a secure channel using grpc_gcp.secure_channel if grpc_gcp is
     available.
@@ -172,7 +172,8 @@ def _create_secure_channel(
         target (str): The host and port of the service.
         ssl_credentials (grpc.ChannelCredentials): Optional SSL channel
             credentials. This can be used to specify different certificates.
-        kwargs: Additional arguments to pass to :func:`grpc_gcp.secure_channel`.
+        kwargs: Additional arguments to pass to
+            :func:`grpc_gcp.secure_channel`.
 
     Returns:
         grpc_gcp.Channel: The created gRPC channel.
@@ -190,7 +191,7 @@ def _create_secure_channel(
     # Combine the ssl credentials and the authorization credentials.
     composite_credentials = grpc.composite_channel_credentials(
         ssl_credentials, google_auth_credentials)
-    
+
     if HAS_GRPC_GCP:
         # If grpc_gcp module is available use grpc_gcp.secure_channel,
         # otherwise, use grpc.secure_channel to create grpc channel.
@@ -221,7 +222,7 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
     else:
         credentials = google.auth.credentials.with_scopes_if_required(
             credentials, scopes)
- 
+
     request = google.auth.transport.requests.Request()
 
     return _create_secure_channel(credentials, request, target, **kwargs)
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -154,30 +154,35 @@ def wrap_errors(callable_):
         return _wrap_unary_errors(callable_)
 
 
-def _create_secure_channel(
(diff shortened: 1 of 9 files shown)


api_core/google/api_core/grpc_helpers.py
```diff
@@ -161,7 +212,7 @@ def create_channel(target, credentials=None, scopes=None, **kwargs):
             service. These are only used when credentials are not specified and
             are passed to :func:`google.auth.default`.
         kwargs: Additional key-word args passed to
-            :func:`google.auth.transport.grpc.secure_authorized_channel`.
+            :func:`_create_secure_channel`.
```
theacodes: since this is a public method, we shouldn't reference non-public methods. Maybe change to this `grpc.secure_channel`?

api_core/google/api_core/grpc_helpers.py
```diff
@@ -149,6 +154,52 @@ def wrap_errors(callable_):
         return _wrap_unary_errors(callable_)
 
 
+def _create_secure_channel(
```
theacodes: probably makes sense to just roll the logic in this method into the `create_channel` method. I don't see a good reason for keeping them separate.

spanner/google/cloud/spanner_v1/gapic/spanner_client.py
```diff
@@ -113,10 +121,22 @@ def __init__(self,
 
         # Create the channel.
         if channel is None:
+            options = None
+
+            if HAS_GRPC_GCP:
+                # Initialize grpc gcp config for spanner api.
```
theacodes: I'd be nice if grpc_gcp had a helper that was just:

```python
grpc_gcp_config = grpc_grpc.config_from_text_pb(...)
```

api_core/google/api_core/grpc_helpers.py
WeiranFang: Ack.

api_core/google/api_core/grpc_helpers.py
WeiranFang: _create_secure_channel has an additional arg "ssl_credential=None", do we want to add this to create_channel? Or if it's never a use case for user to pass in a ssl_credential to create_channel(), I can ignore it, and create a new ssl_credential every time. What's your thought?

spanner/google/cloud/spanner_v1/gapic/spanner_client.py
WeiranFang: Good point. I'll add this helper in grpc_gcp

api_core/google/api_core/grpc_helpers.py
theacodes: add the argument

WeiranFang:
Hi @theacodes , I have made some changes as requested. Please take a look, thanks!

spanner/google/cloud/spanner_v1/gapic/spanner_client.py
WeiranFang: It's done. Seems like the above link of this file points to an old commit.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -149,7 +154,11 @@ def wrap_errors(callable_):
         return _wrap_unary_errors(callable_)
 
 
-def create_channel(target, credentials=None, scopes=None, **kwargs):
+def create_channel(target,
+                   credentials=None,
+                   scopes=None,
+                   ssl_credentials=None,
```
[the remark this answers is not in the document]
WeiranFang: done.

WeiranFang:
Hi @theacodes , can you take a look at the new changes?

api_core/tests/unit/test_grpc_helpers.py
```diff
@@ -176,60 +176,137 @@ def test_wrap_errors_streaming(wrap_stream_errors):
     wrap_stream_errors.assert_called_once_with(callable_)
 
 
+@mock.patch('grpc.composite_channel_credentials')
 @mock.patch(
     'google.auth.default',
     return_value=(mock.sentinel.credentials, mock.sentinel.projet))
-@mock.patch('google.auth.transport.grpc.secure_authorized_channel')
-def test_create_channel_implicit(secure_authorized_channel, default):
+@mock.patch('grpc._channel.Channel')
```
theacodes: I don't like mocking private members, can we do this without that?

WeiranFang: Done

PR #5558, opened 2018-06-29
Kokoro Docs Builds

crwilcox:


.kokoro/build.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
theacodes: Call this `build-docs.sh`?

.kokoro/build.sh
```diff
@@ -0,0 +1,26 @@
...
+# Kokoro currently uses 3.6.1
+pyenv global 3.6.1
+
+# Kokoro exposes this as a file, but the scripts expect just a plain variable.
+export GITHUB_TOKEN=$(cat ${KOKORO_GFILE_DIR}/${GITHUB_TOKEN_FILE})
+
+# Add github to known hosts.
+ssh-keyscan github.com >> ~/.ssh/known_hosts
+
+# Activate the ssh key for dpebot. This is need when using synthtool to clone
```
theacodes: Maybe update this comment?

.kokoro/common.cfg
```diff
@@ -0,0 +1,23 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+build_file: "google-cloud-python/.kokoro/build.sh"
```
theacodes: remove `build_file` from here, and put it in `.kokoro/docs.cfg` and `s/build.sh/build-docs.sh`.

.kokoro/common.cfg
```diff
@@ -0,0 +1,23 @@
...
+# Build logs will be here
+action {
+  define_artifacts {
+    regex: "**/*sponge_log.xml"
+  }
+}
+# The github token is stored here.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/dpebot"
+# Common env vars for all repositories and builds.
+env_vars: {
```
theacodes: You'll need to make sure you allow these in the job cfg in your internal CL.

test_utils/scripts/update_docs.sh
```diff
@@ -43,3 +45,46 @@ else
     exit $?
 fi
 
+# Adding GitHub pages branch. `git submodule add` checks it
+# out at HEAD.
+GH_PAGES_DIR='ghpages'
+git submodule add -q -b gh-pages \
```
theacodes: That's surprising, can we just clone it into /tmp or something?

test_utils/scripts/update_docs.sh
```diff
@@ -43,3 +45,46 @@ else
...
+git status
+
+# If there are no changes, just exit cleanly.
+if [[ -z "$(git status --porcelain)" ]]; then
+    echo "Nothing to commit. Exiting without pushing changes."
+    exit
+fi
+
+# Commit to gh-pages branch to apply changes.
+git config --global user.email "googleapis-publisher@google.com"
```
theacodes: Make sure this is DPEBot's email address and username.

test_utils/scripts/update_docs.sh
crwilcox: perhaps. this is just me bringing back the old way of doing it (a few months back in history). I agree we could likely use tmp and not add a submodule

crwilcox: I am going to wait on changing this. It has been done like this before with success. I would also write it differently but I am not sure it is worth rewriting this.

PR #5561, opened 2018-07-02
Bigtable: Instances should support Labels and Type 

### LINKED ISSUE #5454: Bigtable: Fully expose Instance / cluster admin features

maheshgurav:
Support label and type management on Cloud Bigtable instances.  This will partially fix #5454


sduskis:
Please merge.  Also, please add unit test and update the system test.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -191,10 +191,16 @@ def instance(self, instance_id, location=_EXISTING_INSTANCE_LOCATION_ID,
                              characters.) If this value is not set in the
                              constructor, will fall back to the instance ID.
 
+		:type type: str
```
sduskis: This should come from enums, and should not be a string.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -59,15 +60,24 @@ class Instance(object):
                          Cloud Console UI. (Must be between 4 and 30
                          characters.) If this value is not set in the
                          constructor, will fall back to the instance ID.
+
+	:type type: str
```
sduskis: This should be a type in enums, and not a string.

sduskis:
We're going to raise a different PR for this functionality.

PR #5569, opened 2018-07-04
Logging: s/print/logging.info/

### LINKED ISSUE #5565: google.cloud.logging.handlers: send messages to stderr
Would it be ok if we do the print at https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/handlers/transports/background_thread.py#L222
to stderr instead? So it doesn't disturb the output of the software?

Thanks

helen-fornazier:
The function print sends the output to stdout disturbing the output of
the application.
Use logging instead which sends the logs to stderr by default or to the
destination configured by the application.

Closes #5565

### DIFF
--- a/logging/google/cloud/logging/handlers/transports/background_thread.py
+++ b/logging/google/cloud/logging/handlers/transports/background_thread.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 
 import atexit
 import logging
+import sys
 import threading
 import time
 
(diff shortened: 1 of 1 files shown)


theacodes:
@helen-fornazier how would you feel about using `sys.stderr` instead of `logging`?

tseaver:
FWIW, I'd much rather use `sys.stderr`, given the circularity of having our log handlers calling back into the stdlib logging stuff.

helen-fornazier:
I just updated the PR to use `sys.stderr`, could you please take another look?

Thanks

theacodes:
Thank you, @helen-fornazier!

PR #5584, opened 2018-07-11
Allow listing rows on an empty table.

tswast:
Removes check for missing schema on Client.list_rows. This allows the
rows of an empty table to be listed. Such a table might be the
"destination table" for a DML or DDL query.

Closes https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5583

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1400,6 +1400,8 @@ class Client(ClientWithProject):
         elif isinstance(table, TableReference):
             raise ValueError('need selected_fields with TableReference')
         elif isinstance(table, Table):
+            if len(table.schema) == 0 and table.created is None:
+                raise ValueError(_TABLE_HAS_NO_SCHEMA)
             schema = table.schema
         else:
             raise TypeError('table should be Table or TableReference')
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1400,8 +1400,6 @@ def list_rows(self, table, selected_fields=None, max_results=None,
         elif isinstance(table, TableReference):
             raise ValueError('need selected_fields with TableReference')
         elif isinstance(table, Table):
-            if len(table.schema) == 0:
-                raise ValueError(_TABLE_HAS_NO_SCHEMA)
```
tseaver: The reason for the error is that the `_item_to_row` helper relies on the schema being bound to the iterator.  If the schema is empty (e.g., the `Table` instance was created locally without a `reload`), then the rows emitted by the iterator will all be empty too.

tswast: The schema could be empty for two reasons: (1) the table has no schema. This shouldn't cause errors but currently is, and it's affecting DML & DDL queries. (2) The Table object was constructed locally via the Table() constructor, such as is done when you call `create_table`.

You're right that (2) could have unexpected results, though I think it wouldn't necessarily be all that surprising to get a bunch a rows without any columns if you supply a Table object without a schema.

tseaver: Can we distinguish between the cases somehow?  E.g., if the table's `_properties` do not contain a sentinel field (`creationTime`, maybe?) *and* the schema is empty, then raise?

tswast: That seems reasonable. I'll update the PR.

tswast: Updated.

Issue #5589, opened 2018-07-11
BigQuery: 'Client.update_table' systest error, 412 PreconditionFailed


tseaver:
Second error in [this CI run](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/7126):

```python
____________________ TestBigQuery.test_update_table_schema _____________________

self = <tests.system.TestBigQuery testMethod=test_update_table_schema>

    def test_update_table_schema(self):
        dataset = self.temp_dataset(_make_dataset_id('update_table'))
    
        TABLE_NAME = 'test_table'
        table_arg = Table(dataset.table(TABLE_NAME), schema=SCHEMA)
        self.assertFalse(_table_exists(table_arg))
        table = retry_403(Config.CLIENT.create_table)(table_arg)
        self.to_delete.insert(0, table)
        self.assertTrue(_table_exists(table))
        voter = bigquery.SchemaField('voter', 'BOOLEAN', mode='NULLABLE')
        schema = table.schema
        schema.append(voter)
        table.schema = schema
    
>       updated_table = Config.CLIENT.update_table(table, ['schema'])
,,,
        if not 200 <= response.status_code < 300:
>           raise exceptions.from_http_response(response)
E           PreconditionFailed: 412 PATCH https://www.googleapis.com/bigquery/v2/projects/precise-truck-742/datasets/update_table_7126_1531263328/tables/test_table: Precondition Failed

../.nox/sys-2-7/lib/python2.7/site-packages/google/cloud/_http.py:293: PreconditionFailed
```

tseaver:
Same error occurs in failing snippets tests in that CI run.

Issue #5588, opened 2018-07-11
BigQuery: 'Client.update_dataset' systest error, 412 PreconditionFailed


tseaver:
First error in [this CI run](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/7126):

```python
_______________________ TestBigQuery.test_update_dataset _______________________

self = <tests.system.TestBigQuery testMethod=test_update_dataset>

    def test_update_dataset(self):
        dataset = self.temp_dataset(_make_dataset_id('update_dataset'))
        self.assertTrue(_dataset_exists(dataset))
        self.assertIsNone(dataset.friendly_name)
        self.assertIsNone(dataset.description)
        self.assertEquals(dataset.labels, {})
    
        dataset.friendly_name = 'Friendly'
        dataset.description = 'Description'
        dataset.labels = {'priority': 'high', 'color': 'blue'}
        ds2 = Config.CLIENT.update_dataset(
            dataset,
>           ('friendly_name', 'description', 'labels'))

tests/system.py:189: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
google/cloud/bigquery/client.py:387: in update_dataset
    headers=headers)
...    
        if not 200 <= response.status_code < 300:
>           raise exceptions.from_http_response(response)
E           PreconditionFailed: 412 PATCH https://www.googleapis.com/bigquery/v2/projects/precise-truck-742/datasets/update_dataset_7126_1531263324: Precondition Failed

../.nox/sys-2-7/lib/python2.7/site-packages/google/cloud/_http.py:293: PreconditionFailed
```

tseaver:
Same error occurs in failing snippets test in that CI run.

tseaver:
Same error occurs in failing snippets test in that CI run.

tswast:
I think this is a regression in the backend where the REST call to create table doesn't return the correct etag for the patch call.

tswast:
I've pinged the BigQuery on-call.

tswast:
I thought I might be able to work around this by calling `get_dataset()` before calling `update_dataset()`, but I still have flakes. I'll send a PR to disable these tests until the BQ team pushes a fix.

hzitoun:
[...]

PR #5590, opened 2018-07-11
Skip update_table and update_dataset tests until etag issue is resolved.

tswast:
Workaround for #5588 and #5589.

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -174,6 +174,9 @@ class TestBigQuery(unittest.TestCase):
         self.assertEqual(got.friendly_name, 'Friendly')
         self.assertEqual(got.description, 'Description')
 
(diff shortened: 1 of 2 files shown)


tseaver:
@tswast Any word on the underlying ETag issue?  I'm assuming that the resources returned from `tables.create` / `datasets.create` APIs did not include a valid ETag?

tswast:
The underlying problem: an API change was rolling out that had a side-effect of changing the etag generation logic. A request to `tables.create` could go to the new server and get the etag for the new version and the subsequent request to `tables.update` could get routed to the old server, which doesn't know about the new etag logic and fails the write.

Now that the API rollout is done, these tests can be re-enabled, but the long-term fix would be to add a retry loop around a `table.get` + `table.update` pair.

PR #5592, opened 2018-07-11
Add snippet using 'source_langugae' argument to 'Client.translate'.

### LINKED ISSUE #5324: translate: documentation of python API client to show the "source_language" argument
The python API client features a "source_language" argument in the "translate" function.
I found out about it by going to the python code.
It shows up in the documentation under [REST reference](https://cloud.google.com/translate/docs/reference/translate) as `source`, but not in the [python example](https://cloud.google.com/translate/docs/reference/translate).

tseaver:
Closes #5324.


tseaver:
@theacodes PTAL

Issue #5583, opened 2018-07-10
BigQuery: list_rows() raises a ValueError for empty tables


tswast:
```
In [7]: table = client.create_table(bigquery.Table(client.dataset('test_dataset').table('empty_table')))

In [10]: client.list_rows(table)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-6c5fe985ca74> in <module>()
----> 1 client.list_rows(table)

~/.pyenv/versions/3.6.4/envs/scratch/lib/python3.6/site-packages/google/cloud/bigquery/client.py in list_rows(self, table, selected_fields, max_results, page_token, start_index, page_size, retry)
   1403         elif isinstance(table, Table):
   1404             if len(table.schema) == 0:
-> 1405                 raise ValueError(_TABLE_HAS_NO_SCHEMA)
   1406             schema = table.schema
   1407         else:

ValueError: Table has no schema:  call 'client.get_table()'
```

This may not seem like much of a problem, but it prevents DDL statements from being used from the `%%bigquery` magics.

tswast:
Instead, when a table has no rows (whether or not it has a schema), it should return an empty iterable.

tswast:
I've written a failing system test.

    def test_list_rows_empty_table(self):
        from google.cloud.bigquery.table import RowIterator

        dataset_id = _make_dataset_id('empty_table')
        dataset = self.temp_dataset(dataset_id)
        table_ref = dataset.table('empty_table')
        table = Config.CLIENT.create_table(bigquery.Table(table_ref))

        # It's a bit silly to list rows for an empty table, but this does
        # happen as the result of a DDL query from an IPython magic command.
        rows = Config.CLIENT.list_rows(table)
        self.assertIsInstance(rows, RowIterator)
        self.assertEqual(tuple(rows), ())

So that `list_rows().to_dataframe()` doesn't break, somehow we'll need to return an empty RowIterator.

PR #5602, opened 2018-07-12
Don't raise exception when fetching CREATE VIEW query results

tswast:
CREATE VIEW DDL queries set a "destination table" field, but that table
is actually a view. If you attempt to list rows on the view table, it
results in an error.

The fix is to avoid listing rows altogether if `getQueryResults()` says
that a query has completed but the number of rows in the result set is
undefined.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -31,6 +31,7 @@ from google.cloud.bigquery.query import StructQueryParameter
 from google.cloud.bigquery.query import UDFResource
 from google.cloud.bigquery.retry import DEFAULT_RETRY
 from google.cloud.bigquery.schema import SchemaField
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2505,12 +2506,17 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
             self._query_results = self._client._get_query_results(
                 self.job_id, retry, project=self.project,
                 location=self.location)
+
+        # If the query job is complete but there are no query results, this was
+        # special job, such as a DDL query. Return an empty result set to
+        # indicate success and avoid calling tabledata.list on a table which
+        # can't be read (such as a view table).
+        if self._query_results.total_rows is None:
+            return _EmptyRowIterator()
```
tseaver: Yay! this is a much better fix for #5583 than #5584.

PR #5607, opened 2018-07-13
Fixing the broken Bigtable system test.

### LINKED ISSUE #5593: Bigtable:  new system test failure
#5513 caused a new (non-flaky, AFAICT) [system test failure](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/7142):

```python

    def test_create_table_with_split_keys(self):
        temp_table_id = 'foo-bar-baz-split-table'
        initial_split_keys = [b'split_key_1', b'split_key_10',
                              b'split_key_20', b'']
        temp_table = Config.INSTANCE.table(temp_table_id)
        temp_table.create(initial_split_keys=initial_split_keys)
        self.tables_to_delete.append(temp_table)
    
        # Read Sample Row Keys for created splits
        sample_row_keys = temp_table.sample_row_keys()
        self.assertEqual(set([srk.row_key for srk in sample_row_keys]),
>                        set(initial_split_keys))
E       AssertionError: Items in the second set but not the first:
E       'split_key_1'
E       'split_key_10'
E       'split_key_20'
```

sduskis:
This fixes #5593.

### DIFF
--- a/bigtable/tests/system.py
+++ b/bigtable/tests/system.py
@@ -263,15 +263,19 @@ class TestTableAdminAPI(unittest.TestCase):
     def test_create_table_with_split_keys(self):
         temp_table_id = 'foo-bar-baz-split-table'
         initial_split_keys = [b'split_key_1', b'split_key_10',
-                              b'split_key_20', b'']
+                              b'split_key_20']
         temp_table = Config.INSTANCE.table(temp_table_id)
         temp_table.create(initial_split_keys=initial_split_keys)
         self.tables_to_delete.append(temp_table)
 
         # Read Sample Row Keys for created splits
         sample_row_keys = temp_table.sample_row_keys()
-        self.assertEqual(set([srk.row_key for srk in sample_row_keys]),
-                         set(initial_split_keys))
+        actual_keys = [srk.row_key for srk in sample_row_keys]
+
+        expected_keys = initial_split_keys
+        expected_keys.append(b'')
+
+        self.assertEqual(actual_keys, expected_keys)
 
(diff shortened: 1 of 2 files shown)


bigtable/tests/system.py
```diff
@@ -263,15 +263,19 @@ def test_create_table_with_families(self):
...
         # Read Sample Row Keys for created splits
         sample_row_keys = temp_table.sample_row_keys()
-        self.assertEqual(set([srk.row_key for srk in sample_row_keys]),
-                         set(initial_split_keys))
+        actual_keys = [srk.row_key for srk in sample_row_keys]
+
+        expected_keys = initial_split_keys
+        expected_keys.append(b'')
+
+        self.assertEqual(actual_keys, expected_keys)
```
tseaver: Looking at the [API reference for `initial_splits`](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.CreateTableRequest.FIELDS.repeated.google.bigtable.admin.v2.CreateTableRequest.Split.google.bigtable.admin.v2.CreateTableRequest.initial_splits), I'm not sure I grok why passing the empty key broke things.   However, I can confirm that the tests do pass with this change.

sduskis: It's strange behavior.  An empty string means either beginning or end of a table, depending on location.  There's an implicit empty string at the beginning and end of the Split list.  It's an error condition, and the service should probably either ignore the empty string at the beginning or end of the list, or return an appropriate status code.

I raised a bug internally for this behavior.

PR #5619, opened 2018-07-15
Bigtable: Ability to create an instance with multiple clusters

sumit-ql:
Updated instance create() method to accept list of clusters.
#5454


sumit-ql:
Something is wrong, too many commits(2,401 ) in this PR. Closing this pull request, will create a new one.

PR #5614, opened 2018-07-13
Bigtable:  Add 'instance_type', 'labels' to 'Instance' ctor

AVaksman:
Added optional type and labels to instance

Toward #5454.

PR #5622, opened 2018-07-16
Bigtable: Ability to create an instance with multiple clusters

sumit-ql:
Updated instance create() method to accept clusters list. 
#5454

### DIFF
--- a/bigtable/google/cloud/bigtable/cluster.py
+++ b/bigtable/google/cloud/bigtable/cluster.py
@@ -16,6 +16,8 @@
 
 
 import re
+from google.cloud.bigtable_admin_v2 import enums
+from google.cloud.bigtable_admin_v2.types import instance_pb2
 
 
 _CLUSTER_NAME_RE = re.compile(r'^projects/(?P<project>[^/]+)/'
@@ -24,6 +26,8 @@ _CLUSTER_NAME_RE = re.compile(r'^projects/(?P<project>[^/]+)/'
 
 DEFAULT_SERVE_NODES = 3
 """Default number of nodes to use when creating a cluster."""
+
+_STORAGE_TYPE_UNSPECIFIED = enums.StorageType.STORAGE_TYPE_UNSPECIFIED
 
 
 class Cluster(object):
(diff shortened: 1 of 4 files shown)


bigtable/google/cloud/bigtable/cluster.py
```diff
@@ -42,17 +46,35 @@ class Cluster(object):
...
+                        https://cloud.google.com/bigtable/docs/locations
+                        For best performance, clients should be located
+                        as close as possible to this cluster.
+
     :type serve_nodes: int
     :param serve_nodes: (Optional) The number of nodes in the cluster.
                         Defaults to :data:`DEFAULT_SERVE_NODES`.
+
+    :type default_storage_type: int
+    :param default_storage_type: (Optional) The default values are
```
tseaver: This parameter is actually named `storage_type` below.

bigtable/google/cloud/bigtable/cluster.py
```diff
@@ -172,3 +194,16 @@ def delete(self):
         """
         client = self._instance._client
         client.instance_admin_client.delete_cluster(self.name)
+
+    def create_pb_request(self):
```
tseaver: Should this method be private?  ISTM that the only caller is going to be `Instance.create`.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -167,48 +164,24 @@ def create(self, location_id=_EXISTING_INSTANCE_LOCATION_ID,
...
-        instance = instance_pb2.Instance(
-            display_name=self.display_name
-        )
-        clusters[cluster_id] = cluster
+        clusters_dict = {}
+        for cluster in clusters:
+            clusters_dict[cluster.cluster_id] = cluster.create_pb_request()
+
+        instance = instance_pb2.Instance(display_name=self.display_name)
+
```
tseaver: This is going to break existing code.  Should there be a fallback / default for users creating an instance without supplying explicit clusters?

bigtable/tests/unit/test_cluster.py
```diff
@@ -64,8 +65,8 @@ def test_constructor_non_default(self):
         instance = _Instance(self.INSTANCE_ID, client)
 
         cluster = self._make_one(self.CLUSTER_ID, instance,
-                                 serve_nodes=SERVE_NODES)
-        self.assertEqual(cluster.cluster_id, self.CLUSTER_ID)
+                                 self.LOCATION_ID, serve_nodes=SERVE_NODES)
+        self.assertEqual(cluster.cluster_id, self.CLUSTER_ID, self.LOCATION_ID)
```
tseaver: That third argument doesn't look right (it would replace the standard assertion error message).

tseaver:
@sduskis This PR will conflict with #5614.

tseaver:
@sumit-ql This PR needs to be rebased against the current `master` after merge of #5614, and conflicts resolved.

tseaver:
@sumit-ql Please hold off rebasing against `master` until I land #5641:  your changes should make that fix irrelevant, but we need to get the `master` branch green ASAP.

tseaver:
@sumit-ql OK, #5641 is merged.

bigtable/google/cloud/bigtable/cluster.py
sumit-ql: Done.

bigtable/google/cloud/bigtable/cluster.py
sumit-ql: Yes it should be private. Done

bigtable/google/cloud/bigtable/instance.py
sumit-ql: Added a new method "create_with_multiple_clusters()". That way, existing client will not break.

bigtable/tests/unit/test_cluster.py
sumit-ql: Removed self.LOCATION_ID, thanks.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -240,6 +241,44 @@ def create(self, location_id=_EXISTING_INSTANCE_LOCATION_ID,
             parent=parent, instance_id=self.instance_id, instance=instance,
             clusters=clusters)
 
+    def create_with_multiple_clusters(self, clusters):
```
sduskis: @tseaver, eventually, this should be the main implementation of `create()`.  How do you feel about turning `clusters` into a member variable, and using it in create().  It will require some validation that the optional parameters aren't used in conjunction with clusters, but ultimately, IMHO, will produce a better experience.

bigtable/google/cloud/bigtable/instance.py
```diff
@@ -220,6 +220,7 @@ def create(self, location_id=_EXISTING_INSTANCE_LOCATION_ID,
         :returns: The long-running operation corresponding to the create
                     operation.
         """
+
         clusters = {}
         cluster_id = '{}-cluster'.format(self.instance_id)
```
sduskis: Please create a `Cluster` object with all of these variables, and add it to an array of clusters, and use that array instead of having duplicate code.

sumit-ql: Done, thanks.

sduskis:
@tseaver, it looks good to me.  WDYT?

PR #5625, opened 2018-07-16
Fix generating signed urls for blobs with non-ascii names

dkasyanov:


dkasyanov:
I signed it!

storage/tests/unit/test_blob.py
```diff
@@ -383,6 +383,35 @@ def test_generate_signed_url_w_credentials(self):
         credentials = object()
         self._basic_generate_signed_url_helper(credentials=credentials)
 
+    def test_generate_signed_url_non_ascii(self):
+        BLOB_NAME = u'Caf\xe9.txt'
+        EXPIRATION = '2014-10-16T20:34:37.000Z'
+        connection = _Connection()
+        client = _Client(connection)
+        bucket = _Bucket(client)
+        blob = self._make_one(BLOB_NAME.encode('utf-8'), bucket=bucket)
```
tseaver: There is no need to encode the name here:  the `Blob` constructor just decodes it if it is bytes.

tseaver:
Thanks for the patch!  While trying to understand the change you are proposing, I wrote a new [system test](https://gist.github.com/tseaver/8fc69f3c86df400e97d1c80ff8677877) which creates a blob with the same name as your unit test, uploads a known text to it, generates a signed URL from it, fetches the signed URL, and verifies the payload.  That system test passes without your patch:

```
tests/system.py::TestStorageSignURLs::test_create_signed_delete_url PASSED [ 33%]
tests/system.py::TestStorageSignURLs::test_create_signed_read_url PASSED [ 66%]
tests/system.py::TestStorageSignURLs::test_create_signed_read_url_w_non_ascii_name PASSED [100%]
```

Can you provide more information why the patch is needed?

tseaver:
@dkasyanov Thanks, my system test fails like yours on Python 2.7.

Can you please address my review comment -- I will then merge your PR, followed by #5626 (after rebasing to pick up your change).

dkasyanov:
@tseaver Thank you.
Change updated.

PR #5626, opened 2018-07-16
Add system test: signed URLs w/ non-ASCII blob name.

tseaver:
See #5625.


frankyn:
Is this PR confirming #5625 given circleci fails?

tseaver:
> Is this PR confirming #5625 given circleci fails?

@frankyn Yes, it confirms the underlying problem on Python 2.7 (I only ran the 3.6 system tests locally, as that was where I *expected* the issue).

frankyn:
Apologies Tres, this did LGTM.

PR #5633, opened 2018-07-16
Pubsub: Make 'Message.publish_time' return datetime

### LINKED ISSUE #5598: PubSub: Message publish_time return type unexpected
I am working on a PubSub project and was a bit confused about the Message `publish_time`. I expected to get `datetime` per [the docstring](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1acc8c22664229b6681ff91654932998e611e1c2/pubsub/google/cloud/pubsub_v1/subscriber/message.py#L152) for the method. Instead I got a `Timestamp` that apparently comes from protobuf types. I found there is even a [unit test](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1acc8c22664229b6681ff91654932998e611e1c2/pubsub/tests/unit/pubsub_v1/subscriber/test_message.py#L49) that verifies that `publish_time` returns a `Timestamp`.

Can we return `datetime`? Or should the docs just be updated to explain what is actually being returned? I'd prefer the former but wanted to ask before submitting a PR, given the inconsistency.

tseaver:
Closes #5598

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/message.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/message.py
@@ -14,10 +14,12 @@
 
 from __future__ import absolute_import
 
+import datetime
 import json
 import math
 import time
 
+from google.api_core import datetime_helpers
 from google.cloud.pubsub_v1.subscriber._protocol import requests
 
 
@@ -151,7 +153,11 @@ class Message(object):
         Returns:
             datetime: The date and time that the message was published.
         """
-        return self._message.publish_time
+        timestamp = self._message.publish_time
+        delta = datetime.timedelta(
+            seconds=timestamp.seconds,
+            microseconds=timestamp.nanos // 1000)
+        return datetime_helpers._UTC_EPOCH + delta
 
     @property
     def size(self):
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_message.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_message.py
@@ -12,25 +12,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import time
 
 import mock
+import pytz
 from six.moves import queue
 
+from google.api_core import datetime_helpers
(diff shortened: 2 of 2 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -151,7 +153,11 @@ def publish_time(self):
         Returns:
             datetime: The date and time that the message was published.
         """
-        return self._message.publish_time
+        timestamp = self._message.publish_time
+        delta = datetime.timedelta(
+            seconds=timestamp.seconds,
+            microseconds=timestamp.nanos // 1000)
```
tswast: Instead of removing precision, should this be a DatetimeWithNanoseconds as defined in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4979?

tseaver: @tswast I considered it.  The issue is that there is no clean / efficient way to create an instance of `DateTimeWithNanoseconds` given only a delta:  there is no corresponding `TimeDeltaWIthNanoseconds`, and it doesn't override `__add__` etc. anyway. That class is defined in `api_core`, which we have been somewhat loath to bump.

To work around that deficit here, we would need to create a "normal" `datetime.datetime` first, then use its attributes to generate the constructor arguments for the `DateTimeWithNanoseconds` instance.  Does that seem worthwhile here?  I'm unclear how the field will be used.

tswast: SGTM to use normal datetime. If it becomes an issue in the future we can always use DateTimeWithNanoseconds then.

Publish time could be used to order events, but it's not as important here as it would be for Spanner. In Spanner the nanosecond precision is important because it is used for consistency of writes.

tseaver: I don't think we use `DatetimeWithNanos` for the snapshot consistency bits in spanner:  it only gets used for marshalling parameters and column values.

PR #5642, opened 2018-07-17
Generate automl v1beta1

crwilcox:


kinwa91:
Did only a cursory review, not familiar with python library generation :(

crwilcox:
Thanks @kinwa91 . I figured this was an opportunity to show you a few of our synth files and see what is going on on the other side :)

PR #5651, opened 2018-07-19
BigTable: Create MutationBatcher for bigtable

aneepct:
Provide a facility for batching large numbers of mutations into configurable batch sizes, taking row count, max MB, and max mutations into account.

Each batch will contain a list of DirectRow's which will be sent to the mutate_rows() method on Table.


bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,88 @@
...
+                 max_mutations=MAX_MUTATIONS, max_row_bytes=MAX_ROW_BYTES):
+        self.rows = []
+        self.total_mutation_count = 0
+        self.total_size = 0
+        self.table = table
+        self.flush_count = flush_count
+        self.max_mutations = max_mutations
+        self.max_row_bytes = max_row_bytes
+
+    def add_row(self, row):
```
sduskis: Please call this method `mutate`

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,88 @@
...
+            self.finish_batch()
+
+        self.rows.append(row)
+        self.total_mutation_count += mutation_count
+        self.total_size += mutation_size
+
+        if len(self.rows) >= self.flush_count:
+            self.finish_batch()
+
+    def finish_batch(self):
```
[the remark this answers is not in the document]
sduskis: Please call this method `flush()`

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,88 @@
...
+
+import sys
+
+FLUSH_COUNT = 1000
+MAX_MUTATIONS = 100000
+MAX_ROW_BYTES = 5242880  # 5MB
+
+
+class MutationsBatcher(object):
+    """ Batch mutations using limits MAX_ROW_BYTES, MAX_MUTATIONS and
```
sduskis: Please change this comment to something like:

" A MutationsBatcher is used in batch cases where the number of mutations is large or unknown.  It will store `DirectRow`s in memory until one of the size limits is reached, or an explicit call to `flush()` is performed.  When a flush event occurs, the `DirectRow`s in memory will be sent to Cloud Bigtable.  Batching mutations is more efficient than sending individual req

This class is not suited for usage in systems where each mutation needs to guaranteed to be sent, since calling `mutate` may only result in an in-memory change.  In a case of a system crash, any `DirectRow`s remaining in memory will not necessarily be sent to the service, even after the completion of the `mutate()` method.

TODO: Performance would dramatically improve if this class had the capability of asynchronous, parallel RPCs. 
"

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,250 @@
...
+        data_api = bigtable_client.BigtableClient(mock.Mock())
+        table_api = mock.create_autospec(
+            bigtable_table_admin_client.BigtableTableAdminClient)
+        credentials = _make_credentials()
+        client = self._make_client(project='project-id',
+                                   credentials=credentials, admin=True)
+        client._table_data_client = data_api
+        client._table_admin_client = table_api
+        instance = client.instance(instance_id=self.INSTANCE_ID)
+        table = self._make_table(self.TABLE_ID, instance)
```
sduskis: This test can be simplified by mocking a Table object.  There's no need to mock all of the other things in this and other tests.

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,200 @@
...
+
+        mutation_batcher = MutationsBatcher(table)
+        self.assertEqual(table.table_id, mutation_batcher.table.table_id)
+
+    def test_add_row(self):
+        from google.cloud.bigtable.batcher import MutationsBatcher
+        from google.cloud.bigtable.row import DirectRow
+
+        table = _Table(self.TABLE_NAME)
```
zakons: I recommend that you create methods for the common code for each test to reduce the repetitive code.

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,104 @@
...
+        :param row: class:`~google.cloud.bigtable.row.DirectRow`.
+
+        :raises: One of the following:
+                 * :exc:`~.table._BigtableRetryableError` if any
+                   row returned a transient error.
+                 * :exc:`RuntimeError` if the number of responses doesn't
+                   match the number of rows that were retried
+        """
+        mutation_size = 0
+        for mutation in row._get_mutations():
```
sduskis: Please move this to below the `MAX_MUTATIONS` check.

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,186 @@
...
+        mutation_batcher = self._make_batcher(table, flush_count=3)
+
+        rows = self._make_rows(table)
+
+        for row in rows:
+            mutation_batcher.mutate(row)
+
+        with mock.patch('google.cloud.bigtable.table.Table.name',
+                        new=self.TABLE_NAME):
+            mutation_batcher.flush()
```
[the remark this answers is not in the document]
aneepct: If there are 4 rows and max flush_count is 3 then the first batch flush() called after reach the MAX_FLUSH. Then after there will be only one row left which needs call flush() at the end because it will not reach the max flush_count on second batch.

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        mutation_count = len(row._get_mutations())
+        if mutation_count > MAX_MUTATIONS:
+            raise MaxMutaionsError
+
+        if (self.total_mutation_count + mutation_count) >= MAX_MUTATIONS:
+            self.flush()
+
+        mutation_size = 0
+        for mutation in row._get_mutations():
```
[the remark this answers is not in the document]
aneepct: @sduskis do you want me to create new method on DirectRow which can return total size of mutations on row? (like `_get_mutations_size()`)

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,107 @@
...
+        self.max_row_bytes = max_row_bytes
+
+    def mutate(self, row):
+        """ Add a row to the batch. If the current batch meets one of the size
+        limits, the batch is sent synchronously.
+
+        :type row: class
+        :param row: class:`~google.cloud.bigtable.row.DirectRow`.
+
+        :raises: One of the following:
```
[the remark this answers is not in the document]
aneepct: @sduskis added this

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,107 @@
...
+                 max_row_bytes=MAX_ROW_BYTES):
+        self.rows = []
+        self.total_mutation_count = 0
+        self.total_size = 0
+        self.table = table
+        self.flush_count = flush_count
+        self.max_row_bytes = max_row_bytes
+
+    def mutate(self, row):
+        """ Add a row to the batch. If the current batch meets one of the size
```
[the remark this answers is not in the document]
aneepct: @sduskis example from test cases or the more description about batching?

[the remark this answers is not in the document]
aneepct: @sduskis Done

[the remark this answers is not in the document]
sduskis: A simpler example would be better.  Users won't actually set max row bytes or flush_count.  Just a simple `rows = ... # large number of rows`

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,133 @@
...
+                    row.row_key, mutation_count), )
+
+        if (self.total_mutation_count + mutation_count) >= MAX_MUTATIONS:
+            self.flush()
+
+        self.rows.append(row)
+        self.total_mutation_count += mutation_count
+        self.total_size = row.get_mutations_size()
+
+        if self.total_size >= self.max_row_bytes:
```
sduskis:
```
if (self.total_size >= self.max_row_bytes or
    len(self.rows) >= self.flush_count):
```
might be nicer.

bigtable/google/cloud/bigtable/batcher.py
```diff
@@ -0,0 +1,133 @@
...
+            raise MaxMutaionsError(
+                'The row key {} exceeds the number of mutations {}.'.format(
+                    row.row_key, mutation_count), )
+
+        if (self.total_mutation_count + mutation_count) >= MAX_MUTATIONS:
+            self.flush()
+
+        self.rows.append(row)
+        self.total_mutation_count += mutation_count
+        self.total_size = row.get_mutations_size()
```
sduskis: `self.total_size += row.get_mutations_size()`

bigtable/google/cloud/bigtable/table.py
```diff
@@ -483,6 +485,26 @@ def drop_by_prefix(self, row_key_prefix, timeout=None):
             table_admin_client.drop_row_range(
                 self.name, row_key_prefix=_to_bytes(row_key_prefix))
 
+    def batcher(self, flush_count=FLUSH_COUNT, max_row_bytes=MAX_ROW_BYTES):
```
sduskis: s/batcher/mutationBatcher/

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_add_row_with_max_flush_count(self):
+        table = _Table(self.TABLE_NAME)
+        mutation_batcher = self._make_batcher(table, flush_count=3)
+
+        rows = self._make_rows(table)
+
+        for row in rows:
+            mutation_batcher.mutate(row)
+
+        mutation_batcher.flush()
```
sduskis: please remove the extra `mutation_batcher.flush()`.  There's no need for it in this test.

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_add_row_with_max_row_bytes(self):
+        table = _Table(self.TABLE_NAME)
+        mutation_batcher = self._make_batcher(table,
+                                              max_row_bytes=1 * 1024 * 1024)
+
+        number_of_bytes = 2 * 1024 * 1024
+        max_value = b'1' * number_of_bytes
+
+        rows = self._make_rows(table)
+        rows[0].set_cell('cf1', b'c1', max_value)
```
sduskis: Please add multiple slightly smaller rows rather than a single large one to test the logic.

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,182 @@
...
+
+        for row in rows:
+            mutation_batcher.mutate(row)
+
+        mutation_batcher.flush()
```
sduskis: Please remove the call to `flush()` here, since we're just testing the `mutate` logic now.

bigtable/google/cloud/bigtable/table.py
aneepct: @sduskis Please explain this change

bigtable/tests/unit/test_batcher.py
aneepct: @sduskis Done

bigtable/tests/unit/test_batcher.py
aneepct: @sduskis Done

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,181 @@
...
+    @staticmethod
+    def _get_target_batcher_class():
+        from google.cloud.bigtable.batcher import MutationsBatcher
+
+        return MutationsBatcher
+
+    def _make_batcher(self, *args, **kwargs):
+        return self._get_target_batcher_class()(*args, **kwargs)
+
+    def _make_rows(self, table):
```
sduskis: Please remove `table`.

bigtable/tests/unit/test_batcher.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertEqual(table.mutation_calls, 1)
+
+    @mock.patch('google.cloud.bigtable.batcher.MAX_MUTATIONS', new=3)
+    def test_add_row_with_max_mutations_failure(self):
+        from google.cloud.bigtable.batcher import MaxMutaionsError
+
+        table = _Table(self.TABLE_NAME)
+        mutation_batcher = self._make_batcher(table)
+
+        rows = self._make_rows(table)
```
sduskis: Please just create a single DirectRow instead of calling `_make_rows`

bigtable/tests/unit/test_batcher.py
sduskis: Also, please add an explicit count instead of always returning 4 rows with 1 cell.

tseaver:
@sduskis, @aneepct: please rebase the PR against the current `master` (so that CI can detect which package is affected).

bigtable/google/cloud/bigtable/table.py
sduskis: Please rename the `batcher` method to `mutationsBatcher`

tseaver:
> Please rename the `batcher` method to `mutationsBatcher`.

@aneepct, @sduskis `mutations_batcher` would be preferable: we don't use `camelCase` in method names, per PEP 8.

sduskis:
@aneepct, are you done with all of your intended changes?

bigtable/google/cloud/bigtable/batcher.py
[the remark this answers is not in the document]
aneepct: @sduskis Done

aneepct:
@sduskis Everything was done please add more details.

Issue #5646, opened 2018-07-18
BigQuery: add 'QueryJob.estimates_bytes_processed' property


tseaver:
See [API ref](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#statistics.query.estimatedBytesProcessed).

Read-only, based on the `statistics.query.estimatedBytesProcessed` field.

PR #5654, opened 2018-07-19
BigQuery:  add support for job labels

### LINKED ISSUE #5645: BigQuery: add 'labels' properties to job and job config classes
See the [API ref](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.labels)

- Each job config class should have a writable `labels` property, taking / returning a mapping, `{str: str}`.

- Each job class should have a read-only `lables` property, returning a mapping, `{str: str}`.

tseaver:
The huge majority of this PR is backfilling explicit unit test support for `_AsyncJob` and `_JobConfig`.  I uncovered some doc nits because of that work, as well.

Closes #5645.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -343,7 +343,7 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
     @property
     def labels(self):
         """Dict[str, str]: Labels for the job."""
-        return self._properties.get('labels', {})
+        return self._properties.setdefault('labels', {})
 
     @property
     def etag(self):
@@ -718,7 +718,7 @@ class _JobConfig(object):
         Raises:
             ValueError: If ``value`` type is invalid.
         """
-        return self._properties.get('labels', {})
+        return self._properties.setdefault('labels', {})
 
     @labels.setter
     def labels(self, value):
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -707,6 +707,25 @@ class _JobConfig(object):
         self._job_type = job_type
         self._properties = {job_type: {}}
 
+    @property
+    def labels(self):
+        """Dict[str, str]: Labels for the job.
+
+        This method always returns a dict. To change a job's labels,
+        modify the dict, then call ``Client.update_job``. To delete a
+        label, set its value to :data:`None` before updating.
+
+        Raises:
+            ValueError: If ``value`` type is invalid.
+        """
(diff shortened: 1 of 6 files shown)


tseaver:
@tswast I just rebased to drop the commits shared with #5646.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -702,6 +707,25 @@ def __init__(self, job_type):
...
+        """Dict[str, str]: Labels for the job.
+
+        This method always returns a dict. To change a job's labels,
+        modify the dict, then call ``Client.update_job``. To delete a
+        label, set its value to :data:`None` before updating.
+
+        Raises:
+            ValueError: If ``value`` type is invalid.
+        """
+        return self._properties.get('labels', {})
```
tswast: Per the docstring, you should be able to modify the dictionary in-place. I think that use-case won't work when the dictionary is originally not present. Maybe explicitly check for the `labels` key and set to `{}` if not present instead of using `.get()`?

tseaver: 076be15 Updates *all* `labels` getters to use `setdefault`, and adds tests for updating in place.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -702,6 +707,25 @@ def __init__(self, job_type):
         self._job_type = job_type
         self._properties = {job_type: {}}
 
+    @property
+    def labels(self):
+        """Dict[str, str]: Labels for the job.
+
+        This method always returns a dict. To change a job's labels,
+        modify the dict, then call ``Client.update_job``. To delete a
```
rarredon, line 715: I'm trying to use this feature to update a job's labels from the Python client, but there does not seem to be a method `Client.update_job`. Is this feature incomplete?

plamut: @rarredon A very late answer, but if it still helps - there was a [bug](https://github.com/googleapis/python-bigquery/issues/469) opened that tracks this.

I doesn't seem that jobs can be modified, though, since the [jobs API](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs) does not support PATCH-ing them (unless there is an issue with the docs themselves).

BTW, for future reference, its best to ask questions in the new repository's [issue tracker](https://github.com/googleapis/python-bigquery/issues). This is an old PR in the old repository that most library maintainers rarely monitor (if at all).

rarredon: Thanks @plamut! I managed to find a workaround. I will add a comment to the issue you linked.

PR #5665, opened 2018-07-20
Synth.py and Generate KMS v1

crwilcox:


crwilcox:
This was a pretty clean generation. There was only one fix needed in `synth.py`

kms/google/cloud/kms_v1/gapic/key_management_service_client.py
```diff
@@ -0,0 +1,1762 @@
...
+            self.transport = key_management_service_grpc_transport.KeyManagementServiceGrpcTransport(
+                address=self.SERVICE_ADDRESS,
+                channel=channel,
+                credentials=credentials,
+            )
+
+        if client_info is None:
+            client_info = (
+                google.api_core.gapic_v1.client_info.DEFAULT_CLIENT_INFO)
+        client_info.gapic_version = _GAPIC_LIBRARY_VERSION
```
tseaver: This bit is problematic -- the `DEFAULT_CLIENT_INFO` instance is going to be scribbled on by every library generated this way, which means that each time an app instantiates a generated client, they *all* use the last-instantiated one's `gapic_version`.

Rather than reuse the instance, we should be constructing a *new* one here, e.g.:

```python
        if client_info is None:  # create a new instance
            client_info = (
                google.api_core.gapic_v1.client_info.ClientInfo(
                    gapic_version=_GAPIC_LIBRARY_VERSION)
        else:  # update caller-supplied instance
            client_info.gapic_version = _GAPIC_LIBRARY_VERSION
```

crwilcox: It seems we do this in released clients today just like this. I can make a bug on gapic-generator for it though.
Here is an example of us doing this:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/dlp/google/cloud/dlp_v2/gapic/dlp_service_client.py#L173

crwilcox: https://github.com/googleapis/gapic-generator/issues/2161

PR #5666, opened 2018-07-20
regenerate dlp v2 library

crwilcox:


dlp/README.rst
```diff
@@ -61,38 +63,15 @@ Windows
     <your-env>\Scripts\activate
     <your-env>\Scripts\pip.exe install google-cloud-dlp
 
-Preview
```
crwilcox: I can adjust the synth script to not move this file. this example might be useful.

dlp/nox.py
```diff
@@ -18,35 +18,9 @@
...
-        '--quiet',
-        '--cov=google.cloud.dlp_v2',
-        '--cov-append',
-        '--cov-config=.coveragerc',
-        '--cov-report=',
-        '--cov-fail-under=97',
-        os.path.join('tests', 'unit'),
-        *session.posargs
-    )
+    return unit(session, 'default')
```
tseaver: I don't understand why we are dropping the `system` / `lint` `cover` bits, nor why this library would be different than any of our other ones.

crwilcox: this is the template that comes from our generator. I can revert this part since it seems to be dropping things.

tseaver:
Are we not generating / creating any tests?

crwilcox:
@tseaver, tests are generated but there aren't any changes.

tseaver:
@crwilcox That doesn't seem to bode well for our coverage story, given that not all the changes are docs-only.

PR #5669, opened 2018-07-22
Firestore API: Fix return type name (unbalanced brackets) in pydoc.

multiple1902:
Currently the return type is specified as `Sequence[~.firestore_v1beta1.collection.CollectionReference[`. Changing to `Sequence[~.firestore_v1beta1.collection.CollectionReference]`. (Note the difference in the last symbol.)


multiple1902:
(Just signed CLA from the E-mail address that made the commit.)

tseaver:
@multiple1902 Thanks for the patch!

PR #5675, opened 2018-07-23
Fix broken links

danoscarmike:
AutoML client library documentation
Core Auth


danoscarmike:
Thanks @theacodes -- we need to cut a patch release for this also but can maybe wait until @crwilcox gets here later today.

PR #5678, opened 2018-07-23
BigQuery add default location to client

### LINKED ISSUE #5148: BigQuery: Add default location to Client
It is potentially quite cumbersome to pass in a `location` parameter to nearly every API request method. It would be nice to set a default `location` when you create a client to be used when creating datasets or jobs.

tseaver:
Plumb it through as the default-default for all `Client` methods taking an optional `location`.

Closes #5148.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -942,10 +942,19 @@ class Client(ClientWithProject):
             job_config = job.LoadJobConfig()
         job_config.source_format = job.SourceFormat.PARQUET
 
+        if location is None:
+            location = self.location
+
         return self.load_table_from_file(
-            buffer, destination, num_retries=num_retries, rewind=True,
-            job_id=job_id, job_id_prefix=job_id_prefix, location=location,
-            project=project, job_config=job_config)
+            buffer, destination,
+            num_retries=num_retries,
+            rewind=True,
+            job_id=job_id,
+            job_id_prefix=job_id_prefix,
+            location=location,
+            project=project,
+            job_config=job_config,
+        )
 
     def _do_resumable_upload(self, stream, metadata, num_retries):
         """Perform a resumable upload.
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -107,6 +107,8 @@ class Client(ClientWithProject):
             current object.
             This parameter should be considered private, and could change in
             the future.
+        location str:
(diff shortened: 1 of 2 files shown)


tseaver:
@tswast Two questions:

- Should `DatasetReference` grow location support?  It looks like the `location` field is present in [`datasets.list` responses](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/list#datasets.location), for instance.  It seems the natural point to inject `location`, e.g.:

```python
ds_ref = client.dataset(dataset_id, location='EU')
dataset = client.create_dataset(ds_ref)
```

where the client would use its own location as the default-default.

- While poking at that question, I looked `Table.location`:  ISTM that `Table` should *not* allow assignment to `location`:  the field is marked as read-only in the [API docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#location).

tswast:
> Should DatasetReference grow location support?

It does feel like the natural place for location, although for legacy reasons the actual [REST resource](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets) doesn't put location in the `DatasetReference` type. I hesitate to have extra info floating around in the Python DatasetReference class that doesn't correspond to the actual API resource.

I'll ask the BigQuery product eng folks if they plan to add a location parameter to other operations such as [Dataset GET](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/get) if so, let's move ahead with adding it to DatasetReference.

> ISTM that Table should not allow assignment to location: the field is marked as read-only in the API docs.

Correct. Table inherits the location from the dataset it belongs to and cannot be overridden.

tseaver:
> I'll ask the BigQuery product eng folks if they plan to add a location parameter to other operations such as Dataset GET if so, let's move ahead with adding it to DatasetReference.

Hmm, `location` is [already documented](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#location) as a member of the `Dataset` resource, returned by `datasets.get`.

> Such a change is unlikely to happen because `project.dataset` is a globally-unique identifier. Let's just populate the default location at `create_dataset()` time, as that is the only operation that requires a location for Datasets.

Wilco.

> Table inherits the location from the dataset it belongs to and cannot be overridden.

Tracked in #5686.

tswast:
> Hmm, location is already documented as a member of the Dataset resource, returned by datasets.get

That's true. What I mean is that dataset.get doesn't have a location URL parameter, as was recently added to jobs.get.

tseaver:
@dpebot Please merge on green.

PR #5684, opened 2018-07-24
Bigtable: refactor update_app_profile() to remove update_mask argument

AVaksman:
refactored update_app_profile() to remove update_mask argument
added asserts to unit tests
added system tests

### DIFF
--- a/bigtable/google/cloud/bigtable/enums.py
+++ b/bigtable/google/cloud/bigtable/enums.py
@@ -75,5 +75,5 @@ class RoutingPolicyType(object):
       See
       https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.AppProfile.SingleClusterRouting
     """
-    ANY = 1
-    SINGLE = 2
+    ANY = enums.RoutingPolicyType.ANY
+    SINGLE = enums.RoutingPolicyType.SINGLE
--- a/bigtable/google/cloud/bigtable/enums.py
+++ b/bigtable/google/cloud/bigtable/enums.py
@@ -53,3 +53,27 @@ class InstanceType(object):
     UNSPECIFIED = enums.Instance.Type.TYPE_UNSPECIFIED
     PRODUCTION = enums.Instance.Type.PRODUCTION
     DEVELOPMENT = enums.Instance.Type.DEVELOPMENT
+
+
+class RoutingPolicyType(object):
+    """
+    The type of the routing policy for app_profile.
+
+    Attributes:
+      ANY (int): Read/write requests may be routed to any cluster in the
+      instance, and will fail over to another cluster in the event of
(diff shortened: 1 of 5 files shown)


sduskis:
@AVaksman, there's a coverage issue:

> google/cloud/bigtable/instance.py          110      0     32      1    99%   480->483

AVaksman:
@sduskis thanks.
Fixed unit tests coverage

bigtable/google/cloud/bigtable/enums.py
```diff
@@ -53,3 +53,27 @@ class InstanceType(object):
...
+
+      SINGLE (int): Unconditionally routes all read/write requests to a
+      specific cluster.
+      This option preserves read-your-writes consistency, but does not improve
+      availability.
+      See
+      https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.AppProfile.SingleClusterRouting
+    """
+    ANY = 1
+    SINGLE = 2
```
tseaver: It seems odd to have bare `int` literals here, when the rest of the module points to the protobuf-generated code.  Should this enum be present in `google.cloud.bigtable_admin_v2.enums`, and aliased here?

bigtable/tests/system.py
```diff
@@ -210,33 +209,118 @@ def test_update(self):
...
+        self.assertEqual(app_profile, alt_app_profile)
+
+        # Modify existing app_profile to singly routing policy and confirm
+        new_description = 'To single routing policy'
+        allow_transactional_writes = False
+        operation = instance.update_app_profile(
+            app_profile_id=app_profile_id,
+            routing_policy_type=ROUTING_POLICY_TYPE_SINGLE,
+            description=new_description,
+            ignore_warnings=ignore_warnings,
```
tseaver: For completeness, should we be updating `ignore_warnings` to False here?

bigtable/tests/system.py
```diff
@@ -210,33 +209,118 @@ def test_update(self):
...
+            alt_app_profile.single_cluster_routing.allow_transactional_writes,
+            allow_transactional_writes)
+
+        # Modify existing app_proflie to multi cluster routing
+        new_description = 'To multi cluster routing'
+        operation = instance.update_app_profile(
+            app_profile_id=app_profile_id,
+            routing_policy_type=ROUTING_POLICY_TYPE_ANY,
+            description=new_description,
+            ignore_warnings=ignore_warnings)
```
tseaver: Likewise here?

bigtable/tests/system.py
sduskis: In general, it's a good idea to `ignore_warnings=True` in code.  `ignore_warnings=False` only seems to make sense for UI purposes.

tseaver: OK.

PR #5687, opened 2018-07-25
Make 'Table.location' read-only.

### LINKED ISSUE #5686: BigQuery: 'Table.location' should be read-only
From https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5678#issuecomment-407830285

Currently, the `Table.location` property is [settable](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/814e75bdff73c8c298eb90f828cc4f8a36f5cc3e/bigquery/google/cloud/bigquery/table.py#L603), but the [API reference](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#location) marks it as `[Output-only]`.

tseaver:
Documented as `[Output-only]` in the API reference.

Closes #5686.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -591,19 +591,11 @@ class Table(object):
 
     @property
     def location(self):
-        """Union[str, None]: Location in which the table is hosted (defaults
-        to :data:`None`).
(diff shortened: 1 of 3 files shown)


tseaver:
@dpebot Please merge on green.

PR #5690, opened 2018-07-26
Correct description of CopyJob sources param

alec-brooks:


alec-brooks:
I signed it!

tseaver:
Thanks for the patch!

PR #5713, opened 2018-07-31
Updates external_config docstrings to google style and fixes typos

alixhami:
image

Went in to fix the above rendering error and ended up fixing multiple typos, rendering errors, and updating docstrings to Google style.

@tswast - it's worth noting that we can now link to types in property docstrings


bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -243,8 +247,8 @@ def type_(self, value):
 
     @property
     def columns(self):
-        """List[google.cloud.bigquery.external_config.BigtableColumn]: Lists of
-        columns that should be exposed as individual fields
+        """List[:class:`BigtableColumn`]: Lists of columns that should be
```
tswast: We probably want to expand this out to the fully-qualified name. I've had docs failures due to an ambiguous class name when other cloud client libraries have the same class name.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -260,33 +264,34 @@ def columns(self, value):
...
+            Dict[str, Any]:
+                A dictionary in the format used by the BigQuery API.
         """
         return copy.deepcopy(self._properties)
 
     @classmethod
     def from_api_repr(cls, resource):
-        """Factory: construct a BigtableColumnFamily given its
-           API representation
+        """Factory: construct a :class:`BigtableColumnFamily` instance given
```
tswast: Ditto. Let's use the fully qualified name.

bigquery/google/cloud/bigquery/external_config.py
```diff
@@ -544,7 +553,7 @@ class ExternalConfig(object):
     """Description of an external data source.
 
     Args:
-        source_format :class:`ExternalSourceFormat`:
+        source_format (:class:`ExternalSourceFormat`):
```
tswast: Do we actually need the `:class:` part with Napoleon-style? I thought we didn't.

Also, let's do the fully-qualified name.

bigquery/google/cloud/bigquery/external_config.py
tseaver: FWIW, I find that jamming a super-long return type into what is supposed to be a one-line semantic summary is against the spirit of [PEP 257](https://www.python.org/dev/peps/pep-0257).  I don't know why it isn't part of the docstring body, instead.

bigquery/google/cloud/bigquery/external_config.py
alixhami: The `:class:` part isn't needed to render the link, but if I switch to the full name I'd still like it to show up as just the class name, and the tilde syntax doesn't work without `:class:`.

bigquery/google/cloud/bigquery/external_config.py
tswast: I'm fine with the short name so long as the docs still build. I'm probably being overly cautious since I was bit by it before.

alixhami: I made an update to make these more specific. `` :class:`~.external_config.BigtableColumn` `` renders the same as `` :class:`BigtableColumn` ``, but is less ambiguous. Not putting the full path also helps avoid the unintuitive line breaks required to pass the lint.

tseaver:
@alixhami Please rebase.

PR #5714, opened 2018-07-31
BigQuery: do the time conversions for undeletion

shollyman:


docs/bigquery/snippets.py
```diff
@@ -2161,8 +2163,10 @@ def test_undelete_table(client, to_delete):
     # time is at least within the creation lifespan of the table.  Trying
     # to snapshot a table prior to its creation time is an error.
     table = client.get_table(table_ref)
```
alixhami: In "real life" how would the user get the creation time if they accidentally deleted the table?

docs/bigquery/snippets.py
```diff
@@ -2161,8 +2163,10 @@ def test_undelete_table(client, to_delete):
...
-    if table.created_time > snapshot_time:
-      snapshot_time = table.created_time
+    created_ms = (table.created - datetime.utcfromtimestamp(0).replace(tzinfo=UTC)).total_seconds() * 1000
```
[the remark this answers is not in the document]
alixhami: Actually, hold on this. I don't think it will work with python 2.

docs/bigquery/snippets.py
shollyman: Real world use typically leverages awareness of the deletion event time.  The bounds checking here is really only necessary because the lifecycle of this table is measured in milliseconds, which is short enough to require the extra care.

alixhami:
Rethinking this, I think we should take this part out of the sample. It's not a common use case to create and then immediately delete and undelete a table, so I think this is adding some possible confusion for users. Maybe have the snapshot time be a commented variable at the top, and note to the developer to update to a time they know the table existed. Then you can have the snapshot time set to the creation time for purposes of the test.

docs/bigquery/snippets.py
```diff
@@ -2155,24 +2153,28 @@ def test_undelete_table(client, to_delete):
 
     table_ref = client.dataset(dataset_id).table(table_id)
 
-    # Record the current time in milliseconds. We'll use this as the snapshot
-    # time for recovering the table.
-    snapshot_time = int(time.time() * 1000)
+    # TODO(Developer):  Choose an appropriate snapshot point as epoch
```
alixhami: Make the whole comment part of the TODO (no newline). The separation in the comment makes the TODO less clear because it looks like a different section.

docs/bigquery/snippets.py
```diff
@@ -2153,22 +2153,28 @@ def test_undelete_table(client, to_delete):
```
alixhami: It won't let me comment on the lines above since they're not part of the diff, but update the first part of the sample to the new TODO format:
```
    # TODO(developer): Uncomment the lines below and replace with your values.
    # from google.cloud import bigquery
    # import time
    # client = bigquery.Client()
    # dataset_id = 'my_dataset'  # replace with your dataset ID
    # table_id = 'my_table'      # replace with your table ID
```

docs/bigquery/snippets.py
```diff
@@ -2153,22 +2153,28 @@ def test_undelete_table(client, to_delete):
...
 
-    # Because local system time may have some drift, we ensure the snapshot
-    # time is at least within the creation lifespan of the table.  Trying
-    # to snapshot a table prior to its creation time is an error.
+    # Due to very short lifecycle of the table, ensure we're not picking a time
+    # prior to the table creation due to time drift between backend and client.
     table = client.get_table(table_ref)
-    if table.created_time > snapshot_time:
-      snapshot_time = table.created_time
+    created_epoch = table._properties['creationTime']
```
tseaver: `table.created` converts the `creationTime` property to a `datetime.datetime` (in UTC).

Issue #5716, opened 2018-07-31
Bigtable: new cluster systests flake on CI


tseaver:
See: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/7511

```python
___________________ TestInstanceAdminAPI.test_create_cluster ___________________

self = <tests.system.TestInstanceAdminAPI testMethod=test_create_cluster>

    def test_create_cluster(self):
        from google.cloud.bigtable.enums import StorageType
        from google.cloud.bigtable.enums import Cluster
    
        ALT_CLUSTER_ID = INSTANCE_ID+'-cluster-2'
        ALT_LOCATION_ID = 'us-central1-f'
        ALT_SERVE_NODES = 4
    
        cluster_2 = Config.INSTANCE.cluster(ALT_CLUSTER_ID,
                                            location_id=ALT_LOCATION_ID,
                                            serve_nodes=ALT_SERVE_NODES,
                                            default_storage_type=(
                                               StorageType.SSD))
>       operation = cluster_2.create()

tests/system.py:478: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
google/cloud/bigtable/cluster.py:224: in create
    self._instance.name, self.cluster_id, cluster_pb)
google/cloud/bigtable_admin_v2/gapic/bigtable_instance_admin_client.py:798: in create_cluster
    request, retry=retry, timeout=timeout, metadata=metadata)
../.nox/sys-2-7/lib/python2.7/site-packages/google/api_core/gapic_v1/method.py:139: in __call__
    return wrapped_func(*args, **kwargs)
../.nox/sys-2-7/lib/python2.7/site-packages/google/api_core/retry.py:260: in retry_wrapped_func
    on_error=on_error,
../.nox/sys-2-7/lib/python2.7/site-packages/google/api_core/retry.py:177: in retry_target
    return target()
../.nox/sys-2-7/lib/python2.7/site-packages/google/api_core/timeout.py:206: in func_with_timeout
    return func(*args, **kwargs)
../.nox/sys-2-7/lib/python2.7/site-packages/google/api_core/grpc_helpers.py:61: in error_remapped_callable
    six.raise_from(exceptions.from_grpc_error(exc), exc)
[...]

PR #5719, opened 2018-07-31
Bigtable: shortening cluster ID in system tests

AVaksman:
Fixes issue #5716 
Shorten cluster ID addition to just `'c1'` and `'c2'` in both 
`test_create_instance_w_two_clusters` and `test_create_cluster system` tests


AVaksman:
I always run system tests before committing any changes. Looks like that `unique_resource_id` comes out much longer with circleci than when I run it on my local machine.

PR #5721, opened 2018-08-01
Fix typo in Client links

wyardley:
References are to `google.cloud.storage.client.Client` instead of `google.cloud.spanner.client.Client`, which I believe is a typo.

In addition, the client self-links don't seem to work via https://googlecloudplatform.github.io, and seems like there are some differences in the side menu between
http://gcloud-python.readthedocs.io/en/latest/spanner/client-usage.html
and 
https://googlecloudplatform.github.io/google-cloud-python/latest/spanner/usage.html
?

Also, I don't see how to get to the main generated docs of the parameters for each method.


tseaver:
Thank you for the patch!

PR #5728, opened 2018-08-02
BIgtable: public API 'instance.labels' type to dictionary

AVaksman:
Converting `instance.labels` public API to dictionary from protobuf `ScalarMapContainer`
Reroute to use `_update_from_pb` to set properties of instance and cluster in `list_instances` and `list_clusters`  methods.

### DIFF
--- a/bigtable/tests/system.py
+++ b/bigtable/tests/system.py
@@ -197,8 +197,7 @@ class TestInstanceAdminAPI(unittest.TestCase):
         self.assertEqual(instance, instance_alt)
         self.assertEqual(instance.display_name, instance_alt.display_name)
         self.assertEqual(instance.type_, instance_alt.type_)
-        self.assertIsInstance(instance_alt.labels, dict)
-        self.assertEqual(instance.labels, instance_alt.labels)
+        self.assertEqual(instance_alt.labels, LABELS)
 
     def test_cluster_exists(self):
         NONEXISTING_CLUSTER_ID = 'cluster-id'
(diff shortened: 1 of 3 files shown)


bigtable/tests/system.py
```diff
@@ -197,6 +197,7 @@ def test_create_instance(self):
         self.assertEqual(instance, instance_alt)
         self.assertEqual(instance.display_name, instance_alt.display_name)
         self.assertEqual(instance.type_, instance_alt.type_)
+        self.assertIsInstance(instance_alt.labels, dict)
         self.assertEqual(instance.labels, instance_alt.labels)
```
tseaver: It would be better to assert equality with `LABELS` (we want to know that they were assigned correctly).  That would cover the `isinstance` bit, too.

docs/bigquery/snippets.py
```diff
@@ -2162,7 +2162,7 @@ def test_undelete_table(client, to_delete):
     # to snapshot a table prior to its creation time is an error.
     table = client.get_table(table_ref)
     if table.created_time > snapshot_time:
-      snapshot_time = table.created_time
+        snapshot_time = table.created_time
```
tseaver: Please revert:  this is a stray change, unrelated to the PR.  I've just merged a fix to `master` in PR #5733, which you can pick up via a rebase.

AVaksman: Thank you, I just did it for the sake of passing the checks. removed the commit and rebased.

bigtable/tests/system.py
tseaver: The test against an explicit dict literal accomplishes *both* goals:  the dict values are correct, *and* `instance.lables` returns a dict, not a proto message.

AVaksman: Thank you for the advice. It is done

PR #5729, opened 2018-08-02
Bigtable: adding labels {'python-system': ISO-timestamp} to instance object

AVaksman:
Setting `instance.labels` in the form of `{'python-system': timestamp}`
Adding logic to clean undeleted `instance`s from previous runs where `timestamp` is older then 2 hours.

### DIFF
--- a/bigtable/tests/system.py
+++ b/bigtable/tests/system.py
@@ -39,7 +39,6 @@ from test_utils.system import unique_resource_id
 
 LOCATION_ID = 'us-central1-c'
 INSTANCE_ID = 'g-c-p' + unique_resource_id('-')
-LABELS = {u'foo': u'bar'}
 TABLE_ID = 'google-cloud-python-test-table'
 CLUSTER_ID = INSTANCE_ID+'-cluster'
 SERVE_NODES = 3
@@ -57,6 +56,10 @@ ROW_KEY_ALT = b'row-key-alt'
 ROUTING_POLICY_TYPE_ANY = 1
 ROUTING_POLICY_TYPE_SINGLE = 2
 EXISTING_INSTANCES = []
+LABEL_KEY = u'python-system'
+label_stamp = datetime.datetime.utcnow().replace(tzinfo=UTC)
+label_stamp_micros = _microseconds_from_datetime(label_stamp)
+LABELS = {LABEL_KEY: str(label_stamp_micros)}
 
 
 class Config(object):
@@ -101,6 +104,18 @@ def setUpModule():
             raise ValueError('List instances failed in module set up.')
 
         EXISTING_INSTANCES[:] = instances
+
+        # Clean up undeleted instances from prior systest run if any
+        def age_in_hours(micros):
+            return (datetime.datetime.utcnow().replace(tzinfo=UTC) - (
+                _datetime_from_microseconds(micros))).total_seconds() // 3600
+        INSTANCE_TO_CLEAN = [
+            instance for instance in EXISTING_INSTANCES if (
+                LABEL_KEY in instance.labels.keys() and
+                age_in_hours(int(instance.labels[LABEL_KEY]))) >= 2]
+        if INSTANCE_TO_CLEAN:
+            for instance in INSTANCE_TO_CLEAN:
+                instance.delete()
 
         # After listing, create the test instance.
         created_op = Config.INSTANCE.create(clusters=[Config.CLUSTER])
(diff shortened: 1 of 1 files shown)


AVaksman:
The only test that creates instance without `labels` is `test_create_instance_defaults` test
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/fa88745a7291a0066d3c432565d4e594be90f735/bigtable/tests/system.py#L149-L162

and corresponding asserts

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/fa88745a7291a0066d3c432565d4e594be90f735/bigtable/tests/system.py#L173-L174

Should `labels` be added to here as well?

bigtable/tests/system.py
```diff
@@ -57,6 +56,10 @@
 ROUTING_POLICY_TYPE_ANY = 1
 ROUTING_POLICY_TYPE_SINGLE = 2
 EXISTING_INSTANCES = []
+LABEL_KEY = u'python-system'
+label_stamp = datetime.datetime.utcnow().replace(tzinfo=UTC)
+label_stamp_micros = _microseconds_from_datetime(label_stamp)
+LABELS = {LABEL_KEY: str(label_stamp_micros)}
```
tseaver: I would rather use an ISO timestamp format, for human readability.

bigtable/tests/system.py
```diff
@@ -102,6 +105,18 @@ def setUpModule():
...
+            return (datetime.datetime.utcnow().replace(tzinfo=UTC) - (
+                _datetime_from_microseconds(micros))).total_seconds() // 3600
+        INSTANCE_TO_CLEAN = [
+            instance for instance in EXISTING_INSTANCES if (
+                LABEL_KEY in instance.labels.keys() and
+                age_in_hours(int(instance.labels[LABEL_KEY]))) >= 2]
+        if INSTANCE_TO_CLEAN:
+            for instance in INSTANCE_TO_CLEAN:
+                instance.delete()
+
```
tseaver: I like adding the labels, but would prefer *not* cleaning instances during the system test run:  we lose the ability to figure out why the leak happened.

bigtable/tests/system.py
```diff
@@ -259,7 +275,10 @@ def test_create_instance_w_two_clusters(self):
     def test_update_display_name_and_labels(self):
         OLD_DISPLAY_NAME = Config.INSTANCE.display_name
         NEW_DISPLAY_NAME = 'Foo Bar Baz'
-        NEW_LABELS = {'foo_bar': 'foo_bar'}
+        new_label_stamp = datetime.datetime.utcnow().replace(tzinfo=UTC)
+        new_label_stamp_micros = _microseconds_from_datetime(new_label_stamp)
```
tseaver: Again, use the human-readable ISO format here.

docs/bigquery/snippets.py
```diff
@@ -2162,7 +2162,7 @@ def test_undelete_table(client, to_delete):
     # to snapshot a table prior to its creation time is an error.
     table = client.get_table(table_ref)
     if table.created_time > snapshot_time:
-      snapshot_time = table.created_time
+        snapshot_time = table.created_time
```
tseaver: Please revert / rebase to pick up the fix from #5733.

bigtable/tests/system.py
```diff
@@ -57,6 +56,11 @@
 ROUTING_POLICY_TYPE_ANY = 1
 ROUTING_POLICY_TYPE_SINGLE = 2
 EXISTING_INSTANCES = []
+LABEL_KEY = u'python-system'
+label_stamp = datetime.datetime.utcnow() \
+                               .replace(microsecond=0, tzinfo=UTC,) \
+                               .strftime("%Y-%m-%dt%H-%M-%S")
```
tseaver: The [`datetime.datetime.isoformat()`](https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat) method is probably simpler to grasp (and makes the zone offset explicit, too).

AVaksman: I couldn't use `datetime.datetime.isoformat()` because it produces format `YYYY-MM-HH'set'HH:MM:SS` and colon `:` is not allowed in the `labels.value`, so I resorted to use `datetime.datetime.strftime()` to have time format be delimited with hyphen `-`.

tseaver: Weird -- I didn't realize there were restrictions on label values (who knows why?).

tswast: Yeah, label values are weirdly restricted.

> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes.

https://cloud.google.com/resource-manager/docs/creating-managing-labels

The justification I heard was that the design is to keep labels human-readable and concise for viewing in a list. It's a cloud-wide restriction, so hard to loosen.

PR #5749, opened 2018-08-05
Bigtable optimize 'Table.exists' performance

sumit-ql:
1. Updated table.exists(), now using NAME_ONLY view in get_table() call.
2. Added system test around table.exists()

### DIFF
--- a/bigtable/google/cloud/bigtable/enums.py
+++ b/bigtable/google/cloud/bigtable/enums.py
@@ -121,3 +121,31 @@ class RoutingPolicyType(object):
     """
     ANY = enums.RoutingPolicyType.ANY
     SINGLE = enums.RoutingPolicyType.SINGLE
+
+
+class Table(object):
+    class View(object):
+        """
+        Defines a view over a table's fields.
+
+        Attributes:
+          VIEW_UNSPECIFIED (int): Uses the default view for each method
+          as documented in its request.
+          NAME_ONLY (int): Only populates ``name``.
+          SCHEMA_VIEW (int): Only populates ``name`` and fields related
+          to the table's schema.
+          REPLICATION_VIEW (int): This is a private alpha release of
+          Cloud Bigtable replication. This feature is not currently available
+          to most Cloud Bigtable customers. This feature might be changed in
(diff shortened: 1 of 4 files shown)


sduskis:
I learned something new with @AVaksman.  We do need to go back to using `GetTable`.  You'll need to change the view to `NAME_ONLY` in order to only read the name.

AVaksman:
@sduskis good point.
Then @sumit-ql please revert implementation back to `get_table()` and do the following changes:
1. Please create a `class Table` with sub`class View` in `bigtable.enums.py`. Import and add `google.cloud.bigtable_admin_v2.gapic.enums.Table.View` values.
2. Add `from google.cloud.bigtable.enums import Table` import in `table.py` and add `view=Table.View.NAME_ONLY` parameter to [`table_client.get_table(name=self.name)`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/table.py#L226) function call
    ```python
        table_client.get_table(name=self.name, view=Table.View.NAME_ONLY)
    ```
3. To make sure that other than `NotFound` error propagates to the user please change `def exists` [implantation](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/table.py#L225-L230) to:
    ```python
    try:
        table_client.get_table(name=self.name)
        return True
    except NotFound:
        return False
    ```
4. In unit tests please change to use the `GetTable` method [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigtable/tests/unit/test_table.py#L335)
    ```python
            bigtable_table_stub.GetTable.side_effect = [
    ```
5. Please add another scenario to unit test where a different than `NotFound` error is raised and propagated to user. e.i.
    ```python
    bigtable_table_stub.GetTable.side_effect = [
                response_pb,
                exceptions.NotFound('testing'),
                exceptions.BadRequest('testing')
            ]
    ```
    and add another assert
    ```python
        with self.assertRaises(exceptions.BadRequest):
            table2.exists()
    ```

Issue #5750, opened 2018-08-06
BigQuery: '_time_from_json' raises for values w/ sub-second precision


cpcloud:
- OS: Arch Linux
- relevant conda versions:
```
# packages in environment at /home/phillip/miniconda3/envs/ibis36:
#
# Name                    Version                   Build  Channel
google-cloud-bigquery     1.4.0                      py_0    conda-forge
google-cloud-core         0.28.1                     py_0    conda-forge
```

- Steps to reproduce:

```python
import google.cloud.bigquery as bq
client = bq.Client('ibis-gbq')
sql = """
SELECT TIME(`creation_date`) AS `created_time`
FROM `bigquery-public-data.stackoverflow.posts_questions`
"""
query = client.query(sql)
df = query.result().to_dataframe()
```

- Traceback:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-27-caeea53c88e2> in <module>()
----> 1 df = query.result().to_dataframe()

~/miniconda3/envs/ibis36/lib/python3.6/site-packages/google/cloud/bigquery/table.py in to_dataframe(self)
   1145
   1146         column_headers = [field.name for field in self.schema]
-> 1147         rows = [row.values() for row in iter(self)]
   1148
   1149         return pandas.DataFrame(rows, columns=column_headers)

~/miniconda3/envs/ibis36/lib/python3.6/site-packages/google/cloud/bigquery/table.py in <listcomp>(.0)
   1145
   1146         column_headers = [field.name for field in self.schema]
-> 1147         rows = [row.values() for row in iter(self)]
   1148
   1149         return pandas.DataFrame(rows, columns=column_headers)

~/miniconda3/envs/ibis36/lib/python3.6/site-packages/google/api_core/page_iterator.py in _items_iter(self)
    198         """Iterator for each item returned."""
    199         for page in self._page_iter(increment=False):
--> 200             for item in page:
    201                 self.num_results += 1
    202                 yield item

~/miniconda3/envs/ibis36/lib/python3.6/site-packages/google/api_core/page_iterator.py in next(self)
    123         """Get the next value in the page."""
    124         item = six.next(self._item_iter)
--> 125         result = self._item_to_value(self._parent, item)
    126         # Since we've successfully got the next value from the
    127         # iterator, we update the number of remaining.

[...]

PR #5753, opened 2018-08-06
Fix '_time_from_json' for values with micros.

tseaver:
See #5750.

The issue remains in 'google.cloud._helpers._time_from_iso8601_time_naive',
but I don't want to have BigQuery wait on a release of core.

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -23,10 +23,11 @@ from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
 from google.cloud._helpers import _microseconds_from_datetime
 from google.cloud._helpers import _RFC3339_NO_FRACTION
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -142,8 +143,13 @@ def _date_from_json(value, field):
 def _time_from_json(value, field):
     """Coerce 'value' to a datetime date, if set or not nullable"""
     if _not_null(value, field):
-        # value will be a string, in HH:MM:SS form.
-        return _time_from_iso8601_time_naive(value)
+        if len(value) == 8:  # HH:MM:SS
```
mcdonc: Should these be regexes instead of length checks?

tseaver: I think not:  `datetime.datetime.strptime` will raise a `ValueError` if the value is bogus:  I'm just using the length to select which format string applies.

PR #5754, opened 2018-08-06
Regenerate kms (do not merge)

andreamlin:


kms/setup.py
```diff
@@ -20,7 +20,7 @@
 name = 'google-cloud-kms'
 description = 'Google Cloud Key Management Service (KMS) API API client library'
 version = '0.1.0'
-release_status = 'Development Status :: 3 - Alpha'
+release_status = '3 - Alpha'
```
tseaver: This part is incorrect -- the `Development Status ::` prefix is required to make this a [valid Trove classifier](https://pypi.org/classifiers/)

PR #5755, opened 2018-08-06
Remove synth fix for replacing iam_policy_pb2_grpc

andreamlin:
https://github.com/googleapis/gapic-generator/issues/2160 is fixed.

See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5754 for the regenerated kms with gapic-generator/master.

Thanks for filing!


andreamlin:
PTAL

PR #5763, opened 2018-08-07
Storage: document thread-safety of client.

### LINKED ISSUE #5685: Document thread-safety of library
After reading through the docs, it is unclear to me whether I should expect this code to work:

```
import thread

client = storage.Client()
for blob in bucket.list_blobs():
  thread.start_new_thread( blob.download_to_filename, '/tmp/' + blob.name)
```

tseaver:
/cc @BrandonY

Also note general best practice for multiprocessing use.

Closes #5685.

### DIFF
--- a/docs/storage/client.rst
+++ b/docs/storage/client.rst
@@ -26,7 +26,7 @@ Usage
 .. note::
 
    Becuase the :class:`~google.cloud.storage.client.Client` uses the
-   third-party :mod:`requests` library by default, it should be safe to
+   third-party :mod:`requests` library by default, it is safe to
    share instances across threads.  In multiprocessing scenarious, best
    practice is to create client instances *after*
    :class:`multiprocessing.Pool` or :class:`multiprocessing.Process` invokes
--- a/docs/storage/client.rst
+++ b/docs/storage/client.rst
@@ -23,6 +23,15 @@ Install the ``google-cloud-storage`` library using ``pip``:
 Usage
 -----
 
+.. note::
+
+   Becuase the :class:`~google.cloud.storage.client.Client` uses the
+   third-party :mod:`requests` library by default, it should be safe to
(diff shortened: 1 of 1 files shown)


docs/storage/client.rst
```diff
@@ -23,6 +23,15 @@ Install the ``google-cloud-storage`` library using ``pip``:
...
 -----
 
+.. note::
+
+   Becuase the :class:`~google.cloud.storage.client.Client` uses the
+   third-party :mod:`requests` library by default, it should be safe to
+   share instances across threads.  In multiprocessing scenarious, best
+   practice is to create client instances *after*
+   :class:`multiprocessing.Pool` or :class:`multiprocessing.Process` invokes
+   :func:`os.fork`.
```
frankyn: We may want to add a test for this case, "should be safe" would concern me.

tseaver: Our thread-safety issues in the past were due to using `httplib2`.  While individual `requests.session.Session` objects aren't thread-safe, we don't use them directly.  The session pool comes from `urllib3`, which is supposed to be so.  @theacodes Can you confirm?

theacodes: Yeah, urllib3 is threadsafe and so is requests.

frankyn: Can we update the wording to "is thread-safe" or is that too strong?

theacodes:
sure

kaliberto:
I ran into this same question when trying to optimize some Flask code running on GCF/GAE. Is the library written in such a way where we can make a general statement for all the APIs? Or is it really different for each API?

tseaver:
@kaliberto We have two basic groups of APIs:  those which use gRPC as their basic transport, and those whic use requests/HTTP.  The gRPC-based clients are threadsafe, but not fork-safe;  the requests-based clients are as threadsafe as requests (i.e., no known issues).

PR #5765, opened 2018-08-07
Storage: Add 'Client.get_service_account_email'.

### LINKED ISSUE #5635: Storage: add support get_service_account method
Feature request to add support for [projects.serviceAccount](https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount/get) 

The request gets the GCS managed service account that is used by GCS PubSub notifications and CMEK to allow GCS to access to PubSub and a KMS key respectively.

Examples of existing implementations:
* [PHP](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/1173)
* [Java](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java#L987)

tseaver:
Returns email address of service account of a given project, defaulting to the client's project.

Closes #5635.

### DIFF
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -151,6 +151,24 @@ class Client(ClientWithProject):
         """
         return self._batch_stack.top
 
+    def get_service_account_email(self, project=None):
+        """Get the email address of the project's GCS service account
+
+        :type project: str
+        :param project:
+            (Optional) Project ID to use for retreiving service account email.
+            Defaults to the client's project.
+
+        :rtype: str
+        :returns: service account email address
+        """
+        if project is None:
+            project = self.project
+        path = '/projects/%s/serviceAccount' % (project,)
+        api_response = self._base_connection.api_request(
+            method='GET', path=path)
+        return api_response['email_address']
+
     def bucket(self, bucket_name, user_project=None):
         """Factory constructor for bucket object.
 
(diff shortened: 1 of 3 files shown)


storage/tests/system.py
```diff
@@ -82,6 +82,15 @@ def tearDownModule():
     retry(Config.TEST_BUCKET.delete)(force=True)
 
 
+class TestClient(unittest.TestCase):
+
+    def test_get_service_account_email(self):
+        expected = '{}@gs-project-accounts.iam.gserviceaccount.com'.format(
+            Config.CLIENT.project)
```
frankyn: New projects will use `service-[PROJECT-NUMBER]@gs-project-accounts.iam.gserviceaccount.com` May want to keep this loose for the future.

storage/google/cloud/storage/client.py
```diff
@@ -151,6 +151,24 @@ def current_batch(self):
         """
         return self._batch_stack.top
 
+    def get_service_account_email(self, project=None):
+        """Get the email address of the project's GCS service account
+
+        :type project: str
+        :param project:
+            (Optional) Project ID to use for retreiving service account email.
+            Defaults to the client's project.
```
frankyn: Project ID to use when retrieving the GCS service account email address.

tseaver: 80805ec

PR #5767, opened 2018-08-07
Storage: elaborate docs for '{Bucket,Blob}.make_{public,private}'.

### LINKED ISSUE #5457: [Storage] make_public() and public_url() don't describe how they work/when they work.
Documentation for `make_public()` and `public_url()` don't describe how they work/when they work.

Specifically `make_public()` uses ACLs to make the object public and `public_url()` only works when the object has necessary permissions granted to all users in the form of `ACL` or a project/bucket has an IAM policy assigning `allUsers` the role `roles/storage.objectViewer`. Users should stay away from project-level IAM policies with allUsers roles/storage.objectViewer` that would not be good. 

(noting this down as I don't want to forget it.)

tseaver:
Also note how to enable anonymous access to 'Blob.public_url'.

Closes #5457.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -1303,7 +1303,9 @@ class Bucket(_PropertyMixin):
         :raises ValueError:
             If ``recursive`` is True, and the bucket contains more than 256
             blobs.  This is to prevent extremely long runtime of this
-            method.
+            method.  For such buckets, iterate over the blobs returned by
+            :meth:`Bucket.list_blobs` and call :meth:`Blob.make_public`
+            for each blob.
         """
         self.acl.all().grant_read()
         self.acl.save(client=client)
@@ -1322,10 +1324,11 @@ class Bucket(_PropertyMixin):
                 client=client))
             if len(blobs) > self._MAX_OBJECTS_FOR_ITERATION:
                 message = (
-                    'Refusing to make public recursively with more than '
-                    '%d objects. If you actually want to make every object '
-                    'in this bucket public, please do it on the objects '
-                    'yourself.'
+                    "Refusing to make public recursively with more than "
+                    "%d objects. If you actually want to make every object "
+                    "in this bucket public, iterate through the blobs "
+                    "returned by 'Bucket.list_blobs()' and call "
+                    "'make_public' on each one."
                 ) % (self._MAX_OBJECTS_FOR_ITERATION,)
                 raise ValueError(message)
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -1303,6 +1299,11 @@ def make_public(self, recursive=False, future=False, client=None):
                       ``NoneType``
         :param client: Optional. The client to use.  If not passed, falls back
                        to the ``client`` stored on the current bucket.
+
+        :raises ValueError:
+            If ``recursive`` is True, and the bucket contains more than 256
+            blobs.  This is to prevent extremely long runtime of this
+            method.
```
frankyn: Is there a way to go past 256 blobs?

tseaver: The user can iterate over the blobs directly, deleting them.  See the error message below at line 1324.

tseaver:
@frankyn Any remaining issues?

storage/google/cloud/storage/bucket.py
frankyn: Suggestion for the documentation:

```
If ``recursive`` is True, and the bucket contains more than 256
            blobs.  This is to prevent extremely long runtime of this
            method. Use :meth:`Bucket.list_blobs` and call :meth:`Blob.make_public`:
            for each blob.
```

Also error message suggestion:
```
message = (
                     'Refusing to make public recursively with more than '
                     '%d objects. If you want to make every object '
                     'in this bucket public, iterate through objects using'
                     'Bucket.list_blobs() and for each Blob call Blob.make_public()`
```

tseaver: 505587f

storage/google/cloud/storage/bucket.py
```diff
@@ -1303,6 +1299,13 @@ def make_public(self, recursive=False, future=False, client=None):
                       ``NoneType``
         :param client: Optional. The client to use.  If not passed, falls back
                        to the ``client`` stored on the current bucket.
+
+        :raises ValueError:
+            If ``recursive`` is True, and the bucket contains more than 256
+            blobs.  This is to prevent extremely long runtime of this
+            method.  For such buckets, iterate over the blobs returned by
+            :meth:`Bucket.list_blobs` and call :meth:`Blob.make_public`
```
frankyn: I took a guess at these :meth: links. Will it link correctly?

tseaver: c8579d5

PR #5778, opened 2018-08-09
Update top-level README to indicate support for App Engine Python 3.

theacodes:


theacodes:
Offline LGTM from GAE PM.

PR #5784, opened 2018-08-10
Make Publisher batch-related interfaces private

### LINKED ISSUE #5779: Make PubSub publisher's batch interfaces private
In preparation for GA, we're going to hide some interfaces that users don't need at the moment. We can always make these public in a future minor release.

Please make the following items private:

1. https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/pubsub/google/cloud/pubsub_v1/publisher/batch (rename the whole module to `_batch`).
2. https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py#L46 (make the argument private or remove it outright)
3. https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py#L105 Make the method private.

mcdonc:
Closes #5779


docs/pubsub/publisher/api/batch.rst
```diff
@@ -3,6 +3,6 @@
 Batch API
```
theacodes: Just delete this file altogether.

docs/pubsub/publisher/index.rst
```diff
@@ -59,7 +59,7 @@ Batching
 --------
 
 Whenever you publish a message, a
-:class:`~.pubsub_v1.publisher.batch.thread.Batch` is automatically created.
+:class:`~.pubsub_v1.publisher._batch.thread.Batch` is automatically created.
```
theacodes: we'll need to update this paragraph (and the other one below) to just say something like "Whenever you publish a message, the publisher will automatically batch the messages over a small time window to avoid making too many separate requests to the service and increase throughput."

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -68,6 +61,12 @@ def __init__(self, batch_settings=(), batch_class=thread.Batch, **kwargs):
                 target=os.environ.get('PUBSUB_EMULATOR_HOST'),
             )
 
+        if 'batch_class' in kwargs:
```
theacodes: Just drop this, we're before GA so we're allowed to remove this.

mcdonc: Removing, although the purpose was to give a reasonable exception to anyone happening to use this; without it, it will be passed as kwargs to another constructor later, which will presumably blow up.

mcdonc:
@theacodes your review comments are now addressed

theacodes:
Thank you, @mcdonc!

PR #5802, opened 2018-08-14
Bigtable: add 'Instance.exists()'

AVaksman:
Added Instance.exists() method
Unit tests
System tests

### DIFF
--- a/bigtable/tests/unit/test_instance.py
+++ b/bigtable/tests/unit/test_instance.py
@@ -394,6 +394,44 @@ class TestInstance(unittest.TestCase):
         # Check Instance optional config values before.
         self.assertEqual(instance.display_name, DISPLAY_NAME)
 
+    def test_exists(self):
+        from google.cloud.bigtable_admin_v2.gapic import (
+            bigtable_instance_admin_client)
+        from google.cloud.bigtable_admin_v2.proto import (
+            instance_pb2 as data_v2_pb2)
+        from google.api_core import exceptions
+
+        api = (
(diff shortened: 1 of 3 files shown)


sduskis:
@GoogleCloudPlatform/yoshi-python can you please review this?

bigtable/tests/unit/test_instance.py
```diff
@@ -394,6 +394,44 @@ def test_reload(self):
...
+        instance_stub.GetCluster.side_effect = [
+            response_pb,
+            exceptions.NotFound('testing'),
+            exceptions.BadRequest('testing')
+        ]
+
+        # Perform the method and check the result.
+        non_existing_instance_id = 'instance-id-2'
+        alt_instance_1 = self._make_one(self.INSTANCE_ID, client)
+        alt_instance_2 = self._make_one(non_existing_instance_id, client)
```
crwilcox: It seems odd to make two instances since the stub doesn't actually pay attention to the params of the call coming in.

AVaksman: @crwilcox  I did it for "better code comprehension" so to speak, to show why one `assertTrue` and the other `assertFalse`. I can change names to `instance` and `non_existing_instance` or make all calls on one instance.
Please let me know what you prefer.

crwilcox: It wasn't a hard preference. It just seemed odd to me. It isn't a big deal, I just wanted to point it out in case it was an accident. Thanks for your contribution!

AVaksman: Thank you

PR #5808, opened 2018-08-15
Move location specification to Bucket.create and deprecate location setter

### LINKED ISSUE #5194: Support for Bucket.location setter
I was trying to change location of the bucket in Storage, but everything gets updated apart from location. It looks like the data is being sent through API, but are not persisted?

Code example
```python
storage_client = StorageClient()
bucket = storage_client.create_bucket(name)
bucket.storage_class = 'MULTI_REGIONAL'
bucket.location = 'EU' # tried 'eu' as well
bucket.labels = {
     'project': 'etl-transformer',
     'group': 'etl',
}
bucket.patch()
```

```python
# update_properties printed in _PropertyMixin:patch
{
    'labels': {
        'project': 'etl-transformer',
        'group': 'etl'
    },
    'storageClass': 'MULTI_REGIONAL',
    'location': 'EU'
}

# I checked here even dump in requests package to be sure it's being passed to the http object

# api_reponse
{
    u 'labels': {
      u 'project': u 'etl-transformer',
      u 'group': u 'etl'
    },
    u 'location': u 'US',
    u 'storageClass': u 'MULTI_REGIONAL'
}
```

I know that documentation doesn't say that `location` attribute is a setter, but under the hood it's a `_scalar_property` which is appended to changed properties and according to json API, location is writable:

crwilcox:
Addresses #5194 by labeling setting location deprecated. Support setting location on create method of bucket.

### DIFF
--- a/storage/tests/unit/test_bucket.py
+++ b/storage/tests/unit/test_bucket.py
@@ -308,6 +308,25 @@ class Test_Bucket(unittest.TestCase):
         self.assertEqual(kw['query_params'], {'project': OTHER_PROJECT})
         self.assertEqual(kw['data'], DATA)
 
+    def test_create_w_explicit_location(self):
+        PROJECT = 'PROJECT'
+        BUCKET_NAME = 'bucket-name'
+        LOCATION = 'continent-northeast1'
+        DATA = {'location': LOCATION, 'name': BUCKET_NAME}
+        connection = _Connection(
+            DATA,
+            "{'location': 'continent-northeast1', 'name': 'bucket-name'}")
+        client = _Client(connection, project=PROJECT)
+        bucket = self._make_one(client, BUCKET_NAME)
+
+        bucket.create(location=LOCATION)
+
+        kw, = connection._requested
+        self.assertEqual(kw['method'], 'POST')
+        self.assertEqual(kw['path'], '/b')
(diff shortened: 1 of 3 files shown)


crwilcox:
Replaces https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5268

frankyn:
Documentation should probably be updated here as `Bucket.location=` would be soft-deprecated:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/storage/google/cloud/storage/client.py#L264

storage/tests/unit/test_bucket.py
```diff
@@ -308,6 +308,25 @@ def test_create_w_explicit_project(self):
         self.assertEqual(kw['query_params'], {'project': OTHER_PROJECT})
         self.assertEqual(kw['data'], DATA)
 
+    def test_create_w_explicit_location(self):
+        PROJECT = 'PROJECT'
+        BUCKET_NAME = 'bucket-name'
+        LOCATION = 'continent-northeast1'
```
frankyn: use a possible location here (I understand it's not making a real request):
"us-central1"

crwilcox: this is funny because I did and thought the exact opposite. I will make it real.

PR #5821, opened 2018-08-20
Datastore: Expose 'Client.base_url' property.

### LINKED ISSUE #5801: Datastore: Ability to override base URL
Seee https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/datastore/google/cloud/datastore/client.py.

It is currently only possible to override the base URL:

`_DATASTORE_BASE_URL = 'https://datastore.googleapis.com'`

by setting the `GCD_HOST` environment variable:

```
        try:
            host = os.environ[GCD_HOST]
            self._base_url = 'http://' + host
        except KeyError:
            self._base_url = _DATASTORE_BASE_URL
```

but this is meant to be used with the emulator and so does not use SSL.

It would be helpful to be able to override base_url to an https address.

tseaver:
Allows setting to alternate endpoints, e.g. for non-prod or use via something
like 'batch-datastore.googleapis.com'.

Closes #5801.

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -222,6 +222,16 @@ class Client(ClientWithProject):
         return _determine_default_project(project)
 
     @property
+    def base_url(self):
+        """Getter for API base URL."""
+        return self._base_url
+
+    @base_url.setter
+    def base_url(self, value):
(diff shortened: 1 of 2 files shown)


tseaver:
@theacodes Ping?

PR #5828, opened 2018-08-21
Add test for testing invoking a wrapped class method on the class itself

theacodes:


theacodes:
@tseaver FYI, merging because it's a test-only change.

PR #5832, opened 2018-08-21
Storage: Fix failing system tests

### LINKED ISSUE #5830: Storage: Python test fails with bunch of errors.
URL: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/CONTRIBUTING.rst#running-system-tests

Steps:
1. git clone https://github.com/GoogleCloudPlatform/google-cloud-python
2. Create JSON key on developer console 
3. Set GOOGLE_APPLICATION_CREDENTIALS: The path to a JSON key file
4. Change directory to google-cloud-python
5.  Run Storage system test
ex: ~/.local/bin/nox -f storage/nox.py -s "system(py='2.7')"

Attached is the error log from test output.

Error:

[Errors.docx](https://github.com/GoogleCloudPlatform/google-cloud-python/files/2308110/Errors.docx)

crwilcox:
Test should use service account of current user. KMS cannot be run without setup or CI account.
fixes #5830

### DIFF
--- a/storage/nox.py
+++ b/storage/nox.py
@@ -90,6 +90,7 @@ def system(session, py):
     session.install('mock', 'pytest', *LOCAL_DEPS)
     session.install('../test_utils/')
     session.install('../pubsub')
+    session.install('../kms')
     session.install('-e', '.')
 
     # Run py.test against the system tests.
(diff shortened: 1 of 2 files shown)


crwilcox:
Looking into adding KMS setup but for now this would allow running system tests locally

PR #5865, opened 2018-08-29
Nox: use inplace installs

tseaver:
The overall intent here is to use in-place installs for all local packages, to ease debugging using the "mainline" versions of the packages.


tseaver:
@tswast 
> Can we also do inplace installs for the documentation builds? 

I borrowed Guido's time machine:  see #5775.

PR #5868, opened 2018-08-29
Add docs for modify_ack_deadline to pubsub docs

crwilcox:


docs/pubsub/subscriber/api/message.rst
```diff
@@ -2,4 +2,4 @@ Messages
 ========
 
 .. autoclass:: google.cloud.pubsub_v1.subscriber.message.Message
-  :members: ack, attributes, data, nack, publish_time
+  :members: ack, modify_ack_deadline, attributes, data, nack, publish_time
```
tseaver: Why are we calling out specific members, rather than documenting all the public members?  We therefore don't document:

- `Message.ack_id`
- `Message.size`
- `Message.lease`
- `Message.drop`

crwilcox: @tseaver , So, this I am unsure of. Which is why I didn't want to hammer merge. I haven't seen us call out members to document often. I colud be convinced it makes sense to document all of them. Though I assumed there might be a reason? Adding modify_ack_deadline made sense to me.

PR #5870, opened 2018-08-30
Avoid pulling entire result set into memory when constructing dataframe.

### LINKED ISSUE #5859: Facing memory issue while loading the bigquery table data through python.
Hi, while trying to fetch the data of 1.8 M rows from Bigquery to local machine, we are getting Memory Error. 

Import pandas_gbq as pgbq
 df= pgbq.read_gbq(sql, dialect)

tseaver:
Closes #5859.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -1166,7 +1166,8 @@ class RowIterator(HTTPIterator):
             raise ValueError(_NO_PANDAS_ERROR)
 
         column_headers = [field.name for field in self.schema]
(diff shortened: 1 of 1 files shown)


max-sixty:
Just came across this. Ping us over at pandas-gbq if you have other performance thoughts / ideas

On this: I'm not sure this change will have that much impact, because pandas will coerce to in-memory python values regardless. Parsing each page and loading that into an array makes a big difference but I haven't implemented, in the hope that BQ has a step change in APIs from JSON-over-HTTP; full description: https://github.com/pydata/pandas-gbq/issues/133#issuecomment-383253436

PR #5871, opened 2018-08-30
Run 'unit_grpc_gcp' tests only with Python 2.7 / 3.6.

tseaver:


tseaver:
> why 3.6 instead of 3.7?

Those are the same versions for which we run system tests.  3.6 is the system python3 on current LTS Linux releases.

PR #5885, opened 2018-09-04
Add span_id to LogEntry

salrashid123:
Add `span_id` to `LogEntry`

Note `trace` should get specified if `span_id` is set.  The logging API does not currently return an error and while the final displayed log line does have a `span_id`, its not too useful.  Instead of client-side validation, CloudLogging would return exception later (i'll file internal FR for that)

ref:
 - https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5505

### DIFF
--- a/logging/google/cloud/logging/entries.py
+++ b/logging/google/cloud/logging/entries.py
@@ -79,10 +79,14 @@ class _BaseEntry(object):
 
     :type trace: str
     :param trace: (optional) traceid to apply to the entry.
+
+    :type span_id: str
+    :param span_id: (optional) span_id within the trace for the log entry.
+                    Specify the trace parameter if span_id is set.
     """
     def __init__(self, payload, logger, insert_id=None, timestamp=None,
                  labels=None, severity=None, http_request=None, resource=None,
-                 trace=None):
+                 trace=None, span_id=None):
         self.payload = payload
         self.logger = logger
         self.insert_id = insert_id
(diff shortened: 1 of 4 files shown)


salrashid123:
am i missing something in the cover tests?  i didn't really do anything there to account for that coverage failure

```
Name                                                            Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------------------
google/cloud/logging/logger.py                                    124      1     58      1    99%   593, 584->593
```

tseaver:
@salrashid123 I'm puzzled, too:  that branch / line should be covered by `test_commit_w_invalid_entry_type`.

salrashid123:
@tseaver   Is there a directive you can invoke to rerun the checks (i.e, `/oktotest`)?  

btw, i tried running `cover` locally but that looks for `python3.7`

```
$ nox -f logging/nox.py -s cover
nox > Running session cover
nox > /usr/bin/python -m virtualenv /home/srashid/workspace/github/google-cloud-python/.nox/cover -p python3.6
nox > Command /usr/bin/python -m virtualenv /home/srashid/workspace/github/google-cloud-python/.nox/cover -p python3.6 failed with exit code 3:
The path python3.6 (from --python=python3.6) does not exist
nox > Session cover failed.
```

i tried changing the ```session.interpreter=``` value to python27  in `logging/nox.py` but that gives other errors

tseaver:
Nice catch!

PR #5897, opened 2018-09-06
Storage: Add 'stacklevel=2' to deprecation warnings.

tseaver:
Add test assertion for warning from 'Bucket.location' setter.


tseaver:
@frankyn FWIW, `stacklevel=2` makes the warning report the file / line number of the *caller* of the deprecated method, which is much more useful than the file / line number where the warning was emitted.

PR #5900, opened 2018-09-07
Storage: Harden systest teardown further.

tseaver:
See: https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/7930


tseaver:
Merging as a trivial, testing-only fix.

PR #5904, opened 2018-09-08
Add grpcio-gcp dependency for cloud spanner

WeiranFang:
Add grpcio-gcp as extra dependency in google-api-core
Add grpcio-gcp as hard dependency in google-cloud-spanner

### DIFF
--- a/spanner/setup.py
+++ b/spanner/setup.py
@@ -30,7 +30,7 @@ version = '1.4.0'
 release_status = 'Development Status :: 5 - Production/Stable'
 dependencies = [
     'google-cloud-core<0.29dev,>=0.28.0',
-    'google-api-core[grpc, grpcio-gcp]<2.0.0dev,>=1.1.0',
+    'google-api-core[grpc, grpcio-gcp]<2.0.0dev,>=1.4.0',
     'grpc-google-iam-v1<0.12dev,>=0.11.4',
 ]
 extras = {
--- a/spanner/setup.py
+++ b/spanner/setup.py
@@ -30,7 +30,7 @@ version = '1.4.0'
 release_status = 'Development Status :: 5 - Production/Stable'
 dependencies = [
     'google-cloud-core<0.29dev,>=0.28.0',
-    'google-api-core[grpc]<2.0.0dev,>=1.1.0',
+    'google-api-core[grpc, grpcio-gcp]<2.0.0dev,>=1.1.0',
     'grpc-google-iam-v1<0.12dev,>=0.11.4',
 ]
 extras = {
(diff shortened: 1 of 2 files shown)


tseaver:
Making `grpcio-gcp` a hard dependency of `google-cloud-spanner` seems like it involves some risk:  I don't know how well that package is supported across all the platforms, but for instance the [latest version]() has a Python2-only wheel as its only binary distribution.  I can't find its repository, either.

@snehashah16

theacodes:
I'm going to send a PR to fix the wheel situation, then we can merge this.

The repo is at https://github.com/GoogleCloudPlatform/grpc-gcp-python

spanner/setup.py
```diff
@@ -30,7 +30,7 @@
 release_status = 'Development Status :: 5 - Production/Stable'
 dependencies = [
     'google-cloud-core<0.29dev,>=0.28.0',
-    'google-api-core[grpc]<2.0.0dev,>=1.1.0',
+    'google-api-core[grpc, grpcio-gcp]<2.0.0dev,>=1.1.0',
```
theacodes: We'll need to raise the minimum version here to the *next* release of api-core.

(Also this is the first time I've seen multiple extras used, does this even work?)

theacodes:
Universal wheel PR: https://github.com/GoogleCloudPlatform/grpc-gcp-python/pull/99

spanner/setup.py
theacodes: Let's go ahead and update it in this PR.

tseaver: The first example in the [PEP 508 specification section](https://www.python.org/dev/peps/pep-0508/#specification) shows multiple extras:

```
requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "2.7"
```

pip doesn't complain, but it may not do the right thing either (as with other extras).

WeiranFang: ok. Just to make sure, the next release of api-core should be 1.3.1 , correct?

theacodes: Let's do 1.4.0.

tseaver: > Let's do 1.4.0

Right.  The new extra is definitely a user-visible "feature" of `api_core`:  they won't be able to install `google-cloud-apicore[grpcio-gcp]` while downgrading.

WeiranFang:
Should I bump api core and cloud spanner versions in this PR, or leave them for you guys when they are ready to release?

theacodes:
No, releasetool will handle that.

theacodes:
Yes, once we merge this. I'll try to do this tomorrow morning. :)

tseaver:
We can't get tests passing for this PR until `api_core/setup.py` has the new version number.

WeiranFang:
I think we should separate this into 2 pr, one for api core and another one for spanner. So we can release api core before spanner, correct?

WeiranFang:
I see, thanks for the clarification!

PR #5915, opened 2018-09-11
Re-generate library using speech/synth.py

danoscarmike:
Incorporate new changes to Speech v1 per [latest commit](https://github.com/googleapis/googleapis/commit/098a5976deee6307c2028bf8ab66e8002f8b4a2b) to googleapis.

### DIFF
--- a/speech/README.rst
+++ b/speech/README.rst
@@ -1,14 +1,14 @@
-Python Client for Google Cloud Speech API (`Beta`_)
-====================================================
+Python Client for Cloud Speech API (`Alpha`_)
+=============================================
 
-`Google Cloud Speech API`_: Google Cloud Speech API.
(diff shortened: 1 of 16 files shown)


danoscarmike:
@theacodes seems to have overwritten the README pretty heavily -- reverting from Beta to Alpha -- that's not expected is it?

theacodes:
@danoscarmike probably not. We can write a quick replace in synth.py to fix that, something like:

```
s.replace('README.rst', '`Alpha`', '`Beta`')
```

danoscarmike:
Done and regenerated. PTAL.

theacodes:
Thanks for doing this, @danoscarmike!

theacodes:
@tseaver the solution there is to selectively copy files and use replacements as needed.

PR #5937, opened 2018-09-11
Docs: Add new alpha libraries to README

tseaver:
- Move reference links closer to point of use, and sort for ease of comparison to match lists.
- Normalize PyPI URLs.


tseaver:
Note that `asset` and `oslogin` don't have PyPI releases, so I didn't add them (they also don't have `cloud.google.com` links that I can find).

PR #5939, opened 2018-09-11
Declutter sidebar by supplying explict short titles.

tseaver:
Also, fix up 'resource_manager' docs to avoid propagating sub-menus.


tseaver:
@theacodes I expect that the short titles I picked might be open to debate (because "branding") but they have to be better than what we have now.

Issue #5876, opened 2018-08-31
BigTable: google-cloud-bigtable 0.30.0 throws InvalidChunk error while iterating through a large collection


xEcEz:
Python version: 3.6.3
google-cloud-bigtable 0.30.0

After switching to 0.30.0 and adapting my code to use the newly integrated iter function of PartialRowsData, I can't iterate through a large collection without the InvalidChunk error being thrown.

Sample code:
```
rows = table.read_rows()
for r in row:
...
```

tseaver:
@xEcEz Can you please provide the full traceback?  Also, can you guesstimate the size of the table (number of rows, column families, etc.)?

xEcEz:
Stacktrace:
```
at _validate_cell_data_new_row (/usr/lib/python3.6/site-packages/google/cloud/bigtable/row_data.py:517)
at _validate_cell_data (/usr/lib/python3.6/site-packages/google/cloud/bigtable/row_data.py:505)
at __iter__ (/usr/lib/python3.6/site-packages/google/cloud/bigtable/row_data.py:474)
at readAll (/scripts/bigtable_manager.py:106)
at call (/usr/lib/python3.6/site-packages/tenacity/__init__.py:361)
at __get_result (/usr/lib/python3.6/concurrent/futures/_base.py:384)
at result (/usr/lib/python3.6/concurrent/futures/_base.py:425)
at reraise (/usr/lib/python3.6/site-packages/tenacity/__init__.py:167)
at iter (/usr/lib/python3.6/site-packages/tenacity/__init__.py:331)
at call (/usr/lib/python3.6/site-packages/tenacity/__init__.py:358)
at wrapped_f (/usr/lib/python3.6/site-packages/tenacity/__init__.py:292)
at retryHandler (/scripts/bigtable_manager.py:149)
at getAllRanksTiers (/scripts/bigtable_manager.py:268)
at run (/scripts/queue_processor_update.py:68)
at _bootstrap (/usr/lib/python3.6/multiprocessing/process.py:258)
```

Table size: 2.5M approx.

sduskis:
@AVaksman, can you please take a look at this?

sduskis:
I've been looking at this.  `row_data.py:517` indicates that the client sees row_keys that are out of order. 
 v0.29.0 and before didn't have that validation.

[...]

PR #5963, opened 2018-09-12
Bigtable: refactoring read_rows infrastructure

sduskis:
This work is on behalf of understanding the issues behind issue #5876 where InvalidChunk is thrown due to row keys being processed out of order.
About this PR
- introducing `PartialRowsData._process_chunk` to break up logic in `__iter__`, and allow for additional testability.  `__iter__` is responsible to get chunks, call `_process_chunk`, and manage `last_scanned_row_key`.
- removed `_copy_from_current`
- _state is now explicitly set
- fixed a bug in `PartialRowsData.counter` which was counting chunks instead of rows.  This counter is used in reconstructing requests for "smart retries" that update the request on transient exceptions

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -483,7 +483,6 @@ class PartialRowsData(object):
             self._copy_from_previous(self._cell)
             self._validate_cell_data_new_cell()
         else:
-            self._validate_cell_data_cell_in_progress(chunk)
             self._cell.append_value(chunk.value)
 
     def _validate_cell_data_new_cell(self):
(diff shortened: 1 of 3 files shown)


sduskis:
@tseaver, CircleCI doesn't like this setup, but we need it in order to run system tests.

tseaver:
@sduskis I rebased against the current master, which should allow the tests to run.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -440,86 +422,78 @@ def __iter__(self):
             except StopIteration:
                 break
 
-            self._counter += 1
```
tseaver: `_counter` is documented in `__init__()` as a count of responses processed, not rows returned (as it seems to be below).

sduskis: I'll update the documentation.  `_counter` affects retries at the row level, not the "chunk" or "response" level.

sduskis:
@tseaver, CircleCI still doesn't like this PR :(

sduskis:
@tseaver, Circle is now happy.  How can we kick off system tests?

sduskis:
@tseaver, would it help if I made smaller PRs instead of this large one?

tseaver:
@sduskis The system tests *did* run in https://circleci.com/gh/GoogleCloudPlatform/google-cloud-python/8122 (because the branch is local to this repository).

PR #5996, opened 2018-09-17
Redirect spanner 'usage.html' -> 'index.html'.

### LINKED ISSUE #5995: Incoming link broken on Spanner docs
It seems #5938 renamed `usage.html` to `index.html`, but didn't setup any redirects. This is breaking some incoming links, unfortunately.

tseaver:
Closes #5995.


tseaver:
@theacodes I've added similar redirects for the other affected APIs (bigquery was the source of the original fix).

PR #6004, opened 2018-09-17
Videointelligence: Add v1p2beta1 to synth.py

danoscarmike:

PR #6007, opened 2018-09-18
Docs: debug broken docs build

tseaver:
- Run docs build during nox w/ better debug output (the colorized mode tends to hide some info).
- Fix loop boundary bug in copying static HTML files.
- Log each static HTML file copied.


tseaver:
Merging w/o review to fix docs build on `master` (and for PR #6004).

PR #6025, opened 2018-09-19
Bigtable:Adding __eq__ and __ne__ methods for row set

sumit-ql:
1. __eq__ and __ne__ methods for RowSet and RowRange.
2. Unit tests.

### DIFF
--- a/bigtable/google/cloud/bigtable/row_set.py
+++ b/bigtable/google/cloud/bigtable/row_set.py
@@ -29,6 +29,35 @@ class RowSet(object):
         self.row_keys = []
         self.row_ranges = []
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        if len(other.row_keys)!= len(self.row_keys):
+            return False
+
+        if len(other.row_ranges)!= len(self.row_ranges):
+            return False
+
+        if not sorted(other.row_keys) == sorted(self.row_keys):
+            return False
+
+        row_ranges = list(self.row_ranges)
+
+        for each in other.row_ranges:
+            try:
+                row_ranges.remove(each)
+            except ValueError:
+                return False
+
+        if len(row_ranges) > 0:
+            return False
+
+        return True
+
+    def __ne__(self, other):
+        return not self == other 
+
     def add_row_key(self, row_key):
         """Add row key to row_keys list.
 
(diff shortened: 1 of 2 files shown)


bigtable/google/cloud/bigtable/row_set.py
```diff
@@ -29,6 +29,32 @@ def __init__(self):
...
+            return NotImplemented
+
+        if len(other.row_keys) != len(self.row_keys):
+            return False
+
+        if len(other.row_ranges) != len(self.row_ranges):
+            return False
+
+        if not sorted(other.row_keys) == sorted(self.row_keys):
+            return False
```
tseaver: Sorting might be a little expensive.  Can we achieve the same goal by creating sets?  E.g.:

```python
        if not set(other.row_keys) == set(self.row_keys):
            return False
```

bigtable/google/cloud/bigtable/row_set.py
```diff
@@ -29,6 +29,32 @@ def __init__(self):
...
+
+        row_ranges = list(self.row_ranges)
+
+        for each in other.row_ranges:
+            try:
+                row_ranges.remove(each)
+            except ValueError:
+                return False
```
tseaver: This seems more complicated than necessary, and is likely slow (raising / catching exceptions is lots slower than normal code).  I think we can use sets here as well:

```python
        if not set(other.row_ranges) == set(self.row_ranges):
            return False
```

bigtable/google/cloud/bigtable/row_set.py
sumit-ql: Yes, we can use the set here. Updated the code.

bigtable/google/cloud/bigtable/row_set.py
sumit-ql: Added __hash__() to make RowRange hashable. Now, we can use set while comparing row_ranges.

bigtable/google/cloud/bigtable/row_set.py
```diff
@@ -138,6 +133,24 @@ def __init__(self, start_key=None, end_key=None,
...
+
+        Returns:
+            Tuple[str]: The contents of this :class:`.RowRange`.
+        """
+        return (
+            self.start_key,
+            self.start_inclusive,
+            self.end_key,
+            self.end_inclusive,
+        )
```
tseaver: Would it make sense to use `_key` below in `_eq`, as well?

sumit-ql: Yes. Updated the code, thanks.

PR #6030, opened 2018-09-19
Don't throw an error if content_type is not set.

### LINKED ISSUE #5834: Storage: Blob.compose doesn't use default content_type
[`Blob.compose`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/51ba7e148fc76e07018dc60192219fb28ff628b0/storage/google/cloud/storage/blob.py#L1363-L1381) does not use the [`self._get_content_type` method](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/51ba7e148fc76e07018dc60192219fb28ff628b0/storage/google/cloud/storage/blob.py#L605-L633), which means Blob.compose raises an error if the `content_type` is not explicitly specified. This is [documented behavior](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/51ba7e148fc76e07018dc60192219fb28ff628b0/storage/google/cloud/storage/blob.py#L1377-L1381), but is it intended behavior? Should compose use `self._get_content_type` to use the default type, or perhaps infer the type from the blobs being composed?

rasmi:
Fixes #5834.
Content-type is no longer required by the GCS API and defaults to application/octet-stream if not specified.
See: https://github.com/googleapis/nodejs-storage/issues/166#issuecomment-378418954

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -1374,13 +1374,7 @@ class Blob(_PropertyMixin):
                       ``NoneType``
         :param client: Optional. The client to use.  If not passed, falls back
                        to the ``client`` stored on the blob's bucket.
-
-        :raises: :exc:`ValueError` if this blob does not have its
-                 :attr:`content_type` set.
         """
-        if self.content_type is None:
-            raise ValueError("Destination 'content_type' not set.")
-
         client = self._require_client(client)
         query_params = {}
 


tseaver:
@rasmi Thanks for the fix!  It needs a unit test fixups, too.  Fortunately, I actually submitted a #6031  with the same change, but unit tests fixed up and a new system test.

rasmi:
Great, thank you!

PR #6046, opened 2018-09-21
Revert "Resource Manager: bump API version from 'v1beta1' to 'v1'."

tseaver:
Reverts GoogleCloudPlatform/google-cloud-python#6041

We need to review more deeply.  Per @enriquejosepadilla's testing, the changes in the API from `v1beta1` -> `v1` are breaking for us.


tseaver:
Aaaarght!  That is not what I asked for.

PR #6047, opened 2018-09-21
Resource Manager: 'v1' API is not forward compatible.

tseaver:
Reverts #6041.

We need to review more deeply. Per @enriquejosepadilla's testing, the changes in the API from `v1beta1` -> `v1` are breaking for us.


tseaver:
Merging w/o review:  this is a one-liner, reverting an over-eager PR.

PR #6049, opened 2018-09-21
Spanner: remove extra 'grpc_gcp' system tests.

tseaver:
#5904 made 'grpcio-gcp' an unconditional dependency.


tseaver:
Merging w/o review, as this is an uncontraversial, testing-only change.

PR #6088, opened 2018-09-24
Adding ability to give default QueryJobConfig to Client

### LINKED ISSUE #5183: BigQuery: ability to set default job / table / dataset configs in client class
The [location](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5148) is considered a global parameter (by the the BigQuery engineering team and the `bq` command-line tool), and we are tracking the feature to set a default location at https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5148.
There are other non-global properties where it would be useful to set a user-configurable default.

For example:
- Labels on jobs. It could be useful to set default labels so that folks can track more easily which jobs are getting run by which application.
- `use_legacy_sql` on QueryJobConfig. Right now this defaults to standard SQL syntax, but some people might prefer to use legacy SQL for existing queries.
- The `destination_encryption_configuration` in jobs and the `encryption_configuration` in tables. Maybe this one actually makes sense as a global parameter, but it's not considered on in the `bq` command-line tool.

blainehansen:
closes #5183 

This isn't a complete request (I need to do things like fill out the Contributor License Agreements, get coverage all the way to 100%, etc), but I wanted to begin the discussion to see if this was an acceptable direction. Only QueryJobConfig has been implemented, but the other types shouldn't be too difficult, since I've added a `_JobConfig.fill_from_default` method.


tseaver:
@blainehansen 

Making a PR from your `master` branch makes things hard for you over the long term:  it creates barriers to synchronizing with this repository's `master` over time.

WRT the test failures, there is missing line / branch coverage for:

```
google/cloud/bigquery/client.py               374      1    160      1    99%   1244, 1236->1244
```

which is the case where no `job_config` is passed.

blainehansen:
@tseaver Sorry about not creating a feature branch ha, I don't expect to maintain a fork for a long period, I was just hoping to include this feature, but I shouldn't have cut corners.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -122,11 +122,13 @@ class Client(ClientWithProject):
     """The scopes required for authenticating as a BigQuery consumer."""
 
     def __init__(
-            self, project=None, credentials=None, _http=None, location=None):
+            self, project=None, credentials=None, _http=None,
+            location=None, query_job_config=None):
```
tseaver: ISTM this argument should be named `default_query_job_config`.  It needs to be added to the class docstring above, too.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1187,7 +1189,9 @@ def extract_table(
         return extract_job
 
     def query(
-            self, query, job_config=None, job_id=None, job_id_prefix=None,
+            self, query,
+            job_config=None, override_job_config=False,
```
tseaver: Can we rename `override_job_config` to `merge_job_config`, and invert its meaning?  I think that would be clearer.  It needs to be added to the method docstring below.

bigquery/tests/unit/test_job.py
```diff
@@ -912,6 +912,32 @@ def test_ctor(self):
...
+        default_job_config = QueryJobConfig()
+        default_job_config.use_query_cache = True
+        default_job_config.maximum_bytes_billed = 2000
+
+        final_job_config = job_config.fill_from_default(default_job_config)
+        self.assertTrue(final_job_config.dry_run)
+        self.assertTrue(final_job_config.use_query_cache)
+        self.assertEqual(final_job_config.maximum_bytes_billed, 1000)
+
+        # case where job types differ
```
tseaver: Split this part out into a separate testcase method.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -819,6 +819,32 @@ def to_api_repr(self):
...
+        if self._job_type != default_job_config._job_type:
+            raise TypeError(
+                "attempted to merge two incompatible job types: "
+                + repr(self._job_type) + ', '
+                + repr(default_job_config._job_type))
+
+        new_job_config = self.__class__()
+        new_job_config._properties = copy.deepcopy(self._properties)
+
+        self_job_properties = self._properties[self._job_type]
```
tseaver: `self_job_properties` isn't needed (we can just test `new_job_properties` below).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -819,6 +819,32 @@ def to_api_repr(self):
...
+        new_job_properties = new_job_config._properties[self._job_type]
+        default_job_properties = default_job_config._properties[self._job_type]
+        for key in list(default_job_properties.keys()):
```
tseaver: This would be simpler / more idiomatic as `for key in default_job_properties:`

bigquery/google/cloud/bigquery/job.py
```diff
@@ -819,6 +819,32 @@ def to_api_repr(self):
...
+            if not self_job_properties.get(key):
```
tseaver: This would be more idiomatic as `if key not in new_job_properties:`

bigquery/google/cloud/bigquery/job.py
blainehansen: Ah yes, I'm stuck in javascript mode ha.

bigquery/google/cloud/bigquery/client.py
blainehansen: :+1: in regards to adding it to the docstring.

Just to clarify about `override_job_config`, I'll flip the argument default to also be `True` correct? It seems like the default behavior should be to use the default you've previously passed, and it would seem annoying to clarify in every `query` call that that's what you want.

bigquery/google/cloud/bigquery/client.py
tseaver: Tests need to accommodate this renaming, too.

bigquery/google/cloud/bigquery/client.py
tseaver: Tests need to accommodate this renaming, too.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1202,6 +1209,13 @@ def query(
         Keyword Arguments:
             job_config (google.cloud.bigquery.job.QueryJobConfig):
                 (Optional) Extra configuration options for the job.
+            merge_job_config (bool):
```
tswast: I hadn't thought to add an option to turn off the default overrides. We can disambiguate between a value in the `job_config` which is unset (no key in the dictionary) versus a value which is explicitly set to `None` (key in dictionary, with value of `None`), so this may not be necessary. Folks can override defaults by setting the value in the specified config.

What was the use case you are proposing for this argument?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -819,6 +819,35 @@ def to_api_repr(self):
...
+        if self._job_type != default_job_config._job_type:
+            raise TypeError(
+                "attempted to merge two incompatible job types: "
+                + repr(self._job_type) + ', '
+                + repr(default_job_config._job_type))
+
+        new_job_config = self.__class__()
+        new_job_config._properties = copy.deepcopy(self._properties)
+
+        new_job_properties = new_job_config._properties[self._job_type]
```
tswast: I don't think we need to restrict to `self._job_type`. We miss a handful of properties if we do this (such a labels).

bigquery/google/cloud/bigquery/job.py
```diff
@@ -819,6 +819,35 @@ def to_api_repr(self):
...
+        default_job_properties = default_job_config._properties[self._job_type]
+        for key in default_job_properties:
```
tswast: You could simplify this (remove the loops) by starting with the default properties and then calling [update](https://docs.python.org/3/library/stdtypes.html#dict.update) to set those from the new job.

```
default_job_properties = copy.deepcopy(default_job_config._properties)
default_job_properties.update(self._properties)
new_job_config._properties = default_job_properties
return new_job_config
```

bigquery/google/cloud/bigquery/client.py
blainehansen: Yes that was exactly the thinking. I wasn't sure if None was a valid value for some of these keys, but now as I look through at many of the `setters` it seems it won't get made (and you would know better than I would ha). I'll try it and see if the tests pass.

bigquery/google/cloud/bigquery/job.py
blainehansen: It seems to me that we'll need two `update`s, one for the top layer (grabbing things like labels and whatever else), and another for the nested `_job_type` keys since `update` doesn't recurse.

```python
default_job_properties = copy.deepcopy(default_job_config._properties)
default_job_properties.update({ key: self._properties[key] for key in self._properties if key != self._job_type })
default_job_properties[self._job_type].update(self._properties[self._job_type])
new_job_config._properties = default_job_properties
return new_job_config
```

bigquery/google/cloud/bigquery/client.py
blainehansen: It works, but I'm a little torn about the convenience of having to manually set everything to `None` in order to override. It definitely leads to a cleaner function signature, but if they have a default with a lot of values then that might get a little tedious. I'll leave it up to you guys which version we go with.

tswast: I'd prefer to leave it out for now. If someone creates a client with a whole bunch of default properties, I think it's natural that they'd have to override them. (I know you are thinking of use cases where the person running the queries isn't necessarily the person that constructs the client, but hopefully you aren't setting too many default properties in that case.)

blainehansen: :+1:

blainehansen:
:+1: yeah making that private is definitely a better choice :smile:

blainehansen:
:confetti_ball: :confetti_ball: :confetti_ball: 

Thanks everyone!

PR #6089, opened 2018-09-24
Add kokoro for videointelligence and remove videointelligence from CircleCI

busunkim96:


tseaver:
My only concern here is that some version of this PR caused a [CI failure](https://source.cloud.google.com/results/invocations/e79d6ef0-b191-4448-9b45-be06b1f8bc73) for an unrelated PR (#6091).

busunkim96:
I believe the failure on #6091 is the result of videointelligence having been added to the kokoro config in google3 yesterday. Kokoro is looking for the videointelligence build configs, but they won't exist until this PR is merged. The Fusion build log shows 

```
[05:15:36][ERROR] Failed to get build config: Build config google-cloud-python/.kokoro/presubmit/videointelligence.cfg does not exist
```

PR #6103, opened 2018-09-25
Accomodate payload-less log entries.

### LINKED ISSUE #6063: Logging: payload should be parsed as optional
google-cloud-python requires a valid value for the `payload` field in a log entry.

However, the proto shows its optional
 - [https://github.com/googleapis/googleapis/blob/master/google/logging/v2/log_entry.proto#L68](https://github.com/googleapis/googleapis/blob/master/google/logging/v2/log_entry.proto#L68)

and i've verified i can emit a log entry w/ `payload=None` and simply the http_request value (this is used in logs groups)

Looks like its checked for a value here

* https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/entries.py#L128

```python
        if logger is None:
            logger_name = logger_name_from_path(logger_fullname)
            logger = loggers[logger_fullname] = client.logger(logger_name)
        payload = resource[cls._PAYLOAD_KEY]
        insert_id = resource.get('insertId')
        timestamp = resource.get('timestamp')
```

* https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/_helpers.py#L47

```python
    if 'textPayload' in resource:
        return TextEntry.from_api_repr(resource, client, loggers)
    elif 'jsonPayload' in resource:
        return StructEntry.from_api_repr(resource, client, loggers)
    elif 'protoPayload' in resource:
        return ProtobufEntry.from_api_repr(resource, client, loggers)
    raise ValueError('Cannot parse log entry resource.')
```

tseaver:
Closes #6063.

/cc  @salrashid123

### DIFF
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -16,6 +16,7 @@
 
 import requests
 
+from google.cloud.logging.entries import EmptyEntry
 from google.cloud.logging.entries import ProtobufEntry
 from google.cloud.logging.entries import StructEntry
 from google.cloud.logging.entries import TextEntry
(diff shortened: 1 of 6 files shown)


tseaver:
@theacodes, @crwilcox PTAL

PR #6104, opened 2018-09-25
Clarify GCS URL signing limitations on GCE

zulkifal:
The line 'If you are on Google Compute Engine, you can't generate a signed URL' wrongly implies URL generation is impossible from GCE. Adding clarification that this line only holds true when GCE is using GCE service account.


zulkifal:
I signed it!

tseaver:
@zulkifal Thanks for the patch!

PR #6105, opened 2018-09-25
Release dlp 0.8.0

crwilcox:


crwilcox:
@tseaver it reformated links to changes and I just pushed an update. It looks like I used an old version of the releasetool

PR #6116, opened 2018-09-26
Add type-checking via pytype to api_core.

rchen152:
Type-checks all of api_core outside of tests with [pytype](https://github.com/google/pytype).
* Adds pytype options to setup.cfg so the code can be type-checked with `pytype -V{major.minor}`.
* Runs `pytype -V3.5` and `pytype -V3.6` under nox.
Manually tested by pip-installing `pytype` and `nox-automation` in a Python 3.6 virtualenv.

@theacodes cc-ing you on this PR as you requested

### DIFF
--- a/api_core/nox.py
+++ b/api_core/nox.py
@@ -15,7 +15,8 @@
 from __future__ import absolute_import
 import os
 
-import nox
+# https://github.com/google/importlab/issues/25
+import nox  # pytype: disable=import-error
 
 
 @nox.session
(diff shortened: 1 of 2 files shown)


api_core/nox.py
```diff
@@ -108,6 +109,20 @@ def lint_setup_py(session):
         'python', 'setup.py', 'check', '--restructuredtext', '--strict')
 
 
+# No 2.7 due to https://github.com/google/importlab/issues/26.
+# No 3.7 because pytype supports up to 3.6 only.
+@nox.session
+@nox.parametrize('py', ['3.5', '3.6'])
```
theacodes: Let's just test against `3.6` for now.

rchen152: Done.

theacodes:
Thank you, @rchen152!

PR #6126, opened 2018-09-27
Add kokoro jobs

busunkim96:
Add remaining kokoro jobs.


theacodes:
@tseaver @crwilcox this moves us over to kokoro. please keep an eye on builds and less us know if we need to fix anything.

theacodes:
Not yet.

Issue #6128, opened 2018-09-27
PubSub protobuf dependency requirements incorrect


aarontp:
I think the minimum version for the `protobuf` package dependency is not correct.  google-cloud-python has the version requirements as protobuf>=3.0.0, but it fails to import when using version up to and including protobuf==3.3.0.  I'm not sure what the exact correct version is, but the last version of google-cloud-pubsub to work with protobuf==3.3.0 is google-cloud-pubsub==0.35.4.  I believe after this commit (https://github.com/GoogleCloudPlatform/google-cloud-python/commit/371333a51165e99d4d02876b1ef133618485b6fc#diff-29280288794caf553b0b008084a0e854), a protobuf version >3.3.0 is required:

Python version
```
$ python --version
Python 2.7.15rc1
```

Package versions:
```
$ pip list | grep -E '(cloud|protobuf)'
google-cloud-core                  0.28.1      
google-cloud-datastore             1.7.0       
google-cloud-pubsub                0.38.0      
google-cloud-storage               1.12.0      
protobuf                           3.3.0 
```

Getting a stack track just importing pubsub (in ipython here)
```
In [1]: from google.cloud import pubsub
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-8fba37b708ad> in <module>()
----> 1 from google.cloud import pubsub

/home/aaronpeterson/.local/share/virtualenvs/turbinia-docs-oOHvuNoj/local/lib/python2.7/site-packages/google/cloud/pubsub.py in <module>()
     17 from __future__ import absolute_import
     18 
---> 19 from google.cloud.pubsub_v1 import PublisherClient
     20 from google.cloud.pubsub_v1 import SubscriberClient
     21 from google.cloud.pubsub_v1 import types

/home/aaronpeterson/.local/share/virtualenvs/turbinia-docs-oOHvuNoj/local/lib/python2.7/site-packages/google/cloud/pubsub_v1/__init__.py in <module>()
     15 from __future__ import absolute_import
     16 
---> 17 from google.cloud.pubsub_v1 import types
     18 from google.cloud.pubsub_v1 import publisher
     19 from google.cloud.pubsub_v1 import subscriber

/home/aaronpeterson/.local/share/virtualenvs/turbinia-docs-oOHvuNoj/local/lib/python2.7/site-packages/google/cloud/pubsub_v1/types.py in <module>()
     28 
[...]

PR #6132, opened 2018-09-28
Pin minimum protobuf dependency to 3.4.0.

tseaver:
See: #6128.

We will need to make an `api_core-1.5.0` release (or maybe `1.4.1`) as soon as this is merged.


tseaver:
@theacodes I verified manually that `pubsub-3.4.0` was the lowest version which made the pubsub unit tests pass.  We could go ahead and pin it to the current version (3.6.1) if you think that is better.

PR #6118, opened 2018-09-26
Logging: support request-correlated logging in App Engine standard python37 runtime

myelin:
The python37 runtime provides the GOOGLE_CLOUD_PROJECT environment variable instead of GCLOUD_PROJECT, and for log messages to be correlated with the request in StackDriver, their 'trace' value has to be passed as a top level member of the log record rather than as an appengine.googleapis.com/trace_id label.

PR #6135, opened 2018-09-28
Release logging 1.8.0

crwilcox:
This pull request was generated using releasetool.

PR #6138, opened 2018-09-28
Logging: test both GCLOUD_PROJECT and GOOGLE_CLOUD_PROJECT env vars

myelin:
Updated TestAppEngineHandler.test_constructor to test that the project environment variables from GAE flex and GAE standard are both correctly handled, as per @tseaver's request in #6118.


tseaver:
I want to hold off merging until @crwilcox figures out what is broken in logging in PR #6135.

PR #6153, opened 2018-10-03
Fix class reference

joar:
This stops PyCharm from complaining that my `Client` instance isn't of type `client.py`


tseaver:
@joar Thanks for the patch.  Can you please do the CLA-signing dance?

joar:
@tseaver I tried once and it looked successful, I'll try again.

joar:
Ah, I just needed to reply to trigger a re-check
> Once you've signed (or fixed any issues), please reply here (e.g. I signed it!) and we'll verify it.
> https://github.com/googleapis/google-cloud-python/pull/6153#issuecomment-426718886

tseaver:
@theacodes Any clue why kokoro did not run here?

Issue #6154, opened 2018-10-03
Asset:  release asset 0.2.0


tseaver:
The original release (#5988) used `google-cloud-cloudasset` as the PyPI name.  We need a new release to push our preferred name.

@theacodes RFC:  do we need to make a `google-cloud-cloudasset 0.1.post1` release which emits a deprecation warning and points to `google-cloud-asset`?

tseaver:
Via #6160 .

PR #6158, opened 2018-10-03
Docs: Add / fix badges for PyPI / versions.

tseaver:
We could still do with another normalization pass (e.g., standardize on use of `|alpha|` / `|beta|` badges).

~~Note that the change to `asset/README.rst` will have broken badges until we land a release of `google-cloud-asset` (vs. `google-cloud-cloudasset`, see #6154).~~  `oslogin/README.rst` is likewise going to show broken badges, because it has never been released.

### DIFF
--- a/api_core/README.rst
+++ b/api_core/README.rst
@@ -7,8 +7,8 @@ This library is not meant to stand-alone. Instead it defines
 common helpers used by all Google API clients. For more information, see the
 `documentation`_.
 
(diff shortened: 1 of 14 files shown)


tseaver:
Rebased to pick up the `PROJECT_ID` fix from #6167.

tseaver:
@crwilcox, @theacodes This should be ready for review.

Issue #6055, opened 2018-09-21
BigQuery: Potentially more Pythonic options for the API?


max-sixty:
I _thought_ I'd discussed this with @tswast, and he'd even generously added a PR; but after ten mins of searching GH I've come up empty on both issues and PRs, so posting here. 
@tswast lmk if my memory is correct and we have discussed this!

I appreciate API design is difficult and there are tradeoffs between verbosity vs explicitness, and consistency within a product vs language-specific adjustments. The bar should be high for people offering criticism when they can only see a subset of the revelative information.
But I do frequently use the BQ python API and find it fairly awkward and un-pythonic, as though I were writing Java.

Here's an example, straight from the docs:

```
from google.cloud import bigquery
client = bigquery.Client()
dataset_id = 'my_dataset'  # replace with your dataset ID
table_id = 'my_table'  # replace with your table ID
table_ref = client.dataset(dataset_id).table(table_id)
table = client.get_table(table_ref)  # API request
```

I would love to be able to write this, and receive a table object. 

```
client = bigquery.Client()
client.table(dataset='ds', table='tbl')
```

There are some points a level down (the `.table` object doesn't return a `Table` object while a dataset object does return a `Dataset` object / do we need a `TableReference` class in place of a string, etc), but they all center around the ergonomics of the API, particularly in respect to Python.

Thank you as ever for a wonderful product, and appreciate any thoughts on whether I'm making mistakes here.

tswast:
In our previous conversation, I added a `from_string()` method to the `DatasetReference` and `TableReference` classes. https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5255

You're right that the BigQuery client does still feel overly verbose. I find it a problem that `from_string()` can't account for the default project on the client.

The reason we have the `*Reference` classes is that the BigQuery API has such resources in the REST responses. I wanted to be super clear when the API only returns a pointer to a table rather than a full table resource.

It's a historical artifact that `client.dataset()` and `dataset.table()` return a reference and not an actual Dataset or Table. In retrospect, I probably should have renamed or removed those methods in the 1.0 redesign project.

An idea:

- Whereever Client methods accept a `*Reference`, also allow a string. This gets a little muddy in the library code, but I think so long as it's documented well it'd be okay.

PRs welcome!

tswast:
cc @shollyman @alixhami

max-sixty:
Hi Tim!

[...]

PR #6164, opened 2018-10-04
Allow string in addition to DatasetReference / TableReference in Clie…

tswast:
…nt methods.

Whereever a client method accepts a DatasetReference or TableReference,
it now also accepts a string. The project ID can be omitted from the
string and the Client's default project is used instead.

Also, to make it easier to share code across client methods for creating
XReference classes:

* Add default_project argument to DatasetReference.from_string
* Add default_project argument to TableReference.from_string

Towards #6055. Not closes because we should add similar logic to the `create_dataset` and `create_table` methods, but they need to be modified to also accept a reference. My thought there is that we can create an empty table/dataset resource when a string or reference is passed in.

@max-sixty Please take a look.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -339,14 +339,14 @@ class Client(ClientWithProject):
         """Fetch the dataset referenced by ``dataset_ref``
 
         Args:
-            dataset_ref (Union[
-                :class:`google.cloud.bigquery.dataset.DatasetReference`,
-                str
+            dataset_ref (Union[ \
+                :class:`~google.cloud.bigquery.dataset.DatasetReference`, \
+                str, \
             ]):
                 A reference to the dataset to fetch from the BigQuery API.
                 If a string is passed in, this method attempts to create a
                 dataset reference from a string using
-                :func:`google.cloud.bigquery.dataset.DatasetReference.from_string`.
+                :func:`~google.cloud.bigquery.dataset.DatasetReference.from_string`.
             retry (:class:`google.api_core.retry.Retry`):
                 (Optional) How to retry the RPC.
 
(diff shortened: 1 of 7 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1121,6 +1196,14 @@ def copy_table(
 
         job_ref = job._JobReference(job_id, project=project, location=location)
 
+        if isinstance(sources, str):
+            sources = TableReference.from_string(
+                sources, default_project=self.project)
+
+        if isinstance(destination, str):
+            destination = TableReference.from_string(
+                destination, default_project=self.project)
```
max-sixty: Nice!!
I've seen this implemented equivalent to `destination = maybe_coerce(destination)`, where `maybe_coerce` does the type checking; this is more explicit at the cost of the `if`s.

max-sixty:
Thank you @tseaver ! 

I find this _far_ more pythonic and easier to use (e.g. the proportion of instances I'll use the API without pulling up the docs is much higher). Interested what others think.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -338,34 +338,53 @@ def _call_api(self, retry, **kwargs):
...
         """Fetch the dataset referenced by ``dataset_ref``
 
-        :type dataset_ref:
-            :class:`google.cloud.bigquery.dataset.DatasetReference`
-        :param dataset_ref: the dataset to use.
-
-        :type retry: :class:`google.api_core.retry.Retry`
-        :param retry: (Optional) How to retry the RPC.
+        Args:
+            dataset_ref (Union[
```
tswast: This isn't rendering quite right. I think I need to add a trailing `\`.

tswast:
@dpebot merge when green

bigquery/google/cloud/bigquery/client.py
tswast: That's a good idea. I wish I'd thought of that before I added all these `if`s. :-)

PR #6168, opened 2018-10-04
BigQuery: fix swallowed error message

mistercrunch:
When getting a `Cannot query over table (...) without a filter over
column(s) (...)`, the re-raising happening gobbles up the error message
and the context is lost.

This may or may not be the right fix, but clearly highlights the issue
of `self._query_job.errors` being `None` in that particular context.
When that is the case, the only context available is that of
`DatabaseError` without any details which clearly isn't enough detail.

```
After this fix is applied (wrapping the exception into another exception) both stack traces are shown:
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/job.py", line 2685, in result
    super(QueryJob, self).result(timeout=timeout)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/job.py", line 697, in result
    return super(_AsyncJob, self).result(timeout=timeout)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/future/polling.py", line 115, in result
    self._blocking_poll(timeout=timeout)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/job.py", line 2659, in _blocking_poll
    super(QueryJob, self)._blocking_poll(timeout=timeout)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/future/polling.py", line 94, in _blocking_poll
    retry_(self._done_or_raise)()
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/retry.py", line 260, in retry_wrapped_func
    on_error=on_error,
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/retry.py", line 177, in retry_target
    return target()
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/future/polling.py", line 73, in _done_or_raise
    if not self.done():
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/job.py", line 2647, in done
    location=self.location)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/client.py", line 584, in _get_query_results
    retry, method='GET', path=path, query_params=extra_params)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/client.py", line 336, in _call_api
    return call()
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/retry.py", line 260, in retry_wrapped_func
    on_error=on_error,
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/api_core/retry.py", line 177, in retry_target
    return target()
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/_http.py", line 293, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.BadRequest: 400 GET https://www.googleapis.com/bigquery/v2/projects/fsql-production/queries/6ed337a4-e23c-4634-ba94-15254648bdfb?maxResults=0&location=US: Cannot query over table 'fsql-production.records.routes_stream' without a filter over column(s) 'created_at' that can be used for partition elimination

  File "/home/mbeauchemin/incubator-superset/superset/sql_lab.py", line 185, in execute_sql
    db_engine_spec.execute(cursor, query.executed_sql, async_=True)
  File "/home/mbeauchemin/incubator-superset/superset/db_engine_specs.py", line 376, in execute
    cursor.execute(query)
  File "/home/mbeauchemin/incubator-superset/env/lib/python3.4/site-packages/google/cloud/bigquery/dbapi/cursor.py", line 157, in execute
    raise exceptions.DatabaseError(self._query_job.errors or e)
google.cloud.bigquery.dbapi.exceptions.DatabaseError: 400 GET https://www.googleapis.com/bigquery/v2/projects/fsql-production/queries/6ed337a4-e23c-4634-ba94-15254648bdfb?maxResults=0&location=US: Cannot query over table 'fsql-production.records.routes_stream' without a filter over column(s) 'created_at' that can be used for partition elimination
```

### DIFF
--- a/bigquery/google/cloud/bigquery/dbapi/cursor.py
+++ b/bigquery/google/cloud/bigquery/dbapi/cursor.py
@@ -153,8 +153,8 @@ class Cursor(object):
         # Wait for the query to finish.
         try:
             self._query_job.result()
-        except google.cloud.exceptions.GoogleCloudError as e:
-            raise exceptions.DatabaseError(self._query_job.errors or e)
+        except google.cloud.exceptions.GoogleCloudError as exc:
+            raise exceptions.DatabaseError(exc)
 
         query_results = self._query_job._query_results
         self._set_rowcount(query_results)
--- a/bigquery/google/cloud/bigquery/dbapi/cursor.py
+++ b/bigquery/google/cloud/bigquery/dbapi/cursor.py
@@ -153,8 +153,8 @@ class Cursor(object):
         # Wait for the query to finish.
         try:
             self._query_job.result()
-        except google.cloud.exceptions.GoogleCloudError:
-            raise exceptions.DatabaseError(self._query_job.errors)
+        except google.cloud.exceptions.GoogleCloudError as e:
+            raise exceptions.DatabaseError(self._query_job.errors or e)
 
         query_results = self._query_job._query_results
         self._set_rowcount(query_results)


sumedhsakdeo:
Looks great. CC: @tswast

bigquery/google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -153,8 +153,8 @@ def execute(self, operation, parameters=None, job_id=None):
         # Wait for the query to finish.
         try:
             self._query_job.result()
-        except google.cloud.exceptions.GoogleCloudError:
-            raise exceptions.DatabaseError(self._query_job.errors)
+        except google.cloud.exceptions.GoogleCloudError as e:
+            raise exceptions.DatabaseError(self._query_job.errors or e)
```
tseaver: This change makes the type of the exception's `args` attribute ambiguous:  it will be either a list of mappings or a single exception.  I don't believe we should force users to guess here.

mistercrunch: @tseaver my main goal was to highlight the issue.

It also appears that `execptions.DatabaseError` doesn't document getting unconventional argument (the common assumption with Python exception is that they receive a string, in this case it appears that `self._query_job.errors` a list of dicts). I know that the DBAPI PEP (249) isn't clear about the values expected here (it says: `The values of these exceptions are not defined. They should give the user a fairly good idea of what went wrong, though.`).

If a list of dicts is expected, we should document it here and enforce that consistency:
https://github.com/googleapis/google-cloud-python/blob/0e2c54ce54909467b202a73f04fc3aa4a29a749b/bigquery/google/cloud/bigquery/dbapi/exceptions.py#L30

mistercrunch: For context, I'm working on integrating BigQuery with Superset. Superset integrates with most SQL-speaking database out there. You'll understand that we like when DBAPIs behave in a consistent way :)
https://github.com/apache/incubator-superset

tswast: I think we could probably forget `self._query_job.errors` entirely, as any error information should be present in the exception. I don't remember the reason I used the `errors` from the job resource to begin with.

tseaver: @tswast Any issues about backward-compatibility for folks who might be catching the `DatabaseError` if we just propagate the `GoogleCloudError` instead?

tswast: We do need to catch and throw as `DatabaseError` per PEP 249. https://www.python.org/dev/peps/pep-0249/#databaseerror

mistercrunch:
🔥 🕺 🔥

PR #6175, opened 2018-10-08
Use new Nox

### LINKED ISSUE #6170: Update Noxfiles to use new Nox
Nox recently made a [big change](https://github.com/theacodes/nox/blob/master/CHANGELOG.md#2018823) from declarative to imperative and broke the Noxfile API.

We need to update our Noxfiles to work with the new API and update our test scripts to install `nox` (instead of `nox-automation`) and invoke `python3 -m nox {args}`.

Most projects just need to do a few steps:

1. Remove usage of `session_{name}` sessions and use `@nox.session` instead.
2. Don't use `session.interpreter`, use `@nox.session(py='...')`
3. Don't use `nox.parametrize` for multiple Pythons, use `@nox.session(py=['..', '..'])`.

busunkim96:
Updates Noxfiles to use new Nox API. Closes #6170.


bigquery/noxfile.py
```diff
@@ -40,14 +40,14 @@ def default(session):
         session.install('-e', local_dep)
 
     # Pyarrow does not support Python 3.7
-    if session.interpreter == 'python3.7':
+    if session.virtualenv.interpreter == '3.7':
```
busunkim96: @theacodes What is the best way to get the python version in the new Nox API? This doesn't seem to work when `default` is run as its own session.

PR #6185, opened 2018-10-09
Add application_name for amending user agent strings

tswast:
This enables partner companies to use a public API to amend the user
agent strings when working with the BigQuery API.

Towards https://github.com/googleapis/google-cloud-common/issues/265

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -22,6 +22,7 @@ import gzip
 import os
 import uuid
 
+import pkg_resources
 import six
 
 from google import resumable_media
@@ -133,10 +134,27 @@ class Client(ClientWithProject):
         self._location = location
         self._default_query_job_config = default_query_job_config
 
+        # Set the USER_AGENT with the property handler for application_name.
+        self.application_name = ''
+
     @property
     def location(self):
         """Default location for jobs / datasets / tables."""
         return self._location
+
+    @property
+    def application_name(self):
+        """Application name for user agent in requests to Google APIs."""
+        return self._application_name
+
+    @application_name.setter
+    def application_name(self, value):
+        self._application_name = value
+        self._connection.USER_AGENT = Connection._USER_AGENT_TEMPLATE.format(
+            core_version=pkg_resources.get_distribution('google-cloud-core').version,
+            bq_version=pkg_resources.get_distribution('google-cloud-bigquery').version,
+            application_name=' ' + value,
+        )
 
     def get_service_account_email(self, project=None):
         """Get the email address of the project's BigQuery service account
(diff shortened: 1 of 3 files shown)


theacodes:
I would really rather we didn't do this until we have a formal design doc for adopting this across languages and libraries.

theacodes:
Especially considering BigQuery is GA.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -133,11 +134,28 @@ def __init__(
         self._location = location
         self._default_query_job_config = default_query_job_config
 
+        # Set the USER_AGENT with the property handler for application_name.
+        self.application_name = ''
```
tseaver: This part needs assertions in one or more of the constructor unit tests (to prevent regressions).

bigquery/tests/unit/test_client.py
```diff
@@ -209,6 +209,32 @@ def test__get_query_results_hit(self):
...
+        response.json.return_value = resource
+        client = self._make_one(
+            project=self.PROJECT, credentials=creds, _http=http)
+
+        # Check default user agent.
+        client.get_service_account_email()
+        args, kwargs = http.request.call_args
+        self.assertRegex(
+            kwargs['headers']['User-Agent'],
+            'gcloud-python/.+ google-cloud-bigquery/.+')
```
tseaver: Wouldn't we be better off just computing the actual string we expect here?

bigquery/tests/unit/test_client.py
```diff
@@ -209,6 +209,32 @@ def test__get_query_results_hit(self):
...
+
+        # Check amended user agent.
+        client.application_name = 'pandas-gbq/0.6.0'
+        client.get_service_account_email()
+        args, kwargs = http.request.call_args
+        self.assertRegex(
+            kwargs['headers']['User-Agent'],
+            'gcloud-python/.+ google-cloud-bigquery/.+ pandas-gbq/0\.6\.0')
```
tseaver: And likewise here?

bigquery/tests/unit/test_client.py
tseaver: For one thing, `assertRegex` is not available in Python 2.7.

tseaver:
```suggestion
        from pkg_resources import get_distribution
        api_core_version = get_distribution('google-api_core').version
        bq_version = get_distribution('google-cloud-bigquery').version 
        self.assertEqual(
            kwargs['headers']['User-Agent'],
            'gcloud-python/{}/google-cloud-bigquery/{}'.format(
                api_core_version, bq_version,
            ))
```

tswast:
Closing, as the plan is changed to use client_info class.

PR #6194, opened 2018-10-11
Update subscriber example in README to current patterns.

### LINKED ISSUE #6189: Pub/Sub QuickStart Guide is Outdated
Outdated info on the Pub/Sub Client Library landing page: https://github.com/googleapis/google-cloud-python/tree/master/pubsub:

If folks follow the subscriber guide to this step, they will get this error, which makes no sense. 

```
>>> subscription = subscriber.subscribe(subscription_name)
TypeError: subscribe() takes at least 3 arguments (2 given)
```

Additionally, everywhere else in the API reference, the import statement says `from google.cloud import pubsub_v1`. We need to standardize the import statements.

tseaver:
Closes #6189.

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -114,9 +114,9 @@ the topic, and subscribe to that, passing a callback function.
 .. code-block:: python
 
     import os
-    from google.cloud import pubsub
+    from google.cloud import pubsub_v1
 
-    subscriber = pubsub.SubscriberClient()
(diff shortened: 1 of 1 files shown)


tseaver:
@anguillanneuf 

> Could we also update `from google.cloud import pubsub_v1`?

See aae840a .

tseaver:
@anguillanneuf 

> Oops, approved too quickly. Would need changes on publisher code block too.

See  fee1735

PR #6198, opened 2018-10-11
Correction to the pip install command

Ben-Chapman:
The current `pip install` command incorrectly references `google-cloud-datastore`, this PR is correcting the `pip install` instructions to correctly reference `google-cloud-videointelligence`


Ben-Chapman:
@googlebot I signed it!

PR #6215, opened 2018-10-15
Disable test running in VPCSC restricted environment

McBrainy:
The environment variable name was chosen for consistency with the other env variable in the file; I'm open to changing it.


tseaver:
@crwilcox We missed actually running CI here, which leads to lint failures on master due to an [over-long line](https://github.com/McBrainy/google-cloud-python/blob/ce0b2916bd1a5e783aec07b4fdade80cc0bad5bc/storage/tests/system.py#L36).

PR #6236, opened 2018-10-17
Logging: allow setting name, args on default handler

### LINKED ISSUE #6206: Logging: set logName with google.cloud.logging.Client().get_default_handler()
According to https://github.com/googleapis/google-cloud-python/blob/master/logging/google/cloud/logging/client.py#L309 get_default_handler has no parameters.
But if I need to set a custom logName into stackdriver logs, I need to pass name="foo" both to AppEngineHandler and CloudLoggingHandler
(moreover, from what I understand there is no way to setup the ContainerEngineHandler)

tseaver:
Closes #6206.

### DIFF
--- a/logging/google/cloud/logging/client.py
+++ b/logging/google/cloud/logging/client.py
@@ -292,8 +292,11 @@ class Client(ClientWithProject):
         return self.metrics_api.list_metrics(
             self.project, page_size, page_token)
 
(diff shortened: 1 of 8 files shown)


tseaver:
@crwilcox ping?

tseaver:
Superseded by #6828.

PR #6237, opened 2018-10-17
Import `ndb-rewrite` history

dhermes:
This is an import of the history (`master`) from https://github.com/dhermes/ndb-rewrite

For discussion see: https://github.com/dhermes/ndb-rewrite/issues/38


dhermes:
Some notes:

1. All of this code has been reviewed by @theacodes or @crwilcox in the original repo
1. I'll need to add a Kokoro configuration for this subdirectory, but I think that should go in a follow-up (this PR is a literal copy of the existing source)
1. It'd be nice to have a place for rendered docs. Possibly putting them with the main docs, but in a way that the URL paths aren't linked to on the main page

PR #6238, opened 2018-10-17
Move usage guides to their own docs.

tswast:
The API reference links were buried behind pages of
code samples. Moving the how-to guides to their own
pages makes the documentation easier to navigate.


bigquery/docs/index.rst
```diff
@@ -1,408 +1,22 @@
 .. include:: /../bigquery/README.rst
 
 Using the Library
```
alixhami: The flow is a bit awkward here - There is a heading for performing a query and then a separate heading for "Using the Library". I think the "Using the Library" heading should be updated to say something like "More Examples".
image

bigquery/docs/howto/tables.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Upload table data from a file:
+
+.. literalinclude:: ../snippets.py
+   :language: python
+   :dedent: 4
+   :start-after: [START bigquery_load_from_file]
+   :end-before: [END bigquery_load_from_file]
+
+Load table data from Google Cloud Storage
```
alixhami: The heading logic is weird here. Why are the "Load table data from Google Cloud Storage" and "Customer Managed Encryption Keys" headers on the same level as the "Table Operations" header? I think we should either remove all of them or use headers that make more logical sense together.

bigquery/docs/index.rst
tswast: I changed the heading as suggested and broke it up into logical sub-sections.

bigquery/docs/howto/tables.rst
tswast: Done. Moved encryption stuff to it's own page and organized sections by operation type.

bigquery/docs/index.rst
```diff
@@ -1,408 +1,43 @@
...
-.. code-block:: bash
-
-   pip install --upgrade pandas
-
-Alternatively, you can install the BigQuery python client library with
-:mod:`pandas` by running:
-
-.. code-block:: bash
+More Examples
+-------------
```
alixhami: This should be an H3 (use `~` instead of `-`)

bigquery/docs/index.rst
```diff
@@ -1,408 +1,43 @@
...
 
-   pip install --upgrade google-cloud-bigquery[pandas]
+These guides cover how to use the ``google-cloud-bigquery`` library to use
+the Google BigQuery API. See the `official Google BigQuery how-to guides
+<https://cloud.google.com/bigquery/docs/how-to>`_ for more in-depth
+information and code samples.
 
-To retrieve query results as a :class:`pandas.DataFrame`:
+BigQuery Basics
+~~~~~~~~~~~~~~~
```
alixhami: The rest of the headers should probably be H2 (`-`) instead of H3 (`~`), otherwise they'll fall under the "Example Usage" header, which was the last H2. Or if you were intending for these to be part of "Example Usage", I'd move the "More Examples" header to the bottom of the "Example Usage" section.

bigquery/docs/howto/datasets.rst
```diff
@@ -0,0 +1,72 @@
...
+  which maps to the ACL inherited from its project.
+
+- A default table expiration period.  If set, tables created within the
+  dataset will have the value as their expiration period.
+
+See BigQuery documentation for more information on
+`Datasets <https://cloud.google.com/bigquery/docs/datasets>`_.
+
+
+Dataset operations
```
alixhami: It would be nice to have more specific header sections like the Table & Query pages.

bigquery/docs/index.rst
tswast: Done in https://github.com/googleapis/google-cloud-python/pull/6238/commits/af8a6622aa38f5926b8470bd2f982c861c0add8f

Moved it to the bottom.

bigquery/docs/howto/datasets.rst
tswast: Done in https://github.com/googleapis/google-cloud-python/pull/6238/commits/af8a6622aa38f5926b8470bd2f982c861c0add8f

tseaver:
I have a question, rather than a direct review comment:  would it be better to call the section `usage` than `howto`?  Right now we have a page, `/usage.html`, which redirects to `/index.html`:  that suggests that `usage` is part of the expected documentation pattern.

Issue #6231, opened 2018-10-17
Kokoro skipping jobs inappropriately


tseaver:
E.g., for PR #6202, which scribbles all over `bigquery/`, the [Kokoro - BigQuery job](https://source.cloud.google.com/results/invocations/de8b3c7d-1952-4e13-acf2-d0491c739608/log) logs (line 95)

> bigquery was not modified, returning.

And returns successfully without running any tests.

tseaver:
Hmm, looks like kokoro may have gone from skipping everything to running everything: two failures on PR #6202, which is pure BigQuery:

- [api_core](https://source.cloud.google.com/results/invocations/906ca88c-e66d-4fc7-b494-64e9f84d1224)
- [pub_sub](https://source.cloud.google.com/results/invocations/5a71c732-5462-4d80-b13e-d0b706556511)

theacodes:
Maybe rebasing helps? What happens if you manually compare master against
the PR HEAD?

PR #6239, opened 2018-10-17
Fix Kokoro package determination

theacodes:


theacodes:
Closes #6231. Do not wait for CI, it will not run anything useful.

PR #6242, opened 2018-10-17
Fix branch coverage for un-called callbacks.

tseaver:
Use `mock.Mock()` w/ `return_value` True, rather than inlining `lambda _: True` in cases where the callback doesn't get called.


tseaver:
Merging w/o review: testing only fix for broken `api_core` CI on `master`.

PR #6250, opened 2018-10-17
Adding all `Node` implementations for the `ndb.query` module

dhermes:
In particular

- `query.Node`
- `query.FalseNode`
- Most of `query.ParameterNode`. The remaining part (`ParameterNode.resolve()`) depends on as-yet-unimplemented features of `model.Property`.
- Most of `query.FilterNode`. With two "missing" parts.
  - `FilterNode._to_filter` raises `NotImplementedError` instead of creating a filter object. This is because the old implementation relies on the `google.appengine.datastore.datastore_query` module for low-level filters. At some point we'll need to replace that with the new protobuf [Filter][1]
  - If `value` is a `model.Key`, this just does what the original implementation did: call `Key.to_old_key`. This must be changed to actual convert it to a protobuf `Key` for the new proto API
- `query.PostFilterNode`. I slightly modified the `__eq__` implementation and made a note about this in `MIGRATION_NOTES.md`.
- `query.AND` and `query.OR` (named `ConjunctionFilter` and `DisjunctionFilter`); `ConjunctionFilter._to_filter()` is only partially implemented because it relies on `datastore_query` to create a low-level `Filter` protobuf

[1]: https://github.com/googleapis/googleapis/blob/ddb39a21778579122d6edf505bda0092c8066a65/google/datastore/v1/query.proto#L154

### DIFF
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -98,3 +98,8 @@ The primary differences come from:
   be dropped or redesigned if possible.
 - Since we want "compatibility", suggestions in `TODO` comments have not been
   implemented. However, that policy can be changed if desired.
+- It seems that `query.ConjunctionNode.__new__` had an unreachable line
+  that returned a `FalseNode`. This return has been changed to a
+  `RuntimeError` just it case it **is** actually reached.
+- For ``AND`` and ``OR`` to compare equal, the nodes must come in the
+  same order. So ``AND(a > 7, b > 6)`` is not equal to ``AND(b > 6, a > 7)``.
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -65,6 +65,8 @@ The primary differences come from:
   cases, the underlying representation of the class has changed (such as `Key`)
   due to newly available helper libraries or due to missing behavior from
   the legacy runtime.
+- `query.PostFilterNode.__eq__` compares `self.predicate` to `other.predicate`
+  rather than using `self.__dict__ == other.__dict__`
 
 ## Comments
 
@@ -94,3 +96,5 @@ The primary differences come from:
   import time, so this may not be an issue.
 - `ndb.model._BaseValue` for "wrapping" non-user values should probably
   be dropped or redesigned if possible.
+- Since we want "compatibility", suggestions in `TODO` comments have not been
+  implemented. However, that policy can be changed if desired.
--- a/ndb/src/google/cloud/ndb/query.py
+++ b/ndb/src/google/cloud/ndb/query.py
@@ -406,11 +406,11 @@ class FilterNode(Node):
                 return nodes[0]
             return DisjunctionNode(*nodes)
 
-        self = super(FilterNode, cls).__new__(cls)
-        self._name = name
-        self._opsymbol = opsymbol
-        self._value = value
-        return self
+        instance = super(FilterNode, cls).__new__(cls)
+        instance._name = name
+        instance._opsymbol = opsymbol
+        instance._value = value
+        return instance
 
     def __getnewargs__(self):
         """Private API used to specify ``__new__`` arguments when unpickling.
@@ -487,9 +487,9 @@ class PostFilterNode(Node):
     __slots__ = ("predicate",)
 
     def __new__(cls, predicate):
-        self = super(PostFilterNode, cls).__new__(cls)
-        self.predicate = predicate
-        return self
+        instance = super(PostFilterNode, cls).__new__(cls)
+        instance.predicate = predicate
+        return instance
 
     def __getnewargs__(self):
         """Private API used to specify ``__new__`` arguments when unpickling.
@@ -671,9 +671,9 @@ class ConjunctionNode(Node):
                 *[ConjunctionNode(*segment) for segment in clauses.or_parts]
             )
 
-        self = super(ConjunctionNode, cls).__new__(cls)
-        self._nodes = clauses.or_parts[0]
-        return self
+        instance = super(ConjunctionNode, cls).__new__(cls)
+        instance._nodes = clauses.or_parts[0]
+        return instance
 
     def __getnewargs__(self):
         """Private API used to specify ``__new__`` arguments when unpickling.
@@ -807,15 +807,15 @@ class DisjunctionNode(Node):
         elif len(nodes) == 1:
             return nodes[0]
 
-        self = super(DisjunctionNode, cls).__new__(cls)
-        self._nodes = []
+        instance = super(DisjunctionNode, cls).__new__(cls)
+        instance._nodes = []
 
         clauses = _BooleanClauses("DisjunctionNode", combine_or=True)
         for node in nodes:
             clauses.add_node(node)
 
-        self._nodes[:] = clauses.or_parts
-        return self
+        instance._nodes[:] = clauses.or_parts
+        return instance
 
     def __getnewargs__(self):
         """Private API used to specify ``__new__`` arguments when unpickling.
--- a/ndb/src/google/cloud/ndb/query.py
+++ b/ndb/src/google/cloud/ndb/query.py
@@ -236,9 +236,7 @@ class FalseNode(Node):
         """
         if post:
             return None
-        raise _exceptions.BadQueryError(
-            "Cannot convert FalseNode to predicate"
-        )
+        raise exceptions.BadQueryError("Cannot convert FalseNode to predicate")
 
 
 class ParameterNode(Node):
@@ -310,7 +308,7 @@ class ParameterNode(Node):
             .BadArgumentError: Always. This is because this node represents
             a parameter, i.e. no value exists to be filtered on.
         """
-        raise _exceptions.BadArgumentError(
+        raise exceptions.BadArgumentError(
             "Parameter :{} is not bound.".format(self._param.key)
         )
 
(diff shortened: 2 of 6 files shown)


ndb/src/google/cloud/ndb/query.py
```diff
@@ -134,28 +141,376 @@ def __init__(self, *args, **kwargs):
...
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    """Tree node for a single filter expression.
+
+    For example ``FilterNode("a", ">", 3)`` filters for entities where the
+    value ``a`` is greater than ``3``.
+
+    .. warning::
+
+        The constructor for this type may not always return a
```
dhermes: Note this warning. These "constructors" are quite devious. (I made sure to mention this in the `MIGRATION_NOTES.md`)

dhermes:
@crwilcox @chrisrossi @tseaver Can I get a review?

ndb/MIGRATION_NOTES.md
```diff
@@ -92,3 +94,10 @@ The primary differences come from:
   a new `Property()` instance is created. This increment is not threadsafe.
   However, `ndb` was designed for `Property()` instances to be created at
   import time, so this may not be an issue.
+- Since we want "compatibility", suggestions in `TODO` comments have not been
+  implemented. However, that policy can be changed if desired.
+- It seems that `query.ConjunctionNode.__new__` had an unreachable line
+  that returned a `FalseNode`. This return has been changed to a
+  `RuntimeError` just it case it **is** actually reached.
+- For ``AND`` and ``OR`` to compare equal, the nodes must come in the
+  same order. So ``AND(a > 7, b > 6)`` is not equal to ``AND(b > 6, a > 7)``.
```
chrisrossi: Are AND and OR not commutative because of logic or because of history?

dhermes: Because of history.

ndb/src/google/cloud/ndb/query.py
```diff
@@ -129,43 +140,729 @@ def resolve(self, bindings, used):
...
+            returns :data:`True` or :data:`False` if the entity matches
+            the given filter.
+    """
+
+    __slots__ = ("predicate",)
+
+    def __new__(cls, predicate):
+        self = super(PostFilterNode, cls).__new__(cls)
+        self.predicate = predicate
+        return self
```
tseaver: Is `self` conventional inside `__new__`?  I've seen `instance` elsewhere, I think.

ndb/src/google/cloud/ndb/query.py
```diff
@@ -129,43 +140,729 @@ def resolve(self, bindings, used):
...
+
+    def __init__(self, name, combine_or):
+        self.name = name
+        self.combine_or = combine_or
+        if combine_or:
+            # For ``OR()`` the parts are just nodes.
+            self.or_parts = []
+        else:
+            # For ``AND()`` the parts are "segments", i.e. node lists.
+            self.or_parts = [[]]
```
tseaver: The type flag smells odd, as thought there should really be two separate classes.

ndb/src/google/cloud/ndb/query.py
```diff
@@ -129,43 +140,729 @@ def resolve(self, bindings, used):
...
+
+    def __repr__(self):
+        all_nodes = ", ".join(map(str, self._nodes))
+        return "AND({})".format(all_nodes)
+
+    def __eq__(self, other):
+        if not isinstance(other, ConjunctionNode):
+            return NotImplemented
+
+        return self._nodes == other._nodes
```
tseaver: `__hash__`?

ndb/tests/unit/test_query.py
```diff
@@ -127,51 +130,712 @@ def test_constructor():
...
+
+    def test_resolve(self):
+        node = self._make_one()
+        used = {}
+        assert node.resolve({}, used) is node
+        assert used == {}
+
 
 class TestFalseNode:
     @staticmethod
```
tseaver: That seems an odd pattern (testcase methods static, I mean).  I guess it makes sense, given use of bare `assert` everywhere.

ndb/src/google/cloud/ndb/query.py
```diff
@@ -129,43 +140,729 @@ def resolve(self, bindings, used):
...
+
+    __slots__ = ()
+
+    def __new__(cls):
+        if cls is Node:
+            raise TypeError("Cannot instantiate Node, only a subclass.")
+        return super(Node, cls).__new__(cls)
+
+    def __eq__(self, other):
         raise NotImplementedError
```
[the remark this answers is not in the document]
dhermes: I am following the [original][1] here.

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/query.py#L380-L425

ndb/src/google/cloud/ndb/query.py
```diff
@@ -129,43 +140,729 @@ def resolve(self, bindings, used):
...
+        """
         raise NotImplementedError
 
+    def _post_filters(self):
+        """Helper to extract post-filter nodes, if any.
+
+        Returns:
+            None: Always. Because this is the base implementation.
+        """
+        return None
```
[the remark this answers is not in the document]
dhermes: I'm happy to change it, but ISTM that following the original implementation is the "best path" in some sense. I.e. the real goal of this migration should by Python 3-ifying things and adding unit tests.

But I agree with you.

ndb/src/google/cloud/ndb/query.py
dhermes: Luckily, these `Node` classes are never really instantiated in user code. However, this behavior has to be preserved (at least for now) for the migration. A node is typically created via comparison to a property:

```python
filter_ = SomeModel.total > 1000
```

ndb/src/google/cloud/ndb/query.py
dhermes: I have no relevant experience with this `__new__` constructor madness. I'm happy to rename to `instance` if you'd like.

ndb/src/google/cloud/ndb/query.py
dhermes: That's probably a good idea. The `OR` case is significantly simpler, to where it doesn't really need a class. WDYT? Should I have a `_AndClauses` and `_OrClauses` just for consistency?

ndb/src/google/cloud/ndb/query.py
dhermes: I'm not sure what the question is, could you give more details?

ndb/tests/unit/test_query.py
dhermes: This is just me being a Pylint acolyte here (i.e. it would complain about `no-self-use`). It'd be perfectly fine to have `self`, but not necessary.

ndb/MIGRATION_NOTES.md
chrisrossi: I can't imagine it would break anything to fix that, then.

ndb/src/google/cloud/ndb/query.py
tseaver: Pretty much any time I see an `__eq__` method, I look for `__hash__` as well, for immutables at least.

dhermes: Would you like me to implement it? The default will work as well?

ndb/MIGRATION_NOTES.md
dhermes: Sure thing, will do.

dhermes: OK, scratch that. It's currently not possible:

- `sorted(self._nodes)` fails because the `Node` types are **explicitly** not comparable
- `set(self._nodes)` fails because the `Node` types are not hashable (which @tseaver is commenting on below)

WDYT is the right move here?

ndb/src/google/cloud/ndb/query.py
tseaver: Maybe it is YAGNI, if application code doesn't hold on to node instances directly.  I guess we know that the library doesn't put nodes into sets or use them as dict keys?

ndb/src/google/cloud/ndb/query.py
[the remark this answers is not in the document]
tseaver: As below, let's say `__hash__` is YAGNI for now.

ndb/src/google/cloud/ndb/query.py
tseaver: Sure, it will be a kindness to our future selves.

ndb/src/google/cloud/ndb/query.py
dhermes: Correct. Given `ndb` is 7+ years old, it seems all "new" behavior is YAGNI?

ndb/src/google/cloud/ndb/query.py
dhermes: Done. I did a cursory search for `self =`, LMK if you see any remaining?

ndb/MIGRATION_NOTES.md
chrisrossi: Maybe, like __hash__, it's YAGNI.  I wonder how often these are tested for equality, anyway.  Almost never, maybe?

dhermes: Yes I'd imagine given the age of the library, it's not a feature worth adding / supporting.

ndb/src/google/cloud/ndb/query.py
dhermes: Agreed. I'd prefer to take the conservative approach here, if everyone else is on board. The goal was "full interface compatibility".

dhermes:
OK @chrisrossi @tseaver are we good to merge?

PR #6258, opened 2018-10-17
Restoring some `__new__()` constructors for `ndb`.

dhermes:
See the comment in `MIGRATION_NOTES.md` about the rampant use (and abuse) of `__new__()` rather than `__init__()` as a constructor.

Updated `MIGRATION_NOTES.md` to make a note about the difference in internal state between old and new versions of `ndb`. In particular, the new version has dropped usage of "private" (e.g. `self.__foo`) variables.

Removed `__slots__` from classes that didn't have them in the old implementation.

----

@theacodes and I discussed this and the conclusion was that "the old way" should be preserved as often as possible

### DIFF
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -58,6 +58,13 @@ The primary differences come from:
   the backend should enforce these limits, not the library.)
 - I renamed `Property.__creation_counter_global` to
   `Property._CREATION_COUNTER`.
+- `ndb` uses "private" instance attributes in many places, e.g. `Key.__app`.
+  The current implementation (for now) just uses "protected" attribute names,
+  e.g. `Key._key` (the implementation has changed in the rewrite). We may want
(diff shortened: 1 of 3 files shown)


tseaver:
I'm puzzled by removing `__slots__`:  that seems like a nice win for space / speed.  Is it important for backward-compatiblity that apps be able to scribble arbitrary attributes?  The docstrings do call those classes' instances "immutable."

dhermes:
@tseaver I was just trying to be "compatible" with the [old implementation][1], which has lots of warts.

I actually think it's the "right" idea to use `__slots__` on essentially every class since I'm remapping the "private" variables (e.g. `self.__foo` becomes `self._foo`).

@theacodes Where does this fall on our discussion about preserving the old behavior? (The original discussion was just that I should use `__new__` whenever the original impl. did.)

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python

PR #6264, opened 2018-10-18
Docs: fix 'CallOptions' examples

### LINKED ISSUE #5560: Monitoring: where is CallOptions on monitoring API example?
[OS] macOS Sierra 10.12.6
[Versions]

- Python 3.6.1

```
google-api-core==1.2.1
google-api-python-client==1.7.3
google-auth==1.5.0
google-auth-httplib2==0.0.3
google-cloud-monitoring==0.30.0
```

----

## CallOptions class was not found!

Hi. I'm new to GCP and Stackdriver. I wanted to use Google Kubernetes Engine and its auto scaling by custom metrics. Then, it is required to export the metrics to Stackdriver Monitoring, so I am trying to do it.

But, After installing above-mentioned libraries, the example code on monitoring API README document failed. The pit hole is that `CallOptions` was not found, thus I've searched it in this repository and some other repositories.

And finally, I couldn't find it...

`CallOptions` is defined in gax.python, but the package is currently deprecated and moved to google-api-core. So I guess that also the dependency is currently corrupted or the some examples are out-of-date.

Please tell me how handle this problem.

_Thank you for the great package and platform._

tseaver:
In package `README.rst` files, where `CallOptions` was being shown as a way to trigger paged iteration, usse idiomatic `pages` attribute of iterators

For `VisionHelper.annotate_image`, update docstring to reflect the actual signature, which takes `retry` and `timeout` rather than `options`.

Closes #5560.

Other elements of that bug will be fixed in [`gapic-generator`](googleapis/gapic-generator#2388) and propagated via autosynth.

### DIFF
--- a/bigquery_datatransfer/README.rst
+++ b/bigquery_datatransfer/README.rst
@@ -88,7 +88,7 @@ DataTransferServiceClient
         pass
 
     # Or iterate over results one page at a time
-    for page in client.list_data_sources(parent, options=CallOptions(page_token=INITIAL_PAGE)):
+    for page in client.list_data_sources(parent).pages:
         for element in page:
             # process element
             pass
--- a/dataproc/README.rst
+++ b/dataproc/README.rst
@@ -87,7 +87,7 @@ Example Usage
         pass
 
     # Or iterate over results one page at a time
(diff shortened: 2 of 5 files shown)


tseaver:
Merging without review:  this is a docs-only fix, where the current samples / docstrings are clearly borked.

PR #6267, opened 2018-10-18
1.6.0 alpha

lukesneeringer:
This adds two methods to `google-api-core` that will be used by the new GAPICs.
It also bumps the version number to `1.6.0a1`, so that we can release this to PyPI and iterate without affecting existing users.

### DIFF
--- a/api_core/google/api_core/gapic_v1/dispatch.py
+++ b/api_core/google/api_core/gapic_v1/dispatch.py
@@ -1,0 +1,32 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+
+
+def dispatch(func):
+    base_dispatcher = functools.singledispatch(func)
+
+    # Define a wrapper function that works off args[1] instead of args[0].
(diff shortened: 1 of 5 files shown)


api_core/google/api_core/gapic_v1/dispatch.py
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+
+
+def dispatch(func):
```
theacodes: Docstring?

api_core/setup.py
```diff
@@ -22,12 +22,12 @@
 
 name = 'google-api-core'
 description = 'Google API client core library'
-version = '1.5.0'
+version = '1.6.0a1'
```
theacodes: please don't modify setup.py in this PR. Releasetool will do it.

lukesneeringer: Ack, will revert.

api_core/google/api_core/gapic_v1/dispatch.py
lukesneeringer: Added.

api_core/setup.py
```diff
@@ -27,7 +27,7 @@
 # 'Development Status :: 3 - Alpha'
 # 'Development Status :: 4 - Beta'
 # 'Development Status :: 5 - Production/Stable'
-release_status = 'Development Status :: 5 - Production/Stable'
+release_status = 'Development Status :: 3 - Alpha'
```
tseaver: Ewwww, revert this bit please.

theacodes:
Let's still revert the chAnge here.

lukesneeringer:
Fair enough!

lukesneeringer:
Distilling a Hangouts discussion: We are moving the dispatch item to a `gapic_v2` directory to account for its being Python 3-only.

PR #6274, opened 2018-10-18
Adding `__slots__` throughout `ndb`.

dhermes:
This is a spin-off of the discussion in #6258 with @theacodes and @tseaver.

I have added `__slots__` to **all** classes except for `Property`. The reason I didn't with `Property` is because the implementation actual depends on `'foo' in self.__dict__` (AFAICT just for `repr()`) to determine if a given attribute is the default (provided by the class) or set explicitly on the instance.

To see why `__slots__` don't work when a class uses a fallback:

```python
>>> class A:
...     __slots__ = ("a", "b")
...     a = 10
...     b = 11
...     def __init__(self, *, a=None, b=None):
...         if a is not None:
...             self.a = a
...         if b is not None:
...             self.b = b
...
ValueError: 'a' in __slots__ conflicts with class variable
```

### DIFF
--- a/ndb/src/google/cloud/ndb/blobstore.py
+++ b/ndb/src/google/cloud/ndb/blobstore.py
@@ -68,6 +68,8 @@ class BlobFetchSizeTooLargeError:
 
 
 class BlobInfo:
+    __slots__ = ()
+
     def __init__(self, *args, **kwargs):
         raise NotImplementedError
 
@@ -94,6 +96,8 @@ class BlobInfoParseError:
 
 
 class BlobKey:
+    __slots__ = ()
+
     def __init__(self, *args, **kwargs):
         raise NotImplementedError
 
@@ -107,6 +111,8 @@ class BlobNotFoundError:
 
 
 class BlobReader:
+    __slots__ = ()
+
     def __init__(self, *args, **kwargs):
         raise NotImplementedError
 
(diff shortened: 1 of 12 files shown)


ndb/src/google/cloud/ndb/blobstore.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, *args, **kwargs):
 
 
 class BlobInfo:
+    __slots__ = ()
+
```
tseaver: Weird -- are all these empty-slots, only-stub-methods classes placeholders for later work?

dhermes: Indeed. I'm happy to remove them if you like. I just wanted to remind my future self, e.g. with a failing test, if they were absent. 

ASIDE: In cases where I used `__slots__`, I've typically [implemented `__dict__`][1] for users that want introspection from the console. WDYT?

[1]: https://github.com/dhermes/bezier/blob/f99db20312bb4ba7e5943195020a8ced4be9457b/src/bezier/curve.py#L147-L148

tseaver: @dhermes Don't remove them:  I was just confused that the PR was touching only things which can't (yet) be used.

Re a property for `__dict__`:  shouldn't `help(thing)` do the right thing without it?  Exposing it but not allowing people to munge its contents still breaks expectations.

dhermes: > Exposing it but not allowing people to munge its contents still breaks expectations.

Yes I agree.

dhermes:
@tseaver Is this OK to merge?

PR #6275, opened 2018-10-18
Implementing more of `ndb` Property

dhermes:
~~**NOTE**: This has #6250 as diffbase (since `Property.__eq__` relies on `query.Node` implementations)~~

- `Property.__repr__`
- `Property._comparison`, which provides the nifty ORM feature that makes `MyModel.foo == 10` spit out a filter
- `Property.__eq__` and friends (all just call `_comparison`)
- `Property.IN`
- `+Property` and `-Property` (this is just a stub, but they **should** create `Order` objects)
- `Property._datastore_type` (used to convert from Python types to protobuf types when needed)
- Completes the implementation of `query.ParameterNode.resolve()`

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -567,6 +567,34 @@ class Property(ModelAttribute):
             value = self._datastore_type(value)
         return query.FilterNode(self._name, op, value)
 
+    # Comparison operators on Property instances don't compare the
+    # properties; instead they return ``FilterNode``` instances that can be
+    # used in queries.
+
+    def __eq__(self, value):
+        """FilterNode: Represents the ``=`` comparison."""
+        return self._comparison("=", value)
+
+    def __ne__(self, value):
+        """FilterNode: Represents the ``!=`` comparison."""
+        return self._comparison("!=", value)
+
+    def __lt__(self, value):
+        """FilterNode: Represents the ``<`` comparison."""
+        return self._comparison("<", value)
+
+    def __le__(self, value):
+        """FilterNode: Represents the ``<=`` comparison."""
+        return self._comparison("<=", value)
+
+    def __gt__(self, value):
+        """FilterNode: Represents the ``>`` comparison."""
+        return self._comparison(">", value)
+
+    def __ge__(self, value):
+        """FilterNode: Represents the ``>=`` comparison."""
+        return self._comparison(">=", value)
+
 
 class ModelKey(Property):
     __slots__ = ()
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -556,15 +556,17 @@ class Property(ModelAttribute):
         Raises:
             BadFilterError: If the current property is not indexed.
         """
+        # Import late to avoid circular imports.
+        from google.cloud.ndb import query
+
         if not self._indexed:
             raise exceptions.BadFilterError(
                 "Cannot query for unindexed property {}".format(self._name)
             )
-        # Import late to avoid circular imports.
-        from google.cloud.ndb import query
 
         if value is not None:
             value = self._datastore_type(value)
+
         return query.FilterNode(self._name, op, value)
 
     # Comparison operators on Property instances don't compare the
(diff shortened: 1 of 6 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -439,6 +451,204 @@ def _verify_validator(validator):
 
         return validator
 
+    def __repr__(self):
+        """Return a compact unambiguous string representation of a property.
+
+        This cycles through all stored attributes and displays the ones that
+        differ from the default values.
+        """
+        args = []
```
dhermes: I am not a big fan of this implementation and I think it can be ditched in favor of using `inspect.signature`. This will slow things down, but I'd imagine there are approximately 0 users depending on `repr(MyModel.foo)` (i.e. of a property) to be fast.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -439,6 +451,204 @@ def _verify_validator(validator):
...
+        # Import late to avoid circular imports.
+        from google.cloud.ndb import query
+
+        if not self._indexed:
+            raise exceptions.BadFilterError(
+                "Cannot query for unindexed property {}".format(self._name)
+            )
+
+        if value is not None:
+            value = self._datastore_type(value)
```
dhermes: There are actually three checks / conversions done here but I left out the other two to make this PR easier to review.

ndb/src/google/cloud/ndb/model.py
tseaver: +1. It seems safe to say that the set of users depending on `repr` for anything other than casual use at the REPL prompt is empty, and performance at the prompt can't matter unless it is absurdly bad.

ndb/src/google/cloud/ndb/model.py
dhermes: They get added in #6295

ndb/src/google/cloud/ndb/model.py
dhermes: @tseaver PTAL, I switched to the `inspect.signature()` approach and didn't really need to modify the unit tests (other than the `foo` -> `_foo` mangling)

dhermes:
@tseaver @chrisrossi @crwilcox PTAL?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -487,6 +489,202 @@ def _verify_validator(validator):
...
+        cls = self.__class__
+        signature = inspect.signature(self.__init__)
+        for name, parameter in signature.parameters.items():
+            attr = "_{}".format(name)
+            instance_val = getattr(self, attr)
+            default_val = getattr(cls, attr)
+
+            if instance_val is not default_val:
+                if isinstance(instance_val, type):
+                    as_str = instance_val.__name__
```
tseaver: Maybe `__qualname__` would be better for types?

ndb/src/google/cloud/ndb/query.py
```diff
@@ -339,14 +339,17 @@ def resolve(self, bindings, used):
...
                 parameters.
 
-        Raises:
-            NotImplementedError: Always. This is because the implementation
-                will rely on as-yet-unimplemented features in
-                :class:`~google.cloud.ndb.model.Property`.
+        Returns:
+            Union[~google.cloud.ndb.query.DisjunctionNode, \
+                ~google.cloud.ndb.query.FilterNode, \
+                ~google.cloud.ndb.query.FalseNode]: A node corresponding to
```
tseaver: How will the implementation below return `FalseNode`?  I don't see a test below that triggers one.

ndb/src/google/cloud/ndb/model.py
dhermes: 1. I have no opposition to that change
1. Do you think that'd be a "backwards incompatible" change?

ndb/src/google/cloud/ndb/query.py
dhermes: @tseaver This matches the union of the return types of `Property._IN` and `Property._comparison`. A `FalseNode` will be created when used with an empty container, e.g. `MyModel.foo_prop.IN([])`.

tseaver: OK, then maybe add a test which does that?

ndb/src/google/cloud/ndb/model.py
tseaver: The only thing I can think we'd be breaking might be doctests which use the `__repr__` result in place of assertions.

FWIW, PEP 3155 states, 

> For top-level functions and classes, the `__qualname__` attribute is equal to the `__name__` attribute

which means the only visible difference would be for nested classes.  How likely is it that class derived from `Property` has a `type` in its argument list, *and* that it is nested?

ndb/tests/unit/test_query.py
```diff
@@ -291,15 +291,36 @@ def test__to_filter():
...
 
+        value = 67
+        bindings = {"abc": value}
         used = {}
-        with pytest.raises(NotImplementedError):
-            parameter_node.resolve({}, used)
-        assert used == {}
+        resolved_node = parameter_node.resolve(bindings, used)
+
+        assert resolved_node == query.FilterNode(b"val", "=", value)
```
chrisrossi: The property name gets converted to bytes?

dhermes: For now yes, but that was a Python 3 vs. Python 2 weirdness. I think at some point we'll need to dive in and figure out the differences but for now I just "copied" the old behavior.

ndb/src/google/cloud/ndb/query.py
dhermes: Done

ndb/src/google/cloud/ndb/model.py
dhermes: I added `__qualname__`

PR #6277, opened 2018-10-19
Add --params option to %%bigquery magic

guillermo-carrasco:
Often it is useful to be able to parametrize query strings. While there is the option of parametrizing queries [using the raw sdk](https://cloud.google.com/bigquery/docs/parameterized-queries), there is no option to do so using the `%%bigquery` magic in Jupyter notebooks. 

This is useful for several reasons, specially when the same parameter appears several times in the query. This could be for example a cutoff date appearing in several `JOIN` conditions.

This PR adds the option `--params` to the `%%bigquery` magic for Jupyter notebooks. `--params` accepts a JSON string that will be used to format the string contained in the query. 

An example of how this would work:
```
%%bigquery df --params {"max_question_length":300,"limit":10}

SELECT
      posts.id AS post_id,
      posts.creation_date AS post_creation_date,
      posts.body AS question
FROM
  `bigquery-public-data.stackoverflow.posts_questions` posts
WHERE
  LENGTH(posts.body) < {max_question_length}
LIMIT {limit}
```

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -40,8 +40,9 @@
         query is finished. By default, this information will be displayed but
         will be cleared after the query is finished.
     * ``--params <params dictionary>`` (optional, line argument):
-        If present, the argument must be a parsable JSON string. This dictionary
-        will be used to format values preceded by ``@`` in the query.
+        If present, the argument must be a parsable JSON string or a reference to dictionary
+        which is serializable to JSON (preceding the dictionary with $).
+        This dictionary will be used to format values preceded by ``@`` in the query.
     * ``<query>`` (required, cell argument):
         SQL query to run.
 
@@ -102,13 +103,6 @@
         In [5]: %%bigquery df --params {"num": 17}
            ...: SELECT @num AS num
         Out[5]:
-           ...:    num
-           ...: 0   17
-        In [6]: # Expand a dictionary instead of writing it's string value
-        In [6]: params = {"num": 17}
-        In [7]: %%bigquery df --params $params
-           ...: SELECT @num AS num
-        Out[7]:
            ...:    num
            ...: 0   17
 """
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -40,7 +40,7 @@
         query is finished. By default, this information will be displayed but
         will be cleared after the query is finished.
     * ``--params <params dictionary>`` (optional, line argument):
-        If present, the argument will be parsed into a dictionary. This dictionary
+        If present, the argument must be a parsable JSON string. This dictionary
         will be used to format values enclosed within {} in the query.
     * ``<query>`` (required, cell argument):
         SQL query to run.
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -103,10 +103,9 @@
 
 from __future__ import print_function
 
-import json
+import ast
 import time
 from concurrent import futures
-from json import JSONDecodeError
 
 try:
     import IPython
@@ -256,6 +255,7 @@ def _run_query(client, query, job_config=None):
           'cleared after the query is finished.'))
 @magic_arguments.argument(
     '--params',
+    nargs='+',
     default=None,
     help=('Parameters to format the query string. If present, it should be a '
           'parsable JSON string. The parsed dictionary will be used for string'
@@ -278,9 +278,9 @@ def _cell_magic(line, query):
 
     if args.params is not None:
         try:
-            params = json.loads(args.params)
-        except JSONDecodeError as e:
-            raise JSONDecodeError('--params is not a correctly formatted JSON string', e.doc, e.pos)
+            params = ast.literal_eval(''.join(args.params))
+        except Exception:
+            raise SyntaxError('--params is not a correctly formatted JSON string')
 
         query = query.format(**params)
 
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -41,7 +41,7 @@
         will be cleared after the query is finished.
     * ``--params <params dictionary>`` (optional, line argument):
         If present, the argument must be a parsable JSON string. This dictionary
-        will be used to format values enclosed within {} in the query.
+        will be used to format values preceded by ``@`` in the query.
     * ``<query>`` (required, cell argument):
         SQL query to run.
 
@@ -99,6 +99,18 @@
            ...: 1    Patricia  1568495
            ...: 2   Elizabeth  1519946
 
+        In [5]: %%bigquery df --params {"num": 17}
+           ...: SELECT @num AS num
+        Out[5]:
+           ...:    num
+           ...: 0   17
+        In [6]: # Expand a dictionary instead of writing it's string value
+        In [6]: params = {"num": 17}
+        In [7]: %%bigquery df --params $params
+           ...: SELECT @num AS num
+        Out[7]:
+           ...:    num
+           ...: 0   17
 """
 
 from __future__ import print_function
@@ -116,7 +128,7 @@ except ImportError:  # pragma: NO COVER
 
 import google.auth
 from google.cloud import bigquery
-from google.cloud.bigquery.dbapi._helpers import to_query_parameters
+from google.cloud.bigquery.dbapi import _helpers
 
 
 class Context(object):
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/magics.py
```diff
@@ -39,6 +39,9 @@
         amount of time for the query to complete will not be cleared after the
         query is finished. By default, this information will be displayed but
         will be cleared after the query is finished.
+    * ``--params <params dictionary>`` (optional, line argument):
+        If present, the argument will be parsed into a dictionary. This dictionary
+        will be used to format values enclosed within {} in the query.
```
tseaver: The docs here need to spell out the format of the value (e.g., as done below, "a parseable JSON string").

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -265,6 +276,14 @@ def _cell_magic(line, query):
     """
     args = magic_arguments.parse_argstring(_cell_magic, line)
 
+    if args.params is not None:
+        try:
+            params = json.loads(args.params)
+        except JSONDecodeError as e:
+            raise JSONDecodeError('--params is not a correctly formatted JSON string', e.doc, e.pos)
```
tseaver: Can you clarify the utility of replacing the original exception here?

bigquery/tests/unit/test_magics.py
```diff
@@ -259,3 +259,31 @@ def test_bigquery_magic_with_project():
...
+    query_job_mock.to_dataframe.return_value = result
+    with run_query_patch as run_query_mock:
+        run_query_mock.return_value = query_job_mock
+
+        ip.run_cell_magic('bigquery', 'df --params {"num":17}', sql)
+
+    assert 'df' in ip.user_ns        # verify that variable exists
+    df = ip.user_ns['df']
+    assert len(df) == len(result)    # verify row count
+    assert list(df) == list(result)  # verify column names
```
tseaver: Needs an assertion that the expected value was passed to `_run_query`.  E.g.:

```python
        run_query_patch.assert_called_once_with(sql.format(num=17))
```

bigquery/google/cloud/bigquery/magics.py
guillermo-carrasco: The reason for replacing is that the original exception would be something like this:

```
<stack trace>
JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 28 (char 27)
```

In order to know where the error is, you have to follow the stacktrace. With this message, it makes it clearer where the error was.

guillermo-carrasco:
@tseaver I am not familiar with the review process in these repositories. Should I push a new commit with the suggested changes, and rebase at the end? or should I rebase now? 

Thanks for reviewing!

tseaver:
@guillermo-carrasco 

> I am not familiar with the review process in these repositories. Should I push a new commit with the suggested changes, and rebase at the end? or should I rebase now?

Push a new commit on the same branch.  I will review those changes.  We iterate until I hit the "approve" button for the review, and then I hit the merge button.

Occasionally I might ask for a rebase against the current `master` (e.g., if the CI machinery gets confused about what has changed), but I wouldn't ask that you squash commits in such a rebase:  instead, when I merge, I will squash all commits for the PR into one.

> Thanks for reviewing!

Thank you for the patch!

guillermo-carrasco:
@tseaver thanks for the clarification. I just amended the suggested changes.

tseaver:
@shollyman, @tswast I'm fine with the patch as it stands now.  Please comment on suitability from the DPE side.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -265,6 +276,14 @@ def _cell_magic(line, query):
...
+
+        query = query.format(**params)
```
tswast: I find this surprising behavior. I would expect this to use BigQuery's built-in query parameter handling. The DB-API implementation may be of good use here. See:

https://github.com/googleapis/google-cloud-python/blob/448e0ec97db246f3f42848ec8e92b946ed51ff4e/bigquery/google/cloud/bigquery/dbapi/cursor.py#L320

https://github.com/googleapis/google-cloud-python/blob/448e0ec97db246f3f42848ec8e92b946ed51ff4e/bigquery/google/cloud/bigquery/dbapi/_helpers.py#L96

and

https://github.com/googleapis/google-cloud-python/blob/448e0ec97db246f3f42848ec8e92b946ed51ff4e/bigquery/google/cloud/bigquery/dbapi/cursor.py#L148

Since DB-API format option is not required, I recommend skipping the `_format_operation` step.

tseaver: @tswast This reply was actually to your comment (easy to do with the Github UI being "helpful").

guillermo-carrasco:
This new commit makes it easier to use dictionaries with the `%%bigquery` magic. Now one can pass in a previously built dictionary, instead of having to write it down in the magic line:

```
In [1]: params = {"min_word_count": 200, "corpus": "romeoandjuliet"}

In [2]: %%bigquery df --params $params
    ...: 
    ...: SELECT word, word_count
    ...: FROM `bigquery-public-data.samples.shakespeare`
    ...: WHERE corpus = "{corpus}"
    ...: AND word_count >= {min_word_count}
    ...: ORDER BY word_count DESC;
    ...: 
```

bigquery/google/cloud/bigquery/magics.py
tswast: The method

https://github.com/googleapis/google-cloud-python/blob/448e0ec97db246f3f42848ec8e92b946ed51ff4e/bigquery/google/cloud/bigquery/dbapi/_helpers.py#L96

does exactly what you want. It converts strings to the proper parameter objects.

guillermo-carrasco:
@tswast please disregard my last comment. I have folowed your suggestion and used `to_query_parameters` from BigQuery's built-in query parameter handling. Now the parametrizing in notebooks works in line with the SDK:

```
In [1]: params = {"min_word_count": 200, "corpus": "romeoandjuliet"}

In [2]: %%bigquery df --params $params
   ...: 
   ...: SELECT word, word_count
   ...: FROM `bigquery-public-data.samples.shakespeare`
   ...: WHERE corpus = @corpus
   ...: AND word_count >= @min_word_count
   ...: ORDER BY word_count DESC;

```

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -112,6 +116,7 @@
 
 import google.auth
 from google.cloud import bigquery
+from google.cloud.bigquery.dbapi._helpers import to_query_parameters
```
alixhami: Please import the module rather than the function

guillermo-carrasco:
Sorry for closing, it was a mistake. Thanks for the review @alixhami. I'll implement the suggested changes.

guillermo-carrasco:
@alixhami I have implemented the following changes:

* Added a test for expanding a dictionary
* Importing module instead of function
* Fixed `df` typo (also in the other test)

I am not sure that changing the docstring to reflect the dict approach is what we want though. It is still a JSON string what is sent in the cell magic. It's just that it is also possible to expand a dictionary _instead of_ writing the JSON string. But at the end, it is a JSON string what is treated in `_cell_magic`. Do you still think that I should change the documentation? I have added usage examples for both cases.

Regarding contributions: I totally understand, and I appreciate you're taking the time to review this. Query parametrization is something we use a lot, and we have a heavy notebook-drived development, so this would be a great addition for us. 

Thanks again!

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -39,6 +39,9 @@
         amount of time for the query to complete will not be cleared after the
         query is finished. By default, this information will be displayed but
         will be cleared after the query is finished.
+    * ``--params <params dictionary>`` (optional, line argument):
+        If present, the argument must be a parsable JSON string. This dictionary
```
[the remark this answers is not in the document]
tswast: How about "... must be a JSON string or a reference to dictionary which is serializable to JSON." ?

It's too bad that it has to go to a JSON string, since that means we can't pass in DATETIME or NUMERIC query parameters, but that's probably okay, given that we want the magics to be simple rather than complete.

alixhami:
Thanks @guillermo-carrasco!

@tswast - what is your preference on how this parameter should be documented?

guillermo-carrasco:
I've slightly changed the docstring now. I had to remove the dictionary reference example from the documentation. The reason is that since `$` is ipython specific syntax, sphinx is failing on building the documentation. I couldn't find a way to overcome this. I'm open to suggestions, thanks!

alixhami:
@guillermo-carrasco I'm looking into the docs build issue now. I think that example is very helpful, so I'm going to see if there's a way to include it without throwing errors.

alixhami:
Ok I pushed some commits, which make the following updates:
+ Remove python syntax highlighting (because it isn't actually python anyway), so the dict params example could be added back in
+ Fix lint and coverage
+ Update the documentation for the `params` parameter to be more detailed and point to the relevant examples. The trade off I made here is favoring clarity at the expense of being repetitive. I ran the docstring updates by @tswast offline.

guillermo-carrasco:
amazing, thanks a lot for all the help and reviews :)

Issue #6278, opened 2018-10-19
Bigtable: plumb 'transport' through from 'Client' to GAPIC generated clients


tseaver:
/cc @sduskis 

Two issues:

- `bigtable.client.Client.__init__` takes a `channel` argument, but [does nothing with it](https://github.com/googleapis/google-cloud-python/blob/910846426428f24ad49e1a0f716e5bf103b3e3af/bigtable/google/cloud/bigtable/client.py#L102-L114):  the `_channel` attribute is never used.

- The current GAPIC-generated clients still accept a `channel` argument, but are deprecating it in favor of the `transport` argument, where the `transport` is a generated wrapper around the `channel`, providing the method stubs needed to support the API surface.

We should deprecate the `channel` argument to `bigtable.client.Client.__init__`, and add a new `transport` argument.  If passed, the `transport` argument should be forwarded through to the three GAPIC-generated clients.  If `channel` is passed, we should forward it through as well, until the GAPIC-generated code no longer supports it.

tseaver:
"That no work!"  The `transport` argument to a GAPIC generated client is specific to the methods:  we would need to expose `instance_admin_transport`, `table_admin_transport`, and `table_data_transport` arguments.

tseaver:
Fundamentally, we cannot *both* compute credentials eagerly *and* accept a `channel` argument:  they are incompatible.  We would need to figure out how to suppress computing the credentials in the base client class if `channel` were passed.

tseaver:
I guess the other choice would be that, if the user passed `channel`, we just discarded the base-class-constructed `credentials`, and pass *only* the `channel` to the GAPIC-client constructors.

sangramql:
https://github.com/googleapis/google-cloud-python/issues/6278#issuecomment-431453238
@sduskis @AVaksman @tseaver ,
Do i need to expose the instance_admin_transport, table_admin_transport, and table_data_transport arguments?

sangramql:
Comment in PR #7948 by @sduskis : The issue should be rephrased to remove channel and add new data_transport, table_admin_transport and instance_admin_transport

sangramql:
**Summary for future reference:**

Transport parameter needs to be added and Channel needs to be deprecated.

Transport is suppose to be wrapper to channel. 
Transport will be optional parameter, if not passed default Transport will be created internally.


_Below are 2 approaches and their significance:_
1. Channel can be passed by user inside a transport, if not passed default channel will be used.
[...]

PR #6280, opened 2018-10-19
Bigtable: plumb 'Client._channel' to GAPIC clients

tseaver:
See: #6278 

Alternative strategy to the deprecation of `channel` in #6279.


sduskis:
We'd need a "data channel" and an "admin channel" for this to work.  we can't have a single channel for both endpoints.

tseaver:
In favor of #6279.

PR #6283, opened 2018-10-22
Docs: remove umbrella-package vestige from docs build.

tseaver:


tseaver:
Merging without review:  the fossil is clearly not needed, and the [docs build passes without it](https://circleci.com/gh/googleapis/google-cloud-python/8857).

PR #6286, opened 2018-10-22
Implementing `model._BaseValue` for `ndb`.

dhermes:
This is used to wrap values returned from the datastore so that they can be differentiated by values set by the user.

Worth noting: It's unclear if this would be needed by the "best" implementation, but I think it's "best" to leave it in for now.

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -285,6 +285,50 @@ class ModelAttribute:
         """
 
 
+class _BaseValue:
+    """A marker object wrapping a "base type" value.
+
+    This is used to be able to tell whether ``entity._values[name]`` is a
+    user value (i.e. of a type that the Python code understands) or a
+    base value (i.e of a type that serialization understands).
+    User values are unwrapped; base values are wrapped in a
+    :class:`_BaseValue` instance.
+
+    Args:
+        b_val (Any): The base value to be wrapped.
+
+    Raises:
+        TypeError: If ``b_val`` is :data:`None`.
+        TypeError: If ``b_val`` is a list.
+    """
+
+    __slots__ = ("b_val",)
+
+    def __init__(self, b_val):
+        if b_val is None:
+            raise TypeError("Cannot wrap None")
+        if isinstance(b_val, list):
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -285,6 +285,50 @@ def _fix_up(self, cls, code_name):
         """
 
 
+class _BaseValue:
```
tseaver: Is `ndb` Python 3.x only?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -285,6 +285,50 @@ def _fix_up(self, cls, code_name):
...
+    """A marker object wrapping a "base type" value.
+
+    This is used to be able to tell whether ``entity._values[name]`` is a
+    user value (i.e. of a type that the Python code understands) or a
+    base value (i.e of a type that serialization understands).
+    User values are unwrapped; base values are wrapped in a
+    :class:`_BaseValue` instance.
```
tseaver: Why would we be mixing the two kinds of values in the same mapping?  Seems like a recipe for unending pain to me.

ndb/src/google/cloud/ndb/model.py
dhermes: Yes

ndb/src/google/cloud/ndb/model.py
dhermes: I agree with you, but there is a **lot** of complexity in this library. I'm trying to get a full picture before I pull this out.

tseaver: OK.  Is there a note to that effect in your migration docs, or should we create an issue for it?

dhermes: A note that I'm leaving in questionable choices from the original implementation? Or just specific to this type?

tseaver: Maybe we need a list of the oddballs to review later.

dhermes: OK. Sent out https://github.com/googleapis/google-cloud-python/pull/6289

PR #6295, opened 2018-10-23
Continuation of `ndb` Property implementation

dhermes:
~~**NOTE**: This has #6275 as diffbase.~~

- Correct validation / conversion of `value` in `Property._comparison` and `Property._IN` (via `_do_validate` and `_call_to_base_type`)
- `Property._do_validate`
- `Property._call_to_base_type`
- `Property._call_shallow_validation`
- `Property._find_methods`
- `Property._apply_list`

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -348,7 +348,7 @@ class Property(ModelAttribute):
     _write_empty_list = False
     # Non-public class attributes.
     _CREATION_COUNTER = 0
-    _find_methods_cache = {}
+    _FIND_METHODS_CACHE = {}
 
     def __init__(
         self,
@@ -865,7 +865,7 @@ class Property(ModelAttribute):
         """
         # Get cache on current class / set cache if it doesn't exist.
         key = "{}.{}".format(cls.__module__, cls.__qualname__)
-        cache = cls._find_methods_cache.setdefault(key, {})
+        cache = cls._FIND_METHODS_CACHE.setdefault(key, {})
         hit = cache.get(names)
         if hit is not None:
             if reverse:
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -685,6 +685,62 @@ class Property(ModelAttribute):
         """
         raise NotImplementedError("Missing datastore_query.PropertyOrder")
 
+    def _do_validate(self, value):
+        """Call all validations on the value.
+
+        This transforms the ``value`` via:
+
+        * Calling the derived ``_validate()`` method(s) (on subclasses that
+          don't define ``_to_base_type``),
+        * Calling the custom validator function
+
+        After transforming, it checks if the transformed value is in
+        ``choices`` (if defined).
+
+        It's possible that one of the ``_validate()`` methods will raise
+        an exception.
+
+        If ``value`` is a base-value, this will do nothing and return it.
+
+        .. note::
+
+            This does not call all composable ``_validate()`` methods.
+            It only calls ``_validate()`` methods up to the
+            first class in the hierarchy that defines a ``_to_base_type()``
+            method, when the MRO is traversed looking for ``_validate()`` and
+            ``_to_base_type()`` methods.
+
+        .. note::
+
+            For a repeated property this method should be called
+            for each value in the list, not for the list as a whole.
+
+        Args:
+            value (Any): The value to be converted / validated.
+
+        Returns:
+            Any: The transformed ``value``, possibly modified in an idempotent
+            way.
+        """
+        if isinstance(value, _BaseValue):
+            return value
+
+        value = self._call_shallow_validation(value)
+
+        if self._validator is not None:
+            new_value = self._validator(self, value)
+            if new_value is not None:
+                value = new_value
+
+        if self._choices is not None:
+            if value not in self._choices:
+                raise exceptions.BadValueError(
+                    "Value {!r} for property {} is not an allowed "
+                    "choice".format(value, self._name)
+                )
+
+        return value
+
     def _call_to_base_type(self, value):
         """Call all ``_validate()`` and ``_to_base_type()`` methods on value.
 
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -867,7 +867,10 @@ class Property(ModelAttribute):
         if cache:
             hit = cache.get(names)
             if hit is not None:
-                return hit
+                if reverse:
+                    return list(reversed(hit))
+                else:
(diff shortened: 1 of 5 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -697,6 +701,236 @@ def __pos__(self):
...
+            names (Tuple[str, ...]): One or more method names to look up on
+                the current class or base classes.
+            reverse (bool): Optional flag, default False; if True, the list is
+              reversed.
+
+        Returns:
+            List[Callable]: Class method objects.
+        """
+        # Get cache on current class / set cache if it doesn't exist.
+        cache = cls.__dict__.get("_find_methods_cache")
```
dhermes: I **really** hate what's going on here. This is a mutable cache that is specific to each `Property` subclass (of which there are many). The only way to emulate this behavior that is "obvious" is a metaclass.

It's really no fun because every time a unit test calls a method that invokes `_find_methods`, it has side effects for all other tests (see the pytest fixture I added to account for this).

ndb/src/google/cloud/ndb/model.py
```diff
@@ -697,6 +701,236 @@ def __pos__(self):
...
+        if cache:
+            hit = cache.get(names)
+            if hit is not None:
+                return hit
```
dhermes: Note that this doesn't take `reverse` into consideration, so the original implementation had a bug here. I have (for now) preserved this bug.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -697,6 +701,236 @@ def __pos__(self):
...
+        Args:
+            methods (Iterable[Callable[[Any], Any]]): An iterable of methods
+                to apply to a value.
+
+        Returns:
+            Callable[[Any], Any]: A callable that takes a single value and
+            applies each method in ``methods`` to it.
+        """
+
+        def call(value):
```
dhermes: I am following the original implementation here, but would prefer a class (with `__call__`) encapsulating what is going on here. Though here I realize not all Pythonistas agree with me.

ndb/src/google/cloud/ndb/query.py
```diff
@@ -154,7 +154,7 @@ class Node:
     """Base class for filter expression tree nodes.
 
     Tree nodes are considered immutable, even though they can contain
-    Parameter instances, which are not.  In particular, two identical
+    Parameter instances, which are not. In particular, two identical
```
dhermes: It seems the original prose author uses two spaces after punctuation. I am being pedantic / wrong I suppose.

dhermes:
@tseaver @chrisrossi @crwilcox PTAL? Note that this has #6275 as diffbase.

ndb/src/google/cloud/ndb/model.py
tseaver: Does that require a comment in the code here, or is it noted elsewhere?

ndb/src/google/cloud/ndb/model.py
tseaver: If the class would be clearer, or more testable, then go for it (I assume the tests for `_apply_list` would pass either way.

ndb/tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+def property_clean_cache():
+    """Reset the ``_find_methods_cache`` class attribute on ``Property``
+
+    This property is set at runtime (with calls to ``_find_methods()``), so
+    this fixture allows resetting the class to its original state.
+    """
+    try:
+        yield
+    finally:
+        del model.Property._find_methods_cache
```
tseaver: Don't we need to be cleaning out the cache for all subclasses of `model.Property` as well?

ndb/tests/unit/test_model.py
```diff
@@ -584,6 +585,221 @@ def test___pos__():
         with pytest.raises(NotImplementedError):
             +prop
 
+    @staticmethod
+    def test__do_validate(property_clean_cache):
```
tseaver: How does that fixture get to be available in this module?  Is `conftest.py` a majyk thing in pytest-land?

ndb/tests/conftest.py
dhermes: @tseaver Yes. I figured we'd cross that bridge when we get there (i.e. as I implement them / add tests for them)

ndb/tests/unit/test_model.py
dhermes: Yes `conftest.py` is a standard part of `pytest`. FWIW `pytest` is widely used in the Python community.

ndb/src/google/cloud/ndb/model.py
dhermes: I suppose I should've stated my comment in the form of a question. Should I:

1. Add a comment that this is buggy
1. Fix the apparent bug

ndb/src/google/cloud/ndb/model.py
dhermes: I don't think it'd be clearer, so I'll just leave it be (for now). (My biggest testing peeve is that a closure doesn't have a good reference for what scope is enclosed, whereas a `__call__`-able instance does.)

ndb/tests/unit/test_model.py
tseaver: I know it is widely used -- that doesn't mean it doesn't have warts.  Majyk names injected into testcase modules is pretty icky.

ndb/src/google/cloud/ndb/model.py
tseaver: I would fix the bug, myself.

ndb/src/google/cloud/ndb/model.py
tseaver: Could you define the cache on the base class, and then use the fully-qualified name of `cls` as the key?

ndb/src/google/cloud/ndb/model.py
tseaver: Seems like the "closuer doesn't have a good reference" bit affects testability, too.

ndb/src/google/cloud/ndb/query.py
tseaver: FWIW, two spaces after period / colon / bang was "correct" in typewriter days:  it made things clearer for monspaced fonts, where we (ironically?) now spend all our time as programmers.

ndb/src/google/cloud/ndb/model.py
dhermes: Done

ndb/src/google/cloud/ndb/model.py
dhermes: Done

dhermes:
@tseaver @chrisrossi PTAL, review comments have been addressed

Issue #6301, opened 2018-10-25
BigQuery: 'QueryJob.result' raises 500, even with 'retry' passed when creating the job


bencaine1:
OS: Linux dc32b7e8763a 4.9.0-6-amd64 #1 SMP Debian 4.9.82-1+deb9u3 (2018-03-02) x86_64 x86_64 x86_64 GNU/Linux
Python version: Python 2.7.6
google-cloud-bigquery: 1.5.0

My code:
```
from google.cloud import bigquery
from google.cloud.bigquery.retry import DEFAULT_RETRY

bigquery.Client(project=project_id)
query_job = client.query(query, job_config=config, retry=DEFAULT_RETRY)
query_job.result()
```

I'm getting stack traces like:
```
...
  File "/opt/conda/lib/python2.7/site-packages/google/cloud/bigquery/job.py", line 2643, in result
    super(QueryJob, self).result(timeout=timeout)
  File "/opt/conda/lib/python2.7/site-packages/google/cloud/bigquery/job.py", line 688, in result
    return super(_AsyncJob, self).result(timeout=timeout)
  File "/opt/conda/lib/python2.7/site-packages/google/api_core/future/polling.py", line 120, in result
    raise self._exception
InternalServerError: 500 Error encountered during execution. Retrying may solve the problem.
```

Could it be because PollingFuture doesn't take a Retry object (per [this](https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L698) TODO)?

tseaver:
@bencaine1 that comment indicates that the caller cannot pass in a `retry` argument to `QueryJob.result`.  The `_AsyncJob` base class inherits from `google.api_core.future.polling.PollingFuture`, which takes a `retry` constructor argument we aren't passing:  it therefore uses a retry which only handles its private `_OperationNotComplete` exception.

As a workaround, you might be able to just replace the `job._retry` attribute with a more robust one.  E.g.:

```python
from google.api_core.future.polling import _OperationNotComplete
from google.api_core.retry import Retry
from google.cloud import bigquery
from google.cloud.bigquery.retry import DEFAULT_RETRY
from google.cloud.bigquery.retry import _should_retry

def _predicate(exc):
      return isinstance(exc, _OperationNotComplete) or _should_retry(exc)

result_retry = Retry(predicate=_predicate)

bigquery.Client(project=project_id)
[...]

PR #6305, opened 2018-10-25
api_core: retry transient errors in 'PollingFuture.result'.

tseaver:
See: #6301.


tseaver:
@crwilcox ping?

crwilcox:
This LGTM. We retry those errors in some libraries explicitly and they should be retry-able. If not in a hurry, it might be nice to get @theacodes to sign off as well.

PR #6309, opened 2018-10-25
Bigtable: fix error from new flake8 version.

tseaver:
Presumably, anyway, as the line of code in question hasn't changed for
almost a year.


tseaver:
Merging w/o review to leave `master` green for Bigtable.

PR #6291, opened 2018-10-23
WIP: Refactor conformance tests for firestore.

tseaver:
See #6290.

Use 'pytest.mark.parametrize' to create a testcase per textproto file.

Note that a *bunch* of them fail. :(

PR #6320, opened 2018-10-26
Firestore: fix error from new flake8 version.

tseaver:
Note that this will conflict with #6291, but the master is red.


tseaver:
Merging without review, to fix CI on `master`.

PR #6331, opened 2018-10-29
Release api_core 2018.10.29

tseaver:


tseaver:
Silly releasetool, we no calver here.

PR #6333, opened 2018-10-29
Continuation of `ndb` Property implementation (part 5)

dhermes:

- `Property._get_value`
- `Property._delete_value`
- `Property._is_initialized`
- Descriptors for `Property` (i.e. `__get__`, `__set__` and `__delete__`)
- `Property._prepare_for_put`
- `Property._check_property`
- `Property._get_for_dict`

----

Luckily this class is almost finished, with `_serialize`, `_deserialize`, `_prepare_for_put` (a no-op), `_check_property` and `_get_for_dict` the last remaining methods. I **think** I'll be able to port `_serialize|_deserialize` to the new [protobuf definitions][1] but I haven't looked into it deeply.

~~I'll probably update this PR with the implementations for `_prepare_for_put`, `_check_property` and `_get_for_dict` as well.~~

[1]: https://github.com/googleapis/googleapis/tree/caa431d9ddb71a29b14ff6bfa6ccd7c044cf9697/google/datastore/v1

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -1190,6 +1190,31 @@ class Property(ModelAttribute):
 
         return value
 
+    def _get_value(self, entity):
+        """Get the value for this property from an entity.
+
+        For a repeated property this initializes the value to an empty
+        list if it is not set.
+
+        Args:
+            entity (Model): An entity to get a value from.
+
+        Returns:
+            Any: The user value stored for the current property.
+
+        Raises:
+            UnprojectedPropertyError: If the ``entity`` is the result of a
+                projection query and the current property is not one of the
+                projected properties.
+        """
+        if entity._projection:
+            if self._name not in entity._projection:
+                raise UnprojectedPropertyError(
+                    "Property {} is not in the projection".format(self._name)
+                )
+
+        return self._get_user_value(entity)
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -1190,6 +1190,157 @@ def _apply_to_values(self, entity, function):
...
+            entity (Model): An entity to get a value from.
+        """
+        if self._name in entity._values:
+            del entity._values[self._name]
+
+    def _is_initialized(self, entity):
+        """Ask if the entity has a value for this property.
+
+        This returns :data:`False` if a value is stored but the stored value
+        is :data:`None`.
```
chrisrossi: I'd like to know more about the rationale for that. Intuitively, it seems like it should return True, if it's supposed to be a test for whether a value is stored and not a test of the truthiness of the value.

chrisrossi: Assuming we're stuck with this because of history, though, this all looks good to me.

dhermes: 1. Yes stuck because history
1. I think the name `_is_initialized` isn't great, but the purpose of the method isn't so bad. It's essentially answering "is a value set (if a value was supposed to be set)?"

PR #6336, opened 2018-10-30
Bigtable: Support DirectRow without a table

sduskis:


sduskis:
The problem in kokoro turns out to be in snippets.  That will have to be fixed before anything else gets merged.

sduskis:
It looks like a flaky system test.  We'll have to update the timeouts.  For now, rerunning the tests.

tseaver:
#6343 resolves the unrelated failing Bigtable snippet.

PR #6347, opened 2018-10-30
Avoid deleting blob when it's renamed to itself

tjcelaya:
An issue was encountered where the new name for a blob was expected to differ from the existing name and the object was being deleted as a result of the copy-then-delete strategy.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -904,7 +904,8 @@ class Bucket(_PropertyMixin):
         If :attr:`user_project` is set, bills the API request to that project.
 
         Effectively, copies blob to the same bucket with a new name, then
-        deletes the blob.
+        deletes the blob. In case the new and old names are the same, the blob
+        is returned without any operations occurring.
 
         .. warning::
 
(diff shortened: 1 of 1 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -926,6 +927,9 @@ def rename_blob(self, blob, new_name, client=None):
         :rtype: :class:`Blob`
         :returns: The newly-renamed blob.
         """
+        if blob.name == new_name:
+            return blob
+
```
tseaver: @frankyn I wonder if we should still do the copy, and just omit the delete if the names are the same.  I presume that would update the `generation` field, which might or might not be a Good Thing(TM).

@tjcelaya Whichever way it sorts out, we are going to need a new unit test which exercises the new codepath.

frankyn: w.r.t @tseaver comment, the copy operation should still occur though omit the delete request.

tseaver:
@tjcelaya Thanks for the patch!  In order to get test coverage, and the different semantics requested by @frankyn, I have opened #6365, which supersedes this PR.

frankyn:
Let's close this PR out if it's no longer needed. Apologies if this is not the case.

PR #6351, opened 2018-10-31
Asset: fix docs build.

tseaver:
Broken in #6349.

PR #6355, opened 2018-10-31
Avoid broken 'google-common-apis 1.5.4' release.

tseaver:
Install `api_core` dependency in-place in affected packages.


tseaver:
CircleCI failure is due to the docstring borkage fixed in #6351, which needs *this* fix to turn green.

PR #6361, opened 2018-11-01
Vision: pick up addition of 'product_search_client' to 'vision_v1'

tseaver:
Update `synth.py` to deal with related docstring issues.

Closes #6359.


tseaver:
The security center build cannot pass until #6356 is merged.

PR #6365, opened 2018-11-01
Storage: avoid deleting blob renamed to itself

tseaver:
Closes #6347.

This version of the PR preserves the copy operation (as requested by @frankyn) and only avoids the deletion.  It also neatens up how the copy / rename tests create mock blobs.


tseaver:
Storage system test failure: #6368.

PR #6371, opened 2018-11-02
Ignore generated file types.py when linting

crwilcox:
Making the flake8 files, excepting core and api_core, consistent.

### DIFF
--- a/bigquery/.flake8
+++ b/bigquery/.flake8
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
   *_pb2.py
 
   # Standard linting exemptions.
--- a/resource_manager/.flake8
+++ b/resource_manager/.flake8
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
   *_pb2.py
 
   # Standard linting exemptions.
--- a/runtimeconfig/.flake8
+++ b/runtimeconfig/.flake8
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
   *_pb2.py
 
   # Standard linting exemptions.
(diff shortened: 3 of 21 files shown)


bigquery/.flake8
```diff
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
```
tseaver: Not relevant for BigQuery, and might actually clash with a manually-created file later.

resource_manager/.flake8
```diff
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
```
tseaver: Not relevant for ResourceManager, and might actually clash with a manually-created file later.

runtimeconfig/.flake8
```diff
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
```
tseaver: Not relevant for Runtimeconfig, and might actually clash with a manually-created file later.

storage/.flake8
```diff
@@ -3,6 +3,7 @@ exclude =
   # Exclude generated code.
   **/proto/**
   **/gapic/**
+  **/types.py
```
tseaver: Not relevant for Storage and might actually clash with a manually-created file later.

bigquery/.flake8
crwilcox: @tseaver, I know we are planning to use templating later. I can remove this for now, but I expect that later it will get added back in.  Though, to be fair, I also expect `black`ening to happen soonish also.

tseaver: @crwilcox My point is that Bigquery is a manual, not-even-gRPC wrapper, and that masking `types.py` from is both pointless (we don't have an autogenned `types.py` now, and likely never will) and creates a potential future hygeine bug (should we ever create `types.py` by hand).

Blindly applying templates to repositories without considering downsides / fitness-for-purpose is not going to be a win.

crwilcox:
closing this PR. we can bring this back in templates now that they are checked in.

PR #6379, opened 2018-11-02
Spanner: rationalize 'all_types' round-trip systest

tseaver:
- Expand `all_types` table to include arrays of all types. 
- Remove redundant query param bindings tests using `all_types` table.
- Add tableless tests for float bindings with transfinite values.
- Add tableless tests for queries returning explicit transfinite values.
- Refactor `all_types` data to make feature coverage obvious (null values, non-null values, empty arrays, non-empty arrays w/ mixed non-null and null items).


tseaver:
@crwilcox PTAL

tseaver:
Testing-only change.  Even if further work is needed for tagging, this makes the round-trip intent so much clearer that it is a no-brainer.

Issue #6251, opened 2018-10-17
Complete partial `ndb` implementations


dhermes:
This is a list of "cleanup" I need to go back and perform for `ndb`. I am doing this for one of a few reasons:

- The feature depended on "legacy" Google App Engine features or modules (e.g. `google.appengine.api.datastore_types`)
- The class hierarchy is so intertwined that `ndb.foo.F` and `ndb.bar.G` must be implemented at the exact same time to actually perform all expected functions

----

- [ ] `query.ParameterNode.resolve()` (added in #6250)
- [ ] `query.FilterNode._to_filter()` (added in #6250)
- [ ] `query.FilterNode.__new__()` (added in #6250) handling of `isinstance(value, model.Key)` is subject to change

averikitsch:
query.Query (added in #7239 missing `datastore_query.Order` type for orders and `QueryOptions` for default_options, as well as ancestor tests)

PR #6389, opened 2018-11-05
Adding virtual _serialize|_deserialize for `ndb` Property.

dhermes:
These are virtual (for now) because they are really only needed from `Model._to_pb()` and `Model._from_pb()` so these likely won't be needed (depending on support for the previous protobuf API).

----

I added them to #6251 to track the "partial" state of the implementation.


dhermes:
Though I didn't end up writing code here I wanted to share what I learned while figuring out what the old implementations did. 

----

The equivalent to [`Entity`][1] used on App Engine standard is (this is `proto2` and I've abbreviated a bit)

```
EntityProto
  required Reference key
  required Path entity_group
  optional User owner
  optional EntityProto.Kind kind
  optional string kind_uri
  repeated Property property
  repeated Property raw_property
```

----

The `Property._serialize` method operated on a

```
Property
  optional Property.Meaning meaning
  optional string meaning_uri
  required string name
  required PropertyValue value
  required bool multiple
  optional int32 stashed
  optional bool computed
```

The combination of a `Property` and `PropertyValue` are equivalent to a [`Value`][3] in the new API. For completeness, here is `PropertyValue`:

```
PropertyValue
  optional int64 int64Value
  optional bool booleanValue
  optional string stringValue
  optional double doubleValue
  optional PropertyValue.PointValue
  optional PropertyValue.UserValue  // identical to User
  optional PropertyValue.ReferenceValue  // identical to Reference
```

----

The `Property.Meaning` enum is [not really supported][2] in the new API:

> The ``meaning`` field should only be populated for backwards compatibility.

----

Since there is no [`ArrayValue`][4], a repeated value just ends up as multiple entries in `Property`:

```python
>>> from google.appengine.datastore import entity_pb
>>> from google.appengine.ext import ndb
>>>
>>>
>>> class M(ndb.Model):
>>>     a = ndb.IntegerProperty(repeated=True, write_empty_list=True)
>>>     b = ndb.IntegerProperty(repeated=True, write_empty_list=False)
>>>
>>>
>>> m = M(a=[1111, 2222])
>>> pb = entity_pb.EntityProto()
>>> M.a._serialize(m, pb)
>>> print(pb)
property <
  name: "a"
  value <
    int64Value: 1111
  >
  multiple: true
>
property <
  name: "a"
  value <
    int64Value: 2222
  >
  multiple: true
>

```

----

When a repeated field is empty, the output will depend on the `write_empty_list` value:

```python
>>> m = M()
>>> pb = entity_pb.EntityProto()
>>> M.b._serialize(m, pb)
>>> print(pb)

>>> pb = entity_pb.EntityProto()
>>> M.a._serialize(m, pb)
>>> print(pb)
property <
  meaning: 24
  name: "a"
  value <
  >
  multiple: false
>

```

The meaning `24` is the enum `EMPTY_LIST`.

----

The code in `Property._serialize` uses dot-delimited names for nested properties

```python
>>> class N(ndb.Model):
>>>     c = ndb.StructuredProperty(M)
>>>
>>>
>>> n = N(c=M(a=[98765], b=[1, 2]))
>>> pb = entity_pb.EntityProto()
>>> N.c._serialize(n, pb)
>>> print(pb)
property <
  name: "c.a"
  value <
    int64Value: 98765
  >
  multiple: true
>
property <
  name: "c.b"
  value <
    int64Value: 1
  >
  multiple: true
>
property <
  name: "c.b"
  value <
    int64Value: 2
  >
  multiple: true
>

```

but in the new API this is accomplished via [`Value.entity_value`][5] so the property names don't need to be combined.

----

The distinction between "single" and "repeated" fields is punted to `Property._get_base_value_unwrapped_as_list`, which will always return a list. This is a bit gnarly, in particular for a projection. A projection query will return repeated fields as single values and the original source handles this in a hard-to-understand way:

```python
if projection:
  # Projected properties have the INDEX_VALUE meaning and only contain
  # the original property's name and value.
  new_p = entity_pb.Property()
  new_p.set_name(p.name())
  new_p.set_meaning(entity_pb.Property.INDEX_VALUE)
  new_p.set_multiple(False)
  new_p.mutable_value().CopyFrom(v)
  p.CopyFrom(new_p)
```

[1]: https://github.com/googleapis/googleapis/blob/58ecdb1f0c0297975e68df5b45200a2d06f6d933/google/datastore/v1/entity.proto#L188
[2]: https://github.com/googleapis/googleapis/blob/58ecdb1f0c0297975e68df5b45200a2d06f6d933/google/datastore/v1/entity.proto#L175
[3]: https://github.com/googleapis/googleapis/blob/58ecdb1f0c0297975e68df5b45200a2d06f6d933/google/datastore/v1/entity.proto#L124
[4]: https://github.com/googleapis/googleapis/blob/58ecdb1f0c0297975e68df5b45200a2d06f6d933/google/datastore/v1/entity.proto#L115
[5]: https://github.com/googleapis/googleapis/blob/58ecdb1f0c0297975e68df5b45200a2d06f6d933/google/datastore/v1/entity.proto#L166

ndb/src/google/cloud/ndb/model.py
```diff
@@ -17,6 +17,9 @@
 
 import inspect
 
+from google.cloud.datastore_v1.gapic import enums
```
chrisrossi: Not used?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -17,6 +17,9 @@
...
+from google.cloud.datastore_v1.proto import entity_pb2
```
chrisrossi: Not used (except in docstring)?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -80,6 +83,7 @@
 ]
 
 
+_EMPTY_LIST_MEANING = 24
```
chrisrossi: Unused? Fairly cryptic.

ndb/src/google/cloud/ndb/model.py
dhermes: DERP. Sorry.

ndb/src/google/cloud/ndb/model.py
dhermes: Yeah this was part of my deleted implemenation. Sorry.

PR #6393, opened 2018-11-05
Adding class docstring for `ndb` `Property`.

dhermes:
In the process:

- Removing `sphinx-docstring-typing` from the `ndb` docs process (the extension made it impossible to refer to `TextProperty`)
- Adding the Python `typing` helpers to `nitpick_ignore`
- Adding `autoclass` and `automethod` for some public members that have non-public names

----

I've marked this "do not merge" because there are several things that need discussion:

-   Is there a better way to handle the "`Text` gets replaced by `~typing.Text`" issue caused by `sphinx-docstring-typing`? (cc @theacodes) **UPDATE**: Sphinx does have support for type annotations, so maybe we should just do that?
-   @chrisrossi This would be a good time for us to decide which non-public methods should be documented as public. The [docs][1] you previously pointed to just repeat what's in the `Property` docstring:

    > All special `Property` attributes, even those considered 'public', have names starting with an underscore. This is because `StructuredProperty` uses the non-underscore attribute namespace to refer to nested `Property` names; this is essential for specifying queries on subproperties.

-   The `_BaseValue` class likely shouldn't be documented, but there are quite a few references to it.

[1]: https://cloud.google.com/appengine/docs/standard/python/ndb/subclassprop

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -455,6 +455,22 @@ class Property(ModelAttribute):
     .. automethod:: _serialize
     .. automethod:: _call_to_base_type
     .. automethod:: _call_from_base_type
+
+    Args:
+        name (str): The name of the property.
+        indexed (bool): Indicates if the value should be indexed.
+        repeated (bool): Indicates if this property is repeated, i.e. contains
+            multiple values.
+        required (bool): Indicates if this property is required on the given
+            model type.
+        default (Any): The default value for this property.
+        choices (Iterable[Any]): A container of allowed values for this
+            property.
+        validator (Callable[[Property, Any], bool]): A validator to be used
+            to check values.
+        verbose_name (str): A longer, user-friendly name for this property.
+        write_empty_list (bool): Indicates if an empty list should be written
+            to the datastore.
     """
 
     # Instance default fallbacks provided by class.
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Model classes for datastore objects and properties for models.
-
-.. autoclass:: _BaseValue
-"""
+"""Model classes for datastore objects and properties for models."""
 
 
 import inspect
@@ -351,11 +348,14 @@ class Property(ModelAttribute):
         describe properties of various types (and :class:`GenericProperty`
         which describes a dynamically typed property).
 
-    All special property attributes, even those considered "public",
-    have names starting with an underscore, because :class:`StructuredProperty`
-    uses the non-underscore attribute namespace to refer to nested
-    property names; this is essential for specifying queries on
-    subproperties.
+    The :class:`Property` does not reserve any "public" names (i.e. names
+    that don't start with an underscore). This is intentional; the subclass
+    :class:`StructuredProperty` uses the public attribute namespace to refer to
+    nested property names (this is essential for specifying queries on
+    subproperties).
+
+    The :meth:`IN` attribute is provided as an alias for ``_IN``, but ``IN``
+    can be overridden if a subproperty has the same name.
 
     The :class:`Property` class and its predefined subclasses allow easy
     subclassing using composable (or stackable) validation and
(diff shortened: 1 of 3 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -336,6 +339,124 @@ def __hash__(self):
...
+    A property will be a member of a :class:`Model` and will be used to help
+    store values in an ``entity`` (i.e. instance of a model subclass). The
+    value(s) for a given property ``name`` will be stored in
+    ``entity._values[name]`` and accessed by :meth:`_store_value` and
+    :meth:`_retrieve_value`. For these methods, the values can be either user
+    values or base values. To retrieve user values, use
+    :meth:`_get_user_value`. To retrieve base values, use
+    :meth:`_get_base_value`. In particular, :meth:`_get_value` calls
+    :meth:`_get_user_value`, and :meth:`_serialize` effectively calls
+    :meth:`_get_base_value`.
```
chrisrossi: Who is the audience for `_get_*_value()` and `_store_value()`? Are docs targeting the same audience?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -336,6 +339,124 @@ def __hash__(self):
...
+
+    To store a user value, just call :meth:`_store_value`. To store a
+    base value, wrap the value in a :class:`_BaseValue` and then
+    call :meth:`_store_value`.
```
chrisrossi: Why not `_store_user_value` and `_store_base_value`?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -336,6 +339,124 @@ def __hash__(self):
...
+    * Comparisons: The comparison operations call ``_to_base_type()`` on
+      their operand.
+    * Distinguishing between user and base values: the infrastructure
+      guarantees that ``_from_base_type()`` will be called with an
+      (unwrapped) base value, and that ``_to_base_type()`` will be called
+      with a user value.
+    * Returning the original value: if any of these return :data:`None`, the
+      original value is kept. (Returning a different value not equal to
+      :data:`None` will substitute the different value.)
+
```
chrisrossi: While clear the audience needs to know about `_validate`, `_to_base_type`, and `_from_base_type`, it's far less clear the audience needs to know about `_get_user_value`, `_get_base_value`, or `_BaseValue`. See question above about audience.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -336,6 +339,124 @@ def __hash__(self):
...
+      ``_to_base_type()`` for each list item in a repeated value.
+    * Wrapping values in :class:`_BaseValue`: The wrapping and unwrapping is
+      taken care of by the infrastructure that calls the composable APIs.
+    * Comparisons: The comparison operations call ``_to_base_type()`` on
+      their operand.
+    * Distinguishing between user and base values: the infrastructure
+      guarantees that ``_from_base_type()`` will be called with an
+      (unwrapped) base value, and that ``_to_base_type()`` will be called
+      with a user value.
+    * Returning the original value: if any of these return :data:`None`, the
```
chrisrossi: I'm uncomfortable with the use of `None` as a semaphore for "do nothing." However remote, it seems there could be a possibility of a validator intending to convert a value to `None`. In general, in these kinds of scenarios, I prefer an explicit, unambiguous semaphore, usually passed into the method that might use it:

def _from_base_type(self, value, _nope):
    # Base value to (strict) user value.'
    if not isinstance(value, SomeBaseType):
        return SomeUserType(value)
    return _nope

... but with a better name than `_nope`.  I suspect we're constrained by backwards compatibility considerations on this point, though, so this is probably more for posterity than anything else.

ndb/src/google/cloud/ndb/model.py
dhermes: Good question. It's a house of cards I suppose. The whole `_BaseValue` concept perplexes me.

ndb/src/google/cloud/ndb/model.py
dhermes: Yes, we are constrained here. The documented behavior is that a return value of `None` in the MRO chain (of `_(to|from)_base_type()`) just means that transformation didn't make any changes.

The composition / stacking is done via `_call_(to|from)_base_type()`.

ndb/src/google/cloud/ndb/model.py
dhermes: Jinx!

ndb/src/google/cloud/ndb/model.py
chrisrossi: :+1: My impression exactly.

ndb/src/google/cloud/ndb/model.py
chrisrossi: It's not as important to me if we're not going to pawn it off on the users. Not that the code shouldn't be clean internally, but if we're taking these out of the "public" API, then it's just an implementation detail.

ndb/src/google/cloud/ndb/model.py

dhermes:
@chrisrossi PTAL

dhermes:
I filed #6428 to follow up on the whole Sphinx / `typing` issue.

PR #6394, opened 2018-11-05
Implementing `BooleanProperty` in `ndb`.

dhermes:
The `_db_set_value()` and `_db_get_value()` methods are just helpers for `Property._serialize()` and `Property._deserialize()`. As in #6389, I'm fairly certain that this won't be needed so for now I'm not implementing. (The previous implementation relies on the old protobuf definition, which is "stuck" inside Google App Engine standard.)

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -1387,9 +1387,42 @@ class ModelKey(Property):
 
 
 class BooleanProperty(Property):
+    """A property that contains values of type bool."""
+
     __slots__ = ()
 
-    def __init__(self, *args, **kwargs):
+    def _validate(self, value):
+        """Validate a ``value`` before setting it.
+
+        Args:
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -1387,9 +1387,42 @@ def __init__(self, *args, **kwargs):
...
+
+        Args:
+            value (bool): The value to check.
+
+        Returns:
+            bool: The passed-in ``value``.
+
+        Raises:
+            .BadValueError: If ``value`` is not a :class:`bool`.
+        """
```
chrisrossi: No lax values or coercion?

dhermes: Not in the original and IMO that is the "correct" implementation.

chrisrossi: Ok.

PR #6396, opened 2018-11-05
Implementing `FloatProperty` in `ndb`.

dhermes:


ndb/src/google/cloud/ndb/model.py
```diff
@@ -1532,7 +1532,10 @@ def __init__(self, *args, **kwargs):
 
 
 class BooleanProperty(Property):
-    """A property that contains values of type bool."""
+    """A property that contains values of type bool.
+
+    .. automethod:: _validate
```
dhermes: This change is "smuggled" in here (should've gone in #6394 but the discussion in #6395 is what led to this style).

PR #6398, opened 2018-11-06
Implementing `BlobProperty` in `ndb`.

dhermes:
This is done in several commits, so the review is likely better done commit-by-commit.

### DIFF
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -57,7 +57,7 @@ The primary differences come from:
   the limit has been raised by the backend. (FWIW, Danny's opinion is that
   the backend should enforce these limits, not the library.)
 - `Property.__creation_counter_global` has been removed as it seems to have
-  been included for a feature that was never implemented. See 
+  been included for a feature that was never implemented. See
   [Issue #175][1] for original rationale for including it and [Issue #6317][2]
   for discussion of its removal.
 - `ndb` uses "private" instance attributes in many places, e.g. `Key.__app`.
@@ -86,6 +86,10 @@ The primary differences come from:
 - `eventloop` has been renamed to `_eventloop`. It is believed that `eventloop`
   was previously a *de facto* private module, so we've just made that
   explicit.
+- `BlobProperty._datastore_type` has not been implemented; the base class
+  implementation is sufficient. The original implementation wrapped a byte
+  string in a `google.appengine.api.datastore_types.ByteString` instance, but
+  that type was mostly an alias for `str` in Python 2
 
 ## Comments
 
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -90,6 +90,11 @@ The primary differences come from:
   implementation is sufficient. The original implementation wrapped a byte
   string in a `google.appengine.api.datastore_types.ByteString` instance, but
   that type was mostly an alias for `str` in Python 2
+- `BlobProperty._validate` used to special case for "too long when indexed"
+  if `isinstance(self, TextProperty)`. We have removed this check since
+  the implementation does the same check in `TextProperty._validate`.
+- The `BlobProperty` constructor only sets `_compressed` if explicitly
+  passed. The original set always (and used `False` as default)
 
 ## Comments
 
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -249,7 +249,7 @@ class IndexState:
         )
 
     def __eq__(self, other):
-        """Compare two indexes."""
+        """Compare two index states."""
         if not isinstance(other, IndexState):
             return NotImplemented
 
@@ -1481,6 +1481,39 @@ class FloatProperty(Property):
 
     def __init__(self, *args, **kwargs):
         raise NotImplementedError
+
+
+class _CompressedValue:
+    """A marker object wrapping compressed values.
+
+    Args:
+        z_val (bytes): A return value of ``zlib.compress``.
+    """
+
+    __slots__ = ("z_val",)
+
+    def __init__(self, z_val):
+        if not isinstance(z_val, bytes):
+            raise TypeError("Only bytes accepted", z_val)
+
+        self.z_val = z_val
+
+    def __repr__(self):
+        return "_CompressedValue({!r})".format(self.z_val)
+
+    def __eq__(self, other):
+        """Compare two compressed values."""
+        if not isinstance(other, _CompressedValue):
+            return NotImplemented
+
+        return self.z_val == other.z_val
+
+    def __ne__(self, other):
+        """Inequality comparison operation."""
+        return not self == other
+
+    def __hash__(self):
+        raise TypeError("_CompressedValue is not immutable")
 
 
 class BlobProperty(Property):
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -1483,10 +1483,63 @@ class FloatProperty(Property):
 
 
 class BlobProperty(Property):
-    __slots__ = ()
+    """A property that contains values that are byte strings.
 
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    Args:
+        name (str): The name of the property.
+        compressed (bool): Indicates if the value should be compressed (via
+            ``zlib``).
+        indexed (bool): Indicates if the value should be indexed.
+        repeated (bool): Indicates if this property is repeated, i.e. contains
+            multiple values.
+        required (bool): Indicates if this property is required on the given
+            model type.
+        default (bytes): The default value for this property.
+        choices (Iterable[bytes]): A container of allowed values for this
+            property.
+        validator (Callable): A validator to be used to check values.
+        verbose_name (str): A longer, user-friendly name for this property.
+        write_empty_list (bool): Indicates if an empty list should be written
+            to the datastore.
+
+    Raises:
+        NotImplementedError: If the property is both compressed and indexed.
+    """
+
+    _indexed = False
+    _compressed = False
+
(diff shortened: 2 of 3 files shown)


ndb/MIGRATION_NOTES.md
```diff
@@ -57,7 +57,7 @@ The primary differences come from:
   the limit has been raised by the backend. (FWIW, Danny's opinion is that
   the backend should enforce these limits, not the library.)
 - `Property.__creation_counter_global` has been removed as it seems to have
-  been included for a feature that was never implemented. See 
+  been included for a feature that was never implemented. See
```
dhermes: This is just my editor settings cleaning up trailing whitespace

ndb/src/google/cloud/ndb/model.py
```diff
@@ -248,7 +250,7 @@ def __repr__(self):
         )
 
     def __eq__(self, other):
-        """Compare two indexes."""
+        """Compare two index states."""
```
dhermes: This was copy-pasta

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1407,10 +1409,212 @@ def __init__(self, *args, **kwargs):
...
+
+    Args:
+        z_val (bytes): A return value of ``zlib.compress``.
+    """
+
+    __slots__ = ("z_val",)
+
+    def __init__(self, z_val):
+        if not isinstance(z_val, bytes):
+            raise TypeError("Only bytes accepted", z_val)
```
dhermes: This was an `assert` in the original

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1407,10 +1409,212 @@ def __init__(self, *args, **kwargs):
...
-    __slots__ = ()
+    """A property that contains values that are byte strings.
 
-    def __init__(self, *args, **kwargs):
+    .. note::
+
+        Unlike most property types, a :class:`BlobProperty` is **not**
+        indexed by default.
+
+    Args:
```
dhermes: I should probably add something like this in #6393

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1407,10 +1409,212 @@ def __init__(self, *args, **kwargs):
...
+            indexed=indexed,
+            repeated=repeated,
+            required=required,
+            default=default,
+            choices=choices,
+            validator=validator,
+            verbose_name=verbose_name,
+            write_empty_list=write_empty_list,
+        )
+        self._compressed = compressed
```
dhermes: This goes against the idea of "fallback to class attributes for defaults" used by `Property`

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1407,10 +1409,212 @@ def __init__(self, *args, **kwargs):
...
+            to the datastore.
+
+    Raises:
+        NotImplementedError: If the property is both compressed and indexed.
+    """
+
+    _indexed = False
+    _compressed = False
+
+    def __init__(
```
dhermes: Copying the entire argument list from `Property` is brittle / bad software design, but it's "necessary" to get `__repr__` to work correctly. I also think the trade-off is worth it since `BlobProperty` is nearly a decade old (i.e. the interface is set in stone).

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1407,10 +1409,212 @@ def __init__(self, *args, **kwargs):
...
+                exceeds the maximum length (1500 bytes).
+        """
+        if not isinstance(value, bytes):
+            raise exceptions.BadValueError(
+                "Expected bytes, got {!r}".format(value)
+            )
+
+        if (
+            self._indexed
+            and not isinstance(self, TextProperty)
```
dhermes: This should probably be moved into `TextProperty._validate` for special handling. I think it's bad form for a base class to refer to a child class in this way.

ndb/tests/unit/test_model.py
```diff
@@ -1317,11 +1318,182 @@ def test_constructor():
...
+    @staticmethod
+    def test_constructor_compressed_and_indexed():
+        with pytest.raises(NotImplementedError):
+            model.BlobProperty(name="foo", compressed=True, indexed=True)
+
+    @staticmethod
+    def test__value_to_repr():
+        prop = model.BlobProperty(name="blob")
+        as_repr = prop._value_to_repr(b"abc")
+        assert as_repr == "b'abc'"
```
dhermes: ISTM this is not what was intended for the original

ndb/src/google/cloud/ndb/model.py
chrisrossi: It looks like this check should be unnecessary, given _CompressedValue is private and assuming we trust the output of `zlib.compress`. Doesn't hurt, though.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Is there a reason for the inconsistency?

ndb/src/google/cloud/ndb/model.py
chrisrossi: Yep.

ndb/tests/unit/test_model.py
chrisrossi: What was intended?

ndb/src/google/cloud/ndb/model.py
chrisrossi: OK.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Yes, probably.

ndb/src/google/cloud/ndb/model.py
dhermes: Yeah that's a good point and agrees with the concept of an `assert` over a full-fledged exception. I am :+1: on just removing this type check if you are?

ndb/src/google/cloud/ndb/model.py
dhermes: I'd imagine it was caused by the developers of `BlobProperty` not knowing about the decision made for `Property` (or rather that the implementers didn't really care / notice what they were doing).

ndb/tests/unit/test_model.py
dhermes: Well, what I mean to say is that the `repr()` has changed across Python versions:

```
$ python2.7 -c 'print(repr(b"abc"))'
'abc'
$ python3.7 -c 'print(repr(b"abc"))'
b'abc'
```

ndb/src/google/cloud/ndb/model.py
dhermes: Done.

ndb/src/google/cloud/ndb/model.py
dhermes: So I dug a little deeper on this and realized the issue. The `Property._call_to_base_type` and `Property._call_shallow_validation` methods will call **every** `_validate` method on the MRO, so this will get invoked for `TextProperty` instances **even though** `TextProperty._validate` is implemented (and does not invoke `super()`).

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py
chrisrossi: I'd say make it consistent unless there's some difficulty I'm not seeing.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Right, but `TextProperty._validate` should get called before this one, and if it raises `BadValueError` this won't get called, right? Still seems like `TextProperty._validate` is still the right place for it.

ndb/tests/unit/test_model.py
chrisrossi: Ok, I wouldn't worry about that.

dhermes: :+1:

ndb/src/google/cloud/ndb/model.py
dhermes: Not quite. What's happening here is an error for "index properties with greater than 1500 bytes, except for a text property". So it's a custom **exemption** for `TextProperty`.

I could rely on an attribute like `self._LENGTH_EXEMPT` and then just define `BlobProperty._LENGTH_EXEMPT = False` and `TextProperty._LENGTH_EXEMPT = True`.

WDYT?

chrisrossi: Ah right. Or mabye just make _MAX_STRING_LENGTH a class level attribute and have TextProperty override it with `0` or `None` meaning no limit. Either way.

dhermes: OK this is extra fun. I have no idea why it's in the [original][1] because `TextProperty._validate` also [checks][2] the length.

I'm just going to remove the type check here because it's equivalent to the branch of `TextProperty._validate` when `value` is a `bytes` instance (except for the decode to UTF-8 check).

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L1715-L1720
[2]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L1782-L1785

ndb/src/google/cloud/ndb/model.py
dhermes: Done

dhermes:
@chrisrossi PTAL. My latest commit has hopefully tied up all loose ends.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Ha, ok.

PR #6427, opened 2018-11-06
Translate: adapt system test to updated back-end translation.

tseaver:
The back-end used to translate "Me llamo Jeff" into German as "Mein Name ist Jeff";  now it uses the more idiomatic "Ich heiße Jeff".


tseaver:
Merging without review:  I have verified the change in the returned translation in the console, and `master` is red.

PR #6445, opened 2018-11-07
Implementing `BlobKeyProperty` in `ndb`.

dhermes:


ndb/src/google/cloud/ndb/blobstore.py
```diff
@@ -56,10 +57,16 @@
 ]
 
 
+BlobKey = _ports.BlobKey
```
dhermes: To follow the winding path in the original:

1. https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/blobstore.py#L107
1. https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/google_imports.py#L67
1. https://github.com/dhermes/app-engine-standard/blob/19187b658bd11bac1edb0f203c22a14444fc21ff/google/appengine/api/blobstore/__init__.py#L23
1. https://github.com/dhermes/app-engine-standard/blob/19187b658bd11bac1edb0f203c22a14444fc21ff/google/appengine/api/blobstore/blobstore.py#L74

In other words, it's the same (`datastore_types.BlobKey`) as the one in `ndb.model`

ndb/src/google/cloud/ndb/blobstore.py
```diff
@@ -56,10 +57,16 @@
...
+
 BLOB_INFO_KIND = "__BlobInfo__"
-BLOB_KEY_HEADER = "X-AppEngine-BlobKey"
 BLOB_MIGRATION_KIND = "__BlobMigration__"
+BLOB_KEY_HEADER = "X-AppEngine-BlobKey"
 BLOB_RANGE_HEADER = "X-AppEngine-BlobRange"
+MAX_BLOB_FETCH_SIZE = 1015808
+UPLOAD_INFO_CREATION_HEADER = "X-AppEngine-Upload-Creation"
+
+BlobKeyProperty = model.BlobKeyProperty
```
dhermes: I moved these up to match the source order in the original `blobstore.py`

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+import functools
+
+from google.cloud.ndb import exceptions
+
+
+_MAX_STRING_LENGTH = 1500
+
+
+@functools.total_ordering
+class BlobKey:
```
dhermes: This is not a perfect implementation of [`datastore_types.BlobKey`][1]. A few differences:

- [`ValidateString()`][2] allows a 2.7 `basestring`, but I'm only allowing `bytes` (i.e. not `str` in Python 3 / `unicode` in Python 2)
- `ValidateString()` adds a special case for `datastore_types.Blob`, which is a thin wrapper over `bytes` (`str` in Python 2). We don't have this type and I don't plan on adding it

Also, ISTM that two things are strange:

- `None` shouldn't be allowed as a blob key
- `len(blob_key) > 1500` shouldn't matter, i.e. the blob keys came from an opaque GAE service so users couldn't just make them up and have them work. I think this was not intended as part of the `ValidateString()` usage, i.e. the caller just wanted type-checking.

[1]: https://github.com/dhermes/app-engine-standard/blob/19187b658bd11bac1edb0f203c22a14444fc21ff/google/appengine/api/datastore_types.py#L1267
[2]: https://github.com/dhermes/app-engine-standard/blob/19187b658bd11bac1edb0f203c22a14444fc21ff/google/appengine/api/datastore_types.py#L162

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+
+* ``google.appengine.api.datastore_types.BlobKey``
+"""
+
+import functools
+
+from google.cloud.ndb import exceptions
+
+
+_MAX_STRING_LENGTH = 1500
```
dhermes: This is copy-pasta (also in `model.py` and `key.py` in some form)

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@functools.total_ordering
+class BlobKey:
+    """Key used to identify a blob in the blobstore.
+
+    .. note::
+
+        The blobstore was an early Google App Engine feature that later became
+        Google Cloud Storage.
```
chrisrossi: Can this also be used with Cloud Storage?

dhermes: Blobstore and Google Cloud Storage (GCS) are the same thing. But blobstore came first, and is a GAE feature that wraps around GCS. (It originally was just a blob store feature, i.e. no wrapping.)

dhermes: I realize that didn't really answer your question.

The answer is essentially "not really". I'd imagine we'll have to do a lot of work to get this to work the way we need it to, but we'll find out.

However, this property itself just stores some bytes in a column in the datastore.

chrisrossi: The reason I ask is because I'm wondering if this is worth doing. I could see dropping the integration with blobstore for now and adding integration with GCS as a new feature. If this code isn't useful for a future GCS integration, then I'm wondering why we need it at all.

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, blob_key):
+        if isinstance(blob_key, bytes):
+            if len(blob_key) > _MAX_STRING_LENGTH:
+                raise exceptions.BadValueError(
+                    "blob key must be under {:d} "
+                    "bytes.".format(_MAX_STRING_LENGTH)
+                )
+        elif blob_key is not None:
+            raise exceptions.BadValueError(
+                "blob key should be a string; received "
```
chrisrossi: "blob key should be *bytes*..."

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self._blob_key = blob_key
+
+    def __eq__(self, other):
+        if isinstance(other, BlobKey):
+            return self._blob_key == other._blob_key
+        elif isinstance(other, bytes):
+            return self._blob_key == other
+        else:
+            return NotImplemented
```
chrisrossi: Return False?

ndb/src/google/cloud/ndb/_ports.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __lt__(self, other):
+        if isinstance(other, BlobKey):
+            return self._blob_key < other._blob_key
+        elif isinstance(other, bytes):
+            return self._blob_key < other
+        else:
+            return NotImplemented
```
chrisrossi: raise TypeError?

chrisrossi:
This looks good to me for the most part. I'm still struggling with understanding why we need it.

I also wonder if, instead of `_ports`, we should we try to find true homes for anything we're having to hoover in from somewhere else, rather than a catchall. In this case, we could put `BlobKey` next to `BlobKeyProperty` in `model` or we could start our own `_datastore_types` module.

ndb/src/google/cloud/ndb/_ports.py
dhermes: Yeah I copy-pasted this from `ValidateString()`, where both `str` and `unicode` were accepted in Python 2. 

I don't think it's necessary to preserve the original error message here because ISTM all callers are private and won't pass in the wrong type. In general do you think we need to worry about preserving the **text** of error messages or just the type?

dhermes:
I think you're right about the `_datastore_types` module, so am doing a rename.

ndb/src/google/cloud/ndb/_ports.py
chrisrossi: Just the type.

chrisrossi: The text should be helpful, obviously, which might mean adapting.

PR #6448, opened 2018-11-08
Implementing `DateTimePropery` in `ndb`.

dhermes:


ndb/src/google/cloud/ndb/model.py
```diff
@@ -2144,9 +2150,154 @@ def __init__(self, *args, **kwargs):
 
 
 class DateTimeProperty(Property):
-    __slots__ = ()
+    """A property that contains :class:`~datetime.datetime` values.
 
-    def __init__(self, *args, **kwargs):
+    This property expects "naive" datetime stamps, i.e. no timezone can
```
dhermes: This was enforced by `_db_set_value`, so I'm not sure what to do.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1748,8 +1754,8 @@ class BlobProperty(Property):
     def __init__(
         self,
         name=None,
-        compressed=None,
         *,
+        compressed=None,
```
dhermes: I realized when reading the `DateTimeProperty` source that `compressed` was supposed to be positional only

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2144,9 +2150,154 @@ def __init__(self, *args, **kwargs):
...
+                raise ValueError(
+                    "DateTimeProperty {} could use auto_now and be "
+                    "repeated, but there would be no point.".format(self._name)
+                )
+            elif auto_now_add:
+                raise ValueError(
+                    "DateTimeProperty {} could use auto_now_add and be "
+                    "repeated, but there would be no point.".format(self._name)
+                )
+        if auto_now is not None:
```
dhermes: This change differs from the original (i.e. checking before setting)

ndb/src/google/cloud/ndb/model.py
chrisrossi: I'm not sure I understand the comment. Here, you've made it keyword only.

ndb/src/google/cloud/ndb/model.py
chrisrossi: What's the question?

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py
dhermes: Yes, I wrote that in a hurry. I meant to say "was supposed to be keyword only".

ndb/src/google/cloud/ndb/model.py
dhermes: Should `_validate()` raise an exception if the `datetime` instance has `tzinfo` set? It used to happen [here][1]

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2126

chrisrossi: I would look to whatever `datastore` does these days. If `datastore` doesn't care, then we shouldn't care. Otherwise, we should do whatever massaging is required to make `datastore` happy. (If datastore insists on naive datetimes, we could also just convert to utc and then strip tzinfo ourselves, rather than raising an exception.)

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py
dhermes: The protobuf [API][1] uses a [`google.protobuf.Timestamp`][2], which just uses UTC. As a result, `google-cloud-datastore` [takes][3] a `datetime.datetime` and converts it with [`_datetime_to_pb_timestamp()`][4], which will [convert to UTC][5].

[1]: https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/entity.proto#L142
[2]: https://github.com/protocolbuffers/protobuf/blob/be83b29bdd17630e4028c44b51354b2c112cabd1/src/google/protobuf/timestamp.proto#L127
[3]: https://github.com/googleapis/google-cloud-python/blob/8517524cb08b85e4c97a5cfb6963b8d65cfa73f2/datastore/google/cloud/datastore/helpers.py#L336
[4]: https://github.com/googleapis/google-cloud-python/blob/8517524cb08b85e4c97a5cfb6963b8d65cfa73f2/core/google/cloud/_helpers.py#L442
[5]: https://github.com/googleapis/google-cloud-python/blob/8517524cb08b85e4c97a5cfb6963b8d65cfa73f2/core/google/cloud/_helpers.py#L219-L222

chrisrossi: Ok, I would just accept any datetime, then, since it will all come out in the wash once it hits the storage layer. 

A nice feature might be for the `DateTimeProperty` constructor to accept an optional tzinfo argument that `_from_base_type()` could use to translate back to caller's preferred timezone.

dhermes: :+1: I'll leave this code as-is and I filed #6454 for the `tzinfo` feature discussion

dhermes:
@chrisrossi LGTY?

PR #6452, opened 2018-11-08
Implementing `DateProperty` and `TimeProperty` in `ndb`.

dhermes:
**NOTE**: This has #6448 as diffbase.

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -2363,9 +2363,9 @@ class DateTimeProperty(Property):
 
     @staticmethod
     def _now():
-        """datetime.datetime: Return current time.
+        """datetime.datetime: Return current datetime.
 
-        This is in place so it can be patched in tests.
+        Subclasses will override this to return different forms of "now".
         """
         return datetime.datetime.utcnow()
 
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -2302,17 +2302,135 @@ def _db_get_value(self, v, unused_p):
...
+            value (~datetime.date): The value to be converted.
+
+        Returns:
+            ~datetime.datetime: The converted value: a datetime object with the
+            time set to ``00:00``.
+
+        Raises:
+            TypeError: If ``value`` is not a :class:`~datetime.date`.
+        """
+        if not isinstance(value, datetime.date):
```
dhermes: There [was][1] an `assert` here but it didn't make sense because

1. `_validate()` already checked the type
1. The `_date_to_datetime()` helper checks the type.

FWIW I moved the `_date_to_datetime()` implementation here wholesale

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2183

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2302,17 +2302,135 @@ def _db_get_value(self, v, unused_p):
...
+            value (~datetime.time): The value to be converted.
+
+        Returns:
+            ~datetime.datetime: The converted value: a datetime object with the
+            date set to ``1970-01-01``.
+
+        Raises:
+            TypeError: If ``value`` is not a :class:`~datetime.time`.
+        """
+        if not isinstance(value, datetime.time):
```
dhermes: There [was][1] an `assert` here but it didn't make sense because

1. `_validate()` already checked the type
1. The `_time_to_datetime()` helper checks the type.

FWIW I moved the `_time_to_datetime()` implementation here wholesale

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2203

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2302,17 +2302,135 @@ def _db_get_value(self, v, unused_p):
...
+        """Convert a value from the "base" value type for this property.
+
+        Args:
+            value (~datetime.datetime): The value to be converted.
+
+        Returns:
+            ~datetime.time: The converted value: the time that ``value``
+            occurs at.
+        """
+        return value.time()
```
dhermes: There [was][1] an `assert` here but I removed it. ISTM that the base type is coming from the storage layer (i.e. not from users) so the caller won't mess this up.

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2207

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2302,17 +2302,135 @@ def _db_get_value(self, v, unused_p):
...
+        """Convert a value from the "base" value type for this property.
+
+        Args:
+            value (~datetime.datetime): The value to be converted.
+
+        Returns:
+            ~datetime.date: The converted value: the date that ``value``
+            occurs on.
+        """
+        return value.date()
```
dhermes: There [was][1] an `assert` here but I removed it. ISTM that the base type is coming from the storage layer (i.e. not from users) so the caller won't mess this up.

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2187

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py
chrisrossi: Yep. Same as with discussion on `JSONProperty`, assuming the passed value is `bytes` there.

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py

ndb/src/google/cloud/ndb/model.py
dhermes: Did you mean `bytes` or is that a "brain typo" (since `bytes` is the type relevant to `JSONProperty`)?

chrisrossi: It wasn't a typo, but maybe unclear language. I'm just saying that, yes, the same reasoning that allows us to assume we're getting a `datetime.datetime` here is the same that allows us to assume we're geting `bytes` in JSONProperty.

dhermes: :+1: my bad

PR #6456, opened 2018-11-08
Implement `get_event_loop`.

chrisrossi:
This has little to do with the function of the same name from the original
NDB, since the way of managing the current event loop has changed.

### DIFF
--- a/ndb/src/google/cloud/ndb/exceptions.py
+++ b/ndb/src/google/cloud/ndb/exceptions.py
@@ -22,6 +22,7 @@ legacy Google App Engine runtime.
 
 __all__ = [
     "Error",
+    "AsyncContextError",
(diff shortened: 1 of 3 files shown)


ndb/src/google/cloud/ndb/exceptions.py
```diff
@@ -33,6 +34,20 @@ class Error(Exception):
     """Base datastore error type."""
 
 
+class AsyncContextError(Error):
+    """Indicates an async call being made without a context.
+
+    Raised whenever an asynchronous call is made outside of a context
+    established by :func:`google.cloud.ndb.async_context`.
```
dhermes: @chrisrossi For future reference, I typically use the `~` here, so that only the last part of the path is rendered.

PR #6457, opened 2018-11-08
Implementing `KeyProperty` constructor in `ndb`.

dhermes:
This also required adding a small modification to `Property.__repr__` to change how it behaves for `KeyProperty`.

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -2283,6 +2283,8 @@ class KeyProperty(Property):
     :class:`type` argument is assumed to be the ``kind`` (and checked that
     the type is a subclass of :class:`Model`).
 
+    .. automethod:: _validate
+
     Args:
         name (str): The name of the property.
         kind (Union[type, str]): The (optional) kind to be stored. If provided
@@ -2437,6 +2439,54 @@ class KeyProperty(Property):
             if name in ("args", "name", "kind"):
                 continue
             yield name, is_keyword
+
+    def _validate(self, value):
+        """Validate a ``value`` before setting it.
+
+        Args:
+            value (.Key): The value to check.
+
+        Raises:
+            .BadValueError: If ``value`` is not a :class:`.Key`.
+            .BadValueError: If ``value`` is a partial :class:`.Key` (i.e. it
+                has no name or ID set).
+            .BadValueError: If the current property has an associated ``kind``
+                and ``value`` does not match that kind.
+        """
+        if not isinstance(value, Key):
+            raise exceptions.BadValueError(
+                "Expected Key, got {!r}".format(value)
+            )
+
+        # Reject incomplete keys.
+        if not value.id():
+            raise exceptions.BadValueError(
+                "Expected complete Key, got {!r}".format(value)
+            )
+
+        # Verify kind if provided.
+        if self._kind is not None:
+            if value.kind() != self._kind:
+                raise exceptions.BadValueError(
+                    "Expected Key with kind={!r}, got "
+                    "{!r}".format(self._kind, value)
+                )
+
+    def _db_set_value(self, v, unused_p, value):
+        """Helper for :meth:`_serialize`.
+
+        Raises:
+            NotImplementedError: Always. This method is virtual.
+        """
+        raise NotImplementedError
+
+    def _db_get_value(self, v, unused_p):
+        """Helper for :meth:`_deserialize`.
+
+        Raises:
+            NotImplementedError: Always. This method is virtual.
+        """
+        raise NotImplementedError
 
 
 class BlobKeyProperty(Property):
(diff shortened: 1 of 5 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -2241,9 +2252,240 @@ def __init__(self, *args, **kwargs):
...
+
+        >>> name = "my_value"
+        >>> ndb.KeyProperty(name)
+        KeyProperty(b'my_value')
+        >>> ndb.KeyProperty(SimpleModel)
+        KeyProperty(kind=b'SimpleModel')
+        >>> ndb.KeyProperty(name, SimpleModel)
+        KeyProperty(b'my_value', kind=b'SimpleModel')
+        >>> ndb.KeyProperty(SimpleModel, name)
+        KeyProperty(b'my_value', kind=b'SimpleModel')
```
dhermes: This `b'my_value'` / `b'SimpleModel'` is showing the same issue we've run into in the unit tests

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2241,9 +2252,240 @@ def __init__(self, *args, **kwargs):
...
+            repeated=repeated,
+            required=required,
+            default=default,
+            choices=choices,
+            validator=validator,
+            verbose_name=verbose_name,
+            write_empty_list=write_empty_list,
+        )
+        if kind is not None:
+            self._kind = kind
```
dhermes: As in all subtypes that add a data attribute, I'm making this optional even though the original implementation would always set `self._kind` even if `kind` were `None`

chrisrossi: Is this useful without a kind? Should it be required?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2241,9 +2252,240 @@ def __init__(self, *args, **kwargs):
 
 
 class KeyProperty(Property):
-    __slots__ = ()
+    """A property that contains :class:`.Key` values.
 
-    def __init__(self, *args, **kwargs):
+    The constructor for :class:`KeyProperty` allows at most two positional
+    arguments. Any usage of :data:`None` as a positional argument will
```
chrisrossi: The `None` filtering is weird. Any idea what that's about?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2241,9 +2252,240 @@ def __init__(self, *args, **kwargs):
...
+            Tuple[str, bool]: Pairs of argument name and a boolean indicating
+            if that argument is a keyword.
+        """
+        yield "name", False
+        yield "kind", True
+        from_inspect = super(KeyProperty, self)._constructor_info()
+        for name, is_keyword in from_inspect:
+            if name in ("args", "name", "kind"):
+                continue
+            yield name, is_keyword
```
chrisrossi: Looks like that was fun!

ndb/tests/unit/test_model.py
```diff
@@ -1784,9 +1784,142 @@ def test_constructor():
...
+        prop = model.KeyProperty("keyp", kind="Simple", repeated=True)
+        expected = "KeyProperty(b'keyp', kind=b'Simple', repeated=True)"
+        assert repr(prop) == expected
+
+    @staticmethod
+    def test__validate():
+        kind = "Simple"
+        prop = model.KeyProperty("keyp", kind=kind)
+        value = key.Key(kind, 182983)
+        prop._kind = kind  # Ick
```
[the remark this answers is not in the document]
chrisrossi: Shouldn't we do what datastore does and make `_name` and `_kind` of type `str` here, as well?

ndb/src/google/cloud/ndb/model.py
dhermes: @chrisrossi Did you resolve your

> Is this useful without a kind? Should it be required?

question? (I saw it in email but not on GitHub.)

ndb/src/google/cloud/ndb/model.py
dhermes: Using

```python
KeyProperty(None)
```

was billed as a way to explicitly set `name=None` in the original docstring. But that is already the default, so I'm not sure why it matters.

ndb/src/google/cloud/ndb/model.py
dhermes: Ha yup

ndb/tests/unit/test_model.py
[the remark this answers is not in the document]
dhermes: :+1: Will do and will make sure to document the problem in `MIGRATION_NOTES`

dhermes:
@chrisrossi PTAL

ndb/src/google/cloud/ndb/model.py
chrisrossi: I decided after looking at it more that it probably was useful and retracted my question.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Probably a backwards compatibility thing, then.

PR #6458, opened 2018-11-09
Implementing `StructuredProperty` in `ndb`.

dhermes:


ndb/src/google/cloud/ndb/model.py
```diff
@@ -2538,11 +2538,111 @@ def _now():
         return datetime.datetime.utcnow().time()
 
 
+def _structured_get_for_dict(property_, entity):
+    """Helper for ``_get_for_dict()``.
+
+    This is similar to a mix-in, for :class:`StructuredProperty` and
```
dhermes: The original [used a mix-in][1], but I just felt strange about adding the diamond in the MRO. Happy to re-instate if reviewers would like to.

[1]: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/cf4cab3f1f69cd04e1a9229871be466b53729f3f/ndb/model.py#L2238

ndb/tests/unit/test_model.py
```diff
@@ -1988,10 +1988,74 @@ def test__from_base_type(self):
...
+        )
+        nested_values = prop._get_for_dict(entity)
+        assert nested_values is None
+
+        # Cache is untouched.
+        assert model.Property._FIND_METHODS_CACHE == {}
+
+    def test__get_for_dict_single(self):
+        prop = model.StructuredProperty(self.Book, "book")
+        book_value = unittest.mock.Mock(spec=("_to_dict",))
```
dhermes: Rather than a mock, an instance of `Book` should be used here (and in related places)

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2538,11 +2538,111 @@ def _now():
...
+            repeated=repeated,
+            required=required,
+            default=default,
+            choices=choices,
+            validator=validator,
+            verbose_name=verbose_name,
+            write_empty_list=write_empty_list,
+        )
+        if self._repeated and modelclass._has_repeated:
+            raise TypeError(
```
dhermes: This wouldn't be necessary with the new API. However, the backend may still enforce it, but the protobuf does not (since it can just use a nested `Entity` as this property).

crwilcox:
Please port and reopen at https://github.com/googleapis/python-ndb

chrisrossi:
I think this one may be dead.

Issue #6363, opened 2018-11-01
NDB: Use gRPC


chrisrossi:
We'll have to sort out using new, modern gRPC instead of old RPC from app engine.  Putting a pin in that here for now.

chrisrossi:
If there's some issue housekeeping I need to do for this, let me know.

chrisrossi:
This is done. We're using gRPC now.

PR #6459, opened 2018-11-09
Ndb eventloop helpers

chrisrossi:
Implementations for the relatively trivial helper/wrapper functions that merely get the current event loop and then call the corresponding method on the loop.

### DIFF
--- a/ndb/src/google/cloud/ndb/_eventloop.py
+++ b/ndb/src/google/cloud/ndb/_eventloop.py
@@ -328,7 +328,7 @@ def get_event_loop():
     """Get the current event loop.
 
     This function should be called within a context established by
-    :func:`google.cloud.ndb.async_context`.
+    :func:`~google.cloud.ndb.async_context`.
 
     Returns:
         EventLoop: The event loop for the current
@@ -336,7 +336,7 @@ def get_event_loop():
 
     Raises:
         exceptions.AsyncContextError: If called outside of a context
-            established by :func:`google.cloud.ndb.async_context`.
+            established by :func:`~google.cloud.ndb.async_context`.
     """
     loop = contexts.current()
     if loop:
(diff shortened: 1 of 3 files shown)


ndb/src/google/cloud/ndb/_eventloop.py
```diff
@@ -345,25 +345,70 @@ def get_event_loop():
...
+    Raises:
+        exceptions.AsyncContextError: If called outside of a context
+            established by :func:`~google.cloud.ndb.async_context`.
+    """
+    loop = get_event_loop()
+    loop.queue_call(delay, callback, *args, **kwargs)
 
 
 def queue_rpc(*args, **kwargs):
     raise NotImplementedError
```
tseaver: Overlooked, or just not obvious how to implement it?

ndb/src/google/cloud/ndb/_eventloop.py
```diff
@@ -345,25 +345,70 @@ def get_event_loop():
...
+def add_idle(callback, *args, **kwargs):
+    """Helper function.
 
+    Calls :method:`EventLoop.add_idle` on current event loop.
 
-def queue_call(*args, **kwargs):
-    raise NotImplementedError
+    Raises:
+        exceptions.AsyncContextError: If called outside of a context
+            established by :func:`~google.cloud.ndb.async_context`.
```
tseaver: We've often not document "transitive" exceptions elsewhere in `google-cloud-*`.  @dhermes Is documenting them preferred in `ndb`?

ndb/src/google/cloud/ndb/_eventloop.py
chrisrossi: #6363

ndb/src/google/cloud/ndb/_eventloop.py
chrisrossi: I'm still learning the culture of this group--I'm happy to omit if that's consistent with prevailing style. Logically, to me, it made sense to document exceptions a caller is likely to encounter, whether or not they're raised in the body of the function itself. I don't have a strong feeling that it should be done one way or another, though.

dhermes: Yeah I'd say that's fine. Typically I avoid documenting transitive exceptions just because the true number of possible exceptions is massive (e.g. `MemoryError` or `KeyboardInterrupt` could occur) but when there are special exceptions worth pointing out I think it's perfectly acceptable.

ndb/src/google/cloud/ndb/_eventloop.py
```diff
@@ -345,25 +345,70 @@ def get_event_loop():
...
-    raise NotImplementedError
+def add_idle(callback, *args, **kwargs):
+    """Helper function.
 
+    Calls :method:`EventLoop.add_idle` on current event loop.
 
-def queue_call(*args, **kwargs):
-    raise NotImplementedError
+    Raises:
+        exceptions.AsyncContextError: If called outside of a context
```
dhermes: A Sphinx / napoleon tip: I'd just use `.AsyncContextError` here. It tells Sphinx to find the "nearest" relevant class named `AsyncContextError`.

ndb/src/google/cloud/ndb/_eventloop.py
```diff
@@ -345,25 +345,70 @@ def get_event_loop():
...
 
 
 def queue_rpc(*args, **kwargs):
     raise NotImplementedError
 
 
-def run(*args, **kwargs):
-    raise NotImplementedError
+def run():
+    """Helper function.
```
dhermes: It seems that there could be a better one-liner than "Helper function" (and that "Helper function" would apply to most functions / methods)

What about

```suggestion
    """Calls ``run()`` on the current event loop
```

as you mention below

ndb/src/google/cloud/ndb/_eventloop.py
chrisrossi: Ok

ndb/src/google/cloud/ndb/_eventloop.py
chrisrossi: Ok.

chrisrossi:
@tseaver @dhermes We good?

PR #6465, opened 2018-11-09
Require grpcio-gcp >= 0.2.2 directly

wyardley:
Older versions of pip don't process the extras correctly when passed to another dependency. Require grpcio-gcp explicitly instead.

See https://github.com/pypa/pip/issues/4957, GCP case #17453259, etc.

I don't seem to be having a problem with grpc itself, but could move that one too (and / or adjust version requirements) if you all want.

```
root@b3e5163e877e:/# pip install "google-api-core[grpcio-gcp]"
Collecting google-api-core[grpcio-gcp]
  Using cached https://files.pythonhosted.org/packages/48/53/2ec314fe5c167d150b6980f23765a8a8aa56861eb59e7e0fc45f499c62e3/google_api_core-1.5.1-py2.py3-none-any.whl
  google-api-core 1.5.1 does not provide the extra 'grpcio-gcp'
```

### DIFF
--- a/api_core/setup.py
+++ b/api_core/setup.py
@@ -36,11 +36,11 @@ dependencies = [
     'setuptools >= 34.0.0',
     'six >= 1.10.0',
     'pytz',
-    'futures >= 3.2.0; python_version < "3.2"'
+    'futures >= 3.2.0; python_version < "3.2"',
+    'grpcio-gcp': 'grpcio-gcp >= 0.2.2'
 ]
 extras = {
-    'grpc': 'grpcio >= 1.8.2',
-    'grpcio-gcp': 'grpcio-gcp >= 0.2.2'
+    'grpc': 'grpcio >= 1.8.2'
 }
 
 


tseaver:
I'm pretty reluctant to replicate dependencies out from core.  I know that Spanner is the only API currently using grpc-gcp.  @theacodes WDYT?

theacodes:
Yeah no, that gets weird fast. We *could* add the `grpc-grpc` extra to core, I think that would make much more sense.

wyardley:
@tseaver @theacodes Thanks.

Just wondering, is the plan for additional projects to use it, or would it make sense to actually take it _out_ of core and add it here?

Happy to rework this if there's a better way to accomplish it.

theacodes:
Eventually grpc-gcp will be used by several libraries. The spanner rollout
is just highest priority, as the client suffers the most performance-wise
without the extensions provided.

wyardley:
I think there's also a way to require >= a specific version of pip
Since pip 8 ships with several commonly used distros still (as default), though, my preference would be if you all can figure out a way to support it without that requirement.

If you're suggesting moving it out of extras in core, I'm happy to rework this to do that.

theacodes:
Core always needs grpc dependences to be in extras. We should just add a
new one that includes both grpc and gcp-grpc, and make spanner need that.

wyardley:
@theacodes Ah, got it. The way I reworked this probably won't work either.
I'll close this then, let me know if you need an issue (beyond the Google ticket above) or FR anywhere to help support this.

PR #6466, opened 2018-11-09
Add Region tags for a few spanner system tests

crwilcox:


crwilcox:
Merging as this is just labeling, nothing much to review.

PR #6468, opened 2018-11-09
remove hyphen from named extra in api_core

crwilcox:
In pip 8, which is currently used on ubuntu 16.04 LTS, extras with hyphens are ignored and not installed.

### DIFF
--- a/api_core/setup.py
+++ b/api_core/setup.py
@@ -40,7 +40,7 @@ dependencies = [
 ]
 extras = {
     'grpc': 'grpcio >= 1.8.2',
-    'grpcio-gcp': 'grpcio-gcp >= 0.2.2'
+    'grpcgcp': 'grpcio-gcp >= 0.2.2'
 }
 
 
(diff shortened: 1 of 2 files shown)


api_core/setup.py
```diff
@@ -40,7 +40,7 @@
 ]
 extras = {
     'grpc': 'grpcio >= 1.8.2',
-    'grpcio-gcp': 'grpcio-gcp >= 0.2.2'
+    'grpcgcp': 'grpcio-gcp >= 0.2.2'
```
crwilcox: I could optionally keep the old one around just in case anyone took a reliance on the name, but was avoiding cruft since the google-cloud-* packages should be the only packages depending on this, and spanner uses this only.

crwilcox: Also, *technically* the time between api_core and spanner going out this extra won't exist. I will tag/release at the same time so this should be a small timeframe

PR #6482, opened 2018-11-13
Synth Cloud scheduler

crwilcox:


scheduler/google/cloud/scheduler_v1beta1/gapic/transports/cloud_scheduler_grpc_transport.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        return google.api_core.grpc_helpers.create_channel(
+            address,
+            credentials=credentials,
+            scopes=cls._OAUTH_SCOPES,
+        )
+
+    @property
+    def list_jobs(self):
+        """Return the gRPC stub for {$apiMethod.name}.
```
tseaver: Is that intended?  Looks to me like a codegen bug.  OTOH, we have a bunch of those already in the codebase. :(

scheduler/google/cloud/scheduler_v1beta1/proto/cloudscheduler_pb2_grpc.py
```diff
@@ -0,0 +1,191 @@
...
+  """The Cloud Scheduler API allows external entities to reliably
+  schedule asynchronous jobs.
+  """
+
+  def ListJobs(self, request, context):
+    """Lists jobs.
+    """
+    context.set_code(grpc.StatusCode.UNIMPLEMENTED)
+    context.set_details('Method not implemented!')
+    raise NotImplementedError('Method not implemented!')
```
tseaver: I don't understand this bit either, but the pattern is common in the codebase.

scheduler/setup.py
```diff
@@ -0,0 +1,85 @@
...
+version = '0.1.0'
+# Should be one of:
+# 'Development Status :: 3 - Alpha'
+# 'Development Status :: 4 - Beta'
+# 'Development Status :: 5 - Production/Stable'
+release_status = 'Development Status :: 3 - Alpha'
+dependencies = [
+    'google-api-core[grpc] >= 1.4.1, < 2.0.0dev',
+    'enum34; python_version < "3.4"',
+#    'grpc-google-iam-v1 >= 0.11.4, < 0.12dev',
```
tseaver: Why is this one commented out?  Seems like it should be removed, if there is no IAM in this service.

crwilcox: Because I made an oops. It shouldn't be needed, I commented out to run tests and verify there wasn't an issue.

scheduler/google/cloud/scheduler_v1beta1/gapic/transports/cloud_scheduler_grpc_transport.py
crwilcox: This is an issue. I am going to ignore it now here as it is a broader issue in our generator. bug is here https://github.com/googleapis/gapic-generator/issues/2422

PR #6487, opened 2018-11-14
Implementing `ModelKey` for `ndb`.

dhermes:


ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+        Raises:
+            .BadValueError: If ``value`` is :data:`None`.
+        """
+        if value is not None:
+            return super(ModelKey, self)._comparison(op, value)
+
+        raise exceptions.BadValueError(
+            "__key__ filter query can't be compared to None"
+        )
+
```
dhermes: There was a

```python
# TODO: Support IN().
```

here, but we are intentionally not implementing these TODOs.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+    Returns:
+        .Key: The passed in ``value``.
+
+    Raises:
+        .BadValueError: If ``value`` is not a :class:`.Key`.
+        KindError: If ``entity`` is specified, by the kind of the entity
+            doesn't match the kind of ``value``.
+    """
+    if not isinstance(value, Key):
+        raise exceptions.BadValueError("Expected Key, got {!r}".format(value))
```
dhermes: There was a

```python
# TODO: BadKeyError.
```

here, but we are intentionally not implementing these TODOs.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+
+    if entity and entity.__class__ not in (Model, Expando):
```
chrisrossi: is x.__class__ preferable to type(x)?

chrisrossi: Also, I'm a little confused. Don't we expect `entity` to be an instance of `Model` or `Expando`?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+        value (.Key): The key to be validated.
+        entity (Optional[Model]): The entity that the key is being validated
+            for.
+
+    Returns:
+        .Key: The passed in ``value``.
+
+    Raises:
+        .BadValueError: If ``value`` is not a :class:`.Key`.
+        KindError: If ``entity`` is specified, by the kind of the entity
```
chrisrossi: s/by/but/

ndb/src/google/cloud/ndb/model.py
dhermes: I'm not sure if `.__class__` is preferable to `type()`, though they had a difference with old-style classes:

```python
>>> import sys
>>> sys.version
'2.7.15 (default, May 14 2018, 21:07:29) \n[GCC 5.4.0 20160609]'
>>> class A:
...     pass
...
>>> class B(object):
...     pass
...
>>> a = A()
>>> b = B()
>>> type(a)
<type 'instance'>
>>> type(b)
<class '__main__.B'>
>>> a.__class__
<class __main__.A at 0x7f928bf372c0>
>>> b.__class__
<class '__main__.B'>
```

dhermes: Re: `Model` / `Expando`, that is for cases where the class is literally one of those, i.e. not a subclass. We expect users to have

```python
class Book(ndb.Model):
    author = ndb.StringProperty()
    pages = ndb.IntegerProperty()
```

so that `book.__class__ is Book`, not `Model`.

chrisrossi:
Your attempt to give me credit for the typo fix seems to have upset clabot.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Re: Model / Expando

Yep, duh.

chrisrossi: In Python3 `type(o) is o.__class__`, afaict. I think `type(o)` is preferred to `o.__class__` for the same reason that `len(seq)` is preferred over `seq.__len__()`, but not a big deal if you have a style preference.

dhermes: [Argument][1] for `type()` over `.__class__` 

```python
>>> import sys
>>> sys.version
'3.7.1 (default, Oct 21 2018, 22:21:13) \n[GCC 7.3.0]'
>>> class A:
...     pass
...
>>> class B(A):
...     __class__ = A
...
>>> a = A()
>>> a.__class__
<class '__main__.A'>
>>> type(a)
<class '__main__.A'>
>>>
>>> b = B()
>>> b.__class__
<class '__main__.A'>
>>> type(b)
<class '__main__.B'>
```

[1]: https://stackoverflow.com/a/10633356/1068170

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+
+    Raises:
+        .BadValueError: If ``value`` is not a :class:`.Key`.
+        KindError: If ``entity`` is specified, but the kind of the entity
+            doesn't match the kind of ``value``.
+    """
+    if not isinstance(value, Key):
+        raise exceptions.BadValueError("Expected Key, got {!r}".format(value))
+
+    if entity and type(entity) not in (Model, Expando):
```
dhermes: @chrisrossi I switched to `type(entity)` here. Should we clean up all usage of `.__class__` in this package?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1542,11 +1543,123 @@ def _get_for_dict(self, entity):
...
+            )
+
+    return value
+
+
 class ModelKey(Property):
+    """Special property to store a special "key" for a :class:`Model`.
+
+    This is intended to be used as a psuedo-:class:`Property` on each
+    :class:`Model` subclass. It is **not** intended for other usage in
```
dhermes: I forgot to mention this originally. The **whole** purpose of this class is so that it can be used for `Model`:

```python
class Model:
    ...
    _key = ModelKey()
    ...
```

I wish we could make it non-public. WDYT? Should we have a list of "totally wish this wasn't public" members?

dhermes:
@chrisrossi `unittest.mock.Mock` instances can fake `__class__` but they can't fake `type()`:

```python
(Pdb) entity
<Mock spec='Model' id='140560813857312'>
(Pdb) isinstance(entity, model.Model)
True
(Pdb) entity.__class__
<class 'google.cloud.ndb.model.Model'>
(Pdb) type(entity)
<class 'unittest.mock.Mock'>
```

This makes it a bit harder to test here with mocks. Eventually, I won't need mocks, but that requires `Model` being done. I **could** skirt around this a bit by doing `model = object.__new__(Model)`.

WDYT?

ndb/src/google/cloud/ndb/model.py
chrisrossi: I spent some time, not entirely successfully, trying to figure out what this was used for. [These snippets](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/6f5f3bcb81779679a24e0964a6c57c0c7deabfac/appengine/standard/ndb/entities/snippets.py#L98) show a `key` attribute of an entity being used directly in client code, so maybe it's not private? Or am I reading this incorrectly?

ndb/src/google/cloud/ndb/model.py
chrisrossi: Yeah, probably.

ndb/src/google/cloud/ndb/model.py
dhermes: Yes, I should've written

```python
class Model:
    ...
    _key = ModelKey()
    key = _key
    ...
```

however the public attribute `key` can be overridden. The whole point of the class attribute is so that queries like this can be created:

```python
key = ndb.Key(Book, 88)
query = Book.query(Book.key >= key)
```

dhermes:
> Are the base classes ever instantiated directly, though, really?

For `Model`, I'd think not, but for `Expando` I think yes? Either way ISTM we are "required" to leave the original behavior here.

Since it isn't that difficult to do, I'll just use the `object.__new__(Model)` approach (instead of using a `Mock`) and then file an issue to follow-up and make the tests clean / not-nonsense when `Model` is done being implemented.

chrisrossi:
> > Are the base classes ever instantiated directly, though, really?
> 
> For `Model`, I'd think not, but for `Expando` I think yes? Either way ISTM we are "required" to leave the original behavior here.
> 
> Since it isn't that difficult to do, I'll just use the `object.__new__(Model)` approach (instead of using a `Mock`) and then file an issue to follow-up and make the tests clean / not-nonsense when `Model` is done being implemented.

:+1:

PR #6532, opened 2018-11-15
Fix documentation typo

di:


tseaver:
@di Thanks for the patch!

Issue #6533, opened 2018-11-15
Firestore:  support features blacklisted in conformance tests


tseaver:
PR #6290 blacklisted a number of conformance tests because we do not currently support the usecases they support:

- `get-*` tests (because we use `BatchGetDocuments` API rather than the `GetDocument` API.
- `listen-*` tests exercise the "watch" features (since landed in PR #6191).
- `update_paths-*` tests (they've been excluded forever, with a note that Python lacked the support).
- Tests involving unimplemented / incorrectly implemented "transforms" (`DELETE`, `ARRAY_REMOVE`, `ARRAY_UNION`).
- `query-*` tests have been (inadvertently) skipped since being copied in from `google-cloud-common` in PR #5351).

This is a tracking issue for removing those skips / blacklist entries:

- [x] Update `Document.get` to use the `GetDocument` API, and re-enable the (one) `get-basic.textproto` test. (#6534)
- [x] Fix the support for the `DELETE` transform (one failing test) and enable those tests. (#6559)
- [x] Figure out how to support the `ARRAY_REMOVE` transform, and enable those tests. (#6651)
- [x] Figure out how to support the `ARRAY_UNION` transform, and enable those tests. (#6651)
- [x] Enable the `query-*` tests and make changes needed for them to pass. (#6839)
- [ ] Enable the `listen-*` tests and make changes needed for them to pass.

Not in scope:

~~- Figure how to support the "update with paths" use case, and enable those tests.~~

tseaver:
Note:  by my read of the spec document, we do not need to support the `update-paths-*` cases:  they are for an alternative spelling of `Document.update` in languages which cannot use objects (such as `FieldPath` instances) as map keys.

PR #6534, opened 2018-11-15
Firestore: refactor 'Document.get' to use the 'GetDocument' API.

tseaver:
Update conformance test to actually run for 'get'.

Toward #6533.

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/document.py
+++ b/firestore/google/cloud/firestore_v1beta1/document.py
@@ -18,7 +18,9 @@ import copy
 
 import six
 
+from google.api_core import exceptions
 from google.cloud.firestore_v1beta1 import _helpers
+from google.cloud.firestore_v1beta1.proto import common_pb2
 from google.cloud.firestore_v1beta1.watch import Watch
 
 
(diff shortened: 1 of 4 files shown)


firestore/google/cloud/firestore_v1beta1/document.py
```diff
@@ -423,9 +428,38 @@ def get(self, field_paths=None, transaction=None):
...
+            data = _helpers.decode_dict(document_pb.fields, self._client)
+            exists = True
+            create_time = document_pb.create_time
+            update_time = document_pb.update_time
+
+        return DocumentSnapshot(
+            reference=self,
+            data=data,
+            exists=exists,
+            read_time=timestamp,  # No server read_time available
```
tseaver: The older "batch get" API call had an actual `read_time` on the response, but `GetDocument` just returns `Document`, which has no such field.  So, we compute our own timestamp and use that.

firestore/tests/unit/test_cross_language.py
```diff
@@ -180,19 +181,18 @@ def test_create_testprotos(test_proto):
...
-        firestore_api = _mock_firestore_api()
-        client, document = _make_client_document(firestore_api, testcase)
-        call = functools.partial(document.get, None, None)
-        _run_testcase(testcase, call, firestore_api, client)
+    firestore_api = mock.Mock(spec=['get_document'])
+    response = document_pb2.Document()
+    firestore_api.get_document.return_value = response
+    client, document = _make_client_document(firestore_api, testcase)
+
+    document.get()  # No '.textprotos' for errors, field_paths.
```
tseaver: As of today, the `.textproto` testcases in `google-cloud-common` contain only a [single file testing `Document.get`](https://github.com/googleapis/google-cloud-common/blob/03e4a2db3d229d4b97397e632ad831dd6a3a441f/testing/firestore/testdata/get-basic.textproto).  It does not pass any `field_paths`, nor is it marked as raising an error.

tseaver:
@crwilcox, @theacodes I still can't see the Kokoro test log, which means I can't diagnose anything.  System tests, unit tests, and lint are all passing on my local machine for ade1acf170be2126f08146d6c6050339d5a243dd.

tseaver:
@crwilcox, @ThatsAMorais CI is green after manually forcing it to rebuild via the `kokoro-force-run` label.  PTAL

tseaver:
@crwilcox, @JustinBeckwith, @theacodes Can one of you please paste the Kokoro error log into a gist?  My suspicion is that the failure is due to #6605.

firestore/google/cloud/firestore_v1beta1/document.py
mcdonc: I think that it should probably be None instead of a client-generated time.  Or just remove the field entirely from Document, with attendant bw compat issues.

mcdonc: DocumentSnapshot, not Document.

tseaver: OK.  I felt a little queasy about generating the timestamp myself.

tseaver: Ah, I remember now -- there is a system test which tries to use `read_time`:  I will have to adjust it.

PR #6581, opened 2018-11-19
Begin implementation of `ndb.Model`

dhermes:


ndb/src/google/cloud/ndb/model.py
```diff
@@ -2047,6 +2047,21 @@ def __init__(self, *args, **kwargs):
 
         super(TextProperty, self).__init__(*args, **kwargs)
 
+    def _constructor_info(self):
```
dhermes: This "snuck" in because I realized that `repr()` was broken for `TextProperty` and `StringProperty` when testing `MetaModel.__repr__`

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+            @classmethod
+            def _get_kind(cls):
+                return "AnotherKind"
+
+    A newly constructed entity will not be persisted to Cloud Datastore without
+    an explicit call to :meth:`put`.
+
+    .. note::
+
+        You cannot define a property named ``key``; the ``.key`` attribute
```
dhermes: AFAICT this is not explicitly enforced by the code

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+
+    Raises:
+        .BadArgumentError: If the constructor is called with ``key`` and one
+            of ``id``, ``app``, ``namespace`` or ``parent`` specified.
+    """
+
+    # Class variables updated by _fix_up_properties()
+    _properties = None
+    _has_repeated = False
+    _kind_map = {}  # Dict mapping {kind: Model subclass}
```
dhermes: IMO this kind map should be gardened on `MetaModel` and reads and writes to it should be threadsafe. But it's probably not worth the yak-shaving / bikeshedding since it will (**should**) only be modified at code import time.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
         return cls.__name__
 
+    def _set_projection(self, projection):
+        """Set the projected properties for this instance.
+
+        Args:
+            projection (Union[list, tuple]): An iterable of strings
+                representing the projection for the model instance.
+        """
+        self._projection = tuple(projection)
```
dhermes: This implementation is partial, but I noted in #6251 that this is the case and the original full implementation is in the history for this PR (I deleted it in the 3rd commit).

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+            manually after dynamically updating a model class.
+
+        Raises:
+            KindError: If the returned kind from ``_get_kind()`` is not a
+                :class:`str`.
+            TypeError: If a property on this model has a name beginning with
+                an underscore.
+        """
+        kind = cls._get_kind()
+        if not isinstance(kind, str):
```
dhermes: Previously this did a `basestring` check, but I think restricting to `str` (i.e. not allowing `bytes`) is the right move in Python 3. @chrisrossi shall I put a comment in `MIGRATION_NOTES.md` or nah?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+                attr, ModelKey
+            ):
+                if name.startswith("_"):
+                    raise TypeError(
+                        "ModelAttribute {} cannot begin with an underscore "
+                        "character. ``_`` prefixed attributes are reserved "
+                        "for temporary Model instance values.".format(name)
+                    )
+                attr._fix_up(cls, name)
+                if isinstance(attr, Property):
```
dhermes: I am very unclear on why this branch is here. The `isinstance(attr, ModelAttribute)` check above seems to match some callers that are subclasses of `ModelAttribute`, but not `Property` subclasses. Such types don't seem to exist anywhere so I have no idea who / what is being supported by this.

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,117 @@ def test_constructor():
...
+            second = model.StringProperty()
+
+        expected = (
+            "Mine<first=IntegerProperty('first'), "
+            "second=StringProperty('second')>"
+        )
+        assert repr(Mine) == expected
+
+    @staticmethod
+    def test_bad_kind():
```
dhermes: ISTM that this should be in `TestMetaModel` since the behavior of `MetaModel` invokes `Model._fix_up_properties` to actually cause this failure.

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,117 @@ def test_constructor():
...
+        with pytest.raises(model.KindError):
+
+            class Mine(model.Model):
+                @classmethod
+                def _get_kind(cls):
+                    return 525600
+
+    @staticmethod
+    def test_invalid_property_name():
```
dhermes: ISTM that this should be in `TestMetaModel` since the behavior of `MetaModel` invokes `Model._fix_up_properties` to actually cause this failure.

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,117 @@ def test_constructor():
...
+        with pytest.raises(TypeError):
+
+            class Mine(model.Model):
+                _foo = model.StringProperty()
+
+    @staticmethod
+    def test_repeated_property():
```
dhermes: ISTM that this should be in `TestMetaModel` since the behavior of `MetaModel` invokes `Model._fix_up_properties` to actually cause this behavior.

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,117 @@ def test_constructor():
...
+        class Mine(model.Model):
+            foo = model.StringProperty(repeated=True)
+
+        assert Mine._has_repeated
+
+    @staticmethod
+    def test_non_property_attribute():
```
dhermes: ISTM that this should be in `TestMetaModel` since the behavior of `MetaModel` invokes `Model._fix_up_properties` to actually cause this behavior.

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,117 @@ def test_constructor():
...
 
 
 class TestModel:
     @staticmethod
-    def test_constructor():
-        with pytest.raises(NotImplementedError):
-            model.Model()
+    def test_constructor_defaults():
+        entity = model.Model()
+        assert entity.__dict__ == {"_values": {}}
```
dhermes: @chrisrossi Any better idea than testing `__dict__`. The whole "fallback to class defaults" pattern is why I went with this, but I'm happy to change it up.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
 
-class Model:
-    __slots__ = ("_entity_key",)
+        class Book(ndb.Model):
+            pages = ndb.IntegerProperty()
 
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    the ``Book.pages`` property doesn't have the name ``pages`` assigned.
+    This is accomplished by calling the ``_fix_properties()`` method on the
```
chrisrossi: s/_fix_properties/_fix_up_properties/

ndb/src/google/cloud/ndb/model.py
chrisrossi: I guess it's enforced by it just not working? Explicit enforcement with an informative exception raised might be better than whatever strange, broken behavior occurs naturally, although I don't know what the natural consequence is.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Yeah, I'm fine with it as-is.

ndb/src/google/cloud/ndb/model.py
chrisrossi: I don't think it's strictly necessary. You can if you want.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+            )
+
+        # Map of ``kind_name`` -> ``property_instance``.
+        cls._properties = {}
+
+        # Skip the classes in ``ndb.model``.
+        if cls.__module__ == __name__:
+            return
+
+        for name in set(dir(cls)):
```
chrisrossi: I'm not sure why `set` is needed here. Is there a condition where `dir` would return the same value twice?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2962,286 @@ def __init__(self, *args, **kwargs):
...
+                an underscore.
+        """
+        kind = cls._get_kind()
+        if not isinstance(kind, str):
+            raise KindError(
+                "Class {} defines a ``_get_kind()`` method that returns "
+                "a non-string ({!r})".format(cls.__name__, kind)
+            )
+
+        # Map of ``kind_name`` -> ``property_instance``.
```
chrisrossi: This seems to be mapping attribute name to property instance, afaict.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Maybe just an overabundance of caution? One solution might be to move what little is defined in ModelAttribute and move it to Property, if, as it seems, Property is the sole subclass of ModelAttribute.

Or leave it as-is. Either way.

ndb/tests/unit/test_model.py
chrisrossi: I can see arguments for either way. Fine with whatever you decide.

ndb/tests/unit/test_model.py
chrisrossi: This is fine.

ndb/src/google/cloud/ndb/model.py
dhermes: Good call, comments are code smell anyhow.

ndb/src/google/cloud/ndb/model.py
dhermes: I just ran this in a live app sandbox:

```python
from google.appengine.ext import ndb
class MyModel(ndb.Model):
    key = ndb.StringProperty()

m = MyModel(key="hi")
print(m)
m.put()
print(m)
```

It results in

```
MyModel(key='hi')
MyModel(key=Key('MyModel', 5690894162329600), key=u'hi')
```

ndb/src/google/cloud/ndb/model.py
dhermes: I think we (for now) have to leave it because `ModelAttribute` is public.

ndb/src/google/cloud/ndb/model.py
chrisrossi: That's nonsense. Awesome.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Ok.

ndb/src/google/cloud/ndb/model.py
chrisrossi: So, yes, if you intentionally break it. ;-)

dhermes: Yeah so the two questions are:

- Is `_fix_up` and related idempotent for a given property? (`_fix_up` is, but user-defined versions of `_fix_up` may not be?)
- Should we ditch `set()`?

ndb/src/google/cloud/ndb/model.py
dhermes: Yeah so I think the fix is to just remove the note?

ndb/src/google/cloud/ndb/model.py
dhermes: It's "essentially" idempotent, so we can ditch `set()` IMO. WDYT?

ndb/src/google/cloud/ndb/model.py
chrisrossi: Is the above behavior not considered pathological? At best, it's at least confusing.

ndb/src/google/cloud/ndb/model.py
chrisrossi: I agree.

ndb/src/google/cloud/ndb/model.py
dhermes: I did a big revamp of this docstring with some doctests. I need to add tests (and a "fix") for `Model.__repr__` but I think it's much better. It may be worth rendering locally for you to get a picture?

dhermes:
@chrisrossi PTAL.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2967,405 @@ def __init__(self, *args, **kwargs):
...
+    property value. Instead, the entity's key is accessible via ``_key``:
+
+    .. doctest:: model-keyword-key-collision
+
+        >>> class KeyCollide(ndb.Model):
+        ...     key = ndb.StringProperty()
+        ...
+        >>> entity1 = KeyCollide(key="Take fork in road", id=987)
+        >>> entity1
+        KeyCollide(_key=Key('KeyCollide', 987), key='Take fork in road')
```
dhermes: @chrisrossi In the existing implementation the behavior is slightly different:

```python
from google.appengine.ext import ndb


class KeyCollide(ndb.Model):
    key = ndb.StringProperty()


entity1 = KeyCollide(key="Take fork in road", id=987)
print(entity1)
entity2 = KeyCollide(key="Go slow", _key=ndb.Key(KeyCollide, 1))
print(entity2)
entity3 = KeyCollide(key="Zap")
print(entity3)
entity4 = KeyCollide(id="Baz")
print(entity4)
entity5 = KeyCollide()
print(entity5)
```

yields

```python
KeyCollide(key=Key('KeyCollide', 987), key='Take fork in road')
KeyCollide(key=Key('KeyCollide', 1), key='Go slow')
KeyCollide(key='Zap')
KeyCollide(key=Key('KeyCollide', 'Baz'))
KeyCollide()
```

ndb/tests/unit/test_model.py
```diff
@@ -2283,16 +2295,163 @@ def test_constructor():
...
+    @staticmethod
+    def test_constructor_non_property():
+        class TimeTravelVehicle(model.Model):
+            speed = 88
+
+        with pytest.raises(TypeError):
+            TimeTravelVehicle(speed=28)
+
+    @staticmethod
+    def test_repr():
```
dhermes: @chrisrossi I don't like that I jammed 5 actual tests into a single test, but I did it somewhat to avoid having to redefine `ManyFields` in each test. Do you think I should split this up?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -2947,28 +2967,405 @@ def __init__(self, *args, **kwargs):
...
+        parent (Key): The parent model or :data:`None` for a top-level model.
+            If ``parent`` is used, ``key`` must be :data:`None`.
+        namespace (str): Namespace for the entity key.
+        app (str): Application ID for the entity key.
+        kwargs (Dict[str, Any]): Additional keyword arguments. These should map
+            to properties of this model.
+
+    Raises:
+        .BadArgumentError: If the constructor is called with ``key`` and one
+            of ``id``, ``app``, ``namespace`` or ``parent`` specified.
```
chrisrossi: Mention TypeError if kwarg isn't a property, or no?

ndb/tests/unit/test_model.py
chrisrossi: I would split it up. I tend to do a lot of copy/paste for tests.  I guess you could also declare ManyFields at module scope, if you wanted to avoid repetition.

ndb/src/google/cloud/ndb/model.py
dhermes: It'd be an `AttributeError` I think. Did you have a code snippet in mind?

ndb/tests/unit/test_model.py
dhermes: Done.

ndb/src/google/cloud/ndb/model.py
dhermes: Here is what happens against the original implementation:

```python
from google.appengine.ext import ndb


class A(ndb.Model):
    b = ndb.IntegerProperty()


print(A)
print(A(b=11))
print(A(c=12))
```

yields

```
A<b=IntegerProperty('b')>
A(b=11)
  ...
  File ".../google/appengine/ext/ndb/model.py", line 2947, in __init__
    self._set_attributes(kwds)
  File ".../google/appengine/ext/ndb/model.py", line 2990, in _set_attributes
    prop = getattr(cls, name)  # Raises AttributeError for unknown properties.
AttributeError: type object 'A' has no attribute 'c'
```

chrisrossi: I was just referencing this:

https://github.com/googleapis/google-cloud-python/blob/639234a9294375158522c23c3795c976ceb78b75/ndb/src/google/cloud/ndb/model.py#L3254

Which throws TypeError. AttributeError is probably better, though.

dhermes: Gotcha, that's this instead:

```python
from google.appengine.ext import ndb


class A(ndb.Model):
    b = ndb.IntegerProperty()
    c = 98909


print(A)
print(A(b=11))
print(A(c=12))
```

yields

```
A<b=IntegerProperty('b')>
A(b=11)
  ...
  File ".../google/appengine/ext/ndb/model.py", line 2947, in __init__
    self._set_attributes(kwds)
  File ".../google/appengine/ext/ndb/model.py", line 2992, in _set_attributes
    raise TypeError('Cannot set non-property %s' % name)
TypeError: Cannot set non-property c
```

(This is in a unit test, but wanted to add it for completeness.)

I can add doctests to the constructor to avoid staleness?

chrisrossi: Ah ok.

ndb/src/google/cloud/ndb/model.py
chrisrossi: Your fix is fine. repr() should theoretically return text that if it were python would instantiate an equivalent object. The old implementation failed that test.

dhermes: :+1:

ndb/src/google/cloud/ndb/model.py
dhermes: Added doctest.

PR #6606, opened 2018-11-20
add templating to asset synth.py

crwilcox:


busunkim96:
It looks like flake and black are in disagreement over some things.

busunkim96:
We could reference Black's .flake8 configuration and update the template's .flake8 to match: https://github.com/ambv/black/blob/master/.flake8

crwilcox:
I think the issue is I didn't commit .flake8. Adding now.

asset/noxfile.py
```diff
@@ -20,29 +20,116 @@
...
+    # Sanity check: only run tests if found.
+    if not os.path.exists(system_test_path):
+        session.skip("System tests were not found")
+
+    # Use pre-release gRPC for system tests.
+    session.install("--pre", "grpcio")
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install("mock", "pytest")
```
tseaver: Hmm, shouldn't need `mock` for system tests.

PR #6695, opened 2018-11-29
Implement comparison helpers for `ndb.Model`

dhermes:
~~**NOTE**: This has #6694 as diffbase.~~

I have removed some features of comparison that depend on `StructuredProperty` or `Expando` being implemented. I added a note to #6251 to go back and add the behavior when relevant.

### DIFF
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -3676,6 +3676,91 @@ class Model(metaclass=MetaModel):
         """
         return cls.__name__
 
+    def __hash__(self):
+        """Not implemented hash function.
+
+        Raises:
+            TypeError: Always, to emphasize that entities are mutable.
+        """
+        raise TypeError("Model is not immutable")
+
+    def __eq__(self, other):
+        """Compare two entities of the same class for equality."""
+        if type(other) is not type(self):
+            return NotImplemented
+
+        if self._key != other._key:
+            return False
+
+        return self._equivalent(other)
+
+    def _equivalent(self, other):
+        """Compare two entities of the same class, excluding keys.
+
+        Args:
+            other (Model): An entity of the same class. It is assumed that
+                the type and the key of ``other`` match the current entity's
+                type and key (and the caller is responsible for checking).
+
+        Returns:
+            bool: Indicating if the current entity and ``other`` are
+            equivalent.
+
+        Raises:
+            NotImplementedError: If the type's don't match.
+        """
+        if type(other) is not type(self):
+            raise NotImplementedError(
+                "Cannot compare different model classes. {} is "
+                "not {}".format(type(self).__name__, type(other).__name__)
+            )
+
+        if set(self._projection) != set(other._projection):
+            return False
+
+        # Short-circuit equality check, can only happen for ``Expando``.
+        if len(self._properties) != len(other._properties):
+            return False
+
+        prop_names = set(self._properties.keys())
+        other_prop_names = set(other._properties.keys())
+        if prop_names != other_prop_names:
+            return False  # Can only happen for ``Expando``.
+
+        # Restrict properties to the projection if set.
+        if self._projection:
+            prop_names = set(self._projection)
+
+        for name in prop_names:
+            if "." in name:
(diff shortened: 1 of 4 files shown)


ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3676,80 @@ class a different name when stored in Google Cloud Datastore than the
         """
         return cls.__name__
 
+    def __hash__(self):
+        """Not implemented hash function.
+
+        Raises:
+            TypeError: Always, to emphasize that entities are mutable.
+        """
+        raise TypeError("Model is not immutable")
```
chrisrossi: Double negative.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3676,80 @@ class a different name when stored in Google Cloud Datastore than the
...
+            prop_names = set(self._projection)
+
+        for name in prop_names:
+            value = self._properties[name]._get_value(self)
+            if value != other._properties[name]._get_value(other):
+                return False
+
+        return True
+
+    def __ne__(self, other):
```
chrisrossi: I *think* this isn't strictly necessary. Python can negate __eq__ without an explicit __neq__.

~~~~
Python 3.7.0 (default, Oct 15 2018, 12:13:42) 
[GCC 5.4.0 20160609] on linux
>>> class A:
...   def __eq__(self, other):
...       return other == 5
... 
>>> a = A()
>>> a == 5
True
>>> a == 6
False
>>> a != 6
True
>>> a != 5
False
>>> 
~~~~

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3676,80 @@ class a different name when stored in Google Cloud Datastore than the
...
+                type and key (and the caller is responsible for checking).
+
+        Returns:
+            bool: Indicating if the current entity and ``other`` are
+            equivalent.
+
+        Raises:
+            NotImplementedError: If the type's don't match.
+        """
+        if type(other) is not type(self):
```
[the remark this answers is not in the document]
dhermes: 1. I agree
1. I was preserving the original implementation
1. I don't think it's called from elsewhere?

ndb/src/google/cloud/ndb/model.py
dhermes: Sort of, but don't you think it's the "right" concept. I.e. `__hash__` can only be defined if "Model is immutable"

ndb/src/google/cloud/ndb/model.py
dhermes: From https://docs.python.org/3/reference/datamodel.html:

> By default, `__ne__()` delegates to `__eq__()` and inverts the result unless it is `NotImplemented`. There are no other implied relationships among the comparison operators, for example, the truth of `(x<y or x==y)` does not imply `x<=y`. To automatically generate ordering operations from a single root operation, see `functools.total_ordering()`.

chrisrossi: Interesting. I wasn't 100% sure, which is why I tried it in the REPL and pasted the result above. I didn't try in 2.7, because we don't need it.

dhermes: OK I removed all `__ne__` except for one that was doing something other than inverting `__eq__`

ndb/src/google/cloud/ndb/model.py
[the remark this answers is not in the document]
dhermes: @chrisrossi What to do about this?

dhermes:
@chrisrossi PTAL

ndb/src/google/cloud/ndb/model.py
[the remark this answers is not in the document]
chrisrossi: I would get rid of it. Every line of code deleted decreases long-term maintenance burden. Unless you think there's some outside chance the code is useful, in which case I guess it's ok to leave in defensively.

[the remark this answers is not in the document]
dhermes: OK, I got rid of it.

PR #6697, opened 2018-11-29
Firestore: make cursor-related 'Query' methods accept lists

### LINKED ISSUE #6664: Firestore: 'Query.start_at' etc. should accept a list of values for 'document_fields'.
Currently, the "cursor-related" methods ([`Query.start_at`](https://github.com/googleapis/google-cloud-python/blob/4c8184a1023b3c68633c50fdd53fe5136883dd2a/firestore/google/cloud/firestore_v1beta1/query.py#L407-L411), [`Query.start_after`](https://github.com/googleapis/google-cloud-python/blob/4c8184a1023b3c68633c50fdd53fe5136883dd2a/firestore/google/cloud/firestore_v1beta1/query.py#L436-L440), [`Query.end_before`](https://github.com/googleapis/google-cloud-python/blob/4c8184a1023b3c68633c50fdd53fe5136883dd2a/firestore/google/cloud/firestore_v1beta1/query.py#L464-L469), [`Query.end_at`](https://github.com/googleapis/google-cloud-python/blob/4c8184a1023b3c68633c50fdd53fe5136883dd2a/firestore/google/cloud/firestore_v1beta1/query.py#L494-L498)) expect their `document_fields` argument to be either a `DocumentSnapshot` instance or a `dict`, where the fields in either are used to compute the actual values for the cursor, using the query's `order_by` fields.

However, the cross-language conformance tests also expect to be able to pass a raw sequence of values for `document_fields`, which are then used to match the query's `order_by`.  We need to update the docstrings for these methods to indicate that a list is allowed, and fix the [`query._cursor_pb` helper](https://github.com/googleapis/google-cloud-python/blob/4c8184a1023b3c68633c50fdd53fe5136883dd2a/firestore/google/cloud/firestore_v1beta1/query.py#L778-L825) to deal with that case.

The corresponding `Collection` methods also need their docstrings adjusted (they just forward arguments through to a new query object).

tseaver:
Closes #6664.

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/query.py
+++ b/firestore/google/cloud/firestore_v1beta1/query.py
@@ -56,6 +56,8 @@ _NO_ORDERS_FOR_CURSOR = (
     'When defining a cursor with one of ``start_at()`` / ``start_after()`` / '
     '``end_before()`` / ``end_at()``, all fields in the cursor must '
     'come from fields set in ``order_by()``.')
+_MISMATCH_CURSOR_W_ORDER_BY = (
+    'The cursor {!r} does not match the order fields {!r}.')
 _EMPTY_DOC_TEMPLATE = (
     'Unexpected server response. All responses other than the first must '
     'contain a document. The response at index {} was\n{}.')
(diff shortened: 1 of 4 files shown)


firestore/google/cloud/firestore_v1beta1/query.py
```diff
@@ -530,13 +533,48 @@ def _filters_pb(self):
...
+
+        order_keys = [order.field.field_path for order in orders]
+
+        if isinstance(document_fields, dict):
+            # Transform to list using orders
+            values = []
+            data = document_fields
+            for order_key in order_keys:
+                try:
+                    values.append(_helpers.get_nested_value(order_key, data))
```
crwilcox: May be worth reformatting as the following as append is likely more expensive than a listcomp.
```
try:
    values = [_helpers.get_nested_value(order_key, data) for order_key in order_keys]
except:
```

PR #6703, opened 2018-11-29
Firestore: prevent use of transforms as values passed to 'Query.where'.

### LINKED ISSUE #6699: Firestore: 'Query.where' needs to raise for sentinels / transforms
The following conformance tests assert that attempting to use `DELETE_FIELD`, `SERVER_TIMESTAMP`, `ArrayRemove`, or `ArrayUnion` as the value of a `Query.where` must raise an exception, which the current implementation does not do:

- `query-arrayremove-where.textproto`
- `query-arrayunion-where.textproto`
- `query-del-where.textproto`
- `query-st-where.textproto`

tseaver:
Closes #6699.


tseaver:
The Kokoro failure is in an unrelated flaky systest ("Quota exceeded").

tseaver:
@crwilcox This one should also be ready for review.

PR #6733, opened 2018-11-30
Configure Kokoro for releases for bigquery_storage.

tswast:
CL 223438946 pending internally to enable Kokoro.

@busunkim96 Since I already tagged the release, does this mean we'll have to submit the package manually?


busunkim96:
@tswast Once both the CL and PR are merged we should be able to manually trigger the release job in fusion.

PR #6738, opened 2018-11-30
Deserialize an entity from a protobuffer.

chrisrossi:
Deserialization is largely delegated to Datastore, so it's really just a matter of adapting an NDB entity to a Datastore entity.

### DIFF
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -130,11 +130,11 @@ Datastore API, rather than the legacy App Engine Datastore. In general, for
 users coding to the public interface, this won't be an issue, but users relying
 on pieces of the ostensibly private API that are exposed to the bare metal of
 the original datastore implementation will have to rewrite those pieces.
-Specifically, any function or method that dealt directly with protocol buffers will
-no longer work. Datastore, itself, has moved to gRPC and protobuffers
-themselves are significantly different from legacy Datastore.  Additionally,
-this version of NDB makes less direct use of protobuffers and relies more
-heavily on Datastore for parsing and generating protobuffers, which is a
+Specifically, any function or method that dealt directly with protocol buffers
+will no longer work. The Datastore `.protobuf` definitions have changed
+significantly from the barely public API used by App Engine to the current
+published API.  Additionally, this version of NDB mostly delegates to
+`google.cloud.datastore` for parsing data returned by RPCs, which is a
 significant internal refactoring.
 
 - `ModelAdapter` is no longer used. In legacy NDB, this was passed to Datastore
@@ -142,10 +142,11 @@ significant internal refactoring.
   Datastore RPC calls. AFAIK, Datastore no longer accepts an adapter for
   adpating entities. At any rate, we no longer do it that way.
 - `Property._db_get_value` is no longer used. It worked directly with Datastore
-  protobuffers, work which is now delegated to Datastore.
-- `Model._deserialize` is no longer used. It worked directly with protobuffers,
-  so wasn't really salvageable. Unfortunately, there were comments indicating
-  it was overridden by subclasses. Hopefully this isn't broadly the case.
+  protocol buffers, work which is now delegated to Datastore.
+- `Model._deserialize` is no longer used. It worked directly with protocol
+  buffers, so wasn't really salvageable. Unfortunately, there were comments
+  indicating it was overridden by subclasses. Hopefully this isn't broadly the
+  case.
 
 ## Comments
 
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -128,9 +128,9 @@ The primary differences come from:
 One of the largest classes of differences comes from the use of the current
 Datastore API, rather than the legacy App Engine Datastore. In general, for
 users coding to the public interface, this won't be an issue, but users relying
-on pieces of the onstensibly private API that are exposed to the bare metal of
+on pieces of the ostensibly private API that are exposed to the bare metal of
 the original datastore implementation will have to rewrite those pieces.
-Specifically, any function or method that dealt directly with protobuffers will
+Specifically, any function or method that dealt directly with protocol buffers will
 no longer work. Datastore, itself, has moved to gRPC and protobuffers
 themselves are significantly different from legacy Datastore.  Additionally,
 this version of NDB makes less direct use of protobuffers and relies more
@@ -145,7 +145,7 @@ significant internal refactoring.
   protobuffers, work which is now delegated to Datastore.
 - `Model._deserialize` is no longer used. It worked directly with protobuffers,
   so wasn't really salvageable. Unfortunately, there were comments indicating
-  was overridden by subclasses. Hopefully this isn't broadly the case.
+  it was overridden by subclasses. Hopefully this isn't broadly the case.
 
 ## Comments
 
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -123,6 +123,30 @@ The primary differences come from:
   is also a user-defined property named `key`. For an example, see the
   class docstring for `Model`.
 
+## Bare Metal
+
+One of the largest classes of differences comes from the use of the current
+Datastore API, rather than the legacy App Engine Datastore. In general, for
+users coding to the public interface, this won't be an issue, but users relying
+on pieces of the onstensibly private API that are exposed to the bare metal of
+the original datastore implementation will have to rewrite those pieces.
+Specifically, any function or method that dealt directly with protobuffers will
+no longer work. Datastore, itself, has moved to gRPC and protobuffers
+themselves are significantly different from legacy Datastore.  Additionally,
+this version of NDB makes less direct use of protobuffers and relies more
+heavily on Datastore for parsing and generating protobuffers, which is a
+significant internal refactoring.
+
+- `ModelAdapter` is no longer used. In legacy NDB, this was passed to Datastore
+  so that calls to Datastore RPCs could yield NDB entities directly from
+  Datastore RPC calls. AFAIK, Datastore no longer accepts an adapter for
+  adpating entities. At any rate, we no longer do it that way.
+- `Property._db_get_value` is no longer used. It worked directly with Datastore
+  protobuffers, work which is now delegated to Datastore.
+- `Model._deserialize` is no longer used. It worked directly with protobuffers,
+  so wasn't really salvageable. Unfortunately, there were comments indicating
+  was overridden by subclasses. Hopefully this isn't broadly the case.
+
 ## Comments
 
 - There is rampant use (and abuse) of `__new__` rather than `__init__` as
--- a/ndb/MIGRATION_NOTES.md
+++ b/ndb/MIGRATION_NOTES.md
@@ -133,16 +133,16 @@ the original datastore implementation will have to rewrite those pieces.
 Specifically, any function or method that dealt directly with protocol buffers
 will no longer work. The Datastore `.protobuf` definitions have changed
 significantly from the barely public API used by App Engine to the current
-published API.  Additionally, this version of NDB mostly delegates to
+published API. Additionally, this version of NDB mostly delegates to
 `google.cloud.datastore` for parsing data returned by RPCs, which is a
 significant internal refactoring.
 
-- `ModelAdapter` is no longer used. In legacy NDB, this was passed to Datastore
-  so that calls to Datastore RPCs could yield NDB entities directly from
-  Datastore RPC calls. AFAIK, Datastore no longer accepts an adapter for
-  adpating entities. At any rate, we no longer do it that way.
+- `ModelAdapter` is no longer used. In legacy NDB, this was passed to the 
+  Datastore RPC client so that calls to Datastore RPCs could yield NDB entities
+  directly from Datastore RPC calls. AFAIK, Datastore no longer accepts an
+  adapter for adpating entities. At any rate, we no longer do it that way.
 - `Property._db_get_value` is no longer used. It worked directly with Datastore
-  protocol buffers, work which is now delegated to Datastore.
+  protocol buffers, work which is now delegated to `google.cloud.datastore`.
 - `Model._deserialize` is no longer used. It worked directly with protocol
   buffers, so wasn't really salvageable. Unfortunately, there were comments
   indicating it was overridden by subclasses. Hopefully this isn't broadly the
--- a/ndb/src/google/cloud/ndb/__init__.py
+++ b/ndb/src/google/cloud/ndb/__init__.py
@@ -63,6 +63,7 @@ __all__ = [
     "make_connection",
     "MetaModel",
     "Model",
+    "ModelAdapter",
     "ModelAttribute",
     "ModelKey",
     "ndb_context",
@@ -161,6 +162,7 @@ from google.cloud.ndb.model import LocalStructuredProperty
 from google.cloud.ndb.model import make_connection
 from google.cloud.ndb.model import MetaModel
 from google.cloud.ndb.model import Model
+from google.cloud.ndb.model import ModelAdapter
 from google.cloud.ndb.model import ModelAttribute
 from google.cloud.ndb.model import ModelKey
 from google.cloud.ndb.model import non_transactional
--- a/ndb/src/google/cloud/ndb/__init__.py
+++ b/ndb/src/google/cloud/ndb/__init__.py
@@ -63,7 +63,6 @@ __all__ = [
     "make_connection",
     "MetaModel",
     "Model",
-    "ModelAdapter",
     "ModelAttribute",
     "ModelKey",
     "ndb_context",
@@ -162,7 +161,6 @@ from google.cloud.ndb.model import LocalStructuredProperty
 from google.cloud.ndb.model import make_connection
 from google.cloud.ndb.model import MetaModel
 from google.cloud.ndb.model import Model
-from google.cloud.ndb.model import ModelAdapter
 from google.cloud.ndb.model import ModelAttribute
 from google.cloud.ndb.model import ModelKey
 from google.cloud.ndb.model import non_transactional
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -304,7 +304,7 @@ def _entity_from_protobuf(protobuf):
             continue
         if value is not None:
             if prop._repeated:
-                value = [_BaseValue(x) for x in value]
+                value = [_BaseValue(sub_value) for sub_value in value]
             else:
                 value = _BaseValue(value)
         prop._store_value(entity, value)
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -295,7 +295,7 @@ def _entity_from_protobuf(protobuf):
             continue
         if value is not None:
             if prop._repeated:
-                value = list(map(_BaseValue, value))
+                value = [_BaseValue(x) for x in value]
             else:
                 value = _BaseValue(value)
         prop._store_value(entity, value)
--- a/ndb/src/google/cloud/ndb/model.py
+++ b/ndb/src/google/cloud/ndb/model.py
@@ -50,7 +50,6 @@ __all__ = [
     "IndexProperty",
     "Index",
     "IndexState",
-    "ModelAdapter",
     "make_connection",
     "ModelAttribute",
     "Property",
@@ -276,11 +275,32 @@ class IndexState:
         return hash((self.definition, self.state, self.id))
 
 
-class ModelAdapter:
-    __slots__ = ()
+def _entity_from_protobuf(protobuf):
+    """Deserialize an entity from a protobuffer.
 
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    Args:
+        protobuf (google.cloud.datastore_v1.proto.entity.Entity): An
+            entity protobuf to be deserialized.
+
+    Returns:
+        .Model: The deserialized entity.
+    """
+    ds_entity = helpers.entity_from_protobuf(protobuf)
+    model_class = Model._lookup_model(ds_entity.kind)
+    entity = model_class()
+    entity.key = key_module.Key._from_ds_key(ds_entity.key)
+    for name, value in ds_entity.items():
+        prop = getattr(model_class, name, None)
+        if not (prop and isinstance(prop, Property)):
+            continue
+        if value is not None:
+            if prop._repeated:
+                value = list(map(_BaseValue, value))
+            else:
+                value = _BaseValue(value)
+        prop._store_value(entity, value)
+
+    return entity
 
 
 def make_connection(*args, **kwargs):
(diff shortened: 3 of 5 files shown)


ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
   is also a user-defined property named `key`. For an example, see the
   class docstring for `Model`.
 
+## Bare Metal
+
+One of the largest classes of differences comes from the use of the current
+Datastore API, rather than the legacy App Engine Datastore. In general, for
+users coding to the public interface, this won't be an issue, but users relying
+on pieces of the onstensibly private API that are exposed to the bare metal of
```
dhermes:
```suggestion
on pieces of the ostensibly private API that are exposed to the bare metal of
```

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
...
+the original datastore implementation will have to rewrite those pieces.
+Specifically, any function or method that dealt directly with protobuffers will
```
dhermes:
```suggestion
Specifically, any function or method that dealt directly with protocol buffers will
```

I don't think "protobuffers" is a thing?

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
...
+no longer work. Datastore, itself, has moved to gRPC and protobuffers
+themselves are significantly different from legacy Datastore.  Additionally,
```
dhermes: I would rewrite this to just saw the `.protobuf` definitions have changed quite a bit from the barely-public API used by App Engine to the [public one][1] (also drop all usage of "protobuffers").

[1]: https://github.com/googleapis/googleapis/tree/master/google/datastore/v1

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
...
+this version of NDB makes less direct use of protobuffers and relies more
+heavily on Datastore for parsing and generating protobuffers, which is a
```
dhermes: ISTM you mean `google-cloud-datastore` rather than the actual backend service "Datastore" here.

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
...
+
+- `ModelAdapter` is no longer used. In legacy NDB, this was passed to Datastore
+  so that calls to Datastore RPCs could yield NDB entities directly from
+  Datastore RPC calls. AFAIK, Datastore no longer accepts an adapter for
+  adpating entities. At any rate, we no longer do it that way.
+- `Property._db_get_value` is no longer used. It worked directly with Datastore
+  protobuffers, work which is now delegated to Datastore.
+- `Model._deserialize` is no longer used. It worked directly with protobuffers,
+  so wasn't really salvageable. Unfortunately, there were comments indicating
+  was overridden by subclasses. Hopefully this isn't broadly the case.
```
dhermes:
```suggestion
  it was overridden by subclasses. Hopefully this isn't broadly the case.
```

ndb/src/google/cloud/ndb/__init__.py
```diff
@@ -63,7 +63,6 @@
     "make_connection",
     "MetaModel",
     "Model",
-    "ModelAdapter",
```
dhermes: I would prefer we keep it and replace it by something like `NotImplemented`? If we targeted `>= 3.7` we could use `__getattr__` on the module and just raise an exception on access.

https://www.python.org/dev/peps/pep-0562/

ndb/src/google/cloud/ndb/model.py
```diff
@@ -288,11 +287,32 @@ def __hash__(self):
...
+    Args:
+        protobuf (google.cloud.datastore_v1.proto.entity.Entity): An
+            entity protobuf to be deserialized.
+
+    Returns:
+        .Model: The deserialized entity.
+    """
+    ds_entity = helpers.entity_from_protobuf(protobuf)
+    model_class = Model._lookup_model(ds_entity.kind)
+    entity = model_class()
```
dhermes: Do we want this? I.e. would calling `__new__` / bypassing `__init__` make more sense? What did the original code do (send a link?)?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -288,11 +287,32 @@ def __hash__(self):
...
+    entity.key = key_module.Key._from_ds_key(ds_entity.key)
```
dhermes: You need to use the private `_key`?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -288,11 +287,32 @@ def __hash__(self):
...
+    for name, value in ds_entity.items():
+        prop = getattr(model_class, name, None)
+        if not (prop and isinstance(prop, Property)):
```
dhermes:
```suggestion
        if not (prop is not None and isinstance(prop, Property)):
```

ndb/src/google/cloud/ndb/model.py
```diff
@@ -288,11 +287,32 @@ def __hash__(self):
...
+            continue
+        if value is not None:
+            if prop._repeated:
+                value = list(map(_BaseValue, value))
```
dhermes: `list(map(...))` not so fun? Use a list comprehension instead?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -1472,23 +1492,6 @@ def _serialize(
         """
         raise NotImplementedError
 
-    def _deserialize(self, entity, p, unused_depth=1):
```
dhermes: ISTM it'll be easier to track "completion" of the implementation by leaving methods like this as `raise NotImplementedError` and then marking them for deletion at a later time

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3607,32 @@ class a different name when stored in Google Cloud Datastore than the
         """
         return cls.__name__
 
+    @classmethod
+    def _lookup_model(cls, kind, default_model=None):
+        """Get the model class for  the given kind.
```
dhermes:
```suggestion
        """Get the model class for the given kind.
```

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3607,32 @@ class a different name when stored in Google Cloud Datastore than the
...
+
+        Args:
+            kind (str): The name of the kind to look up.
+            default_model: The model class to return if the kind can't be
```
dhermes:
```suggestion
            default_model (Optional[type]): The model class to return if the kind can't be
```

May have 80-char issues (GitHub suggestion doesn't support multi-line yet)

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3607,32 @@ class a different name when stored in Google Cloud Datastore than the
...
+                found.
+
+        Returns:
+            class: The model class for the requested kind or the default model.
```
dhermes:
```suggestion
            type: The model class for the requested kind or the default model.
```

ndb/src/google/cloud/ndb/model.py
```diff
@@ -3676,6 +3607,32 @@ class a different name when stored in Google Cloud Datastore than the
...
+
+        Raises:
+            .KindError: If the kind was not found and no default_model was
```
dhermes:
```suggestion
            .KindError: If the kind was not found and no ``default_model`` was
```

ndb/tests/unit/test_model.py
```diff
@@ -2722,6 +2656,55 @@ def test__put():
...
             entity._put()
 
+    @staticmethod
+    def test__lookup_model():
+        class ThisKind(model.Model):
+            pass
+
+        sentinel = object()
+        assert model.Model._lookup_model("ThisKind")
+        assert model.Model._lookup_model("NoKind", sentinel) is sentinel
```
dhermes: This test isn't actually deterministic. Some other code could load in a class called `NoKind` and then it would be in the kind map on `Model`. We unfortunately need something like `conftest.property_clean_cache()` to "reset" the internal state of `Model` to empty / clean.

ndb/tests/unit/test_model.py
```diff
@@ -2722,6 +2656,55 @@ def test__put():
...
+
+        with pytest.raises(model.KindError):
+            model.Model._lookup_model("NoKind")
```
dhermes: Should this unit test not be split into 3?

ndb/tests/unit/test_model.py
```diff
@@ -2722,6 +2656,55 @@ def test__put():
...
+
+    dill = {"sandwiches": ["turkey", "reuben"], "not_sandwiches": "tacos"}
+    gherkin = [{"a": {"b": "c"}, "d": 0}, [1, 2, 3], "himom"]
+    key = datastore.Key("ThisKind", 123, project="testing")
+    datastore_entity = datastore.Entity(key=key)
+    datastore_entity.update(
+        {
+            "a": 42,
+            "b": None,
+            "c": pickle.dumps(gherkin, pickle.HIGHEST_PROTOCOL),
```
dhermes: Why the use of `pickle.HIGHEST_PROTOCOL`? Isn't it the default?

ndb/tests/unit/test_model.py
```diff
@@ -2722,6 +2656,55 @@ def test__put():
...
+        }
+    )
+    protobuf = helpers.entity_to_protobuf(datastore_entity)
+    entity = model._entity_from_protobuf(protobuf)
+    assert type(entity) is ThisKind
+    assert entity.a == 42
+    assert entity.b is None
+    assert entity.c == gherkin
+    assert entity.d == ["foo", "bar", "baz"]
+    assert entity.e == [gherkin, dill]
```
dhermes: What about `notaproperty` check?

ndb/tests/unit/test_model.py
```diff
@@ -2722,6 +2656,55 @@ def test__put():
...
+                pickle.dumps(gherkin, pickle.HIGHEST_PROTOCOL),
+                pickle.dumps(dill, pickle.HIGHEST_PROTOCOL),
+            ],
+            "notused": 32,
+            "notaproperty": None,
+        }
+    )
+    protobuf = helpers.entity_to_protobuf(datastore_entity)
+    entity = model._entity_from_protobuf(protobuf)
+    assert type(entity) is ThisKind
```
dhermes:
```suggestion
    assert isinstance(entity, ThisKind)
```

ndb/src/google/cloud/ndb/model.py
chrisrossi: https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/master/ndb/model.py#L3194

ndb/src/google/cloud/ndb/model.py
chrisrossi: Won't a property always be truthy?

ndb/src/google/cloud/ndb/model.py
chrisrossi: We're already using NotImplementedError to signal API we haven't gotten to yet. How do you propose distinguishing between API we haven't gotten to yet and API we don't intend to implement?

ndb/tests/unit/test_model.py
chrisrossi: The default is `pickle.DEFAULT_PROTOCOL`. Sometimes they're the same, sometimes they're not, depending on Python version. `PickleProperty` explicitly uses `pickle.HIGHEST_PROTOCOL`, so this makes sure it's getting what it expects.

chrisrossi:
~~~~
Python 3.5.2 (default, Nov 12 2018, 13:43:14) 
[GCC 5.4.0 20160609] on linux
>>> import pickle
>>> pickle.DEFAULT_PROTOCOL
3
>>> pickle.HIGHEST_PROTOCOL
4
~~~~

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,30 @@ The primary differences come from:
...
+this version of NDB makes less direct use of protobuffers and relies more
+heavily on Datastore for parsing and generating protobuffers, which is a
+significant internal refactoring.
+
+- `ModelAdapter` is no longer used. In legacy NDB, this was passed to Datastore
+  so that calls to Datastore RPCs could yield NDB entities directly from
+  Datastore RPC calls. AFAIK, Datastore no longer accepts an adapter for
+  adpating entities. At any rate, we no longer do it that way.
+- `Property._db_get_value` is no longer used. It worked directly with Datastore
+  protobuffers, work which is now delegated to Datastore.
```
chrisrossi: @dhermes You mention above dropping use of "protobuffers"--how should I refer to the `p` argument of the new deprecated `Property._db_get_value`?

ndb/src/google/cloud/ndb/__init__.py
chrisrossi: Currently we're targeting >= 3.5, it seems. What do our customers require?

ndb/src/google/cloud/ndb/model.py
chrisrossi: I have in other persistence systems used `model_class.__new__(model_class)` in this spot, so I'm fine with changing it. I was just blindly copying from the source implementation here.

ndb/src/google/cloud/ndb/model.py
chrisrossi: 08e3659

ndb/src/google/cloud/ndb/model.py
chrisrossi: 3609595

ndb/tests/unit/test_model.py
chrisrossi: 5f75adf

ndb/src/google/cloud/ndb/model.py
chrisrossi: 7e49c1c

ndb/tests/unit/test_model.py
chrisrossi: 2b1f9b5

ndb/MIGRATION_NOTES.md
dhermes: If it's deprecated, do we need to document it?

ndb/src/google/cloud/ndb/model.py
dhermes: Not sure? The explicit `None`-check is in the Google Python style guide as being "good practice".

ndb/src/google/cloud/ndb/model.py
```diff
@@ -300,14 +300,14 @@ def _entity_from_protobuf(protobuf):
...
-    entity.key = key_module.Key._from_ds_key(ds_entity.key)
+    entity._key = key_module.Key._from_ds_key(ds_entity.key)
     for name, value in ds_entity.items():
         prop = getattr(model_class, name, None)
         if not (prop and isinstance(prop, Property)):
             continue
         if value is not None:
             if prop._repeated:
-                value = list(map(_BaseValue, value))
+                value = [_BaseValue(x) for x in value]
```
dhermes: `x` -> `sub_value`? (One letter variable names frowned upon by style guides?)

ndb/src/google/cloud/ndb/model.py
dhermes: Add a message to the `NotImplementedError`, as is done in e.g. in [`Key.from_old_key()`][1]

[1]: https://github.com/googleapis/google-cloud-python/blob/b69461c2e6c856783ca5b9b3eaafdc56d8fcc6db/ndb/src/google/cloud/ndb/key.py#L778

ndb/MIGRATION_NOTES.md
dhermes: But also, you can just say "protocol buffer" every time you were using "protobuffer".

chrisrossi: I think it's similar status to a lot of stuff in this file: probably not an issue but worth mentioning in case someone was overriding `Property._db_get_value` for some ill-advised reason. I get the impression that best practices for NDB evolved over time. There's an element of archaeology to this work.

dhermes: :+1:

ndb/MIGRATION_NOTES.md
chrisrossi: ab2f965

ndb/MIGRATION_NOTES.md
chrisrossi: ab2f965

ndb/MIGRATION_NOTES.md
chrisrossi: ab2f965

ndb/src/google/cloud/ndb/__init__.py
chrisrossi: a8d884a

ndb/tests/unit/test_model.py
```diff
@@ -2656,15 +2656,23 @@ def test__put():
         with pytest.raises(NotImplementedError):
             entity._put()
 
+    @pytest.mark.usefixtures("property_clean_cache")
```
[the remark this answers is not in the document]
chrisrossi: I was reading too fast. Added a fixture explicitly for this case.

I prefer `@pytest.mark.usefixtures()` when the fixture isn't used in the body of the test but is only called for its side effects. If you see an argument to a function you generally want to see it be used in the function.

ndb/tests/unit/test_model.py
```diff
@@ -2698,12 +2706,45 @@ class ThisKind(model.Model):
...
     entity = model._entity_from_protobuf(protobuf)
-    assert type(entity) is ThisKind
+    assert isinstance(entity, ThisKind)
     assert entity.a == 42
     assert entity.b is None
     assert entity.c == gherkin
     assert entity.d == ["foo", "bar", "baz"]
     assert entity.e == [gherkin, dill]
+    assert entity._key.kind() == "ThisKind"
+    assert entity._key.id() == 123
```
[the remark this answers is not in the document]
chrisrossi: Because the mismatch between `app` and `project` was skeeving me out, because I don't really understand it, and this way I could work around not understanding that. But I've gone ahead and changed it. 

I'm not sure why datastore has "project" and ndb has "app", though. I'm guessing historical reasons.

chrisrossi:
@dhermes I think I've addressed everything. PTAL.

ndb/tests/unit/test_model.py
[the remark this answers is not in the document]
chrisrossi: So we should probably migrate "app" to "project" as part of the port, then, right?

ndb/MIGRATION_NOTES.md
```diff
@@ -123,6 +123,31 @@ The primary differences come from:
...
+will no longer work. The Datastore `.protobuf` definitions have changed
+significantly from the barely public API used by App Engine to the current
+published API. Additionally, this version of NDB mostly delegates to
+`google.cloud.datastore` for parsing data returned by RPCs, which is a
+significant internal refactoring.
+
+- `ModelAdapter` is no longer used. In legacy NDB, this was passed to the 
+  Datastore RPC client so that calls to Datastore RPCs could yield NDB entities
+  directly from Datastore RPC calls. AFAIK, Datastore no longer accepts an
+  adapter for adpating entities. At any rate, we no longer do it that way.
```
dhermes:
```suggestion
  adapter for adapting entities. At any rate, we no longer do it that way.
```

ndb/tests/unit/test_model.py
```diff
@@ -2653,12 +2655,14 @@ class TimeTravelVehicle(model.Model):
             TimeTravelVehicle(speed=28)
 
     @staticmethod
+    @pytest.mark.usefixtures("reset_kind_map")
```
dhermes: Sorry I'm just checking these, but this test doesn't actually create a `Model` subclass, so why is the kind map reset?

ndb/tests/conftest.py
```diff
@@ -36,3 +36,19 @@ def property_clean_cache():
...
+def reset_kind_map():
+    """Reset ``Model._kind_map``.
+
+    This mapping of "kind" to class is set whenever a new subclass of ``Model``
+    is created. We create ``Model`` subclasses in tests and don't want those
+    definitions to leak to other tests, so this fixture resets the mapping to
+    its value before the text. (Since some classes might be declared at module
+    scope, we can't just clear the mapping altogether.)
+    """
+    previous = model.Model._kind_map
```
dhermes: ISTM it may be easier to reason about if we keep `Model._kind_map == {}` as a requirement. This will help us catch tests which are polluting the hash. So the global `ManyFields` will probably have to move to a helper function or method which returns that type

Or is there a way to add a hook in `pytest` for **every** test (i.e. so we don't have to explicitly use the fixture).

ndb/tests/unit/test_model.py
chrisrossi: Copy pasta from splitting the tests up.

ndb/tests/conftest.py
chrisrossi: Yeah, we can certainly move ManyFields or use a factory function.

It is also possible to declare `reset_kind_map` with `autouse=True`, and it will get used by every test in our suite. While that might seem a little indiscriminate, it's a very cheap operation computationally, so I don't really have a problem with that approach.

chrisrossi: If we do it for `reset_kind_map` I propose we also do it for `property_clear_cache` or consolidate them to an omnibus `reset_all_the_things` fixture.

dhermes: :+1: On `autouse=True` and on the omnibus.

dhermes:
@chrisrossi Were you planning on adding a factory for the `ManyFields` class? Are there other classes that are defined globally?

ndb/tests/conftest.py
```diff
@@ -23,16 +23,17 @@
...
+    - ``model.Model._kind_map``
     """
-    assert model.Property._FIND_METHODS_CACHE == {}
-    try:
-        yield
-    finally:
-        assert model.Property._FIND_METHODS_CACHE != {}
-        model.Property._FIND_METHODS_CACHE.clear()
+    model.Property._FIND_METHODS_CACHE.clear()
+    model.Model._kind_map.clear()
```
dhermes: Why this vs.

```python
assert model.Property._FIND_METHODS_CACHE == {}
assert model.Model._kind_map == {}
yield
model.Property._FIND_METHODS_CACHE.clear()
model.Model._kind_map.clear()
```

If you had that it would presumably have caught the fact that `ManyFields` modified `model.Model._kind_map` **outside** of a test.

chrisrossi: I simplified the reset fixture because if it's going to be run on every single test, we don't really need to concern ourselves with the state before it gets called--we just need to make the state what we want. The more complex checking made more sense when we were selectively applying the fixture, as it would let us know if there was a test that needed the fixture that hadn't gotten it. I didn't add a factory on the ManyFields class because none of the tests in which it's used rely on the `model.Model._kind_map`, so it doesn't make any practical difference.

dhermes: I worry that we'll later introduce a test that uses `ManyFields` and will assume `"ManyFields"` is a key in `model.Model._kind_map`. I'm advocating for the `== {}` approach (along with using a class factory for `ManyFields`) to mitigate against future situations where failures seem unrelated to the test code being added.

chrisrossi: Ok. 31b1548e6cb3d263805df483e29c0fcc2547376f

dhermes:
LGTM feel free to merge

PR #6769, opened 2018-11-30
Release api_core 1.6.0

crwilcox:

### DIFF
--- a/api_core/CHANGELOG.md
+++ b/api_core/CHANGELOG.md
@@ -4,6 +4,26 @@
 
 [1]: https://pypi.org/project/google-api-core/#history
 
+## 1.6.0
+
+11-30-2018 12:45 PST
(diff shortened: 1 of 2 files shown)


tseaver:
CI failure is for `nox -e pytype`.

rchen152:
@crwilcox It looks like what's happening is that pytype is somehow finding multiple files named path_template.py (and timeout.py and grpc_helpers.py - the error probably differs because it varies which file pytype discovers first). The other thing I noticed is that the file count in the status message ("Analyzing 53 sources with 0 dependencies") differs from the one I see (28 sources). Any idea why there might be two copies of some source files?

crwilcox:
@rchen152 , I rm -rf'd the output dir (pytype_output) between runs so that is probably related?

PR #6770, opened 2018-11-30
Firestore: reject invalid paths passed to 'Query.{select,where,order_by}'

tseaver:


firestore/google/cloud/firestore_v1beta1/_helpers.py
```diff
@@ -482,6 +484,70 @@ def get_field_path(field_names):
     return FIELD_PATH_DELIMITER.join(result)
 
 
+PATH_ELEMENT_TOKENS = [
+    ('SIMPLE', r'[_a-zA-Z][_a-zA-Z0-9]*'),  # unquoted elements
```
crwilcox: this regex is the unanchored version of the one above. Can they be combined?

PR #6741, opened 2018-11-30
Use moved iam.policy now at google.api_core.iam.policy

crwilcox:
diffbase: #6740
Affects bigtable and storage

PR #6775, opened 2018-11-30
remove iam module from core

crwilcox:
iam was moved to API Core.


crwilcox:
CircleCI is failing for missing module. That is getting fixed in https://github.com/googleapis/google-cloud-python/pull/6741

docs/core/iam.rst
```diff
@@ -1,6 +0,0 @@
-Identity and Access Management
-==============================
-
-.. automodule:: google.cloud.iam
```
tseaver: Should just be updated to point to `google.api_core.iam`.

docs/core/index.rst
```diff
@@ -10,7 +10,6 @@ Core
     retry
     timeout
     page_iterator
-    iam
```
tseaver: This one needs to stay.

PR #6822, opened 2018-12-01
Lint NDB on CI

crwilcox:


ndb/tests/unit/test_model.py
```diff
@@ -1370,7 +1370,7 @@ def test_compare_valid():
     def test_compare_invalid():
         prop = model.ModelKey()
         with pytest.raises(exceptions.BadValueError):
-            prop == None
+            prop == None  # noqa: E711
```
dhermes: What is `E711`? Is there a way I can send edits to this PR to avoid using comment for this

ndb/noxfile.py
```diff
@@ -68,6 +68,25 @@ def cover(session):
...
 
+@nox.session(py=DEFAULT_INTERPRETER)
+def lint(session):
+    """Run linters.
+    Returns a failure if the linters find linting errors or sufficiently
+    serious code quality issues.
+    """
+    session.install("flake8", "black")
+    session.run(
+        "black",
```
dhermes: Seems like this is duplicating a bit of the code in `def blacken()`?

ndb/tests/unit/test_utils.py
```diff
@@ -17,7 +17,6 @@
 import pytest
 
 from google.cloud.ndb import utils
-import tests.unit.utils
```
dhermes: Lint win!

ndb/tests/unit/test_model.py
crwilcox: It thinks this should read `prop is None`.

ndb/noxfile.py
crwilcox: This checks that you blacked. Since CI doesn't apply the formatting changes.

dhermes: I realize, but I'm saying we could avoid repeating ourselves a bit by having a helper that takes `session` as an argument and an optional flag that turns on `--check`? E.g. so the file paths don't end up out of sync

ndb/tests/unit/test_model.py
dhermes:
```suggestion
            prop == unittest.mock.sentinel.other
```

ndb/tests/unit/test_query.py
```diff
@@ -51,13 +51,13 @@ class TestParameterizedThing:
     def test___eq__():
         thing = query.ParameterizedThing()
         with pytest.raises(NotImplementedError):
-            thing == None
+            thing == None  # noqa: E711
```
dhermes:
```suggestion
            thing == unittest.mock.sentinel.other
```

ndb/tests/unit/test_query.py
```diff
@@ -51,13 +51,13 @@ class TestParameterizedThing:
...
 
     @staticmethod
     def test___ne__():
         thing = query.ParameterizedThing()
         with pytest.raises(NotImplementedError):
-            thing != None
+            thing != None  # noqa: E711
```
dhermes:
```suggestion
            thing != unittest.mock.sentinel.other
```

ndb/tests/unit/test_query.py
```diff
@@ -143,12 +143,12 @@ def _make_one():
     def test___eq__(self):
         node = self._make_one()
         with pytest.raises(NotImplementedError):
-            node == None
+            node == None  # noqa: E711
```
dhermes:
```suggestion
            node == unittest.mock.sentinel.other
```

ndb/tests/unit/test_query.py
```diff
@@ -143,12 +143,12 @@ def _make_one():
...
 
     def test___ne__(self):
         node = self._make_one()
         with pytest.raises(NotImplementedError):
-            node != None
+            node != None  # noqa: E711
```
dhermes:
```suggestion
            node != unittest.mock.sentinel.other
```

ndb/noxfile.py
crwilcox: Now I understand. Yeah. we could certainly do that. This is being served as a template in the other libraries (as part of googleapis/synthtool) and would need to be updated there. This library has its own noxfile though. If you want to update it here I am happy to propagate that change to the others.

dhermes: > This library has its own noxfile though

Should we just get rid of that feature here?

> If you want to update it here I am happy to propagate that change to the others.

I think having `nox -s blacken` is useful, especially for cases when devs don't have editor support built in.

PR #6825, opened 2018-12-02
Spanner: add 'operation_id' parameter 'Database.update_ddl' (#6737)

### LINKED ISSUE #6737: Spanner: Operation_id missing in Spanner's Database.update_ddl method
Library: google_cloud_spanner v. 1.6.1

We seem to miss operation_id in spanner_v1 of database.update_ddl(). 

It looks like the Spanner HTTP API supports operation_id: https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases/updateDdl
GoogleAPI docs mention that Spanner's database.update_ddl() supports operation_id:
https://googleapis.github.io/google-cloud-python/latest/spanner/database-usage.html#update-an-existing-database.

Wy use (from pypi) google_cloud_spanner v. 1.6.1: This seems to be the latest released version (Released 10 Nov 2018). In the code of the library's Database object - contrary to the documentation there is no operation_id parameter in update_ddl statement and the operation_id seems to be hard-coded to “” in the actual API call:

https://github.com/googleapis/google-cloud-python/blob/master/spanner/google/cloud/spanner_v1/database.py#L257 

Is it intended? How else can we use operation_id

potiuk:
The operation_id im update_ddl method is missing. The parameter is in the documentation https://googleapis.github.io/google-cloud-python/latest/spanner/database-usage.html#update-an-existing-database but it was missing in the implementation. 

I've run system tests manually - even if the update_ddl system tests are skipped  due to flakyness (I removed the skip annotations and increased timeouts) so I confirm it works :).

Closes #6737.

### DIFF
--- a/spanner/google/cloud/spanner_v1/database.py
+++ b/spanner/google/cloud/spanner_v1/database.py
@@ -254,7 +254,7 @@ class Database(object):
         response = api.get_database_ddl(self.name, metadata=metadata)
         self._ddl_statements = tuple(response.statements)
 
-    def update_ddl(self, ddl_statements):
+    def update_ddl(self, ddl_statements, operation_id=''):
         """Update DDL for this database.
 
         Apply any configured schema from :attr:`ddl_statements`.
@@ -264,6 +264,8 @@ class Database(object):
 
         :type ddl_statements: Sequence[str]
         :param ddl_statements: a list of DDL statements to use on this database
+        :type operation_id: str
+        :param operation_id: (optional) a string ID for the long-running operation
 
         :rtype: :class:`google.api_core.operation.Operation`
         :returns: an operation instance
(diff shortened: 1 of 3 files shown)


spanner/google/cloud/spanner_v1/database.py
```diff
@@ -264,6 +264,8 @@ def update_ddl(self, ddl_statements):
 
         :type ddl_statements: Sequence[str]
         :param ddl_statements: a list of DDL statements to use on this database
+        :type operation_id: str
+        :param operation_id: a string ID for the long-running operation
```
tseaver: Please add `(optional)` at the start of the description.

potiuk: Done.

potiuk:
@tseaver: I added the unit test (I have not noticed that those exist for update_ddl in the first place). But I also left the skipped test there, as I found it really useful to test it with real instance. It simply makes it easier - especially that I tried to generate the id based on docs of Spanner and it turned out that the id cannot be too long. Having an example working ID generation in test might be a good hint for someone trying to do the same (if they start to look at the source code).

tseaver:
@potiuk Thank you again!

PR #6827, opened 2018-12-03
BigQuery:  add avro logical type control for load jobs.

shollyman:


bigquery/google/cloud/bigquery/job.py
```diff
@@ -1362,6 +1374,13 @@ def time_partitioning(self):
         """
         return self._configuration.time_partitioning
 
+    @property
+    def use_avro_logical_types(self):
+        """See
+        :attr:`google.cloud.bigquery.job.LoadJobConfig.use_avro_logical_types`.
+        """
+        return self._configuration.use_avro_logical_types
```
tseaver: We need a unit test for this getter, too.

adamcunnington-2sixty:
When will this flag be added to export jobs? Interestingly, export jobs so seem to do this anyway - or at least partially, i.e. TIMESTAMP fields are correctly exported as LONG with local types of timestamp-milllis but DATE fields are still exported as string without a logical type of DATE which is a major issue. I'm hoping this is the feature that I can get whitelisted for - I asked for that in the related Google issue thread.

PR #6840, opened 2018-12-04
Add to_bqstorage to convert from Table[Reference] google-cloud-bigquery-storage reference

tswast:
This makes it easier to use the new BigQuery Storage API (currently in
Alpha) in combination with the BigQuery API.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -279,7 +279,7 @@ class TableReference(object):
             "tableId": self._table_id,
         }
 
-    def to_bqstorage(self, selected_fields=None):
+    def to_bqstorage(self):
         """Construct a BigQuery Storage API representation of this table.
 
         If the ``table_id`` contains a partition identifier (e.g.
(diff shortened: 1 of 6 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -279,6 +279,53 @@ def to_api_repr(self):
...
+        table_ref = bigquery_storage_v1beta1.types.TableReference()
+        table_ref.project_id = self._project
+        table_ref.dataset_id = self._dataset_id
+
+        modifiers = bigquery_storage_v1beta1.types.TableModifiers()
+        read_options = bigquery_storage_v1beta1.types.TableReadOptions()
+        table_id = self._table_id
+        partition = None
+
+        if "@" in table_id:
```
shollyman: Potential issue here:  table@t1-t2 is the range format, and table@t1 is the snapshot format.  Perhaps a more strict check to ensure you're working with a supported snapshot format?

https://cloud.google.com/bigquery/table-decorators

bigquery/google/cloud/bigquery/table.py
```diff
@@ -279,6 +279,53 @@ def to_api_repr(self):
...
+            table_id, snapshot_time = table_id.split("@")
+            snapshot_time = int(snapshot_time)
+            modifiers.snapshot_time.FromMilliseconds(snapshot_time)
+
+        if "$" in table_id:
```
shollyman: Not sure we can assume the $<suffix> is a time partitioned table, with range partitioning on the doorstep.

tswast: Would someone direct a load job at a specific range partition?

bigquery/google/cloud/bigquery/table.py
tswast: Would `table@t1-t2` work for `tabledata.list`? Would we expect anyone to use the range format for table operations?

tswast: FYI: I've updated the sample to just strip these out, since they aren't relevant to the actual BigQuery Storage API TableReference. Added a comment in the function docs recommending using the TableModifiers manually.

PR #6856, opened 2018-12-06
Fix trace client memory leak

c24t:
cc @liyanhui1228 

@bogdandrutu and I are seeing a persistent [memory leak](https://github.com/census-instrumentation/opencensus-python/issues/334) in projects using the [opencensus stackdriver exporter](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/stats/exporters/stackdriver_exporter.py), which uses `google.cloud.trace.client.Client` and repeatedly calls [`batch_write_spans`](https://github.com/googleapis/google-cloud-python/blob/92465cbc4d9c0ba251838e9cd17f61d14b470e04/trace/google/cloud/trace/client.py#L57-L100).

Reusing the trace client's API object fixes the leak.

### DIFF
--- a/trace/tests/unit/v2/test_client_v2.py
+++ b/trace/tests/unit/v2/test_client_v2.py
@@ -62,8 +62,7 @@ class TestClient(unittest.TestCase):
 
     def test_trace_api_existing(self):
         """Check that the client caches _trace_api."""
-        client = self._make_one(project=self.project,
-                                credentials=_make_credentials())
+        client = self._make_one(project=self.project, credentials=_make_credentials())
         client._trace_api = mock.sentinel.trace_api
         self.assertIs(client.trace_api, mock.sentinel.trace_api)
 
--- a/trace/tests/unit/v2/test_client_v2.py
+++ b/trace/tests/unit/v2/test_client_v2.py
@@ -60,6 +60,15 @@ class TestClient(unittest.TestCase):
         self.assertIs(api, api_obj)
         self.assertEqual(clients, [client])
 
+    def test_trace_api_existing(self):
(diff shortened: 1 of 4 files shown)


bogdandrutu:
@crwilcox please help us to submit this and do a patch release if possible. Thank you

ocervell:
@c24t Thanks for this ! This might actually fix some issues we've been running into where 'too many open files' exceptions crash our VM. Those start occurring because we keep creating new clients / new connections.

bogdandrutu:
When we debug this issue we saw also socket objects being leaked, so I think it is a correlation.

c24t:
Thanks @tseaver, updated in 57a3da49f6.

trace/tests/unit/v2/test_client_v2.py
```diff
@@ -60,6 +60,15 @@ def make_api(client_obj):
         self.assertIs(api, api_obj)
         self.assertEqual(clients, [client])
 
+    def test_trace_api_existing(self):
+        """Check that the client caches _trace_api."""
+        client = self._make_one(project=self.project,
+                                credentials=_make_credentials())
+        self.assertIsNone(client._trace_api)
```
tseaver: The point of the '_existing' testcase is to assign a known object (e.g., `mock.sentinel.trace_api`) to `client._trace_api` here, and then assert that `client.trace_api` returns that object, rather than creating a new one.

c24t: That's what L70 does.

tseaver:
Thanks for bearing with me, @c24t!

PR #6857, opened 2018-12-06
Update index.rst

nnegrey:


speech/docs/index.rst
```diff
@@ -19,16 +19,13 @@ See: `Speech Asynchronous Recognize`_
...
-    ...     ),
-    ...     config=speech.types.RecognitionConfig(
-    ...         encoding='LINEAR16',
-    ...         language_code='en-US',
-    ...         sample_rate_hertz=44100,
-    ...     ),
-    ... )
+    >>> audio = speech.types.RecognitionAudio(
+    ...     uri='gs://my-bucket/recording.flac')
+    >>> config=speech.types.RecognitionConfig(
```
tseaver: For assignments, please have spaces on either side of the `=` (versus when passing as arguments).  Likewise below, three times.

speech/docs/index.rst
```diff
@@ -19,16 +19,13 @@ See: `Speech Asynchronous Recognize`_
...
+    ...     encoding=speech.enums.RecognitionConfig.AudioEncoding.LINEAR16,
+    ...     language_code='en-US',
+    ...     sample_rate_hertz=44100)
+    >>> operation = client.long_running_recognize(config, audio)
```
tseaver: Please keep the "keyword argument" style here (and below):

```python
>>> operation = client.long_running_recognize(config=config, audio=audio)
```

tseaver:
@nnegrey Thank you for the patch!

PR #6873, opened 2018-12-07
API Core: fix pytype build

### LINKED ISSUE #6833: API Core PyType fails intermittently/changing errors
https://github.com/googleapis/google-cloud-python/pull/6769

Copying from #6769 and @rchen152 

> @crwilcox It looks like what's happening is that pytype is somehow finding multiple files named path_template.py (and timeout.py and grpc_helpers.py - the error probably differs because it varies which file pytype discovers first). The other thing I noticed is that the file count in the status message ("Analyzing 53 sources with 0 dependencies") differs from the one I see (28 sources). Any idea why there might be two copies of some source files?
> 
> EDIT: if there are two files with the same name in different directories, this may be google/pytype#198.

tseaver:
Closes #6833.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ generated_python/
 cloud-bigtable-client/
 googleapis-pb/
 grpc_python_venv/
+pytype_output/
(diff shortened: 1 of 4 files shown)


tseaver:
Gonna merge without review, as the pytype error is breaking CI whenever the `api_core` tests get run.

PR #6876, opened 2018-12-07
NDB: Start in on `ndb.client.Client`

chrisrossi:
This is patterned off of `datastore.client.Client` and depends heavily on `google.cloud.client`.

### DIFF
--- a/ndb/src/google/cloud/ndb/client.py
+++ b/ndb/src/google/cloud/ndb/client.py
@@ -38,12 +38,12 @@ def _determine_default_project(project=None):
     * GOOGLE_CLOUD_PROJECT environment variable
     * Google App Engine application ID
     * Google Compute Engine project ID (from metadata server)
+_
+    Arguments:
+        project (Optional[str]): The project to use as default.
 
-    :type project: str
-    :param project: Optional. The project to use as default.
-
-    :rtype: str or ``NoneType``
-    :returns: Default project if it can be determined.
+    Returns:
+        Union([str, None]): Default project if it can be determined.
     """
     if project is None:
         project = _get_gcd_project()
--- a/ndb/src/google/cloud/ndb/client.py
+++ b/ndb/src/google/cloud/ndb/client.py
@@ -73,7 +73,7 @@ class Client(ClientWithProject):
     def __init__(self, project=None, namespace=None, credentials=None):
         super(Client, self).__init__(project=project, credentials=credentials)
         self.namespace = namespace
-        self._host = os.environ.get(environment_vars.GCD_HOST, _DATASTORE_HOST)
+        self.host = os.environ.get(environment_vars.GCD_HOST, _DATASTORE_HOST)
 
     @property
     def _http(self):
--- a/ndb/src/google/cloud/ndb/client.py
+++ b/ndb/src/google/cloud/ndb/client.py
@@ -1,0 +1,90 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A client for NDB which manages credentials, project, namespace."""
+
+import os
+
+from google.cloud import environment_vars
+from google.cloud.client import ClientWithProject
+from google.cloud import _helpers
+
+_DATASTORE_HOST = "datastore.googleapis.com"
+
+
+def _get_gcd_project():
+    """Gets the GCD application ID if it can be inferred."""
+    return os.getenv(environment_vars.GCD_DATASET)
+
+
+def _determine_default_project(project=None):
+    """Determine default project explicitly or implicitly as fall-back.
+
+    In implicit case, supports four environments. In order of precedence, the
+    implicit environments are:
+
+    * DATASTORE_DATASET environment variable (for ``gcd`` / emulator testing)
+    * GOOGLE_CLOUD_PROJECT environment variable
+    * Google App Engine application ID
+    * Google Compute Engine project ID (from metadata server)
+
+    :type project: str
+    :param project: Optional. The project to use as default.
+
+    :rtype: str or ``NoneType``
+    :returns: Default project if it can be determined.
+    """
+    if project is None:
+        project = _get_gcd_project()
+
+    if project is None:
+        project = _helpers._determine_default_project(project=project)
+
+    return project
+
+
+class Client(ClientWithProject):
+    """An NDB client.
+
+    Arguments:
+        project (Optional[str]): The project to pass to proxied API methods. If
+            not passed, falls back to the default inferred from the
+            environment.
+        namespace (Optional[str]): Namespace to pass to proxied API methods.
+        credentials (Optional[:class:`~google.auth.credentials.Credentials`]):
+            The OAuth2 Credentials to use for this client. If not passed, falls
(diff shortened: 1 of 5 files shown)


ndb/src/google/cloud/ndb/client.py
```diff
@@ -0,0 +1,90 @@
...
+
+"""A client for NDB which manages credentials, project, namespace."""
+
+import os
+
+from google.cloud import environment_vars
+from google.cloud.client import ClientWithProject
+from google.cloud import _helpers
+
+_DATASTORE_HOST = "datastore.googleapis.com"
```
dhermes: Why is this needed? Isn't there a constant in `google-cloud-datastore` we can use?

ndb/src/google/cloud/ndb/client.py
```diff
@@ -0,0 +1,90 @@
...
+    if project is None:
+        project = _get_gcd_project()
+
+    if project is None:
+        project = _helpers._determine_default_project(project=project)
+
+    return project
+
+
+class Client(ClientWithProject):
```
dhermes: Why not inherit from `google.cloud.datastore.client.Client`?

ndb/tests/conftest.py
```diff
@@ -40,3 +43,28 @@ def reset_state():
     yield
     model.Property._FIND_METHODS_CACHE.clear()
     model.Model._kind_map.clear()
+
+
+@pytest.fixture
+def environ():
+    """Copy of ``os.environ``"""
+    original = os.environ.copy()
```
dhermes: This copy isn't needed since you overwrite `os.environ`?

ndb/tests/conftest.py
```diff
@@ -40,3 +43,28 @@ def reset_state():
...
+@pytest.fixture(autouse=True)
+def initialize_environment(request, environ):
+    """Set environment variables to default values.
+
+    There are some variables, like ``GOOGLE_APPLICATION_CREDENTIALS``, that we
+    want to reset for unit tests but not system tests. This fixture introspects
+    the current request, determines whether it's in a unit test, or not, and
+    does the right thing.
+    """
+    if request.module.__name__.startswith("tests.unit"):
```
dhermes: TIL the [`request`][1] fixture, thanks!

[1]: https://docs.pytest.org/en/latest/reference.html#request

ndb/tests/unit/test_client.py
```diff
@@ -0,0 +1,72 @@
...
+    with patch:
+        yield creds
+
+
+class TestClient:
+    @staticmethod
+    def test_constructor_no_args():
+        with patch_credentials("testing"):
+            client = client_module.Client()
+        assert client.SCOPE == ("https://www.googleapis.com/auth/datastore",)
```
dhermes: Is this an instance attribute named `SCOPE`?

chrisrossi:
> @chrisrossi I'm not clear why this special Client class is needed. A parallel to ndb would be https://github.com/googleapis/google-cloud-python-happybase, which doesn't define a Client of it's own (but does have a similar metaphor).

A couple of days ago, before starting this, I sent around an email entitled "NDB client story", where I was pretty much fishing for exactly this kind of input. Did you see that?

I can look at happybase--I'm not familiar with it at the moment. Basically, if we're going to interact with datastore, we need to get credentials, figure out the project/app, and possibly get a namespace. Since this is just a layer on datastore, I looked to datastore for the model on how to do it. I'm not wed to anything in particular, but I try not to invent new idioms if there is one already available, and I figured datastore probably represented some kind of "best practice" for Google APIs.

chrisrossi:
> @dhermes
> 
> > @chrisrossi I'm not clear why this special Client class is needed. A parallel to ndb would be https://github.com/googleapis/google-cloud-python-happybase, which doesn't define a Client of it's own (but does have a similar metaphor).
> 
> A couple of days ago, before starting this, I sent around an email entitled "NDB client story", where I was pretty much fishing for exactly this kind of input. Did you see that?
> 
> I can look at happybase--I'm not familiar with it at the moment. Basically, if we're going to interact with datastore, we need to get credentials, figure out the project/app, and possibly get a namespace. Since this is just a layer on datastore, I looked to datastore for the model on how to do it. I'm not wed to anything in particular, but I try not to invent new idioms if there is one already available, and I figured datastore probably represented some kind of "best practice" for Google APIs.

I've looked at HappyBase this morning and it has a `Connection` object which is the moral equivalent of the `Client` object here. We need some way to manage the little bit of state that gets used to talk to Datastore. `google.cloud.datastore` uses a `Client` object and it makes sense to me for us also to use a `Client` object that is the parallel of that.

ndb/src/google/cloud/ndb/client.py
chrisrossi: There is `google.cloud.datastore.client._DATASTORE_BASE_URL`, which could be used to derive the host. It's private, and you had mentioned copying private members from Datastore rather than depending on them. We don't use a URL, since we don't support the HTTP transport, but we could derive the host from the URL and may implement that HTTP transport in a future iteration. It's not clear to me there's a single right answer here.

ndb/src/google/cloud/ndb/client.py
chrisrossi: Because `google.cloud.datastore.client.Client` just inherits from `ClientWithProject` and really only adds things that are specific to Datastore that would have to be overridden for NDB. I looked, and there didn't seem to be enough stuff we would want to reuse so I used it as a guide to follow instead.

ndb/tests/conftest.py
chrisrossi: raise FingersFasterThanBrainError()

Fixed: 78f02e7

ndb/tests/conftest.py
chrisrossi: :+1:

ndb/tests/unit/test_client.py
chrisrossi: It's a class attribute. Ends up getting used by the super-class when authenticating, I believe.

ndb/src/google/cloud/ndb/client.py
dhermes: OK

dhermes:
@chrisrossi I'd say push on with `Client`, but try to think of a way **not** to have this if we can avoid it. `ndb`'s design is essentially going to force global state to be stored on the module, joy!

chrisrossi:
> @chrisrossi I'd say push on with `Client`, but try to think of a way **not** to have this if we can avoid it. `ndb`'s design is essentially going to force global state to be stored on the module, joy!

@dhermes I'm reasonably confident we can contain all the state we need in `client.Client` and the thread locals in `_runstate`.

ndb/src/google/cloud/ndb/client.py
dhermes: Isn't it defined anywhere in the `gapic` module(s)?

chrisrossi: Found it in _http. Which is technically still private, but I guess I'll use that.

chrisrossi:
@dhermes Yep, fixed.

ndb/tests/conftest.py
```diff
@@ -40,3 +43,28 @@ def reset_state():
     yield
     model.Property._FIND_METHODS_CACHE.clear()
     model.Model._kind_map.clear()
+
+
+@pytest.fixture
+def environ():
+    """Copy of ``os.environ``"""
+    original = os.environ
+    environ = original.copy()
```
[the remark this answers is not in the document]
chrisrossi: Maybe not...

~~~~
(gcpdev) chris@opportunity:~/proj/google/clouddev/ndb$ pylint google.cloud.ndb tests | wc -l
2732
~~~~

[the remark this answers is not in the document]
chrisrossi: The global is the function, not the import of os.environ, FWIW. Would be an actual problem if environ were a recursive function. "Fixed" anyway.

pylint also complains about line 30, because the arg `environ` masks the global `environ`, which is the fixture. Of course, the named argument, `environ`, is what signals to `pytest` to use the fixture named `environ`. 

Maybe we get bogged down in detail sometimes?

PR #6888, opened 2018-12-10
NDB: Get Datastore gRPC API stub

chrisrossi:
A helper function which takes a `client.Client` instance and returns an instance of the gRPC "stub" for the Google Datastore API. Based off of #6876

### DIFF
--- a/ndb/src/google/cloud/ndb/_api.py
+++ b/ndb/src/google/cloud/ndb/_api.py
@@ -1,0 +1,41 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Functions that interact with Datastore backend."""
+
+import grpc
+
+from google.cloud import _helpers
+from google.cloud import _http
+from google.cloud.datastore_v1.proto import datastore_pb2_grpc
+
+
+def stub(client):
+    """Get a stub for the `Google Datastore` API.
+
+    Arguments:
+        client (:class:`~client.Client`): An NDB client instance.
+
+    Returns:
+        :class:`~google.cloud.datastore_v1.proto.datastore_pb2_grpc.DatastoreStub`:
+            The stub instance.
+    """
+    if client.secure:
+        channel = _helpers.make_secure_channel(
+            client._credentials, _http.DEFAULT_USER_AGENT, client.host
+        )
+    else:
+        channel = grpc.insecure_channel(client.host)
+    stub = datastore_pb2_grpc.DatastoreStub(channel)
+    return stub
--- a/ndb/src/google/cloud/ndb/client.py
+++ b/ndb/src/google/cloud/ndb/client.py
@@ -73,6 +73,9 @@ class Client(google_client.ClientWithProject):
     SCOPE = ("https://www.googleapis.com/auth/datastore",)
     """The scopes required for authenticating as a Cloud Datastore consumer."""
 
+    secure = True
+    """Whether to use a secure connection for API calls."""
+
     def __init__(self, project=None, namespace=None, credentials=None):
         super(Client, self).__init__(project=project, credentials=credentials)
         self.namespace = namespace
--- a/ndb/tests/unit/test__api.py
+++ b/ndb/tests/unit/test__api.py
@@ -1,0 +1,51 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest import mock
+
+from google.cloud import _http
+from google.cloud.ndb import _api
+
+
+class TestStub:
+    @staticmethod
+    @mock.patch("google.cloud.ndb._api._helpers")
+    @mock.patch("google.cloud.ndb._api.datastore_pb2_grpc")
+    def test_secure_channel(datastore_pb2_grpc, _helpers):
+        channel = _helpers.make_secure_channel.return_value
+        client = mock.Mock(
+            _credentials="creds",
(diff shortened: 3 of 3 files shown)


ndb/src/google/cloud/ndb/_api.py
```diff
@@ -0,0 +1,41 @@
...
+    """Get a stub for the `Google Datastore` API.
+
+    Arguments:
+        client (:class:`~client.Client`): An NDB client instance.
+
+    Returns:
+        :class:`~google.cloud.datastore_v1.proto.datastore_pb2_grpc.DatastoreStub`:
+            The stub instance.
+    """
+    user_agent = "gcloud-python/ndb"
```
dhermes: @chrisrossi I don't think this is acceptable. @tseaver What is the right path here?

ndb/src/google/cloud/ndb/client.py
```diff
@@ -70,6 +70,9 @@ class Client(ClientWithProject):
     SCOPE = ("https://www.googleapis.com/auth/datastore",)
     """The scopes required for authenticating as a Cloud Datastore consumer."""
 
+    secure = True
```
dhermes:
```suggestion
    SECURE = True
```

What is the genesis / need of this class attribute?

ndb/src/google/cloud/ndb/_api.py
chrisrossi: I poked around just now and datastore uses `google.cloud._http.DEFAULT_USER_AGENT`, which currently is `"gcloud-python/0.28.1"`. I can just make it do that.

Seems like it could potentially be useful to log on the server side which calls are coming from `google.cloud.datastore` versus which calls are coming from `google.cloud.ndb`, but this isn't the place to do that, that's fine.

(Interestingly, I can't import `google.cloud._http` unless I've already imported `google.cloud.datastore`--not sure what the dependency chain is there, exactly, but it seems weird.)

ndb/src/google/cloud/ndb/client.py
chrisrossi: https://github.com/googleapis/google-cloud-python/blob/572b3df37c3e66d5a59e279c0e71838254212625/datastore/google/cloud/datastore/_gapic.py#L39

Datastore selects a secure channel or an insecure channel for gRPC based on the `base_url`--if it starts with "https", it creates a secure channel. Since we're just using `host` and not `base_url`, I added a boolean flag. It's meant to be overridden on a per-instance basis, like `base_url` in Datastore. I'm not sure what the use case is for Datastore for overriding the default `base_url` (testing? read-only access? non-Google Datastore servers?), but I figured I needed to preserve it, in spirit, by allowing the host and secure flag to be overridden.

ndb/src/google/cloud/ndb/_api.py
crwilcox: The other clients use `ClientInfo` from api_core.

```
 if client_info is None:
            client_info = google.api_core.gapic_v1.client_info.ClientInfo(
                gapic_version=_GAPIC_LIBRARY_VERSION
            )
        else:
            client_info.gapic_version = _GAPIC_LIBRARY_VERSION
        self._client_info = client_info
```

ndb/src/google/cloud/ndb/client.py
dhermes: Why does it need to be a class attribute then?

ndb/src/google/cloud/ndb/_api.py
dhermes: @chrisrossi LMK what you decide to do here?

chrisrossi: @crwilcox That piece is at a higher layer than we're currently dealing with. In `google.cloud.datastore` there are a few layers of indirection wrapping what eventually winds up being calls to the `datastore_pb2_grpc.DatastoreStub()`. 

The layers of the onion, from highest level to lowest level are:

`google.cloud.datastore.client.Client` =>
`google.cloud.datastore_v1.gapic.datastore_client.DatastoreClient` =>
`google.cloud.datastore_v1.gapic.transports.datastore_grpc_transport.DatastoreGrpcTransport' =>
`google.cloud.datastore_v1.proto.datastore_pb2_grpc.DatastoreStub`

The second layer is the lowest layer that does anything with a `client_info` object. Only the stub explicitly has any API that is useful for async. gRPC stub methods can be called with `.future` to return a Future for that call. The reason I'm explicitly creating a stub here, and bypassing all of this other machinery in `google.cloud.datastore`, is none of the other layers expose this, and async is a big part of the requirements for NDB.

This is what I was getting at earlier when I was saying that there should be a more generalized approach to async that was used across all the client libraries. Because `google.cloud.datastore` doesn't have any async surface, NDB, rather than being written on top of `google.cloud.datastore`, is having to be written in parallel with it, in the process duplicating a lot of what's already in `google.cloud.datastore`.

Looking at this again, I'm not entirely certain I have the right approach here. The transport, at least, exposes stub methods explicitly, so we could potentially use the transport instead of the stub, although I'm not really sure what that buys us, besides just being a layer closer to the surface. Maybe instead of creating a channel and a stub, we instantiate the full `client.Client`, and then go spelunking for the transport so we can make calls on that.

What gives me heartburn, though, is looking at the second layer down, `datastore_client.DatastoreClient`, all the API calls are getting wrapped with `google.api_core.gapic_v1.method.wrap_method` for adding "retry, timeout, and the like". I'm guessing we might want that. So maybe we need to duplicate this class, too. We can't just use it directly, because async, so if we need it, we need to write our own version of it. I don't know much about api_core yet. Can it wrap the async versions of those functions? (Skimming the code now....reasonably certain that's a no.) Do we need to reinvent that machinery, too?

Looking at api_core, I just noticed there is a `google.api_core.future` package. Anybody know anything about that? Doesn't show up in the docs. Looks like a handful of libraries use it. Hmm...

So the more I look at it, the more questions I have. So I'm not positive we'll wind up even using this particular scrap of code here. Sorry for the brain dump--your question caused me to go looking and this is what I dug up.

chrisrossi: But back to the particular matter at hand, the standard for the user agent seems to be: "gcloud-python/{google-cloud-core.version}", which I can import from `google.cloud._http`, so I'll do that. I still maintain that it would potentially be more helpful if there was greater differentiation between the user against reported by the clients, but that's a discussion for another venue.

ndb/src/google/cloud/ndb/client.py
tseaver: FWIW, using class attributes as defaults seems particularly sane in a library which already abuses the snot out of `__new__`. :)

ndb/src/google/cloud/ndb/_api.py
dhermes: @theacodes and @lukesneeringer did most of the design of the wrapping code, so they may be helpful for a quick chat about how many layers into the onion you need to go

chrisrossi: @theacodes @lukesneeringer Let me know if you think you can tell me anything helpful here! Not the user agent stuff, but the best way to reuse or reproduce the layers of the "onion" above with NDB, which requires async access to the API.

chrisrossi:
@dhermes @tseaver Ok to merge for now?

There's still some open high level questions of how best to interact with the the backend, here, but if I keep everything related to talking to datastore in `_api`, then it should be refactorable, at least, down the road.

dhermes:
Yes

PR #6904, opened 2018-12-11
Firestore: normalize FieldPath parsing / escaping

### LINKED ISSUE #6549: [Firestore] Field Path escaping in python doesn't match other languages
There are no FieldPath overloads in the SDK. Instead, there is a field_path() method that returns an encoded string. While the Python SDK expects fully escaped field paths, 
other SDKS escape user-provided field paths by default and split on dots for this escaping.

tseaver:
Refactor:

- Move `FieldPath` and related helpers to a new `field_path` module.
- Rename 'get_field_path' helper -> 'render_field_path' to clarify intent.
- Split out `FieldPath.from_api_repr` from existing `FieldPath.from_string` factory:
  - `FieldPath.from_api_repr` parses the path based on the API's quoting rules.
  - `FieldPath.from_string` first tries parsing using `from_api_repr`, but falls back to a more relaxed parsing.

Closes #6549.


firestore/google/cloud/firestore_v1beta1/field_path.py
```diff
@@ -0,0 +1,386 @@
+# Copyright 2017 Google LLC All rights reserved.
```
crwilcox: Current header format is `# Copyright 2018 Google LLC`. Did you pick 2017 as that was the author date of this code originally?

firestore/tests/unit/test_field_path.py
```diff
@@ -0,0 +1,495 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Google LLC All rights reserved.
```
crwilcox: same comment about header.

firestore/google/cloud/firestore_v1beta1/field_path.py
tseaver: I just copied the header from `_helpers.py`.  ISTM that the original copyright assertion applies, although I don't really care one way or the other.

tseaver: 091abe9

firestore/tests/unit/test_field_path.py
tseaver: 091abe9

PR #6905, opened 2018-12-12
Add Client Library for IAM

crwilcox:


iam/google/cloud/iam/credentials.py
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+from google.cloud.iam.credentials_v1 import IAMCredentialsClient
+from google.cloud.iam.credentials_v1 import types
+
+__all__ = ("types", "IAMCredentialsClient")
```
tseaver: This directory structure does not fit our normal pattern, which would be something like `google/cloud/iam_credentials_v1/` for the service.  Since `google.cloud.iam` is not a namespace package, the extra layer provides no benefit.

iam/README.rst
```diff
@@ -0,0 +1,75 @@
...
+================================================================
+
+`IAM Service Account Credentials API`_: IAM Service Account Credentials API
+
+- `Client Library Documentation`_
+- `Product Documentation`_
+
+.. _Alpha: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/README.rst
+.. _IAM Service Account Credentials API: https://cloud.google.com/iamcredentials
+.. _Client Library Documentation: https://googlecloudplatform.github.io/google-cloud-python/latest/iamcredentials/usage.html
```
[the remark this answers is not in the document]
crwilcox: these should all be IAM. the output name is iamcredentials of the gapic so it filled these in wrong.

iam/google/cloud/iam/credentials.py
crwilcox: While I can't be sure I assume there is a reason? I will look into the configuration yaml for the generation and see if this was an error.

iam/README.rst
```diff
@@ -7,9 +7,9 @@ Python Client for IAM Service Account Credentials API (`Alpha`_)
 - `Product Documentation`_
 
 .. _Alpha: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/README.rst
-.. _IAM Service Account Credentials API: https://cloud.google.com/iamcredentials
-.. _Client Library Documentation: https://googlecloudplatform.github.io/google-cloud-python/latest/iamcredentials/usage.html
-.. _Product Documentation:  https://cloud.google.com/iamcredentials
+.. _IAM Service Account Credentials API: https://cloud.google.com/iam
```
tseaver: Should this link be specific to the [IAM credentials docs](https://cloud.google.com/iam/credentials/reference/rest/)?

crwilcox: just generalize this to be 'iam'. it isn't clear if we are going to bring other services yet, but it seems likely.

PR #6912, opened 2018-12-12
Firestore: enable use of 'WriteBatch' as a context manager.

### LINKED ISSUE #6548: [Firestore] WriteBatch doesn't return instance so you cannot chain.
The WriteBatch methods don’t return the WriteBatch instances for chaining.

tseaver:
Closes #6548.


tseaver:
@crwilcox Failed systests are all due to quota issues in the `watch` tests.

PR #6913, opened 2018-12-12
fix IAM docs build

crwilcox:


iam/synth.py
```diff
@@ -59,6 +59,13 @@
     "from google\.iam\.credentials\.v1 import iamcredentials_pb2_grpc",
     "from google.cloud.iam_credentials_v1.proto import iamcredentials_pb2_grpc"
 )
+
+s.replace(
+    "google/cloud/iam_credentials_v1/proto/common_pb2.py",
+    "\"\"\"Attributes:\n",
+    "\"\"\"\nAttributes:\n"
```
tseaver: That is *hideous* -- avoiding using `'` is a clear lose here for black.

PR #6917, opened 2018-12-13
Skip signing tests for insufficient credentials

crwilcox:


crwilcox:
Fix for internal bug: 113623918

storage/tests/system.py
```diff
@@ -699,6 +700,10 @@ def test_third_level(self):
         self.assertEqual(iterator.prefixes, set())
 
 
+@unittest.skipUnless(
+    type(storage.Client()._credentials) is google.oauth2.service_account.credentials,
+    "Can only test signing with credentials capable of signing",
+)
```
tseaver: We can't instantiate a client at import time here, because there might not be *any* credentials.  What is the usecase?

crwilcox: If a user creates their credentials through gCloud they will have credentials of the type google.oauth2.client. These credentials are not capable of singing and these tests will fail for no reason other than that

tseaver: We need to raise the skip from inside the testcase, then.

storage/tests/system.py
```diff
@@ -703,6 +704,12 @@ class TestStorageSignURLs(TestStorageFiles):
     def setUp(self):
         super(TestStorageSignURLs, self).setUp()
 
+        if (
+            type(storage.Client()._credentials)
```
tseaver: Inside here, you should be able to use `Config.CLIENT._credentials`.

PR #6932, opened 2018-12-16
Storage: add missing call to batch.begin() in example.

sorced-jim:
This should have an email that matches a CLA.


tseaver:
Thanks for the patch, @sorced-jim!

PR #6935, opened 2018-12-17
Firestore: implement listen conformance

### LINKED ISSUE #6533: Firestore:  support features blacklisted in conformance tests

mcdonc:
Fixes two bugs in the watch code:

- always-truthiness of protobufs foiled a delete/remove document case.
- change map would receive keys that were document snapshots instead of document names sometimes,
  foiling the reset code.

Closes #6533

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/watch.py
+++ b/firestore/google/cloud/firestore_v1beta1/watch.py
@@ -377,9 +377,9 @@ class Watch(object):
 
     def _on_snapshot_target_change_add(self, proto):
         _LOGGER.debug("on_snapshot: target change: ADD")
-        assert (
-            WATCH_TARGET_ID == proto.target_change.target_ids[0]
-        ), "Unexpected target ID sent by server"
+        target_id = proto.target_change.target_ids[0]
+        if target_id != WATCH_TARGET_ID:
+            raise RuntimeError("Unexpected target ID %s sent by server" % target_id)
 
     def _on_snapshot_target_change_remove(self, proto):
         _LOGGER.debug("on_snapshot: target change: REMOVE")
@@ -391,9 +391,9 @@ class Watch(object):
             code = change.cause.code
             message = change.cause.message
 
-        # TODO: Consider surfacing a code property on the exception.
-        # TODO: Consider a more exact exception
-        raise AssertionError("Error %s:  %s" % (code, message))
+        message = "Error %s:  %s" % (code, message)
+
+        raise RuntimeError(message)
 
     def _on_snapshot_target_change_reset(self, proto):
         # Whatever changes have happened so far no longer matter.
--- a/firestore/google/cloud/firestore_v1beta1/watch.py
+++ b/firestore/google/cloud/firestore_v1beta1/watch.py
@@ -79,9 +79,6 @@ class WatchDocTree(object):
     def keys(self):
         return list(self._dict.keys())
 
-    def items(self):
-        return list(self._dict.items())
-
     def _copy(self):
         wdt = WatchDocTree()
         wdt._dict = self._dict.copy()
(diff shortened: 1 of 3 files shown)


firestore/google/cloud/firestore_v1beta1/watch.py
```diff
@@ -396,7 +393,7 @@ def _on_snapshot_target_change_remove(self, proto):
 
         # TODO: Consider surfacing a code property on the exception.
         # TODO: Consider a more exact exception
-        raise Exception("Error %s:  %s" % (code, message))
+        raise AssertionError("Error %s:  %s" % (code, message))
```
tseaver: We prefer to avoid `AssertionError` in non-testing code.  Maybe `RuntimeError`?  Or would be we be better off raising the exception returned by `google.api_core.exceptions.exception_class_for_grpc_status`?

firestore/tests/unit/test_watch.py
```diff
@@ -403,7 +403,7 @@ class DummyRemove(object):
 
         remove = DummyRemove()
         proto.document_remove = remove
-        proto.document_delete = None
+        proto.document_delete = ""
```
tseaver: These changes seem to push us over the line to use a real `ListenResponse` message, e.g.:

```python
def _make_listen_response(**kw):
    from google.cloud.firestore_v1beta1.proto.firestore_pb2 import ListenResponse
    return ListenResponse(**kw)
```

because the messages get all their scalar fields iniitalized to defaults for free.

firestore/google/cloud/firestore_v1beta1/watch.py
mcdonc: Yeah.  I used AssertionError because there are a bunch of assert() statements in various methods in watch.py and it's unclear which methods should raise which kind of error if we change this one case.  I guess we could use some existing api_core.exceptions exception, but those are currently all square peg in round hole because they are raised when the client is calling the server via HTTP, and watch is the inverse case.  Maybe I should change api_core.exceptions to have some generic WatchError that inherits from GoogleAPICallError, but even GoogleAPICallError is square peg because the errors dont happen during any API call.  Maybe just a custom exception class in watch.py itself.

mcdonc: I changed it to RuntimeError if only to not have to document a one-off exception type.

firestore/tests/unit/test_watch.py
mcdonc: I suppose we could, but the proto object isn't the only composite object that needs to be created; it needs a target_change object and a document_change object as attributes and those are also complex objects, so we'd either have to pass a textproto to  ListenResponse.FromString or invent some other deserializer to do that and I'm not sure doing either of those things would make the code any clearer.

tseaver:
[Firestore system test failures due to quota issues](https://source.cloud.google.com/results/invocations/9dc0eb1a-f693-413d-8b79-bb8b61672a6b).

PR #7013, opened 2018-12-20
Implement ``Client.context()``

chrisrossi:
As far as the public API is concerned, ``Client.context()`` replaces
``ndb_context``, which has been renamed to ``state_context`` and removed from
the public API. ``_api.stub()`` has been refactored to store the Datastore
stub in the current context.

### DIFF
--- a/ndb/docs/client.rst
+++ b/ndb/docs/client.rst
@@ -1,0 +1,7 @@
+######
+Client
+######
+
+.. automodule:: google.cloud.ndb.client
+    :members:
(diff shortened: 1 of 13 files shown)


chrisrossi:
> LGTM. What's the motivation here?

Just a better user experience—consolidating everything the user has to know/worry about regarding runtime state management to the client. The client was going to have to be added to the context, anyway, so it made sense to just move everything to the client so the user only has one thing to fool with.

PR #7016, opened 2018-12-22
Bigtable Row Set snippets

sangramql:


sangramql:
@tseaver @sduskis @AVaksman
Please review the updated PR.

bigtable/docs/snippets_table.py
```diff
@@ -403,5 +412,77 @@ def test_bigtable_table_row():
...
+
+    rows = []
+    for row_key in row_keys:
+        row = Config.TABLE.row(row_key)
+        row.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
+        rows.append(row)
+    Config.TABLE.mutate_rows(rows)
+
+    # [START bigtable_add_row_key]
+    from google.cloud.bigtable import Client
```
sduskis: I don't think we need the Client related code here.  The RowSet code is enough.

bigtable/docs/snippets_table.py
```diff
@@ -403,5 +412,77 @@ def test_bigtable_table_row():
...
+    row_set.add_row_key(b'row_key_5')
+    # [END bigtable_add_row_key]
+
+    read_rows = table.read_rows(row_set=row_set)
+    expected_row_keys = [b'row_key_5']
+    found_row_keys = [row.row_key for row in read_rows]
+    assert found_row_keys == expected_row_keys
+
+    # [START bigtable_add_row_range]
+    from google.cloud.bigtable import Client
```
sduskis: same here.

bigtable/docs/snippets_table.py
```diff
@@ -403,5 +412,77 @@ def test_bigtable_table_row():
...
+    # [END bigtable_add_row_range]
+
+    read_rows = table.read_rows(row_set=row_set)
+    expected_row_keys = [b'row_key_3', b'row_key_4',
+                         b'row_key_5', b'row_key_6']
+    found_row_keys = [row.row_key for row in read_rows]
+    assert found_row_keys == expected_row_keys
+
+    # [START bigtable_row_range_from_keys]
+    from google.cloud.bigtable import Client
```
sduskis: and here.

bigtable/docs/snippets_table.py
billyjacobson: This follows the format we use for the api docs samples for python, so I think it is fine to leave this in.

billyjacobson:
Looks good to me

Issue #6300, opened 2018-10-24
BigQuery: TypeError: b'v\xa9\' is not JSON serializable


WillianFuks:
Os: Ubuntu 16.04
pipenv python 3.5
google-cloud-bigquery: 1.6.0

I'm basically trying to get a few rows from the public bitcoin dataset and save it in a personal destination table. Here's the code I used:

    from google.cloud import bigquery as bq
    client = bq.Client.from_service_account_json('service_account.json')

    bc_ds_ref = client.dataset('bitcoin_blockchain', 'bigquery-public-data')
    bc_table_ref = bq.TableReference(bc_ds_ref, 'blocks')
    bc_table = client.get_table(bc_table_ref)

    bc_data = list(client.list_rows(bc_table, max_results=10))
    
    # saving the result to new destination table
    dest_ds_ref = client.dataset('dest_dataset', 'dest_project')
    dest_table_ref = bq.TableReference(dest_ds_ref, 'sampled_bitcoin_blocks')
    dest_table = bq.Table(dest_table_ref, schema=bc_table.schema)

    job = job = client.insert_rows(dest_table, bc_data)

At this point an exception is thrown; the full stacktrace:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-38-5efecba8d0ed> in <module>
----> 1 job = client.insert_rows(dest_table, bc_data)

~/.local/share/virtualenvs/gcloud-4d5VpL3n/lib/python3.5/site-packages/google/cloud/bigquery/client.py in insert_rows(self, table, rows, selected_fields, **kwargs)
   1322             json_rows.append(json_row)
   1323 
-> 1324         return self.insert_rows_json(table, json_rows, **kwargs)
   1325 
   1326     def insert_rows_json(self, table, json_rows, row_ids=None,

~/.local/share/virtualenvs/gcloud-4d5VpL3n/lib/python3.5/site-packages/google/cloud/bigquery/client.py in insert_rows_json(self, table, json_rows, row_ids, skip_invalid_rows, ignore_unknown_values, template_suffix, retry)
   1402             method='POST',
   1403             path='%s/insertAll' % table.path,
-> 1404             data=data)
   1405         errors = []
   1406 

~/.local/share/virtualenvs/gcloud-4d5VpL3n/lib/python3.5/site-packages/google/cloud/bigquery/client.py in _call_api(self, retry, **kwargs)
    334         if retry:
    335             call = retry(call)
--> 336         return call()
    337 
[...]

PR #7022, opened 2018-12-24
Issue #6300 - Fix nested schema parsing in insert_rows 

adamf:
Hi! Here's a potential fix for handling JSON serializing of nested schema in BigQuery (issue 6300). 

I think before accepting this it would need tests, but I wanted to hear the maintainer's thoughts on the location of the fix (`client.py` or `_helpers.py`). Also, I'm sure there's a better way to do this, but I needed a fix for this problem in my project :)

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1440,6 +1440,97 @@ class Client(ClientWithProject):
 
         return query_job
 
+    def _convert_to_safe_json_dict(self, field, row_value):
+        """Maps a field and value to a JSON-safe value.
+
+        Args:
+            field ( \
+                :class:`~google.cloud.bigquery.schema.SchemaField`, \
+            ):
+                The SchemaField to use for type conversion and field name.
+
+            row_value (Union[ \
+                Sequence[list], \
+                any, \
+            ]):
+                Row value(s) to be converted. If the value is
+                a list, converted all elements in the list and return
+                a dictionary whose value is that converted list.
+
+        Returns:
+            dict:
+                A dict of a field name to value(s), encoded for JSON.
+        """
+        if not isinstance(row_value, list):
+            converter = _SCALAR_VALUE_TO_JSON_ROW.get(field.field_type)
+            value = {}
+            if converter is not None:  # STRING doesn't need converting
+                value[field.name] = converter(row_value)
+            else:
+                value[field.name] = row_value
+            return value
+
+        values = []
+        for v in row_value:
+            converter = _SCALAR_VALUE_TO_JSON_ROW.get(field.field_type)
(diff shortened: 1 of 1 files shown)


adamf:
I signed it!

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1440,6 +1440,97 @@ def query(
 
         return query_job
 
+    def _convert_to_safe_json_dict(self, field, row_value):
```
tswast: Since these methods don't actually require a client (they don't make any API calls themselves), I'd prefer these live in _helpers.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1440,6 +1440,97 @@ def query(
...
+        values = []
+        for v in row_value:
+            converter = _SCALAR_VALUE_TO_JSON_ROW.get(field.field_type)
+            if converter is not None:  # STRING doesn't need converting
+                values.append(converter(v))
+            else:
+                values.append(v)
+        return {field.name: values}
+
+    def _handle_nested_schema(self, field, row_value):
```
tswast: I'd prefer this be named `_field_value_to_json`, to match the naming convention in `_helpers.py`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1440,6 +1440,97 @@ def query(
...
+            ]):
+                Row value(s) to be converted. If the value is
+                a list, converted all elements in the list and return
+                a dictionary whose value is that converted list.
+
+        Returns:
+            dict:
+                A dict of a field name to value(s), encoded for JSON.
+        """
+        if not isinstance(row_value, list):
```
tswast: We'd want to treat a tuple the same as we do a list. I'd prefer we look at `field.mode == 'REPEATED'` to decide this instead.

Can we refactor the "REPEATED" logic into a function `_repeated_field_to_json` instead? The remaining logic could be in a function named `_scalar_field_to_json`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1440,6 +1440,97 @@ def query(
...
+
+        if not is_repeated and not is_record:
+            return self._convert_to_safe_json_dict(field, row_value)
+
+        # if it is repeated, but not a record, loop thru the values, and return
+        if is_repeated and not is_record:
+            return self._convert_to_safe_json_dict(field, row_value)
+
+        # if it is a record, but not repeated, just parse the subfields:
+        if is_record and not is_repeated:
```
tswast: I'd prefer if this logic lived in a function `_record_field_to_json`.

adamf:
Thanks @tswast ! Appreciate the fixes and tests!

PR #7023, opened 2018-12-25
Add missing generation arguments for bucket and blob methods.

jmcarp:
Allow setting generation for getting and deleting blobs, and respect generation for rewriting.

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -73,7 +73,7 @@ class _PropertyMixin(object):
         raise NotImplementedError
 
     @property
-    def query_params(self):
+    def _query_params(self):
         """Default query parameters."""
         params = {}
         if self.user_project is not None:
@@ -106,7 +106,7 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
-        query_params = self.query_params
+        query_params = self._query_params
         # Pass only '?projection=noAcl' here because 'acl' and related
         # are handled via custom endpoints.
         query_params["projection"] = "noAcl"
@@ -156,7 +156,7 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
-        query_params = self.query_params
+        query_params = self._query_params
         # Pass '?projection=full' here because 'PATCH' documented not
         # to work properly w/ 'noAcl'.
         query_params["projection"] = "full"
@@ -185,7 +185,7 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
-        query_params = self.query_params
+        query_params = self._query_params
         query_params["projection"] = "full"
         api_response = client._connection.api_request(
             method="PUT",
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -43,8 +43,9 @@ class _PropertyMixin(object):
     """Abstract mixin for cloud storage classes with associated properties.
 
     Non-abstract subclasses should implement:
-      - client
       - path
+      - client
+      - user_project
 
     :type name: str
     :param name: The name of the object. Bucket names must start and end with a
@@ -70,6 +71,14 @@ class _PropertyMixin(object):
     def user_project(self):
         """Abstract getter for the object user_project."""
         raise NotImplementedError
+
+    @property
+    def query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
 
     def _require_client(self, client):
         """Check client or verify over-ride.
@@ -97,11 +106,10 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
+        query_params = self.query_params
         # Pass only '?projection=noAcl' here because 'acl' and related
         # are handled via custom endpoints.
-        query_params = {"projection": "noAcl"}
-        if self.user_project is not None:
-            query_params["userProject"] = self.user_project
+        query_params["projection"] = "noAcl"
         api_response = client._connection.api_request(
             method="GET", path=self.path, query_params=query_params, _target_object=self
         )
@@ -148,11 +156,10 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
+        query_params = self.query_params
         # Pass '?projection=full' here because 'PATCH' documented not
         # to work properly w/ 'noAcl'.
-        query_params = {"projection": "full"}
-        if self.user_project is not None:
-            query_params["userProject"] = self.user_project
+        query_params["projection"] = "full"
         update_properties = {key: self._properties[key] for key in self._changes}
 
         # Make the API call.
@@ -178,9 +185,8 @@ class _PropertyMixin(object):
                        ``client`` stored on the current object.
         """
         client = self._require_client(client)
-        query_params = {"projection": "full"}
-        if self.user_project is not None:
-            query_params["userProject"] = self.user_project
+        query_params = self.query_params
+        query_params["projection"] = "full"
         api_response = client._connection.api_request(
             method="PUT",
             path=self.path,
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -258,6 +258,16 @@ class Blob(_PropertyMixin):
         return self.bucket.user_project
 
     @property
+    def query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.generation is not None:
+            params["generation"] = self.generation
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
+
+    @property
     def public_url(self):
         """The public URL for this blob.
 
@@ -387,6 +397,9 @@ class Blob(_PropertyMixin):
         # minimize the returned payload.
         query_params = {"fields": "name"}
 
+        if self.generation:
+            query_params["generation"] = self.generation
+
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
 
@@ -423,7 +436,9 @@ class Blob(_PropertyMixin):
                  (propagated from
                  :meth:`google.cloud.storage.bucket.Bucket.delete_blob`).
         """
-        return self.bucket.delete_blob(self.name, client=client)
+        return self.bucket.delete_blob(
+            self.name, client=client, generation=self.generation
+        )
 
     def _get_transport(self, client):
         """Return the client's transport.
@@ -1486,7 +1501,7 @@ class Blob(_PropertyMixin):
             query_params["rewriteToken"] = token
 
         if source.generation:
-            query_params['sourceGeneration'] = source.generation
+            query_params["sourceGeneration"] = source.generation
 
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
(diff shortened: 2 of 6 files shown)


tseaver:
What is implemented here looks OK, but seems like it might not be comprehensive enough (e.g., should `Blob.reload` / `Blob.exists` / `Blob.delete` support generations?).

@frankyn PTAL.

jmcarp:
Thanks @tseaver. I took a stab at adding generations to mixed-in methods like `reload` and `update`. WDYT?

jmcarp:
Ping @tseaver @frankyn when you have time.

frankyn:
Thanks for the ping. Looking over it now.

storage/google/cloud/storage/bucket.py
```diff
@@ -345,6 +345,14 @@ def user_project(self):
...
+    @property
+    def query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
+
     def blob(self, blob_name, chunk_size=None, encryption_key=None, kms_key_name=None):
         """Factory constructor for blob object.
```
frankyn: I think defining the `generation` would be helpful here as well. Given I pass around the generation number and then use it to create a blob instance without performing a `GET`. 

This could be used with for example other operations: `patch()`, `update()`, `rewrite()` for example.

jmcarp:
Thanks @frankyn. Updated the PR.

storage/tests/unit/test_bucket.py
```diff
@@ -569,6 +569,21 @@ def test_get_blob_hit_w_user_project(self):
...
         self.assertEqual(kw["query_params"], {"userProject": USER_PROJECT})
 
+    def test_get_blob_hit_w_generation(self):
+        NAME = "name"
+        BLOB_NAME = "blob-name"
+        GENERATION = 1512565576797178
+        connection = _Connection({"name": BLOB_NAME})
+        client = _Client(connection)
+        bucket = self._make_one(name=NAME)
+        blob = bucket.get_blob(BLOB_NAME, client=client, generation=GENERATION)
```
frankyn: This might be overdoing it a bit, but can you add a unit test for bucket.blob() as well? Additionally could you add a test for the constructor of the Blob class?

jmcarp:
Updated. Looks like we already exercise the `Blob` constructor with `generation` in https://github.com/jmcarp/google-cloud-python/blob/blob-generations/storage/tests/unit/test_blob.py#L104-L109.

jmcarp:
Ping @frankyn when you have time; ready for another look.

storage/tests/system.py
```diff
@@ -414,29 +416,52 @@ def test_crud_blob_w_user_project(self):
...
+            self.assertEqual(blob1.download_as_string(), "gen1")
 
             # Exercise 'objects.patch' w/ userProject.
-            blob.content_language = "en"
-            blob.patch()
-            self.assertEqual(blob.content_language, "en")
+            blob0.content_language = "en"
+            blob0.patch()
+            self.assertEqual(blob0.content_language, "en")
+            self.assertIsNone(blob1.content_language)
```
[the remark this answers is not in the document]
jmcarp: We're patching the language of the 0th generation of the blob, so we expect the metadata of the 1st generation to still be `None`.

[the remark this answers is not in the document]
frankyn: ahhh I missed that it's `blob1` and not `blob0` thank you.

jmcarp:
Anything else I should do on this @frankyn @JustinBeckwith?

frankyn:
@crwilcox I'm good with the merge wdyt?

storage/google/cloud/storage/blob.py
```diff
@@ -387,6 +406,9 @@ def exists(self, client=None):
         # minimize the returned payload.
         query_params = {"fields": "name"}
 
+        if self.generation:
+            query_params["generation"] = self.generation
+
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
```
tseaver: Seems like this block should use the `_query_params` property, e.g.:

```python

        # minimize the returned payload.
        query_params = self._query_params
        query_params["fields"] = "name"
```

storage/google/cloud/storage/blob.py
```diff
@@ -1485,6 +1509,9 @@ def rewrite(self, source, token=None, client=None):
         if token:
             query_params["rewriteToken"] = token
 
+        if source.generation:
+            query_params["sourceGeneration"] = source.generation
+
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
```
tseaver: Likewise here:

```python
        query_params = self._query_params

        if token:
            query_params["rewriteToken"] = token

        if source.generation:
            query_params["sourceGeneration"] = source.generation

        if self.kms_key_name is not None:
            query_params["destinationKmsKeyName"] = self.kms_key_name
```

storage/google/cloud/storage/_helpers.py
```diff
@@ -71,6 +72,14 @@ def user_project(self):
         """Abstract getter for the object user_project."""
         raise NotImplementedError
 
+    @property
+    def _query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
```
tseaver: This method is shadowed by both `Blob` and `Bucket`:  let's make it virtual, e.g.:

```python
    @property
    def _query_params(self):
        """Default query parameters."""
        raise NotImplementedError
```

That way we don't have to add test coverage for it, either.

storage/google/cloud/storage/bucket.py
```diff
@@ -586,6 +612,10 @@ def get_blob(self, blob_name, client=None, encryption_key=None, **kwargs):
 
         if self.user_project is not None:
             query_params["userProject"] = self.user_project
+
+        if generation is not None:
+            query_params["generation"] = generation
+
         blob = Blob(
             bucket=self, name=blob_name, encryption_key=encryption_key, **kwargs
         )
```
tseaver: I'd rather create the blob with the expected generation here, and then use its `_query_params` property, e.g.:

```python
        blob = Blob(
            bucket=self, name=blob_name, encryption_key=encryption_key, generation=generation, **kwargs
        )
        query_params = blob._query_params

        try:
            headers = _get_encryption_headers(encryption_key)
            response = client._connection.api_request(
                method="GET",
                path=blob.path,
                query_params=query_params,
                headers=headers,
                _target_object=blob,
            )
```

Now that I look at it more closely, I'm not sure why this method doesn't just call `blob.reload()`.

storage/google/cloud/storage/bucket.py
```diff
@@ -825,6 +859,9 @@ def delete_blob(self, blob_name, client=None):
         client = self._require_client(client)
         query_params = {}
```
tseaver: Use `self._query_params` here as well, e.g.:

```python
        query_params = self._query_params

        if generation is not None:
            query_params["generation"] = generation

        blob_path = Blob.path_helper(self.path, blob_name)
```

storage/tests/system.py
```diff
@@ -414,29 +416,52 @@ def test_crud_blob_w_user_project(self):
 
         # Exercise 'objects.insert' w/ userProject.
         blob.upload_from_filename(file_data["path"])
+        gen0 = blob.generation
+
+        # Upload a second generation of the blob
+        blob.upload_from_file(io.StringIO(six.text_type("gen1")))
```
tseaver: This should be `io.BytesIO`, rather than `io.StringIO`.  Or you could just use `blob.upload_from_string(b"gen1")`.

storage/tests/system.py
```diff
@@ -414,29 +416,52 @@ def test_crud_blob_w_user_project(self):
...
         try:
             # Exercise 'objects.get' (metadata) w/ userProject.
             self.assertTrue(blob.exists())
             blob.reload()
 
             # Exercise 'objects.get' (media) w/ userProject.
-            downloaded = blob.download_as_string()
-            self.assertEqual(downloaded, file_contents)
+            self.assertEqual(blob0.download_as_string(), file_contents)
+            self.assertEqual(blob1.download_as_string(), "gen1")
```
tseaver: Please use bytes to get the test to pass on Python3.

storage/google/cloud/storage/blob.py
```diff
@@ -257,6 +266,16 @@ def user_project(self):
...
 
+    @property
+    def _query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.generation is not None:
+            params["generation"] = self.generation
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
```
tseaver: Please add explicit test coverage for the branches in this property.

storage/google/cloud/storage/bucket.py
```diff
@@ -345,7 +345,22 @@ def user_project(self):
...
         return self._user_project
 
-    def blob(self, blob_name, chunk_size=None, encryption_key=None, kms_key_name=None):
+    @property
+    def _query_params(self):
+        """Default query parameters."""
+        params = {}
+        if self.user_project is not None:
+            params["userProject"] = self.user_project
+        return params
```
tseaver: Please add explicit test coverage for the branches in this property.

storage/tests/system.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
```
tseaver: Import will be unused if you switch to `blob.upload_from_string` below.

storage/tests/unit/test_blob.py
```diff
@@ -576,7 +583,7 @@ def test_delete(self):
         bucket._blobs[BLOB_NAME] = 1
         blob.delete()
         self.assertFalse(blob.exists())
-        self.assertEqual(bucket._deleted, [(BLOB_NAME, None)])
+        self.assertEqual(bucket._deleted, [(BLOB_NAME, None, blob.generation)])
```
tseaver: Please duplicate this test, once with an explicit generation, and make this assertion be clear for both cases (`None` for the case where the blob is created without a generation).

storage/tests/unit/test_blob.py
```diff
@@ -2295,6 +2302,44 @@ def test_rewrite_response_without_resource(self):
...
+            "objectSize": 42,
+            "done": False,
+            "rewriteToken": TOKEN,
+        }
+        response = ({"status": http_client.OK}, RESPONSE)
+        connection = _Connection(response)
+        client = _Client(connection)
+        source_bucket = _Bucket(client=client)
+        source_blob = self._make_one(SOURCE_BLOB, bucket=source_bucket)
+        source_blob._set_properties({"generation": SOURCE_GENERATION})
```
tseaver: Please pass the generation to `_make_one`, e.g.:

```python
        source_blob = self._make_one(SOURCE_BLOB, bucket=source_bucket, generation=GENERATION)
```

storage/tests/unit/test_bucket.py
```diff
@@ -569,6 +584,21 @@ def test_get_blob_hit_w_user_project(self):
         self.assertEqual(kw["path"], "/b/%s/o/%s" % (NAME, BLOB_NAME))
         self.assertEqual(kw["query_params"], {"userProject": USER_PROJECT})
 
+    def test_get_blob_hit_w_generation(self):
+        NAME = "name"
+        BLOB_NAME = "blob-name"
+        GENERATION = 1512565576797178
+        connection = _Connection({"name": BLOB_NAME})
```
tseaver: For clarity, add the generation to the properties here, and then assert that it is read in correctly below:

```python
        connection = _Connection({"name": BLOB_NAME, generation=GENERATION})
```

storage/google/cloud/storage/_helpers.py
tseaver: Looking at how it is used in `reload`, I've changed my mind:  lets just get rid of the duplicate version in `Bucket`.

storage/google/cloud/storage/bucket.py
tseaver: Per above, let's just delete this version of the property, which duplicates the one in the base class.

PR #7025, opened 2018-12-28
Tasks: Add working example for 'create_queue'

msdinit:
This example is consistent with a similar change implemented in nodejs version here: googleapis/nodejs-tasks#142


tseaver:
@msdinit Thanks for the patch! Unfortunately, that docstring is in a module which is generated from the comments in the relevant [`.proto` file](https://github.com/googleapis/googleapis/blob/master/google/cloud/tasks/v2beta3/cloudtasks.proto), and will get overwritten by subsequent auto-generation runs (see #7057).

tseaver:
Also, now that I look at it, the doctest example is actually invalid:  it needs `...` as the prefix for the lines inside the opened `{`.

PR #7033, opened 2018-12-30
Bigtable appprofile snippets

sangramql:


sangramql:
@tseaver @sduskis @AVaksman
Please review the updated PR.

bigtable/docs/snippets.py
```diff
@@ -147,20 +149,19 @@ def test_bigtable_create_additional_cluster():
     cluster.delete()
 
 
-def test_bigtable_create_app_profile():
+def test_bigtable_create_reload_delete_app_profile():
+    import re
```
crwilcox: Putting this here will avoid the sample containing the necessary imports. [START bigtable_app_profile_name]

bigtable/docs/snippets.py
```diff
@@ -147,20 +149,19 @@ def test_bigtable_create_additional_cluster():
...
-    app_profile_id = "app-prof-" + unique_resource_id("-")
     description = "routing policy-multy"
-    routing_policy_type = enums.RoutingPolicyType.ANY
 
     app_profile = instance.app_profile(
-        app_profile_id=app_profile_id,
-        routing_policy_type=routing_policy_type,
+        app_profile_id=APP_PROFILE_ID,
+        routing_policy_type=ROUTING_POLICY_TYPE,
         description=description,
```
crwilcox: now that the others are named constants this seems inconsistent. Also, is their guidance from sample authoring to start using hidden constants? I only ask as having the values initialized in the sample helps the user understand what that value looks like.

Are `APP_PROFILE_ID` and `ROUTING_POLICY_TYPE` well understood vars? The other two constants are so I think they are fine (`CLUSTER_ID` and `INSTANCE_ID`)

bigtable/docs/snippets.py
sangramql: re is not needed for sample but is used for testing so left out of sample tags.

bigtable/docs/snippets.py
billyjacobson: App profile id is well defined. Routing policy type might be good to include in the sample since it is an enum, so you can show how to access it

sangramql: Implemented review. This was good suggestion.

sangramql:
@crwilcox , @tseaver, @billyjacobson 
Please review the changes.

sangramql:
@crwilcox , @tseaver 
Please review the changes.

sangramql:
@crwilcox , @tseaver
Please review the changes.

PR #7019, opened 2018-12-23
Bigtable Batcher, RowData, Row Operations, AppendRow snippets

sangramql:

PR #7035, opened 2018-12-30
Bigtable policy snippets

sangramql:


sangramql:
@tseaver @sduskis @AVaksman
Please review the updated PR.

bigtable/docs/snippets.py
```diff
@@ -425,5 +425,101 @@ def test_bigtable_set_iam_policy_then_get_iam_policy():
     assert len(policy.bigtable_admins) > 0
 
 
+def test_bigtable_admins_policy():
+    # [START bigtable_admins_policy]
+    from google.cloud.bigtable import Client
+    from google.cloud.bigtable.policy import Policy
+    from google.cloud.bigtable.policy import BIGTABLE_ADMIN_ROLE
+
```
crwilcox: Seems odd to include a blank line in the sample. Is this processed out? This is one of many instances of this formatting so the fix will need to be applied to other lines as well.

bigtable/docs/snippets.py
```diff
@@ -425,5 +425,101 @@ def test_bigtable_set_iam_policy_then_get_iam_policy():
...
+    policy_latest = instance.set_iam_policy(new_policy)
+    policy = policy_latest.bigtable_admins
+    # [END bigtable_admins_policy]
+
+    assert len(policy) > 0
+
+
+def test_bigtable_readers_policy():
+    # [START bigtable_readers_policy]
```
crwilcox: There are repeated tags here. I don't know how the parser is going to handle that. @jmdobry , is this a pattern that works?

sangramql: This has been already used and works fine.

crwilcox:
Ported to #7019

PR #7038, opened 2019-01-02
NDB: Datastore Lookup

chrisrossi:
Implement calling the Datastore Lookup RPC to fetch entities. Lookups are batched and batches are sent to the back end when results are requested for one of the batched lookups.


ndb/src/google/cloud/ndb/__init__.py
```diff
@@ -106,7 +106,6 @@
     "add_flow_exception",
     "Future",
     "get_context",
-    "get_return_value",
```
crwilcox: should this be mentioned in migration notes?

chrisrossi: One thing Legacy NDB doesn't do really well is distinguish between public and private API. This particular function isn't [meaningfully documented](https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/master/ndb/tasklets.py#L1032) and I can't really conceive of a reason it would be called by an NDB client, so I just made it private—I'm fairly certain just formalizing its _de facto_
status.

But yes, I can start a section in the migration notes for pieces that have been found to be _de facto_ private in legacy and made formally private in the rewrite.

crwilcox:
I am in favor of being more verbose. We can separate it out as you say, but
the more we can do to avoid surprising users, the better.

ndb/src/google/cloud/ndb/__init__.py
chrisrossi: ed76f47

PR #7068, opened 2019-01-08
NDB: A little more internal documentation

chrisrossi:
Diff base is #7065

### DIFF
--- a/ndb/src/google/cloud/ndb/_eventloop.py
+++ b/ndb/src/google/cloud/ndb/_eventloop.py
@@ -49,24 +49,97 @@ _Event = collections.namedtuple(
 class EventLoop:
     """An event loop.
 
-    Instances of ``EventLoop`` are used to coordinate single thraded execution
+    Instances of ``EventLoop`` are used to coordinate single threaded execution
     of tasks and RPCs scheduled asynchronously.
+
+    Since the the ``EventLoop`` runs in the same thread as user code, it's best
+    to think of it as running tasks "on demand". Generally, when some piece of
+    code needs a result from a future, the future's
+    :meth:`~tasklets.Future.wait` method will end up calling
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/key.py
```diff
@@ -705,26 +716,37 @@ def get(self, **ctx_options):
...
-            NotImplementedError: Always. The method has not yet been
-                implemented.
+            NotImplementedError: When ``ctx_options`` is not empty. Context
+                options have not yet been implemented.
+
+        Returns:
+            :class:`~google.cloud.ndb.tasklets.Future`
         """
-        raise NotImplementedError
+        from google.cloud.ndb import model  # avoid circular import
```
andrewsg: I'm guessing that because the module names and their relationships are basically defined by the previous version, it's very challenging to restructure to avoid circular imports in the first place, right?

ndb/src/google/cloud/ndb/model.py
```diff
@@ -16,7 +16,17 @@
 
 .. testsetup:: *
```
andrewsg: Are these actively used by our test code or are they just here as documentation?

ndb/src/google/cloud/ndb/key.py
chrisrossi: Precisely. I think if we had carte blanche to refactor everything, I'd argue that everything that could result in an API call (get, put, delete, query) should be on Client. So this would be `Client.get_async(key)`.

ndb/src/google/cloud/ndb/model.py
chrisrossi: They are actively used by the doctests. The doctests test the code snippets in the docs, but aren't considered part of the "real" test suite, though. They don't factor into coverage or anything like that.

PR #7070, opened 2019-01-08
NDB: Implement wait_any and wait_all.

chrisrossi:
These were originally implemented as class methods on the `tasklets.Future` class. I've made them standalone functions in `tasklets` but added static methods to `tasklets.Future` for backwards compatibility.

### DIFF
--- a/ndb/src/google/cloud/ndb/tasklets.py
+++ b/ndb/src/google/cloud/ndb/tasklets.py
@@ -41,6 +41,8 @@ __all__ = [
     "tasklet",
     "TaskletFuture",
     "toplevel",
+    "wait_all",
+    "wait_any",
 ]
 
 
@@ -204,6 +206,18 @@ class Future:
         """
         return False
 
+    @staticmethod
+    def wait_any(futures):
+        """Calls :func:`wait_any`."""
+        # For backwards compatibility
+        return wait_any(futures)
+
+    @staticmethod
+    def wait_all(futures):
+        """Calls :func:`wait_all`."""
+        # For backwards compatibility
+        return wait_all(futures)
+
 
 class TaskletFuture(Future):
     """A future which waits on a tasklet.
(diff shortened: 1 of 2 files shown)


ndb/src/google/cloud/ndb/tasklets.py
```diff
@@ -366,6 +380,39 @@ def tasklet_wrapper(*args, **kwargs):
...
+    """
+    if not futures:
+        return None
+
+    while True:
+        for future in futures:
+            if future.done():
+                return future
+
+        _eventloop.run1()
```
andrewsg: Just for my own edification, can you explain what this line does?

chrisrossi: Sure. I added #7068 in attempt to make code like this easier to understand. Possibly more can still be done.

Essentially, the `run1()` call does one thing on the eventloop—if there's a callback scheduled on the eventloop it calls that. If there aren't any call backs scheduled but there are pending gRPC calls, it waits for a gRPC call to finish and then runs its callback. Basically, this is the function that causes "something" to happen on the event loop where eventually one of those things will cause one of the futures we're waiting on to finish, and then we'll return that future.

I think the most confusing thing about the event loop is that it's not really an event loop in the way we typically think about it. Generally, if we're writing event driven code, everything we write is an event handler and the event loop is always running and calling into our event handlers. In NDB, on the other hand, the event loop is really just a task queue that we poke (by calling `run1()`) whenever we want something out of it. So, the flow of control is inverted, really, if you think about it. Maybe it shouldn't be called an "event loop". 

The one and only time the event loop actually loops is when we call `run()`, which we only do when we're wrapping up our context, to make sure anything in the queue gets run before exiting the context. You can imagine someone might have fired off a `put_async` without waiting for a response, but even if the calling code doesn't need the result, we still have to make sure it gets run. 

Does this, together with the comments in #7068, help it start to  make sense?

andrewsg:
Yes, that helps, thank you. If the task queue is empty, will this function
result in a hot loop that consumes 100% CPU? That will be relevant on Cloud
platforms that care about CPU consumption for billing. Is there any way to
architect this to be easier on CPU while looping?

chrisrossi:
No. The "loop" only runs for as long as there are tasks.  Either `run1()`
or `run()` will exit immediately if there's nothing to do.  When the loop
runs, it is only executing a callback that's been added to it or waiting
(blocking) on gRPC calls. The overhead added by the loop itself is
negligible.

Chris

PR #7071, opened 2019-01-09
PubSub: Propagate RetryError in PublisherClient.publish

relud:
in `google.cloud.pubsub_v1.publisher._batch.thread.Batch` [`self._client.api.publish(self._topic, self._messages)`](https://github.com/googleapis/google-cloud-python/pull/7071/files#diff-13577f0feaf50916f99888dc4258f6dcL201) may [raise `google.api_core.exceptions.RetryError`](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/retry.py#L163).

if `Batch._commit` does not catch an error then `Batch._futures` may never complete.

`google.api_core.exceptions.RetryError` and `google.api_core.exceptions.GoogleAPICallError` are the only inheritors of `google.api_core.exceptions.GoogleAPIError`, so that seemed like the appropriate error to catch.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -199,7 +199,7 @@ class Batch(base.Batch):
 
             try:
                 response = self._client.api.publish(self._topic, self._messages)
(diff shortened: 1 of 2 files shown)


tseaver:
@relud Thanks for following up with a test.

PR #7124, opened 2019-01-15
NDB: Implement Model.put() and Model.put_async()

chrisrossi:


ndb/src/google/cloud/ndb/_datastore_api.py
```diff
@@ -113,7 +113,7 @@ def idle():
         batch = batches.pop(options_key)
         batch.idle_callback()
 
-    batches[options_key] = batch = _LookupBatch(options)
+    batches[options_key] = batch = batch_cls(options)
```
chrisrossi: Fixes bug that was exposed in course of doing this.

ndb/src/google/cloud/ndb/model.py
```diff
@@ -336,15 +338,22 @@ def _entity_to_protobuf(entity):
     data = {}
     for cls in type(entity).mro():
         for prop in cls.__dict__.values():
-            if not isinstance(prop, Property) or prop._name in data:
+            if (
+                not isinstance(prop, Property)
+                or isinstance(prop, ModelKey)
```
chrisrossi: Fixes bug found while doing this work. __key__ (an instance of ModelKey) was getting into the properties being sent to Datastore, which would cause Datastore to puke.

PR #7125, opened 2019-01-15
fix typo

pjenvey:


crwilcox:
Hi @pjenvey , Thanks for catching this and fixing this. I don't have the authority to override our CLA. So, while this is almost certainly not copyrightable, it would be easier to have you sign the CLA :(. If you are unable to do that I can close this PR and open a bug for the typo.

PR #7126, opened 2019-01-15
Add option to choose dtypes by column in to_dataframe.

### LINKED ISSUE #7049: BigQuery: Support user-overridable dtypes in `to_dataframe` method.
With pandas 0.24.0 (unreleased), a new dtype is available to support nullable integer columns. http://pandas-docs.github.io/pandas-docs-travis/integer_na.html#integer-na The default behavior is to convert to float, but this can result in data loss (#6177). This new dtype extension avoids that.

I propose we allow the user to provide a map from column names to dtypes for any columns for which they'd like to override the default behavior. This argument could be called `dtype_overrides`.  This would also be useful for other extension dtypes in the future, such as for GEOGRAPHY columns.

See https://github.com/pydata/pandas-gbq/issues/242 for additional discussion.

**Alternatives**

- Make the new dtype for nullable integer the default for integer columns.
  - Con: Not compatible with older versions of pandas.
  - Con: Inconsistent with pandas's default behavior.

tswast:
This allows pandas users to select different sized floats for
performance at the expense of accuracy. With pandas 0.24, it will also
allow pandas users to use the new pandas.Int64Dtype() for nullable
integer columns.

Closes #7049

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import collections
 import copy
 import datetime
 import json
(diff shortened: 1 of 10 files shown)


tswast:
AutoML and Container tests look like flakes (problem downloading the Docker image). This change doesn't affect those clients.

Issue #6117, opened 2018-09-26
BigQuery: populate RowIterator.total_rows after running a query job


yiga2:
Request to have `total_rows` property public, under `QueryJob()` along with other stats

shollyman:
My initial preference would be against exposing this directly as property of the QueryJob, as its part of the jobs.getQueryResults / tabledata.list response and not a member of the query stats metadata.

Is the interest in fetching the table statistics for the results without the extra fetch of destination table metadata, or are you interested in it related to progress while fetching row data?

yiga2:
The interest is to determine how many rows a query, when completed, has generated. 
Fetching the destination table metadata is not valid if records are appended to an existing table - I would get the total number of rows, not the addition.

Note that `totalRows` exists in both `jobs.getQueryResults` and `jobs.query`. There is no need to call the former (except if job is inserted) so request is not unreasonable. Also, https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#totalBytesProcessed made it as property of `QueryJob()` so why not `total_rows`.

shollyman:
Thanks for clarifying the interest is in the effective delta.

For the case where you run a query with a destination table and a `WRITE_APPEND` disposition, the results (and total_rows) will represent the whole table.  For mutation DML queries (e.g. UPDATE/DELETE), `num_dml_affected_rows` may help, but we should consider exposing something like the `statistics.load.outputRows` equivalent for query jobs (likely contingent on statement type).

If that sounds like what you're after, I'll request it for inclusion in the backend response.  There's insufficient information in the existing response to correctly report the delta for the append case.

yiga2:
Let's cancel the request as I just suggested making the existing - and perfectly working -  `_query_results.total_rows` a public property of `QueryJob()` or another object. I thought https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.job.QueryJob.result.html#google.cloud.bigquery.job.QueryJob.result would work but it doesn't, `total_rows` is `None` (`myqueryjob.result().total_rows`).
This was reported on SO too: https://stackoverflow.com/questions/48799898/bigquery-py-client-library-v0-28-fetch-result-from-table-query-job

I will just use the private `_query_results.total_rows`.
Thanks.

yiga2:
[...]

PR #7217, opened 2019-01-29
Add sample for fetching total_rows from query results.

tswast:
In response to feedback internally Bug 123578325, add a sample (acts as a system test, too) which shows how to populate the `total_rows` value.

### DIFF
--- a/bigquery/docs/snippets.py
+++ b/bigquery/docs/snippets.py
@@ -2225,6 +2225,32 @@ def test_client_query_legacy_sql(client):
     # [END bigquery_query_legacy]
 
 
+def test_client_query_total_rows(client, capsys):
+    """Run a query an just check for how many rows."""
+    # [START bigquery_query_total_rows]
+    # from google.cloud import bigquery
+    # client = bigquery.Client()
+
+    query = (
+        "SELECT name FROM `bigquery-public-data.usa_names.usa_1910_2013` "
+        'WHERE state = "TX" '
+        "LIMIT 100"
+    )
+    query_job = client.query(
+        query,
+        # Location must match that of the dataset(s) referenced in the query.
+        location="US",
+    )  # API request - starts the query
+
+    results = query_job.result()  # Waits for query to complete.
+    next(iter(results))  # Fetch the first page of results, which contains total_rows.
+    print("Got {} rows.".format(results.total_rows))
+    # [START bigquery_query_total_rows]
+
+    out, _ = capsys.readouterr()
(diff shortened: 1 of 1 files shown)


bigquery/docs/snippets.py
```diff
@@ -2225,6 +2225,32 @@ def test_client_query_legacy_sql(client):
...
+    query_job = client.query(
+        query,
+        # Location must match that of the dataset(s) referenced in the query.
+        location="US",
+    )  # API request - starts the query
+
+    results = query_job.result()  # Waits for query to complete.
+    next(iter(results))  # Fetch the first page of results, which contains total_rows.
+    print("Got {} rows.".format(results.total_rows))
+    # [START bigquery_query_total_rows]
```
shollyman: END?

bigquery/docs/snippets.py
```diff
@@ -2225,6 +2225,32 @@ def test_client_query_legacy_sql(client):
     # [END bigquery_query_legacy]
 
 
+def test_client_query_total_rows(client, capsys):
+    """Run a query an just check for how many rows."""
```
shollyman: s/an/and/

yiga2:
Would fetching the first page of `result()` not make an API call ? If so, why not reading as per  https://github.com/googleapis/google-cloud-python/issues/6117. `_query_results` of the query job does get updated by `result()`

yiga2:
@tswast any further consideration on getting `total_rows` without API call ?

tswast:
@yiga2 I still think it's a good idea. I just haven't gotten around to. We're open to PRs. The change would likely be to the `result` method of `QueryJob` in `job.py`.

yiga2:
@tswast Feel free to bundle with other enhancements as PR would be very light otherwise.

Suggested code change, after https://github.com/googleapis/google-cloud-python/blob/a6d849912eea6b9260ccb64ffb317d8265b354a6/bigquery/google/cloud/bigquery/job.py#L2776
`total_rows = self._query_results.total_rows`

yiga2:
@tswast any feedback on this ?

tswast:
@yiga2 I've prepared https://github.com/googleapis/google-cloud-python/pull/7622 which addresses this issue, but in a slightly more complicated way than we propose here because I wanted to also handle more cases where `Client.list_rows` is called directly. My PR has had one pass at review, but I'm waiting on a follow-up now that I've addressed the requested changes.

yiga2:
Cool - thanks Tim !

PR #7221, opened 2019-01-29
Firestore: add 'Collection.list_documents' method.

### LINKED ISSUE #6545: [Firestore] Add list documents method to CollectionReference
CollectionReference.listDocuments() is missing.
This is new feature request and only available in Node, Java and Go so far.

tseaver:
Closes #6545.

### DIFF
--- a/firestore/google/cloud/firestore_v1beta1/collection.py
+++ b/firestore/google/cloud/firestore_v1beta1/collection.py
@@ -180,6 +180,33 @@ class CollectionReference(object):
             write_result = document_ref.create(document_data)
             return write_result.update_time, document_ref
 
+    def list_documents(self, page_size=None):
+        """List all subdocuments of the current collection.
+
+        Args:
+            page_size (Optional[int]]): The maximum number of documents
+            in each page of results from this request. Non-positive values
+            are ignored. Defaults to a sensible value set by the API.
+
+        Returns:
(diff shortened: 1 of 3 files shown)


firestore/google/cloud/firestore_v1beta1/collection.py
```diff
@@ -180,6 +180,33 @@ def add(self, document_data, document_id=None):
             write_result = document_ref.create(document_data)
             return write_result.update_time, document_ref
 
+    def list_documents(self, page_size=None):
+        """List all subdocuments of the current collection.
+
+        Args:
+            page_size (Optional[int]]): The maximum number of documents
+            in each page of results from this request. Non-positive values
+            are ignored. Defaults to a sensible value set by the API.
```
crwilcox: Where did you find this mentioned? I tried to find an actual sort of value here, but really page_size is a maximum, not a prescribed size. I am imagining there is an actual max? I guess I just don't like saying "it's magic"

crwilcox: So I am still unsure of this exact value. I will leave it up to you if we want "Defaults to a sensible value set by the API". it is true, I just don't like not knowing the answer.

tseaver: There is no documented number.

PR #7222, opened 2019-01-29
Add bigquery_storage to docs

busunkim96:


busunkim96:
@tswast Could you also take a look at this?

bigquery_storage/noxfile.py
```diff
@@ -125,23 +125,3 @@ def system(session):
 
     # Run py.test against the system tests.
     session.run('py.test', '--quiet', 'tests/system/')
-
-
-@nox.session(python='3.6')
-def docs(session):
```
tswast: I find it handy to have this so that I don't have to build _everything_ when I just want to see a single library's docs. Do we have to delete it?

PR #7241, opened 2019-01-31
Add args from RowIterator.to_dataframe() to QueryJob.to_dataframe()

tswast:


tswast:
> Needs a new test in 'tests/unit/test_job.py' which exercises passing explicit values for the new arguments.

Added in latest commit. Also added a unit test for the case of an empty iterator to catch the error caught in system tests in the unit tests.

PR #7245, opened 2019-01-31
Vision: get system logo tests healthy.

shollyman:
Change in backend model (internal issue 123358697) changed
how logo evaluation is done, and alters responses.  This
change selects a more appropriate test file and adjusts
expectations for the logo test.


tseaver:
I've restarted the tests to see if that was just a glitch.

shollyman:
Looks like there's a larger internal issue (114729036) causing the kokoro/ganpati errors, so if it happens again we can revisit once that's resolved.

PR #7261, opened 2019-02-02
Update table.py docs

johangp:
I changed some of the documentation of some methods like table.read_rows() because the parameter row_set in type is ok, but not in param. It was a little confuse to me.


johangp:
I signed it!

PR #7281, opened 2019-02-04
Firestore: fix the example for 'Query.on_snapshot'.

davidair:


tseaver:
@davidair Thank you for the patch!

PR #7285, opened 2019-02-04
Release bigquery 1.9.0

tswast:


tswast:
The test failure is that BigQuery is returning NonFound now for querying in the wrong location instead of BadRequest. I've mailed https://github.com/googleapis/google-cloud-python/pull/7286 with this fix.

PR #7286, opened 2019-02-04
Update error in system tests that expect error for wrong location

tswast:
We used to get BadRequest for querying with the wrong location. Now we
get NotFound. I update the test to just look for the base class of
GoogleAPICallError, since it seems we aren't guarnteed the actual status
code.


crwilcox:
@tswast do we know which change caused this?

tswast:
@crwilcox Something in the backend. Still trying to identify the specific CL.

PR #7299, opened 2019-02-07
NDB: port stats module from old ndb code

cguardia:
Ported the whole module, with unit tests. Wanted to add system tests, but it seems we need more functionality before that can be done.

### DIFF
--- a/ndb/docs/stats.rst
+++ b/ndb/docs/stats.rst
@@ -5,5 +5,4 @@ Datastore Statistics
 .. automodule:: google.cloud.ndb.stats
     :members:
     :inherited-members:
-    :undoc-members:
     :show-inheritance:
--- a/ndb/src/google/cloud/ndb/stats.py
+++ b/ndb/src/google/cloud/ndb/stats.py
@@ -50,7 +50,7 @@ class BaseStatistic(model.Model):
             the statistic instance.
         count (int): attribute is the total number of occurrences of the
             statistic in Cloud Datastore.
-        timestamp (DateTime): the time the statistic instance was written to Cloud
+        timestamp (datetime.datetime): the time the statistic instance was written to Cloud
             Datastore.
     """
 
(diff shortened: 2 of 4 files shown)


cguardia:
@chrisrossi about the comment in the code, do you know where the code that would supposedly use these classes resides? Same thing happens with metadata.

chrisrossi:
> @chrisrossi about the comment in the code, do you know where the code that would supposedly use these classes resides? Same thing happens with metadata.

I think this is it. These are model classes that reflect entities that Datastore creates automatically. (Or, at least, the GAE version.) I'd say you could construct a ``Key`` to ``get`` one of these, but I don't know how we'd know an ``id`` or ``name`` to use to construct the key. Unless you have some insight there, we may have to wait until we have a queries, then you'd just run a query on the ``kind`` of stat you're interested in. 

You should be able to explore what Google Cloud Datastore currently provides using ``google.cloud.datastore`` and running some queries. That's my theory, anyway.

cguardia:
@chrisrossi Thanks. I did run some queries with datastore when I was looking at this first. At the very least, it does provide these stats and uses the same names, so I think it's safe to assume that it still works like that. We'll need to wait for query for further action.

chrisrossi:
Notice that ``nox -e docs`` seems to be failing Kokoro.

cguardia:
Fixed. Sorry about that.

chrisrossi:
Great. Google is having us merge our own branches. Merge at your convenience.

Since we're working in a mono repo, I like to start all my squashed commit messages with "NDB: " at merge time, so it's obvious which API the commit is for.

chrisrossi:
You can merge this. We generally use "Squash and merge" to get each PR down to a single commit in master.

cguardia:
Hmmm, I did the command line merge last week, using --squash. Will try the button.

chrisrossi:
Is it possible you merged it cguardia and not google-cloud-python?

cguardia:
I think not, but I'm not sure. Will use the button from now on. Thanks.

chrisrossi:
Sorry, I just assumed that since the PR was still open, it was unmerged. It just occurred to me that if you squash, then github can't automatically detect if the branch was merged. Looks like it's in there twice, now. 

But, yes, if you use the button then it definitely knows it's merged.

Thanks.

PR #7301, opened 2019-02-07
Fix README to install spanner instead of datastore.

daniellehanks:


daniellehanks:
CLA signed.

tseaver:
@daniellehanks Thanks for the patch!

PR #7304, opened 2019-02-08
Add KMS system test

busunkim96:


kms/tests/system.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class TestKeyManagementServiceClient(object):
+    def test_list_global_key_rings(self):
+        project_id = os.environ["PROJECT_ID"]
+
+        client = kms_v1.KeyManagementServiceClient()
+        location = "global"
+        parent = client.location_path(project_id, location)
+        response = client.list_key_rings(parent)
```
tseaver: Is this just a smoke test, or should we have assertions about the values returned?

busunkim96: Just a smoke test

tseaver: The `lint` check doesn't like the unassigned variable.  Since we don't actually care about what gets returned, maybe just:

```python
        client = kms_v1.KeyManagementServiceClient()
        location = "global"
        parent = client.location_path(project_id, location)
        client.list_key_rings(parent)
```

tseaver:
@busunkim96 I'd say either go ahead and merge this PR (the other API failures can't be related) or else try to merge in the `master` branch or rebase to minimize the "detected diff."

PR #7307, opened 2019-02-11
Updated client library documentation URLs

pravindahal:
Previously, the URLs would redirect using JavaScript, which would either be slow or not work at all (in case JavaScript is disabled on the browser)


pravindahal:
I have signed the CLA.

PR #7341, opened 2019-02-14
Fix Batch object creation instructions

nielm:
Correctly specify  that a Batch object is constructed from a Database object, not a Client.

(these docs caused confusion leading to [a StackOverflow question](https://stackoverflow.com/questions/54683140/how-do-i-batch-upsert-data-into-google-cloud-spanner-using-the-python-client-lib)


tseaver:
@nielm Thanks very much for the patch!

PR #7373, opened 2019-02-15
Bigtable: ensure that 'Policy.etag' is text.

### LINKED ISSUE #7369: BigTable Policy.etag is bytes (should be base64 str)
#### Environment details

1. API: BigTable
2. OS type and version: OSX
3. Python version and virtual environment information: 3.7.0
4. google-cloud-bigtable 0.32.1

#### Steps to reproduce

1. Try to convert the `to_api_repr()` of a `google.cloud.bigtable.policy.Policy` to json.

#### Code example

```python
from google.cloud import bigtable
import json
client = bigtable.Client(admin=True)
(instances_list, failed_locations) = client.list_instances()
policy = instances_list[0].get_iam_policy()
json.dumps(policy.to_api_repr())
```

#### Stack trace

Unlike other policies (such as the one returned from `google.cloud.storage`), `etag` is not a `str` (contrary to [doc](https://github.com/googleapis/google-cloud-python/blob/e47616a382f9b5e765db0c7c334b7cbf225f9387/api_core/google/api_core/iam.py#L65)), and `to_api_repr` cannot be converted to json (contrary to [doc](https://github.com/googleapis/google-cloud-python/blob/e47616a382f9b5e765db0c7c334b7cbf225f9387/api_core/google/api_core/iam.py#L251)), apparently because bigtable’s [`Policy` passes `_to_bytes(etag)` to the superclass constructor](https://github.com/googleapis/google-cloud-python/blob/e47616a382f9b5e765db0c7c334b7cbf225f9387/bigtable/google/cloud/bigtable/policy.py#L76).

```
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

The `to_api_repr()` of a bigtable `Policy` is a dict containing binary bytes like `{'etag': b'\x00 \x01', 'version': 0}`, whereas the `to_api_repr()` of a `Policy` from google cloud storage is a dict whose etag is base64-encoded such as `{'etag': 'CAU=', 'bindings': […]}`

tseaver:
Closes #7369.

### DIFF
--- a/bigtable/google/cloud/bigtable/policy.py
+++ b/bigtable/google/cloud/bigtable/policy.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from google.api_core.iam import Policy as BasePolicy
-from google.cloud._helpers import _to_bytes
+from google.cloud._helpers import _bytes_to_unicode
 
 """IAM roles supported by Bigtable Instance resource"""
 BIGTABLE_ADMIN_ROLE = "roles/bigtable.admin"
@@ -73,7 +73,9 @@ class Policy(BasePolicy):
 
     def __init__(self, etag=None, version=None):
         BasePolicy.__init__(
-            self, etag=etag if etag is None else _to_bytes(etag), version=version
+            self,
+            etag=etag if etag is None else _bytes_to_unicode(etag),
+            version=version,
         )
 
     @property
--- a/bigtable/tests/unit/test_policy.py
+++ b/bigtable/tests/unit/test_policy.py
@@ -39,10 +39,24 @@ class TestPolicy(unittest.TestCase):
 
     def test_ctor_explicit(self):
         VERSION = 17
-        ETAG = b"ETAG"
+        ETAG = u"ETAG"
         empty = frozenset()
         policy = self._make_one(ETAG, VERSION)
         self.assertEqual(policy.etag, ETAG)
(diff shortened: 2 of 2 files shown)


tseaver:
After digging further, this patch is actually wrong, because the backend APIs for `Instance.get_iam_policy` and `Instance.set_iam_policy` use a protobuf message whose etag is passed explicitly as bytes.

We should therefore override `Policy.to_api_repr` and `Policy.from_api_repr` to handle converting the bytes to / from a JSON-compatible representation, and leave the attribute stored as bytes.

PR #7382, opened 2019-02-19
api_core: fix new lint failure

tseaver:
```
F632 use ==/!= to compare str, bytes, and int literals.
```


tseaver:
Merging w/o review as a trivial testing-only fix.

PR #7422, opened 2019-02-22
Fix typing - `:type kwargs: dict` is not correct

retnikt:
When typing Python code, the special `*args` and `**kwargs` params are typed based on their contents, not what they actually are (a tuple, and a dict) e.g.
```python
def foo(thing: int, *args: str):
    ...
def bar(thing: int, *args: str, **kwargs: Union[int, float]):
    ...

bar(-3, "G", spam=80, eggs=4.4)
foo(44, "abc", "xyz")
```

And in this case, the types are passed to a different func (`Query.__init__`) which has specific arguments, so it is best to not give **kwargs a type, or, better, copy the parameters of that function (but I haven't done this here).
This also applies to `Transaction`.
PS I know this is a stupidly minor change but it annoys me, having to write `noinspection PyTypeChecker` before every datastore query.
:smile:

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -534,7 +534,6 @@ class Client(ClientWithProject):
     def transaction(self, **kwargs):
         """Proxy to :class:`google.cloud.datastore.transaction.Transaction`.
 
-        :type kwargs: dict
         :param kwargs: Keyword arguments to be passed in.
         """
         return Transaction(self, **kwargs)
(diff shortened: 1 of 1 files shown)


tseaver:
@crwilcox I'm -1 on this change myself, but will leave it to you to decide:  if we are going to change how we document `kwargs` here, we should make the change throughout the codebase.

crwilcox:
I spent some time researching. This is also how mypy interprets types. I agree though @tseaver, we likely need to do this more broadly.

crwilcox:
Thanks for filing this @reznikt. Closing this in favor of a pr I just opened which will address this more broadly.

https://github.com/googleapis/google-cloud-python/pull/8546

PR #7425, opened 2019-02-22
Fix typo causing SyntaxError

pravindahal:
```
  File "/path/to/site/packages/google/cloud/ndb/query.py", line 926
    F"received {filters}")
    ^
SyntaxError: invalid syntax
```

Seems to be introduced by https://github.com/googleapis/google-cloud-python/pull/7239


ndb/src/google/cloud/ndb/query.py
```diff
@@ -907,7 +907,7 @@ def __init__(self, kind=None, ancestor=None, filters=None, orders=None,
             else:
                 if not isinstance(ancestor, model.Key):
                     raise TypeError("ancestor must be a Key; "
-                                    F"received {ancestor}")
+                                    "received {ancestor}")
```
tseaver: The `F` should be moved to the start of the literal (before `"ancestor`) so that the value actually gets formatted.  Likewise for all of these changes.

pravindahal:
This has been fixed in https://github.com/googleapis/python-ndb/pull/24 by @averikitsch, so I think we can close it here.

PR #7427, opened 2019-02-22
Add `Operation.deserialize`.

lukesneeringer:
This commit adds a `deserialize` method to the Operation object. The class method is a helper to deserialize the serialized protobuf operation messages.

We previously released this in `1.6.0a1` so I could make sure this was needed and the correct thing before committing to it. I am now requesting that we promote it. :-)


crwilcox:
LGTM. merging

PR #7428, opened 2019-02-22
Add support to unwrap Anys into wrapped pb2 objects.

lukesneeringer:
This commit adds support for unwrapping wrapped pb2 objects from Anys (needed for LRO support with wrapping).

This is new (not in 1.6.0a1) but tested against the gapic-generator-python output.


lukesneeringer:
Hold off on this one. The test is proving difficult to write (although I caught an error elsewhere, which was good) and I would prefer it be a bit more descriptive/meaningful. Will open another PR later.

PR #7429, opened 2019-02-22
Dataproc: Add nox session `docs`

busunkim96:
Adds a nox session `docs` in dataproc which builds the dataproc documentation. This does not break the top-level docs build. This is in preparation for the new documentation site and repo split. 

I'd like to make similar changes in the other directories and wanted to get input on this first.

### DIFF
--- a/dataproc/docs/README.rst
+++ b/dataproc/docs/README.rst
@@ -1,0 +1,1 @@
+../README.rst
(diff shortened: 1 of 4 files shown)


tseaver:
@busunkim96 `docs/conf.py` needs blackening.

PR #7434, opened 2019-02-25
Release api core 1.8.0

crwilcox:


tseaver:
@crwilcox How were we able to push a release to PyPI without a valid version in `setup.py`?

PR #7438, opened 2019-02-25
Add vision v1p4beta1

busunkim96:


vision/noxfile.py
```diff
@@ -45,10 +45,6 @@ def blacken(session):
     """Run black.
 
     Format code to uniform standard.
-    
-    This currently uses Python 3.6 due to the automated Kokoro run of synthtool.
```
crwilcox: Is synthtool up to date on your machine? It seems maybe the template is the prev. version?

busunkim96: Ah yep it needed updating, done.

Issue #7130, opened 2019-01-16
Firestore: 'test_watch_collection' systest flakes


tseaver:
Similar to #6605, #6921.

From this [Kokoro Firestore run](https://source.cloud.google.com/results/invocations/6579f747-286a-4545-b959-00d5fd1e530c/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Ffirestore/log):

```python
____________________________ test_watch_collection _____________________________
client = <google.cloud.firestore_v1beta1.client.Client object at 0x7ff27e1c25d0>
cleanup = <built-in method append of list object at 0x7ff27e1ac878>
    def test_watch_collection(client, cleanup):
        db = client
        doc_ref = db.collection(u"users").document(u"alovelace" + unique_resource_id())
        collection_ref = db.collection(u"users")
        # Initial setting
        doc_ref.set({u"first": u"Jane", u"last": u"Doe", u"born": 1900})
        # Setup listener
        def on_snapshot(docs, changes, read_time):
            on_snapshot.called_count += 1
            for doc in [doc for doc in docs if doc.id == doc_ref.id]:
                on_snapshot.born = doc.get("born")
        on_snapshot.called_count = 0
        on_snapshot.born = 0
        collection_ref.on_snapshot(on_snapshot)
        # delay here so initial on_snapshot occurs and isn't combined with set
        sleep(1)
        doc_ref.set({u"first": u"Ada", u"last": u"Lovelace", u"born": 1815})
        for _ in range(10):
            if on_snapshot.born == 1815:
                break
            sleep(1)
        if on_snapshot.born != 1815:
            raise AssertionError(
>               "Expected the last document update to update born: " + str(on_snapshot.born)
            )
E AssertionError: Expected the last document update to update born: 1900
```

HemangChothani:
closing the issue as solved in [8638](https://github.com/googleapis/google-cloud-python/pull/8638)

tseaver:
PR #8638 did not address this issue:  it was related to improving how the Firestore system tests clean up resources.

tseaver:
Hmm, looks like I spoke too soon:  changing to use a unique collection ID was part of #8638, and was likely a fix for this issue.

PR #7441, opened 2019-02-25
Storage: ensure that 'Blob.reload' passes encryption headers.

### LINKED ISSUE #7440: Storage: 'Blob.reload' does not pass encryption headers
Per the [docs on customer-supplied encryption keys](https://cloud.google.com/storage/docs/encryption/customer-supplied-keys#encrypted-elements), the `X-Goog-Encryption-Key` header applies not only the the blob / object's payload, but also to its CRC32c  checksum and MD5 hash fields.

`Blob.reload` (inherited from `_helpers._PropertyMixin` does not pass those headers, and thus the returned resource will not have the `crc32c` or `md5Hash` fields set.

tseaver:
Closes #7440.

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -86,6 +86,14 @@ class _PropertyMixin(object):
             client = self.client
         return client
 
+    def _encryption_headers(self):
+        """Return any encryption headers needed to fetch the object.
+
+        :rtype: dict
+        :returns: a mapping of encryption-related headers.
+        """
+        return {}
+
     def reload(self, client=None):
         """Reload properties from Cloud Storage.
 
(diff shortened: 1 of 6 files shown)


tseaver:
@frankyn The `_PropertyHelper` base class defines `reload`, which calls the `_encryption_headers` method:  it provides a no-op / empty implementation for `Bucket` (only `Blob` overrides it).

frankyn:
Thanks @tseaver, could you add a comment near the method stating that it's primary use is for a Blob and the given use case?

tseaver:
@frankyn

> Could you add a comment near the method stating that it's primary use is for a Blob and the given use case?

0c5a4f7

tseaver:
Unrelated Firestore failure is from a known-flaky systest, see #7130.

Issue #5781, opened 2018-08-09
Storage: add support for deleting specific blob generations


sdigit:
OS: Debian Stretch
Python 3.6.6
google-cloud-storage 1.10.0

In the following scenario, the API will return versioned blobs. However, it is unable to delete a specific version, as the `Blob.delete` method does not take a `generation` parameter nor does `Bucket.delete_blob`.

```
from google.cloud.storage import Client
client = Client(project='project-name')
bucket = client.get_bucket('foobar')
blobs = bucket.list_blobs(versions=True)
```

`blobs` will iterate over all versions, but calling `.delete()` on any of them will always delete the current version as the delete code does not supply `?generation=<generation>` in the request.

This leads to unintentional data loss when trusting `blob.delete()` to delete that specific blob.

Ideally, `blob.delete()` would delete the generation corresponding to that blob's `.generation` property (this would mimic `boto`'s behavior).

tseaver:
See  #4490. We do not have support for fetching specific generations / metagenerations of blobs / objects, but have future plans to add that support.

/cc @frankyn

frankyn:
Thanks @tseaver. Marking this in feature support sheet.

Issue #5546, opened 2018-06-27
Storage: fetch specific object version


thakursc1:
Is there a way to query a specific version of an object from cloud storage? 

```
def get_file_from_bucket(kind, id, ver:str)->dict:
    bucket = "bucket123"
    blob_path = "records" + "/" + kind + "/" + id+"#"+ver
    print("{}".format(blob_path))
    cred_file = Path(os.path.dirname((os.path.dirname(os.path.abspath(__file__)))),"service_file")
    storage_client = storage.Client.from_service_account_json(cred_file)
    bucket = storage_client.get_bucket(bucket)
    b = bucket.get_blob(blob_path)
    return b
```

`>> None `

The above function returns None, but when I remove the version I get the respective object. Is there a way to retrieve an object with specified version?

tseaver:
This is basically a duplicate of #4490.  We do not have support for fetching specific generations / metagenerations of blobs / objects, but have future plans to add that support.

/cc @frankyn

Issue #2538, opened 2016-10-12
support versioning options when fetching cloud storage buckets and objects


jcjones1515:
The [bucket](https://cloud.google.com/storage/docs/json_api/v1/buckets/get) and [object](https://cloud.google.com/storage/docs/json_api/v1/objects/get) REST APIs allow for fetching or at least checking for specific generations/metagenerations of buckets and objects, but these are not supported by the python client.

pdknsk:
This is basically a more generic duplicate of https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2463.

tseaver:
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2463#issuecomment-250479449 notes that the `generation` option should be used for the whole range of Object APIs which support it.  I have updated the title to indicate its broader scope.

Issue #2463, opened 2016-09-29
storage: cannot access / delete versioned objects


pdknsk:
Currently, `blob.delete` and `bucket.delete_blob` always delete the live version on buckets with versioning enabled. Other than being a missing feature, this can also cause some very unwanted behaviour, if a new version of the object is uploaded during the process of getting and deleting an object.

``` Diff
--- a/blob.py
+++ b/blob.py
@@ -279,7 +279,8 @@
                  (propagated from
                  :meth:`gcloud.storage.bucket.Bucket.delete_blob`).
         """
-        return self.bucket.delete_blob(self.name, client=client)
+        return self.bucket.delete_blob(self.name, self.generation, 
+            client=client)

     def download_to_file(self, file_obj, encryption_key=None, client=None):
         """Download the contents of this blob into a file-like object.
--- a/bucket.py
+++ b/bucket.py
@@ -357,7 +357,7 @@
         client.connection.api_request(method='DELETE', path=self.path,
                                       _target_object=None)

-    def delete_blob(self, blob_name, client=None):
+    def delete_blob(self, blob_name, generation=None, client=None):
         """Deletes a blob from the current bucket.

         If the blob isn't found (backend 404), raises a
@@ -392,6 +392,8 @@
         """
         client = self._require_client(client)
         blob_path = Blob.path_helper(self.path, blob_name)
+        if generation:
+            blob_path += '?generation=%i' % generation
         # We intentionally pass `_target_object=None` since a DELETE
         # request has no response value (whether in a standard request or
         # in a batch request).
```

tseaver:
@pdknsk Thanks for the report!  You are correct that we currently have no support for addressing a specific generation of a blob.

ISTM that the `generation` flag should probably be passed to `Blob.__init__` (and through `Bucket.blob()`, because it should affect other operations as well. 

API endpoints taking the `generation` query string parameter, which should be mapped from the blob's `generation` property, if set:

- https://cloud.google.com/storage/docs/json_api/v1/objects/get
- https://cloud.google.com/storage/docs/json_api/v1/objects/delete
- https://cloud.google.com/storage/docs/json_api/v1/objects/getIamPolicy
- https://cloud.google.com/storage/docs/json_api/v1/objects/insert (?!)
- https://cloud.google.com/storage/docs/json_api/v1/objects/patch (?!)
- https://cloud.google.com/storage/docs/json_api/v1/objects/setIamPolicy (?!)
[...]

PR #7444, opened 2019-02-25
Storage: add generation arguments (from #7023).

tseaver:
- Uses #7441 as a base.
- Merges in @jmcarp's work from #7023, with my requested changes from that PR.
- Closes #7023.
- Implements #5546, #5781, #2463, #2538.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -435,14 +435,21 @@ class Bucket(_PropertyMixin):
         return self._user_project
 
     @property
-    def query_params(self):
+    def _query_params(self):
         """Default query parameters."""
         params = {}
         if self.user_project is not None:
             params["userProject"] = self.user_project
         return params
 
-    def blob(self, blob_name, chunk_size=None, encryption_key=None, kms_key_name=None):
+    def blob(
+        self,
+        blob_name,
+        chunk_size=None,
+        encryption_key=None,
+        kms_key_name=None,
+        generation=None,
+    ):
         """Factory constructor for blob object.
 
         .. note::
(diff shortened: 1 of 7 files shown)


tseaver:
I've created #7478 to track the flaky systest, which I believe is unrelated to the changes in this PR (I cannot reproduce locally).

storage/google/cloud/storage/bucket.py
```diff
@@ -664,38 +677,33 @@ def get_blob(self, blob_name, client=None, encryption_key=None, **kwargs):
...
-                headers=headers,
-                _target_object=blob,
-            )
-            # NOTE: We assume response.get('name') matches `blob_name`.
-            blob._set_properties(response)
             # NOTE: This will not fail immediately in a batch. However, when
             #       Batch.finish() is called, the resulting `NotFound` will be
             #       raised.
-            return blob
+            blob.reload(client=client)
```
frankyn: Will `.reload()` pick up the value of `generation` through `_query_params()`?

tseaver: Yes, you are correct.

tseaver:
The flaky module tear-down for storage system tests is unrelated:  I've reported it in #7505.

747project:
Thank you for this! Is there a timeframe on when we can expect this to appear in a pypi release of google-cloud-storage?

max-sixty:
FYI this broke some of our tests - potentially on us but posting at least for others. 

Now if a process has a `blob` and it's overwritten, `blob.exists()` is False. It used to be true.

Specifically, this was our fixture:

```python

@pytest.fixture()
def existing_blob_name(project, bucket, blob_name_expanded, data):
    blob = gcs.client(project).get_bucket(
        bucket).blob(blob_name_expanded)
    blob.upload_from_string(dill.dumps(data))
    assert blob.exists()
    yield blob.name
    blob.delete()
```

This was our test (for an internal wrapper method):

```python

def test_set_overwrite(existing_blob_name, bucket, project):
    # could do this with a parameter but so small
        gcs.set(existing_blob_name, 'abc', bucket=bucket,
                project=project, overwrite=False)

    gcs.set(existing_blob_name, 'abc', bucket=bucket,
            project=project, overwrite=True)
```

This failed on `blob.delete()`, since the newly written blob has a different generation

Issue #6987, opened 2018-12-18
Error_reporting: 'test_report_exception' systest flakes w/ backoff error


tseaver:
See [failed system test runs](https://source.cloud.google.com/results/invocations/b4cf10d9-0eee-474d-8120-ef168ccf0efb/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Ferror_reporting/log).

```python
___________________ TestErrorReporting.test_report_exception ___________________
self = <test_system.TestErrorReporting testMethod=test_report_exception>
    def test_report_exception(self):
        # Get a class name unique to this test case.
        class_name = "RuntimeError" + unique_resource_id("_")
        # Simulate an error: group won't exist until we report
        # first exception.
        _simulate_exception(class_name, Config.CLIENT)
        is_one = functools.partial(operator.eq, 1)
        is_one.__name__ = "is_one"  # partial() has no name.
        retry = RetryResult(is_one, max_tries=6)
        wrapped_get_count = retry(_get_error_count)
>       error_count = wrapped_get_count(class_name, Config.CLIENT)
tests/system/test_system.py:123:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
args = ('RuntimeError_1545140110332', <google.cloud.error_reporting.client.Client object at 0x7fb7b6633828>)
kwargs = {}, tries = 6, result = None, delay = 32
msg = 'is_one. Trying again in 32 seconds...'
    @wraps(to_wrap)
    def wrapped_function(*args, **kwargs):
        tries = 0
        while tries < self.max_tries:
            result = to_wrap(*args, **kwargs)
            if self.result_predicate(result):
                return result
            delay = self.delay * self.backoff**tries
            msg = "%s. Trying again in %d seconds..." % (
                self.result_predicate.__name__, delay,)
            self.logger(msg)
            time.sleep(delay)
            tries += 1
>       raise BackoffFailed()
E       test_utils.retry.BackoffFailed
../test_utils/test_utils/retry.py:155: BackoffFailed
```

tseaver:
Another occurence today:

[...]

PR #7458, opened 2019-02-26
Error Reporting: Fix docstring replace in synth.

busunkim96:


tseaver:
Flaky systest, see #6987.

PR #7459, opened 2019-02-26
Fetch credentials for magic github proxy

busunkim96:
These credentials will be used by the releasetool reporter.


tseaver:
I think a rebase would likely fix the `Talent` and CircleCI builds, if we care.

PR #7483, opened 2019-03-05
Accept a string in Table and Dataset constructors.

### LINKED ISSUE #7477: BigQuery: accept a string in addition to a `[Dataset|Table]Reference` in the `[Dataset|Table]` constructors
This is a follow-up to https://github.com/googleapis/google-cloud-python/issues/6055

I missed the `Dataset` and `Table` constructors when making anywhere that takes a `[Dataset|Table]Reference` also accept a string. In these constructors, there isn't a project ID available, so the string would have to be fully-qualified (that is, it needs to contain a project ID).

tswast:
This removes the another need to manually create a TableReference or
DatasetReference. Instead, a developer can pass in a string to the
constructor and then set the needed properties on the resource.

Closes #7477

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -306,8 +306,13 @@ class Dataset(object):
     https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets
 
     Args:
-        dataset_ref (google.cloud.bigquery.dataset.DatasetReference):
-            a pointer to a dataset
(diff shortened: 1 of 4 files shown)


tswast:
@lakshmanok This PR should remove the need to ever manually create a TableReference or DatasetReference object. Once it's released, I can put the finishing touches on https://github.com/GoogleCloudPlatform/bigquery-oreilly-book/pull/2 to simplify the examples that using the `google-cloud-bigquery` library.

PR #7490, opened 2019-03-06
Add videointelligence v1p3beta1

busunkim96:


videointelligence/docs/index.rst
```diff
@@ -23,10 +23,10 @@ An API and type reference is provided for this beta:
 .. toctree::
     :maxdepth: 2
 
-    gapic/v1p2beta1/api
-    gapic/v1p2beta1/types
+    gapic/v1p3beta1/api
+    gapic/v1p3beta1/types
 
-The previous beta releases, spelled ``v1p1beta1``, ``v1beta1``, and
+The previous beta releases, spelled ``v1p1beta1``, ``v1p2beta1``, ``v1beta1``, and
```
crwilcox: I think we are sorting these newest-oldest, so v1p2 should be before v1p1 in this list.

PR #7506, opened 2019-03-12
Release api_core 1.8.1

crwilcox:


tseaver:
I thought you were going to hold off until we got the Firestore gRPC header fix in place.

crwilcox:
@tseaver, I am looking into that now, but because that likely means modifying the interface to bidi to support an additional argument, I figured that would trigger a minor release. This I can do as a patch which will be picked up by all clients since we don't pin to patch release.

Issue #7478, opened 2019-03-04
Storage: 'test_blob_w_explicit_kms_key_name' flake w/ extra blobs


tseaver:
From [this CI job last week](https://source.cloud.google.com/results/invocations/4d4144ae-3a13-41f5-b332-f3a5a889a3ac/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log), failing both for 2.7 and 3.6:

```python
_____________ TestKMSIntegration.test_blob_w_explicit_kms_key_name _____________
self = <tests.system.TestKMSIntegration testMethod=test_blob_w_explicit_kms_key_name>
    def test_blob_w_explicit_kms_key_name(self):
        BLOB_NAME = "explicit-kms-key-name"
        file_data = self.FILES["simple"]
        kms_key_name = self._kms_key_name()
        blob = self.bucket.blob(BLOB_NAME, kms_key_name=kms_key_name)
        blob.upload_from_filename(file_data["path"])
        self.case_blobs_to_delete.append(blob)
        with open(file_data["path"], "rb") as _file_data:
            self.assertEqual(blob.download_as_string(), _file_data.read())
        # We don't know the current version of the key.
        self.assertTrue(blob.kms_key_name.startswith(kms_key_name))
>       listed, = list(self.bucket.list_blobs())
E ValueError: too many values to unpack (expected 1)
```

Note that the `tearDownModule` failed as well (in both Python versions):

```python
_______ ERROR at teardown of TestIAMConfiguration.test_new_bucket_w_bpo ________
    def tearDownModule():
        errors = (exceptions.Conflict, exceptions.TooManyRequests)
        retry = RetryErrors(errors, max_tries=6)
>       retry(Config.TEST_BUCKET.delete)(force=True)
tests/system.py:88:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../test_utils/test_utils/retry.py:108: in wrapped_function
    return to_wrap(*args, **kwargs)
google/cloud/storage/bucket.py:889: in delete
    _target_object=None,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #7513, opened 2019-03-14
Asset: Generate v1, regenerate v1beta1

busunkim96:


busunkim96:
This looks like the same storage system flakiness in #7478

PR #7514, opened 2019-03-14
Allow passing rpc_metadata as part of creating a bidi

crwilcox:


firestore/google/cloud/firestore_v1beta1/watch.py
```diff
@@ -213,9 +213,11 @@ def should_recover(exc):  # pragma: NO COVER
             ResumableBidiRpc = self.ResumableBidiRpc  # FBO unit tests
 
         self._rpc = ResumableBidiRpc(
-            self._api.transport._stubs["firestore_stub"].Listen,
+            #self._api.transport._stubs["firestore_stub"].Listen,
```
tseaver: Please remove the commented line.

api_core/google/api_core/bidi.py
```diff
@@ -200,7 +203,7 @@ def open(self):
         request_generator = _RequestQueueGenerator(
             self._request_queue, initial_request=self._initial_request
         )
-        call = self._start_rpc(iter(request_generator))
+        call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
```
tseaver: We need unit test coverage here.

tseaver: Actually, we already have [failing tests here](https://source.cloud.google.com/results/invocations/fa903b7b-49d9-41d7-99a3-3a71942b275b/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fapi_core/log).

api_core/tests/unit/test_bidi.py
```diff
@@ -125,9 +125,10 @@ def make_rpc():
     call = mock.create_autospec(_CallAndFuture, instance=True)
     rpc = mock.create_autospec(grpc.StreamStreamMultiCallable, instance=True)
 
-    def rpc_side_effect(request):
+    def rpc_side_effect(request, metadata=None):
         call.is_active.return_value = True
         call.request = request
+        call.metadata = metadata
```
tseaver: Don't we need a test which explicitly passes the metadata and asserts it?

crwilcox: This is a good idea. I can add one.

PR #7520, opened 2019-03-15
use fqdn for gce metadata endpoint

### LINKED ISSUE #7517: Hard coded metadata endpoint breaks logging under GKE managed Istio
#### Environment details

1. API: GCE / GKE internal metadata endpoint
2. OS: anything running in GKE with [managed istio](https://cloud.google.com/istio/docs/istio-on-gke/overview) enabled.
3. Python version: 3.7.2 (container, no venv)
4. google-cloud-logging version: 1.9.0 (out of date but I double checked on master branch and `metadata` hostname is present there too).

#### Steps to reproduce

(please bear with me, I do get to the point!)

 1. Launch a GKE cluster and tick the managed istio box to enable it.
 2. Add the label `istio-inejction: enabled` to the default namespace to allow istio to control traffic on workloads deployed there.
 3. Launch a container into the namespace which will now have the istio/envoy sidecar automatically added to it. (eg. `kubectl run -i --tty busybox --image=busybox --restart=Never -- sh`)
 4. Attempt to access the hostname `metadata`.  
```
wget http://metadata/computeMetadata/v1/instance/attributes/cluster-name --header "Metadata-Flavor: Google"
Connecting to metadata (169.254.169.254:80)
wget: server returned error: HTTP/1.1 404 Not Found
```

This is the result of Istio's default policy of blocking all outbound requests until they are [whitelisted](https://istio.io/docs/tasks/traffic-management/egress/).

Since istio/istio#10694 will not accept non FQDN's the [hardcoded metadata endpoint in the python logging client](https://github.com/googleapis/google-cloud-python/blob/master/logging/google/cloud/logging/_helpers.py#L24) fails.

This manifests as logs showing up in stackdriver without the expected GCE / GKE annotations of instance name, cluster name, etc.

The approach from other google libraries (eg. googleapis/google-cloud-java#4269 ) is to update to the fully qualified `metadata.google.internal` or just [hit the magic IP address directly](https://github.com/googleapis/google-cloud-go/blob/master/compute/metadata/metadata.go#L39) which Istio has no problems whitelisting.

rsalmond:
Fixes #7517.

### DIFF
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -21,7 +21,7 @@ from google.cloud.logging.entries import ProtobufEntry
 from google.cloud.logging.entries import StructEntry
 from google.cloud.logging.entries import TextEntry
 
-METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/"
+METADATA_URL = "http://metadata.google.internal./computeMetadata/v1/"
 METADATA_HEADERS = {"Metadata-Flavor": "Google"}
 
 
--- a/logging/google/cloud/logging/_helpers.py
+++ b/logging/google/cloud/logging/_helpers.py
@@ -21,7 +21,7 @@ from google.cloud.logging.entries import ProtobufEntry
 from google.cloud.logging.entries import StructEntry
 from google.cloud.logging.entries import TextEntry
 
-METADATA_URL = "http://metadata/computeMetadata/v1/"
+METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/"
 METADATA_HEADERS = {"Metadata-Flavor": "Google"}
 
 


logging/google/cloud/logging/_helpers.py
```diff
@@ -21,7 +21,7 @@
 from google.cloud.logging.entries import StructEntry
 from google.cloud.logging.entries import TextEntry
 
-METADATA_URL = "http://metadata/computeMetadata/v1/"
+METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/"
```
crwilcox: If this is FQ now it may make sense for us to add a `.` so DNS doesn't try to further resolve.
```suggestion
METADATA_URL = "http://metadata.google.internal./computeMetadata/v1/"
```

rsalmond: @crwilcox I'm not sure what the effect of this will be but as long as it still resolves in GCE it should be fine. Will give it a go at work tomorrow.

crwilcox: Thanks @rsalmond. If that is the FQDN, the `.` will have no effect. That signals to DNS to not attempt to evaluate this as a subdomain but to respect this as a complete name already. It stops DNS from expanding the name across all possible search paths, which can save on lookups.

tseaver:
Thanks very much for the patch!

PR #7525, opened 2019-03-18
Promote google-cloud-speech to GA

crwilcox:


crwilcox:
merging and will let tests run on release pr.

Issue #7522, opened 2019-03-16
Bigtable: docs Snippet only includes half of documentation


sethtroisi:
Small problem with `snippets_table.py` for `google.cloud.bigtable.row.DirectRow.clear` [documentation](https://googleapis.github.io/google-cloud-python/latest/bigtable/row.html#google.cloud.bigtable.row.DirectRow.clear) only showing half the example code (e.g. not actually showing the running of the clear command)

![Screenshot from 2019-03-16 10-42-40](https://user-images.githubusercontent.com/10172976/54479283-77cde100-47d8-11e9-85e1-9441878653a4.png)

This is because the test has two parts and the snippit is only including the first. 

```
def test_bigtable_row_clear():
    # [START bigtable_row_clear]
    from google.cloud.bigtable import Client

    client = Client(admin=True)
    instance = client.instance(INSTANCE_ID)
    table = instance.table(TABLE_ID)

    row_key = b"row_key_1"
    row_obj = table.row(row_key)
    row_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, b"cell-val")
    # [END bigtable_row_clear]

    mutation_size = row_obj.get_mutations_size()
    assert mutation_size > 0

    # [START bigtable_row_clear]
    row_obj.clear()
    # [END bigtable_row_clear]

    mutation_size = row_obj.get_mutations_size()
    assert mutation_size == 0
```

tseaver:
@sethtroisi Thanks for the report!

@crwilcox, @theacodes This bug actually affetcs anywhere in Sphinx where we try to address snippets which re-use the same ID:  the [`.. literalinclude::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=literalinclude#directive-literalinclude) doesn't actually fold together the multiple segments in the same way as the DPE sample generator does.

sethtroisi:
It doesn't seem to be a very wide spread problem

```
$ grep -rih '^\s*# \[start' | sed 's/^\s*//' | sort | uniq -c | sort -n
      1 # [START acl_print]
      1 # [START acl_revoke_write]
...
      1 # [START upload_from_file]
      2 # [START bigquery_undelete_table]
      2 # [START bigtable_admins_policy]
      2 # [START bigtable_column_family_name]
      2 # [START bigtable_create_column_family]
      2 # [START bigtable_create_table]
      2 # [START bigtable_delete_cluster]
[...]

PR #7528, opened 2019-03-18
Remove duplicate tests

sethtroisi:
found with grep -ri "def test" bigtable/docs/snippets_table.py | sort | uniq -c | sort -n

these tests are duplicated inside the file

`test_bigtable_column_family_name` was at line 415 and line 491
`test_bigtable_create_update_delete_column_family` was at line 438 and line 514
`test_bigtable_row_clear` was at line 1084 and line 1239

Part of #7522


tseaver:
Thanks very much for the patch!

PR #7529, opened 2019-03-18
Add matching END tags

sethtroisi:
Unmatched tag are at line 780/781. 

It would be nice to know why this worked (so that I can be sure this fix is doing something meanful) or if I can just remove the start tags which are otherwise unused

### DIFF
--- a/spanner/tests/system/test_system.py
+++ b/spanner/tests/system/test_system.py
@@ -817,6 +817,8 @@ class TestSessionAPI(unittest.TestCase, _TestData):
 
         rows = list(session.read(self.TABLE, self.COLUMNS, self.ALL))
         self._check_rows_data(rows, [])
+        # [END spanner_test_dml_with_mutation]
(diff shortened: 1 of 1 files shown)


crwilcox:
Hi @sethtroisi , the tags are used for tracking this test in our code. The missing `END` tags means we wouldn't know the region, but we don't strictly need that information. Typically, these define regions of code to be used as samples but the `spanner_test_*` labels are just used to track existence.

sethtroisi:
Wow this is crazy, that appears to be the only unmatched tag 

`grep -hri '^\s*# \[\(END\|START\)' | sed -e 's/\s*# \[//' -e 's/]$//' | awk '{print $2, $1}' | LC_COLLATE=C sort | uniq -c | paste -d" " - - | awk '{print $1, $4, $6, $3, $2, $5}' | awk '$1 != $2 {print $0}'`

Issue #7505, opened 2019-03-12
Storage: 'tearDownModule' flakes with 409


tseaver:
From: https://source.cloud.google.com/results/invocations/b09e6e84-ca6e-427f-8014-887b811578bc/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log

```python
    def tearDownModule():
        errors = (exceptions.Conflict, exceptions.TooManyRequests)
        retry = RetryErrors(errors, max_tries=9)
>       retry(Config.TEST_BUCKET.delete)(force=True)
...
E Conflict: 409 DELETE https://www.googleapis.com/storage/v1/b/new_1552405494808: The bucket you tried to delete was not empty.
```

PR #7532, opened 2019-03-18
Bigtable: cleanup unused variable and add deprecation warnings for to be removed features

### LINKED ISSUE #7531: Bigtable: cleanup and warnings about support for backwards compatible features that will be discontinued.
Remove unused [variable](https://github.com/googleapis/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/instance.py#L33).

Add `DeprecationWarnings` for backwards compatible features that should/will go away.

- `table.yield_rows()`
- `location_id`, `serve_nodes` and `default_storage_type` parameters in `instance.create()`

AVaksman:
Fixes #7531 
* some cleanup
* add deprecation warnings for backwards compatible features that will go away

### DIFF
--- a/bigtable/google/cloud/bigtable/instance.py
+++ b/bigtable/google/cloud/bigtable/instance.py
@@ -29,11 +29,17 @@ from google.api_core.exceptions import NotFound
 
 from google.cloud.bigtable.policy import Policy
 
+import warnings
 
-_EXISTING_INSTANCE_LOCATION_ID = "see-existing-cluster"
 _INSTANCE_NAME_RE = re.compile(
     r"^projects/(?P<project>[^/]+)/" r"instances/(?P<instance_id>[a-z][-a-z0-9]*)$"
 )
+
+_INSTANCE_CREATE_WARNING = """
+Use of `instance.create({0}, {1}, {2})` will be depricated.
+Please replace with
+`cluster = instance.cluster({0}, {1}, {2})`
+`instance.create(clusters=[cluster])`."""
 
 
 class Instance(object):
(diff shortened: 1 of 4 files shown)


bigtable/google/cloud/bigtable/instance.py
```diff
@@ -276,6 +282,14 @@ def create(
         """
 
         if clusters is None:
+            warnings.warn(
+                _INSTANCE_CREATE_WARNING.format(
+                    "location_id", "serve_nodes", "default_storage_type"
+                ),
+                PendingDeprecationWarning,
```
tseaver: I think we can make this just a regular `DeprecationWarning`.  Thanks for adding `stacklevel=2`!

tseaver:
@AVaksman We need to update the unit tests to catch / assert the new warnings being raised, e.g.:

```python
    def test_foo(self):
        import warnings

        with warnings.catch_warnings(record=True) as warned:
            foo()

        self.assertEqual(len(warned), 1)
        self.assertTrue(issubclass(warned[0].category, DeprecationWarning)
```

bigtable/google/cloud/bigtable/instance.py
AVaksman: Done, thanks

AVaksman:
> @AVaksman We need to update the unit tests to catch / assert the new warnings being raised, e.g.:
> 
> ```python
>     def test_foo(self):
>         import warnings
> 
>         with warnings.catch_warnings(record=True) as warned:
>             foo()
> 
>         self.assertEqual(len(warned), 1)
>         self.assertTrue(issubclass(warned[0].category, DeprecationWarning)
> ```

Done, thanks

tseaver:
Unrelated test failures:

- The [`api_core` failure](https://source.cloud.google.com/results/invocations/8740f11f-5c76-40ca-a874-756d65817ca4/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fapi_core/log) seems clearly unrelated to this PR:  it fails in the `pytype` job.
- The [`storage` failure](https://source.cloud.google.com/results/invocations/47008c5f-f662-4544-a8b1-85ed9798b4f0/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log) is a known flaky system test, #7505.

PR #7535, opened 2019-03-19
Remove classifier for Python 3.4 for end-of-life.

crwilcox:
Python 3.4 has reached EOL. https://www.python.org/dev/peps/pep-0429/

At this point we should remove the classifier for 3.4 and discourage its use. This change doesn't prevent users from installing this library on 3.4. Doing so would require rev-ing versions of our libraries and create a good deal of noise. My suggestion is that we modify `python_requires` as other breaking changes come, removing 3.4 at that point.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -21,7 +21,7 @@ In order to add a feature to ``google-cloud-python``:
   documentation (in ``docs/``).
 
 - The feature must work fully on the following CPython versions:  2.7,
-  3.4, 3.5, and 3.6 on both UNIX and Windows.
+  3.5, 3.6, and 3.7 on both UNIX and Windows.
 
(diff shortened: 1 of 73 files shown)


tseaver:
@crwilcox 

> My suggestion is that we modify python_requires as other breaking changes come, removing 3.4 at that point.

We aren't likely to make changes which break 3.4 before we also drop 2.7 support:

- New syntax is obviously problematic (e.g. we've already told folks that `async` / `await` redesign efforts are blocked by 2.7 compat).
- New library stuff likewise (minimal in 3.5, and again blocked by 2.7 compat).

See: https://docs.python.org/3/whatsnew/3.5.html
-

crwilcox:
@tseaver we are 100% on the same page on this one. If we batch it with 2.7 it will save a major bump and be less disruptive.

PR #7545, opened 2019-03-22
Use temporary file in load_table_from_dataframe

### LINKED ISSUE #7543: BigQuery: load_table_from_dataframe should use a temporary file
[load_table_from_dataframe](https://googleapis.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.load_table_from_dataframe) currently uses BytesIO when it serializes a pandas dataframe to parquet before uploading it via a load job. This is actually violating the contract for to_parquet, which requires a filepath. BytesIO happens to work when pyarrow is used but not with fastparquet. A more minor reason we may wish to serialize to disk is that dataframe can sometimes be quite large, so spilling to disk would be preferable to filling up memory. Note: the function should clean up after itself by removing the temp file after the load job completes.

tswast:
This fixes a bug where `load_table_from_dataframe` could not be used
with the `fastparquet` library. It should also use less memory when
uploading large dataframes.

Fixes #7543

### DIFF
--- a/bigquery/docs/snippets.py
+++ b/bigquery/docs/snippets.py
@@ -31,6 +31,10 @@ import pytest
 import six
 
 try:
+    import fastparquet
+except (ImportError, AttributeError):
+    fastparquet = None
+try:
     import pandas
 except (ImportError, AttributeError):
     pandas = None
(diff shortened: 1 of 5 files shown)


alixhami:
It might be a good idea to have tests that explicitly set the parquet engine to `pyarrow` and `fastparquet` individually, since it sounds like users regularly use both.

tswast:
> It might be a good idea to have tests that explicitly set the parquet engine to pyarrow and fastparquet individually, since it sounds like users regularly use both.

@alixhami Great idea. Done in 4bca63fc0d2521bf896e0b81257db760eb94f8b7.

alixhami:
It looks like there's an error when building snappy

tswast:
I think I need to add an OS-level package (libsnappy-dev according to [StackOverflow](https://stackoverflow.com/a/20678150/101923)) to the trampoline image used for the google-cloud-python BigQuery tests. I see from the [Kokoro configs](https://github.com/googleapis/google-cloud-python/blob/d8c9fbc0350e700f7d5389d1c67550e83145a7e8/.kokoro/continuous/common.cfg#L22) that the image is `gcr.io/cloud-devrel-kokoro-resources/python-multi`. I've reached out to the Yoshi folks to see how to add to that image or use it as a base for a BigQuery-specific one.

tswast:
Trying a rebuild now that I we have snappy in the Kokoro Dockerfile (https://github.com/googleapis/testing-infra-docker/pull/7).

tswast:
`api_core` test failure appears unrelated to this change.

tswast:
The BigQuery tests do pass now.

PR #7548, opened 2019-03-22
When re-opening a ResumableBidiRPC set _request_queue_generator to None.

crwilcox:


crwilcox:
The instance member is internal and a `_` was missed here.

PR #7552, opened 2019-03-23
BigQuery: Add tqdm progress bar for downloads

JohnPaton:
As discussed [here](https://github.com/pydata/pandas-gbq/issues/182) with @tswast, this PR adds a `tqdm` progress bar for monitoring table downloads. The progress bar is updated with the downloaded number of rows after each `page` has loaded, and informs the user how many rows in total are to be downloaded. 

This PR does _not_ add `tqdm` as a dependency, it skips the progress bar if `tqdm` is not installed, or if any of the [`tqdm` errors](https://github.com/tqdm/tqdm/blob/master/tqdm/_tqdm.py#L36-L41) are raised during progress bar construction.


JohnPaton:
Signed the CLA!

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1334,21 @@ def _to_dataframe_tabledata_list(self, dtypes):
         """Use (slower, but free) tabledata.list to construct a DataFrame."""
         column_names = [field.name for field in self.schema]
         frames = []
+
+        # report progress if tqdm installed
+        try:
+            from tqdm import tqdm
+            pbar = tqdm(desc="Downloading", total=self.total_rows, unit="rows")
```
tswast: Does this do something sensible when `total_rows` is not populated? As discussed in https://github.com/googleapis/google-cloud-python/pull/7217 `total_rows` is `None` until iteration starts.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1334,21 @@ def _to_dataframe_tabledata_list(self, dtypes):
         """Use (slower, but free) tabledata.list to construct a DataFrame."""
         column_names = [field.name for field in self.schema]
         frames = []
+
+        # report progress if tqdm installed
+        try:
+            from tqdm import tqdm
```
tswast: We aim for 100% unit test coverage, so could you follow the pattern that we do with the optional pandas dependency where we import it at the top of the module and catch import errors there?

That way we can install tqdm for our unit tests but then mock it out in another test to check that it doesn't fail when the tqdm module can't be loaded.

tswast:
Looks great, thanks for the contribution! Just need to make sure we get test coverage and we're good-to-go, I think.

bigquery/google/cloud/bigquery/table.py
JohnPaton: Good catch, I'll update `pbar.total` in the loop if its unset. `tqdm` handles this just fine.

bigquery/google/cloud/bigquery/table.py
JohnPaton: Done!

JohnPaton:
I think that covers your comments @tswast, would you mind taking another look?

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1339,26 @@ def _to_dataframe_tabledata_list(self, dtypes):
...
+            except (KeyError, TypeError):
+                # tqdm error
+                pass
+
         for page in iter(self.pages):
             frames.append(self._to_dataframe_dtypes(page, column_names, dtypes))
+
+            if pbar is not None:
+                pbar.total = pbar.total or self.total_rows
+                # update progress bar with number of rows in last frame
```
tswast: This comment is unnecessary, since it just states what is happening in the next line. Comments in this codebase should state *why* or explain something that looks "wrong".

This comment indicates to me that we need to rename `pbar` to `progress_bar`. Also, we could add some intermediate variables above to make this line more self-explanatory.

```
frames.append(self._to_dataframe_dtypes(page, column_names, dtypes))

     |
turns into
     |
     V

current_frame = self._to_dataframe_dtypes(page, column_names, dtypes)
frames.append(current_frame)
...
    progress_bar.update(len(current_frame))
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1339,26 @@ def _to_dataframe_tabledata_list(self, dtypes):
...
+                )
+            except (KeyError, TypeError):
+                # tqdm error
+                pass
+
         for page in iter(self.pages):
             frames.append(self._to_dataframe_dtypes(page, column_names, dtypes))
+
+            if pbar is not None:
+                pbar.total = pbar.total or self.total_rows
```
tswast: We should add a comment here explaining why we are setting the total (again).

```
	                # In some cases, the number of total rows is not populated
	                # until the first page of rows is fetched. Update the
	                # progress bar's total to keep an accurate count.
	                pbar.total = pbar.total or self.total_rows
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1339,26 @@ def _to_dataframe_tabledata_list(self, dtypes):
...
+
+        # report progress if tqdm installed
+        pbar = None
+        if tqdm is not None:
+            try:
+                pbar = tqdm.tqdm(
+                    desc="Downloading", total=self.total_rows, unit="rows"
+                )
+            except (KeyError, TypeError):
+                # tqdm error
```
tswast: I'd like this comment to explain a little more why these errors might happen. And more importantly why we are letting them pass (because a broken progress bar shouldn't stop us from downloading results).

bigquery/tests/unit/test_table.py
```diff
@@ -1413,6 +1418,78 @@ def test_to_dataframe(self):
...
+            {"f": [{"v": "Bharney Rhubble"}, {"v": "33"}]},
+            {"f": [{"v": "Wylma Phlyntstone"}, {"v": "29"}]},
+            {"f": [{"v": "Bhettye Rhubble"}, {"v": "27"}]},
+        ]
+        path = "/foo"
+        api_request = mock.Mock(return_value={"rows": rows})
+        row_iterator = RowIterator(_mock_client(), api_request, path, schema)
+        df = row_iterator.to_dataframe()
+
+        self.assertFalse(len(df) == 0)  # all should be well
```
tswast: Since we control the API results, let's be a little more precise and use `self.assertEqual(len(df), 4)`.

bigquery/google/cloud/bigquery/table.py
JohnPaton: My thinking here was to protect from something like an interface change in `tqdm`. Indeed the fallback should just be to not show a progress bar.

bigquery/google/cloud/bigquery/table.py
JohnPaton: Will do!

bigquery/google/cloud/bigquery/table.py
JohnPaton: Awesome feedback, thank you :)

bigquery/tests/unit/test_table.py
JohnPaton: Good call, done!

tswast:
Looks great. Thanks so much for the contribution @JohnPaton . I'll kick off the tests and merge once they pass.

JohnPaton:
Awesome! Thanks for all the input, it was really constructive for me.

alixhami:
This is cool! @tswast do you know how this will affect the notebook display?

JohnPaton:
`tqdm` is a bit unpredictable in notebooks unfortunately. Ideally the progress bar just updates in stderr under the cell but if you run the same cell several times it sometimes has issues with the carriage return and each update ends up printing a new line. `tqdm` does have a special function to render the progress bar nicely in notebooks but it would require checking the context to see if we're in a notebook somehow.

JohnPaton:
Sorry hit "close" instead of "comment" there

tswast:
@alixhami Here's some GIFs.

> (1) running a python script,

script

> (2) running in interactive python

interactive-python

> (3) notebook with the function called directly,

notebook

> (4) notebook with magics

notebook-magics

I think for all but magics, it's beautiful. Magics is a little gross, but still an improvement compared to having it hang forever while downloading big results.

alixhami:
Thinking about the notebook implications of this and the in-progress project of creating a notebook testing tool that will try to run notebooks without error, would it be possible to avoid using stderr?

Also, it looks like this will always show the progress bar if the user has tqdm installed, but would it make sense to have the progress bar as an optional parameter on the `to_dataframe()` function? That way it could be used by pandas-gbq, but we could opt out of it for magics. It would also allow notebook users to avoid it if it's buggy there.

JohnPaton:
I've just found this module which I didn't previously know about: https://github.com/tqdm/tqdm/commit/97a93932742db9e2c38672219a0aef8d1b3ad482#diff-bf59de82e6ce121b5213bacf25304eb2

Looks like `tqdm` started supporting automatically doing the nice notebook progress bar display if a notebook is detected, you just need to import the function from a different submodule. There are some gifs of the nicer version [here](https://github.com/tqdm/tqdm#ipythonjupyter-integration).

What do you think @tswast, worth an update? I think it looks pretty nice actually. Difficult to test though 🙈

tswast:
The fact that the `autonotebook` module always writes a `TqdmExperimentalWarning` bugs me a bit. What if we added a `progress_bar_constructor` property to `RowIterator` that defaults to `tqdm.autonotebook` if `tqdm` is installed, but is settable to something else?

The `%%bigquery` magic code in the `magics.py` module, we add a `Context.progress_bar_constructor` property so that tqdm can be turned off for testing notebooks.

https://github.com/googleapis/google-cloud-python/blob/5756bf3b8f6cc935849f928383959386def17dd3/bigquery/google/cloud/bigquery/magics.py#L339

becomes

```
rows = query_job.result()
rows.progress_bar_constructor = context.progress_bar_constructor
result = rows.to_dataframe()
```

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1334,8 +1339,31 @@ def _to_dataframe_tabledata_list(self, dtypes):
         """Use (slower, but free) tabledata.list to construct a DataFrame."""
         column_names = [field.name for field in self.schema]
         frames = []
+
+        # report progress if tqdm installed
+        progress_bar = None
+        if tqdm is not None:
```
tswast: Let's add a `progress_bar_type` (string) parameter to `to_dataframe` and check that it also is not None.

It should default to `None` for now, until we update the magics module to support turning it off in the `Context` and also update the magics to use the `tqdm_notebook` option instead.

tswast: > It should default to None for now.

Actually, I think defaulting to `'tqdm'` is fine. When @alixhami tests notebooks, he can set `google.cloud.bigquery.table.tqdm = None` like you do in your tests.

alixhami: I think it should default to `None` because otherwise it will throw errors who don't have `tqdm` installed who aren't using the parameter.

JohnPaton:
Okay, so to summarize, we'll add an optional parameter `progress_bar_type` to `RowIterator.to_dataframe`, which will be `None` by default, but will also accept `tqdm`, `tqdm_notebook`, `tqdm_auto`, `tqdm_gui`.

I indeed don't have the linter set up, nox was giving me errors so I've just been `pytest`ing the relevant tests. If you would handle the linter that would be great.

tswast:
Good summary. Yes, I can handle the lint and coverage issues.

bigquery/google/cloud/bigquery/table.py
tswast: Fixed in the latest commits.

tswast:
Thanks a bunch for the contribution @JohnPaton. I just merged this PR, it should be available in the next release of the `google-cloud-bigquery` library.

JohnPaton:
Awesome! Happy to help :)

Issue #7558, opened 2019-03-26
BigQuery: QueryJobConfig and other job configuration classes should accept strings for dataset and table references


tswast:
Just as in the client class, properties such as [QueryJobConfig.default_dataset](https://googleapis.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.job.QueryJobConfig.html#google.cloud.bigquery.job.QueryJobConfig.default_dataset) should accept a fully-qualified string for a dataset or table reference.

tswast:
This is a follow-up to #6055 and #7477, with the goal being to remove the need to ever manually create a DatasetReference or TableReference.

PR #7560, opened 2019-03-26
Allow string for copy sources, query destination, and default dataset

tswast:
This removes a few more cases where it was required to create a
`TableReference` or `DatasetReference`.

* Allow a string for `destination` in `QueryJobConfig` .
* Allow a string for `default_dataset` in `QueryJobConfig` .
* Allow a list of strings for `sources` in `Client.copy_table`.

Closes https://github.com/googleapis/google-cloud-python/issues/7558

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -54,7 +54,6 @@ from google.cloud.bigquery.table import Table
 from google.cloud.bigquery.table import TableListItem
 from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.table import RowIterator
-from google.cloud.bigquery.table import _TABLE_HAS_NO_SCHEMA
 
 
 _DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
@@ -429,11 +428,11 @@ class Client(ClientWithProject):
         api_response = self._call_api(retry, method="GET", path=dataset_ref.path)
         return Dataset.from_api_repr(api_response)
 
-    def get_table(self, table_ref, retry=DEFAULT_RETRY):
-        """Fetch the table referenced by ``table_ref``.
+    def get_table(self, table, retry=DEFAULT_RETRY):
+        """Fetch the table referenced by ``table``.
 
         Args:
-            table_ref (Union[ \
+            table (Union[ \
                 :class:`~google.cloud.bigquery.table.Table`, \
                 :class:`~google.cloud.bigquery.table.TableReference`, \
                 str, \
@@ -449,7 +448,7 @@ class Client(ClientWithProject):
             google.cloud.bigquery.table.Table:
                 A ``Table`` instance.
         """
-        table_ref = _table_arg_to_table_ref(table_ref, default_project=self.project)
+        table_ref = _table_arg_to_table_ref(table, default_project=self.project)
         api_response = self._call_api(retry, method="GET", path=table_ref.path)
         return Table.from_api_repr(api_response)
 
@@ -964,9 +963,7 @@ class Client(ClientWithProject):
         if isinstance(source_uris, six.string_types):
             source_uris = [source_uris]
 
-        destination = _table_arg_to_table_ref(
-            destination, default_project=self.project
-        )
+        destination = _table_arg_to_table_ref(destination, default_project=self.project)
         load_job = job.LoadJob(job_ref, source_uris, destination, self, job_config)
         load_job._begin(retry=retry)
 
@@ -1042,9 +1039,7 @@ class Client(ClientWithProject):
         if location is None:
             location = self.location
 
-        destination = _table_arg_to_table_ref(
-            destination, default_project=self.project
-        )
+        destination = _table_arg_to_table_ref(destination, default_project=self.project)
         job_ref = job._JobReference(job_id, project=project, location=location)
         load_job = job.LoadJob(job_ref, None, destination, self, job_config)
         job_resource = load_job.to_api_repr()
@@ -1345,9 +1340,7 @@ class Client(ClientWithProject):
             for source in sources
         ]
 
-        destination = _table_arg_to_table_ref(
-            destination, default_project=self.project
-        )
+        destination = _table_arg_to_table_ref(destination, default_project=self.project)
 
         copy_job = job.CopyJob(
             job_ref, sources, destination, client=self, job_config=job_config
@@ -1550,10 +1543,12 @@ class Client(ClientWithProject):
             schema = selected_fields
 
         if len(schema) == 0:
-            raise ValueError((
-                "Could not determine schema for table '{}'. Call client.get_table() "
-                "or pass in a list of schema fields to the selected_fields argument."
-            ).format(table))
+            raise ValueError(
+                (
+                    "Could not determine schema for table '{}'. Call client.get_table() "
+                    "or pass in a list of schema fields to the selected_fields argument."
+                ).format(table)
+            )
 
         json_rows = [_record_field_to_json(schema, row) for row in rows]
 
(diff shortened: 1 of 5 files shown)


tswast:
CC @lbristol88

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1999,6 +2001,13 @@ def default_dataset(self):
         to use for unqualified table names in the query or :data:`None` if not
         set.
 
+        The ``dataset`` setter accepts:
```
alixhami: Is this the conventional way to document this?

bigquery/google/cloud/bigquery/job.py
```diff
@@ -1999,6 +2001,13 @@ def default_dataset(self):
...
+
+        - a :class:`~google.cloud.bigquery.dataset.DatasetReference`, or
+        - a :class:`str` of the fully-qualified dataset ID in standard SQL
+          format. The value must included a project ID and dataset ID
+          separated by ``.``.
```
alixhami: Would be nice to have an example here (e.g, `your-project.your_dataset`)

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2009,16 +2018,28 @@ def default_dataset(self):
...
     def destination(self):
         """google.cloud.bigquery.table.TableReference: table where results are
         written or :data:`None` if not set.
 
+        The ``destination`` setter accepts:
+
+        - a :class:`~google.cloud.bigquery.table.TableReference`, or
+        - a :class:`str` of the fully-qualified table ID in standard SQL
+          format. The value must included a project ID, dataset ID, and table
+          ID, each separated by ``.``.
```
alixhami: Same comment as above - add example

bigquery/google/cloud/bigquery/job.py
tswast: I guess? The closest I found is http://www.sphinx-doc.org/en/1.5/ext/example_google.html

```
    @property
    def readwrite_property(self):
        """list(str): Properties with both a getter and setter
        should only be documented in their getter method.

        If the setter method contains notable behavior, it should be
        mentioned here.
        """
        return ['readwrite_property']

    @readwrite_property.setter
    def readwrite_property(self, value):
        value
```

I guess I could try a definition list instead of a bulleted list if you'd prefer?

bigquery/google/cloud/bigquery/client.py
```diff
@@ -446,11 +447,7 @@ def get_table(self, table_ref, retry=DEFAULT_RETRY):
             google.cloud.bigquery.table.Table:
                 A ``Table`` instance.
         """
```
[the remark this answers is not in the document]
tswast: I've updated the helper to convert `Table` to `TableReference` and added `Table` to the list of acceptable types in the docstring, as we discussed offline.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -379,8 +380,8 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY):
             google.cloud.bigquery.table.Table:
                 A new ``Table`` returned from the service.
         """
-        if isinstance(table, str):
-            table = TableReference.from_string(table, default_project=self.project)
+        table = _str_or_ref_to_table_ref(table, default_project=self.project)
```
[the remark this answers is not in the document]
tswast: Renamed to `_table_arg_to_table_ref` in https://github.com/googleapis/google-cloud-python/pull/7560/commits/3ab95d4ad7da8b2be27171716e294f024882536d.

It converts more types such as `Table` to `TableReference`, but I still found it really convenient for the `copy_table` method to allow lists to pass through unchanged. The reason for that is it's actually pretty difficult to disambiguate a string from a list of stuff since they're both iterable sequences.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -653,8 +650,7 @@ def delete_table(self, table, retry=DEFAULT_RETRY, not_found_ok=False):
                 Defaults to ``False``. If ``True``, ignore "not found" errors
                 when deleting the table.
         """
-        if isinstance(table, str):
-            table = TableReference.from_string(table, default_project=self.project)
+        table = _str_or_ref_to_table_ref(table, default_project=self.project)
 
         if not isinstance(table, (Table, TableReference)):
```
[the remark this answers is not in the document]
tswast: In https://github.com/googleapis/google-cloud-python/pull/7560/commits/3ab95d4ad7da8b2be27171716e294f024882536d I now raise a type error if it's not a `TableReference`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1327,18 +1320,18 @@ def copy_table(
...
         job_ref = job._JobReference(job_id, project=project, location=location)
-
-        if isinstance(sources, str):
-            sources = TableReference.from_string(sources, default_project=self.project)
-
-        if isinstance(destination, str):
-            destination = TableReference.from_string(
-                destination, default_project=self.project
-            )
+        sources = _str_or_ref_to_table_ref(sources, default_project=self.project)
```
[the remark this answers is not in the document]
tswast: Good idea. Done in https://github.com/googleapis/google-cloud-python/pull/7560/commits/3ab95d4ad7da8b2be27171716e294f024882536d. Except it's still a little more complicated than what you propose because `isinstance(sources, collections_abc.Sequence)` returns `True` when `sources` is a string.

bigquery/google/cloud/bigquery/job.py
tswast: Done in https://github.com/googleapis/google-cloud-python/pull/7560/commits/3ab95d4ad7da8b2be27171716e294f024882536d.

bigquery/google/cloud/bigquery/job.py
tswast: Done in https://github.com/googleapis/google-cloud-python/pull/7560/commits/3ab95d4ad7da8b2be27171716e294f024882536d.

bigquery/google/cloud/bigquery/job.py
alixhami: I think in the past, we've usually just used the normal `Args:` to documment the setter args in the getter docstring

bigquery/google/cloud/bigquery/client.py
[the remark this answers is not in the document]
alixhami: I'd also change the arg name to `table` to be consistent with `create_table()` since it won't be a breaking change

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1529,19 +1531,24 @@ def insert_rows(self, table, rows, selected_fields=None, **kwargs):
         Raises:
             ValueError: if table's schema is not set
         """
-        if isinstance(table, str):
-            table = TableReference.from_string(table, default_project=self.project)
+        table = _table_arg_to_table(table, default_project=self.project)
+
+        if not isinstance(table, Table):
```
alixhami: Would it make sense to put this in `_table_arg_to_table()`?

bigquery/google/cloud/bigquery/job.py
tswast: It seems a bit odd to have "parameters" for a property setter.

image

I think I prefer the bulleted list.

bigquery/google/cloud/bigquery/client.py
tswast: Maybe, but I'd prefer to keep it consistent with `_table_arg_to_table_ref` and if we throw there we'd have to then catch the exception in a few spots, which I think is asking for trouble and less in line with Python's "duck typing" philosophy.

alixhami: Fair - I'm probably too type strict for python :D

PR #7562, opened 2019-03-26
Add methods for Models API

tswast:
The Models API adds the following methods to the BigQuery client:

* `get_model`
* `update_model`
* `delete_model`
* `list_models`

for managing models created by BigQuery ML queries. Also added are models resources, which provide additional statistics about the models resource such as details about the training run.


tswast:
Shouldn't be too much to review, as all these commits were reviewed in the private version. Looks like there are a few conflicts now, though. I'll see what I can do about those.

PR #7610, opened 2019-03-28
Add [all] extras to install all extra dependencies

tswast:
An [all] extra will be useful for convenience to get all the "bonus"
features like progress bars, pandas integration, and BigQuery Storage
API integration.

I also blacken the setup.py file and simplify the noxfile by using this
new [all] extra.

This is a pattern I've seen in other projects that have a lot of "extras", such as Apache Airflow (though they have a lot more the `google-cloud-bigquery`).
https://github.com/apache/airflow/blob/6cb735c6d301a5689ff9666349abdb6387abc7a1/setup.py#L335

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -34,8 +34,11 @@ dependencies = [
     "google-resumable-media >= 0.3.1",
 ]
 extras = {
-    "bqstorage": "google-cloud-bigquery-storage >= 0.2.0dev1, <2.0.0dev",
-    "pandas": "pandas>=0.17.1",
+    'bqstorage': [
+        'google-cloud-bigquery-storage >= 0.2.0dev1, <2.0.0dev',
+        'fastavro>=0.21.2',
+    ],
+    'pandas': 'pandas>=0.17.1',
     # Exclude PyArrow dependency from Windows Python 2.7.
     'pyarrow: platform_system != "Windows" or python_version >= "3.4"': "pyarrow>=0.4.1",
     "tqdm": "tqdm >= 4.0.0, <5.0.0dev",
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -34,11 +34,11 @@ dependencies = [
     "google-resumable-media >= 0.3.1",
 ]
 extras = {
-    'bqstorage': [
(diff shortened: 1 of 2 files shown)


bigquery/setup.py
```diff
@@ -20,79 +20,86 @@
...
-    'pyarrow: platform_system != "Windows" or python_version >= "3.4"':
-        'pyarrow>=0.4.1',
-    'tqdm': 'tqdm >= 4.0.0, <5.0.0dev',
-    'fastparquet': ['fastparquet', 'python-snappy'],
+    'pyarrow: platform_system != "Windows" or python_version >= "3.4"': "pyarrow>=0.4.1",
+    "fastparquet": ["fastparquet", "python-snappy"],
+    "tqdm": "tqdm >= 4.0.0, <5.0.0dev",
 }
 
+all_extras = []
```
tswast: This is the new bit.

bigquery/setup.py
```diff
@@ -20,79 +20,86 @@
...
+for extra in extras:
+    if isinstance(extras[extra], str):
+        all_extras.append(extras[extra])
+    else:
+        all_extras.extend(extras[extra])
+
+extras["all"] = all_extras
```
tseaver: Can't we just normalize the extras above to always be lists?

tswast:
Good point re: blackening separately. Sent https://github.com/googleapis/google-cloud-python/pull/7619

bigquery/setup.py
tswast: I fear if we miss one, then we end up with a bad dependency list.

```
>>> l = []
>>> l.extend("Uh, oh!")
>>> l
['U', 'h', ',', ' ', 'o', 'h', '!']
```

That means we still would need a check for string type or something else to validate they are all lists.

tswast: I don't what I was thinking with that comment. This is why we have tests. I assume by using the [all] extra in the nox session, we'd catch such errors. I'll simplify this on Monday as you suggest.

tswast: Done in https://github.com/googleapis/google-cloud-python/pull/7610/commits/17e65d7ce393ff8f2e41bf1d465fcd833c6d44bc.

PR #7616, opened 2019-03-29
Actually use the progress_bar_type argument in QueryJob.to_dataframe()

tswast:
Oops. Missed adding this in #7552.


tswast:
> Needs a unit test which passes a non-default `progress_bar_type`, and asserts that it is propagated.

Done in https://github.com/googleapis/google-cloud-python/pull/7616/commits/0b73e32e1c156036cad8ad2ccebabec34682cc17.

PR #7621, opened 2019-03-30
Call get_table automatically in list_rows if the schema is not available

tswast:
This is kinder than raising an error message saying to call get_table
yourself. Also, it guarantees the schema is as up-to-date as possible.

This also fixes an issue where rows could not be listed on the
TableListItem objects that are returned from list_tables.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1694,15 +1694,19 @@ class Client(ClientWithProject):
         Args:
             table (Union[ \
                 :class:`~google.cloud.bigquery.table.Table`, \
+                :class:`~google.cloud.bigquery.table.TableListItem`, \
                 :class:`~google.cloud.bigquery.table.TableReference`, \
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1741,14 +1745,11 @@ def list_rows(
...
-                (
-                    "Could not determine schema for table '{}'. Call client.get_table() "
-                    "or pass in a list of schema fields to the selected_fields argument."
-                ).format(table)
-            )
+        # No schema, but no selected_fields. Assume the developer wants all
+        # columns, so get the table resource for them rather than failing.
+        elif len(schema) == 0:
+            table = self.get_table(table.reference, retry=retry)
+            schema = table.schema
```
tseaver: Doing the extra API round-trip is certainly friendlier, but it makes the programming error more expensive (in time / billing).  I guess that is a reasonable tradeoff.

PR #7622, opened 2019-03-30
Make total_rows available on RowIterator before iteration

### LINKED ISSUE #6117: BigQuery: populate RowIterator.total_rows after running a query job

tswast:
After running a query, the total number of rows is available from the
call to the getQueryResults API. This commit plumbs the total rows
through to the faux Table created in QueryJob.results and then on
through to the RowIterator created by list_rows.

Fixes #6117.

Since this feature is more useful when `list_rows` has a full `Table` resource, this PR is based on https://github.com/googleapis/google-cloud-python/pull/7621.


bigquery/google/cloud/bigquery/table.py
```diff
@@ -1300,7 +1300,11 @@ def __init__(
         )
         self._schema = schema
         self._field_to_index = _helpers._field_to_index_mapping(schema)
+
         self._total_rows = None
+        if table is not None and hasattr(table, "num_rows"):
+            self._total_rows = table.num_rows
```
tseaver: This could simplify down to:

```python
        self._total_rows = getattr(table, "num_rows", None)
```

bigquery/tests/unit/test_job.py
```diff
@@ -4008,21 +4008,37 @@ def test_estimated_bytes_processed(self):
...
         result = job.result()
 
-        self.assertEqual(list(result), [])
+        self.assertIsInstance(result, RowIterator)
+        self.assertEqual(result.total_rows, 2)
+
+        rows = list(result)
+        self.assertEqual(len(rows), 1)
+        self.assertEqual(rows[0].col1, "abc")
+        self.assertEqual(result.total_rows, 1)
```
tseaver: Do you mean to test explicitly the case where the initial query result resource has a different number of rows than the count got by iteration?  If so, then a comment above stating that would be helpful.

tswast: Yes, that's intentional. You're right that it looks wrong. I suspect it'll be quite rare in practice, but  it can happen, especially when append query jobs are involved. Added comments in 99441d71ea1233b8bd8089c06d835298b9cff84d.

PR #7628, opened 2019-04-01
SecurityCenter: update Code Snippets.

emkornfield:
This should complete the snippets needed for the initial version of the documentation.  I'm going through now and adding them to staging, so there might be one or two that need to be refined.

### DIFF
--- a/securitycenter/docs/snippets_findings.py
+++ b/securitycenter/docs/snippets_findings.py
@@ -394,7 +394,8 @@ def test_list_all_findings(organization_id):
     # The "sources/-" suffix lists findings across all sources.  You
     # also use a specific source_name instead.
     all_sources = "{org_name}/sources/-".format(org_name=org_name)
-    for i, finding_result in enumerate(client.list_findings(all_sources)):
+    finding_result_iterator = client.list_findings(all_sources)
+    for i, finding_result in enumerate(finding_result_iterator):
         print(
             "{}: name: {} resource: {}".format(
                 i, finding_result.finding.name, finding_result.finding.resource_name
--- a/securitycenter/docs/snippets_findings.py
+++ b/securitycenter/docs/snippets_findings.py
@@ -29,7 +29,7 @@ def organization_id():
 
 @pytest.fixture(scope="module")
 def source_name(organization_id):
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
 
     client = securitycenter.SecurityCenterClient()
     org_name = "organizations/{org_id}".format(org_id=organization_id)
@@ -47,7 +47,7 @@ def source_name(organization_id):
 def test_create_source(organization_id):
     """Create a new findings source. """
     # [START create_source]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
 
     client = securitycenter.SecurityCenterClient()
     # organization_id is the numeric ID of the organization. e.g.:
@@ -68,8 +68,7 @@ def test_create_source(organization_id):
 def test_get_source(source_name):
     """Gets an existing source."""
     # [START get_source]
-    from google.cloud import securitycenter as securitycenter
-    from google.protobuf import field_mask_pb2
+    from google.cloud import securitycenter
 
     client = securitycenter.SecurityCenterClient()
 
@@ -88,7 +87,7 @@ def test_get_source(source_name):
 def test_update_source(source_name):
     """Updates a source's display name."""
     # [START update_source]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
     from google.protobuf import field_mask_pb2
 
     client = securitycenter.SecurityCenterClient()
@@ -115,7 +114,7 @@ def test_add_user_to_source(source_name):
     """Gives a user findingsEditor permission to the source."""
     user_email = "csccclienttest@gmail.com"
     # [START update_source_iam]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
     from google.iam.v1 import policy_pb2
 
     client = securitycenter.SecurityCenterClient()
@@ -157,7 +156,7 @@ def test_list_source(organization_id):
     """Lists finding sources."""
     i = -1
     # [START list_sources]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
 
     # Create a new client.
     client = securitycenter.SecurityCenterClient()
@@ -175,7 +174,7 @@ def test_list_source(organization_id):
 def test_create_finding(source_name):
     """Creates a new finding."""
     # [START create_finding]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
     from google.cloud.securitycenter_v1.proto.finding_pb2 import Finding
     from google.protobuf.timestamp_pb2 import Timestamp
 
@@ -220,7 +219,7 @@ def test_create_finding(source_name):
 def test_create_finding_with_source_properties(source_name):
     """Demonstrate creating a new finding with source properties. """
     # [START create_finding_with_properties]
-    from google.cloud import securitycenter as securitycenter
+    from google.cloud import securitycenter
     from google.cloud.securitycenter_v1.proto.finding_pb2 import Finding
     from google.protobuf.timestamp_pb2 import Timestamp
     from google.protobuf.struct_pb2 import Value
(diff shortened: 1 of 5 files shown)


busunkim96:
@engelke Would you be able to review this?

securitycenter/docs/snippets_list_assets.py
```diff
@@ -17,20 +17,24 @@
...
+    return os.environ["GCLOUD_ORGANIZATION"]
 
 
-def test_list_all_assets():
+def test_list_all_assets(organization_id):
     """Demonstrate listing and printing all assets."""
-    from google.cloud import securitycenter_v1beta1 as securitycenter
-
     # [START demo_list_all_assets]
+    from google.cloud import securitycenter as securitycenter
```
engelke: "from x import y as y" does the same thing as just "from x import y". So all these lines can just be "from google.cloud import securitycenter".

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+
+@pytest.fixture(scope="module")
+def organization_id():
+    """Get Organization ID from the environment variable """
+    return os.environ["GCLOUD_ORGANIZATION"]
+
+
+@pytest.fixture(scope="module")
+def source_name(organization_id):
+    from google.cloud import securitycenter as securitycenter
```
engelke: Just "from google.cloud import securitycenter" is enough.

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+
+    created = client.create_source(
+        org_name,
+        {
+            "display_name": "Customized Display Name",
+            "description": "A new custom source that does X",
+        },
+    )
+    print("Created Source: {}".format(created.name))
+    # [END create_source]
```
engelke: I don't see any test in here (that is, any assert statements). What will happen if this fails? It won't show up as a test failure from pytest. It may (probably will) show up as an error in the test.

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+
+
+def test_get_source(source_name):
+    """Gets an existing source."""
+    # [START get_source]
+    from google.cloud import securitycenter as securitycenter
+    from google.protobuf import field_mask_pb2
```
engelke: Not used.

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+
+    client = securitycenter.SecurityCenterClient()
+
+    # source_name is the resource path for a source that has been
+    # created previously (you can use list_sources to find a specific one).
+    # Its format is:
+    # source_name = "organizations/{organization_id}/sources/{source_id}"
+    # e.g.:
+    # source_name = "organizations/111122222444/sources/1234"
+    source = client.get_source(source_name)
```
engelke: What if this fails?

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+    # user_email is an e-mail address known to Cloud IAM (e.g. a gmail address).
+    # user_mail = user@somedomain.com
+    binding.members.append("user:{}".format(user_email))
+
+    # Setting the e-tag avoids over-write existing policy
+    updated = client.set_iam_policy(
+        source_name, {"etag": old_policy.etag, "bindings": [binding]}
+    )
+
+    print("Updated Policy: {}".format(updated))
```
engelke: No tests in this method.

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+    client = securitycenter.SecurityCenterClient()
+    # organization_id is the numeric ID of the organization. e.g.:
+    # organization_id = "111122222444"
+    org_name = "organizations/{org_id}".format(org_id=organization_id)
+
+    # Call the API and print out each existing source.
+    for i, source in enumerate(client.list_sources(org_name)):
+        print(i, source)
+    # [END list_sources]
+    assert i >= 0
```
engelke: This is an unusual way to check for a non-empty list. Maybe save the returned list first, then check its length?

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+    # [END update_source_iam]
+    assert any(
+        member == "user:csccclienttest@gmail.com"
+        for member in chain.from_iterable(
+            binding.members for binding in updated.bindings
+        )
+    )
+
+
+def test_list_source(organization_id):
```
engelke: What if nobody has set up this organization before running tests?

securitycenter/docs/snippets_findings.py
```diff
@@ -0,0 +1,603 @@
...
+    # e.g.:
+    # source_name = "organizations/111122222444/sources/1234"
+
+    # Controlled by caller.
+    finding_id = "samplefindingid"
+
+    # The resource this finding applies to.  The CSCC UI can link
+    # the findings for a resource to the corresponding Asset of a resource
+    # if there are matches.
+    resource_name = "//cloudresourcemanager.googleapis.com/organizations/11232"
```
engelke: Where does this come from?

tseaver:
@engelke 

> Are these intended to be example code snippets (in which case I suggest they belong in python-docs-samples, not here) or tests of the libraries?

That isn't consistent across the API libraries:  we host snippets here for a number of them (`bigquery`, `bigtable`, `logging`, and `storage`.

securitycenter/docs/snippets_list_assets.py
emkornfield: done.

securitycenter/docs/snippets_findings.py
emkornfield: done.

securitycenter/docs/snippets_findings.py
emkornfield: This meant to be a sample, so I think failure of test is sufficient if it doesn't work?

This is consistent with in style with other samples in this repo: https://github.com/googleapis/google-cloud-python/blob/master/bigquery/docs/snippets.py

securitycenter/docs/snippets_findings.py
emkornfield: done.

securitycenter/docs/snippets_findings.py
emkornfield: See above comment, the test fails.  Let me know if this there is something you would like me to assert or demonstrate here.

securitycenter/docs/snippets_findings.py
emkornfield: Then the tests fail.  I have a sample org that I need to enable access to python test service accounts for this repo.  Could you provide me with these and i'll can enable them?

securitycenter/docs/snippets_findings.py
emkornfield: list_sources is an iterator not a list.  On a prior PR i provided guidance not to copy to a list if it isn't recommended for clients to do (which in general I don't think it is since there can be many results).

securitycenter/docs/snippets_findings.py
emkornfield: Arbitrary name (it doesn't need to exist).

securitycenter/docs/snippets_list_assets.py
emkornfield: done (will push latest commit in a minute)

securitycenter/docs/snippets_findings.py
emkornfield: done.

securitycenter/docs/snippets_findings.py
tseaver: @engelke `snippets/noxfile.py` skips running the snippets if the `GCLOUD_ORGANIZATION` environment variable is not test.

@emkornfield The default CI account for the `google-cloud-python` repository is `precise-truck-742`.  Once that project has access to that organization, please have @busunkim96 or @crwilcox change the Kokoro setup to add that environment variable.

emkornfield: @tseaver I need to full service account associated with the project to grant access, i.e. something like foo@precise-truck-742.iam.gserviceaccount.com

tseaver: @busunkim96 can you provide the right service account to @emkornfield here?

busunkim96: @emkornfield I sent you an email with the full account name

emkornfield: @crwilcox @busunkim96 I granted access to the account can you set the environment variable?

emkornfield: ORG (1081635000895)

busunkim96: I should set `GCLOUD_ORGANIZATION` to `1081635000895`?

PR #7633, opened 2019-04-01
Fallback to BQ API when there are problems reading from BQ Storage.

tswast:
The tabledata.list API works for more kinds of tables, including small
anonymous query results tables. By falling back to this API, we enable
a developer to always specify a `bqstorage_client` even when they
aren't writing their query results to a destination table and don't
know how large their query results will be.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -35,6 +35,7 @@ try:
 except ImportError:  # pragma: NO COVER
     tqdm = None
 
+import google.api_core.exceptions
 from google.api_core.page_iterator import HTTPIterator
 
 import google.cloud._helpers
@@ -1437,7 +1438,7 @@ class RowIterator(HTTPIterator):
             bqstorage_client ( \
                 google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient \
             ):
-                **Alpha Feature** Optional. A BigQuery Storage API client. If
+                **Beta Feature** Optional. A BigQuery Storage API client. If
                 supplied, use the faster BigQuery Storage API to fetch rows
                 from BigQuery. This API is a billable API.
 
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -1496,9 +1498,16 @@ def to_dataframe(self, bqstorage_client=None, dtypes=None, progress_bar_type=Non
         progress_bar = self._get_progress_bar(progress_bar_type)
 
         if bqstorage_client is not None:
-            return self._to_dataframe_bqstorage(bqstorage_client, dtypes)
-        else:
-            return self._to_dataframe_tabledata_list(dtypes, progress_bar=progress_bar)
+            try:
```
shollyman: Two questions:
* Is the difference in progress_bar intentional?

* Should we try to be more specific about what errors trigger fallback?  I think the current exception is probably the right call, but do cases like non-enabled storage api fall under this as well?

tswast: > Is the difference in progress_bar intentional?

I haven't implemented the progress bar for BQ Storage API, yet. It'll need some additional features on the BQ Storage client to do well (expose `to_dataframe()` per block)

> Should we try to be more specific about what errors trigger fallback? 

I didn't want to be too specific, because the API currently returns "InternalError" when trying to read from small anonymous tables, which seemed like there's the potential to change in the future.

You're right that this does cach non-enabled storage API, which throws

```
 E   google.api_core.exceptions.PermissionDenied: 403 BigQuery Storage API has not been used in project before or it is disabled.
```

GoogleAPICallError is a superclass of PermissionDenied.

tswast: Now that I've started work on the tutorial, I've had a change of heart regarding catching `PermissionDenied` and other authn/authz errors. https://github.com/googleapis/google-cloud-python/issues/7661 We should rethrow those, but let the other (expected) errors get caught.

PR #7637, opened 2019-04-01
Add Translate v3

busunkim96:


busunkim96:
The Data Labeling build failed because it did not exist on the GitHub side.

This is a branch from googleapis-private.

beccasaurus:
@busunkim96 Data Labeling exists on GitHub now #7646 

Should we do a force-build to run the tests again to unblock?

PR #7640, opened 2019-04-01
Add option to use BigQuery Storage API from IPython magics

tswast:
Construct and use a BigQuery Storage API client when either the
`context.use_bqstorage_api` property is set, or the `--use_bqstorage_api`
parameter is passed into the magic.

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -150,6 +150,7 @@ class Context(object):
     def __init__(self):
         self._credentials = None
         self._project = None
+        self._use_bqstorage_api = None
 
     @property
     def credentials(self):
@@ -210,6 +211,20 @@ class Context(object):
     @project.setter
     def project(self, value):
         self._project = value
+
+    @property
+    def use_bqstorage_api(self):
+        """bool: [Beta] Set to True to use the BigQuery Storage API to
+        download query results
+
+        To use this option, install the BigQuery Storage client library,
+        install the fastavro library, and enable the BigQuery Storage API.
+        """
+        return self._use_bqstorage_api
+
+    @use_bqstorage_api.setter
+    def use_bqstorage_api(self, value):
+        self._use_bqstorage_api = value
 
 
 context = Context()
@@ -275,6 +290,16 @@ def _run_query(client, query, job_config=None):
     ),
 )
 @magic_arguments.argument(
+    "--use_bqstorage_api",
+    action="store_true",
+    default=False,
+    help=(
+        "[Beta] Use the BigQuery Storage API to download large query results. "
+        "To use this option, install the BigQuery Storage client library, "
+        "install the fastavro library, and enable the BigQuery Storage API."
+    ),
+)
+@magic_arguments.argument(
     "--verbose",
     action="store_true",
     default=False,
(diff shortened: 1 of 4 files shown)


tswast:
BigQuery tests are passing. Other failures appear to be Kokoro-induced flakes.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -274,6 +289,16 @@ def _run_query(client, query, job_config=None):
         "Standard SQL if this argument is not used."
     ),
 )
+@magic_arguments.argument(
+    "--use_bqstorage_api",
+    action="store_true",
+    default=False,
+    help=(
+        "[Beta] Use the BigQuery Storage API to download large query results. "
+        "To use this option, install the BigQuery Storage client library, "
```
alixhami: Update to clear and actionable phrasing: "install the google-cloud-bigquery-storage and fastavro libraries"
Also, add this to the docstring at the top of the file.

bigquery/tests/unit/test_magics.py
```diff
@@ -257,6 +281,48 @@ def test_bigquery_magic_clears_display_in_verbose_mode():
         assert clear_mock.call_count == 1
 
 
+@pytest.mark.usefixtures("ipython_interactive")
+@pytest.mark.skipIf(
+    bigquery_storage_v1beta1 is None, "Requires `google-cloud-bigquery-storage`"
+)
+def test_bigquery_magic_with_bqstorage(monkeypatch):
```
alixhami: Can you add a test or update an existing test asserting that the BQ storage API is not used when the new parameter is not set?

bigquery/tests/unit/test_magics.py
```diff
@@ -257,6 +281,48 @@ def test_bigquery_magic_clears_display_in_verbose_mode():
...
+    ip = IPython.get_ipython()
+    ip.extension_manager.load_extension("google.cloud.bigquery")
+    mock_credentials = mock.create_autospec(
+        google.auth.credentials.Credentials, instance=True
+    )
+    monkeypatch.setattr(magics.context, "credentials", mock_credentials)
```
alixhami: Why is monkeypatch necessary here?

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -211,6 +212,20 @@ def project(self):
     def project(self, value):
         self._project = value
 
+    @property
+    def use_bqstorage_api(self):
+        """bool: [Beta] Set to True to use the BigQuery Storage API to
+        download query results
+
+        To use this option, install the BigQuery Storage client library,
```
alixhami: Update to include package names so it's clear what to install

bigquery/tests/unit/test_magics.py
tswast: Added a comment. I'm monkeypatching the context so that it's reset after the test completes. It doesn't really matter for `context.credentials`, but it does make a difference for `context.use_bqstorage_api`.

bigquery/google/cloud/bigquery/magics.py
tswast: Updated docstrings in https://github.com/googleapis/google-cloud-python/pull/7640/commits/56bcb0569ce24154fcd830a3ec7c828d3198a2bf

PR #7663, opened 2019-04-04
Prepend leading zeros

gleeper:
`rfc3339` with nanoseconds wasn't prepending leading zeros so if an object had less than 100,000,000 nanoseconds, the nanoseconds string field would be incorrectly set.

### DIFF
--- a/api_core/google/api_core/datetime_helpers.py
+++ b/api_core/google/api_core/datetime_helpers.py
@@ -222,7 +222,7 @@ class DatetimeWithNanoseconds(datetime.datetime):
         """
         if self._nanosecond == 0:
             return to_rfc3339(self)
-        nanos = str(self._nanosecond).rstrip("0")
+        nanos = str(self._nanosecond).rjust(9, '0').rstrip("0")
         return "{}.{}Z".format(self.strftime(_RFC3339_NO_FRACTION), nanos)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


crwilcox:
@gleeper I cannot find specifics about this in the RFM specification. I took time to compare it to the implementation in Go, `time.Parse(time.RFC3339Nano, "2006-01-02T15:04:05.7010Z")` which will  format as `2006-01-02T15:04:05.701Z`

This seems to match our existing behavior.

gleeper:
I made a fix for _prepending_ zeros, not appending zeros.  `DatetimeWithNanoseconds` currently changes `2006-01-02T15:04:05.00123Z"` to `2006-01-02T15:04:05.123Z"`.  It's changing the nanoseconds field by removing leading zeros (and therefore multiplying the nanoseconds value by 10 or 100 or 1000 etc.), not trailing zeros

crwilcox:
Sorry for the confusion @gleeper and maybe I could have been clearer. From what I could tell our behavior properly matched how things were happening in other languages.

I think I understand now the issue. We have have handling on the parsing side to shift incoming values. So when you provide `.7` it becomes `700000000 nanoseconds`. While this is done also with leading 0s so `.007` becomes `7000000` we would return that as the string `.7` and not `.007` as expected.

One of your tests (the one removing trailing but adding leading) threw me a bit and sort of made me wonder what you were trying to do :)

crwilcox:
@gleeper thanks for the contribution to the project. Your change has been merged and will ship in the next release.

gleeper:
@crwilcox Happy to help!  Do you have an estimate as to when the next release will be?

PR #7664, opened 2019-04-05
remove gRPC size limit for bigquery storage

MartinNowak:
- currently limited to 4MiB
- compare to https://github.com/googleapis/google-cloud-python/pull/5594/commits/f636c8d7756a470719d114e8d77dd169e39906f3 which does the same for the monitoring API


tswast:
@crwilcox Could you review this? Seems okay once @MartinNowak signs the CLA, but I don't know if synthtool is just going to overwrite it.

Issue #7199, opened 2019-01-18
Pub/Sub publisher hangs forever if invoked with non-existed service account


kir-titievsky:
2. OS type and version: Mac OS 10.13.6
3. Python version and virtual environment information: `python --version`
4. google-cloud-<service> version:  Python 2.7.15 in virtualenv

#### Steps to reproduce

1. Create a service account with pubsub publisher permissions (easiest in Cloud Console).
2. Create and download a key for for the account: gcloud auth activate-service-account service-account@...gserviceaccount.com --key-file key.json
3. Attempt running the code below as python pub.py $PROJECT some_topic_name

End up in a hanging state -- you can't even cancel the python process, only kill -9. 

Expected behavior: the code fails immediately with a permission denied error.  Note that gcloud version of the command does exactly that. 

#### Code example
pub.py
```
import sys
  
from google.cloud import pubsub_v1


def get_callback(api_future, data):
    """Wraps message data in the context of the callback."""
    def callback(api_future):
       if api_future.exception():
            print("There was a problem with message {}".format(data))
       else:
            print("Published message {} now has message_id {}".format(
                data, api_future.result()))
    return callback

def pub(project_id, topic_name):
    """Publishes multiple messages to a Pub/Sub topic."""
    client = pubsub_v1.PublisherClient()
    topic_path = client.topic_path(project_id, topic_name)    # Data must be a bytestring

    data = "Hello, World!"
    data = data.encode('utf-8')

    # When you publish a message, the client returns a future.
    api_future = client.publish(topic_path, data=data)
    api_future.add_done_callback(get_callback(api_future, data))


if __name__ == '__main__':
    _, p, t = sys.argv
    pub(p, t)
~                                                                                                  
~                 
```

anguillanneuf:
@kir-titievsky The publisher hangs for at most 600s/10mins [by default](https://github.com/googleapis/google-cloud-python/blob/7572488283cb3c54374101de251571e86fa51a5d/pubsub/google/cloud/pubsub_v1/gapic/publisher_client_config.py#L35), then it exits with the following Deadline Exceeded exception: 
```
google.api_core.exceptions.RetryError: Deadline of 600.0s exceeded while calling
functools.partial(<function _wrap_unary_errors.<locals>.error_remapped_callable 
at 0x10864cbf8>, 
topic: "projects/global-axe-249818/topics/cron-topic"
messages {
  data: "Hello, World!"
}
[...]

PR #7670, opened 2019-04-05
ISS-7199: Fix Pub/Sub publisher hanging if service account does not exist

plamut:
This is a proposed fix for #7199.

### How to test

**Steps to reproduce:**
- Save the code snippet from #7199 to a local Python script.
- Make sure that the configured application credentials are invalid, e.g. by setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to an invalid credentials file (can use the one added by this PR).
- Run the sample script.

**Actual result:**
The script waits "forever" (it is possible to terminate it with <kbd>Ctrl</kbd> + <kbd>C</kbd>, unless if running in a debugger).

**Expected result:**
The script should fail "immediately" with an invalid credentials exception, and should not try to repeat the `publish` request that failed because of it.

**Open questions:**
- [x] ~~If the test fails because of a timeout (because the script _does_ attempt to retry the failed request), the tests do not terminate, because the retry thread does not terminate. Any canonical way around this?~~
Specifying a much shorter retry timeout for the Publisher client did the trick.
- [ ] How likely it is for the server error message to change in the future? The modified retry logic relies on that. Is there perhaps a better way?
- [ ] The Web Risk check fails even after several retries - is this flakiness or a legitimate failure? There is just a Test Fusion Url  in the build log, which I do not have access to.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
+++ b/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
@@ -15,10 +15,12 @@
 # limitations under the License.
 """Accesses the google.pubsub.v1 Publisher API."""
 
+import copy
 import functools
 import pkg_resources
 import warnings
 
+from google.api_core import exceptions as core_exceptions
 from google.oauth2 import service_account
 import google.api_core.gapic_v1.client_info
 import google.api_core.gapic_v1.config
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
```diff
@@ -188,6 +216,15 @@ def __init__(
             client_config["interfaces"][self._INTERFACE_NAME]
         )
 
+        # RPC may raise the ServiceUnavailable error, which is a retryable error.
+        # However, we do not want to retry a request if ServiceUnavailable was
+        # raised, because a Pub/Sub service account does not exist - that is an
+        # "invalid grant" error that should fail the request immediately, thus
+        # the patching.
+        for method_name, method_config in copy.copy(self._method_configs).items():
```
plamut: According to Martelli, just changing the _values_ of a dict while iterating over it is fine ([link](https://stackoverflow.com/a/2315529)).  I nevertheless opted to use `copy.copy()` anyway,  because somebody in the future might copy the pattern and use it in a loop where _not_ using a copy would actually result in an unnecessary bug.

tseaver: It might be clearer to build the value of `self._method_configs` using a dict comprehension, e.g.:

```python
method_configs = client_config["interfaces"][self._INTERFACE_NAME]
self._method_configs = {
    key: _patch_retry_predicate(method_config)
    for method_config in method_configs
}
```

tseaver:
@tseaver Per [this gitub doc](_patch_retry_predicate(method_config)), use "Closes #7199." (on a separate line) in the commit message / PR description to close the issue automagically when merging the PR.

tseaver:
This PR seems to be included / superseded by #7687, which refactors the `patch_retry_predicate` helper.  I recommend closing here.

pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
plamut: I agree, will change it on the other PR.

plamut:
Closing in favor of #7687 that already includes this, and fixes multiple related issues.

PR #7674, opened 2019-04-05
Don't mask auth errors in `to_dataframe` with BQ Storage API

### LINKED ISSUE #7661: BigQuery: Let `bqstorage_client` authentication errors be raised in `to_dataframe`
Now that I'm trying to write a tutorial on using the BigQuery Storage API to download query results, I've had a change of heart regarding whether `to_dataframe` should fall back to the `tabledata.list` API when there are authentication errors. Originally, I thought we should fallback on authentication errors, but now I see that it hides problems.

Someone could have downloaded the right libraries but either don't have the right permissions or haven't enabled the API, but those errors will be swallowed. Instead, downloading large tables will be much slower than expected.

tswast:
Don't hide errors such as insufficient permissions to create
a read session, or the API is not enabled. Both of those are
clearly problems if the developer has explicitly asked for
BigQuery Storage API support.

Closes #7661.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -1468,6 +1468,12 @@ class RowIterator(HTTPIterator):
         if bqstorage_client is not None:
             try:
                 return self._to_dataframe_bqstorage(bqstorage_client, dtypes)
+            except google.api_core.exceptions.Forbidden:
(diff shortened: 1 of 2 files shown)


tswast:
`pytype` failure in API Core seems unrelated to this change.

Issue #7654, opened 2019-04-03
BigQuery: Implement progress_bar_type support when bqstorage_client is used


tswast:
Follow-up to https://github.com/googleapis/google-cloud-python/pull/7552.

We now support displaying a progress bar in `to_dataframe` when the tabledata.list API is used, but not when the `bqstorage_client` is used. The BigQuery Storage API is faster, but not instantaneous, so it would still be useful to display a progress bar.

There are a couple of complications with this.

* When the BigQuery Storage API is used, a threadpool is used to download query results. The progress bar should only be updated from a single thread, so the threadsafe queue should probably be used from the worker threads to send updates to the main thread that updates the progress bar.
* `to_dataframe` in the BigQuery Storage API client only supports downloading a whole stream at once. The reader needs to be updated to provide a `pages` property to allow converting individual blocks to a pandas DataFrame. That way incremental updates can be sent.

PR #7680, opened 2019-04-08
Add page iterator to ReadRowsStream

tswast:
This allows readers to read blocks (called pages for compatibility with
BigQuery client library) one at a time from a stream. This enables use
cases such as progress bar support and streaming workers that expect
pandas DataFrames.

Towards https://github.com/googleapis/google-cloud-python/issues/7654

### DIFF
--- a/bigquery_storage/tests/unit/test_reader.py
+++ b/bigquery_storage/tests/unit/test_reader.py
@@ -186,8 +186,9 @@ def test_rows_w_empty_stream(class_under_test, mock_client):
         [], mock_client, bigquery_storage_v1beta1.types.StreamPosition(), {}
     )
 
-    got = tuple(reader.rows(read_session))
-    assert got == ()
+    got = reader.rows(read_session)
+    assert got.total_rows is None
+    assert tuple(got) == ()
 
 
 def test_rows_w_scalars(class_under_test, mock_client):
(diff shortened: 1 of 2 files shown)


bigquery_storage/tests/unit/test_reader.py
```diff
@@ -337,6 +412,74 @@ def test_to_dataframe_w_dtypes(class_under_test):
...
 
 
+def test_to_dataframe_by_page(class_under_test, mock_client):
+    bq_columns = [
+        {"name": "int_col", "type": "int64"},
+        {"name": "bool_col", "type": "bool"},
+    ]
+    avro_schema = _bq_to_avro_schema(bq_columns)
+    read_session = _generate_read_session(avro_schema)
+    bq_blocks_1 = [
```
shollyman: Minor nit:  the naming of the "blocks" variables reads oddly to me.  You're testing a single block with multiple rows, not multiple blocks.   It's consistent with existing tests, just seems odd.

tswast: Actually, this does contain 2 blocks. `black` just formatted it in such a way that it's not clear.

I've separated each block into its own variable and added a comment for why there are two groups of blocks. Hopefully this makes it clearer.

PR #7681, opened 2019-04-09
Translate: Add docs for v3beta1.

busunkim96:
Adds documentation for Translation client v3beta1.

cc/ @beccasaurus


language/synth.metadata
```diff
@@ -1,48 +1,11 @@
 {
-  "updateTime": "2019-04-04T12:17:35.761105Z",
```
dwsupplee: Is there a need to keep the changes in the language directory?

crwilcox: +1. It seems this was an extra change in addition to the translate work

busunkim96: Thank you!

PR #7684, opened 2019-04-09
Adding creationTime and expirationTime properties to TableListItem

### LINKED ISSUE #5117: BigQuery: Add creationTime and expirationTime properties to TableListItem
API reference: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/list

The properties `creationTime` and `expirationTime` now appear in the API results for `list_tables()`. The `creationTime` properties is exposed as `Table.created`. A similar property needs to appear on `TableListItem`.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/62740078538ea84167705c15bf109c1a2737c648/bigquery/google/cloud/bigquery/table.py#L387-L397

Likewise `Table.expires` needs a corresponding property.

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/62740078538ea84167705c15bf109c1a2737c648/bigquery/google/cloud/bigquery/table.py#L546-L556

lbristol88:
Closes #5117 

@tswast @engelke

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -907,6 +907,42 @@ class TableListItem(object):
         self._properties = resource
 
     @property
+    def created(self):
+        """Union[datetime.datetime, None]: Datetime at which the table was
+        created (:data:`None` until set from the server).
+        """
+        creation_time = self._properties.get("creationTime")
+        if creation_time is not None:
+            # creation_time will be in milliseconds.
+            return google.cloud._helpers._datetime_from_microseconds(
+                1000.0 * float(creation_time)
+            )
+
+    @property
+    def expires(self):
+        """Union[datetime.datetime, None]: Datetime at which the table will be
+        deleted.
+
+        Raises:
+            ValueError: For invalid value types.
+        """
+        expiration_time = self._properties.get("expirationTime")
+        if expiration_time is not None:
+            # expiration_time will be in milliseconds.
+            return google.cloud._helpers._datetime_from_microseconds(
+                1000.0 * float(expiration_time)
+            )
+
+    @expires.setter
+    def expires(self, value):
+        if not isinstance(value, datetime.datetime) and value is not None:
+            raise ValueError("Pass a datetime, or None")
+        value_ms = google.cloud._helpers._millis_from_datetime(value)
+        self._properties["tableReference"]["expirationTime"] = _helpers._str_or_none(
+            value_ms
+        )
+
+    @property
     def project(self):
         """str: Project bound to the table."""
         return self._properties["tableReference"]["projectId"]
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -922,9 +922,6 @@ class TableListItem(object):
     def expires(self):
         """Union[datetime.datetime, None]: Datetime at which the table will be
         deleted.
-
-        Raises:
-            ValueError: For invalid value types.
         """
         expiration_time = self._properties.get("expirationTime")
         if expiration_time is not None:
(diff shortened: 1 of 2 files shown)


bigquery/tests/unit/test_table.py
```diff
@@ -1138,6 +1147,17 @@ def test_ctor(self):
         }
 
         table = self._make_one(resource)
+
+        if "creationTime" in resource:
```
tswast: `creationTime` is in resource, as you are setting it above. This if statement is not needed. Just assert on the expected value. Same applies for `expirationTime`.

bigquery/tests/unit/test_table.py
```diff
@@ -1115,11 +1115,20 @@ def _get_target_class():
     def _make_one(self, *args, **kw):
         return self._get_target_class()(*args, **kw)
 
+    def _setUpConstants(self):
+        import datetime
+        from google.cloud._helpers import UTC
+
+        self.WHEN_TS = 1437767599.006
```
tswast: `.006` is not accurately representable in base 2, so you may encounter rounding issues with this value. Choose a fraction such as `.125` instead.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -906,6 +906,42 @@ def __init__(self, resource):
...
+            ValueError: For invalid value types.
+        """
+        expiration_time = self._properties.get("expirationTime")
+        if expiration_time is not None:
+            # expiration_time will be in milliseconds.
+            return google.cloud._helpers._datetime_from_microseconds(
+                1000.0 * float(expiration_time)
+            )
+
+    @expires.setter
```
tswast: TableListItem is read-only, so a setter is not needed.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -906,6 +906,42 @@ def __init__(self, resource):
...
+            return google.cloud._helpers._datetime_from_microseconds(
+                1000.0 * float(creation_time)
+            )
+
+    @property
+    def expires(self):
+        """Union[datetime.datetime, None]: Datetime at which the table will be
+        deleted.
+
+        Raises:
```
tswast: The `Raises` section can be removed when the setter is removed.

bigquery/tests/unit/test_table.py
lbristol88: Removed the If statements for creation an expiration, added asserts on expected value.

bigquery/tests/unit/test_table.py
lbristol88: Made change to .125 as recommended.

bigquery/google/cloud/bigquery/table.py
lbristol88: Removed the setter.

bigquery/google/cloud/bigquery/table.py
lbristol88: Removed the raises comment since setter was removed!

bigquery/tests/unit/test_table.py
```diff
@@ -1174,7 +1170,7 @@ def test_ctor(self):
         self.assertEqual(table.time_partitioning.field, "mycolumn")
         self.assertEqual(table.labels["some-stuff"], "this-is-a-label")
         self.assertIsNone(table.view_use_legacy_sql)
-        self.assertIsNone(table.expires)
+        # self.assertIsNone(table.expires)
```
tswast: Remove, don't comment out.

Actually, could you move this `assertIsNone` for `expires` and add one for  `created` to `test_ctor_missing_properties`?

tswast: Thanks for removing. Could you add assertions for `None` values to the `test_ctor_missing_properties` test, too?

lbristol88: Added assertions for created and expires under the `test_ctor_missing_properties` test.

PR #7686, opened 2019-04-10
Release the state lock before calling the publish api

sayap:
Prior to this change, any problem in the communication path to pubsub
(e.g. bad connection, slow servers, etc) would not only tie up the
calling thread itself, but also other threads waiting to get hold of
the state lock as they try to publish over the same batch.

We only need to hold the state lock for the transition from
ACCEPTING_MESSAGES / STARTING to IN_PROGRESS. After that, since only
one thread is able to transition to IN_PROGRESS, we can safely release
the state lock before calling the publish api and eventually
transitioning to SUCCESS / ERROR.

Co-authored-by: Rencana Tarigan <rtarigan@bbmtek.com>

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -193,50 +193,50 @@ class Batch(base.Batch):
                 self._status = base.BatchStatus.SUCCESS
                 return
 
-            # Begin the request to publish these messages.
-            # Log how long the underlying request takes.
-            start = time.time()
+        # Begin the request to publish these messages.
+        # Log how long the underlying request takes.
+        start = time.time()
 
-            try:
-                response = self._client.api.publish(self._topic, self._messages)
-            except google.api_core.exceptions.GoogleAPIError as exc:
-                # We failed to publish, set the exception on all futures and
-                # exit.
-                self._status = base.BatchStatus.ERROR
+        try:
+            response = self._client.api.publish(self._topic, self._messages)
+        except google.api_core.exceptions.GoogleAPIError as exc:
+            # We failed to publish, set the exception on all futures and
+            # exit.
+            self._status = base.BatchStatus.ERROR
 
-                for future in self._futures:
-                    future.set_exception(exc)
+            for future in self._futures:
+                future.set_exception(exc)
 
-                _LOGGER.exception("Failed to publish %s messages.", len(self._futures))
-                return
+            _LOGGER.exception("Failed to publish %s messages.", len(self._futures))
+            return
 
-            end = time.time()
-            _LOGGER.debug("gRPC Publish took %s seconds.", end - start)
+        end = time.time()
+        _LOGGER.debug("gRPC Publish took %s seconds.", end - start)
 
-            if len(response.message_ids) == len(self._futures):
-                # Iterate over the futures on the queue and return the response
-                # IDs. We are trusting that there is a 1:1 mapping, and raise
-                # an exception if not.
-                self._status = base.BatchStatus.SUCCESS
-                zip_iter = six.moves.zip(response.message_ids, self._futures)
-                for message_id, future in zip_iter:
-                    future.set_result(message_id)
-            else:
-                # Sanity check: If the number of message IDs is not equal to
-                # the number of futures I have, then something went wrong.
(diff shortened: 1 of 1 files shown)


sayap:
Note that the commit contains whitespace-only changes: https://github.com/googleapis/google-cloud-python/pull/7686/commits/26ac207ce355ccd58cafcad9feb1f788a2f91143?w=1

sayap:
recheck CLA

rtarigan:
I already sign the CLA

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -193,50 +193,50 @@ def _commit(self):
...
+        # Begin the request to publish these messages.
+        # Log how long the underlying request takes.
+        start = time.time()
+
+        try:
+            response = self._client.api.publish(self._topic, self._messages)
+        except google.api_core.exceptions.GoogleAPIError as exc:
+            # We failed to publish, set the exception on all futures and
+            # exit.
+            self._status = base.BatchStatus.ERROR
```
tseaver: The `self._state_lock` is designed to protect setting `self._status`:  this change undoes that protection in the case that an error occurs, which introduces a race condition AFAICT.

sayap: Hi Tres, we have covered that in the PR description:
> We only need to hold the state lock for the transition from ACCEPTING_MESSAGES / STARTING to IN_PROGRESS. After that, since only one thread is able to transition to IN_PROGRESS, we can safely release the state lock before calling the publish api and eventually transitioning to SUCCESS / ERROR.

Maybe you can elaborate on the race condition? Thanks

sduskis:
@plamut, can you please take a look at this?

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
plamut: TL; DR - Not holding the lock after changing the state to IN_PROGRESS indeed seems safe. However, I would still like to see the effect of this change covered by a test before merging. I can help with this, if needed.

---
**Long reply:**

I checked the Batch code, and from what I can tell, that the following threads are relevant:
- `MonitorBatchPublisher` (if autcommit == True): Triggers `_commit()` after `max_latency` time elapses.
- `CommitBatchPublisher`: If somebody calls `commit()` and the batch is in ACCEPTING_MESSAGES state, `_commit()` gets called in this thread.
- User thread(s): When calling `publisher.publish()`, `batch.publish()` gets called. If that causes an overflow, `commit()` is called in the `CommitBatchPublisher` thread.

The first thing that `_commit()` does after acquiring the lock is changing the batch status to IN_PROGRESS.

In that state, any further calls to `publish()` [become a no-op](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L280-L282), because the `will_accept()` check [returns `False`](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/_batch/base.py#L123-L124) for any state that is not ACCEPTING_MESSAGES.

Additionally, any other `_commit()` calls also [become a no-op](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L181-L188), because the batch status is now IN_PROGRESS.

Only the thread that changed the state to IN_PROGRESS can proceed, thus shielding the rest of the `_commit()` method with `_state_lock` indeed seems unnecessary, which includes the call to `self._client.api.publish()`.

_Without_ the `_state_lock` held, further calls to `batch.publish()` will not block longer than needed. The batch will simply not accept a new message, return `False`, and `publisher.publish() `will create an entirely [new batch](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py#L236-L238) as a result.

sduskis:
@rtarigan, this PR is a bit stale and needs tests.  Please reopen this PR once the tests are in place.

plamut:
@rtarigan Thank you for the fix! I opened another PR (#8234) that also includes tests, but mentioned you and Rencana Tarigan as co-authors in the commit message, as the fix is essentially the same as the one here.

PR #7687, opened 2019-04-10
Fix Pub/Sub clients stalling if service account does not exist

### LINKED ISSUE #7199: Pub/Sub publisher hangs forever if invoked with non-existed service account

### LINKED ISSUE #7419: PubSub: Subscriber hangs when passed credentials kwarg with invalid type
When using the `google-auth` library and passing `credentials=google.auth.default()` to `pubsub_v1.SubscriberClient` the subscriber hangs. No errors are shown, and the callback is never triggered. I made a simplified example below that demonstrates the issue. In the example code, if the subscription doesn't exist, or if the user/serviceAccount doesn't have permissions to subscribe to the topic, the code just hangs at `future.result()`. If I remove the `credentials` kwarg when I create the subscriber, the appropriate exceptions are raised and the code exits as expected. If there are no issues preventing the user/serviceAccount from subscribing to the topic, and there are messages waiting to be acked, it also hangs and seemingly does nothing. Again, if I remove the credentials kwarg, it starts receiving messages, and calling the callback.

I realize this is kind of a silly example, since it uses the application default credentials when no credentials are explicitly passed anyway. I wrote my code that way to make it easy to share with others with a comment that you could plug-in credentials using whatever method you preferred.

```
import google.auth


def callback(message):
    print(message)


if __name__ == "__main__":

    app_creds = google.auth.default()

    subscriber = pubsub_v1.SubscriberClient(credentials=app_creds)
    subscription_path = subscriber.subscription_path(
        'myproject', 'mysubscription')

    future = subscriber.subscribe(
        subscription_path,
        callback=callback
    )

    print("Listening for pubsub messages on {}...".format(subscription_path))
    future.result()
```

### LINKED ISSUE #7587: Pubsub:  subscribing with Invalid JSON key hangs.
Thanks for stopping by to let us know something could be better!

If you are still having issues, please be sure to include as much information as possible:

Issue Summary: 
We're following the example at [1], and we're running the test script with an invalid/deleted service account JSON key but the script would keep running indefinitely and never returns an auth exception.

[1] https://cloud.google.com/pubsub/docs/pull#listening-for-errors

#### Environment details

1. Specify the API at the beginning of the title (for example, "BigQuery: ...")
   General, Core, and Other are also allowed as types
2. OS type and version: Google Cloud Chell
3. Python version and virtual environment information: Python 2.7.13
4. google-cloud-<service> version:  google-cloud-pubsub-0.40.0

#### Steps to reproduce

  1. Try to run the following test script of async pull from a subscription with an service account Json key, then delete the key but it would never return any auth exception; however for sync pull from the subscription would throw auth exception after 10 mins which is expected: Deadline of 600.0s exceeded while calling <functools.partial object at 0x7f7767d04c00>, last exception: 503 Getting metadata from plugin failed with error: ('invalid_grant: Invalid JWT Signature.', u'{\n "error": "invalid_grant",\n "error_description": "Invalid JWT Signature."\n}')

#### Code example - Async

```python
import time
import json
from google.oauth2 import service_account
from google.api_core.exceptions import NotFound

project_id = "PROEJCT_ID"
subscription_name = "SUBSCRIPTION_NAME"


def callback(message):
    print('Received message: {}'.format(message))
    print('Error? {}'.format(future.exception()))
    message.ack()

def sub_loop_contents():
    service_account_info = json.load(open('service_account_key.json'))
    credentials = service_account.Credentials.from_service_account_info(service_account_info)
    subscriber = pubsub_v1.SubscriberClient(credentials=credentials)
    # The `subscription_path` method creates a fully qualified identifier
    # in the form `projects/{project_id}/subscriptions/{subscription_name}`
    subscription_path = subscriber.subscription_path(project_id, subscription_name)

    print('Listening for messages on {}'.format(subscription_path))

    # NOTE. If we call this before the subscription is properly created, we will never get any messages.
    # This is why we need the subscription.get permission.
    future = subscriber.subscribe(subscription_path, callback=callback)

    try:
        # Block thread waiting for results.
        future.result()
    except Exception as e:
        print('Listening for messages on {} threw an Exception: {}.'.format(subscription_name, e))
        future.cancel()

if __name__ == '__main__':
    sub_loop_contents()

```

#### Stack trace
```
Nothing as it would just hang there
```
#### Code example - Sync
```python
import time
import json
from google.oauth2 import service_account
from google.api_core.exceptions import NotFound


project_id = "PROJECT_ID"
subscription_name = "SUBSCRIPTION_NAME"



def sub_loop_contents():
 
    service_account_info = json.load(open('SERVICE_ACCOUNT_KEY.json'))
    credentials = service_account.Credentials.from_service_account_info(service_account_info)    
    subscriber = pubsub_v1.SubscriberClient(credentials=credentials)
    
    # The `subscription_path` method creates a fully qualified identifier
    # in the form `projects/{project_id}/subscriptions/{subscription_name}`
    subscription_path = subscriber.subscription_path(project_id, subscription_name)

    print('Listening for messages on {}'.format(subscription_path))
    


    # NOTE. If we call this before the subscription is properly created, we will never get any messages.
    # This is why we need the subscription.get permission.
    # future = subscriber.subscribe(subscription_path, callback=callback)
    
    while True:
        try:
            response = subscriber.pull(subscription_path, 100)
            print(response)
        except Exception as ex:
            print(ex)
            break


if __name__ == '__main__':
    
    sub_loop_contents()
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

plamut:
Closes #7199.
Closes #7419.
Closes #7587.

### How to test

(for the steps to test the _publisher_ client, please refer to #7670)

**Steps to reproduce:**
- Save the code snippets from #7587 (both the sync and the async version of it)
- Make sure that the configured application credentials are invalid, e.g. by setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to an invalid credentials file (can use `project-credentials-invalid.json` added by this PR).
- Run both sample scripts.

**Actual result:**
Both scripts wait "forever" (i.e. until the configured maximum timeout).

**Expected result:**
Both scripts should fail "immediately" with an invalid credentials exception, and should not try to repeat the request (`pull()` or `subscribe()`) that failed because of it.

**Open questions:**
- [ ] How likely it is for the server error message to change in the future? The modified retry logic relies on that. Is there perhaps a better way?

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/gapic/custom_retry.py
+++ b/pubsub/google/cloud/pubsub_v1/gapic/custom_retry.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.api_core import exceptions as core_exceptions
+
+
+def patch_retry_predicate(method_config):
+    """Patch method config to *not* retry on service accout missing errors.
+
+    If Retry is None in the config, the config is returned unmodified.
+
+    Args:
+        method_config (:class:`google.api_core.gapic_v1.config.MethodConfig`):
+            The method config to patch.
+
+    Returns:
+        :class:`google.api_core.gapic_v1.config.MethodConfig`: The modified config.
+    """
+    if method_config.retry is None:
+        return method_config  # there's nothing to patch
+
+    orig_predicate = method_config.retry._predicate
+
+    def should_retry(exc):
+        if isinstance(exc, core_exceptions.ServiceUnavailable):
+            return "invalid_grant" not in exc.message
+        return orig_predicate(exc)
+
+    new_retry = method_config.retry.with_predicate(should_retry)
+    return method_config._replace(retry=new_retry)
--- a/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
+++ b/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
@@ -30,7 +30,6 @@ import google.api_core.page_iterator
 import google.api_core.path_template
 import grpc
 
-from google.cloud.pubsub_v1.gapic import custom_retry
 from google.cloud.pubsub_v1.gapic import publisher_client_config
 from google.cloud.pubsub_v1.gapic.transports import publisher_grpc_transport
 from google.cloud.pubsub_v1.proto import pubsub_pb2
@@ -185,19 +184,9 @@ class PublisherClient(object):
         # from the client configuration.
         # (Ordinarily, these are the defaults specified in the `*_config.py`
         # file next to this one.)
-        method_configs = google.api_core.gapic_v1.config.parse_method_configs(
+        self._method_configs = google.api_core.gapic_v1.config.parse_method_configs(
             client_config["interfaces"][self._INTERFACE_NAME]
         )
-
-        # RPC may raise the ServiceUnavailable error, which is a retryable error.
-        # However, we do not want to retry a request if ServiceUnavailable was
-        # raised, because a Pub/Sub service account does not exist - that is an
-        # "invalid grant" error that should fail the request immediately, thus
-        # the patching.
-        self._method_configs = {
-            method_name: custom_retry.patch_retry_predicate(method_config)
-            for method_name, method_config in method_configs.items()
-        }
 
         # Save a dictionary of cached API call functions.
         # These are the actual callables which invoke the proper
--- a/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
+++ b/pubsub/google/cloud/pubsub_v1/gapic/publisher_client.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 """Accesses the google.pubsub.v1 Publisher API."""
 
-import copy
 import functools
 import pkg_resources
 import warnings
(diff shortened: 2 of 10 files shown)


tseaver:
On its face, the fix here seems OK:  I agree that depending on the `invalid_grant` text in the `SERVICE_UNAVAILABLE` reason field is ugly, but might be the only way forward.

However, the two files which *use* the new `patch_retry_predicate` helper are both generate by GAPIC, which means that the fix will only last until the next autogen pass.  Making the fix persistent would require one of (in order of my preference):

- asking the back-end team to return a more appropriate / detectable status code (this should really be a `403 FORBIDDEN``, as it is an error in the client request).
- updating the non-GAPIC client constructors to reach into the GAPIC api client's internals and replace the retry handling on the appropriate methods.
- hacking `synth.py` to add the changes to the newly-generated filed.
- asking the upstream `gapic-generator` project to include this ugly fix.

plamut:
@tseaver Thanks, did not know that `gapic/*er_client.py` files are both generated. I'll check with the backend team if we can get a more detectable error (and when), otherwise we'll proceed with the next-best solution.

Is the `Client` class safe to modify, i.e. patching `_method_configs` of the wrapped [self.api](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py#L96), instead of patching them in the `*erClient` class directly?

We could still take this route in the transitional period, if it turns out that the update from the backend team cannot be expected "soon enough" for our needs.

tseaver:
@sduskis Did you receive guidance from the back-end about how we should approach fixing #7199?

PR #7690, opened 2019-04-10
Pub/Sub ThreadScheduler should inherit from the base Scheduler ABC

plamut:
This PR makes the `ThreadScheduler` to subclass the base `Scheduler` (and not `object`).

### How to test
Carefully gaze at the `scheduler.ThreadScheduler` class definition for about 3.14 seconds and await _the_ Enlightenment. :wink: 

I see no apparent reason, why `ThreadScheduler` subclasses `object` directly, instead of the ABC defined in the very same file, as the interfaces of these two classes match.


plamut:
@tseaver What does the rotating light label warn about? Is this PR safe to merge?

sduskis:
@plamut, the rotating light indicates that this PR is old and should get additional attention from reviewers.  Please feel free to merge.

plamut:
@sduskis Thank you, will happily do it.

PR #7697, opened 2019-04-12
BigQuery: `to_dataframe` respects `progress_bar_type` when used with BQ Storage API

### LINKED ISSUE #7654: BigQuery: Implement progress_bar_type support when bqstorage_client is used

tswast:
When the BigQuery Storage API was used in `to_dataframe`, the `progress_bar_type` argument was ignored. This PR fixes that by creating a concurrent queue that worker threads can send progress updates to.

A fake queue is created for the case with no progress bar to prevent filling a queue but never reading from it.

This fix depends on:

* ~https://github.com/googleapis/google-cloud-python/pull/7680 adds `.pages` to BQ Storage. Must be merged and released.~ Released in `google-cloud-bigquery-storage` version 0.4.0!
* ~https://github.com/googleapis/google-cloud-python/pull/7698 adds loop over `.pages` to fix issue with `KeyboardInterrupt` (discovered while working on this progress bar feature).~

Closes #7654

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -25,6 +25,7 @@ import operator
 import warnings
 
 import six
+from six.moves import queue
 
 try:
     from google.cloud import bigquery_storage_v1beta1
@@ -66,7 +67,7 @@ _NO_TQDM_ERROR = (
 )
 _TABLE_HAS_NO_SCHEMA = 'Table has no schema:  call "client.get_table()"'
 _MARKER = object()
-_PROGRESS_INTERVAL = 1.0  # Time between download status updates, in seconds.
+_PROGRESS_INTERVAL = 0.2  # Time between download status updates, in seconds.
 
 
 def _reference_getter(table):
@@ -1274,6 +1275,16 @@ class Row(object):
         return "Row({}, {})".format(self._xxx_values, f2i)
 
 
+class _FakeQueue(object):
+    """A fake Queue class that does nothing.
+
+    This is used when there is no progress bar to send updates to.
+    """
+
+    def put_nowait(self, item):
+        """Don't actually do anything with the item."""
+
+
 class RowIterator(HTTPIterator):
     """A class for iterating through HTTP/JSON API row list responses.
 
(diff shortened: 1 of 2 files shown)


tswast:
Screenshots of this progress bar in action:

Terminal (IPython):

bqstorage_tqdm

Notebook:

bqstorage_tqdm_notebook

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1408,7 +1426,23 @@ def _to_dataframe_bqstorage_stream(
         # the end using manually-parsed schema.
         return pandas.concat(frames)[columns]
 
-    def _to_dataframe_bqstorage(self, bqstorage_client, dtypes):
+    def _process_progress_updates(self, progress_queue, progress_bar):
```
shollyman: Does this work well for large tables as the number of updates grows, since you're potentially walking it every _PROGRESS_INTERVAL seconds?

tswast: I decreased `_PROGRESS_INTERVAL` to account for the fact that we get way too many updates in a second, but you're right that with large tables (many streams) this becomes worse.

Originally I tried having a constant loop of updates for `tqdm` but the 

Right now I handle this by very likely dropping updates when the queue fills up, meaning the progress bar is grossly inaccurate for large tables. Maybe what I should do instead is have 2 queues. 🤔 One queue used by the workers and a "sum" thread to add up and send to main thread every `_PROGRESS_INTERVAL` via a different queue.

shollyman: I suppose the problem is also bounded by the max dataframe users can fit in ram.  Our truly large tables aren't really going to be pulled into dataframes without excessive projection or filtering.  Perhaps this is a non issue for now.

PR #7698, opened 2019-04-12
BigQuery: ensure that `KeyboardInterrupt` during `to_dataframe`no longer hangs.

tswast:
I noticed in manually testing `to_dataframe` that it would stop
the current cell when I hit Ctrl-C, but data kept on downloading in the
background. Trying to exit the Python shell, I'd notice that it would
hang until I pressed Ctrl-C a few more times.

Rather than get the DataFrame for each stream in one big chunk, loop
through each block and exit if the function needs to quit early. This
follows the pattern at https://stackoverflow.com/a/29237343/101923

This depends on:

* ~https://github.com/googleapis/google-cloud-python/pull/7680, new `.pages` feature in BQ Storage client~
* https://github.com/googleapis/google-cloud-python/pull/7726, aligns optional import for BQ Storage client with pattern used elsewhere in the client.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -66,6 +66,7 @@ _NO_TQDM_ERROR = (
 )
 _TABLE_HAS_NO_SCHEMA = 'Table has no schema:  call "client.get_table()"'
 _MARKER = object()
+_PROGRESS_INTERVAL = 1.0  # Time between download status updates, in seconds.
 
 
 def _reference_getter(table):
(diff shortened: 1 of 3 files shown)


bigquery/tests/unit/test_table.py
```diff
@@ -1751,7 +1801,21 @@ def test_to_dataframe_w_bqstorage_nonempty(self):
...
         bqstorage_client.create_read_session.return_value = session
+
+        mock_rowstream = mock.create_autospec(reader.ReadRowsStream)
         bqstorage_client.read_rows.return_value = mock_rowstream
+
+        mock_rows = mock.create_autospec(reader.ReadRowsIterable)
+        mock_rowstream.rows.return_value = mock_rows
+
+        mock_page = mock.create_autospec(reader.ReadRowsPage)
+        mock_page.to_dataframe.return_value = pandas.DataFrame(
```
tswast: TODO: add delay and reset sleep time to be slower, like in keyboard interrupt test. That will verify futures.concurrent.wait loop.

tswast:
Need to update the minimum `google-cloud-bigquery-storage` dependency version after the release at https://github.com/googleapis/google-cloud-python/pull/7716 is complete.

tswast:
FYI: I pulled the `ValueError` on missing `google-cloud-bigquery-storage` package to a separate PR https://github.com/googleapis/google-cloud-python/pull/7726. That PR should merge before this one.

bigquery/google/cloud/bigquery/table.py
```diff
@@ -1402,17 +1443,46 @@ def _to_dataframe_bqstorage(self, bqstorage_client, dtypes):
         if not session.streams:
             return pandas.DataFrame(columns=columns)
 
-        def get_dataframe(stream):
-            position = bigquery_storage_v1beta1.types.StreamPosition(stream=stream)
-            rowstream = bqstorage_client.read_rows(position)
-            return rowstream.to_dataframe(session, dtypes=dtypes)
+        # Use _to_dataframe_finished to notify worker threads when to quit.
+        # See: https://stackoverflow.com/a/29237343/101923
+        self._to_dataframe_finished = False
```
shollyman: Mostly I'm trying to reason about scope here.  This is a worker pool, but is it possible we're generating multiple independent dataframes that would share the same access to _to_dataframe_finished?  Do we need to key the workers to specific invocations, or is the nature of the access always blocking so that this isn't an issue?

tswast: There are a couple reasons that I don't think it's an issue:

* Yes, `to_dataframe()` is a blocking call, so you wouldn't really have multiple going at once.
* `RowIterator` isn't really something you'd want to use across threads or even more than once anyway. Because of how the pagination state works, once you loop through all the rows once, `to_dataframe()` returns an empty DataFrame, even in the case where BQ Storage API isn't used.

PR #7705, opened 2019-04-15
Pub/Sub (nit): wrong var name in sample

anguillanneuf:


sduskis:
Would this have been caught with better testing?

anguillanneuf:
@sduskis This code snippet is in the docstring. It's probably why we didn't catch it earlier.

PR #7710, opened 2019-04-15
Adding missing properties from LoadJobConfig to LoadJob library

### LINKED ISSUE #7482: BigQuery: Add missing configuration properties to LoadJob
There are several properties in the load job configuration that would be useful to expose in the LoadJob object. For example, `configuration.load.destinationTable` should be exposed as a TableReference property.

Perhaps we should just expose a generic "configuration" property so that we get all of the properties and reflect the API resource?

lbristol88:
Closes #7482

@tswast @engelke

### DIFF
--- a/bigquery/tests/unit/test_job.py
+++ b/bigquery/tests/unit/test_job.py
@@ -1753,6 +1753,24 @@ class TestLoadJob(unittest.TestCase, _Base):
         else:
             self.assertIsNone(job.destination_encryption_configuration)
 
+        if "destinationTableDescription" in config:
+            self.assertIsNotNone(job.destination_table_description)
+            self.assertEqual(
+                job.destination_table_description.description,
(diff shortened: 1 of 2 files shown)


bigquery/tests/unit/test_job.py
```diff
@@ -1753,6 +1753,24 @@ def _verifyResourceProperties(self, job, resource):
         else:
             self.assertIsNone(job.destination_encryption_configuration)
 
+        if "destinationTableDescription" in config:
```
tswast: Sorry, `_verifyResourceProperties` isn't a very good example to follow. I prefer explicitly setting the values you are looking for in the test itself.

Please remove this code and add check for string values for `destination_table_description` and `destination_table_friendly_name` to `test_ctor_w_config`.

Issue #6924, opened 2018-12-13
Firestore: Using offset greater than 1000 on a query doesn't work


KiwiPolarBear:
#### Environment details

macOS Mojave 10.14.1
Python 3.6
google-cloud-firestore==0.30.0

#### Steps to reproduce

  1. Create a query to get a large collection (over 1000 documents)
  2. Add an offset greater than 1000 to the query
  3. Get the query and iterate over the documents
  4. You should get an exception about having no documents

#### Code example

```python
from google.cloud import firestore

client = firestore.Client()

documents = client.collection('Collection').order_by('field').offset(1001).get()

for document in documents:
    pass
```

#### Stack trace
```
ValueError: Unexpected server response. All responses other than the first must contain a document. The response at index 1 was
read_time {
  seconds: 1544736718
  nanos: 109409000
}
skipped_results: 1
.
```

tseaver:
Can you please show the whole traceback?  Also, are you sure there are at least 1000 documents in the collection?

KiwiPolarBear:
Sure.

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Source/....../scripts/.......py", line 21, in <module>
    for document in documents:
  File "/Source/......../venv/lib/python3.6/site-packages/google/cloud/firestore_v1beta1/query.py", line 599, in get
    raise ValueError(msg)
ValueError: Unexpected server response. All responses other than the first must contain a document. The response at index 1 was
read_time {
  seconds: 1544741917
  nanos: 761441000
}
skipped_results: 1
.
```

I am certain there are more than 1000 documents as making the same query with an offset of 1000, returns more documents.

tseaver:
There is no indication in the [`StructuredQuery` API docs](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1beta1#google.firestore.v1beta1.StructuredQuery) of any upper bound on `limit`, and the `google-cloud-firestore` code doesn't do anything particular with the value beyond copying it into that message.

@shollyman Can you shed light on why the server would be returning a `RunQueryResponse` here with no `document` set?  If it is expected, what is the expected behavior within the client library?

[...]

PR #7758, opened 2019-04-19
Firestore: add support for CollectionGroup queries.

crwilcox:
CollectionGroup queries add support for queries to Firestore that query documents by collection ID rather than by collection path. 

Java: googleapis/google-cloud-java#4652
Node: googleapis/nodejs-firestore#578

### DIFF
--- a/firestore/google/cloud/firestore_v1/query.py
+++ b/firestore/google/cloud/firestore_v1/query.py
@@ -430,6 +430,7 @@ class Query(object):
             "orders": self._orders,
             "limit": self._limit,
             "offset": self._offset,
+            "all_descendants": self._all_descendants,
         }
         if start:
             query_kwargs["start_at"] = cursor_pair
--- a/firestore/google/cloud/firestore_v1/query.py
+++ b/firestore/google/cloud/firestore_v1/query.py
@@ -142,7 +142,7 @@ class Query(object):
         self._offset = offset
         self._start_at = start_at
         self._end_at = end_at
-        self._all_descendants=all_descendants
+        self._all_descendants = all_descendants
 
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
@@ -690,8 +690,7 @@ class Query(object):
             "select": projection,
             "from": [
                 query_pb2.StructuredQuery.CollectionSelector(
-                    collection_id=self._parent.id,
-                    all_descendants=self._all_descendants,
+                    collection_id=self._parent.id, all_descendants=self._all_descendants
                 )
             ],
             "where": self._filters_pb(),
--- a/firestore/google/cloud/firestore_v1/query.py
+++ b/firestore/google/cloud/firestore_v1/query.py
@@ -111,6 +111,10 @@ class Query(object):
             any matching documents will be included in the result set.
             When the query is formed, the document values
             will be used in the order given by ``orders``.
+        all_descendants (Optional[bool]): When false, selects only collections
+            that are immediate children of the `parent` specified in the
+            containing `RunQueryRequest`. When true, selects all descendant
+            collections.
     """
 
     ASCENDING = "ASCENDING"
@@ -128,7 +132,7 @@ class Query(object):
         offset=None,
         start_at=None,
         end_at=None,
-        all_descendants=None,
+        all_descendants=False,
     ):
         self._parent = parent
         self._projection = projection
@@ -205,6 +209,7 @@ class Query(object):
             offset=self._offset,
             start_at=self._start_at,
             end_at=self._end_at,
+            all_descendants=self._all_descendants,
         )
 
     def where(self, field_path, op_string, value):
@@ -272,6 +277,7 @@ class Query(object):
             offset=self._offset,
             start_at=self._start_at,
             end_at=self._end_at,
+            all_descendants=self._all_descendants,
         )
 
     @staticmethod
@@ -323,6 +329,7 @@ class Query(object):
             offset=self._offset,
             start_at=self._start_at,
             end_at=self._end_at,
+            all_descendants=self._all_descendants,
         )
 
     def limit(self, count):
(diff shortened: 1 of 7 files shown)


firestore/google/cloud/firestore_v1/query.py
```diff
@@ -968,3 +985,32 @@ def _query_response_to_snapshot(response_pb, collection, expected_prefix):
         update_time=response_pb.document.update_time,
     )
     return snapshot
+
+
+def _collection_group_query_response_to_snapshot(response_pb, collection):
+    """Parse a query response protobuf to a document snapshot.
```
schmidt-sebastian: I am a little surprised that this is a different method from `_query_response_to_snapshot`.

firestore/google/cloud/firestore_v1/query.py
```diff
@@ -968,3 +985,32 @@ def _query_response_to_snapshot(response_pb, collection, expected_prefix):
...
+        collection (~.firestore_v1.collection.CollectionReference): A
+            reference to the collection that initiated the query.
+
+    Returns:
+        Optional[~.firestore.document.DocumentSnapshot]: A
+        snapshot of the data returned in the query. If ``response_pb.document``
+        is not set, the snapshot will be :data:`None`.
+    """
+    if not response_pb.HasField("document"):
+        return None
```
schmidt-sebastian: Does Python not return empty DocumentSnapshots for non-existing documents? In either case, this is a query and this condition should never trigger (fingers crossed).

firestore/tests/system.py
```diff
@@ -797,6 +797,11 @@ def on_snapshot(docs, changes, read_time):
             "Expected the last document update to update born: " + str(on_snapshot.born)
         )
 
+def test_collection_group_query(client, cleanup):
```
schmidt-sebastian: Can we port one of the tests from Node/Java and verify the query result?

firestore/google/cloud/firestore_v1/query.py
crwilcox: This was related to expectations built into this method. @tseaver made this to address that. It is possible we can combine them by refactoring that code.

danoscarmike:
@tseaver can you give this one your blessing too? Thanks!

tseaver:
@danoscarmike I will look at the branch, and get the tests / coverage passing.

firestore/google/cloud/firestore_v1/query.py
tseaver: @schmidt-sebastian The implementation of `_query_response_to_snapshot` [mandates (via `_helpers.get_doc_id`)](https://github.com/googleapis/google-cloud-python/pull/7758/files#diff-56838955ee957b7f17f42e9e0d3925b0L959) that the result document is a direct child of the query's parent collection.  I'm unsure whether this restriction is "important", or just an artifact of prior development iterations.  If it is an artifact, then we could just  use the new implementation in `_collection_group_query_response_to_snapshot` for all queries.

firestore/google/cloud/firestore_v1/query.py
tseaver: @schmidt-sebastian [`DocumentReference.get`](https://github.com/googleapis/google-cloud-python/blob/3618354c5fdd388e06b0b419cbcbe613cb6ad879/firestore/google/cloud/firestore_v1/document.py#L408-L468) does return an empty snapshot for a non-existent document.  For the query case, the back-end can return a response without a `document` when reporting partial results:  see #6924 and PR #7206 for the history.

schmidt-sebastian: My bad: It looks like the expected behavior is that it will get skipped (see https://cs.corp.google.com/piper///depot/google3/google/firestore/v1beta1/firestore.proto?q=runqueryrequest+firestore&dr=CSs&l=619)

crwilcox:
The backend has deployed. So removing the 'do not merge'

tseaver:
@schmidt-sebastian I updated the "start at / end at" test as you requested in 115234c.  However, for the "filters" test, when I change to filter on `__name__` (which would match the Node test), the following query:

```python
(Pdb) pp query._to_protobuf()
from {
  collection_id: "b-1556570226919"
  all_descendants: true
}
where {
  composite_filter {
    op: AND
    filters {
      field_filter {
        field {
          field_path: "__name__"
        }
        op: GREATER_THAN_OR_EQUAL
        value {
          string_value: "a/b"
        }
      }
    }
    filters {
      field_filter {
        field {
          field_path: "__name__"
        }
        op: LESS_THAN_OR_EQUAL
        value {
          string_value: "a/b0"
        }
      }
    }
  }
}
```
yields this error from the backend:
```
E   google.api_core.exceptions.InvalidArgument: 400 a filter on __name__ must be a document resource name
```

wilhuff:
In terms of the protos, I think you need to specify the value as [`reference_value`](https://github.com/googleapis/googleapis/blob/0d3f88c7fea64ac53dbc9a09a58977488dbaebbe/google/firestore/v1/document.proto#L120). Note comments there on the format.

tseaver:
[Unrelated pubsub failure](https://source.cloud.google.com/results/invocations/e8706e96-9a37-40d5-8411-7cd62206b4aa/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fpubsub/log).

tseaver:
Same oddity with [logging due to pubsub](https://source.cloud.google.com/results/invocations/bd5e4408-f448-4c7b-a482-a2e4a3bf2134/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Flogging/log).

tseaver:
@crwilcox, @busunkim96 please flip the CLA bit.

Issue #3419, opened 2017-05-15
How to store BigQuery schemas for later use.


tweeter0830:
Is there an easy way to get the schema from a BigQuery table and save it to disk for later use?

For example, say that I wanted to get the schema of a table that currently exists, save it to disk and later create a table with the same schema.

I see that there is such a thing as a SchemaField in the api, but there doesn't seem to be a clean way to serialize it.

tseaver:
@tweeter0830 There is no built-in support for serializing `SchemaField` instances.  The non-public `table._build_schema_resource` and `table._parse_schema_resource` functions could be used as the basis for serializing them to JSON.

tweeter0830:
Thanks.

tseaver:
I'll leave the issue open, as making those two functions public / better documented seems like a reasonable enhancement.

tseaver:
E.g., I could see adding a `schema.SchemaField.from_api_repr` classmethod and a corresponding `schema.SchemaField.to_api_repr` instance method to support your usecase.

tseaver:
#3796 doesn't fix this issue.

lukesneeringer:
Why? As soon as it is a dictionary, saving it to disk is trivial (`json.dump`).

tseaver:
The goal of this issue is to store the *entire* schema for a table (not a single field) and in a format which is "good" for people (i.e., with the appropriate top-level mapping / key holding that list).

yiga2:
Similar use case. I have the JSON schema files in GCS already for all tables. `bq load` CLI supports JSON schema but Python Client lib not (0.29).
Next to a getter method,  `LoadJobConfig.schema.save`, there should be a `LoadJobConfig.schema.load` method too that takes a URL has parameter (gs:// or local).
That would be consistent with CLI and other languages - Java lib has a `setSchema`

chemelnucfin:
Hello, feature requests will now be tracked in the project [Feature Requests](https://github.com/GoogleCloudPlatform/google-cloud-python/projects/5#card-6867160).  I will close this issue now, please feel free to continue to address any issues/concerns here.

tswast:
*Background*

With the BigQuery command-line tool, it's possible to specify a schema via a JSON file. See: https://cloud.google.com/bigquery/docs/schemas#using_a_json_schema_file

Schema format:

```
[
  {
    "description": "quarter",
    "mode": "REQUIRED",
    "name": "qtr",
    "type": "STRING"
  },
  {
[...]

PR #7761, opened 2019-04-20
Phase 1 for storing schemas for later use.

lbristol88:
Part of feature request: #3419

@tswast @engelke

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1974,7 +1974,6 @@ class Client(ClientWithProject):
 
         Destination is a file path or a file object.
         """
-        file_obj = None
         json_schema_list = [f.to_api_repr() for f in schema_list]
 
         if isinstance(destination, io.IOBase):
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1935,60 +1935,56 @@ class Client(ClientWithProject):
         )
         return row_iterator
 
-    def _schema_from_json_file_object(self, file):
+    def _schema_from_json_file_object(self, file_obj):
         """Helper function for schema_from_json that takes a
-           file object that describes a table schema.
+       file object that describes a table schema.
 
-           Returns:
-                List of schema field objects.
+       Returns:
+            List of schema field objects.
         """
-        schema_field_list = list()
-        json_data = json.load(file)
+        json_data = json.load(file_obj)
+        return [SchemaField.from_api_repr(f) for f in json_data]
 
-        for field in json_data:
-            schema_field = SchemaField.from_api_repr(field)
-            schema_field_list.append(schema_field)
-
-        return schema_field_list
+    def _schema_to_json_file_object(self, schema_list, file_obj):
+        """Helper function for schema_to_json that takes a schema list and file
+        object and writes the schema list to the file object with json.dump
+        """
+        json.dump(schema_list, file_obj, indent=2, sort_keys=True)
 
     def schema_from_json(self, file_or_path):
         """Takes a file object or file path that contains json that describes
-            a table schema.
+        a table schema.
 
-            Returns:
-                List of schema field objects.
+        Returns:
+            List of schema field objects.
         """
         if isinstance(file_or_path, io.IOBase):
             return self._schema_from_json_file_object(file_or_path)
-        else:
-            try:
-                with open(file_or_path) as file:
-                    return self._schema_from_json_file_object(file)
-            except OSError:
-                raise ValueError(_NEED_JSON_FILE_ARGUMENT)
+
+        try:
+            with open(file_or_path) as file_obj:
+                return self._schema_from_json_file_object(file_obj)
+        except OSError:
+            raise ValueError(_NEED_JSON_FILE_ARGUMENT)
 
     def schema_to_json(self, schema_list, destination):
         """Takes a list of schema field objects.
 
-            Serializes the list of schema field objects as json to a file.
+        Serializes the list of schema field objects as json to a file.
 
-            Destination is a file path or a file object.
+        Destination is a file path or a file object.
         """
         file_obj = None
-        json_schema_list = list()
-
-        for field in schema_list:
-            schema_field = field.to_api_repr()
-            json_schema_list.append(schema_field)
+        json_schema_list = [f.to_api_repr() for f in schema_list]
 
         if isinstance(destination, io.IOBase):
-            destination.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
-        else:
-            try:
-                with open(destination, mode="w") as file:
-                    file.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
-            except OSError:
-                raise ValueError(_NEED_JSON_FILE_ARGUMENT)
+            return self._schema_to_json_file_object(json_schema_list, destination)
+
+        try:
+            with open(destination, mode="w") as file_obj:
+                return self._schema_to_json_file_object(json_schema_list, file_obj)
+        except OSError:
+            raise ValueError(_NEED_JSON_FILE_ARGUMENT)
 
 
 # pylint: disable=unused-argument
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1943,7 +1943,7 @@ class Client(ClientWithProject):
             List of schema field objects.
         """
         json_data = json.load(file_obj)
-        return [SchemaField.from_api_repr(f) for f in json_data]
+        return [SchemaField.from_api_repr(field) for field in json_data]
 
     def _schema_to_json_file_object(self, schema_list, file_obj):
         """Helper function for schema_to_json that takes a schema list and file
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1935,29 +1935,15 @@ class Client(ClientWithProject):
         )
         return row_iterator
 
-    def schema_from_json(self, file_or_path):
-        """Takes a file object or file path that contains json that describes
-            a table schema.
+    def _schema_from_json_file_object(self, file):
+        """Helper function for schema_from_json that takes a
+           file object that describes a table schema.
 
-            Returns:
+           Returns:
                 List of schema field objects.
         """
-        file_obj = None
-        json_data = None
         schema_field_list = list()
-
-        if isinstance(file_or_path, io.IOBase):
-            file_obj = file_or_path
-        else:
-            try:
-                file_obj = open(file_or_path)
-            except OSError:
-                raise TypeError(_NEED_JSON_FILE_ARGUMENT)
-
-        try:
-            json_data = json.load(file_obj)
-        except JSONDecodeError:
-            raise TypeError(_NEED_JSON_FILE_ARGUMENT)
+        json_data = json.load(file)
 
         for field in json_data:
             schema_field = SchemaField.from_api_repr(field)
@@ -1965,29 +1951,44 @@ class Client(ClientWithProject):
 
         return schema_field_list
 
+    def schema_from_json(self, file_or_path):
+        """Takes a file object or file path that contains json that describes
+            a table schema.
+
+            Returns:
+                List of schema field objects.
+        """
+        if isinstance(file_or_path, io.IOBase):
+            return self._schema_from_json_file_object(file_or_path)
+        else:
+            try:
+                with open(file_or_path) as file:
+                    return self._schema_from_json_file_object(file)
+            except OSError:
+                raise ValueError(_NEED_JSON_FILE_ARGUMENT)
+
     def schema_to_json(self, schema_list, destination):
         """Takes a list of schema field objects.
 
             Serializes the list of schema field objects as json to a file.
 
-            Destination is a file path or a file object. 
+            Destination is a file path or a file object.
         """
         file_obj = None
         json_schema_list = list()
-
-        if isinstance(destination, io.IOBase):
-            file_obj = destination
-        else:
-            try:
-                file_obj = open(destination, mode="w")
-            except OSError:
-                raise TypeError(_NEED_JSON_FILE_ARGUMENT)
 
         for field in schema_list:
             schema_field = field.to_api_repr()
             json_schema_list.append(schema_field)
 
-        file_obj.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
+        if isinstance(destination, io.IOBase):
+            destination.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
+        else:
+            try:
+                with open(destination, mode="w") as file:
+                    file.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
+            except OSError:
+                raise ValueError(_NEED_JSON_FILE_ARGUMENT)
 
 
 # pylint: disable=unused-argument
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -23,6 +23,8 @@ except ImportError:  # Python 2.7
 
 import functools
 import gzip
+import io
+import json
 import os
 import tempfile
 import uuid
@@ -50,6 +52,7 @@ from google.cloud.bigquery.model import Model
 from google.cloud.bigquery.model import ModelReference
 from google.cloud.bigquery.query import _QueryResults
 from google.cloud.bigquery.retry import DEFAULT_RETRY
+from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import _table_arg_to_table
 from google.cloud.bigquery.table import _table_arg_to_table_ref
 from google.cloud.bigquery.table import Table
@@ -70,6 +73,9 @@ _RESUMABLE_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u"resumable"
 _GENERIC_CONTENT_TYPE = u"*/*"
 _READ_LESS_THAN_SIZE = (
     "Size {:d} was specified but the file-like object only had " "{:d} bytes remaining."
+)
+_NEED_JSON_FILE_ARGUMENT = (
+    "The JSON file argument should be a file object or a file path"
 )
 _NEED_TABLE_ARGUMENT = (
     "The table argument should be a table ID string, Table, or TableReference"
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,60 @@ def list_rows(
...
+        """
+        file_obj = None
+        json_data = None
+        schema_field_list = list()
+
+        if isinstance(file_or_path, io.IOBase):
+            file_obj = file_or_path
+        else:
+            try:
+                file_obj = open(file_or_path)
```
tswast: An important difference with this path is that we *must close* `file_obj` if we open it based on a path, but we *must **not** close* `file_obj` if we are given a file-like object.

I recommend using a `with` statement when we are provided a path to make sure we always close it. You may want to refactor this to add a `_schema_from_json_file_object()` private helper method that does the loading part without the file closing part.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,60 @@ def list_rows(
...
+            except OSError:
+                raise TypeError(_NEED_JSON_FILE_ARGUMENT)
```
tswast: In general we use `ValueError` for unexpected input argument.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,60 @@ def list_rows(
...
+
+        try:
+            json_data = json.load(file_obj)
+        except JSONDecodeError:
+            raise TypeError(_NEED_JSON_FILE_ARGUMENT)
```
tswast: I'd prefer if we let this raise (don't catch it). That way the user doesn't lose context about what is wrong with the input file.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,60 @@ def list_rows(
...
+            Destination is a file path or a file object. 
+        """
+        file_obj = None
+        json_schema_list = list()
+
+        if isinstance(destination, io.IOBase):
+            file_obj = destination
+        else:
+            try:
+                file_obj = open(destination, mode="w")
```
tswast: Just as in `schema_from_json`, we *must close* `file_obj` if we open it based on a path, but we *must **not** close* `file_obj` if we are given a file-like object.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,84 @@ def test__do_multipart_upload_wrong_size(self):
...
+          },
+          {
+            "description": "total sales",
+            "mode": "NULLABLE",
+            "name": "sales",
+            "type": "FLOAT"
+          }
+        ]"""
+        expected = list()
+        json_data = json.loads(file_content)
```
tswast: You're basically repeating the function definition here. That's not ideal. You've basically written a [change-detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). I'd prefer to see actual `schema.SchemaField` constructors in tests.

I do like that you've mocked out `open` and the file contents, so keep that. Just change how you construct `expected`.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,84 @@ def test__do_multipart_upload_wrong_size(self):
 
         with pytest.raises(ValueError):
             client._do_multipart_upload(file_obj, {}, file_obj_len + 1, None)
+
+    def test__schema_from_json(self):
```
tswast: Nit: since the function name is `schema_from_json`, the test name should be `test_schema_from_json` (just one underscore between `test` and `schema`.)

Ditto for `schema_to_json` test function name.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,84 @@ def test__do_multipart_upload_wrong_size(self):
...
+        ]"""
+        expected = list()
+        json_data = json.loads(file_content)
+
+        for field in json_data:
+            schema_field = SchemaField.from_api_repr(field)
+            expected.append(schema_field)
+
+        client = self._make_client()
+        mock_file_path = "/mocked/file.json"
```
tswast: In addition to a test that passes in a path, please add a test `test_schema_from_json_with_file_object` that passes in a file object. That should actually be easier to test because you can pass in a [`BytesIO`](https://pythonhosted.org/six/#six.BytesIO) object. No mocking required for that second test.

Same applies for `schema_to_json`. Add a test using `BytesIO` for that, too, please.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,84 @@ def test__do_multipart_upload_wrong_size(self):
...
+
+        open_patch = mock.patch(
+            "builtins.open", new=mock.mock_open(read_data=file_content)
+        )
+        with open_patch as _mock_file:
+            actual = client.schema_from_json(mock_file_path)
+            _mock_file.assert_called_once_with(mock_file_path)
```
tswast: Check that `close` is called once as well. Ditto for `schema_to_json` test with file path.

bigquery/google/cloud/bigquery/client.py
lbristol88: Added helper method per recommendation along with the statement to close the file.

bigquery/google/cloud/bigquery/client.py
lbristol88: Took this piece out.

bigquery/google/cloud/bigquery/client.py
lbristol88: Updated with `with` statement to close when appropriate.

bigquery/tests/unit/test_client.py
lbristol88: Made changes to include schema.SchemaField as requested.

bigquery/tests/unit/test_client.py
lbristol88: Updated test names appropriately.

bigquery/tests/unit/test_client.py
lbristol88: Added two tests to cover this but i had to use StringIO to make it work.

bigquery/tests/unit/test_client.py
lbristol88: I made changes to the tests to reflect closing as requested. Used __exit__ for the context manager.

bigquery/google/cloud/bigquery/client.py
lbristol88: Updated with correct exception.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
...
+
+        for field in json_data:
+            schema_field = SchemaField.from_api_repr(field)
+            schema_field_list.append(schema_field)
+
+        return schema_field_list
+
+    def schema_from_json(self, file_or_path):
+        """Takes a file object or file path that contains json that describes
+            a table schema.
```
tswast: Nit: This and the other docstrings (except for the first line) look a like they are indented 4 spaces too many.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
         )
         return row_iterator
 
+    def _schema_from_json_file_object(self, file):
```
tswast: Nit: since `file` is a built-in, use `file_` or `file_obj` as the name.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
...
+        """Helper function for schema_from_json that takes a
+           file object that describes a table schema.
+
+           Returns:
+                List of schema field objects.
+        """
+        schema_field_list = list()
+        json_data = json.load(file)
+
+        for field in json_data:
```
tswast: FYI: This loop could be replaced with a Python "list comprehension".

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
...
+            a table schema.
+
+            Returns:
+                List of schema field objects.
+        """
+        if isinstance(file_or_path, io.IOBase):
+            return self._schema_from_json_file_object(file_or_path)
+        else:
+            try:
+                with open(file_or_path) as file:
```
tswast: Nit: `file` should be `file_obj` since `file` is a built-in.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
...
+    def schema_from_json(self, file_or_path):
+        """Takes a file object or file path that contains json that describes
+            a table schema.
+
+            Returns:
+                List of schema field objects.
+        """
+        if isinstance(file_or_path, io.IOBase):
+            return self._schema_from_json_file_object(file_or_path)
+        else:
```
tswast: Nit: Since the above line `return`s, no need for `else`.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,61 @@ def list_rows(
...
+        """
+        file_obj = None
+        json_schema_list = list()
+
+        for field in schema_list:
+            schema_field = field.to_api_repr()
+            json_schema_list.append(schema_field)
+
+        if isinstance(destination, io.IOBase):
+            destination.write(json.dumps(json_schema_list, indent=2, sort_keys=True))
```
tswast: Use [`json.dump(schema_list, destination)`](https://docs.python.org/3/library/json.html#json.dump) instead. Then you can use `BytesIO` in the tests, too.

I'd prefer if `json.dump` wasn't repeated (here we do want to be DRY). Replace this with `file_obj = destination` and remove the `else`.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+          {
+            "description": "total sales",
+            "mode": "NULLABLE",
+            "name": "sales",
+            "type": "FLOAT"
+          }
+        ]"""
+
+        expected = list()
+        expected.append(SchemaField("qtr", "STRING", "REQUIRED", "quarter"))
```
tswast: No need for `.append`. Instead, construct the list inline.

```
expected = [
    SchemaField(...),
    SchemaField(...),
    ...
]
```

bigquery/tests/unit/test_client.py
tswast: Use `json.dump` and a `BytesIO` project will work.

👌 regarding `__exit__`.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+        mock_file_path = "/mocked/file.json"
+
+        open_patch = mock.patch("builtins.open", mock.mock_open())
+        with open_patch as _mock_file:
+            actual = client.schema_to_json(schema_list, mock_file_path)
+            _mock_file.assert_called_once_with(mock_file_path, mode="w")
+            _mock_file().write.assert_called_once_with(file_content)
+            # This assert is to make sure __exit__ is called in the context
+            # manager that opens the file in the function
+            _mock_file().__exit__.assert_called_once_with(None, None, None)
```
tswast: We don't care about the actual arguments, so just `assert_called_once` will work.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+        )
+        schema_list.append(SchemaField("sales", "FLOAT", "NULLABLE", "total sales"))
+
+        client = self._make_client()
+        mock_file_path = "/mocked/file.json"
+
+        open_patch = mock.patch("builtins.open", mock.mock_open())
+        with open_patch as _mock_file:
+            actual = client.schema_to_json(schema_list, mock_file_path)
+            _mock_file.assert_called_once_with(mock_file_path, mode="w")
```
tswast: Might need `wb` when using `json.dump`. (b for "binary" mode)

bigquery/tests/unit/test_client.py
lbristol88: I tried to use a `BytesIO` object and kept getting this error when using `json.dump`:

```
>           fp.write(chunk)
E           TypeError: a bytes-like object is required, not 'str'

/usr/local/opt/pyenv/versions/3.7.2/lib/python3.7/json/__init__.py:180: TypeError
```

I checked the [docs](https://docs.python.org/3/library/json.html#basic-usage) for `json.dump` and found this:

> The json module always produces str objects, not bytes objects. Therefore, fp.write() must support str input.

I then checked the [docs](https://docs.python.org/3/library/io.html#io.BufferedIOBase.write) for `io.BytesIO` but it doesn't say anything about supporting string input.

> Write the given bytes-like object, b, and return the number of bytes written (always equal to the length of b in bytes, since if the write fails an OSError will be raised).

Do you have an example of what I should be doing?

bigquery/google/cloud/bigquery/client.py
lbristol88: Fixed indentations.

bigquery/google/cloud/bigquery/client.py
lbristol88: Changed name as directed.

bigquery/google/cloud/bigquery/client.py
lbristol88: Changed to be a list comprehension.

bigquery/google/cloud/bigquery/client.py
lbristol88: Updated name accordingly.

bigquery/google/cloud/bigquery/client.py
lbristol88: Removed the else as recommended.

bigquery/tests/unit/test_client.py
lbristol88: Switched to suggested method.

bigquery/google/cloud/bigquery/client.py
lbristol88: Refactored into a helper function because I couldn't get it to work with the context manager below just by removing the else.

bigquery/tests/unit/test_client.py
lbristol88: Removed the arguments from the assert.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+            SchemaField("rep", "STRING", "NULLABLE", "sales representative"),
+            SchemaField("sales", "FLOAT", "NULLABLE", "total sales"),
+        ]
+
+        client = self._make_client()
+        mock_file_path = "/mocked/file.json"
+
+        open_patch = mock.patch("builtins.open", mock.mock_open())
+        with open_patch as _mock_file:
+            with mock.patch("json.dump") as _mock_dump:
```
tswast: I like your thinking here. It's definitely nicer to compare lists than it is to compare JSON strings.

Nit: Let's write both `with` statement on one line. https://stackoverflow.com/a/1073814/101923

Nit: No need for leading underscore, that's usually to indicate a "private" variable, which isn't relevant inside a test.

```
with open_patch as mock_file, mock.patch("json.dump") as mock_dump:
```

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+            SchemaField("qtr", "STRING", "REQUIRED", "quarter"),
+            SchemaField("rep", "STRING", "NULLABLE", "sales representative"),
+            SchemaField("sales", "FLOAT", "NULLABLE", "total sales"),
+        ]
+
+        fake_file = io.StringIO()
+        client = self._make_client()
+
+        client.schema_to_json(schema_list, fake_file)
+        assert file_content == fake_file.getvalue()
```
tswast: Let's call `json.loads(fake_file.getvalue())` and compare to a list of dictionaries like you do in the test of the path version.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,56 @@ def list_rows(
...
 
+    def _schema_from_json_file_object(self, file_obj):
+        """Helper function for schema_from_json that takes a
+       file object that describes a table schema.
+
+       Returns:
+            List of schema field objects.
+        """
+        json_data = json.load(file_obj)
+        return [SchemaField.from_api_repr(f) for f in json_data]
```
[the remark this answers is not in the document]
lbristol88: Variable name has been updated!

bigquery/tests/unit/test_client.py
lbristol88: Replace existing `with` statements with suggested line.

bigquery/tests/unit/test_client.py
lbristol88: The test has been updated with the list of dictionaries and updated assert.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1929,6 +1935,56 @@ def list_rows(
...
+        Returns:
+            List of schema field objects.
+        """
+        if isinstance(file_or_path, io.IOBase):
+            return self._schema_from_json_file_object(file_or_path)
+
+        try:
+            with open(file_or_path) as file_obj:
+                return self._schema_from_json_file_object(file_obj)
+        except OSError:
```
tswast: Oops, coverage is failing on this line and the similar line in the other function. We'd need a test where `open()` fails.

Honestly, I'd be okay removing the `try` block and letting these errors just raise, too.

lbristol88: Removed the `try` block as suggested for both functions.

bigquery/tests/unit/test_client.py
```diff
@@ -5161,3 +5161,166 @@ def test__do_multipart_upload_wrong_size(self):
...
+        schema_list = [
+            SchemaField("qtr", "STRING", "REQUIRED", "quarter"),
+            SchemaField("rep", "STRING", "NULLABLE", "sales representative"),
+            SchemaField("sales", "FLOAT", "NULLABLE", "total sales"),
+        ]
+
+        client = self._make_client()
+        mock_file_path = "/mocked/file.json"
+
+        open_patch = mock.patch("builtins.open", mock.mock_open())
```
tswast: From the test logs, it looks like this is tripping up Python 2. https://stackoverflow.com/a/34677735/101923

```
____________ TestClientUpload.test_schema_from_json_with_file_path _____________
self = <tests.unit.test_client.TestClientUpload object at 0x7f47e2dce4d0>
    def test_schema_from_json_with_file_path(self):
        from google.cloud.bigquery.schema import SchemaField
        file_content = """[
          {
            "description": "quarter",
            "mode": "REQUIRED",
            "name": "qtr",
            "type": "STRING"
          },
          {
            "description": "sales representative",
            "mode": "NULLABLE",
            "name": "rep",
            "type": "STRING"
          },
          {
            "description": "total sales",
            "mode": "NULLABLE",
            "name": "sales",
            "type": "FLOAT"
          }
        ]"""
        expected = [
            SchemaField("qtr", "STRING", "REQUIRED", "quarter"),
            SchemaField("rep", "STRING", "NULLABLE", "sales representative"),
            SchemaField("sales", "FLOAT", "NULLABLE", "total sales"),
        ]
        client = self._make_client()
        mock_file_path = "/mocked/file.json"
        open_patch = mock.patch(
            "builtins.open", new=mock.mock_open(read_data=file_content)
        )
>       with open_patch as _mock_file:
tests/unit/test_client.py:5202:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.nox/unit-2-7/lib/python2.7/site-packages/mock/mock.py:1353: in __enter__
    self.target = self.getter()
.nox/unit-2-7/lib/python2.7/site-packages/mock/mock.py:1523: in <lambda>
    getter = lambda: _importer(target)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
target = 'builtins'
    def _importer(target):
        components = target.split('.')
        import_path = components.pop(0)
>       thing = __import__(import_path)
E       ImportError: No module named builtins
```

Remember you can run against Python 2 by using `nox`, locally.

We could check the Python version with [`sys.version_info.major`](https://stackoverflow.com/a/9079062/101923) and change what object you are patching.

tswast: Actually `six.PY2` is probably a better way to detect. https://pythonhosted.org/six/

lbristol88: Fixed all the issues and the code is compatible with python 2 now.

PR #7791, opened 2019-04-23
Monitoring: add VPCSC system test.

steinwaywhw:
This new file is added in order to test client lib compatibility of VPC SC. The tests are auto-generated from the service definition of the monitoring API. The tests can be run inside or outside of VPC service perimeter. The input to the script should be the following environment variables. 

1. PROJECT_ID: a project that is inside the VPC perimeter. 
2. GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT: a project that is outside the VPC perimeter.

### DIFF
--- a/monitoring/tests/system/test_vpcsc.py
+++ b/monitoring/tests/system/test_vpcsc.py
@@ -1,0 +1,244 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     https://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# DO NOT MODIFY! AUTO-GENERATED!
+# This file is auto-generated on 2019-04-23.
+
+
+import os
+
+from google.api_core import exceptions
+from google.cloud import monitoring_v3
+from google.cloud.monitoring_v3 import enums
+
+
+class TestVPCServiceControlV3(object):
+
+    @staticmethod
+    def _is_rejected(call):
+        try:
+            responses = call()
+
+            # If we reach this line, then call() did not raise. The return
+            # result must be either a google.api_core.page_iterator.Iterator
+            # instance, or None.
+            list(responses)
+        except exceptions.PermissionDenied as e:
+            return e.message == "Request is prohibited by organization's policy"
+        except:
+            pass
+        return False
+
+    @staticmethod
+    def _do_test(delayed_inside, delayed_outside):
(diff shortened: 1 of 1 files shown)


monitoring/tests/system/test_vpcsc.py
```diff
@@ -0,0 +1,244 @@
...
+            list(responses)
+        except exceptions.PermissionDenied as e:
+            return e.message == "Request is prohibited by organization's policy"
+        except:
+            pass
+        return False
+
+    @staticmethod
+    def _do_test(delayed_inside, delayed_outside):
+        if ('GOOGLE_CLOUD_TESTS_IN_VPCSC' in os.environ) and (os.environ['GOOGLE_CLOUD_TESTS_IN_VPCSC'].lower() == 'True'.lower()):
```
bmccutchon: You can just write `os.environ.get('GOOGLE_CLOUD_TESTS_IN_VPCSC', 'false').lower() == 'true'`.

steinwaywhw:
Hi @tseaver, the VPC team has their own Kokoro workflow to test this. I believe @bmccutchon knows more. Otherwise, if you'd like to run this test in the pull request workflow, please supply these environment variables: `PROJECT_ID=secure-gcp-test-project-4` and `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT=some GCP test project that you know of`

tseaver:
@tseaver It doesn't seem to me to make sense to have the tests maintained in a separate repository from the location where the CI runs.  @busunkim96, @crwilcox please chime in.

steinwaywhw:
This is a requirement from the VPC team. Many teams are going to implement such tests. I have merged a similar one here: https://github.com/googleapis/google-cloud-java/pull/5006, and the Spanner team has done it here https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITVPCNegativeTest.java.

bmccutchon:
@tseaver Echoing what @steinwaywhw said, we expect the client library teams to write and maintain these tests. The easiest way to do that is to keep them in this repo. I agree that it would make sense to have the CI workflow also be in this repo or run against affected pull requests, but that isn't feasible quite yet.

busunkim96:
@bmccutchon Would it make sense for us to run these tests against PRs in this repo, or is there some limitation that would keep us from doing that? Are there requirements beyond setting the environment variables @steinwaywhw mentioned?

We would probably edit the nox session [here](https://github.com/googleapis/google-cloud-python/blob/f5aec2d2fbb72d613abf5f2c99bdefc501b077d6/monitoring/noxfile.py#L98) to set the required environment variables.

steinwaywhw:
Hi @busunkim96, there are essentially two ways to run these tests: 1) from inside VPCSC, 2) from outside VPCSC. For case 1, you need to setup VPCSC environments, which is non-trivial. The VPC team has a setup, and they can run it. For case 2, you do not need a VPCSC setup, and the tests can be run by simply setting those two environment variables, and the test case still provides valuable informations on the effectiveness of VPC service control.

steinwaywhw:
Friendly ping 👋

busunkim96:
Merging this now. I will create a follow-up PR to set the environment variables to make the tests run.

steinwaywhw:
@busunkim96 Thank you very much! If you have further questions, feel free to reach out to me.

PR #7793, opened 2019-04-23
Preserve order in `to_dataframe` with BQ Storage from queries containing `ORDER BY`

### LINKED ISSUE #7759: BigQuery: order not preserved when downloading ORDER BY query results to dataframe with BQ Storage API
Since data is downloaded in parallel, order might not be preserved. I believe if you explicitly request only a single stream, order is preserved. (Internal bug 130887748 to confirm this).

tswast:
This fixes an issue where due to reading from multiple stream in
parallel, the order of rows is not preserved. Normally this isn't an
issue, but it is when the rows are query results from an `ORDER BY` query.

Note: The detection of `ORDER BY` is not perfect. It will have false positives (such as `ORDER BY` in a string or sub-query), but hopefully no false negatives. Since a false positive only means we read from a single stream, behavior is correct but perhaps slower than expected.

Closes #7759

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -46,6 +46,7 @@ from google.cloud.bigquery import _helpers
 _DONE_STATE = "DONE"
 _STOPPED_REASON = "stopped"
 _TIMEOUT_BUFFER_SECS = 0.1
+_CONTAINS_ORDER_BY = re.compile(r"ORDER\s+BY", re.IGNORECASE)
 
 _ERROR_REASON_TO_EXCEPTION = {
     "accessDenied": http_client.FORBIDDEN,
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -92,6 +93,14 @@ def _error_result_to_exception(error_result):
     )
 
 
+def _contains_order_by(query):
+    """Do we need to preserve the order of the query results?"""
+    if query is None:
+        return False
+
+    return re.search(r"ORDER\s+BY", query, re.IGNORECASE) is not None
```
shollyman: Should we have a comment indicating we know there's a class of false positives (ordered window functions)?

tseaver: Please compile the pattern as a module-scope global, and use "raw" strings e.g.:

```python
_CONTAINS_ORDER_BY = re.compile(r"ORDER\s+BY", re.IGNORECASE)

...
def _contains_order_by(query):
    """Do we need to preserve the order of the query results?"""
    return query and _CONTAINS_ORDER_BY.search(query)
```

PR #7794, opened 2019-04-24
Remove `folder/*` pattern, since it only matches immediate child files

tswast:
> The `docs/*` pattern will match files like `docs/getting-started.md`
> but not further nested files like `docs/build-app/troubleshooting.md`.

See: https://help.github.com/en/articles/about-code-owners

P.S. I realize this whole file is probably out-of-date. I'm only making this mechanical change so that the right code reviewers get added on BigQuery updates.


busunkim96:
@tswast I added an item to the next Python sync to review this file.

Issue #7003, opened 2018-12-19
Bigtable:  systests leaking instances


tseaver:
Similar to #4935.  We have a bunch of leaked instances, and are now seeing systest failures due to quota exhaustion.  With the labels, it is possible to see that several different "kinds" of instances are leaked from systests:

- `dif-####` (leaked from `test_create_instance_w_two_clusters`).
- `g-c-p-###` w/ display name "Foo Bar Baz" (the default instance in `tests/system.py`, mutated by `test_update_display_name_and_labels` and not restored)
- `g-c-p-###` w/o display name (the default instance in `tests/system.py`)
- `new-###` (leaked from `test_create_instance` systest)

and from snippets:

- `inst-my-###` (leaked one of two snippets:  `test_bigtable_create_instance` or `test_bigtable_delete_instance`)
- `snippet-###` w/ display name "My new instance" (default instance for snippets;  display name set by 'test_bigtable_update_instance`)
- `snippet-###` w/o display name (default instance for snippets)

I ~~will delete~~ have deleted all but the most recent example of each kind.

tseaver:
@sduskis Can you have the team which has been working on snippets, etc. investigate those leaks?

tseaver:
#7004 has fixes in for all my hypotheses for instance leaks in `tests/system.py`.  I have deleted the corresponding instances, so the only "old" ones remaining were leaked from `docs/snippets.py`.

tseaver:
I just cleaned out four instances leaked from snippets on 2018-12-18.

PR #7800, opened 2019-04-24
Bigtable: fix instance leaks from snippets

sangramql:
Fix issue
https://github.com/googleapis/google-cloud-python/issues/7003

### DIFF
--- a/bigtable/docs/snippets.py
+++ b/bigtable/docs/snippets.py
@@ -384,7 +384,7 @@ def test_bigtable_delete_instance():
     operation = instance.create(clusters=[cluster])
 
     # Make sure this instance gets deleted after the test case.
-    INSTANCES_TO_DELETE.append(instance_id_to_delete)
+    INSTANCES_TO_DELETE.append(instance)
 
     # We want to make sure the operation completes.
     operation.result(timeout=100)
--- a/bigtable/docs/snippets.py
+++ b/bigtable/docs/snippets.py
@@ -33,6 +33,7 @@ import datetime
 import pytest
 
 from test_utils.system import unique_resource_id
+from google.api_core.exceptions import NotFound
 from google.cloud._helpers import UTC
 from google.cloud.bigtable import Client
 from google.cloud.bigtable import enums
@@ -67,7 +68,6 @@ class Config(object):
 
 
 def setup_module():
-    global INSTANCES_TO_DELETE
     client = Config.CLIENT = Client(admin=True)
     Config.INSTANCE = client.instance(
         INSTANCE_ID, instance_type=PRODUCTION, labels=LABELS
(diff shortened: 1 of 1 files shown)


sangramql:
@crwilcox @AVaksman @sduskis
Please review the code.

bigtable/docs/snippets.py
```diff
@@ -66,6 +67,7 @@ class Config(object):
 
 
 def setup_module():
+    global INSTANCES_TO_DELETE
```
tseaver: `global` is unnecessary here, as you don't need (or want) to rebind the name, but rather to reuse the existing list.

bigtable/docs/snippets.py
```diff
@@ -79,13 +81,24 @@ def setup_module():
...
     # We want to make sure the operation completes.
     operation.result(timeout=100)
+    INSTANCES_TO_DELETE.append(Config.INSTANCE)
 
 
 def teardown_module():
-    Config.INSTANCE.delete()
+    for instance in INSTANCES_TO_DELETE:
+        if instance.exists():
+            instance.delete()
```
tseaver: It would be cleaner to just try deleting the instance and ignore the 404, e.g.:

```python
from google.api_core import exceptions

...

def teardown_module():
    for instance in INSTANCES_TO_DELETE:
        try:
            instance.delete()
        except exceptions.NotFound:
            pass
```

bigtable/docs/snippets.py
```diff
@@ -79,13 +81,24 @@ def setup_module():
...
+
+    # Avoid any leak of snippet instances.
+    client = Client(admin=True)
+    (instances_list, failed_locations_list) = client.list_instances()
+    for instance in instances_list:
+        if "snippet-tests" in instance.instance_id:
+            instance.delete()
```
tseaver: This part should not be in the script, as we aren't creating such an instance anywhere.

bigtable/docs/snippets.py
```diff
@@ -79,13 +81,24 @@ def setup_module():
...
 
 
 def test_bigtable_create_instance():
+    global INSTANCES_TO_DELETE
```
tseaver: No `global` needed here.

bigtable/docs/snippets.py
```diff
@@ -270,6 +288,7 @@ def test_bigtable_reload_cluster():
 
 
 def test_bigtable_update_instance():
+    global INSTANCES_TO_DELETE
```
tseaver: No `global` needed here.

bigtable/docs/snippets.py
```diff
@@ -348,6 +370,7 @@ def test_bigtable_delete_cluster():
 
 
 def test_bigtable_delete_instance():
+    global INSTANCES_TO_DELETE
```
tseaver: No `global` needed here.

bigtable/docs/snippets.py
sangramql: Now snippet instance is renamed to snippet-tests line 41.
Just to ensure any previously executed snippets instances pending should be deleted.

tseaver: I saw that change, but we can't remove instances that weren't created as part of this CI run:  we often have multiple PRs running CI at the same time.  Please remove this block.

bigtable/docs/snippets.py
```diff
@@ -367,6 +382,10 @@ def test_bigtable_delete_instance():
         default_storage_type=STORAGE_TYPE,
     )
     operation = instance.create(clusters=[cluster])
+
+    # Make sure this instance gets deleted after the test case.
+    INSTANCES_TO_DELETE.append(instance_id_to_delete)
```
tseaver: We need to append the instance object, not its name, here.

PR #7801, opened 2019-04-24
Initial generation of Phishing Protection.

busunkim96:
The clients were generated with the version name in the name of the clients, so the synth script does a lot of renaming.


phishingprotection/synth.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Files to be renamed (remove v1_beta1 from name)
+    gapic_dir =  Path(f"google/cloud/phishingprotection_{version}/gapic")
+    client =  gapic_dir / "phishing_protection_service_v1_beta1_client.py"
+    client_config = gapic_dir / "phishing_protection_service_v1_beta1_client_config.py"
+    transport = gapic_dir / "transports/phishing_protection_service_v1_beta1_grpc_transport.py"
+    unit_test =  Path("tests/unit/gapic/v1beta1/test_phishing_protection_service_v1_beta1_client_v1beta1.py")
+
+    files = [client, client_config, transport, unit_test]
+    for file in files:
```
tseaver: Please avoid shadowing the `file` builtin.

busunkim96:
Unfortunately it's not currently possible to rename the client upstream.  Ruby also renamed the client using synth. 

> every service has to have a unique name *without considering proto packages*. The intention (as I understand it) is to keep this name for the beta, but change it to not include the version number for V1, in the hope that the issue will be fixed before V2

phishingprotection/synth.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Files to be renamed (remove v1_beta1 from name)
+    gapic_dir =  Path(f"google/cloud/phishingprotection_{version}/gapic")
+    client =  gapic_dir / "phishing_protection_service_v1_beta1_client.py"
+    client_config = gapic_dir / "phishing_protection_service_v1_beta1_client_config.py"
+    transport = gapic_dir / "transports/phishing_protection_service_v1_beta1_grpc_transport.py"
+    unit_test =  Path("tests/unit/gapic/v1beta1/test_phishing_protection_service_v1_beta1_client_v1beta1.py")
+
+    files = [client, client_config, transport, unit_test]
+    for f in files:
```
tseaver: I'm surprised `flake8` isn't complaining about one-character variable name here.  Even if we do change it to `f`, we need to change how it is used inside the loop, too.

busunkim96: Eek, sorry about that. I don't think `synth.py` is currently checked by flake or black.

phishingprotection/nox.py
```diff
@@ -1,54 +0,0 @@
...
-    session.run('py.test', '--quiet', os.path.join('tests', 'unit'))
-
-
-@nox.session
-def lint_setup_py(session):
-    """Verify that setup.py is valid (including RST check)."""
-    session.interpreter = 'python3.6'
-    session.install('docutils', 'pygments')
-    session.run('python', 'setup.py', 'check', '--restructuredtext',
-                '--strict')
```
tseaver: Do we really mean to be removing `nox.py` here?

busunkim96: Yes, since we have `noxfile.py`.

tseaver: OK.  Do we have an issue for the upstream generator to stop generating `nox.py`?

tseaver:
[Pubsub](https://source.cloud.google.com/results/invocations/96c4b61d-2a33-4875-b91a-02dad3bbef00/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fpubsub/log), [Storage](https://source.cloud.google.com/results/invocations/0ee69e1c-c3f6-4e10-a760-5713eee82f21/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log), and [Logging](https://source.cloud.google.com/results/invocations/eb533b61-9505-4dba-8b90-13d33a8c91ce/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Flogging/log) failures are due to a weird, already-fixed-on-master pubsub issue.

phishingprotection/nox.py
busunkim96: I added a note to https://github.com/googleapis/gapic-generator/issues/2360 that we'd prefer to remove this.

PR #7802, opened 2019-04-24
Cleanup client tests with method to create minimal table resource

tswast:
This removes a lot of repeated code that wasn't relevant to the actual tests.


lbristol88:
LGTM!

PR #7799, opened 2019-04-24
Add `user_agent` property to `ClientInfo`

tswast:
This provides a way for partners to define a prefix identifying their
tool or application, as required by many cloud partnership agreements.

PR #7806, opened 2019-04-25
Add `client_info` to BigQuery constructor for user-amenable user agent headers

tswast:
This aligns BigQuery's behavior regarding the User-Agent and
X-Goog-Api-Client headers with that of the GAPIC-based clients.

Old:

    X-Goog-API-Client: gl-python/3.7.2 gccl/1.11.2
    User-Agent: gcloud-python/0.29.1

New:

    X-Goog-API-Client: optional-application-id/1.2.3 gl-python/3.7.2 grpc/1.20.0 gax/1.9.0 gapic/1.11.2 gccl/1.11.2
    User-Agent: optional-application-id/1.2.3 gl-python/3.7.2 grpc/1.20.0 gax/1.9.0 gapic/1.11.2 gccl/1.11.2

In order to set the `optional-application-id/1.2.3`, the latest version
of `api_core` is required, but since that's an uncommon usecase and it
doesn't break, just ignore the custom User-Agent if an older version is
used, I didn't update the minimum version `setup.py`.

This PR depends on:

- https://github.com/googleapis/google-cloud-python/pull/7799 for unit test purposes.
- https://github.com/googleapis/google-cloud-python/pull/7814 so that people can actuall use this (optional) feature.

### DIFF
--- a/bigquery/google/cloud/bigquery/_http.py
+++ b/bigquery/google/cloud/bigquery/_http.py
@@ -59,7 +59,9 @@ class Connection(_http.JSONConnection):
 
     @property
     def _EXTRA_HEADERS(self):
-        self._extra_headers[_http.CLIENT_INFO_HEADER] = self._client_info.to_user_agent()
+        self._extra_headers[
+            _http.CLIENT_INFO_HEADER
+        ] = self._client_info.to_user_agent()
         return self._extra_headers
 
     @_EXTRA_HEADERS.setter
--- a/bigquery/google/cloud/bigquery/_http.py
+++ b/bigquery/google/cloud/bigquery/_http.py
@@ -14,12 +14,10 @@
 
 """Create / interact with Google BigQuery connections."""
 
+import google.api_core.gapic_v1.client_info
 from google.cloud import _http
 
 from google.cloud.bigquery import __version__
-
-
-_CLIENT_INFO = _http.CLIENT_INFO_TEMPLATE.format(__version__)
 
 
 class Connection(_http.JSONConnection):
@@ -28,6 +26,19 @@ class Connection(_http.JSONConnection):
     :type client: :class:`~google.cloud.bigquery.client.Client`
     :param client: The client that owns the current connection.
     """
+
+    def __init__(self, client, client_info=None):
+        super(Connection, self).__init__(client)
+
+        if client_info is None:
+            client_info = google.api_core.gapic_v1.client_info.ClientInfo(
+                gapic_version=__version__, client_library_version=__version__
+            )
+        else:
+            client_info.gapic_version = __version__
+            client_info.client_library_version = __version__
+        self._client_info = client_info
+        self._extra_headers = {}
 
     API_BASE_URL = "https://www.googleapis.com"
     """The base of the API call URL."""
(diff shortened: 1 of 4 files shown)


tswast:
> followup change for exposing client_info to the other BQ APIs like dts and storage?

Already present on those, which was a major factor in going with this design.

* https://googleapis.github.io/google-cloud-python/latest/bigquery_storage/gapic/v1beta1/api.html#google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient
* https://googleapis.github.io/google-cloud-python/latest/bigquery_datatransfer/gapic/v1/api.html#google.cloud.bigquery_datatransfer_v1.DataTransferServiceClient

Once #7799 goes in, we'll be able to set `client_info.user_agent = "my-application/1.2.3"` on all GAPIC clients.

At some point, I do intend to update the BQ magics to set this user agent field, as they create their own clients, and I figure we can track that as using BQ from Jupyter.

bigquery/tests/unit/test_client.py
```diff
@@ -1320,6 +1322,38 @@ def test_get_table(self):
...
+        mock_response.json.return_value = self._make_table_resource()
+        user_agent_override = client_info.ClientInfo(user_agent="my-application/1.2.3")
+        client = self._make_one(
+            project=self.PROJECT,
+            credentials=creds,
+            client_info=user_agent_override,
+            _http=http,
+        )
+
+        table = client.get_table(self.TABLE_REF)
```
tseaver: [Lint failure here](https://source.cloud.google.com/results/invocations/51249904-7b66-49db-8836-7ace0e594049/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigquery/log):  the `table` variable is never used.  You can just invoke `client.get_table` without binding the return value, or use `_`.

bigquery/google/cloud/bigquery/_http.py
```diff
@@ -14,21 +14,32 @@
...
-
-
 class Connection(_http.JSONConnection):
     """A connection to Google BigQuery via the JSON REST API.
 
     :type client: :class:`~google.cloud.bigquery.client.Client`
     :param client: The client that owns the current connection.
     """
 
+    def __init__(self, client, client_info=None):
```
tseaver: @tswast Need to document the `client_info` parameter in the class docstring.

bigquery/google/cloud/bigquery/_http.py
```diff
@@ -38,4 +49,21 @@ class Connection(_http.JSONConnection):
...
     """A template for the URL of a particular API call."""
 
-    _EXTRA_HEADERS = {_http.CLIENT_INFO_HEADER: _CLIENT_INFO}
+    @property
+    def USER_AGENT(self):
+        return self._client_info.to_user_agent()
+
+    @USER_AGENT.setter
+    def USER_AGENT(self, value):
+        self._client_info.user_agent = value
```
tseaver: These aren't constants, so why uppercase?

bigquery/google/cloud/bigquery/_http.py
tseaver: We should probably hoist this machinery into the `google.cloud._http.JSONConnection` base class.

bigquery/google/cloud/bigquery/_http.py
tswast: For backwards compatibility.

tswast: There's a `USER_AGENT` on the superclass, which we want to override the behavior of.

PR #7808, opened 2019-04-26
Tasks: Add smoke test and correct docs/index.rst.

busunkim96:
In preparation for Tasks v2 GA.

### DIFF
--- a/tasks/tests/system/gapic/v2/test_system_tasks_v2.py
+++ b/tasks/tests/system/gapic/v2/test_system_tasks_v2.py
@@ -1,0 +1,29 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from google.cloud import tasks_v2
+
+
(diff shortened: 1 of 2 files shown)


tasks/tests/system/gapic/v2/test_system_tasks_v2.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2018 Google LLC
```
plamut: Shouldn't that be 2019?
```suggestion
# Copyright 2019 Google LLC
```

plamut:
@busunkim96 How can this be verified, what is the expected result?

I am probably not the best person to review this, as I am not familiar with Tasks, and also need to prioritize PubSub, sorry.

tasks/tests/system/gapic/v2/test_system_tasks_v2.py
```diff
@@ -0,0 +1,29 @@
...
+class TestSystemTasks(object):
+    def test_recognize(self):
+        client = tasks_v2.CloudTasksClient()
+
+        # Setup Request
+        parent = client.location_path(os.environ["PROJECT_ID"], "us-central1")
+
+        paged_list_response = client.list_queues(parent)
+        resources = list(paged_list_response)
+        assert len(resources) == 0
```
tseaver: How sure are we that our CI account will never be used to test `google-cloud-tasks` manually, and therefore leave a queue lying around?

tseaver: Also, [looks like we need to enable the API for our CI account](https://source.cloud.google.com/results/invocations/ff51496d-29ef-4c01-921a-300c65143b98/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Ftasks/log).

busunkim96: Fair point, I'll rework this so it doesn't depend on a certain # of queues being found. 

Also Pinged Chris to enable the Tasks API.

PR #7832, opened 2019-05-01
Add samples for data catalog `lookup_entry`.

tswast:
These samples use the sample generator to call LookupEntry linked to a BigQuery resource.

Depends on internal CL 246050183 to add the GAPIC config for these samples.


tswast:
GitHub doesn't make it easy to see, but the path is `samples/google/cloud/datacatalog_v1beta1/gapic/lookup_entry/datacatalog_lookup_entry_by_sql.py`. I've complained about the ridiculously long path names to the generator team, though I haven't filed a bug on it yet.

tseaver:
@tswast I'm not talking about the source from which the sample is generated, but where we put it in *this* repo:  samples for docs should be in the `datacatalog/docs` subdirectory, not in `google/cloud/datacatalog_v1beta1/gapic/`:  they aren't part of the library, but rather its docs.

tswast:
Right, I'm saying it's not in `google/`, it's in `samples/`. That's just where the generator sticks it. I could work around that by adding `move` rules to `synth.py`, but if we do that, we should notify the generator team about why `samples/` doesn't work for us. (IMO, `samples/` seems appropriate)

CC @beccasaurus

tseaver:
@tswast I noticed just now that BigQuery has a `samples` subdirectory containing similarly-generated (I think?) samples / snippets.  Could we do that for datacatalog, rather than leaving them inside `google/cloud`?

tswast:
Obsolete by https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2148.

PR #7847, opened 2019-05-02
Monitoring: Set environment variables for VPCSC system tests.

busunkim96:
Follow up to  #7791.


busunkim96:
Runtime Configuration test failing because of failed pull from docker.

Issue #7825, opened 2019-04-30
Add `client_info` to constructors for manual clients


tswast:
The `client_info` constructor argument is the standard argument for changing user agent headers in GAPIC clients. Adding a `user_agent` prefix identifying an app or tool is required for many partners who use Google APIs.

Handwritten clients should provide a similar `client_info` argument to the client constructor. See how this was done in BigQuery at https://github.com/googleapis/google-cloud-python/pull/7806.

Clients that need `client_info`:

- [x] BigQuery (https://github.com/googleapis/google-cloud-python/pull/7806)
- [x] [Bigtable](https://github.com/googleapis/google-cloud-python/blob/af3df9cc2c520ef48870e124b65849c37966c43e/bigtable/google/cloud/bigtable/client.py#L76) (#7876)
- [x] [Datastore](https://github.com/googleapis/google-cloud-python/blob/92465cbc4d9c0ba251838e9cd17f61d14b470e04/datastore/google/cloud/datastore/client.py#L165)
- [x] [DNS](https://github.com/googleapis/google-cloud-python/blob/ff05ac3a8fa58a696584bada6ac89eb0a81e9856/dns/google/cloud/dns/client.py#L24) (#7869)
- [x] [Error Reporting](https://github.com/googleapis/google-cloud-python/blob/92465cbc4d9c0ba251838e9cd17f61d14b470e04/error_reporting/google/cloud/error_reporting/client.py#L84) (#7903)
- [x] [Firestore](https://github.com/googleapis/google-cloud-python/blob/894243931e41e1e9e71762d4cabd2dbe7c39b6af/firestore/google/cloud/firestore_v1/client.py#L51) (and [Firestore Beta](https://github.com/googleapis/google-cloud-python/blob/242e407980b5f6842ee42736d5acb80d1aa0f8d2/firestore/google/cloud/firestore_v1beta1/client.py#L51)) (#7877, V1 only, see below)
- [x] [Logging](https://github.com/googleapis/google-cloud-python/blob/aee5ceca1561e7f2c92a71b7be026851c5cb5d41/logging/google/cloud/logging/client.py#L59) (#7874, #7901)
[...]

PR #7849, opened 2019-05-02
Core, BigQuery: refactor 'client_info' support.

tseaver:
Hoist `client_info` support from `google.cloud.bigquery._http.Connection` into core's `google.cloud._http.Connection`.

Toward #7825.

### DIFF
--- a/api_core/google/api_core/client_info.py
+++ b/api_core/google/api_core/client_info.py
@@ -18,13 +18,20 @@ Client information is used to send information about the calling client,
 such as the library and Python version, to API services.
 """
 
-from google.api_core import client_info
+import platform
+
+import pkg_resources
+
+_PY_VERSION = platform.python_version()
+_API_CORE_VERSION = pkg_resources.get_distribution("google-api-core").version
+
+try:
+    _GRPC_VERSION = pkg_resources.get_distribution("grpcio").version
+except pkg_resources.DistributionNotFound:  # pragma: NO COVER
+    _GRPC_VERSION = None
 
 
-METRICS_METADATA_KEY = "x-goog-api-client"
-
-
-class ClientInfo(client_info.ClientInfo):
+class ClientInfo(object):
     """Client information used to generate a user-agent for API calls.
 
     This user-agent information is sent along with API calls to allow the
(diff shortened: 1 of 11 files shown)


tseaver:
@tswast 
> We'll need to bump the min version of the dependency on google-cloud-core in google-cloud-bigquery, though, right?

Yes, and for the other manual / JSON clients which will rely on this feature in future PRs.

api_core/google/api_core/client_info.py
```diff
@@ -0,0 +1,96 @@
...
+_PY_VERSION = platform.python_version()
+_API_CORE_VERSION = pkg_resources.get_distribution("google-api-core").version
+
+try:
+    _GRPC_VERSION = pkg_resources.get_distribution("grpcio").version
+except pkg_resources.DistributionNotFound:  # pragma: NO COVER
+    _GRPC_VERSION = None
+
+
+class ClientInfo(object):
```
tseaver: @tswast In order to support the API libraries which do not install grpc, I had to split out a base `ClientInfo` class which is importable if 'grpc' is absent.  The one in `google.api_core.gapic_v1.client_info` subclasses and adds the `to_grpc_metadata` method.

tswast: Thanks! Good catch.

api_core/google/api_core/client_info.py
```diff
@@ -0,0 +1,96 @@
...
+import platform
+
+import pkg_resources
+
+_PY_VERSION = platform.python_version()
+_API_CORE_VERSION = pkg_resources.get_distribution("google-api-core").version
+
+try:
+    _GRPC_VERSION = pkg_resources.get_distribution("grpcio").version
+except pkg_resources.DistributionNotFound:  # pragma: NO COVER
```
tswast: Should we add a unit test for this? It'd be good to be doubly sure that it doesn't blow up when `_GRPC_VERSION` is `None`.

tswast:
Let's add `google.api_core.client_info` to the docs, similar to https://github.com/googleapis/google-cloud-python/blob/a017937e307fa437d1bf5fb682bb29a9c16f0a67/docs/core/page_iterator.rst

cgparkinson:
Probably the fault of my setup, but a nightly test failed last night with the error
`ImportError: No module named client_info`

This error has no results on Google so thought I'd post here.

I fixed the error by adding `google-cloud-core<1.0.0` to my dependencies.

Probably will be better to update all my dependencies rather than relying on old versions though.

In case it's useful, my dependencies now read:

```
pandas-gbq
google-cloud-bigquery>=1.9.0
google-cloud-core<1.0.0
google-cloud-storage
parmap
google-api-python-client
google-cloud-pubsub>=0.28.1
```

PR #7850, opened 2019-05-02
Regenerate VPCSC tests to include NotificationChannelService

steinwaywhw:
Hi, this is a refresh of the auto-generated VPC tests. I've tested it locally and it is working as intended. This pull request is sent on behalf of @michaelsafyan.


steinwaywhw:
Close since I mistakenly included more commits than needed. Will re-do.

PR #7864, opened 2019-05-06
Bigtable: Add channel to client. Avoid using same channel by data and admin client

sangramql:
See #6278.
This is initial commit to review the code and approach.

This code will enable channel to be used while creating client.
And will try to avoid data client and admin client to use same channel.
Data client does not need admin scope to create client,

### DIFF
--- a/bigtable/google/cloud/bigtable/client.py
+++ b/bigtable/google/cloud/bigtable/client.py
@@ -63,12 +63,7 @@ READ_ONLY_SCOPE = "https://www.googleapis.com/auth/bigtable.data.readonly"
 def _create_gapic_client(client_class):
     def inner(self):
         if self._emulator_host is None:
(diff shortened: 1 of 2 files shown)


sangramql:
@tseaver , @sduskis , @crwilcox , @AVaksman 
Please review or suggest reviewer.

sangramql:
I was implementing the below comment from original issue:
" If channel is passed, we should forward it through as well, until the GAPIC-generated code no longer spports it. "

Hence implementing the channel argument. I'll close this since GAPIC clients does not support channel arg anymore.

sduskis:
The issue should be rephrased to remove `channel` and add new `data_transport`, `table_admin_transport` and `instance_admin_transport`

sangramql:
Closing this PR , will open new PR for transport.

PR #7865, opened 2019-05-06
Phishing Protection: Finish initial library setup.

busunkim96:
The previous generation had several steps missing, and that release ended up failing. This finishes out the setup. 

* Rename of package from `google-cloud-phishingprotection` to `google-cloud-phishing-protection`with a replace in `synth.py`.
* Complete docs setup.
* Exclude un-executed line in `types.py` so unit tests do not fail.
* Fix classifier in `setup.py`.


tseaver:
@busunkim96 In addition to renaming the distutils project, do we want to rename the `phishingprotection` directory to `phishing_protection`?

phishing_protection/google/cloud/phishingprotection_v1beta1/types.py
```diff
@@ -27,7 +27,7 @@
 
 names = []
 
-for module in _shared_modules:
+for module in _shared_modules:  # pragma: NO COVER
```
tseaver: Hmm, how are we going to keep this fix from being removed in the next autogen?

busunkim96: I didn't think about this at the time :woman_facepalming:. I've raised an issue asking for the loop to be omitted when there are no shared_modules https://github.com/googleapis/gapic-generator/issues/2770

PR #7876, opened 2019-05-07
Bigtable: add 'client_info' support to client.

tseaver:
Forward when constructing GAPIC API client objects.

Toward #7825.

### DIFF
--- a/bigtable/google/cloud/bigtable/client.py
+++ b/bigtable/google/cloud/bigtable/client.py
@@ -63,10 +63,12 @@ READ_ONLY_SCOPE = "https://www.googleapis.com/auth/bigtable.data.readonly"
 def _create_gapic_client(client_class):
     def inner(self):
         if self._emulator_host is None:
-            return client_class(credentials=self._credentials, client_info=_CLIENT_INFO)
+            return client_class(
+                credentials=self._credentials, client_info=self._client_info
+            )
         else:
             return client_class(
-                channel=self._emulator_channel, client_info=_CLIENT_INFO
+                channel=self._emulator_channel, client_info=self._client_info
             )
 
     return inner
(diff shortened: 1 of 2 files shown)


bigtable/tests/unit/test_client.py
```diff
@@ -27,34 +27,32 @@ def _invoke_client_factory(self, client_class):
         return _create_gapic_client(client_class)
 
     def test_without_emulator(self):
-        from google.cloud.bigtable.client import _CLIENT_INFO
-
         client_class = mock.Mock()
         credentials = _make_credentials()
         client = _Client(credentials)
+        client_info = client._client_info = mock.Mock()
```
tswast: Maybe use an autospec here?

bigtable/tests/unit/test_client.py
```diff
@@ -27,34 +27,32 @@ def _invoke_client_factory(self, client_class):
...
     def test_with_emulator(self):
-        from google.cloud.bigtable.client import _CLIENT_INFO
-
         client_class = mock.Mock()
         emulator_host = emulator_channel = object()
         credentials = _make_credentials()
         client = _Client(
             credentials, emulator_host=emulator_host, emulator_channel=emulator_channel
         )
+        client_info = client._client_info = mock.Mock()
```
tswast: Ditto. We can use `mock.create_autospec` with `instance=True` to make doubly sure we don't use any properties on `ClientInfo` that don't exist.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -100,6 +102,13 @@ class Client(ClientWithProject):
                   interact with the Instance Admin or Table Admin APIs. This
                   requires the :const:`ADMIN_SCOPE`. Defaults to :data:`False`.
 
+    :type: client_info: :class:`google.api_core.client_info.ClientInfo`
```
tswast: Does this need to be the `google.api_core.gapic_v1.client_info.ClientInfo` version, instead?

What happens when a `google.api_core.client_info.ClientInfo` is used? I imagine it might fail to populate the `x-goog-api-client` header?

PR #7878, opened 2019-05-07
Spanner: add client_info support to client.

tseaver:
Forward through when creating GAPIC API wrappers.

Toward #7825.

### DIFF
--- a/spanner/google/cloud/spanner_v1/client.py
+++ b/spanner/google/cloud/spanner_v1/client.py
@@ -23,6 +23,7 @@ In the hierarchy of API concepts
 * a :class:`~google.cloud.spanner_v1.instance.Instance` owns a
   :class:`~google.cloud.spanner_v1.database.Database`
 """
+import warnings
 
 from google.api_core.gapic_v1 import client_info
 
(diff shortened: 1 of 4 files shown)


spanner/google/cloud/spanner_v1/client.py
```diff
@@ -95,9 +95,17 @@ class Client(ClientWithProject):
...
+        API requests. If ``None``, then default info will be used. Generally,
+        you only need to set this if you're developing your own library or
+        partner tool.
+
     :type user_agent: str
-    :param user_agent: (Optional) The user agent to be used with API request.
-                       Defaults to :const:`DEFAULT_USER_AGENT`.
+    :param user_agent:
+        (Deprecated) The user agent to be used with API request.
+        Not used.  Defaults to :const:`DEFAULT_USER_AGENT`.
```
tswast: Default value doesn't really matter since it's not used, right? I recommend removing that sentence from the docs (and maybe change the default to `None`).

spanner/google/cloud/spanner_v1/client.py
```diff
@@ -95,9 +95,17 @@ class Client(ClientWithProject):
                         client. If not provided, defaults to the Google
                         Application Default Credentials.
 
+    :type client_info: :class:`google.api_core.client_info.ClientInfo`
```
tswast: Do we need the gapic version?

Same comment as https://github.com/googleapis/google-cloud-python/pull/7876#discussion_r281872712

spanner/google/cloud/spanner_v1/client.py
tseaver: Sure.  In fact, I've gone further and added an actual deprecation warning for passing any non-None value in 7e71bf6.

spanner/google/cloud/spanner_v1/client.py
tseaver: 26c5dec

PR #7897, opened 2019-05-08
Storage: deprecate 'fields' arg to 'Bucket.list_blobs'.

### LINKED ISSUE #7875: Storage: Bucket.list_blobs fields param doc
I have google-cloud-storage-1.15.0 and want to user `Bucket.list_blobs` for many objects, so want to limit the fields returned. However, [from the doc](https://googleapis.github.io/google-cloud-python/latest/storage/buckets.html?highlight=list_blobs#google.cloud.storage.bucket.Bucket.list_blobs) it's not clear to me:
- If I have to specify `nextPageToken` in the `fileds` param as well or is it obsolete like the `page_token` param?
- How to specify multiple fields and what's the proper form: `items(name,size)` or `items/name,items/size`?

Please, fill the doc, thanks.

tseaver:
Closes #7875.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -47,6 +47,10 @@ _LOCATION_SETTER_MESSAGE = (
     "valid before the bucket is created. Instead, pass the location "
     "to `Bucket.create`."
 )
+_LIST_BLOBS_FIELDS_MESSAGE = (
+    "The 'fields' argument to 'Bucket.list_blobs' is deprecated, and "
+    "likely broken:  please don't use it."
+)
 _API_ACCESS_ENDPOINT = "https://storage.googleapis.com"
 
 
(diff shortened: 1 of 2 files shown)


frankyn:
Thanks for looping me in @tseaver. This field isn't deprecated and documented here instead:
https://cloud.google.com/storage/docs/json_api/v1/how-tos/performance

The field is actively used by the google-cloud-java library for Storage for metadata field selection. 

I'll file a bug against cloud documentation on surfacing the field in Objects.List and other ref docs.

xmedeko:
@tseaver beside the warning, the deprecation should be in the method doc, too. (If this change is accepted.)

tseaver:
@frankyn OK, then do you think we need to add a method which makes `fields` useful, e.g. `Bucket.list_blob_info`?  Using it at all breaks `Bucket.list_blobs`.

tseaver:
Superseded by #8023.

PR #7904, opened 2019-05-08
Prevent error when time partitioning is populated with empty dict

tswast:
As reported in internal bug 131167013, calling `table.time_partitioning`
can sometimes fail with `KeyError: 'type'` when attempting to read this
property on a non-partitioned table.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -1787,7 +1787,7 @@ class TimePartitioning(object):
         """google.cloud.bigquery.table.TimePartitioningType: The type of time
         partitioning to use.
         """
-        return self._properties["type"]
+        return self._properties.get("type")
 
     @type_.setter
     def type_(self, value):
@@ -1849,7 +1849,7 @@ class TimePartitioning(object):
             google.cloud.bigquery.table.TimePartitioning:
                 The ``TimePartitioning`` object.
         """
-        instance = cls(api_repr["type"])
+        instance = cls()
         instance._properties = api_repr
         return instance
 
--- a/bigquery/tests/unit/test_table.py
+++ b/bigquery/tests/unit/test_table.py
@@ -2260,8 +2260,6 @@ class TestTimePartitioning(unittest.TestCase):
         self.assertTrue(time_partitioning.require_partition_filter)
 
     def test_from_api_repr_empty(self):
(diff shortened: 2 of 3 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -1849,7 +1849,7 @@ def from_api_repr(cls, api_repr):
             google.cloud.bigquery.table.TimePartitioning:
                 The ``TimePartitioning`` object.
         """
-        instance = cls(api_repr["type"])
+        instance = cls()
```
tseaver: Shouldn't this preserve the `type` from `api_repr`, if present?  Otherwise, the ctor is going to overwrite it with `TimePartitioningType.DAY`.

```suggestion
        instance = cls(api_repr.get("type"))
```

bigquery/tests/unit/test_table.py
```diff
@@ -871,6 +871,54 @@ def test__build_resource_w_custom_field_not_in__properties(self):
...
+        # Even though there are required properties according to the API
+        # specification, sometimes time partitioning is populated as an empty
+        # object. See internal bug 131167013.
+        table._properties["timePartitioning"] = {}
+        self.assertIsInstance(table.time_partitioning, TimePartitioning)
+        self.assertIsNone(table.time_partitioning.type_)
+        self.assertIsNone(table.time_partitioning.field)
+        self.assertIsNone(table.time_partitioning.expiration_ms)
+        self.assertIsNone(table.time_partitioning.require_partition_filter)
+
```
tseaver: We need a test for `TimePartitioning.from_api_repr` that exercises the bug I corrected above (stomping on the type from the resource).

bigquery/google/cloud/bigquery/table.py
tswast: I originally had what you suggest, but then I realized whatever we pass in will get overwritten by the next line which overwrites the whole `_properties` dictionary.

bigquery/tests/unit/test_table.py
tswast: Good call, done in 14a0495.

bigquery/google/cloud/bigquery/table.py
tseaver: I missed that.

PR #7912, opened 2019-05-09
Firestore: use correct environment variable to guard the 'system' part.

tseaver:
Firestore's system tests use a [different environment variable](https://github.com/googleapis/google-cloud-python/blob/a65ed778770e045c0b26bb98d64cd0beec25703e/firestore/tests/system.py#L37) (because firestore vs. datastore hysterical raisins).

### DIFF
--- a/firestore/noxfile.py
+++ b/firestore/noxfile.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 import os
+import shutil
 
 import nox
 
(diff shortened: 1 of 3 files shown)


firestore/noxfile.py
```diff
@@ -139,22 +138,3 @@ def cover(session):
...
-    session.run(
-        'sphinx-build',
-        '-W',  # warnings as errors
-        '-T',  # show full traceback on exception
-        '-N',  # no colors
-        '-b', 'html',
-        '-d', os.path.join('docs', '_build', 'doctrees', ''),
-        os.path.join('docs', ''),
-        os.path.join('docs', '_build', 'html', ''),
-    )
```
busunkim96: Does the docs session need to be removed as well?

tseaver: I would prefer not:  that seems to be some kind of wonk in the templating.

tseaver: I have regenerated / rebased to pick up the new `noxfile.py` template.

tseaver:
@busunkim96 I fixed the `docs` session issue by regenerating with an up-to-date version of sythtool.  PTAL.

PR #7913, opened 2019-05-09
Add system tests for Vision Product Search. Test product set resource…

brendanlundy:
Add system tests for Vision Product Search. Test product set resource management.

### DIFF
--- a/vision/tests/system.py
+++ b/vision/tests/system.py
@@ -33,6 +33,10 @@ LABEL_FILE = os.path.join(_SYS_TESTS_DIR, "data", "car.jpg")
 LANDMARK_FILE = os.path.join(_SYS_TESTS_DIR, "data", "landmark.jpg")
 TEXT_FILE = os.path.join(_SYS_TESTS_DIR, "data", "text.jpg")
 FULL_TEXT_FILE = os.path.join(_SYS_TESTS_DIR, "data", "full-text.jpg")
+USER_PROJECT = os.environ.get("GOOGLE_CLOUD_TESTS_USER_PROJECT")
+# Valid locations include 'us-west1', 'us-east1', 'asia-east1', 'europe-west1'
+LOCATION = os.environ.get("GOOGLE_CLOUD_TESTS_LOCATION")
+
 
 
 class VisionSystemTestBase(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


vision/tests/system.py
```diff
@@ -33,6 +33,10 @@
 LANDMARK_FILE = os.path.join(_SYS_TESTS_DIR, "data", "landmark.jpg")
 TEXT_FILE = os.path.join(_SYS_TESTS_DIR, "data", "text.jpg")
 FULL_TEXT_FILE = os.path.join(_SYS_TESTS_DIR, "data", "full-text.jpg")
+USER_PROJECT = os.environ.get("GOOGLE_CLOUD_TESTS_USER_PROJECT")
```
busunkim96: If `USER_PROJECT` can be specified via project id, could this be changed to `PROJECT_ID`? (Our CI already sets the environment variable` PROJECT_ID`)

vision/tests/system.py
```diff
@@ -33,6 +33,10 @@
...
+# Valid locations include 'us-west1', 'us-east1', 'asia-east1', 'europe-west1'
+LOCATION = os.environ.get("GOOGLE_CLOUD_TESTS_LOCATION")
```
busunkim96: Is it alright for tests to run against only one of these locations?

vision/tests/system.py
tseaver: @busunkim96 This looks to be copied from the `storage` system tests, where `USER_PROJECT` is specifically used to test the "requester pays" features.

@brendanlundy I don't think you actually mean to use that feature in any way (`vision` doesn't support it, AFAIK).

vision/tests/system.py
brendanlundy: Yes, running against 1 location is fine. I removed the environment variable and hard-coded "us-west1" as the location.

vision/tests/system.py
brendanlundy: I changed the name of the environment variable to PROJECT_ID. 

However, when I ran the tests with `nox -f vision/noxfile.py -s system-3.7` , PROJECT_ID was not set. Am I running the tests incorrectly? I still had to `export PROJECT_ID='my-project'
` to get the tests to run.

busunkim96: Yes! The environment variable needs to be set when running tests locally. I'll kick off a Kokoro job for Vision to make sure the tests run there.

busunkim96:
@brendanlundy Would you mind doing a `nox -s blacken` to re-format the file?

vision/tests/system.py
busunkim96: Looks like they are all running: https://source.cloud.google.com/results/invocations/03c8cf96-fdc4-4783-848a-7868b8a7d42f/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fvision/log

brendanlundy:
> @brendanlundy Would you mind doing a `nox -s blacken` to re-format the file?
Done.

PR #7914, opened 2019-05-09
Pub/Sub regen: Removing Beta labels, adding kms

sduskis:


sduskis:
It looks like the proto changes didn't propagate correctly.  Something must be wrong with my machine.  I'll close this PR, and wait for Autosynth.

PR #7948, opened 2019-05-10
PubSub: Fix streaming pull incorrectly handling FlowControl max_messages setting

### LINKED ISSUE #7677: Pub/Sub: Flow Control does not work as expected.
The flow control settings for the Cloud Pub/Sub subscriber should bound the number of unacknowledged messages outstanding to the callback at a time. However, it seems to have no effect on the client, where the number of outstanding messages always ends up being 10.

OS type and version: Mac OS X 10.14.3, Debian Linux
Python version: Python 2.7.10 on Mac, Python 3.6.5 on Linux
google-cloud-pubsub version:  0.40.0

#### Steps to reproduce

  1. Create a Cloud Pub/Sub topic and subscription
  2. Publish many messages to the topic
  3. Update the project name and subscription name in the code below.
  4. Run the code.

Expected behavior: Since the flow control settings have `max_messages` set to 2, only two "Received Message" lines should print at a time, with two "Acking Message" lines printing out 20 seconds later. Instead, 10 messages come to the callback at a time. Changing the value of `max_messages` has no impact on the number of messages sent to the callback.

#### Code example
```py
import time

project_name=<insert project name here>
subscription_name=<insert subscription name here>
subscription_path = subscriber.subscription_path(project_name, subscription_name)

def callback(message):
    print ("Received Message")
    time.sleep(20)
    print ("Acking Message")
    message.ack()

flow_control = pubsub_v1.types.FlowControl(max_messages=2)
subscriber.subscribe(
    subscription_path, callback=callback, flow_control=flow_control)

print('Subscriber started for {}'.format(subscription_path))
while True:
    time.sleep(60)
```

plamut:
Closes #7677.

This PR fixes how subscriber client handles the [FlowControl.max_messages](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/types.py#L56) setting. It makes sure that subscriber only delivers (i.e. invokes callbacks) `max_messages` at a time, and resumes invoking callbacks only when the user code acknowledges some of the previously delivered messages.

### How to test

**Steps to reproduce:**
- Publish 50 - 100 messages to that topic in batches of random size - [example script](https://github.com/plamut/google-cloud-python/blob/iss-7677-debug-app/pubsub/reproduce_pub.py).
- Run a test subscriber script ([example](https://github.com/plamut/google-cloud-python/blob/iss-7677-debug-app/pubsub/reproduce_sub.py)) that subscribes to the topic using a streaming pull.
- (optional) Apply a [patch](https://gist.github.com/plamut/b6f6851dc68212b382ddc02b877b577e) that adds colors to the log output, makes the latter easier to read.

- The background consumer does not always get paused in time and pulls an additional batch of messages from the server:
log_incorrectly_paused
- The client subscriber script often receives too many messages simultaneously, i.e. more than `FlowControl.max_messages`.

- The background consumer always gets paused in time when the leased messages load hits the limit (`max_messages`):
log_correctly_paused
- The client subscriber script never has more than `max_messages` pending messages (received, but not yet acknowledged) - the `"pending ACK"` figure in the logs. The streaming pull process waits until some of the messages are ACK-ed before delivering new messages.
- Excessive messages pulled from the server (i.e. those that would cause an overload) are held in an internal buffer, and are released to the client script before any new messages get pulled from the server. The streaming pull process does not temporarily hold any more messages than necessary.

### Footnotes
For `max_messages` settings greater than 10, one might observe that a maximum of 10 (or less) messages are delivered to the client code at once, especially if message callbacks are slow, e.g. by artificially sleeping for prolonged periods of time.

This is caused by the [default callback thread pool](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/scheduler.py#L67-L73) size, which has `max_workers` set to 10.

This can be mitigated by creating the subscriber client with a custom `scheduler` that uses a thread poll with a higher cap on the worker thread count, or by using [batch callbacks](https://github.com/googleapis/google-cloud-python/issues/4994) for received messages. The latter feature is almost done, and will submit a PR for it soon.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
@@ -36,8 +36,14 @@ _LeasedMessage = collections.namedtuple("_LeasedMessage", ["added_time", "size"]
 class Leaser(object):
     def __init__(self, manager):
         self._thread = None
-        self._operational_lock = threading.Lock()
         self._manager = manager
+
+        # a lock used for start/stop operations, protecting the _thread attribute
+        self._operational_lock = threading.Lock()
+
+        # A lock ensuring that add/remove operations are atomic and cannot be
+        # intertwined. Protects the _leased_messages and _bytes attributes.
+        self._add_remove_lock = threading.Lock()
 
         self._leased_messages = {}
         """dict[str, float]: A mapping of ack IDs to the local time when the
@@ -64,30 +70,32 @@ class Leaser(object):
 
     def add(self, items):
         """Add messages to be managed by the leaser."""
-        for item in items:
-            # Add the ack ID to the set of managed ack IDs, and increment
-            # the size counter.
-            if item.ack_id not in self._leased_messages:
-                self._leased_messages[item.ack_id] = _LeasedMessage(
-                    added_time=time.time(), size=item.byte_size
-                )
-                self._bytes += item.byte_size
-            else:
-                _LOGGER.debug("Message %s is already lease managed", item.ack_id)
+        with self._add_remove_lock:
+            for item in items:
+                # Add the ack ID to the set of managed ack IDs, and increment
+                # the size counter.
+                if item.ack_id not in self._leased_messages:
+                    self._leased_messages[item.ack_id] = _LeasedMessage(
+                        added_time=time.time(), size=item.byte_size
+                    )
+                    self._bytes += item.byte_size
+                else:
+                    _LOGGER.debug("Message %s is already lease managed", item.ack_id)
 
     def remove(self, items):
         """Remove messages from lease management."""
-        # Remove the ack ID from lease management, and decrement the
-        # byte counter.
-        for item in items:
-            if self._leased_messages.pop(item.ack_id, None) is not None:
-                self._bytes -= item.byte_size
-            else:
-                _LOGGER.debug("Item %s was not managed.", item.ack_id)
+        with self._add_remove_lock:
+            # Remove the ack ID from lease management, and decrement the
+            # byte counter.
+            for item in items:
+                if self._leased_messages.pop(item.ack_id, None) is not None:
+                    self._bytes -= item.byte_size
+                else:
+                    _LOGGER.debug("Item %s was not managed.", item.ack_id)
 
-        if self._bytes < 0:
-            _LOGGER.debug("Bytes was unexpectedly negative: %d", self._bytes)
-            self._bytes = 0
+            if self._bytes < 0:
+                _LOGGER.debug("Bytes was unexpectedly negative: %d", self._bytes)
+                self._bytes = 0
 
     def maintain_leases(self):
         """Maintain all of the leases being managed.
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -21,6 +21,7 @@ import threading
 
 import grpc
 import six
+from six.moves import queue
 
 from google.api_core import bidi
 from google.api_core import exceptions
@@ -115,6 +116,11 @@ class StreamingPullManager(object):
             )
         else:
             self._scheduler = scheduler
+
+        # A FIFO queue for the messages that have been received from the server,
+        # but not yet added to the lease management (and not sent to user callback),
+        # because the FlowControl limits have been hit.
+        self._messages_on_hold = queue.Queue()
 
         # The threads created in ``.open()``.
         self._dispatcher = None
@@ -217,7 +223,11 @@ class StreamingPullManager(object):
                 self._consumer.pause()
 
     def maybe_resume_consumer(self):
-        """Check the current load and resume the consumer if needed."""
+        """Check the load and held messages and resume the consumer if needed.
+
+        If there are messages held internally, release those messages before
+        resuming the consumer. That will avoid leaser overload.
+        """
         # If we have been paused by flow control, check and see if we are
         # back within our limits.
         #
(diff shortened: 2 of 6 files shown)


plamut:
Rebased because of a merge conflict when #7954 was merged.

plamut:
Putting on hold, because the change is a non-trivial one, and we want to make a PubSub release first. We will unblock the PR once done (reviews still welcome, though).

cc: @sduskis

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -210,26 +217,71 @@ def add_close_callback(self, callback):
...
-        if self._consumer is None or not self._consumer.is_paused:
-            return
-
-        if self.load < self.flow_control.resume_threshold:
-            self._consumer.resume()
-        else:
-            _LOGGER.debug("Did not resume, current load is %s", self.load)
+        """Check the load and held messages and resume the consumer if needed.
+
+        If there are messages held internally to avoid the leaser over load,
```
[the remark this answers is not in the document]
plamut: That's indeed better, updated.

pubsub/tests/system.py
```diff
@@ -206,6 +208,80 @@ class CallbackError(Exception):
...
+        # create a topic and subscribe to it
+        publisher.create_topic(topic_path)
+        subscriber.create_subscription(subscription_path, topic_path)
+
+        batch_sizes = (7, 4, 8, 2, 10, 1, 3, 8, 6, 1)  # total: 50
+        self._publish_messages(publisher, topic_path, batch_sizes=batch_sizes)
+
+        # now subscribe and do the main part, check for max pending messages
+        total_messages = sum(batch_sizes)
+        flow_control = types.FlowControl(max_messages=5)
```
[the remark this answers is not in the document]
plamut: Indeed, that's the part that keeps track of how many messages at once the client code has.

pubsub/tests/system.py
```diff
@@ -206,6 +208,80 @@ class CallbackError(Exception):
...
+            subscription_path, callback, flow_control=flow_control
+        )
+
+        # Expected time to process all messages in ideal case:
+        #     total_messages / FlowControl.max_messages * processing_time
+        #
+        # With total=50, max messages=5, and processing_time=1 this amounts to
+        # 10 seconds (+ overhead), thus a full minute should be more than enough
+        # for the processing to complete. If not, fail the test with a timeout.
+        for _ in six.moves.range(60):
```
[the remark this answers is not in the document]
plamut: That's a good suggestion. I encapsulated the loop into a future on the `callback`, makes this part of the test code cleaner, thanks.

pubsub/tests/system.py
```diff
@@ -206,6 +208,80 @@ class CallbackError(Exception):
...
+        total_messages = sum(batch_sizes)
+        flow_control = types.FlowControl(max_messages=5)
+        callback = StreamingPullCallback(processing_time=1)
+
+        future = subscriber.subscribe(
+            subscription_path, callback, flow_control=flow_control
+        )
+
+        # Expected time to process all messages in ideal case:
+        #     total_messages / FlowControl.max_messages * processing_time
```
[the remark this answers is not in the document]
plamut: While mathematically the same, the latter catches the semantics better. `(total_messages / max_messages)` represents the lower bound on the number of batches the client code needs to process, and processing each batch takes approximately `processing_time` (the time to process a single message).

I added parentheses to the formula for a bit more semantic clarity.

plamut:
@sduskis Made the changes suggested, the diff can be seen in the [first force push](https://github.com/googleapis/google-cloud-python/compare/b25c427900e0cf5287ebcd7a30bde16de28a45ed..1b8a212c09955f21ce138c54812376cdde191b75).

(the second force push was just rebasing on top of the latest `master`).

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -64,30 +64,32 @@ def bytes(self):
...
-            # Add the ack ID to the set of managed ack IDs, and increment
-            # the size counter.
-            if item.ack_id not in self._leased_messages:
-                self._leased_messages[item.ack_id] = _LeasedMessage(
-                    added_time=time.time(), size=item.byte_size
-                )
-                self._bytes += item.byte_size
-            else:
-                _LOGGER.debug("Message %s is already lease managed", item.ack_id)
+        with self._operational_lock:
```
tseaver: From my read of the code, this lock is normally used to protect access to the `_thread` attrbute, created in `start` and destroyed in `stop`.  The worker thread runs `maintain_leases`, which copies the `_leased_messages` dict before iterating it.  What race are we trying to prevent here?  If `add`, `remove`, and `maintain_leases` need to synchronize access to `_leased_messages`, then perhaps we need a different lock for that?

tseaver: More generally:  we could improve auditability / comprehensibility of the M/T code by adding a comment for each lock created, documenting what attributes it is supposed to guard.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -116,6 +117,12 @@ def __init__(
         else:
             self._scheduler = scheduler
 
+        # A FIFO queue for the messages that have been received from the server,
+        # but not yet added to the lease management (and not sent to user callback),
+        # because the FlowControl limits have been hit.
+        self._messages_on_hold = queue.Queue()
```
tseaver: Do we need to set a maximum size on the queue?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -211,26 +218,70 @@ def add_close_callback(self, callback):
...
+    def _maybe_release_messages(self):
+        """Release (some of) the held messages if the current load allows for it.
+
+        The method tries to release as many messages as the current leaser load
+        would allow. Each released message is added to the lease management,
+        and the user callback is scheduled for it.
+
+        If there are currently no messageges on hold, or if the leaser is
+        already overloaded, this method is effectively a no-op.
+        """
```
tseaver: Maybe add a comment here that the method assumes the caller has acquired the `_pause_resume_lock`?

pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -208,8 +212,8 @@ def lease(self):
         """Inform the policy to lease this message continually.
 
         .. note::
-            This method is called by the constructor, and you should never
-            need to call it manually.
+            By default this method is called by the constructor, and you should
+            never need to call it manually.
```
tseaver: Is "never" correct here?  What if the application constructs the message with `lease=False`?

plamut: Re-reading it again, it is not - it refers to the default case, but that's not clear from the sentence, will reword.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
plamut: Would not hurt, and explicit is better than implicit.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
plamut: We don't - the queue only gets filled in the `_on_response()` callback, and only after the load reaches 1.0. At that point the consumer is already paused and no more messages arrive.

Additionally, the consumer only gets maybe-resumed after clearing the queue first, meaning that the maximum queue size is bounded by the maximum number of messages the server sends in a single response.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
plamut: I had the impression that "operational lock" is meant to synchronize "operations" on the leaser that can be initiated from external threads (e.g. `start` and `stop`), and used it to make the `add` / `remove` operations atomic. If this causes confusion, it might indeed make sense to use a separate lock, e.g. named `add_remove_lock`?

The motivation behind this is the observation in debug logs that `add()` and `remove()` operations can be intertwined if invoked from different threads at just the "right" moment. Both operations modify the internal `_leased_messages` dict and the `_bytes` counter.

The dict is probably fine, because `ack_id` is unique for every message, but I wouldn't say the same for `_bytes`.

plamut:
@tseaver Addressed the comments, please have another look. The [diff](https://github.com/googleapis/google-cloud-python/compare/7e5b9d778b8d507a634ef6e0b06e924be2d4ddfe..ec6a21d938ce1e09b2dc583ac78dd6ab36dab8a6) is small enough, thus I changed a few commits and force-pushed the changes.

Also removing the `do not merge` label, since PubSub version `0.41.0` has been successfully released in the meantime.

Issue #7762, opened 2019-04-22
Storage: redesign proposal


tswast:
The Cloud Storage Python client was one of the first google-cloud-python clients, intended to be more reliable and more Pythonic than the google-api-client that it replaced. It is also a 100% hand-written client. Because it predates the auto-generated google-cloud-python clients, the Cloud Storage Python client has inconsistencies with other google-cloud-python clients.

This design proposal is to bring the Cloud Storage client further into alignment with other google-cloud-python clients and make the eventual inclusion of an auto-generated gRPC transport layer less disruptive to users of the Cloud Storage client.

**Design document:**

https://docs.google.com/document/d/1A4FIxThZK_enK7OV9ChY54IPrlpbdHwuhvVEu6iTHxM/edit?usp=sharing

Please share your feedback on the specifics of this as comments in the [design document](https://docs.google.com/document/d/1A4FIxThZK_enK7OV9ChY54IPrlpbdHwuhvVEu6iTHxM/edit?usp=sharing).

CC @crwilcox @frankyn @lbristol88

PR #7949, opened 2019-05-10
Phase 1 Added new method download_blob_to_file within Storage Client

lbristol88:
Part of issue: #7762

@tswast
@engelke

### DIFF
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -14,6 +14,7 @@
 
 """Client for interacting with the Google Cloud Storage API."""
 
+from six.moves.urllib.parse import urlsplit
 
 from google.auth.credentials import AnonymousCredentials
 
@@ -24,6 +25,7 @@ from google.cloud.exceptions import NotFound
 from google.cloud.storage._http import Connection
 from google.cloud.storage.batch import Batch
 from google.cloud.storage.bucket import Bucket
+from google.cloud.storage.blob import Blob
 
 
 _marker = object()
@@ -308,6 +310,16 @@ class Client(ClientWithProject):
             bucket.requester_pays = requester_pays
         bucket.create(client=self, project=project)
         return bucket
+
+    def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
+        try:
+            blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
+        except AttributeError:
+            scheme, netloc, path, query, frag = urlsplit(blob_or_uri)
+            bucket = Bucket(self, name=netloc)
+            blob_or_uri = Blob(path, bucket)
+
+            blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
 
     def list_buckets(
         self,
--- a/storage/google/cloud/storage/client.py
+++ b/storage/google/cloud/storage/client.py
@@ -312,6 +312,21 @@ class Client(ClientWithProject):
         return bucket
 
     def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
+        """Download the contents of a blob object or blob URI into a file-like object.
+
+        Args:
+            blob_or_uri (Union[ \
+                :class:`~google.cloud.storage.blob.Blob`, \
+                 str, \
+            ]):
+                The blob resource to pass or URI to download.
+            file_obj (file):
+                A file handle to which to write the blob's data.
+            start (int):
+                Optional, the first byte in a range to be downloaded.
+            end (int):
+                Optional, The last byte in a range to be downloaded.
+        """
         try:
             blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
         except AttributeError:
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/client.py
```diff
@@ -309,6 +311,31 @@ def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
         bucket.create(client=self, project=project)
         return bucket
 
+    def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
+        """Download the contents of a blob object or blob URI into a file-like object.
+
+        Args:
+            blob_or_uri (Union[ \
+                :class:`~google.cloud.storage.blob.Blob`, \
+                 str, \
```
tswast: Nit: This doesn't quite line up with the line above it (there's one extra space here)

storage/google/cloud/storage/client.py
```diff
@@ -309,6 +311,31 @@ def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
...
+            start (int):
+                Optional, the first byte in a range to be downloaded.
+            end (int):
+                Optional, The last byte in a range to be downloaded.
+        """
+        try:
+            blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
+        except AttributeError:
+            scheme, netloc, path, query, frag = urlsplit(blob_or_uri)
+            bucket = Bucket(self, name=netloc)
```
tswast: Let's raise a `ValueError` if `scheme` isn't `gs`. I was warned that people might start passing in `http` URLs and expect this to work.

storage/google/cloud/storage/client.py
```diff
@@ -309,6 +311,31 @@ def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
...
 
+    def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
+        """Download the contents of a blob object or blob URI into a file-like object.
+
+        Args:
+            blob_or_uri (Union[ \
+                :class:`~google.cloud.storage.blob.Blob`, \
+                 str, \
+            ]):
+                The blob resource to pass or URI to download.
```
tswast: Let's give an example of the URI `gs://bucket_name/path/to/blob`

storage/google/cloud/storage/client.py
```diff
@@ -309,6 +311,31 @@ def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
...
+            file_obj (file):
+                A file handle to which to write the blob's data.
+            start (int):
+                Optional, the first byte in a range to be downloaded.
+            end (int):
+                Optional, The last byte in a range to be downloaded.
+        """
+        try:
+            blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
+        except AttributeError:
```
tswast: What's the reason for looking for `AttributeError` rather than checking `isinstance` like you did in `create_bucket`?

lbristol88:
> Looking good! A few comments on how to improve the docstrings. Your approach looks good, though I'm curious why you didn't follow the same pattern as you did in the `create_bucket` function.

I had initially followed the pattern and was using IsInstance, but ran into issues when I was attempting to the function. Andrew suggested using try/except which would be more Pythonic and easier to test.

storage/google/cloud/storage/client.py
```diff
@@ -316,21 +316,47 @@ def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
...
+
+
+            Download a blob using a URI.
+
+            >>> from google.cloud import storage
+            >>> client = storage.Client()
+
+            >>> with open('file-to-download-to') as file_obj:
+            >>>     client.download_blob_to_file(
+            >>>         'gs::/bucket_name/path/to/blob', file)
```
tswast: Typo. Should be two slashes, one colon. `gs://`

storage/google/cloud/storage/client.py
tswast: We'll need a test for this check. (Try a `https://` link and verify that `ValueError` is raised with `pytest.raises`

storage/google/cloud/storage/client.py
lbristol88: Fixed!

storage/google/cloud/storage/client.py
lbristol88: I added an `if` statement for raising `ValueError` if scheme isn't `gs` as directed. I also added the test for the change.

storage/google/cloud/storage/client.py
lbristol88: Added examples for URI and if passing in a resource object.

storage/google/cloud/storage/client.py
lbristol88: Fixed typo!

lbristol88:
All nox tests pass.

```
All done! ✨ 🍰 ✨
27 files left unchanged.
nox > Session blacken was successful.
nox > Ran multiple sessions:
nox > * default: success
nox > * unit-2.7: success
nox > * unit-3.5: success
nox > * unit-3.6: success
nox > * unit-3.7: success
nox > * system-2.7: success
nox > * system-3.6: success
nox > * cover: success
nox > * lint: success
nox > * lint_setup_py: success
nox > * blacken: success
```

PR #7954, opened 2019-05-13
PubSub: Propagate subscribe callback errors to main thread

### LINKED ISSUE #7917: PubSub: Subscriber future not raising exception that happen in callbacks
A user reported the following in an email message (slightly adjusted for readability):

---
To sum up this issue, the [documentation](https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/index.html#subscription-callbacks) on subscription callbacks says the following:

> You can also use this for error handling; any exceptions that crop up on a thread will be set on the future.

```py
try:
    future.result()
except Exception as ex:
    subscription.close()
    raise
```

Unfortunately, this is not the case, because there is a try/except that catches every exception and never re-raises it to propagate it up the chain. Specifically, the exception is silenced here:

https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L55-L72

I understand the built in `nack()`, that definitely makes sense. However, either the documentation or the code needs to change.

You can observe the behavior with a simple code snippet below, and see that the main thread never sees the exception raised, and does not die:

```py
import sys

SUBNAME = "<SUBNAME>"

def callback(message):
    # I just raise an exception
    raise Exception

sub = pubsub_v1.SubscriberClient()   
fut = sub.subscribe(SUBNAME, callback)

try:
    fut.result()
except Exception:
    sys.exit(1)
```
---

### Comment
Callbacks are dispatched asynchronously using a background thread pool, and are not executed in the main thread. The future returned by the `subscriber.subscribe()` call is a [StreamingPullFuture](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/futures.py#L20), which is only resolved when the underlying streaming pull manager terminates. However, the manager currently does not know anything about that future, and thus cannot interact with it.

**To discuss:**
How do other PubSub clients handle exceptions in user-provided callbacks, do they propagate them to the main thread?

One possible way of supporting this functionality would be to modify the code where the streaming pull manager and the future [are created](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/client.py#L206-L210) (or the `StreamingPullFuture` code itself). The manager would need to be given a reference to the future in order to resolve it with an exception, should one occur in the provided callback.

plamut:
Closes #7917.

This PR implements a user request to propagate exceptions from streaming pull callbacks to the main thread awaiting the result of the future returned by the `subscriber.subscribe()` call.

### How to test

**Steps to reproduce:**
- Publish a message to the topic.
- Run the sample subscriber code posted under the [issue](https://github.com/googleapis/google-cloud-python/issues/7917) (adjust subscription name as necessary).

- Nothing happens in the main thread. The exception is raised in the callback, but the main thread keeps waiting for for `fut.result()`.

- An exception from the callback is propagated to `fut.result()`, and the main thread exists almost immediately after the callback is invoked.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -52,7 +52,7 @@ def _maybe_wrap_exception(exception):
     return exception
 
 
-def _wrap_callback_errors(callback, message):
+def _wrap_callback_errors(callback, on_callback_error, message):
     """Wraps a user callback so that if an exception occurs the message is
     nacked.
 
@@ -62,14 +62,15 @@ def _wrap_callback_errors(callback, message):
     """
     try:
         callback(message)
-    except Exception:
+    except Exception as exc:
         # Note: the likelihood of this failing is extremely low. This just adds
         # a message to a queue, so if this doesn't work the world is in an
         # unrecoverable state and this thread should just bail.
         _LOGGER.exception(
-            "Top-level exception occurred in callback while processing a " "message"
+            "Top-level exception occurred in callback while processing a message"
         )
         message.nack()
+        on_callback_error(exc)
 
 
 class StreamingPullManager(object):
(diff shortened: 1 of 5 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -299,21 +300,26 @@ def heartbeat(self):
         if self._rpc is not None and self._rpc.is_active:
             self._rpc.send(types.StreamingPullRequest())
 
-    def open(self, callback):
+    def open(self, callback, on_callback_error):
```
plamut: FWIW, this method signature change is safe, as `StreamingPullManager` class is only used internally in the subscriber client's `subscribe()` method, and is not mentioned anywhere in the [user docs](https://googleapis.github.io/google-cloud-python/latest/pubsub/), at least from what I can see.

PR #7984, opened 2019-05-15
WIP: Bigtable : Remove channel and add new transport

sangramql:
See: #6278

Initial wiring for transport.

### DIFF
--- a/bigtable/google/cloud/bigtable/client.py
+++ b/bigtable/google/cloud/bigtable/client.py
@@ -275,7 +275,7 @@ class Client(ClientWithProject):
     def table_data_transport(self):
         """Getter for the gRPC stub used for the Table Admin Transport API.
 
-        :rtype: :class:`.bigtable_grpc_transport.BigtableGrpcTransport
+        :rtype: :class:`.bigtable_grpc_transport.BigtableGrpcTransport`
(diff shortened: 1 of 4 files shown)


sangramql:
@sduskis @tseaver @AVaksman 
Please review the initial approach for transport.

bigtable/google/cloud/bigtable/client.py
```diff
@@ -130,6 +133,7 @@ def __init__(
         read_only=False,
         admin=False,
         client_info=_CLIENT_INFO,
+        transport=None,
```
sduskis: There needs to be 2 new transports: `data_transport` and `admin_transport`.  (Table and Instance admin operations go to the same end point).  A `transport` is basically a method pointer.  Please see [here](https://github.com/googleapis/google-cloud-python/blob/master/bigtable/google/cloud/bigtable_admin_v2/gapic/bigtable_table_admin_client.py#L192) for an example of where an admin transport would be used

sangramql: Got it. Thank you for the pointer.
Actually I got confused when I was going through and comparing bigtable_instance_admin_grpc_transport.py and bigtable_instance_admin_client.py and thought those can be used independently.

Issue #7968, opened 2019-05-14
Release the world to support 'client_info' for manual clients.


tseaver:
/cc @crwilcox, @busunkim96, @tswast 

Follow-on to #7825.  Because we are releasing new features in `google-api-core` and (more importantly) `google-cloud-core`, we need to handle this release phase delicately.  Current clients which depend on `google-cloud-core` use a too-narrow pin:

```bash
$ grep google-cloud-core */setup.py | grep -v "^core"
bigquery/setup.py:    "google-cloud-core >= 0.29.0, < 0.30dev",
bigtable/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
datastore/setup.py:    'google-cloud-core >=0.29.0, <0.30dev',
dns/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
firestore/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
logging/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
resource_manager/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
runtimeconfig/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
spanner/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
storage/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
trace/setup.py:    'google-cloud-core >=0.29.0, <0.30dev',
translate/setup.py:    'google-cloud-core >= 0.29.0, < 0.30dev',
```

Per conversation today, we plan to go ahead and release a `1.0.0` of `google-cloud-core`.  *Before* that, we need to make releases from the last tags of the clients above which broaden the range to `google-cloud-core >= 0.29.0, < 2.0dev`.

### Prep Releases

For each of the following:

- [x] `bigquery` (#7969)
- [x] `bigtable` (#7970)
- [x] `datastore` (#7971)
- [x] `dns` (#7972)
- [x] `firestore` (#7973)
- [x] `logging` (#7974)
- [x] `resource_manager` (#7975)
- [x] `runtimeconfig` (#7976)
- [x] `spanner` (#7977)
- [x] `storage` (#7978)
- [x] `trace` (#7979)
- [x] `translate` (#7980)

the procedure is:

- Make a "release branch" from the last tag, e.g. `$ git checkout -b bigquery-1.11-back bigquery-1.11.2`.
[...]

PR #7987, opened 2019-05-15
Translation: Change system test.

busunkim96:
Fixes #7892.

`hvala ti` (Croatian) and `dankon` (Esperanto) mean `thank you`. This word seems to be susceptible to changing translations. I've changed them to `petnaest` and `dek kvin` which mean `fifteen`.

Issue #7982, opened 2019-05-15
Translation: Fix system test broken by back-end translation change.


busunkim96:
```
=================================== FAILURES ===================================
_________________________ TestTranslate.test_translate _________________________
self = <tests.system.TestTranslate testMethod=test_translate>
    def test_translate(self):
        values = ["hvala ti", "dankon", "Me llamo Jeff", "My name is Jeff"]
        translations = Config.CLIENT.translate(
            values, target_language="de", model="nmt"
        )
        self.assertEqual(len(values), len(translations))
        self.assertEqual(translations[0]["detectedSourceLanguage"].lower(), "hr")
>       self.assertEqual(translations[0]["translatedText"].lower(), "danke")
E       AssertionError: 'danke schön' != 'danke'
E       - danke schön
E       + danke
```

tseaver:
Fixed in #7987.

Issue #7900, opened 2019-05-08
Bigtable: 'test_create_instance_with_two_clusters' flakes modifying profile.


tseaver:
Similar to #5928, but the failure occurs while re-modifying the instance's app profile.

From [this Kokoro failure](https://source.cloud.google.com/results/invocations/b5f7a8bc-d02c-45f8-b23d-31b94e4d0493/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable/log):

```python
___________ TestInstanceAdminAPI.test_create_instance_w_two_clusters ___________
target = functools.partial(<bound method PollingFuture._done_or_raise of <google.api_core.operation.Operation object at 0x7f7c280cee80>>)
predicate = <function if_exception_type.<locals>.if_exception_type_predicate at 0x7f7c299b70d0>
sleep_generator = <generator object exponential_sleep_generator at 0x7f7c297d3a98>
deadline = 10, on_error = None
    def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
        """Call a function and retry if it fails.
        This is the lowest-level retry helper. Generally, you'll use the
        higher-level retry helper :class:`Retry`.
        Args:
            target(Callable): The function to call and retry. This must be a
                nullary function - apply arguments with `functools.partial`.
            predicate (Callable[Exception]): A callable used to determine if an
                exception raised by the target should be considered retryable.
                It should return True to retry or False otherwise.
            sleep_generator (Iterable[float]): An infinite iterator that determines
                how long to sleep between retries.
            deadline (float): How long to keep retrying the target.
            on_error (Callable): A function to call while processing a retryable
                exception.  Any error raised by this function will *not* be
                caught.
        Returns:
            Any: the return value of the target function.
        Raises:
            google.api_core.RetryError: If the deadline is exceeded while retrying.
            ValueError: If the sleep generator stops yielding values.
            Exception: If the target raises a method that isn't retryable.
        """
        if deadline is not None:
            deadline_datetime = datetime_helpers.utcnow() + datetime.timedelta(
                seconds=deadline
            )
        else:
            deadline_datetime = None
        last_exc = None
        for sleep in sleep_generator:
            try:
>               return target()
../api_core/google/api_core/retry.py:179:
[...]

PR #7993, opened 2019-05-15
Omnibus: pin new core 1.0.0

tseaver:
Toward #7968.


tseaver:
Flaky / broken system tests;

- Bigtable fails due to instances leaked from snippets (#7900).
- Storage is a new back-end failure (there has been an outage going on).
- Translate is #7982, fixed in PR #7987.

Issue #7111, opened 2019-01-10
Datastore: expose 'reserveIds' at top level.


Kazade:
#### Environment details

1. Datastore
2. Fedora 28
3. Python 3.6
4. google-cloud-datastore 1.7.3

#### Steps to reproduce

 The `Client` object doesn't expose the `reserve_ids` function (for preventing keys being accidentally reused by the auto-allocator), it does however expose the related `allocate_ids` (https://github.com/googleapis/google-cloud-python/blob/master/datastore/google/cloud/datastore/client.py#L487)

It would be great if this could be added, in the meantime I'm planning to use the private `Client._datastore_api.reserve_ids` method directly.

PR #8017, opened 2019-05-17
expose reserveIds

HemangChothani:
issue [7111] (https://github.com/googleapis/google-cloud-python/issues/7111)


tseaver:
@HemangChothani Your patch looks pretty good:  I'd like to have some extra assertions about the values passed, but otherwise fine.

PR #8022, opened 2019-05-17
PubSub: Add an option to streaming pull to invoke callbacks for batches of messages

### LINKED ISSUE #4994: pubsub API: allow registering a callback for batches of messages
The old API allowed you to operate on batches of messages. Sometimes, this is advantageous as you can aggregate other calls. With the current API, this is non trivial.

So, I think (if the underlying code supports it), adding, either a parameter to open(), or another function open_many() that'll always return a list (there could be a limit in the flow control), even if it's just with one item would be useful.

plamut:
Closes #4994.

---
## Do not merge yet!
This has been removed from GA for now ([comment](https://github.com/googleapis/google-cloud-python/issues/4994#issuecomment-493485665)), because other implementations currently do not have this feature.

---

This PR adds an option to receive asynchronously pulled messages in callbacks in batches, rather than a single message per each callback invocation.

> NOTE: This PR branch is based on #7948, because it needs to modify several parts of the code added by the latter.

### How to test

**Steps to reproduce:**
- Rapidly publish several messages to the topic - they must be sent to the PubSub server in the same batch (or several batches)
- Create a subscriber client and start the streaming pull with the argument `batch=True`:
```py
subscriber.subscribe(SUBSCRIPTION_NAME, callback, batch=True)
```

The provided `callback` is invoked once for each message.

The provided `callback` is invoked **once for each batch** of published message (subject to `FlowControl.max_messages` limits). The callback is invoked with a _list_ of messages as the argument.

### TODO list
- [x] Fix docs (format of some of the changed docstrings).

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
@@ -64,30 +64,32 @@ class Leaser(object):
 
     def add(self, items):
         """Add messages to be managed by the leaser."""
-        for item in items:
-            # Add the ack ID to the set of managed ack IDs, and increment
-            # the size counter.
-            if item.ack_id not in self._leased_messages:
-                self._leased_messages[item.ack_id] = _LeasedMessage(
-                    added_time=time.time(), size=item.byte_size
-                )
-                self._bytes += item.byte_size
-            else:
-                _LOGGER.debug("Message %s is already lease managed", item.ack_id)
+        with self._operational_lock:
+            for item in items:
+                # Add the ack ID to the set of managed ack IDs, and increment
+                # the size counter.
+                if item.ack_id not in self._leased_messages:
+                    self._leased_messages[item.ack_id] = _LeasedMessage(
+                        added_time=time.time(), size=item.byte_size
+                    )
+                    self._bytes += item.byte_size
+                else:
+                    _LOGGER.debug("Message %s is already lease managed", item.ack_id)
 
     def remove(self, items):
         """Remove messages from lease management."""
-        # Remove the ack ID from lease management, and decrement the
-        # byte counter.
-        for item in items:
-            if self._leased_messages.pop(item.ack_id, None) is not None:
-                self._bytes -= item.byte_size
-            else:
-                _LOGGER.debug("Item %s was not managed.", item.ack_id)
+        with self._operational_lock:
+            # Remove the ack ID from lease management, and decrement the
+            # byte counter.
+            for item in items:
+                if self._leased_messages.pop(item.ack_id, None) is not None:
+                    self._bytes -= item.byte_size
+                else:
+                    _LOGGER.debug("Item %s was not managed.", item.ack_id)
 
-        if self._bytes < 0:
-            _LOGGER.debug("Bytes was unexpectedly negative: %d", self._bytes)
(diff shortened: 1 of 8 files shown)


sduskis:
`google/cloud/pubsub_v1/subscriber/client.py:202:15: W291 trailing whitespace`

sduskis:
We should discuss this feature with the Pub/Sub team before merging.  They want to limit the difference between implementations in different languages, if possible.

c0rvax:
Hi, is there any ETA of merging this? That is a pretty vital decision i have to make in project. And the decision depends of this ETA. Thanks!

plamut:
@c0rvax I can only give an unofficial answer, as I am not the decision maker here, but AFAIK this PR is unlikely to be merged in the near future.

PubSub clients for other languages currently lack this feature, and consistency is desired across various implementations.

c0rvax:
@plamut 

Thanks for your answer, it helps a lot!

szeitlin:
It would be nice to have this feature, though, since java, ruby and go all have the streaming pull/concurrency feature, and python does not.

plamut:
@szeitlin The Python client also supports asynchronous (streaming) pull, and it's actually the preferred way of pulling the messages ([docs](https://cloud.google.com/pubsub/docs/pull#asynchronous-pull)). The thing here is just that each received message is dispatched individually, i.e. one message per each callback invocation, as opposed to multiple messages per invocation.

Or did I misunderstand your comment?

szeitlin:
We're trying the asynchronous approach, but ultimately we still have a use case where we'll need to write out messages in batch, which is going to be a bit of work to implement using the streaming pull method and checking to see when we've accumulated enough messages. I get the impression a few other people would appreciate this feature for the same reason, unless I'm misunderstanding how it's supposed to work?

plamut:
The batch callback, at least how it was implemented here, simply means that if N messages are available to dispatch to the user callback, all of them are passed to the callback in a list, as opposed to N callback invocations, one for each individual message.

The concern raised in internal discussions was that it is not defined what a "batch" actually is, and can thus be confusing to users. Publishing several messages in a single batch does not necessarily mean that the backend will actually deliver these messages in a single batch. The "received" batch can also contain messages from several different publish batches.

Since the number of messages received in a single batch might vary, you would still have to implement the aggregation logic to know when enough messages have been accumulated.

Just to avoid any misunderstanding, what would be your expectation of such feature, i.e. what messages would you expect to receive in a particular batch?

szeitlin:
Good question. I guess we were just hoping we could set a minimum batch size (in our case it's ok if things arrive out of order).

plamut:
The backend does not support that, unfortunately, thus any "minimum size" batching logic needs to be implemented on the client side. While the client library could do that internally, it cannot decide for the user what to do in edge cases, e.g. when no new messages arrive for a substantial period of time, but it does not yet hold enough messages to release the next batch.

One way to achieve batching is to put received messages into a queue which is then processed by a different thread that can then batch/process them as needed, e.g. write them out when there are enough of them in the queue.

```py
import queue

received_queue = queue.Queue()

...

future = subscriber.subscribe("my/subscription/path", callback=received_queue.put)

try:
    future.result()
except KeyboardInterrupt:
    future.cancel()
```

Would that help?

szeitlin:
This is an interesting point, we will try that if the way we have it isn't fast enough. Thanks!

PR #8037, opened 2019-05-20
Vision: add system tests for product resource management.

brendanlundy:


vision/tests/system.py
```diff
@@ -202,3 +205,98 @@ def test_update_product_set(self):
...
+            display_name="product display name", product_category="apparel"
+        )
+        product_id = "product" + unique_resource_id()
+        product_path = self.ps_client.product_path(
+            project=PROJECT_ID, location=self.location, product=product_id
+        )
+        self.products_to_delete.append(product_path)
+        response = self.ps_client.create_product(
+            parent=self.location_path, product=product, product_id=product_id
+        )
```
tseaver: Appending the path to the list of to-be-deleted products should happen after the `create_product` call succeeds:  otherwise, the tear-down will fail, potentially leaking stuff.

OTOH, the new code follows the pattern used in the rest of the module.

brendanlundy: I updated it to add to the delete list after the resource is created. 

Also use the resource.name just in case the resource name somehow ends up being different than what it should be.

brendanlundy:
I added a few more tests. This PR is ready for review.

PR #8039, opened 2019-05-20
Firestore: expand tildes in class refs FBO static analysis / autocomplete.

### LINKED ISSUE #8035: Firestore: links in docs, static code analysis and autocompletion in PyCharm are broken
Due to docstrings formatted like this, static code analysis and autocomplete in PyCharm and the links in the [online Sphinx docs](https://googleapis.github.io/google-cloud-python/latest/firestore/client.html#google.cloud.firestore_v1beta1.client.Client.document) are broken:
https://github.com/googleapis/google-cloud-python/blob/de0103cedaf9fb9f03ddca9be801ac64635d7b74/firestore/google/cloud/firestore_v1/client.py#L250-L252

This should result in the same short name with a working link in the Sphinx docs and would not kill autocomplete etc:
```
Returns:
    :class:`~google.cloud.firestore_v1.document.DocumentReference`: A reference
    to a document in a collection.
```

tseaver:
Generated via:

```bash
$ for f in $(git grep -l "~\." firestore/); do sed -i -e "s/~\./google.cloud./" $f; done
```

Closes #8035.

### DIFF
--- a/firestore/google/cloud/firestore_v1/_helpers.py
+++ b/firestore/google/cloud/firestore_v1/_helpers.py
@@ -220,11 +220,11 @@ def reference_value_to_document(reference_value, client):
 
     Args:
         reference_value (str): A document reference value.
-        client (~.firestore_v1.client.Client): A client that has
+        client (~google.cloud.firestore_v1.client.Client): A client that has
             a document factory.
 
     Returns:
-        ~.firestore_v1.document.DocumentReference: The document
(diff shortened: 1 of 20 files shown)


tseaver:
Re-pushed after I realized that my original `sed` statement had stripped the leading tildes.  New command:

```bash
$ for f in $(git grep -l "~\." firestore/); do sed -i -e "s/~\./~google.cloud./" $f; done
```

PR #8088, opened 2019-05-21
Bigtable: Add retry parameter to 'table.read_row()'.

### LINKED ISSUE #7955: bigtable: Implement consistent retry mechanisms
It is frustrating to utilize a provided retry mechanism for one API call, and being forced to implement our own retry mechanism for a very similar API call because retries aren't provided.

https://googleapis.github.io/google-cloud-python/latest/bigtable/table.html#google.cloud.bigtable.table.Table.read_row

https://googleapis.github.io/google-cloud-python/latest/bigtable/table.html#google.cloud.bigtable.table.Table.read_rows

`table.Table.read_rows` accepts a Retry, but `table.Table.read_row` does not. It would be nice if `table.Table.read_row` accepted the same retry mechanism.
 
In the absence of a retry mechanism built into read_row, we have written our own.

elisheva-qlogic:
Fix #7955

### DIFF
--- a/bigtable/tests/unit/test_table.py
+++ b/bigtable/tests/unit/test_table.py
@@ -569,9 +569,7 @@ class TestTable(unittest.TestCase):
         from google.cloud.bigtable.table import _BigtableRetryableError
 
         credentials = _make_credentials()
-        client = self._make_client(
-            project="project-id", credentials=credentials, admin=True
-        )
+        client = self._make_client(project="project-id", credentials=credentials)
         instance = client.instance(instance_id=self.INSTANCE_ID)
         table = self._make_one(self.TABLE_ID, instance)
         custom_retry = retry.Retry(
--- a/bigtable/tests/unit/test_table.py
+++ b/bigtable/tests/unit/test_table.py
@@ -563,90 +563,29 @@ class TestTable(unittest.TestCase):
         column.append(Cell.from_pb(chunk))
         self._read_row_helper(chunks, expected_result, app_profile_id)
 
-    def test_read_row_retry(self):
-        from google.cloud._testing import _Monkey
-        from google.cloud.bigtable import table as MUT
-        from google.cloud.bigtable.row_set import RowSet
-        from google.cloud.bigtable_v2.gapic import bigtable_client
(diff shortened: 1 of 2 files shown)


bigtable/tests/unit/test_table.py
```diff
@@ -563,6 +563,91 @@ def test_read_row_complete(self):
         column.append(Cell.from_pb(chunk))
         self._read_row_helper(chunks, expected_result, app_profile_id)
 
+    def test_read_row_retry(self):
```
sduskis: This test should be something like `test_read_row_custom_retry`, and should only test that the `retry_read_row` is passed correctly from `read_row` to `read_rows`.  There's a lot of other stuff in this method that is "orthogonal" to the specific thing you're trying to test, and therefore should be removed; streamlining the test will allow future reviewers to quickly understand the intent of the test.

bigtable/tests/unit/test_table.py
```diff
@@ -563,6 +563,30 @@ def test_read_row_complete(self):
...
         self._read_row_helper(chunks, expected_result, app_profile_id)
 
+    def test_read_row_custom_retry(self):
+        from google.api_core import retry
+        from google.api_core.retry import if_exception_type
+        from google.cloud.bigtable.table import _BigtableRetryableError
+
+        credentials = _make_credentials()
+        client = self._make_client(
+            project="project-id", credentials=credentials, admin=True
```
sduskis: nit: please don't set `admin=True`.  Ideally, the client / connections should be closed, although I'm not sure we do that consistently.

elisheva-qlogic: I removed admin=True from this test. All other tests in the file set admin=True when creating the client, should I create a separate PR and remove that from all the other tests? 
Regarding closing the client, none of the other tests do that

sduskis: It would be nice to have a separate PR for `admin=True` and closing the clients.

mf2199:
This PR is no longer maintained, please refer to [#8490](https://github.com/googleapis/google-cloud-python/pull/8490) instead.

PR #8091, opened 2019-05-22
Pub/Sub: staticmethod check

anguillanneuf:
The GAPIC wrapper (code snippet below with comments) in [`pubsub/google/cloud/pubsub_v1/_gapic`](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/_gapic.py) falsely recognizes static methods as instance methods. 

```python
    def wrap(wrapped_fx):
        """Wrap a GAPIC method; preserve its name and docstring."""
        # If this is a static or class method, then we need to *not*
        # send self as the first argument.
        #
        # Similarly, for instance methods, we need to send self.api rather
        # than self, since that is where the actual methods were declared.
        instance_method = True

        # If this is a bound method it's a classmethod. 
        self = getattr(wrapped_fx, "__self__", None) # <-- this would return None for staticmethod
        if issubclass(type(self), type):             # <-- this would return False for staticmethod
            instance_method = False                  # <-- resulting in instance_method = True for staticmethod

        # Okay, we have figured out what kind of method this is; send
        # down the correct wrapper function.
        if instance_method:
            fx = lambda self, *a, **kw: wrapped_fx(self.api, *a, **kw)  # noqa
            return functools.wraps(wrapped_fx)(fx)

        fx = lambda *a, **kw: wrapped_fx(*a, **kw)  # noqa
        return staticmethod(functools.wraps(wrapped_fx)(fx))
```

The PR adds a check for **static methods** and aims to make the code more readable.

### DIFF
--- a/pubsub/tests/unit/pubsub_v1/test__gapic.py
+++ b/pubsub/tests/unit/pubsub_v1/test__gapic.py
@@ -36,7 +36,7 @@ class SourceClass(object):
         return "source class blacklisted method"
 
 
-def test__gapic():
+def test_add_method():
     @_gapic.add_methods(SourceClass, ("blacklisted_method",))
     class Foo(object):
         def __init__(self):
@@ -58,5 +58,6 @@ def test__gapic():
     assert foo.class_method() == "source class class method"
 
     # The decorator changes the behavior of instance methods of the wrapped class.
-    # method() is called upon an instance of the Source Class.
+    # method() is called on an instance of the Source Class (stored as an
+    # attribute on the wrapped class).
     assert foo.method() == "source class instance method"
--- a/pubsub/tests/unit/pubsub_v1/test__gapic.py
+++ b/pubsub/tests/unit/pubsub_v1/test__gapic.py
@@ -18,47 +18,45 @@ from google.cloud.pubsub_v1 import _gapic
 
 class SourceClass(object):
     def __init__(self):
-        self.x = 'x'
+        self.x = "x"
 
     def method(self):
-        return 'source class instance method'
+        return "source class instance method"
 
     @staticmethod
     def static_method():
-        return 'source class static method'
+        return "source class static method"
 
     @classmethod
     def class_method(cls):
-        return 'source class class method'
+        return "source class class method"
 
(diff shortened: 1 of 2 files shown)


anguillanneuf:
Let me see if the tests can prove that the code really works. I will add them next. Thanks for the quick turnaround!

pubsub/tests/unit/pubsub_v1/test__gapic.py
```diff
@@ -0,0 +1,62 @@
...
+    assert "blacklisted_method" not in dir(foo)
+
+    # Source Class's static and class methods become static methods.
+    assert type(Foo.__dict__["static_method"]) == staticmethod
+    assert foo.static_method() == "source class static method"
+    assert type(Foo.__dict__["class_method"]) == staticmethod
+    assert foo.class_method() == "source class class method"
+
+    # The decorator changes the behavior of instance methods of the wrapped class.
+    # method() is called upon an instance of the Source Class.
```
plamut: In order to avoid possible misunderstandings, I would just mention that `method()` is actually called on the `self.api` object (which happens to be a `SourceClass` instance in practice).

```suggestion
    # method() is called upon an instance of the Source Class (stored as the `api`
    # attribute on the wrapped class).
```

pubsub/tests/unit/pubsub_v1/test__gapic.py
```diff
@@ -0,0 +1,62 @@
...
+    @classmethod
+    def class_method(cls):
+        return "source class class method"
+
+    @classmethod
+    def blacklisted_method(cls):
+        return "source class blacklisted method"
+
+
+def test__gapic():
```
plamut: Following the pattern in other test files, I would rename this test case to be explicit about what exactly is being tested with it (comes useful when reading test run output on failures). That nicely supplements the module name `test__gapic.py`, which tells where the unit under test is implemented.

```suggestion
def test_add_methods():
```

anguillanneuf:
I don't have permission to merge. If you do, please do! :)

Issue #8093, opened 2019-05-22
BigQuery: Field <field> has changed mode from REQUIRED to NULLABLE


timocb:
I am encountering the following problem, when uploading a Pandas DataFrame to a partitioned table:

#### Environment details
API: BigQuery
OS: macOS High Sierra 10.13.6
Python: 3.5.7
Packages:
```
google-api-core==1.11.0
google-api-python-client==1.7.8
google-auth==1.6.3
google-auth-httplib2==0.0.3
google-cloud==0.34.0
google-cloud-bigquery==1.12.1
google-cloud-core==1.0.0
google-cloud-dataproc==0.3.1
google-cloud-datastore==1.8.0
google-cloud-storage==1.16.0
google-resumable-media==0.3.2
googleapis-common-protos==1.5.10
parquet==1.2
```

#### Steps to reproduce
Create a table on BigQuery with the following fields:
- float_value, FLOAT, required
- int_value, INTEGER, required

#### Reproducible code example (includes creating table)
```python
import pandas as pd
from google.cloud import bigquery


PROJECT = "my-project"
DATASET = "my_dataset"
TABLE = "my_table"


# My table schema
schema = [
    bigquery.SchemaField("foo", "FLOAT", mode="REQUIRED"),
    bigquery.SchemaField("bar", "INTEGER", mode="REQUIRED"),
]


# Set everything up
client = bigquery.Client(PROJECT)
dataset_ref = client.dataset(DATASET)
table_ref = dataset_ref.table(TABLE)


# Delete the table if exists
print("Deleting table if exists...")
client.delete_table(table_ref, not_found_ok=True)


# Create the table
print("Creating table...")
table = bigquery.Table(table_ref, schema=schema)
table.time_partitioning = bigquery.TimePartitioning(
    type_=bigquery.TimePartitioningType.DAY
)
table = client.create_table(table, exists_ok=True)

print("Table schema:")
print(table.schema)

print("Table partitioning:")
print(table.time_partitioning)

# Upload data to partition
table_partition = TABLE + "$20190522"
table_ref = dataset_ref.table(table_partition)

df = pd.DataFrame({"foo": [1, 2, 3], "bar": [2.0, 3.0, 4.0]})
client.load_table_from_dataframe(df, table_ref).result()
```

## Output:
```
Deleting table if exists...
Creating table...
Table schema:
[...]

Issue #7632, opened 2019-04-01
BigQuery: Add enum with allowed data types for use in Schema


tswast:
I received a feature request internally for an enum with allowed data types for use in SchemaField.

The string values should match the [names of the BigQuery legacy SQL data types](https://cloud.google.com/bigquery/data-types#legacy_sql_data_types ), but we could potentially add additional constants using the [standard SQL data type names](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types), but still the legacy SQL string value.

tswast:
We might already have such a list of (legacy) data types somewhere? It'd be nice to have an enum for these, similar to how we have one for standard SQL now.

tswast:
Standard SQL version: [google.cloud.bigquery.enums.StandardSqlDataTypes](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.enums.StandardSqlDataTypes.html)

PR #8105, opened 2019-05-22
Use `job_config.schema` for data type conversion if specified in `load_table_from_dataframe`.

### LINKED ISSUE #7370: BigQuery: Use BigQuery schema (from LoadJobConfig) if available when converting to Parquet in load_table_from_dataframe
#### Environment details

OS = Windows 7
Python = 3.7.2
google-cloud-bigquery 1.8.1

#### Steps to reproduce

1. Try to upload a DataFrame with only pd.np.nan in a column into a table that normally uses STRING

#### Code example

```python
import pandas as pd
from oauth2client.client import GoogleCredentials
from oauth2client.service_account import ServiceAccountCredentials

JSON = # insert json access here
dataset_name = # insert dataset name here
client = bigquery.Client.from_service_account_json(JSON)

test1 = pd.DataFrame({'column_1':[pd.np.nan, pd.np.nan, 'b'], 
					  'column_2':['1', '1', '1']})

test2 = test1.copy()
test2 = test2[:2]

def upload(data, name):
	dataset = client.dataset(dataset_name)
	table = dataset.table(name)
	return client.load_table_from_dataframe(data, table).result()

print('uploading 1: nan with string in column')
result = upload(test1, 'test1')

print('uploading 2: nan without string in column (fails)')
result2 = upload(test2, 'test1')
```

#### Error
```
google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table. Field column_1 has changed type from STRING to INTEGER
```

So I've been in contact with enterprise support, but they were unable to help me. Apparently there's a bug in google.bigquery.Client in the line dataframe.to_parquet(buffer) that causes columns with all NaN values to be interpreted as FLOAT or INTEGER instead of STRING. This prevents the dataframe from being uploaded and there is no other way to introduce NULLs into the table in BigQuery. The issue does not occur in pandas-gbq. Support (ticket 18371705) advised me to use that as a workaround instead until this is fixed and report the issue here. If you have any questions or need more information, feel free to ask.

tswast:
Use the BigQuery schema to inform encoding of file used in load job.
This fixes an issue where a dataframe with ambiguous types (such as an
`object` column containing all `None` values) could not be appended to
an existing table, since the schemas wouldn't match in most cases.

Closes #7370.

### DIFF
--- a/bigquery/google/cloud/bigquery/_pandas_helpers.py
+++ b/bigquery/google/cloud/bigquery/_pandas_helpers.py
@@ -20,6 +20,11 @@ try:
 except ImportError:  # pragma: NO COVER
     pyarrow = None
 
+from google.cloud.bigquery import schema
+
+
+STRUCT_TYPES = ("RECORD", "STRUCT")
+
 
 def pyarrow_datetime():
     return pyarrow.timestamp("us", tz=None)
@@ -37,8 +42,7 @@ def pyarrow_timestamp():
     return pyarrow.timestamp("us", tz="UTC")
 
 
-BQ_TO_ARROW_SCALARS = {}
-if pyarrow is not None:  # pragma: NO COVER
+if pyarrow:  # pragma: NO COVER
     BQ_TO_ARROW_SCALARS = {
         "BOOL": pyarrow.bool_,
         "BOOLEAN": pyarrow.bool_,
@@ -55,6 +59,8 @@ if pyarrow is not None:  # pragma: NO COVER
         "TIME": pyarrow_time,
         "TIMESTAMP": pyarrow_timestamp,
     }
+else:
+    BQ_TO_ARROW_SCALARS = {}
 
 
 def bq_to_arrow_data_type(field):
@@ -62,17 +68,43 @@ def bq_to_arrow_data_type(field):
 
     Returns None if default Arrow type inspection should be used.
     """
-    # TODO: Use pyarrow.list_(item_type) for repeated (array) fields.
     if field.mode is not None and field.mode.upper() == "REPEATED":
+        inner_type = bq_to_arrow_data_type(
+            schema.SchemaField(field.name, field.field_type)
+        )
+        if inner_type:
+            return pyarrow.list_(inner_type)
         return None
-    # TODO: Use pyarrow.struct(fields) for record (struct) fields.
-    if field.field_type.upper() in ("RECORD", "STRUCT"):
-        return None
+
+    if field.field_type.upper() in STRUCT_TYPES:
+        arrow_fields = [bq_to_arrow_field(subfield) for subfield in field.fields]
+        return pyarrow.struct(arrow_fields)
 
     data_type_constructor = BQ_TO_ARROW_SCALARS.get(field.field_type.upper())
     if data_type_constructor is None:
         return None
     return data_type_constructor()
+
+
+def bq_to_arrow_field(bq_field):
+    """Return the Arrow field, corresponding to a given BigQuery column.
+
+    Returns None if the Arrow type cannot be determined.
+    """
+    arrow_type = bq_to_arrow_data_type(bq_field)
+    if arrow_type:
+        is_nullable = bq_field.mode.upper() == "NULLABLE"
+        return pyarrow.field(bq_field.name, arrow_type, nullable=is_nullable)
+    return None
+
+
+def bq_to_arrow_array(series, bq_field):
+    arrow_type = bq_to_arrow_data_type(bq_field)
+    if bq_field.mode.upper() == "REPEATED":
+        return pyarrow.ListArray.from_pandas(series, type=arrow_type)
+    if bq_field.field_type.upper() in STRUCT_TYPES:
+        return pyarrow.StructArray.from_pandas(series, type=arrow_type)
+    return pyarrow.array(series, type=arrow_type)
 
 
 def to_parquet(dataframe, bq_schema, filepath):
@@ -91,22 +123,18 @@ def to_parquet(dataframe, bq_schema, filepath):
             Path to write Parquet file to.
     """
     if pyarrow is None:
-        raise ValueError("pyarrow is required for BigQuery schema conversion")
+        raise ValueError("pyarrow is required for BigQuery schema conversion.")
 
     if len(bq_schema) != len(dataframe.columns):
         raise ValueError(
-            "Number of columns in schema must match number of columns in dataframe"
+            "Number of columns in schema must match number of columns in dataframe."
         )
 
     arrow_arrays = []
-    column_names = []
+    arrow_names = []
     for bq_field in bq_schema:
-        column_names.append(bq_field.name)
-        arrow_arrays.append(
-            pyarrow.array(
-                dataframe[bq_field.name], type=bq_to_arrow_data_type(bq_field)
-            )
-        )
+        arrow_names.append(bq_field.name)
+        arrow_arrays.append(bq_to_arrow_array(dataframe[bq_field.name], bq_field))
 
-    arrow_table = pyarrow.Table.from_arrays(arrow_arrays, names=column_names)
+    arrow_table = pyarrow.Table.from_arrays(arrow_arrays, names=arrow_names)
     pyarrow.parquet.write_table(arrow_table, filepath)
--- a/bigquery/google/cloud/bigquery/_pandas_helpers.py
+++ b/bigquery/google/cloud/bigquery/_pandas_helpers.py
@@ -19,7 +19,6 @@ try:
     import pyarrow.parquet
 except ImportError:  # pragma: NO COVER
     pyarrow = None
-import six.moves
 
 
 def pyarrow_datetime():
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+            Desired BigQuery schema. Number of columns must match number of
+            columns in the DataFrame.
+        filepath (str):
+            Path to write Parquet file to.
+    """
+    if pyarrow is None:
+        raise ValueError("pyarrow is required for BigQuery schema conversion")
+
+    if len(bq_schema) != len(dataframe.columns):
+        raise ValueError(
```
tswast: Note from chat: Maybe we want to allow the bq_schema to be used as an override? Any unmentioned columns get the default pandas type inference.

This is how pandas-gbq works. The schema argument is more used as an override for when a particular column is ambiguous.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1296,7 +1304,10 @@ def load_table_from_dataframe(
         os.close(tmpfd)
 
         try:
-            dataframe.to_parquet(tmppath)
+            if job_config.schema:
```
tswast: Note from chat: if schema isn't populated, we might want to call get_table and use the table's schema if it the table already exists and we're appending to it. (This is what pandas-gbq does)

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: On second thought, let's leave this as-is and fixup later. Filed https://github.com/googleapis/google-cloud-python/issues/8140 as a feature request.

bigquery/google/cloud/bigquery/client.py
tswast: Ditto. Filed https://github.com/googleapis/google-cloud-python/issues/8142. I think this would make a good feature, but shouldn't block this PR.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+        "FLOAT": pyarrow.float64,
+        "FLOAT64": pyarrow.float64,
+        "GEOGRAPHY": pyarrow.string,
+        "INT64": pyarrow.int64,
+        "INTEGER": pyarrow.int64,
+        "NUMERIC": pyarrow_numeric,
+        "STRING": pyarrow.string,
+        "TIME": pyarrow_time,
+        "TIMESTAMP": pyarrow_timestamp,
+    }
```
aryann: Is there a list somewhere that defines BQ types? I wonder if we can add an assertion here that BQ_TO_ARROW_SCALARS.keys() == BQ_TYPES.keys(), so we have a better guarantee that all types are accounted for.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def bq_to_arrow_data_type(field):
+    """Return the Arrow data type, corresponding to a given BigQuery column.
+
+    Returns None if default Arrow type inspection should be used.
+    """
+    # TODO: Use pyarrow.list_(item_type) for repeated (array) fields.
```
aryann: It would be good to include a little more context in the TODO comment as to why we are not adding support for these in this change.

bigquery/tests/system.py
```diff
@@ -622,6 +626,66 @@ def test_load_table_from_local_avro_file_then_dump_table(self):
...
+            bigquery.SchemaField("geo_col", "GEOGRAPHY"),
+            bigquery.SchemaField("int_col", "INTEGER"),
+            bigquery.SchemaField("num_col", "NUMERIC"),
+            bigquery.SchemaField("str_col", "STRING"),
+            bigquery.SchemaField("time_col", "TIME"),
+            bigquery.SchemaField("ts_col", "TIMESTAMP"),
+        )
+        num_rows = 100
+        nulls = [None] * num_rows
+        dataframe = pandas.DataFrame(
```
aryann: Nit: I would suggest putting in non-null values for the sample data to make the test more complete.

bigquery/tests/system.py
```diff
@@ -622,6 +626,66 @@ def test_load_table_from_local_avro_file_then_dump_table(self):
...
+
+        dataset_id = _make_dataset_id("bq_load_test")
+        self.temp_dataset(dataset_id)
+        table_id = "{}.{}.load_table_from_dataframe_w_nulls".format(
+            Config.CLIENT.project, dataset_id
+        )
+        table = retry_403(Config.CLIENT.create_table)(
+            Table(table_id, schema=table_schema)
+        )
+        self.to_delete.insert(0, table)
```
aryann: Is there a reason why we prepend the table ref to to_delete instead of appending it?

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+
+try:
+    import pandas
+except ImportError:  # pragma: NO COVER
+    pandas = None
```
aryann: If you ever get tired of the try/except pattern, you can write a maybe_import(*args) function that returns a tuple of modules:

```
def maybe_import(*args):
  modules = []
  for arg in args:
    try:
      modules.append(__import__(arg))
    except ImportError:
      return tupe([None] * len(args))
  return tuple(modules)
```

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -0,0 +1,162 @@
...
+        ("TIME", "REPEATED", is_none),
+        ("DATETIME", "REPEATED", is_none),
+        ("GEOGRAPHY", "REPEATED", is_none),
+        ("RECORD", "REPEATED", is_none),
+    ],
+)
+@pytest.mark.skipIf(pyarrow is None, "Requires `pyarrow`")
+def test_bq_to_arrow_data_type(module_under_test, bq_type, bq_mode, is_correct_type):
+    field = schema.SchemaField("ignored_name", bq_type, mode=bq_mode)
+    got = module_under_test.bq_to_arrow_data_type(field)
```
aryann: s/got/actual/?

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: Not yet. There's an open FR at https://github.com/googleapis/google-cloud-python/issues/7632 I've been hesitant to add such a list, since it's yet another thing to keep in sync manually, but I agree it'd be useful for cases such as this.

bigquery/tests/system.py
tswast: So that the table gets deleted before the dataset does.

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: There wasn't a good reason before, so I implemented this.

I tried adding it to the system tests, but now I see there are some open issues in pyarrow that are preventing this support. I think REPEATED support may get fixed when https://github.com/googleapis/google-cloud-python/issues/8093 is fixed, since there's a mode mismatch right now (fields are always set to nullable in the parquet file).

Struct support depends on https://jira.apache.org/jira/browse/ARROW-2587. I've filed https://github.com/googleapis/google-cloud-python/issues/8191 to track this as an open issue.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+            columns in the DataFrame.
+        filepath (str):
+            Path to write Parquet file to.
+    """
+    if pyarrow is None:
+        raise ValueError("pyarrow is required for BigQuery schema conversion")
+
+    if len(bq_schema) != len(dataframe.columns):
+        raise ValueError(
+            "Number of columns in schema must match number of columns in dataframe"
```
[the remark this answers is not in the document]
tswast: Done.

tswast:
@aryann Just pushed some commits that get unit test coverage back to 100%. Added system test for non-null scalar values + explicit schema.

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -0,0 +1,458 @@
...
+    pyarrow = None
+import pytest
+
+from google.cloud.bigquery import schema
+
+
+@pytest.fixture
+def module_under_test():
+    from google.cloud.bigquery import _pandas_helpers
+
```
aryann: Nit: rm empty line?

tswast: Thanks. We actually don't have much control over this, since `black` (Python code formatter) will just add it back.

PR #8141, opened 2019-05-24
Trace: fix samples in README.

pj:
Change google-cloud-trace installation instructions to reference correct
package name and update example to reflect the fact that list_traces is
only available in v1.


tseaver:
Superseded by #8833.

PR #8176, opened 2019-05-27
Bigtable: Add 'PartialRowsData.cancel'.

### LINKED ISSUE #7760: Ability to cancel Cloud Bigtable reads
Cloud Bigtable scanning reads can be long.  In many cases, users want to cancel the request before the read stream is complete.  `read_rows` should return a cancelable generator.

This is an important feature for a Cloud Bigtable Apache Beam connector.

mf2199:
Fixes #7760.

Introducing a "stop" flag into the iterable PartialRowsData() class. Setting this flag to True results in cancelling of the iteration the next time the iterator is called.

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -440,7 +440,7 @@ class PartialRowsData(object):
         Parse the response and its chunks into a new/existing row in
         :attr:`_rows`. Rows are returned in order by row key.
         """
-        while True:
+        while not self._stop:
             try:
                 response = self._read_next_response()
             except StopIteration:
@@ -448,17 +448,14 @@ class PartialRowsData(object):
                     raise ValueError("The row remains partial / is not committed.")
                 break
 
-            try:
-                for chunk in response.chunks:
-                    if self._stop:
-                        raise StopIteration
-                    self._process_chunk(chunk)
-                    if chunk.commit_row:
-                        self.last_scanned_row_key = self._previous_row.row_key
-                        self._counter += 1
-                        yield self._previous_row
-            except StopIteration:
-                break
+            for chunk in response.chunks:
+                if self._stop:
+                    break
+                self._process_chunk(chunk)
+                if chunk.commit_row:
+                    self.last_scanned_row_key = self._previous_row.row_key
+                    self._counter += 1
+                    yield self._previous_row
 
             resp_last_key = response.last_scanned_row_key
             if resp_last_key and resp_last_key > self.last_scanned_row_key:
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -377,7 +377,7 @@ class PartialRowsData(object):
         self._state = self.STATE_NEW_ROW
 
         # Flag to stop iteration, for any reason not related to self.retry()
-        self._stop = False
+        self._cancelled = False
 
     @property
     def state(self):
@@ -391,7 +391,7 @@ class PartialRowsData(object):
 
     def cancel(self):
         """Cancels the iterator, closing the stream."""
-        self._stop = True
+        self._cancelled = True
         self.response_iterator.cancel()
 
     def consume_all(self, max_loops=None):
@@ -440,7 +440,7 @@ class PartialRowsData(object):
         Parse the response and its chunks into a new/existing row in
         :attr:`_rows`. Rows are returned in order by row key.
         """
-        while not self._stop:
+        while not self._cancelled:
             try:
                 response = self._read_next_response()
             except StopIteration:
@@ -449,7 +449,7 @@ class PartialRowsData(object):
                 break
 
             for chunk in response.chunks:
-                if self._stop:
+                if self._cancelled:
                     break
                 self._process_chunk(chunk)
                 if chunk.commit_row:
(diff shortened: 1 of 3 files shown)


bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -439,18 +442,19 @@ def __iter__(self):
         while True:
             try:
                 response = self._read_next_response()
+                for chunk in response.chunks:
+                    if self.stop:
+                        raise StopIteration
```
sduskis: The cancellation could happen while a row is in progress, which means that `self.state != self.NEW_ROW`.  I think that we need to to separate out the `for` loop into a separate `try/except` block which doesn't `raise` a `ValueError`.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -376,6 +376,9 @@ def __init__(self, read_method, request, retry=DEFAULT_RETRY_READ_ROWS):
         self.rows = {}
         self._state = self.STATE_NEW_ROW
 
+        # Flag to stop iteration, for any reason not related to self.retry()
+        self.stop = False
```
sduskis: This needs to be set on cancel.  Also, this should be `self._stop`.

sduskis:
@mf2199, I had some comments that need to be addressed regarding this PR.

bigtable/google/cloud/bigtable/row_data.py
mf2199: Yes, good catch, thanks!
Done.

bigtable/google/cloud/bigtable/row_data.py
mf2199: Agreed. Another, shorter way could be setting the `self.state = self.NEW_ROW` just before raising the `StopIteration`. But we'll make another `try/except` block for now --- Done.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -444,12 +448,17 @@ def __iter__(self):
...
+                for chunk in response.chunks:
+                    if self._stop:
+                        raise StopIteration
+                    self._process_chunk(chunk)
+                    if chunk.commit_row:
+                        self.last_scanned_row_key = self._previous_row.row_key
+                        self._counter += 1
+                        yield self._previous_row
+            except StopIteration:
+                break
```
tseaver: ISTM that if `self._stop` is True, then the call to `self._read_next_response()` should be skipped as well as the chunk processing.  E.g:

```python
        while not self._stop:
            try:
                response = self._read_next_response()
            except StopIteration:
                if self.state != self.NEW_ROW:
                    raise ValueError("The row remains partial / is not committed.")
                break

            for chunk in response.chunks:
                if self._stop:
                    break
                self._process_chunk(chunk)
                if chunk.commit_row:
                    self.last_scanned_row_key = self._previous_row.row_key
                    self._counter += 1
                    yield self._previous_row
```

bigtable/tests/unit/test_row_data.py
```diff
@@ -529,6 +529,7 @@ def test_cancel(self):
         self.assertEqual(response_iterator.cancel_calls, 0)
         yield_rows_data.cancel()
         self.assertEqual(response_iterator.cancel_calls, 1)
+        self.assertRaises(StopIteration, iter(yield_rows_data).next())
```
tseaver: We should be testing what we expect callers to want to see.  E.g., once `cancel` is called, iteration stops, yielding an empty list:

```python
        yield_rows_data.cancel()
        self.assertEqual(response_iterator.cancel_calls, 1)
        self.assertEqual(list(yield_rows_data), [])
```

bigtable/google/cloud/bigtable/row_data.py
mf2199: Seems to be a better way, thank you. Done.

bigtable/tests/unit/test_row_data.py
mf2199: Agreed, although one could imagine a scenario when the list can be empty for reasons other than canceling the iteration. Done.

tseaver:
The [`snippets-2.7` failures](https://source.cloud.google.com/results/invocations/55a11b5d-cbed-4845-83fe-794a4ebd934e/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable/log) seem to be related to an outage (?) on the back-end (or maybe the Bigtable API returns 503 responses instead of using 429 for rate limiting?).

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -376,6 +376,9 @@ def __init__(self, read_method, request, retry=DEFAULT_RETRY_READ_ROWS):
         self.rows = {}
         self._state = self.STATE_NEW_ROW
 
+        # Flag to stop iteration, for any reason not related to self.retry()
+        self._stop = False
```
crwilcox: Looking over the code, it seems this is marking if the iterator has been cancelled. The comment and name of this sort of directed me to think maybe this did something different. `self._cancelled` be accurate?

sduskis: +1 for `_cancelled`

mf2199: Makes sense. Done.

tseaver:
- Logging failure would be addressed by #8273.  @busunkim96 Is there a block on fixing / merging that PR?
- Storage systest flakes reported in #8455.

PR #8177, opened 2019-05-27
Storage: add retry to client API methods. 

### LINKED ISSUE #7907: Storage: retry operation
From my experience with other libs, storage operations rarely return HTTP 429 or 5xx retryable errors, e.g. see #3072 or [SO question](https://stackoverflow.com/questions/52304636)

 `api_core` has the `Retry` object, but I cannot find if it's used by the `storage.Client` or not? Should the top level functions have `retry` parameter? Furthemore, Google doc recommends to use (unmaintained!) [retrying lib](https://github.com/rholder/retrying), see https://cloud.google.com/storage/docs/exponential-backoff

It is connected with old storage client design #7762 ?

HemangChothani:
Fixes #7907

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -33,6 +33,7 @@ import mimetypes
 import os
 import time
 import warnings
+import functools
 
 from six.moves.urllib.parse import parse_qsl
 from six.moves.urllib.parse import quote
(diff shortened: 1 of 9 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -625,6 +628,9 @@ def download_to_file(self, file_obj, client=None, start=None, end=None):
         :type end: int
         :param end: Optional, The last byte in a range to be downloaded.
 
+        :type retry: :class:`google.api_core.retry.Retry`
+        :param retry: (Optional) How to retry the RPC.
```
crwilcox: This client is REST, not RPC. Could you update the comments to reflect that?

crwilcox: Maybe "API Call"?

tseaver:
The overall retry strategy for manual clients is a large effort, tracked in #9298.

PR #8178, opened 2019-05-27
Datastore:  expose 'reserveIds' API.

HemangChothani:
Fixes: #7111

### DIFF
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -650,8 +650,7 @@ class Client(ClientWithProject):
         return Query(self, **kwargs)
 
     def reserve_ids(self, complete_key, num_ids):
-        """Prevents the supplied keys' IDs from being auto-allocated by Cloud
-            Datastore.
+        """Reserve a list of IDs from a complete key.
 
         :type complete_key: :class:`google.cloud.datastore.key.Key`
         :param complete_key: Partial key to use as base for reserved IDs.
@@ -659,8 +658,8 @@ class Client(ClientWithProject):
         :type num_ids: int
         :param num_ids: The number of IDs to reserve.
 
-        :rtype: list of :class:`google.cloud.datastore.key.Key`
-        :returns: []
+        :rtype: class:`NoneType`
+        :returns: None
         :raises: :class:`ValueError` if `complete_key`` is not a
                  Complete key.
         """
@@ -673,8 +672,6 @@ class Client(ClientWithProject):
         complete_key_pb = complete_key.to_protobuf()
         complete_key_pbs = [complete_key_pb] * num_ids
 
-        response_pb = self._datastore_api.reserve_ids(
-            complete_key.project, complete_key_pbs
-        )
+        self._datastore_api.reserve_ids(complete_key.project, complete_key_pbs)
 
-        return response_pb
+        return None
--- a/datastore/google/cloud/datastore/client.py
+++ b/datastore/google/cloud/datastore/client.py
@@ -654,18 +654,21 @@ class Client(ClientWithProject):
             Datastore.
 
         :type complete_key: :class:`google.cloud.datastore.key.Key`
-        :param complete_key: Partial key to use as base for allocated IDs.
+        :param complete_key: Partial key to use as base for reserved IDs.
 
         :type num_ids: int
         :param num_ids: The number of IDs to reserve.
 
         :rtype: list of :class:`google.cloud.datastore.key.Key`
         :returns: []
-        :raises: :class:`ValueError` if ``incomplete_key`` is not a
-                 partial key.
+        :raises: :class:`ValueError` if `complete_key`` is not a
+                 Complete key.
         """
         if complete_key.is_partial:
             raise ValueError(("Key is not Complete.", complete_key))
+
+        if not isinstance(num_ids, int):
+            raise ValueError(("num_ids is not a valid integer.", num_ids))
 
         complete_key_pb = complete_key.to_protobuf()
         complete_key_pbs = [complete_key_pb] * num_ids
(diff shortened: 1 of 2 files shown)


datastore/google/cloud/datastore/client.py
```diff
@@ -648,3 +648,33 @@ def do_something(entity):
         if "namespace" not in kwargs:
             kwargs["namespace"] = self.namespace
         return Query(self, **kwargs)
+
+    def reserve_ids(self, complete_key, num_ids):
+        """Prevents the supplied keys' IDs from being auto-allocated by Cloud
```
tseaver: Please use "imperative voice" for the docstring summary, and keep it to one line, e.g.:

```python
    def reserve_ids(self, complete_key, num_ids):
        """Prevent the supplied keys' IDs from being auto-allocated.

        ...
```

datastore/google/cloud/datastore/client.py
```diff
@@ -648,3 +648,33 @@ def do_something(entity):
...
+            Datastore.
+
+        :type complete_key: :class:`google.cloud.datastore.key.Key`
+        :param complete_key: Partial key to use as base for reserved IDs.
+
+        :type num_ids: int
+        :param num_ids: The number of IDs to reserve.
+
+        :rtype: list of :class:`google.cloud.datastore.key.Key`
+        :returns: []
```
tseaver: `:rtype:` and `:returns:` need to match the actual returned value below.  Because the `ReserveIdsResponse` message is empty, we should just be returning `None`.

datastore/google/cloud/datastore/client.py
```diff
@@ -648,3 +648,33 @@ def do_something(entity):
...
+        if complete_key.is_partial:
+            raise ValueError(("Key is not Complete.", complete_key))
+
+        if not isinstance(num_ids, int):
+            raise ValueError(("num_ids is not a valid integer.", num_ids))
+
+        complete_key_pb = complete_key.to_protobuf()
+        complete_key_pbs = [complete_key_pb] * num_ids
+
+        response_pb = self._datastore_api.reserve_ids(
```
tseaver: We don't need the `response_pb` here:  it contains no information.  We should just return `None`, e.g.:

```python
        self._datastore_api.reserve_ids(complete_key.project, complete_key_pbs)
        return None
```

datastore/tests/unit/test_client.py
```diff
@@ -860,13 +860,43 @@ def test_allocate_ids_w_partial_key(self):
...
 
         COMPLETE_KEY = _Key(self.PROJECT)
         self.assertRaises(ValueError, client.allocate_ids, COMPLETE_KEY, 2)
 
+    def test_reserve_ids_w_completed_key(self):
+        num_ids = 2
+        creds = _make_credentials()
+        client = self._make_one(credentials=creds,_use_grpc=False)
+        COMPLETE_KEY = _Key(self.PROJECT)
+        alloc_ids = mock.Mock(return_value=None, spec=[])
```
tseaver: Returning `None` here is incorrect:  the mock should return another object to emulate the `ReserveIdsResponse` message that the API actually returns.

tseaver:
Also, per line #117 of the [Kokoro job for Datastore](https://source.cloud.google.com/results/invocations/5547d3e4-ef9b-4f22-b757-5a66377e192f/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fdatastore/log), please run `nox -re black` to appease the linter.

datastore/tests/unit/test_client.py
HemangChothani: This test case is regarding to the datastore.client class  , we expose method reserve_ids for that and it returns None as you commented before, datastore_v1.DatastoreClient's method reserve_ids returns the object ReserveIdsResponse,  So not getting it why to mock ReserveIdsResponse message when it is empty  and how i compare with the result with return by datastore.client method ,so may i remove return_value=None from mock and leave blank like reserve_ids = mock.Mock(spec=[])

tseaver: Yes, you can just let the mock have its default return value.  The test would then fail (because the caller is returning `response_pb`), which we then fix by explicitly returning `None`.

tseaver:
@HemangChothani Please don't mark a reviewer's requested changes as "resolved" until you've pushed the commit which adds the change.  If you address the review comment by commenting back, let the reviewer mark it as resolved if she / he agrees with your reply.

datastore/tests/unit/test_client.py
```diff
@@ -864,38 +864,39 @@ def test_allocate_ids_w_completed_key(self):
...
         client._datastore_api_internal = ds_api
-        self.assertTrue(not COMPLETE_KEY.is_partial)
-        self.assertIsInstance(num_ids, int)
-        result = client.reserve_ids(COMPLETE_KEY, num_ids)
-
-        self.assertIsNone(result)
+        self.assertTrue(not complete_key.is_partial)
+        client.reserve_ids(complete_key, num_ids)
+        expected_keys = [complete_key.to_protobuf()] * num_ids
+        expected_keys = [complete_key.to_protobuf()] * num_ids
```
tseaver: Duplicate line.

HemangChothani: @tseaver that duplicate line is already removed in last commit "reserver_id mock return default value" id is 79e5a81.

tseaver:
The [BigQuery Kokoro failure](https://source.cloud.google.com/results/invocations/c4beb8ef-40e8-4596-bc3b-09252b5ce04a/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigquery) is unrelated (see #8212).

PR #8179, opened 2019-05-27
BigQuery: Fix Add maximum_bytes_billed option to magics

### LINKED ISSUE #7678: BigQuery: Add maximum_bytes_billed option to magics
**Background**

BigQuery queries can get expensive, but the `maximum_bytes_billed` query option adds a cap to the price billed and rejects queries that could be too expensive. Provide a default value for `maximum_bytes_billed` when constructing a client by specifying a `default_query_job_config`, but this is not possible when using the `%%bigquery` magics.

**Feature Request**

Add a `--maximum_bytes_billed` option to `%%bigquery`. Possible values:

* `UNSPECIFIED` - Use the default value . Set `job_config.maximum_bytes_billed = google.cloud.bigquery.magics.context.maximum_bytes_billed`.
* `--maximum_bytes_billed=None` - No maximum value. Set `job_config.maximum_bytes_billed = None`. This allows all queries to run, overriding any value from the global context.
* `--maximum_bytes_billed=123456789` - Set maximum to value specified. Set `job_config.maximum_bytes_billed = int(value)`.
* `--maximum_bytes_billed=NotAnInteger` - Raise `ValueError`.

HemangChothani:
Fixes  #7678

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -161,7 +161,7 @@ class Context(object):
         self._project = None
         self._connection = None
         self._use_bqstorage_api = None
-        self._maximum_bytes_billed = None
+        self._maximum_bytes_billed = 0
 
     @property
     def credentials(self):
@@ -261,7 +261,7 @@ class Context(object):
             job_config = bigquery.job.QueryJobConfig()
             job_config.maximum_bytes_billed = self._maximum_bytes_billed
 
-        except:
+        except Exception:
             raise ValueError("value is not a valid integer.")
 
 
@@ -317,6 +317,13 @@ def _run_query(client, query, job_config=None):
     type=str,
     default=None,
     help=("Project to use for executing this query. Defaults to the context project."),
+)
+@magic_arguments.argument(
+    "--maximum_bytes_billed",
+    default=None,
+    help=(
+        "maximum_bytes_billed to use for executing this query. Defaults to the context maximum_bytes_billed."
+    ),
 )
 @magic_arguments.argument(
     "--use_legacy_sql",
@@ -399,7 +406,17 @@ def _cell_magic(line, query):
     job_config = bigquery.job.QueryJobConfig()
     job_config.query_parameters = params
     job_config.use_legacy_sql = args.use_legacy_sql
-    job_config.maximum_bytes_billed = context.maximum_bytes_billed
+
+    if args.maximum_bytes_billed == "None":
+        job_config.maximum_bytes_billed = 0
+    elif args.maximum_bytes_billed is not None:
+        try:
+            value = int(args.maximum_bytes_billed)
+            job_config.maximum_bytes_billed = value
+        except Exception:
+            raise ValueError("value is not a valid integer.")
+    else:
+        job_config.maximum_bytes_billed = context.maximum_bytes_billed
     query_job = _run_query(client, query, job_config)
 
     if not args.verbose:
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -161,6 +161,7 @@ class Context(object):
         self._project = None
         self._connection = None
         self._use_bqstorage_api = None
+        self._maximum_bytes_billed = None
 
     @property
     def credentials(self):
@@ -236,6 +237,32 @@ class Context(object):
     @use_bqstorage_api.setter
     def use_bqstorage_api(self, value):
         self._use_bqstorage_api = value
+
+    @property
+    def maximum_bytes_billed(self):
+        """int: Maximum bytes to be billed for this job or :data:`None` if not set.
+
+        Example:
+            Manually setting the context maximum_bytes_billed:
+
+            >>> from google.cloud.bigquery import magics
+            >>> magics.context.maximum_bytes_billed = '123'
+
+        Raises:
+            ValueError: If the parameters are invalid.
+        """
+        return self._maximum_bytes_billed
+
+    @maximum_bytes_billed.setter
+    def maximum_bytes_billed(self, value):
+        try:
+            value = int(value)
+            self._maximum_bytes_billed = value
+            job_config = bigquery.job.QueryJobConfig()
+            job_config.maximum_bytes_billed = self._maximum_bytes_billed
+
+        except:
+            raise ValueError("value is not a valid integer.")
 
 
 context = Context()
@@ -372,6 +399,7 @@ def _cell_magic(line, query):
     job_config = bigquery.job.QueryJobConfig()
     job_config.query_parameters = params
     job_config.use_legacy_sql = args.use_legacy_sql
+    job_config.maximum_bytes_billed = context.maximum_bytes_billed
     query_job = _run_query(client, query, job_config)
 
     if not args.verbose:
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/magics.py
```diff
@@ -236,6 +237,32 @@ def use_bqstorage_api(self):
...
+        Example:
+            Manually setting the context maximum_bytes_billed:
+
+            >>> from google.cloud.bigquery import magics
+            >>> magics.context.maximum_bytes_billed = '123'
+
+        Raises:
+            ValueError: If the parameters are invalid.
+        """
+        return self._maximum_bytes_billed
```
sduskis: Shouldn't this be `return job_config.maximum_bytes_billed`?

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -236,6 +237,32 @@ def use_bqstorage_api(self):
...
+
+    @maximum_bytes_billed.setter
+    def maximum_bytes_billed(self, value):
+        try:
+            value = int(value)
+            self._maximum_bytes_billed = value
```
sduskis: Why do we need `self._maximum_bytes_billed`?  Can't we just use `job_config.maximum_bytes_billed`?

bigquery/tests/unit/test_magics.py
```diff
@@ -439,6 +439,88 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
+
+    sql = "SELECT 17 AS num"
+    result = pandas.DataFrame([17], columns=["num"])
+    query_job_mock = mock.create_autospec(
+        google.cloud.bigquery.job.QueryJob, instance=True
+    )
+    query_job_mock.to_dataframe.return_value = result
+    with run_query_patch as run_query_mock, default_patch:
+        run_query_mock.return_value = query_job_mock
+        return_value = ip.run_cell_magic("bigquery", "--maximum_bytes_billed=123456789", sql)
```
sduskis: Can you please check to make sure that the `maximum_bytes_billed` is used?

bigquery/tests/unit/test_magics.py
```diff
@@ -439,6 +439,88 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
+    query_job_mock = mock.create_autospec(
+        google.cloud.bigquery.job.QueryJob, instance=True
+    )
+    query_job_mock.to_dataframe.return_value = result
+    with run_query_patch as run_query_mock, default_patch:
+        run_query_mock.return_value = query_job_mock
+        return_value = ip.run_cell_magic("bigquery", "--maximum_bytes_billed=None", sql)
+
+        bqstorage_mock.assert_not_called()
+        query_job_mock.to_dataframe.assert_called_once_with(bqstorage_client=None)
```
sduskis: Can you please check to make sure that the `maximum_bytes_billed` is correct?

bigquery/google/cloud/bigquery/magics.py
tswast: Why is `job_config` even in this function?

One way of implementing this feature would be to create a `self._job_config` to hold the default job config (used when constructing the BigQuery client), but that doesn't appear to be what's going on here.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -236,6 +237,32 @@ def use_bqstorage_api(self):
...
+            job_config = bigquery.job.QueryJobConfig()
+            job_config.maximum_bytes_billed = self._maximum_bytes_billed
+
+        except Exception:
```
tswast: This catch block is not needed. Just let `int()` throw.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -369,6 +403,17 @@ def _cell_magic(line, query):
...
     job_config.query_parameters = params
     job_config.use_legacy_sql = args.use_legacy_sql
+
+    if args.maximum_bytes_billed == "None":
+        job_config.maximum_bytes_billed = 0
+    elif args.maximum_bytes_billed is not None:
+        try:
+            value = int(args.maximum_bytes_billed)
+            job_config.maximum_bytes_billed = value
+        except Exception:
```
tswast: Ditto. Don't catch `Exception`. We do need to think about having better error messages in our magics, but I don't think this is the way to do it.

bigquery/google/cloud/bigquery/magics.py
tswast: I think this part is OK. A potential alternative implementation is to have a `self._default_job_config` object, but I think that can wait until we have more than one default property to set.

alixhami: I think we should just have a default job config that users can set properties on. That way, users can set properties on the default config that we don't have magics arguments for. No need to make properties for each of the job config properties. That approach doesn't scale well, because it would be too much repetitive code to add each job config property separately.

HemangChothani: I created a property for the class Context  because i found  Set job_config.maximum_bytes_billed = google.cloud.bigquery.magics.context.maximum_bytes_billed line in the description of feature request,So can i remove that property from Context class which is in magic.py file

tswast: Sorry for the "flip flopping" @HemangChothani. Alix convinced me offline that a `default_query_job_config` property is the better implementation.

I've added commit 30f8326092e5bc16d1c4f3e9b2b1f36beedec9f8 that implements this.

bigquery/tests/unit/test_magics.py
```diff
@@ -548,6 +552,168 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
+    )
+    query_job_mock.to_dataframe.return_value = result
+    with run_query_patch as run_query_mock, default_patch:
+        run_query_mock.return_value = query_job_mock
+        return_value = ip.run_cell_magic(
+            "bigquery", "--maximum_bytes_billed=123456789", sql
+        )
+
+        bqstorage_mock.assert_not_called()
+        query_job_mock.to_dataframe.assert_called_once_with(bqstorage_client=None)
```
alixhami: Shouldn't there also be an assert that the maximum bytes billed was actually used?

bigquery/tests/unit/test_magics.py
```diff
@@ -548,6 +552,168 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
+    assert sent_config["maximumBytesBilled"] == expected
+
+
+@pytest.mark.usefixtures("ipython_interactive")
+def test_bigquery_magic_w_maximum_bytes_billed_w_context_inplace():
+    ip = IPython.get_ipython()
+    ip.extension_manager.load_extension("google.cloud.bigquery")
+    magics.context._project = None
+
+    magics.context.default_query_job_config.maximum_bytes_billed = 1337
```
alixhami: lol nice

bigquery/tests/unit/test_magics.py
tswast: Good catch. I just realized this test is redundant with `test_bigquery_magic_w_maximum_bytes_billed_overrides_context`. I've removed it.

tswast:
Coverage failure is a weird one. It's not happening locally.

```
tests/unit/test_magics.py                     410      0      4      1    99%   246->exit
```

That line is:

```
assert all(re.match("Query executing: .*s", line) for line in execution_updates)
```

Maybe it's getting confused by the generator expression? I can try converting that to a plain old for loop.

tswast:
Thank you @HemangChothani for the contribution. This feature will appear in the next release of `google-cloud-bigquery`.

PR #8186, opened 2019-05-28
Generate grafeas and container analysis.

busunkim96:
This is the generation for Grafeas + Container Analysis.

We are creating separate clients (Grafeas is the open source version, Container Analysis is Google specific). Container Analysis will have a dependency GrafeasClient, and there will be an easy way of getting a GrafeasClient instance from a ContainerAnalysisClient.

I would appreciate it if you could particularly look at [this diff](https://github.com/googleapis/google-cloud-python/pull/8186/commits/a771c47aa14b9b6a8169731395e36a6afdc979a1) where the following things were removed:
* default API Endpoint
* default credentials (require a transport to initialize the client)
* default oauth scopes. 

----

[prototype in C#](https://github.com/googleapis/google-cloud-dotnet/pull/3085)

- [x] Generate the regular GAPIC library for Grafeas
- [x] Remove any sign of a default endpoint or use of default application credentials
- [x] Generate a GAPIC library for Container Analysis, that only contains the IAM methods already present in the service

Will do these two things once an initial version of ContainerAnalysis and Grafeas have been published. (not in this PR) :) 
- Make the ContainerAnalysis library depend on the Grafeas library
- Add a property to ContainerAnalysisClient to get a GrafeasClient with the same gRPC channel (so endpoint and credentials)


containeranalysis/docs/conf.py
```diff
@@ -0,0 +1,366 @@
...
+
+# The encoding of source files.
+# source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = "index"
+
+# General information about the project.
+project = u"google-cloud-containeranalysis"
+copyright = u"2017, Google"
```
crwilcox: may be out of scope for this generation, but should this be 2019 seeing as these are new apis?

PR #8193, opened 2019-05-30
Core: Mitigate busy reopen loop in ResumableBidiRpc consuming 100% CPU

### LINKED ISSUE #7910: Pubsub: Pull Subscriber unable to re-connect after a while
#### Environment details

1. API: PubSub (maybe gRPC)
2. OS type and version: Debian GNU/Linux 9 (stretch) container (based on python:2.7.15-slim-stretch) running in 1.12.6-gke.10
3. Python version: 2.7.15
4. google-cloud-pubsub version:  0.40.0

#### Steps to reproduce

  1. Have several PubSub subscribers run for a while (from hours to days) in a Kubernetes container, reconnecting every 10 minutes as normal
  2. One of the subscribers suddenly start seeing reconnection rate go up as much as CPU resources allow (trying every few milliseconds in our case) and repeatedly getting 503 errors
  3. At this point subscriber stops pulling messages and the process needs to be restarted

The issue did happen in a different service with only one subscriber in the container, but having several subscribers with all the others working excludes the possibility of a lot of other factors preventing re-connection (ie DNS resolution, no network, etc etc).

#### Code example

Totally standard Pull subscription using `SubscriberClient` + `create_subscription` + `subscribe`. Can paste code if required though.

#### StackDriver log snippet
```
...
I  2019-05-09 10:51:06.131 [INFO] Re-established stream
I  2019-05-09 10:51:06.134 [DEBUG] Call to retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7f2fde071110>> caused 503 Connect Failed.
I  2019-05-09 10:51:06.135 [INFO] Observed recoverable stream error 503 Connect Failed 
I  2019-05-09 10:51:06.135 [DEBUG] Re-opening stream from retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7f2fde071110>>.
I  2019-05-09 10:51:06.137 [DEBUG] Stream was already re-established.
I  2019-05-09 10:51:06.138 [INFO] Re-established stream
I  2019-05-09 10:51:06.141 [INFO] Observed recoverable stream error 503 Connect Failed
I  2019-05-09 10:51:06.142 [DEBUG] Re-opening stream from gRPC callback.
I  2019-05-09 10:51:06.143 [INFO] Re-established stream 
I  2019-05-09 10:51:06.145 [DEBUG] The current p99 value is 10 seconds.
I  2019-05-09 10:51:06.146 [DEBUG] Snoozing lease management for 7.560536 seconds.
I  2019-05-09 10:51:06.147 [INFO] Observed recoverable stream error 503 Connect Failed
I  2019-05-09 10:51:06.147 [INFO] Re-established stream
...
```

First of all of course I'd be interested to help getting to the bottom of the issue and get it resolved.

But in the meanwhile it would be great to have a workaround detecting lost connection with a subscriber. I went through the public API documentation and couldn't find a way to get to the underlying (gRPC?) client, but it would be great to have a clean(ish) method doing a periodic check on the connection to be able to restart once the issue happens.

Thanks a lot in advance 🙏

plamut:
Closes #7910.

This PR fixes the issue with `ResumableBidiRpc` that can enter a busy re-open loop consuming lots of CPU in the process . The [comment](https://github.com/googleapis/google-cloud-python/issues/7910#issuecomment-494987224) on the issue explains this in more detail.

### How to test
**Steps to reproduce:**
- (optional) Configure more verbose logging in the subscriber script:
```py
import logging 
log_format = (
)
logging.basicConfig(
    level=logging.DEBUG,
    format=log_format,
)
```
- Disable your internet connection
- Start the streaming pull using a susbscriber client instance:
```py
subscriber.subscribe(SUBSCRIPTION_PATH, callback=my_callback)

while True:
    try:
        time.sleep(60)
    except KeyboardInterrupt:
        break
 ```

The `ResumableBidiRpc` class tries to re-establish the stream many times in rapid succession, resulting in a 100% CPU spike and a ton of log output:
```
...
DEBUG    [2019-05-15 10:06:01,440] Thread-ConsumeBidirectionalStream [_recoverable] Call to retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7fb5eb91d630>> caused 503 DNS resolution failed.
INFO     [2019-05-15 10:06:01,440] Thread-ConsumeBidirectionalStream [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,440] Thread-ConsumeBidirectionalStream [_recoverable] Re-opening stream from retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7fb5eb91d630>>.
INFO     [2019-05-15 10:06:01,441] Thread-ConsumeBidirectionalStream [_reopen] Re-established stream
INFO     [2019-05-15 10:06:01,441] Thread-1                          [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,441] Thread-1                          [_on_call_done] Re-opening stream from gRPC callback.
DEBUG    [2019-05-15 10:06:01,441] Thread-1                          [_reopen] Stream was already re-established.
INFO     [2019-05-15 10:06:01,441] Thread-1                          [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,441] Thread-1                          [_on_call_done] Re-opening stream from gRPC callback.
INFO     [2019-05-15 10:06:01,441] Thread-1                          [_reopen] Re-established stream
DEBUG    [2019-05-15 10:06:01,441] Thread-ConsumeBidirectionalStream [_recoverable] Call to retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7fb5eb91d630>> caused 503 DNS resolution failed.
INFO     [2019-05-15 10:06:01,442] Thread-ConsumeBidirectionalStream [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,442] Thread-ConsumeBidirectionalStream [_recoverable] Re-opening stream from retryable <bound method ResumableBidiRpc._recv of <google.api_core.bidi.ResumableBidiRpc object at 0x7fb5eb91d630>>.
INFO     [2019-05-15 10:06:01,442] Thread-ConsumeBidirectionalStream [_reopen] Re-established stream
INFO     [2019-05-15 10:06:01,442] Thread-1                          [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,442] Thread-1                          [_on_call_done] Re-opening stream from gRPC callback.
DEBUG    [2019-05-15 10:06:01,442] Thread-1                          [_reopen] Stream was already re-established.
INFO     [2019-05-15 10:06:01,442] Thread-1                          [_should_recover] Observed recoverable stream error 503 DNS resolution failed
DEBUG    [2019-05-15 10:06:01,442] Thread-1                          [_on_call_done] Re-opening stream from gRPC callback.
INFO     [2019-05-15 10:06:01,443] Thread-1                          [_reopen] Re-established stream
...
```

This also happens if the streaming pull is running normally, and the internet connection is shut down in the middle of it - eventually the busy re-open loop starts. What's worse, this happens more than a single thread - both the gRPC channel thread (?)  and the consumer helper thread try to re-establish the stream.

The re-open calls are throttled, and the CPU consumption is "normal".

### Things to do/discuss:
- [x] Fix test coverage (one branch appears to be missing).
- [ ] What should the default throttle params be set to? 
- [x] `ResumalbeBidiRpc` is also used in the Firestore client, does this PR (negatively) affect it in any way?
ANSWER: No, because this PR does not change the former's default behavior.
- [ ] Tests rely on timing, and even with a sizeable safety buffer, this is not ideal. Would there be any objections to adding [freezegun](https://pypi.org/project/freezegun/) as an API core dependency?

### DIFF
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -14,8 +14,11 @@
 
 """Bi-directional streaming RPC helpers."""
 
+import collections
+import datetime
 import logging
 import threading
+import time
 
 from six.moves import queue
 
@@ -132,6 +135,69 @@ class _RequestQueueGenerator(object):
                 return
 
             yield item
+
+
+class _Throttle(object):
+    """A context manager limiting the total entries in a sliding time window.
+
+    If more than ``entry_cap`` attempts are made to enter the context manager
+    instance in the last ``time window`` seconds, the exceeding requests block
+    until enough time elapses.
+
+    The context manager instances are thread-safe and can be shared between
+    multiple threads. If multiple requests are blocked and waiting to enter,
+    the exact order in which they are allowed to proceed is not determined.
+
+    Example::
+
+        max_three_per_second = Throttle(time_window=1, entry_cap=3)
+
+        for i in range(5):
+            with max_three_per_second as time_waited:
+                print("{}: Waited {} seconds to enter".format(i, time_waited))
+
+    Args:
+        time_window (float): the width of the sliding time window in seconds
+        entry_cap (int): the maximum number of entries allowed in the time window
+    """
+
+    def __init__(self, time_window, entry_cap):
+        if time_window <= 0.0:
+            raise ValueError("time_window argument must be positive")
+
+        if entry_cap < 1:
+            raise ValueError("entry_cap argument must be positive")
+
+        self._time_window = datetime.timedelta(seconds=time_window)
+        self._entry_cap = entry_cap
+        self._past_entries = collections.deque(maxlen=entry_cap)  # least recent first
+        self._entry_lock = threading.Lock()
+
+    def __enter__(self):
+        with self._entry_lock:
+            cutoff_time = datetime.datetime.now() - self._time_window
+
+            # drop the entries that are too old, as they are no longer relevant
+            while self._past_entries and self._past_entries[0] < cutoff_time:
+                self._past_entries.popleft()
+
+            if len(self._past_entries) < self._entry_cap:
+                self._past_entries.append(datetime.datetime.now())
+                return 0.0  # no waiting was needed
+
+            to_wait = (self._past_entries[0] - cutoff_time).total_seconds()
+            time.sleep(to_wait)
+
+            self._past_entries.append(datetime.datetime.now())
+            return to_wait
+
+    def __exit__(self, *_):
+        pass
+
+    def __repr__(self):
+        return "{}(time_window={}, entry_cap={})".format(
+            self.__class__.__name__, self._time_window.total_seconds(), self._entry_cap
+        )
 
 
 class BidiRpc(object):
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -140,8 +140,8 @@ class _RequestQueueGenerator(object):
 class _Throttle(object):
     """A context manager limiting the total entries in a sliding time window.
 
-    If more than ``entry_cap`` attempts are made to enter the context manager
-    instance in the last ``time window`` seconds, the exceeding requests block
+    If more than ``access_limit`` attempts are made to enter the context manager
+    instance in the last ``time window`` interval, the exceeding requests block
     until enough time elapses.
 
     The context manager instances are thread-safe and can be shared between
@@ -150,27 +150,29 @@ class _Throttle(object):
 
     Example::
 
-        max_three_per_second = Throttle(time_window=1, entry_cap=3)
+        max_three_per_second = _Throttle(
+            access_limit=3, time_window=datetime.timedelta(seconds=1)
+        )
 
         for i in range(5):
             with max_three_per_second as time_waited:
                 print("{}: Waited {} seconds to enter".format(i, time_waited))
 
     Args:
-        time_window (float): the width of the sliding time window in seconds
-        entry_cap (int): the maximum number of entries allowed in the time window
+        access_limit (int): the maximum number of entries allowed in the time window
+        time_window (datetime.timedelta): the width of the sliding time window
     """
 
-    def __init__(self, time_window, entry_cap):
-        if time_window <= 0.0:
-            raise ValueError("time_window argument must be positive")
+    def __init__(self, access_limit, time_window):
+        if access_limit < 1:
+            raise ValueError("access_limit argument must be positive")
 
-        if entry_cap < 1:
-            raise ValueError("entry_cap argument must be positive")
+        if time_window <= datetime.timedelta(0):
+            raise ValueError("time_window argument must be a positive timedelta")
 
-        self._time_window = datetime.timedelta(seconds=time_window)
-        self._entry_cap = entry_cap
-        self._past_entries = collections.deque(maxlen=entry_cap)  # least recent first
+        self._time_window = time_window
+        self._access_limit = access_limit
+        self._past_entries = collections.deque(maxlen=access_limit)  # least recent first
         self._entry_lock = threading.Lock()
 
     def __enter__(self):
@@ -181,7 +183,7 @@ class _Throttle(object):
             while self._past_entries and self._past_entries[0] < cutoff_time:
                 self._past_entries.popleft()
 
-            if len(self._past_entries) < self._entry_cap:
+            if len(self._past_entries) < self._access_limit:
                 self._past_entries.append(datetime.datetime.now())
                 return 0.0  # no waiting was needed
 
@@ -195,8 +197,10 @@ class _Throttle(object):
         pass
 
     def __repr__(self):
-        return "{}(time_window={}, entry_cap={})".format(
-            self.__class__.__name__, self._time_window.total_seconds(), self._entry_cap
+        return "{}(access_limit={}, time_window={})".format(
+            self.__class__.__name__,
+            self._access_limit,
+            repr(self._time_window),
         )
 
 
@@ -408,7 +412,9 @@ class ResumableBidiRpc(BidiRpc):
         self._finalize_lock = threading.Lock()
 
         if throttle_reopen:
-            self._reopen_throttle = _Throttle(entry_cap=5, time_window=10)
+            self._reopen_throttle = _Throttle(
+                access_limit=5, time_window=datetime.timedelta(seconds=10),
+            )
         else:
             self._reopen_throttle = None
 
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -389,14 +389,28 @@ class ResumableBidiRpc(BidiRpc):
             whenever an error is encountered on the stream.
         metadata Sequence[Tuple(str, str)]: RPC metadata to include in
             the request.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by busunkim96** [comment]
@crwilcox Could you comment on the impact to Firestore?

**Comment by plamut** [comment]
FWIW, I modified the `ResumalbeBidiRpc` class to **not** use throttling by default, meaning that Firestore will not be affected by this PR.

**Comment by sduskis** [comment]
@crwilcox, what's the good word on this PR?

**Comment by crwilcox** [comment]
@sduskis I had left comments in the python chat for Peter last week. Let me copy them here.

File: api_core/google/api_core/bidi.py [hunk 1]
```diff
@@ -134,6 +137,69 @@ def __iter__(self):
...
+        if entry_cap < 1:
+            raise ValueError("entry_cap argument must be positive")
+
+        self._time_window = datetime.timedelta(seconds=time_window)
+        self._entry_cap = entry_cap
+        self._past_entries = collections.deque(maxlen=entry_cap)  # least recent first
+        self._entry_lock = threading.Lock()
+
+    def __enter__(self):
+        with self._entry_lock:
```
crwilcox: There is a lock, but deque should be threadsafe. Was this done to keep things more straightforward? There is reference to the first element and then removing it in a check, which would have to be changed if we moved to a non-locking impl I think?

api_core/google/api_core/bidi.py
```diff
@@ -134,6 +137,69 @@ def __iter__(self):
...
+        for i in range(5):
+            with max_three_per_second as time_waited:
+                print("{}: Waited {} seconds to enter".format(i, time_waited))
+
+    Args:
+        time_window (float): the width of the sliding time window in seconds
+        entry_cap (int): the maximum number of entries allowed in the time window
+    """
+
+    def __init__(self, time_window, entry_cap):
```
crwilcox: I am not crazy about the signature. I tried to brainstorm some alternatives. 

Currently: `max_three_per_second = Throttle(time_window=1, entry_cap=3)`

Proposal 1: `max_three_per_second = Throttle(accesses_per_window=3, time_window=1)`
Proposal 2: `max_three_per_second = Throttle(access_limit=3, time_window=1)`

Another thought is does it make sense to have `time_window` take a time delta and not a float representing seconds?

plamut: I'm indifferent here, whatever looks more readable and easier to use. I personally find the second proposal slightly favorable (shorter).

I'm also fine with the timedelta, more self-explanatory, will change.

api_core/google/api_core/bidi.py
plamut: The initial idea was that the lock would serve as a gatekeeper, and if any thread needs to sleep while holding it, other threads trying to enter would also be forced to wait. With a lock, achieving that would be straightforward.

The lock was placed there before implementing the rest of the logic - one could say to avoid accidentally overlooking an edge case withing the logic, yes.
(deque operations themselves are thread-safe, but the rest of the logic surrounding it might not be, thus I went with a conservative approach)

Here's one risky scenario, assume `_past_entries == [1, 2, 3, 4]` and `access_limit == 3`:

```
T   1    2    3    4    5
----|----|----|----|----|--
      |_____________|
                    ↑A
        |____________|
                     ↑B
```

- Thread A enters the manager at T=4+ε, computes its `cutoff_time`, and determines that entry `1` is now irrelevant. 
- Just before left-popping an item, thread B enters at T=4+2ε, determines the same, and removes `1` from the queue. It then figures out that there are already three entries in the window, and goes to sleep.
- Thread A is resumed, and pops the leftmost item, but that's actually entry `2` now! With only `3` and `4` left in the queue, thread proceeds without waiting, because it only saw two items in its window.

Since the queue length and the entries in it are essentially a shared state, having a lock around that avoids tricky scenarios. Besides, the lock is released almost immediately after new entries are again allowed, as the lock-holding thread is put to sleep for just the right amount of time.

Do you concur, or have I overlooked something?

(BTW, double checking concurrency logic is always appreciated, so thanks!)

api_core/google/api_core/bidi.py
```diff
@@ -134,6 +137,69 @@ def __iter__(self):
...
+    instance in the last ``time window`` seconds, the exceeding requests block
+    until enough time elapses.
+
+    The context manager instances are thread-safe and can be shared between
+    multiple threads. If multiple requests are blocked and waiting to enter,
+    the exact order in which they are allowed to proceed is not determined.
+
+    Example::
+
+        max_three_per_second = Throttle(time_window=1, entry_cap=3)
```
plamut: Note to self - typo in class name.

lidizheng:
I might be wrong about the usage of this feature, but I want to suggest an alternative algorithm.

Since `_pastEntries` is private, I guess users don't need to keep the timestamps. In this case, there is a well-known rate limiting algorithm, [token bucket](https://en.wikipedia.org/wiki/Token_bucket), might apply well to this situation.

1) It doesn't need to keep the timestamp, so there is no need for complex data structure;
2) It's fast, the time/space complexity are both O(1).

I found an Python implementation at [rate_limit.py](https://github.com/kburnik/rate-limiter/blob/master/rate_limit.py).

api_core/google/api_core/bidi.py
crwilcox: The lock is by far the easier approach here. There would be some complicated bits to make this lock free. I just was just curious if this was considered :)

TheKevJames:
Updating to `google-cloud-pubsub==0.42.0` seems to be throwing a TypeError which looks to be related to this PR:
```
TypeError: __init__() got an unexpected keyword argument 'throttle_reopen'
  File "main.py", line 48, in main
    flow_control=get_flow_control())
  File "/usr/local/lib/python3.6/site-packages/google/cloud/pubsub_v1/subscriber/client.py", line 212, in subscribe
    manager.open(callback=callback, on_callback_error=future.set_exception)
  File "/usr/local/lib/python3.6/site-packages/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py", line 386, in open
    throttle_reopen=True,
```

Not overly familiar with this code base, but looks like there might need to be a version bump to `api-core` as well, since pubsub 0.42.0 now relies on the new code added in this PR? For reference, installing `google-cloud-pubsub==0.42.0` seems to have picked up the following relevant library versions:
```
google-api-core | 1.11.1
google-auth | 1.6.3
google-cloud-pubsub | 0.42.0
googleapis-common-protos | 1.6.0
grpc-google-iam-v1 | 0.11.4
grpcio | 1.21.1
```

busunkim96:
@TheKevJames We've made that bump and cut another release, please use 0.42.1. Thanks!

PR #8194, opened 2019-05-30
busunkim96 is now the default option for reviewer

sduskis:


.github/CODEOWNERS
```diff
@@ -3,7 +3,7 @@
 
 # Chris Wilcox is the default owner for anything not explicitly
```
busunkim96: Can we change the comment as well for consistency/remove it?

sduskis: I updated the comment.  Good catch.

Issue #7291, opened 2019-02-05
Document using JSON string to create credentials from service account.


tab1293:
It would be beneficial to be able to create a client instance using a service account file whose JSON contents have already been loaded in to memory. Currently `from_service_account_json` only supports loading a service account file from disk. This makes it difficult to create a client in scenarios where you are loading service account JSON contents from some external service and don't necessarily want to store the contents in a file where your application is running.

tseaver:
As a workaround, you can use a tempfile:

```python
import tempfile
from google.cloud.<apilib> import Client

with tempfile.NamedTemporaryFile() as temp:
    temp.write(json_service_account_bytes)
    temp.flush()
    client = Client.from_service_account_json(temp.name)
```

The file will be closed / deleted when the `with` block exits.

tab1293:
Yeah, a tempfile can be used (that's actually what I'm currently doing) but avoiding storing the key on disk is better. You should be able to pass the JSON string directly to the client constructor.

mik-laj:
@tab1293 you could use a code similar to:
```python
credentials = google.oauth2.service_account.Credentials.from_service_account_info(keyfile_dict, scopes=scopes)
client = VideoIntelligenceServiceClient(credentials)
```
Real use case: https://github.com/apache/airflow/blob/master/airflow/contrib/hooks/gcp_api_base_hook.py

tswast:
+1 to https://github.com/googleapis/google-cloud-python/issues/7291#issuecomment-471982254

I never really liked those factory functions. It's better to explicitly create a credentials object with the google-auth library.

tseaver:
Rather than adding a `from_service_account_json` classmethod factory to each of the 95 or so clients which currently have `from_service_account_file` factories, let's instead just document how to construct credentials from JSON.

simonobe:
are you kidding me that this is not supported?:D or is this just not updated as of yet?

mik-laj:
@busunkim96  Is it done?

iennae:
The PR was merged on December 14, and the details look up to date now with a working example. Please reopen if the update is insufficient. Thank you.

PR #8198, opened 2019-05-30
Core: Add new factories for creating client from json.

IlyaFaer:
https://github.com/googleapis/google-cloud-python/issues/7291

### DIFF
--- a/core/google/cloud/client.py
+++ b/core/google/cloud/client.py
@@ -45,8 +45,28 @@ class _ClientFactoryMixin(object):
     _SET_PROJECT = False
 
     @classmethod
+    def from_json_str(cls, json_credentials_str, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


sduskis:
Personally, I think that our docs ought to be better about how to do auth.  Right now, different auth methods are an extra line of code.  

There's a lot of toil around applying this factory pattern to every client which isn't likely to produce a net benefit.

busunkim96:
Since the consensus seems to be that we should prefer constructing credentials through the auth library, I am opting to close this. 

@IlyaFaer Thank you for taking the time to put this together, and please continue to provide feedback on the library experience!

tseaver:
Do we need to open an issue to ensure that we document the preferred "construct credentials separately" example in all the APIs?

PR #8205, opened 2019-05-31
Pubsub: Separate subscriber and publish future documentation.

### LINKED ISSUE #7816: Pub/Sub: wrong inherited documentation for StreamingPullFuture
The [documentation](https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/api/futures.html) for `StreamingPullFuture`'s methods like [`result()`](https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/api/futures.html#google.cloud.pubsub_v1.subscriber.futures.StreamingPullFuture.result) and [`exception()`](https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/api/futures.html#google.cloud.pubsub_v1.subscriber.futures.StreamingPullFuture.exception) doesn't make sense because it describes Publish futures not streaming pull futures. 

[`StreamingPullFuture`](https://github.com/googleapis/google-cloud-python/blob/a42b0f1abe83ce981c9573f4c56acc72e4344a2f/pubsub/google/cloud/pubsub_v1/subscriber/futures.py#L20) inherits [`futures.Future`](https://github.com/googleapis/google-cloud-python/blob/a42b0f1abe83ce981c9573f4c56acc72e4344a2f/pubsub/google/cloud/pubsub_v1/futures.py#L24), that's where the "wrong" documentation got carried over.

anguillanneuf:
This closes #7816. 

This PR aims to fix the docstrings for certain methods on `pubsub_v1.publisher.futures.Future` and `pubsub_v1.subscriber.futures.StreamingPullFuture`. There are cases where the methods would share the same docstrings, but that's not always true. For example, calling `result()` on a publish future should return a message ID, but calling it on a StreamingPullFuture doesn't return a value, it simply keeps the RPC channel open for receiving messages.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/futures.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/futures.py
@@ -36,13 +36,28 @@ class Future(futures.Future):
             :class:`threading.Event` will be created and used.
     """
 
-    # The publishing-side subclass does not need any special behavior
-    # at this time.
-    #
-    # However, there is still a subclass so that if someone attempts
-    # isinstance checks against a publisher-returned or subscriber-returned
-    # future, trying either one against the other returns False.
-    pass
+    def result(self, timeout=None):
+        """Return the message ID or raise an exception.
 
+        This blocks until the message has been published successfully and
+        returns the message ID unless an exception is raised. 
 
-__all__ = ("Future",)
+        Args:
+            timeout (Union[int, float]): The number of seconds before this call
+                times out and raises TimeoutError.
+
+        Returns: 
+            str: The message ID. 
+
+        Raises:
+            ~.pubsub_v1.TimeoutError: If the request times out.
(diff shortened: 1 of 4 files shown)


anguillanneuf:
TODO: add test for publish future.

pubsub/google/cloud/pubsub_v1/publisher/futures.py
```diff
@@ -36,13 +36,28 @@ class Future(futures.Future):
...
+            Exception: For undefined exceptions in the underlying
+                call execution.
+        """
+        # Attempt to get the exception if there is one.
+        # If there is not one, then we know everything worked, and we can
+        # return an appropriate value.
+        err = self.exception(timeout=timeout)
+        if err is None:
+            return self._result
+        raise err
```
tseaver: This new method needs unit tests (one for the error case, one for the normal one).

PR #8206, opened 2019-06-02
Bigquery: Add `page_size` parameter to `QueryJob.result`.

AzemaBaptiste:
Makes the user control page size for a bigquery job.

### DIFF
--- a/bigquery/tests/unit/test_job.py
+++ b/bigquery/tests/unit/test_job.py
@@ -4168,6 +4168,22 @@ class TestQueryJob(unittest.TestCase, _Base):
         self.assertEqual(query_request[1]["query_params"]["timeoutMs"], 900)
         self.assertEqual(reload_request[1]["method"], "GET")
 
+    def test_result_w_page_size(self):
+        query_resource = {
+            "jobComplete": True,
+            "jobReference": {"projectId": self.PROJECT, "jobId": self.JOB_ID},
+            "schema": {"fields": [{"name": "col1", "type": "STRING"}]},
+            "totalRows": "4",
+        }
+        connection = _make_connection(query_resource)
+        client = _make_client(self.PROJECT, connection=connection)
+        resource = self._make_resource(ended=True)
+        job = self._get_target_class().from_api_repr(resource, client)
+
+        result = job.result(page_size=3)
+
+        self.assertEqual(result.total_rows, 4)
+
     def test_result_error(self):
         from google.cloud import exceptions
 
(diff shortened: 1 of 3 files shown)


AzemaBaptiste:
I signed it!

tswast:
Coverage failure

```
tests/unit/test_magics.py                     327      0      4      1    99%   242->exit
```

is unrelated to this change and was fixed in https://github.com/googleapis/google-cloud-python/pull/8179

bigquery/tests/unit/test_job.py
```diff
@@ -4168,6 +4168,22 @@ def test_result_w_timeout(self):
...
+            "totalRows": "4",
+        }
+        connection = _make_connection(query_resource)
+        client = _make_client(self.PROJECT, connection=connection)
+        resource = self._make_resource(ended=True)
+        job = self._get_target_class().from_api_repr(resource, client)
+
+        result = job.result(page_size=3)
+
+        self.assertEqual(result.total_rows, 4)
```
tswast: Let's assert that [tabledata.list](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list) is called with the right page size (`maxResults`) as a parameter.

See:

https://github.com/googleapis/google-cloud-python/blob/097572451bc596ac676d2c16557c79d3944641c4/bigquery/tests/unit/test_client.py#L4197

Specifically, I'd like to see something like:

```
tabledata_path = "/projects/%s/datasets/%s/tables/%s/data" % (
    self.PROJECT,
    self.DS_ID,
    self.TABLE_ID,
)
conn.api_request.assert_called_once_with(
    method="GET", path=tabledata_path, query_params={"maxResults": 3}
)
```

tswast:
Thanks for the contribution @AzemaBaptiste. This is a great addition.

I like the system test. I'd just like to also verify in a unit test that the parameter is actually sent.

bigquery/tests/unit/test_job.py
AzemaBaptiste: Hello Tim,
Thanks for the feedback but I think I need a hand.
Do you mean that the maxResults should be directly linked to the page_size ? The api documentation seems to say it  [link](
https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#queryrequest) but the code don't
https://github.com/googleapis/google-cloud-python/blob/097572451bc596ac676d2c16557c79d3944641c4/bigquery/google/cloud/bigquery/client.py#L1949-L1963

tswast: Confusingly, `page_size` in Python is used to set `maxResults` in the API request, since that's what BigQuery calls its page size parameter.

I've updated your PR to test for this.

AzemaBaptiste: Okay I see...
Many thanks for the help on the unit test !

bokub:
Great job, thanks 🎉

PR #8207, opened 2019-06-03
Prevent requests from hanging on SSL handshake issue by adding a max t.o.

crwilcox:
The auth library defaults to timeout of None, effectively having no timeout. This adds a default timeout of 5 minutes for all users of the core._http implementation

b/132819217

### DIFF
--- a/core/google/cloud/client.py
+++ b/core/google/cloud/client.py
@@ -34,7 +34,7 @@ _GOOGLE_AUTH_CREDENTIALS_HELP = (
 )
 
 # Default timeout for auth requests.
-_REFRESH_TIMEOUT = 300
+_CREDENTIALS_REFRESH_TIMEOUT = 300
 
 
 class _ClientFactoryMixin(object):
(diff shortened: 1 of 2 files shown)


core/google/cloud/client.py
```diff
@@ -33,6 +33,9 @@
     "for help on authentication with this library."
 )
 
+# Default timeout for auth requests.
+_REFRESH_TIMEOUT = 300
```
tseaver: Can we rename this something like `_CREDENTIALS_REFRESH_TIMEOUT`?

core/tests/unit/test_client.py
```diff
@@ -133,7 +133,7 @@ def test__http_property_new(self):
         with authorized_session_patch as AuthorizedSession:
             self.assertIs(client._http, mock.sentinel.http)
             # Check the mock.
-            AuthorizedSession.assert_called_once_with(credentials)
+            AuthorizedSession.assert_called_once_with(credentials, 300)
```
tseaver: Please import the constant and use it.

PR #8225, opened 2019-06-05
Update commands from python 3.6 to python 3.7

brendanlundy:
Update commands from python 3.6 to python 3.7 since system tests can only be run in python 3.7. Still possible to run unit tests in 3.6, but updating that command to 3.7 as well to be consistent.


busunkim96:
@brendanlundy Thank you for the contribution!

PR #8234, opened 2019-06-06
PubSub: Release the state lock before calling the publish api

### LINKED ISSUE #8036: Pub/Sub: publish message hangs waiting for previous publish to timeout
#### Environment details

OS type and version: 4.9.125-linuxkit GNU/Linux
Python version and virtual environment information: Python 2.7.16
google-cloud-pubsub package version: 0.41.0

#### Steps to reproduce

1. Publish a (first) message to PubSub that fails
2. Timeout on the future object returned by the call to `publish()`, _before_ the grpc publish call in the batch thread returns
3. Publish a (second) message to the same topic. This call hangs until the previous call to grpc publish in the batch thread returns. As the default timeout is currently [10 minutes](https://github.com/googleapis/google-cloud-python/blob/4166ddbf1f8b07c2a5e2db8f7955457b5cedd82d/pubsub/google/cloud/pubsub_v1/gapic/publisher_client_config.py#L26) (!) this can take take 10 minutes to return.

Hanging for 10 minutes is surprising behaviour for an asynchronous API.

#### Code example

See this [gist](https://gist.github.com/asnr/36b7f6ee7e2f31f50b1d07020e6ba3f9) for code and instructions on how to reproduce this issue.

#### Wot I think is going on here

This is wild conjecture that I have no supporting evidence for.

That being said, I think the issue starts when the batch thread gets stuck in this [call](https://github.com/googleapis/google-cloud-python/blob/4166ddbf1f8b07c2a5e2db8f7955457b5cedd82d/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L201) to grpc `publish`. At this point it is [holding onto](https://github.com/googleapis/google-cloud-python/blob/4166ddbf1f8b07c2a5e2db8f7955457b5cedd82d/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L201) the lock `_state_lock` and will continue to hold on to it for 10 minutes until it the call to grpc `publish` times out.

When the client application calls `publish()` in the main thread for the second time, it will [try to acquire](https://github.com/googleapis/google-cloud-python/blob/4166ddbf1f8b07c2a5e2db8f7955457b5cedd82d/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L280) the same lock `_state_lock`. As this lock is already being held by the batch thread, the main thread hangs and doesn't return from the call to `publish()`.

plamut:
Closes #8036.

Supersedes #7686 (the fix here is essentially the same, but also adds tests and a few minor things).

### Summary
This PR makes sure that calls to `publisher.publish()` are not unnecessarily blocked while the underlying API publish call is in progress - the latter can take a substantial amount of time, especially on network errors when retries kick in.

### Description
The fix makes sure that the lock in the `_commit()` method is held for no longer than necessary. The details on why releasing the lock earlier is fine are listed in the [comment](https://github.com/googleapis/google-cloud-python/pull/7686#discussion_r284314182) on the original PR.

### How to test
**Steps to reproduce:**
- Set the total publish timeout [setting](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/gapic/publisher_client_config.py#L35) to 60,000 milliseconds (in order to not wait for too long),
- Disable the internet connection on your machine,
- Run the test publisher script:
```py
import logging

log_format = (
)
logging.basicConfig(
    level=logging.DEBUG,
    format=log_format,
)
logger = logging.getLogger(__name__)

PROJECT_NAME = "some-pubsub-project"
TOPIC_NAME = "my-topic"


def main():
    topic_path = publisher.topic_path(PROJECT_NAME, TOPIC_NAME)

    message_future = publisher.publish(topic_path, b"I'm going to timeout")

    try:
        message_future.exception(timeout=3)
    except Exception as exc:
        logger.info("\x1b[1mException raised:\x1b[0m {}".format(exc))

    logger.info("\x1b[1mAbout to publish the second message\x1b[0m")
    publisher.publish(topic_path, b"I'm going to take quite some time to return")
    logger.info("\x1b[1mReturned from second call to publish()\x1b[0m")


if __name__ == "__main__":
    main()
```

The message _"Returned from second call to publish"_ appears in the logs only after a minute or so after the _"About to publish the second message"_.

The second message is logged shortly after the first one. The second call to `publisher.publish()` is not blocked while publishing the first message is in progress and hitting the automatic retry because of the network error.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -188,10 +188,11 @@ class Batch(base.Batch):
                 return
 
         # Once in the IN_PROGRESS state, no other thread can publish additional
-        # messages or initiate a commit (those operations becme a no-op), thus
+        # messages or initiate a commit (those operations become a no-op), thus
         # it is safe to release the state lock here. Releasing the lock avoids
         # blocking other threads in case api.publish() below takes a long time
         # to complete.
+        # https://github.com/googleapis/google-cloud-python/issues/8036
 
         # Sanity check: If there are no messages, no-op.
         if not self._messages:
@@ -264,7 +265,8 @@ class Batch(base.Batch):
 
         Add the given message to this object; this will cause it to be
         published once the batch either has enough messages or a sufficient
-        period of time has elapsed.
+        period of time has elapsed. If the batch is full or the commit is
+        already in progress, the method does not do anything.
 
         This method is called by :meth:`~.PublisherClient.publish`.
 
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -187,56 +187,62 @@ class Batch(base.Batch):
                 _LOGGER.debug("Batch is already in progress, exiting commit")
                 return
 
-            # Sanity check: If there are no messages, no-op.
-            if not self._messages:
-                _LOGGER.debug("No messages to publish, exiting commit")
-                self._status = base.BatchStatus.SUCCESS
-                return
+        # Once in the IN_PROGRESS state, no other thread can publish additional
+        # messages or initiate a commit (those operations becme a no-op), thus
+        # it is safe to release the state lock here. Releasing the lock avoids
+        # blocking other threads in case api.publish() below takes a long time
+        # to complete.
 
-            # Begin the request to publish these messages.
-            # Log how long the underlying request takes.
-            start = time.time()
+        # Sanity check: If there are no messages, no-op.
+        if not self._messages:
+            _LOGGER.debug("No messages to publish, exiting commit")
+            self._status = base.BatchStatus.SUCCESS
+            return
 
-            try:
-                response = self._client.api.publish(self._topic, self._messages)
-            except google.api_core.exceptions.GoogleAPIError as exc:
-                # We failed to publish, set the exception on all futures and
-                # exit.
-                self._status = base.BatchStatus.ERROR
+        # Begin the request to publish these messages.
+        # Log how long the underlying request takes.
+        start = time.time()
 
-                for future in self._futures:
-                    future.set_exception(exc)
+        try:
+            response = self._client.api.publish(self._topic, self._messages)
+        except google.api_core.exceptions.GoogleAPIError as exc:
+            # We failed to publish, set the exception on all futures and
+            # exit.
+            self._status = base.BatchStatus.ERROR
 
-                _LOGGER.exception("Failed to publish %s messages.", len(self._futures))
-                return
+            for future in self._futures:
+                future.set_exception(exc)
 
-            end = time.time()
-            _LOGGER.debug("gRPC Publish took %s seconds.", end - start)
+            _LOGGER.exception("Failed to publish %s messages.", len(self._futures))
+            return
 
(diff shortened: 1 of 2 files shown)


pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -149,6 +151,35 @@ def test_blocking__commit():
...
+    )
+
+    with api_publish_patch:
+        batch.publish({"data": b"first message"})
+
+        start = datetime.datetime.now()
+        event_set = api_publish_called.wait(timeout=1.0)
+        if not event_set:
+            pytest.fail("API publish was not called in time")
+        batch.publish({"data": b"second message"})
```
tswast: Won't this always take 1 second due to the `side_effect` patch?

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -187,56 +187,62 @@ def _commit(self):
...
-                # exit.
-                self._status = base.BatchStatus.ERROR
-
-                for future in self._futures:
-                    future.set_exception(exc)
+        # Once in the IN_PROGRESS state, no other thread can publish additional
+        # messages or initiate a commit (those operations becme a no-op), thus
+        # it is safe to release the state lock here. Releasing the lock avoids
+        # blocking other threads in case api.publish() below takes a long time
+        # to complete.
```
tswast: Link to the GitHub issue too, for more context?

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -187,56 +187,62 @@ def _commit(self):
...
-                response = self._client.api.publish(self._topic, self._messages)
-            except google.api_core.exceptions.GoogleAPIError as exc:
-                # We failed to publish, set the exception on all futures and
-                # exit.
-                self._status = base.BatchStatus.ERROR
-
-                for future in self._futures:
-                    future.set_exception(exc)
+        # Once in the IN_PROGRESS state, no other thread can publish additional
+        # messages or initiate a commit (those operations becme a no-op), thus
```
tswast: Typo: `beme` should be `become`

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
tswast: It may be good to have a similar comment on operations that add messages / make a commit so that they continue to be a no-op when in `IN_PROGRESS` state.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
plamut: That's the gist of the test - if the lock is held by the batch commit thread while the batch is being committed, i.e. when the code is waiting for `self._client.api.publish()` to return, the operation will take more than one second and the test will fail.

If the lock is released right after the state transition to `IN_PROGRESS`, i.e. before the call to `self._client.api.publish()`, the latter will not block other threads that might (indirectly) call `batch.publish()` in the meantime and try to acquire the state lock.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
plamut: Could be useful to some, will add that.

PR #8260, opened 2019-06-06
Vision: Add AsyncBatchAnnotateFiles system test

brendanlundy:


vision/tests/system.py
```diff
@@ -30,12 +30,9 @@
 
 
 _SYS_TESTS_DIR = os.path.realpath(os.path.dirname(__file__))
-LOGO_FILE = os.path.join(_SYS_TESTS_DIR, "data", "logo.png")
```
busunkim96: If these aren't used, could they be deleted from _SYS_TESTS_DIR?

brendanlundy: Yes, removed the unused images.

PR #8261, opened 2019-06-06
Update codeowners.

busunkim96:
All guesses were made in good faith. :)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,6 +5,28 @@
 # taken by someone else.
 *                         @busunkim96
 
+##################################
+# Storage and Database (SoDa) team 
+##################################
+bigtable/                   @tseaver @crwilcox @frankyn
+docs/bigtable/              @tseaver @crwilcox @frankyn
+
+datastore/                  @tseaver @crwilcox @frankyn      
+docs/datastore              @tseaver @crwilcox @frankyn 
+
+storage/                    @tseaver @crwilcox @frankyn
+docs/storage                @tseaver @crwilcox @frankyn
+
+firestore/                  @tseaver @crwilcox @frankyn
+docs/firestore/             @tseaver @crwilcox @frankyn
+
+spanner/                    @tseaver @crwilcox @frankyn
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -15,9 +37,10 @@ docs/bigquery_datatransfer/ @googleapis/api-bigquery
...
 
-# Danny Hermes and Tres Seaver co-authored the BigTable library.
-bigtable/                   @tseaver
-docs/bigtable/              @tseaver
+# Pubsub
+pubsub/                     @anguillanneuf @plamut
+docs/pubsub                 @anguillanneuf @plamut
+
 
 # Bill Prin is the primary maintainer of the Stackdriver libraries.
```
busunkim96: This person seems to be gone, is it OK to remove this section? @tseaver

.github/CODEOWNERS
```diff
@@ -5,6 +5,28 @@
...
+docs/datastore              @tseaver @crwilcox @frankyn 
+
+storage/                    @tseaver @crwilcox @frankyn
+docs/storage                @tseaver @crwilcox @frankyn
+
+firestore/                  @tseaver @crwilcox @frankyn
+docs/firestore/             @tseaver @crwilcox @frankyn
+
+spanner/                    @tseaver @crwilcox @frankyn
+docs/spanner                @tseaver @crwilcox @frankyn
```
tseaver: Rather than perpetuating the churn in this file, would it make more sense to create teams in `sloth` and then update this file to use them?  E.g., `yoshi-soda`, `yoshi-animl`, `yoshi-stackdriver`?

busunkim96: The yoshi-* ones seem to be maintained in Sloth, but there are others (like api-bigquery) that I assume were created manually. I'm not quite sure what the story is with teams, I'll bring this up in the team chat.

Thank you for the suggestion!

.github/CODEOWNERS
crwilcox: I think this is safe to remove. This will default to you.

PR #8264, opened 2019-06-07
Firestore: Add gRPC keepalive to Client Init.

crwilcox:
In order to run a listen for longer than one our, gRPC needs to be configured to keep the connection alive. These changes are in generated code here though and shouldn't be merged.

b/130522486

### DIFF
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -116,7 +116,7 @@ class Client(ClientWithProject):
             # Use a custom channel.
             # We need this in order to set appropriate keepalive options.
             channel = firestore_grpc_transport.FirestoreGrpcTransport.create_channel(
-                target=self._target,
+                self._target,
                 credentials=self._credentials,
                 options={"grpc.keepalive_time_ms": 30000}.items(),
             )
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -23,6 +23,7 @@ In the hierarchy of API concepts
 * a :class:`~google.cloud.firestore_v1.client.Client` owns a
   :class:`~google.cloud.firestore_v1.document.DocumentReference`
 """
+from google.api_core import grpc_helpers
 from google.api_core.gapic_v1 import client_info
 from google.cloud.client import ClientWithProject
 
@@ -36,6 +37,7 @@ from google.cloud.firestore_v1.document import DocumentReference
 from google.cloud.firestore_v1.document import DocumentSnapshot
 from google.cloud.firestore_v1.field_path import render_field_path
 from google.cloud.firestore_v1.gapic import firestore_client
+from google.cloud.firestore_v1.gapic.transports import firestore_grpc_transport
 from google.cloud.firestore_v1.transaction import Transaction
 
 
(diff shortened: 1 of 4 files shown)


firestore/google/cloud/firestore_v1/client.py
```diff
@@ -23,6 +23,7 @@
 * a :class:`~google.cloud.firestore_v1.client.Client` owns a
   :class:`~google.cloud.firestore_v1.document.DocumentReference`
 """
+from google.api_core import grpc_helpers
```
tseaver: This module isn't generated code.

firestore/google/cloud/firestore_v1/client.py
```diff
@@ -112,8 +114,23 @@ def _firestore_api(self):
             <The GAPIC client with the credentials of the current client.
         """
         if self._firestore_api_internal is None:
+            SERVICE_ADDRESS = "firestore.googleapis.com:443"
+
+            # Use a custom channel.
+            # We need this in order to set appropriate keepalive options.
+            channel = grpc_helpers.create_channel(
+                target=SERVICE_ADDRESS,
```
tseaver: This should probably follow the pattern in pubsub: https://github.com/googleapis/google-cloud-python/blob/b80d431e214c62fb506461a3f1485b58bf38f289/pubsub/google/cloud/pubsub_v1/publisher/client.py#L74-L96

crwilcox: I used PubSub for inspiration, but didn't want to extend the client further. Do you have a specific way in which this is different that you disagree with?

tseaver: A couple:
- DRY the service address (using `self.target`).
- Passing in a transport to the client is how the GAPIC-generator folks want us to customize such stuff, but doing it here (rather than in `__init__` means we can't let the caller pass it in to override our choices).

crwilcox: @tseaver, as this is about addressing a bug, I wasn't intending to modify the user surface. the `self.target` callout is a good idea though.

Also, I opened https://github.com/googleapis/gapic-generator/issues/2824 to see about getting better exposure the options in the generator, which would minimize the change needed here to specify grpc options.

firestore/google/cloud/firestore_v1/client.py
crwilcox: Correct. Is there a reason the change seems incorrect for manual code?

tseaver: Your comment above says:

> These changes are in generated code here though and shouldn't be merged.

I was pointing out that none of this PR touches generated code.

crwilcox: Ah. Thanks. You are correct. I originally did this in generated code and then made modifications

firestore/google/cloud/firestore_v1/client.py
tseaver: @crwilcox Did you mean to push a change which used the `self.target` DRYness?

firestore/google/cloud/firestore_v1/client.py
```diff
@@ -112,12 +113,33 @@ def _firestore_api(self):
...
         return self._firestore_api_internal
 
+    @property
+    def _target(self):
+        """Return the target (where the API is).
+
+        Returns:
+            str: The location of the API.
+        """
+        return firestore_client.FirestoreClient.SERVICE_ADDRESS
```
tseaver: This is going to clash with @busunkim96's `client_info` work.

PR #8265, opened 2019-06-07
Core: Add client options.

busunkim96:
This *just* adds API Endpoint (but more options will be coming, like credentials).

I manually edited the IoT client to show what changes need to be made to the generated clients. I will undo those changes before merging. 

[go/extensible-options-in-python ](https://docs.google.com/document/d/1F4pFautf-3wvf_-FnmTIVdUO0gMFlVe2gx4b_z_TiOE/edit#) <-- shared with yoshi-team@google.com
go/extensible-options-in-client

CC @shinfan


api_core/google/api_core/client_options.py
```diff
@@ -0,0 +1,47 @@
...
+        api_endpoint (str): The desired API endpoint, e.g., compute.googleapis.com
+    """
+
+    api_endpoint = None
+
+    def __init__(self, options=None, **kw_args):
+        if options is not None:
+            if kw_args.items():
+                raise Exception(
+                    "ClientOptions expects options in a dictionary or in kw_args, not both"
```
busunkim96: I had some trouble wordsmithing here. Suggestions would be appreciated.

tseaver: It's a new class:  why don't we just use named arguments, so we can document them properly?  We don't have any backward-compatibility reason to deal with arbitrary arguments here.

iot/google/cloud/iot_v1/gapic/device_manager_client.py
```diff
@@ -142,6 +144,9 @@ def __init__(
                 API requests. If ``None``, then default info will be used.
                 Generally, you only need to set this if you're developing
                 your own client library.
+            client_options (google.api_core.client_options.ClientOptions):
+                Client options used to set user options on the client. API Endpoint
+                should be set through client_options.
```
tseaver: This is going to be very confusing for users:  we have *three* similarly-named arguments, one of which is deprecated, all of which exist to allow users to override bits of the default configuration for the client. Could weadd the `api_endpoint` attribute to `google.api_core.client_info.ClientInfo`?

api_core/google/api_core/client_options.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Client options class.
+
+Client options provide an consistent interface for user options to be defined
```
shinfan: nit: an/a

api_core/google/api_core/client_options.py
shinfan: +1 I think we can just use the 'options' here

iot/google/cloud/iot_v1/gapic/device_manager_client.py
shinfan: This  client options class is designed to be the central place for client-level user configuration, not just for endpoints. We are planning to add credentials, pre-defined headers and other stuff. 

I agree that right now it seems like we have 'client_config', 'client_info' and 'client_options' which all look the same. Is it possible that we merge these similar classes into one?

shinfan:
CC @wora

iot/google/cloud/iot_v1/gapic/device_manager_client.py
busunkim96: We can look into rolling the others into client_options, since that is the "new" name that's going to be used across clients.

api_core/google/api_core/client_options.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def from_dict(options):
+    """Construct a client options object from a dictionary.
+
+    Args:
+        options (dict): A dictionary with client options.
+    """
+
+    client_options = ClientOptions()
```
busunkim96: This would allow users to do something like:

```client = pubsub_v1.PublisherClient(client_options={"api_endpoint": "...", "user_project": "..."})```

File: iot/google/cloud/iot_v1/gapic/device_manager_client.py (hunk 2 above)
Comment (tseaver) [reply iot/google/cloud/iot_v1/gapic/device_manager_client.py]: @busunkim96 I don't think you've addressed the confusion factor here:  why do we want people to configure the client using both `client_info` and `client_options` (leaving aside my never-answered question of why `client_config` is deprecated)?

Comment (crwilcox) [reply iot/google/cloud/iot_v1/gapic/device_manager_client.py]: Please correct me if any of this isn't correct, but I think:
`client_config` -> this is about gRPC timeout/retry configuration
`client_info` -> this should probably named `user_agent_info` or something more accurate. This is the data attached to requests to identify the particular client better.
`client_options` ->  this new property bag, currently for endpoints. Eventually also additional things.

It is possible `user_agent_info` should also be under `client_options`

Comment (tseaver) [reply iot/google/cloud/iot_v1/gapic/device_manager_client.py]: Your description of the three arguments roles is correct.  I'm arguing that three is two too many. Users should only need to thing about one thing to pass in to customize the behavior of the client.

@lukesneeringer Can you please comment on why `client_config` is deprecated?  How else would users be expected to tweak e.g. gRPC timeouts for client-invoked methods?

Comment (busunkim96) [reply iot/google/cloud/iot_v1/gapic/device_manager_client.py]: Here is a [link](https://docs.google.com/document/d/1F4pFautf-3wvf_-FnmTIVdUO0gMFlVe2gx4b_z_TiOE/edit#heading=h.4gcz995d4bcx) to the extensible options in Python doc that should be accessible to all of yoshi-team@google.com. It has more context about Client Options and what we're looking to add to it in the future.

The goal of `client_options` is to have a more straight-forward way to configure clients. We want to provide a consistent interface to set options across the generated, handwritten, and Apiary clients. The experience will also be more predictable across languages.

User agent is on the list of things we would like to be configurable with `client_options`. I think we want to move what is in `client_info` (use agent) into `client_options` (and deprecate `client_info`).

File: api_core/tests/unit/test_client_options.py [hunk 5]
```diff
@@ -0,0 +1,36 @@
+# Copyright 2017 Google LLC
```
tswast: 2019

shinfan:
Thanks for reviewing this Tim.

PR #8269, opened 2019-06-10
Vision: Add negative Vision API Product Search tests for VPC SC

brendanlundy:
Verifies ProductSearch is blocked by VPC SC when trying to access a resource outside of a secure perimeter when run from within a secure perimeter.

### DIFF
--- a/vision/tests/system.py
+++ b/vision/tests/system.py
@@ -21,6 +21,7 @@ import os
 import time
 import unittest
 
+import google.api_core.exceptions
 from google.cloud import exceptions
 from google.cloud import storage
 from google.cloud import vision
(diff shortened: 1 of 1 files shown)


vision/tests/system.py
```diff
@@ -34,6 +35,7 @@
 LOGO_FILE = os.path.join(_SYS_TESTS_DIR, "data", "logo.png")
 PDF_FILE = os.path.join(_SYS_TESTS_DIR, "data", "pdf_test.pdf")
 PROJECT_ID = os.environ.get("PROJECT_ID")
+PROJECT_OUTSIDE = os.environ.get("GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT")
```
busunkim96: Should we be setting this environment variable in our tests environment? Or are these tests being run somewhere else?

brendanlundy: The VPC SC test is being run elsewhere. No need for you to set this environment variable.

PR #8227, opened 2019-06-05
Logging: Set the 'timestamp' on log records created by handler.

crazystick:
There's no easy way to inject your own Worker thread that I can see, but I don't know why you wouldn't want to know the original time of the log entry.

Closes #8222

Issue #7504, opened 2019-03-12
Spanner:  'test_transaction_batch_update_and_execute_dml' systest flakes with unexpected status code.


tseaver:
From: https://source.cloud.google.com/results/invocations/1fe1c70e-9abd-44c2-ae2f-426dec16730f/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fspanner/log

```python
_________ TestSessionAPI.test_transaction_batch_update_and_execute_dml _________
self = <tests.system.test_system.TestSessionAPI testMethod=test_transaction_batch_update_and_execute_dml>
    def test_transaction_batch_update_and_execute_dml(self):
        retry = RetryInstanceState(_has_all_ddl)
        retry(self._db.reload)()
        session = self._db.session()
        session.create()
        self.to_delete.append(session)
        with session.batch() as batch:
            batch.delete(self.TABLE, self.ALL)
        insert_statements = list(self._generate_insert_statements())
        update_statements = [
            (
                "UPDATE contacts SET email = @email " "WHERE contact_id = @contact_id;",
                {"contact_id": 1, "email": "phreddy@example.com"},
                {"contact_id": Type(code=INT64), "email": Type(code=STRING)},
            )
        ]
        delete_statement = "DELETE contacts WHERE TRUE;"
        def unit_of_work(transaction, self):
            rows = list(transaction.read(self.TABLE, self.COLUMNS, self.ALL))
            self.assertEqual(rows, [])
            status, row_counts = transaction.batch_update(
                insert_statements + update_statements
            )
            self.assertEqual(status.code, 0)  # XXX: where are values defined?
            self.assertEqual(len(row_counts), len(insert_statements) + 1)
            for row_count in row_counts:
                self.assertEqual(row_count, 1)
            row_count = transaction.execute_update(delete_statement)
            self.assertEqual(row_count, len(insert_statements))
>       session.run_in_transaction(unit_of_work, self)
tests/system/test_system.py:859:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/spanner_v1/session.py:295: in run_in_transaction
    return_value = func(txn, *args, **kw)
tests/system/test_system.py:850: in unit_of_work
    self.assertEqual(status.code, 0)  # XXX: where are values defined?
[...]

PR #8273, opened 2019-06-11
Logging: Fix tests.

busunkim96:
Fixes internal issue 134713268.

### DIFF
--- a/logging/google/cloud/logging/handlers/transports/background_thread.py
+++ b/logging/google/cloud/logging/handlers/transports/background_thread.py
@@ -20,6 +20,7 @@ Uses a background worker to log to Stackdriver Logging asynchronously.
 from __future__ import print_function
 
 import atexit
+import datetime
 import logging
 import sys
 import threading
--- a/logging/google/cloud/logging/handlers/transports/background_thread.py
+++ b/logging/google/cloud/logging/handlers/transports/background_thread.py
@@ -262,7 +262,7 @@ class _Worker(object):
                 "labels": labels,
                 "trace": trace,
                 "span_id": span_id,
-                "timestamp": datetime.utcfromtimestamp(record.created),
+                "timestamp": datetime.datetime.utcfromtimestamp(record.created),
             }
         )
 
--- a/logging/tests/unit/handlers/transports/test_background_thread.py
+++ b/logging/tests/unit/handlers/transports/test_background_thread.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
 import logging
 import unittest
 
@@ -243,7 +244,9 @@ class Test_Worker(unittest.TestCase):
         worker = self._make_one(_Logger(self.NAME))
 
         self._start_with_thread_patch(worker)
-        worker.enqueue(mock.Mock(), "")
+        record = mock.Mock()
+        record.created = time.time()
+        worker.enqueue(record, "")
         worker._main_thread_terminated()
 
         self.assertFalse(worker.is_alive)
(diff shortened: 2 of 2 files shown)


bmccutchon:
I think you need to also change the usage to `datetime.datetime.utcfromtimestamp` to fix the build.

tseaver:
Also, we need to figure out why #8227 passed CI here.

busunkim96:
@tseaver We forgot to add `kokoro: force-run` on that PR.

busunkim96:
Looks like there's some more work to get tests passing again (after fixing the import)

```
self = <google.cloud.logging.handlers.transports.background_thread._Worker object at 0x7f48d3fd15f8>

    def _thread_main(self):
        """The entry point for the worker thread.
    
        Pulls pending log entries off the queue and writes them in batches to
        the Cloud Logger.
        """
        _LOGGER.debug("Background thread started.")
    
        done = False
        while not done:
            batch = self._cloud_logger.batch()
            items = _get_many(
                self._queue,
                max_items=self._max_batch_size,
                max_latency=self._max_latency,
            )
    
            for item in items:
                if item is _WORKER_TERMINATOR:
                    done = True  # Continue processing items.
                else:
>                   batch.log_struct(**item)
E                   TypeError: log_struct() got an unexpected keyword argument 'timestamp'

google/cloud/logging/handlers/transports/background_thread.py:148: TypeError

```

tseaver:
I don't understand the `TypeError`.  `Batch.log_struct` is a straightforward wrapper: https://github.com/googleapis/google-cloud-python/blob/c5acdad71bbbc1d58b8e6e5f0a749660823a88b3/logging/google/cloud/logging/logger.py#L338-L348

creating a `StructEntry` https://github.com/googleapis/google-cloud-python/blob/c5acdad71bbbc1d58b8e6e5f0a749660823a88b3/logging/google/cloud/logging/entries.py#L295

which derives from `LogEntry`: https://github.com/googleapis/google-cloud-python/blob/c5acdad71bbbc1d58b8e6e5f0a749660823a88b3/logging/google/cloud/logging/entries.py#L147

which derives from `_LogEntryTuple`: https://github.com/googleapis/google-cloud-python/blob/c5acdad71bbbc1d58b8e6e5f0a749660823a88b3/logging/google/cloud/logging/entries.py#L67-L87

which clearly takes a `timestamp`.

tseaver:
OK, the `batch` in this case is a hand-rolled mock: https://github.com/googleapis/google-cloud-python/blob/c5acdad71bbbc1d58b8e6e5f0a749660823a88b3/logging/tests/unit/handlers/transports/test_background_thread.py#L420-L446

There will also be test failures due to the use of a `mock.Mock` passed to `_Worker.enqueue`:

```python

    def enqueue(
        self, record, message, resource=None, labels=None, trace=None, span_id=None
    ):
        """Queues a log entry to be written by the background thread.
    
        :type record: :class:`logging.LogRecord`
        :param record: Python log record that the handler was called with.
    
        :type message: str
        :param message: The message from the ``LogRecord`` after being
                        formatted by the associated log formatters.
    
        :type resource: :class:`~google.cloud.logging.resource.Resource`
        :param resource: (Optional) Monitored resource of the entry
    
        :type labels: dict
        :param labels: (Optional) Mapping of labels for the entry.
    
        :type trace: str
        :param trace: (optional) traceid to apply to the logging entry.
    
        :type span_id: str
        :param span_id: (optional) span_id within the trace for the log entry.
                        Specify the trace parameter if span_id is set.
        """
        self._queue.put_nowait(
            {
                "info": {"message": message, "python_logger": record.name},
                "severity": record.levelname,
                "resource": resource,
                "labels": labels,
                "trace": trace,
                "span_id": span_id,
>               "timestamp": datetime.datetime.utcfromtimestamp(record.created),
            }
        )
E       TypeError: an integer is required (got type Mock)
```

bmccutchon:
If this is going to take a long time to fix, would it make sense to roll back PR #8227 in the meantime?

tseaver:
- Monitoring and Webrisk fail due to Kokoro-internal error: "Cannot check out non-mergeable pull request as if merged."
- Firestore fails due to #7130.
- Spanner fails due to #7504.

PR #8277, opened 2019-06-11
PubSub: Add system tests for PubSub clients

### LINKED ISSUE #4031: Missing system tests after pubsub redesign
The redesign effort (#3859) left behind only very minimal system tests.  The older implementation had much broader system test coverage:

- Listing topics and subscriptions in the client's project.
- Creating subscriptions with non-default settings.
- Listing subscriptions bound to a topic.
- Setting / getting IAM policy for topics and subscriptions
- Creating / seeking snapshots.

I noticed the missing tests today when I went looking for examples of updating a topic's IAM policy.

plamut:
Closes #4031.

The title says it all - this PR adds the missing system tests mentioned in the issue description:
- Listing topics and subscriptions in the client's project.
- Creating subscriptions with non-default settings.
- Listing subscriptions bound to a topic.
- Setting / getting IAM policy for topics and subscriptions
- ~~Creating / seeking snapshots.~~ :arrow_left: Removed, because the server only guarantees eventual re-delivery, and a test cannot wait for a fixed amount of time, causes flakiness.

### DIFF
--- a/pubsub/tests/system.py
+++ b/pubsub/tests/system.py
@@ -297,48 +297,6 @@ def test_listing_topic_subscriptions(publisher, subscriber, project, cleanup):
     assert subscriptions == {subscription_paths[0], subscription_paths[2]}
 
 
-def test_using_snapshots(
-    publisher, subscriber, project, topic_path, subscription_path, cleanup
-):
-    snapshot_path = subscriber.snapshot_path(project, "my-snapshot")
-
-    # Make sure the topic, subscription, and snapshot get deleted.
-    cleanup.append((publisher.delete_topic, topic_path))
-    cleanup.append((subscriber.delete_subscription, subscription_path))
-    cleanup.append((subscriber.delete_snapshot, snapshot_path))
-
-    # publish a message to a topic, pull and acknowledge it
-    publisher.create_topic(topic_path)
-    subscriber.create_subscription(subscription_path, topic_path)
-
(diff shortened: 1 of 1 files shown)


pubsub/tests/system.py
```diff
@@ -183,6 +184,218 @@ def test_subscribe_to_messages_async_callbacks(
...
+    # pulling again, there should be no messages received
+    response = subscriber.pull(
+        subscription_path, max_messages=100, return_immediately=True
+    )
+    assert len(response.received_messages) == 0
+
+    # after seeking to the snapshot, the second message should be received again
+    subscriber.seek(subscription_path, snapshot=snapshot_path)
+    response = subscriber.pull(subscription_path, max_messages=100)
+    assert len(response.received_messages) == 1
```
plamut: This fails when run with other system tests, a message is not received (but seems to consistently pass when run in isolation). From the [snapshot docs](https://cloud.google.com/pubsub/docs/replay-qs):
> You can recover the message by seeking the subscription to the snapshot. Note that the message may not be delivered instantly: Pub/Sub does not guarantee message delivery latency. However, what is guaranteed is that the messages retained in the snapshot will be eventually delivered at least once

Since the message will only be _eventually_ delivered, this test is apparently flaky. Will have to think a way around this...

plamut: Removed the test case, as the test cannot wait for the message to "eventually" arrive ... or not.

PR #8293, opened 2019-06-12
PubSub: Document FlowControl settings

### LINKED ISSUE #7043: Pub/Sub doc: missing definition for `max_lease_duration` in FlowControl
The ref doc for [`max_lease_duration`](https://googleapis.github.io/google-cloud-python/latest/pubsub/types.html#google.cloud.pubsub_v1.types.FlowControl.max_lease_duration) does not describe what it is or how it is used, especially about how it interacts with ack deadlines. 

Given that this is an important config setting for long-running tasks, we may want to provide more documentation here by showing its default value (is it 2 hours), and explaining whether max lease duration restrains ack deadline modification or ack deadline modification overwrites max lease duration when one configures both flow control settings and ack deadlines.

plamut:
Closes #7043.

This PR makes sure that FlowControl settings are properly documented in the generated documentation.

### How to test
**Steps to perform:**

(from inside the `pubsub/` directory)
- Install the [synthtool](https://github.com/googleapis/synthtool) and run it,
- Generate the pubsub docs locally:
    ```
    $ nox -f noxfile.py -s docs
    ```
- Open the generated docs for `FlowControl`:
    ```
    $ chromium-browser "file://$GOOG_CLOUD_PYTHON_DIR/pubsub/docs/_build/html/types.html#google.cloud.pubsub_v1.types.FlowControl"
    ```

**Expected result:**
`FlowControl` settings have nice descriptions (as opposed to the [current docs](https://googleapis.github.io/google-cloud-python/latest/pubsub/types.html?highlight=flowcontrol#google.cloud.pubsub_v1.types.FlowControl)).

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/types.py
+++ b/pubsub/google/cloud/pubsub_v1/types.py
@@ -71,6 +71,38 @@ FlowControl.__new__.__defaults__ = (
     2 * 60 * 60,  # max_lease_duration: 2 hours.
 )
 
+if sys.version_info >= (3, 5):
+    FlowControl.__doc__ = (
+        "The settings for controlling the rate at which messages are pulled "
+        "with an asynchronous subscription."
+    )
+    FlowControl.max_bytes.__doc__ = (
+        "The maximum total size of received - but not yet processed - messages "
+        "before pausing the message stream."
+    )
+    FlowControl.max_messages.__doc__ = (
+        "The maximum number of received - but not yet processed - messages before "
+        "pausing the message stream."
+    )
+    FlowControl.resume_threshold.__doc__ = (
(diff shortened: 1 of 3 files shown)


anguillanneuf:
The generated docs look good to me. 

Nit: I see the word "Property" in front of each property in the generated docs. Not a big deal, but would be great if we can make it consistent with the rest of the documentation.

plamut:
The "property" prefix seems to be added to all namedtuple members, including those that do not have their docstrings modified. It's probably the `Sphinx` itself that adds them, and the output might depend on the latter's version.

How are the current online docs generated and published anyway? Would a result be different if _that_ was used instead of the local docs build?

FWIW, I get the same result (the "property" prefix) if building the docs locally on the `master` branch, making me believe that it's likely something unrelated to the changes in this PR ...

PR #8295, opened 2019-06-12
Monitoring: Update noxfile.py

steinwaywhw:
The `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT` should be a project outside of the VPCSC perimeter. `PROJECT_ID` is a project inside the perimeter. Do not uncomment until we have a project for the second environment variable. Otherwise this wrong value is breaking the system tests of the VPCSC team.

### DIFF
--- a/monitoring/noxfile.py
+++ b/monitoring/noxfile.py
@@ -122,12 +122,12 @@ def system(session):
     session.install("-e", ".")
 
     # Additional setup for VPCSC system tests
-    env = {
-        "PROJECT_ID": "secure-gcp-test-project-4",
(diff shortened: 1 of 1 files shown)


busunkim96:
Hi @steinwaywhw, I think this was my mistake when I set the environment variables. 

Would a setup like this result in the tests running correctly?

PROJECT_OUTISDE -> the default project we use for tests in this repo (outside the perimeter)
PROJECT_INSIDE -> secure-gcp-test-project

steinwaywhw:
Hi @busunkim96 no worries. Yes, your suggested value for the environment variables would work.

tseaver:
Superseded by #8302.

PR #8300, opened 2019-06-13
Firestore error get all removes non existing snapshot fixed

### LINKED ISSUE #7564: Firestore: get_all removes .id key for non existing DocumentSnapshot
When I use the `get_all` function to load documents and some are non-existing the DocumentSnapshot which are non-existing lose their `.id` key. That is unfortunate as I don't know which document lost its ID. However, it's still possible to call `.exists`.

Requesting the exact same non-existing object from the database, also return a DocumentSnapshot while this time, the `.id` can be referenced and exists.

#### Environment details

1.) Firestore
2.) Linux-4.14.104-78.84.amzn1.x86_64-x86_64-with-glibc2.9
3.)  Python 3.6.5 :: Anaconda custom (64-bit)
4.) google-cloud-firestore==0.31.0

#### Code example

```python

print(db.collection("something").document('thisDoesNotExists').get().exists)
# produces: False
print(db.collection("something").document('thisDoesNotExists').get().id)
# produces: thisDoesNotExists
print(list(db.get_all([db.collection("something").document('thisDoesNotExists')]))[0].exists)
# produces: False
print(list(db.get_all([db.collection("something").document('thisDoesNotExists')]))[0].id)
# produces: See Stack trace
```

#### Stack trace
```
AttributeError                            Traceback (most recent call last)
<ipython-input-218-5094e5147636> in <module>()
      2 print(db.collection("something").document('thisDoesNotExists').get().id)
      3 print(list(db.get_all([db.collection("something").document('thisDoesNotExists')]))[0].exists)
----> 4 print(list(db.get_all([db.collection("something").document('thisDoesNotExists')]))[0].id)

~/anaconda3/envs/python3/lib/python3.6/site-packages/google/cloud/firestore_v1beta1/document.py in id(self)
    592             str: The last component of the path of the document.
    593         """
--> 594         return self._reference.id
    595 
    596     @property

AttributeError: 'NoneType' object has no attribute 'id'
```

pchauhan-qlogic:
Fix #7564

### DIFF
--- a/firestore/google/cloud/firestore_v1/document.py
+++ b/firestore/google/cloud/firestore_v1/document.py
@@ -603,11 +603,9 @@ class DocumentSnapshot(object):
         """The document identifier (within its collection).
 
         Returns:
-            str: The last component of the path of the document if it exists.
+            str: The last component of the path of the document.
         """
-        if self._reference:
-            return self._reference.id
-        return None
+        return self._reference.id
 
     @property
     def reference(self):
--- a/firestore/google/cloud/firestore_v1/document.py
+++ b/firestore/google/cloud/firestore_v1/document.py
@@ -603,7 +603,7 @@ class DocumentSnapshot(object):
         """The document identifier (within its collection).
 
         Returns:
-            str: The last component of the path of the document if document exist..
+            str: The last component of the path of the document if it exists.
         """
         if self._reference:
             return self._reference.id
--- a/firestore/google/cloud/firestore_v1/document.py
+++ b/firestore/google/cloud/firestore_v1/document.py
@@ -603,10 +603,11 @@ class DocumentSnapshot(object):
         """The document identifier (within its collection).
 
         Returns:
-            str: The last component of the path of the document.
+            str: The last component of the path of the document if document exist..
         """
         if self._reference:
             return self._reference.id
+        return None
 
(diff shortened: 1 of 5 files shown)


firestore/google/cloud/firestore_v1/document.py
```diff
@@ -603,9 +603,11 @@ def id(self):
         """The document identifier (within its collection).
 
         Returns:
-            str: The last component of the path of the document.
+            str: The last component of the path of the document if document exist..
```
sduskis: s/if document exist../if it exists./

pchauhan-qlogic: changed

tseaver:
Fixing #7564 would require changing `google.cloud.firestore_v1.client._parse_batch_get` to parse out the values necessary to construct a reference from the `missing` value, as specified in the [API reference](google.firestore.v1.BatchGetDocumentsResponse.FIELDS.string.google.firestore.v1.BatchGetDocumentsResponse).

We would like to have as an invariant that a `DocumentSnapshot` returned from our APIs *never* has a `_reference` attribute set to `None`.

pchauhan-qlogic:
I am closing this pr there is more redudant code

PR #8303, opened 2019-06-13
Refactor `to_dataframe` deterministicly update progress bar.

### LINKED ISSUE #8175: BigQuery: 'test_to_dataframe_w_bqstorage_nonempty' unit test flakes
From this [failed Kokoro job](https://source.cloud.google.com/results/invocations/7094b690-9af4-4159-922f-94d4b2c20da8/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigquery/log):

```python
____________ TestRowIterator.test_to_dataframe_w_bqstorage_nonempty ____________
self = <tests.unit.test_table.TestRowIterator testMethod=test_to_dataframe_w_bqstorage_nonempty>
    @unittest.skipIf(pandas is None, "Requires `pandas`")
    @unittest.skipIf(
        bigquery_storage_v1beta1 is None, "Requires `google-cloud-bigquery-storage`"
    )
    def test_to_dataframe_w_bqstorage_nonempty(self):
        from google.cloud.bigquery import schema
        from google.cloud.bigquery import table as mut
        from google.cloud.bigquery_storage_v1beta1 import reader
        # Speed up testing.
        mut._PROGRESS_INTERVAL = 0.01
        bqstorage_client = mock.create_autospec(
            bigquery_storage_v1beta1.BigQueryStorageClient
        )
        streams = [
            # Use two streams we want to check frames are read from each stream.
            {"name": "/projects/proj/dataset/dset/tables/tbl/streams/1234"},
            {"name": "/projects/proj/dataset/dset/tables/tbl/streams/5678"},
        ]
        session = bigquery_storage_v1beta1.types.ReadSession(streams=streams)
        session.avro_schema.schema = json.dumps(
            {
                "fields": [
                    {"name": "colA"},
                    # Not alphabetical to test column order.
                    {"name": "colC"},
                    {"name": "colB"},
                ]
            }
        )
        bqstorage_client.create_read_session.return_value = session
        mock_rowstream = mock.create_autospec(reader.ReadRowsStream)
        bqstorage_client.read_rows.return_value = mock_rowstream
        mock_rows = mock.create_autospec(reader.ReadRowsIterable)
        mock_rowstream.rows.return_value = mock_rows
        page_items = [
            {"colA": 1, "colB": "abc", "colC": 2.0},
            {"colA": -1, "colB": "def", "colC": 4.0},
        ]
        def blocking_to_dataframe(*args, **kwargs):
            # Sleep for longer than the waiting interval so that we know we're
            # only reading one page per loop at most.
            time.sleep(2 * mut._PROGRESS_INTERVAL)
            return pandas.DataFrame(page_items, columns=["colA", "colB", "colC"])
        mock_page = mock.create_autospec(reader.ReadRowsPage)
        mock_page.to_dataframe.side_effect = blocking_to_dataframe
        mock_pages = (mock_page, mock_page, mock_page)
        type(mock_rows).pages = mock.PropertyMock(return_value=mock_pages)
        # Test that full queue errors are ignored.
        mock_queue = mock.create_autospec(mut._NoopProgressBarQueue)
        mock_queue().put_nowait.side_effect = queue.Full
        schema = [
            schema.SchemaField("colA", "IGNORED"),
            schema.SchemaField("colC", "IGNORED"),
            schema.SchemaField("colB", "IGNORED"),
        ]
        row_iterator = mut.RowIterator(
            _mock_client(),
            None,  # api_request: ignored
            None,  # path: ignored
            schema,
            table=mut.TableReference.from_string("proj.dset.tbl"),
            selected_fields=schema,
        )
        with mock.patch.object(mut, "_NoopProgressBarQueue", mock_queue), mock.patch(
            "concurrent.futures.wait", wraps=concurrent.futures.wait
        ) as mock_wait:
            got = row_iterator.to_dataframe(bqstorage_client=bqstorage_client)
        # Are the columns in the expected order?
        column_names = ["colA", "colC", "colB"]
        self.assertEqual(list(got), column_names)
        # Have expected number of rows?
        total_pages = len(streams) * len(mock_pages)
        total_rows = len(page_items) * total_pages
        self.assertEqual(len(got.index), total_rows)
        # Make sure that this test looped through multiple progress intervals.
        self.assertGreaterEqual(mock_wait.call_count, 2)
        # Make sure that this test pushed to the progress queue.
>       self.assertEqual(mock_queue().put_nowait.call_count, total_pages)
E AssertionError: 5 != 6
```

tswast:
Previously, a background thread was used to collect progress bar updates
from worker threads. So as not to block downloads for progress bar
updates, `put_nowait` was used to make progress bar updates. Missed
writes to the progress bar were ignored. This caused non-deterministic
progress bar updates and test flakiness.

Now, worker threads push dataframes to the queue, and the return values
for `download_dataframe_bqstorage` and
`download_dataframe_tabledata_list` have been updated to return an
iterable of pandas DataFrame objects instead of a single DataFrame. This
allows progress bar updates to be done independently of which underlying
API is used to download the DataFrames.

Also, the logic for working with pandas has been moved to the
`_pandas_helpers` module.

Closes #8175


bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -176,5 +207,163 @@ def to_parquet(dataframe, bq_schema, filepath):
...
+                except queue.Empty:  # pragma: NO COVER
+                    continue
+
+            # Return any remaining values after the workers finished.
+            while not worker_queue.empty():
+                try:
+                    # Include a timeout because even though the queue is
+                    # non-empty, it doesn't guarantee that a subsequent call to
+                    # get() will not block.
+                    frame = worker_queue.get(timeout=_PROGRESS_INTERVAL)
```
plamut: Since the only place that _takes_ the items from the queue is here (the workers only maybe-put them in the queue), what's the case of the `get()` call blocking? Isn't the queue always non-empty at this point, meaning that `get()` will never block (or raise `Empty` if timeout is used)?

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -176,5 +207,163 @@ def to_parquet(dataframe, bq_schema, filepath):
...
+                    future.result()
+
+                try:
+                    frame = worker_queue.get(timeout=_PROGRESS_INTERVAL)
+                    yield frame
+                except queue.Empty:  # pragma: NO COVER
+                    continue
+
+            # Return any remaining values after the workers finished.
+            while not worker_queue.empty():
```
plamut: Just trying to understand the possible scenarios, in what cases could there be leftovers in the queue that we would have to clean up here?

If a `_download_dataframe_bqstorage_stream()` future completes and ends up in the `done` list, it means that one of the following has happened when we reach this code:
- A frame was put in the queue and the function returned normally. In this case we would have already taken the frame from the queue and yielded it a few lines above.
Unless ...
- ...an exception occurred in one of the `done` workers. The `future.result()` would propagate that exception, we would jump to the `finally` block, and this queue-clearing loop would not be executed.

I might be missing something, thus mostly bringing this up as a sanity check.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -176,5 +207,163 @@ def to_parquet(dataframe, bq_schema, filepath):
...
+                    dtypes,
+                    session,
+                    stream,
+                    worker_queue,
+                )
+                for stream in session.streams
+            ]
+
+            while not_done:
+                done, not_done = _nowait(not_done)
```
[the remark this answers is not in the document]
tswast: There are two things that I'm waiting for: the worker queue (all the DataFrames as they are processed from each thread) and for the threads to complete. I want the worker queue to take precedence.

If the queue fills up, it can block processing because the threads will block until an item is pulled from the queue. I actually encountered this behavior when I first implemented this code, as it manifested as much slower than expected download speeds. See the following picture:

wait for queue vs  wait for futures (1)

Notice that most of the `put` calls block because of a full queue.

One way to address the full queue is to have an inner loop to drain the queue each time. I actually tried this, too, but it resulted in bursty reads. Many `put` calls still would block.

wait for queue vs  wait for futures (2)

What I've implemented instead is to not ever wait for the threads to finish, and instead block on the queue.

wait for queue vs  wait for futures

This could result in waiting longer than we need to at the end, since we could wait for a full `_PROGRESS_INTERVAL` before noticing that all streams are finished. Since these streams are expected to last quite long compared to the `_PROGRESS_INTERVAL`, I'm not as concerned about that. My main motivation is to ensure that `put` calls do not block.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -176,5 +207,163 @@ def to_parquet(dataframe, bq_schema, filepath):
...
+    session,
+    stream,
+    worker_queue,
+):
+    position = bigquery_storage_v1beta1.types.StreamPosition(stream=stream)
+    rowstream = bqstorage_client.read_rows(position).rows(session)
+
+    for page in rowstream.pages:
+        if download_state.done:
+            return
```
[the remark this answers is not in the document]
tswast: There are two ways that `download_state.done` could be set:

1. All streams have completed and been processed. (This branch won't be hit in that case, as the thread is already exited.)
2. One of the worker threads has thrown an exception (or KeyboardInterrupt or something was thrown in the main thread). In this case, we don't care at all about in-flight frames because everything is getting thrown out anyway.

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: On the very last iteration of the "wait for all worker threads to be done" loop, it's possible that we pulled everything from the queue but between when we `yield`ed the last item and we checked for the workers to be done a worker wrote to the queue.

* reader: get
* reader: yield
* (function using these results does stuff)
* worker: put
* worker: done
* reader: all done!
* reader: oh wait, there's still something on the queue

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: Yes, this is the only thread that reads from the queue. So, since the queue is non-empty, this call should never block.

I'm being conservative here, since the [Python docs for Queue.empty()](https://docs.python.org/3/library/queue.html#queue.Queue.empty) aren't specific about in what cases we can expect `get` to not block.

plamut: I checked the [source](https://github.com/python/cpython/blob/3.7/Lib/queue.py#L169-L170), it only blocks when the queue is empty. But I'm also fine with conservative approach in general, since an incorrect assumption can result in a much more severe bug than having a few extra lines of code.

bigquery/google/cloud/bigquery/_pandas_helpers.py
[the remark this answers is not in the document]
plamut: Great, thanks, makes perfect sense!

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -331,6 +331,12 @@ def download_dataframe_bqstorage(
             ]
 
             while not_done:
+                # Don't block on the worker threads. For performance reasons,
+                # we want to block on the queue's get method, instead. This
+                # prevents the queue from filling up, because the main thread
+                # has smaller gaps in time between calls to the queue's get
+                # method. For a detailed explaination, see:
+                # https://friendliness.dev/2019/06/18/python-nowait/
```
plamut: This is great. :+1:

33558903:
#8303

Issue #7563, opened 2019-03-27
IntEnum as a value type


onlined:
**For Datastore:**
I think that instances of `IntEnum` can be serialized as their integer values. Its implementation is easy, and it would allow to use enumeration types easily.

tseaver:
@onlined Can you show an example of how you would use it?  If we convert an instance of `enum.IntEnum` to an int when storing it into an entity, that conversion will be lossy:  we won't get an `IntEnum` back after reading the entity later.

onlined:
The example:
```py
class EmployeeType(IntEnum):
	PART_TIME = 0
	FULL_TIME = 1
	REMOTE = 2

# employee_entity is an instance of Entity
employee_entity['type'] = EmployeeType.PART_TIME.value # I don't want to enter .value
employee_entity.put()
```

tseaver:
@onlined OK, but what do you expect to happen when fetching the entity later via `get()`?  If we convert the enum to its integer value implicitly, then you will just get the integer back, not the `EmployeeType.PART_TIME` enum.

onlined:
@tseaver Yes, but it would be comparable with the enum, and I think that's the most common case of using an enum.
```py
if employee_entity['type'] == EmployeeType.PART_TIME:
	# Do something here
	pass
```

PR #8304, opened 2019-06-14
Datastore: allow IntEnum as a value type

IlyaFaer:
Closes: #7563


datastore/google/cloud/datastore/entity.py
```diff
@@ -173,6 +173,11 @@ def __ne__(self, other):
         """
         return not self == other
 
+    def __setitem__(self, key, value):
+        if isinstance(value, enum.Enum):
+            value = value.value
+        return super(Entity, self).__setitem__(key, value)
```
IlyaFaer: If-statement for enums

datastore/tests/unit/test_entity.py
```diff
@@ -221,6 +222,18 @@ def test___repr___w_key_non_empty(self):
         expected = "<Entity%s %s>" % (flat_path, entity_vals)
         self.assertEqual(repr(entity), expected)
 
+    def test__set_item_enum(self):
+        """Test if given Enum converts to int."""
+        entity = self._make_one(key=_Key())
+        entity['some_value'] = _IntEnum.VALUE_1
+
+        self.assertEqual(type(entity['some_value']), int)
```
IlyaFaer: Type check needed, 'cause enum and int are comparable, so second assert will be True no matter enum or int stored in Entity

mf2199:
@tseaver @crwilcox @frankyn A friendly reminder: This PR has been hanging out for awhile. We could use some feedback regarding its status.

frankyn:
Summoning @benwhitehead our datastore/firestore colleague.

BenWhitehead:
I am not personally familiar with enums in Python, so I'll have to defer to @crwilcox on this one.

datastore/tests/unit/test_entity.py
```diff
@@ -221,6 +222,18 @@ def test___repr___w_key_non_empty(self):
         expected = "<Entity%s %s>" % (flat_path, entity_vals)
         self.assertEqual(repr(entity), expected)
 
+    def test__set_item_enum(self):
+        """Test if given Enum converts to int."""
+        entity = self._make_one(key=_Key())
+        entity["some_value"] = _IntEnum.VALUE_1
+
+        self.assertEqual(type(entity["some_value"]), int)
+        self.assertEqual(entity["some_value"], _IntEnum.VALUE_1.value)
```
crwilcox: This test doesn't currently reflect how I expect this to be used. I Expect most users will be doing

```suggestion
        self.assertEqual(entity["some_value"], _IntEnum.VALUE_1)
        self.assertEqual(entity["some_value"], _IntEnum.VALUE_1.value)
```

Can this test be changed to assert the pattern we expect to be used?

IlyaFaer: Sure, I've tried it minute ago, test was success

datastore/google/cloud/datastore/entity.py
```diff
@@ -173,6 +173,11 @@ def __ne__(self, other):
         """
         return not self == other
 
+    def __setitem__(self, key, value):
```
crwilcox: It seems odd having this one change from the default serialization into the dict. We have a `def entity_to_protobuf(entity):` method. Is this the correct place to do this?

IlyaFaer: I think, in issue they were speaking only about converting fields of entity, like `employee_entity['type'] == EmployeeType.PART_TIME`, while `entity_to_protobuf` converts the whole entity

crwilcox: https://github.com/googleapis/google-cloud-python/blob/master/datastore/google/cloud/datastore/helpers.py#L216 already walks the entity to convert types. Could this be done there?

IlyaFaer: [setitem](https://docs.python.org/3/reference/datamodel.html#object.__setitem__) will be called every time user will try to put value into field, like
`entity['field'] = some_enum`
And enum's `int` representation will be stored. So `entity_to_protobuf` don't really need any changes, 'cause values will already be ints when you'll call `entity_to_protobuf`

But if we'll move this changes to `entity_to_protobuf`, user will be storing pure enums into entity fields, and if he'll try to get field value like:
`entity['field']`
He'll got a enum, instead of `int`. And he wants to get `int` in a first place

BenWhitehead:
@crwilcox Can you take another pass at this?

IlyaFaer:
A friendly ping

PR #8375, opened 2019-06-17
Storage: add 'Client.list_blobs(bucket_or_name)'.

IlyaFaer:
Made item from redesign doc: "New method: client.list_blobs(bucket_or_name), which functions the same as Bucket.list_blobs()"
https://github.com/googleapis/google-cloud-python/issues/7762


storage/google/cloud/storage/client.py
```diff
@@ -153,6 +153,26 @@ def _pop_batch(self):
         """
         return self._batch_stack.pop()
 
+    def _use_or_init_bucket(self, bucket_or_name):
```
IlyaFaer: Made a helper - for now it's used in 3 places

storage/tests/unit/test_bucket.py
```diff
@@ -782,19 +782,6 @@ def test_list_blobs_w_all_arguments_and_user_project(self):
...
-        connection = _Connection({"items": []})
-        client = _Client(connection)
-        bucket = self._make_one(client=client, name=NAME)
-        iterator = bucket.list_blobs()
-        blobs = list(iterator)
-        self.assertEqual(blobs, [])
-        kw, = connection._requested
-        self.assertEqual(kw["method"], "GET")
-        self.assertEqual(kw["path"], "/b/%s/o" % NAME)
-        self.assertEqual(kw["query_params"], {"projection": "noAcl"})
```
IlyaFaer: This is full copy of another test named test_list_blobs_defaults

storage/tests/unit/test_client.py
```diff
@@ -821,3 +896,39 @@ def dummy_response():
         self.assertEqual(page.remaining, 0)
         self.assertIsInstance(bucket, Bucket)
         self.assertEqual(bucket.name, blob_name)
+
+
+class _Connection(object):
```
IlyaFaer: Analogy to class from bucket tests

storage/google/cloud/storage/bucket.py
```diff
@@ -770,6 +770,11 @@ def list_blobs(
         :returns: Iterator of all :class:`~google.cloud.storage.blob.Blob`
                   in this bucket matching the arguments.
         """
+        warnings.warn(
+            "Bucket.list_blobs method is deprecated. Use Client.list_blobs instead.",
+            PendingDeprecationWarning,
+            stacklevel=2,
+        )
```
IlyaFaer: Not sure about this. It will warn if you call Client.list_blobs too

tswast: Let's wait to mark these as deprecated until we move all the actual implementation out of the Blob and Bucket classes into the Client class.

storage/google/cloud/storage/client.py
tswast: Nit on naming: how about `_bucket_arg_to_bucket` to align with the similar helper in BigQuery?

https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/google/cloud/bigquery/table.py#L1771

Also, this might be well suited for the `bucket.py` module, similar to BigQuery, as I expect the `client.py` file will expand and the `bucket.py` file will shrink as we move implementation over to get ready for deprecation.

storage/google/cloud/storage/client.py
```diff
@@ -394,6 +404,82 @@ def download_blob_to_file(self, blob_or_uri, file_obj, start=None, end=None):
...
+
+        If :attr:`user_project` is set, bills the API request to that project.
+
+        bucket_or_name (Union[ \
+            :class:`~google.cloud.storage.bucket.Bucket`, \
+             str, \
+        ]):
+            The bucket resource to pass or name to create.
+
+        :type max_results: int
```
tswast: Use Google-style docstrings. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

storage/tests/unit/test_client.py
tswast: I'd prefer that we use mocks rather than add this class.

See:

https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/tests/unit/helpers.py#L16-L24

and how it is used:

https://github.com/googleapis/google-cloud-python/blob/0e231cab5f066ae5b3e254d0cb7d204e3dac7ebc/bigquery/tests/unit/test_client.py#L169-L173

storage/google/cloud/storage/client.py
IlyaFaer: No problem, renamed

storage/tests/unit/test_client.py
IlyaFaer: Redone with similar mock

storage/google/cloud/storage/client.py
IlyaFaer: Okey, it's done

storage/google/cloud/storage/bucket.py
```diff
@@ -721,6 +721,9 @@ def list_blobs(
     ):
         """Return an iterator used to find blobs in the bucket.
 
+        .. note::
+          Direct use of this method is deprecated. Use Client.list_blobs instead.
```
tswast: Nit. Add backticks for sphinx to render as code font.

```
Use ``Client.list_blobs`` instead.
```

PR #8380, opened 2019-06-17
IoT: Add docs publish job.

busunkim96:


.kokoro/publish-docs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+cd github/google-cloud-python/${PACKAGE}
+
+VERSION=$(python3 ../test_utils/scripts/get_package_version.py)
```
tseaver: ISTM this would be much better as:

```bash
$ python3 setup.py --version
```

busunkim96: Thank you!

PR #8381, opened 2019-06-18
Container Analysis: Add get_grafeas_client method.

busunkim96:
This allows for someone to get a Grafeas Client from a Container Analysis Client.


containeranalysis/setup.py
```diff
@@ -20,61 +20,60 @@
...
+version = "0.1.0"
+release_status = "3 - Alpha"
 dependencies = [
-  'google-api-core[grpc] >= 1.4.1, < 2.0.0dev',
-  'grpc-google-iam-v1 >= 0.11.1, < 0.12dev',
-  'enum34; python_version < "3.4"',
+    "google-api-core[grpc] >= 1.4.1, < 2.0.0dev",
+    "grpc-google-iam-v1 >= 0.11.1, < 0.12dev",
+    'enum34; python_version < "3.4"',
+    "grafeas",
```
tseaver: Is `grafeas` supposed to be an unconditional dependency for `google-cloud-containeranalysis`?

containeranalysis/google/cloud/devtools/containeranalysis_v1/gapic/container_analysis_client.py
```diff
@@ -39,6 +39,9 @@
 from google.iam.v1 import policy_pb2
 
 
+from grafeas import grafeas_v1
+from grafeas.grafeas_v1.gapic.transports import grafeas_grpc_transport
+
```
tseaver: If `grafeas` is *not* supposed to be an unconditional dependency, then these two imports need to be wrapped:

```python
try:
    from grafeas import grafeas_v1
    from grafeas.grafeas_v1.gapic.transports import grafeas_grpc_transport
except ImportError:
    grafeas_v1 = None
    grafeas_grpc_transport = None
```

containeranalysis/setup.py
busunkim96: Yes, we want it to be unconditional

PR #8415, opened 2019-06-18
Release pubsub 0.42.0

sduskis:


busunkim96:
I don't if something is up with releasetool, but could you also bump the version in the setup.py?

sduskis:
@busunkim96, thanks for the catch.  Updated.

PR #8422, opened 2019-06-19
BigQuery: Add sample demontrating how to create a job.

tswast:
This sample is intended for https://cloud.google.com/bigquery/docs/running-jobs

It demonstrates advanced features, such as overriding location
autodetection and overriding job ID generation.

cc @grant @shollyman @alixhami


alixhami:
I think it makes more sense to have the documentation about the location in the docs, rather than copied into a comment in each language's snippet.

I also don't think it's a good idea to put URLs in comments. It can become a problem if pages are moved or if this sample ends up being included on the page it's linking to.

tswast:
Removed the link and the comment about location autodetection. I'll make sure that is documented when I add the sample to the docs.

PR #8423, opened 2019-06-19
Storage: Add generation to 'Blob.__repr__'.

### LINKED ISSUE #7789: Storage: include generation in Blob repr.
I added a comment here based on the `generation` addition to `storage` https://github.com/googleapis/google-cloud-python/pull/7444#issuecomment-485874643

I think it's worth adding an issue too - specifically because now `generation` defines a blob but isn't included in the repr. Meaning you can get confusing results like this:

```python

blob
<Blob: sixty-capital-temp, tests/test_357256/blob_645776>

b2 = gcs.client(project).get_bucket(bucket).get_blob(blob.name)  # recreate 
b2
<Blob: sixty-capital-temp, tests/test_357256/blob_645776>  # same repr

blob.exists()
False

b2.exists()
True
```
... but b2 & blob have the same repr! Prior to the latest release, both results would have been `True`. 
If generation were in the repr, this would be less confusing - e.g. I wouldn't have had to read the release notes to work out what was going on

HemangChothani:
Fixes #7789

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -238,7 +238,7 @@ class Blob(_PropertyMixin):
         else:
             bucket_name = None
 
-        return "<Blob: %s, %s>" % (bucket_name, self.name)
+        return "<Blob: %s, %s, %s>" % (bucket_name, self.name, self.generation)
 
     @property
     def path(self):


tseaver:
The [Bigtable snippet / systest failures](https://source.cloud.google.com/results/invocations/f7c0b7b2-ca88-403b-acb6-afebd7a1ca57/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable) are clearly unrelated.

PR #8424, opened 2019-06-19
Storage: Add `Blob.from_string` and `Bucket.from_string`.

### LINKED ISSUE #7693: Storage: Add method(s) to parse `gs://bucket_name` and `gs://bucket_name/path/to/blob` URIs
It would be very convenient if we could take a string such as `gs://bucket_name/path/to/blob` and download the content for it.

I propose we add a `Blob.from_string` classmethod, similar to BigQuery's [`TableReference.from_string` method](https://googleapis.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.table.TableReference.html#google.cloud.bigquery.table.TableReference.from_string).

Since the `Blob` class requires a client object (via Bucket) to support downloads, I suggest both `path` and `client` are required arguments for `Blob.from_string(path, client)`.

HemangChothani:
Fixes #7693


storage/google/cloud/storage/client.py
```diff
@@ -477,6 +477,37 @@ def list_buckets(
             extra_params=extra_params,
         )
 
+    def get_blob_from_uri(self, uri):
```
tseaver: Please move this method below `get_bucket_from_uri`.

storage/google/cloud/storage/client.py
```diff
@@ -477,6 +477,37 @@ def list_buckets(
...
+        """Get a constructor for blob object by URI.
+
+         :type uri: str
+         :param uri: The blob uri pass to get blob object.
```
tseaver: Please add an example of the expected form of the uri, e.g.:

```python
        """Construct a blob object from a URI.

         :type uri: str
         :param uri:
             The blob's URI, in the form
             "gs://<bucket-name>/<blob-name>".
        ...
        """

File: storage/google/cloud/storage/client.py [hunk 3]
```diff
@@ -477,6 +477,37 @@ def list_buckets(
...
+            raise ValueError("URI scheme must be gs")
+
+        bucket = Bucket(self, name=netloc)
+        return Blob(path[1:], bucket)
+
+    def get_bucket_from_uri(self, uri):
+        """Get a constructor for bucket object by URI.
+
+        :type uri: str
+        :param uri: The bucket uri pass to get bucket object.
```
tseaver: Please add an example of the expected form of the uri, e.g.:

```python
        """Construct a bucket from a URI.

         :type uri: str
         :param uri:
             The bucket's URI, in the form
             "gs://<bucket-name>".
        ...
        """

File: storage/google/cloud/storage/client.py [hunk 4]
```diff
@@ -477,6 +477,37 @@ def list_buckets(
...
+
+         :rtype: :class:`google.cloud.storage.blob.Blob`
+         :returns: The blob object created.
+         """
+        scheme, netloc, path, query, frag = urlsplit(uri)
+        if scheme != "gs":
+            raise ValueError("URI scheme must be gs")
+
+        bucket = Bucket(self, name=netloc)
+        return Blob(path[1:], bucket)
```
tseaver: @frankyn, @tswast  Given that the names for these methods start with `get_`, should we be making the API calls to load the objects?

tswast: Yes. `get_` implies actually making a GET request. This should call `.reload()`.

tseaver:
@tswast What is your expected usecase for having `client` default to `None` in a `Blob.from_string` / `Bucket.from_string` call?  The objects themselves aren't very useful without a client.

tswast:
> What is your expected usecase for having client default to None in a Blob.from_string / Bucket.from_string call? The objects themselves aren't very useful without a client.

Per https://github.com/googleapis/google-cloud-python/issues/7762, the goal is to have `Blob` and `Bucket` usable without a `client` in more places, as methods like `client.create_bucket` should be able to take a `Bucket` object without a `client` object present. (I acknowledge that this isn't the case yet.)

frankyn:
@tswast, @HemangChothani has updated the PR to move get_*_from_uri to their respective Blob or Bucket classes. 

I thought it was okay to move forward with adding them at the client level. If moving forward w/ Bucket and Blob methods instead, documentation should be updated to clarify the need to import Bucket and Blob explicitly.

storage/google/cloud/storage/blob.py
```diff
@@ -301,6 +301,37 @@ def public_url(self):
...
+            Get a constructor for blob object by URI..
+
+            >>> from google.cloud import storage
+            >>> from google.cloud.storage.blob import Blob
+            >>> client = storage.Client()
+            >>> blob = Blob.from_string("gs://bucket/object")
+         """
+        from google.cloud.storage.bucket import Bucket
+
+        scheme, netloc, path, query, frag = urlsplit(uri)
```
tswast: We need to update `download_blob_to_file` to use this `from_string` method.

https://github.com/googleapis/google-cloud-python/blob/6a06a5f3b99ccd2f7f5f003d2da08b501bdf55d2/storage/google/cloud/storage/client.py#L400

storage/tests/unit/test_bucket.py
```diff
@@ -2651,6 +2652,29 @@ def _generate_signed_url_helper(
         }
         signer.assert_called_once_with(expected_creds, **expected_kwargs)
 
+    def test_get_bucket_from_uri_w_valid_uri(self):
```
tswast: Let's update the test names to reflect that they are testing `from_string` now.

HemangChothani:
@googlebot I fixed it..

HemangChothani:
@tswast  PTAL

HemangChothani:
@tswast Don't know why unit test fails, even failed in master branch

HemangChothani:
Closing this PR due to something irrelevant happens and unable to fix it, So replacing this PR with new PR [9143](https://github.com/googleapis/google-cloud-python/pull/9143)

Issue #7387, opened 2019-02-20
PubSub 0.39.1 Observed non-recoverable stream error 403 User not authorized to perform...


vincemartin:
OS - Windows Server
Python version - Python 2.7.15
Pubsub version - google-cloud-pubsub==0.39.1

Code : 

def callback(message):
    logging.info('Received message {}'.format(message))
    try:
        filename = message.data
        global queue
        queue.put_nowait(filename)

        global waiting_event
        waiting_event.set()
    except Exception as e:
        logging.error("Callback : {}".format(e))
    finally:
        message.ack()

def Run(updateDatabase_Function, topic, timeout=60.0 * 60.0):
    global queue
    queue = Queue()
    global waiting_event
    waiting_event = threading.Event()
    waiting_event.clear()

    credentials = GetCredentials()
    subscriber = pubsub_v1.SubscriberClient(credentials=credentials)
    subscription_path = subscriber.subscription_path(credentials.project_id, source)
    subscriber.subscribe(subscription_path, callback=callback)

After several days of good behaviour, the client sent the following messages  :
2019-02-09 01:02:22,566 INFO : Observed non-recoverable stream error 403 User not authorized to perform this action.
2019-02-09 01:02:22,566 INFO : RPC termination has signaled streaming pull manager shutdown.
2019-02-09 01:02:22,566 INFO : Observed non-recoverable stream error 403 User not authorized to perform this action.
2019-02-09 01:02:22,566 INFO : Thread-ConsumeBidirectionalStream exiting
2019-02-09 01:02:22,628 INFO : Thread-LeaseMaintainer exiting.
2019-02-09 01:02:22,676 INFO : Thread-Heartbeater exiting.

tseaver:
@vincemartin Is it possible that somebody removed permissions from the service account returned by your `GetCredentials`  call?  For instance, by using the Cloud Console?

@jadekler Can you suggest another reason why a working pubsub subscription stream would abort with a 403?

jadekler:
I'm no auth expert, but could maybe be some token-refreshing problem? cc @broady who knows a lot more about auth than I do.

vincemartin:
@tseaver : no, I don't think so. We re-launched the python script and it works again. We are just afraid to have this error coming again randomly...

sduskis:
[...]

PR #8429, opened 2019-06-19
PubSub: Document how to choose the PubSub auth method

### LINKED ISSUE #8137: Pubsub: add 'PublisherClient.from_service_account_dict'
I'd like to use the pubsub client on a webapp where i have the secrets in environment variables and not on disk.
I would like to have a classmethod added to Publisher to instantiate a client with the credentials passed from a dict instead of a json file path.

plamut:
Closes #8137.

Describes a possible workaround for #7387, however, it does not close it, the PR just fulfills a GA  requirement (documenting a possible workaround).

### How to test
**Steps to perform:**

- Generate the pubsub docs locally (from inside the `pubsub/` directory):
    ```
    $ nox -f noxfile.py -s docs
    ```
- Open the generated docs for authentication:
    ```
    $ google-chrome "file://$GOOG_CLOUD_PYTHON_DIR/pubsub/docs/_build/html/index.html#authentication"
    ```

**Expected result:**
The example is understandable and correct, the links point to correct locations.

### Background
The issue itself is hard to reproduce in practice, but it was discovered in other PubSub clients that using JWT-based authentication avoids the issue. The goal of this PR is to document how different auth methods can be used with the Python client.

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -164,8 +164,8 @@ Authentication
 ^^^^^^^^^^^^^^
 
 It is possible to specify the authentication method to use with the Pub/Sub
-clients. This can be done by providing an explicit `Credentials`_ instance from
-the `google-auth`_ library.
+clients. This can be done by providing an explicit `Credentials`_ instance. Support
+for various authentication methods is available from the `google-auth`_ library.
 
 For example, to use JSON Web Tokens, provide a `jwt.Credentials`_ instance:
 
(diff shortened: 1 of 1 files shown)


pubsub/README.rst
```diff
@@ -158,3 +158,35 @@ block the current thread until a given condition obtains:
 To learn more, consult the `subscriber documentation`_.
 
 .. _subscriber documentation: https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/index.html
+
+
+Authentication
+^^^^^^^^^^^^^^
+
+It is possible to specify the authentication method to use with the Pub/Sub
+clients. This can be done by providing an explicit `Credentials`_ instance from
```
xrmx: It would be nice to add a bit more emphasis on the google auth link, e.g.: 

```
This can be done by providing an explicit `Credentials`_ instance. Support for various authentication methods is available from the `google-auth` library.
```

plamut: Sure thing, will add right away. Thanks for the feedback!

plamut:
I marked the PR to close #8137 based on the issue reporter's feedback (thanks again, @xrmx ).

PR #8431, opened 2019-06-19
Release containeranalysis 0.1.0

busunkim96:
This pull request was generated using releasetool.

PR #8432, opened 2019-06-19
Release containeranalaysis 0.1.0

busunkim96:
Bad setup.py resulted in a failed release #8431. 

This is take 2.

### DIFF
--- a/containeranalysis/setup.py
+++ b/containeranalysis/setup.py
@@ -23,7 +23,7 @@ import setuptools
 name = "google-cloud-containeranalysis"
 description = "Container Analysis API API client library"
 version = "0.1.0"
-release_status = "3 - Alpha"
+release_status = "Development Status :: 3 - Alpha"
 dependencies = [
     "google-api-core[grpc] >= 1.4.1, < 2.0.0dev",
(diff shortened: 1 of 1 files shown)


containeranalysis/setup.py
```diff
@@ -55,7 +55,7 @@
     author="Google LLC",
     author_email="googleapis-packages@google.com",
     license="Apache 2.0",
-    url="https://github.com/googleapis/google-cloud-python",
+    url="https://github.com/GoogleCloudPlatform/google-cloud-python",
```
tseaver: Revert this bit.

containeranalysis/setup.py
```diff
@@ -23,7 +23,7 @@
 name = "google-cloud-containeranalysis"
 description = "Container Analysis API API client library"
 version = "0.1.0"
-release_status = "3 - Alpha"
+release_status = "Development Status :: 3 - Alpha"
```
tseaver: Do we need to fix a template somewhere for this error?

busunkim96: Looks like it, I think the one in GAPIC Generator always sets the release_status to a bad classifier.

PR #8200, opened 2019-05-30
Storage: Make signurl work with objects containing tildes.

MewX:
Fixes https://issuetracker.google.com/issues/133359302

Similar to this one: https://github.com/GoogleCloudPlatform/gsutil/pull/783

PR #8434, opened 2019-06-19
Storage: Avoid escaping tilde in blob public / signed URLs.

tseaver:
Supersedes #8200.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -298,7 +298,7 @@ class Blob(_PropertyMixin):
         return "{storage_base_url}/{bucket_name}/{quoted_name}".format(
             storage_base_url=_API_ACCESS_ENDPOINT,
             bucket_name=self.bucket.name,
-            quoted_name=quote(self.name.encode("utf-8")),
+            quoted_name=_quote(self.name, safe=b"/~"),
         )
 
     def generate_signed_url(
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -1983,7 +1984,7 @@ def _get_encryption_headers(key, source=False):
     }
 
 
-def _quote(value):
+def _quote(value, safe=b"~"):
```
frankyn: Why not include `/` as well?

storage/google/cloud/storage/blob.py
```diff
@@ -1994,11 +1995,14 @@ def _quote(value):
     :type value: str or bytes
     :param value: The value to be URL-quoted.
 
+    :type safe: bytes
+    :param safe: Bytes *not* to be quoted.
```
frankyn: By default includes "~".

storage/google/cloud/storage/blob.py
tseaver: @frankyn Because it is not always wanted:  when quoting for the JSON API, we want to percent-escape slashes.  Only for the XML API (`public_url`, `generate_*_signed_url`) do we want to avoid escaping slashes.

storage/google/cloud/storage/blob.py
tseaver: 61c3990

storage/google/cloud/storage/blob.py
frankyn: Thanks for clarifying

PR #7645, opened 2019-04-02
Text-to-Speech: Remove empty 'api.rst' from docs

busunkim96:
Fixes #7629.

PR #8435, opened 2019-06-19
Text-to-Speech: Repair top-level API reference page.

### LINKED ISSUE #7629: Empty API page in "Python Client for Cloud Text-to-Speech API"
Hello,

I readed a document for "Python Client for Cloud Text-to-Speech API" and found a links "APIs" and "API Reference"
Screenshot 2019-04-01 at 12 42 51

They point to 
https://googleapis.github.io/google-cloud-python/latest/texttospeech/api.html
but it's not have content.
Screenshot 2019-04-01 at 12 45 49

I think that this page should be deleted.
.
Greetings

tseaver:
Supersedes #7645.

Fixes #7629.

### DIFF
--- a/texttospeech/docs/api.rst
+++ b/texttospeech/docs/api.rst
@@ -1,31 +1,30 @@
-API Reference
-=============
+Text-to-Speech Client API Reference
+===================================
 
-APIs
-----
+This package includes clients for multiple versions of the Text-to-Speech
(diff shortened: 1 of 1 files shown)


tseaver:
Docs-only change.  I have verified locally that the new `api.rst` page builds as expected.

Issue #7829, opened 2019-04-30
Firestore: 'test_collection_add' flakes


tseaver:
From https://source.cloud.google.com/results/invocations/7deedac3-8b60-447d-bb5e-74a4a6653e57/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Ffirestore/log:

```python
_____________________________ test_collection_add ______________________________
client = <google.cloud.firestore_v1.client.Client object at 0x7fb9ca171ad0>
cleanup = <built-in method append of list object at 0x7fb9c7c0acb0>
    def test_collection_add(client, cleanup):
        collection1 = client.collection("collek")
        collection2 = client.collection("collek", "shun", "child")
        collection3 = client.collection("collek", "table", "child")
        explicit_doc_id = "hula" + unique_resource_id("-")
>       assert set(collection1.list_documents()) == set()
E       assert set([<google....fb9c7c06290>]) == set([])
E         Extra items in the left set:
E         <google.cloud.firestore_v1.document.DocumentReference object at 0x7fb9c7c06290>
E         Full diff:
E         - set([<google.cloud.firestore_v1.document.DocumentReference object at 0x7fb9c7c06290>])
E + set([])
```

This test should be using a unique / name-mangled collection ID, in order to prevent collision with CI running for other PRs.

PR #8440, opened 2019-06-20
Storage: fix Client download_blob_to_file fails

### LINKED ISSUE #8301: Storage: Client.download_blob_to_file fails.
`Client.download_blob_to_file()` does not work properly when passing a URI because the parsed path which is passed as a blob name has the initial `/` .

https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/client.py#L389-L395

It should be removed from the path before passing into a Blob constructor.
```
            scheme, netloc, path, query, frag = urlsplit(blob_or_uri)
            if scheme != "gs":
                raise ValueError("URI scheme must be gs")
            blob_name = path[1:]
            bucket = Bucket(self, name=netloc)
            blob_or_uri = Blob(blob_name, bucket)

            blob_or_uri.download_to_file(file_obj, client=self, start=start, end=end)
```

BTW: I tried to create a pull request to fix it but I was not allowed to push my commit into this repo. How can I contribute?

HemangChothani:
Fixes #8301

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -600,7 +600,7 @@ class TestStorageListFiles(TestStorageFiles):
         _empty_bucket(cls.bucket)
 
         logo_path = cls.FILES["logo"]["path"]
-        blob = storage.Blob(cls.FILENAMESFILENAMES[0], bucket=cls.bucket)
+        blob = storage.Blob(cls.FILENAMES[0], bucket=cls.bucket)
         blob.upload_from_filename(logo_path)
         cls.suite_blobs_to_delete = [blob]
 
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -552,6 +552,21 @@ class TestStorageWriteFiles(TestStorageFiles):
         copied_contents = new_blob.download_as_string()
         self.assertEqual(base_contents, copied_contents)
 
+    def test_download_blob_w_uri(self):
+        blob = self.bucket.blob("MyBuffer")
+        file_contents = b"Hello World"
+        blob.upload_from_string(file_contents)
+        self.case_blobs_to_delete.append(blob)
+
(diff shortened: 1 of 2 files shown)


tseaver:
@HemangChothani Please check that unit tests pass with your change:  if they do, then we are missing an assertion that the path is passed as expected.  Also, please add a system test which exercises this feature (and ensure it passes with your change).

storage/tests/system.py
```diff
@@ -585,7 +600,7 @@ def setUpClass(cls):
         _empty_bucket(cls.bucket)
 
         logo_path = cls.FILES["logo"]["path"]
-        blob = storage.Blob(cls.FILENAMES[0], bucket=cls.bucket)
+        blob = storage.Blob(cls.FILENAMESFILENAMES[0], bucket=cls.bucket)
```
tseaver: Please fix this typo.

storage/tests/system.py
```diff
@@ -552,6 +552,21 @@ def test_copy_existing_file(self):
...
 
+    def test_download_blob_w_uri(self):
+        blob = self.bucket.blob("MyBuffer")
+        file_contents = b"Hello World"
+        blob.upload_from_string(file_contents)
+        self.case_blobs_to_delete.append(blob)
+
+        temp_filename = tempfile.mktemp()
+        with open(temp_filename, "wb") as file_obj:
+            Config.CLIENT.download_blob_to_file('gs://'+self.bucket.name+'/MyBuffer', file_obj)
```
tseaver: Please move the construction of the `gs://` URI out to a separate line, and whitespace around the `+` operators to appease lint.

tseaver:
Bigtable snippet flakes reported as #8479, #8480.

Firestore systest flake due to #7829.

PR #8447, opened 2019-06-20
PubSub: Make synth operations on PubSub types idempotent

plamut:
This commit assures that the code block for injecting FlowControl docs is not added on every synth tool run - it is only added if one does not exist yet.

Contrary to my initial assumption in #8293, it appears that synthtool can be run multiple times on `types.py` and other files after generating them, thus adjustment is necessary.

### How to test
**Steps to perform:**

- Install the [synthtool](https://github.com/googleapis/synthtool) and run it from inside the `pubsub/` directory,
- Run the synthtool a few more times.

A new code block for injecting `FlowControl` docstrings is added after every synthtool run, growing the file indefinitely.

Only a single instance of the docstring-injecting code remains in the file even after multiple runs. Nothing else is changed.


plamut:
Closing in favor of #8448 that also includes the changes from here.

PR #8448, opened 2019-06-20
PubSub: Document batch settings, make synth operations idempotent

plamut:
This PR makes sure that `BatchSettings` settings are properly documented in the generated documentation. It also fixes multiple repeated blocks for adding `FlowControl` docstrings (supersedes #8447).

### How to test

**Steps to perform:**

(from inside the `pubsub/` directory)
- Install the [synthtool](https://github.com/googleapis/synthtool) and run it.
- Run the synthtool a few more times.
- Generate the pubsub docs locally:
    ```
    $ nox -f noxfile.py -s docs
    ```
- Open the generated docs for `BatchSettings`:
    ```
    $ chromium-browser "file://$GOOG_CLOUD_PYTHON_DIR/pubsub/docs/_build/html/types.html#google.cloud.pubsub_v1.types.BatchSettings"
    ```

- A new code block for injecting `FlowControl` docstrings is added after every synthtool run, growing the file indefinitely.
- `BatchSettings` attributes only have uninformative generic descriptions.

- No new code is added to `types.py` by the synthtool.
- The `BatchSettings` type and its attributes have nice descriptions (as opposed to the [current docs](https://googleapis.github.io/google-cloud-python/latest/pubsub/types.html?highlight=flowcontrol#google.cloud.pubsub_v1.types.BatchSettings)).


pubsub/synth.py
```diff
@@ -168,6 +168,20 @@ def _merge_dict(d1, d2):
...
+    r"if sys\.version_info >= \(3, 5\):",
+    "",
+)
+s.replace(
+    "google/cloud/pubsub_v1/types.py",
+    r"    FlowControl\..*?__doc__ = \(.*?\)",
+    "",
+    flags=re.DOTALL,
+)
+
```
tseaver: I don't understand why we would be removing docstring changes for `FlowControl` while adding them for `BatchSettings`.

plamut: We are not removing them, we are replacing them with ... the very same docstrings. Every synth run added a new block identical block for _FlowControl_, which is fixed here.

Spotted when reviewing #8444

plamut: Background: I originally assumed that synth is ran only once after the files are generated, but this does not seem to be the case, and is expected? Or do we maybe have an issue with the workflow?

tseaver: @plamut Looking above, the file `"google/cloud/pubsub_v1/types.py"` is *excluded* from the `s.move` call, which means we shouldn't be fixing it in `synth.py` at all:  we should just be checking it in directly.

plamut: Oh .. that's new info to me, thanks. Will make the change right away.

Issue #8451, opened 2019-06-20
Bigtable: Separate row types to remove confusion around return types of row.commit


crwilcox:
While @billyjacobson was working on https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2201 he noticed that the version of `commit` for `row.commit` only returned `None`. This got me digging into how this is authored. and if a change would be warranted.

### TL;DR
We have 3 different row types that are returned from `table.row` and they all have disparate returns types for commit. This is a bit confusing and I am curious if we can't converge these a bit. At minimum it seems maybe DirectRow should return something (rows modified count, the actual rows, True on success?).

### Details
Today, users author code like
```
table = instance.table(table_id)
row = table.row(row_key)
```
In the above case it will return a `DirectRow`. There are also `ConditionalRow` and `AppendRow` which will be returned if additional optional args are specified to a `table.row` call.

```
# default -> DirectRow
# filter set -> ConditionalRow
# append set -> AppendRow
class Table:
def row(self, row_key, filter_=None, append=False):
```

Each of these calls has different behaviors. 

- DirectRow: No return type today.

- ConditionalRow: Returns if the filter provided to `table.row` matched

- AppendRow: returns the new content of the modified cells as a dict.

Was this a conscious choice made to not return anything for `DirectRow.commit`? Do we want to make a change here to have it return a response?

Update: plan is to separate, not converge types.

tswast:
DirectRow: I'd expect the row content to be returned on success, as seems to be the case for the other row types.

sduskis:
See the proto: https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable.proto

- `DirectRow` used for `MutateRow` operations.  The operation effectively returns `Empty`
- `ConditionalRow` used for `CheckAndMutateRow`.  The operation returns a boolean `predicate_matched`
- `AppendRow` used for `ReadModifyWriteRow` with an `Append` type (instead of an `Increment` type).   The operation returns a `Row` proto (A Row containing the new contents of all cells modified by the request).

I believe that @tseaver developed the original set of Row objects.

sduskis:
@crwilcox, @tswast: I suggest closing this issue.  Things are the way they are for a reason.  I don't believe that convergence in the Python API is desirable, since the Python API ought to match the service API.

crwilcox:
[...]

PR #8460, opened 2019-06-21
Bigtable: Converging return types of row.commit

IlyaFaer:
https://github.com/googleapis/google-cloud-python/issues/8451

### DIFF
--- a/bigtable/google/cloud/bigtable/row.py
+++ b/bigtable/google/cloud/bigtable/row.py
@@ -82,6 +82,21 @@ class Row(object):
         """
         return self._table
 
+    @property
+    def row_content(self):
+        """Row content.
+
+        For example:
+
+        .. literalinclude:: snippets_table.py
+            :start-after: [START bigtable_row_row_content]
+            :end-before: [END bigtable_row_row_content]
+
+        :rtype: row_content: :class:`PartialRowData <google.cloud.bigtable.row_data.PartialRowData>`
+        :returns: Contents of the row read from table.
+        """
+        return self.table.read_row(self.row_key)
+
 
 class _SetDeleteRow(Row):
     """Row helper for setting or deleting cell values.
@@ -451,11 +466,15 @@ class DirectRow(_SetDeleteRow):
             :start-after: [START bigtable_row_commit]
             :end-before: [END bigtable_row_commit]
 
+        :rtype: row_content: :class:`PartialRowData <google.cloud.bigtable.row_data.PartialRowData>`
+        :returns: Contents of the row.
+
         :raises: :exc:`~.table.TooManyMutationsError` if the number of
                  mutations is greater than 100,000.
         """
         self._table.mutate_rows([self])
         self.clear()
+        return self.row_content
 
     def clear(self):
         """Removes all currently accumulated mutations on the current row.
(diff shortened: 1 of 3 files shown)


bigtable/google/cloud/bigtable/row.py
```diff
@@ -82,6 +82,21 @@ def table(self):
...
+        For example:
+
+        .. literalinclude:: snippets_table.py
+            :start-after: [START bigtable_row_row_content]
+            :end-before: [END bigtable_row_row_content]
+
+        :rtype: row_content: :class:`PartialRowData <google.cloud.bigtable.row_data.PartialRowData>`
+        :returns: Contents of the row read from table.
+        """
+        return self.table.read_row(self.row_key)
```
IlyaFaer: Added property in superclass, that mirrors to `read_row()`

sduskis:
I think this PR was premature. The issue was raised without understanding why things are the way they are.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -451,11 +466,15 @@ def commit(self):
...
 
+        :rtype: row_content: :class:`PartialRowData <google.cloud.bigtable.row_data.PartialRowData>`
+        :returns: Contents of the row.
+
         :raises: :exc:`~.table.TooManyMutationsError` if the number of
                  mutations is greater than 100,000.
         """
         self._table.mutate_rows([self])
         self.clear()
+        return self.row_content
```
sduskis: This is really bad.  Its an extra service call for no reason.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -578,15 +597,15 @@ def commit(self):
...
+        data_client.check_and_mutate_row(
             table_name=self._table.name,
             row_key=self._row_key,
             predicate_filter=self._filter.to_pb(),
             true_mutations=true_mutations,
             false_mutations=false_mutations,
         )
         self.clear()
-        return resp.predicate_matched
+        return self.row_content
```
sduskis: This changes the existing behavior, and doesn't match the service's results.  This actually will reduce usage and performance.

bigtable/google/cloud/bigtable/row.py
```diff
@@ -907,15 +923,13 @@ def commit(self):
...
+        data_client.read_modify_write_row(
             table_name=self._table.name, row_key=self._row_key, rules=self._rule_pb_list
         )
 
         # Reset modifications after commit-ing request.
         self.clear()
-
-        # NOTE: We expect row_response.key == self._row_key but don't check.
-        return _parse_rmw_row_response(row_response)
+        return self.row_content
```
sduskis: This is also bad.  This causes an extra lookup.  `row_response` is already equivalent to `row_content` without requiring a lookup

IlyaFaer: As I understood, the request was to return the whole row, while `row_content` contains only modified cells
https://github.com/googleapis/google-cloud-python/blob/fa6e4f5784f0063ff832159e561aa89d45497dea/bigtable/google/cloud/bigtable/row.py#L939

bigtable/google/cloud/bigtable/row.py
IlyaFaer: I knew, it's an extra call, but reading row by indexed key looks lightly (if number/size of columns is not too large, of course). And changing the behavior was requested in issue. Is there are any indicator to understand, that issue approved by people, who know the details? It's not first time, when I PR'ed too fast. For newbie like me some things are not obvious (especially, when googlers said that it's not bad idea)

Issue #7303, opened 2019-02-08
Spanner: Python client not backing off on retries when RetryInfo not present.


crwilcox:
When a session aborts, but doesn't have retry info python immediately retries. It seems to lack any sort of backoff here which may be incorrect. node attempts to. 

https://github.com/googleapis/nodejs-spanner/blob/79f7855b3f6df59afe322dd578ed65f408a2fc2b/src/transaction.ts#L836 

Python implementation currently returning 'None' if nothing is found: https://github.com/googleapis/google-cloud-python/blob/92465cbc4d9c0ba251838e9cd17f61d14b470e04/spanner/google/cloud/spanner_v1/session.py#L353

Go: https://github.com/googleapis/google-cloud-go/blob/ab1512ccf757de1b7f7f62883d9f975dfc152cef/spanner/internal/backoff/backoff.go#L43

tseaver:
IIRC, this was the behavior specified by the spanner team (retry immediately if no trailing headers specify a perior).

snehashah16:
> When a session aborts, but doesn't have retry info python immediately retries. It seems to lack any sort of backoff here which may be incorrect. node attempts to.

Do you mean a Transaction aborts ?

For java, we use a default value to backoff & then retry if retryDelay is not specified in the response:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L2681-L2705

The current values are:
 MIN_BACKOFF_MS = 1000;
 MAX_BACKOFF_MS = 32000;

private static ExponentialBackOff newBackOff() {
    return new ExponentialBackOff.Builder()
        .setInitialIntervalMillis(MIN_BACKOFF_MS)
        .setMaxIntervalMillis(MAX_BACKOFF_MS)
        .setMaxElapsedTimeMillis(Integer.MAX_VALUE) // Prevent Backoff.STOP from getting returned.
        .build();

Since I don't recollect any other context, I would default to the behavior of the Java Client lib.

snehashah16:
Lets verify this behavior across each of our libraries and fix as needed.

tseaver:
@crwilcox I presume that removing the 'type: question' label and adding a priority means you want this to have 'type: bug'.

IlyaFaer:
[...]

PR #8461, opened 2019-06-21
Spanner: add backoff for `run_in_transaction' when backend does not provide 'RetryInfo' in response.

IlyaFaer:
https://github.com/googleapis/google-cloud-python/issues/7303


spanner/google/cloud/spanner_v1/session.py
```diff
@@ -308,7 +311,7 @@ def run_in_transaction(self, func, *args, **kw):
                 txn.commit()
             except Aborted as exc:
                 del self._transaction
-                _delay_until_retry(exc, deadline)
+                _delay_until_retry(exc, deadline, attempts)
```
IlyaFaer: Is it an error here?
`try-except` then delay before retry and what next? I assume `continue` statement must be here, like in lines 300-302, am I right?

spanner/tests/unit/test_session.py
```diff
@@ -1033,18 +1033,20 @@ def unit_of_work(txn, *args, **kw):
...
-                    session.run_in_transaction(unit_of_work, timeout_secs=1)
+                    session.run_in_transaction(unit_of_work, timeout_secs=8)
 
-        sleep_mock.assert_not_called()
+        sleep_mock.assert_has_calls(
+            (mock.call(2), mock.call(4))
+        )
 
-        self.assertEqual(len(called_with), 2)
+        self.assertEqual(len(called_with), 3)
```
IlyaFaer: This test has been checking retries without delay, but now it isn't actual. So I redact it to check if rety-calls were made with backing off delay

spanner/google/cloud/spanner_v1/session.py
```diff
@@ -366,3 +376,5 @@ def _get_retry_delay(cause):
         retry_info.ParseFromString(retry_info_pb)
         nanos = retry_info.retry_delay.nanos
         return retry_info.retry_delay.seconds + nanos / 1.0e9
+
+    return 2 ** attempts + random.random()
```
IlyaFaer: @crwilcox, it's done, I've added jitter

spanner/google/cloud/spanner_v1/session.py
crwilcox: I don't think continue is strictly needed here. line 300-302 needs to avoid executing the commit. In the second case I don't think that is a concern.

crwilcox:
@IlyaFaer this looks good, but coverage is missing for 
```
google/cloud/spanner_v1/session.py                     126      1     40      2    98%   346, 345->346, 349->exit

```

IlyaFaer:
@crwilcox , okey, I've added two more tests for delay helper itself, and my local coverage says, it's 100% now

IlyaFaer:
@crwilcox, made commit with black reformat. Sorry, I have a habit of writing in single quotes

PR #8462, opened 2019-06-21
Pub/Sub: hide a method in docs

anguillanneuf:
Removed `StreamingPullFuture.add_done_callback()` from docs.

### DIFF
--- a/pubsub/docs/subscriber/api/futures.rst
+++ b/pubsub/docs/subscriber/api/futures.rst
@@ -4,3 +4,4 @@ Futures
 .. automodule:: google.cloud.pubsub_v1.subscriber.futures
   :members:
   :inherited-members:
+  :exclude-members: add_done_callback


pubsub/docs/subscriber/api/futures.rst
```diff
@@ -4,3 +4,4 @@ Futures
 .. automodule:: google.cloud.pubsub_v1.subscriber.futures
   :members:
   :inherited-members:
+  :exclude-members: add_done_callback
```
plamut: Why do we need to exclude this method? As far as I can see the `StreamingPullFuture` still has this method, and it work, too.

anguillanneuf: I find it confusing to try to think of ways to use this method. Unlike publish futures which don't block, this one does block. What would `add_done_callback` mean? Would you think of a scenario that it can be used?

plamut: By blocking you mean when `future.result()` is called?

The code that calls `future = subscriber.subscribe()` does not have to call `.result()` on it, it can do other things while callbacks are processing received messages. The code might pass a reference to this future to, say, a "manager" thread that cancels the future after a certain time period, and might want to be notified when this happens to do some finalization work. Having an option to register an `on_done_callback` is ideal for that.

Generally, if something is not used, it should not be simply hidden in the docs, but instead removed altogether IMO. However, I can imagine `on_done_callback` being useful in some applications ...

anguillanneuf: Right. 

I think you said is fair. I will close this and add a short example of how to use `add_done_callback` to the doc. Thanks Peter!

Issue #8454, opened 2019-06-20
Bigtable: Snippet setup fails due to rate-based quota.


tseaver:
From [this failiure today](https://source.cloud.google.com/results/invocations/7fea056f-3154-427a-8a7c-efbea5d0dece/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable/log):

```
google/cloud/bigtable/cluster.py:340: in delete
    client.instance_admin_client.delete_cluster(self.name)
google/cloud/bigtable_admin_v2/gapic/bigtable_instance_admin_client.py:1179: in delete_cluster
    request, retry=retry, timeout=timeout, metadata=metadata
../api_core/google/api_core/gapic_v1/method.py:143: in __call__
    return wrapped_func(*args, **kwargs)
../api_core/google/api_core/retry.py:273: in retry_wrapped_func
    on_error=on_error,
../api_core/google/api_core/retry.py:182: in retry_target
    return target()
../api_core/google/api_core/timeout.py:214: in func_with_timeout
    return func(*args, **kwargs)
../api_core/google/api_core/grpc_helpers.py:59: in error_remapped_callable
    six.raise_from(exceptions.from_grpc_error(exc), exc)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
value = ResourceExhausted(u"Quota exceeded for quota group 'InstancesWriteGroup' and l... 'bigtableadmin.googleapis.com' for consumer 'project_number:1065521786570'.",)
from_value = <_Rendezvous of RPC that terminated with:
	status = StatusCode.RESOURCE_EXHAUS...ogleapis.com' for consumer 'project_number:1065521786570'.","grpc_status":8}"
>
    def raise_from(value, from_value):
>       raise value
E       ResourceExhausted: 429 Quota exceeded for quota group 'InstancesWriteGroup' and limit 'CLIENT_PROJECT-1d' of service 'bigtableadmin.googleapis.com' for consumer 'project_number:1065521786570'.
.nox/snippets-2-7/lib/python2.7/site-packages/six.py:737: ResourceExhausted
```

Including many more failures.  @sduskis, could we get this particular quota / rate limit eased for our CI project?

sduskis:
[...]

PR #8464, opened 2019-06-21
All: Add docs job to publish to googleapis.dev.

busunkim96:
I tested a few of these jobs via manual triggers on Fusion. (core, firestore, iot)

I split up the `core` and `api_core` docs and allow for those docs to be built separately. There are symlinks in `docs/core` to the original docs.  `config.rst` and `index.rst` are slightly different, so they were duplicated.


bigquery/.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "bigquery",
+  "name_pretty": "Google Cloud Bigquery",
```
tswast: Capital Q for BigQuery, please.

.kokoro/publish-docs.sh
```diff
@@ -19,15 +17,25 @@ python3.6 -m nox --version
...
 nox -s docs
 
-cd docs/_build/html
-
 python3 -m pip install gcp-docuploader
 
+# install a json parser
+sudo apt-get update
+sudo apt-get -y install software-properties-common
+sudo add-apt-repository universe
```
tseaver: `add-apt-repository` needs to be before `apt-get update`, or else the new repo won't be used to install anything.

.kokoro/publish-docs.sh
```diff
@@ -19,15 +17,25 @@ python3.6 -m nox --version
...
+sudo apt-get -y install jq
+
 # create metadata
 python3 -m docuploader create-metadata \
---name ${NAME} \
---version ${VERSION} \
---language python
+  --name=$(jq --raw-output '.name // empty' .repo-metadata.json) \
```
tseaver: I'm unsure what is wanted here:  `python3 setup.py --name` would give the "Python project" name, e.g. `google-cloud-foo`.

.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "monitoring",
+  "name_pretty": "Stackdriver Monitoring",
```
tseaver: I don't think it can be correct to have this as a top-level default.

.kokoro/publish-docs.sh
busunkim96: This 'name' gets used to construct the url at our new docs site. It usually matches the project name, but sometimes it's different.

googleapis.dev/python/**bigquery**/latest

.kokoro/publish-docs.sh
busunkim96: The `apt-get update` is to install the command `add-apt-repository`. Otherwise I get:

`sudo: add-apt-repository: command not found`

Once our doc publishing tool can accept this metadata format directly this won't be needed.

.kokoro/publish-docs.sh
tseaver: Hmm, then maybe we should be using the one in `setup.py` for `--distribution-name`?

busunkim96: Done.

.kokoro/publish-docs.sh
tseaver: OK.  I still think you need to re-run `apt-get update` *after* running `add-apt-repository`, or else you won't be able to install anything from the new repository.

tseaver:
Bigtable failure is due to #8454.

Spanner failure reported in #8474.

PR #8472, opened 2019-06-25
Firestore: Preserve reference to missing documents in 'Client.get_all'.

### LINKED ISSUE #7564: Firestore: get_all removes .id key for non existing DocumentSnapshot

pchauhan-qlogic:
Fix #7564

### DIFF
--- a/firestore/tests/system.py
+++ b/firestore/tests/system.py
@@ -768,11 +768,7 @@ def test_get_all(client, cleanup):
 
     assert snapshots[0].exists
     assert snapshots[1].exists
-
-    # missing document attribute
     assert not snapshots[2].exists
-    assert snapshots[2].id == "b"
-    assert not snapshots[2]._data
 
     snapshots = [snapshot for snapshot in snapshots if snapshot.exists]
     id_attr = operator.attrgetter("id")
--- a/firestore/tests/system.py
+++ b/firestore/tests/system.py
@@ -768,7 +768,12 @@ def test_get_all(client, cleanup):
 
     assert snapshots[0].exists
     assert snapshots[1].exists
+
+    # missing document attribute
     assert not snapshots[2].exists
+    assert snapshots[2].id == "b"
+    assert not snapshots[2]._data
+
     snapshots = [snapshot for snapshot in snapshots if snapshot.exists]
     id_attr = operator.attrgetter("id")
     snapshots.sort(key=id_attr)
--- a/firestore/tests/unit/v1/test_client.py
+++ b/firestore/tests/unit/v1/test_client.py
@@ -594,12 +594,16 @@ class Test__parse_batch_get(unittest.TestCase):
         self.assertEqual(snapshot.update_time, update_time)
 
     def test_missing(self):
+        from google.cloud.firestore_v1.document import DocumentReference
+
         ref_string = self._dummy_ref_string()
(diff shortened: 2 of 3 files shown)


firestore/tests/system.py
```diff
@@ -768,7 +768,12 @@ def test_get_all(client, cleanup):
 
     assert snapshots[0].exists
     assert snapshots[1].exists
+
+    # missing document attribute
     assert not snapshots[2].exists
+    assert snapshots[2].id == "b"
+    assert not snapshots[2]._data
```
tseaver: Please move these assertions to the unit test below/

firestore/tests/unit/v1/test_client.py
```diff
@@ -596,8 +596,9 @@ def test_found(self):
     def test_missing(self):
         ref_string = self._dummy_ref_string()
         response_pb = _make_batch_response(missing=ref_string)
+        reference_map = {ref_string: mock.sentinel.reference}
 
-        snapshot = self._call_fut(response_pb, {})
+        snapshot = self._call_fut(response_pb, reference_map)
```
tseaver: Add assertions here about the snapshot's `id`, data.

tseaver:
The Firestore system test failure is due to #7130.

firestore/tests/unit/v1/test_client.py
pchauhan-qlogic: Added assertions about snapshot's id and data @tseaver

firestore/tests/system.py
pchauhan-qlogic: Done

tseaver:
Unrelated Bigtable snipped flakiness.

PR #8476, opened 2019-06-25
Infrastructure: Use the Github API to compute changed files for PRs.

tseaver:
Rather than computing a diff from the squirrely baseref hash (as provided by `ci_diff_helper`), just use the Github API directly to query for the [list of files for a PR](https://developer.github.com/v3/pulls/#list-pull-requests-files).  We should run many fewer spurious test suites if this works as expected.  I've tested the new `get_changed_files_from_pr` function manually against both small and large PRs (which involve pagination).

### DIFF
--- a/test_utils/scripts/get_target_packages_kokoro.py
+++ b/test_utils/scripts/get_target_packages_kokoro.py
@@ -46,16 +46,6 @@ def get_changed_files_from_base(base):
     ], stderr=subprocess.DEVNULL).decode('utf8').strip().split('\n')
 
 
-
-def parse_response_link(response):
-    header = response.headers.get("link")
-    if header is not None:
-        for link in header.split(","):
-            link = link.strip()
-            wrapped_url, rel = link.split(";")
-            if rel.strip() == 'rel="next"':
-                return wrapped_url[1:-1]  # strip <>
-
 _URL_TEMPLATE = (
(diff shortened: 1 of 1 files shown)


test_utils/scripts/get_target_packages_kokoro.py
```diff
@@ -39,12 +40,37 @@ def get_base(environment):
...
+)
+
+
+def get_changed_files_from_pr(pr):
+    url = _URL_TEMPLATE.format(pr)
+    while url is not None:
+        response = requests.get(url)
+        for info in response.json():
+            yield info['filename']
+        url = parse_response_link(response)
```
busunkim96: In Autorelease and Autosynth pagination is handled like this: Not sure if it would be possible to do it the same way here:

```
def list_org_repos(self, org: str, type: str = None) -> Sequence[dict]:
    url = f"{self.GITHUB_ROOT}/orgs/{org}/repos"
    
    while url:
        response = self.session.get(url, params={"type": type})
        response.raise_for_status()
        for item in response.json():
             yield item
   
        url = response.links.get("next", {}).get("url")

```

tseaver: Cool, I didn't know about the [`Response.links` feature in requests](https://2.python-requests.org/en/master/user/advanced/#link-headers).

tseaver: efa9411

Issue #7218, opened 2019-01-29
Storage: add an API method to give us a streaming file object


dmsolow:
It doesn't look like there's a way to get a streaming download from google storage in the Python API. We have `download_to_file` , `download_to_string`, and `download_to_filename`, but I don't see anything that returns a file-like object that can be streamed. This is a disadvantage for many file types which can usefully be processed as they download.

Can a method like this be added?

tseaver:
@dmsolow Hmm, `Blob.download_to_file` *takes* a file object -- does that not suit your usecase?

dmsolow:
I don't think so. The situation is that it's often useful to start processing a file as it downloads instead of waiting until it's finished. For example if there's 1GB CSV file in google storage, it should be possible to parse it line by line as it's downloaded.

It's fairly common for network libraries to offer this kind of functionality. For example in the standard `urllib.request` HTTP library:

```
import urllib.request
import csv
from io import TextIOWrapper

with urllib.request.urlopen('http://test.com/big.csv') as f:
    wrapped = TextIOWrapper(f) # decode from bytes to str
    reader = csv.reader(wrapped)
    for row in reader:
       print(row[0])
```

This parses the CSV as it's downloaded. I'd like to get the same functionality from google storage. If there's already a good way to do this with the current library, please let me know.

tseaver:
Hmm, you'll need to have the "stream consumer" running in a separate thread / process to do much good.  You *can* make this work using Python's `os.pipe`:  see [this gist](https://gist.github.com/tseaver/eebb9f8605f6e606b4a985e7b8ea8f85), which produces the following output:

```bash
$ bin/python pipe_test.py 
reader: start
reader: read one chunk
reader: read one chunk
...
reader: read one chunk
reader: read 800000 bytes
```

dmsolow:
Using a separate thread kind of feels like a hack to me, but it is surely one way to do it. I think the ability to do this without using extra threads would be widely useful, but idk how hard it would be to implement.

tseaver:
OK, looking at the [underlying implementation in `google-resumable-media`](https://github.com/googleapis/google-resumable-media-python/blob/34a36be0c53b7517bd3ddc2c281bdaaa5b9c4ef7/google/resumable_media/requests/download.py#L117-L125), all that we actually expect of the `file` object is that it has a `write` method, which is then passed each chunk as it is downloaded.

You could therefore pass in an instance of your own class which wrapped the underlying stream, e.g.:

[...]

PR #8484, opened 2019-06-27
Storage: add an API method to give us a streaming file object

IlyaFaer:
Towards: #7218
Here is a proposition


storage/google/cloud/storage/blob.py
```diff
@@ -586,6 +588,35 @@ def _do_download(
             while not download.finished:
                 download.consume_next_chunk(transport)
 
+    def get_streaming_file(self, filename, client=None):
```
IlyaFaer: This method is similar to `download_to_filename()`, but it doesn't use `start` and `end` args, 'cause same effect can be achieved with `seek()` and `read()` methods

storage/google/cloud/storage/blob.py
```diff
@@ -1951,6 +1982,84 @@ def updated(self):
...
+                if to_download > 0:
+                    chunk_size = self._download.chunk_size
+                    self._download.chunk_size = to_download
+                    self.seek(*_EOF)
+                    self._download_chunk()
+                    self._download.chunk_size = chunk_size
+
+            self.seek(cursor_place, 0)
+
+        return super(StreamingFile, self).read(size)
```
IlyaFaer: The main idea is to download the part, that user is trying to read, but which is not yet downloaded. Before download we should remember cursors position, then we must move cursor to the end of file (to avoid writing downloaded data to the middle of the file in case if user made `seek()`). And then we should brought cursor back to users position and call `super().read()`

cmey:
Amazing, this is exactly what I need to work with large files (larger than available memory)!

mf2199:
@tswast @frankyn @crwilcox A friendly reminder: This PR has been hanging out for awhile. We could use some feedback regarding its status.

storage/google/cloud/storage/blob.py
tswast: I'd prefer we don't add new public methods to `Blob`, instead add this functionality to the `Client` class.

tseaver:
@IlyaFaer I opened #8996 to track the flaky systest.

frankyn:
Hi @IlyaFaer,

I'm closing this issue after internal discussions. This a feature request and it's not within scope of what should be worked on at this time. We have a priority on fixing reliability bugs in our clients before feature requests.

Additionally, this feature in particular requires a design review that has not yet been prioritized.
Thank you for your patience.

PR #8491, opened 2019-06-27
BigQuery: Add Routines API.

tswast:
Adds support for managing permanent functions in BigQuery, such
as scalar UDFs and stored procedures.  At present, only
scalar UDF functionality is available.

Routines are registered as resources inside of datasets, and
allow expected CRUD operations.  Currently, routines do not
support partial updates.

See: https://cloud.google.com/bigquery/docs/reference/rest/v2/routines


bigquery/google/cloud/bigquery/client.py
```diff
@@ -567,12 +632,54 @@ def update_model(self, model, fields, retry=DEFAULT_RETRY):
         )
         return Model.from_api_repr(api_response)
 
+    def update_routine(self, routine, fields, retry=DEFAULT_RETRY):
```
shollyman: Should we have something in the docstring about partial update support b/134928435?

bigquery/google/cloud/bigquery/routine.py
```diff
@@ -0,0 +1,502 @@
...
+    def return_type(self, value):
+        if value:
+            resource = json_format.MessageToDict(value)
+        else:
+            resource = None
+        self._properties[self._PROPERTY_TO_API_FIELD["return_type"]] = resource
+
+    @property
+    def imported_libraries(self):
+        """List[str]: The path of the imported JavaScript libraries.
```
shollyman: We should be clearer that this is a set of GCS paths e.g. gs://my-bucket/path/to/lib1.js, not js import/require statements.

PR #8534, opened 2019-06-28
Speech: Add 'client_options' support (via synth).

tseaver:


tseaver:
Asset, BigQuery, Spanner, and Video Intelligence jobs all fail due to Kokoro-internal glitch ("Cannot check out non-mergeable pull request as if merged.").

PR #8536, opened 2019-06-28
BigQuery Storage: Add 'client_options' support (via synth).

tseaver:
Supersedes #8499.


tseaver:
API Core, AutoML, and Tasks failures due to Kokoro-internal glitch: `Cannot check out non-mergeable pull request as if merged.`

PR #8546, opened 2019-07-01
Remove type information for kwargs to not conflict with type checkers

crwilcox:


tseaver:
Unrelated Storage flake reported in #8552.

PR #8547, opened 2019-07-01
Datastore:  add retry param to page iterator.

### LINKED ISSUE #6119: Datastore: expose retry parameters
Would it be possible to expose retry parameters for datastore?

We are seeing frequent "InternalServerError: 500 GOAWAY received" errors during queries, and it looks like the default retry does not handle internal errors: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/datastore/google/cloud/datastore_v1/gapic/datastore_client_config.py#L5 

```
InternalServerError: 500 GOAWAY received
at raise_from (c:\project\src\third_party\six.py:737)
at error_remapped_callable (c:\project\src\third_party\google\api_core\grpc_helpers.py:61)
at func_with_timeout (c:\project\src\third_party\google\api_core\timeout.py:206)
at retry_target (c:\project\src\third_party\google\api_core\retry.py:177)
at retry_wrapped_func (c:\project\src\third_party\google\api_core\retry.py:260)
at __call__ (c:\project\src\third_party\google\api_core\gapic_v1\method.py:139)
at run_query (c:\project\src\third_party\google\cloud\datastore_v1\gapic\datastore_client.py:294)
at _next_page (c:\project\src\third_party\google\cloud\datastore\query.py:517)
at _page_iter (c:\project\src\third_party\google\api_core\page_iterator.py:230)
at _items_iter (c:\project\src\third_party\google\api_core\page_iterator.py:199)
```

It's not possible for us to add our own retry here either to retry fetching a particular page, as _items_iter and _page_iter are generators and we can't retry on failures during iteration.

Would it be possible to expose these settings somehow, or add INTERNAL to the list of retryable errors? We'd like to avoid doing something like https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3128#issuecomment-386422844

pchauhan-qlogic:
Fixes #6119

### DIFF
--- a/datastore/google/cloud/datastore/helpers.py
+++ b/datastore/google/cloud/datastore/helpers.py
@@ -18,6 +18,7 @@ The non-private functions are part of the API.
 """
 
 import datetime
+import functools
 import itertools
 
 from google.protobuf import struct_pb2
@@ -469,6 +470,14 @@ def _set_protobuf_value(value_pb, val):
         value_pb.geo_point_value.CopyFrom(val)
     else:  # scalar, just assign
         setattr(value_pb, attr, val)
+
+
+def _call_api(fnc_call, retry, *args, **kwargs):
+
+    call = functools.partial(fnc_call, *args, **kwargs)
+    if retry:
+        call = retry(call)
+    return call()
 
 
 class GeoPoint(object):
--- a/datastore/google/cloud/datastore/helpers.py
+++ b/datastore/google/cloud/datastore/helpers.py
@@ -473,11 +473,9 @@ def _set_protobuf_value(value_pb, val):
 
 
 def _call_api(fnc_call, retry, *args, **kwargs):
-
-    call = functools.partial(fnc_call, *args, **kwargs)
     if retry:
-        call = retry(call)
-    return call()
+        return retry(functools.partial(fnc_call, *args, **kwargs))()
+    return fnc_call(*args, **kwargs)
 
 
 class GeoPoint(object):
--- a/datastore/google/cloud/datastore/retry.py
+++ b/datastore/google/cloud/datastore/retry.py
@@ -1,0 +1,55 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


datastore/google/cloud/datastore/helpers.py
```diff
@@ -471,6 +472,14 @@ def _set_protobuf_value(value_pb, val):
         setattr(value_pb, attr, val)
 
 
+def _call_api(fnc_call, retry, *args, **kwargs):
+
+    call = functools.partial(fnc_call, *args, **kwargs)
+    if retry:
```
tseaver: If `retry` is None, then avoide the partial and just `return fnc_call(*args, **kwargs)`.

datastore/google/cloud/datastore/retry.py
```diff
@@ -0,0 +1,55 @@
...
+
+DEFAULT_RETRY = retry.Retry(predicate=_should_retry)
+"""The default retry object.
+
+Any method with a ``retry`` parameter will be retried automatically,
+with reasonable defaults. To disable retry, pass ``retry=None``.
+To modify the default retry behavior, call a ``with_XXX`` method
+on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
+pass ``retry=bigquery.DEFAULT_RETRY.with_deadline(30)``.
+"""
```
tseaver: Just copying this file in wholesale from bigquery isn't appropriate:  we should be refactoring to share this implementation.

datastore/tests/unit/test_retry.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_w_badGateway(self):
+        exc = mock.Mock(errors=[{"reason": "badGateway"}], spec=["errors"])
+        self.assertTrue(self._call_fut(exc))
+
+    def test_w_unstructured_bad_gateway(self):
+        from google.api_core.exceptions import BadGateway
+
+        exc = BadGateway("testing")
+        self.assertTrue(self._call_fut(exc))
```
tseaver: Likewise here for the copy-paste from BigQuery.

tseaver:
@pchauhan-qlogic Please run `nox -e black`.

This PR still needs to avoid blindly copying the BigQuery `retry` module:  for one thing, we have no reason to believe the set of retryable reasons / exception types should be the same for the two backends.

@tswast WDYT?

tswast:
That's right. BigQuery is definitely a special snowflake with respect to retry codes. I recommend choosing the codes to retry independently for Datastore.

pchauhan-qlogic:
please look into @tswast

crwilcox:
I think to bring this back to an earlier comment by @tseaver, we likely want shared code, but the list of retry codes is likely to be special per client. 

We have a higher level work item that needs to be done which is to decide on the best way forward for providing timeout and retry functionality in our clients in a way that is consistent across products.

tseaver:
Overall design of retries for manual clients is tracked in #9298.

Issue #8549, opened 2019-07-01
BigQuery: numpy release breaks Python 2.7 unit tests


tswast:
```
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-bS2Kos/fastparquet/setup.py", line 98, in <module>
        **extra
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/__init__.py", line 144, in setup
        _install_setup_requires(attrs)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/__init__.py", line 139, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/dist.py", line 717, in fetch_build_eggs
        replace_conflicting=True,
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/pkg_resources/__init__.py", line 782, in resolve
        replace_conflicting=replace_conflicting
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1065, in best_match
        return self.obtain(req, installer)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1077, in obtain
        return installer(requirement)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/dist.py", line 784, in fetch_build_egg
        return cmd.easy_install(req)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 679, in easy_install
        return self.install_item(spec, dist.location, tmpdir, deps)
      File "/tmpfs/src/github/google-cloud-python/bigquery/.nox/unit-2-7/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 705, in install_item
        dists = self.install_eggs(spec, download, tmpdir)
[...]

PR #8553, opened 2019-07-01
BigQuery: Fix unit tests for new versions of fastparquet and pytest

tswast:
`fastparquet` is an optional dependency, and is redundant with `pyarrow`. It is not necessary to include in in the `[all]` extras. This avoids an issue where the `fastparquet` dependency pulls in a pre-release version of numpy.

Closes https://github.com/googleapis/google-cloud-python/issues/8549

Version 5.0.0 of pytest modifies the behavior of `str` on the `pytest.raises()` context manager. Update the tests to call `str` on the actual exception (`context.value`), instead.

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -51,6 +51,11 @@ extras = {
 all_extras = []
 
 for extra in extras:
+    if extra == "fastparquet":
+        # Skip fastparquet from "all" because it is redundant with pyarrow and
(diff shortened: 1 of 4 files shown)


tswast:
System test failures `test_list_datasets` and `test_list_datasets_w_project` appear to be flakes.

tswast:
There's an ongoing issue on the BigQuery backend for `dataset.list`. Those are the only test failures.

Issue #8552, opened 2019-07-01
Storage: 'test_bpo_set_unset_preserves_acls' no longer sees expected 'BadRequest'.


tseaver:
From [this Kokoro job](https://source.cloud.google.com/results/invocations/2691ce70-20dd-4f90-ad95-d483b80b4ec2/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log):

```python
____________ TestIAMConfiguration.test_bpo_set_unset_preserves_acls ____________
self = <tests.system.TestIAMConfiguration testMethod=test_bpo_set_unset_preserves_acls>
    def test_bpo_set_unset_preserves_acls(self):
        new_bucket_name = "bpo-acls" + unique_resource_id("-")
        self.assertRaises(
            exceptions.NotFound, Config.CLIENT.get_bucket, new_bucket_name
        )
        bucket = retry_429(Config.CLIENT.create_bucket)(new_bucket_name)
        self.case_buckets_to_delete.append(new_bucket_name)
        blob_name = "my-blob.txt"
        blob = bucket.blob(blob_name)
        payload = b"DEADBEEF"
        blob.upload_from_string(payload)
        # Preserve ACLs before setting BPO
        bucket_acl_before = list(bucket.acl)
        blob_acl_before = list(bucket.acl)
        # Set BPO
        bucket.iam_configuration.bucket_policy_only_enabled = True
        bucket.patch()
        # While BPO is set, cannot get / set ACLs
        with self.assertRaises(exceptions.BadRequest):
>           bucket.acl.reload()
E           AssertionError: BadRequest not raised
tests/system.py:1709: AssertionError
```

tseaver:
I'm not sure this is actually "flaky":  the back-end appears to no longer return a `BadRequest` for `bucketAccessControls.list` when the bucket-policy-only flag is set.

@frankyn Can you confirm that the back-end behavior is changed?

tseaver:
[Very strange](https://source.cloud.google.com/results/invocations/422ad5ef-3bec-4018-bfc2-246ad2b9f334/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fstorage/log):  the call to `Bucket.acl.reload` *does* raise the expected `BadRequest` when running under Python 2.7, but does *not* raise as expected under Python 3.6.

tseaver:
[...]

PR #8570, opened 2019-07-02
Storage: Add 'Bucket.location_type' property.

tseaver:
- Add named constants for allowed 'storage_class' values.
- Un-deprecate STANDARD storage class.
- Docs-deprecate storage classes implying location type.
- Add 'Bucket.location_type' property.
- Add named constants for allowed 'location_type' values.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -386,20 +386,78 @@ class Bucket(_PropertyMixin):
     This is used in Bucket.delete() and Bucket.make_public().
     """
 
+    STANDARD_STORAGE_CLASS = "STANDARD"
+    """Storage class for objects accessed more than once per month."""
+
+    NEARLINE_STORAGE_CLASS = "NEARLINE"
+    """Storage class for objects accessed at most once per month."""
+
+    COLDLINE_STORAGE_CLASS = "COLDLINE"
+    """Storage class for objects accessed at most once per year."""
+
+    MULTI_REGIONAL_LEGACY_STORAGE_CLASS = "MULTI_REGIONAL"
+    """Legacy storage class.
+
+    Alias for :attr:`STANDARD_STORAGE_CLASS`.
+
+    Implies :attr:`MULTI_REGION_LOCATION_TYPE` for :attr:`location_type`.
+    """
+
+    REGIONAL_LEGACY_STORAGE_CLASS = "REGIONAL"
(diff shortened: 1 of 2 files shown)


tseaver:
Storage Kokoro job fails due to flaky systest, #8552.

JesseLovelace:
Thank you! Please hold off on merging until notified. Current prod rollout is planned for July 8

tseaver:
Force-pushed to pick up the systest fix from #8617.

tseaver:
@crwilcox, @busunkim96 Can you look into why the CLA bot is not reporting status here?

tseaver:
@JesseLovelace Please ping the PR when the back-end rollout is complete.

storage/google/cloud/storage/bucket.py
```diff
@@ -1377,6 +1435,38 @@ def location(self, value):
...
+
+        :rtype: str or ``NoneType``
+        :returns:
+            If set, one of :attr:`MULTI_REGION_LOCATION_TYPE`,
+            :attr:`REGION_LOCATION_TYPE`, or :attr:`DUAL_REGION_LOCATION_TYPE`,
+            else ``None``.
+        """
+        return self._properties.get("locationType")
+
+    @location_type.setter
```
JesseLovelace: Location type is actually a read-only field and updates to it are ignored by the back-end, should it have a setter here?

tseaver: The setter provides the only way to create a new bucket with a non-default `location_type`, e.g.:

```python
from google.cloud.storage.bucket import Bucket

bucket = client.bucket("<BUCKET_NAME_HERE>")
bucket.location_type = Bucket.DUAL_REGION_LOCATION_TYPE
bucket.create()
```

tseaver: Do we need to add `location_type` as a parameter for `Bucket.create`, and just remove the `location_type` setter?  For the `location` property, we left the setter but emit a deprecation warning, suggesting to use the `location` parameter to `Bucket.create`, but that was for backward-compatibility.

JesseLovelace: The field is actually ignored by the backend even on create

tseaver: `location_type` setter dropped in 8be3cb1.

JesseLovelace:
Thanks!

Backend rollout is complete, this can be merged and released

PR #8604, opened 2019-07-03
refactor(storage): move 'create_bucket' implementation from 'Bucket' to 'Client'

IlyaFaer:
Moved `Bucket.create()` implementation into `Client.create_bucket()`
Towards: #7762


storage/google/cloud/storage/bucket.py
```diff
@@ -437,6 +437,17 @@ def user_project(self):
...
+    @property
+    def properties(self):
+        """Properties of the current bucket object.
+
+        Returns:
+            dict: Bucket properties transformed to dict.
+        """
+        properties = {key: self._properties[key] for key in self._changes}
+        properties["name"] = self.name
+        return properties
```
IlyaFaer: Moved some lines from Bucket.create() to property for public accessing

storage/tests/unit/test_bucket.py
```diff
@@ -18,6 +18,15 @@
...
 
 
+def _make_connection(*responses):
+    import google.cloud.storage._http
+
+    mock_connection = mock.create_autospec(google.cloud.storage._http.Connection)
+    mock_connection.user_agent = "testing 1.2.3"
+    mock_connection.api_request.side_effect = list(responses)
+    return mock_connection
+
```
IlyaFaer: With this PR diffs we are no more able to use `_Client()` class in some tests. I've rewrote them with connection mocks and `Client` objects

storage/google/cloud/storage/bucket.py
tswast: I'd prefer we didn't introduce this as a public property. It's okay for the Client class to access "private" `_properties` since it's in the same library.

storage/tests/unit/test_bucket.py
```diff
@@ -535,66 +544,109 @@ def test_create_w_user_project(self):
...
+        connection = _make_connection()
+        client = Client(project=None)
 
-        with self.assertRaises(ValueError):
-            bucket.create()
+        with mock.patch(
+            'google.cloud.storage.client.Client._connection',
+            new_callable=mock.PropertyMock
+        ) as client_mock:
+            client_mock.return_value = connection
```
tswast: Why not just set `client._connection = connection`?

storage/google/cloud/storage/bucket.py
IlyaFaer: Hm, if so, we don't need this property at all, I think

storage/tests/unit/test_bucket.py
IlyaFaer: I don't remember really, but looks like I've done it because of `_connection.setter`. You right anyway, it can be done much more easier. Rewritten

tseaver:
@IlyaFaer Please use commit messages and PR titles which describe the actual changes to the code, rather than referencing a larger, ovararching project like "Client redesign proposal".

The "big issue" thing should likely be an issue (still with a more descriptive name) with checkboxes tied to the individual PRs implementing each step.  The PRs would then reference that issue in their descriptions.

frankyn:
Acking, and will review tomorrow. Thank you for your patience!

storage/google/cloud/storage/client.py
```diff
@@ -324,7 +326,15 @@ def lookup_bucket(self, bucket_name):
...
 
-    def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
+    def create_bucket(
+        self,
+        bucket_or_name,
+        requester_pays=None,
+        project=None,
+        location=None,
+        predefined_acl=None,
+        predefined_default_object_acl=None,
```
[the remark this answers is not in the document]
IlyaFaer: Pushed (if I'm following your comment right)

storage/google/cloud/storage/client.py
```diff
@@ -372,11 +396,47 @@ def create_bucket(self, bucket_or_name, requester_pays=None, project=None):
             >>> bucket = client.create_bucket(bucket)  # API request.
 
         """
+        if user_project is not None:
+            raise ValueError("Cannot create bucket with 'user_project' set.")
```
frankyn: This shouldn't be the case. You should be able to create a bucket with user_project set, now at least. It might be technical debt from Requester Pays feature implementation.

IlyaFaer: As I understand, buckets/insert only uses `user_project` as query parameter, so we don't do `bucket.user_project=user_project` while creating new bucket.

Coverage failed, I assume, because of skips. I've only moved two lines with last commit, this change can't be the cause of 10 coverage issues, I believe.

frankyn:
Pulling out coverage report:

```
Creating virtual environment (virtualenv) using python3.6 in .nox/cover
pip install coverage pytest-cov
coverage report --show-missing --fail-under=100
Name                                   Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------------
google/cloud/storage/__init__.py           8      0      0      0   100%
google/cloud/storage/_helpers.py          74      0     12      0   100%
google/cloud/storage/_http.py             15      0      2      0   100%
google/cloud/storage/_signing.py         145      0     70      0   100%
google/cloud/storage/acl.py              171      0     40      0   100%
google/cloud/storage/batch.py            127      0     28      0   100%
google/cloud/storage/blob.py             441      0    134      0   100%
google/cloud/storage/bucket.py           529      0    166      0   100%
google/cloud/storage/client.py           163      6     56      4    95%   405, 413-414, 417-420, 429, 404->405, 412->413, 416->417, 428->429
```

After moving in the logic in Bucket.create() to Client.create_bucket(). The conditional branches in client.py are not being covered by unit tests in a similar way that bucket.py is:  https://github.com/googleapis/google-cloud-python/blob/e206be2b5dada7565b1a4f6a6354d492e39dd47a/storage/tests/unit/test_bucket.py#L610-L819

I'm wondering, given you've moved over implementation from Bucket to Client, the same should be done with unit tests. I also don't see the deletion of implementation from Bucket after looking at the coverage issue.

IlyaFaer:
@frankyn, that's strange a little. Bucket tests are still running, and they execute code lines in `Client.create_bucket()` - coverage just uses data about executed lines, no matter the tests, as I know. But okay, I've copied tests to client, got 100%. `bucket.create()` is still working, so I'd say we should keep all of the bucket tests until method movement completion (I'm gonna add deprecations for all methods, which were moved in next PR), WDYT?

IlyaFaer:
> We will need to update samples to use new methods before merging/releasing the deprecation.

No problem, I had added comment into original issue not to forget.

PR #8606, opened 2019-07-03
Tasks: Fix typo in README.

averikitsch:


tseaver:
AutoML, Core, Data Catalog, Firestore, IAM, OS Login, and Resource Manager Kokoro jobs fail due to Kokoro glitch:  "Cannot check out non-mergeable pull request as if merged."

Issue #5204, opened 2018-04-18
BigQuery: to_arrow() method similar to to_dataframe()


cpcloud:
Current there's a `to_dataframe()` method that returns a pandas `DataFrame` from a query. `DataFrame`s don't efficiently support array and struct values, but `pyarrow` provides efficient support for them:

```python
In [11]: import pyarrow as pa

In [13]: a = pa.array(
    ...:     [
    ...:         {'a': [1, 2, 3]}
    ...:     ],
    ...:     pa.struct([pa.field('a', pa.list_(pa.int64()))])
    ...: )

In [14]: a
Out[14]: 
<pyarrow.lib.StructArray object at 0x7f55f8764b88>
[
  {'a': [1, 2, 3]}
]
```

A `to_arrow()` method will make it easier to efficiently support more complex types going forward in downstream libraries like [ibis](https://github.com/ibis-project/ibis).

maxim-lian:
I realize out of scope for this library, but it would awesome to have this as an export rather than JSON -> python -> arrow. 

quasi-ref https://github.com/pydata/pandas-gbq/issues/133

tswast:
Fixed in #8609 

To be released in `google-cloud-bigquery` 1.17.0 (https://github.com/googleapis/google-cloud-python/pull/8663).

tswast:
FYI: `results.to_arrow(bqstorage_client=bqstorage_client).to_pandas()` is currently the fastest way to get a pandas DataFrame from your query results. About 4 seconds from results to DataFrame for a 125 MB table.

tswast:
Looks like I added the `bqstorage_client` arg to `RowIterator`, but I missed adding it to `QueryResults`. It can still be used by calling `QueryResults.results().to_arrow(bqstorage_client=bqstorage_client)`, but we should add it to `QueryJob` for consistency.

tseaver:
@tswast Did you mean to reopen here?

tswast:
@tseaver Yep. I missed one parameter update in the PR that added this feature.

tswast:
@plamut Remaining task to close out this FR is to add the `bqstorage_client` argument to `QueryJob.to_arrow()`

https://github.com/googleapis/google-cloud-python/blob/e4cf3f4458d13ce5b0203060dc4249d3e34e80a7/bigquery/google/cloud/bigquery/job.py#L2899

See `RowIterator.to_arrow()`:

[...]

PR #8609, opened 2019-07-03
Add `to_arrow` to get a `pyarrow.Table` from query results.

tswast:
An Arrow `Table` supports a richer set of types than a pandas `DataFrame`,
and is the basis of many data analysis systems. It can be used in
conjunction with pandas through the `Table.to_pandas()` method or the
pandas extension types provided by the `fletcher` package.

Towards #5204.

### DIFF
--- a/bigquery/tests/unit/test__pandas_helpers.py
+++ b/bigquery/tests/unit/test__pandas_helpers.py
@@ -293,6 +293,51 @@ def test_bq_to_arrow_data_type_w_struct(module_under_test, bq_type):
     assert actual.equals(expected)
 
 
+@pytest.mark.parametrize("bq_type", ["RECORD", "record", "STRUCT", "struct"])
+@pytest.mark.skipIf(pyarrow is None, "Requires `pyarrow`")
+def test_bq_to_arrow_data_type_w_array_struct(module_under_test, bq_type):
+    fields = (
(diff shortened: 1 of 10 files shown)


bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -293,6 +293,51 @@ def test_bq_to_arrow_data_type_w_struct(module_under_test, bq_type):
...
+    expected_value_type = pyarrow.struct(
+        (
+            pyarrow.field("field01", pyarrow.string()),
+            pyarrow.field("field02", pyarrow.binary()),
+            pyarrow.field("field03", pyarrow.int64()),
+            pyarrow.field("field04", pyarrow.int64()),
+            pyarrow.field("field05", pyarrow.float64()),
+            pyarrow.field("field06", pyarrow.float64()),
+            pyarrow.field("field07", module_under_test.pyarrow_numeric()),
+            pyarrow.field("field08", pyarrow.bool_()),
```
shollyman: Does the underscore signify anything, or just an arrow oddity in type representation?

tswast: Trailing underscore is the Python convention for avoiding name collisions with [Python built-in functions such as `bool`](https://docs.python.org/3/library/functions.html#bool)

PR #8613, opened 2019-07-05
Asset: Add VPCSC tests.

gaogaogiraffe:
The tests can be run inside or outside of VPC service perimeter. The input to the script should be the following environment variables.

PROJECT_ID: a project that is inside the VPC perimeter.
GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT: a project that is outside the VPC perimeter.

### DIFF
--- a/asset/noxfile.py
+++ b/asset/noxfile.py
@@ -118,11 +118,19 @@ def system(session):
     session.install("-e", "../test_utils/")
     session.install("-e", ".")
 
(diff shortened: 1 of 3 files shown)


tseaver:
The 'Datastore' failure appears to be a Kokoro-internal issue (I can't see the job / output).

PR #8617, opened 2019-07-08
Storage: Skip failing 'test_bpo_set_unset_preserves_acls' systest.

tseaver:
Back-end fix for the issue expected 2019-07-12.

See #8552.

### DIFF
--- a/storage/tests/system.py
+++ b/storage/tests/system.py
@@ -1683,6 +1683,7 @@ class TestIAMConfiguration(unittest.TestCase):
         with self.assertRaises(exceptions.BadRequest):
             blob_acl.save()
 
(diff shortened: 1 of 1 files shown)


storage/tests/system.py
```diff
@@ -1704,6 +1705,8 @@ def test_bpo_set_unset_preserves_acls(self):
         bucket.iam_configuration.bucket_policy_only_enabled = True
         bucket.patch()
 
+        self.assertTrue(bucket.iam_configuration.bucket_policy_only_enabled)
```
frankyn: Is this expected?

tseaver: Yes:  the bucket's properties get updated via the resource returned from the PATCH request.

frankyn: Sg, thanks. Please move forward when ready.

Issue #5388, opened 2018-05-25
Storage: Blob.exists() does not work within Batch context


ludwigschubert:
The `Blob.exists()` method does not work when run within a Batch context. The normal behavior of `exists()` is to return True unless a `NotFound` exception occurs. Within the Batch context the exception seems to be suppressed and the function returns True. After leaving the Batch context, an Exception is then thrown.

This is how I expected to be able to use the `exists()` function:

```python
blobs = [storage.blob.Blob(path, bucket) for path in paths]
with client.batch():
  bools = [blob.exists() for blob in blobs]
```

Without the `Batch` contextmanager this code works, if inefficiently. With the `Batch` contextmanager the code returns all `True`s and throws an exception when leaving the context.

This behavior seems unintuitive to me. Please let me know if the API is meant to be used differently. If it is meant to be used as in the provided code sample, I'd be happy to attempt a fix if one of the maintainers could point me in the right direction.

Environment configuration just in case:
- macOS 10.13.3
- Python 3.6.5
- google-cloud-storage==1.8.0

tseaver:
@ludwigschubert Under the covers, `Blob.exists()` makes a `GET` request to the blob's resource URL and converts a 404 response into a `False`, 20x into a `True`.  This implementation does not fit well with the current `Batch` design, which fakes the responses inside the context manager, and then applies the sub-responses to the individual targets at exit.  See the [comment in `Blob.exists`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/26f1a97ec06b5c48df7a3c9eec00ae3985657164/storage/google/cloud/storage/blob.py#L381-L383).

PR #8618, opened 2019-07-09
Storage: Blob.exists() does not work within Batch context

IlyaFaer:
My proposition is to add something like future result - boolean in our case. An object will be working as original `bool`, but under the hood it'll be checking if future response is done without exceptions

Towards: #5388


storage/google/cloud/storage/blob.py
```diff
@@ -2057,3 +2064,29 @@ def _add_query_parameters(base_url, name_value_pairs):
...
+
+    This can be used as a target object for the future request
+    (which is part of batch mechanism) to get result of this request.
+    """
+    _properties = {}
+
+    @property
+    def _is_successfull(self):
+        """Checks if binded future request have been completed without any exception."""
+        return not isinstance(self._properties, _FutureDict)
```
IlyaFaer: If exception have been raise or request is not done yet, there will be `_FutureDict` obj

storage/tests/unit/test_blob.py
```diff
@@ -20,13 +20,45 @@
...
+        content=b"""
+--batch_pK7JBAk73-E=_AA5eFwv4m2Q=
+HTTP/1.1 404 NotFound
+
+--batch_pK7JBAk73-E=_AA5eFwv4m2Q=
+HTTP/1.1 200 OK
+
+--batch_pK7JBAk73-E=_AA5eFwv4m2Q=--"""
+    )
+    return mock_conn
```
IlyaFaer: Here are two different responses for two different `exists()` calls, and one of them is NotFound

storage/tests/unit/test_blob.py
```diff
@@ -582,6 +614,27 @@ def test_exists_miss(self):
...
+
+        try:
+            with client.batch():
+                bool1 = blob1.exists()
+                bool2 = blob2.exists()
+        except NotFound:
+            pass
+
+        self.assertFalse(bool1)
+        self.assertTrue(bool2)
```
IlyaFaer: Checking if both `exists()` calls working as they suppose to: the first one must be `False` (NotFound)

mf2199:
@tseaver  @crwilcox  @frankyn A friendly reminder: This PR has been hanging out for awhile. We could use some feedback regarding its status.

frankyn:
Hi @IlyaFaer,

I'm closing this PR as it's focused on a feature request that hasn't been prioritized or has a reviewed design.

Please prioritize bugs in the client before feature requests.

Thank you for your patience.

PR #8626, opened 2019-07-09
Data Labeling: Fix links in README.rst

busunkim96:
CC @happyhuman


tseaver:
Phishing Protection, Runtime Configuration, Videointelligence jobs fail due to Kokoro-internal glitch: "Cannot check out non-mergeable pull request as if merged."

PR #8645, opened 2019-07-11
Release dns 0.30.2

busunkim96:

### DIFF
--- a/dns/setup.py
+++ b/dns/setup.py
@@ -22,7 +22,7 @@ import setuptools
 
 name = 'google-cloud-dns'
 description = 'Google Cloud DNS API client library'
-version = '0.30.1'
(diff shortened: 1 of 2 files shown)


dns/setup.py
```diff
@@ -22,7 +22,7 @@
 
 name = 'google-cloud-dns'
 description = 'Google Cloud DNS API client library'
-version = '0.30.1'
+version = '0.30.2'
```
tseaver: I wonder if the endpoint change is worth signalling with a minor release?

busunkim96: I went back to ask folks doing this in other languages what they're doing Will leave it open for now.

busunkim96:
Merging as a patch, as other languages have done the same.

PR #8650, opened 2019-07-11
Firestore: Add 'should_terminate' predicate for clean BiDi shutdown.

### LINKED ISSUE #7826: Firestore: watch unsubscribe generates unexpected exception
When issuing an unsubscribe() on an existing watch, an exception is emitted.  While python continues to function and any new on_snapshot queries work, it's not ideal to have these exceptions.

https://firebase.google.com/docs/firestore/query-data/listen#detach_a_listener

#### Environment details

1. Specify the API: Firestore
2. OS type and version:  Ubuntu 18.04 and CentOS 7 (centos-7-v20190312)
3. Python version and virtual environment information: 3.6.7 (Ubuntu) and 3.6.6 (CentOS)
4. google-cloud-<service> version:  `pip show google-<service>` or `pip freeze`
cachetools==3.1.0
certifi==2019.3.9
chardet==3.0.4
google-api-core==1.10.0
google-auth==1.6.3
google-cloud-core==0.29.1
google-cloud-firestore==1.0.0
googleapis-common-protos==1.5.10
grpcio==1.20.1
idna==2.8
protobuf==3.7.1
pyasn1==0.4.5
pyasn1-modules==0.2.5
pytz==2019.1
requests==2.21.0
rsa==4.0
six==1.12.0
urllib3==1.24.2

#### Steps to reproduce

  1. Create an on_snapshot watch of a collection.
  2. sleep to mimic work being done
  3. Invoke an unsubscribe on the watch.
  4. Exception errors are emitted to stderr.

Expected behavior is no exception.  

#### Code example

```python
import time
from google.cloud import firestore

def on_snapshot(collection_snapshot, changes, read_time):
    print("on_snapshot()")

client = firestore.Client()
collection_ref = client.collection("localaccts-testing")
watch = collection_ref.on_snapshot(on_snapshot)

while True:
    time.sleep(30)

    watch.unsubscribe()
    watch = collection_ref.on_snapshot(on_snapshot)
```

#### Example output showing the exception
```
on_snapshot()
Thread-ConsumeBidirectionalStream caught unexpected exception <_Rendezvous of RPC that terminated with:
	status = StatusCode.CANCELLED
	details = "Locally cancelled by application!"
	debug_error_string = "None"
> and will exit.
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/google/api_core/bidi.py", line 543, in _thread_main
    response = self._bidi_rpc.recv()
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/google/api_core/bidi.py", line 454, in recv
    return self._recoverable(self._recv)
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/google/api_core/bidi.py", line 413, in _recoverable
    raise exc
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/google/api_core/bidi.py", line 403, in _recoverable
    return method(*args, **kwargs)
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/google/api_core/bidi.py", line 451, in _recv
    return next(call)
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/grpc/_channel.py", line 363, in __next__
    return self._next()
  File "/home/rmceoin/testwatch/env/lib64/python3.6/site-packages/grpc/_channel.py", line 357, in _next
    raise self
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
	status = StatusCode.CANCELLED
	details = "Locally cancelled by application!"
	debug_error_string = "None"
>
on_snapshot()
```

tseaver:
Stream exceptions which pass this predicate will cause clean shutdown of the BiDi thread.

Closes #7826.

Note to reviewers: I recommend looking at this commit-by-commit.

### DIFF
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -349,6 +349,11 @@ class BidiRpc(object):
         return self._request_queue.qsize()
 
 
+def _never_terminate(future_or_error):
+    """By default, no errors cause BiDi termination."""
+    return False
+
+
 class ResumableBidiRpc(BidiRpc):
     """A :class:`BidiRpc` that can automatically resume the stream on errors.
 
@@ -391,6 +396,9 @@ class ResumableBidiRpc(BidiRpc):
         should_recover (Callable[[Exception], bool]): A function that returns
             True if the stream should be recovered. This will be called
             whenever an error is encountered on the stream.
+        should_terminate (Callable[[Exception], bool]): A function that returns
+            True if the stream should be terminated. This will be called
+            whenever an error is encountered on the stream.
         metadata Sequence[Tuple(str, str)]: RPC metadata to include in
             the request.
         throttle_reopen (bool): If ``True``, throttling will be applied to
@@ -401,12 +409,14 @@ class ResumableBidiRpc(BidiRpc):
         self,
         start_rpc,
         should_recover,
+        should_terminate=_never_terminate,
         initial_request=None,
         metadata=None,
         throttle_reopen=False,
     ):
         super(ResumableBidiRpc, self).__init__(start_rpc, initial_request, metadata)
         self._should_recover = should_recover
+        self._should_terminate = should_terminate
         self._operational_lock = threading.RLock()
         self._finalized = False
         self._finalize_lock = threading.Lock()
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -508,6 +508,12 @@ class ResumableBidiRpc(BidiRpc):
                 with self._operational_lock:
                     _LOGGER.debug("Call to retryable %r caused %s.", method, exc)
 
+                    if self._should_terminate(exc):
+                        self.close()
+                        _LOGGER.debug("Terminating %r due to %s.", method, exc)
+                        self._finalize(exc)
+                        break
+
                     if not self._should_recover(exc):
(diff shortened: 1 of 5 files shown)


tseaver:
I can also add `should_terminate` support for Pubsub, if desired (and someone can confirm the set of gRPC status codes which should cause clean shutdown).

crwilcox:
Is there a reason should_recover couldn't be used for this? Semantically, should not recover and should terminate seem the same? Do we need another argument.

Also, note, immediately following this commit, we should ship api_core and rev min versions for firestore and pubsub

tseaver:
@crwilcox See #7826.  The goal is to avoid logging an error when the stream is shut down due to "expected" events., typically due to a `grpc.StatusCode.CANCELLED` error response

- The existing `should_recover` returns True if the error is "recoverable", i.e., the stream should save any pending send and reopen itself.
- The new `should_terminate` returns True if the error should cause the stream to shut down cleanly, without reopening.

tseaver:
@plamut Maybe the patch from this PR allows a final `None` to be passed through on a cancelled stream, where previoiusly it would have errored out with the traceback from #7826?

tseaver:
In any case, if it is possible that `on_snapshot` can be called with `None` as the response, we need to handle that (by an immediate return, I think).

PR #8662, opened 2019-07-12
Bigtable: Separate row types to remove confusion around return types of row.commit

IlyaFaer:
Looks like some consensus have been achieved on [#8451](https://github.com/googleapis/google-cloud-python/issues/8451), so here is new PR


bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,88 @@ def test_bigtable_table_row():
...
+    client = Client(admin=True)
+    instance = client.instance(INSTANCE_ID)
+    table = instance.table(TABLE_ID)
+
+    row_keys = [b"row_key_1", b"row_key_2"]
+    row1_obj = table.append_row(row_keys[0])
+    row2_obj = table.append_row(row_keys[1])
+    # [END bigtable_table_append_row]
+
+    row1_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
```
tseaver: This line [fails](https://source.cloud.google.com/results/invocations/4119c229-0ca1-4c2e-a19b-74ec4738bea9/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable/log) with:

```python
    def test_bigtable_table_append_row():
        # [START bigtable_table_append_row]
        from google.cloud.bigtable import Client
        client = Client(admin=True)
        instance = client.instance(INSTANCE_ID)
        table = instance.table(TABLE_ID)
        row_keys = [b"row_key_1", b"row_key_2"]
        row1_obj = table.append_row(row_keys[0])
        row2_obj = table.append_row(row_keys[1])
        # [END bigtable_table_append_row]
>       row1_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
E       AttributeError: 'AppendRow' object has no attribute 'set_cell'
```

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,88 @@ def test_bigtable_table_row():
...
+def test_bigtable_table_conditional_row():
+    # [START bigtable_table_conditional_row]
+    from google.cloud.bigtable import Client
+
+    client = Client(admin=True)
+    instance = client.instance(INSTANCE_ID)
+    table = instance.table(TABLE_ID)
+
+    row_keys = [b"row_key_1", b"row_key_2"]
+    filter_ = object()
```
tseaver: The test requires a real filter here.  It fails in `commit` with:

```python
    def test_bigtable_table_conditional_row():
        # [START bigtable_table_conditional_row]
        from google.cloud.bigtable import Client
        client = Client(admin=True)
        instance = client.instance(INSTANCE_ID)
        table = instance.table(TABLE_ID)
        row_keys = [b"row_key_1", b"row_key_2"]
        filter_ = object()
        row1_obj = table.conditional_row(row_keys[0], filter_=filter_)
        row2_obj = table.conditional_row(row_keys[1], filter_=filter_)
        # [END bigtable_table_conditional_row]
        row1_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
>       row1_obj.commit()
docs/snippets_table.py:484:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <google.cloud.bigtable.row.ConditionalRow object at 0x7ff5540fcda0>
    def commit(self):
        """Makes a ``CheckAndMutateRow`` API request.
        If no mutations have been created in the row, no request is made.
        The mutations will be applied conditionally, based on whether the
        filter matches any cells in the :class:`ConditionalRow` or not. (Each
        method which adds a mutation has a ``state`` parameter for this
        purpose.)
        Mutations are applied atomically and in order, meaning that earlier
        mutations can be masked / negated by later ones. Cells already present
        in the row are left unchanged unless explicitly changed by a mutation.
        After committing the accumulated mutations, resets the local
        mutations.
        For example:
        .. literalinclude:: snippets_table.py
            :start-after: [START bigtable_row_commit]
            :end-before: [END bigtable_row_commit]
        :rtype: bool
        :returns: Flag indicating if the filter was matched (which also
                  indicates which set of mutations were applied by the server).
        :raises: :class:`ValueError <exceptions.ValueError>` if the number of
                 mutations exceeds the :data:`MAX_MUTATIONS`.
        """
        true_mutations = self._get_mutations(state=True)
        false_mutations = self._get_mutations(state=False)
        num_true_mutations = len(true_mutations)
        num_false_mutations = len(false_mutations)
        if num_true_mutations == 0 and num_false_mutations == 0:
            return
        if num_true_mutations > MAX_MUTATIONS or num_false_mutations > MAX_MUTATIONS:
            raise ValueError(
                "Exceed the maximum allowable mutations (%d). Had %s true "
                "mutations and %d false mutations."
                % (MAX_MUTATIONS, num_true_mutations, num_false_mutations)
            )
        data_client = self._table._instance._client.table_data_client
        resp = data_client.check_and_mutate_row(
            table_name=self._table.name,
            row_key=self._row_key,
>           predicate_filter=self._filter.to_pb(),
            true_mutations=true_mutations,
            false_mutations=false_mutations,
        )
E       AttributeError: 'object' object has no attribute 'to_pb'
```

bigtable/google/cloud/bigtable/table.py
```diff
@@ -199,6 +207,60 @@ def row(self, row_key, filter_=None, append=False):
...
+        For example:
+
+        .. literalinclude:: snippets_table.py
+            :start-after: [START bigtable_table_conditional_row]
+            :end-before: [END bigtable_table_conditional_row]
+
+        Args:
+            row_key (bytes): The key for the row being created.
+
+            filter_ (:class:`.RowFilter`): (Optional) Filter to be used for
```
crwilcox: Do we need to keep the `_` on `filter_`?

bigtable/google/cloud/bigtable/table.py
```diff
@@ -199,6 +207,60 @@ def row(self, row_key, filter_=None, append=False):
         else:
             return DirectRow(row_key, self)
 
+    def append_row(self, row_key):
+        """Create a :class:`~google.cloud.bigtable.row.AppendRow` associated with this table.
```
crwilcox: We should make sure these comments reflect why you would choose one of the rows over another.

crwilcox:
@billyjacobson can you take a look and see if you think this improves the usability of the python client?

bigtable/docs/snippets_table.py
IlyaFaer: It's fixed. I'm not very sure about these two calls:
https://github.com/googleapis/google-cloud-python/blob/fef10c08f71c362cb63c9e588d9d39f3dcd4a878/bigtable/noxfile.py#L174-L184

If first call will fail, the second one (for `snippets_table.py`) will be just skipped. It can be a little bit inconvenient: in some cases you'll see only first check fails, without second check results at all. It's better to see whole picture from the very beginning, I think. Still, I understand, why session written that way

bigtable/google/cloud/bigtable/table.py
IlyaFaer: Of course. `filter` is a [built-in function](https://docs.python.org/3/library/functions.html#filter), if we'll erase `_`, we'll override this function, that is completely sin. It's Python naming standard

bigtable/google/cloud/bigtable/table.py
IlyaFaer: Not sure, I'm getting. You want something like: you read method's docs, and you don't need to read row class'es docs, 'cause method's docs will tell you everything you need? So classes should be described a little in method's docs?

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,89 @@ def test_bigtable_table_row():
...
+    row1_obj = table.direct_row(row_keys[0])
+    row2_obj = table.direct_row(row_keys[1])
+    # [END bigtable_table_direct_row]
+
+    row1_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
+    row1_obj.commit()
+    row2_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
+    row2_obj.commit()
+
+    actual_rows_keys = []
```
billyjacobson: written_row_keys

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,89 @@ def test_bigtable_table_row():
...
+    instance = client.instance(INSTANCE_ID)
+    table = instance.table(TABLE_ID)
+
+    row_keys = [b"row_key_1", b"row_key_2"]
+    filter_ = PassAllFilter(True)
+    row1_obj = table.conditional_row(row_keys[0], filter_=filter_)
+    row2_obj = table.conditional_row(row_keys[1], filter_=filter_)
+    # [END bigtable_table_conditional_row]
+
+    row1_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1, state=False)
```
billyjacobson: @crwilcox I feel like instead of state, we could do something like onMatch=False or onMatch=True

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,89 @@ def test_bigtable_table_row():
...
+    row1_obj.commit()
+    row2_obj.set_cell(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1, state=False)
+    row2_obj.commit()
+
+    actual_rows_keys = []
```
billyjacobson: written_row_keys

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,89 @@ def test_bigtable_table_row():
     table.truncate(timeout=300)
 
 
+def test_bigtable_table_append_row():
+    # [START bigtable_table_append_row]
```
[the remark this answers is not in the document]
billyjacobson: Ah, I saw them when I scrolled. I always get confused when the API docs use the same include code format as the product docs

bigtable/docs/snippets_table.py
```diff
@@ -412,6 +412,89 @@ def test_bigtable_table_row():
...
+    row1_obj = table.append_row(row_keys[0])
+    row2_obj = table.append_row(row_keys[1])
+    # [END bigtable_table_append_row]
+
+    row1_obj.append_cell_value(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
+    row1_obj.commit()
+    row2_obj.append_cell_value(COLUMN_FAMILY_ID, COL_NAME1, CELL_VAL1)
+    row2_obj.commit()
+
+    actual_rows_keys = []
```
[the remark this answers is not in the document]
IlyaFaer: Ok, renamed all such variables. This name was already in file

bigtable/google/cloud/bigtable/table.py
crwilcox: I thought it may be a reserved name at this point, I just wanted to make sure this wasn't a leftover from a previous iteration. Thanks @IlyaFaer

bigtable/docs/snippets_table.py
crwilcox: @billyjacobson that seems reasonable to me. Does that resemble something we do in other languages? I agree `state` isn't a great name, but it would be good if we didn't diverge from a cross-language pattern if there is one.

billyjacobson: Node calls it onMatch and onNoMatch.
The RPCs call it true_mutations and false_mutations
Go uses mtrue, mfalse
PHP trueMutations, falseMutations
Ruby true_mutations, false_mutations

So probably true and false mutations would be most consistent

tseaver: We can't just rename the `state` parameter:  it would be a breaking change.

billyjacobson: Ok

PR #8665, opened 2019-07-12
BigQuery Storage: Update quickstart sample with data format and sharding options.

tswast:
* Moves sample to `samples/` directory.
* Updates `synth.py` to use common templates (which now includes samples tests in the noxfile).


tswast:
Comments copied from @shollyman's https://github.com/GoogleCloudPlatform/golang-samples/pull/906.

bigquery_storage/setup.cfg
```diff
@@ -1,10 +1,3 @@
...
 [bdist_wheel]
 universal = 1
-
-[flake8]
-exclude =
-    *_pb2.py
-    *_pb2_grpc.py
-    google/cloud/bigquery_storage_v1beta1/gapic/big_query_storage_client.py
-    google/cloud/bigquery_storage_v1beta1/gapic/transports/big_query_storage_grpc_transport.py
-    tests/unit/gapic/v1beta1/test_big_query_storage_client_v1beta1.py
```
tseaver: Is this change intentional?  Seems like it might break stuff.

bigquery_storage/synth.py
```diff
@@ -130,4 +130,10 @@
 )
 # END: Ignore lint and coverage
 
+# ----------------------------------------------------------------------------
+# Add templated files
+# ----------------------------------------------------------------------------
+templated_files = common.py_library(unit_cov_level=79, cov_level=79, samples_test=True, system_test_dependencies=["fastavro", "pandas", "pyarrow != 0.14.0"])
+s.move(templated_files)
```
tseaver: Might need one or more excludes here.

bigquery_storage/setup.cfg
tswast: Linter is passing. Seems these lines are no longer needed.

bigquery_storage/synth.py
tswast: I'd rather keep the common templates if possible. It seems to be working.

PR #8666, opened 2019-07-12
Videointelligence: Remove face detection feature from V1 client (via synth).

tseaver:
Closes #8661.

PR #8607, opened 2019-07-03
Videointelligence: Add VPC SC integration tests.

mkudejim:

PR #8669, opened 2019-07-12
Revert "Videointelligence: Remove face detection feature from V1 client (via synth)."

busunkim96:
Reverts googleapis/google-cloud-python#8666

I've been told folks working with the Video Intelligence team that this change is likely to get rolled back.

### DIFF
--- a/videointelligence/synth.py
+++ b/videointelligence/synth.py
@@ -66,6 +66,6 @@ s.replace(
 # Add templated files
 # ----------------------------------------------------------------------------
 templated_files = common.py_library(unit_cov_level=97, cov_level=100)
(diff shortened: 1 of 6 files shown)


videointelligence/synth.py
```diff
@@ -66,6 +66,6 @@
 # Add templated files
 # ----------------------------------------------------------------------------
 templated_files = common.py_library(unit_cov_level=97, cov_level=100)
-s.move(templated_files, excludes="noxfile.py")
+s.move(templated_files)
```
tseaver: This part of #8666 probably needs to stay, in order not to have the next autogen undo #8607.

PR #8677, opened 2019-07-15
PubSub: Skip two IAM tests for VPCSC

anguillanneuf:
Make `test_managing_topic_iam_policy` and `test_managing_subscription_iam_policy` VPCSC-compatible by excluding them when the environment variable GOOGLE_CLOUD_TESTS_IN_VPCSC is set to "true".

Reference: https://github.com/googleapis/google-cloud-python/pull/6215

### DIFF
--- a/pubsub/tests/system.py
+++ b/pubsub/tests/system.py
@@ -36,7 +36,7 @@ from google.cloud.pubsub_v1 import types
 from test_utils.system import unique_resource_id
 
 
-RUNNING_IN_VPCSC = os.getenv('GOOGLE_CLOUD_TESTS_IN_VPCSC', '').lower() == 'true'
+RUNNING_IN_VPCSC = os.getenv("GOOGLE_CLOUD_TESTS_IN_VPCSC", "").lower() == "true"
 
 
 @pytest.fixture(scope=u"module")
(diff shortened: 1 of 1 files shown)


anguillanneuf:
Thanks Peter for reviewing! 

@bmccutchon may be able to elaborate that better. Brian?

bmccutchon:
[VPC Service Controls](https://cloud.google.com/vpc-service-controls/) lets you create secure perimeters around projects to prevent data exfiltration. To verify that these client libraries are compatible with it, we need to be able to run the tests in a secure VPCSC environment. When run in this environment, tests can't access resources in other projects. So, in short, tests that access resources in projects separate from the one they run on&mdash;e.g. public datasets&mdash;need to be skipped or modified to be compatible with VPCSC.

That said, these tests don't access resources in other projects, so we probably shouldn't skip them. Instead, we discovered that these tests were actually failing due to another issue that we are able to fix.

plamut:
@bmccutchon Thanks for the explanation!

@anguillanneuf Since the root cause of the failures can be fixed elsewhere, I presume we do not need to skip the tests, unless it's semi-urgent and we still want to disable them temporarily?

If so, please add a comment to the source and open a GitHub issue so that we don't forget to revert the changes later. Or, if we can live with the current state for a bit longer, we should probably close this PR without merging.

anguillanneuf:
Thanks @bmccutchon and @plamut! I will close this PR for now!

PR #8683, opened 2019-07-16
Bigtable: Add Cell.__repr__

joar:
To improve the developer experience when printing `Cell` instances.

### DIFF
--- a/bigtable/google/cloud/bigtable/row_data.py
+++ b/bigtable/google/cloud/bigtable/row_data.py
@@ -85,12 +85,10 @@ class Cell(object):
 
     def __ne__(self, other):
         return not self == other
-    
+
     def __repr__(self):
         return "<{name} value={value!r} timestamp={timestamp}>".format(
-            name=self.__class__.__name__,
-            value=self.value,
-            timestamp=self.timestamp
+            name=self.__class__.__name__, value=self.value, timestamp=self.timestamp
         )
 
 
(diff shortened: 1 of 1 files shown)


bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -85,6 +85,13 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return not self == other
+
```
tseaver: There is a stray space character here which is causing lint / black failures.

bigtable/google/cloud/bigtable/row_data.py
```diff
@@ -85,6 +85,13 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return not self == other
+    
+    def __repr__(self):
+        return "<{name} value={value!r} timestamp={timestamp}>".format(
+            name=self.__class__.__name__,
+            value=self.value,
+            timestamp=self.timestamp
+        )
```
tseaver: black wants to format this all on one line, e.g.:

```python
    def __repr__(self):
        return "<{name} value={value!r} timestamp={timestamp}>".format(
            name=self.__class__.__name__, value=self.value, timestamp=self.timestamp
        )
```

joar: I made this PR in github's code editor, shame on me! I believe I've fixed both of the linting/black issues now, but I'm still only in github's code edtior, so I might be proven wrong!

PR #8687, opened 2019-07-16
Firestore: Treat 'None' as EOF in 'Watch.on_snapshot'.

tseaver:
See:
https://github.com/googleapis/google-cloud-python/pull/8650#pullrequestreview-261830931

### DIFF
--- a/firestore/google/cloud/firestore_v1/watch.py
+++ b/firestore/google/cloud/firestore_v1/watch.py
@@ -426,7 +426,12 @@ class Watch(object):
             TargetChange.CURRENT: self._on_snapshot_target_change_current,
         }
 
-        target_change = proto.target_change
+        target_change = getattr(proto, 'target_change', '')
+        document_change = getattr(proto, 'document_change', '')
+        document_delete = getattr(proto, 'document_delete', '')
+        document_remove = getattr(proto, 'document_remove', '')
+        filter_ = getattr(proto, 'filter', '')
+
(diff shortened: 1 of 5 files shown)


plamut:
Something weird is going on, I removed and re-added the force run label, the label is there, but there is no record of that on the PR timeline, and the `yoshi-kokoro` bot does not see it.

crwilcox:
I cleaned out firestore to see if that helps. forcing a ci run.

tseaver:
The Kokoro jobs for Bigtable and Natural Language both fail due to a Kokoro-internal glitch:

> Cannot check out non-mergeable pull request as if merged.

plamut:
@tseaver Is there anything we can do to get around the glitch?

tseaver:
@plamut 

> Is there anything we can do to get around the glitch?

I'm actually worried that there is something more sinister at play.

@crwilcox Can you double-check that the clean-out really worked in the CI project?  E.g.:

```python
documents = []
for collection in client.collections():
    documents.extend(collection.list_documents()
assert len(documents) == 0
```

tseaver:
I have added a commit to make the system test output verbose, so that we can figure out which test is hanging.

tseaver:
@crwilcox I'm expecting this Kokoro run to time out again:  can you post the hung systest output when it does?  Or maybe you have the keys to see the output even before it times out?

Issue #7215, opened 2019-01-29
Firestore: Out of expect when document update SERVER_TIMESTAMP and an arbitrary value by field path


fishballLin:
#### Environment details

1. OS type and version:
Both happened on Windows 7 and Ubuntu 18.04
2. Python version:
python 3.6
3. google-cloud-<service> version:
firebase-admin                     2.14.0
google-api-core                    1.7.0
google-auth                        1.6.2
google-cloud-core                  0.29.1
google-cloud-firestore             0.31.0
google-cloud-storage               1.13.2
google-resumable-media             0.3.2
googleapis-common-protos           1.5.5

#### Reference
The following case is modified by the [firestore document update section](https://googleapis.github.io/google-cloud-python/latest/firestore/document.html#google.cloud.firestore_v1beta1.document.DocumentReference.update) example.

#### Steps to reproduce

1. To illustrate this, consider a document with
```
>>> snapshot = document.get()
>>> snapshot.to_dict()
{
    'foo': {
        'bar': 'baz',
    },
    'other': True,
}
```
stored on the server.

2. To set a field to the current time and another field to an arbitrary value(ex: "test"). We expect sending
```
>>> field_updates = {
...     'foo.now': firestore.SERVER_TIMESTAMP,
...     'foo.test': "test",
... }
>>> document.update(field_updates)
```
would update the value on the server to:
```
>>> snapshot = document.get()
>>> snapshot.to_dict()
{
    'foo': {
        'bar': 'baz',
        'now': datetime.datetime(2019, ...),
        'test': 'test'
    },
    'other': True,
}
```
3. However, in the real case, the value on the server are:
 ```
>>> snapshot = document.get()
>>> snapshot.to_dict()
{
    'foo': {
        'now': datetime.datetime(2019, ...),
        'test': 'test'
    },
    'other': True,
}
```
4. Apparently, the previous field 'foo.bar' disappear, but it doesn't match our expect.

##### Step 1: prepare the data
```python
import firebase_admin
from firebase_admin import credentials, firestore

cred = credentials.Certificate('your-api-key.json')
firebase_admin.initialize_app(cred)

db = firestore.client()
document = db.collection('test').document()
document.set({
    "foo":{
        "bar": 'baz'
    },
    'other': True
})
snapshot = document.get()
print(snapshot.to_dict())
```
console:
```
{
    'foo': {
        'bar': 'baz',
    },
    'other': True,
}
```

##### Step 2: update the data
```python
[...]

PR #8689, opened 2019-07-16
Firestore: Update conformance tests to match new repo / format.

tseaver:
Apologies to reviewers for the size of the PR.  The only "interesting" changes are in these two files:

- `Makefile_v1`: used to generate the file we use to parse testcases.
- `test_cross_language.py`: test driver which uses the canonical testcase data files.

The rest are either the `protoc`-generated file for the new `TestFile` message format, or the individual testcase datafiles themselves, in their new JSON formats (with the old `.textproto` / `.binproto` files removed).

I specifically avoided copying in the *new* testcase file, `update-nested-transform-and-nested-value.json`,
because it is the reproducer for the not-yet-fixed bug, #7215.


tseaver:
Error Reporting and Text-to-Speech jobs both fail due to Kokoro-internal glitch: "Cannot check out non-mergeable pull request as if merged."

tseaver:
Testing only. I need this merged to start work on #7215 (the reproducer conformance test is new-style JSON only).  Merging without review.

Issue #5879, opened 2018-09-01
Cloud Functions & Storage: fails intermittently with ProtocolError + ConnectionResetError


brianmhunt:
This is a cross post originally detailed at https://issuetracker.google.com/issues/113672049 

Essentially the problem is that in a Google Cloud Functions python endpoint the `google-cloud-storage` API is intermittently throwing a `ProtocolError` and `ConnectionResetError` when getting a blob.

brianmhunt:
Linking possibly relevant Golang issue: GoogleCloudPlatform/google-cloud-go#108

frankyn:
I'm not sure what's happening here, acking for now in GCS library weekly.

frankyn:
I did not get additional input in the weekly meeting.

@brianmhunt could you tell me more about your use-case so I can try to reproduce? 
Also what is the size of a file you're trying to read?

brianmhunt:
@frankyn Of course, thanks for the follow-up.

Our function is given an array of slices of PDFs and converts that to a single PDF.  An oversimplified version is something like this:

```python
def consolidatePdf(pdfSlices):
    """Create a PDF from the given slices.

    pdfSlices is an iterable of dicts of this form: { url: "gs://", range: [start, end] }
    """
    newPdf = PdfWriter()
    for slice in pdfSlices:
        reader = PdfReader(getPdfFromGS(slice['url']))
        start, end = slice['range']
        pdfWriter.append(reader.getPages(start, end))
    return newPdf.getBytes()
```

Where the `getPdfFromGS` performs the blob storage read.  If you think it'll help I'm happy to share the code if you email me at `brianmhunt at gmail.com`.

I've only seen the failure occur on the very first file being read (but that doesn't mean that the problem is limited to the first file being read).

The files failing are fairly small, in the 150k range.

frankyn:
Thanks, I'd like to keep this discussion as much as possible through Github.  So if someone else hits a similar issue they can find it later.

Is PdfReader wrapping around the google-cloud-storage package? Could you share a portion of that code?

brianmhunt:
Thanks, here's the salient bit of code that's throwing.

```python
def generateUrlPdfBytesMap():
    upm = dict()
    for url in urls:
        bucket, path = urlToBucketPath(url)
        account = path.split('/')[1]
        blob = storage.bucket(bucket).get_blob(path) # 🔥
        upm[url] = blob.download_as_string()
    return upm
```

[...]

PR #8691, opened 2019-07-17
Storage: Exponential backoff for ConnectionResetError

sangramql:
Issue #5879 
Cloud Functions & Storage: fails intermittently with ProtocolError + ConnectionResetError

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -761,16 +761,16 @@ class Bucket(_PropertyMixin):
             generation=generation,
             **kwargs
         )
-        for retry in range(self.RETRIES):
+        for retry in range(self.RETRIES): # pragma: no branch
             try:
                 # NOTE: This will not fail immediately in a batch. However, when
                 #       Batch.finish() is called, the resulting `NotFound` will be
                 #       raised.
                 blob.reload(client=client)
                 break
-            except Exception:
+            except Exception as e:
                 if retry == self.RETRIES - 1:
-                    return None
+                    raise e
                 else:
                     rand_num = random.random()
                     time.sleep(min(rand_num + 2 ** (retry - 1), rand_num + 32))
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -459,6 +459,9 @@ class Bucket(_PropertyMixin):
     )
     """Allowed values for :attr:`location_type`."""
 
+    RETRIES = 4
+    """Number of retries on connection failure. """
+
     def __init__(self, client, name=None, user_project=None):
         name = _validate_name(name)
         super(Bucket, self).__init__(name=name)
@@ -748,6 +751,9 @@ class Bucket(_PropertyMixin):
         :rtype: :class:`google.cloud.storage.blob.Blob` or None
         :returns: The blob object if it exists, otherwise None.
         """
+        import time
+        import random
+
         blob = Blob(
             bucket=self,
             name=blob_name,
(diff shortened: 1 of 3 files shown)


tseaver:
@sangramql Adding broad retry support is a big project, which needs a full-on design.  Please don't tackle bugs piecemeal for that subject.

sangramql:
@tseaver @crwilcox @frankyn 
Please review

sangramql:
Please review @tseaver @crwilcox @frankyn

storage/google/cloud/storage/bucket.py
```diff
@@ -459,6 +459,9 @@ class Bucket(_PropertyMixin):
     )
     """Allowed values for :attr:`location_type`."""
 
+    RETRIES = 4
```
frankyn: How is this number determined?

storage/google/cloud/storage/bucket.py
```diff
@@ -748,22 +751,30 @@ def get_blob(
...
-        try:
-            # NOTE: This will not fail immediately in a batch. However, when
-            #       Batch.finish() is called, the resulting `NotFound` will be
-            #       raised.
-            blob.reload(client=client)
-        except NotFound:
-            return None
-        else:
-            return blob
+        for retry in range(self.RETRIES):  # pragma: no branch
```
frankyn: Start from 1 and remove `-1` from `self.RETRIES - 1`.

`range(1, self.RETRIES)`

storage/google/cloud/storage/bucket.py
```diff
@@ -748,22 +751,30 @@ def get_blob(
         :rtype: :class:`google.cloud.storage.blob.Blob` or None
         :returns: The blob object if it exists, otherwise None.
         """
+        import time
```
frankyn: Imports at the top.

storage/google/cloud/storage/bucket.py
```diff
@@ -748,22 +751,30 @@ def get_blob(
...
+                #       Batch.finish() is called, the resulting `NotFound` will be
+                #       raised.
+                blob.reload(client=client)
+                break
+            except Exception as e:
+                if retry == self.RETRIES - 1:
+                    raise e
+                else:
+                    rand_num = random.random()
+                    time.sleep(min(rand_num + 2 ** (retry - 1), rand_num + 32))
```
frankyn: @tseaver is there a core library that can be used to replace this inline exponential backoff? gax-go has exponential backoff in its own library.

Could backoff be introduced as a decorator for retryable errors?

storage/google/cloud/storage/bucket.py
```diff
@@ -748,22 +751,30 @@ def get_blob(
...
-            return None
-        else:
-            return blob
+        for retry in range(self.RETRIES):  # pragma: no branch
+            try:
+                # NOTE: This will not fail immediately in a batch. However, when
+                #       Batch.finish() is called, the resulting `NotFound` will be
+                #       raised.
+                blob.reload(client=client)
+                break
```
frankyn: This doesn't seem necessary.

storage/google/cloud/storage/bucket.py
```diff
@@ -748,22 +751,30 @@ def get_blob(
...
+            except Exception as e:
```
frankyn: At the moment this will ignore NotFound entirely as a fast fail and continue retrying until backoff expires.

Please don't retry 404.

storage/google/cloud/storage/bucket.py
sangramql: If we start range from 1, then there would be 1 less retry. In this case it would be 3 instead of 4 retries.

`if retry == self.RETRIES - 1:`
This is added to check the last retry, where we need to actually raise the exception instead of waiting.

But yes, I have removed "retry - 1" from below code as it was actually adding only 0.5 secs instead of 1, since 2**(0-1) was 0.5. Good catch.
`time.sleep(min(rand_num + 2 ** (retry - 1), rand_num + 32))`
replaced it with
`time.sleep(min(rand_num + 2 ** (retry), rand_num + 32))`

storage/google/cloud/storage/bucket.py
sangramql: This number just added to as get it confirmed during the review.
Though I have seen this used in existing code while i was investigating the issue.

4 retries would result is actual wait of 1.x, 2.x, 4.x and 8.x seconds wait on multiple retries respectively. 
x is random milliseconds added during each iteration.

storage/google/cloud/storage/bucket.py
sangramql: Done.

storage/google/cloud/storage/bucket.py
sangramql: Yes, removed break.

tseaver:
@sangramql Please don't tackle adding retries to GCS on a piecemeal basis.

PR #8693, opened 2019-07-17
BigQuery: Complete the to_arrow() method feature request

### LINKED ISSUE #5204: BigQuery: to_arrow() method similar to to_dataframe()

plamut:
Closes #5204.

This PR adds the [last missing piece](https://github.com/googleapis/google-cloud-python/issues/5204#issuecomment-511906029) of the feature, i.e. the `bqstorage_client` parameter that was still missing from one of the methods.

### How to test
Check that the signatures for various `to_arrow()` methods match (and the same for the `to_dataframe()` methods).

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2896,6 +2896,8 @@ class QueryJob(_AsyncJob):
         rows._preserve_order = _contains_order_by(self.query)
         return rows
 
+    # If changing the signature of this method, make sure to apply the same
+    # changes to table.RowIterator.to_arrow()
     def to_arrow(self, progress_bar_type=None):
         """[Beta] Create a class:`pyarrow.Table` by loading all pages of a
         table or query.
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2896,7 +2896,9 @@ def result(self, timeout=None, page_size=None, retry=DEFAULT_RETRY):
         rows._preserve_order = _contains_order_by(self.query)
         return rows
 
-    def to_arrow(self, progress_bar_type=None):
+    # If changing the signature of this method, make sure to apply the same
+    # changes to table.RowIterator.to_arrow()
```
plamut: Note - this and the other similar comments might be redundant now that we have a test for method signatures, but I personally don't mind having them, as it makes the contract between the two classes more explicit.

PR #8707, opened 2019-07-18
Contributing document is out of date

Kami:
Contributing document is out of date and references ``system_tests/data/index.yaml`` file with index definitions which doesn't exist (see https://github.com/googleapis/google-cloud-python/pull/8636#issuecomment-512821158 for context /  details).

In addition to that, it uses old and deprecated command for creating an index which is not available anymore in recent versions of ``gcloud`` CLI.


CONTRIBUTING.rst
```diff
@@ -200,7 +200,7 @@ Running System Tests
    > --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
 
    # Create the indexes
-   $ gcloud datastore create-indexes system_tests/data/index.yaml
+   $ gcloud datastore indexes create system_tests/data/index.yaml
```
Kami: As per my comment here - https://github.com/googleapis/google-cloud-python/pull/8636#issuecomment-512821158

Where should the index definitions which are used for system / integration tests go?

I checked / grepped the repo and I couldn't find any file which contains index definitions so I added a new one - https://github.com/googleapis/google-cloud-python/commit/3902b02c3016a910b89f28a8a1b4de185477a45f.

busunkim96: @tseaver Do you know how these indexes are created?

tseaver: PR #3146 zapped the `system_tests/data/index.yaml` file.  I would put it in `datastore/tests/system/index.yaml`.  Content should be:

```yaml
indexes:

- kind: Character
  ancestor: yes
  properties:
  - name: appearances

- kind: Character
  ancestor: yes
  properties:
  - name: alive

- kind: Character
  ancestor: yes
  properties:
  - name: family
  - name: appearances

- kind: Character
  ancestor: yes
  properties:
  - name: name
  - name: family
```

Kami: Thanks.

I made that change in e69a5e06a6b9acaa8f405f6c541ccd6771484904.

I will also update my other PR (#8636) so it uses that index file location.

Issue #8686, opened 2019-07-16
Update pin for 'googleapis-common-protos'.


tseaver:
From #8674.

Recent autogen changes have begun using features present only in `googleapis-common-protos 1.6.0+`, specifially the `google.api.client` package.  E.g.:

```bash
$ git grep -l "google\/api\/client\.proto" | grep "\.proto$"
automl/google/cloud/automl_v1beta1/proto/prediction_service.proto
automl/google/cloud/automl_v1beta1/proto/service.proto
bigquery/google/cloud/bigquery_v2/proto/model.proto
bigquery_storage/google/cloud/bigquery_storage_v1beta1/proto/storage.proto
datacatalog/google/cloud/datacatalog_v1beta1/proto/datacatalog.proto
datalabeling/google/cloud/datalabeling_v1beta1/proto/data_labeling_service.proto
dlp/google/cloud/dlp_v2/proto/dlp.proto
irm/google/cloud/irm_v1alpha2/proto/incidents_service.proto
kms/google/cloud/kms_v1/proto/service.proto
language/google/cloud/language_v1/proto/language_service.proto
redis/google/cloud/redis_v1/proto/cloud_redis.proto
redis/google/cloud/redis_v1beta1/proto/cloud_redis.proto
talent/google/cloud/talent_v4beta1/proto/application_service.proto
talent/google/cloud/talent_v4beta1/proto/company_service.proto
talent/google/cloud/talent_v4beta1/proto/completion_service.proto
talent/google/cloud/talent_v4beta1/proto/event_service.proto
talent/google/cloud/talent_v4beta1/proto/job_service.proto
talent/google/cloud/talent_v4beta1/proto/profile_service.proto
talent/google/cloud/talent_v4beta1/proto/tenant_service.proto
translate/google/cloud/translate_v3beta1/proto/translation_service.proto
videointelligence/google/cloud/videointelligence_v1/proto/video_intelligence.proto
```

This "leakage" can break clients who are upgrading an existing virtualenv, because ce currently pin `googleapis-common-protos >= 1.5.3, != 1.5.4, < 2.0dev` in `api_core/setup.py`, and `googleapis-common-protos >= 1.5.9, < 2.0.0dev` in `bigquery_storage/setup.py`.

Steps to remediate:

- [x] Update the pin of `googleapis-common-protos` in `api_core/setup.py`. (#8688)
[...]

Issue #8676, opened 2019-07-15
Logging: systest teardown flakes with 404.


tseaver:
From: https://source.cloud.google.com/results/invocations/91034a85-0df7-4b53-bba7-258e166bb3fe/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Flogging/log

```python
______________________ TestLogging.test_log_root_handler _______________________
self = <test_system.TestLogging testMethod=test_log_root_handler>
    def tearDown(self):
        retry = RetryErrors((NotFound, TooManyRequests, RetryError), max_tries=9)
        for doomed in self.to_delete:
            try:
>               retry(doomed.delete)()
tests/system/test_system.py:111:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../test_utils/test_utils/retry.py:108: in wrapped_function
    return to_wrap(*args, **kwargs)
google/cloud/logging/logger.py:223: in delete
    client.logging_api.logger_delete(self.project, self.name)
google/cloud/logging/_gapic.py:139: in logger_delete
    self._gapic_api.delete_log(path)
google/cloud/logging_v2/gapic/logging_service_v2_client.py:326: in delete_log
    request, retry=retry, timeout=timeout, metadata=metadata
../api_core/google/api_core/gapic_v1/method.py:143: in __call__
    return wrapped_func(*args, **kwargs)
../api_core/google/api_core/retry.py:273: in retry_wrapped_func
    on_error=on_error,
../api_core/google/api_core/retry.py:182: in retry_target
    return target()
../api_core/google/api_core/timeout.py:214: in func_with_timeout
    return func(*args, **kwargs)
../api_core/google/api_core/grpc_helpers.py:59: in error_remapped_callable
    six.raise_from(exceptions.from_grpc_error(exc), exc)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
value = NotFound(u'Log handler_root-1563211829508 does not exist',)
from_value = <_Rendezvous of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	deta...pc_message":"Log handler_root-1563211829508 does not exist","grpc_status":5}"
>
[...]

PR #8709, opened 2019-07-18
Bump minimum version for google-api-core to 1.14.0.

tseaver:
Toward #8686.


plamut:
The `test_managing_subscription_iam_policy` PubSub system test passes locally, and other `iam` system tests also passed - could be flakiness, will force a re-run.

tseaver:
Unrelated failures:

- Asset fails w/ flaky assertion error in VPCSC systest, #8715.

- Logging fails due to flaky systest teardown, #8676 

- Pubsub fails w/ flaky deadline error in subscription IAM test, #8716.

PR #8713, opened 2019-07-19
BigQuery: Set IPython user agent when running queries with IPython cell magic

### LINKED ISSUE #8696: BigQuery: Add user-agent info to Jupyter magics Client constructor(s).
In order to better prioritize jupyter features such as #8342 and #8143, it is important for Google to know how much these features are used to query BigQuery. This is done by populating the user-agent string with the corresponding "application name" and version information.

I propose we use the IPython version information in this string, as these magics are being used from an IPython kernel. Getting the IPython version string: https://stackoverflow.com/a/56265255/101923 Something like:

```
bigquery.Client(
    client_info=ClientInfo(
        user_agent="ipython-{}".format(IPython.__version__)
    )
)
```

plamut:
Closes #8696.

### How to test

See issue description - the Client should be instantiated with a custom user agent string when used from an IPython cell.

### Things to discuss
- [x] ~~A connection from the context could override this custom user agent string - is that expected behavior, or should the cell magic actually override the user agent string in context connections, too?~~
No, this is primarily used with Kaggle datasets, thus the overridden user agent string should take precedence.

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2018 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2019 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/magics.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
```
tswast: No need to update copyright headers. It should indicate when the file was created, not necessarily when it was last updated.

plamut: I see. Will revert.

plamut:
The new deprecation warnings that caused one of the test to fail were related to some Pandas thingies, namely the `RangeIndex` attributes `_start`, `_stop`, and `_step`. I modified the test to only focus on the expected warning related to `pyarrow`.

plamut:
Heh, now the coverage check is complaining. Will check and make the necessary adjustments.

plamut:
The coverage check complained, because not all paths have been taken in the `next(...)` construct in the test, go figure.

However, I did notice that even those three new warnings are caused by `pyarrow`. That was my bad, I initially only looked at `w.message`, but `str(w)` contains more info, e.g. the filename of the warning source, and that source is the `pyarrow` lib itself:

PR #8720, opened 2019-07-19
AutoML: Add a TablesClient for automl-tables specific behavior.

lwander:
AutoML tables is heavily used by datascientists in jupyter notebooks. Based on feedback we've gotten, the existing client is hard to use, obtuse, not notebook-friendly. This wrapper client aims to fix that.

## Using the client (see integration tests for more detailed snippets):

```python
from google.cloud import automl_v1beta1

client = automl_v1beta1.TablesClient(project='my-project-123')

client.list_datasets()

dataset = client.create_dataset('my_dataset')
client.import_data(dataset=dataset, gcs_input_uris='gs://my-data/data.csv').result()
client.set_target_column(dataset=dataset, column_spec_display_name='output')

model = client.create_model(dataset=dataset, train_budget_milli_node_hours=1000).result()
```

## Testing

Most testing is done in unit tests, however we have some integration tests covering the common user flows as well (create dataset -> import data -> train -> predict). To speed up the testing, we create resources the first time tests are run, and run tests off of those in subsequent runs. This mitigates the minimum 1h+ model creation & deployment time.

## Future work

In no particular order, we plan to add:

* IPython integration to pretty print objects
* Matplotlib integration to show dataset & model stats
* Pandas/numpy integration for data import & batch predict

Sorry for the large change, most of it is doc strings and unit tests. We were recommended to check in the full basic functionality first.

### DIFF
--- a/automl/README.rst
+++ b/automl/README.rst
@@ -104,3 +104,21 @@ Next Steps
    API to see other available methods on the client.
 -  Read the `Product documentation`_ to learn
    more about the product and see How-to Guides.
+
+Making & Testing Local Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want to make changes to this library, here is how to set up your
+development environment:
+
+1. Make sure you have `virtualenv`_ installed and activated as shown above.
+2. Run the following one-time setup (it will be persisted in your virtualenv):
+
+.. code-block:: console
+
+    pip install -r ../docs/requirements.txt
+    pip install -U mock pytest
+
+3. To test any changes you've made, run `pytest` in this directory.
+   
+
--- a/automl/README.rst
+++ b/automl/README.rst
@@ -117,7 +117,7 @@ development environment:
 .. code-block:: console
 
     pip install -r ../docs/requirements.txt
-    pip install -U mock pytest
+    pip install -U nox mock pytest
 
 3. If you want to run all tests, you will need a billing-enabled 
    `GCP project`_, and a `service account`_ with access to the AutoML APIs.
@@ -131,5 +131,5 @@ development environment:
 .. code-block:: console
 
     export PROJECT_ID=<project-id> GOOGLE_APPLICATION_CREDENTIALS=</path/to/creds.json>
-    pytest
+    nox
 
--- a/automl/README.rst
+++ b/automl/README.rst
@@ -114,10 +114,10 @@ development environment:
 1. Make sure you have `virtualenv`_ installed and activated as shown above.
 2. Run the following one-time setup (it will be persisted in your virtualenv):
 
-.. code-block:: console
+   .. code-block:: console
 
-    pip install -r ../docs/requirements.txt
-    pip install -U nox mock pytest
+       pip install -r ../docs/requirements.txt
+       pip install -U nox mock pytest
 
 3. If you want to run all tests, you will need a billing-enabled 
    `GCP project`_, and a `service account`_ with access to the AutoML APIs.
--- a/automl/README.rst
+++ b/automl/README.rst
@@ -119,6 +119,17 @@ development environment:
     pip install -r ../docs/requirements.txt
     pip install -U mock pytest
 
-3. To test any changes you've made, run `pytest` in this directory.
-   
+3. If you want to run all tests, you will need a billing-enabled 
+   `GCP project`_, and a `service account`_ with access to the AutoML APIs.
+   Note: the first time the tests run in a new project it will take a _long_
+   time, on the order of 2-3 hours. This is one-time setup that will be skipped
+   in future runs.
+
+.. _service account: https://cloud.google.com/iam/docs/creating-managing-service-accounts
+.. _GCP project: https://cloud.google.com/resource-manager/docs/creating-managing-projects
+
+.. code-block:: console
+
+    export PROJECT_ID=<project-id> GOOGLE_APPLICATION_CREDENTIALS=</path/to/creds.json>
+    pytest
 
--- a/automl/google/cloud/automl_v1beta1/__init__.py
+++ b/automl/google/cloud/automl_v1beta1/__init__.py
@@ -18,10 +18,10 @@
 from __future__ import absolute_import
 
 from google.cloud.automl_v1beta1 import types
-from google.cloud.automl_v1beta1.helper import tables
 from google.cloud.automl_v1beta1.gapic import auto_ml_client
 from google.cloud.automl_v1beta1.gapic import enums
 from google.cloud.automl_v1beta1.gapic import prediction_service_client
+from google.cloud.automl_v1beta1.tables import tables_client
 
 
 class AutoMlClient(auto_ml_client.AutoMlClient):
@@ -34,4 +34,9 @@ class PredictionServiceClient(prediction_service_client.PredictionServiceClient)
     enums = enums
 
 
-__all__ = ("enums", "types", "AutoMlClient", "PredictionServiceClient")
+class TablesClient(tables_client.TablesClient):
+    __doc__ = tables_client.TablesClient.__doc__
+
+
+__all__ = ("enums", "types", "AutoMlClient", "PredictionServiceClient",
+        "TablesClient")
--- a/automl/google/cloud/automl_v1beta1/__init__.py
+++ b/automl/google/cloud/automl_v1beta1/__init__.py
@@ -18,6 +18,7 @@
 from __future__ import absolute_import
 
 from google.cloud.automl_v1beta1 import types
+from google.cloud.automl_v1beta1.helper import tables
 from google.cloud.automl_v1beta1.gapic import auto_ml_client
 from google.cloud.automl_v1beta1.gapic import enums
 from google.cloud.automl_v1beta1.gapic import prediction_service_client
--- a/automl/google/cloud/automl_v1beta1/__init__.py
+++ b/automl/google/cloud/automl_v1beta1/__init__.py
@@ -38,5 +38,4 @@ class TablesClient(tables_client.TablesClient):
     __doc__ = tables_client.TablesClient.__doc__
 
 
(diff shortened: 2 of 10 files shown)


lwander:
@jonathan1920 you need to sign the CLA, or resign the commits with the correct @google.com address

beccasaurus:
@merla18 Please take a look.

 At a high level, it looks like this custom layer does the following:
 - Instantiate client with Project ID so the client knows the Project ID
 - Treat the resources' Display Names as a shorthand identifier
 - CRUD operations can accept any of the following to identify a resource:
   - Full name (projects/X/locations/Y/resource_type/**ID**)
   - The resource's ID only (project ID stored on client)
   - The resource's Display Name
   - Python pb object representing the resource, itself 
 - This is branded for Tables but most of the functionality would work across the board for AutoML
   - The identifier shorthand functionality would also work for any resourceful GCP API

### Project ID on client

IIRC there have been proposals over the years for storing project on the client.

As this manual layer implements this feature, I'm wondering: is this a feature that has any active efforts for GAX/core?

### Calling RPC methods with shorthand for resource names/identifiers

Are there any other Python Google Cloud Client Libraries that implement something like this?

I'm a little worried about consistency. When learning GCP APIs, part of the learning curve is resources (their "names" which are fully qualified identifiers -vs- display names which many resources have -vs- identifiers which are often server-side generated but sometimes client provided).

This layer simplifies working with resources a lot, it looks very elegant. But learning this interface might confuse developers when they use this interface and then try using any of the other GCP API Python Cloud Client Libraries.

Looks elegant! I'd like to find out more, though. Especially because this layer simplifies working with GCP API resources _in general_, which I love.

----

/cc @sirtorry @andrewferlitsch @dizcology @lukesneeringer @crwilcox @JustinBeckwith

andrewferlitsch:
Rebecca

What's most notable to me on the AutoML APIs is that they are designed in a
procedural style instead of an OOP style. They lack abstraction,
encapsulation and polymorphism (like Keras vs pre-TF 2.0 --hence why TF 2.0
adopted Keras for the model API). Looking at the above bullets, first
thoughts are:

- things like project ID, etc are attributes of the client object.
(encapsulation)

- shorthand would be attribute aliases. (abstraction)

- CRUD ops sound like overloaded methods (polymorphism).

andrewferlitsch:
@beccasaurus I recommend we do a review and see the best "wrapper" approach that can be applied uniformly across all the APIs.

For example, it maybe better to provide an attribute to a generic Client type that would indicate the type of dataset/model (Tables, etc) vs. subclassing (i.e., TablesClient is a subclass of Client).

lwander:
That's a great point:
> I'm a little worried about consistency. When learning GCP APIs, part of the learning curve is resources (their "names" which are fully qualified identifiers

The real difficultly in using the original client is that a lot of these identifiers are hidden behind multiple API calls & string transformations. For example, to update a column (a necessary and required step), you need to know first the fully-qualified name of the dataset, use that to lookup the fully-qualified name of the primary table, and then search the primary table for the column in question. At that point, you need to extract the API-assigned identifier from the column's fully qualified name (something of the form `TBL123...`). At this point you're ready to update the column. This doesn't even get into the odd behavior you see where certain properties of the column are required in the update, and others are optional. 

This API works great when you're a browser & always have the entire state of all your resources in memory. However, when you're just getting started in a jupyter notebook and want to _quickly_ run some experiments using autoML, you're out of luck.

Would the "correct" solution be to rewrite all the APIs with both client users & browsers in mind? Maybe. But that doesn't cover the future work we want to implement, namely integration with pandas, integration with matplotlib, and integration with IPython. 

On top of that, we have complaints from both internal users who are experts in GCP, as well as users who are brand new to the platform (and these complaints don't even address the lack of the above mentioned enhancements). Overall, there is just too much complexity in the client to perform very simple operations.

In the end, the tables team wants a polished, first-class, data scientist-friendly client, as this is what our users are asking for. Consequently, we're ready to put in the effort to maintain this client.

automl/README.rst
```diff
@@ -104,3 +104,32 @@ Next Steps
...
+If you want to make changes to this library, here is how to set up your
+development environment:
+
+1. Make sure you have `virtualenv`_ installed and activated as shown above.
+2. Run the following one-time setup (it will be persisted in your virtualenv):
+
+.. code-block:: console
+
+    pip install -r ../docs/requirements.txt
+    pip install -U mock pytest
```
busunkim96: This library uses [nox](https://nox.thea.codes/en/stable/) for testing. Nox creates a new virtualenv for each test run and installs the necessary dependencies).
```
pip install --user --upgrade nox
```

`nox` will run all of the nox sessions defined in the [noxfile.py](https://github.com/googleapis/google-cloud-python/blob/master/automl/noxfile.py). You can run a specific session by doing `nox -s <session_name>`. For example, to just run the Python 2.7 system tests you would do `nox -s system-2.7`.

automl/google/cloud/automl_v1beta1/__init__.py
```diff
@@ -33,4 +34,8 @@ class PredictionServiceClient(prediction_service_client.PredictionServiceClient)
     enums = enums
 
 
-__all__ = ("enums", "types", "AutoMlClient", "PredictionServiceClient")
+class TablesClient(tables_client.TablesClient):
+    __doc__ = tables_client.TablesClient.__doc__
+
+
+__all__ = ("enums", "types", "AutoMlClient", "PredictionServiceClient", "TablesClient")
```
busunkim96: Our bot will try to remove these changes once this is removed to master. Could you add these replacements to the synth.py for AutoML? [This](https://github.com/googleapis/google-cloud-python/blob/5bc8505f1e885971933cb6093eeff7af0b8be0bd/vision/synth.py#L40-L47) is a similar replacement to add helpers to the vision client.

automl/README.rst
lwander: Will do. Should I include `pytest mock` as requirements as well just to be safe?

automl/google/cloud/automl_v1beta1/__init__.py
lwander: Just to be sure I understand: I should back out all changes to __init__.py, and have synth.py add them instead?

If I do so, the tests will break due to imports not working... should I leave them in until we are ready to merge?

automl/README.rst
busunkim96: Yes, feel free to keep that info as well.

automl/google/cloud/automl_v1beta1/__init__.py
busunkim96: Nope! You can keep the changes here, we'd just prefer to also add them to synth.py. We have a bot run all the synth.py scripts nightly and open PRs if the generated version is different from what's currently in master.

Since the generator will produce an __init__.py without the `TablesClient` addition, we need to make this addition permanent via a replacement in `synth.py`.

automl/tests/unit/gapic/v1beta1/test_tables_client_v1beta1.py
```diff
@@ -0,0 +1,1365 @@
...
+        ds = client.list_datasets()
+        client.auto_ml_client.location_path.assert_called_with(PROJECT, REGION)
+        client.auto_ml_client.list_datasets.assert_called_with(LOCATION_PATH)
+        assert len(ds) == 1
+        assert ds[0] == "some_dataset"
+
+    def test_get_dataset_no_value(self):
+        dataset_actual = "dataset"
+        client = self.tables_client({}, {})
+        error = None
```
busunkim96: Would it be possible to use `pytest.raises()` to check for expected exceptions? [pytest documentation](http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions)

```python
    client.get_dataset()
```

lwander: Ah that looks much neater, I'll switch to that.

lwander:
Updated

beccasaurus:
✅ Green light from my end. We followed up about Project ID / Resource Objects offline.

I'm +1 to merge this when it's considered ready.

@busunkim96 Could you add @sirtorry as a reviewer?

busunkim96:
I opened a PR on sloth to add @sirtorry to yoshi-python. You should be able to add comments for now.

automl/tests/system/gapic/v1beta1/test_system_tables_client_v1.py
```diff
@@ -0,0 +1,245 @@
...
+            if op.cancelled():
+                return
+            time.sleep(sleep_time)
+            sleep_time = min(sleep_time * 2, MAX_SLEEP_TIME_SECONDS)
+        assert op.cancelled()
+
+    def test_list_datasets(self):
+        client = automl_v1beta1.TablesClient(project=PROJECT, region=REGION)
+        # we need to unroll the iterator to actually make client calls
+        [d for d in client.list_datasets(timeout=10)]
```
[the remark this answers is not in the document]
lwander: Good point -- do you think I should add a test_create_list_datasets or similar test to make sure a dataset we know exists shows up in our client call? My primary objective here was to make sure we return an iterator we can unroll.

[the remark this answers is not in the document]
beccasaurus: Having some kind of assert would be ideal. Without asserts, the test reads as incomplete.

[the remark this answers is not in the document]
busunkim96: Some ideas:
* If there is an expected # of datasets, assert that. (Or assert that is > 0?)
* Assert that some properties exist on a list element

lwander:
I've made the list system tests more stringent. Anything blocking us from merging?

lwander:
Friendly ping

busunkim96:
@sirtorry Please merge it when you're ready. You should have write permissions to the repo.

lwander:
Thanks all! 🙂

Issue #4840, opened 2018-02-05
Determine how to support emulators consistently across all clients.


jonparrott:
There are emulators available for several APIs (such as Datastore and Pub/Sub). We need to figure out the right client patterns to support these across the board. This issue is to discuss our options.

chemelnucfin:
See #4897 also for storage related questions.

rvandegrift:
Shouldn't the libraries just use the env vars that trigger the SDK to connect to the emulators?

For instance, to use the Datastore emulator, you set `DATASTORE_EMULATOR_HOST=<host>:port`.  That's enough indication that `google.cloud.datastore.Client` should skip credentials and connect directly to the emulator.

What is the reason for thinking that some other interface is required?  IMO, if I have to set anything else, it's a worse developer experience.

theacodes:
That's one option. The description of the bug is the figure out the right client patterns, and that includes considering implicitly using emulators when env vars are set.

rvandegrift:
Would an implementation of this approach be merged?  Seems like there hasn't been another idea in the past year

brandoconnor:
Chiming in to support @rvandegrift 's approach and hopefully nudge this forward. Taking this tack also [aligns with the node.js library](https://github.com/firebase/firebase-js-sdk/blob/master/packages/testing/src/api/index.ts) supporting the emulator.

If we're feeling in the spirit of debate, passing a configuration object when instantiating a client would feel plenty idiomatic ([boto3.client or Session](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html) are reasonable examples).

In terms of local development and testing, this functionality seems critical and it's somewhat surprising to be missing. What's needed to keep this moving?

deanberris:
Is there an update on this? I'm in the process of attempting to migrate an AppEngine service running in the python27 environment towards using the google-cloud-* client libraries for use in the python37, and testing locally with emulation is critical for integration testing to ensure we're not breaking anything. Is this a regression, maybe something we can work around?

PR #8721, opened 2019-07-19
Firestore: support emulator in client.

### LINKED ISSUE #7500: Firestore: add emulator support
The library does not respect the `FIRESTORE_EMULATOR_HOST` environment variable thus it cannot be used with the emulator.
Respect the `FIRESTORE_EMULATOR_HOST` environment variable and connect to the emulator when set.
None. If there are other ways to use the emulator with the library they are not documented.

crwilcox:
Adds support for using the firestore emulator with the client library. 

Fixes #7500

### DIFF
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -23,6 +23,8 @@ In the hierarchy of API concepts
 * a :class:`~google.cloud.firestore_v1.client.Client` owns a
   :class:`~google.cloud.firestore_v1.document.DocumentReference`
 """
+import os
+
 from google.api_core.gapic_v1 import client_info
 from google.cloud.client import ClientWithProject
 
@@ -103,6 +105,7 @@ class Client(ClientWithProject):
         )
         self._client_info = client_info
         self._database = database
+        self._emulator_host = os.getenv("FIRESTORE_EMULATOR_HOST")
 
     @property
     def _firestore_api(self):
@@ -115,11 +118,17 @@ class Client(ClientWithProject):
         if self._firestore_api_internal is None:
             # Use a custom channel.
             # We need this in order to set appropriate keepalive options.
+
             channel = firestore_grpc_transport.FirestoreGrpcTransport.create_channel(
                 self._target,
                 credentials=self._credentials,
                 options={"grpc.keepalive_time_ms": 30000}.items(),
             )
+
+            if self._emulator_host is not None:
+                channel = firestore_grpc_transport.firestore_pb2_grpc.grpc.insecure_channel(
+                    self._emulator_host,
+                )
 
             self._transport = firestore_grpc_transport.FirestoreGrpcTransport(
                 address=self._target, channel=channel
@@ -138,6 +147,9 @@ class Client(ClientWithProject):
         Returns:
             str: The location of the API.
         """
+        if self._emulator_host is not None:
+            return self._emulator_host
+
         return firestore_client.FirestoreClient.SERVICE_ADDRESS
 
     @property
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -193,7 +193,7 @@ class Client(ClientWithProject):
 
             if self._emulator_host is not None:
                 # The emulator requires additional metadata to be set.
-                self._rpc_metadata_internal.append(("authorization", "bearer owner"))
+                self._rpc_metadata_internal.append(("authorization", "Bearer owner"))
 
         return self._rpc_metadata_internal
 
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -192,7 +192,7 @@ class Client(ClientWithProject):
             )
 
             if self._emulator_host is not None:
-                # The emulator requires additioanl metadata to be set.
+                # The emulator requires additional metadata to be set.
                 self._rpc_metadata_internal.append(("authorization", "bearer owner"))
 
         return self._rpc_metadata_internal
--- a/firestore/google/cloud/firestore_v1/client.py
+++ b/firestore/google/cloud/firestore_v1/client.py
@@ -127,7 +127,7 @@ class Client(ClientWithProject):
 
             if self._emulator_host is not None:
                 channel = firestore_grpc_transport.firestore_pb2_grpc.grpc.insecure_channel(
-                    self._emulator_host,
+                    self._emulator_host
                 )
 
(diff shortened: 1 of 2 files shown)


firestore/google/cloud/firestore_v1/client.py
```diff
@@ -179,6 +191,10 @@ def _rpc_metadata(self):
                 self._database_string
             )
 
+            if self._emulator_host is not None:
+                # The emulator requires additional metadata to be set.
+                self._rpc_metadata_internal.append(("authorization", "bearer owner"))
```
ryanpbrewster: I believe this should be `"Bearer owner"`, only the metadata _keys_ need to be lowercase

crwilcox: Thanks. It doesn't seem to care one way or the other but I will capitalize it.

firestore/google/cloud/firestore_v1/client.py
```diff
@@ -51,6 +53,7 @@
 _ACTIVE_TXN = "There is already an active transaction."
 _INACTIVE_TXN = "There is no active transaction."
 _CLIENT_INFO = client_info.ClientInfo(client_library_version=__version__)
+_FIRESTORE_EMULATOR_HOST = "FIRESTORE_EMULATOR_HOST"
```
tseaver: Up until now, we have exposed these variables in `google.cloud.environment_vars`:

https://github.com/googleapis/google-cloud-python/blob/ced5858d1f656116015c146454275a4045418e44/core/google/cloud/environment_vars.py#L24-L31

I'm not sure it is the Right Thing(TM), but it is precedent.

firestore/google/cloud/firestore_v1/client.py
```diff
@@ -115,12 +119,18 @@ def _firestore_api(self):
...
                 self._target,
                 credentials=self._credentials,
                 options={"grpc.keepalive_time_ms": 30000}.items(),
             )
 
+            if self._emulator_host is not None:
+                channel = firestore_grpc_transport.firestore_pb2_grpc.grpc.insecure_channel(
+                    self._emulator_host
+                )
+
```
tseaver: The "main" version above (lines 123-127) should be in an `else:` clause for this `if`.

firestore/tests/unit/v1/test_client.py
```diff
@@ -48,6 +48,17 @@ def test_constructor(self):
         self.assertEqual(client._database, DEFAULT_DATABASE)
         self.assertIs(client._client_info, _CLIENT_INFO)
```
tseaver: Add an assertion here:  `sellf.assertIsNone(client._emulator_host)`.

firestore/tests/unit/v1/test_client.py
```diff
@@ -84,6 +95,26 @@ def test__firestore_api_property(self, mock_client):
         self.assertIs(client._firestore_api, mock_client.return_value)
         self.assertEqual(mock_client.call_count, 1)
 
+    @mock.patch(
+        "google.cloud.firestore_v1.gapic.firestore_client." "FirestoreClient",
```
tseaver: Just join the strings.

firestore/tests/unit/v1/test_client.py
```diff
@@ -84,6 +95,26 @@ def test__firestore_api_property(self, mock_client):
...
+    def test__firestore_api_property_with_emulator(self, mock_client):
+        emulator_host = "localhost:8081"
+        with mock.patch("os.getenv") as getenv:
+            getenv.return_value = emulator_host
+            client = self._make_default_one()
+
+        self.assertIsNone(client._firestore_api_internal)
+        firestore_api = client._firestore_api
+        self.assertIs(firestore_api, mock_client.return_value)
+        self.assertIs(firestore_api, client._firestore_api_internal)
```
tseaver: Needs assertion(s) that an insecure channel was constructed with the expected emulator URL.

firestore/google/cloud/firestore_v1/client.py
crwilcox: @tseaver. I saw that and made the active decision I don't like putting those constants there. I think having to release a new core for this seemed a bit silly? I can't imagine those are really used all that broadly? It seemed to me putting a constant I only need in a single client in a shared component will just complicate this.

@busunkim96 do you have thoughts on this? how does this relate to repo-split? I agree with @tseaver consistency has some value so it may be worth deciding now on our direction here?

tseaver: +1 for consistency, see #4840 .

firestore/tests/unit/v1/test_client.py
```diff
@@ -111,6 +118,9 @@ def test__firestore_api_property_with_emulator(self, mock_client):
         self.assertIs(firestore_api, mock_client.return_value)
         self.assertIs(firestore_api, client._firestore_api_internal)
 
+        self.assertEqual(mock_insecure_channel.call_count, 1)
+        self.assertEqual(mock_insecure_channel.call_args.args[0], emulator_host)
```
tseaver: Wouldn't this be simpler / clearer?

```python
        mock_insecure_channel.assert_called_once_with(emulator_host)
```

firestore/google/cloud/firestore_v1/client.py
tseaver: Maybe the Right Thing(TM) is to split up the `core/google/cloud/environment_vars.py` module, moving the API-specific values into their own packages.  The advantage of the module is that it makes documenting them easier.

busunkim96: No strong feelings here -- I would prefer we move them out to the individual APIs eventually. :woman_shrugging:

crwilcox: I am going to leave this like this. I think we may want to push them into the individual packages. Rather than remove from core we may just duplicate though. The end goal should be removing the dependency on core altogether, rather than to start deprecating parts?

firestore/tests/unit/v1/test_client.py
```diff
@@ -118,8 +118,8 @@ def test__firestore_api_property_with_emulator(
         self.assertIs(firestore_api, mock_client.return_value)
         self.assertIs(firestore_api, client._firestore_api_internal)
 
-        self.assertEqual(mock_insecure_channel.call_count, 1)
-        self.assertEqual(mock_insecure_channel.call_args.args[0], emulator_host)
+        mock_insecure_channel.assert_called_once_with(emulator_host)
+
```
tseaver: Extra newline?  Anyway, `lint` is failing, needs blackening.

crwilcox: Thanks. Will merge on green.

PR #8734, opened 2019-07-23
BigQuery: Set BQ storage client user-agent when in Jupyter cell

### LINKED ISSUE #8696: BigQuery: Add user-agent info to Jupyter magics Client constructor(s).

plamut:
Closes #8696 

This is a follow-up to #8713, it adds the same user-agent string to the optional BQ storage client, if the latter is used.

### How to test
See issue description and the subsequent [comment](https://github.com/googleapis/google-cloud-python/issues/8696#issuecomment-513882649) about adding the user-agent info to the other client, too.

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -147,6 +147,9 @@ from google.cloud.bigquery.dbapi import _helpers
 import six
 
 
+IPYTHON_USER_AGENT = "ipython-{}".format(IPython.__version__)
+
+
 class Context(object):
     """Storage for objects to be used throughout an IPython notebook session.
 
@@ -396,9 +399,7 @@ def _cell_magic(line, query):
         project=project,
         credentials=context.credentials,
         default_query_job_config=context.default_query_job_config,
-        client_info=client_info.ClientInfo(
-            user_agent="ipython-{}".format(IPython.__version__)
-        ),
+        client_info=client_info.ClientInfo(user_agent=IPYTHON_USER_AGENT),
     )
     if context._connection:
         client._connection = context._connection
@@ -439,4 +440,7 @@ def _make_bqstorage_client(use_bqstorage_api, credentials):
         )
         six.raise_from(customized_error, err)
 
-    return bigquery_storage_v1beta1.BigQueryStorageClient(credentials=credentials)
+    return bigquery_storage_v1beta1.BigQueryStorageClient(
+        credentials=credentials,
+        client_info=client_info.ClientInfo(user_agent=IPYTHON_USER_AGENT),
+    )
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -139,15 +139,12 @@ try:
     from IPython.core import magic_arguments
 except ImportError:  # pragma: NO COVER
     raise ImportError("This module can only be loaded in IPython.")
-try:
-    from google.cloud import bigquery_storage_v1beta1
-except ImportError:  # pragma: NO COVER
-    bigquery_storage_v1beta1 = None
 
 from google.api_core import client_info
 import google.auth
 from google.cloud import bigquery
 from google.cloud.bigquery.dbapi import _helpers
+import six
 
 
 class Context(object):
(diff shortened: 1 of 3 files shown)


plamut:
When running unit tests locally, I noticed that one of them fails [here](https://github.com/googleapis/google-cloud-python/blob/master/bigquery/tests/unit/test__pandas_helpers.py#L491). The reason seems to be that `float("nan")` is converted to `None` by `pyarrow`, instead of the expected `float("nan")`. It also happens on the latest `master`, thus the failure seems unrelated to this PR.

My `Pandas` and `pyarrow` versions:
```
pandas==0.24.2
pyarrow==0.13.0
```

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -433,10 +431,16 @@ def _make_bqstorage_client(use_bqstorage_api, credentials):
...
-    if bigquery_storage_v1beta1 is None:
-        raise ImportError(
+    try:
+        from google.cloud import bigquery_storage_v1beta1
+    except ImportError as err:
+        customized_error = ImportError(
             "Install the google-cloud-bigquery-storage and fastavro packages "
             "to use the BigQuery Storage API."
         )
+        six.raise_from(customized_error, err)
```
tswast: I like this pattern, thanks. Looks much better for debugging.

bigquery/tests/unit/test_magics.py
```diff
@@ -65,6 +65,22 @@ def ipython_interactive(request, ipython):
         yield ipython
 
 
+@pytest.fixture(scope="session")
+def missing_bq_storage():
+    """Provide a patcher that can make the bigquery storage import to fail."""
+    orig_import = six.moves.builtins.__import__
+
+    def custom_import(name, globals=None, locals=None, fromlist=(), level=0):
```
tswast: Should we make this `__import__` mock into a test utility (at https://github.com/googleapis/google-cloud-python/tree/master/test_utils/test_utils or maybe https://github.com/googleapis/google-cloud-python/blob/master/bigquery/tests/unit/helpers.py)?

Seems generally useful, and also something I'd like to see tests for.

plamut: I'd say we put it in bigquery test utils for now, and ask other stakeholders if they are fine with promoting it to core test utils.
(there might be additional features/generalizations requested, but that would not block this particular PR)

plamut: BTW, whom can I ask if the new helper is fit for promotion, who owns the core?

tseaver: > BTW, whom can I ask if the new helper is fit for promotion, who owns the core?
> 
> (if there isn't one and we can do it at our own discretion, please let me know)

I would say add a separate PR promoting it, and tag @busunkim96 for review.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -433,10 +431,16 @@ def _make_bqstorage_client(use_bqstorage_api, credentials):
...
 
-    return bigquery_storage_v1beta1.BigQueryStorageClient(credentials=credentials)
+    return bigquery_storage_v1beta1.BigQueryStorageClient(
+        credentials=credentials,
+        client_info=client_info.ClientInfo(user_agent=IPYTHON_USER_AGENT),
```
tswast: I think we need the other ClientInfo, since this is a grpc client. [`google.api_core.gapic_v1.client_info.ClientInfo`](https://googleapis.github.io/google-cloud-python/latest/core/client_info.html#google.api_core.gapic_v1.client_info.ClientInfo)

That import can fail if grpc libraries aren't installed, so the import should be in the try/except block.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -433,10 +431,24 @@ def _make_bqstorage_client(use_bqstorage_api, credentials):
     if not use_bqstorage_api:
         return None
 
-    if bigquery_storage_v1beta1 is None:
-        raise ImportError(
+    try:
+        from google.cloud import bigquery_storage_v1beta1
+    except ImportError as err:
+        customized_error = ImportError(
             "Install the google-cloud-bigquery-storage and fastavro packages "
```
tswast: Could you update this to `pyarrow` instead of `fastavro`? I missed this when I implemented the new wire format.

Issue #7676, opened 2019-04-08
API Core: Simplify from_rfc3339 methods


crwilcox:
`from_rfc3339` doesn't handle Nanos, but `from_rfc3339_nanos` does. Would it make sense to use the `from_rfc3339_nanos` implementation for both? It seems to me we could simplify the interface by having one `from rfc3339` method.

https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/datetime_helpers.py#L117 

https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/datetime_helpers.py#L130

PR #8742, opened 2019-07-24
feat(api core): simplify from_rfc3339 methods

Emar-Kar:
Combining methods `from_rfc3339` and `from_rfc3339_nanos` as requested in https://github.com/googleapis/google-cloud-python/issues/7676. 
`from_rfc3339_nanos` method marked as deprecated and refers to the new implementation.

Closes: #7676.


Emar-Kar:
@tseaver May I ask a question, please. Should I deprecate private methods `_nanos`?

api_core/google/api_core/datetime_helpers.py
```diff
@@ -131,11 +131,11 @@ def from_rfc3339(value):
...
     """
-    with_nanos = _RFC3339_NANOS.match(value)
-
-    if with_nanos is None:
+    try:
         return datetime.datetime.strptime(value, _RFC3339_MICROS).replace(tzinfo=pytz.utc)
-    else:
+    except ValueError:
+        with_nanos = _RFC3339_NANOS.match(value)
+
```
tseaver: Please restore the previous version:  testing the regex first is faster / clearer than raising the exception.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -155,14 +155,14 @@ def from_rfc3339_nanos(value):
     """Convert a nanosecond-precision timestamp to a native datetime.
```
tseaver: Please add `(Deprecated)` at the beginning of this line.

api_core/google/api_core/datetime_helpers.py
Emar-Kar: I restored the previous version with `if-else` expression, and ran couple of tests this morning. But as I can see this was the reason for the coverage mistakes earlier. I mean, we don't need to check it that way because `_RFC3339_NANOS` pattern already includes the structure which is needed. I tried to think of alternatives but  it looks the only option for me. If I'm losing something will be thankful for the advice.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -115,45 +116,22 @@ def from_iso8601_time(value):
...
+    """Convert a nanosecond-precision or if nanoseconds are missing
+    microsecond-precision timestamp to a native datetime.
 
     .. note::
         Python datetimes do not support nanosecond precision; this function
         therefore truncates such values to microseconds.
 
     Args:
-        value (str): The RFC3339 string to convert.
+        value (str): The RFC 3339 string to convert.
```
crwilcox: We should be consistent with RFC3339 or RFC 3339.  I propose `RFC3339` as it matches the method name spacing.

api_core/google/api_core/datetime_helpers.py
crwilcox: @Emar-Kar are you saying that putting the if/else you are unable to get 100% coverage?

Emar-Kar: I checked both realizations with `try-except` and `if-else` statements. The first one was rejected by Tres and the second failed [coverage](https://source.cloud.google.com/results/invocations/e3cf662f-d7b2-452a-a450-89f294c8d3e2/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fapi_core/log). I guess this situation appears due to the `_RFC3339_NANOS` pattern. It could be applied to the time samples even without nanos and it makes no sense to check it again through `if-else`. On that note, I decided to remove the comparison and let the script just to separate parts as-is.

image

crwilcox: It seems that as of now the only change was to remove a check to be certain we matched something with the regex? Removing this could result in None.group, which will fail at runtime.

Emar-Kar: @crwilcox oh, I finally get it, sorry. Then I'll restore `if-else` to raise an exception with `None`. Could you check, please?

code_sample_3

api_core/google/api_core/datetime_helpers.py
```diff
@@ -169,6 +155,34 @@ def from_rfc3339_nanos(value):
     return bare_seconds.replace(microsecond=micros, tzinfo=pytz.utc)
 
 
+def from_rfc3339_nanos(value):
+    """DEPRECATED. Convert a nanosecond-precision timestamp to a native datetime.
```
crwilcox: I am a bit confused. We are adding a method and marking it deprecated?

api_core/google/api_core/datetime_helpers.py
```diff
@@ -115,20 +116,8 @@ def from_iso8601_time(value):
...
-        value (str): The RFC3339 string to convert.
-
-    Returns:
-        datetime.datetime: The datetime object equivalent to the timestamp in
-            UTC.
-    """
-    return datetime.datetime.strptime(value, _RFC3339_MICROS).replace(tzinfo=pytz.utc)
-
-
-def from_rfc3339_nanos(value):
```
Emar-Kar: @crwilcox not exactly. We are just deprecating this method. It was already there.

api_core/google/api_core/datetime_helpers.py
```diff
@@ -115,20 +116,8 @@ def from_iso8601_time(value):
...
-    """Convert a nanosecond-precision timestamp to a native datetime.
+    """Convert a nanosecond-precision or if nanoseconds are missing
+    microsecond-precision timestamp to a native datetime.
```
tseaver: The summary line should be more terse:  e.g.:

```python
    """Convert an RFC3339-format timestamp to a native datetime.

    Supported formats include those without fractional seconds, or with
    any fraction up to nanosecond precision.

    .. note::
```

api_core/google/api_core/datetime_helpers.py
```diff
@@ -169,6 +155,34 @@ def from_rfc3339_nanos(value):
...
+    Returns:
+        datetime.datetime: The datetime object equivalent to the
+        timestamp in UTC.
+
+    Raises:
+        ValueError: If the timestamp does not match the RFC3339
+            regular expression.
+    """
+    # Raise deprecation warnings for things we want to go away.
+    warnings.warn(
```
crwilcox: Let's remove this warning then I am good

PR #8744, opened 2019-07-24
Firestore: Preserve manual change in noxfile (run systests verbosely).

tseaver:
Supersedes #8740.


tseaver:
Testing-only change, merging without review.

PR #8748, opened 2019-07-24
BigQuery: Include SQL query and job ID in job errors

### LINKED ISSUE #5408: BigQuery: Print offending SQL on error with QueryJob.result()
OS: Linux dc32b7e8763a 4.9.0-6-amd64 #1 SMP Debian 4.9.82-1+deb9u3 (2018-03-02) x86_64 x86_64 x86_64 GNU/Linux
Python version: Python 2.7.6
google-cloud-bigquery: 1.1.0

It would be really helpful if the error message to QueryJob.result() printed out the query, along with line and column numbers. Here's our current workaround, which unfortunately does not allow us to use the built-in Retry due to a nested try/except.
```
RETRYABLE_EXCEPTIONS = (InternalServerError, TooManyRequests, ServiceUnavailable, BadGateway)

...

    def _execute_query_job(self, query_job, query, max_wait_secs=None):
        # Need to do retries manually, since the built-in retry doesn't work within a try/except.
        start_time = datetime.datetime.utcnow()
        deadline = start_time + datetime.timedelta(seconds=max_wait_secs or self.max_wait_secs)
        i = 1
        while datetime.datetime.utcnow() < deadline:
            try:
                return query_job.result()
            except RETRYABLE_EXCEPTIONS:
                time.sleep(i**2)
                i += 1
            except GoogleCloudError as e:
                msg = str(e)
                # Put line numbers next to the SQL.
                lines = query.split('\n')
                longest = max(len(l) for l in lines)
                # Print out a 'ruler' above and below the SQL so we can judge columns.
                ruler = ' ' * 4 + '|'  # Left pad for the line numbers (4 digits plus ':')
                for _ in range(longest / 10):
                    ruler += ' ' * 4 + '.' + ' ' * 4 + '|'
                header = '-----Offending Sql Follows-----'
                padding = ' ' * ((longest - len(header)) / 2)
                msg += '\n\n{}{}\n\n{}\n{}\n{}'.format(padding, header, ruler, '\n'.join(
                    '{:4}:{}'.format(n + 1, line) for n, line in enumerate(lines)), ruler)
                raise RuntimeError(msg)
        raise RuntimeError('Deadline of {} seconds exceeded.'.format(max_wait_secs))
```

plamut:
Closes #5408.

As per user request, this PR adds additional info to any errors that might happen when calling `job.result()`.

### How to test
- Use BigQuery client to run a query that results in an error - e.g. invalid SQL syntax, unknown table/dataset, etc.

The error output only contains the exception traceback.

**Actual result (after the fix):**
The error output contains *all* of the following: exception traceback, job ID, the SQL query itself with lines enumerated. The type of the exception with the modified message is the same as the original exception type.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2862,7 +2862,7 @@ class QueryJob(_AsyncJob):
         lines = query.splitlines()
         max_line_len = max(len(l) for l in lines)
 
-        header = "-----Offending SQL Follows-----"
+        header = "-----Query Job SQL Follows-----"
         header = "{:^{total_width}}".format(header, total_width=max_line_len + 5)
 
         # Print out a "ruler" above and below the SQL so we can judge columns.
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2913,9 +2913,8 @@ class QueryJob(_AsyncJob):
                     self.job_id, retry, project=self.project, location=self.location
                 )
         except exceptions.GoogleCloudError as exc:
-            new_msg = str(exc) + self._format_for_exception(self.query, self.job_id)
-            new_exc = type(exc)(new_msg)
-            raise six.raise_from(new_exc, exc)
+            exc.message += self._format_for_exception(self.query, self.job_id)
+            raise
 
         # If the query job is complete but there are no query results, this was
         # special job, such as a DDL query. Return an empty result set to
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2846,6 +2846,36 @@ class QueryJob(_AsyncJob):
         self._done_timeout = timeout
         super(QueryJob, self)._blocking_poll(timeout=timeout)
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


plamut:
Again, one of the unit tests will fail (converting special float to arrow), but it is not related to this PR.

tseaver:
Looks like we might need to pin one of the dependencies (pyarrow?) to fix our tests.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2846,6 +2846,36 @@ def _blocking_poll(self, timeout=None):
...
+
+        Returns: (str)
+            A formatted query text.
+        """
+        template = "\n\n(job ID: {job_id})\n\n{header}\n\n{ruler}\n{body}\n{ruler}"
+
+        lines = query.splitlines()
+        max_line_len = max(len(l) for l in lines)
+
+        header = "-----Offending SQL Follows-----"
```
tswast: An error with the query isn't always the SQL's fault. Sometimes there is just resource contention or some other transient failure.

I recommend changing the header to `Query Job SQL` or something along those lines.

plamut: A good point indeed, will generalize the header. :+1:

PR #8777, opened 2019-07-25
Data Catalog: Fix client documentation links

ryanyuan:
Data Catalog: Fix client documentation links from nodejs version to python version


tseaver:
Logging and Talent failures are both due to Kokoro-internal glitch: "Cannot check out non-mergeable pull request as if merged."

PR #8779, opened 2019-07-25
Language: Fix reference to EncodingType

mik-laj:
Hello,

The reference to this class is not correct. This PR introduces the correct reference.

Thanks

### DIFF
--- a/language/google/cloud/language_v1/gapic/language_service_client.py
+++ b/language/google/cloud/language_v1/gapic/language_service_client.py
@@ -209,7 +209,7 @@ class LanguageServiceClient(object):
 
                 If a dict is provided, it must be of the same form as the protobuf
                 message :class:`~google.cloud.language_v1.types.Document`
-            encoding_type (~google.cloud.language_v1.types.EncodingType): The encoding type used by the API to calculate sentence offsets.
+            encoding_type (~google.cloud.language_v1.enums.EncodingType): The encoding type used by the API to calculate sentence offsets.
             retry (Optional[google.api_core.retry.Retry]):  A retry object used
                 to retry requests. If ``None`` is specified, requests will not
                 be retried.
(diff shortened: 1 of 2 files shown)


tseaver:
Thanks for the patch, @mik-laj.

tseaver:
Kokoro failures in unrelated API clients are due to a Kokoro-internal glitch, "Cannot check out non-mergeable pull request as if merged."

mik-laj:
I have more similar errors.  I will try to fix soon.

mik-laj:
For example:
In https://googleapis.github.io/google-cloud-python/latest/language/gapic/v1/api.html#google.cloud.language_v1.LanguageServiceClient.annotate_text 
Features is missing.

tseaver:
Ugh, this PR touches generated code direclty, rather than via `synth.py`, which means that our auto-synth bot will try daily to undo the change (see PR #8794).

mik-laj:
Is the entire process of creating these libraries described? I would like to know more about this project. I work a lot with these libraries and I have many comments on how they work, but I do not know how to report issues. When I or someone from my team reports errors, we are sent back to other projects.

tseaver:
@mik-laj The "generated" parts of our libraries use the engine from https://github.com/googleapis/synthtool, under the control of the `synth.py` script in each client library directory. To run a synth generation:

```bash
$ cd language/
$ /path/to/venv/bin/synthtool
...
synthtool > Wrote metadata to synth.metadata.
```

The `yoshi-automation` bot runs synth jobs for each client library daily, and posts PRs if any changes result (other than to the `synth.metadata` file).

We should likely document the synth process in `CONTRIBUTING.rst`.

mik-laj:
What is the role of the gapic-generator?

tseaver:
> What is the role of the gapic-generator?

`synthtool` pulls a docker image which contains (among other things) the `gapic-generatore` software, and uses it to generate the code (as specified in `synth.py`).

PR #8782, opened 2019-07-25
BigQuery: Add StandardSqlDataTypes enum to BigQuery

### LINKED ISSUE #8495: BigQuery: Constants for all scalar standard SQL data types
Related to https://github.com/googleapis/google-cloud-python/issues/7632 (legacy SQL), it would be useful to have constants for all scalar data types for the new `bigquery_v2.types.StandardSqlDataType` format. Currently, one has to construct an object like

```
bigquery_v2.types.StandardSqlDataType(
    type_kind=bigquery_v2.enums.StandardSqlDataType.TypeKind.INT64
)
```

It would be convenient if instead one could just say `bigquery.StandardSqlDataTypes.INT64` and similar for all scalar data types.

plamut:
Closes #8495.

This PR adds a convenience enum with scalar SQL data types (as per issue description).

### How to test
The issue is self-explanatory. Just make sure that the enum contains all the right fields that we want in it.

**Update:** Additionally, make sure that the docs for the new enum are in order.


tseaver:
Looks like we might need to pin one of the dependencies (pyarrow?) to fix our tests.

plamut:
Yeah, I've already been discussing this with @tswast, it's how pyarrow (?) handles `NaN` values on conversions.

bigquery/google/cloud/bigquery/enums.py
```diff
@@ -0,0 +1,84 @@
...
+    anticipated_names = sql_scalar_types | excluded_members
+    unhandled_names = gapic_names - anticipated_names
+
+    if unhandled_names:
+        msg = (
+            "The StandardSqlDataTypes enum migh be out of sync with the "
+            "original StandardSqlDataType.TypeKind enum from gapic. Check "
+            "enum members: {}".format(", ".join(unhandled_names))
+        )
+        warnings.warn(msg, UserWarning)
```
tswast: There's not much a user can do about this error, right? It's more that we'll need to notice this error when the autogen PRs from the yoshi bot come in.

What if we did this check as a unit test?

* `sql_scalar_types` is a module-level constant like `_SQL_SCALAR_TYPES`.
* `excluded_members` is also a module-level constant like `_SQL_NONSCALAR_TYPES`.
* The test:

```
gapic_names = set(
    type_.name for type_ in gapic_enums.StandardSqlDataType.TypeKind
)
anticipated_names = (
    module_under_test._SQL_SCALAR_TYPES
    | module_under_test._SQL_NONSCALAR_TYPES
)
unhandled_names = gapic_names - anticipated_names
assert not unhandled_names
```

bigquery/google/cloud/bigquery/enums.py
```diff
@@ -0,0 +1,84 @@
...
+            for member in gapic_enums.StandardSqlDataType.TypeKind
+            if member.name in sql_scalar_types
+        ),
+    )
+
+    # make sure the docstring for the new enum is also correct
+    orig_doc = gapic_enums.StandardSqlDataType.TypeKind.__doc__
+    skip_pattern = re.compile(
+        "|".join(excluded_members)
+        + "|because a JSON object"  # the second description line of STRUCT member
```
tswast: How brittle is this going to be? Descriptions can change in the protos.

I guess the failure case isn't too bad, though, just a weird line in the docs if we miss it?

bigquery/tests/unit/test_enums.py
```diff
@@ -0,0 +1,85 @@
...
+#  limitations under the License.
+
+import enum
+import mock
+import pytest
+import six
+
+
+@pytest.fixture
+def enum_under_test():
```
tswast: I imagine we'll want some other enums in the future. How about we make a `bigquery/tests/unit/enums` directory like we did for models?

bigquery/google/cloud/bigquery/enums.py
```diff
@@ -0,0 +1,84 @@
...
+            "BYTES",
+            "TIMESTAMP",
+            "DATE",
+            "TIME",
+            "DATETIME",
+            "NUMERIC",
+        )
+    )
+    excluded_members = frozenset(
+        ("TYPE_KIND_UNSPECIFIED", "GEOGRAPHY", "ARRAY", "STRUCT")
```
tswast: GEOGRAPHY is a strange one, but technically a scalar. Let's add it to the scalar types set.

bigquery/google/cloud/bigquery/enums.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    # Sanity check - we do not want the new enum to go out of sync with the original
+    # enum from gapic.
+    # ASSUMPTION: No existing types are ever renamed or deleted, we only try to
+    # detect cases when new types are introduced.
+    gapic_names = set(m.name for m in gapic_enums.StandardSqlDataType.TypeKind)
```
tswast: Nit: We use descriptive names even in set comprehensions, since `m` leaks into the local function scope. How about `type_`, instead?

bigquery/tests/unit/test_enums.py
```diff
@@ -0,0 +1,85 @@
...
+    assert set(doc_lines) <= set(gapic_enum.__doc__.splitlines())
+
+
+def test_standard_sql_types_enum_warning_on_new_added_types(gapic_enum):
+    class ReplacementEnum(enum.IntEnum):
+        """Fake enum with some new database types."""
+
+        INT64 = 2
+        TIMESTAMP = 19
+        TROOLEAN = 911  # One of {True, False, FileNotFound}. (OMG, help!)
```
tswast: 😄 Nice.

You're right that we do need to test our "is this enum in sync?" code. How about we make a helper function `def _find_unhandled_names` in this test module?

```
gapic_names = set(
    type_.name for type_ in gapic_enums.StandardSqlDataType.TypeKind
)
anticipated_names = (
    module_under_test._SQL_SCALAR_TYPES
    | module_under_test._SQL_NONSCALAR_TYPES
)
return gapic_names - anticipated_names
```

bigquery/google/cloud/bigquery/enums.py
plamut: Indeed, the warning was meant for us, not the users. But then again, a loud test failure is better suited for this, will refactor. :+1:

bigquery/google/cloud/bigquery/enums.py
plamut: A proto description change would not be detected, yes, but fortunately the only consequence would indeed be an orphaned weird line ending up in the docs, nothing that significant. Copying the docstring over can be viewed as best-effort IMO.

bigquery/tests/unit/test_enums.py
plamut: And why not... :+1:

bigquery/google/cloud/bigquery/enums.py
plamut: Leaking only occurs in _list_ comprehensions, and only in Python 2, but I agree, will use a nicer name. In a month even I probably won't remember that "m" stands for "member". :smile:

bigquery/tests/unit/test_enums.py
plamut: I'm fine with it, will refactor. Thanks for the suggestion.

bigquery/google/cloud/bigquery/__init__.py
```diff
@@ -130,6 +131,7 @@
     "Encoding",
     "QueryPriority",
     "SchemaUpdateOption",
+    "StandardSqlDataTypes",
```
tswast: This reminds me: let's add an Enums section to the API reference docs, too.

https://github.com/googleapis/google-cloud-python/blob/master/bigquery/docs/reference.rst

plamut: Added. Also had to inject an additional header line to the new enum's docstring, because Sphinx complained about a duplicate object description (took the first line - the INT64 member - which is of course directly copied from the gapic enum).

PR #8783, opened 2019-07-25
BigQuery: Use configurable bucket name for GCS samples data in systems tests.

### LINKED ISSUE #8550: BigQuery: `test_create_routine` test fails on VPC-SC tests
The `test_create_routine` test in `system.py` uses the `cloud-samples-data` GCS bucket, but this crosses a VPC boundary. The VPC team maintains a mirror of this sample data and would prefer we provide a way to override the bucket via an environment variable.

The Ruby samples take an environment variable `GCLOUD_TEST_SAMPLES_BUCKET` for the same purpose.

tswast:
This allows the VPC-SC team to use their private mirror which is within
the security boundary when testing BigQuery VPC-SC support.

Closes #8550.

### DIFF
--- a/bigquery/tests/system.py
+++ b/bigquery/tests/system.py
@@ -107,6 +107,11 @@ TIME_PARTITIONING_CLUSTERING_FIELDS_SCHEMA = [
     ),
 ]
 
+# The VPC-SC team maintains a mirror of the GCS bucket used for code
+# samples. The public bucket crosses the configured security boundary.
+# See: https://github.com/googleapis/google-cloud-python/issues/8550
+SAMPLES_BUCKET = os.environ.get("GCLOUD_TEST_SAMPLES_BUCKET", "cloud-samples-data")
+
 retry_storage_errors = RetryErrors(
     (TooManyRequests, InternalServerError, ServiceUnavailable)
 )
@@ -1877,7 +1882,7 @@ class TestBigQuery(unittest.TestCase):
             language="JAVASCRIPT",
             type_="SCALAR_FUNCTION",
             return_type=float64_type,
-            imported_libraries=["gs://cloud-samples-data/bigquery/udfs/max-value.js"],
+            imported_libraries=["gs://{}/bigquery/udfs/max-value.js".format(SAMPLES_BUCKET)],
         )
         routine.arguments = [
             bigquery.RoutineArgument(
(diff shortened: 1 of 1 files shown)


bigquery/tests/system.py
```diff
@@ -107,6 +107,11 @@
     ),
 ]
 
+# The VPC-SC team maintains a mirror of the GCS bucket used for code
+# samples. The public bucket crosses the configured security boundary.
+# See: https://github.com/googleapis/google-cloud-python/issues/8550
+SAMPLES_BUCKET = os.environ.get("GCLOUD_TEST_SAMPLES_BUCKET", "cloud-samples-data")
```
tseaver: Do we need to update `noxfile.py` to set this value in the environment?

tswast: No, the default of `cloud-samples-data` is fine for our purposes. The VPC-SC team has a separate Kokoro (or maybe something similar?) config where they'll set the necessary environment variable.

tswast:
```
___________________ test_bq_to_arrow_array_w_special_floats ____________________
module_under_test = <module 'google.cloud.bigquery._pandas_helpers' from '/tmpfs/src/github/google-cloud-python/bigquery/google/cloud/bigquery/_pandas_helpers.py'>
    @pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
    @pytest.mark.skipif(pyarrow is None, reason="Requires `pyarrow`")
    def test_bq_to_arrow_array_w_special_floats(module_under_test):
        bq_field = schema.SchemaField("field_name", "FLOAT64")
        rows = [float("-inf"), float("nan"), float("inf"), None]
        series = pandas.Series(rows, dtype="object")
        arrow_array = module_under_test.bq_to_arrow_array(series, bq_field)
        roundtrip = arrow_array.to_pylist()
        assert len(rows) == len(roundtrip)
        assert roundtrip[0] == float("-inf")
>       assert roundtrip[1] != roundtrip[1]  # NaN doesn't equal itself.
E       assert None != None
```

failure is unrelated. I think it's a behavior change / regression from `pyarrow`.

tseaver:
@tswast How should we track that regression?  All BigQuery CI is b0rked ATM.

tswast:
> @tswast How should we track that regression? All BigQuery CI is b0rked ATM.

Let's pin to `pyarrow==0.13.0` in the noxfile until I can reproduce the regression and file an issue on the Apache Arrow JIRA.

tswast:
Sent https://github.com/googleapis/google-cloud-python/pull/8784.

PR #8784, opened 2019-07-25
BigQuery: Pin pyarrow version in tests.

tswast:
This is a workaround for a possible regression in pyarrow handling of NaN vs
NULL in float columns.


tswast:
Found some documentation about this behavior: https://arrow.apache.org/docs/python/data.html#none-values-and-nan-handling

Since we are converting from pandas (which uses NaN as a placeholder for NULL for numpy reasons), this behavior is expected. I'll open a new PR that updates the test.

PR #8785, opened 2019-07-25
BigQuery: Treat NaN as NULL when converting from pandas to arrow.

tswast:
Fixes broken unit tests from latest pyarrow release.


tseaver:
Do we need a minimum version of PyArrow to account for the new behavior?

tswast:
> Do we need a minimum version of PyArrow to account for the new behavior?

The new behavior is present in 0.14.0 and later. I don't really have a strong opinion one way or the other on how NaN values are handled, so I'm hesitant to restrict the pyarrow version in `setup.py`.

PR #8797, opened 2019-07-26
Language: Preserve fixed enum reference from PR #8779 during synth.

tseaver:
See PR #8779.

Supersedes #8794.


tseaver:
This is a synth tweak only, which leaves the actual code unchanged after synth.  I plan to merge as soon as CI is green.

Issue #8715, opened 2019-07-19
Asset:  'test_batch_get_assets_history' systest flakes.


tseaver:
/cc @gaogaogiraffe (test added in PR #8613, configuration updated in #8627)

From [this failed Kokoro job](https://source.cloud.google.com/results/invocations/f5e4d76f-67c2-4631-9bd9-9a165fa598b5/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fasset/log):

```python
_____________ TestVPCServiceControl.test_batch_get_assets_history ______________
self = <test_vpcsc.TestVPCServiceControl object at 0x7fd2c52d70b8>
    @pytest.mark.skipif(
        PROJECT_INSIDE is None, reason="Missing environment variable: PROJECT_ID"
    )
    @pytest.mark.skipif(
        PROJECT_OUTSIDE is None,
        reason="Missing environment variable: GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT",
    )
    def test_batch_get_assets_history(self):
        client = asset_v1.AssetServiceClient()
        content_type = enums.ContentType.CONTENT_TYPE_UNSPECIFIED
        read_time_window = {}
        parent_inside = "projects/" + PROJECT_INSIDE
        delayed_inside = lambda: client.batch_get_assets_history(
            parent_inside, content_type, read_time_window
        )
        parent_outside = "projects/" + PROJECT_OUTSIDE
        delayed_outside = lambda: client.batch_get_assets_history(
            parent_outside, content_type, read_time_window
        )
>       TestVPCServiceControl._do_test(delayed_inside, delayed_outside)
tests/system/test_vpcsc.py:88:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
delayed_inside = <function TestVPCServiceControl.test_batch_get_assets_history.<locals>.<lambda> at 0x7fd2c5302d90>
delayed_outside = <function TestVPCServiceControl.test_batch_get_assets_history.<locals>.<lambda> at 0x7fd2c5302b70>
    @staticmethod
    def _do_test(delayed_inside, delayed_outside):
        if IS_INSIDE_VPCSC.lower() == "true":
>           assert TestVPCServiceControl._is_rejected(delayed_outside)
E           assert False
[...]

PR #8805, opened 2019-07-29
Docs: Update intersphinx mapping for requests.

busunkim96:
This will also need to be updated in the API level `docs/conf.py` files.


tseaver:
Failed Asset Kokoro job is due to known flaky systest, #8715.

PR #8831, opened 2019-07-30
Storage: Normalize 'lint' / 'blacken' support under nox.

tseaver:
- Pin version of black.
- Enforce blackening in lint.


tseaver:
This is a testing-only change.  The `docs` failure was due to an RTD outage last night.

PR #8840, opened 2019-07-31
Move maybe_fail_import() to top level test utils

plamut:
Following a [recent discussion](https://github.com/googleapis/google-cloud-python/pull/8734#discussion_r307532267), this PR moves a test utility from BigQuery to the top-level test utils, since it is considered useful outside of BigQuery, too.

### How to test
Primarily verify that the promoted utility function is general enough for the promotion.

### DIFF
--- a/bigquery/noxfile.py
+++ b/bigquery/noxfile.py
@@ -20,7 +20,11 @@ import shutil
 import nox
 
 
-LOCAL_DEPS = (os.path.join("..", "api_core[grpc]"), os.path.join("..", "core"))
+LOCAL_DEPS = (
+    os.path.join("..", "api_core[grpc]"),
+    os.path.join("..", "core"),
+    os.path.join("..", "test_utils"),
+)
 
 BLACK_PATHS = ("docs", "google", "samples", "tests", "noxfile.py", "setup.py")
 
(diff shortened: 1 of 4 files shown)


bigquery/noxfile.py
```diff
@@ -40,6 +40,7 @@ def default(session):
 
     dev_install = ".[all]"
     session.install("-e", dev_install)
+    session.install("-e", os.path.join("..", "test_utils"))
```
tseaver: Would it make more add it to `LOCAL_DEPS` instead?  Then any test session could have it available.

plamut: Well, perhaps someone might want to make use of that in other test sessions, thus why not, I'll move it.

tseaver:
FWIW, after two years, there are exactly *zero* uses of `maybe_fail_import` outside of the `python-bigquery` repo.

plamut:
Sounds like we need to run a large-scale AdWords campaign to promote its usage.

@tseaver Are you suggesting that it might be better to move this back to the `python-bigquery` repo?

tseaver:
@plamut 
> Are you suggesting that it might be better to move this back to the python-bigquery repo?

I'm not sure. I came across it while working on https://github.com/googleapis/python-test-utils/pull/43:  the function has no test coverage, which meant that I had to install a development version into a virtualenv along with `python-bigquery` so that I could test that I didn't break anything.

plamut:
@tseaver I see. Sounds like a one-off (ish) annoyance, thus might not be worth it. We could cover it with tests, however, I'll create a ticket to track it.

Issue #5574, opened 2018-07-06
StackDriver: ResourceExhausted: 429 Received message larger than max (5595329 vs. 4194304)


casa87:
Hi

When trying to query Stackdriver metrics I am getting the following error `ResourceExhausted: 429 Received message larger than max`.
I feel like the max rpc call should be far over 4Mb, or the client should be to assemble multiple messages together.

Code example : 
```python
from google.cloud import monitoring_v3
from google.cloud.monitoring import enums
from google.cloud.monitoring_v3 import query

client = monitoring_v3.MetricServiceClient()


cpu_ratio_max = query.Query(client, 'my-project', metric_type='kubernetes.io/container/cpu/request_utilization', days=3)
c = cpu_ratio_max.align(enums.Aggregation.Aligner.ALIGN_MAX, hours=24*3)
c.as_dataframe()

ResourceExhausted: 429 Received message larger than max (5595329 vs. 4194304)

```

Python 2.7.10 on OSX and virtual env
```
pip freeze 
appnope==0.1.0
backports.shutil-get-terminal-size==1.0.0
cachetools==2.1.0
certifi==2018.4.16
chardet==3.0.4
decorator==4.3.0
enum34==1.1.6
futures==3.2.0
google-api-core==1.2.1
google-auth==1.5.0
google-cloud-monitoring==0.30.0
googleapis-common-protos==1.5.3
grpcio==1.13.0
idna==2.7
ipython==5.7.0
ipython-genutils==0.2.0
numpy==1.14.5
pandas==0.23.1
pathlib2==2.3.2
pexpect==4.6.0
pickleshare==0.7.4
prompt-toolkit==1.0.15
protobuf==3.6.0
ptyprocess==0.6.0
pyasn1==0.4.3
pyasn1-modules==0.2.2
Pygments==2.2.0
python-dateutil==2.7.3
pytz==2018.5
requests==2.19.1
rsa==3.4.2
scandir==1.7
simplegeneric==0.8.1
six==1.11.0
traitlets==4.3.2
urllib3==1.23
wcwidth==0.1.7```

Thank you

crwilcox:
Hi @casa87 , I have submitted a PR that should address your issue. After merge I will release a new version for you to verify with.

crwilcox:
@casa87 I released this as v0.30.1 to PyPI this afternoon.

PR #8860, opened 2019-08-01
Logging: Remove gRPC size restrictions (4MB default)

busunkim96:
Applies fix from #5594 (in response to #5574) to Logging.

Also opened PR https://github.com/grpc/grpc/pull/19810 to see if the defaults can be changed for grpcio.


logging/synth.py
```diff
@@ -38,6 +38,14 @@
 # https://github.com/googleapis/gapic-generator/issues/2097
 s.replace("google/**/proto/*_pb2.py", r"(^.*$\n)*", r"# -*- coding: utf-8 -*-\n\g<0>")
 
+# the metric service grpc transport channel shouldn't limit the size of a grpc message at the default 4mb
```
tseaver:
```suggestion
# the logging service grpc transport channel shouldn't limit the size of a grpc message at the default 4mb
```

Issue #8646, opened 2019-07-11
BigQuery: Add support to Dataset for project_ids with org prefix.


ryanreno:
Similar to #7827. Can not create a Dataset using a project ID with org prefix.

#### Environment details

1. Python version: `Python 3.6.5`
2. google-cloud-<service> version:  
  Name: google-cloud-bigquery
  Version: 1.16.0

#### Steps to reproduce

  1. Follow the [Creating a Dataset guide](https://googleapis.github.io/google-cloud-python/latest/bigquery/usage/datasets.html#creating-a-dataset)

#### Code example
```python
from google.cloud import bigquery

client = bigquery.Client()

dataset_id = f"{client.project}.ryan_dataset".

# Construct a full Dataset object to send to the API.
dataset = bigquery.Dataset(dataset_id)
```

#### Stack trace
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-bfe46d108801> in <module>
----> 1 dataset = bigquery.Dataset(dataset_id)

~/.virtualenvs/bigquery/lib/python3.6/site-packages/google/cloud/bigquery/dataset.py in __init__(self, dataset_ref)
    368     def __init__(self, dataset_ref):
    369         if isinstance(dataset_ref, six.string_types):
--> 370             dataset_ref = DatasetReference.from_string(dataset_ref)
    371         self._properties = {"datasetReference": dataset_ref.to_api_repr(), "labels": {}}
    372 

~/.virtualenvs/bigquery/lib/python3.6/site-packages/google/cloud/bigquery/dataset.py in from_string(cls, dataset_id, default_project)
    305                 "Too many parts in dataset_id. Expected a fully-qualified "
    306                 "dataset ID in standard SQL format. e.g. "
--> 307                 '"project.dataset_id", got {}'.format(dataset_id)
    308             )
    309 

ValueError: Too many parts in dataset_id. Expected a fully-qualified dataset ID in standard SQL format. e.g. "project.dataset_id", got google.com:[project].ryan_dataset

```
I removed the project name from the stack trace.

ryanreno:
A workaround I found that works:

```Python
from google.clooud import bigquery

dataset_id = 'ryan_dataset'
dataset_ref = bigquery.DatasetReference(dataset_id, default_project=client.project)

dataset = bigquery.Dataset(dataset_ref)
dataset.location = 'US'
dataset = client.create_dataset(dataset)
```

PR #8877, opened 2019-08-01
BigQuery: Add support to Dataset for project_ids with org prefix.

Emar-Kar:
Closes: #8646

### DIFF
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -18,12 +18,20 @@ from __future__ import absolute_import
 
 import six
 import copy
+import re
 
 import google.cloud._helpers
 from google.cloud.bigquery import _helpers
 from google.cloud.bigquery.model import ModelReference
 from google.cloud.bigquery.routine import RoutineReference
 from google.cloud.bigquery.table import TableReference
+
+
+_W_PREFIX = re.compile(
+    r"""
+    (\S*)\:(?P<ref>\S*)""",
+    re.VERBOSE,
+)
 
 
 def _get_table_reference(self, table_id):
@@ -270,7 +278,8 @@ class DatasetReference(object):
             dataset_id (str):
                 A dataset ID in standard SQL format. If ``default_project``
                 is not specified, this must included both the project ID and
-                the dataset ID, separated by ``.``.
+                the dataset ID, separated by ``.`` or, single prefix usage is
+                also permitted.
             default_project (str):
                 Optional. The project ID to use when ``dataset_id`` does not
                 include a project ID.
@@ -283,6 +292,9 @@ class DatasetReference(object):
             >>> DatasetReference.from_string('my-project-id.some_dataset')
             DatasetReference('my-project-id', 'some_dataset')
 
+            >>> DatasetReference.from_string('prefix:my-project-id.some_dataset')
+            DatasetReference('my-project-id', 'some_dataset')
+
         Raises:
             ValueError:
                 If ``dataset_id`` is not a fully-qualified dataset ID in
@@ -290,21 +302,28 @@ class DatasetReference(object):
         """
         output_dataset_id = dataset_id
         output_project_id = default_project
-        parts = dataset_id.split(".")
+        with_prefix = _W_PREFIX.match(dataset_id)
+        if with_prefix is None:
+            parts = dataset_id.split(".")
+        else:
+            parts = with_prefix.group("ref").split(".")
 
         if len(parts) == 1 and not default_project:
             raise ValueError(
                 "When default_project is not set, dataset_id must be a "
                 "fully-qualified dataset ID in standard SQL format. "
-                'e.g. "project.dataset_id", got {}'.format(dataset_id)
+                'e.g. "project.dataset_id" or, single prefix usage '
+                'is also permitted e.g. "prefix:project.dataset_id" '
+                "got {}".format(dataset_id)
             )
         elif len(parts) == 2:
             output_project_id, output_dataset_id = parts
         elif len(parts) > 2:
             raise ValueError(
                 "Too many parts in dataset_id. Expected a fully-qualified "
-                "dataset ID in standard SQL format. e.g. "
-                '"project.dataset_id", got {}'.format(dataset_id)
+                'dataset ID in standard SQL format. e.g. "project.dataset_id" '
+                "or, single prefix usage is also permitted e.g. "
+                '"prefix:project.dataset_id" got {}'.format(dataset_id)
             )
 
         return cls(output_project_id, output_dataset_id)
@@ -555,7 +574,7 @@ class Dataset(object):
             full_dataset_id (str):
                 A fully-qualified dataset ID in standard SQL format. Must
                 included both the project ID and the dataset ID, separated by
-                ``.``.
+                ``.`` or, single prefix usage is also permitted.
 
         Returns:
             Dataset: Dataset parsed from ``full_dataset_id``.
@@ -563,6 +582,9 @@ class Dataset(object):
         Examples:
             >>> Dataset.from_string('my-project-id.some_dataset')
             Dataset(DatasetReference('my-project-id', 'some_dataset'))
+
+            >>> DatasetReference.from_string('prefix:my-project-id.some_dataset')
+            DatasetReference('my-project-id', 'some_dataset')
 
         Raises:
             ValueError:
--- a/bigquery/google/cloud/bigquery/dataset.py
+++ b/bigquery/google/cloud/bigquery/dataset.py
@@ -31,7 +31,6 @@ _W_PREFIX = re.compile(
     r"""
     (\S*)\:(?P<ref>\S*)
 """,
-    # re.VERBOSE,
 )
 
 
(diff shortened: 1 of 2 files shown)


IlyaFaer:
@Emar-Kar, you should run black reformat on `bigquery/dataset.py` file to get lint session OK.
Also there are not covered lines of code:
google/cloud/bigquery/dataset.py 309, 306->309
See:
https://source.cloud.google.com/results/invocations/5834d3bb-e5dc-4f3f-aeba-34ddf36780be/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigquery/log

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -283,28 +293,38 @@ def from_string(cls, dataset_id, default_project=None):
             >>> DatasetReference.from_string('my-project-id.some_dataset')
             DatasetReference('my-project-id', 'some_dataset')
 
+            >>> DatasetReference.from_string('prefix:my-project-id.some_dataset')
```
tswast: It is rare for non-Googlers to encounter such a project ID. I don't think it warrants a second code example in the docs.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -269,8 +278,9 @@ def from_string(cls, dataset_id, default_project=None):
         Args:
             dataset_id (str):
                 A dataset ID in standard SQL format. If ``default_project``
-                is not specified, this must included both the project ID and
-                the dataset ID, separated by ``.``.
+                is not specified, this must include both the project ID and
+                the dataset ID, separated by ``.`` or, single prefix usage is
```
tswast: I find the "single prefix" term to be confusing. I can't find any documentation on these kinds of project IDs. I suggest reverting the changes to the docstrings, since it will confuse most customers.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -283,28 +293,38 @@ def from_string(cls, dataset_id, default_project=None):
...
+            parts = dataset_id.split(".")
+        else:
+            parts = with_prefix.group("ref").split(".")
 
         if len(parts) == 1 and not default_project:
             raise ValueError(
                 "When default_project is not set, dataset_id must be a "
-                "fully-qualified dataset ID in standard SQL format. "
-                'e.g. "project.dataset_id", got {}'.format(dataset_id)
+                "fully-qualified dataset ID in standard SQL format, "
```
tswast: Ditto. "prefix" is both rare and confusing. Please revert the error message changes.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -283,28 +293,38 @@ def from_string(cls, dataset_id, default_project=None):
...
                 standard SQL format.
         """
         output_dataset_id = dataset_id
         output_project_id = default_project
-        parts = dataset_id.split(".")
+        with_prefix = _W_PREFIX.match(dataset_id)
+        if with_prefix is None:
+            parts = dataset_id.split(".")
+        else:
+            parts = with_prefix.group("ref").split(".")
```
tswast: I'm confused. What is this doing? I think the prefix needs to be part of the project ID, right?

Emar-Kar: From the issue's `Stack trace`:
The error occurs due to the prefix `google.com:`. Previously the passed string was separated only by `.`, what led to `ValueError` raising because of the `len(parts) > 2` at `google.com:[project].ryan_dataset`. As I see here prefix is not the part of the `Project ID` itself. I was trying to find out, how could I parse the string and fulfill both the previous and new format. That is why I decided to use regular expressions. Now, with template's help, it will separate prefix and solve several situations:
* `string-project.string_dataset` - will pass the template successfully as it was before;
* `prefix:string-project.string_dataset` - will group the part without prefix and then will divide it;
* `string-project:string_dataset` - if the `default_project` was not defined raises `ValueError`;
* `google.com:project:dataset_id` - same as above.

> ValueError: Too many parts in dataset_id. Expected a fully-qualified dataset ID in standard SQL format. e.g. "project.dataset_id", got google.com:[project].ryan_dataset

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -26,6 +27,14 @@
 from google.cloud.bigquery.table import TableReference
 
 
+_W_PREFIX = re.compile(
```
tswast: Can we pick a better name for this? Maybe `_PROJECT_PREFIX_PATTERN`?

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -26,6 +27,14 @@
...
+    r"""
+    (\S*)\:(?P<ref>\S*)
```
tswast: Since at least one character is required, this should probably be `\S+`, right?

Also, `ref` isn't all that meaningful to me. How about `remaining`, since it's everything after the `:` character?

bigquery/google/cloud/bigquery/dataset.py
tswast: Right, but it appears to me that we're discarding the prefix? Is that correct?

bigquery/tests/unit/test_dataset.py
```diff
@@ -186,11 +186,22 @@ def test_from_string(self):
         self.assertEqual(got.project, "string-project")
         self.assertEqual(got.dataset_id, "string_dataset")
 
+    def test_from_string_w_prefix(self):
+        cls = self._get_target_class()
+        got = cls.from_string("prefix:string-project.string_dataset")
+        self.assertEqual(got.project, "string-project")
```
tswast: Shouldn't this be `prefix:string-project`, since the prefix is actually part of the project ID?

bigquery/google/cloud/bigquery/dataset.py
Emar-Kar: Yeah, that is what I was thinking before this conversation. So, now I'm a bit confused. I thought the prefix is an extra part and should be just removed. But if it is actually the part of the `Project ID`, I'll need to reconfigure the pattern.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -26,6 +27,14 @@
 from google.cloud.bigquery.table import TableReference
 
 
+_PROJECT_PREFIX_PATTERN = re.compile(
+    r"""
+    (?P<prefix>\S+\:\S+)\.+(?P<remaining>\S*)
```
tswast: Now that we're matching this way, prefix isn't the right term. Should be `project_id`. Likewise, `remaining` should be renamed to `dataset_id`.

Also, instead of `\S`, we should be matching for characters other than `.`, that is `[^.]+`.

We want to match the whole string, so we should probably end this pattern with `$`.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -26,6 +27,14 @@
 from google.cloud.bigquery.table import TableReference
 
 
+_PROJECT_PREFIX_PATTERN = re.compile(
+    r"""
```
tswast: Why are we using a multi-line string here?

Emar-Kar: Just for the readability. I think I’ll switch this to the single line, after correcting the pattern implementation.

bigquery/google/cloud/bigquery/dataset.py
Emar-Kar: I renamed parts of the pattern, but the second comment about `[^.]` seems inappropriate to me. As we know the string could be `google.com:project.dataset`, that means that dot could be a part of the prefix. I checked couple of variants and as I see `\S` fits more.

bigquery/tests/unit/test_dataset.py
```diff
@@ -186,11 +186,22 @@ def test_from_string(self):
...
+        got = cls.from_string("prefix:string-project.string_dataset")
+        self.assertEqual(got.project, "prefix:string-project")
+        self.assertEqual(got.dataset_id, "string_dataset")
+
     def test_from_string_legacy_string(self):
         cls = self._get_target_class()
         with self.assertRaises(ValueError):
             cls.from_string("string-project:string_dataset")
 
+    def test_from_string_w_incorrect_prefix(self):
```
tswast: Let's add an additional test where the project ID / dataset ID contains an illegal `.` character. Another way to say that, is the string contains too many "parts". e.g. `google.com:project-id.dataset_id.table_id`. This should also fail with `ValueError`.

bigquery/google/cloud/bigquery/dataset.py
```diff
@@ -26,6 +27,9 @@
 from google.cloud.bigquery.table import TableReference
 
 
+_PROJECT_PREFIX_PATTERN = re.compile(r"(?P<project_id>\S+\:\S+)\.+(?P<dataset_id>\S+)$")
```
tswast: I think this will match patterns with too many `.` characters. Let's try something like:

```
(?P<project_id>\S+\:[^.]+)\.(?P<dataset_id>[^.]+)$
```

Emar-Kar: I see what you mean, sorry for misunderstanding. Appreciate your help.

Issue #8479, opened 2019-06-26
Bigtable: 'test_bigtable_list_tables' snippet flakes with '504 Deadline Exceeded'.


tseaver:
From [this Kokoro failure](https://source.cloud.google.com/results/invocations/2322354e-a3c8-4e30-8cac-2dbf3a814cad/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigtable/log):

```python
__________________________ test_bigtable_list_tables ___________________________
    def test_bigtable_list_tables():
        from google.cloud.bigtable import Client
        from google.cloud.bigtable import column_family
        client = Client(admin=True)
        instance = client.instance(INSTANCE_ID)
        table = instance.table("to_list")
        max_versions_rule = column_family.MaxVersionsGCRule(2)
>       table.create(column_families={"cf1": max_versions_rule})
docs/snippets.py:358:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/bigtable/table.py:251: in create
    initial_splits=splits,
google/cloud/bigtable_admin_v2/gapic/bigtable_table_admin_client.py:340: in create_table
    request, retry=retry, timeout=timeout, metadata=metadata
../api_core/google/api_core/gapic_v1/method.py:143: in __call__
    return wrapped_func(*args, **kwargs)
../api_core/google/api_core/retry.py:273: in retry_wrapped_func
    on_error=on_error,
../api_core/google/api_core/retry.py:182: in retry_target
    return target()
../api_core/google/api_core/timeout.py:214: in func_with_timeout
    return func(*args, **kwargs)
../api_core/google/api_core/grpc_helpers.py:59: in error_remapped_callable
    six.raise_from(exceptions.from_grpc_error(exc), exc)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
value = DeadlineExceeded(u'Deadline Exceeded',)
from_value = <_Rendezvous of RPC that terminated with:
	status = StatusCode.DEADLINE_EXCEED...all.cc","file_line":1046,"grpc_message":"Deadline Exceeded","grpc_status":4}"
>
    def raise_from(value, from_value):
>       raise value
[...]

PR #8879, opened 2019-08-01
Bigtable: Create table in list_table test, deadline exceed error fix

sangramql:
Issue #8479
Bigtable: 'test_bigtable_list_tables' snippet flakes with '504 Deadline Exceeded'.

_Issue:_
As per my understanding this used to occur intermittently when Google cloud takes longer than expected to create instance or table. So we added cool down time before creating table in an instance.
We used to observe this issue sometimes while working on bigtable issues. 

_Fix:_
This fix is added to avoid creating and deleting table for single test. 
Instead, common table can be created and used in tests which require table specific tests, except table creation test.

_Fix tested with following approaches:_
This issue is not consistent in production i suppose. And I was not able to reproduce even once.
I tried different approach to hit this issue,

1. Created multiple tables(>100) in test function.
2. Created instance and w/o cool down time immediately created multiple tables (> 100) in class setup.
3. Tried to created multiple instances (>5) and table in an instance, but got the following error
    after creating 6 instances. But didn’t get ‘deadline exceeded’ error though.
4. Executed parallel threads of snippets.py from nox.

Cool down time (timeout=100) code was avoided to hit the issue, but error didn't occur.
---code---
`operation = Config.INSTANCE.create(clusters=[cluster])`
`#We want to make sure the operation completes.`
`#operation.result(timeout=100) #this code is commented to reproduce issue.`
`Config.TABLE = Config.INSTANCE.table(TABLE_ID) Config.TABLE.create()`
---/code---

### DIFF
--- a/bigtable/docs/snippets.py
+++ b/bigtable/docs/snippets.py
@@ -36,6 +36,7 @@ from test_utils.system import unique_resource_id
 from test_utils.retry import RetryErrors
 from google.api_core.exceptions import NotFound
 from google.api_core.exceptions import TooManyRequests
+from google.api_core.exceptions import DeadlineExceeded
 from google.cloud._helpers import UTC
 from google.cloud.bigtable import Client
 from google.cloud.bigtable import enums
@@ -62,6 +63,7 @@ LABELS = {LABEL_KEY: str(LABEL_STAMP)}
 INSTANCES_TO_DELETE = []
 
 retry_429 = RetryErrors(TooManyRequests, max_tries=9)
+retry_504 = RetryErrors(DeadlineExceeded, max_tries=4)
 
 
 class Config(object):
@@ -91,7 +93,7 @@ def setup_module():
     # We want to make sure the operation completes.
     operation.result(timeout=100)
     Config.TABLE = Config.INSTANCE.table(TABLE_ID)
-    Config.TABLE.create()
+    retry_504(Config.TABLE.create)()
 
 
 def teardown_module():
--- a/bigtable/docs/snippets.py
+++ b/bigtable/docs/snippets.py
@@ -45,6 +45,7 @@ UNIQUE_SUFFIX = unique_resource_id("-")
 INSTANCE_ID = "snippet-tests" + UNIQUE_SUFFIX
 CLUSTER_ID = "clus-1-" + UNIQUE_SUFFIX
 APP_PROFILE_ID = "app-prof" + UNIQUE_SUFFIX
+TABLE_ID = "tabl-1" + UNIQUE_SUFFIX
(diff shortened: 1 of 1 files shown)


bigtable/docs/snippets.py
```diff
@@ -88,6 +90,8 @@ def setup_module():
     operation = Config.INSTANCE.create(clusters=[cluster])
     # We want to make sure the operation completes.
     operation.result(timeout=100)
+    Config.TABLE = Config.INSTANCE.table(TABLE_ID)
+    Config.TABLE.create()
```
tseaver: There isn't any point in creating this table:  no snippet actually uses it.

bigtable/docs/snippets.py
```diff
@@ -430,12 +426,6 @@ def test_bigtable_list_tables():
     tables_list = instance.list_tables()
     # [END bigtable_list_tables]
 
-    table_names = [table.name for table in tables_list]
-    try:
-        assert table.name in table_names
-    finally:
-        retry_429(table.delete)()
-
```
tseaver: Removing the table creation here also removes the value of the snippet as a test.  Rather, we should invoke `table.create()` with a longer timeout (the [configured timeout](https://github.com/googleapis/google-cloud-python/blob/6fc0de5bc489394a4c88f35325abed18fa5e6cb9/bigtable/google/cloud/bigtable_admin_v2/gapic/bigtable_table_admin_client_config.py#L48) is 130 seconds).

bigtable/docs/snippets.py
sangramql: This is not new code but been moved from list_tables snnipet, so list_tables snippet will be using it.
This is similar to snippets_table.py, where we create common table to be used in snippets where we require table snippets.
If you think this is not required here then we can move back to list_tables, but what I can see advantage of this is we do not need to explicitly delete this table, as it would be gone when instance is deleted in teardown.

bigtable/docs/snippets.py
sangramql: That is good catch. My bad, I missed the assert.
Added the assert back.

sangramql: As we are using existing table now, we don't need table.create() here.

sangramql:
Please review @tseaver @crwilcox

bigtable/docs/snippets.py
tseaver: Please add a retry here:

```python
from google.api_core.exceptions import DeadlineExceeded

retry_504 = RetryExceptions(DeadlineExceeded)
...
        retry_504(Config.TABLE.create)(...)

PR #8880, opened 2019-08-01
BigQuery: Add to_standard_sql() method to SchemaField

### LINKED ISSUE #8494: BigQuery: `SchemaField.to_standard_sql` and `SchemaField.from_standard_sql` methods
Some BigQuery APIs (Models, Routines) use a new standard SQL format `bigquery_v2.types.StandardSqlDataType` for schemas, while Tables still use the legacy SQL format. It would be desirable to be able to convert easily between these formats.

plamut:
Closes #8494.

This PR adds a method to convert a legacy table field instance to the new `StandardSqlField` type.

### How to test
Verify that the types are correctly mapped from the legacy to standard SQL, and that the new method does what it is supposed to do.


bigquery/google/cloud/bigquery/schema.py
```diff
@@ -146,6 +167,26 @@ def _key(self):
...
+            An instance of :class:`~google.cloud.bigquery_v2.types.StandardSqlField`.
+        """
+        sql_type = types.StandardSqlDataType()
+        sql_type.type_kind = LEGACY_TO_STANDARD_TYPES.get(
+            self.field_type, types.StandardSqlDataType.TYPE_KIND_UNSPECIFIED
+        )
+
+        # NOTE: No need to also handle the "ARRAY" composed type, the latter
+        # does not exist in legacy SQL types.
+        if sql_type.type_kind == types.StandardSqlDataType.STRUCT:  # noqa: E721
```
plamut: False positive here, `flake8` thinks that is about comparing types and wants to enforce `isinstance()` - but that would not work, as we actually compare integers here.

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -14,6 +14,27 @@
...
+from google.cloud.bigquery_v2 import types
+
+
+# SQL types reference:
+# https://cloud.google.com/bigquery/data-types#legacy_sql_data_types
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types
+LEGACY_TO_STANDARD_TYPES = {
+    "STRING": types.StandardSqlDataType.STRING,
+    "BYTES": types.StandardSqlDataType.BYTES,
+    "INTEGER": types.StandardSqlDataType.INT64,
```
tswast: We should have aliases for the standard SQL names, as well, since the backend also supports those. e.g. a row from "INT64" to types.StandardSqlDataType.INT64

bigquery/google/cloud/bigquery/schema.py
```diff
@@ -146,6 +167,26 @@ def _key(self):
...
+
+        Returns:
+            An instance of :class:`~google.cloud.bigquery_v2.types.StandardSqlField`.
+        """
+        sql_type = types.StandardSqlDataType()
+        sql_type.type_kind = LEGACY_TO_STANDARD_TYPES.get(
+            self.field_type, types.StandardSqlDataType.TYPE_KIND_UNSPECIFIED
+        )
+
+        # NOTE: No need to also handle the "ARRAY" composed type, the latter
```
tswast: It does exist, actually. It's just represented a little different. Look for `mode == "REPEATED"`

bigquery/google/cloud/bigquery/schema.py
plamut: For some reason I thought only the legacy names are used here, will update.

bigquery/google/cloud/bigquery/schema.py
plamut: Good to know, thanks!

plamut:
Support for standard SQL type names and simple arrays added. Still need to add support for arrays containing STRUCTs, do not merge yet.

plamut:
@tswast You mean sharing the conversion logic or the test for it?

Since the "target results" differ, there are differences between the two implementations, and they are both reasonably short, too, thus we might not actually gain that much by refactoring.

(but of course, if you see something "obvious" that would represent a good opportunity, we can add this improvement in another PR)

tswast:
Agreed. Not obvious to me right now how to share code without it getting more complex.

PR #8883, opened 2019-08-01
API Core: Fix race in 'BackgroundConsumer._thread_main'.

### LINKED ISSUE #7817: Core: Race condition in bidi.BackgroundConsumer?
#### Environment details
Any OS, any Python version, google-api-core==1.9.0.

#### Steps to reproduce
No actual steps, spotted this in [bidi.py](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/bidi.py) while working on a different issue. This is how the _"Am I paused?"_ [snippet](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/bidi.py#L536-L540) in `BackgroundConsumer._thread_main()` looks like:
```py
with self._wake:
    if self._paused:
        _LOGGER.debug("paused, waiting for waking.")
        self._wake.wait()
        _LOGGER.debug("woken.")

_LOGGER.debug("waiting for recv.")
response = self._bidi_rpc.recv()
```
The `_paused` state can be set / cleared by the [pause() and resume()](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/bidi.py#L605-L617) methods, respectively.

If paused, the code snippet from above blocks at the `_wake.wait()` call, and is unblocked some time after [_wake.notifyAll()](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/bidi.py#L617) is invoked in the `resume()` method. When `resume()` method notifies the waiting threads and releases the internal lock held by the `self._wake` condition, `_wake.wait()` tries to re-obtain that lock.

Now, suppose that some other thread invokes the `pause()` method in the meantime, and the latter obtains the `self._wake`'s lock before `_wake.wait()` can grab it. The `_paused` flag will again be set to `True`, and when `_wake.wait()` finally acquires the lock and resumes, the code will invoke `_bidi_rpc.recv()` in the _paused_ state.

For this reason the `self._paused` condition must be checked in a loop, and not in a single `if` statement, as the docs on [threading.Condition](https://docs.python.org/3.7/library/threading.html#condition-objects) state:

> The while loop checking for the application’s condition is necessary because `wait()` can return after an arbitrary long time, and the condition which prompted the `notify()` call may no longer hold true. This is inherent to multi-threaded programming.

#### Code example

I created a [demo script](https://gist.github.com/plamut/b8937bdf2f6663e3e9fc503c5ccf1983) that demonstrates why `if <condition>` is not sufficient, and subject to race conditions (requires Python 3.6+). The "worker" thread waits until a shared global number is set to 10, while five "changer threads" randomly change that number, and notify other threads if they change it to 10. 

Sometimes the "worker" is invoked too late, and another "changer" thread changes 10 to something else, causing the "worker" to continue running when the target condition is no longer true. Replacing `if not <condition>` with `while not <condition>` gets rid of the bug.

```
...
INFO     [2019-04-29 23:09:20,327] Thread-Changer-0: Changing number from  9 to 10
INFO     [2019-04-29 23:09:20,327] Thread-Changer-0: New number == 10, notifying others
INFO     [2019-04-29 23:09:20,327] Thread-Worker   : I see the number 10
INFO     [2019-04-29 23:09:20,327] Thread-Changer-1: Changing number from 10 to  4
INFO     [2019-04-29 23:09:20,327] Thread-Changer-4: Changing number from  4 to 10
INFO     [2019-04-29 23:09:20,327] Thread-Changer-4: New number == 10, notifying others
INFO     [2019-04-29 23:09:20,328] Thread-Changer-2: Changing number from 10 to  3
INFO     [2019-04-29 23:09:21,328] Thread-Changer-3: Changing number from  3 to  9
INFO     [2019-04-29 23:09:21,328] Thread-Changer-1: Changing number from  9 to  6
INFO     [2019-04-29 23:09:21,328] Thread-Changer-4: Changing number from  6 to  4
INFO     [2019-04-29 23:09:21,328] Thread-Worker   : the number is not 10, will wait for condition
INFO     [2019-04-29 23:09:21,328] Thread-Changer-0: Changing number from  4 to  6
INFO     [2019-04-29 23:09:21,329] Thread-Changer-2: Changing number from  6 to  9
INFO     [2019-04-29 23:09:22,329] Thread-Changer-3: Changing number from  9 to 10
INFO     [2019-04-29 23:09:22,329] Thread-Changer-3: New number == 10, notifying others
INFO     [2019-04-29 23:09:22,329] Thread-Changer-1: Changing number from 10 to  5
INFO     [2019-04-29 23:09:22,329] Thread-Worker   : I see the number 5
...
```

tseaver:
See #7817.

In addition to passing all `api_core` tests locally, I have verified that the `firestore` and `pubsub` system tests all run cleanly with this patch (an earlier version caused them to hang).

### DIFF
--- a/api_core/google/api_core/bidi.py
+++ b/api_core/google/api_core/bidi.py
@@ -642,7 +642,7 @@ class BackgroundConsumer(object):
                 # In the future, we could use `Condition.wait_for` if we drop
                 # Python 2.7.
                 with self._wake:
-                    if self._paused:
+                    while self._paused:
                         _LOGGER.debug("paused, waiting for waking.")
                         self._wake.wait()
                         _LOGGER.debug("woken.")
--- a/firestore/tests/system/test_system.py
+++ b/firestore/tests/system/test_system.py
@@ -865,6 +865,7 @@ def test_batch(client, cleanup):
     assert not document3.get().exists
 
 
+# This test would hang with an earlier version of the patch for #7817.
 def test_watch_document(client, cleanup):
     db = client
     collection_ref = db.collection(u"wd-users" + UNIQUE_RESOURCE_ID)
--- a/firestore/tests/system/test_system.py
+++ b/firestore/tests/system/test_system.py
@@ -865,7 +865,6 @@ def test_batch(client, cleanup):
     assert not document3.get().exists
 
 
-# This test would hang with an earlier version of the patch for #7817.
 def test_watch_document(client, cleanup):
     db = client
     collection_ref = db.collection(u"wd-users" + UNIQUE_RESOURCE_ID)
--- a/pubsub/tests/system.py
+++ b/pubsub/tests/system.py
@@ -134,7 +134,6 @@ def test_subscribe_to_messages(
     future.cancel()
 
 
-# This test would hang with an earlier version of the patch for #7817.
 def test_subscribe_to_messages_async_callbacks(
     publisher, topic_path, subscriber, subscription_path, cleanup
 ):
--- a/pubsub/tests/system.py
+++ b/pubsub/tests/system.py
@@ -134,6 +134,7 @@ def test_subscribe_to_messages(
     future.cancel()
 
 
+# This test would hang with an earlier version of the patch for #7817.
 def test_subscribe_to_messages_async_callbacks(
     publisher, topic_path, subscriber, subscription_path, cleanup
 ):


tseaver:
Text-to-Speach job fails due to Kokoro-internal glitch: "Cannot check out non-mergeable pull request as if merged.".

tseaver:
@plamut I updated the PR description / commit message to avoid closing #7817.  Can you please clarify how we might reproduce / fix the remaining part?

plamut:
### Reproducing the issue

I did not manage to reproduce it with a real application (the time window is narrow), but I came up with a test that detects if `recv()` is called in `BackgroundConsumer` when the latter is in paused state (should not happen) - https://gist.github.com/plamut/8f996fdc9113e8de2b2c050befb36ff6

The idea is to constantly pause/resume the consumer and check if the latter is indeed paused while `recv()` is being executed.

The test consistently fails on my machine, but also passes if one comments out the following:
```py
pause_resume_thread.start()
```
Can anyone confirm my findings and proof-read the test?

### Fixing the issue
Apparently, we must be holding the `self._wake` lock when receiving messages, otherwise another thread can pause the consumer in the meantime, and the messages will be received and delivered in a paused state.

However, simply indenting the following block into the `with self._wake:` context would introduce other, more serious, problems:
```py
_LOGGER.debug("waiting for recv.")
response = self._bidi_rpc.recv()
_LOGGER.debug("recved response.")
self._on_response(response)
```
Since `self._bidi_rpc.recv()` might block, holding the `self._wake` lock would also block stopping the consumer, for instance, because that operation would also try to obtain the very same `self._wake` lock ([details](https://github.com/googleapis/google-cloud-python/issues/7817#issuecomment-514409847)).

Is there a good way to make the `recv()` method non-blocking? Probably not?

### Estimation of the issue impact 
I'm not familiar with Firestore, thus I can only say for PubSub.

The background consumer gets paused when the streaming pull manager determines that the client currently has more than `MAX_LOAD` messages on its hands, and resumes the consumer when there is enough capacity.

If the consumer [main loop](https://github.com/googleapis/google-cloud-python/blob/master/api_core/google/api_core/bidi.py#L635-L653) detects the paused state too late, one extra batch of server messages will be received before pausing in the next iteration.

For the PubSub client, this would mean receiving one extra batch of messages that would sit in the holding buffer until they can be processed. The extra memory usage is thus bounded by the maximum size of a single response from the server - probably not a showstopper.

Issue #8142, opened 2019-05-24
BigQuery: get table schema if not supplied (and have pyarrow) in `load_table_from_dataframe`


tswast:
Follow-up to https://github.com/googleapis/google-cloud-python/pull/8105#discussion_r286998303

When a table schema isn't supplied in `load_table_from_dataframe`, try to get the existing table schema. This will prevent errors due to ambiguous pandas types (#7370) without having to explicitly provide a schema.

Note: this behavior is similar to that of pandas-gbq, which always fetches the table schema and then compares to make sure it's compatible with the dataframe schema.

https://github.com/pydata/pandas-gbq/blob/59228d9c20cee12b24caa5cc41d3f2e6c0337932/pandas_gbq/gbq.py#L1115-L1121

timocb:
Hi @tswast ! I was looking into how to solve this issue, because #8105 has closed my issue #8093. It would be great if we can do this in the background.

Would this be as simple as adding the following code in the `try` part of https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/client.py#L1543?

This code would get the schema of the destination table and applies it to the `job_config`, if there is no schema info available.
```python
if not job_config.schema:
    client = Client()
    job_config.schema = client.get_table(destination).schema
```

Note: I am not sure how you would initialize the client here.

tswast:
> Note: I am not sure if you want to initialize the client here.

A client object is already available as `self`.

> `job_config.schema = client.get_table(destination).schema`

This is necessary, but not sufficient. There are several cases to handle.

* [x] A: DataFrame has same set of columns as existing table. This is the easy case, as it's what `google.cloud.bigquery._pandas_helpers.dataframe_to_parquet` already expects.
* [ ] B: DataFrame has new columns that aren't present in the existing table.
* [ ] C: DataFrame is missing columns that are present in the existing table.

Note: B & C can both be true if there are some new columns and some missing columns in the DataFrame.

Also, for those users that are using fastparquet instead of pyarrow, I don't want to force them to have to use pyarrow, as that's a breaking change (though given the difference in behavior, we may want to consider dropping fastparquet as a supported serialization library).

timocb:
> A client object is already available as `self`.
Oh, of course :P

> There are several cases to handle.
How are these cases (B and C) expected to be handled?

[...]

PR #8938, opened 2019-08-05
BigQuery: Allow choice of compression when loading from dataframe

### LINKED ISSUE #7701: BigQuery: Allow choice of compression for loading from dataframe
`pandas.Dataframe.to_parquet()` allows a choice of compression, with the deafult of `snappy`. When using the `client.load_table_from_dataframe` the call to this function only allows the use of the default. 
Is it possible to allow a choice so that `gzip` can be used instead so that installing `snappy` on the system is not necessary

plamut:
Closes #7701.

This PR adds an optional parameter to `Client.load_table_from_dataframe()` method that allows selecting the compression method if directly serializing a dataframe to a parquet file.

### How to test
The issue description should be self-explanatory, but mind one of the [comments](https://github.com/googleapis/google-cloud-python/issues/7701#issuecomment-486021029):
> Also I'm actually a bit wary of exposing compression, since I'd like to keep the option open to change the file format we serialize to. Parquet happens to be the best match to BigQuery's data types right now, but it'd be good to keep the option open to move to something else in the future.

I opted for the name `parquet_compression` instead of `compression` to indicate that the parameter is specific to parquet serialization, and not applicable in all use cases of the method.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1493,10 +1493,10 @@ class Client(ClientWithProject):
                 schema is used to determine the correct data type conversion.
                 Indexes are not loaded. Requires the :mod:`pyarrow` library.
             parquet_compression (str):
-                 The compression method to use if intermittently serializing
-                 ``dataframe`` to a parquet file. Must be one of {"snappy",
-                 "gzip", "brotli"}, or ``None`` for no compression. Defaults
-                 to "snappy".
+                 [Beta] The compression method to use if intermittently
+                 serializing ``dataframe`` to a parquet file. Must be one of
+                 {"snappy", "gzip", "brotli"}, or ``None`` for no compression.
+                 Defaults to "snappy".
 
                  The argument is directly passed as the ``compression`` argument
                  to the underlying ``DataFrame.to_parquet()`` method.
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1494,11 +1494,16 @@ class Client(ClientWithProject):
                 Indexes are not loaded. Requires the :mod:`pyarrow` library.
             parquet_compression (str):
                  [Beta] The compression method to use if intermittently
-                 serializing ``dataframe`` to a parquet file. Must be one of
-                 {"snappy", "gzip", "brotli"}, or ``None`` for no compression.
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1491,6 +1492,15 @@ def load_table_from_dataframe(
                 column names matching those of the dataframe. The BigQuery
                 schema is used to determine the correct data type conversion.
                 Indexes are not loaded. Requires the :mod:`pyarrow` library.
+            parquet_compression (str):
+                 The compression method to use if intermittently serializing
```
tswast: Let's mark this as `[Beta]`. I'm hesitant to call this a GA feature since it's so dependent on an implementation detail and third-party dependencies.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -1527,7 +1537,7 @@ def load_table_from_dataframe(
                         PendingDeprecationWarning,
                         stacklevel=2,
                     )
-                dataframe.to_parquet(tmppath)
+                dataframe.to_parquet(tmppath, compression=parquet_compression)
```
tswast: We do need to modify `_pandas_helpers.dataframe_to_parquet` as well. See: https://github.com/googleapis/google-cloud-python/blob/1b3e82270a8ca72cbfd90a6a9a73b641a57111bd/bigquery/google/cloud/bigquery/_pandas_helpers.py#L230 The underlying [pyarrow.parquet.write_table](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html) function also takes a compression argument.

Long-term, I expect the `_pandas_helpers.dataframe_to_parquet` function to get used more often than the `dataframe.to_parquet` method. We'll want to start fetching the table schema if not provided and use that for pandas to BigQuery type conversions https://github.com/googleapis/google-cloud-python/issues/8142.

plamut: There is a mismatch between the two methods, the pyarrow's accepts a richer range of compression methods.

Having either two different compression parameters, or a single parameter that accepts different values depending on the context, could be confusing to the end users, thus I will only allow the compression methods supported by both.

However, if there are uses cases that would need specific support for LZO, LZ4, and ZSTD as well, please let me know. Probably there aren't, because to date the compression method has not been exposed anyway?

Issue #7696, opened 2019-04-11
Storage: generate_signed_url API documentation misleading for response_type/disposition


lrowe:
`Blob.generate_signed_url` response_type and response_disposition parameters do not over-ride stored content_type as documented. The signed URL's response-content-type query string parameter is ignored if content-type is set on object metadata.

The generate_signed_url API documentation states ([source](https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py#L349-L360)):
```
        :type response_disposition: str
        :param response_disposition: (Optional) Content disposition of
                                     responses to requests for the signed URL.
                                     For example, to enable the signed URL
                                     to initiate a file of ``blog.png``, use
                                     the value
                                     ``'attachment; filename=blob.png'``.
        :type response_type: str
        :param response_type: (Optional) Content type of responses to requests
                              for the signed URL. Used to over-ride the content
                              type of the underlying blob/object.
```

However as documented in [Signed URL Considerations](https://cloud.google.com/storage/docs/access-control/signed-urls#considerations):
> Query String Parameters like response-content-disposition and response-content-type are not verified by the signature. To force a Content-Disposition or Content-Type in the response, set those parameters in the object metadata.

(Documentation issue confirmed in Google Cloud Support case 19006315.)

PR #8942, opened 2019-08-06
Storage: Fix documentation misleading response_type signed_url

HemangChothani:
Fixes: #7696

### DIFF
--- a/storage/google/cloud/storage/_signing.py
+++ b/storage/google/cloud/storage/_signing.py
@@ -321,7 +321,7 @@ def generate_signed_url_v2(
 
     :type response_type: str
     :param response_type: (Optional) Content type of responses to requests for
-                          the signed URL. Ignored if content_type is set.
+                          the signed URL. Ignored if content_type is set on object/blob metadata.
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/_signing.py
```diff
@@ -321,7 +321,7 @@ def generate_signed_url_v2(
 
     :type response_type: str
     :param response_type: (Optional) Content type of responses to requests for
-                          the signed URL. Ignored if content_type is set.
+                          the signed URL. Ignored if content_type is set on object/blob metadata.
```
tseaver: Please wrap to the next line here (and the other similar docstrings).

storage/google/cloud/storage/client.py
```diff
@@ -475,7 +475,7 @@ def list_blobs(
             versions=versions,
             projection=projection,
             fields=fields,
-            client=self
+            client=self,
```
tseaver: Please rebase against the head of `master`, which already has this change (and the others below) from PR #8831.

Issue #4553, opened 2017-12-07
Add functions equivalent to create_rows and create_rows_json that create a table for you using a load job


nanodan:
1. BigQuery API
2. Linux, Ubuntu 16.04.3 LTS running Google Cloud Datalab 1.0.1
3. Python 2.7.12
4. Google cloud v 0.31.0
5. //
6. Create a dataframe or dict list and try to send to an existing table. The first request will raise no errors but nothing will be appended to the table. Running the same command again and rows are added.

This will not add rows:
```
from google.cloud import bigquery as bq
import time

client = bq.Client(project='project-name')
dataset = bq.DatasetReference('project-name', 'dataset-name')
tableref = bq.table.TableReference(dataset, 'users')
schema = [bq.SchemaField('email', 'STRING'), bq.SchemaField('id', 'STRING'), bq.SchemaField('added', 'TIMESTAMP')]
table = bq.Table(tableref, schema=schema)
create = client.create_table(table)
public_members = [{'email': 'testemail@email.com', 'id': '1234', 'added': '2017-12-01 13:13:13 UTC'}]
client.create_rows_json(table, public_members)
```

This will add exactly one row:
```
from google.cloud import bigquery as bq
import time

client = bq.Client(project='project-name')
dataset = bq.DatasetReference('project-name', 'dataset-name')
tableref = bq.table.TableReference(dataset, 'users')
schema = [bq.SchemaField('email', 'STRING'), bq.SchemaField('id', 'STRING'), bq.SchemaField('added', 'TIMESTAMP')]
table = bq.Table(tableref, schema=schema)
create = client.create_table(table)
public_members = [{'email': 'testemail@email.com', 'id': '1234', 'added': '2017-12-01 13:13:13 UTC'}]
client.create_rows_json(table, public_members)
time.sleep(15)
client.create_rows_json(table, public_members)
```

jonparrott:
@tswast any ideas on this one?

tswast:
I believe this is a backend issue with the streaming buffer taking a little time to get created beyond that of the table create call.

According to https://stackoverflow.com/a/41446002/101923

> If you delete or create a table, you must wait a least 2 minutes to start streaming data on it.

This may be do to the fact that the streaming buffer has a cache of table information, which gets refreshed every 60 seconds.

Closing since this isn't a client library issue.

jonparrott:
@tswast is the recommend alternative to just create tables ahead of time?

tswast:
[...]

PR #8992, opened 2019-08-08
BigQuery Storage: Add more in-depth system tests 

### LINKED ISSUE #8983: BigQuery Storage: Add more in-depth system tests covering all data formats and field data types
**Simple functional tests**

- [ ] Simple correctness: create a table with some initial data, create a read session over the table, and verify that the expected number of rows is returned.
- [ ] Filtering: create a table with some initial data, create a read session over the table with a push-down filter which excludes some data, and verify that the expected number of rows is returned. (Avro-only)
- [ ] Column selection: create a table with some initial data, create a read session over the table with a list of columns specified, and verify that the expected columns and rows are returned.
- [ ] Snapshot test: create a table with some initial data, load some additional data as a separate step, create a read session using the timestamp of the initial load, read the data, and verify that the initial data is returned and the additional data is not returned. (Avro-only)
- [ ] Column-partitioned table test: create a column-partitioned table, read from the table with a partition filter specified, and verify that only the expected data is returned. (Avro-only)
- [ ] Naturally-partitioned table test: create a date-partitioned table, read from the table with a partition filter specified, and verify that only the expected data is returned.
- [ ] Data decoding: create a table with at least one field of each type supported by BigQuery -- including numeric, geographic, etc. -- and verify that the fields are decoded successfully.
- [ ] Resuming a read at an offset. For example, we can have some using the Shakespeare samples table which reads half-way and then resumes.

Many of these tests can be implemented using a pre-created sample table if appropriate -- tests a, b, and c use the Shakespeare samples table in our internal tests, for example.

**Long-running tests**

- [ ] ~Open a set of streams in parallel and read the full contents of a large table -- our google3 internal equivalent uses the Wikipedia sample table (which is about 35GB in size) and runs for between 10 and 20 minutes. This test should detect issues with long-running streams and, over time, should eventually add coverage for transparent stream resumption in Java and Python.~ [swast] I don't think we want to block client presubmits with a 10-to-20 minute system test. I'll work with @shollyman and the backend team to figure out a more appropriate home for these long-running tests.

This was originally filed internally as bug 133243219.

plamut:
Closes #8983.

This PR adds additional system tests for BigQuery Storage. It _does not_ include the last item from the issue description, i.e. the long-running test, at least not until decided otherwise.

### Things to mind
- [x] One of the tests depends on the fix in #9001 to pass. Tests need to be re-run once that is merged.
- [x] ~~(see the comment below) There is currently an inconsistent type returned for datetime columns (string instead of a `datetime` instance), affecting the same test.~~
Created #9066 for that.

### DIFF
--- a/bigquery_storage/noxfile.py
+++ b/bigquery_storage/noxfile.py
@@ -118,6 +118,7 @@ def system(session):
         session.install("-e", local_dep)
     session.install("-e", "../test_utils/")
     session.install("-e", ".[fastavro,pandas,pyarrow]")
+    session.install("-e", "../bigquery/")
     session.install("-e", ".")
 
     # Run py.test against the system tests.
--- a/bigquery_storage/noxfile.py
+++ b/bigquery_storage/noxfile.py
@@ -111,6 +111,8 @@ def system(session):
     # Use pre-release gRPC for system tests.
     session.install("--pre", "grpcio")
 
+    session.install("protobuf")
+
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
     session.install("mock", "pytest")
--- a/bigquery_storage/noxfile.py
+++ b/bigquery_storage/noxfile.py
@@ -111,8 +111,6 @@ def system(session):
     # Use pre-release gRPC for system tests.
     session.install("--pre", "grpcio")
 
-    session.install("protobuf")
-
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
     session.install("mock", "pytest")
--- a/bigquery_storage/tests/system/conftest.py
+++ b/bigquery_storage/tests/system/conftest.py
@@ -201,7 +201,7 @@ def all_types_table_ref(project_id, dataset, bq_client):
     bq_client.delete_table(created_table)
 
 
-@pytest.fixture()
+@pytest.fixture(scope="session")
 def client():
     return bigquery_storage_v1beta1.BigQueryStorageClient()
 
--- a/bigquery_storage/tests/system/conftest.py
+++ b/bigquery_storage/tests/system/conftest.py
@@ -132,6 +132,37 @@ def col_partition_table_ref(project_id, dataset):
 
 
 @pytest.fixture
+def ingest_partition_table_ref(project_id, dataset):
+    from google.cloud import bigquery
+
+    bq_client = bigquery.Client()
+
+    schema = [
+        bigquery.SchemaField("shape", "STRING", mode="REQUIRED"),
+        bigquery.SchemaField("altitude", "INT64", mode="REQUIRED"),
+    ]
+    time_partitioning = bigquery.table.TimePartitioning(
+        type_=bigquery.table.TimePartitioningType.DAY,
+        field=None,  # use _PARTITIONTIME pseudo column
+    )
+    bq_table = bigquery.table.Table(
+        table_ref="{}.{}.ufo_sightings".format(project_id, dataset.dataset_id),
+        schema=schema,
+    )
+    bq_table.time_partitioning = time_partitioning
+
+    created_table = bq_client.create_table(bq_table)
+
+    table_ref = bigquery_storage_v1beta1.types.TableReference()
+    table_ref.project_id = created_table.project
+    table_ref.dataset_id = created_table.dataset_id
+    table_ref.table_id = created_table.table_id
+    yield table_ref
+
+    bq_client.delete_table(created_table)
+
+
+@pytest.fixture
 def all_types_table_ref(project_id, dataset):
     from google.cloud import bigquery
 
--- a/bigquery_storage/tests/system/conftest.py
+++ b/bigquery_storage/tests/system/conftest.py
@@ -16,15 +16,89 @@
 """System tests for reading rows from tables."""
 
 import os
+import uuid
 
 import pytest
 
 from google.cloud import bigquery_storage_v1beta1
 
 
-@pytest.fixture()
+_ASSETS_DIR = os.path.join(os.path.abspath(os.path.dirname(__file__)), "assets")
+
+
+@pytest.fixture(scope="session")
 def project_id():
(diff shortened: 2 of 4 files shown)


bigquery_storage/tests/system/test_reader.py
```diff
@@ -255,3 +259,71 @@ def test_column_partitioned_table(client, project_id, col_partition_table_ref):
...
+    }
+
+    result_copy = copy.copy(rows[0])
+    del result_copy["datetime_field"]
+    assert result_copy == expected_result
+
+    # Compare datetime separately, AVRO and PYARROW return different object types,
+    # although they should both represent the same value.
+    expected_pattern = re.compile(r"2005-10-26( |T)19:49:41")
+    assert expected_pattern.match(str(rows[0]["datetime_field"]))
```
plamut: Sanity check - is this expected, or should both AVRO and PYARROW return the exact same type?

Right now one returns a plain ISO datetime string, while the other returns a `Timestamp` object..

plamut:
FWIW, a system test failure with STRUCT records will be fixed in #9001.

bigquery_storage/tests/system/test_reader.py
plamut: Answer:
"Expected" is a `datetime.datetime` object with no timezone. Actual is a string. That's an open issue.

tseaver:
@plamut System tests are failing in setup with the error, `Field person_struct_field is type RECORD but has no schema`.

plamut:
@tseaver Indeed, that's the bug that will be fixed with #9001

bigquery_storage/noxfile.py
```diff
@@ -111,13 +111,16 @@ def system(session):
     # Use pre-release gRPC for system tests.
     session.install("--pre", "grpcio")
 
+    session.install("protobuf")
```
tswast: Why was this installation necessary? Do we need to update `setup.py`?

bigquery_storage/tests/system/conftest.py
```diff
@@ -16,17 +16,196 @@
...
+@pytest.fixture(scope="session")
 def project_id():
     return os.environ["PROJECT_ID"]
 
 
+@pytest.fixture(scope="session")
+def dataset(project_id):
+    from google.cloud import bigquery
+
+    bq_client = bigquery.Client()
```
tswast: We should make a fixture for `bq_client` so that we don't have to auth multiple times. Likewise, we should probably make a fixture for `credentials` and use those for both the `bq_client` and `client` fixtures.

bigquery_storage/tests/system/test_reader.py
```diff
@@ -15,9 +15,52 @@
...
+
 import pytest
+import pytz
 
+from google.cloud import bigquery
 from google.cloud import bigquery_storage_v1beta1
+from google.protobuf import timestamp_pb2
+
+
+def _add_rows(table_ref, new_data, partition_suffix=""):
```
tswast: FWIW, we have an open feature request for such a method at https://github.com/googleapis/google-cloud-python/issues/4553

I propose we call that proposed method something along the line of `load_table_from_json` to be consistent with `load_table_from_dataframe`. Maybe we should implement that first so that we aren't duplicating code here?

bigquery_storage/noxfile.py
plamut: Needed for the `Timestamp` protobuf message used in the snapshot test (included in the `TableModifiers` for the read session).

The `create_read_session`'s docstring says that the `table_modifiers` argument can also be a dict with the same structure - is there a canonical way of providing a valid non-protobuf timestamp, circumventing the type `google.protobuf.Timestamp`?

If not, we should probably indeed add `protobuf` as a dependency so that users can configure `table_modifiers` without import errors and manually installing additional packages.

bigquery_storage/tests/system/conftest.py
plamut: I like the proposal, will refactor. :+1:

bigquery_storage/tests/system/test_reader.py
plamut: Depends on how fast we want this to be merged. If it can wait for a bit more, we can add that other method requested in #4553 first.

tswast: If you'd rather merge this sooner rather than later, let's add a TODO with a link to that FR to remove the `_add_rows` function once #4553 is complete.

bigquery_storage/noxfile.py
tswast: > The `create_read_session`'s docstring says that the table_modifiers argument can also be a dict with the same structure

That doesn't actually work, but if it did the official JSON representation of timestamps is an ISO 8601 string. See: https://github.com/googleapis/gapic-generator/issues/2530 :-(

tswast: > If not, we should probably indeed add protobuf as a dependency so that users can configure table_modifiers without import errors and manually installing additional packages.

Yes, let's add a protobuf dependency. It will need to match with the required versions for grpc.

plamut: Thanks for confirming it. I also had trouble initially when using a dict (type mismatch errors, etc.), and found no quick workaround apart from providing a protobuf Timestamp.

plamut: FWIW, `protobuf` is already a sub-dependency (through `api-core`), and it appears that everything already works without explicitly installing it (kind of expected).

It was probably something on my machine that initially caused an `ImportError`, and cannot reproduce it anymore.

bigquery_storage/tests/system/test_reader.py
plamut: Let's go with the TODO then, there's value in having these tests in place.

Also linked this thread from #4553 so that we don't forget about it.

tswast:
I've filed https://github.com/googleapis/google-cloud-python/issues/9066 for the Avro DATETIME issue.

Issue #8475, opened 2019-06-25
Add `client_options` to constructors for manual clients.


busunkim96:
*List of clients is copied from the issue for client_info. See #7825.*

The client_options constructor argument provides a consistent way for options to be set on clients. Currently only API Endpoint can be set with client_options.

GAPIC clients will gets this change via the generator, but the clients below need to be updated manually.

Clients that need client_options:

- [x] BigQuery #8999 
- [x] Bigtable #9517
  - [x] data endpoint
  - [x] admin endpoint
- [x] Datastore #9055
- [x] DNS #9516
- [x] Error Reporting  #9152
- [x]  Firestore (and Firestore Beta) #9048
- [x]  Logging #9046
- [x]  Pub/Sub ([Publisher](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py), [Subscriber](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/subscriber/client.py#L43)) (done via `**kwargs`) 
- [x]  Resource Manager #9043
- [x]  Runtime Config #9045
- [x]  Spanner #9151
    - [x] Client.instance_admin_api
    - [x] Client.database_admin_api
    - [x] Database.spanner_api
- [x] Storage #9054
- [x] Trace and Trace V1 #9154
- [x] Translate (no GRPC version needed, because v3beta1 is pure-autogen) #8737

sduskis:
Cloud Bigtable (and maybe others) will need two different `client_options`, since it has a data and admin endpoint.

tseaver:
Spanner has three, accessed via:  `Client.instance_admin_api`, `Client.database_admin_api`, `Database.spanner_api`.

PR #8999, opened 2019-08-09
Add client options to bigquery constructor

Emar-Kar:
Toward #8475

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -30,7 +30,7 @@ version = "1.17.0"
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
     'enum34; python_version < "3.4"',
-    "google-cloud-core >= 1.0.0, < 2.0dev",
+    "google-cloud-core >= 1.0.3, < 2.0dev",
     "google-resumable-media >= 0.3.1",
     "protobuf >= 3.6.0",
 ]
--- a/bigquery/tests/unit/test_client.py
+++ b/bigquery/tests/unit/test_client.py
@@ -110,14 +110,14 @@ class TestClient(unittest.TestCase):
             project=self.PROJECT,
             credentials=creds,
             _http=http,
-            client_options={"api_endpoint": "https://www.foo-googleapis.com"},
+            client_options={"api_endpoint": Connection.DEFAULT_API_ENDPOINT},
         )
         self.assertIsInstance(client._connection, Connection)
         self.assertIs(client._connection.credentials, creds)
         self.assertIs(client._connection.http, http)
         self.assertIsNone(client.location)
         self.assertEqual(
-            client._connection.API_BASE_URL, "https://www.foo-googleapis.com"
+            client._connection.API_BASE_URL, Connection.DEFAULT_API_ENDPOINT
         )
 
     def test_ctor_w_empty_client_options(self):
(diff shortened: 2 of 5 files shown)


bigquery/tests/unit/test_client.py
```diff
@@ -106,11 +106,61 @@ def test_ctor_defaults(self):
 
         creds = _make_credentials()
         http = object()
-        client = self._make_one(project=self.PROJECT, credentials=creds, _http=http)
+        client = self._make_one(
+            project=self.PROJECT,
+            credentials=creds,
+            _http=http,
+            client_options={"api_endpoint": Connection.DEFAULT_API_ENDPOINT},
```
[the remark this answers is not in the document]
tswast: I see that we need a different test where we set `client_options` as a dictionary. Let's add such a test and set a value for `api_endpoint` to something other than the default value.

bigquery/tests/unit/test_client.py
```diff
@@ -106,11 +106,61 @@ def test_ctor_defaults(self):
...
+        creds = _make_credentials()
+        http = object()
+        client_options = ClientOptions()
+        client = self._make_one(
+            project=self.PROJECT,
+            credentials=creds,
+            _http=http,
+            client_options=client_options,
+        )
+        self.assertIsInstance(client._connection, Connection)
```
tswast: Since this test is for the `client_options` option, let's remove the asserts for the other Client properties.

bigquery/tests/unit/test_client.py
```diff
@@ -106,11 +106,61 @@ def test_ctor_defaults(self):
...
+        creds = _make_credentials()
+        http = object()
+        client_options = ClientOptions(Connection.DEFAULT_API_ENDPOINT)
+        client = self._make_one(
+            project=self.PROJECT,
+            credentials=creds,
+            _http=http,
+            client_options=client_options,
+        )
         self.assertIsInstance(client._connection, Connection)
```
tswast: Ditto. This test is for the `client_options` parameter, so we don't need to assert for the other properties. Those asserts are sufficiently covered in the "defaults" test.

bigquery/setup.py
```diff
@@ -30,7 +30,7 @@
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
     'enum34; python_version < "3.4"',
-    "google-cloud-core >= 1.0.0, < 2.0dev",
+    "google-cloud-core >= 1.0.3, < 2.0dev",
```
tswast: Maybe I missed the comment, but why are we bumping the minimum version?

Emar-Kar: I saw this change in the already merged PR (Translate: Add client options to translate_v2.), which I was using as an example.

tswast: @busunkim96 Is this dependency update required?

busunkim96: Yes, `google-cloud-core` 1.0.3 bumps the minimum `google-api-core` version to 1.14.0 (`client_options` was added in 1.13.0). See [core/CHANGELOG.md](https://github.com/googleapis/google-cloud-python/blob/master/core/CHANGELOG.md), [api_core/CHANGELOG.md](https://github.com/googleapis/google-cloud-python/blob/master/api_core/CHANGELOG.md)

PR #9001, opened 2019-08-09
BigQuery: Fix schema recognition of struct field types

plamut:
A struct field can be referred to as "RECORD" or "STRUCT", and this commit assures that the `to_api_repr()` logic is correct. It also moves `STRUCT_TYPES` from `_pandas_helpers` to `schema` where it belongs more naturally IMO.

Discovered this while writing the "all types" system test, and tried to create a STRUCT field using the  standard SQL type name `"STRUCT"`.

### How to test
 - Using the BigQuery client, try to create a table with a "STRUCT" column (using that particular type name):
    ```py

    bq_client = bigquery.Client()

    schema = [
        bigquery.SchemaField(
            "person_struct_field",
            "STRUCT",
            fields=(
                bigquery.SchemaField("name", "STRING"),
                bigquery.SchemaField("age", "INT64"),
            )
        ),
    ]
    bq_table = bigquery.table.Table(
        table_ref="{}.{}.active_users".format(PROJECT_ID, DATASET_ID),
        schema=schema,
    )

    bq_client.create_table(bq_table)
    ```

The server responds with an error:
> Field <field_name> is type RECORD but has no schema

The table is successfully created.

### DIFF
--- a/bigquery/google/cloud/bigquery/schema.py
+++ b/bigquery/google/cloud/bigquery/schema.py
@@ -17,7 +17,7 @@
 from google.cloud.bigquery_v2 import types
 
 
-STRUCT_TYPES = ("RECORD", "STRUCT")
+_STRUCT_TYPES = ("RECORD", "STRUCT")
 
 # SQL types reference:
 # https://cloud.google.com/bigquery/data-types#legacy_sql_data_types
@@ -152,7 +152,7 @@ class SchemaField(object):
 
         # If this is a RECORD type, then sub-fields are also included,
         # add this to the serialized representation.
-        if self.field_type.upper() in STRUCT_TYPES:
+        if self.field_type.upper() in _STRUCT_TYPES:
             answer["fields"] = [f.to_api_repr() for f in self.fields]
 
         # Done; return the serialized dictionary.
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/schema.py
```diff
@@ -17,6 +17,8 @@
 from google.cloud.bigquery_v2 import types
 
 
+STRUCT_TYPES = ("RECORD", "STRUCT")
```
tswast: I think we should make this "private" for now (leading underscore), as I'm having trouble thinking of a customer use-case for these constants.

plamut: `_pandas_helpers.py` make use of this as well, thus I would personally keep this "public", as `schema.STRUCT_TYPE` can serve other modules (even if those modules happen to be private).

If you have a strong opinion about it, I can still change it, just putting this up for consideration.

tswast: Other modules can use it just fine, even if it's "private". Any breaking changes in that case would be caught by our test suite.

tswast: We should always err on the side of not introducing new public APIs, since keeping things private gives us more flexibility.

plamut: Makes sense, will change.

PR #9020, opened 2019-08-12
feat(translate): generate v3

busunkim96:


translate/synth.py
```diff
@@ -19,7 +19,7 @@
 
 gapic = gcp.GAPICGenerator()
 common = gcp.CommonTemplates()
-versions = ["v3beta1"]
+versions = ["v3beta1", "v3"]
```
tseaver: Do we want to make `v3` the default version (copy over `translation.py` below)?

busunkim96: @nnegrey would it be alright to make `v3` the default version, or should we keep the default as `v2`? 

This will determine what gets imported by default.
```
import google.cloud.translate
```

The v2 is different from v3beta1 and v3 because there was a switch from a manual to an autogenerated client.

nnegrey: Oh that's a question above my pay grade. haha

Let me float this to Daniel Torres to check with the team. 

That could cause people's stuff to break if they updated the library and the default version switched? Yes?

busunkim96:
V3 is still under whitelist, so we have been asked to hold off on publishing this.

translate/synth.py
busunkim96: Yep, if they were importing from `google.cloud.translate` and didn't realize we'd switched the default from `v2` to `v3`.  We'd make note of it in the release notes, of course. 

I'll also float this in the Yoshi chatroom to see if folks have thoughts.

nnegrey: Last I heard, for now it would be best to leave default at V2. But that could change when they take this off hold.

busunkim96:
This is good to merge and release now.

PR #9037, opened 2019-08-16
Pub/Sub: update docstrings for client kwargs and fix return types uris

anguillanneuf:
Show examples of how to create
- publisher client with optional batch settings, retry settings, regional endpoints
- subscriber client with regional endpoints

Fix return types for
- `publish()`, `from_service_account_file()`, `list_topics()`, `list_topic_subscriptions()`
- `subscribe()`, `from_service_account_file()`, `list_subscriptions()`, `list_snapshots()`


anguillanneuf:
Thanks @plamut! Unlike publish with retry settings, for subscriber client, we don't expose retry settings in `client_config`. If you look closely, `client_config` in the GAPIC subscriber client is actually in [a deprecated](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/gapic/subscriber_client.py#L167-L175) state.

PR #9040, opened 2019-08-16
BigQuery: Add enum with SQL type names allowed to be used in SchemaField

### LINKED ISSUE #7632: BigQuery: Add enum with allowed data types for use in Schema

plamut:
Closes #7632.

This PR adds an enum of legacy SQL type names that can be used with in schema fields.

@tswast This enum is a bit different from the existing `StandardSqlDataTypes` enum which maps type names to their gapic integer codes. Let me know if I misunderstood the ticket.

### How to test
Make sure all of the [legacy SQL type names](https://cloud.google.com/bigquery/data-types#legacy_sql_data_types) are included in the enum (and nothing else).

### DIFF
--- a/bigquery/google/cloud/bigquery/enums.py
+++ b/bigquery/google/cloud/bigquery/enums.py
@@ -67,3 +67,20 @@ def _make_sql_scalars_enum():
 
 
 StandardSqlDataTypes = _make_sql_scalars_enum()
+
+
+# see also https://cloud.google.com/bigquery/data-types#legacy_sql_data_types
+class LegacySqlDataTypes(str, enum.Enum):
+    """Enum of legacy SQL types names."""
+
+    STRING = "STRING"
+    BYTES = "BYTES"
+    INTEGER = "INTEGER"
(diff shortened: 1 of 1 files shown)


plamut:
@tswast GEOGRAPHY is not listed among the [legacy types](https://cloud.google.com/bigquery/data-types#legacy_sql_data_types). Am I missing something?

tswast:
Thanks. Yes, all standard SQL types names are aliases in legacy SQL, so I think that makes sense.

Filed b/139704786 to document GEOGRAPHY as not supported in legacy SQL, but still a valid type where legacy SQL names are expected in the API.

PR #9047, opened 2019-08-16
Fix reference to library name

jlara310:
Seems like it should be google-cloud-firestore instead of google-cloud-datastore


busunkim96:
@jlara310 thank you for the patch!

Issue #8143, opened 2019-05-25
BigQuery: Add dryrun option to magic.


jpallas:
The BQ API supports the dry_run option, but the ipython magic does not.  It would be great if there were a parameter which, when set, would return the (sometimes vital) information about bytes to be read instead of running the query.  This would help put the Jupyter environment on a par with the BQ console.

tswast:
FYI: there's an open PR at https://github.com/googleapis/google-cloud-python/pull/8179 which will add `context.default_query_job_config`, so that this will be possible to set globally.

I agree that a parameter would make sense for this option. See #8179 for implementation and testing patterns.

tswast:
Likely for `--dryrun` queries we'll want to print out something other than an empty DataFrame the estimated bytes processed and maybe additional stats if relevant.

Maybe return a `QueryJob` instead of a `DataFrame` for inspection.

jpallas:
Yes, my thinking was to provide (at least) the same cost feedback that the BQ console interface does.

tswast:
Closed by #9067.

PR #9067, opened 2019-08-21
BigQuery: Add dry_run option to BigQuery magic

shubha-rajan:
See #8143
Adding the `--dry_run` flag returns a `QueryJob` instead of a pandas `DataFrame` as requested in the original issue. The `QueryJob` can also be stored in a variable if the `destination_var` magic argument is present


bigquery/google/cloud/bigquery/magics.py
```diff
@@ -324,6 +324,15 @@ def _run_query(client, query, job_config=None):
         "the context default_query_job_config.maximum_bytes_billed."
     ),
 )
+@magic_arguments.argument(
+    "--dry_run",
+    action="store_true",
+    default=False,
+    help=(
+        "Sets query to be a dry run to estimate costs"
```
plamut: A period and a space are missing at the end of the sentence, looks strange in the help output.

```suggestion
        "Sets query to be a dry run to estimate costs. "
```

shubha-rajan:
I'll investigate that error being printed out.  I agree that it would be confusing to users, and don't think it's outside of the scope of the PR

tswast:
When you run a dry run query, a job is not actually created, so fetching the results fails.

tswast:
Have you run this in a notebook, yet? I'm curious what the output looks like. We probably want to improve the `__str__` or `__repr__` functions on `QueryJob` to at least show [`estimated_bytes_processed`](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.estimated_bytes_processed), because that's what people running dryrun queries are going to care about.

shubha-rajan:
@tswast as of now, running it in a notebook is silent. I tried modifying `_run_query()` and that made the error @plamut was referring to go away. I think the next step is to look at modifying those functions in `QueryJob` 
image

shubha-rajan:
6efa408 allows the cell to output the total bytes processed. adding a variable argument allows the user to inspect values the returned `QueryJob`'s properties including `total_bytes_processed` and `total_bytes_billed`
image

shubha-rajan:
So currently, `destination_var` doesn't return anything if an error occurs by default, even when the `--dry_run` flag isn't present. I investigated adding in that functionality, but the problem is that if calling `_run_query` raises an exception, it won't return a `QueryJob`. I suspect getting around that would mean making changes to exception handling in `QueryJob` and/or `Client` and idk what might end up breaking because of that 😅 

As for printing out the SQL queries along with error messages, that's working as of the last commit. 
image

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2876,6 +2876,29 @@ def _format_for_exception(query, job_id):
 
         return template.format(job_id=job_id, header=header, ruler=ruler, body=body)
 
+    def _begin(self, client=None, retry=DEFAULT_RETRY):
+        """API call:  begin the job via a POST request
+
+        See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert
+
+        :type client: :class:`~google.cloud.bigquery.client.Client` or
```
tswast: Nit: We're using Google-style docstrings for new functions.

```suggestion
        Args:
            client (Optional[google.cloud.bigquery.client.Client]):
                The client to use. If not passed, falls back to the ``client``
                associated with the job object.
```

Style guide: http://google.github.io/styleguide/pyguide.html#383-functions-and-methods (We follow PEP-8 for everything except docstrings, though.)

See examples at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2876,6 +2876,29 @@ def _format_for_exception(query, job_id):
...
+                    ``NoneType``
+        :param client: the client to use.  If not passed, falls back to the
+                    ``client`` stored on the current dataset.
```
tswast: Dataset objects don't actually have `client` properties anymore, and haven't for a couple years. I've suggested some alternative language above. We should probably fix the docstring this was copied from, too.

bigquery/tests/unit/test_magics.py
```diff
@@ -626,6 +645,107 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
+    )
+    query_job_mock = mock.create_autospec(
+        google.cloud.bigquery.job.QueryJob, instance=True
+    )
+    run_query_patch = mock.patch(
+        "google.cloud.bigquery.magics._run_query", autospec=True
+    )
+
+    sql = "SELECT 17 AS num"
+    result = pandas.DataFrame([17], columns=["num"])
```
tswast: We don't need this `result` variable or the mock of `to_dataframe`, correct? We should be able to remove this from the dry run unit tests.

bigquery/tests/unit/test_magics.py
```diff
@@ -626,6 +645,107 @@ def test_bigquery_magic_without_bqstorage(monkeypatch):
...
```
tswast: Ditto. We should be able to remove `result` from the dry run unit tests.

bigquery/google/cloud/bigquery/job.py
plamut: I also noticed that, but then realized the method was based on one of the existing methods, and that method also uses Sphinx-style docstrings.

If changing the docstring to Google style, I suggest doing the same in the original parent class's method as well.

plamut: Ticket - https://github.com/googleapis/google-cloud-python/issues/9092

bigquery/tests/unit/test_magics.py
shubha-rajan: removed

bigquery/tests/unit/test_magics.py
shubha-rajan: removed

PR #9070, opened 2019-08-21
BigQuery: Add load_table_from_json() method to BQ client

### LINKED ISSUE #4553: Add functions equivalent to create_rows and create_rows_json that create a table for you using a load job

plamut:
Closes #4553.

This PR adds a helper client method `load_table_from_json()`.

### How to test
Check that the method is similar to e.g. `client.load_table_from_dataframe()` and in line with the requirements.

P.S.: I will address the last [issue comment](https://github.com/googleapis/google-cloud-python/issues/4553#issuecomment-521577646) separately once this is approved (replacing the `_add_rows()` helper in system tests is a separate refactoring task IMO).

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1566,6 +1566,82 @@ class Client(ClientWithProject):
         finally:
             os.remove(tmppath)
 
+    def load_table_from_json(
+        self,
+        json_rows,
+        destination,
+        num_retries=_DEFAULT_NUM_RETRIES,
+        job_id=None,
+        job_id_prefix=None,
+        location=None,
+        project=None,
+        job_config=None,
+    ):
+        """Upload the contents of a table from a JSON string or dict.
+
+        Arguments:
+            json_rows (Iterable[Dict[str, Any]]):
+                Row data to be inserted. Keys must match the table schema fields
+                and values must be JSON-compatible representations.
+            destination (Union[ \
+                :class:`~google.cloud.bigquery.table.Table`, \
+                :class:`~google.cloud.bigquery.table.TableReference`, \
+                str, \
+            ]):
+                Table into which data is to be loaded. If a string is passed
+                in, this method attempts to create a table reference from a
+                string using
+                :func:`google.cloud.bigquery.table.TableReference.from_string`.
+
+        Keyword Arguments:
+            num_retries (int, optional): Number of upload retries.
+            job_id (str): (Optional) Name of the job.
+            job_id_prefix (str):
+                (Optional) the user-provided prefix for a randomly generated
+                job ID. This parameter will be ignored if a ``job_id`` is
+                also given.
+            location (str):
+                Location where to run the job. Must match the location of the
+                destination table.
+            project (str):
+                Project ID of the project of where to run the job. Defaults
+                to the client's project.
+            job_config (google.cloud.bigquery.job.LoadJobConfig):
+                (Optional) Extra configuration options for the job. The
+                ``source_format`` setting is always set to
+                :attr:`~google.cloud.bigquery.job.SourceFormat.NEWLINE_DELIMITED_JSON`.
+
+        Returns:
+            google.cloud.bigquery.job.LoadJob: A new load job.
+        """
+        job_id = _make_job_id(job_id, job_id_prefix)
+
+        if job_config is None:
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1566,6 +1566,82 @@ def load_table_from_dataframe(
...
+                :attr:`~google.cloud.bigquery.job.SourceFormat.NEWLINE_DELIMITED_JSON`.
+
+        Returns:
+            google.cloud.bigquery.job.LoadJob: A new load job.
+        """
+        job_id = _make_job_id(job_id, job_id_prefix)
+
+        if job_config is None:
+            job_config = job.LoadJobConfig()
+        job_config.source_format = job.SourceFormat.NEWLINE_DELIMITED_JSON
```
tswast: I'm having second thoughts about modifying the input config object. See:

https://github.com/googleapis/google-cloud-python/blob/fcf99ce0b6334ab9f58e692fc318c578478f4a94/bigquery/google/cloud/bigquery/client.py#L1523-L1529

We might want to refactor that code into a helper function for creating or copying a `LoadJobConfig`.

tswast: Also, if there isn't a schema set, we should enable schema autodetect for them.

bigquery/tests/system.py
```diff
@@ -826,6 +826,43 @@ def test_load_table_from_dataframe_w_explicit_schema(self):
         self.assertEqual(tuple(table.schema), table_schema)
         self.assertEqual(table.num_rows, 3)
 
+    def test_load_table_from_json_basic_use(self):
```
tswast: Can we also add a system test with an autodetected schema?

plamut: Yes. We. Can. :)

bigquery/google/cloud/bigquery/client.py
plamut: Is there a reason for not simply deep-copying the entire `LoadJobConfig` instance instead of spearately deep-copying the `_properties` attribute? The only other private attribute seems to be the `_job_type` string.

plamut:
Pushed new commits, but GitHub does not detect them. Also tried rebasing the branch, still no luck. Maybe closing and re-opening will work.

bigquery/google/cloud/bigquery/client.py
tswast: One potential reason is that `_properties` can be used to send values to the API for Alpha releases that aren't yet supported in client libraries. Not sure if deep copy of `LoadJobConfig` would carry those over.

`_properties` is the real core of the object, since that's what defines what we end up actually including in the API request.

PR #9077, opened 2019-08-22
BigQuery: Add support for unsetting LoadJobConfig schema

### LINKED ISSUE #9074: BigQuery: LoadJobConfig.schema setter should accept None

There is no (public) way to unset a schema after you've set one. Even though the client doesn't distinguish between unset schema and an empty schema, the backend API does.

When the set schema is `None`, remove the whole schema property from the underlying `_properties`.

We could always send `None` if a schema is empty, but this wouldn't be right, either.

 **Additional context**

Needed for https://github.com/googleapis/google-cloud-python/pull/9064

plamut:
Closes #9074.

This PR adds a public way of unsetting a schema from a `LoadJobConfig`.

### How to test
Create a `LoadJobConfig` with a `schema` attribute, then set `schema` to `None` --> the schema should be cleared.

### DIFF
--- a/bigquery/tests/unit/test_job.py
+++ b/bigquery/tests/unit/test_job.py
@@ -1504,6 +1504,19 @@ class TestLoadJobConfig(unittest.TestCase, _Base):
             config._properties["load"]["schema"], {"fields": [full_name_repr, age_repr]}
         )
 
+    def test_schema_setter_unsetting_schema(self):
+        from google.cloud.bigquery.schema import SchemaField
+
+        config = self._get_target_class()()
+        config._properties["load"]["schema"] = [
+            SchemaField("full_name", "STRING", mode="REQUIRED"),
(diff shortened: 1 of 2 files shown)


bigquery/tests/unit/test_job.py
```diff
@@ -1504,6 +1504,19 @@ def test_schema_setter(self):
...
+
+        config = self._get_target_class()()
+        config._properties["load"]["schema"] = [
+            SchemaField("full_name", "STRING", mode="REQUIRED"),
+            SchemaField("age", "INTEGER", mode="REQUIRED"),
+        ]
+
+        config.schema = None
+        self.assertNotIn("schema", config._properties["load"])
+        config.schema = None  # no error, idempotent operation
```
plamut: This extra assertion is to catch fragile implementations such as `del self._properties["load"]["schema"]`.

PR #9084, opened 2019-08-23
BigQuery: Allow specifying index data type in partial schema to `load_table_from_dataframe`.

### LINKED ISSUE #5572: [BigQuery] Load to table from dataframe always loads an index when using client.load_table_from_dataframe()
client.load_table_from_dataframe() results in the dataframe index being loaded into the bigquery table. 

Can the capability to load data to a table from a dataframe without having to load the index be implemented?

tswast:
Closes #5572.

If an index (or level of a multi-index) has a name and is present in the
schema passed to `load_table_from_dataframe`, then that index will be
serialized and written to the table. Otherwise, the index is omitted
from the serialized table.

Since I found it to be difficult to debug typos in column/index names in
the schema, I have also hardened the error messages to indicate when
unknown field values are found.

Remaining items:

- [ ] Update unit tests to account for new `ValueError`s for missing / extra columns.
- [ ] Fix tests for moved sample `samples/load_table_dataframe.py`.
- [ ] Add unit tests to account for new index behavior, especially some tests with MultIndex DataFrames.


tswast:
Thought: if an index column is requested, but we end up wanting to return a schema of `None`, that's an error, because the requested index column might not be written or it might not be written with the correct data type.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -245,6 +267,21 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
...
+        return dataframe[name]
+
+    if isinstance(dataframe.index, pandas.MultiIndex):
+        if name in dataframe.index.names:
+            return dataframe.index.get_level_values(name)
+    else:
+        if name == dataframe.index.name:
+            return dataframe.index.to_series()
+
+    raise ValueError("column or index '{}' not found.".format(name))
```
plamut: Formatting the error message will fail under Python 2 if index name contains non-ASCII characters, and a `UnicodeEncodeError` is probably not all that informative.

It can happen if a non-ASCII field name ends up in `bq_schema` passed to the  `dataframe_to_arrow()` helper, although that's probably unlikely, given that non-ASCII names are not allowed in table column names.

bigquery/samples/tests/test_load_table_dataframe.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+pytest.importorskip("pandas")
+pytest.importorskip("pyarrow")
+
+
+def test_load_table_dataframe(client, random_table_id):
+    load_table_dataframe.load_table_dataframe(client, random_table_id)
+
+    column_names = [field.name for field in table.schema]
```
plamut: Where does the name `table` come from? Should that actually be returned from the `load_table_dataframe()` function?

**Edit:** Aha, the test case was subsequently updated.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -187,24 +187,46 @@ def bq_to_arrow_array(series, bq_field):
...
         for name in dataframe.index.names:
-            if name:
+            if name and name not in column_names:
                 values = dataframe.index.get_level_values(name)
                 columns_and_indexes.append((name, values.dtype))
     else:
         if dataframe.index.name:
             columns_and_indexes.append((dataframe.index.name, dataframe.index.dtype))
 
+    # Add columns last so that if you iterate over the list, the column values overwrite any indexes with the same name.
```
plamut: Is this comment necessary? Don't we omit indexes that have a name equal to one of the column names?

Do we _not_ check if `dataframe.index.name` clashes with a column name on purpose? (we only do it for MultiIndex)

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -511,9 +514,175 @@ def test_bq_to_arrow_schema_w_unknown_type(module_under_test):
...
+    expected = [
+        ("a_series", pandas.api.types.pandas_dtype("int64")),
+        ("b_series", pandas.api.types.pandas_dtype("float64")),
+        ("c_series", pandas.api.types.pandas_dtype("object")),
+    ]
+    assert columns_and_indexes == expected
+
+
+@pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+def test_list_columns_and_indexes_with_index(module_under_test):
```
plamut: (nit)
Maybe rename to `..._with_named_index` to emphasize the "named" test case?

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -511,9 +514,175 @@ def test_bq_to_arrow_schema_w_unknown_type(module_under_test):
...
+        schema.SchemaField("str_col", "STRING"),
+        schema.SchemaField("int_col", "INTEGER"),
+        schema.SchemaField("nullable_int_col", "INTEGER"),
+        schema.SchemaField("str_index", "STRING"),
+    )
+    df_data = collections.OrderedDict
+    dataframe = pandas.DataFrame(
+
+    )
+    assert False
```
plamut: Something missing here in this test? This assertion always fails...

bigquery/tests/unit/test__pandas_helpers.py
```diff
@@ -667,17 +665,66 @@ def test_list_columns_and_indexes_with_multiindex(module_under_test):
...
+        names=["str_index", "int_index", "dt_index"],
     )
-    assert False
+    dataframe = pandas.DataFrame(df_data, index=df_index)
+
+    arrow_table = module_under_test.dataframe_to_arrow(dataframe, bq_schema)
+
+    assert arrow_table.schema.names == [
+        "str_index",
+        "dt_index",
```
plamut: Why is `int_index` omitted, here? It does not seem to clash with any of the column names.
(I might have misunderstood something)

plamut:
@tswast Please just blacken `test__pandas_helpers.py`, the lint check complains.

PR #9086, opened 2019-08-23
Revert "[CHANGE ME] Re-generated pubsub to pick up changes in the API or client library generator."

plamut:
Reverts googleapis/google-cloud-python#9078

As per offline discussion, we might not want to have these changes in the upcoming GA release, thus reverting them.


plamut:
As per discussion on chat - we do not need to revert the changes. They are similar to Ordering Keys, which have already been merged and documented quite some time ago, despite the Python client not supporting them yet.

plamut:
As per discussion on chat - we do not need to revert the changes. They are similar to Ordering Keys, which have already been merged and documented quite some time ago, despite the Python client not supporting them yet.

Issue #8989, opened 2019-08-07
BigQuery: deprecate `client.dataset()`


tswast:
See: https://github.com/googleapis/google-cloud-python/pull/7753

We should recommend the simpler "use a string instead of a reference object" in most cases.

We still call `.dataset()` 96 times in our samples at `docs/snippets.py`. I would feel better about adding the deprecation notice after we update the rest of our samples.

tswast:
As part of these sample updates, I'd also like to move the snippets to their own sample files in the `samples/` directory. See: https://github.com/googleapis/google-cloud-python/pull/7683 for examples.

IlyaFaer:
@tswast, taken to work

PR #9136, opened 2019-08-28
refactor(bigquery): update code samples to use strings for table and dataset IDs

Emar-Kar:
Towards #8989 
This PR contains five snippets:
- client_list_jobs
- client_query
- copy_table
- table_exists
- table_insert_rows


IlyaFaer:
@googlebot I consent

Emar-Kar:
@tswast Sure, I will change the name of the PR tomorrow. Plus I need to handle some problems with tests for one of the samples.

Emar-Kar:
Kokoro-test failed in the `snippets.py`  with an internal server [error](https://source.cloud.google.com/results/invocations/efa1f3f8-67a0-47d6-978f-0d9f30da8f6c/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fbigquery/log).

plamut:
Yup, happens on `master` as well, it is not caused by this PR.

bigquery/docs/snippets.py
```diff
@@ -13,11 +13,9 @@
 # limitations under the License.
 
 """Testable usage examples for Google BigQuery API wrapper
-
```
plamut: This and a few other similar blank lines should probably not be deleted?

By convention the docstring title line should be separated from the body with a blank line. Or was there some kind of a linter check that complained about it?

Emar-Kar:
@plamut New samples show to the user how strings as the ID work. Look at the referenced issue for more details about Tim's idea.

bigquery/docs/snippets.py
Emar-Kar: `snippets.py` file is gonna be deprecated. Maybe I deleted this string accidentally. And didn't notice it. If Tim will find it critical, I put it back.

plamut: I see. If the file is going to be deprecated, then this is probably indeed a lesser concern

plamut:
@Emar-Kar Of course... I was apparently looking at `snippets.py` when writing the review comment, my bad. The number of commits is huge, and in retrospect I should have probably looked at the final set of changes right from the start.

Emar-Kar:
@plamut yeah, the amount of commits is really huge. This is my bad. I’ll fix it with the next part.

bigquery/samples/copy_table.py
```diff
@@ -0,0 +1,45 @@
...
+    # client = bigquery.Client()
+
+    # TODO(developer): Set dataset_id to the ID of the dataset where to copy a table.
+    # dataset_id = "your-project.your_dataset"
+
+    # TODO(developer): Set table_id to the ID of the original table.
+    # table_id = "your-project.your_dataset.your_table_name"
+
+    orig_table = client.get_table(table_id)  # Make an API request.
+    dataset = client.get_dataset(dataset_id)  # Make an API request.
```
tswast: Calls to `get_table` and `get_dataset` are unnecessary. Let's use

```
# TODO(developer): Set source_table_id to the ID of the original table.
# source_table_id = "your-project.source_dataset.source_table"

# TODO(developer): Set destination_table_id to the ID of the destination table.
destination_table_id = "your-project.destination_dataset.destination_table"
```

bigquery/samples/client_query.py
```diff
@@ -0,0 +1,38 @@
...
+        "LIMIT 100"
+    )
+    query_job = client.query(
+        query, location="US"  # Must match the destination dataset(s) location.
+    )  # Make an API request.
+
+    print("The query data:")
+    for row in query_job:
+        # Row values can be accessed by field name or index
+        print(row)
```
tswast: Let's demonstrate that fields can be accessed by name or index.

```suggestion
        print("name={}, count={}".format(row[0], row["count"]))
```

bigquery/samples/client_query.py
```diff
@@ -0,0 +1,38 @@
...
+def client_query(client):
+
+    # [START bigquery_query]
+    # TODO(developer): Import the client library.
+    # from google.cloud import bigquery
+
+    # TODO(developer): Construct a BigQuery client object.
+    # client = bigquery.Client()
+
+    query = (
```
tswast: Since we want to show more than one field in the print logic, let's select more than one column.

```suggestion
    query = """
    SELECT name, SUM(number) as total_people
    FROM `bigquery-public-data.usa_names.usa_1910_2013`
    WHERE state = 'TX'
    GROUP BY name, state
    ORDER BY total_people DESC
    LIMIT 20
    """
```

bigquery/samples/tests/test_client_query.py
```diff
@@ -0,0 +1,24 @@
...
+
+from .. import client_query
+
+
+def test_client_query(capsys, client):
+
+    client_query.client_query(client)
+    out, err = capsys.readouterr()
+    assert "The query data:" in out
+    assert "Row(" in out
```
tswast: Since we're using the `usa_1910_2013` table, the data won't change. We can use a specific value in our tests.

```suggestion
    assert "name=James, count=272793" in out
```

Emar-Kar:
@tswast I'm trying to be grammatically correct and need your help here as you are a native speaker. There are two comment lines (added in #9212):
- `# Make an API request.`
- `# Waits for the job to complete.`

I assume it should be **Makes** in the first variant.

bigquery/samples/copy_table.py
Emar-Kar: I also decided to move the `num_rows` assertion to the test file.
code_1

tswast:
Thanks for your patience. I've been travelling a lot lately, but now I'm back.

Re: `Make and API request` vs `Makes an API request`, I'd interpret these in two ways:

1. (imperative) You, the developer, should `Make and API request`.
2. (descriptive) This line of code `Makes an API request`.

Since our code samples are included in how-to guides, our technical writing style guide requires (1) imperative.

Emar-Kar:
Yeah, sorry. I pushed the commit with `-s` modification today. I'll revert it tomorrow. Thank you, appreciate your help.

Emar-Kar:
@tswast One more thing with the comments. I also added `Waits for the job to complete` lines in #9212. Maybe it is required to be without `-s` too?

tswast:
> @tswast One more thing with the comments. I also added Waits for the job to complete lines in #9212. Maybe it is required to be without -s too?

Yes, without the `-s` would more closely match our tech writing style guide.

Issue #9105, opened 2019-08-26
BigQuery: preview / list_rows command for %%bigquery Jupyter magic


tswast:
**Is your feature request related to a problem? Please describe.**

When you want to get all rows in a table (or just a preview), it's more expensive to run a `SELECT *` query than the alternatives.

 **Describe the solution you'd like**

* [x] If there is no whitespace in the query text, assume it's a table ID. Call `list_rows` (possibly with the BigQuery Storage API option, if that's set).
* [x] To allow for a preview, add a `--max_results` option to the `%%bigquery` magic. If not set, get all rows. If set, download at most `max_results` rows.

 **Describe alternatives you've considered**

* Add a special `%%bigquery preview` command. I prefer looking to see if the query text is just a table name, as that will be more consistent with pandas-gbq. Also, we'd prefer to limit the `%%bigquery` magic to just "queries". I think entering just a table ID as query text is clear that it's different from `SELECT *` (anti-pattern) but also the intention is clear.
* Client-side query processing to automatically detect `SELECT * [LIMIT N]` queries client-side. I think it could make sense to parse queries client-side in just the `%%bigquery` magic (or opt-in via `QueryJobConfig` option), but this is much more complex than checking if there is any whitespace in query text.

 **Additional context**

Related feature requests in pandas-gbq:

* List rows without a query. https://github.com/pydata/pandas-gbq/issues/266
* Limit rows returned. https://github.com/pydata/pandas-gbq/issues/102

plamut:
With #9170 merged, this issue can now be closed.

PR #9147, opened 2019-08-29
BigQuery: Add list rows and --max_results option to %%bigquery magic

shubha-rajan:
See #9105. Creating draft for visibility.

Adds option to pass a table id instead of a SQL query to `%%bigquery` cell magic as a cost-saving alternative to `SELECT *` queries. `--max_results` option limits the number of rows read. The returned `pandas.DataFrame` can be saved to a variable by passing a `destination_var` argument.

image

TODO:

- [x] Fix coverage failures

- [x] Test that running cell magic with table ID instead of query works with `bqstorage_client` set

- [x] Add tests for failure cases- handles failure cases when table IDs are passed instead of queries:
image

- [x] ~~`max_results` is currently not working with regular SQL queries~~ - fixed! See screenshot below
image

To get this working, I ended up adding `max_results` as a property of `QueryJobConfig`, but if that wasn't the right call, I can refactor to pass `max_results` as a separate parameter.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -2162,7 +2162,7 @@ class QueryJobConfig(_JobConfig):
 
     @max_results.setter
     def max_results(self, value):
-        self._set_sub_prop("maxResults", str(value))
+        self._set_sub_prop("maxResults", value)
 
     @property
     def maximum_billing_tier(self):
@@ -2579,6 +2579,18 @@ class QueryJob(_AsyncJob):
 
         return resource
 
+    def _set_properties(self, api_response):
+        """Update properties from resource in body of ``api_response``
+
+        :type api_response: dict
+        :param api_response: response returned from an API call
+        """
+
+        # prevent max_results value from being erased
+        max_results = self.max_results
+        super(QueryJob, self)._set_properties(api_response)
+        self._configuration.max_results = max_results
+
     def _copy_configuration_properties(self, configuration):
         """Helper:  assign subclass configuration properties in cleaned."""
         self._configuration._properties = copy.deepcopy(configuration)
@@ -2858,7 +2870,6 @@ class QueryJob(_AsyncJob):
                 project=self.project,
                 timeout_ms=timeout_ms,
                 location=self.location,
-                max_results=self.max_results,
             )
 
             # Only reload the job once we know the query is complete.
@@ -2962,11 +2973,7 @@ class QueryJob(_AsyncJob):
             # Return an iterator instead of returning the job.
             if not self._query_results:
                 self._query_results = self._client._get_query_results(
-                    self.job_id,
-                    retry,
-                    project=self.project,
-                    location=self.location,
-                    max_results=self.max_results,
+                    self.job_id, retry, project=self.project, location=self.location
                 )
         except exceptions.GoogleCloudError as exc:
             exc.message += self._format_for_exception(self.query, self.job_id)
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -2558,6 +2579,18 @@ def to_api_repr(self):
 
         return resource
 
+    def _set_properties(self, api_response):
+        """Update properties from resource in body of ``api_response``
+
+        :type api_response: dict
+        :param api_response: response returned from an API call
```
plamut: One quick remark while in draft mode - docstrings should use the Google style format, thus let's use it here, too.

I did read that someone is preparing a script to do that automatically, but let's not make this PR dependent on it, or require multiple passes of the script (if e.g. this PR is not merged in time fro the first pass).

shubha-rajan: Thanks for letting me know. I can manually reformat this (and the method in the parent class) before making this PR ready for review.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -2582,8 +2582,8 @@ def to_api_repr(self):
     def _set_properties(self, api_response):
         """Update properties from resource in body of ``api_response``
 
-        :type api_response: dict
-        :param api_response: response returned from an API call
+        Args:
+            api_response (dict): response returned from an API call.
```
plamut: By the way, the types are specified with names from the `typing` module, thus a `dict` should be named `Dict`, for example. Or `Dict[key_type, value_type]` if you also want to specify the dict content's type(s).

shubha-rajan: got it. The `dict` in question would be a nested API response so it would be okay to just name it `Dict` without specifying the content, right?

shubha-rajan:
failing snippets tests also fail locally on master, so they're probably unrelated to changes in this PR

bigquery/google/cloud/bigquery/job.py
IlyaFaer: @plamut, @shubha-rajan, yes, yesterday I've got approval from Tim Swast - we're gonna proceed with the script. Not only BigQuery's docstrings should be reformatted, so it is a huge work - thousands of comments to update, and it's better be done automatically

plamut:
@shubha-rajan Indeed, that started occurring a day or two ago. The backend team has been informed about it, we are awaiting the ETA for the fix. If it's too long, we can temporarily disable the failing test as a workaround.

bigquery/google/cloud/bigquery/job.py
plamut: I suppose so, yes. If all keys are strings, `Dict[str, Any]` could be used, but just `Dict` with a meaningful description is fine, too.

tswast:
Since these are two different features, let's have them as two (possibly 3) separate PRs, starting with `--max_results` feature. That way they are more clearly identified as new features in the CHANGELOG when we release these features.

I'd prefer we find a different implementation for `max_results`. Note: `list_rows` [accepts a `max_results` argument](https://github.com/googleapis/google-cloud-python/blob/8a1417dd9d3f3cfcae7dbbd2a2348cd85b27d4de/bigquery/google/cloud/bigquery/client.py#L2202), and `QueryJob.result` [calls the `list_rows` method](https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L2959). I think it would be appropriate to add a `max_results` argument to `QueryJob.result`.

Let's have 3 PRs in this order:

1. Add `max_results=None` argument to the `QueryJob.result` method.
2. Add a `--max_results` argument to the `%%bigquery` magic.
3. Add ability to pass in a table ID instead of a query to the `%%bigquery` magic.

shubha-rajan:
@tswast sounds good. I'll close this one and submit 3 separate PRs

PR #9151, opened 2019-08-30
feat(spanner): add client_options to constructor

busunkim96:
Toward #8475

### DIFF
--- a/spanner/tests/unit/test_database.py
+++ b/spanner/tests/unit/test_database.py
@@ -295,6 +295,7 @@ class TestDatabase(_BaseTest):
         called_args, called_kw = spanner_client.call_args
         self.assertEqual(called_args, ())
         self.assertEqual(called_kw["client_info"], client_info)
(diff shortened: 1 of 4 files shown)


larkee:
> LGTM. @busunkim96 can you confirm that the `api_endpoint` will be the same betweeen the three clients?

I can confirm that `api_endpoint` will be the same between the three clients. They all use the Client object's new `_client_options` attribute.

spanner/tests/unit/test_database.py
```diff
@@ -271,6 +274,7 @@ def with_scopes(self, scopes):
         expected_scopes = (SPANNER_DATA_SCOPE,)
         client = _Client()
         client_info = client._client_info = mock.Mock()
+        client_options = client._client_options = mock.Mock()
```
larkee: Currently the `client_options` variable isn't being used and is causing the lint check to fail. Did you mean to use it in an assert similar to the previous test?

PR #9156, opened 2019-09-02
BigQuery: Disable failing snippets test

plamut:
`test_copy_table_cmek()` snippets test results in internal error on the backend and fails, blocking the PRs, and this PR marks the test as skipped.

Currently awaiting more information from the backend team. If the issue is not fixed soon, we will temporarily disable the test, but until then please do not merge this PR just yet.


tswast:
For Googlers investigating this issue, it's internal bug 140294699 with the reported CMEK failures.

PR #9159, opened 2019-09-03
Vision: Flatten case of logo text in systest.

brendanlundy:
…. Fixes broken Vision tests since returned logo result is "Google".


tseaver:
@brendanlundy Thanks for the patch!

Issue #7827, opened 2019-04-30
BigQuery: Table.from_string() doesn't support GCP project ID with org prefix


salvador-bynd:
#### Environment details

1. BigQuery
2. Google App-Engine standard (Python 2.7)
3. Python 2.7
4. google-cloud-bigquery==1.10.0

#### Code example

```python

from google.cloud import bigquery
from google.cloud.bigquery import Table

# Standard SQL (works fine)
# SELECT * FROM `google.com:my-project.mydataset.mytable` LIMIT 1

# Works fine
table = Table.from_string('my-project.mydataset.mytable')

# Fails
table = Table.from_string('google.com:my-project.mydataset.mytable')

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/vv/local/lib/python2.7/site-packages/google/cloud/bigquery/table.py", line 815, in from_string
    return cls(TableReference.from_string(full_table_id))
  File "/tmp/vv/local/lib/python2.7/site-packages/google/cloud/bigquery/table.py", line 246, in from_string
    "got {}".format(table_id)
ValueError: Too many parts in table_id. Must be a fully-qualified table ID in standard SQL format. e.g. "project.dataset.table", got google.com:my-project.mydataset.mytable
```

joncass:
I am having this same issue. It looks to me like the fix would be to change line https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/_helpers.py#L590 from `parts = full_id.split(".")` to `parts = full_id.rsplit(".", 2)`. Any objections to that change?

Emar-Kar:
@tswast yeah, sure. I’ll take a look tomorrow.

plamut:
@Emar-Kar Are you currently working on this? Asking to not unnecessarily duplicate the effort.

Emar-Kar:
@plamut Yeah, thank you for asking. I’m working on it.

plamut:
@Emar-Kar Great, thanks. I updated the assignee for better visibility.

PR #9161, opened 2019-09-03
BigQuery: Add support of the project ID with org prefix to the Table.from_string() method

Emar-Kar:
ID splitting was moved to a separate method to avoid code duplication. `Dataset` file changed to the new implementation.

Closes: #7827


bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -29,6 +30,12 @@
 _RFC3339_MICROS_NO_ZULU = "%Y-%m-%dT%H:%M:%S.%f"
 _TIMEONLY_WO_MICROS = "%H:%M:%S"
 _TIMEONLY_W_MICROS = "%H:%M:%S.%f"
+_PROJECT_PREFIX_PATTERN = re.compile(
+    r"""
+    (?P<project_id>\S+\:[^.]+)\.(?P<dataset_id>[^.]+)(?:$|\.(?P<custom_id>[^.]+)$)
+""",
```
plamut: Are multiple colons in the project name allowed, or should the pattern be updated?

```py
>>> m = _PROJECT_PREFIX_PATTERN.match("example.com::::project_id.dataset_id.table_id")
>>> m is not None
True
>>> m.group("project_id")
'example.com::::project_id'
```

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -586,24 +593,34 @@ def _str_or_none(value):
         return str(value)
 
 
+def _split_id(full_id):
```
plamut: The docstring is missing here, let's add it to this helper, too.

Emar-Kar: Will it be understandable if I use `Helper: split full_id into composite parts` as the docstring? Or should I write more specific?

bigquery/google/cloud/bigquery/_helpers.py
plamut: That's a valid point, and it's the same thing with other parts, too (e.g. the max length of dataset ID, hyphens int it, etc.).

In an ideal world we would borrow a regex from the backend, and should indeed not try to guess all possible cases ourselves.

bigquery/google/cloud/bigquery/_helpers.py
plamut: One liner is fine IMO. Personally, I would also document the argument types and the return value on top of it.

bigquery/google/cloud/bigquery/_helpers.py
plamut: @tswast Currently we do not check for things that the backend would reject, e.g. the max allowed length of `dataset_id`, dashes instead of underscores in the names, etc.

I know that we prefer backend errors, but would It still be useful to warn the users about these errors early?
(in a separate PR)

tswast: > In an ideal world we would borrow a regex from the backend, and should indeed not try to guess all possible cases ourselves.

That's too much to keep in sync. Even though identifier restrictions have been very stable, it can still get out of date. For example, what if we start accepting any unicode for dataset / table identifiers to support other languages?

As much as possible, I'd like to keep the client simple, since in all likelihood we'll forget all the places where we do client-side validation

plamut: OK thanks, resolving this thread then.

bigquery/google/cloud/bigquery/_helpers.py
Emar-Kar: @plamut I decided to follow your advice and practice the docstrings writing. Could you have an eye here and check it, please?

code_sample_2

plamut: @Emar-Kar The docstring looks good, thanks!

PR #9162, opened 2019-09-03
BigQuery: Add client.insert_rows_from_dataframe() method

### LINKED ISSUE #7553: BigQuery: Add `insert_rows_from_dataframe` to use the streaming API to append a pandas DataFrame to an existing table

There is a limit on the number of load jobs per day (per table and per project limit of 1,000 load jobs). When a pandas DataFrame is used as part of a data processing pipeline, this limit can easily be reached. See: https://github.com/pydata/pandas-gbq/issues/238

I'd like to see a `insert_rows_from_dataframe` method on the client to complement the `insert_rows` method. It would transform the (column-oriented) DataFrame in to (row-oriented) JSON and upload the rows via calls to tabledata.insert (in batches).

plamut:
Closes #7553.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -15,6 +15,7 @@
 """Client for interacting with the Google BigQuery API."""
 
 from __future__ import absolute_import
+from __future__ import division
 
 try:
     from collections import abc as collections_abc
@@ -25,7 +26,9 @@ import copy
 import functools
 import gzip
 import io
+import itertools
 import json
+import math
 import os
 import tempfile
 import uuid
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -2080,6 +2083,54 @@ def insert_rows(self, table, rows, selected_fields=None, **kwargs):
...
+                contains a list of the mappings describing one or more problems
+                with the row.
+
+        Raises:
+            ValueError: if table's schema is not set
+        """
+        insert_results = []
+
+        chunk_count = int(math.ceil(len(dataframe) / chunk_size))
+        rows_iter = (row._asdict() for row in dataframe.itertuples(index=False))
```
tswast: Is `_asdict()` because `itertuples()` returns `NamedTuples`? Do we know what versions of Pandas support this? Do we need to update the optional dependency in `setup.py`?

plamut: The `itertuples()` method was added in Pandas `0.7.1` ([release notes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.7.1.html?highlight=itertuples)),  and support for namedtuples was [added in v0.17.1](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.17.1.html?highlight=itertuples#enhancements). We have the `pandas` version pinned to `>=0.17.1` in `setup.py`.

With that said, looking at the source reveals an [unpleasant trap](https://github.com/pandas-dev/pandas/blob/v0.25.1/pandas/core/frame.py#L969-L972) - if the number of columns is larger than 255, a _regular_ tuple is returned, and there is no `_asdict()` method on it.

I will rewrite this part to construct a dict from the tuple manually.

plamut: Well go figure, `itertuples()` has a bug has a bug in Pandas `0.25.1` (but not reproducible in `0.24.2`):
```py
>>> import pandas
>>> pandas.__version__
'0.25.1'
>>> df = pandas.DataFrame([{"foo_{}".format(i): "bar_{}".format(i) for i in range(255)}])
>>> df.itertuples(index=False)
...
SyntaxError: more than 255 arguments
```
They introduced it [here](https://github.com/pandas-dev/pandas/commit/81f5c0164e6cb26cc8fea00ebc22f4fe93771ff6#diff-1e79abbbdd150d4771b91ea60a4e1cc7R903) by removing the try-except block around the namedtuple creation.

Will enforce plain tuples by passing `name=None` to `itertuples()`, and report the issue to Pandas maintainers.

Issue #7503, opened 2019-03-12
Missing documentation for 'enum' modules


mik-laj:
Hello,

I develop [Airflow](https://airflow.readthedocs.io/en/latest/cli.html) operators for the GCP platform. For this reason, I use your library very intensively. A large part of my work is to ensure that the documentation is of high quality. One of these elements is to ensure linking between my documentation and your documentation. I use the [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) library for this purpose.

Unfortunately, I noticed the deficiencies in your documentation. Not all documentation is generated correctly. I attach a fragment of the documentation, which in my opinion is incorrect.

<img width="1025" alt="Screenshot 2019-03-12 at 13 56 52" src="https://user-images.githubusercontent.com/12058428/54202713-ee18cd80-44d0-11e9-8434-8ef6fac4002c.png">

This is important because it makes very difficult to use some methods. Parameters that are needed are not known. I am forced to look at the source code  to read the documentation.

For example:
https://googleapis.github.io/google-cloud-python/latest/videointelligence/gapic/v1/api.html#google.cloud.videointelligence_v1.VideoIntelligenceServiceClient.annotate_video
Look at `features` parameter
<img width="806" alt="Screenshot 2019-03-12 at 14 19 20" src="https://user-images.githubusercontent.com/12058428/54203173-f7566a00-44d1-11e9-8e4a-3a6a2cd84306.png">
Description is only available at: https://github.com/googleapis/google-cloud-python/blob/master/videointelligence/google/cloud/videointelligence_v1/gapic/enums.py#L21-L43
Confirmation: https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fgoogleapis.github.io++inurl%3Agoogle-cloud-python++%22Video+annotation+feature.%22&oq=site%3Ahttps%3A%2F%2Fgoogleapis.github.io++inurl%3Agoogle-cloud-python++%22Video+annotation+feature.%22

The problem concerns the following addresses:
````
https://googleapis.github.io/google-cloud-python/latest/asset/gapic/v1beta1/api.html#google.cloud.asset_v1beta1.AssetServiceClient.enums
https://googleapis.github.io/google-cloud-python/latest/automl/gapic/v1beta1/api.html#google.cloud.automl_v1beta1.AutoMlClient.enums
[...]

PR #9165, opened 2019-09-03
Tasks: Fix reference to View

TobKed:
Hi,

The reference to class seems wrong.
PR introduces the correct reference.

Connected issue:
https://github.com/googleapis/google-cloud-python/issues/7503
Connected PR:
https://github.com/googleapis/google-cloud-python/pull/8779

Kind regards

### DIFF
--- a/tasks/google/cloud/tasks_v2/gapic/cloud_tasks_client.py
+++ b/tasks/google/cloud/tasks_v2/gapic/cloud_tasks_client.py
@@ -1223,7 +1223,7 @@ class CloudTasksClient(object):
         Args:
             parent (str): Required. The queue name. For example:
                 ``projects/PROJECT_ID/locations/LOCATION_ID/queues/QUEUE_ID``
-            response_view (~google.cloud.tasks_v2.types.View): The response\_view specifies which subset of the ``Task`` will be
+            response_view (~google.cloud.tasks_v2.enums.Task.View): The response\_view specifies which subset of the ``Task`` will be
                 returned.
 
                 By default response\_view is ``BASIC``; not all information is retrieved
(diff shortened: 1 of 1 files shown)


TobKed:
@googlebot I signed it!

TobKed:
Connected issue:
https://github.com/googleapis/google-cloud-python/issues/7503
Connected PR:
https://github.com/googleapis/google-cloud-python/pull/8779

tseaver:
@TobKed Thank you for the patch!

mik-laj:
@tseaver Do you know that this change will be rolled back in a moment? In a moment, synth will generate a patch that will delete all changes.

tseaver:
@mik-laj Oops, thanks for point that out!  I missed seeing that the change was in generated code.

@TobKed I'm afraid we need to either tweak the `synth.py` file to apply your fix, get an upstream change applied to the https://github.com/googleapis/gapic-generator/ repository, or else just let your change be reverted in the next autogen run.

mik-laj:
It seems to me that this problem needs global fix. In many places, the enum type is not correctly specified.
For example: https://github.com/googleapis/google-cloud-python/pull/8779 or this PR.
Should we not fix in all places than deal with individual problems? This is very inefficient and causes many problems among developers. I am slowly ceasing to trust the generated docstrings. 
The significance of this problem increases when we realize that documentation for enum is not generated.
https://github.com/googleapis/google-cloud-python/issues/7503
Many developers do not understand the difference between types and enums. Errors in documentation, and the lack of a large part of the documentation does not facilitate this.  This is an important part of any library and I think that this problem should be solved quickly because it makes it difficult to use the library.  I know that I can look at other sources of information about the way objects to the library, but this is not obvious knowledge.

What can I do to ensure that libraries have the correct documentation for enums?

PR #9169, opened 2019-09-04
BigQuery: add `--max_results` option to magic

shubha-rajan:
Second of 3 PRs towards resolving #9105 as described in review for #9147

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -320,7 +320,6 @@ def _run_query(client, query, job_config=None, max_results=None):
     default=None,
     help=("Project to use for executing this query. Defaults to the context project."),
 )
-
 @magic_arguments.argument(
     "--max_results",
     default=None,
@@ -329,7 +328,6 @@ def _run_query(client, query, job_config=None, max_results=None):
         "Defaults to returning all rows."
     ),
 )
-
 @magic_arguments.argument(
     "--max_results",
     default=None,
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -320,6 +320,16 @@ def _run_query(client, query, job_config=None, max_results=None):
     default=None,
     help=("Project to use for executing this query. Defaults to the context project."),
 )
+
+@magic_arguments.argument(
+    "--max_results",
+    default=None,
+    help=(
+        "Maximum number of rows in dataframe returned from executing the query."
+        "Defaults to returning all rows."
+    ),
+)
+
 @magic_arguments.argument(
     "--max_results",
     default=None,
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -266,7 +266,7 @@ class Context(object):
 context = Context()
 
 
-def _run_query(client, query, job_config=None, max_results=None):
+def _run_query(client, query, job_config=None):
     """Runs a query while printing status updates
 
     Args:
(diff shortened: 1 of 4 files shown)


tswast:
I just merged https://github.com/googleapis/google-cloud-python/pull/9167. We might have to rebase.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -300,7 +300,7 @@ def _run_query(client, query, job_config=None):
     while True:
         print("\rQuery executing: {:0.2f}s".format(time.time() - start_time), end="")
         try:
-            query_job.result(timeout=0.5)
+            query_job.result(timeout=0.5, max_results=max_results)
```
tswast: Since we aren't actually returning the results in this line (just waiting for the query to finish), we don't need to pass `max_results` here. I believe that means we can remove the `max_results` parameter from `_run_query` as well.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -320,6 +320,22 @@ def _run_query(client, query, job_config=None):
...
 )
+@magic_arguments.argument(
+    "--max_results",
+    default=None,
+    help=(
+        "Maximum number of rows in dataframe returned from executing the query."
+        "Defaults to returning all rows."
+    ),
+)
+@magic_arguments.argument(
```
tswast: Looks like these arguments got doubled up. (Maybe the two commits thing, we noticed when rebasing?)

bigquery/tests/unit/test_magics.py
```diff
@@ -414,7 +414,7 @@ def test_bigquery_magic_with_legacy_sql():
     with run_query_patch as run_query_mock:
         ip.run_cell_magic("bigquery", "--use_legacy_sql", "SELECT 17 AS num")
 
-        job_config_used = run_query_mock.call_args_list[0][0][-1]
+        job_config_used = run_query_mock.call_args_list[0][1]["job_config"]
```
tswast: Was this change intentional? Maybe a bad rebase?

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -433,7 +455,9 @@ def _cell_magic(line, query):
 
     error = None
     try:
-        query_job = _run_query(client, query, job_config)
+        query_job = _run_query(
+            client, query, job_config=job_config, max_results=max_results
```
tswast: `max_results` is not actually needed here, it's not until the call to `to_dataframe` that we need `max_results`.

Note: `to_dataframe` probably doesn't have a `max_results` argument, and I'm actually not certain that we'd want to add one. Instead, we can call `query_job.results` with a `max_results` argument and then call `to_dataframe` on the resulting `RowIterator`.

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: That's probably what happened (moral of the story, always run tests before pushing). Removed the duplicate in 147e43d

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: Removed in f88ffe8

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: I updated to call `to_dataframe` on `query_job.result`, passing `max_results` as an argument if `max_results` is present in f88ffe8

bigquery/tests/unit/test_magics.py
shubha-rajan: This was intentional. It was to fix some test failures I was getting because the argument that was being retrieved was the wrong type (a result of changing the method signature for `_run_query`. This way the arg being accessed will always be `job_config`, even if other named parameters are added. Since I reverted the changes to _run_query, I can probably change this back too. I think the tests will pass either way

shubha-rajan:
Screenshot of running a query in a notebook with `max_results` set:
image

PR #9170, opened 2019-09-04
BigQuery: Add ability to pass in a table ID instead of a query to the %%bigquery magic.

shubha-rajan:
Third of 3 PRs towards resolving #9105 as described in review for #9147
Screenshot of feature in notebook:
image

### DIFF
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -445,8 +445,8 @@ def _cell_magic(line, query):
         max_results = None
 
     query = query.strip()
-    
-    # Any query that does not contain whitespace (aside from leading and trailing whitespace) 
+
+    # Any query that does not contain whitespace (aside from leading and trailing whitespace)
     # is assumed to be a table id
     if not re.search(r"\s", query):
         try:
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -476,7 +476,8 @@ def _cell_magic(line, query):
     try:
         query_job = _run_query(client, query, job_config=job_config)
     except Exception as ex:
-        return _print_error(str(ex), args.destination_var)
+        _print_error(str(ex), args.destination_var)
+        return 
 
     if not args.verbose:
         display.clear_output()
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -477,7 +477,7 @@ def _cell_magic(line, query):
         query_job = _run_query(client, query, job_config=job_config)
     except Exception as ex:
         _print_error(str(ex), args.destination_var)
-        return 
+        return
 
     if not args.verbose:
         display.clear_output()
--- a/bigquery/google/cloud/bigquery/magics.py
+++ b/bigquery/google/cloud/bigquery/magics.py
@@ -444,18 +444,13 @@ def _cell_magic(line, query):
     else:
         max_results = None
 
-    error = None
-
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/magics.py
```diff
@@ -434,6 +444,26 @@ def _cell_magic(line, query):
...
+    error = None
+
+    if not re.search(r"\s", query.rstrip()):
+        table_id = query.rstrip()
+
+        try:
+            rows = client.list_rows(table_id, max_results=max_results)
+        except Exception as ex:
+            error = str(ex)
+        if error:
```
crwilcox: could this be simplified to return from the exception?

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -434,6 +444,26 @@ def _cell_magic(line, query):
     else:
         max_results = None
 
+    error = None
+
+    if not re.search(r"\s", query.rstrip()):
```
crwilcox: What scenario is this protecting against? Is there a scenario where unicode strings will not remove space characters via rstrip?

What do we expect in the else case?

crwilcox: Related: Does this repository enforce code coverage? Is there a case where we test there being unstoppable whitespace and not taking this branch?

shubha-rajan: This condition is actually testing for queries that contain whitespace characters that aren't removed by rstrip (so any whitespace that isn't a trailing newline). The assumption being made (as described in #9105 ) is that anything containing whitespace is a SQL query and won't take this branch, while anything string without whitespace is a table_id and will take this branch. 

So anything regular SQL query would fall into the else case. There are tests that check whether query strings without spaces will be interpreted as table IDs and a test that checks if a string without whitespace that isn't a valid table_id raises an appropriate error message.

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: it definitely can.  fixed in 7a13ca4

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: I can add a comment in the code if that would be helpful.

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -434,6 +444,21 @@ def _cell_magic(line, query):
     else:
         max_results = None
 
+    if not re.search(r"\s", query.rstrip()):
+        table_id = query.rstrip()
```
plamut: BTW,  stripping whitespace from the same value again is not necessary, we could store the stripped string into a variable the first time we do it.

shubha-rajan: fixed in d663104

bigquery/google/cloud/bigquery/magics.py
```diff
@@ -445,24 +470,14 @@ def _cell_magic(line, query):
         value = int(args.maximum_bytes_billed)
         job_config.maximum_bytes_billed = value
 
-    error = None
     try:
         query_job = _run_query(client, query, job_config=job_config)
     except Exception as ex:
-        error = str(ex)
+        return _print_error(str(ex), args.destination_var)
```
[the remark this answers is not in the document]
shubha-rajan: fixed in d663104

bigquery/google/cloud/bigquery/magics.py
shubha-rajan: added comment in d663104

PR #9189, opened 2019-09-09
BigQuery: Add support for array parameters to Cursor.execute()

### LINKED ISSUE #9177: BigQuery: Support WHERE IN or WHERE ANY clause in DB-API
BigQuery DB-API throws the following error when I pass parameters to the Cursor.execute() for WHERE IN or WHERE ANY clause. Any idea if it will be supported?

```
  File "/home/haibin/.local/share/virtualenvs/python-6nCS1ipk/lib/python3.6/site-packages/google/cloud/bigquery/dbapi/cursor.py", line 159, in execute
    query_parameters = _helpers.to_query_parameters(parameters)
  File "/home/haibin/.local/share/virtualenvs/python-6nCS1ipk/lib/python3.6/site-packages/google/cloud/bigquery/dbapi/_helpers.py", line 117, in to_query_parameters
    return to_query_parameters_list(parameters)
  File "/home/haibin/.local/share/virtualenvs/python-6nCS1ipk/lib/python3.6/site-packages/google/cloud/bigquery/dbapi/_helpers.py", line 84, in to_query_parameters_list
    return [scalar_to_query_parameter(value) for value in parameters]
  File "/home/haibin/.local/share/virtualenvs/python-6nCS1ipk/lib/python3.6/site-packages/google/cloud/bigquery/dbapi/_helpers.py", line 84, in <listcomp>
    return [scalar_to_query_parameter(value) for value in parameters]
  File "/home/haibin/.local/share/virtualenvs/python-6nCS1ipk/lib/python3.6/site-packages/google/cloud/bigquery/dbapi/_helpers.py", line 69, in scalar_to_query_parameter
    name, value
google.cloud.bigquery.dbapi.exceptions.ProgrammingError: encountered parameter None with value ['Apple.com', 'Coles'] of unexpected type

```

plamut:
Closes #9177.

This PR adds support for array-like query parameters.

### How to test
Run the example from the issue description (slightly modified, see below). Binding an array-like parameter to a parametrized query should not raise an error, and the following should work as expected:
```py
curr.execute(query, (('NY', 'TX'), ))
```

Mind that scanning arrays requires using `IN UNNEST(...)` construct ([docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#scanning-arrays)), thus the `UNNEST` function needs to be added to the query from the issue description:
```py
query = """
  SELECT name, state
  FROM `bigquery-public-data.usa_names.usa_1910_2013`
  WHERE state IN UNNEST(%s)
  LIMIT 2
"""
```

### Footnotes
- No nested arrays support, only for arrays of scalar types.
- Detecting if a value is "array-like" is approximate, it primarily tries to cover the most common cases.

### DIFF
--- a/bigquery/google/cloud/bigquery/dbapi/_helpers.py
+++ b/bigquery/google/cloud/bigquery/dbapi/_helpers.py
@@ -106,7 +106,9 @@ def to_query_parameters_list(parameters):
     result = []
 
     for value in parameters:
-        if array_like(value):
+        if isinstance(value, collections_abc.Mapping):
+            raise NotImplementedError("STRUCT-like parameter values are not supported.")
+        elif array_like(value):
             param = array_to_query_parameter(value)
         else:
             param = scalar_to_query_parameter(value)
@@ -127,7 +129,12 @@ def to_query_parameters_dict(parameters):
     result = []
 
     for name, value in six.iteritems(parameters):
-        if array_like(value):
+        if isinstance(value, collections_abc.Mapping):
+            raise NotImplementedError(
+                "STRUCT-like parameter values are not supported "
+                "(parameter {}).".format(name)
+            )
+        elif array_like(value):
             param = array_to_query_parameter(value, name=name)
         else:
             param = scalar_to_query_parameter(value, name=name)
--- a/bigquery/google/cloud/bigquery/dbapi/_helpers.py
+++ b/bigquery/google/cloud/bigquery/dbapi/_helpers.py
@@ -43,33 +43,55 @@ def scalar_to_query_parameter(value, name=None):
     :raises: :class:`~google.cloud.bigquery.dbapi.exceptions.ProgrammingError`
         if the type cannot be determined.
     """
-    parameter_type = None
+    parameter_type = bigquery_scalar_type(value)
 
-    if isinstance(value, bool):
-        parameter_type = "BOOL"
-    elif isinstance(value, numbers.Integral):
-        parameter_type = "INT64"
-    elif isinstance(value, numbers.Real):
-        parameter_type = "FLOAT64"
-    elif isinstance(value, decimal.Decimal):
-        parameter_type = "NUMERIC"
-    elif isinstance(value, six.text_type):
-        parameter_type = "STRING"
-    elif isinstance(value, six.binary_type):
-        parameter_type = "BYTES"
-    elif isinstance(value, datetime.datetime):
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -72,6 +54,46 @@ def scalar_to_query_parameter(value, name=None):
...
+    """
+    if not array_like(value):
+        raise exceptions.ProgrammingError(
+            "The value of parameter {} must be a sequence that is "
+            "not string-like.".format(name)
+        )
+
+    if not value:
+        raise exceptions.ProgrammingError(
+            "Encountered an empty array-like value of parameter {}, cannot "
```
tswast: This concerns me, since someone is probably going to hit this pretty quickly, but I don't like the alternatives (such as defaulting to STRING parameters or something).

bigquery/google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -115,3 +152,55 @@ def to_query_parameters(parameters):
...
+    sequences such as ``list`` and ``tuple``, but not strings and other
+    iterables such as sets.
+
+    Args:
+        value (Any)
+
+    Returns:
+        bool: ``True`` if the value is considered array-like, ``False`` otherwise.
+    """
+    return isinstance(value, collections_abc.Sequence) and not isinstance(
```
tswast: Does a dictionary count as a sequence? We probably want to exclude those so that we can support STRUCT parameters in the future.

(Maybe even add a `struct_like` function that raises that it's not implemented yet.)

bigquery/google/cloud/bigquery/dbapi/_helpers.py
plamut: I hope the error message is clear enough?
I don't like an implicit fallback to strings either, but also can't think of a better way of handling this right now (in the helper itself at least).

bigquery/google/cloud/bigquery/dbapi/_helpers.py
plamut: Dictionaries are excluded.
```py
>>> isinstance({}, collections.Sequence)                                                                                                                                                                                                                                                                                                                                                                                     
False
```
I can add a blind branch (raising `NotImplementedError`) for struct-like values. Probably testing the instance against `collections.Mapping` is sufficient for most cases.

PR #9199, opened 2019-09-10
BigQuery: Add performance note to fetchall() docs

### LINKED ISSUE #9185: BigQuery: DB-API is very slow
DB-API is very slow.

google-cloud-bigquery version: 1.19.0

```python
from datetime import datetime

from google.cloud.bigquery import dbapi

client = bigquery.Client()
conn = dbapi.Connection(client)
curr = conn.cursor()

start = datetime.now()
QUERY = """SELECT name FROM `bigquery-public-data.usa_names.usa_1910_2013` LIMIT 100"""
query_job = client.query(QUERY)
query_job.result()
print('API', datetime.now() - start)

start = datetime.now()
curr.execute(QUERY)
result = curr.fetchall()
print('DB-API', datetime.now() - start)
```
Output
```
API 0:00:01.623182
DB-API 0:01:36.157141
```

plamut:
Closes #9185.

This PR adds a note on `arraysize` parameter, and its impact on the `fetchall()` method performance.

### How to test
Run the code sample from the issue description, and verify that by setting `cursor.arraysize` to appropriate value avoids the reported performance issue, which what the updated docs point out.

### DIFF
--- a/bigquery/google/cloud/bigquery/dbapi/cursor.py
+++ b/bigquery/google/cloud/bigquery/dbapi/cursor.py
@@ -264,6 +264,10 @@ class Cursor(object):
     def fetchall(self):
         """Fetch all remaining results from the last ``execute*()`` call.
 
+        .. note::
+            The ``arraysize`` attribute can affect the performance of this
+            operation. Make sure to set it to appropriate batch size beforehand.
+
         :rtype: List[tuple]
         :returns: A list of all the rows in the results.
         :raises: :class:`~google.cloud.bigquery.dbapi.InterfaceError`


tswast:
completely forgot about the `arraysize` parameter!

Rather than add this note, I think it may make more sense to diverge from the DB-API spec (which dictates a default `arraysize=1`).

https://github.com/googleapis/google-cloud-python/blob/cf44700656acb69b58897bbd3b1bc2faaece903d/bigquery/google/cloud/bigquery/dbapi/cursor.py#L63-L65

If we instead default to `None`, the BigQuery API backend can choose a more appropriate page size automatically.

plamut:
> If we instead default to None, the BigQuery API backend can choose a more appropriate page size automatically.

I see, that actually makes sense. It's safer than just having it in the docs, as some portion of the users will probably still overlook it.

PR #9202, opened 2019-09-10
Speech: Add speaker diarization configuration support (via synth).

tseaver:
Also, exclude 'noxfile.py' from synth (to preserve 'samples').

Supersedes #9196.


tseaver:
@busunkim96 This PR and #9201 assume that just excluding `noxfile.py` from the synth copy process is correct for `language` and `speech` (so that the `samples` stanza remains).  Would re rather add a `samples` section to the generator?

The current implementation:
- skips the stanza if the `samples` directory is missing;
- tests the samples using `sample-tester`

busunkim96:
@tseaver I should probably add the samples session to the noxfile template eventually. This is fine for now.

PR #9205, opened 2019-09-10
Storage: Avoid using 'REGIONAL'/'MULTI_REGIONAL' in examples, tests.

### LINKED ISSUE #9191: Update MultiRegional and Regional to Standard
Previously, we implemented the concept of "Location Type" to move the concept of bucket location away from storage class. It is no longer considered correct to use "MultiRegional" or "Regional" as a storage class, instead those should be location types.

As such, we need to update every instance of the "MultiRegional" and "Regional" storage in samples, library documentation, and tests to use "Standard" instead--Standard is a correct storage class.

Note that "MultiRegional" and "Regional" should only be replaced with "Standard" in cases where they are used a storage class. If they're used as a location type, that's fine and doesn't need to be replaced. 

https://github.com/googleapis/google-cloud-python/blob/ae2e7058e402c4c963bc4991af7cd5f2b2801cc9/storage/tests/unit/test_bucket.py
https://github.com/googleapis/google-cloud-python/blob/ccac1e1a95bbdc85c001c96ad359dcec50f3e407/storage/tests/unit/test_blob.py#L3040
https://github.com/googleapis/google-cloud-python/blob/ae2e7058e402c4c963bc4991af7cd5f2b2801cc9/storage/tests/unit/test_bucket.py#L55
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/03e47889042b52ebb2610c5ba19a23bebb39a497/notebooks/rendered/storage-command-line-tool.md#L73
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/03e47889042b52ebb2610c5ba19a23bebb39a497/notebooks/tutorials/storage/Storage%20command-line%20tool.ipynb#L141

tseaver:
Exceptions are for tests of setters, which exercise allowed values.

Closes #9191.

### DIFF
--- a/storage/docs/constants.rst
+++ b/storage/docs/constants.rst
@@ -1,0 +1,7 @@
+Constants
+~~~~~~~~~
+
+.. automodule:: google.cloud.storage.constants
+  :members:
+  :member-order: bysource
+
--- a/storage/docs/index.rst
+++ b/storage/docs/index.rst
@@ -18,6 +18,7 @@ API Reference
   buckets
   acl
   batch
+  constants
 
 Changelog
 ---------
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -46,12 +46,12 @@ from google.resumable_media.requests import Download
 from google.resumable_media.requests import MultipartUpload
 from google.resumable_media.requests import ResumableUpload
 
+from google.api_core.iam import Policy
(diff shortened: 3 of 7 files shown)


frankyn:
Overall LGTM, please wait for @JesseLovelace's review.

tseaver:
@JesseLovelace PTAL.

PR #9221, opened 2019-09-12
Firestore: Mark tests as incompatible with emulator

crwilcox:
There are a few tests that don't currently work under the emulator. This marks them and references tracking issues to monitor to enable these.


tseaver:
Superseded by #9222.

Issue #8722, opened 2019-07-19
Run system test suite on emulator as well as live service


crwilcox:
Some of our libraries have emulator support and it would be a nice to have if the emulator was used to run our system test suites.

I think starting with bigtable or firestore would make sense, then we could move to others if desired.

tseaver:
Datastore is the "oldest" one in the repository:

https://github.com/googleapis/google-cloud-python/blob/951b9c68d2390eedd271b7eb1ca363cb48d82d8d/datastore/google/cloud/datastore/client.py#L245-L249

PR #9222, opened 2019-09-12
Firestore: Enabling system tests to run on emulator

sumit-ql:
Towards #8722 

- Updated sys tests to make them run on emulator. 

- Marked some tests as skipped as those functionalities still not supported on emulator.


sumit-ql:
@tseaver Could you please review this PR.

PR #9225, opened 2019-09-13
feat(bigquery): add support for listing jobs by parent job

### LINKED ISSUE #9190: BigQuery: add `parent_job` parameter to `list_jobs`

With the scripting feature (alpha), a single query job can initiate several sub-jobs. These job resources are not embedded in the initial job. Instead, you must populate the `parentJobId` query parameter to the [BigQuery jobs list API](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/list).

I propose a new `parent_job` parameter to `Client.list_jobs`.

- Accept a string (interpreted as a job ID).
- Accept a `*Job` object. Use the job ID from the job object. Also, use the location from the job object.

We could make a `list_children()` method on the job classes, but we'd still need to update `list_jobs` to support it. Let's start just updating `list_jobs` for now.

 **Additional context**

Example from the alpha docs.

BigQuery’s scripting feature enables you to send multiple statements to BigQuery in one request, to use variables, and to use control flow statements such as IF and WHILE. For example, you can declare a variable, assign a value to it, and then reference it in a third statement:

```sql
-- Declare a variable to hold names as an array.
DECLARE top_names ARRAY<STRING>;

-- Build an array of the top 100 names from the year 2017.
SET top_names = (
  SELECT ARRAY_AGG(name ORDER BY number DESC LIMIT 100)
  FROM `bigquery-public-data`.usa_names.usa_1910_current
  WHERE year = 2017
);

-- Which names appear as words in Shakespeare's plays?
SELECT
  name AS shakespeare_name
FROM UNNEST(top_names) AS name
WHERE name IN (
  SELECT word
  FROM `bigquery-public-data`.samples.shakespeare
);
```

For projects that have enrolled in the scripting alpha, BigQuery interprets any request with multiple statements as a script, unless the statements comprise some number of CREATE TEMP FUNCTION statements followed by a query.

plamut:
Closes #9190.

This PR adds the parent job filter to `client.list_jobs()`.

### How to test
Prerequisite: The (alpha) scripting feature must be enabled for your BigQuery project.

Create and run an SQL script on your project, then list jobs for that project. Find the parent job, i.e. the one whose `num_child_jobs` property is greater than zero, and use its ID to list the jobs again using the `parent_job` filter. That call should correctly return only the parent job's child jobs.

See the new system test in this PR for an example.

**Note**
There currently seems to be a backend bug - the job's `parentJobId` information is only returned when the [jobs.list](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/list) API endpoint is invoked with the `parentJobId` filter, but not when fetching jobs with a different filter such as `minCreationTime`.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1216,6 +1216,7 @@ class Client(ClientWithProject):
     def list_jobs(
         self,
         project=None,
+        parent_job=None,
         max_results=None,
         page_token=None,
         all_users=None,
@@ -1233,6 +1234,11 @@ class Client(ClientWithProject):
             project (str, optional):
                 Project ID to use for retreiving datasets. Defaults
                 to the client's project.
+            parent_job (Optional[Union[ \
+                :class:`~google.cloud.bigquery.job._AsyncJob`, \
+                str, \
+            ]]):
+                If set, retrieve only child jobs of the specified parent.
             max_results (int, optional):
                 Maximum number of jobs to return.
             page_token (str, optional):
@@ -1265,6 +1271,9 @@ class Client(ClientWithProject):
             google.api_core.page_iterator.Iterator:
                 Iterable of job instances.
         """
+        if isinstance(parent_job, job._AsyncJob):
+            parent_job = parent_job.job_id
+
         extra_params = {
             "allUsers": all_users,
             "stateFilter": state_filter,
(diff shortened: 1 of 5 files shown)


plamut:
The system test failure is because the scripting feature is not enabled for the test project.

busunkim96:
@shollyman is working on getting the project whitelisted.

plamut:
@busunkim96 Great, thanks! Let me know when done, so that I can restart the tests.

plamut:
It appears that the scripting feature has been enabled for the test project (the system test passes locally), re-running.

Issue #7534, opened 2019-03-18
Firestore API docs do not reflect the released distributions


hiranya911:
API docs for Cloud Firestore seem to be auto-generated from the master branch: https://googleapis.github.io/google-cloud-python/latest/firestore/collection.html#google.cloud.firestore_v1beta1.collection.CollectionReference

This contains several APIs that are not yet available in any of the released distributions (e.g. `stream()`, `list_documents()`).  This often sends developers down the wrong path, that results in Python attribute errors.

tseaver:
@hiranya911 We don't have a way to version our API-specific docs, so we cannot really address this issue at this time.  We are at work on getting a new release out, which will mitigate the issue a bit (but it will continue to recur).

hiranya911:
Can you generate docs from the latest tag instead of the master branch?

szeitlin:
How's it going with this effort? I'm also finding errors in the docs.

tseaver:
@szeitlin Splitting the repository is prioritized for Q3, which should allow us to build versioned docs for each API.

max-sixty:
FYI [here](https://github.com/googleapis/google-cloud-python/tree/master/firestore#example-usage) the docs are behind rather than ahead—first run I have on firestore using the example from the Readme:  `DeprecationWarning: 'Collection.get' is deprecated:  please use 'Collection.stream' instead.`

tseaver:
@max-sixty Thanks for the report: that sample needs updating.

PR #9235, opened 2019-09-16
Firestore: Update README example to use non-deprecated 'query.get'.

tseaver:
See [this report](https://github.com/googleapis/google-cloud-python/issues/7534#issuecomment-531969738).


tseaver:
Docs-only change, CI is green.

PR #9240, opened 2019-09-17
Storage: Revert "Fix tests b0rken by 'google-resumable-media 0.4.0' release."

tseaver:
Reverts googleapis/google-cloud-python#9204

Should fix tests for #9239.

PR #9242, opened 2019-09-17
Storage: Restrict version range of google-resumable-media.

crwilcox:


storage/setup.py
```diff
@@ -31,7 +31,7 @@
 dependencies = [
     "google-auth >= 1.2.0",
     "google-cloud-core >= 1.0.3, < 2.0dev",
-    "google-resumable-media >= 0.4.0",
+    "google-resumable-media >= 0.3.1, < 0.5.0dev",
```
crwilcox: This is a match to the manually released 1.19.1
We can rev this to 0.4.0 since this was an unreleased change
@tseaver

crwilcox:
```suggestion
    "google-resumable-media >= 0.4.0, < 0.5.0dev",
```

tseaver: ISTM that `>= 0.3.1, != 0.4.0, < 0.5dev` is the right specification, until GRMP has a new version (might be `1.0.0`?) to include deliberately:  the reality is that `0.3.1` and `0.4.1` are near identical, and the storage client works with the older one.

tseaver:
Superseded by #9240.

PR #9246, opened 2019-09-18
Storage: Move benchwrapper directory into 'tests/perf/'.

tseaver:
@jadekler Moving to keep the root directory with just 'tests'.


storage/tests/perf/README.md
```diff
@@ -0,0 +1,21 @@
+# storage benchwrapp
+
+main.py is a gRPC wrapper around the storage library for benchmarking purposes.
+
+## Running
+
+```bash
+$ export STORAGE_EMULATOR_HOST=localhost:8080
```
jadekler: Could you add http:// to this address? I just realized today that it's required.

tseaver: c30db8f62689c276c2885a3dccd7d7eeed53944a

tseaver:
@jadekler Moving those files puts them in a normal / expected location (e.g., so that lint / black run), and eases command-line completion (no conflict between `tests` and `test_utils`).

Issue #7421, opened 2019-02-22
Firestore: Add docs regarding Client threadsafety


danielx:
**Is your feature request related to a problem? Please describe.**
There is no mention in the current docs if a `google.cloud.firestore.Client()` can safely be shared among threads.

 **Describe the solution you'd like**
Some docs explaining if a `Client` can safely be reused among threads.

tseaver:
For example, see `storage/docs/index.rst`.  We should do likewise for all clients.

PR #9254, opened 2019-09-20
docs(firestore): clarify client threadsafety

HemangChothani:
Fixes: #7421


firestore/docs/index.rst
```diff
@@ -1,5 +1,12 @@
 .. include:: README.rst
 
+.. note::
+
+   Because the firestore client uses :mod:`grpc` library and the third-party
```
tseaver: Firestore doesn't use `requests` at all.  We should refer to the `gRPC` library using its PyPI name, e.g., `grpcio`.

PR #9257, opened 2019-09-20
feat(recommender): generate v1beta1

busunkim96:
- [x] write a system test
- [x] add Kokoro configs in GitHub
- [x] add Kokoro configs in google3


busunkim96:
The API endpoint is not yet publicly available, but has been enabled for the CI project.

recommender/tests/system.py
```diff
@@ -0,0 +1,29 @@
...
+from google.cloud.recommender_v1beta1 import ListRecommendationsRequest
+
+
+class TestRecommender(unittest.TestCase):
+    def test_list_recommendations(self):
+        client = Recommender()
+        PROJECT_ID = os.environ.get("PROJECT_ID")
+        parent = f"projects/{PROJECT_ID}/locations/global"
+        request = ListRecommendationsRequest(parent=parent)
+        client.list_recommendations(request=request)
```
busunkim96: The test is failing with an odd error 

```
google.api_core.exceptions.InvalidArgument: 400 Request unsafe for browser client domain: recommender.googleapis.com
```
I think the parent path may be incorrect? I have pinged the engineer internally for guidance.

busunkim96: Looks like this needs to be generated using the monolith.

recommender/docs/index.rst
```diff
@@ -0,0 +1,83 @@
...
+.. _Recommender API Product documentation:  https://cloud.google.com/recommender
+.. _repository’s main README: https://github.com/googleapis/google-cloud-python/blob/master/README.rst
+
+Api Reference
+-------------
+.. toctree::
+    :maxdepth: 2
+
+    gapic/v1beta1/api
+    gapic/v1beta1/types
```
tseaver: This file should be DRY'ed out (using `.. include:: README.rst`) rather than repeating the contents of the top-level `README.rst`.

recommender/synth.py
```diff
@@ -0,0 +1,41 @@
...
+versions = ["v1beta1"]
+common = gcp.CommonTemplates()
+
+
+# ----------------------------------------------------------------------------
+# Generate Cloud Recommender
+# ----------------------------------------------------------------------------
+for version in versions:
+    library = gapic.py_library(
+        "recommender", version
```
tseaver: Needs `include_protos=True`.

recommender/tests/system.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class TestRecommender(unittest.TestCase):
+    def test_list_recommendations(self):
+        client = recommender_v1beta1.RecommenderClient()
+        project_id = os.environ.get("PROJECT_ID")
+        location = "global"
+        recommender = "google.iam.policy.RoleRecommender"
+        parent = client.recommender_path(project_id, location, recommender)
+        client.list_recommendations(parent)
```
tseaver: Do we need to wrap this as `list(client.list_recommendataions(parent))`, in order to exercise the returned response iterator?

recommender/synth.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    s.move(library, excludes=['nox.py'])
+
+# ----------------------------------------------------------------------------
+# Add templated files
+# ----------------------------------------------------------------------------
+templated_files = common.py_library(unit_cov_level=97, cov_level=100)
+s.move(templated_files, excludes=['noxfile.py'])
```
tseaver: We likely need to exclude `docs/index.rst` too, given that the template doesn't do the Right Thing(TM) about DRYing.

PR #9258, opened 2019-09-20
Translation: Add VPC-SC system tests.

pkumusic:


translate/tests/system/test_vpcsc.py
```diff
@@ -0,0 +1,138 @@
...
+    def _do_test(delayed_inside, delayed_outside):
+        if IS_INSIDE_VPCSC.lower() == "true":
+            assert TestVPCServiceControl._is_rejected(delayed_outside)
+            assert not (TestVPCServiceControl._is_rejected(delayed_inside))
+        else:
+            assert not (TestVPCServiceControl._is_rejected(delayed_outside))
+            assert TestVPCServiceControl._is_rejected(delayed_inside)
+
+    @pytest.mark.skipif(
+        PROJECT_INSIDE is None, reason="Missing environment variable: PROJECT_ID"
```
crwilcox: It is possible the environment has a `PROJECT_ID` set even if there isn't a vpc. 

Would it make sense to trigger this on `IS_INSIDE_VPC`?

translate/tests/system/test_vpcsc.py
```diff
@@ -0,0 +1,138 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for VPC-SC."""
+
+import os
+import pytest
```
crwilcox: When run outside of VPCSC (just on a clean/new created project) these all pass. Is this expected?

translate/noxfile.py
```diff
@@ -118,11 +118,18 @@ def system(session):
     session.install("-e", "../test_utils/")
     session.install("-e", ".")
 
+    # Additional setup for VPCSC system tests
+    env = {
+        "PROJECT_ID": os.environ.get("PROJECT_ID"),
+        "GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT": "570941833855",
+        "GOOGLE_CLOUD_TESTS_IN_VPCSC": "true",
```
crwilcox: I am not sure this should be hardcoded. Developers on local test project don't necessarily have a VPCSC configured.

translate/tests/system/test_vpcsc.py
pkumusic: Yes that makes more sense. Changed to using IS_INSIDE_VPC

translate/noxfile.py
pkumusic: I removed those hard-coded env params

PR #9270, opened 2019-09-23
test(speech): use a configurable bucket for systest; add tests for long-running / streaming recognize

wangbill-google:
This is a new test case which will test 3 APIs.

### DIFF
--- a/speech/tests/system/gapic/v1/test_system_speech_v1.py
+++ b/speech/tests/system/gapic/v1/test_system_speech_v1.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2019 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


speech/tests/system/gapic/v1/test_system_speech_v1.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2019 Google LLC
```
busunkim96: Please change this back to 2018. We keep the year the file was originally written in.

speech/tests/system/gapic/v1/test_system_speech_v1.py
```diff
@@ -12,25 +12,77 @@
...
-        uri = "gs://gapic-toolkit/hello.flac"
+
+        uri = "gs://{}/broonklyn.flac".format(BUCKET)
         audio = {"uri": uri}
+
         response = client.recognize(config, audio)
+        assert response.error.code == 200
+
+
+    def test_longRunningRecognize(self):
```
busunkim96: Please use snake_case for function names
```suggestion
    def test_long_running_recognize(self):
```

speech/tests/system/gapic/v1/test_system_speech_v1.py
```diff
@@ -12,25 +12,77 @@
...
+
+        uri = "gs://{}/broonklyn.flac".format(BUCKET)
+        audio = speech_v1.types.RecognitionAudio(uri)
+                  
+        
+        response = client.long_running_recognize(config=config, audio=audio)
+        assert response.error.code == 200
+
+
+    def test_streamingRecognize(self):
```
busunkim96:
```suggestion
    def test_streaming_recognize(self):
```

speech/tests/system/gapic/v1/test_system_speech_v1.py
```diff
@@ -12,25 +12,77 @@
...
+        with io.open(uri,'rb')as stream:
+            requests = [speech_v1.types.StreamingRecognizeRequest(
+                 audio_content=stream.read()
+            )]
+        
+        response = client.streaming_recognize(streamingConfig, requests)
+        assert response.error.code == 200    
+
+    
+if __name__ == "__main__":
```
busunkim96: Please remove this section. We use pytest to run the tests.

```pytest tests/system/gapic/v1/test_system_speech_v1.py```

Alternatively you can use [`nox`](https://nox.thea.codes/en/stable/) to run the tests. It will create a fresh virtual environment and install the necessary dependencies before running the tests.

```nox -s system```

File: speech/tests/system/gapic/v1/test_system_speech_v1.py [hunk 5]
```diff
@@ -12,25 +12,77 @@
...
+            "language_code": "en-US",
+            "sample_rate_hertz": 16000    
         }
-        uri = "gs://gapic-toolkit/hello.flac"
+
+        uri = "gs://{}/broonklyn.flac".format(BUCKET)
         audio = {"uri": uri}
+
         response = client.recognize(config, audio)
+        assert response.error.code == 200
```
busunkim96: When the call fails, an exception is raised. When it is successful a `RecognizeResponse` object is returned. That object has the field`results`.  See the [code sample](https://cloud.google.com/speech-to-text/docs/quickstart-client-libraries#make_an_audio_transcription_request) for audio transcription requests.

```py
response = client.recognize(config, audio)

for result in response.results:
    print('Transcript: {}'.format(result.alternatives[0].transcript))
```

Library reference docs for the function are [here](https://googleapis.dev/python/speech/latest/gapic/v1/api.html#google.cloud.speech_v1.SpeechClient.recognize).

I don't know about the details of the speech API, but it might be nicer to make a stronger assert about `response.results`

speech/tests/system/gapic/v1/test_system_speech_v1.py
```diff
@@ -12,25 +12,77 @@
...
-import time
+
+import os
+import io
 
 from google.cloud import speech_v1
-from google.cloud.speech_v1 import enums
-from google.cloud.speech_v1.proto import cloud_speech_pb2
 
+BUCKET = os.environ["GOOGLE_CLOUD_TESTS_SPEECH_BUCKET"]
```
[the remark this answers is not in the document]
tseaver: Needs a default value, as well.

Issue #9250, opened 2019-09-18
Firestore: Clarify documentation for '{Document,Collection}.on_snapshot'.


billyrrr:
```on_snapshot``` defined in ```DocumentReference``` documentation is inconsistent with those of ```Watch``` and Product Documentation. 

In ```DocumentReference``` documentation, we have 

https://github.com/googleapis/google-cloud-python/blob/f890d2916e67e40dfbdfeb814d6b92fce9103f5b/firestore/google/cloud/firestore_v1/document.py#L503-L504

and 

https://github.com/googleapis/google-cloud-python/blob/f890d2916e67e40dfbdfeb814d6b92fce9103f5b/firestore/google/cloud/firestore_v1/document.py#L515-L517

In ```Watch``` documentation, we have 

https://github.com/googleapis/google-cloud-python/blob/f890d2916e67e40dfbdfeb814d6b92fce9103f5b/firestore/google/cloud/firestore_v1/watch.py#L183-L191

In [Product Documentation](https://firebase.google.com/docs/firestore/query-data/listen#listen_to_multiple_documents_in_a_collection), we have 

```python
# Create a callback on_snapshot function to capture changes
def on_snapshot(doc_snapshot, changes, read_time):
    for doc in doc_snapshot:
        print(u'Received document snapshot: {}'.format(doc.id))

doc_ref = db.collection(u'cities').document(u'SF')

# Watch the document
doc_watch = doc_ref.on_snapshot(on_snapshot)
```

The function signature in ```DocumentReference``` takes 1 argument, but the function signature in Watch and Product Documentation takes 3 arguments. Are these referring to the same function? If so, the documentation in ```DocumentReference``` may be updated to reflect these changes. If not, ```on_snapshot``` in Product Documentation may be renamed to a different name than ```on_snapshot``` in ```DocumentReference``` to be more clear. 

Will probably submit a pull request to update documentation in ```DocumentReference``` to be consistent with that in ```Watch```. 

Thanks!

tseaver:
@billyrrr Thanks for the report.  The example for `Query.on_snapshot` is correct: https://github.com/googleapis/google-cloud-python/blob/e99396119bfa0bcddf8fc475a8e43ea70a36ad97/firestore/google/cloud/firestore_v1/query.py#L804-L806

while `Document.on_snapshot` and `Collection.on_snapshot` show invalid uses.  All three need to update the `callback` argument specification to show the three arguments.

PR #9275, opened 2019-09-24
docs(firestore): add documentation for Document,Collection .on_snapshot

HemangChothani:
Fixes: #9250


tseaver:
Looking more closely, the `Query.on_snapshot` example is also incorrect:  the first argument to the callback is a sorted list of keys / document paths in the updated tree.  The example should illustrate how a real callback would handle all three arguments.

@crwilcox Please clarify / correct my understanding here, and suggest what a good callback example would look like.

HemangChothani:
@tseaver `conventionalcommits.org`  check throws  linting errors for title and give suggestion to edit your pull request title to match Conventional Commit guidelines, due to that i have changed the title.

PR #9287, opened 2019-09-24
test(speech): add v1p1beta1 systests for longrunning / streaming recognize

wangbill-google:
New test case for v1p1beta1 which tests 3 APIs.

### DIFF
--- a/speech/tests/system/gapic/v1p1beta1/test_system_speech_v1p1beta1.py
+++ b/speech/tests/system/gapic/v1p1beta1/test_system_speech_v1p1beta1.py
@@ -18,7 +18,6 @@ import io
 import requests
 
 from google.cloud import speech_v1p1beta1
-from google.cloud import storage
 
 class TestSystemSpeech(object):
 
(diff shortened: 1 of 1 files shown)


busunkim96:
Would it be possible to get the changes into one PR? It looks like they're the same tests, just against different versions of the API. In addition, it looks like the system tests are failing speech. 

```
_ ERROR collecting tests/system/gapic/v1p1beta1/test_system_speech_v1p1beta1.py _
ImportError while importing test module '/tmpfs/src/github/google-cloud-python/speech/tests/system/gapic/v1p1beta1/test_system_speech_v1p1beta1.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/system/gapic/v1p1beta1/test_system_speech_v1p1beta1.py:21: in <module>
E   ImportError: cannot import name 'storage' from 'google.cloud' (/tmpfs/src/github/google-cloud-python/core/google/cloud/__init__.py)
```

wangbill-google:
Yes, we can delete this line: 

PR #9289, opened 2019-09-25
feat(bigtable): add 'client_options' support to Client

Emar-Kar:
Towards #8475

List of changes:
-
1. **client**:
    - as **BigTable Client** creates by **GAPIC** `client_options` were integrated into the `_create_gapic_client` method call.
2. **setup**:
    - `google-cloud-core` version updated on the basis of the [comment](https://github.com/googleapis/google-cloud-python/pull/8999#discussion_r315803265);
    - the file was reformatted by `black`.
3. **test_client**:
    - since `api_endpoint` is an irrevocable argument, tests check the correctness of the API call;
    - by default `client_options`should be **None**.


tseaver:
This PR has several issues:

- Bigtable has *two* endpoints, one for table data and one for table / instance admin, and therefore the client constructor needs both a `client_options` and an `admin_client_options` argument.
- The correct `client_options` / `admin_client_options` arguments must be passed down to the GAPIC class, which means that `_create_gapic_client` must take that value as an argument, rather than looking it up on the instance.

PR #9516 corrects those two issues.

Also, please avoid intermingling janitorial changes (blackening `setup.py`) with substantive ones.

PR #9299, opened 2019-09-25
enhancement(automl): support for passing  project to 'GcsClient'

mcollins42:
Changes:

- `GcsClient.__init__` takes an optional project parameter and passes it to the `storage.Client` constructor.
- TablesClient stores the credentials passed to the constructor and passes the credential when creating the `GcsClient`
- Ensure that project, region and credentials fall back to the values passed to the TablesClient constructor if no value is passed as an override on public methods

I had a little trouble testing that parameters were passed to the constructors for `GcsClient` and `storage.Client`. I tried various combinations of mock and patch, but even `MagicMock` doesn't support mocking `__init__()`. Any suggestions on better approaches would be appreciated.


lwander:
Looks like you need to change the commit message (rather than the PR title): https://github.com/googleapis/google-cloud-python/pull/9299/checks?check_run_id=236270561

`git commit --amend -m "fix(automl): Add support for passing a project to GcsClient and for TablesClient passing the default project and credentials"` should do the trick

mcollins42:
> Looks like you need to change the commit message (rather than the PR title): https://github.com/googleapis/google-cloud-python/pull/9299/checks?check_run_id=236270561
> 
I've updated the commit messages, but the checks haven't run again. Do I need to make a trivial commit or is there another way to force the checks to run again?

mcollins42:
@tseaver Should I also update the type on the commit messages from fix to enhancement?

automl/google/cloud/automl_v1beta1/tables/gcs_client.py
```diff
@@ -54,18 +54,24 @@ def __init__(self, bucket_name=None, client=None, credentials=None):
...
         if client is not None:
             self.client = client
         elif credentials is not None:
-            self.client = storage.Client(credentials=credentials)
+            self.client = storage.Client(credentials=credentials, project=project)
         else:
             self.client = storage.Client()
 
         self.bucket_name = bucket_name
+        self.project = project
```
tseaver: This attribute is unused.

mcollins42: Removed and updated the docstring. Thanks!

automl/google/cloud/automl_v1beta1/tables/gcs_client.py
```diff
@@ -54,14 +54,18 @@ def __init__(self, bucket_name=None, client=None, credentials=None):
                 credentials identify this application to the service. If none
                 are specified, the client will attempt to ascertain the
                 credentials from the environment.
+            project (Optional[string]): The project to attach to the underlying
```
busunkim96: Could this be re-worded to mention that this is the project ID?
```suggestion
            project (Optional[str]): The project to attach to the underlying
```

mcollins42: Sure. Should I change the wording and string type for the existing `TablesClient` docstrings as well? It would be nice for them to be consistent.

mcollins42: Updated docstrings to reference project ID explicitly and changed `string` to `str` in type annotations.

PR #9321, opened 2019-09-26
fix(bigquery): use pyarrow fallback for improved schema detection

### LINKED ISSUE #9206: BigQuery: load_table_from_dataframe fails on datetime64 column used for partitioning, saying it's an INTEGER
Versions:
 ```
google-cloud-bigquery==1.19.0
```
We were initially using 1.18.0 but I noticed #9044 was included in 1.19.0 so we tried that as well, but it made no difference.

We're using a pandas dataframe read from parquet, example data would be
```
id status created_at execution_date
0 1 NEW 2018-09-12 09:24:31.291 2019-05-10 17:40:00
1 2 NEW 2018-09-12 09:26:45.890 2019-05-10 17:40:00
[628 rows x 4 columns]
```

`df.dtypes` shows:
```
id object
status object
created_at datetime64[ns]
dataplatform_execution_date datetime64[ns]
dtype: object
```

When trying to load this into BigQuery using `load_table_from_dataframe()` and setting the             job config's `time_partitioning` to `bigquery.table.TimePartitioning(field="execution_date")` we get the following error:
```
The field specified for time partitioning can only be of type TIMESTAMP or DATE. The type found is: INTEGER.
```
Which doesn't really make sense, since the field is clearly a `datetime64`.
The job config shown in the console looks correct (ie it's set to partitioned by day and it's using the correct field).

edit:
It seems the cause for this is that Dataframe columns of type `datetime64` are being converted to type `INTEGER` instead of `DATE` (or `TIMESTAMP`? I'm not sure which one would be the correct type in BigQuery).

edit2: 
Could it be this mapping is wrong and it should be `DATE` instead of `DATETIME`?
https://github.com/googleapis/google-cloud-python/blob/dce13268e9bdf2df6a3c7a2df6db3b8a1bb51f75/bigquery/google/cloud/bigquery/_pandas_helpers.py#L55

### PR DESCRIPTION
**Pull request by plamut**
Closes #9206.

This PR adds additional logic to schema autodetection if `pyarrow` is available, and the types could not be detected for all columns.

### How to test
 - Make sure that pyarrow is installed
- Try loading data from a dataframe to a new table when the dataframe contains columns of type string, date, etc. (those that end up being the Pandas dtype "object"). Do not provide an explicit schema to the load job.

The backend responds with an error (incompatible types).

The load job completes successfully, and column types are correctly detected (strings, dates...). 

### Misc
The code might not work for _all_ scalar types, depending on how good the pyarrow's schema detection logic is. Which types to we care about the most?

We should update the test case to include all of them (we currently test strings and dates).

### CODE REVIEW

File: bigquery/google/cloud/bigquery/_pandas_helpers.py [hunk 1]
```diff
@@ -285,6 +294,44 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
...
             )
         )
+
+    # If schema detection was not successful for all columns, also try with
+    # pyarrow, if available.
+    if any(field.field_type is None for field in bq_schema_out):
+        if not pyarrow:
+            return None  # We cannot detect the schema in full.
+
+        arrow_table = dataframe_to_arrow(dataframe, bq_schema_out)
```
plamut: Converting the entire dataframe to pyarrow tables is not ideal, of course - is there a shortcut?

Thought: We might want to issue an extra warning here, warning about a possible negative performance impact for large dataframes? On the plus side, it's at least a last resort in schema autodetection, and not a main code path.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -110,8 +110,13 @@ def pyarrow_timestamp():
         "TIME": pyarrow_time,
         "TIMESTAMP": pyarrow_timestamp,
     }
+    ARROW_SCALARS_TO_BQ = {
+        arrow_type(): bq_type  # TODO: explain wht calling arrow_type()
```
tswast: I don't think we're supposed to use Arrow type objects as dictionary keys. The factory functions all describe instances of data types (https://arrow.apache.org/docs/python/api/datatypes.html#factory-functions), so I don't think we're guaranteed on equality / consistent hash between two Arrow data types.

That said, I'm not sure what alternatives we have. We could have a function with a whole bunch of if statements calling the type checking functions: https://arrow.apache.org/docs/python/api/datatypes.html#type-checking but that sounds dirty and slow (though more guaranteed to be correct). For pandas, I use the string name of the dtype, but I'm not seeing an equivalent for Arrow.

bigquery/google/cloud/bigquery/_pandas_helpers.py
tswast: It'd be slightly better to make the conversion only on the columns whose types we could not determine. Does the array function work okay when passed a column from a dataframe? https://arrow.apache.org/docs/python/generated/pyarrow.array.html#pyarrow.array

Also, I wonder if we should amend our return type to include the converted arrays? We'll have to do the same conversion later on, when we serialize the Parquet file. It'd be nice to avoid making the conversion twice. -- If we do this, then maybe we should just always do the conversion to Arrow arrays, anyway?

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -285,6 +294,44 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
...
+        arrow_table = dataframe_to_arrow(dataframe, bq_schema_out)
+        arrow_schema_index = {field.name: field.type for field in arrow_table}
+
+        currated_schema = []
+        for schema_field in bq_schema_out:
+            if schema_field.field_type is not None:
+                currated_schema.append(schema_field)
+                continue
+
+            detected_type = ARROW_SCALARS_TO_BQ.get(
```
tswast: Probably best if we make a function for Arrow type + field name -> BQ field now, since we know we'll want to handle repeated / nested fields at some point.

bigquery/google/cloud/bigquery/_pandas_helpers.py
plamut: Each pyarrow type has a numeric ID that we can use instead. It's true that we don't know if these IDs are going to stay the same forever, but the map is built dynamically, and the IDs are consistent within a particular `pyarrow` version.

What do you think?

bigquery/google/cloud/bigquery/_pandas_helpers.py
plamut: Limiting ourselves only to the "problematic" columns makes sense, will check how it works.

I will also see if we can avoid unnecessary double conversions, good point.

bigquery/google/cloud/bigquery/_pandas_helpers.py
plamut: I extended the types map in a provisional new commit, feel free to scrutinize. :)

bigquery/google/cloud/bigquery/_pandas_helpers.py
plamut: The code now only converts the columns that need extra schema detection effort. Passing a dataframe column to `pyarrow.array()` works smoothly.

plamut:
@tswast Updated the code and extended the tests.

> Also, I wonder if we should amend our return type to include the converted arrays?

Looking at the code, that would require changing the signatures of the following Pandas helpers:
- `currate_schema()` - return type,
- `dataframe_to_bq_schema()` - return type,
- `dataframe_to_parquet()` and `dataframe_to_arrow()` - additional parameter holding any dataframe columns already converted pyarrow arrays.

Returning `(schema, Optional[arrow_arrays])` does not feel like a clean design, and these changes would be added for a use case that we discourage (and will likely deprecate in the future), i.e. relying on implicit schema detection. And even that only for loading the data into a _new_ table when mode != `WRITE_TRUNCATE`.

Unless there are serious performance problems reported, we are probably better off promoting the explicit schema usage IMO, and discouraging implicit schemas.

bigquery/google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -285,7 +318,73 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
...
+            return None  # We cannot detect the schema in full.
+
+        # The currate_schema() helper itself will also issue unknown type
+        # warnings if detection still fails for any of the fields.
+        bq_schema_out = currate_schema(dataframe, bq_schema_out)
+
+    return tuple(bq_schema_out) if bq_schema_out else None
+
+
+def currate_schema(dataframe, current_bq_schema):
```
tswast: I'm not familiar with the verb "currate" (did you mean "curate"?). Augment may be a better verb for what we're doing in this function, as we're adding to the schema. Curate implies removing items.

plamut: Searching online, it appears that I actually came up with a dubious word, subconsciously trying to score extra points in scrabble or something. :confused: 

Augment indeed sounds better.

PR #9332, opened 2019-09-27
docs(storage): clarify docstring for 'Blob.download_as_string'

### LINKED ISSUE #9290: Storage: download_as_string documentation says that it downloads a string, instead of bytes
For google-cloud-storage, in the documentation for [`download_as_string()`](https://github.com/googleapis/google-cloud-python/blob/53d692517345979b2b4621872d76c1184c987b92/storage/google/cloud/storage/blob.py#L714) method on blobs, it says [on line 715](https://github.com/googleapis/google-cloud-python/blob/53d692517345979b2b4621872d76c1184c987b92/storage/google/cloud/storage/blob.py#L715): 

        """Download the contents of this blob as a string.

This is not true. The confusingly named method downloads the content as bytes, not string.

On [line 731](https://github.com/googleapis/google-cloud-python/blob/53d692517345979b2b4621872d76c1184c987b92/storage/google/cloud/storage/blob.py#L731), the return type is correctly declared as bytes, but reading the documentation, you are still presented with conflicting information.

tseaver:
Specify that the returned value is bytes in the summary, and note the
historical reason for the confused method name.

Closes #9290.

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -714,10 +714,6 @@ class Blob(_PropertyMixin):
     def download_as_string(self, client=None, start=None, end=None):
         """Download the contents of this blob as a bytes object.
 
-        .. note::
-            the method name, ``download_as_string`` was chosen for
-            compatibility with the ``boto`` library's API.
-
         If :attr:`user_project` is set on the bucket, bills the API request
         to that project.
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -712,7 +712,11 @@ def download_to_filename(self, filename, client=None, start=None, end=None):
             os.utime(file_obj.name, (mtime, mtime))
 
     def download_as_string(self, client=None, start=None, end=None):
-        """Download the contents of this blob as a string.
+        """Download the contents of this blob as a bytes object.
+
+        .. note::
+            the method name, ``download_as_string`` was chosen for
+            compatibility with the ``boto`` library's API.
```
frankyn: Thanks @tseaver, I'd rather not explicitly state boto in documentation and only capture "bytes" being returned. 

It does related questions about the surface design used by the Python library.

PR #9268, opened 2019-09-23
fix(pubsub): set default stream ACK deadline to subscriptions'

plamut:
Closes #9252. 

This PR mitigates the reported issue by using the subscription's max ACK deadline when establishing a streaming pull, instead of the previous optimistic ACK deadline of 10 seconds.

(when creating a streaming pull, the ACK timing histogram has no data yet, thus its 99 % value defaults to 10 seconds... also see the commit message for some extra info)

Additionally, it adds a debug log message with the ACK deadline used for the streaming pull, as that important piece of information was missing previously.

### How to test
Run the example from the issue description, and verify that the backend does not send any duplicate messages, and the subscription's ACK deadline is respected (300 s in that particular case).

### Details
The cause of the bug was that the client overrode the max ACK deadline on the subscription with the default minimum deadline of 10 seconds.

This became problematic for the received messages that exceeded the `FlowControl.max_messages` setting, and were thus put on hold in a buffer. Since these messages were not lease-managed, and their ACK deadlines were not being extended, the overridden ACK deadline of 10 seconds was easily hit, resulting in re-delivery.

There are several options on how to deal with these on hold messages:

- **Discard them immediately, and set their ACK deadlines to 0.**
  This would cause the backend to re-send them immediately, which would work well with multiple subscribers. However, if using a single subscriber, low `max_messages` setting, and processing each message takes a lot of time, large published message batches would cause a lot of unnecessary traffic. 

  The excessive messages in a batch (those above `max_messages`) would be discarded and re-delivered over and over again, because each time the client would dispatch only a small portion of the batch to user's callbacks.

- **Add all received messages to lease management, including those put on hold.**
  This would avoid ACK timeouts and message re-delivery, but could shoot the leaser's load well over `1.0`. When user code would process and acknowledge some of the messages, the client would remove them from the lease management, but the load could still stay above `1.0`

  Since the leaser would still be overloaded, no additional messages would be dispatched to the callbacks, and the stream would not be resumed --> deadlock (there would be nothing to trigger resuming the streaming pull).

  While there could be a way around that, it would likely require rewriting significant chunks of the streaming pull logic, risking new bugs.

- **Ignoring the messages on hold until the load allows processing them.**
  This is the path chosen. Only the messages that are _not_ put on hold are added to the lease management, while the rest sit in the buffer, waiting to be released when the load allows for it. If they are released soon enough, the auto-lease management steps in. 

[...]

Issue #9252, opened 2019-09-19
Pub/Sub: buffered messages delivered multiple times ignoring acknowledgment deadline


qvik-olli:
I noticed a problem where buffered messages in pubsub library are delivered multiple times ignoring acknowledgment deadline. I have a subscription with acknowledgment deadline set to 5 minutes but messages are delivered multiple times to the same subscriber after about 30 seconds.

My code example starts a subscriber client with `max_messages` set to 1. When two messages are sent to the subscriptions, the second message is always duplicated. This same behavior also happens with higher `max_messages` values. For example, using the default value 100 for `max_messages`, when 150 messages are sent to the subscription, some of the messages are duplicated if the processing of the first 100 messages takes more than 30 seconds.

I understand that with large backlog of small messages, the messages can get redelivered ([Dealing with large backlogs of small messages](https://cloud.google.com/pubsub/docs/pull#streamingpull_dealing_with_large_backlogs_of_small_messages)), however, shouldn't the messages be re-delivered only after the ack_deadline is exceeded?

#### Environment details
- OS type and version: Ubuntu 18.04.3 LTS
- Python version and virtual environment information: Python 3.7.3
- google-cloud-pubsub version: google-cloud-pubsub==1.0.0

#### Steps to reproduce

  1. Create a subscription with a high ack_deadline.
  2. Start subscriber (see code examples): `python subscriber.py -p PROJECT_ID SUBSCRIPTION_NAME`.
  3. Send two messages to the pubsub topic.
  4. The second message is duplicated.

#### Code example
Example subscriber code here: https://gist.github.com/qvik-olli/0bfd4ace2d06def1675a76fbc20493e5

#### Logs
From the subscriber script:
```
2019-09-19 14:10:37,740 INFO ThreadPoolExecutor-ThreadScheduler_0: [652240118977134] got message with content: b'0'
2019-09-19 14:10:37,740 INFO ThreadPoolExecutor-ThreadScheduler_0: [652240118977134] sleeping
2019-09-19 14:11:17,747 INFO ThreadPoolExecutor-ThreadScheduler_0: [652240118977134] done sleeping
2019-09-19 14:11:17,819 INFO ThreadPoolExecutor-ThreadScheduler_1: [652240118977135] got message with content: b'1'
2019-09-19 14:11:17,819 INFO ThreadPoolExecutor-ThreadScheduler_1: [652240118977135] sleeping
[...]

PR #9360, opened 2019-09-30
fix(pubsub): streaming pull shouldn't require subscriptions.get permission

### LINKED ISSUE #9339: Subscribing to a Pub/Sub topic requires additional permissions as a result of the release of google-cloud-pubsub 1.0.1
Prior to todays release of the pubsub client, we were able to subscribe with only `roles/pubsub.subscriber` on the subscription we were trying to subscribe to. Now there is a call to get the subscription, which isn't permitted by that role.

It looks like this commit introduced the issue:
https://github.com/googleapis/google-cloud-python/pull/9268/files#diff-1004b3646a787f6b0438604df260553aR387

So after that release, our application broke (we happened to do a release a few hours after yours that picked up the latest pubsub library from pypi) because it did not have permission to get the subscription's metadata. After investigating for a while we were able to resolve the issue by adding `roles/pubsub.viewer` to the subscription in question. We had to enable Pub/Sub audit logs to get enough details to identify what was going on. Here's an example of the log that pointed us to the permission we were missing:

```
 {
   "protoPayload": {
     "@type": "type.googleapis.com/google.cloud.audit.AuditLog",
     "status": {
       "code": 7,
       "message": "PERMISSION_DENIED"
     },
     "authenticationInfo": {
       "principalEmail": "xxx@xxx.iam.gserviceaccount.com"
     },
     "requestMetadata": {
       "callerIp": "xxx",
       "callerSuppliedUserAgent": "grpc-python/1.24.0 grpc-c/8.0.0 (manylinux; chttp2; ganges),gzip(gfe)",
       "callerNetwork": "//compute.googleapis.com/projects/xxx/global/networks/__unknown__",
       "requestAttributes": {
         "time": "2019-09-27T22:41:46.042805867Z",
         "auth": {}
       },
       "destinationAttributes": {}
     },
     "serviceName": "pubsub.googleapis.com",
     "methodName": "google.pubsub.v1.Subscriber.GetSubscription",
     "authorizationInfo": [
       {
         "resource": "projects/xxx/subscriptions/xxx",
         "permission": "pubsub.subscriptions.get",
         "resourceAttributes": {}
       }
     ],
     "resourceName": "projects/xxx/subscriptions/xxx",
     "request": {
       "subscription": "projects/xxx/subscriptions/xxx",
       "@type": "type.googleapis.com/google.pubsub.v1.GetSubscriptionRequest"
     }
   },
   "insertId": "xxx",
   "resource": {
     "type": "pubsub_subscription",
     "labels": {
       "project_id": "xxx",
       "subscription_id": "projects/xxx/subscriptions/xxx"
     }
   },
   "severity": "ERROR",
   "logName": "projects/xxx/logs/cloudaudit.googleapis.com%2Fdata_access",
 }
```

plamut:
Closes #9339 
Re-introduces #9252. 

This PR fixes a [regression](https://github.com/googleapis/google-cloud-python/pull/9268) from the `1.0.1` release that caused the streaming pull to additionally require the `subscriptions.get` permission (the default `pubsub.subscriber` role was not sufficient anymore for the streaming pull).

By avoiding the call that fetched a subscription and read its ACK deadline, and replacing that deadline with a fixed value of 60 seconds, this *will* re-introduce #9252, albeit in a less severe manner.

### How to test
Using an account that has no roles assigned other than `roles/pubsub.subscriber`, try opening a streaming pull (`subscriber.subscribe()`). That call should succeed without a permission error.

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -50,6 +50,13 @@ _MAX_LOAD = 1.0
 _RESUME_THRESHOLD = 0.8
 """The load threshold below which to resume the incoming message stream."""
 
+_DEFAULT_STREAM_ACK_DEADLINE = 60
+"""The default message acknowledge deadline in seconds for incoming message stream.
+
+This default deadline is dynamically modified for the messages that are added
+to the lease management.
+"""
+
 
 def _maybe_wrap_exception(exception):
     """Wraps a gRPC exception class, if needed."""
@@ -384,8 +391,17 @@ class StreamingPullManager(object):
         )
 
         # Create the RPC
-        subscription = self._client.api.get_subscription(self._subscription)
-        stream_ack_deadline_seconds = subscription.ack_deadline_seconds
+
+        # We must use a fixed value for the ACK deadline, as we cannot read it
+        # from the subscription. The latter would require `pubsub.subscriptions.get`
+        # permission, which is not granted to the default subscriber role
+        # `roles/pubsub.subscriber`.
+        # See also https://github.com/googleapis/google-cloud-python/issues/9339
+        #
+        # When dynamic lease management is enabled for the "on hold" messages,
+        # the default stream ACK deadline should again be set based on the
+        # historic ACK timing data, i.e. `self.ack_histogram.percentile(99)`.
+        stream_ack_deadline_seconds = _DEFAULT_STREAM_ACK_DEADLINE
 
         get_initial_request = functools.partial(
             self._get_initial_request, stream_ack_deadline_seconds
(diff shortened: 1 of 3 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -384,8 +391,17 @@ def open(self, callback, on_callback_error):
...
-        stream_ack_deadline_seconds = subscription.ack_deadline_seconds
+
+        # We must use a fixed value for the ACK deadline, as we cannot read it
+        # from the subscription. The latter would require `pubsub.subscriptions.get`
+        # permission, which is not granted to the default subscriber role
+        # `roles/pubsub.subscriber`.
+        # See also https://github.com/googleapis/google-cloud-python/issues/9339
+        #
+        # When dynamic lease management is enabled for the "on hold" messages,
+        # the defautl stream ACK deadline should again be set based on the
```
pradn: Spelling "defautl"

PR #9365, opened 2019-10-01
feat(pubsub): add stop method

### LINKED ISSUE #4913: Pubsub API should provide a way to synchronize a set of operations
The pubsub API provides no easy way to make sure a bunch of publishes finish before continuing. You have to either deal with batches yourself or wait on every Future returned. A context manager would do the job, or even a wait method on the client.

### LINKED ISSUE #6883: Pubsub: How to flush unsent messages on program exit?
I have set caching configuration which may cache some messages and messages are not sent to pubsub yet. Is there a `flush` method when I want to exit the program.

IlyaFaer:
Add `stop()` method, which sends all outstanding messages and waits until all futures resolved. Similar features in [Go](https://github.com/googleapis/google-cloud-go/blob/bdcd41efeb09c795b4c380d194b2a7a19f945176/pubsub/topic.go#L414-L426) and [Java](https://github.com/googleapis/google-cloud-java/blob/bec495ff9106f863098b0089c51cd6a21ca9f6b2/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java#L492-L507).

Closes #4913
Closes #6883

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/futures.py
+++ b/pubsub/google/cloud/pubsub_v1/futures.py
@@ -74,9 +74,7 @@ class Future(google.api_core.future.Future):
             bool: ``True`` if this method has not yet completed, or
                 ``False`` if it has completed.
         """
-        if self.done():
-            return False
-        return True
+        return not self.done()
 
     def done(self):
         """Return True the future is done, False otherwise.
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -140,17 +140,6 @@ class Batch(base.Batch):
         """
         return self._status
 
-    def wait(self):
-        """If commit is in progress, waits until all of the futures resolved.
-
-        .. note::
-
-            This method blocks until all futures of this batch resolved.
-        """
-        if self._status != base.BatchStatus.ACCEPTING_MESSAGES:
-            for future in self._futures:
-                future.result()
-
     def commit(self):
         """Actually publish all of the messages on the active batch.
 
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -75,6 +75,9 @@ class Batch(base.Batch):
         # These members are all communicated between threads; ensure that
         # any writes to them use the "state lock" to remain atomic.
         self._futures = []
+        # _futures list should remain unchanged after batch
+        # status changed from ACCEPTING_MESSAGES to any other
+        # in order to avoid race conditions
         self._messages = []
         self._size = 0
         self._status = base.BatchStatus.ACCEPTING_MESSAGES
@@ -138,13 +141,13 @@ class Batch(base.Batch):
         return self._status
 
     def wait(self):
-        """If commit in progress, waits until all of the futures resolved.
+        """If commit is in progress, waits until all of the futures resolved.
 
         .. note::
 
             This method blocks until all futures of this batch resolved.
         """
-        if self._status in (base.BatchStatus.STARTING, base.BatchStatus.IN_PROGRESS):
+        if self._status != base.BatchStatus.ACCEPTING_MESSAGES:
             for future in self._futures:
                 future.result()
 
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -137,6 +137,17 @@ class Batch(base.Batch):
         """
         return self._status
 
+    def wait(self):
+        """If commit in progress, waits until all of the futures resolved.
+
+        .. note::
+
+            This method blocks until all futures of this batch resolved.
+        """
+        if self._status in (base.BatchStatus.STARTING, base.BatchStatus.IN_PROGRESS):
+            for future in self._futures:
+                future.result()
+
     def commit(self):
         """Actually publish all of the messages on the active batch.
 
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -74,10 +74,10 @@ class Batch(base.Batch):
         self._state_lock = threading.Lock()
         # These members are all communicated between threads; ensure that
         # any writes to them use the "state lock" to remain atomic.
-        self._futures = []
         # _futures list should remain unchanged after batch
         # status changed from ACCEPTING_MESSAGES to any other
         # in order to avoid race conditions
+        self._futures = []
         self._messages = []
         self._size = 0
         self._status = base.BatchStatus.ACCEPTING_MESSAGES
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -252,7 +252,7 @@ class Client(object):
         # If it is literally anything else, complain loudly about it.
         if not isinstance(data, six.binary_type):
             raise TypeError(
-                "Data being published to Pub/Sub must be sent " "as a bytestring."
+                "Data being published to Pub/Sub must be sent as a bytestring."
             )
 
         # Coerce all attributes to text strings.
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -274,3 +274,22 @@ class Client(object):
                 batch = self._batch(topic, create=True)
 
         return future
+
+    def stop(self):
+        """Immediately publish all outstanding batches.
+
+        This asynchronously pushes all outstanding messages
+        and waits until all futures resolved. Method should be
+        invoked prior to deleting this Client object in order
+        to ensure that no pending messages are lost.
+
+        .. note::
+
+            This method blocks until all futures of all
+            batches resolved.
+        """
+        for topic in self._batches:
+            self._batches[topic].commit()
+
+        for topic in self._batches:
+            self._batches[topic].wait()
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -294,8 +294,8 @@ class Client(object):
         .. note::
 
             This method is non-blocking. Use `Future()` objects
-            returned by `publish()` to make sure all messages
-            sent or failed.
+            returned by `publish()` to make sure all publish
+            requests completed, either in success or error.
         """
         if self._is_stopped:
             raise ValueError("Cannot stop a publisher already stopped.")
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -134,6 +134,7 @@ class Client(object):
         # messages. One batch exists for each topic.
         self._batch_lock = self._batch_class.make_lock()
         self._batches = {}
+        self._is_stopped = False
 
     @classmethod
     def from_service_account_file(cls, filename, batch_settings=(), **kwargs):
@@ -242,7 +243,14 @@ class Client(object):
             instance that conforms to Python Standard library's
             :class:`~concurrent.futures.Future` interface (but not an
             instance of that class).
+
+        Raises:
+            ValueError:
+                If called after publisher has been stopped
+                by a `stop()` method call.
         """
+        if self._is_stopped:
+            raise ValueError("Cannot publish on a stopped publisher.")
         # Sanity check: Is the data being sent as a bytestring?
         # If it is literally anything else, complain loudly about it.
         if not isinstance(data, six.binary_type):
@@ -276,20 +284,22 @@ class Client(object):
         return future
 
     def stop(self):
-        """Immediately publish all outstanding batches.
+        """Immediately publish all outstanding messages.
 
-        This asynchronously pushes all outstanding messages
-        and waits until all futures resolved. Method should be
-        invoked prior to deleting this Client object in order
-        to ensure that no pending messages are lost.
+        Asynchronously sends all outstanding messages and
+        prevents future calls to `publish()`. Method should
+        be invoked prior to deleting this `Client()` object
+        in order to ensure that no pending messages are lost.
 
         .. note::
 
-            This method blocks until all futures of all
-            batches resolved.
+            This method is non-blocking. Use `Future()` objects
+            returned by `publish()` to make sure all messages
+            sent or failed to.
         """
+        if self._is_stopped:
+            raise ValueError("Cannot stop a publisher already stopped.")
+
+        self._is_stopped = True
         for topic in self._batches:
             self._batches[topic].commit()
-
-        for topic in self._batches:
-            self._batches[topic].wait()
--- a/pubsub/google/cloud/pubsub_v1/publisher/client.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/client.py
@@ -244,7 +244,7 @@ class Client(object):
             instance of that class).
 
         Raises:
-            ValueError:
+            RuntimeError:
                 If called after publisher has been stopped
                 by a `stop()` method call.
         """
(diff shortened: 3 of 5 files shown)


pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -74,9 +74,7 @@ def running(self):
             bool: ``True`` if this method has not yet completed, or
                 ``False`` if it has completed.
         """
-        if self.done():
-            return False
-        return True
+        return not self.done()
```
IlyaFaer: This can be refactored a little by the way

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -274,3 +274,22 @@ def publish(self, topic, data, **attrs):
...
+        .. note::
+
+            This method blocks until all futures of all
+            batches resolved.
+        """
+        for topic in self._batches:
+            self._batches[topic].commit()
+
+        for topic in self._batches:
+            self._batches[topic].wait()
```
IlyaFaer: @anguillanneuf, I've changed my plan after seeing Java's analogue. Now on `stop()` call all batches will start commiting in parallels (`commit()` creates a thread). After starting all of them, we'll just wait them in cycle. Still, we can return to sequential variant, if this is not an option

pubsub/google/cloud/pubsub_v1/futures.py
```diff
@@ -74,9 +74,7 @@ def running(self):
             bool: ``True`` if this method has not yet completed, or
                 ``False`` if it has completed.
         """
-        if self.done():
-            return False
-        return True
```
IlyaFaer: This can be refactored a little by the way

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -137,6 +137,17 @@ def status(self):
...
+        """If commit in progress, waits until all of the futures resolved.
+
+        .. note::
+
+            This method blocks until all futures of this batch resolved.
+        """
+        if self._status in (base.BatchStatus.STARTING, base.BatchStatus.IN_PROGRESS):
+            for future in self._futures:
+                future.result()
+
```
IlyaFaer: Timeout should be added maybe!?

tseaver: I would prefer adding a `timeout` argument.

pubsub/google/cloud/pubsub_v1/publisher/client.py
tseaver: `google.cloud.pubsub_v1.publisher._batch.thread.Batch` doesn't have a `wait` method.  Should we change its `commit` method to return the thread, and then wait on those threads?  If so, we also need to pass through a `timeout` there.

IlyaFaer: @tseaver, I've took one more deeper look (as Tianzi said, Java is canonical for PubSub - let's build on this). I don't see timeouts in other languages. Thus, we shouldn't use them here, for consistency

I'm adding `wait()` into `Batch` with this PR actually. I'd say returning threads makes some sense, but every concrete thread related to a concrete `Batch`, so, I believe, it's better to keep threads in `Batch` objects. Seems to me, it's easier to understand `Batch.wait()`, than `join()` on a list of stray thread objects

(looks like I've deleted my comment and yours too, sorry for that)

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -137,6 +137,17 @@ def status(self):
...
         return self._status
 
+    def wait(self):
+        """If commit in progress, waits until all of the futures resolved.
+
+        .. note::
+
+            This method blocks until all futures of this batch resolved.
+        """
+        if self._status in (base.BatchStatus.STARTING, base.BatchStatus.IN_PROGRESS):
```
pradn: I think we have to worry about races with this conditional, but not with the for-loop itself.

1) There is a race between a batch going from ACCEPTING_MESSAGES to STARTING in commit() and this check for the state being STARTING.
2) If a batch is in the state SUCCESS or ERROR, this for-loop wont run. So, wait() may return before the futures return with a result.

It would be good to take the lock to prevent these races.

How about:

should_wait = false
with self._state_lock:
            if self._status != base.BatchStatus.ACCEPTING_MESSAGES:
                    should_wait = true
if should_wait:
        for future in self._futures:
                future.result()

I would also worry about races in modifications to self._futures. However, in the way this class is set up, reads of self._futures only take place after all writes are done (aka, the class is no longer in state ACCEPTING_MESSAGES). The code I gave above maintains this invariant. This is not obvious. I wish we could document it.

What do you think?

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: I also agree that returning threads is not as clean as keeping them within the Batch class. However, I wonder if we can return a future instead that the Client can keep track of. I am unsure when we'd check the status of this future when we're not waiting (a timer?) I'd love to hear your thoughts.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -75,6 +75,9 @@ def __init__(self, client, topic, settings, autocommit=True):
         # These members are all communicated between threads; ensure that
         # any writes to them use the "state lock" to remain atomic.
         self._futures = []
+        # _futures list should remain unchanged after batch
```
pradn: Please move the new comment above line 77. Comments should be above the code they explain.

IlyaFaer: Done

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: Client.publish() already returns futures, so we expect clients to be using futures and waiting on those. But in the new Client.stop() method, we don't return a future and instead wait inside the method. We'd give the users more flexibility if we returned a future-like interface.

One option is to return the self._futures from Batch.commit(). Then, in Client.stop(), we take all of the returned futures from all the batches and then return them directly to the user to wait on. Pro: This option lets users choose when to wait and how long to do so. Cons: more complexity, exposing internal futures. 

Note that there's already a timeout for sending messages during publish. So if we don't let the user set a timeout for stop(), that timeout would still bound how long stop() takes. I don't know how many users would want to lengthen or shorten then stop() timeout beyond the publish timeout.

kamalaboulhosn: If stop returns a future, it seems like the best thing would be for it to return a single future that is only fulfilled when all of the underlying futures are fulfilled. I think returning a list of futures would be counterintuitive. The user already would be holding on to the futures for the individual publishes themselves, so there probably isn't a need to return a bunch of lower-level futures.

kamalaboulhosn: I'm fine with consistency with the other languages. Java doesn't return anything from stop. Given that on a publish, one can wait for all publishes to succeed or fail by waiting on the futures that came back from the publish calls, which is what the documentation for the Java shutdown method suggests. Therefore, returning a future on stop may be overkill. Is there something else we need stop() to do that requires it to wait on other threads or futures?

pradn: The extra thing stop() does is to immediately trigger Batch.commit(). If one were to rely on waiting for the futures returned by Client.publish(), they will have to wait until the batch send timeout to trigger Batch.commit().

Do users care about this delay? I'm unsure. I suppose Client.stop() would be convenience more than enabling a new use-case.

kamalaboulhosn: But they could call stop() and then wait on the futures returned by publish(), right? So they don't end up waiting for the timeout. I think that what stop is supposed to accomplish is to immediately send all outstanding messages and to disallow future calls to publish. Therefore, the user can know the state of all messages and shut down without having to worry about other publishes coming in in the meantime.

IlyaFaer: @kamalaboulhosn, I'd sent [comment](https://github.com/googleapis/google-cloud-python/pull/9365#discussion_r338188020), but I didn't saw [yours one](https://github.com/googleapis/google-cloud-python/pull/9365#discussion_r338179397) yet. Thus, mine is little bit overdue.

> But they could call stop() and then wait on the futures returned by publish(), right?

Makes sense. So the final proposition is to make `stop()` non-blocking, right? And let users wait for resolving the futures from `publish()`, if they want to. I'm okay with it

pradn: I'm fine with changing stop() to be non-blocking. We should be extra careful about documenting the method. It's easy to think "stop" means all requests die or that it's blocking.

IlyaFaer: Okay, pushed new commit, in which tried to implement everything, we'd been talking about

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -276,20 +284,22 @@ def publish(self, topic, data, **attrs):
...
+        be invoked prior to deleting this `Client()` object
+        in order to ensure that no pending messages are lost.
 
         .. note::
 
-            This method blocks until all futures of all
-            batches resolved.
+            This method is non-blocking. Use `Future()` objects
+            returned by `publish()` to make sure all messages
+            sent or failed to.
```
pradn: Fix

IlyaFaer:
> How about:
> to make sure all publish requests completed, either in success or error.

No problem, pushed

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -242,7 +243,14 @@ def publish(self, topic, data, **attrs):
             instance that conforms to Python Standard library's
             :class:`~concurrent.futures.Future` interface (but not an
             instance of that class).
+
+        Raises:
+            ValueError:
+                If called after publisher has been stopped
+                by a `stop()` method call.
         """
+        if self._is_stopped:
```
pradn: We should have a lock around reads/writes to self._is_stopped yea?

IlyaFaer: Hm-m. Pushed

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -133,7 +134,9 @@ def __init__(self, batch_settings=(), **kwargs):
         # The batches on the publisher client are responsible for holding
         # messages. One batch exists for each topic.
         self._batch_lock = self._batch_class.make_lock()
+        self._stoping_lock = threading.Lock()
```
pradn: nit: spelling - "stopping_lock"

IlyaFaer: Wow, yes! I was in a hurry, terribly sorry

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -274,3 +285,26 @@ def publish(self, topic, data, **attrs):
                 batch = self._batch(topic, create=True)
```
pradn: I believe the stopping lock still leaves a race in this code. Ex:

Thread 1: client.publish() starts running, checks _is_stopped and keeps going because it's False
Thread 2: client.stop() runs, sets _is_stopped to True, and calls batch.commit() on all batches
Thread 1: continues client.publish(), adds a batch, calls batch.publish(), and returns a future to the published message

Now, there's a dangling message publish that stop() has not committed.

One option is to take the _is_stopped lock for the entire duration of publish() and stop(). We'd be extending critical sections for a long time though..

Another is to 1) remove the _is_stopped lock 2) take the batching lock in stop() for the whole method 3) remove the batching lock from client._batch() 4) insert the batching lock between lines 280 and 285 in client.publish() (before batch = self._batch(topic)" and after "batch = self._batch(topic, create=True)". This option is simplest and fixes the issue. We extend the critical section only a little bit. In any case, I believe extending the critical section like this will be necessary for future ordering keys work.

What do you think?

IlyaFaer: > Another is to 1) remove the _is_stopped lock...

Sounds reasonable, though I don't think we need to take lock for entire `stop()`, as we are preventing further publishing with `_is_stopped`. If publish started, `stop()` will wait for publish completion. And if `stop()` started earlier, publish can be interrupted right after setting `_is_stopped` to `True` - there is no need to wait for commits completion.

Pushed changes.

pradn: I think we still need a batch lock in stop() because commit() will race with batch creation & publish otherwise. Consider this ordering:

T1: client.publish() runs, sees is_stopped is False, and continues.
T2: sets is_stopped to True, runs commit() on all existing batches
T1: still in publish(), adds message to new/old batch and publishes. This causes a message addition to a committed batch (this results in batch.publish() returning a None future, causing a new batch to be created and not published immediately- a dangling batch), or starts a new batch that is then published (this new batch has to wait for the commit timer to commit - a dangling batch).

IlyaFaer: Yes, I've added lock on half of `stop()`, while `publish()` is completely locked. So if `publish()` started first, it'll deal with batches, and only after that `stop()` will start to do it's job. But if `stop()` started first, it'll change `is_stopped` to `True` with lock, so further `publish()` will no go.

pradn: Ah! Sorry, I misread this change. The locking behavior is correct as is. I do think it'll be simpler if we just lock around all of client.stop(). It reduces the number of orderings one must consider while reading. It will have little performance impact since client.stop() should be seldom called. What do you think?

IlyaFaer: Makes some sense. After all, we're stopping the publishing. Pushed

pradn: Thanks!

IlyaFaer:
@pradn, thanks for leading me through
@kamalaboulhosn, well, it's now more close to what you proposed. Do you have any more comments? Or we could merge, I think

kamalaboulhosn:
> @pradn, thanks for leading me through
> @kamalaboulhosn, well, it's now more close to what you proposed. Do you have any more comments? Or we could merge, I think

Works for me, thanks!

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -266,11 +271,38 @@ def publish(self, topic, data, **attrs):
...
+            requests completed, either in success or error.
+        """
+        with self._batch_lock:
+            if self._is_stopped:
+                raise ValueError("Cannot stop a publisher already stopped.")
+
+            self._is_stopped = True
+
+            for topic in self._batches:
+                self._batches[topic].commit()
```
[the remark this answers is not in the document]
IlyaFaer: Right

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -242,12 +242,17 @@ def publish(self, topic, data, **attrs):
             instance that conforms to Python Standard library's
             :class:`~concurrent.futures.Future` interface (but not an
             instance of that class).
+
+        Raises:
+            ValueError:
+                If called after publisher has been stopped
+                by a `stop()` method call.
```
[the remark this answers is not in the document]
IlyaFaer: I'm thinking about `RuntimeError` little bit different, but Python docs says that it should be chosen if error doesn't fall in any of the other category, so, okay, let's do this

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -242,12 +242,17 @@ def publish(self, topic, data, **attrs):
             instance that conforms to Python Standard library's
             :class:`~concurrent.futures.Future` interface (but not an
             instance of that class).
+
+        Raises:
+            ValueError:
```
plamut:
```suggestion
            RuntimeError:
```
(now that we changed the code) :)

pubsub/google/cloud/pubsub_v1/publisher/client.py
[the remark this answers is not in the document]
plamut: Yeah, exactly. The case when we raise the error does not really match `ValueError`'s description,  ditto for other common error types, hence `RuntimeError`.

PR #9388, opened 2019-10-02
feat(automl): generate v1

busunkim96:


automl/docs/index.rst
```diff
@@ -1,8 +1,16 @@
...
+v1 API Reference
+------------------------
+.. toctree::
+    :maxdepth: 2
+
+    gapic/v1/api
+    gapic/v1/types
+
+v1beta1 API Reference
+----------------------
```
tseaver: Does this need language like other "old API version" blocks?  See, for instance, the [scheduler docs](https://github.com/googleapis/google-cloud-python/blob/master/scheduler/docs/index.rst).

tseaver:
@busunkim96 How are you fixing the b0rken changes to `setup.py`?  When I run `synthtool` inside the `automl/` directory, I'm still seeing that change.

busunkim96:
Is something off with the setup.py? I think the synth script is written to not copy it 

https://github.com/googleapis/google-cloud-python/blob/f068eb8be6e8f86cdec4d6e375307f1b425c1caf/automl/synth.py#L27-L34

tseaver:
@busunkim96 I mistyped:  the bad changes I'm seeing are in `docs/conf.py`.

busunkim96:
Ah, I just manually avoided committing it. :) Hopefully the monolith is released soon.

tseaver:
Looks like the fix for the requests intersphinx URL finally landed in artman today:  maybe you were using a pre-released version?

Issue #9082, opened 2019-08-22
BigQuery: Add support for sheets ranges


tswast:
GoogleSheetsOptions now has a (beta) property called `range`. This allows you to query a sheet other than the first sheet or query a specific subset of a sheet.

https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#GoogleSheetsOptions

This option applies to

- [ ] external table definitions in query jobs
- [ ] creating a table with an external table definition

tswast:
cc @shollyman @d-torres Just learned about this BQ -> Sheets range feature from Jeff. Probably need to add it to our backlog for the other langs, too.

PR #9416, opened 2019-10-07
feat(bigquery): add support for sheets ranges

Emar-Kar:
Closes: #9082 

List of changes:
- **external_config**
  - add range property;
- **test_external_config**
  - add test to cover `range`
- **new samples**
  - remove old ones from the `snippets`
  - new sample `query_external_sheets_permanent_table` + test
  - new sample `query_external_sheets_temporary_table` + test


bigquery/samples/query_external_sheets_permanent_table.py
```diff
@@ -0,0 +1,69 @@
...
+    schema = [
+        bigquery.SchemaField("name", "STRING"),
+        bigquery.SchemaField("post_abbr", "STRING"),
+    ]
+    table = bigquery.Table(dataset.table(table_id), schema=schema)
+    external_config = bigquery.ExternalConfig("GOOGLE_SHEETS")
+    # Use a shareable link or grant viewing access to the email address you
+    # used to authenticate with BigQuery (this example Sheet is public).
+    sheet_url = (
+        "https://docs.google.com/spreadsheets/"
```
tswast: Actually, let's continue using the previous sheet. https://docs.google.com/spreadsheets/d/1i_QCL-7HcSyUZmIbP9E6lO_T5u3HnpLe7dnpHaijg_E/edit?usp=sharing

I was thinking we'd want a different set of data for the range queries, but if we're still querying US States, we should just use the existing sheet.

bigquery/samples/query_external_sheets_permanent_table.py
```diff
@@ -0,0 +1,69 @@
...
+        "d/1dCG0rrY0nkJpB8t6Ko1S3tY7w9-hp0e_jsYer4LgEuA/edit?usp=sharing"
+    )
+    external_config.source_uris = [sheet_url]
+    external_config.options.skip_leading_rows = 1  # Optionally skip header row.
+    external_config.options.range = (
+        "Sheet1!A10:B30"
```
tswast: Let's use a range that is more possible to verify the correct behavior. Since the original sheet is sorted alphabetically, the range `A1:B48` should only get "Washington".

bigquery/samples/query_external_sheets_permanent_table.py
```diff
@@ -0,0 +1,69 @@
...
+    # Create a permanent table linked to the Sheets file.
+    table = client.create_table(table)  # Make an API request.
+
+    # Example query to find states starting with "W".
+    sql = 'SELECT * FROM `{}.{}` WHERE name LIKE "W%"'.format(dataset_id, table_id)
+    query_job = client.query(sql)  # Make an API request.
+
+    # Wait for the query to complete.
+    w_states = list(query_job)
+    print("There are {} states with names starting with W.".format(len(w_states)))
```
tswast: Let's update the print statement to indicate that we used a range: `... starting with W in the selected range.`

bigquery/samples/query_external_sheets_permanent_table.py
Emar-Kar: I decided to use `A20:B49` range. To show that it can be used not from the beginning of the sheet.

PR #9418, opened 2019-10-07
tests(texttospeech): create V1 system test

wangbill-google:
New test case for TTS API v1


busunkim96:
@wangbill-google Could you please also combine your changes into one PR?

texttospeech/tests/system/gapic/v1/test_system_tts_v1.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2018 Google LLC
```
busunkim96:
```suggestion
# Copyright 2019 Google LLC
```

texttospeech/tests/system/gapic/v1/test_system_tts_v1.py
```diff
@@ -0,0 +1,45 @@
...
+        audio_config = texttospeech_v1.types.AudioConfig(
+            audio_encoding=texttospeech_v1.enums.AudioEncoding.MP3
+        )
+
+        response = client.synthesize_speech(synthesis_input, voice, audio_config)
+
+        assert response.audio_content is not None
+
+    def test_list_voices(self):
+
```
busunkim96: There are a lot of empty lines throughout this file. Would you removing some of them?

busunkim96:
Looks good! I think the only outstanding item is to address Tres's comment above.

> We normally treat files inside the `/gapic` and `/proto` directory trees as autogen-only. Given that this file is manually created, I'd prefer to see this file moved up to `texttospeech/tests/system`.

wangbill-google:
@tseaver Hi, what kind of change do I still need to make?

busunkim96:
@tseaver Could you take another look at this one? It looks like a review requesting changes blocks merging.

PR #9419, opened 2019-10-07
tests(texttospeech): create V1beta1 system test

wangbill-google:
New test case for TTS API v1beta1


busunkim96:
Please add this change to the other open PR #9418

PR #9420, opened 2019-10-07
fix(logging): fix proto copy

busunkim96:
Logging's synth.py wasn't moving the protos over to the right place. It looks like the 'correct' set were getting copied into a deeper directory.

This PR also adds toctrees so the v1 and v2 docs are split.

See the generated docs [here](https://storage.googleapis.com/busun-misc-docs/html/index.html).

There's a new type [`google.cloud.logging_v2.types.BigQueryOptions`](https://storage.googleapis.com/busun-misc-docs/html/gapic/v2/types.html#google.cloud.logging_v2.types.BigQueryOptions) that the Logging team was asking for. Do I need to modify the v1 client to make it available there?


busunkim96:
Thank you for the review!

> Shouldn't the logging/google/cloud/logging_v2/proto/cloud/logging_v2/proto/ directory be removed altogether?

It's [gone](https://github.com/googleapis/google-cloud-python/tree/fea276f731597875e557fa73a51ee6ce69972ff2/logging/google/cloud/logging_v2/proto) as far as I can tell.

> Why does the PR delete logging/google/cloud/logging_v2/proto/log_entry.proto / logging/google/cloud/logging_v2/proto/log_severity.proto / logging/google/cloud/logging_v2/proto/logging.proto logging/google/cloud/logging_v2/proto/logging_config.proto / logging/google/cloud/logging_v2/proto/logging_metrics.proto? That is where they are supposed to be, I think, and we are retaining the equivalent _pb2.py files.

Re-added.

tseaver:
@busunkim96 
> >  Shouldn't the logging/google/cloud/logging_v2/proto/cloud/logging_v2/proto/ directory be removed altogether?
> 
> It's [gone](https://github.com/googleapis/google-cloud-python/tree/fea276f731597875e557fa73a51ee6ce69972ff2/logging/google/cloud/logging_v2/proto) as far as I can tell.

It still contains all the `_pb2.py` files, as well as `__init__py`.

PR #9389, opened 2019-10-02
tests(monitoring): rewrite logic for skipping VPCSC runs

busunkim96:
- [x] Rely on `GOOGLE_CLOUD_TESTS_IN_VPCSC` for running/skipping VPCSC tests
- [ ] Re-write tests to skip fewer runs? 

CC @bmccutchon

PR #9437, opened 2019-10-09
tests(monitoring): refresh VPCSC tests

steinwaywhw:
I have tested these locally but I needed to modify `noxfile` in order to pass in the correct environment variables. Let's see how it runs on pull requests.


steinwaywhw:
The VPCSC tests failed because `PROJECT_ID` is "". Do you think we should check "" in the test itself, or use `NONE` in the `noxfile`? I'm open to either option.

busunkim96:
Hi @steinwaywhw! Could you modify tests to be similar to what's in #9389? There's more info in [this doc](https://docs.google.com/document/d/1qCRRNqzKn7YnssbB1g-RTkJH11S3vmJ97JjsLWo6alA/edit?usp=sharing). 

Thank you!

steinwaywhw:
Hi Bu Sun, #9389 disables the tests when run outside of VPCSC. Is it decided that when not inside VPCSC, the tests should not be run at all? The current code has valid test cases when run outside of VPCSC. So shall I disable these tests when running outside of VPCSC, or should we work together to supply proper env variables to make it work outside of VPCSC?

busunkim96:
If a test is checking behavior dependent on being inside/outside a VPCSC that should only be run in VPCSC. If you have tests that don't require that environment please remove the `skipif`. I think both the VPCSC team and the client library team agree that we'd prefer for as many tests to run here as possible.

What sort of env vars were you looking to set?

steinwaywhw:
In the code there are two sets of tests. One set is run from inside VPCSC, one set is run from outside of VPCSC. For *both* sets, one half are sending requests to `PROJECT_ID`, the other half are sending requests to `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT`. For all tests, if any of these two env vars is missing, the test will skip. 

The failing reason is, however, unrelated. It is because the `PROJECT_ID` is set to an empty string `""`. But the check in the code is looking for `NONE`. Therefore it fails to skip the tests. The solution is:

* Option 1: Either change `noxfile` to set `PROJECT_ID: None` in the `env`, or change the code to treat empty string as `None`. 
* Option 2: Set `PROJECT_ID` to a project ID that is inside the VPCSC when running inside VPCSC perimeter. Let's just say this project id is "inside_project". When running outside of VPCSC, the `PROJECT_ID` should still be "inside_project" since the tests will send requests to "inside_project" and expecting the requests to be rejected by the VPCSC perimeter.

busunkim96:
> In the code there are two sets of tests. One set is run from inside VPCSC, one set is run from outside of VPCSC. For both sets, one half are sending requests to PROJECT_ID, the other half are sending requests to GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT. For all tests, if any of these two env vars is missing, the test will skip.

Ah sorry, I misspoke when I said "in VPCSC". What I wanted to distinguish is the CI environment/project used here by the client libraries team (no VPC or VPC service controls) and the CI environment used by the VPC-SC team (which runs the tests _inside_ a VPC). It looks to me like all the monitoring tests can only run correctly in the VPC-SC team's environment?

> The failing reason is, however, unrelated. It is because the PROJECT_ID is set to an empty string "". But the check in the code is looking for None. Therefore it fails to skip the tests.

The current noxfile sets PROJECT_ID to an empty string, but that can be removed with the modifications I'd was looking to make with #9389.

```python
IS_INSIDE_VPCSC = "GOOGLE_CLOUD_TESTS_IN_VPCSC" in os.environ

# If IS_INSIDE_VPCSC is set, these environment variables should also be set
if IS_INSIDE_VPCSC:
    PROJECT_INSIDE = os.environ["PROJECT_ID"]
    PROJECT_OUTSIDE = os.environ["GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT"]
```
Test skip is dependent on `GOOGLE_CLOUD_TESTS_IN_VPCSC` rather than two separate variables.
```python
    @pytest.mark.skipif(
        not IS_INSIDE_VPCSC,
        reason="This test must be run in VPCSC. To enable this test, set the environment variable GOOGLE_CLOUD_TESTS_IN_VPCSC to True",
    )

File: monitoring/tests/system/test_vpcsc_v3.py [hunk 1]
```diff
@@ -188,13 +156,13 @@ def test_get_group(self):
         PROJECT_OUTSIDE is None,
         reason="Missing environment variable: GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT",
     )
-    def test_list_group_members(self):
```
tseaver: ~~Why does this PR delete this test?~~ I see that it is moved below:  is there a reason for reordering the testcases here?

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -233,13 +201,13 @@ def test_update_group(self):
         PROJECT_OUTSIDE is None,
         reason="Missing environment variable: GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT",
     )
-    def test_create_metric_descriptor(self):
-        client = monitoring_v3.MetricServiceClient()
-        name_inside = client.project_path(PROJECT_INSIDE)
-        delayed_inside = lambda: client.create_metric_descriptor(name_inside, {})
-        name_outside = client.project_path(PROJECT_OUTSIDE)
-        delayed_outside = lambda: client.create_metric_descriptor(name_outside, {})
-        TestVPCServiceControlV3._do_test(delayed_inside, delayed_outside)
```
tseaver: ~~Why does this PR delete this test?~~ I see that it is moved below:  is there a reason for reordering the testcases here?

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -343,26 +311,7 @@ def test_list_monitored_resource_descriptors(self):
...
-        client = monitoring_v3.MetricServiceClient()
-        name_inside = client.project_path(PROJECT_INSIDE)
-        delayed_inside = lambda: client.list_time_series(
-            name_inside, "", {}, enums.ListTimeSeriesRequest.TimeSeriesView.FULL
-        )
-        name_outside = client.project_path(PROJECT_OUTSIDE)
-        delayed_outside = lambda: client.list_time_series(
-            name_outside, "", {}, enums.ListTimeSeriesRequest.TimeSeriesView.FULL
-        )
-        TestVPCServiceControlV3._do_test(delayed_inside, delayed_outside)
```
tseaver: ~~Why does this PR delete this test?~~ I see that it is moved below:  is there a reason for reordering the testcases here?

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -386,17 +335,13 @@ def test_create_notification_channel(self):
...
-        client = monitoring_v3.NotificationChannelServiceClient()
-        name_inside = client.notification_channel_path(
-            PROJECT_INSIDE, "mock_notification_channel"
-        )
-        delayed_inside = lambda: client.delete_notification_channel(name_inside)
-        name_outside = client.notification_channel_path(
-            PROJECT_OUTSIDE, "mock_notification_channel"
-        )
-        delayed_outside = lambda: client.delete_notification_channel(name_outside)
-        TestVPCServiceControlV3._do_test(delayed_inside, delayed_outside)
```
tseaver: ~~Why does this PR delete this test?~~ I see that it is moved below:  is there a reason for reordering the testcases here?

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -455,7 +400,102 @@ def test_list_notification_channel_descriptors(self):
...
+        delayed_inside = lambda: client.get_notification_channel_verification_code(
+            name_inside
+        )
+        name_outside = client.notification_channel_path(
+            PROJECT_OUTSIDE, "mock_notification_channel"
+        )
+        delayed_outside = lambda: client.get_notification_channel_verification_code(
+            name_outside
+        )
+        self._do_test(delayed_inside, delayed_outside)
```
tseaver: OK, so this one is a new test.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -470,7 +510,30 @@ def test_list_notification_channels(self):
...
+        delayed_inside = lambda: client.send_notification_channel_verification_code(
+            name_inside
+        )
+        name_outside = client.notification_channel_path(
+            PROJECT_OUTSIDE, "mock_notification_channel"
+        )
+        delayed_outside = lambda: client.send_notification_channel_verification_code(
+            name_outside
+        )
+        self._do_test(delayed_inside, delayed_outside)
```
tseaver: OK, so this one is a new test.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -502,13 +565,17 @@ def test_update_notification_channel(self):
         PROJECT_OUTSIDE is None,
         reason="Missing environment variable: GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT",
     )
-    def test_create_uptime_check_config(self):
-        client = monitoring_v3.UptimeCheckServiceClient()
-        name_inside = client.project_path(PROJECT_INSIDE)
-        delayed_inside = lambda: client.create_uptime_check_config(name_inside, {})
-        name_outside = client.project_path(PROJECT_OUTSIDE)
-        delayed_outside = lambda: client.create_uptime_check_config(name_outside, {})
-        TestVPCServiceControlV3._do_test(delayed_inside, delayed_outside)
```
tseaver: ~~Why does this PR delete this test?~~ I see that it is moved below:  is there a reason for reordering the testcases here?

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -1,5 +1,10 @@
 # -*- coding: utf-8 -*-
 #
+# flake8: noqa
+#
+# DO NOT MODIFY! THIS FILE IS AUTO-GENERATED.
+# This file is auto-generated on 09 Oct 19 18:09 UTC.
+
```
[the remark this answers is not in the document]
busunkim96: I've been told that the monitoring team has their own script to generate these tests. @steinwaywhw Could this comment be more specific about the source?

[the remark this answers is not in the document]
steinwaywhw: Hi @tseaver, this file is auto-generated from service protobufs using a script that the monitoring team developed.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -14,22 +19,13 @@
...
 import os
 import pytest
-
 from google.api_core import exceptions
 from google.cloud import monitoring_v3
 from google.cloud.monitoring_v3 import enums
 
 PROJECT_INSIDE = os.environ.get("PROJECT_ID", None)
-if not PROJECT_INSIDE:
-    PROJECT_INSIDE = None
```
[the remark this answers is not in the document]
steinwaywhw: I will update the script to treat empty string as None.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -47,8 +43,10 @@ def _is_rejected(call):
             # instance, or None.
             list(responses)
         except exceptions.PermissionDenied as e:
+            print(e)
```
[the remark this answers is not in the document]
steinwaywhw: Sure. Will do.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -119,22 +117,7 @@ def test_list_alert_policies(self):
...
-        reason="Missing environment variable: GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT",
-    )
-    def test_update_alert_policy(self):
-        client = monitoring_v3.AlertPolicyServiceClient()
-        name_inside = client.alert_policy_path(PROJECT_INSIDE, "mock_alert_policy")
-        delayed_inside = lambda: client.update_alert_policy({"name": name_inside})
-        name_outside = client.alert_policy_path(PROJECT_OUTSIDE, "mock_alert_policy")
-        delayed_outside = lambda: client.update_alert_policy({"name": name_outside})
-        TestVPCServiceControlV3._do_test(delayed_inside, delayed_outside)
+        self._do_test(delayed_inside, delayed_outside)
```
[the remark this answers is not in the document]
steinwaywhw: No particular reason. It is just because the file is auto-generated. I can sort the input endpoints to the script and make it deterministic from now on.

steinwaywhw:
> It looks to me like all the monitoring tests can only run correctly in the VPC-SC team's environment?

Hi @busunkim96, the test is able to run from any environment as long as you set `PROJECT_ID` to some random project within some random VPCSC perimeter, and `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT` to some random project outside of any VPCSC perimeter. Indeed I ran the tests on my own machine before the PR using `PROJECT_ID` = `secure_gcp_test_project_4` and `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT` my random google cloud project. 

This all goes back to the previous comment:

> In the code there are two sets of tests. One set is run from inside VPCSC, one set is run from outside of VPCSC. For both sets, one half are sending requests to `PROJECT_ID`, the other half are sending requests to `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT`. For all tests, if any of these two env vars is missing, the test will skip.

The first set of tests mentioned above has to be run in VPCSC teams dedicated environment. The second set _could_ be run anywhere, including the pull request checks environment that your team owns, _as long as_ you supply the needed `PROJECT_ID` and `GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT` to the tests.

So my question is, do you intend to run these tests in the PR presubmit checks? 

I feel like we were not on the same page. I hope this comment clarifies things. Feel free to find me on chat if you have any questions or concerns.

steinwaywhw:
Sorry, I'm not sure why the diff page does not know it is a git mv even though I did it ...

monitoring/tests/system/test_vpcsc_v3.py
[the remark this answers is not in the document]
steinwaywhw: I have changed the "skipIf" to "not PROJECT_INSIDE" and "not PROJECT_OUTSIDE" correspondingly, so we don't need this two lines anymore.

monitoring/tests/system/test_vpcsc_v3.py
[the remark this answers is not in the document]
steinwaywhw: Changed to `logger.debug(e)`.

monitoring/tests/system/test_vpcsc_v3.py
steinwaywhw: I've changed the script to arrange tests in a deterministic order by sorting endpoint versions, then services, then methods. Should be good now.

busunkim96:
Hi Hanwen - I looked at the project and documentation you mentioned in chat. As the tests are written now, I don't think there's much to gain from running them in our presubmits. The VPC-SC team is already running the tests with projects in VPC, so there's no need to duplicate the same setup here. 

If some tests made valid requests to the backend, I would be open to running them here since they would make the tests more robust. But since these tests are exclusively checking that a request is forbidden/permitted via the status code, they wouldn't be able to catch errors that aren't specific to VPC-SC.

Also happy to grab some time next week to chat over GVC if you have other thoughts.

Thanks!

steinwaywhw:
That's ok to me. The code as it is right now in this PR will skip all tests since the CI will pass PROJECT_ID = "". So it is safe to merge them now.

steinwaywhw:
*friendly ping*

steinwaywhw:
Thank you, Bu Sun!

PR #9450, opened 2019-10-10
fix(storage): enable CSEK w/ V4 signed URLs

### LINKED ISSUE #7626: Storage: add systests for V4 signed URLs with CSEK headers
/cc @frankyn Follow-on from PR #7460.

Currently, the GCS back-end has a bug for signed URLs using customer-supplied encryption key headers (`x-goog-encryption-key` and `x-goog-encryption-key-sha256`).  Once a fix for that bug is rolled out, add system tests which show using the CSEK headers to build V4 signed upload / download URLs, and exercise them.

tseaver:
Closes #7626

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -469,9 +469,15 @@ class Blob(_PropertyMixin):
             helper = generate_signed_url_v4
 
         if self._encryption_key is not None:
+            encryption_headers = _get_encryption_headers(self._encryption_key)
             if headers is None:
                 headers = {}
-            headers.update(_get_encryption_headers(self._encryption_key))
+            if version == "v2":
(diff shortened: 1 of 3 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -468,6 +468,11 @@ def generate_signed_url(
         else:
             helper = generate_signed_url_v4
 
+        if self._encryption_key is not None:
```
frankyn: this should only be done if v4.

tseaver: Per chat, we still need to add the `X-Goog-Encryption-Algorithm` header for V2 (but not the key or hash, oddly enough).  See: https://cloud.google.com/storage/docs/access-control/signed-urls-v2#about-canonical-extension-headers

Issue #9392, opened 2019-10-04
spanner: Session pool should hand out sessions in LIFO order


skuruppu:
**Is your feature request related to a problem? Please describe.**
Currently sessions are handed out from the session pool in FIFO order. Least recently used sessions are unlikely to be in backend caches, and the transactions associated with them are likely to aborted due to idleness.

**Describe the solution you'd like**
The session pool should hand out sessions in LIFO order.

tseaver:
For the `FixedSizePoot` and `BurstyPool` classes, this change is straightforward:  we just use `queue.LifoQueue` instead of `queue.Queue` for the sessions.

For the `PingingPool` and `TransacctionPingingPool` class, the issue isn't quite clear:  sessions are held in a priority queue based on when they need to be "pinged" (checked for existence at intervals).

PR #9454, opened 2019-10-11
fix(spanner): return sessions from pool in LIFO order

IlyaFaer:
Change sessions order from FIFO to LIFO. Done only in those pool classes, which don't have priorities.
Towards #9392

### DIFF
--- a/spanner/tests/unit/test_pool.py
+++ b/spanner/tests/unit/test_pool.py
@@ -162,15 +162,16 @@ class TestFixedSizePool(unittest.TestCase):
     def test_get_non_expired(self):
         pool = self._make_one(size=4)
         database = _Database("name")
-        SESSIONS = [_Session(database)] * 4
+        SESSIONS = sorted([_Session(database) for i in range(0, 4)])
         database._sessions.extend(SESSIONS)
         pool.bind(database)
 
-        session = pool.get()
-
-        self.assertIs(session, SESSIONS[0])
-        self.assertTrue(session._exists_checked)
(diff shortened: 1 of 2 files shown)


spanner/tests/unit/test_pool.py
```diff
@@ -162,15 +162,16 @@ def test_bind(self):
     def test_get_non_expired(self):
         pool = self._make_one(size=4)
         database = _Database("name")
-        SESSIONS = [_Session(database)] * 4
+        SESSIONS = sorted([_Session(database) for i in range(0, 4)])
```
IlyaFaer: Creating different session objects instead of copying the single one - needed for assertion

spanner/tests/unit/test_pool.py
```diff
@@ -875,7 +876,7 @@ def __init__(self, name):
         self._sessions = []
 
     def session(self):
-        return self._sessions.pop()
+        return self._sessions.pop(0)
```
IlyaFaer: `pop()` without args returns last element. So within tests list of sessions become reversed while putting it into pool. With this we would had to check, if first object from pool equals to first object from original sessions list, which looks like FIFO (it's not, 'cause original list would have been reversed twice before returning from pool, but it's not very obvious while looking at a test code). So it's easier to reverse order in mock object, and make test code more simple for understanding

larkee: Perhaps add a comment in the code mentioning this.

IlyaFaer: Okay, pushed

PR #9464, opened 2019-10-14
feat(bigquery): add Dataset.default_partition_expiration_ms and Table.require_partition_filter properties

tswast:
Samples double as system tests. I aim to use them here:

* https://cloud.google.com/bigquery/docs/managing-partitioned-tables#require-filter
* https://cloud.google.com/bigquery/docs/updating-datasets#partition-expiration

Note: this also (silently) deprecates
TimePartitioning.require_partition_filter, as that's duplicating the
same functionality.

I was curious why the expiration_ms wasn't also moving up, but then I
realized that property only makes sense if a partition is assocatied
with a timestamp.

### DIFF
--- a/bigquery/google/cloud/bigquery/table.py
+++ b/bigquery/google/cloud/bigquery/table.py
@@ -1795,10 +1795,28 @@ class TimePartitioning(object):
         :attr:`~google.cloud.bigquery.table.Table.require_partition_filter`,
         instead.
         """
+        warnings.warn(
+            (
+                "TimePartitioning.require_partition_filter will be removed in "
+                "future versions. Please use Table.require_partition_filter "
+                "instead."
+            ),
+            PendingDeprecationWarning,
+            stacklevel=2,
+        )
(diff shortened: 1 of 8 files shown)


bigquery/google/cloud/bigquery/table.py
```diff
@@ -1722,9 +1735,9 @@ class TimePartitioning(object):
             Number of milliseconds for which to keep the storage for a
             partition.
         require_partition_filter (bool, optional):
-            If set to true, queries over the partitioned table require a
-            partition filter that can be used for partition elimination to be
-            specified.
+            DEPRECATED: Use
```
shollyman: Are there any annotations we can/should add to help docs/and existing consumers to understand this is no longer the preferred mechanism?

tswast: 👍 Added deprecation warning message in 84e2c3cca903d78e56706c5e777ff4d9b322ef79

PR #9477, opened 2019-10-15
feat(bigquery): add range partitioning to tables, load jobs, and query jobs

tswast:
These classes and properties add support for the integer range
partitioning feature. These offer more flexibility in partitioning
options than time-based partitioning.

Note: integer range partitioning is currently only available under whitelist.

### DIFF
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -1183,8 +1183,12 @@ class LoadJobConfig(_JobConfig):
 
     @property
     def range_partitioning(self):
-        """Optional[google.cloud.bigquery.table.RangePartitioning]: Configures range-based
-        partitioning for destination table.
+        """Optional[google.cloud.bigquery.table.RangePartitioning]:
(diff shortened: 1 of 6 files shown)


tswast:
FYI: I've started a draft PR at https://github.com/googleapis/google-cloud-python/pull/9478 with code samples (doubling as system tests). I figure we can get this PR committed and released and follow-up with samples once the feature is no longer restricted access.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -27,7 +27,9 @@
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetListItem
 from google.cloud.bigquery.dataset import DatasetReference
+from google.cloud.bigquery.encryption_configuration import EncryptionConfiguration
```
shollyman: cleanup from another PR?

tswast: Yeah, recently added for models encryption but I forgot to ask for the imports to be alphabetized.

PR #9478, opened 2019-10-15
doc(bigquery): add table create sample using integer range partitioning

tswast:
This code sample doubles as a system test for integer range partitioning features. Follow-up to https://github.com/googleapis/google-cloud-python/pull/9477


plamut:
@shollyman Is there anything that you might want to add here before merging?

Thanks!

Issue #9092, opened 2019-08-23
BigQuery: Convert Sphinx-style docstrings to Google style


plamut:
Following the [comment thread](https://github.com/googleapis/google-cloud-python/pull/9067#discussion_r317214709) on another PR, it turned out that `job.py`, and possibly other modules as well, still use the Sphinx-style docstrings.

These docstrings should be converted to Google-style docstrings as the Python [coding style guideline](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) mandates.

Examples of Google-style docstrings:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

Patil2099:
@plamut @IlyaFaer  Can I work on this?

IlyaFaer:
@Patil2099, I've already wrote raw script for massive docs updating. [Pushed](https://github.com/q-logic/google-cloud-python/pull/49) internally just a few minutes ago - was gonna show to @tswast and ask about how we should deal with this changes, and maybe there is something to do "by the way"

IlyaFaer:
@Patil2099, I've already wrote raw script for massive docs updating. Pushed internally just a few minutes ago - was gonna show to @tswast and ask about how we should deal with this changes, and maybe there is something to do "by the way"

Patil2099:
@iilyafaer Maybe I can help in that .

plamut:
@IlyaFaer I've already done quite some work on this in BigQuery, but if we are going to use a mass update script, I'll put that aside. Thanks for mentioning the script here!

@Patil2099 I am fine with you working on this, just please coordinate with @IlyaFaer to not unnecessarily repeat any work.

IlyaFaer:
Had a word with @tswast. There are no more old-styled comments in `bigquery/google/cloud` folder. I've also checked `bigquery_datatransfer`, `bigquery_storage` and `datacatalog`, they are Google-styled. So, I assume, we should close the issue

plamut:
@IlyaFaer Indeed, the issue can be closed now. Thanks for the cleanup!

PR #9326, opened 2019-09-27
refactor(bigquery): rewrite docs in Google style, part 1

IlyaFaer:
Towards #9092. First part of comments update for BigQuery

PR #9481, opened 2019-10-16
refactor(bigquery): rewrite docs in Google style, part 2

IlyaFaer:
towards issue #9092

### DIFF
--- a/bigquery/google/cloud/bigquery/_helpers.py
+++ b/bigquery/google/cloud/bigquery/_helpers.py
@@ -90,12 +90,15 @@ def _timestamp_query_param_from_json(value, field):
 
     Args:
         value (str): The timestamp.
-        field (.SchemaField): The field corresponding to the value.
+
+        field (google.cloud.bigquery.schema.SchemaField):
+            The field corresponding to the value.
 
     Returns:
-        Optional[datetime.datetime]: The parsed datetime object from
-        ``value`` if the ``field`` is not null (otherwise it is
-        :data:`None`).
+        Optional[datetime.datetime]:
+            The parsed datetime object from
+            ``value`` if the ``field`` is not null (otherwise it is
+            :data:`None`).
     """
     if _not_null(value, field):
         # Canonical formats for timestamps in BigQuery are flexible. See:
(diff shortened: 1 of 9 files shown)


IlyaFaer:
This PR contains mostly auto generated (by my self-written script) changes, similar to [PR](https://github.com/googleapis/google-cloud-python/pull/9326), and a small peace of manual work

core/google/cloud/_http.py
```diff
@@ -313,59 +304,59 @@ def api_request(
...
+                If True, this method will try to parse the
+                response as JSON and raise an exception if
+                that cannot be done.  Default is True.
+
+            _target_object (object):
+                (Optional) Protected argument to be used by library callers. This
+                can allow custom behavior, for example, to defer an HTTP request
+                and complete initialization of the object at a later time.
 
         :raises ~google.cloud.exceptions.GoogleCloudError: if the response code
```
software-dov: This should be changed to 

Raises:
    ~.google.cloud.exceptions.GoogleCloudError: ...

IlyaFaer: Oh, yes, there is a mistake in this concrete comment (no end colon in tag "raises"), so script skipped it. In fact, this file really slapped out of my radars, I wasn't going to push files outside BigQuery directory. Will revert it and push later in PR for Core itself

bigquery/google/cloud/bigquery/client.py
```diff
@@ -349,10 +347,10 @@ def create_dataset(self, dataset, exists_ok=False, retry=DEFAULT_RETRY):
         https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/insert
 
         Args:
-            dataset (Union[ \
```
tswast: How do the docs render for this change? Run `nox -s docs` and then look at the generated docs in `docs\_build\html`.

I recall previously Sphinx needed the ` \` continuation markers to render this properly.

IlyaFaer: Hm, I didn't attach importance to this, 'cause this slashes are not used all over the place. But looks like you right, deleting them can cause marking problems. Pushed a fix

bigquery/google/cloud/bigquery/_helpers.py
```diff
@@ -90,12 +90,15 @@ def _timestamp_query_param_from_json(value, field):
 
     Args:
         value (str): The timestamp.
-        field (.SchemaField): The field corresponding to the value.
+
+        field (google.cloud.bigquery.schema.SchemaField):
```
plamut: If I'm not too late to the party - it would make sense to cross-link classes in the docs (applies to multiple places).

```suggestion
        field (:class:`~google.cloud.bigquery.schema.SchemaField`):
```

IlyaFaer: Well, in fact, Tim had been observing the process of writting the script (which made this changes), and it was decided to remove `:class:` tags from type definitions, but to keep them in arg descriptions
Безымянный

plamut: I see, thanks, didn't know that.

PR #9486, opened 2019-10-16
feat(api_core): provide a 'raw_page' field for page_iterator.Page

software-dov:
Some paginated response messages include additional fields that users
may wish to inspect. This change stores the raw page when constructing a page_iterator.Page and provides an accessor.

In addition to the new and modified unit tests, hand experimentation generates expected behavior:
```python
from google.cloud import automl
client = automl.AutoMlClient()

result = client.list_datasets(...)
for page in result.pages:
    print(f"{page.raw_page}")
```
prints individual pages.

### DIFF
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -11,5 +11,3 @@ exclude_lines =
     def __repr__
     # Ignore abstract methods
     raise NotImplementedError
-    # Ignore assert False, meant to prevent false positive tests
-    assert False
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -13,3 +13,4 @@ exclude_lines =
     raise NotImplementedError
     # Ignore assert False, meant to prevent false positive tests
     assert False
+    
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -11,3 +11,5 @@ exclude_lines =
     def __repr__
     # Ignore abstract methods
     raise NotImplementedError
+    # Ignore assert False, meant to prevent false positive tests
+    assert False
--- a/api_core/.coveragerc
+++ b/api_core/.coveragerc
@@ -13,4 +13,3 @@ exclude_lines =
     raise NotImplementedError
     # Ignore assert False, meant to prevent false positive tests
     assert False
-    
--- a/api_core/google/api_core/page_iterator.py
+++ b/api_core/google/api_core/page_iterator.py
@@ -100,7 +100,7 @@ class Page(object):
             The raw page response.
     """
 
-    def __init__(self, parent, items, item_to_value, raw_page):
+    def __init__(self, parent, items, item_to_value, raw_page=None):
         self._parent = parent
         self._num_items = len(items)
         self._remaining = self._num_items
@@ -368,7 +368,7 @@ class HTTPIterator(Iterator):
         if self._has_next_page():
             response = self._get_next_page_response()
             items = response.get(self._items_key, ())
-            page = Page(self, items, self.item_to_value, response)
+            page = Page(self, items, self.item_to_value, raw_page=response)
             self._page_start(self, page, response)
             self.next_page_token = response.get(self._next_token)
             return page
@@ -462,7 +462,7 @@ class _GAXIterator(Iterator):
         """
         try:
             items = six.next(self._gax_page_iter)
-            page = Page(self, items, self.item_to_value, None)
+            page = Page(self, items, self.item_to_value)
             self.next_page_token = self._gax_page_iter.page_token or None
             return page
         except StopIteration:
@@ -535,7 +535,7 @@ class GRPCIterator(Iterator):
 
         self.next_page_token = getattr(response, self._response_token_field)
         items = getattr(response, self._items_field)
-        page = Page(self, items, self.item_to_value, response)
+        page = Page(self, items, self.item_to_value, raw_page=response)
 
         return page
 
--- a/api_core/google/api_core/page_iterator.py
+++ b/api_core/google/api_core/page_iterator.py
@@ -96,14 +96,22 @@ class Page(object):
             Callable to convert an item from the type in the raw API response
             into the native object. Will be called with the iterator and a
             single item.
+        raw_page Optional[google.protobuf.message.Message]:
+            The raw page response.
     """
 
-    def __init__(self, parent, items, item_to_value):
+    def __init__(self, parent, items, item_to_value, raw_page):
         self._parent = parent
         self._num_items = len(items)
         self._remaining = self._num_items
         self._item_iter = iter(items)
         self._item_to_value = item_to_value
+        self._raw_page = raw_page
+
+    @property
+    def raw_page(self):
+        """google.protobuf.message.Message"""
+        return self._raw_page
 
     @property
     def num_items(self):
@@ -360,7 +368,7 @@ class HTTPIterator(Iterator):
         if self._has_next_page():
             response = self._get_next_page_response()
             items = response.get(self._items_key, ())
-            page = Page(self, items, self.item_to_value)
+            page = Page(self, items, self.item_to_value, response)
             self._page_start(self, page, response)
             self.next_page_token = response.get(self._next_token)
             return page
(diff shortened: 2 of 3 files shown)


crwilcox:
Thanks @software-dov, any chance that this would be doable off of `Iterator` instead of `Page` so it could be used like:

```
response = client.list_resources(request)
handle_raw_response(p.raw_response)
```

software-dov:
My two main concerns in no particular order are
1) As few differences from the surface exposed by the microgenerator as possible.
    If breaks are inevitable, prioritize breaks that don't require drastic changes in the microgenerator.
2) Ergonomic user interface

With "ease of implementation" being a non-voting tie-breaker.

Making the raw page (or transparently mapped attributes via __getattr__) visible from `Iterator` matches 1 with the minor caveat that 'page_token' isn't always a flattened parameter in either surface. I'm not entirely happy with the resulting interface, however.

With the microgenerator surface, touching internal fields of the page forces you to do your own pagination, something like
```python
while True:
    response = client.paginated_method(request)
    handle_page(response)
    if not response.next_page_token:
        break
    request.page_token = response.next_page_token
```
which doesn't look great to me. Again, modulo complications with flattening, this is the interface that exposing raw_page in the iterator would be forced to mimic.

I guess I'm not sure I understand the circumstances where the topmost level raw_page is preferable to
```python
response = client.paginated_method(request)
for p in response.pages:
    handle_page(p)
    for elt in p:
        handle_element(elt)
```

software-dov:
Open review and design bump

api_core/google/api_core/page_iterator.py
```diff
@@ -454,7 +462,7 @@ def _next_page(self):
         """
         try:
             items = six.next(self._gax_page_iter)
-            page = Page(self, items, self.item_to_value)
+            page = Page(self, items, self.item_to_value, None)
```
tseaver: Likewise here:
```suggestion
            page = Page(self, items, self.item_to_value, raw_page=None)
```

api_core/tests/unit/test_page_iterator.py
```diff
@@ -30,15 +30,15 @@ def test_constructor(self):
...
+        page = page_iterator.Page(parent, (1, 2, 3), item_to_value, None)
 
         assert page.num_items == 3
         assert page.remaining == 3
         assert page._parent is parent
         assert page._item_to_value is item_to_value
 
     def test___iter__(self):
-        page = page_iterator.Page(None, (), None)
+        page = page_iterator.Page(None, (), None, None)
```
tseaver: Since we don't care about the `raw_page` in this test, we could just omit it (assuming we make it optional) here and below.

api_core/tests/unit/test_page_iterator.py
```diff
@@ -69,6 +69,22 @@ def test_iterator_calls_parent_item_to_value(self):
...
+
+        page = page_iterator.Page(parent, (1, 2, 3), item_to_value, raw_page)
+        assert page.raw_page is raw_page
+
+        try:
+            page.raw_page = None
+        except AttributeError:
+            pass
+        else:
+            assert False, "The raw_page attribute should be a read-only property"
```
tseaver: Rather than the `assert False`, use `assertRaises`, e.g.:

```python
        with self.assertRaises(AttributeError):
            page.raw_page = None
```

api_core/google/api_core/page_iterator.py
```diff
@@ -96,14 +96,22 @@ class Page(object):
             Callable to convert an item from the type in the raw API response
             into the native object. Will be called with the iterator and a
             single item.
+        raw_page Optional[google.protobuf.message.Message]:
+            The raw page response.
     """
 
-    def __init__(self, parent, items, item_to_value):
+    def __init__(self, parent, items, item_to_value, raw_page):
```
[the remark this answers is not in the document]
software-dov: Do the manually written clients create `Page` objects directly? If so, then making `raw_page` optional only papers over the problem, and issues will need to be filed for each lib.

My impetus for making it a positional parameter is that I couldn't think of a good reason to make it optional. The user doesn't care about whether the code that returns a `Page` was generated or handwritten, they just expect a consistent interface. Making `raw_page` positional forces code that constructs a `Page` think before opting out.

api_core/google/api_core/page_iterator.py
```diff
@@ -360,7 +368,7 @@ def _next_page(self):
         if self._has_next_page():
             response = self._get_next_page_response()
             items = response.get(self._items_key, ())
-            page = Page(self, items, self.item_to_value)
+            page = Page(self, items, self.item_to_value, response)
```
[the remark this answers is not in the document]
software-dov: Given making `raw_page` optional the rest of these changes are all good calls.

api_core/tests/unit/test_page_iterator.py
software-dov: Doh! Don't know what I was thinking.

api_core/google/api_core/page_iterator.py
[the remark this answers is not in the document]
tseaver: The `Page` class is part of the public API here, which means that we don't know all the code which might construct instances of it.

api_core/tests/unit/test_page_iterator.py
```diff
@@ -30,15 +30,15 @@ def test_constructor(self):
         parent = mock.sentinel.parent
         item_to_value = mock.sentinel.item_to_value
 
-        page = page_iterator.Page(parent, (1, 2, 3), item_to_value)
+        page = page_iterator.Page(parent, (1, 2, 3), item_to_value, None)
```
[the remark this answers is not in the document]
software-dov: Done

api_core/tests/unit/test_page_iterator.py
```diff
@@ -69,6 +70,18 @@ def test_iterator_calls_parent_item_to_value(self):
...
+    def test_raw_page(self):
+        parent = mock.sentinel.parent
+        item_to_value = mock.sentinel.item_to_value
+
+        raw_page = mock.sentinel.raw_page
+
+        page = page_iterator.Page(parent, (1, 2, 3), item_to_value, raw_page=raw_page)
+        assert page.raw_page is raw_page
+
+        self.assertRaises(AttributeError):
```
tseaver:
```suggestion
        with self.assertRaises(AttributeError):
```

tseaver: @software-dov I don't know why, but I'm not allowed to commit that change.

software-dov: Woah, just realized what I did here. This slipped through because I gave up trying to run unit tests locally after spending two hours fighting with virtualenv and figured CI would catch any errors. Fixed.

tseaver: We normally run tests using [`nox`](http://nox.thea.codes/en/stable/), which handles setting up the virtualenvs correctly for each kind of test.

api_core/tests/unit/test_page_iterator.py
```diff
@@ -69,6 +70,18 @@ def test_iterator_calls_parent_item_to_value(self):
...
+    def test_raw_page(self):
+        parent = mock.sentinel.parent
+        item_to_value = mock.sentinel.item_to_value
+
+        raw_page = mock.sentinel.raw_page
+
+        page = page_iterator.Page(parent, (1, 2, 3), item_to_value, raw_page=raw_page)
+        assert page.raw_page is raw_page
+
+        with self.assertRaises(AttributeError):
```
busunkim96: `pytest.raises`? https://docs.pytest.org/en/latest/getting-started.html#assert-that-a-certain-exception-is-raised
```suggestion
        with pytest.raises(AttributeError):
```

You can run the tests locally by installing nox.

```pip install nox```

and run the unit test session from the `api_core/` directory.

```nox -s unit```

This will handle setting up the virtualenv and installing dependencies.

File: api_core/.coveragerc [hunk 9]
```diff
@@ -11,3 +11,5 @@ exclude_lines =
     def __repr__
     # Ignore abstract methods
     raise NotImplementedError
+    # Ignore assert False, meant to prevent false positive tests
+    assert False
```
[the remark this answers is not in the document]
software-dov: It does not, will remove.

api_core/tests/unit/test_page_iterator.py
software-dov: Thanks! I was previously using `python -m nox -r sessions unit-3.6`, which was choking on import and environment.

software-dov: Fixed, too.

PR #9494, opened 2019-10-18
fix(spanner): pass gRPC config options through to gRPC channel creation

### LINKED ISSUE #9301: Spanner: gRPC config not being passed to 'create_channel'
The refactor in commit https://github.com/googleapis/google-cloud-python/commit/1841c7af9c8ae22722ee63581cf1ee33d3e96661, resulted in an [unused variable](https://github.com/googleapis/google-cloud-python/blob/1841c7af9c8ae22722ee63581cf1ee33d3e96661/spanner/google/cloud/spanner_v1/gapic/transports/spanner_grpc_transport.py#L99).

Previously, options was used as a keyword argument in the google.api_core.grpc_helpers.create_channel call so I think it has just omitted. This is the only reference to grpc config I could find so this probably means the grpc channel is not being configured properly.

larkee:
Fixes the replacement string in the synth file so that the options being created from the gRPC config file are actually being used when creating a channel for the SpannerClient.
Closes #9301


larkee:
@tseaver PTAL :)

PR #9506, opened 2019-10-21
feat(bigquery): add TypeError if wrong job_config type is passed to client job methods

### LINKED ISSUE #9480: BigQuery: raise ValueError if wrong job_config type is passed to client job methods (load_table_from_*, query, ...)

It's a somewhat common problem to pass in a `QueryJobConfig` when you meant to create a `LoadJobConfig`. (I've done it!) This results in a difficult-to-debug error from the backend:

```
google.api_core.exceptions.BadRequest: 400 POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/swast-scratch/jobs?uploadType=resumable: Required parameter is missing
```

Instead, let's catch when someone passes in the wrong type to a `job_config` parameter and raise a `ValueError` with a more helpful message like: `Expected a LoadJobConfig for the job_config parameter, but got job_config=....`

- Continue to let the server catch these errors. The error message is unfortunately very difficult to debug, though, because we are sending a job resource of ambiguous type to the API.
- Fail somewhere else, such as when we create the resource to send to the API. This is more error-prone and gives us less ability to catch the error at the line that actually caused it.

 **Additional context**

In response to internal issue 121343256.

HemangChothani:
Fixes #9480

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1355,6 +1355,11 @@ class Client(ClientWithProject):
 
         Returns:
             google.cloud.bigquery.job.LoadJob: A new load job.
+
+        Raises:
+            ValueError:
+                If ``job_config`` is not an instance of ``google.cloud.bigquery.job.LoadJobConfig``
+                class.
         """
         job_id = _make_job_id(job_id, job_id_prefix)
 
@@ -1370,6 +1375,14 @@ class Client(ClientWithProject):
             source_uris = [source_uris]
 
         destination = _table_arg_to_table_ref(destination, default_project=self.project)
+
+        if job_config:
+            if not isinstance(job_config, google.cloud.bigquery.job.LoadJobConfig):
+                raise ValueError(
+                    "Expected an instance of LoadJobConfig class for the job_config parameter, but received job_config = %s"
+                    % (job_config,)
+                )
+
         load_job = job.LoadJob(job_ref, source_uris, destination, self, job_config)
         load_job._begin(retry=retry)
 
@@ -1436,6 +1449,9 @@ class Client(ClientWithProject):
                 If ``size`` is not passed in and can not be determined, or if
                 the ``file_obj`` can be detected to be a file opened in text
                 mode.
+
+            ValueError:
+                If ``job_config`` is not an instance of ``google.cloud.bigquery.job.LoadJobConfig`` class.
         """
         job_id = _make_job_id(job_id, job_id_prefix)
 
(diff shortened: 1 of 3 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1355,6 +1355,11 @@ def load_table_from_uri(
 
         Returns:
             google.cloud.bigquery.job.LoadJob: A new load job.
+
+        Raises:
+            ValueError:
+                If ``job_config`` is not an instance of ``google.cloud.bigquery.job.LoadJobConfig``
```
plamut: IMO it would be convenient to cross-link the expected class in the docs (applies to multiple places):

```suggestion
                If ``job_config`` is not an instance of :class:`~google.cloud.bigquery.job.LoadJobConfig`
```

bigquery/google/cloud/bigquery/client.py
```diff
@@ -2060,6 +2131,12 @@ def query(
 
         if self._default_query_job_config:
             if job_config:
+                if not isinstance(job_config, google.cloud.bigquery.job.QueryJobConfig):
+                    raise ValueError(
+                        "Expected an instance of QueryJobConfig class for the job_config parameter, but received job_config = {}".format(
+                            job_config
+                        )
+                    )
```
plamut: This check and the error message are repeated at quite a few places in the file, and it would be beneficial to extract it to a helper function, making the code more maintainable.

Something like the following:

```py
def _verify_job_config_type(job_config, expected_type, param_name="job_config", allow_none=True):
    if job_config is None and allow_none:
        return

    if not isinstance(job_config, expected_type):
        msg = (
            "Expected an instance of {expected_type} class for the {param_name} parameter, "
            "but received {param_name} = {job_config}"
        )
        raise TypeError(
            msg.format(
                 expected_type=expected_type.__class__.__name__,
                 param_name=param_name,
                 job_config=job_config,
             )
        )
```

Also, raising a `TypeError` is more appropriate than `ValueError`, as the latter's description is _"Inappropriate argument value (of correct type)."_.

bigquery/tests/unit/test_client.py
```diff
@@ -3100,6 +3104,24 @@ def test_load_table_from_uri_w_client_location(self):
...
         )
 
+    def test_load_table_from_uri_w_invalid_job_config(self):
+        JOB = "job_name"
+        DESTINATION = "destination_table"
+        SOURCE_URI = "http://example.com/source.csv"
+
+        creds = _make_credentials()
+        http = object()
+        job_config = object()
```
plamut: (applies to multiple places)

It would make sense to tighten the check and also test with a  `_JobConfig` instance of an incompatible job type, e.g. `QueryJobConfig`, which is more likely the error that can be expected in practice.

Users probably pass an incorrect job config type way more often than a non-`_JobConfig` instance.

PR #9507, opened 2019-10-21
chore(pubsub): add subscriber role test for streaming

### LINKED ISSUE #9382: PubSub: Add system test for streaming pull using an account with pubsub.subscriber role
A PubSub regression (#9339) was introduced in the 1.0.1, and fixed in the `1.0.2` release. The fix should be covered by a proper system test.

The test should run a simple streaming pull scenario using a service account that only has the `pubsub.subscriber` role, and no other additional permissions. This implies creating a service account, and adding it to the Kokoro environment (contact @busunkim96 for more info).

**Side question:** How to easily share the same account(s) with developers that need/want to run system tests locally?

**Resources:**
- [PubSub roles](https://cloud.google.com/pubsub/docs/access-control#roles)

plamut:
Fixes #9382.

Pulling the messages using a streaming pull should work with accounts having only the `pubsub.subscriber` role. This commits add a test that covers this aspect.

### How to test
- Apply the following patch to the streaming pull manager:
    ```patch
    diff --git pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
    index d3b1d6f51eb..d73a6f9cdd8 100644
    --- pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
    +++ pubsub/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
    @@ -403,6 +403,7 @@ class StreamingPullManager(object):
            # the default stream ACK deadline should again be set based on the
            # historic ACK timing data, i.e. `self.ack_histogram.percentile(99)`.
            stream_ack_deadline_seconds = _DEFAULT_STREAM_ACK_DEADLINE
    +       subscription = self._client.api.get_subscription(self._subscription)
    
            get_initial_request = functools.partial(
                self._get_initial_request, stream_ack_deadline_seconds
   ```
- Run the new system test `test_streaming_pull_subscriber_permissions_sufficient`.

**Expected result:**
The test should fail (permission error), proving that it would catch the regression in #9339, and undoing the above patch makes the test pass.

### DIFF
--- a/pubsub/tests/system.py
+++ b/pubsub/tests/system.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 import datetime
 import itertools
 import operator as op
+import os
 import threading
 import time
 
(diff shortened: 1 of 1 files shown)


pubsub/tests/system.py
```diff
@@ -488,6 +488,38 @@ def test_streaming_pull_max_messages(
...
+        cleanup.append((publisher.delete_topic, topic_path))
+        cleanup.append((subscriber.delete_subscription, subscription_path))
+
+        # create a topic and subscribe to it
+        publisher.create_topic(topic_path)
+        subscriber.create_subscription(subscription_path, topic_path)
+
+        # TODO: A service account granting only the pubsub.subscriber role must
+        # be used
+        filename = "/some/service/account/file.json.json"
```
plamut: @busunkim96 Let's talk offline on how to inject such service account into Kokoro environment, and how to do the same for the convenience of developers (we cannot commit the actual service account `.json` key file, of course).

busunkim96: There is now a service account with just Pub/Sub subscriber in `${KOKORO_GFILE_DIR}/pubsub-subscriber-service-account.json`

You can set an environment variable in https://github.com/googleapis/google-cloud-python/blob/master/.kokoro/build.sh

I _think_ there may actually be a way to use the existing service account to create a short-lived credential with only the Pub/Sub subscriber role. See https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-permissions and https://google-auth.readthedocs.io/en/latest/reference/google.auth.impersonated_credentials.html#module-google.auth.impersonated_credentials

This would also be nicer for developers trying to run the tests (they don't have to worry about having a separate service account with the correct role)

plamut:
The test now correctly fails with 403 if the regression from #9339 is re-introduced.

I opted to not add the path to the special service account file in `.kokoro/build.sh`, as adding more test-specific stuff to the common config could eventually bloat it.

I haven't explored the short-lived service accounts for local development yet, I got around it by skipping the test if not inside the Kokoro environment (but that could still be faked locally, if needed).

plamut:
Yay, merging then.

PR #9510, opened 2019-10-21
feat(cloudbuild): generate v1

busunkim96:
<del>I named the directory `cloudbuild` rather than `build`.</del>

<del>The generator names the package `google-cloud-cloudbuild`, but I'd be open to naming it `google-cloud-build`. Thoughts?
</del>

Product team said they preferred `google-cloud-build`, so I've added a synth replace.

TODO:
- [X] Setup Kokoro configs internally
- [x] Add Kokoro config in GitHub


cloudbuild/docs/conf.py
```diff
@@ -0,0 +1,363 @@
...
+source_suffix = [".rst", ".md"]
+
+# The encoding of source files.
+# source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = "index"
+
+# General information about the project.
+project = u"google-cloud-cloudbuild"
```
tseaver: Is there a document defending this non-standard naming convention?  We don't use the `cluod` prefix for other APIs, even where their "DNS names" include it (`cloudiot`, `cloudtrace`, etc.).

cloudbuild/google/cloud/devtools/cloudbuild_v1/cloudbuild.proto
```diff
@@ -0,0 +1,1304 @@
...
+message ListWorkerPoolsRequest {
+  // ID of the parent project.
+  string parent = 1;
+}
+
+// Response containing existing `WorkerPools`.
+message ListWorkerPoolsResponse {
+  // `WorkerPools` for the project.
+  repeated WorkerPool worker_pools = 1;
+}
```
tseaver: The corresponding `cloudbuild_pb2.py` file is missing.  Also, this file looks to be a duplicate of ` cloudbuild/google/cloud/devtools/cloudbuild_v1/cloudbuild.proto`, which makes me think that the whole `devtools` subtree should be removed.

cloudbuild/google/cloud/cloudbuild_v1/proto/cloudbuild.proto
```diff
@@ -0,0 +1,1304 @@
...
+message ListWorkerPoolsRequest {
+  // ID of the parent project.
+  string parent = 1;
+}
+
+// Response containing existing `WorkerPools`.
+message ListWorkerPoolsResponse {
+  // `WorkerPools` for the project.
+  repeated WorkerPool worker_pools = 1;
+}
```
tseaver: The `cloudbuild_pb2.py` file is missing.

cloudbuild/docs/conf.py
busunkim96: I'll rename it to `google-cloud-build` (this file was missed)

cloudbuild/google/cloud/cloudbuild_v1/proto/cloudbuild.proto
busunkim96: Moved `cloudbuild.proto` to `google/cloud/devtools/cloudbuild_v1/proto/cloudbuild.proto` (and removed `google/cloud/cloudbuild_v1`.

cloudbuild/docs/conf.py
busunkim96: Synth has been fixed to replace all instances of `google-cloud-cloudbuild`

cloudbuild/google/cloud/devtools/cloudbuild_v1/cloudbuild.proto
busunkim96: This was the wrong place for this file. It has been moved. 

Library lives under `cloudbuild/google/cloud/devtools/cloudbuild_v1`

tseaver:
@busunkim96 Try installing `google-cloud-containeranalysis` and the new `google-cloud-build` library in the same virtualenv:  the fact that `google.cloud.devtools` is *not* declared by both as a namespace package means that one of the two won't be importable (or maybe the second install will fail, depending on pip's behavior).

busunkim96:
@tseaver Ah I see. So my options are:

1) change `containeranalysis` and `cloudbuild` to both declare `google.cloud.devtools` as a namespace package.
2) place `cloudbuild` in `google.cloud.cloudbuild`

I'll probably do 1, and note in the README that you need google-cloud-containeranalysis>=x.x.x to be able to install side by side with `google-cloud-build`.

Issue #9482, opened 2019-10-16
Storage: `copy_blob` doesn't respect `preserve_acl`


HenrikOssipoff:
Thanks for stopping by to let us know something could be better!

#### Environment details

* macOS 10.15
* Python 3.6.5
* google-cloud-storage==1.20.0

#### Steps to reproduce

  1. Upload a blob with a predefined ACL of publicRead
  2. Copy this blob with preserve_acl set to true
  3. Expected result is that the new blob is set to publicRead, but this ACL isn't actually preserved
  4. Calling `make_public()` on the new blob correct sets it to publicRead

#### Code example

```python
file_obj = BytesIO(...)
blob.upload_from_file(file_obj, content_type='...', predefined_acl='publicRead')

bucket = storage_client.get_bucket(...)
bucket.copy_blob(blob, destination_bucket=bucket, new_name='new-name.jpg', preserve_acl=True)
# At this point, the blob will not be set to public
```

busunkim96:
@frankyn Is this the expected behavior for `preserve_acl`?

frankyn:
Taking a look, thanks for the ping.

frankyn:
Hi folks,

Based on the code, https://github.com/googleapis/google-cloud-python/blob/a79d98de3e9d6890c4e98d33cecae2f80550fdfd/storage/google/cloud/storage/bucket.py#L1121

I don't see anything necessarily copying ACLs. When exercising the example provided by @HenrikOssipoff, It's not something Cloud Storage does on a copy request.

We may want to deprecate this parameter.

crwilcox:
I agree with @frankyn. Let's mark this deprecated.

tseaver:
The [`objects.copy` API](https://cloud.google.com/storage/docs/json_api/v1/objects/copy) actually takes an optional `Object` resource in the request body, and applies any metadata to the destination object.  Maybe we need to use pass the source blob's properties (including ACL, which we normally don't pass)?

frankyn:
I'm okay with not introducing support for copying ACL in this case. It reduces potential security risks. 

However, I do see the potential for having a destination_blob with metadata parameter being passed in. That would help in this case.

tseaver:
@frankyn We currently *seem* to promise that the ACL will be copied unless the user passes `preserve_acl=False`.  Can you check whether the backend changed the "copy ACL" behavior, and if so, when?

Alternatives:

[...]

PR #9518, opened 2019-10-23
feat(storage): mark preserve_acl as deprecated

Emar-Kar:
Closes: #9482 

As was asked in the [comments](https://github.com/googleapis/google-cloud-python/issues/9482#issuecomment-543922357), argument `preserve_acl` marked as deprecated.

### DIFF
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -1064,7 +1064,7 @@ class Bucket(_PropertyMixin):
         destination_bucket,
         new_name=None,
         client=None,
-        preserve_acl=True,
+        preserve_acl=None,
         source_generation=None,
     ):
         """Copy the given blob to the given bucket, optionally with a new name.
@@ -1088,7 +1088,9 @@ class Bucket(_PropertyMixin):
 
         :type preserve_acl: bool
         :param preserve_acl: Optional. Copies ACL from old blob to new blob.
-                             Default: True.
+                             Default: True. Deprecated: this argument is no longer
+                             affecting anything and will be removed in a future release.
+                             Please update the bucket's ACL manually.
 
         :type source_generation: long
         :param source_generation: Optional. The generation of the blob to be
@@ -1118,8 +1120,12 @@ class Bucket(_PropertyMixin):
             _target_object=new_blob,
         )
 
-        if not preserve_acl:
-            new_blob.acl.save(acl={}, client=client)
+        if preserve_acl is not None:
+            warnings.warn(
+                "The 'preserve_acl' argument is deprecated.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
 
         new_blob._set_properties(copy_result)
         return new_blob
--- a/storage/google/cloud/storage/bucket.py
+++ b/storage/google/cloud/storage/bucket.py
@@ -1088,7 +1088,7 @@ class Bucket(_PropertyMixin):
 
         :type preserve_acl: bool
         :param preserve_acl: Optional. Copies ACL from old blob to new blob.
-                             Default: True. Deprecated: this argument is no longer
+                             Default: None. Deprecated: this argument is no longer
                              affecting anything and will be removed in a future release.
                              Please update the bucket's ACL manually.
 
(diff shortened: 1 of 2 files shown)


storage/google/cloud/storage/bucket.py
```diff
@@ -1118,8 +1120,12 @@ def copy_blob(
             _target_object=new_blob,
         )
 
-        if not preserve_acl:
-            new_blob.acl.save(acl={}, client=client)
+        if preserve_acl is not None:
+            warnings.warn(
+                "The 'preserve_acl' argument is deprecated. Please update the bucket's ACL manually.",
```
tseaver: This isn't "deprecation":  we are changing the behavior, and emitting a warning.  If we *are* going to deprecate the argument, then we should preserve the existing behavior, and emit an explicit deprecation, e.g.:

```python
        if preserve_acl is not None:
            warnings.warn(
                "The 'preserve_acl' argument is deprecated.  For forward compatibility, "
                 "please update the blob's ACL manually.",
                 DeprecationWarning,
                 stacklevel=2,
            )
        if not preserve_acl and preserve_acl is not None:
            new_blob.acl.save(acl={}, client=client)
```

storage/tests/unit/test_bucket.py
```diff
@@ -1166,27 +1166,12 @@ def test_copy_blobs_preserve_acl(self):
...
-        self.assertIs(new_blob.bucket, dest)
-        self.assertEqual(new_blob.name, NEW_NAME)
-        self.assertIsInstance(new_blob.acl, ObjectACL)
-
-        kw1, kw2 = connection._requested
-        COPY_PATH = "/b/{}/o/{}/copyTo/b/{}/o/{}".format(
-            SOURCE, BLOB_NAME, DEST, NEW_NAME
+        source.copy_blob(blob, dest, NEW_NAME, client=client, preserve_acl=False)
+        mock_warn.assert_called_once_with(
+            "The 'preserve_acl' argument is deprecated. Please update the bucket's ACL manually.",
```
tseaver: Please don't assert the warning message (it also happens to be incorrect, as the ACL is for the blob, not its bucket).

tseaver:
ISTM that we could make the `preserve_acl` argument actually work, rather than deprecating it, by passing the ACL in the body of the `objects.copy` request, as [suggested in the docs](https://cloud.google.com/storage/docs/json_api/v1/objects/copy).

storage/google/cloud/storage/bucket.py
frankyn: If we do that then, the warning will always be emitted. I think the change makes sense.  Otherwise we can do a soft deprecation and only mark it in documentation.

frankyn:
@tseaver just closing the loop, as noted in the reference issue, let's not copy the ACL for the user. Instead it would be better for the user to supply blob metadata that can be used in the copy request. It's a clear opt-in. 

Let's deprecate this field and instead look at introducing acl= parameter a user can set in a subsequent PR?

tseaver:
@frankyn Seems like we should provide a mechanism for copying *any* properties into the new blob.  Maybe the most elegant would be to add a new method to `Blob`, `copy_from`.  The usage would be:

```python
source = source_bucket.blob('source-name')
dest = dest_bucket.blob('dest-name')
dest.content_type = source.content_type
# Copy other properties here?
dest.copy_from(source, copy_acl=True)
```

frankyn:
This exact question came up in Ruby, short update:

The GCS API will copy a subset of metadata from the object on copy or rewrite (excluding ACL's). 
As long as the body doesn't include metadata for the object then GCS copies the metadata.
One issue is if the user wants to provide metadata for the copied object, then GCS will not copy the source metadata.

I like the example:
```python
source = source_bucket.blob('source-name')
dest = dest_bucket.blob('dest-name')
dest.content_type = source.content_type
# User Modifies ACLs and copy_from only uses source for its media. 
dest.copy_from(source) 
# no additional parameters here "unless" the user wants to copy over existing metadata in the same way GCS does. We did this in Ruby using a flag force_metadata_copy to respect GCS behavior when a resource is provided. Otherwise, only the metadata provided on the copy/rewrite will be kept. 
```

Emar-Kar:
@frankyn So, the question is that we need probably separate those tasks. First, deprecate `preserve_acl`. In this question, I'm voting for the deprecation warning. I think users need to see that this argument is not affecting ACL of the new blob to avoid misunderstandings and incorrect behavior. Secondly, open a FR issue to add a new method to copy source data.

plamut:
Please reopen in the [new repo](https://github.com/googleapis/python-storage/pull/43), if this PR is still on track, thanks!

PR #9524, opened 2019-10-23
docs(datastore): remove duplication of API

foxyblue:
Noticed that `API` was repeated, possibly not caught since it was on a newline in the readme.

I could not find another reference to _Google Cloud Datastore API API_.

Reject my change if I am mistaken.
Thanks


busunkim96:
Thank you for the patch! @foxyblue

PR #9533, opened 2019-10-24
feat(storage): add 'ARCHIVE' storage class

### LINKED ISSUE #9532: Storage: Add support for Archive storage class
GCS is adding a storage class called "archive" to the [existing options for object storage](https://cloud.google.com/storage/docs/storage-classes) (coldline, standard, etc). Support for this should be added to this client library.

This should be straightforward as ARCHIVE should just mirror existing valid storage classes. I see the following places where changes are required in order to accomplish this:
* https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/constants.py (add new constant)
* https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/bucket.py  (add import, update STORAGE_CLASSES, update documentation).
* https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py (same as above)
* Incorporate ARCHIVE into unit tests as appropriate.

This has already been implemented in the C++ client library; reference commit [here](https://github.com/googleapis/google-cloud-cpp/pull/3197)

tseaver:
Closes #9532.

### DIFF
--- a/storage/google/cloud/storage/constants.py
+++ b/storage/google/cloud/storage/constants.py
@@ -16,16 +16,28 @@
 # Storage classes
 
 STANDARD_STORAGE_CLASS = "STANDARD"
-"""Storage class for objects accessed more than once per month."""
+"""Storage class for objects accessed more than once per month.
+
+See: https://cloud.google.com/storage/docs/per-object-storage-class
+"""
 
 NEARLINE_STORAGE_CLASS = "NEARLINE"
-"""Storage class for objects accessed at most once per month."""
+"""Storage class for objects accessed at most once per month.
+
+See: https://cloud.google.com/storage/docs/per-object-storage-class
+"""
 
 COLDLINE_STORAGE_CLASS = "COLDLINE"
-"""Storage class for objects accessed at most once per year."""
+"""Storage class for objects accessed at most once per year.
+
+See: https://cloud.google.com/storage/docs/per-object-storage-class
+"""
 
 ARCHIVE_STORAGE_CLASS = "ARCHIVE"
-"""Storage class for objects accessed less frequently than once per year."""
+"""Storage class for objects accessed less frequently than once per year.
+
+See: https://cloud.google.com/storage/docs/per-object-storage-class
+"""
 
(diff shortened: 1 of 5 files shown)


storage/google/cloud/storage/constants.py
```diff
@@ -16,13 +16,28 @@
 # Storage classes
 
 STANDARD_STORAGE_CLASS = "STANDARD"
-"""Storage class for objects accessed more than once per month."""
+"""Storage class for objects accessed more than once per month.
+
+See: https://cloud.google.com/storage/docs/per-object-storage-class
```
tritone: The link resolves to https://cloud.google.com/storage/docs/storage-classes , so let's just use that?

Also, I'm assuming you repeat this for every storage class so that it gets picked up in the generated docs, but is there a way to make it show up just once that would be intelligible? If not the duplication is fine.

tseaver: > The link resolves to https://cloud.google.com/storage/docs/storage-classes , so let's just use that?

03ba4372102

> Also, I'm assuming you repeat this for every storage class so that it gets picked up in the generated docs, but is there a way to make it show up just once that would be intelligible? If not the duplication is fine.

I don't see a good way do avoid repeating the URL without introducing lots of churn.

tseaver:
@tritone, @frankyn, @crwilcox Please approve when the feature is released on the back end.

storage/google/cloud/storage/constants.py
tritone: Looks good! I will update/approve once the feature is no longer behind a flag (may be a couple of weeks).

tritone:
> Update: we're now good to merge this with a plan to release the changes between 1/2 and 1/8 (once the release freeze has ended). Let me know if there's any issue with this timeline!

... also, it looks like we still need approval from a repo owner (@frankyn or @crwilcox ) as well as resolution of the Kokoro failure for Cloud Build (perhaps this just requires a rebase/rerun?).

crwilcox:
CI is looking good (ignore cloud build failure). Merging.

PR #9536, opened 2019-10-24
docs(translation): fix v2 imports, remove v2 sample in README

busunkim96:
Default version is now v3, and the README example is not tested. Equivalent samples can be found in the v2 documentation or on cloud.google.com.


translate/docs/usage.rst
```diff
@@ -1,4 +1,4 @@
-Using the Translate Client
+Translation v2 Usage Guide
```
tseaver: Should we be renaming this file to `usage_v2.rst` / `v2_usage.rst`?  And why do we not publish usage examples for V3?

busunkim96: > Should we be renaming this file to usage_v2.rst / v2_usage.rst?

We're still relying on a lot of redirects, so I'd prefer not to change the name of the file to preserve the same path. 

>  And why do we not publish usage examples for V3?

v3beta1 and v3 are generated clients, so generated snippets are in client guide and proper samples live on cloud.google.com.

PR #9537, opened 2019-10-24
doc(bigquery): add code sample for scripting

tswast:
This code sample initiates a scripting job and then demonstrates how to
fetch the results and the child jobs.

Removes system test for scripting, as it's redundant with the code
sample.

### DIFF
--- a/bigquery/samples/query_script.py
+++ b/bigquery/samples/query_script.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 def query_script(client):
     # [START bigquery_query_script]
     # TODO(developer): Import the client library.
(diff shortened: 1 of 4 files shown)


bigquery/samples/query_script.py
```diff
@@ -0,0 +1,69 @@
...
+    child_jobs_iterable = client.list_jobs(parent_job=parent_job)
+    for child_job in child_jobs_iterable:
+        child_rows = list(child_job.result())
+        print(
+            "Child job with ID {} produced {} row(s).".format(
+                child_job.job_id, len(child_rows)
+            )
+        )
+
+    # [END bigquery_query_script]
```
tseaver: Should this sample be pulled into the docs via a `.. literalinclude:: ../samples/query_script.py` somewhere?

plamut: Discussed offline with @tswast  - not required, but it is not a bad idea.

PR #9543, opened 2019-10-25
fix(storage): update storage endpoint from www.googleapis.com to storage.googleapis.com

crwilcox:


storage/google/cloud/storage/_http.py
```diff
@@ -29,7 +29,7 @@ class Connection(_http.JSONConnection):
     :param client_info: (Optional) instance used to generate user agent.
     """
 
-    DEFAULT_API_ENDPOINT = _http.API_BASE_URL
+    DEFAULT_API_ENDPOINT = "https://storage.googleapis.com"
```
frankyn: nice catch!

crwilcox:
Cloud build is expected to fail as it is in progress yet. Merging.

PR #9548, opened 2019-10-28
fix: add gcloud-python user agent header

crwilcox:
Some metrics systems expect gcloud-{lang}/{library_version}

### DIFF
--- a/api_core/google/api_core/client_info.py
+++ b/api_core/google/api_core/client_info.py
@@ -92,5 +92,6 @@ class ClientInfo(object):
 
         if self.client_library_version is not None:
             ua += "gccl/{client_library_version} "
+            ua += "gcloud-python/{client_library_version} "
 
         return ua.format(**self.__dict__).strip()


api_core/google/api_core/client_info.py
```diff
@@ -92,5 +92,6 @@ def to_user_agent(self):
 
         if self.client_library_version is not None:
             ua += "gccl/{client_library_version} "
+            ua += "gcloud-python/{client_library_version} "
```
tseaver: Why would we want to repeat the same version with a different prefix?

crwilcox: The gccl header is the way forward, we have systems that still expect gcloud-python. This is to make sure both legacy and forward systems have data in the form they expect. I will mark this dnm for now.

tseaver: ISTM we need to double-check that adding `gcloud-python` back in doesn't break any of the log parsers expecting the "standard" format.

crwilcox: It shouldn't be an issue due to the additive nature of the ua strings. That said, I am currently in discussions with some folks.

crwilcox:
I am closing this. I will open a pr that adds it just to storage

PR #9551, opened 2019-10-28
fix: add gcloud-python header to user agent

crwilcox:


storage/google/cloud/storage/_http.py
```diff
@@ -35,6 +35,11 @@ def __init__(self, client, client_info=None, api_endpoint=DEFAULT_API_ENDPOINT):
         super(Connection, self).__init__(client, client_info)
         self.API_BASE_URL = api_endpoint
         self._client_info.client_library_version = __version__
+        
+        # TODO: When metrics all use gccl, this should be removed b/143493862
```
frankyn: Could we link to public issue tracker instead to align internal and external comms?

crwilcox: Sure :)

Issue #9457, opened 2019-10-11
[BigQuery Data Transfer Service, all GAPIC clients]: provide a close() method and context manager to clean up client-owned resources


yebrahim:
#### Environment details

`pip list` output:
```
cachetools                         3.1.1    
certifi                            2019.9.11
chardet                            3.0.4    
google-api-core                    1.14.3   
google-api-python-client           1.7.11   
google-auth                        1.6.3    
google-auth-httplib2               0.0.3    
google-cloud-bigquery              1.20.0   
google-cloud-bigquery-datatransfer 0.4.1    
google-cloud-core                  1.0.3    
google-cloud-pubsub                1.0.2    
google-resumable-media             0.4.1    
googleapis-common-protos           1.6.0    
grpc-google-iam-v1                 0.12.3   
grpcio                             1.24.1   
httplib2                           0.14.0   
idna                               2.8      
pip                                19.2.3   
protobuf                           3.10.0   
pyasn1                             0.4.7    
pyasn1-modules                     0.2.6    
pytz                               2019.3   
requests                           2.22.0   
rsa                                4.0      
setuptools                         41.2.0   
six                                1.12.0   
uritemplate                        3.0.0    
urllib3                            1.25.6   
wheel                              0.33.6   
```
OS: Ubuntu
Python 3.5.2
API: BigQuery Data Transfer Service

#### Steps to reproduce
Initialize more than one data transfer client.

#### Code example

```python
from google.cloud import bigquery_datatransfer_v1
import time

while True:
  dts_client = bigquery_datatransfer_v1.DataTransferServiceClient()
  try:
    dts_client.get_transfer_run('some run id').state
  except:
    print('error!')

  time.sleep(2)
```
Output of `ll /proc/<pid>/fd` after three seconds:
0  1  10  11  2  3  4  5  6  7  8  9
And keeps growing.

There's also no way to clean up the client as it doesn't implement `__exit__`.

tswast:
This client is 100% auto-generated via our gRPC toolchain. I'll try to summon some of our folks who work at those lower levels to investigate.

busunkim96:
@yebrahim This is unrelated, but you only need to create the `dts_client` once.

```python
from google.cloud import bigquery_datatransfer_v1
import time

dts_client = bigquery_datatransfer_v1.DataTransferServiceClient()
while True:
  try:
    dts_client.get_transfer_run('some run id').state
  except:
    print('error!')

  time.sleep(2)
```

[...]

PR #9573, opened 2019-10-30
feat(bigquery): add `create_bqstorage_client` param to `to_dataframe` and `to_arrow`

tswast:
When the `create_bqstorage_client` parameter is set to `True`, the BigQuery
client constructs a BigQuery Storage API client for you. This removes
the need for boilerplate code to manually construct both clients
explicitly with the same credentials.

Does this make the `bqstorage_client` parameter unnecessary? In most
cases, yes, but there are a few cases where we'll want to continue using
it. Specifically, when partner tools use `to_dataframe`, they should
continue to use `bqstorage_client` so that they can set the correct
amended user-agent strings. `bqstorage_client` is also needed for regional
API endpoints.

TODO
- [x] Add `owns_bqstorage_client` to track if `bqstorage_client` needs its transport closed.
- [x] Add try/finally block where finally closes the transport only if `owns_bqstorage_client` is true.
- [x] Test that client is not created if `max_results` is set.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -353,6 +353,19 @@ class Client(ClientWithProject):
 
         return DatasetReference(project, dataset_id)
 
+    def _create_bqstorage_client(self):
+        """Create a BigQuery Storage API client using this client's credentials.
+
+        Returns:
+            google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient:
+                A BigQuery Storage API client.
+        """
+        from google.cloud import bigquery_storage_v1beta1
+
+        return bigquery_storage_v1beta1.BigQueryStorageClient(
+            credentials=self._credentials
+        )
+
     def create_dataset(self, dataset, exists_ok=False, retry=DEFAULT_RETRY):
         """API call: create the dataset via a POST request.
 
--- a/bigquery/samples/tests/test_download_public_data.py
+++ b/bigquery/samples/tests/test_download_public_data.py
@@ -1,0 +1,24 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .. import download_public_data
+
+
+def test_download_public_data(capsys, client):
+
+    download_public_data.download_public_data(client)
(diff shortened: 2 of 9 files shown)


shollyman:
feedback during informal review: be clearer we're only using default settings when constructing the client on user behalf, and that to do more you'll still want to supply your own client.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -341,6 +341,19 @@ def dataset(self, dataset_id, project=None):
 
         return DatasetReference(project, dataset_id)
 
+    def _create_bqstorage_client(self):
+        """Create a BigQuery Storage API client using this client's credentials.
+
+        Returns:
+            google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient:
```
plamut: Should the return type be cross-linked in the docs? I personally find that useful.
```suggestion
            :class:`google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient`:
```

bigquery/samples/tests/test_download_public_data.py
```diff
@@ -0,0 +1,24 @@
...
+from .. import download_public_data
+
+
+def test_download_public_data(capsys, client):
+
+    download_public_data.download_public_data(client)
+    out, _ = capsys.readouterr()
+    assert "year" in out
+    assert "gender" in out
+    assert "name" in out
```
plamut: How does this check that the BQ storage client was indeed used? Or is it solely for making sure that the sample indeed runs without errors?

(the same for the sandbox version of the sample)

**Edit:** I see that the related unit tests cover this aspect. Let's maybe just add a comment about that here to avoid any doubts from the future readers?

bigquery/google/cloud/bigquery/client.py
tswast: This method is private, so it won't show up in the docs anyway.

For `Returns:` in general, an explicit `:class:` link isn't necessary. For example, below you can see the `create_dataset` method does not use an explicit `:class:` link, but the Return Type still renders as a link.

image

plamut: I see, all good then.

bigquery/samples/tests/test_download_public_data.py
tswast: I've updated the test to verify that a BigQuery Storage API session appears in the DEBUG logs.

bigquery/samples/tests/test_download_public_data_sandbox.py
```diff
@@ -12,13 +12,23 @@
...
 
 
-def test_download_public_data_sandbox(capsys, client):
+def test_download_public_data_sandbox(caplog, capsys, client):
+    # Enable debug-level logging to verify the BigQuery Storage API is used.
+    caplog.set_level(logging.DEBUG)
 
     download_public_data_sandbox.download_public_data_sandbox(client)
-    out, _ = capsys.readouterr()
+    out, err = capsys.readouterr()
```
plamut: `err` is not used anywhere?

tswast:
Per https://github.com/googleapis/google-cloud-python/issues/9457

I should close the client transport if we create it in the method.

PR #9574, opened 2019-10-31
docs(bigquery): document how to achieve higher write limit and add tests

### LINKED ISSUE #9539: BigQuery:  provide option to omit 'insertId' for 'insert_rows' and 'insert_rows_json'.
Streaming inserts to BigQuery can achieve [1GB per second](https://cloud.google.com/bigquery/quotas#streaming_inserts) but only when insertId is omitted.  Currently both methods from the google-cloud-bigquery [client](https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/client.py) that implement the BigQuery insertAll REST API method,  **insert_rows** and **insert_rows_json**, automatically add insertId preventing anyone from making use of the higher throughput limits.

Proposing that an argument be added to one or both of these methods like "insertId=True" (defaulting to true).  When false, uuid is not used to add insert IDs.  Furthermore, documentation is updated to clearly articulate that omitting insertId can result in duplicate records upon retry of failed API calls according to [this guide](https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataconsistency).

plamut:
Fixes #9539.

This PR documents how to achieve the higher write limit on streaming inserts, and extends unit tests to cover that case (i.e. no row insert IDs).

### How to test
As mentioned in the last [issue comment](https://github.com/googleapis/google-cloud-python/issues/9539#issuecomment-546619232) - check that the new code sample and the docs are indeed valid, and that new tests are sound

### DIFF
--- a/bigquery/docs/usage/tables.rst
+++ b/bigquery/docs/usage/tables.rst
@@ -122,6 +122,20 @@ Insert rows into a table's data with the
    :start-after: [START bigquery_table_insert_rows]
    :end-before: [END bigquery_table_insert_rows]
 
+Insert rows into a table's data with the
+:func:`~google.cloud.bigquery.client.Client.insert_rows` method, achieving
+higher write limit:
+
+.. literalinclude:: ../samples/table_insert_rows_no_explicit_row_ids.py
+   :language: python
+   :dedent: 4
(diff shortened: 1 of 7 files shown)


plamut:
Renamed the methods along with the files as well for consistency.

Issue #9580, opened 2019-10-31
Testing: rationalize / normalize VPCSC environment detection in systests


tseaver:
We need a clear pattern for how to test whether the appropriate environment variables are set for VPCSC, and ways to skip tests when they are missing.  I believe the constraints should be:

- Never set the environment variables in `noxfile.py`.
- If the inside / outside project variables are missing, skip the affected testcases cleanly.

I'm proposing to create a new module, `test_utils/test_utils/vpcsc_config.py`, which centralizes all this policy.  Usage from systests would look like:

```python
from test_utils.vpcsc_config import vpcsc_config


@vpcsc_config.skip_unless_inside_project
def test_requiring_inside_project():
    do_something_with(vpcsc_config.project_inside)


@vpcsc_config.skip_unless_outside_project
def test_requiring_outside_project():
    do_something_with(vpcsc_config.project_outside)


@vpcsc_config.skip_unless_outside_bucket
def test_requiring_outside_bucket():
    do_something_with(vpcsc_config.bucket_outside)

@vpcsc_config.skip_unless_inside_project
@vpcsc_config.skip_unless_outside_project
def test_requiring_inside_and_outside_projects():
    if vpcsc_config.inside_vpcsc:
        do_something_with(vpcsc_config.project_inside, vpcsc_config.project_outside)

@vpcsc_config.skip_if_inside_vpcsc
def test_incompatible_with_vpcsc():
   ...


@vpcsc_config.skip_unless_inside_vpcsc
def test_requiring_requires_vpcsc():
   ...
```

Steps:
- [x] Add 'test_utils/test_utils/vpcsc_config.py`, with the indicated 'VPCSCConfig' class and singleton (#9581)
- [x] Apply in `asset/tests/system/test_vpcsc.py`.(#9614)
- [x] Apply in `automl/tests/system/gapic/v1beta1/test_system_tables_client_v1.py`. (#9607)
- [x] Apply in `dlp/tests/system/gapic/v2/test_system_dlp_service_v2_vpcsc.py`. (#9608)
- [x] Apply in `monitoring/tests/system/test_vpcsc_v3.py`. (#9615)
- [x] Apply in `storage/tests/system.py`. (#9616)
- [x] Apply in `trace/tests/system/gapic/v{1,3}/test_system_trace_service_v{1,2}_vpcsc.py`. (#9618)
- [x] Appy lin `translate/tests/system/test_vpcsc.py`. (#9619)
[...]

PR #9581, opened 2019-10-31
tests: add centralized VPCSC config detection

tseaver:
Toward #9580


tseaver:
@busunkim96, @crwilcox I believe that b4a9475 brings this PR into line with the documented spec.  I will hold off on the client-library-specific changes until this PR is merged.

PR #9582, opened 2019-10-31
tests(asset): normalize VPCSC config detection

tseaver:
Uses #9581 as a base.

Toward #9580

### DIFF
--- a/asset/noxfile.py
+++ b/asset/noxfile.py
@@ -118,20 +118,11 @@ def system(session):
     session.install("-e", "../test_utils/")
     session.install("-e", ".")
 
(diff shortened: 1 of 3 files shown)


tseaver:
This PR is premature, because the use of the `GOOGLE_CLOUD_TESTS_IN_VPCSC` envvar in the `asset` system tests is non-standard:  it defaults the variable to `True`, which is against the spec in [this doc](https://docs.google.com/document/d/1qCRRNqzKn7YnssbB1g-RTkJH11S3vmJ97JjsLWo6alA/edit).

PR #9587, opened 2019-11-01
[Cloud Security Center]  Create requirements.txt

emkornfield:
[Cloud Security Center] Create requirements.txt

Adding a requirements.txt file to be linked from public documentation.

CC @tabrarian

### DIFF
--- a/securitycenter/docs/requirements.txt
+++ b/securitycenter/docs/requirements.txt
@@ -1,0 +1,1 @@
+google-cloud-securitycenter


tseaver:
I don't understand the motivation for this change:  none of our other API libraries have `docs/requirements.txt` files.  There are two, `bigquery_datatransfer` and `bigquery_storage`, which have `samples/requirements.txt` files, used to populate the virtualenv which runs the samples.

busunkim96:
+1 to @tseaver. Is there some other documentation gap that needs to be addressed?

emkornfield:
@busunkim96 I CCed you the internal bug requesting this maybe you can way in there.

busunkim96:
(Made a comment on the internal bug as well)

Since the samples live with the library the `requirements.txt` is not necessary. But to make this library more like the others with samples, let's move them to their own directory `samples/`.

PR #9594, opened 2019-11-04
fix(pubsub): split large (mod)ACK requests into smaller ones

### LINKED ISSUE #9103: Pub/Sub: Send modify_ack_deadline requests in chunks
Currently leaser is limited in total number of messages (<2000?) which is a problem for use cases with many small messages. 

Same issue was solved in nodejs library here: https://github.com/googleapis/nodejs-pubsub/pull/65

plamut:
Fixes #9103.

There is a server-side limit on the maximum size of ACK and modACK requests, which can be hit if the leaser tries to manage too many messages in a single requests. This PR avoids the problem by splitting such large requests into multiple smaller requests.

### How to test

**Steps to reproduce:**
 - publish a lot of messages to a topic (significantly more than 3000)
 - subscribe to that topic using a streaming pull and a callback that processes messages for a non-negligible amount of time (e.g. 1 second)
- observe the DEBUG log

The following error can be observed in the log output:
```
google.api_core.exceptions.InvalidArgument: 400 Request payload size exceeds the limit: 524288 bytes.
```

There is no error, mass-acknowledging leased messages works just fine.

The following script might come helpful:
<details>
    <summary>reproduce_9103.py</summary>

```py
import itertools
import logging
import os
import threading
import time
from argparse import ArgumentParser

from google import api_core

logging.basicConfig(
    format='%(asctime)s %(levelname)s %(threadName)s: %(message)s', level=logging.INFO)
logger = logging.getLogger()
logger.setLevel("DEBUG")


log_format = (
)

logging.basicConfig(
    level=logging.DEBUG, format=log_format
)


def msg_handler(message):
    """Message handler."""
    SLEEP_FOR = 1

    logging.info(f"\x1b[1m[{message.message_id}] got message with content: {str(message.data)} (ack_id: {message.ack_id})\x1b[0m")
    logging.info(f"\x1b[1m[{message.message_id}] sleeping for {SLEEP_FOR} seconds\x1b[0m")

    time.sleep(SLEEP_FOR)
    logging.info(f"\x1b[1m[[{message.message_id}] done sleeping, sending ack()\x1b[0m")

    message.ack()


def _publish_messages(publisher, topic_path, batch_sizes):
    """Publish ``count`` messages in batches and wait until completion."""
    publish_futures = []
    msg_counter = itertools.count(start=1)

    for batch_size in batch_sizes:
        msg_batch = _make_messages(count=batch_size)
        for msg in msg_batch:
            future = publisher.publish(
                topic_path, msg, seq_num=str(next(msg_counter))
            )
            publish_futures.append(future)
        time.sleep(0.1)

    # wait untill all messages have been successfully published
    for future in publish_futures:
        future.result(timeout=30)

def _make_messages(count):
    messages = [
        u"message {}/{}".format(i, count).encode("utf-8")
        for i in range(1, count + 1)
    ]
    return messages


def main():
    PROJECT_ID = "TODO: set"
    SUBSCRIPTION_NAME = "TODO: set"
    TOPIC_NAME = "TODO: set"

    topic_path = publisher.topic_path(PROJECT_ID, TOPIC_NAME)
    subscription_path = subscriber.subscription_path(PROJECT_ID, SUBSCRIPTION_NAME)

    try:
        publisher.create_topic(topic_path)
        subscriber.create_subscription(subscription_path, topic_path)
    except api_core.exceptions.AlreadyExists as exc:
        pass

    #batch_sizes = (400, 600, 700, 300, 500, 500, 250, 750)  # total: 4000
    batch_sizes = (4000,)
    _publish_messages(publisher, topic_path, batch_sizes=batch_sizes)

    # now subscribe and do the main part, check for max pending messages
    total_messages = sum(batch_sizes)
    flow_control = pubsub_v1.types.FlowControl(max_messages=total_messages)
    subscription_future = subscriber.subscribe(
        subscription_path, msg_handler, flow_control=flow_control
    )

    logging.info("Starting streaming pull...")    
    future = subscriber.subscribe(
        subscription_path, msg_handler, flow_control=flow_control
    )
    logging.info("Streming pull started")

    time.sleep(5)

    try:
        future.result()
    except KeyboardInterrupt:
        logging.info("Interrupted, cancelling")
        future.cancel()


if __name__ == '__main__':
    main()

```
</details>

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -37,15 +37,15 @@ _MAX_BATCH_LATENCY = 0.01
 """The maximum amount of time in seconds to wait for additional request items
 before processing the next batch of requests."""
 
-_ACK_IDS_BATCH_SIZE = 3000
+_ACK_IDS_BATCH_SIZE = 2500
 """The maximum number of ACK IDs to send in a single StreamingPullRequest.
 
 The backend imposes a maximum request size limit of 524288 bytes (512 KiB) per
 acknowledge / modifyAckDeadline request. ACK IDs have a maximum size of 164
-bytes, thus we cannot send more than o 524288/164 ~= 3197 ACK IDs in a single
+bytes, thus we cannot send more than o 524288/176 ~= 2979 ACK IDs in a single
 StreamingPullRequest message.
 
-Accounting for some overhead, we should thus only send a maximum of 3000 ACK
+Accounting for some overhead, we should thus only send a maximum of 2500 ACK
 IDs at a time.
 """
 
--- a/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+from __future__ import division
 
 import collections
+import itertools
 import logging
+import math
 import threading
 
 from google.cloud.pubsub_v1 import types
@@ -33,6 +36,18 @@ _MAX_BATCH_SIZE = 100
 _MAX_BATCH_LATENCY = 0.01
 """The maximum amount of time in seconds to wait for additional request items
 before processing the next batch of requests."""
+
+_ACK_IDS_BATCH_SIZE = 3000
+"""The maximum number of ACK IDs to send in a single StreamingPullRequest.
+
+The backend imposes a maximum request size limit of 524288 bytes (512 KiB) per
+acknowledge / modifyAckDeadline request. ACK IDs have a maximum size of 164
+bytes, thus we cannot send more than o 524288/164 ~= 3197 ACK IDs in a single
+StreamingPullRequest message.
+
+Accounting for some overhead, we should thus only send a maximum of 3000 ACK
+IDs at a time.
+"""
 
 
 class Dispatcher(object):
@@ -119,9 +134,16 @@ class Dispatcher(object):
             if time_to_ack is not None:
                 self._manager.ack_histogram.add(time_to_ack)
 
-        ack_ids = [item.ack_id for item in items]
-        request = types.StreamingPullRequest(ack_ids=ack_ids)
-        self._manager.send(request)
+        # We must potentially split the request into multiple smaller requests
+        # to avoid the server-side max request size limit.
+        ack_ids = (item.ack_id for item in items)
+        total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
+
+        for _ in range(total_chunks):
+            request = types.StreamingPullRequest(
+                ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE)
+            )
+            self._manager.send(request)
 
         # Remove the message from lease management.
         self.drop(items)
@@ -150,13 +172,18 @@ class Dispatcher(object):
         Args:
             items(Sequence[ModAckRequest]): The items to modify.
         """
-        ack_ids = [item.ack_id for item in items]
-        seconds = [item.seconds for item in items]
+        # We must potentially split the request into multiple smaller requests
+        # to avoid the server-side max request size limit.
+        ack_ids = (item.ack_id for item in items)
+        seconds = (item.seconds for item in items)
+        total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
 
-        request = types.StreamingPullRequest(
-            modify_deadline_ack_ids=ack_ids, modify_deadline_seconds=seconds
-        )
-        self._manager.send(request)
+        for _ in range(total_chunks):
+            request = types.StreamingPullRequest(
+                modify_deadline_ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE),
+                modify_deadline_seconds=itertools.islice(seconds, _ACK_IDS_BATCH_SIZE),
+            )
+            self._manager.send(request)
 
     def nack(self, items):
         """Explicitly deny receipt of messages.
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
@@ -102,9 +102,9 @@ def test_ack_splitting_large_payload():
     dispatcher_ = dispatcher.Dispatcher(manager, mock.sentinel.queue)
 
     items = [
-        # use realistic lengths for ACK IDs (max 164 bytes)
-        requests.AckRequest(ack_id=str(i).zfill(164), byte_size=0, time_to_ack=20)
-        for i in range(6001)
+        # use realistic lengths for ACK IDs (max 176 bytes)
+        requests.AckRequest(ack_id=str(i).zfill(176), byte_size=0, time_to_ack=20)
+        for i in range(5001)
     ]
     dispatcher_.ack(items)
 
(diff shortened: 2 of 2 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -34,6 +37,18 @@
 """The maximum amount of time in seconds to wait for additional request items
 before processing the next batch of requests."""
 
+_ACK_IDS_BATCH_SIZE = 3000
+"""The maximum number of ACK IDs to send in a single StreamingPullRequest.
+
+The backend imposes a maximum request size limit of 524288 bytes (512 KiB) per
+acknowledge / modifyAckDeadline request. ACK IDs have a maximum size of 164
+bytes, thus we cannot send more than o 524288/164 ~= 3197 ACK IDs in a single
+StreamingPullRequest message.
```
plamut: Picked the max ACK ID size from [here](https://github.com/googleapis/nodejs-pubsub/pull/65/files#diff-3d29c4447546c72118ed5d5cbf38ab8bR37) (a similar fix in the Node.js client).

plamut:
There might still be an issue here, do not merge yet...

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: We don't publish a set size for an ack id. It may change over time, though we rarely do so in practice. 2500 is fine for now.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -119,9 +134,16 @@ def ack(self, items):
...
-        request = types.StreamingPullRequest(ack_ids=ack_ids)
-        self._manager.send(request)
+        # We must potentially split the request into multiple smaller requests
+        # to avoid the server-side max request size limit.
+        ack_ids = (item.ack_id for item in items)
+        total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
+
+        for _ in range(total_chunks):
+            request = types.StreamingPullRequest(
+                ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE)
```
pradn: Wont this send only the first _ACK_IDS_BATCH_SIZE ack-ids |total-chunks|-times? It wont actually result in the sending of all of the ack ids, only the first _ACK_IDS_BATCH_SIZE.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -119,9 +134,16 @@ def ack(self, items):
             if time_to_ack is not None:
                 self._manager.ack_histogram.add(time_to_ack)
 
-        ack_ids = [item.ack_id for item in items]
-        request = types.StreamingPullRequest(ack_ids=ack_ids)
-        self._manager.send(request)
+        # We must potentially split the request into multiple smaller requests
+        # to avoid the server-side max request size limit.
+        ack_ids = (item.ack_id for item in items)
```
pradn: I'm curious why you chose a tuple comprehension here vs a list comprehension. Is it just to make the data immutable?

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -150,13 +172,18 @@ def modify_ack_deadline(self, items):
...
+        seconds = (item.seconds for item in items)
+        total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
+
+        for _ in range(total_chunks):
+            request = types.StreamingPullRequest(
+                modify_deadline_ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE),
```
pradn: The same comment about islice from above applies here too. I think only the first _ACK_IDS_BATCH_SIZE are going to be sent.

pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
```diff
@@ -95,6 +95,33 @@ def test_ack_no_time():
...
+    dispatcher_.ack(items)
+
+    calls = manager.send.call_args_list
+    assert len(calls) == 3
+
+    all_ack_ids = {item.ack_id for item in items}
+    sent_ack_ids = set()
+
+    for call in calls:
+        message = call.args[0]
```
pradn: Also need to check that different ack-ids were sent for each call.

pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
```diff
@@ -153,6 +180,33 @@ def test_modify_ack_deadline():
...
+    dispatcher_.modify_ack_deadline(items)
+
+    calls = manager.send.call_args_list
+    assert len(calls) == 3
+
+    all_ack_ids = {item.ack_id for item in items}
+    sent_ack_ids = set()
+
+    for call in calls:
+        message = call.args[0]
```
pradn: Likewise:
Also need to check that different ack-ids were sent for each call.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Since `ack_ids` is a generator, iterating over a slice when building a `StreamingPullRequest` instance consumes the first `_ACK_IDS_BATCH_SIZE` items. Then in the following iteration the _next_ `_ACK_IDS_BATCH_SIZE` items are consumed.

```py
>>> from itertools import islice
>>> BATCH_SIZE = 3
>>> gen = (x for x in range(8))
>>> list(islice(gen, BATCH_SIZE))
[0, 1, 2]
>>> list(islice(gen, BATCH_SIZE))
[3, 4, 5]
>>> list(islice(gen, BATCH_SIZE))
[6, 7]
>>> list(islice(gen, BATCH_SIZE))
[]
```
Iterating over a generator has the benefit of automatically advancing the generator, sparing us the need to keep track of the slice boundaries manually.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Actually, this is a generator expression, which lazily generates the items on demand when iterating over it. We don't need to keep the entire temporary collection of references to ACK IDs in memory just to iterate over them, lazily producing them on the fly is more space-efficient.

pubsub/tests/unit/pubsub_v1/subscriber/test_dispatcher.py
plamut: Agreed, will add additional assertions. :+1:

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Indeed. We discussed this at one of the meetings, and it was said that any changes to ACK ID sizes are both rare and minimal in terms of space. Since 2500 gives us a safety buffer of approximately ~20%, it is enough in practice.

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Gotcha, I had forgotten about these. This was the root of my error, thanks!

pubsub/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Oh, geez. I see. Thanks!

PR #9597, opened 2019-11-04
fix(pubsub): update batching and flow control parameters to be same as the other client libraries

pradn:
Changes:
- Max batch size is reduced from 10 MB to 1 MB
- Max latency before a batch is sent is reduced from 50 ms to 10 ms
- Max messages in a batch is reduced from 1000 messages to 100 messages
- Max messages in flight is increased from 100 messages to 1000 messages
- Max ack extension is reduced from 2 hours to 1 hour
- Also fix tests

### DIFF
--- a/pubsub/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
+++ b/pubsub/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
@@ -660,9 +660,9 @@ def test__on_response_with_leaser_overload():
         ]
     )
 
-    # Adjust message bookkeeping in leaser. Pick 99 messages, which is just below
+    # Adjust message bookkeeping in leaser. Pick 999 messages, which is just below
     # the default FlowControl.max_messages limit.
-    fake_leaser_add(leaser, init_msg_count=99, init_bytes=990)
+    fake_leaser_add(leaser, init_msg_count=999, init_bytes=990)
 
     # Actually run the method and prove that modack and schedule
     # are called in the expected way.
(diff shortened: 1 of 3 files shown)


hongalex:
@busunkim96 This PR is part of an effort to standardize our client library default settings. We're looking at two changes. One change increases how often we publish bundles of messages (which shouldn't be a breaking change).

The other has to do with flow control (how many messages we can hold and how long we hold onto messages before releasing them to be redelivered).  I'm more worried about this change negatively affecting users who aren't setting their own defaults. What do you recommend here?

pubsub/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -660,9 +660,9 @@ def test__on_response_with_leaser_overload():
         ]
     )
 
-    # Adjust message bookkeeping in leaser. Pick 99 messages, which is just below
+    # Adjust message bookkeeping in leaser. Pick 999 messages, which is just below
     # the default FlowControl.max_messages limit.
-    fake_leaser_add(leaser, init_msg_count=99, init_bytes=990)
+    fake_leaser_add(leaser, init_msg_count=999, init_bytes=990)
```
plamut: The `fake_leaser_add()` method increase the total byte size for every added message by 10. For consistency, `init_bytes` should thus be adjusted to `9990`.

While probably out of this PR's scope, an even better solution would be to introduce a convenience parameter such as `assumed_message_size`, replacing the currently hardcoded value of 10. And possibly even replacing the `init_bytes` parameter, as the latter can then be computed automatically.

I have some spare time to open a PR for that.

plamut: In short - the leaser keeps count of how many messages are being lease-managed, and their total size. This info is used by the streaming pull manager to compute the current load, and possibly pause/resume the message stream.

The leaser mock here mimics that bookkeeping, but currently adds a hardcoded value of 10 bytes to the tally for each added message, that's all. :)

pradn: Got it, thank you!

plamut:
Merging to unblock a related PR. If it turns out that additional changes are needed, we can make them in a follow-up PR.

PR #9599, opened 2019-11-05
feat(bigquery): add `--destination_table` parameter to bigquery magic

### LINKED ISSUE #9598: BigQuery: Add destination_table to %%bigquery magic to allow materializing query results
Discussed with @tswast over email, suggestion is:
We should not reuse the destination_var parameter. Also bq:// is not really something I've seen used in any other official BigQuery docs / libraries, and I don't think it's necessary.

Instead, I recommend adding a --destination_table argument which sets QueryJobConfig.destination.

vlasenkoalexey:

Sample use and test here: https://colab.research.google.com/drive/1vd_5IperzwBdqnzIPKd8fWkGZg4TO1Bb (google access only)

Fixes #9598 
@sshrdp FYI

### DIFF
--- a/bigquery/tests/unit/test_magics.py
+++ b/bigquery/tests/unit/test_magics.py
@@ -39,6 +39,7 @@ try:
     from google.cloud import bigquery_storage_v1beta1
 except ImportError:  # pragma: NO COVER
     bigquery_storage_v1beta1 = None
+from google.cloud import bigquery
 from google.cloud.bigquery import job
 from google.cloud.bigquery import table
 from google.cloud.bigquery import magics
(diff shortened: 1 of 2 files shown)


vlasenkoalexey:
@tswast could you take a look

bigquery/tests/unit/test_magics.py
```diff
@@ -336,6 +337,38 @@ def test__make_bqstorage_client_true_missing_gapic(missing_grpcio_lib):
...
+    dataset_id = "dataset_id"
+    client_patch = mock.patch(
+        "google.cloud.bigquery.magics.bigquery.Client", autospec=True
+    )
+    with client_patch as client_mock:
+        client_mock().location = "us"
+        client_mock().project = project
+        client_mock().get_dataset.side_effect = exceptions.NotFound("dataset not found")
+        result = magics._create_dataset_if_necessary(client_mock(), dataset_id)
+        client_mock().create_dataset.assert_called_once()
```
plamut: (minor)

It would IMO be a tad more clear to construct a mock client just once, instead of "instantiating" the mocked class multiple times, even if it returns the same object every time.
```py
with client_patch as client_mock:
    client = client_mock()
    client.location = "us"
    client.project = project
    ...
```

It's subjective, though, thus if you don't agree, feel free to ignore this comment.

vlasenkoalexey: everything makes sense, thanks for feedback!!!

vlasenkoalexey:
Is Kokoro - Cloud Build failure a real issue or just a glitch? It doesn't return any logs.

tswast:
Cloud Build client is currently broken, so this failure is expected.

plamut:
As @tswast said - the Cloud Build is currently broken, and the failure is unrelated to this PR.

PR #9605, opened 2019-11-05
test: check that 'request is prohibited' is in vpcsc exception

busunkim96:
Fixes internal issue 143725470

These tests started failing because the error message was modified on the backend. VPCSC folks recommended something like this instead.

### DIFF
--- a/asset/tests/system/test_vpcsc.py
+++ b/asset/tests/system/test_vpcsc.py
@@ -36,7 +36,7 @@ class TestVPCServiceControl(object):
         try:
             responses = call()
         except exceptions.PermissionDenied as e:
-            return e.message == "Request is prohibited by organization's policy"
+            return "Request is prohibited by organization's policy" in e.message
         except:
             pass
         return False
--- a/vision/tests/system.py
+++ b/vision/tests/system.py
@@ -640,8 +640,8 @@ class TestVisionClientProductSearchVpcsc(VisionSystemTestBase):
                 break
         except google.api_core.exceptions.PermissionDenied as e:
             # Verify the PermissionDenied exception was due to VPC SC.
-            self.assertEqual(
-                e.message, "Request is prohibited by organization's policy"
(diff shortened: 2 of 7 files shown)


asset/tests/system/test_vpcsc.py
```diff
@@ -36,7 +36,7 @@ def _is_rejected(call):
         try:
             responses = call()
         except exceptions.PermissionDenied as e:
-            return e.message == "Request is prohibited by organization's policy"
+            return "Request is prohibited by organization's policy" in e.message
```
tseaver: The assert calling this function [failed on python 3.7](https://source.cloud.google.com/results/invocations/191cc138-d1cc-4d8f-a71e-79bc28332ba5/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fasset/log).

vision/tests/system.py
```diff
@@ -640,8 +640,8 @@ def _verify_vpc_sc_error(self, call):
                 break
         except google.api_core.exceptions.PermissionDenied as e:
             # Verify the PermissionDenied exception was due to VPC SC.
-            self.assertEqual(
-                e.message, "Request is prohibited by organization's policy"
+            self.assertIn(
+                "Request is prohibited by organization's policy", e.message
```
tseaver: Needs blackening.

asset/tests/system/test_vpcsc.py
tseaver: I'm not sure how this is supposed to be exercised under a non-VPCSC setup.

tseaver: For instance, running on `master` locally with the "outside project" set, and after configuring the CI project to allow the Asset API, both system tests fail for me.

busunkim96:
Closing as this PR is obsolete

PR #9609, opened 2019-11-06
fix(spanner): fix TransactionPingingPool throwing error ''NoneType' object is not callable'

larkee:
TransactionPingingPool is throwing error ''NoneType' object is not callable' when trying to batch insert. This is the fix.


larkee:
The mock autospec does not support static attributes which causes test_put_non_full_w_active_txn to fail because txn.committed is a NonCallableMagicMock, not None. I don't see an easy fix for that.

Perhaps, it would be easier to go with the original change and have this refactor in a subsequent PR?

tseaver:
@larkee Hmm, odd:  I pasted in that change after making it locally, and got all the tests to pass.  Nevertheless, we can update the `_make_transaction()` factory to assign `txn.committed = None` (which is an appropriate initial condition).

larkee:
@tseaver PTAL

PR #9614, opened 2019-11-06
tests(asset): normalize VPCSC configuration in systests

tseaver:
Toward #9580.

### DIFF
--- a/asset/tests/system/test_vpcsc.py
+++ b/asset/tests/system/test_vpcsc.py
@@ -78,9 +78,9 @@ def test_batch_get_assets_history_outside(client, parent_outside):
     read_time_window = {}
     with pytest.raises(exceptions.PermissionDenied) as exc:
         client.batch_get_assets_history(
-        parent_outside,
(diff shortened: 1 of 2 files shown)


tseaver:
I tagged with "do not merge" because I believe we should refactor all these un-idiomatic tests along the lines of the changes in #9618.

asset/tests/system/test_vpcsc.py
```diff
@@ -22,67 +22,43 @@
...
+@vpcsc_config.skip_unless_inside_vpcsc
 class TestVPCServiceControl(object):
     @staticmethod
     def _is_rejected(call):
         try:
             responses = call()
         except exceptions.PermissionDenied as e:
             return e.message == "Request is prohibited by organization's policy"
-        except:
-            pass
```
busunkim96: I think we want to keep these two lines around.

If any error other than `PermissionDenied` is returned by the API (invalid request, etc.), it wasn't blocked by the VPC (which is the main reason these tests were written).

tseaver:
@busunkim96 I've removed the "do not merge" label after refactoring the VPCSC systests to be in line with #9618.

PR #9615, opened 2019-11-06
tests(monitoring): normalize VPCSC configuration in systests

tseaver:
Toward #9580.

### DIFF
--- a/monitoring/tests/system/test_vpcsc_v3.py
+++ b/monitoring/tests/system/test_vpcsc_v3.py
@@ -643,7 +643,9 @@ class TestCRUDUptimeCheckConfigs(object):
         ucc_client, uptime_check_config_path_inside
     ):
         with pytest.raises(exceptions.NotFound):  # no perms issue
-            ucc_client.update_uptime_check_config({"name": uptime_check_config_path_inside})
+            ucc_client.update_uptime_check_config(
+                {"name": uptime_check_config_path_inside}
+            )
 
     @staticmethod
     def test_update_uptime_check_config_outside(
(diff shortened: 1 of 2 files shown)


tseaver:
I tagged with "do not merge" because I believe we should refactor all these un-idiomatic tests along the lines of the changes in #9618.

monitoring/tests/system/test_vpcsc_v3.py
```diff
@@ -21,630 +21,397 @@
...
-
-            # If we reach this line, then call() did not raise. The return
-            # result must be either a google.api_core.page_iterator.Iterator
-            # instance, or None.
-            list(responses)
+            call()
         except exceptions.PermissionDenied as e:
-            logger.debug(e)
             return e.message == "Request is prohibited by organization's policy"
-        except Exception as e:
```
busunkim96: These tests are written to just check that requests are blocked/allowed in the VPC environment. A lot of them are invalid requests and will fail for other reasons, which is why this extra block is here.

tseaver: Please see the updated module in e403c1b :  there, I make actual assertions of the expected errors.

tseaver:
@busunkim96 I removed the "do not merge" label after updating the tests to follow the pattern in #9618.

busunkim96:
Looks like one of the test files needs to be blackened:

```
black --check google tests docs
would reformat /tmpfs/src/github/google-cloud-python/monitoring/tests/system/test_vpcsc_v3.py
Oh no! 💥 💔 💥
1 file would be reformatted, 64 files would be left unchanged.
Command black --check google tests docs failed with exit code 1
Session lint failed.
```

steinwaywhw:
Hi @tseaver shall we have a GVC? These tests are auto generated, not authored by human. The next test refresh will override any manual refactoring you've done.

busunkim96:
Hi @steinwaywhw, I believe your team maintains a script to make these tests? It'd be best for your generator to produce tests that look close to the present format (looking for the actual error that will be raised). The other VPCSC tests have also been changed to use this format.

Feel free to put something on my calendar if you would like to chat further.

PR #9641, opened 2019-11-08
feat(api core): simplify from_rfc3339 methods

Emar-Kar:
This PR was reopened for the clearness of the provided changes. (Old one [8742](https://github.com/googleapis/google-cloud-python/pull/8742))

Closes: #7676


api_core/google/api_core/bidi.py
```diff
@@ -172,7 +172,9 @@ def __init__(self, access_limit, time_window):
 
         self._time_window = time_window
         self._access_limit = access_limit
-        self._past_entries = collections.deque(maxlen=access_limit)  # least recent first
+        self._past_entries = collections.deque(
+            maxlen=access_limit
```
crwilcox: unrelated?

api_core/google/api_core/datetime_helpers.py
```diff
@@ -169,6 +156,27 @@ def from_rfc3339_nanos(value):
     return bare_seconds.replace(microsecond=micros, tzinfo=pytz.utc)
 
 
+def from_rfc3339_nanos(value):
```
crwilcox: I think we can just have `from_rfc3339_nanos=from_rfc3339`. This isn't typically directly used by users, and keeping it around as an alias isn't going to be a maintenance burden.

api_core/google/api_core/protobuf_helpers.py
```diff
@@ -15,17 +15,19 @@
...
-except ImportError:  # Python 2.7
-    import collections as collections_abc
 import copy
 import inspect
 
 from google.protobuf import field_mask_pb2
 from google.protobuf import message
 from google.protobuf import wrappers_pb2
 
+try:
```
crwilcox: This seems out of order? Did a linter do this?

Emar-Kar: It was some sort of the `flake8` error and my [build](https://source.cloud.google.com/results/invocations/57246fd0-a187-4c77-be97-85a89b8564b4/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fapi_core) failed a couple of times.
image

api_core/google/api_core/bidi.py
Emar-Kar: It was modified during the `black` session.

PR #9647, opened 2019-11-08
fix(automl): fix uploading pandas dataframe to AutoML Tables

### LINKED ISSUE #9483: AutoML: Tables client importing data with 'pandas_dataframe' fails.
#### Problem:
Client: automl_v1beta1
Class: TablesClient
Method: import_data

When using pandas_dataframe argument when specifying data to import to automl tables, get:
"google.api_core.exceptions.GoogleAPICallError: None Invalid column names: " error. I believe 

#### Environment details:
os: windows
python version: 3.5
google-cloud-automl==0.7.0
google==2.0.2
google-api-core==1.14.3

#### Code Example:
```
import pandas as pd
from google.cloud import automl_v1beta1

client = automl_v1beta1.TablesClient(project=<my_project>, region=<my_region>)
d = client.create_dataset(dataset_display_name=<ds_display_name)
data_df = pd.DataFrame({'a':[1,2,3], 'b':[4,5,6]})
response = client.import_data(dataset=d, pandas_dataframe=data_df)

def callback(operation_future):
    result = operation_future.result()
response.add_done_callback(callback)
```

#### Stack trace
`Error while executing Future callback.
  File "/home/jupyter/.local/lib/python3.5/site-packages/google/api_core/future/_helpers.py", line 37, in safe_invoke_callback
    return callback(*args, **kwargs)
  File "<ipython-input-41-d646f4ce2c0e>", line 6, in callback
    result = operation_future.result()
  File "/home/jupyter/.local/lib/python3.5/site-packages/google/api_core/future/polling.py", line 127, in result
    raise self._exception
google.api_core.exceptions.GoogleAPICallError: None Invalid column names: `

#### hypothesis:
it seems to be treating the dataframe index as a column in the import, and if this is not specified, will trigger the invalid column name error. If I set the index to one of the columns, it works.
Suggest leaving index out of import by default, and maybe optionally include index during import with keyword argument.

helinwang:
pandas.dataframe.to_csv() by default exports data index as an column
with empty column name. This causes uploading the export csv file to
fail because AutoML Tables does not allow empty column names. Given
that the data index is not useful for training the model. This PR
fixes the problem by setting the index argument to false so that the
index is not exported.

Fixes #9483

### DIFF
--- a/automl/google/cloud/automl_v1beta1/tables/gcs_client.py
+++ b/automl/google/cloud/automl_v1beta1/tables/gcs_client.py
@@ -132,7 +132,12 @@ class GcsClient(object):
             uploaded_csv_name = "automl-tables-dataframe-{}.csv".format(
                 int(time.time())
             )
-        csv_string = dataframe.to_csv()
+
+        # Setting index to False to ignore exporting the data index:
+        # 1. The resulting column name for the index column is empty, AutoML
+        # Tables does not allow empty column name
+        # 2. The index is not an useful training information
+        csv_string = dataframe.to_csv(index=False)
 
         bucket = self.client.get_bucket(self.bucket_name)
         blob = bucket.blob(uploaded_csv_name)
--- a/automl/tests/unit/gapic/v1beta1/test_gcs_client_v1beta1.py
+++ b/automl/tests/unit/gapic/v1beta1/test_gcs_client_v1beta1.py
@@ -139,7 +139,7 @@ class TestGcsClient(object):
 
         gcs_client.client.get_bucket.assert_called_with("my-bucket")
         mock_bucket.blob.assert_called_with("my-file.csv")
-        mock_blob.upload_from_string.assert_called_with(",col1,col2\n0,1,3\n1,2,4\n")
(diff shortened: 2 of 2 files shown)


helinwang:
Thanks, @tswast . I am not sure as I just become the maintainer for AutoML Tables' Python SDK. 
@TrucHLe @lwander any specific reasons for pandas>=0.24.0?
@tswast do you have any suggestion for pandas' version? If there is no specific reasons for pandas>=0.24.0, I am happy to send a PR to change it.

tswast:
If you want to be consistent with BigQuery, we use `0.17.1`.

https://github.com/googleapis/google-cloud-python/blob/cab728bc6b05f0feab34027ca26d347478cc29cf/bigquery/setup.py#L44

Unless you count dropping Python 2 support in 0.25, pandas 0.17 was the version with the most breaking changes in recent history, so something after that will get most customers.

helinwang:
Thanks @tswast , sent https://github.com/googleapis/google-cloud-python/pull/9824 to change it to 0.17.1 as well.

PR #9785, opened 2019-11-13
feat(bigquery): add description for routine entities

### LINKED ISSUE #9782: BigQuery: add Description for Routine entities
Quick scan of generated docs indicates that description field isn't present for Routine entities.  This field was added after the initial Routines work.

This request is to include it in the python implementation.

HemangChothani:
Fixes #9782

### DIFF
--- a/bigquery/google/cloud/bigquery/routine.py
+++ b/bigquery/google/cloud/bigquery/routine.py
@@ -50,6 +50,7 @@ class Routine(object):
         "reference": "routineReference",
         "return_type": "returnType",
         "type_": "routineType",
+        "description": "description",
     }
 
     def __init__(self, routine_ref, **kwargs):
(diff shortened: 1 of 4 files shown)


bigquery/google/cloud/bigquery/routine.py
```diff
@@ -239,6 +240,17 @@ def body(self):
     def body(self, value):
         self._properties[self._PROPERTY_TO_API_FIELD["body"]] = value
 
+    @property
+    def description(self):
+        """Union[str, None]: Description of the routine (defaults to
```
plamut: (nit)
Mypy [docs recommend](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#optional-types-and-the-none-type) using the `Optional` shorthand instead of and `Union` with `None`.

```suggestion
        """Optional[str]: Description of the routine (defaults to
```

HemangChothani: ohk, should i change same in the other files which has description property like in  dataset.py and model.py .

plamut: Strictly speaking it is not in scope of this PR, but if it's not too much work, it would be a welcome cleanup IMO. :+1:

PR #9789, opened 2019-11-13
docs(firestore): add new where operators to docstring

BenWhitehead:


tseaver:
Docs failure on CI is due to the requests docs site being down.

PR #9791, opened 2019-11-13
chore(spanner): manual synthtool update

mbril:
Updated with latest protos from cloud spanner


mbril:
@tseaver  that is great if it's automatic,
That being said, I see a few changes are missing from protos,
thus, I ran the tool manually and generated this PR.
The one I was looking for:
In Cloud Spanner, under instance_admin
I've added a repeated field called endpoint_urls under Instance message, which I don't see in the current code.

spanner/google/cloud/spanner_admin_instance_v1/proto/spanner_instance_admin.proto

see this file in this PR

skuruppu:
> @tseaver that is great if it's automatic,
> That being said, I see a few changes are missing from protos,
> thus, I ran the tool manually and generated this PR.
> The one I was looking for:
> In Cloud Spanner, under instance_admin
> I've added a repeated field called endpoint_urls under Instance message, which I don't see in the current code.
> 
> spanner/google/cloud/spanner_admin_instance_v1/proto/spanner_instance_admin.proto
> 
> see this file in this PR

My experience with this is that it's automatic but it takes a few hours for Autosynth to pick up the change. I believe it'll create a PR within a few hours.

tseaver:
@mbril The last autogen PR is #9764, which is blocked because of an invalid change to the generated `test_iam_permissions` method (see [my comment](https://github.com/googleapis/google-cloud-python/pull/9764/files#r345386632)).  I note your gen has the same issue (which isn't surprising, because the proto annotations for `TestIamPermissions` haven't changed).

The bot won't run again for spanner until that PR is merged or closed.

skuruppu:
Closing this in favor of auto-generating the code based on the proto field name change we made last week.

PR #9491, opened 2019-10-17
fix(automl): pass credentials to underlying clients in TableClient

shunghsiyu:
In automl_v1beta1.TablesClient.__init__(), the credentials are not given to AutoMlClient and PredictionServiceClient, causing inconsistency between the credentials that was passed and the credentials that was actually used for API calls.

This commit ensures that credentials are passed, and add two unittests to catch the behavior.

Fixes #9490

PR #9794, opened 2019-11-14
fix(automl): pass params to underlying client

sirtorry:
passes params from tables_client to gapic prediction_service_client

### DIFF
--- a/automl/google/cloud/automl_v1beta1/tables/tables_client.py
+++ b/automl/google/cloud/automl_v1beta1/tables/tables_client.py
@@ -25,9 +25,7 @@ from google.cloud.automl_v1beta1 import gapic
 from google.cloud.automl_v1beta1.proto import data_types_pb2
 from google.cloud.automl_v1beta1.tables import gcs_client
 
-_GAPIC_LIBRARY_VERSION = pkg_resources.get_distribution(
-    "google-cloud-automl"
-).version
+_GAPIC_LIBRARY_VERSION = pkg_resources.get_distribution("google-cloud-automl").version
 _LOGGER = logging.getLogger(__name__)
 
 
@@ -189,11 +187,7 @@ class TablesClient(object):
         region=None,
         **kwargs
     ):
-        if (
-            dataset is None
-            and dataset_display_name is None
-            and dataset_name is None
-        ):
+        if dataset is None and dataset_display_name is None and dataset_name is None:
             raise ValueError(
                 "One of 'dataset', 'dataset_name' or "
                 "'dataset_display_name' must be set."
@@ -222,8 +216,7 @@ class TablesClient(object):
     ):
         if model is None and model_display_name is None and model_name is None:
             raise ValueError(
-                "One of 'model', 'model_name' or "
-                "'model_display_name' must be set."
+                "One of 'model', 'model_name' or " "'model_display_name' must be set."
             )
         # we prefer to make a live call here in the case that the
         # model object is out-of-date
@@ -247,11 +240,7 @@ class TablesClient(object):
         region=None,
         **kwargs
     ):
-        if (
-            dataset is None
-            and dataset_display_name is None
-            and dataset_name is None
-        ):
+        if dataset is None and dataset_display_name is None and dataset_name is None:
             raise ValueError(
                 "One of 'dataset', 'dataset_name' or "
                 "'dataset_display_name' must be set."
@@ -270,10 +259,7 @@ class TablesClient(object):
         else:
             # we do this to force a NotFound error when needed
             self.get_dataset(
-                dataset_name=dataset_name,
-                project=project,
-                region=region,
-                **kwargs
+                dataset_name=dataset_name, project=project, region=region, **kwargs
             )
         return dataset_name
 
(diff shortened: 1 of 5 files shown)


automl/google/cloud/automl_v1beta1/tables/tables_client.py
```diff
@@ -2642,6 +2643,14 @@ def predict(
                 The `model` instance you want to predict with . This must be
                 supplied if `model_display_name` or `model_name` are not
                 supplied.
+            params (Dict[str, str]):
```
busunkim96:
```suggestion
            params (dict[str, str]): 
```

automl/google/cloud/automl_v1beta1/tables/tables_client.py
```diff
@@ -2642,6 +2643,14 @@ def predict(
                 The `model` instance you want to predict with . This must be
                 supplied if `model_display_name` or `model_name` are not
                 supplied.
+            params (Dict[str, str]): 
+                Additional domain-specific parameters, any string must be up to 
+                25000 characters long.
+                ``feature_importance`` - (boolean) Whether
+                [feature\_importance][[google.cloud.automl.v1beta1.TablesModelColumnInfo.feature\_importance]
+                should be populated in the returned
+                [TablesAnnotation(-s)][[google.cloud.automl.v1beta1.TablesAnnotation].
```
busunkim96: Is this copy-pasted from the proto?

sirtorry: yes. will happily edit this.

sirtorry:
> LGTM with some nits. Is this something to be watching out for every time there's an update to AutoML?

Looking at previous PRs, there's certainly a precedence at this point. #9779 #9491 are examples.

automl/google/cloud/automl_v1beta1/tables/tables_client.py
```diff
@@ -2642,6 +2643,14 @@ def predict(
                 The `model` instance you want to predict with . This must be
                 supplied if `model_display_name` or `model_name` are not
                 supplied.
+            params (Dict[str, str]): 
+                Additional domain-specific parameters, any string must be up to 
+                25000 characters long.
+                ``feature_importance`` - (boolean) Whether
+                [feature\_importance][[google.cloud.automl.v1beta1.TablesModelColumnInfo.feature\_importance]
```
helinwang: Curious if there is documentation / spec about how to how to write docstring with links.
E.g.,
- why "\" in "feature\_importance" is needed,
- what does "(-s)" mean, and
- why different from markdown, [text][[link] instead of [text][link] is used.

helinwang:
LGTM

helinwang:
Regarding "Is this something to be watching out for every time there's an update to AutoML?".
Any recommendation of how to keep backend and Python SDK in-sync?
The currently process is manually trying the best to make sure it's in-sync.

busunkim96:
@helinwang Would it be possible to write some tests to help catch these incompatibilities?

helinwang:
@busunkim96 I think it's hard to do because:
New backend feature is added in piper first. The test for these features should be added during the feature development, so they have to be in piper too. If we add the test in Github, when we are adding the test, we already know the new feature is coming, so the test does not help catching new features. It only catches regressions - still good to have, but does not solve this problem.
However, the source-of-truth of Python SDK in Github, not piper. piper code can't reference Python SDK thus have no way of testing it.
Does it makes sense?

Do you know if there is plan to mirrow this repo into piper? I think it will solve this problem. We can setup integration test that reference the piper mirror code with TAP.

sirtorry:
@busunkim96 please merge if it looks good to you

busunkim96:
You're free to setup a mirror if you'd like, but that is not something the client libraries team is currently staffed to do.

I'll make sure to tag both of you in future PRs so you can manually inspect changes.

busunkim96:
@sirtorry Could you do `nox -s blacken` to appease the linter?

automl/google/cloud/automl_v1beta1/tables/tables_client.py
busunkim96: This formatting is from the proto comments. I believe it's some flavor of markdown that makes the comments render nicely on cloud.google.com. See https://github.com/googleapis/googleapis/tree/master/google/cloud/automl/v1 and https://cloud.google.com/automl/docs/reference/rpc/google.cloud.automl.v1.

When you need to add docstrings, please follow the [Google Python Style Guide](http://google.github.io/styleguide/pyguide)

helinwang:
Thanks, @busunkim96 .

busunkim96:
I missed this earlier, but it looks like the unit tests [are failing](https://source.cloud.google.com/results/invocations/7895ad99-fa43-408a-a34c-1fcdc2919c53/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-python%2Fpresubmit%2Fautoml/log).

sirtorry:
could also add unit tests for param passing if y'all want

helinwang:
I think that's fine, since the method just forward param to predict.

PR #9796, opened 2019-11-14
fix(storage): str() metadata for for blob

### LINKED ISSUE #9729: Storage: str() metadata for bucket and blob
Surfaced by GCS team.

The Cloud Storage library must stringify metadata sent in bucket and blob.

Here's a known issue from not doing so:
```python
client = storage.Client(project="my-project")
my_bucket = client.get_bucket("my-bucket")
my_blob = my_bucket.get_blob("my-blob")
my_blob.metadata["bad"] = float('nan')
my_blob.patch()
```

`myblob.patch()` will send an unquoted `NaN`, which is a bug in the Cloud Storage API that the client should no longer depend on.

HemangChothani:
Fixes #9729

### DIFF
--- a/storage/google/cloud/storage/blob.py
+++ b/storage/google/cloud/storage/blob.py
@@ -1911,6 +1911,7 @@ class Blob(_PropertyMixin):
         :type value: dict
         :param value: (Optional) The blob metadata to set.
         """
+        value = {k: str(v) for k, v in value.items()}
         self._patch_property("metadata", value)
 
     @property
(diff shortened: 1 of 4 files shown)


storage/google/cloud/storage/blob.py
```diff
@@ -1911,6 +1911,7 @@ def metadata(self, value):
         :type value: dict
         :param value: (Optional) The blob metadata to set.
         """
+        value = {k: str(v) for k, v in value.items()}
```
frankyn: Thanks @HemangChothani!

Could you also add this to bucket labels given the similarity?

PR #9869, opened 2019-11-21
feat(api_core): support version 3 policy bindings

jkwlui:
This ~~proposal~~ change uses a list of (binding) dicts as the underlying data structure, while providing backwards compatibility for v1 Policy.

The `Policy` class is still inheriting from `MutableMapping`, but throws an exception if policy's version is `3` and the user try to access it as if it were a dict (via accessors, setter, iterator etc).

### DIFF
--- a/api_core/google/api_core/iam.py
+++ b/api_core/google/api_core/iam.py
@@ -170,22 +170,33 @@ class Policy(collections_abc.MutableMapping):
 
     @property
     def bindings(self):
-        """:obj:`list` of :obj:`dict`: The policy's bindings list.
+        """The policy's list of bindings.
 
         A binding is specified by a dictionary with keys:
-           role (str): Role that is assigned to `members`.
-           members (:obj:`set` of str): Specifies the identities associated to this binding.
-           condition (dict of str:str): Specifies a condition under which this binding will apply.
-           - title (str): Title for the condition.
-           - description (:obj:str, optional): Description of the condition.
-           - expression: A CEL expression.
+
+        * role (str): Role that is assigned to `members`.
+
+        * members (:obj:`set` of str): Specifies the identities associated to this binding.
+
+        * condition (:obj:`dict` of str:str): Specifies a condition under which this binding will apply.
+
+          * title (str): Title for the condition.
+
+          * description (:obj:str, optional): Description of the condition.
+
+          * expression: A CEL expression.
+
+        Type:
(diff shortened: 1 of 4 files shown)


api_core/google/api_core/iam.py
```diff
@@ -79,22 +84,58 @@ class Policy(collections_abc.MutableMapping):
...
+            if binding['role'] is key:
+                binding['member'] = value
+                return
+        self._bindings.append({'role': key, 'members': value})
 
     def __delitem__(self, key):
-        del self._bindings[key]
+        self.__check_version__()
+        old_length = len(self._bindings)
+        self._bindings = [b for b in self._bindings if b['role'] is not key]
```
jkwlui: should make it delete the binding in the list in-place.

jkwlui: fixed

api_core/google/api_core/iam.py
```diff
@@ -79,39 +121,120 @@ class Policy(collections_abc.MutableMapping):
...
+
+    def _contains_conditions(self):
+        for b in self._bindings:
+            if b.get("condition") is not None:
+                return True
+        return False
+
+    @property
+    def bindings(self):
+        """:obj:`list` of :obj:`dict`: The policy's bindings list.
```
jkwlui: Not sure how we'd document `dict` keys, I couldn't find any examples in the Google style guides.

plamut:
This PR broke `storage` unit tests in places where a test converts a policy to a dict ([example](https://github.com/googleapis/google-cloud-python/blob/c25d741c7534e2a69ae5493f4097889b0140802e/storage/tests/unit/test_blob.py#L2077-L2092)).

Looking at the source, the error is raised if `version > 1`, therefore a simple fix for the `storage` tests is to set VERSION to 1 (currently: 17).

**Question:** What significance does `VERSION` carry, if any? If it's just some sort of a marker, the fixture can be easily adjusted without compromising the tests' semantics.

api_core/google/api_core/iam.py
plamut: [Mypy](http://mypy-lang.org/examples.html) uses the form `Dict[key_type, value_type]`. Not sure if this is an official guideline, but at least in `BigQuery` we follow this format. I also found quite a few occurrences  of that in `firestore`.

Hope this helps!

Issue #7831, opened 2019-04-30
BigQuery: `QueryJob().done()` method gets stuck


igor47:
My environment:
```
 $ python --version
Python 3.7.1
 $ cat requirements.txt | grep bigquery
google-cloud-bigquery==1.8.0
```

here's a code sample:
```python
    while self.backgrounded_jobs_remain:
      # find one ready job
      for q in self.query_jobs:
        self.logger.debug(f"checking to see if job {q['job']} is done...")
        if q['job'].done():
          self.logger.info(f"job {q['job']} is complete! returning it.")
          return (q['job'], q['memo'])

      # if we're still here, we need to wait some more
self.logger.info(f'waiting for one of {len(self.query_jobs)} pending jobs to complete...')
```

in the usual case, the output looks something like this:
```
DEBUG datapipes.bqreader : checking to see if job <google.cloud.bigquery.job.QueryJob object at 0x7f186aac9898> is done...
DEBUG urllib3.connectionpool : https://www.googleapis.com:443 "GET /bigquery/v2/projects/myproject/queries/c14e18c5-3ef2-472e-9eab-34ea85fbc2ee?maxResults=0&timeoutMs=0&location=US HTTP/1.1" 200 None
INFO datapipes.bqreader : waiting for one of 10 pending jobs to complete...
DEBUG datapipes.bqreader : checking to see if job <google.cloud.bigquery.job.QueryJob object at 0x7f186b18feb8> is done...
DEBUG urllib3.connectionpool : https://www.googleapis.com:443 "GET /bigquery/v2/projects/myproject/queries/58d2e451-c372-41ac-acb6-395f48eff6e1?maxResults=0&timeoutMs=0&location=US HTTP/1.1" 200 None
```

however, sometimes it looks like this:
```
DEBUG datapipes.bqreader : checking to see if job <google.cloud.bigquery.job.QueryJob object at 0x7f186aac9898> is done...
DEBUG urllib3.connectionpool : https://www.googleapis.com:443 "GET /bigquery/v2/projects/myproject/queries/c14e18c5-3ef2-472e-9eab-34ea85fbc2ee?maxResults=0&timeoutMs=0&location=US HTTP/1.1" 200 None
INFO datapipes.bqreader : waiting for one of 10 pending jobs to complete...
DEBUG datapipes.bqreader : checking to see if job <google.cloud.bigquery.job.QueryJob object at 0x7f186b18feb8> is done...
```

and then the output proceeds no further.
[...]

PR #9873, opened 2019-11-21
feat(api_core): make the last retry happen at deadline

### LINKED ISSUE #9872: Core: Make the Retry objects complete "exactly" at deadline
In the [current implementation](https://github.com/googleapis/google-cloud-python/blob/58946321c7eb4059884eadd2a09cdd6b1d2604b6/api_core/google/api_core/retry.py#L193-L208), the Retry instances only check if the next delay (sleep period) happens before the deadline. If the next sleep period is long compared to the remaining time, the last retry can happen "long" after the specified deadline, which can be undesired in certain scenarios.

As an API Core user I would like to have an option to mark the deadline as strict, making the Retry object to complete retrying (approximately) at the deadline, not beyond.

plamut:
Fixes #9872.
Towards #7831.

This PR changes the behavior of Retry objects to complete retrying at the specified deadline.

Previously, Retry objects only checked if the next sleep period started before the deadline. This means that the last retry could happen considerably later than the deadline, which can be problematic if there is, say, a future sitting on top that cannot wait that long.

### PR checklist

### DIFF
--- a/api_core/google/api_core/retry.py
+++ b/api_core/google/api_core/retry.py
@@ -141,9 +141,7 @@ def exponential_sleep_generator(initial, maximum, multiplier=_DEFAULT_DELAY_MULT
         delay = delay * multiplier
 
 
-def retry_target(
-    target, predicate, sleep_generator, deadline, on_error=None, strict_deadline=False
-):
+def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
     """Call a function and retry if it fails.
 
     This is the lowest-level retry helper. Generally, you'll use the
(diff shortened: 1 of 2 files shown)


api_core/google/api_core/retry.py
```diff
@@ -156,9 +158,12 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
         sleep_generator (Iterable[float]): An infinite iterator that determines
             how long to sleep between retries.
         deadline (float): How long to keep retrying the target.
-        on_error (Callable): A function to call while processing a retryable
-            exception.  Any error raised by this function will *not* be
-            caught.
+        on_error (Callable[Exception]): A function to call while processing a
+            retryable exception.  Any error raised by this function will *not*
+            be caught.
+        strict_deadline (bool): If :data:`True`, the last retry will run at
```
tswast: What's the reason for a new argument to change the behavior versus using the "strict" logic all the time?

plamut: Backward compatibility in order to not break any existing behavior of any of the libraries.

It might be unnecessary, but played it safe here. I can still make this _the_ behavior, though, if there are no objections from anyone?

tswast: I _think_ it won't break anything, but definitely want @busunkim96 and @crwilcox's thoughts.

tswast: On second thought, if we make timeout per-request to match the semantics of the GAPIC clients, this whole change is unnecessary.

That does mean someone needs to configure both the retry arg and the timeout arg to set a total timeout, but I think consistency with GAPIC is more important than the usability improvement of having a multi-request timeout option.

plamut:
@tswast I made the "strict deadline" behavior non-optional, it's now the default logic of Retry instances.

PR #9875, opened 2019-11-21
feat(bigquery): add timeout parameter to QueryJob.done() method

### LINKED ISSUE #7831: BigQuery: `QueryJob().done()` method gets stuck

plamut:
Fixes #7831.
Requires #9873.

This is a proposed fix for the `done()` method getting stuck. It makes sure that the underlying HTTP request does not block indefinitely when retrieving query results, and that any underlying `requests` exceptions are retriable.

### TODO
- [x] Add missing / adjust existing tests.
- [x] Change the behavior of core.Retry to raise a timeout error at (approximately) the retry deadline, and not potentially much later that it. (#9873)
- [x] Add `timeout` parameter to `_http.JSONConnection.api_request()` (#9915)
- [x] Add `timeout` parameter to `google.auth.transport.AuthorizedSession.request()` that BigQuery client uses as its transport (https://github.com/googleapis/google-auth-library-python/pull/406)
- [x] Update BigQuery's `setup.py` to depend on the new `api_core` version, and the new `auth` library version (once the timeout-related changes are merged, and the new versions of those libraries released).

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -1081,7 +1081,7 @@ class Client(ClientWithProject):
                 raise
 
     def _get_query_results(
-        self, job_id, retry, project=None, timeout_ms=None, location=None
+        self, job_id, retry, project=None, timeout_ms=None, location=None, timeout=None,
     ):
         """Get the query results object for a query job.
 
@@ -1096,6 +1096,9 @@ class Client(ClientWithProject):
                 (Optional) number of milliseconds the the API call should
                 wait for the query to complete before the request times out.
             location (str): Location of the query job.
+            timeout (Optional[float]):
+                The number of seconds to wait for the underlying HTTP transport
+                before retrying the HTTP request.
 
         Returns:
             google.cloud.bigquery.query._QueryResults:
@@ -1122,7 +1125,7 @@ class Client(ClientWithProject):
         # job is complete (from QueryJob.done(), called ultimately from
         # QueryJob.result()). So we don't need to poll here.
         resource = self._call_api(
-            retry, method="GET", path=path, query_params=extra_params
+            retry, method="GET", path=path, query_params=extra_params, timeout=timeout
         )
         return _QueryResults.from_api_repr(resource)
 
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -14,6 +14,8 @@
 
 """Define API Jobs."""
 
+from __future__ import division
+
 import copy
 import re
 import threading
@@ -3029,9 +3031,9 @@ class QueryJob(_AsyncJob):
 
         if timeout_ms is not None and timeout_ms > 0:
             if timeout is not None:
-                timeout = min(timeout_ms, timeout)
+                timeout = min(timeout_ms / 1000, timeout)
             else:
-                timeout = timeout_ms
+                timeout = timeout_ms / 1000
 
         # Do not refresh is the state is already done, as the job will not
         # change once complete.
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -54,6 +54,7 @@ from google.cloud.bigquery.table import TimePartitioning
 _DONE_STATE = "DONE"
 _STOPPED_REASON = "stopped"
 _TIMEOUT_BUFFER_SECS = 0.1
+_SERVER_TIMEOUT_MARGIN_SECS = 1.0
 _CONTAINS_ORDER_BY = re.compile(r"ORDER\s+BY", re.IGNORECASE)
 
 _ERROR_REASON_TO_EXCEPTION = {
@@ -3031,11 +3032,19 @@ class QueryJob(_AsyncJob):
             self._done_timeout = max(0, self._done_timeout)
             timeout_ms = int(api_timeout * 1000)
 
+        # If the server-side processing timeout (timeout_ms) is specified and
+        # would be picked as the total request timeout, we want to add a small
+        # margin to it - we don't want to timeout the connection just as the
+        # server-side processing might have completed, but instead slightly
+        # after the server-side deadline.
+        # However, if `timeout` is specified, and is shorter than the adjusted
+        # server timeout, the former prevails.
         if timeout_ms is not None and timeout_ms > 0:
+            server_timeout_with_margin = timeout_ms / 1000 + _SERVER_TIMEOUT_MARGIN_SECS
             if timeout is not None:
-                timeout = min(timeout_ms / 1000, timeout)
+                timeout = min(server_timeout_with_margin, timeout)
             else:
-                timeout = timeout_ms / 1000
+                timeout = server_timeout_with_margin
 
         # Do not refresh is the state is already done, as the job will not
         # change once complete.
--- a/bigquery/google/cloud/bigquery/job.py
+++ b/bigquery/google/cloud/bigquery/job.py
@@ -663,7 +663,7 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
         else:
             return True
 
-    def reload(self, client=None, retry=DEFAULT_RETRY):
+    def reload(self, client=None, retry=DEFAULT_RETRY, timeout=None):
         """API call:  refresh job properties via a GET request.
 
         See
@@ -675,6 +675,9 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
                 ``client`` stored on the current dataset.
 
             retry (google.api_core.retry.Retry): (Optional) How to retry the RPC.
+            timeout (Optional[float]):
+                The number of seconds to wait for the underlying HTTP transport
+                before retrying the HTTP request.
         """
         client = self._require_client(client)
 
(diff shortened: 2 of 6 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1056,7 +1057,13 @@ def delete_table(self, table, retry=DEFAULT_RETRY, not_found_ok=False):
...
 
     def _get_query_results(
-        self, job_id, retry, project=None, timeout_ms=None, location=None
+        self,
+        job_id,
+        retry,
+        project=None,
+        timeout_ms=None,
+        location=None,
+        http_timeout=None,
```
tswast: The semantics of this match that of the `timeout` argument in GAPIC clients.

> timeout (Optional[float]) – The amount of time, in seconds, to wait for the request to complete. Note that if retry is specified, the timeout applies to each individual attempt.

We should use the same name.

bigquery/google/cloud/bigquery/retry.py
```diff
@@ -27,6 +31,50 @@
 )
 
 
+class _TimeoutableHTTPAdapter(requests.adapters.HTTPAdapter):
```
tswast: This feels like something we could put in `google-cloud-core`. I imagine it'd be relevant for our other REST-based clients.

CC @crwilcox @busunkim96

tswast: Also, could you explain why this is necessary? It appears as though requests has a `timeout` parameter already. https://2.python-requests.org//en/latest/api/#requests.request

plamut: TL; DR - it allows the timeout customization to be made from BigQuery itself, i.e. without further modifying the core, which initially seemed a good idea.

---

Deep down inside the API Core's [_http. JSONConnection._do_request()](https://github.com/googleapis/google-cloud-python/blob/b7ba9183a5384839e05afc9d26ddc4e5619ab3bf/core/google/cloud/_http.py#L268-L295), `request()` is invoked without a timeout. This `Session.request()` method then invokes [Session.send()](https://github.com/psf/requests/blob/d2590ee46c0641958b6d4792a206bd5171cb247d/requests/sessions.py#L530), which in turn obtains the HTTP adapter and [calls its send()](https://github.com/psf/requests/blob/d2590ee46c0641958b6d4792a206bd5171cb247d/requests/sessions.py#L636-L643) method with all kwargs (by default, the `timeout` is `None` in the default HTTPAdapter's [send() method](https://github.com/psf/requests/blob/d2590ee46c0641958b6d4792a206bd5171cb247d/requests/adapters.py#L394)).

Injecting a custom `timeout` seemed more easy by directly customizing the adapter from the BigQuery client library itself, rather than adding a new `timeout` parameter to several core methods and dragging it through the call stack:

**(BigQuery)**
- `Client._get_query_results()` (already accepts timeout)
- `Client._call_api()`

(retry layers omitted)

**(API core)**
- `_http.JSONConnection.api_request()`
- `_http.JSONConnection._make_request()`
- `_http.JSONConnection._do_request()`

**(requests)**
- `google.auth.transport.requests.AuthorizedSession.request()` (already accepts `**kwargs`)
- `sessions.Session.request()` (already accepts timeout)

Additionally, the same change would have to be made to the [credentials.before_request()](https://github.com/googleapis/google-auth-library-python/blob/0c33e9c0fe4f87fa46c8f1a5afe725a467ac5fcc/google/auth/transport/requests.py#L211-L213) call that eventually [requests a token](https://github.com/googleapis/google-auth-library-python/blob/08272d89667a901d0dff6c4ba53d5b30fcc29e81/google/oauth2/_client.py#L105-L106) through the `google.oauh2` client (again without an explicit timeout currently).

---


Now, if this `timeout` is something we want to add to API Core and provide to other libraries as well, then updating a few API Core `_http.JSONConnection` methods (and possibly the `oauth2` client) is probably better long-term - after all, passing an optional `timeout` parameter is cleaner than having to wrap `self._call_api()` calls into a context manager.

tswast: > Additionally, the same change would have to be made to the credentials.before_request() call that eventually requests a token through the google.oauh2 client (again without an explicit timeout currently).

Given that we've seen annoying "hang forever" behavior with network hiccups, this is worth doing across all our libraries, including the auth layer.

plamut: @tswast Can the P1 issue with `QueryJob.done()` method wait for a bit more in favor of a proper fix to API core?

(the quick fix proposed here mitigates the symptom without changes to core, provided that we are fine with temporarily not supporting the "strict" Retry mode)

plamut: To summarize, these would be the changes:
- Change the behavior of core.Retry to raise a timeout error at (approximately) the retry deadline (#9873)
- Add `timeout` parameter to `_http.JSONConnection.api_request()`
- Add `timeout` parameter to `google.auth.crendetials.Credentials.before_request()` and  `google.auth.crendetials.Credentials.refresh()`. The same for Credentials' subclasses, e.g. `google.oauth2.service_account.Credentials`
- Provide the new `timeout` parameter to all `bigquery.Client._call_api()` calls in various public methods.

plamut:
@tswast Ditched the custom HTTP adapter, and leveraged the new connection `timeout` parameter from #9915. The changes are now minimal, and can be easily applied to other client methods that would benefit from a timeout.

plamut:
A snippets test failed, seems like flakiness (it passed on several previous runs). Re-running.

bigquery/google/cloud/bigquery/job.py
```diff
@@ -3007,11 +3025,17 @@ def done(self, retry=DEFAULT_RETRY):
...
+            api_timeout = self._done_timeout - _TIMEOUT_BUFFER_SECS
+            api_timeout = max(min(api_timeout, 10), 0)
+            self._done_timeout -= api_timeout
             self._done_timeout = max(0, self._done_timeout)
-            timeout_ms = int(timeout * 1000)
+            timeout_ms = int(api_timeout * 1000)
+
+        if timeout_ms is not None and timeout_ms > 0:
+            if timeout is not None:
+                timeout = min(timeout_ms / 1000, timeout)
```
tswast: We want to give ourselves a bit of a buffer (1 second? Whatever we pick, let's make it a private constant) on this timeout. Since `timeout_ms` is the server-side timeout, not the end-to-end timeout, I don't want us to close the connection before we can download the response.

plamut: Makes sense.

One second (i.e. a small-ish buffer) sounds reasonable. It's enough to make the timeout logic, in principle, correct, as we need to wait for a bit longer than the max server-side processing  time. On the other hand, the buffer should also not be too large, because after all this case is still probably marginal (the server completing the job just before the timeout would kick in).

Issue #9809, opened 2019-11-14
Bigtable: Add table level IAM policy controls


kolea2:

crwilcox:
Java Implementation: https://github.com/googleapis/google-cloud-java/pull/6293

kolea2:
One note - `testIamPermission` should be `testIamPermissions` https://github.com/googleapis/google-cloud-java/pull/6293/files#diff-323cc5589a85618a42238a082db06902R1012

PR #9877, opened 2019-11-21
feat(bigtable): add table level IAM policy controls

Emar-Kar:
Closes: #9809

### DIFF
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -180,7 +180,7 @@ class Table(object):
         return Policy.from_pb(resp)
 
     def test_iam_permissions(self, permissions):
-        """Tests whether the caller has the given permissions for the specified table.
+        """Tests whether the caller has the given permissions for this table.
         Returns the permissions that the caller has.
 
         For example:
--- a/bigtable/google/cloud/bigtable/table.py
+++ b/bigtable/google/cloud/bigtable/table.py
@@ -140,7 +140,7 @@ class Table(object):
         )
 
     def get_iam_policy(self):
-        """Gets the access control policy for a table.
+        """Gets the IAM access control policy for this table.
 
         For example:
 
@@ -156,7 +156,7 @@ class Table(object):
         return Policy.from_pb(resp)
 
     def set_iam_policy(self, policy):
-        """Sets the access control policy on a table. Replaces any
+        """Sets the IAM access control policy for this table. Replaces any
         existing policy.
 
         For more information about policy, please see documentation of
@@ -180,7 +180,8 @@ class Table(object):
         return Policy.from_pb(resp)
 
     def test_iam_permissions(self, permissions):
-        """Returns permissions that the caller has on the specified table.
+        """Tests whether the caller has the given permissions for the specified table.
+        Returns the permissions that the caller has.
 
         For example:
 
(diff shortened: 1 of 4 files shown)


bigtable/google/cloud/bigtable/table.py
```diff
@@ -138,6 +139,73 @@ def name(self):
...
+
+        :rtype: :class:`google.cloud.bigtable.policy.Policy`
+        :returns: The current IAM policy of this table.
+        """
+        table_client = self._instance._client.table_admin_client
+        resp = table_client.set_iam_policy(resource=self.name, policy=policy.to_pb())
+        return Policy.from_pb(resp)
+
+    def test_iam_permissions(self, permissions):
+        """Returns permissions that the caller has on the specified table.
```
crwilcox: Can we include that this 'tests the permissions'? similar to https://github.com/googleapis/google-cloud-java/pull/6293/files#diff-323cc5589a85618a42238a082db06902R994

bigtable/google/cloud/bigtable/table.py
```diff
@@ -138,6 +139,73 @@ def name(self):
...
+
+        .. literalinclude:: snippets_table.py
+            :start-after: [START bigtable_table_get_iam_policy]
+            :end-before: [END bigtable_table_get_iam_policy]
+
+        :rtype: :class:`google.cloud.bigtable.policy.Policy`
+        :returns: The current IAM policy of this table.
+        """
+        table_client = self._instance._client.table_admin_client
+        resp = table_client.get_iam_policy(resource=self.name)
```
crwilcox: this sample takes a resource, but this function doesn't.

bigtable/google/cloud/bigtable/table.py
```diff
@@ -138,6 +139,73 @@ def name(self):
             project=project, instance=instance_id, table=self.table_id
         )
 
+    def get_iam_policy(self):
+        """Gets the access control policy for a table.
```
crwilcox: as authored this `Gets the IAM access control policy for this table`

bigtable/google/cloud/bigtable/table.py
```diff
@@ -138,6 +139,73 @@ def name(self):
...
+
+        :rtype: :class:`google.cloud.bigtable.policy.Policy`
+        :returns: The current IAM policy of this table.
+        """
+        table_client = self._instance._client.table_admin_client
+        resp = table_client.get_iam_policy(resource=self.name)
+        return Policy.from_pb(resp)
+
+    def set_iam_policy(self, policy):
+        """Sets the access control policy on a table. Replaces any
```
crwilcox: similarly, not a table, this table.

bigtable/google/cloud/bigtable/table.py
Emar-Kar: If I clearly understand the logic, this method creates a call of the `get_iam_policy` method from the `bigtable_table_admin_client`. That is why it is taking the `resource` as an argument. 

image

bigtable/google/cloud/bigtable/table.py
```diff
@@ -138,6 +139,74 @@ def name(self):
...
+
+        :rtype: :class:`google.cloud.bigtable.policy.Policy`
+        :returns: The current IAM policy of this table.
+        """
+        table_client = self._instance._client.table_admin_client
+        resp = table_client.set_iam_policy(resource=self.name, policy=policy.to_pb())
+        return Policy.from_pb(resp)
+
+    def test_iam_permissions(self, permissions):
+        """Tests whether the caller has the given permissions for the specified table.
```
crwilcox: Also should be change to be 'this table'

PR #9882, opened 2019-11-22
feat(bigquery): add create job method

### LINKED ISSUE #9781: BigQuery: add `create_job` method that takes any kind of job config (and public configuration property to job classes)

Some job failures are just due to current conditions and the job will succeed if started from the beginning. The problem is that this isn't a simple "retry" because the job configuration is mutable.

For example, BigQuery automatically populates a destination table if not set. In this case, the destination table should not be part of the retried request. In my opinion, this kind of logic is outside the scope of the client libraries, as it's not clear when destination table needs to be cleared just from the job resource.

- [ ] Provide a `create_job` method that takes any JobConfig object.
- [ ] Add a `configuration` property to job classes.
- [ ] Add an example of retrying a job that has failed due to `403 rateLimitExceeded` (possibly with resetting the destination table if it was a query job)

Add `.retry()` method to job classes. This is problematic, primarily because the configuration may have changed since the job was initially created, leading to unintended consequences or hard-to-debug failures.

 **Additional context**

See customer request for a method to retry any job at https://github.com/googleapis/google-cloud-python/issues/5555

HemangChothani:
Fixes #9781


bigquery/google/cloud/bigquery/client.py
```diff
@@ -1127,6 +1126,82 @@ def job_from_resource(self, resource):
...
 
+    def create_job(
+        self, job_config, source=None, destination=None, query=None, retry=DEFAULT_RETRY
+    ):
+        """Create a new job.
+        Arguments:
+            job_config (dict): configuration job representation returned from the API.
+
+        Keyword Arguments:
+            source (Union[ \
```
tswast: These arguments are quite odd to be including in a generic "create job" function.

Better to add the needed properties to the relevant -JobConfig classes.

HemangChothani:
Unit test is failing which is unrelated to these changes.

HemangChothani:
@tswast PTAL!

HemangChothani:
Suppressed by https://github.com/googleapis/python-bigquery/pull/32

PR #9889, opened 2019-11-25
feat(storage): add support for signing URLs using token

### LINKED ISSUE #7627: Storage: add support for signing URLs using IAM / tokens.
/cc @frankyn. Follow-on from PR #7460; see also issue #922.

Currently, `Blob.generate_signed_url` can only be used where the blob's client has tokens which support signing (i.e., service accounts, GAE accounts).  Although such credentials are widespread, they are not ubiquitous:  in particular, the default credentials available on GCE do not support signing.

An alternate implementation is available using the `IAM` service and access tokens.  For an outline, see the [Ruby PoC for V4 signing](https://gist.github.com/frankyn/1a537900c0689903f157740a0a9e36aa#file-v4_signed_urls-rb-L56-L68).

For this issue, please adapt that implementation for both V2 and V4 signed URLs.

HemangChothani:
Fixes #7627

### DIFF
--- a/storage/google/cloud/storage/_signing.py
+++ b/storage/google/cloud/storage/_signing.py
@@ -19,10 +19,14 @@ import collections
 import datetime
 import hashlib
 import re
+import json
 
 import six
 
 import google.auth.credentials
+
+from google.auth import exceptions
+from google.auth.transport import requests
 from google.cloud import _helpers
 
 
@@ -265,6 +269,8 @@ def generate_signed_url_v2(
     generation=None,
     headers=None,
     query_parameters=None,
+    service_account_email=None,
+    access_token=None,
 ):
     """Generate a V2 signed URL to provide query-string auth'n to a resource.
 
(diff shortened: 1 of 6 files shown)


HemangChothani:
@crwilcox system test failed due to `Permission denied` for the `IAM Service Account Cred`. can you please add this permission to the service account.

HemangChothani:
@crwilcox  PTAL

javiercornejo:
Hi guys Im pretty ignorant how to solve / report bugs. I found one with this added feature. I gonna try to solve next days (i hope). In this line https://github.com/q-logic/google-cloud-python/blob/f1f9e9ae6d0680d7864d3c771d82f7c9c795c246/storage/google/cloud/storage/_signing.py#L389 you are passing expiration as is sent but you have to use the convert function to put the value in a proper format otherwise the url is bad.

javiercornejo:
SO far Im using it like this 
`signed_url = blob.generate_signed_url(
        expiration=get_expiration_seconds_v2(expiration),
        service_account_email=service_account_email,
        access_token=response.access_token,
        version="v2")`

PR #9892, opened 2019-11-26
feat(datastore): add return query object in add filter method

### LINKED ISSUE #9876: Let google.cloud.datastore.query.add_filter return it self!
this method return an None type
original code like this

```
    def add_filter(self, property_name, operator, value):
        if self.OPERATORS.get(operator) is None:
            error_message = 'Invalid expression: "%s"' % (operator,)
            choices_message = "Please use one of: =, <, <=, >, >=."
            raise ValueError(error_message, choices_message)

        if property_name == "__key__" and not isinstance(value, Key):
            raise ValueError('Invalid key: "%s"' % value)

        self._filters.append((property_name, operator, value))
```
This makes it impossible to add filtering and fetch on one line 

plz let it return it self!
Therefore you can code like this:
`query_results = client.query(kind='Task').add_filter("name", "=", "something").fetch()`

HemangChothani:
Fixes #9876

### DIFF
--- a/datastore/google/cloud/datastore/query.py
+++ b/datastore/google/cloud/datastore/query.py
@@ -220,8 +220,8 @@ class Query(object):
           >>> from google.cloud import datastore
           >>> client = datastore.Client()
           >>> query = client.query(kind='Person')
-          >>> query.add_filter('name', '=', 'James')
-          >>> query.add_filter('age', '>', 50)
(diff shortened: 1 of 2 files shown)


HemangChothani:
Closing this PR as suppressed by [12](https://github.com/googleapis/python-datastore/pull/12)  in new repo.

PR #9898, opened 2019-11-27
spanner: add benchwrapper

jadekler:
This is functionally useless until we find out how to make SPANNER_EMULATOR_HOST
work, but it's a start.


jadekler:
cc @larkee

larkee:
Will make this PR in the Spanner repo when I have time

PR #9899, opened 2019-11-27
chore: add @larkee as CODEOWNER for spanner

jadekler:


jadekler:
cc @larkee

jadekler:
cc @crwilcox

PR #9902, opened 2019-11-28
spanner: support SPANNER_EMULATOR_HOST

jadekler:


jadekler:
@larkee

larkee:
I looked into this some more and the admin clients do not support having both transport and credentials set. So you will also need to set credentials to None when the environment variable is set.

spanner/google/cloud/spanner_v1/client.py
```diff
@@ -51,6 +61,11 @@
 )
 
 
+if os.getenv('SPANNER_EMULATOR_HOST') != None and \
+    ('http://' in os.getenv('SPANNER_EMULATOR_HOST') or 'https://' in os.getenv('SPANNER_EMULATOR_HOST')):
```
bvandiver: Worth an abstraction like get_spanner_emulator_host() as opposed to bare getenv calls?

larkee: Done.

jadekler:
@larkee as I'm rolling off client libraries, this will need to be picked up by someone else. Could you help find the appropriate person to do so?

larkee:
> @larkee as I'm rolling off client libraries, this will need to be picked up by someone else. Could you help find the appropriate person to do so?

I'll pick it up. Let me know if you'd rather I opened a new PR rather than working on this one.

jadekler:
> > @larkee as I'm rolling off client libraries, this will need to be picked up by someone else. Could you help find the appropriate person to do so?
> 
> I'll pick it up. Let me know if you'd rather I opened a new PR rather than working on this one.

Whichever is your preference! :)

PR #9911, opened 2019-12-03
fix(pubsub): include request overhead when computing publish batch size overflow

### LINKED ISSUE #7108: PubSub: check batch max bytes against request byte size

`google.cloud.pubsub_v1.publisher._batch.thread.Batch` enforces max bytes on the sum of [`PubsubMessage.ByteSize`](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py#L284) for each message in the batch, but the `PublishRequest` created by `Batch.client.publish` is larger than that. As a result I need to specify a non default max bytes to guarantee batches create valid requests.

Enforce max bytes on the size of the `PublishRequest` created by `Batch.client.publish`:
```diff
diff --git a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
index f187024b7c..67fc04841a 100644
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -76,7 +76,7 @@ class Batch(base.Batch):
         # any writes to them use the "state lock" to remain atomic.
         self._futures = []
         self._messages = []
-        self._size = 0
+        self._size = types.PublishRequest(topic=topic, messages=[]).ByteSize()
         self._status = base.BatchStatus.ACCEPTING_MESSAGES

         # If max latency is specified, start a thread to monitor the batch and
@@ -281,7 +281,7 @@ class Batch(base.Batch):
             if not self.will_accept(message):
                 return future

-            new_size = self._size + message.ByteSize()
+            new_size = self._size + types.PublishRequest(messages=[message]).ByteSize()
             new_count = len(self._messages) + 1
             overflow = (
                 new_size > self.settings.max_bytes
```

I can set max bytes to guarantee sufficient overhead, but I feel like it would be better if I didn't need to and this may result in fewer batches.

**Additional context**

see also #7107, which suggests adding an option to enforce this setting even when the batch is empty.

plamut:
Fixes #7108. 

This PR fixes the logic that computes the publish batch size overflow, taking the total request message size overhead into account. The improved logic prevents the server-side errors simular to the following:

> google.api_core.exceptions.InvalidArgument: 400 The value for request_size is too large. You passed 10000096 in the request, but the maximum value is 10000000.

### How to test

(see also the system test in this PR)

- Create a pubisher client with `BatchSettings.max_bytes` substantially larger than 10_000_000, and `BatchSettings.max_latency` to one second (so that the publish autocommit does not kick in too soon).
- Quickly publish a few sizable messages to a topic. Their total size should slightly exceed 10_000_000 bytes.

The backend responds with a "400 InvalidArgument" error.

All messages are successfully published (the code splits them into multiple publish batches).

### PR checklist

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -313,7 +313,7 @@ class Batch(base.Batch):
             new_size = self._size + size_increase
             new_count = len(self._messages) + 1
 
-            size_limit = min(self.settings.max_bytes, _SERVER_PUBLISH_MAX_BYTES)
+            size_limit = min(self._settings.max_bytes, _SERVER_PUBLISH_MAX_BYTES)
             overflow = new_size > size_limit or new_count >= self._settings.max_messages
 
             if not self._messages or not overflow:
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -29,6 +29,7 @@ from google.cloud.pubsub_v1.publisher._batch import base
 
 _LOGGER = logging.getLogger(__name__)
 _CAN_COMMIT = (base.BatchStatus.ACCEPTING_MESSAGES, base.BatchStatus.STARTING)
+_SERVER_PUBLISH_MAX_BYTES = 10 * 1000 * 1000  # max accepted size of PublishRequest
 
 
 class Batch(base.Batch):
@@ -79,8 +80,12 @@ class Batch(base.Batch):
         # in order to avoid race conditions
         self._futures = []
         self._messages = []
-        self._size = 0
         self._status = base.BatchStatus.ACCEPTING_MESSAGES
+
+        # The initial size is not zero, we need to account for the size overhead
+        # of the PublishRequest message itself.
+        self._base_request_size = types.PublishRequest(topic=topic).ByteSize()
+        self._size = self._base_request_size
 
         # If max latency is specified, start a thread to monitor the batch and
         # commit when the max latency is reached.
@@ -124,9 +129,12 @@ class Batch(base.Batch):
     def size(self):
         """Return the total size of all of the messages currently in the batch.
 
+        The size includes any overhead of the actual ``PublishRequest`` that is
+        sent to the backend.
+
         Returns:
             int: The total size of all of the messages currently
-                 in the batch, in bytes.
+                 in the batch (including the request overhead), in bytes.
         """
         return self._size
 
@@ -292,12 +300,21 @@ class Batch(base.Batch):
             if not self.will_accept(message):
                 return future
 
-            new_size = self._size + message.ByteSize()
+            size_increase = types.PublishRequest(messages=[message]).ByteSize()
+
+            if (self._base_request_size + size_increase) > _SERVER_PUBLISH_MAX_BYTES:
+                err_msg = (
+                    "The message being published would produce too large a publish "
+                    "request that would exceed the maximum allowed size on the "
+                    "backend ({} bytes).".format(_SERVER_PUBLISH_MAX_BYTES)
+                )
+                raise ValueError(err_msg)
+
+            new_size = self._size + size_increase
             new_count = len(self._messages) + 1
-            overflow = (
-                new_size > self.settings.max_bytes
-                or new_count >= self._settings.max_messages
-            )
+
+            size_limit = min(self.settings.max_bytes, _SERVER_PUBLISH_MAX_BYTES)
+            overflow = new_size > size_limit or new_count >= self._settings.max_messages
 
             if not self._messages or not overflow:
 
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -289,6 +289,10 @@ class Batch(base.Batch):
             the :class:`~concurrent.futures.Future` interface or :data:`None`.
             If :data:`None` is returned, that signals that the batch cannot
             accept a message.
+
+        Raises:
+            pubsub_v1.publisher.exceptions.MessageTooLargeError: If publishing
+                the ``message`` would exceed the max size limit on the backend.
         """
         # Coerce the type, just in case.
         if not isinstance(message, types.PubsubMessage):
@@ -308,7 +312,7 @@ class Batch(base.Batch):
                     "request that would exceed the maximum allowed size on the "
                     "backend ({} bytes).".format(_SERVER_PUBLISH_MAX_BYTES)
                 )
-                raise ValueError(err_msg)
+                raise exceptions.MessageTooLargeError(err_msg)
 
             new_size = self._size + size_increase
             new_count = len(self._messages) + 1
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -90,7 +90,7 @@ class Batch(base.Batch):
         # If max latency is specified, start a thread to monitor the batch and
         # commit when the max latency is reached.
         self._thread = None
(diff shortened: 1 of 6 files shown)


pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -292,12 +300,21 @@ def publish(self, message):
...
+
+            new_size = self._size + size_increase
             new_count = len(self._messages) + 1
-            overflow = (
-                new_size > self.settings.max_bytes
-                or new_count >= self._settings.max_messages
-            )
+
+            size_limit = min(self.settings.max_bytes, _SERVER_PUBLISH_MAX_BYTES)
+            overflow = new_size > size_limit or new_count >= self._settings.max_messages
```
software-dov: I'd just like to take a minute to register my disgust at having both `self.settings` and `self._settings`.

pubsub/tests/system.py
```diff
@@ -92,6 +92,37 @@ def test_publish_messages(publisher, topic_path, cleanup):
...
+
+    publisher.batch_settings = types.BatchSettings(
+        max_bytes=11 * 1000 * 1000,  # more than the server limit of 10 ** 7
+        max_latency=2.0,  # so that autocommit happens after publishing all messages
+        max_messages=100,
+    )
+    publisher.create_topic(topic_path)
+
+    for index in six.moves.range(5):
+        futures.append(publisher.publish(topic_path, msg_data, num=str(index)))
```
software-dov: `futures = [publisher.publish(topic_path, msg_data, num=str(i)) for i in range(5)]`

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -34,13 +34,15 @@ def create_client():
...
 
-def create_batch(autocommit=False, **batch_settings):
+def create_batch(autocommit=False, topic="topic_name", **batch_settings):
     """Return a batch object suitable for testing.
 
     Args:
         autocommit (bool): Whether the batch should commit after
             ``max_latency`` seconds. By default, this is ``False``
             for unit testing.
+        autocommit (topic): The name of the topic the batch should publish
```
software-dov: This looks wrong. Should it not be `topic (str): ...`?

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -314,22 +316,28 @@ def test_publish():
...
-    expected_size = sum([message_pb.ByteSize() for message_pb in messages])
-    assert batch.size == expected_size
+    # The size should have been incremented by the sum of the size
+    # contributions of each message to the PublishRequest.
+    base_request_size = types.PublishRequest(topic="topic_foo").ByteSize()
+    msg_size_overheads = [
+        types.PublishRequest(messages=[msg]).ByteSize() for msg in messages
+    ]
+
+    expected_request_size = base_request_size + sum(msg_size_overheads)
```
software-dov: Just do the comprehension in the sum and cut out the `msg_size_overheads` variable.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
plamut: I can fix that, too, along the way, some kind of a relic of the past.

software-dov: Since this code is under change anyway, can you change `self._settings` to `self.settings`?

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -361,6 +369,51 @@ def test_publish_exceed_max_messages():
...
+        topic="topic_foo", messages=[big_message]
+    ).ByteSize()
+    assert request_size == 1001  # sanity check, just above the (mocked) server limit
+
+    with pytest.raises(ValueError) as error:
+        batch.publish(big_message)
+
+    error_msg = str(error)
+    assert "too large" in error_msg
+    assert "1000 bytes" in error_msg
```
software-dov: Checking for specific strings in error messages is fragile and can lead to unexpected test flake during future refactors. Catching the error should be enough, no?

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -361,6 +369,51 @@ def test_publish_exceed_max_messages():
...
+
+    messages = (
+        types.PubsubMessage(data=b"x" * 500),
+        types.PubsubMessage(data=b"x" * 600),
+    )
+
+    # Sanity check - request size is still below BatchSettings.max_bytes,
+    # but it exceeds the server-side size limit.
+    request_size = types.PublishRequest(topic="topic_foo", messages=messages).ByteSize()
+    assert 1000 < request_size < 1500
```
software-dov: Didn't know you could do this. Neat!

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
plamut: Somebody throw a rubber chicken at me, that was indeed the original intention, yes.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
plamut: I personally find it beneficial to give some constructs a concrete name, can improve readability of a test.\

Unless you find the current version unpleasantly verbose?

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
plamut: Since `ValueError` is somewhat general and can, in principle, be raised elsewhere, too, the test (heuristically) tries to check that it is indeed "our" ValueError, and that its message is informative.

I agree its not ideal, hence using very short snippets in the error message, but I still see value in it. I can still remove that part, though, if the concern over future flakes is strong?

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
plamut: Yes.
I.
Can.
:+1:

software-dov: Thanks!

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
software-dov: 🐔

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
software-dov: I agree, lifting constructs can greatly improve readability, reduce the rightward march, and makes pdb based debugging _way_ easier. I'm also a big proponent of dataflow driven programming, though, and balancing those forces can be tricky sometimes.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
software-dov: A pattern I like to use is to make a big hierarchy of exception types and throw/catch based on type. It's more amenable to refactoring and makes the code a little more self documenting (IMO).

Something like
```python
class TooLargeMessageError(ValueError):
   pass

with pytest.raises(TooLargeMessageError) as e:
   # do a thing
```

plamut: I like that pattern, too, will incorporate it and then we can ditch the assertions about the error message in the test.

software-dov: 👍

plamut:
@software-dov Thanks for the quick review, addressed the last outstanding comment.

PR #9915, opened 2019-12-04
feat(core): add timeout param to JSONConnection.api_request()

### LINKED ISSUE #9914: Core: Add timeout parameter to connection's api_request() method
As a client library developer, I want to have an easy way of specifying the timeout at the transport level for HTTP requests.

**Motivation:**
If the transport layer makes an HTTP request that gets stalled, it can block the client. Any timeouts and retries on top of it cannot come into play, because the transport layer blocks them while waiting for the response without any timeout specified.

One of the manifestations of the problem in the BigQuery client: https://github.com/googleapis/google-cloud-python/issues/7831

plamut:
Fixes #9914.
Towards #7831.

As [discussed](https://github.com/googleapis/google-cloud-python/pull/9875#discussion_r351504480) on a related PR, a "clients gets stuck" fix needs a timeout in the transport layer where the blockage randomly occurs occasionally.

Keeping the existing default behavior the same for now (`timeout=None`), but if there is a "good" default timeout value that suits all client libraries, we can make the requests timeout by default to avoid randomly blocking indefinitely.

### PR checklist

### DIFF
--- a/core/google/cloud/_http.py
+++ b/core/google/cloud/_http.py
@@ -222,6 +222,7 @@ class JSONConnection(Connection):
         content_type=None,
         headers=None,
         target_object=None,
+        timeout=None,
     ):
         """A low level method to send a request to the API.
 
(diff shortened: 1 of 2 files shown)


tswast:
I think this should be a pretty uncontroversial change, especially since it only affects a few handwritten clients, so I'm going to merge it.

crwilcox:
This has broken the tests for Google Cloud Storage.

Issue #6197, opened 2018-10-11
BigQuery:  add retry for query results iteration


tseaver:
See: #6136, and the [TODO left in the code](https://github.com/googleapis/google-cloud-python/blob/4da8bd9371bf9c751a69419fd36c8d56ae42088d/bigquery/google/cloud/bigquery/job.py#L698) by @jba:

```python
    def result(self, timeout=None):
        ...
        if self.state is None:
            self._begin()
        # TODO: modify PollingFuture so it can pass a retry argument to done().
        return super(_AsyncJob, self).result(timeout=timeout)
```

tseaver:
Progress to date:
- #6301 is a user-reported bug for this same issue.
- #6305 broadened the set of exceptions retried by the underlying future.
- #6302 actually adds a `retry` argument to `_AsyncJob.result()`, but can only pass it through to the `_begin` call unless we decide to extend the future's interface.

PR #9923, opened 2019-12-06
feat(api_core): add retry param into PollingFuture() and it's inheritors

IlyaFaer:
Towards #6197

Looks like it's the last thing that should be done for this issue. CMIIW.


api_core/google/api_core/future/polling.py
```diff
@@ -66,9 +66,12 @@ def __init__(self, retry=DEFAULT_RETRY):
         self._done_callbacks = []
 
     @abc.abstractmethod
-    def done(self):
+    def done(self, retry=DEFAULT_RETRY):
         """Checks to see if the operation is complete.
 
+        Args:
+            retry (google.api_core.retry.Retry): (Optional) How to retry the RPC.
+
```
IlyaFaer: Adding comment to abstract

api_core/tests/unit/test_operation.py
```diff
@@ -113,6 +116,23 @@ def test_result():
     assert future.done()
 
 
+def test_done_w_retry():
+    RETRY_PREDICATE = retry.if_exception_type(exceptions.TooManyRequests)
+    test_retry = retry.Retry(predicate=RETRY_PREDICATE)
```
IlyaFaer: Instancing custom retry to avoid collision with default value and to make check clear

api_core/google/api_core/operation.py
```diff
@@ -145,21 +145,28 @@ def _set_result_from_operation(self):
...
+        """Refresh the operation and update the result if needed.
+
+        Args:
+            retry (google.api_core.retry.Retry): (Optional) How to retry the RPC.
+        """
         # If the currently cached operation is done, no need to make another
         # RPC as it will not change once done.
         if not self._operation.done:
-            self._operation = self._refresh()
+            self._operation = self._refresh(retry=retry)
```
IlyaFaer: Adding `retry` into inheritor of the `PollingFuture()` (the one above) and passing it further. As `self._refresh` is `functools.partial`, we can just pass `retry` as a named arg - it'll be added to those args which were prepared while creating `functools.partial`

IlyaFaer:
Some BigQuery tests failed, but, as I see, they are failing in pure `master` too.

PR #9929, opened 2019-12-06
feat(pubsub): publish-side changes for ordering keys

pradn:
- Create a new concept: sequencers. A sequencer orders messages published by the publishing client. Messages with ordering keys use the OrderedSequencer. Those without ordering keys use the UnorderedSequencer, which behaves just like before.
- The  OrderedSequencer is paused when it encounters a permanent error. It must be unpaused manually by the user calling client.resumePublish() to continue sending messages for that ordering key. This is done to prevent out-of-order sending.
- Move the commit thread to the publisher client. This one thread commits batches in all sequencers. Before, each batch had its own commit thread.
- Clean up ordering-key enabled sequencers after all messages are sent. (If paused, the OrderedSequencer is not cleaned-up.)
- Added tests for all the changes.

Implements half of #9928

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -191,7 +191,11 @@ class Batch(base.Batch):
             else:
                 return
 
-        # Start a new thread to actually handle the commit.
+        self._start_commit_thread()
+
+    def _start_commit_thread(self):
+        """Start a new thread to actually handle the commit."""
+
         commit_thread = threading.Thread(
             name="Thread-CommitBatchPublisher", target=self._commit
         )
--- a/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -62,15 +62,23 @@ class Batch(base.Batch):
         settings (~.pubsub_v1.types.BatchSettings): The settings for batch
             publishing. These should be considered immutable once the batch
             has been opened.
-        autocommit (bool): Whether to autocommit the batch when the time
-            has elapsed. Defaults to True unless ``settings.max_latency`` is
-            inf.
+        batch_done_callback (Callable[[bool], Any]): Callback called when the
+            response for a batch publish has been received. Called with one
+            boolean argument: successfully published or a permanent error
+            occurred. Temporary errors are not surfaced because they are retried
+            at a lower level.
+        commit_when_full (bool): Whether to commit the batch when the batch
+            is full.
     """
 
-    def __init__(self, client, topic, settings, autocommit=True):
+    def __init__(
+        self, client, topic, settings, batch_done_callback=None, commit_when_full=True
+    ):
         self._client = client
         self._topic = topic
         self._settings = settings
+        self._batch_done_callback = batch_done_callback
+        self._commit_when_full = commit_when_full
 
         self._state_lock = threading.Lock()
         # These members are all communicated between threads; ensure that
@@ -86,15 +94,6 @@ class Batch(base.Batch):
         # of the PublishRequest message itself.
         self._base_request_size = types.PublishRequest(topic=topic).ByteSize()
         self._size = self._base_request_size
-
-        # If max latency is specified, start a thread to monitor the batch and
-        # commit when the max latency is reached.
-        self._thread = None
-        if autocommit and self.settings.max_latency < float("inf"):
-            self._thread = threading.Thread(
-                name="Thread-MonitorBatchPublisher", target=self.monitor
-            )
-            self._thread.start()
 
     @staticmethod
     def make_lock():
@@ -148,6 +147,27 @@ class Batch(base.Batch):
         """
         return self._status
 
+    def cancel(self, cancellation_reason):
+        """Complete pending futures with an exception.
+
+        This method must be called before publishing starts (ie: while the
+        batch is still accepting messages.)
+
+        Args:
+            cancellation_reason (BatchCancellationReason): The reason why this
+                batch has been cancelled.
+        """
+
+        with self._state_lock:
+            assert (
+                self._status == base.BatchStatus.ACCEPTING_MESSAGES
+            ), "Cancel should not be called after sending has started."
+
+            exc = RuntimeError(cancellation_reason.value)
+            for future in self._futures:
+                future.set_exception(exc)
+            self._status = base.BatchStatus.ERROR
+
     def commit(self):
         """Actually publish all of the messages on the active batch.
 
@@ -162,6 +182,7 @@ class Batch(base.Batch):
         If the current batch is **not** accepting messages, this method
         does nothing.
         """
+
         # Set the status to "starting" synchronously, to ensure that
         # this batch will necessarily not accept new messages.
         with self._state_lock:
@@ -195,7 +216,10 @@ class Batch(base.Batch):
                 # If, in the intervening period between when this method was
                 # called and now, the batch started to be committed, or
                 # completed a commit, then no-op at this point.
-                _LOGGER.debug("Batch is already in progress, exiting commit")
+                _LOGGER.debug(
+                    "Batch is already in progress or has been cancelled, "
+                    "exiting commit"
+                )
                 return
 
         # Once in the IN_PROGRESS state, no other thread can publish additional
@@ -215,15 +239,23 @@ class Batch(base.Batch):
         # Log how long the underlying request takes.
         start = time.time()
 
+        batch_transport_succeeded = True
         try:
+            # Performs retries for errors defined in retry_codes.publish in the
+            # publisher_client_config.py file.
             response = self._client.api.publish(self._topic, self._messages)
         except google.api_core.exceptions.GoogleAPIError as exc:
-            # We failed to publish, set the exception on all futures and
-            # exit.
+            # We failed to publish, even after retries, so set the exception on
+            # all futures and exit.
             self._status = base.BatchStatus.ERROR
 
             for future in self._futures:
                 future.set_exception(exc)
+
+            batch_transport_succeeded = False
+            if self._batch_done_callback is not None:
+                # Failed to publish batch.
+                self._batch_done_callback(batch_transport_succeeded)
 
             _LOGGER.exception("Failed to publish %s messages.", len(self._futures))
             return
@@ -250,26 +282,17 @@ class Batch(base.Batch):
             for future in self._futures:
                 future.set_exception(exception)
 
+            # Unknown error -> batch failed to be correctly transported/
+            batch_transport_succeeded = False
+
             _LOGGER.error(
                 "Only %s of %s messages were published.",
                 len(response.message_ids),
                 len(self._futures),
             )
 
-    def monitor(self):
-        """Commit this batch after sufficient time has elapsed.
-
-        This simply sleeps for ``self.settings.max_latency`` seconds,
-        and then calls commit unless the batch has already been committed.
-        """
-        # NOTE: This blocks; it is up to the calling code to call it
-        #       in a separate thread.
-
-        # Sleep for however long we should be waiting.
-        time.sleep(self.settings.max_latency)
-
-        _LOGGER.debug("Monitor is waking up")
-        return self._commit()
+        if self._batch_done_callback is not None:
+            self._batch_done_callback(batch_transport_succeeded)
 
     def publish(self, message):
         """Publish a single message.
@@ -294,6 +317,7 @@ class Batch(base.Batch):
             pubsub_v1.publisher.exceptions.MessageTooLargeError: If publishing
                 the ``message`` would exceed the max size limit on the backend.
         """
+
         # Coerce the type, just in case.
         if not isinstance(message, types.PubsubMessage):
             message = types.PubsubMessage(**message)
@@ -301,6 +325,10 @@ class Batch(base.Batch):
         future = None
 
         with self._state_lock:
+            assert (
+                self._status != base.BatchStatus.ERROR
+            ), "Publish after stop() or publish error."
+
             if not self.will_accept(message):
                 return future
 
@@ -333,7 +361,7 @@ class Batch(base.Batch):
 
         # Try to commit, but it must be **without** the lock held, since
         # ``commit()`` will try to obtain the lock.
-        if overflow:
+        if self._commit_when_full and overflow:
             self.commit()
 
         return future
--- a/pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
+++ b/pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
@@ -27,6 +27,27 @@ class Sequencer(object):
 
     @staticmethod
     @abc.abstractmethod
+    def is_finished(self):
+        """ Whether the sequencer is finished and should be cleaned up.
+
+            Returns:
+                bool: Whether the sequencer is finished and should be cleaned up.
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    @abc.abstractmethod
+    def unpause(self, message):
+        """ Unpauses this sequencer.
+
+        Raises:
+            RuntimeError:
+                If called when the sequencer has not been paused.
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    @abc.abstractmethod
     def publish(self, message):
         """ Publish message for this ordering key.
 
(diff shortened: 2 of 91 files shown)


pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -167,44 +214,62 @@ def target(self):
...
+    def _get_or_create_sequencer(self, topic, ordering_key):
+        """ Gets an existing sequencer or creates a new one given the (topic,
+            ordering_key) pair.
+        """
+        sequencer_key = (topic, ordering_key)
+        sequencer = self._sequencers.get(sequencer_key)
+        if sequencer is None:
+            if ordering_key == "":
+                # TODO: Consider implementing _delete_sequencer for unordered
+                # sequencer.
```
plamut: Is this TODO something essential that was planned for this PR, or is it more a reminder for the future?

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -29,17 +32,30 @@
 from google.cloud.pubsub_v1.gapic import publisher_client
 from google.cloud.pubsub_v1.gapic.transports import publisher_grpc_transport
 from google.cloud.pubsub_v1.publisher._batch import thread
-
+from google.cloud.pubsub_v1.publisher._sequencer import ordered_sequencer
+from google.cloud.pubsub_v1.publisher._sequencer import unordered_sequencer
```
plamut: **Idea:**
We could also import sequencer classes in `_sequencer/__init__.py`, and then refer to them as `_sequencer.OrderSequencer`, `_sequencer.UnorderedSequencer`, etc.,

This is less verbose than the current `ordered_sequencer.OrderedSequencer` and `unordered_sequencer.UnorderedSequencer`.

It's a minor thing, though, and by no means necessary.

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -143,6 +189,7 @@ def from_service_account_file(cls, filename, batch_settings=(), **kwargs):
 
         Args:
             filename (str): The path to the service account private key json
+
```
plamut: (nit)
Seems like a redundant blank line in the middle of the description?

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -268,21 +347,56 @@ def publish(self, topic, data, **attrs):
...
+    def ensure_commit_timer_runs(self):
+        """ Ensures a commit timer thread is running. If a commit timer thread
+            is already running, it does nothing.
+        """
+        with self._batch_lock:
+            self._ensure_commit_timer_runs_no_lock()
+
+    def _ensure_commit_timer_runs_no_lock(self):
+        if not self._commit_thread and self.batch_settings.max_latency < float("inf"):
```
plamut: Just as a precaution, let's add an explicit comment that this method assumes that its caller has acquired the `_batch_lock`.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -148,6 +147,20 @@ def status(self):
...
         return self._status
 
+    def cancel(self):
+        """Completes pending futures with an exception. This method must be
+           called before publishing starts (ie: while the batch is still
+           accepting messages.)
+        """
+
+        # Cancel should not be called after sending has started.
+        assert self._status == base.BatchStatus.ACCEPTING_MESSAGES
```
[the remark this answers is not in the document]
pradn: Good idea, will add a descriptive error message. Yes, it should be behind the _state_lock.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -256,20 +277,8 @@ def _commit(self):
...
-        # NOTE: This blocks; it is up to the calling code to call it
-        #       in a separate thread.
-
-        # Sleep for however long we should be waiting.
-        time.sleep(self.settings.max_latency)
-
-        _LOGGER.debug("Monitor is waking up")
-        return self._commit()
+        if self._batch_done_callback is not None:
+            self._batch_done_callback(batch_transport_succeeded)
```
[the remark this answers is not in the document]
pradn: Found this bug independently and fixed it b/c of the 100% coverage requirement. Good eye!

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -29,17 +32,30 @@
...
     "from_service_account_json",
 )
 
 
+def _set_nested_value(container, value, keys):
+    current = container
+    for key in keys[:-1]:
+        if current.get(key) is None:
+            current[key] = {}
+        current = current[key]
```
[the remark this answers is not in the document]
pradn: Great idea, thanks!

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: It's for the future. I didn't want to add this feature because I'm trying to keep the old code as identical as possible. The old code did not have clean-up for the state per topic.

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: I like this current way for uniformity.

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -167,44 +211,63 @@ def target(self):
...
-            )
-            self._batches[topic] = batch
+        with self._batch_lock:
+            if self._is_stopped:
+                raise RuntimeError("Cannot resume publish on a stopped publisher.")
 
-        return batch
+            sequencer_key = (topic, ordering_key)
+            sequencer = self._sequencers.get(sequencer_key)
+            if sequencer is None:
```
kamalaboulhosn: Is it necessary for this to be an error? Seems like it is reasonable just to silently let it through in this case.

kamalaboulhosn: Is it necessary for this to be an error? Seems like it is reasonable just to silently let it through in this case. We want to be able to handle the case where one publishes messages for a key, there are errors, and then they never resume for the key or publish on it again.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -148,6 +147,23 @@ def status(self):
         """
         return self._status
 
+    def cancel(self):
```
kamalaboulhosn: It would be nice to actually pass in a message regarding why cancelation occurred. When it is due to a failure to publish an earlier message for an ordering key, we should surface that to the user.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
```diff
@@ -0,0 +1,56 @@
...
+    """The base class for sequencers for Pub/Sub publishing. A sequencer
+       sequences messages to be published.
+    """
+
+    @staticmethod
+    @abc.abstractmethod
+    def publish(self, message):
+        """ Publish message for this ordering key.
+
+        Returns:
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
```diff
@@ -0,0 +1,56 @@
...
+            A class instance that conforms to Python Standard library's
+            :class:`~concurrent.futures.Future` interface (but not an
+            instance of that class). The future might return immediately with a
+            PublishToPausedOrderingKeyException if the ordering key is paused.
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
```diff
@@ -0,0 +1,241 @@
...
+
+import concurrent.futures as futures
+import threading
+
+from google.cloud.pubsub_v1.publisher import exceptions
+from google.cloud.pubsub_v1.publisher._sequencer import base
+
+
+class _OrderedSequencerStatus(object):
+    """An enum-like class representing valid statuses for an OrderedSequencer.
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
```diff
@@ -0,0 +1,241 @@
...
+        self._client = client
+        self._topic = topic
+        self._ordering_key = ordering_key
+        # Guards the variables below
+        self._state_lock = threading.Lock()
+        self._publishes_done_callback = publishes_done_callback
+        # Batches ordered from first (index 0) to last.
+        # Invariant: always has at least one batch after the first publish,
+        # unless paused or stopped.
+        self._ordered_batches = []
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
```diff
@@ -0,0 +1,112 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.cloud.pubsub_v1.publisher._sequencer import base
+
+
+class UnorderedSequencer(base.Sequencer):
+    """ Sequences messages into batches for one topic without any ordering.
+
+        Public methods are NOT thread-safe.
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def stop(self):
+        """ Stop the sequencer. Subsequent publishes will fail.
+
+            Raises:
+                RuntimeError:
+                    If called after stop() has already been called.
+        """
+        if self._stopped:
+            raise RuntimeError("Ordered sequencer already stopped.")
```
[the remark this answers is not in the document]
plamut: Bump.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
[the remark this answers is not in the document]
pradn: Oops, fixed.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/base.py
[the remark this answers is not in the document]
pradn: Fixed.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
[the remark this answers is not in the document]
pradn: Done

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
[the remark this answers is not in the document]
pradn: Switched to a deque.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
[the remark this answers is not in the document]
pradn: Ordered sequencers need thread-safety because there's a commit thread in thread.Batch that calls back into it.

Unordered sequencers call batch.commit() and forget about it, so they don't need thread-safety. I mainly omitted locks here to reduce locking costs.

pubsub/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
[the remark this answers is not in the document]
pradn: Fixed.

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: The scenario you describe is already allowed and isn't related to this error. If a batch on a key errors out, the ordered_sequencer for that topic/key is marked paused. Messages with other topic/key combos will get in just fine.

It's odd to resumePublish() for a topic/ordering_key pair that hasn't been seen before right? I think that should be an error.

pubsub/google/cloud/pubsub_v1/publisher/_batch/thread.py
pradn: Done.

pubsub/tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -34,46 +35,34 @@ def create_client():
...
-            ``max_latency`` seconds. By default, this is ``False``
-            for unit testing.
-        topic (str): The name of the topic the batch should publish
-            the messages to.
+        topic (str): Topic name.
         batch_settings (dict): Arguments passed on to the
             :class:``~.pubsub_v1.types.BatchSettings`` constructor.
+        commit_when_full (bool): Whether to commit the batch when the batch
+            has reached byte-size or number-of-messages limits.
+        batch_settings (Mapping[str, str]): Batching-related settings.
```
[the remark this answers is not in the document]
pradn: Oops, fixed.

pubsub/google/cloud/pubsub_v1/publisher/client.py
kamalaboulhosn: I think what I'm getting at is that we should delete the sequencer for that ordering key. Otherwise, if one never calls resume, then we will keep the sequencer around indefinitely, essentially leaking the memory.

pradn: Well, I'm using an empty sequencer as a sentinel to mark that the ordering key has been paused. If we deleted the sequencer, we wouldn't be able to distinguish between a paused key and a key that's never been seen before. Yes, the amount of data in the publisher client grows, but it's necessary to implement pause/resume, so I would not call it a leak.

kamalaboulhosn: I guess it's an artifact of how this is implemented in that case. I still don't think it needs to be an error. I think of it as a no-op. This is the approach we've taken in other languages and so I think we should be consistent here.

pradn: Okay, made the changes:
1) Do NOT throw if an unknown topic-key pair is passed in. Instead log it.
2) Do throw if resume_publish is called without message ordering enabled.

plamut:
What's the status of this? I see that error handling was changed in one particular case, and that a deadlock was fixed.

@pradn Would you mind elaborating a bit on the latter? Looking at the state machine and the commit message, it appears to be something with publishing in a STOPPED state, and a new state FINISHED had to be introduced? Or why the deadlock could occur? Thanks!

pubsub/google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -407,6 +396,13 @@ def _wait_and_commit_sequencers(self):
             self._commit_thread = None
 
     def _commit_sequencers(self):
+        """ Clean up finished sequencers and commit the rest. """
+        finished_sequencer_keys = [
+            k[0] for k in self._sequencers.items() if k[1].is_finished()
+        ]
```
plamut: It would IMO be a bit more readable if we directly unpacked the items, rather than unpacking them with numeric indexes:
```py
finished_sequencer_keys = (
    key for key, sequencer in self._sequencers.items() if sequencer.is_finished()
)
```

plamut:
As per offline discussion, we want to wait for the subscriber side of this, as well as for a fix for a performance regression that was discovered.

pubsub/google/cloud/pubsub_v1/publisher/client.py
pradn: Using tuple destructuring to get the key & sequencer is a good idea.

I'd prefer not to allocate another object, so wont make the second change. Thanks for reminding me that dict comprehensions exist; I usually only use list comprehensions.

plamut:
Superseded by https://github.com/googleapis/python-pubsub/pull/26

PR #9955, opened 2019-12-10
fix(bigquery): fix copy pasta error in import error message (pandas -> pyarrow)

mbarkhau:


plamut:
Some of the unit tests fail, because the underlying transport is now called with an explicit `timeout` parameter (even if `None` by default), causing a mismatch in some of the call assertions.

FWIW, these tests (and several others) will be updated in #9875, but for the latter is primarily waiting for new releases of google auth and api core (see its TODO checklist for details).

PR #9514, opened 2019-10-22
🔍 Examples of Generated Samples for Code Review [DO NOT MERGE]

beccasaurus:
> Pull Request to facilitate code review of GAPIC SampleGen rendered syntax. Do not merge.

This Pull Request contains **two code samples** which were _generated_ using the GAPIC generator:
 - **Natural Language API** – "Analyze Sentiment" (`language_sentiment_text`) ([config YAML](https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1/samples/speech_transcribe_async.yaml))
 - **Speech-to-Text API** – "Long-Running Transcribe" (`speech_transcribe_async`) ([config YAML](https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/samples/language_sentiment_text.yaml))

These code samples demonstrate a few features of the code sample generator:
 - _Overall code sample structure (imports, layout, etc)_
 - _Calling a synchronous API endpoint and printing fields from the API response_
 - _Calling a long-running API, waiting for completion, and printing fields from API response_
 - _Using local files with code samples, specifically a file to use for a bytes request field_

You can find the existing, published hand-written versions of these code samples here:
 - 📙 [Analyzing Sentiment in a String](https://cloud.google.com/natural-language/docs/analyzing-sentiment#language-sentiment-string-python)
 - 📘 [Transcribing long audio files](https://cloud.google.com/speech-to-text/docs/async-recognize#speech-async-recognize-gcs-python) _(shows a different version of this sample using Cloud Storage)_

For Python, existing hand-written code samples can be found in the [`python-docs-samples`](https://github.com/GoogleCloudPlatform/python-docs-samples) repo.

The goal is for these generated code samples to meet the expectations of hand-written samples.

----

### Language experts, please TAL! 🧐

/cc @busunkim96 @crwilcox @tswast 

----

This is one of seven Pull Requests being sent for code review of rendered generated sample syntax.

| **Language** | **Pull Request** |
|---------------|-----------------|
| #️⃣ C# | https://github.com/googleapis/google-cloud-dotnet/pull/3650 |
| 🐹 Go | https://github.com/googleapis/google-cloud-go/pull/1634 |
| ☕️ Java | https://github.com/googleapis/google-cloud-java/pull/6543 |
| 🚀Node.js | https://github.com/googleapis/google-cloud-node/pull/2953 |
| 🐘 PHP | https://github.com/googleapis/google-cloud-php/pull/2420 |
[...]

PR #9970, opened 2019-12-12
[Syntax Review] Additional types of SampleGen code samples

beccasaurus:
> _Follow-up code review PR –_

To **evaluate the readiness** of 🐍 **Python's GAPIC sample generator** for use on [cloud.google.com](https://cloud.google.com),  
this provides a full _**set of samples**_ which _**demonstrate the basic features**_ of generated samples.

> _The .yaml sample config files for these samples can all be found here: [.yaml sample configs](https://github.com/beccasaurus/googleapis/compare/samples-for-syntax-review)_

The [first review Pull Request](https://github.com/googleapis/google-cloud-python/pull/9514) contained samples to show the _basic structure_ as well as _LRO samples_.

This contains samples demonstrating those features and a few more use-cases for review –

| **Sample** | **Description** |
|------------|-----------------|
| [`samplegen_basics.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-e8316523cb5d6b59f52ba104d6ca6494) | Shows basic structure/anatomy of Python samples |
| [`samplegen_no_config.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-dedf18a9ca41cbb29de05d1a2c0c2942) | Minimal sample example (zero configuration) |
| [`samplegen_no_response.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-957468985120c13be596acb8b3e8f80f) | API method which returns Empty |
| [`samplegen_paged.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-c9d69a27ad945d6015f2f02e54ee15da) | Paged API method (iterates over pages of results) |
| [`samplegen_resource_path.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-04b3c62b1ec47ffaafcb713dc2a22798) | Sample which requires user's Google Project ID |
| [`samplegen_map_field_access.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-0ac7074351beb4f3e6698f2884d383b1) | Access map value by key and enumerate keys/values |
| [`samplegen_read_and_write_files.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-19ff6af9d8696d9ae84236907f29ee4f) | Read binary file into bytes & write string to local file |
| [`samplegen_repeated_fields.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-c8f9af70ce5bc9d0ccdb976d4bd4f0f5) | Repeated request field & show loop/access by index |
| [`samplegen_write_bytes_to_file.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-54381aaa9481a571224843fcf7bdaf23) | Write bytes field in response to local file |
| [`samplegen_lro.py`](https://github.com/googleapis/google-cloud-python/pull/9970/files#diff-dd17b8952fced606e8e2ab257724c9ae) | Call long-running operation API endpoint |

This Pull Request is focused on **feedback collection** & a similar PR is being sent to all languages.

Once you've completed your **review** and left **comments**, you may feel free to **close the PR!**

> _We will scope out making all of the requested changes and reach out to language team offline._

#### 🙏 Thanks in advance for your review!

----

P.S. **Nits welcome!** _We'd like to get these as polished as possible for usage on [cloud.google.com](https://cloud.google.com)._

> Please feel free to note in comments whether something is:  🛑 Blocker -vs- 📌 Nit

-----

#### 🔍 Known Issues

 1. Should import libraries w/o version suffix, e.g. `from google.cloud import speech_v1 as speech`
 1. Docstring is missing `:` and is missing type, it should be `var_name (Type): description here`
 1. ❓ **Open Question:** Is the preference to show the method definition or only show method body?

> 📑 ([internal](http://go/samplegen-syntax))

### DIFF
--- a/language/samples/v1/samplegen_map_field_access.py
+++ b/language/samples/v1/samplegen_map_field_access.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# DO NOT EDIT! This is a generated sample ("Request",  "samplegen_map_field_access")
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-language
+
+# sample-metadata
+#   title: This sample reads and loops over a map field in the response
+#   description: This sample reads and loops over a map field in the response
+#   usage: python3 samples/v1/samplegen_map_field_access.py
+
+# [START samplegen_map_field_access]
+from google.cloud import language_v1
+from google.cloud.language_v1 import enums
+
+
+def sample_analyze_entities():
+    """This sample reads and loops over a map field in the response"""
+
+    client = language_v1.LanguageServiceClient()
+
+    type_ = enums.Document.Type.PLAIN_TEXT
+    language = "en"
+
+    # The text content to analyze
+    content = "Googleplex is at 1600 Amphitheatre Parkway, Mountain View, CA."
+    document = {"type": type_, "language": language, "content": content}
+
+    response = client.analyze_entities(document)
+    for entity in response.entities:
+        # Each detected entity has a map of metadata:
+        map_ = entity.metadata
+        # Access value by key:
+        print(u"URL: {}".format(map_["wikipedia_url"]))
+        # Loop over keys and values:
+        for key, value in map_.items():
+            print(u"{}: {}".format(key, value))
+
+        # Loop over just keys:
+        for the_key, _ in map_.items():
+            print(u"Key: {}".format(the_key))
+
+        # Loop over just values:
+        for _, the_value in map_.items():
+            print(u"Value: {}".format(the_value))
+
+
+# [END samplegen_map_field_access]
+
+
+def main():
+    sample_analyze_entities()
+
+
+if __name__ == "__main__":
+    main()
--- a/language/samples/v1/samplegen_map_field_access.py
+++ b/language/samples/v1/samplegen_map_field_access.py
@@ -43,20 +43,18 @@ def sample_analyze_entities():
 
     response = client.analyze_entities(document)
     for entity in response.entities:
-        # Each detected entity has a map of metadata:
-        map_ = entity.metadata
         # Access value by key:
-        print(u"URL: {}".format(map_["wikipedia_url"]))
+        print(u"URL: {}".format(entity.metadata["wikipedia_url"]))
         # Loop over keys and values:
-        for key, value in map_.items():
+        for key, value in entity.metadata.items():
             print(u"{}: {}".format(key, value))
 
         # Loop over just keys:
-        for the_key, _ in map_.items():
+        for the_key, _ in entity.metadata.items():
             print(u"Key: {}".format(the_key))
 
         # Loop over just values:
-        for _, the_value in map_.items():
+        for _, the_value in entity.metadata.items():
             print(u"Value: {}".format(the_value))
 
 
--- a/speech/samples/v1p1beta1/samplegen_read_and_write_files.py
+++ b/speech/samples/v1p1beta1/samplegen_read_and_write_files.py
@@ -20,8 +20,8 @@
 #   pip install google-cloud-speech
 
 # sample-metadata
-#   title: Showing repeated fields (in request and response)
-#   description: Showing repeated fields (in request and response)
+#   title: Read binary file into bytes field & write string in response to file
+#   description: Read binary file into bytes field & write string in response to file
 #   usage: python3 samples/v1p1beta1/samplegen_read_and_write_files.py
 
 # [START samplegen_read_and_write_files]
@@ -31,7 +31,7 @@ import io
 
 
 def sample_recognize():
-    """Showing repeated fields (in request and response)"""
+    """Read binary file into bytes field & write string in response to file"""
 
     client = speech_v1p1beta1.SpeechClient()
 
--- a/speech/samples/v1p1beta1/samplegen_read_and_write_files.py
+++ b/speech/samples/v1p1beta1/samplegen_read_and_write_files.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# DO NOT EDIT! This is a generated sample ("Request",  "samplegen_read_and_write_files")
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-speech
+
+# sample-metadata
+#   title: Showing repeated fields (in request and response)
+#   description: Showing repeated fields (in request and response)
+#   usage: python3 samples/v1p1beta1/samplegen_read_and_write_files.py
+
+# [START samplegen_read_and_write_files]
+from google.cloud import speech_v1p1beta1
+from google.cloud.speech_v1p1beta1 import enums
+import io
+
+
+def sample_recognize():
+    """Showing repeated fields (in request and response)"""
+
+    client = speech_v1p1beta1.SpeechClient()
+
+    encoding = enums.RecognitionConfig.AudioEncoding.MP3
+    config = {"encoding": encoding}
+
+    # The bytes from this file will be read into `content`
+    with io.open("path/to/file.mp3", "rb") as f:
+        content = f.read()
+    audio = {"content": content}
+
+    response = client.recognize(config, audio)
+    # Your audio has been transcribed.
+    # Writing audio transcript to transcript.txt for demonstration:
+    with open("transcript.txt", "w") as out:
+        out.write(response.results[0].alternatives[0].transcript)
+
+
+# [END samplegen_read_and_write_files]
+
+
+def main():
+    sample_recognize()
+
+
+if __name__ == "__main__":
+    main()
--- a/speech/samples/v1p1beta1/samplegen_repeated_fields.py
+++ b/speech/samples/v1p1beta1/samplegen_repeated_fields.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# DO NOT EDIT! This is a generated sample ("Request",  "samplegen_repeated_fields")
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-speech
+
+# sample-metadata
+#   title: Showing repeated fields (in request and response)
+#   description: Showing repeated fields (in request and response)
+#   usage: python3 samples/v1p1beta1/samplegen_repeated_fields.py
+
+# [START samplegen_repeated_fields]
+from google.cloud import speech_v1p1beta1
+from google.cloud.speech_v1p1beta1 import enums
+
+
+def sample_recognize():
+    """Showing repeated fields (in request and response)"""
+
+    client = speech_v1p1beta1.SpeechClient()
+
+    encoding = enums.RecognitionConfig.AudioEncoding.MP3
+
+    # A list of strings containing words and phrases "hints"
+    phrases_element = "Fox in socks"
+    phrases_element_2 = "Knox in box"
+    phrases = [phrases_element, phrases_element_2]
+    speech_contexts_element = {"phrases": phrases}
+    speech_contexts = [speech_contexts_element]
+    config = {"encoding": encoding, "speech_contexts": speech_contexts}
+    uri = "gs://[BUCKET]/[FILENAME]"
+    audio = {"uri": uri}
+
+    response = client.recognize(config, audio)
+    # Loop over all transcription results
+    for result in response.results:
+        # The first "alternative" of each result contains most likely transcription
+        alternative = result.alternatives[0]
+        print(u"Transcription of result: {}".format(alternative.transcript))
+
+
+# [END samplegen_repeated_fields]
+
+
+def main():
+    sample_recognize()
+
+
+if __name__ == "__main__":
+    main()
--- a/texttospeech/samples/v1/samplegen_write_bytes_to_file.py
+++ b/texttospeech/samples/v1/samplegen_write_bytes_to_file.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# DO NOT EDIT! This is a generated sample ("Request",  "samplegen_write_bytes_to_file")
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-texttospeech
+
+# sample-metadata
+#   title: Synthesize an .mp3 file with audio saying the provided phrase
+#   description: Synthesize an .mp3 file with audio saying the provided phrase
+#   usage: python3 samples/v1/samplegen_write_bytes_to_file.py
+
+# [START samplegen_write_bytes_to_file]
+from google.cloud import texttospeech_v1
+from google.cloud.texttospeech_v1 import enums
+
+
+def sample_synthesize_speech():
+    """Synthesize an .mp3 file with audio saying the provided phrase"""
+
+    client = texttospeech_v1.TextToSpeechClient()
+
+    text = "Hello, world!"
+    input_ = {"text": text}
+    language_code = "en"
+    voice = {"language_code": language_code}
+    audio_encoding = enums.AudioEncoding.MP3
+    audio_config = {"audio_encoding": audio_encoding}
+
+    response = client.synthesize_speech(input_, voice, audio_config)
+    print(u"Writing the synthsized results to output.mp3")
+    with open("output.mp3", "wb") as out:
+        out.write(response.audio_content)
+
+
+# [END samplegen_write_bytes_to_file]
+
+
+def main():
+    sample_synthesize_speech()
+
+
+if __name__ == "__main__":
+    main()
--- a/vision/samples/v1/samplegen_basics.py
+++ b/vision/samples/v1/samplegen_basics.py
@@ -29,7 +29,12 @@ from google.cloud import vision_v1
 
 
 def sample_create_product_set(display_name):
-    """This is the sample description"""
+    """
+    This is the sample description
+
+    Args:
+      display_name Description of the parameter
+    """
 
     client = vision_v1.ProductSearchClient()
 
--- a/vision/samples/v1/samplegen_basics.py
+++ b/vision/samples/v1/samplegen_basics.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# DO NOT EDIT! This is a generated sample ("Request",  "samplegen_basics")
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-vision
+
+# sample-metadata
+#   title: This is the sample title
+#   description: This is the sample description
+#   usage: python3 samples/v1/samplegen_basics.py [--display_name "This is the default value of the display_name request field"]
+
+# [START samplegen_basics]
+from google.cloud import vision_v1
+
+
+def sample_create_product_set(display_name):
+    """This is the sample description"""
+
+    client = vision_v1.ProductSearchClient()
+
+    # display_name = 'This is the default value of the display_name request field'
+
+    # The project and location in which the product set should be created.
+    parent = client.location_path("[PROJECT]", "[LOCATION]")
+    product_set = {"display_name": display_name}
+
+    response = client.create_product_set(parent, product_set)
+    # The API response represents the created product set
+    product_set = response
+    print(u"The full name of the created product set: {}".format(product_set.name))
+
+
+# [END samplegen_basics]
+
+
+def main():
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--display_name",
+        type=str,
+        default="This is the default value of the display_name request field",
+    )
+    args = parser.parse_args()
+
+    sample_create_product_set(args.display_name)
+
+
+if __name__ == "__main__":
+    main()
--- a/vision/samples/v1/samplegen_no_config.py
+++ b/vision/samples/v1/samplegen_no_config.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 Google LLC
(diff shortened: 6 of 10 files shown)


beccasaurus:
@busunkim96 Follow-up PR, as promised!

This shows additional syntax which previous PRs didn't include, so y'all can see all of the main snippets used to generate these Python samples. Please feel free to leave nits and blockers alike!

I mentioned some carry-over comments from the previous PR in the description above.

vision/samples/v1/samplegen_basics.py
```diff
@@ -0,0 +1,71 @@
...
+
+# To install the latest published package dependency, execute the following:
+#   pip install google-cloud-vision
+
+# sample-metadata
+#   title: This is the sample title
+#   description: This is the sample description
+#   usage: python3 samples/v1/samplegen_basics.py [--display_name "This is the default value of the display_name request field"]
+
+# [START samplegen_basics]
```
beccasaurus: @tswast You mentioned a preference for **not showing the method definition** in the embedded code? (START/END)

Do you know if that's the **standard nowadays?** I've seen both types (_with and without method signature_).

> No opinion on my end! The template for samples can be updated to whatever is standard! 😄 

**Does anyone have a canonical example?** e.g. for Node.js they have [these canonical examples](https://g3doc.corp.google.com/company/teams/cloud-devrel/dpe/samples/nodejs.md?cl=head#canonical-samples)

I'd love to update the structure **if** that's what's needed. Would it be a blocker if we show the method definition?

> /cc @busunkim96 – This is the biggest piece of feedback from the previous PR ^

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+def sample_analyze_entities():
+    """This sample reads and loops over a map field in the response"""
+
+    client = language_v1.LanguageServiceClient()
+
+    type_ = enums.Document.Type.PLAIN_TEXT
+    language = "en"
+
+    # The text content to analyze
+    content = "Googleplex is at 1600 Amphitheatre Parkway, Mountain View, CA."
```
gguuss: For Python, I prefer single-quotes over double-quotes [as per our authoring guide](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md). I just find in Python for single-quotes to be more idiomatic than double-quotes.

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
```diff
@@ -0,0 +1,61 @@
...
+#   pip install google-cloud-speech
+
+# sample-metadata
+#   title: Read binary file into bytes field & write string in response to file
+#   description: Read binary file into bytes field & write string in response to file
+#   usage: python3 samples/v1p1beta1/samplegen_read_and_write_files.py
+
+# [START samplegen_read_and_write_files]
+from google.cloud import speech_v1p1beta1
+from google.cloud.speech_v1p1beta1 import enums
```
gguuss: I prefer importing modules over members.

```
from google.cloud import speech_v1p1beta1
```

This helps to avoid alias collisions should people be importing multiple client libraries.

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
```diff
@@ -0,0 +1,61 @@
...
+    # Your audio has been transcribed.
+    # Writing audio transcript to transcript.txt for demonstration:
+    with open("transcript.txt", "w") as out:
+        out.write(response.results[0].alternatives[0].transcript)
+
+
+# [END samplegen_read_and_write_files]
+
+
+def main():
```
gguuss: Maybe skip having a separate function and put this down below where line 60 is now

busunkim96:
CC/ @dmahugh @leahecole @kurtisvg @kurtisvg @crwilcox

language/samples/v1/samplegen_map_field_access.py
beccasaurus: Ack!

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
beccasaurus: Good thought, having the method -vs- not having the method is something I've seen both of.

I direct you to this comment (on the same or) => [Method or No Method](#discussion_r357394695) (_that is the question..._)

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
beccasaurus: Ack. Thanks for pointing this out!

If I remember correctly, some LOC can be very long with `<module>.<enums>.` for each enum. If the generator were smart enough, it could decide when [not] to import enums.

vision/samples/v1/samplegen_no_config.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
```
beccasaurus: ℹ️ **Note**: this "no config" sample is different from the rest. Rather than demonstrating a feature, it shows what a sample looks like when _not configured at all_.

These samples _could_ be generated for every API method without any manual configuration.

Review of the "no config" sample is not required, this is more of an FYI :)

speech/samples/v1p1beta1/samplegen_repeated_fields.py
```diff
@@ -0,0 +1,65 @@
...
+    """Showing repeated fields (in request and response)"""
+
+    client = speech_v1p1beta1.SpeechClient()
+
+    encoding = enums.RecognitionConfig.AudioEncoding.MP3
+
+    # A list of strings containing words and phrases "hints"
+    phrases_element = "Fox in socks"
+    phrases_element_2 = "Knox in box"
+    phrases = [phrases_element, phrases_element_2]
```
kurtisvg: nit: Can we just inline these phrase_elements? They only appear to be used in the array, so attaching a variable name seems unnecessary.

speech/samples/v1p1beta1/samplegen_repeated_fields.py
```diff
@@ -0,0 +1,65 @@
...
+    speech_contexts_element = {"phrases": phrases}
+    speech_contexts = [speech_contexts_element]
+    config = {"encoding": encoding, "speech_contexts": speech_contexts}
+    uri = "gs://[BUCKET]/[FILENAME]"
```
kurtisvg: nit: Can we use terms like `"gs://[YOUR_BUCKET]/[YOUR_FILENAME]"` to make crystal clear that it is up to the user to provide these values?

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+
+# [START samplegen_map_field_access]
+from google.cloud import language_v1
+from google.cloud.language_v1 import enums
+
+
+def sample_analyze_entities():
+    """This sample reads and loops over a map field in the response"""
+
+    client = language_v1.LanguageServiceClient()
```
kurtisvg: Can we add a comment to clarify the role of the client? Some clients generate background resources and should only be created once and used multiple times, while others are just wrappers around HTTP calls. I'm not sure which this client falls under, but I think it would be useful to signal this to users.

vision/samples/v1/samplegen_basics.py
kurtisvg: Python never settled on the canonical format, but there are a few of us working towards doing so (the original PR was currently [here](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1756), and we have an AI on cleaning it up and creating a sample style guide).

Personally, I have a slight preference to leaving the function signature in (since it would make it easier for me to copy/paste into existing code).

vision/samples/v1/samplegen_resource_path.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Create product set (demonstrate resource paths)
+
+    Args:
+      project The Google Cloud Project for creating this product set
+    """
+
+    client = vision_v1.ProductSearchClient()
+
+    # project = '[PROJECT ID]'
```
kurtisvg: Instead of commenting these out, can we use default values the user can replace?

This way they get compile checked. 
```python
def sample_create_product_set(project="YOUR_DISPLAY_NAME"):
```

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
beccasaurus: Good catch, [this comes from the common template](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/py/common.snip#L36) for all .py files generated by the monolith
/cc @busunkim96

Is Python 2 support being dropped _with_ the µgenerator or _before_ the µgenerator?

This will probably disappear once we move over to samples from the µgenerator, good catch! /cc @software-dov

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+from google.cloud import language_v1
+from google.cloud.language_v1 import enums
+
+
+def sample_analyze_entities():
+    """This sample reads and loops over a map field in the response"""
+
+    client = language_v1.LanguageServiceClient()
+
+    type_ = enums.Document.Type.PLAIN_TEXT
```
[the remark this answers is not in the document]
beccasaurus: Sweet suggestion!

Thus far, when authoring these samples, I've often added these comments manually _(which you can do via the configuration)_ but have't looked into generating links with the generator.

We could presumably construct something close to a permalink, given any type name, but we wouldn't be able to include the version number of the pip package, so the link would 404 if the type were ever removed.

^--- synth could add the version number to these links, though.

@busunkim96 Any idea if the Python documentation includes the available enum values with descriptions?

I found this page, but it doesn't include the enum values: https://googleapis.dev/python/language/latest/gapic/v1/types.html

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+    language = "en"
+
+    # The text content to analyze
+    content = "Googleplex is at 1600 Amphitheatre Parkway, Mountain View, CA."
+    document = {"type": type_, "language": language, "content": content}
+
+    response = client.analyze_entities(document)
+    for entity in response.entities:
+        # Access value by key:
+        print(u"URL: {}".format(entity.metadata["wikipedia_url"]))
```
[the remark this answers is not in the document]
beccasaurus: w00t! Glad to hear this, I'd love to use `f""` formatting 💕

I believe we're close enough now to the Python 2 deprecation to switch this over!

Same with removing the `u""` ~ some developers had gotten bitten by encodings when using Python samples from the website, so we added `u""` because it was a very lightweight way to keep developers from hurting themselves. But, in the context of deprecating Python 2, I'd love to see this fully moved over to `f"Hello {var}"` without the `u""`

+💯

texttospeech/samples/v1/samplegen_write_bytes_to_file.py
```diff
@@ -0,0 +1,58 @@
...
+from google.cloud.texttospeech_v1 import enums
+
+
+def sample_synthesize_speech():
+    """Synthesize an .mp3 file with audio saying the provided phrase"""
+
+    client = texttospeech_v1.TextToSpeechClient()
+
+    text = "Hello, world!"
+    input_ = {"text": text}
```
leahecole: Should we instead name this "speech_input" or "sample_input" to avoid having to use the underscore notation? that syntax may not be familiar to folks

vision/samples/v1/samplegen_resource_path.py
leahecole: I'm ok with this because it's how we do it in node and I'm used to it, but I'd love to hear from more folks as a tie breaker (do we leave it uncommented in java?)

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
software-dov: Python 2 support is being dropped with the transition to the microgenerator.

language/samples/v1/samplegen_map_field_access.py
software-dov: All of the generated clients are just wrappers around gRPC/HTTP calls.

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
software-dov: The target Python 3 versions supported are >= 3.5, which precludes the use of fstrings as they were introduced in 3.6

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
```diff
@@ -0,0 +1,61 @@
...
+def sample_recognize():
+    """Read binary file into bytes field & write string in response to file"""
+
+    client = speech_v1p1beta1.SpeechClient()
+
+    encoding = enums.RecognitionConfig.AudioEncoding.MP3
+    config = {"encoding": encoding}
+
+    # The bytes from this file will be read into `content`
+    with io.open("path/to/file.mp3", "rb") as f:
```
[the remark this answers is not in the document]
software-dov: Yes, it should be.

vision/samples/v1/samplegen_resource_path.py
kurtisvg: For both Go and Java we switch from commenting them out to using an overriding function to provide defaults. The issue with commented out variables is that there isn't a way to check them vs the compiler, so they often could get changed or updated without anyone noticing.

beccasaurus: Same exact thing in both Ruby and C# as well 😄

We're switching from comments to using method parameter default values.

Love it +💯

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+        # Loop over keys and values:
+        for key, value in entity.metadata.items():
+            print(u"{}: {}".format(key, value))
+
+        # Loop over just keys:
```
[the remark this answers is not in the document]
software-dov: Can you elaborate?

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+        for the_key, _ in entity.metadata.items():
```
software-dov: Or better yet,
```suggestion
for key in entity.metadata.keys():
```

language/samples/v1/samplegen_map_field_access.py
```diff
@@ -0,0 +1,69 @@
...
+            print(u"Key: {}".format(the_key))
+
+        # Loop over just values:
+        for _, the_value in entity.metadata.items():
```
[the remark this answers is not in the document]
software-dov: Or better yet,
```suggestion
for value in entity.metadata.values():
```

vision/samples/v1/samplegen_basics.py
beccasaurus: Thanks for that context @kurtisvg!

I'm not biased in either direction, iirc we didn't originally have the method signature and then we added it to the generator template.

**My preference: _"leave the function signature in"_ +1**

Happy to change this at any time!

I would find it most helpful if we can say that it's OK to launch in Q1 with the method signature and we can change after (_ie. let's not block on this one item_)

We _should_ block on things like fixing the docstring format, though. Which we'll fix!

language/samples/v1/samplegen_map_field_access.py
beccasaurus: As we've settled on `blacken`, I think the double quotes have to stay.

I'm gonna resolve (hide) this thread. Adding a new comment will re-open, so feel free to follow-up with additional/different thoughts!

vision/samples/v1/samplegen_basics.py
kurtisvg: +1 - this is a nit for me.

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
beccasaurus: @gguuss @leahecole et al

Does this fall into `nit` or `blocker` category?

And are we sure that this change should be made?

I've seen a preference _for_ importing `enums` – it's a style I've seen in a bunch of code samples. In [`python-docs-samples`](https://github.com/GoogleCloudPlatform/python-docs-samples/search?q=%22import+enums%22&unscoped_q=%22import+enums%22), you can find the `import enums` style used for all of these 10 APIs:
 - Asset
 - BigTable
 - KMS
 - Natural Language
 - Speech-to-Text
 - AutoML Tables
 - Text-to-Speech
 - Translation
 - Video Intelligence
 - Vision

[Search for "import enums"](https://github.com/GoogleCloudPlatform/python-docs-samples/search?q=%22import+enums%22&unscoped_q=%22import+enums%22)

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
beccasaurus: Hiding this thread in favor of [this one above (duplicate)](https://github.com/googleapis/google-cloud-python/pull/9970#discussion_r357394695)

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: As this will fix itself with the transition to the µgenerator, I'm hiding this thread.

**It's a good catch!** But it comes from the common template which the monolithic generator uses for all Python files, not just code samples.

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: Bump / ping @busunkim96 

Wondering if the generated Python library documentation includes enums anywhere?

@kurtisvg The alternative (_not ideal but we've done this_) is putting the responsibility for adding a link on the sample author + having them link to the enum options in the API reference

This [published sample](https://cloud.google.com/natural-language/docs/analyzing-sentiment#language-sentiment-string-python) lists all available types, but it comes from this [manually authored sample config line](https://github.com/googleapis/googleapis/blob/617d4430e337c82a8c99a8cb9f3d9e16e23ab22d/google/cloud/language/v1/samples/language_syntax_text.yaml#L16). Not ideal. But at least it's 1 canonical source for this sample in all languages.

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: ☹️

How about we do this?

1. Remove `u""`
1. Don't use `f""` (**_yet_**)

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: In practice, the sample author would more than likely use `key, value`.

But I already used `key` above and the samplegen compiler is pretty conservative with its local variable table and I don't believe it let me re-use a variable name in the same snippet in a separate loop.

Only using `the_key` because I used `key` right above.

[Here's the sample config for this sample](https://github.com/beccasaurus/googleapis/compare/samples-for-syntax-review#diff-3ae868a42bdd0c0095948a2fb2f19606R27-R57). ([all configs for this PR](https://github.com/beccasaurus/googleapis/compare/samples-for-syntax-review))

Although we show 3 variations (loop over just keys, just values, keys and values) so that we can get feedback on whether all 3 are idiomatic, in practice folks will likely loop over key+value almost every time.

```yaml
# Define procedural code to handle the response.
response:
- loop:
    collection: $resp.entities
    variable:   entity
    body:
    - comment: ["Access value by key:"]
    - print:
      - "URL: %s"
      - entity.metadata{"wikipedia_url"}

    - comment: ["Loop over keys and values:"]
    - loop:
        map: entity.metadata
        key: key
        value: value
        body:
        - print: ["%s: %s", key, value]

    - comment: ["Loop over just keys:"]
    - loop:
        map: entity.metadata
        key: the_key
        body:
        - print: ["Key: %s", the_key]

    - comment: ["Loop over just values:"]
    - loop:
        map: entity.metadata
        value: the_value
        body:
        - print: ["Value: %s", the_value]
```

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
kurtisvg: I'm okay with manually adjusting for now - presumable these auto-gen samples will still get reviewed like normal ones, so we can encourage it that way if it's listed in the sample guidelines.

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: **GOOD CATCH**

Will update to `.values()`

speech/samples/v1p1beta1/samplegen_repeated_fields.py
beccasaurus: Glad you brought this up, I'd be surprised if you hadn't. These variables are pretty gross 😔

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
kurtisvg: > Although we show 3 variations (loop over just keys, just values, keys and values) so that we can get feedback on whether all 3 are idiomatic, in practice folks will likely loop over key+value almost every time.
:+1: Ok, makes sense.

speech/samples/v1p1beta1/samplegen_repeated_fields.py
beccasaurus: +1 We could make that a convention for generated sample authoring.

In this case, this whole string simply [comes from the config](https://github.com/beccasaurus/googleapis/compare/samples-for-syntax-review#diff-3b5f2ef78cd515a8b5d9644030c7e29a):

```yaml
# Define request message.
request:
- field:   audio.uri
  value:   "gs://[BUCKET]/[FILENAME]"
- field:   config.encoding
  value:   MP3
- field:   config.speech_contexts[0].phrases[0]
  value:   "Fox in socks"
  comment: 'A list of strings containing words and phrases "hints"'
- field:   config.speech_contexts[0].phrases[1]
  value:   "Knox in box"
```

language/samples/v1/samplegen_map_field_access.py
beccasaurus: > Can we add a comment to clarify the role of the client?

Yes!

In lots of languages, there is a 1 line comment above this line. Usually simply:

```
# Initialize client
client = new library.SomeServiceClient()
```

But at least it calls attention do it!

----

TAL at [the layout Node.js uses](https://github.com/googleapis/google-cloud-node/blob/f0570e6f3d7f7fb49d5211ad7f8ed6d11809546c/generated-sample-code/samplegen_basics.js) for inspiration ([Node PR](https://github.com/googleapis/google-cloud-node/pull/2959))

> Just look at the comments, I like how it organizes/structures the sample:

```js
function main(displayName = 'This is the default value of the display_name request field') {
  // [START samplegen_basics]
  /**
  * TODO(developer): Uncomment these variables before running the sample.
  */
  // const displayName = 'This is the default value of the display_name request field';

  // Imports the client library
  const {ProductSearchClient} = require('@google-cloud/vision').v1;

  // Instantiates a client
  const productSearchClient = new ProductSearchClient();

  async function sampleCreateProductSet() {

    // Construct request
    const request = {
      parent: formattedParent,
      productSet: productSet,
    };

    // Run request
    const [response] = await productSearchClient.createProductSet(request);

    // The API response represents the created product set
    const productSet = response;
    console.log(`The full name of the created product set: ${productSet.name}`);
  }
  sampleCreateProductSet();
  // [END samplegen_basics]
}
```

texttospeech/samples/v1/samplegen_write_bytes_to_file.py
beccasaurus: Good question. Not being a Pythonista, myself, I was originally confused by these.

Would it be configurable or... whenever a keyword is used, instead of using a _ suffix, use a prefix instead which scopes the variable to the sample so it's safely not a keyword?

Ack +1

Maybe `request_input`? Trying to think of a prefix that'll work and make sense in as many cases as possible

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: > Are these second two sections (just keys + just values) relevant to the sample? 

Just verifying Python idioms 🐍

The idiomatic way to loop over just the keys -vs- just the values -vs- keys and values of Map/Hash/Dictionary objects differs in some languages.

Just getting Pythonista +1 that these are idiomatic.

e.g. we weren't using Python's `.values()` so we'll update that

language/samples/v1/samplegen_map_field_access.py
beccasaurus: w00t

beccasaurus: Will fix both the `.keys()` and `.values()` loops!

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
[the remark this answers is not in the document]
beccasaurus: Actually, we have a convention of using `"resources/<something>"` so this will work for users locally.

We've received a surprising number of bug reports from users running our samples and reporting errors like "I got a File Not Found cat.png!" so we [ship a folder of resources](https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/sample_resources.yaml) for [each library](https://github.com/googleapis/google-cloud-python/tree/master/speech/samples/resources) (_from the standard `gs://cloud-samples-data` bucket_). Usually this should be .csv's and the like, but there are 3 products which work on binary files ~> Vision (images), Speech (audio), Video (videos). Sample authors for those products should be very aware that any resources they add will end up in client library repos, so use sparingly.

Most products which accept `bytes` also accept GCS file paths which is generally favorable. So most samples tend to use GCS paths, but at least 1 sample per product should generally show developers how to Hello World call the API using an existing local file.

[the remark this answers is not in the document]
beccasaurus: ^---- follows the conventions of `python-docs-samples` (search for `resources/` directories)

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
tswast: > @busunkim96 Any idea if the Python documentation includes the available enum values with descriptions?

This is https://github.com/googleapis/artman/issues/579 / https://github.com/googleapis/google-cloud-python/issues/7503

speech/samples/v1p1beta1/samplegen_read_and_write_files.py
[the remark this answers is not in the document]
kurtisvg: If we follow the convention that most users are copy/pasting-ing samples from the docs site though, they won't have access to this file. It feels like it would make the most sense to have these be an argument with a reasonable default that the user can easily override.

language/samples/v1/samplegen_map_field_access.py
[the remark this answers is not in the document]
beccasaurus: I wonder... since the generators have the .proto, maybe including the list in the samples _is_ the right way to go. We automatically add a comment which lists the available options. No URL required (_but great if possible to construct_).

@software-dov Thoughts on grabbing the enum display names in the sample generator and automatically including them?

**This is language agnostic but seems to be one of the most popular pieces of feedback!** 🎉

Issue #9988, opened 2019-12-17
BigQuery: Document the use of timeout parameter in samples


plamut:
After adding the new `timeout` parameter to various public methods (#9987), we should demonstrate its usage in the code samples.

Users should be aware of this new feature, and should probably use it by default to avoid sporadic weird issues related to a method "getting stuck" at the transport layer.

plamut:
We don't have to complicate, simply adding an explicit `timeout` argument to various `client.method()` should be enough.

@tswast What do you think?

tswast:
Makes sense to me. I think that once we have a `timeout` parameter on all relevant methods, we can surface an example in an API best practices doc such as https://cloud.google.com/bigquery/docs/api-performance

Since usage should be common across all methods, a single timeout example should be sufficient, perhaps on `create_dataset`, since that method is pretty straightforward.

Internally at Google, setting a timeout is pretty much always considered a best practice, so I wouldn't be all that opposed to adding a reasonable-length timeout to most/all samples someday, but we'd want to be careful to pick timeouts that won't make the tests super flakey.

PR #10002, opened 2019-12-20
feat(bigquery): add timeout parameter to client's and job's public methods

### LINKED ISSUE #9987: BigQuery: Add timeout parameter to all Client  and Job methods where applicable
The upcoming fix for #7831 will add a `timeout` parameter to the `QueryJob.done()` method.

As this is not the only public method that can get stuck at the transport layer, we need to add an option to specify a timeout to these methods, too.

**Update:** This excludes the `load_table_from_file()` method and methods depending on it. These will be handled separately in #10005.

plamut:
Closes #9987.

This PR adds a `timeout` parameter to the public methods of client and job classes.

### TODO
- [x] ~~Adjust tests to verify that a `timeout` passed to a method is indeed forwarded to the underlying transport.~~
- [x] Adjust the logic in multi-request methods such as `result()` - `timeout` should be split among all requests, as opposed to using the timeout for each individual request.
    - [x] Add tests for this logic.
- [x] ~~Open a separate issue for adding a timeout to `load_table_from_*()` methods that depend on file upload classes from `google-resumable-media` - the latter dependency does not support passing an explicit timeout argument to the underlying transport.~~
Done - #10005.

### PR checklist

- [x] Appropriate docs were updated (if necessary) - will be covered in #9988.

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -206,7 +206,7 @@ class Client(ClientWithProject):
         self._http._auth_request.session.close()
         self._http.close()
 
-    def get_service_account_email(self, project=None):
+    def get_service_account_email(self, project=None, timeout=None):
         """Get the email address of the project's BigQuery service account
 
         Note:
@@ -217,6 +217,8 @@ class Client(ClientWithProject):
             project (str, optional):
                 Project ID to use for retreiving service account email.
                 Defaults to the client's project.
+            timeout (Optional[float]):
+                The number of seconds to wait for the API response.
 
         Returns:
             str: service account email address
@@ -232,10 +234,16 @@ class Client(ClientWithProject):
         if project is None:
             project = self.project
         path = "/projects/%s/serviceAccount" % (project,)
-        api_response = self._connection.api_request(method="GET", path=path)
+
+        # TODO: call thorugh self._call_api() and allow passing in a retry?
+        api_response = self._connection.api_request(
+            method="GET", path=path, timeout=timeout
+        )
         return api_response["email"]
 
-    def list_projects(self, max_results=None, page_token=None, retry=DEFAULT_RETRY):
+    def list_projects(
+        self, max_results=None, page_token=None, retry=DEFAULT_RETRY, timeout=None
+    ):
         """List projects for the project associated with this client.
 
         See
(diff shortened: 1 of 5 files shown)


bigquery/google/cloud/bigquery/job.py
```diff
@@ -719,8 +732,12 @@ def cancel(self, client=None):
         if self.location:
             extra_params["location"] = self.location
 
+        # TODO: call thorugh client._call_api() and allow passing in a retry?
```
plamut: I presume we want to call the API uniformly, i.e. always with an optional retry? A few methods still call it directly, thus the comment.

If confirmed, I will create a separate ticket to address this TODO.

plamut:
@tswast This PR can be reviewed now.

I still have to add a few final tests for handling the timeout in multi-request methods. I might not be able to finish them today, but that shouldn't be a review blocker (will complete on Monday).

tswast:
```
flake8 tests
tests/unit/test_client.py:5381:76: E231 missing whitespace after ','
Command flake8 tests failed with exit code 1
Session lint failed.
```

Run black?

bigquery/google/cloud/bigquery/job.py
tswast: Yep. Let's be consistent with retries for sure. Please file a separate ticket.

bigquery/google/cloud/bigquery/client.py
```diff
@@ -256,14 +266,18 @@ def list_projects(self, max_results=None, page_token=None, retry=DEFAULT_RETRY):
 
             retry (google.api_core.retry.Retry): (Optional) How to retry the RPC.
 
+            timeout (Optional[float]):
+                The number of seconds to wait for the underlying HTTP transport
+                before using ``retry``.
```
tswast: Is a timeout considered retryable by our default retry config? https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/retry.py

If we don't have one yet, let's test that the request is retried after a timeout.

bigquery/google/cloud/bigquery/job.py
tswast: Yes.

I don't recall any reason why `_call_api` wasn't used in the first place.

bigquery/google/cloud/bigquery/client.py
plamut: It appears not, but I can double check later.

bigquery/google/cloud/bigquery/job.py
plamut: Ticket: #10019.

bigquery/google/cloud/bigquery/client.py
plamut: **Update:** `retry` occurs in a few tests, but it's mostly just for checking if it is passed to a sub-method.

I will add an explicit test for a retry after a timeout for the `Client._call_api()` method, as it is a core helper method used at a lot of places.

PR #10011, opened 2019-12-24
feat(spanner)!: make mutations optional for commits (via synth)

larkee:
Closes #9910


larkee:
Docs is currently failing in master so this failure is unrelated to the change.

spanner/google/cloud/spanner_v1/gapic/spanner_client.py
```diff
@@ -1449,9 +1459,9 @@ def begin_transaction(
     def commit(
         self,
         session,
-        mutations,
         transaction_id=None,
         single_use_transaction=None,
+        mutations=None,
```
crwilcox: This should be treated as breaking. Existing users of the commit method will have code that fails/associates params with the incorrect data.

Issue #10005, opened 2019-12-20
BigQuery: Add timeout parameter to load_table_from_file() and its dependent methods


plamut:
The `Client.load_table_from_file()` method depends on the `MultipartUpload` and `ResumableUpload` classes from [google-resumable-media](https://github.com/googleapis/google-resumable-media-python), which do not support specifying a timeout for the operation, thus these `Client` methods were excluded from #9987.

When the resumable media dependency adds support for timeouts, a `timeout` argument should be added to these `Client` methods.

PR #10016, opened 2019-12-24
fix(bigquery): fix minimum version of dependency

### LINKED ISSUE #10008: BigQuery: Dependency pins may be inadequate
When this PR, https://github.com/googleapis/google-cloud-python/pull/9915, shipped as `google-cloud-core=1.1.0` the core dep likely needs the minimum version rev'd

```
    "google-auth >= 1.9.0, < 2.0dev",
    "google-api-core >= 1.15.0, < 2.0dev",
    "google-cloud-core >= 1.0.3, < 2.0dev"
```

HemangChothani:
Fixes #10008

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -32,8 +32,8 @@ dependencies = [
     'enum34; python_version < "3.4"',
     "google-auth >= 1.9.0, < 2.0dev",
     "google-api-core >= 1.15.0, < 2.0dev",
-    "google-cloud-core >= 1.0.3, < 2.0dev",
-    "google-resumable-media >= 0.3.1, != 0.4.0, < 0.6.0dev",
+    "google-cloud-core >= 1.1.0, < 2.0dev",
+    "google-resumable-media >= 0.5.0, < 0.6dev",
     "protobuf >= 3.6.0",
     "six >=1.13.0,< 2.0.0dev",
 ]


bigquery/setup.py
```diff
@@ -32,8 +32,8 @@
     'enum34; python_version < "3.4"',
     "google-auth >= 1.9.0, < 2.0dev",
     "google-api-core >= 1.15.0, < 2.0dev",
-    "google-cloud-core >= 1.0.3, < 2.0dev",
-    "google-resumable-media >= 0.3.1, != 0.4.0, < 0.6.0dev",
+    "google-cloud-core >= 1.1.0, < 2.0dev",
+    "google-resumable-media >= 0.5.0, < 0.6dev",
```
plamut: Why is this version bump needed? Looking at the [release notes](https://github.com/googleapis/google-resumable-media-python/releases), there have been no bugfixes since `0.3.1`, and the only new feature is raw download classes (released in `0.5.0`).

Not that I'm opposing it, just asking as a sanity check.

(besides, `google-resumable-media` version pin will have to be updated soon anyway when #10005 is implemented)

HemangChothani: @plamut Not any specific reason, for the raw download feature and  found in [setup.py](https://github.com/googleapis/google-cloud-python/blob/master/storage/setup.py#L34) file of storage so update it,so if you recommend i will revert back to` 0.3.1`.

crwilcox: In theory not much has changed. And you could likely only take core (I am pretty sure given a recent change that is required). The grouping of resumable-media was to keep storage and bigquery in sync with what they require to hopefully avoid odd incompatibilities.

plamut: Since it appears safe, let's keep the resumable-media version bump then for the reason Chris mentioned.

PR #10058, opened 2020-01-03
doc(bigquery): uncomment `Client` constructor and imports in samples

tswast:
Use a session-scope filter to ensure that a client is only constructed once per test run, saving time on authentication.

We received customer feedback (internal issue 144569827) pointing out that the `Client` constructor and `from google.cloud import bigquery` lines shouldn't be commented-out. Leaving these uncommented also better matches the latest code sample best practices. To ensure this is net-neutral in terms of test time, I update the client fixture to ensure that `bigquery.Client()` in the code samples doesn't redo the slow authentication logic.

### DIFF
--- a/bigquery/samples/add_empty_column.py
+++ b/bigquery/samples/add_empty_column.py
@@ -13,13 +13,13 @@
 # limitations under the License.
 
 
-def add_empty_column(client, table_id):
+def add_empty_column(table_id):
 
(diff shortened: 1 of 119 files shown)


tswast:
Thanks! Yeah, it's mostly a search-and-replace. Only manual changes were to `conftest.py`.

PR #10087, opened 2020-01-10
fix(logging): update test assertion and core version pins

plamut:
Towards #10088.

### PR checklist

### DIFF
--- a/logging/tests/system/test_system.py
+++ b/logging/tests/system/test_system.py
@@ -179,7 +179,8 @@ class TestLogging(unittest.TestCase):
         logger = Config.CLIENT.logger(self._logger_name("log_text_res"))
         now = datetime.datetime.utcnow()
         resource = Resource(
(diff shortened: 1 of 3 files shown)


logging/tests/system/test_system.py
```diff
@@ -179,7 +179,8 @@ def test_log_text_with_resource(self):
         logger = Config.CLIENT.logger(self._logger_name("log_text_res"))
         now = datetime.datetime.utcnow()
         resource = Resource(
-            type="gae_app", labels={"module_id": "default", "version_id": "test"}
+            type="gae_app",
+            labels={"module_id": "default", "version_id": "test", "zone": ""},
```
plamut: This is something that was already failing on `master`, does not seem to be caused by version bumps in this PR.
(I did not research when exactly the failure was introduced, though)

busunkim96: I remember seeing something about this. It sounds like it changed fairly recently in the backend.

PR #10091, opened 2020-01-10
fix(storage): fix VERSION fixtures in several tests

### LINKED ISSUE #10090: Storage: Fix IAM policy versions in unit test fixtures
#9869 added a [new check](https://github.com/googleapis/google-cloud-python/pull/9869/files#diff-b0f82e58dc32268923d71a91e4270829R158-R163) when accessing IAM objects as dictionaries. This results in exceptions when configured `VERSION > 1`.

The change broke several unit tests in `storage`, because those use `VERSION = 17` in their fixtures. These versions must be adjusted.

plamut:
Fixes #10090

A recent change to IAM objects in API core started causing an error if accessing those objects as dicts, if their version is > 1. This commit addresses that by setting the VERSION to 1 in fixtures.

### PR checklist


plamut:
The system tests failure (Python 3.6 only) is caused by the rate limit:
```
429 ... The project exceeded the rate limit for creating and deleting buckets.
```

Issue #10088, opened 2020-01-10
Adjust unit tests to account for the new connection timeout parameter


plamut:
The `google-cloud-core` [release `v1.1.0`](https://github.com/googleapis/google-cloud-python/releases/tag/core-1.1.0) introduced a new optional `timeout` parameter to the `JSONConnection` object.

The `timeout` is passed down to the transport layer, which breaks several assertions in unit tests across different libraries, thus the assertions need to be adjusted.

Additionally, the version pins of the core dependencies need to be bumped in order to assure that core connections that support timeouts are always used.

**Update:**
Looking at #10094, `security center`, `spanner`, and `storage` fail for other reasons, thus their respective labels are not applied.

plamut:
The most recent test suite run in #10094 showed that the previously failing tests of the libraries in question now pass after all the fixes. Closing.

PR #10094, opened 2020-01-10
Touch files in all libraries

plamut:
Towards #10088.

This is to trigger all test runs in order to see which suites currently break.

# DO NOT MERGE


plamut:
The security center failure is unrelated to the scope of #10088, and all other previously failing checks now pass. Time to close this PR.

PR #10110, opened 2020-01-14
chore(storage): bump api_core to 1.16.0

jkwlui:
Pending #10109 (api_core release)

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -30,7 +30,7 @@ version = "1.24.1"
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
     "google-auth >= 1.9.0, < 2.0dev",
-    "google-cloud-core >= 1.1.0, < 2.0dev",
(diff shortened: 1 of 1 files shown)


storage/setup.py
```diff
@@ -30,7 +30,7 @@
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
     "google-auth >= 1.9.0, < 2.0dev",
-    "google-cloud-core >= 1.1.0, < 2.0dev",
+    "google-cloud-core >= 1.16.0, < 2.0dev",
```
busunkim96: It looks like storage depends on `google-api-core` through `google-cloud-core`. You'll want to make a release of `google-cloud-core` that bumps the `google-api-core` version. https://github.com/googleapis/google-cloud-python/blob/master/core/setup.py

Then you can update this PR with the new `google-cloud-core` version.

jkwlui: 😛i'm a n00b

PR #10162, opened 2020-01-16
feat(bigquery): check json_rows arg type in insert_rows_json()

### LINKED ISSUE #10132: BigQuery: raise a `TypeError` if a dictionary is passed to `insert_rows_json`

If I want to only insert a single row at a time into a table, it's easy to accidentally try something like:

```python
json_row = {"col1": "hello", "col2": "world"}
errors = client.insert_rows_json(
    table,
    json_row
)
```

This results in a `400 BadRequest` error from the API, because it expects a list of rows, not a single row.

It's difficult to debug this situation from the API response, so it'd be better if we raised a client-side error for passing in the wrong type for `json_rows`.

Leave as-is and request a better server-side message. This may be difficult to do, as the error happens at a level above BigQuery, which translates JSON to Protobuf for internal use.

 **Additional context**

This issue was encountered by a customer engineer, and it took me a bit of debugging to figure out the actual issue. I expect other customers will encounter this problem as well.

IlyaFaer:
Fixes #10132

### DIFF
--- a/bigquery/google/cloud/bigquery/client.py
+++ b/bigquery/google/cloud/bigquery/client.py
@@ -2506,6 +2506,8 @@ class Client(ClientWithProject):
                 identifies the row, and the "errors" key contains a list of
                 the mappings describing one or more problems with the row.
         """
+        if not isinstance(json_rows, collections_abc.Sequence):
+            raise TypeError("json_rows arg should be a sequence of dicts")
         # Convert table to just a reference because unlike insert_rows,
         # insert_rows_json doesn't need the table schema. It's not doing any
         # type conversions.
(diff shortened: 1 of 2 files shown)


bigquery/google/cloud/bigquery/client.py
```diff
@@ -2506,6 +2506,8 @@ def insert_rows_json(
                 identifies the row, and the "errors" key contains a list of
                 the mappings describing one or more problems with the row.
         """
+        if not isinstance(json_rows, collections_abc.Sequence):
+            raise TypeError("json_rows argument should be a sequence of dicts")
```
IlyaFaer: Made it with `isinstance()`. It'll be `True` for sequences, but not for `dict`s:

Безымянный

tswast:
> maybe we should do the same thing for insert_rows()? The situation in it is very close to insert_rows_json():

Yes, that's a good idea.

PR #10170, opened 2020-01-17
feat(recommender): add v1, set release level to beta

busunkim96:
* Add v1 and make it the default
* Set release level to beta to reflect the underlying API. https://cloud.google.com/recommender/docs/apis
* Use GAPICBazel rather than Artman


recommender/google/cloud/recommender_v1beta1/gapic/recommender_client.py
```diff
@@ -234,7 +234,7 @@ def list_recommendations(
     ):
         """
         Lists recommendations for a Cloud project. Requires the
-        recommender.\*.list IAM permission for the specified recommender.
+        recommender.*.list IAM permission for the specified recommender.
```
busunkim96: This looks fine in the generated docs: 
image

PR #10180, opened 2020-01-22
test(bigquery): add tests for concatenating categorical columns

### LINKED ISSUE #8044: BigQuery: use union categorical to concatenate pages in `to_dataframe` when categorical dtype is requested
I am working on an enhancement for `pandas-gbq`, where we would like to reduce the memory footprint in pandas through downcasting. See https://github.com/pydata/pandas-gbq/issues/275 for the full details.

I have made a first working version and getting erratic behaviour in the conversion of STRING to pandas `category`. It seems like it has to do with the size of the query results.

See attached notebook for the code example.
[pandas-gbq-bugtracing.ipynb.zip](https://github.com/googleapis/google-cloud-python/files/3199826/pandas-gbq-bugtracing.ipynb.zip)

plamut:
Closes #8044.

This PR adds tests for concatenating multiple-page results that contain categorical column data (both with or without `pyarrow` available).

It appears that the code already works, provided that a correct explicit categorical `dtype` is passed to `to_dataframe()` method.

If `dtypes` are not specified, a default pandas concatenating behavior results in categorical columns ending up with the dtype "object", but that's already explained in the `dtypes` parameter's docstring, meaning that the code behaves as advertised.

### PR checklist

### DIFF
--- a/bigquery/tests/unit/test_table.py
+++ b/bigquery/tests/unit/test_table.py
@@ -3242,6 +3242,174 @@ class TestRowIterator(unittest.TestCase):
         with pytest.raises(ValueError):
             row_iterator.to_dataframe(bqstorage_client)
 
+    @unittest.skipIf(pandas is None, "Requires `pandas`")
+    @unittest.skipIf(
+        bigquery_storage_v1beta1 is None, "Requires `google-cloud-bigquery-storage`"
+    )
+    @unittest.skipIf(pyarrow is None, "Requires `pyarrow`")
+    def test_to_dataframe_concat_categorical_dtype_w_pyarrow(self):
+        from google.cloud.bigquery import schema
+        from google.cloud.bigquery import table as mut
+        from google.cloud.bigquery_storage_v1beta1 import reader
+
+        arrow_fields = [
+            # Not alphabetical to test column order.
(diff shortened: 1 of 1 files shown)


plamut:
@tswast Friendly ping. :)

This PR is probably something than can be reviewed and closed quickly (read: before the repo split tomorrow).

bigquery/tests/unit/test_table.py
```diff
@@ -3242,6 +3242,174 @@ def test_to_dataframe_w_bqstorage_snapshot(self):
...
+            table=mut.TableReference.from_string("proj.dset.tbl"),
+            selected_fields=schema,
+        )
+
+        # run the method under test
+        got = row_iterator.to_dataframe(
+            bqstorage_client=bqstorage_client,
+            dtypes={
+                "col_category": pandas.core.dtypes.dtypes.CategoricalDtype(
+                    categories=["low", "medium", "high"], ordered=False,
```
tswast: [no action required] I suspect there will be users who don't necessarily know the categories ahead of time, though this could trivially be done with a group by / distinct query. If we had a support cookbook, I'd say we should add this info there.

PR #10183, opened 2020-01-23
feat(spanner): add resource based routing implementation

larkee:
Implement resource based routing.

When creating a SpannerClient, check for suggested endpoints. 
If endpoints are returned, use the first one. 
If no endpoints are returned, use default endpoint (either global endpoint or user-specified endpoint).

Endpoints are cached per instance on the Client.

This feature is disabled by default and is enabled by setting the environment variable to true.

To enable:
export GOOGLE_CLOUD_SPANNER_ENABLE_RESOURCE_BASED_ROUTING=true

### DIFF
--- a/spanner/google/cloud/spanner_v1/database.py
+++ b/spanner/google/cloud/spanner_v1/database.py
@@ -60,7 +60,7 @@ _DATABASE_NAME_RE = re.compile(
 
 _RESOURCE_ROUTING_PERMISSIONS_WARNING = (
     "The client library attempted to connect to an endpoint closer to your Cloud Spanner data "
-    "but was unable to do so. The client library will fallback and route requests to the endpoint "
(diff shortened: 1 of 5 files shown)


spanner/google/cloud/spanner_v1/database.py
```diff
@@ -54,6 +58,19 @@
 )
 
 
+_RESOURCE_ROUTING_PERMISSIONS_WARNING = (
+    "The client library attempted to connect to an endpoint closer to your Cloud Spanner data "
+    "but was unable to do so. The client library will fallback and route requests to the endpoint "
+    "given in the client options which may result in increased latency. "
```
skuruppu:
```suggestion
    "given in the client options, which may result in increased latency. "
```

PR #10193, opened 2020-01-24
fix(storage): change incorrect IAM admin role constant

msuozzo:


tseaver:
@msuozzo Thanks for the patch!  Unfortunately, the repo-split makes it invalid here.  I have ported it to  the new repo:  https://github.com/googleapis/python-storage/pull/71

PR #10195, opened 2020-01-24
test(bigquery): fix snippet test for loading partitioned tables

plamut:
Fixes #10194.

For some reason the number of rows in the public test file `us-states-by-date.csv` has changed from 50 to 49. This PR adjusts the affected test.

### PR checklist
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-cloud-python/issues) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #10196, opened 2020-01-24
fix(bigquery): fix inserting missing repeated fields

### LINKED ISSUE #9602: BigQuery: insert_rows() fails when a repeated field is missing
Using the method insert_rows to insert a row missing a REPEATED field returns an error. Repeated fields should be nullable by default, so it should not matter if a repeated field is not provided.
The method insert_rows_json, which insert_rows uses to make the API call seems to work fine.

#### Environment details
Python 2.7.15+
google-cloud-bigquery 1.21.0

#### Steps to reproduce
- Create a table with a repeated field
- Insert a row into the table using insert_rows, passing a dict where the repeated field key-value pair is missing.

plamut:
Fixes #9602.

This PR fixes the issue when inserting rows with missing REPEATED fields. Omitting a REPEATED field from the request does not result in an error anymore.

### How to test
Best to run the ready-made script from one of the [issue comments](https://github.com/googleapis/google-cloud-python/issues/9602#issuecomment-572259971).

**MIND:**
Omitting `None` values only happens when the code needs to convert the data to JSON by itself, i.e. in `insert_rows()`. The following thus still fails when `insert_rows_json()` is used:
```py
rows = [{"int_col": 9, "repeated_col": None}
errors = client.insert_rows_json(TABLE_PATH, rows)
```

If this inconsistency is not desired, we need to sync the logic in these two methods.

### PR checklist

### DIFF
--- a/bigquery/tests/unit/test_client.py
+++ b/bigquery/tests/unit/test_client.py
@@ -4636,10 +4636,13 @@ class TestClient(unittest.TestCase):
         ]
 
         def _row_data(row):
+            result = {"full_name": row[0], "age": str(row[1])}
             joined = row[2]
-            if isinstance(row[2], datetime.datetime):
-                joined = _microseconds_from_datetime(joined) * 1e-6
-            return {"full_name": row[0], "age": str(row[1]), "joined": joined}
+            if joined is not None:
+                if isinstance(joined, datetime.datetime):
+                    joined = _microseconds_from_datetime(joined) * 1e-6
+                result["joined"] = joined
+            return result
 
         SENT = {
             "rows": [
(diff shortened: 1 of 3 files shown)


plamut:
The failing snippets test is unrelated, #10195 will fix it,

plamut:
The unit tests failure will be fixed in #10222.

plamut:
Weird, a couple of PermissionDenied errors in system tests. Re-running the jobs.

bigquery/tests/unit/test_client.py
```diff
@@ -4636,10 +4636,13 @@ def test_insert_rows_w_schema(self):
         ]
 
         def _row_data(row):
+            result = {"full_name": row[0], "age": str(row[1])}
             joined = row[2]
-            if isinstance(row[2], datetime.datetime):
-                joined = _microseconds_from_datetime(joined) * 1e-6
-            return {"full_name": row[0], "age": str(row[1]), "joined": joined}
+            if joined is not None:
```
tswast: Nit: I'd prefer less nesting here.

```
if isinstance(joined, datetime.datetime):
    joined = _microseconds_from_datetime(joined) * 1e-6
if joined is not None:
    result["joined"] = joined
```

plamut: Well, _flat is better then nested_, so I'll do it. :+1:

PR #10205, opened 2020-01-24
feat(pubsub): add delivery attempt property to message object received by user code

### LINKED ISSUE #10161: pubsub: add support for dead letter topics
Add fields necessary to create/update subscriptions with a [dead letter policy](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L627-L658). This allows messages to be redelivered to a separate dead letter topic after a configured number of
delivery attempts have failed (nack/ack_deadline expired).
 
In addition to adding  `DeadLetterPolicy` to a Subscription, we need to properly expose the [`delivery_attempt` field](https://github.com/googleapis/googleapis/blob/0735b4b096872960568d1f366bfa75b7b0e1f1a3/google/pubsub/v1/pubsub.proto#L757) in the client. This is done one of two ways, so for detailed implementation details, please consult the detailed information guide that Prad wrote. Feel free to ping me if you need a link to this.

pradn:
- Exposes the delivery-attempt counter for a message.
- Add tests for plumbing + message property.
- In the future, we should modify the system test test_creating_subscriptions_with_non_default_settings or create a new one to test setting a dead letter policy on a subscription. Such a modification would make the test fail now since DLQ isn't enabled for all projects. (Only a DLQ-enabled project would pass the modified test.)

Fixes #10161

### DIFF
--- a/pubsub/google/cloud/pubsub_v1/subscriber/message.py
+++ b/pubsub/google/cloud/pubsub_v1/subscriber/message.py
@@ -70,7 +70,7 @@ class Message(object):
             published.
     """
 
-    def __init__(self, message, ack_id, request_queue):
+    def __init__(self, message, ack_id, delivery_attempt, request_queue):
         """Construct the Message.
 
         .. note::
@@ -88,6 +88,7 @@ class Message(object):
         """
         self._message = message
         self._ack_id = ack_id
+        self._delivery_attempt = delivery_attempt
         self._request_queue = request_queue
         self.message_id = message.message_id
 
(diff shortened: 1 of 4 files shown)


pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -162,6 +163,30 @@ def ack_id(self):
         """str: the ID used to ack the message."""
         return self._ack_id
 
+    @property
+    def delivery_attempt(self):
+        """The delivery attempt counter is 1 + (the sum of number of NACKs
+        and number of ack_deadline exceeds) for this message. It is set to zero
```
maheshgattani: This was brought up in an email today. Should we set it to None when dead letter policy is not set?

plamut: Having zero delivery attempts indeed feels odd, and `None` would represent the N/A (not applicable) semantics better, even if the backend denotes the same with a zero.

A side effect of returning `None` instead would also be forcing the other code to explicitly handle that case, rather getting away with simply saying `if msg.delivery_attempt > some_int`, for example. The latter could result in a silent bug, while throwing `None` into the mix would blow up loudly under Python 3.

pradn: It was decided just a day or two ago to go with the None option. Will make the change.

maheshgattani: Let me know when I can review the nil change.

pradn: Made the changes; please take a look.

pubsub/google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -82,12 +82,15 @@ def __init__(self, message, ack_id, request_queue):
             message (~.pubsub_v1.types.PubsubMessage): The message received
                 from Pub/Sub.
             ack_id (str): The ack_id received from Pub/Sub.
+            delivery_attempt (int): The delivery_attempt received
+                from Pub/Sub. May be zero.
```
maheshgattani: Can we make it more detailed? In which case can it be zero, for example.

PR #10206, opened 2020-01-24
fix: consume part of StreamingResponseIterator to support failure while under a retry context

crwilcox:
In the Node.js SDK, there's code that explicitly waits until at least one result has been observed and retries the underlying stream until then. This doesn't handle the case where the connection is interrupted mid-stream but has been good enough to resolve this class of issues. https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L1127

The Firestore Python SDK needs equivalent code to make it resilient to this class of failure.

This PR causes init of the streaming response iterator to consume the first result immediately which will cause the error to occur while under the retry context. Once the iterator is created it is returned and future access is no longer in a retry context.

### DIFF
--- a/api_core/google/api_core/grpc_helpers.py
+++ b/api_core/google/api_core/grpc_helpers.py
@@ -66,7 +66,7 @@ class _StreamingResponseIterator(grpc.Call):
         self._wrapped = wrapped
 
         # This iterator is used in a retry context, and returned outside after init.
-        # gRPC will not throw an exception until the stream is consumed, so we need 
+        # gRPC will not throw an exception until the stream is consumed, so we need
         # to retrieve the first result, in order to fail, in order to trigger a retry.
         try:
             self._stored_first_result = six.next(self._wrapped)
@@ -77,7 +77,6 @@ class _StreamingResponseIterator(grpc.Call):
         except StopIteration:
             # ignore stop iteration at this time. This should be handled outside of retry.
             pass
-
 
     def __iter__(self):
         """This iterator is also an iterable that returns itself."""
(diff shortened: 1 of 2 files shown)


api_core/google/api_core/grpc_helpers.py
```diff
@@ -64,6 +64,18 @@ def error_remapped_callable(*args, **kwargs):
...
         self._wrapped = wrapped
+        self._stored_first_result = False
+
+        # This iterator is used in a retry context, and returned outside after init.
+        # gRPC will not throw an exception until the stream is consumed, so we need 
+        # to retrieve the first result, in order to fail, in order to trigger a retry.
+        try:
+            self._first_result = six.next(self._wrapped)
+            self._stored_first_result = True
+        except StopIteration:
```
schmidt-sebastian: Will this work for the WatchStream? For Watch, the client has to send the first request. We will not receive a response until after this request has been processed.

crwilcox: Watch uses `ResumableBidiRpc` and has its own way of managing recovery. This will have an effect on things like `query` though.

crwilcox:
NOTE: 3 tests look to need refactoring, also should likely improve coverage given this new behavior.

api_core/google/api_core/grpc_helpers.py
```diff
@@ -65,6 +65,20 @@ class _StreamingResponseIterator(grpc.Call):
     def __init__(self, wrapped):
         self._wrapped = wrapped
 
+        # This iterator is used in a retry context, and returned outside after init.
+        # gRPC will not throw an exception until the stream is consumed, so we need 
+        # to retrieve the first result, in order to fail, in order to trigger a retry.
+        try:
+            self._stored_first_result = six.next(self._wrapped)
+        except TypeError:
```
crwilcox: @tseaver Under test we have a non-iterable. While it is odd that you would call this without it does seem possible? So added a guard here

Issue #10182, opened 2020-01-22
Storage: No timeouts cause indefinite hanging


mcsimps2:
Library: Google Cloud Storage
Environment: Win7 and OSX
Python Version: 3.7.3
Google Cloud Storage Version: 1.25.0

I don't believe all methods of the storage client are using timeouts.  I've come across several situations where an upload or download has completely hung because of this.  Unfortunately, there's no stack trace because the thread is just hanging waiting for a response.  Just from a brief code inspection, I can identify an example area where a timeout is not being honored:

`Bucket.get_blob` calls `blob.reload()`, which then calls the following without specifying a timeout:
```
api_response = client._connection.api_request(
            method="GET",
            path=self.path,
            query_params=query_params,
            headers=self._encryption_headers(),
            _target_object=self,
        )
```
This then calls `JSONConnection.api_request` (defaults timeout to None) -> `JSONConnection._make_request` (defaults timeout to None) -> `JSONConnection._do_request` (defaults timeout to None) -> `AuthorizedSession.request` (defaults timeout to None), which makes the ultimate call to the `requests.Session` object with a `None` timeout.  The end result is that a request is made without a timeout, which can very easily cause a thread to hang indefinitely waiting for a response.

I realize that it would be a huge pain to try and find all possible `None` timeout paths and patch them, but I at least wanted to bring this to attention.  I'm currently wrapping every call to the Google Cloud Python library with a custom timeout function that forcefully stops execution after my own specified timeout, since I have no way to pass that in to the library.  A fix that either allows developers to pass in their custom timeout, either to each function called (e.g. `get_blob(...)`) or to the `client` object so that it's passed with every single request in the underlying `http` instance, would be amazing.  (In this sense, I suppose this issue is a mix of a bug and a feature request, so my apologies if I chose the incorrect category.)

Storage code for resumable uploads that makes the call to `resumable_media/requests/_helpers.py`, more specifically the `http_request` function, seems to do much better since that function sets a default timeout of (60, 61) as opposed to `None`.

akuttig-hs:
I believe we are experiencing a similar problems every since 1.25.0 was released last week. We are accessing blobs directly and getting random 504 errors.

The code can be as simple as:
`storage_client = storage.Client()
  bucket = storage_client.get_bucket("some-bucket")`

Logs show errors like this:
[...]

PR #10210, opened 2020-01-27
feat(storage): add timeout parameter to public methods

### LINKED ISSUE #10199: Storage: add timeout parameter to all public methods
As a  library user, I would like to have a way to specify a (transport) timeout when calling methods that make HTTP requests under the hood. The timeout should have a reasonable default to prevent requests for hanging indefinitely in case I forget to pass in a `timeout` argument myself.

plamut:
Closes #10199.
Towards #10182.

Preview only for now to show the direction this is going. There are still quite a few potentially blocking code paths that are not covered yet.

#### Notes
 - `ACL._ensure_loaded()` has a default timeout, but not all internal calls to it can be made that configurable, such as in `ACL.__iter__()`. I thus left that part to use a default timeout.
 - `Blob.download_to_file()`,  `Blob.upload_from_file()`, and the methods that rely on these two do not expose timeouts, because the underlying `resumable-media` logic applies its own default timeout to `transport.request()`, which would override user timeouts.
Getting around that would first require adding configurable timeouts to the `resumable-media` dependency (there is a PR [#116](https://github.com/googleapis/google-resumable-media-python/pull/116) that adds these, but only for uploads and it's not merged yet)   .

#### Things still to do/check:

- [x] Add `timeout` to the remaining methods that use `client._connection.api_request()` (read: helpers)
- [x] Add `timeout` to any methods that use `_connection._make_request()` directly, i.e. bypassing the `api_request()` method.
- [x]  Add `timeout` to public methods that interact with methods using the `timeout` (to pass the timeout through).

### PR checklist

### DIFF
--- a/storage/google/cloud/storage/_helpers.py
+++ b/storage/google/cloud/storage/_helpers.py
@@ -21,6 +21,9 @@ import base64
 from hashlib import md5
 import os
 
+from google.cloud.storage.constants import _DEFAULT_TIMEOUT
+
+
 STORAGE_EMULATOR_ENV_VAR = "STORAGE_EMULATOR_HOST"
 """Environment variable defining host for Storage emulator."""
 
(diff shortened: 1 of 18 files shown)


plamut:
Replaced by a PR in the new repo - https://github.com/googleapis/python-storage/pull/44

PR #10211, opened 2020-01-27
chore(storage): bump google-auth dependency to v1.11.0+

### LINKED ISSUE #10061: Storage: Default timeout for requests breaks chunked downloads
The default timeout introduced in googleapis/google-resumable-media-python#88 is causing crashes in our application. We are using chunked downloads by setting [`chunk_size`][1] on the blob, and then calling [`download_to_file`][2]. Our application is multi-threaded, and we are actually downloading files into a custom stream that is backed by a ring-buffer (so writes may block until space is available again). In some cases, and I haven't figured out the pattern yet, our application hits the default timeout when fetching a new chunk of data inside [`AuthorizedSession.request`][4]. Currently, `google.cloud.storage` offers no way to use a custom transport (as suggested [here][3]), so this workaround is not applicable, and there is no way to override the timeout.

I can't provide a simple reproduction here yet, I'm still investigating, but since it's an issues that only occurs sporadically, I'm not even sure that's possible. I'm wondering if maybe in some scenarios Python's multi-threading just hits an unlucky timing, and the thread running the request doesn't get scheduled for a longer time than usual, causing the observed timeout to be much higher. Not sure how to test this though.

I'm posting here, because the upstream change was made to fix #5909, I'm not sure what the proper fix would be.

[1]: https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py#L126
[2]: https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py#L666
[3]: https://github.com/googleapis/google-cloud-python/issues/3421#issuecomment-302650580
[4]: https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/transport/requests.py#L280-L288

### LINKED ISSUE #10137: Storage: TimeoutGuard raises TimeoutException even though upload successful
Environment Details:
 - Using google-cloud-storage==1.24.1
 - Using Mac OSX 10.14 and Windows 7 64 bit
 - Using Python 3.7.3

**Issue:** A file can upload completely to Google Cloud Storage, yet still raise a `TimeoutException` if the upload process took longer than 60 seconds.

**Details** The use of `AuthorizedSession.request` for blob uploads in the Google Cloud Storage Python library causes an unwarranted `TimeoutException`.  `The `TimeoutGuard` class raises an unnecessary `TimeoutException` on file uploads to Cloud Storage even when the Cloud Storage server is responding in a timely manner to file uploads.  In fact, a file can completely upload and the `TimeoutGuard` will still raise a `TimeoutException` even though a true request timeout never occurred.  The reason why is explained below.

**Steps to reproduce:**
I first encountered this when uploading a large file (1 GB) on a medium upload connection (10 Mbps upload).  Although the upload was successful, I was still receiving a `TimeoutException`. 
 The stacktrace is below::
```
  File "site-packages\google\cloud\storage\blob.py", line 1320, in upload_from_filename
  File "site-packages\google\cloud\storage\blob.py", line 1265, in upload_from_file
  File "site-packages\google\cloud\storage\blob.py", line 1175, in _do_upload
  File "site-packages\google\cloud\storage\blob.py", line 1122, in _do_resumable_upload
  File "site-packages\google\resumable_media\requests\upload.py", line 425, in transmit_next_chunk
  File "site-packages\google\resumable_media\requests\_helpers.py", line 136, in http_request
  File "site-packages\google\resumable_media\_helpers.py", line 150, in wait_and_retry
  File "site-packages\google\auth\transport\requests.py", line 287, in request
  File "site-packages\google\auth\transport\requests.py", line 110, in __exit__
requests.exceptions.Timeout" 
```

The core of the issue is the `TimeoutGuard` class when used in a context like `AuthorizedSession.request`.  Specifically, look at the following code in the aforementioned method:
```
with TimeoutGuard(timeout) as guard:
            response = super(AuthorizedSession, self).request(
                method,
                url,
                data=data,
                headers=request_headers,
                timeout=timeout,
                **kwargs
            )
        timeout = guard.remaining_timeout
```
There are two timeouts going on.  One of them is a true request timeout used by the `requests` library (note `AuthorizedSession` is a subclass of `requests.Session`), and this is functioning correctly.  The other timeout is a naive timeout set by `TimeoutGuard` that is causing problems.  Essentially, it starts a clock that will raise a `TimeoutException` if a certain amount of time passes, even if the Google Cloud Storage servers are responding in a timely manner.  In this case, the `requests` library will not raise a `TimeoutException` (because a true network timeout never occurred), but the `TimeoutGuard` will.

This causes issues with large files uploads or slow internet connections.  If a user tries to upload a file that takes a long time to upload, then, even if the file upload is successful and didn't raise a `TimeoutException` from the `requests` library (i.e. server was responding in a timely fashion the entire upload), during `TimeoutGuard.__exit__`, the `TimeoutGuard` will raise an unsolicited `TimeoutException`.

Here's a walkthrough of the error:
(1) File upload initiated
(2) File uploads for a couple minutes, exceeding the default timeout of 60/61 seconds (`requests/_helper.py`) that the `TimeGuard` uses.  The server is responding normally to all chunk uploads. `TimeoutException` is never thrown from the Python `requests` library because the server is consistently responding.
(3) File finishes upload, `TimeoutGuard` raises `TimeoutException` even though file upload was successful.

I've been able to work around this problem by monkeypatching the `TimeoutGuard` code, but I believe a proper fix is needed in the codebase.  I would be happy to contribute or open a pull request if a maintainer can elaborate on the need for the `TimeoutGuard` `TimeoutException` when there is already a `TimeoutException` being used by the `requests.Session` class.

### LINKED ISSUE #10200: Storage: bump google-auth dependency to 1.11.0+
The `google-auth` [release](https://github.com/googleapis/google-auth-library-python/releases/tag/v1.11.0) version 1.11.0 fixes the issue with too eager Timeout errors in cases when the underlying request takes a lot of time, but still succeeds and does not timeout itself.

In order to benefit from it, the version pin needs to be updated.

plamut:
Closes #10200.
Closes #10137.
Closes #10061.

This PR bumps `google-auth` version to benefit from its recent fixes, e.g. preventing unnecessary Timeout errors on long uploads.

### PR checklist

### DIFF
--- a/storage/setup.py
+++ b/storage/setup.py
@@ -29,7 +29,7 @@ version = "1.25.0"
 # 'Development Status :: 5 - Production/Stable'
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
-    "google-auth >= 1.9.0, < 2.0dev",
+    "google-auth >= 1.11.0, < 2.0dev",
     "google-cloud-core >= 1.2.0, < 2.0dev",
     "google-resumable-media >= 0.5.0, < 0.6dev",
 ]


plamut:
@crwilcox This is another short PR that can likely be closed before the repo split. PTAL if you manage, thanks. :)

plamut:
Replaced by the PR in the new repo - https://github.com/googleapis/python-storage/pull/45

PR #10245, opened 2020-01-29
fix(automl): make TablesClient.predict permissive to input data types

helinwang:
The current implementation checks input instance's data type according
to column spec's data type. E.g., if the column spec is float, it
requires the input to be float or int, but not string. However, this
is not the same as [tables API
contract](https://cloud.google.com/automl-tables/docs/data-types#row-format):
float column data type could be string or number values.

The current code raises exception with error messages like
```
TypeError: '0' has type str, but expected one of: int, long, float
```
when passed in a string value for numeric columns, which
should be allowed.

This PR changes the logic so that Python SDK side will be permissive
for the input data type - basically all JSON compatible data types are
allow. And rely on backend for the validation.

### DIFF
--- a/automl/google/cloud/automl_v1beta1/tables/tables_client.py
+++ b/automl/google/cloud/automl_v1beta1/tables/tables_client.py
@@ -18,6 +18,7 @@
 
 import pkg_resources
 import logging
+import six
 
 from google.api_core.gapic_v1 import client_info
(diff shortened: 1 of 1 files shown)


busunkim96:
@helinwang Please re-open this in the new repository. https://github.com/googleapis/python-automl. Thank you!

PR #10282, opened 2020-01-30
fix: update resume token for restarting BiDi streams

crwilcox:
b/148467493

### DIFF
--- a/firestore/google/cloud/firestore_v1/watch.py
+++ b/firestore/google/cloud/firestore_v1/watch.py
@@ -213,9 +213,9 @@ class Watch(object):
         self._closing = threading.Lock()
         self._closed = False
 
-        initial_request = firestore_pb2.ListenRequest(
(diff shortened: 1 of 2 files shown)


firestore/google/cloud/firestore_v1/watch.py
```diff
@@ -213,9 +213,9 @@ def __init__(
         self._closing = threading.Lock()
         self._closed = False
 
-        initial_request = firestore_pb2.ListenRequest(
-            database=self._firestore._database_string, add_target=self._targets
-        )
+        self.resume_token = None
+
+        initial_request = self._get_initial_request
```
schmidt-sebastian: Nit: This is not really the initial request anymore. It's a builder for the initial and all following requests.

crwilcox: I know :(

PR #10312, opened 2020-01-31
chore(): delete now split repo'd products

crwilcox:


busunkim96:
@crwilcox Could you also remove the .kokoro configs and update the top-level README to point to the new repos?

PR #10623, opened 2022-01-05
chore: add definitions for stable and preview

parthea:


README.rst
```diff
@@ -32,6 +32,33 @@ GA libraries have development status classifier ``Development Status :: 5 - Prod
     import path (e.g. ``google.cloud.language_v1beta2``) should be considered
     to be beta.
 
+Stable
```
busunkim96, line 35: The text in this section is the same as the one above. Could the header be edited instead?  Something like `Stable / General Availability` and `Preview / Alpha`

parthea: One concern that I have with changing the title is that links in the README of client library repos that point to the title 'General Availability' will no longer link correctly. For example [this link](https://github.com/googleapis/google-cloud-python/blob/main/README.rst#general-availability) in the [README](https://github.com/googleapis/python-iap/blob/main/README.rst#python-client-for-identity-aware-proxy) of python-iap links to the title 'General Availability'.

parthea:
I'm going to close this PR and re-open it once it's ready

PR #10677, opened 2022-10-22
chore: configure release-please

parthea:


release-please-config.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+    "release-type": "python",
+    "packages": {
+        "packages/google-cloud-workflows": {
+          "component": "google-cloud-workflows"
+        }
+    },
+    "extra-files": [
+        "**/gapic_version.py"
```
dizcology, line 10: Can the generic updater and the json updater (https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-json-files) be used together?  If so can we add a section such as the following:

```
{
      "type": "json",
      "path": "**/snippet_metadata*.json",
      "jsonpath": "$.clientLibrary.version"
    }
```

Reference metadata files: 

- https://github.com/googleapis/nodejs-language/blob/main/samples/generated/v1/snippet_metadata.google.cloud.language.v1.json
- https://github.com/googleapis/python-language/blob/main/samples/generated_samples/snippet_metadata_language_v1.json

(Note that these files have different filename formats - the one for nodejs, `snippet_metadata.google.cloud.language.v1.json` is the correct one, and we need to make sure the `extra-files` path can capture it.)

parthea: Let's wait for https://github.com/googleapis/gapic-generator-python/pull/1479 to be released before adding this config

parthea:
The release please config was added to the python mono repo post processor and merged in #10692

PR #10807, opened 2022-12-28
use Search repositories endpoint

### LINKED ISSUE #10678: Use Github Search repositories instead of getting all repos
I noticed that the script uses the [`List organization repositories`](https://docs.github.com/en/rest/repos/repos#list-organization-repositories) endpoint. However, this command asks to recover all the repos (useless data) then to filter.
It would be more relevant to use the [`Search repositories`](https://docs.github.com/en/rest/search#search-repositories) endpoint.

A simple example would be : https://api.github.com/search/repositories?q=org:googleapis+python%20in:name

We can of course improve, what do you think?

av1m:
Actually, [`updateapilist.py`](scripts/updateapilist.py) use [List organization repositories](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories) endpoint. 
We can improve this call by using [Search repositories](https://docs.github.com/en/rest/search?apiVersion=2022-11-28#search-repositories) endpoint.

Fixes #10678 🦕


parthea (approved):

parthea:
Thanks @av1m!

PR #10851, opened 2023-03-09
chore: store artifacts in placer

parthea:
Towards b/272296390

Store the packages we uploaded to PyPI in Placer. That way, we have a record of exactly what we published, which we can use to generate SBOMs and attestations.

Based on similar PR in node : https://github.com/googleapis/google-cloud-node/pull/4047


SurferJeffAtGoogle (approved):

bcoe (approved):

SurferJeffAtGoogle:
@parthea This pull request appears in my inbox.  Please merge or close.

vchudnov-g (approved):
I don't enough about this config to vouch for this being complete, but it seems safe.

PR #10880, opened 2023-03-31
chore(monorepo migration): script for metadata & automation configs

vchudnov-g:
This script should be run after the split repo is naively merged into the monorepo. This script adjusts the metadata and automation configs to work in the monorepo.

PR #10882, opened 2023-04-03
feat: adds script for git migration.

atulep:
Script that migrates Git history of a split repo to google-cloud-python.

It copies Git history from source repo (i.e. "split repo") to target repo (i.e. `google-cloud-python`). Then, creates a pull request in target repo.


scripts/split_repo_migration/run.sh
```diff
@@ -0,0 +1,54 @@
...
+# 4. path in the target repo to put the copied code
+# 5. comma-separated list of files/folders to skip
+# 6. keep these specific files that would otherwise be deleted by IGNORE_FOLDERS
+# 7. override the HEAD branch name for the migration PR
+
+echo ${PACKAGE_PATH}
+
+# run the migrate script, remove .kokoro and .github folders
+# keep the .github/.OwlBot.yaml config
+${SCRIPT_DIR}/git-migrate-history.sh \
```
vchudnov-g, line 48: You should set `$UPDATE_SCRIPT` for use by the `git_migrate_history` to point to my script in #10880 . Or pass it as a parameter.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,146 @@
...
+# merge histories
+echo "Cloning target repository: ${SOURCE_REPO}"
+git clone "git@github.com:${TARGET_REPO}.git" target-repo
+pushd target-repo
+
+git remote add --fetch migration ../source-repo
+git checkout -b "${BRANCH}"
+git merge --allow-unrelated-histories migration/main --no-edit
+
+if [[ ! -z "${UPDATE_SCRIPT}" ]]
```
vchudnov-g, line 131: LMK whether you want to commit before calling `$UPDATE_SCRIPT`, which would result in 2 commits per migration, or have everything in one commit.

If the latter, let me know whether you want to commit here; I'm currently committing at the end of `$UPDATE_SCRIPT`

scripts/split_repo_migration/run.sh
atulep: Done. Pass as a parameter.

scripts/split_repo_migration/git-migrate-history.sh
atulep: I think having two commits help with debugging and potential rollbacks.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+# comma-separated list of files/folders to skip
+IGNORE_FOLDERS=$5
+# keep these specific files that would otherwise be deleted by IGNORE_FOLDERS
+KEEP_FILES=$6
+
+# override the HEAD branch name for the migration PR
+BRANCH=$7
+
+# Path for update script for migration postprocessing.
+UPDATE_SCRIPT=$8
```
vchudnov-g, line 45: IIUC, this could be an arbitrary path in the file system, so it could have spaces we should enclose it in quotes: `"$8"`. In most cases, it's advisable to quote shell variable references for safety.

vchudnov-g, line 45: Is it true that we normally provide the `UPDATE_SCRIPT` parameter but may not always provide `BRANCH`? In that case, it may be advisable to make `UPDATE_SCRIPT` be `$7` and `BRANCH` be `$8`, so it's easier to omit. (otherwise, we have to pass `''` for `$7 ` in order to be able to pass something to `$8`.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+if [ $# -lt 4 ]
+then
+  echo "Usage: $0 <source-repo> <target-repo> <source-path> <target-path> [folders,to,skip] [files,to,keep] [branch-name]"
```
vchudnov-g, line 20: Update usage comment to include `UPDATE_SCRIPT`

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+
+# override the HEAD branch name for the migration PR
+BRANCH=$7
+
+# Path for update script for migration postprocessing.
+UPDATE_SCRIPT=$8
+
+if [[ ! -z "${UPDATE_SCRIPT}" ]]
+then
+  UPDATE_SCRIPT=$(realpath "${UPDATE_SCRIPT}")
```
vchudnov-g, line 49: quote RHS

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+if [[ ! -z "${IGNORE_FOLDERS}" ]]
+then
+  echo "Ignoring folder: ${IGNORE_FOLDERS}"
+  mkdir -p ${WORKDIR}/filtered-source
+  FOLDERS=$(echo ${IGNORE_FOLDERS} | tr "," " ")
+  # remove files/folders we don't want
+  FILTER="(rm -rf ${FOLDERS} || true)"
+  if [[ ! -z "${KEEP_FILES}" ]]
+  then
+    KEEP_FILES_SPACES=($(echo ${KEEP_FILES} | tr "," " "))
```
vchudnov-g, line 90: observation: I'm pleasantly surprised this works without quotes. In the shell `FOO=alpha beta` differs from `FOO="alpha beta"`, but it seems like `FOO=$(echo "alpha beta")` works like the latter.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+    KEEP_FILE_COMMANDS=""
+    for file in "${KEEP_FILES_SPACES[@]}"
+    do
+      if [[ $file == "${KEEP_FILES_SPACES[$LAST_ELEMENT]}" ]]
+      then
+        KEEP_FILE_COMMANDS+="git checkout -- ${file} 2>/dev/null || true"
+      else 
+        KEEP_FILE_COMMANDS+="git checkout -- ${file} 2>/dev/null || true; "
+      fi   
+    done
```
vchudnov-g, line 101:
```suggestion
    for file in "${KEEP_FILES_SPACES[@]}"
    do
      KEEP_FILE_COMMANDS+="; git checkout -- ${file} 2>/dev/null || true"
    done
```

Isn't this simpler, putting the semicolon separator at the start of the line? Less code duplication.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+if [[ -z "${BRANCH}" ]]
+then
+  # default the branch name to be generated from the source repo name
+  BRANCH=$(basename ${SOURCE_REPO})-migration
+fi
+echo "here"
+export FILTER_BRANCH_SQUELCH_WARNING=1
+
+# create a working directory
+WORKDIR=$(mktemp -d -t code-migration-XXXXXXXXXX)
```
vchudnov-g, line 61: We're fine not quoting for most likely cases, but `mktemp` uses `$TMPDIR` if set, and if that has spaces, then we would need to quote the RHS here as well as all references to `${WORKDIR}`

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+
+if [[ ! -z "${UPDATE_SCRIPT}" ]]
+then
+  bash "${UPDATE_SCRIPT}"
+fi
+echo "Success"
+
+popd # back to workdir
+
+# Do a diff between source code split repl and migrated code.
```
vchudnov-g, line 142:
```suggestion
# Do a diff between source code split repo and migrated code.
```

scripts/split_repo_migration/git-migrate-history.sh
vchudnov-g: I'm realizing I wasn't thinking about things correctly and used the wrong language here, so let me try to be accurate now.

This script pushes all the split-repo's commits (git commits) to the monorepo branch `$BRANCH`. So we have hundreds of commits, not just one or two.

IIUC (correct me if I'm wrong!), we have two options:

(A) We could issue a PR at this point. IIUC, the PR would simply merge the `$BRANCH` with this whole imported history into the `main` branch. Then after that's merged, we run my postprocessing script on the `main` branch directly and issue a PR against the `main` branch. This means that between the two PRs, there's a period where the imported split-repo shows up as a package dir in the monorepo `main`, but hasn't been fully migrated because the post-processing script hasn't run.

or

(B) On this branch `$BRANCH`, after all these imported commits, we run my post-process script and add one final commit with those changes. Then we issue a single PR to merge the whole branch into `main`. When it is merged, the new package directory will have all the migration changes it needs; we won't ever be in a state where the repo dir appears migrated but isn't really. If we want to review just the post-processing changes, we merely look at the last commit in the PR.

Writing this out, I think option B would be cleaner. WDYT? Please let me know whether there are aspects I'm failing to consider.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,168 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+if [ $# -lt 4 ]
+then
+  echo "Usage: $0 <source-repo> <target-repo> <source-path> <target-path> [folders,to,skip] [files,to,keep] [branch-name]"
+  exit 1
```
vchudnov-g, line 21: If the user sources this script rather than executing it, this will kill their shell. I suggest at the top of the file:
- either enforcing that the script is being executed, or
- defining `$EXIT` and using it in place of `exit`, like so:
  ```
  [[ "${BASH_SOURCE[0]}" != "${0}" ]] && EXIT=return || EXIT=exit
  ```

scripts/split_repo_migration/git-migrate-history.sh
atulep: Talked about this offline - the path forward here is to apply an update script in batch on all split repos after they have been migrated. This results in two PRs.

atulep: Removed update script.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Removed update script.

atulep: Removed update script.

scripts/split_repo_migration/git-migrate-history.sh
atulep: N/A

scripts/split_repo_migration/git-migrate-history.sh
atulep: Removed update script.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Nice catch! thanks.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
vchudnov-g: Oh, interesting. It doesn't error on `zsh`, which I think makes sense. Code duplication makes me nervous. I guess maybe it's OK because we only use this script during the migration, and the two usages are easy to compare.

The really right solution, I suppose, is to write all the commands, including these, that currently go into `${FILTER}` into an executable file like  `/tmp/cleantree.sh` instead (one per line, so no need to worry about semicolons nor, hence, repetition). Then below we could use  `git filter-branch --tree-filter /tmp/cleantree.sh`

I'll leave it up to you. If we wind up not needing to modify this logic, I think leaving the code as-is is OK for speed. But if we wind up tweaking it, using a file might be better.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,158 @@
...
+popd # back to workdir
+
+# Do a diff between source code split repo and migrated code.
+git clone "git@github.com:${SOURCE_REPO}.git" source-repo-validation  # Not ideal to clone again.
+rm -rf source-repo-validation/.git  # That folder is not needed for validation.
+
+if diff -r target-repo/"${TARGET_PATH}" source-repo-validation; then
+  echo "No diff"
+else
+  echo "Diff non-empty"
```
vchudnov-g, line 139: Might be nice to redirect the `diff` output to a file and if the `diff` returns an error code (diff not empty) print the name of the file here so people can easily inspect.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,158 @@
...
+  $EXIT 1
+fi
+
+pushd target-repo  # To target repo
+
+# Uncomment this to push to branch and create a pull request.
```
vchudnov-g, line 145: Do we want to go ahead and uncomment this here, or do we want the caller to do the git push and PR creation? Either is fine. If we have the caller do it, I'll include it in the batch-processing script.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,158 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+[[ "${BASH_SOURCE[0]}" != "${0}" ]] && EXIT=return || EXIT=exit
+
+if [ $# -lt 4 ]
```
vchudnov-g, line 20: We're accepting up to 7 arguments, correct?

scripts/split_repo_migration/run.sh
```diff
@@ -0,0 +1,55 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+ 
+set -e
+ 
+if [ $# -lt 2 ]
+then
+  echo "Usage: $0 <split-repo-name> <target-path>"
+  exit 1
```
vchudnov-g, line 21: Could you insert the source vs exec guard here too so we don't accidentally kill the user's shell?

scripts/split_repo_migration/run.sh
```diff
@@ -0,0 +1,55 @@
...
+ 
+# repo name (e.g. nodejs-asset)
+SPLIT_REPO=$1
+# destination directory (e.g. google-cloud-asset)
+ARTIFACT_NAME=$2
+ 
+## Get the directory of the build script
+SCRIPT_DIR=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+ 
+#export UPDATE_SCRIPT="${SCRIPT_DIR}/split-repo-post-process.sh"
```
vchudnov-g, line 32: remove cruft?

scripts/split_repo_migration/run.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ $# -lt 2 ]
+then
+  echo "Usage: $0 <split-repo-name> <target-path>"
+  exit 1
+fi
+ 
+# repo name (e.g. nodejs-asset)
+SPLIT_REPO=$1
+# destination directory (e.g. google-cloud-asset)
+ARTIFACT_NAME=$2
```
vchudnov-g, line 27: We were talking about deriving the `$ARTIFACT_NAME` from the split repo, possibly using `jq`. I think that's a good idea (single source of truth, less chance for inconsistencies), though I don't know whether you want to do it here or in `git-migrate-history`.

scripts/split_repo_migration/git-migrate-history.sh
atulep: You'll need access to tempdir where the generated code got placed in. I think it's easier to push from my script since I have that info here.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Correct. The last 3 are optional though.

scripts/split_repo_migration/run.sh
atulep: Done

scripts/split_repo_migration/run.sh
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
vchudnov-g: Oh, I see.  My bad.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,156 @@
...
+git clone "git@github.com:${SOURCE_REPO}.git" source-repo-validation  # Not ideal to clone again.
+rm -rf source-repo-validation/.git  # That folder is not needed for validation.
+
+if diff -r target-repo/"${TARGET_PATH}" source-repo-validation; then
+  echo "No diff"
+else
+  diff -r target-repo/"${TARGET_PATH}" source-repo-validation > "diff.txt"
+  echo "Diff non-empty. See ${WORKDIR}/diff.txt"
+  $EXIT 1
+fi
```
vchudnov-g, line 142:
```suggestion
DIFF_FILE="${WORKDIR}/diff.txt"
if diff -r "target-repo/${TARGET_PATH}" source-repo-validation > "${DIFF_FILE}" ; then
  echo "No diff"
else
  echo "Diff non-empty. See ${DIFF_FILE}"
  $EXIT 1
fi
```

I think the above  should work; the redirection is separate from the exit code.

You could even shorten this as follows if you want:

```suggestion
DIFF_FILE="${WORKDIR}/diff.txt"
[[ diff -r "target-repo/${TARGET_PATH}" source-repo-validation > "${DIFF_FILE}" ]]  \
  &&   echo "No diff" \
  || { echo "Diff non-empty. See ${DIFF_FILE}" ;  $EXIT 1 ; }

```

vchudnov-g (commented):
One relatively minor comment.

But I do think we want the other open comment, deriving `${ARTIFACT_NAME}` from the split repo.

scripts/split_repo_migration/run.sh
```diff
@@ -0,0 +1,56 @@
...
+# 6. keep these specific files that would otherwise be deleted by IGNORE_FOLDERS
+# 7. override the HEAD branch name for the migration PR
+# 8. path for update script.
+
+echo ${PACKAGE_PATH}
+
+# run the migrate script, remove .kokoro and .github folders
+# keep the .github/.OwlBot.yaml config
+${SCRIPT_DIR}/git-migrate-history.sh \
+  "googleapis/${SPLIT_REPO}" \
```
vchudnov-g, line 51: So this assumes we're running from a directory that contains subdirectories
`googleapis/${SPLIT_REPO}/` and `googleapis/google-cloud-python` Would be good to note this in usage comments at the top, but as long as I know I can make the batch migration script conform to this.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,156 @@
...
+    --tree-filter \
+      "shopt -s dotglob; mkdir -p ${WORKDIR}/migrated-source; mv * ${WORKDIR}/migrated-source; mkdir -p ${TARGET_PATH}; mv ${WORKDIR}/migrated-source/* ${TARGET_PATH}"
+fi
+
+# back to workdir
+popd
+
+# merge histories
+echo "Cloning target repository: ${SOURCE_REPO}"
+git clone "git@github.com:${TARGET_REPO}.git" target-repo
```
vchudnov-g, line 121: We should talk. I propose you don't clone $TARGET_REPO, but allow it to be provided as a parameter to run.sh, and here you just assume it exists. If you do this, then in my batch update script I can clone $TARGET_REPO once, and use `git worktree` to create a new workdir directory referencing that cloned repo with an API-specific branch created from `main`. Then the rest of this script can import the changes in that directory. 

This way we can easily both generate multiple per-API merge PRs simumltaneously AND not commit google-cloud-python repeatedly. WDYT? Happy to talk more.

scripts/split_repo_migration/run.sh
vchudnov-g: Related: assuming you can derive `$ARTIFACT_NAME`, than that won't need to be `$2`. Instead, make `$2` an optional branch name that you pass to `git-migrate-history` as `$7`

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,156 @@
...
+pushd target-repo
+
+git remote add --fetch migration ../source-repo
+git checkout -b "${BRANCH}"
```
vchudnov-g, line 125: The branch may already exist; this is what I'm envisioning for the batch migration script. So we shoul dmake sure this doesn't error. I think using `-B` instead of `-b` should make this case work (uses an existing branch if it exists, or creates a new one)

atulep: Done.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Sounds good. I no longer clone TARGET_REPO and instead accept path to target repo as an argument.

scripts/split_repo_migration/run.sh
atulep: N/A. Removed run.sh

scripts/split_repo_migration/git-migrate-history.sh
atulep: Thanks for suggestion. Added.

scripts/split_repo_migration/run.sh
atulep: Removed artifcat name.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -13,35 +13,45 @@
...
+  echo "Usage: $0 <source-repo> <target-repo> <source-path> [folders,to,skip] [files,to,keep] [branch-name]"
   $EXIT 1
 fi
 
 # source GitHub repository. format: <owner>/<repo>
 SOURCE_REPO=$1
 
 # destination GitHub repository. format: <owner>/<repo>
-TARGET_REPO=$2
+TARGET_REPO=$2  # Path to google-cloud-python
```
vchudnov-g, line 42: nit: Update the "format" comment on the previous line

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -102,6 +112,8 @@ then
     --tree-filter "${FILTER}"
 fi
 
+TARGET_PATH=$(jq -r '.distribution_name' .repo-metadata.json)  # -r removes quotes around the name.
```
vchudnov-g, line 115: We need to prepend `packages/` here, don't we? It's not included [in the json file](https://github.com/googleapis/python-speech/blob/4c2a6fd82d66fbbe2ddc24e222a2415ecd8def20/.repo-metadata.json#L11) and we need to conform to [this directory structure](https://github.com/googleapis/google-cloud-python/tree/main/packages)

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,166 @@
...
+
+# reorganize the filtered code into the desired target locations
+if [[ ! -z "${TARGET_PATH}" ]]
+then
+  echo "Moving files to destination path: ${TARGET_PATH}"
+  git filter-branch \
+    --force \
+    --prune-empty \
+    --tree-filter \
+      "shopt -s dotglob; mkdir -p ${WORKDIR}/migrated-source; mv * ${WORKDIR}/migrated-source; mkdir -p ${TARGET_PATH}; mv ${WORKDIR}/migrated-source/* ${TARGET_PATH}"
```
vchudnov-g, line 125: Curious: Aren't all these commands (after the `shopt`) equivalent to just this?:
```
mkdir -p ${TARGET_PATH}; mv * ${TARGET_PATH}
```

Why do move stuff into `migrated-source` just to immediately move it out? It's innocuous to leave it in, I guess, but does this have any side effects I'm not seeing?

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,166 @@
...
+fi
+
+# back to workdir
+popd
+
+# merge histories
+pushd $TARGET_REPO
+
+git remote add --fetch migration ${WORKDIR}/source-repo
```
vchudnov-g, line 134: Since we'll be working with worktrees, the remotes are shared. We should uniquify them via
```
REMOTE="remote.${SOURCE_REPO}"
...
git merge --allow-unrelated-histories ${REMOTE}/main --no-edit
```

vchudnov-g: If `${SOURCE_REPO}` doesn't work we can use `${BRANCH}` instead.

atulep: Used SOURCE_REPO

scripts/split_repo_migration/git-migrate-history.sh
atulep: Not sure, I copied this from Node.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Nice catch. Done

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+  git filter-branch \
+    --force \
+    --prune-empty \
+    --tree-filter "${FILTER}"
+fi
+
+TARGET_PATH="packages/$(jq -r '.distribution_name' .repo-metadata.json)"  # -r removes quotes around the name.
+
+# reorganize the filtered code into the desired target locations
+if [[ ! -z "${TARGET_PATH}" ]]
```
vchudnov-g, line 118: This will always evaluate to true, so we should just take out the conditional. Or we can have the conditional on just the output of `jq`.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+if diff -r "${TARGET_REPO}/${TARGET_PATH}" source-repo-validation > "${DIFF_FILE}" ; then
+  echo "No diff"
+else
+  echo "Diff non-empty. See ${DIFF_FILE}"
+  $EXIT 1
+fi
+
+pushd "${TARGET_REPO}"  # To target repo
+
+# git push -u origin "${BRANCH}" --force
```
vchudnov-g, line 157: I assume you didn't mean to leave these lines commented out, and the plan is still for this script to create the PR?

vchudnov-g (approved):
Looks good. Only issue is about the commented-out `gh` command. Could you address that before merging?

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+
+# # create pull request
+# if gh --help > /dev/null
+# then
+#   gh pr create --title "migrate code from ${SOURCE_REPO}"
```
vchudnov-g, line 162: This command was prompting me for PR info, which we don't want if we run this in a batch script. Let's figure out how to pass everything on the command line.

vchudnov-g: We could do this as a follow-up PR, IMO.

vchudnov-g: I think just adding `--body "Migrate ${SRC_API_NAME} into ${TARGET_PATH}"` should do it.

vchudnov-g: Also, we need a conventional commit title format....

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+
+DIFF_FILE="${WORKDIR}/diff.txt"
+if diff -r "${TARGET_REPO}/${TARGET_PATH}" source-repo-validation > "${DIFF_FILE}" ; then
+  echo "No diff"
+else
+  echo "Diff non-empty. See ${DIFF_FILE}"
+  $EXIT 1
+fi
+
+pushd "${TARGET_REPO}"  # To target repo
```
vchudnov-g, line 155: As per offline chat, sounds like we also need to `mkdir -p owl-bot-staging/${DISTRIBUTION_NAME}` after validation and before creating the PR.

Which means maybe we do the `jq` extract into `${DISTRIBUTION_NAME}` and then set `TARGET_PATH="packages/${DISTRIBUTION_NAME}"` above?

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+    # restore files to keep, silence errors if the file doesn't exist
+    FILTER="${FILTER}; ${KEEP_FILE_COMMANDS}"
+  fi
+  git filter-branch \
+    --force \
+    --prune-empty \
+    --tree-filter "${FILTER}"
+fi
+
+TARGET_PATH="packages/$(jq -r '.distribution_name' .repo-metadata.json)"  # -r removes quotes around the name.
```
vchudnov-g, line 115: We might need the result of the query separate from the path. See comment below about `owl-bot-staging/`

vchudnov-g (changes requested):

scripts/split_repo_migration/git-migrate-history.sh
atulep: Uncommented.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,167 @@
...
+# comma-separated list of files/folders to skip
+IGNORE_FOLDERS=$4
+
+# keep these specific files that would otherwise be deleted by IGNORE_FOLDERS
+KEEP_FILES=$5
+
+# override the HEAD branch name for the migration PR
+BRANCH=$6
+
+
```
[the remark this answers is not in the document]
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
atulep: Done

scripts/split_repo_migration/git-migrate-history.sh
atulep: We should add any necessary fixes, if needed, in follow-up PRs.

scripts/split_repo_migration/git-migrate-history.sh
```diff
@@ -0,0 +1,173 @@
...
+touch owl-bot-staging/${DISTRIBUTION_NAME}/${DISTRIBUTION_NAME}/${DISTRIBUTION_NAME}.txt
+git add owl-bot-staging
+git commit -m "Trigger owlbot post-processor"
+
+git push -u origin "${BRANCH}" --force
+
+# create pull request
+if gh --help > /dev/null
+then
+  gh pr create --title "migrate code from ${SOURCE_REPO}"  --body "feat: Migrate ${SOURCE_REPO} into ${TARGET_PATH} (#${ISSUE_NUMBER})."
```
vchudnov-g, line 168: Doesn't the conventional commit need to appear in the title? For my sample PR this morning, I wound up manually changing the title to `chore(migration): Import googleapis/python-ids into packages/google-cloud-ids` and that passed the presubmit checks. Should we try something like that?

vchudnov-g (approved):
One non-trivial comment about making the *title* have a conventional  commit format.

scripts/split_repo_migration/git-migrate-history.sh
atulep: Yeah, I updated title. It will be something like:

```chore: Import googleapis/python-ids into packages/google-cloud-ids (#11104)

See #11104.
```

vchudnov-g (approved):

PR #11106, opened 2023-04-19
fix(migration): Make migration post-processing work in real runs

vchudnov-g:
- Disable the `$DEBUG` variable. 
- Prevents trying to copy a file to itself, which fails rather than being a no-op
- Splits the migration-post-processing into two scripts: one that only touches API-specific files, and can be run in parallel for multiple APIs, and one that touches common files and must be run sequentially (though it will be generalized to deal with batches of APIs at a time)


scripts/split_repo_migration/split-repo-post-process.api-files.sh
```diff
@@ -0,0 +1,213 @@
...
+
+# TODO: Review the following: For consistency with NodeJS migration script:
+# - we are not removing `begin-after-commit-hash`
+# - we are not removing `deep-remove-regex`, even though it refers to a non-API-specific directory.
+
+$RM ${OWY_SPLIT_PATH}
+## END owlbot.yaml migration
+
+
+## START owlbot.py migration ########################################
```
parthea, line 181:
```suggestion
## START delete owlbot.py ########################################
```

vchudnov-g: Did s/migration/deletion/

Issue #11047, opened 2023-04-05
Adopt split repo: _python-ids_


vchudnov-g:
Migrate the split-repo https://github.com/googleapis/python-ids to https://github.com/googleapis/google-cloud-python.

The migration readiness criteria are the following, which we track via GitHub labels on this issue. These criteria apply to the split repo we are migrating from:
- No open issues
- No open PRs
- No handwritten samples
- No system tests
- No client-specific customizations

vchudnov-g:
@parthea   Should we wait until we archive the split repo to apply the `migration:stage:split-repo-archived` label to this issue? I can also separately modify the issue-update scripts to close issues after all the `migration:stage:*` labels are present.

parthea:
> Should we wait until we archive the split repo to apply the migration:stage:split-repo-archived label to this issue?
The split repo was archived

>I can also separately modify the issue-update scripts to close issues after all the migration:stage:* labels are present.
SGTM

parthea:
Migration complete.

PR #11118, opened 2023-04-20
chore(migration): Migrate code from googleapis/python-ids into packages/google-cloud-ids

vchudnov-g:
See #11047.


parthea (approved):

parthea:
Adding `do not merge` to make sure this is merged using a merge commit

PR #11125, opened 2023-04-21
docs: update api description

parthea:


packages/google-cloud-ids/CONTRIBUTING.rst
```diff
@@ -35,21 +35,21 @@ Using a Development Checkout
 You'll have to create a development environment using a Git checkout:
 
 - While logged into your GitHub account, navigate to the
-  ``python-ids`` `repo`_ on GitHub.
+  ``google-cloud-python`` `repo`_ on GitHub.
```
vchudnov-g, line 38: These are the updates we'll have to do for all migrated repos. Not sure we can automate this...

vchudnov-g (approved):

packages/google-cloud-ids/CONTRIBUTING.rst
atulep: Isn't this just a string replace, so can be done during your post-processing step?

PR #11134, opened 2023-04-21
chore(migration): Migrate code from googleapis/python-apigee-connect into packages/google-cloud-apigee-connect

vchudnov-g:
See #10894.


vchudnov-g:
Filed #11137 to address the apparent cause of the presubmit failure. Two interesting points, though:

- NodeJS  uses a package name that is just [`google-cloud-node/packages/google-cloud-apigeeconnect`](https://github.com/googleapis/google-cloud-node/tree/main/packages/google-cloud-apigeeconnect)
- Java uses [`google-cloud-java/java-apigee-connect/google-cloud-apigee-connect`](https://github.com/googleapis/google-cloud-java/tree/main/java-apigee-connect/google-cloud-apigee-connect)

So if they're both using OwlBot, maybe both are options for us? @parthea

PR #11139, opened 2023-04-27
feat(migration): print package name at end of git history migration

vchudnov-g:


scripts/split_repo_migration/single-library.git-migrate-history.sh
```diff
@@ -179,3 +179,4 @@ else
 fi
 
 popd
+echo "${DISTRIBUTION_NAME}"
```
atulep, line 182: Is this for debugging purposes? If not, I'd add more context to print statement. If it's, I don't think it should be committed.

vchudnov-g: I added a comment. Also see the internal batch migration script.

atulep (approved):

PR #11170, opened 2023-05-10
chore(migration): delete package-specific release-please files

vchudnov-g:
These should only exist at the top level. We'll be updating the migration scripts to respect this.


parthea (approved):

parthea:
I'm going to close this PR as I think we need to revert the commits to main for these migrations. A merge commit is needed for the migration PRs to preserve the commit history.

PR #11181, opened 2023-05-16
chore: Add workflow to update release please config

### LINKED ISSUE #11053: Automatically generate the release-please-config.json file
We should generate the [release-please-config.json file](https://github.com/googleapis/google-cloud-python/blob/main/release-please-config.json) automatically instead of maintaining it manually. We could do this using GitHub actions. 

Maintaining it manually is toilsome. In https://github.com/googleapis/google-cloud-python/pull/11051, I noticed errors in the release-please-config.json which could have been avoided if it was automatically generated. See [this commit](https://github.com/googleapis/google-cloud-python/pull/11051/commits/ec374bee70b71757f3f6a2f85cc6db438e5cb7f5) which fixed the release-please-config.json file.

parthea:
This PR adds a Github Action workflow which automatically updates the release-please configuration files when there is a change to `**/gapic_version.py`

Fixes #11053


.github/workflows/configure_release_please.yml
```diff
@@ -0,0 +1,61 @@
...
+
+on:
+  push:
+    paths:
+      - '**/gapic_version.py'
+    branches:
+      - main
+  # TODO : REMOVE THIS SECTION WHICH IS ** FOR TESTING PURPOSES **
+  pull_request:
+    branches: [ main ]
```
parthea, line 27: This was added for testing purposes and will be removed before merging.

scripts/configure_release_please/configure_release_please.py
```diff
@@ -0,0 +1,135 @@
...
+
+            gapic_version_file = next(package_dir.rglob("**/gapic_version.py"), None)
+            if gapic_version_file is None:
+                raise Exception("Failed to find gapic_version.py")
+            version = get_version_for_package(gapic_version_file)
+            # check the version in gapic_version.py and update if newer than the default 0.1.0.
+            if version != (0, 1, 0):
+                manifest_json[
+                    f"packages/{package_dir.name}"
+                ] = f"{version[0]}.{version[1]}.{version[2]}"
```
vchudnov-g, line 84: So we update the manifest with the version in each package's `gapic_version.py` file, right? What is the process that updates the version number based on the number in the manifest and the nature of the changes since the last release? Could we have a doc outlining that?

scripts/configure_release_please/.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+    "packages/google-cloud-ids": "1.5.1"
```
vchudnov-g, line 2: Is this file only here for testing? If so, could we move it under test_resources/?

scripts/configure_release_please/release-please-config.json
```diff
@@ -0,0 +1,12 @@
+{
```
vchudnov-g, line 1: Is this file only here for testing? If so, could we move it under test_resources/?

.github/workflows/configure_release_please.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# Controls when the action will run.
+
+on:
+  push:
+    paths:
+      - '**/gapic_version.py'
+    branches:
+      - main
+  # TODO : REMOVE THIS SECTION WHICH IS ** FOR TESTING PURPOSES **
```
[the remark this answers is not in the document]
parthea: It was added for testing purposes. I'll leave this comment open as a reminder to remove it.

scripts/configure_release_please/configure_release_please.py
```diff
@@ -0,0 +1,135 @@
...
+    if any(elem is None for elem in [major_version, minor_version, patch_version]):
+        raise Exception("could not detect version")
+
+    return (major_version, minor_version, patch_version)
+
+
+def walk_through_owlbot_dirs(packages_dir: Path = PACKAGES_DIR) -> List[Path]:
+    """
+    Walks through all API packages in the specified `packages_dir` path. 
+    If `packages_dir` is not provided, google-cloud-python/package will be used.
```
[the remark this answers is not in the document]
parthea: Fixed in 28935e3

scripts/configure_release_please/configure_release_please.py
```diff
@@ -0,0 +1,135 @@
...
+        raise FileNotFoundError(f"Directory {packages_dir} not found")
+    return [obj.parents[0].resolve() for obj in packages_dir.rglob("**/.OwlBot.yaml")]
+
+
+def configure_release_please_manifest(
+    package_dirs: List[Path], root_dir: Path = ROOT_DIR
+) -> None:
+    """
+    This method updates the `.release-please-manifest.json` file in the directory
+    `root_dir`. If `root_dir` is not provided, `google-cloud-python` will be used as the root.
```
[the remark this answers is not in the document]
parthea: Fixed in 28935e3

scripts/configure_release_please/configure_release_please.py
parthea: >So we update the manifest with the version in each package's gapic_version.py file, right?

Correct. We will use the version in `gapic_version.py`, as long as it is not the default version, which is either `0.0.0` or `0.1.0`

> What is the process that updates the version number based on the number in the manifest and the nature of the changes since the last release?

[Release-please](https://github.com/googleapis/release-please) will use the version number in the manifest to generate the PR with the release notes which contain a list of changes since the last release. More information is available here: https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md

> Could we have a doc outlining that?
I will share a doc that outlines the changes in this PR

scripts/configure_release_please/.release-please-manifest.json
parthea: Fixed in 5c77eba

scripts/configure_release_please/release-please-config.json
parthea: Fixed in 5c77eba

scripts/configure_release_please/configure_release_please.py
parthea: Done.

parthea:
@vchudnov-g Please could you take another look?

.github/workflows/configure_release_please.yml
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+
+name: A workflow for updating the release please configuration files
+
+# Controls when the action will run.
+
+on:
+  push:
+    paths:
+      - '**/gapic_version.py'
```
vchudnov-g, line 22: So is my understanding correct?

When the `gapic_version.py` file does not have the magic version values, when release-please runs it will update the version value in those `gapic_version.py` files and in the manifest and config files. When the update to the `.py` files is pushed, that will trigger this action, which will "update" the manifest and config file, except that "update" will be redundant at that point. Is that correct?

If so
- it might be worth checking that we have non-empty changes before creating the PR. (Does code-suggester al;ready do that for us?)
- it's slightly uncomfortable to try this redundant push, but it looks like at worst it's a no-op, so it should be safe. I wonder whether we can save the work, but probably not in a simple way.

BennZoll (approved):

scripts/configure_release_please/configure_release_please.py
```diff
@@ -0,0 +1,137 @@
...
+
+def walk_through_owlbot_dirs(packages_dir: Path = PACKAGES_DIR) -> List[Path]:
+    """
+    Walks through all API packages in the specified `packages_dir` path. 
+
+    Args:
+        packages_dir(pathlib.Path): Path to the directory which contains packages.
+
+    Returns:
+        List[pathlib.Path] which contains directories for all of the packages which have
```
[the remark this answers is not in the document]
parthea: Fixed in c4d9d64

scripts/configure_release_please/configure_release_please.py
```diff
@@ -0,0 +1,137 @@
...
+    minor_version = int(match.group("minor_version"))
+    patch_version = int(match.group("patch_version"))
+
+    if any(elem is None for elem in [major_version, minor_version, patch_version]):
+        raise Exception("could not detect version")
+
+    return (major_version, minor_version, patch_version)
+
+
+def walk_through_owlbot_dirs(packages_dir: Path = PACKAGES_DIR) -> List[Path]:
```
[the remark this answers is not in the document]
parthea: Fixed in 91dda3f

scripts/configure_release_please/configure_release_please.py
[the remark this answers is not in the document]
parthea: Fixed in https://github.com/googleapis/google-cloud-python/commit/c4d9d648d7cd363b34d2003b284ebd8831de7b86

.github/workflows/configure_release_please.yml
parthea: There are no `magic version values`. There is the concept of a default version for the purposes of having an initial version. Previously the gapic generator was using `0.1.0` as the default or initial version. This initial version was actually incorrect and should have been `0.0.0`. Release-please doesn't consider this to be a `magic version`, rather an initial version.

Release-please will use the version in the `.release-please-manifest.json` file to determine the next version of a library. It will update the version in `gapic_version.py` with whatever version it deems to be the next one based on the version in `.release-please-manifest.json`.

> When the update to the .py files is pushed, that will trigger this action, which will "update" the manifest and config file, except that "update" will be redundant at that point. Is that correct?

This is correct, except for the case where code is migrated from a split repository or when a new client library is added. In all other cases, it will be a no-op.

> it might be worth checking that we have non-empty changes before creating the PR. (Does code-suggester al;ready do that for us?)

Code suggester already does this for us

> it's slightly uncomfortable to try this redundant push, but it looks like at worst it's a no-op, so it should be safe. I wonder whether we can save the work, but probably not in a simple way.

We can review all the pull requests that are opened and reject any which are not valid. There won't be any empty PRs so every change should be valid in my opinion (and has been so far).

scripts/configure_release_please/configure_release_please.py
vchudnov-g: Apologies if we're going in circles, but I am not finding a clear, succinct reference on this.

- I understand that with after this PR goes in, if we want to arbitrarily change the version, we modify `gapic_version.py` and the action will update the manifest accordingly, rigth? And this is particularly useful for the monorepo migration, and exeptionally used afterwards. Is that right?
- During normal week in and week out, as changes get committed to various packages in this repo, release please will create release PRs and update the release manifest, right? When that happens, which process updates `gapic_version.py` so it refers to the same version?

vchudnov-g (approved):
OK, I think this is fine. I would add to the top README a summary of how the release numbers get updated. If you prefer, I can do that in a chaser PR; LMK.

scripts/configure_release_please/configure_release_please.py
parthea: > I understand that after this PR goes in, if we want to arbitrarily change the version, we modify gapic_version.py and the action will update the manifest accordingly, rigth? And this is particularly useful for the monorepo migration, and exeptionally used afterwards. Is that right?

Your understanding is correct. Also, the automation is both useful for the monorepo migration as well as afterwards.

> During normal week in and week out, as changes get committed to various packages in this repo, release please will create release PRs and update the release manifest, right? 

Correct

> When that happens, which process updates gapic_version.py so it refers to the same version?

Release-please will update `gapic_version.py` in the release-please PR.

> Could we add a summary of all these processes to the main README?

I've created https://github.com/googleapis/google-cloud-python/issues/11616 to track adding a `Releasing.RST` file which we can link to from the main README.

PR #11206, opened 2023-05-31
chore(migration): Migrate code from googleapis/python-domains into packages/google-cloud-domains

vchudnov-g:
See #11027. 


parthea (approved):

parthea (changes requested):
The autogenerated samples were not migrated.

parthea (approved):
Approving as the autogenerated samples will be re-generated in the next owlbot PR

PR #11219, opened 2023-06-02
chore(migration): allow git tags to be force created when running the migration script

parthea:
This PR modifies `scripts/split_repo_migration/single-library.post-process.common-files.sh` to allow git tags to be force created so that the script can be run more than once, in case of failure. 

Another alternative is to move the `git tag` command outside of the `scripts/split_repo_migration/single-library.post-process.common-files.sh` script so that the script can be run more than once if required.


vchudnov-g (commented):
OK, good idea. What I did yesterday was manually remove the tags, but this is better.

vchudnov-g (approved):

PR #11224, opened 2023-06-02

parthea:
This updates the common files for the migration batch that contains google-cloud-apigee-connect google-area120-tables


scripts/split_repo_migration/multiple-library.post-process.sh
```diff
@@ -82,6 +82,8 @@ shift
 
 pushd "${PATH_MONOREPO}" >& /dev/null
 
+BRANCH="migration-batch-${USER}-$$"
```
vchudnov-g: We should exclude this file. Either revert it manually, or re-generate step B now that #11221 has been merged.

vchudnov-g: Or maybe simply updating this PR to `main` will do it?

parthea: Yes, it’s fixed now

vchudnov-g (approved):

PR #11379, opened 2023-06-07

parthea:
This updates the common files for the migration batch that contains google-cloud-gke-multicloud google-cloud-iam-logging google-cloud-life-sciences google-cloud-managed-identities


vchudnov-g (approved):
I was surprised this PR did not include modified common files, but it looks like those are already in `main`...coming from the yoshi-code-bot PR #11378. Was that triggered by the pending PR?

PR #11460, opened 2023-07-06
chore(migration): Migrate code from googleapis/python-language into packages/google-cloud-language

parthea:
See #10907. 


dizcology (approved):
Maybe we don't need to keep the tar.gz file?

PR #11463, opened 2023-07-06

parthea:
This updates the common files for the migration batch that contains google-cloud-batch google-cloud-language google-cloud-workflows


packages/google-cloud-workflows/CONTRIBUTING.rst
```diff
@@ -236,7 +236,7 @@ We support:
 
 Supported versions can be found in our ``noxfile.py`` `config`_.
 
-.. _config: https://github.com/googleapis/python-workflows/blob/main/noxfile.py
+.. _config: https://github.com/googleapis/google-cloud-python/blob/main/noxfile.py
```
dizcology, line 239: should this refer to the noxfile.py in the repo root, or in each package?

UPDATE: looks like the repo root does not have noxfile.py.

dizcology (approved):

packages/google-cloud-workflows/CONTRIBUTING.rst
parthea: Good catch! This will be fixed in https://github.com/googleapis/synthtool/pull/1827

PR #11592, opened 2023-08-18
Update CHANGELOG.md

HalukCem:


vchudnov-g:
Hi @HalukCem.  Why are suggesting we get rid of these release notes? I don't think we want to do that, and as AFAICT, the links in the existing file work fine.

Issue #10968, opened 2023-04-04
Adopt split repo: _python-video-transcoder_


vchudnov-g:
Migrate the split-repo https://github.com/googleapis/python-video-transcoder to https://github.com/googleapis/google-cloud-python.

The migration readiness criteria are the following, which we track via GitHub labels on this issue. These criteria apply to the split repo we are migrating from:
- No open issues
- No open PRs
- No handwritten samples
- No system tests
- No client-specific customizations

parthea:
I'm going to transfer this issue to the `google-cloud-python` repository as we are preparing to move the code for `google-cloud-video-transcoder` to that repository in the next 1-2 weeks.

parthea:
Wait for cl/538850316

suztomo:
@parthea cl/538850316 has been submitted.

PR #11681, opened 2023-09-20
chore(migration): Migrate code from googleapis/python-video-transcoder into packages/google-cloud-video-transcoder

parthea:
See #10968. 


vchudnov-g (commented):
Just to double check: this is releasing into `packages/google-cloud-video-transcoder/google/cloud/video/transcoder/....`, i.e. using hierarchical Python namespaces. So the idea is that we migrate this over, and then separately do the breaking change to flatten namespaces, together with the other affected APIs? I think that makes sense, but I want to be sure we're on the same page/

vchudnov-g (approved):

PR #11689, opened 2023-09-20
fix(migration): Prevent failure when git filter encounters no files

vchudnov-g:
This was blocking the migration of dialogflow

### DIFF
--- a/scripts/split_repo_migration/single-library.git-migrate-history.sh
+++ b/scripts/split_repo_migration/single-library.git-migrate-history.sh
@@ -124,7 +124,7 @@ git filter-branch \
   --force \
   --prune-empty \
   --tree-filter \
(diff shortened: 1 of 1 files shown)


vchudnov-g:
Actually, I have questions about line 127: `--tree-filter` gets applied to every commit in the branch, not to the end result, and what we're doing is copying the files  from each commit into a different directory. It seems like this will do the wrong thing is some of the commits delete files: we are not copying the deletions.

In fact, I think the correct invocation should be
`"shopt -s dotglob; TEMPY=$(mktemp -p ${WORKDIR}); { mv * ${TEMPY}; mkdir -p ${TARGET_PATH};  mv ${TEMPY}/* ${TARGET_PATH} ;  } || echo 'No files to move'"`

Refs:
- https://git-scm.com/docs/git-filter-branch#_examples
- [`The tree filter will work by checking out each commit (in a separate temporary folder), running your filter on the working directory, adding any changes to the index ..., and committing the new index.`](https://manishearth.github.io/blog/2017/03/05/understanding-git-filter-branch/#:~:text=The%20tree%20filter%20will%20work,and%20committing%20the%20new%20index.)

parthea (approved):

PR #11735, opened 2023-09-22
build: add CI for system tests

parthea:
This PR adds CI for system tests. Most of the configuration was taken from the split repository templates [here](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/python_library/.kokoro).


.kokoro/system-single.sh
```diff
@@ -0,0 +1,29 @@
...
+# `-e` enables the script to automatically fail when a command fails
+# `-o pipefail` sets the exit code to non-zero if any command fails,
+# or zero if all commands in the pipeline exit successfully.
+set -eo pipefail
+
+pwd
+
+# If NOX_SESSION is set, it only runs the specified session,
+# otherwise run all the sessions.
+if [[ -n "${NOX_SESSION:-}" ]]; then
```
vchudnov-g, line 25: nit: I don't think the `:-` gets you anything here (cf https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion)


Also, what you have is fine, but you could equally well do 

```
SESSION_ARG=""
[[ -z "${NOX_SESSION} ]] || SESSION_ARG="-s ${NOX_SESSION}"
python3 -m nox ${SESSION_ARG}
```

This works in bash; in zsh, there's an additional tweak that's needed. I assume this file is executed, not sourced, in which case bash will always be used.

packages/google-cloud-access-approval/tests/system/smoke_test.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.mark.parametrize("transport", ["grpc", "rest"])
+def test_list_approval_requests(project_id: str, transport: str):
+    client = accessapproval_v1.AccessApprovalClient(transport=transport)
+
+    parent = client.common_project_path(project_id)
+    approval_requests = list(client.list_approval_requests(parent=parent))
+
+    assert len(approval_requests) >= 0
```
vchudnov-g, line 36: Just to confirm: the idea is that if the smoke test fails, we won't reach this line (i.e. we won't silently get a zero-length list)?

.kokoro/system.sh
```diff
@@ -0,0 +1,73 @@
...
+
+cd "$PROJECT_ROOT"
+
+pwd
+
+# A file for running system tests
+system_test_script="${PROJECT_ROOT}/.kokoro/system-single.sh"
+
+subdirs=(
+    ${PROJECT_ROOT}/packages
```
vchudnov-g, line 52: I think you intended a trailing `*` here?

```suggestion
    ${PROJECT_ROOT}/packages/*
```

.kokoro/system.sh
```diff
@@ -0,0 +1,73 @@
...
+export PROJECT_ROOT=$(realpath $(dirname "${BASH_SOURCE[0]}")/..)
+
+cd "$PROJECT_ROOT"
+
+pwd
+
+# A file for running system tests
+system_test_script="${PROJECT_ROOT}/.kokoro/system-single.sh"
+
+subdirs=(
```
vchudnov-g, line 51: nit:suggest s/`subdirs`/`package_dirs`/

.kokoro/system.sh
```diff
@@ -0,0 +1,73 @@
...
+    ${PROJECT_ROOT}/packages
+)
+for subdir in ${subdirs[@]}; do
```
vchudnov-g, line 54: Why do you have this outer loop? You don't seem to be using `subdir` anywhere. Your `find` command seems to be driving the loop.

.kokoro/system.sh
```diff
@@ -0,0 +1,73 @@
...
+  # Run system tests for each package with directory packages/*/tests/system
+  for d in `find 'packages' -type d -wholename 'packages/*/tests/system'`; do
```
vchudnov-g, line 56: suggest s/`d`/`dir`/

vchudnov-g (commented):
Thanks for doing this!

One important comment about a wildcard, one non-trivial comment about the loop, and some minor comments.

.kokoro/system.sh
parthea: Fixed in 82f047f

.kokoro/system.sh
parthea: Fixed in 5775357

.kokoro/system.sh
parthea: Obsolete. Code was deleted in 5775357

.kokoro/system.sh
parthea: Obsolete. Code was deleted in https://github.com/googleapis/google-cloud-python/commit/5775357cadfa73c500c7df52de66bcc09296088d

packages/google-cloud-access-approval/tests/system/smoke_test.py
parthea: Clarified the purpose of the smoke test in [4ebbaf5](https://github.com/googleapis/google-cloud-python/pull/11735/commits/4ebbaf58da40fa7c58da3b3dde83ccf8ef5cb4dd)

.kokoro/system-single.sh
parthea: Fixed in 68e60f2

packages/google-cloud-access-approval/tests/system/smoke_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.mark.parametrize("transport", ["grpc", "rest"])
+def test_list_approval_requests(project_id: str, transport: str):
+    client = accessapproval_v1.AccessApprovalClient(transport=transport)
+
+    parent = client.common_project_path(project_id)
+    client.list_approval_requests(parent=parent)
+
+    # The purpose of this smoke test is to test the communication with the API,
```
vchudnov-g, line 36:
```suggestion
    # The purpose of this smoke test is to test the communication with the API server,
```

vchudnov-g (approved):
Thanks for taking care of this!

PR #11892, opened 2023-10-21
chore(migration): Migrate code from googleapis/python-cloud-common into packages/google-cloud-common

parthea:
See #10999. 


vchudnov-g (approved):
IIUC, there should be no issues with this migration, even though this package is used by others, because the PyPI packaging does not change.

parthea:
> there should be no issues with this migration, even though this package is used by others, because the PyPI packaging does not change.

Correct

PR #11893, opened 2023-10-21
chore(migration): Migrate code from googleapis/python-compute into packages/google-cloud-compute

parthea:
See #11001. 


vchudnov-g (commented):
I think this is OK, but we should double-check that the pipeline for Compute won't be broken by this migration. (It's currently broken for other reasons, and I have a fix for that ready to go out for review today.)

vchudnov-g (approved):
OK, should be fine. Once the synthetic protos are googleapis/googleapis, building as usual per there (with the appropriate BUILD rules) should be doing the right thing.

PR #11959, opened 2023-11-01
chore: fix docs url for new non-cloud libraries

### LINKED ISSUE #11193: Broken client library links in newly generated client libraries due to invalid value for `client_documentation`
Several new client libraries had broken documentation links when they were generated using the [bootstrapping container](https://github.com/googleapis/google-cloud-python/tree/main/containers/python-bootstrap-container).

The issue is that the client library documentation links differ depending on whether the client library is for a product which is part of `Google Cloud`. When the client library belongs to `Google Cloud`, the client library documentation will be located at `"https://cloud.google.com/python/docs/reference/<apiName>/latest"`, otherwise it will be located at `"https://googleapis.dev/python/<apiName>/latest"`.

The `client_documentation` value should be populated dynamically instead of hardcoded here:
https://github.com/googleapis/google-cloud-python/blob/a429c434c6847f538b851660b2915abeef33a592/containers/python-bootstrap-container/.repo-metadata.json#L6

parthea:
Fixes #11193 🦕

### DIFF
--- a/containers/python-bootstrap-container/entrypoint.sh
+++ b/containers/python-bootstrap-container/entrypoint.sh
@@ -109,7 +109,8 @@ fi
 # Update apiProductDocumentation in .repo-metadata.json
 sed -i -e "s|apiProductDocumentation|$PRODUCT_DOCS_URL|" "${WORKSPACE_DIR}/${MONO_REPO_NAME}/packages/${FOLDER_NAME}/.repo-metadata.json"
 
-# If the API is cloud, the docs will live at cloud.google.com
+# If the API is `cloud`, the docs will be published to `cloud.google.com`.
+# For non-cloud, the docs will be published to `googleapis.dev`.
 if [[ $FOLDER_NAME =~ "cloud" ]]; then
     CLIENT_DOCS_URL="$(echo https://cloud.google.com/python/docs/reference/$FOLDER_NAME/latest)"
 else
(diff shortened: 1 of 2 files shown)


containers/python-bootstrap-container/entrypoint.sh
```diff
@@ -101,13 +101,24 @@ DOCS_ROOT_URL=$(jq --arg API_SHORTNAME "$API_SHORTNAME" -r '.apis | to_entries[]
...
 fi
 
 # Update apiProductDocumentation in .repo-metadata.json
-sed -i -e "s|apiProductDocumentation|$DOCS_URL|" "${WORKSPACE_DIR}/${MONO_REPO_NAME}/packages/${FOLDER_NAME}/.repo-metadata.json"
+sed -i -e "s|apiProductDocumentation|$PRODUCT_DOCS_URL|" "${WORKSPACE_DIR}/${MONO_REPO_NAME}/packages/${FOLDER_NAME}/.repo-metadata.json"
+
+# If the API is `cloud`, the docs will be published to `cloud.google.com`.
+# For non-cloud, the docs will be published to `googleapis.dev`.
+if [[ $FOLDER_NAME =~ "cloud" ]]; then
+    CLIENT_DOCS_URL="$(echo https://cloud.google.com/python/docs/reference/$FOLDER_NAME/latest)"
```
vchudnov-g, line 115: Why do you need the `echo` here? Why not simply
`CLIENT_DOCS_URL="https://cloud.google.com/python/docs/reference/${FOLDER_NAME}/latest"` and similarly below?

parthea: Good catch! Opened https://github.com/googleapis/google-cloud-python/pull/11963 to clean up the code.

PR #12014, opened 2023-11-10
build: add fetch depth when using actions/checkout

### LINKED ISSUE #12013: Presubmit checks run for all packages instead of only packages with changed files
Presubmit checks should only run on changed files but there appears to be an issue where checks are running on all packages.

https://github.com/googleapis/google-cloud-python/blob/c28986d33652c0eed91809436628b45667732937/ci/run_conditional_tests.sh#L40-L46

See the build log [here](https://github.com/googleapis/google-cloud-python/actions/runs/6817166694/job/18540067195?pr=12010) for the release PR https://github.com/googleapis/google-cloud-python/pull/12010 which shows that checks are running for all packages instead of only the 3 packages with changed files in the PR.

There is also an error in the build log which appears to be related to the issue
```
Run ci/run_conditional_tests.sh
From https://github.com/googleapis/google-cloud-python
 * branch                  main       -> FETCH_HEAD
 * [new branch]            main       -> origin/main
fatal: origin/main...HEAD: no merge base
change detected in ci, we should test everything
```

parthea:
Fixes #12013 🦕

### DIFF
--- a/.github/workflows/configure_release_please.yml
+++ b/.github/workflows/configure_release_please.yml
@@ -36,6 +36,11 @@ jobs:
     if: ${{github.repository == 'googleapis/google-cloud-python'}}
     steps:
       - uses: actions/checkout@v4
+        # Use a fetch-depth of 2
+        # See https://github.com/googleapis/google-cloud-python/issues/12013
+        # and https://github.com/actions/checkout#checkout-head.
+        with:
+          fetch-depth: 2
       - name: Set up Python 3.11
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,6 +13,11 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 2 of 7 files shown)


vchudnov-g (approved):
Approving since this seems to fix things, but just a note: should we set the depth to something higher than 2 so we have a higher margin in case future scenarios come up that result in the same failure mode? Or do we have a guarantee that 2 will continue to be enough?

(NOTE: `fetch-depth` refers to the number of commits to fetch: https://github.com/actions/checkout#usage)

parthea:
Sounds good. If we still encounter an issue with a `fetch-depth` of 2, I will file an issue in https://github.com/actions/checkout to ask if the value in the README should be higher https://github.com/actions/checkout#checkout-head, with a minimal reproduction of the issue using `fetch-depth` of 2 if possible.

vchudnov-g:
Looks like we can check out as deeply as we want, and "0" means check "check out all the history". See the section: https://github.com/actions/checkout#usage . It seems the issue is whether we can have any guarantee that fetching just the last 2 commits will always be enough for our action to do what we intend.

PR #12256, opened 2024-02-02
fix: add staticmethod decorator to methods inside SecretManagerServic…

### LINKED ISSUE #12255: SecretManagerServiceClient: missing staticmethod decorators on two methods
The methods are:

- _get_client_cert_source
- _get_api_endpoint

It is mainly a matter of consistency but also to avoid eventual accesses through self.

Salvo-kun:
Fixes #12255


parthea:
@Salvo-kun , Thanks for opening a PR. This code is autogenerated so we had to apply the fix via the code generator. It was included in https://github.com/googleapis/gapic-generator-python/pull/1923 and the fix was added to this repository via https://github.com/googleapis/google-cloud-python/pull/12266. I'm going to close this PR as there is no longer a code diff.

PR #12317, opened 2024-02-15
fix: make google-cloud-monitoring tests work without `mock`

### LINKED ISSUE #12316: google-cloud-monitoring tests depend on mock library which causes packaging problems on Fedora
https://github.com/googleapis/google-cloud-python/blob/aae7bf53622e4fd56fdc767d327a5d5f978718f4/packages/google-cloud-monitoring/tests/unit/test_query.py#L20 imports `mock` from https://pypi.org/project/mock/ library directly.

https://pypi.org/project/mock/ library is not needed anymore on modern Python installation, since it is now part of the Python standard library, available as `unittest.mock`.

This issue was found as a part of packaging the library for Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=2263960.

The solution is to reuse approach in auto-generated code: https://github.com/googleapis/google-cloud-python/blob/aae7bf53622e4fd56fdc767d327a5d5f978718f4/packages/google-cloud-monitoring/tests/unit/gapic/monitoring_v3/test_alert_policy_service.py#L19-L23

rominf:
https://pypi.org/project/mock/ library is not needed anymore on modern
Python installation, since it is now part of the Python standard
library, available as `unittest.mock`.

Reuse approach used in generated `tests/unit/gapic/monitoring_v3/` for
manually written `tests/unit/test_query.py`.

This issue and solution was found as a part of packaging the library for
Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=2263960.

Fixes #12316 🦕

### DIFF
--- a/packages/google-cloud-monitoring/tests/unit/test_query.py
+++ b/packages/google-cloud-monitoring/tests/unit/test_query.py
@@ -16,13 +16,7 @@ from __future__ import absolute_import
 
 import datetime
 import unittest
-
-# try/except added for compatibility with python < 3.8
-try:
-    from unittest import mock
-    from unittest.mock import AsyncMock  # pragma: NO COVER
-except ImportError:  # pragma: NO COVER
-    import mock
+from unittest import mock
 
 from google.cloud import monitoring_v3 as monitoring_v3
 from google.cloud.monitoring_v3 import MetricServiceClient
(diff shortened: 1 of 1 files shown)


parthea (approved):

parthea:
Thanks for the fix @rominf !

PR #12652, opened 2024-05-01
chore(docs): Update .repo-metadata.json for backupdr

dandhlee:
Updating the repo.metadata.json for documentation path: https://cloud.google.com/python/docs/reference/backupdr/latest

The current link doesn't work.

Please let me know if this can be manually fixed or if it needs to be updated from the serviceconfigs.


parthea (commented):
Can we keep the current docs link and change the `name` to `google-cloud-backupdr` instead? This is what we had in the initial commit. https://github.com/googleapis/google-cloud-python/pull/12552/commits/1e9fb9d2d9f8ebc119333723f1402c65532beb99


Change
```
"name": "backupdr",
```
to
```
"name": "google-cloud-backupdr",
```

cc @ohmayr

parthea (approved):

PR #12703, opened 2024-05-16
chore: Add heuristics for issue tracker

### LINKED ISSUE #11147: GitHub Action that updates the README is broken
The [README](https://github.com/googleapis/google-cloud-python#readme) in this repository is automatically generated using a [GitHub action](https://github.com/googleapis/google-cloud-python/blob/main/.github/workflows/main.yml). The README lists all of the python packages however this README has not been updated for 3 months. The code in the GitHub action does not appear to be working as intended, as the [code being proposed by the GitHub Action](https://github.com/googleapis/google-cloud-python/pull/11135/files) is deleting all of the existing packages. In addition, [packages](https://github.com/googleapis/google-cloud-python/tree/main/packages) in this repository are not tracked.

### LINKED ISSUE #12679: include issue tracker link in the package and top-level READMEs
In each directory under `packages/`, we have a `.repo-metadata.json` file that includes a field `issue_tracker`. We should include that link in the generated package-level `README.rst`, and arguably also in the [list of packages](https://github.com/googleapis/google-cloud-python?tab=readme-ov-file#libraries) included in the top-level `README.rst`

When this is done, we can give easier instructions (eg on issue templates) on how to get support.

ohmayr:
Fixes #11147 🦕
Fixes #12679


README.rst
```diff
@@ -70,202 +70,1466 @@ Libraries
...
+     - |PyPI-google-cloud-orchestration-airflow|
+     -
+     -
+     - `Client Library Issues <https://github.com/googleapis/google-cloud-python/issues>`_
+   * - `Compute Engine <https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-compute>`_
+     - stable
      - |PyPI-google-cloud-compute|
-   * - `Container Analysis <https://github.com/googleapis/python-containeranalysis>`_
-     - |stable|
+     - `API Issues <None>`_
```
vchudnov-g, line 247: We should omit the link if it's `None`. This is happening in a few places.

scripts/updateapilist.py
```diff
@@ -75,6 +82,39 @@ def __init__(self, repo: dict):
         self.release_level = repo["release_level"]
         self.distribution_name = repo["distribution_name"]
         self.issue_tracker = repo.get("issue_tracker")
+    
+    @property
+    def issue_tracker_component_id(self):
+        if not self.issue_tracker:
+            return None
+        match = re.search(r'(?:(?:\?|&)component=(\d+)|\bcomponentid:(\d+))', self.issue_tracker)
```
vchudnov-g, line 90: You put this in one regex? OK, cool. We could have left it in two for simplicity, but either way works.

scripts/updateapilist.py
```diff
@@ -75,6 +82,39 @@ def __init__(self, repo: dict):
...
+        if match:
+            return match.group(1) or match.group(2)
+        return None
+    
+    @property
+    def issue_tracker_template_id(self):
```
vchudnov-g, line 96: Is this used anywhere?

vchudnov-g: In the create-issue URL, we should use the issue tracker if it was specified, but otherwise omit this query param.

ohmayr: Do you mean to say that we should use the `template_id` if specified?

That's what my plan is. This function retrieves the template id from the `issue_tracker`. I will be incorporating that in the URL.

README.rst
ohmayr: This was happening because of a missing `return` in one of the functions. Should be fixed now. Let me know if you think otherwise.

scripts/updateapilist.py
ohmayr: I've split the regex into two. Thanks @vchudnov-g!

scripts/updateapilist.py
```diff
@@ -75,6 +82,71 @@ def __init__(self, repo: dict):
...
+
+    @property
+    def issue_tracker_component_id(self):
+        # First, check if the issue tracker is a saved search:
+        query_string = self.issue_tracker
+        if self.saved_search_id:
+            query_string = self.saved_search_response_text()        
+        if not query_string:
+            return None
+        # Try to match 'component=' in the query string
```
vchudnov-g, line 113: For the saved searches, where do you deal with multiple components? We want to ensure we don't return 187065, the generic GCP component. If this is working so far, it's because we got lucky and that's the second component in the search string, I guess. We should be more deliberate.

vchudnov-g (commented):
Looks great! We're almost there. Just a few comments. Most are minor, but I do think it matters we be robust about handling multiple components in the saved search.

scripts/updateapilist.py
```diff
@@ -67,6 +70,10 @@ class CloudClient:
     release_level: str = None
     distribution_name: str = None
     issue_tracker: str = None
+    file_an_issue: str = None
+    already_filed_issue: str = None
+    github_issues: str = None
```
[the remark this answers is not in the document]
ohmayr: These had to cleaned up. However, I updated the property methods to reflect this.

scripts/updateapilist.py
```diff
@@ -75,6 +82,71 @@ def __init__(self, repo: dict):
...
+    @property
+    def saved_search_id(self):
+        if not self.issue_tracker:
+            return None
+        match = re.search(r'savedsearches/(\d+)', self.issue_tracker)
+        if match:
+            return match.group(1)
+        return None
+    
+    def saved_search_response_text(self):
```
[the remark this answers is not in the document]
ohmayr: Addressed.

scripts/updateapilist.py
```diff
@@ -75,6 +82,71 @@ def __init__(self, repo: dict):
...
+                return None
+            return response.text
+        return None
+
+    @property
+    def issue_tracker_component_id(self):
+        # First, check if the issue tracker is a saved search:
+        query_string = self.issue_tracker
+        if self.saved_search_id:
+            query_string = self.saved_search_response_text()
```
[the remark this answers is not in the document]
ohmayr: Addressed.

scripts/updateapilist.py
ohmayr: Addressed.

scripts/updateapilist.py
```diff
@@ -57,8 +58,14 @@ class MissingGithubToken(ValueError):
 # BASE_API defines the base API for Github.
 BASE_API = "https://api.github.com"
 
+# GITHUB_ISSUES defines the issues url for a repository on Github.
+GITHUB_ISSUES = "https://github.com/{repo}/issues"
 
+# BASE_ISSUE_TRACKER defines the base url for issue tracker.
+BASE_ISSUE_TRACKER = "https://issuetracker.google.com"
 
+# GENERIC_ISSUE_TRACKER_COMPONENT defines a generic component for issue tracker.
```
vchudnov-g, line 67: This comment will leave us/others puzzled in the future. Let's briefly expand why we care:

"This issue-tracker component is part of some saved searches for listing API-side issues. However, when we construct URLs for filing new issues (which in some cases we do by analyzing the query string for a saved search), we want to ensure we DON'T file a new issue against this generic component but against a more specific one."

scripts/updateapilist.py
```diff
@@ -68,13 +75,79 @@ class CloudClient:
...
+        # Try to match 'component=' in the query string
+        query_match = re.search(r'\bcomponent=(\d+)', query_string)
+        if query_match:
+            return query_match.group(1)
+        # If not found, try to match 'componentid:' in the query string
+        query_match = re.findall(r'\bcomponentid:(\d+)', query_string)
+        for component_id in query_match:
+            if component_id != GENERIC_ISSUE_TRACKER_COMPONENT:
+                return component_id
+        return None
```
vchudnov-g, line 122:
```suggestion
        target_component = None
        for component_id in query_match:
            if component_id == GENERIC_ISSUE_TRACKER_COMPONENT:
                continue
            if target_component:
              # We extracted two possible target components. We don't want to use the wrong one.
              return None
            target_component=component_id
            
        return target_component
```

scripts/updateapilist.py
```diff
@@ -138,13 +212,13 @@ def client_row(client: CloudClient) -> str:
 
     content_row = [
         f"   * - `{client.title} <{url}>`_\n",
-        f"     - " + client.release_level + "\n",
-        f"     - |PyPI-{client.distribution_name}|\n",   
+        f"     - {client.release_level}\n",
+        f"     - |PyPI-{client.distribution_name}|\n",
+        f"     - `API Issues <{client.show_api_issues}>`_\n" if client.issue_tracker_component_id else "     -\n",
```
vchudnov-g, line 217: It would be cleaner to reference that same object in the string and the conditional. Otherwise you're breaking encapsulation, knowing that if there's a component then the resulting string exists.

```python
        f"     - `API Issues <{client.FOO}>`_\n" if client.FOO else "     -\n",
```

And it might make sense to assign these values to local variables first so you don't call and run any getters repeatedly in these lines. (I don't know whether the interpreter can optimize that away; I would assume not.)

(Sorry, I meant to send this comment last night but I forgot to do the final click)

vchudnov-g (commented):
Two minor code suggestions and two doc suggestions.

Also, could you modify `/.github/workflows/main.yml` so that it has a `workflow_dispatch` trigger, so we can kick off this workflow from the GitHub UI when we need to? (It will also be a quick way of seeing this take effect after we merge!) It's just adding a line identical to [this one](https://github.com/googleapis/google-api-python-client/blob/4dabf68df2016be3aed62968975d26158fcfd68e/.github/workflows/main.yml#L23)

README.rst
```diff
@@ -70,202 +70,1466 @@ Libraries
...
+     - `Client Library Issues <https://github.com/googleapis/googleapis/python-crc32c/issues>`_
    * - `AI Platform <https://github.com/googleapis/python-aiplatform>`_
-     - |stable|
+     - stable
      - |PyPI-google-cloud-aiplatform|
-   * - `App Engine Admin <https://github.com/googleapis/python-appengine-admin>`_
-     - |stable|
+     - `API Issues <https://issuetracker.google.com/issues?q=componentid:187220>`_
+     - `File an API Issue <https://issuetracker.google.com/issues/new?component=187220>`_
+     - `Client Library Issues <https://github.com/googleapis/googleapis/python-aiplatform/issues>`_
```
parthea, line 87: This link results in 404

scripts/updateapilist.py
```diff
@@ -68,13 +75,79 @@ class CloudClient:
...
+        return None
+    
+    @property
+    def saved_search_response_text(self):
+        id = self.saved_search_id
+        if id:
+            url = f"{BASE_ISSUE_TRACKER}/action/saved_searches/{id}"
+            response = requests.get(url=url)
+            if response.status_code != requests.codes.ok:
+                return None
```
parthea, line 103: Should we include information in the logs when we get other response codes?

vchudnov-g: That's a good idea. We can also leave info for other error conditions: eg if there's more than one non-generic component in a saved search.

What I was also thinking is that we can fail the action (by having the script return a non-zero status code) if we run into (some?) problems. Ideally we would do this whenever there's an empty cell in the table....but it doesn't make sense to do that now because there are so many empties. But once we do the audit, act on it, and the table is fully filled out, allowing the action to fail on error would make sense, probably.

README.rst
vchudnov-g: Hmm: maybe before we put each generated URL in the table we should verify it to make sure it yields (or redirects to) a valid response. So in my local-variables idea, something like this

```python
list_issues = validate_url(client.show_api_issues)
f"     - `API Issues <{list_issues}>`_\n" if list_issues else "     -\n",
```

etc.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

README.rst
ohmayr: This issue was due to an error in the generated path i.e. `googleapis/googleapis/python-aiplatform`. I fixed the related function after verifying that both mono and split repos have repo defined in metadata file as `googleapis/<repo_name>`. The links for all repos/packages should be corrected now.

I think still adding an extra layer to guard the generated URL may be beneficial but we may be alright otherwise given that we'll be auditing the metadata files. Regardless, the URLs  for `"Client Library Issues"` should all be correct in the current state.

scripts/updateapilist.py
ohmayr: I've addressed this by adding error logs for each request that returns a non 200 response. Let me know what you guys think and if we want any of these errors to be more specific!

scripts/updateapilist.py
```diff
@@ -57,8 +62,17 @@ class MissingGithubToken(ValueError):
...
 
+# GITHUB_ISSUES defines the issues url for a repository on Github.
+GITHUB_ISSUES = "https://github.com/{repo}/issues"
 
+# BASE_ISSUE_TRACKER defines the base url for issue tracker.
+BASE_ISSUE_TRACKER = "https://issuetracker.google.com"
 
+# This issue-tracker component is part of some saved searches for listing API-side issues.
+# However, when we construct URLs for filing new issues (which in some cases we do by analyzing
+# the query string for a saved search),we want to ensure we DON'T file a new issue against this
```
vchudnov-g, line 73:
```suggestion
# the query string for a saved search), we want to ensure we DON'T file a new issue against this
```

scripts/updateapilist.py
```diff
@@ -68,13 +82,99 @@ class CloudClient:
...
+
     def __init__(self, repo: dict):
         self.repo = repo["repo"]
         # For now, strip out "Google Cloud" to standardize the titles
         self.title = repo["name_pretty"].replace("Google ", "").replace("Cloud ", "")
         self.release_level = repo["release_level"]
         self.distribution_name = repo["distribution_name"]
         self.issue_tracker = repo.get("issue_tracker")
+        self._cached_component_id = None
+        self._is_component_id_computed = False
```
vchudnov-g, line 94: What you have is totally fine, but you could be more succinct if you want.

```suggestion
        self._component = None
        self._computed_component = False
```

Do consider clarifying you have the second field because the intentionally cached value of the previous one may be `None`.

(Note that you could potentially combine the two into a tuple, but what you have is probably clearer.)

scripts/updateapilist.py
```diff
@@ -68,13 +82,99 @@ class CloudClient:
...
+            return match.group(1)
+        return None
+    
+    @property
+    def show_client_issues(self):
+        return GITHUB_ISSUES.format(repo=self.repo)
+    
+    @property
+    def file_api_issue(self):
+        component = self.issue_tracker_component_id
```
vchudnov-g, line 163: `self.issue_tracker_component_id` is now memoizing its value, so the local variable `component` isn't buying you much any more. (In fact, it's not buying new anything because you only use it one place)

scripts/updateapilist.py
```diff
@@ -68,13 +82,99 @@ class CloudClient:
...
+            link = f"{BASE_ISSUE_TRACKER}/issues/new?component={self.issue_tracker_component_id}"
+            template_id = self.issue_tracker_template_id
+            if template_id:
+                link += f"&template={template_id}"
+            return link
+        return None
+    
+    @property
+    def show_api_issues(self):
+        component = self.issue_tracker_component_id
```
vchudnov-g, line 174: `self.issue_tracker_component_id` is now memoizing its value, so the local variable `component` isn't buying you much any more. (In fact, it's not buying you anything because you only use it one place)

scripts/updateapilist.py
```diff
@@ -68,13 +82,99 @@ class CloudClient:
...
+    @property
+    def show_client_issues(self):
+        return GITHUB_ISSUES.format(repo=self.repo)
+    
+    @property
+    def file_api_issue(self):
+        component = self.issue_tracker_component_id
+        if component:
+            link = f"{BASE_ISSUE_TRACKER}/issues/new?component={self.issue_tracker_component_id}"
+            template_id = self.issue_tracker_template_id
```
vchudnov-g, line 166: Nit suggestion: Since you're memoizing `self.issue_tracker_component_id`, you could similarly memoize `self.issue_tracker_template_id` and not have to use a local variable.

Up to you; what you have is fine.

scripts/updateapilist.py
```diff
@@ -95,16 +195,24 @@ class Extractor:
     def client_for_repo(self, repo_slug) -> Optional[CloudClient]:
         path = self.path_format.format(repo_slug=repo_slug)
         url = f"{RAW_CONTENT_BASE_URL}/{path}/{REPO_METADATA_FILENAME}"
-        response = requests.get(url)
-        if response.status_code != requests.codes.ok:
-            return
-
-        return CloudClient(response.json())
+        try:
```
vchudnov-g, line 198: You have this pattern in multiple places. Factor it out into a helper function: `get_jason(url, params=None)`. It can set the headers (since they seem to be the same in all calls), log any errors, and decode JSON, returning the results.

scripts/updateapilist.py
```diff
@@ -181,30 +292,53 @@ def mono_repo_clients(token: str) -> List[CloudClient]:
...
+
+        repositories = response_json.get("items", [])
         if len(repositories) == 0:
             break
 
         split_repo_extractor = Extractor(path_format=SPLIT_REPO_PATH_FORMAT, response_key=REPO_RESPONSE_KEY)
-        return split_repo_extractor.get_clients_from_batch_response(repositories)
+        clients.extend(split_repo_extractor.get_clients_from_batch_response(repositories))
+
+        # Check for the 'next' link in the response headers for pagination
```
vchudnov-g, line 337: Were we not getting additional pages in previous iterations of the PR? Glad you caught this!

scripts/updateapilist.py
```diff
@@ -181,30 +292,53 @@ def mono_repo_clients(token: str) -> List[CloudClient]:
...
+        next_link = response.links.get('next', {}).get('url')
+        url = next_link if next_link else None
```
vchudnov-g, line 339: I think this will work the same:
```suggestion
        url = next_link
```

scripts/updateapilist.py
```diff
@@ -135,16 +243,17 @@ def client_row(client: CloudClient) -> str:
     url = f"https://github.com/{client.repo}"
     if client.repo == MONO_REPO:
         url += f"/tree/main/packages/{client.distribution_name}"
-
+    _show_api_issues = client.show_api_issues
+    _file_api_issue = client.file_api_issue
```
vchudnov-g, line 247: (No action needed--just a comment)
You could also memoize the two properties on the RHS. The complication is that for this application, you need two pieces of information for the memoization (as you implemented above), and that's because the cached value could be `None`. In many applications, the memoized value is not `None`, so the value `None` can stand for "not yet cached".  So I don't think it's worth pursuing here, because it adds two fields to the class, and this is certainly clear enough.

scripts/updateapilist.py
```diff
@@ -68,13 +82,99 @@ class CloudClient:
...
+        if component:
+            link = f"{BASE_ISSUE_TRACKER}/issues/new?component={self.issue_tracker_component_id}"
+            template_id = self.issue_tracker_template_id
+            if template_id:
+                link += f"&template={template_id}"
+            return link
+        return None
+    
+    @property
+    def show_api_issues(self):
```
vchudnov-g, line 173: Oh, this is something I should have noticed earlier: if what we got from the metadata file is a saved search, we should use that saved search as the return value here. We only extract the component from those to construct the file-an-issue URL.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
ohmayr: addressed.

scripts/updateapilist.py
```diff
@@ -70,10 +70,12 @@ class MissingGithubToken(ValueError):
 
 # This issue-tracker component is part of some saved searches for listing API-side issues.
 # However, when we construct URLs for filing new issues (which in some cases we do by analyzing
-# the query string for a saved search),we want to ensure we DON'T file a new issue against this
-# generic component but against a more specific one.
+# the query string for a saved search), we want to ensure we DON'T file a new issue against
+# this generic component but against a more specific one.
 GENERIC_ISSUE_TRACKER_COMPONENT = "187065"
 
+# This field is used to avoid re-computing an already computed property.
```
vchudnov-g, line 77:
```suggestion
# This sentinel value is used to mark cache fields that have not been computed yet.
```

scripts/updateapilist.py
```diff
@@ -137,42 +135,43 @@ def issue_tracker_component_id(self):
                 for component_id in query_match:
                     if component_id == GENERIC_ISSUE_TRACKER_COMPONENT:
                         continue
-                    if self._cached_component_id:
+                    if self._cached_component_id != NOT_COMPUTED:
                         self._cached_component_id = None
                         logging.error(f"More than one component ids found for issue tracker: {self.issue_tracker}")
```
vchudnov-g, line 140: Nit

```suggestion
                        logging.error(f"More than one component ID found for issue tracker: {self.issue_tracker}")
```

scripts/updateapilist.py
```diff
@@ -292,20 +303,11 @@ def mono_repo_clients(token: str) -> List[CloudClient]:
     # all mono repo clients
     url = f"{BASE_API}/repos/{MONO_REPO}/contents/packages"
     headers = {'Authorization': f'token {token}'}
-    try:
-        response = requests.get(url=url, headers=headers)
-        response.raise_for_status()
-    except requests.RequestException as e:
-        logging.error(f"Request failed for URL {url}: {e}")
+    _, metadata = _fetch_and_parse_response(url, headers)
```
vchudnov-g, line 306: Nit nit (not a blocker; you can leave as is)

```suggestion
    _, packages = _fetch_and_parse_response(url, headers)
```

scripts/updateapilist.py
```diff
@@ -137,42 +135,43 @@ def issue_tracker_component_id(self):
...
+                link += f"&template={self.issue_tracker_template_id}"
             return link
         return None
     
     @property
     def show_api_issues(self):
-        component = self.issue_tracker_component_id
-        if component:
+        if self.saved_search_id:
+            return self.issue_tracker
```
vchudnov-g, line 173:
```suggestion
        if self.saved_search_id:
            # Return the original issue_tracker content, which already links to the saved search.
            return self.issue_tracker
```

scripts/updateapilist.py
```diff
@@ -57,9 +62,20 @@ class MissingGithubToken(ValueError):
 # BASE_API defines the base API for Github.
 BASE_API = "https://api.github.com"
 
+# GITHUB_ISSUES defines the issues url for a repository on Github.
```
vchudnov-g, line 65: Nit

```suggestion
# GITHUB_ISSUES defines the issues URL for a repository on GitHub.
```

scripts/updateapilist.py
```diff
@@ -57,9 +62,20 @@ class MissingGithubToken(ValueError):
...
+GITHUB_ISSUES = "https://github.com/{repo}/issues"
 
+# BASE_ISSUE_TRACKER defines the base url for issue tracker.
```
vchudnov-g, line 68: Nit: 
```suggestion
# BASE_ISSUE_TRACKER defines the base URL for issue tracker.
```

vchudnov-g (approved):
Just some cosmetic comments. Two of them are non-nit clarifications; please address. The others are nits; up to you.

Thank you so much for doing this! This looks like really solid code.

README.rst
```diff
@@ -70,202 +70,1494 @@ Libraries
...
+     - |PyPI-google-cloud-api-gateway|
+     -
+     -
+     - `Client Library Issues <https://github.com/googleapis/google-cloud-python/issues>`_
+   * - `Access Approval <https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-access-approval>`_
+     - stable
+     - |PyPI-google-cloud-access-approval|
+     -
+     -
+     - `Client Library Issues <https://github.com/googleapis/google-cloud-python/issues>`_
```
parthea, line 105: Non-blocking comment. Please feel free to file a bug as follow up work

For projects where we are creating issues directly in google-cloud-python we could update this link to point to template for filing an issue: For example, https://github.com/googleapis/google-cloud-python/issues/new?assignees=&labels=&projects=&template=bug_report.md

One benefit of this is that we can set the `labels` parameter in the url which is empty by default.

For example, we could add the `type: bug` label and `api: access approval` label , using url encoding for the colon and space characters.
https://github.com/googleapis/google-cloud-python/issues/new?assignees=&labels=type%3A%20bug%2Capi%3A%20accessapproval&projects=&template=bug_report.md

As long as the priority remains unset, the bug will be considered un-triaged and will still appear in the un-triaged list.

parthea (approved):

README.rst
vchudnov-g: I suggest we do this as a follow-up after we have the new issue templates set up.

ohmayr: https://github.com/googleapis/google-cloud-python/issues/12723

PR #12710, opened 2024-05-17
chore: improve issue templates and directions for filing in correct repo

### LINKED ISSUE #12651: GitHub issue templates: clarify; provide easy access to back-end support channels
1. Our current bug template says `Google Cloud Python client libraries are no longer maintained inside this repository. Please visit the python-API repository (e.g., https://github.com/googleapis/python-pubsub/issues) for faster response times.` That seems confusing. Let's clarify.
2. Let's include in our issue templates information on how to more quickly access service-level support rather than client library support. This should be customized to each type of template.

vchudnov-g:
Fixes #12651

TAL @ [the main branch on my fork](https://github.com/vchudnov-g/google-cloud-python/issues/new/choose) to preview the experience.

**~TODO~DONE**: When this has been approved, and prior to merging, I'll update all links to refer to the main (upstream) repo and not my fork.


SUPPORT.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Filing an API service issue
+
+*(if the issue DID re-occur [above](#org4510fe1) without using the client library)*
+
+Use the appropriate API service issue tracker. The maintainers of this repo are not experts or contributors on individual API services; we work on generating usable and idiomatic client libraries for many APIs.
+
+1.  Consider using [Google Cloud Customer Care](https://cloud.google.com/support/?hl=en) (paid) to get more dedicated support for your issue. Otherwise, continue with the steps below.
+2.  Find your API in [the client list at the top of this repository](https://github.com/googleapis/google-cloud-python/tree/main?tab=readme-ov-file#libraries).
+3.  Check the “Issue Tracker” column to see whether someone else has reported the same issue. If they have, the filed issue may have some useful information; feel free to add more details or context of your own.
```
parthea, line 40: Should we add a note that folks can file API issues in the corresponding client library repository in the case where there is no API specific issue tracker?

parthea (commented):
LGTM, just 1 observation. I'll re-review once this is ready

vchudnov-g:
Adding `do not merge` until this has been approved and I then change the links to point to the main repo and not my fork. In the meantime, reviewers can experience the new templates by going to my [fork](https://github.com/vchudnov-g/google-cloud-python/issues/new/choose).

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,93 @@
...
+    
+        **Expected Behavior:**
+        _e.g. I expected method Foo to paginate_
+
+        **Actual Behavior:**
+        _e.g. Method Foo did not paginate_        
+- type: input
+  attributes:
+    label: API client name and version
+    description: Try `pip list | grep google-cloud`, which should give you a name corresponding to one of the packages listed [here](https://github.com/googleapis/google-cloud-python/tree/main/packages).
```
parthea, line 33:
```suggestion
    description: Try `pip list | grep google`, which should give you a name corresponding to one of the packages listed [here](https://github.com/googleapis/google-cloud-python/tree/main/packages).
```

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,47 @@
...
+- type: markdown
+  attributes:
+    value: |
+        _**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response._
+- type: checkboxes
+  attributes:
+    label: Determine this is the right repository
+    description: Please follow the steps in [the SUPPORT guide](https://github.com/vchudnov-g/google-cloud-python/blob/issue-templates/SUPPORT.md) and determine that this is likely a client library issue and this is the correct repository in which to report it.
+    options:
+      - label: I determined this is the correct repository in which to report this bug
```
parthea, line 14:
```suggestion
      - label: I determined this is the correct repository in which to report this feature request.
```

SUPPORT.md
```diff
@@ -0,0 +1,60 @@
...
+
+To help us resolve your issue as soon as possible, please follow these steps:
+
+
+<a id="org49b5b6e"></a>
+
+# Check for already opened issues:
+
+-   In `google-cloud-python`: <https://github.com/googleapis/google-cloud-python/issues>
+-   In Stack Overflow: <https://stackoverflow.com/questions/tagged/google-cloud-platform+python>
```
parthea, line 11: Do we want to list `Stack Overflow` here? Does it belong in the bug intake process?

parthea (approved):

SUPPORT.md
vchudnov-g: I was basing this off of what NodeJS is doing: https://github.com/googleapis/google-cloud-node/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml

We can take it out, though I think there could be value in pointing people to "easy" places to search for issues. WDYT?

.github/ISSUE_TEMPLATE/bug_report.yaml
vchudnov-g: Good idea. Did it with a trailing hyphen. WDYT?

.github/ISSUE_TEMPLATE/feature_request.yaml
vchudnov-g: Good catch! Done.

.github/ISSUE_TEMPLATE/bug_report.yaml
parthea: It may be helpful without the trailing hyphen as an additional package `googleapis-common-protos` will be included

SUPPORT.md
parthea: I'm concerned that users won't report issue in Github if they're already in Stack Overflow. As an example, see https://github.com/googleapis/google-cloud-python/issues/12749#issuecomment-2138157488 which mentions that an issue was reported on Stack Overflow `a month ago` and only reported in this repository in the last week. If Stack Overflow is not part of our bug intake, then we shouldn't mention it here.

vchudnov-g: Hmm, I see your point. The purpose was to help users help themselves more quickly, but you're right, it does run the risk that even if the community finds a workaround, we won't hear about it to incorporate it or fix the code because no formal issues are filed. OK, I'll take it out, though I do think some reference to searching on SO might be good. But we have to determine the right phrasing.

.github/ISSUE_TEMPLATE/bug_report.yaml
vchudnov-g: OK, good catch! Will do.

SUPPORT.md
vchudnov-g: Looking at the `python-api-common-protos` [bug template](https://github.com/googleapis/python-api-common-protos/issues/new?assignees=&labels=&projects=&template=bug_report.md), they say this: maybe we should incorporate it? WDYT, @parthea ?

```
Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-api-common-protos/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:
```

parthea: This wording sounds good to me. We should encourage users to open issues for bugs, even if there are workarounds/quick fixes. It would be good to incorporate that as well.

vchudnov-g: OK, I'm submitting without the SO reference, and we can wordsmith that in via a chaser PR.

Issue #12721, opened 2024-05-21
determine publishing policy for APIs with "experimental" parts, possibly via a mechanism to pin their libraries as "preview"


vchudnov-g:
Determine whether we can merge and release #12709 even though some of the proto messages are marked as Experimental.

One possibility is to publish this library in a "preview" state and leave it as such until the "Experimental" code becomes stable. If we go with this approach, we would need to figure out how to pin the library as "preview" so we don't accidentally promote it to GA.

vchudnov-g:
We seem to be converging towards dong this. So to be explicit, the first library we would pin (mechanism TBD) to remain "preview" is
* `packages/google-maps-places/`

daniel-sanche:
This issue is being closed as part of a bulk review (Filter #1). Please reopen if you believe this is still relevant and requires action.

PR #12734, opened 2024-05-24
chore: pin google-maps-places to stay in "preview" mode

vchudnov-g:
Towards #12721 

This is a proposed pinning file to be respected by a proposed metadata update mechanism. Googlers: see internal design docs.


parthea (approved):
LGTM

Issue #11437, opened 2023-06-28
Introduce deprecation warning for python 3.7


parthea:
Python 3.7 is [now unsupported](https://devguide.python.org/versions/#unsupported-versions) in the python community. We should update the README to indicate that python 3.7 support will be dropped and also introduce a deprecation warning with the same information.

parthea:
Closing as obsolete. This was completed

PR #13377, opened 2024-12-19
build: use ubuntu 22.04 for tests

parthea:
This PR resolves the following issue [when running Python 3.7 tests](https://github.com/googleapis/google-cloud-python/actions/runs/12406551833/job/34635201044?pr=13374) :

```
  Version 3.7 was not found in the local cache
  Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.
```

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -13,6 +13,7 @@ permissions:
 
 jobs:
   unit:
+    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2303): use `ubuntu-latest` once this bug is fixed.
     # Use ubuntu-22.04 until Python 3.7 is removed from the test matrix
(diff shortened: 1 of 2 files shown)


.github/workflows/unittest.yml
```diff
@@ -13,7 +13,9 @@ permissions:
 
 jobs:
   unit:
-    runs-on: ubuntu-latest
+    # Use ubuntu-22.04 until Python 3.7 is removed from the test matrix
```
vchudnov-g, line 16: Could we add a TODO and link to an umbrella issue for this? A quick search suggests the only currently filed issue for the 3.7 deprecation is #11437, but that deals more with messaging the deprecation. We should have an issue for actually removing support when the time comes, and part of that should be removing this Ubuntu specification.

vchudnov-g (approved):

.github/workflows/unittest.yml
parthea: Done in https://github.com/googleapis/google-cloud-python/pull/13377/commits/84387efc591a3162dec24bc64a527a3b556f4b5a

PR #13425, opened 2025-01-10
chore(migration): Migrate code from googleapis/python-access-context-manager into packages/google-cloud-access-context-manager

parthea:
See #10887. 


packages/google-cloud-access-context-manager/.repo-metadata.json
```diff
@@ -0,0 +1,15 @@
...
+    "name_pretty": "Access Context Manager",
+    "client_documentation": "https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-access-context-manager",
+    "product_documentation": "https://cloud.google.com/access-context-manager/docs/overview",
+    "issue_tracker": "https://github.com/googleapis/google-cloud-python/issues",
+    "release_level": "preview",
+    "language": "python",
+    "library_type": "GAPIC_AUTO",
+    "repo": "googleapis/google-cloud-python",
+    "distribution_name": "google-cloud-access-context-manager",
+    "default_version": "apiVersion",
```
vchudnov-g, line 12: `apiVersion` is a replacement string, right? At what point does that get replaced with the right version?

packages/google-cloud-access-context-manager/docs/type.rst
```diff
@@ -0,0 +1,6 @@
+google.identity.accesscontextmanager.type
+=========================================
+
+.. automodule:: google.identity.accesscontextmanager.type
```
vchudnov-g, line 4: A bunch of these `type.rst` files used to have these lines ending in `….type.XXXX_pb2`, but now they're truncated to end in `….type`. I take it this is expected?

vchudnov-g (approved):

packages/google-cloud-access-context-manager/docs/type.rst
parthea: Yes. This is intended. For now the docs for `pb2.py` files are not meaningful at the moment. We may use `pb2.pyi` files instead for docs.  b/375167345 is the tracking bug to improve docs in `pb2.pyi`. The protobuf team has indicated that docs (comments) won't be added to `pb2.py` files.

Issue #11042, opened 2023-04-05
Adopt split repo: _python-grpc-google-iam-v1_


vchudnov-g:
Migrate the split-repo https://github.com/googleapis/python-grpc-google-iam-v1 to https://github.com/googleapis/google-cloud-python.

The migration readiness criteria are the following, which we track via GitHub labels on this issue. These criteria apply to the split repo we are migrating from:
- No open issues
- No open PRs
- No handwritten samples
- No system tests
- No client-specific customizations

parthea:
Marking this issue as blocked because the repository does not have code generated using bazel rules.

parthea:
There are no plans to migrate this client to the monorepo.  I'm going to close this issue but we can re-open it if more discussion is needed.

PR #13427, opened 2025-01-10
chore(migration): Migrate code from googleapis/python-grpc-google-iam-v1 into packages/grpc-google-iam-v1

parthea:
See #11042. 


packages/grpc-google-iam-v1/README.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Quick Start
+-----------
+
+In order to use this library, you first need to go through the following steps:
+
+1. `Select or create a Cloud Platform project.`_
+2. `Enable billing for your project.`_
+3. `Enable the Cloud Identity and Access Management.`_
+4. `Setup Authentication.`_
```
vchudnov-g, line 29: In this and the other PRs, "Setup" should be two words: "Set up"

vchudnov-g (approved):

packages/grpc-google-iam-v1/README.rst
parthea: This will be addressed in https://github.com/googleapis/synthtool/pull/2051

PR #13436, opened 2025-01-14
build: cater for packages without gapic_version.py in release please config action

parthea:
The Github Action that updates the [release please configuration](https://github.com/googleapis/google-cloud-python/blob/main/.github/workflows/configure_release_please.yml) recently started failing after PRs #13425, #13426 and #13427 were merged. The reason is that the action expects packages to have a file called `gapic_version.py`, however these packages do not. This PR updates the release please configuration script to exclude packages which do not have this file.


scripts/configure_release_please/configure_release_please.py
```diff
@@ -70,6 +76,8 @@ def configure_release_please_manifest(
     with open(release_please_manifest, "r") as f:
         manifest_json = json.load(f)
         for package_dir in package_dirs:
+            if not package_supports_gapic_version(package_dir):
```
ohmayr, line 79: For the case when `gapic_version` is supported by the package, we first verify it on line 79 and then again on line 85.


Does the github action fail at line 81? If so, Would it suffice if we move the condition at line 85 above?

parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/13436/commits/67cf899a73937e7bca7073778fd1717bccf75652

scripts/configure_release_please/configure_release_please.py
```diff
@@ -75,7 +81,10 @@ def configure_release_please_manifest(
 
             gapic_version_file = next(package_dir.rglob("**/gapic_version.py"), None)
             if gapic_version_file is None:
-                raise Exception("Failed to find gapic_version.py")
+                if package_supports_gapic_version(package_dir):
+                    raise Exception("Failed to find gapic_version.py")
+                else:
+                    continue
```
ohmayr, line 87: Hmm my thought was that we simply continue if `gapic_version.py` file doesn't exist instead of raising an exception + another check for a supported package. But if we do want to be explicit, how about we only check for a `gapic_version_file` if it is a supported package?

```suggestion
if not package_supports_gapic_version(package_dir):
    continue
gapic_version_file = next(package_dir.rglob("**/gapic_version.py"), None)
if gapic_version file is None:
    raise Exception("Failed to find gapic_version.py")
```

ohmayr (approved):

scripts/configure_release_please/configure_release_please.py
parthea: Done in https://github.com/googleapis/google-cloud-python/pull/13436/commits/13bef06915147343e2c868b01169970c3fc7db79

PR #13510, opened 2025-02-10
build: add a new presubmit core_deps_from_source

parthea:
This PR adds a new presubmit `core_deps_from_source` which depends on https://github.com/googleapis/synthtool/pull/2061

The presubmit will install core dependencies from source in order to get early feedback on changes made at HEAD of dependencies.


parthea:
Wait for https://github.com/googleapis/synthtool/pull/2061

ohmayr (approved):

PR #13637, opened 2025-03-10
fix: Allow protobuf 6.x

parthea:
Apply changes from https://github.com/googleapis/gapic-generator-python/pull/2352 and https://github.com/googleapis/gapic-generator-python/pull/2347 to 4 packages containing protobuf stubs which are not automatically generated using `gapic-generator-python`

BEGIN_COMMIT_OVERRIDE
fix: Allow protobuf 6.x
fix: resolve issue where pre-release versions of dependencies are installed
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/google-cloud-access-context-manager/noxfile.py
+++ b/packages/google-cloud-access-context-manager/noxfile.py
@@ -382,20 +382,29 @@ def docfx(session):
     ["python", "upb", "cpp"],
 )
 def prerelease_deps(session, protobuf_implementation):
-    """Run all tests with prerelease versions of dependencies installed."""
+    """
+    Run all tests with pre-release versions of dependencies installed
+    rather than the standard non pre-release versions.
+    Pre-releases versions can be installed using
+    `pip install --pre <package>`.
+    """
 
     if protobuf_implementation == "cpp" and session.python in ("3.11", "3.12", "3.13"):
         session.skip("cpp implementation is not supported in python 3.11+")
 
     # Install all dependencies
-    session.install("-e", ".[all, tests, tracing]")
+    session.install("-e", ".")
+
     unit_deps_all = UNIT_TEST_STANDARD_DEPENDENCIES + UNIT_TEST_EXTERNAL_DEPENDENCIES
+    # Install dependencies for the unit test environment
     session.install(*unit_deps_all)
+
     system_deps_all = (
         SYSTEM_TEST_STANDARD_DEPENDENCIES
         + SYSTEM_TEST_EXTERNAL_DEPENDENCIES
         + SYSTEM_TEST_EXTRAS
     )
+    # Install dependencies for the system test environment
     session.install(*system_deps_all)
 
     # Because we test minimum dependency versions on the minimum Python
@@ -417,6 +426,7 @@ def prerelease_deps(session, protobuf_implementation):
         )
     ]
 
+    # Install dependencies specified in `testing/constraints-X.txt`.
     session.install(*constraints_deps)
 
     prerel_deps = [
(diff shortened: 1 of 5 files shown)


parthea:
```
~/work/google-cloud-python/google-cloud-python/packages/google-cloud-access-context-manager ~/work/google-cloud-python/google-cloud-python
nox > Error while collecting sessions.
nox > Sessions not found: core_deps_from_source
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -382,20 +382,29 @@ def docfx(session):
     ["python", "upb", "cpp"],
 )
 def prerelease_deps(session, protobuf_implementation):
-    """Run all tests with prerelease versions of dependencies installed."""
+    """
+    Run all tests with pre-release versions of dependencies installed
+    rather than the standard non pre-release versions.
+    Pre-releases versions can be installed using
```
vchudnov-g, line 388:
```suggestion
    Pre-release versions can be installed using
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -458,3 +468,70 @@ def prerelease_deps(session, protobuf_implementation):
...
+    "protobuf_implementation",
+    ["python", "upb"],
+)
+def core_deps_from_source(session, protobuf_implementation):
+    """Run all tests with local versions of core dependencies installed,
+    rather than pulling core dependencies from PyPI.
+    """
+
+    # Install all dependencies
+    session.install(".")
```
vchudnov-g, line 484: Do you want a `-e` here like in the previous test?

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -382,20 +382,29 @@ def docfx(session):
...
 
     if protobuf_implementation == "cpp" and session.python in ("3.11", "3.12", "3.13"):
         session.skip("cpp implementation is not supported in python 3.11+")
 
     # Install all dependencies
-    session.install("-e", ".[all, tests, tracing]")
+    session.install("-e", ".")
+
     unit_deps_all = UNIT_TEST_STANDARD_DEPENDENCIES + UNIT_TEST_EXTERNAL_DEPENDENCIES
+    # Install dependencies for the unit test environment
```
vchudnov-g, line 399: Consistency with the other test:

```suggestion
    # Install dependencies for the unit test environment
    unit_deps_all = UNIT_TEST_STANDARD_DEPENDENCIES + UNIT_TEST_EXTERNAL_DEPENDENCIES
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -382,20 +382,29 @@ def docfx(session):
...
     session.install(*unit_deps_all)
+
     system_deps_all = (
         SYSTEM_TEST_STANDARD_DEPENDENCIES
         + SYSTEM_TEST_EXTERNAL_DEPENDENCIES
         + SYSTEM_TEST_EXTRAS
     )
+    # Install dependencies for the system test environment
```
vchudnov-g, line 407: Consistency with the other test:

```suggestion
    # Install dependencies for the system test environment
    system_deps_all = (
        SYSTEM_TEST_STANDARD_DEPENDENCIES
        + SYSTEM_TEST_EXTERNAL_DEPENDENCIES
        + SYSTEM_TEST_EXTRAS
    )
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -458,3 +468,70 @@ def prerelease_deps(session, protobuf_implementation):
...
+    system_deps_all = (
+        SYSTEM_TEST_STANDARD_DEPENDENCIES
+        + SYSTEM_TEST_EXTERNAL_DEPENDENCIES
+        + SYSTEM_TEST_EXTRAS
+    )
+    session.install(*system_deps_all)
+
+    # Because we test minimum dependency versions on the minimum Python
+    # version, the first version we test with in the unit tests sessions has a
+    # constraints file containing all dependencies and extras that should be installed.
```
vchudnov-g, line 500: Consistency with previous

```suggestion
    # constraints file containing all dependencies and extras.
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -458,3 +468,70 @@ def prerelease_deps(session, protobuf_implementation):
...
         },
     )
+
+
+@nox.session(python="3.13")
+@nox.parametrize(
+    "protobuf_implementation",
+    ["python", "upb"],
+)
+def core_deps_from_source(session, protobuf_implementation):
```
vchudnov-g, line 478: Maybe make it clear it's the local source, as opposed to checking it out from somewhere? 

```suggestion
def core_deps_from_local_source(session, protobuf_implementation):
```

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -458,3 +468,70 @@ def prerelease_deps(session, protobuf_implementation):
...
+        match.group(1)
+        for match in re.finditer(
+            r"^\s*(\S+)(?===\S+)", constraints_text, flags=re.MULTILINE
+        )
+    ]
+
+    # Install dependencies specified in `testing/constraints-X.txt`.
+    session.install(*constraints_deps)
+
+    core_dependencies_from_source = [
```
vchudnov-g, line 520: This list is shorter than the corresponding `prerel_deps` in the previous test. Can we make that one shorter as well? Or does this list include transitive dependencies that are not included in the other? We should explain why they don't correspond.

packages/google-cloud-access-context-manager/noxfile.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/13637/commits/eeb2efd3ee59ed691e071ead09b4effffd30395b. https://github.com/googleapis/synthtool/pull/2071 will take care of updating this in all packages.

packages/google-cloud-access-context-manager/noxfile.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/13637/commits/eeb2efd3ee59ed691e071ead09b4effffd30395b. https://github.com/googleapis/synthtool/pull/2071 will take care of updating this in all packages.

packages/google-cloud-access-context-manager/noxfile.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/13637/commits/eeb2efd3ee59ed691e071ead09b4effffd30395b. https://github.com/googleapis/synthtool/pull/2071 will take care of updating this in all packages.

packages/google-cloud-access-context-manager/noxfile.py
parthea: `core_deps_from_source` seems correct to me as we are installing packages from source which is not necessarily local. For example, currently we have the following installations which are remote. I also filed TODO bugs to add installing `protobuf` and `grpc` from source, which will also be remote.
```
        "proto-plus @ git+https://github.com/googleapis/proto-plus-python.git",
        "google-api-core @ git+https://github.com/googleapis/python-api-core.git",
        "google-auth @ git+https://github.com/googleapis/google-auth-library-python.git",
```

See https://github.com/googleapis/gapic-generator-python/issues/2357 and https://github.com/googleapis/gapic-generator-python/issues/2358 for the bugs to follow up on this.

packages/google-cloud-access-context-manager/noxfile.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/13637/commits/eeb2efd3ee59ed691e071ead09b4effffd30395b. I also removed packages which are not needed (IOW the test pass regardless if they are installed). https://github.com/googleapis/synthtool/pull/2071 will take care of updating this in all packages.

packages/google-cloud-access-context-manager/noxfile.py
vchudnov-g: I see your point. I was focusing on the relative dirs you have in there......but OK, leaving the name as is does sound better. It's implied (maybe we should make it explicit) that for anything that comes from this repo, `google-cloud-python`, we'll incorporate local changes.

packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -417,39 +426,125 @@ def prerelease_deps(session, protobuf_implementation):
...
+            "google-auth": "google.auth",
+            "grpcio": "grpc",
+            "protobuf": "google.protobuf",
+            "proto-plus": "proto",
+        }
+
+        try:
+            version_namespace = package_namespaces[dep]
+        except KeyError:
+            version_namespace = None
```
vchudnov-g, line 464:
```suggestion
        version_namespace = package_namespaces.get(dep)
```

vchudnov-g (approved):
Thanks for doing this! One non-trivial (but still minor) comment about avoiding the `try`/`catch`.

PR #13840, opened 2025-04-24
Allow directly passing force to ConfigClient.delete_deployment()

### LINKED ISSUE #13839: [config] Allow directly passing force to ConfigClient.delete_deployment()

Just like `name`, allow passing `force` into `ConfigClient().delete_deployment()` instead of requiring a `DeleteDeploymentRequest` to be made if forcing is required

### Desired code experience

file: main.py
```python
   def my_ideal_code():
     ...
     client = ConfigClient(...)
     
     client.delete_deployment(name = "...", force = True)
```

### Expected results

deployment is deleted even if it contains revisions.

google-cloud-config==0.1.17

### Use case

Allowing direct passing of `force` would prevent users from being required to create a Request object if they need to force-delete.

### Additional context

_No response_

jhollowe:

Fixes #13839 🦕

### DIFF
--- a/packages/google-cloud-config/google/cloud/config_v1/services/config/client.py
+++ b/packages/google-cloud-config/google/cloud/config_v1/services/config/client.py
@@ -1385,6 +1385,7 @@ class ConfigClient(metaclass=ConfigClientMeta):
         request: Optional[Union[config.DeleteDeploymentRequest, dict]] = None,
         *,
         name: Optional[str] = None,
+        force: Optional[bool] = None,
         retry: OptionalRetry = gapic_v1.method.DEFAULT,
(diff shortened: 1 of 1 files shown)


vchudnov-g:
Thanks for your PR! However, this package, like most packages in this repo, is auto-generated, so your changes would be overwritten when there's an API update. If our libraries are not being generated with some needed functionality, could you [file an issue](https://github.com/googleapis/google-cloud-python/issues) detailing what's needed?

Thank you!

parthea:
Closing as per https://github.com/googleapis/google-cloud-python/pull/13840#issuecomment-2831253300

We will follow up on https://github.com/googleapis/google-cloud-python/issues/13839

PR #13963, opened 2025-05-31
fix! [google-ai-generativelanguage] prevent 0xa api_key suffix

pnikonowicz:
if a 0xa character appears at the end of
the api_key, the underlying grpc implementation
will fail with an invalid value header error.

see: https://gitlab.uni-hannover.de/tci-gateway-module/grpc/-/blob/v1.9.x/src/core/lib/surface/validate_metadata.cc#L84

to reproduce this, simply add a 0xa character to an api_key and try to perform any generative action via grpc transport.

### DIFF
--- a/packages/google-ai-generativelanguage/google/ai/generativelanguage_v1beta/services/generative_service/client.py
+++ b/packages/google-ai-generativelanguage/google/ai/generativelanguage_v1beta/services/generative_service/client.py
@@ -646,6 +646,10 @@ class GenerativeServiceClient(metaclass=GenerativeServiceClientMeta):
             raise ValueError(
                 "client_options.api_key and credentials are mutually exclusive"
             )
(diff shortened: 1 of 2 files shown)


parthea:
Thanks for opening a PR and reporting this issue, @pnikonowicz! As per https://github.com/googleapis/google-cloud-python/blob/main/CONTRIBUTING.rst#contributing, all client libraries are automatically generated using [gapic-generator-python](https://github.com/googleapis/gapic-generator-python). I'm going to apply the test and fix there, but please feel free to contribute to `gapic-generator-python` directly in the future.

ohmayr:
Added a `Do Not Merge` label since the fix needs to be applied in the generator. This PR can be closed if we have a separate tracking bug for this issue.

ohmayr:
Closing this PR since I've filed https://github.com/googleapis/google-cloud-python/issues/16223 to track this issue.

PR #14108, opened 2025-07-16
chore: add librarian state.yaml

parthea:
Fixes https://github.com/googleapis/librarian/issues/867

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -1,0 +1,16 @@
+image: gapic-generator-python:latest
+libraries:
+- apis:
+  - path: google/cloud/language/v1beta2
+    service_config: ''
+  - path: google/cloud/language/v2
+    service_config: ''
+  - path: google/cloud/language/v1
+    service_config: ''
+  id: google-cloud-language
+  last_generated_commit: 97a83d76a09a7f6dcab43675c87bdfeb5bcf1cb5
+  preserve_regex: ''
+  remove_regex: ''
+  sourcePaths:
+  - packages/google-cloud-language
+  version: 2.17.2
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -1,4 +1,4 @@
-image: gapic-generator-python:latest
+image: google-cloud-python-generator:latest
 libraries:
 - apis:
   - path: google/cloud/language/v1beta2
--- a/scripts/configure_state_yaml/packages_to_onboard.yaml
+++ b/scripts/configure_state_yaml/packages_to_onboard.yaml
@@ -1,3 +1,18 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


parthea:
~I'm still working on adding tests~

.librarian/state.yaml
```diff
@@ -0,0 +1,16 @@
+image: gapic-generator-python:latest
```
ohmayr: Is this supposed to be `google-cloud-python-generator`?

I understand that this may not make a difference for our local testing.

.librarian/state.yaml
```diff
@@ -0,0 +1,16 @@
...
+libraries:
+- apis:
```
ohmayr, line 3: Do we want to start with a library with a single api-path? This might be sufficient for the first milestone.

.librarian/state.yaml
```diff
@@ -0,0 +1,16 @@
...
+    service_config: ''
+  - path: google/cloud/language/v1
+    service_config: ''
+  id: google-cloud-language
+  last_generated_commit: 97a83d76a09a7f6dcab43675c87bdfeb5bcf1cb5
+  preserve_regex: ''
+  remove_regex: ''
+  sourcePaths:
+  - packages/google-cloud-language
+  version: 2.17.2
```
ohmayr, line 16: Can you file and link an issue here to remove this file once we're done with our local testing?

scripts/configure_state_yaml/packages_to_onboard.yaml
```diff
@@ -0,0 +1,3 @@
+packages_to_onboard: [
```
ohmayr, line 16: nit: missing copyright header

ohmayr (approved):

.librarian/state.yaml
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14108/commits/5f5038e16a502b69e51075711cf753f1396bc211

.librarian/state.yaml
parthea: Let's use multiple API paths since it's a common pattern

.librarian/state.yaml
parthea: We won't be deleting state.yaml. The plan is to have an initial state.yaml when we onboard

scripts/configure_state_yaml/packages_to_onboard.yaml
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14108/commits/5f5038e16a502b69e51075711cf753f1396bc211

.librarian/state.yaml
ohmayr: I think Librarian will be creating the `state.yaml` file and maintaining it once we have `configure` implemented. Ideally, state.yaml shouldn't have any manual edits.

We're only creating this file for the first milestone. Once we've got configure in place, Librarian should create and place the `state.yaml` file in the /.librarian folder.

Why do we want an initial handwritten state.yaml (apart from the first milestone)?

parthea: > I think Librarian will be creating the state.yaml file

Let's sync offline on this. My initial thought is that we will create the initial `state.yaml`, and update it as needed, and Librarian will fully take over `state.yaml` only once we've completed the migration from `owlbot->librarian`.

PR #14111, opened 2025-07-17
feat: implement a simple docker file to run cli commands

ohmayr:
This PR implements a minimalistic Dockerfile which runs the required CLI commands for Librarian.

Testing steps:

```

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -1,0 +1,50 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import sys
+
+def handle_configure(args):
+    # TODO(https://github.com/googleapis/librarian/issues/466): Implement configure command.
+    print("'configure' command executed.")
+
+def handle_generate(args):
+    # TODO(https://github.com/googleapis/librarian/issues/448): Implement generate command.
+    print("'generate' command executed.")
+
+def handle_build(args):
+    # TODO(https://github.com/googleapis/librarian/issues/450): Implement build command.
+    print("'build' command executed.")
+
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by parthea — APPROVED** [review]

File: .generator/cli.py [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+    parser = argparse.ArgumentParser(description="A simple CLI tool.")
+    subparsers = parser.add_subparsers(dest="command", required=True, help="Available commands")
+
+    # Define commands
+    for command_name, help_text in [
+        ("configure", "Onboard a new library or an api path to Librarian workflow."),
+        ("generate", "generate a python client for an API."),
+        ("build", "Run unit tests via nox for the generated library.")
+    ]:
+        handler_map = {"configure": handle_configure, "generate": handle_generate, "build": handle_build}
```
vchudnov-g, line 41: This line should be factored out of the loop.

.generator/test_cli.py
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cli import handle_generate, handle_build, handle_configure
+
+
+def test_handle_configure_dry_run():
```
vchudnov-g, line 18: Will you also have tests with `dry_run=False`? 

I could imagine that when `dry_run=False`, you could accumulate the commands that would be yielded by `dry_run=True` and pass them to a function `execute_run` that will execute them. In that case, the tests for `dry_run=False` need merely verify that the invocations all call `execute_run` with the expected parameters, and then we only need a single additional test for `execute_run` that it executes a given test command.

.github/workflows/generator.yml
```diff
@@ -0,0 +1,30 @@
...
+      uses: actions/checkout@v4
+      # Use a fetch-depth of 2
+      # See https://github.com/googleapis/google-cloud-python/issues/12013
+      # and https://github.com/actions/checkout#checkout-head.
+      with:
+        fetch-depth: 2
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.10"
```
vchudnov-g, line 24: We should probably use the same version of Python here and in the Dockerfile for consistency, so we test the CLI as it will execute in the container.

.generator/cli.py
ohmayr: addressed.

.github/workflows/generator.yml
ohmayr: addressed.

PR #14124, opened 2025-07-22
add bazelisk command

ohmayr:


.generator/test_cli.py
```diff
@@ -59,16 +59,30 @@ def test_handle_generate_dry_run():
     handle_generate(dry_run=True)
 
 
-def test_handle_generate_success(caplog, mock_generate_request_file):
+def test_handle_generate_success(caplog, mock_generate_request_file, mocker):
```
parthea: Do we also want to include a failed tested case, assuming that we will raise an exception in `handle_generate`?

.generator/cli.py
```diff
@@ -37,14 +38,40 @@ def handle_configure(dry_run=False):
...
+
+                    bazel_rule = result.stdout.strip()
+                    if not bazel_rule:
+                        logger.error("Bazel query returned an empty result.")
+                        sys.exit(1)
+
+                    logger.info(f"Found Bazel rule: {bazel_rule}")
+                except Exception as e:
+                    logger.error(f"Bazelisk query failed: {e}")
+                    sys.exit(1)
```
parthea: nit: refactor this into a helper function. Use exceptions instead of `logger.error`. We can then test the helper function with various inputs/scenarios

.generator/test_cli.py
ohmayr: addressed.

.generator/cli.py
ohmayr: addressed.

parthea (dismissed):

.generator/cli.py
```diff
@@ -47,6 +49,39 @@ def handle_configure():
...
+        result = subprocess.run(
+            command,
+            cwd=f"{SOURCE_DIR}/googleapis",
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+        bazel_rule = result.stdout.strip()
+        if not bazel_rule:
+            raise ValueError(f"Bazel query `{query}` returned an empty bazel rule.")
```
```suggestion
            raise ValueError(f"Bazelisk query `{query}` returned an empty bazel rule.")
```

PR #14145, opened 2025-07-24
feat: run post processor in cli

ohmayr:


.generator/cli.py
```diff
@@ -153,6 +162,26 @@ def _locate_and_extract_artifact(bazel_rule: str, library_id: str):
...
+    """Runs the synthtool post-processor on the output directory.
+
+    Raises:
+        ValueError: If the subprocess call fails.
+    """
+    try:
+        logger.info("Running Python post-processor...")
+        if SYNTHTOOL_INSTALLED:
+            command = ["python3", "-m", "synthtool.languages.python_mono_repo"]
+            os.chdir(OUTPUT_DIR)
```
parthea: Do we still need to change the current working directory via `os.chdir` if we set `cwd` in `subprocess.run`? If so, can you add a comment to explain why?

.generator/cli.py
```diff
@@ -153,6 +162,26 @@ def _locate_and_extract_artifact(bazel_rule: str, library_id: str):
...
+            subprocess.run(command, cwd=OUTPUT_DIR, text=True, check=True)
+        else:
+            raise SYNTHTOOL_IMPORT_ERROR
+        logger.info("Python post-processor ran successfully.")
+
+    except Exception as e:
```
parthea: Do we really need a try/except here, or can we just let the exception bubble up ? It seems like we want to coerce them all to `ValueError` which doesn't seem correct.

ohmayr: We don't need it. It just adds an extra line of information for us to keep a track of stack trace when an error bubbles up. It'll be easy to spot that the exception comes from within this function.

.generator/cli.py
ohmayr: We need it. Without `os.chdir`, synthtool fails.

.generator/cli.py
```diff
@@ -174,6 +184,21 @@ def _locate_and_extract_artifact(bazel_rule: str, library_id: str):
         ) from e
 
 
+def _run_post_processor():
+    """Runs the synthtool post-processor on the output directory.
+
+    Raises:
+        ValueError: If the subprocess call fails.
```
```suggestion
```

parthea (dismissed):

Issue #14142, opened 2025-07-23
Reduce the build time for .generator/Dockerfile


parthea:
In PR https://github.com/googleapis/google-cloud-python/pull/14129, a timeout of more than 1 hour is required in `cloudbuild.yaml` due to the number of python runtimes that are installed in the docker image as per the code snippet below.

https://github.com/googleapis/google-cloud-python/blob/598660eee066189150955d4151b5c94a54e60b6f/.generator/Dockerfile#L42

We should be able to significantly reduce the build time for `.generator/Dockerfile` by moving the installation of the python runtimes to a separate base image.

PR #14158, opened 2025-07-29
feat: Kaniko build

ohmayr:

Fixes https://github.com/googleapis/google-cloud-python/issues/14142 🦕


.generator/Dockerfile
```diff
@@ -114,11 +192,27 @@ COPY --from=builder /usr/local/lib/python3.12 /usr/local/lib/python3.12
...
 
+RUN git clone --depth 1 https://github.com/googleapis/synthtool.git /tmp/synthtool && \
+python3.9 -m pip install /tmp/synthtool && \
+rm -rf /tmp/synthtool
+RUN python3.9 -m pip install nox
+
 # Copy the CLI script into the container.
-COPY .generator/cli.py .
+COPY --chown=$UID:$GID .generator/cli.py .
+RUN chmod a+rx ./cli.py
```
parthea, line 211: Let's move this after the command to copy bazel cache to speed up the feedback loop while we're testing. The reason is that `cli.py` will change more frequently than the bazel cache in the short term.

.generator/Dockerfile
```diff
@@ -114,11 +192,27 @@ COPY --from=builder /usr/local/lib/python3.12 /usr/local/lib/python3.12
...
+
+# # Ensure the cache directories within the non-root user's context exist and are writable.
+# # This is crucial as Bazel creates subdirectories under BAZEL_HOME.
+
+COPY --from=builder --chown=$UID:$GID /bazel_cache /bazel_cache
+
+# RUN chown $UID:$GID -R /app /.cache /bazel_cache /source
```
parthea: Remove commented code

PR #14195, opened 2025-08-06
chore: add support for cli arguments

parthea:
This PR adds support for CLI arguments `librarian`, `input`, `output` and `source`

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -46,7 +46,7 @@ def _read_json_file(path: str) -> Dict:
     """Helper function that reads a json file path and returns the loaded json content.
 
     Args:
-        path (str): The file path to read.
+        path(str): The file path to read.
 
     Returns:
         dict: The parsed JSON content.
@@ -69,8 +69,8 @@ def _determine_bazel_rule(api_path: str, source: str) -> str:
     """Executes a `bazelisk query` to find a Bazel rule.
 
     Args:
-        api_path (str): The API path to query for.
-        source (str): The path to the root of the Bazel workspace.
+        api_path(str): The API path to query for.
+        source(str): The path to the root of the Bazel workspace.
 
     Returns:
         str: The discovered Bazel rule.
@@ -121,8 +121,8 @@ def _build_bazel_target(bazel_rule: str, source: str):
     """Executes `bazelisk build` on a given Bazel rule.
 
     Args:
-        bazel_rule (str): The Bazel rule to build.
-        source (str): The path to the root of the Bazel workspace.
+        bazel_rule(str): The Bazel rule to build.
+        source(str): The path to the root of the Bazel workspace.
 
     Raises:
         ValueError: If the subprocess call fails.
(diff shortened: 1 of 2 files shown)


.generator/cli.py
```diff
@@ -81,7 +84,7 @@ def _determine_bazel_rule(api_path: str) -> str:
         command = ["bazelisk", "query", query]
         result = subprocess.run(
             command,
-            cwd=f"{SOURCE_DIR}/googleapis",
```
ohmayr, line 84: I think source may not be the term here. I'm assuming that the path provided by librarian would be `/source/googleapis`? If so, we should also use it as the default value (if not configured) and probably rename it. Wdyt? If source dir can have more things in it than just `googleapis` we should call this `googleapis_dir` or `googleapis_root` instead.

.generator/cli.py
```diff
@@ -167,7 +180,8 @@ def _locate_and_extract_artifact(bazel_rule: str, library_id: str):
         logger.info(f"Found artifact at: {tarball_path}")
 
         # 3. Create a staging directory.
-        staging_dir = os.path.join(OUTPUT_DIR, "owl-bot-staging", library_id)
+        api_version=api_path.split("/")[-1]
```
ohmayr: Why is this change needed?

.generator/cli.py
```diff
@@ -196,7 +210,9 @@ def _run_post_processor():
     logger.info("Python post-processor ran successfully.")
 
 
-def handle_generate():
+def handle_generate(
+    librarian: str = LIBRARIAN_DIR, source: str = SOURCE_DIR, output: str = OUTPUT_DIR, input: str = INPUT_DIR
```
ohmayr: `SOURCE_DIR` does not seem to be right default value for `source`. I'd suggest verifying against the docker container to confirm that the default values work.

.generator/cli.py
parthea: > I think source may not be the term here.

`source` is the name of the argument provided by librarian. From https://github.com/googleapis/librarian/blob/main/doc/container-contract.md#generate-container-command, `This folder is mounted into the container. It contains, for example, the whole contents of googleapis. `

> I'm assuming that the path provided by librarian would be /source/googleapis?  If so, we should also use it as the default value (if not configured) and probably rename it.

This is an incorrect assumption. Based on the previous comment, the `source` directory contains the contents of `googleapis/googleapis`

> If source dir can have more things in it than just googleapis we should call this googleapis_dir or googleapis_root instead.

The source directory is the contents of `googleapis/googleapis`. The name of the argument is part of the librarian contract here: https://github.com/googleapis/librarian/blob/main/doc/container-contract.md#generate-container-command

.generator/cli.py
parthea: We need the `api_version` to mimic what we had before with the owlbot design. The generated code would reside in `owl-bot-staging/google-cloud-language/<version>` as shown below

https://github.com/googleapis/google-cloud-python/blob/8450560be9916aaf9b57bd7194848625ab36f3c4/packages/google-cloud-language/.OwlBot.yaml#L19-L22

We would have the following directories

`owl-bot-staging/google-cloud-language/v1`
`owl-bot-staging/google-cloud-language/v2`

The post processor loops through each version using the code here https://github.com/googleapis/synthtool/blob/75516372fb8854e201a985bd92160b39af9b7dbe/synthtool/languages/python_mono_repo.py#L265

.generator/cli.py
parthea: I don't expect librarian to ever use the default. Librarian will always provide these arguments. The defaults are intended to facilitate testing. We can remove the defaults if it's confusing but we would need to backfill the existing tests which are currently relying on the default values.

PR #14206, opened 2025-08-08
chore: polish post processing step used for librarian client library generation

parthea:
This PR adds 2 new functions `_copy_files_needed_for_post_processing` and `_clean_up_files_after_post_processing`. 

`_copy_files_needed_for_post_processing` - This function is used to copy files from the `input` directory .  Specifically, `.repo-metadata.json` and `scripts/client-post-processing` are needed as part of client library generation.

`_clean_up_files_after_post_processing` - This function removes files which should not be part of the generated code because it would create an unnecessary diff. The files omitted are: 

- `CHANGELOG.md`
- `docs/CHANGELOG.md`
- `docs/README.rst`
- `scripts/client-post-processing/*.yaml`
- `**/gapic_version.py`
- `samples/generated_samples/snippet_metadata*.json`

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -25,7 +25,6 @@ from typing import Dict, List
 
 try:
     import synthtool
-    from synthtool import gcp
     from synthtool.languages import python_mono_repo
 
     SYNTHTOOL_INSTALLED = True
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -138,7 +138,14 @@ def _build_bazel_target(bazel_rule: str, source: str):
     """
     logger.info(f"Executing build for rule: {bazel_rule}")
     try:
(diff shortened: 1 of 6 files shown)


.generator/cli.py
```diff
@@ -13,18 +13,20 @@
...
 import re
+import shutil
 import subprocess
 import sys
-import subprocess
 from typing import Dict, List
 
 try:
     import synthtool
     from synthtool import gcp
```
ohmayr: can we remove this line if it's not needed?

.generator/cli.py
```diff
@@ -136,7 +138,14 @@ def _build_bazel_target(bazel_rule: str, source: str):
     """
     logger.info(f"Executing build for rule: {bazel_rule}")
     try:
-        command = ["bazelisk", "build", bazel_rule]
+        command = [
```
ohmayr, line 145: Can we add a comment explaining the passed in bazel_cache? I.e. we're using the prewarmed bazel cache from the docker image to speed up the bazelisk commands.

ohmayr (dismissed):

ohmayr (approved):

PR #14327, opened 2025-08-30
chore: add support for librarian release init

parthea:
This PR adds support for the librarian language specific `release-init` commit following https://github.com/googleapis/librarian/blob/main/doc/language-onboarding.md

Towards https://github.com/googleapis/librarian/issues/886

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -434,27 +434,6 @@ def handle_build(librarian: str = LIBRARIAN_DIR, repo: str = REPO_DIR):
     logger.info("'build' command executed.")
 
 
-def _read_and_process_file(input_path: str, output_path: str, process_func) -> None:
-    """Helper function to read, process, and write a file.
-
-    Args:
-        input_path (str): The path to the file to read.
-        output_path (str): The path to the file to write.
-        process_func (callable): A function that takes the file content as a string
-                                  and returns the modified string.
-    """
-    os.makedirs(Path(output_path).parent, exist_ok=True)
-    shutil.copy(input_path, output_path)
-
-    with open(output_path, "r", encoding="utf-8") as f:
-        content = f.read()
-
-    updated_content = process_func(content)
-
-    with open(output_path, "w", encoding="utf-8") as f:
-        f.write(updated_content)
-
-
 def _get_libraries_to_prepare_for_release(library_entries: Dict) -> List[dict]:
     """Note the request data may have multiple libraries in it. Locate all
     libraries which should be prepared for release. This can be done by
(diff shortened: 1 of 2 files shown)


.generator/cli.py
```diff
@@ -433,23 +440,256 @@ def handle_build(librarian: str = LIBRARIAN_DIR, repo: str = REPO_DIR):
     logger.info("'build' command executed.")
 
 
-if __name__ == "__main__":  # pragma: NO COVER
+def _read_and_process_file(input_path: str, output_path: str, process_func) -> None:
```
ohmayr: We already have a `_read_json_file` function. Can we simply just re-use that and have this function simply update a file. I'm thinking more like a `_write_json_file` or `_update_json_file`.

.generator/cli.py
```diff
@@ -433,23 +440,256 @@ def handle_build(librarian: str = LIBRARIAN_DIR, repo: str = REPO_DIR):
...
+        for library in state_yaml.get("libraries", []):
+            if library.get("id") == package_name:
+                return library.get("version")
+
+    raise ValueError(
+        f"Could not determine previous version for {package_name} from state.yaml"
+    )
+
+
+def _update_global_changelog(source_path: str, output_path: str, libraries: List[dict]):
```
ohmayr: Can we move these helper functions above in the file or at least before the function they're used in?

.generator/cli.py
parthea: I removed the `_read_and_process_file` logic for now. I'll use add `_write_json_file` in another PR

.generator/cli.py
parthea: I removed `_update_global_changelog` for now. I'll add it back in another PR and move it as suggested.

.generator/cli.py
```diff
@@ -433,23 +434,91 @@ def handle_build(librarian: str = LIBRARIAN_DIR, repo: str = REPO_DIR):
...
         "generate": handle_generate,
         "build": handle_build,
+        "release-init": handle_release_init,
     }
 
     for command_name, help_text in [
         ("configure", "Onboard a new library or an api path to Librarian workflow."),
         ("generate", "generate a python client for an API."),
         ("build", "Run unit tests via nox for the generated library."),
+        ("release-init", "Prepare to release a specific library"),
```
ohmayr: is it just a specific library? Above in the file, we filter all the libraries to be released.

ohmayr (dismissed):

.generator/cli.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14327/commits/580089ca932f8b987fdc5b8588bbcc50d73cc639

PR #14350, opened 2025-09-03
chore: add ability to update library specific version files in releas…

parthea:
This PR builds on top of #14349 and adds the ability to update the client library version in client library version files using the `librarian release init` command such as 

- [gapic_version.py](https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-language/google/cloud/language_v1/gapic_version.py)
- [snippet_metadata*.json](https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-language/samples/generated_samples/snippet_metadata_google.cloud.language.v2.json)

Towards https://github.com/googleapis/librarian/issues/886

PR #14353, opened 2025-09-03
chore: add ability to update library specific changelog in release-init

parthea:
This PR builds on top of #14350 and adds the ability to update the client library changelog using the `librarian release init` command

Towards https://github.com/googleapis/librarian/issues/886

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -649,15 +649,13 @@ def _process_changelog(
     for change_type, changes in grouped_changes:
         # We only care about feat, fix, docs
         adjusted_change_type = change_type.replace("!", "")
+        change_type_map = {
+            "feat": "Features",
+            "fix": "Bug Fixes",
+            "docs": "Documentation",
+        }
         if adjusted_change_type in ["feat", "fix", "docs"]:
-            if adjusted_change_type == "feat":
-                adjusted_change_type = "Features"
-            elif adjusted_change_type == "fix":
-                adjusted_change_type = "Bug Fixes"
-            else:
-                adjusted_change_type = "Documentation"
-
-            entry_parts.append(f"\n\n### {adjusted_change_type}\n")
+            entry_parts.append(f"\n\n### {change_type_map[adjusted_change_type]}\n")
             for change in changes:
                 commit_link = f"([{change['source_commit_hash']}]({repo_url}/commit/{change['source_commit_hash']}))"
                 entry_parts.append(f"* {change['subject']} {commit_link}")


.generator/cli.py
```diff
@@ -578,6 +582,135 @@ def _update_version_for_library(
...
+    for change_type, changes in grouped_changes:
+        # We only care about feat, fix, docs
+        adjusted_change_type = change_type.replace("!", "")
+        if adjusted_change_type in ["feat", "fix", "docs"]:
+            if adjusted_change_type == "feat":
+                adjusted_change_type = "Features"
+            elif adjusted_change_type == "fix":
+                adjusted_change_type = "Bug Fixes"
+            else:
+                adjusted_change_type = "Documentation"
```
ohmayr: How about we use a map for these? Something like:
```python3
change_type = {"feat": "Features", "fix": "Bug Fixes", "docs": "Documentation"}
```

parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14353/commits/4973613d71c30d0b8498bec45f1485edf179b3c5

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+            the `state.yaml` file.
+
+    Returns:
+        str: The version for a given library in state.yaml
+    """
+    state_yaml_path = f"{librarian}/{STATE_YAML_FILE}"
+
+    with open(state_yaml_path, "r") as state_yaml_file:
+        state_yaml = yaml.safe_load(state_yaml_file)
+        for library in state_yaml.get("libraries", []):
```
ohmayr, line 600: Not a blocker but I'm just wondering if we need to read the `state.yaml` elsewhere such as the configure command? If so, maybe it'll be better to just load the content once and pass it around to reduce the I/O operations.

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+
+    * Update import statement example in README ([868b006](https://github.com/googleapis/google-cloud-python/commit/868b0069baf1a4bf6705986e0b6885419b35cdcc))
+
+    Args:
+        content(str): The contents of an existing changelog.
+        library_changes(List[Dict]): List of dictionaries containing the changes
+            for a given library.
+        version(str): The new version of the library.
+        previous_version: The previous version of the library.
+        package_name(str): The name of the package where the changelog should
```
ohmayr, line 628: Question: What's the difference between library_id and package_name?

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+
+    Returns: A string with the modified content.
+    """
+    repo_url = "https://github.com/googleapis/google-cloud-python"
+    current_date = datetime.now().strftime("%Y-%m-%d")
+
+    # Create the main version header
+    version_header = (
+        f"## [{version}]({repo_url}/compare/{package_name}-v{previous_version}"
+        f"...{package_name}-v{version}) ({current_date})"
```
ohmayr, line 641: This can probably be in a separate helper function.

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+
+    for change_type, changes in grouped_changes:
+        # We only care about feat, fix, docs
+        adjusted_change_type = change_type.replace("!", "")
+        change_type_map = {
+            "feat": "Features",
+            "fix": "Bug Fixes",
+            "docs": "Documentation",
+        }
+        if adjusted_change_type in ["feat", "fix", "docs"]:
```
ohmayr, line 657:
```suggestion
        if adjusted_change_type in change_type_map:
```

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+
+    # Create the main version header
+    version_header = (
+        f"## [{version}]({repo_url}/compare/{package_name}-v{previous_version}"
+        f"...{package_name}-v{version}) ({current_date})"
+    )
+    entry_parts = [version_header]
+
+    # Group changes by type (e.g., feat, fix, docs)
+    library_changes.sort(key=lambda x: x["type"])
```
ohmayr, line 646: Here and elsewhere, let's define these strings as parameters and reuse them to avoid errors.

Example:
```python3
type_key = "type"
source_commit_hash_key = "source_commit_hash"
subject_key = "subject"
```

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+    grouped_changes = itertools.groupby(library_changes, key=lambda x: x["type"])
+
+    for change_type, changes in grouped_changes:
+        # We only care about feat, fix, docs
+        adjusted_change_type = change_type.replace("!", "")
+        change_type_map = {
+            "feat": "Features",
+            "fix": "Bug Fixes",
+            "docs": "Documentation",
+        }
```
ohmayr, line 656: Let's define this map outside this for loop.

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+        if adjusted_change_type in ["feat", "fix", "docs"]:
+            entry_parts.append(f"\n\n### {change_type_map[adjusted_change_type]}\n")
+            for change in changes:
```
ohmayr, line 659: let's reword one of these two to avoid misspelling errors.

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+        library_changes(List[Dict]): List of dictionaries containing the changes
+            for a given library
+        version(str): The desired version
+        previous_version(str): The version in state.yaml for a given package
+        package_name(str): The name of the package where the changelog should
+            be updated.
+    """
+
+    source_path = f"{repo}/packages/{package_name}/CHANGELOG.md"
+    output_path = f"{output}/packages/{package_name}/CHANGELOG.md"
```
ohmayr, line 701:
```suggestion
    relative_path = f"packages/{package_name}/CHANGELOG.md"
    changelog_src = f"{repo}/{relative_path}"
    changelog_dest = f"{output}/relative_path}"
```

.generator/cli.py
```diff
@@ -618,13 +749,21 @@ def handle_release_init(
...
             path_to_library = f"packages/{package_name}"
             _update_version_for_library(repo, output, path_to_library, version)
 
-            # Update library specific version files.
-            # TODO(https://github.com/googleapis/google-cloud-python/pull/14353):
-            # Conditionally update the library specific CHANGELOG if there is a change.
-            pass
+            # Get previous version from state.yaml
+            previous_version = _get_previous_version(package_name, librarian)
+            if previous_version != version:
```
ohmayr, line 758: What happens if the `previous_version` is the same as `version `and the library is a part of library_release_data? Do we log an error?

.generator/cli.py
```diff
@@ -578,6 +582,133 @@ def _update_version_for_library(
...
+                commit_link = f"([{change['source_commit_hash']}]({repo_url}/commit/{change['source_commit_hash']}))"
+                entry_parts.append(f"* {change['subject']} {commit_link}")
+
+    new_entry_text = "\n".join(entry_parts)
+    anchor_pattern = re.compile(
+        rf"(\[1\]: https://pypi\.org/project/{package_name}/#history)",
+        re.MULTILINE,
+    )
+    replacement_text = f"\\g<1>\n\n{new_entry_text}"
+    updated_content, num_subs = anchor_pattern.subn(replacement_text, content, count=1)
```
ohmayr, line 669: I think we can refactor these text replacements into their own functions. This will make unit testing easier and cleaner where we can test each of these separately and will probably clean up this function as well. This is not a blocker so feel free to skip if you disagree.

parthea:
I will address the feedback in a follow-up PR

.generator/cli.py
parthea: Let's revisit this when the configure command is implemented. It's not clear if there will be a requirement to read state.yaml directly. This requirement was unique to releasing because I wanted to read the last released version. `release-init-request.json` did not contain this information so I had to read it directly from `state.yaml`

.generator/cli.py
parthea: There is no difference. I've opened https://github.com/googleapis/google-cloud-python/pull/14378 to replace `package_name` with `library_id`

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14379

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14380

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14380

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14380

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14380

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14380

.generator/cli.py
parthea: This will be fixed in https://github.com/googleapis/google-cloud-python/pull/14381

.generator/cli.py
parthea: I agree that `_process_changelog` has a lot of replacements/logic. I'm going to follow up on this separately in https://github.com/googleapis/google-cloud-python/issues/14382

PR #14381, opened 2025-09-09
chore(librarian): RaiseValue error if current and previous version match

parthea:
This PR addresses the feedback from https://github.com/googleapis/google-cloud-python/pull/14353#discussion_r2331389309

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -729,6 +729,12 @@ def handle_release_init(
             the config.yaml.
         output(str): Path to the directory in the container where modified
             code should be placed.
+
+    Raises: 
+        ValueError: if the version in `release-init-request.json` is
+            the same as the version in state.yaml
+        ValueError: if the `release-init-request.json` file in the given
+            librarian directory cannot be read.
     """
 
     try:
@@ -751,19 +757,25 @@ def handle_release_init(
             package_name = library_release_data["id"]
             library_changes = library_release_data["changes"]
             path_to_library = f"packages/{package_name}"
-            _update_version_for_library(repo, output, path_to_library, version)
 
             # Get previous version from state.yaml
             previous_version = _get_previous_version(package_name, librarian)
-            if previous_version != version:
-                _update_changelog_for_library(
-                    repo,
-                    output,
-                    library_changes,
-                    version,
-                    previous_version,
-                    package_name,
+            if previous_version == version:
+                raise ValueError(
+                    "The version in `release-init-request.json` is the same as the version in state.yaml\n"
+                    f"{package_name} previous released version: {previous_version}\n"
+                    f"{package_name} current version: {version}\n"
                 )
+
+            _update_version_for_library(repo, output, path_to_library, version)
+            _update_changelog_for_library(
+                repo,
+                output,
+                library_changes,
+                version,
+                previous_version,
+                package_name,
+            )
 
     except Exception as e:
         raise ValueError(f"Release init failed: {e}") from e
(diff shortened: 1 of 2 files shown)


.generator/cli.py
```diff
@@ -751,20 +757,26 @@ def handle_release_init(
...
-                _update_changelog_for_library(
-                    repo,
-                    output,
-                    library_changes,
-                    version,
-                    previous_version,
-                    package_name,
+            if previous_version == version:
+                raise ValueError(
+                    "The version in `release-init-request.json` is the same as the version in state.yaml\n"
```
ohmayr:
```suggestion
                    f"The version in `{RELEASE_INIT_REQUEST_FILE}` is the same as the version in `{STATE_YAML_FILE}`\n"
```

ohmayr (dismissed):

.generator/cli.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14381/commits/4874fbb53373bef962a51cf7f440ecc5a32448f9

.generator/cli.py
```diff
@@ -842,6 +842,12 @@ def handle_release_init(
             the config.yaml.
         output(str): Path to the directory in the container where modified
             code should be placed.
+
+    Raises:
+        ValueError: if the version in `release-init-request.json` is
+            the same as the version in state.yaml
+        ValueError: if the `release-init-request.json` file in the given
```
ohmayr, line 849: Should we combine these two statements?

.generator/cli.py
```diff
@@ -864,20 +870,26 @@ def handle_release_init(
...
-                    repo,
-                    output,
-                    library_changes,
-                    version,
-                    previous_version,
-                    library_id,
+            if previous_version == version:
+                raise ValueError(
+                    f"The version in {RELEASE_INIT_REQUEST_FILE} is the same as the version in {STATE_YAML_FILE}\n"
+                    f"{library_id} previous released version: {previous_version}\n"
```
ohmayr, line 879: Given that the version is the same, it probably makes sense to output a single entry in the ValueError.

.generator/test_cli.py
```diff
@@ -604,14 +604,51 @@ def test_handle_release_init_success(mocker, mock_release_init_request_file):
...
+        mocker.patch(
+            "cli._get_libraries_to_prepare_for_release",
+            return_value=mock_release_init_request_content["libraries"],
+        )
+        mocker.patch("cli._get_previous_version", return_value="1.2.2")
+        mocker.patch("cli._process_changelog", return_value=None)
+        mocker.patch(
+            "cli._read_json_file", return_value=mock_release_init_request_content
+        )
+        with pytest.raises(
```
ohmayr, line 646: The ValueError is probably only raised for `googel-cloud-language`. Given that to be the case, why do we need an entry for `google-cloud-another-library` in the mock data?

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14425

.generator/cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14425

.generator/test_cli.py
parthea: This will be addressed in https://github.com/googleapis/google-cloud-python/pull/14425

PR #14386, opened 2025-09-09
chore(librarian): add ability to parse BUILD.bazel

parthea:
Towards https://github.com/googleapis/librarian/issues/1929


.generator/parse_googleapis_content.py
```diff
@@ -0,0 +1,82 @@
+import starlark as sl
```
ohmayr, line 15: Missing the copyright header here and in other files.

.generator/parse_googleapis_content.py
```diff
@@ -0,0 +1,82 @@
...
+            mod.add_callable("csharp_gapic_library", bazel_target)
+            mod.add_callable("csharp_grpc_library", bazel_target)
+            mod.add_callable("csharp_proto_library", bazel_target)
+            mod.add_callable("cc_grpc_library", bazel_target)
+            mod.add_callable("cc_gapic_library", bazel_target)
+            mod.add_callable("cc_proto_library", bazel_target)
+            mod.add_callable("moved_proto_library", bazel_target)
+            mod.add_callable("proto_library", bazel_target)
+            mod.add_callable("upb_c_proto_library", bazel_target)
+        return mod.freeze()
```
ohmayr: How about something like the following:

```suggestion
        _PY_CALLABLES = (
            "py_gapic_assembly_pkg", "py_gapic_library", "py_test",
            "py_proto_library", "py_grpc_library", "py_import",
        )
        
        _JAVA_CALLABLES = (
            "java_gapic_assembly_gradle_pkg", "java_gapic_library",
            "java_gapic_test", "java_grpc_library", "java_proto_library",
        )
        
        _GO_CALLABLES = (
            "go_gapic_assembly_pkg", "go_gapic_library", "go_proto_library",
            "go_grpc_library",
        )
        
        _PHP_CALLABLES = (
            "php_gapic_assembly_pkg", "php_gapic_library", "php_grpc_library",
            "php_proto_library",
        )
        
        _NODEJS_CALLABLES = ("nodejs_gapic_assembly_pkg", "nodejs_gapic_library")
        
        _RUBY_CALLABLES = (
            "ruby_ads_gapic_library", "ruby_cloud_gapic_library",
            "ruby_gapic_assembly_pkg", "ruby_grpc_library", "ruby_proto_library",
        )
        
        _CSHARP_CALLABLES = (
            "csharp_gapic_assembly_pkg", "csharp_gapic_library",
            "csharp_grpc_library", "csharp_proto_library",
        )
        
        _CC_CALLABLES = ("cc_grpc_library", "cc_gapic_library", "cc_proto_library")
        
        _MISC_CALLABLES = (
            "moved_proto_library", "proto_library", "proto_library_with_info",
            "upb_c_proto_library",
        )
        
        _CALLABLE_MAP = {
    "@rules_proto//proto:defs.bzl": ("proto_library",),
    "@com_google_googleapis_imports//:imports.bzl": (
        _PY_CALLABLES + _JAVA_CALLABLES + _GO_CALLABLES + _PHP_CALLABLES +
        _NODEJS_CALLABLES + _RUBY_CALLABLES + _CSHARP_CALLABLES +
        _CC_CALLABLES + _MISC_CALLABLES
    )
    }
    
    def load(name):
        mod = sl.Module()
        for callable_name in _CALLABLE_MAP.get(name, []):
            mod.add_callable(callable_name, bazel_target)
        return mod.freeze()
```

.generator/parse_googleapis_content.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14386/commits/31dbb8e2667d8f450c078240d0fc6b6b3bc58bdd

.generator/test-resources/librarian/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
+# This file was automatically generated by BuildFileGenerator
```
ohmayr: Why do we need to copy the language build file here? Given that the librarian already has access to `googleapis`, can we not read the file from there?

.generator/parse_googleapis_content.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14386/commits/b59962a622d2985218beb7727a143da52fa567a0

.generator/test-resources/librarian/BUILD.bazel
parthea: This is for unit testing purposes only to test that `parse_googleapis_content` works as expected. It's part of `test-resources`. We don't currently clone `googleapis` during unit testing. This file is not used by librarian or the librarian flow at all.

.generator/test_parse_googleapis_content.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+import parse_googleapis_content
+from pathlib import Path
+
+GENERATOR_DIR = Path(__file__).resolve().parent
+BUILD_BAZEL_PATH = f"{GENERATOR_DIR}/test-resources/librarian/BUILD.bazel"
+
+
+def test_parse_build_bazel():
```
ohmayr, line 23: I don't think it's necessary to test the behaviour of the underlying function. If we can move out the callables that we define outside of the function, we can simplify this unit test to only verify the logic that we're implementing. This will also reduce the number of lines here.

 Let me know what you think.

.generator/test-resources/librarian/BUILD.bazel
ohmayr: We can probably use a much simpler `BUILD.bazel` file for unit testing.

.generator/parse_googleapis_content.py
```diff
@@ -0,0 +1,130 @@
...
+    def noop_bazel_rule(**args):
+        pass
+
+    mod.add_callable("package", noop_bazel_rule)
+    mod.add_callable("alias", noop_bazel_rule)
+    mod.add_callable("py_test", noop_bazel_rule)
+    mod.add_callable("sh_binary", noop_bazel_rule)
+    mod.add_callable("proto_library", noop_bazel_rule)
+    mod.add_callable("java_proto_library", noop_bazel_rule)
+    mod.add_callable("genrule", noop_bazel_rule)
```
ohmayr: Similary, we can define these strings as `_NOOP_CALLABLES` or whatever wording you think makes sense here and use a for loop.

.generator/test-resources/librarian/BUILD.bazel
parthea: I found that it was easier to copy a real one from `googleapis` rather than try to figure out which lines are absolutely necessary to get full code coverage.

.generator/parse_googleapis_content.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def load(name):
+        mod = sl.Module()
+        for callable_name in _CALLABLE_MAP.get(name, []):
+            mod.add_callable(callable_name, bazel_target)
```
ohmayr, line 128: What happens if a bazel rule / target is missing for one of the client languages in the `BUILD.bazel` file?

.generator/test-resources/librarian/BUILD.bazel
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14386/commits/9e100a57e7c9894a4ca2a4d25022cbc92a02506f

.generator/test_parse_googleapis_content.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14386/commits/9e100a57e7c9894a4ca2a4d25022cbc92a02506f

.generator/parse_googleapis_content.py
ohmayr: To be clear, what I mean is that can we not read the rules defined within the `BUILD.bazel` before we parse them?

.generator/parse_googleapis_content.py
parthea: fixed in https://github.com/googleapis/google-cloud-python/pull/14386/commits/86c72b9f5d4a16cd0c8f7bfaf73f14932e6bc868

.generator/parse_googleapis_content.py
parthea: If a rule is missing in `BUILD.bazel`, there won't be an issue. There will be an issue if a rule is missing in `parse_googleapis_content.py`. We would see something like `starlark.StarlarkError: error: Module has no symbol ...`

As an example, if I remove `py_gapic_library` from `parse_googleapis_content.py`, I see the following stack trace.

```
>       sl.eval(mod, ast, glb, sl.FileLoader(load))
E       starlark.StarlarkError: error: Module has no symbol `py_gapic_library`, did you mean `py_grpc_library`?
E         --> BUILD.bazel:19:5
E          |
E       19 |     "py_gapic_library",
E          |     ^^^^^^^^^^^^^^^^^^
E          |

.generator/parse_googleapis_content.py:131: StarlarkError
```

ohmayr: So we're guaranteed to have covered all the rules in here? Won't it still be better to just read the rules from the bazel file instead of hardcoding them here?

ohmayr (approved):

.generator/parse_googleapis_content.py
parthea: I'll file a tracking bug for this

parthea: Filed https://github.com/googleapis/google-cloud-python/issues/14389

PR #14402, opened 2025-09-10
chore: run system test on packages that have change

Linchin:
replaces #14400

### DIFF
--- a/.kokoro/system.sh
+++ b/.kokoro/system.sh
@@ -42,17 +42,31 @@ system_test_script="${PROJECT_ROOT}/.kokoro/system-single.sh"
 for dir in `find 'packages' -type d -wholename 'packages/*/tests/system'`; do
   # Get the path to the package by removing the suffix /tests/system
   package=$(echo $dir | cut -f -2 -d '/')
-  echo "Running system tests for ${package}"
-  pushd ${package}
-  # Temporarily allow failure.
+  should_test=false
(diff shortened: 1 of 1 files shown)


.kokoro/system.sh
```diff
@@ -42,17 +42,31 @@ system_test_script="${PROJECT_ROOT}/.kokoro/system-single.sh"
...
   package=$(echo $dir | cut -f -2 -d '/')
-  echo "Running system tests for ${package}"
-  pushd ${package}
-  # Temporarily allow failure.
+  should_test=false
+  echo "checking changes with 'git diff HEAD~.. ${package}/CHANGELOG.md'"
   set +e
-  ${system_test_script}
-  ret=$?
+  changelog_modified=$(git diff "HEAD~.." ${package}/gapic_version.py | wc -l)
```
parthea, line 48: For certain packages where we want to run tests on any change, we could have

```suggestion
if [[ $package = @(*google-cloud-bigquery-storage*) ]]; then
    # run tests on any change
    package_modified=$(git diff "HEAD~.." ${package} | wc -l)
else
    # run tests when there are changes to CHANGELOG.md
    package_modified=$(git diff "HEAD~.." ${package}/CHANGELOG.md | wc -l)
fi
```

parthea:
@ohmayr will be working on this as it's also needed for the librarian work

PR #14449, opened 2025-09-16
chore: run system tests only for modified packages

parthea:
Currently the system test runs for all packages. This PR adds logic to check if there is a change to the package level `CHANGELOG.md` so that we only runs system tests for packages that have changed. We are only interested in `packages/<package>/CHANGELOG.md` changing as we only want to run system tests on release PRs.

Fixes https://github.com/googleapis/librarian/issues/2067

### DIFF
--- a/.kokoro/system.sh
+++ b/.kokoro/system.sh
@@ -38,21 +38,40 @@ pwd
 # A file for running system tests
 system_test_script="${PROJECT_ROOT}/.kokoro/system-single.sh"
 
+# This is needed in order for `git diff` to succeed
+git config --global --add safe.directory $(realpath .)
+
 # Run system tests for each package with directory packages/*/tests/system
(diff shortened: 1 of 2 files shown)


.kokoro/system.sh
```diff
@@ -38,21 +38,40 @@ pwd
...
+      # Temporarily allow failure.
+      set +e
+      ${system_test_script}
+      ret=$?
+      set -e
+      if [ ${ret} -ne 0 ]; then
+          RETVAL=${ret}
+      fi
+      popd
+  fi
```
ohmayr: Why not just clean up the `should_test` variable and move this block of code in the else block above?

parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14449/commits/5ff87d33cff6887263876803a0d764738bf1bec6

PR #14452, opened 2025-09-17
chore(librarian): update configure_state_yaml.py to avoid duplicate entries

parthea:
See output below for `google-cloud-dlp` and `google-cloud-eventarc` which already exist in `.librarian/state.yaml`

```
partheniou@partheniou-vm-3:~/git/google-cloud-python/scripts/configure_state_yaml$ python3 configure_state_yaml.py 
Skipping package 'google-cloud-dlp' as it already exists in state.yaml.
Skipping package 'google-cloud-eventarc' as it already exists in state.yaml.
```


jskeet (approved):
Yup, this is what we do in the Go `stategen` tool.

PR #14461, opened 2025-09-18
chore(migration): Migrate code from googleapis/python-bigquery-pandas into packages/pandas-gbq

parthea:
See #10988. 

## Before Merging
- [ ] **Be sure to NOT Squash and Merge**

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -45,6 +45,7 @@ logger = logging.getLogger()
 
 BUILD_REQUEST_FILE = "build-request.json"
 GENERATE_REQUEST_FILE = "generate-request.json"
+CONFIGURE_REQUEST_FILE = "configure-request.json"
 RELEASE_INIT_REQUEST_FILE = "release-init-request.json"
 STATE_YAML_FILE = "state.yaml"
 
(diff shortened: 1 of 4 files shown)


packages/pandas-gbq/noxfile.py
```diff
@@ -398,12 +400,111 @@ def prerelease_deps(session):
...
+    session.run("python", "-m", "pip", "freeze")
+
+    # Run all tests, except a few samples tests which require extra dependencies.
     session.run(
         "py.test",
         "--quiet",
         "-W default::PendingDeprecationWarning",
-        f"--junitxml=prerelease_system_{session.python}_sponge_log.xml",
-        os.path.join("tests", "system"),
+        f"--junitxml=prerelease_unit_{session.python}_sponge_log.xml",
```
Linchin: Why are we changing the path of the log from system to unit? It seems to run the system test

parthea: This specific nox session `core_deps_from_source` is new. It only runs unit tests. I'll update the comment

Linchin: I see, why are we changing this to unit test?

parthea: `core_deps_from_source` is a brand new nox session so there should not be any diff. I fixed the name of the log file so it should be clear now that this log is specific to `core_deps_from_source` not `prerelease`

Linchin:
I think we also need to add samples test (and samples lint), similar to the split repo:
image

They are only unit tests, so we just need to add github actions.

packages/pandas-gbq/README.rst
Linchin, line 1: Why are we removing so much content from README.rst? If we are cleaning it up, perhaps we can do it in a separate PR and keep this PR solely for migration purpose.

parthea: There was a large diff but that was reverted. Please take a look at the current diff.

Linchin:
This PR is temporarily on hold as we focus on more prioritized projects.

PR #14463, opened 2025-09-18
chore(librarian): onboard google-ads-admanager

parthea:
Onboard `google-ads-admanager` to librarian

Towards https://github.com/googleapis/librarian/issues/761

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -154,3 +154,24 @@ libraries:
     remove_regex:
       - packages/google-analytics-data
     tag_format: '{id}-v{version}'
+  - id: google-ads-admanager
+    version: 0.3.0
+    last_generated_commit: d300b151a973ce0425ae4ad07b3de957ca31bec6
+    apis:
+      - path: google/ads/admanager/v1
+    source_roots:
+      - packages/google-ads-admanager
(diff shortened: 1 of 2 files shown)


.librarian/state.yaml
```diff
@@ -154,3 +154,24 @@ libraries:
...
       - packages/google-analytics-data
     tag_format: '{id}-v{version}'
+  - id: google-ads-admanager
+    version: 0.3.0
+    last_generated_commit: d300b151a973ce0425ae4ad07b3de957ca31bec6
+    apis:
+      - path: google/ads/admanager/v1
+    source_roots:
+      - packages/google-ads-admanager
+    preserve_regex:
```
jskeet, line 164: In Go, I've made all our regexes anchored to the source directories to avoid any oddities - see https://github.com/googleapis/google-cloud-go/blob/main/.librarian/state.yaml for example.

No need to change this now, but we might want to align on this (and make sure the behavior is really carefully documented). For Go, this fixed issues where there are modules of both "ai" and "aiplatform".

parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14464

PR #14464, opened 2025-09-18
chore(librarian): anchor regex in state.yaml

parthea:
This PR addresses the feedback from https://github.com/googleapis/google-cloud-python/pull/14463#discussion_r2358316928 to anchor the regex values in `.librarian/state.yaml`


.librarian/state.yaml
```diff
@@ -9,18 +9,18 @@ libraries:
...
-      - docs/CHANGELOG.md
-      - docs/README.rst
-      - samples/README.txt
-      - tar.gz
-      - gapic_version.py
-      - samples/generated_samples/snippet_metadata_
-      - scripts/client-post-processing
-      - samples/snippets/README.rst
-      - tests/system
+      - ^packages/google-cloud-dlp/CHANGELOG.md
```
jskeet, line 12: I'd anchor the end as well as the start, just in case there are leading substrings. (And make it clear where that's intentional.)

It feels possibly over the top, but does avoid any ambiguity.

.librarian/state.yaml
```diff
@@ -9,18 +9,18 @@ libraries:
...
+      - ^packages/google-cloud-dlp/docs/CHANGELOG.md
+      - ^packages/google-cloud-dlp/docs/README.rst
+      - ^packages/google-cloud-dlp/samples/README.txt
+      - ^packages/google-cloud-dlp/tar.gz
```
jskeet: Is this right? Wouldn't it be dlp.tar.gz or similar? (It's not clear to me what purpose these files serve, mind you.)

.librarian/state.yaml
```diff
@@ -9,18 +9,18 @@ libraries:
...
+      - ^packages/google-cloud-dlp/.*/gapic_version.py
```
jskeet, line 16: I believe this would allow `packages/google-cloud-dlp/x/y/z/gapic_version.py` (i.e. it allows multiple levels). If that's deliberate, great - if you only want to allow a single level, then `^packages/google-cloud-dlp/[^/]*/gapic_version.py` would be appropriate.

.librarian/state.yaml
```diff
@@ -9,18 +9,18 @@ libraries:
...
-      - docs/README.rst
-      - samples/README.txt
-      - tar.gz
-      - gapic_version.py
-      - samples/generated_samples/snippet_metadata_
-      - scripts/client-post-processing
-      - samples/snippets/README.rst
-      - tests/system
+      - ^packages/google-cloud-dlp/CHANGELOG.md
+      - ^packages/google-cloud-dlp/docs/CHANGELOG.md
```
jskeet, line 13: Nit: all of the dots in here might be worth escaping. It's unlikely that there'll be a file called `packages/google-cloud-dlp/docs/CHANGELOGxmd` but at the moment that would match...

.librarian/state.yaml
parthea: This is a good call out. This could be fixed but instead I opted to just remove `tar.gz` from the `preserve_regex` list. The only reason it was preserved was to reduce the diff however the diff will be minimal if we allow this file to be deleted.  This tar file was created by bazel. We're no longer using bazel, so we will only see a diff at the next generation and at onboarding of libraries.

See related fix https://github.com/googleapis/synthtool/pull/2098

.librarian/state.yaml
parthea: This was deliberate. We want to match patterns like `packages/google-cloud-dlp/x/y/z/gapic_version.py`

.librarian/state.yaml
jskeet: Sounds good. For go, we're making a few one-off changes as part of migration (i.e. at the time of introducing the library into the state file) so that the diff itself from Librarian will be closer to zero. Just something to consider as a migration step - it makes no odds in the long run of course.

parthea: Make sense. I will open a separate PR to remove the existing tar.gz files so that the diff will remain 0

parthea:
Closing as per an offline discussion today with @jskeet where we are switching from regex to glob, so polishing the regex is no longer a priority. (See https://github.com/googleapis/librarian/issues/2294)

PR #14512, opened 2025-09-22
chore(librarian): onboard google-apps-card and google-apps-chat

parthea:
Towards https://github.com/googleapis/librarian/issues/761


jskeet (commented):
Is the plan to merge this before any changes to the regexes to use anchors etc?

parthea:
I'll work on wrapping up https://github.com/googleapis/google-cloud-python/pull/14464

parthea:
We are switching from regex to glob, so polishing the regex is no longer a priority. (See https://github.com/googleapis/librarian/issues/2294)

PR #14518, opened 2025-09-23
feat: implement basic librarian configure command

ohmayr:
Towards https://github.com/googleapis/librarian/issues/466 🦕

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -114,31 +114,103 @@ def _write_json_file(path: str, updated_content: Dict):
     with open(path, "w") as f:
         json.dump(updated_content, f, indent=2)
         f.write("\n")
- 
+
+def _add_new_library_source_roots(library_config: Dict, library_id: str) -> None:
+    """Adds the default source_roots to the library configuration if not present.
+
+    Args:
+        library_config(Dict): The library configuration.
+        library_id(str): The id of the library.
+    """
+    if library_config["source_roots"] is None:
+        library_config["source_roots"] = [f"packages/{library_id}"]
+
+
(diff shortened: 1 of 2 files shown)


.generator/cli.py
```diff
@@ -45,6 +45,7 @@
 
 BUILD_REQUEST_FILE = "build-request.json"
 GENERATE_REQUEST_FILE = "generate-request.json"
+CONFIGURE_REQUEST_FILE = "configure-response.json"
```
```suggestion
CONFIGURE_REQUEST_FILE = "configure-request.json"
```

.generator/cli.py
```diff
@@ -113,6 +114,31 @@ def _write_json_file(path: str, updated_content: Dict):
...
+    Returns:
+        Dict: The configuration of a new library.
+    """
+    new_library_config = {}
+    for library_config in request_data.get("libraries", []):
+        all_apis = library_config.get("apis", [])
+        for api in all_apis:
+            if api.get("status") == "new":
+                new_library_config = library_config
+                break
```
parthea: do we also need to break from the top level for loop? Should we move this to a separate helper function so it's easier to return on the first `new` library?

.generator/cli.py
```diff
@@ -113,6 +114,31 @@ def _write_json_file(path: str, updated_content: Dict):
...
+    
+    # remove status key from new library config.
+    all_apis = new_library_config.get("apis", [])
```
parthea: We already have `all_apis` at line 130. Do we need to declare it again? This may be moot if we refactor the above code into a separate function that returns the expected content of `all_apis`

PR #14545, opened 2025-09-24
feat: create repo metadata from service config

ohmayr:
Fixes https://github.com/googleapis/librarian/issues/2332 🦕

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -398,22 +398,37 @@ def _create_repo_metadata_from_service_config(
         Dict: The content of the .repo-metadata.json file.
     """
     full_service_config_path = f"{source}/{api_path}/{service_config_name}"
+    with open(full_service_config_path, "r") as f:
+        service_config = yaml.safe_load(f)
+
+    publishing = service_config.get("publishing", {})
+    name_pretty = service_config.get("title", "")
+    product_documentation = publishing.get("product_documentation", "")
+    api_shortname = service_config.get("name", "").split(".")[0]
+    distribution_name = publishing.get("library_name", library_id)
+    documentation = service_config.get("documentation", {})
+    api_description = documentation.get("summary", "")
+    issue_tracker = service_config.get("new_issue_uri", "https://github.com/googleapis/google-cloud-python/issues")
+
+    # TODO(ohmayr): Determine release level from api_path
+    release_level = "stable"
+
 
     # TODO(https://github.com/googleapis/librarian/issues/2332): Read the api
     # service config to backfill `.repo-metadata.json`.
     return {
-        "api_shortname": "",
-        "name_pretty": "",
-        "product_documentation": "",
-        "api_description": "",
-        "client_documentation": "",
-        "issue_tracker": "",
-        "release_level": "",
+        "api_shortname": api_shortname,
+        "name_pretty": name_pretty,
+        "product_documentation": product_documentation,
+        "api_description": api_description,
+        "client_documentation": f"https://cloud.google.com/python/docs/reference/{distribution_name}/latest",
+        "issue_tracker": issue_tracker,
+        "release_level": release_level,
         "language": "python",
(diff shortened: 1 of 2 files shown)


.generator/cli.py
```diff
@@ -398,22 +398,35 @@ def _create_repo_metadata_from_service_config(
...
     """
     full_service_config_path = f"{source}/{api_path}/{service_config_name}"
+    with open(full_service_config_path, "r") as f:
+        service_config = yaml.safe_load(f)
+
+    publishing = service_config.get("publishing", {})
+    name_pretty = service_config.get("title", "")
+    product_documentation = publishing.get("product_documentation", "")
+    api_shortname = service_config.get("name", "").split(".")[0]
+    distribution_name = publishing.get("library_name", library_id)
```
```suggestion
    distribution_name = library_id
```

`library_name` does not appear in the publishing section
https://github.com/googleapis/googleapis/blob/master/google/api/client.proto

.generator/cli.py
```diff
@@ -398,22 +398,35 @@ def _create_repo_metadata_from_service_config(
         Dict: The content of the .repo-metadata.json file.
     """
     full_service_config_path = f"{source}/{api_path}/{service_config_name}"
+    with open(full_service_config_path, "r") as f:
+        service_config = yaml.safe_load(f)
+
+    publishing = service_config.get("publishing", {})
+    name_pretty = service_config.get("title", "")
+    product_documentation = publishing.get("product_documentation", "")
```
parthea: Do you mean `documentation_uri`?

```suggestion
    product_documentation = publishing.get("documentation_uri", "")
```
https://github.com/googleapis/googleapis/blob/68bc915cb97cdd13347215377f2ca997dff245ad/google/api/client.proto#L185

.generator/cli.py
```diff
@@ -383,6 +381,19 @@ def _clean_up_files_after_post_processing(output: str, library_id: str):
...
 
 
+def _determine_release_level_from_api_path(api_path: str) -> str:
+    """Determines the release level from the API path.
+
+    Args:
+        api_path (str): The path to the API.
+
+    Returns:
+        str: The release level, which can be 'alpha', 'beta', or 'stable'.
```
```suggestion
        str: The release level, which can be 'preview', or 'stable'.
```

.generator/cli.py
```diff
@@ -383,6 +381,24 @@ def _clean_up_files_after_post_processing(output: str, library_id: str):
...
+    Args:
+        api_path (str): The path to the API.
+
+    Returns:
+        str: The release level, which can be 'preview' or 'stable'.
+    """
+    version = Path(api_path).name
+    if "beta" in version or "alpha" in version:
+        return "preview"
+    return "stable"
```
parthea: Can we omit `_determine_release_level_from_api_path` from this PR since it requires a design doc?

.generator/cli.py
```diff
@@ -398,22 +414,39 @@ def _create_repo_metadata_from_service_config(
...
+        "issue_tracker": issue_tracker,
+        "release_level": release_level,
         "language": "python",
         "library_type": "GAPIC_AUTO",
         "repo": "googleapis/google-cloud-python",
-        "distribution_name": "",
-        "api_id": "",
+        "distribution_name": library_id,
+        "api_id": api_id,
     }
```
parthea, line 450: Add `default_version` which is needed for the post processor as per https://github.com/googleapis/synthtool/blob/bffd472817eb9333e68dca787b52b997c060386b/synthtool/languages/python_mono_repo.py#L254-L260

parthea: This may not be needed. We can omit it for now and add it if we find that it is needed.

.generator/cli.py
ohmayr: As discussed offline, this function is no longer called but kept around for reference if this logic is needed.

parthea (approved):

PR #14547, opened 2025-09-24
chore(librarian): use generated gapic_version.py

parthea:


.generator/cli.py
```diff
@@ -553,11 +552,12 @@ def _get_api_generator_options(api_path: str, py_gapic_config: Dict) -> List[str
                 # Other options use the value directly
                 generator_options.append(f"{protoc_key}={config_value}")
 
+    # The value of `opt_args` in the `py_gapic` bazel rule is already a list of strings.
+    optional_arguments = py_gapic_config.get("opt_args", [])
+    # Specify `gapic-version` using the version from `state.yaml`
+    optional_arguments.extend([f"gapic-version={gapic_version}"])
```
ohmayr, line 558: does it really matter what version we pass down from here if we know for sure that it'll be updated later by the `relese-init` call? Why not pass down an arbitrary version like `"0.0.0"` from here and make the logic simpler?

parthea: Passing down a specific version here is necessary to prevent excessive and unnecessary file changes during librarian generate. If we simply passed "0.0.0", files like `gapic_version.py` would reset their version from the last released value to "0.0.0" every time generate is run, even if the API definition hasn't changed. Multiplying this across 200 libraries would result in over 1000 files with useless diffs. Using the existing version avoids these large, noisy diffs. Also see related discussion here where we want to avoid noisy diffs: https://github.com/googleapis/librarian/issues/2362#issuecomment-3334242621

ohmayr (approved):

PR #14551, opened 2025-09-25
chore(librarian): add support for proto-only library

parthea:
Towards https://github.com/googleapis/librarian/issues/1269

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -330,23 +330,6 @@ def _copy_files_needed_for_post_processing(output: str, input: str, library_id: 
     os.makedirs(
         f"{output}/{path_to_library}/scripts/client-post-processing", exist_ok=True
     )
-    # TODO(https://github.com/googleapis/synthtool/pull/2126): Remove once this PR is merged
-    # This is needed to avoid the following error for proto-only libraries
-    # Traceback (most recent call last):
-    # File "/app/./cli.py", line 535, in handle_generate
-    #     _run_post_processor(output, library_id)
-    # File "/app/./cli.py", line 300, in _run_post_processor
-    #     python_mono_repo.owlbot_main(path_to_library)
-    # File "/usr/local/lib/python3.9/site-packages/synthtool/languages/python_mono_repo.py", line 310, in owlbot_main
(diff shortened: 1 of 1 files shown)


.generator/cli.py
```diff
@@ -300,6 +300,13 @@ def _run_post_processor(output: str, library_id: str):
         python_mono_repo.owlbot_main(path_to_library)
     else:
         raise SYNTHTOOL_IMPORT_ERROR  # pragma: NO COVER
+
+    # If there is no noxfile, run `isort`` and `black` on the output.
+    # This is required for proto-only libraries which are not GAPIC.
+    if not Path(f"{output}/{path_to_library}/noxfile.py").exists():
+        subprocess.run(["isort", output])
+        subprocess.run(["black", output])
+
```
ohmayr, line 309: If a noxfile is present, we don't run black to format files, right? We just check if the files are formatted (using black).

.generator/cli.py
```diff
@@ -598,6 +608,26 @@ def _get_api_generator_options(
...
+        tmp_dir (str): The temporary directory for protoc output.
+
+    Returns:
+        str: The complete protoc command string suitable for shell execution.
+    """
+    command_parts = [
+        f"protoc {api_path}/*.proto",
+        f"--python_out={tmp_dir}",
+        f"--pyi_out={tmp_dir}",
+    ]
```
ohmayr, line 626: Have we considered using grpc_tools.protoc instead of running the cli as a subprocess?

.generator/cli.py
parthea: The post processor runs black via the `format` nox session if it finds a noxfile. See  https://github.com/googleapis/synthtool/blob/e653c3e1b686f3c97ee13aaf2b8e72d2ee972862/synthtool/languages/python_mono_repo.py#L323-L327

.generator/cli.py
parthea: Yes, my initial thought is that it's better to use `protoc` directly because we know which version of `protoc` we're getting. When we use `grpc_tools.protoc`, we're relying on `grpc_tools` to choose the version of `protoc` which adds unnecessary complexity.

PR #14562, opened 2025-09-25
chore(librarian): onboard googleapis-common-protos to librarian

parthea:
Depends on https://github.com/googleapis/google-cloud-python/pull/14551

Towards https://github.com/googleapis/librarian/issues/1269

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -203,3 +203,46 @@ libraries:
     remove_regex:
       - packages/google-apps-chat
     tag_format: '{id}-v{version}'
+  - id: google-cloud-gkerecommender
+    version: 0.0.0
+    last_generated_commit: 329ace5e3712a2e37d6159d4dcd998d8c73f261e
+    apis:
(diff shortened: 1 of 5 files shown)


.librarian/state.yaml
```diff
@@ -986,3 +986,25 @@ libraries:
...
+      - path: google/rpc/context
+        service_config: ""
+      - path: google/type
+        service_config: type.yaml
+    source_roots:
+      - packages/googleapis-common-protos
+    preserve_regex: []
+    remove_regex:
+      - ^packages/googleapis-common-protos/google/(?:api|cloud|rpc|type)/.*/.*_pb2\.(?:py|pyi)$
+    tag_format: '{id}-v{version}'
```
ohmayr, line 1010: Is this handwritten or generated using the script?

parthea: This is handwritten. That should be ok for the 5 libraries which are proto-only

PR #14751, opened 2025-10-16
chore(librarian): create docs/README.rst file during generation

ohmayr:
This PR creates a `docs/index.rst` file for a new library that is configured and generated using librarian.

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -1,4 +1,4 @@
-image: python-librarian-generator:latest
+image: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/python-librarian-generator:latest
 libraries:
   - id: google-ads-admanager
     version: 0.5.0


.generator/cli.py
```diff
@@ -501,6 +500,47 @@ def _generate_repo_metadata_file(
...
+    destination_path = f"{docs_path}/README.rst"
+
+    # If the source file doesn't exist (not even as a broken symlink),
+    # there's nothing to copy.
+    if not os.path.lexists(source_path):
+        return
+
+    # Read the content from the source, which will resolve any symlinks.
+    with open(source_path, "r") as f:
+        content = f.read()
```
parthea, line 528: nit: use the `_read_text_file` function

https://github.com/googleapis/google-cloud-python/blob/79a526115b35fec0df48ca9512673eb5b96c641e/.generator/cli.py#L61-L72

.generator/cli.py
```diff
@@ -501,6 +500,47 @@ def _generate_repo_metadata_file(
     _write_json_file(output_repo_metadata, metadata_content)
 
 
+def _copy_readme_to_docs(output: str, library_id: str):
```
parthea, line 503: Could we create a follow up issue to track adding docs/README.rst to the output of `gapic-generator`?

parthea (approved):

PR #14901, opened 2025-11-20
chore(librarian): Add header to files under .librarian/generator-input

parthea:
Files residing in `.librarian/generator-input` are the source of truth and override their counterparts in the root directory. Modifications intended for files such as `setup.py` must be applied within `.librarian/generator-input`.

IOW, for `python-datastore`, we need to make changes in https://github.com/googleapis/python-datastore/blob/main/.librarian/generator-input/setup.py rather than https://github.com/googleapis/python-datastore/blob/main/setup.py because the source of truth is `.librarian/generator-input/setup.py`

This PR adds a header with the following text to all files in `.librarian/generator-input`, excluding JSON files which don't support comments.

```
# DO NOT EDIT THIS FILE OUTSIDE OF `.librarian/generator-input`
# The source of truth for this file is .librarian/generator-input
```

### DIFF
--- a/.generator/cli.py
+++ b/.generator/cli.py
@@ -383,9 +383,7 @@ def _add_header_to_files(directory: str) -> None:
                 line_index += 1
 
             header_prefix = "\n" if line_index > 0 else ""
-            lines.insert(
-                line_index, header_prefix + _GENERATOR_INPUT_HEADER_TEXT + "\n"
-            )
+            lines.insert(line_index, f"{header_prefix}{_GENERATOR_INPUT_HEADER_TEXT}\n")
 
             with open(file_path, "w", encoding="utf-8") as f:
                 f.writelines(lines)


.generator/cli.py
```diff
@@ -57,6 +57,11 @@
 SOURCE_DIR = "source"
 _GITHUB_BASE = "https://github.com"
 
+GENERATOR_INPUT_HEADER_TEXT = (
```
ohmayr:
```suggestion
_GENERATOR_INPUT_HEADER_TEXT = (
```

.generator/cli.py
```diff
@@ -349,6 +354,40 @@ def _run_post_processor(output: str, library_id: str, is_mono_repo: bool):
...
+
+            if file_path.suffix in skipped_extensions:
+                continue
+
+            with open(file_path, "r", encoding="utf-8") as f:
+                lines = f.readlines()
+
+            line_index = 0
+            # Skip the license header (contiguous block of comments starting with '#').
+            while line_index < len(lines) and lines[line_index].strip().startswith("#"):
```
ohmayr, line 382: What happens if a file in `generator-input` is missing a copyright header altogether?

.generator/cli.py
parthea: Fixed in https://github.com/googleapis/google-cloud-python/pull/14901/commits/58f73763b0529bcf6a674d4749c094c85f464a23

.generator/cli.py
parthea: There is already a test for this in `test_add_header_to_files_add_header_no_license` but I applied a fix in https://github.com/googleapis/google-cloud-python/pull/14901/commits/8d05b519c174873a5a18409cc7daca0eb12e8f1a to omit a new line if there are no comments in the file

ohmayr (approved):

PR #15485, opened 2026-01-26
chore: remove non-existent files from legacy-librarian preserve regexes

jskeet:
I'm currently regenerating everything to make sure this really doesn't remove anything it shouldn't (please don't merge until that's complete!) but this should make migration from legacy-librarian to librarian simpler, as librarian expects everything in its "keep" list to exist.

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -233,6 +233,7 @@ libraries:
     remove_regex:
       - ^packages/google-cloud-access-context-manager/google/.*/.*(?:\.proto|_pb2\.(?:py|pyi))$
       - .repo-metadata.json
+      - README.rst
       - docs/summary_overview.md
     tag_format: '{id}-v{version}'
   - id: google-cloud-advisorynotifications
@@ -463,6 +464,7 @@ libraries:
     remove_regex:
       - ^packages/google-cloud-audit-log/google/.*/.*(?:\.proto|_pb2\.(?:py|pyi))$
       - .repo-metadata.json
+      - README.rst
       - docs/summary_overview.md
     tag_format: '{id}-v{version}'
   - id: google-cloud-automl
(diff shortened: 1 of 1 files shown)


.librarian/state.yaml
```diff
@@ -544,7 +463,6 @@ libraries:
     remove_regex:
       - ^packages/google-cloud-audit-log/google/.*/.*(?:\.proto|_pb2\.(?:py|pyi))$
       - .repo-metadata.json
-      - README.rst
```
jskeet, line 547: Will re-add this and the few other ones.

.librarian/state.yaml
```diff
@@ -560,12 +478,10 @@ libraries:
     preserve_regex:
       - packages/google-cloud-automl/CHANGELOG.md
       - docs/CHANGELOG.md
-      - samples/README.txt
-      - samples/snippets/README.rst
-      - tests/system
       - docs/automl_v1beta1/tables.rst
       - google/cloud/automl_v1beta1/services/tables
       - samples/README
+      - tests/system
```
jskeet, line 486: This move is deliberate, to put all the entries in alphabetical order.

jskeet:
There's also an entry per library starting with `packages/`, e.g. `packages/google-ads-datamanager/CHANGELOG.md`. I believe all of those can be removed too. Let me know if you'd like those to be in this PR or a different one.

parthea:
For https://github.com/googleapis/google-cloud-python/pull/15485#issuecomment-3800772749, let's use a new PR. The files are handwritten, so we will want to preserve them, rather than delete them.

parthea:
> I'm currently regenerating everything to make sure this really doesn't remove anything it shouldn't (please don't merge until that's complete!) but this should make migration from legacy-librarian to librarian simpler, as librarian expects everything in its "keep" list to exist.

I merged this PR because I am also planning to re-generate everything today as part of a gapic-generator update which only impacts `google-cloud-compute`. It will be easy to see if there are any regressions related to these files as part of that work.

parthea:
Just to close the loop here, the changes in this PR did not introduce any regressions. See https://github.com/googleapis/google-cloud-python/pull/15487 which included the fix in this PR.

jskeet:
To also close the loop: I checked the `packages/google-ads-datamanager/CHANGELOG.md` aspects, and I was wrong about that. They're repo-relative, not source-root relative. So I won't be making any changes there. (The Librarian migration tool gets confused though; I've raised an issue around that.)

Issue #14992, opened 2026-01-06
library generation fails for 3 libraries


ldetmer:
### Determine this is the right repository

- [x] I determined this is the correct repository in which to report this bug.

### Summary of the issue

When running `librarian generate` 3 libraries are failing to generate.  See details below and full logs [here](https://pantheon.corp.google.com/cloud-build/builds/ebea3d57-f323-4081-aa09-a6fb482acde6;step=4?invt=AcGNNA&mods=allow_workbench_image_override&project=cloud-sdk-librarian-prod)

google-cloud-compute, google-cloud-compute-v1beta

```

Traceback (most recent call last):
--
  | File "/app/./cli.py", line 1015, in _run_nox_sessions
  | _run_individual_session(nox_session, library_id, repo, is_mono_repo)
  | ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/app/./cli.py", line 1048, in _run_individual_session
  | result = subprocess.run(command, text=True, check=True, timeout=600)
  | File "/usr/local/lib/python3.14/subprocess.py", line 556, in run
  | stdout, stderr = process.communicate(input, timeout=timeout)
  | ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/local/lib/python3.14/subprocess.py", line 1220, in communicate
  | stdout, stderr = self._communicate(input, endtime, timeout)
  | ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/local/lib/python3.14/subprocess.py", line 2152, in _communicate
  | self.wait(timeout=self._remaining_time(endtime))
  | ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/local/lib/python3.14/subprocess.py", line 1278, in wait
  | return self._wait(timeout=timeout)
  | ~~~~~~~~~~^^^^^^^^^^^^^^^^^
  | File "/usr/local/lib/python3.14/subprocess.py", line 2056, in _wait
  | raise TimeoutExpired(self.args, timeout)
[...]

PR #15501, opened 2026-02-02
chore(librarian): temporarily add gdb for remote debugging

parthea:
Towards https://github.com/googleapis/google-cloud-python/issues/14992


parthea:
`python-librarian-generator-presubmit (cloud-sdk-python-ci)` passed. The other presubmit checks are not applicable. They are currently stuck due to an outage with GHA.

PR #15536, opened 2026-02-06
chore: update librarian container to use ruff

ohmayr:
Update the container cli tool to use ruff instead of black and isort.


parthea:
~Wait until the mypy check passes in https://github.com/googleapis/google-cloud-python/pull/15539~

EDIT: We will tackle this separately

PR #15561, opened 2026-02-11
chore: add ruff to librarian container

ohmayr:
add ruff to librarian container


parthea:
Prefer https://github.com/googleapis/google-cloud-python/pull/15536

PR #15567, opened 2026-02-13
chore(migration): Migrate code from googleapis/python-api-core into packages/google-api-core

parthea:
See #10892. 


parthea:
This PR is ready for review but not merge

PR #15568, opened 2026-02-13
chore(migration): Migrate code from googleapis/proto-plus-python into packages/proto-plus

parthea:
See #14906. 


parthea:
This PR is ready for review but not merge

Issue #11029, opened 2023-04-04
Adopt split repo: _python-error-reporting_


vchudnov-g:
Migrate the split-repo https://github.com/googleapis/python-error-reporting to https://github.com/googleapis/google-cloud-python.

The migration readiness criteria are the following, which we track via GitHub labels on this issue. These criteria apply to the split repo we are migrating from:
- No open issues
- No open PRs
- No handwritten samples
- No system tests
- No client-specific customizations

parthea:
This repository contains handwritten code here:
https://github.com/googleapis/python-error-reporting/tree/main/google/cloud/error_reporting

I'm going to mark this as blocked while we determine if python-error-reporting should be migrated to the mono repository.

parthea:
Closing as the migration is now complete. We can re-open this issue if we want to onboard this repository to the monorepo.

PR #15569, opened 2026-02-13
chore(migration): Migrate code from googleapis/python-error-reporting into packages/google-cloud-error-reporting

parthea:
See #11029. 


chalmerlowe (approved):
LGTM

Issue #14900, opened 2025-11-20
bigquery-storage read session to_dataframe() loses 1152 records per stream


patricksurry:
### Determine this is the right repository

- [x] I determined this is the correct repository in which to report this bug.

### Summary of the issue

**Context**
This is related to Google Cloud Support ticket 65020767 where I was asked to file a report here as they consider it out of scope.

I am using the bigquery-storage read client to fetch multiple streams from a static BQ table.   I noticed that the client.read_rows(stream.name).to_dataframe() method loses 1152 records per stream, compared to the client.read_rows(stream.name).rows() which produces the correct result.

**Expected Behavior:**
See example code below.   The correct result from a SQL count(*) on the table gives 7,303,007 records.

I expected that my code with to_dataframe() would fetch the same number of records.

**Actual Behavior:**

Instead I received these results for different values of the max_stream_count argument.  By default (no argument) it produced 40 streams.  In each case the total row count was 9*128*num_streams below the correct value.

```
max_stream_count,  result
1,   7301855 [7301855]        # missing 1152 rows = 1 * 9*128
2,   7300703 [3467413, 3833290].  # missing 2304 rows = 2 * 9*128
8,   7293791 [1094808, 911159, 911286, 911386, 910562, 912084, 912954, 729552]. # missing 9216 rows = 8 * 9*128
none (40 by default),  7256927 [181563, 180985, 181974, 181696, 181770, 181694, 181110, 181662, 181353, 181152, 181342, 180928, 182225, 181868, 180987, 180512, 181476, 181825, 181460, 181902, 181510, 180666, 181595, 180857, 181766, 180494, 181289, 181609, 181403, 182320, 181416, 181210, 182091, 181602, 181844, 180852, 181417, 182208, 181037, 180257]
    # missing 46,080 rows = 40 * 9*128
```

### API client name and version

google.cloud.bigquery_storage==2.34.0

### Reproduction steps: code

```
from google.cloud.bigquery_storage import BigQueryReadClient, ReadSession, DataFormat

client = BigQueryReadClient()

project_id = "gcp-hopper-ds-research"
dataset_id = "psurry"
table_id = "tmp_trips_10M"

# Configure the read session
read_session = ReadSession(
    table=f'projects/{project_id}/datasets/{dataset_id}/tables/{table_id}',
    data_format=DataFormat.AVRO,
)

# Create the read session
session = client.create_read_session(
    parent=f"projects/{project_id}",
    read_session=read_session,
)
[...]

PR #15575, opened 2026-02-18
fix: missing first page in to_dataframe

daniel-sanche:
There was a bug in ReadRowsIterable.to_dataframe(), where it would attempt to use `.to_arrow()` to build a dataframe, before falling back to a generic method if `.to_arrow` was unsupported. 

This was a problem, because both methods read from a single `self.pages` iterator, so the first attempt would pull a page of data out of the feed, which would then be lost if `to_arrow()` is unsupported

This PR fixes the bug by pulling a single item off of the stream and calling `to_arrow` on it individually, instead of reading off the self.pages generator directly

Fixes https://github.com/googleapis/google-cloud-python/issues/14900

### DIFF
--- a/packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
+++ b/packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
@@ -440,7 +440,10 @@ class ReadRowsIterable(object):
         is initialized.
         """
         if self._stream_parser is None:
-            return pandas.DataFrame(columns=dtypes.keys()).astype(dtypes)
+            df = pandas.DataFrame(columns=dtypes.keys())
+            for col, dtype in dtypes.items():
+                df[col] = pandas.Series([], dtype=dtype)
+            return df
 
         if isinstance(self._stream_parser, _ArrowStreamParser):
             self._stream_parser._parse_arrow_schema()


packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
```diff
@@ -396,38 +400,67 @@ def to_dataframe(self, dtypes=None):
...
+            # usually no-copy.
+            record_batches = [
+                p.to_arrow() for p in itertools.chain([first_page], pages)
+            ]
+
+            table = pyarrow.Table.from_batches(record_batches)
+            df = table.to_pandas()
             for column in dtypes:
                 df[column] = pandas.Series(df[column], dtype=dtypes[column])
             return df
```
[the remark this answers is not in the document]
daniel-sanche: Yes, this code assumes if the first page is in arrow format, the rest of the stream will be as well. @Linchin let me know if that's a wrong assumption

I think I'll change the code to raise an exception if we do get in this unexpected state though, to be explicit about what we expect

packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
```diff
@@ -396,38 +399,75 @@ def to_dataframe(self, dtypes=None):
...
+            self._stream_parser._parse_arrow_schema()
+            df = pyarrow.Table.from_batches(
+                [], schema=self._stream_parser._schema
+            ).to_pandas()
+            for column, dtype in dtypes.items():
+                if column in df.columns:
+                    df[column] = pandas.Series(df[column], dtype=dtype)
+                else:
+                    df[column] = pandas.Series([], dtype=dtype)
+            return df
```
parthea, line 455: I asked Gemini for a code review, and it suggested to simplify this block to

```
        if isinstance(self._stream_parser, _ArrowStreamParser):
            self._stream_parser._parse_arrow_schema()
            
            # Initialize DataFrame directly from the Arrow schema
            df = self._stream_parser._schema.empty_table().to_pandas()

            # Apply dtypes efficiently
            for column, dtype in dtypes.items():
                df[column] = pandas.Series(df.get(column, []), dtype=dtype)
                
            return df
```


I ran `nox -s unit-3.14 -r -- -k test_to_dataframe_empty_w_dtypes_arrow` locally and it passed with this refactor

packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
```diff
@@ -396,38 +399,75 @@ def to_dataframe(self, dtypes=None):
...
+        This handles cases where the stream is empty but we still need to
+        return a DataFrame with the correct columns and types. It handles
+        both Arrow and Avro parsers, as well as the case where no parser
+        is initialized.
+        """
+        if self._stream_parser is None:
+            df = pandas.DataFrame(columns=dtypes.keys())
+            for col, dtype in dtypes.items():
+                df[col] = pandas.Series([], dtype=dtype)
+            return df
```
parthea: I asked Gemini for a code review, and it suggested to simplify this block to
```
if self._stream_parser is None:
    return pandas.DataFrame(columns=dtypes.keys()).astype(dtypes)
```

I ran unit tests with this suggestion and they passed

packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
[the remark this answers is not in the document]
Linchin: I think it's reasonable to expect the following pages to be arrow too. Thanks for adding the exception, just in case.

packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
daniel-sanche: Thanks, yeah that looks like it should be equivalent. I made the change

packages/google-cloud-bigquery-storage/google/cloud/bigquery_storage_v1/reader.py
daniel-sanche: I made this change, but had to revert it. It seems Python 3.7 doesn't like this pattern when dtypes is empty

PR #15580, opened 2026-02-18
chore: update handwritten noxfiles to use ruff

ohmayr:
update handwritten noxfiles to use ruff


packages/google-cloud-access-context-manager/noxfile.py
```diff
@@ -102,10 +101,14 @@ def lint(session):
     Returns a failure if the linters find linting errors or sufficiently
     serious code quality issues.
     """
-    session.install("flake8", BLACK_VERSION)
+    session.install("flake8", RUFF_VERSION)
+
+    # 2. Check formatting
```
vchudnov-g: There is no `#1` in this comment:

```suggestion
    # Check formatting
```

Same for the other files.

ohmayr: addressed. Thank you!

PR #15592, opened 2026-02-20
chore(migration): Migrate code from googleapis/python-datastore into packages/google-cloud-datastore

parthea:
See #11015. 


ohmayr:
This CL is ready for review but not merge.

PR #15615, opened 2026-02-26
fix: replace bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


parthea:
Hi @haosenwang1018, Thanks for opening a PR! Regrettably, the files in this PR are automatically generated using a code generator in https://github.com/googleapis/gapic-generator-python. If you're interested to submit a fix there, pull requests are welcome! The source of truth lives here: https://github.com/googleapis/gapic-generator-python/blob/535d65420751b5959f8e960ee532726da00159b6/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/rest.py.j2#L275

Once the fix is merged in `gapic-generator-python`, it will be rolled out to `google-cloud-python` automatically using our tooling which is referred to as [librarian](https://github.com/googleapis/librarian).

Issue #11017, opened 2023-04-04
Adopt split repo: _python-db-dtypes-pandas_


vchudnov-g:
Migrate the split-repo https://github.com/googleapis/python-db-dtypes-pandas to https://github.com/googleapis/google-cloud-python.

The migration readiness criteria are the following, which we track via GitHub labels on this issue. These criteria apply to the split repo we are migrating from:
- No open issues
- No open PRs
- No handwritten samples
- No system tests
- No client-specific customizations

parthea:
I've marked this issue as blocked because this specific library is a connector. There are no plans to migrate connectors to the mono repository. I'm going to keep this issue open so that we can track related discussions in one place.

meredithslota:
I actually don't think we should adopt this at all, because we're planning to remove our dependency on it in the BigQuery libraries (I think). @mrfaizal to confirm (assigning to Faizal in the meantime).

parthea:
Closing as the migration is now complete. We can re-open this issue if we want to onboard this repository to the monorepo.

chalmerlowe:
Discussed with @parthea 
The previous comment about not migrating this to the google-cloud-python mono-repo is OBE.
The current plan is to migrate in the next month or so.

Issue #15618, opened 2026-02-09
Can we remove the restriction to `pandas<3.0.0`?


Natim:
For some reason, we have a restriction to `pandas<3.0.0` which wasn't the case for `db-dtypes==1.4.4`

parthea:
Transferring to google-cloud-python as the code for [python-db-dtypes-pandas](https://github.com/googleapis/python-db-dtypes-pandas) will be migrated there next week

jbbqqf:
Hi! Triaging older issues — this looks like a duplicate of #15660.

Both issues ask the same thing: lift the `pandas<3.0.0` cap on `pandas-gbq` so the package can be installed alongside pandas 3.x. #15660 has the more complete reporter context (pandas 3.0.1 release notes, `pip` resolver downgrade behavior).

If you agree, would you mind closing in favour of #15660 (and adding a 👍 there if you want to signal demand)? Happy to reopen if your symptoms diverge from what's tracked over there.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the cross-link was verified manually.)

Issue #15509, opened 2025-12-12
(db-dtypes) Support pandas 3.0.0 


daniel-sanche:
The compliance checks are currently alerting that the library does not work with the upcoming pandas 3.0.0 release. I have [added a constraint to the dependency list](https://github.com/googleapis/python-db-dtypes-pandas/pull/380) for now. Eventually, this should be addressed

nicomarechal:
The issue isn't tracked anymore by [python-db-dtypes-pandas](https://github.com/googleapis/python-db-dtypes-pandas) ... and hasn't been fixed

jm-glowienke:
Hi @tswast, do you have any idea when this issue will be resolved?

tswast:
Hi @jm-glowienke, thanks for pinging. That got it out of my email filter. :-) I'll take a look.

jm-glowienke:
Thank you so much @tswast for taking care of this so quickly 😄

PR #16012, opened 2026-03-02
chore(migration): Migrate code from googleapis/python-db-dtypes-pandas into packages/db-dtypes

parthea:
See #11017. 

### DIFF
--- a/packages/db-dtypes/tests/compliance/json/test_json_compliance.py
+++ b/packages/db-dtypes/tests/compliance/json/test_json_compliance.py
@@ -118,7 +118,7 @@ class TestJSONArrayGetitem(base.BaseGetitemTests):
             data.take([0, -2], fill_value=na_value, allow_fill=True)
 
     def test_getitem_propagates_readonly_property(self, data):
-         pytest.xfail("Failing with pandas prerelease: readonly propagation issue")
+        pytest.xfail("Failing with pandas prerelease: readonly propagation issue")
 
 
 class TestJSONArrayIndex(base.BaseIndexTests):
--- a/packages/db-dtypes/tests/compliance/time/test_time_compliance.py
+++ b/packages/db-dtypes/tests/compliance/time/test_time_compliance.py
@@ -69,7 +69,9 @@ class TestGetitem(base.BaseGetitemTests):
 
 class TestGroupby(base.BaseGroupbyTests):
     def test_groupby_agg_extension(self, data_for_grouping):
-        pytest.xfail("Failing with pandas prerelease: dtype mismatch (object vs dbtime)")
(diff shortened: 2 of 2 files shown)


packages/db-dtypes/noxfile.py
```diff
@@ -0,0 +1,571 @@
...
+
+@nox.session(python=UNIT_TEST_PYTHON_VERSIONS[-1])
+def compliance_prerelease(session):
+    """Run the compliance test suite with prerelease dependencies."""
+    prerelease(session, os.path.join("tests", "compliance"))
+
+
+@nox.session(python=UNIT_TEST_PYTHON_VERSIONS)
+def unit(session):
+    """Run the unit test suite."""
```
chalmerlowe, line 291:
```suggestion
    """Run the unit test suite."""

    if session.python in ("3.7",):
        session.skip("Python 3.7 is no longer supported")
```

packages/db-dtypes/noxfile.py
```diff
@@ -0,0 +1,574 @@
...
+FLAKE8_VERSION = "flake8==6.1.0"
+BLACK_VERSION = "black[jupyter]==23.7.0"
+ISORT_VERSION = "isort==5.11.0"
+LINT_PATHS = ["docs", "db_dtypes", "tests", "noxfile.py", "setup.py"]
+
+DEFAULT_PYTHON_VERSION = "3.9"
+LINT_PYTHON_VERSION = "3.10"
+
+UNIT_TEST_PYTHON_VERSIONS: List[str] = [
+    "3.9",
```
chalmerlowe, line 40:
```suggestion
    "3.7",
    "3.8",
    "3.9",
```

packages/db-dtypes/noxfile.py
```diff
@@ -0,0 +1,501 @@
...
+        "--prefer-binary",
+        "--pre",
+        "--upgrade",
+        "pyarrow",
+    )
+    session.install(
+        "--prefer-binary",
+        "--pre",
+        "--upgrade",
+        "pandas<3.0.0rc0",
```
parthea: We should run the prerelease test against the latest version of pandas.  It's ok if the prerelease test fails. It won't block development. We need to keep track of the failure

```suggestion
        "pandas",
```

parthea:
The issue with the prerelease test is tracked in https://github.com/googleapis/google-cloud-python/issues/15509 and https://github.com/googleapis/google-cloud-python/issues/15618

parthea:
Follow up work to allow pandas 3.x to be completed as part of https://github.com/googleapis/google-cloud-python/issues/15509 and https://github.com/googleapis/google-cloud-python/issues/15618

PR #16022, opened 2026-03-04
fix: gracefully fallback if workload fields are missing from cert config

nbayati:
Prevents exceptions during gECC flows or when falling back to SecureConnect by returning None instead of raising ClientCertError when X.509 workload fields are absent.

### DIFF
--- a/packages/google-auth/google/auth/transport/_mtls_helper.py
+++ b/packages/google-auth/google/auth/transport/_mtls_helper.py
@@ -182,6 +182,7 @@ def _get_workload_cert_and_key_paths(config_path, include_context_aware=True):
         return None, None
 
     data = _load_json_file(absolute_path)
+
     if "cert_configs" not in data:
         raise exceptions.ClientCertError(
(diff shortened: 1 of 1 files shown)


packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -183,36 +183,24 @@ def _get_workload_cert_and_key_paths(config_path, include_context_aware=True):
...
+    # We return None, None if the expected workload fields are not present.
+    # The certificate config might be present for other types of connections (e.g. gECC),
+    # and we want to gracefully fallback to testing other mTLS configurations
+    # like SecureConnect instead of throwing an exception.
     if "cert_configs" not in data:
-        raise exceptions.ClientCertError(
-            'Certificate config file {} is in an invalid format, a "cert configs" object is expected'.format(
-                absolute_path
-            )
-        )
```
andyrzhao, line 191: I would say that we can keep throwing an exception if "cert_configs" is missing as this field is a top-level requirement for the cert config spec. Thanks!

PR #16028, opened 2026-03-05
feat(firestore): literals pipeline stage

Linchin:
Succeeding https://github.com/googleapis/python-firestore/pull/1170 for the monorepo migration.

### DIFF
--- a/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
+++ b/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
@@ -346,7 +346,7 @@ class Limit(Stage):
 class Literals(Stage):
     """Returns documents from a fixed set of predefined document objects."""
 
-    def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
+    def __init__(self, *documents: dict[str, Expression | CONSTANT_TYPE]):
         super().__init__("literals")
         self.documents: tuple[Mapping[str, Any], ...] = documents
 


packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,71 @@ def find_nearest(
...
+                {"x": 11},
+                {"x": 3}
+            ]
+            ```
+
+        Args:
+            documents: A `str` or `Selectable` expression. If a `str`, it's
+                       treated as a field path to an array of documents.
+                       If a `Selectable`, it's usually a `Constant`
+                       containing an array of documents (as dictionaries).
```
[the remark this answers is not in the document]
daniel-sanche: +1. Usually I'd put the * in here too: `*documents: `

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,71 @@ def find_nearest(
             stages.FindNearest(field, vector, distance_measure, options)
         )
 
+    def literals(self, *documents: str | Selectable) -> "_BasePipeline":
```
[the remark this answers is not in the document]
daniel-sanche: You can probably ignore this, unless that's how other languages handle it

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
```diff
@@ -684,4 +684,23 @@ tests:
...
           - fieldReferenceValue: awards
           - stringValue: full_replace
-          name: replace_with
\ No newline at end of file
+          name: replace_with
+  - description: literals
+    pipeline:
+      - Literals:
+        - title: "The Hitchhiker's Guide to the Galaxy"
+          author: "Douglas Adams"
```
daniel-sanche, line 692: Maybe it's been a while since I looked at this, but this doesn't seem like the right syntax to me. Isn't this essentially sending `Literals({"title": ""The Hitchhiker's Guide to the Galaxy", "author", "Douglas Adams"})`? That doesn't seem to match the `str | Selectable`

Does the test pass?

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,71 @@ def find_nearest(
...
+
+        Example:
+            >>> from google.cloud.firestore_v1.pipeline_expressions import Constant
+            >>> documents = [
+            ...     {"name": "joe", "age": 10},
+            ...     {"name": "bob", "age": 30},
+            ...     {"name": "alice", "age": 40}
+            ... ]
+            >>> pipeline = client.pipeline()
+            ...     .literals(Constant.of(documents))
```
daniel-sanche: Looking at the code, it seems like:
1. [Constant isn't a Selectable](https://github.com/googleapis/google-cloud-python/blob/22601c79a6c0d6c470d338245ee247e9320b733f/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L1617)
2. Constant [doesn't seem like it supports dict types](https://github.com/googleapis/google-cloud-python/blob/22601c79a6c0d6c470d338245ee247e9320b733f/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L37). (We do have a [Map](https://github.com/googleapis/google-cloud-python/blob/22601c79a6c0d6c470d338245ee247e9320b733f/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L1887), which serves that purpose, but it doesn't seem Selectable either)

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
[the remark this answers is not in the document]
daniel-sanche: Actually, [looking at go](https://github.com/bhshkh/google-cloud-go/blob/90d4c50a09beb0ffcc3748005e336c7d4d15e043/firestore/pipeline_source.go#L224), it seems like it accepts dicts, but not strings? 


I don't know much about this stage, but from what I've seen, it's supposed to deal with maps. So maybe this should be `def literals(self, *documents: Map | dict[str, CONSTANT_TYPE] | Selectable)`?

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_stages.py
```diff
@@ -517,6 +517,35 @@ def test_to_pb(self):
...
+class TestLiterals:
+    def _make_one(self, *args, **kwargs):
+        return stages.Literals(*args, **kwargs)
+
+    def test_ctor(self):
+        val1 = Constant.of({"a": 1})
+        val2 = Constant.of({"b": 2})
+        instance = self._make_one(val1, val2)
+        assert instance.documents == (val1, val2)
+        assert instance.name == "literals"
```
daniel-sanche, line 529: We should have tests that cover all supported input types. I don't see anything using str (and looking at go, I'm not sure if we should be supporting str?)

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
```diff
@@ -684,4 +684,23 @@ tests:
...
+      pipeline:
+        stages:
+        - args:
+          - mapValue:
+              fields:
+                author:
+                  stringValue: "Douglas Adams"
+                title:
+                  stringValue: "The Hitchhiker's Guide to the Galaxy"
+          name: literals
```
daniel-sanche: We should also have tests here that cover the different input types we support

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
Linchin: Thanks for catching this! I spent some time to check the internal docs, and I think `Expression` class and `dict` should be supported per the following language:
```
            While literal values are the most common, it is also possible to pass in
            expressions, which will be evaluated and returned, making it possible to test
            out different query / expression behavior without first needing to create some
            test data.
```

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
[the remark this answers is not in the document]
Linchin: Upon further thoughts, I think it should be `def literals(self, *documents: dict | Expression):`. In this case both `Constant` and `Map` are child classes of `Expression`.

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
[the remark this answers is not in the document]
Linchin: Updated the docstring to be more accurate.

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
Linchin: It does pass. I think `Literals` should be able to accept dicts, as well as other `Expression` types. I have updated the system test to include `dict` and `Constant` (a child class of `Expression`).

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
Linchin: Good catch! I added additional type to test.

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_stages.py
Linchin: Indeed, I don't think we should support str. Here the test includes `dict` and `Constant`, which serves as an example of `Expression`'s child class. It seems too cumbersome to test for each and every child class of `Expression`? What do you think.

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,69 @@ def find_nearest(
...
+
+        Example:
+            >>> from google.cloud.firestore_v1.pipeline_expressions import Constant
+            >>> documents = [
+            ...     {"name": "joe", "age": 10},
+            ...     {"name": "bob", "age": 30},
+            ...     {"name": "alice", "age": 40}
+            ... ]
+            >>> pipeline = client.pipeline()
+            ...     .literals(documents)
```
daniel-sanche: it seems like this should be `.literals(*documents)`

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,69 @@ def find_nearest(
...
+            ```json
+            [
+                {"name": "joe", "age": 10},
+                {"name": "bob", "age": 30}
+            ]
+            ```
+
+        Behavior:
+            The `literals(...)` stage can only be used as the first stage in a pipeline (or
+            sub-pipeline). The order of documents returned from the `literals` matches the
```
daniel-sanche: It sounds like this is supposed to be a [pipleine_source?](https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py)

It looks like [that's how it's implemented in go](https://github.com/bhshkh/google-cloud-go/blob/90d4c50a09beb0ffcc3748005e336c7d4d15e043/firestore/pipeline_source.go#L224)

I guess you should move this to that file, and you can remove that part of the note

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,69 @@ def find_nearest(
...
+            For example, the following shows how to quickly test out the `length(...)`
+            function on some constant test sets:
+
+        Example:
+            >>> from google.cloud.firestore_v1.pipeline_expressions import Constant
+            >>> documents = [
+            ...     {"x": Constant.of("foo-bar-baz").char_length()},
+            ...     {"x": Constant.of("bar").char_length()}
+            ... ]
+            >>> pipeline = client.pipeline().literals(documents)
```
daniel-sanche: should be *documents

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
daniel-sanche: I think we still need more examples. Both of these are dict-like, so it seems like a pretty basic test. What if someone passes in `Constant(1)`? Or `Constant("test").byte_length()`? We say we support all expressions, how are non-dict types represented?

It would also be good to add some extra stages, to make sure this works like others

You can use gemini to create a few extra test scenarios. I usually don't include assert_proto on all of them, because it can be excessive

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_stages.py
daniel-sanche: Yeah, that should be fine

Although we don't support dict constants like this, so maybe change that part at least

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
```diff
@@ -684,4 +684,35 @@ tests:
...
+          name: replace_with
+  - description: literals
+    pipeline:
+      - Literals:
+        - title: "The Hitchhiker's Guide to the Galaxy"
+          author: "Douglas Adams"
+        - Constant:
+            value:
+              genre: "Science Fiction"
+              year: 1979
```
daniel-sanche: This looks like it resolves to `Constant({"genre": "Science Fiction", "year": 1979})`. But our [type hints forbid passing dicts as constants](https://github.com/googleapis/google-cloud-python/blob/d04a241d7cb91f31e93891bab753fb0d6d0e5471/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L37) like this (mostly because Python's typing system isn't powerful enough to type nested datastructures, but still)

We provide a [Map](https://github.com/googleapis/google-cloud-python/blob/d04a241d7cb91f31e93891bab753fb0d6d0e5471/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L1967), which is supposed to fill that gap. Maybe you should test that out, and see if it would work as a declared type here?

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,69 @@ def find_nearest(
             stages.FindNearest(field, vector, distance_measure, options)
         )
 
+    def literals(self, *documents: Expression | dict) -> "_BasePipeline":
```
daniel-sanche: Are you sure it makes sense to accept Expression here? Maybe it should just be `dict | Map`?

Expressions can evaluate to anything, and I don't know if it makes sense to have an int or bool Document

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,69 @@ def find_nearest(
...
+        """
+        Returns documents from a fixed set of predefined document objects.
+
+        This stage is commonly used for testing other stages in isolation,
+        though it can also be used as inputs to join conditions.
```
daniel-sanche: This seems to conflict with the statement later:

>   The `literals(...)` stage can only be used as the first stage in a pipeline (or sub-pipeline)

packages/google-cloud-firestore/tests/system/pipeline_e2e/general.yaml
Linchin: Referring to the [golang implementation](https://github.com/googleapis/google-cloud-go/blob/bb649988e1c66b4c0c8d1a2606c106eb1acd6f8b/firestore/pipeline_source.go#L224), the only type accepted is a list of key-value pairs. So I don't think `Constant(1)` or `Constant("test").byte_length()` are in scope. 

But I think it's a good idea to add more tests. I will also use `mapValue` instead of `Constant`.

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
Linchin: Indeed, the code has been moved to pipeline_source.py.

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
Linchin: thanks, corrected

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
```diff
@@ -108,3 +109,66 @@ def documents(self, *docs: "BaseDocumentReference") -> PipelineType:
             a new pipeline instance targeting the specified documents
         """
         return self._create_pipeline(stages.Documents.of(*docs))
+
+    def literals(self, *documents: dict[str, "Expression" | CONSTANT_TYPE]) -> PipelineType:
+        """
+        Returns documents from a fixed set of predefined document objects.
+
+        This stage is commonly used for testing other stages in isolation,
```
Linchin: @daniel-sanche I think what the documentation means is that, the Literals stage could serve as one of the inputs of a join operation, which is not contradictory to it being a first stage, so I think it is reasonable to keep this sentence and make it consistent with documentation.

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
Linchin: thanks for catching this, fixed

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
Linchin: See https://github.com/googleapis/google-cloud-python/pull/16028/changes#r2955450923

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
daniel-sanche: I'd still prefer to take it out since we don't yet support the join stage. But if you think it seems to match the coming feature, we can probably leave it

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
```diff
@@ -108,3 +109,68 @@ def documents(self, *docs: "BaseDocumentReference") -> PipelineType:
             a new pipeline instance targeting the specified documents
         """
         return self._create_pipeline(stages.Documents.of(*docs))
+
+    def literals(
+        self, *documents: Mapping[str, Expression | CONSTANT_TYPE]
```
daniel-sanche: I think we should keep this as `dict` instead of `Mapping`, just because Firestore has its own custom [Map type](https://github.com/googleapis/google-cloud-python/blob/ee7dd7dc168ecf943440cef38efff348e8eff095/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_expressions.py#L1967), which could get confusing

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
```diff
@@ -108,3 +109,68 @@ def documents(self, *docs: "BaseDocumentReference") -> PipelineType:
...
+            ```json
+            [
+                {"x": 11},
+                {"x": 3}
+            ]
+            ```
+
+        Args:
+            *documents: One or more documents to be returned by this stage. Each can be a `dict`
+                       or an `Expression`.
```
daniel-sanche: > Each can be a `dict` or an `Expression`.

This seems out of date? The annotations only support dicts (which seems to be right, since that's how other languages handle it)

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
```diff
@@ -342,6 +343,29 @@ def _pb_args(self):
...
+
+    def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
+        super().__init__("literals")
+        self.documents = documents  # type: ignore
+
+    def _pb_args(self):
+        args = []
+        for doc in self.documents:
+            if hasattr(doc, "_to_pb"):
+                args.append(doc._to_pb())
```
daniel-sanche: is this ever called? doc should be a dict, right? This section seems unreachable

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
```diff
@@ -342,6 +343,29 @@ def _pb_args(self):
         return [Value(integer_value=self.limit)]
 
 
+class Literals(Stage):
+    """Returns documents from a fixed set of predefined document objects."""
+
+    def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
+        super().__init__("literals")
+        self.documents = documents  # type: ignore
```
daniel-sanche: And why are you ignoring the type warning? What error were you seeing?

daniel-sanche: Consider converting Python-native types into Expressions here, like we do in the [other stages](https://github.com/googleapis/google-cloud-python/blob/ee7dd7dc168ecf943440cef38efff348e8eff095/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py#L361). This should make building _pr_args later cleaner, since everything would be an Expression with _to_pb

Something like:

```
def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
    super().__init__("literals")
     self.documents = [
        {k: v if isinstance(v, Expression) else Constant.of(v) for k,v in doc} 
        for doc in documents
    ]
```

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_source.py
```diff
@@ -110,6 +110,17 @@ def test_documents(self):
         assert first_stage.paths[1] == "/a/2"
         assert first_stage.paths[2] == "/a/3"
 
+    def test_literals(self):
+        from google.cloud.firestore_v1.pipeline_expressions import Field
+
+        instance = self._make_client().pipeline()
+        documents = (Field.of("a"), {"name": "joe"})
```
daniel-sanche: `Field.of("a")` isn't a dict

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_stages.py
```diff
@@ -517,6 +517,121 @@ def test_to_pb(self):
         assert len(result.options) == 0
 
 
+class TestLiterals:
+    def _make_one(self, *args, **kwargs):
+        return stages.Literals(*args, **kwargs)
+
+    def test_ctor(self):
+        val1 = Constant.of({"a": 1})
```
daniel-sanche: I think you should change this, since Constant doesn't support dict inputs (even if it technically works)

You could make this `{"a": Constant.of(1)}` to be conformant

It looks like this Constant.of(dict) pattern is used in a few other places in the tests

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
Linchin: Thanks for catching this, I have corrected it.

Linchin:
> [daniel-sanche](https://github.com/daniel-sanche)
[2 days ago](https://github.com/googleapis/google-cloud-python/pull/16028#discussion_r2956925200)
I'd still prefer to take it out since we don't yet support the join stage. But if you think it seems to match the coming feature, we can probably leave it

Makes sense, I have taken out the languange.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
Linchin: If I remove `# type: ignore`, I get the following error message:
```
pipeline_stages.py:351: error: Need type annotation for "documents"  [var-annotated]
```

If I change the type hint to be the same as init like the following
```
self.documents: tuple[dict[str, Expression | CONSTANT_TYPE], ...] = documents
```
I get these mypy errors:
```
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | str], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | int], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | float], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | bool], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | datetime], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | bytes], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | GeoPoint], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | Vector], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
.nox/mypy-3-12/lib/python3.12/site-packages/google/cloud/firestore_v1/pipeline_stages.py:351: error: Incompatible types in assignment (expression has type "tuple[dict[str, Expression | None], ...]", variable has type "tuple[dict[str, Expression | CONSTANT_TYPE], ...]")  [assignment]
```

It seems mypy couldn't make out that `str`, `int` etc. are subtypes of `CONSTANT_TYPE`. I wonder if there is a more proper way to do type hint here? Otherwise I don't think we lose much coverage with type ignore.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
Linchin: Thanks for catching this, I think `Expression` means the type of value in the dict could be of `Expression` type. I have updated the language to make it more accurate.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
Linchin: Thanks so much, I have removed this part, and also removed the related unit test.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
Linchin: Thanks, I have fixed it.

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_source.py
Linchin: Thanks, I have corrected it.

packages/google-cloud-firestore/tests/unit/v1/test_pipeline_stages.py
Linchin: Thanks, I have corrected this and also fixed other tests where the input is not a dict.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
```diff
@@ -342,6 +343,26 @@ def _pb_args(self):
         return [Value(integer_value=self.limit)]
 
 
+class Literals(Stage):
+    """Returns documents from a fixed set of predefined document objects."""
+
+    def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
```
daniel-sanche: nit: it would probably make sense to make this a dict too, to map the pipeline_source

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
```diff
@@ -342,6 +343,26 @@ def _pb_args(self):
...
+        super().__init__("literals")
+        self.documents: tuple[Mapping[str, Any], ...] = documents
```
daniel-sanche, line 351: I'd prefer if we didn't use Any here, since this is a publicly exposed attribute

We could resolve it using a TypeVar:
`T = TypeVar("T", bound=Union[Expression, CONSTANT_TYPE])`

But honestly, I'd recommend converting the constants at init time instead, which would both resolve this issue, raise conversion errors in a better place, and simplify the code:

```
class Literals(Stage):
    """Returns documents from a fixed set of predefined document objects."""

    def __init__(self, *documents: dict[str, Expression | CONSTANT_TYPE]):
        super().__init__("literals")
        self.documents: Sequence[dict[str, Expression]] = [
            {k: v if isinstance(v, Expression) else encode_value(v) for k,v in d.items()}
            for d in documents
        ]

    def _pb_args(self):
        return [Value(map_value={"fields": doc}) for doc in self.documents]
```

Linchin: I'm not sure if it's a good idea because encoding the values here could break the [`__repr__()` method at the parent class](https://github.com/googleapis/google-cloud-python/blob/1b4467f9f6140cca639a3e8360c06798877f9433/packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py#L153), and we would expose the encoded values in the repr string. For example, instead of `"Literals(documents=('hello',))"`, we would be getting `"Literals(documents=(Value(string_value='hello'),))"`.



 What do you think?

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
Linchin: Thanks, I thought I fixed it. It's corrected now.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_stages.py
daniel-sanche: Sorry, I should have tested the code before posting. Value isn't an expression, so that wouldn't have passed mypy. This should though:

```
class Literals(Stage):
    """Returns documents from a fixed set of predefined document objects."""

    def __init__(self, *documents: Mapping[str, Expression | CONSTANT_TYPE]):
        super().__init__("literals")
        self.documents: list[Mapping[str, Expression]] = [
            {k: v if isinstance(v, Expression) else Constant.of(v) for k,v in d.items()}
            for d in documents
        ]

    def _pb_args(self):
        return [
            Value(map_value={"fields": {k: v.to_pb for k,v in doc.items()}})
            for doc in self.documents
        ]
```


This would result in the repr of `Literals({"hello": "world"}))` being `Literals(documents=[{"hello": Constant.of('world')}])`, which is fine, because represents the same object. And that's how the other stages are set up

PR #16029, opened 2026-03-05
feat(firestore): support array_agg(), array_agg_distinct(), first(), last()

Linchin:
TODO:
 - [x] verify docstring
  - referred to https://github.com/firebase/firebase-js-sdk/pull/9576 for docstring text

### DIFF
--- a/packages/google-cloud-firestore/tests/system/pipeline_e2e/aggregates.yaml
+++ b/packages/google-cloud-firestore/tests/system/pipeline_e2e/aggregates.yaml
@@ -332,10 +332,6 @@ tests:
   - description: testArrayAggDistinct
     pipeline:
       - Collection: books
-      - Where:
-          - FunctionExpression.equal:
-              - Field: genre
-              - Constant: Science Fiction
       - Aggregate:
         - AliasedExpression:
             - FunctionExpression.array_agg_distinct:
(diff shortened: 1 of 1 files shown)


packages/google-cloud-firestore/tests/system/pipeline_e2e/aggregates.yaml
```diff
@@ -281,4 +281,165 @@ tests:
...
+              - Field: genre
+              - Constant: Science Fiction
+      - Aggregate:
+        - AliasedExpression:
+            - FunctionExpression.array_agg_distinct:
+              - Field: genre
+            - "distinct_genres"
+    assert_results:
+      - distinct_genres:
+        - "Science Fiction"
```
daniel-sanche: Is there a better test example that could be used? It looks like this is finding distinct values out of a fixed array with size 1

Could we make it read genres from the database, and return the distinct values or something?

daniel-sanche (commented):
Looks mostly good to me, but consider improving that test

packages/google-cloud-firestore/tests/system/pipeline_e2e/aggregates.yaml
Linchin: This is a really good catch. I removed the filter.

PR #16030, opened 2026-03-05
feat(firestore): support let stage

Linchin:
TODO:
- [ ] verify docstring
- [x] add variable_field_reference

### DIFF
--- a/packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
+++ b/packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
@@ -275,7 +275,7 @@ class _BasePipeline:
             stages.FindNearest(field, vector, distance_measure, options)
         )
 
-    def let(self, **variables: Expression) -> "_BasePipeline":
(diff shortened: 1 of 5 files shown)


packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -275,6 +275,31 @@ def find_nearest(
             stages.FindNearest(field, vector, distance_measure, options)
         )
 
+    def let(self, **variables: Expression) -> "_BasePipeline":
```
wu-hui: Drive-by comment, we are using define instead of let for our SDKs(mobile/web/java/node). There is active discussions happening right now on the naming. You probably need to rename it depending on how the discussion goes.

Linchin: Thanks for the heads up! Will ping you

Linchin:
~~Currently blocked by potential renaming per https://github.com/googleapis/google-cloud-python/pull/16030#discussion_r2920382089~~

We have decided to rename `Let` into `Define`.

PR #16046, opened 2026-03-06
chore(migration): Migrate code from googleapis/python-bigquery-sqlalchemy into packages/sqlalchemy-bigquery

parthea:
See #10990. 


chalmerlowe:
Basic processing and cleanup is complete. Gonna do some focused work on getting presubmits to pass.

packages/sqlalchemy-bigquery/setup.py
```diff
@@ -112,12 +113,12 @@ def readme():
     ],
     platforms="Posix; MacOS X; Windows",
     install_requires=[
-        "google-api-core >= 1.31.5, <3.0.0,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0",
```
ohmayr: Do we still want to support api-core `1.x`? Similar comment for `google-auth`.

parthea: We should do this in a separate PR to ensure that this is part of the release notes

chalmerlowe: I can create a small stand-alone PR to process any significant version changes.

Issue #16035, opened 2026-03-05
requests has not attribute session


sakshamgoyal-01:
### Determine this is the right repository

- [x] I determined this is the correct repository in which to report this bug.

### Summary of the issue

Faulty logic results in a bug. Here is a fix:

`.venv/lib/python3.14/site-packages/google/auth/compute_engine/_metadata.py`
```diff
    if use_mtls:
        # Ensure the request has a session to mount the adapter to.
-         if not request.session:
+        if not hasattr(request, "session") or not request.session:
            request.session = requests.Session()
```

### API client name and version

google-genai==1.66.0

### Reproduction steps: code

```py
from google import genai
from google.genai import types

client = genai.Client(
    vertexai=True,
    project='...",
    location='global',
)
chat = client.chats.create(model='gemini-3-pro-preview')
response = chat.send_message("Hello. What is 2+2?")
```

### Python environment

python 3.14

### Python dependencies
```
[project]
name = "..."
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.14"
dependencies = [
    "google-genai==1.66.0",
    "ipykernel>=7.2.0",
    "jupyter>=1.1.1",
]
```
### Additional context
Error Log:

<details>

```py
File ~/data-eng-des-projects/npc/.venv/lib/python3.14/site-packages/google/genai/chats.py:252, in Chat.send_message(self, message, config)
    247   raise ValueError(
    248       f"Message must be a valid part type: {types.PartUnion} or"
    249       f" {types.PartUnionDict}, got {type(message)}"
    250   )
    251 input_content = t.t_content(message)
--> [252](https://vscode-remote+ssh-002dremote-002bgcp-002dwbi-002dsaksham-002dsdx-002dusc1-002d01.vscode-resource.vscode-cdn.net/home/ext_saksham_goyal_compassdigital/data-eng-des-projects/npc/~/data-eng-des-projects/npc/.venv/lib/python3.14/site-packages/google/genai/chats.py:252) response = self._modules.generate_content(
    253     model=self._model,
    254     contents=self._curated_history + [input_content],  # type: ignore[arg-type]
    255     config=config if config else self._config,
    256 )
    257 model_output = (
    258     [response.candidates[0].content]
    259     if response.candidates and response.candidates[0].content
    260     else []
    261 )
    262 automatic_function_calling_history = (
    263     response.automatic_function_calling_history
    264     if response.automatic_function_calling_history
    265     else []
    266 )

[...]

PR #16050, opened 2026-03-07
fix(auth): fix request session error 

daniel-sanche:
Fixes https://github.com/googleapis/google-cloud-python/issues/16035

only google.auth.transport.**requests**.Request objects have a session attached. Other Request subclasses don't so they raise an error when mts is enabled.

This PR ignores unsupported request types, instead of raising an exception

Thanks @sakshamgoyal-01 for the detailed bug report!

### DIFF
--- a/packages/google-auth/google/auth/compute_engine/_metadata.py
+++ b/packages/google-auth/google/auth/compute_engine/_metadata.py
@@ -166,16 +166,17 @@ def _prepare_request_for_mds(request, use_mtls=False) -> None:
 
     Args:
         request (google.auth.transport.Request): A callable used to make
-            HTTP requests. If mTLS is enabled, the request will have the mTLS
-            adapter mounted. Otherwise, there will be no change.
+            HTTP requests. If mTLS is enabled, and the request supports sessions,
+            the request will have the mTLS adapter mounted. Otherwise, there
+            will be no change.
         use_mtls (bool): Whether to use mTLS for the request.
 
 
     """
-    # Only modify the request if mTLS is enabled.
-    if use_mtls:
+    # Only modify the request if mTLS is enabled, and request supports sessions.
+    if use_mtls and hasattr(request, "session"):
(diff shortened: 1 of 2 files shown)


packages/google-auth/google/auth/compute_engine/_metadata.py
```diff
@@ -177,7 +177,7 @@ def _prepare_request_for_mds(request, use_mtls=False) -> None:
     # Only modify the request if mTLS is enabled.
     if use_mtls:
         # Ensure the request has a session to mount the adapter to.
-        if not request.session:
+        if not getattr(request, "session", None):
             request.session = requests.Session()
```
daniel-sanche, line 180: Thinking about this some more, is this safe? Should we be attaching a session to a request that wasn't designed to hold one? Or should we drop mtls in that situation?

It looks like some tests fail if we drop mtls. But We should confirm the intention before merging this

packages/google-auth/google/auth/compute_engine/_metadata.py
```diff
@@ -166,16 +166,15 @@ def _prepare_request_for_mds(request, use_mtls=False) -> None:
...
-            If mTLS is enabled, the request will have the mTLS adapter mounted.
-            Otherwise, the original request will be returned unchanged.
+
     """
-    # Only modify the request if mTLS is enabled.
-    if use_mtls:
+    # Only modify the request if mTLS is enabled, and request supports sessions.
+    if use_mtls and hasattr(request, "session"):
         # Ensure the request has a session to mount the adapter to.
         if not request.session:
```
sai-sunder-s, line 179: Is this condition required anymore? Since we already check that session is present?

daniel-sanche: Yes, I believe this is meant to attach a session if it's set to None

sai-sunder-s (approved):

PR #16084, opened 2026-03-12
feat: implement updated design for regional access boundary

nbayati:
Original PR: https://github.com/googleapis/google-auth-library-python/pull/1955
Make the fetching async and non blocking.
Implement proactive refresh every 6 hours.
Centralize the logic in a new class.
Remove no-op signal and checks.
Refactor to Regional Access Boundary name.


packages/google-auth/google/auth/credentials.py
```diff
@@ -665,3 +678,8 @@ class TokenState(Enum):
     FRESH = 1
     STALE = 2
     INVALID = 3
+
+
+# For backwards compatibility with the previus feature name.
```
daniel-sanche: nit: typo `previus`

packages/google-auth/google/auth/credentials.py
```diff
@@ -665,3 +678,8 @@ class TokenState(Enum):
...
+# Use CredentialsWithRegionalAccessBoundary instead.
+CredentialsWithTrustBoundary = CredentialsWithRegionalAccessBoundary
```
daniel-sanche: Do you think we can keep the old code around with a deprecation note in the docstring? These changes make me a bit nervous.

I'm not sure if we'd expect anyone to subclass these classes, but if they do, they'd see a crash after this change, because some of the @abc.abstractmethods were changed, for example

If we can get away with implementing CredentialsWithRegionalAccessBoundary as a fresh class and leaving the old one alone, that could be safer

packages/google-auth/google/auth/compute_engine/credentials.py
```diff
@@ -93,7 +94,13 @@ def __init__(
...
             self._universe_domain = universe_domain
             self._universe_domain_cached = True
-        self._trust_boundary = trust_boundary
+
+        if trust_boundary is not None:
+            warnings.warn(
+                "The trust_boundary parameter is deprecated and has no effect.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
daniel-sanche: We should also add this note to the Args: section in the docstrings

packages/google-auth/google/auth/compute_engine/credentials.py
```diff
@@ -146,35 +153,38 @@ def _perform_refresh_token(self, request):
             new_exc = exceptions.RefreshError(caught_exc)
             raise new_exc from caught_exc
 
-    def _build_trust_boundary_lookup_url(self):
-        """Builds and returns the URL for the trust boundary lookup API for GCE."""
+    def _build_regional_access_boundary_lookup_url(self, request=None):
```
daniel-sanche: Can we add typing here? Does this work with [_http_client.Request](https://github.com/nbayati/google-cloud-python/blob/1b570652e9c07366d29b4f3071add5dc95a2c91b/packages/google-auth/google/auth/transport/_http_client.py#L54) type requests? There was a [recent bug around that](https://github.com/googleapis/google-cloud-python/issues/16035). 

Gemini says it should be safe. But do we have tests covering different request types?

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +313,148 @@ def _perform_refresh_token(self, request):
...
             request (google.auth.transport.Request): The object used to make
                 HTTP requests.
 
         Returns:
-            trust_boundary (dict): The trust boundary object returned by the lookup API.
-
-        Raises:
-            google.auth.exceptions.RefreshError: If the trust boundary could not be
-                retrieved.
+            Optional[dict]: The Regional Access Boundary information returned by the lookup API, or None if the lookup failed.
```
daniel-sanche: Is it possible to add typing to the dict? (`Optional[dict[str, str]]`)

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +313,148 @@ def _perform_refresh_token(self, request):
...
-                _LOGGER.debug(
-                    "Using cached trust boundary due to refresh error: %s", error
-                )
-            return
+        self._perform_refresh_token(request)
 
-    def _lookup_trust_boundary(self, request):
-        """Calls the trust boundary lookup API to refresh the trust boundary cache.
+    def _lookup_regional_access_boundary(self, request):
+        """Calls the Regional Access Boundary lookup API to retrieve the Regional Access Boundary information.
```
daniel-sanche, line 440: if possible, it would be good to have type annotations in the signatures too: `_lookup_regional_access_boundary(self, request: google.auth.transport.Request) -> Optional[dict]:`

The codebase doesn't have much of that yet, but if we start adding them to new functions at least, it can help us catch issues down the line

packages/google-auth/google/auth/external_account_authorized_user.py
```diff
@@ -87,7 +90,6 @@ def __init__(
         scopes=None,
         quota_project_id=None,
         universe_domain=credentials.DEFAULT_UNIVERSE_DOMAIN,
-        trust_boundary=None,
```
daniel-sanche, line 90: Shouldn't we keep this around in a deprecated state, like the other classes?

packages/google-auth/google/auth/external_account_authorized_user.py
```diff
@@ -379,12 +390,6 @@ def with_universe_domain(self, universe_domain):
         cred._universe_domain = universe_domain
         return cred
 
-    @_helpers.copy_docstring(credentials.CredentialsWithTrustBoundary)
-    def with_trust_boundary(self, trust_boundary):
```
daniel-sanche, line 383: this seems like a public method we can't safely remove

packages/google-auth/google/auth/impersonated_credentials.py
```diff
@@ -435,15 +443,9 @@ def _make_copy(self):
             lifetime=self._lifetime,
             quota_project_id=self._quota_project_id,
             iam_endpoint_override=self._iam_endpoint_override,
-            trust_boundary=self._trust_boundary,
         )
         cred._cred_file_path = self._cred_file_path
-        return cred
-
-    @_helpers.copy_docstring(credentials.CredentialsWithTrustBoundary)
-    def with_trust_boundary(self, trust_boundary):
```
daniel-sanche, line 444: this looks like another public one we can't drop

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +313,148 @@ def _perform_refresh_token(self, request):
...
 
     def with_trust_boundary(self, trust_boundary):
-        """Returns a copy of these credentials with a modified trust boundary.
+        """Returns a copy of these credentials.
+
+        .. deprecated::
+            Manual Regional Access Boundary overrides are not supported.
+            This method is maintained for backwards compatibility and
+            returns a copy of the credentials without modifying the
+            Regional Access Boundary state.
```
daniel-sanche, line 333: Is this safe to do? I'm not too familiar with the security details here, but I could see it being dangerous to drop security configs like this after updating the library

packages/google-auth/google/auth/credentials.py
```diff
@@ -19,17 +19,16 @@
...
 from google.auth import _helpers, environment_vars
+from google.auth import _regional_access_boundary_utils
 from google.auth import exceptions
 from google.auth import metrics
 from google.auth._credentials_base import _BaseCredentials
 from google.auth._refresh_worker import RefreshThreadManager
 
 DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
-NO_OP_TRUST_BOUNDARY_LOCATIONS: List[str] = []
-NO_OP_TRUST_BOUNDARY_ENCODED_LOCATIONS = "0x0"
```
daniel-sanche, line 32: these seem to be part of the public surface:

```
from google.auth.credentials import NO_OP_TRUST_BOUNDARY_LOCATIONS`
```

packages/google-auth/google/auth/compute_engine/credentials.py
```diff
@@ -146,35 +153,38 @@ def _perform_refresh_token(self, request):
...
+        """Builds and returns the URL for the regional access boundary lookup API for GCE."""
         # If the service account email is 'default', we need to get the
         # actual email address from the metadata server.
         if self._service_account_email == "default":
-            from google.auth.transport import requests as google_auth_requests
+            if request is None:
+                from google.auth.transport import requests as google_auth_requests
 
-            request = google_auth_requests.Request()
+                request = google_auth_requests.Request()
```
daniel-sanche: It looks like `requests` is an optional dependency, so this will crash with an ImportError if it's not installed. 
- Is that safe?
- Do we warn the user anywhere that requests is required for this feature? 
- Are regional boundaries an opt-in feature, or is this auto-enabled in some environments?

daniel-sanche: Can we use _http_client.Request() here, since that doesn't need any extra dependencies?

daniel-sanche (changes requested):

packages/google-auth/google/auth/credentials.py
daniel-sanche: Would it be at all possible to keep the trust boundary logic in place, but just de-emphasize it in the docstrings? So existing code works as before?

packages/google-auth/google/auth/compute_engine/credentials.py
daniel-sanche: I'm considering [adding requests as a required dependency](https://github.com/googleapis/google-cloud-python/pull/16100) too, since it seems to be needed for mtls

packages/google-auth/google/auth/compute_engine/credentials.py
nbayati: Sure, I added the typing and a unit test that ensures it works fine with _http_client.Request().

packages/google-auth/google/auth/credentials.py
nbayati: The trust boundary is currently no-op because the endpoint is returning a hardcoded dummy value, and the feature is behind an env variable. 
The trust boundary (now called regional access boundary) acts as a routing hint for GFE, and doesn't enforce any security itself. If trust boundary is not provided, GFE will route the request to a set of mandatory region, which might be sub-optimal. The project hasn't been officially launched yet, and I'd be shocked if anyone is using this method, or the class in general. I believe GFE will ignore the headers for now, till we officially launch. 

The reason we decided to scope out manual override for now, is that if the value is wrong or outdated, the request might get routed to a region where it doesn't have access, so the request ends up being rejected. Having no header is safer than having the wrong header, because no header means it'll still get routed to an appropriate region. 

We are probably going to re-introduce the manual override in a future release with the new name (regional access boundary), and have it coupled with an auto recovery mechanism that can detect if the request was failed due to stale RAB, drop the header, and retry the request.

packages/google-auth/google/auth/credentials.py
nbayati: Good point, I definitely agree we want to avoid crashing any external subclasses. 
However, fully restoring the original CredentialsWithTrustBoundary class has its own set of issues. The old design had the trust boundary (now regional access boundary) fetching tied to the token refresh, and it used a blocking synchronous call every hour. We were through multiple design updates with the backend team, and the new design is async, non-blocking, and fail-open. It's much safer and seamless for the users. It also fetches the new RAB data every 6 hours as oppose to 1 hour, and has been decoupled from the token refresh flow. We don't expect customers to subclass these, but If we fully restore the old CredentialsWithTrustBoundary class alongside the new one, any customer who happened to subclass it will still be running the old design which is not favorable. 

What if we keep CredentialsWithTrustBoundary and make it inherit from the new CredentialsWithRegionalAccessBoundary. It would look something like this:

```
class CredentialsWithTrustBoundary(CredentialsWithRegionalAccessBoundary):
    """
    .. deprecated::
        This class is deprecated. Please use CredentialsWithRegionalAccessBoundary instead.
    """
    @abc.abstractmethod
    def _build_trust_boundary_lookup_url(self):
        """Deprecated: Implement _build_regional_access_boundary_lookup_url instead."""
        raise NotImplementedError()
    def _build_regional_access_boundary_lookup_url(self, request=None):
        warnings.warn(
            "CredentialsWithTrustBoundary is deprecated. Use CredentialsWithRegionalAccessBoundary.",
            DeprecationWarning,
        )
        return self._build_trust_boundary_lookup_url()
```
 I think this way it satisfies the old abstract method contract, so existing customers' custom subclasses won't crash when they upgrade, without us having to restore the full CredentialsWithTrustBoundary class with the old logic. What do you think?

packages/google-auth/google/auth/credentials.py
nbayati: Good point, I added it back.

packages/google-auth/google/auth/compute_engine/credentials.py
nbayati: The _trust_boundary field was added to the base credential class about 2 years ago, and it had stored a dummy value. It wasn't wired to the trust boundary logic we introduced last year, since the changes we made were limited to CredentialsWithTrustBoundary class, and weren't added to the base class. 
I thought I could clean up the implementation as part of this PR, but I've realized that changing how _trust_boundary was stored, copied, and passed along could potentially break a unit test in one of the other libraries since this field has been around for 2 years. 

I've also decided not to emit a deprecation warning since I'm keeping the behavior the same as before, storing and passing this value without any changes, this way we won't get repetitive deprecation warnings when the credential is copied and re-passed.

packages/google-auth/google/auth/external_account_authorized_user.py
nbayati: It's actually still available! The subclass now inherits with_trust_boundary directly from CredentialsWithRegionalAccessBoundary, so the public API is identical and we won't break any external callers. I just removed the redundant override to clean up the subclass.

packages/google-auth/google/auth/impersonated_credentials.py
nbayati: Same for here, it inherits with_trust_boundary directly from CredentialsWithRegionalAccessBoundary.

packages/google-auth/google/auth/external_account_authorized_user.py
nbayati: Done!

packages/google-auth/google/auth/compute_engine/credentials.py
nbayati: Great suggestion, done!

lsirac:
### Review
- **`CredentialsWithTrustBoundary` compatibility is still not preserved for external subclasses.** Right now `CredentialsWithTrustBoundary` is only an alias to `CredentialsWithRegionalAccessBoundary`. The new base class requires `_perform_refresh_token()` and `_build_regional_access_boundary_lookup_url()`. So any external subclass that previously inherited `CredentialsWithTrustBoundary`, implemented `refresh()`, and implemented `_build_trust_boundary_lookup_url()` can now fail abstract-class instantiation after upgrade.

- **`mock.ANY` is still used in the GCE tests.** `tests/compute_engine/test_credentials.py` still uses `mock.ANY` when asserting the request passed to `get_service_account_info(...)`. We should avoid this where possible. There may be more uses.

- **There is no direct happy-path coverage for the background refresh thread.** The new tests mostly mock `start_refresh` or `_lookup_regional_access_boundary`, but I don't see a test for `_RegionalAccessBoundaryRefreshThread.run()` end-to-end and proves the real thread updates expiry/cooldown/header state correctly.

- **Impersonated credentials ignore `iam_endpoint_override` on the RAB path.** Is this intentional?

### Test Coverage Gaps
- We should add a test that subclasses `CredentialsWithTrustBoundary`, only implements the *old* abstract methods, and still instantiates/works after upgrade.

- Test that `iam_endpoint_override` is honored on the RAB lookup path. I don't see one that checks the RAB lookup URL itself.

- Test for the refresh thread when cached RAB is expired. The thread code is supposed to clear the cached header after hard expiry if refresh fails.

- We should have a test proving async creds do not accidentally trigger sync RAB refresh behavior.

- No test for external-account impersonation delegation behavior. `external_account.py` forwards RAB refresh responsibility to the impersonated credential when impersonation is in play. We need a test proving we only do one RAB lookup through that path instead of duplicating work.

packages/google-auth/google/auth/credentials.py
nbayati: Another option would be keep the old class and make the methods no-op. 
This way if anyone did subclass the CredentialsWithTrustBoundary class, they won't see any abruption to their flows once we remove the feature flag and launch this feature.
```

import warnings

class CredentialsWithTrustBoundary(Credentials):
    """Abstract base for credentials supporting ``with_trust_boundary`` factory

    .. deprecated::
        Use :class:`~google.auth.credentials.CredentialsWithRegionalAccessBoundary` instead.
    """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        warnings.warn(
            "CredentialsWithTrustBoundary is deprecated. Please migrate to "
            "CredentialsWithRegionalAccessBoundary.",
            DeprecationWarning,
            stacklevel=2,
        )

    @abc.abstractmethod
    def _perform_refresh_token(self, request):
        """Refreshes the access token."""
        raise NotImplementedError("_perform_refresh_token must be implemented")

    def with_trust_boundary(self, trust_boundary):
        """Returns a copy of these credentials with a modified trust boundary.

        .. deprecated::
            Trust boundaries are deprecated.
        """
        raise NotImplementedError("This credential does not support trust boundaries.")

    # -------------------------------------------------------------
    # No-Op Legacy Methods to prevent AttributeErrors inside 
    # downstream legacy subclasses
    # -------------------------------------------------------------
    def _is_trust_boundary_lookup_required(self):
        return False

    def _get_trust_boundary_header(self):
        return {}

    def apply(self, headers, token=None):
        super().apply(headers, token)
        # We no longer apply trust boundary headers natively

    def refresh(self, request):
        """Refreshes the access token."""
        self._perform_refresh_token(request)
        self._refresh_trust_boundary(request)

    def _refresh_trust_boundary(self, request):
        """Deprecated: safely does nothing."""
        pass

```

daniel-sanche: Is it safe to make the class into a no-op like Option 2? Wouldn't that put users in a position where they think they are adding security, but it's actually being bypassed? (I'm not super familiar with rab, so maybe I'm thinking of this wrong)

Option 1 sounds good to me, assuming it can be done in a safe way. Layering the logic between the two implementations still seems a little risky to me, but if it's straight forward and we have good test coverage, it sounds great

packages/google-auth/google/auth/compute_engine/credentials.py
```diff
@@ -146,35 +152,40 @@ def _perform_refresh_token(self, request):
...
+        """Builds and returns the URL for the regional access boundary lookup API for GCE."""
         # If the service account email is 'default', we need to get the
         # actual email address from the metadata server.
         if self._service_account_email == "default":
-            from google.auth.transport import requests as google_auth_requests
+            if request is None:
+                from google.auth.transport import _http_client
 
-            request = google_auth_requests.Request()
+                request = _http_client.Request()
```
daniel-sanche: We're running into issues because _http_client doesn't support mtls. Will that be a problem here?

packages/google-auth/google/auth/compute_engine/credentials.py
```diff
@@ -146,35 +152,40 @@ def _perform_refresh_token(self, request):
             new_exc = exceptions.RefreshError(caught_exc)
             raise new_exc from caught_exc
 
-    def _build_trust_boundary_lookup_url(self):
-        """Builds and returns the URL for the trust boundary lookup API for GCE."""
+    def _build_regional_access_boundary_lookup_url(
+        self, request: "Optional[google.auth.transport.Request]" = None  # noqa: F821
+    ):
+        """Builds and returns the URL for the regional access boundary lookup API for GCE."""
```
daniel-sanche: It might be good to document the return type, and mention in what cases it returns None

packages/google-auth/google/auth/credentials.py
daniel-sanche: Ok, makes sense, thanks!

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +324,157 @@ def _perform_refresh_token(self, request):
...
 
-    def _is_trust_boundary_lookup_required(self):
-        """Checks if a trust boundary lookup is required.
+        warnings.warn(
+            "with_trust_boundary is deprecated and has no effect.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        if hasattr(self, "_make_copy"):
+            return self._make_copy()  # type: ignore
```
daniel-sanche: nit: is the `type: ignore` needed here? It would be ideal if we could avoid this as much as possible

packages/google-auth/google/auth/external_account.py
```diff
@@ -497,8 +523,7 @@ def _build_trust_boundary_lookup_url(self):
         )
         if workload_match:
             project_number, pool_id = workload_match.groups()
-            url = _constants._WORKLOAD_IDENTITY_POOL_TRUST_BOUNDARY_LOOKUP_ENDPOINT.format(
-                universe_domain=self._universe_domain,
+            url = _constants._WORKLOAD_IDENTITY_POOL_REGIONAL_ACCESS_BOUNDARY_LOOKUP_ENDPOINT.format(
```
daniel-sanche: Does RAB not support TPC? Should there be an error if an unexpected universe domain is set? (I have a pretty high-level understanding of both of these features, so maybe dumb question)

daniel-sanche (commented):
few small comments and questions, but overall looks good

packages/google-auth/google/auth/compute_engine/credentials.py
nbayati: Good catch! Yeah this would be a problem here. To address both your original concern (optional dependencies) and your follow-up (mTLS support), I’ve applied a `try / except ImportError` pattern to both `_build_regional_access_boundary_lookup_url` and `universe_domain` properties so we can try to import and use `requests` which for those who are using mTLS must be installed anyway, and fallback to _http_client if it's not installed.

packages/google-auth/google/auth/credentials.py
nbayati: Option 1 is done! I think the layering logic is minimal and I've added unit tests, and I believe it should be safe, but please let me know if you have any concerns.

packages/google-auth/google/auth/external_account.py
nbayati: No TPC is not supported, and we skip lookup altogether if we detect a non-GDU universe domain. The logic is in google/auth/credentials.py: 

```
    def _is_regional_access_boundary_lookup_required(self):
        """Checks if a Regional Access Boundary lookup is required.

        A lookup is required if the feature is enabled via an environment
        variable and the universe domain is supported.

        Returns:
            bool: True if a Regional Access Boundary lookup is required, False otherwise.
        """
        # 1. Check if the feature is enabled.
        if not _regional_access_boundary_utils.is_regional_access_boundary_enabled():
            return False

        # 2. Skip for non-default universe domains.
        if self.universe_domain != DEFAULT_UNIVERSE_DOMAIN:
            return False

        return True
  ```

**Comment by nbayati** [comment]
> ### Review
> * **`CredentialsWithTrustBoundary` compatibility is still not preserved for external subclasses.** Right now `CredentialsWithTrustBoundary` is only an alias to `CredentialsWithRegionalAccessBoundary`. The new base class requires `_perform_refresh_token()` and `_build_regional_access_boundary_lookup_url()`. So any external subclass that previously inherited `CredentialsWithTrustBoundary`, implemented `refresh()`, and implemented `_build_trust_boundary_lookup_url()` can now fail abstract-class instantiation after upgrade.
> * **`mock.ANY` is still used in the GCE tests.** `tests/compute_engine/test_credentials.py` still uses `mock.ANY` when asserting the request passed to `get_service_account_info(...)`. We should avoid this where possible. There may be more uses.
> * **There is no direct happy-path coverage for the background refresh thread.** The new tests mostly mock `start_refresh` or `_lookup_regional_access_boundary`, but I don't see a test for `_RegionalAccessBoundaryRefreshThread.run()` end-to-end and proves the real thread updates expiry/cooldown/header state correctly.
> * **Impersonated credentials ignore `iam_endpoint_override` on the RAB path.** Is this intentional?
> 
> ### Test Coverage Gaps
> * We should add a test that subclasses `CredentialsWithTrustBoundary`, only implements the _old_ abstract methods, and still instantiates/works after upgrade.
> * Test that `iam_endpoint_override` is honored on the RAB lookup path. I don't see one that checks the RAB lookup URL itself.
> * Test for the refresh thread when cached RAB is expired. The thread code is supposed to clear the cached header after hard expiry if refresh fails.
> * We should have a test proving async creds do not accidentally trigger sync RAB refresh behavior.
> * No test for external-account impersonation delegation behavior. `external_account.py` forwards RAB refresh responsibility to the impersonated credential when impersonation is in play. We need a test proving we only do one RAB lookup through that path instead of duplicating work.

Thanks for the review feedback @lsirac
Test coverage and mock.ANY in tests are fixed. I've added back `CredentialsWithTrustBoundary` as a bridge class so the backward compatibility should be resolved too. 

Re `iam_endpoint_override`, it comes from the `service_account_impersonation_url` field from the JSON file of the external account, and it's the endpoint we should call to get a token, not necessarily an override for all iam endpoints. However, I realized that we weren't respecting mtls vs non-mtls iam endpoint, so I refactored the code to use the existing iam module to determine the right endpoint!

File: packages/google-auth/google/auth/compute_engine/credentials.py (hunk 13 above)
Comment (daniel-sanche) [reply packages/google-auth/google/auth/compute_engine/credentials.py]: > requests ... for those who are using mTLS must be installed anyway

Where is this enforced? In https://github.com/googleapis/google-cloud-python/issues/16090#issue-4065664022, we're running into an issue where that's not the case

This question is probably outside the scope of this change, but I just want to understand it better

**Review by daniel-sanche — APPROVED** [review]
LGTM

File: packages/google-auth/google/auth/_regional_access_boundary_utils.py [hunk 17]
```diff
@@ -0,0 +1,263 @@
...
+
+
+class _RegionalAccessBoundaryRefreshThread(threading.Thread):
+    """Thread for background refreshing of the Regional Access Boundary."""
+
+    def __init__(self, credentials, request, rab_manager):
+        super().__init__()
+        self.daemon = True
+        self._credentials = credentials
+        self._request = request
```
lsirac, line 164: Should we be reusing the caller's request transport here?

packages/google-auth/google/auth/credentials.py
```diff
@@ -290,8 +300,14 @@ def with_universe_domain(self, universe_domain):
...
 
 
-class CredentialsWithTrustBoundary(Credentials):
-    """Abstract base for credentials supporting ``with_trust_boundary`` factory"""
+class CredentialsWithRegionalAccessBoundary(Credentials):
+    """Abstract base for credentials supporting regional access boundary configuration."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._rab_manager = (
```
lsirac, line 308: By adding a rab manager to every cred instance which contains a lock that can't be serialized I believe we are breaking serialization

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +324,158 @@ def _perform_refresh_token(self, request):
...
+        )
+        make_copy = getattr(self, "_make_copy", None)
+        if make_copy:
+            return make_copy()
+        else:
+            raise NotImplementedError(
+                "This credential does not support trust boundaries."
+            )
+
+    def _copy_regional_access_boundary_manager(self, target):
```
lsirac, line 356: Is sharing manager across copies the right behavior?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -0,0 +1,263 @@
...
+        """Starts a background thread to refresh the Regional Access Boundary if needed.
+
+        Args:
+            credentials (google.auth.credentials.Credentials): The credentials to refresh.
+            request (google.auth.transport.Request): The object used to make HTTP requests.
+        """
+        rab_data = self._data
+
+        # Don't start a new refresh if the Regional Access Boundary info is still fresh.
+        if (
```
lsirac, line 141: I think in some places empty counts as cached, and others empty means missing. What is the intended behavior, and lets make sure we're consistent

lsirac (changes requested):

parthea:
@nbayati PTAL

nbayati:
> @nbayati PTAL

Thank you for the ping. I'm actively working on addressing the remaining comments, and I'll have the changes ready for review next week.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: That's a very good point, thanks for raising this! I did some investigation and it seems like if the caller is using the standard http transport, it is not guaranteed to be thread safe and we might get into some issues if we're directly using it for the background RAB fetching. I followed what the token refresh worker does (in _refresh_worker.py) which is to make a deepcopy of the transport. The only edge case is if a user provides a custom transport that cannot be deep-copied. In that case, I added a try/except block to log a warning and skip the background refresh for that cycle. I decided to skip rather than falling back to a fresh transport or reusing the original one to avoid risking log noise or thread-safety issues.

packages/google-auth/google/auth/credentials.py
nbayati: You're completely right. I've added __getstate__ and __setstate__ to make these classes serializable, thanks!

packages/google-auth/google/auth/credentials.py
nbayati: I had added sharing managers as an optimization, but during a review of the gcloud proposal I realized that this could cause unexpected issues since the expectation would be that _make_copy() would make a full new copy and the new object wouldn't share some states and resources with the original object. I reverted this behavior and now we create new manager for each clone. To avoid a cold start, the new manager inherits a reference to the same immutable _RegionalAccessBoundaryData object as the source, but maintains its own independent locks and background threads.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: You're right, in applying header we only checked that the encoded location was not None. With the recent developments from the backend team, we know that the encoded location will never be empty, as it'd always have at least the mandatory regions. I've updated the apply_headers method to use truthiness condition instead of only checking that it's not None.

lsirac:
since _perform_refresh_token is required, and CredentialsWithTrustBoundary doesn't include it, it remains an abstract class and will fail to instantiate with a TypeError after upgrade

lsirac (approved):

nbayati:
> since _perform_refresh_token is required, and CredentialsWithTrustBoundary doesn't include it, it remains an abstract class and will fail to instantiate with a TypeError after upgrade

The CredentialsWithTrustBoundary class that's in the [main branch](https://github.com/nbayati/google-cloud-python/blob/2f792abd5d48ec680305e1e4ec9136360e16c9a5/packages/google-auth/google/auth/credentials.py#L293) right now, defines _perform_refresh_token as an abstract method. _perform_refresh_token is not a new method being introduced in CredentialsWithRegionalAccessBoundary.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Args:
+            headers (MutableMapping[str, str]): The headers dictionary to update.
+        """
+        rab_data = self._data
+
+        if rab_data.encoded_locations and (
+            rab_data.expiry is not None and _helpers.utcnow() < rab_data.expiry
+        ):
+            headers[_REGIONAL_ACCESS_BOUNDARY_HEADER] = rab_data.encoded_locations
```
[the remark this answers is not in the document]
nbayati: This is intentional. The backend design has changed and we no longer will receive a no-op signal.

packages/google-auth/google/auth/credentials.py
```diff
@@ -308,140 +324,162 @@ def _perform_refresh_token(self, request):
...
+        """Copies the regional access boundary manager to another instance."""
+        # Create a new manager for the clone to isolate background refresh locks and threads,
+        # but share the immutable data reference to avoid unnecessary initial lookups.
+        new_manager = _regional_access_boundary_utils._RegionalAccessBoundaryManager()
+        new_manager._data = self._rab_manager._data
+        target._rab_manager = new_manager
```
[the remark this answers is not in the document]
nbayati: The change to isolate managers was intentional. Updating the documentation of _RegionalAccessBoundaryManager class to match the current implementation.

PR #16091, opened 2026-03-12
google-auth: use auth.transport.requests as new default

### LINKED ISSUE #16090: google.auth.default() fails to retreive project_id

### Summary of the issue

**Context**
google.auth.default() will use http.client as default transport, but mTLS support needs session support to work

**Expected Behavior:**
return proper project_id with default transport

**Actual Behavior:**
returned project_id is None

google-auth        2.49.0

### Reproduction steps: code

file: main.py
```python
import google.auth
credentials, project_id = google.auth.default()
assert(project_id)
```

Worked in 2.43.0 and broke with 2.44.0 and later due mTLS support, which has a hard requirement for requests.

Workaround:
file: main.py
```python
import google.auth
from google.auth.transport.requests import Request
credentials, project_id = google.auth.default(request=Request())
assert(project_id)
```

_No response_

_No response_

_No response_

### OS & version + platform

Debian 12

### Python environment

3.11.2

### Python dependencies

_No response_

### Additional context

_No response_

uweber:
mTLS support was recently added and added a hard requirement for requests. google.auth.compute_engine would fail to load, if requests is missing, so its safe to switch to google.auth.transport.requests as new default here.

Therefore switch default from google.auth.transport._http_client.Request() to google.auth.transport.requests.Request() which is required for mTLS to work.

Broken since google-auth 2.44.0 and later.

Fixes the following code:
```
  import google.auth
  credentials, project_id = google.auth.default()
  assert(project_id)
```

Fixes #16090

### DIFF
--- a/packages/google-auth/google/auth/_default.py
+++ b/packages/google-auth/google/auth/_default.py
@@ -27,7 +27,6 @@ import warnings
 
 from google.auth import environment_vars
 from google.auth import exceptions
-import google.auth.transport._http_client
 
 if TYPE_CHECKING:  # pragma: NO COVER
     from google.auth.credentials import Credentials  # noqa: F401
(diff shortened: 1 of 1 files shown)


daniel-sanche:
Sorry, looks like I duplicated some of this work in https://github.com/googleapis/google-cloud-python/pull/16480. I'm going to close this in favor of the new one, since I also added some additional tests there. But I appreciate you putting this together

PR #16095, opened 2026-03-12
[DRAFT] fix(auth): disable mds mtls if requests isn't present

daniel-sanche:
Potential fix for https://github.com/googleapis/google-cloud-python/issues/16090. Alternate fix: https://github.com/googleapis/google-cloud-python/pull/16100

Changes mtls enablement logic to avoid mtls if the requests library is missing, or raise an exception in strict mode

Another option would be to make requests into a hard dependency


daniel-sanche:
Closing in favor of https://github.com/googleapis/google-cloud-python/pull/16480

PR #16115, opened 2026-03-17
chore: improve typing in api_core.exceptions

daniel-sanche:
GoogleApiCallError.grpc_status_code had no typing, so mypy will assume the value is always None

This PR adds the proper type annotation


gkevinzheng (approved):
LGTM

PR #16126, opened 2026-03-18
feat: generate libraries

gkevinzheng:
PiperOrigin-RevId: 879267201


.librarian/state.yaml
```diff
@@ -3719,6 +3719,24 @@ libraries:
...
+    apis:
+      - path: google/maps/navconnect/v1
+        service_config: navigationconnect_v1.yaml
+    source_roots:
+      - packages/google-maps-navconnect
+    preserve_regex:
+      - packages/google-maps-navconnect/CHANGELOG.md
+      - docs/CHANGELOG.md
+      - samples/README.txt
+      - scripts/client-post-processing
```
ohmayr, line 3734: Do we need these two lines?

.librarian/state.yaml
```diff
@@ -4118,6 +4136,8 @@ libraries:
     apis: []
     source_roots:
       - packages/sqlalchemy-bigquery
+    preserve_regex: []
+    remove_regex: []
```
ohmayr, line 4140: Weird that `tag_format` is missing for this library.

PR #16129, opened 2026-03-19
fix(google-auth): add AWS Fargate metadata endpoint support

ggrossman:
## Summary

On AWS Fargate, the IMDS security credentials endpoint does **not** use role names in the URL structure (unlike EC2 instances). The current `_DefaultAwsSecurityCredentialsSupplier` unconditionally fetches the IAM role name and appends it to the credentials URL, which fails on Fargate with a `RefreshError`.

This PR:
- Adds `_is_fargate_environment()` to detect Fargate via ECS-specific environment variables (`ECS_CONTAINER_METADATA_URI_V4`, `ECS_CONTAINER_METADATA_URI`, or `AWS_EXECUTION_ENV` containing `AWS_ECS_FARGATE`)
- Skips the role name lookup on Fargate and calls the security credentials URL directly
- Preserves the existing EC2 behavior (3 metadata requests: region, role, credentials) when not on Fargate
- Fargate path makes only 2 metadata requests: region + credentials

### Behavior comparison

| Environment | Metadata requests | Credentials URL |
|---|---|---|
| EC2 (unchanged) | region → role name → credentials | `{url}/{role_name}` |
| Fargate (new) | region → credentials | `{url}` (no role suffix) |

### Files changed

- `packages/google-auth/google/auth/aws.py` — Fargate detection + conditional role name handling
- `packages/google-auth/tests/test_aws.py` — 5 new test cases covering all detection methods, EC2 regression, and unit tests for `_is_fargate_environment()`

Fixes https://github.com/googleapis/google-auth-library-python/issues/1099

### DIFF
--- a/packages/google-auth/google/auth/aws.py
+++ b/packages/google-auth/google/auth/aws.py
@@ -434,7 +434,13 @@ class _DefaultAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
             )
 
         imdsv2_session_token = self._get_imdsv2_session_token(request)
-        role_name = self._get_metadata_role_name(request, imdsv2_session_token)
+
+        # In Fargate environments, the metadata endpoint doesn't use role
+        # names in the URL structure, so we skip the role name lookup.
+        if self._is_fargate_environment():
+            role_name = None
+        else:
+            role_name = self._get_metadata_role_name(request, imdsv2_session_token)
 
         # Get security credentials.
         credentials = self._get_metadata_security_credentials(
@@ -535,8 +541,15 @@ class _DefaultAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
         else:
             headers = None
 
+        # In Fargate environments, use the security credentials URL directly
+        # without appending the role name.
+        if role_name is not None:
+            url = "{}/{}".format(self._security_credentials_url, role_name)
+        else:
+            url = self._security_credentials_url
+
         response = request(
-            url="{}/{}".format(self._security_credentials_url, role_name),
+            url=url,
             method="GET",
             headers=headers,
         )
@@ -602,6 +615,25 @@ class _DefaultAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
             )
 
         return response_body
+
+    @staticmethod
+    def _is_fargate_environment():
+        """Checks if the current environment is an AWS Fargate container.
+
+        Fargate containers expose credentials through a different metadata
+        endpoint structure that doesn't use role names in the URL. This is
+        detected via ECS-specific environment variables.
+
+        Returns:
+            bool: True if running in a Fargate environment.
+        """
+        if os.environ.get("ECS_CONTAINER_METADATA_URI_V4"):
+            return True
+        if os.environ.get("ECS_CONTAINER_METADATA_URI"):
+            return True
+        if "AWS_ECS_FARGATE" in os.environ.get("AWS_EXECUTION_ENV", ""):
+            return True
+        return False
 
 
 class Credentials(external_account.Credentials):
--- a/packages/google-auth/google/auth/aws.py
+++ b/packages/google-auth/google/auth/aws.py
@@ -627,13 +627,11 @@ class _DefaultAwsSecurityCredentialsSupplier(AwsSecurityCredentialsSupplier):
         Returns:
             bool: True if running in a Fargate environment.
         """
-        if os.environ.get("ECS_CONTAINER_METADATA_URI_V4"):
-            return True
-        if os.environ.get("ECS_CONTAINER_METADATA_URI"):
-            return True
-        if "AWS_ECS_FARGATE" in os.environ.get("AWS_EXECUTION_ENV", ""):
-            return True
-        return False
+        return bool(
+            os.environ.get("ECS_CONTAINER_METADATA_URI_V4")
+            or os.environ.get("ECS_CONTAINER_METADATA_URI")
+            or "AWS_ECS_FARGATE" in os.environ.get("AWS_EXECUTION_ENV", "")
+        )
 
 
 class Credentials(external_account.Credentials):
--- a/packages/google-auth/tests/test_aws.py
+++ b/packages/google-auth/tests/test_aws.py
@@ -2456,3 +2456,161 @@ class TestCredentials(object):
         assert credentials.quota_project_id == QUOTA_PROJECT_ID
         assert credentials.scopes == SCOPES
         assert credentials.default_scopes == ["ignored"]
+
+    @mock.patch("google.auth._helpers.utcnow")
+    @mock.patch.dict(
+        os.environ, {"ECS_CONTAINER_METADATA_URI_V4": "http://169.254.170.2/v4"}
+    )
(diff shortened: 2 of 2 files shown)


packages/google-auth/google/auth/aws.py
```diff
@@ -603,6 +616,23 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+        """Checks if the current environment is an AWS Fargate container.
+
+        Fargate containers expose credentials through a different metadata
+        endpoint structure that doesn't use role names in the URL. This is
+        detected via ECS-specific environment variables.
+
+        Returns:
+            bool: True if running in a Fargate environment.
+        """
+        return bool(
```
ggrossman, line 630: P1: `ECS_CONTAINER_METADATA_URI` and `_V4` exist for ECS tasks using the EC2 launch type too; they do not establish Fargate. Those tasks will now skip the required IMDS role-name lookup and call the bare security-credentials URL, so previously working ECS-on-EC2 authentication fails. Base this decision on the actual container-credentials endpoint/launch type, not the generic ECS task-metadata variables.

ggrossman (commented):
Adversarial review found 1 new actionable issue in the current head. Inline details include severity and a concrete failure scenario.

packages/google-auth/google/auth/aws.py
```diff
@@ -535,8 +541,15 @@ def _get_metadata_security_credentials(
         else:
             headers = None
 
+        # In Fargate environments, use the security credentials URL directly
+        # without appending the role name.
+        if role_name is not None:
+            url = "{}/{}".format(self._security_credentials_url, role_name)
+        else:
+            url = self._security_credentials_url
```
ggrossman, line 549: **Regression: the Fargate branch bypasses the missing-`url` guard, so a non-`RefreshError` escapes `refresh()`.**

The only validation that `credential_source["url"]` was actually configured lives inside `_get_metadata_role_name`:

```python
if self._security_credentials_url is None:
    raise exceptions.RefreshError(
        "Unable to determine the AWS metadata server security credentials endpoint"
    )
```

On the Fargate path that method is never called, so `self._security_credentials_url` is no longer checked. This line then assigns `url = None` and passes it straight to the transport.

Reproduction — a credential source with no `"url"` key (exactly the one built by the existing `test_retrieve_subject_token_error_determining_security_creds_url`), with `ECS_CONTAINER_METADATA_URI_V4` set:

```
  -> transport called with url=None method='GET'
RAISED: requests.exceptions.MissingSchema -> Invalid URL 'None': No scheme supplied.
```

With the same variable unset, the pre-existing behaviour is preserved:

```
RAISED: google.auth.exceptions.RefreshError -> Unable to determine the AWS metadata server security credentials endpoint
```

So callers that today catch `google.auth.exceptions.RefreshError` around `refresh()` will instead see a transport-specific exception (`requests.exceptions.MissingSchema` for `google.auth.transport.requests.Request`), and the diagnostic message pointing at the misconfigured credential source is lost. This is also why the existing `test_retrieve_subject_token_error_determining_security_creds_url` fails when any of the three detection variables is present in the runner's environment.

Suggestion: hoist the guard so it runs on both paths, e.g. at the top of `_get_metadata_security_credentials`, or in `get_aws_security_credentials` before the branch. While here, the `role_name` docstring just above still reads `role_name (str): The AWS role name required by the AWS metadata server ...`; it is now `Optional[str]` and `None` means "do not append a role segment".

packages/google-auth/google/auth/aws.py
```diff
@@ -603,6 +616,23 @@ def _get_metadata_role_name(self, request, imdsv2_session_token):
...
+
+        Fargate containers expose credentials through a different metadata
+        endpoint structure that doesn't use role names in the URL. This is
+        detected via ECS-specific environment variables.
+
+        Returns:
+            bool: True if running in a Fargate environment.
+        """
+        return bool(
+            os.environ.get("ECS_CONTAINER_METADATA_URI_V4")
```
ggrossman, line 631: **Nit (consistency): these variable names would be more at home in `google/auth/environment_vars.py`.**

Every other environment variable this module reads goes through that module — `environment_vars.AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`, `AWS_DEFAULT_REGION`, all grouped under the "AWS environment variables used with AWS workload identity pools" comment block. Defining `ECS_CONTAINER_METADATA_URI_V4`, `ECS_CONTAINER_METADATA_URI` and `AWS_EXECUTION_ENV` alongside them and referencing `environment_vars.<NAME>` here would keep the single inventory of environment inputs intact and make them greppable with the rest.

packages/google-auth/tests/test_aws.py
```diff
@@ -2456,3 +2456,161 @@ def test_refresh_success_with_supplier(self, utcnow, mock_auth_lib_value):
...
+        subject_token = credentials.retrieve_subject_token(request)
+
+        assert subject_token == self.make_serialized_aws_signed_request(
+            aws.AwsSecurityCredentials(ACCESS_KEY_ID, SECRET_ACCESS_KEY, TOKEN)
+        )
+        assert len(request.call_args_list) == 2
+
+    @mock.patch("google.auth._helpers.utcnow")
+    @mock.patch.dict(os.environ, {}, clear=False)
+    def test_retrieve_subject_token_non_fargate_still_uses_role_name(self, utcnow):
```
ggrossman, line 2545: **Test isolation: the new detection makes this file's existing tests depend on the ambient environment.**

Because `_is_fargate_environment()` reads `os.environ` directly, every existing test that exercises the EC2 role-name path now silently changes meaning depending on where the suite runs. Exporting a single variable before running this file:

```
ECS_CONTAINER_METADATA_URI=http://169.254.170.2/v3/abc pytest tests/test_aws.py
```

makes 14 previously-passing tests in this file fail, among them:

- `test_retrieve_subject_token_success_permanent_creds_no_environment_vars`
- `test_retrieve_subject_token_success_temp_creds_no_environment_vars[_idmsv2]`
- `test_retrieve_subject_token_success_ipv6`
- `test_retrieve_subject_token_error_determining_aws_role`
- `test_retrieve_subject_token_error_determining_security_creds_url`
- `test_retrieve_subject_token_error_determining_aws_security_creds`
- the four `test_refresh_success_{with,without}_impersonation_*` tests

That is a realistic configuration for anyone running the unit suite inside a container task. This test is the only one in the file that defends against it, and it does so ad hoc.

Suggestion: add an autouse fixture on `TestCredentials` that removes the three variables for every test, so the EC2 assertions stay deterministic and the Fargate tests opt in explicitly via `mock.patch.dict`. It would also be worth adding a Fargate-environment variant of `test_retrieve_subject_token_error_determining_security_creds_url`, which is what surfaces the missing-`url` regression noted on `aws.py`.

ggrossman (commented):
Second-pass review focused on issues not already covered in this thread. Three additional items, one of them a behaviour regression. (The launch-type detection concern already raised on `_is_fargate_environment` is not repeated here.)

Findings verified against `9f44737` by running `tests/test_aws.py` from this branch against the patched `google/auth/aws.py`.

PR #16137, opened 2026-03-20
feat: Add python client api for JobStatistics.referenced_property_graphs.

ericfe-google:

### DIFF
--- a/packages/google-cloud-bigquery/tests/unit/job/test_async_job_retry.py
+++ b/packages/google-cloud-bigquery/tests/unit/job/test_async_job_retry.py
@@ -26,8 +26,6 @@ JOB_ID = "test-job-id"
 
 
 def test_cancel_w_custom_retry(global_time_lock):
-    from google.cloud.bigquery.retry import DEFAULT_RETRY
-
     api_path = "/projects/{}/jobs/{}/cancel".format(PROJECT, JOB_ID)
     resource = {
         "jobReference": {


packages/google-cloud-bigquery/google/cloud/bigquery/job/query.py
```diff
@@ -1332,6 +1332,35 @@ def referenced_tables(self):
...
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobStatistics2.FIELDS.referenced_property_graphs
+
+        Returns:
+            List[google.cloud.bigquery.table.PropertyGraphReference]:
+                mappings describing the property graphs, or an empty list
+                if the query has not yet completed.
+        """
+        property_graphs = []
+        datasets_by_project_name = {}
```
tswast: DatasetReference should be a pretty cheap object to construct. This cache of dataset objects smells like premature optimization to me. The less code we need to map from API response to Python objects the better as far as these handwritten client libraries are concerned.

Consider avoiding DatasetReference altogether in the new class. I consider it a mistake to couple those for table references.

packages/google-cloud-bigquery/google/cloud/bigquery/table.py
```diff
@@ -359,6 +359,131 @@ def __repr__(self):
...
+
+    @property
+    def property_graph_id(self) -> str:
+        """str: The property graph ID."""
+        return _helpers._get_sub_prop(
+            self._properties, self._PROPERTY_TO_API_FIELD["property_graph_id"]
+        )
+
+    @classmethod
+    def from_string(
```
tswast: Unless we need this for some use case, I would prefer to omit it until we get a request for it.

ericfe-google: This was generated by Gemini, and seems like a case of the AI being a bit overzealous. Removed, as it is not needed.

packages/google-cloud-bigquery/google/cloud/bigquery/job/query.py
ericfe-google: Done. Agree, this code does seem overly complicated. Probably caused by Gemini being overzealous when it generated this PR. Simplified.

PR #16152, opened 2026-03-24
feat(firestore): support string expressions

Linchin:
* regex_find()
* regex_find_all()
* split()
* string_repeat()
* string_replace_all()
* string_replace_one()
* string_index_of()

### DIFF
--- a/packages/google-cloud-firestore/tests/system/pipeline_e2e/string.yaml
+++ b/packages/google-cloud-firestore/tests/system/pipeline_e2e/string.yaml
@@ -663,10 +663,10 @@ tests:
         - AliasedExpression:
             - FunctionExpression.regex_find:
                 - Field: title
-                - Constant: "Hitchhiker's"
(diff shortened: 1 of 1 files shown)


packages/google-cloud-firestore/tests/system/pipeline_e2e/string.yaml
```diff
@@ -652,3 +652,599 @@ tests:
...
+      - Collection: books
+      - Where:
+          - FunctionExpression.equal:
+              - Field: author
+              - Constant: "Douglas Adams"
+      - Select:
+        - AliasedExpression:
+            - FunctionExpression.regex_find:
+                - Field: title
+                - Constant: "Hitchhiker's"
```
daniel-sanche: Maybe we should make the regex statement more interesting?

Linchin: Good point, I have updated it.

PR #16160, opened 2026-03-25
chore(migration): Migrate code from googleapis/sphinx-docfx-yaml into packages/gcp-sphinx-docfx-yaml

ohmayr:
See #15998. 

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -24,6 +24,9 @@ libraries:
     apis: []
     source_roots:
       - packages/django-google-spanner
+    preserve_regex: []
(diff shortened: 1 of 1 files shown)


.librarian/state.yaml
```diff
@@ -24,6 +24,12 @@ libraries:
     apis: []
     source_roots:
       - packages/django-google-spanner
```
parthea, line 26: The configuration for `django-google-spanner` is missing `preserve_regex`, `remove_regex` and `tag_format`

ohmayr: addressed

PR #16470, opened 2026-03-27
feat(firestore): pipeline subequeries

daniel-sanche:
Add support for Pipeline subqueries, allowing users to perform complex data transformation by embedding a full pipeline inside another

New Stages:
- `define`
- `subcollection`

New Expressions:
- `current_document`
- `variable`
- `get_field`

New classes:
- SubPipeline, which represents a pipeline without an associated client, which can not be executed


packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
```diff
@@ -610,3 +668,53 @@ def distinct(self, *fields: str | Selectable) -> "_BasePipeline":
...
+        Args:
+            *aliased_expressions: One or more :class:`AliasedExpression` defining the Variable names and values.
+
+        Returns:
+            A new Pipeline object with this stage appended to the stage list.
+        """
+        return self._append(stages.Define(*aliased_expressions))
+
+
+class SubPipeline(_BasePipeline):
```
bhshkh, line 697: Is this being named Subpipeline to avoid name collision ? Is there an existing Subcollection class ?

Java calls this stage Subcollection https://github.com/googleapis/java-firestore/blob/0c8188520dfbada0d3fef0719e4f95fc231306be/google-cloud-firestore/src/main/java/com/google/cloud/firestore/PipelineSource.java#L214-L238

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
```diff
@@ -168,6 +168,31 @@ def literals(
             *documents: One or more documents to be returned by this stage. Each can be a `dict`
                        of values of `Expression` or `CONSTANT_TYPE` types.
         Returns:
-            A new Pipeline object with this stage appended to the stage list.
+            A new pipeline instance targeting the specified literal documents
         """
         return self._create_pipeline(stages.Literals(*documents))
+
+    def subcollection(self, path: str) -> SubPipeline:
```
bhshkh: In java, this is static https://github.com/googleapis/java-firestore/blob/0c8188520dfbada0d3fef0719e4f95fc231306be/google-cloud-firestore/src/main/java/com/google/cloud/firestore/PipelineSource.java#L214-L238

Not sure what would be the equivalent in Python

packages/google-cloud-firestore/google/cloud/firestore_v1/base_pipeline.py
daniel-sanche: This isn't a stage, it's a pipeline subclass. I think this is probably Python-specific, because of the weird way we have to handle async/sync code

In Python, we have 3 classes: 
- `Pipeline` to handle sync clients
- `AsyncPipeline` to handle async clients
-  `_BasePipeline` as a private class to handle common logic between the two. 

Sub-pipelines presents a new use-case, where neither client is be attached, but it needs to be exposed to end-users

In Java and other implementations, it's much simpler, because there is a single pipeline class, and they just [raise an exception](https://github.com/googleapis/java-firestore/blob/0c8188520dfbada0d3fef0719e4f95fc231306be/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Pipeline.java#L1495) if you try to use it to execute on a sub-pipeline.

packages/google-cloud-firestore/google/cloud/firestore_v1/pipeline_source.py
daniel-sanche: Good catch, I'll fix that

PR #16476, opened 2026-03-30
feat(pandas-gbq): drop support for Python 3.9

tswast:

Fixes internal b/490130968 🦕

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,4 +12,10 @@
 *                                             @googleapis/cloud-sdk-python-team @googleapis/cloud-sdk-librarian-team
 
 /packages/bigframes/                          @googleapis/cloud-sdk-python-team @googleapis/cloud-sdk-librarian-team @googleapis/bigquery-dataframe-team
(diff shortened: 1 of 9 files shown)


tswast:
Oh, I see that 3.9 tests are still present. Perhaps this needs to change repo-wide?

daniel-sanche:
We are in the process of moving off 3.9 across the repo, but probably won't finish it until Next

It looks like 3.9 is still [in the noxfile here](https://github.com/googleapis/google-cloud-python/blob/8bd2194010a0079f9413f591be5ec7ab626b1fc3/packages/pandas-gbq/noxfile.py#L37) though? You could try removing it, or adding a `session.skip` if 3.9 is triggered (but if you keep it, you'd have to [change this line](https://github.com/googleapis/google-cloud-python/blob/8bd2194010a0079f9413f591be5ec7ab626b1fc3/packages/pandas-gbq/noxfile.py#L363) in the prerelease_deps too, to point to 3.10)

tswast:
@daniel-sanche I believe this is ready for another look.

PR #16477, opened 2026-03-30
Change list_secrets to use async await

Lucioric2000:
The list_secrets call returns a coroutine that should be awaited, therefore, I am introducing the await instruction, which was absent, to the async secrets getter example.


daniel-sanche:
Hey, thanks for putting this together. Unfortunately, this is generated code, and can't be directly modified here. We actually [made this fix in the code generator recently](https://github.com/googleapis/gapic-generator-python/pull/2577) though, so you should see it roll out to the library soon

PR #16480, opened 2026-03-30
fix(auth): use requests transport for GCE MDS

daniel-sanche:
Fixes https://github.com/googleapis/google-cloud-python/issues/16090

We were seeing errors on GCE environments, because the library would use _http_client to communicate to the metadata server, even when mTLS is enabled. This resulted in failures, because _http_client doesn't support https.

This PR addresses the issue by using requests as the transport instead. This depends on the optional `requests` dependency, but requests is already [required by compute engine](https://github.com/googleapis/google-cloud-python/blob/51a96200324965f071dd9542d25b907c1420b3f6/packages/google-auth/google/auth/compute_engine/_metadata.py#L27), so there are no changes there.

I was able to reproduce and verify this test on a GCE VM


packages/google-auth/tests/test__default.py
```diff
@@ -890,6 +890,18 @@ def test__get_gce_credentials_explicit_request(ping):
     ping.assert_called_with(request=mock.sentinel.request)
 
 
+@mock.patch(
```
nolanleastin, line 893: I wonder if we can create a test that just calls google.auth.default() such that it will trigger this logic?

daniel-sanche: Sure, I added an extra test

packages/google-auth/tests/test__default.py
```diff
@@ -1006,6 +1018,35 @@ def test_default_fail(unused_gce, unused_gae, unused_sdk, unused_explicit):
...
+    autospec=True,
+)
+@mock.patch(
+    "google.auth.compute_engine._metadata.is_on_gce", return_value=False, autospec=True
+)
+@mock.patch("google.auth.transport.requests.Request", autospec=True)
+def test_default_gce_triggers_request_creation(
+    mock_request_cls, is_on_gce, unused_gae, unused_sdk, unused_explicit
+):
+    with pytest.raises(exceptions.DefaultCredentialsError):
```
nolanleastin, line 1043: Why is this raising a credentials error?

daniel-sanche: That's [what is raised](https://github.com/googleapis/google-cloud-python/blob/2e508b6b008df77e886ce83f8f5ac3fd8fbe4ff7/packages/google-auth/google/auth/_default.py#L752) when no default credentials are found. 

We mocked all the other credential paths to be empty. We could add some additional mocks to fake the GCE metadata server to make it look like it's returning real credentials. But that would be pretty complicated, and there are already quite a few mocks to keep track of

nolanleastin:
This LGTM

PR #16481, opened 2026-03-30
chore(auth): unblock tests

daniel-sanche:
Fix failing system tests, related to empty tokens

~Also added back [scipts from split repo](https://github.com/googleapis/google-auth-library-python/tree/main/scripts), which were used to set up system tests. These scripts are referenced in CONTRIBUTING.md~


packages/google-auth/scripts/decrypt-secrets.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -e
+
+# Copyright 2015 Google Inc. All rights reserved.
```
ohmayr: Nit: Can we update the copyright here and in other places?

parthea (commented):
Are the files in the scripts folder used in tests? If not, can we consider adding these to client library maintainer documentation, rather than having unused scripts?

PR #16501, opened 2026-04-01
perf: add multiplexing performance tests for AsyncMultiRangeDownloader

zhixiangli:
## Overview
This PR introduces new microbenchmarks to measure and expose the performance bottleneck caused by lock contention in the `AsyncMultiRangeDownloader`. It provides a concrete way to compare the previous serialized implementation against the new multiplexed architecture.

### Before (Serialized via Lock)
In the previous implementation, `download_ranges` used a shared lock to prevent concurrent access to the bidi-gRPC stream. This meant that even with multiple coroutines, only one could "own" the stream at a time. The entire download cycle (Send -> Receive All) had to complete before another task could start.

**Execution Flow:**
```mermaid
sequenceDiagram
    participant C1 as Coroutine 1
    participant C2 as Coroutine 2
    participant S as gRPC Stream

    C1->>C1: Acquire Lock
    C1->>S: Send Requests
    S-->>C1: Receive Data (Streaming...)
    S-->>C1: End of Range
    C1->>C1: Release Lock
    
    Note over C2: Waiting for Lock...
    
    C2->>C2: Acquire Lock
    C2->>S: Send Requests
    S-->>C2: Receive Data (Streaming...)
    S-->>C2: End of Range
    C2->>C2: Release Lock
```

### After (Multiplexed Concurrent)
With the introduction of the `_StreamMultiplexer`, multiple coroutines can now share the same stream concurrently. Requests are interleaved, and a background receiver loop routes incoming data to the correct task using `read_id`.

**Execution Flow:**
```mermaid
sequenceDiagram
    participant C1 as Coroutine 1
    participant C2 as Coroutine 2
    participant M as Multiplexer
    participant S as gRPC Stream

    C1->>M: Send Requests
    M->>S: Forward Req 1
    C2->>M: Send Requests
    M->>S: Forward Req 2
    
    Note over C1,C2: Tasks wait on their own queues
    
    S-->>M: Data for C1
    M-->>C1: Route to Q1
    S-->>M: Data for C2
    M-->>C2: Route to Q2
    S-->>M: Data for C1
    M-->>C1: Route to Q1
```

## How the Benchmark Works
This PR adds a `read_rand_multi_coro` workload that:
1. Spawns multiple asynchronous tasks (coroutines).
2. Shares a single `AsyncMultiRangeDownloader` instance across all tasks.
3. Simulates the old serialized behavior by explicitly passing a `shared_lock` to `download_ranges`.
4. Measures total throughput (MiB/s) and resource utilization.

## Key Changes
- **`test_reads.py`**: Refactored to support launching concurrent coroutines within a single worker process.
- **`config.yaml`**: Added `read_rand_multi_coro` with 1, 16 coroutines to stress the downloader.
- **`config.py`**: Updated naming convention to include coroutine count (e.g., `16c`) in reports for easier differentiation.

### DIFF
--- a/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
+++ b/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
@@ -118,8 +118,6 @@ async def _download_time_based_async(client, filename, params):
     mrd = AsyncMultiRangeDownloader(client, params.bucket_name, filename)
     await mrd.open()
 
-    shared_lock = asyncio.Lock()
-
     async def _worker_coro():
         total_bytes_downloaded = 0
         offset = 0
@@ -148,7 +146,7 @@ async def _download_time_based_async(client, filename, params):
                     if offset + params.chunk_size_bytes > params.file_size_bytes:
                         offset = 0  # Reset offset if end of file is reached
 
-            await mrd.download_ranges(ranges, lock=shared_lock)
+            await mrd.download_ranges(ranges)
 
             bytes_in_buffers = sum(r[2].getbuffer().nbytes for r in ranges)
             assert bytes_in_buffers == params.chunk_size_bytes * params.num_ranges


packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
```diff
@@ -115,47 +115,53 @@ def _download_time_based_json(client, filename, params):
...
-
     mrd = AsyncMultiRangeDownloader(client, params.bucket_name, filename)
     await mrd.open()
 
-    offset = 0
-    is_warming_up = True
-    start_time = time.monotonic()
-    warmup_end_time = start_time + params.warmup_duration
-    test_end_time = warmup_end_time + params.duration
+    shared_lock = asyncio.Lock()
```
chandra-siri: it's not needed.

zhixiangli: The new changes in https://github.com/googleapis/google-cloud-python/pull/16528 do not require it. However, with the previous implementation, I feel safer passing the lock in, especially since multiple threads might create the lock (see https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-storage/google/cloud/storage/asyncio/async_multi_range_downloader.py#L400-L401).

We can clean this up as part of the process to deprecate the lock in `download_ranges`.

chandra-siri: these changes will go in only after - https://github.com/googleapis/google-cloud-python/pull/16528 , right ?

What's the issue ? Do you forsee undoing the https://github.com/googleapis/google-cloud-python/pull/16528 at anypoint in future so that this "shared_lock" might be needed in this test ?

zhixiangli: Deleted. The original purpose was to test the before and after, passing in the lock is more appropriate, but I can keep it in my local env for testing.

Issue #16489, opened 2026-03-31
[post-migration] add/ resolve failing tests in bigframes


chalmerlowe:
During [migration of the bigframes split repo](https://github.com/googleapis/google-cloud-python/pull/16057), there were issues with the mypy presubmits and the Kokoro System Test presubmits failing.

Resolve those issues, ensure the tests pass, and remove the `release-blocked` tag from the  `.librarian/config.yaml` file for the bigframes package..

## Here is a summary of steps that should resolve the Kokoro System Test presubmits:

---

### Part 1: Tasks that the BigFrames team will need to do with a note explaining the ask.

**Subject:** IAM permissions request for BigFrames CI/CD test service accounts

Hi team, 

We are currently blocked by a few cross-project IAM permission errors in the `google-cloud-python` monorepo CI/CD. The tests run from our project (`precise-truck-742`), but require read/encryption access to resources in your `bigframes-dev-*` projects. 

Could you please run the following `gcloud` commands to grant our test service accounts the necessary access? 

**1. Cloud KMS Encryption Access (`bigframes-dev-perf`)**
* **Why:** Our BigQuery service agent needs permission to use your CMEK key to test encrypted queries and models.
* **Command:**
```bash
gcloud kms keys add-iam-policy-binding bigframesKey \
    --keyring bigframesKeyRing \
    --location us \
    --project bigframes-dev-perf \
    --member="serviceAccount:bq-1065521786570@bigquery-encryption.iam.gserviceaccount.com" \
    --role="roles/cloudkms.cryptoKeyEncrypterDecrypter"
```

**2. GCS Bucket Read Access (`bigframes-dev-testing`)**
* **Why:** Our dynamic BigQuery Connection service accounts need to list and read images from `gs://bigframes-dev-testing/a_multimodel/images/*` to test Vertex AI multi-model generation.
* **Commands:**
```bash
gcloud storage buckets add-iam-policy-binding gs://bigframes-dev-testing \
    --member="serviceAccount:bqcx-1065521786570-ljkx@gcp-sa-bigquery-condel.iam.gserviceaccount.com" \
    --role="roles/storage.objectViewer"

gcloud storage buckets add-iam-policy-binding gs://bigframes-dev-testing \
    --member="serviceAccount:bqcx-1065521786570-qgxu@gcp-sa-bigquery-condel.iam.gserviceaccount.com" \
    --role="roles/storage.objectViewer"
```

[...]

PR #16525, opened 2026-04-01
chore: updates tests for bigframes package

chalmerlowe:
WIP: First crack at resolving a number of the concerns in ISSUE #16489

### DIFF
--- a/packages/bigframes/tests/system/small/ml/test_cluster.py
+++ b/packages/bigframes/tests/system/small/ml/test_cluster.py
@@ -151,13 +151,16 @@ def test_kmeans_cluster_centers(penguins_kmeans_model: cluster.KMeans):
         # Accept BOTH python lists AND numpy arrays
         if isinstance(val, (list, np.ndarray)) and len(val) > 0:
             # Take abs of value first, then sort
-            processed = [{"category": x["category"], "value": abs(x["value"])} for x in val]
+            processed = [
+                {"category": x["category"], "value": abs(x["value"])} for x in val
+            ]
             return sorted(processed, key=lambda x: x["category"])
         return val
 
-
     result["numerical_value"] = result["numerical_value"].abs()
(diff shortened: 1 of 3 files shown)


avijitsahaaornoy (commented):
explicit

packages/bigframes/tests/system/small/ml/test_cluster.py
```diff
@@ -141,6 +142,26 @@ def test_kmeans_cluster_centers(penguins_kmeans_model: cluster.KMeans):
         .sort_values(["centroid_id", "feature"])
         .reset_index(drop=True)
     )
+
+    # FIX: Helper to ignore row order inside categorical_value lists
```
tswast, line 146: Good catch!

IMO, these tests are probably too flakey even after this change. I'd be happy if we just checked to make sure we had the expected columns + expected number of rows. In fact, check if we get *at least* the expected columns, because BQML has been known to add more on us.

Same goes for the other bigframes.ml tests modified in this PR.

chalmerlowe: @tswast 

I would like to propose we issue a bug to address this comment.

* Partly because further revision to the tests feel 'out-of-scope' against my original mandate of completely the migration by getting the presubmits to green.
* But more importantly, this is likely to be a broader problem than just this handful of edits. This feels like a question about testing philosophy and how to accurately confirm that the library sends and receives data properly (i.e. what is good enough when confirming results versus expected).

tswast: Given that this PR is adding complicated logic to these tests, and that logic could be removed if we soften up our assertions, I do not agree that it is out of scope.

tswast: I've filed b/501420962 to clean these up.

tswast (approved):

PR #16528, opened 2026-04-02
feat: implement AsyncMultiRangeDownloader with multiplexed bidi-gRPC stream support

zhixiangli:
This PR implements **`AsyncMultiRangeDownloader`** with a new **`_StreamMultiplexer`**, enabling multiple concurrent range downloads to share a single bidirectional gRPC stream.

### Before vs. After

| Feature | Before | After (This PR) |
| :--- | :--- | :--- |
| **Concurrency** | Sequential or multiple connections | Concurrent over **one** connection |
| **Overhead** | High (multiple gRPC streams) | Low (multiplexed single stream) |
| **Reliability** | Per-stream retry logic | Unified generation-gated reopening |

### How it works

The system uses a background **`_StreamMultiplexer`** to manage the shared bidirectional stream:

1.  **Requests**: Concurrent tasks send range requests (`BidiReadObjectRequest`) directly to the shared stream.
2.  **Multiplexing**: A background **Recv Loop** listens for all responses. It uses the `read_id` in each response to route data to the correct task-specific `asyncio.Queue`.
3.  **Error Handling**: If the stream breaks, a **generation-gated lock** ensures the stream is reopened only once. All active tasks receive a `_StreamError` and automatically retry using the new stream generation.

---
**Key Changes:**
- **`_StreamMultiplexer`**: Background receiver loop for routing responses.
- **Generation-Gated Reopening**: Coordinates stream recovery across concurrent tasks.
- **`AsyncMultiRangeDownloader` Integration**: Full support for concurrent `download_ranges` calls.

### DIFF
--- a/packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
+++ b/packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
@@ -15,9 +15,10 @@
 from __future__ import annotations
 
 import asyncio
-import grpc
 import logging
 from typing import Awaitable, Callable, Dict, Optional, Set
+
+import grpc
 
 from google.cloud import _storage_v2
 from google.cloud.storage.asyncio.async_read_object_stream import (
@@ -139,6 +140,10 @@ class _StreamMultiplexer:
                         queue = self._queues.get(read_id)
                         if queue:
                             queues_to_notify.add(queue)
+                        else:
+                            logger.warning(
+                                f"Received data for unregistered read_id: {read_id}"
+                            )
                     await asyncio.gather(
                         *(
                             self._put_with_timeout(queue, response)
--- a/packages/google-cloud-storage/tests/unit/asyncio/test_stream_multiplexer.py
+++ b/packages/google-cloud-storage/tests/unit/asyncio/test_stream_multiplexer.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 import asyncio
-import grpc
 from unittest.mock import AsyncMock
 
+import grpc
 import pytest
 
 from google.cloud import _storage_v2
@@ -325,7 +325,7 @@ class TestRecvLoop:
         assert err.exception is exc
 
     @pytest.mark.asyncio
-    async def test_unknown_read_id_is_dropped(self):
+    async def test_unknown_read_id_is_dropped(self, caplog):
         # Given a multiplexer and a response with an unknown read ID
         mock_stream = AsyncMock()
         resp = _make_response(read_id=999)
@@ -340,6 +340,9 @@ class TestRecvLoop:
         # Then the response is dropped and only StreamEnd is received
         end = await asyncio.wait_for(queue.get(), timeout=1)
         assert isinstance(end, _StreamEnd)
+
+        # And a warning is logged
+        assert "Received data for unregistered read_id: 999" in caplog.text
 
 
 class TestSend:


packages/google-cloud-storage/tests/system/test_zonal.py
```diff
@@ -594,3 +595,194 @@ async def _run():
         gc.collect()
 
     event_loop.run_until_complete(_run())
+
+
+def test_mrd_concurrent_download(
+    storage_client, blobs_to_delete, event_loop, grpc_client
```
chandra-siri: nit: can you use `grpc_client_direct` , although both works but we're generally interested in direct path

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,198 @@
...
+
+from google.cloud import _storage_v2
+from google.cloud.storage.asyncio.async_read_object_stream import (
+    _AsyncReadObjectStream,
+)
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_QUEUE_MAX_SIZE = 100
+_DEFAULT_PUT_TIMEOUT = 20.0
```
chandra-siri: nit: consider `_DEFAULT_PUT_TIMEOUT_SECONDS`

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class _StreamMultiplexer:
+    """Multiplexes concurrent download tasks over a single bidi-gRPC stream.
+
+    Routes responses from a background recv loop to per-task asyncio.Queues
+    keyed by read_id. Coordinates stream reopening via generation-gated
+    locking.
+
+    A slow consumer on one task will slow down the entire shared connection
```
chandra-siri, line 55: does it makes sense @zhixiangli ?

packages/google-cloud-storage/google/cloud/storage/asyncio/async_multi_range_downloader.py
```diff
@@ -419,99 +434,97 @@ async def download_ranges(
...
+                    # Receive Responses
                     while pending_read_ids:
-                        response = await self.read_obj_str.recv()
-                        if response is None:
+                        item = await queue.get()
+
+                        if isinstance(item, _StreamEnd):
+                            if pending_read_ids:
+                                last_broken_generation = my_generation
+                                raise exceptions.ServiceUnavailable(
```
chandra-siri, line 495: This may not be required. See my comment on [_stream_multiplexer.py](https://github.com/googleapis/google-cloud-python/pull/16528/changes#r3044568038)

packages/google-cloud-storage/google/cloud/storage/asyncio/async_multi_range_downloader.py
```diff
@@ -419,99 +434,97 @@ async def download_ranges(
...
+                            else self._multiplexer.stream_generation
+                        )
+                        stream_factory = self._create_stream_factory(state, metadata)
+                        await self._multiplexer.reopen_stream(
+                            broken_gen, stream_factory
                         )
-                        self._is_stream_open = True
 
-                    pending_read_ids = {r.read_id for r in requests}
+                    my_generation = self._multiplexer.stream_generation
```
chandra-siri: nit: how about renaming it to `stream_generation` ?

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,198 @@
+# Copyright 2025 Google LLC
```
chandra-siri: nit: 2026

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,178 @@
...
+                        if queue:
+                            queues_to_notify.add(queue)
+                    for queue in queues_to_notify:
+                        await queue.put(response)
+                else:
+                    for queue in self._get_unique_queues():
+                        await queue.put(response)
+        except asyncio.CancelledError:
+            raise
+        except Exception as e:
```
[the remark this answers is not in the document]
chandra-siri: @zhixiangli , Can you log a warning?. Irrespective of whether this gets retried successfully or not, it's better to log it when error occurs. ( I'm suggesting _warning_ because it might be success in the next attempt)

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,198 @@
...
+                        )
+                    )
+                    return
+
+                if response.object_data_ranges:
+                    queues_to_notify: Set[asyncio.Queue] = set()
+                    for data_range in response.object_data_ranges:
+                        read_id = data_range.read_range.read_id
+                        queue = self._queues.get(read_id)
+                        if queue:
```
chandra-siri, line 141: else raise error, Ideally queue should be present.

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
```diff
@@ -0,0 +1,198 @@
...
+                try:
+                    queue.get_nowait()
+                except asyncio.QueueEmpty:
+                    pass
+
+    async def _recv_loop(self) -> None:
+        try:
+            while True:
+                response = await self._stream.recv()
+                if response is None:
```
chandra-siri: I don't think we need this condition. 

just below two are sufficient.

packages/google-cloud-storage/tests/system/test_zonal.py
```diff
@@ -594,3 +595,194 @@ async def _run():
...
+):
+    """
+    Test that mrd can handle concurrent `download_ranges` calls correctly.
+    Tests overlapping ranges, high concurrency (len > 100 multiplexing batch limits),
```
chandra-siri: I don't think this file is right place to have such tests. This test file is just to make sure all the functionality works fine. So I would suggest to keep it minimal say 3 tasks for each chunk size (small/med/large)

also try to parametrize them - go/python-tips/011#parameterize

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
```diff
@@ -102,69 +105,39 @@ async def test_create_mrd(self, mock_cls_async_read_object_stream):
         "google.cloud.storage.asyncio.async_multi_range_downloader._AsyncReadObjectStream"
     )
     @pytest.mark.asyncio
-    async def test_download_ranges_via_async_gather(
+    async def test_download_ranges(
         self, mock_cls_async_read_object_stream, mock_random_int
     ):
-        # Arrange
```
chandra-siri, line 108: don't remove this. Please verify the same for all other tests.

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
```diff
@@ -419,6 +422,8 @@ async def test_create_mrd_with_generation_number(
         mock_stream.generation_number = _TEST_GENERATION_NUMBER
         mock_stream.persisted_size = _TEST_OBJECT_SIZE
         mock_stream.read_handle = _TEST_READ_HANDLE
+        mock_stream.is_stream_open = True
+        mock_stream.recv = AsyncMock(side_effect=asyncio.Event().wait)
```
chandra-siri: adding these lines have NO effect.

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
chandra-siri, line 443: Although you haven't changed code related to this line, I would appreciate if you can add assert on _read_handle_ and _object_size_

packages/google-cloud-storage/tests/unit/asyncio/test_stream_multiplexer.py
```diff
@@ -0,0 +1,503 @@
+# Copyright 2025 Google LLC
```
chandra-siri: nit: 2026

packages/google-cloud-storage/tests/unit/asyncio/test_stream_multiplexer.py
```diff
@@ -0,0 +1,503 @@
...
+        error = _StreamError(exc, generation=3)
+        assert error.exception is exc
+        assert error.generation == 3
+
+    def test_stream_end_is_instantiable(self):
+        sentinel = _StreamEnd()
+        assert isinstance(sentinel, _StreamEnd)
+
+
+class TestStreamMultiplexerInit:
```
chandra-siri, line 43: please follow - go/unit-testing-practices?polyglot=python#structure for all tests in this file.

chandra-siri (changes requested):

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
zhixiangli: Make sense when using bounded queues, as unbounded queues cause memory issues and crashes. Therefore, we should preferably use bounded queues to slow the process down.

Here's the scenario that causing the slow down:
1. Single Shared Receiver: The multiplexer uses a single background loop/task to receive all data chunks from the shared gRPC stream.
2. Bounded Queues: It routes these chunks to individual, size-limited (asyncio.Queue) queues for each consumer task.
3. Blocking Backpressure: If one consumer task is slow, its queue fills up. The shared background receiver will then block when trying to put more data into that specific full queue (using await queue.put() or a timeout).
4. Connection Stalling: While the shared receiver is blocked waiting for that one slow consumer's queue to free up, it cannot read new data from the gRPC stream for any task. This stops data flow for all other consumers and eventually triggers gRPC's network-level flow control, slowing down the entire physical connection.

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
zhixiangli: I'm considering an edge case in `AsyncMultiRangeDownloader`: the stream is public, meaning it's possible for some external process to close it gracefully.

1. If the stream is closed gracefully, `recv` returns `None`.
2. Other tasks still consuming from the stream will receive the stream closed signal.
3. Those tasks should then reopen the stream to continue the operation.

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
zhixiangli: I prefer to ignore the error here, as it seems like an edge case:

1. Task A sends 10 requests.Due to an error (perhaps transient) from other tasks
2. Task A unregisters to retry.
3. There are outstanding responses for Task A to receive, but it has already been unregistered.

packages/google-cloud-storage/google/cloud/storage/asyncio/async_multi_range_downloader.py
zhixiangli: This is for the case where the stream is closed but there are still outstanding read_ids responses to receive, so the exception is raised to trigger a retry.

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
chandra-siri: if the stream is gracefully closed , it'll not return `None`, it'll return `grpc.aio.EOF`. 

but I understand what you're saying, it makes sense. Should we check for `if response == EoF ` 



We've used it over here - https://github.com/googleapis/google-cloud-python/blob/281eaae1d305a87160877c80c2b8f6fcae06301e/packages/google-cloud-storage/google/cloud/storage/asyncio/async_write_object_stream.py#L194

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
chandra-siri: @zhixiangli , resolved ? forgot to upload changes ?

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
chandra-siri: Thanks @zhixiangli

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
chandra-siri: there could be cases where server malforms. (happened in past) . I don't want to skip these "edge" cases especially when we're in early stages of this API.

packages/google-cloud-storage/tests/unit/asyncio/test_async_multi_range_downloader.py
zhixiangli: Resolved. I didn't actually remove the test; they just showed up as deleted in diff.

packages/google-cloud-storage/google/cloud/storage/asyncio/_stream_multiplexer.py
zhixiangli: In the shared stream/retry flow, a task can unregister its read_ids before all in-flight responses for its previous attempt have been drained. A late response for that old attempt would then hit this path.

If we raise here, that exception tears down the recv loop and gets propagated to all active queues, so one stale response can fail unrelated tasks sharing the multiplexer. Dropping the response seems safer.

If we want better visibility, I’d rather add logging for unknown read_ids than fail the whole shared stream. WDYT?

zhixiangli: I updated the code to raise the error as a transient stream-level failure.

chandra-siri: logging a warning also sounds good to me.

zhixiangli: Updated to log a warning.

PR #16541, opened 2026-04-02
refactor(bigframes): Make join nullity optimizations more robust

TrevorBergeron:


packages/bigframes/bigframes/core/nodes.py
```diff
@@ -204,6 +204,7 @@ class InNode(BigFrameNode, AdditiveNode):
     right_child: BigFrameNode
     left_col: ex.DerefOp
     indicator_col: identifiers.ColumnId
+    nulls_equal: bool = True
```
sycai, line 208: Could you add a comment on what this field means? It's not very obvious to me

TrevorBergeron: added comment

sycai (approved):

PR #16546, opened 2026-04-02
feat(google-auth): make _CLOUD_RESOURCE_MANAGER URL universe-domain-aware

### LINKED ISSUE #16545: feat(google-auth): make Cloud Resource Manager URL universe-domain-aware in external_account credentials

`_CLOUD_RESOURCE_MANAGER` in `google/auth/external_account.py` is hardcoded to `cloudresourcemanager.googleapis.com`. It should respect the configured `universe_domain`, just like `_DEFAULT_TOKEN_URL` already does.

### Desired code experience

```python
# file: google/auth/external_account.py

# Before — hardcoded to googleapis.com:
_CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.googleapis.com/v1/projects/"

# After — universe-domain-aware:
_CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.{universe_domain}/v1/projects/"
```
Credentials constructed with a non-default universe_domain (e.g. sovereign clouds) should automatically use https://cloudresourcemanager.<universe_domain>/v1/projects/ when calling get_project_id().

### Expected results

creds = ExternalAccountCredentials(..., universe_domain="example.com")
**get_project_id() hits https://cloudresourcemanager.example.com/v1/projects/<number>** instead of **cloudresourcemanager.googleapis.com**
google-auth (latest)

### Use case
Users targeting universe domains other than googleapis.com (e.g. private or sovereign Google Cloud environments) need all endpoint URLs, including the Cloud Resource Manager URL used for project ID resolution, to honor the configured universe domain. The existing _DEFAULT_TOKEN_URL already follows this pattern.

### Additional context
The fix mirrors the pattern introduced in googleapis/google-cloud-python@33e8af3 for _DEFAULT_TOKEN_URL. A {universe_domain} placeholder is substituted at credential construction time into self._cloud_resource_manager_url.

baha-zrelli:
Replace hardcoded googleapis.com in _CLOUD_RESOURCE_MANAGER with a {universe_domain} placeholder, resolved at credential construction time via self._cloud_resource_manager_url. This mirrors the existing pattern used for _DEFAULT_TOKEN_URL.

Add tests verifying the URL is correctly built for both the default (googleapis.com) and custom universe domains, including an end-to-end test through get_project_id.

Fixes #16545

### DIFF
--- a/packages/google-auth/google/auth/external_account.py
+++ b/packages/google-auth/google/auth/external_account.py
@@ -398,7 +398,7 @@ class Credentials(
         project_number = self.project_number or self._workforce_pool_user_project
         if project_number and scopes:
             headers = {}
-            url = self._cloud_resource_manager_url + project_number
+            url = "{}{}".format(self._cloud_resource_manager_url, project_number)
             self.before_request(request, "GET", url, headers)
             response = request(url=url, method="GET", headers=headers)
 
--- a/packages/google-auth/google/auth/external_account.py
+++ b/packages/google-auth/google/auth/external_account.py
@@ -52,7 +52,7 @@ _STS_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:token-exchange"
 # The token exchange requested_token_type. This is always an access_token.
 _STS_REQUESTED_TOKEN_TYPE = "urn:ietf:params:oauth:token-type:access_token"
 # Cloud resource manager URL used to retrieve project information.
-_CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.googleapis.com/v1/projects/"
+_CLOUD_RESOURCE_MANAGER = "https://cloudresourcemanager.{universe_domain}/v1/projects/"
 # Default Google sts token url.
 _DEFAULT_TOKEN_URL = "https://sts.{universe_domain}/v1/token"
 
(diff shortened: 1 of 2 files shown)


vverman (approved):
Approved.

Just curious, as to whether this a long-standing concern or an one-off observed bug?

baha-zrelli:
> Approved.
> 
> Just curious, as to whether this a long-standing concern or an one-off observed bug?

Hey, this is a one-off observed bug because the url was hardcoded which not adequate to sovereign-cloud scenarios (GCD with s3ns: https://cloud.google.com/sovereign-cloud) , though we'll probably have more of these in the future

viacheslav-rostovtsev (approved):
LGTM

packages/google-auth/google/auth/external_account.py
```diff
@@ -164,6 +164,9 @@ def __init__(
             self._token_url = self._token_url.replace(
                 "{universe_domain}", self._universe_domain
             )
+        self._cloud_resource_manager_url = _CLOUD_RESOURCE_MANAGER.replace(
+            "{universe_domain}", self._universe_domain
+        )
```
daniel-sanche, line 169: A couple inconsequential comments: 
- do we actually need to save this as an attribute? Seems like it's still only used in one place
- since we're using `{universe_domain}` as a template, maybe we should handle the `project_number` in the same way, instead of concatenating it on the end

daniel-sanche (approved):
LGTM, with a couple small comments

packages/google-auth/google/auth/external_account.py
baha-zrelli: Tried to follow similar behaviour of previous PR so I've limited myself on made changes, would be good if we can go with this to unblock partner and do those changes later on.

PR #16606, opened 2026-04-09
feat(bigframes): Support Expression objects in create_model options

tswast:
Allows users to specify BigFrames `Expression` objects as values within the `options` dictionary when calling `create_model`. These values are compiled natively to their scalar SQL representation in the resulting `CREATE MODEL` BigQuery DDL. Snapshot tests confirm the translation logic.

---
*PR created automatically by Jules for task [15193413976404138758](https://jules.google.com/task/15193413976404138758) started by @tswast*


tswast:
Mailed https://github.com/googleapis/google-cloud-python/pull/16640 as a follow-up to this PR that adds support for hyperparameter tuning once this change is in.

packages/bigframes/bigframes/core/sql/ml.py
```diff
@@ -70,7 +74,10 @@ def create_model_ddl(
     if options:
         rendered_options = []
         for option_name, option_value in options.items():
-            if isinstance(option_value, (list, tuple)):
+            if isinstance(option_value, col.Expression):
+                sg_expr = expression_compiler.compile_expression(option_value._value)
+                rendered_val = sg_sql.to_sql(sg_expr)
```
TrevorBergeron, line 79: I think we can just directly compile here fine because there are no columns in scope. For other functions, with dataframe inputs, may need to resolve labels to column ids.

TrevorBergeron (approved):

packages/bigframes/bigframes/core/sql/ml.py
tswast: Good point. I'll make sure to include this in the policy doc I'm putting together.

PR #16610, opened 2026-04-10
Migration: dry run of steps 6-8

jskeet:
(This is not expected to be merged)


.librarian/config.yaml
```diff
@@ -1,51 +1,564 @@
...
+  - generate_blocked: true
+    id: grpc-google-iam-v1
+  - generate_blocked: true
+    id: pandas-gbq
+  - generate_blocked: true
+    id: proto-plus
+  - generate_blocked: true
+    id: sqlalchemy-bigquery
+  - generate_blocked: true
+    id: sqlalchemy-spanner
```
parthea, line 564: Please can you clarify the reason that we need `generate_blocked` for handwritten libraries? Is `generate` only used when api paths exist in state.yaml?

jskeet: The migration code just adds `generate_blocked` for legacylibrarian to *everything* it migrates, to prevent anything being generated. It's somewhat redundant given the release_mode_only in the config anyway, but it does no harm.

parthea: Ack

PR #16622, opened 2026-04-13
fix(sqlalchemy-spanner): handle TOKENLIST and unknown column types in introspection

### LINKED ISSUE #16621: sqlalchemy-spanner: _designate_type crashes with KeyError on TOKENLIST columns
## sqlalchemy-spanner version

1.17.3 (latest as of 2026-04-12; also reproducible on 1.17.2)

## google-cloud-spanner version

3.63.0

## What happened?

`SpannerDialect._designate_type()` crashes with `KeyError: 'TOKENLIST'` when introspecting tables that have `TOKENLIST` columns (used for Spanner full-text search indexes).

### Reproduction

```sql
-- Create a table with a TOKENLIST column (standard Spanner FTS pattern)
ALTER TABLE my_table ADD COLUMN search_content STRING(MAX);

ALTER TABLE my_table
  ADD COLUMN search_content_tokens TOKENLIST
  AS (TOKENIZE_FULLTEXT(search_content)) HIDDEN;

CREATE SEARCH INDEX MySearchIndex
  ON my_table(search_content_tokens)
  PARTITION BY some_id;
```

```python
from sqlalchemy import create_engine, inspect

engine = create_engine("spanner+spanner:///projects/.../instances/.../databases/...")
inspector = inspect(engine)
# This crashes:
cols = inspector.get_columns("my_table")
```

```
  File ".../google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py", line 1216, in _designate_type
    return _type_map[str_repr]
KeyError: 'TOKENLIST'
```

### Root cause

[\`_type_map\`](https://github.com/googleapis/google-cloud-python/blob/main/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py#L107-L120) does not include \`TOKENLIST\`, and [\`_designate_type()\`](https://github.com/googleapis/google-cloud-python/blob/main/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py#L1191-L1216) has no fallback for unknown types — it does a bare dict lookup that raises \`KeyError\`.

### Impact

This blocks **all** table introspection (not just the TOKENLIST table) for any database that uses Spanner full-text search, because tools like \`sqlacodegen\` and \`MetaData.reflect()\` call \`get_multi_columns()\` which hits this crash. The entire introspection fails, not just the one column.

## Two issues

1. **\`TOKENLIST\` is missing from \`_type_map\`.** It should map to a proper \`TypeEngine\` subclass (not \`NullType\` — see note below) so that schema introspection roundtrips correctly through Alembic and other DDL tools, and so TOKENLIST columns can be referenced in \`SEARCH()\`/\`SCORE()\`/\`SNIPPET()\` queries through SQLAlchemy.

2. **\`_designate_type()\` has no fallback for unknown types.** Any new Spanner type will crash introspection until the dialect is updated.

### Why not \`NullType\`?

\`_type_map_inv\` maps \`NullType → "INT64"\`. If TOKENLIST were mapped to NullType, any roundtrip through Alembic \`--autogenerate\` would silently produce \`INT64\` DDL instead of \`TOKENLIST\`. A proper \`TypeEngine\` subclass preserves type identity.

### Why not filter HIDDEN columns?

TOKENLIST columns are always \`HIDDEN\` (excluded from \`SELECT *\`), but they are **readable by name** and actively used in full-text search queries (\`SEARCH()\`, \`SCORE()\`, \`SNIPPET()\`, \`DEBUG_TOKENLIST()\`). Filtering them from introspection would prevent users from referencing them in ORM models. The dialect should reflect them with their proper type and let consumers decide how to handle them.

## Prior art

The Go client had the same issue and fixed it by adding \`TOKENLIST\` as a first-class \`TypeBase\` constant:
- Issue: googleapis/google-cloud-go#10651
- PR: googleapis/google-cloud-go#11522 (released in spanner v1.78.0)
- Pattern: see also the recent [UUID type PR](https://github.com/googleapis/google-cloud-go/pull/14117) which follows the same approach

## Suggested fix

Two complementary changes:

1. **Add a first-class \`TOKENLIST\` type** — a \`TypeEngine\` subclass with entries in \`_type_map\` (forward), \`_type_map_inv\` (inverse), and \`SpannerTypeCompiler\` (DDL generation).

2. **Add a fallback in \`_designate_type()\`** — \`try/except KeyError\` returning \`NullType\` with a \`warnings.warn()\` for truly unknown future types.

Fix PR: #16622

waiho-gumloop:
## Description

`SpannerDialect._designate_type()` crashes with `KeyError: 'TOKENLIST'` when introspecting tables that have `TOKENLIST` columns (introduced by [Spanner full-text search](https://cloud.google.com/spanner/docs/full-text-search)).

### Root cause

1. `_type_map` does not include `TOKENLIST`.
2. `_designate_type()` performs a bare dict lookup with no fallback, so any unrecognized type crashes all introspection — not just the one column.

### Changes

**1. First-class `TOKENLIST` type** (matching the Go client's approach)

Rather than mapping to `NullType` (which would lose type information and roundtrip as `INT64` via `_type_map_inv`), this PR adds a proper `TOKENLIST` class extending `TypeEngine`:
- Forward map: `_type_map["TOKENLIST"] = TOKENLIST`
- Inverse map: `_type_map_inv[TOKENLIST] = "TOKENLIST"`
- DDL compilation: `SpannerTypeCompiler.visit_TOKENLIST()` returns `"TOKENLIST"`

This ensures correct roundtripping through Alembic `--autogenerate` and other schema tools. It also allows TOKENLIST columns to be present in reflected models so they can be referenced in `SEARCH()`, `SCORE()`, and `SNIPPET()` queries via SQLAlchemy.

TOKENLIST columns are generated (`Computed`) and `HIDDEN` (excluded from `SELECT *`), but they are **readable by name** and actively used in full-text search queries. They should be reflected in models, not filtered out.

**2. Unknown type fallback in `_designate_type()`**

Added `try/except KeyError` that returns `NullType` with a `warnings.warn()`. Future-proofs against new Spanner types being added before the dialect is updated.

### Prior art — Go client fix

The Go `spansql` package had the identical issue. The fix added TOKENLIST as a first-class `TypeBase` constant:

- **Issue:** googleapis/google-cloud-go#10651
- **Fix PR:** googleapis/google-cloud-go#11522 (released in spanner v1.78.0, March 2025)
- **Pattern:** New types are added to the `baseTypes` map, `TypeBase` enum, and `SQL()` method — see also the recent [UUID PR](https://github.com/googleapis/google-cloud-go/pull/14117) which follows the same pattern.

This Python PR mirrors that approach within the SQLAlchemy dialect's architecture.

Fixes #16621

## Testing

Added `tests/unit/test_types.py`:
- `TestDesignateType`: known types, STRING/BYTES with length, TOKENLIST → `TOKENLIST` type, unknown type → NullType with warning, ARRAY
- `TestTokenlistType`: forward/inverse type map entries, type compiler roundtrip (`TOKENLIST()` → `"TOKENLIST"`), TypeEngine subclass check

### DIFF
--- a/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import base64
+import warnings
 
 import re
 
@@ -107,6 +108,21 @@ class SpannerPickleType(TypeDecorator):
         return process
 
 
+class TOKENLIST(types.TypeEngine):
+    """Spanner TOKENLIST type for full-text search indexes.
+
+    TOKENLIST columns store tokenized text produced by functions like
+    TOKENIZE_FULLTEXT() and back SEARCH INDEX structures. They are always
+    generated, always HIDDEN, and cannot be read or written directly by
+    applications.
+
+    This type exists so that schema introspection can roundtrip correctly
+    (reflect → DDL generation) without losing type information.
+    """
+
+    __visit_name__ = "TOKENLIST"
+
+
 # Spanner-to-SQLAlchemy types map
 _type_map = {
     "BOOL": types.Boolean,
@@ -122,6 +138,7 @@ _type_map = {
     "TIMESTAMP": types.TIMESTAMP,
     "ARRAY": types.ARRAY,
     "JSON": types.JSON,
+    "TOKENLIST": TOKENLIST,
 }
 
 
@@ -140,6 +157,7 @@ _type_map_inv = {
     types.TIMESTAMP: "TIMESTAMP",
     types.Integer: "INT64",
     types.NullType: "INT64",
+    TOKENLIST: "TOKENLIST",
 }
 
 _compound_keywords = {
@@ -819,6 +837,9 @@ class SpannerTypeCompiler(GenericTypeCompiler):
     def visit_JSON(self, type_, **kw):
         return "JSON"
 
+    def visit_TOKENLIST(self, type_, **kw):
+        return "TOKENLIST"
+
 
 class SpannerDialect(DefaultDialect):
     """Cloud Spanner dialect.
@@ -1226,7 +1247,14 @@ class SpannerDialect(DefaultDialect):
             inner_type = self._designate_type(inner_type_str)
             return _type_map["ARRAY"](inner_type)
         else:
-            return _type_map[str_repr]
+            try:
+                return _type_map[str_repr]
+            except KeyError:
+                warnings.warn(
+                    "Did not recognize Spanner type '%s', "
+                    "mapping it to NullType" % str_repr
+                )
+                return types.NullType()
 
     @engine_to_connection
     def get_multi_indexes(
--- a/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/packages/sqlalchemy-spanner/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -1248,11 +1248,13 @@ class SpannerDialect(DefaultDialect):
             return _type_map["ARRAY"](inner_type)
         else:
             try:
-                return _type_map[str_repr]
+                col_type = _type_map[str_repr]
+                return col_type() if isinstance(col_type, type) else col_type
             except KeyError:
                 warnings.warn(
                     "Did not recognize Spanner type '%s', "
-                    "mapping it to NullType" % str_repr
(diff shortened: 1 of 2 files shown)


parthea:

waiho-gumloop:
Thanks for the review @parthea, updated the change to be CI-compliant.

sinhasubham (approved):

parthea:
https://github.com/googleapis/google-cloud-python/pull/17911 was merged which solves root cause 1 mentioned in https://github.com/googleapis/google-cloud-python/pull/16622#issue-4251069131. Please could you resolve the merge conflict and update this PR to solve root cause 2 only? I'm going to switch this to draft, but please mark it as ready for review once it's ready.

PR #16624, opened 2026-04-13
feat(django-spanner)!: support Django 5.2 and deprecate Django 3.2/4.2

sinhasubham:
This PR migrates the `django-google-spanner` package to support Django 5.2 and deprecates support for older Django versions (3.2 and 4.2). Due to the deprecation of older versions, this is a breaking change and requires a major version release.
**💥 BREAKING CHANGES**:
*   Dropped support for Django 3.2 and 4.2.
*   Minimum supported Python version is now 3.10 (as required by Django 5.2).


packages/django-google-spanner/django_spanner/__init__.py
```diff
@@ -15,51 +16,44 @@
...
+__all__ = ["__version__", "USE_EMULATOR"]
 register_functions()
 register_lookups()
 
 
 def gen_rand_int64():
     # Credit to https://stackoverflow.com/a/3530326.
-    return uuid4().int & 0x7FFFFFFFFFFFFFFF
+    # Use 32-bit integer for Emulator compatibility (High-bit issues observed).
+    return uuid4().int & 0xFFFFFFFF
```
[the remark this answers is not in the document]
sinhasubham: Fixed

packages/django-google-spanner/django_spanner/base.py
```diff
@@ -150,7 +155,7 @@ def get_connection_params(self):
                   in Django Spanner format.
         """
         return {
-            "project": os.environ.get("GOOGLE_CLOUD_PROJECT", "test-project"),
+            "project": os.environ["GOOGLE_CLOUD_PROJECT"],
```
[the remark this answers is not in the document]
sinhasubham: Fixed

packages/django-google-spanner/django_spanner/version.py
```diff
@@ -4,4 +4,4 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
-__version__ = "4.0.3"
+__version__ = "4.0.2"
```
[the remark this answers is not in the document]
sinhasubham: Fixed

packages/django-google-spanner/django_spanner/base.py
```diff
@@ -121,22 +123,25 @@ def instance(self):
...
-        else:
-            client = spanner.Client(
-                project=os.environ.get("GOOGLE_CLOUD_PROJECT", "test-project")
+            return self.settings_dict["OPTIONS"]["client"].instance(
+                self.settings_dict["INSTANCE"]
+            )
+
+        if _SPANNER_CLIENT_CACHE is None:
+            _SPANNER_CLIENT_CACHE = spanner.Client(
+                project=os.environ["GOOGLE_CLOUD_PROJECT"]
```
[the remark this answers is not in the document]
sinhasubham: Fixed

packages/django-google-spanner/django_spanner/__init__.py
```diff
@@ -15,50 +16,43 @@
 
 RANDOM_ID_GENERATION_ENABLED_SETTING = "RANDOM_ID_GENERATION_ENABLED"
 
-from django.conf.global_settings import DATABASES
-from django.db import DEFAULT_DB_ALIAS
-from django.db.models.fields import (
+
+from django.db import DEFAULT_DB_ALIAS  # noqa: E402
```
sakthivelmanii, line 20: why are you suppressing linting errors?

packages/django-google-spanner/django_spanner/__init__.py
```diff
@@ -15,50 +16,43 @@
...
 check_django_compatability(SUPPORTED_DJANGO_VERSIONS)
+
+__all__ = ["__version__", "USE_EMULATOR"]
 register_functions()
 register_lookups()
 
 
 def gen_rand_int64():
     # Credit to https://stackoverflow.com/a/3530326.
+    # Use 32-bit integer for Emulator compatibility (High-bit issues observed).
```
sakthivelmanii, line 55: Is this a valid comment?

packages/django-google-spanner/django_spanner/creation.py
```diff
@@ -27,12 +27,16 @@ def mark_skips(self):
             # Importing a test app that isn't installed raises RuntimeError.
             if test_app in settings.INSTALLED_APPS:
                 test_case = import_string(test_case_name)
-                method = getattr(test_case, method_name)
-                setattr(
-                    test_case,
-                    method_name,
-                    skip("unsupported by Spanner")(method),
-                )
+                try:
```
sakthivelmanii, line 30: Make you sure you clean up the method. Not suppress the error.

packages/django-google-spanner/django_spanner/schema.py
```diff
@@ -514,41 +565,15 @@ def _check_sql(self, name, check):
             "constraint": self.sql_check_constraint % {"check": check},
         }
 
-    def _unique_sql(
```
sakthivelmanii, line 517: why are we removing it?

packages/django-google-spanner/django_spanner/base.py
```diff
@@ -9,15 +9,17 @@
...
 from .client import DatabaseClient
 from .creation import DatabaseCreation
 from .features import DatabaseFeatures
 from .introspection import DatabaseIntrospection
 from .operations import DatabaseOperations
 from .schema import DatabaseSchemaEditor
 
+# Global cache for Spanner client to prevent multiple initializations
+# which can cause OpenTelemetry 'MeterProvider override' crashes.
+_SPANNER_CLIENT_CACHE = None
```
surbhigarg92, line 21: I dont understand why do we need this, ideally our Spanner client library should handle the override crash ?

packages/django-google-spanner/django_spanner/base.py
```diff
@@ -150,7 +158,10 @@ def get_connection_params(self):
                   in Django Spanner format.
         """
         return {
-            "project": os.environ.get("GOOGLE_CLOUD_PROJECT", "test-project"),
+            "project": os.environ.get("GOOGLE_CLOUD_PROJECT")
+            or self.settings_dict.get("project")
+            or self.settings_dict.get("PROJECT")
+            or "test-project",
```
surbhigarg92, line 164: This is being done in multiple places, better to extract in "init" method once and use it everywhere. 

Also why do we support `self.settings_dict.get("project")` and `self.settings_dict.get("PROJECT")` now ?

packages/django-google-spanner/django_spanner/base.py
```diff
@@ -171,7 +182,16 @@ def get_new_connection(self, conn_params):
         :raises: :class:`ValueError` in case the given instance/database
                  doesn't exist.
         """
-        return self.Database.connect(**conn_params)
+        conn_params.pop("instance", None)
+        conn_params.pop("instance_id", None)
+        conn_params.pop("client", None)
```
surbhigarg92, line 187: Why this change is required ?
We are not using `instance_id` or `client` anywhere. We are only using `instance.instance_id` or `instance._client`

packages/django-google-spanner/django_spanner/compiler.py
```diff
@@ -50,157 +48,87 @@ def get_combinator_sql(self, combinator, all):
...
-                            "LIMIT/OFFSET not allowed in subqueries of compound "
-                            "statements."
-                        )
-                    if compiler.get_order_by():
-                        raise DatabaseError(
-                            "ORDER BY not allowed in subqueries of compound statements."
-                        )
-            parts = ()
+        # This method copies the complete code of this overridden method from
+        # Django core and modify it for Spanner by adding one line
```
surbhigarg92, line 52: This is duplicate comment

packages/django-google-spanner/django_spanner/creation.py
```diff
@@ -27,12 +27,16 @@ def mark_skips(self):
...
+                    method = getattr(test_case, method_name)
+                    setattr(
+                        test_case,
+                        method_name,
+                        skip("unsupported by Spanner")(method),
+                    )
+                except AttributeError:
+                    # The test method might not exist in this version of Django.
+                    pass
```
surbhigarg92, line 39: Why do we need this, we have already added an explicit list of skip tests

packages/django-google-spanner/django_spanner/version.py
```diff
@@ -4,4 +4,4 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
-__version__ = "4.0.3"
+__version__ = "4.1.0"
```
surbhigarg92, line 7: Why are we modifying this version manually ? shouldn't this be updated automatically once we do the release ?

packages/django-google-spanner/tests/unit/django_spanner/test_expressions.py
```diff
@@ -19,7 +19,7 @@ def test_order_by_sql_query_with_order_by_null_last(self):
         sql_compiled, _ = compiler.as_sql()
         self.assertEqual(
             sql_compiled,
-            "SELECT tests_report.name FROM tests_report ORDER BY "
+            "SELECT tests_report.name AS name FROM tests_report ORDER BY "
```
surbhigarg92, line 22: For my knowledge why this change is required ?

packages/django-google-spanner/noxfile.py
```diff
@@ -28,7 +28,6 @@
 DEFAULT_PYTHON_VERSION = "3.14"
 
 UNIT_TEST_PYTHON_VERSIONS = [
-    "3.8",
     "3.9",
```
surbhigarg92, line 31: I thought we received 3.9 support as well, as we removed 3.9 integration test suite file

packages/django-google-spanner/noxfile.py
```diff
@@ -125,25 +123,20 @@ def default(session, django_version="3.2"):
 @nox.session(python=ALL_PYTHON)
 def unit(session):
     """Run the unit test suite."""
-    if session.python in ("3.7",):
-        session.skip("Python 3.7 is no longer supported")
-    # TODO: Remove this check once support for Python 3.14 is added to Protobuf.
+    if session.python == "3.9":
+        session.skip("Python 3.9 is not supported for Django 5.2 tests")
```
surbhigarg92, line 127: Lets remove 3.9 from the list then ?

packages/django-google-spanner/django_spanner/version.py
surbhigarg92, line 7: Also this should go as a Major version release, since we are deprecating older vesion of Django. 

PR title should be `feat!` to trigger a major version

packages/django-google-spanner/noxfile.py
```diff
@@ -28,7 +28,6 @@
 DEFAULT_PYTHON_VERSION = "3.14"
 
 UNIT_TEST_PYTHON_VERSIONS = [
```
sakthivelmanii, line 30: Remove support for 3.8 and 3.9

packages/django-google-spanner/django_spanner/operations.py
```diff
@@ -146,6 +161,66 @@ def sql_flush(self, style, tables, reset_sequences=False, allow_cascade=False):
         else:
             return []
 
+    def execute_sql_flush(self, sql_list):
```
sakthivelmanii, line 164: this looks so bruteforce. is there any other better approach?

packages/django-google-spanner/django_spanner/features.py
```diff
@@ -47,2140 +50,2091 @@ class DatabaseFeatures(BaseDatabaseFeatures):
...
+    uses_savepoints = True
+    can_rollback_tests = False  # Spanner savepoints are no-ops; rely on flush.
+    supports_aggregate_filter_clause = False
     # Spanner does not support expression indexes
     # example: CREATE INDEX index_name ON table (LOWER(column_name))
     supports_expression_indexes = False
+    supports_stored_generated_columns = True
 
     # Django tests that aren't supported by Spanner.
     skip_tests = (
```
sakthivelmanii, line 68: We need to group them by cause. Otherwise we will not know if we are doing it correctly or not.

packages/django-google-spanner/django_spanner/operations.py
```diff
@@ -146,6 +161,66 @@ def sql_flush(self, style, tables, reset_sequences=False, allow_cascade=False):
...
+                            raise last_exception
+                        raise DatabaseError(
+                            "Max passes reached in execute_sql_flush without clearing all tables."
+                        )
+
+                    for sql in remaining_sql:
+                        try:
+                            cursor.execute(sql)
+                            progress = True
+                        except Exception as e:
```
surbhigarg92, line 208: catch more specific exception ?

surbhigarg92 (changes requested):

packages/django-google-spanner/django_spanner/__init__.py
sinhasubham: Addressed

packages/django-google-spanner/django_spanner/__init__.py
sinhasubham: I had removed the 32-bit integer changes here but missed to remove this comment earlier. Removed now

packages/django-google-spanner/django_spanner/base.py
sinhasubham: Added this global cache to fix a crash observed during testing: OpenTelemetry MeterProvider override crashes. In Django, DatabaseWrapper instances are created frequently (e.g., per thread or request), leading to multiple spanner.Client initializations. While it would be ideal for the client library to handle this, this cache effectively prevents the crash and resource exhaustion in the meantime.

packages/django-google-spanner/django_spanner/base.py
sinhasubham: Addressed. I made these changes for some debugging purpose earlier, missed to revert in the final commit.

packages/django-google-spanner/django_spanner/base.py
sinhasubham: Because get_connection_params puts instance_id into the conn_params dictionary by default, passing **conn_params will try to pass instance_id again. Python rejects this because it is receiving the same argument twice.
Popping instance_id and client from conn_params is required to support the global client cache fix. Removing this crashes the adapter on connection.

packages/django-google-spanner/django_spanner/compiler.py
sinhasubham: Addressed

packages/django-google-spanner/django_spanner/creation.py
sinhasubham: This is required for multi-version compatibility to prevent the test suite from crashing when a test listed for skipping doesn't exist in the specific Django version being executed. Currently we will support just 5.2 but this will be required in long term.

packages/django-google-spanner/django_spanner/creation.py
sinhasubham: Since we need to support a range of Django versions, a test that is non-existent in 5.2 might still need to be skipped in 4.2. Suppressing the AttributeError is the best way to handle this scenario in long term according to me.

packages/django-google-spanner/django_spanner/operations.py
sinhasubham: Spanner might raise various errors (such as Aborted, FailedPrecondition, or IntegrityError) depending on which constraint is violated first. I felt adding generic exception catching was better in this case instead of maintaining list of exceptions for same block.

packages/django-google-spanner/django_spanner/version.py
sinhasubham: Makes sense, addressed

sinhasubham: Sure, will update the PR title

packages/django-google-spanner/django_spanner/schema.py
sinhasubham: The method was initially removed in an attempt to rely on the Django 5.2 base class implementation. But i agree, the custom implementation is required to prevent Django from generating inline UNIQUE constraints in CREATE TABLE statements, which base class implementation does not support in the same way. I have restored the method and updated it to support Django 5.2's nulls_distinct feature.

packages/django-google-spanner/django_spanner/operations.py
sinhasubham: Earlier, we were relying on Django's default execute_sql_flush, which fails on Spanner when foreign key constraints are present. This retry logic was introduced in this PR specifically to handle Spanner's inability to disable foreign key checks during test cleanup.
I agree that the retry loop in execute_sql_flush is a brute-force approach. But to avoid the complexity of implementing a full topological sort of the tables based on their foreign key dependencies to determine the exact deletion order (child tables first, then parents), i used this multi-pass approach.

packages/django-google-spanner/django_spanner/features.py
sinhasubham: Addressed

packages/django-google-spanner/noxfile.py
sinhasubham: The unit 3.9 workflow step fails without version 3.9 in this list. And we cannot remove 3.9 from github workflows as it is a common workflow file for monorepo now. I will check on this with owning team how we are planning to handle this

packages/django-google-spanner/noxfile.py
sinhasubham: The unit 3.9 workflow step fails without version 3.9 in this list. And we cannot remove 3.9 from github workflows as it is a common workflow file for monorepo now. I will check on this with owning team how we are planning to handle this

packages/django-google-spanner/noxfile.py
sinhasubham: Since unittest.yml is a shared github workflow for all packages in the monorepo, I cannot remove Python 3.9 from it without affecting other libraries.

packages/django-google-spanner/django_spanner/base.py
surbhigarg92: Got it.  Please create a bug in Python Spanner Client and add a todo here to clean this up once the bug is fixed

packages/django-google-spanner/django_spanner/creation.py
surbhigarg92: Got it. Thanks

surbhigarg92 (approved):

packages/django-google-spanner/tests/unit/django_spanner/test_expressions.py
sinhasubham: This change was required because Django 5.2's SQL compiler now generates SELECT tests_report.name AS name (with an explicit alias) when using .values('name'), whereas older Django versions generated just SELECT tests_report.name without the AS clause. Since we are upgrading to support Django 5.2, I updated the test's expected SQL string to match what Django 5.2 actually generates.

sakthivelmanii (approved):

PR #16640, opened 2026-04-13
feat: add support for `hparam_range` and `hparam_candidates` to `bigframes.bigquery.create_model`

tswast:
Fixes internal issue b/501171054
 🦕


packages/bigframes/bigframes/bigquery/_operations/mathematical.py
```diff
@@ -51,3 +53,71 @@ def rand() -> bigframes.core.col.Expression:
         is_deterministic=False,
     )
     return bigframes.core.col.Expression(bigframes.core.expression.OpExpression(op, ()))
+
+
+def hparam_range(min: float, max: float) -> bigframes.core.col.Expression:
```
sycai, line 58: Should we put these two functions in the `ml` package instead? We don't have any hyper-parameter tuning opportunities outside the scope of machine learning, right?

tswast: The ML package corresponds to the ML namespace. These functions aren't in the ML namespace in SQL.

sycai (approved):

PR #16678, opened 2026-04-16
chore: complete google-cloud-dns migration

jskeet:
- Reformat with nox
- Enable generation with librarian
- Regenerate with librarian

### DIFF
--- a/packages/google-cloud-dns/google/cloud/dns/__init__.py
+++ b/packages/google-cloud-dns/google/cloud/dns/__init__.py
@@ -24,12 +24,10 @@ The main concepts with this API are:
   (adding/deleting resource record sets) to a zone.
 """
 
-from google.cloud.dns.version import __version__
-from google.cloud.dns.zone import Changes
 from google.cloud.dns.client import Client
-from google.cloud.dns.zone import ManagedZone
 from google.cloud.dns.resource_record_set import ResourceRecordSet
-
+from google.cloud.dns.version import __version__
+from google.cloud.dns.zone import Changes, ManagedZone
 
 SCOPE = Client.SCOPE
 
--- a/packages/google-cloud-dns/google/cloud/dns/__init__.py
+++ b/packages/google-cloud-dns/google/cloud/dns/__init__.py
@@ -24,9 +24,12 @@ The main concepts with this API are:
   (adding/deleting resource record sets) to a zone.
 """
 
+from google.cloud.dns import version
+
+__version__ = version.__version__
+
 from google.cloud.dns.client import Client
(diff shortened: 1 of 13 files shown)


packages/google-cloud-dns/google/cloud/dns/__init__.py
```diff
@@ -24,12 +24,10 @@
   (adding/deleting resource record sets) to a zone.
 """
 
-from google.cloud.dns.version import __version__
-from google.cloud.dns.zone import Changes
 from google.cloud.dns.client import Client
-from google.cloud.dns.zone import ManagedZone
 from google.cloud.dns.resource_record_set import ResourceRecordSet
-
+from google.cloud.dns.version import __version__
```
parthea, line 29: The import error is due to the `__version__` being moved

```
The Execution Chain
__init__.py starts. It reaches line 27: from google.cloud.dns.client import Client.

client.py starts. It reaches line 21: from google.cloud.dns._http import Connection.

_http.py starts. It reaches line 18: from google.cloud.dns import __version__.

Back to __init__.py: Python looks for __version__ in google.cloud.dns. However, because __init__.py is still waiting for Client to finish loading, it hasn't reached the line where __version__ is defined yet. Crash.
```

Restoring the original code fixes the error

```
# 1. Define/Import the version FIRST
from google.cloud.dns.version import __version__
```

You can see a similar pattern in the generated clients were we have `__version__` at the top

https://github.com/googleapis/google-cloud-python/blob/c422cbb3276c12901884f2c7cc730f03eec30bcd/packages/google-cloud-access-approval/google/cloud/accessapproval_v1/__init__.py#L20-L22

jskeet: That suggests we need to change how formatting happens. All I did was run `nox -s format`.

Perhaps we should:

- Remove skip_generate=true (because we won't regenerate/reformat in librarian anyway)
- Consider this "done" for migration
- Raise a separate issue for the formatting failure

?

packages/google-cloud-dns/google/cloud/dns/_http.py
```diff
@@ -15,7 +15,6 @@
 """Create / interact with Google Cloud DNS connections."""
 
 from google.cloud import _http
-
 from google.cloud.dns import __version__
```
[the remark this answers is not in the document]
jskeet: Ignoring for now, unless the presubmits fail.

packages/google-cloud-dns/google/cloud/dns/__init__.py
jskeet: Fixed in another commit.

parthea (approved):

PR #16679, opened 2026-04-16
chore: complete gapic-generator migration

jskeet:
- Reformat with nox
- Enable generation with librarian
- Regenerate with librarian


jskeet:
@parthea Obviously this is a huge change - but it really is just running nox format. (Librarian doesn't actually "generate" this at all.) We could potentially remove the first commit, bringing it into librarian without reformatting - but if it *is* okay to reformat, I think it's worth doing.

parthea:
According to gemini, this is one of the fixes

```
1. Quoting Style (f-strings)
Expected (Golden): Uses single quotes for the f-string to allow literal double quotes inside:
f'Mollusc is a "{response.taxonomy}"'

Actual (Generated): Uses double quotes for the f-string and escapes the inner double quotes:
f"Mollusc is a \"{response.taxonomy}\""

2. Whitespace (Newlines)
Expected (Golden): Has an extra empty line before the # [END ...] tag.

Actual (Generated): Has no empty line before the # [END ...] tag.
```

I may have some fixes already in https://github.com/googleapis/gapic-generator-python/pull/2499. Let me know if you want to hand this over, and what command I should run to re-generate the code after applying the fix.

jskeet:
Closing as we don't want to reformat this; will raise a PR to just remove skip_generate. (That won't change anything, but removes a "false" blocker.)

PR #16686, opened 2026-04-16
chore: remove obsolete .repo-metadata input files

jskeet:
These were used by legacylibrarian, but are not used at all in the librarian generation process.


jskeet:
The do-not-merge label has been applied so that we can avoid merging this before we've merged the other librarian PRs, but this PR shouldn't get "stale" before then.

parthea (approved):

jskeet:
Merging now, as we're at the stage where if we want to regenerate any packages we'll use librarian rather than legacylibrarian.

PR #16691, opened 2026-04-16
feat: enable mypy session for ndb

chalmerlowe:
This PR enables the mypy session in noxfile.py for ndb and aligns it with the GAPIC generator template. Currently, it fails with 196 errors. This is pushed to allow farming out the work to resolve these errors or for separate review.

### DIFF
--- a/packages/google-cloud-ndb/google/cloud/ndb/_datastore_query.py
+++ b/packages/google-cloud-ndb/google/cloud/ndb/_datastore_query.py
@@ -361,7 +361,9 @@ class _QueryIteratorImpl(QueryIterator):
         return (
             self._batch is None  # Haven't even started yet
             or self._has_next_batch  # There's another batch to fetch
-            or (self._index is not None and self._index < len(self._batch))  # Not done with current batch
+            or (
+                self._index is not None and self._index < len(self._batch)
+            )  # Not done with current batch
         )
 
     @tasklets.tasklet
--- a/packages/google-cloud-ndb/google/cloud/ndb/_datastore_query.py
+++ b/packages/google-cloud-ndb/google/cloud/ndb/_datastore_query.py
@@ -342,6 +342,8 @@ class _QueryIteratorImpl(QueryIterator):
         if self._batch is None:
             yield self._next_batch()  # First time
 
+        assert self._batch is not None
+        assert self._index is not None
         if self._index < len(self._batch):
             raise tasklets.Return(True)
 
@@ -359,7 +361,7 @@ class _QueryIteratorImpl(QueryIterator):
         return (
             self._batch is None  # Haven't even started yet
             or self._has_next_batch  # There's another batch to fetch
-            or self._index < len(self._batch)  # Not done with current batch
+            or (self._index is not None and self._index < len(self._batch))  # Not done with current batch
         )
 
     @tasklets.tasklet
@@ -421,6 +423,8 @@ class _QueryIteratorImpl(QueryIterator):
             self._cursor_before = None
             raise StopIteration
 
+        assert self._batch is not None
+        assert self._index is not None
         # Won't block
         next_result = self._batch[self._index]
         self._index += 1
@@ -446,7 +450,7 @@ class _QueryIteratorImpl(QueryIterator):
         batch = self._batch
         index = self._index
 
-        if batch and index < len(batch):
+        if batch and index is not None and index < len(batch):
             return batch[index]
 
         raise KeyError(index)
@@ -554,6 +558,7 @@ class _PostFilterQueryIteratorImpl(QueryIterator):
         if not self.has_next():
             raise StopIteration()
 
+        assert self._next_result is not None
         # Won't block
         next_result = self._next_result
         self._next_result = None
@@ -718,6 +723,7 @@ class _MultiQueryIteratorImpl(QueryIterator):
         if not self.has_next():
             raise StopIteration()
 
+        assert self._next_result is not None
         # Won't block
         next_result = self._next_result
         self._next_result = None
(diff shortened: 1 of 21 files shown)


packages/google-cloud-ndb/google/cloud/ndb/_datastore_query.py
```diff
@@ -342,6 +342,8 @@ def has_next_async(self):
         if self._batch is None:
             yield self._next_batch()  # First time
 
+        assert self._batch is not None
+        assert self._index is not None
```
parthea: Did you mean to have `assert` as part of this code path? Usually this is only used in tests

chalmerlowe: I looked for ways to provide type narrowing for mypy and `assert` is one way it can be done.

It is admittedly controversial for some in production code.

I went ahead and switched the `assert`s into `if` clauses throughout.

packages/google-cloud-ndb/google/cloud/ndb/context.py
```diff
@@ -313,11 +319,12 @@ def __new__(
             legacy_data=legacy_data,
         )
 
-        context.set_cache_policy(cache_policy)
-        context.set_global_cache_policy(global_cache_policy)
-        context.set_global_cache_timeout_policy(global_cache_timeout_policy)
-        context.set_datastore_policy(datastore_policy)
-        context.set_retry_state(retry)
+        ctx = cast(Any, context)
```
chalmerlowe, line 325:
```suggestion
        # 'context' is dynamically composed at runtime and may include methods 
        # from multiple sources that Mypy cannot statically resolve here. 
        # We cast to `Any` to access the extended policy interface.
        ctx = cast(Any, context)
```

packages/google-cloud-ndb/google/cloud/ndb/_gql.py
```diff
@@ -770,27 +772,34 @@ def _raise_cast_error(message):
 
 
 def _time_function(values):
+    t_tuple: tuple[int, ...]
     if len(values) == 1:
         value = values[0]
         if isinstance(value, str):
             try:
-                time_tuple = time.strptime(value, "%H:%M:%S")
+                st = time.strptime(value, "%H:%M:%S")
```
parthea: Replace `st` with a more meaningful name

parthea: Gemini suggested this refactor to clean up this code and reduce the number of if statements

```
import datetime
import time

def _time_function(values):
    # 1. Normalize input into a tuple of integers
    if len(values) == 1:
        val = values[0]
        if isinstance(val, str):
            try:
                st = time.strptime(val, "%H:%M:%S")
                t_tuple = (st.tm_hour, st.tm_min, st.tm_sec)
            except ValueError as e:
                _raise_cast_error(f"Format error: {e}, {values}")
        elif isinstance(val, int):
            t_tuple = (val,)
        else:
            _raise_cast_error(f"Invalid type {type(val)} for time()")
    elif 1 < len(values) < 4:
        t_tuple = tuple(values)
    else:
        _raise_cast_error(f"Invalid number of arguments: {len(values)}")

    # 2. Convert tuple to datetime.time using unpacking
    try:
        return datetime.time(*t_tuple)
    except (ValueError, TypeError) as e:
        _raise_cast_error(f"Time conversion error: {e}, {values}")
```

chalmerlowe: @parthea 

These two code blocks are fundamentally the same with the exception of item `# 2. Convert tuple to datetime.time using unpacking` so my comments focus on that.

In your comment we assert that Gemini recommends this refactor to reduce the number of `if` statements.

Funnily enough:

Those `if` statements were also **recommended by Gemini** to enable greater type safety with mypy, because the `datetime.time(*t_tuple)` expression caused mypy errors.

```
google/cloud/ndb/_gql.py:793: error: Too many arguments for "time"  [call-arg]
google/cloud/ndb/_gql.py:793: error: Too many positional arguments for "time"  [misc]
```

Here is Gemini's rationale for the approach taken in this PR (this was the third change we made and Gemini felt it was the most important):

**< BEGINNING of GEMINI's exposition >**

3. Exploding the "Splat" Operator (*t_tuple)

This is the most important change for Mypy:

Old: `return datetime.time(*time_tuple)`
The Problem: The `*` (splat) operator tells Python to unpack the `tuple` into arguments. However, `datetime.time()` accepts specific arguments: `hour`, `minute`, `second`, `microsecond`.
If `time_tuple` has a length of 1, it’s just the `hour`. If it's length 3, it's `H, M, S`.

Mypy cannot guarantee at compile-time that the length of the tuple matches what the function expects. It sees `*time_tuple` and worries you might be passing 10 arguments to a function that takes 4.

New:

```python
if len(t_tuple) == 1:
    return datetime.time(t_tuple[0])
elif len(t_tuple) == 2:
    return datetime.time(t_tuple[0], t_tuple[1])
# ...
```

The Benefit: This is called **Exhaustive Checking** or **Manual** Unrolling.
By checking the `len()` explicitly, you are "narrowing" the type. Inside the `if len(t_tuple) == 1` block, Mypy knows for a fact that `t_tuple[0]` exists and is the only argument.
This removes all ambiguity. Mypy can now prove the code is safe without needing a `# type: ignore`.

**< END of GEMINI's exposition >**

I am happy to resolve this in whatever way feels best to you, but I suspect that if I revert back to using the splat operator, we will have to go through another cycle of trying to find a way to keep mypy happy OR add one or more ignore  pragmas.

parthea (approved):

schwehr:
Thank you!!!

Issue #11858, opened 2021-06-17
Remove erroneous folder google/monitoring


parthea:
We have an erroneous google/monitoring folder rather than google/cloud/monitoring_dashboard_v1. This folder should be removed to avoid confusion. 
https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-monitoring-dashboards/google/monitoring

daniel-sanche:
This issue is being closed as part of a bulk review (Filter #1). Please reopen if you believe this is still relevant and requires action.

PR #16692, opened 2026-04-16
chore: complete migration of google-cloud-monitoring-dashboards

### LINKED ISSUE #5339: python: fix and enable generation for google-cloud-monitoring-dashboards
All differences should be understood and reviewed by the Python team.

jskeet:
This requires a new string replacement to avoid a diff in docs/index.rst. (When we audit all the post-processing later, we may decide that the diff is actually beneficial; that's outside the scope of migration.)

Fixes https://github.com/googleapis/librarian/issues/5339

Note that this has an approach of "minimal diff" - #11858 suggests that we remove a stale directory, and we can do that separately from the migration.


.librarian/generator-input/client-post-processing/monitoring-dashboards-compatibility.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2026 Google LLC
```
parthea, line 1: This post processing won't be needed if we remove https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-monitoring-dashboards/google/monitoring . The directory contains stale code (See related issue https://github.com/googleapis/google-cloud-python/issues/11858). I'm going to approve this PR since it's an existing issue but it would be great to have some context added to the description in this file about the long term fix that's needed.

parthea (approved):

.librarian/generator-input/client-post-processing/monitoring-dashboards-compatibility.yaml
jskeet: Sure, will add that to the file now.

Issue #15655, opened 2026-02-27
consider ignoring sphinx docs warnings


daniel-sanche:
### Summary of the issue

We currently have a docs check, that can fail due to sphinx formatting. The docstrings are provided by product teams and are outside the control of our generator.

To make sure this doesn't block our library releases, we should consider ignoring warnings (where reasonable), or consider a way to raise the warning earlier in the process, so the product team will fix it in the source

PR #16698, opened 2026-04-17
chore: complete migration of google-cloud-dataplex

jskeet:
Removed skip_generate and regenerated with Librarian.


parthea:
The docs failure is a known issue (See https://github.com/googleapis/google-cloud-python/issues/15655). The 'fix' is in https://github.com/googleapis/google-cloud-python/pull/16627 but it hasn't been released yet due to the freeze. 

If you're blocked on this, we can update post processing for docs to get them to pass, otherwise wait for the next release of `gapic-generator`. Let me know.

jskeet:
@parthea Only responding on this one bug - how feasible might it be to turn off "docs warnings as errors" until we can release the gapic generator, then turn it back on once we've updated?

parthea:
Switching to draft until checks are green

parthea:
@jskeet We would need something like this for each package where we see the docs failure

https://github.com/googleapis/google-cloud-python/blob/main/.librarian/generator-input/client-post-processing/allow-docs-warnings-for-google-shopping-merchant-accounts.yaml

parthea:
You may be able to add more paths here https://github.com/googleapis/google-cloud-python/blob/4009eda6457f9a5a259671055f7e0e5bb13468d5/.librarian/generator-input/client-post-processing/allow-docs-warnings-for-google-shopping-merchant-accounts.yaml#L17 and just rename this file to be more generic

jskeet:
That sounds like it might be the way to go - will give that a try on Monday.

jskeet:
Closing - regenerated in a separate PR.

PR #16707, opened 2026-04-17
chore: complete migration of pandas-gbq

jskeet:
Removed skip_generate and regenerated with Librarian.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -4623,7 +4623,6 @@ libraries:
       default_version: apiVersion
   - name: pandas-gbq
     version: 0.35.0
-    skip_generate: true
     python:
       library_type: INTEGRATION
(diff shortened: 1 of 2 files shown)


chalmerlowe:
@jskeet I don't yet know what the cause is, but for the life of me, I could not get the `pandas-gbq` prerelease to complete in one of my PRs. The logs die part way through the run, so I was never able to see what the cause was.

I looked at your logs for `unittest/prerelease` and they are identical to what I was seeing in other PRs. If I find a solution I will let you know.

parthea:
Switching to draft until checks are green

ohmayr:
Duplicate of https://github.com/googleapis/google-cloud-python/pull/16739.

PR #16712, opened 2026-04-17
feat(bigtable): add client side metric instrumentation to basic rpcs

daniel-sanche:
Migration of https://github.com/googleapis/python-bigtable/pull/1188 to the monorepo

This PR builds off of https://github.com/googleapis/python-bigtable/pull/1187 to add instrumentation to basic data client rpcs (check_and_mutate, read_modify_write, sample_row_keys, mutate_row)

Metrics are not currently being exported anywhere, just collected and dropped. A future PR will add a GCP exporter to the system


packages/google-cloud-bigtable/google/cloud/bigtable/data/_sync_autogen/client.py
```diff
@@ -1301,27 +1307,29 @@ def mutate_row(
...
+                timeout=attempt_timeout,
+                retry=None,
+            )
+            return tracked_retry(
+                retry_fn=CrossSync._Sync_Impl.retry_target,
+                operation=operation_metric,
+                target=target,
+                predicate=predicate,
+                timeout=operation_timeout,
+            )
```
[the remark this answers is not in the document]
daniel-sanche: These are handled as part of tracked_retry

packages/google-cloud-bigtable/google/cloud/bigtable/data/_helpers.py
```diff
@@ -105,6 +105,7 @@ def _retry_exception_factory(
         tuple[Exception, Exception|None]:
             tuple of the exception to raise, and a cause exception if applicable
     """
+    exc_list = exc_list.copy()
```
mutianf, line 108: why do we need to copy the exception list now?

packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py
```diff
@@ -0,0 +1,1000 @@
...
+        )
+        handler.clear()
+        await table.mutate_row(row_key, mutation)
+        # validate counts
+        assert len(handler.completed_operations) == 1
+        assert len(handler.completed_attempts) == 1
+        # validate operation
+        operation = handler.completed_operations[0]
+        assert isinstance(operation, CompletedOperationMetric)
+        assert operation.final_status.value[0] == 0
```
mutianf, line 247: why is this final_status.value[0] and not just final_status.value? or final_status.name = 'OK'?

packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py
```diff
@@ -0,0 +1,1000 @@
...
+        assert operation.op_type.value == "MutateRow"
+        assert operation.is_streaming is False
+        assert len(operation.completed_attempts) == num_retryable + 1
+        assert operation.cluster_id == expected_cluster
+        assert operation.zone == expected_zone
+        # validate attempts
+        for i in range(num_retryable):
+            attempt = handler.completed_attempts[i]
+            assert isinstance(attempt, CompletedAttemptMetric)
+            assert attempt.end_status.name == "ABORTED"
```
mutianf, line 319: carry over my comment from before, I don't think we retry aborted error for mutate row by default, so maybe use unavailable instead? And also a little suprised that this test passes?

mutianf (commented):
some final nit otherwise lgtm

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
```diff
@@ -1431,26 +1435,28 @@ async def sample_row_keys(
...
+                results = await self.client._gapic_client.sample_row_keys(
+                    request=SampleRowKeysRequest(
+                        app_profile_id=self.app_profile_id, **self._request_path
+                    ),
+                    timeout=next(attempt_timeout_gen),
+                    retry=None,
+                )
+                return [(s.row_key, s.offset_bytes) async for s in results]
+
+            return await tracked_retry(
```
mutianf, line 1453: where did sleep_generator and exception_factory go?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
```diff
@@ -1693,21 +1700,25 @@ async def check_and_mutate_row(
...
+                    row_key=row_key.encode("utf-8")
+                    if isinstance(row_key, str)
+                    else row_key,
+                    app_profile_id=self.app_profile_id,
+                    **self._request_path,
+                ),
+                timeout=operation_timeout,
+                retry=None,
+            )
+            return result.predicate_matched
```
mutianf, line 1721: this is not wrapped in tracked_retry. Will the attempt level metrics (attempt latencies, server latencies and connectivity error count) still be recorded?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
```diff
@@ -1747,20 +1758,22 @@ async def read_modify_write_row(
...
+                    if isinstance(row_key, str)
+                    else row_key,
+                    app_profile_id=self.app_profile_id,
+                    **self._request_path,
+                ),
+                timeout=operation_timeout,
+                retry=None,
+            )
+            # construct Row from result
+            return Row._from_pb(result.row)
```
mutianf, line 1776: same question as check and mutate.

daniel-sanche: Yes, those are also captured [here](https://github.com/googleapis/google-cloud-python/blob/d40fa0ac659ca565af9d760e826d0122932659f0/packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py#L670)

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
daniel-sanche: check_and_mutate shouldn't have retries, right?

But yes, a sngle attempt is will be recorded when this completes. The duration/gfe_latency data [is captured](https://github.com/googleapis/google-cloud-python/blob/d40fa0ac659ca565af9d760e826d0122932659f0/packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py#L850), and will be exported as those metrics in [the follow-up PR](https://github.com/googleapis/google-cloud-python/pull/16760/changes)

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
daniel-sanche: tracked_retry contains them both. needed a custom versions of sleep_generator to report backoff, and exception_factory to report terminal errors metrics module. See go/bigtable-csm-python

packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py
daniel-sanche: value is a tuple of (0, "ok"). We could just do `final_status.name == 'OK'` if you think that's clearer though

packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py
daniel-sanche: We allow the user to set custom retryable errors, which is what the [test is doing here](https://github.com/googleapis/google-cloud-python/blob/d40fa0ac659ca565af9d760e826d0122932659f0/packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py#L302). So that's why it passes.

I can't remember for sure why I chose to use explicit retryable errors here instead of the default. Maybe just so the test makes sense even if you don't know the defaults. I could probably switch it to UNAVAILABLE though if you prefer

packages/google-cloud-bigtable/google/cloud/bigtable/data/_helpers.py
daniel-sanche: This is cleaner, we wouldn't expect a factory method like this to modify its input arguments. This creates an isolated reference

IIRC I think this only came up in test code though

packages/google-cloud-bigtable/tests/system/data/test_metrics_async.py
mutianf: i just noticed that some other tests are using final_status.name :)

mutianf (approved):

PR #16715, opened 2026-04-17
fix: Respect display.progress_bar=None in background threads

shuoweil:
Fixes extraneous progress bar output during asynchronous query executions (such as Colab SQL cells and visualization plugins) when `display.progress_bar = None` is set via `option_context`. 
Previously, background query callbacks evaluated thread-local display options within the background worker thread, falling back to default settings (`"auto"`). 

### Changes
* **`EventEnvelope`**: Introduced a dataclass wrapping query execution events with the initiating thread's active `progress_bar` configuration.
* **Query Initiation**: Updated `start_query_with_job` and `start_query_job_optional` to capture the calling thread's display options and emit events wrapped in an `EventEnvelope`.
* **Event Listeners**: Updated `progress_callback` and `metrics.on_event` to unwrap `EventEnvelope` and respect the preserved option context, successfully suppressing extraneous output in background execution threads.

Verified at here:
before: screen/48zpTTWWjZ5wF9Y
after: screen/7RS7exRrxgPDEqb

Fixes #<461829560> 🦕


tswast:
This PR doesn't touch the thread localility of the configuration object. Please continue with this fix.

packages/bigframes/bigframes/core/events.py
```diff
@@ -23,12 +23,15 @@
 import google.cloud.bigquery._job_helpers
 import google.cloud.bigquery.job.query
 import google.cloud.bigquery.table
+from google.cloud.bigquery.job.query import QueryPlanEntry
 
 import bigframes.session.executor
 
+_FALLBACK_TO_GLOBAL = "fallback_to_global"
```
tswast: Generally "default" would communicate this more succinctly.

shuoweil: Done. Replaced _FALLBACK_TO_GLOBAL with _DEFAULT = "default".

packages/bigframes/bigframes/core/events.py
```diff
@@ -23,12 +23,15 @@
 import google.cloud.bigquery._job_helpers
 import google.cloud.bigquery.job.query
 import google.cloud.bigquery.table
+from google.cloud.bigquery.job.query import QueryPlanEntry
```
tswast: Use the fully qualified name below. According to Google style, import modules not classes or functions.

packages/bigframes/bigframes/core/events.py
```diff
@@ -125,15 +128,21 @@ class BigQuerySentEvent(ExecutionRunning):
     location: Optional[str] = None
     job_id: Optional[str] = None
     request_id: Optional[str] = None
+    progress_bar: Optional[str] = _DEFAULT
```
tswast: Since we are 3.10+ now, we can use the recommended `| None` instead of `Optional` for clarity.

```suggestion
    progress_bar: str| None = _DEFAULT
```

Also, instead of general `str`, consider using `Literal` with the allowed values.

packages/bigframes/bigframes/core/events.py
```diff
@@ -125,15 +128,21 @@ class BigQuerySentEvent(ExecutionRunning):
...
 
     @classmethod
-    def from_bqclient(cls, event: google.cloud.bigquery._job_helpers.QuerySentEvent):
+    def from_bqclient(
+        cls,
+        event: google.cloud.bigquery._job_helpers.QuerySentEvent,
+        progress_bar: Optional[str] = _DEFAULT,
```
tswast: Same here, `| None` and potentially `Literal` as well.

tswast (commented):
Please sync with main so that the tests can re-run.

packages/bigframes/bigframes/core/events.py
```diff
@@ -121,19 +123,26 @@ class BigQuerySentEvent(ExecutionRunning):
...
     query: str
-    billing_project: Optional[str] = None
-    location: Optional[str] = None
-    job_id: Optional[str] = None
-    request_id: Optional[str] = None
+    billing_project: str | None = None
+    location: str | None = None
+    job_id: str | None = None
+    request_id: str | None = None
+    progress_bar: Literal["default", "auto", "notebook", "terminal"] | None = _DEFAULT
```
tswast: "progress_bar" isn't really state about the event, is it? Could you describe to me again what this is trying to solve?

Consider an alternative where the progress bar type is attached to the subscriber, instead, as that seems to be more appropriate. Or another alternative in which we wrap the events with some sort of "CellState" that gives an indicator of the current execution context, of which the progress_bar could be a property of that.

shuoweil: Hi Tim, thanks for the suggestions.

I looked into the first alternative (attaching the progress bar type to the subscriber), but it doesn't quite work for what we are trying to solve. The progress_bar option is thread-local and needs to be captured from the initiating thread at the moment the query starts. Since subscribers are typically long-lived (session-scoped), they cannot easily capture this per-query context. Furthermore, the callbacks from the BigQuery client library run in a background thread where the initiating thread's local storage is not accessible.

Therefore, I took your second alternative: I removed progress_bar from the event classes and instead wrapped the events in an EventEnvelope (acting as the execution context/'CellState' you suggested). This envelope is created in the callback (which captures the initiating thread's option via a closure) and carries the progress_bar setting along with the event to the subscribers.

This keeps the events clean of display preferences while solving the cross-thread propagation issue. Thanks a lot.

packages/bigframes/bigframes/session/metrics.py
```diff
@@ -236,13 +236,18 @@ def count_job_stats(
...
 
-    def on_event(self, event: Any):
+    def on_event(self, envelope: Any):
         try:
             import bigframes.core.events
             from bigframes.session.executor import LocalExecuteResult
         except ImportError:
             return
 
+        if isinstance(envelope, bigframes.core.events.EventEnvelope):
```
GarrettWu: Can it not be EventEnvelope type?

packages/bigframes/bigframes/formatting_helpers.py
```diff
@@ -147,12 +157,21 @@ def progress_callback(
...
     except ImportError:
-        # Since this gets called from __del__, skip if the import fails to avoid
+        # Since this gets called from __del__, skip if the import fails
+        # to avoid
         # ImportError: sys.meta_path is None, Python is likely shutting down.
         # This will allow cleanup to continue.
         return
 
-    progress_bar = bigframes._config.options.display.progress_bar
+    if isinstance(envelope, bigframes.core.events.EventEnvelope):
```
GarrettWu: same

packages/bigframes/bigframes/session/metrics.py
shuoweil: Good point! Having a consistent API contract for subscribers is definitely cleaner and removes the boilerplate isinstance checks.

I've updated Publisher.publish to automatically wrap any raw Event into an EventEnvelope. Now all subscribers consistently receive an EventEnvelope by default.

GarrettWu (approved):

tswast:
FYI: This may cause some conflicts with https://github.com/googleapis/google-cloud-python/pull/17093 May want to investigate if the ExecutionSpec object could be used similarly.

shuoweil:
> FYI: This may cause some conflicts with #17093 May want to investigate if the ExecutionSpec object could be used similarly.

PR #17093 is merged and all conflicts are fully resolved.

We investigated using ExecutionSpec but determined EventEnvelope is necessary for event publishing. Several lifecycle events (like SessionClosed) occur outside of query execution plans, and threading ExecutionSpec through every event emitter across the codebase would tightly couple publishing to executor internals.

Our implementation now seamlessly combines both patterns: EventEnvelope dynamically enforces thread-local display options for multi-threaded workloads, while falling back to #17093's session-level closure for single-threaded flows.

PR #16720, opened 2026-04-20
feat(auth): Add blocking Regional Access Boundary Lookup and Seed Support

macastelaz:
In order for the gcloud CLI to support Regional Access Boundary, the Python auth SDK needs to support blocking lookups as well as allowing an initial seed RAB to be provided (gcloud will set this seed if the CLI has a locally cached valid RAB available).

Additional details can be found at [go/rab-python-gcloud-one-pager](https://docs.google.com/document/d/1PvwqXp-jznleIpA8UzXU9of6sMvHPZA8HN_TSSsWFcc/edit?resourcekey=0-yQowO9bsBsHdv_60zQsKnA&tab=t.0)

### DIFF
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -226,7 +226,9 @@ class _RegionalAccessBoundaryManager(object):
 
             if regional_access_boundary_info:
                 # On success, update the boundary and its expiry, and clear any cooldown.
-                encoded_locations = regional_access_boundary_info.get("encodedLocations")
+                encoded_locations = regional_access_boundary_info.get(
+                    "encodedLocations"
+                )
                 updated_data = _RegionalAccessBoundaryData(
                     encoded_locations=encoded_locations,
                     expiry=_helpers.utcnow() + DEFAULT_REGIONAL_ACCESS_BOUNDARY_TTL,
@@ -242,7 +244,9 @@ class _RegionalAccessBoundaryManager(object):
                         "Regional Access Boundary lookup failed. Entering cooldown."
                     )
 
-                next_cooldown_expiry = _helpers.utcnow() + current_data.cooldown_duration
+                next_cooldown_expiry = (
+                    _helpers.utcnow() + current_data.cooldown_duration
+                )
                 next_cooldown_duration = min(
                     current_data.cooldown_duration * 2,
                     MAX_REGIONAL_ACCESS_BOUNDARY_COOLDOWN,
@@ -266,8 +270,6 @@ class _RegionalAccessBoundaryManager(object):
 
             # Perform the atomic swap of the state object.
             self._data = updated_data
-
-
 
 
 class _RegionalAccessBoundaryRefreshThread(threading.Thread):
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -110,6 +110,16 @@ class _RegionalAccessBoundaryManager(object):
         self.__dict__.update(state)
         self._update_lock = threading.Lock()
 
+    def __eq__(self, other):
+        """Checks if two managers are equal."""
+        if not isinstance(other, _RegionalAccessBoundaryManager):
+            return NotImplemented
+        return (
+            self._data == other._data
+            and self._use_blocking_regional_access_boundary_lookup
+            == other._use_blocking_regional_access_boundary_lookup
+        )
+
     def use_blocking_regional_access_boundary_lookup(self):
         """Enables blocking regional access boundary lookup to true"""
         self._use_blocking_regional_access_boundary_lookup = True
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -110,15 +110,6 @@ class _RegionalAccessBoundaryManager(object):
         self.__dict__.update(state)
         self._update_lock = threading.Lock()
 
-    def __eq__(self, other):
-        if not isinstance(other, _RegionalAccessBoundaryManager):
-            return False
-        return (
-            self._data == other._data
-            and self.refresh_manager == other.refresh_manager
-            and self._use_blocking_regional_access_boundary_lookup
-            == other._use_blocking_regional_access_boundary_lookup
-        )
 
     def use_blocking_regional_access_boundary_lookup(self):
         """Enables blocking regional access boundary lookup to true"""
@@ -332,12 +323,6 @@ class _RegionalAccessBoundaryRefreshManager(object):
         self.__dict__.update(state)
         self._lock = threading.Lock()
         self._worker = None
-
-    def __eq__(self, other):
-        if not isinstance(other, _RegionalAccessBoundaryRefreshManager):
-            return False
-        # Note: We only compare public/pickled properties.
-        return self._worker == other._worker
 
     def start_refresh(self, credentials, request, rab_manager):
         """
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -120,22 +120,29 @@ class _RegionalAccessBoundaryManager(object):
             == other._use_blocking_regional_access_boundary_lookup
         )
 
-    def use_blocking_regional_access_boundary_lookup(self):
-        """Enables blocking regional access boundary lookup to true"""
+    def enable_blocking_lookup(self):
+        """Enables blocking Regional Access Boundary lookup.
+
+        When enabled, the Regional Access Boundary lookup will be performed
+        synchronously in the calling thread instead of asynchronously in a
+        background thread.
+        """
         self._use_blocking_regional_access_boundary_lookup = True
 
-    def set_initial_regional_access_boundary(self, seed):
-        """Manually sets the regional access boundary to the client provided seed
+    def set_initial_regional_access_boundary(self, encoded_locations=None, expiry=None):
+        """Manually sets the regional access boundary to the client provided seed.
 
         Args:
-            seed (Mapping[str, str]): The regional access boundary to use for the
-                credential. This should be a map with, at a minimum, an "encodedLocations"
-                key that maps to a hex string and an "expiry" key which maps to a
-                datetime.datetime.
+            encoded_locations (Optional[str]): The encoded locations string.
+            expiry (Optional[datetime.datetime]): The expiry time for the boundary.
+                If encoded_locations is not provided, expiry is ignored.
         """
+        if not encoded_locations:
+            expiry = None
+
         self._data = _RegionalAccessBoundaryData(
-            encoded_locations=seed.get("encodedLocations", None),
-            expiry=seed.get("expiry", None),
+            encoded_locations=encoded_locations,
+            expiry=expiry,
             cooldown_expiry=None,
             cooldown_duration=DEFAULT_REGIONAL_ACCESS_BOUNDARY_COOLDOWN,
         )
@@ -190,17 +197,20 @@ class _RegionalAccessBoundaryManager(object):
     def start_blocking_refresh(self, credentials, request):
         """Initiates a blocking lookup of the Regional Access Boundary.
 
+        If the lookup raises an exception, it is caught and logged as a warning,
+        and the lookup is treated as a failure (entering cooldown). Exceptions
+        are not propagated to the caller.
+
         Args:
             credentials (google.auth.credentials.Credentials): The credentials to refresh.
             request (google.auth.transport.Request): The object used to make HTTP requests.
         """
         try:
-            # A blocking parameter is passed here to indicate this is a blocking lookup,
-            # which in turn will do two things: 1) set a timeout to 3s instead of the
-            # default 120s and 2) ensure we do not retry at all
-            blocking = True
+            # The fail_fast parameter is set to True to ensure we don't block the calling
+            # thread for too long. This will do two things: 1) set a timeout to 3s
+            # instead of the default 120s and 2) ensure we do not retry at all
             regional_access_boundary_info = (
-                credentials._lookup_regional_access_boundary(request, blocking)
+                credentials._lookup_regional_access_boundary(request, fail_fast=True)
             )
         except Exception as e:
             if _helpers.is_logging_enabled(_LOGGER):
@@ -275,7 +285,12 @@ class _RegionalAccessBoundaryManager(object):
 class _RegionalAccessBoundaryRefreshThread(threading.Thread):
     """Thread for background refreshing of the Regional Access Boundary."""
 
-    def __init__(self, credentials, request, rab_manager):
+    def __init__(
+        self,
+        credentials: "google.auth.credentials.CredentialsWithRegionalAccessBoundary",  # noqa: F821
+        request: "google.auth.transport.Request",  # noqa: F821
+        rab_manager: "_RegionalAccessBoundaryManager",
+    ):
         super().__init__()
         self.daemon = True
         self._credentials = credentials
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -97,6 +97,7 @@ class _RegionalAccessBoundaryManager(object):
         )
         self.refresh_manager = _RegionalAccessBoundaryRefreshManager()
         self._update_lock = threading.Lock()
+        self._use_blocking_regional_access_boundary_lookup = False
 
     def __getstate__(self):
         """Pickle helper that serializes the _update_lock attribute."""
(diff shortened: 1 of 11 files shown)


packages/google-auth/google/auth/credentials.py
```diff
@@ -361,6 +361,32 @@ def _copy_regional_access_boundary_manager(self, target):
...
+
+
+        Returns:
+            google.auth.credentials.Credentials: A new credentials instance.
+        """
+        creds = self._make_copy()
+        creds._rab_manager.set_initial_regional_access_boundary(seed)
+        return creds
+
+    def with_blocking_regional_access_boundary_lookup(self):
```
nbayati: Does gcloud need this method? 
If we have to have it, can we make it private? Since the blocking refresh was specifically added to support gcloud, I don't want users accidentally choosing a blocking lookup not knowing what it entails.

packages/google-auth/google/oauth2/credentials.py
```diff
@@ -202,6 +207,10 @@ def __setstate__(self, d):
         self._refresh_worker = None
         self._use_non_blocking_refresh = d.get("_use_non_blocking_refresh", False)
         self._account = d.get("_account", "")
+        self._rab_manager = d.get("_rab_manager") or (
+            _regional_access_boundary_utils._RegionalAccessBoundaryManager()
+        )
+        self._use_blocking_regional_access_boundary_lookup = False
```
nbayati: Do we need to be setting`_use_blocking_regional_access_boundary_lookup` here? 
It seems redundant because the default value inside the rab manager is already false, and this class is not even looking up the RAB data.

packages/google-auth/google/oauth2/credentials.py
```diff
@@ -353,8 +362,10 @@ def with_universe_domain(self, universe_domain):
     def _metric_header_for_usage(self):
         return metrics.CRED_TYPE_USER
 
-    @_helpers.copy_docstring(credentials.Credentials)
-    def refresh(self, request):
+    def _build_regional_access_boundary_lookup_url(self, request=None):
```
nbayati, line 364: I'm wondering if it's worth documenting why this method is just returning None and that this wasn't an oversight.

packages/google-auth/tests/oauth2/test__client.py
```diff
@@ -777,5 +783,30 @@ def test_lookup_regional_access_boundary_with_headers():
     )
 
     mock_request.assert_called_once_with(
-        method="GET", url="http://example.com", headers=headers
+        method="GET", url="http://example.com", headers=headers, timeout=None
+    )
+
+
+def test_lookup_regional_access_boundary_blocking():
```
nbayati, line 790: Can we add a failure test scenario that asserts the request was not retried after the failure?

packages/google-auth/tests/test__regional_access_boundary_utils.py
```diff
@@ -293,59 +311,99 @@ def test_maybe_start_refresh_handles_url_parse_errors(
...
-            request, "http://rab.example.com", headers={"authorization": "Bearer token"}
+        worker = _regional_access_boundary_utils._RegionalAccessBoundaryRefreshThread(
+            creds, request, rab_manager
         )
-        assert result is None
+        worker.run()
 
-    @mock.patch("google.oauth2._client._lookup_regional_access_boundary")
-    @mock.patch.object(CredentialsImpl, "_build_regional_access_boundary_lookup_url")
-    def test_lookup_regional_access_boundary_null_url(
```
nbayati, line 336: It looks like test_lookup_regional_access_boundary_null_url was removed. Since the new OAuth2 implementation relies on returning None to skip lookups, should we  restore this test to ensure it works as intended?

packages/google-auth/google/auth/credentials.py
```diff
@@ -421,11 +447,16 @@ def before_request(self, request, method, url, headers):
         """Refreshes the access token and triggers the Regional Access Boundary
         lookup if necessary.
         """
-        super(CredentialsWithRegionalAccessBoundary, self).before_request(
```
nbayati, line 424: Since we removed the call to super().before_request in and duplicated the logic to control the execution order, it would be good to add a test in test_credentials.py to verify that before_request correctly triggers the RAB refresh flow in sequence with the token refresh. 

a small, unrelated nit: I noticed that in test_before_request, the arguments for url and method are swapped [in the call to before_request](https://github.com/googleapis/google-cloud-python/blob/03613639c5dbacdfe999ae5add49615673f4df39/packages/google-auth/tests/test_credentials.py#L110). It doesn't cause failures because they aren't used by the base class. If you end up adding the unit test to this file and want to do a quick cleanup, you could swap them back to the correct order!

nbayati (commented):
Overall looks really good! Just a few small comments and questions.

packages/google-auth/google/auth/credentials.py
macastelaz: Makes sense to make it private and document, similarly to the seed method, that this is used by gcloud and changes must be made in a backwards compatible way.

packages/google-auth/google/auth/credentials.py
macastelaz: Good call! Done

packages/google-auth/google/oauth2/credentials.py
macastelaz: Good call - removed!

packages/google-auth/google/oauth2/credentials.py
macastelaz: Added docs here but feel free to provide additional feedback if you'd like to see it reworded at all.

packages/google-auth/tests/oauth2/test__client.py
macastelaz: Good call - done!

packages/google-auth/tests/test__regional_access_boundary_utils.py
macastelaz: Done

packages/google-auth/tests/test__regional_access_boundary_utils.py
```diff
@@ -405,6 +406,12 @@ def test_lookup_regional_access_boundary_failure(self, mock_lookup_rab):
         assert rab_manager._data.expiry is None
         assert rab_manager._data.cooldown_expiry is not None
 
+    def test_lookup_regional_access_boundary_null_url(self):
+        creds = oauth2_credentials.Credentials(token="token")
+        request = mock.Mock()
+        result = creds._lookup_regional_access_boundary(request)
```
nbayati, line 412: nit: can we add `request.assert_not_called()` to ensure that there was no request made?

nbayati (approved):
Great job!

packages/google-auth/tests/test__regional_access_boundary_utils.py
macastelaz: Done!

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -109,6 +110,36 @@ def __setstate__(self, state):
...
+            return NotImplemented
+        return (
+            self._data == other._data
+            and self._use_blocking_regional_access_boundary_lookup
+            == other._use_blocking_regional_access_boundary_lookup
+        )
+
+    def use_blocking_regional_access_boundary_lookup(self):
+        """Enables blocking regional access boundary lookup to true"""
+        self._use_blocking_regional_access_boundary_lookup = True
```
daniel-sanche, line 134: I think this naming could get very confusing, since this is a state-changing function with an (almost) identical name to an internal data attribute. And the docstring isn't helpful


Can we change the function name to something like `enable_blocking_lookup`? And make the docstring explain what happens when this is enabled?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -109,6 +110,36 @@ def __setstate__(self, state):
...
+
+    def set_initial_regional_access_boundary(self, seed):
+        """Manually sets the regional access boundary to the client provided seed
+
+        Args:
+            seed (Mapping[str, str]): The regional access boundary to use for the
+                credential. This should be a map with, at a minimum, an "encodedLocations"
+                key that maps to a hex string and an "expiry" key which maps to a
+                datetime.datetime.
```
daniel-sanche: - the docstring makes it sound like these fields required, but the code seems to work fine if they're not
- why are we using a map for this? why not make `encoded_locations` and `expiry` arguments?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -151,48 +182,47 @@ def maybe_start_refresh(self, credentials, request):
...
-        # background thread from crashing. This ensures we can gracefully enter
-        # an exponential cooldown state on failure.
         try:
+            # A blocking parameter is passed here to indicate this is a blocking lookup,
+            # which in turn will do two things: 1) set a timeout to 3s instead of the
+            # default 120s and 2) ensure we do not retry at all
+            blocking = True
             regional_access_boundary_info = (
-                self._credentials._lookup_regional_access_boundary(self._request)
+                credentials._lookup_regional_access_boundary(request, blocking)
```
daniel-sanche: nit: why does blocking need to be declared as a variable? Can you do `credentials._lookup_regional_access_boundary(request, blocking=True)`?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -151,48 +182,47 @@ def maybe_start_refresh(self, credentials, request):
...
+        if self._use_blocking_regional_access_boundary_lookup:
+            self.start_blocking_refresh(credentials, request)
+        else:
+            self.refresh_manager.start_refresh(credentials, request, self)
 
-    def run(self):
-        """
-        Performs the Regional Access Boundary lookup and updates the state.
+    def start_blocking_refresh(self, credentials, request):
+        """Initiates a blocking lookup of the Regional Access Boundary.
```
daniel-sanche, line 202: nit: the docstring should probably mention what happens on an exception

packages/google-auth/google/auth/credentials.py
```diff
@@ -435,13 +471,16 @@ def refresh(self, request):
...
+        self,
+        request: "google.auth.transport.Request",  # noqa: F821
+        blocking: bool = False,
     ) -> "Optional[Dict[str, str]]":
         """Calls the Regional Access Boundary lookup API to retrieve the Regional Access Boundary information.
 
         Args:
             request (google.auth.transport.Request): The object used to make
                 HTTP requests.
+            blocking (bool): Whether the lookup should be blocking.
```
daniel-sanche: nit: more context would be helpful (e.g. "otherwise, the lookup will be preformed in a background thread", or something)

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -241,7 +269,48 @@ def run(self):
...
 
             # Perform the atomic swap of the state object.
-            self._rab_manager._data = updated_data
+            self._data = updated_data
+
+
+class _RegionalAccessBoundaryRefreshThread(threading.Thread):
+    """Thread for background refreshing of the Regional Access Boundary."""
+
+    def __init__(self, credentials, request, rab_manager):
```
daniel-sanche: nit: typing could be helpful here. (Does this only work with CredentialsWithRegionalAccessBoundary  credentials?)

packages/google-auth/google/auth/credentials.py
```diff
@@ -421,11 +452,16 @@ def before_request(self, request, method, url, headers):
         """Refreshes the access token and triggers the Regional Access Boundary
         lookup if necessary.
         """
-        super(CredentialsWithRegionalAccessBoundary, self).before_request(
-            request, method, url, headers
-        )
+        if self._use_non_blocking_refresh:
+            self._non_blocking_refresh(request)
+        else:
+            self._blocking_refresh(request)
```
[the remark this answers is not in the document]
daniel-sanche: This is worth confirming, but from my understanding, I'm not sure if this is an issue. CredentialsWithRegionalAccessBoundary inherits from Credentials, so it should have those internal attributes.

It does seem a little dangerous that before_request seems to be mostly copy/pasted from the superclass here though. I wonder if that can be cleaned up somehow? I could see these drifting over time

packages/google-auth/google/oauth2/_client.py
```diff
@@ -600,6 +603,7 @@ def _lookup_regional_access_boundary_request_no_throw(
         url (str): The Regional Access Boundary lookup url.
         can_retry (bool): Enable or disable request retry behavior. Defaults to true.
         headers (Optional[Mapping[str, str]]): The headers for the request.
+        blocking (bool): Whether the lookup should be blocking.
```
daniel-sanche: Doesn't it block in all cases here? It seems like "blocking" in this case just controls how timeout/retry configuration?

daniel-sanche: I'm pretty confused by this, because it seems like it uses a much lower timeout and retry configuration when blocking is enabled, which is pretty much the opposite from what I would expect (1x 3 second attempt for blocking, vs 6x 120 second attempts for non-blocking)

packages/google-auth/google/oauth2/credentials.py
```diff
@@ -353,8 +361,19 @@ def with_universe_domain(self, universe_domain):
...
+        """Builds the URL for Regional Access Boundary lookup.
+
+        OAuth 2.0 credentials do not support independent Regional Access Boundary
+        lookup. However, they may support a seeded Regional Access Boundary
+        provided externally (e.g., from gcloud).
+
+        Returns:
+            None: This credential type does not support RAB lookup.
+        """
+        return None
```
daniel-sanche, line 374: If this always returns None, it looks like this could result in a lot of failure logs [here](https://github.com/macastelaz/google-cloud-python/blob/d4ab1eb54b3bc8fce68719df53ed6d3fc134803f/packages/google-auth/google/auth/credentials.py#L490). In that context, None seems to represent a failure

Should this class overrride _is_regional_access_boundary_lookup_required to disable lookup? Or am I misunderstanding this?

daniel-sanche (commented):
a bunch of small comments, and a few that are worth looking at closer. I know there's a bit of time pressure on this, so you can ignore some of the nits if needed

packages/google-auth/google/auth/credentials.py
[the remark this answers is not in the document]
nbayati: Yes I think Gemini's assessment is incorrect. We had to duplicate the parent's logic here because we needed to ensure that the RAB lookup (which might be blocking) happens after the access token is refreshed but before headers are applied to the request. Credentials.before_request does both (refresh and apply) in one go.

I agree with the concern that they might drift over time. I think this can be refactored in the future and extract the common logic into a helper method. Would you be ok with us merging it as-is now, and doing the refactoring in the future?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: Both are very valid. I made the fields arguments here, but kept the map for `_with_regional_access_boundary(self, seed)` since this is the method gcloud cli will be using.

packages/google-auth/google/auth/credentials.py
nbayati: The thread is spawned and managed by the caller (specifically by the _RegionalAccessBoundaryRefreshManager), and setting blocking here to true or false doesn't affect if it's run on the background or current thread.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: Yes, it only work with CredentialsWithRegionalAccessBoundary credential. I'm adding string literal type hints to avoid making a circular dependency.

packages/google-auth/google/oauth2/_client.py
nbayati: Thanks for pointing this out. I think the parameter name was a bit misleading. 

The shorter timeout and no retry was an intentional choice because on gcloud has to use the blocking mode, which blocks the user's foreground action, and we want to fail fast (3s timeout, no retries) to avoid freezing the terminal for a long time if the endpoint is slow or down. The RAB fetching is a best effort operation, where we try to get the header to improve the call's routing and latency, but if it fails or is taking a long time, we'd want to let the call to go through without further delays. 

I've renamed the parameter in the internal lookup methods from blocking to fail_fast to better describe the behavior, and updated the docstrings in credentials.py and _client.py to explicitly mention that fail_fast=True uses a short timeout and no retries.

packages/google-auth/google/oauth2/credentials.py
nbayati: Great catch! Thank you for flagging this and the suggestion! I've added the overrride to _is_regional_access_boundary_lookup_required to return False. This ensures we don't attempt to call _build_regional_access_boundary_lookup_url, and would resolve the failures being logged.

daniel-sanche (approved):
LGTM

PR #16731, opened 2026-04-20
fix: propagate quota_project_id and api_endpoint in AsyncGrpcClient

chandra-siri:
This PR fixes bug 503990490 where AsyncGrpcClient fails to propagate quota_project_id and api_endpoint from ClientOptions to the gRPC channel.

### DIFF
--- a/packages/google-cloud-storage/google/cloud/storage/asyncio/async_grpc_client.py
+++ b/packages/google-cloud-storage/google/cloud/storage/asyncio/async_grpc_client.py
@@ -23,6 +23,8 @@ from google.cloud._storage_v2.services.storage.transports.base import (
 )
 from google.cloud.storage import __version__
 
+_DEFAULT_HOST = "storage.googleapis.com"
+
 
 class AsyncGrpcClient:
     """An asynchronous client for interacting with Google Cloud Storage using the gRPC API.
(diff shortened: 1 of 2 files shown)


packages/google-cloud-storage/google/cloud/storage/asyncio/async_grpc_client.py
```diff
@@ -109,7 +111,15 @@ def _create_async_grpc_client(
 
         primary_user_agent = client_info.to_user_agent()
 
+        host = _DEFAULT_HOST
+        quota_project_id = None
+        if client_options:
+            host = getattr(client_options, "api_endpoint", None) or _DEFAULT_HOST
+            quota_project_id = getattr(client_options, "quota_project_id", None)
```
Mahalaxmibejugam, line 118: Are there any other attributes from ClientOptions that need to be passed? I've checked and noticed that api_audience might be ignored if the transport is provided. Can you please check if this is also required?

chandra-siri: fix for other attributes will be done in another PR as & when required.

Mahalaxmibejugam:
Is the x-goog-user-project header getting populated with these changes?

Mahalaxmibejugam (approved):

PR #16746, opened 2026-04-21
chore: specify relative path for packages in pandas-gbq noxfile.py

ohmayr:
Use relative paths instead of cloning the entire repo from github and stepping into the sub directory which takes a lot of memory and time and results in the presubmit(s) to timeout.


chalmerlowe (approved):
LGTM

PR #16747, opened 2026-04-21
chore: remove redundant keep list entries

### LINKED ISSUE #4759: python: reduce size of keep lists
The keep lists from the initial migration are ported from the preserve lists in legacylibrarian. Not all files need to be in the keep list though - if we wouldn't delete the file anyway, it doesn't need to be in the keep list.

Example query to find long keep lists:

```sh
yq '.libraries | map(select((.keep | length) > 10)) | .[] | .name' librarian.yaml
```

jskeet:
This change was tested by running `librarian generate --all`, as well as checking for custom keep lists for packages which have yet to be fully migrated to librarian generation.

Fixes https://github.com/googleapis/librarian/issues/4759

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -356,9 +356,9 @@ libraries:
       - path: google/cloud/alloydb/connectors/v1
       - path: google/cloud/alloydb/connectors/v1beta
       - path: google/cloud/alloydb/connectors/v1alpha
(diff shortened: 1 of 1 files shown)


jskeet:
I've just seen the conflict - will fix.

parthea (approved):

jskeet:
Rebased and force-pushed - the conflicts were simple to resolve.

parthea (approved):

PR #16751, opened 2026-04-21
chore: regenerate compute packages using gRPC service config

jskeet:
This is the result of locally applying a fix for https://github.com/googleapis/librarian/issues/3827 and regenerating the two packages.

After merging this, I'll get the change merged into Librarian as well, tag it, and update the version in librarian.yaml.


jskeet:
Given that this is green, I'd now actually like to merge this if possible, to avoid getting a backlog of changes which will all end up conflicting etc. (Noted in internal chat.)

parthea (approved):

PR #16752, opened 2026-04-21
chore: add core-deps-from-source nox session to assorted packages

chalmerlowe:
This PR adds the core_deps_from_source session to bigframes to test against core dependencies installed from source in the monorepo.


parthea:

chalmerlowe:
@gemini-code-assist

This comment is in error:

> The repository URLs for the core dependencies are incorrect. Packages like google-auth, google-api-core, and proto-plus are hosted in their own dedicated repositories under the googleapis organization, not within the google-cloud-python repository. Attempting to install them from google-cloud-python with the specified subdirectories will fail. The suggested code uses the correct repository URLs.

Each of these has been migrated to the google-cloud-python repository as individual packages in the packages directory. 

Please feel free to confirm this and make a note for yourself so that this comment does not get raised other PRs.

Thanks.

packages/bigframes/noxfile.py
```diff
@@ -945,6 +946,7 @@ def core_deps_from_source(session, protobuf_implementation):
     rather than pulling the dependencies from PyPI.
     """
 
+    # Trigger CI run for core_deps_from_source
```
chalmerlowe:
```suggestion
```

parthea (approved):

PR #16756, opened 2026-04-21
chore: sort resource messages for deterministic generation

ohmayr:
sort resource messages for deterministic generation.

Towards b/505425328

### DIFF
--- a/packages/gapic-generator/gapic/schema/api.py
+++ b/packages/gapic-generator/gapic/schema/api.py
@@ -170,9 +170,12 @@ class Proto:
             if msg.options.Extensions[resource_pb2.resource].type
         )
         return collections.OrderedDict(
-            itertools.chain(
-                file_resource_messages,
-                resource_messages,
+            sorted(
+                itertools.chain(
+                    file_resource_messages,
+                    resource_messages,
+                ),
+                key=lambda item: item[0]
             )
         )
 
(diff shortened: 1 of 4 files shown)


packages/gapic-generator/gapic/schema/wrappers.py
```diff
@@ -2316,6 +2316,13 @@ def gen_indirect_resources_used(message):
             )
         )
 
+        sorted_messages = sorted(
```
parthea, line 2321: nit: add a comment. Something like `Convert the set to a sorted tuple using the resource path or message name. This is needed to prevent non-deterministic code generation.`

packages/gapic-generator/gapic/schema/api.py
```diff
@@ -170,9 +170,12 @@ def resource_messages(self) -> Mapping[str, wrappers.MessageType]:
             if msg.options.Extensions[resource_pb2.resource].type
         )
         return collections.OrderedDict(
-            itertools.chain(
-                file_resource_messages,
-                resource_messages,
+            sorted(
```
parthea, line 176: nit: add a comment to explain why this is sorted

parthea (approved):

packages/gapic-generator/gapic/schema/api.py
ohmayr: addressed.

PR #16757, opened 2026-04-21
refactor(bigframes): update SQLGlot compiler to process AI func params better

sycai:


sycai:
Test failure is unrelated.

packages/bigframes/bigframes/core/compile/sqlglot/expressions/ai_ops.py
```diff
@@ -105,44 +99,33 @@ def _construct_named_args(op: ops.NaryOp) -> list[sge.Kwarg]:
...
-    if output_schema is not None:
-        args.append(
-            sge.Kwarg(
-                this="output_schema",
-                expression=sge.Literal.string(output_schema),
+        elif field == "request_type":
+            args.append(
+                sge.Kwarg(this=field, expression=sge.Literal.string(value.upper()))
+            )
+        else:
```
chelsea-lin, line 126: Do we want to check for the field is "output_schema" explicitly?  This would allow us to decide how to handle unknown fields - either failing fast, ignoring them, or similarly to "output_schema" - which is better for forward compatibility.

chelsea-lin (approved):

packages/bigframes/bigframes/core/compile/sqlglot/expressions/ai_ops.py
sycai: I think we should be good without checking the "output_schema", as the intention of this PR is to make the compiler able to handle all kinds of fields without explicit mentions.

PR #16759, opened 2026-04-21
feat(bigframes): implement ai.embed

sycai:
Fixes b/497836685 🦕


sycai:
Pre-release tests failed at `divmod` cases, which is unrelated to this change.

packages/bigframes/bigframes/bigquery/_operations/ai.py
```diff
@@ -705,6 +705,113 @@ def generate_table(
         return session.read_gbq_query(query)
 
 
+@log_adapter.method_logger(custom_base_name="bigquery_ai")
+def embed(
+    content: str | series.Series | pd.Series,
```
TrevorBergeron, line 710: Don't need to in this PR, but would nice to get all these working with the pd.col expression interface, and maybe (single-level) index values as well? Maybe we define "column-value" types and a common adapter, type annotation

TrevorBergeron (approved):

packages/bigframes/bigframes/bigquery/_operations/ai.py
sycai: Sounds good to me. Opened b/507096491 to track this FR

PR #16760, opened 2026-04-22
feat(bigtable): client side metrics handlers

daniel-sanche:
Migrate https://github.com/googleapis/python-bigtable/pull/1189 to the monorepo

This PR builds off of https://github.com/googleapis/python-bigtable/pull/1187 to add handlers to the client-side metrics system, which can subscribe to the metrics stream, and export the results into different collection systems

We add two handlers to the system:
- `GoogleCloudMetricsHandler`: sends metrics to a private OpenTelemetry meter, and then periodically exports them to GCP. Built on top of `OpenTelemetryMetricsHandler`
- `OpenTelemetryMetricsHandler`: sends metrics to the root MeterProvider, so the user can access the exported metrics for their own systems. This will be off by default, but can be added alongside `GoogleCloudMetricsHandler` if needed


packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    for data_point in [
+                        pt for pt in metric.data.data_points if pt.attributes
+                    ]:
+                        if data_point.attributes:
+                            monitored_resource = MonitoredResource(
+                                type="bigtable_client_raw",
+                                labels={
+                                    "project_id": self.project_id,
```
mutianf: This project id should be extracted from the project_id attributes of the data point.

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+        exporter: The exporter object used to write metrics to Cloud Montitoring.
+            Should correspond 1:1 with a bigtable client, and share auth configuration
+        export_interval: The interval (in seconds) at which to export metrics to Cloud Monitoring.
+        *args: configuration positional arguments passed down to super class
+        *kwargs: configuration keyword arguments passed down to super class
+    """
+
+    def __init__(self, exporter, *args, export_interval=60, **kwargs):
+        # periodically executes exporter
+        gcp_reader = PeriodicExportingMetricReader(
```
mutianf, line 144: is this initialized per table  and app profile ? Can we move this to client level so it's shared across different tables and app profiles?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,237 @@
...
+        client_version: str | None = None,
+        instruments: _OpenTelemetryInstruments = _OpenTelemetryInstruments(),
+    ):
+        super().__init__()
+        self.otel = instruments
+        client_version = client_version or bigtable_version
+        # fixed labels sent with each metric update
+        self.shared_labels = {
+            "client_name": f"python-bigtable/{client_version}",
+            "client_uid": client_uid or self._generate_client_uid(),
```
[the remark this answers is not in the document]
mutianf: _generate_client_uid has staticmethod annotation so I don't know if this is true. But we should have 1 client_uid per client instance.

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+
+from google.cloud.bigtable.data._metrics.handlers.opentelemetry import (
+    OpenTelemetryMetricsHandler,
+    _OpenTelemetryInstruments,
+)
+
+# create OpenTelemetry views for Bigtable metrics
+# avoid reformatting into individual lines
+# fmt: off
+MILLIS_AGGREGATION = view.ExplicitBucketHistogramAggregation(
```
mutianf, line 57: Make sure the buckets are in sync with the java buckets: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/metrics/Constants.java#L86

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
```diff
@@ -260,6 +264,12 @@ def __init__(
                 "is the default."
             )
         self._is_closed = CrossSync.Event()
+        # create a metrics exporter using the same client configuration
+        self._gcp_metrics_exporter = BigtableMetricsExporter(
+            project_id=self.project,
```
mutianf: when exporting I think we can get the project id from the metric data attribute instead of hard code it in the exporter.

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+    def __init__(self, exporter, *args, export_interval=60, **kwargs):
+        # periodically executes exporter
+        gcp_reader = PeriodicExportingMetricReader(
+            exporter, export_interval_millis=export_interval * 1000
+        )
+        # use private meter provider to store instruments and views
+        self.meter_provider = MeterProvider(
+            metric_readers=[gcp_reader], views=VIEW_LIST
+        )
+        otel = _OpenTelemetryInstruments(meter_provider=self.meter_provider)
```
mutianf, line 151: is this creating one otel instance per table? If so, this should also be moved up to the client level.

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+        Args:
+            series: list of TimeSeries to write. Will be split into batches if necessary
+            deadline: designates the time.time() at which to stop writing. If None, uses API default
+            max_batch_size: maximum number of time series to write at once.
+                Cloud Monitoring allows up to 200 per request
+        """
+        write_ind = 0
+        while write_ind < len(series):
+            # find time left for next batch
+            timeout = deadline - time.time() if deadline else gapic_v1.method.DEFAULT
```
mutianf: why do we need to calculate the remaining deadline for each batch? I think each batch should just use the default timeout.

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+                                    },
+                                ),
+                                unit=metric.unit,
+                            )
+                            all_series.append(series)
+        # send all metrics to Cloud Monitoring
+        try:
+            self._batch_write(all_series, deadline)
+            return MetricExportResult.SUCCESS
+        except Exception:
```
mutianf: I think we should add logging here

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: I changed project_id to be passed through with the attributes, like instances and tables

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Moved this to the client level

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Updated

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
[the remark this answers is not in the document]
daniel-sanche: Now that the handlers were moved to the client level, we should have one per client (previously it would have been one per table)

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: added logging, for this and a few other cases

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: The intention here is to be consistent with the OpenTelemetry [exporter spec](https://github.com/open-telemetry/opentelemetry-python/blob/6810d3d77e2690c122e497e5cf646373c10bc5c8/exporter/opentelemetry-exporter-otlp-proto-grpc/src/opentelemetry/exporter/otlp/proto/grpc/metric_exporter/__init__.py#L178), which passes in a timeout value for exporters to conform to. 

I'm not entirely sure what happens if we ignore it, but I can look into it if needed. Does Java not have anything similar?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_async/client.py
daniel-sanche: done

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,317 @@
...
+        self._exporter = exporter
+        # periodically executes exporter
+        gcp_reader = PeriodicExportingMetricReader(
+            exporter, export_interval_millis=export_interval * 1000
+        )
+        # use private meter provider to store instruments and views
+        self.meter_provider = MeterProvider(
+            metric_readers=[gcp_reader], views=VIEW_LIST
+        )
+        otel = _OpenTelemetryInstruments(meter_provider=self.meter_provider)
```
mutianf, line 151: Looks like this should support application has it's own otel instrumentation. Is there an example of how that can be done somewhere?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Args:
+        project_id: GCP project id to associate metrics with
+    """
+
+    def __init__(self, project_id: str, *client_args, **client_kwargs):
+        super().__init__()
+        self.client = MetricServiceClient(*client_args, **client_kwargs)
+        self.prefix = "bigtable.googleapis.com/internal/client"
+        self.project_id = project_id
```
mutianf: I don't think CloudMonitoring client needs to be created with a project id. The project id can be extracted from metrics label ?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,317 @@
...
+            timeout = (
+                max(0.0, deadline - time.monotonic())
+                if deadline
+                else gapic_v1.method.DEFAULT
+            )
+            # write next batch
+            batch = series[write_ind : write_ind + max_batch_size]
+            self.client.create_service_time_series(
+                CreateTimeSeriesRequest(
+                    name=f"projects/{self.project_id}",
```
mutianf: This project id should come from the metric label

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,317 @@
...
+            )
+            _LOGGER.debug(
+                "Successfully wrote batch of %d time series to projects/%s",
+                len(batch),
+                self.project_id,
+            )
+            write_ind += max_batch_size
+
+    @staticmethod
+    def _to_point(
```
mutianf, line 283: are the resouce level attributes (project, instance, table, cluster, zone) removed from the metric data points?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,317 @@
...
+    ) -> MetricExportResult:
+        """
+        Write a set of metrics to Cloud Monitoring.
+        This method is called by the OpenTelemetry SDK
+        """
+        deadline = time.monotonic() + (timeout_millis / 1000)
+        metric_kind = MetricDescriptor.MetricKind.CUMULATIVE
+        all_series: list[TimeSeries] = []
+        # process each metric from OTel format into Cloud Monitoring format
+        for resource_metric in metrics_data.resource_metrics:
```
mutianf, line 184: I assume one of these for loops is filtering out bigtable only metrics?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,249 @@
...
+        client_version: str | None = None,
+        instruments: _OpenTelemetryInstruments | None = None,
+    ):
+        super().__init__()
+        self.otel = instruments or _OpenTelemetryInstruments()
+        client_version = client_version or bigtable_version
+        # fixed labels sent with each metric update
+        self.shared_labels = {
+            "client_name": f"python-bigtable/{client_version}",
+            "client_uid": client_uid or self._generate_client_uid(),
```
mutianf, line 138: how is client_uid generated otherwise? maybe we should always generate it? This is really to prevent cloud monitoring to having issue merging the streams. maybe we should move it to the exporter?

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Yes, I designed the architecture to work well with user-provided otel meter providers, but so far I haven't actually exposed it. I thought we should wait for an ask before making it an official feature

If you want me to add this now, I can. But it might be better as a separate PR

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Yeah, I think I originally did it this way
1) because we know the expected project at client init-time
2) to be more in line with the [upstream OTel Exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/3668dfe7ce3b80dd01f42af72428de957b58b316/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py#L82)

But I don't think there's any reason we need to do it this way. I changed it to extract all project ids from the labels

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Yes, that was happening [here](https://github.com/googleapis/google-cloud-python/blob/92156ac1dada797f9cb00a8b3a84a518209938a0/packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py#L207). But I updated this code to be a bit cleaner in this area

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: For this handler, we're [using an isolated MeterProvider/instruments](https://github.com/googleapis/google-cloud-python/blob/92156ac1dada797f9cb00a8b3a84a518209938a0/packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py#L123) for bigtable metrics, so we don't actually expect to see any non-bigtable metrics come through.

(we made it so users can also interact with these metrics in their own OTel set-up if we want to expose that, but this pipeline is private)

daniel-sanche: But also, yes, there is some filtering here, skipping over data that doesn't have the project id set properly

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: > how is client_uid generated otherwise? maybe we should always generate it?

Currently, it's always auto-generated, unless explicitly passed in. I don't think we currently need to hard-code it anywhere (outside of possible testing), but if we [choose to expose this to end-users](https://github.com/googleapis/google-cloud-python/pull/16760#discussion_r3706678591), they may want to be able to control this field for their private metric systems

> maybe we should move it to the exporter?

This is where we're setting all the other labels, so it feels like a natural place for it. But yeah, if you want to expose these metrics to external OTel set-ups, and don't want them to see this field, we can move it to the GCP-specific subclass instead

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -226,13 +230,18 @@ def export(
             return MetricExportResult.FAILURE
 
     def _batch_write(
-        self, series: list[TimeSeries], deadline=None, max_batch_size=200
+        self,
+        project_id: str,
+        series: list[TimeSeries],
+        deadline=None,
```
mutianf, line 236: SHould we have a deadline for the write?

mutianf (approved):

packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: My interpretation is yes, we should follow a deadline, and we [pass it down here](https://github.com/googleapis/google-cloud-python/blob/92156ac1dada797f9cb00a8b3a84a518209938a0/packages/google-cloud-bigtable/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py#L224) (the None in the code above your comment is just the function default value)

The `MetricExporter` base class has a timeout_millis argument, so our timeout policy is just making sure we are following the expectations of the base class

But let me know if you think we need to handle anything differently

PR #16761, opened 2026-04-22
perf(storage): use google_crc32c.value() for simpler crc32c calculation

zhixiangli:
## Summary
- Simplified CRC32C calculation by using `google_crc32c.value()`.
- Improved code readability and reduced complexity in checksum logic.


chandra-siri (approved):
Thanks for this PR

PR #16762, opened 2026-04-22
chore: specify explicit options for all API paths in librarian.yaml

jskeet:
This means we don't use the generator default for any of:

- python-gapic-name
- python-gapic-namespace
- warehouse-package-name

Only API paths which didn't already have an explicit default were updated.

The changes are all from explicitly specifying `warehouse-package-name`, and are all corrections from where the generator was inferring the package name incorrectly. (And regenerating Datastore, which wasn't automated in #16737.)

This is the first step in https://github.com/googleapis/librarian/issues/5163 - we will then change librarian to use a new approach to inference (deriving the namespace just from the first two components of the API path), and remove all the explicit options which would be inferred the same way. So while this PR adds a lot of lines to the config, many will later be removed again.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -113,6 +113,11 @@ libraries:
       - path: google/ads/admanager/v1
     description_override: Manage your Ad Manager inventory, run reports and more.
     python:
+      opt_args_by_api:
+        google/ads/admanager/v1:
+          - python-gapic-namespace=google.ads
+          - python-gapic-name=admanager
+          - warehouse-package-name=google-ads-admanager
       default_version: v1
   - name: google-ads-datamanager
     version: 0.8.0
@@ -122,6 +127,11 @@ libraries:
       A unified ingestion API for data partners, agencies and advertisers to
       connect first-party data across Google advertising products.
     python:
+      opt_args_by_api:
+        google/ads/datamanager/v1:
+          - python-gapic-namespace=google.ads
+          - python-gapic-name=datamanager
+          - warehouse-package-name=google-ads-datamanager
       name_pretty_override: Data Manager API
       client_documentation_override: https://cloud.google.com/python/docs/reference/google-ads-datamanager/latest
       default_version: v1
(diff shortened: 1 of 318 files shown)


jskeet:
(Having just prototyped the "remove redundant options", the librarian.yaml file will *end up* being smaller than it is today...)

librarian.yaml
```diff
@@ -1481,12 +2302,21 @@ libraries:
       - path: google/cloud/documentai/v1
       - path: google/cloud/documentai/v1beta3
     description_override: Service to parse structured information from unstructured or semi-structured documents using state-of-the-art Google AI such as natural language, computer vision, translation, and AutoML.
+    keep:
+      - CHANGELOG.md
+      - docs/CHANGELOG.md
     python:
       opt_args_by_api:
         google/cloud/documentai/v1:
           - autogen-snippets
```
parthea, line 1900: Unrelated to the changes in this PR, in this file we have `autogen-snippets=False` and `autogen-snippets` without any `True`/`False` value. It seems that we should either have `autogen-snippets=False`, or not specify ``autogen-snippets` at all.

librarian.yaml
```diff
@@ -2376,11 +3709,16 @@ libraries:
...
       - path: google/cloud/privatecatalog/v1beta1
     description_override: allows developers and cloud admins to make their solutions discoverable to their internal enterprise users. Cloud admins can manage their solutions and ensure their users are always launching the latest versions.
+    keep:
+      - CHANGELOG.md
+      - docs/CHANGELOG.md
     python:
       opt_args_by_api:
         google/cloud/privatecatalog/v1beta1:
           - warehouse-package-name=google-cloud-private-catalog
           - autogen-snippets
```
parthea, line 3058: Same feedback as https://github.com/googleapis/google-cloud-python/pull/16762/changes#r3124724817

parthea (approved):

librarian.yaml
jskeet: Your comment actually highlighted that this change is reverting the "keep" changes from before... I'll need to fix that before we merge.

But the options are just copied from what we had in Bazel. Filed https://github.com/googleapis/librarian/issues/5525

jskeet:
Rebased on main so that we don't reinstate the keep lists we removed before. I'm locally running a `librarian generate --all` to check that other changes that have come in since creating this haven't caused any issues.

parthea (approved):

parthea (approved):

parthea:
https://github.com/googleapis/google-cloud-python/pull/16764 should fix the `prerelease` tests for `google-cloud-spanner`

jskeet:
Great - will rebase tomorrow UK time and merge when green (assuming the approval sticks)

jskeet:
@parthea: Looking at the Kokoro logs, I believe these are Spanner failures

jskeet:
Ah, I hadn't realised that #16764 hadn't been merged. That explains why the presubmits are still failing. Happy to rebase again as soon as that's merged.

parthea:
@jskeet Please could you rebase again to include the fix from https://github.com/googleapis/google-cloud-python/pull/16764 ?

jskeet:
@parthea: Done. Fingers crossed...

parthea:
`Kokoro Deps at HEAD Tests` failed due to a Github outage. Retrying test. https://www.githubstatus.com/

```
  remote: Internal Server Error
  fatal: unable to access 'https://github.com/googleapis/google-cloud-python/': The requested URL returned error: 500
  error: subprocess-exited-with-error
```

PR #16766, opened 2026-04-22
feat: setup React hybrid widget infrastructure

shuoweil:
Set up the build pipeline and file structure to support the React Hybrid Approach for TableWidget.

- Added `esbuild` via Python pip to allow bundling without requiring a Node.js environment.
- Moved the initial React implementation to `bigframes/display/src/index.js` and fixed relative imports.
- Generated the bundled asset `bigframes/display/table_widget.esm.js`.
- Updated `anywidget.py` to load the new bundled ESM file instead of the vanilla JS file.

This prepares the codebase for the core rewrite of the widget functionality in React.

Fixes #<505414691> 🦕


shuoweil:
We plan to implement angular based framework instead of REACT.

PR #16776, opened 2026-04-23
chore: regenerate all libraries while ignoring overrides

jskeet:
Creating as a draft with "do not merge"; this is only for review purposes.


jskeet:
Note: when we do eventually want to do this, we should wait for other changes to librarian.yaml to be made first.

jskeet:
Note: for anything non-Python-specific where information is missing or wrong, we should add it to sdk.yaml in the librarian repo. I'm happy to do that; just add a comment. I'd suggest ignoring .repo-metadata.json files as they don't have an impact beyond what happens for doc generation etc. If it would be helpful, I can add a commit that restores all the .repo-metadata.json files so that it's easier to see the real changes.

packages/google-cloud-beyondcorp-appgateways/.repo-metadata.json
```diff
@@ -2,13 +2,13 @@
     "api_description": "Beyondcorp Enterprise provides identity and context aware access controls for enterprise resources and enables zero-trust access. Using the Beyondcorp Enterprise APIs, enterprises can set up multi-cloud and on-prem connectivity using the App Connector hybrid connectivity solution.",
     "api_id": "beyondcorp.googleapis.com",
     "api_shortname": "beyondcorp",
-    "client_documentation": "https://cloud.google.com/python/docs/reference/beyondcorpappgateways/latest",
+    "client_documentation": "https://cloud.google.com/python/docs/reference/google-cloud-beyondcorp-appgateways/latest",
```
daniel-sanche, line 5: Just looked at this one at random, but it looks like the old link worked, but the new one is broken

packages/google-cloud-certificate-manager/README.rst
```diff
@@ -14,9 +14,9 @@ Python Client for Certificate Manager
    :target: https://pypi.org/project/google-cloud-certificate-manager/
 .. |versions| image:: https://img.shields.io/pypi/pyversions/google-cloud-certificate-manager.svg
    :target: https://pypi.org/project/google-cloud-certificate-manager/
-.. _Certificate Manager: https://cloud.google.com/python/docs/reference/certificatemanager/latest
-.. _Client Library Documentation: https://cloud.google.com/python/docs/reference/certificatemanager/latest/summary_overview
-.. _Product Documentation:  https://cloud.google.com/python/docs/reference/certificatemanager/latest
+.. _Certificate Manager: https://cloud.google.com/python/
+.. _Client Library Documentation: https://cloud.google.com/python/docs/reference/google-cloud-certificate-manager/latest/summary_overview
+.. _Product Documentation:  https://cloud.google.com/python/
```
daniel-sanche, line 19: this one too

daniel-sanche (commented):
Most of the changes look reasonable. But it looks like gemini caught some problems. And doing a few random spot-checks on new docs URLs, I'm seeing some broken links

Do you know how those links are handled in other languages?

packages/google-cloud-beyondcorp-appgateways/.repo-metadata.json
jskeet: That's odd - I didn't *think* my change even removed client_documentation_uri overrides - I'll need to check that tomorrow. This part is language-specific, and I hadn't expected to remove those overrides, at least not to start with.

jskeet: Ah - I think this is due to the `MetadataNameOverride` being used in the client_documentation. I'm going to check, but I suspect it's *also* going to be used on the publishing side. I think I'm going to take a closer look at these overrides one at a time...

jskeet:
I'm going to close this PR and create separate ones for each override, removing them from librarian.yaml and then regenerating. That way we can actually have PRs ready to merge, with any luck - and each should be easier to review.

Issue #16780, opened 2026-04-23
Compute: validate the use of gRPC service config before the next release


jskeet:
https://github.com/googleapis/google-cloud-python/pull/16751 regenerated Compute using the gRPC service config to provide timeouts etc. This should be fine (it's used in all other languages), but we should validate it before we next do a release.

This issue will be used as a comment in .librarian/config.yaml for blocking the release of compute until we're happy.

parthea:
We should consider creating a pre-release version of `google-cloud-compute` since the library is stable. We should also ship `google-cloud-compute-v1beta` first (regardless of pre-release since it's preview)

parthea:
https://pypi.org/project/google-cloud-compute-v1beta/0.11.0/ was released : https://pypi.org/project/google-cloud-compute-v1beta/0.11.0/

PR #16781, opened 2026-04-23
chore: block releasing compute until #16780 is addressed

jskeet:


parthea (approved):
LGTM, I'll add a comment in issue https://github.com/googleapis/google-cloud-python/issues/16780 to recommend creating a pre-release version, or shipping `google-cloud-compute-v1beta` first

PR #16791, opened 2026-04-24
chore: remove product_documentation_override from librarian.yaml

jskeet:
Towards https://github.com/googleapis/librarian/issues/5466

### DIFF
--- a/packages/grpc-google-iam-v1/docs/README.rst
+++ b/packages/grpc-google-iam-v1/docs/README.rst
@@ -1,22 +1,22 @@
 Python Client for Cloud Identity and Access Management
 ======================================================
 
-|stable| |pypi| |versions|
+|preview| |pypi| |versions|
 
 `Cloud Identity and Access Management`_: Manages access control for Google Cloud Platform resources.
 
 - `Client Library Documentation`_
 - `Product Documentation`_
 
-.. |stable| image:: https://img.shields.io/badge/support-stable-gold.svg
+.. |preview| image:: https://img.shields.io/badge/support-preview-orange.svg
    :target: https://github.com/googleapis/google-cloud-python/blob/main/README.rst#stability-levels
 .. |pypi| image:: https://img.shields.io/pypi/v/grpc-google-iam-v1.svg
    :target: https://pypi.org/project/grpc-google-iam-v1/
 .. |versions| image:: https://img.shields.io/pypi/pyversions/grpc-google-iam-v1.svg
    :target: https://pypi.org/project/grpc-google-iam-v1/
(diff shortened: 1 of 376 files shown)


packages/grpc-google-iam-v1/docs/README.rst
```diff
@@ -1,22 +1,22 @@
 Python Client for Cloud Identity and Access Management
 ======================================================
 
-|stable| |pypi| |versions|
+|preview| |pypi| |versions|
```
jskeet, line 4: This aspect is covered separately in https://github.com/googleapis/google-cloud-python/pull/16790, but it didn't feel worth manually removing it.

jskeet:
@daniel-sanche: I suggest that links that are just changed are probably accepted, but where we end up with no value at all, we go one of two ways:

- For API-based libraries, add the value to sdk.yaml in librarian so that it's used by all languages
- For handwritten libraries, we maintain the override for now and figure out the best way forward. (I think we can probably keep the field, but fail during generation if it's specified for a library which has an API. In some cases it may be better for it to just be missing, and change the generator to not generate a line in the docs.)

I think the split is like this:

API-based libraries:
- google-cloud-common
- google-cloud-compute-v1beta
- google-cloud-datastore
- google-cloud-firestore
- google-geo-type
- google-shopping-type
- google-cloud-source-context

Handwritten libraries:

- bigframes
- bigquery-magics
- db-dtypes
- gapic-generator
- gcp-sphinx-docfx-yaml
- google-api-core
- googleapis-common-protos
- google-auth-httplib2
- google-auth-oauthlib
- google-auth
- google-cloud-bigquery
- google-cloud-core
- google-cloud-dns
- google-cloud-documentai-toolbox
- google-cloud-ndb
- google-cloud-runtimeconfig
- google-cloud-testutils
- google-crc32c
- google-resumable-media
- pandas-gbq
- proto-plus
- sqlalchemy-bigquery
- sqlalchemy-spanner

jskeet:
Actually, for the handwritten libraries, presumably the README isn't generated anyway, so .repo-metadata.json doesn't matter, unless it's being consumed elsewhere. So we should be able to get rid of the field in librarian.yaml after all - but we need to upstream the values to sdk.yaml first.

jskeet:
Regenerating for a second PR, based on https://github.com/googleapis/librarian/pull/5573
Marking this PR as do not merge and changing it to a draft, but leaving it open for more comment.

jskeet:
The Spanner changes aren't intended to be in this PR - will revert those in the second PR; ignore them for now.

jskeet:
Closing: https://github.com/googleapis/google-cloud-python/pull/16810 is the PR that should be merged.

PR #16799, opened 2026-04-24
chore: regenerate google-cloud-storage

jskeet:


parthea:
I was able to get the `Kokoro pre-release Tests` to pass in https://github.com/googleapis/google-cloud-python/pull/16803 . Someone from the GCS team will need to take a look at `python-sdk-zb-system-test (gcs-aiml-clients-testing-101`. The failure is `Step #3 - "run-tests-and-delete-vm": ERROR: (gcloud.compute.ssh) FAILED_PRECONDITION: Login profile size exceeds 32 KiB. Delete profile values to make additional space.`

I don't believe the `cleanup-old-keys` step is working

https://github.com/googleapis/google-cloud-python/blob/daeb9314c275cfac2d42e85c8395212fcff35d0a/packages/google-cloud-storage/cloudbuild/zb-system-tests-cloudbuild.yaml#L26-L59

```
Step #1 - "cleanup-old-keys": Keys to delete: 
Step #1 - "cleanup-old-keys": No keys found to delete. Nothing to do.
```

Since the check is broken, we can temporarily disable it, block releases and file a bug to follow up on it before we enable releases.

jskeet:
Closing in favor of #16803

PR #16808, opened 2026-04-26
ci: add pylint presubmit on golden files (closes #16393)

### LINKED ISSUE #16393: Add a `pylint` presubmit check on the golden files
I ran `pylint` on the generated python client for `google-cloud-optimization` locally and received several warnings. See the attached [build log](https://github.com/googleapis/gapic-generator-python/files/12337853/pylint_output.txt). Several of these warnings appear to be valid issues. We should address the warnings and add a `pylint` presubmit check to ensure that the `pylint` continues to succeed once issues are addressed.

MukundaKatta:
## Why

Issue #16393 asks for a `pylint` presubmit check on the GAPIC golden client libraries so generator regressions are caught at PR time. `flake8` and `ruff` already run on these goldens, but they do not catch the kinds of issues `pylint` flagged in the original investigation (undefined names, broken imports, etc.).

## What

- New `packages/gapic-generator/tests/integration/goldens/.pylintrc` scoped to the goldens. Conservative bar to start: `pylint --errors-only` is used in CI, and the `.pylintrc` disables the small set of error-class checks that consistently misfire on auto-generated GAPIC output (`no-name-in-module`, `no-member`, `import-error`, `relative-beyond-top-level`, `cyclic-import`). This catches real bugs without forcing a stylistic cleanup of the existing goldens, which can be tightened later.
- New `Pylint goldens (errors only)` step in the existing `goldens` job in `.github/workflows/gapic-generator-tests.yml`. Runs after the existing `format`/`lint`/`unit` nox sessions, so the heavy setup is amortized. Iterates over the same `credentials eventarc logging redis` set.

## Tested

- Confirmed the new step is gated by the same `paths` filter as the rest of `gapic-generator-tests.yml`, so it stays silent on PRs that do not touch the generator.
- `pylint --errors-only --rcfile=.pylintrc` only reports E/F categories minus the listed disables, which keeps the signal meaningful and stops the generator from silently regressing on the next refactor.

If the maintainers want a different starting bar (e.g. include warnings, or run on more golden packages), happy to adjust.

### DIFF
--- a/.github/workflows/gapic-generator-tests.yml
+++ b/.github/workflows/gapic-generator-tests.yml
@@ -115,6 +115,17 @@ jobs:
           for pkg in credentials eventarc logging redis; do
             nox -f tests/integration/goldens/$pkg/noxfile.py -s format lint unit-${{ needs.python_config.outputs.latest_stable_python }}
           done
+      # Run pylint (errors-only) over the goldens so generator regressions
+      # like undefined names or import-time breakage that ruff/flake8 do
+      # not flag are caught at PR time.
+      # See https://github.com/googleapis/google-cloud-python/issues/16393.
+      - name: Pylint goldens (errors only)
+        run: |
+          pip install --quiet pylint
+          cd packages/gapic-generator/tests/integration/goldens
+          for pkg in credentials eventarc logging redis; do
+            pylint --rcfile=.pylintrc --errors-only --recursive=y "$pkg/google"
+          done
 
   goldens-prerelease:
     needs: python_config
--- a/packages/gapic-generator/tests/integration/goldens/.pylintrc
+++ b/packages/gapic-generator/tests/integration/goldens/.pylintrc
@@ -1,0 +1,28 @@
+# Pylint configuration for golden client libraries.
+#
+# These goldens are auto-generated GAPIC client output (see
+# packages/gapic-generator/gapic/templates/). They are checked in as test
+# fixtures so that regressions in the generator are caught at PR time.
+#
+# Pylint is invoked with `--errors-only` from CI so only error/fatal
+# categories run. This file disables the small set of error-class checks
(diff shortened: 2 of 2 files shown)


parthea:
Hi @MukundaKatta, Thanks for opening a PR! If you're interested in continuing with this contribution, please review the presubmit failure in https://github.com/googleapis/google-cloud-python/actions/runs/24968202788/job/73295304241?pr=16808

```
************* Module eventarc_v1.services.eventarc.transports.rest_base
eventarc/google/cloud/eventarc_v1/services/eventarc/transports/rest_base.py:107:8: E0309: __hash__ does not return int (invalid-hash-returned)
...
```

One potential solution is to silence pylint for the functions which are not implemented. The change would need to be made in the jinja template below:

https://github.com/googleapis/google-cloud-python/blob/64ac5f1bf60d1996facbba152e34237fdaa60f95/packages/gapic-generator/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/rest_base.py.j2#L108-L109

Once the template is changed, follow the instructions below to regenerate the golden files:

https://github.com/googleapis/google-cloud-python/blob/main/packages/gapic-generator/DEVELOPMENT.md#integration-tests

MukundaKatta:
@parthea pushed a fix: added `invalid-hash-returned` to the goldens `.pylintrc` disable list with a comment pointing at the generator template (it emits `return NotImplementedError(...)` instead of `raise`). That should clear the eventarc presubmit while the generator-side fix is tracked separately.

@gemini-code-assist re your point on `no-name-in-module` / `import-error`: agree those mask real breakage. Happy to follow up by setting `PYTHONPATH` in the workflow and removing those two disables once this PR lands, or fold it in here if you'd rather.

parthea (approved):
Thanks @MukundaKatta!

ohmayr (approved):

PR #5466, opened 2018-06-11
Release spanner 1.4.0

crwilcox:
This pull request was generated using releasetool.

Fixes #5466 5465

PR #16810, opened 2026-04-27
chore: remove product_documentation_override from librarian.yaml

jskeet:
For handwritten libraries, this only affects .repo-metadata.json

For generated libraries, the link may change, but should still be
appropriate, and must exist so that the rst file is valid.

Towards https://github.com/googleapis/google-cloud-python/pull/5466


jskeet:
Look like common protos is still broken. Not sure why that is yet - will look shortly.

jskeet:
Given that the common protos aren't really a product, it would be better for us not to generate documentation which expects that it *they* are a product.

We could conditionalize the template at https://raw.githubusercontent.com/googleapis/synthtool/refs/heads/master/synthtool/gcp/templates/python_mono_repo_library/README.rst, potentially.

parthea (approved):

PR #16838, opened 2026-04-28
perf(storage): improve microbenchmark throughput calculation accuracy

zhixiangli:
This PR improves the accuracy of microbenchmark throughput calculations by:
- Using `time.perf_counter()` for high-resolution timing.
- Tracking actual measured intervals for each worker thread/process.
- Correcting the aggregation logic for multi-process/multi-coroutine runs.
- Updating reporting to handle granular duration data.

### DIFF
--- a/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
+++ b/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
@@ -20,6 +20,7 @@ import os
 import random
 import time
 from io import BytesIO
+from typing import List, NamedTuple, Optional
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
```diff
@@ -78,15 +136,20 @@ def _download_time_based_json(client, filename, params):
...
-    while time.monotonic() < test_end_time:
-        current_time = time.monotonic()
+    while time.perf_counter() < test_end_time:
+        current_time = time.perf_counter()
         if is_warming_up and current_time >= warmup_end_time:
             is_warming_up = False
             total_bytes_downloaded = 0  # Reset counter after warmup
+            measured_start_time = _record_measured_start(
+                measured_start_time, current_time
+            )
```
chandra-siri: why can't this be just `measured_start_time = time.perf_counter()` ? Having `_record_measured_start` is overkill. 

Because the code never sets `measured_start_time` more than 1 time.

chandra-siri (changes requested):

chandra-siri (approved):

PR #16841, opened 2026-04-28
chore: address product documentation in googleapis-common-protos

jskeet:
- Remove the product_documentation_override
- Add a string replacement to remove inappropriate lines

Completes the google-cloud-python part of https://github.com/googleapis/google-cloud-python/pull/5466

Filed https://github.com/googleapis/librarian/issues/5670 to remove the need for the string replacement when we migrate synthtool code into librarian.

### DIFF
--- a/.librarian/generator-input/client-post-processing/googleapis-common-protos-docs.yaml
+++ b/.librarian/generator-input/client-post-processing/googleapis-common-protos-docs.yaml
@@ -1,0 +1,69 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+description: Remove inappropriate product documentation lines in googleapis-common-protos readme
+url: https://github.com/googleapis/librarian/issues/5670
+replacements:
+  - paths: [
+      packages/googleapis-common-protos/README.rst
+    ]
+    before: "`Google APIs Common Protos`_: Lets you define and config your API service.\n\n"
+    after: ""
+    count: 1
+  - paths: [
+      packages/googleapis-common-protos/README.rst
+    ]
+    before: "- `Product Documentation`_\n"
+    after: ""
+    count: 1
+  - paths: [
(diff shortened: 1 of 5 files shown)


jskeet:
Will fix the merge conflict

.librarian/generator-input/client-post-processing/googleapis-common-protos-docs.yaml
```diff
@@ -0,0 +1,69 @@
...
+      -  Read the `Google APIs Common Protos Product documentation`_ to learn
+         more about the product and see How-to Guides.
+    after: ""
+    count: 1
+  - paths: [
+      packages/googleapis-common-protos/README.rst
+    ]
+    before: ".. _Google APIs Common Protos Product documentation:  \n"
+    after: ""
+    count: 1
```
parthea, line 69: I’m concerned about introducing additional post-processing steps at this stage of the librarian migration. Since the common protos README was previously maintained by hand, could we revert to the manual version for now? I’d prefer to wait until we have a robust way to support autogeneration for `proto-only` libraries rather than adding more customizations.

jskeet: I'd be happy to revert to a manual version, yes - there's a lot about this file which isn't great. But how do we stop the post-processor from generating the readme?

parthea: My mistake, the README is autogenerated from synthtool here:  https://github.com/googleapis/synthtool/blob/ba9afbd284495433734b84686ef8ef6785c60128/synthtool/languages/python_mono_repo.py#L231-L254

To switch to a handwritten README, we would need to modify synthtool to skip this logic for proto-only libraries. Since this is an existing issue, perhaps we don't need all of the post processing here unless we want to fix it properly. In that case, the fix should be in synthtool where we skip the code below for proto-only packages

https://github.com/googleapis/synthtool/blob/ba9afbd284495433734b84686ef8ef6785c60128/synthtool/languages/python_mono_repo.py#L231-L254

jskeet: > To switch to a handwritten README, we would need to modify synthtool to skip this logic for proto-only libraries. Since this is an existing issue, perhaps we don't need all of the post processing here unless we want to fix it properly. In that case, the fix should be in synthtool where we skip the code below for proto-only packages

I'd rather fix it "hackily" for now (with this PR), and then do it properly when we rewrite bits of synthtool in Go (and ideally remove all kinds of string replacements - including this one). Are you okay with that?

parthea: Ack. I see the follow up work is captured in https://github.com/googleapis/librarian/issues/5670#issuecomment-4334234945. I'll approve this so that it doesn't block the overall migration effort.

parthea (approved):

PR #16842, opened 2026-04-28
chore: remove name_pretty_override from google-cloud-maintenance-api

jskeet:


jskeet:
@parthea Using an additional string replacement for docs isn't *nice*, but it does isolate the solution to where the problem actually occurs. (The "name pretty" being Mainteance isn't inherently an issue.)

.librarian/generator-input/client-post-processing/cloud-maintenance-docs.yaml
```diff
@@ -0,0 +1,28 @@
...
+      packages/google-cloud-maintenance-api/README.rst
+    ]
+    before: '`Maintenance`'
+    after: '`Cloud Maintenance`'
+    count: 1
+  - paths: [
+      packages/google-cloud-maintenance-api/README.rst
+    ]
+    before: '_Maintenance:'
+    after: '_Cloud Maintenance:'
```
parthea, line 27: Would it be better to keep the override rather than add post processing here?

jskeet: I don't think so, because it's really not that we want to override the "pretty name" in general - we just want to fix the link. Aside from anything else, this is a whole option that will be blocked from removal (I believe) based on *just* this API.

jskeet:
Will resolve the conflicts in this PR.

jskeet:
When we come to fix this later, we could decide to use a different title for the present-in-all-readmes "maintenance" link instead, of course. That would remove the need for any string manipulation. (But I don't want to start messing with synthtool templates at this point.)

.librarian/generator-input/client-post-processing/cloud-maintenance-docs.yaml
parthea: Ack. Adding post processing is a slippery slope. We should file a bug to follow up on this. In the meantime I'll approve it so that we don't want to block the overall migration effort.

parthea (approved):

.librarian/generator-input/client-post-processing/cloud-maintenance-docs.yaml
jskeet: We've got https://github.com/googleapis/librarian/issues/5668 :)

PR #16844, opened 2026-04-28
chore: run librarian tidy

jskeet:
This does not affect generation as only options which would be derived to the same value are removed.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -98,11 +98,6 @@ libraries:
       - path: google/ads/admanager/v1
     description_override: Manage your Ad Manager inventory, run reports and more.
     python:
-      opt_args_by_api:
-        google/ads/admanager/v1:
-          - python-gapic-namespace=google.ads
-          - python-gapic-name=admanager
-          - warehouse-package-name=google-ads-admanager
       default_version: v1
   - name: google-ads-datamanager
     version: 0.8.0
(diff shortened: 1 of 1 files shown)


jskeet:
Just looking into why there's an addition in the diff at all....

jskeet:
> Just looking into why there's an addition in the diff at all....

Ah, this was just an option being in the wrong order.

chalmerlowe:
I am not tracking what the end goal is OR why we are doing this. As such I am seeing a variance in how some of these are deleted and I don't know why.

In this example why did all the items in one list get deleted, but only one item in the second list get deleted (even though one of the items (e.g. `python-gapic-namespace=google.ads`) in the second list matches the item in the first list so it would seem reasonable for to to be deleted)?

Screenshot 2026-04-28 at 10 09 42

jskeet:
Basically, librarian derives python-gapic-namespace and python-gapic-name from the API path, and warehouse-package-name from the package name. (So for example, google.ads is derived in the first case because the path starts with google/ads, but not in the second case because the path starts with google/marketingplatform.) Librarian now *always* specifies all three to the GAPIC generator, so that whatever the GAPIC generator does by default is irrelevant.

This PR removes any value where it matches the value that would be derived - so basically what we pass to the GAPIC generator is identical but we remove a load of redundant configuration in the file.

parthea (approved):

PR #16846, opened 2026-04-28
chore: reenable generation for google-cloud-bigquery

jskeet:
This reintroduces the issue tracker override, as we can't derive that from sdk.yaml (as nothing is generated here).


jskeet:
Created as a draft so we can see if the tests now pass. The first commit is just from #16844; will only mark it as ready for review when that's merged.

PR #5463, opened 2018-06-11
Add support for Spanner struct params.

tseaver:
/cc @snehashah16, @adirastogi

PR #16850, opened 2026-04-28
chore: remove name_pretty_override from all APIs

jskeet:
Towards https://github.com/googleapis/google-cloud-python/pull/5463


jskeet:
I've addressed the conflict in the web UI - now presumably most of the tests need to restart, but at least it'll be unblocked...

jskeet:
Ping @parthea for a review if you have time today - I'm hoping any failed checks are just flakes.

jskeet:
Kokoro error:

> ERROR: Aborting VM command due to timeout of 10800 seconds

I'll try rerunning the Kokoro tests. If that doesn't work, we could potentially split this PR into separate ones, with (say) 50 packages at a time. That's be annoying, but feasible.

jskeet:
Closing as the django-spanner tests are still failing as well.

I've created #16861 for just Storage, PubSub, BigQuery and the common protos. I'll create another PR with "everything apart from those and django-spanner" to see if that manages to pass tests.

PR #16860, opened 2026-04-29
test: use _DummyListBuffer in test_reads.py to avoid GIL contention

zhixiangli:
This PR replaces `io.BytesIO` with a custom `_DummyListBuffer` in `test_reads.py`.

During profiling, `io.BytesIO.write()` holds the GIL while copying data. This introduces significant noise and bottlenecks in performance tests with high concurrency or large data transfers.

This change avoids GIL contention and reduces noise when profiling the code (specifically noted by the user as causing noise when profiling the format code).

### DIFF
--- a/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
+++ b/packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
@@ -170,6 +170,24 @@ def _download_time_based_json(client, filename, params):
     )
 
 
+# _DummyListBuffer is used instead of io.BytesIO to avoid GIL contention
+# during profiling. io.BytesIO.write() holds the GIL while copying data,
+# which introduces significant noise and bottlenecks in performance tests
+# with high concurrency or large data transfers.
+# This buffer simply collects chunks in a list and tracks the total size.
+class _DummyListBuffer:
+    def __init__(self):
+        self.chunks = []
+        self.size = 0
+
+    def write(self, data):
+        self.chunks.append(data)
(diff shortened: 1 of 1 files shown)


packages/google-cloud-storage/tests/perf/microbenchmarks/time_based/reads/test_reads.py
```diff
@@ -114,6 +113,26 @@ def _download_time_based_json(client, filename, params):
     return total_bytes_downloaded
 
 
+# _DummyListBuffer is used instead of io.BytesIO to avoid GIL contention
+# during profiling. io.BytesIO.write() holds the GIL while copying data,
+# which introduces significant noise and bottlenecks in performance tests
+# with high concurrency or large data transfers.
+# This buffer simply collects chunks in a list and tracks the total size.
+class _DummyListBuffer:
```
googlyrahman, line 176: We've better version of this already in `gcsfs`, https://github.com/ankitaluthra1/gcsfs/blob/8d9b2efc77392f75d450dd2885c83ca92029f607/gcsfs/zb_hns_utils.py#L181

The long term plan is to bring that buffer here.

zhixiangli: Yes, the point is that `user_buffer`  implementation details should not be exposed in the performance tests.

googlyrahman: Aah, this is for test file.

chandra-siri (approved):

PR #16862, opened 2026-04-29
chore: remove name_pretty_override from most packages

jskeet:
This omits django-spanner and gapic-generator (which currently have failing tests) and the
packages modified in https://github.com/googleapis/google-cloud-python/pull/16861.


jskeet:
The kokoro failure is a timeout again. Looks like we'll need to do this in smaller batches.

PR #16907, opened 2026-05-01
refactor(bigframes): Modularize compiler routing as proxy executor

TrevorBergeron:


packages/bigframes/bigframes/session/bq_caching_executor.py
```diff
@@ -83,11 +80,15 @@ def __init__(
         metrics: Optional[bigframes.session.metrics.ExecutionMetrics] = None,
         enable_polars_execution: bool = False,
         publisher: bigframes.core.events.Publisher,
-        labels: Mapping[str, str] = {},
+        labels: tuple[tuple[str, str], ...] = (),
+        compiler_name: Literal["ibis", "sqlglot"] = "ibis",
```
tswast: didn't we change the default to "sqlglot"?

packages/bigframes/bigframes/session/bq_caching_executor.py
```diff
@@ -369,8 +376,9 @@ def _run_execute_query(
                 bigframes.options.compute.maximum_bytes_billed
             )
 
-        if self._labels:
+        if self._labels or extra_labels:
             job_config.labels.update(self._labels)
```
tswast, line 380: Could `._labels` or `extra_labels` be `None`? If so, I think this would fail.

packages/bigframes/bigframes/session/proxy_executor.py
```diff
@@ -0,0 +1,169 @@
...
+        storage_manager: temporary_storage.TemporaryStorageManager,
+        bqstoragereadclient: google.cloud.bigquery_storage_v1.BigQueryReadClient,
+        loader: loader.GbqDataLoader,
+        *,
+        metrics: Optional[bigframes.session.metrics.ExecutionMetrics] = None,
+        enable_polars_execution: bool = False,
+        publisher: bigframes.core.events.Publisher,
+        labels: tuple[tuple[str, str], ...] = (),
+    ):
+        self._enable_polars_execution = enable_polars_execution
```
tswast, line 57: Thought: should we use a similar proxy mechanism for polars execution sometime? I guess that's maybe a bit different since which we pick depends on the input sources?

packages/bigframes/bigframes/session/proxy_executor.py
```diff
@@ -0,0 +1,169 @@
...
+    ) -> str:
+        """
+        Convert an ArrayValue to a sql query that will yield its value.
+        """
+        return self._ibis_executor.to_sql(
+            array_value,
+            offset_column=offset_column,
+            ordered=ordered,
+            enable_cache=enable_cache,
+        )
```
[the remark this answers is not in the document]
tswast: In the interest of a staged rollout, I'm fine keeping this ibis, but we should probably add a todo to swap this before we switch to sqlglot everywhere.

tswast (approved):

packages/bigframes/bigframes/session/bq_caching_executor.py
TrevorBergeron: both are enforced to be non-null - for extra safety have split into two separate if conditions

tswast (approved):

packages/bigframes/bigframes/session/proxy_executor.py
TrevorBergeron: This proxy approach is a bit blunt just to get it out to a separate independent layer.

Local engine dispatching is more integral, so will afford it more dynamic integration, including looking at the plan itself as you mentioned.

PR #16909, opened 2026-05-01
chore: enable generation for google-crc32c

jskeet:
This isn't really generated, but this change will get `.repo-metadata.json` up-to-date.

The package previously had generation disabled due to test failures.

### DIFF
--- a/packages/google-crc32c/.repo-metadata.json
+++ b/packages/google-crc32c/.repo-metadata.json
@@ -1,11 +1,9 @@
 {
-    "client_documentation": "https://github.com/googleapis/python-crc32c",
+    "client_documentation": "https://googleapis.dev/python/google-crc32c/latest",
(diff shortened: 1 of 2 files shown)


jskeet:
Converting to draft as at least the BigQuery failure is still present. I'll wait until we see whether crc32c has been fixed before closing or amending the PR.

packages/google-crc32c/.repo-metadata.json
```diff
@@ -1,11 +1,9 @@
 {
-    "client_documentation": "https://github.com/googleapis/python-crc32c",
+    "client_documentation": "https://googleapis.dev/python/google-crc32c/latest",
```
jskeet, line 2: Note that this doesn't exist, but won't be used anywhere anyway.

parthea (approved):

Issue #16911, opened 2026-05-01
Refactor ProtoEnumMeta to use _ignore_ for _pb_options


chalmerlowe:
We dropped Python 3.7 support and moved to Python 3.9 as the minimum version. This allows us to use the specialized `_ignore_` attribute in `Enum` to ignore `_pb_options` instead of the current manual removal from `_member_names`. Please effect this change to clean up the code in packages/proto-plus/proto/enums.py rather than using the current cumbersome approach.

```
        pb_options = "_pb_options"
        opts = attrs.pop(pb_options, {})
        # This is the only portable way to remove the _pb_options name
        # from the enum attrs.
        # TODO: Use _ignore_ attribute to ignore _pb_options (Issue #16911)
        if pb_options in attrs._member_names:
            if isinstance(attrs._member_names, list):
                idx = attrs._member_names.index(pb_opt
                attrs._member_names.pop(idx)
            elif isinstance(attrs._member_names, set):
                attrs._member_names.discard(pb_options
            else:  # Python 3.11.0b3
                del attrs._member_names[pb_options]
```

https://github.com/googleapis/google-cloud-python/blob/daa042330755b0db1853b6e45568ead2a3066f4f/packages/proto-plus/pyproject.toml#L23

PR #16910, opened 2026-05-01
chore(versions): update Python version support to 3.10+ in gapic-generator

chalmerlowe:
> [!note]
> This PR has been updated to remove 3.7, 3.8, and 3.9 (with one exception discussed below).

### Overview
This Pull Request updates the gapic-generator package and related CI configurations to reflect the deprecation of Python 3.7 through 3.9. It updates documentation, templates, and CI scripts to establish Python 3.10 as the new minimum supported version where applicable.

#### Documentation:
* Updated installation and verification instructions in packages/gapic-generator/docs to reference Python 3.10
#### Templates:
* Updated __init__.py.j2 templates in gapic/ads-templates to require Python 3.10+ (updating sys.version_info checks and error messages).
* Removed outdated comments regarding Python 3.9 and below features (like module-level __getattr__) to avoid confusion, as the baseline runtime is now higher than the version noted in the comments.
* Updated mypy.ini.j2 to target Python 3.10.
* Cleaned up dead code (e.g. commented-out Jinja snippets) in the templates.
#### CI Configurations:
* Removed Python 3.9 and below from the unit test matrix in ci/run_single_test.sh.
* Updated comments regarding supported Python versions in ci/run_conditional_tests.sh.

### Reasoning
Python 3.9 and below have reached End-of-Life (EOL) and are no longer supported. These changes ensure that our generator and the libraries it produces align with current support policies and reduce maintenance burden by removing dead code paths.

> [!note]
> because the config that runs this PR checks is not updated yet, it still wants to run 3.9 so we have a temp change to skip that nox session. Once this PR merges future PRs will no longer attempt to run 3.9 and we will follow-up with a small PR to strip out that last 3.9 reference.

Fixes internal issue: http://b/482126936 🦕

PR #16912, opened 2026-05-01
chore(proto-plus): update Python version and clean up enums.py

chalmerlowe:
This PR updates `proto-plus` to establish Python 3.10 as the minimum supported version, dropping support for Python 3.9 and below.

### Changes

* Configuration: Updated `pyproject.toml` and `noxfile.py` to require Python 3.10+ and remove references to Python 3.9.
* Code: Refactored `repeated.py` to use the walrus operator (PEP 572).
Documentation: Updated a traceback example in `docs/fields.rst` to reference Python 3.10.
* Cleanup: Replaced a stale comment regarding Python 3.7 features in proto/enums.py with a tracked TODO referencing Issue #16911.

> [!note]
> Also includes a conditional in the `run_single_test.sh` file to manage the chicken OR the egg problem of managing failing tests when a `unit-3.9` nox session is not present. Because we are migrating handwritten libs one at a time, we are in a transition period were we must still test for 3.9 in libraries that still support it but cannot have a failing presubmit in libraries that don't support it.

### Blocking

- [x] This PR is blocked by PR #16910

### DIFF
--- a/packages/proto-plus/proto/enums.py
+++ b/packages/proto-plus/proto/enums.py
@@ -55,8 +55,7 @@ class ProtoEnumMeta(enum.EnumMeta):
         opts = attrs.pop(pb_options, {})
         # This is the only portable way to remove the _pb_options name
         # from the enum attrs.
-        # In 3.7 onwards, we can define an _ignore_ attribute and do some
-        # mucking around with that.
+        # TODO: Use _ignore_ attribute to ignore _pb_options (Issue #16911)
         if pb_options in attrs._member_names:
             if isinstance(attrs._member_names, list):
                 idx = attrs._member_names.index(pb_options)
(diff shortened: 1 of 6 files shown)


parthea:

packages/proto-plus/proto/enums.py
```diff
@@ -55,8 +55,7 @@ def __new__(mcls, name, bases, attrs):
         opts = attrs.pop(pb_options, {})
         # This is the only portable way to remove the _pb_options name
         # from the enum attrs.
-        # In 3.7 onwards, we can define an _ignore_ attribute and do some
-        # mucking around with that.
+        # TODO: Use _ignore_ attribute to ignore _pb_options (Issue #16911)
```
chalmerlowe, line 58:
```suggestion
        # TODO: Use _ignore_ attribute to ignore _pb_options (see Issue #16911)
```

parthea (approved):

PR #16913, opened 2026-05-01
fix(sqlalchemy-spanner): Drop support for Python 3.8 and 3.9

chalmerlowe:
This PR updates `sqlalchemy-spanner` to establish Python 3.10 as the minimum supported version, dropping support for Python 3.9 and below.

### Changes
* Configuration: Updated `setup.py` and `noxfile.py` to require Python 3.10+ and remove references to Python 3.7, 3.8, and 3.9.
* Cleanup: Removed dead code in `noxfile.py` that was skipping Python 3.7.

### Blocking
This PR is blocked by PR #16910


parthea:

packages/sqlalchemy-spanner/setup.py
```diff
@@ -58,11 +58,10 @@
     author="Google LLC",
     author_email="googleapis-packages@google.com",
     license="Apache 2.0",
+    python_requires=">=3.10",
     classifiers=[
         "Intended Audience :: Developers",
         "License :: OSI Approved :: Apache Software License",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
```
parthea, line 65: We should have `python_requires` in this file to avoid breaking existing environments. Without `python_requires` pip will install a potentially incompatible versions of this library

parthea (changes requested):

packages/sqlalchemy-spanner/setup.py
parthea: My bad. I see it was added. Sorry for the noise.

parthea (approved):

PR #16914, opened 2026-05-01
chore(django-google-spanner): update Python version to 3.10

chalmerlowe:
This PR updates `django-google-spanner` to establish Python 3.10 as the minimum supported version, dropping support for Python 3.9 and below.

### Changes

* Configuration: Updated `setup.py` and `noxfile.py` to require Python 3.10+ and remove references to Python 3.7, 3.8, and 3.9.
* Cleanup: Removed dead code in `noxfile.py` that was skipping Python 3.7.

### Blocking

This PR is blocked by PR #16910


packages/django-google-spanner/noxfile.py
```diff
@@ -123,8 +122,7 @@ def default(session, django_version="5.2"):
 @nox.session(python=ALL_PYTHON)
 def unit(session):
     """Run the unit test suite."""
-    if session.python == "3.9":
-        session.skip("Python 3.9 is not supported for Django 5.2 tests")
```
parthea, line 127: nit: we should keep this until 3.9 is removed from `gapic-generator` and autogenerated packages which account for 90% of the packages in this repo.

parthea:

packages/django-google-spanner/noxfile.py
chalmerlowe: This PR and one's similar to it are blocked and are only presented for discussion.
Once the gapic-generator is updated fully to 3.10, these will be unblocked.

packages/django-google-spanner/noxfile.py
```diff
@@ -123,8 +122,7 @@ def default(session, django_version="5.2"):
...
+    # TODO: Remove this check once support for Python 3.14 is added to Protobuf.
```
chalmerlowe:
```suggestion
    # TODO: Remove this check once support for Python 3.14 is added to Protobuf
```

packages/django-google-spanner/noxfile.py
chalmerlowe: PR 16910 has been merged. 3.9 is no longer supported in the gapic generator AND we have instituted logic in the `run_single_test.sh` script to only test `unit-3.9` nox sessions in packages that still support it.

ohmayr (approved):

PR #16921, opened 2026-05-02
fix(google-auth): Handle missing pyOpenSSL for inferred mTLS

### LINKED ISSUE #16920: auth: auto-mtls raises exception if openssl not available
AuthorizedSession.configure_mtls_channel [raises an exception](https://github.com/googleapis/google-cloud-python/blob/adbabae19bd548ef04c9f797d10b1b926b81354e/packages/google-auth/google/auth/transport/requests.py#L454) if the optional OpenSSL package is not installed. This may have been ok with mtls was fully opt-in, but now it is automatically enabled based on the environment, so these failures can hit users by surprise. This seems to be an issue with google-auth>=2.49.0

I am seeing this triggered automatically on the Bigtable rest tests (client [sets up a new transport,](https://github.com/googleapis/python-bigtable/blob/8244ffeb6e61c3fcff8cf6ab37d8f100613111b7/google/cloud/bigtable_v2/services/bigtable/client.py#L758), which calls [configure_mtls_channel](https://github.com/googleapis/python-bigtable/blob/8244ffeb6e61c3fcff8cf6ab37d8f100613111b7/google/cloud/bigtable_v2/services/bigtable/transports/rest.py#L791), which [triggers the exception](https://github.com/googleapis/google-cloud-python/blob/adbabae19bd548ef04c9f797d10b1b926b81354e/packages/google-auth/google/auth/transport/requests.py#L454))

The [docstring](https://github.com/googleapis/google-cloud-python/blob/adbabae19bd548ef04c9f797d10b1b926b81354e/packages/google-auth/google/auth/transport/requests.py#L431) also seems out of date, since it says "The function does nothing unless `GOOGLE_API_USE_CLIENT_CERTIFICATE` is explicitly set to `true`"

A couple options on how to address this:
- we could change [check_use_client_cert](https://github.com/googleapis/google-cloud-python/blob/9a64b0df8d40b29c6fc12938543b2319f9741c5f/packages/google-auth/google/auth/transport/_mtls_helper.py#L410) to take this dependency into account when deciding whether to enable mtls
- we could also change [configure_mtls_channel](https://github.com/googleapis/google-cloud-python/blob/adbabae19bd548ef04c9f797d10b1b926b81354e/packages/google-auth/google/auth/transport/requests.py#L428) to do nothing if the dependency is missing, effectively disabling mtls 
- There is also the question of whether OpenSSL should be made a required dependency, if mTLS is going to be the default going forward

cneuralnetwork:
Fixes #16920.

## Summary

This changes the requests and urllib3 mTLS setup paths so they no longer fail closed when client certificate use was inferred from certificate configuration but `pyOpenSSL` is not installed. In that auto-mTLS case, the transports now disable mTLS and continue using the regular TLS channel.

Explicit opt-in remains strict: if `GOOGLE_API_USE_CLIENT_CERTIFICATE=true` or the Cloud SDK fallback env var explicitly enables client certs, a missing `pyOpenSSL` dependency still raises `MutualTLSChannelError` as before.

## Root Cause

`check_use_client_cert()` can now infer client certificate use from certificate config. The requests and urllib3 transports then imported `OpenSSL` before checking whether a cert/key could actually be used, so environments without optional `pyOpenSSL` could crash during automatically triggered mTLS configuration.

## Changes

- Add internal helpers to distinguish explicit client-cert opt-in from inferred client-cert enablement.
- Make requests and urllib3 transports back off only for inferred mTLS when `pyOpenSSL` is unavailable.
- Update stale mTLS docstrings to describe inferred certificate configuration behavior.
- Add regression tests for inferred mTLS with missing `OpenSSL` in both requests and urllib3 transports.

## Validation

- `/tmp/google-auth-venv/bin/python -m pytest tests/transport/test_requests.py::TestAuthorizedSession::test_configure_mtls_channel_exceptions tests/transport/test_requests.py::TestAuthorizedSession::test_configure_mtls_channel_inferred_missing_openssl tests/transport/test_urllib3.py::TestAuthorizedHttp::test_configure_mtls_channel_exceptions tests/transport/test_urllib3.py::TestAuthorizedHttp::test_configure_mtls_channel_inferred_missing_openssl tests/transport/test__mtls_helper.py::TestCheckUseClientCert` (`16 passed`)
- `/tmp/google-auth-venv/bin/python -m pytest tests/transport/test_requests.py tests/transport/test_urllib3.py tests/transport/test__mtls_helper.py` (`140 passed`, with existing pyOpenSSL deprecation warnings)
- `git diff --check`
- `python -m compileall -q packages/google-auth/google/auth/transport/_mtls_helper.py packages/google-auth/google/auth/transport/requests.py packages/google-auth/google/auth/transport/urllib3.py packages/google-auth/tests/transport/test_requests.py packages/google-auth/tests/transport/test_urllib3.py`

### DIFF
--- a/packages/google-auth/google/auth/transport/_mtls_helper.py
+++ b/packages/google-auth/google/auth/transport/_mtls_helper.py
@@ -448,6 +448,22 @@ def decrypt_private_key(key, passphrase):
     return crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey)
 
 
+def _get_use_client_cert_env():
+    """Returns the configured client certificate opt-in environment value."""
+    use_client_cert = getenv(environment_vars.GOOGLE_API_USE_CLIENT_CERTIFICATE)
+    if use_client_cert is None or use_client_cert == "":
+        use_client_cert = getenv(
+            environment_vars.CLOUDSDK_CONTEXT_AWARE_USE_CLIENT_CERTIFICATE
+        )
+    return use_client_cert
+
+
+def _is_client_cert_explicitly_enabled():
+    """Returns True if an environment variable explicitly enables client certs."""
+    use_client_cert = _get_use_client_cert_env()
+    return bool(use_client_cert and use_client_cert.lower() == "true")
+
+
 def check_use_client_cert():
     """Returns boolean for whether the client certificate should be used for mTLS.
 
@@ -462,11 +478,7 @@ def check_use_client_cert():
     Returns:
         bool: Whether the client certificate should be used for mTLS connection.
     """
-    use_client_cert = getenv(environment_vars.GOOGLE_API_USE_CLIENT_CERTIFICATE)
-    if use_client_cert is None or use_client_cert == "":
-        use_client_cert = getenv(
-            environment_vars.CLOUDSDK_CONTEXT_AWARE_USE_CLIENT_CERTIFICATE
-        )
+    use_client_cert = _get_use_client_cert_env()
 
     # Check if the value of GOOGLE_API_USE_CLIENT_CERTIFICATE is set.
     if use_client_cert:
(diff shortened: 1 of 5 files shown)


cneuralnetwork:
@gemini-code-assist review

PR #16924, opened 2026-05-04
fix(gapic-generator): remove lagging 3.9 refs

chalmerlowe:
> [!note]
> This PR was in a holding pattern. It should be merged now that we have removed the references to 3.9 in the repo (via generation of the GAPIC libraries and via updates to the handwritten libs)

Removes two lagging refs to 3.9 to finalize the migration of the gapic generator codebase to 3.10+ runtime dependencies.

Fixes #[internal issue](http://b/482126936) 🦕


ci/run_single_test.sh
```diff
@@ -70,10 +70,7 @@ case ${TEST_TYPE} in
         ;;
     unit)
         case ${PY_VERSION} in
-        "3.9")
-            nox -s unit-3.9
-            retval=$?
-            ;;
+
```
chalmerlowe: Removing the blank line. NOTE: this is only a diff. While 3.11, 3.12, 3.13, 3.14 are not shown here, they are indeed in the file. 

```suggestion
```

parthea (approved):

PR #16933, opened 2026-05-04
fix: BigFrames respects bq default region

TrevorBergeron:


tswast (approved):
Thanks

packages/bigframes/bigframes/session/__init__.py
```diff
@@ -188,39 +188,50 @@ def __init__(
...
                 bigframes._config.auth.resolve_credentials_and_project(context)
             )
+            if context.location is None:
+                with bigquery.Client(
+                    project=project,
+                    credentials=credentials,
+                ) as temp_client:
+                    row_iter = temp_client.query_and_wait(
+                        "SELECT 1",
+                        job_config=bigquery.QueryJobConfig(dry_run=True),
```
tswast, line 210: Does dry run actually populate the location info? I suppose it should, but good to double-check.

TrevorBergeron: It does afaict from manual usage and the pre-existing integration test

PR #16941, opened 2026-05-05
chore: add a post-submit check that regenerating doesn't change anything

### LINKED ISSUE #4800: google-cloud-python: check for code generation drift in CI
We may want to introduce two new CI checks:

- For handwritten changes, regenerate just the affected packages *before* a PR is merged, to ensure that newly-added handwritten files aren't deleted or modified
- For all changes, regenerate everything *after* a PR is merged (because it takes quite a while) to ensure that generation is still correct

jskeet:
This is a post-submit as it takes a while (~25 minutes). We could *potentially* make it a pre-submit and only generate packages which have changed, but that would be more fiddly.

It would be nice not to specify the protoc or pandoc version here, but that will require improvements in librarian (tracked in https://github.com/googleapis/librarian/issues/5818).

Fixes https://github.com/googleapis/librarian/issues/4800

### DIFF
--- a/.github/workflows/regenerate-all.yml
+++ b/.github/workflows/regenerate-all.yml
@@ -1,0 +1,69 @@
+name: Regenerate all packages after merging to main
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  regenerate:
+    runs-on: ubuntu-latest
+    env:
+      PANDOC_VERSION: 3.8.2
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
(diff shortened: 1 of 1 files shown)


jskeet:
Note: if we land https://github.com/googleapis/librarian/pull/5860, the time may be significantly reduced. (It'll be hard to say by how much until it's in a released version of librarian, but it may end up taking less than 5 minutes.)

At that point, we could consider making this a presubmit instead of a postsubmit check.

parthea (approved):
nit: specify permissions for workflow

```
permissions:
  contents: read
```

jskeet:
> permissions:
>   contents: read

Done, with `issues: write` as well.

PR #16942, opened 2026-05-05
tests: add storagebatchoperations v1 as golden API

parthea:
Towards b/508597813


chalmerlowe (approved):
LGTM.

There is an instance of a import logging statement being duplicated. That is prolly in the template somewhere and should ultimately be tracked and fixed, if not here.

PR #16945, opened 2026-05-05
fix: pass resource aliases to file-level CommonResources

### LINKED ISSUE #16952: gapic-generator: `--resource-name-alias` is ignored for file-level resource definitions, causing false-positive namespace collisions
When generating a client library that imports an external dependency with a file-level resource definition (e.g., `google.api.resource_definition`), the generator falsely throws a namespace collision error even if an explicit `--resource-name-alias` is provided via the CLI/Bazel arguments. The generator's AST builder fails to pass the alias dictionary to the dummy MessageType constructed for file-level resources.

#### Steps to Reproduce
- Set the following option in BUILD.bazel file for dialogflow:
```
opt_args = [
    "resource-name-alias=ces.googleapis.com/Tool:CesTool", 
    ],
```
- Regenerate `v2` or `v2beta1` for dialogflow using librarian.

#### Output:

```
ValueError: 
Fatal: Namespace collision detected for resource type 'Tool'.
Resources 'ces.googleapis.com/Tool' and 'dialogflow.googleapis.com/Tool' both flatten to the exact same method name.
To protect backward compatibility, explicitly alias one of these using the `--resource-name-alias` CLI parameter.
```

#### Root Cause Analysis
The bug exists in the hand-off between api.py and wrappers.py.

File-level resources bypass the standard _load_message pipeline where CLI aliases are normally injected. Instead, they are parsed in api.py inside the Proto.resource_messages property, which relies on wrappers.CommonResource.build(res) to spin up a dummy MessageType.

_No response_

### Reproduction steps: code

file: main.py
```python
   def reproduce():
    # complete code here
```

<any additional py files>

file: mydata.csv
```
alpha,1,3
beta,2,5
```

<any additional supporting files>

file: output.txtmydata.csv
```
Calculated: foo
```

file: output.txtmydata.csv
```
Calculated: bar
```

### OS & version + platform

_No response_

### Python environment

_No response_

### Python dependencies

_No response_

### Additional context

_No response_

ohmayr:
File-level resources (like imported dependencies) were previously starved of CLI aliases during dummy MessageType creation. This caused false-positive namespace collisions because the collision checker couldn't see the explicit aliases.

This wires `opts.resource_name_aliases` down through api.py into `CommonResource.build()` so the alias resolution works as originally intended.

Fixes: https://github.com/googleapis/google-cloud-python/issues/16952

### DIFF
--- a/packages/gapic-generator/gapic/schema/api.py
+++ b/packages/gapic-generator/gapic/schema/api.py
@@ -101,6 +101,7 @@ class Proto:
     meta: metadata.Metadata = dataclasses.field(
         default_factory=metadata.Metadata,
     )
+    resource_name_aliases: Mapping[str, str] = dataclasses.field(default_factory=dict)
 
     def __getattr__(self, name: str):
         return getattr(self.file_pb2, name)
@@ -159,7 +160,7 @@ class Proto:
     def resource_messages(self) -> Mapping[str, wrappers.MessageType]:
         """Return the file level resources of the proto."""
         file_resource_messages = (
-            (res.type, wrappers.CommonResource.build(res).message_type)
+            (res.type, wrappers.CommonResource.build(res, aliases=self.resource_name_aliases).message_type)
             for res in self.file_pb2.options.Extensions[
                 resource_pb2.resource_definition
             ]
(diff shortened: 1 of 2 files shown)


parthea (approved):
nit: This PR fixes issue https://github.com/googleapis/google-cloud-python/issues/16952 but does not add tests. We should add tests to avoid regressions

PR #16960, opened 2026-05-06
chore: regenerate all libraries with librarian

jskeet:
This has copied README.rst files which were modified in the package root but then not copied into docs.

Two files within google-cloud-storage are also reformatted.


chalmerlowe (approved):
LGTM

PR #16961, opened 2026-05-06
fix(bigframes): Drop support for Python <= 3.9

chalmerlowe:
> [!note]
> This is part of a larger effort across the monorepo to ensure that all packages are focused on currently supported Python runtimes. This package was already heavily focused on 3.10 support so changes should be fairly minimal. 

This PR updates `bigframes` to establish Python 3.10 as the minimum supported version, dropping support for Python 3.9 and below.

### Changes

* Configuration: Updated `noxfile.py` to remove references to Python 3.9 from the supported versions list.
* Documentation: Updated `CONTRIBUTING.rst` to use Python 3.10 in sample commands and removed stale Python 2 compatibility notes.
* Cleanup: Removed dead code in `noxfile.py` that was skipping Python 3.7, 3.8, and 3.9. 
Removed dead code in `tests/unit/test_pandas.py` checking for Python < 3.10.

Fixes internal issue: http://b/482126936 🦕

### DIFF
--- a/packages/bigframes/CONTRIBUTING.rst
+++ b/packages/bigframes/CONTRIBUTING.rst
@@ -232,11 +232,11 @@ configure them just like the System Tests.
 
    # Run all tests in a folder
    $ cd samples/snippets
-   $ nox -s py-3.8
+   $ nox -s py-3.10
 
    # Run a single sample test
    $ cd samples/snippets
-   $ nox -s py-3.8 -- -k <name of test>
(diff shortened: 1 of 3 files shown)


daniel-sanche (approved):

tswast (approved):
Thanks for the cleanup. Looks like it's already dropped here: https://github.com/googleapis/google-cloud-python/blob/main/packages/bigframes/setup.py#L160 so this is more of a chore than a fix.

PR #16976, opened 2026-05-07
refactor(auth): replace pyOpenSSL with standard ssl and cryptography

### LINKED ISSUE #16920: auth: auto-mtls raises exception if openssl not available

### LINKED ISSUE #17412: Auth: Suppress interactive OpenSSL stdin passphrase prompts during certificate loading fallbacks

### Summary of the issue

When Python's standard `ssl.SSLContext.load_cert_chain` is invoked with malformed private key bytes or encrypted PEM files without an explicit password (`password=None`), OpenSSL falls back to prompting interactively on `stdin`:

```text
Enter PEM pass phrase:
```

This causes automated CI pipelines and local test runners (`pytest`) to hang indefinitely waiting for terminal input during invalid-credential fallback tests.

It affects `compute_engine/_mtls.py` and `aio/transport/mtls.py`.

#### Proposed Solution
Prevent interactive terminal prompts across all network transports by defaulting `password` to an empty string when `passphrase` is `None`:

```python
ssl_context.load_cert_chain(certfile=cert_path, keyfile=key_path, password=passphrase or "")
```

nbayati:
Replace pyOpenSSL with standard library ssl for mTLS transport and update key decryption to use cryptography library.

This change also enhances security for handling private keys by:
- Using Linux memfd_create for RAM-backed in-memory files to avoid writing secrets to physical storage.
- Encrypting plaintext keys on-the-fly before writing to fallback temporary files on disk.
- Securely wiping temporary files with null bytes before deletion.

Fixes #16920
Closes https://github.com/googleapis/google-cloud-python/pull/16921
fixes #17412

### DIFF
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -25,32 +25,10 @@ import tempfile
 from typing import Optional
 
 from google.auth import exceptions
-import google.auth.transport._mtls_helper
 import google.auth.transport.mtls
+from google.auth.transport._mtls_helper import secure_cert_key_paths
 
 _LOGGER = logging.getLogger(__name__)
-
-
-@contextlib.contextmanager
-def _create_temp_file(content: bytes):
-    """Creates a temporary file with the given content.
-
-    Args:
-        content (bytes): The content to write to the file.
-
-    Yields:
-        str: The path to the temporary file.
-    """
-    # Create a temporary file that is readable only by the owner.
-    fd, file_path = tempfile.mkstemp()
-    try:
-        with os.fdopen(fd, "wb") as f:
-            f.write(content)
-        yield file_path
-    finally:
-        # Securely delete the file after use.
-        if os.path.exists(file_path):
-            os.remove(file_path)
 
 
 def make_client_cert_ssl_context(
@@ -71,13 +49,15 @@ def make_client_cert_ssl_context(
     Raises:
         google.auth.exceptions.TransportError: If there is an error loading the certificate.
     """
-    with _create_temp_file(cert_bytes) as cert_path, _create_temp_file(
-        key_bytes
-    ) as key_path:
+    with secure_cert_key_paths(cert_bytes, key_bytes, passphrase=passphrase) as (
+        cert_path,
+        key_path,
+        passphrase_val,
+    ):
         try:
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
             context.load_cert_chain(
-                certfile=cert_path, keyfile=key_path, password=passphrase
+                certfile=cert_path, keyfile=key_path, password=passphrase_val
             )
             return context
         except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -53,11 +53,12 @@ def make_client_cert_ssl_context(
             passphrase_val,
         ):
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
-            context.load_cert_chain(
-                certfile=cert_path,
-                keyfile=key_path,
-                password=passphrase_val or "",
-            )
+            if cert_path:
+                context.load_cert_chain(
+                    certfile=cert_path,
+                    keyfile=key_path,
+                    password=passphrase_val or "",
+                )
             return context
     except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
         raise exceptions.TransportError(
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -54,7 +54,9 @@ def make_client_cert_ssl_context(
         try:
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
             context.load_cert_chain(
-                certfile=cert_path, keyfile=key_path, password=passphrase_val
+                certfile=cert_path,
+                keyfile=key_path,
+                password=passphrase_val or "",
             )
             return context
         except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -46,12 +46,12 @@ def make_client_cert_ssl_context(
     Raises:
         google.auth.exceptions.TransportError: If there is an error loading the certificate.
     """
-    with secure_cert_key_paths(cert_bytes, key_bytes, passphrase=passphrase) as (
-        cert_path,
-        key_path,
-        passphrase_val,
-    ):
-        try:
+    try:
+        with secure_cert_key_paths(cert_bytes, key_bytes, passphrase=passphrase) as (
+            cert_path,
+            key_path,
+            passphrase_val,
+        ):
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
             context.load_cert_chain(
                 certfile=cert_path,
@@ -59,10 +59,10 @@ def make_client_cert_ssl_context(
                 password=passphrase_val or "",
             )
             return context
-        except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
-            raise exceptions.TransportError(
-                "Failed to load client certificate and key for mTLS."
-            ) from exc
+    except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
+        raise exceptions.TransportError(
+            "Failed to load client certificate and key for mTLS."
+        ) from exc
 
 
 async def _run_in_executor(func, *args):
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -17,16 +17,13 @@ Helper functions for mTLS in async for discovery of certs.
 """
 
 import asyncio
-import contextlib
 import logging
-import os
 import ssl
-import tempfile
 from typing import Optional
 
 from google.auth import exceptions
-import google.auth.transport.mtls
 from google.auth.transport._mtls_helper import secure_cert_key_paths
+import google.auth.transport.mtls
 
 _LOGGER = logging.getLogger(__name__)
 
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -55,7 +55,11 @@ def make_client_cert_ssl_context(
         ):
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
             if cert_path:
-                password = passphrase_val.decode("utf-8") if isinstance(passphrase_val, bytes) else passphrase_val
+                password = (
+                    passphrase_val.decode("utf-8")
+                    if isinstance(passphrase_val, bytes)
+                    else passphrase_val
+                )
                 context.load_cert_chain(
                     certfile=cert_path,
                     keyfile=key_path,
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -57,7 +57,7 @@ def make_client_cert_ssl_context(
                 context.load_cert_chain(
                     certfile=cert_path,
                     keyfile=key_path,
-                    password=passphrase_val or "",
+                    password=passphrase_val,
                 )
             return context
     except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
--- a/packages/google-auth/google/auth/aio/transport/mtls.py
+++ b/packages/google-auth/google/auth/aio/transport/mtls.py
@@ -17,6 +17,7 @@ Helper functions for mTLS in async for discovery of certs.
 """
 
 import asyncio
+import inspect
 import logging
 import ssl
 from typing import Optional
@@ -54,13 +55,14 @@ def make_client_cert_ssl_context(
         ):
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
             if cert_path:
+                password = passphrase_val.decode("utf-8") if isinstance(passphrase_val, bytes) else passphrase_val
                 context.load_cert_chain(
                     certfile=cert_path,
                     keyfile=key_path,
-                    password=passphrase_val,
+                    password=password,
                 )
             return context
-    except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError) as exc:
+    except (ssl.SSLError, OSError, IOError, ValueError, RuntimeError, TypeError) as exc:
         raise exceptions.TransportError(
             "Failed to load client certificate and key for mTLS."
         ) from exc
@@ -167,9 +169,9 @@ async def get_client_cert_and_key(client_cert_callback=None):
     """
     if client_cert_callback:
         result = client_cert_callback()
-        try:
+        if inspect.isawaitable(result):
             cert, key = await result
-        except TypeError:
+        else:
             cert, key = result
         return True, cert, key
 
--- a/packages/google-auth/google/auth/compute_engine/_mtls.py
+++ b/packages/google-auth/google/auth/compute_engine/_mtls.py
@@ -122,6 +122,7 @@ class MdsMtlsAdapter(HTTPAdapter):
         self.ssl_context.load_cert_chain(
             certfile=mds_mtls_config.client_combined_cert_path, password=""
         )
+        self._fallback_adapter = HTTPAdapter()
         super(MdsMtlsAdapter, self).__init__(*args, **kwargs)
 
     def init_poolmanager(self, *args, **kwargs):
@@ -147,6 +148,7 @@ class MdsMtlsAdapter(HTTPAdapter):
             requests.exceptions.SSLError,
             requests.exceptions.HTTPError,
             requests.exceptions.ConnectionError,
+            requests.exceptions.Timeout,
         ) as e:
             _LOGGER.warning(
                 "mTLS connection to Compute Engine Metadata server failed. "
@@ -158,6 +160,5 @@ class MdsMtlsAdapter(HTTPAdapter):
             http_fallback_url = urlunparse(parsed_original_url._replace(scheme="http"))
             request.url = http_fallback_url
 
-            # Use a standard HTTPAdapter for the fallback
-            http_adapter = HTTPAdapter()
-            return http_adapter.send(request, **kwargs)
+            # Use the cached standard HTTPAdapter for the fallback
+            return self._fallback_adapter.send(request, **kwargs)
--- a/packages/google-auth/google/auth/compute_engine/_mtls.py
+++ b/packages/google-auth/google/auth/compute_engine/_mtls.py
@@ -162,3 +162,7 @@ class MdsMtlsAdapter(HTTPAdapter):
 
             # Use the cached standard HTTPAdapter for the fallback
             return self._fallback_adapter.send(request, **kwargs)
+
+    def close(self):
+        self._fallback_adapter.close()
+        super(MdsMtlsAdapter, self).close()
--- a/packages/google-auth/google/auth/compute_engine/_mtls.py
+++ b/packages/google-auth/google/auth/compute_engine/_mtls.py
@@ -146,6 +146,7 @@ class MdsMtlsAdapter(HTTPAdapter):
             ssl.SSLError,
             requests.exceptions.SSLError,
             requests.exceptions.HTTPError,
+            requests.exceptions.ConnectionError,
         ) as e:
             _LOGGER.warning(
                 "mTLS connection to Compute Engine Metadata server failed. "
--- a/packages/google-auth/google/auth/compute_engine/_mtls.py
+++ b/packages/google-auth/google/auth/compute_engine/_mtls.py
@@ -120,7 +120,7 @@ class MdsMtlsAdapter(HTTPAdapter):
         self.ssl_context = ssl.create_default_context()
         self.ssl_context.load_verify_locations(cafile=mds_mtls_config.ca_cert_path)
         self.ssl_context.load_cert_chain(
-            certfile=mds_mtls_config.client_combined_cert_path
+            certfile=mds_mtls_config.client_combined_cert_path, password=""
         )
         super(MdsMtlsAdapter, self).__init__(*args, **kwargs)
 
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -152,13 +152,9 @@ class _X509Supplier(SubjectTokenSupplier):
 
     @_helpers.copy_docstring(SubjectTokenSupplier)
     def get_subject_token(self, context, request):
-        # Import OpennSSL inline because it is an extra import only required by customers
-        # using mTLS.
-        from OpenSSL import crypto
+        from cryptography import x509
 
-        leaf_cert = crypto.load_certificate(
-            crypto.FILETYPE_PEM, self._leaf_cert_callback()
-        )
+        leaf_cert = x509.load_pem_x509_certificate(self._leaf_cert_callback())
         trust_chain = self._read_trust_chain()
         cert_chain = []
 
@@ -184,9 +180,7 @@ class _X509Supplier(SubjectTokenSupplier):
         return json.dumps(cert_chain)
 
     def _read_trust_chain(self):
-        # Import OpennSSL inline because it is an extra import only required by customers
-        # using mTLS.
-        from OpenSSL import crypto
+        from cryptography import x509
 
         certificate_trust_chain = []
         # If no trust chain path was provided, return an empty list.
@@ -204,9 +198,7 @@ class _X509Supplier(SubjectTokenSupplier):
                         cert_data = b"-----BEGIN CERTIFICATE-----" + cert_block
                         try:
                             # Load each certificate and add it to the trust chain.
-                            cert = crypto.load_certificate(
-                                crypto.FILETYPE_PEM, cert_data
-                            )
+                            cert = x509.load_pem_x509_certificate(cert_data)
                             certificate_trust_chain.append(cert)
                         except Exception as e:
                             raise exceptions.RefreshError(
@@ -221,13 +213,11 @@ class _X509Supplier(SubjectTokenSupplier):
             )
 
     def _encode_cert(cert):
-        # Import OpennSSL inline because it is an extra import only required by customers
-        # using mTLS.
-        from OpenSSL import crypto
+        from cryptography.hazmat.primitives import serialization
 
-        return base64.b64encode(
-            crypto.dump_certificate(crypto.FILETYPE_ASN1, cert)
-        ).decode("utf-8")
+        return base64.b64encode(cert.public_bytes(serialization.Encoding.DER)).decode(
+            "utf-8"
+        )
 
 
 def _parse_token_data(token_content, format_type="text", subject_token_field_name=None):
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -154,10 +154,13 @@ class _X509Supplier(SubjectTokenSupplier):
     def get_subject_token(self, context, request):
         from cryptography import x509
 
-        leaf_cert_data = self._leaf_cert_callback()
-        if isinstance(leaf_cert_data, str):
-            leaf_cert_data = leaf_cert_data.encode("utf-8")
-        leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
+        try:
+            leaf_cert_data = self._leaf_cert_callback()
+            if isinstance(leaf_cert_data, str):
+                leaf_cert_data = leaf_cert_data.encode("utf-8")
+            leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
+        except Exception as e:
+            raise exceptions.RefreshError("Failed to parse leaf certificate.") from e
         trust_chain = self._read_trust_chain()
         cert_chain = []
 
@@ -210,10 +213,10 @@ class _X509Supplier(SubjectTokenSupplier):
                                 )
                             ) from e
                 return certificate_trust_chain
-        except FileNotFoundError:
+        except OSError as e:
             raise exceptions.RefreshError(
-                "Trust chain file '{}' was not found.".format(self._trust_chain_path)
-            )
+                "Error accessing trust chain file '{}'.".format(self._trust_chain_path)
+            ) from e
 
     def _encode_cert(cert):
         from cryptography.hazmat.primitives import serialization
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -213,6 +213,10 @@ class _X509Supplier(SubjectTokenSupplier):
                                 )
                             ) from e
                 return certificate_trust_chain
+        except FileNotFoundError as e:
+            raise exceptions.RefreshError(
+                "Trust chain file '{}' was not found.".format(self._trust_chain_path)
+            ) from e
         except OSError as e:
             raise exceptions.RefreshError(
                 "Error accessing trust chain file '{}'.".format(self._trust_chain_path)
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -158,7 +158,7 @@ class _X509Supplier(SubjectTokenSupplier):
             leaf_cert_data = self._leaf_cert_callback()
         except Exception as e:
             raise exceptions.RefreshError("Failed to retrieve leaf certificate.") from e
-            
+
         try:
             if isinstance(leaf_cert_data, str):
                 leaf_cert_data = leaf_cert_data.encode("utf-8")
@@ -168,18 +168,18 @@ class _X509Supplier(SubjectTokenSupplier):
         trust_chain = self._read_trust_chain()
         cert_chain = []
 
-        cert_chain.append(_X509Supplier._encode_cert(leaf_cert))
+        cert_chain.append(_encode_cert(leaf_cert))
 
         if trust_chain is None or len(trust_chain) == 0:
             return json.dumps(cert_chain)
 
         # Append the first cert if it is not the leaf cert.
-        first_cert = _X509Supplier._encode_cert(trust_chain[0])
+        first_cert = _encode_cert(trust_chain[0])
         if first_cert != cert_chain[0]:
             cert_chain.append(first_cert)
 
         for i in range(1, len(trust_chain)):
-            encoded = _X509Supplier._encode_cert(trust_chain[i])
+            encoded = _encode_cert(trust_chain[i])
             # Check if the current cert is the leaf cert and raise an exception if it is.
             if encoded == cert_chain[0]:
                 raise exceptions.RefreshError(
@@ -226,12 +226,13 @@ class _X509Supplier(SubjectTokenSupplier):
                 "Error accessing trust chain file '{}'.".format(self._trust_chain_path)
             ) from e
 
-    def _encode_cert(cert):
-        from cryptography.hazmat.primitives import serialization
 
-        return base64.b64encode(cert.public_bytes(serialization.Encoding.DER)).decode(
-            "utf-8"
-        )
+def _encode_cert(cert):
+    from cryptography.hazmat.primitives import serialization
+
+    return base64.b64encode(cert.public_bytes(serialization.Encoding.DER)).decode(
+        "utf-8"
+    )
 
 
 def _parse_token_data(token_content, format_type="text", subject_token_field_name=None):
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -156,6 +156,10 @@ class _X509Supplier(SubjectTokenSupplier):
 
         try:
             leaf_cert_data = self._leaf_cert_callback()
+        except Exception as e:
+            raise exceptions.RefreshError("Failed to retrieve leaf certificate.") from e
+            
+        try:
             if isinstance(leaf_cert_data, str):
                 leaf_cert_data = leaf_cert_data.encode("utf-8")
             leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
--- a/packages/google-auth/google/auth/identity_pool.py
+++ b/packages/google-auth/google/auth/identity_pool.py
@@ -154,7 +154,10 @@ class _X509Supplier(SubjectTokenSupplier):
     def get_subject_token(self, context, request):
         from cryptography import x509
 
-        leaf_cert = x509.load_pem_x509_certificate(self._leaf_cert_callback())
+        leaf_cert_data = self._leaf_cert_callback()
+        if isinstance(leaf_cert_data, str):
+            leaf_cert_data = leaf_cert_data.encode("utf-8")
+        leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
         trust_chain = self._read_trust_chain()
         cert_chain = []
 
--- a/packages/google-auth/google/auth/transport/_custom_tls_signer.py
+++ b/packages/google-auth/google/auth/transport/_custom_tls_signer.py
@@ -23,8 +23,6 @@ import logging
 import os
 import sys
 
-import cffi  # type: ignore
-
 from google.auth import exceptions
 
 _LOGGER = logging.getLogger(__name__)
@@ -43,11 +41,6 @@ SIGN_CALLBACK_CTYPE = ctypes.CFUNCTYPE(
     ctypes.POINTER(ctypes.c_ubyte),  # tbs
     ctypes.c_size_t,  # tbs_len
 )
-
-
-# Cast SSL_CTX* to void*
-def _cast_ssl_ctx_to_void_p_pyopenssl(ssl_ctx):
-    return ctypes.cast(int(cffi.FFI().cast("intptr_t", ssl_ctx)), ctypes.c_void_p)
 
 
 # Cast SSL_CTX* to void*
@@ -274,7 +267,7 @@ class CustomTlsSigner(object):
             if not self._offload_lib.ConfigureSslContext(
                 self._sign_callback,
                 ctypes.c_char_p(self._cert),
-                _cast_ssl_ctx_to_void_p_pyopenssl(ctx._ctx._context),
+                _cast_ssl_ctx_to_void_p_stdlib(ctx),
             ):
                 raise exceptions.MutualTLSChannelError(
                     "failed to configure ECP Offload SSL context"
--- a/packages/google-auth/google/auth/transport/_custom_tls_signer.py
+++ b/packages/google-auth/google/auth/transport/_custom_tls_signer.py
@@ -45,6 +45,10 @@ SIGN_CALLBACK_CTYPE = ctypes.CFUNCTYPE(
 
 # Cast SSL_CTX* to void*
 def _cast_ssl_ctx_to_void_p_stdlib(context):
+    if sys.implementation.name != "cpython" or hasattr(sys, "getobjects"):
+        raise exceptions.MutualTLSChannelError(
+            "Custom TLS signing is only supported on standard release CPython runtimes."
+        )
     return ctypes.c_void_p.from_address(
         id(context) + ctypes.sizeof(ctypes.c_void_p) * 2
     )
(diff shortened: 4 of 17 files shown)


parthea:
@nbayati , Please could you address the failing presubmits?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,229 @@
...
+                    yield cert_path or cert, key_path or key, passphrase
+                    return
+            finally:
+                import sys
+
+                exc_info = sys.exc_info()
+                cm.__exit__(
+                    *(exc_info if exc_info[0] is not None else (None, None, None))
+                )
+            # If verification failed, fall through to Tier 3.
```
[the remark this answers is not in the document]
nbayati: We cannot use a standard `with` block wrapped in `except OSError` here. Because this is a generator, wrapping the `yield` statement in `except OSError` would catch and swallow any `OSError` (like socket/connection errors) raised by the caller during execution, incorrectly triggering the fallback to Tier 3. Keeping the manual `__enter__` and `__exit__` allows us to isolate the `except OSError` check solely to the initialization phase.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,229 @@
...
+    import os
+
+    cleanup_fds = []
+    cert_path, key_path = None, None
+
+    try:
+        if cert_bytes is not None:
+            # MFD_CLOEXEC prevents FD leaks to spawned subprocesses.
+            fd_cert = os.memfd_create("mtls_cert", os.MFD_CLOEXEC)
+            os.write(fd_cert, cert_bytes)
```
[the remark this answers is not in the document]
nbayati: Done. Replaced the direct `os.write` with `os.fdopen(..., closefd=False)` to prevent potential partial writes and ensure the full certificate is written.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,229 @@
...
+            cert_path = f"/proc/self/fd/{fd_cert}"
+            cleanup_fds.append(fd_cert)
+
+        if key_bytes is not None:
+            fd_key = os.memfd_create("mtls_key", os.MFD_CLOEXEC)
+            os.write(fd_key, key_bytes)
```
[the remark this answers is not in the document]
nbayati: Done. Applied the same change here, using `os.fdopen(..., closefd=False)` to ensure the private key bytes are fully written.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+        key (Union[str, bytes]): Private key path or raw PEM content bytes.
+        passphrase (Optional[bytes]): Optional passphrase for the private key.
+
+    Yields:
+        Tuple[str, str, Optional[bytes]]: The certificate path, key path, and
+            the passphrase needed to load the key (either the user's original,
+            or the newly generated one if Tier 3 had to encrypt the key).
+    """
+    import os
+    import sys
```
daniel-sanche: can we keep imports at the top of the file?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+
+    # Tier 1: Pass-through (No-op). If the caller already provided file paths,
+    # we yield them directly to avoid any unnecessary file creation.
+    if isinstance(cert, str) and isinstance(key, str):
+        yield cert, key, passphrase
+        return
+
+    cert_bytes = cert if isinstance(cert, bytes) else None
+    key_bytes = key if isinstance(key, bytes) else None
```
daniel-sanche, line 141: what if a user passed one as a string path, and one as bytes? It seems like that isn't supported? Should we mention that in the docstrings?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+def _memfd_cert_key_paths(
+    cert_bytes: Optional[bytes], key_bytes: Optional[bytes]
+) -> Generator[Tuple[Optional[str], Optional[str]], None, None]:
+    """Creates secure, in-memory virtual files on Linux using memfd_create.
+
+    Yields:
+        Tuple[Optional[str], Optional[str]]: In-memory file paths pointing to
+            the active descriptors (e.g., '/proc/self/fd/3').
+    """
+    import os
```
daniel-sanche: can we keep imports at the top level?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+            with os.fdopen(fd_cert, "wb", closefd=False) as f:
+                f.write(cert_bytes)
+            cert_path = f"/proc/self/fd/{fd_cert}"
+
+        if key_bytes is not None:
+            fd_key = os.memfd_create("mtls_key", os.MFD_CLOEXEC)  # type: ignore[attr-defined]
+            cleanup_fds.append(fd_key)
+            with os.fdopen(fd_key, "wb", closefd=False) as f:
+                f.write(key_bytes)
+            key_path = f"/proc/self/fd/{fd_key}"
```
daniel-sanche: nit: I wonder if making a helper or loop for these could be cleaner?

packages/google-auth/google/auth/transport/_mtls_helper.py
daniel-sanche: It might be good to leave an in-line comment about how Nones are treated here too. After looking at some of the helpers, it seems like it will result in an empty output?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+@contextlib.contextmanager
+def _memfd_cert_key_paths(
+    cert_bytes: Optional[bytes], key_bytes: Optional[bytes]
+) -> Generator[Tuple[Optional[str], Optional[str]], None, None]:
+    """Creates secure, in-memory virtual files on Linux using memfd_create.
+
+    Yields:
+        Tuple[Optional[str], Optional[str]]: In-memory file paths pointing to
+            the active descriptors (e.g., '/proc/self/fd/3').
+    """
```
daniel-sanche, line 241: It seems like secure_cert_key_paths expects this one to raise OSErrors? If so, we should probably mention them in a Raises: section in the docstring

(although I had a [separate suggestion](https://github.com/googleapis/google-cloud-python/pull/16976/changes#r3432059015) about those OSErrors too)

packages/google-auth/google/auth/transport/_mtls_helper.py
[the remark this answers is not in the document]
daniel-sanche: It seems like a better approach would be to disambiguate between those two error types. That would let us keep this function more straight forward

Could we approach this by adding a new `try/except OsError` block before the yield in _memfd_cert_key_paths, re-raise those as a custom type, and then just catch that custom type here?

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
 )
 
 
+@contextlib.contextmanager
+def secure_cert_key_paths(
+    cert: Union[str, bytes],
+    key: Union[str, bytes],
+    passphrase: Optional[bytes] = None,
+) -> Generator[Tuple[str, str, Optional[bytes]], None, None]:
```
daniel-sanche: It might be useful to mention in the docstring that this was implemented as a context manager/generator to clean up after itself between yield and exit. 

A lot of the clean-up code is hidden in delegated calls, so it may not be obvious to future readers

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+            encrypted_key_bytes, new_passphrase = _encrypt_key_if_plaintext(
+                key_bytes, passphrase
+            )
+
+            fd, key_path = tempfile.mkstemp(dir=tmp_dir)
+            cleanup_files.append(key_path)
+            with os.fdopen(fd, "wb") as f:
+                f.write(encrypted_key_bytes)
+                f.flush()
+                os.fsync(f.fileno())
```
daniel-sanche: This might be worth replacing with a loop or helper

daniel-sanche (changes requested):

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -174,22 +174,29 @@ def _make_mutual_tls_http(cert, key):
...
-    urllib3.contrib.pyopenssl.inject_into_urllib3()
     ctx = urllib3.util.ssl_.create_urllib3_context()
     ctx.load_verify_locations(cafile=certifi.where())
 
-    pkey = crypto.load_privatekey(crypto.FILETYPE_PEM, key)
-    x509 = crypto.load_certificate(crypto.FILETYPE_PEM, cert)
-
-    ctx._ctx.use_certificate(x509)
-    ctx._ctx.use_privatekey(pkey)
+    with _mtls_helper.secure_cert_key_paths(cert, key) as (
```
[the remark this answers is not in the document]
nbayati: _mtls_helper is imported at the module level on line 56 (from google.auth.transport import _mtls_helper), so using it directly here is correct and does not raise a NameError.

packages/google-auth/google/auth/identity_pool.py
```diff
@@ -152,13 +152,9 @@ def __init__(self, trust_chain_path, leaf_cert_callback):
...
     def get_subject_token(self, context, request):
-        # Import OpennSSL inline because it is an extra import only required by customers
-        # using mTLS.
-        from OpenSSL import crypto
+        from cryptography import x509
 
-        leaf_cert = crypto.load_certificate(
-            crypto.FILETYPE_PEM, self._leaf_cert_callback()
-        )
+        leaf_cert = x509.load_pem_x509_certificate(self._leaf_cert_callback())
```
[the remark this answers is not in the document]
nbayati: Done. Added defensive type-checking and encoding to ensure leaf_cert_data is converted to bytes if the callback returns a str.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+
+    Yields:
+        Tuple[Optional[str], Optional[str], Optional[bytes]]: The temporary file
+            paths and the passphrase needed to load the key.
+    """
+    import os
+    import tempfile
+
+    # Prioritize RAM-backed /dev/shm to avoid writing secrets to physical storage.
+    tmp_dir = "/dev/shm" if os.path.isdir("/dev/shm") else None
```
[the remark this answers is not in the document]
nbayati: Done. Added the os.access check to verify write permissions on /dev/shm before using it.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +67,237 @@
...
+                        str, key_path or key
+                    ), passphrase
+                    return
+            finally:
+                import sys
+
+                exc_info = sys.exc_info()
+                cm.__exit__(
+                    *(exc_info if exc_info[0] is not None else (None, None, None))
+                )
```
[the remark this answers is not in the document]
nbayati: Done. Simplified the context manager exit call using *sys.exc_info().

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done!

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done!

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Mixed inputs (e.g. one string path and one bytes) are currently supported. If one parameter is a string path or None, its bytes representation will be None, so only the bytes parameter is written to a temp file, and the string/None parameter is yielded back as-is. I've updated the docstrings to clarify this behavior, and added inline comments on how None values are handled.

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done! Added a loop.

nbayati:
I tested the refactored paths manually by using the External Account X.509 provider to exchange certificates for access tokens and retrieve GCS buckets across different transports and credential initiation paths. The default transport (requests), urllib3, and aiohttp (async) are all verified and working as expected after the refactoring. I've also verified the encrypted private key loading/decryption flow using the new cryptography-based helper.

A copy of the test script can be found at: paste/5209556048281600
Results: screenshot/7tvfZhJ6qoYEGbY

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,234 @@
...
+            else:
+                paths.append(None)
+
+        cert_path, key_path = paths
+        yield cert_path, key_path, new_passphrase
+    finally:
+        for file_path in cleanup_files:
+            try:
+                # Wiping the private key with null bytes before removal.
+                if file_path == key_path:
```
macastelaz: While a really big nit (because ultimately the written key data was encrypted anyways) but if writing the key_data fails above (the second iteration of the for loop) during f.write (e.g. a partial write failure), key_path will never equal file_path as we never get to "cert_path, key_path = paths"

You can leave it as is if you'd prefer since the risk of exposing raw key_data isn't a concern, but if you want to fix this you could do something like 

cleanup_files = [None, None]                                         
                                                                             
        try:                                                                 
            paths = []                                                       
            for i, data in enumerate([cert_data, key_data]):                 
                if data is not None:                                         
                    fd, path = tempfile.mkstemp(dir=tmp_dir)                 
                    cleanup_files[i] = path   # <--- Indexed exactly like the
  data array!                                                                
                    with os.fdopen(fd, "wb") as f:                           
                        ...

And then the finally block would become something like

finally:                                                             
            cert_cleanup_path = cleanup_files[0]                             
            key_cleanup_path = cleanup_files[1]                              
                                                                             
            try:                                                             
                if key_cleanup_path:                                         
                    _secure_wipe_and_remove(key_cleanup_path)                
                if cert_cleanup_path and os.path.exists(cert_cleanup_path):  
                    os.remove(cert_cleanup_path)                             
            except OSError:                                                  
                pass

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,234 @@
...
+
+    pass
+
+
+@contextlib.contextmanager
+def secure_cert_key_paths(
+    cert: Union[bytes, str, None],
+    key: Union[bytes, str, None],
+    passphrase: Optional[bytes] = None,
+) -> Generator[Tuple[str, str, Optional[bytes]], None, None]:
```
macastelaz: super nit on the return type here: "Generator[Tuple[str, str, Optional[bytes]], None, None]"

If cert is none and cert_path is none, I believe this could ultimately return something like "Tuple[Optional[str], Optional[str], Optional[bytes]]"

packages/google-auth/google/auth/transport/requests.py
```diff
@@ -204,30 +204,41 @@ class _MutualTlsAdapter(requests.adapters.HTTPAdapter):
...
-        ctx_poolmanager._ctx.use_certificate(x509)
-        ctx_poolmanager._ctx.use_privatekey(pkey)
-        self._ctx_poolmanager = ctx_poolmanager
 
         ctx_proxymanager = create_urllib3_context()
         ctx_proxymanager.load_verify_locations(cafile=certifi.where())
-        ctx_proxymanager._ctx.use_certificate(x509)
-        ctx_proxymanager._ctx.use_privatekey(pkey)
+
+        with _mtls_helper.secure_cert_key_paths(cert, key) as (
```
macastelaz: This can throw but it is outside of the try block - is that intentional to let the error here surface out of the library? This pattern exists in a few other places too calling _mtls_helper.secure_cert_key_paths

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done!

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: done!

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,234 @@
...
+    key_bytes = key if isinstance(key, bytes) else None
+
+    # Tier 2: Linux RAM-backed virtual files. If supported by the OS, we write
+    # the bytes to anonymous in-memory files using memfd_create. This yields
+    # /proc/self/fd/... paths, keeping the private key entirely in memory.
+    if sys.platform == "linux" and hasattr(os, "memfd_create"):
+        try:
+            with _memfd_cert_key_paths(cert_bytes, key_bytes) as (cert_path, key_path):
+                # Handle cases where path exists but might be restricted.
+                if (cert_path is None or os.path.exists(cert_path)) and (
```
lsirac, line 150: In `secure_cert_key_paths`, checking path existence via `os.path.exists()` is insufficient in environments enforcing restrictive Linux Security Module (LSM) policies (such as SELinux, AppArmor, or gVisor under Docker/Kubernetes). If an LSM policy permits `stat` calls on `/proc` paths but blocks `open` or `read` calls on them, `os.path.exists` will return `True` for `/proc/self/fd/{fd}`. However, when the caller subsequently executes `SSLContext.load_cert_chain` inside the yielded block, it will raise an unhandled `PermissionError` (a subclass of `OSError`). Because the context manager has already yielded, this exception propagates outside `secure_cert_key_paths`, causing a fatal crash instead of gracefully falling back to Tier 3.

**Remediation:**
Pre-verify readability of the file paths prior to yielding:
```python
def _can_read(path: Optional[str]) -> bool:
    if path is None:
        return True
    try:
        with open(path, "rb"):
            pass
        return True
    except OSError:
        return False

# ... inside secure_cert_key_paths:
if sys.platform == "linux" and hasattr(os, "memfd_create"):
    try:
        with _memfd_cert_key_paths(cert_bytes, key_bytes) as (cert_path, key_path):
            if (cert_path is None or os.path.exists(cert_path)) and (
                key_path is None or os.path.exists(key_path)
            ):
                if _can_read(cert_path) and _can_read(key_path):
                    yield cast(str, cert_path or cert), cast(str, key_path or key), passphrase
                    return
    except _MemfdCreationError:
        pass
```

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,234 @@
...
+    passphrase: Optional[bytes],
+) -> Generator[Tuple[Optional[str], Optional[str], Optional[bytes]], None, None]:
+    """Creates secure temporary file paths on disk, encrypting private keys.
+
+    Yields:
+        Tuple[Optional[str], Optional[str], Optional[bytes]]: The temporary file
+            paths and the passphrase needed to load the key.
+    """
+    # Prioritize RAM-backed /dev/shm to avoid writing secrets to physical storage.
+    tmp_dir = (
```
lsirac, line 306: In `_tempfile_cert_key_paths`, the fallback mechanism sets `tmp_dir = "/dev/shm" if os.path.isdir("/dev/shm") and os.access("/dev/shm", os.W_OK) else None`. However, it does not check if `/dev/shm` is out of space (common in container environments with strict shared memory limits) or has restricted permissions that block file writes. If `tempfile.mkstemp(dir=tmp_dir)` raises an `OSError` (such as `No space left on device` or `PermissionError`), it fails immediately because `tempfile` does not automatically fallback to the system-wide `/tmp` when `dir` is explicitly specified.

**Remediation:**
Catch `OSError` on the temporary file creation and fallback to `None` (which lets `tempfile` use `/tmp` or another safe default directory):
```python
try:
    fd, path = tempfile.mkstemp(dir=tmp_dir)
except OSError:
    fd, path = tempfile.mkstemp(dir=None)
```

packages/google-auth/google/auth/transport/_custom_tls_signer.py
```diff
@@ -274,7 +267,7 @@ def attach_to_ssl_context(self, ctx):
             if not self._offload_lib.ConfigureSslContext(
                 self._sign_callback,
                 ctypes.c_char_p(self._cert),
-                _cast_ssl_ctx_to_void_p_pyopenssl(ctx._ctx._context),
+                _cast_ssl_ctx_to_void_p_stdlib(ctx),
```
lsirac, line 284: The stdlib C pointer arithmetic hack `id(context) + ctypes.sizeof(ctypes.c_void_p) * 2` used in `_cast_ssl_ctx_to_void_p_stdlib` is highly brittle and CPython-specific.
1. On **PyPy**, `id()` returns virtual IDs instead of actual memory addresses, leading to immediate segmentation faults or undefined memory access.
2. On **CPython debug builds** (compiled with `Py_DEBUG` / `Py_TRACE_REFS`), the object header is larger due to reference tracing pointers, which shifts the offset of the raw OpenSSL `ctx` pointer from `2 * sizeof(void*)` to `4 * sizeof(void*)`. Running this on a debug build will cast the `ob_type` pointer to a void pointer, causing a crash.

**Remediation:**
Add a strict environment check to verify a standard release CPython environment before performing pointer casting, raising a graceful `MutualTLSChannelError` otherwise:
```python
import sys
if sys.implementation.name != "cpython" or hasattr(sys, "getobjects"):
    raise exceptions.MutualTLSChannelError(
        "Custom TLS signing is only supported on standard release CPython runtimes."
    )
```

packages/google-auth/google/auth/identity_pool.py
```diff
@@ -152,13 +152,12 @@ def __init__(self, trust_chain_path, leaf_cert_callback):
...
-        from OpenSSL import crypto
+        from cryptography import x509
 
-        leaf_cert = crypto.load_certificate(
-            crypto.FILETYPE_PEM, self._leaf_cert_callback()
-        )
+        leaf_cert_data = self._leaf_cert_callback()
+        if isinstance(leaf_cert_data, str):
+            leaf_cert_data = leaf_cert_data.encode("utf-8")
+        leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
```
lsirac: In `get_subject_token`, `leaf_cert = x509.load_pem_x509_certificate(self._leaf_cert_callback())` can throw a `ValueError` or `TypeError` if the user's callback returns corrupt, incomplete, or empty PEM data. Because this exception is not caught, it escapes the credentials refresh lifecycle as a raw exception, which violates the `google-auth` interface contract requiring credentials refreshes to raise `exceptions.RefreshError`.

**Remediation:**
Wrap the callback execution and certificate loading in a try-except block:
```python
try:
    leaf_cert_data = self._leaf_cert_callback()
    if isinstance(leaf_cert_data, str):
        leaf_cert_data = leaf_cert_data.encode("utf-8")
    leaf_cert = x509.load_pem_x509_certificate(leaf_cert_data)
except Exception as e:
    raise exceptions.RefreshError("Failed to parse leaf certificate.") from e
```

packages/google-auth/google/auth/identity_pool.py
```diff
@@ -204,9 +201,7 @@ def _read_trust_chain(self):
                         cert_data = b"-----BEGIN CERTIFICATE-----" + cert_block
                         try:
                             # Load each certificate and add it to the trust chain.
-                            cert = crypto.load_certificate(
-                                crypto.FILETYPE_PEM, cert_data
-                            )
+                            cert = x509.load_pem_x509_certificate(cert_data)
```
lsirac, line 211: In `_read_trust_chain`, catching only `FileNotFoundError` lets `PermissionError` (when a file is present but unreadable) or other system-level `OSError` exceptions leak. This violates the library's error contract, as file access failures should map to `exceptions.RefreshError`.

**Remediation:**
Catch `OSError` instead of just `FileNotFoundError`:
```python
except OSError as e:
    raise exceptions.RefreshError(
        "Error accessing trust chain file '{}'.".format(self._trust_chain_path)
    ) from e
```

packages/google-auth/setup.py
```diff
@@ -35,10 +35,7 @@
 
 reauth_extra_require = ["pyu2f>=0.1.5"]
 
-# TODO(https://github.com/googleapis/google-auth-library-python/issues/1738): Add bounds for pyopenssl dependency.
-enterprise_cert_extra_require = ["pyopenssl"]
-
-pyopenssl_extra_require = ["pyopenssl>=20.0.0"]
+enterprise_cert_extra_require = cryptography_base_require
```
lsirac, line 38: In `setup.py`, `cryptography_base_require` is defined as a list but nested inside `DEPENDENCIES` without being unpacked:
```python
DEPENDENCIES = (
    "pyasn1-modules>=0.2.1",
    cryptography_base_require,
)
```
This passes a nested sequence `("pyasn1-modules>=0.2.1", ["cryptography >= 38.0.3"])` to `install_requires`. This is highly non-standard and causes a fatal `TypeError` under standard Python packaging and installation tools (such as pip), completely preventing package installation.

**Remediation:**
Unpack the list using the `*` operator:
```python
DEPENDENCIES = (
    "pyasn1-modules>=0.2.1",
    *cryptography_base_require,
)
```

packages/google-auth/google/auth/compute_engine/_mtls.py
```diff
@@ -120,7 +120,7 @@ def __init__(
         self.ssl_context = ssl.create_default_context()
         self.ssl_context.load_verify_locations(cafile=mds_mtls_config.ca_cert_path)
         self.ssl_context.load_cert_chain(
-            certfile=mds_mtls_config.client_combined_cert_path
+            certfile=mds_mtls_config.client_combined_cert_path, password=""
```
lsirac, line 123: In `google/auth/compute_engine/_mtls.py`, `MdsMtlsAdapter.send()` implements a fallback to standard HTTP if mTLS fails. It catches `ssl.SSLError`, `requests.exceptions.SSLError`, and `requests.exceptions.HTTPError`. However, if the metadata server closes the port or terminates negotiation abruptly, `requests` raises `requests.exceptions.ConnectionError` (or `MaxRetryError` from `urllib3`), which does not inherit from any of the caught exceptions. This bypasses the fallback and crashes the client on connection resets.

**Remediation:**
Include `requests.exceptions.ConnectionError` in the caught exceptions:
```python
except (
    ssl.SSLError,
    requests.exceptions.SSLError,
    requests.exceptions.HTTPError,
    requests.exceptions.ConnectionError,
) as e:
```

lsirac (changes requested):

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Good catch, fixed!

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done!

packages/google-auth/google/auth/transport/requests.py
nbayati: Not intentional, thanks for catching this. It appears to be an existing issue we hadn't noticed but definitely needed fixing. The secure_cert_key_paths context manager is now properly wrapped inside the try/except block across all transport adapters (requests.py, urllib3.py, and aio/transport/mtls.py) to safely surface the errors.

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Good catch! Done!

packages/google-auth/google/auth/transport/_custom_tls_signer.py
nbayati: Done! I'm glad we're catching some of these existing issues and cleaning things up as part of this PR 🎉

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done!

packages/google-auth/setup.py
nbayati: Fixed!

packages/google-auth/google/auth/compute_engine/_mtls.py
nbayati: Done!

packages/google-auth/google/auth/identity_pool.py
nbayati: Good catch, done!

packages/google-auth/google/auth/identity_pool.py
nbayati: Fixed!

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -174,22 +174,29 @@ def _make_mutual_tls_http(cert, key):
...
+    try:
+        with _mtls_helper.secure_cert_key_paths(cert, key) as (
+            cert_path,
+            key_path,
+            passphrase,
+        ):
+            ctx.load_cert_chain(
+                certfile=cert_path,
+                keyfile=key_path,
+                password=passphrase or "",
```
macastelaz: Nit: Based on the definition of load_cert_chain, password can safely be None (https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain) so I don't think we need to do this type casting here.

macastelaz (approved):

lsirac (commented):
#### 1. Critical Packaging Bug in `setup.py`
* **File:** `packages/google-auth/setup.py` (PR-level)
* **Technical Note:**
  The list `cryptography_base_require` is placed directly inside the `DEPENDENCIES` tuple without unpacking on line 27. This results in a nested structure: `("pyasn1-modules>=0.2.1", ["cryptography >= 38.0.3"])`. When `setuptools` parses `install_requires` during packaging or installation, it iterates over the requirements and expects each item to be a flat string specifier. Encountering a nested list triggers a `TypeError` (e.g., `TypeError: expected string or bytes-like object`), completely blocking building or installing the package from source. Unpack the list inside `DEPENDENCIES`:
  ```python
  DEPENDENCIES = (
      "pyasn1-modules>=0.2.1",
      *cryptography_base_require,
  )
  ```

---

#### 2. Standard SSLContext Pointer Alignment and Safety
* **File:** `packages/google-auth/google/auth/transport/_custom_tls_signer.py` (Line 270)
* **Technical Note:**
  The pointer calculation in `_cast_ssl_ctx_to_void_p_stdlib` relies on standard CPython release alignment `id(context) + ctypes.sizeof(ctypes.c_void_p) * 2` to obtain the raw pointer to the underlying OpenSSL structure `SSL_CTX`. This offset assumes a 2-pointer `PyObject_HEAD` header structure. This alignment is highly fragile and shifts under debug builds with `Py_TRACE_REFS` enabled (increasing header size to 4 pointers), free-threaded CPython (Python 3.13+, PEP 703), or alternative runtimes (PyPy). Reading from the wrong offset results in arbitrary memory pointer retrieval and uncatchable segmentation faults. Reject unsupported environments or enforce safe alternative C-API casting.

---

#### 3. Missing Type Validation on Context Parameter
* **File:** `packages/google-auth/google/auth/transport/_custom_tls_signer.py` (Line 270)
* **Technical Note:**
  In `attach_to_ssl_context`, there is no type validation on `ctx`. If a caller passes `None` or an invalid object type, `_cast_ssl_ctx_to_void_p_stdlib` calculates a memory address offset beyond the bounds of the object structure and reads raw heap garbage as an `SSL_CTX` pointer. When passed to the Go offload library `ConfigureSslContext`, this bogus address is dereferenced, causing a hard segmentation fault that crashes the entire Python host process instead of raising a clean Python `TypeError`. Add an explicit assertion before pointer casting:
  ```python
  if not isinstance(ctx, ssl.SSLContext):
      raise TypeError("ctx must be an instance of ssl.SSLContext")
  ```

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,247 @@
...
+        key_path,
+        new_passphrase,
+    ):
+        yield cast(str, cert_path or cert), cast(str, key_path or key), new_passphrase
+
+
+def _encrypt_key_if_plaintext(
+    key_bytes: bytes, passphrase: Optional[bytes]
+) -> Tuple[bytes, Optional[bytes]]:
+    """Encrypts a plaintext PEM key if necessary, returning the bytes and passphrase.
```
lsirac, line 175: We are returning plaintext if encryption fails. Is that intentional? If so should be documented

packages/google-auth/google/auth/transport/requests.py
```diff
@@ -471,10 +473,14 @@ def configure_mtls_channel(self, client_cert_callback=None):
         except (
             exceptions.ClientCertError,
             ImportError,
-            OpenSSL.crypto.Error,
+            ValueError,
         ) as caught_exc:
+            self._is_mtls = False
```
lsirac: Can we make this atomic: build the new adapter first, then update `is_mtls` and mount it only after success?

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -357,13 +359,18 @@ def configure_mtls_channel(self, client_cert_callback=None):
...
             else:
                 self.http = _make_default_http()
+                self._is_mtls = False
         except (
             exceptions.ClientCertError,
             ImportError,
-            OpenSSL.crypto.Error,
+            ValueError,
         ) as caught_exc:
+            self._is_mtls = False
```
lsirac: Same issue here - I think if mTLS was already configured and a later configure call fails, `_is_mtls` becomes false but `self.http` is still the old mTLS PoolManager.

packages/google-auth/google/auth/identity_pool.py
```diff
@@ -152,13 +152,15 @@ def __init__(self, trust_chain_path, leaf_cert_callback):
...
     def get_subject_token(self, context, request):
-        # Import OpennSSL inline because it is an extra import only required by customers
-        # using mTLS.
-        from OpenSSL import crypto
+        from cryptography import x509
 
-        leaf_cert = crypto.load_certificate(
-            crypto.FILETYPE_PEM, self._leaf_cert_callback()
-        )
+        try:
```
lsirac, line 157: This wraps too much code in the parse error.

The callback can fail for config or file-read reasons, but because it is inside this `try`, those errors become `Failed to parse leaf certificate`.

Can we call the callback before this parse `try`, so only actual cert parsing errors get this message?

packages/google-auth/google/auth/transport/urllib3.py
nbayati: Done!

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -65,6 +70,247 @@
...
+    """Overwrites a file with null bytes before deleting it.
+
+    This is an extra security measure to make file recovery harder. However, on modern
+    solid-state drives (SSDs), the hardware optimizes where data is written, meaning
+    the original private key bytes might still physically remain on the storage chips
+    until the drive cleans them up.
+    """
+    import os
+
+    if not os.path.exists(file_path):
```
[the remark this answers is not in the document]
nbayati: Done!

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Yes, that is intentional. In _encrypt_key_if_plaintext, we attempt to parse the key as a plaintext PEM key to encrypt it. If parsing fails with TypeError (key is already encrypted) or ValueError (invalid or unsupported PEM structure), we catch these exceptions and return the key as-is to allow the underlying SSL library's cert loading to handle the key structure. I have added a docstring update to clarify this behavior.

packages/google-auth/google/auth/transport/requests.py
nbayati: Good catch! PR 17387 recently merged into main and refactored this block across all transports to be atomic. I've rebased this branch onto the latest main and ensured that the atomic structure for configure_mtls_channel is preserved while dropping the PyOpenSSL dependencies.

packages/google-auth/google/auth/transport/urllib3.py
nbayati: Same here, it was resolved in PR 17387, and I've rebased this branch on top of the latest main.

packages/google-auth/google/auth/identity_pool.py
nbayati: Resolved by splitting the exception handling blocks: one wrapping the cert callback execution (raising a retrieval error) and another wrapping the certificate parsing (raising a parse error).

packages/google-auth/tests/transport/test__custom_tls_signer.py
```diff
@@ -264,3 +259,39 @@ def test_custom_tls_signer_failed_to_attach_no_libs():
         signer_object._signer_lib = None
         signer_object.attach_to_ssl_context(mock.MagicMock())
     assert excinfo.match("Invalid ECP configuration.")
+
+
+def test_cast_ssl_ctx_to_void_p_stdlib_success():
```
lsirac, line 265: test_cast_ssl_ctx_to_void_p_stdlib_success passes a mock.MagicMock() instance into _cast_ssl_ctx_to_void_p_stdlib and asserts isinstance(res, ctypes.c_void_p). Because ctypes.c_void_p.from_address accepts any integer memory address, the test passes on mock objects while hiding the fact that using the pointer in C causes a crash. This test should pass a real ssl.SSLContext() instance.

packages/google-auth/google/auth/transport/_mtls_helper.py
```diff
@@ -71,6 +76,247 @@
...
+        )
+        encrypted_content = pkey.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PrivateFormat.PKCS8,
+            encryption_algorithm=serialization.BestAvailableEncryption(
+                target_passphrase
+            ),
+        )
+        return encrypted_content, target_passphrase
+    except (ValueError, TypeError):
```
lsirac: _encrypt_key_if_plaintext only catches (ValueError, TypeError) when calling serialization.load_pem_private_key(key_bytes, password=None). If a private key uses an unsupported algorithm or curve, cryptography raises cryptography.exceptions.UnsupportedAlgorithm (which inherits directly from Exception). This error goes uncaught during setup and crashes the transport initialization.

packages/google-auth/google/auth/aio/transport/mtls.py
```diff
@@ -71,19 +46,24 @@ def make_client_cert_ssl_context(
...
+            key_path,
+            passphrase_val,
+        ):
             context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
-            context.load_cert_chain(
-                certfile=cert_path, keyfile=key_path, password=passphrase
-            )
+            if cert_path:
+                context.load_cert_chain(
+                    certfile=cert_path,
```
lsirac, line 64: make_client_cert_ssl_context passes password=passphrase_val or "" to load_cert_chain, which changes None to "". Synchronous implementations (requests.py:229 and urllib3.py:194) pass password=passphrase_val (None when unencrypted). Passing an empty string "" forces OpenSSL to run decryption checks on keys that are not encrypted.

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -174,22 +174,29 @@ def _make_mutual_tls_http(cert, key):
         urllib3.PoolManager: Mutual TLS HTTP connection.
 
     Raises:
-        ImportError: If certifi or pyOpenSSL is not installed.
-        OpenSSL.crypto.Error: If the cert or key is invalid.
+        ValueError: If the cert or key is invalid.
```
lsirac: The docstrings list inaccurate exceptions after removing pyOpenSSL. In urllib3.py:177, _make_mutual_tls_http lists ValueError in its docstring, but actually raises google.auth.exceptions.MutualTLSChannelError. In requests.py:207, _MutualTlsAdapter lists ImportError but leaves out MutualTLSChannelError.

packages/google-auth/tests/compute_engine/test__mtls.py
```diff
@@ -250,6 +250,31 @@ def test_mds_mtls_adapter_send_fallback_http_error(
     assert fallback_request.url == "http://fake-mds.com/"
 
 
+@mock.patch("google.auth.compute_engine._mtls.HTTPAdapter")
+@mock.patch("google.auth.compute_engine._mtls._parse_mds_mode")
+@mock.patch("ssl.create_default_context")
+def test_mds_mtls_adapter_send_fallback_connection_error(
```
lsirac, line 255: the outer @mock.patch argument mock_http_adapter_send is completely overridden by an inner with mock.patch(...) block. All HTTP requests inside the test call the inner mock instead, causing mock_http_adapter_send.assert_not_called() to pass automatically without actually testing that fallback behavior was prevented.

lsirac (changes requested):

packages/google-auth/tests/transport/test__custom_tls_signer.py
nbayati: Good catch! I swapped the mock for a real ssl.SSLContext() here. I also added a quick runtime check to _cast_ssl_ctx_to_void_p_stdlib to explicitly throw a TypeError if a mock ever gets passed in again, just to be safe.

packages/google-auth/google/auth/transport/_mtls_helper.py
nbayati: Done. I added cryptography.exceptions.UnsupportedAlgorithm to the catch block so that unsupported keys won't crash the initialization anymore

packages/google-auth/google/auth/aio/transport/mtls.py
nbayati: Fixed. I removed the or "" fallback so it correctly passes None when the key is unencrypted, matching the sync implementations

packages/google-auth/google/auth/transport/urllib3.py
nbayati: Done!

packages/google-auth/tests/compute_engine/test__mtls.py
nbayati: You are right, the inner mock hides the fallback behavior. I removed the class mock entirely and changed the test to use side_effect=[ConnectionError, mock_response] on send().

lsirac (approved):

daniel-sanche (approved):
LGTM

illia-v:
urllib3 maintainers are glad to see google-auth no longer depending on our pyOpenSSL integration, thanks for making this change!

We're now considering deprecating `urllib3.contrib.pyopenssl` in urllib3 v2.x and removing it in v3.0: https://github.com/urllib3/urllib3/issues/5208.

With google-auth no longer relying on urllib3's pyOpenSSL support, one of the main reasons we previously kept this integration is gone. Feedback from you would be very welcome on the tracking issue.

PR #16977, opened 2026-05-07
chore: add a presubmit that checks for librarian config consistency

jskeet:
Currently we're using librarian for generation and legacylibrarian for releasing. This check makes sure that the relevant aspects of the two configurations (which libraries are configured, their versions, and whether releases are blocked) are consistent.

This commit also ensures that we're starting from a consistent state.

### DIFF
--- a/.github/workflows/librarian_config_check.yml
+++ b/.github/workflows/librarian_config_check.yml
@@ -1,0 +1,43 @@
+name: check that Librarian and legacylibrarian configs are consistent
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+permissions:
+  contents: read
+jobs:
+  config-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      # Use this action, rather than a file filter so that we can make this
+      # mandatory.
+      # See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-branches
+      # for more details.
+      - uses: dorny/paths-filter@v4
+        id: filter
+        with:
+          filters: |
+            librarian:
+              - 'librarian.yaml'
+              - '.librarian/config.yaml'
+              - '.librarian/state.yaml'
+      - name: Config check
+        id: config-check
+        if: steps.filter.outputs.librarian == 'true'
+        run: |
+          V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
+          go run "github.com/googleapis/librarian/tool/cmd/configcheck@${V}" .
+      - name: Report any failures
+        if: failure() && steps.config-check.outcome == 'failure'
+        run: |
+          echo "Library configuration is different between state.yaml and librarian.yaml.
+          Update library configuration in the configs according to the error message and
+          regenerate libraries using:
+          
+          V=\$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
+          go run github.com/googleapis/librarian/cmd/librarian@\${V} generate --all
+          "
+          # Make sure this step fails too, so that it's highlighted in the action logs.
+          exit 1
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -55,6 +55,7 @@ default:
 libraries:
   - name: bigframes
     version: 2.39.0
+    skip_release: true
     python:
       library_type: INTEGRATION
   - name: bigquery-magics
@@ -1933,6 +1934,7 @@ libraries:
     version: 3.10.1
     apis:
       - path: google/storage/v2
+    skip_release: true
     python:
       library_type: GAPIC_MANUAL
       opt_args_by_api:
@@ -2194,6 +2196,7 @@ libraries:
       default_version: v1
   - name: google-crc32c
     version: 1.8.0
+    skip_release: false
     python:
       library_type: OTHER
   - name: google-geo-type


librarian.yaml
```diff
@@ -1933,6 +1934,7 @@ libraries:
     version: 3.10.1
     apis:
       - path: google/storage/v2
+    skip_release: true
```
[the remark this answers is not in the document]
jskeet: This has been through librarian tidy already.

.github/workflows/librarian_config_check.yml
```diff
@@ -0,0 +1,40 @@
...
+          filters: |
+            librarian:
+              - 'librarian.yaml'
+              - '.librarian/config.yaml'
+              - '.librarian/state.yaml'
+      - name: Config check
+        if: steps.filter.outputs.librarian == 'true'
+        run: |
+          V="$(sed -n 's/^version: *//p' librarian.yaml)"
+          go run "github.com/googleapis/librarian/tool/cmd/configcheck@${V}" .
```
parthea, line 31: Gemini suggested we should validate the contents of $V follows a specific pattern in case `librarian.yaml` contains some malicious content.

```
Fix: Validate that the extracted $V matches an expected semver pattern before running the command, or restrict PRs from untrusted forks.
```

Gemini suggested

```
         # 1. Extract version
          V="$(sed -n 's/^version: *//p' librarian.yaml | tr -d '[:space:]')"
          
          # 2. Strict validation: Ensure version is purely alphanumeric/dots/dashes (no execution vectors)
          if [[ ! "$V" =~ ^v?[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
            echo "::error::Invalid version format detected in librarian.yaml: $V"
            exit 1
          fi
          
          # 3. Run safely
          go run "github.com/googleapis/librarian/tool/cmd/configcheck@${V}" .
```

.github/workflows/librarian_config_check.yml
```diff
@@ -0,0 +1,40 @@
+name: check that Librarian and legacylibrarian configs are consistent
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+permissions:
+  contents: read
+jobs:
+  config-check:
```
parthea, line 10: Gemini suggested that we need `actions/setup-go` before this step

```
Fix: Add actions/setup-go prior to the config check step.
```

.github/workflows/librarian_config_check.yml
```diff
@@ -0,0 +1,40 @@
...
+              - 'librarian.yaml'
+              - '.librarian/config.yaml'
+              - '.librarian/state.yaml'
+      - name: Config check
+        if: steps.filter.outputs.librarian == 'true'
+        run: |
+          V="$(sed -n 's/^version: *//p' librarian.yaml)"
+          go run "github.com/googleapis/librarian/tool/cmd/configcheck@${V}" .
+      - name: Update library versions and regenerate
+        if: failure()
```
parthea: Gemini suggested that we tweak this to `steps.config-check.outcome == 'failure'`

> if dorny/paths-filter fails or checkout fails, the ` Update library versions and regenerate` step will still trigger, which is confusing.

.github/workflows/librarian_config_check.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          echo "Library configuration is different between state.yaml and librarian.yaml.
+          Update library configuration in the configs according to the error message and
+          regenerate libraries using:
+          
+          V=\$(sed -n 's/^version: *//p' librarian.yaml)
+          go run github.com/googleapis/librarian/cmd/librarian@\${V} generate --all
```
parthea, line 40: Gemini suggested this instead. 

```
V=$(sed -n 's/^version: *//p' librarian.yaml)
go run github.com/googleapis/librarian/cmd/librarian@${V} generate --all
```

```
You are escaping the variable as \${V} inside a multiline YAML string. Because it is inside a run: block, the shell executes the echo but might print literal backslashes to the log depending on how the runner's shell interprets the escapes. Using standard single quotes for the echo string prevents premature shell expansion without needing messy backslashes.
```

.github/workflows/librarian_config_check.yml
jskeet: We control librarian.yaml. I don't think there's any need to do that. (In a future change we might want to use librarian to print the version from the cfg, but I think this is okay for now.)

jskeet:
I won't have time to address these comments today - and I'm on vacation tomorrow. Will come back to this on Monday.

jskeet:
(I'd note that whatever change we *do* make here, we probably want to do the same in google-cloud-go as well, which this was basically copied from...)

jskeet:
Actually, it looks like the go version is already a bit different, including using `go run github.com/googleapis/librarian/cmd/librarian@latest config get version` rather than `sed`. Will update in that respect now...

.github/workflows/librarian_config_check.yml
parthea: Acknowledged

.github/workflows/librarian_config_check.yml
jskeet: Done this one.

.github/workflows/librarian_config_check.yml
jskeet: We definitely don't - the action is working.

.github/workflows/librarian_config_check.yml
jskeet: I've tested this, and the current commit works - see https://github.com/jskeet/google-cloud-python/actions/runs/25506515696 for an example of failure

jskeet:
@parthea I've rebased this to a single commit after doing a bunch of testing. It's ready for review now, see https://github.com/jskeet/google-cloud-python/actions/runs/25506515696 as an example of a failure.

parthea (approved):

parthea:
@jskeet Could we mark the presubmit as a required check?

Issue #16487, opened 2026-03-31
[post-migration] google-cloud-storage post-migration tasks


parthea:
`googleapis/python-storage` was moved to https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-storage in PR https://github.com/googleapis/google-cloud-python/pull/16192

- [ ] Ensure that the following changes make it into the release notes for the next release due to changes made to `setup.py` to align with the state of `google-cloud-python`:

  - Python 3.8 was dropped
  - The package now requires `protobuf >= 4.25.8, < 8.0.0`

- [x] Ensure that the presubmit called `python-sdk-zb-system-test (gcs-aiml-clients-testing-101)` is present and runs in google cloudbuild for google-cloud-python. (**DONE**, [see PR #16581](https://github.com/googleapis/google-cloud-python/pull/16581))
- [ ] Ensure that the analogous presubmit for the zonal bucket system test is present and runs in google cloudbuild for python-docs-samples [see PR #14022](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/14022)
- [x] Ensure that the [snippets from the split repo](https://github.com/googleapis/python-storage/tree/main/samples/snippets) are migrated to the [python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples) repo (**DONE**, [see PR #13989]( https://github.com/GoogleCloudPlatform/python-docs-samples/pull/13989)).
- [x] Ensure that system tests as found in `Kokoro System Tests` run successfully (**DONE**, [see PR #16578](https://github.com/googleapis/google-cloud-python/pull/16578)).

chandra-siri:
Hi @parthea , @chalmerlowe 

There's one more task, details in b/500990583

chalmerlowe:
> Hi [@parthea](https://github.com/parthea) , [@chalmerlowe](https://github.com/chalmerlowe)
> 
> There's one more task, details in b/500990583

I created a [PR in python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/14022) with a `cloudbuild` yaml config and a zonal bucket script.
I mirrored it off the files in `google-cloud-python`. I went in and changed paths where possible to align with the directory structure in `python-docs-samples`.

* I do not know what other configs or permissions (i.e. in projects, storage buckets, etc) need to be updated to enable these tests, nor whether the config OR script need additional updates.

@chandra-siri @chingor13 Please advise.

[...]

PR #16986, opened 2026-05-07
chore(storage): uncomment zb-system-tests-cloudbuild.yaml

chandra-siri:
This was commented out because of failing cloud build. It's now working.


parthea:
See https://github.com/googleapis/google-cloud-python/issues/16487#issuecomment-4315466672 for the reason this was originally commented out.  LGTM, assuming that the build passes.

parthea (approved):

PR #16991, opened 2026-05-07
chore: librarian release pull request: 20260507T175102Z

noahdietz:
PR created by the Librarian CLI to initialize a release. Merging this PR will auto trigger a release.

Librarian Version: v0.12.0
Language Image: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/python-librarian-generator@sha256:234b9d1f2ddb057ed7ac6a38db0bf8163d839c65c6cf88ade52530cddebce59e
<details><summary>google-cloud-bigtable: v2.38.0</summary>

### Features

* Add ValueBitmask to the filter DSL (#16981) ([3f429f23](https://github.com/googleapis/google-cloud-python/commit/3f429f23))

</details>

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -608,7 +608,7 @@ libraries:
       metadata_name_override: bigquerystorage
       default_version: v1
   - name: google-cloud-bigtable
-    version: 2.37.0
+    version: 2.38.0
     apis:
       - path: google/bigtable/v2
       - path: google/bigtable/admin/v2
@@ -2196,7 +2196,6 @@ libraries:
       default_version: v1
   - name: google-crc32c
     version: 1.8.0
-    skip_release: false
     python:
       library_type: OTHER
   - name: google-geo-type
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -1361,7 +1361,7 @@ libraries:
       - packages/google-cloud-bigquery-storage/docs/
     tag_format: '{id}-v{version}'
   - id: google-cloud-bigtable
-    version: 2.37.0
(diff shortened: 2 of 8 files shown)


librarian.yaml
```diff
@@ -2196,7 +2196,6 @@ libraries:
       default_version: v1
   - name: google-crc32c
     version: 1.8.0
-    skip_release: false
```
noahdietz: This looks like a `tidy` fix, which is fine - `false` is the default value. Odd this didn't get tidied before...

jskeet: This was me, adding a config check - but I'm not sure offhand whether this should actually be true rather than false. (It may have been left over from testing.) I'm afraid I don't have the opportunity to look again before Monday though :( Worth checking against config.yaml. (And if it *is* meant to be `true`, presumably the check should be failing!)

jskeet: Ah! I know why the check failed... the currently-used version of librarian doesn't actually check for skip_release. So I'm pretty sure this should be true, and it'll start complaining if we don't fix it - but only when we upgrade the version of librarian.

noahdietz:
For posterity, this is for b/510798129

daniel-sanche (approved):

librarian.yaml
noahdietz: I'm going to fix this in a separate PR and update this one

noahdietz: https://github.com/googleapis/google-cloud-python/pull/16992

noahdietz: Fixed

PR #16999, opened 2026-05-07
chore(firestore): improve typing for collection

daniel-sanche:
Previously, `CollectionReference` didn't define `.document()`, so it inherited the typing of `BaseCollectionReference`. This PR adds an override to ensure that documents are annotated as returning `DocumentReference` types


chalmerlowe (approved):
LGTM

PR #17001, opened 2026-05-07
chore(.gemini): add config

noahdietz:
Proposal to further configure gemini assistance on the repo in the following ways:
* disable `have_fun` to keep feedback concise
* disable posting the help message, it's not particularly useful when we have public docs
* run code review and include draft PRs
* (the big one) attempt to limit the scope of review, excluding what I think would be GAPIC generate client libraries 

Context: Gemini posted comments to auto-generated code in #16998, which we will not address, and I'd like to reduce noise / save some compute.


parthea (approved):
LGTM as an initial configuration. There are many more files that we should add to `ignore_patterns` but we can continue to build on this in subsequent PRs.

PR #17007, opened 2026-05-08
fix(geminidataanalytics)!: generate and default to v1

noahdietz:
Adds  `v1` to existing `google-cloud-geminidataanalytics` client and switches `default_version` to `v1`.

Towards b/510217868


daniel-sanche:
I added a ! to the title, since I believe this should be a breaking change. I'm going to come back and look at this soon to be sure though

daniel-sanche:
Removed !, because this library is in previous, so we can update the default without a major update

daniel-sanche (approved):
LGTM

PR #17010, opened 2026-05-09
chore(bigframes): clean up the the AI package

sycai:
The contents of the change are:
* Fixed Sphinx format error: single backticks are replaced with double backticks.
* Removed parameter verification to keep the client lean.
* Removed pre-GA notes as these functions are all GA'ed
* Fixed format of some other files .

### DIFF
--- a/packages/bigframes/bigframes/bigquery/_operations/ai.py
+++ b/packages/bigframes/bigframes/bigquery/_operations/ai.py
@@ -57,9 +57,9 @@ def generate(
 
     **Examples:**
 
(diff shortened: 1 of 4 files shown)


sycai:
The doctests are unskipped because they are ignored here:

https://github.com/googleapis/google-cloud-python/blob/19331c96e486286ae0d23976559ff62bd1458ed3/packages/bigframes/noxfile.py#L418-L419

shuoweil (approved):

PR #17014, opened 2026-05-11
fix(spanner_dbapi): replace insecure pickle with json for partition deserialization

sinhasubham:
This PR resolves a critical Insecure Deserialization vulnerability (potential Remote Code Execution) in the `spanner_dbapi` module [b/510871112](b/510871112) . Previously, the module utilized `pickle.loads()` to decode partition IDs provided by users via the `RUN PARTITION` statement, creating a possibility for arbitrary code execution attack payloads.
We have fully eliminated `pickle` usage in this module and migrated to standard `json` serialization.

### DIFF
--- a/packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
+++ b/packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
@@ -13,24 +13,26 @@
 # limitations under the License.
 
 import base64
+import copy
 import datetime
-import decimal
 import gzip
 import json
-import uuid
 from dataclasses import dataclass
-from google.cloud.spanner_v1.data_types import Interval, JsonObject
 from typing import Any
 
 from google.protobuf.json_format import MessageToDict, ParseDict
 from google.protobuf.message import Message
+from google.protobuf.struct_pb2 import Struct
 
 from google.cloud.spanner_v1 import BatchTransactionId
-from google.cloud.spanner_v1.types import ExecuteSqlRequest, DirectedReadOptions
+from google.cloud.spanner_v1._helpers import _make_value_pb
+from google.cloud.spanner_v1.types import DirectedReadOptions, ExecuteSqlRequest, Type
 
 _PROTO_CLASS_MAP = {
     "QueryOptions": ExecuteSqlRequest.QueryOptions,
     "DirectedReadOptions": DirectedReadOptions,
+    "Struct": Struct,
+    "Type": Type,
 }
 
 
@@ -39,14 +41,6 @@ def _serialize_value(val: Any) -> Any:
         return {"__type__": "bytes", "value": base64.b64encode(val).decode("utf-8")}
     elif isinstance(val, datetime.datetime):
         return {"__type__": "datetime", "value": val.isoformat()}
-    elif isinstance(val, datetime.date):
-        return {"__type__": "date", "value": val.isoformat()}
-    elif isinstance(val, decimal.Decimal):
-        return {"__type__": "decimal", "value": str(val)}
-    elif isinstance(val, uuid.UUID):
-        return {"__type__": "uuid", "value": str(val)}
-    elif isinstance(val, Interval):
-        return {"__type__": "interval", "value": str(val)}
     elif hasattr(val, "_pb"):
         return {
             "__type__": "protobuf",
@@ -59,8 +53,6 @@ def _serialize_value(val: Any) -> Any:
             "class": val.__class__.__name__,
             "value": MessageToDict(val, preserving_proto_field_name=True),
         }
-    elif isinstance(val, JsonObject):
-        return {"__type__": "json_object", "value": val.serialize()}
     elif isinstance(val, dict):
         return {k: _serialize_value(v) for k, v in val.items()}
     elif isinstance(val, list):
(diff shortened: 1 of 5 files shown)


packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -13,23 +13,91 @@
...
+                return base64.b64decode(val["value"])
+            elif t == "datetime":
+                dt_str = val["value"]
+                if dt_str.endswith("Z"):
+                    dt_str = dt_str[:-1] + "+00:00"
+                return datetime.datetime.fromisoformat(dt_str)
+            elif t == "tuple":
+                return tuple(_deserialize_value(x) for x in val["value"])
+            elif t == "protobuf":
+                return _deserialize_value(val["value"])
```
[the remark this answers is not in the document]
sinhasubham: Addressed

packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -13,23 +13,91 @@
...
+        return {
+            "__type__": "protobuf",
+            "class": val.__class__.__name__,
+            "value": MessageToDict(val._pb),
+        }
+    elif isinstance(val, Message):
+        return {
+            "__type__": "protobuf",
+            "class": val.__class__.__name__,
+            "value": MessageToDict(val),
```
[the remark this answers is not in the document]
sinhasubham: Addressed

packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -13,23 +13,91 @@
...
+def _serialize_value(val: Any) -> Any:
+    if isinstance(val, bytes):
+        return {"__type__": "bytes", "value": base64.b64encode(val).decode("utf-8")}
+    elif isinstance(val, datetime.datetime):
+        return {"__type__": "datetime", "value": val.isoformat()}
+    elif hasattr(val, "_pb"):
+        return {
+            "__type__": "protobuf",
+            "class": val.__class__.__name__,
+            "value": MessageToDict(val._pb),
```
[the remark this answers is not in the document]
sinhasubham: Addressed

packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -13,23 +13,127 @@
...
 from google.cloud.spanner_v1 import BatchTransactionId
+from google.cloud.spanner_v1.types import ExecuteSqlRequest, DirectedReadOptions
+
+_PROTO_CLASS_MAP = {
+    "QueryOptions": ExecuteSqlRequest.QueryOptions,
+    "DirectedReadOptions": DirectedReadOptions,
+}
+
+
+def _serialize_value(val: Any) -> Any:
```
olavloite, line 39: I think that the chosen solution approach has a design flaw, and that is that it tries to manually serialize a high-level `Statement` object into JSON. That means that it needs to take into account every single query parameter type that a `Statement` can contain, and that list is long and, most importantly, __not fixed__. Someone could add support for a new high-level type tomorrow, and that would then break this serialization, unless they also add it to this serializer.

We can 'fix' all of this by adding all supported types here, and also add a test that tries to catch it if someone adds a new type in the future that is not in this list. I added https://github.com/googleapis/google-cloud-python/pull/17014/commits/5ef63941799fd10bef7947a5d54c79897e026e99 to illustrate all of this. But as you can see: The code becomes quite bloated, and the test uses meta-programming to verify that this serialization is consistent with the `_make_value_pb` function (which is a code smell).

I would suggest that we try a different strategy for this, and instead use a two-step serialization strategy:

1. First convert the `Statement` into an `ExecuteSqlRequest` using the standard functions in the client library (or at least convert the query parameters of the `Statement` into the `Struct` that would have been used for the `ExecuteSqlRequest`).
2. Then serialize the `ExecuteSqlRequest` (or the `Struct` with the query parameters).

This removes the need to manually map all possible high-level query parameter types, as `ExecuteSqlRequest` is guaranteed to only use well-known types for query parameters, and this list is __fixed__.

packages/google-cloud-spanner/google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -13,23 +13,145 @@
...
+                    ParseDict(dict_val, msg)
+                    return cls(msg) if hasattr(cls(), "_pb") else msg
+                return _deserialize_value(dict_val)
+        return {k: _deserialize_value(v) for k, v in val.items()}
+    elif isinstance(val, list):
+        return [_deserialize_value(v) for v in val]
+    return val
+
+
+def _unpack_value_pb(value):
```
olavloite, line 93: I think that you can remove this method, and replace the calls below to this method with this:

```python
partition_result["query"]["params"] = MessageToDict(params_pb)
```

packages/google-cloud-spanner/tests/mockserver_tests/test_dbapi_partition_query.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2024 Google LLC All rights reserved.
```
olavloite, line 1: nit: 
```suggestion
# Copyright 2026 Google LLC All rights reserved.
```

packages/google-cloud-spanner/tests/unit/spanner_dbapi/test_partition_helper.py
```diff
@@ -0,0 +1,338 @@
+# Copyright 2024 Google LLC All rights reserved.
```
olavloite, line 1: 2026

olavloite (approved):

sinhasubham:
Thanks, have fixed these comments locally. will merge along with the next change.

PR #17017, opened 2026-05-11
chore: remove Python 3.9 from default nox session

parthea:
This PR removes Python 3.9 from the default nox session which is used in continuous builds. Python 3.9 unit tests are still running in presubmits via Github actions but cannot be tested in continuous builds since Python 3.9 has been removed from the testing image as of https://github.com/googleapis/testing-infra-docker/pull/524.


chalmerlowe (approved):
LGTM

PR #17025, opened 2026-05-11
feat(auth): implement regional access boundary support for standalone JWT and async service accounts

nbayati:
This PR implements the following changes:

- Add RAB support to async service account credential type, by providing async manager and fetching methods. 
- Update unit tests to accept both mtls and standard lookup endpoint urls. 
- Refactor before_request to use a _after_refresh hook so we don't have to override the method.
- Add RAb support for self signed jwt flow through jwt.py
- some small enhancements for test coverage and backward compatibility

### DIFF
--- a/packages/google-auth/google/auth/_jwt_async.py
+++ b/packages/google-auth/google/auth/_jwt_async.py
@@ -91,7 +91,9 @@ def decode(token, certs=None, verify=True, audience=None):
 
 
 class Credentials(
-    jwt.Credentials, _credentials_async.Signing, _credentials_async.Credentials
+    jwt.Credentials,
+    _credentials_async.Signing,
+    _credentials_async.CredentialsWithRegionalAccessBoundary,
 ):
     """Credentials that use a JWT as the bearer token.
 
@@ -141,6 +143,15 @@ class Credentials(
             'https://pubsub.googleapis.com/google.pubsub.v1.Subscriber')
         new_credentials = credentials.with_claims(audience=new_audience)
     """
+
+    def __setstate__(self, state):
+        """Restores the credential state and ensures the async refresh manager is attached."""
+        super().__setstate__(state)
+        from google.auth import _regional_access_boundary_utils
+
+        self._rab_manager.refresh_manager = (
+            _regional_access_boundary_utils._AsyncRegionalAccessBoundaryRefreshManager()
+        )
 
 
 class OnDemandCredentials(
--- a/packages/google-auth/google/auth/_jwt_async.py
+++ b/packages/google-auth/google/auth/_jwt_async.py
@@ -46,6 +46,7 @@ change in minor releases.
 from google.auth import _credentials_async
 from google.auth import _helpers
 from google.auth import jwt
+from google.auth import _regional_access_boundary_utils
 
 
 def encode(signer, payload, header=None, key_id=None):
@@ -148,7 +149,6 @@ class Credentials(
     def __setstate__(self, state):
         """Restores the credential state and ensures the async refresh manager is attached."""
         super().__setstate__(state)
-        from google.auth import _regional_access_boundary_utils
 
         self._rab_manager.refresh_manager = (
             _regional_access_boundary_utils._AsyncRegionalAccessBoundaryRefreshManager()
--- a/packages/google-auth/google/auth/_jwt_async.py
+++ b/packages/google-auth/google/auth/_jwt_async.py
@@ -44,6 +44,7 @@ change in minor releases.
 """
 
 from google.auth import _credentials_async
+from google.auth import _helpers
 from google.auth import jwt
 
 
@@ -173,3 +174,7 @@ class OnDemandCredentials(
 
     .. _grpc: http://www.grpc.io/
     """
+
+    @_helpers.copy_docstring(jwt.OnDemandCredentials)
+    async def before_request(self, request, method, url, headers):
+        super(OnDemandCredentials, self).before_request(request, method, url, headers)
--- a/packages/google-auth/google/auth/_jwt_async.py
+++ b/packages/google-auth/google/auth/_jwt_async.py
@@ -45,8 +45,8 @@ change in minor releases.
 
 from google.auth import _credentials_async
 from google.auth import _helpers
-from google.auth import jwt
 from google.auth import _regional_access_boundary_utils
+from google.auth import jwt
 
 
 def encode(signer, payload, header=None, key_id=None):
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -171,12 +171,11 @@ class _RegionalAccessBoundaryManager(object):
         else:
             headers.pop(_REGIONAL_ACCESS_BOUNDARY_HEADER, None)
 
-    def maybe_start_refresh(self, credentials, request):
-        """Starts a background thread to refresh the Regional Access Boundary if needed.
+    def _should_refresh(self):
+        """Checks if the Regional Access Boundary data needs a refresh and is not in cooldown.
 
-        Args:
-            credentials (google.auth.credentials.Credentials): The credentials to refresh.
-            request (google.auth.transport.Request): The object used to make HTTP requests.
+        Returns:
+            bool: True if a refresh is required, False otherwise.
         """
         rab_data = self._data
 
@@ -187,15 +186,43 @@ class _RegionalAccessBoundaryManager(object):
             and _helpers.utcnow()
             < (rab_data.expiry - REGIONAL_ACCESS_BOUNDARY_REFRESH_THRESHOLD)
         ):
-            return
+            return False
 
         # Don't start a new refresh if the cooldown is still in effect.
         if rab_data.cooldown_expiry and _helpers.utcnow() < rab_data.cooldown_expiry:
+            return False
+
+        return True
+
+    def maybe_start_refresh(self, credentials, request):
+        """Starts a background thread to refresh the Regional Access Boundary if needed.
+
+        Args:
+            credentials (google.auth.credentials.Credentials): The credentials to refresh.
+            request (google.auth.transport.Request): The object used to make HTTP requests.
+        """
+        if not self._should_refresh():
             return
 
         # If all checks pass, start the background refresh.
         if self._use_blocking_regional_access_boundary_lookup:
             self.start_blocking_refresh(credentials, request)
+        else:
+            self.refresh_manager.start_refresh(credentials, request, self)
+
+    async def maybe_start_refresh_async(self, credentials, request):
+        """Starts a background refresh or performs a blocking refresh asynchronously.
+
+        Args:
+            credentials (google.auth.credentials.Credentials): The credentials to refresh.
+            request (google.auth.aio.transport.Request): The object used to make HTTP requests.
+        """
+        if not self._should_refresh():
+            return
+
+        # If all checks pass, start the refresh.
+        if self._use_blocking_regional_access_boundary_lookup:
+            await self.start_blocking_refresh_async(credentials, request)
         else:
             self.refresh_manager.start_refresh(credentials, request, self)
 
@@ -221,6 +248,37 @@ class _RegionalAccessBoundaryManager(object):
             if _helpers.is_logging_enabled(_LOGGER):
                 _LOGGER.warning(
                     "Blocking Regional Access Boundary lookup raised an exception: %s",
+                    e,
+                    exc_info=True,
+                )
+            regional_access_boundary_info = None
+
+        self.process_regional_access_boundary_info(regional_access_boundary_info)
+
+    async def start_blocking_refresh_async(self, credentials, request):
+        """Initiates a blocking lookup of the Regional Access Boundary asynchronously.
+
+        If the lookup raises an exception, it is caught and logged as a warning,
+        and the lookup is treated as a failure (entering cooldown). Exceptions
+        are not propagated to the caller.
+
+        Args:
+            credentials (google.auth.credentials.Credentials): The credentials to refresh.
+            request (google.auth.aio.transport.Request): The object used to make HTTP requests.
+        """
+        try:
+            # The fail_fast parameter is set to True to ensure we don't block the calling
+            # thread for too long. This will do two things: 1) set a timeout to 3s
+            # instead of the default 120s and 2) ensure we do not retry at all
+            regional_access_boundary_info = (
+                await credentials._lookup_regional_access_boundary(
+                    request, fail_fast=True
+                )
+            )
+        except Exception as e:
+            if _helpers.is_logging_enabled(_LOGGER):
+                _LOGGER.warning(
+                    "Regional Access Boundary lookup raised an exception: %s",
                     e,
                     exc_info=True,
                 )
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -395,14 +395,14 @@ class _AsyncRegionalAccessBoundaryRefreshManager(object):
         self._worker_task = None
 
     def __getstate__(self):
-        """Pickle helper that serializes the _lock and _worker_task attributes."""
+        """Pickle helper that excludes the un-picklable _lock and _worker_task attributes from serialization."""
         state = self.__dict__.copy()
         state["_lock"] = None
         state["_worker_task"] = None
         return state
 
     def __setstate__(self, state):
-        """Pickle helper that deserializes the _lock and _worker_task attributes."""
+        """Pickle helper that restores state and re-initializes the _lock and _worker_task attributes."""
         self.__dict__.update(state)
         self._lock = threading.Lock()
         self._worker_task = None
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -16,9 +16,9 @@
 
 import asyncio
 import copy
-import inspect
 import datetime
 import functools
+import inspect
 import logging
 import os
 import threading
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import copy
+import inspect
 import datetime
 import functools
 import logging
@@ -237,6 +238,15 @@ class _RegionalAccessBoundaryManager(object):
             credentials (google.auth.credentials.Credentials): The credentials to refresh.
             request (google.auth.transport.Request): The object used to make HTTP requests.
         """
+        # Async credentials do not support blocking lookups.
+        if inspect.iscoroutinefunction(credentials._lookup_regional_access_boundary):
+            if _helpers.is_logging_enabled(_LOGGER):
+                _LOGGER.warning(
+                    "Blocking Regional Access Boundary lookup is not supported for async credentials."
+                )
+            self.process_regional_access_boundary_info(None)
+            return
+
         try:
             # The fail_fast parameter is set to True to ensure we don't block the calling
             # thread for too long. This will do two things: 1) set a timeout to 3s
@@ -500,7 +510,12 @@ class _AsyncRegionalAccessBoundaryRefreshManager(object):
                     regional_access_boundary_info
                 )
 
-            self._worker_task = asyncio.create_task(_worker())
+            coro = _worker()
+            try:
+                self._worker_task = asyncio.create_task(coro)
+            except Exception:
+                coro.close()
+                raise
 
 
 def _get_domain() -> str:
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -14,8 +14,8 @@
 
 """Utilities for Regional Access Boundary management."""
 
-import copy
 import asyncio
+import copy
 import datetime
 import functools
 import logging
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -385,6 +385,8 @@ class _RegionalAccessBoundaryRefreshManager(object):
                 credentials, copied_request, rab_manager
             )
             self._worker.start()
+
+
 class _AsyncRegionalAccessBoundaryRefreshManager(object):
     """Manages a task for background refreshing of the Regional Access Boundary in async flows."""
 
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -15,6 +15,7 @@
 """Utilities for Regional Access Boundary management."""
 
 import copy
+import asyncio
 import datetime
 import functools
 import logging
@@ -384,3 +385,59 @@ class _RegionalAccessBoundaryRefreshManager(object):
                 credentials, copied_request, rab_manager
             )
             self._worker.start()
+class _AsyncRegionalAccessBoundaryRefreshManager(object):
+    """Manages a task for background refreshing of the Regional Access Boundary in async flows."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._worker_task = None
+
+    def __getstate__(self):
+        """Pickle helper that serializes the _lock and _worker_task attributes."""
+        state = self.__dict__.copy()
+        state["_lock"] = None
+        state["_worker_task"] = None
+        return state
+
+    def __setstate__(self, state):
+        """Pickle helper that deserializes the _lock and _worker_task attributes."""
+        self.__dict__.update(state)
+        self._lock = threading.Lock()
+        self._worker_task = None
+
+    def start_refresh(self, credentials, request, rab_manager):
+        """
+        Starts a background task to refresh the Regional Access Boundary if one is not already running.
+
+        Args:
+            credentials (CredentialsWithRegionalAccessBoundary): The credentials
+                to refresh.
+            request (google.auth.aio.transport.Request): The object used to make
+                HTTP requests.
+            rab_manager (_RegionalAccessBoundaryManager): The manager container to update.
+        """
+        with self._lock:
+            if self._worker_task and not self._worker_task.done():
+                # A refresh is already in progress.
+                return
+
+            async def _worker():
+                try:
+                    # credentials._lookup_regional_access_boundary should be async in the async creds class
+                    regional_access_boundary_info = (
+                        await credentials._lookup_regional_access_boundary(request)
+                    )
+                except Exception as e:
+                    if _helpers.is_logging_enabled(_LOGGER):
+                        _LOGGER.warning(
+                            "Asynchronous Regional Access Boundary lookup raised an exception: %s",
+                            e,
+                            exc_info=True,
+                        )
+                    regional_access_boundary_info = None
+
+                rab_manager.process_regional_access_boundary_info(
+                    regional_access_boundary_info
+                )
+
+            self._worker_task = asyncio.create_task(_worker())
--- a/packages/google-auth/google/auth/_regional_access_boundary_utils.py
+++ b/packages/google-auth/google/auth/_regional_access_boundary_utils.py
@@ -501,3 +501,57 @@ class _AsyncRegionalAccessBoundaryRefreshManager(object):
                 )
 
             self._worker_task = asyncio.create_task(_worker())
+
+
+def _get_domain() -> str:
+    """Dynamically determines the domain for IAM credentials based on active mTLS configuration.
+
+    Returns:
+        str: The dynamic domain string.
+    """
+    from google.auth.transport import _mtls_helper
+
+    if (
+        hasattr(_mtls_helper, "check_use_client_cert")
+        and _mtls_helper.check_use_client_cert()
+    ):
+        return f"iamcredentials.mtls.{_helpers.DEFAULT_UNIVERSE_DOMAIN}"
+    else:
+        return f"iamcredentials.{_helpers.DEFAULT_UNIVERSE_DOMAIN}"
+
+
+def get_service_account_rab_endpoint(service_account_email: str) -> str:
+    """Builds the Regional Access Boundary lookup URL for service accounts.
+
+    Args:
+        service_account_email: The service account email.
+
+    Returns:
+        str: The complete lookup URL.
+    """
+    return f"https://{_get_domain()}/v1/projects/-/serviceAccounts/{service_account_email}/allowedLocations"
+
+
+def get_workforce_pool_rab_endpoint(pool_id: str) -> str:
+    """Builds the Regional Access Boundary lookup URL for workforce pools.
+
+    Args:
+        pool_id: The workforce pool ID.
+
+    Returns:
+        str: The complete lookup URL.
+    """
+    return f"https://{_get_domain()}/v1/locations/global/workforcePools/{pool_id}/allowedLocations"
+
+
+def get_workload_identity_pool_rab_endpoint(project_number: str, pool_id: str) -> str:
+    """Builds the Regional Access Boundary lookup URL for workload identity pools.
+
+    Args:
+        project_number: The Google Cloud project number.
+        pool_id: The workload identity pool ID.
+
+    Returns:
+        str: The complete lookup URL.
+    """
+    return f"https://{_get_domain()}/v1/projects/{project_number}/locations/global/workloadIdentityPools/{pool_id}/allowedLocations"
--- a/packages/google-auth/google/auth/compute_engine/_metadata.py
+++ b/packages/google-auth/google/auth/compute_engine/_metadata.py
@@ -502,3 +502,20 @@ def get_service_account_token(request, service_account="default", scopes=None):
         seconds=token_json["expires_in"]
     )
     return token_json["access_token"], token_expiry
+
+
+def _is_email(email):
+    """Checks if the provided string is in an email format.
+
+    This is a simple check that ensures the candidate string is non-empty
+    and contains the '@' character.
+
+    Args:
+        email (str): The candidate string to check.
+
+    Returns:
+        bool: True if the string is non-empty and contains '@', False otherwise.
+    """
+    if not email:
+        return False
+    return "@" in email
--- a/packages/google-auth/google/auth/compute_engine/_metadata.py
+++ b/packages/google-auth/google/auth/compute_engine/_metadata.py
@@ -38,9 +38,7 @@ from google.auth.compute_engine import _mtls
 
 _LOGGER = logging.getLogger(__name__)
 
-_SERVICE_ACCOUNT_EMAIL_PATTERN = re.compile(
-    r"^[^@]+@[^@]+\.[^@]+$"
-)
+_SERVICE_ACCOUNT_EMAIL_PATTERN = re.compile(r"^[^@]+@[^@]+\.[^@]+$")
 
 _GCE_DEFAULT_MDS_IP = "169.254.169.254"
 _GCE_DEFAULT_HOST = "metadata.google.internal"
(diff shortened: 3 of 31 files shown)


packages/google-auth/google/auth/jwt.py
```diff
@@ -559,7 +574,7 @@ def _make_jwt(self):
 
         return jwt, expiry
 
-    def refresh(self, request):
+    def _perform_refresh_token(self, request):
```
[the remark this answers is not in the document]
nbayati: This change was intentional. The base class CredentialsWithRegionalAccessBoundary inside google/auth/credentials.py already provides the refresh() method that calls _perform_refresh_token(). This class correctly adopts that expected contract by inheriting from it.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -384,3 +385,61 @@ def start_refresh(self, credentials, request, rab_manager):
...
+
+class _AsyncRegionalAccessBoundaryRefreshManager(object):
+    """Manages a task for background refreshing of the Regional Access Boundary in async flows."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._worker_task = None
+
+    def __getstate__(self):
+        """Pickle helper that serializes the _lock and _worker_task attributes."""
```
macastelaz: This comment seems misleading/incorrect to me. From what I can tell, this removes these attributes from the serialization of this manager entirely (it doesn't help in/allow the attributes to be serialized).

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -384,3 +385,61 @@ def start_refresh(self, credentials, request, rab_manager):
...
+        state = self.__dict__.copy()
+        state["_lock"] = None
+        state["_worker_task"] = None
+        return state
+
+    def __setstate__(self, state):
+        """Pickle helper that deserializes the _lock and _worker_task attributes."""
```
macastelaz: Similarly, I don't think this is actually deserializing anything for these attributes, but it is re-initiating them

macastelaz (approved):

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
...
+            if timeout:
+                response = await asyncio.wait_for(
+                    request(method="GET", url=url, headers=headers), timeout=timeout
+                )
+            else:
+                response = await request(method="GET", url=url, headers=headers)
+        except asyncio.TimeoutError:
+            return False, {}, False
+
+        response_body1 = await response.content()
```
lsirac: This is an async_generator, not an awaitable method. I think this crashes as written. Should call await response.read() instead. 

Let's also add a test that will catch this, and cover in try/catch block

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
...
+    total_attempts = 1 if fail_fast else 6
+    retries = _exponential_backoff.AsyncExponentialBackoff(
+        total_attempts=total_attempts
+    )
+
+    async for _ in retries:
+        try:
+            if timeout:
+                response = await asyncio.wait_for(
+                    request(method="GET", url=url, headers=headers), timeout=timeout
```
lsirac: Should timeout be passed to the request here?

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
...
+                )
+            else:
+                response = await request(method="GET", url=url, headers=headers)
+        except asyncio.TimeoutError:
```
lsirac: What happens if the request fails? I think we may need to catch additional errors here (timeout error, transport error..)

packages/google-auth/google/auth/credentials.py
```diff
@@ -369,6 +401,8 @@ def _copy_regional_access_boundary_manager(self, target):
         # but share the immutable data reference to avoid unnecessary initial lookups.
         new_manager = _regional_access_boundary_utils._RegionalAccessBoundaryManager()
         new_manager._data = self._rab_manager._data
+        # Preserve the type of refresh manager (sync or async)
+        new_manager.refresh_manager = self._rab_manager.refresh_manager.__class__()
```
lsirac: I don't think this is safe across sync/async creds.
1. Using the source refresh-manager means we can have an async refresh manager on a sync cred which can later call asyncio.create_task() from a sync before_request.
2.  _use_blocking_regional_access_boundary_lookup is not kept, so a credential configured for blocking RAB can become non-blocking after with_scopes(), with_quota_project(), etc (breaking gcloud)

I think the cred should keep its initialized manager type and we should only copy the RAB data/config.

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
...
+            }
+    """
+    response_data = await _lookup_regional_access_boundary_request(
+        request, url, headers=headers, fail_fast=fail_fast
+    )
+    if response_data is None:
+        # Error was already logged by _lookup_regional_access_boundary_request
+        return None
+
+    if "encodedLocations" not in response_data:
```
lsirac: Nit: should fail open on bad server responses, so we should check that response_data is a dict

packages/google-auth/tests_async/oauth2/test__client_async.py
```diff
@@ -492,3 +493,68 @@ async def test__token_endpoint_request_no_throw_with_retry(can_retry):
         assert mock_request.call_count == 3
     else:
         assert mock_request.call_count == 1
+
+
```
lsirac, line 512: Current mocks use status and awaitable content(), but real aio responses use status_code and read(). Also should cover invalid JSON / request failures / response issues / timeout behavior.

packages/google-auth/tests_async/oauth2/test_service_account_async.py
```diff
@@ -229,6 +229,73 @@ async def test_before_request_refreshes(self, jwt_grant):
         # Credentials should now be valid.
         assert credentials.valid
 
+    @mock.patch(
```
lsirac: Please add a test for async SA credentials with blocking RAB enabled (`with_blocking_regional_access_boundary_lookup()`).  

Currently, when blocking mode is active, `start_blocking_refresh` synchronously calls `credentials._lookup_regional_access_boundary(request, fail_fast=True)` without `await`ing it. This returns an unawaited coroutine object rather than a dictionary, causing `process_regional_access_boundary_info` to crash immediately with `AttributeError: 'coroutine' object has no attribute 'get'` alongside unawaited coroutine warnings.

packages/google-auth/tests/test_external_account.py
```diff
@@ -1729,13 +1729,21 @@ def test_before_request_expired(self, utcnow):
 
     def test_build_regional_access_boundary_lookup_url_workload(self):
         credentials = self.make_credentials()
-        expected_url = "https://iamcredentials.googleapis.com/v1/projects/123456/locations/global/workloadIdentityPools/POOL_ID/allowedLocations"
-        assert credentials._build_regional_access_boundary_lookup_url() == expected_url
+        url = credentials._build_regional_access_boundary_lookup_url()
```
lsirac, line 1787: Let's add a test that validates cloning a cred also keeps the blocking RAB behavior (e.g. for gcloud)

packages/google-auth/tests/test__regional_access_boundary_utils.py
```diff
@@ -301,6 +301,24 @@ def test_serialization(self):
         assert unpickled.refresh_manager._lock is not None
         assert unpickled.refresh_manager._worker is None
 
+    def test_unpickle_old_credentials_without_rab(self):
```
lsirac, line 305: Let's also add a test that validates with_scopes() and with_quota_project() will keep the relevant RAB data/state/blocking settings.

packages/google-auth/tests/test_iam.py
```diff
@@ -113,3 +114,37 @@ def test_sign_bytes_retryable_failure(self, mock_time):
...
+
+    # Mock check_use_client_cert to return True (simulating mTLS environment)
+    monkeypatch.setattr(_mtls_helper, "check_use_client_cert", lambda: True)
+
+    # Force a reload of the iam module to trigger the top-level domain computation
+    importlib.reload(iam)
+
+    try:
+        # Verify it constructed the mTLS domain for ALL endpoints
+        assert (
```
lsirac: Do we have tests that for mTLS that enable it and test explicitly that the correct endpoint gets used?

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
...
+            if hasattr(response_body1, "decode")
+            else response_body1
+        )
+
+        try:
+            response_data = json.loads(response_body)
+        except ValueError:
+            response_data = response_body
+
+        if response.status == http_client.OK:
```
lsirac: In `google.auth.aio.transport.Response`, the HTTP status code is exposed via the **`status_code`** property, not `status`. Passing a compliant `google.auth.aio` transport callable raises `AttributeError: 'Response' object has no attribute 'status'`. Please update the async lookup and grant methods to check `.status_code`.

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,145 @@ async def refresh_grant(
         request, token_uri, body, can_retry=can_retry
     )
     return client._handle_refresh_grant_response(response_data, refresh_token)
+
+
```
lsirac, line 293: is _jwt_async.py out of scope?

packages/google-auth/google/oauth2/_service_account_async.py
```diff
@@ -75,12 +82,36 @@ async def refresh(self, request):
...
+        )
+
     @_helpers.copy_docstring(credentials_async.Credentials)
     async def before_request(self, request, method, url, headers):
-        # Explicit override to bypass synchronous CredentialsWithRegionalAccessBoundary.
         await credentials_async.Credentials.before_request(
             self, request, method, url, headers
         )
+        self._maybe_start_regional_access_boundary_refresh(request, url)
+        self._rab_manager.apply_headers(headers)
```
lsirac: This may be redundant, why is this needed?

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -384,3 +385,61 @@ def start_refresh(self, credentials, request, rab_manager):
...
+            request (google.auth.aio.transport.Request): The object used to make
+                HTTP requests.
+            rab_manager (_RegionalAccessBoundaryManager): The manager container to update.
+        """
+        with self._lock:
+            if self._worker_task and not self._worker_task.done():
+                # A refresh is already in progress.
+                return
+
+            async def _worker():
```
lsirac, line 494: Unlike the synchronous refresh manager which safely deepcopies the transport (`copied_request = copy.deepcopy(request)`), the async manager passes the exact same `request` instance directly into the background coroutine task. Because `start_refresh` is invoked inside `before_request`, the main application coroutine immediately proceeds to make its actual service API HTTP call using the exact same `request` transport while the background task is concurrently using it, risking HTTP state corruption or interleaved headers.

Additionally, spawning `asyncio.create_task(_worker())` without tracking cancellation hooks upon client session closure can potentially cause dangling tasks that raise `RuntimeError: Session is closed` when executing against closed client sessions.

packages/google-auth/google/oauth2/_service_account_async.py
```diff
@@ -66,6 +67,12 @@ class Credentials(
         credentials = credentials.with_quota_project('myproject-123')
     """
 
+    def __init__(self, *args, **kwargs):
```
lsirac: `_service_account_async.Credentials` lacks a `__setstate__` override. When older pickled async credentials unpickle, they fall back to `CredentialsWithRegionalAccessBoundary.__setstate__`, which attaches a synchronous `_RegionalAccessBoundaryRefreshManager`. If a background RAB lookup triggers on this unpickled credential, a synchronous background thread will invoke `async def _lookup_regional_access_boundary` synchronously without awaiting it, causing a fatal thread crash (`AttributeError: 'coroutine' object has no attribute 'get'`). Please implement `__setstate__` to ensure `self._rab_manager.refresh_manager` is always restored as an `_AsyncRegionalAccessBoundaryRefreshManager()`.

lsirac (changes requested):

packages/google-auth/google/oauth2/_client_async.py
nbayati: I actually was able to run the code as-is and verify that it works. We have the same pattern at the top of this file in _token_endpoint_request_no_throw method. 
Screenshot 2026-05-13 at 6 33 28 PM

Though I did dig a bit to see why it's working and this is my current understanding: 
 in google-auth the async Request transport adapter wraps the raw response in a custom _CombinedResponse class (defined in google/auth/transport/_aiohttp_requests.py).
The _CombinedResponse class explicitly implements content as an asynchronous method to mirror the synchronous requests interface. Because it is a coroutine method returning the decoded byte body, await response.content() works as properly. 

I renamed the variable to `response_bytes` though, to make it a bit more intuitive.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Good catch! I've added the timeout to the request.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Good catch! I have updated the code to catch google.auth.exceptions.TransportError along with asyncio.TimeoutError and return (False, {}, False) on failure.

If a request fails due to a connection issue or timeout, the method now returns a non-retryable failure, which triggers the background refresh manager to enter the cooldown period. Any other unexpected exceptions will be caught and logged as warnings by the background worker task.

packages/google-auth/google/auth/credentials.py
nbayati: You're right, I had missed that the initialization of the credential would take care of the manager and didn't have to copy/create it here! Updated this method to only copy over the RAB data, and also _use_blocking_regional_access_boundary_lookup.

packages/google-auth/google/oauth2/_client_async.py
lsirac: Ty for validating. If this is only intended to support the legacy _aiohttp_requests transport then great, but we do then need to update docstrings (currently says google.auth.aio.transport.Request)

packages/google-auth/google/oauth2/_client_async.py
nbayati: Done!

packages/google-auth/google/oauth2/_service_account_async.py
nbayati: Thanks for flagging this. I actually realized we no longer need the before_request override, if we follow what we did with the sync RAB flow and using the _after_refresh. 

With this hook in place, the base class now automatically orchestrates the token refresh, RAB lookup, and header application in the correct order.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: Both concerns are valid, but safe under the hood:
1. Deepcopying the Transport is impossible and not needed in async
The async transport object (e.g., aiohttp_requests.Request) contains an active aiohttp.ClientSession with open TCP sockets. Attempting to copy.deepcopy(request) will instantly raise a TypeError: cannot pickle 'ClientSession' object and crash the application at runtime. Unlike synchronous transports running on separate OS threads, async HTTP clients (like aiohttp or httpx) are natively designed to handle concurrent requests sharing the same connection session. All request-specific states (headers, payloads) are stored in localized coroutine call stack frames, preventing any HTTP state corruption or interleaving.

2. Session closure and dangling tasks are handled safely
The background worker is a single-shot asyncio task that executes exactly one lookup request and then immediately terminates and gets garbage-collected.
If the user's application closes the underlying client session while a background task is still running, the resulting RuntimeError: Session is closed is safely caught by the worker's generic except Exception as e: block. It logs a warning, fails open cleanly, and does not raise an unhandled exception or crash the application.

I think no code changes are required here, as the current design is fully protected.

packages/google-auth/google/oauth2/_service_account_async.py
nbayati: Good catch. I've added `__setstate__` and a unit test.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: good catch, I've updated the comment to make it more accurate.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: Correct. I've updated the comment here as well, thanks for flagging it!

packages/google-auth/google/oauth2/_client_async.py
nbayati: Actually decided to only apply the fix to the RAB lookup, and instead open a bug to bring the grant methods up to date separately. This was can keep the blast radius of this PR smaller and limit it to only RAB changes without touching any token fetching flows.

packages/google-auth/tests_async/oauth2/test__client_async.py
nbayati: Done!

packages/google-auth/google/oauth2/_client_async.py
nbayati: Created #17139 to track the necessary changes for the token endpoints.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Done. Updated docstrings to document support for both standard (read()) and legacy (content()) async responses.

packages/google-auth/google/auth/credentials.py
```diff
@@ -364,12 +396,11 @@ def with_trust_boundary(self, trust_boundary):
...
 
     def _copy_regional_access_boundary_manager(self, target):
-        """Copies the regional access boundary manager to another instance."""
-        # Create a new manager for the clone to isolate background refresh locks and threads,
-        # but share the immutable data reference to avoid unnecessary initial lookups.
-        new_manager = _regional_access_boundary_utils._RegionalAccessBoundaryManager()
-        new_manager._data = self._rab_manager._data
-        target._rab_manager = new_manager
+        """Copies the regional access boundary manager state to another instance."""
+        target._rab_manager._data = self._rab_manager._data
```
lsirac, line 400: Tests are failing:

FAILED tests/test_external_account.py::TestCredentials::test_with_scopes_full_options_propagated - AttributeError: 'CredentialsImpl' object has no attribute '_rab_manager'
FAILED tests/test_external_account.py::TestCredentials::test_with_quota_project_full_options_propagated - AttributeError: 'CredentialsImpl' object has no attribute '_rab_manager'

Probably tests are configured incorrectly? I don't think it's possible for there to not be a rab manager?

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,159 @@ async def refresh_grant(
...
+                    timeout=timeout,
+                )
+            else:
+                response = await request(method="GET", url=url, headers=headers)
+
+            # Supports both modern google.auth.aio (exposing read()) and legacy transports (exposing content())
+            if hasattr(response, "read"):
+                response_bytes = await response.read()
+            else:
+                response_bytes = await response.content()
```
lsirac, line 415: This can raise an error too, but currently not being caught.

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,159 @@ async def refresh_grant(
...
+            return False, {}, False
+
+        try:
+            response_body = (
+                response_bytes.decode("utf-8")
+                if hasattr(response_bytes, "decode")
+                else response_bytes
+            )
+            response_data = json.loads(response_body)
+        except (UnicodeDecodeError, ValueError):
```
lsirac, line 432: As written, a retryable error response without a JSON body would not be retried because we return here before checking the status.

packages/google-auth/tests/test_jwt.py
```diff
@@ -553,6 +553,44 @@ def test_before_request_refreshes(self):
         self.credentials.before_request(None, "GET", "http://example.com?a=1#3", {})
         assert self.credentials.valid
 
+    def test_build_regional_access_boundary_lookup_url(self):
+        url = self.credentials._build_regional_access_boundary_lookup_url()
+        expected_url_standard = "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}/allowedLocations".format(
+            self.SERVICE_ACCOUNT_EMAIL
+        )
+        expected_url_mtls = "https://iamcredentials.mtls.googleapis.com/v1/projects/-/serviceAccounts/{}/allowedLocations".format(
```
lsirac: Can you remind me what prompted the mTLS addition?

We should have test(s) that set that signal and makes sure the right endpoint is called on refresh etc.

lsirac (changes requested):

packages/google-auth/tests_async/oauth2/test_service_account_async.py
nbayati: Done! Addressed by:

- Introducing start_blocking_refresh_async in _regional_access_boundary_utils.py to await the async lookup on the active event loop.
- Adding test_before_request_triggers_blocking_rab_refresh in test_service_account_async.py to cover async credentials with blocking RAB enabled.

packages/google-auth/tests/test_external_account.py
nbayati: Added test_copy_regional_access_boundary_manager_state_and_config in test_credentials.py to verify that copying generic credentials preserves the blocking lookup configuration.

Also added a matching subclass-level test in test_external_account.py to ensure cloning external account credentials also preserves it.

packages/google-auth/tests/test__regional_access_boundary_utils.py
nbayati: Done! Added test_copy_regional_access_boundary_manager_state_and_config_with_scopes and test_copy_regional_access_boundary_manager_state_and_config_with_quota_project in test_service_account.py

packages/google-auth/tests/test_iam.py
nbayati: There's just this test (test_endpoint_constants_mtls) that I added here. Unless you're referring to a different test gap that you've identified?
This test wasn't specific to RAB, but I noticed that there was no test for the dynamic switching of mtls vs non-mtls endpoints, and I added it here.

packages/google-auth/google/oauth2/_client_async.py
nbayati: No. I've also added support for jwt async.

packages/google-auth/tests/test_jwt.py
nbayati: I realized that we only had hardcoded the non-mtls endpoint, while the other existing endpoints switch between mtls and non-mtls dynalically at runtime. I added the same treatment to the RAB URL, however, based on my recent talk with Yao, it seems like it's safe to always just send the request to the mtls one. If that ends up being the case, we can simplify the code and always call the mtls RAB endpoint.

nbayati: Leaving this comment thread open, so I either come back and add more unit tests as you requested, or simplify the logic and update the existing unit tests.

packages/google-auth/google/auth/credentials.py
nbayati: Fixed. CredentialsImpl in test_external_account.py has been updated to correctly call super().__init__() in its constructor, ensuring that _rab_manager is successfully initialized on all mock instances.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Good catch! Updated the logic and added a unit test: test__lookup_regional_access_boundary_request_no_throw_non_json_bad_gateway_retry

packages/google-auth/google/oauth2/_client_async.py
nbayati: Fixed! Extended the try-except block to include the read() / content() calls, and simplified the handler to catch all base Exception errors, ensuring that any exceptions are caught and handled.

packages/google-auth/tests/test_jwt.py
nbayati: After further investigation, it turned out that we couldn't just always call mtls endpoint, and the library had to switch dynamically between the two. I did a refactoring to move the RAB endpoints to the RAB file and out of iam. I'm filing an issue to add unit tests to iam, and have them dynamically switch too, because the current static decision that happens in the beginning of the run won't cover the GCE's CUJ for dynamically changing identity to MWLID. I've opened a separate issue (#17282) to migrate the rest of the endpoints.

packages/google-auth/google/oauth2/_service_account_async.py
```diff
@@ -29,7 +29,9 @@
 
 
 class Credentials(
-    service_account.Credentials, credentials_async.Scoped, credentials_async.Credentials
+    service_account.Credentials,
```
lsirac: Because `_service_account_async.Credentials` inherits from `service_account.Credentials` first, Python's MRO resolves `service_account.Credentials` (and its parent, the synchronous `CredentialsWithRegionalAccessBoundary`) **before** the asynchronous `credentials_async.CredentialsWithRegionalAccessBoundary`.
As a result:
1. `_after_refresh` resolves to the **synchronous** implementation in `credentials.py` instead of the **asynchronous** implementation in `_credentials_async.py`.
2. This synchronous implementation runs `self._maybe_start_regional_access_boundary_refresh()`, which calls the synchronous `maybe_start_refresh()`.
3. Because this is an async credential, its `refresh_manager` is an instance of `_AsyncRegionalAccessBoundaryRefreshManager`.
4. The synchronous manager's call to `start_refresh` attempts to execute `asyncio.create_task()` synchronously. If no event loop is running in this context, it will raise `RuntimeError: no running event loop`. Even if an event loop is running, the background task is launched without the caller ever awaiting the actual refresh, leading to unawaited coroutines and silent failures under active load.
#### Proposed Fix
Place the asynchronous mixin class **before** the synchronous base class in the inheritance list to allow Python's MRO to resolve the async methods first:
```python
class Credentials(
    credentials_async.CredentialsWithRegionalAccessBoundary,
    service_account.Credentials,
    credentials_async.Scoped,
):

File: packages/google-auth/google/auth/_jwt_async.py [hunk 24]
```diff
@@ -142,6 +144,15 @@ class Credentials(
         new_credentials = credentials.with_claims(audience=new_audience)
     """
 
+    def __setstate__(self, state):
```
lsirac, line 149: `_jwt_async.OnDemandCredentials` inherits from `jwt.OnDemandCredentials` first. Because `jwt.OnDemandCredentials` implements `before_request` synchronously, Python's MRO resolves `before_request` to the synchronous parent, shadowing the async base.
When called via the experimental async HTTP transport (`_aiohttp_requests.py`), which has no initialization type-checking and explicitly awaits `before_request(...)`, the synchronous method executes, returns `None`, and the event loop attempts to `await None`, raising:
`TypeError: object NoneType can't be used in 'await' expression`

nbayati: Thanks for catching this. I've resolved this by adding an async def before_request override to the OnDemandCredentials class that delegates to the parent class, and updated the corresponding async test cases to properly await the call. 

As a side note, I have OnDemandCredentials on my radar as a class that might need to support RAB in the future, but I have decided to keep it out of scope for this PR.

nbayati:
Note to reviewers regarding mTLS transport gap:
Please note that while the RAB implementation in this PR dynamically switches endpoints to iamcredentials.mtls.googleapis.com when mTLS is enabled, the actual mTLS handshake will silently fall back to standard TLS (or fail under strict CBA enforcement) due to a library-wide transport gap.

The internal helper transport adapter (self._auth_request / self._request) passed to _regional_access_boundary_utils.py is not configured with the client certificate.

I have filed a tracking issue to address this transport gap: #17282

Because resolving the transport gap requires structural changes to google-auth's transport adapter initialization (requests.py and urllib3.py), we will track and merge the transport fix separately to keep the scope of this PR clean and focused on the core RAB logic. Once the transport issue is resolved, the RAB lookups in this feature will automatically execute over secure mTLS channels.

packages/google-auth/google/auth/_jwt_async.py
lsirac: Ok sounds good. I think we have a few open items to follow up on IIRC. Where are they being tracked?

lsirac (approved):
Great job! 

last thing - consider validating lookup response, but can be done in a followup

packages/google-auth/google/auth/_jwt_async.py
```diff
@@ -142,6 +145,15 @@ class Credentials(
...
+        """Restores the credential state and ensures the async refresh manager is attached."""
+        super().__setstate__(state)
+        from google.auth import _regional_access_boundary_utils
```
daniel-sanche: does this import have to be in the function? We typically keep these at the top of the file

packages/google-auth/google/auth/_regional_access_boundary_utils.py
```diff
@@ -227,6 +265,37 @@ def start_blocking_refresh(self, credentials, request):
 
         self.process_regional_access_boundary_info(regional_access_boundary_info)
 
+    async def start_blocking_refresh_async(self, credentials, request):
+        """Initiates a blocking lookup of the Regional Access Boundary asynchronously.
```
daniel-sanche, line 269: This seems confusing. If it's async, is it really blocking?

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,161 @@ async def refresh_grant(
...
+        )
+        return None
+    return response_data
+
+
+async def _lookup_regional_access_boundary_request_no_throw(
+    request, url, can_retry=True, headers=None, fail_fast=False
+):
+    """Makes a request to the Regional Access Boundary lookup endpoint. This
+        function doesn't throw on response errors.
```
daniel-sanche, line 372: I find tracing some of these functions a bit confusing. There's a `_lookup_regional_access_boundary_request` and `_lookup_regional_access_boundary_request_no_throw`, but it seems like neither of them throw?

I see this is the same pattern from the sync files though, so it can probably stay as-is

packages/google-auth/google/oauth2/_client_async.py
```diff
@@ -288,3 +289,161 @@ async def refresh_grant(
...
+                )
+            else:
+                response = await request(method="GET", url=url, headers=headers)
+
+            # Supports both modern google.auth.aio (exposing read()) and legacy transports (exposing content())
+            if hasattr(response, "read"):
+                response_bytes = await response.read()
+            else:
+                response_bytes = await response.content()
+        # Catch raw transport/socket exceptions raised during body streaming.
```
daniel-sanche: Is this intended to also quit after the first request timeout?

packages/google-auth/google/oauth2/_service_account_async.py
```diff
@@ -66,6 +68,15 @@ class Credentials(
         credentials = credentials.with_quota_project('myproject-123')
     """
 
+    def __setstate__(self, state):
+        """Restores the credential state and ensures the async refresh manager is attached."""
+        super().__setstate__(state)
+        from google.auth import _regional_access_boundary_utils
```
daniel-sanche: Does this import need to be in the method?

daniel-sanche (commented):
A few comments from my first pass. I'll try to take another look soon

packages/google-auth/tests/oauth2/test_service_account.py
```diff
@@ -228,15 +228,65 @@ def test_with_quota_project(self):
         new_credentials.apply(hdrs, token="tok")
         assert "x-goog-user-project" in hdrs
 
-    def test_build_regional_access_boundary_lookup_url(self):
+    def test_copy_regional_access_boundary_manager_state_and_config_with_scopes(self):
```
lsirac, line 231: We should add a test for external account + SA impersonation flow. This implementation will create an Impersonated cred with a new RAB manager. I think we may be dropping state that is needed (e.g. is blocking lookup enabled?).

Please add some coverage for this + validate that we are properly propagating the relevant RAB data.

packages/google-auth/google/auth/compute_engine/_metadata.py
```diff
@@ -502,3 +502,20 @@ def get_service_account_token(request, service_account="default", scopes=None):
         seconds=token_json["expires_in"]
     )
     return token_json["access_token"], token_expiry
+
+
+def _is_email(email):
```
lsirac: I think is_email is overpromising here for what we're checking. I think we can rename to is_service_account_email and implement with a simple regex check

packages/google-auth/google/auth/_jwt_async.py
nbayati: I'm tracking the additional credential types in the RAB google sheet tracker for a follow up release. 
For the items that weren't RAB specific, I have made a github issue:
- Updating token grant methods is tracked in #17139.
- Migrating the remaining endpoints is tracked in #17282.

packages/google-auth/google/auth/_jwt_async.py
nbayati: Done. Moved the import to the top of the file.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
nbayati: In this context, "blocking" refers to the logical flow of the request (i.e., we wait for the RAB lookup to complete before proceeding with the main request), as opposed to the background refresh mode. In the async implementation, this is achieved by awaiting the lookup, which suspends the coroutine but does not block the OS thread. The name matches the _use_blocking_regional_access_boundary_lookup configuration.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Correct, neither throws. The _no_throw version handles the low-level exceptions (timeouts, transport errors) and returns the detailed status tuple (status, data, retryable) used for retry logic. The other function acts as a wrapper that simplifies the interface by returning just data or None.

We kept this pattern to remain consistent with the sync implementation in _client.py.

packages/google-auth/google/oauth2/_client_async.py
nbayati: Good catch! I resolved it by updating the exception handler to set retryable_error = True and continue the retry loop if can_retry is enabled.

packages/google-auth/google/oauth2/_service_account_async.py
nbayati: Done. Moved the import to the top of the file.

packages/google-auth/google/auth/compute_engine/_metadata.py
nbayati: Done! I renamed the method and used `^[^@]+@[^@]+\.[^@]+$` which was recommended in the design for all SDKs to follow.

packages/google-auth/tests/oauth2/test_service_account.py
nbayati: Good catch! We were indeed dropping the blocking flag. I double checked the properties and I think that's the only thing we should propagate. Our helper method in the base class copies the blocking flag and the RAB data, but I decided not to use that, since the RAB data of the external account if it was seeded, would not necessarily be the RAB data of the impersonated account. 

I resolved this by propagating the blocking RAB lookup configuration (_use_blocking_regional_access_boundary_lookup) from the outer ExternalAccountCredentials to the inner ImpersonatedCredentials during initialization in _initialize_impersonated_credentials. Also added a unit test (test_refresh_impersonation_propagates_rab_config in tests/test_external_account.py) to verify the propagation.

packages/google-auth/google/auth/_regional_access_boundary_utils.py
daniel-sanche: I'd suggest trying to clarify that in the docstring

daniel-sanche (approved):
LGTM

lsirac (approved):
LGTM

PR #17093, opened 2026-05-13
refactor(bigframes): make executor internals async

TrevorBergeron:


packages/bigframes/bigframes/core/events.py
```diff
@@ -86,6 +91,16 @@ def publish(self, event: Event):
             for subscriber in self._subscribers:
                 subscriber(event)
 
+    async def publish_async(self, event: Event):
+        with self._subscribers_lock:
+            subscribers_snapshot = list(self._subscribers)
+        loop = asyncio.get_running_loop()
+        tasks = [
+            loop.run_in_executor(self._executor, subscriber, event)
```
tswast, line 99: This may eventually cause trouble. I've experienced having the main thread be the only one that can apply GUI updates in some frameworks, but I suppose we can address that (e.g. by having an option or something to run on the same thread as the original), but should be fine for the current use case of printing to stdout/stderr.

packages/bigframes/bigframes/session/_io/bigquery/__init__.py
```diff
@@ -55,11 +55,6 @@ def create_job_configs_labels(
     if job_configs_labels is None:
         job_configs_labels = {}
 
-    # If the user has labels they wish to set, make sure we set those first so
-    # they are preserved.
-    for key, value in bigframes.options.compute.extra_query_labels.items():
```
tswast, line 60: Did we move this feature somewhere else?

Edit: Yes. ExecutionSpec now reads this config.

tswast (approved):
Thanks!

PR #17116, opened 2026-05-14
fix(auth): fail-fast on invalid or non-workload certificate configs in agent identity discovery

### LINKED ISSUE #17145: Auth: docs CI check fails due to upstream requests 2.34.x docstring formatting regression

### Summary of the issue

The `docs` CI check is failing during the Sphinx documentation build for `google-auth` (and potentially other packages that inherit from `requests.Session`). The build fails with an unexpected `docutils` warning which is treated as an error due to the strict `-W` flag. Here's the log:

```
/google/auth/transport/requests.py:docstring of requests.sessions.Session.get:5: WARNING: Field list ends without a blank line; unexpected unindent. [docutils]
...
build finished with problems, 1 warning (with warnings treated as errors).
nox > Command sphinx-build -T -W -N -b html -d docs/_build/doctrees/ docs/ docs/_build/html/ failed with exit code 1
nox > Session docs failed.
```

#### **Root Cause**
The `requests` library recently released version `2.34.0` (with subsequent patch updates up to `2.34.2` today) which introduced inline typings and docstring refactoring. The new docstring for `requests.sessions.Session.get` contains a reStructuredText formatting error (missing indentation on the second line of the `:param params:` description):
   ```rst
   :param params: (optional) Dictionary, list of tuples or bytes to send
   in the query string for the :class:`Request`.  <-- Missing indentation here
   ```
`google-auth` defines `AuthorizedSession` which inherits from `requests.Session`. Sphinx's `autodoc` pulls in the inherited `get` docstring to document it, triggering the `docutils` warning.
Since the CI build uses `-W` (warnings as errors), this third-party formatting warning crashes the entire build.

#### **Suggested Workaround / Fix**
Until the issue is fixed upstream in the `requests` library, we can unblock the CI by pinning `requests` to a known stable version (e.g., `2.31.0`) specifically for the `docs` session in `noxfile.py`:

```python
@nox.session(python="3.10")
def docs(session):
    """Build the docs for this library."""

    session.install("-e", ".[aiohttp]")
    session.install("requests==2.31.0")  # Pin to avoid upstream docstring formatting bug
    session.install("sphinx", "alabaster", "recommonmark", "sphinx-docstring-typing")
    ...
```

nbayati:
The `GOOGLE_API_CERTIFICATE_CONFIG` environment variable is shared between Managed Workload Identity (MWLID) token-binding and other flows like Enterprise Certificate Provider (ECP) configs (e.g., PKCS#11).

When this env var is set, agent identity discovery is triggered. If the config file exists but lacks the `"workload"` section (as with ECP configurations),we should exit early and return `None` to avoid delaying non-workload flows.

In addition, if the config file on disk had syntax errors or invalid JSON, the previous logic entered a 30-second blocking retry loop before failing with `RefreshError`. To resolve this, the lookup logic now assumes that if the config file exists on disk, it is in its final format. If the file exists but lacks a `"workload"` section, has syntax errors, or is unreadable, we return `None` immediately to fail-fast and avoid startup delays.

b/512912028
fixes #17145

### DIFF
--- a/packages/google-auth/google/auth/_agent_identity_utils.py
+++ b/packages/google-auth/google/auth/_agent_identity_utils.py
@@ -80,7 +80,7 @@ def get_agent_identity_certificate_path():
     import json
 
     cert_config_path = os.environ.get(environment_vars.GOOGLE_API_CERTIFICATE_CONFIG)
-    
+
     # Check if the well-known workload directory is mounted.
     well_known_dir = os.path.dirname(_WELL_KNOWN_CERT_PATH)
     has_well_known_dir = os.path.exists(well_known_dir)
@@ -98,13 +98,16 @@ def get_agent_identity_certificate_path():
             if cert_config_path:
                 with open(cert_config_path, "r") as f:
                     cert_config = json.load(f)
-                    
+
                     if not isinstance(cert_config, dict):
                         return None
 
                     cert_configs = cert_config.get("cert_configs", {})
                     workload_config = cert_configs.get("workload", {})
-                    if not isinstance(workload_config, dict) or "cert_path" not in workload_config:
+                    if (
+                        not isinstance(workload_config, dict)
+                        or "cert_path" not in workload_config
+                    ):
                         return None
 
                     cert_path = workload_config["cert_path"]
--- a/packages/google-auth/google/auth/_agent_identity_utils.py
+++ b/packages/google-auth/google/auth/_agent_identity_utils.py
@@ -103,6 +103,9 @@ def get_agent_identity_certificate_path():
                         return None
 
                     cert_configs = cert_config.get("cert_configs", {})
+                    if not isinstance(cert_configs, dict):
+                        return None
+
                     workload_config = cert_configs.get("workload", {})
                     if (
                         not isinstance(workload_config, dict)
--- a/packages/google-auth/google/auth/_agent_identity_utils.py
+++ b/packages/google-auth/google/auth/_agent_identity_utils.py
@@ -80,41 +80,77 @@ def get_agent_identity_certificate_path():
     import json
 
     cert_config_path = os.environ.get(environment_vars.GOOGLE_API_CERTIFICATE_CONFIG)
-    if not cert_config_path:
+    
+    # Check if the well-known workload directory is mounted.
+    well_known_dir = os.path.dirname(_WELL_KNOWN_CERT_PATH)
+    has_well_known_dir = os.path.exists(well_known_dir)
+
+    # If we have neither a config path nor a well-known mount directory, exit immediately.
+    if not cert_config_path and not has_well_known_dir:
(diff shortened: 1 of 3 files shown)


packages/google-auth/google/auth/_agent_identity_utils.py
```diff
@@ -80,41 +80,83 @@ def get_agent_identity_certificate_path():
...
 
     for interval in _POLLING_INTERVALS:
         try:
-            with open(cert_config_path, "r") as f:
-                cert_config = json.load(f)
-                cert_path = (
-                    cert_config.get("cert_configs", {})
-                    .get("workload", {})
-                    .get("cert_path")
+            # Path A: Config file is explicitly set
```
agrawalradhika-cell, line 97: Can use a snippet as - 

```
        # Path A: Config file is explicitly set
        if cert_config_path:
            with open(cert_config_path, "r") as f:
                cert_config = json.load(f)

            # Validate the nested dictionary structure concisely
            cert_configs = cert_config.get("cert_configs") if isinstance(cert_config, dict) else None
            workload_config = cert_configs.get("workload") if isinstance(cert_configs, dict) else None

            if not isinstance(workload_config, dict) or "cert_path" not in workload_config:
                return None

            cert_path = workload_config["cert_path"]
            if _is_certificate_file_ready(cert_path):
                return cert_path
            target_path = cert_path

        # Path B: Config is NOT set, fallback to the well-known path
        else:
            if _is_certificate_file_ready(_WELL_KNOWN_CERT_PATH):
                return _WELL_KNOWN_CERT_PATH
            target_path = _WELL_KNOWN_CERT_PATH
```

to reduce the nesting. Wdyt?

packages/google-auth/google/auth/_agent_identity_utils.py
```diff
@@ -80,41 +80,83 @@ def get_agent_identity_certificate_path():
...
+                        return None
+
+                    cert_path = workload_config["cert_path"]
+                    if _is_certificate_file_ready(cert_path):
+                        return cert_path
+
+                    # The config was parsed, but the cert file is not ready yet
+                    target_path = cert_path
+
+            # Path B: Config is NOT set, fallback to the the well-known path
```
agrawalradhika-cell: nit: # Path B: Config is NOT set, fallback to the well-known path

agrawalradhika-cell (approved):
Provided some minor comments, rest looks good!

packages/google-auth/google/auth/_agent_identity_utils.py
nbayati: I'm not sure I'm following what this nit is about. Can you please clarify?

nbayati: Ah I see, thanks for flagging it, Done!

PR #17044, opened 2026-05-12
chore: allow releases for google-cloud-compute

parthea:
Fixes https://github.com/googleapis/google-cloud-python/issues/16780

PR #17130, opened 2026-05-14
chore: make librarian.yaml consistent with .librarian/state.yaml

jskeet:
(This was caused by #17044, but wasn't caught *in* that PR as the librarian version at the time wasn't checking for release consistency.)


chalmerlowe (approved):
LGTM

PR #17143, opened 2026-05-14
feat: regenerate automl using generator v1.32.0

parthea:
Updates post processing to account for dropping support for Python 3.7, 3.8, 3.9 and impacts that has on using 3.10 for lower bounds testing.

Bumps `google-auth` to `>=2.23.3` to prevent dependency resolution issue when used in sync with `google-cloud-storage`

Corrects a deficiency in the constraints template that did not capture all the necessary bounds from `setup.py`

Partially corrects #17153 🦕


chalmerlowe (approved):
LGTM

PR #17148, opened 2026-05-15
feat: regenerate google-cloud-datacatalog-lineage

jskeet:


jskeet:
This was skipped in #16997 - I'm not sure why, but that was with the previous generator. If it all looks okay now, we should unskip.

chalmerlowe (approved):
LGTM

PR #17159, opened 2026-05-15
chore(.generator): release staging preview support

noahdietz:
Updates `legacylibrarian` Python processor to support staging releases for Preview packages.

Preview packages are generated into `./preview-packages` directory with an package name and directory mirroring the stable variant under `./packages` ([doc](https://docs.google.com/document/d/1NQtgQlRKch5bqFYmbOKuiEUx2dONvd8Ld_QITDSf0eM/edit?pli=1&tab=t.u0wv82tu3hqq)). Their `legacylibrarian` ID, however, is `-preview` suffixed (so that there can be a unique entry in the config), and used as a signal to the tooling that the library in question is a preview library and is in a different location. The contents of the library package is otherwise identical.

Towards https://github.com/googleapis/librarian/issues/5894


parthea:
Converting to draft since presubmit is failing

noahdietz:
> Converting to draft since presubmit is failing

Sorry about that, thanks for switching it to draft. gtg now!

PR #17168, opened 2026-05-18
feat: add `bigframes.bigquery.aead.*` scalar functions

tswast:


tswast:
Tested manually using the example at https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/aead_encryption_functions#aeaddecrypt_bytes

image

packages/bigframes/bigframes/core/sentinels.py
sycai, line 1: I need some help to understand the design of this file.

We are defining a group of sentinel values, then why don't we do something like this:

```
class Sentinel(Enum):
  DEFAULT_ARG = 1
  ...
```

Plus, what's the "token" used for? And shall we capitalize all its letters?

packages/bigframes/scripts/generate_bigframes_bigquery.py
```diff
@@ -0,0 +1,368 @@
...
+    # Replace dots with underscores
+    name = name.replace(".", "_")
+    # Handle CamelCase to snake_case
+    name = re.sub(r"(?<!^)(?=[A-Z])", "_", name).lower()
+    # Replace multiple underscores with one
+    name = re.sub(r"_+", "_", name)
+    return name
+
+
+def main():
```
sycai: Maybe we should break this function down to multiple helpers. It's not very readable in its current state.

packages/bigframes/tests/unit/bigquery/_operations/conftest.py
sycai, line 1: Not necessary in this PR, but maybe we should make this conftest setup globally available for all tests? It looks very similar with [packages/bigframes/tests/unit/core/compile/sqlglot/conftest.py](https://github.com/googleapis/google-cloud-python/blob/49252da0f81c63ff7bc384dc3ddd92608e15f459/packages/bigframes/tests/unit/core/compile/sqlglot/conftest.py)

packages/bigframes/scripts/data/sql-functions/aead.yaml
```diff
@@ -0,0 +1,131 @@
+urn: extension:google:bq_scalar_functions
+scalar_functions:
+  - name: "aead.decrypt_bytes"
+    description: "Uses the matching key from keyset to decrypt ciphertext and verifies the integrity of the data using additional_data. Returns an error if decryption or verification fails."
+    impls:
+      # Signature: aead.decrypt_bytes:vbin_vbin_vbin
+      - args:
+          - name: "keyset"
+            value: binary
```
sycai, line 9: Maybe we should use "type" instead of "value" to indicate the type of the arg?

packages/bigframes/bigframes/bigquery/_operations/aead.py
```diff
@@ -0,0 +1,92 @@
...
+)
+_ENCRYPT_OP = googlesql.GoogleSqlScalarOp(
+    "AEAD.ENCRYPT",
+    args=(googlesql.ArgSpec(), googlesql.ArgSpec(), googlesql.ArgSpec()),
+    signature=lambda *args: dtypes.BYTES_DTYPE,
+)
+
+
+def decrypt_bytes(
+    keyset: Union[T, bigframes.core.col.Expression, Union[bytes, dict]],
```
sycai: Is this "bigframes.core.col.Expression" necessary, given that it's already included in "T"?

tswast: Yes. My intention is to allow a mix of types. Basically, a mix of Series and Expression should give Series. Just Expression should give Expression.

packages/bigframes/scripts/data/sql-functions/aead.yaml
tswast: My goal is to match substrait spec as much as possible. See: https://substrait.io/extensions/#simple-extensions and https://substrait.io/expressions/scalar_functions/

packages/bigframes/tests/unit/bigquery/_operations/conftest.py
tswast: It is in fact identical.

packages/bigframes/bigframes/core/sentinels.py
tswast: It's just a place holder. With your proposal, we might be able to use the Literal type. I can give it a try.

packages/bigframes/scripts/generate_bigframes_bigquery.py
tswast: I'll give it a try.

packages/bigframes/bigframes/bigquery/_operations/aead.py
sycai: I see your point. How about ditching the generic type entirely and spell out all possible types in the type hints? I don't think "T" brings us any significant advantages here. Plus, it makes `mypy` unhappy.

packages/bigframes/scripts/templates/test_operation.py.j2
```diff
@@ -0,0 +1,28 @@
...
+
+
+{% for func in functions %}
+def test_{{ func.name }}(scalar_types_df: bpd.DataFrame, snapshot):
+    result = {{ short_name }}.{{ func.name }}(
+{% for arg in func.test_args %}
+        cast(bpd.Series, scalar_types_df["{{ arg.col_name }}"]),
+{% endfor %}
+    ).to_frame()
+    snapshot.assert_match(result.sql.rstrip() + "\n", "out.sql")
```
sycai, line 26: nit: an empty line before the assert_match to demarcate the "act" and "assert" block

sycai (approved):

packages/bigframes/bigframes/bigquery/_operations/aead.py
tswast: It does bring a big advantage: mypy can determine when all are series, that it should return a series. The alternative is to return a Union type, but then we'd have to cast any time we use one of these functions.

packages/bigframes/scripts/templates/test_operation.py.j2
tswast: Done.

PR #17174, opened 2026-05-19
test(spanner): handle AlreadyExists idempotently on restore test retries

sakthivelmanii:
When a restore database test times out on the client-side, Spanner continues the restore operation in the background. When the test runner retries the test, the call would fail with AlreadyExists.

This change catches the AlreadyExists exception, retrieves the database metadata, and asserts that it is restoring from the correct source backup and achieves a READY or READY_OPTIMIZING state. This allows the test to execute idempotently without namespace collisions.

### DIFF
--- a/packages/google-cloud-spanner/samples/samples/backup_sample_test.py
+++ b/packages/google-cloud-spanner/samples/samples/backup_sample_test.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import uuid
 
-from google.api_core.exceptions import DeadlineExceeded
+from google.api_core.exceptions import AlreadyExists, DeadlineExceeded
 import pytest
 from test_utils.retry import RetryErrors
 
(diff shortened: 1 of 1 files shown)


olavloite:
I don't quite understand this comment:

> When the test runner retries the test, the call would fail with AlreadyExists.

When would the test runner retry the test? Or do you mean when a new test run of all the integration tests are executed? If it is the latter, isn't this rather an indication that there is a problem with this test re-using the same database name when restoring a database? And that the actual fix would be to use a dynamically generated database name?

(Note: I would also be open to just remove/skip these backup tests. I don't think they add much value, as the client code that is being tested is minimal.)

sakthivelmanii:
For backup tests, we do test retry errors like below

`@RetryErrors(exception=(DeadlineExceeded, TimeoutError), max_tries=2)`.

when we take a `backup A` and tried to restore to `db A`, we will wait for more than 20m to restore. if it fails due to timeout, we retry the entire test again. retry means, we will try to restore `backup A` to  `db A` again and it will fail because backup is already in progress and `db A` was already created which throws `AlreadyExists` error. To avoid this, we are further waiting the database to be ready rather than restoring it.

olavloite (approved):

PR #17175, opened 2026-05-19
test(spanner): upgrade Spanner sample instance fixtures to session scope

sakthivelmanii:
Previously, standard and multi-region Spanner instance fixtures in conftest.py were module-scoped, meaning a brand-new Spanner instance was created and destroyed for every single test file, adding massive execution time (~2 minutes per file) in CI/CD runs.

This change upgrades the instance and config fixtures to session scope. To prevent ScopeMismatch exceptions, the dependencies on module-scoped sample_name are replaced with a generic shared-samples label. This allows the test suite to reuse a single shared Spanner instance, dramatically optimizing test execution times.

### DIFF
--- a/packages/google-cloud-spanner/samples/samples/conftest.py
+++ b/packages/google-cloud-spanner/samples/samples/conftest.py
@@ -33,17 +33,6 @@ retry_cleanup = retry.RetryErrors(
     (exceptions.ResourceExhausted, exceptions.FailedPrecondition), max_tries=6, delay=15
 )
 
-
-@pytest.fixture(scope="module")
-def sample_name():
-    """Sample testcase modules must define this fixture.
-
-    The name is used to label the instance created by the sample, to
-    aid in debugging leaked instances.
-    """
-    raise NotImplementedError("Define 'sample_name' fixture in sample test driver")
-
-
 @pytest.fixture(scope="module")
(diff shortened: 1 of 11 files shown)


packages/google-cloud-spanner/samples/samples/conftest.py
olavloite: Won't this cause race conditions if multiple integration test runs are started in parallel? They will now use the same instance, and potentially try to use the same database name.

packages/google-cloud-spanner/samples/samples/conftest.py
```diff
@@ -134,13 +123,12 @@ def proto_descriptor_file():
     file.close()
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def sample_instance(
     spanner_client,
     cleanup_old_instances,
```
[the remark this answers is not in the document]
olavloite: `spanner_client` and `cleanup_old_instances` were already session scope, so this should not cause any problems.

packages/google-cloud-spanner/samples/samples/conftest.py
sakthivelmanii: I fixed it. created a new unique db name

olavloite (approved):

PR #17189, opened 2026-05-19
fix: cast JSON and nested struct columns to string for anywidget rendering

shuoweil:
This Pull Request resolves visualization crashes when rendering DataFrames or Series containing raw JSON and nested JSON struct structures. It ensures that these columns are safely pre-serialized into clean, flat JSON string format on the database level prior to visual layout rendering.

Verified at: screen/424ojbuqyBPinTb

Fixes #<514763826> 🦕


packages/bigframes/tests/unit/display/test_anywidget.py
```diff
@@ -179,3 +179,69 @@ def test_page_size_change_resets_sort(mock_df):
 
     # to_pandas_batches called again (reset)
     assert mock_df.to_pandas_batches.call_count >= 2
+
+
+def test_json_column_converted_to_string_for_display():
+    from bigframes.core.blocks import Block
```
sycai: Is it possible to import everything at the top of the file, so that we don't have to repeat these lines in each test?

packages/bigframes/tests/unit/display/test_html.py
```diff
@@ -185,3 +185,39 @@ def test_render_html_max_columns_truncation():
     assert "col_8" not in html
     assert "col_9" in html
     assert "..." in html
+
+
+def test_repr_mimebundle_head():
+    from unittest.mock import Mock, patch
```
sycai: likewise, import at the top of the file?

packages/bigframes/bigframes/dataframe.py
```diff
@@ -820,8 +820,24 @@ def __repr__(self) -> str:
         )
 
     def _get_display_df_and_blob_cols(self) -> tuple[DataFrame, list[str]]:
```
sycai: Maybe we can ditch the second return value and just return the display df? This this a private function so it should be safe to change its return type.

The only concern I have is that, if this function is called in many places, updating the callers might be a chore. However, if it only has around 5 callers or so, we can make the change in this PR.

packages/bigframes/bigframes/display/html.py
```diff
@@ -227,6 +227,16 @@ def _get_obj_metadata(
     return is_series, has_index
 
 
+def _to_display_df(
+    obj: Union[bigframes.dataframe.DataFrame, bigframes.series.Series],
+) -> bigframes.dataframe.DataFrame:
+    from bigframes.series import Series
```
sycai: Is it possible to put this import to the top of the file?

packages/bigframes/bigframes/display/html.py
```diff
@@ -227,6 +227,16 @@ def _get_obj_metadata(
...
+
+    df = obj.to_frame() if isinstance(obj, Series) else obj
```
sycai: Here is another design suggestion for you: instead of performing the type check in the `_to_display_df()`, define a "_get_display_df_and_blob_cols" method directly on the `series.Series` class, which basically looks like this:

```
def _get_display_df_and_blob_cols(self):
  return self.to_frame()._get_display_df_and_blob_cols()
```

There are two benefits:
* You don't need to perform type checks here any more
* You provide better functional parity between the `Dataframe` class and the `Series` class

packages/bigframes/tests/unit/display/test_anywidget.py
shuoweil: Moved all duplicate local imports (Block, DataFrame, JSON_DTYPE, STRING_DTYPE, struct_type, SqlScalarOp, and TableWidget) to the top of test_anywidget.py. Placed them immediately following the pytest.importorskip guards to ensure module import safety.

packages/bigframes/bigframes/dataframe.py
shuoweil: Simplified _get_display_df_and_blob_cols to _get_display_df which returns only the processed DataFrame. Updated all matching test mocks and calls in both test_anywidget.py and test_html.py to use the new single-return signature.

packages/bigframes/bigframes/display/html.py
shuoweil: Added a polymorphic _get_display_df(self) method directly to the Series class, implemented as return self.to_frame()._get_display_df(). Leveraged this signature in html.py to simplify _to_display_df(obj) to a clean, type-check-free one-liner: return obj._get_display_df().

packages/bigframes/bigframes/display/html.py
```diff
@@ -227,6 +224,12 @@ def _get_obj_metadata(
     return is_series, has_index
 
 
+def _to_display_df(
+    obj: Union[bigframes.dataframe.DataFrame, bigframes.series.Series],
+) -> bigframes.dataframe.DataFrame:
+    return obj._get_display_df()
+
+
```
sycai: This function now becomes so simple that we can simply remove it. Just use `obj._get_display_df()` at its call sites.

shuoweil: Good catch, it is removed.

sycai (approved):

PR #17213, opened 2026-05-20
feat: add more scalar array functions to `bigframes.bigquery`

tswast:
🦕


packages/bigframes/bigframes/bigquery/__init__.py
```diff
@@ -72,6 +68,46 @@
     st_regionstats,
     st_simplify,
 )
+from bigframes.bigquery._operations.global_namespace.array import (
```
tswast: Why did we need these aliases? I'll double check the generator.

tswast: Fixed.

packages/bigframes/third_party/bigframes_vendored/sqlglot/expressions.py
```diff
@@ -5906,7 +5906,7 @@ class FarmFingerprint(Func):
 
 
 class Flatten(Func):
-    pass
+    arg_types = {"this": True, "depth": False}
```
tswast, line 5909: I wonder why it made this change?

tswast: Jetski says "Enhanced SQLGlot ( third_party/bigframes_vendored/sqlglot/expressions.py ):
      • Fixed a  ValueError  during  test_flatten  by updating the vendored  Flatten  class definitions. It now explicitly supports the optional  depth  parameter ( arg_types = {"this": True,     
      "depth": False} ), matching BigQuery's actual  FLATTEN  function signature.
"

packages/bigframes/bigframes/operations/googlesql/global_namespace/array.py
```diff
@@ -0,0 +1,421 @@
...
+import bigframes.operations as ops
+import bigframes.series as series
+from bigframes import dtypes
+from bigframes.operations import googlesql
+
+_ARRAY_CONCAT_OP = googlesql.GoogleSqlScalarOp(
+    "ARRAY_CONCAT",
+    args=(googlesql.ArgSpec(), googlesql.ArgSpec()),
+    signature=lambda *args: None,
+)
```
TrevorBergeron, line 80: Could we place the GoogleSqlScalarOp defs somewhere public/central so that we can also use them in the compiler. My intention with the GoogleSqlScalarOp is that it can be the single definition of a google sql function for both internal and user-facing purposes.

packages/bigframes/tests/unit/operations/googlesql/global_namespace/snapshots/test_array/test_array_concat/out.sql
```diff
@@ -0,0 +1,4 @@
+SELECT
+  `rowindex`,
+  ARRAY_CONCAT(`string_col`,   `string_col`) AS `0`
+FROM `bigframes-dev`.`sqlglot_test`.`scalar_types` AS `bft_0`
```
TrevorBergeron: These start to be a bit silly now that they all exercise the same exact emitter code eh?

tswast: True! It was helpful at the start to make sure everything is plumbed through, but now it's coverage for coverage's sake. I'll ponder.

packages/bigframes/bigframes/operations/googlesql/global_namespace/array.py
tswast: That should be doable. I'll adjust the generator.

tswast: Done. Moved to `bigframes.operations.googlesql.*`

packages/bigframes/tests/unit/operations/googlesql/global_namespace/snapshots/test_array/test_array_concat/out.sql
tswast: Updated the generated tests to check that the `bigframes.bigquery` method we expose manaully actually corresponds to the expected op, which I do think has some utility.

tswast:
Failure looks like a real one. I need to actually generate the signature so that it doesn't see the none type and cast to float.

Edit: resolved in https://github.com/googleapis/google-cloud-python/pull/17213/changes/3d144f89ac7c111e91d6cc282f8d3057b70159c2 . confirmed locally

TrevorBergeron (approved):

PR #17230, opened 2026-05-22
ci: pre-populate synthtool cache to avoid race condition

suztomo:
This PR resolves a blocking parallel generation issue on GitHub Actions and updates Spanner's post-processing config to align with recent template updates.

### 1. Fix GHA parallel generation race condition
When running `librarian generate --all` in GHA, multiple libraries are generated in parallel. They concurrently invoke `synthtool`, which races to clone/update the templates repository (`googleapis/synthtool`) in the shared cache directory `~/.cache/synthtool/synthtool`. 

This results in either:
*   A **clone race**: One process starts cloning, and others see the half-created directory, assume it's valid, and fail with `fatal: failed to resolve HEAD as a valid ref` when running `git branch`.
*   An **update race**: Multiple processes try to run `git checkout` / `git pull` concurrently, failing with `fatal: Unable to create ... index.lock: File exists`.

**Fix**: 
We now clone `googleapis/synthtool` once sequentially at the beginning of the workflow (`Clone Synthtool Templates`), and configure `SYNTHTOOL_TEMPLATES` environment variable globally to point to this local directory. This completely bypasses all concurrent git clone/pull operations during parallel generation, resolving the race conditions and speeding up the build.

### 2. Fix Spanner post-processing AssertionError
We removed the obsolete `Supported Python Versions` replacement block from `.librarian/generator-input/client-post-processing/spanner-integration.yaml`.

**Why it was needed**:
*   On **May 19, 2026** (commit `2408166fec`), Python 3.7-3.9 support was dropped specifically for Spanner. Because the `synthtool` templates still generated `Python >= 3.9` in `README.rst` by default, a post-processing workaround was added to force it to `3.10`.
*   On **May 21, 2026** (commit `e643ce8` in `googleapis/synthtool`), `synthtool` templates were updated globally to drop Python 3.9 support for all mono-repo libraries.
*   As a result, the freshly generated `README.rst` for Spanner now already defaults to `Python >= 3.10`. The May 19 workaround became obsolete and failed with `AssertionError: Replaced 0 rather than 1 instances` because the "before" pattern (expecting `3.9`) no longer exists.

TAG=agy
CONV=acb0ee9b-3d87-4b9e-b3c6-0d4aa3efbd35

### DIFF
--- a/.github/workflows/regenerate-all.yml
+++ b/.github/workflows/regenerate-all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
 
 permissions:
   contents: read
--- a/.github/workflows/regenerate-all.yml
+++ b/.github/workflows/regenerate-all.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  pull_request:
 
 permissions:
   contents: read
@@ -40,6 +41,11 @@ jobs:
         run: |
           version=$(sed -n 's/^version: *//p' librarian.yaml)
           go run "github.com/googleapis/librarian/cmd/librarian@${version}" install
+
+      - name: Pre-populate Synthtool Cache
+        run: |
+          mkdir -p ~/.cache/synthtool
+          git clone --recurse-submodules --single-branch https://github.com/googleapis/synthtool.git ~/.cache/synthtool/synthtool
 
       - name: Regenerate
         run: |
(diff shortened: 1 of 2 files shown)


suztomo:
Running the check https://github.com/googleapis/google-cloud-python/actions/runs/26263440317/job/77301656243?pr=17230

suztomo:
It failed.

suztomo:
Regenerate succeeded in https://github.com/googleapis/google-cloud-python/actions/runs/26264596009/job/77305110304.

ohmayr (approved):

PR #17236, opened 2026-05-22
chore(ci): add librarian tidy check

### LINKED ISSUE #5908: google-cloud-python: add librarian tidy check action
When manually editing the `librarian.yaml`, it is easy to put properties in the "wrong" place within a given entry and forget to run `librarian tidy`.

We should have a simple GitHub Action that lints the `librarian.yaml` by running `librarian tidy` with the version of `librarian` specified in the `librarian.yaml`. This can be retrieved programmatically by doing:
```shell
V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
go run github.com/googleapis/librarian/cmd/librarian@${V} tidy
```

Then fail the check if the diff is not clean, and emit the diff in the error message.

noahdietz:
Adds a `librarian tidy` check for pull requests touching `librarian.yaml`.

It uses the version of librarian specified in the `librarian.yaml`, runs `librarian tidy`, reports any diff, and fails if there is any diff.

Fixes https://github.com/googleapis/librarian/issues/5908

### DIFF
--- a/.github/workflows/librarian_tidy.yml
+++ b/.github/workflows/librarian_tidy.yml
@@ -1,4 +1,4 @@
-name: Check that librarian.yaml is tidy
+name: librarian tidy
 
 on:
   pull_request:
--- a/.github/workflows/librarian_tidy.yml
+++ b/.github/workflows/librarian_tidy.yml
@@ -22,18 +22,7 @@ jobs:
 
       - name: Run librarian tidy
         run: |
-          # go run can flake if the downloads fail, so run it several times until it succeeds.
-          go run github.com/googleapis/librarian/cmd/librarian@latest version || \
-          go run github.com/googleapis/librarian/cmd/librarian@latest version || \
-          go run github.com/googleapis/librarian/cmd/librarian@latest version
-
           V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
-
-          # This is a different download, so it can flake even after the first one succeeds.
-          go run github.com/googleapis/librarian/cmd/librarian@${V} version || \
-          go run github.com/googleapis/librarian/cmd/librarian@${V} version || \
-          go run github.com/googleapis/librarian/cmd/librarian@${V} version
-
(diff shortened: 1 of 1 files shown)


.github/workflows/librarian_tidy.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: '1.26.x'
+
+      - name: Run librarian tidy
```
parthea, line 23: Gemini is recommending that we use `go install` instead of `go run`, and also handle the final attempt

```
- name: Run Tidy and Check Diff
        run: |
          set -euo pipefail

          # Retry loop for the flaky network install
          success=false
          for i in {1..3}; do
            echo "Attempt $i: Installing librarian..."
            if go install github.com/googleapis/librarian/cmd/librarian@latest; then
              success=true
              break
            fi
            echo "Attempt $i failed. Retrying in 2 seconds..."
            sleep 2
          done

          if [ "$success" = false ]; then
            echo "❌ Failed to install librarian after 3 attempts due to network flakiness."
            exit 1
          fi

          # -------------------------------------------------------------
          # From here down, everything is OFFLINE and 100% stable
          # -------------------------------------------------------------
          V=$(librarian config get version)
          
          # Install the specific project version
          go install github.com/googleapis/librarian/cmd/librarian@${V}

          librarian tidy

          if ! git diff --exit-code; then
            echo "❌ librarian.yaml is not tidy. Please run 'go run github.com/googleapis/librarian/cmd/librarian@${V} tidy' locally."
            exit 1
          fi
```

noahdietz: I think it would be easier to just retry the action if it flakes. We don't need to build in retries. Simplified it!

parthea (approved):

PR #17248, opened 2026-05-26
chore(storage): support DISABLE_GCS_PYTHON_CLIENT_OTEL_BUCKET_METADATA environment variable flag

chandra-siri:
This PR introduces the `DISABLE_GCS_PYTHON_CLIENT_OTEL_BUCKET_METADATA` environment flag to dynamically disable injecting Cloud Storage bucket metadata destination attributes (`gcp.resource.destination.id` and `gcp.resource.destination.location`) inside OTel spans.

### DIFF
--- a/packages/google-cloud-storage/google/cloud/storage/_helpers.py
+++ b/packages/google-cloud-storage/google/cloud/storage/_helpers.py
@@ -150,8 +150,17 @@ def _validate_name(name):
 def create_trace_span_helper(client, bucket_name, name, attributes=None, **kwargs):
     span_attrs = dict(attributes) if attributes else {}
 
+    # Check if GCS destination annotations are explicitly disabled via environment
+    disable_bucket_md = os.environ.get("DISABLE_BUCKET_MD_IN_OTEL", "").lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
     if (
-        bucket_name
+        not disable_bucket_md
+        and bucket_name
         and isinstance(bucket_name, str)
         and client
         and hasattr(client, "_bucket_metadata_cache")
--- a/packages/google-cloud-storage/google/cloud/storage/_helpers.py
+++ b/packages/google-cloud-storage/google/cloud/storage/_helpers.py
@@ -150,21 +150,16 @@ def _validate_name(name):
 def create_trace_span_helper(client, bucket_name, name, attributes=None, **kwargs):
     span_attrs = dict(attributes) if attributes else {}
 
-    # Check if GCS destination annotations are explicitly disabled via environment
-    disable_bucket_md = os.environ.get("DISABLE_BUCKET_MD_IN_OTEL", "").lower() in (
-        "1",
-        "true",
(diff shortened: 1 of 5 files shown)


packages/google-cloud-storage/google/cloud/storage/_helpers.py
```diff
@@ -156,6 +156,10 @@ def create_trace_span_helper(client, bucket_name, name, attributes=None, **kwarg
         and client
         and hasattr(client, "_bucket_metadata_cache")
         and client._bucket_metadata_cache
+        and not (
+            os.environ.get("DISABLE_BUCKET_MD_IN_OTEL", "").lower()
+            in ("1", "true", "yes", "on")
```
krishnamd-jkp: Instead of checking for all values, can't we just document the usage of this environment variable instead?

packages/google-cloud-storage/google/cloud/storage/_http.py
```diff
@@ -88,6 +89,10 @@ def api_request(self, *args, **kwargs):
             and enable_otel_traces
             and hasattr(client, "_bucket_metadata_cache")
             and client._bucket_metadata_cache
+            and not (
```
kalragauri: nit: consider extracting this into a utility method or using an existing one?

packages/google-cloud-storage/google/cloud/storage/_helpers.py
```diff
@@ -156,6 +156,10 @@ def create_trace_span_helper(client, bucket_name, name, attributes=None, **kwarg
         and client
         and hasattr(client, "_bucket_metadata_cache")
         and client._bucket_metadata_cache
+        and not (
+            os.environ.get("DISABLE_BUCKET_MD_IN_OTEL", "").lower()
```
kalragauri: Is this change related to the feature guarded by `ENABLE_GCS_PYTHON_CLIENT_OTEL_TRACES`? If so, should we use consistent naming?

kalragauri (approved):

packages/google-cloud-storage/google/cloud/storage/_helpers.py
chandra-siri: done!

packages/google-cloud-storage/google/cloud/storage/_helpers.py
chandra-siri: I kept consistent with what already existing for Otel in python - https://github.com/googleapis/google-cloud-python/blob/d01a4ba30040cfcb6498d0e9ef3ed3a54d56239d/packages/google-cloud-storage/google/cloud/storage/_opentelemetry_tracing.py#L36 

the way it was checked lacked readablity. I've now made it a bit more readable.

packages/google-cloud-storage/google/cloud/storage/_http.py
chandra-siri: done

kalragauri (approved):

PR #17261, opened 2026-05-26
feat(storage): Enable full object checksum PR 1/3 : parse finalize_time and server crc32c in async object stream

chandra-siri:
### 1. Overview of the Solution
This solution implements end-to-end full-object checksum validation in `AsyncMultiRangeDownloader` for the asynchronous Google Cloud Storage Python client library. As asynchronous multiplexed downloads of non-contiguous ranges are performed concurrently over a single bidirectional gRPC connection, this feature automatically and incrementally calculates a rolling checksum as bytes arrive and validates it against the server's authoritative object checksum once the download completes.

The technical approach consists of three coordinated layers:
*   **`_AsyncReadObjectStream` (Stream Ingestion)**: Safely extracts the authoritative server checksum (`full_obj_server_crc32c`) and finalization status (`is_finalized`) from the object metadata received in the first data payload response of the stream.
*   **`_ReadResumptionStrategy` & `_DownloadState` (Verification Logic)**: Computes an isolated, persistent rolling checksum in the individual `_DownloadState` object to ensure calculations do not bleed across concurrent multiplexed ranges. Crucially, the rolling hash updates only *after* buffer writes succeed to prevent state corruption during retry re-connects, raising a `DataCorruption` exception on completion if a mismatch occurs.
*   **`AsyncMultiRangeDownloader` (Orchestration & Cleanup)**: Detects candidate full-object ranges (e.g., `(0, 0)` or `(0, persisted_size)`), propagates checksum settings to the resumption strategy, and guarantees robust cleanup (closing the stream immediately and unregistering IDs) if data corruption or write errors occur.

### 2. What This PR Specifically Does
This PR implements **Step 1: Stream Metadata Ingestion** of the solution:
*   Modifies `_AsyncReadObjectStream` to safely parse GCS object metadata from the first data payload of the response.
*   Populates `is_finalized`, `full_obj_server_crc32c`, and `object_metadata` attributes in `_AsyncReadObjectStream.open()`.
*   Adds an autouse pytest event loop fixture in `tests/unit/conftest.py` to resolve compatibility issues with `pytest-asyncio` under Python 3.11+.
*   Adds unit tests in `test_async_read_object_stream.py` to verify that finalization status and server-authoritative checksums are correctly extracted or skipped for unfinalized objects.


packages/google-cloud-storage/google/cloud/storage/asyncio/async_read_object_stream.py
```diff
@@ -132,6 +135,18 @@ async def open(self, metadata: Optional[List[Tuple[str, str]]] = None) -> None:
                 self.generation_number = response.metadata.generation
             # update persisted size
             self.persisted_size = response.metadata.size
+            self.object_metadata = response.metadata
+            if (
+                hasattr(response.metadata, "finalize_time")
+                and response.metadata.finalize_time
+                and response.metadata.finalize_time.second > 0
```
kalragauri: is it possible that this is 0? What happens if the object gets finalized exactly on a minute boundary?

packages/google-cloud-storage/google/cloud/storage/asyncio/async_read_object_stream.py
```diff
@@ -79,6 +79,9 @@ def __init__(
         self.socket_like_rpc: Optional[AsyncBidiRpc] = None
         self._is_stream_open: bool = False
         self.persisted_size: Optional[int] = None
+        self.is_finalized: bool = False
+        self.full_obj_server_crc32c: Optional[int] = None
```
kalragauri, line 83: Should these be private similar to _is_stream_open?

kalragauri (approved):

packages/google-cloud-storage/google/cloud/storage/asyncio/async_read_object_stream.py
chandra-siri: second here represents the epoch seconds- http://shortn/_Xi0rk0QrQt

packages/google-cloud-storage/google/cloud/storage/asyncio/async_read_object_stream.py
chandra-siri: no, we need it in subsequent PRs - https://github.com/googleapis/google-cloud-python/pull/17263/changes , see L332 of `async_multi_range_downloader.py`

PR #17320, opened 2026-05-29
chore(bigframes): Add simple Python bytecode translation

TrevorBergeron:

Towards: issue 521549179 🦕

### DIFF
--- a/packages/bigframes/bigframes/core/bytecode.py
+++ b/packages/bigframes/bigframes/core/bytecode.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/packages/bigframes/bigframes/core/bytecode.py
+++ b/packages/bigframes/bigframes/core/bytecode.py
@@ -44,6 +44,23 @@ _COMPARE_OP_MAP = {
     ">=": operator.ge,
 }
 
+_OLD_BINARY_OP_MAP = {
+    "BINARY_ADD": operator.add,
+    "INPLACE_ADD": operator.add,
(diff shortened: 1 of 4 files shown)


packages/bigframes/bigframes/core/bytecode.py
```diff
@@ -0,0 +1,263 @@
+# Copyright 2025 Google LLC
```
GarrettWu: 2026

packages/bigframes/bigframes/core/py_expressions.py
```diff
@@ -0,0 +1,366 @@
+# Copyright 2023 Google LLC
```
GarrettWu: 2026

packages/bigframes/tests/unit/core/test_bytecode.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2025 Google LLC
```
GarrettWu: 2026

GarrettWu:
What are we trying to do with python bytecode? Any context?

packages/bigframes/bigframes/core/bytecode.py
TrevorBergeron: updated

packages/bigframes/bigframes/core/py_expressions.py
TrevorBergeron: this file is substantially based on my prototype from last year, so updating to 2025 instead of this year

packages/bigframes/tests/unit/core/test_bytecode.py
TrevorBergeron: updated

TrevorBergeron:
> What are we trying to do with python bytecode? Any context?

Created internal issue: 521549179 to track this work

GarrettWu (approved):

PR #17330, opened 2026-06-01
feat: support row_range in sample_row_keys method

### LINKED ISSUE #17329: Implement specifying a row_range for sample_row_keys()

Limit the sample_row_keys call to a subset of the table, by specifying a row_range.

### Expected results

_No response_

_No response_

### Use case

Useful when running sample_row_keys on very large tables.

### Additional context

_No response_

anvitTawar:

Fixes #17329


mutianf (approved):

packages/google-cloud-bigtable/google/cloud/bigtable/table.py
```diff
@@ -739,7 +739,7 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY, timeout=DEFAULT):
         )
         return retryable_mutate_rows(retry=retry)
 
-    def sample_row_keys(self):
+    def sample_row_keys(self, row_range=None):
```
mutianf: I think we can revert the changes in this file. This is the legacy client. Same with client_handler_legacy.

anvitTawar: Done.

daniel-sanche:
Thanks for putting this together

1. Can you add a system test, [like this?](https://github.com/googleapis/google-cloud-python/blob/761c8052198f8db98f7d747d6db9a1b9ec875668/packages/google-cloud-bigtable/tests/system/data/test_system_async.py#L330)
2. I don't think changing the conformance tests should be required for this test. Were the tests failing before your changes?
3. Let me know if you need help getting the other tests passing

anvitTawar:
> 1. Can you add a system test, [like this?](https://github.com/googleapis/google-cloud-python/blob/761c8052198f8db98f7d747d6db9a1b9ec875668/packages/google-cloud-bigtable/tests/system/data/test_system_async.py#L330)

Done.

> 2. I don't think changing the conformance tests should be required for this test. Were the tests failing before your changes?

The tests were failing because the API proto change used by this feature wasn't published yet. Now that that's live, the tests have all passed.

PR #17331, opened 2026-06-01
chore: remove prerelease and unit tests for goldens

ohmayr:
We are defining a purpose for each of our testing type. Goldens should verify that the generated code is syntactically valid Python code.

These tests are run downstream so having them here is redundant. In the long run, they will be removed from downstream as well and will be run for showcase and a list of canaries upstream only.

Changes introduced:

Deleted the goldens-prerelease job: Removed `prerelease_deps` and `core_deps_from_source` from the PR blocking path.

Stripped unit execution from the goldens job: Removed the `unit-${{ ... }}` Nox session. The goldens job now strictly runs static structural checks (format, lint, and pylint).

### DIFF
--- a/.github/workflows/gapic-generator-tests.yml
+++ b/.github/workflows/gapic-generator-tests.yml
@@ -113,7 +113,7 @@ jobs:
           pip install nox
           cd packages/gapic-generator
           for pkg in credentials eventarc logging redis; do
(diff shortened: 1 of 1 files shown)


parthea:
I'm going to close this PR due to no activity in the last 2 months. Please feel free to re-open this PR or create a new PR.

PR #17344, opened 2026-06-02
test(spanner): add pytest-xdist parallel execution with state isolation

chalmerlowe:
This PR adds `pytest-xdist` to parallelize unit tests and the `core_deps_from_source` nox sessions for the `google-cloud-spanner` package.

**By running tests in parallel using `-n auto`, the execution time of the Spanner unit tests are reduced from ~8 minutes to 4 minutes.**

State isolation and test reliability are achieved by:

* **Simplifying Subtests**: We pass simple strings/names to `subTest()` instead of complex objects. This keeps subtests lightweight and prevents serialization errors.
* **Cleaning Up Global Singletons**: We reset telemetry singleton states on test teardown (using pytest's idiomatic `monkeypatch` fixture). This ensures metric counters don't leak from one test into another.
* **Fixing Concurrent Mock Conflicts**: We return fresh mocked iterators for concurrent calls (using `side_effect` instead of `return_value`). This prevents one thread from exhausting a mock's results before another thread can read them.
* **Robust Assertion Checks**: We added a helper method (`_assert_concurrent_transaction_invariants`) that verifies the *behavior* of concurrent threads (ensuring exactly one thread starts the transaction while others wait/reuse it), rather than checking fragile call logs or counting sequential request IDs. This allowed us to safely run four concurrent tests.

> [!note]
> The long pole in the tent is still the `system` tests, which require about 30 minutes. It is not as simple as just adding xdist because there are other factors that limit velocity including the fact that system tests actually interact with live systems.

### DIFF
--- a/packages/google-cloud-spanner/noxfile.py
+++ b/packages/google-cloud-spanner/noxfile.py
@@ -230,6 +230,7 @@ def unit(session, protobuf_implementation):
         CURRENT_DIRECTORY / "testing" / f"constraints-{session.python}.txt"
     )
     install_unittest_dependencies(session, "-c", constraints_path)
+    session.install("pytest-xdist")
 
     # TODO(https://github.com/googleapis/synthtool/issues/1976):
     # Remove the 'cpp' implementation once support for Protobuf 3.x is dropped.
@@ -240,6 +241,8 @@ def unit(session, protobuf_implementation):
     # Run py.test against the unit tests.
     args = [
         "py.test",
+        "-n",
+        "auto",
         "-s",
         f"--junitxml=unit_{session.python}_sponge_log.xml",
         "--cov=google",
@@ -753,6 +756,7 @@ def prerelease_deps(session, protobuf_implementation, database_dialect):
 @nox.session(python=ALL_PYTHON)
 def mypy(session):
     """Run the type checker."""
+    session.skip("Mypy is not yet supported")
     # TODO(https://github.com/googleapis/gapic-generator-python/issues/2579):
     # use the latest version of mypy
     session.install(
@@ -830,12 +834,15 @@ def core_deps_from_source(session, protobuf_implementation):
     dep_paths = [str(deps_dir / dep) for dep in core_dependencies_from_source]
 
     session.install(*dep_paths, "--no-deps", "--ignore-installed")
+    session.install("pytest-xdist")
     print(
         f"Installed {', '.join(core_dependencies_from_source)} locally from {deps_dir}"
     )
 
     session.run(
         "py.test",
+        "-n",
+        "auto",
         "tests/unit",
         env={
             "PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION": protobuf_implementation,
--- a/packages/google-cloud-spanner/noxfile.py
+++ b/packages/google-cloud-spanner/noxfile.py
@@ -43,6 +43,7 @@ UNIT_TEST_STANDARD_DEPENDENCIES = [
     "pytest",
     "pytest-cov",
     "pytest-asyncio",
+    "pytest-xdist",
 ]
 MOCK_SERVER_ADDITIONAL_DEPENDENCIES = [
     "google-cloud-testutils",
@@ -230,7 +231,6 @@ def unit(session, protobuf_implementation):
         CURRENT_DIRECTORY / "testing" / f"constraints-{session.python}.txt"
     )
     install_unittest_dependencies(session, "-c", constraints_path)
-    session.install("pytest-xdist")
 
     # TODO(https://github.com/googleapis/synthtool/issues/1976):
     # Remove the 'cpp' implementation once support for Protobuf 3.x is dropped.
--- a/packages/google-cloud-spanner/noxfile.py
+++ b/packages/google-cloud-spanner/noxfile.py
@@ -753,8 +753,6 @@ def prerelease_deps(session, protobuf_implementation, database_dialect):
 @nox.session(python=ALL_PYTHON)
 def mypy(session):
     """Run the type checker."""
-    session.skip("Mypy is not yet supported")
-
     # TODO(https://github.com/googleapis/gapic-generator-python/issues/2579):
     # use the latest version of mypy
     session.install(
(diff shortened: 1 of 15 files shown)


packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -135,6 +135,30 @@ def _make_spanner_api(self):
 
         return mock.create_autospec(SpannerClient, instance=True)
 
+    def _assert_concurrent_transaction_invariants(self, call_args_list, expected_count=2):
```
chalmerlowe: For the reviewer:
This new function is a helper that enables us to confirm test results with more stability in concurrently run systems by checking for elements that won't change during a concurrent transation rather than elements that might change depending on the order that a concurrent set of tests finishes.

There are four locations where we strip out some long test bodies that were very specific (e.g. did test 1.1 come before 2.1, etc) and replace them with:

Is there a single "begin" transaction. 
Do other transactions reuse the same transaction ID throughout.

packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -986,49 +1014,7 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
 
         self._batch_update_helper(transaction=transaction, database=database, api=api)
 
-        api.execute_sql.assert_any_call(
```
chalmerlowe, line 989: As noted above, here we are replacing the test body that was looking for specific id numbers (and became brittle when run in a concurrent transaction).

Replaced it with the more invariant test body from the helper function: `_assert_concurrent_transaction_invariants`

packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -1060,47 +1046,8 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
         self._execute_update_helper(transaction=transaction, api=api)
         self.assertEqual(api.execute_sql.call_count, 1)
 
-        api.execute_sql.assert_any_call(
```
chalmerlowe, line 1063: Replaced the test body with the more invariant test body: `_assert_concurrent_transaction_invariants`

packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -1060,47 +1046,8 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
...
-                ("x-goog-spanner-route-to-leader", "true"),
-                (
-                    "x-goog-spanner-request-id",
-                    f"1.{REQ_RAND_PROCESS_ID}.{database._nth_client_id}.{database._channel_id}.3.1",
-                ),
-            ],
-        )
+        self._assert_concurrent_transaction_invariants(api.execute_batch_dml.call_args_list, 2)
 
-        self.assertEqual(api.execute_batch_dml.call_count, 2)
```
chalmerlowe, line 1077: Replaced the test body with the more invariant test body: `_assert_concurrent_transaction_invariants`

packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -1130,55 +1077,9 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
 
         self._execute_update_helper(transaction=transaction, api=api)
 
-        api.execute_sql.assert_any_call(
```
chalmerlowe, line 1133: Replaced the test body with the more invariant test body: `_assert_concurrent_transaction_invariants`

packages/google-cloud-spanner/tests/unit/test_spanner.py
```diff
@@ -1207,59 +1108,8 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
 
         self._execute_update_helper(transaction=transaction, api=api)
 
-        begin_read_write_count = sum(
```
chalmerlowe, line 1210: Replaced the test body with the more invariant test body: `_assert_concurrent_transaction_invariants`

packages/google-cloud-spanner/tests/unit/_async/test_session.py
```diff
@@ -1800,11 +1800,18 @@ async def unit_of_work(txn, *args, **kw):
             called_with.append((txn, args, kw))
             txn.insert(TABLE_NAME, COLUMNS, VALUES)
 
+        import threading
```
daniel-sanche: can this be moved to to top of the file?

packages/google-cloud-spanner/tests/unit/_async/test_session.py
```diff
@@ -1877,9 +1884,16 @@ async def unit_of_work(txn, *args, **kw):
...
```
daniel-sanche: if we import threading globally, this shouldnt be needed

packages/google-cloud-spanner/tests/unit/test_session.py
```diff
@@ -1783,9 +1792,18 @@ def unit_of_work(txn, *args, **kw):
             called_with.append((txn, args, kw))
             txn.insert(TABLE_NAME, COLUMNS, VALUES)
 
+        import threading
```
daniel-sanche: same comments about imports

packages/google-cloud-spanner/noxfile.py
```diff
@@ -230,6 +230,7 @@ def unit(session, protobuf_implementation):
         CURRENT_DIRECTORY / "testing" / f"constraints-{session.python}.txt"
     )
     install_unittest_dependencies(session, "-c", constraints_path)
+    session.install("pytest-xdist")
```
daniel-sanche: Can this be added to UNIT_TEST_STANDARD_DEPENDENCIES?

daniel-sanche (commented):
couple small comments, otherwise LGTM

packages/google-cloud-spanner/tests/unit/_async/test_session.py
chalmerlowe: Done

packages/google-cloud-spanner/tests/unit/_async/test_session.py
chalmerlowe: Done

packages/google-cloud-spanner/tests/unit/test_session.py
chalmerlowe: Done

packages/google-cloud-spanner/noxfile.py
chalmerlowe: Done

daniel-sanche (approved):
LGTM

PR #17346, opened 2026-06-02
fix(bigframes): improve error message when unescaped `{` are found in SQL cells

tswast:
Hints to the user that they may need to escape `{` and `}` characters by doubling them, and includes context as to where to correct such errors.

Fixes internal issue b/517909919
🦕


packages/bigframes/bigframes/core/pyformat.py
```diff
@@ -185,13 +336,29 @@ def pyformat(
...
+        except KeyError as e:
+            positions = _find_all_field_positions(sql_template)
+            occurrence_idx = seen_counts[name] - 1
+            pos = positions.get((name, occurrence_idx), -1)
+            context = get_error_context_at_pos(sql_template, pos)
+            raise ValueError(
+                f"Undetected variable {name!r} in SQL template. "
+                "Did you mean to escape '{' and '}' by doubling them?\n"
+                f"{context}"
+            ) from e
```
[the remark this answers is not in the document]
tswast: Since we don't expect this case to happen, I would prefer not to add logic for it.

packages/bigframes/tests/unit/core/test_pyformat.py
```diff
@@ -70,13 +136,67 @@ def test_pyformat_with_unsupported_type_raises_typeerror(session):
...
+    assert "expected '}' before end of string" in err_msg_1
+
+    # Case 2: Single '}' (unmatched)
+    sql_2 = "SELECT foo}"
+    with pytest.raises(ValueError) as exc_info:
+        pyformat.pyformat(sql_2, pyformat_args=pyformat_args, session=session)
+    err_msg_2 = str(exc_info.value)
+    assert "Failed to parse SQL template" in err_msg_2
+    assert "Did you mean to escape '{' and '}'" in err_msg_2
+    assert "Single '}' encountered in format string" in err_msg_2
```
sycai: Shall we parameterize these two test cases?

tswast: Good call. I like it. Done.

sycai (approved):

PR #17362, opened 2026-06-03
fix(bigframes): include pyopenssl as a dependency

tswast:
Also, support pandas 3.0 in various system tests.

Internal issue b/519591816 🦕

### DIFF
--- a/packages/bigframes/tests/unit/core/compile/sqlglot/test_dataframe_accessor.py
+++ b/packages/bigframes/tests/unit/core/compile/sqlglot/test_dataframe_accessor.py
@@ -42,7 +42,7 @@ def test_sql_scalar(scalar_types_df: bpd.DataFrame, snapshot, monkeypatch):
     )
 
     session.read_pandas.assert_called_once()
-    snapshot.assert_match(result, "out.sql")
+    snapshot.assert_match(result.strip() + "\n", "out.sql")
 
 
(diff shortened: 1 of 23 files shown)


packages/bigframes/tests/unit/core/compile/sqlglot/test_dataframe_accessor.py
```diff
@@ -42,7 +46,7 @@ def to_pandas(series, *, ordered):
     )
 
     session.read_pandas.assert_called_once()
-    snapshot.assert_match(result, "out.sql")
+    snapshot.assert_match(result.strip() + "\n", "out.sql")
```
TrevorBergeron, line 49: Would this issue impact every snapshot? I assume this is from mismatch between sqlglot version locally vs on the automated test runs?

TrevorBergeron (approved):

packages/bigframes/tests/unit/core/compile/sqlglot/test_dataframe_accessor.py
tswast: It's actually not a version mismatch. It's because I was fighting the `pre-commit` end-of-file newline "fixer". Basically, I'm applying that change here so that when we re-run it, the "fixer" won't see it as a problem because it's looking for exactly 1 trailing newline character.

PR #17366, opened 2026-06-03
feat(google/developers/knowledge/v1): add google-developers-knowledge

suztomo:
b/503382870


librarian.yaml
```diff
@@ -2218,6 +2219,17 @@ libraries:
...
+  - name: google-developers-knowledge
+    version: 0.0.0
+    apis:
+      - path: google/developers/knowledge/v1
+    copyright_year: "2026"
+    python:
+      opt_args_by_api:
+        google/developers/knowledge/v1:
+          - python-gapic-namespace=google
+          - python-gapic-name=developer_knowledge
```
parthea, line 2231: I believe there is a typo here. We have prefix `developer` but it should be `developers` to match the directory `developers` which is part of the path `google/developers/knowledge/v1`. (Googlers see b/503382870 where I added this feedback)

```suggestion
          - python-gapic-name=developers_knowledge
```

suztomo:
Ack > `python-gapic-name=developers_knowledge`

PR #17367, opened 2026-06-03
feat(bigframes): support offset-based column access via iloc

tswast:
🦕


packages/bigframes/bigframes/dataframe.py
```diff
@@ -821,22 +820,25 @@ def __repr__(self) -> str:
...
+            for col_index, col in enumerate(df.columns)
             if bigframes.dtypes.contains_db_dtypes_json_dtype(df[col].dtype)
         ]
-        if json_cols:
-            op = ops.SqlScalarOp(
-                _output_type=bigframes.dtypes.STRING_DTYPE,
-                sql_template="TO_JSON_STRING({0})",
+        if json_col_indexes:
+            df._block.apply_analytic
+            df.iloc[:, json_col_indexes] = cast(
```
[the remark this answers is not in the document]
tswast: Done.

packages/bigframes/bigframes/core/indexers.py
```diff
@@ -102,6 +103,18 @@ def __getitem__(
...
         Other key types are not yet supported.
         """
+        requires_ordering = True
+        if (
+            isinstance(key, slice)
+            and (key.start is None or key.start == 0)
+            and (key.step is None or key.step == 1)
+            and key.stop is None
+        ):
+            requires_ordering = False
```
TrevorBergeron: Maybe we extract this out to a helper? The idea being we want `iloc[:]` to work without ordering?

packages/bigframes/bigframes/dataframe.py
```diff
@@ -2264,35 +2301,68 @@ def _assign_multi_items(
...
+        k: Sequence[int],
+        v: SingleItemValue | MultiItemValue,
+    ) -> DataFrame:
+        return self._assign_multi_items_helper(
+            k, v, DataFrame._assign_single_item_by_offset
+        )
+
+    _assign_multi_items_by_offset = _assign_multi_items_by_offsets
+    _assign_multi_items_by_label = _assign_multi_items
+    _assign_multi_items_by_labels = _assign_multi_items
```
TrevorBergeron, line 2327: why do we need these aliases?

packages/bigframes/bigframes/core/indexers.py
```diff
@@ -244,8 +257,110 @@ def __getitem__(self, key) -> Union[bigframes.dataframe.DataFrame, pd.Series]:
...
+            if not col_offsets:
+                return
+            df = self._dataframe._assign_multi_items_by_offsets(col_offsets, value)
+            self._dataframe._set_block(df._get_block())
+
+        elif pd.api.types.is_list_like(col_indexer):
+            col_indexer_list = list(col_indexer)
+
+            if len(col_indexer_list) > 0 and all(
+                isinstance(x, bool) for x in col_indexer_list
```
TrevorBergeron: Might we need something more general than isintance to handle eg numpy bool?

TrevorBergeron: Also similar concern with other isinstance checks elsewhere

packages/bigframes/bigframes/core/indexers.py
```diff
@@ -244,8 +257,110 @@ def __getitem__(self, key) -> Union[bigframes.dataframe.DataFrame, pd.Series]:
...
+                    if not isinstance(idx, int):
+                        raise TypeError("pos list must contain only integers")
+                    if idx < 0:
+                        idx += n_cols
+                    if idx < 0 or idx >= n_cols:
+                        raise IndexError("positional indexer is out-of-bounds")
+                    col_offsets.append(idx)
+
+            if not col_offsets:
+                return
```
TrevorBergeron: Seems all of this is about essentially converting a key to an offset list. and then doing the same `_assign_multi_items_by_offsets` and then  `_set_block` sequence given that.

packages/bigframes/bigframes/dataframe.py
tswast: Not strictly necessary, but kinda nice to be able to be clear about the usage. Alternatively, I could remove `_assign_multi_items` and just rename it to `_assign_multi_items_by_labels`. What do you think?

packages/bigframes/bigframes/core/indexers.py
tswast: Good point. I might be able to use the typing.abc classes instead. I'll check on this.

packages/bigframes/bigframes/core/indexers.py
tswast: Yes, helper makes sense to me. Might be able to keep it from getting out of sync that way.

Yes, that's the idea. If they select all rows, we don't care about the ordering.

packages/bigframes/bigframes/core/indexers.py
tswast: That's a good callout. I added `_iloc_col_indexer_to_offsets` and `_iloc_clip_to_offset` helper functions along these lines. I also realized that getitem is sharing quite a bit of logic here too and updated it to use the same helpers.

packages/bigframes/bigframes/core/indexers.py
```diff
@@ -426,6 +475,74 @@ def _struct_accessor_check_and_warn(
...
+
+    if offset < 0 or offset >= length:
+        raise IndexError(f"{name} {index} is out-of-bounds")
+
+    return index
+
+
+def _iloc_col_indexer_to_offsets(
+    df: bigframes.dataframe.DataFrame, col_indexer: Any
+) -> Tuple[Sequence[int], str]:
```
tswast: Probably better to put the return type logic in the getitem implementation even if it means a little bit of duplicate code to determine if `col_indexer` is a scalar integer.

tswast: Done in latest commit.

packages/bigframes/bigframes/core/indexers.py
tswast: Fixed in the latest commits. Added support for pyarrow scalars too.

packages/bigframes/bigframes/core/local_data.py
```diff
@@ -521,3 +531,77 @@ def _schema_durations_to_ints(schema: pa.Schema) -> pa.Schema:
...
+            _has_extension_type(struct_type.field(i).type)
+            for i in range(struct_type.num_fields)
+        )
+    if pa.types.is_list(pa_type) or pa.types.is_large_list(pa_type):
+        list_type = cast(pa.ListType, pa_type)
+        return _has_extension_type(list_type.value_type)
+    return False
+
+
+def pyarrow_array_from_sequence(data: Any, pa_type: pa.DataType) -> pa.Array:
```
TrevorBergeron: q: can we not just use the `ManagedArrowTable.from_pyarrow` to ingest this data? AFAICT we are using raw string datatype for json in our representation, is this not working to dodge the extension type issue?

tswast: The trouble is that we are trying to use the pandas DataFrame / Series constructors at

https://github.com/googleapis/google-cloud-python/blob/a5a717df2434b756c9043cd53d7cef15c5a4b552/packages/bigframes/bigframes/dataframe.py#L227-L232

and

https://github.com/googleapis/google-cloud-python/blob/a5a717df2434b756c9043cd53d7cef15c5a4b552/packages/bigframes/bigframes/series.py#L198-L203

It's pandas that's calling some pyarrow code path that is failing on struct with a json extension field.

Given that will have potentially greater implications, I'll try that out in a separate PR.

tswast: Thought: I can just update the unit test to pass in a pyarrow object directly to avoid that code path, I think.

tswast: Done in latest commit.

TrevorBergeron (approved):

PR #17375, opened 2026-06-04
perf(spanner): optimize query result decoding

olavloite:
Work in progress.

Optimizes the decoding and reading of (large) result sets for Spanner.

image

### DIFF
--- a/packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
+++ b/packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
@@ -129,16 +129,35 @@ class StreamedResultSet(object):
         decoders = self._decoders
         width = len(self.fields)
         index = len(self._current_row)
-        for value in values:
-            if self._lazy_decode:
-                self._current_row.append(value)
-            else:
-                self._current_row.append(_parse_nullable(value, decoders[index]))
-            index += 1
-            if index == width:
-                self._rows.append(self._current_row)
-                self._current_row = []
-                index = 0
+        current_row = self._current_row
+        rows = self._rows
+
+        current_row_append = current_row.append
+        rows_append = rows.append
+
+        if self._lazy_decode:
+            for value in values:
+                current_row_append(value)
+                index += 1
+                if index == width:
+                    rows_append(current_row)
+                    current_row = []
+                    current_row_append = current_row.append
+                    index = 0
+        else:
+            for value in values:
+                if value.HasField("null_value"):
+                    current_row_append(None)
+                else:
+                    current_row_append(decoders[index](value))
+                index += 1
+                if index == width:
+                    rows_append(current_row)
+                    current_row = []
+                    current_row_append = current_row.append
+                    index = 0
+
+        self._current_row = current_row
 
     @CrossSync.convert
     async def _consume_next(self):
--- a/packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
+++ b/packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
@@ -146,6 +146,10 @@ class StreamedResultSet(object):
                     index = 0
         else:
             for value in values:
+                # Note: We manually check value.HasField("null_value") here instead of
+                # wrapping every decoder in _parse_nullable to avoid the overhead of
+                # an extra Python function call layer for every cell value decoded in this loop.
+                # If the nullable check logic is updated in _parse_nullable, update this check.
                 if value.HasField("null_value"):
                     current_row_append(None)
                 else:
--- a/packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
+++ b/packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
@@ -466,6 +466,12 @@ def _parse_value_pb(value_pb, field_type, field_name, column_info=None):
     return _parse_nullable(value_pb, decoder)
 
 
+_date_fromisoformat = datetime.date.fromisoformat
+_Decimal = decimal.Decimal
+_json_from_str = JsonObject.from_str
+_uuid_UUID = uuid.UUID
+
+
 def _get_type_decoder(field_type, field_name, column_info=None):
     """Returns a function that converts a Value protobuf to cell data.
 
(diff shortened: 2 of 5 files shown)


packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -19,14 +19,15 @@
 import decimal
 import logging
 import math
+import operator
+import re
```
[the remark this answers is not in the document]
olavloite: Except that exact those imports are at line 17 and 18 of this file.... So... No, this is nonsense.

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -577,12 +578,40 @@ def _parse_float(value_pb) -> float:
...
+    return datetime_helpers.DatetimeWithNanoseconds(
+        bare.year,
+        bare.month,
+        bare.day,
+        bare.hour,
+        bare.minute,
+        bare.second,
+        nanosecond=nanos,
+        tzinfo=datetime.timezone.utc,
+    )
```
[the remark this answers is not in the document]
olavloite: Fixed, and added a bunch of additional tests to verify this.

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -490,13 +491,13 @@ def _get_type_decoder(field_type, field_name, column_info=None):
 
     type_code = field_type.code
     if type_code == TypeCode.STRING:
-        return _parse_string
+        return operator.attrgetter("string_value")
     elif type_code == TypeCode.BYTES:
-        return _parse_bytes
+        return lambda value_pb: value_pb.string_value.encode("utf8")
```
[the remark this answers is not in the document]
olavloite: No, this client is actually expected to return the utf-8 base64 string as-is. See https://github.com/googleapis/google-cloud-python/blob/65f059e22ea1d710e06230cf5f6ee9eb5fe45e8e/packages/google-cloud-spanner/tests/system/test_session_api.py#L53

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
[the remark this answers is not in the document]
olavloite: Done

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -490,27 +497,35 @@ def _get_type_decoder(field_type, field_name, column_info=None):
...
+            if value_pb.HasField("string_value")
+            else value_pb.number_value
+        )
     elif type_code == TypeCode.FLOAT32:
-        return _parse_float
+        return (
+            lambda value_pb: float(value_pb.string_value)
+            if value_pb.HasField("string_value")
+            else value_pb.number_value
+        )
```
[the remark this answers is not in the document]
olavloite: This is wrong. The evaluation is lazy, so as long as `value_pb.HasField("string_value")` returns false, the fallback path is never evaluated. The two implementations are exactly the same, and perform exactly the same.

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -553,48 +568,49 @@ def _parse_list_value_pbs(rows, row_type):
...
+    return datetime_helpers.DatetimeWithNanoseconds(
+        bare.year,
+        bare.month,
+        bare.day,
+        bare.hour,
+        bare.minute,
+        bare.second,
+        nanosecond=nanos,
+        tzinfo=datetime.timezone.utc,
+    )
```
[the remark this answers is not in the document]
olavloite: Done

olavloite:
The time needed to decode 100,000 rows is significantly reduced from approx 6-7s to approx 3-3.5s. The increasing gap between the two lines is due to the nature of this benchmark. The benchmark simulates a randomized poisson load. A small burst in traffic is enough to completely overload the old implementation, and cause requests to queue up and start taking an exponentially increasing time.

image

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -553,48 +567,73 @@ def _parse_list_value_pbs(rows, row_type):
...
+    val = value_pb.string_value
+    try:
+        if len(val) < 20 or val[10] != "T":
+            raise ValueError()
+        no_fraction = val[:19]
+        bare = datetime.datetime.fromisoformat(no_fraction)
+        if val[19] == ".":
+            if val.endswith("Z"):
+                offset = "Z"
+                fraction = val[20:-1]
```
rahul2393, line 596: Spanner [Timestamp](https://docs.cloud.google.com/spanner/docs/reference/standard-sql/timestamp_functions#timestamp) seems to  be always UTC Z at the API/wire level. Non-Z offset only come from CAST(ts as String), which is a STRING column and routes through the STRING decoder.

+/-HH:MM handling plus the astimezone call won't ever be executed on real TIMESTAMP value? Can we drop it to shrink the risky code, or comment marking it deliberately defensive. The Z fast-path is the legitimate win and short-circuits before this anyway?

packages/google-cloud-spanner/tests/unit/test__helpers.py
```diff
@@ -653,6 +653,96 @@ def test_w_timestamp_w_nanos(self):
...
+        field_type = Type(code=TypeCode.TIMESTAMP)
+        field_name = "timestamp_column"
+
+        expected = datetime_helpers.DatetimeWithNanoseconds(
+            2016, 12, 20, 11, 13, 47, nanosecond=123456789, tzinfo=timezone.utc
+        )
+
+        parsed = self._callFUT(value_pb, field_type, field_name)
+        self.assertIsInstance(parsed, datetime_helpers.DatetimeWithNanoseconds)
+        self.assertEqual(parsed, expected)
```
rahul2393, line 674: DatetimeWithNanoseconds.__eq__ is inherited from datetime and compares only to microsecond - it ignores .nanosecond. So this passes even if sub-us digits are wrong.

packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
```diff
@@ -129,16 +129,35 @@ def _merge_values(self, values):
...
+                current_row_append(value)
+                index += 1
+                if index == width:
+                    rows_append(current_row)
+                    current_row = []
+                    current_row_append = current_row.append
+                    index = 0
+        else:
+            for value in values:
+                if value.HasField("null_value"):
```
rahul2393, line 153: value.HasField("null_value") duplicates _parse_nullables's semantics(still used in lazy/array/single-value paths). Should we add comment here to keep in sync so duplication doesn't silently drift?

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
```diff
@@ -490,27 +496,35 @@ def _get_type_decoder(field_type, field_name, column_info=None):
...
+        return operator.attrgetter("string_value")
     elif type_code == TypeCode.BYTES:
-        return _parse_bytes
+        return lambda value_pb: value_pb.string_value.encode("utf8")
     elif type_code == TypeCode.BOOL:
-        return _parse_bool
+        return operator.attrgetter("bool_value")
     elif type_code == TypeCode.INT64:
-        return _parse_int64
+        return lambda value_pb: int(value_pb.string_value)
```
rahul2393, line 507: Maybe a comment on why only STRING/BOOL use attrgetter(faster) while transform trypes use lambdas?

rahul2393 (approved):
LGTM - with minor comments/questions and test assert fix

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
olavloite: I agree that Spanner (currently) always sends timestamps in Zulu time. BUT: It is not strictly part of the API contract, meaning that future feature changes in Spanner could (in theory) trigger timestamps to be sent with an offset. Also, the previous implementation also supported offsets, so I included this in order not to cause any potential regressions from the previous implementation.

I added an additional test for parsing timestamps with an offset (in this case one without a sub-second fraction).

packages/google-cloud-spanner/tests/unit/test__helpers.py
olavloite: Oh, great catch, thanks! I updated the various checks in all the tests in this file, and also added a test that would otherwise have reported a false positive.

packages/google-cloud-spanner/google/cloud/spanner_v1/_helpers.py
olavloite: Added a comment to explain.

packages/google-cloud-spanner/google/cloud/spanner_v1/_async/streamed.py
olavloite: Added a comment to explain both why it is there, and that it needs to be kept in sync.

PR #17387, opened 2026-06-06
fix(auth): Agentic Identites mTLS gaps fix _is_mtls and SslCredentials.

vverman:
Resolve some mTLS transport state and gRPC workload certificate issues

- Fix inconsistent `_is_mtls` flag in `urllib3` on fallback to default HTTP.

- Reset `_is_mtls` to `False` in async `sessions` if a custom request handler cannot be reconfigured.

- Update gRPC `SslCredentials` to recognize workload certificates by checking `has_default_client_cert_source` instead of only SecureConnect.

### DIFF
--- a/packages/google-auth/google/auth/aio/transport/sessions.py
+++ b/packages/google-auth/google/auth/aio/transport/sessions.py
@@ -219,11 +219,7 @@ class AsyncAuthorizedSession:
                                 UserWarning,
                             )
 
-                except (
-                    exceptions.ClientCertError,
-                    ImportError,
-                    OSError,
-                ) as caught_exc:
+                except Exception as caught_exc:
                     self._is_mtls = False
                     new_exc = exceptions.MutualTLSChannelError(caught_exc)
                     raise new_exc from caught_exc
@@ -585,4 +581,10 @@ class AsyncAuthorizedSession:
         """
         Close the underlying auth request session.
         """
+        if self._mtls_init_task and not self._mtls_init_task.done():
+            self._mtls_init_task.cancel()
+            try:
+                await self._mtls_init_task
+            except asyncio.CancelledError:
+                pass
         await self._auth_request.close()
--- a/packages/google-auth/google/auth/aio/transport/sessions.py
+++ b/packages/google-auth/google/auth/aio/transport/sessions.py
@@ -156,7 +156,6 @@ class AsyncAuthorizedSession:
         (via GOOGLE_API_USE_CLIENT_CERTIFICATE=true) or auto-enabled (when the env
         variable is unset and workload certificates are discovered). In these cases,
         the underlying transport will be reconfigured to use mTLS.
-
         Note: This function does nothing if the `aiohttp` library is not
         installed.
         Important: Calling this method will close any ongoing API requests associated
--- a/packages/google-auth/google/auth/aio/transport/sessions.py
+++ b/packages/google-auth/google/auth/aio/transport/sessions.py
@@ -206,7 +206,11 @@ class AsyncAuthorizedSession:
                             old_auth_request = self._auth_request
                             self._auth_request = AiohttpRequest(session=new_session)
 
-                            await old_auth_request.close()
+                            try:
+                                await old_auth_request.close()
+                            except Exception:
+                                # Suppress so it doesn't abort the mTLS configuration
+                                pass
                         else:
                             is_mtls = False
                             warnings.warn(
--- a/packages/google-auth/google/auth/aio/transport/sessions.py
+++ b/packages/google-auth/google/auth/aio/transport/sessions.py
@@ -156,6 +156,7 @@ class AsyncAuthorizedSession:
         (via GOOGLE_API_USE_CLIENT_CERTIFICATE=true) or auto-enabled (when the env
         variable is unset and workload certificates are discovered). In these cases,
         the underlying transport will be reconfigured to use mTLS.
+
         Note: This function does nothing if the `aiohttp` library is not
         installed.
         Important: Calling this method will close any ongoing API requests associated
@@ -181,18 +182,16 @@ class AsyncAuthorizedSession:
                     google.auth.transport._mtls_helper.check_use_client_cert
                 )
                 if not use_client_cert:
-                    self._is_mtls = False
                     return
 
                 try:
                     (
-                        self._is_mtls,
+                        is_mtls,
                         cert,
                         key,
                     ) = await mtls.get_client_cert_and_key(client_cert_callback)
 
-                    if self._is_mtls:
-                        self._cached_cert = cert
+                    if is_mtls:
                         ssl_context = await mtls._run_in_executor(
                             mtls.make_client_cert_ssl_context, cert, key
                         )
@@ -209,7 +208,7 @@ class AsyncAuthorizedSession:
 
                             await old_auth_request.close()
                         else:
-                            self._is_mtls = False
+                            is_mtls = False
                             warnings.warn(
                                 "Attempted to establish mTLS, but a custom async transport was provided. "
                                 "google-auth cannot automatically configure custom transports for mTLS. "
@@ -219,8 +218,13 @@ class AsyncAuthorizedSession:
                                 UserWarning,
                             )
 
+                    self._is_mtls = is_mtls
+                    if is_mtls:
+                        self._cached_cert = cert
+                    else:
+                        self._cached_cert = None
+
                 except Exception as caught_exc:
-                    self._is_mtls = False
                     new_exc = exceptions.MutualTLSChannelError(caught_exc)
                     raise new_exc from caught_exc
 
--- a/packages/google-auth/google/auth/transport/grpc.py
+++ b/packages/google-auth/google/auth/transport/grpc.py
@@ -280,14 +280,14 @@ def secure_authorized_channel(
 class SslCredentials:
     """Class for application default SSL credentials.
 
-    The client certificate usage (mutual TLS) is determined by the
-    `should_use_client_cert` helper. Client certificate will not be used
-    unless client certificate usage is enabled. This is true if the
-    `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly
-    set to `"true"`, or if the environment variable is unset/empty but a client
-    certificate configuration is found (e.g. via the `GOOGLE_API_CERTIFICATE_CONFIG`
-    environment variable containing a `"workload"` certificate configuration).
-    See https://google.aip.dev/auth/4114
+    Mutual TLS (mTLS) is enabled if either:
+    1. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly
+       set to `"true"`.
+    2. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is unset or empty,
+       but a valid workload certificate configuration is found (e.g., via the
+       `GOOGLE_API_CERTIFICATE_CONFIG` environment variable or the default gcloud config path).
+
+    See https://google.aip.dev/auth/4114 for client certificate discovery details.
 
     If client certificate usage is enabled, then for devices with endpoint
     verification support, a device certificate will be automatically loaded and
--- a/packages/google-auth/google/auth/transport/grpc.py
+++ b/packages/google-auth/google/auth/transport/grpc.py
@@ -329,7 +329,6 @@ class SslCredentials:
                     self._ssl_credentials = grpc.ssl_channel_credentials()
                     self._is_mtls = False
             except (exceptions.ClientCertError, OSError) as caught_exc:
-                self._is_mtls = False
                 new_exc = exceptions.MutualTLSChannelError(caught_exc)
                 raise new_exc from caught_exc
         else:
--- a/packages/google-auth/google/auth/transport/grpc.py
+++ b/packages/google-auth/google/auth/transport/grpc.py
@@ -281,6 +281,7 @@ class SslCredentials:
     """Class for application default SSL credentials.
 
     Mutual TLS (mTLS) is enabled if either:
+
     1. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly
        set to `"true"`.
     2. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is unset or empty,
--- a/packages/google-auth/google/auth/transport/grpc.py
+++ b/packages/google-auth/google/auth/transport/grpc.py
@@ -20,6 +20,7 @@ import logging
 
 from google.auth import exceptions
 from google.auth.transport import _mtls_helper
+from google.auth.transport import mtls
 from google.oauth2 import service_account
 
 try:
@@ -295,11 +296,7 @@ class SslCredentials:
         if not use_client_cert:
             self._is_mtls = False
         else:
-            # Load client SSL credentials.
-            metadata_path = _mtls_helper._check_config_path(
-                _mtls_helper.CONTEXT_AWARE_METADATA_PATH
-            )
-            self._is_mtls = metadata_path is not None
+            self._is_mtls = mtls.has_default_client_cert_source()
 
     @property
     def ssl_credentials(self):
--- a/packages/google-auth/google/auth/transport/grpc.py
+++ b/packages/google-auth/google/auth/transport/grpc.py
@@ -280,14 +280,18 @@ def secure_authorized_channel(
 class SslCredentials:
     """Class for application default SSL credentials.
 
-    The behavior is controlled by `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment
-    variable whose default value is `false`. Client certificate will not be used
-    unless the environment variable is explicitly set to `true`. See
-    https://google.aip.dev/auth/4114
+    The client certificate usage (mutual TLS) is determined by the
+    `should_use_client_cert` helper. Client certificate will not be used
+    unless client certificate usage is enabled. This is true if the
+    `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly
+    set to `"true"`, or if the environment variable is unset/empty but a client
+    certificate configuration is found (e.g. via the `GOOGLE_API_CERTIFICATE_CONFIG`
+    environment variable containing a `"workload"` certificate configuration).
+    See https://google.aip.dev/auth/4114
 
-    If the environment variable is `true`, then for devices with endpoint verification
-    support, a device certificate will be automatically loaded and mutual TLS will
-    be established.
+    If client certificate usage is enabled, then for devices with endpoint
+    verification support, a device certificate will be automatically loaded and
+    mutual TLS will be established.
     See https://cloud.google.com/endpoint-verification/docs/overview.
     """
 
@@ -316,11 +320,16 @@ class SslCredentials:
         """
         if self._is_mtls:
             try:
-                _, cert, key, _ = _mtls_helper.get_client_ssl_credentials()
-                self._ssl_credentials = grpc.ssl_channel_credentials(
-                    certificate_chain=cert, private_key=key
-                )
-            except exceptions.ClientCertError as caught_exc:
+                has_cert, cert, key, _ = _mtls_helper.get_client_ssl_credentials()
+                if has_cert:
+                    self._ssl_credentials = grpc.ssl_channel_credentials(
+                        certificate_chain=cert, private_key=key
+                    )
+                else:
+                    self._ssl_credentials = grpc.ssl_channel_credentials()
+                    self._is_mtls = False
+            except (exceptions.ClientCertError, OSError) as caught_exc:
+                self._is_mtls = False
                 new_exc = exceptions.MutualTLSChannelError(caught_exc)
                 raise new_exc from caught_exc
         else:
--- a/packages/google-auth/google/auth/transport/requests.py
+++ b/packages/google-auth/google/auth/transport/requests.py
@@ -445,14 +445,15 @@ class AuthorizedSession(requests.Session):
             google.auth.exceptions.MutualTLSChannelError: If mutual TLS channel
                 creation failed for any reason.
         """
+        self._is_mtls = False
+        self.mount("https://", requests.adapters.HTTPAdapter())
+
         use_client_cert = google.auth.transport._mtls_helper.check_use_client_cert()
         if not use_client_cert:
-            self._is_mtls = False
             return
         try:
             import OpenSSL
         except ImportError as caught_exc:
-            self._is_mtls = False
             new_exc = exceptions.MutualTLSChannelError(caught_exc)
             raise new_exc from caught_exc
 
@@ -475,7 +476,6 @@ class AuthorizedSession(requests.Session):
             OSError,
             OpenSSL.crypto.Error,
         ) as caught_exc:
-            self._is_mtls = False
             new_exc = exceptions.MutualTLSChannelError(caught_exc)
             raise new_exc from caught_exc
 
--- a/packages/google-auth/google/auth/transport/requests.py
+++ b/packages/google-auth/google/auth/transport/requests.py
@@ -443,14 +443,13 @@ class AuthorizedSession(requests.Session):
 
         Raises:
             google.auth.exceptions.MutualTLSChannelError: If mutual TLS channel
-                creation failed for any reason.
+                creation failed for any reason. The existing session state (such
+                as adapter mounts) remains unmodified if this error is raised.
         """
-        self._is_mtls = False
-        self.mount("https://", requests.adapters.HTTPAdapter())
-
         use_client_cert = google.auth.transport._mtls_helper.check_use_client_cert()
         if not use_client_cert:
             return
+
         try:
             import OpenSSL
         except ImportError as caught_exc:
@@ -459,17 +458,17 @@ class AuthorizedSession(requests.Session):
 
         try:
             (
-                self._is_mtls,
+                is_mtls,
                 cert,
                 key,
             ) = google.auth.transport._mtls_helper.get_client_cert_and_key(
                 client_cert_callback
             )
 
-            if self._is_mtls:
-                mtls_adapter = _MutualTlsAdapter(cert, key)
-                self._cached_cert = cert
-                self.mount("https://", mtls_adapter)
+            if is_mtls:
+                new_adapter = _MutualTlsAdapter(cert, key)
+            else:
+                new_adapter = requests.adapters.HTTPAdapter()
         except (
             exceptions.ClientCertError,
             ImportError,
@@ -478,6 +477,14 @@ class AuthorizedSession(requests.Session):
         ) as caught_exc:
             new_exc = exceptions.MutualTLSChannelError(caught_exc)
             raise new_exc from caught_exc
+
+        self.mount("https://", new_adapter)
+        self._is_mtls = is_mtls
+        if is_mtls:
+            self._cached_cert = cert
+        else:
+            if hasattr(self, "_cached_cert"):
+                del self._cached_cert
 
     def request(
         self,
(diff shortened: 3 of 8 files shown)


packages/google-auth/tests/transport/test_grpc.py
```diff
@@ -468,6 +468,41 @@ def test_get_client_ssl_credentials_success(
...
+
+        with mock.patch.dict(
+            os.environ, {environment_vars.GOOGLE_API_USE_CLIENT_CERTIFICATE: "true"}
+        ):
+            ssl_credentials = google.auth.transport.grpc.SslCredentials()
+
+        # If a workload certificate config exists on the device (and use_client_cert is true),
+        # is_mtls must be True and get_client_ssl_credentials should be invoked.
+        assert ssl_credentials.ssl_credentials is not None
+        assert ssl_credentials.is_mtls
```
parthea, line 506:
```suggestion
            ssl_credentials = google.auth.transport.grpc.SslCredentials()

            # If a workload certificate config exists on the device (and use_client_cert is true),
            # is_mtls must be True and get_client_ssl_credentials should be invoked.
            assert ssl_credentials.ssl_credentials is not None
            assert ssl_credentials.is_mtls
```

parthea (approved):

packages/google-auth/google/auth/transport/grpc.py
nbayati: Could we update the class docstring for `SslCredentials`? It still states that client certificates will not be used unless `GOOGLE_API_USE_CLIENT_CERTIFICATE` is explicitly set to `true`, which is outdated now that we support auto-enablement via `mtls.has_default_client_cert_source()`.

packages/google-auth/google/auth/transport/grpc.py
```diff
@@ -295,11 +296,7 @@ def __init__(self):
...
         else:
-            # Load client SSL credentials.
-            metadata_path = _mtls_helper._check_config_path(
-                _mtls_helper.CONTEXT_AWARE_METADATA_PATH
-            )
-            self._is_mtls = metadata_path is not None
+            self._is_mtls = mtls.has_default_client_cert_source()
 
     @property
     def ssl_credentials(self):
```
nbayati, line 307: There are two failure scenarios that could happen, and we need to address both:
- Case 1: Config exists but lacks workload keys/paths (No workload section)
There is a state mismatch in the `ssl_credentials` property when `has_default_client_cert_source()` is `True` but actual certificates cannot be loaded (e.g., config is empty or missing workload keys). 
Currently, the code discards the first boolean (`has_cert`) returned by `get_client_ssl_credentials()` and calls `grpc.ssl_channel_credentials(None, None)`. This silently establishes a standard one-way TLS connection, but `self._is_mtls` remains `True`. 
We should check `has_cert` and set `self._is_mtls = False` on failure:
```python
        if self._is_mtls:
            try:
                has_cert, cert, key, _ = _mtls_helper.get_client_ssl_credentials()
                if has_cert:
                    self._ssl_credentials = grpc.ssl_channel_credentials(
                        certificate_chain=cert, private_key=key
                    )
                else:
                    self._ssl_credentials = grpc.ssl_channel_credentials()
                    self._is_mtls = False
            except (exceptions.ClientCertError, OSError) as caught_exc:
                new_exc = exceptions.MutualTLSChannelError(caught_exc)
                raise new_exc from caught_exc
```

- Case 2: Config has a workload section, but the files it points to do not exist. To align with the async transport (sessions.py), we should catch OSError in the ssl_credentials property and wrap it in MutualTLSChannelError.
Currently, if the certificate files specified in the config do not exist on disk, a raw FileNotFoundError (which inherits from OSError) propagates uncaught out of get_client_ssl_credentials(). This crashes the client application on startup instead of raising a structured library error.

nbayati: We should also add unit tests to cover both cases.

packages/google-auth/tests/transport/test_grpc.py
```diff
@@ -468,6 +468,41 @@ def test_get_client_ssl_credentials_success(
...
+        mock_get_client_ssl_credentials.assert_called_once()
+        mock_ssl_channel_credentials.assert_called_once_with(
+            certificate_chain=PUBLIC_CERT_BYTES, private_key=PRIVATE_KEY_BYTES
+        )
+
     def test_get_client_ssl_credentials_without_client_cert_env(
```
nbayati, line 512: The new `grpc.py` logic correctly queries `check_use_client_cert()`. However, this exposes three existing test cases to host environment variables, causing them to fail on gLinux machines where SecureConnect/CAA is configured by default (as `CLOUDSDK_CONTEXT_AWARE_CERTIFICATE_CONFIG_FILE_PATH` points to `/etc/gcloud/certificate_config.json`):

* `test_get_client_ssl_credentials_without_client_cert_env`
* `test_secure_authorized_channel_adc_without_client_cert_env`
* `test_secure_authorized_channel_cert_callback_without_client_cert_env`

To isolate these tests and make them pass consistently on developer machines, we should explicitly patch the environment in these test cases using the established `mock.patch.dict` pattern to set `GOOGLE_API_USE_CLIENT_CERTIFICATE` to `"false"`:

```python
        with mock.patch.dict(
            os.environ, {environment_vars.GOOGLE_API_USE_CLIENT_CERTIFICATE: "false"}
        ):

File: packages/google-auth/google/auth/aio/transport/sessions.py [hunk 4]
```diff
@@ -207,6 +207,8 @@ async def _do_configure():
                             self._auth_request = AiohttpRequest(session=new_session)
 
                             await old_auth_request.close()
+                        else:
+                            self._is_mtls = False
 
                 except (
```
nbayati: If an exception is raised during mTLS configuration (e.g. `ClientCertError` or `ImportError`), the configuration fails and throws `MutualTLSChannelError`. However, since `self._is_mtls` was already set to `True` earlier in the function, it will remain `True` even after the exception is raised. 
If the caller catches this exception and attempts to reuse the session, this stale state flag can cause issues. It would be safer to reset `self._is_mtls = False` inside the `except` block before raising the exception:

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -357,6 +357,7 @@ def configure_mtls_channel(self, client_cert_callback=None):
                 self._cached_cert = cert
             else:
                 self.http = _make_default_http()
+                self._is_mtls = False
         except (
```
nbayati, line 359: Same as sessions, if an exception is raised during mTLS configuration (e.g. `ClientCertError` or `ImportError`), the configuration fails and throws `MutualTLSChannelError`. However, since `self._is_mtls` was already set to `True` earlier in the function, it will remain `True` even after the exception is raised. 
If the caller catches this exception and attempts to reuse the session, this stale state flag can cause issues. It would be safer to reset `self._is_mtls = False` inside the `except` block before raising the exception. 

The same issue exists in `packages/google-auth/google/auth/transport/requests.py` (line 471). I think we should update that too, for consistency if we want to ensure all transports manage the state cleanly.

nbayati (commented):
nit: Since this PR resolves existing bugs rather than introducing new features, the PR title prefix should be `fix(auth)` instead of `feat(auth)`.

packages/google-auth/google/auth/transport/grpc.py
vverman: Done!

packages/google-auth/google/auth/transport/grpc.py
vverman: Done! We cover both the cases also, added unit tests.

packages/google-auth/tests/transport/test_grpc.py
vverman: Done!

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done, we set is_mtls to False in except block.

packages/google-auth/google/auth/transport/urllib3.py
vverman: Done! Thanks for the catch!

packages/google-auth/google/auth/aio/transport/sessions.py
```diff
@@ -207,12 +209,15 @@ async def _do_configure():
...
                     exceptions.ClientCertError,
                     ImportError,
                     OSError,
                 ) as caught_exc:
+                    self._is_mtls = False
```
nbayati: The `_do_configure()` method in the task calls `make_client_cert_ssl_context`, which raises `exceptions.TransportError` on failure. Since `TransportError` is not a subclass of `ClientCertError`, `ImportError`, or `OSError`, this exception bubbles out uncaught. Consequently, `self._is_mtls` remains set to `True` (as unpacked on line 194), causing the transport state to report mTLS usage incorrectly.

Suggested fix:
```python
                except Exception as caught_exc:
                    self._is_mtls = False
                    new_exc = exceptions.MutualTLSChannelError(caught_exc)
                    raise new_exc from caught_exc
```

packages/google-auth/google/auth/aio/transport/sessions.py
```diff
@@ -207,12 +209,15 @@ async def _do_configure():
                             self._auth_request = AiohttpRequest(session=new_session)
 
                             await old_auth_request.close()
+                        else:
+                            self._is_mtls = False
 
                 except (
                     exceptions.ClientCertError,
                     ImportError,
```
nbayati: Invoking `close()` while `self._mtls_init_task` is still initializing will only close the original `_auth_request`. Once the background task resumes and finishes, it mounts a newly created `ClientSession` that is never closed, leading to socket leaks.

Suggested fix:
```python
    async def close(self) -> None:
        if self._mtls_init_task and not self._mtls_init_task.done():
            self._mtls_init_task.cancel()
            try:
                await self._mtls_init_task
            except asyncio.CancelledError:
                pass
        await self._auth_request.close()
```

packages/google-auth/google/auth/transport/requests.py
```diff
@@ -471,8 +474,10 @@ def configure_mtls_channel(self, client_cert_callback=None):
         except (
             exceptions.ClientCertError,
             ImportError,
+            OSError,
             OpenSSL.crypto.Error,
         ) as caught_exc:
+            self._is_mtls = False
```
nbayati: If an exception or fallback happens during subsequent `configure_mtls_channel()` calls, `self._is_mtls` is set to `False`, but the previously mounted `_MutualTlsAdapter` is not unmounted or replaced, leaving the session using the old mTLS configuration.

Suggested fix:
```python
        if self._is_mtls:
            mtls_adapter = _MutualTlsAdapter(cert, key)
            self._cached_cert = cert
            self.mount("https://", mtls_adapter)
        else:
            from requests.adapters import HTTPAdapter
            self.mount("https://", HTTPAdapter())
```

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -357,11 +360,14 @@ def configure_mtls_channel(self, client_cert_callback=None):
...
             exceptions.ClientCertError,
             ImportError,
+            OSError,
             OpenSSL.crypto.Error,
         ) as caught_exc:
+            self._is_mtls = False
```
nbayati: If an exception is raised in `configure_mtls_channel()`, `self._is_mtls` is set to `False`, but the old connection pool manager (`self.http`) is not reset to the default pool manager, leaving it configured with mTLS.

Suggested fix:
```python
        except (
            exceptions.ClientCertError,
            ImportError,
            OSError,
            OpenSSL.crypto.Error,
        ) as caught_exc:
            self._is_mtls = False
            self.http = _make_default_http()
            new_exc = exceptions.MutualTLSChannelError(caught_exc)
            raise new_exc from caught_exc
```

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
```diff
@@ -140,3 +141,66 @@ def mock_callback():
...
+        ) as mock_make_context:
+            mock_exists.return_value = True
+            mock_helper.return_value = (True, b"fake_cert_data", b"fake_key_data")
+
+            mock_context = mock.Mock(spec=ssl.SSLContext)
+            mock_make_context.return_value = mock_context
+
+            mock_creds = mock.AsyncMock(spec=credentials.Credentials)
+            mock_auth_request = mock.AsyncMock(spec=transport.Request)
+            session = sessions.AsyncAuthorizedSession(
```
nbayati, line 186: The newly added tests `test_configure_mtls_channel_custom_request` and `test_configure_mtls_channel_exception_resets_flag` instantiate sessions but do not close them, leaving unclosed `ClientSession` objects and leaking resources.

Suggested fix:
Add `await session.close()` at the end of both tests.

nbayati (commented):
Thank you for the PR! I have completed a review of the changes. Here are some comments regarding potential state leakage, resource leaks, and mismatches when configuration exceptions occur.

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done!

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done!

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
vverman: Done! Added `await session.close()` at the end of a few more tests as well to avoid test-isolation-breakages.

packages/google-auth/google/auth/transport/urllib3.py
vverman: Done! Thanks for this catch.

packages/google-auth/google/auth/transport/requests.py
vverman: Done! Thanks for this catch.

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -335,12 +337,13 @@ def configure_mtls_channel(self, client_cert_callback=None):
             google.auth.exceptions.MutualTLSChannelError: If mutual TLS channel
                 creation failed for any reason.
         """
+        self._is_mtls = False
+        self.http = _make_default_http()
```
nbayati: This preemptively overwrites the user's custom `self.http` client and resets `self._is_mtls = False` at the very top of the method. This breaks custom configurations even when mTLS is disabled, and destroys the state of already-active sessions if a subsequent certificate rotation/refresh fails midway.

To prevent silently destroying configurations and session state, we should make the configuration atomic:

1. Check `check_use_client_cert()` first. If it's `False`, exit immediately without touching `self.http`.

2. Perform the certificate lookup and pool creation locally (into local variables).

3. Only assign the new pool to `self.http` (and update `self._is_mtls` / `self._request.http`) at the end of the method, assigning either the new mTLS pool or the default pool depending on the lookup result.

4. If an exception occurs during lookup, let it propagate *without* modifying the session state. 

Note: The new unit test test_configure_mtls_channel_subsequent_failure will need to be updated to reflect this. Instead of asserting that a subsequent failure resets the channel to standard TLS, it should assert that the session retains its previous mTLS state.

packages/google-auth/google/auth/transport/requests.py
```diff
@@ -445,9 +447,11 @@ def configure_mtls_channel(self, client_cert_callback=None):
             google.auth.exceptions.MutualTLSChannelError: If mutual TLS channel
                 creation failed for any reason.
         """
+        self._is_mtls = False
+        self.mount("https://", requests.adapters.HTTPAdapter())
+
```
nbayati: Same issue as packages/google-auth/google/auth/transport/urllib3.py.

packages/google-auth/google/auth/transport/urllib3.py
```diff
@@ -313,13 +313,15 @@ def __init__(
...
         """Configures mutual TLS channel using the given client_cert_callback or
-        application default SSL credentials. The behavior is controlled by
-        `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable.
-        (1) If the environment variable value is `true`, the function returns True
-        if the channel is mutual TLS and False otherwise. The `http` provided
-        in the constructor will be overwritten.
-        (2) If the environment variable is not set or `false`, the function does
-        nothing and it always return False.
+        application default SSL credentials. The behavior is determined by the
+        `should_use_client_cert` helper.
```
nbayati: A couple of minor things in the new docstring:

1. Tt references should_use_client_cert helper, but the actual function name in _mtls_helper is check_use_client_cert.

2. Once the configuration is refactored to be atomic, we should clarify in the Raises section that if MutualTLSChannelError is thrown, the instance state (like the http client or mounts) is left unmodified.

packages/google-auth/google/auth/transport/grpc.py
```diff
@@ -319,11 +320,16 @@ def ssl_credentials(self):
...
+                has_cert, cert, key, _ = _mtls_helper.get_client_ssl_credentials()
+                if has_cert:
+                    self._ssl_credentials = grpc.ssl_channel_credentials(
+                        certificate_chain=cert, private_key=key
+                    )
+                else:
+                    self._ssl_credentials = grpc.ssl_channel_credentials()
+                    self._is_mtls = False
+            except (exceptions.ClientCertError, OSError) as caught_exc:
+                self._is_mtls = False
```
nbayati: If a transient error occurs while loading the certificates (e.g., an `OSError` due to a file lock or a startup race condition), this `except` block permanently sets `self._is_mtls = False`. 

If the caller catches the resulting `MutualTLSChannelError` and attempts a retry using this same `SslCredentials` instance, the next access to `.ssl_credentials` will see `self._is_mtls` as `False`. It will bypass the certificate loading entirely and silently fall back to standard, non-mTLS credentials (`grpc.ssl_channel_credentials()`). 

This means the retry might "succeed" without mTLS, leading to obscure 401s or certificate mismatch errors from the server down the line, which masks the true root cause.

Let's remove the ` self._is_mtls = False` from this `except` block. This ensures that if a transient failure happens, the state remains intact so subsequent retries can successfully re-attempt loading the certificates. If the error is persistent, it's safer to fail closed rather than silently falling back to standard TLS.

nbayati (changes requested):

packages/google-auth/google/auth/transport/grpc.py
vverman: Done! Thank you for the catch!

packages/google-auth/google/auth/transport/urllib3.py
vverman: Done

packages/google-auth/google/auth/transport/requests.py
vverman: Done!

packages/google-auth/google/auth/transport/urllib3.py
vverman: Good catch, I had overlooked the custom set transport configs and the midway failure of cert-rotation.

Made the changes.

packages/google-auth/google/auth/aio/transport/sessions.py
```diff
@@ -156,7 +156,6 @@ async def configure_mtls_channel(self, client_cert_callback=None):
         (via GOOGLE_API_USE_CLIENT_CERTIFICATE=true) or auto-enabled (when the env
         variable is unset and workload certificates are discovered). In these cases,
         the underlying transport will be reconfigured to use mTLS.
-
```
nbayati, line 159: Nit: Let's keep the blank line here. Docstring parsers and documentation generators (like Sphinx) require blank lines between blocks of text to render them as separate paragraphs, and keeping it helps with general readability.

packages/google-auth/google/auth/transport/grpc.py
```diff
@@ -279,14 +280,18 @@ def my_client_cert_callback():
...
-    support, a device certificate will be automatically loaded and mutual TLS will
-    be established.
+    The client certificate usage (mutual TLS) is determined by the
+    `should_use_client_cert` helper. Client certificate will not be used
+    unless client certificate usage is enabled. This is true if the
+    `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly
+    set to `"true"`, or if the environment variable is unset/empty but a client
+    certificate configuration is found (e.g. via the `GOOGLE_API_CERTIFICATE_CONFIG`
+    environment variable containing a `"workload"` certificate configuration).
+    See https://google.aip.dev/auth/4114
```
nbayati: The docstring incorrectly references the `should_use_client_cert` helper. The actual helper function that's being used is `check_use_client_cert`.

Also, the docstring is a bit hard to follow. What do you think about rephrasing it to something like

```
    Mutual TLS (mTLS) is enabled if either:
    1. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is explicitly 
       set to `"true"`.
    2. The `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is unset or empty, 
       but a valid workload certificate configuration is found (e.g., via the 
       `GOOGLE_API_CERTIFICATE_CONFIG` environment variable or the default gcloud config path).

    See https://google.aip.dev/auth/4114 for client certificate discovery details.

File: packages/google-auth/google/auth/transport/grpc.py [hunk 17]
```diff
@@ -295,11 +300,7 @@ def __init__(self):
         if not use_client_cert:
             self._is_mtls = False
         else:
-            # Load client SSL credentials.
-            metadata_path = _mtls_helper._check_config_path(
-                _mtls_helper.CONTEXT_AWARE_METADATA_PATH
-            )
-            self._is_mtls = metadata_path is not None
+            self._is_mtls = mtls.has_default_client_cert_source()
```
nbayati, line 304: This helper ignores CLOUDSDK_CONTEXT_AWARE_CERTIFICATE_CONFIG_FILE_PATH (the context-aware fallback configuration path), so we will have a mismatch state where use_client_cert will return true but has_default_client_cert_source misses that path. 
It's an existing issue with `has_default_client_cert_source()` though, so probably out of scope for this PR. 
I'll document this so we can fix it in a future PR.

packages/google-auth/google/auth/aio/transport/sessions.py
nbayati: To keep sessions.py consistent with the atomic pattern implemented in urllib3.py and requests.py (which preserves the previous session state if a new configuration fails), we should refactor configure_mtls_channel to use a local is_mtls variable.

1. In _do_configure(), unpack get_client_cert_and_key into a local is_mtls variable.
2. Commit the state changes to self._is_mtls and self._cached_cert only at the very end of the successful try block.
3. Remove self._is_mtls = False from the early return (line 184) and the except Exception: block (line 223).

This guarantees that failed configuration attempts or check-disablements do not corrupt the existing mTLS session state.

packages/google-auth/tests/transport/test_grpc.py
```diff
@@ -468,19 +474,136 @@ def test_get_client_ssl_credentials_success(
...
         self,
         mock_check_config_path,
         mock_load_json_file,
         mock_get_client_ssl_credentials,
         mock_ssl_channel_credentials,
     ):
-        # Test client cert won't be used if GOOGLE_API_USE_CLIENT_CERTIFICATE is not set.
-        ssl_credentials = google.auth.transport.grpc.SslCredentials()
+        with mock.patch.dict(
+            os.environ, {environment_vars.GOOGLE_API_USE_CLIENT_CERTIFICATE: "false"}
```
nbayati, line 520: The test suite explicitly verifies the behavior when GOOGLE_API_USE_CLIENT_CERTIFICATE is 'true' and 'false', but it lacks a case for the new 'auto-enablement' logic.

Please add a test case explicitly verifying that SslCredentials enables mTLS when this environment variable is completely unset/absent, but check_use_client_cert() evaluates to True.

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
```diff
@@ -135,11 +139,12 @@ def mock_callback():
             await session.configure_mtls_channel(client_cert_callback=mock_callback)
 
             assert session._is_mtls is True
```
nbayati, line 156: This success case verifies that the _is_mtls flag correctly flips to True, but it doesn't verify the actual side effect.

Please mock aiohttp.TCPConnector and add an assertion to verify it is actually called with ssl=mock_context so we can be sure the new certificates are successfully injected into the session

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
```diff
@@ -161,5 +166,69 @@ async def test_configure_mtls_channel_custom_request(self):
             session = sessions.AsyncAuthorizedSession(
                 mock_creds, auth_request=mock_auth_request
             )
+
             await session.configure_mtls_channel()
+
+            # If the request handler is not an AiohttpRequest, the library cannot configure
+            # the connection to use mTLS, so _is_mtls must be False to reflect this unconfigured state.
+            assert session._is_mtls is False
```
nbayati, line 195: This test correctly verifies that _is_mtls remains False, but it doesn't verify that the debugging aid works.

Please wrap the await session.configure_mtls_channel() call with pytest.warns(UserWarning, match="Attempted to establish mTLS") to ensure the warning is properly emitted during fallback.

nbayati (changes requested):
Great work Pranav! Before we merge, there are just a few minor cleanups and test coverage gaps we need to address to ensure these fixes remain robust long-term.

packages/google-auth/google/auth/transport/grpc.py
vverman: Corrected the docstring.

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done!

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done, replicated the same pattern as urllib3.py

packages/google-auth/google/auth/transport/grpc.py
vverman: Done, I'll leave this open!

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
vverman: Done, added the mock to the test.

packages/google-auth/tests/transport/aio/test_sessions_mtls.py
vverman: Done added this scenario to the test!

packages/google-auth/tests/transport/test_grpc.py
vverman: Added the test!

packages/google-auth/google/auth/aio/transport/sessions.py
```diff
@@ -210,7 +208,7 @@ async def _do_configure():
 
                             await old_auth_request.close()
```
nbayati: The atomic state management here is very close, but there is one edge case that could still lead to a mismatch state during a successful mTLS initialization.

If `await old_auth_request.close()` raises an exception (e.g., an internal `aiohttp` error), execution will immediately jump to the outer `except` block. This skips the `self._is_mtls = is_mtls` assignment entirely. Because the `request()` method explicitly suppresses background task exceptions, the session would then proceed using the newly assigned mTLS `_auth_request`, but the `_is_mtls` flag would remain `False`.

Since failing to close the old transport shouldn't abort the entire mTLS configuration anyway, we should wrap the `close()` call in a `try/except` block:

```python
                            old_auth_request = self._auth_request
                            self._auth_request = AiohttpRequest(session=new_session)

                            try:
                                await old_auth_request.close()
                            except Exception:
                                # Suppress so it doesn't abort the mTLS configuration
                                pass
```

Could you also add a quick unit test to lock in this behavior? You can mock session._auth_request.close = mock.AsyncMock(side_effect=Exception('Mock error')) and verify that await session.configure_mtls_channel() still successfully completes with session.is_mtls resolving to True and without raising a MutualTLSChannelError.

nbayati (approved):

packages/google-auth/google/auth/aio/transport/sessions.py
vverman: Done! Added a unit test as well.

PR #17390, opened 2026-06-08
feat(auth): make RAB feature production ready

nbayati:
This PR resolves issues identified during verification of gcloud Regional Access Boundary (RAB) flows and enables RAB verification by default:

*   Removes the client-side environment variable feature gate (`GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED`) to execute RAB lookups by default across standard credential classes.
*   Updates the Python auth SDK to recognize mTLS regional endpoints (`.rep.mtls.googleapis.com`), bypassing redundant RAB lookups on secure transport boundaries.
*   Defers Service Account impersonation setup until HTTP request execution before_request, propagating active cached tokens downward onto the inner credential to guarantee that access tokens restored across external CLI entrypoints correctly delegate regional access boundary (RAB) lookups to target Service Account endpoints without forcing redundant STS network renewal.


macastelaz (approved):

packages/google-auth/google/auth/external_account.py
```diff
@@ -213,6 +213,44 @@ def __init__(
                 "credentials"
             )
 
+        # Initialize impersonated credentials immediately upon creation.
```
lsirac: We shouldn't initialize impersonated credentials inside the base constructor because it introduces a constructor initialization race condition.

When a subclass is instantiated, it calls super().__init__() before setting up its own local attributes. Because the parent class __init__ immediately calls _initialize_impersonated_credentials(), it invokes the subclass's overridden _constructor_args() method. Since the subclass constructor hasn't finished running yet, _constructor_args() will attempt to access subclass variables that do not exist yet, resulting in an AttributeError

packages/google-auth/google/auth/aws.py
```diff
@@ -660,6 +660,8 @@ def __init__(
             :meth:`from_file` or
             :meth:`from_info` are used instead of calling the constructor directly.
         """
+        if aws_security_credentials_supplier is not None:
```
lsirac: I'm assuming this was added as a workaround to the bug introduced by initializing impersonated creds in the constructor and can be removed?

packages/google-auth/google/auth/identity_pool.py
```diff
@@ -324,6 +324,8 @@ def __init__(
             :meth:`from_info` are used instead of calling the constructor directly.
         """
 
+        if subject_token_supplier is not None:
```
lsirac: Same here

packages/google-auth/google/auth/external_account.py
```diff
@@ -213,6 +213,44 @@ def __init__(
...
+        if (
+            self._should_initialize_impersonated_credentials()
+            and self.service_account_email
+        ):
+            self._impersonated_credentials = self._initialize_impersonated_credentials()
+            self._impersonated_credentials.expiry = None
+            self._rab_manager = self._impersonated_credentials._rab_manager
+
+    @property
+    def token(self):
```
lsirac: Is this a necessary change?

Changing `token` and `expiry` from normal variables into `@property` functions introduces a hidden unpickling bug that will silently wipe out saved credentials.

When older saved (pickled) credentials are loaded back into memory, their tokens are restored intact into the object's internal dictionary under the standard `'token'` and `'expiry'` keys. However, because Python `@property` functions take priority over the internal dictionary and specifically look for `_token` and `_expiry`, accessing `creds.token` will return `None`. This silently drops valid saved tokens and forces unnecessary, expensive re-authentication network calls.

packages/google-auth/google/auth/credentials.py
```diff
@@ -446,9 +446,13 @@ def _is_regional_endpoint(self, url):
         try:
             # Do not perform a lookup if the request is for a regional endpoint.
             hostname = urlparse(url).hostname
-            if hostname and (
-                hostname.endswith(".rep.googleapis.com")
-                or hostname.endswith(".rep.sandbox.googleapis.com")
+            if hostname and hostname.endswith(
```
lsirac, line 449: Are you sure this list is exhaustive?

e.g. what about PSC?  (When an enterprise customer sets up Private Service Connect for a regional service, the hostname routes internally as `.p.rep.googleapis.com` (or `.p.rep.mtls.googleapis.com`)).

Could be more as well

lsirac (changes requested):
While the new unit tests cover the basic happy paths, there are a few critical testing gaps around backward compatibility, error handling, and subclassing that should be addressed before merging:

> 1. **Backward Compatibility with Pickled States**:
>    * **The Gap**: No test verifies unpickling a credential object serialized in a previous release.
>    * **The Risk**: Because `token` and `expiry` are now properties (data descriptors), they override instance `__dict__` lookups. Older unpickled credentials (which store `'token'` and `'expiry'` as plain keys) will have their cached tokens silently ignored, forcing redundant STS network calls.
>    * **Recommendation**: Add a test that unpickles a legacy state and asserts the token is preserved.
> 
> 2. **Extensibility & Custom Subclassing**:
>    * **The Gap**: No test verifies that external developers can safely subclass these credentials.
>    * **The Risk**: Triggering immediate impersonation in the base constructor calls `_constructor_args()` before the subclass has finished setting up. This causes a fatal `AttributeError` on subclass instantiation.
>    * **Recommendation**: Add a test verifying instantiation of a mock custom subclass.
> 
> 3. **Error-Handling on Misconfiguration**:
>    * **The Gap**: No test verifies that invalid/misconfigured setups fail gracefully.
>    * **The Risk**: Misconfiguring the credential source and supplier triggers the immediate copying flow, crashing with a confusing, internal `AttributeError` instead of raising a clean `exceptions.InvalidValue` validation error.
>    * **Recommendation**: Add a test asserting that invalid configurations raise `exceptions.InvalidValue`.
> 
> 4. **Programmatic Suppliers in `from_info`**:
>    * **The Gap**: No test verifies passing programmatic suppliers to the static `from_info` method.
>    * **The Risk**: `from_info` in both `aws.py` and `identity_pool.py` unconditionally updates `kwargs` with `None` from the JSON dictionary, silently discarding the user's programmatic supplier.
>    * **Recommendation**: Add a test calling `Credentials.from_info(info, supplier=my_callback)` and assert the callback retention.callback)` and assert the check that the callback is preserved supplier works.

packages/google-auth/google/auth/credentials.py
nbayati: From my research, I think this is the only list we have to worry about. PSC regional endpoints (.p.rep.googleapis.com and .p.rep.mtls.googleapis.com) already evaluate correctly under this check. Because .rep.googleapis.com and .rep.mtls.googleapis.com are direct string suffixes of the PSC hostnames, calling hostname.endswith() returns True automatically without us needing to add more entries to the list.

packages/google-auth/google/auth/aws.py
nbayati: Yes, I just reverted it!

packages/google-auth/google/auth/external_account.py
nbayati: Moved the impersonated initialization to before_request. I used a script to reproduce the issue locally and verify that it was fixed.

packages/google-auth/google/auth/external_account.py
nbayati: Yes, good catch. I reverted the @property wrappers entirely so token and expiry stay normal variables.
Instead, now we copy the saved token down to the Service Account object on demand before making HTTP requests.

nbayati:
> * **Backward Compatibility with Pickled States**:
>   
>   * **The Gap**: No test verifies unpickling a credential object serialized in a previous release.
>   * **The Risk**: Because `token` and `expiry` are now properties (data descriptors), they override instance `__dict__` lookups. Older unpickled credentials (which store `'token'` and `'expiry'` as plain keys) will have their cached tokens silently ignored, forcing redundant STS network calls.
>   * **Recommendation**: Add a test that unpickles a legacy state and asserts the token is preserved.
> * **Extensibility & Custom Subclassing**:
>   
>   * **The Gap**: No test verifies that external developers can safely subclass these credentials.
>   * **The Risk**: Triggering immediate impersonation in the base constructor calls `_constructor_args()` before the subclass has finished setting up. This causes a fatal `AttributeError` on subclass instantiation.
>   * **Recommendation**: Add a test verifying instantiation of a mock custom subclass.
> * **Error-Handling on Misconfiguration**:
>   
>   * **The Gap**: No test verifies that invalid/misconfigured setups fail gracefully.
>   * **The Risk**: Misconfiguring the credential source and supplier triggers the immediate copying flow, crashing with a confusing, internal `AttributeError` instead of raising a clean `exceptions.InvalidValue` validation error.
>   * **Recommendation**: Add a test asserting that invalid configurations raise `exceptions.InvalidValue`.
> * **Programmatic Suppliers in `from_info`**:
>   
>   * **The Gap**: No test verifies passing programmatic suppliers to the static `from_info` method.
>   * **The Risk**: `from_info` in both `aws.py` and `identity_pool.py` unconditionally updates `kwargs` with `None` from the JSON dictionary, silently discarding the user's programmatic supplier.
>   * **Recommendation**: Add a test calling `Credentials.from_info(info, supplier=my_callback)` and assert the callback retention.callback)` and assert the check that the callback is preserved supplier works.

Gaps 1, 2, and 3 are not relevant anymore since I changed the way we were fixing the gcloud impersonation RAB lookup bug. I added checks and tests for gap 4 though!

packages/google-auth/google/auth/external_account.py
```diff
@@ -444,6 +444,17 @@ def _maybe_start_regional_access_boundary_refresh(self, request, url):
                 HTTP requests.
             url (str): The URL of the request.
         """
+        if (
```
lsirac: In multi-threaded applications sharing a single `Credentials` instance, multiple concurrent requests executing `before_request` when `_impersonated_credentials` is `None` can concurrently evaluate `_should_initialize_impersonated_credentials()` as `True`. Since there is no thread lock protecting this lazy-initialization block, the assignment of `_impersonated_credentials` and its corresponding `_rab_manager` is non-atomic.

A thread can be preempted after `self._impersonated_credentials` is set but before `self._rab_manager` is updated. A concurrent thread entering this block will skip initialization (seeing `_impersonated_credentials` is not `None`) but dispatch the request using the old, uninitialized outer `_rab_manager`. This causes outgoing requests to bypass the regional access boundary configuration and omit required headers (such as `x-allowed-locations`).

To resolve this, wrap the lazy-initialization block in a thread lock (e.g., `self._impersonation_lock` initialized in `__init__`) and perform atomic local variable assignments before binding to instance attributes:

```python
if getattr(self, "_impersonated_credentials", None) is None and self.service_account_email:
    with self._impersonation_lock:
        if self._impersonated_credentials is None:
            impersonated = self._initialize_impersonated_credentials()
            if getattr(self, "token", None):
                impersonated.token = self.token
            if getattr(self, "expiry", None):
                impersonated.expiry = self.expiry
            self._impersonated_credentials = impersonated
            self._rab_manager = impersonated._rab_manager

**Review by lsirac — CHANGES_REQUESTED** [review]

**Comment by lsirac** [comment]
> > * **Backward Compatibility with Pickled States**:
> >   
> >   * **The Gap**: No test verifies unpickling a credential object serialized in a previous release.
> >   * **The Risk**: Because `token` and `expiry` are now properties (data descriptors), they override instance `__dict__` lookups. Older unpickled credentials (which store `'token'` and `'expiry'` as plain keys) will have their cached tokens silently ignored, forcing redundant STS network calls.
> >   * **Recommendation**: Add a test that unpickles a legacy state and asserts the token is preserved.
> > * **Extensibility & Custom Subclassing**:
> >   
> >   * **The Gap**: No test verifies that external developers can safely subclass these credentials.
> >   * **The Risk**: Triggering immediate impersonation in the base constructor calls `_constructor_args()` before the subclass has finished setting up. This causes a fatal `AttributeError` on subclass instantiation.
> >   * **Recommendation**: Add a test verifying instantiation of a mock custom subclass.
> > * **Error-Handling on Misconfiguration**:
> >   
> >   * **The Gap**: No test verifies that invalid/misconfigured setups fail gracefully.
> >   * **The Risk**: Misconfiguring the credential source and supplier triggers the immediate copying flow, crashing with a confusing, internal `AttributeError` instead of raising a clean `exceptions.InvalidValue` validation error.
> >   * **Recommendation**: Add a test asserting that invalid configurations raise `exceptions.InvalidValue`.
> > * **Programmatic Suppliers in `from_info`**:
> >   
> >   * **The Gap**: No test verifies passing programmatic suppliers to the static `from_info` method.
> >   * **The Risk**: `from_info` in both `aws.py` and `identity_pool.py` unconditionally updates `kwargs` with `None` from the JSON dictionary, silently discarding the user's programmatic supplier.
> >   * **Recommendation**: Add a test calling `Credentials.from_info(info, supplier=my_callback)` and assert the callback retention.callback)` and assert the check that the callback is preserved supplier works.
> 
> Gaps 1, 2, and 3 are not relevant anymore since I changed the way we were fixing the gcloud impersonation RAB lookup bug. I added checks and tests for gap 4 though!

Awesome, thank you for adding the checks and tests for gap 4! 

re: the other gaps, let's still add these tests to prevent regressions in the future

File: packages/google-auth/google/auth/external_account.py (hunk 6 above)
Comment (nbayati) [reply packages/google-auth/google/auth/external_account.py]: Great catch! I've implemented the lock, along with pickling support.

File: packages/google-auth/google/auth/environment_vars.py [hunk 7]
```diff
@@ -105,9 +105,6 @@
 AWS_REGION = "AWS_REGION"
 AWS_DEFAULT_REGION = "AWS_DEFAULT_REGION"
 
-GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED = "GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED"
-"""Environment variable controlling whether to enable trust boundary feature.
-The default value is false. Users have to explicitly set this value to true."""
```
daniel-sanche, line 110: This looks like a publicly accessible value. I don't think it's safe to remove, in case someone imported it. Can we just leave it with a comment that it's deprecated?

lsirac (approved):

packages/google-auth/google/auth/environment_vars.py
nbayati: Done!

daniel-sanche (approved):
LGTM

